PR #2998, opened 2024-12-03
RemoteNet -> ExistingNet

ndr-ds:
## Motivation

I think I originally implemented this, it was originally running tests against the devnet on GCP. I don't remember what the name was then.
Regardless, now this has changed, and it actually just connects to any existing/running network that has a faucet.
I thought `RunningNet` sounded weird, `ExistingNet` sounded better. A `WithFaucet` suffix or something like that also sounded wrong. So I landed on `ExistingNet`.

## Proposal

Rename all remote net things into existing net. The feature for this also had kubernetes related stuff as a dependency, even though it uses nothing kubernetes related.
So removed that.

## Test Plan

CI

## Release Plan

- Nothing to do / These changes follow the usual release cycle.

PR #3067, opened 2024-12-20
ExistingNet -> NetWithFaucet

ndr-ds:
## Motivation

As much as `ExistingNet` is better than the previous name, it's still not great.

## Proposal

`NetWithFaucet` seems better and more descriptive of the network properties.

## Test Plan

CI

## Release Plan

### DIFF
--- a/linera-service/Cargo.toml
+++ b/linera-service/Cargo.toml
@@ -54,7 +54,7 @@ kubernetes = [
     "dep:pathdiff",
     "dep:fs_extra",
 ]
-existing-net = []
+faucet-net = []
 metrics = ["prometheus", "linera-base/metrics", "linera-client/metrics"]
 storage-service = ["linera-client/storage-service", "linera-storage-service"]
 
--- a/linera-service/tests/linera_net_tests.rs
+++ b/linera-service/tests/linera_net_tests.rs
@@ -7,7 +7,7 @@
     feature = "scylladb",
     feature = "storage-service",
     feature = "kubernetes",
-    feature = "existing-net"
+    feature = "faucet-net"
 ))]
 
 mod common;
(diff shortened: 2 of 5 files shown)


ma2bd:
It makes more sense but it doesn't end up with "net" like the others. I still prefer the old name remote-net.

ndr-ds:
Maybe `FaucetNet`?

linera-service/tests/linera_net_tests.rs
```diff
@@ -377,7 +377,7 @@ impl AmmApp {
 #[cfg_attr(feature = "scylladb", test_case(LocalNetConfig::new_test(Database::ScyllaDb, Network::Grpc) ; "scylladb_grpc"))]
 #[cfg_attr(feature = "dynamodb", test_case(LocalNetConfig::new_test(Database::DynamoDb, Network::Grpc) ; "aws_grpc"))]
 #[cfg_attr(feature = "kubernetes", test_case(SharedLocalKubernetesNetTestingConfig::new(Network::Grpc, BuildArg::Build) ; "kubernetes_grpc"))]
-#[cfg_attr(feature = "existing-net", test_case(ExistingNetTestingConfig::new(None) ; "existing_net_grpc"))]
+#[cfg_attr(feature = "net-with-faucet", test_case(NetWithFaucetTestingConfig::new(None) ; "net_with_faucet_grpc"))]
```
ma2bd, line 380:
```
#[cfg_attr(feature = "faucet", test_case(FaucetNetTestingConfig::new(None) ; "faucet_grpc"))]
```

linera-service/Cargo.toml
```diff
@@ -54,7 +54,7 @@ kubernetes = [
     "dep:pathdiff",
     "dep:fs_extra",
 ]
-existing-net = []
+net-with-faucet = []
```
ma2bd, line 57: `faucet = []`

ndr-ds: Wouldn't `faucet-net` be better here? Since this isn't general `faucet` functionality, but specific to the faucet net 🤔 I'll do `faucet-net`, but I can switch to `faucet` if you disagree

ma2bd: None of that matters because we want to change things in depth

ndr-ds: Ok, just left it as `faucet-net` for now then!

ma2bd:
I still prefer the old name (until we refactor tests entirely, notably removing local-net) so I suggest we revert #2998 instead with PR #3073

PR #3064, opened 2024-12-19
Move `Hashed` to linera-base.

afck:
## Motivation

We are going to use `Hashed` for blobs, too, not only for certificate values.

## Proposal

Move it to `linera-base`.

Remove the `lite` method and create a `LiteValue::new` instead.

Also, rename `CertificateValueT` to `CertificateValue`.

## Test Plan

CI should catch regressions.

## Release Plan

- Nothing to do / These changes follow the usual release cycle.

## Links

- [reviewer checklist](https://github.com/linera-io/linera-protocol/blob/main/CONTRIBUTING.md#reviewer-checklist)

PR #3068, opened 2024-12-20
`impl<T: CertificateValue> From<&Hashed<T>> for LiteValue`

afck:
## Motivation

This addresses https://github.com/linera-io/linera-protocol/pull/3064#discussion_r1893807182.

## Proposal

`impl<T: CertificateValue> From<&Hashed<T>> for LiteValue`

Since we use the reference and don't consume the `Hashed<T>`, we still need to call `LiteValue::from(&value)` explicitly, or `(&value).into()`, but `value.into()` wouldn't work.

## Test Plan

Minor refactoring. CI will catch regressions.

## Release Plan

## Links

- Original comment: https://github.com/linera-io/linera-protocol/pull/3064#discussion_r1893807182

### DIFF
--- a/linera-chain/src/certificate/generic.rs
+++ b/linera-chain/src/certificate/generic.rs
@@ -115,7 +115,7 @@ impl<T> GenericCertificate<T> {
         T: CertificateValue,
     {
         crate::certificate::LiteCertificate {
-            value: LiteValue::new(&self.value),
+            value: LiteValue::from(&self.value),
             round: self.round,
             signatures: std::borrow::Cow::Borrowed(&self.signatures),
         }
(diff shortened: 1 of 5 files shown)


deuszx:
I'm just gonna paste here what I wrote in the other thread: personally I don't like From conversions which are not obvious. It makes it difficult to track the flow - it's impossible to look for w/ IDE's "find usage/references" feature.

bart-linera (approved):
This makes sense to me, but I would be fine with leaving the conversion as a constructor, too.

ma2bd:
Should be use a small extension trait then, so that a method ´fn lite()' can be added in a later module?

afck:
An extension trait is also an option. To be honest, I don't have a strong opinion here.

ma2bd:
Can we go for the trait? @deusz has a point here about From conversions

ma2bd:
Or maybe we just keep the current code and close this PR...

PR #3075, opened 2024-12-29
feat: Make the communicate_with_quorum grace period configurable

### LINKED ISSUE #2836: Make the `communicate_with_quorum` grace period configurable.
Also, make the default `0.1` rather than `0.2`, to decrease latency.

And don't add the grace period at all in the failure cases (in particular `highest_key_score + remaining_votes < committee.quorum_threshold()`).

huitseeker:
- Renamed `GRACE_PERIOD` to `DEFAULT_GRACE_PERIOD` ~~and altered its value~~,
- added grace period argument in communicate with quorum,
- added that value to `ChainClientOptions` and threaded through the `Client` builder,
- Refactored the `end_time` setting logic in the `communicate_with_quorum` function to improve the handling of unreachable quorum on any key,
- ~~also corrects a handful of typos (in comments and error messages) as I was reading in the area~~

Fixes #2836. /cc @afck for review @Twey for context

### DIFF
--- a/linera-client/src/client_options.rs
+++ b/linera-client/src/client_options.rs
@@ -158,7 +158,7 @@ pub struct ClientOptions {
     pub restrict_chain_ids_to: Option<HashSet<ChainId>>,
 
     /// An additional delay, after reaching a quorum, to wait for additional validator signatures,
-    /// as a fraction of time taken to reach quorum. If `None`, uses the default value.
+    /// as a fraction of time taken to reach quorum. If omitted, uses the default value.
     #[arg(long)]
     pub grace_period: Option<f64>,
 }
--- a/linera-client/src/client_options.rs
+++ b/linera-client/src/client_options.rs
@@ -18,7 +18,7 @@ use linera_base::{
     },
     ownership::{ChainOwnership, TimeoutConfig},
 };
-use linera_core::client::BlanketMessagePolicy;
+use linera_core::{client::BlanketMessagePolicy, DEFAULT_GRACE_PERIOD};
 use linera_execution::{
     committee::ValidatorName, ResourceControlPolicy, WasmRuntime, WithWasmDefault as _,
 };
@@ -158,9 +158,9 @@ pub struct ClientOptions {
     pub restrict_chain_ids_to: Option<HashSet<ChainId>>,
 
     /// An additional delay, after reaching a quorum, to wait for additional validator signatures,
-    /// as a fraction of time taken to reach quorum. If omitted, uses the default value.
-    #[arg(long)]
-    pub grace_period: Option<f64>,
+    /// as a fraction of time taken to reach quorum.
+    #[arg(long, default_value_t = DEFAULT_GRACE_PERIOD)]
+    pub grace_period: f64,
 }
 
 impl ClientOptions {
--- a/linera-client/src/client_options.rs
+++ b/linera-client/src/client_options.rs
@@ -156,6 +156,11 @@ pub struct ClientOptions {
     /// an empty string.
     #[arg(long, value_parser = util::parse_chain_set)]
     pub restrict_chain_ids_to: Option<HashSet<ChainId>>,
+
+    /// An additional delay, after reaching a quorum, to wait for additional validator signatures,
+    /// as a fraction of time taken to reach quorum. If `None`, uses the default value.
+    #[arg(long)]
+    pub grace_period: Option<f64>,
 }
 
 impl ClientOptions {
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -1111,6 +1111,7 @@ where
                         .await
                 })
             },
+            None,
         )
         .await?;
         Ok(())
@@ -1146,6 +1147,7 @@ where
                 let action = action.clone();
                 Box::pin(async move { updater.send_chain_update(action).await })
             },
+            None,
         )
         .await?;
         ensure!(
@@ -1516,6 +1518,7 @@ where
                         .await
                 })
             },
+            None,
         )
         .await;
         let received_certificate_batches = match result {
@@ -1673,6 +1676,7 @@ where
                         .await
                 }
             },
+            None,
         )
         .await?;
 
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -158,8 +158,8 @@ where
     /// Whether to block on cross-chain message delivery.
     cross_chain_message_delivery: CrossChainMessageDelivery,
     /// An additional delay, after reaching a quorum, to wait for additional validator signatures,
-    /// as a fraction of time taken to reach quorum. If `None`, uses the default value.
-    grace_period: Option<f64>,
+    /// as a fraction of time taken to reach quorum.
+    grace_period: f64,
     /// Chains that should be tracked by the client.
     // TODO(#2412): Merge with set of chains the client is receiving notifications from validators
     tracked_chains: Arc<RwLock<HashSet<ChainId>>>,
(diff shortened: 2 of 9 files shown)


linera-core/src/updater.rs
```diff
@@ -31,9 +31,9 @@ use crate::{
     remote_node::RemoteNode,
 };
 
-/// The amount of time we wait for additional validators to contribute to the result, as a fraction
-/// of how long it took to reach a quorum.
-const GRACE_PERIOD: f64 = 0.2;
+/// The default amount of time we wait for additional validators to contribute 
+/// to the result, as a fraction of how long it took to reach a quorum.
+const DEFAULT_GRACE_PERIOD: f64 = 0.1;
```
ma2bd, line 36: Please keep the current value

linera-core/src/updater.rs
```diff
@@ -97,14 +97,16 @@ pub enum CommunicationError<E: fmt::Debug> {
...
+/// Tries to stop early when a quorum is reached. If `grace_period` is specified, other validators
+/// are given additional time to contribute to the result. The grace period is calculated as a fraction
+/// (defaulting to `DEFAULT_GRACE_PERIOD`) of the time taken to reach quorum.
 pub async fn communicate_with_quorum<'a, A, V, K, F, R, G>(
     validator_clients: &'a [RemoteNode<A>],
     committee: &Committee,
     group_by: G,
     execute: F,
+    // Grace period as a fraction of time taken to reach quorum
+    grace_period: impl Into<Option<f64>>,
```
ma2bd, line 109: We generally avoid `impl Into` (to keep type-checking strong and the code normalized).

linera-core/src/updater.rs
```diff
@@ -167,13 +170,15 @@ where
...
-        // If a key reaches a quorum or it becomes clear that no key can, wait for the grace
-        // period to collect more values or error information and then stop.
-        if end_time.is_none()
-            && (highest_key_score >= committee.quorum_threshold()
-                || highest_key_score + remaining_votes < committee.quorum_threshold())
-        {
-            end_time = Some(Instant::now() + start_time.elapsed().mul_f64(GRACE_PERIOD));
+        // If it becomes clear that no key can reach a quorum, break early.
+        if highest_key_score + remaining_votes < committee.quorum_threshold() {
+            break 'vote_wait;
```
ma2bd, line 175: cc @afck for reviewing this part

ma2bd:
Thanks @huitseeker for the PR!
* One cannot change the protocol behavior (here default value) in a PR called "refactor"
* Consider using several distinct commits in the PR next time :) to separate the easy stuff from the more serious refactoring
* Seems like you need to call `cargo +nightly fmt`

linera-core/src/updater.rs
huitseeker: The value was changed following the advice in #2836.

huitseeker:
- removed typo fixes,
- split change of DEFAULT_GRACE_PERIOD to another commit.

linera-core/src/updater.rs
huitseeker: reverted to `Option<f64>`

huitseeker:
> One cannot change the protocol behavior (here default value) in a PR called "refactor" (also I think 0.2 is fine for now)

Ah, yeah, i had not noticed this was public. I would actually advise switching this low level function to `pub(crate)`

linera-core/src/updater.rs
afck: Makes sense to me. :+1: 
(An alternative would be to set `end_time` to `now()`, but maybe the `break` is cleaner.)

linera-core/src/updater.rs
afck: I don't have a strong opinion; sorry for the confusion!
(The idea was that since `0.2` means we make the _happy path_ 20% slower it improves user experience to reduce this. And we will need additional tools to help slow validators catch up anyway. But we can always change the default later.)

afck (approved):
Looks good to me, but it doesn't actually make it configurable yet, i.e. add a config option.

huitseeker:
Thanks for the feedback @ma2bd and @afck. I have:
- removed the commit that changed the default value of the grace period (now at the original value of 0.2)
- added a commit that includes the `grace_period` parameter as part of the `ChainClientOptions` and threads that value through the `Client` builder.

linera-core/src/client/mod.rs
```diff
@@ -157,6 +157,9 @@ where
     message_policy: MessagePolicy,
     /// Whether to block on cross-chain message delivery.
     cross_chain_message_delivery: CrossChainMessageDelivery,
+    /// Grace period for communicate_with_quorum as a fraction of time taken to reach quorum.
+    /// If None, uses the default value.
```
afck, line 161:
```suggestion
    /// Grace period for `communicate_with_quorum` as a fraction of time taken to reach quorum.
    /// If `None`, uses the default value.
```

Or maybe we should not mention the name of the method. "An additional delay, after reaching a quorum, to wait for additional validator signatures."?

afck (commented):
Thanks; but that still doesn't make it user-configurable: It doesn't add a CLI option yet. (I think that needs to be added to `client_options.rs`.)

linera-core/src/client/mod.rs
```diff
@@ -1516,6 +1527,7 @@ where
                         .await
                 })
             },
+            None,
```
afck, line 1530: Did you mean to use `self.options.grace_period`, or is this deliberately `None`?

linera-client/src/client_options.rs
```diff
@@ -156,6 +156,11 @@ pub struct ClientOptions {
     /// an empty string.
     #[arg(long, value_parser = util::parse_chain_set)]
     pub restrict_chain_ids_to: Option<HashSet<ChainId>>,
+
+    /// An additional delay, after reaching a quorum, to wait for additional validator signatures,
+    /// as a fraction of time taken to reach quorum. If `None`, uses the default value.
+    #[arg(long)]
+    pub grace_period: Option<f64>,
```
afck, line 163: (We could also make this a percentage. Not sure if users will find it more intuitive to specify `--grace-period 20` or `--grace-period 0.2`… or `--grace-period 20%`? Anyway, we can always change this later.)

linera-client/src/client_options.rs
```diff
@@ -156,6 +156,11 @@ pub struct ClientOptions {
     /// an empty string.
     #[arg(long, value_parser = util::parse_chain_set)]
     pub restrict_chain_ids_to: Option<HashSet<ChainId>>,
+
+    /// An additional delay, after reaching a quorum, to wait for additional validator signatures,
+    /// as a fraction of time taken to reach quorum. If `None`, uses the default value.
```
afck, line 161: Since this is user-visible maybe we shouldn't mention Rust values like `None`. ("If omitted"?)

afck (approved):
No blockers from my side!

linera-client/src/client_options.rs
ma2bd: please no percentage. We don't need to create footguns

linera-core/src/client/mod.rs
```diff
@@ -157,6 +157,9 @@ where
     message_policy: MessagePolicy,
     /// Whether to block on cross-chain message delivery.
     cross_chain_message_delivery: CrossChainMessageDelivery,
+    /// An additional delay, after reaching a quorum, to wait for additional validator signatures,
+    /// as a fraction of time taken to reach quorum. If `None`, uses the default value.
+    grace_period: Option<f64>,
```
ma2bd, line 162: Note that we could remove the `Option` entirely by injecting the default value during CLI parsing (with or without moving the constant).

huitseeker: That would work for the CLI.  Clients created in `linera-service` would also need to inject the default.
 
I've added a commit that shows what that looks like. This `linera-service` concern also shows why it may make sense to keep `DEFAULT_GRACE_PERIOD` in the core crate.

ma2bd (approved):
Thanks!

Issue #2980, opened 2024-11-28
Investigate "validator forgot a certificate they signed before" issue.


deuszx:
When running tests [1], we get 
```
2024-11-28T12:04:40.529396Z  WARN linera::main: linera_core::remote_node: Validator 95fe9f76cdcd8e364033eb330babad3d0b9421b50bd8ad389a3f29d7b50ebe50 forgot a certificate value that they signed before
2024-11-28T12:04:40.529483Z  WARN linera::main: linera_core::remote_node: Validator 02a9b7afb957f4d54490cd808d6f41880b763cc853c588b7f9e2788c80de2d5c forgot a certificate value that they signed before
2024-11-28T12:04:40.529595Z  WARN linera::main: linera_core::remote_node: Validator 8f00a9a78f77041752d58605400d50aea4167c1ae6f02de4e1e5e32e7f9e2659 forgot a certificate value that they signed before
2024-11-28T12:04:40.529667Z  WARN linera::main: linera_core::remote_node: Validator f83b11500e9874d20d6cfe020def3f004949cb1653ae045755b7cc59f7785200 forgot a certificate value that they signed before
```

We need to investigate why and fix.
------------

[1] In my case it was
```
cargo run --release -p linera-storage-service -- memory --endpoint $LINERA_STORAGE_SERVICE &\
  cargo test --features storage-service,unstable-oracles -- storage_service_tcp --nocapture
```

afck:
It looks like after `create_final_vote` we're not caching the signed `ConfirmedBlock` value.

(In `vote_for_block_proposal` we cache everything as expected.)

ndr-ds:
Running `linera benchmark` also spews out a lot of these pretty consistently

afck:
I think the executed block cache isn't working at all:
We _insert_ executed block based on the hash from `Hashed<ExecutedBlock>`, but then we look them up based on `Hashed<ConfirmedBlock>` or `Hashed<ValidatedBlock>`, which are different?

ndr-ds:
I actually don't see us doing it for `create_final_vote` 🤔 and I have code locally adding it, and the warnings go away for `linera benchmark`

afck:
Really??
I think it's being inserted here: https://github.com/linera-io/linera-protocol/blob/69268f8b1eb8791f05335a4f5b2bd7b1c9fae90a/linera-core/src/chain_worker/state/attempted_changes.rs#L212

[...]

PR #3076, opened 2024-12-30
Serialize Confirmed and Validated blocks as their inner blocks.

deuszx:
## Motivation

As described in https://github.com/linera-io/linera-protocol/issues/2980 we've started to see a lot of logs about validator forgetting a block it had already signed. As @afck correctly observed - the problem was that when inserting a `ConfirmedBlock` (and `ValidatedBlock`) into the `ExecutedBlock` cache, they were inserted under a hash of the wrapper type (`ConfirmedBlock`/`ValidatedBlock`) rather than the hash of `ExecutedBlock` they contained. This meant that when looking up the cached value we always failed to find it (since the keys were different) – hence the error log.

## Proposal

Here we serialize `ConfirmedBlock` and `ValidatedBlock` as their inner `ExecutedBlock`s they contain. This means that their hashes are now equivalent (which fixes the original problem with the cache) but could also mean their serialized forms are now interchangeable. This is potentially problematic but not critical (i.e. for the protocol). Note that the _votes_ are still different - i.e. vote for a `ConfirmedBlock` will be different from a vote for `ValidatedBlock` containing the same `ExecutedBlock` (TODO: add link to the `test_signed_values` unit test).

Interestingly, we now drop the `ConfirmedBlock` and `ValidatedBlock` types from the RPC (see [Update RPC format](https://github.com/linera-io/linera-protocol/pull/3076/commits/1a8fb06be48aebae7500b8647a8cf94d18154b13)) which actually might be considered a change for the better since they were a "intermediate", wrapper types and the _kind_ of the certificate is encoded in the certificates' types themselves - `ValidatedBlockCertificate` and `ConfirmedBlockCertificate`.

## Test Plan

I've updated the `test_hashes` unit test and verified that we no longer get `"validator forgot a certificate they signed before"` in the logs by running
```
cargo test test_end_to_end_publish_data_blob_in_cli::storage_test_service_grpc -- --nocapture 2>&1 | grep forgot
```

## Release Plan

Since this doesn't affect the correctness (or safety) of the protocol, I propose:

## Links

Fixes https://github.com/linera-io/linera-protocol/issues/2980

### DIFF
--- a/linera-chain/src/block.rs
+++ b/linera-chain/src/block.rs
@@ -17,29 +17,10 @@ use thiserror::Error;
 use crate::{data_types::ExecutedBlock, ChainError};
 
 /// Wrapper around an `ExecutedBlock` that has been validated.
-#[derive(Debug, PartialEq, Eq, Clone)]
+#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
+#[serde(transparent)]
 pub struct ValidatedBlock {
     executed_block: Hashed<ExecutedBlock>,
-}
-
-impl Serialize for ValidatedBlock {
-    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
-    where
-        S: serde::Serializer,
-    {
-        self.executed_block.serialize(serializer)
-    }
-}
-
-impl<'de> Deserialize<'de> for ValidatedBlock {
-    fn deserialize<D>(deserializer: D) -> Result<ValidatedBlock, D::Error>
-    where
-        D: serde::Deserializer<'de>,
-    {
-        Ok(ValidatedBlock {
-            executed_block: Hashed::deserialize(deserializer)?,
-        })
-    }
 }
 
 impl ValidatedBlock {
@@ -88,30 +69,11 @@ impl ValidatedBlock {
 impl<'de> BcsHashable<'de> for ValidatedBlock {}
 
 /// Wrapper around an `ExecutedBlock` that has been confirmed.
-#[derive(Debug, PartialEq, Eq, Clone)]
+#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
+#[serde(transparent)]
 pub struct ConfirmedBlock {
     // The executed block contained in this `ConfirmedBlock`.
     executed_block: Hashed<ExecutedBlock>,
-}
-
-impl Serialize for ConfirmedBlock {
-    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
-    where
-        S: serde::Serializer,
-    {
-        self.executed_block.serialize(serializer)
-    }
-}
-
-impl<'de> Deserialize<'de> for ConfirmedBlock {
-    fn deserialize<D>(deserializer: D) -> Result<ConfirmedBlock, D::Error>
-    where
-        D: serde::Deserializer<'de>,
-    {
-        Ok(ConfirmedBlock {
-            executed_block: Hashed::deserialize(deserializer)?,
-        })
-    }
 }
 
 #[async_graphql::Object(cache_control(no_cache))]
--- a/linera-chain/src/block.rs
+++ b/linera-chain/src/block.rs
@@ -19,34 +19,30 @@ use crate::{data_types::ExecutedBlock, ChainError};
 /// Wrapper around an `ExecutedBlock` that has been validated.
 #[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
 #[serde(transparent)]
-pub struct ValidatedBlock {
-    executed_block: Hashed<ExecutedBlock>,
-}
+pub struct ValidatedBlock(Hashed<ExecutedBlock>);
 
 impl ValidatedBlock {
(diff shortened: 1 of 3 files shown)


linera-chain/src/block.rs
```diff
@@ -17,11 +17,31 @@ use thiserror::Error;
...
+impl<'de> Deserialize<'de> for ValidatedBlock {
+    fn deserialize<D>(deserializer: D) -> Result<ValidatedBlock, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        Ok(ValidatedBlock {
+            executed_block: Hashed::deserialize(deserializer)?,
+        })
+    }
+}
```
afck, line 43: Is this equivalent to `#[serde(transparent)]`?

linera-chain/src/unit_tests/data_types_tests.rs
```diff
@@ -59,7 +59,7 @@ fn test_hashes() {
     let confirmed_hashed = Hashed::new(ConfirmedBlock::new(executed_block.clone()));
     let validated_hashed = Hashed::new(ValidatedBlock::new(executed_block));
 
-    assert_ne!(confirmed_hashed.hash(), validated_hashed.hash());
+    assert_eq!(confirmed_hashed.hash(), validated_hashed.hash());
```
afck, line 62: Maybe we should have a test for the concern that this wan was originally meant to address: Check that you can't use a signature from a vote/certificate for a validated block inside a vote/certificate for the same _confirmed_ block; signature verification should fail.

deuszx: In the test above we check that votes for Confirmed and Validated blocks (containing the same executed block inside) are not equal. Do you want me to extend that to also check that signatures for those votes aren't interchangeable?

linera-chain/src/block.rs
deuszx: Ah, it is. Thanks for pointing it out.

linera-chain/src/unit_tests/data_types_tests.rs
afck: Ah, right, I forgot about the other test.
But yes, I think that would be best: Check that the signature for the validated block doesn't work as a signature for the confirmed block.

linera-chain/src/block.rs
```diff
@@ -17,11 +17,31 @@ use thiserror::Error;
...
+
 impl ValidatedBlock {
```
ma2bd, line 45: Do we need the type `ValidatedBlock` and `ConfirmedBlock` and all this boilerplate in the first place?

deuszx: I've thought about that (i.e. removal of `ValidatedBlock` and `ConfirmedBlock`) but decided not to do that (yet) b/c:
- they still seem useful when distinguishing between different type of blocks we're dealing with in various steps of the protocol flow
- we'd have to decide what to do with the `Timeout` type which is a corresponding type for the `TimeoutCertificate`.
- didn't want to do that now as it seemed not a priority.

linera-chain/src/block.rs
deuszx: Done in [use serde(transparent)](https://github.com/linera-io/linera-protocol/pull/3076/commits/9387762edffef6c883f7dbaeaae827c39c566c17). 

I also refactored these types to newtype structs.

linera-chain/src/unit_tests/data_types_tests.rs
deuszx: That will pass though b/c when signing we sign `LiteValue` type which contains only the _hash_ of the block (so the `ExecutedBlock` type) and we add the _kind_ of the block/certificate separately in `LiteVote`.

Separately from that - I just noticed that [we don't sign `ChainId` ](https://github.com/linera-io/linera-protocol/blob/7799087241b2bc667244c60782c73cf82fa7bc60/linera-chain/src/data_types.rs#L822-L823) and I think we should.

afck: That's what I mean: I'd add that test just to make sure we never accidentally remove the kind from the sign data.

The block already contains chain ID, so that should be fine.

deuszx: Got it now, I thought you meant to change how the types look like.

Right, hash already "contains" the chain_id b/c it was part of the struct. Maybe we can "lift" it to `LiteVote` type so that the "metadata" of where the vote comes from is easily available. RN it's impossible to know what chain that vote is for/from.

afck: More generally, you're saying: Maybe it would be more useful if the lite vote contained the (TBD) _header_ of the block rather than just the _hash_?

Definitely worth discussing, but probably something for a separate PR?

deuszx: I meant that we keep the hash (of the block) but add the chainId explicitly, alongside, rest of the data. But maybe you're right and it'll be more useful to include the whole block header there. That's how it'll work when bridging (and validating w/ light clients) so it makes sense to have the same approach everywhere.

Definitely not in this PR.

afck (approved):

PR #3081, opened 2025-01-03
Reinstate kind workflow

christos-h:
## Motivation

Better testing for ScyllaDB validators.

## Proposal

Add `kind` workflow back in.

### DIFF
--- a/.github/workflows/kubernetes.yml
+++ b/.github/workflows/kubernetes.yml
@@ -64,7 +64,7 @@ jobs:
           timeout_minutes: 30
           command: |
             kind get clusters | xargs -I {} kind delete cluster --name {}
-            RUST_LOG=linera=info cargo test --locked -p linera-service --features scylladb,kubernetes --test end_to_end_tests -- kubernetes --nocapture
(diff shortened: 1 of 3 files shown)


christos-h:
> Nice! what was the fix?

`seastar` does arithmetic based on available resources on whether to initialise or not (even in developer mode). There were some assertions failing when it was initialising (on my machine as well).

It looks like increasing the default resouirces allocated to the `scylla-manager` pods in `scylla-manager.value.yaml` fixed this.

.github/workflows/kubernetes.yml
```diff
@@ -0,0 +1,91 @@
...
+      - name: Run e2e tests
+        uses: nick-fields/retry@v2
+        with:
+          # Port forwarding sometimes dies, which makes all requests timeout
+          # Which is why we need retries
+          max_attempts: 3
+          timeout_minutes: 30
+          command: |
+            kind get clusters | xargs -I {} kind delete cluster --name {}
+            RUST_LOG=linera=info cargo test --locked -p linera-service --features scylladb,kubernetes --test linera_net_tests -- kubernetes --nocapture
```
ndr-ds, line 67: A good follow up here would be to see if we can fix the port forwarding issue, so that we don't have to do these retries and save some github money

PR #3084, opened 2025-01-06
Introduce the functionality of getting the list of blobs.

MathieuDutSik:
## Motivation

It is a useful functionality for obtaining the list of storage blobs. We introduce it here.

## Proposal

It is important to access the list of ChainIDs and BlobIDs from the command line. Accessing
the ChainIDs is more complicated than expected since we need to access the list of root keys
and the functionality is missing right now. So we limit ourselves to BlobIDs.

The BlobIDs are accessed through the `find_keys_by_prefix`. The problem is that the keys are
obtained from the serialization of the `BaseKey`. Therefore, in order to avoid leaking the inner
logic of the serialization into the code, we introduce a `to_bytes` function.

The BlobIDs access is added to the storage and also to the validator functionality in a
straightforward way.

## Test Plan

A test has been added for the functionality. We only check for the BlobsIDs to be contained. We cannot
check for equality before 

## Release Plan

This functionality could be useful for operators on DevNet / TestNet.

## Links

None.

### DIFF
--- a/linera-client/src/client_options.rs
+++ b/linera-client/src/client_options.rs
@@ -434,6 +434,9 @@ pub enum ClientCommand {
         chain_id: Option<ChainId>,
     },
 
+    /// Show the list of BlobIds of the validator
+    ListAllBlobIds,
+
     /// Show the version and genesis config hash of a new validator, and print a warning if it is
     /// incompatible. Also print some information about the given chain while we are at it.
     QueryValidator {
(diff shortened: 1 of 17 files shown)


linera-client/src/client_options.rs
```diff
@@ -434,6 +434,9 @@ pub enum ClientCommand {
         chain_id: Option<ChainId>,
     },
 
+    /// Show the list of BlobIds of the validator
```
ma2bd, line 437: The local node (not necessarily a validator)

MathieuDutSik: Ok.

MathieuDutSik:
This PR is killed because the functionality is essentially used for debugging and putting it in the protocol is not great.
The replacement PR is https://github.com/linera-io/linera-protocol/pull/3161

PR #3087, opened 2025-01-07
Switch alloy support from `linera_alloy` to `alloy`.

MathieuDutSik:
## Motivation

We introduced the `linera_alloy` when we needed a version of `alloy` from crates.io
In the meantime, `alloy` has been fully published in crates.io
Therefore, we switch to this version.

## Proposal

We switch from `linera_alloy` to `alloy` version 0.9.2 (the latest one) in the crates.io
We also remove the `get_wallet` which is not used.

## Test Plan

The CI does the testing.

## Release Plan

This is currently not being used in the TestNet / DevNet, so it does not really matter.

## Links

None


ma2bd (approved):
Nice!

PR #3101, opened 2025-01-08
New Block structure

deuszx:
## Motivation

Verifying Linera (micro)chains with a light client requires following the chain's blocks and subsequently confirming whether an event, a message, state change, were part of that block. Our current block definition - `ExecutedBlock` type - was not made for this purpose:
- there is no distinction between block's "metadata" and the "body" - which would allow for _light_ verification of the blocks.
- the events/messages/operations are all layed out, flat, in the `BlockExecutionOutcome` w/o a way to easily verify whether a particular event is part of that set other than sending the whole vector.

## Proposal

In this PR we introduce the new block structures. The new `Block` type consists of two elements: 
- (block) header
- (block) body.

`BlockHeader` is a succint representation of the chain's new block. It contains metadata that suffices to track the chain's progress:
```rust
pub struct BlockHeader {
    /// The block version.
    pub version: u8, // TODO: More granular versioning. #3078
    /// The chain to which this block belongs.
    pub chain_id: ChainId,
    /// The number identifying the current configuration.
    pub epoch: Epoch,
    /// The block height.
    pub height: BlockHeight,
    /// The timestamp when this block was created.
    pub timestamp: Timestamp,
    /// The hash of the chain's execution state after this block.
    pub state_hash: CryptoHash,
    /// Certified hash of the previous block in the
    /// chain, if any.
    pub previous_block_hash: Option<CryptoHash>,
    /// The user signing for the operations in the block and paying for their execution
    /// fees. If set, this must be the `owner` in the block proposal. `None` means that
    /// the default account of the chain is used. This value is also used as recipient of
    /// potential refunds for the message grants created by the operations.
    pub authenticated_signer: Option<Owner>,

    // Inputs to the block, chosen by the block proposer.
    /// Cryptographic hash of all the incoming bundles in the block.
    pub bundles_hash: CryptoHash,
    /// Cryptographic hash of all the operations in the block.
    pub operations_hash: CryptoHash,

    // Outcome of the block execution.
    /// Cryptographic hash of all the messages in the block.
    pub messages_hash: CryptoHash,
    /// Cryptographic hash of all the oracle responses in the block.
    pub oracle_responses_hash: CryptoHash,
    /// Cryptographic hash of all the events in the block.
    pub events_hash: CryptoHash,
}
```

`BlockBody` holds the data that is not necessary for (light) verification but was either an input to the block's execution (incoming bundles, operations) or its output (oracle responses, messages sent, events emitted). Each vector is deterministically hashed (into `CryptoHash`) and the result of that hashing is put into the `BlockHeader`. This way it's possible to verify inclusion of any specific element w/o having to sent all of the body.

```rust
/// The body of a block containing all the data included in the block.
#[derive(Debug, PartialEq, Eq, Hash, Clone, Serialize, Deserialize, SimpleObject)]
pub struct BlockBody {
    /// A selection of incoming messages to be executed first. Successive messages of same
    /// sender and height are grouped together for conciseness.
    pub incoming_bundles: Vec<IncomingBundle>,
    /// The operations to execute.
    pub operations: Vec<Operation>,
    /// The list of outgoing messages for each transaction.
    pub messages: Vec<Vec<OutgoingMessage>>,
    /// The record of oracle responses for each transaction.
    pub oracle_responses: Vec<Vec<OracleResponse>>,
    /// The list of events produced by each transaction.
    pub events: Vec<Vec<EventRecord>>,
}
```

## Test Plan

All of the tests have been updated to work with the new types.

## Release Plan

- These changes follow the usual release cycle. Although we may want to defer this update since this change is not backwards compatible.

## Links

- Closes https://github.com/linera-io/linera-protocol/issues/3077


linera-chain/src/manager.rs
```diff
@@ -369,10 +369,10 @@ impl ChainManager {
         &self,
         certificate: &ValidatedBlockCertificate,
     ) -> Result<Outcome, ChainError> {
-        let new_block = &certificate.executed_block().block;
+        let new_block = certificate.block();
         let new_round = certificate.round;
         if let Some(Vote { value, round, .. }) = &self.confirmed_vote {
-            if value.inner().executed_block().block == *new_block && *round == new_round {
+            if value.inner().block() == new_block && *round == new_round {
```
deuszx, line 375: TODO: Double-check. Previously we were comparing the `Proposal` and now we compare the new `Block` type. It means we also compare the execution results now.

linera-core/src/client/mod.rs
```diff
@@ -2012,8 +2012,8 @@ where
 
         match self.process_pending_block_without_prepare().await? {
             ClientOutcome::Committed(Some(certificate))
-                if certificate.executed_block().block
-                    == confirmed_value.inner().executed_block().block =>
+                if certificate.block() == confirmed_value.inner().block() =>
+            // TODO: double check this. Previously we were comparing the Proposal only.
```
deuszx, line 2016: TODO: Double-check. Same here as above.

linera-chain/src/chain.rs
```diff
@@ -147,8 +148,9 @@ static STATE_HASH_COMPUTATION_LATENCY: LazyLock<HistogramVec> = LazyLock::new(||
     )
 });
 
-/// The BCS-serialized size of an empty `ExecutedBlock`.
-const EMPTY_EXECUTED_BLOCK_SIZE: usize = 91;
+/// The BCS-serialized size of an empty [`Block`].
+// TODO: UPDATE
+const EMPTY_BLOCK_SIZE: usize = 91;
```
deuszx, line 153: TODO: Update new empty block size.

linera-chain/src/chain.rs
```diff
@@ -1235,5 +1241,5 @@ fn empty_executed_block_size() {
         outcome: crate::data_types::BlockExecutionOutcome::default(),
     };
     let size = bcs::serialized_size(&executed_block).unwrap();
-    assert_eq!(size, EMPTY_EXECUTED_BLOCK_SIZE);
+    assert_eq!(size, EMPTY_BLOCK_SIZE);
```
deuszx, line 1244: TODO: Test for empty block (new) size.

linera-base/src/crypto.rs
```diff
@@ -172,6 +180,38 @@ impl Serialize for CryptoHash {
...
+        // if serializer.is_human_readable() {
+        //     let s = self
+        //         .0
+        //         .iter()
+        //         .map(|hash| hash.to_string())
+        //         .collect::<Vec<_>>();
+        //     serializer.collect_seq(s)
+        // } else {
+        serializer.serialize_newtype_struct("CryptoHashVec", &self.0)
+        // }
```
deuszx, line 197: TODO: Should be safe to just have the non-human serialization. We don't expose the vector anywhere, it's a temporary value used internally.

linera-core/src/unit_tests/worker_tests.rs
```diff
@@ -712,18 +711,18 @@ where
...
-            .value()
-            .inner()
-            .executed_block()
-            .block,
-        &block_proposal0.content.block
-    ); // Should be confirmed after handling the certificate.
+    let executed_block: ExecutedBlock = chain
+        .manager
+        .get()
+        .validated_vote()
```
deuszx, line 717: This was the mistake 🙈 I changed _confirmed_ to _validated_.

linera-base/src/crypto.rs
```diff
@@ -172,6 +180,29 @@ impl Serialize for CryptoHash {
     }
 }
 
+impl Serialize for CryptoHashVec {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::ser::Serializer,
+    {
+        serializer.serialize_newtype_struct("CryptoHashVec", &self.0)
```
ma2bd, line 188: Another `#[transparent]`?

linera-chain/src/certificate/confirmed.rs
```diff
@@ -13,19 +13,19 @@ use serde::{ser::SerializeStruct, Deserialize, Deserializer, Serialize};
...
-    pub fn executed_block(&self) -> &ExecutedBlock {
-        self.inner().executed_block()
+    pub fn block(&self) -> &Block {
+        self.inner().block()
     }
 
     /// Returns whether this value contains the message with the specified ID.
     pub fn has_message(&self, message_id: &MessageId) -> bool {
-        self.executed_block().message_by_id(message_id).is_some()
+        self.block().message_by_id(message_id).is_some()
```
ma2bd, line 28: Would it make sense the handle the renaming `executed_block` -> `block` first in a separate PR?

linera-base/src/crypto.rs
deuszx: Ah, good point.

linera-chain/src/certificate/confirmed.rs
deuszx: Actually, I didn't get rid of `ExecutedBlock`. I tried but the `Proposal`/`BlockExecutionOutcome` is used in a couple of places (we execute before proposing, we execute when we receive a proposal, we re-execute when we receive a confirmed certificate, etc.) so I kept the `ExecutedBlock` as an internal data structure that caries the result of execution.

linera-base/src/crypto.rs
deuszx: Done in https://github.com/linera-io/linera-protocol/pull/3101/commits/dca5da358b8cda100075937b57c5c9ac61169a4e.

deuszx: Actually, this won't work, b/c we will start getting the errors that we can call `type_name()` only on `struct` or `newtype` (b/c we'll call `Deserialize` on the `Vec<_>` and that will fail).

ma2bd: Also why do we need this in the first place?

linera-chain/src/chain.rs
deuszx: Done in [Testy empty Block size.](https://github.com/linera-io/linera-protocol/pull/3101/commits/5fe042587db04596109fa57d5d66cb29303122bc)

linera-chain/src/chain.rs
deuszx: Done in [Testy empty Block size.](https://github.com/linera-io/linera-protocol/pull/3101/commits/5fe042587db04596109fa57d5d66cb29303122bc)

afck (changes requested):

linera-chain/src/block.rs
```diff
@@ -192,3 +206,286 @@ pub enum ConversionError {
     #[error("Expected a `TimeoutCertificate` value")]
     Timeout,
 }
+
+/// Block defines the atomic unit of growth of the Linera chain.
+#[derive(Debug, PartialEq, Eq, Hash, Clone, Serialize, Deserialize, SimpleObject)]
```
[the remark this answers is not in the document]
afck: Yes, that would unblock it from my side. :+1: 
(Not sure if others find hashing on deserialization controversial.)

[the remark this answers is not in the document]
deuszx: Done in [Check hash equality when deserializing Block](https://github.com/linera-io/linera-protocol/pull/3101/commits/f71b9446d765e62723b018793e222845ac78a403)

> (Not sure if others find hashing on deserialization controversial.)

that's unavoidable - either we never send `*_hash`-es and always hash ourselves or we send both and hash and check.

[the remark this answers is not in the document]
deuszx: Take a look at [Custom (de)serializer for BlockHeader.](https://github.com/linera-io/linera-protocol/pull/3101/commits/f3c8e49164563f79ba3da880cfd1847bb2a2b282) where I use a custom (de)serializer for the `BlockHeader` when that's (de)serialized within `Block` type.

linera-chain/src/block.rs
```diff
@@ -192,3 +206,286 @@ pub enum ConversionError {
...
+pub struct Block {
+    /// Header of the block containing metadata of the block.
+    pub header: BlockHeader,
+    /// Body of the block containing all of the data.
+    pub body: BlockBody,
+}
+
+/// Succint representation of a block.
```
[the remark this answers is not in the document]
deuszx: Done in [Address typos and renames](https://github.com/linera-io/linera-protocol/pull/3101/commits/65a80f6e96b88194d85b2949cadf17b79e68db13).

linera-core/src/unit_tests/worker_tests.rs
```diff
@@ -712,18 +711,18 @@ where
...
+        .get()
+        .confirmed_vote()
+        .unwrap()
+        .value()
+        .inner()
+        .block()
+        .clone()
+        .into();
+
+    assert_eq!(&executed_block.proposal, &block_proposal0.content.proposal); // Should be confirmed after handling the certificate.
```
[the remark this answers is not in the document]
deuszx: Done in [Address typos and renames](https://github.com/linera-io/linera-protocol/pull/3101/commits/65a80f6e96b88194d85b2949cadf17b79e68db13).

linera-indexer/example/tests/test.rs
```diff
@@ -149,8 +149,8 @@ async fn test_end_to_end_operations_indexer(config: impl LineraNetConfig) {
     );
 
     // checking indexer operation
-    let executed_block = last_block.value.executed_block;
-    let last_operation = executed_block.block.operations[0].clone();
+    let exeuted_block = last_block.value.block;
```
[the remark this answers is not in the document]
deuszx: Done in [Address typos and renames](https://github.com/linera-io/linera-protocol/pull/3101/commits/65a80f6e96b88194d85b2949cadf17b79e68db13).

linera-chain/src/block.rs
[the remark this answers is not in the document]
afck: Thanks! I like the custom deserializer. :+1: 

> that's unavoidable

I guess an alternative would be to not make `Block` serializable at all. Instead, you always send header and body separately. The fields of `Block` are then not public, and the only way to construct a block is via a constructor that checks that the hashes match.

But I think I prefer the deserialization code you just added.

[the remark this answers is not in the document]
deuszx: Can we resolve this comment now?

linera-chain/src/data_types.rs
```diff
@@ -395,16 +399,16 @@ impl OutgoingMessage {
     }
 }
 
-/// A [`Block`], together with the outcome from its execution.
+/// A [`Proposal`], together with the outcome from its execution.
 #[derive(Debug, PartialEq, Eq, Hash, Clone, Serialize, Deserialize, SimpleObject)]
 pub struct ExecutedBlock {
```
afck, line 404: Maybe `ProposalWithOutcome`. Or even inline it, if it's not used in many places anymore?

linera-chain/src/data_types.rs
```diff
@@ -76,7 +76,7 @@ pub struct Block {
     pub previous_block_hash: Option<CryptoHash>,
 }
 
-impl Block {
+impl Proposal {
```
afck, line 79: Now a `BlockProposal` contains a `ProposalContent` which contains a `Proposal`.

Maybe call this type `ProposedBlock`?

deuszx: I didn't call it `ProposedBlock` on purpose b/c we have the `BlockProposal` which sounds basically equivalent that is the encompassing type. It seemed weird that `BlockProposal` contains `ProposedBlock`.

linera-chain/src/block.rs
```diff
@@ -192,3 +206,286 @@ pub enum ConversionError {
     #[error("Expected a `TimeoutCertificate` value")]
     Timeout,
 }
+
+/// Block defines the atomic unit of growth of the Linera chain.
```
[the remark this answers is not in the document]
deuszx: I updated the comment a bit. I didn't go into too much details though as the `BlockBody` or `BlockHeader` have more of that already.

linera-chain/src/data_types.rs
afck: But isn't a `BlockProposal` containing a `Proposal` even worse?

Maybe `BlockInput`? `BlockTemplate`? `BlockPlan`? :confused:

linera-chain/src/block.rs
```diff
@@ -192,3 +206,362 @@ pub enum ConversionError {
...
+/// As part of the block body, contains all the incoming messages
+/// and operations to executed which define state transition of the chain.
```
afck, line 212:
```suggestion
/// and operations to execute which define a state transition of the chain.
```

linera-chain/src/block.rs
```diff
@@ -192,3 +206,362 @@ pub enum ConversionError {
     #[error("Expected a `TimeoutCertificate` value")]
     Timeout,
 }
+
+/// Block defines the atomic unit of growth of the Linera chain.
+/// As part of the block body, contains all the incoming messages
```
afck, line 211: I think it's common to add a blank line (i.e. new paragraph in markdown; also below).

```suggestion
///
/// As part of the block body, contains all the incoming messages
```

linera-chain/src/chain.rs
deuszx: You're right - it doesn't make sense to keep this test for the `ExecutedBlock` type.

linera-chain/src/block.rs
```diff
@@ -192,3 +206,363 @@ pub enum ConversionError {
...
+impl Block {
+    pub fn new(proposal: Proposal, outcome: BlockExecutionOutcome) -> Self {
+        let bundles_hash = hashing::hash_vec(&proposal.incoming_bundles);
+        let messages_hash = hashing::hash_vec_vec(&outcome.messages);
+        let operations_hash = hashing::hash_vec(&proposal.operations);
+        let oracle_responses_hash = hashing::hash_vec_vec(&outcome.oracle_responses);
+        let events_hash = hashing::hash_vec_vec(&outcome.events);
+
+        let header = BlockHeader {
+            version: 1,
```
afck, line 349: As discussed let's add versioning later, as a separate feature.

linera-chain/src/manager.rs
afck: Either way is okay; we could even just compare the round: There can't be more than one validated block certificate per round anyway, unless there is no quorum of honest validators.

linera-chain/src/test.rs
```diff
@@ -17,29 +17,31 @@ use linera_execution::{
...
+        epoch: parent_header.epoch,
+        chain_id: parent_header.chain_id,
         incoming_bundles: vec![],
         operations: vec![],
         previous_block_hash: Some(parent.hash()),
-        height: parent_block.height.try_add_one().unwrap(),
-        authenticated_signer: parent_block.authenticated_signer,
-        timestamp: parent_block.timestamp,
+        height: parent_header.height.try_add_one().unwrap(),
+        authenticated_signer: parent_header.authenticated_signer,
```
afck, line 37: (Weird! What if this is proposed by someone else? Not new in this PR, of course.)

linera-chain/src/unit_tests/chain_tests.rs
```diff
@@ -171,11 +171,11 @@ async fn test_block_size_limit() {
 
     // The valid block is accepted...
     let outcome = chain.execute_block(&valid_block, time, None).await.unwrap();
-    let executed_block = outcome.with(valid_block);
+    let block: Block = Block::new(valid_block, outcome);
```
afck, line 174:
```suggestion
    let block = Block::new(valid_block, outcome);
```

(Or `outocome.with(valid_block)`? You still use that below.)

deuszx: `outcome-.with(valid_block)` still returns `ExecutedBlock` while here we want `Block`.

linera-chain/src/block.rs
```diff
@@ -192,3 +206,363 @@ pub enum ConversionError {
...
+    /// The number identifying the current configuration.
+    pub epoch: Epoch,
+    /// The block height.
+    pub height: BlockHeight,
+    /// The timestamp when this block was created.
+    pub timestamp: Timestamp,
+    /// The hash of the chain's execution state after this block.
+    pub state_hash: CryptoHash,
+    /// Certified hash of the previous block in the
+    /// chain, if any.
```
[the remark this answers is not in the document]
deuszx: We should really make cargo fmt do that. I hate manually fixing comments' length.

linera-chain/src/unit_tests/chain_tests.rs
deuszx: Done in [Address review comments from Andreas](https://github.com/linera-io/linera-protocol/pull/3101/commits/89062f6fb4b18b5d5a3b037719617210e82123e4)

linera-chain/src/block.rs
deuszx: Done in [Address review comments from Andreas](https://github.com/linera-io/linera-protocol/pull/3101/commits/89062f6fb4b18b5d5a3b037719617210e82123e4)

linera-chain/src/block.rs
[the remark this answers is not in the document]
deuszx: Done in [Address review comments from Andreas](https://github.com/linera-io/linera-protocol/pull/3101/commits/89062f6fb4b18b5d5a3b037719617210e82123e4)

linera-chain/src/block.rs
```diff
@@ -192,3 +206,363 @@ pub enum ConversionError {
     #[error("Expected a `TimeoutCertificate` value")]
     Timeout,
 }
+
+/// Block defines the atomic unit of growth of the Linera chain.
+///
+/// As part of the block body, contains all the incoming messages
+/// and operations to execute which define state transition of the chain.
```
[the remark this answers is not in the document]
deuszx: Done in [Address review comments from Andreas](https://github.com/linera-io/linera-protocol/pull/3101/commits/89062f6fb4b18b5d5a3b037719617210e82123e4)

linera-chain/src/block.rs
```diff
@@ -84,34 +94,38 @@ impl ConfirmedBlock {
...
+    pub fn from_hashed(block: Hashed<Block>) -> Self {
+        Self(block)
     }
 
-    pub fn inner(&self) -> &Hashed<ExecutedBlock> {
+    pub fn inner(&self) -> &Hashed<Block> {
         &self.0
     }
 
+    pub fn take_inner(self) -> Hashed<Block> {
```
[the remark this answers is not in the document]
deuszx: Done in [Address review comments from Andreas](https://github.com/linera-io/linera-protocol/pull/3101/commits/89062f6fb4b18b5d5a3b037719617210e82123e4)

linera-chain/src/block.rs
```diff
@@ -84,34 +94,38 @@ impl ConfirmedBlock {
...
+
     /// Returns a reference to the `ExecutedBlock` contained in this `ConfirmedBlock`.
-    pub fn executed_block(&self) -> &ExecutedBlock {
+    pub fn block(&self) -> &Block {
         self.0.inner()
     }
 
     /// Consumes this `ConfirmedBlock`, returning the `ExecutedBlock` it contains.
-    pub fn into_inner(self) -> ExecutedBlock {
+    pub fn into_inner(self) -> Block {
```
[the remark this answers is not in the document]
deuszx: Done in [Address review comments from Andreas](https://github.com/linera-io/linera-protocol/pull/3101/commits/89062f6fb4b18b5d5a3b037719617210e82123e4)

afck (approved):
No blockers from my side, but:
* Can we remove `ExecutedBlock` now? (Maybe in another PR.)
* One test is still failing.
* Since this is a very central type, it would be great if we could get a second review.

deuszx:
> * Can we remove `ExecutedBlock` now? (Maybe in another PR.)

Discussed offline - and decided not to do that. `ExecutedBlock` is still being used as a return type from execution. it's also more useful as the internal, intermediate type b/c it retains the original `proposal <> outcome` distinction (`Block` is the _end_ type) and that is used in client (due to pre-execution before sending the proposal) and on the server to differentiate between _initial_ proposal and a _retry_.

> * One test is still failing.

I think I fixed that now.

> * Since this is a very central type, it would be great if we could get a second review.

I agree, let's wait.

ma2bd (approved):

PR #3102, opened 2025-01-08
Have a Cursor inside MessageBundle

ndr-ds:
## Motivation

`MessageBundle` already has the fields for a `Cursor`, and we even had a conversion function.

## Proposal

Just have a `Cursor` in there to make things cleaner

## Test Plan

CI

## Release Plan

### DIFF
--- a/linera-chain/src/lib.rs
+++ b/linera-chain/src/lib.rs
@@ -14,7 +14,7 @@ pub mod types {
 
 mod chain;
 pub mod data_types;
-mod inbox;
+pub mod inbox;
 pub mod manager;
 mod outbox;
 #[cfg(with_testing)]
--- a/linera-core/src/data_types.rs
+++ b/linera-core/src/data_types.rs
@@ -254,8 +254,10 @@ impl CrossChainRequest {
         match self {
             CrossChainRequest::UpdateRecipient { bundle_vecs, .. } => {
                 bundle_vecs.iter().any(|(_, bundles)| {
-                    debug_assert!(bundles.windows(2).all(|w| w[0].1.height <= w[1].1.height));
-                    matches!(bundles.first(), Some((_, h)) if h.height <= height)
+                    debug_assert!(bundles
+                        .windows(2)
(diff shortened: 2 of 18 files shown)


ndr-ds:
* **#3102** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/3102?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/3102?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

linera-core/src/data_types.rs
```diff
@@ -254,8 +254,10 @@ impl CrossChainRequest {
         match self {
             CrossChainRequest::UpdateRecipient { bundle_vecs, .. } => {
                 bundle_vecs.iter().any(|(_, bundles)| {
-                    debug_assert!(bundles.windows(2).all(|w| w[0].1.height <= w[1].1.height));
-                    matches!(bundles.first(), Some((_, h)) if h.height <= height)
+                    debug_assert!(bundles
+                        .windows(2)
+                        .all(|w| w[0].1.cursor.height <= w[1].1.cursor.height));
```
afck, line 259: Might as well assert that the cursors are strictly increasing.

```suggestion
                        .all(|w| w[0].1.cursor < w[1].1.cursor));
```

linera-chain/src/lib.rs
```diff
@@ -14,7 +14,7 @@ pub mod types {
 
 mod chain;
 pub mod data_types;
-mod inbox;
+pub mod inbox;
```
afck, line 17: So `Cursor` was _not_ public?
Then I'd actually leave it that way, and not do this PR, I think. (Happy to let others chime in.)

afck (commented):
Not sure about this anymore: It adds lots of lines of code and makes the `Cursor` type public.

PR #3104, opened 2025-01-08
Fix Witty memory allocation alignment

jvff:
## Motivation

When allocating heap memory in Witty, the alignment was incorrectly hard-coded to `1`. This causes issues when the allocated memory does not start at an aligned address. The code to store types will automatically add padding, which offsets the start of the write and can then overflow the allocated region.

## Proposal

Specify the allocation alignment whenever needed.

## Test Plan

Unit tests were added, including one that reproduces the issue `wit_store::test_list_fields` and now passes with the fix.

## Release Plan

      SDK soon.

## Links

### DIFF
--- a/linera-witty/tests/wit_load.rs
+++ b/linera-witty/tests/wit_load.rs
@@ -353,6 +353,16 @@ fn test_heap_allocated_fields() {
     );
 }
 
+/// Check that a [`Vec`] type is properly loaded from memory and lifted from its flat
+/// layout.
+#[test]
+fn test_vec() {
+    let expected = vec![SimpleWrapper(false), SimpleWrapper(true)];
+
+    test_load_from_memory(&[8, 0, 0, 0, 2, 0, 0, 0, 0, 1], expected.clone());
+    test_lift_from_flat_layout(hlist![0_i32, 2_i32], expected, &[0, 1]);
+}
+
 /// Tests that the type `T` and wrapped versions of it can be loaded from an `input` sequence of
 /// bytes in memory and that it matches the `expected` value.
 fn test_load_from_memory<T>(input: &[u8], expected: T)
--- a/linera-witty/tests/wit_load.rs
+++ b/linera-witty/tests/wit_load.rs
@@ -374,10 +374,30 @@ fn test_list_fields() {
             SimpleWrapper(true),
             SimpleWrapper(false),
         ],
+        second_vec: vec![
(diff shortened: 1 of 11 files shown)


linera-witty/tests/wit_load.rs
```diff
@@ -353,6 +354,52 @@ fn test_heap_allocated_fields() {
     );
 }
 
+/// Check that a [`Vec`] type is properly loaded from memory and lifted from its flat
+/// layout.
+#[test]
+fn test_vec() {
+    let expected = vec![SimpleWrapper(false), SimpleWrapper(true)];
+
+    test_load_from_memory(&[8, 0, 0, 0, 2, 0, 0, 0, 0, 1], expected.clone());
```
afck, line 363: Why 8, 2 and 1? Would this have been different without this change?
(Sorry, I'm still not familiar enough with the inner workings of Witty. :sweat_smile:)

linera-witty/tests/wit_load.rs
```diff
@@ -353,6 +354,52 @@ fn test_heap_allocated_fields() {
...
+    test_lift_from_flat_layout(hlist![0_i32, 2_i32], expected, &[0, 1]);
```
afck, line 364: (So the `input` here is: "`2` bytes at address `0`". And those two bytes are `0` for `false` and `1` for `true`?)

linera-witty/tests/wit_load.rs
```diff
@@ -353,6 +354,52 @@ fn test_heap_allocated_fields() {
     );
 }
 
+/// Check that a [`Vec`] type is properly loaded from memory and lifted from its flat
```
afck, line 357:
```suggestion
/// Checks that a [`Vec`] type is properly loaded from memory and lifted from its flat
```

(Also below.)

afck (approved):
The change looks good to me, but I don't understand all the values in the tests.

linera-witty/tests/wit_load.rs
jvff: The vec is stored in memory as a tuple of address and size (each four bytes), and the contents are stored right after the tuple. That means:

- 0..4: address (8, right after the tuple)
- 4..8: size (2)
- 8: `SimpleWrapper(false)` (0)
- 9: `SimpleWrapper(true)` (1)

linera-witty/tests/wit_load.rs
jvff: Exactly :raised_hands:

Issue #3048, opened 2024-12-17
Support uploading blobs separately


afck:

afck:
With https://github.com/linera-io/linera-protocol/pull/3108, this is solved for confirmed blocks.

For block proposals, we _do_ want to have the published blobs available at validation time, to check their size. And for validated blocks, we want to avoid storing the blobs in the chain manager, but also can't write them to storage yet.

So we add to the system execution state (as views; shown as simplified types here):
* `locked_blobs: Map<BlobId, Blob>` with the blobs that belong to the chain manager's `locked` block. (Remove from chain manager.)
* `proposal_blobs: Map<BlobId, Blob>` with the blobs that belong to the chain manager's `proposed` block. (Remove from chain manager.)
* `pending_validated_blobs: Option<(ValidatedBlockCertificate, Map<BlobId, Option<Blob>)>` with the validated block in the current round if it's still missing blobs.
* `pending_proposal_blobs: Map<Owner, (BlockProposal, Map<BlobId, Option<Blob>>)>` with any pending proposals and their sets of published blobs. (Each owner is only allowed to make one proposal at a time. If the proposal is a re-proposal, all _required_ blobs are included.)

There is a new endpoint `UploadPendingBlob(ChainId, BlobContent)` that adds the blob to `pending_proposal_blobs` and `pending_validated_blobs` wherever it matches, handles any that now have their complete set of blobs, and then returns the `ChainManagerInfo`. So clients can now:
* First upload their proposal/validated block.
* Then upload all the blobs that belong to it.

Conversely, if a client receives a chain manager with a proposal or a locked block that they are missing blobs for, they request it with a new endpoint `DownloadPendingBlob(ChainId, BlobId)`, which looks up the blob in `locked_blobs` and `proposal_blobs`. (_Not_ in the "pending" fields.)

These changes should be done in two separate PRs:
* [ ] Extract `locked_blobs` and `proposed_blobs` from the chain manager, add a `DownloadPendingBlob` endpoint; do staged downloads.
* [ ] Add `pending_validated_blobs`, `pending_proposal_blobs`, and an `UploadPendingBlob` endpoint; do staged uploads.

PR #3108, opened 2025-01-09
Upload blobs for confirmed certificates separately.

afck:
## Motivation

Including blobs with the gRPC message that contains a block proposal or certificate severely limits the total size of the blobs. (See #3048.)

## Proposal

As a first step, remove the blobs from the `handle_confirmed_certificate` functions and messages.

Instead, when a validator sees a fully signed confirmed block it creates the blob states in its local storage even if it doesn't have the blobs yet. The client can then upload the blobs one by one, and the validator will accept them. Finally, the client can retry sending the certificate.

We don't do this for block proposals or validated blocks yet: These will need to be handled differently, because in these cases the blob has not necessarily been successfully published yet, so we should _not_ create a blob state. Instead, we will put these blobs into a temporary cache.

## Test Plan

The existing tests are now using the new flow for confirmed block certificates.

## Release Plan

## Links

- Part of #3048

### DIFF
--- a/linera-core/src/chain_worker/state/attempted_changes.rs
+++ b/linera-core/src/chain_worker/state/attempted_changes.rs
@@ -239,7 +239,6 @@ where
     pub(super) async fn process_confirmed_block(
         &mut self,
         certificate: ConfirmedBlockCertificate,
-        blobs: &[Blob],
         notify_when_messages_are_delivered: Option<oneshot::Sender<()>>,
     ) -> Result<(ChainInfoResponse, NetworkActions), WorkerError> {
         let executed_block = certificate.executed_block();
@@ -298,33 +297,30 @@ where
         );
 
         let required_blob_ids = executed_block.required_blob_ids();
-        // Verify that no unrelated blobs were provided.
-        self.state
-            .check_for_unneeded_blobs(&required_blob_ids, blobs)?;
-        let blobs = self.state.get_required_blobs(executed_block, blobs).await?;
-        let blobs = blobs.into_values().collect::<Vec<_>>();
+        let blobs_result = self.state.get_required_blobs(executed_block, &[]).await;
+        let blobs_result = blobs_result.map(|blobs| blobs.into_values().collect::<Vec<_>>());
 
-        let certificate_hash = certificate.hash();
-
-        self.state
-            .storage
-            .write_blobs_and_certificate(&blobs, &certificate)
-            .await?;
+        if let Ok(blobs) = &blobs_result {
+            self.state
+                .storage
+                .write_blobs_and_certificate(blobs, &certificate)
+                .await?;
+        }
 
         // Update the blob state with last used certificate hash.
         let blob_state = BlobState {
-            last_used_by: certificate_hash,
+            last_used_by: certificate.hash(),
             chain_id: block.chain_id,
             block_height,
             epoch: block.epoch,
         };
+        let overwrite = blobs_result.is_ok(); // Overwrite only if we wrote the certificate.
+        let blob_ids = required_blob_ids.into_iter().collect::<Vec<_>>();
         self.state
             .storage
-            .maybe_write_blob_states(
-                &required_blob_ids.into_iter().collect::<Vec<_>>(),
-                blob_state,
-            )
+            .maybe_write_blob_states(&blob_ids, blob_state, overwrite)
             .await?;
+        blobs_result?;
 
         // Execute the block and update inboxes.
         self.state.chain.remove_bundles_from_inboxes(block).await?;
--- a/linera-core/src/chain_worker/state/attempted_changes.rs
+++ b/linera-core/src/chain_worker/state/attempted_changes.rs
@@ -297,8 +297,11 @@ where
         );
 
         let required_blob_ids = executed_block.required_blob_ids();
-        let blobs_result = self.state.get_required_blobs(executed_block, &[]).await;
-        let blobs_result = blobs_result.map(|blobs| blobs.into_values().collect::<Vec<_>>());
+        let blobs_result = self
+            .state
+            .get_required_blobs(executed_block, &[])
+            .await
+            .map(|blobs| blobs.into_values().collect::<Vec<_>>());
 
         if let Ok(blobs) = &blobs_result {
             self.state
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -45,8 +45,9 @@ use linera_chain::{
         MessageAction,
     },
     types::{
-        CertificateValue, ConfirmedBlock, ConfirmedBlockCertificate, GenericCertificate,
-        LiteCertificate, Timeout, TimeoutCertificate, ValidatedBlock, ValidatedBlockCertificate,
+        CertificateKind, CertificateValue, ConfirmedBlock, ConfirmedBlockCertificate,
+        GenericCertificate, LiteCertificate, Timeout, TimeoutCertificate, ValidatedBlock,
+        ValidatedBlockCertificate,
     },
     ChainError, ChainExecutionContext, ChainStateView,
 };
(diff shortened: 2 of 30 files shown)


linera-core/src/client/mod.rs
```diff
@@ -395,9 +396,24 @@ where
...
     ) -> Result<ChainInfoResponse, LocalNodeError> {
-        self.local_node
-            .handle_certificate(certificate, blobs, &self.notifier)
-            .await
+        if T::KIND == CertificateKind::Confirmed {
+            let result = self
+                .local_node
+                .handle_certificate(certificate.clone(), vec![], &self.notifier)
+                .await;
+            if let Err(LocalNodeError::BlobsNotFound(_)) = &result {
```
deuszx, line 404: Can you explain why we do that only for the `Confirmed` certificate kind? Is that b/c only for confirmed blocks we are sure that the blobs should be written to the storage?

linera-core/src/node.rs
```diff
@@ -103,6 +102,10 @@ pub trait ValidatorNode {
     /// Subscribes to receiving notifications for a collection of chains.
     async fn subscribe(&self, chains: Vec<ChainId>) -> Result<Self::NotificationStream, NodeError>;
 
+    // Uploads a blob content. Returns an error if the validator has not seen a
+    // certificate using this blob.
+    async fn upload_blob_content(&self, content: BlobContent) -> Result<BlobId, NodeError>;
```
deuszx, line 107: Personally, I'd just drop the "content" parts. This is "user-facing" feature and they are thinking about _blobs_ and not really bother w/ the distinction that "blob" has a cached hash etc.

linera-rpc/src/grpc/client.rs
```diff
@@ -349,6 +347,12 @@ impl ValidatorNode for GrpcClient {
         Ok(client_delegate!(self, get_genesis_config_hash, req)?.try_into()?)
     }
 
+    #[instrument(target = "grpc_client", skip(self), err, fields(address = self.address))]
```
deuszx, line 350: Does that end up being a Prometheus metric or just a log?

linera-service/src/proxy/main.rs
```diff
@@ -299,6 +300,15 @@ where
             GenesisConfigHashQuery => Ok(Some(RpcMessage::GenesisConfigHashResponse(Box::new(
                 self.genesis_config.hash(),
             )))),
+            UploadBlobContent(content) => {
+                let blob = Blob::new(*content);
+                let id = blob.id();
+                ensure!(
+                    self.storage.maybe_write_blobs(&[blob]).await?[0],
+                    "Blob not found"
```
deuszx, line 308: This ends up being a panic?

linera-core/src/chain_worker/state/attempted_changes.rs
```diff
@@ -298,33 +297,30 @@ where
...
         let required_blob_ids = executed_block.required_blob_ids();
-        // Verify that no unrelated blobs were provided.
-        self.state
-            .check_for_unneeded_blobs(&required_blob_ids, blobs)?;
-        let blobs = self.state.get_required_blobs(executed_block, blobs).await?;
-        let blobs = blobs.into_values().collect::<Vec<_>>();
-
-        let certificate_hash = certificate.hash();
+        let blobs_result = self.state.get_required_blobs(executed_block, &[]).await;
+        let blobs_result = blobs_result.map(|blobs| blobs.into_values().collect::<Vec<_>>());
```
deuszx, line 301: Can we chain the two lines instead of creating a variable with the same name? So
```rust
self.state.get_required_blobs(executed_block, &[]).await.map(|blobs| ...)
```

linera-core/src/chain_worker/state/attempted_changes.rs
```diff
@@ -298,33 +297,30 @@ where
...
 
         // Update the blob state with last used certificate hash.
         let blob_state = BlobState {
-            last_used_by: certificate_hash,
+            last_used_by: certificate.hash(),
             chain_id: block.chain_id,
             block_height,
             epoch: block.epoch,
         };
+        let overwrite = blobs_result.is_ok(); // Overwrite only if we wrote the certificate.
```
deuszx, line 317: Why would we want to _overwrite_ the blobs ever? Shouldn't that be idempotent?

linera-service/src/proxy/main.rs
afck: It returns an `Err`. I think the proxy will just return an error to the request then (otherwise the other `?`s in this function wouldn't make sense).

linera-rpc/src/grpc/client.rs
afck: Just a log.

linera-core/src/client/mod.rs
afck: I added a paragraph to the `Proposal` section in the PR description.

deuszx:
I can see that you've updated the tests but we should also have one where we publish a certificate, that gets rejected with `BlobsNotFound` after which we upload the blobs and try again.

linera-core/src/chain_worker/state/attempted_changes.rs
deuszx: Ah, I missed the _state_ part.

linera-rpc/src/grpc/client.rs
deuszx: Ok, I was thinking that it might be useful to track blobs' sizes.

deuszx (approved):
LGTM

linera-core/src/node.rs
```diff
@@ -103,6 +102,10 @@ pub trait ValidatorNode {
     /// Subscribes to receiving notifications for a collection of chains.
     async fn subscribe(&self, chains: Vec<ChainId>) -> Result<Self::NotificationStream, NodeError>;
 
+    // Uploads a blob content. Returns an error if the validator has not seen a
+    // certificate using this blob.
```
deuszx, line 106: Where do we do that? i.e. check that validator has not seen the certificate yet?

linera-core/src/chain_worker/state/attempted_changes.rs
```diff
@@ -298,33 +297,33 @@ where
         );
 
         let required_blob_ids = executed_block.required_blob_ids();
-        // Verify that no unrelated blobs were provided.
-        self.state
-            .check_for_unneeded_blobs(&required_blob_ids, blobs)?;
```
deuszx, line 303: How do you plan on making sure that only _needed_ blobs are uploaded?

linera-core/src/node.rs
afck: In proxy/grpc.rs:485: If the validator hasn't seen it yet, there's no blob state, so it doesn't write the blob.

linera-core/src/chain_worker/state/attempted_changes.rs
afck: The validator doesn't accept blobs for which there is no blob state. Only blobs that have been seen as part of a confirmed block certificate have a blob state.

linera-rpc/src/grpc/client.rs
afck: Probably, yes. We should consider adding that at some point.

linera-storage/src/db_storage.rs
```diff
@@ -576,6 +577,25 @@ where
         Ok(())
     }
 
+    async fn maybe_write_blobs(&self, blobs: &[Blob]) -> Result<Vec<bool>, ViewError> {
```
deuszx, line 580: nitpick: I'd drop the `maybe_prefix` and just explain in the comment that the `Vec<bool>` tracks which blobs from the input argument were stored/new.

linera-storage/src/db_storage.rs
```diff
@@ -576,6 +577,25 @@ where
...
+        }
+        let blob_state_keys = blobs
+            .iter()
+            .map(|blob| bcs::to_bytes(&BaseKey::BlobState(blob.id())))
+            .collect::<Result<_, _>>()?;
+        let blob_states = self.store.contains_keys(blob_state_keys).await?;
+        let mut batch = Batch::new();
+        for (blob, has_state) in blobs.iter().zip(&blob_states) {
+            if *has_state {
+                Self::add_blob_to_batch(blob, &mut batch)?;
```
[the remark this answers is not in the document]
afck: I agree, but that's for a separate PR.

linera-storage/src/db_storage.rs
deuszx: Sounds like the client could then figure out the blobs that need to be written and then call the `write_blobs`.

afck: It's just that we use this logic—check the blob states to figure out which ones to write—in multiple places, so I wanted to deduplicate it.

linera-service/src/proxy/main.rs
deuszx: I missed that this function returns `anyhow::Result`.

PR #3109, opened 2025-01-09
Pin ScyllaDB minor version.

christos-h:
## Motivation

Pin the ScyllaDB minor version to avoid accidentally pulling upstream changes.

## Proposal

Pin the image version in Docker Compose.

## Test Plan

CI will catch regressions.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -1,6 +1,6 @@
 services:
   scylla:
-    image: scylladb/scylla:6.1
+    image: scylladb/scylla:6.1.3
     container_name: scylla
     volumes:
       - linera-scylla-data:/var/lib/scylla


docker/docker-compose.yml
```diff
@@ -1,6 +1,6 @@
 services:
   scylla:
-    image: scylladb/scylla:6.1
+    image: scylladb/scylla:6.1.3
```
ndr-ds, line 3: On the co-work yesterday you mentioned something about pinning a commit version. Were you talking about this or are we not gonna do that here?

ndr-ds: Talked on Slack, will follow up with pinning the commit version

ndr-ds (approved):

PR #3114, opened 2025-01-10
Create a tool to extract a bash script from a Linera README file

### LINKED ISSUE #1456: Add a tool to extract bash commands from README files
# Motivation

Currently the [README test](https://github.com/linera-io/linera-protocol/blob/main/linera-service/tests/readme_test.rs) uses the `QuotedBashScript` utility to extract the commands from a README file, so that they can be tested.

This functionality would also be useful for development, to create a script from the README that quickly sets up a network with the deployed application.

# Proposal

* Move `QuotedBashScript` into an `extract_bash` binary of its own within `linera-service`.
* Apply it to all the example `README.md` files and produce a `run_readme.sh` in the same directory.
* Add a check to `rust.yml` so that CI verifies that all the `run_readme.sh` are in sync with their corresponding `README`.
* Simplify the README test so that it directly uses the scripts.

ma2bd:
## Motivation

This was suggested by the community and team members also hacked a local solution at times.

fixes #1456

## Proposal

* Some light refactoring
* Add a command `linera extract-script-from-markdown`. The command is hidden because we may arbitrarily change or remove it in the future.

## Test Plan

CI
```
cargo build
target/debug/linera extract-script-from-markdown examples/fungible/README.md | bash
```

## Release Plan

### DIFF
--- a/linera-service/src/util.rs
+++ b/linera-service/src/util.rs
@@ -88,6 +88,7 @@ where
     pub fn extract_bash_script_to(
         self,
         mut output: impl std::io::Write,
+        pause_after_linera_service: Option<Duration>,
         pause_after_gql_mutations: Option<Duration>,
     ) -> Result<(), std::io::Error> {
         let mut lines = self.buffer.lines();
@@ -108,9 +109,10 @@ where
                         quote += &line;
                         quote += "\n";
 
-                        if line.contains("linera service") {
-                            quote += "sleep 3";
-                            quote += "\n";
+                        if let Some(pause) = pause_after_linera_service {
+                            if line.contains("linera service") {
+                                quote += &format!("sleep {}\n", pause.as_secs());
+                            }
                         }
                     }
                     writeln!(output, "{}", quote)?;
@@ -133,13 +135,13 @@ where
                 )?;
                 writeln!(output, "QUERY_RESULT=$(curl -w '\\n' -g -X POST -H \"Content-Type: application/json\" -d \"$JSON_QUERY\" {uri} \
                         | tee /dev/stderr \
-                        | jq -e .data \n)"
+                        | jq -e .data)"
                 )?;
 
                 if let Some(pause) = pause_after_gql_mutations {
                     // Hack: let's add a pause after mutations.
                     if quote.starts_with("mutation") {
-                        writeln!(output, "\nsleep {}", pause.as_secs())?;
+                        writeln!(output, "sleep {}\n", pause.as_secs())?;
                     }
                 }
             }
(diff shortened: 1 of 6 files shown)


linera-service/src/util.rs
```diff
@@ -77,47 +69,29 @@ macro_rules! test_name {
...
-        Ok(Self { tmp_dir, path })
-    }
-
-    pub fn tmp_dir(&self) -> &Path {
-        self.tmp_dir.as_ref()
-    }
-
-    pub fn path(&self) -> &Path {
-        &self.path
+impl Markdown<std::io::BufReader<fs_err::File>> {
```
afck, line 76: (Nit: I'd `use std::io` and remove all those `std::`.)

linera-service/src/util.rs
```diff
@@ -77,47 +69,29 @@ macro_rules! test_name {
...
-        reader: impl std::io::BufRead,
+    pub fn extract_bash_script_to(
+        self,
+        mut output: impl std::io::Write,
+        pause_after_linera_service: Option<Duration>,
         pause_after_gql_mutations: Option<Duration>,
-    ) -> std::io::Result<Vec<String>> {
-        let mut result = Vec::new();
-        let mut lines = reader.lines();
+    ) -> Result<(), std::io::Error> {
```
afck, line 93:
```suggestion
    ) -> io::Result<()> {
```

linera-service/src/util.rs
```diff
@@ -153,24 +128,27 @@ impl QuotedBashAndGraphQlScript {
...
-                result.push(format!("QUERY=\"{}\"", quote.replace('"', "\\\"")));
-                result.push("JSON_QUERY=$(jq -n --arg q \"$QUERY\" '{\"query\": $q}')".to_string());
-                result.push(format!(
-                    "QUERY_RESULT=$(curl -w '\\n' -g -X POST -H \"Content-Type: application/json\" -d \"$JSON_QUERY\" {uri} \
+                writeln!(output, "QUERY=\"{}\"", quote.replace('"', "\\\""))?;
+                writeln!(
+                    output,
+                    "JSON_QUERY=$(jq -n --arg q \"$QUERY\" '{{\"query\": $q}}')"
+                )?;
+                writeln!(output, "QUERY_RESULT=$(curl -w '\\n' -g -X POST -H \"Content-Type: application/json\" -d \"$JSON_QUERY\" {uri} \
```
afck, line 136: (Line too long.)

afck (approved):

linera-service/src/util.rs
ma2bd: Not sure I like `use std::io` so much. I added a few others instead.

gorinaI: nice project!good idea!

Issue #2890, opened 2024-11-14
Convert the README tests to use a faucet


ma2bd:
We won't be able to control the `CHAIN_ID`s but that shouldn't be an issue after the latest updates.

PR #3115, opened 2025-01-13
Allow running a faucet from linera net up

ma2bd:
## Motivation

First step in addressing #2890

## Proposal

* Make a function of `find_owned_chains` out of the code in `ClientWrapper::drop()`
* Add an option `--with-faucet-chain N` to `linera net up` as well as `--faucet-port` and `--faucet-amount`
* Add a simple test

The only caveat is that `N` cannot be a chain-id (since we wouldn't know what to write) so we use the index amongst root chains (typically 1 for the first non-admin initial chain).

## Test Plan

CI

### DIFF
--- a/linera-service/src/cli_wrappers/wallet.rs
+++ b/linera-service/src/cli_wrappers/wallet.rs
@@ -1012,14 +1012,12 @@ impl Drop for ClientWrapper {
         let chain_ids = match self.find_owned_chains() {
             Ok(ids) => ids,
             Err(err) => {
-                warn!("Skipping closing chains because of error: {}", err);
+                warn!("Not closing chains because an error occurred: {}", err);
                 return;
             }
         };
-
         let binary_path = self.binary_path.lock().unwrap();
-        let binary_path = binary_path.as_ref().expect("Binary was used successfully");
-
+        let binary_path = binary_path.as_ref().expect("Binary was used before");
         let working_directory = self.path_provider.path();
 
         for chain_id in chain_ids {
--- a/linera-service/src/cli_wrappers/wallet.rs
+++ b/linera-service/src/cli_wrappers/wallet.rs
@@ -32,7 +32,7 @@ use serde::{de::DeserializeOwned, ser::Serialize};
 use serde_json::{json, Value};
 use tempfile::TempDir;
 use tokio::process::{Child, Command};
-use tracing::{info, warn};
+use tracing::{error, info, warn};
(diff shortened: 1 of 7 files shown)


linera-service/src/cli_wrappers/wallet.rs
```diff
@@ -101,6 +101,53 @@ impl ClientWrapper {
         }
     }
 
+    /// Runs `linera wallet show --short --owned` to find the list of chain ids in the
```
afck, line 104:
```suggestion
    /// Runs `linera wallet show --short --owned` to find the list of chain IDs in the
```

linera-service/src/cli_wrappers/wallet.rs
```diff
@@ -101,6 +101,53 @@ impl ClientWrapper {
...
+            wallet_show_command.arg(&*argument);
+        }
+
+        let Ok(wallet_show_output) = wallet_show_command
+            .current_dir(working_directory)
+            .args(["wallet", "show", "--short", "--owned"])
+            .output()
+        else {
+            bail!("Failed to execute `wallet show --short` to list chains in the wallet");
+        };
```
afck, line 129:
```suggestion
        .context("Failed to execute `wallet show --short` to list chains in the wallet")?;
```

linera-service/src/linera/net_up_utils.rs
```diff
@@ -286,9 +318,29 @@ async fn net_up(
...
     }
 
+    // Run the faucet,
+    let faucet_service = if let Some(faucet_chain) = with_faucet_chain {
+        let chain_ids = client
+            .find_owned_chains()
+            .expect("Cannot retrieve chains from linera wallet");
+        let chain_id = chain_ids.get(faucet_chain as usize).expect(
+            "Was expecting a number `faucet_chain` in the range `0..=num_other_initial_chains`",
+        );
```
afck, line 328: Why not `ChainID::root(faucet_chain)`? That's closer to what the documentation says (the Nth root chain), so it would still be true if we changed the order in which chains appeared in the wallet.

afck (approved):

linera-service/src/linera/net_up_utils.rs
ma2bd: 🤦 Yes of course

gorinaI: good

PR #3121, opened 2025-01-13
Fetch locked blobs individually, not in `ChainManagerInfo`.

afck:
## Motivation

Ultimately we want to transfer all blobs separately, rather than in a single message. (https://github.com/linera-io/linera-protocol/issues/3048)

This PR is one step towards that goal: When the client fetches the locked block from a validator, it now requests the corresponding blobs one by one, rather than all at once.

## Proposal

Add a `DownloadPendingBlob` endpoint; remove the locked blobs from the `ChainManagerInfo`.

## Test Plan

Existing tests exercise this scenario, e.g. `test_finalize_locked_block_with_blobs`. 

## Release Plan

## Links

- Part of #3048.

### DIFF
--- a/linera-core/src/chain_worker/actor.rs
+++ b/linera-core/src/chain_worker/actor.rs
@@ -14,7 +14,7 @@ use linera_base::{
     crypto::CryptoHash,
     data_types::{Blob, BlockHeight, Timestamp, UserApplicationDescription},
     hashed::Hashed,
-    identifiers::{ChainId, UserApplicationId},
+    identifiers::{BlobId, ChainId, UserApplicationId},
 };
 use linera_chain::{
     data_types::{Block, BlockProposal, ExecutedBlock, MessageBundle, Origin, Target},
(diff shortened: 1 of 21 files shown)


linera-core/src/chain_worker/actor.rs
```diff
@@ -142,6 +142,13 @@ where
         callback: oneshot::Sender<Result<(ChainInfoResponse, NetworkActions), WorkerError>>,
     },
 
+    /// Get a blob if it belongs to the current locked block or pending proposal.
+    DownloadPendingBlob {
```
deuszx, line 146: Just a thought - if there will be more than one "type" of blob being downloaded (pending, locked, proposed, etc.) then maybe we can have a single message `DownloadBlob` with a field denoting type/source for the blob.

deuszx (approved):

linera-core/src/chain_worker/actor.rs
afck: I was thinking about this, but went with the separate request, because `DownloadBlob` is handled directly by the proxy whereas `DownloadPendingBlob` has to be forwarded to a worker.

PR #3133, opened 2025-01-14
Fix pending change after flushing `RegisterView`

jvff:
## Motivation

@afck ran into an issue where an assertion caught unexpected pending changes after voting for a block proposal. After investigating the issue, the cause seems to be that `RegisterView::flush` had an edge case where it wouldn't clear it's `update` field.

## Proposal

Clear the `update` field after successfully flushing.

## Test Plan

A unit test was added to reproduce the issue.

## Release Plan

## Links

### DIFF
--- a/linera-views/src/views/register_view.rs
+++ b/linera-views/src/views/register_view.rs
@@ -99,6 +99,7 @@ where
             self.stored_value = value;
         }
         self.delete_storage_first = false;
+        self.update = None;
         Ok(delete_view)
     }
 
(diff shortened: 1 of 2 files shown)


afck (approved):
Thank you! :tada: :pray:

linera-views/src/views/register_view.rs
```diff
@@ -99,6 +99,7 @@ where
             self.stored_value = value;
         }
         self.delete_storage_first = false;
+        self.update = None;
```
ma2bd, line 102: Crazy that this bug survived so long

PR #3135, opened 2025-01-15
Make the chain manager a `View`.

afck:
## Motivation

Currently the chain manager is serialized as a whole and stored in a `RegisterView` in the chain state view. Since it contains blobs it can be very large, and the blobs are not needed every time the chain manager is loaded.

## Proposal

Make the chain manager a `View`, and put the blobs in a `MapView`.

## Test Plan

This doesn't change any logic, so CI should catch regressions. (In fact, it already did: https://github.com/linera-io/linera-protocol/pull/3133)

## Release Plan

## Links

- In preparation for: https://github.com/linera-io/linera-protocol/issues/3048

### DIFF
--- a/linera-chain/src/manager.rs
+++ b/linera-chain/src/manager.rs
@@ -499,6 +499,16 @@ where
         Ok(())
     }
 
+    /// Returns the requested blob if it belongs to the proposal or the locked block.
+    pub async fn pending_blob(&self, blob_id: &BlobId) -> Result<Option<Blob>, ViewError> {
+        if let Some(proposal) = self.proposed.get() {
+            if let Some(blob) = proposal.blobs.iter().find(|blob| blob.id() == *blob_id) {
+                return Ok(Some(blob.clone()));
+            }
+        }
+        self.locked_blobs.get(blob_id).await
+    }
+
     /// Updates `current_round` and `round_timeout` if necessary.
     ///
     /// This must be after every change to `timeout`, `locked` or `proposed`.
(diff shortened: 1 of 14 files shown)


linera-service-graphql-client/gql/service_schema.graphql
```diff
@@ -361,6 +401,14 @@ type Entry_AccountOwner_Amount_aaf96548 {
 	value: Amount
 }
 
+"""
+A GraphQL-visible map item, complete with key.
+"""
+type Entry_BlobId_Blob_9f0b41f3 {
```
deuszx, line 407: What's this weird suffix?

linera-chain/src/manager.rs
```diff
@@ -99,85 +106,93 @@ pub enum Outcome {
...
     /// Having a leader timeout certificate in any given round causes the next one to become
     /// current. Seeing a validated block certificate or a valid proposal in any round causes that
     /// round to become current, unless a higher one already is.
-    pub current_round: Round,
+    #[graphql(skip)]
+    pub current_round: RegisterView<C, Round>,
     /// The owners that take over in fallback mode.
-    #[debug(skip_if = BTreeMap::is_empty)]
-    pub fallback_owners: BTreeMap<Owner, (PublicKey, u64)>,
+    pub fallback_owners: RegisterView<C, BTreeMap<Owner, (PublicKey, u64)>>,
```
deuszx, line 164: That's a lot of mechanical work. I know it's not for this PR but it'd be nice if `derive(View)` wrapped all of the fields into propery `*View` variants.

afck: I don't know if that's possible in general. E.g. some but not all maps should turn into `MapView`s.

deuszx: sensible deafults + overrides with attributes.

PR #3137, opened 2025-01-15
fix: Optimized file Update build.rs

mdqst:
## Motivation

I noticed that using `format!` was unnecessary since `display()` already returns an object that can be passed directly to `File::create`.

I simplified and optimized the code to remove this redundancy.


ma2bd (approved):

ma2bd (changes requested):
~~This doesn't look correct actually. (Remove `.display()`)~~

afck:
@ma2bd: I think it's necessary: https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.display

ma2bd:
Thanks again for the PR. It is correct but I think using `format!` is just fine.

PR #3138, opened 2025-01-15
Use Keccak256 when hashing for signing

### LINKED ISSUE #3125: Switch hashing to Keccak

deuszx:
## Motivation

We want to verify Linera blocks on Ethereum (and other EVMs) but they don't have opcodes/precompiles for Sha256 that we're using .

## Proposal

Replace Sha256 with Keccak256.

Notice that I'm only changing hashing scheme used in `CryptoHash` struct - i.e. there are still places in the code where we use sha266 (like for example linera views and cache).

## Release Plan

This is backwards incompatible change.

## Links

Closes #3125 


linera-base/src/crypto.rs
```diff
@@ -300,24 +299,22 @@ impl TryFrom<&[u8]> for CryptoHash {
...
     fn from(integers: [u64; 4]) -> Self {
-        CryptoHash(u64_array_to_le_bytes(integers).into())
+        CryptoHash(u64_array_to_be_bytes(integers).into())
     }
 }
 
 impl From<CryptoHash> for [u64; 4] {
     fn from(crypto_hash: CryptoHash) -> Self {
-        le_bytes_to_u64_array(&crypto_hash.0)
+        be_bytes_to_u64_array(crypto_hash.0.as_ref())
```
afck, line 317: Is that so that a BCS-encoded `CryptoHash` has the same sequence of bytes as the `B256`?

deuszx: Keccak256 internally is big-endian so I thought we should be consistent with that.

slava1310:
Testnet

MathieuDutSik (commented):
I have only one objection to the PR.
We agree that SHA-3 is almost the same as Keccak256.
However, the implementations are distinct. Could you put a runtime comparison of the relative speed of SHA-3 vs Keccak256.

linera-base/src/crypto.rs
```diff
@@ -167,7 +166,7 @@ impl Serialize for CryptoHash {
         if serializer.is_human_readable() {
             serializer.serialize_str(&self.to_string())
         } else {
-            serializer.serialize_newtype_struct("CryptoHash", &self.0)
+            serializer.serialize_newtype_struct("CryptoHash", &self.0 .0)
```
afck, line 169: Maybe `CryptoHash` deserves an `as_bytes` method to avoid the `.0 .0`.

linera-base/src/crypto.rs
```diff
@@ -396,6 +393,20 @@ where
     }
 }
 
+// Temporary struct to extend `Keccak256` with `io::Write`.
```
afck, line 396:
```suggestion
/// Temporary struct to extend `Keccak256` with `io::Write`.
```

linera-base/src/crypto.rs
```diff
@@ -396,6 +393,20 @@ where
...
+struct Keccak256Ext(Keccak256);
```
afck, line 397: Is this used in places other than `CryptoHash::new`? Otherwise we might as well just call `update` directly there.

linera-base/src/crypto.rs
```diff
@@ -747,6 +754,18 @@ fn le_bytes_to_u64_array(bytes: &[u8]) -> [u64; 4] {
...
+    integers[0] = u64::from_be_bytes(bytes[0..8].try_into().expect("incorrect indices"));
+    integers[1] = u64::from_be_bytes(bytes[8..16].try_into().expect("incorrect indices"));
+    integers[2] = u64::from_be_bytes(bytes[16..24].try_into().expect("incorrect indices"));
+    integers[3] = u64::from_be_bytes(bytes[24..32].try_into().expect("incorrect indices"));
+
+    integers
+}
+
 /// Returns the bytes that represent the `integers` in little-endian.
 fn u64_array_to_le_bytes(integers: [u64; 4]) -> [u8; 32] {
```
afck, line 770: Are we still using this? (And _should_ we be using this at all?)

linera-base/src/identifiers.rs
```diff
@@ -1026,15 +1026,15 @@ mod tests {
     fn chain_ids() {
```
afck, line 1026: It might be helpful to add root chains 1 and 2 here as well: They are probably used in a few READMEs, so this serves as documentation where one can look up the hash. It might also simplify any future changes to how chain IDs are computed.

linera-execution/tests/test_execution.rs
```diff
@@ -1333,22 +1333,23 @@ async fn test_multiple_messages_from_different_applications() -> anyhow::Result<
 
     // The registration message for the first destination chain
```
afck, line 1334:
```suggestion
    // The registration message for the second destination chain
```

linera-execution/tests/test_execution.rs
```diff
@@ -1333,22 +1333,23 @@ async fn test_multiple_messages_from_different_applications() -> anyhow::Result<
...
         authenticated: false,
         grant: Amount::ZERO,
         kind: MessageKind::Simple,
         message: SystemMessage::RegisterApplications {
-            applications: vec![sending_target_description.clone(), caller_description],
+            applications: vec![sending_target_description.clone()],
         },
     };
+
     // The registration message for the second destination chain
```
afck, line 1345:
```suggestion
    // The registration message for the first destination chain
```

linera-execution/tests/test_execution.rs
afck: Or we could rename `first_message` and `second_message` to `message0` and `message1` and the destination chains similarly. (But _keep_ the `*_registration_message`s.)

linera-base/src/crypto.rs
deuszx: We hash in Signature::new which uses interface from Write trait.

deuszx:
> I have only one objection to the PR. We agree that SHA-3 is almost the same as Keccak256. However, the implementations are distinct.
> 
> Could you put a runtime comparison of the relative speed of SHA-3 vs Keccak256?

How is it relevant for the decision? We're not making a switch because of the performance reasons.

linera-base/src/crypto.rs
deuszx: It already has `as_bytes` but it will be `as_bytes().0` anyway (the second `.0` is for extracting `[u8; N]` from `FixedBytes`.

linera-execution/tests/test_execution.rs
deuszx: Right, sorry for that. I thought it's about the order in which these variables appear in the code - not about the destination chain.

linera-base/src/identifiers.rs
deuszx: Good idea. What I can see now is that we enumerate from `1` READMEs (i.e. `ChainId::root(0)` is `CHAINID_1`).

linera-base/src/crypto.rs
afck: Might be less confusing to use the same endianness there, too.

deuszx:
> I have only one objection to the PR. We agree that SHA-3 is almost the same as Keccak256. However, the implementations are distinct.
> 
> Could you put a runtime comparison of the relative speed of SHA-3 vs Keccak256?

@MathieuDutSik here's the comparison:
```
// keccak                  time:   [165.70 ns 165.77 ns 165.86 ns]
// Found 13 outliers among 100 measurements (13.00%)
// 5 (5.00%) high mild
// 8 (8.00%) high severe

// sha256                  time:   [141.33 ns 141.44 ns 141.55 ns]
// Found 6 outliers among 100 measurements (6.00%)
// 2 (2.00%) high mild
// 4 (4.00%) high severe
```

linera-base/src/crypto.rs
deuszx: Please create a separate issue for that.

linera-witty/src/runtime/wasmer/mod.rs
```diff
@@ -160,11 +162,13 @@ pub type ReentrantInstance<'a, UserData> = FunctionEnvMut<'a, Environment<UserDa
...
     type UserData = UserData;
-    type UserDataReference<'a> = MutexGuard<'a, UserData>
+    type UserDataReference<'a>
+        = MutexGuard<'a, UserData>
     where
         Self::UserData: 'a,
         Self: 'a;
-    type UserDataMutReference<'a> = MutexGuard<'a, UserData>
+    type UserDataMutReference<'a>
+        = MutexGuard<'a, UserData>
```
afck, line 171: Why are these necessary? Is your rustfmt version different from CI?

linera-core/benches/hashing_benchmarks.rs
```diff
@@ -0,0 +1,29 @@
...
+fn keccak_benchmark(c: &mut Criterion) {
+    c.bench_function("keccak", |b| b.iter(|| keccak(black_box(20))));
+}
+
+fn sha256_benchmark(c: &mut Criterion) {
+    c.bench_function("sha256", |b| b.iter(|| sha256(black_box(20))));
+}
+
+criterion_group!(benches, keccak_benchmark, sha256_benchmark);
+criterion_main!(benches);
```
afck, line 29: (Not sure if we need to commit this file. Not a blocker: We can always remove it again later.)

linera-base/src/crypto.rs
deuszx: Actually, even without that it'd require pretty big rewrite:
```rust
impl<'de, T, Hasher> Hashable<Hasher> for T
where
    T: BcsHashable<'de>,
    Hasher: io::Write,
{
    fn write(&self, hasher: &mut Hasher) {
        let name = <Self as HasTypeName>::type_name();
        // Note: This assumes that names never contain the separator `::`.
        write!(hasher, "{}::", name).expect("Hasher should not fail");
        bcs::serialize_into(hasher, &self).expect("Message serialization should not fail");
    }
}
```
write now we serialize `T` using the `Hashable` impl above. If I were to remove the wrapper (for which there exists a `io::Write` implementation then the `fn write` code above would have to be replicated in `CryptoHash::new` itself

linera-witty/src/runtime/wasmer/mod.rs
deuszx: I have no idea how that happened - I always run the same `cargo fmt` command.

deuszx: Yes yes, I do too, I don't know how this happened.

deuszx: Reverted.

MathieuDutSik (approved):

linera-core/benches/hashing_benchmarks.rs
Vatson17: Thank you very much for your work! We hope to need it someday.

PR #3139, opened 2025-01-15
Removed linera-admin Kubernetes service account.

christos-h:
## Motivation

We have a `linera-admin` Service Account which was used for testing. This is no longer necessary.

We may need to add this back in when we have autoscaling.

## Proposal

Remove the `linera-admin` service account.

## Test Plan

Kubernetes workflow will catch regressions.

## Release Plan


ndr-ds (approved):

ripon85464:
good

ripon85464:
ohhgxdf vxzwsa5tfgrds

Issue #3134, opened 2025-01-15
Fast block proposal is locked, but can be overwritten.


afck:
The fast block proposal counts as a locked block but is stored in the `proposed` field. If the same block is re-proposed in a later round, this gets overwritten, so the block isn't locked anymore.

We need to keep track of both the locked proposal from the `Fast` round _and_ the actual most recent proposal.

Also, just like with validated blocks, the client needs to be able to set its locked block to the `Fast` round proposal even if there are already proposals in later rounds.

Issue #2971, opened 2024-11-27
Take note of block proposals in the client without handling them.


afck:
[The client currently asks validators for pending block proposals and then handles them locally.](https://github.com/linera-io/linera-protocol/blob/3d05b4de633efef92f6a3e21e381257d9555951d/linera-core/src/client/mod.rs#L1715) This serves two purposes:
* If the proposal is in `Round::Fast`, it counts as locked, so the client _must_ re-propose that same block in a later round.
* If it is in `Round::MultiLeader(r)` and the client wants to propose a block, it _should_ not do that in the same round, but it can make a proposal in round `r + 1` without waiting for a timeout.

In the first case, the block cannot use any oracles.
In the second case, however, the block could be using oracles and therefore handling it locally could fail. On the other hand, we don't really care about the block itself there: We only need to know _that_ a proposal was made, not _which_ one.

Let's add a way for the client to keep track of proposals in the current round without executing them (but still checking the proposer's signature), so that the protocol works as intended in the second case even for blocks using oracles.

## Alternative: Use Timeouts?

We could also just rely on timeouts instead, and by default configure all multi-leader rounds to have timeout 0. That requires one more round-trip in the second case, but simplifies the protocol and brings the multi leader mode in line with the single leader one.

afck:
A test (maybe in the `client_tests` module) for this could look as follows:
* Players A and B play `hex-game` on a chain owned by both. (Regular owners, lots of multi-leader rounds.)
* Player A proposes a block in multi-leader round 0, the validators sign it, but then player A doesn't send (or the validators don't process) the validated block certificate.
* The clock is then advanced by more than the `start_time` configured in the `hex-game` settings (60 seconds by default).
* Now player B should be able to claim victory. However, B's client needs to know that it must propose in multi-leader round 1. This would currently fail, because it would fail to acknowledge A's proposal due to the `assert_before` call in it.

PR #3140, opened 2025-01-15
Remember a fast block proposal: it remains locked even if re-proposed.

afck:
## Motivation

In the fast round, the validators sign to confirm right away, so the block proposal must count as locked.

However, if the same block is re-proposed in a later round, the `proposed` field is overwritten and, due to the different round, is now not considered locked anymore.

## Proposal

Keep track of the fast round proposal as part of the `locked` field, too.

## Test Plan

The problem was reproduced by extending `test_fast_proposal_is_locked`.

## Release Plan

## Links

- Fixes https://github.com/linera-io/linera-protocol/issues/3134.

### DIFF
--- a/linera-chain/src/manager.rs
+++ b/linera-chain/src/manager.rs
@@ -108,8 +108,8 @@ pub enum Outcome {
 
 pub type ValidatedOrConfirmedVote<'a> = Either<&'a Vote<ValidatedBlock>, &'a Vote<ConfirmedBlock>>;
 
-/// The current locked block: Validators are only allowed to sign any proposal with a different
-/// block if they see a `ValidatedBlock` certificate with a higher round.
+/// The current locked block: Validators are allowed to sign a different block (from the locked
+/// block) iff they see a `ValidatedBlockCertificate` for it with a higher round.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[cfg_attr(with_testing, derive(Eq, PartialEq))]
 pub enum LockedBlock {
--- a/linera-chain/src/manager.rs
+++ b/linera-chain/src/manager.rs
@@ -450,13 +450,17 @@ where
             self.set_locked(LockedBlock::Fast(proposal.clone()), blobs)?;
         }
 
+        // If we are a client, we record the proposed block, in case it affects the current
+        // round number. That way, we don't make another proposal in the same round.
         let Some(key_pair) = key_pair else {
-            // Record the proposed block, in case it affects the current round number.
-            self.set_proposed(proposal);
-            self.update_current_round(local_time);
+            if proposal.content.round < Round::SingleLeader(0) {
+                self.set_proposed(proposal);
+                self.update_current_round(local_time);
+            }
             return Ok(None);
         };
 
+        // Otherwise we are a validator:
         self.check_proposal_round(&proposal)?;
         // Record the proposed block, so it can be supplied to clients that request it.
         self.set_proposed(proposal);
--- a/linera-chain/src/manager.rs
+++ b/linera-chain/src/manager.rs
@@ -404,13 +404,15 @@ where
         // We don't compare to `current_round` here: Non-validators must update their locked block
         // even if it is older than the current round. Validators will only sign in the current
         // round, though. (See `create_final_vote` below.)
-        if let Some(LockedBlock::Regular(locked)) = self.locked.get() {
-            if locked.hash() == certificate.hash() && locked.round == certificate.round {
-                return Ok(Outcome::Skip);
+        if let Some(locked) = self.locked.get() {
+            if let LockedBlock::Regular(locked_cert) = locked {
+                if locked_cert.hash() == certificate.hash() && locked.round() == certificate.round {
+                    return Ok(Outcome::Skip); // We already handled this certificate.
+                }
             }
             ensure!(
-                new_round > locked.round,
-                ChainError::InsufficientRoundStrict(locked.round)
+                new_round > locked.round(),
+                ChainError::InsufficientRoundStrict(locked.round())
             );
         }
         Ok(Outcome::Accept)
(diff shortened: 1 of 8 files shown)


linera-chain/src/manager.rs
```diff
@@ -108,6 +108,35 @@ pub enum Outcome {
 
 pub type ValidatedOrConfirmedVote<'a> = Either<&'a Vote<ValidatedBlock>, &'a Vote<ConfirmedBlock>>;
 
+/// The current locked block: Validators are only allowed to sign any proposal with a different
+/// block if they see a `ValidatedBlock` certificate with a higher round.
```
deuszx, line 112: Do you mean:
_Validators are allowed to sign a different block (from the locked block) iff they see a `ValidatedBlockCertificate` for it with a higher round_

if so, consider replacing the comment as my version reads better (IMHO ofc).

linera-chain/src/manager.rs
```diff
@@ -404,7 +404,7 @@ where
         // We don't compare to `current_round` here: Non-validators must update their locked block
         // even if it is older than the current round. Validators will only sign in the current
         // round, though. (See `create_final_vote` below.)
-        if let Some(locked) = self.locked.get() {
+        if let Some(LockedBlock::Regular(locked)) = self.locked.get() {
```
deuszx, line 407: What if we have a `::Fast` block here? Should that be unreachable?

linera-chain/src/manager.rs
```diff
@@ -424,51 +424,60 @@ where
...
-                .map_or(true, |locked| locked.round < lite_cert.round)
+                .map_or(true, |locked| locked.round() < lite_cert.round)
             {
                 let value = Hashed::new(ValidatedBlock::new(executed_block.clone()));
                 if let Some(certificate) = lite_cert.clone().with_value(value) {
-                    self.set_locked(certificate, blobs)?;
+                    self.set_locked(LockedBlock::Regular(certificate), blobs)?;
                 }
             }
+        } else if proposal.content.round.is_fast() {
```
deuszx, line 446: nitpick: you already have the `let round = ...` above. Either reuse it or remove the variable and inline.

linera-chain/src/manager.rs
afck: Not unreachable, but if our locked block is in round `Fast`, then `new_round` will always be greater: `Fast` is the lowest round, and validated block certificates only exist in other rounds.

deuszx: OK. My point is - we're silently ignoring that variant here. I'd be more comfortable if we pattern match on both variants and handle them explicitly. For `Fast` variant maybe add some assertions/comments. The code is quite complex already.

linera-chain/src/manager.rs
afck: Done in https://github.com/linera-io/linera-protocol/pull/3140/commits/3876f9ad5dd59e4c6e4121e8b9078590fa627c81.

linera-chain/src/manager.rs
afck: Done in https://github.com/linera-io/linera-protocol/pull/3140/commits/3876f9ad5dd59e4c6e4121e8b9078590fa627c81.

I just do the check in either case, even though if the locked block is fast it will be trivially satisfied.

linera-chain/src/manager.rs
```diff
@@ -424,51 +426,60 @@ where
...
         }
 
+        let Some(key_pair) = key_pair else {
+            // Record the proposed block, in case it affects the current round number.
+            self.set_proposed(proposal);
+            self.update_current_round(local_time);
+            return Ok(None);
+        };
+
+        self.check_proposal_round(&proposal)?;
```
deuszx, line 460: So we check the proposal's vailidity w.r.t. to the round only when we're actively validating? i.e. shouldn't this check be before the `let Some(key_pair) = key_pair else { ... }` which will return early?

linera-chain/src/manager.rs
```diff
@@ -424,51 +426,60 @@ where
...
         // Record the proposed block, so it can be supplied to clients that request it.
-        self.proposed.set(Some(proposal));
+        self.set_proposed(proposal);
```
deuszx, line 462: Can we move this (and line below) to line 452?

linera-chain/src/manager.rs
afck: No, because even if the proposal isn't from the current round, on the client side we still want to add it to the `proposed` field so that in multi leader rounds we know we shouldn't make a proposal in the same round ourselves. (This _partially_ addresses https://github.com/linera-io/linera-protocol/issues/2971.)

linera-chain/src/manager.rs
afck: No, we can't update the current round before checking against the current round.

linera-chain/src/manager.rs
deuszx: That's confusing TBH. That the logic here is makes these distinction b/c of some hidden requirements (i.e. being invoked on client vs server side).

linera-chain/src/manager.rs
deuszx: Yes, this was conditional on the acceptance of the previous comment.

linera-chain/src/manager.rs
afck: I agree. As I mentioned on Slack, I think we should make that distinction more explicit. But I think that will be a bigger change, and I'm not sure if it wouldn't make this PR even harder to review. Happy to give it a try, though, if you prefer.

deuszx: No, that's fine. Just maybe leave that part in addition to the comment above (in the else { ... } clause).

deuszx (approved):

linera-chain/src/manager.rs
afck: Actually, I'll also add another check so that we only add the proposal if it's before the single leader rounds. For those we don't need it, and we also shouldn't accept it before the round has begun. Sorry for the confusion! I hope we can separate some of the client vs. validator logic soon, and clarify all of this.

Issue #1069, opened 2023-09-25
Clarify the schema of `ScyllaDb`


MathieuDutSik:
The schema of `ScyllaDb` appears too complicated with the use of `dummy`.
We need to see if we can improve that and if not why is it so.

PR #3142, opened 2025-01-16
Implement `list_root_keys` to the `AdminKeyValueStore`

### LINKED ISSUE #3085: Introduce the `get_root_keys` to the `AdminKeyValueStore`
We are currently using the `root_key` in the `AdminKeyValueStore`. This is a formalization of the notion of partition keys in databases.

What is missing is the ability to get the list of root keys. That is we would like to have
```rust
fn get_root_keys(&self) -> Result<Vec<Vec<u8>>, Self::Error>
```

There are two use cases for that:
* Accessing to the list of chains of a validator.
* Migrating data from one storage to another.

There are obstacles to that:
* For DynamoDb and ScyllaDb, we could access the root keys easily since they are partition keys.
* The `RocksDb` does not have a notion of partition key.
  - Instead, we simply concatenate the keys like `root_key + key`.
  - This design was chosen because having one RocksDB instance per root key would lead to an overflow in the number of opened files.
  - One idea would be to add some entry indicating the creation of a new root key when doing a `clone_with_root_key`. This would make the `clone_with_root_key` async but that is fine.
* The support in storage service would also be somewhat difficult. But a priori possible.

MathieuDutSik:
## Motivation

We need to have the feature of accessing to the list of root keys in order to have mutation function and accessing to the list of chain ids of a storage.

Fixes #3085 

## Proposal

The implementation causes some problems:
* For ScyllaDb the implementation is very easy since the root_key is used as a partition key.
* For DynamoDb we can implement the root_key as a partition key, but DynamoDb forbids iterating and determining all the partition keys. So, we need to keep track of the keys.
* For RocksDb / StorageService / IndexedDb we need to keep track of the list of root_keys.
* For RocksDb, this led to a simplification since the edge case of having a key of the form `[255, ..., 255]` disappears. This corrects a problem and a test is added to detect it.
* The result of the `list_root_keys` will not be the same on different storage. If storage has been created with `fn create`, some keys were written but later deleted then in DynamoDb, RocksDb, storage-service, IndexedDb the root key will show up as existing while in ScyllaDb, the root key will not be visible.

The writing of the root key occurs when a `write_batch` is done.

## Test Plan

One test has been added for this feature.

## Release Plan

No impact on the TestNet / DevNet. It can follow the normal release plan.

## Links

None.


linera-views/src/backends/dynamo_db.rs
```diff
@@ -92,11 +92,19 @@ async fn get_config_internal(
     }
 }
 
+/// The DynamoDb forbids the iteration over the partition keys.
```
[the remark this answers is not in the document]
MathieuDutSik: Ok, thanks.

linera-views/src/backends/dynamo_db.rs
```diff
@@ -92,11 +92,19 @@ async fn get_config_internal(
...
+/// Therefore we use a special partition_key named [1] for storing
+/// the root keys. For normal root_keys, we simply put a 0 in front
+/// therefore no intersection is possible.
```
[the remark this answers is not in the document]
MathieuDutSik: Yes, corrected.

linera-views/src/backends/dynamo_db.rs
```diff
@@ -159,9 +167,9 @@ const MAX_TRANSACT_WRITE_ITEM_SIZE: usize = 100;
 
 /// Keys of length 0 are not allowed, so we extend by having a prefix on start
 fn extend_root_key(root_key: &[u8]) -> Vec<u8> {
-    let mut vec = vec![0];
-    vec.extend(root_key);
-    vec
+    let mut start_key = vec![0];
```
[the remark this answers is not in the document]
MathieuDutSik: Ok.

linera-views/src/backends/dynamo_db.rs
```diff
@@ -159,9 +167,9 @@ const MAX_TRANSACT_WRITE_ITEM_SIZE: usize = 100;
...
+    start_key.extend(root_key);
+    start_key
```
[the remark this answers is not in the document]
MathieuDutSik: I am actually against that. I first used the term `extended_root_key` but it is not great. The `start_key` is short enough to be easy to use. It is also used in the backends in a fairly uniform way with the same meaning when it is used.

The inconvenience of `prefix` and `key_prefix` is that they are used for searching purposes. I prefer not to have a collision of the meaning of names.

linera-views/src/backends/dynamo_db.rs
```diff
@@ -308,7 +314,7 @@ impl TransactionBuilder {
         value: Vec<u8>,
         store: &DynamoDbStoreInternal,
     ) -> Result<(), DynamoDbStoreInternalError> {
-        let transact = store.build_put_transact(&self.root_key, key, value)?;
+        let transact = store.build_put_transact(&self.start_key, key, value)?;
```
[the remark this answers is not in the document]
MathieuDutSik: Ok, I used `transaction` in the DynamoDb code.

linera-client/src/client_options.rs
```diff
@@ -964,6 +964,14 @@ pub enum DatabaseToolCommand {
         #[arg(long = "storage")]
         storage_config: String,
     },
+
+    /// List the rot keys of the database
```
afck, line 968:
```suggestion
    /// List the root keys of the database
```

linera-client/src/storage.rs
```diff
@@ -554,6 +554,32 @@ impl StoreConfig {
             }
         }
     }
+
+    /// Lists all root keys of the storage
+    pub async fn list_root_keys(self) -> Result<Vec<Vec<u8>>, ViewError> {
+        match self {
+            StoreConfig::Memory(_, _) => Err(ViewError::StoreError {
+                backend: "memory".to_string(),
+                error: "list_all is not supported for the memory storage".to_string(),
```
afck, line 563:
```suggestion
                error: "list_all_root_keys is not supported for the memory storage".to_string(),
```

linera-storage-service/proto/key_value_store.proto
```diff
@@ -139,6 +139,24 @@ message ReplyListAll {
...
+}
+
+
+message RequestInsertRootKey {
+  bytes namespace = 1;
+  bytes root_key = 2;
+}
+
+message ReplyInsertRootKey {
+}
```
afck, line 157: (We could use `google.protobuf.Empty` instead of all those empty types. Not sure if that's preferable, but we do it in `rpc.proto`.)

linera-views/tests/admin_tests.rs
```diff
@@ -7,27 +7,53 @@ use linera_views::dynamo_db::DynamoDbStore;
...
+    root_key_admin_test::<RocksDbStore>().await;
+}
+
+#[cfg(with_dynamodb)]
+#[tokio::test]
+async fn root_key_admin_test_dynamo_db() {
+    root_key_admin_test::<DynamoDbStore>().await;
+}
+
+#[cfg(with_scylladb)]
```
afck, line 55: Might be more readable to use `test_case` for this.

linera-views/src/store.rs
```diff
@@ -174,11 +174,18 @@ pub trait LocalAdminKeyValueStore: WithError + Sized {
...
 
     /// Takes a connection and creates a new one with a different `root_key`.
-    fn clone_with_root_key(&self, root_key: &[u8]) -> Result<Self, Self::Error>;
+    async fn clone_with_root_key(&self, root_key: &[u8]) -> Result<Self, Self::Error>;
 
     /// Obtains the list of existing namespaces.
     async fn list_all(config: &Self::Config) -> Result<Vec<String>, Self::Error>;
 
+    /// Lists the root keys of the namespace.
+    /// It is possible that some root_keys have no keys.
```
afck, line 183: Do you mean "some namespaces have no root keys"?

afck (commented):
Sorry, I think I need some more explanations (or better: code comments, or pointers where to find them) before I can review this. I'm confused how the keys are structured and represented in the different databases. E.g. I thought the first byte was there to distinguish between actual key-value pairs, and markers for the existence of namespaces; in what way does that relate to root keys at all, and why is it changing now?

linera-client/src/client_options.rs
MathieuDutSik: Thank you.

linera-client/src/storage.rs
MathieuDutSik: Thanks.

linera-views/src/store.rs
MathieuDutSik: No, I really mean that.
What happens is that for each key, there is at least one root key.
But for some root key, there is no corresponding key.
* For ScyllaDb, we will have that the insertion is done via `(root_ley, key)` which lead to the fact that for each key we have a matching root_key.
* For others the root key has to be stored as a key. That creation is done just at `fn connect` or `fn clone_with_root_key`. So, the creation itself creates root keys. But the key is not yet written.

linera-views/src/test_utils/mod.rs
```diff
@@ -700,3 +703,70 @@ where
...
+        }
+    }
+
+    let read_root_keys = S::list_root_keys(&config, &namespace)
+        .await
+        .expect("read_root_keys");
+    let set_root_keys = root_keys.iter().cloned().collect::<HashSet<_>>();
+    for read_root_key in &read_root_keys {
+        assert!(set_root_keys.contains(read_root_key));
+    }
```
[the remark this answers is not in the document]
MathieuDutSik: No.
For a database like ScyllaDb, the root keys would be inserted only if there is a corresponding key. This is why there is a `size_select` that could be 0.

linera-views/src/backends/scylla_db.rs
```diff
@@ -715,6 +718,33 @@ impl AdminKeyValueStore for ScyllaDbStoreInternal {
...
+        // See TODO(#1069).
+        let query = format!("SELECT root_key FROM kv.{} ALLOW FILTERING", namespace);
+
+        // Execute the query
+        let rows = session.query_iter(query, &[]).await?;
+        let mut rows = rows.rows_stream::<(Vec<u8>,)>()?;
+        let mut root_keys = HashSet::new();
+        while let Some(row) = rows.next().await {
+            let (root_key,) = row?;
+            let root_key = root_key[1..].to_vec();
```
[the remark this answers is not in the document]
MathieuDutSik: In ScyllaDb, a key has to hava non-zero length. Therefore, we have a function `get_big_root_key` that adds a 0 at the first entry.

linera-views/src/backends/rocks_db.rs
```diff
@@ -193,52 +188,28 @@ impl RocksDbStoreExecutor {
...
-                    if let Excluded(upper_bound) = get_upper_bound(&key_prefix) {
-                        let mut full_key1 = self.root_key.to_vec();
-                        full_key1.extend(key_prefix);
-                        let mut full_key2 = self.root_key.to_vec();
-                        full_key2.extend(upper_bound);
-                        inner_batch.delete_range(&full_key1, &full_key2);
-                    }
+                    let mut full_key1 = self.start_key.to_vec();
+                    full_key1.extend(&key_prefix);
+                    let full_key2 = get_upper_bound_option(&full_key1).unwrap();
```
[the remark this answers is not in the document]
MathieuDutSik: Yes, exactly.
Added, documentation.

linera-views/tests/admin_tests.rs
MathieuDutSik: Yes, thank you.

linera-storage-service/proto/key_value_store.proto
MathieuDutSik: Thank you, that is great.

linera-views/src/backends/rocks_db.rs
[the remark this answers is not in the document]
ma2bd: You may remove the comment and replace `.unwrap()` by `.expect("the first entry cannot be 255");`

linera-views/src/backends/rocks_db.rs
```diff
@@ -96,7 +91,7 @@ fn check_key_size(key: &[u8]) -> Result<(), RocksDbStoreInternalError> {
 #[derive(Clone)]
 struct RocksDbStoreExecutor {
     db: Arc<DB>,
-    root_key: Vec<u8>,
+    start_key: Vec<u8>,
```
ma2bd, line 94: what is a `start_key`? Seems like we're missing field documentation here.

linera-views/src/backends/rocks_db.rs
```diff
@@ -425,19 +401,30 @@ impl AdminKeyValueStore for RocksDbStoreInternal {
...
     }
 
-    fn clone_with_root_key(&self, root_key: &[u8]) -> Result<Self, RocksDbStoreInternalError> {
+    async fn clone_with_root_key(
+        &self,
+        root_key: &[u8],
+    ) -> Result<Self, RocksDbStoreInternalError> {
         let mut store = self.clone();
-        store.executor.root_key = root_key.to_vec();
+        store.executor.start_key = vec![1];
```
ma2bd, line 421: I feel that we should define two global constants instead of `0` and `1`.

ma2bd: The notion of "key space" or "domain" comes to mind.

linera-views/src/backends/scylla_db.rs
```diff
@@ -715,6 +718,33 @@ impl AdminKeyValueStore for ScyllaDbStoreInternal {
...
+    ) -> Result<Vec<Vec<u8>>, ScyllaDbStoreInternalError> {
+        Self::check_namespace(namespace)?;
+        let session = SessionBuilder::new()
+            .known_node(config.uri.as_str())
+            .build()
+            .boxed()
+            .await?;
+        // Create a table if it doesn't exist
+        // The schema appears too complicated for non-trivial reasons.
+        // See TODO(#1069).
```
ma2bd, line 733: #1069 is already closed. Do we need to the TODO?

linera-client/src/client_options.rs
```diff
@@ -964,6 +964,14 @@ pub enum DatabaseToolCommand {
         #[arg(long = "storage")]
         storage_config: String,
     },
+
+    /// List the root keys of the database
+    #[command(name = "list_root_keys")]
+    ListRootKeys {
```
ma2bd, line 970: In `linera-client`, we know that the "root keys" are (basically) chain ids. So shouldn't we list the chain ids instead?

ma2bd: And similarly we should list ApplicationId, BlobId, etc as such.

ma2bd: Btw you may need to add conversion function in `linera-storage` for this.

linera-views/src/backends/rocks_db.rs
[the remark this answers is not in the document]
MathieuDutSik: Ok.

linera-views/src/backends/rocks_db.rs
MathieuDutSik: It is a fact that what is used externally as `root_key` does not translate exactly like that in the specific code. Building it everytime it is needed would cost a little, which is why it is better to store the key that is going to be used.

We need to add something and it varies according to the context. This happens for `RocksDb`, `DynamoDb`, `IndexDb`, `LineraStorageService`. Description could be added but there would not be unifying explanation as this is all case specific.

About the name, I thought `effective_root_key`, `extended_root_key`. But they are quite long. Which is why I liked `start_key` which is short and reasonable in what it expresses.

linera-views/src/backends/rocks_db.rs
MathieuDutSik: Ok, except for the second where I prefer `STORE_ROOT_KEYS_PREFIX`.

linera-views/src/backends/scylla_db.rs
MathieuDutSik: Yes, thank you.

linera-client/src/client_options.rs
MathieuDutSik: Yes, this is why I wanted to do it in a subsequent PR.

linera-views/src/backends/scylla_db.rs
```diff
@@ -715,6 +718,30 @@ impl AdminKeyValueStore for ScyllaDbStoreInternal {
...
+            .known_node(config.uri.as_str())
+            .build()
+            .boxed()
+            .await?;
+        let query = format!("SELECT root_key FROM kv.{} ALLOW FILTERING", namespace);
+
+        // Execute the query
+        let rows = session.query_iter(query, &[]).await?;
+        let mut rows = rows.rows_stream::<(Vec<u8>,)>()?;
+        let mut root_keys = HashSet::new();
```
afck, line 736: If we use a `BTreeSet`, it will return the keys sorted, rather than in a random order.

linera-views/src/backends/dynamo_db.rs
```diff
@@ -91,11 +92,17 @@ async fn get_config_internal(
     }
 }
 
+/// DynamoDb forbids the iteration over the partition keys.
+/// Therefore we use a special partition_key named [1] for storing
+/// the root keys. For normal root_keys, we simply put a 0 in front
```
afck, line 97:
```suggestion
/// DynamoDB forbids the iteration over the partition keys.
/// Therefore we use a special partition key named `[1]` for storing
/// the root keys. For normal root keys, we simply put a `[0]` in front
```

linera-views/src/backends/dynamo_db.rs
```diff
@@ -352,28 +370,34 @@ impl AdminKeyValueStore for DynamoDbStoreInternal {
...
+        store.write_root_key(root_key).await?;
+        Ok(store)
     }
 
-    fn clone_with_root_key(&self, root_key: &[u8]) -> Result<Self, DynamoDbStoreInternalError> {
+    async fn clone_with_root_key(
+        &self,
+        root_key: &[u8],
+    ) -> Result<Self, DynamoDbStoreInternalError> {
+        self.write_root_key(root_key).await?;
```
afck, line 389: Does that mean `DbStorage::load_chain` will now write a root key?

E.g. if we handle a certificate only to realize during execution that it's invalid, it would still write that chain state view's root key to storage?

linera-views/src/backends/rocks_db.rs
```diff
@@ -425,19 +405,30 @@ impl AdminKeyValueStore for RocksDbStoreInternal {
...
+    async fn clone_with_root_key(
+        &self,
+        root_key: &[u8],
+    ) -> Result<Self, RocksDbStoreInternalError> {
         let mut store = self.clone();
-        store.executor.root_key = root_key.to_vec();
+        store.executor.start_key = STORED_ROOT_KEYS_PREFIX.to_vec();
+        let mut batch = Batch::new();
+        batch.put_key_value_bytes(root_key.to_vec(), vec![]);
+        store.write_batch(batch).await?;
```
afck, line 428: (See ScyllaDB: Not sure we should write here.)

linera-views/src/backends/dynamo_db.rs
MathieuDutSik: Ok.

linera-views/src/backends/dynamo_db.rs
MathieuDutSik: Yes.
That is the design with the root keys. The running of the system would still be deterministic.

linera-views/src/backends/rocks_db.rs
MathieuDutSik: Can you precise?
RocksDD does not have any sort of partition key. So, we need to implement them manually.

linera-views/src/backends/scylla_db.rs
MathieuDutSik: That is true, that does not really matter, but we can do it that way.

linera-views/src/backends/dynamo_db.rs
afck: But we're calling `load_chain` quite early in the validation process: You could spam a validator with completely invalid certificates with random bogus chain IDs, and it would create a database entry for each of these IDs.

ma2bd: Yes, this sounds much better to delay the registration of a root view to the first `write_batch`.

MathieuDutSik: Ok, done.

linera-views/src/backends/dynamo_db.rs
```diff
@@ -972,6 +958,25 @@ impl DirectWritableKeyValueStore for DynamoDbStoreInternal {
     type Batch = SimpleUnorderedBatch;
 
     async fn write_batch(&self, batch: Self::Batch) -> Result<(), DynamoDbStoreInternalError> {
+        let write_root_key = {
+            let mut root_key_written = self.root_key_written.lock().expect("root_key_written should not be poisoned");
+            if !*root_key_written {
+                *root_key_written = false;
```
afck, line 968: Did you mean to set it to `true`?

linera-storage-service/src/common.rs
```diff
@@ -28,6 +28,8 @@ pub enum KeyTag {
     Key = MIN_VIEW_TAG,
     /// Prefix for the storage of existence or not of the namespaces.
     Namespace,
+    /// Prefix for the root key
+    RootKey,
```
ma2bd, line 32: The first variant should probably be called `RootView`.
* "Prefix for the storage of the keys of the map" -> "Key prefix for the storage of the root views"
* etc

linera-views/src/backends/dynamo_db.rs
```diff
@@ -91,11 +96,17 @@ async fn get_config_internal(
...
+/// DynamoDB forbids the iteration over the partition keys.
+/// Therefore we use a special partition key named `[1]` for storing
+/// the root keys. For normal root keys, we simply put a `[0]` in
+/// front therefore no intersection is possible.
+const PARTITION_KEY_ROOT_KEY: &[u8] = &[1];
+
 /// The attribute name of the partition key.
 const PARTITION_ATTRIBUTE: &str = "item_partition";
 
 /// A root key being used for testing existence of tables
```
ma2bd, line 108: Is this a "key" or a "key prefix"?

linera-storage-service/src/client.rs
```diff
@@ -240,8 +240,20 @@ impl WritableKeyValueStore for ServiceStoreClientInternal {
             return Ok(());
         }
         let mut statements = Vec::new();
+        let mut full_key = self.start_key.clone();
+        full_key[0] = KeyTag::RootKey as u8;
+        let operation = Operation::Put(KeyValue {
+            key: full_key,
+            value: vec![],
+        });
```
ma2bd, line 248: It may not be ideal to overwrite this key over and over. (E.g. it might evict other keys faster from any cache)
Would it be possible to only schedule one such write per connect?

MathieuDutSik: Yes, we can do that.
Only write at the first, this requires a `Arc<Mutex<bool>>` which I did for DynamoDb, but not for others as I thought it was needed for DynamoDb.

linera-views/src/backends/rocks_db.rs
```diff
@@ -193,55 +192,35 @@ impl RocksDbStoreExecutor {
...
+                    full_key1.extend(&key_prefix);
+                    let full_key2 =
+                        get_upper_bound_option(&full_key1).expect("the first entry cannot be 255");
+                    inner_batch.delete_range(&full_key1, &full_key2);
                 }
             }
         }
+        let mut full_key = self.start_key.to_vec();
+        full_key[0] = STORED_ROOT_KEYS_PREFIX;
+        inner_batch.put(&full_key, vec![]);
```
ma2bd, line 223: Same. Do we have to write this every time?

ma2bd: That one is called "keys_prefix" lol

linera-views/src/backends/dynamo_db.rs
MathieuDutSik: Yes, thank you.

linera-views/src/backends/dynamo_db.rs
MathieuDutSik: It is a key.
The partition key coming from root keys start with 0, the partition key storing the root keys is exactly `[1]`.

linera-views/src/backends/rocks_db.rs
MathieuDutSik: Corrected. Now we write every time only for IndexedDB.

linera-storage-service/src/client.rs
```diff
@@ -54,14 +57,18 @@ const MAX_KEY_SIZE: usize = 1000000;
...
+//   to indicate the existence of a root key.
 #[derive(Clone)]
 pub struct ServiceStoreClientInternal {
     channel: Channel,
     semaphore: Option<Arc<Semaphore>>,
     max_stream_queries: usize,
-    namespace: Vec<u8>,
+    prefix_len: usize,
     start_key: Vec<u8>,
+    root_key_written: Arc<Mutex<bool>>,
```
afck, line 71: Maybe an `AtomicBool` would be lower overhead. Also, you could do `.fetch_or(true)` below to set and get the value atomically.

linera-storage-service/src/client.rs
```diff
@@ -308,7 +337,7 @@ impl ServiceStoreClientInternal {
     }
 
     fn namespace_as_vec(namespace: &str) -> Result<Vec<u8>, ServiceStoreError> {
-        let mut key = vec![KeyTag::Key as u8];
+        let mut key = vec![];
         bcs::serialize_into(&mut key, namespace)?;
         Ok(key)
```
afck, line 342: I think that's equivalent to:
```suggestion
        Ok(bcs::to_bytes(namespace)?)
```

Alternatively, maybe this should take a `u8` as an argument to prepend to the namespace?

linera-views/src/backends/dynamo_db.rs
```diff
@@ -320,7 +327,8 @@ pub struct DynamoDbStoreInternal {
     namespace: String,
     semaphore: Option<Arc<Semaphore>>,
     max_stream_queries: usize,
-    root_key: Vec<u8>,
+    start_key: Vec<u8>,
+    root_key_written: Arc<Mutex<bool>>,
```
afck, line 331: Same here: Should probably be `AtomicBool`.

linera-views/src/backends/dynamo_db.rs
```diff
@@ -400,6 +411,24 @@ impl AdminKeyValueStore for DynamoDbStoreInternal {
...
     }
 
+    async fn list_root_keys(
+        config: &Self::Config,
+        namespace: &str,
+    ) -> Result<Vec<Vec<u8>>, DynamoDbStoreInternalError> {
+        let mut store = Self::connect(config, namespace, &[]).await?;
+        store.start_key = PARTITION_KEY_ROOT_KEY.to_vec();
+
+        let key_prefix = &[0];
```
afck, line 421: Should we use `EMPTY_ROOT_KEY` here?

linera-views/src/backends/rocks_db.rs
```diff
@@ -193,55 +196,41 @@ impl RocksDbStoreExecutor {
...
-                if get_upper_bound(key_prefix) == Bound::Unbounded {
-                    for short_key in self.find_keys_by_prefix_internal(key_prefix.to_vec())? {
-                        let mut full_key = self.root_key.clone();
-                        full_key.extend(key_prefix);
-                        full_key.extend(short_key);
-                        keys.push(full_key);
-                    }
-                }
-            }
-        }
```
afck, line 214: Does this change the semantics of a batch?

In general, if we have a `DeletePrefix` in a batch, and later a `Put` in that prefix, I guess we _do_ want that to be written, right? (So the old code violated that.)

linera-views/src/backends/rocks_db.rs
```diff
@@ -254,6 +243,7 @@ pub struct RocksDbStoreInternal {
     _path_with_guard: PathWithGuard,
     max_stream_queries: usize,
     spawn_mode: RocksDbSpawnMode,
+    root_key_written: Arc<Mutex<bool>>,
```
afck, line 246:
```suggestion
    root_key_written: Arc<AtomicBool>,
```

linera-views/src/backends/rocks_db.rs
```diff
@@ -402,9 +397,24 @@ impl WritableKeyValueStore for RocksDbStoreInternal {
...
+                .root_key_written
+                .lock()
+                .expect("root_key_written should not be poisoned");
+            if !*root_key_written {
+                *root_key_written = true;
+                true
+            } else {
+                false
+            }
+        };
```
afck, line 411:
```suggestion
        let write_root_key = !root_key_written.fetch_or(true);
```

linera-views/src/test_utils/mod.rs
```diff
@@ -700,3 +703,67 @@ where
             .expect("A successful deletion");
     }
 }
+
+/// Testing of the root key functionality
```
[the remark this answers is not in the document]
MathieuDutSik: I prefer the second.

linera-views/src/backends/rocks_db.rs
MathieuDutSik: That is great. But the writing is
```
let write_root_key = !self.root_key_written.fetch_or(true, Ordering::SeqCst);
```

which is still shorter than the original code.

linera-views/src/backends/rocks_db.rs
MathieuDutSik: Yes, that is simpler.

linera-views/src/backends/dynamo_db.rs
MathieuDutSik: Yes, that works.

linera-storage-service/src/client.rs
MathieuDutSik: Thanks, this lead to the elimination of `namespace_as_vec`.

linera-storage-service/src/client.rs
MathieuDutSik: Yes, thanks.

linera-storage-service/src/common.rs
MathieuDutSik: I am against putting the word "view". It is true that storage can be used for views but that is not the only usage.
I understand that the `MIN_VIEW_TAG` can give the wrong impression and so I removed it.

linera-storage-service/src/client.rs
afck: It's still a mutex in this file.

ma2bd: I was going to ask why :)

linera-views/src/backends/dynamo_db.rs
MathieuDutSik: Yes, we can.

linera-views/src/backends/rocks_db.rs
MathieuDutSik: Yes, the semantics changes. I added a test that is passed by the new code and failed by the old.

PR #2631, opened 2024-10-15
Improve system api to perform HTTP requests

jvff:
## Motivation

Linera applications can perform HTTP requests to external services, but the available `http_post` API was very limited, and prevented applications from configuring the request. This is often needed in order to add custom headers, for example for authentication.

## Proposal

Replace the `http_post` API with a broader `http_request` API, which allows sending more configurable `http::Request`s, and returns a more detailed `http::Response`.

## Test Plan

TODO

## Release Plan

- Backporting is not possible but we may want to deploy a new `devnet` and release a new
      SDK soon.
- Contains breaking changes to the WIT interface and to the `linera-sdk` API.

## Links

- [reviewer checklist](https://github.com/linera-io/linera-protocol/blob/main/CONTRIBUTING.md#reviewer-checklist)

PR #3144, opened 2025-01-17
Use service to query ethereum node in Ethereum Tracker example

jvff:
## Motivation

While working on PR #2631, the end-to-end test for the Ethereum Tracker example started failing. That happened because the updated HTTP request API also included the response headers in the oracle recording, and a "Date" header made the response non-deterministic.

## Proposal

Use the service to perform the HTTP request, and only send back to the contract the actual deterministic information that matters.

In order to do so, the `EthereumClient` had to be split in two, so that it could use either the contract system APIs or the service system APIs.

## Test Plan

The end-to-end test passes locally in my repository when this fix is applied to the changes in PR #2631.

## Release Plan

- Nothing to do / These changes follow the usual release cycle, as it only changes the example and the SDK.

## Links

### DIFF
--- a/examples/ethereum-tracker/src/contract.rs
+++ b/examples/ethereum-tracker/src/contract.rs
@@ -5,7 +5,8 @@
 
 mod state;
 
-use ethereum_tracker::{EthereumTrackerAbi, InstantiationArgument, U256Cont};
+use alloy::primitives::U256;
+use ethereum_tracker::{EthereumTrackerAbi, InstantiationArgument};
 use linera_sdk::{
     base::WithContractAbi,
     ethereum::{ContractEthereumClient, EthereumDataType, EthereumQueries as _},
@@ -46,10 +47,16 @@ impl Contract for EthereumTrackerContract {
             contract_address,
             start_block,
         } = argument;
+
         self.state.ethereum_endpoint.set(ethereum_endpoint);
         self.state.contract_address.set(contract_address);
         self.state.start_block.set(start_block);
-        self.read_initial(start_block).await;
+        self.state
+            .save()
+            .await
+            .expect("Failed to write updated storage");
+
+        self.read_initial().await;
     }
 
     async fn execute_operation(&mut self, operation: Self::Operation) -> Self::Response {
@@ -76,28 +83,36 @@ impl EthereumTrackerContract {
         (ethereum_client, contract_address)
     }
 
-    async fn read_initial(&mut self, start_block: u64) {
-        let event_name_expanded = "Initial(address,uint256)";
-        let (ethereum_client, contract_address) = self.get_endpoints();
-        let events = ethereum_client
-            .read_events(
-                &contract_address,
-                event_name_expanded,
-                start_block,
-                start_block + 1,
-            )
-            .await
-            .expect("Read the Initial event");
-        assert_eq!(events.len(), 1);
-        let event = events[0].clone();
-        let EthereumDataType::Address(address) = event.values[0].clone() else {
-            panic!("wrong type for the first entry");
+    /// Reads the initial event emitted by the Ethereum contract, with the initial account and its
+    /// balance.
+    async fn read_initial(&mut self) {
+        let request = async_graphql::Request::new("query { readInitialEvent }");
+
+        let application_id = self.runtime.application_id();
+        let response = self.runtime.query_service(application_id, request);
+
+        let async_graphql::Value::Object(data_object) = response.data else {
+            panic!("Unexpected response from `readInitialEvent`: {response:#?}");
         };
-        let EthereumDataType::Uint256(value) = event.values[1] else {
-            panic!("wrong type for the second entry");
+        let async_graphql::Value::Object(ref initial_event) = data_object["readInitialEvent"]
+        else {
+            panic!("Unexpected response data from `readInitialEvent`: {data_object:#?}");
         };
-        let value = U256Cont { value };
-        self.state.accounts.insert(&address, value).unwrap();
+        let async_graphql::Value::String(ref address) = initial_event["address"] else {
+            panic!("Unexpected address in initial event: {initial_event:#?}");
+        };
+        let async_graphql::Value::String(ref balance_string) = initial_event["balance"] else {
+            panic!("Unexpected balance in initial event: {initial_event:#?}");
+        };
+
+        let balance = balance_string
+            .parse::<U256>()
+            .expect("Balance could not be parsed");
+
+        self.state
+            .accounts
+            .insert(address, balance.into())
+            .expect("Failed to insert initial balance");
     }
 
     async fn update(&mut self, to_block: u64) {
--- a/examples/ethereum-tracker/src/contract.rs
+++ b/examples/ethereum-tracker/src/contract.rs
@@ -9,7 +9,6 @@ use alloy::primitives::U256;
 use ethereum_tracker::{EthereumTrackerAbi, InstantiationArgument};
 use linera_sdk::{
     base::WithContractAbi,
-    ethereum::ContractEthereumClient,
     views::{RootView, View},
     Contract, ContractRuntime,
 };
@@ -76,13 +75,6 @@ impl Contract for EthereumTrackerContract {
 }
 
 impl EthereumTrackerContract {
-    fn get_endpoints(&self) -> (ContractEthereumClient, String) {
-        let url = self.state.ethereum_endpoint.get().clone();
-        let contract_address = self.state.contract_address.get().clone();
-        let ethereum_client = ContractEthereumClient { url };
-        (ethereum_client, contract_address)
-    }
-
     /// Reads the initial event emitted by the Ethereum contract, with the initial account and its
     /// balance.
     async fn read_initial(&mut self) {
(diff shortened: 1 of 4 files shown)


examples/ethereum-tracker/src/contract.rs
```diff
@@ -69,69 +75,96 @@ impl Contract for EthereumTrackerContract {
...
+            r#"query {{ readTransferEvents(endBlock: {end_block}) }}"#
+        ));
+
+        let application_id = self.runtime.application_id();
+        let response = self.runtime.query_service(application_id, request);
+
+        let async_graphql::Value::Object(data_object) = response.data else {
+            panic!("Unexpected response from `readTransferEvents`: {response:#?}");
+        };
+        let async_graphql::Value::List(ref events) = data_object["readTransferEvents"] else {
```
MathieuDutSik, line 126: Isn't the "ref variable" something that we want to avoid?

examples/ethereum-tracker/src/contract.rs
```diff
@@ -69,69 +75,96 @@ impl Contract for EthereumTrackerContract {
...
-            .expect("Read a transfer event");
-        *start_block = to_block;
-        for event in events {
-            let EthereumDataType::Address(from) = event.values[0].clone() else {
-                panic!("wrong type for the first entry");
+    /// Updates the accounts based on the transfer events emitted up to the `end_block`.
+    async fn update(&mut self, end_block: u64) {
+        let request = async_graphql::Request::new(format!(
+            r#"query {{ readTransferEvents(endBlock: {end_block}) }}"#
+        ));
```
MathieuDutSik, line 118: The code in main is reading events from a `start_block` to a `to_block` (renamed `end_block` here).
In the new code, there isn't anymore a `start_code`. How do we avoid repeating the operations?

examples/ethereum-tracker/src/contract.rs
```diff
@@ -69,69 +75,96 @@ impl Contract for EthereumTrackerContract {
...
+                panic!("Unexpected destination address in transfer event: {event:#?}");
             };
-            let EthereumDataType::Uint256(value) = event.values[2] else {
-                panic!("wrong type for the third entry");
+            let async_graphql::Value::String(ref value_string) = event["value"] else {
+                panic!("Unexpected balance in transfer event: {event:#?}");
             };
+
+            let value = U256::from_str_radix(&value_string[2..], 16)
+                .expect("Balance could not be parsed as a hex string");
```
MathieuDutSik, line 146: It is a little unfortunate to have this rather low level conversion occurring in this part of the code.

examples/ethereum-tracker/src/service.rs
```diff
@@ -40,11 +46,134 @@ impl Service for EthereumTrackerService {
...
+    /// Reads the transfer events emitted by the monitored Ethereum contract.
+    async fn read_transfer_events(&self, end_block: u64) -> Vec<TransferEvent> {
+        let start_block = *self.0.state.start_block.get();
+        let events = self
+            .read_events(
+                "Transfer(address indexed,address indexed,uint256)",
+                start_block,
+                end_block,
+            )
+            .await;
```
MathieuDutSik, line 114: What I find a little odd is that now the code has two ways to access the events.
Here, we have the `read_transfer_events` and in `fn update` a different code.

examples/ethereum-tracker/src/contract.rs
```diff
@@ -46,10 +46,16 @@ impl Contract for EthereumTrackerContract {
...
         } = argument;
+
         self.state.ethereum_endpoint.set(ethereum_endpoint);
         self.state.contract_address.set(contract_address);
         self.state.start_block.set(start_block);
-        self.read_initial(start_block).await;
+        self.state
+            .save()
+            .await
+            .expect("Failed to write updated storage");
```
MathieuDutSik, line 56: What is exactly the justification for adding this `self.state.save()` here? This is the only smart contract where a `self.state.save()` is added to the `fn instantiate` function. Why is that so?
For example, the `crowd-funding` and `fungible` are doing meaningful operations in their `fn instantiate` functions. Why haven't they some `self.state.save()`.
Side question 1: When is the `fn store` function used?
Side question 2: Now, the `fn save` function is doing the writing to the batch and if the batch is empty, then nothing is done. So, could the `save`  could be added by the linera-rpc?

examples/ethereum-tracker/src/contract.rs
```diff
@@ -69,69 +75,96 @@ impl Contract for EthereumTrackerContract {
...
-            .await
-            .expect("Read the Initial event");
-        assert_eq!(events.len(), 1);
-        let event = events[0].clone();
-        let EthereumDataType::Address(address) = event.values[0].clone() else {
-            panic!("wrong type for the first entry");
+        let application_id = self.runtime.application_id();
+        let response = self.runtime.query_service(application_id, request);
+
+        let async_graphql::Value::Object(data_object) = response.data else {
```
afck, line 86: We could avoid all those `async_graphql::Value::Variant` matches by using `response.data.into_json()` instead. The JSON `Value` type has a few more convenient methods like `as_str` etc. (And we'd rarely need `as_object` because e.g. `json_data["readInitialEvent"]["address"]` works, too.)

(Anyway, maybe that's outside the scope of this PR.)

examples/ethereum-tracker/src/contract.rs
afck: I'd imagine that `FromStr` would work on the whole string here, and interpret it as hexadecimal because it starts with `0x`? I.e. maybe we can just use `value_string.parse::<U256>()`?

examples/ethereum-tracker/src/service.rs
afck: But `update` is using `read_transfer_events`. How else could we move this functionality into the service and use it from the contract?

examples/ethereum-tracker/src/contract.rs
afck: The service has access to the state, too, and `read_transfer_events` uses the `start_block` field. No need to pass it in here.

examples/ethereum-tracker/src/service.rs
```diff
@@ -40,11 +46,134 @@ impl Service for EthereumTrackerService {
...
+
+        events
+            .into_iter()
+            .map(|event| {
+                let EthereumDataType::Address(source) = event.values[0].clone() else {
```
afck, line 119: We aren't using `event` anymore after this, so we might as well use `event.values.into_iter()` instead of cloning the values.

examples/ethereum-tracker/src/state.rs
```diff
@@ -5,7 +5,7 @@ use ethereum_tracker::U256Cont;
 use linera_sdk::views::{linera_views, MapView, RegisterView, RootView, ViewStorageContext};
 
 /// The application state.
-#[derive(RootView, async_graphql::SimpleObject)]
+#[derive(RootView)]
```
afck, line 8: Alternatively, we could use `ComplexObject` and remove the manual implementations of the four getters.

linera-sdk/src/ethereum.rs
```diff
@@ -14,25 +14,27 @@ pub use linera_ethereum::{
 use linera_ethereum::{client::JsonRpcClient, common::EthereumServiceError};
 use serde::{Deserialize, Serialize};
 
-use crate::contract::wit::contract_system_api;
+use crate::{contract::wit::contract_system_api, service::wit::service_system_api};
+
+// TODO(#3143): Unify the two types into a single `EthereumClient` type.
```
afck, line 19: Does the `Contract` one still make sense at all? With the planned change for the HTTP oracle, I imagine it wouldn't work anymore?

gabu1096 (approved):

examples/ethereum-tracker/src/contract.rs
```diff
@@ -69,69 +75,97 @@ impl Contract for EthereumTrackerContract {
...
+        };
+
+        assert!(
+            balance_string.starts_with("0x"),
+            "Balance is not a hex string starting with `0x`"
+        );
+
+        let balance = balance_string
+            .parse::<U256>()
+            .expect("Balance could not be parsed as a hex string");
```
afck, line 107: Here we could probably also parse it with the `0x` still attached.

examples/ethereum-tracker/src/contract.rs
afck: I still see `from_str_radix`.

afck (approved):

examples/ethereum-tracker/src/contract.rs
deuszx: Why doesn't service access the cache (which is an overlay over persistent state + the things that contract did) ?

jvff: Because the cache is in memory inside the contract's Wasm environment, and when the service is called a separate Wasm environment is created. The way the update is sent from the contract to the service is by "persisting" data (which actually just updates the cache in the host's memory, which is only truly persisted if the block executes successfully).

jvff: No, A will only see the response B sent to it. If A calls B again, then B's changes will still be available because its runtime is reused. This is also why if A calls B and then B calls A, A doesn't have to save its state before calling B.

I don't think persisting before queries would be a good idea. The problem is that if we do that invisibly, we might cause bugs where unexpected state is persisted. It's a tricky situation :confused: I'm starting to think splitting the runtime function into two functions with more verbose names, like `save_state_and_query_service` and `query_service_without_saving_state`, would at least reduce confusion and the risk of (potentially serious) state persistence bugs :thinking:

PR #3147, opened 2025-01-20
feat(amm): Implement Swap Bids functionality

### LINKED ISSUE #969: Figure out and implement Swap Bids on AMM
Right now we only have "market orders" implemented, and we need "limit orders". We should explore combining the Matching Engine with the AMM and have one application with all functionality

crStiv:
Added SwapBid operation to allow users to place bids with minimum output amount requirements.
This ensures users get at least their specified minimum amount of tokens in return,
protecting them from unfavorable price movements.

Resolves #969


ma2bd:
Nice try but this won't compile without an actual implementation

PR #3153, opened 2025-01-20
Upload blobs for validated block certificates in separate messages.

### LINKED ISSUE #3152: Upload validated blocks' blobs separately.

afck:
## Motivation

Ultimately we want to transfer all blobs separately, rather than in a single message. (https://github.com/linera-io/linera-protocol/issues/3048)

This PR is another step towards that goal: The blobs required by a validated block certificate are now uploaded separately, rather than in the same message as the certificate itself.

## Proposal

Add a map of missing blobs for the highest-round validated block to the chain state, and a `HandlePendingBlob` endpoint to populate that map.

## Test Plan

There are already tests covering different scenarios with validated blocks' blobs. Where necessary, these were updated.

## Release Plan

## Links

- Closes #3152.

### DIFF
--- a/linera-chain/src/chain.rs
+++ b/linera-chain/src/chain.rs
@@ -13,12 +13,13 @@ use futures::stream::{self, StreamExt, TryStreamExt};
 use linera_base::{
     crypto::CryptoHash,
     data_types::{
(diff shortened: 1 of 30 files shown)


bart-linera (approved):
As far as I can tell, looks good :+1:

PR #3154, opened 2025-01-20
Reduced the number of validators in the Local Kube tests to 1.

christos-h:
## Motivation

Having 4 validators for the `kind` tests is unnecessary. Storage service tests will catch any consensus related regressions.

## Proposal

Reduce the number of validators to 1.

## Test Plan

CI will catch regressions.

## Release Plan

### DIFF
--- a/linera-service/src/cli_wrappers/local_kubernetes_net.rs
+++ b/linera-service/src/cli_wrappers/local_kubernetes_net.rs
@@ -97,7 +97,7 @@ impl SharedLocalKubernetesNetTestingConfig {
             testing_prng_seed: Some(37),
             num_other_initial_chains: 2,
             initial_amount: Amount::from_tokens(2000),
-            num_initial_validators: 4,
+            num_initial_validators: 1,
             num_shards: 4,
             binaries,
             no_build: false,


linera-service/src/cli_wrappers/local_kubernetes_net.rs
```diff
@@ -97,7 +97,7 @@ impl SharedLocalKubernetesNetTestingConfig {
             testing_prng_seed: Some(37),
             num_other_initial_chains: 2,
             initial_amount: Amount::from_tokens(2000),
-            num_initial_validators: 4,
+            num_initial_validators: 1,
```
ndr-ds, line 100: We could even do 2, I think, and it would probably still mitigate the issue (if we want to keep `kind` testing more than 1 validator still, but maybe that's not needed)

ndr-ds (approved):
Wait for CI and
[shipit](https://app.graphite.dev//settings/meme-library?org=linera-io)

linera-service/src/cli_wrappers/local_kubernetes_net.rs
ma2bd: In terms of coverage, 2 is not better than 1

PR #3159, opened 2025-01-21
Increase linera-core helmfile release timeout

ndr-ds:
## Motivation

Seems like we're timing out when trying to run `helmfile sync`

## Proposal

Increase the timeout for the `linera-core` release, which is the one timing out

## Test Plan

CI

## Release Plan

### DIFF
--- a/kubernetes/linera-validator/helmfile.yaml
+++ b/kubernetes/linera-validator/helmfile.yaml
@@ -12,7 +12,7 @@ releases:
   - name: linera-core
     namespace: default
     chart: .
-    timeout: 900
+    timeout: 1200
     needs:
       - scylla/scylla
     values:


ndr-ds:
* **#3159** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/3159?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/3159?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ndr-ds:
There's an actual issue with the deployment causing it to not get into a healthy state, so this isn't just something that is taking a bit more time. So increasing the timeout isn't an actual solution here

PR #3160, opened 2025-01-21
Use owners instead of public keys.

### LINKED ISSUE #3165: Remove `PublicKey` from `ChainOwnership`.
If we have permissionless consensus rounds, we cannot use `ChainOwnership` as a lookup table for public keys anymore. It should only care about `Owner`s.

We are also using public keys in a few other places where we should be using the `Owner` type. This is an opportunity to clean those up as well.

afck:
## Motivation

For #3162 we need to stop using `ChainOwnership` as a lookup table for public keys.

In general, we are using `PublicKey` in several places where we should be using the `Owner` type.

## Proposal

Remove public keys from `ChainOwnership`. Add the signer's public key directly to the `BlockProposal` instead.

When creating key pairs, assigning chains and changing chain ownership via CLI commands, node service mutations or system API calls, use `Owner` instead of `PublicKey`.

## Test Plan

Several tests have been updated.

## Release Plan

## Links

- Closes #3165.


linera-chain/src/manager.rs
```diff
@@ -576,31 +576,30 @@ where
 
     /// Returns the public key of the block proposal's signer, if they are a valid owner and allowed
     /// to propose a block in the proposal's round.
-    pub fn verify_owner(&self, proposal: &BlockProposal) -> Option<PublicKey> {
-        if let Some(public_key) = self.ownership.get().super_owners.get(&proposal.owner) {
-            return Some(*public_key);
+    pub fn verify_owner(&self, proposal: &BlockProposal) -> bool {
```
christos-h, line 572: Need to update method description here?

christos-h (approved):
Very nice!

ma2bd (approved):

linera-chain/src/manager.rs
EidoS021: 1 year

EidoS021:
Not a bad project

PR #3161, opened 2025-01-21
Enable the listing of `BlobIds` from the command line.

MathieuDutSik:
## Motivation

Knowing the blobs that are in storage is useful for debugging. We implement this feature here.

## Proposal

The functionality is implemented from the `db_storage`. The problem is that the implementation depends on the serialization so a test is added for this.

## Test Plan

The CI.

For the feature itself, it was tested locally from the command line for the non-fungible test.

## Release Plan

A new functionality is added, but no break of storage schema occurs.

## Links

None.


afck (approved):
Looks good to me.

But the PR title needs to be fixed: it doesn't _download_ them.

PR #3166, opened 2025-01-22
Improve the tests to the Views.

MathieuDutSik:
## Motivation

Recently, a long-standing bug was found in the `RegisterView`. 

## Proposal

Some implementing of `TestView` were done for `SetView` / `QueueView`.

The test that detects the found problem for the `RegisterView` was extended to all the view supporting `TestView`.

## Test Plan

Simply improve the tests.

## Release Plan

Nothing relevant.

## Links

None


jvff (approved):
Looks good! I left an optional suggestion, but we can also merge it like this.

Issue #3162, opened 2025-01-22
App-based permissionless consensus rounds


afck:
Add "permissionless" consensus rounds where _anyone_ can propose blocks. Expose round numbers to contracts as an oracle.

# Motivation

It is already possible to configure an application that has to be called in every block. Such an application can decide whether the block is valid, a functionality we are already using in some examples. (E.g. in the Hex game: [the block is invalid if it's not the proposer's turn in the game running on the temporary chain](https://github.com/linera-io/linera-protocol/blob/8ed5ddb84651d87c95a3c87069e3fcfd49ffc960/examples/hex-game/src/contract.rs#L115)).

This freedom of applications to decide who is allowed to propose which kind of block, and when, is currently restricted, because the proposer must always be one of the chain owners. Lifting that restriction and exposing round numbers to contracts would open up new possibilities:
* An application could maintain a set of chain owners itself, assign them to rounds based on time stamps, alternative random number generators, or VDFs.
* A chain could even be completely "permissionless" if an application used other means of preventing too many conflicting simultaneous proposals, e.g. Proof-of-Work.

PR #3168, opened 2025-01-23
Add an option for open (permissionless) multi-leader rounds.

### LINKED ISSUE #3163: Permissionless consensus rounds
Add a flag to `ChainOwnership` that completely removes the ownership check in multi-leader rounds.

afck:
## Motivation

For https://github.com/linera-io/linera-protocol/issues/3162, some chains need to be "permissionless", i.e. admit any signer as a block proposer, as far as the chain manager is concerned.

## Proposal

Add an `open_multi_leader_rounds` option. If it is `true`, the multi-leader rounds are not restricted to the chain owners anymore. (All other round types still are restricted as usual.)

## Test Plan

A test was added.

## Release Plan

## Links

- Closes #3163.

### DIFF
--- a/CLI.md
+++ b/CLI.md
@@ -200,6 +200,7 @@ Open (i.e. activate) a new multi-owner chain deriving the UID from an existing o
 
    If they are specified there must be exactly one weight for each owner. If no weights are given, every owner will have weight 100.
 * `--multi-leader-rounds <MULTI_LEADER_ROUNDS>` — The number of rounds in which every owner can propose blocks, i.e. the first round number in which only a single designated leader is allowed to propose blocks
+* `--open-multi-leader-rounds` — Whether the multi-leader rounds are unrestricted, i.e. not limited to chain owners. This should only be `true` on chains with restrictive application permissions and an application-based mechanism to select block proposers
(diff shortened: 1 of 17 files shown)


CLI.md
```diff
@@ -200,6 +200,7 @@ Open (i.e. activate) a new multi-owner chain deriving the UID from an existing o
 
    If they are specified there must be exactly one weight for each owner. If no weights are given, every owner will have weight 100.
 * `--multi-leader-rounds <MULTI_LEADER_ROUNDS>` — The number of rounds in which every owner can propose blocks, i.e. the first round number in which only a single designated leader is allowed to propose blocks
+* `--open-multi-leader-rounds` — Whether the multi-leader rounds are unrestricted, i.e. not limited to chain owners. This should only be `true` on chains with restrictive application permissions and an application-based mechanism to select block proposers
```
bart-linera, line 203: Should we add some mechanism that would disallow setting this on chains without restrictive application permissions? (Not necessarily now, but maybe a nice to have in the future?)

afck: I don't think so: Even if you _do_ set app permissions, we can't control what the app does—it might still allow anyone to propose blocks.

So I think we just need to guard against spam on the validator side, and otherwise the rule is always that your we only guarantee _safety_; bad chain owners or chain configs can brick a chain.

PR #3175, opened 2025-01-23
Validation round number oracle

### LINKED ISSUE #3162: App-based permissionless consensus rounds

### LINKED ISSUE #3164: Round-number oracle
Expose the round number where a proposal was first _validated_ as an oracle to contracts.

(Note that this is _not_ necessarily the same round where it gets finalized.)

afck:
## Motivation

For most uses of permissionless rounds (#3162), applications need to have access to the round number, at least in multi-leader rounds.

## Proposal

Expose the round number in which the block is being validated to the contract.

This is only `Some` in multi-leader rounds.

## Test Plan

An example application using this will be added in a separate PR.

## Release Plan

## Links

- Closes #3164.
- Closes #3162.

### DIFF
--- a/linera-sdk/wit/contract-system-api.wit
+++ b/linera-sdk/wit/contract-system-api.wit
@@ -31,7 +31,7 @@ interface contract-system-api {
     assert-data-blob-exists: func(hash: crypto-hash);
     log: func(message: string, level: log-level);
     consume-fuel: func(fuel: u64);
-    validation-round: func() -> round;
+    validation-round: func() -> option<u32>;
 
     record account {
         chain-id: chain-id,
@@ -126,13 +126,6 @@ interface contract-system-api {
         messages: u32,
         message-size: u32,
         storage-size-delta: u32,
-    }
-
-    variant round {
-        fast,
-        multi-leader(u32),
-        single-leader(u32),
-        validator(u32),
     }
 
     record send-message-request {
--- a/linera-sdk/wit/contract-system-api.wit
+++ b/linera-sdk/wit/contract-system-api.wit
@@ -31,6 +31,7 @@ interface contract-system-api {
     assert-data-blob-exists: func(hash: crypto-hash);
     log: func(message: string, level: log-level);
     consume-fuel: func(fuel: u64);
+    validation-round: func() -> round;
 
     record account {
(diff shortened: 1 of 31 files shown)


linera-sdk/wit/contract-system-api.wit
```diff
@@ -127,6 +128,13 @@ interface contract-system-api {
         storage-size-delta: u32,
     }
 
+    variant round {
+        fast,
```
ma2bd, line 132: If we're in a fast round, should we always fail this oracle call?

afck: We kind of do, since oracles are always disabled in fast rounds.

Du you think we should not even expose the variant to contracts? I.e. have a different `Round` type in the SDK with one less variant?

afck: I like it! That also solves the edge case with the timeout: Whenever we do have to wait for a timeout, it can't be a multi-leader round anyway.

afck: Done in https://github.com/linera-io/linera-protocol/pull/3175/commits/5361eb2a276465a9b019d2db656a7b71083eac62.

ma2bd (approved):

Issue #3170, opened 2025-01-23
Split the `locked` field in the chain manager.


afck:
In validators, this is meant to track the locked block, i.e. the latest proposal that this validator signed to confirm.

In the client, it is used for a different purpose: to track the highest validated block we have seen, so that the client knows which block it should re-propose in the next round.

This is confusing; the logic would be easier to follow if these were two separate fields, and blocks were handled on the client side the same way as on the validator side.

# Alternative

It is actually sufficient if even the validators, just like the clients, keep track of the highest validated block they have seen, and only of the _round_ of their locked block, i.e. the round of the last `ConfirmedBlock` vote they have cast.

_Except_ for the fast block: In the fast round it is possible for validators to sign different `ConfirmedBlock` votes, so each of them needs to know which one it signs.

The following would work:
* Keep track of the latest vote to confirm: Don't clear the `confirmed_vote` when setting `validated_vote`. That way we always know our locked block.
* When handling a validated block, always accept it if it is from a later round than the one we have. But only vote to confirm if it is from the _current_ round.
* When handling a block proposal: If `confirmed_vote` is `Some`, the proposal must include a validated block certificate no earlier than that vote, _or_ it must match the content of `confirmed_vote`.

PR #3178, opened 2025-01-27
Simplify chain manager logic.

afck:
## Motivation

We are using the chain manager's `locked` field for two purposes:
* In validators, this is meant to track _this validator's_ locked block, i.e. the latest proposal it signed to confirm.
* In the client, it tracks the highest validated block (or fast proposal) we have seen, i.e. the highest block that _might_ be locked for any honest validator, so that the client knows which block it should re-propose in the next round.

This is confusing, and makes it more difficult to tackle issues like https://github.com/linera-io/linera-protocol/issues/2971.

## Proposal

Use the `locked` field for the first purpose, both in clients and validators: to track the highest block that _might_ be locked for any of the validators.

In a validator's chain manager, the block that is locked for _that validator_ is the one it has last voted to confirm, i.e. the `confirmed_vote` field.

## Test Plan

The tests already cover lots of scenarios involving locked blocks, and should catch any regressions. They have been updated where appropriate.

## Release Plan

## Links

- Closes https://github.com/linera-io/linera-protocol/issues/3170.

### DIFF
--- a/linera-chain/src/block.rs
+++ b/linera-chain/src/block.rs
@@ -23,6 +23,7 @@ use crate::{
         BlockExecutionOutcome, EventRecord, ExecutedBlock, IncomingBundle, Medium, MessageBundle,
         OutgoingMessage, ProposedBlock,
     },
+    types::CertificateValue,
     ChainError,
 };
 
@@ -153,6 +154,28 @@ impl ConfirmedBlock {
                 actual: hashed_certificate_value.hash(),
             })
         }
+    }
+
+    /// Returns whether this block matches the proposal.
+    pub fn matches_proposal(&self, block: &ProposedBlock) -> bool {
+        let ProposedBlock {
+            chain_id,
+            epoch,
+            incoming_bundles,
+            operations,
+            height,
+            timestamp,
+            authenticated_signer,
+            previous_block_hash,
+        } = block;
+        *chain_id == self.chain_id()
+            && *epoch == self.epoch()
+            && *incoming_bundles == self.block().body.incoming_bundles
+            && *operations == self.block().body.operations
+            && *height == self.block().header.height
+            && *timestamp == self.block().header.timestamp
+            && *authenticated_signer == self.block().header.authenticated_signer
+            && *previous_block_hash == self.block().header.previous_block_hash
     }
 }
 
--- a/linera-chain/src/block.rs
+++ b/linera-chain/src/block.rs
@@ -157,7 +157,7 @@ impl ConfirmedBlock {
     }
 
     /// Returns whether this block matches the proposal.
-    pub fn matches_proposal(&self, block: &ProposedBlock) -> bool {
+    pub fn matches_proposed_block(&self, block: &ProposedBlock) -> bool {
         let ProposedBlock {
             chain_id,
             epoch,
--- a/linera-chain/src/manager.rs
+++ b/linera-chain/src/manager.rs
@@ -108,19 +108,20 @@ pub enum Outcome {
 
 pub type ValidatedOrConfirmedVote<'a> = Either<&'a Vote<ValidatedBlock>, &'a Vote<ConfirmedBlock>>;
 
-/// The current locked block: Validators are allowed to sign a different block (from the locked
-/// block) iff they see a `ValidatedBlockCertificate` for it with a higher round.
+/// The latest block that validators may have voted to confirm: this is either the block proposal
+/// from the fast round or a validated block certificate. Validators are allowed to vote for this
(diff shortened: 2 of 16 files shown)


linera-chain/src/block.rs
```diff
@@ -154,6 +155,28 @@ impl ConfirmedBlock {
             })
         }
     }
+
+    /// Returns whether this block matches the proposal.
+    pub fn matches_proposal(&self, block: &ProposedBlock) -> bool {
```
ma2bd, line 160: super nit: maybe `matches_proposed_block`?

linera-chain/src/manager.rs
```diff
@@ -158,8 +158,8 @@ where
     /// proposals in the same round, this contains only the first one.
     #[graphql(skip)]
     pub proposed: RegisterView<C, Option<BlockProposal>>,
-    /// Latest validated proposal that we have voted to confirm (or would have, if we are not a
-    /// validator).
+    /// Latest validated proposal that a validator may have voted to confirm. This is either the
+    /// latest `ValidatedBlock` we have seen, or the proposal from the `Fast` round.
     #[graphql(skip)]
     pub locked: RegisterView<C, Option<LockedBlock>>,
```
ma2bd, line 164: Maybe `externally_locked`, `validator_locked`, or `network_locked`?
Then, the container `LockedBlock` could be renamed accordingly.

linera-chain/src/manager.rs
```diff
@@ -283,43 +283,68 @@ where
...
             }
         }
+        // The round of our validation votes is only allowed to increase.
+        if let Some(vote) = self.validated_vote() {
+            ensure!(
+                new_round > vote.round,
+                ChainError::InsufficientRoundStrict(vote.round)
+            );
+        }
+        // A proposal that isn't newer than the locked block is not relevant anymore.
```
ma2bd, line 329: Is this restriction just making sense or is it a safety requirement?

linera-chain/src/manager.rs
```diff
@@ -283,43 +283,68 @@ where
...
+        // If we have voted to confirm we cannot vote to validate a different block anymore, except
+        // if there is a validated block certificate from a later round.
+        if let Some(vote) = self.confirmed_vote() {
+            ensure!(
+                if let Some(validated_cert) = proposal.validated_block_certificate.as_ref() {
+                    vote.round <= validated_cert.round
+                } else {
+                    vote.round.is_fast() && vote.value().inner().matches_proposal(new_block)
+                },
+                ChainError::HasLockedBlock(new_block.height, vote.round)
```
ma2bd, line 345: Maybe the error should be renamed to `HasIncompatibleConfirmedVote`?

ma2bd (approved):
Couple of questions and nits but otherwise LGTM

linera-chain/src/manager.rs
afck: It's not a safety requirement anymore, since for safety we now check `confirmed_vote` directly.

linera-chain/src/manager.rs
```diff
@@ -283,43 +283,68 @@ where
...
         );
-        // TODO(#2971): The client still needs to make a note of proposals that fail here:
-        match self.locked.get().as_ref() {
-            None => {} // No locked block; any proposal is allowed.
-            Some(LockedBlock::Fast(old_proposal)) => {
-                // We have a locked block in the `Fast` round. Either the proposal contains a
-                // validated block certificate, or it must propose the same block.
+        let current_round = self.current_round();
+        match new_round {
+            Round::Fast => {}
```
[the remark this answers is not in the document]
afck: We can't return early in that case here, because the `Fast` proposal may still be relevant as a locked block, even if the current round is already a later one.

linera-chain/src/manager.rs
```diff
@@ -283,43 +283,68 @@ where
...
+            ensure!(
+                new_round > vote.round,
+                ChainError::InsufficientRoundStrict(vote.round)
+            );
+        }
+        // A proposal that isn't newer than the locked block is not relevant anymore.
+        if let Some(locked) = self.locked.get() {
+            ensure!(
+                locked.round() < new_round,
+                ChainError::HasLockedBlock(new_block.height, locked.round())
```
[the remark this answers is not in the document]
afck: I went with `HasIncompatibleConfirmedVote`.

linera-chain/src/manager.rs
afck: How would you feel about `UnlockingBlock`?

The point of it is that it can _unlock_ validators that have already voted to confirm.

Issue #3003, opened 2024-12-03
A Request For Quotes application


bart-linera:
Design and implement an application enabling users to exchange different kinds of tokens without locking up funds on a chain which might stop being live, thus causing the user to lose access to their funds. That is, the application should always use a separate chain owned by both parties involved in the exchange to perform said exchange.

PR #3179, opened 2025-01-27
Add a Request For Quotes example application

bart-linera:
## Motivation

Current example solutions for atomic swaps require lots of manual setup and can only be used once per application instance. We could use an application that automates some of the rough edges and enables performing multiple atomic swaps.

## Proposal

The RFQ application works by having users request quotes from other users, and automatically setting up an atomic swap on a temporary chain when both sides agree upon a quote.

## Test Plan

Manual testing by running the scenario described in the README has been performed.

## Release Plan

## Links

- Issue #3003 

### DIFF
--- a/examples/rfq/README.md
+++ b/examples/rfq/README.md
@@ -1,0 +1,327 @@
+<!-- cargo-rdme start -->
+
+# Automated Market Maker (AMM) Example Application
+
+This example implements an Automated Market Maker (AMM) which demonstrates DeFi capabilities of the
+Linera protocol. Prerequisite for the AMM application is the `fungible` application, as we will
+be adding/removing liquidity and also performing a swap.
+
+# How it works
+
+It supports the following operations. All operations need to be executed remotely.
+
+- Swap: For a given input token and an input amount, it swaps that token amount for an
+  amount of the other token calculated based on the current AMM ratio.
+
+- Add Liquidity: This operation allows adding liquidity to the AMM. Given a maximum
+  `token0` and `token1` amount that you're willing to add, it adds liquidity such that you'll be
+  adding at most `max_token0_amount` of `token0` and `max_token1_amount` of `token1`. The amounts
+  will be calculated based on the current AMM ratio. The owner, in this case, refers to the user
+  adding liquidity, which currently can only be a chain owner.
+
+- Remove Liquidity: This withdraws tokens from the AMM. Given the index of the token you'd
+  like to remove (can be 0 or 1), and an amount of that token that you'd like to remove, it calculates
+  how much of the other token will also be removed based on the current AMM ratio. Then it removes
+  the amounts from both tokens as a removal of liquidity. The owner, in this context, is the user
+  removing liquidity, which currently can only be a chain owner.
+
+# Usage
+
+## Setting Up
+
+Before getting started, make sure that the binary tools `linera*` corresponding to
+your version of `linera-sdk` are in your PATH. For scripting purposes, we also assume
+that the BASH function `linera_spawn_and_read_wallet_variables` is defined.
+
+From the root of Linera repository, this can be achieved as follows:
+
+```bash
+export PATH="$PWD/target/debug:$PATH"
+source /dev/stdin <<<"$(linera net helper 2>/dev/null)"
+```
+
+To start the local Linera network:
+
+```bash
+linera_spawn_and_read_wallet_variables linera net up --testing-prng-seed 37
+```
+
+We use the test-only CLI option `--testing-prng-seed` to make keys deterministic and simplify our
+explanation.
+
+```bash
+OWNER_1=d2115775b5b3c5c1ed3c1516319a7e850c75d0786a74b39f5250cf9decc88124
+OWNER_2=a477cb966190661c0dfbe50602616a78a48d2bef6cb5288d49deb3e05585d579
+CHAIN_1=673ce04da4b8ed773ee7cd5828a2083775bea4130498b847c5b34b2ed913b07f
+CHAIN_2=69705f85ac4c9fef6c02b4d83426aaaf05154c645ec1c61665f8e450f0468bc0
+CHAIN_AMM=e476187f6ddfeb9d588c7b45d3df334d5501d6499b3f9ad5595cae86cce16a65
+OWNER_AMM=7136460f0c87ae46f966f898d494c4b40c4ae8c527f4d1c0b1fa0f7cff91d20f
+```
+
+Now we have to publish and create the fungible applications. The flag `--wait-for-outgoing-messages` waits until a quorum of validators has confirmed that all sent cross-chain messages have been delivered.
+
+```bash
+(cd examples/fungible && cargo build --release --target wasm32-unknown-unknown)
+
+FUN1_APP_ID=$(linera --wait-for-outgoing-messages \
+  publish-and-create examples/target/wasm32-unknown-unknown/release/fungible_{contract,service}.wasm \
+    --json-argument "{ \"accounts\": {
+        \"User:$OWNER_AMM\": \"100.\"
+    } }" \
+    --json-parameters "{ \"ticker_symbol\": \"FUN1\" }" \
+)
+
+FUN2_APP_ID=$(linera --wait-for-outgoing-messages \
+  publish-and-create examples/target/wasm32-unknown-unknown/release/fungible_{contract,service}.wasm \
+    --json-argument "{ \"accounts\": {
+        \"User:$OWNER_AMM\": \"100.\"
+    } }" \
+    --json-parameters "{ \"ticker_symbol\": \"FUN2\" }" \
+)
+
+(cd examples/amm && cargo build --release --target wasm32-unknown-unknown)
+AMM_APPLICATION_ID=$(linera --wait-for-outgoing-messages \
+  publish-and-create examples/target/wasm32-unknown-unknown/release/amm_{contract,service}.wasm \
+  --json-parameters "{\"tokens\":["\"$FUN1_APP_ID\"","\"$FUN2_APP_ID\""]}" \
+  --required-application-ids $FUN1_APP_ID $FUN2_APP_ID)
+```
+
+## Using the AMM Application
+
+First, a node service for the current wallet has to be started:
+
+```bash
+PORT=8080
+linera service --port $PORT &
+```
+
+### Using GraphiQL
+
+Type each of these in the GraphiQL interface and substitute the env variables with their actual
+values that we've defined above.
+
+To properly setup the tokens in the proper chains, we need to do some transfer operations:
+
+- Transfer 50 FUN1 from `$OWNER_AMM` in `$CHAIN_AMM` to `$OWNER_1` in `$CHAIN_1`, so they're in the proper chain.
+  Run `echo "http://localhost:8080/chains/$CHAIN_AMM/applications/$FUN1_APP_ID"` to print the URL
+  of the GraphiQL interface for the FUN1 app. Navigate to that URL and enter:
+
+```gql,uri=http://localhost:8080/chains/$CHAIN_AMM/applications/$FUN1_APP_ID
+    mutation {
+        transfer(
+            owner: "User:$OWNER_AMM",
+            amount: "50.",
+            targetAccount: {
+                chainId: "$CHAIN_1",
+                owner: "User:$OWNER_1",
+            }
+        )
+    }
+```
+
+- Transfer 50 FUN1 from `$OWNER_AMM` in `$CHAIN_AMM` to `$OWNER_2` in `$CHAIN_2`, so they're in the proper chain:
+
+```gql,uri=http://localhost:8080/chains/$CHAIN_AMM/applications/$FUN1_APP_ID
+    mutation {
+        transfer(
+            owner: "User:$OWNER_AMM",
+            amount: "50.",
+            targetAccount: {
+                chainId: "$CHAIN_2",
+                owner: "User:$OWNER_2",
+            }
+        )
+    }
+```
+
+- Transfer 50 FUN2 from `$OWNER_AMM` in `$CHAIN_AMM` to `$OWNER_1` in `$CHAIN_1`, so they're in the proper chain.
+  Since this is the other token, FUN2, we need to go to its own GraphiQL interface:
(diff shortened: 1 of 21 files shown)


linera-execution/src/runtime.rs
```diff
@@ -1201,6 +1201,7 @@ impl ContractSyncRuntimeHandle {
             &mut contract
                 .instance
                 .try_lock()
+                .map_err(|_| println!("{:?}", self.inner().call_stack))
```
afck, line 1204: Is this a leftover from debugging?

linera-base/src/ownership.rs
```diff
@@ -187,6 +187,14 @@ pub enum CloseChainError {
     NotPermitted,
 }
 
+/// Errors that can happen when attempting to change the application permissions.
+#[derive(Clone, Copy, Debug, Error, WitStore, WitType)]
+pub enum ChangeApplicationPermissionsError {
+    /// Authenticated signer wasn't allowed to change the application permissions.
```
afck, line 193: Is it the authenticated signer or the application that doesn't have permission? (Or does it depend on the context?)

examples/rfq/README.md
```diff
@@ -0,0 +1,278 @@
...
+
+Each user is supposed to run the application on their own chain. Once user A wants to exchange
+tokens with another user B, they need to look up user B's chain ID and submit a `RequestQuote`
+operation. Such an operation defines the pair of tokens that are intended to be exchanged by their
+application IDs, as well as the amount to be exchanged.
+
+Once user B receives the request, they can respond with a quote using the `ProvideQuote` operation.
+In it, they specify the price they are willing to offer, as well as their owner ID (which will be
+required for setting up the temporary chain for the atomic swap). It is possible that multiple
+requests could have been received: the user specify which one they are responding to using a request
```
afck, line 17: (I think it should be "the user specifies"… even though it is "they specify", of course. :thinking: E.g. [Merriam-Webster writes](https://www.merriam-webster.com/wordplay/singular-nonbinary-they) "This is my friend, Jay. I met them at work." Not "This are my friend, Jay.")

examples/rfq/README.md
```diff
@@ -0,0 +1,278 @@
...
+`CancelRequest` operation, or accept it using the `AcceptQuote` operation. Cancelling the request
+removes it from the application state and notifies the other party. Accepting the request launches
+the exchange process.
+
+Initially, a temporary chain for the atomic swap is automatically created. An instance of the
+Matching Engine application is created on the chain, tokens are transferred to the chain (they are
+initially assumed to exist on user A's chain - if they don't, the operation will fail), a Matching
+Engine order is submitted and user B is notified.
+
+The temporary chain is owned by both users, which mean both users can withdraw from the exchange at
```
afck, line 30:
```suggestion
The temporary chain is owned by both users, which means both users can withdraw from the exchange at
```

examples/rfq/README.md
```diff
@@ -0,0 +1,278 @@
...
+
+# Usage
+
+## Setting Up
+
+Before getting started, make sure that the binary tools `linera*` corresponding to
+your version of `linera-sdk` are in your PATH. For scripting purposes, we also assume
+that the BASH function `linera_spawn_and_read_wallet_variables` is defined.
+
+From the root of Linera repository, this can be achieved as follows:
```
afck, line 53:
```suggestion
From the root of the Linera repository, this can be achieved as follows:
```

examples/rfq/README.md
```diff
@@ -0,0 +1,278 @@
...
+We also need to make sure that both users can access the RFQ application, so we have to request it
+on the second user's chain.
+
+```bash
+linera -w 1 request-application $APP_RFQ
+```
+
+## Using the RFQ Application
+
+First, node services for the both users' wallets have to be started:
```
afck, line 123:
```suggestion
First, node services for both users' wallets have to be started:
```

examples/rfq/README.md
```diff
@@ -0,0 +1,278 @@
...
+linera -w 0 service --port 8080 &
+linera -w 1 service --port 8081 &
+```
+
+### Using GraphiQL
+
+Type each of these in the GraphiQL interface and substitute the env variables with their actual
+values that we've defined above.
+
+First, user B has to claim their tokens on their chain, since they were created on user A's chain.
```
afck, line 135: Would it simplify things if B created the FUN0 app?

It's probably also sufficient if A only has FUN1 and B only has FUN2. (Not sure if that feels less natural as an example.)

examples/rfq/README.md
```diff
@@ -0,0 +1,278 @@
...
+    amount: "500.",
+    targetAccount: {
+      chainId: "$CHAIN_1",
+      owner: "User:$OWNER_1"
+    }
+  )
+}
+```
+
+- Claim 500 FUN2 from `$OWNER_1` in `$CHAIN_0` to `$OWNER_1` in `$CHAIN_1`, so they're in the proper chain.
```
afck, line 157: I'd remove the dashes: Paragraphs separated by a new line won't be indented anymore, even though I think they conceptually belong to this point?

examples/rfq/README.md
```diff
@@ -0,0 +1,278 @@
...
+```gql,uri=http://localhost:8080/chains/$CHAIN_0/applications/$APP_RFQ
+mutation {
+  closeRequest(
+    requestId: {
+      otherChainId:"$CHAIN_1",
+      seqNum:0
+    },
+  )
+}
+```
```
afck, line 259: Only one of them has to make that request, I assume?

examples/rfq/README.md
```diff
@@ -0,0 +1,278 @@
...
+`http://localhost:8081/chains/$CHAIN_1/applications/$APP_ID_0` (user B's FUN1 account) and
+`http://localhost:8081/chains/$CHAIN_1/applications/$APP_ID_1` (user B's FUN2 account) and
+performing the following query:
+
+```gql
+query {
+    accounts {
+        entries { key value }
+    }
+}
```
afck, line 274: We've been using 2 spaces to indent GraphQL elsewhere.

linera-rpc/tests/snapshots/format__format.yaml.snap
```diff
@@ -1,5 +1,6 @@
 ---
 source: linera-rpc/tests/format.rs
+assertion_line: 68
```
afck, line 3: Weird! Is that due to a different `insta` version?

examples/rfq/README.md
bart-linera: You're right, of course - I must have made a typo :sweat_smile:

linera-base/src/ownership.rs
bart-linera: Hmm, good point. I might have copied this comment from the `CloseChain` error. But if not, then it's probably just wrong, but I'll double check.

linera-execution/src/runtime.rs
bart-linera: Looks like it :grimacing:

examples/rfq/README.md
bart-linera: Both, actually - because this request also transfers the tokens, and the transfer requires the operation to be signed by the owner of the tokens. It's a bit suboptimal UX, but I didn't have any ideas how to fix that.

linera-rpc/tests/snapshots/format__format.yaml.snap
bart-linera: No idea! I can try removing it and seeing if it doesn't cause issues.

examples/rfq/README.md
bart-linera: Huh. I changed that now, and I see just two spaces on my side, but it still displays as 4 spaces here :thinking:

examples/rfq/README.md
afck: Can't we make it so that the Matching Engine owns the tokens? Its `CloseChain` operation seems to be able to send back all tokens to their owners in one go:
https://github.com/linera-io/linera-protocol/blob/a918bc92c9f458d918063ff2389da8780c973bc7/examples/matching-engine/src/contract.rs#L94

bart-linera: I thought about it, but there is an issue. After the orders are matched, ME immediately returns the tokens, and it returns them to the temporary chain, because that's where it got them from. So they are not under ME control anymore.

But now that I think about it, maybe it's RFQ that should be owning the tokens :thinking: Then ME would be returning them to RFQ, and RFQ would return them to the users.

afck: Right, and that could be done synchronously.

Actually, this implementation feels like it doesn't really need the Matching Engine at all: It already knows all the parameters, so it could just swap the tokens directly.

bart-linera: Hmm, that is true. I'm not really using much of what the Matching Engine does in the end :grimacing:

parfenukajnur:
🤔

Andrey11012:
Good proekt

linera-rpc/tests/snapshots/format__format.yaml.snap
bart-linera: Looks like it wasn't needed :+1:

linera-base/src/ownership.rs
bart-linera: It was, in fact, the application. The same applied to CloseChainError, so I changed the comment there as well.

examples/rfq/README.md
bart-linera: I removed the Matching Engine entirely. It simplified some logic and improved the UX significantly.

examples/rfq/README.md
afck: Looks correct now.

examples/rfq/src/state.rs
```diff
@@ -0,0 +1,268 @@
...
+            | RequestState::QuoteProvided(QuoteProvided { token_pair, .. }) => token_pair.clone(),
+            _ => panic!("invalid state for reading the token pair"),
+        }
+    }
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize, SimpleObject)]
+pub struct RequestData {
+    state: RequestState,
+    pub we_requested: bool,
```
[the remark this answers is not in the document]
bart-linera: Unfortunately no, most states are valid on both sides of the exchange. I could theoretically make different states for the requesting side and for the providing side, but the distinction is not always important and would lead to duplication, so I thought this way made more sense.

examples/rfq/src/lib.rs
```diff
@@ -0,0 +1,141 @@
...
+impl ServiceAbi for RfqAbi {
+    type Query = Request;
+    type QueryResponse = Response;
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, SimpleObject, InputObject)]
+#[graphql(input_name = "TokenPairInput")]
+pub struct TokenPair {
+    pub token_offered: ApplicationId,
+    pub token_asked: ApplicationId,
```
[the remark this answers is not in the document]
bart-linera: It would, but it doesn't work for some reason - the derive macros start throwing fits.

examples/rfq/src/contract.rs
```diff
@@ -0,0 +1,413 @@
...
+                quote,
+                quoter_owner,
+            } => {
+                let request_data = self
+                    .state
+                    .request_data(&request_id)
+                    .await
+                    .expect("Request not found!");
+                if request_data.we_requested {
+                    return;
```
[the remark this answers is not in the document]
bart-linera: Right, I'm not sure which is better in the context of an application. As far as I can tell, just returning or panicking are the only two options (there is no way to return an error), right? But maybe always panicking on an error would be more consistent.

examples/rfq/src/contract.rs
```diff
@@ -0,0 +1,413 @@
...
+            Operation::RequestQuote {
+                target,
+                token_pair,
+                amount,
+            } => {
+                let seq_number = self
+                    .state
+                    .create_new_request(target, token_pair.clone(), amount);
+                let message = Message::RequestQuote {
+                    seq_number,
```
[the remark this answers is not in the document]
bart-linera: Hmm, good point, I think you're right.
The decision to use the other party's chain ID was deliberate, because I thought that would identify the request uniquely, and that's the only part they really need to store - each one knows their own chain ID.
I think I would opt for making `we_requested` a part of the ID :+1:

examples/rfq/src/lib.rs
```diff
@@ -124,17 +138,24 @@ pub enum Message {
...
+            | Message::ChainClosed { request_id } => request_id.clone(),
+            Message::CancelRequest {
+                seq_number,
+                recipient_requested,
+            } => RequestId::new(other_chain_id, *seq_number, *recipient_requested),
             Message::TokensSent { .. } | Message::CloseChain => {
-                // not important, we can just return 0
-                0
+                // unused
+                RequestId::new(other_chain_id, 0, false)
```
afck, line 158: (Then we could even say `unreachable!()` here? Your call.)

afck (approved):

examples/rfq/src/lib.rs
bart-linera: I thought about it, but it would complicate `handle_message` in `contract.rs` - as it is now, I can just have a single call to `message.request_id()` on top of the `handle_message` function, but I'd have to have one call per message variant if I put `unreachable!` here.

afck (changes requested):
The README should be added to `readme_test.rs`.

afck (approved):

linera-service/tests/readme_test.rs
```diff
@@ -30,6 +30,7 @@ use tokio::process::Command;
 #[test_case::test_case("../examples/non-fungible" ; "non-fungible")]
 #[test_case::test_case("../examples/matching-engine" ; "matching engine")]
 #[test_case::test_case("../examples/meta-counter" ; "meta counter")]
+#[test_case::test_case("../examples/rfq" ; "requests for quotes")]
```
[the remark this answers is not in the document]
bart-linera: Should it be singular? I was thinking, it's an application for making potentially multiple requests, hence I chose to use the plural - but maybe the naming convention here is different?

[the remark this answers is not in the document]
afck: Good point!

PR #3183, opened 2025-01-28
Fix misleading error message

intls:
## Proposal

Corrected error message in the Message::Swap handler so that it reflects the swap operation rather than adding liquidity.

## Release Plan

    - be released in a new SDK,
    - be released in a new SDK,
      SDK soon.

## Links

### DIFF
--- a/examples/amm/src/contract.rs
+++ b/examples/amm/src/contract.rs
@@ -32,8 +32,6 @@ impl Contract for AmmContract {
     type Message = Message;
     type InstantiationArgument = ();
     type Parameters = Parameters;
-    type Operation = Operation;
-    type Response = ();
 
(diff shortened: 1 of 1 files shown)


examples/amm/src/contract.rs
```diff
@@ -32,6 +32,8 @@ impl Contract for AmmContract {
     type Message = Message;
     type InstantiationArgument = ();
     type Parameters = Parameters;
+    type Operation = Operation;
+    type Response = ();
```
ma2bd, line 36: `WithContractAbi` above is taking care of this

intls: Got it! will revert

intls:
@ma2bd Thanks! I've reverted!

afck (approved):

PR #3186, opened 2025-01-28
Fix Placeholder in NFT Error Messages

donatik27:
### Changes Made

1. **File:** `examples/gen-nft/src/contract.rs`
   - **Old Code:** `.expect("NFT {token_id} not found")`
   - **New Code:** `.expect("NFT not found")`

2. **File:** `examples/non-fungible/src/contract.rs`
   - **Old Code:** `.expect("NFT {token_id} not found")`
   - **New Code:** `.expect("NFT not found")`

### Reason for Changes

Improved clarity and consistency of error messages. The previous message included a placeholder `{token_id}` that was not replaced, leading to confusion. The new message "NFT not found" is clear and concise, aiding in debugging.

### DIFF
--- a/examples/crowd-funding/README.md
+++ b/examples/crowd-funding/README.md
@@ -29,7 +29,7 @@ Optionally, contributors may also be able to create a block to accept a new epoc
 (i.e. a change of validators).
 
 <!--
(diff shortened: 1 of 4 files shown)


afck (approved):
Good catch!

afck (approved):

Issue #3203, opened 2025-01-29
Limit total pending blob size.


afck:
In permissionless chains, anyone can upload proposals and supply pending blobs.

Even in permissioned chains, an owner can upload a proposal with lots of associated maximum-size blobs.

We should limit the number of `PendingBlobsView`s as well as the total size of pending blobs in each individual view.

malboro3435:
super, go go go

Utso123:
??

tufinger:
.

luchkoolga21:
+1

ex001xp:
hmm Limit number of pending

Cave11:
Привет!

AngryBuka:
Which terminals does it work best with?

PR #3204, opened 2025-01-29
Upload proposals' blobs separately.

### LINKED ISSUE #3202: Upload proposals' blobs separately.

afck:
## Motivation

Including blobs with the gRPC message that contains a block proposal or certificate severely limits the total size of the blobs. (See https://github.com/linera-io/linera-protocol/issues/3048.)

## Proposal

Remove the `blobs` field from block proposals, and handle blobs separately, with one message each.

I added a `maximum_published_blobs` limit per block, and limited the number of pending proposals with blobs to 1 for permissionless chains. (See https://github.com/linera-io/linera-protocol/issues/3203.)

## Test Plan

The tests have been updated where necessary. Otherwise they already cover different scenarios involving proposals with blobs.

## Release Plan

## Links

- Closes #3202

### DIFF
--- a/linera-chain/src/pending_blobs.rs
+++ b/linera-chain/src/pending_blobs.rs
@@ -26,6 +26,12 @@ where
 {
     /// The round in which the block is validated.
     pub round: RegisterView<C, Round>,
+    /// Whether these blobs were already validated.
+    ///
+    /// This is only `false` for _new_ block proposals, not when re-proposing blocks from earlier
+    /// rounds or when handling validated block certificates. If it is false, the pending blobs are
+    /// only the ones published by the new block, not the ones that are only read.
+    pub validated: RegisterView<C, bool>,
     /// The map of blobs needed to process the block.
     pub pending_blobs: MapView<C, BlobId, Option<Blob>>,
 }
@@ -51,6 +57,7 @@ where
     pub async fn update(
         &mut self,
         round: Round,
+        validated: bool,
         maybe_blobs: BTreeMap<BlobId, Option<Blob>>,
     ) -> Result<(), ChainError> {
         let existing_round = *self.round.get();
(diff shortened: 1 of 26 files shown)


linera-chain/src/pending_blobs.rs
```diff
@@ -0,0 +1,78 @@
...
+
+    pub async fn maybe_insert(&mut self, blob: &Blob) -> Result<(), ViewError> {
+        let blob_id = blob.id();
+        if let Some(maybe_blob) = self.pending_blobs.get_mut(&blob_id).await? {
+            if maybe_blob.is_none() {
+                *maybe_blob = Some(blob.clone());
+            }
+        }
+        Ok(())
+    }
```
bart-linera, line 55: Would it be useful to signal somehow if the blob was _not_ inserted?

afck: I guess so: We could then return to the client whether the blob was inserted in any of the collections. I'll do that in a follow-up PR.

afck: Actually, not sure it would be useful at all:

`HandlePendingBlob` currently returns a `ChainInfo`. I'm not sure it would be more useful to replace this with a `bool`, or worth it to add another response struct that contains both a `bool` and a `ChainInfo`.

Because on the client side, there isn't necessarily an expectation that we get a `true` response: E.g. other clients might currently be sending the same validated block certificate and also upload the blobs, so we would get back `false` because now the validator already had it.

Anyway, happy to discuss this next week!

bart-linera: I was thinking more along the lines of: if we try to insert an unexpected blob, maybe we'd like to know? Even if just to log it on the validator side, because that indicates either malicious behavior or a bug. It shouldn't happen, but if it does happen, it might be better to know than just silently ignore it.

PR #3205, opened 2025-01-29
Added more GraphQL docs to assist LLM agents.

christos-h:
## Motivation

Scalars in our codebase are not documented implicitly by GraphQL introspection.

## Proposal

Add explicit documentation of the structure of some of the scalars we have.

### DIFF
--- a/linera-base/src/data_types.rs
+++ b/linera-base/src/data_types.rs
@@ -1206,7 +1206,17 @@ impl Event {
 impl<'de> BcsHashable<'de> for Event {}
 
 doc_scalar!(Bytecode, "A WebAssembly module's bytecode");
-doc_scalar!(Amount, "A non-negative amount of tokens.");
+doc_scalar!(
+    Amount,
+    r#"
+A non-negative amount of tokens.
+
+An example of an `Amount`:
+```
+"1.234"
+```
+"#
+);
 doc_scalar!(BlockHeight, "A block height to identify blocks in a chain");
 doc_scalar!(
     Timestamp,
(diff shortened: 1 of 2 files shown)


linera-execution/src/graphql.rs
```diff
@@ -21,7 +21,18 @@ doc_scalar!(
     Epoch,
     "A number identifying the configuration of the chain (aka the committee)"
 );
-doc_scalar!(Recipient, "The recipient of a transfer");
+doc_scalar!(Recipient, r#"
+The recipient of a transfer.
+
+An example of a `Recipient`:
+```graphql
```
ma2bd, line 28: Technically this is json. Perhaps the comment could mention GraphQL?

ma2bd, line 28: Technically I believe this is json. Perhaps the comment could mention GraphQL?

linera-base/src/data_types.rs
```diff
@@ -1134,7 +1134,14 @@ impl<'a> Deserialize<'a> for Blob {
 }
 
 doc_scalar!(Bytecode, "A WebAssembly module's bytecode");
-doc_scalar!(Amount, "A non-negative amount of tokens.");
+doc_scalar!(Amount, r#"
+A non-negative amount of tokens.
+
+An example of an `Amount`:
+```
```
ma2bd, line 1141: json?

ma2bd (approved):
Great! Just need to fix CI

linera-execution/src/graphql.rs
christos-h: I don't think this is JSON. JSON always has string keys.

E.g.:

```
{
    "Account" : {
        "chain_id": <chain_id>
    }
}
```

ma2bd: gotcha

PR #3209, opened 2025-01-30
Improve Variable Naming for Clarity in next_timestamp Function

Hopium21:
The variable name `block_time` was changed to `current_time` to enhance code clarity. The new name better reflects the purpose of the variable, indicating that it holds the current time rather than the time associated with a block. This change improves the readability and maintainability of the code.


afck:
There's a difference between the block timestamp and the time at which the block was validated (see e.g. https://github.com/linera-io/linera-protocol/blob/c74ad68e4d545d74c98a4ba958e3fd0b6cf9008c/examples/hex-game/src/contract.rs#L116), so I'd say `current_time` is _less_ precise.

PR #3217, opened 2025-01-31
Change mutations to schedule operations using a system API

jvff:
## Motivation

Previously, services were required to use GraphQL in order to support creating blocks with operations. The design was that each mutation query would return a serialized operation as a vector of bytes. This is a constraint that's not really needed, and requires extra work by application developers.

## Proposal

Add a `schedule_operation` system API, which adds an operation to a list of scheduled operations. More than one operation can be scheduled to be added to the proposed block. The list is then returned back to the client (in the node service), which then uses it to build a block proposal and send it to the validators.

The SDK was updated to add `ServiceRuntime::schedule_raw_operation` to send the operation as a vector of bytes, and also add a `ServiceRuntime::schedule_operation` method which accepts any `Serialize` type and handles serialization for the application.

All examples were updated to use the new API, either manually if the mutation object is implemented manually or by the changes to the `GraphQLMutationRoot` derive macro. Unfortunately, `async_graphql` requires the methods to return _something_ that implements `OutputType`, so for now they return an empty array of bytes.

## Test Plan

CI should catch any regressions by the switch to use the new API.

## Release Plan

      SDK soon.

In theory only the client is affected, but since the API is available to services, when running a contract the validators could query a service that runs as an oracle and the behavior would differ if this change is included in the validator or not. In both cases an error will be returned and the block rejected, but the errors would be different. To mitigate risk, this should be released together with validator upgrades.

## Links

### DIFF
--- a/examples/counter/src/service.rs
+++ b/examples/counter/src/service.rs
@@ -57,9 +57,9 @@ struct MutationRoot {
 
 #[Object]
 impl MutationRoot {
-    async fn increment(&self, value: u64) -> Vec<u8> {
+    async fn increment(&self, value: u64) -> [u8; 0] {
         self.runtime.schedule_operation(&value);
-        bcs::to_bytes(&value).unwrap()
+        []
     }
 }
 
--- a/examples/counter/src/service.rs
+++ b/examples/counter/src/service.rs
@@ -5,6 +5,8 @@
 
 mod state;
 
+use std::sync::Arc;
+
 use async_graphql::{EmptySubscription, Object, Request, Response, Schema};
 use linera_sdk::{base::WithServiceAbi, views::View, Service, ServiceRuntime};
 
@@ -12,6 +14,7 @@ use self::state::CounterState;
 
 pub struct CounterService {
     state: CounterState,
+    runtime: Arc<ServiceRuntime<Self>>,
 }
 
 linera_sdk::service!(CounterService);
@@ -27,7 +30,10 @@ impl Service for CounterService {
         let state = CounterState::load(runtime.root_view_storage_context())
             .await
             .expect("Failed to load state");
-        CounterService { state }
+        CounterService {
+            state,
+            runtime: Arc::new(runtime),
+        }
     }
 
     async fn handle_query(&self, request: Request) -> Response {
@@ -35,7 +41,9 @@ impl Service for CounterService {
             QueryRoot {
                 value: *self.state.value.get(),
             },
-            MutationRoot {},
+            MutationRoot {
+                runtime: self.runtime.clone(),
+            },
             EmptySubscription,
         )
         .finish();
@@ -43,11 +51,14 @@ impl Service for CounterService {
     }
 }
 
-struct MutationRoot;
+struct MutationRoot {
+    runtime: Arc<ServiceRuntime<CounterService>>,
+}
 
 #[Object]
 impl MutationRoot {
     async fn increment(&self, value: u64) -> Vec<u8> {
+        self.runtime.schedule_operation(&value);
         bcs::to_bytes(&value).unwrap()
     }
 }
--- a/examples/counter/src/service.rs
+++ b/examples/counter/src/service.rs
@@ -65,6 +65,8 @@ impl QueryRoot {
 
 #[cfg(test)]
 mod tests {
+    use std::sync::Arc;
+
     use async_graphql::{Request, Response, Value};
     use futures::FutureExt as _;
     use linera_sdk::{util::BlockingWait, views::View, Service, ServiceRuntime};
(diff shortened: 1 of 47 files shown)


examples/counter/src/service.rs
```diff
@@ -27,28 +30,36 @@ impl Service for CounterService {
...
+    runtime: Arc<ServiceRuntime<CounterService>>,
+}
 
 #[Object]
 impl MutationRoot {
-    async fn increment(&self, value: u64) -> Vec<u8> {
-        bcs::to_bytes(&value).unwrap()
+    async fn increment(&self, value: u64) -> [u8; 0] {
+        self.runtime.schedule_operation(&value);
+        []
```
ma2bd, line 62: In principle, we could return something here, for instance the future value.

ma2bd (approved):
nice

linera-execution/src/runtime.rs
```diff
@@ -959,21 +963,27 @@ impl<UserInstance> BaseRuntime for SyncRuntimeInternal<UserInstance> {
...
+        } else {
+            let context = QueryContext {
+                chain_id: self.chain_id,
+                next_block_height: self.height,
+                local_time: self.local_time,
             };
+            let sender = self.execution_state_sender.clone();
+            ServiceSyncRuntime::new(sender, context).run_query(application_id, query)?
+        };
+        self.scheduled_operations.extend(operations);
```
afck, line 986: This is probably only needed in the `else` case above. Or actually, should we check whether we're currently in a service query, and never add to `scheduled_operations` if we're in an operation?

linera-sdk-derive/src/lib.rs
```diff
@@ -62,16 +67,25 @@ fn generate_mutation_root_code(input: ItemEnum, crate_root: &str) -> TokenStream
                     field_names.push(name);
                 }
                 methods.push(quote! {
-                    async fn #function_name(&self, #(#fields,)*) -> Vec<u8> {
-                        #crate_root::bcs::to_bytes(&#enum_name::#variant_name ( #(#field_names,)* ))
-                            .unwrap()
+                    async fn #function_name(&self, #(#fields,)*) -> [u8; 0] {
```
afck, line 70: So GraphQL requires a return type other than `()`, I guess?

linera-execution/src/lib.rs
```diff
@@ -783,9 +786,44 @@ pub enum Query {
...
 }
 
+/// The outcome of the execution of a query.
+#[derive(Debug, PartialEq, Eq, Hash, Clone, Serialize, Deserialize)]
+pub struct QueryOutcome<Response = QueryResponse> {
+    pub response: Response,
+    pub operations: Vec<Operation>,
+}
+
+impl From<QueryOutcome<SystemResponse>> for QueryOutcome {
```
afck, line 796: Are we actually using this?

afck (approved):
Looks good! No blockers from my side.

examples/gen-nft/src/service.rs
```diff
@@ -175,43 +177,49 @@ impl QueryRoot {
...
     }
 
     async fn transfer(
         &self,
         source_owner: AccountOwner,
         token_id: String,
         target_account: Account,
-    ) -> Vec<u8> {
-        bcs::to_bytes(&Operation::Transfer {
+    ) -> [u8; 0] {
```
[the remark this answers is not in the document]
Twey: It doesn't implement `OutputType` :(

Apollo recommends returning a struct:

```rust
struct Response {
  success: bool,
  status: u16,
  message: String,
}
```

alluding to the result of the mutation, but since we're doing the mutation async we obviously can't. Maybe it's just `true` though.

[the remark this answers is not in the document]
jvff: The response will be ignored by the node service, so I figured an empty array would be the "most explicit" valid response that indicates that :thinking:

linera-execution/src/runtime.rs
jvff: I was thinking about changing the `transaction_tracker` in a separate PR, so that it is an `enum` with  the current `TransactionTracker` and a new `QueryTracker` or something, so that it would then schedule operations only in queries (and fail otherwise) and maybe only replay oracles if it's being called from a contract. I need to think more about this, so I left it for later :sweat_smile:

examples/counter/src/service.rs
jvff: It will get dropped by the node service, because it returns the hash of the new block.

linera-sdk-derive/src/lib.rs
jvff: Yes, unfortunately :/

linera-execution/src/lib.rs
jvff: Yes, [here](https://github.com/linera-io/linera-protocol/pull/3217/files/f4e9b4a21fa7d977f4b5774b7f48de40f766151b#diff-c8417387ee85ef169137900d33ff4ac909c49dce26129c49a994d748fc9c9c53R493).

linera-execution/src/runtime.rs
afck: Makes sense! :+1:

linera-execution/src/lib.rs
jvff: Ah, the force push removed it :/ Here's a more permanent link:

https://github.com/jvff/linera-protocol-archive/blob/13dbdf447c0f505f6d0644fb1cc47e0d29a4d659/linera-execution/src/execution.rs#L493

afck: So there's exactly one system query, and it returns the chain ID and balance? :thinking:

Twey:
Fixes https://github.com/linera-io/linera-protocol/pull/3217.

PR #3229, opened 2025-02-02
removed duplicate

Fallengirl:
examples/hex-game/README.md
second player player at - removed duplicate

linera-chain/src/block.rs
of same - of the same


afck (approved):

afck:
You need to make the same change in `schema.graphql` to make the tests pass. Alternatively, run:

```
cargo run --locked --bin linera-schema-export > linera-service-graphql-client/gql/service_schema.graphql
```

PR #3232, opened 2025-02-03
Limit waiting for a single node when downloading blobs.

afck:
## Motivation

When trying to download a blob sequentially from each validator, a single unresponsive validator can considerably slow down the client. (Even worse if they maliciously respond, after a delay, with a gRPC error code that makes the client retry the request.)

## Proposal

After 1s (configurable) try a second, after 4s try a third validator in parallel, after 9s etc.

## Test Plan

## Release Plan

## Links

### DIFF
--- a/linera-core/src/remote_node.rs
+++ b/linera-core/src/remote_node.rs
@@ -309,7 +309,11 @@ impl<N: ValidatorNode> RemoteNode<N> {
     }
 
     #[instrument(level = "trace", skip(validators))]
-    async fn download_blob(validators: &[Self], blob_id: BlobId) -> Option<Blob> {
+    async fn download_blob(
+        validators: &[Self],
+        blob_id: BlobId,
+        timeout: Duration,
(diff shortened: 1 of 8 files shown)


linera-core/src/remote_node.rs
```diff
@@ -309,12 +309,24 @@ impl<N: ValidatorNode> RemoteNode<N> {
...
         validators.shuffle(&mut rand::thread_rng());
-        for remote_node in validators {
-            if let Some(blob) = remote_node.try_download_blob(blob_id).await {
+        let mut stream = validators
+            .into_iter()
+            .zip(0..)
+            .map(|(remote_node, i)| async move {
+                tokio::time::sleep(timeout * i * i).await;
+                remote_node.try_download_blob(blob_id).await
+            })
```
deuszx, line 326: Genius.

deuszx (approved):

PR #3240, opened 2025-02-04
Faucet

Twey:
## Motivation

We would like to be able to create wallets from the Web.  Our current preferred flow for creating wallets involves contacting the faucet for the desired network, but the faucet is currently only available within `linera-service`.

## Proposal

Create new crates with client and server for the faucet, on which `linera-service` and `linera-web` both depend.

## Test Plan

Refactor: handled by existing tests.

## Release Plan

## Links


linera-faucet/server/Cargo.toml
```diff
@@ -0,0 +1,29 @@
+[package]
+name = "linera-faucet-server"
+version = "0.1.0"
```
afck, line 3: Not sure if this should follow the workspace version (`version.workspace = true`) instead?

afck (approved):

linera-faucet/server/Cargo.toml
Twey: It definitely should, thank you :)

PR #3241, opened 2025-02-04
compose.sh exits eagerly if building the images fails.

christos-h:
## Motivation

`compose.sh` will continue executing even if the build process fails.

## Proposal

Don't.

## Test Plan

CI will catch regressions.


Twey:
This seems to add a lot of config files as well?

PR #2668, opened 2024-10-21
Serialize BTreeMap with keys as strings

andresilva91:
## Motivation

When we try to save the wallet, the `pending_blobs` inside of `UserChain` will need to be serialized to JSON, but that map is currently a `BlobId`.

## Proposal

Write a custom serialization module for BTreeMaps that will serialize the keys as strings
Fixes #2665

## Test Plan

CI + stacked #2646 on top of this and ran the benchmark test, it did not fail in the serialization anymore

## Release Plan

- Nothing to do / These changes follow the usual release cycle.

PR #3242, opened 2025-02-04
Don't forget pending blobs if retrying at the next height.

afck:
## Motivation

This is a port of #3238 to the main branch.

## Proposal

Always set the pending blobs together with the pending block.

Also simplify some more blob-related client code.

## Test Plan

`test_re_propose_locked_block_with_blobs` was modified to re-propose a `PublishDataBlob` operation. This fails without the fix.

## Release Plan

## Links

- Original PR: #3238 

### DIFF
--- a/linera-client/src/unit_tests/wallet.rs
+++ b/linera-client/src/unit_tests/wallet.rs
@@ -1,8 +1,6 @@
 // Copyright (c) Zefchain Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-use std::collections::BTreeMap;
-
 use anyhow::anyhow;
 use linera_base::{
     crypto::KeyPair,
@@ -45,14 +43,12 @@ async fn test_save_wallet_with_pending_blobs() -> anyhow::Result<()> {
         WalletState::create_from_file(&wallet_path, Wallet::new(genesis_config, Some(37)))?;
     let mut context = ClientContext::new_test_client_context(storage, wallet);
     let blob = Blob::new_data(b"blob".to_vec());
-    let mut pending_blobs = BTreeMap::new();
-    pending_blobs.insert(blob.id(), blob);
     context
         .update_wallet_for_new_chain_with_pending_blobs(
             chain_id,
             Some(key_pair),
             clock.current_time(),
-            pending_blobs,
+            vec![blob],
         )
         .await?;
     context.save_wallet().await?;
--- a/linera-client/src/unit_tests/wallet.rs
+++ b/linera-client/src/unit_tests/wallet.rs
@@ -3,27 +3,34 @@
 
 use anyhow::anyhow;
 use linera_base::{
-    crypto::KeyPair,
-    data_types::{Amount, Blob},
-    identifiers::ChainId,
+    data_types::{Amount, Blob, BlockHeight},
+    identifiers::{ChainDescription, ChainId},
 };
-use linera_core::test_utils::{MemoryStorageBuilder, StorageBuilder, TestBuilder};
-use rand::SeedableRng as _;
+use linera_chain::data_types::ProposedBlock;
+use linera_core::{
+    client::PendingProposal,
+    test_utils::{MemoryStorageBuilder, StorageBuilder, TestBuilder},
+};
+use linera_execution::committee::Epoch;
+use rand::{rngs::StdRng, SeedableRng as _};
 
 use super::util::make_genesis_config;
-use crate::{client_context::ClientContext, config::WalletState, wallet::Wallet};
+use crate::{
+    client_context::ClientContext,
+    config::WalletState,
+    wallet::{UserChain, Wallet},
+};
 
 /// Tests whether we can correctly save a wallet that contains pending blobs.
 #[test_log::test(tokio::test)]
 async fn test_save_wallet_with_pending_blobs() -> anyhow::Result<()> {
-    let mut rng = rand::rngs::StdRng::seed_from_u64(42);
+    let mut rng = StdRng::seed_from_u64(42);
     let storage_builder = MemoryStorageBuilder::default();
     let clock = storage_builder.clock().clone();
     let mut builder = TestBuilder::new(storage_builder, 4, 1).await?;
     let chain_id = ChainId::root(0);
     builder.add_root_chain(0, Amount::ONE).await?;
     let storage = builder.make_storage().await?;
-    let key_pair = KeyPair::generate_from(&mut rng);
 
     let genesis_config = make_genesis_config(&builder);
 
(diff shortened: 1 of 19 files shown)


linera-core/src/client/chain_client_state.rs
```diff
@@ -50,7 +53,7 @@ impl ChainClientState {
         timestamp: Timestamp,
         next_block_height: BlockHeight,
         pending_block: Option<ProposedBlock>,
-        pending_blobs: BTreeMap<BlobId, Blob>,
+        pending_blobs: impl IntoIterator<Item = Blob>,
```
deuszx, line 56: In theory, this iterator can be non-empty for `pending_block = None`. What do you think about asserting that they're consistent?

linera-core/src/updater.rs
```diff
@@ -300,15 +298,26 @@ where
...
-                        .filter(|blob_id| !published_blobs.iter().any(|blob| blob.id() == *blob_id))
-                        .collect::<Vec<_>>();
-                    let missing_blob_ids = self.remote_node.node.missing_blob_ids(blob_ids).await?;
-                    let local_storage = self.local_node.storage_client();
+                    let published_blob_ids =
+                        BTreeSet::from_iter(proposal.content.block.published_blob_ids());
+                    blob_ids.retain(|blob_id| !published_blob_ids.contains(blob_id));
+                    let mut published_blobs = Vec::new();
+                    {
+                        let chain = self.local_node.chain_state_view(chain_id).await?;
```
deuszx, line 306: This change - from querying a `remote_node` to a `local_node` - seems like a win as well.

Was it even correct before?

deuszx: Ah, it was simply moved down below.

linera-core/src/client/mod.rs
```diff
@@ -2172,7 +2124,7 @@ where
         let committee = self.local_committee().await?;
         let max_size = committee.policy().maximum_block_proposal_size;
         block.check_proposal_size(max_size)?;
```
deuszx, line 2126: Do we check that `blobs` for this particular `block` are also within the size limits?

linera-core/src/client/mod.rs
```diff
@@ -2067,16 +2017,17 @@ where
...
 
     /// Executes a new block.
     ///
     /// This must be preceded by a call to `prepare_chain()`.
-    #[instrument(level = "trace", skip(operations))]
+    #[instrument(level = "trace", skip(operations, blobs))]
     async fn execute_block(
         &self,
         operations: Vec<Operation>,
+        blobs: impl IntoIterator<Item = Blob>,
```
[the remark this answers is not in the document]
afck: That's pretty much what we had before this PR: The blobs were put into a map and then looked up. That separated them from the publishing operations in the code and made it easy to cause this bug.

This way, the to-be-published blobs are kept together with the operations all the way. I think it's safer like this.

We could put the operations and blobs into a new struct—`BlockTemplate`? But maybe in a separate PR?

linera-core/src/client/mod.rs
deuszx: Even though they're threaded separately, they are part of the same block proposal.

linera-core/src/client/mod.rs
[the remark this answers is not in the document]
deuszx: I think the underlying problem was that we simply missed that the `pending_*` is actually represented with two variables - separately block and blobs. Maybe it'd be easier to not miss it if they were bundled together into a struct `PendingProposal { block, blobs }`.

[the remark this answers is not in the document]
afck: I'm working on `PendingProposal` and it simplifies things indeed.

But this particular function will still remain as it is, since we haven't created a `Block` yet: This is just operations, together with the published blobs.

linera-core/src/client/mod.rs
afck: Done in https://github.com/linera-io/linera-protocol/pull/3242/commits/f097f389c5da0c91605bdad1e68ccb47a319de41.

linera-core/src/client/mod.rs
```diff
@@ -2124,6 +2125,14 @@ where
         let committee = self.local_committee().await?;
         let max_size = committee.policy().maximum_block_proposal_size;
         block.check_proposal_size(max_size)?;
+        for blob in &blobs {
+            if published_blob_ids.contains(&blob.id()) {
+                committee
+                    .policy()
+                    .check_blob_size(blob.content())
```
deuszx, line 2124: Is there a global limit for published blobs' size? I.e. not only the per-blob size limit but also sum of all blobs published as part of the single proposal

afck: Only implicitly: There is a maximum blob size and a maximum number of published blobs per block. So the product of them is the maximum total size.

linera-client/src/unit_tests/wallet.rs
```diff
@@ -39,18 +46,29 @@ async fn test_save_wallet_with_pending_blobs() -> anyhow::Result<()> {
...
+            chain_id,
+            epoch: Epoch::ZERO,
+            incoming_bundles: vec![],
+            operations: vec![],
+            height: BlockHeight::ZERO,
+            timestamp: clock.current_time(),
+            authenticated_signer: None,
+            previous_block_hash: None,
+        },
+        blobs: vec![Blob::new_data(b"blob".to_vec())],
```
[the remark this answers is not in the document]
afck: It should be referenced, yes. But this is only a regression test for https://github.com/linera-io/linera-protocol/pull/2668, which was about serializing the wallet, so I'm not sure how much code we should add here just to make it a realistic scenario.

linera-core/src/client/mod.rs
```diff
@@ -2480,35 +2484,22 @@ where
...
-                            local_node
-                                .get_locking_blobs(&blob_ids, self.chain_id)
-                                .await?
-                                .ok_or_else(|| {
-                                    ChainClientError::InternalError("Missing local locking blobs")
-                                })?
-                        } else {
-                            self.pending_blobs().iter().cloned().collect()
-                        };
+                    LocalNodeError::BlobsNotFound(_) => {
```
deuszx, line 2502: How is it that we no longer need the code from the left?

afck: We know which blobs we need based on whether we are re-proposing a locking block or making a new proposal. So we fetch those in line 2453. (These are probably exactly the blobs in the chain manager, so there's room for more simplification, but this PR is already getting a bit large.)

linera-client/src/unit_tests/wallet.rs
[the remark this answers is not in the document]
deuszx: Got it. I thought that maybe there's an option for a smart constructor of some sort that "fills in" the `published_blobs` field based on the operations.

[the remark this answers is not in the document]
afck: Yes, maybe it should have that. But then where would it look up the blobs?

I'd also leave that for later.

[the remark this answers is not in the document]
deuszx: Ok, you've done enough here already 💪

linera-core/src/client/mod.rs
deuszx: Thanks.

deuszx (approved):

PR #3244, opened 2025-02-04
Return an error if a pending blob is not needed by any pending block.

afck:
## Motivation

We currently don't detect if the blob in `HandlePendingBlob` was needed at all.
(See [this discussion](https://github.com/linera-io/linera-protocol/pull/3204#discussion_r1939274824))

## Proposal

Return an error if no pending block needed it.

## Test Plan

CI should catch regressions.

## Release Plan

## Links

- Discussion: https://github.com/linera-io/linera-protocol/pull/3204#discussion_r1939274824

### DIFF
--- a/linera-chain/src/pending_blobs.rs
+++ b/linera-chain/src/pending_blobs.rs
@@ -44,14 +44,16 @@ where
         Ok(self.pending_blobs.get(blob_id).await?.flatten())
     }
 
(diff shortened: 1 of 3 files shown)


deuszx (approved):
Can you modify some existing test to verify that we do indeed return the new error?

deuszx (approved):

PR #3235, opened 2025-02-03
Remove max_in_flight from linera benchmark

ndr-ds:
## Motivation

We're gonna slowly change the design of this. Limiting the amount of in flight requests goes against our goal of stress testing a network with this

## Proposal

Remove that constraint

## Test Plan

Ran `linera benchmark` with a few parameter variations

## Release Plan

- Nothing to do / These changes follow the usual release cycle.

PR #3245, opened 2025-02-04
Parallelize process_inboxes_and_force_validator_updates

ndr-ds:
## Motivation

## Proposal

Paralellize initial validator sync step

## Test Plan

CI + ran locally

## Release Plan

### DIFF
--- a/linera-client/src/client_context.rs
+++ b/linera-client/src/client_context.rs
@@ -18,8 +18,9 @@ use linera_chain::types::ConfirmedBlockCertificate;
 use linera_core::{
     client::{BlanketMessagePolicy, ChainClient, Client, MessagePolicy},
     data_types::ClientOutcome,
-    join_set_ext::{JoinSet, JoinSetExt as _},
+    join_set_ext::JoinSet,
     node::CrossChainMessageDelivery,
+    JoinSetExt,
 };
 use linera_rpc::node_provider::{NodeOptions, NodeProvider};
 use linera_storage::Storage;
@@ -566,14 +567,44 @@ where
     W: Persist<Target = Wallet>,
 {
     pub async fn process_inboxes_and_force_validator_updates(&mut self) {
-        for chain_id in self.wallet.owned_chain_ids() {
-            let chain_client = self
-                .make_chain_client(chain_id)
-                .expect("chains in the wallet must exist");
-            self.process_inbox(&chain_client).await.unwrap();
-            chain_client.update_validators(None).await.unwrap();
-            self.update_wallet_from_client(&chain_client).await.unwrap();
+        let chain_clients = self
+            .wallet
+            .owned_chain_ids()
+            .iter()
+            .map(|chain_id| {
+                self.make_chain_client(*chain_id)
+                    .expect("chains in the wallet must exist")
+            })
+            .collect::<Vec<_>>();
+
+        let mut join_set = task::JoinSet::new();
+        for chain_client in chain_clients {
+            join_set.spawn(async move {
+                Self::process_inbox_without_updating_wallet(&chain_client)
+                    .await
+                    .expect("Processing inbox should not fail!");
+                chain_client
+            });
         }
+
+        let chain_clients = join_set.join_all().await;
+        for chain_client in &chain_clients {
+            self.update_wallet_from_client(chain_client).await.unwrap();
+        }
+    }
+
+    async fn process_inbox_without_updating_wallet(
+        chain_client: &ChainClient<NodeProvider, S>,
+    ) -> Result<Vec<ConfirmedBlockCertificate>, Error> {
+        // Try processing the inbox optimistically without waiting for validator notifications.
+        chain_client.synchronize_from_validators().await?;
+        let (certificates, maybe_timeout) = chain_client.process_inbox_without_prepare().await?;
+        assert!(
+            maybe_timeout.is_none(),
+            "Should not timeout within benchmark!"
+        );
+
+        Ok(certificates)
     }
 
     /// Creates chains if necessary, and returns a map of exactly `num_chains` chain IDs
--- a/linera-service/src/linera/main.rs
+++ b/linera-service/src/linera/main.rs
@@ -764,10 +764,15 @@ impl Runnable for Job {
                 transactions_per_block,
                 fungible_application_id,
             } => {
+                let start = Instant::now();
                 // Below all block proposals are supposed to succeed without retries, we
                 // must make sure that all incoming payments have been accepted on-chain
                 // and that no validator is missing user certificates.
                 context.process_inboxes_and_force_validator_updates().await;
+                info!(
+                    "Processed inboxes and forced validator updates in {} ms",
+                    start.elapsed().as_millis()
+                );
 
                 let key_pairs = context
                     .make_benchmark_chains(num_chains, tokens_per_chain)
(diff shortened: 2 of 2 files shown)


ndr-ds:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/3245?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#3245** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/3245?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/3245?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

linera-client/src/client_context.rs
```diff
@@ -411,6 +412,46 @@ where
...
+
+    pub async fn start_listening_for_notifications(
+        &mut self,
+        chain_client: &ChainClient<NodeProvider, S>,
+    ) -> Result<NotificationStream, ChainClientError> {
+        // Start listening for notifications, so we learn about new rounds and blocks.
+        let (listener, _listen_handle, notification_stream) = chain_client.listen().await?;
+        self.chain_listeners.spawn_task(listener);
+        Ok(notification_stream)
+    }
```
afck, line 453: Do these need to be `pub`? And shouldn't they be in the `#[cfg(feature = "benchmark")]` impl below?

linera-client/src/client_context.rs
```diff
@@ -411,6 +412,46 @@ where
...
+    ) -> Result<Vec<ConfirmedBlockCertificate>, Error> {
+        let mut certificates = Vec::new();
+        // Try processing the inbox optimistically without waiting for validator notifications.
+        let (new_certificates, maybe_timeout) = {
+            chain_client.synchronize_from_validators().await?;
+            let result = chain_client.process_inbox_without_prepare().await;
+            result?
+        };
+        certificates.extend(new_certificates);
+        if maybe_timeout.is_none() {
```
afck, line 427: Can this ever be `Some` in the benchmark? Aren't these all single-user chains, where we expect the first proposal to succeed?

linera-client/src/client_context.rs
```diff
@@ -584,14 +625,44 @@ where
...
+            .wallet
+            .owned_chain_ids()
+            .iter()
+            .map(|chain_id| {
+                self.make_chain_client(*chain_id)
+                    .expect("chains in the wallet must exist")
+            })
+            .collect::<Vec<_>>();
+
+        let mut chain_client_and_notification_stream = Vec::new();
```
afck, line 638:
```suggestion
        let mut chain_clients_and_notification_streams = Vec::new();
```

linera-client/src/client_context.rs
ndr-ds: Good point! They're all single super owner chains. This should simplify the code quite a bit :)

ndr-ds:
This one is significantly faster, 850ms in parallel vs 3500ms serially

linera-client/src/client_context.rs
```diff
@@ -566,16 +567,55 @@ where
...
+                chain_client.update_validators(None).await.unwrap();
+                chain_client
+            });
+        }
+
+        let chain_clients = join_set.join_all().await;
+        for chain_client in &chain_clients {
+            self.update_wallet_from_client(chain_client).await.unwrap();
+        }
+        if chain_clients.is_empty() {
```
afck, line 595: Do you mean if it's _not_ empty?

linera-client/src/client_context.rs
```diff
@@ -566,16 +567,55 @@ where
...
+    async fn process_inbox_without_updating_wallet(
+        chain_client: &ChainClient<NodeProvider, S>,
+    ) -> Result<Vec<ConfirmedBlockCertificate>, Error> {
+        let mut certificates = Vec::new();
+        // Try processing the inbox optimistically without waiting for validator notifications.
+        let (new_certificates, maybe_timeout) = {
+            chain_client.synchronize_from_validators().await?;
+            let result = chain_client.process_inbox_without_prepare().await;
+            result?
+        };
```
afck, line 609: Why the curly brackets?

```suggestion
        chain_client.synchronize_from_validators().await?;
        let (new_certificates, maybe_timeout) = chain_client.process_inbox_without_prepare().await?;
```

linera-client/src/client_context.rs
```diff
@@ -566,16 +567,55 @@ where
...
+        }
+        if chain_clients.is_empty() {
+            self.save_wallet().await.unwrap();
         }
     }
 
+    async fn process_inbox_without_updating_wallet(
+        chain_client: &ChainClient<NodeProvider, S>,
+    ) -> Result<Vec<ConfirmedBlockCertificate>, Error> {
+        let mut certificates = Vec::new();
```
afck, line 603: This is unnecessary. Just return `new_certificates` and rename it to `certificates`.

linera-client/src/client_context.rs
ndr-ds: So `update_wallet_from_client` already calls `save_wallet` inside of it. If we have any chain clients, we will have saved the wallet already. But actually, if chain clients is empty, then we don't even have to save anything just in case, actually, because that'll mean that none of this code before got executed, basically, so there will be nothing to save 🤔 gonna delete this if statement then

linera-client/src/client_context.rs
ndr-ds: Ah lol my bad, good catch. Overlooked that while refactoring 😅

linera-client/src/client_context.rs
ndr-ds: True, I guess now it's not necessary anymore. Thanks!

linera-client/src/client_context.rs
```diff
@@ -566,14 +567,48 @@ where
...
+            .collect::<Vec<_>>();
+
+        let mut join_set = task::JoinSet::new();
+        for chain_client in chain_clients {
+            join_set.spawn(async move {
+                Self::process_inbox_without_updating_wallet(&chain_client)
+                    .await
+                    .expect("Processing inbox should not fail!");
+                chain_client
+                    .update_validators(None)
```
afck, line 587: I wonder if that's even necessary. `process_inbox_without_prepare` already updates the validators about the new certificates.

afck (approved):

linera-client/src/client_context.rs
ndr-ds: It's probably an unlikely case, but from what I looked, `process_inbox_without_prepare` will only execute a new block (and then update the validators) if we have pending message bundles. If that's not the case, then we won't update the validators.
`update_validators` in this case will only update anything if our committee has changed. So I guess in the case where we're running this against the testnet or something, and some validator gets added/removed, we should keep this around to make sure things will be properly updated in that case.

afck: But if the committee changed, we _do_ receive a message about it, so there _will_ be a pending message bundle. (At least until we have event streams—but even if, I'd say it isn't the benchmark's task to update the validators about the admin chain.)

ndr-ds:
### Merge activity

* **Feb 6, 1:28 PM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/linera-io/linera-protocol/3245).

PR #3246, opened 2025-02-04
make_benchmark_chains improvements

ndr-ds:
## Motivation

## Proposal

Paralellize creating the benchmark chains

## Test Plan

CI + ran locally

## Release Plan

### DIFF
--- a/linera-client/src/client_context.rs
+++ b/linera-client/src/client_context.rs
@@ -614,79 +614,137 @@ where
         num_chains: usize,
         balance: Amount,
     ) -> Result<HashMap<ChainId, KeyPair>, Error> {
-        let mut key_pairs = HashMap::new();
+        let mut benchmark_chains = HashMap::new();
+        let start = Instant::now();
         for chain_id in self.wallet.owned_chain_ids() {
-            if key_pairs.len() == num_chains {
+            if benchmark_chains.len() == num_chains {
                 break;
             }
-            let Some(key_pair) = self
+            // This should never panic, because `owned_chain_ids` only returns the owned chains that
+            // we have a key pair for.
+            let key_pair = self
                 .wallet
                 .get(chain_id)
                 .and_then(|chain| chain.key_pair.as_ref().map(|kp| kp.copy()))
-            else {
-                continue;
-            };
+                .unwrap();
             let chain_client = self.make_chain_client(chain_id)?;
+
             let ownership = chain_client.chain_info().await?.manager.ownership;
             if !ownership.owners.is_empty() || ownership.super_owners.len() != 1 {
                 continue;
             }
-            key_pairs.insert(chain_id, key_pair);
+            benchmark_chains.insert(chain_client.chain_id(), key_pair);
         }
 
+        info!(
+            "Got {} chains from the wallet in {} ms",
+            benchmark_chains.len(),
+            start.elapsed().as_millis()
+        );
+
+        let chains_from_wallet = benchmark_chains.len();
+        let start = Instant::now();
         let default_chain_id = self
             .wallet
             .default_chain()
             .expect("should have default chain");
+        let num_chains_to_create = num_chains - benchmark_chains.len();
+        let operations_per_block = 900; // Over this we seem to hit the block size limits.
+
+        let mut key_pairs = Vec::new();
+        for _ in (0..num_chains_to_create).step_by(operations_per_block) {
+            key_pairs.push(self.wallet.generate_key_pair());
+        }
+        let mut key_pairs_iter = key_pairs.into_iter();
+        let admin_id = self.wallet.genesis_admin_chain();
         let chain_client = self.make_chain_client(default_chain_id)?;
-        while key_pairs.len() < num_chains {
-            let key_pair = self.wallet.generate_key_pair();
-            let (epoch, committees) = chain_client.epoch_and_committees(default_chain_id).await?;
-            let epoch = epoch.expect("default chain should be active");
-            // Put at most 1000 OpenChain operations in each block.
-            let num_new_chains = (num_chains - key_pairs.len()).min(1000);
-            let config = OpenChainConfig {
-                ownership: ChainOwnership::single_super(key_pair.public().into()),
-                committees,
-                admin_id: self.wallet.genesis_admin_chain(),
-                epoch,
+        for i in (0..num_chains_to_create).step_by(operations_per_block) {
+            let num_new_chains = operations_per_block.min(num_chains_to_create - i);
+            let key_pair = key_pairs_iter.next().unwrap();
+
+            let certificate = Self::execute_open_chains_operations(
+                num_new_chains,
+                &chain_client,
                 balance,
-                application_permissions: Default::default(),
-            };
-            let operations = iter::repeat(Operation::System(SystemOperation::OpenChain(config)))
-                .take(num_new_chains)
-                .collect();
-            let certificate = chain_client
-                .execute_operations(operations, vec![])
-                .await?
-                .expect("should execute block with OpenChain operations");
+                &key_pair,
+                admin_id,
+            )
+            .await?;
+            info!("Block executed successfully");
+
             let block = certificate.block();
             let timestamp = block.header.timestamp;
+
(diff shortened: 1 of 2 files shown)


ndr-ds:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/3246?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#3246** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/3246?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/3246?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

linera-client/src/client_context.rs
```diff
@@ -672,79 +672,181 @@ where
...
+            if let Some((chain_id, key_pair)) = result.unwrap()? {
+                benchmark_chains.insert(chain_id, key_pair);
             }
-            key_pairs.insert(chain_id, key_pair);
         }
 
+        join_set.abort_all();
+
+        info!(
+            "Got {} chains from the wallet in {} ms",
```
afck, line 719: Have you tried out if the parallelization makes a difference here? After all, these are just accessing the local database so far.

linera-client/src/client_context.rs
```diff
@@ -672,79 +672,181 @@ where
...
+        }
+
+        let chains_from_wallet = benchmark_chains.len();
+        let start = Instant::now();
         let default_chain_id = self
             .wallet
             .default_chain()
             .expect("should have default chain");
+        let num_chains_to_create = num_chains - benchmark_chains.len();
+        let operations_per_block = 900; // Over this we seem to hit the block size limits.
```
afck, line 734: Would be good to have a test that verifies that this number of `OpenChain` operations does indeed fit in a block. Because `OpenChain` contains e.g. `ChainOwnership` and `ApplicationPermissions`, and these structures might get additional fields in the future.

Also, the block size limit is configurable: Maybe we should get it from the network and calculate `operations_per_block` based on it?

linera-client/src/client_context.rs
```diff
@@ -672,79 +672,181 @@ where
...
+            join_set.spawn(async move {
+                child_client.process_inbox().await?;
+                Ok(child_client)
+            });
+        }
+        let chain_clients = join_set
+            .join_all()
+            .await
+            .into_iter()
+            .collect::<Result<Vec<_>, Error>>()?;
```
afck, line 808: I think you could use the `try_join!` macro for this.

(And in general, the `join!` macro instead of `JoinSet`, for tasks that don't return results. Might also be a bit shorter, but it's a matter of taste.)

linera-client/src/client_context.rs
```diff
@@ -672,79 +672,181 @@ where
...
+                Self::execute_open_chains_operations(
+                    num_new_chains,
+                    chain_client,
+                    balance,
+                    key_pair,
+                    admin_id,
+                )
+                .await
+            });
+        }
```
afck, line 758: It doesn't make sense to do this concurrently: These are all on the same chain, will conflict with each other, and in the end get executed sequentially.

afck (changes requested):

linera-client/src/client_context.rs
ndr-ds: Since I'm trying to paralellize things, I don't think this is correct. From the [documentation](https://docs.rs/tokio/latest/tokio/macro.join.html) of `join!`:
> The join! macro takes a list of async expressions and evaluates them concurrently on the same task.

With `JoinSet`, I can spawn async work to be done in different tasks, which will be run by the tokio runtime in different threads, thus paralellizing the work amongst the CPU cores. If I use `join!`, it'll have the same behavior as `FuturesUnordered`, which is just executing the futures concurrently in the same task/thread, which is not what I want in this case.

linera-client/src/client_context.rs
ndr-ds: I guess that's true, they will be executed sequentially, so maybe the parallelization isn't doing much here. The fact that they conflict with each other IMHO is not a problem, is it? As long as we get the chains created, the order in which the blocks that have the open chain operations appear in the chain shouldn't matter in this case, right?
Anyways, I can time this and confirm we don't get much gain from the parallelization. If that's the case, I'll stop doing it here

linera-client/src/client_context.rs
ndr-ds: Ah, good point, I guess I can try to calculate the maximum `operations_per_block` from the limit. If that's too involved, I can at least add the test

linera-client/src/client_context.rs
ndr-ds: This is mostly for parallelizing the chain info queries, but I can time it as well to see if it makes a difference.

linera-client/src/client_context.rs
afck: It _is_ a problem: Not sure if we're using super owners or regular owners here. If it's super owners it can simply brick the chains, so it will never complete at all. If it's regular owners there will be lots of conflicts that actually make it _slower_ than doing it sequentially.

linera-client/src/client_context.rs
afck: But shouldn't tokio use at least as many threads as there are CPUs anyway, so that even `join!` would distribute the tasks across all CPUs?

ndr-ds: The documentation also says this:
> If parallelism is required, spawn each async expression using tokio::spawn and pass the join handle to join!.

`join!` takes a list of futures. It will execute these futures concurrently on the same task. So there's no parallelism involved. If you need parallelism, you need to spawn different tasks, and then pass these join handles to `join!` instead. In that case IMHO it seems like a much more straight forward solution (and also seems more correct), to just use a `JoinSet` which was made for exactly this purpose, AFAIU.
Spawn the tasks in the `JoinSet`, then await all of them in parallel.

linera-client/src/client_context.rs
ndr-ds: Wow, you're right. Parallelizing it makes it slower 🤦🏻‍♂️ That's on me for not actually measuring it. I'll go through all the parallelizations on these PRs and make sure they're actually speeding things up, I should've done this in the first place. My bad! And thanks for the nudge

linera-client/src/client_context.rs
ndr-ds: So, what fails here when I increase this number is the size of the executed block. As we move this forward, the actual size of the executed block will vary because we'll be processing messages with transfers from other chains and stuff like that. So it'll actually be pretty hard to calculate the maximum `operations_per_block` beforehand when we're building the actual block.
So it's better to just go with the test route IMHO

linera-client/src/client_context.rs
```diff
@@ -623,79 +623,162 @@ where
...
+            .into_iter()
+            .map(|chain_id| {
+                // This should never panic, because `owned_chain_ids` only returns the owned chains that
+                // we have a key pair for.
+                let key_pair = self
+                    .wallet
+                    .get(chain_id)
+                    .and_then(|chain| chain.key_pair.as_ref().map(|kp| kp.copy()))
+                    .unwrap();
+                Ok::<(KeyPair, ChainClient<NodeProvider, S>), Error>((
```
afck, line 639: Is that annotation necessary? Or maybe only `<_, Error>` is needed?

afck (approved):

linera-client/src/client_context.rs
ndr-ds: True, doesn't need to be this verbose 😅

ndr-ds:
### Merge activity

* **Feb 6, 1:28 PM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/linera-io/linera-protocol/3246).

PR #3247, opened 2025-02-04
supply_fungible_tokens improvements

ndr-ds:
## Motivation

Will start paralellizing some parts of this existing flow

## Proposal

Paralellize supplying the fungible tokens

## Test Plan

CI + ran locally

## Release Plan

- Nothing to do / These changes follow the usual release cycle.

PR #3248, opened 2025-02-04
Parallelize staging block execution

ndr-ds:
## Motivation

## Proposal

Paralellize staging the block execution

## Test Plan

CI + ran locally

## Release Plan

### DIFF
--- a/linera-service/src/linera/main.rs
+++ b/linera-service/src/linera/main.rs
@@ -802,17 +802,28 @@ impl Runnable for Job {
                     fungible_application_id,
                 );
                 let num_proposal = proposals.len();
-                let mut values = HashMap::new();
 
+                let mut join_set = JoinSet::new();
                 for rpc_msg in &proposals {
                     if let RpcMessage::BlockProposal(proposal) = rpc_msg {
-                        let executed_block = context
-                            .stage_block_execution(proposal.content.block.clone(), None)
-                            .await?;
-                        let value = Hashed::new(ConfirmedBlock::new(executed_block));
-                        values.insert(value.hash(), value);
+                        let block = proposal.content.block.clone();
+                        let client = context.client.clone();
+                        join_set.spawn(async move {
+                            let executed_block = client
+                                .local_node()
+                                .stage_block_execution(block, None)
+                                .await?
+                                .0;
+                            let value = Hashed::new(ConfirmedBlock::new(executed_block));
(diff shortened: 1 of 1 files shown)


ndr-ds:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/3248?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#3248** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/3248?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/3248?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

linera-service/src/linera/main.rs
```diff
@@ -793,17 +793,28 @@ impl Runnable for Job {
...
+                            let executed_block = client
+                                .local_node()
+                                .stage_block_execution(block, None)
+                                .await?
+                                .0;
+                            let value = Hashed::new(ConfirmedBlock::new(executed_block));
+                            Ok::<_, anyhow::Error>((value.hash(), value))
+                        });
                     }
                 }
```
ndr-ds, line 812: I guess for this we can only parallelize the work on different chains, so I need to change this to take that into account

afck: I think originally we only had one proposal per chain.

If we do multiple, they have to be sent in the order of their block heights, yes.

ndr-ds: Ah, good point, then maybe we're good (for now), because we're still doing just one proposal per chain

afck (approved):
That only makes sense if there are no two proposals on the same chain. But that's currently the case, right?

deuszx:
> That only makes sense if there are no two proposals on the same chain. But that's currently the case, right?

If there are never more than two propoals - what's the point of parallelising?

Also, have we measured the perf improvements?

ndr-ds:
> That only makes sense if there are no two proposals on the same chain. But that's currently the case, right?

Exactly! We create one proposal per chain, currently.

ndr-ds:
This actually is slower, surprisingly

PR #3253, opened 2025-02-05
Fix incorrect comment refers

paulfisherz:
## Release Plan

    - be released in a new SDK,
    - be released in a new SDK,
      SDK soon.

## Links


linera-views/src/batch.rs
```diff
@@ -168,7 +168,7 @@ impl UnorderedBatch {
 /// That is, there should not be any two prefixes p1 and p2 such that p1 < p2 for
 /// the lexicographic ordering on `Vec<u8>` entries.
 /// Under this condition we have equivalence between the following two statements:
-/// * There is an key_prefix in `key_prefiw_set` that matches `key`.
+/// * There is an key_prefix in `key_prefix_set` that matches `key`.
```
afck, line 171: Also:

```suggestion
/// * There is a prefix in `key_prefix_set` that matches `key`.
```

afck (approved):

PR #3254, opened 2025-02-05
Use configured client options in `assign_new_chain_to_key`.

### LINKED ISSUE #3237: Use configured client options in `assign_new_chain_to_key`.
We use the default option there.

Investigate why we can't just use `context.client`. If we can't, at least use the same option.

afck:
## Motivation

In `assign_new_chain_to_key` we need a temporary `Client` instance tracking the new chain and its parent. We currently set all options to their defaults.

## Proposal

Use the configured options instead.

## Test Plan

This should at best negligibly change performance when assigning a key to a new chain. CI will catch regressions.

## Release Plan

## Links

- Closes #3237

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -223,6 +223,41 @@ impl<P, S: Storage + Clone> Client<P, S> {
         }
     }
 
+    /// Returns a clone with a different set of tracked chains.
+    pub fn clone_with(
+        &self,
+        validator_node_provider: P,
+        name: impl Into<String>,
+        tracked_chains: impl IntoIterator<Item = ChainId>,
+        long_lived_services: bool,
+    ) -> Self {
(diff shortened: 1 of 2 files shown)


linera-core/src/client/mod.rs
```diff
@@ -223,6 +223,41 @@ impl<P, S: Storage + Clone> Client<P, S> {
         }
     }
 
+    /// Returns a clone with a different set of tracked chains.
+    pub fn clone_with(
+        &self,
+        validator_node_provider: P,
+        name: impl Into<String>,
+        tracked_chains: impl IntoIterator<Item = ChainId>,
+        long_lived_services: bool,
```
deuszx, line 232: I think the set of input arguments for overriding is bit arbitrary (or rather dictated by the current need) and a more generic approach with the builder pattern would be better in long run but it's a lot of typing and can wait.

deuszx (approved):

PR #3257, opened 2025-02-05
`linera-client`: make sure to use `linera_base::time`

Twey:
## Motivation

`std::time` will panic if used on `wasm32-unknown-unknown`, since without making additional assumptions about the execution environment no clock syscalls can be used.  A `std::time::Duration` is used in `linera_client::client_options`.

## Proposal

We have time primitives exported from `linera_base::time` that are Web-safe (using `web-time` on the Web).  Replace `std::time::Duration` with `linera_base::time::Duration` in `linera-client`. 

## Release Plan

## Links


afck (approved):
(My bad, probably!)

PR #3264, opened 2025-02-06
make_benchmark_block_proposals now returns BlockProposals

ndr-ds:
## Motivation

This `RpcMessage` type is needed only when calling `mass_broadcast`, and it makes the code a bit more awkward because you have to check if it's a block proposal, when it'll always be

## Proposal

Return the actual `BlockProposal`s instead, and only convert them when needed.

## Test Plan

CI + ran locally

## Release Plan

- Nothing to do / These changes follow the usual release cycle.

PR #3249, opened 2025-02-04
Make generated block proposals cancel each other out

ndr-ds:
## Motivation

The block proposals are supposed to rotate and cancel each other out, but there's currently a bug in how we do it.

## Proposal

Fix the bug

## Test Plan

CI + ran locally

## Release Plan

- Nothing to do / These changes follow the usual release cycle.

PR #3265, opened 2025-02-06
Introduce long running mode to linera benchmark

ndr-ds:
## Motivation

We need a way of proposing blocks to a network at a fixed BPS rate

## Proposal

Introduce a long running mode that you can specify a BPS rate, and it'll attempt to reach that. This is the most naive version, doesn't scale very well.
We need to figure out how much we're gonna paralellize these before having another binary run this in different processes with different wallets.

## Test Plan

Ran it locally against a local kind network.

## Release Plan


ndr-ds:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/3265?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#3265** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/3265?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/3265?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3264** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/3264?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3249** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/3249?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

burakacar7:
perfect

linera-client/src/client_context.rs
```diff
@@ -41,28 +41,47 @@ use {
...
     },
     linera_sdk::abis::fungible,
     std::{collections::HashMap, iter},
     tokio::task,
-    tracing::{error, trace},
+    tracing::{error, trace, warn},
 };
 
+#[cfg(feature = "benchmark")]
+fn deserialize_response(response: RpcMessage) -> Option<ChainInfoResponse> {
```
afck, line 71: (You could make it a static method of `ChainClient` below; one less `cfg` attribute.)

linera-client/src/client_options.rs
```diff
@@ -602,6 +602,11 @@ pub enum ClientCommand {
         /// If none is specified, the benchmark uses the native token.
         #[arg(long)]
         fungible_application_id: Option<linera_base::identifiers::ApplicationId>,
+
+        /// If provided, will be long running, and block proposals will be send at the
```
afck, line 606:
```suggestion
        /// If provided, will be long running, and block proposals will be sent at the
```

linera-client/src/client_context.rs
```diff
@@ -566,6 +585,206 @@ where
...
+                    );
+                } else {
+                    tokio::time::sleep(Duration::from_secs(1) - elapsed).await;
+                    info!("Achieved {} BPS/{} TPS", bps, bps * transactions_per_block);
+                }
+                start = Instant::now();
+                num_sent_proposals = 0;
+            }
+        }
+    }
```
afck, line 786: This will be slower than running `run_benchmark` in a loop. What's the advantage?

ndr-ds: The goal here is having block proposals be sent at a given/controlled BPS/TPS. This will allow us to ramp this up while watching metrics (client's resources like CPU/Memory/IO, and validator metrics like proxy latency, etc). This way we can figure out what the max TPS is for different network configurations, and also play around with parameters to find potential bottlenecks.
But like I said, this is version 1 from the doc. It's not very scalable. So this will become more useful as we build the next versions.

ndr-ds: I'm keeping the old version around btw because I think if we ever replace `linera-benchmark` with something that calls this, we'll need the old version for CI, probably, as we won't want it to run continuously

afck (commented):
This is not a blocker from my side, but I don't think we'll need two separate implementations for this in the long run. A limit of blocks per chain that can be set to 1 would make `run_benchmark` a special case of `run_benchmark_long_running`.

ndr-ds:
> This is not a blocker from my side, but I don't think we'll need two separate implementations for this in the long run. A limit of blocks per chain that can be set to 1 would make `run_benchmark` a special case of `run_benchmark_long_running`.

True, we can think about unifying it at some point, but it just felt easier for now to keep them separate. I can look into that in a follow up PR

ndr-ds:
Talked with @afck offline and decided to merge the follow up PR changes here

afck (approved):
(As discussed, this temporarily removes the benchmark implementation that uses the chains in parallel.)

ndr-ds:
I'll do one task per chain as a follow up

ndr-ds:
### Merge activity

* **Feb 13, 1:05 PM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/linera-io/linera-protocol/3265).

Issue #2025, opened 2024-05-14
Client eagerly audits chains


jvff:
# Motivation

The client always eagerly executes entire chains it keeps track of. This takes time and uses resources, and in many cases full execution of the chain isn't needed.

# Proposal

When receiving a message, for example, the client only needs to check that the certificate of the block that sent the message is valid. Therefore, if the certificate is signed by validators that provide enough stake/weight to be considered a quorum of the current epoch, the client can trust that the incoming message is valid. Otherwise, the client can only trust the certificate if there is a sequence of certificates starting from it and ending with a certificate that is signed by a quorum of the current epoch.

In both cases, the blocks don't need to be executed.

afck:
Duplicate of #3969.

PR #3289, opened 2025-02-10
Refactorings to make separation of execution from certificate handling easier.

afck:
## Motivation

To make https://github.com/linera-io/linera-protocol/issues/2025 possible, we need to better separate the code that verifies a certificate and handles outboxes and inboxes from the block execution code.

## Proposal

Don't call `execute_init_message` from `execute_block` at all: That is already unnecessary because it is called from `process_confirmed_block` and `receive_message_bundle`.

Extract some functions from `execute_block` and `process_confirmed_block` to make them easier to read, and deduplicate some of the logic.

## Test Plan

CI

## Release Plan

## Links

- Related to #2025 

### DIFF
--- a/linera-chain/src/chain.rs
+++ b/linera-chain/src/chain.rs
@@ -565,13 +565,35 @@ where
         }
     }
 
+    /// Verifies that the block's first message is `OpenChain`. Initializes the chain if necessary.
+    async fn execute_init_message_from(
+        &mut self,
+        block: &ProposedBlock,
+        local_time: Timestamp,
+    ) -> Result<(), ChainError> {
+        let (in_bundle, posted_message, config) = block
+            .starts_with_open_chain_message()
+            .ok_or_else(|| ChainError::InactiveChain(block.chain_id))?;
+        if self.is_active() {
+            return Ok(()); // Already initialized.
+        }
+        let message_id = MessageId {
+            chain_id: in_bundle.origin.sender,
+            height: in_bundle.bundle.height,
+            index: posted_message.index,
+        };
+        self.execute_init_message(message_id, config, block.timestamp, local_time)
+            .await
+    }
+
+    /// Initializes the chain using the given configuration.
     pub async fn execute_init_message(
         &mut self,
         message_id: MessageId,
         config: &OpenChainConfig,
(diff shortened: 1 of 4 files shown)


linera-chain/src/chain.rs
```diff
@@ -684,31 +708,7 @@ where
...
-            if !self.is_active() {
-                let message_id = MessageId {
-                    chain_id: in_bundle.origin.sender,
-                    height: in_bundle.bundle.height,
-                    index: posted_message.index,
-                };
-                self.execute_init_message(message_id, config, block.timestamp, local_time)
-                    .await?;
-            }
-        }
```
deuszx, line 711: it's very difficult to see if we indeed call `execute_init_message` prior to calling this `execute_block` function.

deuszx (approved):
Do we have tests covering these call paths?

linera-chain/src/chain.rs
afck: It's called in `receive_message_bundle` [here](https://github.com/linera-io/linera-protocol/blob/a592336640d7d9d0ffd833fe99952fc533d1df2c/linera-chain/src/chain.rs#L504), so if we add an `OpenChain` message to an inbox, it is called, and we don't need to call it again when processing the first block.

If we don't see `OpenChain` before the first block _proposal_, we reject the proposal anyway.

If we don't see `OpenChain` before the first block and the block is already committed, then [we call it explicitly here](https://github.com/linera-io/linera-protocol/blob/a592336640d7d9d0ffd833fe99952fc533d1df2c/linera-core/src/chain_worker/state/attempted_changes.rs#L332) in `process_confirmed_block`, before executing the block.

So the call in `execute_block` was already redundant.

afck:
We have lots of tests that create child chains. In most of them, the `OpenChain` message is received before the first block is processed.

To test the other case, `test_end_to_end_assign_greatgrandchild_chain` assigns a new chain on a client, which automatically downloads the parent chain. But since this is a great-grandchild of a root chain, the grandparent chain is _not_ downloaded. So the parent chain's first block is processed without the corresponding `OpenChain` message having been received first.

PR #3290, opened 2025-02-10
update lib.rs / transaction.rs 

XxAlex74xX:
exchange of token - fix - exchange for token `2x`
An message - fix - A message


afck (approved):

linera-execution/src/lib.rs
```diff
@@ -1400,6 +1400,6 @@ pub struct InvalidWasmRuntime(String);
 doc_scalar!(Operation, "An operation to be executed in a block");
 doc_scalar!(
     Message,
-    "An message to be sent and possibly executed in the receiver's block."
```
afck, line 1403: The exact same change needs to be made in `linera-service-graphql-client/gql/service_schema.graphql` to make the tests pass.

PR #3298, opened 2025-02-12
Fix dangling endpoint to chain worker actor that failed to start

### LINKED ISSUE #3295: Invalid chain worker endpoint can end up in the cache
# Description

When a request for a chain worker that's not already running is made, the following steps are made:

1. Create a channel to communicate with the actor
2. Add the sender endpoint to the chain worker cache
3. Start the actor with the receiver endpoint

However, if step 3 fails, the actor doesn't run and the receiver is immediately dropped. This means that any subsequent queries to the same chain will get the sender endpoint and fail to send a message.

jvff:
## Motivation

Testnet is currently experiencing some panics when an attempt is made to send a request to a chain worker actor. The panic message says that the endpoint is closed. One possible cause is #3295, where a storage error could cause an endpoint to be added to the chain worker cache but its respective actor task wouldn't start.

## Proposal

Refactor the `ChainWorkerActor` so that it has a `run` method that loads the chain state from storage and then handles the incoming requests. If loading the state fails, then the error is reported to the next requester, and loading will be reattempted while the actor is running and receiving requests.

## Test Plan

CI should catch any regressions caused by this refactor.

#3299 tracks having a regression test for the issue.

## Release Plan

This needs to be hotfixed because it may fix a bug that's in production.

## Links

- Closes #3295

### DIFF
--- a/linera-core/src/chain_worker/actor.rs
+++ b/linera-core/src/chain_worker/actor.rs
@@ -176,6 +176,44 @@ impl<StorageClient> ChainWorkerActor<StorageClient>
 where
     StorageClient: Storage + Clone + Send + Sync + 'static,
 {
+    /// Runs the [`ChainWorkerActor`], first by loading the chain state from `storage` then
+    /// handling all `incoming_requests` as they arrive.
+    ///
+    /// If loading the chain state fails the next request will receive the error reported by the
+    /// `storage`, and the actor will then try again to load the state.
+    pub async fn run(
+        config: ChainWorkerConfig,
+        storage: StorageClient,
+        executed_block_cache: Arc<ValueCache<CryptoHash, Hashed<Block>>>,
+        tracked_chains: Option<Arc<RwLock<HashSet<ChainId>>>>,
+        delivery_notifier: DeliveryNotifier,
+        chain_id: ChainId,
+        mut incoming_requests: mpsc::UnboundedReceiver<ChainWorkerRequest<StorageClient::Context>>,
+    ) {
+        let actor = loop {
+            let load_result = Self::load(
+                config.clone(),
+                storage.clone(),
+                executed_block_cache.clone(),
+                tracked_chains.clone(),
+                delivery_notifier.clone(),
+                chain_id,
+            )
+            .await
+            .inspect_err(|error| warn!("Failed to load chain state: {error:?}"));
+
+            match load_result {
+                Ok(actor) => break actor,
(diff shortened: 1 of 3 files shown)


linera-core/src/chain_worker/actor.rs
```diff
@@ -176,8 +176,45 @@ impl<StorageClient> ChainWorkerActor<StorageClient>
...
+                delivery_notifier.clone(),
+                chain_id,
+            )
+            .await;
+
+            match load_result {
+                Ok(actor) => break actor,
+                Err(error) => match incoming_requests.recv().await {
+                    Some(request) => request.send_error(error),
+                    None => return,
```
afck, line 208: Should we log the error here?

afck (approved):

linera-core/src/chain_worker/actor.rs
```diff
@@ -176,8 +176,45 @@ impl<StorageClient> ChainWorkerActor<StorageClient>
...
+                tracked_chains.clone(),
+                delivery_notifier.clone(),
+                chain_id,
+            )
+            .await;
+
+            match load_result {
+                Ok(actor) => break actor,
+                Err(error) => match incoming_requests.recv().await {
+                    Some(request) => request.send_error(error),
```
deuszx, line 207: What happens to failed requests?

linera-core/src/chain_worker/actor.rs
jvff: Yes indeed! Thanks for catching this. I added a `warn!` above, to log the error as soon as it is received, otherwise we would only log when the error is sent to a request, which would lose timing information.

linera-core/src/chain_worker/actor.rs
jvff: I believe it depends on the request, but most of them abort the operation :thinking:

deuszx: The current code will try to load a chain worker for every incoming request and fail them - one-by-one. I wonder whether the alternatives:
- failing all of them at once
- blocking until loaded
- panic(?)

were considered? Not being able to load a chain worker is a fatal error, right? It practically stalls the chain.

jvff: I didn't give those alternatives much thought because I figured that a storage error could be temporary (e.g., network issue to the database). The option to block until the chain state is loaded is tricky, because of the mixture between `async` code (to load it) and synchronous code (the endpoint cache).

PR #3302, opened 2025-02-12
Introduce the tombstone for `RocksDb`

MathieuDutSik:
## Motivation

The tombstone problem has emerged as a problem for `RocksDb`. We therefore introduce the test here for this store. We also do some corrections and simplification to the testing framework.

## Proposal

We propose the following changes:
* Add the testing of read keys from the store as test.
* Iterate the reading of the keys from the store to see the evolution of the performance.
* Introduce the test for RocksDb.

Besides, there are a number of changes for the tests:
* Simplification of the `MemoryContext::new_for_testing` function since the calls always took the same arguments.
* Elimination of requirement for `Debug` in the error type since it was not used.
* Put the creation of the testing store in the benchmark code.

## Test Plan

The CI is the object of those changes.

## Release Plan

Not relevant to TestNet / DevNet.

## Links

None.

### DIFF
--- a/linera-views/src/context.rs
+++ b/linera-views/src/context.rs
@@ -276,26 +276,19 @@ pub type MemoryContext<E> = ViewContext<E, MemoryStore>;
 /// Provides a `MemoryContext<()>` that can be used for tests.
 #[cfg(with_testing)]
 pub fn create_test_memory_context() -> MemoryContext<()> {
-    let namespace = crate::random::generate_test_namespace();
-    let root_key = &[];
-    MemoryContext::new_for_testing(
-        crate::memory::TEST_MEMORY_MAX_STREAM_QUERIES,
-        &namespace,
-        root_key,
-        (),
-    )
+    MemoryContext::new_for_testing(())
 }
 
 impl<E> MemoryContext<E> {
     /// Creates a [`Context`] instance in memory for testing.
     #[cfg(with_testing)]
     pub fn new_for_testing(
-        max_stream_queries: usize,
-        namespace: &str,
-        root_key: &[u8],
         extra: E,
     ) -> Self {
-        let store = MemoryStore::new_for_testing(max_stream_queries, namespace, root_key).unwrap();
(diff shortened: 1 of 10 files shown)


linera-views/src/test_utils/mod.rs
```diff
@@ -478,15 +497,41 @@ pub async fn tombstone_triggering_test<C: LocalRestrictedKeyValueStore>(key_valu
...
     run_test_batch_from_blank(&key_value_store, key_prefix.clone(), batch_insert).await;
+    println!("run_test_batch in {} ms", t1.elapsed().as_millis());
+
     // Deleting them all
+    let t1 = Instant::now();
     key_value_store.write_batch(batch_delete).await.unwrap();
-    // Reading everything and seeing that it is now cleaned.
-    let key_values = read_key_values_prefix(&key_value_store, &key_prefix).await;
-    assert_eq!(key_values, remaining_key_values);
+    println!("batch_delete in {} ms", t1.elapsed().as_millis());
```
ma2bd, line 513: We generally use `tracing` including in tests.

ma2bd (commented):
Can you fix the title of the PR? Afaict one would expect the word "test" or "testing" in the title.

ma2bd (changes requested):
back to your queue for the comment

linera-views/src/context.rs
```diff
@@ -276,19 +276,16 @@ pub type MemoryContext<E> = ViewContext<E, MemoryStore>;
 /// Provides a `MemoryContext<()>` that can be used for tests.
 #[cfg(with_testing)]
 pub fn create_test_memory_context() -> MemoryContext<()> {
```
ma2bd, line 278: Can we remove this function then?

MathieuDutSik: Yes, we could.
But the function occurs 393 times in the code, so at that size, maybe be prudent.
The replacement is 6 more characters.
If ok, I could replace it in another PR.

ma2bd: Next PR sounds good

ma2bd (approved):

PR #3306, opened 2025-02-12
Move `KeyPair` type to `ed25519.rs` and rename to `Ed25519*`.

deuszx:
## Motivation

We will be having more than Ed25519-only scheme of signatures.

## Proposal

To improve readability, move `KeyPair` type to `ed25519.rs` and rename to `Ed25519*`.

## Test Plan

N/A

## Release Plan

## Links


linera-base/src/crypto/ed25519.rs
```diff
@@ -4,21 +4,255 @@
 
 //! Defines ed25519 signature primitives used by the Linera protocol.
```
afck, line 5:
```suggestion
//! Defines Ed25519 signature primitives used by the Linera protocol.
```

afck (commented):
No blockers, but I'm wondering whether we should keep the algorithms themselves as almost an implementation detail inside the `crypto` module, and elsewhere:
* for now use type aliases like `KeyPair` and `Signature`, and
* later maybe use enums that support multiple algorithms?
(Not sure; just a thought.)

deuszx:
> No blockers, but I'm wondering whether we should keep the algorithms themselves as almost an implementation detail inside the `crypto` module, and elsewhere:
> 
> * for now use type aliases like `KeyPair` and `Signature`, and
> * later maybe use enums that support multiple algorithms?
>   (Not sure; just a thought.)

This might be a good idea. I've had to do a lot of rename-fixes in the follow-up PR where I tried to use secp256k1 instead of Ed25519. I'll see what I can do.

deuszx:
> No blockers, but I'm wondering whether we should keep the algorithms themselves as almost an implementation detail inside the `crypto` module, and elsewhere:
> 
> * for now use type aliases like `KeyPair` and `Signature`, and
> * later maybe use enums that support multiple algorithms?
>   (Not sure; just a thought.)

Ok, I re-exported the `Ed*` parts under generic names: `PublicKey`, `Signature` and `SigningKey`. Note that I decided to _not_ use `KeyPair` as previously it was used for `Ed25519PrivateKey` (which is not a `KeyPair` in a sense that it's not a public _and_ private key at the same time).

Issue #365, opened 2023-01-24
Replace broadcast channels by "event streams" so that producers don't need to accept subscriptions explicitly


ma2bd:
Event streams will be similar to streams of "events" in traditional blockchains.
* They will be indexed in storage to provide fast read-only access from other microchains.
* Applications that need to "tail" a stream in a transaction will have to declare the stream as a dependency so that block-producers (aka wallets) can subscribe (off-chain) to stream notifications and pass the last known index of a stream as transaction argument.

* [x] #2229
* [x] #2305
* [x] Add event reading oracle. (Contains whole event, not just ID.)
* [x] Add epoch change events.
* [x] Add `ProcessCommitteeChange(EventId)` operation.
* [x] Clients listen for new admin chain blocks, scan for epoch change events, propose that event in all their chains.
* [x] Remove committee change messages, admin chain subscriptions.
* [x] #2308
* [x] Add `subscribe_to_events` and `unsubscribe_from_events` contract runtime methods.
* [x] Add `UpdateStreams` system operation to update local event counts.
* [x] Add `Contract::EventValue` and make `emit` and `read_event` type safe.
* [x] Clients listen for new blocks on chains according to subscriptions, scan for events, submit `UpdateStreams` to subscriber chains.
* [x] Add `process_streams` contract endpoint.
* [x] Call `process_streams` on `UpdateStreams`, and whenever subscribers change.
* [x] Port `social` example to use events.
* [x] Remove pub-sub channels.
* **See sub-issues below.**

PR #3309, opened 2025-02-12
Move events and subscriptions out of `RawExecutionOutcome`.

afck:
## Motivation

The fact that events are in `RawExecutionOutcome` makes #365 more complicated. They also conceptually don't need to be in the structure whose purpose is mainly to associate messages with the correct application ID and authenticated signer.

## Proposal

Remove `events` from `RawExecutionOutcome`. Also remove `subscribe` and `unsubscribe`; the same argument applies to them.

## Test Plan

Refactoring; CI should catch regressions.

## Release Plan

## Links

- Related to #365

### DIFF
--- a/linera-base/src/data_types.rs
+++ b/linera-base/src/data_types.rs
@@ -19,7 +19,7 @@ use std::{
 };
 
 use anyhow::Context as _;
-use async_graphql::InputObject;
+use async_graphql::{InputObject, SimpleObject};
 use base64::engine::{general_purpose::STANDARD_NO_PAD, Engine as _};
 use custom_debug_derive::Debug;
 use linera_witty::{WitLoad, WitStore, WitType};
@@ -34,8 +34,8 @@ use crate::{
     crypto::{BcsHashable, CryptoHash},
     doc_scalar, hex_debug,
     identifiers::{
-        ApplicationId, BlobId, BlobType, BytecodeId, Destination, GenericApplicationId, MessageId,
-        UserApplicationId,
+        ApplicationId, BlobId, BlobType, BytecodeId, ChainId, Destination, EventId,
+        GenericApplicationId, MessageId, StreamId, UserApplicationId,
     },
     limited_writer::{LimitedWriter, LimitedWriterError},
     time::{Duration, SystemTime},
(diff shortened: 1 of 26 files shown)


linera-base/src/data_types.rs
```diff
@@ -1156,6 +1156,34 @@ impl<'a> Deserialize<'a> for Blob {
     }
 }
 
+/// An event recorded in an executed block.
+#[derive(Debug, PartialEq, Eq, Hash, Clone, Serialize, Deserialize, SimpleObject)]
+pub struct EventRecord {
```
deuszx, line 1161: Why not just `Event`?

afck: Right… the original reason (from the original PR) was that we already had an `Event` type with a different meaning. But now I can rename it…

linera-execution/src/transaction_tracker.rs
```diff
@@ -25,15 +25,36 @@ pub struct TransactionTracker {
...
+    pub oracle_responses: Vec<OracleResponse>,
+    #[debug(skip_if = Vec::is_empty)]
+    pub outcomes: Vec<ExecutionOutcome>,
+    pub next_message_index: u32,
+    /// Events recorded by contracts' `emit` calls.
+    pub events: Vec<EventRecord>,
+    /// Subscribe chains to channels.
+    pub subscribe: Vec<(ChannelFullName, ChainId)>,
+    /// Unsubscribe chains from channels.
+    pub unsubscribe: Vec<(ChannelFullName, ChainId)>,
```
deuszx, line 49: What happens if there are multiple actions to subscribe/unsubscribe for the same channel<>chainId? Will we try to "cancel them out" before handling?

afck: No, we just handle all of them.

These will go away anyway when we finally have event streams and remove pub-sub channels.

deuszx: It's wasteful to handle if they will cancel out each other... If you say we'll remove them then it doesn't matter though.

deuszx (approved):

PR #3310, opened 2025-02-12
Rename `EventRecord` → `Event`.

afck:
## Motivation

`EventRecord` has its name because back then there was still another `Event` type. That is not the case anymore.

## Proposal

Rename `EventRecord` to `Event`.

## Test Plan

Only renaming; CI would catch regressions.

## Release Plan

## Links


deuszx (approved):

Twey (approved):

alordk:
I wish you success in your work!

Issue #3300, opened 2025-02-12
Unexpected gRPC status: status: Internal, message: "h2 protocol error: error reading a body from connection"


kikakkz:
**Description**
We tested with machines in different areas (Hongkong, Frankfurt), and we always get this error. Sometimes node service stopped to get new notifications from validators. But I'm not sure it's caused by this error.

**Steps**
Till now I don't get any specific steps to reproduce the error. We just run node service, then it'll happen.

**Screenshot**
![Image](https://github.com/user-attachments/assets/72a4719c-d121-4b1c-8263-dcfa8b4fc3ea)

PR #3313, opened 2025-02-12
Include validator address in gRPC status log.

afck:
This is a port of #3312.

## Motivation

The gRPC status log currently does not show which validator connection it is about.

## Proposal

Include the validator address in the log.

## Test Plan

CI

## Release Plan

- Nothing to do.

## Links

- Related to #3300
- Testnet Archimedes version: #3312

### DIFF
--- a/linera-rpc/src/grpc/client.rs
+++ b/linera-rpc/src/grpc/client.rs
@@ -316,6 +316,7 @@ impl ValidatorNode for GrpcClient {
         })
         .flatten();
 
+        let span = tracing::info_span!("notification stream");
         // The stream of `Notification`s that inserts increasing delays after retriable errors, and
         // terminates after unexpected or fatal errors.
         let notification_stream = endlessly_retrying_notification_stream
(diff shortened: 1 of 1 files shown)


Twey:
Maybe we should _also_ instrument the future?

Twey (approved):

afck:
> Maybe we should also instrument the future?

That's probably a bigger issue: Should we consider this for all `async move` blocks?

linera-rpc/src/grpc/client.rs
```diff
@@ -330,9 +331,11 @@ impl ValidatorNode for GrpcClient {
                     retry_count = 0;
                     return future::Either::Left(future::ready(true));
                 };
+                let _enter_span = span.enter();
                 if !Self::is_retryable(status) || retry_count >= max_retries {
                     return future::Either::Left(future::ready(false));
                 }
+                drop(_enter_span);
```
ndr-ds, line 338: Would something like this would've also worked? 🤔
```suggestion
                if !Self::is_retryable(status).instrumented(span.clone()).inner() || retry_count >= max_retries {
                    return future::Either::Left(future::ready(false));
                }
```

ndr-ds: This is my attempt at that https://github.com/linera-io/linera-protocol/pull/3329

PR #3315, opened 2025-02-12
spelling error

Vitaliyr888:
modelization - modellization


afck:
Wiktionary says "modelization" is fine.

PR #3322, opened 2025-02-13
Add Secp256k1 keys

deuszx:
## Motivation

N/A

## Proposal

Adds `Secp256k1SecretKey` and `Secp256k1PublicKey` structs. Creates a static `SECP256K1` context for reuse, instead of initializing new context for every operation. Note that the same effect could be used by using `global-context` feature of the `secp256k1` crate but that is mutually exclusive with other features (like `alloc`) that we might want to use in the future.

## Test Plan

N/A

## Release Plan

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4369,6 +4369,7 @@ dependencies = [
  "is-terminal",
  "linera-base",
  "linera-witty",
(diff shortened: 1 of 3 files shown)


deuszx:
Replaced by #3345

PR #3324, opened 2025-02-13
Introduce --with-faucet for net up using the default chain

ndr-ds:
## Motivation

Right now we have a `--with-faucet-chain` option that both enables starting a faucet and also specifies what chain the faucet will be running on.
`linera faucet` however has a slightly different behavior, in which if you don't specify the chain, it just uses the default chain.

## Proposal

Introduce a `--with-faucet` option that controls whether a faucet will be started, which will by default start a faucet on the default chain, unless the `--faucet-chain` argument is provided.

## Test Plan

Run it locally, see the faucet being started using the default chain as expected, and previous behavior also works

## Release Plan

### DIFF
--- a/linera-client/src/client_options.rs
+++ b/linera-client/src/client_options.rs
@@ -1099,10 +1099,16 @@ pub enum NetCommand {
         #[arg(long, default_value = "grpc")]
         external_protocol: String,
 
-        /// If present, a faucet is started using the given chain root number (0 for the
-        /// admin chain, 1 for the first non-admin initial chain, etc).
+        /// If present, a faucet is started using the chain provided by --faucet-chain, or
+        /// `ChainId::root(1)` if not provided, as root 0 is usually the admin chain.
+        #[arg(long, default_value = "false")]
+        with_faucet: bool,
+
+        /// When using --with-faucet, this specifies the chain on which the faucet will be started.
+        /// The chain is specified by its root number (0 for the admin chain, 1 for the first
+        /// non-admin initial chain, etc).
         #[arg(long)]
-        with_faucet_chain: Option<u32>,
+        faucet_chain: Option<u32>,
 
         /// The port on which to run the faucet server
         #[arg(long, default_value = "8080")]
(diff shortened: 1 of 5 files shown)


linera-client/src/client_options.rs
```diff
@@ -1099,10 +1099,15 @@ pub enum NetCommand {
         #[arg(long, default_value = "grpc")]
         external_protocol: String,
 
+        /// If present, a faucet is started using the chain provided by --faucet-chain, or the
+        /// default chain if not provided.
```
ma2bd, line 1103: I don't get it. We just created a local test network and an admin wallet. The default chain is root chain 0.

ma2bd: Oh ok so you want `--faucet` instead of `--with-faucet-chain 0` but we're dealing with chain IDs here so the API was never going to be standard.

ma2bd: Also `0` (the admin chain) is not the best choice for the faucet chain.

ma2bd (changes requested):
I don't think `0` is a good default value.

linera-client/src/client_options.rs
ndr-ds: Ah, good point. Let me fix that.

linera-client/src/client_options.rs
```diff
@@ -1099,10 +1099,15 @@ pub enum NetCommand {
         #[arg(long, default_value = "grpc")]
         external_protocol: String,
 
+        /// If present, a faucet is started using the chain provided by --faucet-chain, or
+        /// `ChainId::root(1)` if not provided, as root 0 is usually the admin chain.
+        #[arg(long, default_value = "false")]
+        with_faucet: bool,
+
         /// If present, a faucet is started using the given chain root number (0 for the
         /// admin chain, 1 for the first non-admin initial chain, etc).
```
ma2bd, line 1108: outdated

ndr-ds: I had updated this but I guess I forgot to push? Not sure what happened. Let me push again

ma2bd (approved):

ndr-ds:
### Merge activity

* **Feb 17, 2:15 PM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/linera-io/linera-protocol/3324).

PR #3339, opened 2025-02-14
Correcting the Typo in `entrypoint2`

aso20455:
There’s a small typo in `entrypoint2`: replace `Ok(self.view1.get(&key).await?)` with `Ok(self.view2.get(&key).await?)` because `view2` is the MapView that requires a key.


afck (approved):
Good catch!

(We should have a test for this template, ideally.)

PR #3345, opened 2025-02-15
Add Secp256k1 keys

deuszx:
## Motivation

We want to support secp256k1 signatures - meaning we need public/pivate keys for them.

## Proposal

Adds `Secp256k1SecretKey` and `Secp256k1PublicKey` structs. Creates a static `SECP256K1` context for reuse, instead of initializing new context for every operation. Note that the same effect could be used by using `global-context` feature of the `secp256k1` crate but that is mutually exclusive with other features (like `alloc`) that we might want to use in the future.

## Test Plan

N/A

## Release Plan


linera-base/Cargo.toml
```diff
@@ -44,6 +44,7 @@ getrandom = { workspace = true, optional = true }
 hex.workspace = true
 is-terminal.workspace = true
 linera-witty = { workspace = true, features = ["macros"] }
+once_cell.workspace = true
```
afck, line 47: [`OnceCell` is in the standard library now.](https://doc.rust-lang.org/std/cell/struct.OnceCell.html)

linera-base/src/crypto/secp256k1.rs
```diff
@@ -4,39 +4,181 @@
...
+
 /// A secp256k1 signature.
 #[derive(Eq, PartialEq, Copy, Clone)]
 pub struct Secp256k1Signature(pub secp256k1::ecdsa::Signature);
 
+impl PartialEq for Secp256k1SecretKey {
+    fn eq(&self, other: &Self) -> bool {
+        self.0 == other.0
+    }
+}
```
afck, line 43: Isn't that the same as `derive(PartialEq)`?

linera-base/src/crypto/secp256k1.rs
```diff
@@ -4,39 +4,181 @@
...
+        }
+    }
+}
+
+impl FromStr for Secp256k1PublicKey {
+    type Err = CryptoError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let pk = secp256k1::PublicKey::from_str(s)
+            .map_err(|_| CryptoError::IncorrectPublicKeySize(0))?;
```
afck, line 92: It could also be invalid characters. Should we just wrap `secp256k1::Error`? (Also below.)

linera-base/src/crypto/secp256k1.rs
```diff
@@ -4,39 +4,181 @@
...
+        Ok(Secp256k1PublicKey(pk))
+    }
+}
+
+impl From<[u8; 33]> for Secp256k1PublicKey {
+    fn from(value: [u8; 33]) -> Self {
+        let pk = secp256k1::PublicKey::from_slice(&value).expect("Invalid public key");
```
afck, line 99: Is it guaranteed that this can't fail if the value is 33 bytes? Looking at the code I would say it's not.

Either way, I'd use `secp256k1::constants::PUBLIC_KEY_SIZE` instead of `33`.

linera-base/src/crypto/secp256k1.rs
```diff
@@ -4,39 +4,181 @@
...
+impl fmt::Display for Secp256k1PublicKey {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let s = hex::encode(self.0.serialize());
+        write!(f, "{}", s)
+    }
+}
+
+impl fmt::Debug for Secp256k1PublicKey {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", hex::encode(&self.0.serialize()[0..9]))
```
afck, line 123: What would `self.0.fmt(f)` look like? I'd expect that to also just be the key in hexadecimal?

linera-base/src/crypto/secp256k1.rs
```diff
@@ -4,39 +4,181 @@
...
+    }
+}
+
+impl Secp256k1KeyPair {
+    #[cfg(all(with_getrandom, with_testing))]
+    /// Generates a new key-pair.
```
afck, line 129:
```suggestion
    /// Generates a new key-pair.
    #[cfg(all(with_getrandom, with_testing))]
```

linera-base/src/crypto/secp256k1.rs
```diff
@@ -4,39 +4,181 @@
...
+    pub fn generate() -> Self {
+        let mut rng = rand::rngs::OsRng;
+        Self::generate_from(&mut rng)
+    }
+
+    #[cfg(with_getrandom)]
+    /// Generates a new key-pair from the given RNG. Use with care.
```
afck, line 136:
```suggestion
    /// Generates a new key-pair from the given RNG. Use with care.
    #[cfg(with_getrandom)]
```

linera-base/src/crypto/secp256k1.rs
```diff
@@ -4,39 +4,181 @@
...
+    }
+}
+
+impl Secp256k1PublicKey {
+    /// Returns a public key for the given secret key.
+    #[allow(dead_code)]
+    fn from_secret_key(secret: &Secp256k1SecretKey) -> Self {
+        secret.to_public()
+    }
+}
```
afck, line 159: Do we need these? Are we storing only the secret key to save space?

linera-base/src/crypto/secp256k1.rs
```diff
@@ -4,39 +4,181 @@
...
-    pub fn new<'de, T>(value: &T, secret: &secp256k1::SecretKey) -> Self
+    pub fn new<'de, T>(value: &T, secret: &Secp256k1SecretKey) -> Self
     where
         T: BcsSignable<'de>,
     {
-        let secp = secp256k1::Secp256k1::new();
+        let secp = secp256k1::Secp256k1::signing_only();
         let message = Message::from_digest(CryptoHash::new(value).as_bytes().0);
-        let signature = secp.sign_ecdsa(&message, secret);
+        let signature = secp.sign_ecdsa(&message, &secret.0);
```
afck, line 170: Why not use `SecretKey::sign_ecdsa` directly? You are using `SECP256K1` above, so I assume we require the `global-context` feature anyway?

linera-base/src/crypto/secp256k1.rs
```diff
@@ -4,39 +4,181 @@
...
-    pub fn check<'de, T>(&self, value: &T, author: &secp256k1::PublicKey) -> Result<(), CryptoError>
+    pub fn check<'de, T>(&self, value: &T, author: &Secp256k1PublicKey) -> Result<(), CryptoError>
     where
         T: BcsSignable<'de> + fmt::Debug,
     {
-        let secp = secp256k1::Secp256k1::new();
         let message = Message::from_digest(CryptoHash::new(value).as_bytes().0);
-        secp.verify_ecdsa(&message, &self.0, author)
+        SECP256K1
+            .verify_ecdsa(&message, &self.0, &author.0)
```
afck, line 181: There's also `PublicKey::verify`.

linera-base/src/crypto/secp256k1.rs
deuszx: That's "our" SECP256K1 and not from the crate. That feature (`global-context`) is mutually exclusive with some other features that we use.

linera-base/src/crypto/secp256k1.rs
deuszx: Yeah, you're right.

linera-base/src/crypto/secp256k1.rs
afck: Ah, right, makes sense!

linera-base/src/crypto/secp256k1.rs
afck: (No, because we have to use our own global context, right?)

deuszx: Yes. But this wil probably get some overhaul after the "JS compatibility" discussions.

linera-base/src/crypto/secp256k1.rs
deuszx: `from_secret_key` is not necessary - I'll remove it. The `fn to_public()` will be used (and renamed to `public()`) as it will be now used for validator keys.

linera-base/Cargo.toml
deuszx: That one isn't safe to use between threads and `static`s require that 🤔

deuszx: Ah, I forgot about this one (and I was reading about it just a week ago).

linera-base/src/crypto/secp256k1.rs
deuszx: Yeah, I was actually supposed to fix these errors. On it now.

linera-base/Cargo.toml
deuszx: Done in [Review part1](https://github.com/linera-io/linera-protocol/pull/3345/commits/f8ad395f83f175631105fba95a5a046aeb61fd41).

linera-base/src/crypto/secp256k1.rs
deuszx: Done in [Review part1](https://github.com/linera-io/linera-protocol/pull/3345/commits/f8ad395f83f175631105fba95a5a046aeb61fd41).

linera-base/src/crypto/secp256k1.rs
deuszx: Done in [Review part1](https://github.com/linera-io/linera-protocol/pull/3345/commits/f8ad395f83f175631105fba95a5a046aeb61fd41).

linera-base/src/crypto/secp256k1.rs
deuszx: Done in [Review part1](https://github.com/linera-io/linera-protocol/pull/3345/commits/f8ad395f83f175631105fba95a5a046aeb61fd41).

linera-base/src/crypto/secp256k1.rs
deuszx: Done in [Review part1](https://github.com/linera-io/linera-protocol/pull/3345/commits/f8ad395f83f175631105fba95a5a046aeb61fd41).

linera-base/src/crypto/secp256k1.rs
deuszx: Done in [Review part1](https://github.com/linera-io/linera-protocol/pull/3345/commits/f8ad395f83f175631105fba95a5a046aeb61fd41).

linera-base/src/crypto/secp256k1.rs
deuszx: Done in [Review part1](https://github.com/linera-io/linera-protocol/pull/3345/commits/f8ad395f83f175631105fba95a5a046aeb61fd41).

linera-base/src/crypto/secp256k1.rs
```diff
@@ -4,39 +4,163 @@
...
+
+/// A secp256k1 secret key.
+#[derive(Eq, PartialEq)]
+pub struct Secp256k1SecretKey(pub secp256k1::SecretKey);
+
+/// A secp256k1 public key.
+#[derive(Eq, PartialEq, Copy, Clone)]
+pub struct Secp256k1PublicKey(pub secp256k1::PublicKey);
+
+/// Secp256k1 public/private key pair.
```
afck, line 26:
```suggestion
/// Secp256k1 public/secret key pair.
```

linera-base/src/crypto/secp256k1.rs
```diff
@@ -4,39 +4,163 @@
...
+use std::{fmt, str::FromStr, sync::LazyLock};
 
-use secp256k1::{self, Message};
+use secp256k1::{self, All, Message, Secp256k1};
 use serde::{Deserialize, Serialize};
 
 use super::{BcsSignable, CryptoError, CryptoHash, HasTypeName};
 use crate::doc_scalar;
 
+/// Static Secp256k1 context for reuse.
```
afck, line 15:
```suggestion
/// Static secp256k1 context for reuse.
```

linera-base/src/crypto/secp256k1.rs
```diff
@@ -4,39 +4,163 @@
...
+}
+
+impl fmt::Debug for Secp256k1PublicKey {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", hex::encode(&self.0.serialize()[0..9]))
+    }
+}
+
+impl Secp256k1KeyPair {
+    /// Generates a new key-pair.
```
afck, line 118: I'd use "key pair" with a space everywhere. (Looks like we're already a bit inconsistent there.)

afck (approved):

Issue #1793, opened 2024-03-18
Use GRPC streams for `linera-storage-service`.


MathieuDutSik:
To deal with the limit of 4M (for incoming and outgoing requests), we proposed to implement some chunking mechanisms. The right solution is to use GRPC streams.
An additional benefit of such streams is that we could avoid the memory leak that comes from cancelled or terminated requests.

ma2bd:
> An additional benefit of such streams is that we could avoid the memory leak that comes from cancelled or terminated requests.

@MathieuDutSik What do you mean?

ma2bd:
Where does this 4M come from? I believe we changed that for the main GRPC service

PR #3346, opened 2025-02-16
Use gRPC streams for linera storage service.

usagi32:
addresses #1793

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4993,6 +4993,7 @@ dependencies = [
  "test-strategy",
  "thiserror 1.0.69",
  "tokio",
+ "tokio-stream",
  "tonic",
  "tonic-build",
  "tracing",
--- a/linera-storage-service/Cargo.toml
+++ b/linera-storage-service/Cargo.toml
@@ -19,7 +19,7 @@ test = ["linera-views/test"]
 
 [[bin]]
 name = "linera-storage-server"
-path = "src/server.rs"
+path = "src/main.rs"
 
 [dependencies]
 anyhow.workspace = true
@@ -34,6 +34,7 @@ prost.workspace = true
 serde.workspace = true
 thiserror.workspace = true
 tokio = { workspace = true, features = ["full"] }
+tokio-stream.workspace = true
 tonic = { workspace = true, features = ["codegen", "prost", "transport"] }
 tracing.workspace = true
 tracing-subscriber = { workspace = true, features = ["fmt"] }
--- a/linera-storage-service/proto/key_value_store.proto
+++ b/linera-storage-service/proto/key_value_store.proto
@@ -3,25 +3,6 @@ package key_value_store.v1;
 
 import "google/protobuf/empty.proto";
 
-message KeyValue {
-  bytes key = 1;
-  bytes value = 2;
-}
-
-message KeyValueAppend {
-  bytes key = 1;
-  bytes value = 2;
-  bool last = 3;
-}
-
-message Statement {
-  oneof Operation {
-    bytes delete = 1;
-    KeyValue put = 2;
-    KeyValueAppend append = 3;
-    bytes delete_prefix = 4;
-  }
-}
 
 message OptValue {
   optional bytes value = 1;
@@ -34,8 +15,6 @@ message RequestReadValue {
 
 message ReplyReadValue {
   optional bytes value = 1;
-  int64 message_index = 2;
-  int32 num_chunks = 3;
 }
 
 
@@ -53,7 +32,7 @@ message RequestContainsKeys {
 }
 
 message ReplyContainsKeys {
-  repeated bool tests = 1;
+  repeated bool tests = 1 [packed = true];
 }
 
 
@@ -63,8 +42,6 @@ message RequestReadMultiValues {
 
 message ReplyReadMultiValues {
   repeated OptValue values = 1;
-  int64 message_index = 2;
-  int32 num_chunks = 3;
 }
 
 
@@ -74,8 +51,6 @@ message RequestFindKeysByPrefix {
 
 message ReplyFindKeysByPrefix {
   repeated bytes keys = 1;
-  int64 message_index = 2;
-  int32 num_chunks = 3;
 }
 
 
@@ -84,24 +59,12 @@ message RequestFindKeyValuesByPrefix {
 }
 
 message ReplyFindKeyValuesByPrefix {
-  repeated KeyValue key_values = 1;
-  int64 message_index = 2;
-  int32 num_chunks = 3;
+  bytes key_values = 1;
 }
 
 
-message RequestWriteBatchExtended {
-  repeated Statement statements = 1;
-}
-
-
-message RequestSpecificChunk {
-  int64 message_index = 1;
-  int32 index = 2;
-}
-
-message ReplySpecificChunk {
-  bytes chunk = 1;
+message RequestWriteBatch {
+  bytes stream = 1;
 }
 
 
@@ -139,18 +102,17 @@ message ReplyListRootKeys {
 
 
 service StoreProcessor {
-  rpc ProcessReadValue (RequestReadValue) returns (ReplyReadValue) {}
+  rpc ProcessReadValue (RequestReadValue) returns (stream ReplyReadValue) {}
   rpc ProcessContainsKey (RequestContainsKey) returns (ReplyContainsKey) {}
-  rpc ProcessContainsKeys (RequestContainsKeys) returns (ReplyContainsKeys) {}
-  rpc ProcessReadMultiValues (RequestReadMultiValues) returns (ReplyReadMultiValues) {}
-  rpc ProcessFindKeysByPrefix (RequestFindKeysByPrefix) returns (ReplyFindKeysByPrefix) {}
-  rpc ProcessFindKeyValuesByPrefix (RequestFindKeyValuesByPrefix) returns (ReplyFindKeyValuesByPrefix) {}
-  rpc ProcessWriteBatchExtended (RequestWriteBatchExtended) returns (google.protobuf.Empty) {}
-  rpc ProcessSpecificChunk (RequestSpecificChunk) returns (ReplySpecificChunk) {}
+  rpc ProcessContainsKeys (stream RequestContainsKeys) returns (ReplyContainsKeys) {}
+  rpc ProcessReadMultiValues (stream RequestReadMultiValues) returns (stream ReplyReadMultiValues) {}
+  rpc ProcessFindKeysByPrefix (RequestFindKeysByPrefix) returns (stream ReplyFindKeysByPrefix) {}
+  rpc ProcessFindKeyValuesByPrefix (RequestFindKeyValuesByPrefix) returns (stream ReplyFindKeyValuesByPrefix) {}
+  rpc ProcessWriteBatch (stream RequestWriteBatch) returns (google.protobuf.Empty) {}
   rpc ProcessCreateNamespace (RequestCreateNamespace) returns (google.protobuf.Empty) {}
   rpc ProcessExistsNamespace (RequestExistsNamespace) returns (ReplyExistsNamespace) {}
   rpc ProcessDeleteNamespace (RequestDeleteNamespace) returns (google.protobuf.Empty) {}
-  rpc ProcessListAll (google.protobuf.Empty) returns (ReplyListAll) {}
+  rpc ProcessListAll (google.protobuf.Empty) returns (stream ReplyListAll) {}
   rpc ProcessListRootKeys (RequestListRootKeys) returns (ReplyListRootKeys) {}
   rpc ProcessDeleteAll (google.protobuf.Empty) returns (google.protobuf.Empty) {}
 }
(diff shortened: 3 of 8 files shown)


usagi32:
@ma2bd

linera-views/src/test_utils/mod.rs
```diff
@@ -286,8 +286,10 @@ pub fn get_random_test_scenarios() -> Vec<Vec<(Vec<u8>, Vec<u8>)>> {
         get_random_key_values1(7, 3),
         get_random_key_values1(150, 3),
         get_random_key_values1(30, 10),
+        get_random_key_values1(10, 4_500_000),
         get_random_key_values2(30, 4, 10),
         get_random_key_values2(30, 4, 100),
+        //get_random_key_values2(10, 200, 3_999_700),
```
afck, line 292: Does that value have a particular significance?
Anyway, let's either remove the line or the `//`.

linera-storage-service/Cargo.toml
```diff
@@ -19,7 +19,7 @@ test = ["linera-views/test"]
 
 [[bin]]
 name = "linera-storage-server"
-path = "src/server.rs"
+path = "src/main.rs"
```
afck, line 22: Why?

linera-storage-service/proto/key_value_store.proto
```diff
@@ -53,7 +32,7 @@ message RequestContainsKeys {
 }
 
 message ReplyContainsKeys {
-  repeated bool tests = 1;
+  repeated bool tests = 1 [packed = true];
```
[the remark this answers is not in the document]
usagi32: Yes, its in proto3, it was not in proto2.

linera-storage-service/src/client.rs
```diff
@@ -98,17 +97,18 @@ impl ReadableKeyValueStore for ServiceStoreClientInternal {
...
-            self.read_entries(message_index, num_chunks).await
+        let mut incoming_stream = response.into_inner();
+
+        let mut reformed_value = Vec::new();
+        while let Some(ReplyReadValue { value }) = incoming_stream.message().await? {
+            if value.is_none() {
+                return Ok(None);
+            }
+
+            reformed_value.append(&mut value.unwrap());
```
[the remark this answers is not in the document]
usagi32: Append takes an exclusive reference? 
No, its fine. None is actually part of the message optional field. If key is not found None is send and we return from it.

linera-storage-service/src/client.rs
```diff
@@ -134,8 +134,28 @@ impl ReadableKeyValueStore for ServiceStoreClientInternal {
...
         }
-        let query = RequestContainsKeys { keys: full_keys };
-        let request = tonic::Request::new(query);
+
+        let mut queries = Vec::new();
+        let mut grouped_keys = Vec::new();
+        let mut chunk_size = 0;
+        for full_key in full_keys {
+            if MAX_PAYLOAD_SIZE >= full_key.len() + chunk_size {
+                chunk_size += full_key.len();
```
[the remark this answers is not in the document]
usagi32: Actually, gRPC proto encodes a little more than even the length.
There were two extreme cases
- keys are closer to MAX_KEY_SIZE, then there is a lot more than enough room in MAX_PAYLOAD_SIZE.
- keys are small like 10 bytes, then there will be like hundreds of thousands of keys to test before the limit, even in synthetic tests we are not doing that.

Both are unrealistic cases, and there will performance and efficiency tradeoffs. Extra encodings are also variable for each element as non primitive types are not packed together. Its like tag + length + data where tag can be anything from 1 to 5 bytes depending on the field number and same for length.
Just a simple buffer of like 10-20 bytes will dominate the size of very small keys.

Let me know.

linera-storage-service/src/client.rs
```diff
@@ -156,24 +176,75 @@ impl ReadableKeyValueStore for ServiceStoreClientInternal {
...
+            }
+
+            if !outgoing_message.is_empty() {
+                let _ = tx.send(RequestReadMultiValues {
+                    keys: outgoing_message,
+                });
+            }
+        };
+
+        let _ = std::thread::spawn(spawner);
```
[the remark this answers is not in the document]
usagi32: We have to get to the receiver at the end of functions body to take the advantages of streaming for switching the compute bound and IO bound operations. Preprocessing should be avoided as possible.
I just did that as there was no future to be awaited in the closure.
It just hit me that the receiver will likely be awaited by the tonic, we can just do this like a normal future.

linera-views/src/test_utils/mod.rs
usagi32: Large key lengths or number of keys are exponentially slow in the tests. Likely due to the way `run_reads` is designed.
Just left this for the thoughts.

linera-storage-service/Cargo.toml
usagi32: It was quite handy for me during the testing. We get a `server` library function we can spawn any time to perform tests using `#[tokio::test]` without building everything.

linera-storage-service/src/client.rs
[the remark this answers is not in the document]
afck: We'd just like to avoid `unwrap` wherever it's easily possible. `if let else` works just as well here, without the `unwrap`.

Right, the `append` could probably be replaced with `extend`.

linera-storage-service/src/client.rs
```diff
@@ -156,24 +176,75 @@ impl ReadableKeyValueStore for ServiceStoreClientInternal {
...
+                            num_messages = usize::from_be_bytes(bytes.try_into().unwrap());
+                            continue;
+                        }
+
+                        value.append(&mut bytes);
+                        num_messages -= 1;
+
+                        if 0 == num_messages {
+                            values.push(Some(value));
+                            value = Vec::new();
```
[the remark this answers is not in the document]
afck: I mean after the `while`, inside the outer loop's body: That would remove the need to reset it. I'd also find it a bit more readable, because it conceptually tracks the progress of the inner loop.

linera-storage-service/src/client.rs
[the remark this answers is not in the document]
usagi32: > We'd just like to avoid `unwrap` wherever it's easily possible. `if let else` works just as well here, without the `unwrap`.

Actually, was talking about the need to log the None. It's fine and expected.

> Right, the `append` could probably be replaced with `extend`.

That would be quite bad, `extend` will iterate move things on the stack for each element, `append` moves things in bulk in-place on the heap. It will be like O(1) vs O(n), where n is at least 6 orders of magnitude greater. Sometimes the compiler inlines and optimizes it away so both are same, but let's not rely on that.

linera-storage-service/src/client.rs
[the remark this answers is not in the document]
usagi32: Actually, it needs to be persisted across the while loop. You may refer to the server module.

linera-views/src/test_utils/mod.rs
MathieuDutSik: No, you should remove the // and add a comment to the `run_reads` function complexity. Indeed it is slow.
A test does not need to be super fast, but it should be testing useful functionality.

The code should not be cryptic and if there is an underlying reason that takes more than 5 minutes to understand then it should be written clearly.

linera-storage-service/src/client.rs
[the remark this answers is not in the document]
MathieuDutSik: The standard size is 4M for gRPC. But the MAX_PAYLOAD_SIZE if 4000000 which leaves room for additional small stuff to be put.

linera-storage-service/src/server.rs
```diff
@@ -310,181 +329,365 @@ impl StoreProcessor for ServiceStoreServer {
...
+
+        let mut state = Vec::new();
+        let mut key = Vec::new();
+        let mut value = Vec::new();
+        let mut len_bytes = Vec::with_capacity(8);
+        while let Some(RequestWriteBatch { stream: message }) = incoming_stream.message().await? {
+            let mut message = VecDeque::from(message);
+            while !message.is_empty() || state.get(2).is_some_and(|len| 0 == *len) {
+                match state.len() {
+                    0 => {
```
MathieuDutSik, line 609: Instead of 0, 1, 2, 3, an enum indicating the significance of the choice being made should be used.

linera-storage-service/src/server.rs
```diff
@@ -310,181 +329,365 @@ impl StoreProcessor for ServiceStoreServer {
...
+                                .drain(..(state[2] - key.len()).min(message.len()))
+                                .collect::<Vec<_>>(),
+                        );
+                        match (state[0], state[2] == key.len()) {
+                            (1, true) => {
+                                batch.delete_key(key);
+                                key = Vec::new();
+                                let _ = state.pop();
+                                let _ = state.pop();
+                                let _ = state.pop();
```
MathieuDutSik, line 655: It is a little unfortunate to have the 3 pops. But the alternativer with `split_off` or `drain` is not much simpler.

MathieuDutSik (commented):
It is a nice progress but a little cryptic.

linera-storage-service/src/client.rs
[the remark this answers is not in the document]
usagi32: It should be fine for the most part. Although, it can be broken if one tries intentionally, but I'm sure never in any real-life, practical scenario.
I'll do some few tweaks if I ever feel otherwise.

linera-views/src/test_utils/mod.rs
usagi32:
```suggestion
        get_random_key_values2(10, 200, 3_999_700),   // slow
```

linera-storage-service/src/server.rs
```diff
@@ -310,181 +329,365 @@ impl StoreProcessor for ServiceStoreServer {
...
-        Ok(Response::new(response))
+
+        let _ = std::thread::spawn(spawner);
+
+        Ok(Response::new(UnboundedReceiverStream::new(rx)))
     }
 
-    #[instrument(target = "store_server", skip_all, err, fields(n_statements = ?request.get_ref().statements.len()))]
-    async fn process_write_batch_extended(
+    #[instrument(target = "store_server", skip_all, err)]
```
usagi32, line 593:
```suggestion
    // Encoding for this takes place in a pattern that mimics bcs serialization.
    // just using simple numbers for Operation's 3 variants.
    // simple length prefixed vectors.
    // and a stack state based depth first de-serialization.
    #[instrument(target = "store_server", skip_all, err)]
```

linera-views/src/test_utils/mod.rs
```diff
@@ -286,8 +286,10 @@ pub fn get_random_test_scenarios() -> Vec<Vec<(Vec<u8>, Vec<u8>)>> {
         get_random_key_values1(7, 3),
         get_random_key_values1(150, 3),
         get_random_key_values1(30, 10),
+        get_random_key_values1(10, 4_500_000),
         get_random_key_values2(30, 4, 10),
         get_random_key_values2(30, 4, 100),
+        get_random_key_values2(10, 200, 3_999_700),   // slow
```
usagi32, line 292:
```suggestion
        //get_random_key_values2(10, 200, 3_999_700), 
```

linera-views/src/test_utils/mod.rs
```diff
@@ -286,8 +286,10 @@ pub fn get_random_test_scenarios() -> Vec<Vec<(Vec<u8>, Vec<u8>)>> {
         get_random_key_values1(7, 3),
         get_random_key_values1(150, 3),
         get_random_key_values1(30, 10),
+        get_random_key_values1(10, 4_500_000),
         get_random_key_values2(30, 4, 10),
         get_random_key_values2(30, 4, 100),
+        //get_random_key_values2(10, 200, 3_999_700),
```
usagi32, line 292:
```suggestion
        get_random_key_values2(10, 200, 3_999_700),
```

linera-views/src/test_utils/mod.rs
usagi32: It's timing out for web tests in store_tests.rs as limit is 300. I guess `run_reads` is not too good with large or many keys. I am commenting it out to revisit in future.

linera-views/src/test_utils/mod.rs
```diff
@@ -286,8 +286,10 @@ pub fn get_random_test_scenarios() -> Vec<Vec<(Vec<u8>, Vec<u8>)>> {
         get_random_key_values1(7, 3),
         get_random_key_values1(150, 3),
         get_random_key_values1(30, 10),
+        get_random_key_values1(10, 4_500_000),
         get_random_key_values2(30, 4, 10),
         get_random_key_values2(30, 4, 100),
+        get_random_key_values2(10, 200, 3_999_700),
```
usagi32, line 292:
```suggestion
        // get_random_key_values2(10, 200, 3_999_700), slow may re-visit in future.
```

rannlangel (commented):
Adreno_gpu

Cargo.lock
YoungWhizz, line 1: nice

ma2bd (changes requested):
Let's pause this

PR #3347, opened 2025-02-16
Make `query_service` contract-only.

usagi32:
Removed query service from the BaseRuntime and moved it to the ContractRuntime as the ServiceRuntime doesn't need it.


usagi32:
@ma2bd

afck (approved):
Good point, yes! That only really makes sense in the contract.

afck:
Could you apply `cargo fmt`? The test failed.

Issue #1416, opened 2023-12-22
Support parallel reads in synchronous runtime


ma2bd:
# Motivation

After #1398, we lose some concurrency on the storage reads.

# Solution (tentatively)

* Add a configuration option to specify how many parallel reads are allowed
* Wrap `&mut ExecutionStateView` in a `RwLock` while the execution state actor is running and implement `ExecutionStateView::handle_request` on `this: &RwLock<&mut Self>` instead of `&mut self`.
* use `try_for_each_concurrent` to run the event handling instead of a for loop.

Issue #2037, opened 2024-05-15
Support composed operations (aka scripts)


ma2bd:
See also #1200

ma2bd:
Example of design: https://docs.sui.io/concepts/transactions/prog-txn-blocks

PR #3348, opened 2025-02-16
Support concurrent I/O.

usagi32:
addresses #1416

### DIFF
--- a/linera-execution/src/execution.rs
+++ b/linera-execution/src/execution.rs
@@ -6,7 +6,7 @@ use std::{
     mem, vec,
 };
 
-use futures::{stream::FuturesOrdered, FutureExt, StreamExt, TryStreamExt};
+use futures::{stream::FuturesOrdered, StreamExt, TryStreamExt};
 use linera_base::{
     data_types::{Amount, BlockHeight, Timestamp},
     identifiers::{Account, AccountOwner, ChainId, Destination, Owner},
@@ -238,9 +238,7 @@ where
 
         contract_runtime_task.send(code)?;
 
-        while let Some(request) = execution_state_receiver.next().await {
-            self.handle_request(request).await?;
-        }
+        let _ = self.handle_requests(&mut execution_state_receiver).await?;
 
         let (controller, txn_tracker_moved) = contract_runtime_task.join().await?;
         *txn_tracker = txn_tracker_moved;
@@ -540,9 +538,7 @@ where
 
         service_runtime_task.send(code)?;
 
-        while let Some(request) = execution_state_receiver.next().await {
-            self.handle_request(request).await?;
-        }
+        let _ = self.handle_requests(&mut execution_state_receiver).await?;
 
         service_runtime_task.join().await
     }
@@ -557,8 +553,7 @@ where
         >,
         runtime_request_sender: &mut std::sync::mpsc::Sender<ServiceRuntimeRequest>,
     ) -> Result<QueryOutcome<Vec<u8>>, ExecutionError> {
-        let (outcome_sender, outcome_receiver) = oneshot::channel();
-        let mut outcome_receiver = outcome_receiver.fuse();
+        let (outcome_sender, mut outcome_receiver) = oneshot::channel();
 
         runtime_request_sender
             .send(ServiceRuntimeRequest::Query {
@@ -569,16 +564,12 @@ where
             })
             .expect("Service runtime thread should only stop when `request_sender` is dropped");
 
+        // manual fusing.
+        let mut is_done = false;
         loop {
-            futures::select! {
-                maybe_request = incoming_execution_requests.next() => {
-                    if let Some(request) = maybe_request {
-                        self.handle_request(request).await?;
-                    }
-                }
-                outcome = &mut outcome_receiver => {
-                    return outcome.map_err(|_| ExecutionError::MissingRuntimeResponse)?;
-                }
+            tokio::select! {
+                result = self.handle_requests(incoming_execution_requests), if !is_done => { is_done = true; let _ = result?; },
+                outcome = &mut outcome_receiver => { return outcome.map_err(|_| ExecutionError::MissingRuntimeResponse)?; }
             }
         }
     }
--- a/linera-execution/src/execution_state_actor.rs
+++ b/linera-execution/src/execution_state_actor.rs
@@ -7,7 +7,7 @@
 use std::sync::LazyLock;
 
 use custom_debug_derive::Debug;
-use futures::channel::mpsc;
+use futures::{channel::mpsc::{self, UnboundedReceiver}, StreamExt};
 #[cfg(with_metrics)]
 use linera_base::prometheus_util::{bucket_latencies, register_histogram_vec, MeasureLatency as _};
 use linera_base::{
@@ -21,6 +21,7 @@ use oneshot::Sender;
 #[cfg(with_metrics)]
 use prometheus::HistogramVec;
 use reqwest::{header::CONTENT_TYPE, Client};
+use tokio::sync::RwLock;
 
 use crate::{
     system::{CreateApplicationResult, OpenChainConfig, Recipient},
@@ -60,7 +61,7 @@ where
     C::Extra: ExecutionRuntimeContext,
 {
     pub(crate) async fn load_contract(
-        &mut self,
+        &self,
         id: UserApplicationId,
     ) -> Result<(UserContractCode, UserApplicationDescription), ExecutionError> {
         #[cfg(with_metrics)]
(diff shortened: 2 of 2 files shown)


usagi32:
@ma2bd

afck (changes requested):
I believe the requests are actually _sent_ sequentially currently, so this wouldn't change anything.

If they were sent in parallel, we would need to make sure that all changes to the execution state are made in a deterministic order that doesn't depend on the timing of I/O.

usagi32:
Yes, it was also odd to me as I was trying to go along with issue description.

We have two blocking problems for it,
- The channel is created in the same function (`run_user_action_with_runtime`) in which `handle_request `is called. Reference of UnboundReceiver should be passed instead.
- There will be only one mutable reference of execution state view.

But this will only benefit if multiple runtime instances are simultaneously accessing the same execution state view. But I guess executions of transactions are happening sequentially?

afck:
Yes, exactly, that's also sequential. And even if they weren't, we'd again have to be very careful to keep execution deterministic. :thinking:

usagi32:
Let me ask this, is sequential execution of transactions required? There is this #2037 means each transaction was proposed keeping in mind the execution state of the last block, and therefore should not modify the state for other transactions in the same block.

I guess we have to be clear about how we want things to behave.

afck:
It's not planned right now to execute transactions in parallel. #2037's motivation is that the "result of the first operation is passed to the second operation", so that, at least, would have to be sequential.

For #1416 we should just try to make reads from the application state parallel where possible. I think I was actually wrong that this approach doesn't address that, sorry! But let's avoid `unsafe`.

usagi32:
But isn't each individual transaction proposed with the regards to the parent block execution state, as transactions likely don't know what outcomes they will produce, and past transactions may also invalidate the future ones. So state should not change 
in-between processing successive transactions in a block.

I understand what #2037 is about, but what I meant that its the only specific case that would require the changed state to persist in-between operations, as its currently done.

And isn't that unsafe safe because we know what we are doing? Maybe something with FuturesUnordered or LocalSet can be done, but we don't know the number of futures, its dynamic.

afck:
> isn't each individual transaction proposed with the regards to the parent block execution state

No, it's applied to whatever the execution state looks like after the previous transaction. It's the same in most blockchains, isn't it? Anyway, in Linera this is even less of an issue than on other platforms, because it's always a single client proposing the whole block, rather than adding a transaction to a mempool.

> And isn't that unsafe safe because we know what we are doing?

(Famous last words. :sweat_smile:)
Sure, but we'd just like to minimize, in general, the number of `unsafe` and `unwrap` and `expect`s etc. wherever possible.

usagi32:
> isn't each individual transaction proposed with the regards to the parent block execution state

I meant according to the client with non executed block, who doesn't know transaction outcome. Its "supposed" to be like this.

Like bitcoin and ethereum are free to alter index of their transaction and disregard state created by other transactions. Bitcoin keeps the utxo set commutative while ethereum doesn't alter the nonce for the same account. So its fine for them.

I guess this should be fine, if the client uses a proper wallet which will track the state changes by each transaction. It should not be a problem unless the client manually forms the block raw.

I'll try to do something about the unsafe thing. The issue is number of futures is dynamic, so a JoinSet was a perfect candidate which is requires 'static. Both FuturesUnorderd and LocalSet are done with fixed number of futures as you have to manually poll or run them again. 
It should be possible though.

linera-execution/src/execution.rs
```diff
@@ -569,16 +564,12 @@ where
             })
             .expect("Service runtime thread should only stop when `request_sender` is dropped");
 
+        // manual fusing.
```
jvff, line 567: I think we could simplify this and remove the `loop`, if we `use std::pin::pin;` and `use futures::future::{self, Either};`:

```rust
let request_handler = pin!(self.handle_requests(incoming_execution_requests));

let outcome = match future::select(outcome_receiver, request_handler).await {
    Either::Left((outcome, _)) => outcome,
    Either::Right((result, outcome_receiver)) => {
        result?;
        outcome_receiver.await
    }
};

outcome.map_err(|_| ExecutionError::MissingRuntimeResponse)?
```

linera-execution/src/execution_state_actor.rs
```diff
@@ -7,7 +7,7 @@
 use std::sync::LazyLock;
 
 use custom_debug_derive::Debug;
-use futures::channel::mpsc;
+use futures::{channel::mpsc::{self, UnboundedReceiver}, StreamExt};
```
jvff, line 10: You'll need to fix the formatting here for CI to pass:

```suggestion
use futures::{
    channel::mpsc::{self, UnboundedReceiver},
    StreamExt,
};
```

linera-execution/src/execution_state_actor.rs
```diff
@@ -360,6 +362,400 @@ where
...
+    ) -> Result<(), ExecutionError> {
+        use ExecutionRequest::*;
+
+        let request_handle = |request, this: &'static RwLock<&'static mut Self>| async {
+            match request {
+                #[cfg(not(web))]
+                LoadContract { id, callback } => {
+                    let actor = this.read().await;
+                    let  response = actor.load_contract(id).await?;
+                    drop(actor);
```
jvff, line 378: There's a pattern here that we can simplify and remove the `drop` by doing:

```suggestion
                    let response = this.read().await.load_contract(id).await?;
```

linera-execution/src/execution_state_actor.rs
```diff
@@ -89,6 +90,7 @@ where
         Ok((code, description))
     }
 
+    #[allow(dead_code)]
```
jvff, line 93: It would be best if we could update this method with the contents of the closure in `handle_requests` instead of keeping the old code.

linera-execution/src/execution_state_actor.rs
```diff
@@ -360,6 +362,400 @@ where
...
+                    write_actor.system.assert_blob_exists(blob_id).await?;
+                    callback.respond(write_actor.system.blob_used(None, blob_id).await?)
+                }
+            }
+
+            Ok(())
+        };
+
+        // alternative is to use FuturesUnordered
+        let self_transmuted: &'static mut ExecutionStateView<C> = unsafe { std::mem::transmute(self) };
```
jvff, line 722: We try to avoid `unsafe` as much as possible. Maybe `FuturesUnordered` (without spawning Tokio tasks) would work. I need to think a bit more about this :thinking:

afck:
> I meant according to the client with non executed block, who doesn't know transaction outcome.

OK, you're right, it _is_ different than in e.g. Ethereum: There, the client cares about their account, and uses a nonce.

In Linera, the block height of your user chain has a similar function to the nonce; your client tracks your entire chain and proposes whole blocks.

usagi32:
> > I meant according to the client with non executed block, who doesn't know transaction outcome.
> 
> OK, you're right, it _is_ different than in e.g. Ethereum: There, the client cares about their account, and uses a nonce.
> 
> In Linera, the block height of your user chain has a similar function to the nonce; your client tracks your entire chain and proposes whole blocks.

Yes, this should not be a problem at least in the single owner chains as long as the client uses proper state tracking and does not proposes a block raw block by hand.

For other types of chains this can be looked upon to achieve some kind of deterministic parallelism. Where some kinds of transactions are required to be serial like in #2037, and others can parallelized.

Let's see.

usagi32:
@jvff can you review it now? I had did it last week only also requested a review from you.

linera-execution/src/execution_state_actor.rs
```diff
@@ -262,30 +300,40 @@ where
...
+
                 callback.respond(messages)
             }
 
             CloseChain {
                 application_id,
                 callback,
             } => {
-                let app_permissions = self.system.application_permissions.get();
+                let mut write_actor = this.write().await;
```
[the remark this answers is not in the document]
Twey: The one-use variable is unnecessary actually: this block can actually be nested directly inside `respond` with no extra indent :)

I don't think it's too important to manually drop the guard since the send is an instantaneous sync call and it will be dropped immediately after. Maybe in the multi-threaded case it will avoid one block if the receiver immediately tries to take the lock?

linera-execution/src/execution_state_actor.rs
```diff
@@ -141,9 +161,10 @@ where
                     )
                     .await?;
```
Twey, line 162: I'm very scared about holding the `RwLock` guard over an await point. This means that (now as part of the external API of this module) the caller must know all the  possible holders of the lock and make sure to only poll them all at once, since otherwise they will deadlock when taking the lock.

PR #3349, opened 2025-02-16
Global state for non-fungible tokens.

usagi32:
## Motivation
Global state of linera is difficult to realize. That's because its quite an inherent drawback of the the linera's microchain nature and
about the actor model like systems altogether.
For example, currently, there is no standard definition of what a non fungible token is unlike other blockchains. There should be a common behaviour standard, functionality and uniqueness of all kinds tokens in linera like Ethereum's ERC 721 or Stacks blockchain's SIP009 or SIP010 for fungibles.
Also, currently, anyone is able to mint a same token twice without any consequences, as there can be same contract bytcodes with same behaviours on different chains, further decreasing the ability to canonicalize the originality. 
Solana solves this problem quite beautifully, with its separate and unique token accounts, metadata accounts and mint accounts.
But that's more of a workaround, rather than a solution, to their stateless nature smart contracts. And that would be overkill for linera.

## Todo
- Implement the standard for fungible tokens.
- Discuss for the semi fungible tokens.

### DIFF
--- a/linera-base/src/identifiers.rs
+++ b/linera-base/src/identifiers.rs
@@ -100,6 +100,38 @@ impl FromStr for Account {
     }
 }
 
+/// The mint associated with this asset.
+#[derive(Eq, PartialEq, Ord, PartialOrd, Debug, Copy, Clone, Serialize, Deserialize, WitLoad, WitStore, WitType)]
+pub struct NftMint(pub CryptoHash);
+
+/// Handy elegant Alias
+pub type Mint = NftMint;
+
+/// Necessary minimal data associated with a minted asset.
+#[derive(Eq, PartialEq, Ord, PartialOrd, Debug, Clone, Serialize, Deserialize, WitLoad, WitStore, WitType)]
+pub struct NftMetadata {
+    /// The mint associated with this asset.
+    pub mint: Mint,
+    /// Link to the asset's data and metadata json file.
+    pub uri: String,
+    /// A name for the asset.
+    pub name: String,
+    /// A samll description of the asset.
+    pub description: String,
+}
+
+/// Handy elegant Alias
+pub type Metadata = NftMetadata;
+
+/// Just for convinience
+pub fn hash_bytes(bytes: &[u8]) -> CryptoHash {
+    use sha3::digest::Digest;
+
+    let mut hasher = sha3::Sha3_256::new();
+    hasher.update(bytes);
+    hasher.finalize()[..].try_into().unwrap()
+}
+
 /// How to create a chain.
 #[derive(Eq, PartialEq, Ord, PartialOrd, Copy, Clone, Hash, Debug, Serialize, Deserialize)]
 pub enum ChainDescription {
--- a/linera-chain/src/chain.rs
+++ b/linera-chain/src/chain.rs
@@ -23,6 +23,7 @@ use linera_base::{
     ownership::ChainOwnership,
 };
 use linera_execution::{
+    global_state::GlobalContext,
     committee::{Committee, Epoch, ValidatorName},
     system::OpenChainConfig,
     ExecutionOutcome, ExecutionRuntimeContext, ExecutionStateView, Message, MessageContext,
@@ -334,6 +335,7 @@ where
         &mut self,
         local_time: Timestamp,
         query: Query,
+        global_context: &GlobalContext,
         service_runtime_endpoint: Option<&mut ServiceRuntimeEndpoint>,
     ) -> Result<QueryOutcome, ChainError> {
         let context = QueryContext {
@@ -342,7 +344,12 @@ where
             local_time,
         };
         self.execution_state
-            .query_application(context, query, service_runtime_endpoint)
+            .query_application(
+                context,
+                query,
+                global_context,
+                service_runtime_endpoint
+            )
             .await
             .with_execution_context(ChainExecutionContext::Query)
     }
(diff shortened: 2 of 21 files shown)


usagi32:
@ma2bd

ma2bd:
> currently, anyone is able to mint a same token twice

That is not correct. The ApplicationID will be different.

ma2bd:
> currently, anyone is able to mint a same token twice without any consequences

Ok so for fungible tokens, we have a unique TokenID which is the ApplicationID.

For the NFT demo in `examples/non-fungible` (not quite a standard but certainly a first step), we also generate a unique TokenID (see `create_token_id).

usagi32:
@ma2bd 

Actually, that's not what I was talking about.

Let's say there is an artwork to be minted. Nothing is stopping from minting the same artwork data 100 times. This is a classic problem in blockchains. Even though the things like ContractID, ApplicationID etc can be different, and that's also is not given, and the original artwork can be traced back to the first mint, but this all includes uncertainity and multiple levels of indirection and is not possible for run of the mill users.
Usually what happens is, the same artworks are minted, listed and sold on different marketplaces by frauders and there is no way to check it, if there is no knowledge about it to the original owner, marketplace or the buyer. This leads to common nft frauds, as this is just an example, this nature usually becomes the breeding hotspot for all kind of nft related frauds.

Secondly, nfts in blockchains like Ethereum and Stacks are just represented by the pair tuple (ContractId, TokenId) where TokenID is a uint256. It is this pair tuple that is considered globally unique, even though the TokenID has same amount of entropy (256 bits) and space as a Sha-256 hash and the potential to be globally unique by itself, it is local to its respective contract and just represents a mint index.
The mapping that relates TokenID to its metadata (like to the artwork or json files) is a philosophical discussion and a internal matter of the contract itself, and is not required to be immutable or be there at all. Artists can any time mutate it as willed. And other blockchain do not require it to be present at all leading to minter related frauds and rugpulls. That's what I meant when I was referring to Solana's usage of accounts (metadata accounts also have an optional field is_mutable). Solana has a 1 : 1 mapping between the tokens and their metadata built in the protocol itself.

Any modern blockchain should be free of such loopholes and flaws, if it's taking its ease of use and security seriously.

ma2bd:
@usagi32 I appreciate the detail answer, but going back to this PR, I'm not sure to see the problem that we want to address. We have unique TokenIDs as everyone and (if we wanted) we could allow parts of an NFT to be mutable. In this respect, microchains are not limiting here because NFT metadata mutations would be driven by a single writer (the minting chain). There is a little bit of work to "broadcast" the mutations everywhere but we have a solution in progress ("event streams").

usagi32:
Maybe I am not clear in my wordings,
I see microchains to be at greater disadvantage than other blockchains is this problem.

- Nft creators will likely have their own applications and algorithms on how their TokenIDs are generated
- An artwork image (that is already a nft somewhere) serialized and hashed can be copied to another contract or application, minted there, and sold as same piece of art on other marketplaces, where the original image is not listed yet, without the permission of the artist or owner, leading to frauds. This pr assures that only a single real world entity exists and is minted for once on the Protocol irrespective of the ChainID and ApplicationID, as they are not considered in the minting process. Only the real entity that is to be tokenized is used as a unique identity.
- Mutations in nfts and metadata are usually discouraged and seen as less desirable, unless nfts are made for that purpose. Nft creators, with their own applications, can, let's take an example - make a nft application, sell a nft to another user, go back to their original application and as they are owners of that application can any time mutate the metadata, doing rugpulls and  common frauds in nft space, as the original nft is worthless now.
- As creators have their own custom applications, there is no mandation of metadata to be present
- No enforcing that metadata is 1 : 1 mapped, usually in something like ethereum uint256 ID is stored in map like structure corresponding to its metadata. This map is the only thing that links the tokens (which are nothing but a uint256 integer to the virtual machine) to its original data like an image. But this map is not necessary to be part of the contract. This is an old philosophical question that "what really links an outside image to a unique token on the blockchain". And this is what Solana does with a associated metadata account with each mint account.

If there is another solution to these problems then, its good.

afck:
This approach would create a single bottleneck in the database, in the form of the global state view. Linera is all about avoiding that, and I don't think there is a solution to the problem you describe that does not need a single global state.

Also, I'd argue the protection gained from that is quite small: One can always change some pixel in some NFT image and sell it as a different NFT. Or mint it again on a completely different platform.

I'll close this. Feel free to reopen if you disagree, but please as a draft PR (since it's not fully implemented).

usagi32:
Yes but workers are not sharded yet, I guess.
Even after that the memory footprint of hashmaps not really makes them an ideal solution.

And yes, I was thinking about that only when referring to lock-free data structure. It will be made into a view. After #3350, the root node will have a branching factor of 256, so it can be easily managed which branch is currently active and where at the storage level. Further, concurrent and lock-free access will be available till at the last depth level above the leaf. Since the data resides at the leaves only, the cursor or the walker is free to roam around and travel anywhere else guaranteeing instant IO.

usagi32:
Actually, I want to discuss about these things thoroughly with you, and I would also love to get your thoughts about #3351

afck:
IMO #3351 is something we may revisit much later; we have to prioritize.

Workers _are_ already sharded. Each validator runs multiple worker processes on multiple machines.

ma2bd:
> @afck Any possibility to get involved with the team?

@usagi32 Thanks for asking. Can you send me a resume and references on telegram? (same handle as github)

Issue #460, opened 2023-02-20
Allow the state of chain to be check-pointed


ma2bd:
The goal is that the history of the chain before the checkpoint doesn't need to be downloaded any more by clients.
Possible ingredients to a solution include:
* cross-chain messages to the chain itself
* exporting the storage state(s) in a blob.

PR #3350, opened 2025-02-16
Block Header and Merkle Proofs

usagi32:
For this pr, I define:
- Client - A normal node peer storing, if wished, the entire chain of concern. For example, a desktop client node.
- Light Client - Node peers with limited and constrained resources which are only required to store the Block-Headers. For example, a web or mobile wallet.
- Stateless Client - For example, a smart contract in a chain in the outside world. They only need a single good known trusted hash to verify the zero knowledge proofs till that point in time. 
<br/>
A proper Block-Header is a single point of consensus, among all peers, which acts as a snapshot of the entire history of the respective blockchain state, up until the corresponding block height, in a compressed form. 

It should be:
- A unique identity for a block.
- Compact, Precise and Elegant as it will be the face of the blockchain for the outside world.
- Simple and easy to develop around protocols as it will act as a bridge between the outside world.
- Able to verify the entire history of the chain state with minimal amount of resources possible in at most sub-linear complexity for both the time and the space.
- Able to allow the light clients to prune the chain state.

Other than these, the surrounding blockchain should support quick, generation of validity proofs, adding a new block with its header to the chainstate in sub-linear time, small proof sizes, etc.
<br/>
Fortunately, unlike other blockchains, the way linera is, we don't have to send the blockheader to other peers. It can be derived at the either ends, saving the block space, as its not part of the consensus(yet, there is a small space where it can be, for the better).
And also improving on the recent implementation, we don't have to send the block execution outcomes, as they are deterministic and dependent on the inputs, greatly saving the precious block space.
But, what's the incentive for the clients to cryptographically commit to the block-headers?
The state of the chain can be check-pointed, a snapshot, and the history of the chain before that point can be safely pruned.
<br/>
Further, to maintain those, likely to be millions of, transaction hashes, a index data structure is needed. After careful considerations, a merkle trie is a most likely candidate. A trie like data structure is used by every major blockchain, Ethereum uses a modified merkle patricia tree, Solana uses a form of concurrent tree, etc. And we require a trie data structure at many places anyway, like in linera-views, and in future to manage Accounts and chain ids.
<br/>
For our purposes, a best fit would an Adaptive Radix Merkle Trie (ART). Details about it can be found in the original paper<sup>[[1]](https://db.in.tum.de/~leis/papers/ART.pdf)</sup> below.
Adding a new block to the index require minimal samples 
x
Reads are O(1) 
Writes are O(1)
Adding a new block is O(logn)
Generating a proof is O(logn)
Verifying a proof is O(logn)
Proof size is O(logn)
<br/>

Lastly, a merkle skip list is used to calculate the canonical block hash, which is the concatenation of the current root hash with the consensus hashes N - 1, N - 2, N - 4, N - 8 ... and so on.
The benefits to this can be many due to the magical nature of numbers and cryptrography, but two main reasons are 
- Early detection of forks and malicious peers at the point of fork.
- Possibility of Existence of stateless clients.
## Links
- [1] https://db.in.tum.de/~leis/papers/ART.pdf

### DIFF
--- a/linera-base/src/codec.rs
+++ b/linera-base/src/codec.rs
@@ -1,0 +1,172 @@
+//! Defines a universal consensus codec for Linera.
+
+use core::{error, fmt};
+use std::io::{self, Read, Write};
+
+use crate::{crypto::CryptoHash, data_types::BlockHeight, identifiers::ChainId};
+
+/// Error type for serialization related operations.
+#[derive(Debug)]
+pub enum Error {
+    /// Failed to encode
+    SerializeError(String),
+    /// Failed to read
+    ReadError(io::Error),
+    /// Failed to decode
+    DeserializeError(String),
+    /// Failed to write
+    WriteError(io::Error),
+    /// Underflow -- not enough bytes
+    UnderflowError(String),
+    /// Overflow -- too big
+    OverflowError(String),
+    /// Array is too big
+    ArrayTooLong,
+    /// Generic error
+    GenericError(String),
+}
+
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match *self {
+            Error::SerializeError(ref s) => fmt::Display::fmt(s, f),
+            Error::DeserializeError(ref s) => fmt::Display::fmt(s, f),
+            Error::ReadError(ref io) => fmt::Display::fmt(io, f),
+            Error::WriteError(ref io) => fmt::Display::fmt(io, f),
+            Error::UnderflowError(ref s) => fmt::Display::fmt(s, f),
+            Error::OverflowError(ref s) => fmt::Display::fmt(s, f),
+            Error::ArrayTooLong => write!(f, "Array too long"),
+            Error::GenericError(ref s) => fmt::Display::fmt(s, f),
+        }
+    }
+}
+
+impl error::Error for Error {
+    fn source(&self) -> Option<&(dyn error::Error + 'static)> {
+        match *self {
+            Error::SerializeError(ref _s) => None,
+            Error::ReadError(ref io) => Some(io),
+            Error::DeserializeError(ref _s) => None,
+            Error::WriteError(ref io) => Some(io),
+            Error::UnderflowError(ref _s) => None,
+            Error::OverflowError(ref _s) => None,
+            Error::ArrayTooLong => None,
+            Error::GenericError(ref _s) => None,
+        }
+    }
+}
+
+/// Types implementing a single, universally agreed consensus.
+/// Helps to make serialization and deserialization a single pipeline
+/// creating a confidence of compatibilty among all types implementing it.
+pub trait Codec {
+    /// Serialize directly into a writable file descriptor.
+    fn consensus_serialize<W: Write>(&self, fd: &mut W) -> Result<(), Error>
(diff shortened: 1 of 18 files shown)


usagi32:
@ma2bd

afck:
Thank you for the suggestions!

I'll mark this PR as a draft since it's not meant to be merged in the current state (lots of `todo!()`s) and requires a few discussions and product decisions.

usagi32:
As @afck requested me to put some things in this conversation, I'll just point out the why our current recent approach is wrong and few enhancements this pr will bring.

Currently,
- We are wasting a lot of precious block space in including transaction execution outcomes. As they are deterministic and are dependent variables on the inputs (operations and messages). We only need the independent input variables to be propagated to the network.
- The block headers themselves can be constructed at either ends and don't have to be included in the proposal. Unlike other blockchains where these headers are different for everyone and they have to agree on a canonical consensus, block headers are deterministic in linera and we only have to propagate the minimal data necessary to the network. Few extra things can be made part of consensus, to make sure the cryptographic commitment, but I doubt the necessity of even that as that's a philosophical discussion to be.
- Hashing is done linearly. This is a naive approach and completely disregards the purpose of a block header. If we still have to the send the whole block to prove the inclusion on an element, we might as well not use block headers at all, rather just a trusted hash of the block.

We do need an index for all transactions, as they ought to be globally unique. In Bitcoin, a transaction index is a part of the protocol itself. In previous versions of the bitcoin core, this was part of the consensus but it's optional now, likely used by api nodes, rpc nodes, wallet service providers etc.

Other than these problems, this pr contributes some to the solutions of
- State compression
- Pruned nodes (see #460)
- Light clients like mobile wallets and web extensions
- Stateless clients

usagi32:
>  We include them so that consumers don't need to re-execute the whole block to get them (which would require syncing the state and getting the transactions that produced those outcomes)

We never send the outcomes, we just query the validator for them and the corresponding proofs of inclusion in that block. I guess that's how they work in ethereum (transaction receipts), they are not included in the block proposal themself, but just a hash in the header.

> assumption is that since Linera's microchains are small 

Yes, actually i did considered that assumption. But there is just that.. I know there are all kinds of clients. And for some reasons, some mint 1000s of NFT's at the same time and all that. Plus there eventually be services built on top of a single microchain like, just for example, a layer 2. And a merkle hash is not bad at all even for small blocks.

 > but will perform more hashing

It doesn't? Its still log(n).
>vector approach it will do only 1 hash (of the whole vector)

Hashing of the whole vector? Isn't that only what's we are trying to avoid?

usagi32:
I know that, and that's still all under the assumption that microchains are always small. We don't want that much performance when there are only like 2 transactions in a block, then it's all same. And we will lose out where it all really matters.

>  4 times to verify

Let's just say there are 256 transactions in a block, and that by no margin is a big number and it's completely expected daily. To still prove some thing at index 3 we are essentially sending the whole block, while hashing it all at receiver's end.
Just using a tree, we just need 8 hashes to hash.
Now, let's say transaction count is someday double (512), if we assume our clients are built around low number, there will be network congestions. Still just using a tree we only need 1 more hash (i.e 9 hashes).

usagi32:
Yeah, I meant the vector. I won't want that.

Its something too simple and important to delay it.

PR #3354, opened 2025-02-17
Rename `WasmExecutionError`.

MathieuDutSik:
## Motivation

Subsequent to https://github.com/linera-io/linera-protocol/pull/3327 it was requested that the renaming done in a separate PR. This is done here.

## Proposal

Following changes are done:
* Rename `WasmExecutionError` as `VmExecutionError`.
* Rename `WasmError` as `VmError`.
* Move the `VmExecutionError` and `{CONTRACT,SERVICE}_INSTANTIATION_LATENCY` to `linera_execution/src/lib.rs`

## Test Plan

No change to CI as it is simply a renaming.

## Release Plan

Normal release schedule.

## Links

None.

### DIFF
--- a/linera-execution/src/wasm/mod.rs
+++ b/linera-execution/src/wasm/mod.rs
@@ -21,16 +21,14 @@ mod wasmer;
 mod wasmtime;
 
 use linera_base::data_types::Bytecode;
-use thiserror::Error;
 #[cfg(with_wasmer)]
 use wasmer::{WasmerContractInstance, WasmerServiceInstance};
 #[cfg(with_wasmtime)]
 use wasmtime::{WasmtimeContractInstance, WasmtimeServiceInstance};
 #[cfg(with_metrics)]
 use {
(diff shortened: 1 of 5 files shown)


linera-execution/src/wasm/mod.rs
```diff
@@ -268,40 +246,10 @@ const _: () = {
     }
 };
 
-/// Errors that can occur when executing a user application in a WebAssembly module.
-#[cfg(any(with_wasmer, with_wasmtime))]
-#[derive(Debug, Error)]
-pub enum WasmExecutionError {
```
ma2bd, line 274: I feel that this one should stay as it is, and you would just create a new type of Error for the EVM.

ma2bd: It's fair normal for each module to have its Error type.

MathieuDutSik: Ok, let's kill the PR.

MathieuDutSik:
The merging of errors into a `VmExecutionError` was deemed a bad idea and so the PR is killed.

PR #3355, opened 2025-02-17
Implement the functionalities of listing the `ChainIds`.

MathieuDutSik:
## Motivation

We implement here the functionality of listing the Chain Ids from a storage.

## Proposal

The implementation follows the same path as from the implementation of the listing of the BlobIds by using the `list_root_keys`.

## Test Plan

The functionality was tested locally by running the non-fungible test and listing the ChainIds.

## Release Plan

Normal release schedule.

## Links

None.

### DIFF
--- a/CLI.md
+++ b/CLI.md
@@ -57,6 +57,7 @@ This document contains the help content for the `linera` command-line program.
 * [`linera storage initialize`↴](#linera-storage-initialize)
 * [`linera storage list_namespaces`↴](#linera-storage-list_namespaces)
 * [`linera storage list_blob_ids`↴](#linera-storage-list_blob_ids)
+* [`linera storage list_chain_ids`↴](#linera-storage-list_chain_ids)
 * [`linera storage list_root_keys`↴](#linera-storage-list_root_keys)
 
 ## `linera`
@@ -964,7 +965,8 @@ Operation on the storage
 * `check_absence` — Check absence of a namespace in the database
 * `initialize` — Initialize a namespace in the database
 * `list_namespaces` — List the namespaces of the database
-* `list_blob_ids` — List the blobs of the database
+* `list_blob_ids` — List the blob ids of the database
+* `list_chain_ids` — List the chain ids of the database
 * `list_root_keys` — List the root keys of the database
 
 
@@ -1043,9 +1045,21 @@ List the namespaces of the database
 
 ## `linera storage list_blob_ids`
 
-List the blobs of the database
+List the blob ids of the database
 
 **Usage:** `linera storage list_blob_ids --storage <STORAGE_CONFIG>`
+
+###### **Options:**
+
+* `--storage <STORAGE_CONFIG>` — Storage configuration for the blockchain history
+
+
+
+## `linera storage list_chain_ids`
+
+List the chain ids of the database
+
+**Usage:** `linera storage list_chain_ids --storage <STORAGE_CONFIG>`
 
 ###### **Options:**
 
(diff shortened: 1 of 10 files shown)


CLI.md
```diff
@@ -57,6 +57,7 @@ This document contains the help content for the `linera` command-line program.
 * [`linera storage initialize`↴](#linera-storage-initialize)
 * [`linera storage list_namespaces`↴](#linera-storage-list_namespaces)
 * [`linera storage list_blob_ids`↴](#linera-storage-list_blob_ids)
+* [`linera storage list_chain_ids`↴](#linera-storage-list_chain_ids)
 * [`linera storage list_root_keys`↴](#linera-storage-list_root_keys)
```
ma2bd, line 61: do we need to list root keys?

MathieuDutSik: I think this can be useful for debugging, but if you think not, then so be it.

CLI.md
```diff
@@ -1053,11 +1053,11 @@ List the blobs of the database
 
 
 
-## `linera storage list_root_keys`
+## `linera storage list_chain_ids`
 
-List the root keys of the database
+List the chain ids of the database
```
ma2bd, line 1058: IIRC @afck prefers `IDs` ?

linera-storage/src/db_storage.rs
```diff
@@ -296,24 +300,52 @@ mod tests {
...
+        assert_eq!(key[0], INDEX_CHAIN_ID);
+        assert_eq!(key.len(), 1 + CHAIN_ID_LENGTH);
     }
 }
 
-/// Lists the blobs of the storage.
+/// Lists the blob ids of the storage.
 pub async fn list_all_blob_ids<S: KeyValueStore>(store: &S) -> Result<Vec<BlobId>, ViewError> {
-    let prefix = &[INDEX_BLOB];
+    let prefix = &[INDEX_BLOB_ID];
```
ma2bd, line 320: Thanks for the clarification

linera-storage/src/db_storage.rs
```diff
@@ -296,24 +300,52 @@ mod tests {
...
 
+/// Lists the chain ids of the storage.
+pub async fn list_all_chain_ids<S: AdminKeyValueStore>(
+    config: &S::Config,
+    namespace: &str,
+) -> Result<Vec<ChainId>, ViewError> {
+    let root_keys = S::list_root_keys(config, namespace).await?;
+    let mut chain_ids = Vec::new();
+    for root_key in root_keys {
+        if root_key.len() == 1 + CHAIN_ID_LENGTH && root_key[0] == INDEX_CHAIN_ID {
```
ma2bd, line 340: Should we print a warning if some root keys don't match?

CLI.md
MathieuDutSik: Ok, changed (and also in some other places.

linera-client/src/client_options.rs
```diff
@@ -992,17 +992,17 @@ pub enum DatabaseToolCommand {
         storage_config: String,
     },
 
-    /// List the blobs of the database
+    /// List the blob IDs of the database
```
afck, line 995: (Maybe "in the database" would be clearer? Also below for chain IDs.)

linera-storage/src/db_storage.rs
```diff
@@ -277,17 +277,21 @@ enum BaseKey {
     Event(EventId),
 }
 
-const INDEX_BLOB: u8 = 3;
-const BLOB_LENGTH: usize = std::mem::size_of::<BlobId>();
+const INDEX_CHAIN_ID: u8 = 0;
+const INDEX_BLOB_ID: u8 = 3;
+const CHAIN_ID_LENGTH: usize = std::mem::size_of::<ChainId>();
+const BLOB_ID_LENGTH: usize = std::mem::size_of::<BlobId>();
```
afck, line 283: It's actually about the BCS-serialized size, isn't it? That that's the in-memory size is just accidental.

Or rather, if we want the key to be exactly the chain/blob ID, why do we use BCS at all? Why not use the IDs' bytes directly (which is probably the same anyway)?

linera-client/src/client_options.rs
MathieuDutSik: Ok.

afck (approved):

PR #3357, opened 2025-02-17
Use secp256k1 features depending on the target arch.

deuszx:
## Motivation

We want to have `thread::rng` in `std` environments but `Secp256k1` constructor feature-gates its usage based on the target arch which does NOT randomize the secp256k1 context if `feature = "std"` is not enabled.
```rust
#[cfg(all(
                not(target_arch = "wasm32"),
                feature = "rand",
                feature = "std",
                not(feature = "global-context-less-secure")
            ))]
            {
                ctx.randomize(&mut rand::thread_rng());
            }
```

To get around the compilation of `linera-base` in wasm32 environments (like indexer or examples) I used `alloc` feature instead of `std` but that means secp256k1 context was NOT randomized properly which could have lead to some cryptography-related problems.

## Proposal

Enable `secp256k1` features conditionally - depending on the target architecture for which it is being compiled.

## Test Plan

N/A

## Release Plan

## Links

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1585,6 +1585,22 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5e764a1d40d510daf35e07be9eb06e75770908c27d411ee6c92109c9840eaaf7"
 
 [[package]]
+name = "bitcoin-io"
+version = "0.1.3"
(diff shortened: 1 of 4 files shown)


jvff (approved):

Twey (approved):

deuszx:
Not needed for now.

PR #3359, opened 2025-02-17
Cleanup graceful shutdown a bit

ndr-ds:
## Motivation

This could be written sligthly better IMHO

## Proposal

Clean it up a bit

## Test Plan

CI + run it locally

## Release Plan


ndr-ds:
I'm gonna stop saving to the wallet at all, so this will be irrelevant with that PR

PR #3366, opened 2025-02-18
Refactor: Separate authority and chain owner keys.

deuszx:
## Motivation

In order to sign blocks with secp256k1 signatures (while keeping ed25519 for chain owners/block proposals) we need to be able to differentiate between different type of keys. At the moment we have two types of actors in the system: validators and chain owners. Fact that both use the same public-key cryptography (ed25519) is orthoganal and will change soon. Authorities (validators) will need to be able to act as chain owners in the cases where they become _fallback_ chain owners. They will also receive rewards for their work so they need to be able to transfer these tokens.

## Proposal

Use type aliases `Authority*` and `Account*` (`*PrivateKey`, `*PublicKey`, `*Signature`) for validators and chain owners respectively. Right now it's a type alias (without a newtype wrapper) as these types will become enums in the near future.

## Test Plan

CI should catch any regressions (but there should not be any, it's just a rename).

## Release Plan

## Links

### DIFF
--- a/linera-base/src/crypto/mod.rs
+++ b/linera-base/src/crypto/mod.rs
@@ -16,12 +16,12 @@ pub use hash::*;
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
 
-/// The public key of an authority.
-pub type AuthorityPublicKey = ed25519::Ed25519PublicKey;
-/// The private key of an authority.
-pub type AuthorityPrivateKey = ed25519::Ed25519SecretKey;
-/// The signature of an authority.
-pub type AuthoritySignature = ed25519::Ed25519Signature;
+/// The public key of a validator.
+pub type ValidatorPublicKey = ed25519::Ed25519PublicKey;
+/// The private key of a validator.
+pub type ValidatorPrivateKey = ed25519::Ed25519SecretKey;
+/// The signature of a validator.
+pub type ValidatorSignature = ed25519::Ed25519Signature;
 
 /// The public key of a chain owner.
 /// Corresponding private key is allowed to propose blocks
--- a/linera-base/src/crypto/mod.rs
+++ b/linera-base/src/crypto/mod.rs
@@ -11,13 +11,26 @@ mod secp256k1;
 use std::{io, num::ParseIntError};
 
 use alloy_primitives::FixedBytes;
-pub use ed25519::{
-    Ed25519PublicKey as PublicKey, Ed25519SecretKey as SigningKey, Ed25519Signature as Signature,
-};
 use ed25519_dalek::{self as dalek};
 pub use hash::*;
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
+
+/// The public key of an authority.
+pub type AuthorityPublicKey = ed25519::Ed25519PublicKey;
+/// The private key of an authority.
+pub type AuthorityPrivateKey = ed25519::Ed25519SecretKey;
+/// The signature of an authority.
+pub type AuthoritySignature = ed25519::Ed25519Signature;
+
+/// The public key of a chain owner.
+/// Corresponding private key is allowed to propose blocks
+/// on the chain and transfer account's tokens.
+pub type AccountPublicKey = ed25519::Ed25519PublicKey;
+/// The private key of a chain owner.
+pub type AccountPrivateKey = ed25519::Ed25519SecretKey;
+/// The signature of a chain owner.
+pub type AccountSignature = ed25519::Ed25519Signature;
 
 /// Error type for cryptographic errors.
 #[derive(Error, Debug)]
--- a/linera-base/src/crypto/mod.rs
+++ b/linera-base/src/crypto/mod.rs
@@ -24,7 +24,7 @@ pub type ValidatorPrivateKey = ed25519::Ed25519SecretKey;
 pub type ValidatorSignature = ed25519::Ed25519Signature;
 
 /// The public key of a chain owner.
-/// Corresponding private key is allowed to propose blocks
+/// The corresponding private key is allowed to propose blocks
 /// on the chain and transfer account's tokens.
 pub type AccountPublicKey = ed25519::Ed25519PublicKey;
 /// The private key of a chain owner.
--- a/linera-base/src/crypto/mod.rs
+++ b/linera-base/src/crypto/mod.rs
@@ -19,7 +19,7 @@ use thiserror::Error;
 /// The public key of a validator.
 pub type ValidatorPublicKey = ed25519::Ed25519PublicKey;
 /// The private key of a validator.
-pub type ValidatorPrivateKey = ed25519::Ed25519SecretKey;
+pub type ValidatorSecretKey = ed25519::Ed25519SecretKey;
 /// The signature of a validator.
 pub type ValidatorSignature = ed25519::Ed25519Signature;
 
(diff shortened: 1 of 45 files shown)


linera-base/src/crypto/mod.rs
```diff
@@ -11,14 +11,27 @@ mod secp256k1;
...
 use alloy_primitives::FixedBytes;
-pub use ed25519::{
-    Ed25519PublicKey as PublicKey, Ed25519SecretKey as SigningKey, Ed25519Signature as Signature,
-};
 use ed25519_dalek::{self as dalek};
 pub use hash::*;
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
 
+/// The public key of an authority.
```
afck, line 19: Let's use the term "validator" instead of "authority" for consistency.

linera-base/src/crypto/mod.rs
```diff
@@ -11,14 +11,27 @@ mod secp256k1;
...
+pub type AuthorityPublicKey = ed25519::Ed25519PublicKey;
+/// The private key of an authority.
+pub type AuthorityPrivateKey = ed25519::Ed25519SecretKey;
+/// The signature of an authority.
+pub type AuthoritySignature = ed25519::Ed25519Signature;
+
+/// The public key of a chain owner.
+/// Corresponding private key is allowed to propose blocks
```
afck, line 27:
```suggestion
/// The corresponding private key is allowed to propose blocks
```

linera-chain/src/manager.rs
```diff
@@ -500,7 +500,7 @@ where
     pub fn create_final_vote(
         &mut self,
         validated: ValidatedBlockCertificate,
-        key_pair: Option<&SigningKey>,
+        key_pair: Option<&AuthorityPrivateKey>,
```
afck, line 503: Don't we actually want key pairs instead of secret keys here?

linera-base/src/crypto/mod.rs
```diff
@@ -11,14 +11,27 @@ mod secp256k1;
...
-};
 use ed25519_dalek::{self as dalek};
 pub use hash::*;
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
 
+/// The public key of an authority.
+pub type AuthorityPublicKey = ed25519::Ed25519PublicKey;
+/// The private key of an authority.
+pub type AuthorityPrivateKey = ed25519::Ed25519SecretKey;
```
afck, line 22: Let's use the term "secret key", because "private key" can be abbreviated PK, which is confusing.

But maybe we should only use key _pairs_ almost everywhere anyway?

linera-sdk/src/test/validator.rs
```diff
@@ -67,7 +67,7 @@ impl Clone for TestValidator {
 impl TestValidator {
     /// Creates a new [`TestValidator`].
     pub async fn new() -> Self {
-        let key_pair = SigningKey::generate();
```
afck, line 70: Shouldn't this be a validator key?

linera-rpc/src/grpc/conversions.rs
```diff
@@ -1207,21 +1209,20 @@ pub mod tests {
...
             content: ProposalContent {
                 block: get_block(),
                 round: Round::SingleLeader(4),
                 outcome: Some(outcome),
             },
-            owner: Owner::from(public_key),
-            public_key,
-            signature: Signature::new(&Foo("test".into()), &SigningKey::generate()),
+            public_key: key_pair.public(),
+            signature: AccountSignature::new(&Foo("test".into()), &key_pair),
```
afck, line 1225: A block proposal should be signed by an account key.

linera-chain/src/data_types.rs
```diff
@@ -285,9 +288,8 @@ pub enum Medium {
 #[cfg_attr(with_testing, derive(Eq, PartialEq))]
 pub struct BlockProposal {
     pub content: ProposalContent,
-    pub owner: Owner,
```
afck, line 288: Not sure if we want to remove that, or have the owner explicitly in the proposal.

I actually like the idea, since it removes a way in which a proposal could be invalid (and which could be missed on validation). But it should be added to the PR description.

(Originally the idea was that since we are using `Owner` in most user-facing places now, it is more self-explanatory if the user is an explicit field.)

linera-chain/src/unit_tests/chain_tests.rs
```diff
@@ -93,9 +93,9 @@ fn make_admin_message_id(height: BlockHeight) -> MessageId {
 }
 
 fn make_open_chain_config() -> OpenChainConfig {
-    let committee = Committee::make_simple(vec![PublicKey::test_key(1).into()]);
+    let committee = Committee::make_simple(vec![AuthorityPublicKey::test_key(1).into()]);
     OpenChainConfig {
-        ownership: ChainOwnership::single(PublicKey::test_key(0).into()),
+        ownership: ChainOwnership::single(AuthorityPublicKey::test_key(0).into()),
```
afck, line 98: That's an account.

linera-rpc/src/grpc/conversions.rs
deuszx: And that is what's happening here, isn't it? I use `&key_pair` which is generated as `AccountPrivateKey::generate()`.

linera-chain/src/manager.rs
deuszx: We can, but it was the PrivateKey-only before this as well and public can be derived.

linera-rpc/src/grpc/conversions.rs
afck: Ah, right. :facepalm:

linera-chain/src/data_types.rs
deuszx: I can extract it to a separate PR if you prefer?

deuszx: This change was reverted in [Revert "Remove Owner from BlockProposal."](https://github.com/linera-io/linera-protocol/pull/3366/commits/c1bb7b452e7501e326dfecc2e8f0b30cfb27b606). I will create a separate PR for it.

linera-base/src/crypto/mod.rs
deuszx: Renamed.

linera-base/src/crypto/mod.rs
deuszx: Done.

linera-chain/src/unit_tests/chain_tests.rs
deuszx: Done.

linera-sdk/src/test/validator.rs
deuszx: Done in [Use proper keys in tests](https://github.com/linera-io/linera-protocol/pull/3366/commits/70fe86eb4bed9152d1526f990edcf99735f48c97)

linera-chain/src/manager.rs
afck: OK, something for a separate PR as well then.

I was just wondering whether there's any computational overhead if we don't have the public key available here.

afck (approved):

linera-chain/src/manager.rs
deuszx: Probably there is - especially that for secp256k1 keys it will go through FFI. We can change that later.

PR #3367, opened 2025-02-18
Stop updating the wallet during linera benchmark

ndr-ds:
## Motivation

For the chains we're using during the benchmark, we were currently adding them to our wallet. However, that is not necessary.

## Proposal

Create chains for the benchmark, and don't add them to any wallet, just use them for the benchmark then close them.

## Test Plan

Ran locally, saw that things worked correctly

## Release Plan

### DIFF
--- a/linera-client/src/client_context.rs
+++ b/linera-client/src/client_context.rs
@@ -8,7 +8,7 @@ use std::{collections::HashSet, sync::Arc};
 use async_trait::async_trait;
 use futures::Future;
 use linera_base::{
-    crypto::AccountSecretKey,
+    crypto::{AccountSecretKey, CryptoHash},
     data_types::{BlockHeight, Timestamp},
     identifiers::{Account, ChainId},
     ownership::ChainOwnership,
@@ -16,7 +16,7 @@ use linera_base::{
 };
 use linera_chain::types::ConfirmedBlockCertificate;
 use linera_core::{
-    client::{BlanketMessagePolicy, ChainClient, Client, MessagePolicy},
+    client::{BlanketMessagePolicy, ChainClient, Client, MessagePolicy, PendingProposal},
     data_types::ClientOutcome,
     join_set_ext::JoinSet,
     node::CrossChainMessageDelivery,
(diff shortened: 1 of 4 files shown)


ndr-ds:
* **#3367** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/3367?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/3367?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

linera-client/src/client_context.rs
```diff
@@ -873,6 +921,7 @@ where
         &mut self,
         key_pairs: &HashMap<ChainId, SigningKey>,
         application_id: ApplicationId,
+        chain_states: &HashMap<ChainId, (Option<CryptoHash>, BlockHeight, Timestamp)>,
```
afck, line 924: Or should we keep the chain clients themselves, rather than recreating them here?

ndr-ds: Ah, good point, I can just always pass the chain clients around 😅

linera-client/src/client_context.rs
```diff
@@ -826,18 +799,75 @@ where
...
+                .handle_chain_info_query(query)
+                .await?
+                .info;
+
+            let chain_client = self.make_chain_client_internal(
+                *chain_id,
+                vec![key_pair.copy()],
+                chain_info.block_hash,
+                chain_info.timestamp,
+                chain_info.next_block_height,
```
afck, line 823: This will always be

```rust
                None,
                timestamp_of_certificate_in_line_767,
                BlockHeight::ZERO,
```

won't it? I think the timestamp doesn't even matter as long as the local clock isn't reset to an earlier time, so we could set that to zero or `now()`.

afck (approved):

linera-client/src/client_context.rs
ndr-ds: Good point!

linera-service/src/linera/main.rs
```diff
@@ -749,18 +749,26 @@ impl Runnable for Job {
...
+                    .wallet
+                    .default_chain()
+                    .expect("should have default chain");
+                let chain_client = context.make_chain_client(default_chain_id)?;
                 // Below all block proposals are supposed to succeed without retries, we
                 // must make sure that all incoming payments have been accepted on-chain
                 // and that no validator is missing user certificates.
-                context.process_inboxes_and_force_validator_updates().await;
+                context
+                    .process_inbox_without_updating_wallet(&chain_client)
```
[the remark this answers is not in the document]
ndr-ds: True, I guess if something else changes it (if it's used for something other than the benchmark). Good point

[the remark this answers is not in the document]
afck: Also because of the blocks we created on the default chain to open the benchmark chains. The default chain has a greater block height after the benchmark than before.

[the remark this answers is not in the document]
ndr-ds: True 😅

afck (approved):

PR #3370, opened 2025-02-19
Correct the caching of the solidity compiler.

MathieuDutSik:
## Motivation

The caching is not working for Solc. We attempt to correct it.

## Proposal

Correct the CI code in order to have the solc compiler.

## Test Plan

The CI.

## Release Plan

Not relevant.

## Links

None

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -177,6 +177,8 @@ jobs:
       uses: pontem-network/get-solc@master
       with:
         version: v0.8.25
+    - name: Check if solc is in PATH
+      run: which solc || echo "Solc not found in PATH"
     - name: Install Protoc
       uses: arduino/setup-protoc@v1
       with:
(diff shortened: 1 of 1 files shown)


ndr-ds (approved):
Accepting to unblock, but please verify with your test that uses `solc` directly that this works before merging

.github/workflows/rust.yml
```diff
@@ -177,6 +179,14 @@ jobs:
       uses: pontem-network/get-solc@master
       with:
         version: v0.8.25
+    - name: Add Solc to PATH
+      run: echo "/home/runner/.solc" >> $GITHUB_PATH
+    - name: List the content of the Solc cache
+      run: ls /home/runner/.solc/v0.8.25/solc-static-linux
+    - name: Create a Solc symbolic link
+      run: ln -sf /home/runner/.solc/v0.8.25/solc-static-linux /home/runner/.solc/solc
```
deuszx, line 187: We could extract version to an env var and reuse everywhere.

PR #3374, opened 2025-02-19
Have one thread per chain on benchmark

ndr-ds:
## Motivation

Right now on `linera benchmark` we do everything from a single thread. That doesn't scale much.

## Proposal

Spawn one task per chain, and split the desired BPS across the number of chains. For example, if the desired BPS is 10, and you have 10 chains, each task will need to reach 1 BPS.

Every task just tries to send as many block proposals as fast as possible. We don't control the BPS in the tasks, but we have a BPS control task for that. Once a task has sent the BPS amount of blocks successfully, it sends a message to the BPS control task. It uses a `crossbeam` channel for that, as it allows bounded channels with buffer size of 0.
If we have a buffer size of 0, that means that if the BPS control task is sleeping or processing another message, the sender task will be blocked. This is exactly what we want to properly control the BPS behavior. These are faster than `std` channels and comparable speed of `tokio` channels, but `tokio` channels don't allow a buffer size of 0, and also don't block the sender task as sending is async.
The BPS control task will have a timer going. Once it receives `num_chains` messages, that means that the total number of blocks to be sent was reached. If more than a second elapsed, we failed to achieve the desired BPS. Otherwise, we achieved the desired BPS.

This all runs sharing the same wallet and `ClientContext`. The only resource the different threads could race for is the wallet, but since we don't save the benchmark chains to the wallet anymore, that's no longer a concern. So using the same wallet for multiple threads is feasible, because the wallet is only used to create the chains at the start using its default chain, then the wallet isn't used by the threads anymore.

Shutdown signals also continue to work (all threads are gracefully shutdown), and we continue to close all the chains on exit.

## Test Plan

Ran with a few different BPS/TPB variations, seems to work. Shutdown also seems to work correctly.

## Release Plan

### DIFF
--- a/linera-client/src/benchmark.rs
+++ b/linera-client/src/benchmark.rs
@@ -1,0 +1,322 @@
+// Copyright (c) Zefchain Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+use std::{collections::HashMap, iter};
+
+use linera_base::{
+    crypto::{AccountPublicKey, AccountSecretKey},
+    data_types::{Amount, Timestamp},
+    hashed::Hashed,
+    identifiers::{AccountOwner, ApplicationId, ChainId, Owner},
+    listen_for_shutdown_signals,
+    time::Instant,
+};
+use linera_chain::{
+    data_types::{BlockProposal, ProposedBlock},
+    types::ConfirmedBlock,
+};
+use linera_core::{client::ChainClient, local_node::LocalNodeClient};
+use linera_execution::{
+    committee::{Committee, Epoch},
+    system::{Recipient, SystemOperation},
+    Operation,
+};
+use linera_rpc::node_provider::NodeProvider;
+use linera_sdk::abis::fungible;
+use linera_storage::Storage;
+use tokio::{runtime::Handle, task, time};
+use tokio_util::sync::CancellationToken;
+use tracing::{debug, info, warn, Instrument as _};
+
+use crate::Error;
+
+pub struct Benchmark<Storage>
+where
+    Storage: linera_storage::Storage,
+{
+    _phantom: std::marker::PhantomData<Storage>,
+}
+
+impl<S> Benchmark<S>
+where
+    S: Storage + Clone + Send + Sync + 'static,
+{
+    #[allow(clippy::too_many_arguments)]
+    pub async fn run_benchmark(
+        num_chains: usize,
+        transactions_per_block: usize,
+        bps: Option<usize>,
+        chain_clients: HashMap<ChainId, ChainClient<NodeProvider, S>>,
+        epoch: Epoch,
+        blocks_infos: Vec<(ChainId, Vec<Operation>, AccountSecretKey)>,
+        committee: Committee,
+        local_node: LocalNodeClient<S>,
+    ) -> Result<(), Error> {
+        let shutdown_notifier = CancellationToken::new();
+        tokio::spawn(listen_for_shutdown_signals(shutdown_notifier.clone()));
+
+        // The bps control task will control the BPS from the threads. `crossbeam_channel` is used
+        // for two reasons:
+        // 1. it allows bounded channels with zero sized buffers.
+        // 2. it blocks the current thread if the message can't be sent, which is exactly
+        //    what we want to happen here. `tokio::sync::mpsc` doesn't do that. `std::sync::mpsc`
+        //    does, but it is slower than `crossbeam_channel`.
+        // Number 1 is the main reason. `tokio::sync::mpsc` doesn't allow 0 sized buffers.
+        // With a channel with a buffer of size 1 or larger, even if we have already reached
+        // the desired BPS, the tasks would continue sending block proposals until the channel's
+        // buffer is filled, which would cause us to not properly control the BPS rate.
+        let (sender, receiver) = crossbeam_channel::bounded(0);
+        let bps_control_task = tokio::spawn(async move {
+            let mut recv_count = 0;
+            let mut start = time::Instant::now();
+            while let Ok(()) = receiver.recv() {
+                recv_count += 1;
+                if recv_count == num_chains {
+                    let elapsed = start.elapsed();
+                    if let Some(bps) = bps {
+                        if elapsed > time::Duration::from_secs(1) {
+                            warn!(
+                                "Failed to achieve {} BPS/{} TPS in {} ms",
+                                bps,
+                                bps * transactions_per_block,
+                                elapsed.as_millis(),
+                            );
+                        } else {
+                            time::sleep(time::Duration::from_secs(1) - elapsed).await;
+                            info!(
+                                "Achieved {} BPS/{} TPS in {} ms",
+                                bps,
+                                bps * transactions_per_block,
+                                elapsed.as_millis(),
+                            );
+                        }
+                    } else {
+                        let achieved_bps = num_chains as f64 / elapsed.as_secs_f64();
+                        info!(
+                            "Achieved {} BPS/{} TPS in {} ms",
+                            achieved_bps,
+                            achieved_bps * transactions_per_block as f64,
+                            elapsed.as_millis(),
+                        );
+                    }
+
+                    recv_count = 0;
+                    start = time::Instant::now();
+                }
+            }
+
+            info!("Exiting logging task...");
+        });
+
+        let mut bps_remainder = bps.unwrap_or_default() % num_chains;
+        let bps_share = bps.map(|bps| bps / num_chains);
+
+        let mut join_set = task::JoinSet::<Result<(), Error>>::new();
+        for (chain_id, operations, key_pair) in blocks_infos {
+            let bps_share = if bps_remainder > 0 {
+                bps_remainder -= 1;
+                Some(bps_share.unwrap() + 1)
+            } else {
+                bps_share
+            };
+
+            let shutdown_notifier = shutdown_notifier.clone();
+            let sender = sender.clone();
+            let handle = Handle::current();
+            let committee = committee.clone();
+            let local_node = local_node.clone();
+            let chain_client = chain_clients[&chain_id].clone();
+            let short_chain_id = format!("{:?}", chain_id);
+            join_set.spawn_blocking(move || {
+                handle.block_on(
+                    async move {
+                        Self::run_benchmark_internal(
+                            bps_share,
+                            operations,
+                            key_pair,
+                            epoch,
+                            chain_client,
+                            shutdown_notifier,
+                            sender,
+                            committee,
+                            local_node,
+                        )
+                        .await?;
+
+                        Ok(())
+                    }
+                    .instrument(tracing::info_span!(
+                        "benchmark_chain_id",
+                        chain_id = short_chain_id
+                    )),
+                )
+            });
+        }
+
+        join_set
+            .join_all()
+            .await
+            .into_iter()
+            .collect::<Result<Vec<_>, _>>()?;
+        drop(sender);
+        info!("All benchmark tasks completed");
+        bps_control_task.await?;
+
+        Ok(())
+    }
+
+    #[allow(clippy::too_many_arguments)]
+    async fn run_benchmark_internal(
+        bps: Option<usize>,
+        operations: Vec<Operation>,
+        key_pair: AccountSecretKey,
+        epoch: Epoch,
+        chain_client: ChainClient<NodeProvider, S>,
+        shutdown_notifier: CancellationToken,
+        sender: crossbeam_channel::Sender<()>,
+        committee: Committee,
+        local_node: LocalNodeClient<S>,
+    ) -> Result<(), Error> {
+        let chain_id = chain_client.chain_id();
+        info!(
+            "Starting benchmark at target BPS of {:?}, for chain {:?}",
+            bps, chain_id
+        );
+        let cross_chain_message_delivery = chain_client.options().cross_chain_message_delivery;
+        let mut num_sent_proposals = 0;
+        loop {
+            if shutdown_notifier.is_cancelled() {
+                info!("Shutdown signal received, stopping benchmark");
+                break;
+            }
+            let block = ProposedBlock {
+                epoch,
+                chain_id,
+                incoming_bundles: Vec::new(),
+                operations: operations.clone(),
+                previous_block_hash: chain_client.block_hash(),
+                height: chain_client.next_block_height(),
+                authenticated_signer: Some(Owner::from(key_pair.public())),
+                timestamp: chain_client.timestamp().max(Timestamp::now()),
+            };
+            let executed_block = local_node
+                .stage_block_execution(block.clone(), None)
+                .await?
+                .0;
+
+            let value = Hashed::new(ConfirmedBlock::new(executed_block));
+            let proposal =
+                BlockProposal::new_initial(linera_base::data_types::Round::Fast, block, &key_pair);
+
+            chain_client
+                .submit_block_proposal(&committee, Box::new(proposal), value)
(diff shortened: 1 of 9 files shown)


ndr-ds:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/3374?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#3374** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/3374?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/3374?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3367** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/3367?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

linera-service/src/linera/main.rs
```diff
@@ -806,31 +816,118 @@ impl Runnable for Job {
...
+                                    time::sleep(time::Duration::from_secs(1) - elapsed).await;
+                                    info!(
+                                        "Achieved {} BPS/{} TPS in {} ms",
+                                        bps.unwrap(),
+                                        bps.unwrap() * transactions_per_block,
+                                        elapsed.as_millis(),
+                                    );
+                                }
+                            } else {
+                                let achieved_bps = num_chains as f64 / elapsed.as_secs_f64();
```
afck, line 859: Why not `recv_count / elapsed`?

linera-service/src/linera/main.rs
```diff
@@ -806,31 +816,118 @@ impl Runnable for Job {
...
+                // buffer is filled, which would cause us to not properly control the BPS rate.
+                let (sender, receiver) = crossbeam_channel::bounded(0);
+                let bps_control_task = tokio::spawn(async move {
+                    let mut recv_count = 0;
+                    let mut start = time::Instant::now();
+                    while let Ok(()) = receiver.recv() {
+                        recv_count += 1;
+                        if recv_count == num_chains {
+                            let elapsed = start.elapsed();
+                            if bps.is_some() {
```
afck, line 841:
```suggestion
                            if let Some(bps) = bps {
```

Then we don't need the `unwrap`s.

linera-service/src/linera/main.rs
```diff
@@ -806,31 +816,118 @@ impl Runnable for Job {
...
                 }
+
+                join_set
+                    .join_all()
+                    .await
+                    .into_iter()
+                    .collect::<Result<Vec<_>, _>>()?;
+                drop(sender);
+                info!("All benchmark tasks completed");
+                bps_control_task.await?;
```
afck, line 930: We should extract all of this into a separate function.

linera-service/src/linera/main.rs
```diff
@@ -806,31 +816,118 @@ impl Runnable for Job {
...
+                        Some(bps_share.unwrap() + 1)
+                    } else {
+                        bps_share
+                    };
+
+                    let shutdown_notifier = shutdown_notifier.clone();
+                    let sender = sender.clone();
+                    let handle = Handle::current();
+                    let context = context.clone();
+                    let clients = clients.clone();
```
afck, line 893: Why clone all the clients if we only need one?

linera-service/src/linera/main.rs
```diff
@@ -806,31 +816,118 @@ impl Runnable for Job {
...
+                    let chain_client = chain_clients[&chain_id].clone();
+                    let short_chain_id = format!("{:?}", chain_id);
+                    join_set.spawn_blocking(move || {
+                        handle.block_on(
+                            async move {
+                                context
+                                    .run_benchmark(
+                                        bps_share,
```
afck, line 901: Do we really need this in addition to the BPS control task? Can't each chain task go as fast as possible because it will be blocked by the channel anyway if the desired BPS is reached?

ndr-ds: If I do it like this that means that the number of messages the BPS control task has to process is directly proportional to the BPS value, which is not ideal, specially when we start reaching really high BPS values. If each task has a BPS share it is supposed to achieve, and they only send a message when that share is achieved, then we have at most `num_chains` messages per second, always, and that can't really become a bottleneck so easily.

linera-service/src/linera/main.rs
ndr-ds: These clients are the validator clients, we need to send the block proposals to all validators, no?

linera-service/src/linera/main.rs
afck: I see; so each chain does at most one block per second?

linera-client/src/client_context.rs
```diff
@@ -838,35 +1030,19 @@ where
...
-            .join_all()
-            .await
-            .into_iter()
-            .collect::<Result<Vec<_>, _>>()?;
+        chain_client
+            .execute_operation(Operation::System(SystemOperation::CloseChain))
+            .await?
+            .expect("Close chain operation should not fail!");
 
         info!(
```
afck, line 1043: Do we really want to log this at INFO level? If we use hundreds of chains, this will make the log message that actually matters, the one about the BPS, hard to find.

ndr-ds: I can make it `debug!`

linera-client/src/client_context.rs
```diff
@@ -583,41 +587,255 @@ where
...
+                            time::sleep(time::Duration::from_secs(1) - elapsed).await;
+                            info!(
+                                "Achieved {} BPS/{} TPS in {} ms",
+                                bps,
+                                bps * transactions_per_block,
+                                elapsed.as_millis(),
+                            );
+                        }
+                    } else {
+                        let achieved_bps = num_chains as f64 / elapsed.as_secs_f64();
```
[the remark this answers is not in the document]
ndr-ds: In the previous behavior we created one block proposal per chain, and tried to send them all in parallel. This is exactly what we do here. We could have 1000 chains and send all 1000 block proposals in 0.3 seconds, which would give us 3k+ BPS.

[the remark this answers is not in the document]
afck: Sure, that's fine for now. But if it turns out that each chain gets its block proposal done in, say, less then 0.2 seconds, then we might as well make five proposals per chain, instead of requiring five times as many chains for a given BPS. Anyway, that can be done in a later PR. :+1:

linera-client/src/benchmark.rs
```diff
@@ -0,0 +1,342 @@
...
+                .join_all()
+                .await
+                .into_iter()
+                .collect::<Result<Vec<_>, _>>()?
+                .into_iter()
+                .filter_map(|response| {
+                    let vote = response.info.manager.pending?;
+                    vote.clone().with_value(value.clone())
+                })
+                .collect::<Vec<_>>();
```
afck, line 234: Doesn't this mean that we're always waiting for the slowest validator?

If we want to wait for only a quorum, we might as well just call `chain_client.execute_operations` directly.

linera-client/src/benchmark.rs
```diff
@@ -0,0 +1,342 @@
...
+                .await?;
+
+            num_sent_proposals += 1;
+            if let Some(bps) = bps {
+                if num_sent_proposals == bps {
+                    sender.send(())?;
+                    num_sent_proposals = 0;
+                }
+            } else {
+                break;
```
afck, line 271: Might as well `sender.send(())` here, then we don't need the `if` below.

linera-client/src/client_context.rs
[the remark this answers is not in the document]
afck: No, I was misunderstanding. (I had assumed omitting `--bps` would mean "go as fast as possible", but it actually means "make one block per chain and then quit".)

afck (approved):

linera-client/src/benchmark.rs
ndr-ds: Good point!

linera-service/src/linera/main.rs
afck: I can't imagine that a bunch of `()` messages could become a bottleneck… but I've been proven wrong before. 😅

afck (approved):

deuszx:
> as we don't do any mutable operations to either since we don't alter the wallet for benchmarks anymore.

If we don't do any mutable operations (don't update the wallet/state/db) is it an honest benchmark?

linera-client/src/client_context.rs
```diff
@@ -583,135 +574,73 @@ where
...
+    > {
+        let start = Instant::now();
+        // Below all block proposals are supposed to succeed without retries, we
+        // must make sure that all incoming payments have been accepted on-chain
+        // and that no validator is missing user certificates.
+        self.process_inboxes_and_force_validator_updates().await;
+        info!(
+            "Processed inboxes and forced validator updates in {} ms",
+            start.elapsed().as_millis()
+        );
```
deuszx, line 600: As a reader I'm slightly confused - the method is called `prepare_for_benchmark` but here we force the processing of inboxes and create blocks?

linera-client/src/client_context.rs
```diff
@@ -583,135 +574,73 @@ where
...
-                .collect::<Result<Vec<_>, _>>()?;
+        let start = Instant::now();
+        let (key_pairs, chain_clients) = self
+            .make_benchmark_chains(num_chains, tokens_per_chain)
+            .await?;
+        info!(
+            "Got {} chains in {} ms",
+            key_pairs.len(),
+            start.elapsed().as_millis()
+        );
```
deuszx, line 610: Why not `trace!`?

linera-client/src/client_context.rs
deuszx, line 600: Looks like a `TRACE`-level log.

linera-client/src/client_context.rs
```diff
@@ -583,135 +574,73 @@ where
...
+        let committee = committees
+            .get(&epoch)
+            .expect("current epoch should have a committee");
+        let blocks_infos = self.make_benchmark_block_info(
+            key_pairs,
+            transactions_per_block,
+            fungible_application_id,
+        );
+
+        Ok((chain_clients, epoch, blocks_infos, committee.clone()))
```
deuszx, line 640: This method is doing at least four things:
- clearing out inboxes, creating blocks
- creating new chains for benchmark
- supplying a test application with some tokens
- calling `make_benchmark_block_info` on a default chain
-

linera-core/src/client/mod.rs
```diff
@@ -1123,7 +1129,7 @@ where
 
     /// Broadcasts certified blocks to validators.
     #[instrument(level = "trace", skip(committee, delivery))]
-    async fn communicate_chain_updates(
+    pub async fn communicate_chain_updates(
```
deuszx, line 1132: If we add `pub` then they become part of the API contract and we no longer can change them.

linera-client/src/benchmark.rs
```diff
@@ -0,0 +1,259 @@
...
+        // 1. it allows bounded channels with zero sized buffers.
+        // 2. it blocks the current thread if the message can't be sent, which is exactly
+        //    what we want to happen here. `tokio::sync::mpsc` doesn't do that. `std::sync::mpsc`
+        //    does, but it is slower than `crossbeam_channel`.
+        // Number 1 is the main reason. `tokio::sync::mpsc` doesn't allow 0 sized buffers.
+        // With a channel with a buffer of size 1 or larger, even if we have already reached
+        // the desired BPS, the tasks would continue sending block proposals until the channel's
+        // buffer is filled, which would cause us to not properly control the BPS rate.
+        let (sender, receiver) = crossbeam_channel::bounded(0);
+        let bps_control_task = tokio::spawn(async move {
```
deuszx, line 68: It looks like `bps_control_task` is spawned on the same logical "level" as tasks that send the blocks. Personally I think it'd make more sense if the "control task" (supervisor) was spawning those tasks.

linera-client/src/benchmark.rs
```diff
@@ -0,0 +1,259 @@
...
+            let mut recv_count = 0;
+            let mut start = time::Instant::now();
+            while let Ok(()) = receiver.recv() {
+                recv_count += 1;
+                if recv_count == num_chains {
```
deuszx, line 73: It's hard to see that the code here is called just before finishing the benchmark. If the BPS control task spawned the "worker tasks" but itself stayed in the "main thread" it'd be more linear.

linera-service/src/linera/main.rs
```diff
@@ -749,87 +749,32 @@ impl Runnable for Job {
...
                 }
 
-                let default_chain_id = context
-                    .wallet
-                    .default_chain()
-                    .expect("should have default chain");
-                let default_chain_client = context.make_chain_client(default_chain_id)?;
-                let (epoch, committees) = default_chain_client
-                    .epoch_and_committees(default_chain_id)
+                let (chain_clients, epoch, blocks_infos, committee) = context
```
deuszx, line 758: I find it weird that benchmark load (`operations` etc) are generated by a client context. Why is it not the benchmark process itself deciding what and how the benchmark should look like?

ndr-ds:
@deuszx I think I didn't express myself well there. All I meant to say was that we used to alter the wallet, and now we don't. The only resource the different threads could race for is the wallet, but since we don't save the benchmark chains to the wallet anymore, that's no longer a concern. So using the same wallet for multiple threads is feasible, because the wallet is only used to create the chains at the start using its default chain, then the wallet isn't used by the threads anymore.

deuszx:
> @deuszx I think I didn't express myself well there. All I meant to say was that we used to alter the wallet, and now we don't. The only resource the different threads could race for is the wallet, but since we don't save the benchmark chains to the wallet anymore, that's no longer a concern. So using the same wallet for multiple threads is feasible, because the wallet is only used to create the chains at the start using its default chain, then the wallet isn't used by the threads anymore.

I see, thanks. It was my understanding then – I totally forgot that by _wallet_ we mean local wallet of the client who tracks chains.

linera-client/src/client_context.rs
ndr-ds: That's what we need to do to prepare for running the benchmark :)

ndr-ds: This is not production code, it's a benchmarking tool. I want to know how long things are running for without having to turn on trace or debug logs

linera-client/src/benchmark.rs
ndr-ds: The main task deals with awaiting the different spawned tasks. I don't see what the issue is. The control task controls the BPS, it is not supposed to create other things.

linera-client/src/benchmark.rs
ndr-ds: What you're describing only happens when we don't specify `--bps`. All senders will be closed, and this will exit

linera-service/src/linera/main.rs
ndr-ds: Makes sense, I can move the stuff that doesn't need the client context into `Benchmark` instead

linera-client/src/benchmark.rs
deuszx: It's the BPS task that is controlling when the job "is done" though. But OK, it doesn't necessarily mean it has to spawn the workers.

ndr-ds:
### Merge activity

* **Feb 21, 10:27 AM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/linera-io/linera-protocol/3374).

PR #3379, opened 2025-02-20
Adjust latency metrics buckets

ndr-ds:
## Motivation

We don't need to track latencies below 1ms IMHO. Also, for the server latency, I noticed it would very often just shoot to 50ms and stay there the whole time. If you're using few shards (1, like me, for example), shard latency can actually be fairly high.

## Proposal

Adjust things

## Test Plan

Check the changes on a local network

## Release Plan


afck (approved):

ndr-ds:
### Merge activity

* **Feb 20, 6:54 PM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/linera-io/linera-protocol/3379).

Issue #2362, opened 2024-08-08
update_execution_outcomes_with_app_registrations invalidates message IDs


afck:
[`update_execution_outcomes_with_app_registrations` inserts an outgoing message, moving existing messages and changing their message IDs](https://github.com/linera-io/linera-protocol/blob/0a2b0c10101671c2e5f224e61785f6971678a379/linera-execution/src/execution.rs#L229). This can be a problem if any following messages used their own message ID, e.g. for opening a chain or creating an application.

https://github.com/linera-io/linera-protocol/issues/2167 should make this unnecessary, so that this function can be removed.

bart-linera:
Closed by #3382

Issue #2167, opened 2024-06-21
Replace ApplicationDescription by blobs


afck:
Blobs make the `request-application` command unnecessary: Users should now be able to just start using an existing application on any chain:

_Creating_ an application should create a new application description blob, whose key is the application ID.

_Registering_ an application with a given ID on a chain involves reading the blob with that ID, and storing the description in the chain state's application registry. `SystemMessage::RegisterApplications` can be removed now, since this should happen whenever an application is first used on a chain.

(Registering an application still needs to recursively register all its dependencies, too.)

bart-linera:
Closed by #3382

PR #3382, opened 2025-02-20
Remove the need for `request_application`

bart-linera:
## Motivation

Until now, applications were only available on chains that published them and ones that explicitly requested to use them using `request_application`. This meant there was an awkward step for users that wanted to use others' applications on their own chains.

## Proposal

This PR changes `UserApplicationDescription` into a blob. Publishing an application means storing a blob, which is then accessible to all microchains right away. This removes the need for `request_application` and makes it possible to remove its functionality entirely.

## Test Plan

The example usage flow of the Request For Quotes application was tested manually in order to ensure that the RFQ app was available on other chains without it being explicitly requested.

The readme tests will ensure that this will be tested in CI as well.

## Release Plan

## Links

- #2167
- #2362  


linera-core/src/client/mod.rs
```diff
@@ -412,10 +412,17 @@ where
             }
             let mut result = self.handle_certificate(certificate.clone()).await;
 
-            if let Err(LocalNodeError::BlobsNotFound(blob_ids)) = &result {
+            let mut old_missing_blob_ids = vec![];
+            while let Err(LocalNodeError::BlobsNotFound(blob_ids)) = &result {
+                if blob_ids == &old_missing_blob_ids {
+                    // looks like we haven't been able to download any missing blobs
```
ma2bd, line 418: This is a little loose (in case there are lots of blobs pending) but ok for now.

linera-core/src/client/mod.rs
```diff
@@ -2198,12 +2205,19 @@ where
...
+        loop {
+            let result = self
+                .client
+                .local_node
+                .query_application(self.chain_id, query.clone())
+                .await;
+            if let Err(LocalNodeError::BlobsNotFound(blob_ids)) = &result {
+                self.receive_certificates_for_blobs(blob_ids.clone())
+                    .await?;
+                continue; // We found the missing blob: retry.
```
ma2bd, line 2217: should we give sometime?

linera-execution/src/applications.rs
```diff
@@ -1,167 +0,0 @@
-// Copyright (c) Zefchain Labs, Inc.
```
ma2bd, line 1: nice!

linera-execution/src/execution_state_actor.rs
```diff
@@ -79,10 +94,24 @@ where
...
         id: UserApplicationId,
+        txn_tracker: Option<&mut TransactionTracker>,
     ) -> Result<(UserServiceCode, UserApplicationDescription), ExecutionError> {
         #[cfg(with_metrics)]
         let _latency = LOAD_SERVICE_LATENCY.measure_latency();
-        let description = self.system.registry.describe_application(id).await?;
+        let blob_id = BlobId::new(
+            id.application_description_hash,
+            BlobType::ApplicationDescription,
+        );
```
ma2bd, line 104: nit: This could be a method of `UserApplicationId` I suppose.

linera-service/src/node_service.rs
```diff
@@ -1000,6 +1000,7 @@ where
             .map_err(|_| NodeServiceError::UnknownChainId {
                 chain_id: chain_id.to_string(),
             })?;
+        client.synchronize_from_validators().await?;
```
ma2bd, line 1003: Actually I don't see the race condition here. Maybe add a comment?

ma2bd (commented):
This looks great!

linera-core/src/client/mod.rs
bart-linera: It might not even be needed, strictly speaking. I added this when I thought some blobs weren't downloaded on the first pass, but I think eventually it turned out that the issue was elsewhere. Unless you think this is useful, I could revert this change and re-run the tests.

linera-core/src/client/mod.rs
bart-linera: I copied this approach from another method - but yes, I think it might be good to have some kind of, say, attempt counter and give up after too many attempts. Since we already did this elsewhere, maybe ticket it for another PR?

linera-execution/src/execution_state_actor.rs
bart-linera: Good idea :+1:

linera-service/src/node_service.rs
bart-linera: The issue was that if we tried to propose a block too soon after a cross-chain message was sent, we wouldn't include it in our block and a transaction could potentially fail. This happens in the tests where we transfer some tokens and then immediately try to transfer them from the target chain, so it's rather unlikely to happen in production. We could also fix this in the test code, but I opted for this approach, because while unlikely, it is not impossible in production. (Then again, in production we could just re-propose if we fail the first time... Not sure which approach is better.)

linera-core/src/client/mod.rs
ma2bd: yes totally. Maybe add a `TODO(#task)`

bart-linera: I started creating the issue, but I found myself hesitating if this needs changing. `receive_certificates_for_blobs` will either succeed, which means we won't be missing any blobs on the next pass, or it will fail, in which case the `?` operator will return the error. In either case, we should be looping at most twice, unless I'm missing something?

ma2bd (approved):
Thanks for the epic work!

linera-base/src/data_types.rs
```diff
@@ -784,13 +784,17 @@ pub enum OracleResponse {
 
 impl<'de> BcsHashable<'de> for OracleResponse {}
 
-/// Description of the necessary information to run a user application.
-#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Hash, Serialize)]
+/// Description of the necessary information to run a user application used within blobs.
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Hash, Serialize, WitType, WitStore)]
```
afck, line 788: Is this type exposed to contracts now?

linera-base/src/data_types.rs
```diff
@@ -801,10 +805,19 @@ pub struct UserApplicationDescription {
...
+            CryptoHash::new(&BlobContent::new_application_description(description)),
+            description.module_id,
+        )
+    }
+}
+
+impl BcsHashable<'_> for UserApplicationDescription {}
+
+impl UserApplicationDescription {
+    /// Gets the serialized bytes for this `BlobUserApplicationDescription`.
```
afck, line 818:
```suggestion
    /// Gets the serialized bytes for this `UserApplicationDescription`.
```

linera-base/src/identifiers.rs
```diff
@@ -749,58 +751,51 @@ impl<A> Copy for ApplicationId<A> {}
...
+
+impl<A> fmt::Debug for ApplicationId<A> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("ApplicationId")
+            .field(
+                "application_description_hash",
+                &self.application_description_hash,
+            )
+            .field("module_id", &self.module_id)
+            .finish()
```
afck, line 790: Isn't that exactly what `derive(Debug)` would do?

linera-base/src/identifiers.rs
```diff
@@ -865,10 +868,19 @@ impl<A> ApplicationId<A> {
...
     pub fn forget_abi(self) -> ApplicationId {
         ApplicationId {
+            application_description_hash: self.application_description_hash,
             module_id: self.module_id.forget_abi(),
-            creation: self.creation,
         }
     }
+
+    /// Converts the application ID to the ID of the blob containing the
+    /// UserApplicationDescription.
```
afck, line 877:
```suggestion
    /// [`UserApplicationDescription`].
```

linera-core/src/chain_worker/state/attempted_changes.rs
```diff
@@ -337,9 +338,10 @@ where
         );
 
         let required_blob_ids = executed_block.required_blob_ids();
```
afck, line 340: But that doesn't include the created blobs' IDs, does it? And we need to store those, too?

linera-core/src/client/mod.rs
afck: Yes, I'd revert this. If we don't manage to download them the first time we try, we should error.

linera-core/src/client/mod.rs
afck: No, you're right. We could replace the `loop` with an `if let`, but then we'd have to duplicate the `query_application` call.

linera-execution/src/execution.rs
```diff
@@ -69,24 +66,20 @@ where
...
             authenticated_caller_id: None,
-            height: application_description.creation.height,
+            height: application_description.block_height,
             round: None,
             index: Some(0),
         };
 
         let action = UserAction::Instantiate(context, instantiation_argument);
-        let next_message_index = application_description.creation.index + 1;
+        let next_message_index = application_description.application_index + 1;
```
afck, line 80: Would it make more sense to set this to 0? The next message index is different from the application index, isn't it?

linera-execution/src/execution.rs
```diff
@@ -251,74 +248,6 @@ where
...
-            ..RawExecutionOutcome::default()
-        };
-
-        // Insert the message before the first user outcome.
-        let index = results
-            .iter()
-            .position(|outcome| matches!(outcome, ExecutionOutcome::User(_, _)))
-            .unwrap_or(results.len());
-        // TODO(#2362): This inserts messages in front of existing ones, invalidating their IDs.
-        results.insert(index, ExecutionOutcome::System(system_outcome));
```
afck, line 317: Awesome! :tada: 
Don't forget to add https://github.com/linera-io/linera-protocol/issues/2362 to the issues that will be closed by this PR.

linera-execution/src/runtime.rs
```diff
@@ -401,10 +401,16 @@ impl SyncRuntimeInternal<UserContractInstance> {
             }
             #[cfg(not(web))]
             hash_map::Entry::Vacant(entry) => {
-                let (code, description) = self
+                let txn_tracker_moved = mem::take(&mut self.transaction_tracker);
```
afck, line 404: (We really need to find a better way to handle the transaction tracker. Not in this PR, of course!)

linera-execution/src/runtime.rs
```diff
@@ -1370,10 +1382,14 @@ impl ContractRuntime for ContractSyncRuntimeHandle {
                 };
                 let sender = this.execution_state_sender.clone();
 
+                let txn_tracker = TransactionTracker::default()
+                    .with_blobs(this.transaction_tracker.get_blobs_cache().clone());
```
afck, line 1386: (Is this so that we can query a service in the same block where we created the application?)

linera-execution/src/system.rs
```diff
@@ -646,37 +634,23 @@ where
...
+                    txn_tracker: txn_tracker_moved,
                 } = self
                     .create_application(
-                        next_message_id,
+                        context.chain_id,
+                        context.height,
                         module_id,
                         parameters,
                         required_application_ids,
+                        txn_tracker_moved,
```
afck, line 648: Would it be easier if `create_application` accepted a `&mut TransactionTracker` instead? Then we wouldn't need the `mem::take` here?

linera-execution/src/system.rs
```diff
@@ -1028,59 +980,157 @@ where
...
-            module_id,
-            creation: next_message_id,
-        };
-        let mut blobs_to_register = vec![];
-        for application in required_application_ids.iter().chain(iter::once(&id)) {
+        let application_index = txn_tracker.next_application_index();
+        for application_module in required_application_ids
+            .iter()
+            .map(|app_id| &app_id.module_id)
+            .chain(iter::once(&module_id))
```
afck, line 994: That misses indirect dependencies, does it?

linera-execution/src/system.rs
```diff
@@ -423,6 +409,8 @@ pub enum SystemExecutionError {
     OracleResponseMismatch,
     #[error("No recorded response for oracle query")]
     MissingOracleResponse,
+    #[error("Invalid index for operation: {0:?}")]
+    InvalidOperationIndex(Option<u32>),
```
afck, line 413: Where is that used?

linera-execution/src/system.rs
```diff
@@ -1028,59 +980,157 @@ where
...
+            txn_tracker,
         })
     }
 
+    async fn check_required_applications(
+        &mut self,
+        application_description: &UserApplicationDescription,
+        mut txn_tracker: Option<&mut TransactionTracker>,
+    ) -> Result<(), SystemExecutionError> {
+        // Make sure that referenced applications ids have been registered.
```
afck, line 1030:
```suggestion
        // Make sure that referenced applications IDs have been registered.
```

linera-execution/src/system.rs
```diff
@@ -1028,59 +980,157 @@ where
...
+    }
+
+    /// Retrieves an application's description.
+    pub async fn describe_application(
+        &mut self,
+        id: UserApplicationId,
+        mut txn_tracker: Option<&mut TransactionTracker>,
+    ) -> Result<UserApplicationDescription, SystemExecutionError> {
+        let blob_id = id.description_blob_id();
+        let blob_content = self.read_blob_content(blob_id).await?;
```
afck, line 1044: Should this check the blob cache first, so it already works in the same block?

linera-execution/src/system.rs
```diff
@@ -1028,59 +980,157 @@ where
...
+        self.blob_used(txn_tracker.as_deref_mut(), service_bytecode_blob_id)
+            .await?;
+
+        self.check_required_applications(&description, txn_tracker)
+            .await?;
+
+        Ok(description)
+    }
+
+    /// Retrieves the recursive dependencies of applications and apply a topological sort.
```
afck, line 1063:
```suggestion
    /// Retrieves the recursive dependencies of applications and applies a topological sort.
```

linera-execution/src/system.rs
```diff
@@ -1028,59 +980,157 @@ where
...
+                if !seen.contains(child) {
+                    stack.push(*child);
+                }
+            }
+        }
+        Ok(result)
+    }
+
+    /// Retrieves applications' descriptions preceded by their recursive dependencies.
+    pub async fn describe_applications_with_dependencies(
```
afck, line 1103: Is that still used?

linera-execution/tests/test_execution.rs
```diff
@@ -1313,7 +1393,19 @@ async fn test_multiple_messages_from_different_applications() -> anyhow::Result<
...
+        0,
+        0,
+        Some(
+            caller_blobs
+                .iter()
+                .chain(silent_blobs.iter())
+                .chain(sending_blobs.iter())
+                .copied()
+                .map(OracleResponse::Blob)
+                .collect(),
```
afck, line 1406: This is repeated a lot. Maybe some helper method that turns an iterator of blobs into a vector of oracle responses would be nice.

linera-execution/src/transaction_tracker.rs
```diff
@@ -106,6 +128,14 @@ impl TransactionTracker {
         });
     }
 
+    pub fn add_created_blob(&mut self, blob: Blob) {
+        self.blobs.insert(blob.id(), blob);
+    }
+
+    pub fn get_blobs_cache(&self) -> &BTreeMap<BlobId, Blob> {
```
afck, line 135: Or just `blobs_cache`, `blobs` or `created_blobs`?

afck (approved):
Looks great! :tada:

linera-base/src/data_types.rs
bart-linera: Good question. Doesn't look like it - removing this doesn't break anything. Must have been needed for something I reverted later.

linera-base/src/identifiers.rs
bart-linera: Hm. I thought I had some issues with just deriving `Debug`, because I don't think I'd have done this otherwise... but it seems that it works now? So I'll just switch to deriving :+1:

linera-core/src/chain_worker/state/attempted_changes.rs
bart-linera: I had to double check, but it looks like t does? :sweat_smile:  Only `BlockProposal::required_blob_ids` does not include the created blobs, because at that stage they don't even exist yet.

linera-execution/src/execution.rs
bart-linera: This is in test code, so it doesn't matter much, anyway, I think? But it probably would make sense to change it to 0 :+1:

linera-execution/src/runtime.rs
bart-linera: Yes, exactly :+1:

linera-execution/src/system.rs
bart-linera: In a way, yes. But it would make it harder to handle in `ContractSyncRuntimeHandle::create_application`. There, we need to take the tracker and send it to the execution worker, and it needs to return it back to us. If we made `create_application` take a reference, it couldn't return the modified tracker in the result, which would complicate the code there.

linera-execution/src/system.rs
bart-linera: This part does, yes. But then `check_required_applications` calls `describe_application` for each dependency, and `describe_application` calls `check_required_applications`.

In fact, I think some checking is duplicated here - we only need to check the created application's bytecodes at this point, because the dependencies are taken care of by `check_required_applications`.

linera-execution/src/system.rs
bart-linera: Ah, I probably added it as a part of an earlier approach :sweat_smile: I'll remove it.

linera-core/src/chain_worker/state/attempted_changes.rs
afck: Right. Maybe the two functions shouldn't have the same name anymore then.

linera-execution/src/system.rs
bart-linera: `TransactionTracker` is per transaction, anyway... but maybe it still should, so that if an operation creates multiple applications (which is possible now that `create_application` is exposed to contracts), it still works :+1:

linera-execution/src/system.rs
bart-linera: Huh. I thought it was, but apparently it isn't. Good catch :+1:

bart-linera: But actually... Maybe `describe_application` recursively checking dependencies isn't correct? It may be a waste most of the time. So maybe we would want not to check dependencies there, but use this function in `create_application`?

linera-core/src/chain_worker/state/attempted_changes.rs
bart-linera: I'm not sure what a better name could be. Especially since `Block` has a method with the same name, which returns the same blobs, and they _are_ required in a sense there (as in, when you process a `Block` and don't have those blobs, you need to get them from somewhere).

Here they are the blobs that need to be written... but I'm not sure if `blobs_to_be_written` is a better name?

linera-execution/src/system.rs
afck: I think we want to make sure that once we start using an application ~on one chain~—or rather: on one _client_—, we also have all its dependencies (as blobs), even if the application wasn't created there.

But the whole application dependencies topic also needs further discussion anyway, so maybe not even this is a blocker for this PR.

linera-base/src/identifiers.rs
bart-linera: Turns out `UserApplicationDescription` is not in scope in that file, and this broke `cargo doc` :grimacing: I removed the link (brackets), but left the backticks for formatting.

PR #3380, opened 2025-02-20
Adjust dashboards to show 1m metrics

ndr-ds:
## Motivation

`1h` metrics honestly don't make much sense, because p99 for example could take forever to go down, even though the network is fine.

## Proposal

`1m` seems more reasonable.

## Test Plan

Deploy a local network and check

## Release Plan

- Nothing to do / These changes follow the usual release cycle.

PR #3384, opened 2025-02-21
Use linear buckets in some places

ndr-ds:
## Motivation

We currently use exponential buckets everywhere. As much as they're good in the sense that they generate less buckets, and can be cheaper on Grafana Cloud, etc, they make our buckets super wide, which makes our Prometheus data less accurate.

## Proposal

I need to spend some time later looking at the `testnet` data for these different metrics to adjust the buckets, but for not just changing proxy/server latencies to use linear buckets instead.
I also changed the default starting value to 0.001 as 1 microsecond should be enough for most of what we're measuring, and will take at least one bucket away from metrics using this.

## Test Plan

Run a validator locally, see the metrics exported with the new buckets

## Release Plan

### DIFF
--- a/linera-base/src/data_types.rs
+++ b/linera-base/src/data_types.rs
@@ -27,7 +27,9 @@ use serde::{Deserialize, Deserializer, Serialize, Serializer};
 use thiserror::Error;
 
 #[cfg(with_metrics)]
-use crate::prometheus_util::{bucket_latencies, register_histogram_vec, MeasureLatency};
+use crate::prometheus_util::{
+    exponential_bucket_latencies, register_histogram_vec, MeasureLatency,
+};
 use crate::{
     crypto::{BcsHashable, CryptoHash},
     doc_scalar, hex_debug, http,
@@ -1168,7 +1170,7 @@ static BYTECODE_COMPRESSION_LATENCY: LazyLock<HistogramVec> = LazyLock::new(|| {
         "bytecode_compression_latency",
         "Bytecode compression latency",
         &[],
-        bucket_latencies(10.0),
+        exponential_bucket_latencies(10.0),
     )
 });
 
@@ -1179,7 +1181,7 @@ static BYTECODE_DECOMPRESSION_LATENCY: LazyLock<HistogramVec> = LazyLock::new(||
         "bytecode_decompression_latency",
         "Bytecode decompression latency",
         &[],
-        bucket_latencies(10.0),
+        exponential_bucket_latencies(10.0),
     )
 });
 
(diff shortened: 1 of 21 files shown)


ndr-ds:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/3384?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#3384** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/3384?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/3384?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3374** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/3374?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3380** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/3380?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3379** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/3379?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

ndr-ds:
Why is exponential better for latency?

MathieuDutSik:
> Why is exponential better for latency? It gives us less accurate latency values, as I explained in the PR description

I thought the latency was more in an exponential curves, but if you see it differently fine.
But what about the number of buckets? Could we have too much data in the Prometheus.

If it were up to me, I would remove the buckets altogether. I see it mostly as something useful for presentation.

ndr-ds:
So the issue here is that the wider our buckets, the less accurate our quantiles will be (p50, p90, p99, etc), which is not good, because that's what we'll be looking at in our dashboards to monitor the system. So we need to make our buckets as narrow as possible, without generating too many buckets, which can be expensive when using Grafana Cloud, for example. These were starting at 0.0001 ms and it had a base 3 exponential growth. Now it starts at 1 ms, and grows linearly for both the `proxy` and `server` latencies. We're probably not interested in latencies below 1 ms for these, so this is fine. This will generate less buckets than before, and the buckets will be narrower so our data will be more accurate.
So this is a win on every aspect IMHO.

MathieuDutSik (approved):
I approve because there is less buckets with this system than the preceding.

ndr-ds:
### Merge activity

* **Feb 24, 11:56 AM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/linera-io/linera-protocol/3384).

PR #3385, opened 2025-02-21
Create command `request-chain` to allow using testnets in bash

ma2bd:
## Motivation

* Allow README files to run on the testnet
* Do not hardcode chain IDs

## Proposal

* Create a command `request-chain`
* Use it in the main README file
* Also create a comment `follow-chain` while we're at it

## Test Plan

Tested new CLI commands manually (together with `forget-keys` and `forget-chain`) + CI

## Release Plan

These changes COULD be backported to the latest `testnet` branch, then be released in a new SDK.

### DIFF
--- a/CLI.md
+++ b/CLI.md
@@ -40,6 +40,8 @@ This document contains the help content for the `linera` command-line program.
 * [`linera wallet show`↴](#linera-wallet-show)
 * [`linera wallet set-default`↴](#linera-wallet-set-default)
 * [`linera wallet init`↴](#linera-wallet-init)
+* [`linera wallet request-chain`↴](#linera-wallet-request-chain)
+* [`linera wallet follow-chain`↴](#linera-wallet-follow-chain)
 * [`linera wallet forget-keys`↴](#linera-wallet-forget-keys)
 * [`linera wallet forget-chain`↴](#linera-wallet-forget-chain)
 * [`linera project`↴](#linera-project)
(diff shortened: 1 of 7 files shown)


linera-client/src/client_options.rs
```diff
@@ -1199,7 +1199,25 @@ pub enum WalletCommand {
...
+        /// The address of a faucet.
+        #[arg(long)]
+        faucet: String,
+
+        /// Whether this chain should become the default chain.
+        #[arg(long)]
+        set_default: bool,
+    },
+
+    /// Add a new followed chain (a chain without keypair) to the follow.
```
afck, line 1213:
```suggestion
    /// Add a new followed chain (a chain without keypair) to follow.
```

afck (approved):
Good idea! :+1:

deuszx:
We need the same for owners as well.

linera-client/src/client_options.rs
sam31101975: good

CLI.md
```diff
@@ -794,9 +798,34 @@ Initialize a wallet from the genesis configuration
 
 
 
+## `linera wallet request-chain`
+
+Request a chain from a faucet and add it to the wallet
```
deuszx, line 803: I'd be more precise here - we're requesting _creation_ of a new chain from the wallet.

ma2bd:
> We need the same for owners as well - everything that has hash hardcoded in READMEs.

Yes. Everytime we generate a key pair, we should print the owner/address on stdout.

PR #3386, opened 2025-02-21
Make Evm a part of the supported virtual machines.

MathieuDutSik:
## Motivation

We extend here the support for EVM. This means extending the `BytecodeId` by adding the `VmRuntime`to it.

## Proposal

The following changes have been done:
* A type `VmRuntime` has been introduced that covers both `Wasm` (Native linera) and `Evm`.
* The `VmRuntime` is added to the `BytecodeId`. This is a reasonable way to make the VM a part of the type description.
* We cannot put the `VmRuntime` in the `UserApplicationDescription` since this forces the `create_application` to have a `VmRuntime` type. While innocent, this changes is not adequate since the `create_application` is used in smart contract code and we want to avoid this part to seep into the smart contract code.
* The `VmRuntime,` does not  depend on any feature. This is because they have to pass the wit barrier and the wit-types cannot depend on features.
* The `VmRuntime` has been added to the `linera-base`. This is unavoidable because `BytecodeId` depends on it.
* The `load_contract/load_service` is simplified into one single code that panics when relevant features have not been installed.
* The `--vm-runtime` is added to the function calls like `publish_application`. When missing it defaults to `VmRuntime::Wasm`.
* A typo in `contract_runtime_apis.rs` for `ApplicationDescription` is corrected.

## Test Plan

The CI. No test has been added to test the functionality.

## Release Plan

Normal release plan.

## Links

None.


CLI.md
```diff
@@ -117,7 +117,6 @@ A Byzantine-fault tolerant sidechain with low-latency finality and high throughp
 * `--max-pending-message-bundles <MAX_PENDING_MESSAGE_BUNDLES>` — The maximum number of incoming message bundles to include in a block proposal
 
   Default value: `10`
-* `--wasm-runtime <WASM_RUNTIME>` — The WebAssembly runtime to use
```
ma2bd, line 120: @MathieuDutSik We said that we are keeping this

linera-execution/src/test_utils/mod.rs
```diff
@@ -47,9 +48,14 @@ pub fn create_dummy_user_application_description(
         compressed_bytes: b"service".to_vec(),
     });
 
+    let vm_runtime = VmRuntime::default();
```
ma2bd, line 51: nit: just inline the value

linera-base/src/identifiers.rs
```diff
@@ -350,6 +351,8 @@ pub struct BytecodeId<Abi = (), Parameters = (), InstantiationArgument = ()> {
     pub contract_blob_hash: CryptoHash,
     /// The hash of the blob containing the service bytecode.
     pub service_blob_hash: CryptoHash,
```
ma2bd, line 353: @afck @jvff Should we make the service blob optional?

linera-sdk/wit/service-system-api.wit
```diff
@@ -41,6 +41,7 @@ interface service-system-api {
     record bytecode-id {
         contract-blob-hash: crypto-hash,
         service-blob-hash: crypto-hash,
+        vm-runtime: vm-runtime,
```
ma2bd, line 44: @afck @jvff I'm worried that we expose too much detail about BytecodeID (and generally applications) to the SDK. Then, we will have to maintain them for ever.

Perhaps, this BytecodeID could be a hexadecimal string. Alternatively, we could also remove `create_application` for now.

linera-client/src/client_options.rs
```diff
@@ -773,6 +774,10 @@ pub enum ClientCommand {
         /// Path to the Wasm file for the application "service" bytecode.
         service: PathBuf,
 
+        /// The virtual machine runtime to use.
+        #[arg(long)]
+        vm_runtime: Option<VmRuntime>,
```
ma2bd, line 779: nit: We could also handle the default here.

linera-execution/src/lib.rs
```diff
@@ -1344,19 +1344,6 @@ pub trait WithWasmDefault {
 }
 
 impl WasmRuntime {
-    #[cfg(with_wasm_runtime)]
-    pub fn default_with_sanitizer() -> Self {
```
ma2bd, line 1348: Why? Is this unused?

ma2bd:
@MathieuDutSik LGTM. I believe you need to update the summary.

linera-execution/src/lib.rs
MathieuDutSik: Yes.

linera-base/src/identifiers.rs
MathieuDutSik: Maybe, if so it would have to be in a separate PR I think.

linera-client/src/client_options.rs
MathieuDutSik: You are correct, thank you.

linera-sdk/wit/service-system-api.wit
MathieuDutSik: Maybe we could change. But in another PR.

linera-execution/src/test_utils/mod.rs
MathieuDutSik: Ok, done.

linera-sdk/wit/service-system-api.wit
jvff: I agree that it's not ideal and will make adding new runtimes harder because it will always require changing the WIT interface.

I'm starting to lean more towards having a single bytecode blob with metadata including the VM runtime and pointing to the actual blobs to form the bytecode.

But yeah, it doesn't have to be done in this PR.

linera-storage/src/lib.rs
```diff
@@ -278,41 +280,57 @@ pub trait Storage: Sized {
...
+                cfg_if::cfg_if! {
+                    if #[cfg(with_revm)] {
+                        let evm_runtime = EvmRuntime::Revm;
+                        Ok(EvmContractModule::new(_contract_bytecode, evm_runtime)
+                           .await?
+                           .into())
+                    } else {
+                        panic!(
+                            "An Evm runtime is required to load user applications. \
+                             Please enable the `revm` feature flags \
```
jvff, line 319: Nit: typo

```suggestion
                             Please enable the `revm` feature flag \
```

linera-storage/src/lib.rs
```diff
@@ -321,28 +339,49 @@ pub trait Storage: Sized {
...
+                    }
+                }
+            }
+            VmRuntime::Evm => {
+                cfg_if::cfg_if! {
+                    if #[cfg(with_revm)] {
+                        let evm_runtime = EvmRuntime::Revm;
+                        Ok(EvmServiceModule::new(_service_bytecode, evm_runtime)
+                           .await?
+                           .into())
```
jvff, line 374: I'm not really sure what it means to have an Evm service. We could either return an error saying Evm services are not supported, or figure out a way to query the Evm state, either exposing some read-only entrypoints or a single generic service to inspect the state.

linera-storage/src/lib.rs
```diff
@@ -321,28 +339,49 @@ pub trait Storage: Sized {
...
+                    } else {
+                        panic!(
+                            "An Evm runtime is required to load user applications. \
+                             Please enable the `revm` feature flags \
```
jvff, line 378: Nit: typo

```suggestion
                             Please enable the `revm` feature flag \
```

linera-storage/src/lib.rs
```diff
@@ -278,41 +280,57 @@ pub trait Storage: Sized {
         let compressed_contract_bytecode = CompressedBytecode {
             compressed_bytes: contract_blob.into_bytes().to_vec(),
         };
-        let contract_bytecode =
+        let _contract_bytecode =
```
jvff, line 283: Could we add an attribute instead of the underscore. The pattern is to add the underscore to unused bindings, but in this case it might be used.

```suggestion
        #[cfg_attr(not(any(with_wasm_runtime, with_revm)), allow(unused_variables))]
        let contract_bytecode =
```

linera-storage/src/lib.rs
```diff
@@ -321,28 +339,49 @@ pub trait Storage: Sized {
...
-    async fn load_service(
-        &self,
-        _application_description: &UserApplicationDescription,
-    ) -> Result<UserServiceCode, ExecutionError> {
-        panic!(
-            "A Wasm runtime is required to load user applications. \
-            Please enable the `wasmer` or the `wasmtime` feature flags \
-            when compiling `linera-storage`."
-        );
+        let _service_bytecode =
```
jvff, line 342: Same here:

```suggestion
        #[cfg_attr(not(any(with_wasm_runtime, with_revm)), allow(unused_variables))]
        let service_bytecode =
```

jvff (approved):

linera-base/src/identifiers.rs
```diff
@@ -674,10 +690,15 @@ impl<'de, Abi, Parameters, InstantiationArgument> Deserialize<'de>
 
 impl BytecodeId {
     /// Creates a bytecode ID from contract/service hashes.
-    pub fn new(contract_blob_hash: CryptoHash, service_blob_hash: CryptoHash) -> Self {
+    pub fn new(
```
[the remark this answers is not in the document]
MathieuDutSik: Ok, thank you.

linera-base/src/vm.rs
```diff
@@ -0,0 +1,58 @@
...
+    Evm,
+}
+
+impl FromStr for VmRuntime {
+    type Err = InvalidVmRuntime;
+
+    fn from_str(string: &str) -> Result<Self, Self::Err> {
+        match string {
+            "wasm" => Ok(VmRuntime::Wasm),
+            "revm" => Ok(VmRuntime::Evm),
```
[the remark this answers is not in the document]
MathieuDutSik: Yes, correct.

linera-storage/src/lib.rs
MathieuDutSik: Thank you.

linera-base/src/identifiers.rs
ma2bd: Ok personally I was upvoting the enum where each VM variant can have a variable number of blobs. (I don't how easy it is to support, though)

MathieuDutSik: Currently, we have two bytecodes in the EVM for service and contract which are identical.
We have discussed merging contract/service for regular Linera applications. This may be an impetus for doing that.

linera-storage/src/lib.rs
MathieuDutSik: Thanks.

linera-storage/src/lib.rs
MathieuDutSik: What it means to have service means to call the smart contract with the contract being executed without the result being written to the blockchain.

The corresponding function in the Ethereum blockchain is `eth_call`.

linera-storage/src/lib.rs
MathieuDutSik: Ok, corrected.

linera-base/src/vm.rs
```diff
@@ -0,0 +1,58 @@
...
+    Deserialize,
+    WitType,
+    WitStore,
+    WitLoad,
+    Debug,
+)]
+#[cfg_attr(with_testing, derive(test_strategy::Arbitrary))]
+/// The virtual machine runtime
+pub enum VmRuntime {
+    /// The Wasm for the virtual machine
```
ma2bd, line 34: drop "for the" --> "The Wasm virtual machine"
alos below

linera-sdk/wit/base-runtime-api.wit
```diff
@@ -64,10 +64,11 @@ interface base-runtime-api {
         height: block-height,
         index: u32,
     }
-    
+
     record bytecode-id {
         contract-blob-hash: crypto-hash,
         service-blob-hash: crypto-hash,
+        vm-runtime: vm-runtime,
     }
```
ma2bd, line 73: @afck @jvff The coupling between the SDK and the details of runtime VMs worries me a little. Could we use abstract bytes for BytecodeId inside the SDK.

linera-base/src/vm.rs
MathieuDutSik: Ok.

linera-base/src/vm.rs
```diff
@@ -0,0 +1,58 @@
...
+            "wasm" => Ok(VmRuntime::Wasm),
+            "evm" => Ok(VmRuntime::Evm),
+            unknown => Err(InvalidVmRuntime(unknown.to_owned())),
+        }
+    }
+}
+
+scalar!(VmRuntime);
+
+/// Attempts to create an invalid [`VmRuntime`] instance from a string.
```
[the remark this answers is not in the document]
MathieuDutSik: Ok.

ma2bd (approved):
Please update the summary

PR #3397, opened 2025-02-23
generalize --with-wallet N to --with-wallet ALPHANUM

ma2bd:
## Motivation

After #3393, wallet names do not need to be numbers

## Proposal

Support alphanumerical wallet names

## Test Plan

CI


Twey (approved):
I'd still like this to be a path rather than a substring of an environment variable — but no problems with this as an interim step

ma2bd:
> I'd still like this to be a path rather than a substring of an environment variable — but no problems with this as an interim step

Yes, that's a possible end result depending how the next refactoring works out

PR #3400, opened 2025-02-23
Make storage configs clonable and serializable

ma2bd:
## Motivation

We'd like to simplify config management from databases.

## Proposal

The first step is to make storage config objects serializable.

## Test Plan

CI

### DIFF
--- a/linera-views/src/backends/dynamo_db.rs
+++ b/linera-views/src/backends/dynamo_db.rs
@@ -36,6 +36,7 @@ use aws_sdk_dynamodb::{
 use aws_smithy_types::error::operation::BuildError;
 use futures::future::{join_all, FutureExt as _};
 use linera_base::ensure;
+use serde::{Deserialize, Serialize};
 use thiserror::Error;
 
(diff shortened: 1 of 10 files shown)


linera-views/src/backends/dynamo_db.rs
```diff
@@ -334,15 +321,26 @@ pub struct DynamoDbStoreInternal {
...
 }
 
-/// The initial configuration of the system
-#[derive(Debug)]
+/// The initial configuration of the system.
+#[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct DynamoDbStoreInternalConfig {
-    /// The AWS configuration
-    config: aws_sdk_dynamodb::Config,
+    /// Whether to use local stack or not.
```
afck, line 327:
```suggestion
    /// Whether to use LocalStack or not.
```

afck (approved):

MathieuDutSik (approved):
Thank you for implementing this.

PR #3401, opened 2025-02-24
linera windows

usagi32:
## Motivation
Helps with compiling linera on windows.


usagi32:
As I use windows. Always had to do some hacky things to get linera to compile. This will be helpful for me.

ma2bd (approved):

PR #3363, opened 2025-02-18
Make sure we include end_value in generated Prometheus buckets

ndr-ds:
## Motivation

Right now we're generating the buckets, but it's not in an inclusive way, and it should be. For latency metrics, for example, you have `end_value` being 500ms, but one of the buckets is 159. Since we're using a multiplication factor of 3, the next bucket would go over 500ms, so we don't include it...

## Proposal

Make sure we always include the `end_value` bucket

## Test Plan

CI + start a network locally and see the correct buckets exported

## Release Plan

- Nothing to do / These changes follow the usual release cycle.

PR #2878, opened 2024-11-13
Simplify the Prometheus stuff

MathieuDutSik:
## Motivation

The Prometheus entries is creating a lot of copy-pasted code which is bad taste. 

## Proposal

This PR does two things:
* Change the use statements which lead to smaller code.
* Introduce functions `bucket_latencies` and `bucket_interval` for creating the bucket of values in a more organized way.

## Test Plan

The CI

## Release Plan

The intervals have changed a bit. So, if the operation code strictly depends on the exact values of the buckets then
something else would have to be done.

## Links

None.

PR #3405, opened 2025-02-24
(Port of #3363) Make sure we include end_value in generated Prometheus buckets

ndr-ds:
## Motivation

Right now we're generating the buckets, but it's not in an inclusive way, and it should be. For latency metrics, for example, you have `end_value` being 500ms, but one of the buckets is 159. Since we're using a multiplication factor of 3, the next bucket would go over 500ms, so we don't include it...

## Proposal

Make sure we always include the `end_value` bucket

## Test Plan

CI + start a network locally and see the correct buckets exported

## Release Plan

### DIFF
--- a/linera-base/src/prometheus_util.rs
+++ b/linera-base/src/prometheus_util.rs
@@ -4,8 +4,8 @@
 //! This module defines util functions for interacting with Prometheus (logging metrics, etc)
 
 use prometheus::{
-    histogram_opts, register_histogram_vec, register_int_counter_vec, Error, HistogramVec,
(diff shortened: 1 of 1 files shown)


ndr-ds:
A lot of changes happened in this prometheus code that we didn't backport (mainly https://github.com/linera-io/linera-protocol/pull/2878). Backporting it would be a huge pain, so I'm just gonna change the buckets manually for now, and when we cut a new testnet we'll have the proper stuff in

ndr-ds:
https://github.com/linera-io/linera-protocol/pull/3406

PR #3407, opened 2025-02-25
Replace rust_secp256k1 with k256

deuszx:
## Motivation

`rust_secp256k1` did not work well with Wasm.

## Proposal

Replace with `k256`.

## Test Plan

CI should catch any regressions and @Twey will check if this indeed works with web client.

## Release Plan

## Links


.github/workflows/rust.yml
```diff
@@ -289,7 +289,7 @@ jobs:
     - uses: actions-rust-lang/setup-rust-toolchain@v1
     - name: Install cargo-machete
       run: |
-        cargo install cargo-machete --locked
+        cargo install cargo-machete@0.7.0 --locked
```
deuszx, line 292: There's a new version of cargo-machete (0.8.0) which migrated to Rust edition2024.

linera-base/src/crypto/secp256k1.rs
```diff
@@ -34,17 +45,33 @@ pub struct Secp256k1KeyPair {
...
+
+    /// Decodes the bytes into the public key.
+    /// Expects the bytes to be of compressed representation.
+    ///
+    /// Panics if the encoding can't be done in a constant time.
+    pub fn from_bytes(bytes: &[u8]) -> Self {
+        Self(
+            k256::PublicKey::from_encoded_point(&EncodedPoint::from_bytes(bytes).unwrap())
+                .expect("Decode in constant time.")
+                .into(),
```
deuszx, line 73: TODO @deuszx : fix this.

linera-base/src/crypto/secp256k1.rs
```diff
@@ -119,30 +143,31 @@ impl FromStr for Secp256k1PublicKey {
...
 }
 
 impl TryFrom<&[u8]> for Secp256k1PublicKey {
     type Error = CryptoError;
 
     fn try_from(value: &[u8]) -> Result<Self, Self::Error> {
-        let pk = secp256k1::PublicKey::from_slice(value).map_err(CryptoError::Secp256k1Error)?;
-        Ok(Secp256k1PublicKey(pk))
+        let pk = k256::PublicKey::from_encoded_point(&EncodedPoint::from_bytes(value).unwrap())
+            .expect("Decode in constant time.");
```
deuszx, line 155: TODO @deuszx : use Self::from_bytes.

.github/workflows/rust.yml
Twey: I'm going to make a separate PR for this just because this one will end up squashed — so it would be really nice to be able to track/revert/etc this change in a separate commit.

linera-base/src/crypto/secp256k1.rs
Twey: If this TODO (and others) isn't to be fixed in the PR, can we get an issue + comment in the code?

.github/workflows/rust.yml
Twey: Wait shouldn't `--locked` take care of this?  Oh well, no harm done :sweat_smile:

linera-base/src/crypto/secp256k1.rs
deuszx: I'm working on these right now. Apologies for confusion.

linera-base/src/crypto/secp256k1.rs
```diff
@@ -187,9 +213,8 @@ impl Secp256k1Signature {
     where
         T: BcsSignable<'de>,
     {
-        let secp = secp256k1::Secp256k1::signing_only();
-        let message = Message::from_digest(CryptoHash::new(value).as_bytes().0);
-        let signature = secp.sign_ecdsa(&message, &secret.0);
+        let prehash = CryptoHash::new(value).as_bytes().0;
+        let (signature, _rid) = secret.0.sign_prehash_recoverable(&prehash).unwrap();
```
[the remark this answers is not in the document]
deuszx: Working on this right now. Simply changing this to return `Result<Self, Error>` affects a lot of the code - `LiveVote`/`Vote` creation etc.

Cargo.toml
```diff
@@ -122,6 +122,9 @@ rand = { version = "0.8.5", default-features = false }
 rand_chacha = { version = "0.3.1", default-features = false }
 rand_distr = { version = "0.4.3", default-features = false }
 ruzstd = "0.7.1"
+k256 = { version = "0.13.4", default-features = false, features = [ 
+    "ecdsa", "pem", "sha256", "serde", "arithmetic" 
+] }
```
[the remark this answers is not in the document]
deuszx: That's how it's done for `reqwest` and `test-table` (among others).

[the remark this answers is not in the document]
Twey: Nevertheless.  It is a nit, though, so feel free to ignore :)

Twey:
Modulo one non-nit comment I'm happy with this, but I'd appreciate a review from someone more familiar with elliptic curve cryptography.

Cargo.toml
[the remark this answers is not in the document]
deuszx: I'd prefer to ignore this stylistic change if it doesn't follow the already-established "standard". If we want to do it that way, let's update everything at once.

linera-base/src/crypto/secp256k1.rs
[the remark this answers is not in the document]
deuszx: After talking to @afck we decided that it's "ok" to panic here. From reading the code it looks like the sources of the errors here are not controllable by us (like a wrong choice of `k` for the secp256k1 signing algorithm, which is an internal details of `k256` anyway) and would most likely either lead to all signature creation fail or would be non-deterministically unsafe.

linera-base/src/crypto/mod.rs
```diff
@@ -50,14 +49,19 @@ pub enum CryptoError {
...
+        len: usize,
+        expected: usize,
+    },
     #[error("Could not parse integer: {0}")]
     ParseIntError(#[from] ParseIntError),
     #[error("secp256k1 error: {0}")]
-    Secp256k1Error(::secp256k1::Error),
+    Secp256k1Error(k256::ecdsa::Error),
+    #[error("Could not parse public key: {0}. Public key point at inifinity.")]
+    Secp256k1PointAtIfinity(String),
```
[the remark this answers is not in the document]
deuszx: `.context("")` works only on `anyhow` errors though, right?

linera-base/src/crypto/mod.rs
```diff
@@ -50,14 +49,19 @@ pub enum CryptoError {
...
+        len: usize,
+        expected: usize,
+    },
     #[error("Could not parse integer: {0}")]
     ParseIntError(#[from] ParseIntError),
     #[error("secp256k1 error: {0}")]
-    Secp256k1Error(::secp256k1::Error),
+    Secp256k1Error(k256::ecdsa::Error),
+    #[error("could not parse public key: {0}: key point at inifinity.")]
+    Secp256k1PointAtIfinity(String),
```
Twey, line 64:
```suggestion
    #[error("could not parse public key `{0}`: point at infinity")]
    Secp256k1PointAtInfinity(String),
```
(re-commenting suggestion since it was partially applied)

linera-base/src/crypto/mod.rs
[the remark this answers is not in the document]
deuszx: That requires adding new dependency to `linera-base`. What is this improving?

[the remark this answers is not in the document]
Twey: It allows for a generalized way to report the context of errors, meaning we can attach context (which is very helpful for debugging) to errors without adding O(mn) error variants for `error condition × code path`.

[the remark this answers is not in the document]
deuszx: I understand that but in this particular case? When this error can originate in only one codepath.

linera-base/src/crypto/mod.rs
```diff
@@ -50,14 +49,19 @@ pub enum CryptoError {
     )]
     IncorrectHashSize(usize),
     #[error(
-        "Byte slice has length {0} but a `PublicKey` requires exactly {expected} bytes",
-        expected = dalek::PUBLIC_KEY_LENGTH,
+        "Byte slice has length {len} but a {scheme} `PublicKey` requires exactly {expected} bytes"
     )]
-    IncorrectPublicKeySize(usize),
+    IncorrectPublicKeySize {
+        scheme: String,
```
afck, line 55: This could probably just be a `&'static str`?

linera-base/src/crypto/mod.rs
afck, line 64:
```suggestion
    Secp256k1PointAtInfinity(String),
```

linera-base/src/crypto/secp256k1.rs
```diff
@@ -34,17 +51,42 @@ pub struct Secp256k1KeyPair {
...
-
         let mut rng = rand::rngs::StdRng::seed_from_u64(seed as u64);
-        let secp = secp256k1::Secp256k1::signing_only();
-        Self(secp.generate_keypair(&mut rng).1)
+        let sk = k256::SecretKey::random(&mut rng);
+        Self(sk.public_key().into())
+    }
+
+    /// Returns the bytes of the public key in compressed representation.
+    pub fn as_bytes(&self) -> Vec<u8> {
```
afck, line 67: Should this be a `[u8; 33]`?

linera-base/src/crypto/secp256k1.rs
```diff
@@ -34,17 +51,42 @@ pub struct Secp256k1KeyPair {
...
+        self.0.to_encoded_point(true).as_bytes().to_vec()
+    }
+
+    /// Decodes the bytes into the public key.
+    /// Expects the bytes to be of compressed representation.
+    ///
+    /// Panics if the encoding can't be done in a constant time.
```
afck, line 74: Why "in a constant time"?
(I guess this is constant time, but it's not like it can ever fail because it could decode it but would take longer?)

linera-base/src/crypto/secp256k1.rs
```diff
@@ -119,30 +163,29 @@ impl FromStr for Secp256k1PublicKey {
...
+        Self::from_bytes(value)
     }
 }
 
 impl fmt::Display for Secp256k1PublicKey {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let s = hex::encode(self.0.serialize());
-        write!(f, "{}", s)
+        let str = hex::encode(self.0.to_encoded_point(true).to_bytes());
+        write!(f, "{}", str)
```
afck, line 181: Would `str.fmt(f)` respect formatting options (e.g. `format!("{:8}", pub_key)`)?

linera-base/src/crypto/secp256k1.rs
```diff
@@ -119,30 +163,29 @@ impl FromStr for Secp256k1PublicKey {
...
     }
 }
 
 impl fmt::Debug for Secp256k1PublicKey {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "{}", hex::encode(&self.0.serialize()[0..9]))
+        let str = hex::encode(&self.0.to_encoded_point(true).to_bytes()[0..9]);
+        write!(f, "{}..", str)
```
afck, line 188: …then you could simplify this:

```suggestion
        write!(f, "{:9}..", self)
```

(Both comments also apply to signatures below.)

linera-base/src/crypto/secp256k1.rs
```diff
@@ -187,9 +231,13 @@ impl Secp256k1Signature {
...
-        let secp = secp256k1::Secp256k1::signing_only();
-        let message = Message::from_digest(CryptoHash::new(value).as_bytes().0);
-        let signature = secp.sign_ecdsa(&message, &secret.0);
+        use k256::ecdsa::signature::hazmat::PrehashSigner;
+
+        let prehash = CryptoHash::new(value).as_bytes().0;
+        let (signature, _rid) = secret
+            .0
+            .sign_prehash(&prehash)
+            .expect("Failed to sign prehased data"); // NOTE: This is a critical error we don't control.
```
afck, line 240:
```suggestion
            .expect("Failed to sign prehashed data"); // NOTE: This is a critical error we don't control.
```

linera-base/src/crypto/secp256k1.rs
```diff
@@ -349,4 +417,58 @@ mod tests {
...
+        assert_eq!(key, key2);
+    }
+
+    #[test]
+    fn bytes_repr_compact_public_key() {
+        use crate::crypto::secp256k1::Secp256k1PublicKey;
+        let key_in: Secp256k1PublicKey = Secp256k1PublicKey::test_key(0);
+        let bytes = key_in.as_bytes();
+        assert!(
+            bytes.len() == 33,
```
afck, line 455:
```suggestion
            bytes.len() == SECP256K1_PUBLIC_KEY_SIZE,
```

afck (approved):
Looks good! No blockers from my side.

linera-base/src/crypto/secp256k1.rs
deuszx: The conversion is fallible but I guess it should work since we're operating on a valid key here.

afck: Yes, and the type guards its invariant, right? Both in all constructors and on deserialization?

linera-base/src/crypto/secp256k1.rs
deuszx: My version was slightly more performant since we were cutting of the unneeded bytes but this is more elegant.

linera-base/src/crypto/secp256k1.rs
deuszx: Not necessarily. This is a compressed representation - 33 bytes - and the uncompressed one can be 65 bytes IIRC so - in theory - this conversion could fail if we ever switched to uncompressed representation. But there's a unit test for that.

linera-base/src/crypto/secp256k1.rs
afck: (I don't understand where we cut off the bytes? Anyway, just a thought; I'm fine with either variant, of course!)

deuszx: Not here ofc but in the Debug variants.

linera-base/src/crypto/secp256k1.rs
deuszx: Done in [Apply review suggestions.](https://github.com/linera-io/linera-protocol/pull/3407/commits/8f530c26831f602c5a89abc74abc7445a02389d6).

linera-base/src/crypto/secp256k1.rs
deuszx: Done in [Apply review suggestions.](https://github.com/linera-io/linera-protocol/pull/3407/commits/8f530c26831f602c5a89abc74abc7445a02389d6).

linera-base/src/crypto/secp256k1.rs
deuszx: Done in [Apply review suggestions.](https://github.com/linera-io/linera-protocol/pull/3407/commits/8f530c26831f602c5a89abc74abc7445a02389d6).

linera-base/src/crypto/secp256k1.rs
deuszx: Done in [Apply review suggestions.](https://github.com/linera-io/linera-protocol/pull/3407/commits/8f530c26831f602c5a89abc74abc7445a02389d6).

linera-base/src/crypto/mod.rs
deuszx: Done in [Apply review suggestions.](https://github.com/linera-io/linera-protocol/pull/3407/commits/8f530c26831f602c5a89abc74abc7445a02389d6).

linera-base/src/crypto/mod.rs
deuszx: Done in [Apply review suggestions.](https://github.com/linera-io/linera-protocol/pull/3407/commits/8f530c26831f602c5a89abc74abc7445a02389d6).

linera-base/src/crypto/mod.rs
deuszx: Done in [Apply review suggestions.](https://github.com/linera-io/linera-protocol/pull/3407/commits/8f530c26831f602c5a89abc74abc7445a02389d6).

linera-base/src/crypto/secp256k1.rs
afck: Ah, good point, I totally didn't think of that! Feel free to revert, if you prefer.

And if you don't: I also missed that 9 bytes is 18 characters, not 9 characters. :facepalm:

linera-base/src/crypto/mod.rs
Twey: The modified version you committed still has trailing punctuation, against the `Error` guidelines :sweat_smile: 

The reason for this guideline is that it's common to embed the error message inside a larger error string at a caller site up the stack, and having trailing punctuation breaks the grammar, which can cause it to be difficult to decide where one error ends and the next begins when reading the error.  This hurts user (or developer) experience.

linera-base/src/crypto/mod.rs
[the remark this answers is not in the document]
Twey: I think it's generally good practice to keep error context out of the error enum, even so.  Particularly, it's good to make sure errors are context-enabled because it drastically lowers the barrier to adding context to an error (not just in this module but also potentially in consumers), which encourages people to add more context, which results in easier debugging down the line.  But I understand if you think this is too much work for this PR: that's why I marked it a nit :)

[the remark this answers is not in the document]
deuszx: My way of using `#[error(...)]` and errors in general is different then. Here, the error is self-describing so there's no need IMHO for the additional `Display` impl (via `#[error` attribute). As for the context - I think it's unnecessary if there's only one place where it can be returned.

Also, none of the variants in `CryptoError` uses that context either.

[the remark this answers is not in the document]
Twey: I'm a bit confused, you want to remove the `Display` implementation?  `Display` is required for `std::error::Error`, which is why `thiserror`'s `error` attribute generates one.  Without an `Error` implementation you won't be able to use this enum ‘as an error’ in a variety of standard-library contexts, and it won't get backtraces, et cetera.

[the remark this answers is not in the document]
deuszx: I don't want to remove it. I'm just saying that it's not always making any difference. If error is self-describing then it doesn't need a whole sentence. If there's one place where it can be thrown from it doesn't need the context. In my opinion.

linera-base/src/crypto/mod.rs
deuszx: OK but let's not introduce new guidelines for one variant out of 10. If we want to agree how it should be done let's all agree on some common approach and introduce it for everything in a single PR. If I do what you're suggesting then this is still inconsistent and who's making a decision which one is better.

linera-base/src/crypto/mod.rs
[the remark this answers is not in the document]
Twey: It's a little orthogonal to the discussion, since this variant does already have context baked in anyway and the original message was not one but two full sentences :), but in general the `Display` implementation of an error is intended to be a friendly explanation to the end user of what's gone wrong and why.

deuszx:
I think I've updated everything and the only unaddressed comment was the stylistic suggestion that I'd prefer to apply for the whole codebase (or at least the crate) rather than one variant. 

Merging.

PR #3410, opened 2025-02-25
spelling errors metering.rs

comfsrt:
_read_value_number_none_cases - _read_value_none_cases

### DIFF
--- a/linera-views/src/backends/metering.rs
+++ b/linera-views/src/backends/metering.rs
@@ -158,7 +158,7 @@ impl KeyValueStoreMetrics {
         let entry2 = format!("{} delete latency", title_name);
         let delete_latency = register_histogram_vec(&entry1, &entry2, &[], None);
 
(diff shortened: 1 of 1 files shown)


linera-views/src/backends/metering.rs
```diff
@@ -158,7 +158,7 @@ impl KeyValueStoreMetrics {
         let entry2 = format!("{} delete latency", title_name);
         let delete_latency = register_histogram_vec(&entry1, &entry2, &[], None);
 
-        let entry1 = format!("{}_read_value_number_none_cases", var_name);
+        let entry1 = format!("{}_read_value_none_cases", var_name);
         let entry2 = format!("{} read value number none cases", title_name);
```
afck, line 162: Good catch! This one, too:

```suggestion
        let entry2 = format!("{} read value none cases", title_name);
```

afck (approved):

Issue #3409, opened 2025-02-25
Migrate from `cargo-sort` to `taplo` for CI


Twey:
We currently use `cargo-sort` in CI to normalize our `Cargo.toml` files.  However, this project is on life support, requiring us to use a fork to fix bugs and stay up to date with the TOML spec.

An alternative is [Taplo](https://taplo.tamasfe.dev/), a TOML utility that provides formatting functionality out of the box, with [significantly more customization options](https://taplo.tamasfe.dev/configuration/formatter-options.html).

wellcast:
Hey @Twey, I took the freedom to take this first chore to my best understanding. Please let correct/feedback is welcome.

PR #3413, opened 2025-02-25
Migrate from cargo-sort to Taplo (#3409)

wellcast:
## Motivation

A fork of cargo-sort used in CI to normalize Cargo.toml files as the project is on life support. #3409

## Proposal

Migrate from cargo-sort to [Taplo](https://taplo.tamasfe.dev/), while current lint step should verify the order of dependencies alongside schema validation. 

## Test Plan

CI

## Release Plan

## Links

- https://taplo.tamasfe.dev/cli/usage/validation.html

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -338,18 +338,18 @@ jobs:
       run: |
         ln -sf toolchains/nightly/rust-toolchain.toml
     - uses: actions-rust-lang/setup-rust-toolchain@v1
-    - name: Install cargo-sort
+    - name: Install taplo-sort
       run: |
-        RUSTFLAGS='' cargo install cargo-sort --git https://github.com/Twey/cargo-sort/ --tag linera
+        RUSTFLAGS='' cargo install taplo-cli --locked
     - name: Check if Cargo.toml files are sorted
       run: |
-        cargo sort --check --workspace --grouped
+        taplo fmt --check
     - name: Check if example Cargo.toml files are sorted
       run: |
         cd examples
-        cargo sort --check --workspace --grouped
+        taplo fmt --check
         cd agent
-        cargo sort --check
+        taplo fmt --check
 
   lint-check-for-outdated-readme:
     runs-on: ubuntu-latest
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -340,15 +340,9 @@ jobs:
     - uses: actions-rust-lang/setup-rust-toolchain@v1
     - name: Install taplo-sort
       run: |
-        RUSTFLAGS='' cargo install taplo-cli --locked
+        RUSTFLAGS='' cargo install taplo-cli@0.9.3 --locked
     - name: Check if Cargo.toml files are sorted
       run: |
-        taplo fmt --check
-    - name: Check if example Cargo.toml files are sorted
-      run: |
-        cd examples
-        taplo fmt --check
-        cd agent
         taplo fmt --check
 
   lint-check-for-outdated-readme:
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -330,7 +330,7 @@ jobs:
 
   lint-cargo-sort:
     runs-on: ubuntu-latest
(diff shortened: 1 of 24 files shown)


Twey:
Hey, thanks for this super speedy contribution! 

It's looking great, but I think you've accidentally removed the `k256` crate. Once the checks pass I'm happy to merge this :)

wellcast:
> Hey, thanks for this super speedy contribution!
> 
> It's looking great, but I think you've accidentally removed the `k256` crate. Once the checks pass I'm happy to merge this :)

Happy to contribute! The removed crate should be back now.

Twey:
There's also a dependency missing.  I recommend running `cargo build` locally to check everything checks out!

Did you reformat using `taplo fmt` or by hand?  If `taplo fmt` changed the set of dependencies I'd be a little concerned.

wellcast:
`taplo fmt` was used and worked fine. Fixing a conflict between re-ordered dependencies and changes on main left some duplicated entries. It looks Ok now.

Twey:
Ah that's good to hear, thank you :)

.github/workflows/rust.yml
```diff
@@ -318,18 +318,18 @@ jobs:
...
-        cargo sort --check --workspace --grouped
+        taplo fmt --check
     - name: Check if example Cargo.toml files are sorted
       run: |
         cd examples
-        cargo sort --check --workspace --grouped
+        taplo fmt --check
         cd agent
-        cargo sort --check
+        taplo fmt --check
```
Twey, line 332: I think this separate job should be unnecessary now — let's just include these files in the `.taplo.toml`.

.taplo.toml
```diff
@@ -0,0 +1,11 @@
+include = ["Cargo.toml"]
```
Twey, line 1: This causes Taplo to only check the root `Cargo.toml` — so the subcrates' `Cargo.toml`s are no longer being checked.

wellcast: afaik that was already the case. but we can go and cascade-test all examples then.

wellcast: jfyi I initially did not include every lenera-* directory as it seemed to not check those prior to these changes. I will push a revised commit.

Twey: The `--workspace` flag causes `cargo sort` to check every `Cargo.toml` file in the workspace.  We only have to do the examples separately because they're in their own workspace.

.github/workflows/rust.yml
```diff
@@ -294,42 +294,22 @@ jobs:
...
     - uses: actions/checkout@v3
     - name: Put lint toolchain file in place
       run: |
         ln -sf toolchains/nightly/rust-toolchain.toml
     - uses: actions-rust-lang/setup-rust-toolchain@v1
-    - name: Install cargo-sort
+    - name: Install taplo-sort
       run: |
-        RUSTFLAGS='' cargo install cargo-sort --git https://github.com/Twey/cargo-sort/ --tag linera
+        RUSTFLAGS='' cargo install taplo-cli --locked
```
Twey, line 309: It's probably also a good idea to pin the `taplo-cli` version here.

.taplo.toml
```diff
@@ -0,0 +1,11 @@
+include = ["Cargo.toml", "**/Cargo.toml", "examples/**/Cargo.toml"]
+
+[formatting]
+reorder_keys = false
+
+[[rule]]
+include = ["Cargo.toml", "**/Cargo.toml", "examples/**/Cargo.toml"]
```
Twey, line 7: Is it necessary to duplicate the file listing here?  Surely without this line it will apply to all files anyway.

.taplo.toml
```diff
@@ -0,0 +1,11 @@
+include = ["**/Cargo.toml"]
```
wellcast, line 1: @Twey this is obviously covering every Cargo.toml in the repo (). Please let me know if that is not the desired state. 
```
ls -1 **/Cargo.toml | wc -l
      47
```

I also covered the common dependencies keys below. Which reflected in more files changes.

tesleo:
Cool )

PR #3414, opened 2025-02-25
Add --policy-config argument to create-genesis-config

ndr-ds:
## Motivation

Right now we pass everything manually, but there's already a config that can make this easier

## Proposal

User the policy config instead

## Test Plan

CI

## Release Plan

### DIFF
--- a/CLI.md
+++ b/CLI.md
@@ -491,47 +491,31 @@ Create genesis configuration for a Linera deployment. Create initial user chains
 
   Default value: `0`
 * `--start-timestamp <START_TIMESTAMP>` — The start timestamp: no blocks can be created before this time
-* `--block-price <BLOCK_PRICE>` — Set the base price for creating a block
+* `--policy-config <POLICY_CONFIG>` — Configure the resource control policy (notably fees) according to pre-defined settings
 
-  Default value: `0`
-* `--fuel-unit-price <FUEL_UNIT_PRICE>` — Set the price per unit of fuel
+  Default value: `no-fees`
 
-  Default value: `0`
-* `--read-operation-price <READ_OPERATION_PRICE>` — Set the price per read operation
+  Possible values: `no-fees`, `testnet`
 
-  Default value: `0`
-* `--write-operation-price <WRITE_OPERATION_PRICE>` — Set the price per write operation
-
-  Default value: `0`
-* `--byte-read-price <BYTE_READ_PRICE>` — Set the price per byte read
-
-  Default value: `0`
-* `--byte-written-price <BYTE_WRITTEN_PRICE>` — Set the price per byte written
-
-  Default value: `0`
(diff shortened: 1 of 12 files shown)


ndr-ds:
* **#3414** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/3414?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/3414?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

CLI.md
```diff
@@ -491,39 +491,23 @@ Create genesis configuration for a Linera deployment. Create initial user chains
...
-* `--block-price <BLOCK_PRICE>` — Set the base price for creating a block
+* `--policy-config <POLICY_CONFIG>` — Configure the resource control policy (notably fees) according to pre-defined settings
 
-  Default value: `0`
-* `--fuel-unit-price <FUEL_UNIT_PRICE>` — Set the price per unit of fuel
+  Default value: `default`
 
-  Default value: `0`
-* `--read-operation-price <READ_OPERATION_PRICE>` — Set the price per read operation
+  Possible values: `default`, `only-fuel`, `fuel-and-block`, `all-categories`, `devnet`
```
afck, line 498: I'm not sure we need all these, and I don't think the names should be user-visible like this. I think we should only have two user-visible ones:
* `testnet` (renamed from `devnet`), and
* `no-fees` (renamed from `default`, where everything is zero and limits are `MAX`).

The other two are, I think, only used in tests, and should be feature-gated?

afck (commented):
Great idea! :+1:

linera-client/src/client_options.rs
```diff
@@ -640,49 +640,54 @@ pub enum ClientCommand {
...
 
+        /// Configure the resource control policy (notably fees) according to pre-defined
+        /// settings.
+        #[arg(long, default_value = "default")]
+        policy_config: ResourceControlPolicyConfig,
+
         /// Set the base price for creating a block.
-        #[arg(long, default_value = "0")]
-        block_price: Amount,
+        #[arg(long)]
```
deuszx, line 649: I think there's an option in `clap` to mark arguments mutually exclusive - i.e. that one doesn't have to be provided if another one is (here path to the config file). 

Unless we want to allow for overwriting the config file with CLI args but if we do I'd suggest adding that explicitly in the comments.

ndr-ds: I'll add it to the comments

ndr-ds:
### Merge activity

* **Feb 28, 9:56 AM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/linera-io/linera-protocol/3414).

PR #3415, opened 2025-02-26
Limit network actions per batch

jvff:
## Motivation

There's a risk of `O(n^2)` behavior in the chain worker, because when network actions are created (for cross-chain messages) **all** outboxes are inspected in order to figure out which actions should be created. However, the outboxes are only emptied after a confirmation of the receipt from the receiver. This means that until that confirmation is received, the same network actions might be created again and again, stressing the cross-chain channels.

## Proposal

Limit the number of outboxes inspected to create the network actions. The limit is configured on server shards to be equal to the cross-shard communication channel queue size. If the limit is reached, random outboxes are skipped when creating a new batch of network actions.

## Test Plan

TODO: should we add a stress test for this scenario?

## Release Plan

Because this could be related to a performance issue the testnet is running into.

## Links

### DIFF
--- a/linera-core/src/chain_worker/state/mod.rs
+++ b/linera-core/src/chain_worker/state/mod.rs
@@ -32,6 +32,7 @@ use linera_execution::{
 };
 use linera_storage::{Clock as _, Storage};
 use linera_views::views::{ClonableView, ViewError};
+use rand::{prelude::SliceRandom as _, thread_rng};
 use tokio::sync::{oneshot, OwnedRwLockReadGuard, RwLock};
 
 #[cfg(test)]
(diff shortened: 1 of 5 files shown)


linera-core/src/chain_worker/state/mod.rs
```diff
@@ -423,6 +424,18 @@ where
...
+                targets.drain(0..elements_to_discard);
+            } else {
+                targets.partial_shuffle(&mut thread_rng(), limit);
+                targets.truncate(limit);
+            }
+        }
+
         if let Some(tracked_chains) = self.tracked_chains.as_ref() {
             let publishers = self
                 .chain
```
ma2bd, line 441: Personally, I was only worried about returning too many `NetworkAction`s, so I'm a little bit on fence regarding the randomization idea. It could also be counter-productive.

linera-core/src/chain_worker/state/mod.rs
```diff
@@ -423,6 +424,18 @@ where
     async fn create_network_actions(&self) -> Result<NetworkActions, WorkerError> {
         let mut heights_by_recipient = BTreeMap::<_, BTreeMap<_, _>>::new();
         let mut targets = self.chain.outboxes.indices().await?;
+
+        if let Some(limit) = self.config.network_actions_batch_limit {
+            if limit > targets.len() / 2 {
+                let elements_to_discard = targets.len().saturating_sub(limit);
+                targets.partial_shuffle(&mut thread_rng(), elements_to_discard);
```
afck, line 431: Why not just use `choose_multiple`? That should be faster, not require the case distinction, and the _order_ of the selected elements doesn't really matter that much. I guess the downside is that it clones the entries, but they are small, I think?

jvff:
Closing because this was intended to be a quick fix, but it's no longer high-priority, so a better solution can be designed instead.

PR #3416, opened 2025-02-26
Export linera_execution::system::Recipient to application tests

kikakkz:
## Motivation

Let application unit tests be able to call `block.with_native_token_transfer` which need `Recipient`.

## Proposal

Export `linera_execution::system::Recipient` in `linera-sdk/src/test` crate.

## Test Plan

CI

## Release Plan


jvff (approved):
Thanks!

PR #3419, opened 2025-02-26
 Discrepancies in Comments

dmytroheknt:
There's a mismatch in the second doc comment: it references `find_keys_by_prefix` 
but the function is actually named `find_key_values_by_prefix`.
Also, the top comment mentions generating about 2000 keys of length 11, 
whereas the code sets NUM_ENTRIES to 200 and LEN_KEY to 10.


ma2bd:
Closing this PR for now as it seems to be stale

PR #3420, opened 2025-02-26
Pin the version of the `base64ct` indirect dependency

jvff:
## Motivation

Version 1.7.0 breaks semver compatibility by requiring the usage of a new Rust edition. More information is available in https://github.com/RustCrypto/formats/issues/1684.

## Proposal

We should pin it to the last version that we still support at least until we manage to update to the new Rust edition.

## Test Plan

## Release Plan

We may have to make a new SDK release with a new patch version.

## Links

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4858,6 +4858,7 @@ dependencies = [
  "anyhow",
  "async-graphql",
  "async-trait",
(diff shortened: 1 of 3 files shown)


deuszx (approved):
🤞

PR #3426, opened 2025-02-27
spelling error

pendrue:
changed:

test works - tests work


afck:
Good catch! You'll also have to make the same change in `lib.rs`, to make the tests work (no pun intended). CI compares them.

afck (approved):

PR #3427, opened 2025-02-27
Only allow HTTP requests authorized by the committee

jvff:
## Motivation

For security reasons, we would like to initially limit the hosts that applications can send HTTP requests to.

## Proposal

Add an allow-list of hosts to the `ResourceControlPolicy` configured by a committee. Only perform HTTP requests in the execution state actor if the target host is in the list.

## Test Plan

Tests that exercise the allow list were added in PR #3509.

## Release Plan

- Nothing to do / These changes follow the usual release cycle, because this is a new feature that should be included in the next release and testnet.

## Links


linera-execution/src/execution_state_actor.rs
```diff
@@ -336,8 +336,24 @@ where
                     .map(|http::Header { name, value }| Ok((name.parse()?, value.try_into()?)))
                     .collect::<Result<HeaderMap, ExecutionError>>()?;
 
+                let url = Url::parse(&request.url)?;
+                let host = url
+                    .host_str()
+                    .ok_or_else(|| ExecutionError::UnauthorizedHttpRequest(url.clone()))?;
```
deuszx, line 342: Did you mean to use `InvalidUrlForHttpRequest` here instead?

linera-execution/src/execution_state_actor.rs
```diff
@@ -336,8 +336,24 @@ where
...
+                let allowed_hosts = &committee.policy().http_request_allow_list;
+
+                ensure!(
+                    allowed_hosts.contains(host),
+                    ExecutionError::UnauthorizedHttpRequest(url)
+                );
+
                 let response = Client::new()
-                    .request(request.method.into(), request.url)
+                    .request(request.method.into(), url)
```
deuszx, line 356: Should we whitelist also methods?

linera-execution/src/execution_state_actor.rs
jvff: Not really, honestly I think either is fine :thinking: . However the `UnauthorizedHttpRequest` provides the URL in the error :man_shrugging:

linera-execution/src/execution_state_actor.rs
jvff: I don't think that's needed. We could make the allow list more complicated, mapping hosts to a set of allowed methods, but where do we draw the line? We could also limit headers, ports, etc. Right now I'm assuming that the allowed HTTP servers will be safe to communicated with.

ma2bd (approved):

PR #3431, opened 2025-02-27
Preparations for epoch change events.

afck:
## Motivation

To make epoch changes scalable, we want to use events instead of messages (https://github.com/linera-io/linera-protocol/issues/365).

This will be done in an upcoming PR. To make that easier to review, I separated some of the changes I need to make as a preparation.

The follow-up PR will add `ProcessNewEpoch` and `ProcessRemovedEpoch` system operations and an `Oracle::Response::Event` variant. Clients will be _notified_ about new epochs because they will always follow/synchronize the admin chain and thus store the epoch change events locally. They will then run `process_inbox` which, in addition to incoming messages, will look for epoch change events, and include the new operations in the proposal accordingly. Processing the operations will check for the event, add the oracle response, and update the chain's epoch and committees.

## Proposal

* Add an `Event` oracle response variant, to record read event values in the block.
* Don't fail in `make_chain_client` if the admin chain is not in the wallet. Clients will always need to sync/follow the admin chain to learn about epoch change events.
* Return early from `try_synchronize_chain_state_from` if the remote node has no new blocks for us.
* Always initialize the root chain in the `linera-core` tests, even if it is not mentioned in the test case.
* Add events methods to the execution runtime context: We will need these in `system.rs` to load the epoch change events.
* Add more assertions to `test_end_to_end_reconfiguration`. This was useful for debugging.
* Other minor cleanups.

## Test Plan

No new behavior was added. CI will catch regressions.

## Release Plan

## Links

- Part of #365.

### DIFF
--- a/linera-base/src/data_types.rs
+++ b/linera-base/src/data_types.rs
@@ -780,8 +780,6 @@ pub enum OracleResponse {
     Assert,
     /// The block's validation round.
     Round(Option<u32>),
-    /// An event was read.
-    Event(EventId, Vec<u8>),
 }
 
 impl<'de> BcsHashable<'de> for OracleResponse {}
--- a/linera-base/src/data_types.rs
+++ b/linera-base/src/data_types.rs
@@ -780,6 +780,8 @@ pub enum OracleResponse {
     Assert,
     /// The block's validation round.
     Round(Option<u32>),
+    /// An event was read.
+    Event(EventId, Vec<u8>),
 }
 
 impl<'de> BcsHashable<'de> for OracleResponse {}
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -1011,8 +1011,7 @@ where
             // For chains with any owner other than ourselves, we could be missing recent
             // certificates created by other owners. Further synchronize blocks from the network.
             // This is a best-effort that depends on network conditions.
-            let nodes = self.validator_nodes().await?;
-            info = self.synchronize_chain_state(&nodes, self.chain_id).await?;
+            info = self.synchronize_chain_state(self.chain_id).await?;
         }
 
         let result = self
@@ -1700,17 +1699,20 @@ where
 
     /// Downloads and processes any certificates we are missing for the given chain.
     #[instrument(level = "trace", skip_all)]
-    pub async fn synchronize_chain_state(
+    async fn synchronize_chain_state(
         &self,
-        validators: &[RemoteNode<P::Node>],
         chain_id: ChainId,
     ) -> Result<Box<ChainInfo>, ChainClientError> {
         #[cfg(with_metrics)]
         let _latency = metrics::SYNCHRONIZE_CHAIN_STATE_LATENCY.measure_latency();
 
-        let committee = self.local_committee().await?;
+        let (epoch, mut committees) = self.epoch_and_committees(chain_id).await?;
+        let committee = committees
+            .remove(&epoch.ok_or(LocalNodeError::InvalidChainInfoResponse)?)
+            .ok_or(LocalNodeError::InvalidChainInfoResponse)?;
+        let validators = self.make_nodes(&committee)?;
         communicate_with_quorum(
-            validators,
+            &validators,
             &committee,
             |_: &()| (),
             |remote_node| {
@@ -1749,6 +1751,9 @@ where
             .with_sent_certificate_hashes_in_range(range)
             .with_manager_values();
         let info = remote_node.handle_chain_info_query(query).await?;
+        if info.next_block_height < local_info.next_block_height {
+            return Ok(());
+        }
 
         let certificates: Vec<ConfirmedBlockCertificate> = remote_node
             .download_certificates(info.requested_sent_certificate_hashes)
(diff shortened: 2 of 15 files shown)


linera-service/tests/local_net_tests.rs
```diff
@@ -123,7 +124,7 @@ async fn test_end_to_end_reconfiguration(config: LocalNetConfig) -> Result<()> {
     client.query_validators(Some(chain_1)).await?;
 
     if matches!(network, Network::Grpc) {
-        assert_eq!(faucet.current_validators().await?.len(), 5);
+        assert_eq!(faucet.current_validators().await.unwrap().len(), 5);
```
jvff, line 127: Are we back to using `unwrap` in tests instead of `?` with `anhow::Error`s?

jvff (approved):

linera-core/src/client/mod.rs
```diff
@@ -1700,17 +1699,20 @@ where
...
+    async fn synchronize_chain_state(
         &self,
-        validators: &[RemoteNode<P::Node>],
         chain_id: ChainId,
     ) -> Result<Box<ChainInfo>, ChainClientError> {
         #[cfg(with_metrics)]
         let _latency = metrics::SYNCHRONIZE_CHAIN_STATE_LATENCY.measure_latency();
 
-        let committee = self.local_committee().await?;
+        let (epoch, mut committees) = self.epoch_and_committees(chain_id).await?;
```
deuszx, line 1709: epoch and committee should be the same for all chains?

linera-base/src/data_types.rs
```diff
@@ -780,6 +780,8 @@ pub enum OracleResponse {
     Assert,
     /// The block's validation round.
     Round(Option<u32>),
+    /// An event was read.
+    Event(EventId, Vec<u8>),
```
deuszx, line 784: I thought that `OracleResponses` are for things that are non-deterministic and we need to either enforce or check that they match. Why is reading an event - by its id - an oracle event?

linera-service/src/cli_wrappers/wallet.rs
```diff
@@ -1239,6 +1243,17 @@ impl NodeService {
         Ok(bytecode_id.with_abi())
     }
 
+    pub async fn query_committees(&self, chain_id: &ChainId) -> Result<BTreeMap<Epoch, Committee>> {
+        let query = format!(
+            "query {{ chain(chainId:\"{chain_id}\") {{
+                executionState {{ system {{ committees }} }}
+            }} }}"
+        );
```
deuszx, line 1251: I wish this could be typesafe. Hardcoding strings into these queries had caused a lot of debugging headaches in the past.

linera-core/src/client/mod.rs
afck: No, that's exactly what the epoch change messages—and soon the epoch change events—are about: Every chain always needs to have each epoch's last block confirmed by the old epoch's validators, so it needs to explicitly acknowledge the change.

deuszx (commented):
This new Event API (add_event, get_event) is not used anywhere yet?

linera-base/src/data_types.rs
afck: Because a validator may or may not have seen that event yet.

In the future we will probably write client code to update validators about missing events, but for now it is sufficient if everyone synchronizes the admin chain.

linera-service/src/cli_wrappers/wallet.rs
afck: I agree. But I just moved this code, and we have lots of strings like this, so I'd like to not address that in this PR.

afck:
> This new Event API (add_event, get_event) is not used anywhere yet?

No, this is just preparation for an upcoming PR. I'm hoping that will make the actual logic changes easier to review.

linera-base/src/data_types.rs
deuszx: events are results of executing a block - won't validator _have to_ see the event if he executed the block (has the certificates)?

deuszx: And you're planning to use it for what exactly? User chains checking for committee/epoch changes on the admin chain?

I understand the desire to make the follow-up PRs easier to review but it's hard to understand/review these changes in isolation b/c we're basically reviewing part of a design w/o a design doc and full picture.

afck: OK, let me undo this part. Maybe this belongs in the follow-up PR.

(A user chain processing an epoch change will create this oracle response, yes.)

afck: I removed it, and updated the PR description as discussed.

deuszx:
> They will then run process_inbox which, in addition to incoming messages, will look for epoch change events, and include the new operations in the proposal accordingly. Processing the operations will check for the event, add the oracle response, and update the chain's epoch and committees.

We should think about prioritising these "system" operations over user operations.

deuszx (approved):

afck:
> We should think about prioritising these "system" operations over user operations.

In my WIP code that's simply separate: If you run some other command (e.g. a transfer), we create a block just for that, without any other operations. If you run `process-inbox`, we create a block with no operations other than the epoch change ones.

deuszx:
You could still have many more incoming messages than can fit into one (many) blocks.

afck:
Right, yes, we may want to prioritize the system operations over the _messages_ then. I think we actually already do that, too, in that block execution staging loop…

PR #3432, opened 2025-02-27
Use events for adding and removing committees.

afck:
## Motivation

To make epoch changes scalable, we want to use events instead of messages (https://github.com/linera-io/linera-protocol/issues/365).

## Proposal

Remove the committee messages and add add `ProcessNewEpoch` and `ProcessRemovedEpoch` system _operations_ instead. The committee change operations on the admin change create events now, and the two new operations consume those events, i.e. verify that they exist, change the epoch and committees on the local chain, and create an `OracleResponse::Event` entry in the execution outcome.

Clients get notified about new epochs because they always follow/synchronize the admin chain and thus store the epoch change events locally. They will then run `process_inbox` which, in addition to incoming messages, looks for epoch change events, and includes the new operations in the proposal accordingly.

## Test Plan

Several existing tests verify that reconfigurations are processed correctly. These check the new mechanism now instead of the message-based one.

## Release Plan

## Links

- Part of #365.

### DIFF
--- a/linera-base/src/data_types.rs
+++ b/linera-base/src/data_types.rs
@@ -1002,6 +1002,11 @@ impl BlobContent {
         BlobContent::new(BlobType::ApplicationDescription, bytes)
     }
 
+    /// Creates a new committee [`BlobContent`] from the provided serialized committee.
+    pub fn new_committee(committee: impl Into<Box<[u8]>>) -> Self {
+        BlobContent::new(BlobType::Committee, committee)
+    }
+
     /// Gets a reference to the blob's bytes.
     pub fn bytes(&self) -> &[u8] {
         &self.bytes
--- a/linera-base/src/data_types.rs
+++ b/linera-base/src/data_types.rs
@@ -780,6 +780,8 @@ pub enum OracleResponse {
     Assert,
     /// The block's validation round.
     Round(Option<u32>),
+    /// An event was read.
+    Event(EventId, Vec<u8>),
 }
 
 impl<'de> BcsHashable<'de> for OracleResponse {}
--- a/linera-client/src/chain_listener.rs
+++ b/linera-client/src/chain_listener.rs
@@ -1,10 +1,14 @@
 // Copyright (c) Zefchain Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-use std::{collections::HashSet, sync::Arc, time::Duration};
+use std::{
+    collections::{BTreeSet, HashSet},
+    sync::Arc,
+    time::Duration,
+};
 
 use async_trait::async_trait;
-use futures::{lock::Mutex, StreamExt};
+use futures::{lock::Mutex, stream, StreamExt};
 use linera_base::{
     crypto::AccountSecretKey,
     data_types::Timestamp,
@@ -12,8 +16,8 @@ use linera_base::{
 };
 use linera_core::{
     client::{ChainClient, ChainClientError},
-    node::ValidatorNodeProvider,
-    worker::Reason,
+    node::{NotificationStream, ValidatorNodeProvider},
+    worker::{Notification, Reason},
 };
 use linera_execution::{Message, OutgoingMessage, SystemMessage};
 use linera_storage::{Clock as _, Storage};
@@ -100,7 +104,12 @@ impl ChainListener {
     where
         C: ClientContext,
     {
-        let chain_ids = context.lock().await.wallet().chain_ids();
+        let chain_ids = {
+            let guard = context.lock().await;
+            let mut chain_ids = BTreeSet::from_iter(guard.wallet().chain_ids());
+            chain_ids.insert(guard.wallet().genesis_admin_chain());
+            chain_ids
+        };
         for chain_id in chain_ids {
             Self::run_with_chain_id(
                 chain_id,
@@ -155,6 +164,12 @@ impl ChainListener {
         let client = context.lock().await.make_chain_client(chain_id)?;
         let (listener, _listen_handle, local_stream) = client.listen().await?;
         let mut local_stream = local_stream.fuse();
+        let admin_listener: NotificationStream = if client.admin_id() == chain_id {
+            Box::pin(stream::pending())
+        } else {
+            Box::pin(client.subscribe_to(client.admin_id()).await?)
+        };
+        let mut admin_listener = admin_listener.fuse();
         client.synchronize_from_validators().await?;
         drop(linera_base::task::spawn(listener.in_current_span()));
         let mut timeout = storage.clock().current_time();
@@ -169,6 +184,16 @@ impl ChainListener {
                     } else {
                         break;
                     }
+                }
+                maybe_notification = admin_listener.next() => {
+                    // A new block on the admin chain may mean a new committee. Set the timer to
+                    // process the inbox.
+                    if let Some(
+                        Notification { reason: Reason::NewBlock { .. }, .. }
+                    ) = maybe_notification {
+                        timeout = storage.clock().current_time();
+                    }
+                    continue;
                 }
                 () = sleep => {
                     timeout = Timestamp::from(u64::MAX);
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -3053,8 +3053,6 @@ where
         #[cfg(with_metrics)]
         let _latency = metrics::PROCESS_INBOX_WITHOUT_PREPARE_LATENCY.measure_latency();
 
-        self.synchronize_chain_state(self.admin_id).await?;
-
         let (mut min_epoch, mut next_epoch) = {
             let query = ChainInfoQuery::new(self.chain_id).with_committees();
             let info = *self
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -3053,6 +3053,29 @@ where
         #[cfg(with_metrics)]
         let _latency = metrics::PROCESS_INBOX_WITHOUT_PREPARE_LATENCY.measure_latency();
 
+        let mut epoch_change_ops = self.collect_epoch_changes().await?.into_iter();
+
+        let mut certificates = Vec::new();
+        loop {
+            let incoming_bundles = self.pending_message_bundles().await?;
+            let block_operations = epoch_change_ops.next().into_iter().collect::<Vec<_>>();
+            if incoming_bundles.is_empty() && block_operations.is_empty() {
+                return Ok((certificates, None));
+            }
+            match self.execute_block(block_operations, vec![]).await {
+                Ok(ExecuteBlockOutcome::Executed(certificate))
+                | Ok(ExecuteBlockOutcome::Conflict(certificate)) => certificates.push(certificate),
+                Ok(ExecuteBlockOutcome::WaitForTimeout(timeout)) => {
+                    return Ok((certificates, Some(timeout)));
+                }
+                Err(error) => return Err(error),
+            };
+        }
+    }
+
+    /// Returns operations to process all pending epoch changes: first the new epochs, in order,
+    /// then the revoked epochs, in order.
+    async fn collect_epoch_changes(&self) -> Result<Vec<Operation>, ChainClientError> {
         let (mut min_epoch, mut next_epoch) = {
             let query = ChainInfoQuery::new(self.chain_id).with_committees();
             let info = *self
@@ -3084,24 +3107,7 @@ where
             )));
             min_epoch.try_add_assign_one()?;
         }
-        let mut epoch_change_ops = epoch_change_ops.into_iter();
-
-        let mut certificates = Vec::new();
-        loop {
-            let incoming_bundles = self.pending_message_bundles().await?;
-            let block_operations = epoch_change_ops.next().into_iter().collect::<Vec<_>>();
-            if incoming_bundles.is_empty() && block_operations.is_empty() {
-                return Ok((certificates, None));
-            }
-            match self.execute_block(block_operations, vec![]).await {
-                Ok(ExecuteBlockOutcome::Executed(certificate))
-                | Ok(ExecuteBlockOutcome::Conflict(certificate)) => certificates.push(certificate),
-                Ok(ExecuteBlockOutcome::WaitForTimeout(timeout)) => {
-                    return Ok((certificates, Some(timeout)));
-                }
-                Err(error) => return Err(error),
-            };
-        }
+        Ok(epoch_change_ops)
     }
 
     /// Returns whether the system event on the admin chain with the given stream name and key
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -3074,7 +3074,7 @@ where
     }
 
     /// Returns operations to process all pending epoch changes: first the new epochs, in order,
-    /// then the revoked epochs, in order.
+    /// then the removed epochs, in order.
     async fn collect_epoch_changes(&self) -> Result<Vec<Operation>, ChainClientError> {
         let (mut min_epoch, mut next_epoch) = {
             let query = ChainInfoQuery::new(self.chain_id).with_committees();
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -33,8 +33,8 @@ use linera_base::{
     ensure,
     hashed::Hashed,
     identifiers::{
-        Account, AccountOwner, ApplicationId, BlobId, BlobType, ChainId, MessageId, ModuleId,
-        Owner, UserApplicationId,
+        Account, AccountOwner, ApplicationId, BlobId, BlobType, ChainId, EventId, MessageId,
+        ModuleId, Owner, StreamId, UserApplicationId,
     },
     ownership::{ChainOwnership, TimeoutConfig},
 };
@@ -55,8 +55,8 @@ use linera_chain::{
 use linera_execution::{
     committee::{Committee, Epoch},
     system::{
-        AdminOperation, OpenChainConfig, Recipient, SystemChannel, SystemOperation,
-        OPEN_CHAIN_MESSAGE_INDEX,
+        AdminOperation, OpenChainConfig, Recipient, SystemOperation, EPOCH_STREAM_NAME,
+        OPEN_CHAIN_MESSAGE_INDEX, REMOVED_EPOCH_STREAM_NAME,
     },
     ExecutionError, Operation, Query, QueryOutcome, QueryResponse, SystemExecutionError,
     SystemQuery, SystemResponse,
@@ -702,6 +702,12 @@ impl<P: 'static, S: Storage> ChainClient<P, S> {
         self.chain_id
     }
 
+    /// Gets the ID of the admin chain.
+    #[instrument(level = "trace", skip(self))]
+    pub fn admin_id(&self) -> ChainId {
+        self.admin_id
+    }
+
     /// Gets the hash of the latest known block.
     #[instrument(level = "trace", skip(self))]
     pub fn block_hash(&self) -> Option<CryptoHash> {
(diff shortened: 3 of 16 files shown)


linera-execution/src/system.rs
```diff
@@ -191,6 +195,10 @@ pub enum SystemOperation {
     },
     /// Operations that are only allowed on the admin chain.
     Admin(AdminOperation),
+    /// Processes an event about a new epoch and committee.
+    ProcessNewEpoch(Epoch),
+    /// Processes an event about a removed epoch and committee.
+    ProcessRemovedEpoch(Epoch),
```
deuszx, line 201: Why do we even need this? I.e. creating a new epoch (and changing the committee) should "remove" the previous epoch?

linera-execution/src/system.rs
```diff
@@ -684,6 +674,64 @@ where
...
+                    None => self.context().extra().get_event(event_id.clone()).await?,
+                    Some(OracleResponse::Event(recorded_event_id, bytes))
+                        if recorded_event_id == event_id =>
+                    {
+                        bytes
+                    }
+                    Some(_) => return Err(SystemExecutionError::OracleResponseMismatch),
+                };
+                let committee = bcs::from_bytes(&bytes)?;
+                txn_tracker.add_oracle_response(OracleResponse::Event(event_id, bytes));
```
deuszx, line 702: Do we need the whole committee in the oracle event here? `EventId` should uniquely identify the committee.

deuszx (commented):
First pass through the code. It looks good overall but I had some more general questions. I think the general approach is slightly hacky (i.e. looping through the storage in search for interesting events). It could be improved if we had an option to subscribe for specific event streams in the client.

I am not giving an approval yet - I want to see the answers and then will re-review it.

linera-client/src/chain_listener.rs
```diff
@@ -150,27 +155,48 @@ impl ChainListener {
...
         client.synchronize_from_validators().await?;
         drop(linera_base::task::spawn(listener.in_current_span()));
         let mut timeout = storage.clock().current_time();
         loop {
-            let sleep = Box::pin(storage.clock().sleep_until(timeout));
-            let notification = match future::select(local_stream.next(), sleep).await {
-                Either::Left((Some(notification), _)) => notification,
-                Either::Left((None, _)) => break,
-                Either::Right(((), _)) => {
+            let mut sleep = Box::pin(futures::FutureExt::fuse(
```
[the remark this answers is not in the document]
afck: I know, but it also uses plenty of local variables, which is why I haven't found a good way to refactor it yet. :thinking:

linera-client/src/chain_listener.rs
```diff
@@ -150,27 +155,48 @@ impl ChainListener {
...
+                storage.clock().sleep_until(timeout),
+            ));
+            let notification = futures::select! {
+                maybe_notification = local_stream.next() => {
+                    if let Some(notification) = maybe_notification {
+                        notification
+                    } else {
+                        break;
```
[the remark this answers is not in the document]
afck: ~That means the local stream has been terminated. It's the one with the abort handle.~

Actually, the abort handle isn't used. Both `None` cases should be unreachable.

We should probably consider using the abort handles in the node service and faucet. (Unrelated to this PR, though.)

linera-core/src/unit_tests/worker_tests.rs
```diff
@@ -2872,16 +2768,19 @@ where
...
-                        committee: committee.clone(),
-                    })],
-                    vec![channel_admin_message(SystemMessage::RemoveCommittee {
-                        epoch: Epoch::from(0),
-                    })],
+                messages: vec![vec![]; 2],
+                events: vec![
+                    vec![Event {
+                        value: bcs::to_bytes(&committee).unwrap(),
+                        stream_id: StreamId::system(EPOCH_STREAM_NAME),
```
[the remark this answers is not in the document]
afck: `ADDED_EPOCH_STREAM_NAME`?

(On the other hand, the `REMOVED` one will go away eventually, when we implement proper long range attack protection: Epochs will have to expire based on time, not due to an event, I think.)

linera-execution/src/system.rs
afck: Multiple epochs can be trusted at the same time. Removing them means no longer trusting them. In the future this will have to be based on an expiry time instead.

linera-execution/src/system.rs
afck: I guess we could put the committee into a blob, and make the event only contain the blob ID… :thinking:

linera-client/src/chain_listener.rs
```diff
@@ -150,27 +155,48 @@ impl ChainListener {
...
         drop(linera_base::task::spawn(listener.in_current_span()));
         let mut timeout = storage.clock().current_time();
         loop {
-            let sleep = Box::pin(storage.clock().sleep_until(timeout));
-            let notification = match future::select(local_stream.next(), sleep).await {
-                Either::Left((Some(notification), _)) => notification,
-                Either::Left((None, _)) => break,
-                Either::Right(((), _)) => {
+            let mut sleep = Box::pin(futures::FutureExt::fuse(
+                storage.clock().sleep_until(timeout),
```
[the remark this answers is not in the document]
deuszx: Wouldn't it be more idiomatic if the "process inbox" was a separate process that gets _notified_ (via a channel or an async mutex shared with notifiers) when to process it? Right now it's "time based" and the timer has to be manipualted properly which I find bit hacky.

linera-core/src/unit_tests/worker_tests.rs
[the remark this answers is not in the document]
deuszx: OK, that's what I was trying to get at with this and the other comment about the `RemoveEpoch` event.

Let's rename for now please.

linera-execution/src/system.rs
deuszx: Or that. I think the committee in the event is unnecessary - it can be retrieved if needed but I don't think that's even the case right now - and it bloats the event size.

linera-core/src/client/mod.rs
```diff
@@ -3052,30 +3104,23 @@ where
...
-    #[instrument(level = "trace")]
-    pub async fn unsubscribe_from_new_committees(
-        &self,
-    ) -> Result<ClientOutcome<ConfirmedBlockCertificate>, ChainClientError> {
-        let operation = SystemOperation::Unsubscribe {
-            chain_id: self.admin_id,
-            channel: SystemChannel::Admin,
-        };
-        self.execute_operation(Operation::System(operation)).await
+        match self.client.storage.read_event(event_id).await {
```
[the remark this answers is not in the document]
deuszx: https://github.com/linera-io/linera-protocol/issues/3437

linera-execution/src/system.rs
afck: You mean we could read the original admin chain block instead?

We could, but it's a bit more complicated (iterate over the operations to find the one with the committee), and reads more data (the whole blob instead of just the committee).

Another option is to store the committees separately in storage, as yet another type of data. Then we don't even need them in the chain states? :thinking:

linera-client/src/chain_listener.rs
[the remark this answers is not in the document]
afck: …or maybe we should just finally extract the inbox processing logic from the chain listener and call it directly instead of setting the timeout to `now`.

(But we'll always need a timer for the case where we have to wait for the round timeout to be able to produce a block.)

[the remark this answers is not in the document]
deuszx: Doesn't seem like a chain listener is proper place for that though?

linera-execution/src/system.rs
deuszx: > You mean we could read the original admin chain block instead?

It's already in the storage, right? We just need to be able to get it but we already do - you're using it in this PR.

linera-client/src/chain_listener.rs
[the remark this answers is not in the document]
afck: Yes, maybe it should move into the `ChainClient` itself, in the form of some kind of queue. Also a bigger refactoring, though.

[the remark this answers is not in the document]
deuszx: Yes, but this is the first time we're talking about it. If you agree, please ticket with the details (you know best anyway).

linera-execution/src/system.rs
afck: > It's already in the storage, right?

Yes, as part of the admin chain block.

Would you prefer to only put the admin chain block hash into the event, and then load the block, and go through its iterations to find the operation that contains the committee?

Then we'd have to do a similar thing on the validator side, instead of using a mechanism that we'll use for all events.

And when we change the admin chain logic (e.g. using a staking contract), we would have to change that mechanism again, both for clients and validators, instead of just emitting an event from there.

afck:
Closing this in favor of https://github.com/linera-io/linera-protocol/issues/3453.
(The implementation will probably use some of this PR's code, but also change a lot.)

afck:
Reopening as a draft. I will mark it ready for review when https://github.com/linera-io/linera-protocol/issues/3453 is fully implemented.

linera-base/src/data_types.rs
```diff
@@ -780,6 +780,8 @@ pub enum OracleResponse {
     Assert,
     /// The block's validation round.
     Round(Option<u32>),
+    /// An event was read.
+    Event(EventId, Vec<u8>),
```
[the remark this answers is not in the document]
afck: Not sure: This oracle response _does_ re-certify the event itself (including its content), just not the certificate that created it.

linera-client/src/chain_listener.rs
```diff
@@ -150,27 +155,48 @@ impl ChainListener {
...
-        drop(guard);
         // If the client is not present, we can request it.
         let client = context.lock().await.make_chain_client(chain_id)?;
-        let (listener, _listen_handle, mut local_stream) = client.listen().await?;
+        let (listener, _listen_handle, local_stream) = client.listen().await?;
+        let mut local_stream = local_stream.fuse();
+        let admin_listener: NotificationStream = if client.admin_id() == chain_id {
+            Box::pin(stream::pending())
+        } else {
+            Box::pin(client.subscribe_to(client.admin_id()).await?)
```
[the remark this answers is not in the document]
afck: No, just a local entry in the notifier.

linera-core/src/client/mod.rs
```diff
@@ -3035,13 +3051,49 @@ where
         #[cfg(with_metrics)]
         let _latency = metrics::PROCESS_INBOX_WITHOUT_PREPARE_LATENCY.measure_latency();
 
+        self.synchronize_chain_state(self.admin_id).await?;
```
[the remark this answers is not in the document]
afck: Yes, I need to revisit whether it's actually necessary, once we've got all the tests working.

linera-client/src/chain_listener.rs
```diff
@@ -105,7 +105,12 @@ impl ChainListener {
     where
         C: ClientContext,
     {
-        let chain_ids = context.lock().await.wallet().chain_ids();
+        let chain_ids = {
+            let guard = context.lock().await;
+            let mut chain_ids = BTreeSet::from_iter(guard.wallet().chain_ids());
+            chain_ids.insert(guard.wallet().genesis_admin_chain());
```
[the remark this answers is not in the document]
afck: I could add a getter to the wallet that returns the wallet's chains plus the admin chain.

linera-core/src/client/mod.rs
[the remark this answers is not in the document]
afck: It's the only way to learn about admin chain events now. It doesn't send us a message anymore.

We could make it a separate command, but I'd rather not: It means that more users won't end up migrating to new epochs.

linera-base/src/data_types.rs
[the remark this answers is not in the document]
ma2bd: Does the client know that?

linera-execution/src/system.rs
```diff
@@ -349,6 +353,8 @@ pub enum SystemExecutionError {
     ArithmeticError(#[from] ArithmeticError),
     #[error(transparent)]
     ViewError(ViewError),
+    #[error(transparent)]
+    BcsError(#[from] bcs::Error),
```
MathieuDutSik, line 357: The BcsError could be put in the `ViewError`. Non blocker of course.

linera-core/src/client/mod.rs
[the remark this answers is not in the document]
afck: Done in https://github.com/linera-io/linera-protocol/pull/3432/commits/34d744edb62e48eb4d637000498a240bb643eb40.

linera-execution/src/system.rs
afck: Not sure: I've also heard the argument that we should create errors only in their own module.

linera-execution/src/system.rs
```diff
@@ -530,35 +530,33 @@ where
...
-                            message: SystemMessage::CreateCommittee { epoch, committee },
-                        };
-                        outcome.messages.push(message);
+                        txn_tracker.add_event(
+                            StreamId::system(EPOCH_STREAM_NAME),
+                            bcs::to_bytes(&epoch)?,
+                            bcs::to_bytes(&blob_hash)?,
+                        );
                     }
                     AdminOperation::RemoveCommittee { epoch } => {
```
MathieuDutSik, line 550: It is a little strange to have such a formulation.
I would expect `RemoveCommittee { committee }`.
Then maybe rename `RemoveFromCommittee`.
Not blocker of course.

afck: I agree, it should probably be called `RevokeEpoch`? But then I'd rename the other one also from `CreateCommittee` to `CreateEpoch`, even though the latter contains a committee and the epoch is implied.

Not sure what the best names are.

linera-base/src/data_types.rs
```diff
@@ -1000,6 +1002,11 @@ impl BlobContent {
         BlobContent::new(BlobType::ApplicationDescription, bytes)
     }
 
+    /// Creates a new committee [`BlobContent`] from the provided serialized committee.
+    pub fn new_committee(committee: impl Into<Box<[u8]>>) -> Self {
```
deuszx, line 1006: This API should be more type-safe. Right now it accepts any bytes blindly a s Committee which can led to problems. Why not pass the actual committee instead and serialise it here?

linera-core/src/client/mod.rs
```diff
@@ -3020,13 +3053,47 @@ where
...
+        while self
+            .has_admin_event(REMOVED_EPOCH_STREAM_NAME, min_epoch)
+            .await?
+        {
+            epoch_change_ops.push(Operation::System(SystemOperation::ProcessRemovedEpoch(
+                min_epoch,
+            )));
+            min_epoch.try_add_assign_one()?;
+        }
+        let mut epoch_change_ops = epoch_change_ops.into_iter();
```
deuszx, line 3087: Can we extract this to a helper method? Sth like `collect_committee_changes` ?

linera-sdk/src/test/validator.rs
```diff
@@ -201,10 +201,19 @@ impl TestValidator {
 
         let admin_chain = self.get_chain(&ChainId::root(0));
 
+        let committee_blob = Blob::new(BlobContent::new_committee(
+            bcs::to_bytes(&committee).unwrap(),
+        ));
+        let blob_hash = committee_blob.id().hash;
+        self.storage
+            .write_blob(&committee_blob)
```
deuszx, line 209: I think changing the resource policy should not require re-publishing a new committee. There should be a separate mechanism for that. If we change the cost of a Wasm multiplication will we also publish a new committee? It should be stored somewhere in the "system storage" and its updates should require quorum of validators.

afck: But it also needs to be applied separately to each chain, just like committee changes. It makes sense to me to treat all changes with that property as one thing. Anyway, this wasn't changed in this PR, so it's probably a separate discussion.

linera-execution/src/system.rs
```diff
@@ -652,10 +644,73 @@ where
...
+            }
             ReadBlob { blob_id } => {
                 self.read_blob_content(blob_id).await?;
                 self.blob_used(Some(txn_tracker), blob_id).await?;
             }
+            ProcessNewEpoch(epoch) => {
+                let chain_next_epoch = self.epoch.get().expect("chain is active").try_add_one()?;
+                ensure!(
+                    epoch == chain_next_epoch,
+                    SystemExecutionError::InvalidCommitteeCreation
```
deuszx, line 658: Maybe add a `String` argument do the error for a bit more context? Or a separate error.

linera-execution/src/system.rs
```diff
@@ -652,10 +644,73 @@ where
...
+                let event_id = EventId {
+                    chain_id: admin_id,
+                    stream_id: StreamId::system(REMOVED_EPOCH_STREAM_NAME),
+                    key: bcs::to_bytes(&epoch)?,
+                };
+                let bytes = match txn_tracker.next_replayed_oracle_response()? {
+                    None => self.context().extra().get_event(event_id.clone()).await?,
+                    Some(OracleResponse::Event(recorded_event_id, bytes))
+                        if recorded_event_id == event_id =>
+                    {
```
deuszx, line 704: Same here.

linera-execution/src/system.rs
afck: I created https://github.com/linera-io/linera-protocol/issues/3516 for this.

linera-client/src/chain_listener.rs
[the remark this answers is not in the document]
afck: I created https://github.com/linera-io/linera-protocol/issues/3517, but I don't know yet how that will look in detail.

linera-execution/src/system.rs
```diff
@@ -652,10 +644,73 @@ where
...
+                let event_id = EventId {
+                    chain_id: admin_id,
+                    stream_id: StreamId::system(EPOCH_STREAM_NAME),
+                    key: bcs::to_bytes(&epoch)?,
+                };
+                let bytes = match txn_tracker.next_replayed_oracle_response()? {
+                    None => self.context().extra().get_event(event_id.clone()).await?,
+                    Some(OracleResponse::Event(recorded_event_id, bytes))
+                        if recorded_event_id == event_id =>
+                    {
```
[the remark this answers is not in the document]
afck: I will add the comment to `next_replayed_oracle_response` itself.

linera-execution/src/system.rs
```diff
@@ -652,16 +640,82 @@ where
...
+                let admin_id = self
+                    .admin_id
+                    .get()
+                    .ok_or_else(|| SystemExecutionError::InactiveChain)?;
+                let event_id = EventId {
+                    chain_id: admin_id,
+                    stream_id: StreamId::system(EPOCH_STREAM_NAME),
+                    key: bcs::to_bytes(&epoch)?,
+                };
+                let bytes = match txn_tracker.next_replayed_oracle_response()? {
```
bart-linera, line 661: This pattern of checking for oracle response mismatches is now repeated in a few different places, maybe it could somehow be extracted into a method on `TransactionTracker`? (Although it might not be so simple, because one of the parameters would have to be a pattern... it would need some thought.)

Anyway, not a blocker.

bart-linera (approved):

linera-execution/src/system.rs
afck: Yes, that's a good point: We could pass in the closure that extracts the data by matching the pattern.

What's probably less easy to do is to also pass in the closure that uses the actual oracle, since that's going to be async and often borrow a lot of references.

But yes, I'd prefer to do that in a separate PR.

I created https://github.com/linera-io/linera-protocol/issues/3519.

linera-base/src/data_types.rs
afck: I created https://github.com/linera-io/linera-protocol/issues/3520 for this.

linera-execution/src/system.rs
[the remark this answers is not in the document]
afck: Done in https://github.com/linera-io/linera-protocol/pull/3432/commits/94cccbc7b392e0e2cd96b233bcbf8d2ad0a43a61, and I created https://github.com/linera-io/linera-protocol/issues/3519 as well to approve the API.

linera-execution/src/system.rs
afck: Done in https://github.com/linera-io/linera-protocol/pull/3432/.

linera-execution/src/system.rs
afck: I added more detail to it in https://github.com/linera-io/linera-protocol/pull/3432/commits/0af786dbf224d42661039c0d74e09dae7979a32d.

deuszx (approved):
LGTM

PR #3456, opened 2025-03-03
fix: fix command for removing Docker containers and volumes

hexcow:
## Description
I’ve fixed an issue with the previous command not filtering containers correctly based on the attached volume. The original approach using `docker ps -a -q --filter volume=$SCYLLA_VOLUME` does not work as intended because `docker ps` cannot filter containers by volume directly. I’ve updated the command to correctly remove the containers and associated volume using:

```
docker ps -a -q --filter "volume=$SCYLLA_VOLUME" | xargs docker rm -f
docker volume rm $SCYLLA_VOLUME
```

This should now work as expected.

## Motivation  
The original command failed to filter Docker containers by the attached volume, which led to unexpected behavior when trying to remove containers and volumes.

## Proposal  
Update the command to use `xargs` to pipe the container IDs and ensure correct deletion of containers attached to the specified volume, followed by the removal of the volume itself.

## Test Plan  
1. Verify that the containers attached to the specified volume are correctly listed and removed by running the updated command.
2. Confirm that the volume is successfully removed after deleting the containers.
3. Ensure no unexpected errors occur and that the command runs smoothly in different environments.

### DIFF
--- a/docker/compose.sh
+++ b/docker/compose.sh
@@ -19,7 +19,7 @@ cleanup() {
     rm wallet.json
     SCYLLA_VOLUME=docker_linera-scylla-data
     SHARED_VOLUME=docker_linera-shared
-    docker rm -f $(docker ps -a -q --filter volume=$SCYLLA_VOLUME)
+    docker ps -a -q --filter "volume=$SCYLLA_VOLUME" | xargs docker rm -f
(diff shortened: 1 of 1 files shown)


ma2bd:
I think we're good as it is

PR #3463, opened 2025-03-04
Allow reading native balances in integration tests

### LINKED ISSUE #3417: Feature request: Can we support `chain_balance` and `owner_balance` for `ActiveChain` ?
**Description**
Now there is no method to query balance of native tokens in ActiveChain. For the use case which can accept native tokens to mint fungible tokens, application cannot write tests easily to check native tokens. Of course we can write `native_balance_of` in application `service.rs` then query it in the integration tests, but for me it's not quite native due to every application has to create their own `native_balance_of` function.

**Suggestion**
Can we support `chain_balance` and `owner_balance` for `ActiveChain` then any application can call from their integration tests natively ?

jvff:
## Motivation

When writing applications that use native tokens, it helps to write integration tests if there's an easy way to query the on-chain native token balances.

## Proposal

Add a few helper methods to read the native account balances on an `ActiveChain`.

## Test Plan

Integration tests that use the new helpers were added to the Native Fungible Token application example.

## Release Plan

- Nothing to do / These changes follow the usual release cycle, because this introduces a new backwards compatible feature.

## Links

- Closes #3417

### DIFF
--- a/linera-sdk/src/test/chain.rs
+++ b/linera-sdk/src/test/chain.rs
@@ -14,14 +14,17 @@ use std::{
 use cargo_toml::Manifest;
 use linera_base::{
     crypto::{AccountPublicKey, AccountSecretKey},
-    data_types::{Blob, BlockHeight, Bytecode, CompressedBytecode},
+    data_types::{Amount, Blob, BlockHeight, Bytecode, CompressedBytecode},
     identifiers::{ApplicationId, BytecodeId, ChainDescription, ChainId, MessageId},
     vm::VmRuntime,
 };
 use linera_chain::{types::ConfirmedBlockCertificate, ChainError, ChainExecutionContext};
 use linera_core::{data_types::ChainInfoQuery, worker::WorkerError};
 use linera_execution::{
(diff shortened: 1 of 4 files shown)


linera-sdk/src/test/chain.rs
```diff
@@ -89,6 +93,103 @@ impl ActiveChain {
...
+    }
+
+    /// Reads the current account balance on this microchain of an [`AccountOwner`].
+    pub async fn owner_balance(&self, owner: &AccountOwner) -> Option<Amount> {
+        let chain_state = self
+            .validator
+            .worker()
+            .chain_state_view(self.id())
+            .await
+            .expect("Failed to read chain state");
```
afck, line 121: Would it be useful for testing to (in addition or instead of this) also have a method that returns the whole chain state view?

afck (approved):

linera-sdk/src/test/chain.rs
jvff: I'm not sure if it would be too much of a leaky abstraction to application developers. I think it might make more sense to provide individual methods and/or a GraphQL (like?) query interface. But we can think more about this for a future PR.

PR #3465, opened 2025-03-04
Make Account* keys an enum of Ed25519 and Secp256k1.

deuszx:
## Motivation

We want users to use their EVM wallets (like Metamask or others) when interacting with Linera (which technically means allowing for secp256k1 keys to be used as chain owners).

## Proposal

Change `Account*` abstractions from a newtype struct to an enum with two variants: Ed25519 and secp256k1.

## Test Plan

A crowdfunding example was modified to have two chains with different key schemes for chains' owners. This unfortunately didn't excersise signature verification as we don't do that for the Linera application testing framework.

Benchmarks were also modified to generate chains with both key variants and exersise what was not possible in the example.

## Release Plan

### DIFF
--- a/linera-base/src/crypto/mod.rs
+++ b/linera-base/src/crypto/mod.rs
@@ -12,9 +12,10 @@ use std::{fmt::Display, io, num::ParseIntError, str::FromStr};
 
 use alloy_primitives::FixedBytes;
 use custom_debug_derive::Debug;
-use ed25519::{Ed25519PublicKey, Ed25519Signature};
+pub use ed25519::{Ed25519PublicKey, Ed25519SecretKey, Ed25519Signature};
 pub use hash::*;
 use linera_witty::{WitLoad, WitStore, WitType};
+pub use secp256k1::{Secp256k1PublicKey, Secp256k1SecretKey, Secp256k1Signature};
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
 
@@ -133,31 +134,6 @@ impl AccountSecretKey {
                 let signature = secp256k1::Secp256k1Signature::new(value, secret);
                 AccountSignature::Secp256k1(signature)
             }
-        }
-    }
-
-    #[cfg(all(with_getrandom, with_testing))]
-    /// Generates a new key pair.
-    ///
-    /// Uses `OsRng` for that. If you want control over the RNG, use `generate_from`.
-    pub fn generate() -> Self {
-        use rand::RngCore;
-        let mut rng = rand::rngs::OsRng;
-        if rng.next_u32() % u32::MAX == 0 {
-            AccountSecretKey::Ed25519(ed25519::Ed25519SecretKey::generate())
-        } else {
-            AccountSecretKey::Secp256k1(secp256k1::Secp256k1KeyPair::generate().secret_key)
-        }
-    }
-
-    #[cfg(with_getrandom)]
-    /// Generates a new key pair from the given RNG. Use with care.
-    pub fn generate_from<R: CryptoRng>(rng: &mut R) -> Self {
-        //TODO: Use a better way to choose between the two schemes.
-        if rng.next_u32() % u32::MAX == 0 {
-            AccountSecretKey::Ed25519(ed25519::Ed25519SecretKey::generate_from(rng))
-        } else {
-            AccountSecretKey::Secp256k1(secp256k1::Secp256k1KeyPair::generate_from(rng).secret_key)
         }
     }
 }
@@ -320,6 +296,42 @@ impl TryFrom<&[u8]> for AccountSignature {
 
     fn try_from(bytes: &[u8]) -> Result<Self, Self::Error> {
         AccountSignature::from_slice(bytes)
+    }
+}
+
+impl From<ed25519::Ed25519PublicKey> for AccountPublicKey {
+    fn from(public_key: ed25519::Ed25519PublicKey) -> Self {
+        AccountPublicKey::Ed25519(public_key)
+    }
+}
+
+impl From<secp256k1::Secp256k1PublicKey> for AccountPublicKey {
+    fn from(public_key: secp256k1::Secp256k1PublicKey) -> Self {
+        AccountPublicKey::Secp256k1(public_key)
+    }
+}
+
+impl From<ed25519::Ed25519Signature> for AccountSignature {
+    fn from(signature: ed25519::Ed25519Signature) -> Self {
+        AccountSignature::Ed25519(signature)
+    }
+}
+
+impl From<secp256k1::Secp256k1Signature> for AccountSignature {
+    fn from(signature: secp256k1::Secp256k1Signature) -> Self {
+        AccountSignature::Secp256k1(signature)
+    }
+}
+
+impl From<secp256k1::Secp256k1SecretKey> for AccountSecretKey {
+    fn from(secret_key: secp256k1::Secp256k1SecretKey) -> Self {
+        AccountSecretKey::Secp256k1(secret_key)
+    }
+}
+
+impl From<ed25519::Ed25519SecretKey> for AccountSecretKey {
+    fn from(secret_key: ed25519::Ed25519SecretKey) -> Self {
+        AccountSecretKey::Ed25519(secret_key)
     }
 }
 
--- a/linera-base/src/crypto/mod.rs
+++ b/linera-base/src/crypto/mod.rs
@@ -12,6 +12,7 @@ use std::{fmt::Display, io, num::ParseIntError, str::FromStr};
 
 use alloy_primitives::FixedBytes;
 use custom_debug_derive::Debug;
+use ed25519::{Ed25519PublicKey, Ed25519Signature};
 pub use hash::*;
 use linera_witty::{WitLoad, WitStore, WitType};
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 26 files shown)


linera-base/src/crypto/mod.rs
```diff
@@ -28,6 +30,36 @@ pub type ValidatorSignature = secp256k1::Secp256k1Signature;
...
+    Secp256k1,
+}
+
+impl SignatureScheme {
+    /// Returns the flag of the signature scheme.
+    pub fn flag(&self) -> u8 {
+        match self {
+            SignatureScheme::Ed25519 => 0x00,
+            SignatureScheme::Secp256k1 => 0x01,
+        }
```
afck, line 48: Would it make sense to use `#[repr(u8)]` so we could just use `scheme as u8` as the flag?
(And should it be "tag" rather than "flag"?)

linera-base/src/crypto/mod.rs
```diff
@@ -28,6 +30,36 @@ pub type ValidatorSignature = secp256k1::Secp256k1Signature;
 /// The key pair of a validator.
 pub type ValidatorKeypair = secp256k1::Secp256k1KeyPair;
 
+/// Signature scheme used for the public key.
```
afck, line 33: Or "…for account keys"? (I mean, the scheme is used by the secret key as well…)

linera-base/src/crypto/mod.rs
```diff
@@ -135,25 +263,31 @@ impl FromStr for AccountPublicKey {
...
         let value = hex::decode(s)?;
-        Ok(AccountPublicKey((value.as_slice()).try_into()?))
+        AccountPublicKey::from_slice(value.as_slice())
     }
 }
 
 impl Display for AccountPublicKey {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        self.0.fmt(f)
+        write!(f, "{}", hex::encode(self.as_bytes()))
```
afck, line 272: We could probably still do this, to keep the formatting parameters:

```suggestion
        hex::encode(self.as_bytes()).fmt(f)
```

linera-base/src/crypto/secp256k1.rs
```diff
@@ -187,6 +192,120 @@ impl fmt::Debug for Secp256k1PublicKey {
...
+
+impl From<&Secp256k1PublicKey> for Owner {
+    fn from(value: &Secp256k1PublicKey) -> Self {
+        Self(CryptoHash::new(value))
+    }
+}
+
+impl<'de> BcsHashable<'de> for Secp256k1PublicKey {}
+
+impl WitType for Secp256k1PublicKey {
```
afck, line 209: Can't the Witty types be derived for the two key types?

linera-base/src/ownership.rs
```diff
@@ -198,13 +198,13 @@ pub enum ChangeApplicationPermissionsError {
...
 mod tests {
     use super::*;
-    use crate::crypto::AccountSecretKey;
+    use crate::crypto::{Ed25519SecretKey, Secp256k1SecretKey};
 
     #[test]
     fn test_ownership_round_timeouts() {
-        let super_pub_key = AccountSecretKey::generate().public();
+        let super_pub_key = Ed25519SecretKey::generate().public();
         let super_owner = Owner::from(super_pub_key);
```
afck, line 206: Don't we want the `Owner` type to be derived from the general `PublicKey` enum? Or is this step temporary, until `Owner` gets replaced by a new address type?

linera-service/src/server.rs
```diff
@@ -288,7 +291,7 @@ fn make_server_config<R: CryptoRng>(
     options: ValidatorOptions,
 ) -> anyhow::Result<persistent::File<ValidatorServerConfig>> {
     let validator_keypair = ValidatorKeypair::generate_from(rng);
-    let account_secret = AccountSecretKey::generate_from(rng);
+    let account_secret = AccountSecretKey::from(Ed25519SecretKey::generate_from(rng));
```
afck, line 294: Would it make sense to have an `AccountSecretKey::generate_from` for tests, that chooses a random key type?

afck (approved):

linera-base/src/crypto/secp256k1.rs
deuszx: TBH I just did the same thing that was already done for Ed25519.

linera-service/src/server.rs
deuszx: I had that initially but then thought it might not be desirable to _not_ know what type of keys you'll get. Although I guess that's OK if you can choose a specific key scheme anyway (like this).

I'm not sure if it's worth adding now, given that the code has been updated and compiles.

linera-base/src/crypto/secp256k1.rs
afck: We should find out why it was done explicitly there. Or was that even before the `derive` macros for the Witty types were implemented?

linera-base/src/ownership.rs
afck: But if `Owner` can only be derived from an Ed25519 key, then how can a secp256k1 key ever sign a block proposal?

deuszx: I think Owner can be derived from both. There are impls for `Secp256k1` types as well.

linera-base/src/crypto/mod.rs
```diff
@@ -46,87 +78,183 @@ pub type ValidatorKeypair = secp256k1::Secp256k1KeyPair;
...
+        match self {
+            AccountPublicKey::Ed25519(_) => SignatureScheme::Ed25519,
+            AccountPublicKey::Secp256k1(_) => SignatureScheme::Secp256k1,
+        }
     }
 
     /// Returns the byte representation of the public key.
-    pub fn to_bytes(&self) -> Vec<u8> {
-        self.0.as_bytes()
+    pub fn as_bytes(&self) -> Vec<u8> {
```
ma2bd, line 151: Why are re-defining a serialization function here? (also it turns out the bytes are exactly the BCS bytes)

linera-base/src/crypto/mod.rs
```diff
@@ -46,87 +78,183 @@ pub type ValidatorKeypair = secp256k1::Secp256k1KeyPair;
...
+        }
+
+        bytes
     }
 
     /// Parses the byte representation of the public key.
     ///
-    /// Returns error if the byte slice has incorrect length.
+    /// Returns error if the byte slice has incorrect length or the flag is not recognized.
     pub fn from_slice(bytes: &[u8]) -> Result<Self, CryptoError> {
```
ma2bd, line 170: same question

linera-base/src/crypto/mod.rs
```diff
@@ -28,6 +30,36 @@ pub type ValidatorSignature = secp256k1::Secp256k1Signature;
...
+pub enum SignatureScheme {
+    /// Ed25519
+    Ed25519,
+    /// secp256k1
+    Secp256k1,
+}
+
+impl SignatureScheme {
+    /// Returns the flag of the signature scheme.
+    pub fn flag(&self) -> u8 {
```
ma2bd, line 44: why do we need this at? See other comments

linera-base/src/crypto/mod.rs
```diff
@@ -46,87 +78,183 @@ pub type ValidatorKeypair = secp256k1::Secp256k1KeyPair;
...
+        match self {
+            AccountSignature::Ed25519(signature) => {
+                let public_key = match author {
+                    AccountPublicKey::Ed25519(public_key) => public_key,
+                    _ => {
+                        return Err(CryptoError::InvalidSignature {
+                            error: "invalid public key type".to_string(),
+                            type_name: std::any::type_name::<T>().to_string(),
+                        })
+                    }
```
ma2bd, line 207: It seems like we could `match (self, author) { .. }` and avoid the code duplication

linera-base/src/crypto/mod.rs
deuszx: Are we re-defining? I'd rather see serialization code call into this. I don't like leaving something as important as serialization to serde. We don't really control what's going on inside and might even not notice when it changes (between versions for example).

linera-base/src/crypto/mod.rs
deuszx: Done in [Use bcs instead of manually crafting the (de)serialization](https://github.com/linera-io/linera-protocol/pull/3465/commits/f7d33d04b25150fac5fcb5e45b5d914366c13aaa).

linera-base/src/crypto/mod.rs
deuszx: Done in [Use bcs instead of manually crafting the (de)serialization](https://github.com/linera-io/linera-protocol/pull/3465/commits/f7d33d04b25150fac5fcb5e45b5d914366c13aaa).

linera-base/src/crypto/mod.rs
```diff
@@ -165,6 +299,42 @@ impl TryFrom<&[u8]> for AccountSignature {
...
+        AccountSecretKey::Secp256k1(secret_key)
+    }
+}
+
+impl From<ed25519::Ed25519SecretKey> for AccountSecretKey {
+    fn from(secret_key: ed25519::Ed25519SecretKey) -> Self {
+        AccountSecretKey::Ed25519(secret_key)
+    }
+}
+
```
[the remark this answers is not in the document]
deuszx: OK, let me see what happens when I remove them. It was mostly for easier tests migration.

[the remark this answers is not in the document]
deuszx: OK, there's a lot of them (~100) but I can fix that.

PR #3470, opened 2025-03-04
End value fix for linear_bucket_interval

ndr-ds:
## Motivation

So I was under the impression that getting the ceiling there would make us always include the end value, but I was wrong :)

## Proposal

Doing the same fix here for the end value that we did for the exponential buckets

## Test Plan

Deployed a local network, saw that we actually see 500 as the max value now for example for proxy latency

## Release Plan

### DIFF
--- a/linera-base/src/prometheus_util.rs
+++ b/linera-base/src/prometheus_util.rs
@@ -62,11 +62,11 @@ pub fn exponential_bucket_latencies(max_latency: f64) -> Option<Vec<f64>> {
 /// Construct the bucket interval linearly starting from a value and an ending value.
 pub fn linear_bucket_interval(start_value: f64, width: f64, end_value: f64) -> Option<Vec<f64>> {
     let count = (end_value - start_value) / width;
-    let count = count.ceil() as usize;
-    Some(
-        linear_buckets(start_value, width, count)
-            .expect("Linear buckets creation should not fail!"),
-    )
+    let count = count.round() as usize;
+    let mut buckets = linear_buckets(start_value, width, count)
+        .expect("Linear buckets creation should not fail!");
+    buckets.push(end_value);
+    Some(buckets)
 }
 
 /// Construct the latencies linearly starting from 1 and ending at the maximum latency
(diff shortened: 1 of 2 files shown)


linera-base/src/prometheus_util.rs
```diff
@@ -63,10 +63,14 @@ pub fn exponential_bucket_latencies(max_latency: f64) -> Option<Vec<f64>> {
...
-        linear_buckets(start_value, width, count)
-            .expect("Linear buckets creation should not fail!"),
-    )
+    let mut buckets = linear_buckets(start_value, width, count)
+        .expect("Linear buckets creation should not fail!");
+    if let Some(last) = buckets.last() {
+        if *last < end_value {
+            buckets.push(end_value);
+        }
+    }
```
afck, line 72: Wouldn't all of this be unnecessary if we just used `round()` instead of `ceil()` above?

ndr-ds: I don't think so. `ceil` will give us a higher count on average than `round`, because `round` can also round down. And even with `ceil`, we were still not getting the `end_value` in there. So I think we need this fix

afck: Ah, because it should actually be `(end_value - start_value) / width + 1`:

If you have start value 1, end value 4 and width 1, you need to get 4 buckets, but (4 - 1) / 1 = 3.

ndr-ds: Ah, good point. But isn't it better if the last bucket is always the `end_value` limit? 🤔 doing the `+ 1` will probably frequently give us a last bucket that is greater than `end_value`, I think, which would make it so that the function isn't really creating buckets in the provided range, which seems wrong 🤔

afck: I'd rather say it's a weird signature for this kind of function: What _is_ the expected behavior if the user calls it in a way where `end_value - start_value` is not an integer multiple of `width`?

I'd rather just make `count` itself a parameter, and compute `width` from it:

`width = (end_value - start_value) / (count - 1);`

ndr-ds: I think it's easier to think "what range of values do I think this metric will be in" than "how many buckets do I want to create in this range", tbh. A lot of these just use `linear_bucket_latencies`, because as the user I just want to say if it's linear or exponential, and say the range I'm interested in, and the buckets (and even width) can be figured out by the function.
Seems like a better user experience, no?

afck: But then I'd rather return an error or panic if it isn't (close to) a multiple of the width?
I certainly wouldn't want buckets with values `[1.0, 1.999, 2.998, 3.0]`.

ndr-ds: For linear buckets that should never happen, I can even make the parameter a usize. For exponential buckets, it might happen, depends on the starting value you use

afck (approved):
Looks good to me!
(Did you try out that the result is as expected?)

afck (approved):

afck (approved):

ndr-ds:
### Merge activity

* **Mar 5, 1:24 PM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/linera-io/linera-protocol/3470).

PR #2786, opened 2024-11-02
Split rust CI into more jobs

andresilva91:
## Motivation

CI takes a long time

## Proposal

Split into more jobs, which will make us do roughly the same amount of work, but more parallelized. So hopefully this reduces CI time without increasing github costs much.

## Test Plan

CI
With this and the PR before this one (#2785), we went from having the `test` job sometimes take up to an hour, to now, running many jobs in parallel, the test that takes the longest is `check-all-features`, which takes around 30 minutes.

## Release Plan

- Nothing to do / These changes follow the usual release cycle.

PR #3472, opened 2025-03-04
Restore CI job to check WIT files

jvff:
## Motivation

The WIT files are automatically generated by Witty from the Rust code. Therefore, they must be kept up-to-date with the Rust code for applications to work correctly.

There was a CI job to check this before, but it was accidentally removed when the CI jobs were refactored (#2786).

## Proposal

Add a new CI job to run `wit-generator -c`.

## Test Plan

CI should show if the job executed successfully.

## Release Plan

- Nothing to do because this only affects CI. Even though the WIT files were unsorted, they were correct.

## Links


ndr-ds (approved):
🤦🏻‍♂️🤦🏻‍♂️🤦🏻‍♂️ my bad

PR #3476, opened 2025-03-05
Deserialize `Account` from its string representation

### LINKED ISSUE #3462: GraphQL cannot parse `Account` structure from its format string
**Description**
When define argument with `Account` in `service.rs`, GraphQL cannot parse `Account` struct from its format string.

**Query**
```
let query = format!(                                                                                                                    
      "query {{ balanceOf(owner: \"{}\")}}",                                                                                                                                                                                                                                    
      Account {                                                                                                                           
          chain_id: meme_application_id.creation.chain_id,                                                                                
          owner: Some(AccountOwner::Application(meme_application_id.forget_abi())),                                                       
      },                                                                                                                                  
  );                                                                                                                                      
  let QueryOutcome { response, .. } = meme_chain.graphql_query(meme_application_id, query).await;                                         
  assert_eq!(                                                                                                                             
      Amount::from_str(response["balanceOf"].as_str().unwrap()).unwrap(),                                                                 
      initial_supply.try_sub(initial_liquidity).unwrap(),                                                                                 
  );
```

**Error**
```
Failed to parse \"Account\": invalid type: string \"59ac87e09448b72d4b910ecc20855a6292d9f29c32a7894e592f8a7d1b51cb11:Application:2e5649769255ce2eeb353c1b1962edec449fb19f00c7a283f2f7db68ea782dc1012e2b736ebef10e98b41ed5cde4bab0fb91a12950a7772f426137d80c2e533359ac87e09448b72d4b910ecc20855a6292d9f29c32a7894e592f8a7d1b51cb11040000000000000000000000\", expected struct Account
```

jvff:
## Motivation

It is useful to be able to use an `Account`'s string representation emitted by its `Display` implementation in a GraphQL request. However, that requires the type to be able to be deserialized from a string.

## Proposal

Manually implement `Deserialize` for `Account` so that it can also be deserialized from its string representation.

## Test Plan

Added unit tests for the roundtrip serialization of `Account` to GraphQL, in both its object and string representations.

## Release Plan

- These changes follow the usual release cycle, because it introduces a new backwards compatible feature.

## Links

- Closes #3462


linera-base/src/unit_tests.rs
```diff
@@ -47,6 +48,36 @@ where
         .expect("Flattening WIT roundtrip test failed");
 }
 
+/// Test GraphQL serialization roundtrip.
+#[test_case(account_test_case(); "of_account")]
```
deuszx, line 52: Are there more types that could use these type of tests? Maybe we can enforce it? i.e. for example when we generate `*.graphql` schemas we could run these tests over all types that are outputted there.

linera-base/src/identifiers.rs
```diff
@@ -101,6 +99,132 @@ impl FromStr for Account {
...
+                        Accessor::Error::duplicate_field("chain_id")
+                    );
+                    chain_id = Some(accessor.next_value()?);
+                }
+                AccountFieldDiscriminator::Owner => {
+                    ensure!(owner.is_none(), Accessor::Error::duplicate_field("owner"));
+                    owner = Some(accessor.next_value()?);
+                }
+            }
+        }
```
deuszx, line 157: alternatively sth like:
```rust
match (accessor.next_key()?, accessor.next_key()?) { ... }
```

jvff: I'm not sure that's possible, because I think the API requires a call to `next_value` after the `next_key` if you want that entry :thinking:

ma2bd:
> An alternative would be to just use strings if `is_human_readable`, like we do e.g. for `Amount` and other types. I think that would be a way simpler implementation?

Yes, please. Binary formats for composed types should follow Serde.

ma2bd (changes requested):
As discussed, let's handle human readable and binary encodings separately.

PR #3484, opened 2025-03-06
Add exists chain to TestValidator to test new created chain

kikakkz:
## Motivation

If we create new chain with `open_chain` in `contract.rs`, it won't be maintained by `TestValidator` in integration tests. Thus we won't be able to test some functionalities of the new chain in integration tests. For example, if we would like to call operation from another user chain to the created chain's application, the user chain should register the application firstly, then call it. At that time we'll get error `Chain not found` with the `TestValidator`.

## Proposal

If we could add exists chain to `TestValidator`, then I think we'll be able to test those created chains in integration tests.

## Test Plan

CI

## Release Plan

### DIFF
--- a/linera-sdk/src/test/validator.rs
+++ b/linera-sdk/src/test/validator.rs
@@ -189,7 +189,7 @@ impl TestValidator {
         chain
     }
 
-    /// Add an exists [`ActiveChain`]
+    /// Add an existing [`ActiveChain`]
     pub fn add_chain(&self, chain: ActiveChain) {
         self.chains.insert(chain.id(), chain.clone());
     }
--- a/linera-sdk/src/test/validator.rs
+++ b/linera-sdk/src/test/validator.rs
@@ -189,7 +189,7 @@ impl TestValidator {
         chain
     }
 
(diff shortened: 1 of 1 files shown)


linera-sdk/src/test/validator.rs
```diff
@@ -189,6 +189,11 @@ impl TestValidator {
         chain
     }
 
+    /// Add an exists [`ActiveChain`]
```
afck, line 192:
```suggestion
    /// Add an existing [`ActiveChain`]
```

linera-sdk/src/test/validator.rs
```diff
@@ -195,6 +195,11 @@ impl TestValidator {
         self.new_chain_with_keypair(key_pair).await
     }
 
+    /// Add an existing [`ActiveChain`]
```
jvff, line 198: Nit: changing the verb conjugation to follow the convention and adding a period at the end.

```suggestion
    /// Adds an existing [`ActiveChain`].
```

jvff (approved):

PR #3485, opened 2025-03-06

usagi32:


linera-client/src/benchmark.rs
```diff
@@ -189,7 +189,7 @@ where
             join_set.spawn_blocking(move || {
                 handle.block_on(
                     async move {
-                        Self::run_benchmark_internal(
+                        Box::pin(Self::run_benchmark_internal(
```
ma2bd, line 192: Did you get an overflow? or a lint error?

ma2bd:
Thanks for the PR. Can you please link to the GH issue and fill the summary as the team does?

linera-client/src/benchmark.rs
usagi32: Yeah, a lint error(Clippy future is too large. There is a lint about it in its lib.rs). Its occurs only for me and not for others. Don't know why.

linera-core/src/chain_worker/state/temporary_changes.rs
```diff
@@ -169,6 +169,10 @@ where
             .remove_bundles_from_inboxes(block.timestamp, &block.incoming_bundles)
             .await?;
         let outcome = if let Some(outcome) = outcome {
+            ensure!(
+                block.operations.len() == outcome.user_action_outcomes.len(),
```
ma2bd, line 173: Please replace `user_action_outcome` everywhere by `operation_outcome`.

afck: `operation_results` or `operation_outcomes` sounds good to me.

afck:
I'd make the length equal to the number of operations, because messages don't return a value, yes.

linera-core/src/chain_worker/state/temporary_changes.rs
usagi32: Wouldn't there be ambiguity if its `system_operation` or `user_operation`.

ma2bd: Both types of operations should be able to return a result.

usagi32: The issue was specifically about collecting the results of `execute_operation` contract entrypoint. System operations don't do that and the few things they return are already collected by the `TransactionTracker` as something else. So, this only concerns user operations.

afck: That's true, at least for now system operations don't return anything. But I'd still prefer if the index in the outcome vector matched the index in the operations vector, I guess.

ma2bd: I hear you: for now, it will be always an empty result for system operations.

usagi32: Yeah index can match.. but what about the name?

usagi32: As this is result is only returned by user applications

ma2bd: not in the future

PR #3487, opened 2025-03-06
chore: fix incorrect string comparison in JSON-RPC check

annwag:
## Motivation  
The comparison of `"2.0"` with `result.jsonrpc` was incorrect. Since `"2.0"` is already a string literal, the dereference (`*`) was unnecessary and didn’t make sense. This could lead to unexpected behavior.  

## Proposal  
Removed the unnecessary dereference to ensure proper string comparison in the JSON-RPC check.  

## Test Plan  
Verified that the comparison now works correctly by testing with valid and invalid `jsonrpc` versions. Ensured no regressions in existing functionality.  

## Release Plan  
No breaking changes. This is a straightforward fix that improves correctness. Safe to include in the next release.

### DIFF
--- a/linera-ethereum/src/client.rs
+++ b/linera-ethereum/src/client.rs
@@ -45,7 +45,7 @@ pub trait JsonRpcClient {
         let res = serde_json::from_str(raw.get())?;
         ensure!(id == result.id, EthereumQueryError::IdIsNotMatching);
         ensure!(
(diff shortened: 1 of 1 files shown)


jvff (approved):

jvff:
> [jvff](https://github.com/jvff) pls merge it

Thanks for the PR! I'll do that soon, I just want to get the CI passing first.

Issue #3488, opened 2025-03-06
Test that events cannot be overwritten.


afck:
Add a test with a faulty app that tries to emit two events with the same ID but different value. Assert that this fails.

mrgbomb:
this solution cal help

afck:
Obsolete because of https://github.com/linera-io/linera-protocol/pull/3672.

Issue #3172, opened 2025-01-23
Don't compare full executed block to determine whether there was a conflict.


afck:
With #3164, the execution outcome can depend on the current round.

If this oracle is used, we shouldn't try to discern `ExecuteBlockOutcome::Executed` from `ExecuteBlockOutcome::Conflict` by comparing the exact executed block.

One alternative is to check that the `ProposedBlock` matches (i.e. the operations and messages are the ones we were meaning to commit) and that we are the block's signer. (https://github.com/linera-io/linera-protocol/issues/456 might be a prerequisite for doing that.)

PR #3489, opened 2025-03-06
Check for overwriting events, fix typos, duplications, and minor inconsistencies.

afck:
## Motivation

Working on https://github.com/linera-io/linera-protocol/pull/3432 I found a few issues that can be fixed separately.

## Proposal

* Move the operation inspection code from the two `published_blob_ids` implementations to `Operation` and deduplicate it.
* Use the `select!` macro in the client listener. This will make it easier to add another stream, to listen for admin chain events.
* Fix a typo where a blob ID was called `chain_id`.
* Add a check that events don't get overwritten. Testing this will be complicated, and possibly require more SDK features: https://github.com/linera-io/linera-protocol/issues/3488
* Treat `stage_new_committee` like other client operations. The issue with conflicting blocks needs to be addressed for all operations anyway, and admin operations are always on a single-user chain in our case. https://github.com/linera-io/linera-protocol/issues/3172
* Use `stage_new_committee` in the node service` instead of manually creating the operation. There is no reason to have different semantics here in CLI vs. the node service API, and to have the user input the current epoch manually.

## Test Plan

CI should catch regressions. I created #3488 for the events check.

## Release Plan

## Links

- Found while working on https://github.com/linera-io/linera-protocol/pull/3432.

### DIFF
--- a/linera-core/src/chain_worker/state/mod.rs
+++ b/linera-core/src/chain_worker/state/mod.rs
@@ -567,7 +567,7 @@ fn missing_blob_ids(maybe_blobs: &BTreeMap<BlobId, Option<Blob>>) -> Vec<BlobId>
     maybe_blobs
         .iter()
         .filter(|(_, maybe_blob)| maybe_blob.is_none())
-        .map(|(chain_id, _)| *chain_id)
+        .map(|(blob_id, _)| *blob_id)
         .collect()
 }
 
--- a/linera-core/src/chain_worker/state/temporary_changes.rs
+++ b/linera-core/src/chain_worker/state/temporary_changes.rs
@@ -3,6 +3,8 @@
 
 //! Operations that don't persist any changes to the chain state.
 
+use std::collections::HashMap;
+
 use linera_base::{
     data_types::{ArithmeticError, Timestamp, UserApplicationDescription},
     ensure,
@@ -14,7 +16,7 @@ use linera_chain::data_types::{
 };
 use linera_execution::{ChannelSubscription, Query, QueryOutcome};
 use linera_storage::{Clock as _, Storage};
-use linera_views::views::View;
+use linera_views::views::{View, ViewError};
 #[cfg(with_testing)]
 use {
     linera_base::{crypto::CryptoHash, data_types::BlockHeight},
@@ -171,6 +173,26 @@ where
         } else {
             Box::pin(chain.execute_block(block, local_time, round.multi_leader(), None)).await?
         };
+
+        // Verify that no event values are overwritten.
+        let mut new_events = HashMap::new();
+        for event in outcome.events.iter().flatten() {
+            let event_id = event.id(chain.chain_id());
+            if let Some(old_value) = new_events.insert(event_id.clone(), &event.value) {
+                ensure!(
+                    *old_value == event.value,
+                    WorkerError::OverwritingEvent(Box::new(event_id))
+                );
+            }
+            match self.0.storage.read_event(event_id.clone()).await {
+                Ok(old_value) => ensure!(
+                    old_value == event.value,
+                    WorkerError::OverwritingEvent(Box::new(event_id))
+                ),
+                Err(ViewError::EventsNotFound(_)) => {}
+                Err(err) => return Err(err.into()),
+            }
+        }
 
         let executed_block = outcome.with(block.clone());
         ensure!(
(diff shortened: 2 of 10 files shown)


linera-core/src/client/mod.rs
```diff
@@ -2987,29 +2987,11 @@ where
         &self,
         committee: Committee,
     ) -> Result<ClientOutcome<ConfirmedBlockCertificate>, ChainClientError> {
-        loop {
```
deuszx, line 2990: We're removing the loop here so it's not only a simplification.

deuszx: Ahh, I see the same is done inside `execute_operation`.

linera-service/src/node_service.rs
```diff
@@ -512,12 +512,21 @@ where
...
         &self,
         chain_id: ChainId,
-        epoch: Epoch,
         committee: Committee,
     ) -> Result<CryptoHash, Error> {
-        let operation =
-            SystemOperation::Admin(AdminOperation::CreateCommittee { epoch, committee });
-        self.execute_system_operation(operation, chain_id).await
+        Ok(self
+            .apply_client_command(&chain_id, move |client| {
```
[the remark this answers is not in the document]
afck: Yes, we should decide whether callers of `ChainClient` methods should construct the `Operation` instances, or whether they should be constructed by the chain client itself. Currently the latter is the case in most places, but it wasn't the case here.

linera-core/src/chain_worker/state/mod.rs
```diff
@@ -567,7 +567,7 @@ fn missing_blob_ids(maybe_blobs: &BTreeMap<BlobId, Option<Blob>>) -> Vec<BlobId>
     maybe_blobs
         .iter()
         .filter(|(_, maybe_blob)| maybe_blob.is_none())
-        .map(|(chain_id, _)| *chain_id)
+        .map(|(blob_id, _)| *blob_id)
```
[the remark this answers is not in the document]
deuszx: Not really :)

linera-service/src/node_service.rs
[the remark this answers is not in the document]
deuszx: Then maybe there should be no `execute_system_operation` available on `self.`

linera-core/src/chain_worker/state/temporary_changes.rs
```diff
@@ -172,6 +174,26 @@ where
             Box::pin(chain.execute_block(block, local_time, round.multi_leader(), None)).await?
         };
 
+        // Verify that no event values are overwritten.
+        let mut new_events = HashMap::new();
+        for event in outcome.events.iter().flatten() {
+            let event_id = event.id(chain.chain_id());
+            if let Some(old_value) = new_events.insert(event_id.clone(), &event.value) {
```
deuszx, line 181: How could this happen though? This should be a fatal error b/c it means there's problem with the code

linera-core/src/client/mod.rs
deuszx: Similarly to the other discussion - we're trying to achieve the same thing using different layers of abstractions and that's a sign of bad separation of concerns IME.

linera-service/src/node_service.rs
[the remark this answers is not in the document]
afck: Good point. I created https://github.com/linera-io/linera-protocol/issues/3493 for it.

linera-core/src/chain_worker/state/temporary_changes.rs
afck: As long as we only have this particular system event, yes.

But we may add other operations where the user can cause a particular event to be written, or allow contracts to create events. In those cases, we should simply fail the block.

deuszx: Like I wrote many times - I think events should have index that identifies its "order" in the block. Then we'll never have this problem.

linera-core/src/client/mod.rs
afck: This loop was probably meant to work around https://github.com/linera-io/linera-protocol/issues/3172, which needs a proper solution. (See PR description.)

It's not really useful in this particular case since the admin chain is single-owner anyway.

linera-core/src/chain_worker/state/temporary_changes.rs
afck: Yes, I think you're right. We should remove the key entirely for now, and just make it a number that increments each time we write an event to a channel.

(Also a bigger change though, that I won't be able to do this week.)

deuszx (approved):
I don't see any blockers for this PR specifically but I think it's uncovered few more problems that are being fixed here.

PR #3494, opened 2025-03-06

### LINKED ISSUE #3378: Put operation results in executed block
`execute_operation` (might get renamed: https://github.com/linera-io/linera-protocol/issues/3375) returns a result which is only used in cross-application calls, but dropped in operations.

Let's store the results as a separate `Vec<Vec<u8>>` (length `== operations.len()`) in the execution outcome.

usagi32:
## Motivation
resolves #3378 

## Test Plan

CI

## Release Plan

### DIFF
--- a/linera-chain/src/chain.rs
+++ b/linera-chain/src/chain.rs
@@ -172,7 +172,7 @@ static NUM_OUTBOXES: LazyLock<HistogramVec> = LazyLock::new(|| {
 });
 
 /// The BCS-serialized size of an empty [`Block`].
-const EMPTY_BLOCK_SIZE: usize = 91;
+const EMPTY_BLOCK_SIZE: usize = 92;
 
 /// An origin, cursor and timestamp of a unskippable bundle in our inbox.
 #[derive(Debug, Clone, Serialize, Deserialize, async_graphql::SimpleObject)]
@@ -776,6 +776,7 @@ where
         let mut oracle_responses = Vec::new();
         let mut events = Vec::new();
         let mut messages = Vec::new();
+        let mut operation_results = Vec::new();
         for (txn_index, transaction) in block.transactions() {
             let chain_execution_context = match transaction {
                 Transaction::ReceiveMessages(_) => ChainExecutionContext::IncomingBundle(txn_index),
@@ -889,6 +890,16 @@ where
             oracle_responses.push(txn_outcome.oracle_responses);
             messages.push(txn_messages);
             events.push(txn_outcome.events);
+
+            if let Transaction::ExecuteOperation(_) = transaction {
+                resource_controller
+                    .track_block_size_of(&(&txn_outcome.operation_result))
+                    .with_execution_context(chain_execution_context)?;
+                resource_controller
+                    .track_executed_block_size_sequence_extension(operation_results.len(), 1)
+                    .with_execution_context(chain_execution_context)?;
+                operation_results.push(txn_outcome.operation_result.into());
+            }
         }
 
         // Finally, charge for the block fee, except if the chain is closed. Closed chains should
@@ -918,6 +929,7 @@ where
             state_hash,
             oracle_responses,
             events,
+            operation_results,
         };
         Ok(outcome)
     }
--- a/linera-chain/src/data_types.rs
+++ b/linera-chain/src/data_types.rs
@@ -10,7 +10,7 @@ use std::{
 use async_graphql::SimpleObject;
 use custom_debug_derive::Debug;
 use linera_base::{
-    bcs,
+    bcs, bcs_scalar,
     crypto::{
         AccountPublicKey, AccountSecretKey, AccountSignature, BcsHashable, BcsSignable,
         CryptoError, CryptoHash, ValidatorPublicKey, ValidatorSecretKey, ValidatorSignature,
@@ -381,7 +381,7 @@ impl From<Vec<u8>> for OperationResult {
 
 impl<'de> BcsHashable<'de> for OperationResult {}
 
-doc_scalar!(
+bcs_scalar!(
     OperationResult,
     "The execution result of a single operation."
 );
--- a/linera-chain/src/data_types.rs
+++ b/linera-chain/src/data_types.rs
@@ -381,7 +381,10 @@ impl From<Vec<u8>> for OperationResult {
 
 impl<'de> BcsHashable<'de> for OperationResult {}
 
-doc_scalar!(OperationResult, "The execution result of a single operation.");
+doc_scalar!(
+    OperationResult,
+    "The execution result of a single operation."
+);
 
 /// A [`ProposedBlock`], together with the outcome from its execution.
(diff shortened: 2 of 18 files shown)


linera-chain/src/data_types.rs
```diff
@@ -364,6 +365,24 @@ impl OutgoingMessage {
...
+
+impl From<Vec<u8>> for OperationResult {
+    fn from(operation_outcome: Vec<u8>) -> Self {
+        OperationResult(operation_outcome)
+    }
+}
+
+impl<'de> BcsHashable<'de> for OperationResult {}
+
+doc_scalar!(OperationResult, "The outcome of a single operation.");
```
ma2bd, line 384: @afck Should this be `bcs_scalar`?

linera-execution/src/execution.rs
```diff
@@ -241,13 +241,17 @@ where
             self.handle_request(request).await?;
         }
 
-        let (controller, txn_tracker_moved) = contract_runtime_task.join().await?;
+        let (action_outcome, controller, txn_tracker_moved) = contract_runtime_task.join().await?;
```
ma2bd, line 244: replace `action_outcome` with `result`

ma2bd:
To update the formats, you may have to run something `INSTA_UPDATE=always cargo insta test -p linera-rpc`

usagi32:
> To update the formats, you may have to run something `INSTA_UPDATE=always cargo insta test -p linera-rpc`

They are already updated

linera-chain/src/chain.rs
```diff
@@ -889,6 +890,16 @@ where
             oracle_responses.push(txn_outcome.oracle_responses);
             messages.push(txn_messages);
             events.push(txn_outcome.events);
+
+            if let Transaction::ExecuteOperation(_) = transaction {
+                resource_controller
+                    .track_block_size_of(&(&txn_outcome.operation_result))
+                    .with_execution_context(chain_execution_context)?;
+                resource_controller
+                    .track_executed_block_size_sequence_extension(operation_results.len(), 1)
```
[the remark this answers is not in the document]
afck: Right, we currently have multiple sequences of the same length, so we could just add a multiplier to `track_executed_block_size_sequence_extension`. Probably something for a separate PR, though. Also, the _new_ sequence doesn't have the same length, because it's only as long as the operations, not all transactions.

linera-core/src/chain_worker/state/temporary_changes.rs
```diff
@@ -169,6 +169,10 @@ where
             .remove_bundles_from_inboxes(block.timestamp, &block.incoming_bundles)
             .await?;
         let outcome = if let Some(outcome) = outcome {
+            ensure!(
+                block.operations.len() == outcome.operation_results.len(),
+                WorkerError::InvalidBlockProposal(format!("{:?}", content))
+            );
```
afck, line 175: That's not necessary: We don't re-execute so we can't fully validate the outcome at this point anyway, and it is trusted because it is signed by a quorum of validators (this was checked before the `validate_proposal_content` call).

(If there were really a faulty quorum, or a severe bug: We _do_ compare the full outcome when handling the confirmed block in the end, so it would still get caught.)

linera-core/src/unit_tests/wasm_worker_tests.rs
```diff
@@ -296,6 +298,7 @@ where
             events: vec![Vec::new()],
             state_hash: creator_state.crypto_hash().await?,
             oracle_responses: vec![Vec::new()],
+            operation_results: vec![vec![0x0F, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00].into()],
```
afck, line 301: Is this the current counter value?

```suggestion
            operation_results: vec![bcs::to_bytes(15u64)?.into()],
```

linera-chain/src/data_types.rs
```diff
@@ -364,6 +365,27 @@ impl OutgoingMessage {
...
+
+impl From<Vec<u8>> for OperationResult {
+    fn from(operation_outcome: Vec<u8>) -> Self {
+        OperationResult(operation_outcome)
+    }
+}
+
+impl<'de> BcsHashable<'de> for OperationResult {}
+
+doc_scalar!(
```
afck, line 384: I agree with @ma2bd here:

```suggestion
bcs_scalar!(
```

afck (commented):
Looks great! :+1:

linera-core/src/chain_worker/state/temporary_changes.rs
usagi32: Yeah that was just me being over cautious.

linera-core/src/unit_tests/wasm_worker_tests.rs
usagi32: Sure as you wish. As its just a constant.

linera-core/src/chain_worker/state/temporary_changes.rs
```diff
@@ -169,6 +169,10 @@ where
             .remove_bundles_from_inboxes(block.timestamp, &block.incoming_bundles)
             .await?;
         let outcome = if let Some(outcome) = outcome {
+            ensure!(
```
usagi32, line 172:
```suggestion
```

linera-core/src/chain_worker/state/temporary_changes.rs
```diff
@@ -169,6 +169,10 @@ where
...
+                block.operations.len() == outcome.operation_results.len(),
```
usagi32, line 173:
```suggestion
```

linera-core/src/chain_worker/state/temporary_changes.rs
```diff
@@ -169,6 +169,10 @@ where
...
+                WorkerError::InvalidBlockProposal(format!("{:?}", content))
```
usagi32, line 174:
```suggestion
```

linera-core/src/chain_worker/state/temporary_changes.rs
usagi32, line 175:
```suggestion
```

afck (approved):
Looks good to me! :+1: 
(I'll let @ma2bd take another look, but from my side it's good to merge.)

usagi32:
> Looks good to me! 👍 (I'll let @ma2bd take another look, but from my side it's good to merge, once the remaining tests are fixed.)

Ok so `bcs_scalar!` was the one failing the CI (source of failure was the indexer test from storage service tests and the other test that failed was probably just a random failure)
I reverted it back to `doc_scalar`

linera-chain/src/chain.rs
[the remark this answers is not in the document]
ma2bd: @afck We should over-approximate this. IIRC it will never go over 5 bytes per vector.

linera-chain/src/data_types.rs
ma2bd: Why was this resolved?

ma2bd:
@usagi32 I landed your changes as part of #3535

PR #3496, opened 2025-03-07
Add the `cache_key_absence` to the LruCachingConfig.

MathieuDutSik:
## Motivation

The PR on epoch-events revealed that caching absence is creating some problems when another worker is writing data.
The error occurred because one worker had in its cache that a key was absent, then another worker wrote it but since in the cache it was indicating it as absent then it was absent.

## Proposal

* Instead of completely disabling the caching of absence, it is added as an option. We have to be realistic, caching is not going to get any simpler.
* The feature `cache_key_absence` is added to the `LruCachingConfig` and to the `CommonStoreConfig`. The default value is false.

The fact that this works is that the values being cached are Blobs which are immutable. 

## Test Plan

The test with `cache_key_absence = true` is added to the `LruMemoryStorage`. But for the other tests, the `cache_key_absence = false` is used.

## Release Plan

In relation to the work on `epoch-event`, fairly important.

## Links

None.

### DIFF
--- a/linera-views/src/backends/lru_caching.rs
+++ b/linera-views/src/backends/lru_caching.rs
@@ -87,8 +87,20 @@ impl<'a> LruPrefixCache {
 
     /// Marks cached keys that match the prefix as deleted. Importantly, this does not create new entries in the cache.
     pub fn delete_prefix(&mut self, key_prefix: &[u8]) {
-        for (_, value) in self.map.range_mut(get_interval(key_prefix.to_vec())) {
-            *value = None;
+        if self.cache_key_absence {
+            for (_, value) in self.map.range_mut(get_interval(key_prefix.to_vec())) {
+                *value = None;
+            }
+        } else {
+            let keys = self
+                .map
+                .range(get_interval(key_prefix.to_vec()))
+                .map(|(key, _)| key.to_vec())
+                .collect::<Vec<_>>();
+            for key in keys {
+                self.map.remove(&key);
+                self.queue.remove(&key);
+            }
         }
     }
 
@@ -134,11 +146,9 @@ where
         {
             let lru_read_values = lru_read_values.lock().unwrap();
             if let Some(value) = lru_read_values.query(key) {
-                if lru_read_values.cache_key_absence || value.is_some() {
-                    #[cfg(with_metrics)]
-                    NUM_CACHE_SUCCESS.with_label_values(&[]).inc();
-                    return Ok(value.clone());
-                }
+                #[cfg(with_metrics)]
+                NUM_CACHE_SUCCESS.with_label_values(&[]).inc();
+                return Ok(value.clone());
             }
         }
         #[cfg(with_metrics)]
@@ -201,25 +211,17 @@ where
         {
             let lru_read_values = lru_read_values.lock().unwrap();
             for (i, key) in keys.into_iter().enumerate() {
-                let cache_fault = if let Some(value) = lru_read_values.query(&key) {
-                    if lru_read_values.cache_key_absence || value.is_some() {
-                        #[cfg(with_metrics)]
-                        NUM_CACHE_SUCCESS.with_label_values(&[]).inc();
-                        result.push(value.clone());
-                        false
-                    } else {
-                        true
-                    }
+                if let Some(value) = lru_read_values.query(&key) {
+                    #[cfg(with_metrics)]
+                    NUM_CACHE_SUCCESS.with_label_values(&[]).inc();
+                    result.push(value.clone());
                 } else {
-                    true
-                };
-                if cache_fault {
                     #[cfg(with_metrics)]
                     NUM_CACHE_FAULT.with_label_values(&[]).inc();
                     result.push(None);
                     cache_miss_indices.push(i);
                     miss_keys.push(key);
-                }
+                };
             }
         }
         if !miss_keys.is_empty() {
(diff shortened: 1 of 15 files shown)


linera-views/src/backends/lru_caching.rs
```diff
@@ -346,38 +363,49 @@ where
...
         Self {
             store,
             lru_read_values,
         }
     }
 
-    /// Gets the `cache_size`
-    pub fn cache_size(&self) -> usize {
+    /// Gets the `cache_config`
+    pub fn cache_config(&self) -> (usize, bool) {
```
afck, line 400: (I'm fine with this for now, but in the long run, if we keep the option, we should probably make this a struct rather than a tuple.)

afck (approved):
Looks good to me, but I'm a bit surprised you apply the flag on the _reading_ side of things: If I'm reading the code correctly, regardless of the flag, you record absence in the LRU cache, but then you ignore `None` values when retrieving things from the cache? I would have expected we'd just not put `None` values into the cache in the first place.

linera-views/src/backends/lru_caching.rs
```diff
@@ -130,11 +132,13 @@ where
...
         // First inquiring in the read_value_bytes LRU
         {
-            let lru_read_values_container = lru_read_values.lock().unwrap();
-            if let Some(value) = lru_read_values_container.query(key) {
-                #[cfg(with_metrics)]
-                NUM_CACHE_SUCCESS.with_label_values(&[]).inc();
-                return Ok(value.clone());
+            let lru_read_values = lru_read_values.lock().unwrap();
+            if let Some(value) = lru_read_values.query(key) {
+                if lru_read_values.cache_key_absence || value.is_some() {
```
bart-linera, line 137: Wouldn't it be better to guard the `insert` call with the `cache_key_absence` check instead of the read?

linera-views/src/backends/lru_caching.rs
```diff
@@ -195,13 +199,21 @@ where
...
         {
-            let lru_read_values_container = lru_read_values.lock().unwrap();
+            let lru_read_values = lru_read_values.lock().unwrap();
             for (i, key) in keys.into_iter().enumerate() {
-                if let Some(value) = lru_read_values_container.query(&key) {
-                    #[cfg(with_metrics)]
-                    NUM_CACHE_SUCCESS.with_label_values(&[]).inc();
-                    result.push(value.clone());
+                let cache_fault = if let Some(value) = lru_read_values.query(&key) {
+                    if lru_read_values.cache_key_absence || value.is_some() {
```
bart-linera, line 205: Same here.

linera-views/src/backends/lru_caching.rs
MathieuDutSik: Yes, that works better.

linera-views/src/backends/lru_caching.rs
MathieuDutSik: Ok, next PRs.

linera-views/src/backends/lru_caching.rs
```diff
@@ -130,8 +144,8 @@ where
         };
         // First inquiring in the read_value_bytes LRU
         {
-            let lru_read_values_container = lru_read_values.lock().unwrap();
-            if let Some(value) = lru_read_values_container.query(key) {
+            let lru_read_values = lru_read_values.lock().unwrap();
+            if let Some(value) = lru_read_values.query(key) {
```
bart-linera, line 148: This can return `Some(None)` because of line 158 now, no? I thought we would be guarding that line. Something like:
```rust
if lru_read_values.cache_key_absence || value.is_some() {
    lru_read_values.insert(key.to_vec(), value.clone());
}
```

linera-views/src/backends/lru_caching.rs
```diff
@@ -195,9 +209,9 @@ where
         let mut cache_miss_indices = Vec::new();
         let mut miss_keys = Vec::new();
         {
-            let lru_read_values_container = lru_read_values.lock().unwrap();
+            let lru_read_values = lru_read_values.lock().unwrap();
             for (i, key) in keys.into_iter().enumerate() {
-                if let Some(value) = lru_read_values_container.query(&key) {
+                if let Some(value) = lru_read_values.query(&key) {
```
bart-linera, line 214: Same as above.

linera-views/src/store.rs
```diff
@@ -29,6 +29,8 @@ pub struct CommonStoreConfig {
     pub max_stream_queries: usize,
     /// The cache size being used.
     pub cache_size: usize,
+    /// Caching the absence of keys.
+    pub cache_key_absence: bool,
```
ma2bd, line 33: This belongs to the `LruCachingConfig`

ma2bd: Actually, maybe it should be in `LruCachingStore` actually.
It should be safe to set this flag to true every time a store is returned by `<LruCachingStore as AdminKeyValueStore>::connect` and `clone_with_root_key`.

ma2bd (changes requested):
I would rather land #3502 (which also applies to journaling) and maybe keep the work of this PR on testing.

Since this PR doesn't touch `linera-storage`, it just deactivates absence-caching across the board. In my view, users (here `linera-storage`) should use different *connection mode* instead of different configurations for blobs and views.

PR #3501, opened 2025-03-07
Do not pass a root key when calling the method `connect`

ma2bd:
## Motivation

* First step in distinguishing immutable shared data from root views
* Simplify code

## Proposal

Do not pass a root key when calling the method `connect`.

The current code will use the empty vector, but this could also change later to enforce a separation between the locked mode (with root key -- for root views) and the default mode (no root key -- for blobs).

## Test Plan

CI

### DIFF
--- a/linera-client/src/storage.rs
+++ b/linera-client/src/storage.rs
@@ -116,9 +116,6 @@ pub enum StorageConfig {
     },
 }
 
-/// The `root_key` used at startup before the `clone_with_root_key`.
-const ROOT_KEY: &[u8] = &[0];
-
 impl StorageConfig {
     #[cfg(feature = "rocksdb")]
     pub fn is_rocks_db(&self) -> bool {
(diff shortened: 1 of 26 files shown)


linera-client/src/storage.rs
```diff
@@ -116,9 +116,6 @@ pub enum StorageConfig {
     },
 }
 
-/// The `root_key` used at startup before the `clone_with_root_key`.
-const ROOT_KEY: &[u8] = &[0];
```
ma2bd, line 120: This value was arbitrary because we are also padding the key in `scylladb.rs`.

rust-toolchain.toml
```diff
@@ -1 +1 @@
-toolchains/stable/rust-toolchain.toml
\ No newline at end of file
+toolchains/nightly/rust-toolchain.toml
```
afck, line 1: Is this necessary?

MathieuDutSik (approved):
Quite a neat change. Thank you.

rust-toolchain.toml
ma2bd: missed that! I reverted the change on the main branch

ma2bd:
> Quite a neat change. Thank you.
> 
> EDIT: I checked that the `list_all_blob_ids` and `list_all_chain_ids` are still working.

Thanks for the extra testing

Issue #3453, opened 2025-03-03
Write committees as blobs to storage.


afck:
### Problem

The admin chain subscriptions for committee changes don't scale well: Every new chain sends a cross-chain message to the admin chain to subscribe to notifications about new committees, and whenever there is a new epoch (introducing a new committee), or an old epoch is removed, a cross-chain message is sent to every subscriber (i.e. almost every chain in the system).

Currently these cross-chain messages, as well as every chain state, contains a copy of the committees.

### Requirements

All chains still need to have a well-defined last block in each epoch, since at each block height, we need to know the validators are before agreeing on the next block.

They don't need to explicitly acknowledge old epochs expiring, however. (In fact, if an epoch is not trusted anymore, that is immediately the case for all chains; there cannot be an option to delay that.)

### Proposal

We are going to replace the old mechanism and get rid of the system channel bottleneck, so we:

* Remove the `CreateCommittee` and `RemoveCommittee` system message variants, and the corresponding system channel subscriptions.
* Remove the committees from all chain states.

Instead, we will use an event stream and blobs:

* Whenever a committee is created on the admin chain, it is stored in a new blob, and an event is emitted with:
  - a stream name `NEW_EPOCH_STREAM_NAME = &[0]`,
  - the new epoch number as the event ID, and
  - the blob ID as the payload.
* Whenever a committee is removed on the admin chain, a new event is emitted with:
  - a stream name `REMOVED_EPOCH_STREAM_NAME = &[1]`,
  - the old epoch number as the event ID, and
  - an empty payload.
* These events can be handled on other chains with a new `ProcessNewEpoch(Epoch)` or `ProcessRemovedEpoch(Epoch)` system operation. Execution of that operation:
  - reads the event from storage and creates an oracle response with its event ID (key and stream name) and content (the blob ID or empty), and
  - updates the chain's set of epochs and committees.  
* Every client will now follow/sync the admin chain:
  - When processing any chain's inbox on the CLI, the admin chain is synced as well, and if there are new epoch events, they are handled, too, in addition to the incoming messages.
  - When running the node service or faucet service, the client makes a gRPC subscription to the admin chain, so it is notified about new blocks and therefore new events. Whenever there is a new epoch event, it is handled on all its owned chains.

This still means that every currently running node service (and faucet) will subscribe to the admin chain, so it still puts some load on the validators, but it removes the on-chain subscriptions and all the cross-chain messages.

afck:
[...]

PR #3502, opened 2025-03-08
Fix blob storage

ma2bd:
## Motivation

* As noted by @MathieuDutSik in #3496, LRU-caching is incorrect outside views.
* However, journaling outside of views is also incorrect

## Proposal

On top of #3501, we simply deactivate the features that require exclusive access to an object in storage.
* Journaling is not authorized.
* LRU cache should never insert `None` but instead forget about the deleted key. (Not that we delete blobs but who knows in the future.)

We may want to rename `connect` and `clone_with_root_keys` later in another PR.

## Test Plan

* CI
* Verified that this solves the issue with #3453

## Release Plan

In principle, this chould be backported to the latest `testnet` branch.

### DIFF
--- a/linera-views/src/backends/journaling.rs
+++ b/linera-views/src/backends/journaling.rs
@@ -45,7 +45,7 @@ pub enum JournalConsistencyError {
     #[error("The journal block could not be retrieved, it could be missing or corrupted.")]
     FailureToRetrieveJournalBlock,
 
-    #[error("Refusing to use the journal with exclusive database access to the root object.")]
+    #[error("Refusing to use the journal without exclusive database access to the root object.")]
     JournalRequiresExclusiveAccess,
 }
 
--- a/linera-views/src/backends/journaling.rs
+++ b/linera-views/src/backends/journaling.rs
@@ -40,10 +40,13 @@ sa::const_assert!(JOURNAL_TAG < MIN_VIEW_TAG);
 
 /// Data type indicating that the database is not consistent
 #[derive(Error, Debug)]
+#[allow(missing_docs)]
 pub enum JournalConsistencyError {
-    /// The journal block could not be retrieved, it could be missing or corrupted
-    #[error("the journal block could not be retrieved, it could be missing or corrupted")]
+    #[error("The journal block could not be retrieved, it could be missing or corrupted.")]
     FailureToRetrieveJournalBlock,
+
+    #[error("Refusing to use the journal with exclusive database access to the root object.")]
+    JournalRequiresExclusiveAccess,
 }
 
 #[repr(u8)]
@@ -102,6 +105,8 @@ struct JournalHeader {
 pub struct JournalingKeyValueStore<K> {
     /// The inner store.
     store: K,
+    /// Whether we have exclusive R/W access to the keys under root key.
+    has_exclusive_access: bool,
 }
 
 impl<K> DeletePrefixExpander for &JournalingKeyValueStore<K>
@@ -184,12 +189,18 @@ where
 
     async fn connect(config: &Self::Config, namespace: &str) -> Result<Self, Self::Error> {
         let store = K::connect(config, namespace).await?;
-        Ok(Self { store })
+        Ok(Self {
+            store,
+            has_exclusive_access: false,
+        })
     }
 
     fn clone_with_root_key(&self, root_key: &[u8]) -> Result<Self, Self::Error> {
         let store = self.store.clone_with_root_key(root_key)?;
-        Ok(Self { store })
+        Ok(Self {
+            store,
+            has_exclusive_access: true,
+        })
     }
 
     async fn list_all(config: &Self::Config) -> Result<Vec<String>, Self::Error> {
@@ -233,6 +244,9 @@ where
         if Self::is_fastpath_feasible(&batch) {
             self.store.write_batch(batch).await
         } else {
+            if !self.has_exclusive_access {
+                return Err(JournalConsistencyError::JournalRequiresExclusiveAccess.into());
+            }
             let header = self.write_journal(batch).await?;
             self.coherently_resolve_journal(header).await
         }
@@ -406,6 +420,9 @@ where
 impl<K> JournalingKeyValueStore<K> {
     /// Creates a new journaling store.
     pub fn new(store: K) -> Self {
-        Self { store }
+        Self {
+            store,
+            has_exclusive_access: false,
+        }
     }
 }
--- a/linera-views/src/backends/lru_caching.rs
+++ b/linera-views/src/backends/lru_caching.rs
@@ -397,12 +397,9 @@ impl<K> LruCachingStore<K> {
 
     /// Sets the value `has_exclusive_access` to `true`, if applicable.
     pub fn enable_exclusive_access(&self) {
-        match &self.lru_read_values {
-            None => (),
-            Some(lru_read_values) => {
-                let mut lru_read_values = lru_read_values.lock().unwrap();
-                lru_read_values.has_exclusive_access = true;
-            }
+        if let Some(lru_read_values) = &self.lru_read_values {
+            let mut lru_read_values = lru_read_values.lock().unwrap();
+            lru_read_values.has_exclusive_access = true;
         }
     }
 }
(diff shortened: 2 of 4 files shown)


linera-views/src/backends/journaling.rs
```diff
@@ -40,10 +40,13 @@ sa::const_assert!(JOURNAL_TAG < MIN_VIEW_TAG);
...
 /// Data type indicating that the database is not consistent
 #[derive(Error, Debug)]
+#[allow(missing_docs)]
 pub enum JournalConsistencyError {
-    /// The journal block could not be retrieved, it could be missing or corrupted
-    #[error("the journal block could not be retrieved, it could be missing or corrupted")]
+    #[error("The journal block could not be retrieved, it could be missing or corrupted.")]
     FailureToRetrieveJournalBlock,
+
+    #[error("Refusing to use the journal with exclusive database access to the root object.")]
```
afck, line 48:
```suggestion
    #[error("Refusing to use the journal without exclusive database access to the root object.")]
```

rust-toolchain.toml
```diff
@@ -1 +1 @@
-toolchains/stable/rust-toolchain.toml
\ No newline at end of file
+toolchains/nightly/rust-toolchain.toml
```
afck, line 1: Is that necessary?

linera-views/src/backends/lru_caching.rs
```diff
@@ -36,19 +36,16 @@ static NUM_CACHE_FAULT: LazyLock<IntCounterVec> =
...
-/// It is inspired by the crate `lru-cache`.
+/// Stores the data for simple `read_values` queries.
 ///
-/// We cannot apply this crate directly because the batch operation
-/// need to update the cache. In the case of `DeletePrefix` we have to
-/// handle the keys by prefixes. And so we need to have a `BTreeMap` to
-/// keep track of this.
-
-/// The data structures
+/// This data-structure is inspired by the crate `lru-cache` but was modified to support
```
afck, line 41:
```suggestion
/// This data structure is inspired by the crate `lru-cache` but was modified to support
```

afck (commented):
Looks good to me!

The assumption that after `clone_with_root_key` we have exclusive access is not something that storage guarantees, is it? It's just that in our use case, all cloned stores are for root views, and those are always only accessed by a single worker?

linera-views/src/backends/lru_caching.rs
```diff
@@ -36,19 +36,16 @@ static NUM_CACHE_FAULT: LazyLock<IntCounterVec> =
...
+/// range deletions.
 struct LruPrefixCache {
     map: BTreeMap<Vec<u8>, Option<Vec<u8>>>,
     queue: LinkedHashMap<Vec<u8>, (), RandomState>,
     max_cache_size: usize,
+    /// Whether we have exclusive R/W access to the keys under the root key of the store.
+    has_exclusive_access: bool,
```
MathieuDutSik, line 48: The use of the terminology `has_exclusive_access` is adequate for the terminology.
But the use in the case `!use_exclusive_access` makes sense only in a system where data is written in keys are only written, never modified, never erased. That needs to be said. But yes, it is better than `cache_key_absence` or PR #3496.

The problem of this design is that it forces an API. Conceivably, we could do a `clone_with_root_key` and not have exclusive access to it. Also, we could access to a store and have access to it. Therefore, I think it would make sense to have that in the input parameter to the `LruCachingConfig`.

linera-views/src/backends/journaling.rs
```diff
@@ -237,6 +244,9 @@ where
         if Self::is_fastpath_feasible(&batch) {
             self.store.write_batch(batch).await
         } else {
+            if !self.has_exclusive_access {
+                return Err(JournalConsistencyError::JournalRequiresExclusiveAccess.into());
+            }
```
MathieuDutSik, line 249: The action is indeed necessary for the system to work correctly.
However, I would say that this kills the interest of journaling.

The problem is that what is stored as blobs is large data like Wasm compiled code. Given the limits on size for the fastpath this could be a problem. Still, I have not been able to enable the problem.

Right now, I do not see how to address this issue and have journal and non-exclusive access.

rust-toolchain.toml
ma2bd: oh noo

ma2bd: fixed on the main branch

linera-views/src/backends/lru_caching.rs
ma2bd: We need to rename `clone_with_root_key`. In fact the types returned by `connect` and `clone_with_root_key` should probably be different.

linera-views/tests/store_tests.rs
```diff
@@ -248,9 +271,12 @@ async fn test_indexed_db_big_write_read() {
 #[cfg(with_dynamodb)]
 #[tokio::test]
 async fn test_dynamo_db_big_write_read() {
+    use linera_views::store::AdminKeyValueStore as _;
+
     let store = linera_views::dynamo_db::DynamoDbStore::new_test_store()
         .await
         .unwrap();
+    let store = store.clone_with_root_key(&[]).unwrap();
```
MathieuDutSik, line 279: I do not think that is needed as the `run_big_write_read` is very much not cached.

linera-views/tests/store_tests.rs
```diff
@@ -75,17 +75,34 @@ async fn test_reads_rocks_db() {
...
+            .await
+            .unwrap();
+        let store = store.clone_with_root_key(&[]).unwrap();
+        run_reads(store, scenario).await;
+    }
+}
+
+#[cfg(with_scylladb)]
+#[tokio::test]
+async fn test_reads_scylla_db_no_journaling() {
```
MathieuDutSik, line 105: Running locally those tests, it is impossible to see some differences in the timing. And running both tests takes the same runtime as one single test.

Also, I object to the `no_journaling` terminology:
* The journaling is still present in the journaling. If we are in exclusive access then operation fail.
* The difference is potentially about caching not journaling. And there is still caching, just caching of values, not absence of values.
* But actually there is no difference. The difference between exclusive access and no exclusive access is about the checking of absence, and that caching of absence is not going to be present.

linera-views/src/backends/lru_caching.rs
```diff
@@ -377,6 +394,17 @@ impl<K> LruCachingStore<K> {
...
+    /// Sets the value `has_exclusive_access` to `true`, if applicable.
+    pub fn enable_exclusive_access(&self) {
+        match &self.lru_read_values {
+            None => (),
+            Some(lru_read_values) => {
+                let mut lru_read_values = lru_read_values.lock().unwrap();
+                lru_read_values.has_exclusive_access = true;
+            }
+        }
+    }
```
MathieuDutSik, line 407: This can be shortened as
```[rust]
if let Some(lru_read_values) = &self.lru_read_values {
   let mut lru_read_values = lru_read_values.lock().unwrap();
   lru_read_values.has_exclusive_access = true;
}
```

This could also be inlined in the `clone_with_root_key`.

MathieuDutSik (approved):
Thanks for addressing the issue.

linera-views/tests/store_tests.rs
ma2bd: This seems to be required:
https://github.com/linera-io/linera-protocol/actions/runs/13734903398/job/38417381624

ma2bd: It's more about unsafe journaling actually

linera-views/tests/store_tests.rs
ma2bd: ok fair. let me rename to `test_reads_scylla_db_no_root_key`

linera-views/src/backends/lru_caching.rs
ma2bd: thanks

ma2bd:
> Looks good to me!
> 
> The assumption that after `clone_with_root_key` we have exclusive access is not something that storage guarantees, is it? It's just that in our use case, all cloned stores are for root views, and those are always only accessed by a single worker?

Yes that's correct

PR #3505, opened 2025-03-08
Support custom `ResourceControlPolicy` in integration tests

jvff:
## Motivation

In integration tests, the `ResourceControlPolicy` was hardcoded to use the default value. With the addition of an HTTP allow list (#3427), the tests aren't able to test any HTTP requests.

## Proposal

Allow changing the `ResourceControlPolicy` used inside a test, allowing to add hosts to the allow list. For the policy change to work, a new committee and epoch must be published. Therefore the epoch and committee can't be assumed to be constant anymore.

## Test Plan

Wrote some (upcoming) tests that try to perform an HTTP request without changing the policy and after adding the host to the policy, and ensured that the former fails while the latter succeeds.

## Release Plan

- These changes follow the usual release cycle, because this will help with tests that use a feature that has not yet been released.

## Links

### DIFF
--- a/linera-sdk/src/test/validator.rs
+++ b/linera-sdk/src/test/validator.rs
@@ -22,8 +22,10 @@ use linera_base::{
 use linera_core::worker::WorkerState;
 use linera_execution::{
     committee::{Committee, Epoch},
-    system::{OpenChainConfig, SystemOperation, OPEN_CHAIN_MESSAGE_INDEX},
-    WasmRuntime,
+    system::{
+        AdminOperation, OpenChainConfig, SystemChannel, SystemOperation, OPEN_CHAIN_MESSAGE_INDEX,
+    },
+    ResourceControlPolicy, WasmRuntime,
 };
 use linera_storage::{DbStorage, Storage, TestClock};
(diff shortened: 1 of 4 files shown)


linera-sdk/src/test/validator.rs
```diff
@@ -45,7 +50,7 @@ use crate::ContractAbi;
 pub struct TestValidator {
     validator_secret: ValidatorSecretKey,
     account_secret: AccountSecretKey,
-    committee: Committee,
+    committee: Arc<Mutex<(Epoch, Committee)>>,
```
afck, line 53: Should a validator have a `committee` field at all? In production, it's kind of per chain.

jvff: This is just a "cached" view of the latest epoch and committee the test is using. Right now all chains should switch epochs in lock-step. This is a limitation and I agree that it should be removed later, but it requires some broader changes.

afck (approved):

linera-sdk/src/test/validator.rs
jvff: Added #3507 to track this, thanks!

PR #3508, opened 2025-03-08
Tracing improvements

### LINKED ISSUE #3499: `ChainWorkerRequests` execute in the wrong tracing context
They execute in the context of the call that spawned the actor task, not the call that made the request.  We should attach a tracing context to each request and enter it for execution.

Twey:
## Motivation
https://github.com/linera-io/linera-protocol/issues/3499

## Proposal
Pass a span with each request.

## Test Plan
CI.

## Release Plan

## Links
- Fixes #3499.

### DIFF
--- a/linera-core/src/chain_worker/actor.rs
+++ b/linera-core/src/chain_worker/actor.rs
@@ -282,7 +282,7 @@ where
 
     /// Runs the worker until there are no more incoming requests.
     #[instrument(
-        name = "ChainWorkerActor",
+        name = "ChainWorkerActor::handle_requests",
         skip_all,
         fields(chain_id = format!("{:.8}", self.worker.chain_id())),
     )]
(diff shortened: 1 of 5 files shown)


linera-core/src/worker.rs
```diff
@@ -811,7 +815,7 @@ where
     }
 
     /// Processes a confirmed block certificate.
-    #[instrument(skip_all, fields(
+    #[instrument(skip_all,  fields(
```
ma2bd, line 818: typo

linera-core/src/chain_worker/state/mod.rs
```diff
@@ -204,6 +204,7 @@ where
         &mut self,
         proposal: BlockProposal,
     ) -> Result<(ChainInfoResponse, NetworkActions), WorkerError> {
+        tracing::debug!("handling block proposal: {proposal:#?}");
```
ma2bd, line 207: leftover? We could certainly add TRACE-level log for all handlers.

ma2bd (commented):
LGTM but maybe @afck and @jvff wnat to weigh in

linera-core/src/chain_worker/actor.rs
```diff
@@ -285,18 +288,21 @@ where
...
-        while let Some(request) = incoming_requests.recv().await {
+        while let Some((request, span)) = incoming_requests.recv().await {
             // TODO(#2237): Spawn concurrent tasks for read-only operations
             trace!("Handling `ChainWorkerRequest`: {request:?}");
 
             let responded = match request {
                 #[cfg(with_testing)]
                 ChainWorkerRequest::ReadCertificate { height, callback } => callback
-                    .send(self.worker.read_certificate(height).await)
+                    .send(self.worker.read_certificate(height).instrument(span).await)
```
afck, line 305: Should we deduplicate all the `instrument(span)`s (and also prevent that we forget adding it to calls in the future) by calling `handle_requests(…).instrument(span)` from `handle_requests_with_span`?

linera-core/src/chain_worker/state/mod.rs
Twey: I think this should be done with `tracing::instrument` though.

linera-core/src/worker.rs
Twey: Oh I was hoping `cargo fmt` would fix this one for me :sweat_smile:

linera-core/src/chain_worker/actor.rs
Twey: I was on the fence about doing this anyway :sweat_smile:

afck (approved):

PR #3509, opened 2025-03-08
Add a "How-To" example for performing HTTP requests

jvff:
## Motivation

Performing HTTP requests from applications is an important feature for Linera. There are a few different ways for applications to perform them, and there are tradeoffs between these ways. Therefore, this feature should be documented and a "How-To" example demonstrating how to use and test the feature would be useful.

## Proposal

Add an example that performs HTTP requests in all different ways, and test them.

## Test Plan

Unit and integration tests were added to test not just the application, but also to serve as testing examples and to also test the HTTP allow-list.

## Release Plan

- Nothing to do, because this is just a new example being added to the repository.

## Links

### DIFF
--- a/examples/how-to/perform-http-requests/README.md
+++ b/examples/how-to/perform-http-requests/README.md
@@ -1,0 +1,140 @@
+# How to perform HTTP requests
+
+This example application demonstrates how to perform HTTP requests from the service and from the
+contract, in a few different ways:
+
+- From the service while handling a mutation.
+- From the contract directly.
+- From the service when it is being used as an oracle by the contract.
+
+## HTTP requests from the service
+
+The service is executed either on the client when requested by the user or on validators when the
+service is queried as an oracle by the contract. In this first usage scenario, the HTTP request is
+executed only in the client.
+
+The HTTP response can then be used by the service to either prepare a query response to the caller
+or to prepare operations to be executed by the contract in a block proposal.
+
+## HTTP requests from the contract
+
+The contract can perform HTTP requests as well, but the responses must always be the same. The
+requests are executed on the client and on all the validators. That means that the client and each
+validator perform the HTTP request independently. The responses must all match (or at least match
+in a quorum of validators) for the block the be confirmed.
+
+If the response varies per request (as a simple example, due to the presence of a "Date" timestamp
+header in the response), the block proposal may end up being rejected by the validators. If there's
+a risk of that happening, the contract should instead call the service as an oracle, and let the
+service perform the HTTP request and return only the deterministic parts of the response.
+
+## HTTP requests using the service as an oracle
+
+The contract may call the service as an oracle. That means that that contracts sends a query to the
+service and waits for its response. The execution of the contract is metered by executed
+instruction, while the service executing as an oracle is metered by a coarse-grained timer. That
+allows the service to execute non-deterministically, and as long as it always returns a
+deterministic response back to the contract, the validators will agree on its execution and reach
+consensus.
+
+In this scenario, the contract requests the service to perform the HTTP request. The HTTP request
+is also executed in each validator.
+
+## Recommendation
+
+It is recommended to minimize the number of HTTP requests performed in total, in order to reduce
+costs. Whenever possible, it's best to perform the request in the client using the service, and
+forward only the HTTP response to the contract. The contract should then verify that the response
+can be trusted.
+
+If there's no way to verify an off-chain HTTP response in the contract, then the request should be
+made in the contract. However, if there's a risk of receiving different HTTP responses among the
+validators, the contract should use the service as oracle to perform the HTTP request and return to
+the contract only the data that is deterministic. Using the service as an oracle is more expensive,
+so it should be avoided if possible.
+
+## Usage
+
+### Setting Up
+
+Before getting started, make sure that the binary tools `linera*` corresponding to
+your version of `linera-sdk` are in your PATH.
+
+For the test, a simple HTTP server will be executed in the background.
+
+```bash
+HTTP_PORT=9090
+cd examples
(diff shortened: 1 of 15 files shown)


linera-base/src/crypto/secp256k1.rs
```diff
@@ -63,7 +63,7 @@ pub struct Secp256k1Signature(pub Signature);
 
 impl Secp256k1PublicKey {
     /// A fake public key used for testing.
-    #[cfg(with_testing)]
+    #[cfg(all(with_testing, not(target_arch = "wasm32")))]
```
deuszx, line 66: Did you encounter problems with compilation here?

examples/how-to/perform-http-requests/tests/http_requests.rs
```diff
@@ -0,0 +1,248 @@
...
+        });
+
+    assert_eq!(bytes, HTTP_RESPONSE_BODY.as_bytes());
+
+    Ok(())
+}
+
+/// Tests if service query can't perform HTTP requests to hosts that aren't allowed.
+#[test_log::test(tokio::test)]
+#[should_panic(expected = "UnauthorizedHttpRequest")]
```
deuszx, line 65: Does it expect a "Rust panic" in the service part? How will that behave on the validator node (client)? Will it propagate the panic and shut down the node?

examples/how-to/perform-http-requests/tests/http_requests.rs
```diff
@@ -0,0 +1,248 @@
...
+                .http_request_allow_list
+                .insert("localhost".to_owned());
+        })
+        .await;
+
+    chain
+        .graphql_mutation(application_id, "mutation { performHttpRequest }")
+        .await;
+
+    Ok(())
```
deuszx, line 102: It looks like it's strictly less test than `service_query_performs_http_request`?

examples/how-to/perform-http-requests/tests/http_requests.rs
```diff
@@ -0,0 +1,248 @@
...
+    validator
+        .change_resource_control_policy(|policy| {
+            policy
+                .http_request_allow_list
+                .insert("localhost".to_owned());
+        })
+        .await;
+
+    chain
+        .graphql_mutation(application_id, "mutation { performHttpRequest }")
```
deuszx, line 130: It'd be nice to be able to really see that the block proposal failed for _exactly_ the expected reason here.

deuszx (commented):
A more general question: 

Why do we need all these ways to perform an HTTP query? If only the service was allowed to make queries then it could be used both on~ and off~ chain:
- A client node calls the service only and consumes the results to construct a block
- A contract calls into its service and consumes the result (acting as an oracle)

examples/how-to/perform-http-requests/README.md
```diff
@@ -0,0 +1,125 @@
...
+
+```bash
+export PATH="$PWD/target/debug:$PATH"
+source /dev/stdin <<<"$(linera net helper 2>/dev/null)"
+```
+
+To start the local Linera network:
+
+```bash
+linera_spawn_and_read_wallet_variables linera net up --testing-prng-seed 37
```
ma2bd, line 74: Can you use the new blurb and bash setup? (see main README file)

ma2bd (changes requested):
see comment

Also the example is not run in CI, is it?

linera-base/src/crypto/secp256k1.rs
jvff: Yes, I believe when it tried to compile the tests for `wasm32-unknown-unknown`.

examples/how-to/perform-http-requests/README.md
jvff: Done, thanks!

jvff:
> Also the example is not run in CI, is it?

It should be: https://github.com/linera-io/linera-protocol/actions/runs/13744827986/job/38438473644#step:5:968 :thinking:

examples/how-to/perform-http-requests/README.md
```diff
@@ -0,0 +1,128 @@
...
+
+```bash
+linera_spawn linera net up --with-faucet --faucet-port 8080
+
+# Remember the URL of the faucet.
+FAUCET_URL=http://localhost:8080
+```
+
+We use the test-only CLI option `--testing-prng-seed` to make keys deterministic and simplify our
+explanation.
```
ma2bd, line 81: remove this

examples/how-to/perform-http-requests/README.md
```diff
@@ -0,0 +1,128 @@
...
+
+```bash
+INFO=($(linera wallet request-chain --faucet $FAUCET_URL))
+CHAIN="${INFO1[0]}"
```
ma2bd, line 85: typo

ma2bd: I suspect this file is not tested in CI. You may need to update `readme_test.rs`

ma2bd (approved):

examples/how-to/perform-http-requests/README.md
jvff: Thanks! Now that it's running as a test, I realized there's an issue. We need an HTTP server for the test to work :sweat_smile: And we also need to allow that HTTP server to be accessed from the localnet. I'll look into this tomorrow.

ma2bd: I suppose the faucet is also an http server :)

jvff: It needed a specific response for the test to pass, so I created a simple HTTP server for the test.

examples/how-to/perform-http-requests/README.md
jvff: Done.

PR #3510, opened 2025-03-08
[nit] some renaming in LRU cache

ma2bd:
## Motivation

According to [style guidelines](https://github.com/linera-io/linera-protocol/blob/main/CONTRIBUTING.md#naming-conventions), the proper variable names for values of `Option<Arc<Mutex<LruPrefixCache>>>` is
`lru_prefix_cache` or `cache`.

## Proposal

Rename fields and variables accordingly.

## Test Plan

CI

### DIFF
--- a/linera-views/src/backends/lru_caching.rs
+++ b/linera-views/src/backends/lru_caching.rs
@@ -87,7 +87,8 @@ impl<'a> LruPrefixCache {
         }
     }
 
-    /// Marks cached keys that match the prefix as deleted. Importantly, this does not create new entries in the cache.
+    /// Marks cached keys that match the prefix as deleted. Importantly, this does not
+    /// create new entries in the cache.
     pub fn delete_prefix(&mut self, key_prefix: &[u8]) {
         if self.has_exclusive_access {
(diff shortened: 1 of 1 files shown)


linera-views/src/backends/lru_caching.rs
```diff
@@ -204,17 +205,17 @@ where
         &self,
         keys: Vec<Vec<u8>>,
     ) -> Result<Vec<Option<Vec<u8>>>, Self::Error> {
-        let Some(lru_read_values) = &self.lru_read_values else {
+        let Some(cache) = &self.cache else {
```
MathieuDutSik, line 208: Why sometimes we have `let Some(cache) = ...` and sometime we have `let Some(values) = ...`?

MathieuDutSik (approved):
I approve as it is a movement in the right direction, but there is still some inconsistencies.

linera-views/src/backends/lru_caching.rs
ma2bd: good catch

ma2bd:
> I approve as it is a movement in the right direction, but there is still some inconsistencies.

fixed

PR #3511, opened 2025-03-08
Add the test for the behavior of the cache in the shared scenario.

MathieuDutSik:
## Motivation

Following the merging of PR #3502 the test was missing.

## Proposal

Add the test from PR #3496.

## Test Plan

The test is the whole point.

## Release Plan

Not relevant.

## Links

None

### DIFF
--- a/linera-views/src/test_utils/mod.rs
+++ b/linera-views/src/test_utils/mod.rs
@@ -774,15 +774,16 @@ where
     assert_eq!(keys, read_keys);
 }
 
-/// Due to the fact that some store are caching values, it is possible
-/// that the following scenario occurs:
-/// * Store 1 deletes a key and mark it as missing in its cache.
-/// * Store 2 writes the key
-/// * Store 1 reads the key, see it as missing and returns false
-///   erroneously.
+/// A store can be in exclusive access where it stores the absence of values
+/// or in shared access where only values are stored and (key, value) once
+/// written are never modified nor erased.
 ///
-/// This test checks that the behavior is correct.
-pub async fn test_cache_check_absence_admin_test<S: TestKeyValueStore>()
+/// In case of no exclusive access the following scenarion is checked
+/// * Store 1 deletes a key and does not mark it as missing in its cache.
+/// * Store 2 writes the key
+/// * Store 1 reads the key, but since it is no in the cache it can read
+///   it correctly.
+pub async fn no_exclusive_access_admin_test<S: TestKeyValueStore>()
 where
     S::Error: Debug,
 {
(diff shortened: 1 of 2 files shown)


linera-views/src/test_utils/mod.rs
```diff
@@ -773,3 +773,71 @@ where
...
+///   it correctly.
+pub async fn no_exclusive_access_admin_test<S: TestKeyValueStore>()
+where
+    S::Error: Debug,
+{
+    let config = S::new_test_config().await.expect("config");
+    let namespace = generate_test_namespace();
+    S::create(&config, &namespace).await.expect("creation");
+    let key = vec![42];
+    //
```
ma2bd, line 794: nit: We use simple empty lines in the rest of the code.

ma2bd (approved):

linera-views/src/test_utils/mod.rs
```diff
@@ -773,3 +773,71 @@ where
...
 }
+
+/// A store can be in exclusive access where it stores the absence of values
+/// or in shared access where only values are stored and (key, value) once
+/// written are never modified nor erased.
+///
+/// In case of no exclusive access the following scenarion is checked
+/// * Store 1 deletes a key and does not mark it as missing in its cache.
+/// * Store 2 writes the key
+/// * Store 1 reads the key, but since it is no in the cache it can read
```
afck, line 784:
```suggestion
/// * Store 1 reads the key, but since it is not in the cache it can read
```

linera-views/src/test_utils/mod.rs
```diff
@@ -773,3 +773,71 @@ where
...
+    store2.write_batch(batch2).await.expect("write batch2");
+    //
+    assert!(store1.contains_key(&key).await.unwrap());
+}
+
+/// In case of exclusive access. We have the following scenario:
+/// * Store 1 deletes a key and mark it as missing in its cache.
+/// * Store 2 writes the key (it should not be doing it)
+/// * Store 1 reads the key, see it as missing.
+pub async fn exclusive_access_admin_test<S: TestKeyValueStore>()
```
afck, line 812: You could deduplicate these tests, and use a `exclusive_access: bool` argument. They only differ in the setup and the final assertion, I think?

afck (approved):

linera-views/src/test_utils/mod.rs
MathieuDutSik: Ok, done.

linera-views/src/test_utils/mod.rs
MathieuDutSik: Ok, thanks.

linera-views/src/test_utils/mod.rs
MathieuDutSik: Yes, it was possible to do that.

PR #3513, opened 2025-03-10
Fix message ordering

### LINKED ISSUE #3467: Outgoing messages are in the wrong order.
We put outgoing messages into `RawExecutionOutcome`s and 

The expected outcome of the following would be messages A0, B0, S, B1, A1, in that order:

* App A sends A0. Then calls into app B.
* B sends B0.
* B makes e.g. a native token transfer and sends S.
* B sends B1, then returns to A.
* A sends A1.

But currently A0 and A1 would be in one `RawExecutionOutcome`, B0 and B1 in another, and S in yet another one, I believe, and the order will be S, B0, B1, A0, A1.

We can probably remove `RawExecutionOutcome` entirely, and add messages directly to the transaction tracker, in the correct order.

ma2bd:
## Motivation

fixes #3467

## Proposal

Use the transaction tracker for messages as well

## Test Plan

CI

### DIFF
--- a/linera-execution/src/runtime.rs
+++ b/linera-execution/src/runtime.rs
@@ -12,7 +12,7 @@ use custom_debug_derive::Debug;
 use linera_base::{
     crypto::CryptoHash,
     data_types::{
-        Amount, ApplicationPermissions, ArithmeticError, BlockHeight, OracleResponse, Resources,
+        Amount, ApplicationPermissions, ArithmeticError, BlockHeight, OracleResponse,
         SendMessageRequest, Timestamp,
     },
     ensure, http,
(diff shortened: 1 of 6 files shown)


linera-execution/tests/test_execution.rs
```diff
@@ -1372,33 +1249,22 @@ async fn test_multiple_messages_from_different_applications() -> anyhow::Result<
         txn_outcome.outcomes,
         &[
             ExecutionOutcome::User(
-                silent_target_id,
-                RawExecutionOutcome::default().with_refund_grant_to(Some(account)),
-            ),
-            ExecutionOutcome::User(
-                sending_target_id,
+                caller_id,
```
ma2bd, line 1252: The reordering is visible here.

linera-execution/src/runtime.rs
```diff
@@ -1232,16 +1180,28 @@ impl ContractRuntime for ContractSyncRuntimeHandle {
...
+            RawOutgoingMessage::from(message).into_priced(&this.resource_controller.policy)?;
+        this.resource_controller.track_grant(message.grant)?;
+
+        let outcome = RawExecutionOutcome {
+            authenticated_signer,
+            refund_grant_to: this.refund_grant_to,
+            messages: vec![message],
+        };
+        this.transaction_tracker
+            .add_user_outcome(application_id, outcome)?;
```
afck, line 1197: (Not in this PR, but: This could just become `add_message`, and take an `OutgoingMessage`, which whould have to be moved to `linera-execution`, could it? We don't need `RawExecutionOutcome` at all, and after this PR it will always contain only a single message anyway.)

afck (approved):

PR #3518, opened 2025-03-10
Fix message ordering

### LINKED ISSUE #3467: Outgoing messages are in the wrong order.

usagi32:
## Motivation
resolves #3467

## Test Plan

e2e

## Release Plan

### DIFF
--- a/linera-execution/src/runtime.rs
+++ b/linera-execution/src/runtime.rs
@@ -119,8 +119,6 @@ struct ApplicationStatus {
     description: UserApplicationDescription,
     /// The authenticated signer for the execution thread, if any.
     signer: Option<Owner>,
-    /// The current execution outcome of the application.
-    outcome: RawExecutionOutcome<Vec<u8>, Resources>,
 }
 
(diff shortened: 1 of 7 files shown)


ma2bd:
@usagi32 Nice work. Sorry there was a little bit of a race condition as we're preparing our next testnet.

ma2bd (changes requested):
Feel free to close this PR, or rebase to propose some improvements in our tests

PR #3535, opened 2025-03-11

### LINKED ISSUE #3378: Put operation results in executed block

ma2bd:
## Motivation

resolves #3378

## Proposal

This is a rebase of #3494 authored by @usagi32 

## Test Plan

CI

### DIFF
--- a/linera-rpc/src/grpc/server.rs
+++ b/linera-rpc/src/grpc/server.rs
@@ -518,11 +518,12 @@ where
         } = request.into_inner().try_into()?;
         trace!(?certificate, "Handling lite certificate");
         let (sender, receiver) = wait_for_outgoing_messages.then(oneshot::channel).unzip();
-        match self
-            .state
-            .clone()
-            .handle_lite_certificate(certificate, sender)
-            .await
+        match Box::pin(
+            self.state
+                .clone()
+                .handle_lite_certificate(certificate, sender),
+        )
+        .await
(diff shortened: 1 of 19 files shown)


linera-rpc/src/grpc/server.rs
```diff
@@ -518,11 +518,12 @@ where
         } = request.into_inner().try_into()?;
         trace!(?certificate, "Handling lite certificate");
         let (sender, receiver) = wait_for_outgoing_messages.then(oneshot::channel).unzip();
-        match self
-            .state
-            .clone()
-            .handle_lite_certificate(certificate, sender)
-            .await
+        match Box::pin(
```
deuszx, line 521: We didn't need `Box::pin` before, what changed?

usagi32: Hey @deuszx, it was a lint error for me. Not sure why it didn't occur for the else. To be pedantic, it was a clippy future too large, the lint level is defined at the respective `lib.rs`

ma2bd: yes

Issue #3533, opened 2025-03-10
Use metering to limit service oracle execution time


jvff:
# Motivation

Service as oracles are not metered per instruction, but by execution time. However, services should not be allowed to run indefinitely before determining if the block exceeded its service oracle execution time limit.

This was previously attempted by running the service oracle in a separate thread, and using a timeout to receive the response from that thread. That [did not work](https://github.com/linera-io/linera-protocol/issues/2127#issuecomment-2710840953) because there was no safe way to cancel the thread when the timeout happens.

Instead, the decision was made to use the metering to periodically check if the service had reach its execution time limit.

# Proposal

Enable metering for all services, but when the method attached for consuming fuel is called (#3530), check if the service ran out of execution time instead.

MathieuDutSik:
If service calls are free for low runtime, like 1 minute, then there is no barrier for a contract to offload their computations to services for free.

Therefore, services need to be paid for by contract users. There are several ways to do that:
* Put a price associated with the runtime. That is a maximum runtime of service calls. One needs to put a price on the time. 
* Put a price identical to the contract calls. Of course, the problem is that the runtime could be different on different validators. Therefore, the caller of services should put a maximum cost on the service call. If it goes above the limit, then the block construction has failed.

Putting a price on the time introduces another variable, while what is measured is the same: The use of computer resources. So, using the time is not something that should be preferred.

Service calls are also used for outside calls. This is a similar use case to `eth_call` in the Ethereum world. The way the validator handles that situation is by putting a maximum fuel usage. We could use the same strategy here.

The problem is that this introduces a complication to the user: When calling a service, for example, by `query_service` the contract must put a maximum fuel cost on his call.

PR #3536, opened 2025-03-11
Service as oracle execution time limit configuration

### LINKED ISSUE #3528: Add a configuration option for the service timeout
# Motivation

Running services as oracles should not exceed a maximum execution time agreed upon between the validators. Unfortunately, this requires some larger changes because the service runs in a synchronous thread that's hard to cancel.

The solution discussed was to use Wasm metering to periodically check if the time budget was exceeded.

# Proposal

Add a field to `ResourceControlPolicy` with the `maximum_service_oracle_execution_time` per block.

jvff:
## Motivation

Services should not execute more than a maximum time limit. The limit should be agreed upon among the validators.

## Proposal

Add a `maximum_service_oracle_execution_ms` field to `ResourceControlPolicy`. Update the `ResourceTracker` to track service as oracles execution times in blocks.

Note that although blocks are rejected if the total service oracle execution time exceed the limit, there is no timeout mechanism to preempt a service that is taking too long. This requires more work and will be done later (#3533).

## Test Plan

Some unit tests were added using `MockApplication`s to test service as oracle calls.

## Release Plan

- Backporting is not possible but we may want to deploy a new `devnet` and `testnet` and release a new
      SDK soon, because this contains backwards incompatible changes to the chain state.

## Links

- Depends on #3534 
- Closes #3528


linera-chain/src/unit_tests/chain_tests.rs
```diff
@@ -284,3 +286,191 @@ async fn test_application_permissions() -> anyhow::Result<()> {
...
     Ok(())
 }
+
+/// Tests if services can execute as oracles if the total execution time is less than the limit.
+#[test_case(&[100]; "single service as oracle call")]
+#[test_case(&[50, 50]; "two service as oracle calls")]
+#[test_case(&[90, 10]; "long and short service as oracle calls")]
+#[test_case(&[33, 33, 33]; "three service as oracle calls")]
+#[tokio::test]
+async fn test_service_as_oracles(service_oracle_execution_times_ms: &[u64]) -> anyhow::Result<()> {
```
deuszx, line 296: Couldn't we merge the two test cases into one by accepting the `maximum_service_oracle_execution_ms` as the second argument to the test?

linera-execution/src/runtime.rs
```diff
@@ -460,6 +461,45 @@ impl SyncRuntimeInternal<UserContractInstance> {
...
+        let _timeout = self
+            .resource_controller
+            .remaining_service_oracle_execution_time()?;
+        let execution_start = Instant::now();
+        let result = service_runtime.run_query(application_id, query);
+
+        // Always track the execution time, irrespective to whether the service ran successfully or
+        // timed out
+        self.resource_controller
+            .track_service_oracle_execution(execution_start.elapsed())?;
```
deuszx, line 493: so the timeout is shared across all calls within the block? it's not per-call.

jvff: That is correct.

linera-chain/src/unit_tests/chain_tests.rs
jvff: We could, but I try to avoid having conditional assertions in tests, especially in unit tests. I try to stick to the "setup, execute, assert" pattern, and make these as obvious and streamlined as possible. I'd rather try to create some test helper functions to make the tests shorter instead. What do you think?

deuszx: If you prefer it that way - sure.

jvff: Done. At least it's a start, we can improve this more later as well.

PR #3541, opened 2025-03-11
Add timeouts for HTTP requests

### LINKED ISSUE #3522: Epoch configurable HTTP timeouts
# Motivation

HTTP requests can take a long time to complete, which could be abused by malicious applications.

# Proposal

Add a `maximum_http_timeout` field to `ResourceControlPolicy` and ensure that HTTP requests respect it.

jvff:
## Motivation

HTTP requests can take a long time to complete, or attempt to stay open indefinitely. This should not be allowed because it may stall a chain.

## Proposal

Add a `http_request_timeout` policy to `ResourceControlPolicy`, and configure all HTTP requests to use that timeout.

## Test Plan

Wrote some [tests](https://github.com/jvff/linera-protocol-archive/compare/95da70cabb..test-http-timeouts-p1) out of this branch, possibly to be added to #3509.

## Release Plan

- Backporting is not possible but we may want to deploy a new `devnet` and `testnet` and release a new
      SDK soon, because this contains backwards incompatible changes to the chain state.

## Links

- Closes #3522 

### DIFF
--- a/linera-execution/src/execution_state_actor.rs
+++ b/linera-execution/src/execution_state_actor.rs
@@ -5,6 +5,8 @@
 
 #[cfg(with_metrics)]
 use std::sync::LazyLock;
+#[cfg(not(web))]
+use std::time::Duration;
 
 use custom_debug_derive::Debug;
 use futures::channel::mpsc;
@@ -395,12 +397,20 @@ where
                     ExecutionError::UnauthorizedHttpRequest(url)
                 );
 
-                let response = Client::new()
+                #[cfg_attr(web, allow(unused_mut))]
+                let mut request = Client::new()
                     .request(request.method.into(), url)
                     .body(request.body)
-                    .headers(headers)
-                    .send()
-                    .await?;
+                    .headers(headers);
+                #[cfg(not(web))]
+                {
+                    request = request.timeout(Duration::from_millis(
+                        committee.policy().http_request_timeout_ms,
+                    ));
+                }
+
+                let response = request.send().await?;
+
                 callback.respond(http::Response::from_reqwest(response).await?);
             }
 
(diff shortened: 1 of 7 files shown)


linera-execution/src/execution_state_actor.rs
```diff
@@ -395,12 +397,20 @@ where
...
-                let response = Client::new()
+                #[cfg_attr(web, allow(unused_mut))]
+                let mut request = Client::new()
                     .request(request.method.into(), url)
                     .body(request.body)
-                    .headers(headers)
-                    .send()
-                    .await?;
+                    .headers(headers);
+                #[cfg(not(web))]
```
afck, line 405: …or use the same timeout in either case? Might be less surprising, too.

linera-execution/src/test_utils/mock_application.rs
afck, line 1: Why these changes?

afck (approved):

linera-execution/src/execution_state_actor.rs
```diff
@@ -395,12 +397,20 @@ where
                     ExecutionError::UnauthorizedHttpRequest(url)
                 );
 
-                let response = Client::new()
+                #[cfg_attr(web, allow(unused_mut))]
+                let mut request = Client::new()
```
deuszx, line 401: So on a web client there won't be any timeouts?

linera-execution/src/execution_state_actor.rs
jvff: We can't, `reqwest` doesn't support timeouts on Wasm :/

linera-execution/src/execution_state_actor.rs
jvff: Only the default browser timeouts. Unfortunately we can't add timeouts with `reqwest` for Wasm, we'd need to use Javascript APIs directly :(

PR #3544, opened 2025-03-11
Add runtime API for services to stop executing on timeout

### LINKED ISSUE #3530: Add function to check service execution time
# Motivation

Since we can't easily cancel the thread executing the service as an oracle, we need a different way to limit its execution time.

The solution we have currently 

# Proposal

Stub a `check_execution_time` function to the `ServiceRuntimeApi`, so that it's available for Wasm services to call periodically.

jvff:
## Motivation

Services running as oracles should run in bounded time. Since adding a timeout was not possible, the Wasm runtimes' metering services need to be used. The goal is to instrument the services in order to force them to periodically check if they have exceeded their execution time, and stop executing if they have. For this to be possible, the runtime needs a new function to be called from the services.

## Proposal

Add a `ServiceRuntime::check_execution_time` runtime API for services, and export it to the WIT interface. Add an optional `deadline` to the `SyncRuntimeInternal` to be used by services running as oracles. When the new runtime API is called, return an error if it is has passed its deadline.

## Test Plan

A unit test was added to check that execution stops early.

## Release Plan

    - be released in a new SDK,
    - be released in a new SDK,
      SDK soon.

## Links

- Depends on #3536 
- Closes #3530 

### DIFF
--- a/linera-chain/src/unit_tests/chain_tests.rs
+++ b/linera-chain/src/unit_tests/chain_tests.rs
@@ -3,7 +3,11 @@
 
 #![allow(clippy::large_futures)]
 
-use std::{collections::BTreeMap, iter, thread, time::Duration};
+use std::{
+    collections::BTreeMap,
+    iter, thread,
+    time::{Duration, Instant},
+};
 
 use assert_matches::assert_matches;
 use linera_base::{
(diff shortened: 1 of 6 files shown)


bart-linera (commented):
Looks good, but I have a question - how do we make sure that services call this new `check_execution_time` function? Is this supposed to be taken care of by this PR, or will it be added later? (I'm probably missing something - likely how metering works :sweat_smile: )

jvff:
This will be done through instrumentation. It will be the same that is done for contracts, configured here:

https://github.com/linera-io/linera-protocol/blob/c324ed5eca4e55f0e8cf0b502b98416617b8b323/linera-execution/src/wasm/wasmer.rs#L205-L245

But for contracts the callback function is `consume_fuel`:

https://github.com/linera-io/linera-protocol/blob/c324ed5eca4e55f0e8cf0b502b98416617b8b323/linera-execution/src/wasm/runtime_api.rs#L602-L612

bart-linera (approved):

PR #3546, opened 2025-03-11
Fees for publishing and reading blobs.

### LINKED ISSUE #2172: Publishing and execution fees

afck:
## Motivation

Publishing and reading blobs creates cost on the validator side and therefore needs to be paid for with fees.

## Proposal

Add fee categories for publishing `Data` and `Bytecode`, and reading `Data` blobs.

Make `PublishCommitteeBlob` an `AdminOperation`: It is free, so users should not be allowed to do it.

## Test Plan

A client test was added.

## Release Plan

## Links

- Closes #2172.


CLI.md
```diff
@@ -452,6 +452,8 @@ View or update the resource control policy
 * `--write-operation <WRITE_OPERATION>` — Set the price per write operation
 * `--byte-read <BYTE_READ>` — Set the price per byte read
 * `--byte-written <BYTE_WRITTEN>` — Set the price per byte written
+* `--blob-byte-read <BLOB_BYTE_READ>` — Set he price to read a blob, per byte
+* `--blob-byte-published <BLOB_BYTE_PUBLISHED>` — The price to publish a blob, per byte
```
deuszx, line 456: Why are these two not the same as `--byte-read` and `--byte-written`?

deuszx: Also, I'm confused - the argument is called `blob byte read` (note absence of _price_ word), the comment says it sets the price (per byte) AND there's another setting called `blob byte read PRICE`.

linera-storage/src/lib.rs
```diff
@@ -429,8 +429,18 @@ where
         }
     }
 
-    async fn get_blob(&self, blob_id: BlobId) -> Result<Blob, ViewError> {
-        self.storage.read_blob(blob_id).await
+    async fn get_blobs(&self, blob_ids: BlobId) -> Result<Blob, ViewError> {
+        let maybe_blobs = self.storage.read_blobs(blob_ids).await;
```
deuszx, line 433: How does it work - `blob_ids` is still a single `Blobid` type.

afck: (Just wasn't done yet. Fixed now.)

linera-core/src/chain_worker/actor.rs
```diff
@@ -85,6 +85,7 @@ where
     StageBlockExecution {
         block: ProposedBlock,
         round: Option<u32>,
+        published_blobs: Vec<Blob>,
```
ma2bd, line 88: why is this added now?

ma2bd:
Need to update summary

linera-core/src/chain_worker/actor.rs
afck: Up until now we didn't need the published blobs inside `ChainStateView::execute_block`: We were checking their size before that, during execution only needed to make a note of their IDs (not their content), and then we store them after execution.

Now we handle fees (and the size check) inside `execute_block`.

linera-core/src/unit_tests/client_tests.rs
```diff
@@ -2334,10 +2334,51 @@ where
...
+        ) if matches!(**error, ExecutionError::BlobTooLarge))
     );
 
     Ok(())
 }
+
+#[test_case(MemoryStorageBuilder::default(); "memory")]
+#[cfg_attr(feature = "storage-service", test_case(ServiceStorageBuilder::new().await; "storage_service"))]
+#[test_log::test(tokio::test)]
+async fn test_blob_fees<B>(storage_builder: B) -> anyhow::Result<()>
```
bart-linera, line 2350: Should we also test that doing other stuff, like publishing bytecodes, or creating an application, does not generate fees?

bart-linera (approved):

linera-base/src/data_types.rs
```diff
@@ -277,6 +277,14 @@ pub struct Resources {
     pub bytes_to_read: u32,
     /// A number of bytes to write.
     pub bytes_to_write: u32,
+    /// A number of blobs to read.
+    pub blobs_to_read: u32,
+    /// A number of blobs to publish.
+    pub blobs_to_publish: u32,
```
deuszx, line 283: I'm not sure these two should have seaprate config values. I think the blobs reading/writing (publishing) should be limited by the blob bytes limits. I.e. it shouldn't matter if we're reading 1 blob twice the size of two smaller blobs half the size.

afck: I imagine that the cost of reading a 10-byte blob is almost the same as reading a 0-byte blob: Both require a roundtrip to the database, and very little payload. Also, reading a 0-byte blob shouldn't be free, since it does incur a cost. And we need to charge something for when a contract asserts the existence of a blob.

deuszx: There should be a constant cost for reaching to a db and then the per-byte cost.

afck: We are.
This struct here is only used for _grants_, where an application wants to pay for the execution of a message elsewhere, up to a limit. That limit is specified in resources.

linera-execution/src/system.rs
```diff
@@ -556,11 +561,14 @@ where
             PublishDataBlob { blob_hash } => {
                 self.blob_published(&BlobId::new(blob_hash, BlobType::Data))?;
             }
-            PublishCommitteeBlob { blob_hash } => {
-                self.blob_published(&BlobId::new(blob_hash, BlobType::Committee))?;
-            }
             ReadBlob { blob_id } => {
-                self.read_blob_content(blob_id).await?;
+                let content = self.read_blob_content(blob_id).await?;
+                if blob_id.blob_type == BlobType::Data {
```
deuszx, line 566: Why do we not track the other two `BlobType`s as well?

linera-base/src/data_types.rs
deuszx: 👍 

I'd make all limits gas-based.

linera-execution/src/system.rs
afck: * We want all other chains to keep with the latest epoch as much as possible, and the only way to read a `Committee` blob is to do exactly that. Each chain can only read each `Committee` blob exactly once (or: as often as it occurs in different epochs), and we want them all to do that.
* For application description and bytecode blobs it's a bit debatable, but it's not a great user experience if the price of executing a transaction or operation, before the execution even starts, already depends on the size of the bytecode or parameters. Also, we cache modules in the chain worker, I think, so the reading cost for the validator only occurs whenever they had evicted that chain worker and are reloading it, which isn't deterministic.

linera-core/src/unit_tests/client_tests.rs
```diff
@@ -1132,7 +1132,19 @@ async fn test_change_voting_rights<B>(storage_builder: B) -> anyhow::Result<()>
 where
     B: StorageBuilder,
 {
-    let mut builder = TestBuilder::new(storage_builder, 4, 1).await?;
+    // To test that no fees are paid for reading or publishing committee blobs, we set the price
+    // higher than the chain balance.
```
deuszx, line 1136: To be honest I can't see where this is set. I see `30_000` and `Amount::from_tokens(10)`. Is `maximum_fuel_per_block` denominated in different units than the rest of config values?

afck: No, it's just that we set the chain balances to 3 resp. 1 below (lines 1148 and 1149), so they couldn't pay for any of these and the blocks would fail.

deuszx: I'd extract to a variable then - like `let initial_balance = Amount::from_tokens(3)` and then set the price that is a multiple of that. It'll be clearer.

linera-execution/src/execution_state_actor.rs
```diff
@@ -402,12 +403,25 @@ where
...
 
             AssertBlobExists { blob_id, callback } => {
                 self.system.assert_blob_exists(blob_id).await?;
+                // Treating this as reading a size-0 blob for fee purposes.
+                if blob_id.blob_type == BlobType::Data {
+                    resource_controller
+                        .with_state(&mut self.system)
+                        .await?
+                        .track_blob_read(0)?;
+                }
```
deuszx, line 424: This won't get recorded in the case when blob does not exist.

linera-execution/src/execution_state_actor.rs
```diff
@@ -402,12 +403,25 @@ where
 
             ReadBlobContent { blob_id, callback } => {
                 let blob = self.system.read_blob_content(blob_id).await?;
+                if blob_id.blob_type == BlobType::Data {
+                    resource_controller
+                        .with_state(&mut self.system)
+                        .await?
+                        .track_blob_read(blob.bytes().len() as u64)?;
+                }
```
deuszx, line 411: There should be some constant cost for trying to read a blob even when it doesn't exist - here, if the blob is not found we'll return early and not track any reading-related cost.

linera-core/src/chain_worker/state/mod.rs
```diff
@@ -210,13 +211,18 @@ where
...
+        let Some(published_blobs) = ChainWorkerStateWithAttemptedChanges::new(&mut *self)
             .await
             .validate_block(&proposal)
-            .await?;
+            .await?
+        else {
+            // Skipping: We already voted for this block.
+            let info = ChainInfoResponse::new(&self.chain, self.config.key_pair());
+            return Ok((info, NetworkActions::default()));
+        };
```
deuszx, line 222: I don't think it's a good design where `validate_block` returns `Option<published_blobs>` and when that's `None` it means we already voted for this block. That's at least two separate responsibilities mixed in one method.

deuszx (commented):
No hard blockers but I find our gas-accounting (resource control) very hard to follow. Some actions are tracked in `execute_block`, some in `execute_operation`. `ExecutionRequest` has `ReadBlobContent` and `AssertBlobExists` variants that we also track but no variant related to publishing blob. I'm sure I'm missing something but it'd be simpler to validate if those things (resource control/gas accounting) were in a single place.

linera-execution/src/execution_state_actor.rs
afck: But then we fail the whole block anyway.

linera-execution/src/execution_state_actor.rs
afck: Yes, then the whole block is invalid.

linera-execution/src/execution_state_actor.rs
deuszx: Ah, I wanted to say it should be paid for up until that point in execution but we don't have invalid blocks.

afck: That's a discussion that goes well beyond the scope of this PR. Currently failing operations make the block invalid.

deuszx (approved):
No blockers.

afck:
Thank you! I'll merge this and see if I can address some of your comments in a follow-up PR, specifically improving the test readability and the `validate_block` return type.

Regarding the gas accounting I'm not sure yet. I'll probably pass the published blobs into `execute_operation`, so the fees are calculated when the operation is processed.
But some fees, of course, necessarily remain in `execute_block`, like the base block fee and block size fee.

PR #3547, opened 2025-03-11
Implement multiplem address for EVM.

MathieuDutSik:
## Motivation

In order to execute EVM smart contract, we need to have access to the Linera address and to the EVM address.
The call in the EVM to another EVM smart contract is done via using the EVM address.
The implementation would be done (in next PR) in the following way:
* The cross contract call is intercepted by the inspector.
* The inspector translate the EVM address into Linera address.
* The inspector calls the other contract via the `try_call_application`.
* The inspector returns the value.

This PR is the first step in the conversion.

## Proposal

The following is done:
* The Address is added to the `Evm` option of the enum. For Wasm, nothing is needed as only one address is needed.
* The `Address` needs to be encapsulated into a `EncapsulateAddress` in order to implement the `Wit*` traits.

For the effective implementation there are two options:
* One is that the contracts that are needed are encoded into the `required_application_ids`. This could be accessed from the `BaseRuntime`. This would of course limit the list of available contracts to the one that are put in advance (a limitation which is not in EVM).
* Or any address could be used. This would also require adding an entry to the `system` of `ExecutionStateView` mapping EVM address to Linera address and this could be accessed by adding functionalities to the `BaseRuntime`.

Regardless, these two options depend on the insertion of the address into the `ModuleId`.

The first goal is to implement the EVM to EVM calls where there is no opportunity for change.

## Test Plan

Basic tests are done in the `vm.rs`.

## Release Plan

## Links

None

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4396,6 +4396,7 @@ dependencies = [
 name = "linera-base"
 version = "0.14.0"
 dependencies = [
+ "alloy",
  "alloy-primitives",
  "anyhow",
  "async-graphql",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -338,6 +338,7 @@ dependencies = [
  "const-hex",
  "derive_more",
  "foldhash",
+ "getrandom",
  "hashbrown 0.15.2",
  "indexmap 2.7.0",
  "itoa",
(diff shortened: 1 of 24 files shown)


MathieuDutSik:
This is not the direction that we want to pursue and so the PR is closed.

PR #3548, opened 2025-03-11
Charge for every query to a service as oracle

### LINKED ISSUE #3529: Charge a fixed fee to call services as oracles
# Motivation

Executing a service as an oracle requires changing the metering model from instruction based to time based. This switch also has a cost, so it must be charged.

# Proposal

Add a fee for calling a service as an oracle.

jvff:
## Motivation

There's a non-negligible performance impact due to the context switch from the contract to the service, and this should be charged.

## Proposal

Charge a fixed fee for every service as an oracle query. Allow the price to be charged to be determined as a policy.

## Test Plan

Some fee calculation unit tests were added.

## Release Plan

      SDK soon, because this contains breaking changes to the on-chain resource control policy.

## Links

- Closes #3529 
- Depends on #3544


bart-linera (approved):

linera-execution/src/runtime.rs
```diff
@@ -1414,6 +1414,8 @@ impl ContractRuntime for ContractSyncRuntimeHandle {
 
         let response =
             if let Some(response) = this.transaction_tracker.next_replayed_oracle_response()? {
+                this.resource_controller
+                    .track_service_oracle_execution(Duration::ZERO)?;
```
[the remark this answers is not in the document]
jvff: It doesn't in terms of execution time, but it does update the balance. The goal was to charge the fee in one place in the code, but now in hindsight I think this was the wrong choice. Not just because it's confusing, but because it hides that the call is being charged and because it charges _after_ the call has been made.

I'll post an update to fix this, thanks!

linera-base/src/data_types.rs
```diff
@@ -284,6 +284,8 @@ pub struct Resources {
     pub message_size: u32,
     /// An increase in the amount of storage space.
     pub storage_size_delta: u32,
+    /// A number of service as oracle requests to be performed.
```
deuszx, line 287:
```suggestion
    /// A number of service-as-oracle requests to be performed.
```

It somehow feels better this way.

linera-execution/src/policy.rs
```diff
@@ -41,6 +41,8 @@ pub struct ResourceControlPolicy {
     pub message: Amount,
     /// The additional price for each byte in the argument of a user message.
     pub message_byte: Amount,
+    /// The price to query a service as an oracle.
```
deuszx, line 44:
```suggestion
    /// The price per query to a service as an oracle.
```
?

deuszx (approved):
No blockers.

PR #3551, opened 2025-03-12
fix: fix type issue in `test_key` function

atakoraka:
## Motivation

There was an inconsistency in the code where `SecretKey` was used in one place, while other parts of the code expected `SigningKey`. This could lead to compilation issues or runtime errors. The goal is to ensure consistency in the types being used across the codebase.

## Proposal

Change `k256::SecretKey::random` to `k256::ecdsa::SigningKey::random` and update the method for getting the public key to `verifying_key()`. This ensures the code works as expected and maintains type consistency.

## Test Plan

- Run the unit tests to ensure that no compilation errors or runtime issues occur after the change.
- Verify that the `test_key` function now correctly returns a public key from a `SigningKey`.

### DIFF
--- a/linera-base/src/crypto/secp256k1.rs
+++ b/linera-base/src/crypto/secp256k1.rs
@@ -67,8 +67,8 @@ impl Secp256k1PublicKey {
     pub fn test_key(seed: u8) -> Self {
         use rand::SeedableRng;
         let mut rng = rand::rngs::StdRng::seed_from_u64(seed as u64);
-        let sk = k256::SecretKey::random(&mut rng);
-        Self(sk.public_key().into())
+        let sk = k256::ecdsa::SigningKey::random(&mut rng);
+        Self(sk.verifying_key().into())
     }
 
     /// Returns the bytes of the public key in compressed representation.


deuszx:
SecretKey/SigningKey in `k256` create are interchangeable in a sense. Their usage in the method is correct. if it wasn't the code wouldn't have compiled.

atakoraka:
> SecretKey/SigningKey in `k256` create are interchangeable in a sense. Their usage in the method is correct. if it wasn't the code wouldn't have compiled.

I get that `SecretKey` and `SigningKey` are interchangeable in `k256`, but using `SigningKey` directly here makes the code clearer and more consistent. Since we eventually need a `SigningKey` to call `verifying_key()`, it makes sense to initialize it upfront.  

This also avoids implicit conversions and makes the intent of the code more obvious. Plus, if the crate changes in the future, using the correct type directly reduces the risk of subtle bugs.  

I believe this approach is cleaner and more maintainable—what do you think?

deuszx:
> > SecretKey/SigningKey in `k256` create are interchangeable in a sense. Their usage in the method is correct. if it wasn't the code wouldn't have compiled.
> 
> I get that `SecretKey` and `SigningKey` are interchangeable in `k256`, but using `SigningKey` directly here makes the code clearer and more consistent. Since we eventually need a `SigningKey` to call `verifying_key()`, it makes sense to initialize it upfront.
> 
> This also avoids implicit conversions and makes the intent of the code more obvious. Plus, if the crate changes in the future, using the correct type directly reduces the risk of subtle bugs.
> 
> I believe this approach is cleaner and more maintainable—what do you think?

I think it doesn't make any difference because the caller gets the same type and it's an internal detail whether we use `Secret*` or `Signing*` types. They're aliases in k256 anyway.

atakoraka:
> I think it doesn't make any difference because the caller gets the same type and it's an internal detail whether we use `Secret*` or `Signing*` types. They're aliases in k256 anyway.

That makes sense since the caller gets the same type, but I still think using `SigningKey` directly has benefits:  

1. **Clarity**: It explicitly reflects the intended use (signing) rather than relying on an alias.  
2. **Consistency**: Other parts of the code already work with `SigningKey`, so aligning here avoids mental overhead.  
3. **Future-proofing**: If `k256` changes the relationship between these types, using the correct one directly reduces potential breakages.

ma2bd:
Closing this following @deuszx' comments

PR #3553, opened 2025-03-12
Limit HTTP response size

### LINKED ISSUE #3521: Limit size of HTTP responses
# Motivation

HTTP responses can have large or indeterminate size (for example with streaming responses). This is a security issue, because if the validators have very fast connections, applications can generate a lot of traffic before a timeout happens.

# Proposal

Ensure that HTTP response sizes (including headers) are limited, with a new `maximum_http_response_size` configuration in `ResourceControlPolicy`.

jvff:
## Motivation

Applications could cause excessive bandwidth costs to validators by performing HTTP requests to obtain large amounts of data as fast as possible.

## Proposal

Ensure that HTTP responses are limited in size to a value agreed among the validators.

## Test Plan

Wrote some unit tests in a separate branch, using the code from #3509:

https://github.com/jvff/linera-protocol-archive/blob/test-http-response-limits-p1/examples/how-to/perform-http-requests/tests/response_size_tests.rs

## Release Plan

      SDK soon, because this contains changes to the consensus critical `ResourceControlPolicy`.

## Links

- Closes #3521 


linera-execution/src/policy.rs
```diff
@@ -61,6 +61,8 @@ pub struct ResourceControlPolicy {
     pub maximum_bytes_read_per_block: u64,
     /// The maximum data to write per block
     pub maximum_bytes_written_per_block: u64,
+    /// The maximum size of a received HTTP response.
+    pub maximum_http_response_size: u64,
```
deuszx, line 65: Can you add a unit to both the name and comment?

linera-execution/src/lib.rs
```diff
@@ -234,6 +234,8 @@ pub enum ExecutionError {
     MaximumFuelExceeded,
     #[error("Serialized size of the executed block exceeds limit")]
     ExecutedBlockTooLarge,
+    #[error("HTTP response exceeds the size limit")]
+    HttpResponseSizeLimitExceeded,
```
deuszx, line 238: Can we add the limit to the error?

linera-execution/src/execution_state_actor.rs
```diff
@@ -425,6 +431,70 @@ where
...
+    ) -> Result<http::Response, ExecutionError> {
+        let status = response.status().as_u16();
+        let maybe_content_length = response.content_length();
+
+        let headers = mem::take(response.headers_mut())
+            .into_iter()
+            .scan(None, |previous_name, (maybe_name, value)| {
+                let name = maybe_name
+                    .or(previous_name.take())
+                    .expect("All headers should have names");
```
deuszx, line 455: Do we catch this panic?

deuszx: OK but we shouldn't let panic in this crate crash our nodes. I'd say it's the problem with the server that responds to the query so the execution should simply fail.

linera-execution/src/policy.rs
jvff: Done.

linera-execution/src/lib.rs
jvff: Done.

linera-execution/src/lib.rs
```diff
@@ -236,6 +236,8 @@ pub enum ExecutionError {
     MaximumServiceOracleExecutionTimeExceeded,
     #[error("Serialized size of the executed block exceeds limit")]
     ExecutedBlockTooLarge,
+    #[error("HTTP response exceeds the size limit of {limit} bytes, having at least {size} bytes")]
+    HttpResponseSizeLimitExceeded { limit: u64, size: u64 },
```
ma2bd, line 240: Should we make this a limit for all types of oracle responses?

linera-execution/src/execution_state_actor.rs
deuszx: I'm gonna insist that we don't panic here.

jvff: This is not caused by a problem with the server, but the presence of a bug in the `reqwest` crate.

But let me see if I can refactor it to avoid the `expect`.

deuszx: We can return an error here and handle that on the caller site.

linera-execution/src/execution_state_actor.rs
```diff
@@ -430,6 +436,79 @@ where
...
+                limit: size_limit,
+                size: total_header_size,
+            },
+        )?;
+
+        if let Some(content_length) = maybe_content_length {
+            if content_length > remaining_bytes {
+                return Err(ExecutionError::HttpResponseSizeLimitExceeded {
+                    limit: size_limit,
+                    size: content_length + (size_limit - remaining_bytes),
```
deuszx, line 485: I'd just return `content_length + total_header_size`.

deuszx (changes requested):
Let's not panic if the header is misformatted.

linera-execution/src/execution_state_actor.rs
jvff: I did something different. I think it's better now.

linera-execution/src/execution_state_actor.rs
```diff
@@ -430,6 +435,71 @@ where
...
+        let headers = response
+            .headers()
+            .iter()
+            .map(|(name, value)| http::Header::new(name.to_string(), value.as_bytes()))
+            .collect::<Vec<_>>();
+
+        let total_header_size = headers
+            .iter()
+            .map(|header| (header.name.as_bytes().len() + header.value.len()) as u64)
+            .sum();
```
deuszx, line 463: I wonder - could we collect the sizes when we map `response.headers()` already? we have the data there (`name`, `value`, etc.).

jvff:
> Let's not panic if the header is misformatted.

I'll insist. It did _not_ panic if the header is misformatted. It would have panicked if the _http_ crate had a bug that would return `None` for the first header name. Here is the code:

https://docs.rs/http/0.2.12/src/http/header/map.rs.html#3039

Note that `IntoIter.next` starts as `None`: https://docs.rs/http/0.2.12/src/http/header/map.rs.html#1994

linera-execution/src/execution_state_actor.rs
jvff: Done.

linera-execution/src/execution_state_actor.rs
jvff: We could, but I think doing it separately is cleaner.

linera-execution/src/lib.rs
jvff: Oh right, I'll open an issue to limit the oracle response size as well.

linera-execution/src/execution_state_actor.rs
deuszx: but not faster :)

deuszx (approved):

linera-execution/src/lib.rs
jvff: Issue #3563 created.

linera-execution/src/execution_state_actor.rs
jvff: Premature optimization is the root of all evil ;)

PR #3560, opened 2025-03-13
Make Chain a variant of AccountOwner

deuszx:
## Motivation

We want to fix and expand our notion of Address (currently Owner, GenericApplicationId, etc.) to different type of addresses (32-byte Linera/Solana, 20-byte EVM). In order to do that we need to prepare the code for the introduction of new variants.

## Proposal

Inline a `Chain` into `AccountOwner` enum to identify cases where transactions (mostly token transfers) are targeting or using chain's account balance. Previously that case was handled with the usage of `Option<AccountOwner>`. This made refactoring more difficult.

## Test Plan

CI should catch regressions.

## Release Plan

## Links


linera-execution/src/lib.rs
```diff
@@ -1028,15 +1028,14 @@ impl<Message> RawOutgoingMessage<Message, Resources> {
 }
 
 impl OperationContext {
-    fn refund_grant_to(&self) -> Option<Account> {
-        // TODO: return `Account` instead.
-        Some(Account {
+    fn refund_grant_to(&self) -> Account {
```
deuszx, line 1031: TODO: double-check. I think this is wrong b/c previously `None` was kept as `None` and rn we use `AccountOwner::Chain`.

deuszx: This is correct. It always returned `Some(_)` and `owner: None` was already refunding chain.

ma2bd: The option type controls whether we refund or not.

deuszx: Yes, and that part is kept intact. Whereas this method used to always return `Some(_)` I know wrap the returned value with `Some(_)`.

examples/amm/src/contract.rs
```diff
@@ -326,6 +326,9 @@ impl AmmContract {
                     "Unauthorized"
                 )
             }
+            AccountOwner::Chain => {
+                unreachable!("Chains cannot be used to authenticate")
+            }
```
bart-linera, line 331: Those `unreachable!`s suggest to me that we might be conflating two different things here: who can authenticate a transaction, and where tokens can be transferred to. Maybe those should be represented by separate types? (Not necessarily something for this PR.)

bart-linera (commented):
I'm not sure I agree with the premise. It seems like a chain as a transfer target doesn't really belong as a variant of `AccountOwner`. If we want to be more explicit, I'd just make a type like `enum TransferTarget { AccountOwner(AccountOwner), Chain }` and replace `Option<AccountOwner>` with it. Then `AccountOwner` would play the role of identifying someone owning tokens that can authenticate transfers, and `TransferTarget` would represent, well, a target of a transfer.

I'm just not sure if that helps with solving the problem this was intended to solve, but I guess new potential variants of "addresses" could then be new variants of `TransferTarget`?

examples/amm/src/contract.rs
deuszx: I think the problem is deeper - we seem to be authenticating actions on a wrong layer and using fragile mechanism. We shouldn't rely on a "type" of the code living under an address to decide whether an action can be authorized or not.

deuszx:
> I'm not sure I agree with the premise. It seems like a chain as a transfer target doesn't really belong as a variant of `AccountOwner`. If we want to be more explicit, I'd just make a type like `enum TransferTarget { AccountOwner(AccountOwner), Chain }` and replace `Option<AccountOwner>` with it. Then `AccountOwner` would play the role of identifying someone owning tokens that can authenticate transfers, and `TransferTarget` would represent, well, a target of a transfer.
> 
> I'm just not sure if that helps with solving the problem this was intended to solve, but I guess new potential variants of "addresses" could then be new variants of `TransferTarget`?

I think the _ultimate_ solution would be to simply remove that variant entirely and either hardcode that every chain has an account with some hardcoded address (`[0u8; 32]`) where we store its balance or remove that entirely - i.e. chains cannot have token balance at all.

linera-execution/src/lib.rs
```diff
@@ -1176,7 +1179,7 @@ impl Operation {
     /// Creates a new user application operation following the `application_id`'s [`Abi`].
     #[cfg(with_testing)]
     pub fn user<A: Abi>(
-        application_id: UserApplicationId<A>,
+        application_id: ApplicationId<A>,
```
bart-linera, line 1182: This (and below) seems to be a part of a different change?

linera-execution/src/system.rs
```diff
@@ -711,6 +707,7 @@ where
                 authenticated_application_id == Some(owner),
                 ExecutionError::UnauthenticatedClaimOwner
             ),
+            AccountOwner::Chain => unreachable!(),
```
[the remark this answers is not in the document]
deuszx: Sounds reasonable to mirror logic of `transfer` here. I didn't decide to do it b/c I would change the logic - i.e. the `source` here wasn't wrapped with `Option`.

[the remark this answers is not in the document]
deuszx: cc @afck

examples/amm/src/contract.rs
ma2bd: `unreachable!` is inaccurate and the error message is misleading.
```
panic!("Using chain balance is not authorized")
```

examples/fungible/src/contract.rs
```diff
@@ -143,6 +143,9 @@ impl FungibleTokenContract {
                     "The requested transfer is not correctly authenticated."
                 )
             }
+            AccountOwner::Chain => {
+                unreachable!("Chains cannot be used to authenticate")
```
ma2bd, line 147: `panic!("Chain account is not supported")`

ma2bd: We could totally support it by the way: it would authorized from operations and unauthorized from messages.

examples/native-fungible/src/contract.rs
```diff
@@ -144,6 +144,9 @@ impl NativeFungibleTokenContract {
                 );
             }
             AccountOwner::Application(_) => panic!("Applications not supported yet"),
+            AccountOwner::Chain => {
+                panic!("Chains cannot be used to authenticate")
```
ma2bd, line 148: `"Chain accounts are not supported"`

examples/native-fungible/src/contract.rs
```diff
@@ -144,6 +144,9 @@ impl NativeFungibleTokenContract {
                 );
             }
             AccountOwner::Application(_) => panic!("Applications not supported yet"),
```
ma2bd, line 146: We should fix that.

linera-base/src/identifiers.rs
```diff
@@ -306,7 +304,7 @@ pub struct ApplicationId<A = ()> {
 
 /// Alias for `ApplicationId`. Use this alias in the core
 /// protocol where the distinction with the more general enum `GenericApplicationId` matters.
-pub type UserApplicationId<A = ()> = ApplicationId<A>;
+pub type UserApplicationId = ApplicationId<()>;
```
ma2bd, line 307: I'm ok with this.

examples/non-fungible/src/contract.rs
```diff
@@ -144,6 +144,9 @@ impl NonFungibleTokenContract {
                     "The requested transfer is not correctly authenticated."
                 )
             }
+            AccountOwner::Chain => {
+                unreachable!("Chains cannot be used to authenticate")
```
ma2bd, line 148: `panic!("Chain accounts are not supported")`

linera-base/src/identifiers.rs
```diff
@@ -990,6 +991,7 @@ impl Display for AccountOwner {
         match self {
             AccountOwner::User(owner) => write!(f, "User:{}", owner)?,
             AccountOwner::Application(app_id) => write!(f, "Application:{}", app_id)?,
+            AccountOwner::Chain => write!(f, "Chain")?,
```
ma2bd, line 994: I look forward to remove these tags

linera-core/src/chain_worker/state/temporary_changes.rs
```diff
@@ -219,9 +219,12 @@ where
...
         }
-        if let Some(owner) = query.request_owner_balance {
-            info.requested_owner_balance =
-                chain.execution_state.system.balances.get(&owner).await?;
+        match query.request_owner_balance {
+            owner @ AccountOwner::Application(_) | owner @ AccountOwner::User(_) => {
+                info.requested_owner_balance =
+                    chain.execution_state.system.balances.get(&owner).await?;
+            }
+            _ => {}
```
ma2bd, line 227: Let's return the chain balance!

linera-execution/src/system.rs
```diff
@@ -419,7 +417,7 @@ where
                     .transfer(
                         context.authenticated_signer,
                         None,
-                        owner.map(AccountOwner::User),
+                        owner.map(AccountOwner::User).unwrap_or(AccountOwner::Chain),
```
ma2bd, line 420: We should instead modify `SystemOperation` to use `SystemOperation::Transfer { owner: Owner .. }` and resolve the default value way earlier.

linera-execution/src/lib.rs
```diff
@@ -1299,7 +1302,7 @@ impl Query {
 
     /// Creates a new user application query following the `application_id`'s [`Abi`].
     pub fn user<A: Abi>(
-        application_id: UserApplicationId<A>,
+        application_id: ApplicationId<A>,
```
ma2bd, line 1305: yes this is great

linera-execution/src/runtime.rs
```diff
@@ -1272,7 +1272,7 @@ impl ContractRuntime for ContractSyncRuntimeHandle {
 
     fn transfer(
         &mut self,
-        source: Option<AccountOwner>,
+        source: AccountOwner,
```
ma2bd, line 1275: yes

linera-execution/src/system.rs
```diff
@@ -375,7 +373,7 @@ where
         use SystemOperation::*;
         let mut outcome = RawExecutionOutcome {
             authenticated_signer: context.authenticated_signer,
-            refund_grant_to: context.refund_grant_to(),
+            refund_grant_to: Some(context.refund_grant_to()),
```
ma2bd, line 376: No. We should not force refunds.

ma2bd (changes requested):
Thanks for the PR.

No major issue except that `refund_grand_to` should be an option.

linera-execution/src/system.rs
ma2bd: `unwrap_or(AccountOwner::Chain)` is a little bit of a red flag. (After this PR, only CLI should be allowed to default to a chain account)

linera-core/src/chain_worker/state/temporary_changes.rs
deuszx: Good point.

linera-execution/src/system.rs
deuszx: Ah, you're right. We actually had two `Option`s previously here - the outer one (return type) and inner one (`Some(Account { owner: None })`) and previously `Some(None)` would not refund at all.

linera-base/src/identifiers.rs
deuszx: The `::Chain` variant might not be so easy to remove b/c there are flow paths where the `Chain` is still a valid identifier. Right now I can see two ways to get rid of it:
1. Do not let chain own a balance.
2. Turn `::Chain` variant into a special case of address (`[u8; 32]`).

The first one isn't just a refactor (internal change) as it requires making some high-level design decisions around opening new chains etc.

I think the second option is the easier one and we should try to do it. We could hash a string like `"Linera Chain Address"` (note it cannot simply be `[0u8; 32]` b/c in Ed25519 a private key for _all zeros public key_ is [well-known](https://substrate.stackexchange.com/questions/982/why-does-the-all-0-public-key-have-a-known-private-key-in-sr25519-and-ed25519)).

linera-core/src/chain_worker/state/temporary_changes.rs
deuszx: Done in [Return chain balance](https://github.com/linera-io/linera-protocol/pull/3560/commits/f04abbc1ac87f1131d16cfb9c78252381a57b1c8)

examples/amm/src/contract.rs
deuszx: Done in [Replace unreachables with panics](https://github.com/linera-io/linera-protocol/pull/3560/commits/f8d5e33ec6a0610a99a2b4a9524e399cf37e60c4).

examples/fungible/src/contract.rs
deuszx: Done in [Replace unreachables with panics](https://github.com/linera-io/linera-protocol/pull/3560/commits/f8d5e33ec6a0610a99a2b4a9524e399cf37e60c4).

examples/native-fungible/src/contract.rs
deuszx: Done in [Replace unreachables with panics](https://github.com/linera-io/linera-protocol/pull/3560/commits/f8d5e33ec6a0610a99a2b4a9524e399cf37e60c4).

examples/non-fungible/src/contract.rs
deuszx: Done in [Replace unreachables with panics](https://github.com/linera-io/linera-protocol/pull/3560/commits/f8d5e33ec6a0610a99a2b4a9524e399cf37e60c4).

linera-execution/src/system.rs
deuszx: I agree it looks dangerous. So I tracked the places where I use `AccountOwner::Chain` as the default - there are two _(actually three but there are two instances where I use `::Chain` as `source` when constructing a `SystemMessage::Credit` so that's one for the purpose of this discussion)_:
1. In `send_refund` we [previously](https://github.com/linera-io/linera-protocol/blob/524c61a02c7a1b1fea62eec0174777cb37e72649/linera-execution/src/execution.rs#L404) were using `None` as `source` and [now](https://github.com/linera-io/linera-protocol/blob/9a67aed0aec636e947be960ccc3e766b4e5b6848/linera-execution/src/execution.rs#L402-L407) we use `AccountOwner::Chain`.
2. In `system.rs#execute_operation` we [previously](https://github.com/linera-io/linera-protocol/blob/524c61a02c7a1b1fea62eec0174777cb37e72649/linera-execution/src/system.rs#L422) were using `None` as `source` for the transfer and [now](https://github.com/linera-io/linera-protocol/blob/cleanup-account-owner-chain/linera-execution/src/system.rs#L420) we default to `::Chain`.

In both those cases `None` resolved to the "chain" already:
1. [main](https://github.com/linera-io/linera-protocol/blob/524c61a02c7a1b1fea62eec0174777cb37e72649/linera-execution/src/system.rs#L774-L778) vs [this PR](https://github.com/linera-io/linera-protocol/blob/cleanup-account-owner-chain/linera-execution/src/system.rs#L774-L777)
2. [main](https://github.com/linera-io/linera-protocol/blob/524c61a02c7a1b1fea62eec0174777cb37e72649/linera-execution/src/system.rs#L665-L668) vs [this PR](https://github.com/linera-io/linera-protocol/blob/9a67aed0aec636e947be960ccc3e766b4e5b6848/linera-execution/src/system.rs#L661-L664).

linera-base/src/identifiers.rs
ma2bd: > note it cannot simply be [0u8; 32] b/c in Ed25519 a private key for all zeros public key is [well-known](https://substrate.stackexchange.com/questions/982/why-does-the-all-0-public-key-have-a-known-private-key-in-sr25519-and-ed25519))

Addresses are not public keys so `[0u8; 32]` would work. I proposed `0x0` (`Reserved(0)`) in the internal doc.

ma2bd (approved):

PR #3568, opened 2025-03-13
Charge a fee to perform an HTTP request

### LINKED ISSUE #3564: Charge a fixed fee per HTTP request
# Motivation

Performing HTTP requests incurs in potential costs to the validators, as well as an impact to block latency. Therefore, these costs should be charged when executing the block.

# Proposal

Add a fee to charge every HTTP request performed by the application.

jvff:
## Motivation

Performing HTTP requests incurs in potential costs to the validators, as well as an impact to block latency. Therefore, these costs should be charged when executing the block.

## Proposal

Charge a fixed fee for every HTTP request performed.

## Test Plan

Added some unit tests to ensure that the HTTP requests are charged.

## Release Plan

      SDK soon, because this contains changes to the consensus critical `ResourceControlPolicy` and WIT interface.

## Links

- Closes #3564 


linera-execution/src/policy.rs
```diff
@@ -41,6 +41,8 @@ pub struct ResourceControlPolicy {
     pub message: Amount,
     /// The additional price for each byte in the argument of a user message.
     pub message_byte: Amount,
+    /// The price for a performing an HTTP request.
+    pub http_request: Amount,
```
deuszx, line 45:
```suggestion
    pub per_http_request_price: Amount,
```

We should also think about having a separate type for prices. And start calling it weights. See https://github.com/paritytech/polkadot-sdk/blob/master/substrate/frame/revive/src/weights.rs for reference.

linera-base/src/data_types.rs
```diff
@@ -284,6 +284,8 @@ pub struct Resources {
     pub message_size: u32,
     /// An increase in the amount of storage space.
     pub storage_size_delta: u32,
+    /// A number of HTTP requests to be performed.
+    pub http_requests: u32,
```
deuszx, line 288: Why do we need to track this?

linera-execution/src/runtime.rs
```diff
@@ -969,6 +969,7 @@ impl<UserInstance> BaseRuntime for SyncRuntimeInternal<UserInstance> {
                     })?
                     .recv_response()?
             };
+        self.resource_controller.track_http_request()?;
```
deuszx, line 972: If we have multiple HTTP requests within a single operation/proposal - won't this charge multiples? i.e. 1. First `perform_http_request` call will charge for 1 request
2. Second call will charge for 2
3. Third call will charge for 3

In total, instead of charging 3 multiples we would charge 6 🤔

jvff: No, there's a test for that in the unit tests (it includes three HTTP requests).

linera-base/src/data_types.rs
jvff: This type is used to calculate grants. Basically you configure an instance of this with whichever operations you expect to be executed when the message is received, and then the runtime calculates how much that would cost.

linera-execution/src/policy.rs
jvff: That breaks the convention (see the fields above). But I agree we could improve the names. Feel free to open an issue to track this.

bart-linera (approved):

deuszx (approved):

PR #3569, opened 2025-03-13
Add configurable permissions for making HTTP requests

### LINKED ISSUE #3531: Configurable permissions for performing HTTP requests
# Motivation

Shared chains may want to prevent applications from executing HTTP requests because it might negatively affect the chain's latency.

# Related Issues

- This is similar to #3526

bart-linera:
## Motivation

Chains may want to prevent applications from performing HTTP requests.

## Proposal

This adds a new field to ApplicationPermissions that controls which applications can make HTTP requests. If the calling application is not on the list, the call will fail with an UnauthorizedApplication error.

## Test Plan

Unit tests were added that exercise new functionality.

## Release Plan

## Links

- Closes #3531 

### DIFF
--- a/linera-execution/tests/contract_runtime_apis.rs
+++ b/linera-execution/tests/contract_runtime_apis.rs
@@ -907,7 +907,7 @@ async fn test_query_service(authorized_apps: Option<Vec<()>>) -> Result<(), Exec
         .expect("should register mock application");
 
     let call_service_as_oracle =
-        authorized_apps.map(|apps| apps.into_iter().map(|_| application_id).collect());
+        authorized_apps.map(|apps| apps.into_iter().map(|()| application_id).collect());
 
     view.system
         .application_permissions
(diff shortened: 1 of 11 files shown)


linera-execution/tests/contract_runtime_apis.rs
```diff
@@ -952,3 +954,81 @@ async fn test_query_service(authorized_apps: Option<Vec<()>>) -> Result<(), Exec
...
+    let app_desc_blob_id = application_description_blob.id();
+
+    let (application_id, application) = view
+        .register_mock_application_with(application_description, contract_blob, service_blob)
+        .await
+        .expect("should register mock application");
+
+    let make_http_requests = authorized_apps
+        .into_iter()
+        .map(|_| application_id)
```
jvff, line 990: Optional minor nit: we could use `()` here directly to make it fail to compile if we change the test parameter but forget to update it here.

```suggestion
        .map(|()| application_id)
```

jvff (approved):

jvff (approved):

PR #3570, opened 2025-03-13
Separate types for Application and ApplicationId

deuszx:
## Motivation

We want to fix and expand our notion of `Address` (currently `Owner`, `GenericApplicationId`, etc.) to different type of addresses (32-byte Linera/Solana, 20-byte EVM). In order to do that we need to prepare the code for the introduction of new variants.

## Proposal

Add a separate type for `ApplicationId(hash)` vs `Application` (a hash + type-safe ABI). Previously the presence of `A` type parameter in "address" enum made it more difficult to refactor. After this PR we have separate types for application's address and application with ABI (used mostly in tests and CLI now).

## Test Plan

CI should catch regressions.

## Release Plan

## Links

### DIFF
--- a/examples/ethereum-tracker/src/contract.rs
+++ b/examples/ethereum-tracker/src/contract.rs
@@ -80,7 +80,7 @@ impl EthereumTrackerContract {
     async fn read_initial(&mut self) {
         let request = async_graphql::Request::new("query { readInitialEvent }");
 
-        let application_id = self.runtime.application_id();
+        let application_id = self.runtime.application();
         let response = self.runtime.query_service(application_id, request);
 
         let async_graphql::Value::Object(data_object) = response.data else {
@@ -113,7 +113,7 @@ impl EthereumTrackerContract {
             r#"query {{ readTransferEvents(endBlock: {end_block}) }}"#
         ));
 
-        let application_id = self.runtime.application_id();
+        let application_id = self.runtime.application();
         let response = self.runtime.query_service(application_id, request);
 
         let async_graphql::Value::Object(data_object) = response.data else {
--- a/examples/amm/src/contract.rs
+++ b/examples/amm/src/contract.rs
@@ -8,7 +8,7 @@ mod state;
 use amm::{AmmAbi, Message, Operation, Parameters};
 use fungible::{Account, FungibleTokenAbi};
 use linera_sdk::{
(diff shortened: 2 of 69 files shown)


ma2bd:
We haven't really discussed what to do with ApplicationId yet. I was hoping we don't need to change this.

examples/ethereum-tracker/src/contract.rs
```diff
@@ -113,7 +113,7 @@ impl EthereumTrackerContract {
             r#"query {{ readTransferEvents(endBlock: {end_block}) }}"#
         ));
 
-        let application_id = self.runtime.application_id();
+        let application_id = self.runtime.application();
```
ma2bd, line 116: This PR ruins the field/variable naming convention in many places.

ma2bd (changes requested):
Thanks for the PR.

I'd rather keep `application_id: ApplicationId<A>` for the SDK: `Application` feels a bit vague and requires a massive refactoring of variable names -- including for our SDK users.

Why not use `UserApplicationId` for the (internal) use case without a type parameter?

deuszx:
> I'd rather keep `application_id: ApplicationId<A>` for the SDK: `Application` feels a bit vague and requires a massive refactoring of variable names -- including for our SDK users.
> 

In my mind `*Id` is just for identification and so the `ApplicationId<A: Abi>` whose purpose (and difference compared to `ApplicationId<()>`) was to convey the underlying application's ABI did not seem fit that as the additional `Abi` type provided information about the application that goes beyond its address.

> Why not use `UserApplicationId` for the (internal) use case without a type parameter?

What does the `User*` prefix stand for? An application without the typed ABI (`ApplicationId<()>`) becomes a `User*` one when we erase it (the ABI). The same application (the same underlying address) is either an _application_ (when the ABI is typed - `ApplicationId<A>`) and _user application_ (when it's not typed - `ApplicationId<()>`). I can use `UserApplicationId` for the untyped struct but this comment would be valid still.

deuszx:
Closing this in favor of https://github.com/linera-io/linera-protocol/pull/3582 which has far fewer changes (avoids client-facing renames).

examples/ethereum-tracker/src/contract.rs
deuszx: See https://github.com/linera-io/linera-protocol/pull/3582 for alternative.

PR #3577, opened 2025-03-15
Introduce one smart contract without GraphQL and the corresponding end-to-end test.

MathieuDutSik:
## Motivation

GraphQL is used all over the Linera code. This led to the question whether we can program a Linera application without GraphQL. This PR demonstrates that it is possible.

## Proposal

The following was done:
* Introduce an application `counter-no-graphql` that works without GraphQL but still uses the `linera-sdk`.
* The `ApplicationWrapper` has been restructured to separate between json queries and GraphQL queries.

This PR led to some further questions to address in next PRs:
* The `counter-no-graphql` application does not use GraphQL, but it still typed. We should also have a `counter_untyped` for which `execute_operation` takes a `Vec<u8>` and does not use the `linera-sdk` macros.
* Different serializations are being used in the code: `bcs::to_bytes` and `serde_json::to_vec`. No effort was made to unify this.
* The `raw_operation` function is returning some `CryptoHash` instead of a `Vec<u8>`. It was not clear how to obtain the result of the `execute_operation` from the resulting entry.
* The EVM application are untyped. Moving to the `raw_application` seems premature.

## Test Plan

The test running without GraphQL has been inserted.

## Release Plan

This should be included in next TestNet release.

## Links

None.

### DIFF
--- a/examples/counter-no-graphql/src/service.rs
+++ b/examples/counter-no-graphql/src/service.rs
@@ -1,0 +1,79 @@
+// Copyright (c) Zefchain Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+#![cfg_attr(target_arch = "wasm32", no_main)]
+
+mod state;
+
+use std::sync::Arc;
+
+use counter_no_graphql::{CounterOperation, CounterRequest};
+use linera_sdk::{linera_base_types::WithServiceAbi, views::View, Service, ServiceRuntime};
+
+use self::state::CounterState;
+
+pub struct CounterService {
+    state: CounterState,
+    runtime: Arc<ServiceRuntime<Self>>,
+}
+
+linera_sdk::service!(CounterService);
+
+impl WithServiceAbi for CounterService {
+    type Abi = counter_no_graphql::CounterNoGraphQlAbi;
+}
+
+impl Service for CounterService {
+    type Parameters = ();
+
+    async fn new(runtime: ServiceRuntime<Self>) -> Self {
+        let state = CounterState::load(runtime.root_view_storage_context())
+            .await
+            .expect("Failed to load state");
+        CounterService {
+            state,
+            runtime: Arc::new(runtime),
+        }
+    }
+
+    async fn handle_query(&self, request: CounterRequest) -> u64 {
+        match request {
+            CounterRequest::Query => *self.state.value.get(),
+            CounterRequest::Increment(value) => {
+                let operation = CounterOperation::Increment(value);
+                self.runtime.schedule_operation(&operation);
+                0
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use std::sync::Arc;
+
+    use futures::FutureExt as _;
+    use linera_sdk::{util::BlockingWait, views::View, Service, ServiceRuntime};
+
+    use super::{CounterRequest, CounterService, CounterState};
+
+    #[test]
+    fn query() {
+        let value = 61_098_721_u64;
+        let runtime = Arc::new(ServiceRuntime::<CounterService>::new());
+        let mut state = CounterState::load(runtime.root_view_storage_context())
+            .blocking_wait()
+            .expect("Failed to read from mock key value store");
+        state.value.set(value);
+
+        let service = CounterService { state, runtime };
+        let request = CounterRequest::Query;
+
+        let response = service
+            .handle_query(request)
+            .now_or_never()
+            .expect("Query should not await anything");
+
+        assert_eq!(response, value)
+    }
+}
--- a/linera-service/src/cli_wrappers/wallet.rs
+++ b/linera-service/src/cli_wrappers/wallet.rs
@@ -1376,39 +1376,37 @@ pub struct ApplicationWrapper<A> {
 }
 
 impl<A> ApplicationWrapper<A> {
-    pub async fn raw_query(&self, query: impl AsRef<str>) -> Result<Value> {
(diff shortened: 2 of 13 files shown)


ma2bd:
> This is importance since GraphQL is a very peripheral API in finance or blockchain world.

This is subjective and hardly the point here. What matters is to use the right tool for the job. GraphQL happens to a bit heavy at times for on-chain applications, especially those in Solidity.

MathieuDutSik:
> > This is importance since GraphQL is a very peripheral API in finance or blockchain world.
> 
> This is subjective and hardly the point here. What matters is to use the right tool for the job. GraphQL happens to a bit heavy at times for on-chain applications, especially those in Solidity.

Ok, rewritten.

examples/counter-no-graphql/src/service.rs
```diff
@@ -0,0 +1,65 @@
...
+    type Parameters = ();
+
+    async fn new(runtime: ServiceRuntime<Self>) -> Self {
+        let state = CounterState::load(runtime.root_view_storage_context())
+            .await
+            .expect("Failed to load state");
+        CounterService { state }
+    }
+
+    async fn handle_query(&self, _request: ()) -> u64 {
```
ma2bd, line 32: We need a demo app that actually processes queries: either to answer queries or to support block building (aka mutations in GraphQL speak).

linera-service/src/node_service.rs
```diff
@@ -923,13 +928,19 @@ where
...
+            axum::routing::get(util::graphiql).post(Self::raw_application_handler);
 
         let app = Router::new()
             .route("/", index_handler)
             .route(
                 "/chains/:chain_id/applications/:application_id",
                 application_handler,
             )
+            .route(
+                "/chains/:chain_id/raw_applications/:application_id",
```
ma2bd, line 941: Why do we need this in the first place?

ma2bd (changes requested):
It should be easier to do the right thing after https://github.com/linera-io/linera-protocol/pull/3581

examples/counter-no-graphql/src/service.rs
MathieuDutSik: Ok, I extended the request type which is now an enum.
The first case just returns a value, the second test creates an operation with `schedule_operation`.

linera-service/src/cli_wrappers/wallet.rs
```diff
@@ -1376,39 +1376,37 @@ pub struct ApplicationWrapper<A> {
...
 impl<A> ApplicationWrapper<A> {
-    pub async fn raw_query(&self, query: impl AsRef<str>) -> Result<Value> {
+    pub async fn json_query(&self, query: impl AsRef<str>) -> Result<Value> {
+        let query = query.as_ref();
+        let value = self.raw_query(json!({ "query": query })).await?;
+        Ok(value["data"].clone())
+    }
+
+    pub async fn raw_query<T: Serialize>(&self, query: T) -> Result<Value> {
         const MAX_RETRIES: usize = 5;
```
ma2bd, line 1387: Can we serialize only once?
I would add this line:
`let query = serde_json::to_string(&query)?;`

linera-service/src/cli_wrappers/wallet.rs
```diff
@@ -1376,39 +1376,37 @@ pub struct ApplicationWrapper<A> {
 }
 
 impl<A> ApplicationWrapper<A> {
-    pub async fn raw_query(&self, query: impl AsRef<str>) -> Result<Value> {
+    pub async fn json_query(&self, query: impl AsRef<str>) -> Result<Value> {
```
ma2bd, line 1379: `json` is not the right concept. I would suggest something like `run_graphql_query` or `send_graphql_query`

linera-service/src/cli_wrappers/wallet.rs
```diff
@@ -1376,39 +1376,37 @@ pub struct ApplicationWrapper<A> {
...
+        let query = query.as_ref();
+        let value = self.raw_query(json!({ "query": query })).await?;
+        Ok(value["data"].clone())
+    }
+
+    pub async fn raw_query<T: Serialize>(&self, query: T) -> Result<Value> {
```
ma2bd, line 1385: _This_ is the JSON one: `run_json_query` or `send_json_query`.

examples/counter-no-graphql/src/service.rs
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    async fn handle_query(&self, request: CounterRequest) -> u64 {
+        match request {
+            CounterRequest::Query => *self.state.value.get(),
+            CounterRequest::Increment(value) => {
+                let operation = CounterOperation::Increment(value);
+                self.runtime.schedule_operation(&operation);
+                0
+            }
```
ma2bd, line 46: nice

ma2bd (approved):
two comments but otherwise LGTM

linera-service/src/cli_wrappers/wallet.rs
MathieuDutSik: Ok, I was unsure about the naming.

linera-service/src/cli_wrappers/wallet.rs
MathieuDutSik: Ok, I was unsure about the naming.

linera-service/src/cli_wrappers/wallet.rs
MathieuDutSik: Ok, changed.

PR #3578, opened 2025-03-15
feat: Implement limit parameter for blocks page

VolodymyrBg:
## Motivation

This PR implements a missing feature in the explorer UI that was marked with a TODO comment. The blocks page previously had a hardcoded limit of 20 blocks, but there was no way for users to adjust this limit through the UI.

## Proposal

The changes implement the TODO comment in linera-explorer/src/lib.rs by:
- Adding a select input in the Blocks.vue component that allows users to choose how many blocks to display (10, 20, 50, or 100)
- Updating the blocks function in lib.rs to use the limit parameter from the URL
- Updating the routing function to pass the limit parameter from the URL to the blocks function
- Adding logic to the Blocks.vue component to show the currently selected limit from the URL
These changes improve the user experience by giving users control over how many blocks they want to view at once.

## Test Plan

I manually tested the changes by:
- Loading the blocks page with the default limit (20)
- Changing the limit using the select input and verifying that the correct number of blocks is displayed
- Refreshing the page and verifying that the selected limit is preserved in the URL
- Navigating to other pages and back to the blocks page to ensure the limit is maintained

### DIFF
--- a/linera-explorer/src/components/Block.vue
+++ b/linera-explorer/src/components/Block.vue
@@ -1,125 +1,89 @@
 <script setup lang="ts">
 import { HashedConfirmedBlock } from '../../gql/service'
-import Json from './Json.vue'
-import Op from './Op.vue'
+import { onMounted, ref } from 'vue'
 
-defineProps<{block: HashedConfirmedBlock, title: string}>()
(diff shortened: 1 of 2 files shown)


ma2bd:
@VolodymyrBg Thanks for the PR. We're going to have a look

PR #3581, opened 2025-03-16
Simplify service requests

ma2bd:
## Motivation

After #3217, we do not need to introspect service queries to applications.

## Proposal

* Minimize assumptions about application APIs. After this PR, we only assume UTF8 bytes as input and JSON as output.
* Simplify requests handlers
* Somehow this revealed silently failing queries in our README files. Also fix that.

Note the assumption on JSON outputs could be relaxed if we stopped returning the hash of the certified block and return the query response instead. However, this is a breaking change so this should be a separate PR.

## Test Plan

CI

### DIFF
--- a/linera-service/src/node_service.rs
+++ b/linera-service/src/node_service.rs
@@ -4,8 +4,8 @@
 use std::{borrow::Cow, iter, net::SocketAddr, num::NonZeroU16, sync::Arc};
 
 use async_graphql::{
-    futures_util::Stream, resolver_utils::ContainerType, Error, MergedObject, OutputType, Request,
-    ScalarType, Schema, ServerError, SimpleObject, Subscription,
+    futures_util::Stream, resolver_utils::ContainerType, Error, MergedObject, OutputType,
+    ScalarType, Schema, SimpleObject, Subscription,
 };
 use async_graphql_axum::{GraphQLRequest, GraphQLResponse, GraphQLSubscription};
 use axum::{extract::Path, http::StatusCode, response, response::IntoResponse, Extension, Router};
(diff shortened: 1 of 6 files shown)


linera-service/src/node_service.rs
```diff
@@ -950,53 +886,24 @@ where
...
-                .iter()
-                .map(|e| e.to_string())
-                .collect();
-            return Err(NodeServiceError::ApplicationServiceError { errors });
+        if operations.is_empty() {
+            return Ok(response);
         }
-        trace!("Operations: {operations:?}");
 
+        trace!("Query requested a new block with operations: {operations:?}");
```
afck, line 906: Maybe there should be a way for the user to tell the client that it does _not_ want to create a new block, regardless of what the application returns.

linera-service/src/node_service.rs
```diff
@@ -1018,17 +925,17 @@ where
             })?;
             util::wait_for_next_round(&mut stream, timeout).await;
         };
-        Ok(async_graphql::Response::new(hash.to_value()))
+        let response = async_graphql::Response::new(hash.to_value());
+        Ok(serde_json::to_vec(&response)?)
```
deuszx, line 929: I'm not sure if this is really an improvement - sure the type is `Vec<u8>` but it is a serialized JSON. If it can't be anything else I think the previous type was better b/c it more precisely defined how to handle the data. Right now callers have to read the docs, previously it was in the type.

afck: Actually, wouldn't it be more consistent to not return the block hash here, and instead return the `response` from the `QueryOutcome`, just like in line 903 above?

ma2bd: please guys read the summary

bart-linera (approved):

linera-service/src/node_service.rs
ma2bd: Yes we're certainly not going to auto-sign block proposals like this in the future but instead properly integrate with wallets.

Issue #3587, opened 2025-03-18
Print right url info for faucet service


kikakkz:
**Description**
It always print `localhost` but listen on `0.0.0.0` currently.

**Screenshot**
![Image](https://github.com/user-attachments/assets/c68dea1b-07cd-4cfc-ae20-68ef326b7723)

kikakkz:
> Technically, localhost (aka 127.0.0.1) is a subset of the addresses designated by 0.0.0.0

Yeah. Here I meant that we can give (maybe) two urls to users: `http://localhost:8080` and `http://$LAN_IP:8080`.

I create a draft PR here: https://github.com/linera-io/linera-protocol/pull/3588

PR #3588, opened 2025-03-18
Print right url info for faucet service

kikakkz:
## Motivation

Faucet always print `http://localhost:8080` to users.

## Proposal

I think we can provide urls with localhost and $LAN_IP to users here.

## Test Plan

CI

## Release Plan

## Links

https://github.com/linera-io/linera-protocol/issues/3587


kikakkz:
Just create to explain issue #3587 so I close it directly.

PR #3592, opened 2025-03-18
`UserApplicationId` is `MultiAddress`

deuszx:
## Proposal

Replace usage of `UserApplicationId` with `MultiAddress`. `UserApplicationId` was an address of an application that internally was `CryptoHash` and now we can identify all entities (accounts, applications) in the system with `MultiAddress` type.

## Release Plan

    - be released in a new SDK,
    - be released in a new SDK,
      SDK soon.

## Links


ma2bd (commented):
It seems like we can replace `UserApplicationId` entirely in the SDK (using `Address` and `ApplicationId`) but I'm not sure about the representations of an untyped application ID inside the Linera protocol.

Also technically `UserApplicationId` (or however we call it in the end) will probably be a fixed size hash (possibly 20-bytes) ==> we can spare the tag in binary encoding.

linera-base/src/data_types.rs
```diff
@@ -718,34 +718,34 @@ pub struct ApplicationPermissions {
     /// If it is `Some`, only operations from the specified applications are allowed, and
     /// no system operations.
     #[debug(skip_if = Option::is_none)]
-    pub execute_operations: Option<Vec<UserApplicationId>>,
+    pub execute_operations: Option<Vec<MultiAddress>>,
```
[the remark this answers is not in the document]
deuszx: Why? I think we should stop thinking about user vs system applications' addresses and it's all just an address. So here, it makes sense that there's a set of addresses that are allowed to `execute_operations` (etc).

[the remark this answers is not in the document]
afck: …and we could even slightly change the meaning of this one:
You are allowed to execute an operation if
* you yourself are in this list, _or_
* the operation belongs to an application in the list.

linera-service/src/cli_wrappers/wallet.rs
```diff
@@ -412,7 +408,7 @@ impl ClientWrapper {
         module_id: &ModuleId<Abi, Parameters, InstantiationArgument>,
         parameters: &Parameters,
         argument: &InstantiationArgument,
-        required_application_ids: &[UserApplicationId],
+        required_application_ids: &[MultiAddress],
```
afck, line 411: Maybe this should just be called `required_applications` now everywhere.

In general, a variable of type `MultiAddress` should probably be called `app[lication]` or `app[lication]_addr[ess]`.

afck (approved):
No blockers from my side, once the remaining discussions are resolved.

linera-base/src/data_types.rs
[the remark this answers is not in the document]
ma2bd: ok this specific one works but some other places in the protocol will make only sense for an app id

linera-base/src/data_types.rs
```diff
@@ -718,34 +718,34 @@ pub struct ApplicationPermissions {
...
     /// At least one operation or incoming message from each of these applications must occur in
     /// every block.
     #[graphql(default)]
     #[debug(skip_if = Vec::is_empty)]
-    pub mandatory_applications: Vec<UserApplicationId>,
+    pub mandatory_applications: Vec<MultiAddress>,
```
ma2bd, line 726: ok so afaict this one makes no sense for user addresses

deuszx: There's no such thing as "user addresses" anymore. It's just address (both for accounts and applications - system and user).

linera-service/src/cli_wrappers/wallet.rs
deuszx: Yes, there's a lot of renames still to do.

ma2bd: This is another instance where using addresses looks imprecise.

examples/gen-nft/src/lib.rs
```diff
@@ -122,7 +122,7 @@ impl Display for TokenId {
 impl Nft {
     pub fn create_token_id(
         chain_id: &ChainId,
-        application_id: &UserApplicationId,
+        application_id: &MultiAddress,
```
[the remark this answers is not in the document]
deuszx: There's no need to know the Abi of this application here.

PR #3595, opened 2025-03-18
Remove `RawExecutionOutcome` and `RawOutgoingMessage`.

afck:
## Motivation

As of https://github.com/linera-io/linera-protocol/pull/3513, `RawExecutionOutcome` and `RawOutgoingMessage` are not really needed anymore.

## Proposal

Remove them.

In the first commit, their usage in production is replaced by directly creating `OutgoingMessage`s. The `linera-execution` tests still passed, so the behavior didn't change.

In the second commit, their usage is also removed from the tests.

## Test Plan

Existing tests were updated.

## Release Plan

## Links

### DIFF
--- a/linera-execution/src/execution.rs
+++ b/linera-execution/src/execution.rs
@@ -29,9 +29,9 @@ use super::{runtime::ServiceRuntimeRequest, ExecutionRequest};
 use crate::{
     resources::ResourceController, system::SystemExecutionStateView, ContractSyncRuntime,
     ExecutionError, ExecutionRuntimeConfig, ExecutionRuntimeContext, Message, MessageContext,
-    MessageKind, Operation, OperationContext, Query, QueryContext, QueryOutcome,
-    RawExecutionOutcome, RawOutgoingMessage, ServiceSyncRuntime, SystemMessage, TransactionTracker,
-    UserApplicationDescription, UserApplicationId,
+    MessageKind, Operation, OperationContext, OutgoingMessage, Query, QueryContext, QueryOutcome,
+    ServiceSyncRuntime, SystemMessage, TransactionTracker, UserApplicationDescription,
+    UserApplicationId,
 };
 
 /// A view accessing the execution state of a chain.
@@ -317,7 +317,7 @@ where
         match message {
             Message::System(message) => {
                 let outcome = self.system.execute_message(context, message).await?;
-                txn_tracker.add_system_outcome(outcome)?;
+                txn_tracker.add_outgoing_messages(outcome)?;
             }
             Message::User {
                 application_id,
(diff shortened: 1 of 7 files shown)


linera-execution/src/transaction_tracker.rs
```diff
@@ -149,6 +98,16 @@ impl TransactionTracker {
...
     }
 
+    pub fn add_outgoing_messages(
+        &mut self,
+        messages: impl IntoIterator<Item = OutgoingMessage>,
+    ) -> Result<(), ArithmeticError> {
+        for message in messages {
+            self.add_outgoing_message(message)?;
+        }
+        Ok(())
```
deuszx, line 108: Could also do `checked_add(messages.len())` here and `outgoing_messages.extend(messages)`?

linera-execution/src/system.rs
```diff
@@ -486,17 +477,13 @@ where
...
-                        id: context.chain_id,
-                        subscription,
-                    },
+                let message = SystemMessage::Subscribe {
+                    id: context.chain_id,
+                    subscription,
                 };
-                outcome.messages.push(message);
+                txn_tracker.add_outgoing_message(
+                    OutgoingMessage::new(chain_id, message).with_kind(MessageKind::Protected),
```
deuszx, line 485: Are some messages `Protected` by design? Or should they be? I can imagine `Subscribe` could be one.

linera-execution/src/execution.rs
```diff
@@ -347,41 +347,14 @@ where
...
-                    kind: MessageKind::Bouncing,
-                    message: bytes,
-                });
-                txn_tracker.add_user_outcome(application_id, outcome)?;
-            }
-        }
+        txn_tracker.add_outgoing_message(OutgoingMessage {
+            destination: Destination::Recipient(context.message_id.chain_id),
+            authenticated_signer: context.authenticated_signer,
+            refund_grant_to: context.refund_grant_to.filter(|_| !grant.is_zero()),
```
deuszx, line 353: I checked that we ignore the `refund_grant_to` if `grant == 0`.

deuszx (approved):

linera-execution/src/execution.rs
afck: Yes, but it doesn't hurt to already omit it from the messages and save a few bytes.

linera-execution/src/transaction_tracker.rs
afck: Then we'd need to make `messages` a `Vec` or `impl ExactSizeIterator` which would make it more cumbersome to use. And the performance improvement should be negligible.

Issue #3532, opened 2025-03-10
Update `resource-control-policy` command with missing configuration parameters


jvff:
# Motivation

The `linera resource-control-policy` command allows changing some options in the `ResourceControlPolicy` specified in an `Epoch`. However, some configuration options are missing from the command.

# Proposal

Tweak the implementation to do something like:

```rust
let existing_policy = // From committee
let new_policy = ResourceControlPolicy {
    policy: policy_option.unwrap_or(existing_policy.policy),
    // etc.
};
```

So that new fields added to `ResourceControlPolicy` force new options to be written for the CLI.

PR #3597, opened 2025-03-19
Limit the size of oracle responses

### LINKED ISSUE #3563: Limit the size of oracle responses
# Motivation

Oracle responses are stored in blocks, so they should not be allowed to be very large.

# Proposal

Add a `maximum_oracle_response_bytes` field to `ResourceControlPolicy`, and enforce it whenever an oracle response is recorded.

jvff:
## Motivation

Services can be called as oracles, and the responses they provide must be bounded because they are stored in blocks.

## Proposal

Implement a configurable limit to the size of oracle responses, and enforce that limit in contract HTTP requests and service-as-oracle responses.

To distinguish between contracts and services in compile time, an extension trait was used to allow minimal change to the `BaseRuntime` implementation.

## Test Plan

Unit test cases were added to check if the limits were being respected.

## Release Plan

- These changes follow the usual release cycle, because services-as-oracles is still feature gated by `unstable-oracles`.

## Links

- Closes #3563


linera-execution/src/policy.rs
```diff
@@ -67,6 +67,8 @@ pub struct ResourceControlPolicy {
     pub maximum_bytes_read_per_block: u64,
     /// The maximum data to write per block
     pub maximum_bytes_written_per_block: u64,
+    /// The maximum size in bytes of an oracle response.
+    pub maximum_oracle_response_bytes: u64,
```
afck, line 71: Let's make sure all resource policy options can be set via the command line. (See https://github.com/linera-io/linera-protocol/pull/3598.)

jvff: I was going to do this in a separate PR that would also catch these issues: #3532.

afck: (cc @MathieuDutSik)

linera-chain/src/test/http_server.rs
```diff
@@ -0,0 +1,57 @@
...
+use std::{future::IntoFuture, net::Ipv4Addr};
+
+use axum::Router;
+use futures::FutureExt as _;
+use tokio::{net::TcpListener, sync::oneshot};
+
+/// A handle to a running HTTP server.
+///
+/// The server is gracefully shutdown when this handle is dropped.
+pub struct HttpServer {
```
deuszx, line 15: I vageuly remember seeing this already in one of your previous PRs 🤔

jvff: Yes, in #3509. I'll resume work on that one soon, and use the type created here instead.

linera-execution/src/resources.rs
```diff
@@ -321,6 +321,19 @@ where
 
         Ok(())
     }
+
+    /// Tracks the size of a response produced by an oracle.
+    pub(crate) fn track_service_oracle_response(
```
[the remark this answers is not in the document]
jvff: No. I think we could, but I'm not sure if it makes sense. My reasoning is that we have a limit and a fixed cost per call. That's like an upper bound on the cost. Tracking the response size could make it more cost efficient, but would require extra tweaking and might not be worth it. I opened #3600 to track this.

linera-execution/src/policy.rs
jvff: I'll wait for #3598 to be merged first, than I can rebase and fix it here as well.

jvff: Done, thanks!

linera-execution/src/resources.rs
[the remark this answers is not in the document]
jvff: Sure, I guess we can do something like that. But I think that's out of the scope of this PR.

linera-chain/src/test/http_server.rs
```diff
@@ -0,0 +1,57 @@
...
+    port: u16,
+    _shutdown_sender: oneshot::Sender<()>,
+}
+
+impl HttpServer {
+    /// Spawns a task with an HTTP server serving the routes defined by the [`Router`].
+    ///
+    /// Returns a [`HttpServer`] handle to keep the server running in the background.
```
afck, line 23:
```suggestion
    /// Returns an [`HttpServer`] handle to keep the server running in the background.
```

linera-chain/src/test/http_server.rs
```diff
@@ -0,0 +1,57 @@
...
+    pub async fn start(router: Router) -> anyhow::Result<Self> {
+        let (shutdown_sender, shutdown_receiver) = oneshot::channel();
+        let shutdown_signal = shutdown_receiver.map(|_| ());
```
afck, line 26: Isn't the type already `()`? Why the `map`?

linera-execution/src/runtime.rs
```diff
@@ -918,6 +926,25 @@ impl<UserInstance> BaseRuntime for SyncRuntimeHandle<UserInstance> {
     }
 }
 
+/// An extension trait to determine in compile time the different behaviors between contract and
+/// services in the implementation of [`BaseRuntime`].
+trait ContractAndServiceBaseRuntimeDifferences {
```
afck, line 931: Maybe just `ContractOrServiceRuntime`?
(One could argue that a trait is always about the differences between the different implementations.)

afck (approved):

linera-execution/src/runtime.rs
jvff: Good point, thanks!

linera-chain/src/test/http_server.rs
jvff: The `oneshot::Receiver<()>` [implements](https://docs.rs/oneshot/latest/oneshot/struct.Receiver.html#impl-Future-for-Receiver%3CT%3E) `Future<Output = Result<(), oneshot::RecvError>>`, so this just treats the error caused by a dropped sender as the shutdown signal as well.

PR #2500, opened 2024-09-16
Improvements to the LRU cache

MathieuDutSik:
## Motivation

The LRU is currently caching only the `read_value` operations. This leaves some optimizations away.

## Proposal

The following was done:
* The `map` was renamed as `map_value`.
* A map for find is introduced keeping track of the `find_keys_by_prefix` and `find_key_values_by_prefix`.
* For the `find_keys_by_prefix` we keep the map as suffix closed and we use the existing entries to get some value. So, if we already computed for prefix `[0]` then we can deduce the result for `[0,2]`.
* Similarly for `find_key_values_by_prefix`.
* Write batch tries to update as much as possible the cache.
* The `max_cache_size` which before was a maximum on the number of entries becomes a maximum on the number of bytes in the cache (and is expanded from 1000 to 10000000).
* The maximum number of entries in the cache is set to 1000.

Measurement setup:
* Done with the `test_wasm_end_to_end_amm` on `remote-net`.
* Storage used is `ScyllaDb` which has LRU caching.
* Code compiled in `--release` which takes about 1H to compile.
* Done 10 times with the 1st entry discarded.
* Scripting available at https://github.com/MathieuDutSik/linera_parsing_tools 

Averaged results:
* **Cache hit**. Old code 82.9%, new code: 84.8% for values (_contains_key_ and _read_value_), 91% for finds (_find_keys_by_prefix_ and _find_key_values_by_prefix_).
* **read_value_bytes**. 0.0276ms (Old code) vs 0.019 ms (New code).
* **contains_key**: 0.25 ms (Old code) vs 0.076 ms (New code)
* **find_keys_by_prefix**: 1.17 ms (Old code) vs 0.087 ms (New code)
* **find_key_values_by_prefix**: 1.08 ms (Old code) vs 0.13 ms (New code)
* **write_batch**: 2.22 ms (Old code) vs 2.52 ms (New code)
* **total runtime**: 16.63 second (Old code) vs 15.83 seconds (New code)

## Test Plan

The CI.

## Release Plan

Not relevant.

## Links

- [reviewer checklist](https://github.com/linera-io/linera-protocol/blob/main/CONTRIBUTING.md#reviewer-checklist)

PR #3599, opened 2025-03-19
Introduce the caching of `contains_key` operations.

MathieuDutSik:
## Motivation

The LRU caching is very minimal and can be extended.

## Proposal

The PR #2500 implements several improvements to the LRU caching,
but it proved impossible to review. Splitting the changes more atomically
might be mergeable into main.

The following was done:
* The `Option<Vec<u8>>` is replaced by a `CacheEntry` that encodes the status if missing.
* The lifetime is removed from the `fn query`, but that does not matter since it was cloned just after.
* The cache success/fault is split with read_value/contains_key.

## Test Plan

The CI.

## Release Plan

Normal release plan.

## Links

None.

### DIFF
--- a/linera-views/src/backends/lru_caching.rs
+++ b/linera-views/src/backends/lru_caching.rs
@@ -178,7 +178,9 @@ impl LruPrefixCache {
     /// Returns `Some(true)` or `Some(false)` if we know that the entry does or does not
     /// exist in the database. Returns `None` if that information is not in the cache.
     pub fn query_contains_key(&self, key: &[u8]) -> Option<bool> {
-        self.map.get(key).map(|entry| !matches!(entry, CacheEntry::DoesNotExist))
+        self.map
+            .get(key)
+            .map(|entry| !matches!(entry, CacheEntry::DoesNotExist))
     }
 }
 
--- a/linera-views/src/backends/lru_caching.rs
+++ b/linera-views/src/backends/lru_caching.rs
@@ -27,28 +27,96 @@ use crate::{
 use crate::{memory::MemoryStore, store::TestKeyValueStore};
 
 #[cfg(with_metrics)]
-/// The total number of cache faults
-static NUM_CACHE_FAULT: LazyLock<IntCounterVec> =
-    LazyLock::new(|| register_int_counter_vec("num_cache_fault", "Number of cache faults", &[]));
+/// The total number of cache read value faults
+static NUM_CACHE_READ_VALUE_FAULT: LazyLock<IntCounterVec> = LazyLock::new(|| {
+    register_int_counter_vec(
+        "num_cache_read_value_fault",
+        "Number of cache read value faults",
+        &[],
+    )
+});
 
 #[cfg(with_metrics)]
-/// The total number of cache successes
-static NUM_CACHE_SUCCESS: LazyLock<IntCounterVec> =
-    LazyLock::new(|| register_int_counter_vec("num_cache_success", "Number of cache success", &[]));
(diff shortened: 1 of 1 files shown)


linera-views/src/backends/lru_caching.rs
```diff
@@ -27,28 +27,96 @@ use crate::{
...
+    )
+});
+
+enum CacheEntry {
+    DoesNotExist,
+    Exists,
+    Value(Vec<u8>),
+}
+
+impl CacheEntry {
```
ma2bd, line 75: nit: To me, these functions are removing the benefits of using an enum `CacheEntry`. I would try to inline them.

ma2bd:
TBH there's almost too much stuff going on in this PR already.

linera-views/src/backends/lru_caching.rs
MathieuDutSik: Ok, I inline the functions that are used only once.

linera-views/src/backends/lru_caching.rs
```diff
@@ -27,28 +27,74 @@ use crate::{
...
+    fn get_contains_key(&self) -> bool {
```
afck, line 76: One of the uses is negated anyway, so maybe `matches!(entry, CacheEntry::DoesNotExist)` (or even `entry == CacheEntry::DoesNotExist`) would be easier to read, and we don't need this method either.

linera-views/src/backends/lru_caching.rs
```diff
@@ -107,9 +171,21 @@ impl<'a> LruPrefixCache {
         }
     }
 
-    /// Gets the entry from the key.
-    pub fn query(&'a self, key: &'a [u8]) -> Option<&'a Option<Vec<u8>>> {
-        self.map.get(key)
+    /// Gets the read_value_bytes from the key.
```
afck, line 174:
```suggestion
    /// Returns the cached value, or `Some(None)` if the entry does not exist in the
    /// database. If `None` is returned, the entry might exist in the database but is
    /// not in the cache.
```

linera-views/src/backends/lru_caching.rs
```diff
@@ -107,9 +171,21 @@ impl<'a> LruPrefixCache {
...
+            None => None,
+            Some(entry) => match entry {
+                CacheEntry::DoesNotExist => Some(None),
+                CacheEntry::Exists => None,
+                CacheEntry::Value(vec) => Some(Some(vec.clone())),
+            },
+        }
+    }
+
+    /// Gets the contains_key from the key.
```
afck, line 186:
```suggestion
    /// Returns `Some(true)` or `Some(false)` if we know that the entry does or does not
    /// exist in the database. Returns `None` if that information is not in the cache.
```

linera-views/src/backends/lru_caching.rs
MathieuDutSik: Thank you, done.

linera-views/src/backends/lru_caching.rs
MathieuDutSik: Thank you!

linera-views/src/backends/lru_caching.rs
MathieuDutSik: Thank you!

afck (approved):
Looks good to me. But I think @ma2bd wasn't sure yet whether we want to cache `contains_key` `false` results at all.

ma2bd (approved):

PR #3611, opened 2025-03-20
(nit) add symlink to make binary formats more discoverable

ma2bd:
## Motivation

Tracking binary formats is important but the corresponding file is buried in the directory tree and looks like just another test snapshot.

## Proposal

* Create a symlink to make the file more visible

This more stable link could also be used later for `serde_generate`

## Test Plan

Nothing to do


Twey (approved):

Ehsanbad:
Like

PR #3612, opened 2025-03-20
Refactor `ChainStateView::execute_block`.

afck:
## Motivation

`ChainStateView::execute_block` has many local variables that together represent the intermediate state of a block execution.

## Proposal

Make the intermediate state explicit as a `linera_execution::BlockExecutor`.

## Test Plan

Only refactoring.

## Release Plan

## Links

### DIFF
--- a/linera-chain/src/chain.rs
+++ b/linera-chain/src/chain.rs
@@ -1,12 +1,9 @@
 // Copyright (c) Zefchain Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
+use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
 #[cfg(with_metrics)]
 use std::sync::LazyLock;
-use std::{
-    collections::{BTreeMap, BTreeSet, HashMap, HashSet},
-    sync::Arc,
-};
 
 use async_graphql::SimpleObject;
 use futures::stream::{self, StreamExt, TryStreamExt};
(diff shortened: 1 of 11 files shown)


linera-execution/src/block_executor.rs
```diff
@@ -0,0 +1,328 @@
+// Copyright (c) Zefchain Labs, Inc.
```
deuszx, line 1: It'd be awesome if we could have a rustdoc here explaining what's the idea for `BlockExecutor`.

linera-execution/src/block_executor.rs
```diff
@@ -0,0 +1,328 @@
...
+);
+
+impl<'de> BcsHashable<'de> for OperationResult {}
+
+doc_scalar!(
+    OperationResult,
+    "The execution result of a single operation."
+);
+
+pub struct BlockExecutor<'a, C> {
```
deuszx, line 37: Or here.

linera-execution/src/block_executor.rs
```diff
@@ -0,0 +1,328 @@
...
+    pub resource_controller: ResourceController<Option<Owner>>,
+    replaying_oracle_responses: Option<vec::IntoIter<Vec<OracleResponse>>>,
+    next_message_index: u32,
+    next_application_index: u32,
+    local_time: Timestamp,
+    pub oracle_responses: Vec<Vec<OracleResponse>>,
+    pub events: Vec<Vec<Event>>,
+    pub blobs: Vec<Vec<Blob>>,
+    pub messages: Vec<Vec<OutgoingMessage>>,
+    pub operation_results: Vec<OperationResult>,
```
deuszx, line 48: If these are supposed to be set only by the `BlockExecutor` but only accessed (read) from everywhere else I'd drop the `pub` and add the getters. Otherwise, if you have `mut block_executor` instance you could push data that is not supposed to be here.

linera-execution/src/block_executor.rs
```diff
@@ -0,0 +1,329 @@
...
+    pub fn new(
+        state: &'a mut ExecutionStateView<C>,
+        account: Option<Owner>,
+        replaying_oracle_responses: Option<Vec<Vec<OracleResponse>>>,
+        local_time: Timestamp,
+    ) -> Result<Self, ExecutionError> {
+        let (_, committee) = state
+            .system
+            .current_committee()
+            .ok_or_else(|| ExecutionError::InactiveChain)?;
```
deuszx, line 67: I we don't merge `new` and `track_block` I'd extract this out and make `new` non-fallible. Nothing else here can fail and we won't need to add `with_execution_context` on the call site.

linera-chain/src/chain.rs
```diff
@@ -849,114 +858,71 @@ where
...
 
         #[cfg(with_metrics)]
-        Self::track_block_metrics(&resource_controller.tracker);
+        Self::track_block_metrics(&_resource_controller.tracker);
+
+        let txn_count = block.incoming_bundles.len() + block.operations.len();
+        assert_eq!(messages.len(), txn_count);
+        assert_eq!(events.len(), txn_count);
+        assert_eq!(oracle_responses.len(), txn_count);
+        assert_eq!(blobs.len(), txn_count);
```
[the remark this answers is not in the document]
afck: E.g. `events` is a `Vec<Vec<Event>>`. The outer vector always contains one element per transaction.

afck:
Closing in favor of https://github.com/linera-io/linera-protocol/pull/3641 for now.

PR #3620, opened 2025-03-21
Further tune RocksDB options

ndr-ds:
## Motivation

The client became a bottleneck in the benchmarks, more specifically while processing the confirmed certificates locally. Upon further investigation, I saw that saving the chain (in different places of the code) was taking a huge amount of time (seconds).
Given the high load we're putting the client under, it made sense to increase the in memory writes state for RocksDB, to be able to handle that load. 

## Proposal

Increase in memory writes state for RocksDB. That indeed seemed to fix the issue. I changed a few other config options as well to increase parallelism and things like that, so we can make good use of the CPU to prevent RocksDB from slowing down.
After these changes, the client was not a bottleneck anymore.

## Test Plan

Deployed a network and tested this + also CI

## Release Plan


ma2bd (approved):

ndr-ds:
### Merge activity

* **Mar 21, 4:08 PM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/linera-io/linera-protocol/3620).

PR #3622, opened 2025-03-22
Update to Rust version 1.85

### LINKED ISSUE #3196: Application contracts don't load with Rust 1.82 or later
# Description

When using Rust 1.82 or newer, application contracts fail to load to be executed. The error that appears is:

```
thread 'simple_test' panicked at /home/user/.cargo/registry/src/index.crates.io-6f17d22bba15001f/linera-sdk-0.13.1/src/test/chain.rs:93:14:
Failed to execute block.: Execution error: Failed to load contract Wasm module: Invalid table reference (128) during Operation(0)
```

# Steps to Reproduce

```bash
cargo install linera-service
linera project new letest
cd letest
sed -i -e '/channel/s/"[^"]*"/"1.82.0"/' rust-toolchain.toml

cat >> Cargo.toml << EOF

[target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]
anyhow = "*"
tokio = "*"
linera-sdk = { version = "0.13.1", features = ["test", "wasmer"] }
EOF

mkdir tests
cat >> tests/simple.rs << EOF
use letest::LetestAbi;
use linera_sdk::test::TestValidator;

#[tokio::test]
async fn simple_test() -> anyhow::Result<()> {
    let _ = TestValidator::with_current_application::<LetestAbi, _, _>((), ()).await;
    Ok(())
}
EOF

cargo test
```

jvff:
## Motivation

The code was previously stuck with Rust version 1.81 because newer versions would cause applications to fail to load (#3196). However, with that issue fixed ([PR #1](https://github.com/linera-io/parity-wasm/pull/1)), it is now possible to update the code to the latest stable Rust version.

## Proposal

Update the minimum supported Rust version to 1.85, and apply the suggestions from the newer versions of the compiler and the linter.

Also update the nightly version used in CI to a newer version so that it's not so far behind the stable version to the point it causes compilation issues.

## Test Plan

## Release Plan

- These changes follow the usual release cycle, because it's backwards compatible and allows downstream dependencies to use newer Rust versions.

## Links

- Future work: update to use the new Rust 2024 edition.
- Closes #3196 


linera-client/src/persistent/file.rs
```diff
@@ -69,7 +69,7 @@ impl Lock {
 
 impl Drop for Lock {
     fn drop(&mut self) {
-        if let Err(error) = self.0.file().unlock() {
+        if let Err(error) = fs4::FileExt::unlock(self.0.file()) {
```
ma2bd, line 72: nit: Should we have `use fs4::FileExt` above and write `FileExt::unlock()` then?

ma2bd (approved):

linera-client/src/persistent/file.rs
jvff: Indeed! Good catch, thanks!

PR #3624, opened 2025-03-22
Reduce size of `Operation` type

jvff:
## Motivation

While migrating to a newer nightly version of Rust (see #3622), some new lints were suggested. One of them was that the `Operation` type had unbalanced variants. More specifically, the `User` variant was small and had a heap allocation with the bytes for the user operation, while the `System` variant had a large `SystemOperation` type inside.

## Proposal

Apply the suggested fix by the linter to move the `SystemOperation` to the heap by boxing it.

## Test Plan

## Release Plan

- Nothing to do, because this is an internal refactor.

## Links

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -2758,7 +2758,7 @@ where
             let mut owners = ownership.owners.into_iter().collect::<Vec<_>>();
             owners.extend(ownership.super_owners.into_iter().zip(iter::repeat(100)));
             owners.push((new_owner, new_weight));
-            let operations = vec![Operation::System(SystemOperation::ChangeOwnership {
+            let operations = vec![Operation::system(SystemOperation::ChangeOwnership {
                 super_owners: Vec::new(),
                 owners,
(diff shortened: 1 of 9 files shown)


linera-core/src/client/mod.rs
```diff
@@ -2073,12 +2073,12 @@ where
     }
 
     /// Executes an operation.
-    #[instrument(level = "trace")]
```
jvff, line 2076: Note: keeping this would require the `operation` parameter to also have the `Debug` constraint. This would be possible, but I thought it was simpler to just remove it since this method just calls `execute_operations` anyway, which already has the tracing instrumentation as well.

ma2bd (approved):
LGTM

Not sure if the test failure is related

linera-core/src/client/mod.rs
```diff
@@ -2073,12 +2073,12 @@ where
...
     pub async fn execute_operation(
         &self,
-        operation: Operation,
+        operation: impl Into<Operation>,
```
ma2bd, line 2078: We generally try not to do this

PR #3626, opened 2025-03-23
remove alias `UserApplicationId` and rename `UserApplicationDescription`

ma2bd:
## Motivation

Attempts to use different types for `ApplicationId<A>` from `ApplicationId<()>` show that the problem is more complicated than we thought. In the meantime, we have the "worst of both worlds".

## Proposal

* Remove the alias `UserApplicationId` and enjoy the short names.
* Rename `UserApplicationDescription` into `ApplicationDescription`.

## Test Plan

CI

### DIFF
--- a/linera-base/src/data_types.rs
+++ b/linera-base/src/data_types.rs
@@ -824,7 +824,7 @@ pub enum OracleResponse {
 
 impl<'de> BcsHashable<'de> for OracleResponse {}
 
-/// Description of the necessary information to run a user application used within blobs.
+/// Description of a user application.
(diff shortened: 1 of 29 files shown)


linera-base/src/data_types.rs
```diff
@@ -1234,7 +1230,7 @@ doc_scalar!(
     "A blob of binary data, with its content-addressed blob ID."
 );
 doc_scalar!(
-    UserApplicationDescription,
+    ApplicationDescription,
     "Description of the necessary information to run a user application"
```
afck, line 1234: (Not sure if we want to change the wording to "run an application".)

afck (approved):

PR #3628, opened 2025-03-23
fix: error in calling the `chain_id()` method

EricNorberg:
## Motivation

 There are no brackets after `chain_id`, which leads to the fact that the function itself is passed instead of the value, not the result of its execution.
It must call the `chain_id()` method with parentheses
## Release Plan

    - be released in a new SDK,
    - be released in a new SDK,
      SDK soon.

## Links


afck:
@ma2bd: This change looks correct to me? There's no field `chain_id`, only a function.

PR #3632, opened 2025-03-24
Add links to previous blocks with the same message recipients.

### LINKED ISSUE #3630: Sender chain skip lists
A client that wants to receive a message from a chain A in chain B currently has to download all of A up to the point where that message was sent.

For every outgoing message in a block, add the hash of the most recent certificate in the same block that sent a message to the same destination chain. That way we can optimize the client in the future to download only the relevant blocks of chain A.

afck:
## Motivation

We are planning to optimize the client so it does not download all message sender chains (https://github.com/linera-io/linera-protocol/issues/3630). This requires adding data to the `Block` structure itself.

## Proposal

Add the new data to the blocks now, so the optimization is a client-side change and can be added even after the next testnet launch.

## Test Plan

The tests were updated to expect the new entries.

## Release Plan

## Links

- Closes #3630.

### DIFF
--- a/linera-chain/src/chain.rs
+++ b/linera-chain/src/chain.rs
@@ -32,6 +32,7 @@ use linera_execution::{
 use linera_views::{
     context::Context,
     log_view::LogView,
+    map_view::MapView,
     queue_view::QueueView,
     reentrant_collection_view::ReentrantCollectionView,
     register_view::RegisterView,
@@ -171,7 +172,7 @@ static NUM_OUTBOXES: LazyLock<HistogramVec> = LazyLock::new(|| {
 });
 
 /// The BCS-serialized size of an empty [`Block`].
-const EMPTY_BLOCK_SIZE: usize = 93;
+const EMPTY_BLOCK_SIZE: usize = 94;
 
 /// An origin, cursor and timestamp of a unskippable bundle in our inbox.
 #[derive(Debug, Clone, Serialize, Deserialize, async_graphql::SimpleObject)]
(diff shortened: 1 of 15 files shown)


linera-core/src/unit_tests/worker_tests.rs
```diff
@@ -340,9 +340,22 @@ where
         .take(block.operations.len())
         .collect();
     let state_hash = system_state.into_hash().await;
+    let previous_message_blocks = messages
```
deuszx, line 343: What if there was no message for `recipient` in the previous block? Shouldn't we put in the latest one, regardless of how far way it would be?

deuszx (commented):
LGTM so far.

linera-chain/src/chain.rs
```diff
@@ -239,6 +240,8 @@ where
     pub unskippable_bundles: QueueView<C, TimestampedBundleInInbox>,
     /// Unskippable bundles that have been removed but are still in the queue.
     pub removed_unskippable_bundles: SetView<C, BundleInInbox>,
+    /// The heights of previous blocks that sent messages to the same recipients.
+    pub previous_message_blocks: MapView<C, ChainId, BlockHeight>,
```
deuszx, line 244: It's persisted, right? and restored on process startup?

deuszx (approved):

linera-chain/src/chain.rs
afck: Yes, it's part of the chain state, which is persisted whenever we process a confirmed block.

PR #3634, opened 2025-03-24
Remove system channels.

afck:
## Motivation

As of https://github.com/linera-io/linera-protocol/pull/3432 we are not using system channels anymore.

## Proposal

Remove the `SystemChannel` type and any system channel-specific logic, in particular the "off-chain subscription handling".

## Test Plan

The `social` example tests make sure that _user_ pub-sub channels still work.

## Release Plan

## Links

### DIFF
--- a/linera-chain/src/chain.rs
+++ b/linera-chain/src/chain.rs
@@ -18,7 +18,8 @@ use linera_base::{
     },
     ensure,
     identifiers::{
-        AccountOwner, ApplicationId, BlobType, ChainId, ChannelFullName, Destination, MessageId,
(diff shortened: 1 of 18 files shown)


linera-chain/src/chain.rs
```diff
@@ -1167,8 +1157,15 @@ where
                         message.grant == Amount::ZERO,
                         ChainError::GrantUseOnBroadcast
                     );
+                    let GenericApplicationId::User(application_id) =
```
deuszx, line 1160: Is `GenericApplicationId::System` used anywhere else now?

afck: Yes, in events.

PR #3636, opened 2025-03-24
Remove unused `Subscribe` and `Unsubscribe` messages.

afck:
## Motivation

`SystemMessage::Subscribe` and `Unsubscribe` were only used for system channels, which are now gone.

## Proposal

Remove them, and the `subscribers` collection in the system execution state view.

## Test Plan

CI

## Release Plan

## Links

- System channels were removed in https://github.com/linera-io/linera-protocol/pull/3634.

### DIFF
--- a/linera-chain/src/chain.rs
+++ b/linera-chain/src/chain.rs
@@ -532,30 +532,27 @@ where
                 }
             }
         }
-
-        if bundle.goes_to_inbox() {
-            // Process the inbox bundle and update the inbox state.
-            let mut inbox = self.inboxes.try_load_entry_mut(origin).await?;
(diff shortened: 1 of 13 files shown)


ma2bd (approved):
nice!

linera-chain/src/chain.rs
```diff
@@ -532,30 +532,27 @@ where
                 }
             }
         }
-
-        if bundle.goes_to_inbox() {
```
deuszx, line 536: Do I understand correctly that we can remove this b/c all bundles now goe to the inbox?

deuszx (approved):

PR #3642, opened 2025-03-25
Network release workflow.

christos-h:
## Motivation

We'd like to be able to tag git commits on network branches (i.e. `testnet_archimedes`) and trigger new releases to notify validators of hotfixes.

## Proposal

Add a workflow which pattern matches on git tags and triggers releases.

## Test Plan

Need to test manually.


christos-h:
Closing in favor of #3694 .

PR #3647, opened 2025-03-25
Standardize Wasm fuel metering

### LINKED ISSUE #1967: Use `wasm-instrument` in all Wasm runtimes
Currently we are using `wasm-instrument` to provide metering support in the Wasmer runtime.  However, we should try to make our metering identical between runtimes.  An easy way to do this, and one that saves us a fair bit of code, is to use instrumentation uniformly across runtimes, moving the instrumentation logic out of `linera-execution/src/wasm/wasmer.rs` into a shared namespace.

jvff:
## Motivation

Wasmer and Wasmtime had different fuel metering implementations. When using Wasmer, the code used bytecode instrumentation, altering the Wasm bytecode to call back to the host to consume fuel. When using Wasmtime, the code used Wasmtime's own native fuel metering implementation. This led to inconsistencies and required a sanitizer to try to ensure both runtimes measured fuel consumption the same way.

## Proposal

Use the same bytecode instrumentation for all runtimes. This allows removing the sanitizer.

## Test Plan

## Release Plan

- These changes follow the usual release cycle, because this only contains an internal refactor that should not affect any visible behavior.

## Links

- Closes #1967 


ma2bd (approved):
Nice!

PR #3654, opened 2025-03-26
Implement the `EvmQuery` for the EVM contracts

MathieuDutSik:
## Motivation

We use a hacky system for the processing of the query in `handle_query` of REVM. We want to streamline the
mechanism.

## Proposal

The code is adapted from a separate branch that implements the calls `EVM->{EVM,Wasm}` and `Wasm->EVM`
and the changes reflect that.

The following is done:
* The `EvmQuery` is introduced as a type for the input of queries, which reflects what it is.
* The `EvmQuery` is introduced in `linera-base` instead of `linera-execution` since this would force a dependency on `linera-execution` for `linera-sdk`.
* The tests `test_fuel_for_counter_revm_application` and `test_evm_end_to_end_counter` are simplified accordingly.
* The Query type of `EvmAbi` is changed to `EvmQuery`.

## Test Plan

No tests are added, CI suffices.

## Release Plan

Normal release schedule.

## Links

None.

### DIFF
--- a/linera-base/src/vm.rs
+++ b/linera-base/src/vm.rs
@@ -57,11 +57,11 @@ scalar!(VmRuntime);
 #[error("{0:?} is not a valid virtual machine runtime")]
 pub struct InvalidVmRuntime(String);
 
-/// The possible types of queries for an Evm contract
+/// The possible types of queries for an EVM contract
 #[derive(Clone, Debug, Deserialize, Serialize)]
 pub enum EvmQuery {
-    /// The ordinary queries.
+    /// A read-only query.
     Query(Vec<u8>),
-    /// The operation being scheduled.
-    Operation(Vec<u8>),
+    /// A request to schedule an operation that can mutate the application state.
+    Mutation(Vec<u8>),
 }
--- a/linera-base/src/vm.rs
+++ b/linera-base/src/vm.rs
@@ -56,3 +56,12 @@ scalar!(VmRuntime);
 #[derive(Clone, Debug, Error)]
 #[error("{0:?} is not a valid virtual machine runtime")]
 pub struct InvalidVmRuntime(String);
+
+/// The possible types of queries for an Evm contract
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub enum EvmQuery {
+    /// The ordinary queries.
+    Query(Vec<u8>),
+    /// The operation being scheduled.
+    Operation(Vec<u8>),
+}
--- a/linera-execution/src/revm.rs
+++ b/linera-execution/src/revm.rs
@@ -472,7 +472,6 @@ where
         _context: OperationContext,
         operation: Vec<u8>,
     ) -> Result<Vec<u8>, ExecutionError> {
-        let operation = bcs::from_bytes::<Vec<u8>>(&operation)?;
         ensure!(
             operation.len() >= 4,
             ExecutionError::EvmError(EvmExecutionError::OperationIsTooShort)
@@ -489,7 +488,6 @@ where
             unreachable!("It is impossible for a Choice::Call to lead to an Output::Create");
         };
         let output = output.as_ref().to_vec();
-        let output = bcs::to_bytes(&output)?;
         Ok(output)
     }
 
@@ -608,7 +606,6 @@ where
             EvmQuery::Query(vec) => vec,
             EvmQuery::Operation(operation) => {
                 let mut runtime = self.db.runtime.lock().expect("The lock should be possible");
-                let operation = bcs::to_bytes(&operation)?;
                 runtime.schedule_operation(operation)?;
                 return Ok(Vec::new());
             }
--- a/linera-execution/src/revm.rs
+++ b/linera-execution/src/revm.rs
@@ -603,9 +603,7 @@ where
         _context: QueryContext,
         argument: Vec<u8>,
     ) -> Result<Vec<u8>, ExecutionError> {
-        tracing::info!("revm::handle_query, step 1");
         let evm_query: EvmQuery = serde_json::from_slice(&argument)?;
-        tracing::info!("revm::handle_query, step 2");
         let query = match evm_query {
             EvmQuery::Query(vec) => vec,
             EvmQuery::Operation(operation) => {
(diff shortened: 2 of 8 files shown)


linera-execution/src/revm.rs
```diff
@@ -472,6 +472,7 @@ where
         _context: OperationContext,
         operation: Vec<u8>,
     ) -> Result<Vec<u8>, ExecutionError> {
+        let operation = bcs::from_bytes::<Vec<u8>>(&operation)?;
```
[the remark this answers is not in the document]
MathieuDutSik: This is needed for `Wasm->EVM` calls.
The API of `EvmAbi` has the type of operation being `Vec<u8>`. However, that needs to be serialized with BCS like the handle query by `serde_json`. By doing that, we have that the `try_call_application` / `try_query_application` are working properly.

I am not sure what you mean by "JS client". For sure, a lot needs to be done for the support of the Ethereum API. I doubt that the calls will be made directly to `execute_operation`. For a `Vec<u8>`, the BCS serialization is just adding the ULEB128 of the length, which can certainly be done in JS.

There are alternatives to that:
* Having the BCS return identity on the top type being `Vec<u8>`. But I cannot imagine the complexity of this move.
* Having a function besides `try_call_application` named `try_call_untyped_application` that is taking just the `Vec<u8>`. But that would require extensive discussion.

What I think is that it is a reasonable solution that is future proof with respect to the EVM support. Surely, there is room for improvement, as the deserialization of something that was just serialized is not great. But within the scope of this PR, I think that this is the best that can be done.

linera-execution/src/revm.rs
```diff
@@ -604,20 +606,21 @@ where
...
-            let mut runtime = self.db.runtime.lock().expect("The lock should be possible");
-            runtime.schedule_operation(operation)?;
-            let answer = serde_json::json!({"data": ""});
-            let answer = serde_json::to_vec(&answer).unwrap();
-            return Ok(answer);
-        }
-        let argument = argument[10..argument.len() - 3].to_string();
-        let argument = hex::decode(&argument).unwrap();
-        let tx_data = Bytes::copy_from_slice(&argument);
+        tracing::info!("revm::handle_query, step 1");
```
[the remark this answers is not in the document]
MathieuDutSik: Sorry, forgotten print statements.

linera-execution/src/revm.rs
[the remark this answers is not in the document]
ma2bd: We can define `try_call_evm_application` and `try_query_evm_application` in the Rust SDK to avoid forcing BCS in this case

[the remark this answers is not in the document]
MathieuDutSik: Right now (that is before this PR), the EVM apps have an ABI, the one of `EvmAbi`, and the serialization rules apply.

But I agree it is somewhat of a fiction with the types being `Vec<u8>` and this PR introduces the `EvmQuery` which is a little cleaner than what was before. But I agree that this translates into some BCS serialization/deserialization which are annoying.

Getting around that, if we are ready for `call_evm_application`, then maybe we could have a `try_call_untyped_application`. Another idea could be to disable the use of BCS for `EvmAbi`. I will think about that.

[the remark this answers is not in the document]
ma2bd: The low-level runtime system call `contract_wit::try_call_application` is serialization-agnostic and already usable from Rust/Wasm. What we want is building meaning conventions and a high-level runtime method on top of it.

linera-base/src/vm.rs
```diff
@@ -56,3 +56,12 @@ scalar!(VmRuntime);
 #[derive(Clone, Debug, Error)]
 #[error("{0:?} is not a valid virtual machine runtime")]
 pub struct InvalidVmRuntime(String);
+
+/// The possible types of queries for an Evm contract
```
afck, line 60:
```suggestion
/// The possible types of queries for an EVM contract
```

linera-execution/src/revm.rs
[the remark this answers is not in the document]
MathieuDutSik: After merging of https://github.com/linera-io/linera-protocol/pull/3679
the PR has been rebased and is indeed simpler.

linera-base/src/vm.rs
```diff
@@ -56,3 +56,12 @@ scalar!(VmRuntime);
...
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub enum EvmQuery {
+    /// The ordinary queries.
+    Query(Vec<u8>),
+    /// The operation being scheduled.
```
afck, line 65: Maybe something like:

```suggestion
    /// A read-only query.
    Query(Vec<u8>),
    /// A request to schedule an operation that can mutate the application state.
```

linera-execution/src/test_utils/solidity.rs
```diff
@@ -109,3 +110,34 @@ contract ExampleCounter {
...
+    let mut vec: Vec<u8> = Vec::new();
+    for val in value.as_array().unwrap() {
+        let val = val.as_u64().unwrap();
+        let val = val as u8;
+        vec.push(val);
+    }
+    vec
+}
+
+pub fn read_evm_u64_entry(value: Value) -> anyhow::Result<u64> {
```
afck, line 137:
```suggestion
pub fn read_evm_u64_entry(value: Value) -> u64 {
```

linera-service/tests/linera_net_tests.rs
```diff
@@ -350,33 +352,24 @@ async fn test_evm_end_to_end_counter(config: impl LineraNetConfig) -> Result<()>
...
-    let app_file = "app.json";
-    let app_path = path.join(app_file);
-    {
-        std::fs::write(app_path.clone(), &module)?;
-    }
+    let module = get_evm_example_counter()?;
 
-    let contract = app_path.to_path_buf();
-    let service = app_path.to_path_buf();
+    let (contract, service, _dir) = get_contract_service_paths(module)?;
```
afck, line 372: Are we publishing and storing two distinct blobs (contract and service) with identical content? We should deduplicate that for EVM contracts somehow. (Not in this PR.)

linera-base/src/vm.rs
MathieuDutSik: Ok.

linera-execution/src/test_utils/solidity.rs
MathieuDutSik: Yes, correct.

linera-service/tests/linera_net_tests.rs
MathieuDutSik: Will look at that.

afck: Yes, maybe we need to distinguish, and have a separate `BlobType::EvmBytecode` instead?

afck (approved):

PR #3656, opened 2025-03-26
Fix chain execution context for rejected messages.

afck:
## Motivation

`ChainExecutionContext` annotates execution errors with information about where they occurred: in a particular operation, message, or when handling the block in general.

For rejected messages we currently use `Block`, even though they belong to the rejected incoming bundle.

## Proposal

Change the context to `IncomingBundle`.

## Test Plan

CI

## Release Plan

## Links


ma2bd (changes requested):
No this was intended

PR #3657, opened 2025-03-26
Change the way that `UserApplicationId` is computed for `ApplicationDescription` of EVM smart contracts.

MathieuDutSik:
## Motivation

EVM applications should have a hash that is of length 20. We can achieve that by setting the last 12 bytes to 0.

## Proposal

The changes are done to the computation in the hash in the `data_types.rs`. It requires matching the `ApplicationDescription` for the blob type.
In that case, we need to deserialize the description in order to obtain the module_id and whether they are EVM or not.

It has to be done in two places.

## Test Plan

The CI should be doing the adequate test.

## Release Plan

Normal testing release.

## Links

None.

### DIFF
--- a/linera-base/src/crypto/hash.rs
+++ b/linera-base/src/crypto/hash.rs
@@ -46,6 +46,13 @@ impl CryptoHash {
         &self.0
     }
 
+    /// set the hash as the one from EVM by setting the last 12 bytes to zero.
+    pub fn set_as_evm(&mut self) {
+        for index in 20..32 {
+            self.0 .0[index] = 0;
+        }
+    }
+
     /// Returns the hash of `TestString(s)`, for testing purposes.
     #[cfg(with_testing)]
     pub fn test_hash(s: impl Into<String>) -> Self {
--- a/linera-base/src/crypto/hash.rs
+++ b/linera-base/src/crypto/hash.rs
@@ -46,11 +46,9 @@ impl CryptoHash {
         &self.0
     }
 
-    /// set the hash as the one from EVM by setting the last 12 bytes to zero.
-    pub fn set_as_evm(&mut self) {
-        for index in 20..32 {
-            self.0 .0[index] = 0;
-        }
+    /// Force the last 12 bytes of the hash to be zeroes. This is currently used for EVM compatibility
+    pub fn make_evm_compatible(&mut self) {
+        self.0[20..32].fill(0);
     }
 
     /// Returns the hash of `TestString(s)`, for testing purposes.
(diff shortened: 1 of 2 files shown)


linera-base/src/crypto/hash.rs
```diff
@@ -46,6 +46,13 @@ impl CryptoHash {
         &self.0
     }
 
+    /// set the hash as the one from EVM by setting the last 12 bytes to zero.
+    pub fn set_as_evm(&mut self) {
```
ma2bd, line 50:
```suggestion
    /// Force the last 12 bytes of the hash to be zeroes. This is currently used for EVM compatibility.
    pub fn make_evm_compatible(&mut self) {
```

linera-base/src/data_types.rs
```diff
@@ -845,9 +846,11 @@ pub struct ApplicationDescription {
 
 impl From<&ApplicationDescription> for ApplicationId {
     fn from(description: &ApplicationDescription) -> Self {
-        ApplicationId::new(CryptoHash::new(&BlobContent::new_application_description(
-            description,
-        )))
+        let mut hash = CryptoHash::new(&BlobContent::new_application_description(description));
+        if matches!(description.module_id.vm_runtime, VmRuntime::Evm) {
+            hash.set_as_evm();
```
ma2bd, line 851:
```suggestion
            hash.make_evm_compatible();
```

linera-base/src/data_types.rs
```diff
@@ -1076,7 +1079,14 @@ pub struct Blob {
 impl Blob {
     /// Computes the hash and returns the hashed blob for the given content.
     pub fn new(content: BlobContent) -> Self {
-        let hash = CryptoHash::new(&content);
+        let mut hash = CryptoHash::new(&content);
+        if matches!(content.blob_type, BlobType::ApplicationDescription) {
+            let application_description = bcs::from_bytes::<ApplicationDescription>(&content.bytes)
+                .expect("to obtain an application description");
+            if matches!(application_description.module_id.vm_runtime, VmRuntime::Evm) {
+                hash.set_as_evm();
```
ma2bd, line 1087:
```suggestion
                hash.make_evm_compatible();
```

ma2bd (approved):
Accepting to unblock. It's probably cleaner to make `ApplicationId` 20-byte long at some point.

linera-base/src/crypto/hash.rs
```diff
@@ -46,6 +46,13 @@ impl CryptoHash {
...
+        for index in 20..32 {
+            self.0 .0[index] = 0;
```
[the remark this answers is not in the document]
afck: Alternatively maybe `self.0[20..32].fill(0)` would also work?

linera-base/src/data_types.rs
MathieuDutSik: Yes, of course. You do not have to precise this as it is implied by your previous change.

linera-base/src/crypto/hash.rs
MathieuDutSik: This is good, thank you.

PR #3672, opened 2025-03-27
Instead of a key, use an incrementing `index` for events.

afck:
## Motivation

The system needs to enforce that events are never overwritten anyway. It makes them much harder to use for applications if the applications need to keep track of that.

## Proposal

Replace the general `key: Vec<u8>` with `index: u32`. Instead of the application, the system now chooses the index, and always uses the next available number.

## Test Plan

The tests have been updated, and already use system events.

The user application events will be tested more thoroughly later, when we have an example (probably `social`) that uses them.

## Release Plan

## Links

### DIFF
--- a/linera-sdk/src/contract/runtime.rs
+++ b/linera-sdk/src/contract/runtime.rs
@@ -276,8 +276,8 @@ where
     }
 
     /// Adds a new item to an event stream.
-    pub fn emit(&mut self, name: StreamName, key: &[u8], value: &[u8]) {
-        contract_wit::emit(&name.into(), key, value);
+    pub fn emit(&mut self, name: StreamName, index: u32, value: &[u8]) {
+        contract_wit::emit(&name.into(), index, value);
     }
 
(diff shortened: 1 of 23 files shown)


linera-sdk/src/contract/runtime.rs
```diff
@@ -276,8 +276,8 @@ where
     }
 
     /// Adds a new item to an event stream.
-    pub fn emit(&mut self, name: StreamName, key: &[u8], value: &[u8]) {
-        contract_wit::emit(&name.into(), key, value);
+    pub fn emit(&mut self, name: StreamName, value: &[u8]) -> u32 {
```
deuszx, line 279: Maybe add to the comment what's the `u32` returned.

afck: Done in https://github.com/linera-io/linera-protocol/pull/3672/commits/9cf84642f6db6a6ef96b15601d7726d820f1a59b.

PR #3674, opened 2025-03-27
PoC: Sign block proposals with EVM compatible wallets/keys.

deuszx:
## Motivation

We want to make it easier to onboard into Linera ecosystem from EVM world.

## Proposal

Add a new Account* key variant - `EvmSecp256k1` - which will be used for keys/signatures from EVM wallets, like Metamask. The expectation is that if a signature is of `Evm*` type it was produced using `personal_sign` (`eth_sign`) as described in[ EIP-191](https://eips.ethereum.org/EIPS/eip-191) and needs to be verified accordingly.

## Test Plan

_To be filled_

## Release Plan

## Links

### DIFF
--- a/linera-base/src/crypto/ed25519.rs
+++ b/linera-base/src/crypto/ed25519.rs
@@ -284,8 +284,8 @@ impl Ed25519Signature {
     where
         T: BcsSignable<'de>,
     {
-        let preimage = CryptoHash::new(value).as_bytes().0;
-        let signature = secret.0.sign(&preimage);
+        let prehash = CryptoHash::new(value).as_bytes().0;
+        let signature = secret.0.sign(&prehash);
         Ed25519Signature(signature)
     }
 
@@ -311,9 +311,9 @@ impl Ed25519Signature {
     where
         T: BcsSignable<'de>,
     {
-        let preimage = CryptoHash::new(value).as_bytes().0;
+        let prehash = CryptoHash::new(value).as_bytes().0;
         let public_key = dalek::VerifyingKey::from_bytes(&author.0)?;
-        public_key.verify(&preimage, &self.0)
+        public_key.verify(&prehash, &self.0)
     }
 
     /// Checks a signature.
--- a/linera-base/src/crypto/secp256k1/evm.rs
+++ b/linera-base/src/crypto/secp256k1/evm.rs
@@ -1,8 +1,7 @@
-// Copyright (c) Facebook, Inc. and its affiliates.
 // Copyright (c) Zefchain Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-//! Defines secp256k1 signature primitives used by the Linera protocol.
+//! Defines EIP-191 compatible signature primitives used by the Linera protocol.
 
 use std::{
     borrow::Cow,
@@ -119,7 +118,7 @@ impl EvmPublicKey {
 
 impl fmt::Debug for EvmSecretKey {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "<redacted for Secp256k1 secret key>")
+        write!(f, "<redacted for secp256k1 secret key>")
     }
 }
 
--- a/linera-base/src/identifiers.rs
+++ b/linera-base/src/identifiers.rs
@@ -385,7 +385,7 @@ impl From<Ed25519PublicKey> for AccountOwner {
 
 impl From<EvmPublicKey> for AccountOwner {
     fn from(public_key: EvmPublicKey) -> Self {
-        AccountOwner::Address32(CryptoHash::new(&public_key))
+        AccountOwner::Address20(alloy_primitives::Address::from_public_key(&public_key.0).into())
     }
 }
 
--- a/linera-chain/src/data_types.rs
+++ b/linera-chain/src/data_types.rs
@@ -843,7 +843,7 @@ mod signing {
     fn proposal_content_singing() {
         use std::str::FromStr;
 
-        // Generated in metamask.
+        // Generated in MetaMask.
         let pk = "f77a21701522a03b01c111ad2d2cdaf2b8403b47507ee0aec3c2e52b765d7a66";
 
         let signer: AccountSecretKey = AccountSecretKey::EvmSecp256k1(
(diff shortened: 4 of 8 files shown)


linera-base/src/identifiers.rs
```diff
@@ -378,6 +379,12 @@ impl From<Ed25519PublicKey> for AccountOwner {
     }
 }
 
+impl From<EvmPublicKey> for AccountOwner {
+    fn from(public_key: EvmPublicKey) -> Self {
+        AccountOwner::Address32(CryptoHash::new(&public_key))
```
ma2bd, line 384: I believe this is where we should create a 20-byte address in the Ethereum way, so Linera applications show the same user address as Metamask

deuszx: Done in [EvmPublicKey maps to AccountOwner::Address20](https://github.com/linera-io/linera-protocol/pull/3674/commits/c981041ff3578149275cd15b2b17c07d91ed3fc2)

linera-service/benches/transfers.rs
```diff
@@ -67,8 +67,10 @@ async fn setup_native_token_balances(
         .then(|idx| {
             let key_pair = if idx % 2 == 0 {
                 AccountSecretKey::Secp256k1(Secp256k1SecretKey::generate())
-            } else {
+            } else if idx % 3 == 0 {
                 AccountSecretKey::Ed25519(Ed25519SecretKey::generate())
+            } else {
+                AccountSecretKey::EvmSecp256k1(EvmSecretKey::generate())
             };
```
afck, line 74: Now secp256k1 has 50%, Ed25519 17% and EVM-secp256k1 33%.

Maybe simpler: `match idx % 3`

linera-rpc/tests/format.rs
```diff
@@ -41,6 +41,17 @@ fn get_registry() -> Result<Registry> {
         );
         tracer.trace_value(&mut samples, &validator_keypair.public_key)?;
         tracer.trace_value(&mut samples, &validator_signature)?;
+
+        // We also record a separate samples for EVM-compatible keys,
```
afck, line 45:
```suggestion
        // We also record separate samples for EVM-compatible keys,
```

linera-core/src/unit_tests/worker_tests.rs
```diff
@@ -407,8 +408,12 @@ fn direct_credit_message(recipient: ChainId, amount: Amount) -> OutgoingMessage
 fn generate_key_pairs(count: usize) -> Vec<AccountSecretKey> {
     let mut key_pairs = iter::repeat_with(Secp256k1SecretKey::generate)
         .map(AccountSecretKey::Secp256k1)
-        .take(count)
+        .take(count - 2)
         .collect::<Vec<_>>();
+    key_pairs.extend(vec![
+        AccountSecretKey::EvmSecp256k1(EvmSecretKey::generate()),
+        AccountSecretKey::EvmSecp256k1(EvmSecretKey::generate()),
+    ]);
```
afck, line 416: What if `count < 2`? Maybe just do a single iteration and alternate the variants instead?

linera-chain/src/data_types.rs
```diff
@@ -963,3 +963,51 @@ doc_scalar!(
...
+    };
+    use linera_execution::committee::Epoch;
+
+    use crate::data_types::{ProposalContent, ProposedBlock};
+
+    #[test]
+    fn proposal_content_singing() {
+        use std::str::FromStr;
+
+        // Generated in metamask.
```
afck, line 982:
```suggestion
        // Generated in MetaMask.
```

linera-chain/src/data_types.rs
```diff
@@ -963,3 +963,51 @@ doc_scalar!(
...
+            previous_block_hash: None,
+        };
+
+        let proposal = ProposalContent {
+            block: proposed_block,
+            round: Round::SingleLeader(11),
+            outcome: None,
+        };
+
+        // personal_sign of the `proposal_hash` done via Metamask.
```
afck, line 1006:
```suggestion
        // personal_sign of the `proposal_hash` done via MetaMask.
```

linera-base/src/crypto/secp256k1/evm.rs
```diff
@@ -0,0 +1,635 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
+// Copyright (c) Zefchain Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Defines secp256k1 signature primitives used by the Linera protocol.
```
afck, line 5:
```suggestion
//! Defines EVM-compatible secp256k1 signature primitives used by the Linera protocol.
```

(Or "Ethereum-compatible" or "MetaMask-compatible"?)

linera-base/src/crypto/secp256k1/evm.rs
```diff
@@ -0,0 +1,635 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
```
afck, line 1: (I think we omit this for new files? Or is it because it copies some code from the other file?)

linera-base/src/crypto/secp256k1/evm.rs
```diff
@@ -0,0 +1,635 @@
...
+                let error = CryptoError::Secp256k1PointAtInfinity(hex::encode(bytes));
+                Err(error)
+            }
+        }
+    }
+}
+
+impl fmt::Debug for EvmSecretKey {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "<redacted for Secp256k1 secret key>")
```
afck, line 122:
```suggestion
        write!(f, "<redacted for secp256k1 secret key>")
```

linera-base/src/crypto/ed25519.rs
```diff
@@ -312,10 +311,9 @@ impl Ed25519Signature {
     where
         T: BcsSignable<'de>,
     {
-        let mut message = Vec::new();
-        value.write(&mut message);
+        let preimage = CryptoHash::new(value).as_bytes().0;
```
afck, line 314: Did you mean `prehash`?

Do we need to change the `new` implementation for the two other algorithms as well to avoid unnecessary double hashing?

deuszx: Yes, `prehash`. Other impls should already match - `secp256k1` was doing that already and for EIP-191

linera-base/src/crypto/secp256k1/evm.rs
deuszx: It's EVM-wide standard but this is not entirely correct either. I should mention the EIP-191 here.

linera-base/src/crypto/secp256k1/evm.rs
deuszx: Ah, this particular line you mean? I didn't check in details the error message.

afck: Yes; I guess if it's all new code, the copyright is only Zefchain.

linera-base/src/identifiers.rs
```diff
@@ -380,6 +381,12 @@ impl From<Ed25519PublicKey> for AccountOwner {
     }
 }
 
+impl From<EvmPublicKey> for AccountOwner {
+    fn from(public_key: EvmPublicKey) -> Self {
+        AccountOwner::Address20(alloy_primitives::Address::from_public_key(&public_key.0).into())
```
ma2bd, line 386: nice!

PR #3679, opened 2025-03-28
Introduce the `{de}serialize_{operation,response}` to the ContractAbi.

MathieuDutSik:
## Motivation

The EVM apps do not have a type associated with them, instead, the `Operation` / `Response` are `Vec<u8>`.
Using BCS for them is not adequate as it leads to some serialization followed by deserialization situations which
are suboptimal.

## Proposal

The following is done:
* The `{de}serialization_{operation,response}` is introduced in the `ContractAbi`.
* They use strings as error types. Using `bcs::Error` is not the point. And `anyhow::Error` is for tests and binaries.
* There is a standard implementation in the trait, which is overloaded in the `EvmAbi`.
* The use of BCS serialization is removed in the place of the code where it was and replaced by the introduced functions. They `unwrap()` as the use case of those functions always panics in case of errors.
* Whenever a serialization error occurs, the error and the input are printed in the error message.

## Test Plan

The CI.

## Release Plan

Normal release plan.

## Links

None

### DIFF
--- a/linera-sdk/src/contract/runtime.rs
+++ b/linera-sdk/src/contract/runtime.rs
@@ -262,8 +262,7 @@ where
         application: ApplicationId<A>,
         call: &A::Operation,
     ) -> A::Response {
-        let call_bytes = bcs::to_bytes(call)
-            .expect("Failed to serialize `Operation` type for a cross-application call");
+        let call_bytes = A::serialize_operation(call).unwrap();
 
         let response_bytes = contract_wit::try_call_application(
             authenticated,
@@ -271,8 +270,7 @@ where
             &call_bytes,
         );
 
-        bcs::from_bytes(&response_bytes)
-            .expect("Failed to deserialize `Response` type from cross-application call")
+        A::deserialize_response(response_bytes).unwrap()
     }
 
     /// Adds a new item to an event stream. Returns the new event's index in the stream.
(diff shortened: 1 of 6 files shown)


ma2bd (approved):
LGTM!

linera-sdk/src/contract/runtime.rs
```diff
@@ -262,17 +262,15 @@ where
         application: ApplicationId<A>,
         call: &A::Operation,
     ) -> A::Response {
-        let call_bytes = bcs::to_bytes(call)
-            .expect("Failed to serialize `Operation` type for a cross-application call");
```
ma2bd, line 266: The message in the expect is a bit more precise.

MathieuDutSik: Yes, that is right. We lose the context that it was a "cross-application call". I think that this is not a major problem. 

Note also that the error messages that we obtain are better in two ways:
* We get the error that occurred in the serialization.
* We get the input.

MathieuDutSik: If that is viewed as a major problem, then we could pass the context in the function call.

ma2bd: How about keeping all the `.expect` as they are, and use `bcs::..().map_err(String::from)` in the abi methods?

MathieuDutSik: Using the `expect` is a good idea indeed as the error show up.
Using `String::from` is not a good idea since if we want to debug serialization error, we need to have the input.

afck (approved):

PR #3682, opened 2025-03-29
Add mdbook anchors to various source files

ma2bd:
## Motivation

Help with Rust quotes in the dev manual

## Proposal

* Add mdbook anchors
* Rewrite some of the comments

## Test Plan

CI + used in a next PR in the mdbook

## Release Plan

- These changes should be backported to the latest `mainnet` branch and the next testnet branch (`babbage`)

### DIFF
--- a/linera-sdk/src/contract/runtime.rs
+++ b/linera-sdk/src/contract/runtime.rs
@@ -225,12 +225,15 @@ where
     }
 
     /// Calls another application.
+    // ANCHOR: call_application
     pub fn call_application<A: ContractAbi + Send>(
         &mut self,
         authenticated: bool,
         application: ApplicationId<A>,
         call: &A::Operation,
-    ) -> A::Response {
+    ) -> A::Response
+// ANCHOR_END: call_application
+    {
         let call_bytes = bcs::to_bytes(call)
             .expect("Failed to serialize `Operation` type for a cross-application call");
 
--- a/linera-sdk/src/lib.rs
+++ b/linera-sdk/src/lib.rs
@@ -81,13 +81,25 @@ doc_scalar!(DataBlobHash, "Hash of a Data Blob");
 /// storage and is gas-metered.
 ///
 /// Below we use the word "transaction" to refer to the current operation or message being
-/// executed.
+/// executed. Operations are created by users and added to blocks, serving as the starting
+/// point for an application's execution. Messages are executed when a message created by
+/// the same application is received from another chain and accepted in a block.
+///
+/// `instantiate` is only called once when the application is created and only on the
+/// microchain that created the application. To share configuration data on every chain,
+/// use [`Contract::Parameters`] instead of [`Contract::InstantiationArgument`].
+///
+/// Messages are mean to be sent across chains. They are created and received by the same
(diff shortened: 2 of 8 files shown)


Twey:
How about instead of annotating our source files we use the [`include-rs`](https://crates.io/crates/mdbook-include-rs) plugin to include specific functions?

ma2bd:
> Instead of annotating our source files maybe we could use the [`include-rs`](https://crates.io/crates/mdbook-include-rs) plugin to include specific functions?

Interesting, let me give it a try

ma2bd:
> > Instead of annotating our source files maybe we could use the [`include-rs`](https://crates.io/crates/mdbook-include-rs) plugin to include specific functions?
> 
> Interesting, let me give it a try

@afck @Twey So I tried `mdbook-include-rs`. It's quite promising but also very new (literally 2 weeks old!). You have to `cargo install --git` from the tip of the repo, and (understandably) there are still some limitations: e.g. I couldn't find
* how to include the body of a method in a trait implementation, (as opposed to a function)
* how to include a trait without the top comment.

linera-sdk/src/contract/runtime.rs
```diff
@@ -225,12 +225,15 @@ where
...
     /// Calls another application.
+    // ANCHOR: call_application
     pub fn call_application<A: ContractAbi + Send>(
         &mut self,
         authenticated: bool,
         application: ApplicationId<A>,
         call: &A::Operation,
-    ) -> A::Response {
+    ) -> A::Response
+// ANCHOR_END: call_application
```
afck, line 235:
```suggestion
    // ANCHOR_END: call_application
```

linera-sdk/src/lib.rs
```diff
@@ -81,69 +81,58 @@ doc_scalar!(DataBlobHash, "Hash of a Data Blob");
...
-/// executed.
+/// executed. Operations are created by users and added to blocks, serving as the starting
+/// point for an application's execution. Messages are executed when a message created by
+/// the same application is received from another chain and accepted in a block.
+///
+/// `instantiate` is only called once when the application is created and only on the
+/// microchain that created the application. To share configuration data on every chain,
+/// use [`Contract::Parameters`] instead of [`Contract::InstantiationArgument`].
+///
+/// Messages are mean to be sent across chains. They are created and received by the same
```
afck, line 92:
```suggestion
/// Messages are meant to be sent across chains. They are created and received by the same
```

afck (approved):

Twey:
> how to include the body of a method in a trait implementation, (as opposed to a function)

From the README:
> `impl StructName::method_name` - includes a specific method from an impl block

I'm not sure about stripping off the doc comments though.  I'm okay with this for now/until this crate gains the features we need.

linera-sdk/src/contract/runtime.rs
ma2bd: I believe this one comes from rustfmt

ma2bd:
> From the README:
> 
> > `impl StructName::method_name` - includes a specific method from an impl block

I tried that and it didn't work for trait implementations (yet?)

PR #3684, opened 2025-03-31
Do not drop notifiers silently if the query may succeed

ma2bd:
## Motivation

Resolve suspicious `ERROR`s in logs

## Proposal

* Do not drop notifiers silently if the query may succeed
* The error is still useful and should never happen.

## Test Plan

Run tests locally => no more errors

## Release Plan

### DIFF
--- a/linera-core/src/chain_worker/state/attempted_changes.rs
+++ b/linera-core/src/chain_worker/state/attempted_changes.rs
@@ -299,6 +299,12 @@ where
         }
         if tip.next_block_height > height {
             // Block was already confirmed.
+            if let Some(notifier) = notify_when_messages_are_delivered {
+                // Nothing to wait for.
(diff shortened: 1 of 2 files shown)


linera-core/src/chain_worker/state/attempted_changes.rs
```diff
@@ -299,6 +299,12 @@ where
         }
         if tip.next_block_height > height {
             // Block was already confirmed.
+            if let Some(notifier) = notify_when_messages_are_delivered {
+                // Nothing to wait for.
+                if let Err(()) = notifier.send(()) {
+                    warn!("Failed to notify message delivery to caller");
+                }
+            }
```
afck, line 307: Right, good catch!

afck (approved):

PR #3692, opened 2025-04-01
Add basic user events, without subscriptions.

afck:
## Motivation

We want to replace pub-sub channels with event streams (https://github.com/linera-io/linera-protocol/issues/365).

This is a complex task, so I am splitting it up into multiple PRs. This one does not fully implement subscriptions yet, but adds `read_event`.

## Proposal

* Make `emit` type safe using an associated `EventValue` type.
* Add `read_event`.
* Track event stream subscriptions. (But the client does not act on them yet.)
* Add a minimal assertion to the `social` example to test `read_event`.
* Charge reading and writing fees for events; add a TODO for a separate fee category.

## Test Plan

`read_event` is now tested. The other features will be tested later, when subscriptions are implemented and the `social` example is fully migrated to events.

## Release Plan

- 
## Links

- Closes https://github.com/linera-io/linera-protocol/issues/2308.
- Part of https://github.com/linera-io/linera-protocol/issues/365.

### DIFF
--- a/linera-execution/src/lib.rs
+++ b/linera-execution/src/lib.rs
@@ -745,7 +745,7 @@ pub trait ContractRuntime: BaseRuntime {
     fn subscribe_to_events(
         &mut self,
         chain_id: ChainId,
-        application_id: GenericApplicationId,
+        application_id: ApplicationId,
         stream_name: StreamName,
     ) -> Result<(), ExecutionError>;
 
@@ -753,7 +753,7 @@ pub trait ContractRuntime: BaseRuntime {
     fn unsubscribe_from_events(
         &mut self,
         chain_id: ChainId,
-        application_id: GenericApplicationId,
+        application_id: ApplicationId,
         stream_name: StreamName,
     ) -> Result<(), ExecutionError>;
 
--- a/linera-sdk/wit/contract-runtime-api.wit
+++ b/linera-sdk/wit/contract-runtime-api.wit
@@ -17,8 +17,8 @@ interface contract-runtime-api {
     try-call-application: func(authenticated: bool, callee-id: application-id, argument: list<u8>) -> list<u8>;
     emit: func(name: stream-name, value: list<u8>) -> u32;
     read-event: func(chain-id: chain-id, name: stream-name, index: u32) -> list<u8>;
-    subscribe-to-events: func(chain-id: chain-id, application-id: generic-application-id, name: stream-name);
-    unsubscribe-from-events: func(chain-id: chain-id, application-id: generic-application-id, name: stream-name);
+    subscribe-to-events: func(chain-id: chain-id, application-id: application-id, name: stream-name);
+    unsubscribe-from-events: func(chain-id: chain-id, application-id: application-id, name: stream-name);
     query-service: func(application-id: application-id, query: list<u8>) -> list<u8>;
     consume-fuel: func(fuel: u64);
     validation-round: func() -> option<u32>;
@@ -96,11 +96,6 @@ interface contract-runtime-api {
     variant destination {
         recipient(chain-id),
         subscribers(channel-name),
-    }
-
-    variant generic-application-id {
-        system,
-        user(application-id),
     }
 
     record message-id {
--- a/linera-execution/src/runtime.rs
+++ b/linera-execution/src/runtime.rs
@@ -1348,7 +1348,7 @@ impl ContractRuntime for ContractSyncRuntimeHandle {
     fn subscribe_to_events(
         &mut self,
         chain_id: ChainId,
-        application_id: GenericApplicationId,
+        application_id: ApplicationId,
         stream_name: StreamName,
     ) -> Result<(), ExecutionError> {
         let this = self.inner();
@@ -1358,7 +1358,7 @@ impl ContractRuntime for ContractSyncRuntimeHandle {
         );
         let stream_id = StreamId {
             stream_name,
-            application_id,
+            application_id: application_id.into(),
         };
         let subscriber_app_id = this.current_application().id;
         this.execution_state_sender
@@ -1375,7 +1375,7 @@ impl ContractRuntime for ContractSyncRuntimeHandle {
     fn unsubscribe_from_events(
         &mut self,
         chain_id: ChainId,
-        application_id: GenericApplicationId,
+        application_id: ApplicationId,
         stream_name: StreamName,
     ) -> Result<(), ExecutionError> {
         let this = self.inner();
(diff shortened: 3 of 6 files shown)


examples/ethereum-tracker/src/contract.rs
```diff
@@ -30,6 +30,7 @@ impl Contract for EthereumTrackerContract {
     type Message = ();
     type InstantiationArgument = InstantiationArgument;
     type Parameters = ();
+    type EventValue = ();
```
deuszx, line 33: Maybe just `Event`?

examples/meta-counter/src/contract.rs
```diff
@@ -40,8 +41,10 @@ impl Contract for MetaCounterContract {
         // Send a no-op message to ourselves. This is only for testing contracts that send messages
         // on initialization. Since the value is 0 it does not change the counter value.
         let this_chain = self.runtime.chain_id();
-        self.runtime
-            .emit(StreamName(b"announcements".to_vec()), b"instantiated");
+        self.runtime.emit(
+            StreamName(b"announcements".to_vec()),
+            &b"instantiated".to_vec(),
```
deuszx, line 46: This will eventually by serialize `Self::EventValue`?

deuszx: Ah, it already is - it's just this particular example.

afck: This _is_ `Self::EventValue`. I made it `Vec<u8>` for the meta counter. Looking at this, maybe `String` would make more sense?

examples/ethereum-tracker/src/contract.rs
afck: We already have an `Event` type which includes the event ID. I wanted to avoid confusion with that. Not sure.

linera-execution/src/lib.rs
```diff
@@ -731,6 +731,30 @@ pub trait ContractRuntime: BaseRuntime {
     /// Adds a new item to an event stream. Returns the new event's index in the stream.
     fn emit(&mut self, name: StreamName, value: Vec<u8>) -> Result<u32, ExecutionError>;
 
+    /// Reads an event from a stream. Returns the event's value.
```
deuszx, line 734: Does absence become an error? If so, please add that to a comment.

I'd personally prefer if it was an `Option<Vec<u8>>` but we can leave that for a separate PR.

afck: It has to be an error, because we want to allow reading events in fast blocks eventually. For that, we'll need to make execution fully deterministic, i.e. the outcome cannot depend on whether a validator is already seeing the event or not. Instead, it needs to return an `EventNotFound` error so that the client can re-send the event (i.e. the block that created it) and retry.

deuszx: I don't understand how returning a `None` - instead of an `Error` - affects the (non)determinism but I can imagine that how the return value is _handled_ could affect it

linera-execution/src/system.rs
```diff
@@ -172,6 +176,8 @@ pub enum SystemOperation {
     ProcessNewEpoch(Epoch),
     /// Processes an event about a removed epoch and committee.
     ProcessRemovedEpoch(Epoch),
+    /// Updates the event stream trackers.
+    UpdateStreams(Vec<(ChainId, StreamId, u32)>),
```
deuszx, line 180: I can't find where this is emitted 👀

linera-execution/src/lib.rs
afck: The runtime method doesn't return an error. It fails the block execution.

linera-execution/src/system.rs
afck: It isn't. That will follow in the next PR.

deuszx: OK, slightly confusing it's here now.

afck: Yes; I was trying to minimize the diff of the upcoming PR, because that will already contain somewhat complicated code. But I could remove it for now, if you prefer.

deuszx (approved):

linera-execution/src/system.rs
deuszx: It's easier to review stuff when there's a context for its usage.

linera-sdk/src/lib.rs
```diff
@@ -101,6 +101,9 @@ pub trait Contract: WithContractAbi + ContractAbi + Sized {
     /// instead.
     type InstantiationArgument: Serialize + DeserializeOwned + Debug;
 
+    /// Event values for streams created by this application.
+    type EventValue: Serialize + DeserializeOwned + Debug;
```
ma2bd, line 105: How about `type StreamName` here?

ma2bd: (could also be another PR)

linera-sdk/wit/contract-runtime-api.wit
```diff
@@ -16,6 +16,9 @@ interface contract-runtime-api {
     create-application: func(module-id: module-id, parameters: list<u8>, argument: list<u8>, required-application-ids: list<application-id>) -> application-id;
     try-call-application: func(authenticated: bool, callee-id: application-id, argument: list<u8>) -> list<u8>;
     emit: func(name: stream-name, value: list<u8>) -> u32;
+    read-event: func(chain-id: chain-id, name: stream-name, index: u32) -> list<u8>;
+    subscribe-to-events: func(chain-id: chain-id, application-id: generic-application-id, name: stream-name);
+    unsubscribe-from-events: func(chain-id: chain-id, application-id: generic-application-id, name: stream-name);
```
ma2bd, line 21: In the spirit of system calls, we could have a separate triplet of methods for each system stream (which currently is not even needed), then a single API for all user streams.

ma2bd: In this way, `linera_execution::StreamId` could be like `Operation`:
```
enum StreamId {
    System(SystemStreamId),
    User(ApplicationId, Vec<u8>),
}
```

linera-sdk/src/lib.rs
afck: `StreamName`s can also be used by other applications, for subscriptions. So `process_events`, at least, will have to use `StreamId`s that only contain `Vec<u8>`, because they could be from different applications.

linera-sdk/wit/contract-runtime-api.wit
afck: I changed it to user app IDs in https://github.com/linera-io/linera-protocol/pull/3692/commits/04b1b74f52844ecd3c844e84e02d3d4c256d8bbf.

afck: (I didn't change the `StreamId` type yet, because I'll probably have to do that in a later PR anyway.)

PR #3693, opened 2025-04-01
Count messages, not transactions, in the outgoing message counter.

afck:
## Motivation

The message counter currently counts how many `Vec<OutgoingMessage>` vectors there are in a block, which is the number of transactions.

## Proposal

Count the number of messages instead.
(Edit: This isn't actually used in Linera itself at all, but it it exposed as part of the chain state, e.g. for block explorers.)

## Test Plan

The change is small and only affects metrics, so there is no dedicated test.

## Release Plan

## Links

### DIFF
--- a/linera-chain/src/chain.rs
+++ b/linera-chain/src/chain.rs
@@ -324,7 +324,8 @@ impl ChainTipState {
             .ok_or(ArithmeticError::Overflow)?;
 
         let num_outgoing_messages =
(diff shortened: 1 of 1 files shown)


ma2bd (approved):

ndr-ds:
This metric was indeed suspiciously low when benchmarking :) thank you for the fix!

PR #3696, opened 2025-04-01
Extract process of signing into a trait

deuszx:
## Motivation

Currently client processes own a private key (`AccountSecretKey`) which is potentially insecure but also prevents implementation of external signers (wallet extensions, hardware wallets, etc.). This PR addresses that.

## Proposal

We achieve that by the introduction of a new `Signer` trait that encapsulates the actions of signing and getting a public key for an `AccountOwner` instance. A couple of other changes were made/introduced to support that:
- `Signer` is passed around as `Box<dyn Signer>` to hide the implementation details of the actual `Signer` instance.
- `Signer::sign` signs a `CryptoHash` (rather than `T: BcsSignable`).
- (required by the above ☝️ ) New `sign_prehash(self, CryptoHash)` methods added on all secret keys types in the `linera-crypto`
- An `InMemSigner` was introduced for backwards-compatibility and intermediate usage in native and web clients.
- Removed `assigned_keys` and `unassignd_keys` from the `Wallet`. Now the `Signer` is the source of truth about which keys are available.

## Test Plan

All tests have been updated to pass.

## Release Plan

## Links

### DIFF
--- a/linera-base/src/crypto/mod.rs
+++ b/linera-base/src/crypto/mod.rs
@@ -8,6 +8,7 @@ mod ed25519;
 mod hash;
 #[allow(dead_code)]
 mod secp256k1;
+mod signer;
 use std::{fmt::Display, io, num::ParseIntError, str::FromStr};
 
 use alloy_primitives::FixedBytes;
@@ -20,6 +21,7 @@ pub use secp256k1::{
     Secp256k1PublicKey, Secp256k1SecretKey, Secp256k1Signature,
 };
 use serde::{Deserialize, Serialize};
+pub use signer::*;
 use thiserror::Error;
 
 /// The public key of a validator.
@@ -133,10 +135,34 @@ impl AccountSecretKey {
         }
     }
 
+    /// Creates a signature for the `value`.
+    pub fn sign_prehash(&self, value: CryptoHash) -> AccountSignature {
+        match self {
+            AccountSecretKey::Ed25519(secret) => {
+                let signature = Ed25519Signature::sign_prehash(secret, value);
+                AccountSignature::Ed25519(signature)
+            }
+            AccountSecretKey::Secp256k1(secret) => {
+                let signature = secp256k1::Secp256k1Signature::sign_prehash(secret, value);
+                AccountSignature::Secp256k1(signature)
+            }
+            AccountSecretKey::EvmSecp256k1(secret) => {
+                let signature = secp256k1::evm::EvmSignature::sign_prehash(secret, value);
+                AccountSignature::EvmSecp256k1(signature)
+            }
+        }
+    }
+
     #[cfg(all(with_testing, with_getrandom))]
     /// Generates a new key pair using the operating system's RNG.
     pub fn generate() -> Self {
         AccountSecretKey::Ed25519(Ed25519SecretKey::generate())
+    }
+
+    #[cfg(with_getrandom)]
+    /// Generates a new key pair from the given RNG. Use with care.
+    pub fn generate_from<R: CryptoRng>(rng: &mut R) -> Self {
+        AccountSecretKey::Ed25519(Ed25519SecretKey::generate_from(rng))
     }
 }
 
--- a/linera-base/src/crypto/signer.rs
+++ b/linera-base/src/crypto/signer.rs
@@ -1,0 +1,315 @@
+// Copyright (c) Zefchain Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+use async_trait::async_trait;
+pub use in_mem::InMemorySigner;
+
+use super::CryptoHash;
+use crate::{
+    crypto::{AccountPublicKey, AccountSignature},
+    identifiers::AccountOwner,
+};
+
+/// A trait for signing keys.
+#[async_trait]
+pub trait Signer: Send + Sync {
+    /// Creates a signature for the given `value` using the provided `owner`.
+    // DEV: We sign `CryptoHash` type, rather than `&[u8]` to make sure we don't sign
+    // things accidentally. See [`CryptoHash::new`] for how the type's name is included
+    // in the resulting hash, providing the canonicity of the hashing process.
+    async fn sign(
+        &self,
+        owner: &AccountOwner,
+        value: &CryptoHash,
+    ) -> Result<AccountSignature, Box<dyn std::error::Error>>;
+
+    /// Returns the public key corresponding to the given `owner`.
+    async fn get_public_key(
+        &self,
+        owner: &AccountOwner,
+    ) -> Result<AccountPublicKey, Box<dyn std::error::Error>>;
+
+    /// Returnes whether the given `owner` is a known signer.
+    async fn contains_key(&self, owner: &AccountOwner) -> Result<bool, Box<dyn std::error::Error>>;
+}
+
+#[async_trait]
+impl Signer for Box<dyn Signer> {
+    async fn sign(
+        &self,
+        owner: &AccountOwner,
+        value: &CryptoHash,
+    ) -> Result<AccountSignature, Box<dyn std::error::Error>> {
+        (**self).sign(owner, value).await
+    }
+
+    async fn get_public_key(
+        &self,
+        owner: &AccountOwner,
+    ) -> Result<AccountPublicKey, Box<dyn std::error::Error>> {
+        (**self).get_public_key(owner).await
+    }
+
+    async fn contains_key(&self, owner: &AccountOwner) -> Result<bool, Box<dyn std::error::Error>> {
+        (**self).contains_key(owner).await
+    }
+}
+
+/// In-memory implementatio of the [`Signer`] trait.
+mod in_mem {
+    use std::{
+        collections::BTreeMap,
+        sync::{Arc, RwLock},
+    };
+
+    use async_trait::async_trait;
+    use serde::{Deserialize, Serialize};
+
+    #[cfg(with_getrandom)]
+    use crate::crypto::CryptoRng;
+    use crate::{
+        crypto::{AccountPublicKey, AccountSecretKey, AccountSignature, CryptoHash, Signer},
+        identifiers::AccountOwner,
+    };
+
+    /// In-memory signer.
+    #[derive(Clone)]
+    pub struct InMemorySigner(Arc<RwLock<InMemSignerInner>>);
+
+    impl InMemorySigner {
+        /// Creates a new `InMemSigner` seeded with `prng_seed`.
+        /// If `prng_seed` is `None`, an `OsRng` will be used.
+        #[cfg(with_getrandom)]
+        pub fn new(prng_seed: Option<u64>) -> Self {
+            InMemorySigner(Arc::new(RwLock::new(InMemSignerInner::new(prng_seed))))
+        }
+
+        /// Creates a new `InMemSigner`.
+        #[cfg(not(with_getrandom))]
+        pub fn new() -> Self {
+            InMemorySigner(Arc::new(RwLock::new(InMemSignerInner::new())))
+        }
+
+        /// Generates a new key pair from Signer's RNG. Use with care.
+        #[cfg(with_getrandom)]
+        pub fn generate_new(&mut self) -> AccountPublicKey {
+            let mut inner = self.0.write().unwrap();
+            let secret = AccountSecretKey::generate_from(&mut inner.rng_state.prng);
+            if inner.rng_state.testing_seed.is_some() {
+                // Generate a new testing seed for the case when we need to store the PRNG state.
+                // It provides a "forward-secrecy" property for the testing seed.
+                // We do not do that for the case when `testing_seed` is `None`, because
+                // we default to the usage of OsRng in that case.
+                inner.rng_state.testing_seed = Some(inner.rng_state.prng.next_u64());
+            }
+            let public = secret.public();
+            let owner = AccountOwner::from(public);
+            inner.keys.insert(owner, secret);
+            public
+        }
+
+        /// Returns the public key corresponding to the given `owner`.
+        pub fn keys(&self) -> Vec<(AccountOwner, Vec<u8>)> {
+            let inner = self.0.read().unwrap();
+            inner.keys()
+        }
+    }
+
+    /// In-memory signer.
+    struct InMemSignerInner {
+        keys: BTreeMap<AccountOwner, AccountSecretKey>,
+        #[cfg(with_getrandom)]
+        rng_state: RngState,
+    }
+
+    #[cfg(with_getrandom)]
+    struct RngState {
+        prng: Box<dyn CryptoRng>,
+        #[cfg(with_getrandom)]
+        testing_seed: Option<u64>,
+    }
+
+    #[cfg(with_getrandom)]
+    impl RngState {
+        fn new(prng_seed: Option<u64>) -> Self {
+            let prng: Box<dyn CryptoRng> = prng_seed.into();
+            RngState {
+                prng,
+                #[cfg(with_getrandom)]
+                testing_seed: prng_seed,
+            }
+        }
+    }
+
+    impl InMemSignerInner {
+        /// Creates a new `InMemSignerInner` seeded with `prng_seed`.
+        /// If `prng_seed` is `None`, an `OsRng` will be used.
+        #[cfg(with_getrandom)]
+        pub fn new(prng_seed: Option<u64>) -> Self {
+            InMemSignerInner {
+                keys: BTreeMap::new(),
+                rng_state: RngState::new(prng_seed),
+            }
+        }
+
+        /// Creates a new `InMemSignerInner`.
+        #[cfg(not(with_getrandom))]
+        pub fn new() -> Self {
+            InMemSignerInner {
+                keys: BTreeMap::new(),
+            }
+        }
+
+        pub fn keys(&self) -> Vec<(AccountOwner, Vec<u8>)> {
+            self.keys
+                .iter()
+                .map(|(owner, secret)| {
+                    (
+                        *owner,
+                        serde_json::to_vec(secret).expect("serialization should not fail"),
+                    )
+                })
+                .collect()
+        }
+    }
+
+    #[async_trait]
+    impl Signer for InMemorySigner {
+        /// Creates a signature for the given `value` using the provided `owner`.
+        async fn sign(
+            &self,
+            owner: &AccountOwner,
+            value: &CryptoHash,
+        ) -> Result<AccountSignature, Box<dyn std::error::Error>> {
+            let inner = self.0.read().unwrap();
+            if let Some(secret) = inner.keys.get(owner) {
+                let signature = secret.sign_prehash(*value);
+                Ok(signature)
+            } else {
+                Err("No key found for the given owner".into())
+            }
+        }
+
+        /// Returns the public key corresponding to the given `owner`.
+        async fn get_public_key(
+            &self,
+            owner: &AccountOwner,
+        ) -> Result<AccountPublicKey, Box<dyn std::error::Error>> {
+            let inner = self.0.read().unwrap();
+            match inner.keys.get(owner).map(|s| s.public()) {
+                Some(public) => Ok(public),
+                None => Err("No key found for the given owner".into()),
+            }
+        }
+
+        /// Returnes whether the given `owner` is a known signer.
+        async fn contains_key(
+            &self,
+            owner: &AccountOwner,
+        ) -> Result<bool, Box<dyn std::error::Error>> {
+            let inner = self.0.read().unwrap();
+            Ok(inner.keys.contains_key(owner))
+        }
+    }
+
+    impl FromIterator<(AccountOwner, AccountSecretKey)> for InMemorySigner {
+        fn from_iter<T>(input: T) -> Self
+        where
+            T: IntoIterator<Item = (AccountOwner, AccountSecretKey)>,
+        {
+            InMemorySigner(Arc::new(RwLock::new(InMemSignerInner {
+                keys: BTreeMap::from_iter(input),
+                #[cfg(with_getrandom)]
+                rng_state: RngState::new(None),
+            })))
+        }
+    }
+
+    impl Default for InMemSignerInner {
+        fn default() -> Self {
+            #[cfg(with_getrandom)]
+            let signer = InMemSignerInner::new(None);
+            #[cfg(not(with_getrandom))]
+            let signer = InMemSignerInner::new();
+            signer
+        }
+    }
+
+    impl Serialize for InMemorySigner {
+        fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+        where
+            S: serde::Serializer,
+        {
+            let inner = self.0.read().unwrap();
+            InMemSignerInner::serialize(&*inner, serializer)
+        }
+    }
+
+    impl<'de> Deserialize<'de> for InMemorySigner {
+        fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+        where
+            D: serde::Deserializer<'de>,
+        {
+            let inner = InMemSignerInner::deserialize(deserializer)?;
+            Ok(InMemorySigner(Arc::new(RwLock::new(inner))))
+        }
+    }
+
+    impl Serialize for InMemSignerInner {
+        fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+        where
+            S: serde::Serializer,
+        {
+            #[derive(Serialize, Debug)]
+            struct Inner<'a> {
+                keys: &'a Vec<(AccountOwner, Vec<u8>)>,
+                #[cfg(with_getrandom)]
+                prng_seed: Option<u64>,
+            }
+
+            #[cfg(with_getrandom)]
+            let prng_seed = self.rng_state.testing_seed;
+
+            let inner = Inner {
+                keys: &self.keys(),
+                #[cfg(with_getrandom)]
+                prng_seed,
+            };
+
+            Inner::serialize(&inner, serializer)
+        }
+    }
+
+    impl<'de> Deserialize<'de> for InMemSignerInner {
+        fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+        where
+            D: serde::Deserializer<'de>,
+        {
+            #[derive(Deserialize)]
+            struct Inner {
+                keys: Vec<(AccountOwner, Vec<u8>)>,
+                #[cfg(with_getrandom)]
+                prng_seed: Option<u64>,
+            }
+
+            let inner = Inner::deserialize(deserializer)?;
+
+            let keys = inner
+                .keys
+                .into_iter()
+                .map(|(owner, secret)| {
+                    let secret =
+                        serde_json::from_slice(&secret).map_err(serde::de::Error::custom)?;
+                    Ok((owner, secret))
+                })
+                .collect::<Result<BTreeMap<_, _>, _>>()?;
+
+            let signer = InMemSignerInner {
+                keys,
+                #[cfg(with_getrandom)]
+                rng_state: RngState::new(inner.prng_seed),
+            };
+            Ok(signer)
+        }
+    }
+}
--- a/linera-chain/src/data_types.rs
+++ b/linera-chain/src/data_types.rs
@@ -2,15 +2,18 @@
 // Copyright (c) Zefchain Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-use std::collections::{BTreeMap, BTreeSet, HashSet};
+use std::{
+    collections::{BTreeMap, BTreeSet, HashSet},
+    error::Error,
+};
 
 use async_graphql::SimpleObject;
 use custom_debug_derive::Debug;
 use linera_base::{
     bcs,
     crypto::{
-        AccountPublicKey, AccountSecretKey, AccountSignature, BcsHashable, BcsSignable,
-        CryptoError, CryptoHash, ValidatorPublicKey, ValidatorSecretKey, ValidatorSignature,
+        AccountPublicKey, AccountSignature, BcsHashable, BcsSignable, CryptoError, CryptoHash,
+        Signer, ValidatorPublicKey, ValidatorSecretKey, ValidatorSignature,
     },
     data_types::{Amount, Blob, BlockHeight, Epoch, Event, OracleResponse, Round, Timestamp},
     doc_scalar, ensure, hex_debug,
@@ -490,26 +493,34 @@ pub struct ProposalContent {
 }
 
 impl BlockProposal {
-    pub fn new_initial(round: Round, block: ProposedBlock, secret: &AccountSecretKey) -> Self {
+    pub async fn new_initial(
+        owner: AccountOwner,
+        round: Round,
+        block: ProposedBlock,
+        signer: &(impl Signer + ?Sized),
+    ) -> Result<Self, Box<dyn Error>> {
         let content = ProposalContent {
             round,
             block,
             outcome: None,
         };
-        let signature = secret.sign(&content);
-        Self {
+        let signature = signer.sign(&owner, &CryptoHash::new(&content)).await?;
+        let public_key = signer.get_public_key(&owner).await?;
+
+        Ok(Self {
             content,
-            public_key: secret.public(),
+            public_key,
             signature,
             validated_block_certificate: None,
-        }
+        })
     }
 
-    pub fn new_retry(
+    pub async fn new_retry(
+        owner: AccountOwner,
         round: Round,
         validated_block_certificate: ValidatedBlockCertificate,
-        secret: &AccountSecretKey,
-    ) -> Self {
+        signer: &(impl Signer + ?Sized),
+    ) -> Result<Self, Box<dyn Error>> {
         let lite_cert = validated_block_certificate.lite_certificate().cloned();
         let block = validated_block_certificate.into_inner().into_inner();
         let (block, outcome) = block.into_proposal();
@@ -518,13 +529,15 @@ impl BlockProposal {
             round,
             outcome: Some(outcome),
         };
-        let signature = secret.sign(&content);
-        Self {
+        let signature = signer.sign(&owner, &CryptoHash::new(&content)).await?;
+
+        let public_key = signer.get_public_key(&owner).await?;
+        Ok(Self {
             content,
-            public_key: secret.public(),
+            public_key,
             signature,
             validated_block_certificate: Some(lite_cert),
-        }
+        })
     }
 
     pub fn check_signature(&self) -> Result<(), CryptoError> {
--- a/linera-client/src/chain_listener.rs
+++ b/linera-client/src/chain_listener.rs
@@ -14,9 +14,9 @@ use futures::{
     FutureExt as _, StreamExt,
 };
 use linera_base::{
-    crypto::{AccountSecretKey, CryptoHash},
+    crypto::{CryptoHash, Signer},
     data_types::{ChainDescription, Timestamp},
-    identifiers::{BlobType, ChainId},
+    identifiers::{AccountOwner, BlobType, ChainId},
     task::NonBlockingFuture,
 };
 use linera_core::{
@@ -72,10 +72,12 @@ pub trait ClientContext: 'static {
     async fn make_chain_client(&self, chain_id: ChainId)
         -> Result<ContextChainClient<Self>, Error>;
 
+    fn client(&self) -> &linera_core::client::Client<Self::Environment>;
+
     async fn update_wallet_for_new_chain(
         &mut self,
         chain_id: ChainId,
-        key_pair: Option<AccountSecretKey>,
+        owner: Option<AccountOwner>,
         timestamp: Timestamp,
     ) -> Result<(), Error>;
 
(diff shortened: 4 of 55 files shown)


linera-base/src/crypto/mod.rs
```diff
@@ -213,6 +229,58 @@ impl TryFrom<&[u8]> for AccountSignature {
...
+        AccountSecretKey::public(self)
+    }
+
+    #[cfg(all(with_testing, with_getrandom))]
+    fn generate_new() -> Self {
+        AccountSecretKey::generate()
+    }
+}
+
+impl SigningKey for &AccountSecretKey {
```
Twey, line 267: I'm a bit confused about the purpose of this impl.  All the available methods (apart from the one that's been stubbed out) are already `&self`, so I think we shouldn't gain any power by adding an impl for `&AccountSecretKey` — and insofar as we do want such a thing it can be written as a blanket `impl<K: SigningKey> SigningKey for &K`.

linera-base/src/crypto/mod.rs
```diff
@@ -71,6 +72,15 @@ pub enum AccountSecretKey {
...
 
+impl Clone for AccountSecretKey {
+    fn clone(&self) -> Self {
+        match self {
+            AccountSecretKey::Ed25519(secret) => AccountSecretKey::Ed25519(secret.copy()),
+            AccountSecretKey::Secp256k1(secret) => AccountSecretKey::Secp256k1(secret.copy()),
+        }
+    }
+}
+
```
Twey, line 83: It was a deliberate choice not to provide `Clone` for secret keys in the codebase.  I don't have strong opinions about it either way, but I'd suggest having a discussion.

deuszx: I know but if it wasn't needed by the _already-existing_ code I wouldn't have added it. Right now I'm just trying to get a green CI with the new abstractions but without changing anything else.

linera-base/src/crypto/mod.rs
```diff
@@ -11,6 +11,7 @@ mod secp256k1;
 use std::{fmt::Display, io, num::ParseIntError, str::FromStr};
 
 use alloy_primitives::FixedBytes;
+use async_trait::async_trait;
```
Twey, line 14: We support async traits natively in Rust now, so this mostly isn't needed any more.  Instead, try `trait-variant` for just the functionality of conditionally providing a `Send`/`?Send` definition of the trait.

linera-base/src/crypto/mod.rs
```diff
@@ -213,6 +229,58 @@ impl TryFrom<&[u8]> for AccountSignature {
     }
 }
 
+/// A trait for signing keys.
+#[cfg_attr(not(web), async_trait)]
+#[cfg_attr(web, async_trait(?Send))]
+pub trait SigningKey {
```
Twey, line 235: I do suggest that this be named `Signer` or some other name that doesn't imply that it actually contains a key.

linera-base/src/crypto/mod.rs
```diff
@@ -213,6 +229,58 @@ impl TryFrom<&[u8]> for AccountSignature {
...
+    }
+
+    fn public(&self) -> AccountPublicKey {
+        AccountSecretKey::public(self)
+    }
+
+    #[cfg(all(with_testing, with_getrandom))]
+    fn generate_new() -> Self {
+        AccountSecretKey::generate()
+    }
```
Twey, line 264: Rather than including this in this trait, we can provide an implementation of `StandardUniform: Distribution<AccountSecretKey>`.

deuszx: I don't think that'll work as the currently-existing code (that operates on `AccountSecret`) is expecting the type to have this method.

We can do that but that's orthogonal to this change.

linera-base/src/crypto/mod.rs
```diff
@@ -213,6 +229,58 @@ impl TryFrom<&[u8]> for AccountSignature {
...
 
+/// A trait for signing keys.
+#[cfg_attr(not(web), async_trait)]
+#[cfg_attr(web, async_trait(?Send))]
+pub trait SigningKey {
+    /// The public key type corresponding to this signing key.
+    type PublicKey: Into<AccountPublicKey> + std::fmt::Debug;
+
+    /// Signs the given value and returns the signature.
+    fn sign<'a, 'b, A: BcsSignable<'b>>(&'a self, value: &A) -> AccountSignature;
```
Twey, line 240: In general, I think that signing may update the internal state of the signer, so I'd be inclined to make this `&mut self`.

In the interest of keeping the concerns separate, I think it might be better to take the bytes here directly, so that this trait remains independent of `BcsSignable`.

linera-client/src/client_context.rs
```diff
@@ -81,45 +84,54 @@ where
 
 #[cfg_attr(not(web), async_trait)]
 #[cfg_attr(web, async_trait(?Send))]
-impl<S, W> chain_listener::ClientContext for ClientContext<S, W>
+impl<S, W, K> chain_listener::ClientContext for ClientContext<S, W, K>
```
Twey, line 87: This PR becomes _a lot_ smaller if we use a trait object here rather than making it part of the type.

linera-base/src/crypto/mod.rs
```diff
@@ -213,6 +229,58 @@ impl TryFrom<&[u8]> for AccountSignature {
     }
 }
 
+/// A trait for signing keys.
+#[cfg_attr(not(web), async_trait)]
+#[cfg_attr(web, async_trait(?Send))]
+pub trait SigningKey {
+    /// The public key type corresponding to this signing key.
+    type PublicKey: Into<AccountPublicKey> + std::fmt::Debug;
```
Twey, line 237: I think this is an antipattern: the things that are `Into<AccountPublicKey>` are roughly the contents of the `AccountPublicKey` enum, so this associated type doesn't add much information over just returning an `AccountPublicKey` (as I see you have, indeed!).

linera-base/src/crypto/mod.rs
Twey: Also, since the wallet takes only one signer, the signer must support signing for multiple identities, since different chains may be signed by different keys, so this should take as parameter the identity we want to sign for.  (Alternatively the wallet could take a `HashMap<Identity, Box<dyn SigningKey>>` instead, but I think that's a less elegant solution.)

linera-base/src/crypto/mod.rs
Twey: Oh I missed that it already exists — no worries then.

linera-base/src/crypto/mod.rs
Twey: No I think using `AccountPublicKey` is exactly right.  I think the associated `PublicKey` type is what isn't necessary, and it makes it harder to make a trait object in the `ClientContext` later on.

linera-base/src/crypto/mod.rs
deuszx: Done - although, force-pushed.

linera-base/src/crypto/mod.rs
deuszx: Renamed.

linera-base/src/crypto/mod.rs
deuszx: Removed.

linera-base/src/crypto/mod.rs
deuszx: > In the interest of keeping the concerns separate, I think it might be better to take the bytes here directly, so that this trait remains independent of `BcsSignable` (and, for example, we don't force redundant reserialization if the caller already has serialized bytes available).

Definitely a good point - although, as previously, this API is compatible with the already-existing `AccountSecretKey` type (which in turn was made compatible with the crypto we have and how we use `BcsSignable` types).

All in all, good point but orthogonal to this PR.

deuszx: > In general, I think that signing may update the internal state of the signer, so I'd be inclined to make this `&mut self`. We can always strengthen it later to `&self` if it turns out nothing does this, but it's much harder to go the other way.

I can't think of an example where that'd be the case. Hopefully that'll be never true as intuitively it could be dangerous - consective signatures shouldn't rely on previous ones as that could lead some secret (metadata).

linera-base/src/crypto/mod.rs
Twey: I'm a bit confused how it could be needed by the existing code, since as far as I know there previously wasn't such an implementation for `AccountSecretKey` (and I can see you've removed some `copy()` calls below that are how we used to copy these).

deuszx: We used to `.copy()` [here](https://github.com/linera-io/linera-protocol/blob/main/linera-core/src/client/mod.rs#L1025).

deuszx: This `Clone` impl is now removed.

linera-client/src/client_context.rs
deuszx: Updated.

linera-base/src/crypto/signer.rs
```diff
@@ -0,0 +1,178 @@
...
+    fn from(input: T) -> Self {
+        InMemSigner(BTreeMap::from_iter(input))
+    }
+}
+
+impl Default for InMemSigner {
+    fn default() -> Self {
+        InMemSigner::new()
+    }
+}
```
Twey, line 111: `#[derive(Default)]`

linera-base/src/crypto/signer.rs
```diff
@@ -0,0 +1,178 @@
...
+
+impl Clone for InMemSigner {
+    fn clone(&self) -> Self {
+        let mut map = BTreeMap::new();
+        for (owner, secret) in self.0.iter() {
+            map.insert(*owner, secret.copy());
+        }
+        InMemSigner(map)
+    }
+}
```
Twey, line 121: I can't help but feel that if we have `Clone` for this we might as well just have `Clone` for `AccountSecretKey` (and then this is `#[derive(Clone)]`.  I know @ma2bd has strong feelings about this so I'll defer.

linera-chain/src/data_types.rs
```diff
@@ -614,16 +614,19 @@ pub struct ProposalContent {
 }
 
 impl BlockProposal {
-    pub fn new_initial(round: Round, block: ProposedBlock, secret: &AccountSecretKey) -> Self {
+    pub fn new_initial(round: Round, block: ProposedBlock, signer: &Box<dyn Signer>) -> Self {
```
Twey, line 617: Clippy is right, I'd even say `&impl Signer`.

deuszx: Done, albeit force-pushed.

linera-base/src/crypto/signer.rs
```diff
@@ -0,0 +1,178 @@
...
+    pub fn new() -> Self {
+        InMemSigner(BTreeMap::new())
+    }
+}
+
+impl<T: IntoIterator<Item = (AccountOwner, AccountSecretKey)>> From<T> for InMemSigner {
+    fn from(input: T) -> Self {
+        InMemSigner(BTreeMap::from_iter(input))
+    }
+}
```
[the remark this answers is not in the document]
deuszx: Personally, I prefer this slightly more explicit way (i.e. we see right away that the target type is `BTreeMap`.

linera-base/src/crypto/signer.rs
deuszx: Got rid of this.

linera-base/src/crypto/signer.rs
```diff
@@ -0,0 +1,97 @@
...
+
+impl Signer for Box<dyn Signer> {
+    #[cfg(with_getrandom)]
+    fn generate_new(&mut self) -> AccountPublicKey {
+        (**self).generate_new()
+    }
+
+    fn sign(&self, owner: &AccountOwner, value: &CryptoHash) -> Option<AccountSignature> {
```
Twey, line 68: I'm not sure I understand why this takes `CryptoHash` instead of `&[u8]`.  Could you add a comment?

linera-chain/src/data_types.rs
```diff
@@ -612,16 +612,23 @@ pub struct ProposalContent {
...
 impl BlockProposal {
-    pub fn new_initial(round: Round, block: ProposedBlock, secret: &AccountSecretKey) -> Self {
+    pub fn new_initial(
+        round: Round,
+        block: ProposedBlock,
+        signer: &(impl Signer + ?Sized),
+    ) -> Self {
+        let owner = block
+            .authenticated_signer
+            .expect("Block author is required for new block proposals");
```
Twey, line 622: Since we're changing the signature of this function anyway, it would be nice to encode this in the type system by taking an `AccountOwner` as parameter.

linera-client/src/config.rs
```diff
@@ -148,7 +142,9 @@ impl<W: Persist<Target = Wallet>> Persist for WalletState<W> {
 
 #[cfg(feature = "fs")]
 impl WalletState<persistent::File<Wallet>> {
-    pub fn create_from_file(path: &std::path::Path, wallet: Wallet) -> Result<Self, Error> {
+    /// Reads the wallet from the given path, creating it if it does not exist.
+    /// The wallet is created with the given `wallet` value.
+    pub fn read_or_create(path: &std::path::Path, wallet: Wallet) -> Result<Self, Error> {
```
Twey, line 147: My bad, thanks for fixing this :)

linera-service/src/linera/main.rs
```diff
@@ -133,11 +136,11 @@ impl Runnable for Job {
...
-                    Some(owner) => (owner, None),
+                let new_owner = match owner {
+                    Some(owner) => owner,
                     None => {
-                        let key_pair = context.wallet.generate_key_pair();
-                        (key_pair.public().into(), Some(key_pair))
+                        let public_key = context.signer.generate_new();
+                        public_key.into()
                     }
                 };
```
Twey, line 145: `new_owner = owner.unwrap_or_else(|| context.signer.generate_new().into())`

linera-client/src/client_context.rs
```diff
@@ -317,24 +332,24 @@ where
     pub async fn update_wallet_for_new_chain(
         &mut self,
         chain_id: ChainId,
-        key_pair: Option<AccountSecretKey>,
+        owner: Option<AccountOwner>,
```
Twey, line 335: I think this should now be just non-optional `AccountOwner` (and down the callstack).  I don't think it makes sense to have a chain without an owner.

linera-client/src/persistent/dirty.rs
```diff
@@ -2,11 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #[derive(Clone, Debug, PartialEq, Eq, derive_more::Deref, derive_more::DerefMut)]
-pub struct Dirty(
-    #[deref]
-    #[deref_mut]
-    bool,
-);
+pub struct Dirty(#[deref] bool);
```
Twey, line 5: What happened to `deref_mut`?

linera-core/src/client/chain_client_state.rs
```diff
@@ -97,21 +89,21 @@ impl ChainClientState {
         }
     }
 
-    pub fn known_key_pairs(&self) -> &BTreeMap<AccountOwner, AccountSecretKey> {
-        &self.known_key_pairs
+    #[allow(clippy::borrowed_box)]
+    pub fn signer(&self) -> &Box<dyn Signer> {
+        &self.signer
+    }
```
Twey, line 95: I think `&dyn Signer` is the way here?

linera-core/src/client/mod.rs
```diff
@@ -283,7 +285,7 @@ impl<P, S: Storage + Clone> Client<P, S> {
     pub fn create_chain_client(
         self: &Arc<Self>,
         chain_id: ChainId,
-        known_key_pairs: Vec<AccountSecretKey>,
+        signer: Box<dyn Signer>,
```
Twey, line 288: Since we take an owner in the `Signer` trait, I think that we can hoist this right up into `Client`?  Then we shouldn't need to clone the `Signer` (modulo tests that re-use the signer for different clients).

linera-core/src/client/signer.rs
```diff
@@ -0,0 +1,168 @@
...
+use std::collections::BTreeMap;
+
+use linera_base::{
+    crypto::{AccountPublicKey, AccountSecretKey, AccountSignature, CryptoHash, CryptoRng, Signer},
+    identifiers::AccountOwner,
+};
+use serde::{Deserialize, Serialize};
+
+/// In-memory signer.
+pub struct InMemSigner {
```
Twey, line 13: How did this end up in `linera-core` while the trait is in `linera-base`? :thinking: I'd expect these to be next to each other.

linera-core/src/client/signer.rs
```diff
@@ -0,0 +1,168 @@
...
+    prng: Box<dyn CryptoRng>,
+    keys: BTreeMap<AccountOwner, AccountSecretKey>,
+}
+
+impl InMemSigner {
+    /// Creates a new `InMemSigner`.
+    pub fn new(prng_seed: Option<u64>) -> Self {
+        let prng: Box<dyn CryptoRng> = prng_seed.into();
+        InMemSigner {
+            prng_seed,
```
Twey, line 25: I think it's only necessary to store the `prng`, not the `prng_seed`.  When we create a new `InMemSigner` we can create the `prng` from the `prng_seed`, and when we save the state we generate a new `prng_seed` from the `prng` (this gives us forward secrecy).

linera-core/src/client/signer.rs
```diff
@@ -0,0 +1,168 @@
...
+use linera_base::{
+    crypto::{AccountPublicKey, AccountSecretKey, AccountSignature, CryptoHash, CryptoRng, Signer},
+    identifiers::AccountOwner,
+};
+use serde::{Deserialize, Serialize};
+
+/// In-memory signer.
+pub struct InMemSigner {
+    prng_seed: Option<u64>,
+    keys_generated: u64,
```
Twey, line 15: I'm not sure I understand the purpose of this field.

linera-base/src/crypto/signer.rs
```diff
@@ -0,0 +1,97 @@
...
+    fn get_public(&self, owner: &AccountOwner) -> Option<AccountPublicKey>;
+
+    /// Returnes whether the given `owner` is a known signer.
+    fn contains_key(&self, owner: &AccountOwner) -> bool;
+
+    /// Removes the key for the given `owner`.
+    fn remove(&mut self, owner: &AccountOwner) -> bool;
+
+    /// Returns a clone of the `Signer` as a boxed trait object.
+    fn clone_box(&self) -> Box<dyn Signer>;
```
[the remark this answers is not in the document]
Twey: In fact we might not need `Clone` at all — see below.

linera-base/src/crypto/signer.rs
```diff
@@ -0,0 +1,97 @@
...
+    fn sign(&self, owner: &AccountOwner, value: &CryptoHash) -> Option<AccountSignature>;
+
+    /// Returns the public key corresponding to the given `owner`.
+    fn get_public(&self, owner: &AccountOwner) -> Option<AccountPublicKey>;
+
+    /// Returnes whether the given `owner` is a known signer.
+    fn contains_key(&self, owner: &AccountOwner) -> bool;
+
+    /// Removes the key for the given `owner`.
+    fn remove(&mut self, owner: &AccountOwner) -> bool;
```
Twey, line 47: I guess that these are temporary, for easing transition from the old API?  A `TODO` would be nice if so.

linera-client/src/client_options.rs
```diff
@@ -44,6 +44,10 @@ pub struct ClientContextOptions {
     #[arg(long = "wallet")]
     pub wallet_state_path: Option<PathBuf>,
 
+    /// Sets the file storing the keystore state.
+    #[arg(long = "keystore")]
```
Twey, line 48: As a small but significant naming change, I think that after this split what you've called ‘keystore’ is what other people like to call ‘wallet’, and what we're now calling ‘wallet’ is actually some kind of client state+configuration.

linera-base/src/crypto/signer.rs
```diff
@@ -0,0 +1,97 @@
...
+    pub fn new(bytes: Vec<u8>) -> Self {
+        SignableBytes(bytes)
+    }
+
+    /// Returns the inner bytes.
+    pub fn as_bytes(&self) -> &[u8] {
+        &self.0
+    }
+}
+impl BcsSignable<'_> for SignableBytes {}
```
ma2bd, line 27: What is this for? This is defeating the purpose of `BcsSignable`.

linera-base/src/crypto/signer.rs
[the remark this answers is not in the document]
Twey: Sorry, I wasn't suggesting that we replace `BTreeMap::from_iter(input)` with `input.collect()` (I don't have strong feelings about this), I was suggesting that we replace `<T: IntoIterator> From<T> for InMemSigner` with `FromIterator for InMemSigner` ([some prior discussion](https://users.rust-lang.org/t/why-is-fromiterator-necessary-is-from-into-iterator-not-good-enough/119505/4)).

linera-client/src/persistent/dirty.rs
deuszx: ups! I think I introduced `deref_mut` intermittently in this PR, then removed it and went too far with this.

linera-base/src/crypto/signer.rs
deuszx: Yes, this should be removed now - i.e. the whole `SignableBytes` struct.

linera-client/src/client_context.rs
```diff
@@ -866,14 +881,14 @@ where
         num_new_chains: usize,
         chain_client: &ChainClient<NodeProvider, S>,
         balance: Amount,
-        key_pair: &AccountSecretKey,
+        pub_key: &AccountPublicKey,
```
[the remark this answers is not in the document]
Twey: You mean you can't recover the pubkey from the `AccountOwner`?  You've retrieved the `AccountOwner` from the pubkey a few lines down :)

This function doesn't do anything with the pubkey though, the only use of `pub_key` is to convert it into an `AccountOwner`.

linera-base/src/crypto/signer.rs
```diff
@@ -0,0 +1,97 @@
...
+    fn keys(&self) -> Vec<(AccountOwner, Vec<u8>)> {
+        (**self).keys()
+    }
+}
+
+// impl Default for Box<dyn Signer> {
+//     fn default() -> Self {
+//         Box::new(InMemSigner::new(None))
+//     }
+// }
```
deuszx, line 97: Leftover

linera-chain/src/data_types.rs
deuszx: Account owner here is taken from the already-constructed block where this field is optional.

linera-core/src/client/signer.rs
deuszx: Me neither anymore.

linera-core/src/client/signer.rs
deuszx: Actually - `Box<dyn CryptoRng>` is not cloneable and in `Clone for InMemSigner` I don't have the reference to `&mut self` (just `&self`) so I can't generate the _next_ `u64` seed for the clone of `InMemSigner`.

linera-core/src/client/signer.rs
deuszx: Actually, I do now. See https://github.com/linera-io/linera-protocol/pull/3696#discussion_r2035877539

linera-client/src/config.rs
```diff
@@ -174,14 +170,75 @@ impl WalletState<persistent::IndexedDb<Wallet>> {
 
 impl<W: Deref<Target = Wallet>> WalletState<W> {
     pub fn new(wallet: W) -> Self {
-        Self {
-            prng: wallet.make_prng(),
-            wallet,
-        }
+        Self { wallet }
+    }
+}
```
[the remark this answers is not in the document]
deuszx: oh, I thought it was also some kind of abstraction around persistable `Wallet`.

linera-client/src/client_context.rs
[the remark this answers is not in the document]
deuszx: You're right. I must have been thinking about some other place where I made a similar change.

linera-base/src/crypto/signer.rs
deuszx: Added the comment on the `Signer` trait.

linera-client/src/client_context.rs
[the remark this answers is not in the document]
deuszx: Fixed.

linera-client/src/persistent/dirty.rs
deuszx: Reverted.

linera-core/src/client/chain_client_state.rs
deuszx: Changed.

linera-base/src/crypto/signer.rs
deuszx: Removed.

linera-client/src/client_options.rs
deuszx: To be honest, I don't think there's a fully-agreed upon dictionary here. Some people call the whole "blockchain" as _ledger_/wallet. I think we can fix the names later.

linera-base/src/crypto/signer.rs
deuszx: I found that I don't need the _remove_ per-se and I can replace it with `contains_key` for now.

linera-base/src/crypto/signer.rs
```diff
@@ -0,0 +1,97 @@
...
+    fn contains_key(&self, owner: &AccountOwner) -> bool;
+
+    /// Removes the key for the given `owner`.
+    fn remove(&mut self, owner: &AccountOwner) -> bool;
+
+    /// Returns a clone of the `Signer` as a boxed trait object.
+    fn clone_box(&self) -> Box<dyn Signer>;
+
+    /// Returns an iterator over the keys in the signer.
+    fn keys(&self) -> Vec<(AccountOwner, Vec<u8>)>;
```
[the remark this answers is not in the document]
deuszx: Yes, this is just so that I can persist the `InMemSigner` for now.

linera-base/src/crypto/signer.rs
```diff
@@ -0,0 +1,97 @@
...
+    }
+}
+impl BcsSignable<'_> for SignableBytes {}
+
+/// A trait for signing keys.
+#[trait_variant::make(Send + Sync)]
+pub trait Signer {
+    /// Generates a new signing key for the Self type.
+    /// New secret key is inserted into Signer's memory and the `AccountPublicKey` is returned.
+    #[cfg(with_getrandom)]
```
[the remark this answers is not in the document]
deuszx: Not only that - in `Signer for InMemSigner` we call into `AccountSecretKey::generate_from` which is also feature-flagged.

linera-base/src/crypto/signer.rs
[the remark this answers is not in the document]
deuszx: Moved to `impl InMemSigner` - it's enough to have it there.

linera-base/src/crypto/signer.rs
deuszx: Removed the `remove`.

linera-core/src/client/signer.rs
deuszx: The reason I didn't do it initially was that it requires lots of feature-flagging the `CryptoRng` usage b/c `linera-base` is dependency for web crates.

I will move it back and add those feature flags though.

deuszx: Done

linera-base/src/crypto/signer.rs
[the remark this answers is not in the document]
deuszx: Done

linera-base/src/crypto/signer.rs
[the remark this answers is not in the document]
deuszx: I'd prefer to avoid magic code like this. It wasn't difficult (or long) to implement this by hand.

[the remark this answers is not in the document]
Twey: It's not magic, it's exactly what you've done here, just standardized under a well-known name so the reader doesn't have to piece together the intention, and factored into a separate base trait to keep the separation of concerns :)

linera-base/src/crypto/signer.rs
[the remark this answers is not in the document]
Twey: I think it needn't be?  Since you have a RNG.

linera-core/src/client/signer.rs
Twey: Eek.  I guess that if we need to keep `Clone for InMemSigner` we should probably `Arc<Mutex>` the RNG.

linera-core/src/client/signer.rs
Twey: `linera-core` is also a dependency for Web crates :) (but not on-chain builds)

linera-client/src/client_context.rs
Twey: And non-user chains might have a more complicated ownership structure than just one `AccountOwner`, right.

linera-base/src/crypto/signer.rs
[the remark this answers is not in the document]
deuszx: `CryptoRng` is feature-gated. `Ed25519::generate_key` was feature-gated even before "my times". I didn't want to change that.

linera-core/src/client/signer.rs
Twey: Actually — what you're doing here is just as crypto-unsafe as `Clone`!  The seed is now shared between different chains which makes the output predictable.  We _must_ add a mutex either in the RNG or (better IMO) the `Client` so we can mutably share the RNG state.

linera-core/src/client/signer.rs
deuszx: Possibly, but if we support keys' removal then no.

linera-core/src/client/signer.rs
deuszx: Yes, I agree. `RNG` needs to be Mutex'ed and the `keys` behind `Arc`. 

I will be fixing that as soon as I fix the other error I'm hitting in CI.

linera-base/src/crypto/signer.rs
```diff
@@ -109,9 +111,9 @@ mod in_mem {
 
     /// In-memory signer.
     pub struct InMemSigner {
-        keys: BTreeMap<AccountOwner, AccountSecretKey>,
+        keys: Arc<BTreeMap<AccountOwner, AccountSecretKey>>,
         #[cfg(with_getrandom)]
```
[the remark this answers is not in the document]
deuszx: I'm not sure I understand. I cannot `use crypto::CryptoRng` in the web client (i.e. anything behind the `cfg(with_getrandom)` is not available in those builds) so I have to feature-gate the code that uses it - imports, new key-generations etc. 

All our crypto also feature-gates the access to new key generation, so this requirements is inherited.

linera-base/src/crypto/signer.rs
deuszx: Cloning of `Signer` is no longer necessary.

linera-base/src/crypto/signer.rs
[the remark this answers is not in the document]
deuszx: Got rid of this method entirely.

linera-base/src/crypto/signer.rs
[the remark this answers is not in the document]
deuszx: key generation is now a method available on `InMemSigner` type, not the `Signer` trait.

linera-base/src/crypto/signer.rs
```diff
@@ -0,0 +1,97 @@
...
+}
+impl BcsSignable<'_> for SignableBytes {}
+
+/// A trait for signing keys.
+#[trait_variant::make(Send + Sync)]
+pub trait Signer {
+    /// Generates a new signing key for the Self type.
+    /// New secret key is inserted into Signer's memory and the `AccountPublicKey` is returned.
+    #[cfg(with_getrandom)]
+    fn generate_new(&mut self) -> AccountPublicKey;
```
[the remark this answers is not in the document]
deuszx: Removed from `Signer`.

linera-core/src/client/mod.rs
deuszx: Moved to `Client`.

linera-chain/src/data_types.rs
deuszx: This was incorrect - `block.authenticated_signer` was allowed to be `None`. I modified the constructors of `BlockProposal` to take the proposer's `AccountOwner` value explicitly. All tests have been updated.

ma2bd: The None authenticated signer was allowed on purpose

ma2bd: You could call that an "unauthenticated block"

deuszx: Yes, I learned that later on. This change is no longer part of the latest diff.

linera-core/src/client/signer.rs
deuszx: This should be fixed now. The (inner) state of the `InMemSigner` is wrapped with `Arc<RwLock<InMemSignerInner>>`.

linera-base/src/crypto/signer.rs
```diff
@@ -0,0 +1,282 @@
...
+                #[cfg(with_getrandom)]
+                testing_seed: prng_seed,
+            }
+        }
+    }
+
+    impl InMemSignerInner {
+        /// Creates a new `InMemSignerInner` seeded with `prng_seed`.
+        /// If `prng_seed` is `None`, an `OsRng` will be used.
+        #[cfg(with_getrandom)]
```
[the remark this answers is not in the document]
deuszx:
```rust
#[cfg(with_getrandom)]
impl From<Option<u64>> for Box<dyn CryptoRng> {
    fn from(seed: Option<u64>) -> Self {
        use rand::SeedableRng;

        match seed {
            Some(seed) => Box::new(rand::rngs::StdRng::seed_from_u64(seed)),
            None => Box::new(rand::rngs::OsRng),
        }
    }
}
```
This is what requires the flag.

linera-base/src/crypto/signer.rs
```diff
@@ -0,0 +1,282 @@
...
+    struct InMemSignerInner {
+        keys: BTreeMap<AccountOwner, AccountSecretKey>,
+        #[cfg(with_getrandom)]
+        rng_state: RngState,
+    }
+
+    #[cfg(with_getrandom)]
+    struct RngState {
+        prng: Box<dyn CryptoRng>,
+        #[cfg(with_getrandom)]
```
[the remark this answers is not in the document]
deuszx: Acutally, I cannot do that - i.e. `RngState` has to be conditionally-compiled. `trait CryptoRng` is behind a feature flag (`with_getrandom`) so the compilation of `linera-base` , when it's a dependency in contracts, will fail as this type is not available there.

linera-base/src/crypto/signer.rs
```diff
@@ -0,0 +1,282 @@
...
+    #[cfg(with_getrandom)]
+    use crate::crypto::CryptoRng;
+    use crate::{
+        crypto::{AccountPublicKey, AccountSecretKey, AccountSignature, CryptoHash, Signer},
+        identifiers::AccountOwner,
+    };
+
+    /// In-memory signer.
+    #[derive(Clone)]
+    pub struct InMemSigner(Arc<RwLock<InMemSignerInner>>);
```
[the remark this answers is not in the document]
deuszx: Done in [Rename, comment update.](https://github.com/linera-io/linera-protocol/pull/3696/commits/9aabaa78bd12c7fc84639216ed89e061ca9b6fea).

linera-base/src/crypto/mod.rs
```diff
@@ -133,11 +135,35 @@ impl AccountSecretKey {
         }
     }
 
+    /// Creates a signature for the `value` using provided `secret`.
```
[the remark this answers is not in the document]
deuszx: Done in [Rename, comment update.](https://github.com/linera-io/linera-protocol/pull/3696/commits/9aabaa78bd12c7fc84639216ed89e061ca9b6fea).

linera-base/src/crypto/signer.rs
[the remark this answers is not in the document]
Twey: Why is `CryptoRng` behind `with_getrandom`? :thinking: That feels a little odd.

linera-base/src/crypto/signer.rs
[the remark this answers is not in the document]
Twey: What I mean to say is, it's only the `None` branch that requires `getrandom`.  So we can remove the conditional compilation on `CryptoRng` and replace this function with `new(rng: impl CryptoRng)` to save a whole bunch of conditional compilation.

linera-base/src/crypto/signer.rs
```diff
@@ -0,0 +1,315 @@
...
+    impl Serialize for InMemSignerInner {
+        fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+        where
+            S: serde::Serializer,
+        {
+            #[derive(Serialize, Debug)]
+            struct Inner<'a> {
+                keys: &'a Vec<(AccountOwner, Vec<u8>)>,
+                #[cfg(with_getrandom)]
+                prng_seed: Option<u64>,
```
Twey, line 267: I think since we're not trying to allow recreation of seed states we don't need to store this one, and can just generate a new seed from the RNG when we save :thinking:

linera-client/src/chain_listener.rs
```diff
@@ -406,7 +417,9 @@ impl<C: ClientContext> ChainListener<C> {
             .process_inbox_without_prepare()
             .await
         {
-            Err(ChainClientError::CannotFindKeyForChain(_)) => {}
+            Err(ChainClientError::CannotFindKeyForChain(chain_id)) => {
+                debug!(%chain_id, "Cannot find key for chain");
```
Twey, line 421: It might be better to replace this with `err(warn)` instrumentation on `process_inbox_without_prepare`.

linera-base/src/crypto/signer.rs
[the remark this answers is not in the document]
Twey: Discussed offline: this is probably suboptimal, but stems from `CryptoRng` being conditionally compiled, which was not part of this work.

linera-service/tests/local_net_tests.rs
```diff
@@ -878,12 +883,13 @@ async fn test_wasm_end_to_end_ethereum_tracker(config: impl LineraNetConfig) ->
     let owner1 = {
         let wallet = client.load_wallet()?;
         let user_chain = wallet.get(chain).unwrap();
-        user_chain.key_pair.as_ref().unwrap().public().into()
+        *user_chain.owner.as_ref().unwrap()
```
Twey, line 886: Nit: `user_chain.owner.copied().unwrap()`

Twey (approved):
Some comments, but I'm pretty happy with this.  Looks good!

Some additional work will be needed when binding for the Web since I think we need to be more careful about our `Send`/`Sync`ness.  But this is a great start.

linera-client/src/chain_listener.rs
afck: I originally decided to just ignore this error, because it can happen that one of our chains changes ownership, and then we're not responsible for processing its inbox anymore.

But maybe we should detect it after all, and also remove it from `tracked_chains` in that case.

PR #3699, opened 2025-04-01
tentative fix

ma2bd:
## Motivation

We're seeing some issues in e2e tests with newly created chains.

## Proposal

Fix the suspicious code

## Test Plan

CI +  manual test TBD

## Release Plan

- These changes should be backported to the latest `main` branch

### DIFF
--- a/linera-core/src/chain_worker/state/attempted_changes.rs
+++ b/linera-core/src/chain_worker/state/attempted_changes.rs
@@ -496,7 +496,7 @@ where
         &mut self,
         latest_heights: Vec<(Target, BlockHeight)>,
     ) -> Result<(), WorkerError> {
-        let mut height_with_fully_delivered_messages = BlockHeight::ZERO;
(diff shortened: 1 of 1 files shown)


afck (approved):
Good catch!!

linera-core/src/chain_worker/state/attempted_changes.rs
```diff
@@ -496,7 +496,7 @@ where
         &mut self,
         latest_heights: Vec<(Target, BlockHeight)>,
     ) -> Result<(), WorkerError> {
-        let mut height_with_fully_delivered_messages = BlockHeight::ZERO;
+        let mut height_with_fully_delivered_messages = None;
```
afck, line 499: (Alternatively: `next_delivery_height: BlockHeight::ZERO`, and add `1` below.)

ma2bd: But we would to subtract 1 as well.

PR #3701, opened 2025-04-02
Pass CertificateValue to Vote::new. Get rid of extra hashing.

deuszx:
## Motivation

Andreas noticed that we're unnecessarily hashing `ConfirmedBlock` before creating a `Vote` for it. Intuitiveily it shouldn't be necessary - `ConfirmedBlock` already contains `Hashed<Block>` and when creating a vote for it we're adding the certificate's kind as well.

## Proposal

Change the code to operate on `T: CertificateValue` rather than `Hashed<T>`. All `CertificateValue` impls are required now to also have a `fn hash() -> CryptoHash`.

## Test Plan

CI

## Release Plan

## Links

### DIFF
--- a/linera-chain/src/certificate/generic.rs
+++ b/linera-chain/src/certificate/generic.rs
@@ -139,6 +139,5 @@ impl<T: CertificateValue + Eq + PartialEq> PartialEq for GenericCertificate<T> {
         self.hash() == other.hash()
             && self.round == other.round
             && self.signatures == other.signatures
-        // && self::KIND == other::KIND
     }
 }
--- a/linera-chain/src/certificate/generic.rs
+++ b/linera-chain/src/certificate/generic.rs
@@ -6,7 +6,6 @@ use custom_debug_derive::Debug;
 use linera_base::{
     crypto::{CryptoHash, ValidatorPublicKey, ValidatorSignature},
     data_types::Round,
-    hashed::Hashed,
 };
 use linera_execution::committee::Committee;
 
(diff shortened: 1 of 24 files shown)


linera-chain/src/certificate/generic.rs
```diff
@@ -37,37 +36,31 @@ impl<T> GenericCertificate<T> {
     }
 
     /// Returns a reference to the `Hashed` value contained in this certificate.
```
afck, line 38:
```suggestion
    /// Returns a reference to the value contained in this certificate.
```

linera-chain/src/certificate/generic.rs
```diff
@@ -139,12 +132,13 @@ impl<T: Clone> Clone for GenericCertificate<T> {
...
-impl<T: Eq + PartialEq> Eq for GenericCertificate<T> {}
+impl<T: CertificateValue + Eq + PartialEq> Eq for GenericCertificate<T> {}
 #[cfg(with_testing)]
-impl<T: Eq + PartialEq> PartialEq for GenericCertificate<T> {
+impl<T: CertificateValue + Eq + PartialEq> PartialEq for GenericCertificate<T> {
     fn eq(&self, other: &Self) -> bool {
         self.hash() == other.hash()
             && self.round == other.round
             && self.signatures == other.signatures
+        // && self::KIND == other::KIND
```
afck, line 142: Does this comment mean that the type system already implies that the kinds match?

afck (approved):

linera-chain/src/certificate/generic.rs
deuszx, line 142:
```suggestion
```

deuszx: Removed in [RM commented-out line.](https://github.com/linera-io/linera-protocol/pull/3701/commits/3b22e9f6e34afdf2199e22e8809bcf925a5c4fd7)

PR #3702, opened 2025-04-02
Extract some logic from `select` in the chain listener.

afck:
## Motivation

The `select!` statement in the chain listener is very long and hard to read. It is also counterintuitive that we set `timeout` to "now" to process the inbox. (Thank Mateusz for pointing this out!)

## Proposal

Extract processing the inbox from the code, and call the new function instead of setting the timeout to now.

To remove some more code from the macro, filter the admin chain notifications using `StreamExt::filter`.

## Test Plan

CI

## Release Plan

- 
## Links

### DIFF
--- a/linera-client/src/chain_listener.rs
+++ b/linera-client/src/chain_listener.rs
@@ -278,6 +278,13 @@ impl ChainListener {
         }
     }
 
+    /// Processes the inbox, unless `skip_process_inbox` is set.
+    ///
+    /// If no block can be produced because we are not the round leader, a timeout is returned
+    /// for when to retry; otherwise `u64::MAX` is returned.
+    ///
+    /// The wallet is persisted with any blocks that processing the inbox added. An error
+    /// is returned if persisting the wallet fails.
     async fn maybe_process_inbox<C>(
         config: &ChainListenerConfig,
         client: &ChainClient<C::ValidatorNodeProvider, C::Storage>,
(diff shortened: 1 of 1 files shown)


linera-client/src/chain_listener.rs
```diff
@@ -167,12 +167,15 @@ impl ChainListener {
         let admin_listener: NotificationStream = if client.admin_id() == chain_id {
             Box::pin(stream::pending())
         } else {
-            Box::pin(client.subscribe_to(client.admin_id()).await?)
+            let stream = client.subscribe_to(client.admin_id()).await?;
+            Box::pin(stream.filter(|notification: &Notification| {
```
deuszx, line 171: Does the future need to be pinned?

afck: Yes, the `fuse()` call below seems to require that.

linera-client/src/chain_listener.rs
```diff
@@ -297,4 +277,37 @@ impl ChainListener {
             linera_base::time::timer::sleep(Duration::from_millis(delay_ms)).await;
         }
     }
+
+    async fn maybe_process_inbox<C>(
```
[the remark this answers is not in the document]
afck: Done in https://github.com/linera-io/linera-protocol/pull/3702/commits/bd23adb84b7d9412b785fe1a449bd60d0c1a0e5d.

deuszx (approved):

PR #3705, opened 2025-04-02
Replace `#[should_panic]` with assertions in HTTP example tests

jvff:
## Motivation

The integration tests for the HTTP Requests How-To example (#3509) used the `#[should_panic]` attribute for testing failures. However, as [pointed out during review](https://github.com/linera-io/linera-protocol/pull/3509#discussion_r1986888526), these are quite broad and may lead to mistakes. Being able to assert the exact error leads to better tests and a better example for external developers.

## Proposal

Add `try_query`, `try_graphql_query` and `try_graphql_mutation` helper methods to the `ActiveChain` test interface. Use the new methods to replace usages of `#[should_panic]` with error type assertions.

To reduce boilerplate code, a `WorkerError::expect_execution_error` test helper method was added, that dives into the internal `ExecutionError` stored inside a `WorkerError`.

## Test Plan

CI should catch any regressions caused by this refactor to the tests.

## Release Plan

- These changes follow the usual release cycle, because this only includes small backward-compatible API additions to the public API and SDK.

## Links

### DIFF
--- a/linera-chain/src/lib.rs
+++ b/linera-chain/src/lib.rs
@@ -179,6 +179,7 @@ impl From<ViewError> for ChainError {
 }
 
 #[derive(Copy, Clone, Debug)]
+#[cfg_attr(with_testing, derive(Eq, PartialEq))]
 pub enum ChainExecutionContext {
     Query,
     DescribeApplication,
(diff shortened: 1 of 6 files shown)


linera-chain/src/lib.rs
```diff
@@ -178,7 +178,7 @@ impl From<ViewError> for ChainError {
     }
 }
 
-#[derive(Copy, Clone, Debug)]
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
```
ma2bd, line 181: `#[cfg_attr(with_testing, derive(Eq, PartialEq))]` ?

ma2bd (approved):

Issue #1401, opened 2023-12-18
Cache staged executions of the next block


ma2bd:
Motivation:
* Most validators executing a certified block have already voted on the proposal.
* Similarly, clients call `handle_block_proposal` locally at the beginning of block creation and call `handle_certificate` at the end.
* On top of that, clients might stage the execution of a block even before calling `handle_block_proposal` locally (for a final verification).

Concretely, we could cache the value of `ChainStateView` just before it is rolled back in `handle_block_proposal`.

afck:
I will add a single cache to `WorkerState`, next to the block cache, for now.
(Both caches would work equally well if they were split up into one cache per chain, of course. This would prevent one chain evicting another chain's entries, but it would make it harder to control the _total_ cache size.)

Minor issue: `ValueCache` requires `V: Clone`. `ExecutionStateView` isn't `Clone`, because `ClonableView::clone_unchecked` can fail. This is due to the `ReentrantByteCollectionView`, which tries locking an async `RwLock`.
So I'm going to remove the `V: Clone` requirement from ValueCache, and move it to only the `get` method. (Needed because the actual `LruCache` is inside a mutex.) I'll add a new `take` method that _removes_ the value. In the case of a block proposal, I'll just re-insert it. In the case of a confirmed block, we want to remove it anyway.

PR #3708, opened 2025-04-03
Cache execution states.

afck:
## Motivation

Currently chain workers executes almost every block twice, both in the client's local node and the validators: Once when handling the proposal, and then again when handling the confirmed block.

## Proposal

Cache the `ExecutionStateView` when handling the proposal, and use the cached value for the confirmed block.

This is only done for blocks where no pub-sub channel subscribers are added or removed. (But these will be replaced by event streams soon anyway.)

The second part of https://github.com/linera-io/linera-protocol/issues/1401, caching intermediate states when _staging_ a block, is not part of this PR yet.

## Test Plan

Most existing tests will use this, so CI would fail if it broke anything.

I locally added some logging to verify that a client test (`test_re_propose_locked_block_with_blobs`) has a lot of cache hits. Metrics are already in place for the `ValueCache`, so we should be able to observe the numbers in production, too.

## Release Plan

## Links

- Partly addresses https://github.com/linera-io/linera-protocol/issues/1401.

### DIFF
--- a/linera-core/src/chain_worker/actor.rs
+++ b/linera-core/src/chain_worker/actor.rs
@@ -22,7 +22,8 @@ use linera_chain::{
     ChainStateView,
 };
 use linera_execution::{
-    committee::Epoch, Query, QueryContext, QueryOutcome, ServiceRuntimeEndpoint, ServiceSyncRuntime,
+    committee::Epoch, ExecutionStateView, Query, QueryContext, QueryOutcome,
+    ServiceRuntimeEndpoint, ServiceSyncRuntime,
 };
 use linera_storage::Storage;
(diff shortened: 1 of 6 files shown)


deuszx (approved):
LGTM

linera-core/src/chain_worker/actor.rs
```diff
@@ -181,10 +182,14 @@ where
     ///
     /// If loading the chain state fails the next request will receive the error reported by the
     /// `storage`, and the actor will then try again to load the state.
+    #[allow(clippy::too_many_arguments)]
```
ma2bd, line 185: I believe we've started using `expect` in this case?

PR #3718, opened 2025-04-04
Allow the node service to select arbitrary ports

jvff:
## Motivation

Tests currently rely on a port finder to find an available port for the node service to listen on. However, this prevents the integration tests to run in parallel because the tests race to find available ports.

## Proposal

Change the node service to support using port `0` when binding, allowing the operating system to assign an arbitrary available port. The tests must then read the node service's output until it finds the port number.

## Test Plan

Manually ran the integration tests, and then ran them in parallel.

## Release Plan

- These changes follow the usual release cycle, because this changes the `linera service` CLI command in a non-backwards compatible way. Now not specifying the port forces the node service to listen on an arbitrary port, not the fallback `8080` port.

## Links

### DIFF
--- a/linera-indexer/example/tests/test.rs
+++ b/linera-indexer/example/tests/test.rs
@@ -142,7 +142,7 @@ async fn test_end_to_end_operations_indexer(config: impl LineraNetConfig) {
     .await
     .unwrap()
     .block
-    .unwrap_or_else(|| panic!("no block found"));
+    .expect("no block found");
     let last_hash = last_block.clone().hash;
 
     let indexer_state = request::<State, _>(&req_client, "http://localhost:8081", state::Variables)
--- a/linera-indexer/example/tests/test.rs
+++ b/linera-indexer/example/tests/test.rs
@@ -39,7 +39,7 @@ fn reqwest_client() -> reqwest::Client {
         .unwrap()
 }
 
-async fn run_indexer(path_provider: &PathProvider) -> Child {
+async fn run_indexer(path_provider: &PathProvider, service_port: u16) -> Child {
     let port = 8081;
     let path = resolve_binary("linera-indexer", "linera-indexer-example")
         .await
@@ -48,7 +48,7 @@ async fn run_indexer(path_provider: &PathProvider) -> Child {
     command
         .current_dir(path_provider.path())
         .kill_on_drop(true)
-        .args(["run"]);
+        .args(["run", "--service-port", &service_port.to_string()]);
     let child = command.spawn().unwrap();
     let client = reqwest_client();
     for i in 0..10 {
@@ -73,7 +73,7 @@ fn indexer_running(child: &mut Child) {
     }
 }
 
-async fn transfer(client: &reqwest::Client, from: ChainId, to: Account, amount: &str) {
+async fn transfer(client: &reqwest::Client, port: u16, from: ChainId, to: Account, amount: &str) {
     let variables = transfer::Variables {
         chain_id: from,
         owner: AccountOwner::CHAIN,
@@ -81,7 +81,7 @@ async fn transfer(client: &reqwest::Client, from: ChainId, to: Account, amount: 
         recipient_account: to.owner,
         amount: Amount::from_str(amount).unwrap(),
     };
-    request::<Transfer, _>(client, "http://localhost:8080", variables)
+    request::<Transfer, _>(client, &format!("http://localhost:{port}"), variables)
         .await
         .unwrap();
 }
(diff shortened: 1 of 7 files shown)


linera-service/src/cli_wrappers/wallet.rs
```diff
@@ -450,9 +452,17 @@ impl ClientWrapper {
...
-            .args(["--port".to_string(), port.to_string()])
-            .spawn_into()?;
+        if let Some(requested_port) = port.into() {
+            command.args(["--port".to_string(), requested_port.to_string()]);
+        }
+        let mut child = command.spawn_into()?;
+        let port = Self::parse_node_service_port(
+            child
+                .stdout
+                .take()
```
jvff, line 462: The downside of this is that the node service logs are no longer printed with the test output. Not sure what should be done here. One option is to spawn a task that keeps reading it and printing it out with `eprintln` :thinking:

ma2bd: All machine-readable results must go to stdout.

linera-service/src/node_service.rs
```diff
@@ -828,8 +831,20 @@ where
...
-        let port = self.port.get();
-        let index_handler = axum::routing::get(util::graphiql).post(Self::index_handler);
+        let requested_port = self.port.map(NonZeroU16::get).unwrap_or_default();
+        let listener =
+            tokio::net::TcpListener::bind(SocketAddr::from(([127, 0, 0, 1], requested_port)))
+                .await?;
+        let port = NonZeroU16::try_from(listener.local_addr()?.port())
+            .expect("Sockets should never bind to port zero");
+
+        info!("GraphiQL IDE: http://localhost:{}", port);
```
ma2bd, line 841: `println!`

linera-indexer/example/tests/test.rs
```diff
@@ -133,11 +134,15 @@ async fn test_end_to_end_operations_indexer(config: impl LineraNetConfig) {
...
-        .unwrap_or_else(|| panic!("no block found"));
+    let last_block = request::<Block, _>(
+        &req_client,
+        &format!("http://localhost:{node_service_port}"),
+        variables,
+    )
+    .await
+    .unwrap()
+    .block
+    .unwrap_or_else(|| panic!("no block found"));
```
afck, line 145:
```suggestion
    .expect("no block found");
```

linera-service/src/linera/main.rs
```diff
@@ -1348,7 +1348,10 @@ fn log_file_name_for(command: &ClientCommand) -> Cow<'static, str> {
         ClientCommand::Project { .. } => "project".into(),
         ClientCommand::Watch { .. } => "watch".into(),
         ClientCommand::Storage { .. } => "storage".into(),
-        ClientCommand::Service { port, .. } => format!("service-{port}").into(),
+        ClientCommand::Service {
+            port: Some(port), ..
+        } => format!("service-{port}").into(),
+        ClientCommand::Service { port: None, .. } => "service".into(),
```
afck, line 1354: Is there a danger that multiple services with dynamically assigned ports now use the same log file?

linera-indexer/example/tests/test.rs
christos-h: Doesn't clippy prefer `unwrap_or_else(|| ...)` because it's lazy?

afck: Yes, if the `foo` in `unwrap_or(foo())` does any work. But in this case it's just a `&'static str`.

linera-service/src/linera/main.rs
jvff: Yes, but this only matters when using the `LINERA_LOGS_DIR` environment variable. Long term we should probably update the tests to pass an `--id` to the node services.

linera-service/src/node_service.rs
```diff
@@ -828,8 +831,21 @@ where
...
     #[instrument(name = "node_service", level = "info", skip(self), fields(port = ?self.port))]
     pub async fn run(self) -> Result<(), anyhow::Error> {
-        let port = self.port.get();
-        let index_handler = axum::routing::get(util::graphiql).post(Self::index_handler);
+        let requested_port = self.port.map(NonZeroU16::get).unwrap_or_default();
+        let listener =
+            tokio::net::TcpListener::bind(SocketAddr::from(([127, 0, 0, 1], requested_port)))
+                .await?;
+        let port = NonZeroU16::try_from(listener.local_addr()?.port())
+            .expect("Sockets should never bind to port zero");
```
ma2bd, line 839: very nice

ma2bd (approved):

PR #3722, opened 2025-04-06
Create trait `RunnableWithStore` to simplify the database tool

ma2bd:
## Motivation

Minimize boilerplate + clean up names

follows #3724

## Proposal
 
* (light) refactor how `--config` is handled, notably making it a global option
* create a trait `RunnableWithStore` so that the database command can be written in a single match.
* `linera storage initialize` now takes a genesis config (similar to `linera-server initialize` which could be killed next)
* a few minor code improvements

## Test Plan

CI

### DIFF
--- a/linera-service/src/linera/main.rs
+++ b/linera-service/src/linera/main.rs
@@ -1475,9 +1475,9 @@ impl RunnableWithStore for DatabaseToolJob<'_> {
         S: KeyValueStore + Clone + Send + Sync + 'static,
         S::Error: Send + Sync,
     {
+        let start_time = Instant::now();
         match self.0 {
             DatabaseToolCommand::DeleteAll => {
-                let start_time = Instant::now();
                 S::delete_all(&config).await?;
                 info!(
                     "All namespaces deleted in {} ms",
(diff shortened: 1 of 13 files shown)


linera-service/src/linera/main.rs
```diff
@@ -1448,6 +1459,110 @@ impl ClientOptions {
...
+                    println!("The database does exist");
+                    return Ok(0);
+                } else {
+                    println!("The database does not exist");
+                    return Ok(1);
+                }
+            }
+            DatabaseToolCommand::CheckAbsence => {
+                let start_time = Instant::now();
+                let test = S::exists(&config, &namespace).await?;
```
afck, line 1511:
```suggestion
                let exists = S::exists(&config, &namespace).await?;
```

linera-service/src/linera/main.rs
```diff
@@ -1448,6 +1459,110 @@ impl ClientOptions {
...
+                    "Absence of a namespace checked in {} ms",
+                    start_time.elapsed().as_millis()
+                );
+                if test {
+                    println!("The database does exist");
+                    return Ok(1);
+                } else {
+                    println!("The database does not exist");
+                    return Ok(0);
+                }
```
afck, line 1522: Do we really need both commands, if they are just each other's negation?

linera-service/src/linera/main.rs
```diff
@@ -1448,6 +1459,110 @@ impl ClientOptions {
...
+            }
+            DatabaseToolCommand::CheckAbsence => {
+                let start_time = Instant::now();
+                let test = S::exists(&config, &namespace).await?;
+                info!(
+                    "Absence of a namespace checked in {} ms",
+                    start_time.elapsed().as_millis()
+                );
+                if test {
+                    println!("The database does exist");
```
afck, line 1517: Would it be better to print something more machine-readable, and log this as `info!` instead?

linera-service/src/linera/main.rs
```diff
@@ -1448,6 +1459,110 @@ impl ClientOptions {
...
+        config: S::Config,
+        namespace: String,
+    ) -> Result<Self::Output, anyhow::Error>
+    where
+        S: KeyValueStore + Clone + Send + Sync + 'static,
+        S::Error: Send + Sync,
+    {
+        match self.0 {
+            DatabaseToolCommand::DeleteAll => {
+                let start_time = Instant::now();
```
afck, line 1479: (This could be deduplicated. But it looks like you moved it into the `match` arms on purpose?)

linera-service/src/linera/main.rs
```diff
@@ -1448,6 +1459,110 @@ impl ClientOptions {
...
+                );
+            }
+            DatabaseToolCommand::ListNamespaces => {
+                let start_time = Instant::now();
+                let namespaces = S::list_all(&config).await?;
+                info!(
+                    "Namespaces listed in {} ms",
+                    start_time.elapsed().as_millis()
+                );
+                println!("The list of namespaces is {:?}", namespaces);
```
afck, line 1544: In all three `List*` commands, too: Should the `println!` output be machine-readable?

linera-storage/src/db_storage.rs
```diff
@@ -887,22 +856,63 @@ where
...
+    pub async fn list_chain_ids(
+        config: &Store::Config,
+        namespace: &str,
+    ) -> Result<Vec<ChainId>, ViewError> {
+        let root_keys = Store::list_root_keys(config, namespace).await?;
+        let mut chain_ids = Vec::new();
+        for root_key in root_keys {
+            if root_key.len() == 1 + CHAIN_ID_LENGTH && root_key[0] == INDEX_CHAIN_ID {
+                let root_key_red = &root_key[1..1 + CHAIN_ID_LENGTH];
+                let chain_id = bcs::from_bytes(root_key_red)?;
```
afck, line 911:
```suggestion
                let chain_id = bcs::from_bytes(&root_key[1..])?;
```

linera-storage/src/db_storage.rs
```diff
@@ -887,22 +856,63 @@ where
...
+        config: &Store::Config,
+        namespace: &str,
+    ) -> Result<Vec<BlobId>, ViewError> {
+        let store = Store::maybe_create_and_connect(config, namespace).await?;
+        let prefix = &[INDEX_BLOB_ID];
+        let keys = store.find_keys_by_prefix(prefix).await?;
+        let mut blob_ids = Vec::new();
+        for key in keys.iterator() {
+            let key = key?;
+            let key_red = &key[..BLOB_ID_LENGTH];
```
afck, line 888: (This should probably always be the whole key?)

afck (approved):

MathieuDutSik (approved):
LGTM (after the merge)

PR #3724, opened 2025-04-06
move CLI command and storage back to linera-service

ma2bd:
## Motivation

Minimize the code exposed to `linera-web` thereby facilitating future refactorings.

## Proposal

* Split `ClientOptions`
* Move `Command` and `storage.rs` back to linera-service

## Test Plan

CI


Twey (approved):
I'm not sure about the `storage.rs` move long-term but I'm okay with it for now.

PR #3725, opened 2025-04-07
Remove event subscription functions from contract runtime.

afck:
## Motivation

In https://github.com/linera-io/linera-protocol/pull/3692 I added basic user events, but subscriptions are not implemented yet. However, I optimistically added the contract runtime functions for subscriptions in that PR, too, so they are now user-visible but don't work yet.

## Proposal

Remove them from the contract runtime, so that if we release a new SDK, they don't show up yet.

## Test Plan

CI

## Release Plan

## Links

- Reverts a few lines of https://github.com/linera-io/linera-protocol/pull/3692.

### DIFF
--- a/linera-sdk/src/contract/runtime.rs
+++ b/linera-sdk/src/contract/runtime.rs
@@ -296,26 +296,6 @@ where
         bcs::from_bytes(&event).expect("Failed to deserialize event")
     }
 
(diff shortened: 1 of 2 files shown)


deuszx (approved):
LGTM if CI agrees.

PR #3726, opened 2025-04-07
Refactor the chain listener

afck:
## Motivation

The chain listener is currently one big complex loop. It is about to get more complicated when we add event stream subscriptions.

## Proposal

Split the loop into smaller methods. Move some of the local variables as fields into `ChainListener`. Introduce a `ChainClientListener` that tracks the local variables that are specific to a single chain.

## Test Plan

CI

## Release Plan

## Links

### DIFF
--- a/linera-client/src/chain_listener.rs
+++ b/linera-client/src/chain_listener.rs
@@ -10,7 +10,7 @@ use std::{
 use async_trait::async_trait;
 use futures::{channel::mpsc, lock::Mutex, SinkExt as _, StreamExt};
 use linera_base::{
-    crypto::AccountSecretKey,
+    crypto::{AccountSecretKey, CryptoHash},
     data_types::Timestamp,
     identifiers::{ChainId, Destination},
 };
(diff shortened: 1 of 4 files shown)


linera-client/src/chain_listener.rs
```diff
@@ -85,57 +85,68 @@ pub trait ClientContext: 'static {
...
-        chain_id: ChainId,
-        context: Arc<Mutex<C>>,
-        storage: C::Storage,
-        config: ChainListenerConfig,
-        listening: Arc<Mutex<HashSet<ChainId>>>,
-    ) where
-        C: ClientContext,
-    {
+    fn run_with_chain_id(&self, chain_id: ChainId) {
+        if !self.listening.lock().unwrap().insert(chain_id) {
```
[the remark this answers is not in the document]
afck: (It's only used in this place, and not held across an `await` point.)

linera-client/src/chain_listener.rs
```diff
@@ -85,57 +85,68 @@ pub trait ClientContext: 'static {
...
-    listening: Arc<Mutex<HashSet<ChainId>>>,
+pub struct ChainListener<C: ClientContext> {
+    context: Arc<Mutex<C>>,
+    storage: C::Storage,
+    config: Arc<ChainListenerConfig>,
+    listening: Arc<std::sync::Mutex<HashSet<ChainId>>>,
 }
 
-impl ChainListener {
+impl<C: ClientContext> Clone for ChainListener<C> {
```
Twey, line 95: Isn't this `#[derive(Clone)]`?

linera-client/src/chain_listener.rs
```diff
@@ -85,57 +85,68 @@ pub trait ClientContext: 'static {
...
+    fn run_with_chain_id(&self, chain_id: ChainId) {
+        if !self.listening.lock().unwrap().insert(chain_id) {
+            // If we are already listening to notifications, there's nothing to do.
+            // This can happen if we download a child before the parent
+            // chain, and then process the OpenChain message in the parent.
+            return;
+        }
+        let listener = self.clone();
         let _handle = linera_base::task::spawn(
             async move {
```
Twey, line 148: Maybe this should be `in_current_span`?

Twey (approved):

linera-client/src/chain_listener.rs
afck: Only if `C: Clone`, which we don't actually need.

linera-client/src/chain_listener.rs
afck: Ah, good catch! :+1:

afck: Done in https://github.com/linera-io/linera-protocol/pull/3726/commits/b62d410eed39377d1847d40bf21c1b4a88d1747d.

PR #3731, opened 2025-04-08
display BlobTypes without quotes

ma2bd:
## Motivation

Nicer output for `linera storage list_blob_ids`

## Proposal

* Use the derived `Debug` to display `BlobType`s without quotes
* Keep using `serde_json` for parsing (as opposed to importing `clap` or other crates)

## Test Plan

CI

### DIFF
--- a/linera-base/src/identifiers.rs
+++ b/linera-base/src/identifiers.rs
@@ -183,10 +183,7 @@ pub enum BlobType {
 
 impl Display for BlobType {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        match serde_json::to_string(self) {
-            Ok(s) => write!(f, "{}", s),
-            Err(_) => Err(fmt::Error),
-        }
+        write!(f, "{:?}", self)
     }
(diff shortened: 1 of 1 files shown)


linera-base/src/identifiers.rs
```diff
@@ -183,18 +183,16 @@ pub enum BlobType {
...
     }
 }
 
 impl FromStr for BlobType {
     type Err = anyhow::Error;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        serde_json::from_str(s).with_context(|| format!("Invalid BlobType: {}", s))
+        serde_json::from_str(&format!("\"{s}\""))
+            .with_context(|| format!("Invalid BlobType: {}", s))
```
afck, line 195: Doing this manually with `match` would avoid the string allocation.

afck (approved):

linera-base/src/identifiers.rs
ma2bd: Well I didn't want to write all the case

PR #3745, opened 2025-04-10
Rewrite the chain listener as a single loop.

afck:
## Motivation

For https://github.com/linera-io/linera-protocol/issues/365 the chain listener will need to react to new block notifications on a publisher chain by processing the inboxes of all subscriber chains. That is difficult to implement with the current approach: we run a notification processing loop in a separate task for each chain.

## Proposal

Rewrite the chain listener so that there is a single top-level loop processing all events on all chains. That will make it easier to update chains based on events from _other_ chains.

`process_new_events` already does this, to update all non-admin chains for new epochs.

## Test Plan

CI

## Release Plan

## Links

- Preparation for #365.

### DIFF
--- a/linera-base/src/task.rs
+++ b/linera-base/src/task.rs
@@ -138,10 +138,10 @@ mod implementation {
     }
 
     /// The type of a future awaiting another task.
-    pub type NonblockingFuture<R> = oneshot::Receiver<R>;
+    pub type NonBlockingFuture<R> = oneshot::Receiver<R>;
 
     /// Spawns a new task on the current thread.
-    pub fn spawn<F: Future + 'static>(future: F) -> NonblockingFuture<F::Output> {
+    pub fn spawn<F: Future + 'static>(future: F) -> NonBlockingFuture<F::Output> {
         let (send, recv) = oneshot::channel();
         wasm_bindgen_futures::spawn_local(async {
             let _ = send.send(future.await);
--- a/linera-client/src/chain_listener.rs
+++ b/linera-client/src/chain_listener.rs
@@ -9,11 +9,16 @@ use std::{
 };
 
 use async_trait::async_trait;
-use futures::{future::select_all, lock::Mutex, FutureExt as _, StreamExt};
+use futures::{
+    future::{join_all, select_all},
+    lock::Mutex,
+    FutureExt as _, StreamExt,
+};
 use linera_base::{
     crypto::{AccountSecretKey, CryptoHash},
     data_types::Timestamp,
     identifiers::{ChainId, Destination},
+    task::NonBlockingFuture,
 };
 use linera_core::{
     client::{AbortOnDrop, ChainClient, ChainClientError},
@@ -22,6 +27,7 @@ use linera_core::{
 };
 use linera_execution::{Message, OutgoingMessage, SystemMessage};
 use linera_storage::{Clock as _, Storage};
+use tokio_util::sync::CancellationToken;
 use tracing::{debug, info, instrument, warn, Instrument as _};
 
 use crate::{wallet::Wallet, Error};
@@ -93,7 +99,9 @@ struct ListeningClient<C: ClientContext> {
     /// The chain client.
     client: ContextChainClient<C>,
     /// The abort handle for the task that listens to the validators.
-    _abort_handle: AbortOnDrop,
+    abort_handle: AbortOnDrop,
+    /// The listening task's join handle.
+    join_handle: NonBlockingFuture<()>,
     /// The stream of notifications from the local node.
     notification_stream: Arc<Mutex<NotificationStream>>,
     /// This is only `< u64::MAX` when the client is waiting for a timeout to process the inbox.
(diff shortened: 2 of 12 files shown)


linera-client/src/chain_listener.rs
```diff
@@ -122,141 +137,58 @@ impl<C: ClientContext> ChainListener<C> {
...
-impl<C: ClientContext> ChainClientListener<C> {
-    /// Creates a new chain client listener.
-    fn new(client: ContextChainClient<C>, listener: ChainListener<C>) -> Self {
-        Self {
-            client,
-            listener,
-            timeout: Timestamp::from(u64::MAX),
+            self.listen(chain_id).await?;
         }
+        let _handler = linera_base::task::spawn(self.run_notification_loop().in_current_span());
```
Twey, line 150: I'm not totally clear what this spawn is for.  Can we not let the caller decide how to poll this chain listener?

linera-client/src/error.rs
```diff
@@ -39,6 +39,8 @@ pub(crate) enum Inner {
     #[cfg(feature = "benchmark")]
     #[error("Benchmark error: {0}")]
     Benchmark(#[from] BenchmarkError),
+    #[error("internal error; this is a bug: {0}")]
+    Internal(String),
```
Twey, line 43: Could we just panic?

linera-client/src/chain_listener.rs
afck: Done in https://github.com/linera-io/linera-protocol/pull/3745/commits/ba3a9fad87b93ac1c44adf5c95c182ec186b0ef7.

linera-client/src/unit_tests/chain_listener.rs
```diff
@@ -142,8 +142,12 @@ async fn test_chain_listener() -> anyhow::Result<()> {
         .update_wallet_for_new_chain(chain_id0, Some(key_pair), clock.current_time())
         .await?;
     let context = Arc::new(Mutex::new(context));
-    let listener = ChainListener::new(config, context, storage);
-    listener.run().await;
+    let _handle = linera_base::task::spawn(async move {
```
ma2bd, line 145: `let _ = ...` is generally a bad idea as it disables important clippy warnings.

linera-service/src/node_service.rs
```diff
@@ -843,14 +843,15 @@ where
...
-        serve_fut.await?;
+
+        let chain_listener = ChainListener::new(self.config, self.context, self.storage).run();
+        let tcp_listener =
+            tokio::net::TcpListener::bind(SocketAddr::from(([0, 0, 0, 0], port))).await?;
+        let server = axum::serve(tcp_listener, app).into_future();
+        futures::select! {
+            result = Box::pin(chain_listener).fuse() => result?,
+            result = Box::pin(server).fuse() => result?,
+        };
```
ma2bd, line 854: cool

linera-client/src/chain_listener.rs
```diff
@@ -274,37 +199,143 @@ impl<C: ClientContext> ChainClientListener<C> {
...
             debug!("Not processing inbox due to listener configuration");
             return Ok(());
         }
+        let listening_client = self.listening.get_mut(&chain_id).expect("missing client");
         debug!("Processing inbox");
-        match self.client.process_inbox_without_prepare().await {
+        listening_client.timeout = Timestamp::from(u64::MAX);
+        match listening_client
+            .client
+            .process_inbox_without_prepare()
```
ma2bd, line 336: Does it mean that we can finally make client functions like `process_inbox_without_prepare ` take `&mut self` again?

ma2bd (commented):
looks nice

linera-client/src/chain_listener.rs
afck: I'm not sure. Was the chain listener the only reason we changed that, @Twey?

linera-client/src/unit_tests/chain_listener.rs
afck: But that's the point here: `spawn` starts running the task immediately in the background, but it also returns a future that can be `await`ed to join the task. We don't want to join in this case, so we don't use that future, which is what Clippy would complain about.

linera-client/src/chain_listener.rs
```diff
@@ -274,37 +199,143 @@ impl<C: ClientContext> ChainClientListener<C> {
...
+                if chain_id != admin_id {
+                    self.maybe_process_inbox(chain_id).await?;
+                }
             }
         }
         Ok(())
     }
 
+    /// Start listening for notifications about the given chain.
+    async fn listen(&mut self, chain_id: ChainId) -> Result<(), Error> {
```
deuszx, line 238: Can we dynamically add new chains to listen to? i.e. can we call `listen` from the outisde and this process will add new subscription streams after the `self.run` has been called?

afck: To do that, we'll have to add a third arm to the `select!` statement that notifies about a new chain ID to listen to. Should be doable if we need it.

linera-client/src/unit_tests/chain_listener.rs
ma2bd: I see. We could also signal the listener to stop at the end of the test and wait for it cleanly.

afck: Yes, maybe I'll add the join handle to `ListeningClient`. Unfortunately we can't just join on `Drop`, because it's `async`. :thinking:

afck: Done in https://github.com/linera-io/linera-protocol/pull/3745/commits/40831db1c676f3528dc410c0bd47f250623306b6.

linera-client/src/chain_listener.rs
Twey: Pedantically we can, but it wouldn't change much semantically since chain clients can be taken from the `Client` with a mere `&` reference.  This is probably still necessary because `ChainClient::listen` still shares the client.

linera-base/src/task.rs
```diff
@@ -138,10 +138,10 @@ mod implementation {
     }
 
     /// The type of a future awaiting another task.
-    pub type NonblockingFuture<R> = oneshot::Receiver<R>;
+    pub type NonBlockingFuture<R> = oneshot::Receiver<R>;
```
Twey, line 141: https://en.wiktionary.org/wiki/nonblocking https://books.google.com/ngrams/graph?content=nonblocking%2Cnon-blocking&year_start=1800&year_end=2022&corpus=en&smoothing=3 :) (I don't really mind though, if we want to make ‘non-blocking’ our standard spelling.)

afck: I had to change it because the _other_ (non-`web`) conditionally compiled type was `NonBlockingFuture`. Happy to change the other one instead, if you prefer.

PR #3748, opened 2025-04-10
make the block exporter context visible in the Storage trait

ma2bd:
## Motivation

#3729 was not enough to re-use `run_with_storage` in the block exporter

## Proposal

* Move the new definition to the trait.
* Add the necessary associated type.

This is a little ugly but the alternative is to have different variants of the `Storage` trait depending on the app using it. This could be done later.

## Test Plan

CI


Twey (approved):
We want to refactor this anyway — accept for now

PR #3755, opened 2025-04-11
Store network information in DB

ma2bd:
## Motivation

We currently pass around a genesis configuration when deploying a validator but this is not really needed.
This is the first step to simplify things.

## Proposal

* Create a notion of network description to be persisted in DB
* Use it in the proxy

Incidentally, we won't compute the genesis config hash over and over again.

## Test Plan

CI

## Release Plan

### DIFF
--- a/linera-storage/src/db_storage.rs
+++ b/linera-storage/src/db_storage.rs
@@ -43,7 +43,7 @@ use {
     prometheus::{HistogramVec, IntCounterVec},
 };
 
-use crate::{ChainRuntimeContext, Clock, Storage};
+use crate::{ChainRuntimeContext, Clock, NetworkDescription, Storage};
 
 /// The metric counting how often a blob is tested for existence from storage
 #[cfg(with_metrics)]
@@ -207,6 +207,28 @@ pub static WRITE_EVENT_COUNTER: LazyLock<IntCounterVec> = LazyLock::new(|| {
     )
 });
 
+/// The metric counting how often the network description is read from storage.
+#[cfg(with_metrics)]
+#[doc(hidden)]
+pub static READ_NETWORK_DESCRIPTION: LazyLock<IntCounterVec> = LazyLock::new(|| {
+    register_int_counter_vec(
+        "network_description",
+        "The metric counting how often the network description is read from storage",
+        &[],
+    )
+});
+
+/// The metric counting how often the network description is written to storage.
+#[cfg(with_metrics)]
+#[doc(hidden)]
+pub static WRITE_NETWORK_DESCRIPTION: LazyLock<IntCounterVec> = LazyLock::new(|| {
+    register_int_counter_vec(
+        "write_event",
+        "The metric counting how often the network description is written to storage",
+        &[],
+    )
+});
+
 trait BatchExt {
     fn add_blob(&mut self, blob: &Blob) -> Result<(), ViewError>;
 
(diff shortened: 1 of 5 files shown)


linera-rpc/proto/rpc.proto
```diff
@@ -120,6 +120,12 @@ message VersionInfo {
     string wit_hash = 6;
 }
 
+message NetworkDescription {
+    string name = 1;
+    CryptoHash genesis_config_hash = 2;
+    uint64 timestamp = 3;
+}
```
deuszx, line 127: `name` and `genesis_config_hash` are self-explanatory but `timestamp` is not IMHO. Is it a timestamp at the time of generating the instance of `NetworkDescription`? Is it a timestamp of creating the message? Is it a timestamp of generating the genesis config? 

Similarly for the `NetworkDescription` struct.

linera-storage/src/lib.rs
```diff
@@ -339,6 +349,16 @@ pub trait Storage: Sized {
     ) -> Result<Self::BlockExporterContext, ViewError>;
 }
 
+/// A description of the current Linera network to be stored in every node's database.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[cfg_attr(with_testing, derive(Eq, PartialEq))]
+pub struct NetworkDescription {
+    pub name: String,
+    pub genesis_config_hash: CryptoHash,
+    pub timestamp: Timestamp,
```
deuszx, line 358:
```suggestion
    pub genesis_timestamp: Timestamp,
```
Then it becomes self-explanatory like the rest of the fields.

deuszx (approved):
Apart from the comments/rename - LGTM.

linera-service/src/linera/main.rs
```diff
@@ -380,12 +380,15 @@ impl Runnable for Job {
...
-                    ),
+                match node.get_network_description().await {
+                    Ok(description) => {
+                        if description.genesis_config_hash != genesis_config_hash {
+                            error!(
+                                "Validator's genesis config hash {} does not match our own: {}.",
+                                description.genesis_config_hash, genesis_config_hash
+                            );
+                        }
+                    }
```
afck, line 391: Here it might be useful to also print both network names. (Also below.)

linera-storage/src/db_storage.rs
```diff
@@ -207,6 +207,28 @@ pub static WRITE_EVENT_COUNTER: LazyLock<IntCounterVec> = LazyLock::new(|| {
...
+#[doc(hidden)]
+pub static READ_NETWORK_DESCRIPTION: LazyLock<IntCounterVec> = LazyLock::new(|| {
+    register_int_counter_vec(
+        "network_description",
+        "The metric counting how often the network description is read from storage",
+        &[],
+    )
+});
+
+/// The metric counting how often the network description is written to storage.
```
afck, line 221: (This one should really be always 1, should it? :sweat_smile:)

afck (approved):

linera-storage/src/lib.rs
ma2bd: yes good idea

linera-storage/src/db_storage.rs
ma2bd: That's right. We could also refuse to overwrite the network description btw

linera-service/src/linera/main.rs
ma2bd: Mmm I'm going to send another PR for this.

linera-storage/src/db_storage.rs
ma2bd: This may change in the future when initialization becomes idempotent after @bart-linera's work.

ma2bd: Turns out the dual store does call initialize several times

PR #3764, opened 2025-04-14
Bump the version of alloy in order to address the `getrandom` problem.

MathieuDutSik:
## Motivation

It has emerged that the `alloy` depends on `getrandom` that causes some compilation problems under Wasm.

## Proposal

The following was done:
* Switch the version of `alloy-sol-types` to 1.0.0 which address the problem.
* Switch from `alloy` to `alloy-primitives` at version 1.0.0 so as to reduce the dependencies.

## Test Plan

The CI. 

## Release Plan

Normal release process.

## Links

None.


deuszx (commented):
Why not switch it for the root (`Cargo.toml`) as well? RN it's bit weird that we have pre-1.0.0 for all linera crates but in examples we have 1.0.0

deuszx (approved):

PR #3765, opened 2025-04-14
Remove the Localstack word from the source code

### LINKED ISSUE #932: Eliminate `localstack` from the code.
We used to use localstack for testing the `DynamoDb` functionality and so the code is littered with its mention.
However, we are no longer using it so this should be cleaned up.

MathieuDutSik:
## Motivation

We have not been using Localstack for several years and instead we use DynamoDB local.

Fixes #932 

## Proposal

The renaming is done in a straightforward way.

## Test Plan

The CI

## Release Plan

Normal release schedule.

## Links

None.

### DIFF
--- a/.github/workflows/dynamodb.yml
+++ b/.github/workflows/dynamodb.yml
@@ -58,6 +58,6 @@ jobs:
         AWS_REGION: us-east-1
         AWS_ACCESS_KEY_ID: test
         AWS_SECRET_ACCESS_KEY: test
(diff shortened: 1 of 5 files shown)


ma2bd (approved):
Thanks for the cleanup. I wonder if we could just make the ENDPOINT part of the config as a replacement for the boolean value.

MathieuDutSik:
> Thanks for the cleanup. I wonder if we could just make the ENDPOINT part of the config as a replacement for the boolean value.

Ok, will see next.

PR #3767, opened 2025-04-14
Increase the durations in `test_service_as_oracle_timeout_early_stop`.

### LINKED ISSUE #3766: `test_service_as_oracle_timeout_early_stop` fails locally.
Maybe the expected execution time and delays just need to be increased?

afck:
## Motivation

`cargo test -p linera-chain` fails for me locally most of the time.

(Note that it only fails if I run _all_ the `linera-chain` tests. If it's _only_ `test_service_as_oracle_timeout_early_stop`, it passes.)

## Proposal

Multiply the durations in the test by 10, to make them less flaky.

## Test Plan

With this change, `cargo test -p linera-chain` works for me, even in debug mode.

## Release Plan

## Links

- Closes #3766.

### DIFF
--- a/linera-chain/src/unit_tests/chain_tests.rs
+++ b/linera-chain/src/unit_tests/chain_tests.rs
@@ -400,17 +400,17 @@ async fn test_service_as_oracle_exceeding_time_limit(
 }
 
 /// Tests if execution fails early if services call `check_execution_time`.
-#[test_case(&[120]; "single service as oracle call")]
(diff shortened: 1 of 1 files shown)


jvff (approved):
LGTM! We can also reduce this later if the tests take too long to run.

afck:
(I think it's still negligible compared to the rest of all our tests, so I went with a factor 10, to make sure it's not flaky anymore.)

Issue #3782, opened 2025-04-16
Social example: All subscribers should see likes and comments.


afck:
Currently in the social example, if A likes B's post, C will only see that like if it is subscribed not only to B, but also to A.

It would be better if the like were sent as a direct message from A to B, and then B would publish it via the event stream (formerly channel).

zibosvet:
I agree with that

PR #3784, opened 2025-04-16
Implement event stream subscriptions; port social example.

afck:
## Motivation

We want to replace pub-sub channels with event streams for better scalability (https://github.com/linera-io/linera-protocol/issues/365).

Events can already be emitted and read; this PR adds subscriptions.

## Proposal

Handle the `UpdateStreams` system operation. Chain owners can inform applications on a chain about new events in streams published by other chains this way. Also add back `subscribe_to_events` and `unsubscribe_from_events`.

Keep a note in the `TransactionTracker` about any changes to event subscriptions or new events in relevant streams. As part of the same transaction, call `process_streams` in all subscriber contracts.

When synchronizing a chain, and in the chain listener, detect new events and commit `UpdateStreams` as appropriate. (This is part of `process_inbox` now.)

Port the `social` example to use event streams instead of pub-sub channels, without changing any of the business logic (which should be fixed, too, but that is a separate issue: https://github.com/linera-io/linera-protocol/issues/3782).

## Test Plan

CI; the two pub-sub channel tests have been adapted.

## Release Plan

## Links

- Part of https://github.com/linera-io/linera-protocol/issues/365.

### DIFF
--- a/examples/social/src/contract.rs
+++ b/examples/social/src/contract.rs
@@ -6,7 +6,7 @@
 mod state;
 
 use linera_sdk::{
-    linera_base_types::{ChainId, StreamId, WithContractAbi},
+    linera_base_types::{ChainId, StreamUpdate, WithContractAbi},
     views::{RootView, View},
     Contract, ContractRuntime,
 };
@@ -72,30 +72,25 @@ impl Contract for SocialContract {
         panic!("Unexpected message");
     }
 
-    async fn process_streams(&mut self, streams: Vec<(ChainId, StreamId, u32)>) {
-        for (chain_id, stream_id, next_index) in streams {
-            assert_eq!(stream_id.stream_name, STREAM_NAME.into());
+    async fn process_streams(&mut self, updates: Vec<StreamUpdate>) {
+        for update in updates {
+            assert_eq!(update.stream_id.stream_name, STREAM_NAME.into());
             assert_eq!(
-                stream_id.application_id,
+                update.stream_id.application_id,
                 self.runtime.application_id().forget_abi().into()
             );
-            let stored_count = self
-                .state
-                .processed_events
-                .get_mut_or_default(&chain_id)
-                .await
-                .expect("Failed to read from state");
-            let count = *stored_count;
-            *stored_count = next_index;
-            for index in count..next_index {
-                let event = self.runtime.read_event(chain_id, STREAM_NAME.into(), index);
+            for index in update.new_indices() {
+                let event = self
+                    .runtime
+                    .read_event(update.chain_id, STREAM_NAME.into(), index);
                 match event {
                     Event::Post { post, index } => {
-                        self.execute_post_event(chain_id, index, post);
+                        self.execute_post_event(update.chain_id, index, post);
                     }
                     Event::Like { key } => self.execute_like_event(key).await,
                     Event::Comment { key, comment } => {
-                        self.execute_comment_event(key, chain_id, comment).await;
+                        self.execute_comment_event(key, update.chain_id, comment)
+                            .await;
                     }
                 }
             }
--- a/examples/social/src/contract.rs
+++ b/examples/social/src/contract.rs
@@ -14,8 +14,8 @@ use serde::{Deserialize, Serialize};
 use social::{Comment, Key, Operation, OwnPost, Post, SocialAbi};
 use state::SocialState;
 
-/// The channel name the application uses for cross-chain messages about new posts.
-const POSTS_CHANNEL_NAME: &[u8] = b"posts";
+/// The stream name the application uses for events about posts, likes and comments.
+const STREAM_NAME: &[u8] = b"posts";
 
 pub struct SocialContract {
     state: SocialState,
@@ -51,12 +51,12 @@ impl Contract for SocialContract {
             Operation::Subscribe { chain_id } => {
                 let app_id = self.runtime.application_id().forget_abi();
                 self.runtime
-                    .subscribe_to_events(chain_id, app_id, POSTS_CHANNEL_NAME.into());
+                    .subscribe_to_events(chain_id, app_id, STREAM_NAME.into());
             }
             Operation::Unsubscribe { chain_id } => {
                 let app_id = self.runtime.application_id().forget_abi();
                 self.runtime
-                    .unsubscribe_from_events(chain_id, app_id, POSTS_CHANNEL_NAME.into());
+                    .unsubscribe_from_events(chain_id, app_id, STREAM_NAME.into());
             }
             Operation::Post { text, image_url } => {
                 self.execute_post_operation(text, image_url).await
@@ -74,7 +74,7 @@ impl Contract for SocialContract {
 
     async fn process_streams(&mut self, streams: Vec<(ChainId, StreamId, u32)>) {
         for (chain_id, stream_id, next_index) in streams {
-            assert_eq!(stream_id.stream_name, POSTS_CHANNEL_NAME.into());
+            assert_eq!(stream_id.stream_name, STREAM_NAME.into());
             assert_eq!(
                 stream_id.application_id,
                 self.runtime.application_id().forget_abi().into()
@@ -88,12 +88,10 @@ impl Contract for SocialContract {
             let count = *stored_count;
             *stored_count = next_index;
             for index in count..next_index {
-                let event = self
-                    .runtime
-                    .read_event(chain_id, POSTS_CHANNEL_NAME.into(), index);
+                let event = self.runtime.read_event(chain_id, STREAM_NAME.into(), index);
                 match event {
-                    Event::Post { post } => {
-                        self.execute_post_event(chain_id, next_index, post);
+                    Event::Post { post, index } => {
+                        self.execute_post_event(chain_id, index, post);
                     }
                     Event::Like { key } => self.execute_like_event(key).await,
                     Event::Comment { key, comment } => {
@@ -117,19 +115,19 @@ impl SocialContract {
             text,
             image_url,
         };
+        let index = self.state.own_posts.count().try_into().unwrap();
         self.state.own_posts.push(post.clone());
         self.runtime
-            .emit(POSTS_CHANNEL_NAME.into(), &Event::Post { post });
+            .emit(STREAM_NAME.into(), &Event::Post { post, index });
     }
 
     async fn execute_like_operation(&mut self, key: Key) {
-        self.runtime
-            .emit(POSTS_CHANNEL_NAME.into(), &Event::Like { key });
+        self.runtime.emit(STREAM_NAME.into(), &Event::Like { key });
     }
 
     async fn execute_comment_operation(&mut self, key: Key, comment: String) {
         self.runtime
-            .emit(POSTS_CHANNEL_NAME.into(), &Event::Comment { key, comment });
+            .emit(STREAM_NAME.into(), &Event::Comment { key, comment });
     }
 
     fn execute_post_event(&mut self, chain_id: ChainId, index: u32, post: OwnPost) {
@@ -195,7 +193,7 @@ impl SocialContract {
 #[derive(Debug, PartialEq, Serialize, Deserialize)]
 pub enum Event {
     /// A new post was created
-    Post { post: OwnPost },
+    Post { post: OwnPost, index: u32 },
     /// A user liked a post
     Like { key: Key },
     /// A user commented on a post
--- a/examples/social/src/contract.rs
+++ b/examples/social/src/contract.rs
@@ -6,11 +6,12 @@
 mod state;
 
 use linera_sdk::{
-    linera_base_types::{ChainId, ChannelName, Destination, MessageId, WithContractAbi},
+    linera_base_types::{ChainId, StreamId, WithContractAbi},
     views::{RootView, View},
     Contract, ContractRuntime,
 };
-use social::{Comment, Key, Message, Operation, OwnPost, Post, SocialAbi};
+use serde::{Deserialize, Serialize};
+use social::{Comment, Key, Operation, OwnPost, Post, SocialAbi};
 use state::SocialState;
 
 /// The channel name the application uses for cross-chain messages about new posts.
@@ -28,10 +29,10 @@ impl WithContractAbi for SocialContract {
 }
 
 impl Contract for SocialContract {
-    type Message = Message;
+    type Message = ();
     type InstantiationArgument = ();
     type Parameters = ();
-    type EventValue = OwnPost;
+    type EventValue = Event;
 
     async fn load(runtime: ContractRuntime<Self>) -> Self {
         let state = SocialState::load(runtime.root_view_storage_context())
@@ -46,9 +47,17 @@ impl Contract for SocialContract {
     }
 
     async fn execute_operation(&mut self, operation: Operation) -> Self::Response {
-        let (destination, message) = match operation {
-            Operation::Subscribe { chain_id } => (chain_id.into(), Message::Subscribe),
-            Operation::Unsubscribe { chain_id } => (chain_id.into(), Message::Unsubscribe),
+        match operation {
+            Operation::Subscribe { chain_id } => {
+                let app_id = self.runtime.application_id().forget_abi();
+                self.runtime
+                    .subscribe_to_events(chain_id, app_id, POSTS_CHANNEL_NAME.into());
+            }
+            Operation::Unsubscribe { chain_id } => {
+                let app_id = self.runtime.application_id().forget_abi();
+                self.runtime
+                    .unsubscribe_from_events(chain_id, app_id, POSTS_CHANNEL_NAME.into());
+            }
             Operation::Post { text, image_url } => {
                 self.execute_post_operation(text, image_url).await
             }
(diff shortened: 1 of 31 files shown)


examples/social/src/contract.rs
```diff
@@ -91,56 +108,32 @@ impl Contract for SocialContract {
...
         };
-        let index = self.runtime.emit(POSTS_CHANNEL_NAME.into(), &post);
+        let index = self.state.own_posts.count().try_into().unwrap();
         self.state.own_posts.push(post.clone());
-        (
-            ChannelName::from(POSTS_CHANNEL_NAME.to_vec()).into(),
-            Message::Post { index, post },
-        )
+        self.runtime
+            .emit(STREAM_NAME.into(), &Event::Post { post, index });
```
deuszx, line 121: I would say that `Post` shouldn't need an index and the events should be indexed internally - but that's orthogonal.

examples/social/src/state.rs
```diff
@@ -12,4 +15,6 @@ pub struct SocialState {
     pub own_posts: LogView<OwnPost>,
     /// Posts we received from authors we subscribed to.
     pub received_posts: CustomMapView<Key, Post>,
+    /// Other chains events we have already processed.
+    pub processed_events: MapView<ChainId, u32>,
```
deuszx, line 19: Normally this kind of tracking would be done in service, not in the contract - right? Otherwise the indexing over all emitted events in `process_stream` is unnecessarily expensive.

afck: It has to be the contract that keeps this map: When `process_streams` gets called, it needs to know which new events to read.

examples/social/src/contract.rs
afck: The other chains need a way to refer to this specific post. I originally had used the event index here, but that's a bit wasteful: Some events aren't new posts.
(But the whole contract will need significant changes for https://github.com/linera-io/linera-protocol/issues/3782 anyway.)

examples/social/src/state.rs
deuszx: I thought this decision - which events to read - will be done externally as well. Otherwise contract execution pays for detecting what to ignore and ignoring that.

linera-client/src/chain_listener.rs
```diff
@@ -140,6 +139,8 @@ pub struct ChainListener<C: ClientContext> {
     storage: C::Storage,
     config: Arc<ChainListenerConfig>,
     listening: BTreeMap<ChainId, ListeningClient<C>>,
+    /// The set of chains that subscribe to at least one event stream on the given chain.
+    event_subscribers: BTreeMap<ChainId, BTreeSet<ChainId>>,
```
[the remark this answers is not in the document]
afck: We don't need to persist it because we recreate it on client startup from the information in the chain's execution state.

linera-service/tests/linera_net_tests.rs
```diff
@@ -913,19 +913,9 @@ async fn test_wasm_end_to_end_social_user_pub_sub(config: impl LineraNetConfig)
...
         .await?;
 
-    node_service1.process_inbox(&chain1).await?;
-
-    // The returned hash should now be the latest one.
-    let query = format!("query {{ chain(chainId: \"{chain2}\") {{ tipState {{ blockHash }} }} }}");
-    for node_service in [&node_service2, &node_service1] {
-        let response = node_service.query_node(&query).await?;
-        assert_eq!(hash, response["chain"]["tipState"]["blockHash"]);
-    }
```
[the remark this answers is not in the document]
afck: Good point; I guess even though it's not a message anymore, the other node service still has to sync the chain to get the event. I can add it back.

afck:
> I thought this decision - which events to read - will be done externally as well.

No, the current design (see LIP-20) just updates the contracts about the highest index; the rest is up to the contract.

examples/social/src/state.rs
ma2bd: @afck The system has to maintain a similar map for deduplication purposes, right? So what we could do is pass the previous stream index:
```
async fn process_streams(&mut self, streams: Vec<(ChainId, StreamId, /*previous*/ Option<u32>, /*new*/ u32)>
```

ma2bd: Btw. we should probably define something like `struct StreamUpdate { chain_id, stream_id, previous_index, new_index }`.

linera-execution/src/lib.rs
```diff
@@ -466,6 +475,41 @@ pub struct MessageContext {
     pub message_id: MessageId,
 }
 
+#[derive(Clone, Copy, Debug)]
+pub struct ProcessStreamsContext {
+    /// The current chain ID.
+    pub chain_id: ChainId,
+    /// The authenticated signer of the operation that created the message, if any.
+    #[debug(skip_if = Option::is_none)]
+    pub authenticated_signer: Option<AccountOwner>,
```
ma2bd, line 484: I believe this should always be `None` -- since stream updates are automatically processed (like incoming messages) by standard clients.

examples/social/src/state.rs
ma2bd: (and ofc we can also use `previous_index_plus_one` aka `previous_next_index` to avoid the `Option`)

ma2bd: Based on the `UpdateStreams` handling in `system.rs`, it looks like this should be easy. I'm excited :)

ma2bd: This way, contracts have the freedom to handle every element of a stream exactly once -- or just the last one whenever a notification arrives. In previous designs, it was hard to support both modes efficiently.

afck: > The system has to maintain a similar map for deduplication purposes, right?

Not really, no. But I can add that. I agree that would be more user-friendly.

ma2bd: oh I see. `EventSubscriptions::next_index` is a bit different

linera-execution/src/system.rs
```diff
@@ -517,9 +517,31 @@ where
                 for (chain_id, stream_id, next_index) in streams {
                     let subscriptions = self
                         .event_subscriptions
-                        .get_mut_or_default(&(chain_id, stream_id))
+                        .get_mut_or_default(&(chain_id, stream_id.clone()))
+                        .await?;
+                    if subscriptions.next_index >= next_index {
+                        continue;
+                    }
+                    subscriptions.next_index = next_index;
```
deuszx, line 525: Maybe we should update the index after we act on it.

linera-execution/src/execution.rs
```diff
@@ -327,6 +336,8 @@ where
                 .await?;
             }
         }
+        self.process_subscriptions(txn_tracker, resource_controller, context.into())
```
deuszx, line 339: Why do we process subscriptions after both execution of messages and operations? My understanding was that messages can be only used to send information between two instances of the same app on different chains. I don't see how events (emittion, subscription) could be manipulated here.

linera-execution/src/execution.rs
```diff
@@ -495,4 +506,50 @@ where
         }
         Ok(applications)
     }
+
+    async fn process_subscriptions(
```
deuszx, line 510: It'd be good to have a comment on this method describing what it does - `process_subscriptions` doesn't say anything really.

linera-execution/src/system.rs
```diff
@@ -517,9 +517,31 @@ where
                 for (chain_id, stream_id, next_index) in streams {
                     let subscriptions = self
                         .event_subscriptions
-                        .get_mut_or_default(&(chain_id, stream_id))
+                        .get_mut_or_default(&(chain_id, stream_id.clone()))
+                        .await?;
+                    if subscriptions.next_index >= next_index {
```
deuszx, line 522: How could it happen that we have an `UpdateStream` call that tries to consume an event with index lower than what it is currently? It'd mean that that this event stream has already been updated by some other block proposer on this chain?

linera-execution/src/system.rs
```diff
@@ -517,9 +517,31 @@ where
...
+                    let index = next_index
+                        .checked_sub(1)
+                        .ok_or(ArithmeticError::Underflow)?;
+                    self.context()
+                        .extra()
+                        .get_event(EventId {
+                            chain_id,
+                            stream_id: stream_id.clone(),
+                            index,
+                        })
```
deuszx, line 543: This is just to make sure that we have that event locally?

linera-core/src/client/mod.rs
```diff
@@ -1526,6 +1588,29 @@ where
...
+            .chain_state_view()
+            .await?
+            .execution_state
+            .system
+            .event_subscriptions
+            .indices()
+            .await?
+            .iter()
+            .map(|(chain_id, _)| *chain_id)
+            .filter(|chain_id| *chain_id != self.chain_id && *chain_id != self.admin_id)
```
deuszx, line 1603: Shouldn't we synchronize the admin chain always?

deuszx (commented):
First pass.

linera-core/src/client/mod.rs
afck: Yes, but we already do that in `synchronize_from_validators`, so we avoid doing it again here. I'll add a comment.

linera-execution/src/system.rs
afck: Yes. Event values are supposed to be small anyway, but I can replace this with something like `contains_event`.

linera-execution/src/system.rs
afck: Yes, exactly. It's basically an error on the client side. We could disallow it, or just take the maximum.

linera-execution/src/execution.rs
afck: A contract could call `subscribe_to_events` at any point in its code, inside `execute_operation` or `execute_message`—or even `process_streams`.

linera-client/src/chain_listener.rs
```diff
@@ -277,8 +288,28 @@ impl<C: ClientContext> ChainListener<C> {
...
+        Ok(publishers)
+    }
+
+    /// Updates the event subscribers map, and returns all publisher chains we need to listen to.
+    async fn update_event_subscriptions(
+        &mut self,
+        chain_id: ChainId,
+    ) -> Result<BTreeSet<ChainId>, Error> {
+        let listening_client = self.listening.get_mut(&chain_id).expect("missing client");
+        if !listening_client.client.is_tracked() {
```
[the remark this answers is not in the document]
afck: Tracked chains are the ones that we want to receive messages on.
I think it's fair to assume that they are also the ones we want to receive events on.
So when there is a new event, we want to know about it if the _subscriber_ is tracked, regardless of whether the _publisher_ is tracked.

linera-client/src/chain_listener.rs
```diff
@@ -240,31 +246,36 @@ impl<C: ClientContext> ChainListener<C> {
...
-            let chain_ids = self.listening.keys().cloned().collect::<Vec<_>>();
-            for chain_id in chain_ids {
-                if chain_id != admin_id {
-                    self.maybe_process_inbox(chain_id).await?;
-                }
-            }
+        let Some(subscribers) = self.event_subscribers.get(&chain_id).cloned() else {
+            return Ok(());
+        };
+        for subscriber_id in subscribers {
```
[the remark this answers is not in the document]
afck: You mean you'd rename this to `subscriber_chain_id`? And `subscribers` to `subscriber_chains`?

But then `event_subscribers` should also be `event_subscriber_chains`?

If I understand correctly you're saying the words "publisher" and "subscriber" should apply to `(ChainId, ApplicationId)`, and therefore we should use "publishing chain" and "subscribing chain" for `ChainId`? I can try to make those changes across the board, just want to make sure I'm not misunderstanding.

linera-client/src/chain_listener.rs
```diff
@@ -240,31 +246,36 @@ impl<C: ClientContext> ChainListener<C> {
...
 
-    /// Start listening for notifications about the given chain.
-    async fn listen(&mut self, chain_id: ChainId) -> Result<(), Error> {
+    /// Starts listening for notifications about the given chain.
+    ///
+    /// Returns all publisher chains, that we also need to listen to.
+    async fn listen(&mut self, chain_id: ChainId) -> Result<BTreeSet<ChainId>, Error> {
         if self.listening.contains_key(&chain_id) {
-            return Ok(());
+            return Ok(BTreeSet::new());
```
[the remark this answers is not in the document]
afck: It shouldn't be possible, because if we're already listening, we would get a `NewBlock` notification, and whenever we get that we call `update_event_subscriptions`.

linera-execution/src/system.rs
afck: I moved this line to after we put the information into the transaction tracker.

Or do you mean we should update it only after calling `process_streams` in all the applications? That would complicate the code quite a bit, because we use the same code to call `process_streams`, regardless of whether we just processed `UpdateStreams` or any other operation or message that may have added a new subscription.

linera-core/src/client/mod.rs
deuszx: It'd be nice if we didn't have to think about this. Maybe a separate abstraction (struct) responsible for synchronizing other chains that holds the necessary information (like tracked chains) and a simple API that we can use to trigger the sync?

linera-execution/src/system.rs
deuszx: I was wondering whether this call here is needed at all? The assumption is that we have the events, it's not like we need to sync it here

afck: In the client maybe, but the server needs to verify that the events the block proposer claims exist actually exist.

linera-core/src/client/mod.rs
afck: Not sure what's the best abstraction here. This function is only called in one place, in `synchronize_from_validators`.

But actually, maybe we don't have to synchronize the admin chain there explicitly, and can just always consider the admin chain to be a publishing chain…

afck: Done in https://github.com/linera-io/linera-protocol/pull/3784/commits/50d0eaee76f7125aad57f0d4a5e9a0e06c76d7c5.

ma2bd (approved):

afck:
I mentioned yesterday that I might modify the `social` example to remove existing `received_posts` entries when we unsubscribe. However, `received_posts` is a map view, so if we re-subscribe, there is no danger of duplicating posts, so it's less important.

(I could do it anyway, if we think these posts should disappear on unsubscribing, but I'm not sure. It's also kind of unrelated to this PR now.)

linera-client/src/chain_listener.rs
```diff
@@ -196,6 +196,13 @@ impl<C: ClientContext> ChainListener<C> {
                 self.update_validators(&notification).await?;
                 self.update_wallet(notification.chain_id).await?;
                 self.add_new_chains(*hash).await?;
+                let publishers = self
+                    .update_event_subscriptions(notification.chain_id)
+                    .await?;
+                if !publishers.is_empty() {
```
deuszx, line 202: This if shouldn't be necessary - if `publishers` is empty then `listen_recursively` will return early.

afck: That's true. But maybe we should still avoid needlessly calling `process_inbox`.

linera-execution/src/system.rs
```diff
@@ -517,9 +517,36 @@ where
...
+                        stream_id,
+                        index,
+                    };
+                    ensure!(
+                        self.context()
+                            .extra()
+                            .contains_event(event_id.clone())
+                            .await?,
+                        ExecutionError::EventNotFound(event_id)
+                    );
```
deuszx, line 549: Wouldn't it make more sense to first check if we have the event, even before we add it to the txn_tracker?

linera-execution/src/runtime.rs
```diff
@@ -1361,14 +1364,22 @@ impl ContractRuntime for ContractSyncRuntimeHandle {
             application_id: application_id.into(),
         };
         let subscriber_app_id = this.current_application().id;
-        this.execution_state_sender
+        let next_index = this
+            .execution_state_sender
```
deuszx, line 1365: It's really unfortunate that we need to communicate via async channels to find out the next index of the event. Seems like an overkill and potential source of bugs (thing like this should. be atomic and single-threaded).

linera-execution/src/transaction_tracker.rs
```diff
@@ -160,6 +164,62 @@ impl TransactionTracker {
         self.operation_result = result
     }
 
+    pub fn add_stream_to_process(
```
deuszx, line 167: Maybe just `add_new_event_stream`?

linera-execution/src/transaction_tracker.rs
```diff
@@ -160,6 +164,62 @@ impl TransactionTracker {
...
+            .or_default()
+            .entry((chain_id, stream_id))
+            .and_modify(|(pi, ni)| {
+                *pi = (*pi).min(previous_index);
+                *ni = (*ni).max(next_index);
+            })
+            .or_insert_with(|| (previous_index, next_index));
+    }
+
+    pub fn remove_stream_to_process(
```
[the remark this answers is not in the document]
deuszx: Maybe `remove_event_stream` ?

deuszx (commented):
Another batch.

linera-execution/src/transaction_tracker.rs
afck: But this isn't necessarily a new stream: It can be called either because of a new subscription to a stream, or because of new events in an existing subscription's stream.

linera-execution/src/runtime.rs
afck: I agree! This execution state actor vs. runtime split is making a lot of things awkward. I want to look into this soon, but it'll be quite a large change (if it's feasible to resolve that at all).

linera-execution/src/system.rs
afck: It doesn't matter because if this fails, the transaction tracker is dropped anyway. And I can't easily reorder this because `self` is borrowed. I'd have to clone the whole `subscriptions.applications` collection.

linera-execution/src/transaction_tracker.rs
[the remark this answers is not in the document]
afck: Same as below: We're not really removing an event stream, just the note that we need to update a particular app about a particular stream. So I'm not sure the changed name would be better.

linera-storage/src/db_storage.rs
```diff
@@ -783,6 +783,14 @@ where
         maybe_value.ok_or_else(|| ViewError::EventsNotFound(vec![event_id]))
     }
 
+    async fn contains_event(&self, event_id: EventId) -> Result<bool, ViewError> {
+        let event_key = bcs::to_bytes(&BaseKey::Event(event_id))?;
+        let exists = self.store.contains_key(&event_key).await?;
+        #[cfg(with_metrics)]
+        CONTAINS_BLOB_COUNTER.with_label_values(&[]).inc();
```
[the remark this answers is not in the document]
afck: Oh, this should actually be a separate counter, not the blobs one! :facepalm: 

Anyway, good point, we're not really consistent there. For now I'd keep this the same way as for blobs, and count only successes?

linera-core/src/client/mod.rs
```diff
@@ -1526,6 +1591,31 @@ where
         }
     }
 
+    /// Synchronizes all chains that any application on this chain subscribes to.
+    /// We always consider the admin chain a relevant publishing chain, for new epochs.
+    async fn synchronize_publisher_chains(&self) -> Result<(), ChainClientError> {
```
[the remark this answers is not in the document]
afck: But here we specifically synchronize those chains that our own chain is subscribed to. We don't want to synchronize all tracked chains here.

linera-core/src/client/mod.rs
```diff
@@ -755,6 +755,26 @@ where
...
+            .system
+            .event_subscriptions
+            .indices()
+            .await?
+            .into_iter()
+            .map(|(chain_id, _)| chain_id)
+            .collect::<BTreeSet<_>>();
+        if self.chain_id != self.admin_id {
+            publishers.insert(self.admin_id);
+        }
```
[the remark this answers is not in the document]
afck: I don't understand how we could get rid of that: If our chain _is_ the admin chain, we don't need to synchronize the admin chain again here. If it isn't, we do.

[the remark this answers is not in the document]
deuszx: I wanted to say that we would filter out our own chain "deeper" in the callstack but in the other comment you wrote that sometimes we do want.

deuszx (approved):

Issue #2390, opened 2024-08-20
Replace `ChainDescription` with blobs or events


andresilva91:
Needs discussion as to wether we'll use blobs or events

afck:
As discussed, we should consider having a default chain ID for every account, which does not need to be created by any parent chain and gets initialized as soon as it receives a message.

ma2bd:
> As discussed, we should consider having a default chain ID for every account, which does not need to be created by any parent chain and gets initialized as soon as it receives a message.

That's a different task though

Issue #2351, opened 2024-08-06
Don't set the committee before checking it in `process_confirmed_block`.


afck:
[`process_confirmed_block` currently initializes the chain, including the committee](https://github.com/linera-io/linera-protocol/blame/9ee2d1d3e72e32ca2ab08d4c41845f0d42eeb575/linera-core/src/chain_worker/state/attempted_changes.rs#L261) before [checking the signatures based on that committee](https://github.com/linera-io/linera-protocol/blame/9ee2d1d3e72e32ca2ab08d4c41845f0d42eeb575/linera-core/src/chain_worker/state/attempted_changes.rs#L289).

ma2bd:
Good catch. This is recent (removal of OpenChain dependency for confirmed blocks). I believe the fix is to put committees in an event stream (#365)

ma2bd:
@afck Moving this to testnet2 due to the dependency to event streams. Perhaps we could add a `TODO(#2351)` in the code  now though?

bart-linera:
#3787 is currently introducing similar issues in other places, I'll tag them with this issue's number and fix them all collectively in a follow-up PR.

afck:
After #3787 the fix should now simply mean to strictly follow the rule again:

> The client should never put blobs into storage without first verifying the signatures of a certificate that uses that blob, and storing that certificate. This currently means executing the certificate (and all its ancestors), but after https://github.com/linera-io/linera-protocol/pull/3989 it only means adding that one block to the `unprocessed_blocks` set.

In addition, we can't consider a certificate from the chain itself to be proving its own chain description blob, unless we compare the committee to the admin chain.

PR #3787, opened 2025-04-16
Change ChainDescriptions into blobs

bart-linera:
## Motivation

This will allow for easier initialization of the network at genesis, among other things.

## Proposal

`ChainDescription` will become a blob, and `OpenChain` messages will be removed. Calling `open_chain` will create the blob, and the chain will actually get initialized when a block is proposed for it.

## Test Plan

Should be covered by CI.

## Release Plan

## Links

- #2390

### DIFF
--- a/examples/hex-game/tests/hex_game.rs
+++ b/examples/hex-game/tests/hex_game.rs
@@ -8,7 +8,7 @@
 use hex_game::{HexAbi, Operation, Timeouts};
 use linera_sdk::{
     linera_base_types::{
-        AccountSecretKey, Amount, ChainDescription, Secp256k1SecretKey, TimeDelta,
+        AccountSecretKey, Amount, BlobType, ChainDescription, Secp256k1SecretKey, TimeDelta,
     },
     test::{ActiveChain, QueryOutcome, TestValidator},
 };
@@ -34,8 +34,15 @@ async fn hex_game() {
         .await;
 
     let block = certificate.inner().block();
-    let message_id = block.message_id_for_operation(0, 0).unwrap();
-    let description = ChainDescription::Child(message_id);
+    let description = block
+        .created_blobs()
+        .into_iter()
+        .filter_map(|(blob_id, blob)| {
+            (blob_id.blob_type == BlobType::ChainDescription)
+                .then(|| bcs::from_bytes::<ChainDescription>(blob.content().bytes()).unwrap())
+        })
+        .next()
+        .unwrap();
     let mut chain = ActiveChain::new(key_pair1.copy(), description, validator);
 
     chain
@@ -108,8 +115,15 @@ async fn hex_game_clock() {
         .await;
 
     let block = certificate.inner().block();
-    let message_id = block.message_id_for_operation(0, 0).unwrap();
-    let description = ChainDescription::Child(message_id);
+    let description = block
+        .created_blobs()
+        .into_iter()
+        .filter_map(|(blob_id, blob)| {
+            (blob_id.blob_type == BlobType::ChainDescription)
+                .then(|| bcs::from_bytes::<ChainDescription>(blob.content().bytes()).unwrap())
+        })
+        .next()
+        .unwrap();
     let mut chain = ActiveChain::new(key_pair1.copy(), description, validator.clone());
 
     chain
--- a/linera-base/src/data_types.rs
+++ b/linera-base/src/data_types.rs
@@ -9,6 +9,7 @@ use std::ops;
 #[cfg(with_metrics)]
 use std::sync::LazyLock;
 use std::{
+    collections::BTreeMap,
     fmt::{self, Display},
     fs,
     hash::Hash,
@@ -37,6 +38,7 @@ use crate::{
         ApplicationId, BlobId, BlobType, ChainId, EventId, GenericApplicationId, ModuleId, StreamId,
     },
     limited_writer::{LimitedWriter, LimitedWriterError},
+    ownership::ChainOwnership,
     time::{Duration, SystemTime},
     vm::VmRuntime,
 };
@@ -706,6 +708,30 @@ impl Amount {
     }
 }
 
+/// What created a chain.
+#[derive(Eq, PartialEq, Ord, PartialOrd, Copy, Clone, Hash, Debug, Serialize, Deserialize)]
+pub enum ChainOrigin {
+    /// The chain was created by the genesis configuration.
+    Root(u32),
+    /// The chain was created by a call from another chain.
+    Child {
+        /// The parent of this chain.
+        parent: ChainId,
+        /// The block height in the parent at which this chain was created.
+        block_height: BlockHeight,
+        /// The index of this chain among chains created at the same block height in the parent
+        /// chain.
+        chain_index: u32,
+    },
+}
+
+impl ChainOrigin {
+    /// Whether the chain was created by another chain.
+    pub fn is_child(&self) -> bool {
+        matches!(self, ChainOrigin::Child { .. })
+    }
+}
+
 /// A number identifying the configuration of the chain (aka the committee).
 #[derive(Eq, PartialEq, Ord, PartialOrd, Copy, Clone, Hash, Default, Debug)]
 pub struct Epoch(pub u32);
@@ -784,12 +810,77 @@ impl Epoch {
     }
 }
 
+/// The configuration for a new chain.
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Clone, Serialize, Deserialize)]
+pub struct OpenChainConfig {
+    /// The ownership configuration of the new chain.
+    pub ownership: ChainOwnership,
+    /// The ID of the admin chain.
+    pub admin_id: Option<ChainId>,
+    /// The epoch in which the chain is created.
+    pub epoch: Epoch,
+    /// Blob IDs of the committees corresponding to epochs.
+    pub committees: BTreeMap<Epoch, Vec<u8>>,
+    /// The initial chain balance.
+    pub balance: Amount,
+    /// The initial application permissions.
+    pub application_permissions: ApplicationPermissions,
+}
+
+/// How to create a chain.
+#[derive(Eq, PartialEq, Ord, PartialOrd, Clone, Hash, Debug, Serialize, Deserialize)]
+pub struct ChainDescription {
+    origin: ChainOrigin,
+    timestamp: Timestamp,
+    config: OpenChainConfig,
+}
+
+impl ChainDescription {
+    /// Creates a new [`ChainDescription`].
+    pub fn new(origin: ChainOrigin, config: OpenChainConfig, timestamp: Timestamp) -> Self {
+        Self {
+            origin,
+            config,
+            timestamp,
+        }
+    }
+
+    /// Returns the [`ChainId`] based on this [`ChainDescription`].
+    pub fn id(&self) -> ChainId {
+        ChainId::from(self)
+    }
+
+    /// Returns the [`ChainOrigin`] describing who created this chain.
+    pub fn origin(&self) -> ChainOrigin {
+        self.origin
+    }
+
+    /// Returns a reference to the [`OpenChainConfig`] of the chain.
+    pub fn config(&self) -> &OpenChainConfig {
+        &self.config
+    }
+
+    /// Returns the timestamp of when the chain was created.
+    pub fn timestamp(&self) -> Timestamp {
+        self.timestamp
+    }
+
+    /// Whether the chain was created by another chain.
+    pub fn is_child(&self) -> bool {
+        self.origin.is_child()
+    }
+}
+
+impl BcsHashable<'_> for ChainDescription {}
+
 /// Permissions for applications on a chain.
 #[derive(
     Default,
     Debug,
     PartialEq,
     Eq,
+    PartialOrd,
+    Ord,
     Hash,
     Clone,
     Serialize,
@@ -1137,6 +1228,13 @@ impl BlobContent {
         BlobContent::new(BlobType::Committee, committee)
     }
 
+    /// Creates a new chain description [`BlobContent`] from a [`ChainDescription`].
+    pub fn new_chain_description(chain_description: &ChainDescription) -> Self {
+        let bytes = bcs::to_bytes(&chain_description)
+            .expect("Serializing a ChainDescription should not fail!");
+        BlobContent::new(BlobType::ChainDescription, bytes)
+    }
+
     /// Gets a reference to the blob's bytes.
     pub fn bytes(&self) -> &[u8] {
         &self.bytes
@@ -1198,7 +1296,7 @@ impl Blob {
         Blob::new(BlobContent::new_data(bytes))
     }
 
-    /// Creates a new contract bytecode [`BlobContent`] from the provided bytes.
+    /// Creates a new contract bytecode [`Blob`] from the provided bytes.
     pub fn new_contract_bytecode(compressed_bytecode: CompressedBytecode) -> Self {
         Blob::new(BlobContent::new_contract_bytecode(compressed_bytecode))
     }
@@ -1208,17 +1306,26 @@ impl Blob {
         Blob::new(BlobContent::new_evm_bytecode(compressed_bytecode))
     }
 
-    /// Creates a new service bytecode [`BlobContent`] from the provided bytes.
+    /// Creates a new service bytecode [`Blob`] from the provided bytes.
     pub fn new_service_bytecode(compressed_bytecode: CompressedBytecode) -> Self {
         Blob::new(BlobContent::new_service_bytecode(compressed_bytecode))
     }
 
-    /// Creates a new application description [`BlobContent`] from the provided
-    /// description.
+    /// Creates a new application description [`Blob`] from the provided description.
     pub fn new_application_description(application_description: &ApplicationDescription) -> Self {
         Blob::new(BlobContent::new_application_description(
             application_description,
         ))
+    }
+
+    /// Creates a new committee [`Blob`] from the provided bytes.
+    pub fn new_committee(committee: impl Into<Box<[u8]>>) -> Self {
+        Blob::new(BlobContent::new_committee(committee))
+    }
+
+    /// Creates a new chain description [`Blob`] from a [`ChainDescription`].
+    pub fn new_chain_description(chain_description: &ChainDescription) -> Self {
+        Blob::new(BlobContent::new_chain_description(chain_description))
     }
 
     /// A content-addressed blob ID i.e. the hash of the `Blob`.
@@ -1352,6 +1459,7 @@ doc_scalar!(
     Round,
     "A number to identify successive attempts to decide a value in a consensus protocol."
 );
+doc_scalar!(ChainDescription, "How to create a chain");
 doc_scalar!(OracleResponse, "A record of a single oracle response.");
 doc_scalar!(BlobContent, "A blob of binary data.");
 doc_scalar!(
--- a/linera-base/src/data_types.rs
+++ b/linera-base/src/data_types.rs
@@ -811,7 +811,7 @@ impl Epoch {
 }
 
 /// The initial configuration for a new chain.
-#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Clone, Serialize, Deserialize)]
+#[derive(Debug, PartialEq, Eq, Hash, Clone, Serialize, Deserialize)]
 pub struct InitialChainConfig {
     /// The ownership configuration of the new chain.
     pub ownership: ChainOwnership,
@@ -827,8 +827,8 @@ pub struct InitialChainConfig {
     pub application_permissions: ApplicationPermissions,
 }
 
-/// How to create a chain.
-#[derive(Eq, PartialEq, Ord, PartialOrd, Clone, Hash, Debug, Serialize, Deserialize)]
+/// Initial chain configuration and chain origin.
+#[derive(Eq, PartialEq, Clone, Hash, Debug, Serialize, Deserialize)]
 pub struct ChainDescription {
     origin: ChainOrigin,
     timestamp: Timestamp,
@@ -1459,7 +1459,10 @@ doc_scalar!(
     Round,
     "A number to identify successive attempts to decide a value in a consensus protocol."
 );
-doc_scalar!(ChainDescription, "How to create a chain");
+doc_scalar!(
+    ChainDescription,
+    "Initial chain configuration and chain origin."
+);
 doc_scalar!(OracleResponse, "A record of a single oracle response.");
 doc_scalar!(BlobContent, "A blob of binary data.");
 doc_scalar!(
--- a/linera-base/src/data_types.rs
+++ b/linera-base/src/data_types.rs
@@ -810,16 +810,16 @@ impl Epoch {
     }
 }
 
-/// The configuration for a new chain.
+/// The initial configuration for a new chain.
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Clone, Serialize, Deserialize)]
-pub struct OpenChainConfig {
+pub struct InitialChainConfig {
     /// The ownership configuration of the new chain.
     pub ownership: ChainOwnership,
     /// The ID of the admin chain.
     pub admin_id: Option<ChainId>,
     /// The epoch in which the chain is created.
     pub epoch: Epoch,
-    /// Blob IDs of the committees corresponding to epochs.
+    /// Serialized committees corresponding to epochs.
     pub committees: BTreeMap<Epoch, Vec<u8>>,
     /// The initial chain balance.
     pub balance: Amount,
@@ -832,12 +832,12 @@ pub struct OpenChainConfig {
 pub struct ChainDescription {
     origin: ChainOrigin,
     timestamp: Timestamp,
-    config: OpenChainConfig,
+    config: InitialChainConfig,
 }
 
 impl ChainDescription {
     /// Creates a new [`ChainDescription`].
-    pub fn new(origin: ChainOrigin, config: OpenChainConfig, timestamp: Timestamp) -> Self {
+    pub fn new(origin: ChainOrigin, config: InitialChainConfig, timestamp: Timestamp) -> Self {
         Self {
             origin,
             config,
@@ -855,8 +855,8 @@ impl ChainDescription {
         self.origin
     }
 
-    /// Returns a reference to the [`OpenChainConfig`] of the chain.
-    pub fn config(&self) -> &OpenChainConfig {
+    /// Returns a reference to the [`InitialChainConfig`] of the chain.
+    pub fn config(&self) -> &InitialChainConfig {
         &self.config
     }
 
--- a/linera-base/src/identifiers.rs
+++ b/linera-base/src/identifiers.rs
@@ -994,33 +994,33 @@ mod tests {
     use assert_matches::assert_matches;
 
     use super::{AccountOwner, BlobType};
+    use crate::{
+        data_types::{Amount, ChainDescription, ChainOrigin, Epoch, InitialChainConfig, Timestamp},
+        ownership::ChainOwnership,
+    };
 
-    /* TODO: figure out if this can be fixed.
-    /// Verifies that chain IDs that are explicitly used in some example and test scripts don't
-    /// change.
+    /// Verifies that the way of computing chain IDs doesn't change.
     #[test]
-    fn chain_ids() {
-        assert_eq!(
-            &ChainId::root(0).to_string(),
-            "aee928d4bf3880353b4a3cd9b6f88e6cc6e5ed050860abae439e7782e9b2dfe8"
+    fn chain_id_computing() {
+        let example_chain_origin = ChainOrigin::Root(0);
+        let example_chain_config = InitialChainConfig {
+            admin_id: None,
+            epoch: Epoch::ZERO,
+            ownership: ChainOwnership::single(AccountOwner::Reserved(0)),
+            balance: Amount::ZERO,
+            committees: [(Epoch::ZERO, vec![])].into_iter().collect(),
+            application_permissions: Default::default(),
+        };
+        let description = ChainDescription::new(
+            example_chain_origin,
+            example_chain_config,
+            Timestamp::from(0),
         );
         assert_eq!(
-            &ChainId::root(1).to_string(),
-            "a3edc33d8e951a1139333be8a4b56646b5598a8f51216e86592d881808972b07"
+            description.id().to_string(),
+            "f8c2f02adc0ac763ffd74a32d02aa37490d869605c356a2da55e1d09e7d253bf"
         );
-        assert_eq!(
-            &ChainId::root(2).to_string(),
-            "678e9f66507069d38955b593e93ddf192a23a4087225fd307eadad44e5544ae3"
-        );
-        assert_eq!(
-            &ChainId::root(9).to_string(),
-            "63620ea465af9e9e0e8e4dd8d21593cc3a719feac5f096df8440f90738f4dbd8"
-        );
-        assert_eq!(
-            &ChainId::root(999).to_string(),
-            "5487b70625ce71f7ee29154ad32aefa1c526cb483bdb783dea2e1d17bc497844"
-        );
-    }*/
+    }
 
     #[test]
     fn blob_types() {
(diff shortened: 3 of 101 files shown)


linera-base/src/data_types.rs
```diff
@@ -707,12 +709,179 @@ impl Amount {
...
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Clone, Serialize, Deserialize)]
+pub struct OpenChainConfig {
+    /// The ownership configuration of the new chain.
+    pub ownership: ChainOwnership,
+    /// The ID of the admin chain.
+    pub admin_id: Option<ChainId>,
+    /// The epoch in which the chain is created.
+    pub epoch: Epoch,
+    /// Blob IDs of the committees corresponding to epochs.
+    pub committees: BTreeMap<Epoch, Vec<u8>>,
```
ma2bd, line 824: Why `Vec<u8>` instead of a blob id?

linera-base/src/data_types.rs
```diff
@@ -707,12 +709,179 @@ impl Amount {
...
+
+    /// Tries to add one to this epoch's number. Returns an error if an overflow happens.
+    #[inline]
+    pub fn try_add_assign_one(&mut self) -> Result<(), ArithmeticError> {
+        self.0 = self.0.checked_add(1).ok_or(ArithmeticError::Overflow)?;
+        Ok(())
+    }
+}
+
+/// The configuration for a new chain.
```
ma2bd, line 814: nit: "The initial configuration..."

Perhaps `InitialChainConfig` is a better name as a result?

linera-base/src/identifiers.rs
```diff
@@ -1139,8 +1118,9 @@ mod tests {
 
     use assert_matches::assert_matches;
 
-    use super::{AccountOwner, BlobType, ChainId};
+    use super::{AccountOwner, BlobType};
 
+    /* TODO: figure out if this can be fixed.
```
ma2bd, line 1123: AFAICT you can remove

linera-core/src/client/mod.rs
```diff
@@ -287,6 +294,25 @@ where
...
+        validators: &[RemoteNode<impl ValidatorNode>],
+    ) -> Result<Box<ChainInfo>, LocalNodeError> {
+        match self.local_node.chain_info(chain_id).await {
+            Ok(info) => Ok(info),
+            Err(LocalNodeError::BlobsNotFound(blob_ids)) => {
+                let blobs =
+                    RemoteNode::download_blobs(&blob_ids, validators, self.blob_download_timeout)
+                        .await
+                        .ok_or(LocalNodeError::BlobsNotFound(blob_ids))?;
+                self.local_node.storage_client().write_blobs(&blobs).await?;
```
afck, line 309: But now the remote node could have tricked us, and the chain might not actually have been created.

I think we need to use something like `ChainClient::update_local_node_with_blobs_from`, which downloads proof from the validators that the blob was created. Except if we already have at least one correctly signed certificate from this chain—its first block can also count as a proof, I guess (if we double-check the committee information with the admin chain).

linera-core/src/client/mod.rs
```diff
@@ -296,10 +322,10 @@ where
...
     ) -> Result<Box<ChainInfo>, ChainClientError> {
         // Sequentially try each validator in random order.
-        let mut validators = validators.iter().collect::<Vec<_>>();
-        validators.shuffle(&mut rand::thread_rng());
-        for remote_node in validators {
-            let info = self.local_node.chain_info(chain_id).await?;
+        let mut validators_vec = validators.iter().collect::<Vec<_>>();
+        validators_vec.shuffle(&mut rand::thread_rng());
+        for remote_node in validators_vec {
+            let info = self.chain_info(chain_id, validators).await?;
```
afck, line 328: Actually, at this point we only need the next block height. Why not just set it to zero if we don't have the chain at all?

linera-core/src/client/mod.rs
```diff
@@ -403,6 +429,61 @@ where
...
+        let admin_epoch = admin_epoch.ok_or(ChainClientError::CommitteeSynchronizationError)?;
+        let remote_committee = admin_committees
+            .get(&admin_epoch)
+            .ok_or_else(|| ChainClientError::CommitteeDeprecationError)?;
+        // Recover history from the network.
+        let nodes = self.make_nodes(remote_committee)?;
+        let blob = RemoteNode::download_blob(&nodes, chain_desc_id, self.blob_download_timeout)
+            .await
+            .ok_or(LocalNodeError::BlobsNotFound(vec![chain_desc_id]))?;
+        self.local_node.storage_client().write_blob(&blob).await?;
```
afck, line 484: Same as above: We can't trust that blob yet.

linera-core/src/client/mod.rs
```diff
@@ -823,35 +909,8 @@ where
...
-            && info
-                .description
-                .ok_or_else(|| LocalNodeError::InactiveChain(self.chain_id))?
-                .is_child()
-        {
-            // The first incoming message of any child chain must be `OpenChain`. We must have it in
-            // our inbox, and include it before all other messages.
-            rearranged = IncomingBundle::put_openchain_at_front(&mut pending_message_bundles);
-            ensure!(rearranged, LocalNodeError::InactiveChain(self.chain_id));
-        }
```
afck, line 845: (Nice! Much less messy now! :tada:)

linera-core/src/remote_node.rs
```diff
@@ -314,7 +314,7 @@ impl<N: ValidatorNode> RemoteNode<N> {
     }
 
     #[instrument(level = "trace", skip(validators))]
-    async fn download_blob(
+    pub async fn download_blob(
```
afck, line 317: Let's add a doc comment that this downloads a blob but doesn't verify that it has actually been published and accepted by a quorum of validators.

linera-core/src/local_node.rs
```diff
@@ -294,6 +294,35 @@ where
...
+                    "could not find certificate with block chain ID and height {}",
+                    (chain_id, block_height),
+                )
+            })?;
+        Ok(certificate)
+    }
+
     /// Obtains the certificate containing the specified message.
     #[instrument(level = "trace", skip(self))]
     pub async fn certificate_for(
```
afck, line 328: Maybe we don't need this anymore now?

linera-chain/src/chain.rs
```diff
@@ -445,11 +443,30 @@ where
     }
 
     /// Invariant for the states of active chains.
-    pub fn ensure_is_active(&self) -> Result<(), ChainError> {
+    pub async fn ensure_is_active(&mut self, local_time: Timestamp) -> Result<(), ChainError> {
         if self.is_active() {
             Ok(())
```
afck, line 448: Maybe less indenting in the rest of the function.

```suggestion
            return Ok(());
```

linera-base/src/data_types.rs
mtpo201: great

ma2bd: The comment is inaccurate then, right?

afck: Ah, right! Yes, it is.

linera-core/src/client/mod.rs
bart-linera: We also need to return the info if `target_next_block_height` is zero, so we need to get it. (Maybe we could do a shortcut if it's not zero, but I'm not sure if it's worth it...)

examples/hex-game/tests/hex_game.rs
```diff
@@ -34,8 +34,15 @@ async fn hex_game() {
...
 
     let block = certificate.inner().block();
-    let message_id = block.message_id_for_operation(0, 0).unwrap();
-    let description = ChainDescription::Child(message_id);
+    let description = block
+        .created_blobs()
+        .into_iter()
+        .filter_map(|(blob_id, blob)| {
+            (blob_id.blob_type == BlobType::ChainDescription)
+                .then(|| bcs::from_bytes::<ChainDescription>(blob.content().bytes()).unwrap())
```
afck, line 42: Check the operation results instead, and make the operation return the child chain ID as a result? (Also below.)

linera-base/src/ownership.rs
```diff
@@ -21,7 +21,20 @@ use crate::{
 };
 
 /// The timeout configuration: how long fast, multi-leader and single-leader rounds last.
-#[derive(PartialEq, Eq, Clone, Hash, Debug, Serialize, Deserialize, WitLoad, WitStore, WitType)]
+#[derive(
+    PartialEq,
+    Eq,
+    PartialOrd,
```
afck, line 27: Is `[Partial]Ord` really necessary? (Also below.)

linera-chain/src/chain.rs
```diff
@@ -505,16 +522,20 @@ where
...
-            if let Some(config) = posted_message.message.matches_open_chain() {
-                if self.execution_state.system.description.get().is_none() {
-                    let message_id = chain_and_height.to_message_id(posted_message.index);
-                    self.execute_init_message(message_id, config, bundle.timestamp, local_time)
-                        .await?;
-                }
+        match self.ensure_is_active(local_time).await {
+            Ok(_) => (),
+            // if the only issue was that we couldn't initialize the chain because of a
+            // missing chain description blob, we might still want to process the inbox
```
afck, line 528:
```suggestion
            // missing chain description blob, we might still want to update the inbox
```

linera-chain/src/chain.rs
```diff
@@ -697,7 +672,7 @@ where
     /// Executes a block: first the incoming messages, then the main operation.
     /// Does not update chain state other than the execution state.
     #[expect(clippy::too_many_arguments)]
-    pub async fn execute_block_inner(
+    async fn execute_block_inner(
```
afck, line 675: (Good catch!)

linera-chain/src/unit_tests/chain_tests.rs
```diff
@@ -63,70 +60,108 @@ where
...
+    let mut env = TestEnvironment::new();
+
     let time = Timestamp::from(0);
-    let message_id = make_admin_message_id(BlockHeight(3));
-    let chain_id = ChainId::child(message_id);
-    let mut chain = ChainStateView::new(chain_id).await;
 
     // The size of the executed valid block below.
-    let maximum_block_size = 856;
+    let maximum_block_size = 260;
```
afck, line 159: (Nice! :star_struck:)

linera-chain/src/unit_tests/chain_tests.rs
```diff
@@ -63,70 +60,108 @@ where
...
+        height: u64,
+        config: InitialChainConfig,
+    ) -> ChainDescription {
+        let origin = ChainOrigin::Child {
+            parent: self.admin_id(),
+            block_height: BlockHeight(height),
+            chain_index: 0,
+        };
+        let config = InitialChainConfig {
+            admin_id: Some(self.admin_id()),
```
afck, line 142: This probably doesn't need to be an option anymore.

linera-client/src/client_context.rs
```diff
@@ -84,7 +85,7 @@ pub struct ClientContext<Env: Environment, W> {
 #[cfg_attr(web, async_trait(?Send))]
 impl<Env: Environment, W> chain_listener::ClientContext for ClientContext<Env, W>
 where
-    W: Persist<Target = Wallet> + 'static,
+    W: Persist<Target = Wallet> + Sync + 'static,
```
afck, line 88: Will this be an issue for the web client, @Twey?

linera-client/src/client_context.rs
```diff
@@ -432,28 +468,22 @@ where
         };
 
         // Download the parent chain.
```
afck, line 470: We don't have to do that at all anymore, I think.

linera-client/src/client_context.rs
```diff
@@ -432,28 +468,22 @@ where
...
+        let target_height = parent_height.try_add_one()?;
         self.client
-            .download_certificates(&nodes, message_id.chain_id, target_height)
+            .download_certificates(&nodes, parent_id, target_height)
             .await
             .context("downloading parent chain")?;
 
+        let config = chain_description.config();
+
         // The initial timestamp for the new chain is taken from the block with the message.
```
afck, line 479: The timestamp should also be in the chain description, right? So no need for this.

linera-core/src/chain_worker/state/attempted_changes.rs
```diff
@@ -303,13 +303,7 @@ where
         }
         let local_time = self.state.storage.clock().current_time();
         // TODO(#2351): This sets the committee and then checks that committee's signatures.
```
afck, line 305: (Right… that's still the case now, is it?)

linera-core/src/client/mod.rs
afck: This one I'd consider a blocker, or at least it should have a GitHub issue and be addressed right after this PR.

linera-core/src/client/mod.rs
afck: (Also kind of a blocker.)

linera-chain/src/unit_tests/chain_tests.rs
bart-linera: I think it has to be - an admin chain is still its own admin chain and it's impossible to store its own ID in its own `ChainDescription` :grimacing:

linera-execution/src/system.rs
```diff
@@ -318,6 +325,19 @@ where
         Some((*epoch, committee))
     }
 
+    /// Returns a map of epochs to blob IDs of corresponding committee blobs.
```
afck, line 328:
```suggestion
    /// Returns a map of epochs to serialized committees.
```

linera-client/src/client_context.rs
bart-linera: Good point!

linera-execution/src/system.rs
```diff
@@ -711,42 +737,48 @@ where
...
 
     /// Initializes the system application state on a newly opened chain.
-    pub fn initialize_chain(
-        &mut self,
-        message_id: MessageId,
-        timestamp: Timestamp,
-        config: OpenChainConfig,
-    ) {
+    pub async fn initialize_chain(&mut self, chain_id: ChainId) -> Result<(), ExecutionError> {
         // Guaranteed under BFT assumptions.
```
afck, line 748: Is that still the case? This isn't based on a cross-chain message anymore.
E.g. can it happen that we first synchronize a chain (without its parent), and then later also synchronize its parent, arriving at the block that created the chain description blob? Maybe we should instead just return if it's already initialized.

linera-core/src/chain_worker/state/attempted_changes.rs
bart-linera: Yes, I think it is :grimacing:

linera-execution/src/system.rs
```diff
@@ -711,42 +737,48 @@ where
...
+            .into_iter()
+            .map(|(epoch, serialized_committee)| {
+                let committee = bcs::from_bytes(&serialized_committee)
+                    .expect("Deserializing a committee shouldn't fail");
+                (epoch, committee)
+            })
+            .collect();
         self.committees.set(committees);
-        self.admin_id.set(Some(admin_id));
+        // If `admin_id` is `None`, this chain is its own admin chain.
```
afck, line 776: Right, nice! (Ignore my other comment about making this non-optional; doesn't work for the admin ID, of course.)

linera-execution/src/system.rs
```diff
@@ -769,27 +801,29 @@ where
...
+    ) -> Result<ChainId, ExecutionError> {
+        let chain_index = txn_tracker.next_chain_index();
+        let chain_origin = ChainOrigin::Child {
+            parent,
+            block_height,
+            chain_index,
+        };
+        let committees = self.get_committees();
+        let init_chain_config = config.init_chain_config(
+            (*self.epoch.get()).unwrap_or(Epoch::ZERO),
```
afck, line 817: (`None` should be unreachable. Might as well throw an error.)

linera-base/src/ownership.rs
bart-linera: IIRC at some point I used some `BTreeSet` with `ChainDescription`s and that created this requirement, but it does seem to be unnecessary now :+1:

linera-base/src/data_types.rs
```diff
@@ -784,12 +810,77 @@ impl Epoch {
...
+    pub epoch: Epoch,
+    /// Serialized committees corresponding to epochs.
+    pub committees: BTreeMap<Epoch, Vec<u8>>,
+    /// The initial chain balance.
+    pub balance: Amount,
+    /// The initial application permissions.
+    pub application_permissions: ApplicationPermissions,
+}
+
+/// How to create a chain.
```
afck, line 830: Maybe just "Initial chain configuration and chain origin."

linera-execution/tests/test_execution.rs
```diff
@@ -1174,45 +1204,57 @@ async fn test_open_chain() -> anyhow::Result<()> {
         ValidatorPublicKey::test_key(0),
         AccountPublicKey::test_key(0),
     )]);
-    let committees = BTreeMap::from([(Epoch::ZERO, committee)]);
+    let committees = BTreeMap::from([(
+        Epoch::ZERO,
+        bcs::to_bytes(&committee).expect("serializing a committee should not fail"),
```
afck, line 1209: It's fine to just `unwrap()` in tests. And this one even returns a result anyway:

```suggestion
        bcs::to_bytes(&committee)?,
```

afck, line 1209: It's fine to just `unwrap()` in tests. And this one even returns a result anyway:

```suggestion
        bcs::to_bytes(&committee)?,
```

linera-execution/tests/test_execution.rs
```diff
@@ -411,7 +421,9 @@ async fn test_rejecting_block_from_finalize() -> anyhow::Result<()> {
 #[tokio::test]
 async fn test_rejecting_block_from_called_applications_finalize() -> anyhow::Result<()> {
     let mut state = SystemExecutionState::default();
-    state.description = Some(ChainDescription::Root(0));
+    let description = dummy_chain_description(0);
+    let chain_id = description.id();
+    state.description = Some(description);
```
afck, line 426: We should extract this to a function, too. It's repeated in a lot of tests.

linera-execution/tests/test_execution.rs
```diff
@@ -1228,37 +1270,50 @@ async fn test_open_chain() -> anyhow::Result<()> {
...
+    let new_blob = &txn_outcome.blobs[0];
+    assert_eq!(new_blob.id().hash, child_id.0);
+    assert_eq!(new_blob.id().blob_type, BlobType::ChainDescription);
+    let created_description: ChainDescription = bcs::from_bytes(&new_blob.clone().into_bytes())
+        .expect("should deserialize a chain description");
+    assert_eq!(created_description.config().balance, Amount::ONE);
+    assert_eq!(created_description.config().ownership, child_ownership);
+    assert_eq!(created_description.config().committees, committees);
 
     // Initialize the child chain using the config from the message.
```
afck, line 1290:
```suggestion
    // Initialize the child chain using the new blob.
```

linera-execution/tests/test_execution.rs
```diff
@@ -1382,22 +1433,22 @@ async fn test_message_receipt_spending_chain_balance(
     let amount = Amount::ONE;
     let super_owners = receiving_chain_owner.into_iter().collect();
 
-    let mut view = SystemExecutionState {
-        description: Some(ChainDescription::Root(0)),
-        balance: amount,
-        ownership: ChainOwnership {
+    let description = dummy_chain_description_with_ownership_and_balance(
```
afck, line 1436: (…or maybe we need a `ChainDescriptionBuilder` for tests or something.)

linera-core/src/unit_tests/client_tests.rs
```diff
@@ -1953,6 +1889,9 @@ where
     Ok(())
 }
 
+// TODO: this test is currently intermittently failing if the faulty validators respond to
```
afck, line 1892: Please create an issue for it and cite it here.
We're trying to have no naked `TODO`s in the code, only `TODO(#1234)`.

linera-core/src/unit_tests/test_utils.rs
```diff
@@ -699,20 +700,13 @@ impl GenesisStorageBuilder {
...
             storage
-                .create_chain(
-                    initial_committee.clone(),
-                    admin_id,
-                    account.description,
-                    account.public_key.into(),
-                    account.balance,
-                    Timestamp::from(0),
-                )
+                .create_chain(account.description.clone())
```
afck, line 709: Are `public_key` and `balance` in `GenesisAccount` now redundant?

linera-core/src/unit_tests/test_utils.rs
```diff
@@ -807,77 +801,84 @@ where
         balance: Amount,
     ) -> anyhow::Result<ChainClient<B::Storage>> {
         // Make sure the admin chain is initialized.
-        if self.genesis_storage_builder.accounts.is_empty() && index != 0 {
+        if self.admin_description.is_none() && index != 0 {
             Box::pin(self.add_root_chain(0, Amount::ZERO)).await?;
```
afck, line 805: This was for tests that don't explicitly initialize the admin chain. Maybe it isn't needed anymore? Either way, it doesn't make sense to just initialize an arbitrary root chain here.

linera-core/src/unit_tests/test_utils.rs
```diff
@@ -807,77 +801,84 @@ where
...
         // Remember what's in the genesis store for future clients to join.
         self.genesis_storage_builder
-            .add(description, public_key, balance);
+            .add(description.clone(), public_key, balance);
         for validator in &self.validator_clients {
             let storage = self
                 .validator_storages
                 .get_mut(&validator.public_key)
                 .unwrap();
             if validator.fault_type().await == FaultType::Malicious {
```
afck, line 840: I actually think we should come up with a different fault instead. It would be better if they did have the same chain IDs, but somehow disagreed on the amount anyway.

linera-core/src/unit_tests/worker_tests.rs
```diff
@@ -79,294 +81,349 @@ use crate::{
...
-    I: IntoIterator<Item = (ChainDescription, AccountOwner, Amount)>,
     S: Storage + Clone + Send + Sync + 'static,
 {
-    let (committee, worker) = init_worker(
-        storage, /* is_client */ false, /* has_long_lived_services */ false,
-    );
-    for (description, owner, balance) in balances {
-        worker
+    async fn new(storage: S, is_client: bool, has_long_lived_services: bool) -> Self {
+        Self::new_with_amount(storage, is_client, has_long_lived_services, Amount::MAX).await
```
afck, line 111: (Another `Amount::MAX`? And people are surprised there's inflation!)

linera-core/src/unit_tests/worker_tests.rs
```diff
@@ -79,294 +81,349 @@ use crate::{
...
-    description: ChainDescription,
-    owner: AccountOwner,
-    balance: Amount,
-) -> (Committee, WorkerState<S>)
-where
-    S: Storage + Clone + Send + Sync + 'static,
-{
-    init_worker_with_chains(storage, [(description, owner, balance)]).await
-}
+    async fn add_child_chain(
```
afck, line 203: Do we ever actually do that? Insert a child chain just deus ex machina, without an `OpenChain` operation?

linera-core/src/unit_tests/worker_tests.rs
```diff
@@ -1298,67 +1295,31 @@ where
...
-                    balance,
-                    application_permissions: Default::default(),
-                })))
-                .to_posted(0, MessageKind::Protected),
-            ],
-        },
-        action: MessageAction::Accept,
-    };
+    let description = env
+        .add_child_chain(chain_2_desc.id(), sender_key_pair.public().into(), balance)
```
afck, line 1304: I see… but now this test makes less sense because there's no `OpenChain` message. (And the name is wrong, too.) I'd almost say this test can be removed. And then maybe we don't need `add_child_chain` anymore.

linera-execution/src/runtime.rs
```diff
@@ -110,6 +129,8 @@ pub struct SyncRuntimeInternal<UserInstance> {
     refund_grant_to: Option<Account>,
     /// Controller to track fuel and storage consumption.
     resource_controller: ResourceController,
+    /// Additional context for the runtime.
+    user_context: UserInstance::UserContext,
```
afck, line 133: Is it really worth creating that trait and associated type for this? We already have several optional fields in here that are `None` in some cases (operation vs. message vs. service call etc.).

linera-execution/src/runtime.rs
```diff
@@ -1431,27 +1459,31 @@ impl ContractRuntime for ContractSyncRuntimeHandle {
...
+        let (chain_id, txn_tracker_moved) = self
+            .inner()
             .execution_state_sender
             .send_request(|callback| ExecutionRequest::OpenChain {
                 ownership,
                 balance,
-                next_message_id: message_id,
+                parent_id,
+                block_height,
+                timestamp,
```
afck, line 1477: And the timestamp is actually not necessary here, because it's part of the system execution state.

linera-storage/src/lib.rs
```diff
@@ -181,41 +180,18 @@ pub trait Storage: Sized {
...
-    async fn create_chain(
-        &self,
-        committee: Committee,
-        admin_id: ChainId,
-        description: ChainDescription,
-        owner: AccountOwner,
-        balance: Amount,
-        timestamp: Timestamp,
-    ) -> Result<(), ChainError>
+    async fn create_chain(&self, description: ChainDescription) -> Result<(), ChainError>
```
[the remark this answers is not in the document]
bart-linera: There are quite a few call sites, so I think it still makes sense to have such a function - but I agree that it seems to be a bit outside of storage's scope. I'm not sure how to clean it up best right now, though.

linera-service/src/linera/net_up_utils.rs
```diff
@@ -282,17 +280,17 @@ async fn print_messages_and_create_faucet(
...
-        let faucet_chain = if let Some(faucet_chain) = faucet_chain {
-            ChainId::root(faucet_chain)
-        } else {
-            assert!(
-                num_other_initial_chains > 1,
-                "num_other_initial_chains must be greater than 1 if with_faucet is true"
-            );
-            ChainId::root(1)
-        };
+        let faucet_chain_idx = faucet_chain.unwrap_or(1);
```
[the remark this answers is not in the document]
bart-linera: Good point! The chain IDs are keys in a BTreeMap, so they would be sorted just by the numeric value of the hash, I guess, so it is possible that the chain at index 1 will happen to be the admin chain. I'll fix this :+1:

linera-service/src/linera/main.rs
```diff
@@ -150,8 +153,16 @@ impl Runnable for Job {
...
                 let timestamp = certificate.block().header.timestamp;
+                storage
+                    .write_blobs(
+                        &certificate
+                            .block()
+                            .created_blobs()
+                            .into_values()
+                            .collect::<Vec<_>>(),
+                    )
+                    .await?;
```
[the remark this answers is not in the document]
bart-linera: Good point. It looks like it already does it. I'll remove this part and if CI fails, I will be unpleasantly surprised :sweat_smile:

linera-service/src/linera/main.rs
```diff
@@ -1714,7 +1710,30 @@ async fn run(options: &ClientOptions) -> Result<i32, Error> {
...
+                admin_id: None,
+                balance: *initial_funding,
+                committees,
+                epoch: Epoch::ZERO,
+                application_permissions: Default::default(),
+                ownership: ChainOwnership::single(admin_key_pair.public().into()),
+            };
+            let admin_chain_description =
+                ChainDescription::new(origin, admin_config.clone(), timestamp);
+            let admin_id = admin_chain_description.id();
```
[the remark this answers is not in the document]
bart-linera: You're probably right, but I don't see an obvious way of doing this now - I'll take a note of it for later.

linera-execution/src/system.rs
bart-linera: The comment is wrong, but the only call site of this method checks if the chain is initialized first and only calls `initialize_chain` if it's not. But I guess we can return a boolean in the `Ok` variant telling the caller if the chain was already initialized, saving ourselves that check at the call site.

linera-sdk/src/test/validator.rs
```diff
@@ -298,26 +324,36 @@ impl TestValidator {
...
+            ownership: ChainOwnership::single(key_pair.public().into()),
+            committees: [(
+                epoch,
+                bcs::to_bytes(&committee).expect("Serializing a committee should not fail!"),
+            )]
+            .into_iter()
+            .collect(),
+            admin_id: None,
+            epoch,
+            balance: Amount::MAX,
```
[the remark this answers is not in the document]
bart-linera: I just needed _some_ value for the tests :sweat_smile: But the `MAX` was causing problems in a few places, let me change that.

linera-core/src/unit_tests/test_utils.rs
bart-linera: Balance, yes, but the `public_key` is still needed, because the information about it is lost in the description (it gets converted into an `AccountOwner` in the ownership).

linera-storage/src/lib.rs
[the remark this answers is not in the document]
bart-linera: #3856

afck (approved):
Approving, since as discussed the comments can also be addressed in follow-up PRs.

linera-service/src/linera/main.rs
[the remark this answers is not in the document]
bart-linera: #3857

linera-core/src/unit_tests/test_utils.rs
bart-linera: #3858

linera-core/src/unit_tests/worker_tests.rs
bart-linera: It's also used in one more test to simulate creating a block that opens a chain, without actually using it. So while we could remove this test, this function would still have some use.

linera-execution/src/runtime.rs
bart-linera: I'm not sure, honestly, but it does avoid some awkward unwrapping or returning errors from codepaths that should be unreachable.
I might even have started with an option and then decided that this is better, but I don't remember if that was how I introduced this, TBH :grimacing:

linera-execution/src/runtime.rs
bart-linera: I'm not sure what you mean :sweat_smile:  I believe we need this to initialize the new system execution state with the correct timestamp (and it might be different than the one stored in the current chain's system execution state until now), or am I missing something?

linera-base/src/data_types.rs
```diff
@@ -707,12 +709,179 @@ impl Amount {
...
+#[derive(Eq, PartialEq, Ord, PartialOrd, Clone, Hash, Debug, Serialize, Deserialize)]
+pub struct ChainDescription {
+    origin: ChainOrigin,
+    timestamp: Timestamp,
+    config: OpenChainConfig,
```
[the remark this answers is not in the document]
bart-linera: #3859

linera-client/src/client_context.rs
bart-linera: I actually removed it :sweat_smile: My thinking is that the proof will be downloaded in `ensure_has_chain_description` if we don't have the relevant blob (of course, currently it doesn't check the proof, but will be fixed as a part of #2351 ).

linera-execution/src/runtime.rs
afck: This is while executing the block on the parent chain, is it? Specifically, executing the `OpenChain` operation?

The timestamp we want to assign to the child chain is the timestamp of the current block. And yes, that is exactly the timestamp in the system execution state, because we update that at the _beginning_ of block execution.

bart-linera: Ah, I see! That makes sense. I'll try to apply your suggestion, then - could be that it will let me remove this whole trait thing :+1:

linera-service/tests/linera_net_tests.rs
```diff
@@ -3280,7 +3279,7 @@ async fn test_end_to_end_faucet_with_long_chains(config: impl LineraNetConfig) -
 
     // Since the faucet chain exceeds the configured maximum length, the faucet should have
     // switched after the first new chain.
-    assert!(other_outcome.message_id.chain_id != outcome.message_id.chain_id);
+    assert!(other_outcome.chain_id != outcome.chain_id);
```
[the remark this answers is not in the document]
bart-linera: Oops, you're right! :grimacing: Thanks!

PR #3794, opened 2025-04-17
correction README.md

strmfos:
that that - that

### DIFF
--- a/examples/how-to/perform-http-requests/README.md
+++ b/examples/how-to/perform-http-requests/README.md
@@ -30,7 +30,7 @@ service perform the HTTP request and return only the deterministic parts of the 
 
 ## HTTP requests using the service as an oracle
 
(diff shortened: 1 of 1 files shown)


examples/how-to/perform-http-requests/README.md
```diff
@@ -30,7 +30,7 @@ service perform the HTTP request and return only the deterministic parts of the
 
 ## HTTP requests using the service as an oracle
 
-The contract may call the service as an oracle. That means that that contracts sends a query to the
+The contract may call the service as an oracle. That means that contracts sends a query to the
```
afck, line 33:
```suggestion
The contract may call the service as an oracle. That means that contract sends a query to the
```

ALINASIR07: The contract may call the service as an oracle. That means that that contracts sends a query to the

afck (approved):

PR #3818, opened 2025-04-22
Add new applications to `used_blobs`; change `applications` query.

afck:
## Motivation

The `applications` query uses `used_blobs`, but newly created applications are not actually added to `used_blobs`.

## Proposal

Add new applications to `used_blobs`.

Make the `applications` query return the applications with an execution state.

## Test Plan

I added a regression test.

I also tested it locally:

```
cargo run --bin linera -- net up --with-faucet --faucet-port 8080
```

```
cargo run --bin linera -- wallet init --with-new-chain --faucet http://localhost:8080
cargo run --bin linera -- project publish-and-create examples/counter --json-argument "42"
cargo run --bin linera -- service --port 8081
```

Then `list_applications` doesn't show the new app. With this fix, it does.

## Release Plan

Since this changes the execution semantics, we could consider a different workaround for the testnet: Instead of using `used_blobs`, show the keys of `ExecutionStateView::users`.

## Links

### DIFF
--- a/linera-core/src/unit_tests/wasm_worker_tests.rs
+++ b/linera-core/src/unit_tests/wasm_worker_tests.rs
@@ -296,6 +296,10 @@ where
         )
         .await?;
     creator_state.system.timestamp.set(Timestamp::from(3));
+    creator_state
+        .system
+        .used_blobs
+        .insert(&application_description_blob_id)?;
     let run_block_proposal = ConfirmedBlock::new(
         BlockExecutionOutcome {
             messages: vec![Vec::new()],
@@ -303,7 +307,7 @@ where
             events: vec![Vec::new()],
             blobs: vec![Vec::new()],
             state_hash: creator_state.crypto_hash().await?,
-            oracle_responses: vec![vec![OracleResponse::Blob(application_description_blob_id)]],
+            oracle_responses: vec![vec![]],
             operation_results: vec![OperationResult(bcs::to_bytes(&15u64)?)],
         }
         .with(run_block),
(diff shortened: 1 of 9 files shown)


linera-execution/src/system.rs
```diff
@@ -799,7 +799,9 @@ where
         self.check_required_applications(&application_description, Some(&mut txn_tracker))
             .await?;
 
-        txn_tracker.add_created_blob(Blob::new_application_description(&application_description));
+        let blob = Blob::new_application_description(&application_description);
+        self.used_blobs.insert(&blob.id())?;
+        txn_tracker.add_created_blob(blob);
```
deuszx, line 804: Alternatively we could add all blobs at the end of the transaction, from the transactiont tracker. Rather than having this logic in multiple places.

linera-core/src/unit_tests/wasm_worker_tests.rs
```diff
@@ -296,14 +296,18 @@ where
...
+        .insert(&application_description_blob_id)?;
     let run_block_proposal = ConfirmedBlock::new(
         BlockExecutionOutcome {
             messages: vec![Vec::new()],
             previous_message_blocks: BTreeMap::new(),
             events: vec![Vec::new()],
             blobs: vec![Vec::new()],
             state_hash: creator_state.crypto_hash().await?,
-            oracle_responses: vec![vec![OracleResponse::Blob(application_description_blob_id)]],
+            oracle_responses: vec![vec![]],
```
deuszx, line 310: I don't understand what has changed in the code that this blob was moved from `oracle_responses` to `used_blobs` collection. The latter is not verified by the protocol?

afck: Since we already add it to `used_blobs` when creating the application, no further oracle responses are generated when it is used later.

(At first I thought this would cause a problem, but `requires_blob` already takes the created blobs into account, too, so no oracle response should be necessary here.)

PR #3819, opened 2025-04-22
Implement `BucketLogView` as a view and in the protocol.

MathieuDutSik:
## Motivation

The `LogView` that are used in the protocol have a type T that is constant in size. So, this can be grouped into a single bucket so as to avoid having too many keys.

## Proposal

The following is done:
* The `BucketQueueView` is introduced.
* A random test is introduced.
* The test is implemented in the protocol.

This is simpler than the `BucketQueueView`.

## Test Plan

A benchmark has been added that demonstrates that the Bucket version is 80% faster than the native version: 43.9ms for `memory_log_view` vs 24.2ms for `memory_bucket_log_view`.

The CI.

## Release Plan

Normal release.

## Links

None.

### DIFF
--- a/linera-views/src/views/bucket_log_view.rs
+++ b/linera-views/src/views/bucket_log_view.rs
@@ -135,6 +135,7 @@ where
         let mut delete_view = false;
         if self.delete_storage_first {
             batch.delete_key_prefix(self.context.base_key());
+            self.stored_data.clear();
             self.stored_count = 0;
             delete_view = true;
         }
--- a/linera-views/src/views/bucket_log_view.rs
+++ b/linera-views/src/views/bucket_log_view.rs
@@ -4,6 +4,7 @@
 #[cfg(with_metrics)]
 use std::sync::LazyLock;
 use std::{
+    cmp::min,
     collections::{BTreeSet, HashMap},
     ops::{Bound, Range, RangeBounds},
 };
@@ -347,8 +348,13 @@ where
     }
 
     async fn read_context(&self, range: Range<usize>) -> Result<Vec<T>, ViewError> {
+        println!("range={:?} count={} stored_count={}", range, self.count(), self.stored_count);
         let (i_bucket_start, position_start) = self.get_position(range.start).unwrap();
-        let (i_bucket_end, position_end) = self.get_position(range.end).unwrap();
+        let effective_end = min(range.end, self.stored_count);
+        let (i_bucket_end, position_end) = match self.get_position(effective_end) {
+            Some(pair) => pair,
+            None => (self.stored_data.len(), 0)
+        };
 
         let mut keys = Vec::new();
         let mut symbols = Vec::new();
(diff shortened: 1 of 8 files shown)


linera-views/src/views/bucket_log_view.rs
```diff
@@ -0,0 +1,492 @@
...
+    Index,
+}
+
+/// The `StoredIndices` contains the description of the stored buckets.
+#[derive(Clone, Debug, Default, Serialize, Deserialize)]
+struct StoredSizes {
+    /// The stored buckets with the first index being the size (at most N) and the
+    /// second one is the index in the storage. If the index is 0 then it corresponds
+    /// with the first value (entry `KeyTag::Front`), otherwise to the keys with
+    /// prefix `KeyTag::Index`.
```
afck, line 55: Is this comment outdated?

linera-views/src/views/bucket_log_view.rs
```diff
@@ -0,0 +1,492 @@
...
+        }
+    }
+}
+
+/// A view that supports logging values of type `T`.
+#[derive(Debug)]
+pub struct BucketLogView<C, T, const N: usize> {
+    context: C,
+    delete_storage_first: bool,
+    stored_data: Vec<Bucket<T>>,
```
afck, line 79: (I wonder at what length this becomes less efficient than the regular log view for e.g. just accessing the last element. We now always have to load the whole list of bucket sizes.)

linera-views/src/views/bucket_log_view.rs
MathieuDutSik: Yes, corrected.

linera-views/src/views/bucket_log_view.rs
```diff
@@ -0,0 +1,492 @@
...
+            }
+        } else {
+            Ok(
+                self.new_values[(start - effective_stored_count)..(end - effective_stored_count)]
+                    .to_vec(),
+            )
+        }
+    }
+
+    /// Read all the elements of the bucket log view
```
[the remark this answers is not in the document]
MathieuDutSik: Thank you.

afck (commented):
TBH I fear that the actual bucket sizes will always be 1 for `confirmed_log`, and _almost_ always 1 for the other two in practice, in which case I believe this would make things slower rather than faster?

ma2bd:
> TBH I fear that the actual bucket sizes will always be 1 for `confirmed_log`, and _almost_ always 1 for the other two in practice, in which case I believe this would make things slower rather than faster?

Do you mean the number of buckets?

linera-views/src/views/bucket_log_view.rs
```diff
@@ -0,0 +1,492 @@
...
+        }
+        Ok(result)
+    }
+
+    async fn read_context(&self, range: Range<usize>) -> Result<Vec<T>, ViewError> {
+        let mut indices = Vec::new();
+        for index in range {
+            indices.push(index);
+        }
+        let values = self.multi_get(indices).await?;
```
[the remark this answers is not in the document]
MathieuDutSik: Ok, done.

MathieuDutSik:
> TBH I fear that the actual bucket sizes will always be 1 for `confirmed_log`, and _almost_ always 1 for the other two in practice, in which case I believe this would make things slower rather than faster?

Indeed, if the size of blocks are going to be of size almost 1 then we will not gain anything. 
The PR could be merged without changing the protocol code. Comments welcomed.

PR #3822, opened 2025-04-23
Make sure `process_inbox` never creates empty blocks.

afck:
## Motivation

`process_inbox_without_prepare` collects the pending incoming messages and then calls `execute_block`. The latter first finalizes any pending blocks before actually creating the new block, so it collects the pending messages _again_, in case they have changed (a new block could have received some of them).

Since the mutex lock is only obtained inside `execute_block`, it can happen even just inside the local client that `process_inbox_without_prepare` sees pending messages, but those all disappeared once `execute_block` actually puts together the block, resulting in an empty block.

## Proposal

Acquire the lock earlier.

## Test Plan

```
export LINERA_FAUCET_URL="https://faucet.testnet-babbage.linera.net"
cargo test -p linera-service test_end_to_end_repeated_transfers::remote_net_grpc --features remote-net
```

failed for me, but passes with this fix.

## Release Plan

- Possibly release a new SDK patch version.

## Links

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -2103,9 +2103,6 @@ where
         blobs: Vec<Blob>,
         guard: Option<OwnedMutexGuard<()>>,
     ) -> Result<ExecuteBlockOutcome, ChainClientError> {
-        #[cfg(with_metrics)]
-        let _latency = metrics::EXECUTE_BLOCK_LATENCY.measure_latency();
-
         let _guard = if let Some(guard) = guard {
             guard
         } else {
@@ -2114,6 +2111,10 @@ where
             let mutex = self.state().client_mutex();
             mutex.lock_owned().await
         };
+
+        #[cfg(with_metrics)]
+        let _latency = metrics::EXECUTE_BLOCK_LATENCY.measure_latency();
+
         match self.process_pending_block_without_prepare().await? {
             ClientOutcome::Committed(Some(certificate)) => {
                 return Ok(ExecuteBlockOutcome::Conflict(certificate))


linera-core/src/client/mod.rs
```diff
@@ -2101,12 +2101,18 @@ where
...
 
-        let mutex = self.state().client_mutex();
-        let _guard = mutex.lock_owned().await;
+        let _guard = match guard {
+            Some(guard) => guard,
+            None => {
+                let mutex = self.state().client_mutex();
+                mutex.lock_owned().await
+            }
+        };
```
deuszx, line 2115: Can you add comment here explaining why we acquire a different lock if `guard` is None?

afck: Done in https://github.com/linera-io/linera-protocol/pull/3822/commits/c93bbbe8d391bd81fe128f567830d9b46470d819.

linera-core/src/client/mod.rs
```diff
@@ -3090,12 +3098,17 @@ where
 
         let mut certificates = Vec::new();
         loop {
+            let mutex = self.state().client_mutex();
+            let guard = mutex.lock_owned().await;
```
deuszx, line 3102: Why not lock at the beginning of this method and release when we leave it?

afck: We could, but then we don't allow other things to happen in between at all. E.g. if there are many incoming messages to process, this could prevent users from doing other mutations in the node service in between.

linera-core/src/client/mod.rs
```diff
@@ -2101,12 +2101,20 @@ where
         &self,
         operations: Vec<Operation>,
         blobs: Vec<Blob>,
+        guard: Option<OwnedMutexGuard<()>>,
```
Twey, line 2104: I guess the `Option` is just to allow the profiling to include taking the lock?  I think I'd rather it always did or always didn't, especially since there's no label in the metric to indicate whether we are taking the lock or not.

deuszx: It's for the case when we already took the lock in the outside context or not b/c this method is called in more than one place (and sometimes we have the lock and sometimes we don't).

afck: It's just so that I don't have to explicitly acquire the lock in the other call sites where we don't want the guard to live longer than the `execute_operation` call.

Twey: Right, I understand that, but in that case it should be equivalent to taking the lock and passing the (owned) guard as an argument, which also gives the metric the semantics of always being after the lock acquisition.

If the code to acquire the lock is too long or we expect it to change, we can always factor it into a function.

afck: Good point, fixed in https://github.com/linera-io/linera-protocol/pull/3822/commits/9aaf71c13004e8780fa57607c01d992b7d1bb481.

afck: This is hopefully only temporary anyway: I will try (on the main branch) to remove the whole client mutex, but that will be a much bigger change.

PR #3823, opened 2025-04-23
Introduce the execute_message for EVM applications.

### LINKED ISSUE #3760: Implement `execute_message` for EVM
The `execute_message` is the functionality that distinguishes Linera. So, it has to be coded and an operation `send_message` has to be coded for accessing the precompile.

MathieuDutSik:
## Motivation

We want to enable the multi-chain feature to the EVM chain. This is done here.

Fixes #3760 

## Proposal

The following design was adopted:
* An application hat is working in different chains is designed in the following way:
  - It is instantiated in one chain
  - The smart contract is then copied on other chains where it start from a blank state.
* In order to reproduce that design we adopt the following design:
  - EVM applications are of two types, the classic one which have one constructor and are created just once. This is the one running on Ethereum.
  - EVM applications running on multiple chains on Linera. They need to be different from the classic ones.
* The difference is introduced by a different `EvmInstantiation` type.
* The EVM apps running on multiple chains need to have the following features:
  - Their constructor need to be trivial. This is because when an app is instantiated on other chains, the constructor cannot be run.
  - They need to have a function `function instantiate(bytes input)` that is run on the initial chain.
  - They need to have a function `function execute_message(bytes input)` that execute the messages.

Regarding the code, the following changes are made:
* We have access to the EVM bytecode, but this is not the same as the published bytecode. Therefore, we have access to the module, but we need to instantiate to do that.
* For the contract, this is easy, but for the service not so much since we cannot write to the storage.
* Therefore, we introduce the `changes` entry in the `DatabaseRuntime<Runtime>`. For normal contract writing this markedly simplifies the code, since the effective writing is done outside of the code.
* For service, the changes are obtained from the first initial write which is then used for the `basic_ref` calls. For the `storage_ref` nothing needs to be done since the returned value is zero by default.

## Test Plan

An end-to-end test has been introduced.

## Release Plan

The goal is to publish the EVM features for the next TestNet.

## Links

None.


linera-service/tests/linera_net_tests.rs
```diff
@@ -708,6 +714,130 @@ async fn test_evm_call_wasm_end_to_end_counter(config: impl LineraNetConfig) ->
...
+    assert_eq!(counter_value, original_value);
+    let result = application2.run_json_query(query.clone()).await?;
+    let counter_value = read_evm_u64_entry(result);
+    assert_eq!(counter_value, 0);
+
+    // Second: executing the movement of assets
+
+    let chain_id: [u64; 4] = <[u64; 4]>::from(chain2.0);
+    let chain_id: [u8; 32] = linera_base::crypto::u64_array_to_be_bytes(chain_id);
+    let chain_id: B256 = chain_id.into();
```
ma2bd, line 812: Should we have a conversion function from ChainId to B256 ?

linera-execution/src/evm/mod.rs
```diff
@@ -35,3 +40,12 @@ pub enum EvmExecutionError {
     #[error("The operation was halted")]
     Halt { gas_used: u64, reason: HaltReason },
 }
+
+/// The possible types of instantiating an EVM contract
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub enum EvmInstantiation {
+    /// A classic EVM contract which is not shared between chains.
+    Classic(Vec<u8>),
```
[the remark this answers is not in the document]
MathieuDutSik: It is needed if we want to be able to run existing Ethereum applications on Linera.
Of course, we cannot prevent the application from being instantiated on different chains. But that is not an issue as they would not be able to communicate.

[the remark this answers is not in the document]
ma2bd: Other applications would communicate and could act as a intended bridge.

MathieuDutSik:
> Is the distinction necessary at all?
> 
> E.g. we could decide to always pass the parameters into the constructor. (And possibly a flag indicating whether this is the creation chain.)
> 
> And we could not call `initialize` if the initialization arguments are empty, so it wouldn't fail if it doesn't implement `initialize` either.
> 
> And a contract that doesn't implement `execute_message` should simply fail if it receives a message. (Which shouldn't happen anyway, unless it _sends_ a message.)

Overall, that seems like a good idea:
* I was not daring to pass the constructor argument into the parameters, but if that is ok, then so be it.
* About the flag, I propose to add instead a function `Linera.is_creation_chain()` to the precompile that calls the runtime and  determines whether `runtime.chain_id() == runtime.application_creator_chain_id()`.
* Instead of `initialize` I think you mean `instantiate`. And yes, running it conditionally is good.
* And a contract that does not implement `execute_message` and does not do `send_message` would still work.

The good thing is that it allows a classic EVM application to still run fine on Linera.

linera-execution/src/evm/mod.rs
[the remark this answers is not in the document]
MathieuDutSik: Yes, that would also be possible.
In general, I think it is a good idea to offer different options to developpers.

MathieuDutSik:
> Is the distinction necessary at all?
> 
> E.g. we could decide to always pass the parameters into the constructor. (And possibly a flag indicating whether this is the creation chain.)
> 
> And we could not call `initialize` if the initialization arguments are empty, so it wouldn't fail if it doesn't implement `initialize` either.
> 
> And a contract that doesn't implement `execute_message` should simply fail if it receives a message. (Which shouldn't happen anyway, unless it _sends_ a message.)

Ok, I changed the code according to this design.

linera-service/tests/linera_net_tests.rs
```diff
@@ -708,6 +707,123 @@ async fn test_evm_call_wasm_end_to_end_counter(config: impl LineraNetConfig) ->
...
+    sol! {
+        function move_value_to_chain(bytes32 chain_id, uint64 moved_value);
+        function get_value();
+    }
+
+    let constructor_argument = Vec::new();
+
+    let instantiation_argument: Vec<u8> = u64::abi_encode(&original_value);
+
+    // Publishing the EVM contract and instantiating it on the two chains.
```
afck, line 748: Applications are _instantiated_ only on one chain. (I.e. `instantiate` is called only on the creator chain, at least for Wasm apps.)

linera-execution/src/evm/revm.rs
```diff
@@ -223,12 +260,16 @@ impl UserServiceModule for EvmServiceModule {
 // functionalities accessed from the EVM.
 const PRECOMPILE_ADDRESS: Address = address!("000000000000000000000000000000000000000b");
 
-fn u8_slice_to_application_id(vec: &[u8]) -> ApplicationId {
+fn u8_slice_to_cryptohash(vec: &[u8]) -> CryptoHash {
```
afck, line 263: There's already an `impl TryFrom<&[u8]> for CryptoHash`.

linera-execution/src/evm/revm.rs
```diff
@@ -239,12 +280,18 @@ fn address_to_user_application_id(address: Address) -> ApplicationId {
...
     TryCallApplication,
-    /// Key prefix for the try_query_application
+    /// Key prefix for try_query_application
     TryQueryApplication,
+    /// Key prefix for send_message
+    SendMessage,
+    /// Key prefix for message_id
+    MessageId,
+    /// Key prefix for message_is_bouncing
+    MessageIsBouncing,
```
afck, line 294:
```suggestion
   /// Key prefix for `try_call_application`
    TryCallApplication,
    /// Key prefix for `try_query_application`
    TryQueryApplication,
    /// Key prefix for `send_message`
    SendMessage,
    /// Key prefix for `message_id`
    MessageId,
    /// Key prefix for `message_is_bouncing`
    MessageIsBouncing,
```

linera-execution/src/evm/revm.rs
```diff
@@ -586,6 +688,32 @@ where
         Self { module, db }
     }
 
+    /// Executes the transaction. If needed instantiate the contract
```
afck, line 691:
```suggestion
    /// Executes the transaction. If needed initializes the contract.
```

linera-execution/src/evm/revm.rs
```diff
@@ -586,6 +688,32 @@ where
...
+    fn init_transact_commit_tx_data(
+        &mut self,
+        ch: Choice,
+        vec: &[u8],
+    ) -> Result<ExecutionResultSuccess, ExecutionError> {
+        // In case of a not yet instanstated application, we need first to create it.
```
afck, line 697: Isn't the application necessarily already instantiated, in Linera's sense? It's about initializing the EVM contract, isn't it?

linera-execution/src/evm/revm.rs
```diff
@@ -586,6 +688,32 @@ where
...
+        if !self.db.is_initialized()? {
+            self.initialize_contract()?;
+        }
+        self.transact_commit_tx_data(ch, vec)
+    }
+
+    /// Executes the transaction. If needed instantiate the contract
```
afck, line 704: Wrong comment?

linera-execution/src/evm/revm.rs
```diff
@@ -696,21 +826,51 @@ impl<Runtime> RevmServiceInstance<Runtime>
 where
     Runtime: ServiceRuntime,
 {
-    fn transact_tx_data(&mut self, vec: &[u8]) -> Result<ExecutionResultSuccess, ExecutionError> {
+    fn init_transact_tx_data(
```
afck, line 829: Doesn't "tx" stand for "transaction", too? :sweat_smile:

linera-execution/src/evm/database.rs
```diff
@@ -107,39 +96,101 @@ where
...
+                    tracing::info!("basic_ref Some: |code|={}", code.len());
+                } else {
+                    tracing::info!("basic_ref Some: |code|=None");
+                }
+            } else {
+                tracing::info!("basic_ref None");
+            }
+            Ok(account_info)
+        } else {
+            Ok(Some(AccountInfo::default()))
```
afck, line 142: Maybe early return?

linera-execution/src/evm/database.rs
```diff
@@ -107,39 +96,101 @@ where
...
+        &self,
+        _code_hash: B256,
+    ) -> Result<revm::primitives::Bytecode, ExecutionError> {
+        panic!("Functionality code_by_hash_ref not implemented");
+    }
+
+    fn storage_ref(&self, address: Address, index: U256) -> Result<U256, ExecutionError> {
+        tracing::info!("storage_ref call at address={address:?} index={index:?}");
+        if !self.changes.is_empty() {
+            let account = self.changes.get(&address).unwrap();
```
afck, line 156: Is it okay to panic here?

linera-execution/src/evm/database.rs
```diff
@@ -107,39 +96,101 @@ where
...
+            return Ok(match account.storage.get(&index) {
+                None => U256::ZERO,
+                Some(slot) => slot.present_value(),
+            });
         }
+        let val = self.get_contract_address_key(&address);
+        let Some(val) = val else {
+            panic!("There is no storage associated to Externally Owned Account");
```
afck, line 164: …and here. Is this all really unreachable, even with malicious EVM code?

```suggestion
            panic!("There is no storage associated to externally owned account");
```

afck, line 164: …and here. Is this all really unreachable, even with malicious EVM code?

```suggestion
            panic!("There is no storage associated to externally owned account");
```

linera-service/tests/linera_net_tests.rs
MathieuDutSik: Removed.

linera-execution/src/evm/revm.rs
MathieuDutSik: Ok, changed.

linera-execution/src/evm/revm.rs
MathieuDutSik: Corrected.

linera-execution/src/evm/revm.rs
MathieuDutSik: Yes, that is that. Linera instantiated, but not according to the EVM.
So, I changed the comment.

linera-execution/src/evm/database.rs
MathieuDutSik: Ok, if you prefer.

linera-execution/src/evm/database.rs
MathieuDutSik: Ok, the terminology "EOA" is common in Ethereum world.

linera-execution/src/evm/database.rs
MathieuDutSik: Yes, the query is about a specific contract address.
But in that case, there is just one (actually two with the ZERO contract which exists for a mysterious reason) and the other contract.

All the other contracts are being called via the precompile or the contract call interception.

linera-execution/solidity/linera.sol
```diff
@@ -5,8 +5,101 @@ pragma solidity ^0.8.0;
...
+    }
+    return (pos + 4, value);
+  }
+
+
+
+
+
+
+
```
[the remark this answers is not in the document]
MathieuDutSik: Not really.
Now, I applied the `forge fmt` for reformatting the solidity source code.

linera-execution/solidity/linera.sol
```diff
@@ -5,8 +5,101 @@ pragma solidity ^0.8.0;
...
+// 3: message_id
+// 4: message_is_bouncing
 
 library Linera {
+  function bcs_deserialize_offset_bool(uint256 pos, bytes memory input) internal pure returns (uint256, bool) {
+    uint8 val = uint8(input[pos]);
+    bool result = false;
+    if (val == 1) {
+      result = true;
+    }
```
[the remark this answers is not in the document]
MathieuDutSik: Ok. Corrected here.
This should be done as well in `serde-reflection`.

linera-execution/solidity/linera.sol
```diff
@@ -5,8 +5,101 @@ pragma solidity ^0.8.0;
...
 // 0: try_call_application
 // 1: try_query_application
+// 2: send_message
+// 3: message_id
+// 4: message_is_bouncing
 
 library Linera {
+  function bcs_deserialize_offset_bool(uint256 pos, bytes memory input) internal pure returns (uint256, bool) {
+    uint8 val = uint8(input[pos]);
+    bool result = false;
```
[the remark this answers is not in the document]
MathieuDutSik: Not really. The EVM does bassically those check for us. Any access out of bounds will create a "Revert" statement, that is abort the transaction.

linera-execution/src/evm/revm.rs
MathieuDutSik: Ok, thanks.

linera-execution/solidity/linera.sol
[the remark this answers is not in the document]
afck: Ah, great! But then we don't need the `requires` in the other functions either, do we?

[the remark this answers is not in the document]
MathieuDutSik: Yes, correct, eliminated.

linera-execution/solidity/linera.sol
```diff
@@ -5,8 +5,101 @@ pragma solidity ^0.8.0;
...
+    if (val == 1) {
+      result = true;
+    }
+    return (pos + 1, result);
+  }
+
+  function bcs_deserialize_offset_bytes32(uint256 pos, bytes memory src) internal pure returns (uint256, bytes32) {
+    bytes32 dest;
+    require(src.length >= pos + 32, "out of bounds");
+    assembly { dest := mload(add(add(src, 0x20), pos)) }
```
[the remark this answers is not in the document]
MathieuDutSik: It is correct.

The bytes is actually decomposed in the following way:
* The first 32 bytes contains the size of the bytes. And Ethereum being Ethereum they encoded the length as a 256 bytes which is suboptimal but so be it.
* The data of the bytes.

So, the first addition of 32 is to skip the bytes containing the size. The next one is for the position.

To be sure, I created a solidity code that uses this code and then back to bytes. The comparison of the value show that it is ok.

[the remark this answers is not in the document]
afck: I see! Should we then assert that the first 32 bytes have the value 0x000…0020?

afck (approved):
No blockers, but I can't easily see the changes you made based on my comments. Clicking on the first "force-push" link doesn't work.

linera-execution/src/evm/revm.rs
MathieuDutSik: Ok, thank you. Missed that.
I restructured the code accordingly.

linera-execution/src/evm/database.rs
MathieuDutSik: The address in question is the one of the running contract.
It can only be Zero or the one of the contract in question. This cannot be passed back by another application.

Issue #3007, opened 2024-12-04
Generalize `linera-storage` to an injection mechanism for all `linera-core` dependencies


Twey:
We would like to do dependency injection for a variety of effectful dependencies of `linera-core` in order to keep it sans-I/O:
- the storage system
- the clock
- the wallet storage
- the network
- (soon, probably) the block signer

Currently this process is handled for the storage and clock by the (now misnamed) `linera-storage` crate, which defines one big dependency interface for `linera-client`, and `linera-service`, which performs injection of the dependencies.  The wallet storage, rather than being injected, is duplicated, with concurrent tasks that listen for changes to the in-memory wallet and update the stored wallet; and the network implementation is statically injected into the constructor.

We should at the least rename `linera-storage` to reflect its generalization from just storage, and ideally figure out a unified way to inject the dependencies individually at the provider.

PR #3828, opened 2025-04-24
Bundle client dependencies into an `Environment` trait

Twey:
## Motivation

Working towards https://github.com/linera-io/linera-protocol/issues/3007.  By introducing an ‘end-to-end principle’ for our client dependencies, we can restructure our dependencies as necessary without rewriting huge amounts of code each time.

## Proposal

Introduce a new `Environment` trait in `linera-core` that ~provides~ will provide all external runtime dependencies necessary for the client to run.  By propagating this trait bound instead of the individual dependencies, only the provider and consumer need to be aware of the comprising dependencies, with the code in between agnostic.

## Test Plan

CI.

## Release Plan

## Links

### DIFF
--- a/linera-core/src/environment.rs
+++ b/linera-core/src/environment.rs
@@ -43,6 +43,7 @@ impl<S: Storage, N: Network> Environment for Impl<S, N> {
     fn storage(&self) -> &S {
         &self.storage
     }
+
     fn network(&self) -> &N {
         &self.network
     }


linera-core/src/client/mod.rs
```diff
@@ -487,13 +476,10 @@ pub struct ChainClientOptions {
 /// * As a rule, operations are considered successful (and communication may stop) when
 ///   they succeeded in gathering a quorum of responses.
 #[derive(Debug)]
-pub struct ChainClient<ValidatorNodeProvider, Storage>
-where
-    Storage: linera_storage::Storage,
-{
+pub struct ChainClient<Env: Environment> {
```
ma2bd, line 479: To be clear, we still would like to remove the trait constraint (forced by async graphql IIRC) eventually.

ma2bd: nit: Could we use `E` instead of `Env` ?

ma2bd (commented):
Looks great directionally. I imagine we should be able to move the `Clock` (and maybe other things) out of `linera-storage` after that so that this crate is only about .. storage.

linera-core/src/client/mod.rs
Twey: Right, absolutely.  There's a neat trick for that: wherever we have

```rust
struct Foo<T: HasBar> {
    x: T,
    y: T::Bar,
}
```

we can transform it into:

```rust
struct Foo<T, TBar = <T as HasBar>::Bar> {
  x: T,
  y: TBar,
}
```

with no loss of ergonomics for the user.  But currently it's not worth doing this because the bounds are forced anyway by both `async-graphql` and also `linera-views`, which requires a type that `: Context` in its views, propagating the bound up from the `chain_worker`.

Twey:
When I went to rename `linera-storage` I realized that the main functionality in `linera-storage` is actually currently the Wasm module cache — all the pure storage functionality is already handled in `linera-views`.  I'd like to take the clock and Wasm runtime out using this framework and then move the Wasm module cache into `linera-core`, absorbing `linera-storage`.

linera-core/src/client/mod.rs
Twey: On `E` rather than `Env`: we use (and Rust uses, generally) `E` to mean `Error` a lot, so I wanted to distinguish this parameter.

linera-storage/src/db_storage.rs
```diff
@@ -267,7 +267,7 @@ impl BatchExt for Batch {
 
 /// Main implementation of the [`Storage`] trait.
 #[derive(Clone)]
-pub struct DbStorage<Store, Clock> {
+pub struct DbStorage<Store, Clock = WallClock> {
```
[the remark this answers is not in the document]
Twey: It will, but I'm trying to keep the PRs smallish :)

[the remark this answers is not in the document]
Twey: (I realize that this comment is actually on just such an extra change — oops :sweat_smile: but it's a rather small one)

linera-core/src/environment.rs
```diff
@@ -0,0 +1,57 @@
...
+impl<S: Storage, N: Network> Environment for Impl<S, N> {
+    type Storage = S;
+    type Network = N;
+    type ValidatorNode = N::Node;
+    type StorageContext = S::Context;
+
+    fn storage(&self) -> &S {
+        &self.storage
+    }
+    fn network(&self) -> &N {
```
afck, line 46:
```suggestion
    }

    fn network(&self) -> &N {
```

linera-core/src/environment.rs
```diff
@@ -0,0 +1,57 @@
...
+        &self.network
+    }
+}
+
+cfg_if::cfg_if! {
```
afck, line 51: Is that just to avoid repeating `#[cfg(with_testing)]` three times?

afck (approved):
Good idea! :+1:

linera-core/src/environment.rs
Twey: Pretty much, yes.

PR #3833, opened 2025-04-27
Add the computation of the BlockEnv for EVM.

MathieuDutSik:
## Motivation

Several variables such as block height are available in the `block` variable in the EVM. Support was missing.

## Proposal

The `BlockEnv` is computed from the `runtime`. The `gas_limit` could not be put since the limit in `Resources`
does not appear to be available from the `runtime`.

## Test Plan

The CI.

## Release Plan

Normal release plan.

## Links

Documentation on the `BlockEnv` is available in https://docs.rs/revm-primitives/15.2.0/revm_primitives/env/struct.BlockEnv.html

### DIFF
--- a/linera-execution/src/evm/database.rs
+++ b/linera-execution/src/evm/database.rs
@@ -274,7 +274,10 @@ where
         let prevrandao = keccak256(entry.as_bytes());
         // The blob excess gas and price is not relevant to the execution
         // on Linera. We set up a default value as in REVM.
-        let entry = BlobExcessGasAndPrice { excess_blob_gas: 0, blob_gasprice: 1 };
+        let entry = BlobExcessGasAndPrice {
+            excess_blob_gas: 0,
+            blob_gasprice: 1,
+        };
         let blob_excess_gas_and_price = Some(entry);
         Ok(BlockEnv {
             number: block_height_evm,
--- a/linera-execution/src/evm/database.rs
+++ b/linera-execution/src/evm/database.rs
@@ -252,9 +252,8 @@ where
         // The block height being used
         let block_height_linera = runtime.block_height()?;
         let block_height_evm = U256::from(block_height_linera.0);
-        // The coinbase receiving the minted ethereum in Proof Of Work.
-        // Not relevant anymore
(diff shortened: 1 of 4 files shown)


linera-execution/src/evm/database.rs
```diff
@@ -241,3 +242,48 @@ where
...
+        let gas_limit = U256::MAX;
+        // The timestamp. Both the EVM and Linera use the same UNIX epoch.
+        // But the Linera epoch is in microseconds since the start and the
+        // Ethereum epoch is in seconds
+        let timestamp_linera = runtime.read_system_timestamp()?;
+        let timestamp_evm = U256::from(timestamp_linera.micros() / 1_000_000);
+        // The basefee is the minimum feee for executing. We have no such
+        // concept in Linera
+        let basefee = U256::ZERO;
+        // The randomness beacon being used.
```
afck, line 271: This should probably be moved two lines down.

MathieuDutSik: Ok.

linera-execution/src/evm/database.rs
```diff
@@ -241,3 +242,52 @@ where
...
+    Runtime: BaseRuntime,
+{
+    pub fn get_block_env(&self) -> Result<BlockEnv, ExecutionError> {
+        let mut runtime = self.runtime.lock().expect("The lock should be possible");
+        // The block height being used
+        let block_height_linera = runtime.block_height()?;
+        let block_height_evm = U256::from(block_height_linera.0);
+        // The coinbase receiving the minted ethereum in Proof Of Work.
+        // Not relevant anymore
+        let coinbase = address!("00000000000000000000000000000000000000bb");
```
deuszx, line 257: In[ the docs](https://reth.rs/docs/reth/revm/revm/context/struct.BlockEnv.html) it says
> This is the receiver address of all the gas spent in the block.

which I can imagine this being useful for us at some point in the future.

Let's rename to _beneficiary_ and copy-paste the comment from the REVM as explanation.

deuszx (approved):
other than the rename and comment - LGTM

Issue #3839, opened 2025-04-28
Fetch timeout certificates when synchronizing from validators.


afck:
Currently timeout certificates are not fetched from the validators, so the client doesn't know the latest round.

PR #3840, opened 2025-04-28
When syncing, also process timeout certificates.

afck:
## Motivation

When syncing a chain, a client needs to also process the latest timeout certificate from the validators, so that it knows what the current round is, and it can propose blocks. This is currently missing in `try_synchronize_chain_state_from`.

## Proposal

Process the timeout certificate.

## Test Plan

A client test was extended to verify that client will process the timeout certificate requested by another client.

## Release Plan

- These changes should be released in a new SDK, and
- ported to `main`.

## Links

- #3839

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -619,6 +619,14 @@ pub enum ChainClientError {
 
     #[error(transparent)]
     BcsError(#[from] bcs::Error),
+
+    #[error("Unexpected quorum: got {hash}, {round}, expected {expected_hash}, {expected_round}")]
+    UnexpectedQuorum {
+        hash: CryptoHash,
+        round: Round,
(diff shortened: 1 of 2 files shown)


deuszx (approved):
We could extract the synchronization logic into a separate struct for easier testing.

afck:
> We could extract the synchronization logic into a separate struct for easier testing.

How? It uses the local node, and some of the chain client's fields.

deuszx:
> > We could extract the synchronization logic into a separate struct for easier testing.
> 
> How? It uses the local node, and some of the chain client's fields.

IDK how yet but there's got to be a way (injecting externalities as dependencies; passing as args to function). This is a quite complicated logic that is contained in a couple of methods and require quite eloborate test setup.

PR #3842, opened 2025-04-28
Update actions/checkout to v4 

PixelPil0t1:
This PR updates the actions/checkout GitHub Action from v3 to v4

Reference: [actions/checkout v4 release](https://github.com/actions/checkout/releases)


PixelPil0t1:
> Is that all of them? Would be good to do this in a single PR.

Sorry about that! Yes, this is all of them.

PR #3847, opened 2025-04-29
Provide EVM support for some linera features

MathieuDutSik:
## Motivation

The Linera features need to be provided for EVM. Some of this is done here.

## Proposal

The following is provided:
* validation_round
* chain_ownership
* chain_id
* application_creator_chain_id
* read_data_blob
* assert_data_blob_exists

The precompile is extended for that purpose and the handling of the various cases is reorganized for clarity.

## Test Plan

A test for the features has been provided in the CI.

## Release Plan

The goal is to provide all this for TestNet 3.

## Links

None.


linera-execution/src/evm/revm.rs
```diff
@@ -287,6 +302,50 @@ enum PrecompileTag {
...
+/// Key prefixes used to transmit precompiles.
+#[derive(Debug)]
+enum PrecompileTag {
+    Base(BasePrecompileTag),
+    Contract(ContractPrecompileTag),
+    Service(ServicePrecompileTag),
+}
+
+impl PrecompileTag {
+    fn from_u8(tag: u8) -> Result<Self, String> {
```
ma2bd, line 321: Is there a way we could avoid this boilerplate?

ma2bd: (For instance using `bincode` or `bcs` serialization)

MathieuDutSik: Yes, there are ways, but they are not necessarily as simple as 10 lines.

ma2bd: We consistently avoid boilerplate in the rest of the codebase

MathieuDutSik: Ok, changed.

linera-execution/src/evm/revm.rs
```diff
@@ -272,13 +273,28 @@ fn address_to_user_application_id(address: Address) -> ApplicationId {
...
+    ChainId,
+    /// Key prefix for `application_creator_chain_id`
+    ApplicationCreatorChainId,
+    /// Key prefix for `chain_ownership`
+    ChainOwnership,
+    /// Key prefix for `read_data_blob`
+    ReadDataBlob,
+    /// Key prefix for `assert_data_blob_exists`
+    AssertDataBlobExists,
+}
```
deuszx, line 289: This is still missing lots of methods. What's the plan for adding it?

linera-execution/solidity/linera.sol
```diff
@@ -3,13 +3,53 @@
...
+// (1,3): message_id
+// (1,4): message_is_bouncing
+// (2,0): try_query_application
 library Linera {
+    function bcs_deserialize_offset_bytes32(uint256 pos, bytes memory src) internal pure returns (uint256, bytes32) {
+        bytes32 dest;
+        assembly {
+            dest := mload(add(add(src, 0x20), pos))
+        }
+        uint256 new_pos = pos + 32;
```
deuszx, line 23: Both `0x20` and `32` are magic numbers. Please add at least comments explaining what they are for.

linera-execution/solidity/linera.sol
```diff
@@ -18,16 +58,297 @@ library Linera {
...
+        input1[1] = bytes1(uint8(4));
+        bytes memory input2 = abi.encodePacked(input1, hash);
+        (bool success, bytes memory output1) = precompile.call(input2);
+        require(success);
+        assert(output1.length == 0);
+    }
+
+    struct OptionU32 {
+        bool has_value;
+        uint32 value;
```
deuszx, line 351: Is the code above generated or hand-written?

linera-execution/src/evm/revm.rs
MathieuDutSik: It is generally preferred to have smaller PRs.
Some functions are not going to be inserted like `fn application_parameters` because it contains the constructor parameters.
But yes, other functions could to be inserted:
* Maybe `read_system_timestamp` since it in micros vs second for EVM.
* The `read_chain_balance`.
* The `read_owner_balance` is normally available directly in EVM.
* The `read_owner_balances` and `read_balances_owner` could be done

The above is contingent on having correct understanding of balances in EVM.

Then the `http_request` requires more work and thinking for a good test case.

The events require their own work as well.

The `create_application` requires more work in relation to `create / create2`.

linera-execution/solidity/linera.sol
MathieuDutSik: Partially computer generated, partially hand written.

linera-execution/solidity/linera.sol
MathieuDutSik: In that case 0x20 = 32. But yes, I can document that.

linera-execution/solidity/linera.sol
deuszx: Could we split them then? Manually-written from the generated ones and tag the generated file with an appropriate notice in the header about:
- how it was generated (command)
- when it was generated (timestamp)
- that it MUST NOT be modified manually (only via fixing the source generator)
?

Otherwise it's impossible to tell what should be generated and what has to be manually implemented.

For example, I don't know if I should review the functions that (de)serialize like bcs does b/c those _should_ be generated.

MathieuDutSik: Many changes were made to the generated code:
* I changed some types which were cumbersome and replaced them with "bytes32" or similar. And I want to change appropriately the `serde-reflection`.
* I also added checks for the nature of the object for boolean and also this will go into `serde-reflection`.
* I removed the functions that are unused.
* I reformatted the generated code with "forge fmt"
* Some types like `MessageIsBouncing` are a simpler version than the generated code.
* I avoided having `CryptoHash` and instead put directly "bytes32"
* Some name like "function chain_id" causes a namespace collision with other "chain_id" in the code that had to be renamed.

So, I am not sure how to document all that.

deuszx: I'd say that all of of the changes that you made to the generated code should be made to the generator itself. We should not modify the generated code. 

If you feel like there needs to be some "translation layer"/adapter between the generated code and this codebase then it should be added _in addition_ to the generated one AND in a separate file, with proper comments explaining why the generated output is not sufficient.

MathieuDutSik: Well, we need to have some contextual changes. How is the generated code supposed to know that we have a `function chain_id()`, and so that there is a namespace collision?

How is the `serde-generate` supposed to know which functions are used?

Otherwise, yes I agree that some of the changes will find its way into `serde-generate`. But that is not immediate.

deuszx: I didn't mean that the Linera API should end up in the generated code but all the mappings between Solidity<>Linera serialization could and should.

MathieuDutSik: In the end, the computer-generated types are not going to look great. Where is the harm in writing specific types?

linera-execution/solidity/linera.sol
```diff
@@ -113,35 +401,43 @@ library Linera {
...
+        bytes memory input1 = new bytes(2);
+        input1[0] = bytes1(uint8(1));
+        input1[1] = bytes1(uint8(1));
+        (bool success, bytes memory output1) = precompile.call(input1);
         require(success);
-        return output;
+        uint32 val = bcs_deserialize_uint32(output1);
+        bool has_value = true;
+        uint32 value = 0;
+        if (val == 0) {
```
deuszx, line 420: Aren't rounds 0-indexed? i.e. first round has value=0.

linera-execution/solidity/linera.sol
```diff
@@ -150,11 +446,20 @@ library Linera {
 
     function message_is_bouncing() internal returns (MessageIsBouncing) {
         address precompile = address(0x0b);
-        bytes memory input1 = new bytes(1);
-        input1[0] = bytes1(uint8(4));
+        bytes memory input1 = new bytes(2);
+        input1[0] = bytes1(uint8(1));
+        input1[1] = bytes1(uint8(4));
```
deuszx, line 451: Is this code generated or hand-written? Where did you get the bytes from? Also, it seems like previous version was incorrect? Or why did it change? I don't see any test being updated that would help understanding.

MathieuDutSik: As far as I know, the previous version was correct.
The matching changes are in `revm.rs`. The `PrecompileTag` has been reorganized in order to handle those various cases separately and more cleanly than before.

In one way or another we have two codes: The rust code and the solidity code. It is not going to be super nice. So, we need a translation layer and that translation layer is `linera.sol`.

linera-execution/solidity/Linera.sol
```diff
@@ -0,0 +1,130 @@
...
+// (0,1): application_creator_chain_id
+// (0,2): chain_ownership
+// (0,3): read data blob
+// (0,4): assert data blob exists
+// (1,0): try_call_application
+// (1,1): validation round
+// (1,2): send_message
+// (1,3): message_id
+// (1,4): message_is_bouncing
+// (2,0): try_query_application
```
deuszx, line 21: What's the reasoning for choosing this particular encoding? each element of the tuple is `uint8` so holds numbers up to 255. Wouldn't it be easier to simply keep the indexes monotonically increase?

MathieuDutSik: It has to match the bcs encoding of the type `PrecompileTag`.

MathieuDutSik: This organization of the `PrecompileTag` is the one before this PR.

The problem is that some runtime operations are available in both `Contract` and `Service`, some are available only in `Contract` and some only in `Service`.

With the proposed reorganization, things become much clearer and the code is better organized:
```rust
enum PrecompileTag {
    Base(BasePrecompileTag),
    Contract(ContractPrecompileTag),
    Service(ServicePrecompileTag),
}
```

But since we do not want to have boilerplate code, this implies having 2 bytes instead of just 1.

deuszx: I see. I still think it'd be good to have some better safety precautions here.

Maybe for the time being add a longer comment here that these indices have to match the serde encoding of Precompile tag. Also, maybe a comment on `PrecompileTag` with a warning that things should not be reorganised b/c they can break the integration.

deuszx (approved):
Approving but please update the comment on the indices for the Linera FFI.

PR #3848, opened 2025-04-29
Automatically rotate the faucet chain once it has a maximum length.

### LINKED ISSUE #3846: Rotate faucet chain.
To improve performance for `wallet init --with-new-chain`, the faucet should, whenever its chain gets too long, create a new one and move all its tokens there.

afck:
## Motivation

If the faucet chain gets very long, requesting a new chain becomes very slow for new wallets.

## Proposal

Make the faucet switch chains every 100 blocks.

## Test Plan

I extended `test_end_to_end_faucet_with_long_chains`.

## Release Plan

    - be deployed as the faucet.

## Links

- Closes #3846.

### DIFF
--- a/linera-faucet/server/src/lib.rs
+++ b/linera-faucet/server/src/lib.rs
@@ -163,6 +163,7 @@ where
                 .open_chain(ownership, ApplicationPermissions::default(), balance)
                 .await?
                 .try_unwrap()?;
+            // TODO(#1795): Move the remaining tokens to the new chain.
             client.close_chain().await?.try_unwrap()?;
             let chain_id = ChainId::child(message_id);
             info!("Switching to a new faucet chain {chain_id:8}; remaining balance: {balance}");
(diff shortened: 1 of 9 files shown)


linera-faucet/server/src/lib.rs
```diff
@@ -150,16 +153,32 @@ where
...
+            // TODO(#1795): Move the remaining tokens to the new chain.
+            client.close_chain().await?.try_unwrap()?;
+            let chain_id = ChainId::child(message_id);
+            info!("Switching to a new faucet chain {chain_id:8}; remaining balance: {balance}");
+            self.context
+                .lock()
+                .await
+                .update_wallet_for_new_chain(
+                    chain_id,
+                    Some(key_pair),
```
christos-h, line 175: We use the same keypair on child chains?

christos-h (approved):
LGTM :) Nice!

linera-faucet/server/src/lib.rs
afck: It was easiest to do it that way, and maybe it's also better: That way you don't have to keep backing up new key pairs.

deuszx: This will change in a moment with the new `Signer` abstraction anyway.

Issue #3841, opened 2025-04-28
Message grants should be able to pay all fees.


afck:
If all incoming messages are covered by grants, the chain owner should not have any fees left to pay.

* [ ] Remove the block fee.
* [ ] Disallow empty blocks.
* [ ] Charge fees for published blobs as part of the publishing transaction.

afck:
Related: #3605

PR #3850, opened 2025-04-29
Remove block fee, disallow empty blocks.

afck:
## Motivation

If all incoming messages are covered by grants, the chain owner should not have any fees left to pay.

## Proposal

Remove the base block fee. Disallow empty blocks.

## Test Plan

TBD

## Release Plan

## Links

- Part of https://github.com/linera-io/linera-protocol/issues/3841.


.github/workflows/rust.yml
```diff
@@ -63,8 +63,6 @@ jobs:
     - name: Create two epochs and run the faucet
       run: |
         cargo build --bin linera
-        cargo run --bin linera -- resource-control-policy --block 0.0000001
-        cargo run --bin linera -- resource-control-policy --block 0.000000
```
afck, line 67: Right, we should have documented this: This was to test a regression where some tests failed if we aren't in epoch 0. Is it necessary to remove this? (If it is, that's probably a worrying sign.)

linera-execution/tests/fee_consumption.rs
```diff
@@ -206,37 +206,36 @@ async fn test_fee_consumption(
...
-        byte_written: Amount::from_tokens(13),
-        byte_stored: Amount::from_tokens(17),
-        operation: Amount::from_tokens(19),
-        operation_byte: Amount::from_tokens(23),
-        message: Amount::from_tokens(29),
-        message_byte: Amount::from_tokens(31),
-        service_as_oracle_query: Amount::from_millis(37),
-        http_request: Amount::from_tokens(41),
+        evm_fuel_unit: Amount::from_tokens(2),
+        read_operation: Amount::from_tokens(3),
```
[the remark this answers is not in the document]
deuszx: I didn't know there's a pattern here. I think this is a result of my merge with `main`?

linera-execution/src/resources.rs
```diff
@@ -108,9 +108,11 @@ where
     /// and `other` to `self`.
     pub fn merge_balance(&mut self, initial: Amount, other: Amount) -> Result<(), ExecutionError> {
         if other <= initial {
+            let sub_amount = initial.try_sub(other).expect("other <= initial");
             self.account
                 .try_sub_assign(initial.try_sub(other).expect("other <= initial"))
```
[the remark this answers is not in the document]
deuszx: Done.

.github/workflows/rust.yml
afck: Yes, sorry about that! It was https://github.com/linera-io/linera-protocol/pull/2835.

We could add them back and set another fee category to non-zero and then zero. As a comment, maybe something like:
```
# This is a regression test for #2835, and in general to make sure the faucet
# works well even if it was started in a later epoch.
```

Alternatively: Maybe it _is_ fine to remove it. Mainnet won't have a faucet, and the way we handle new epochs and new chains has fundamentally changed since anyway.

afck (commented):
No blockers from my side! But I can't approve because I wrote parts of it. Needs another reviewer.

.github/workflows/rust.yml
deuszx: Ok, I will do that.

deuszx (approved):

PR #3851, opened 2025-04-29
Make some operations exempt from app permissions.

afck:
## Motivation

Chains with restricted block operations can't process e.g. new events because system operations are not allowed.

## Proposal

Allow `Process*Epoch` and `UpdateStreams`.
Enforce that `UpdateStreams` actually has new events.

## Test Plan

A test was extended.

## Release Plan

    - be released in a new SDK,

## Links

- Fixes https://github.com/linera-io/linera-protocol/issues/3849.

### DIFF
--- a/linera-core/src/unit_tests/client_tests.rs
+++ b/linera-core/src/unit_tests/client_tests.rs
@@ -9,9 +9,9 @@ mod wasm;
 use assert_matches::assert_matches;
 use futures::StreamExt;
 use linera_base::{
-    crypto::AccountSecretKey,
+    crypto::{AccountSecretKey, CryptoHash},
     data_types::*,
-    identifiers::{Account, AccountOwner, ChainId, MessageId},
+    identifiers::{Account, AccountOwner, ApplicationId, ChainId, MessageId},
     ownership::{ChainOwnership, TimeoutConfig},
 };
 use linera_chain::{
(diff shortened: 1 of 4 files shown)


linera-core/src/unit_tests/client_tests.rs
```diff
@@ -1236,6 +1236,20 @@ where
...
+    )))
+    .await?;
+
+    let committee = Committee::new(validators, ResourceControlPolicy::default());
+    admin.stage_new_committee(committee).await.unwrap();
+    assert_eq!(admin.epoch().await.unwrap(), Epoch::from(3));
+
+    user.synchronize_from_validators().await?;
+    user.process_inbox().await?;
+    assert_eq!(user.epoch().await.unwrap(), Epoch::from(3));
```
deuszx, line 1251: Can you add a comment to this new part with the explanation?

deuszx (approved):
No blockers but let's just add explanatory comment.

linera-core/src/unit_tests/client_tests.rs
afck: Done in https://github.com/linera-io/linera-protocol/pull/3851/commits/b969c8a44dbba791f55aec0b4edfccd10d0fc908.

Issue #1948, opened 2024-04-24
Remove `Context` trait


Twey:
The `linera_views::common::Context` trait is implicitly just a triple holding:
- an instance of `KeyValueStore + Clone`
- a base key
- an ‘extra’ field

Each of these can be removed, meaning the trait as a whole can be factored right out.

# `KeyValueStore`
We can remove this by replacing any `Context` type parameter with a ‘storage’ parameter that implements `KeyValueStore`.

# Base key
Each `View` already implicitly owns a `base_key`.  We currently keep it in an associated `Context`, which we modify for each `View`, but it could easily be inlined into the `View` trait.

# ‘Extra’ data
This is used in only one place, and exists because we currently don't have the capability to add non-`View` fields to structs deriving `View`.  If we add a ‘skip’ attribute to the `View` derive macro, the need for this part of `Context` goes away.

PR #3854, opened 2025-04-29
`linera-views-derive`: minor cleanups

Twey:
## Motivation

I wanted to [add a `view(skip)` directive to `linera-views-derive`](https://github.com/linera-io/linera-protocol/pull/3845) in order to facilitate the [removal of the `Extra` type from `linera_views::Context`](https://github.com/linera-io/linera-protocol/issues/1948), but it turns out to be more effort than it's worth to get skipped fields into the struct when loading.

However, while I was there I made a couple of tweaks to `linera-views-derive` that I think are improvements.

## Proposal

- add new constraints to `*View` implementations where they're required, rather than relying on the consumer to add them: recursive constraints on the field types, and `Self: Send + Sync` since we rely on `async_trait`
  - this should make the library more robust in the face of different uses, and also provide slightly better errors when misused
- use `deluxe` for attribute parsing instead of doing it by hand: this is less of a win than it was originally since without `view(skip)` we only have one attribute to parse, but at least makes it easier to add more in the future
- parse types for the `context` attribute rather than strings
- use the [`quote` repetition syntax](https://docs.rs/quote/latest/quote/macro.quote.html#interpolation) to save some code

## Test Plan

CI.

## Release Plan

## Links

### DIFF
--- a/linera-views-derive/src/lib.rs
+++ b/linera-views-derive/src/lib.rs
@@ -6,10 +6,13 @@
 use proc_macro::TokenStream;
 use proc_macro2::{Span, TokenStream as TokenStream2};
 use quote::{format_ident, quote};
-use syn::{
-    parse_macro_input, parse_quote, punctuated::Punctuated, ItemStruct, Token, Type, TypePath,
-    WhereClause,
-};
+use syn::{parse_macro_input, parse_quote, ItemStruct, Type, TypePath};
+
+#[derive(Debug, deluxe::ParseAttributes)]
+#[deluxe(attributes(view))]
+struct StructAttrs {
+    context: Option<syn::Type>,
+}
 
 fn get_seq_parameter(generics: syn::Generics) -> Vec<syn::Ident> {
     let mut generic_vect = Vec::new();
@@ -21,20 +24,14 @@ fn get_seq_parameter(generics: syn::Generics) -> Vec<syn::Ident> {
     generic_vect
 }
 
-#[derive(Debug, deluxe::ExtractAttributes)]
-#[deluxe(attributes(view))]
-struct StructAttrs {
-    context: Option<syn::Type>,
-}
-
 fn context_and_constraints(
-    item: &mut syn::ItemStruct,
+    item: &syn::ItemStruct,
     template_vect: &[syn::Ident],
-) -> (Type, WhereClause) {
-    let attrs: StructAttrs = deluxe::extract_attributes(item).unwrap();
+) -> (Type, Vec<syn::WherePredicate>) {
+    let attrs: StructAttrs = deluxe::parse_attributes(item).unwrap();
 
     if let Some(context) = attrs.context {
-        (context, empty_where_clause())
+        (context, vec![])
     } else {
         let context = Type::Path(TypePath {
             qself: None,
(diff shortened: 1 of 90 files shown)


linera-views-derive/src/lib.rs
```diff
@@ -21,61 +24,29 @@ fn get_seq_parameter(generics: syn::Generics) -> Vec<syn::Ident> {
...
-
 fn context_and_constraints(
-    attributes: &[Attribute],
+    item: &syn::ItemStruct,
     template_vect: &[syn::Ident],
-) -> (Type, WhereClause) {
-    let context;
-    let constraints;
+) -> (Type, Vec<syn::WherePredicate>) {
+    let attrs: StructAttrs = deluxe::parse_attributes(item).unwrap();
```
MathieuDutSik, line 31: We do not like to put types like that.
Here we prefer
```rust
    let attrs = deluxe::parse_attributes::<_, StructAttrs>(item).unwrap();
```

linera-views-derive/src/lib.rs
```diff
@@ -223,22 +198,18 @@ fn generate_view_code(input: ItemStruct, root: bool) -> TokenStream2 {
 }
 
 fn generate_save_delete_view_code(input: ItemStruct) -> TokenStream2 {
```
MathieuDutSik, line 200: The term `generate_save_delete_view_code` is inherited from the first version where we had a `delete` functionality. This is no longer the case.
So, maybe rename the function to `generate_rootview_code`.

linera-views-derive/src/lib.rs
```diff
@@ -90,18 +61,14 @@ fn get_extended_entry(e: Type) -> TokenStream2 {
...
-    let mut where_clause = maybe_where_clause
-        .cloned()
-        .unwrap_or_else(empty_where_clause);
-    where_clause
-        .predicates
-        .extend(context_constraints.predicates);
+    let input_constraints = maybe_where_clause
+        .map(|w| w.predicates.iter().cloned())
+        .into_iter()
+        .flatten();
```
MathieuDutSik, line 71: That code section is duplicated in `generate_view_code`, `generate_hash_view_code` and in general all the function generating code.
So, I think it would make sense to add it to the `context_and_constraints`

MathieuDutSik (approved):
What I like about the PR:
* The better handling of the constraints.
* Some simplification of the code.
* The replacement of ViewStorageContext by something that is no longer a string.

About the `Send + Sync`, I am not sure but so be it.

linera-views-derive/src/lib.rs
Twey: I'm not sure I like that (at least in this case) because it makes the use of the type variable much less clear. The fact we passed `StructAttrs` to `parse_attributes` doesn't tell us what it's used for, whereas by typing the variable we know that it's the type of the resulting variable.

Twey:
Thanks for the feedback! I'm not too concerned about the snapshot size, since they're auto-generated anyway and we mostly care about their diffs.

(Great idea to snapshot-test this macro crate though — I find them super useful when changing the derive macros!)

linera-views-derive/src/lib.rs
Twey: https://github.com/linera-io/linera-protocol/pull/3854/commits/4b4ef12d18e22db810555a52ec405ecbf95fde85

linera-views-derive/src/lib.rs
Twey: https://github.com/linera-io/linera-protocol/pull/3854/commits/3419ca31d54572b0fe8a956f0a624a8f9bdd2e85

linera-views-derive/src/lib.rs
MathieuDutSik: As you want.
As mentioned in the `CONTRIBUTE.md`, we put type annotations only when the compiler requires it, and if so, we prefer to do it in the right-hand side.

PR #3861, opened 2025-04-29
`linera_views::Context`: extract `KeyValueStore` functions

Twey:
## Motivation
The `Context` trait currently duplicates the `ReadableKeyValue` and `WritableKeyValue` traits.

## Proposal
Instead, provide direct access to the store so there's no need to pass through these calls.  This shrinks the `Context` trait and its implementers quite considerably in preparation for https://github.com/linera-io/linera-protocol/issues/1948.

While we're at it, factor `BaseKey` out into a separate type for the same reason.

Additionally, I noticed we've maintained the `Local`/`Send` split for our traits, a distinction that we chose to abandon in the codebase in favour of conditional compilation, so I merged those.  This led to some knock-on effects:

- we get to remove `async-trait` in a lot of places, replacing its trait transformation functionality with `trait-variant` where necessary — this simplifies the implementation of the traits, since implementers don't need to re-iterate the `Send`ness of the trait
- since we can now no longer implement things for `*Store` without implementing them for `Local*Store`, we need to be more careful about the propagation of `Send`/`Sync` traits.  Notably this required making GraphQL support conditional, which could even be extended to a feature flag in the future (just by modifying `Cargo.toml` and `build.rs`, since I made use of the `with_*` alias pattern).

In addition it seems that `cargo machete` detected some unused dependencies in our workspace and decided to remove them.

## Test Plan

CI.

## Release Plan

## Links


MathieuDutSik (commented):
The PR actually combines more than what it claims:
* The splitting of the `Context` into a `BaseKey` and a `Store`. It does reflect the functionality. The view code becomes more complicated. But breaking down a big class into smaller ones is generally a welcomed thing.
* The removal of some `async_trait` for some traits and the replacement by something else.
* The restructuration of the Store traits.
* Removal of `secp256k1` from the Cargo.toml.

I would like all the points of the PR to be shortly discussed in the description.

Also, it would be good to discuss the further plan for the `Context` to be discussed in the description if there are any.

MathieuDutSik:
> You're right, getting tests passing resulted in a few more changes than originally intended 😅 I'll update the description. (The `Cargo.toml` changes were I think a rebase artefact — oops!)

To be clear, removing unused crates is an unconditional positive. I just would like it to be documented.

afck (commented):
Looks great! But why are so many snap files getting deleted? Were these unused all along or did some tests get disabled due to conditional compilation?

Twey:
> To be clear, removing unused crates is an unconditional positive. I just would like it to be documented.

Done!  I'm not entirely sure how that change got in there, be it through rebase or `cargo machete`, but I guess since CI is happy they weren't necessary.

> Looks great! But why are so many snap files getting deleted? Were these unused all along or did some tests get disabled due to conditional compilation?

They were unused all along.   I realized by accident when in order to reset the snapshots I deleted them and then reran the tests.  The only feature flag that affects these snapshots is `metrics`.

linera-views/src/store.rs
```diff
@@ -78,8 +78,8 @@ pub trait WithError {
 }
 
 /// Low-level, asynchronous read key-value operations. Useful for storage APIs not based on views.
-#[trait_variant::make(ReadableKeyValueStore: Send)]
-pub trait LocalReadableKeyValueStore: WithError {
+#[cfg_attr(not(web), trait_variant::make(Send))]
```
ma2bd, line 81: Nice!

ma2bd (approved):

PR #3862, opened 2025-04-29
Improve download certificate performance.

christos-h:
## Release Plan

    - be released in a new SDK,
    - be released in a new SDK,
      SDK soon.

## Links

### DIFF
--- a/linera-client/src/client_context.rs
+++ b/linera-client/src/client_context.rs
@@ -412,30 +412,23 @@ where
         // Take the latest committee we know of.
         let admin_chain_id = self.wallet.genesis_admin_chain();
         let query = ChainInfoQuery::new(admin_chain_id).with_committees();
-        let nodes: Vec<_> = if let Some(validators) = validators {
-            node_provider
-                .make_nodes_from_list(validators)?
-                .map(|(public_key, node)| RemoteNode { public_key, node })
-                .collect()
-        } else {
-            let info = self
-                .client
-                .local_node()
-                .handle_chain_info_query(query)
-                .await?;
-            let committee = info
-                .latest_committee()
-                .ok_or(error::Inner::ChainInfoResponseMissingCommittee)?;
-            node_provider
-                .make_nodes(committee)?
-                .map(|(public_key, node)| RemoteNode { public_key, node })
-                .collect()
-        };
+        let info = self
+            .client
+            .local_node()
+            .handle_chain_info_query(query)
+            .await?;
+        let committee = info
+            .latest_committee()
+            .ok_or(error::Inner::ChainInfoResponseMissingCommittee)?;
+        let nodes: Vec<_> = node_provider
+            .make_nodes(committee)?
+            .map(|(public_key, node)| RemoteNode { public_key, node })
+            .collect();
 
         // Download the parent chain.
         let target_height = message_id.height.try_add_one()?;
         self.client
-            .download_certificates(&nodes, message_id.chain_id, target_height)
+            .download_certificates(&nodes, message_id.chain_id, target_height, committee.clone())
             .await
             .context("downloading parent chain")?;
 
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -282,6 +282,7 @@ impl<P, S: Storage + Clone> Client<P, S> {
     }
 }
 
+
 impl<P, S> Client<P, S>
 where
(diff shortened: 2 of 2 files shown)


linera-client/src/client_context.rs
```diff
@@ -412,30 +412,23 @@ where
         // Take the latest committee we know of.
         let admin_chain_id = self.wallet.genesis_admin_chain();
         let query = ChainInfoQuery::new(admin_chain_id).with_committees();
-        let nodes: Vec<_> = if let Some(validators) = validators {
```
afck, line 415: Isn't this needed for the case where we request a chain from the faucet in a network where the genesis validators have already disappeared?

We should probably sync the admin chain first, then take the latest committee, and then use that one to download the certificates for the assigned chain.

linera-core/src/client/mod.rs
```diff
@@ -282,6 +282,7 @@ impl<P, S: Storage + Clone> Client<P, S> {
     }
 }
 
+
```
afck, line 285:
```suggestion
```

linera-core/src/client/mod.rs
```diff
@@ -294,32 +295,51 @@ where
...
-                target_next_block_height,
-            )
-            .await?;
-        }
+        // Request certificates in parallel from 1/3 + 1 of total validators in terms of votes.
+        let mut validators_and_votes = validators.iter()
+            .map(|validator| (validator, committee.validators.get(&validator.public_key).expect("validator should always be in committee").votes))
+            .collect::<Vec<_>>();
+        validators_and_votes.shuffle(&mut rand::thread_rng());
+        let sufficient_threshold = (committee.total_votes() / 3) + 1;
```
afck, line 305: Let's use `committee.validity_threshold` here.

(It's actually not 1/3 + 1, just 1/3 rounded up.)

linera-core/src/client/mod.rs
```diff
@@ -294,32 +295,51 @@ where
...
+        let sufficient_validator_subset: Vec<_> = validators_and_votes
+            .into_iter()
+            .scan(0u64, move |sum, (validator, votes)| {
+                let prev = *sum;
+                *sum += votes;
+                (prev < sufficient_threshold).then(|| validator)
+            })
```
afck, line 312: Or `take_while`? Then we don't keep iterating once we've reached the threshold.

linera-core/src/client/mod.rs
```diff
@@ -294,32 +295,51 @@ where
...
+            .collect();
         let info = self.local_node.chain_info(chain_id).await?;
         if target_next_block_height <= info.next_block_height {
-            Ok(info)
-        } else {
-            Err(ChainClientError::CannotDownloadCertificates {
-                chain_id,
-                target_next_block_height,
-            })
+            return Ok(info);
```
afck, line 316: Then we don't need the above computation, so I'd move this to the beginning of the function.

linera-core/src/client/mod.rs
```diff
@@ -294,32 +295,51 @@ where
...
+            .iter()
+            .map(|remote_node| {
+                let info = info.clone();
+                async move {
+                    self.try_download_certificates_from(
+                        remote_node,
+                        chain_id,
+                        info.next_block_height,
+                        target_next_block_height,
+                    ).await
```
bart-linera, line 328: If we need to download multiple certificates, this will request all of them from each node in parallel, if I understand this correctly - which seems to be a bit wasteful? It would probably be better to try and divide the block height range into some batches and download different batches from different validators... but of course that would also be much more complex, and we would still want to try other nodes if some batch didn't succeed with the node we chose at first :grimacing:

afck: Yes, we're currently erring on the side of not duplicating requests, but since some validators are very slow, this impacts client performance whenever one of those is randomly selected.

This approach prioritizes latency over bandwidth. But at least we can limit the duplication to ⅓ of the validators, rather than all of them.

Any better ideas are welcome, of course, but we'd like to improve the current testnet's performance as soon as possible, so if they take longer to implement they're maybe something for the third testnet instead.

linera-core/src/client/mod.rs
bart-linera: I think `scan` also stops iterating once it returns `None`, doesn't it?

afck: Ah, you're right!

bart-linera:
Closing as being superseded by #3913

PR #3863, opened 2025-04-30
Find current chain on faucet restart.

afck:
## Motivation

As of https://github.com/linera-io/linera-protocol/pull/3852, the faucet switches chains every 100 blocks. That means it cannot be restarted with the same command line, because the original chain may be closed (and broke).

## Proposal

On restart, while the current chain is closed, keep switching to the last chain that it opened until we find the most recent one.

Also, change the default maximum length to 500 blocks to reduce the overhead and the number of chains that end up in the faucet wallet.

## Test Plan

`test_end_to_end_faucet_with_long_chains` restarts the faucet now.

## Release Plan

- These changes should be ported to the `main` branch.
- The faucet should be restarted with this.

## Links

- See https://github.com/linera-io/linera-protocol/pull/3852.

### DIFF
--- a/CLI.md
+++ b/CLI.md
@@ -605,7 +605,7 @@ Run a GraphQL service that exposes a faucet where users can claim tokens. This g
 * `--limit-rate-until <LIMIT_RATE_UNTIL>` — The end timestamp: The faucet will rate-limit the token supply so it runs out of money no earlier than this
 * `--max-chain-length <MAX_CHAIN_LENGTH>` — The maximum number of blocks in the faucet chain, before a new one is created
 
-  Default value: `100`
+  Default value: `500`
 * `--listener-skip-process-inbox` — Do not create blocks automatically to receive incoming messages. Instead, wait for an explicit mutation `processInbox`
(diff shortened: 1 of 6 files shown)


CLI.md
```diff
@@ -605,7 +605,7 @@ Run a GraphQL service that exposes a faucet where users can claim tokens. This g
 * `--limit-rate-until <LIMIT_RATE_UNTIL>` — The end timestamp: The faucet will rate-limit the token supply so it runs out of money no earlier than this
 * `--max-chain-length <MAX_CHAIN_LENGTH>` — The maximum number of blocks in the faucet chain, before a new one is created
 
-  Default value: `100`
+  Default value: `1000`
```
ma2bd, line 608: Hard to say what the best value could be. Maybe `500` ?

ma2bd (approved):

CLI.md
afck: Done in https://github.com/linera-io/linera-protocol/pull/3863/commits/3c779b48b189df2839fabd7ae3c134153f70d211.

PR #3864, opened 2025-04-30
Don't fail if faucet chain was opened but old one not closed.

afck:
## Motivation

On the testnet, for some reason the faucet failed to close the old chain after migrating to a new one.

## Proposal

Make the faucet handle that case on startup: Also migrate if the old chain has no tokens, even if it is not closed.

## Test Plan

A new test for all these cases was added.

## Release Plan

- These changes should be ported to the main branch.
- We should try this on the faucet with the `a3edc33d8e951a1139333be8a4b56646b5598a8f51216e86592d881808972b07` chain.

## Links

- Extending the functionality from #3863.

### DIFF
--- a/linera-client/src/client_context.rs
+++ b/linera-client/src/client_context.rs
@@ -342,6 +342,11 @@ where
                 })
             })
             .await?;
+        } else {
+            self.mutate_wallet(|w| {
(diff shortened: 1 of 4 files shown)


afck:
This approach to faucet rotation was reverted.

PR #3868, opened 2025-05-01
Alloy version bump and removal of `alloy-signer-local`

MathieuDutSik:
## Motivation

The versions of alloy used in examples/Cargo.toml and Cargo.toml had diverged and needed to be reunified.
Also the `alloy-signer-local` was unused and needed to be removed.

## Proposal

Both changes are done.

The elimination of `alloy-signer-local` is more important than it looks like. Version 0.9.2 depends on `c-kzg`.
Unfortunately, version 22.0.1 of REVM depends on `c-kzg` as well and a native library can be present only
one time in the linking process.

In other words, one block for the use of the next version of Revm has been cleared. Two more obstacles remain:
* The resolution of the issue https://github.com/bluealloy/revm/issues/2425
* The upgrade of Revm from version 22.0.1 to the next one with this bug correction and the change of the PR https://github.com/bluealloy/revm/pull/2464

When this is done, we could finally use Linera ApplicationId (the 20 bytes long version of course) in EVM contracts.

## Test Plan

The CI.

## Release Plan

## Links

None.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -101,13 +101,13 @@ dependencies = [
 
 [[package]]
 name = "alloy-chains"
-version = "0.1.54"
+version = "0.1.69"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d38fdd69239714d7625cda1e3730773a3c1a8719d506370eb17bb0103b7c2e15"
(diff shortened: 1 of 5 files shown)


deuszx (approved):
LGTM

PR #2734, opened 2024-10-29
Introduce `dual-store`support

MathieuDutSik:
## Motivation

The `dual.rs` provides support for the dual stores. However, it was not implemented. This PR addresses that.

## Proposal

The following was done:
* The `fn create` should creates an error if the namespace already exists. This was added to the `admin_test` admin tests.
* This led to some correction to the `linera-storage-service`, `dual`, `memory` and `rocksdb` backends.
* The `DualStore` was introduced to the `linera-client/src/storage.rs`.
* The `DEFAULT_ROOT_KEY` and `DEFAULT_NAMESPACE` are introduced in the `linera-storage`and used.
* The wallet.rs is restructured. One consequence is that the initialization puts the chains in all the shards for RocksDb. 

## Test Plan

(A) The CI should provide the consistency check.

(B) The usage of the dualstore  in the tests adds more work to the CI, so it was limited to the counter and amm tests.

(C) The runtime is not optimistic. For 19 tests the runtime is 1415 seconds with dualstoreRocksDb_Scylla and 928 seconds for running with the ScyllaDb. Therefore, the dual store is not recommended to be used. But its insertion is I think valuable to the source code.

## Release Plan

The implementation of the system into TestNet / DevNet is quite premature. So, right now, it is 

## Links

- [reviewer checklist](https://github.com/linera-io/linera-protocol/blob/main/CONTRIBUTING.md#reviewer-checklist)

PR #3870, opened 2025-05-01
Simplification storage initialization of the dual-store database backend

ma2bd:
## Motivation

After #3787, it is no longer needed to initialize the shard of the dual-store databases separately.

## Proposal

Remove special code added in #2734

## Test Plan

CI


ndr-ds (approved):
Nice!

PR #3881, opened 2025-05-02
Improve code for validator checks

ma2bd:
## Motivation

Refactor the code for query_validator(s)

## Proposal

- Introduce some helper functions in `linera_client`

@Twey I made the new functions available to `web` , adding a dependency to `linera-version`. I figured version checks could be helpful in the web as well.

## Test Plan

CI + tested manually

### DIFF
--- a/linera-service/src/linera/main.rs
+++ b/linera-service/src/linera/main.rs
@@ -521,7 +521,11 @@ impl Runnable for Job {
                 if !faulty_validators.is_empty() {
                     println!("{:#?}", faulty_validators);
                 }
-                println!("{}/{} OK.", num_ok_validators, committee.validators().len());
+                info!(
+                    "{}/{} validators are OK.",
+                    num_ok_validators,
+                    committee.validators().len()
+                );
             }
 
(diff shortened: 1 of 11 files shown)


linera-service/src/linera/main.rs
```diff
@@ -517,59 +485,47 @@ impl Runnable for Job {
...
+                            .entry((name, address))
+                            .or_default()
+                            .push(e);
                     }
                 }
+                let num_ok_validators = committee.validators().len() - faulty_validators.len();
                 if !faulty_validators.is_empty() {
                     println!("{:#?}", faulty_validators);
                 }
-                println!("{}/{} OK.", num_ok_validators, committee.validators().len());
```
ma2bd, line 572: @christos-h Are we parsing this somewhere? Is it ok to turn it into `info!`?

ma2bd: context: We generally print only machine-parsable outputs on `stdout` (preferably in JSON or newline-separated JSON)

afck (approved):

PR #3882, opened 2025-05-03
correction client_options.rs

strmfos:
hi! 
fraction of time taken - fraction of the time taken
protected message - protected messages


linera-client/src/client_options.rs
```diff
@@ -185,7 +185,7 @@ pub struct ChainOwnershipConfig {
     )]
     timeout_increment: TimeDelta,
 
-    /// The age of an incoming tracked or protected message after which the validators start
+    /// The age of an incoming tracked or protected messages after which the validators start
```
afck, line 188: "an incoming tracked or protected message" _is_ singular?

afck:
I'm not even sure "fraction of time taken" is wrong. The other fix is wrong anyway, and CI is failing, too. Feel free to reopen if you disagree.

PR #3883, opened 2025-05-04
Add access to `ResourceControlPolicy`

MathieuDutSik:
## Motivation

The execution of EVM smart contracts requires access to the price of gas. It is a perfectly reasonable requirement but missing from the Runtime. This PR adds that.

## Proposal

The following is done:
* The `resource_control_policy` is added to the `ContractRuntime`. Maybe it could have been added instead to the `BaseRuntime`.
* The `ResourceControlPolicy` has to be moved to the `linera-base`, which makes sense since `Resources` is also there.
* The `resource_control_policy` has been added to the Wasm execution model.
* A Wasm test has been added for checking that the Wasm
* The newly available resource control policy has not been added yet as some other functionalities are still missing.

The movement of `ResourceControlPolicy` revealed that the oracle costs are not accounted for.

## Test Plan

The CI and the additional test.

## Release Plan

## Links

None


examples/test-runtime/README.md
```diff
@@ -0,0 +1,16 @@
+# Counter No GraphQL Example Application
+
```
ma2bd, line 2: What is this title? Is this intended?

examples/test-runtime/README.md
```diff
@@ -0,0 +1,16 @@
...
+This example application implements a simple counter contract, it is initialized with an
+unsigned integer that can be increased by the `increment` operation. In contrast with the
+counter application, it works without GraphQL 
+
+## How It Works
+
+It is a very basic Linera application, which is initialized by a `u64` which can be incremented
```
ma2bd, line 9: "very basic" => "simple".

linera-base/src/policy.rs
```diff
@@ -6,17 +6,25 @@
...
+    Clone,
+    Debug,
+    Serialize,
+    Deserialize,
+    InputObject,
+    WitLoad,
+    WitStore,
+    WitType,
+)]
 pub struct ResourceControlPolicy {
```
ma2bd, line 28: Do we need to expose the entire vector `ResourceControlPolicy` to the SDK?

ma2bd: What exactly are EVM contracts trying to achieve?

examples/test-runtime/README.md
MathieuDutSik: No, of course, corrected.

linera-base/src/policy.rs
MathieuDutSik: I think that smart contracts should be able to see the price of the gas when running. Smart contracts need to know the cost of their operation and decide based on that. So, I do not see the harm in providing too much information.

MathieuDutSik: The data that should be provided to the run is the gas price in the struct `TxEnv`.

Right now, that cost is not available in the `ResourceControlPolicy`. It should be after the PR https://github.com/linera-io/linera-protocol/pull/3715

MathieuDutSik:
The exposition of `ResourseControlPolicy` appears to be a wrong idea and so the PR is closed.

PR #3887, opened 2025-05-05
Test that follow-chain also works with child chains.

afck:
## Motivation

There were some reports of `sync` after `follow-chain` not working. `follow-chain` is only used directly after `forget-chain` (for the same chain) in the tests currently, so that the chain is still in the database.

## Proposal

Also test `follow_chain` in `test_end_to_end_assign_greatgrandchild_chain`.

## Test Plan

This seems fine on `main`. I will backport this to `testnet_babbage`.

## Release Plan

- Backport to `testnet_babbage` and fix, if necessary.

## Links

### DIFF
--- a/linera-service/tests/linera_net_tests.rs
+++ b/linera-service/tests/linera_net_tests.rs
@@ -3073,6 +3073,9 @@ async fn test_end_to_end_assign_greatgrandchild_chain(config: impl LineraNetConf
     let client2 = net.make_client().await;
     client2.wallet_init(&[], FaucetOption::None).await?;
 
+    let client3 = net.make_client().await;
+    client3.wallet_init(&[], FaucetOption::None).await?;
+
     let chain1 = *client1.load_wallet()?.chain_ids().first().unwrap();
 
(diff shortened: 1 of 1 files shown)


linera-service/tests/linera_net_tests.rs
```diff
@@ -3097,6 +3100,11 @@ async fn test_end_to_end_assign_greatgrandchild_chain(config: impl LineraNetConf
     client2.process_inbox(chain2).await?;
     assert!(client2.local_balance(account2).await? > Amount::ZERO);
 
+    // Verify that a third party can also follow the chain.
+    client3.follow_chain(chain2).await?;
```
deuszx, line 3104: Intuitively, _follow chain_ would start syncing immediately.

deuszx (approved):

linera-service/tests/linera_net_tests.rs
afck: Good point, let's discuss with the team. (Would be a separate PR, though.)

PR #3888, opened 2025-05-05
Fix syncing after `follow-chain` for child chains.

### LINKED ISSUE #3891: `follow-chain` does not work with child chains. (Testnet Babbage)
The backport of https://github.com/linera-io/linera-protocol/pull/3887 fails on `testnet_babbage`.

The main problem seems to be that we try to use "local" validators (i.e. the ones in the to-be-synchronized chain's state) in several places, and these are not defined for the uninitialized chain yet.

afck:
## Motivation

`sync` after `follow-chain` doesn't work.

## Proposal

Add a test (backport of https://github.com/linera-io/linera-protocol/pull/3887). Handle the case where a chain is uninitialized.

## Test Plan

The test was backported.

## Release Plan

## Links

- Closes #3891.
- Test for `main` (doesn't have the bug): https://github.com/linera-io/linera-protocol/pull/3887

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -1046,9 +1046,7 @@ where
 
         let mut info = self.synchronize_until(self.next_block_height()).await?;
 
-        if self.state().has_other_owners(&info.manager.ownership)
-            || info.next_block_height == BlockHeight::ZERO
-        {
+        if self.state().has_other_owners(&info.manager.ownership) {
             // For chains with any owner other than ourselves, we could be missing recent
             // certificates created by other owners. Further synchronize blocks from the network.
             // This is a best-effort that depends on network conditions.
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -922,9 +922,7 @@ where
             .await?
             .info;
         let epoch = info.epoch;
-        let committees = info
-            .requested_committees
-            .ok_or(LocalNodeError::InvalidChainInfoResponse)?;
+        let committees = info.requested_committees.unwrap_or_else(BTreeMap::new);
         Ok((epoch, committees))
     }
 
@@ -951,7 +949,7 @@ where
     /// Obtains the committee for the latest epoch.
     #[instrument(level = "trace")]
     pub async fn latest_committee(&self) -> Result<Committee, LocalNodeError> {
-        let (mut committees, epoch) = self.known_committees().await?;
+        let (epoch, mut committees) = self.known_committees().await?;
         committees
             .remove(&epoch)
             .ok_or(LocalNodeError::InactiveChain(self.chain_id))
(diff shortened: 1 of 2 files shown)


linera-core/src/client/mod.rs
```diff
@@ -922,9 +922,7 @@ where
             .await?
             .info;
         let epoch = info.epoch;
-        let committees = info
-            .requested_committees
-            .ok_or(LocalNodeError::InvalidChainInfoResponse)?;
+        let committees = info.requested_committees.unwrap_or_else(BTreeMap::new);
```
deuszx, line 925: I'm not sure I agree this is a good change. I understand it's necessary for the fix but:
- it's not possible for the protocol to execute correctly with empty committee
- if we requested it then it shouldn't be empty.

deuszx: On the call-site we only pattern match on the first element of the tuple and interpret the `None` as a chain not-active. Maybe we can return `Ok(None)` entirely.

linera-core/src/client/mod.rs
```diff
@@ -1750,9 +1750,12 @@ where
         #[cfg(with_metrics)]
         let _latency = metrics::SYNCHRONIZE_CHAIN_STATE_LATENCY.measure_latency();
 
-        let (epoch, mut committees) = self.epoch_and_committees(chain_id).await?;
+        let (epoch, mut committees) = match self.epoch_and_committees(chain_id).await? {
+            (Some(epoch), committees) => (epoch, committees),
+            (None, _) => self.known_committees().await?, // Chain is not active yet.
```
deuszx, line 1755: This is a bit confusing for a number of reasons:
- both `epoch_and_committees` and `known_committees` return the same types
- `known_commitees` calls `epoch_commitees(chain_id)` as well

IIUC, in the case of "chain not active yet" we want to fetch the known committee from the admin chain, is that right?

afck: Yes. TBH I'm not sure what the "right committee" is in all those cases where we just need _nodes_ to synchronize from, as opposed to actual validators to ask to sign something. But maybe that's beyond the scope of this PR, and something we should just fix on `main` instead?

linera-core/src/client/mod.rs
afck: This also triggers a bunch of other changes, and I was trying to keep this PR small.

I'll look into cleaning up all those committee getters a bit; if it's not too big I'll add it to this PR.

linera-core/src/client/mod.rs
deuszx: OK, agreed. BUT an empty committee can't be a right committee ever, right?

deuszx (dismissed):
As discussed - approving this PR in the interest of keeping the backport minimal and fixing it on the `main` instead.

linera-core/src/client/mod.rs
afck: That's true, yes.

afck:
Closing this for now, since the fix breaks another test. I created https://github.com/linera-io/linera-protocol/issues/3891.

PR #3892, opened 2025-05-05
Fix a bug where clients don't recover from conflicting proposals.

afck:
## Motivation

If two clients make conflicting proposals, it can happen that both of them have their own proposal in their local node, so when retrying, they will retry in the same round, because their own pending proposal isn't treated as a "conflict".

## Proposal

Use the next round, even if the proposal is our own.

## Test Plan

The added test reproduces the issue.

## Release Plan

    - be released in a new SDK.

## Links

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -2361,7 +2361,7 @@ impl<Env: Environment> ChainClient<Env> {
         // Using the round number during execution counts as an oracle.
         // Accessing the round number in single-leader rounds where we are not the leader
         // is not currently supported.
-        let round = match Self::round_for_new_proposal(&info, &identity, &proposed_block, true)? {
+        let round = match Self::round_for_new_proposal(&info, &identity, true)? {
             Either::Left(round) => round.multi_leader(),
             Either::Right(_) => None,
         };
@@ -2716,7 +2716,7 @@ impl<Env: Environment> ChainClient<Env> {
             // Use the round number assuming there are oracle responses.
             // Using the round number during execution counts as an oracle.
             let proposed_block = pending_proposal.block;
-            let round = match Self::round_for_new_proposal(&info, &owner, &proposed_block, true)? {
+            let round = match Self::round_for_new_proposal(&info, &owner, true)? {
                 Either::Left(round) => round.multi_leader(),
                 Either::Right(_) => None,
             };
@@ -2731,12 +2731,7 @@ impl<Env: Environment> ChainClient<Env> {
 
         let has_oracle_responses = block.has_oracle_responses();
         let (proposed_block, outcome) = block.into_proposal();
-        let round = match Self::round_for_new_proposal(
-            &info,
-            &owner,
-            &proposed_block,
-            has_oracle_responses,
-        )? {
+        let round = match Self::round_for_new_proposal(&info, &owner, has_oracle_responses)? {
             Either::Left(round) => round,
             Either::Right(timeout) => return Ok(ClientOutcome::WaitForTimeout(timeout)),
         };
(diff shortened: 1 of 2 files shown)


linera-core/src/unit_tests/client_tests.rs
```diff
@@ -1532,6 +1532,79 @@ where
...
+    for i in 0..4 {
+        let info = builder
+            .node(i)
+            .chain_info_with_manager_values(chain_id)
+            .await?;
+        assert_eq!(
+            AccountOwner::from(info.manager.requested_proposed.unwrap().public_key),
+            if i < 2 { owner1 } else { owner2 },
+        );
+    }
```
deuszx, line 1599: We can also check that _before_ resetting all validators to `Honest` again, right? IMHO that'd be cleaner:
- we split the network into two
- two clients communicate their proposals to two different non-major subsets
- the two validator subsets have different `requested_proposed` blocks set
- after fixing the network, it recovers.

linera-core/src/client/mod.rs
```diff
@@ -2855,16 +2850,17 @@ impl<Env: Environment> ChainClient<Env> {
...
         // If there is a conflicting proposal in the current round, we can only propose if the
         // next round can be started without a timeout, i.e. if we are in a multi-leader round.
         // Similarly, we cannot propose a block that uses oracles in the fast round.
-        let conflict = manager.requested_proposed.as_ref().is_some_and(|proposal| {
-            proposal.content.round == manager.current_round && proposal.content.block != *block
-        }) || (manager.current_round.is_fast() && has_oracle_responses);
+        let conflict = manager
+            .requested_proposed
+            .as_ref()
+            .is_some_and(|proposal| proposal.content.round == manager.current_round)
```
deuszx, line 2862: So now it's always a conflict if it's for the same round, even if it's equivalent to what we're trying to propose? In the new test though the clients are proposing different things (burn different amount of tokens).

linera-core/src/unit_tests/client_tests.rs
afck: You mean like this? https://github.com/linera-io/linera-protocol/pull/3892/commits/6017994e4b4bebd1774a0c2ca93ba580a611d197

deuszx: Yes. 👍

linera-core/src/client/mod.rs
deuszx: I see - so the _client_ sees a conflicting proposal from its node but it's his own proposal? So it's conflicting with itself?

afck: Kind of. The problem is that currently we can't easily distinguish in that part of the client code whether:
* Our own proposal is the only one the validators know of, or
* they know of an actual conflicting proposal.

In the second case, we have to try in a new round, in the first case it would be fine to retry in the same round. But it's always safe to retry in a new round anyway.

Making the client keep track of each validator's views would add a lot of complexity, but not only that: it also wouldn't really help! We can't wait for all validators' answers because some might be offline, so we can only reliably wait for 67%. But then we don't know if those 67% are honest, so we _never_ know for sure that our own proposal can still get a quorum. That's why I think it's always best to just try in the next round instead.

linera-core/src/unit_tests/client_tests.rs
afck: I had to comment it out for now anyway because the test utils currently don't allow _reliably_ creating this exact scenario. (https://github.com/linera-io/linera-protocol/issues/3894)

deuszx (approved):

linera-core/src/client/mod.rs
afck: In single-leader mode yes. In multi-leader mode it can be made immediately.

(Either way, the _common_ case if there are no serious network issues is that each proposal goes through on the first attempt. All of the consensus logic is just to handle the cases where it doesn't.)

bart-linera (approved):

PR #3893, opened 2025-05-06
Rotate the faucet chain automatically.

afck:
## Motivation

If the faucet chain gets very long, requesting a new chain becomes very slow for new wallets.

## Proposal

Make the faucet switch chains every 100 blocks.

The previous attempt was reverted: https://github.com/linera-io/linera-protocol/pull/3852
This new approach makes it less problematic to restart the faucet:

The faucet now only uses the main chain (which in practice will usually be one of our root chains in the testnet) to create temporary child chains. It only provides each child with enough funds to serve 100 user requests. (That number is a config option.) Once the child's funds are exhausted, it is closed and removed from the wallet, and a new child is created from the main chain.

In particular, the main chain is never drained of all funds at once, it never gets closed, and it is never used to directly serve user requests. All chains requested by users are _grandchildren_ of the main chain, because they are created by one of the temporary chains. Since these are never longer than 100 blocks, synchronization for new users should be much faster.

## Test Plan

`test_end_to_end_faucet_chain_limit`

## Release Plan

- These changes should be deployed on Testnet Babbage.
- They _could_ be ported to `main`.

## Links

- Closes https://github.com/linera-io/linera-protocol/issues/3846.

### DIFF
--- a/linera-base/src/data_types.rs
+++ b/linera-base/src/data_types.rs
@@ -371,7 +371,7 @@ macro_rules! impl_wrapped_number {
             }
 
             /// Saturating addition.
-            pub fn saturating_add(self, other: Self) -> Self {
+            pub const fn saturating_add(self, other: Self) -> Self {
                 let val = self.0.saturating_add(other.0);
                 Self(val)
             }
@@ -392,7 +392,7 @@ macro_rules! impl_wrapped_number {
             }
 
             /// Saturating subtraction.
-            pub fn saturating_sub(self, other: Self) -> Self {
+            pub const fn saturating_sub(self, other: Self) -> Self {
                 let val = self.0.saturating_sub(other.0);
                 Self(val)
             }
@@ -413,7 +413,7 @@ macro_rules! impl_wrapped_number {
             }
 
             /// Saturating in-place addition.
-            pub fn saturating_add_assign(&mut self, other: Self) {
+            pub const fn saturating_add_assign(&mut self, other: Self) {
                 self.0 = self.0.saturating_add(other.0);
             }
 
@@ -427,7 +427,7 @@ macro_rules! impl_wrapped_number {
             }
 
             /// Saturating multiplication.
-            pub fn saturating_mul(&self, other: $wrapped) -> Self {
+            pub const fn saturating_mul(&self, other: $wrapped) -> Self {
                 Self(self.0.saturating_mul(other))
             }
 
@@ -662,37 +662,37 @@ impl Amount {
     pub const ONE: Amount = Amount(10u128.pow(Amount::DECIMAL_PLACES as u32));
 
     /// Returns an `Amount` corresponding to that many tokens, or `Amount::MAX` if saturated.
-    pub fn from_tokens(tokens: u128) -> Amount {
+    pub const fn from_tokens(tokens: u128) -> Amount {
         Self::ONE.saturating_mul(tokens)
     }
 
     /// Returns an `Amount` corresponding to that many millitokens, or `Amount::MAX` if saturated.
-    pub fn from_millis(millitokens: u128) -> Amount {
+    pub const fn from_millis(millitokens: u128) -> Amount {
         Amount(10u128.pow(Amount::DECIMAL_PLACES as u32 - 3)).saturating_mul(millitokens)
     }
 
     /// Returns an `Amount` corresponding to that many microtokens, or `Amount::MAX` if saturated.
-    pub fn from_micros(microtokens: u128) -> Amount {
+    pub const fn from_micros(microtokens: u128) -> Amount {
         Amount(10u128.pow(Amount::DECIMAL_PLACES as u32 - 6)).saturating_mul(microtokens)
     }
 
     /// Returns an `Amount` corresponding to that many nanotokens, or `Amount::MAX` if saturated.
-    pub fn from_nanos(nanotokens: u128) -> Amount {
+    pub const fn from_nanos(nanotokens: u128) -> Amount {
         Amount(10u128.pow(Amount::DECIMAL_PLACES as u32 - 9)).saturating_mul(nanotokens)
     }
 
     /// Returns an `Amount` corresponding to that many attotokens.
-    pub fn from_attos(attotokens: u128) -> Amount {
+    pub const fn from_attos(attotokens: u128) -> Amount {
         Amount(attotokens)
     }
 
     /// Helper function to obtain the 64 most significant bits of the balance.
-    pub fn upper_half(self) -> u64 {
+    pub const fn upper_half(self) -> u64 {
         (self.0 >> 64) as u64
     }
 
     /// Helper function to obtain the 64 least significant bits of the balance.
-    pub fn lower_half(self) -> u64 {
+    pub const fn lower_half(self) -> u64 {
         self.0 as u64
     }
 
--- a/linera-faucet/server/src/lib.rs
+++ b/linera-faucet/server/src/lib.rs
@@ -25,6 +25,9 @@ use serde::Deserialize;
 use tower_http::cors::CorsLayer;
 use tracing::info;
 
+/// A rough estimate of the maximum fee for a block.
+const MAX_FEE: Amount = Amount::from_millis(100);
+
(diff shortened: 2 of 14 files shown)


linera-base/src/data_types.rs
```diff
@@ -662,37 +662,37 @@ impl Amount {
...
 
     /// Helper function to obtain the 64 most significant bits of the balance.
-    pub fn upper_half(self) -> u64 {
+    pub const fn upper_half(self) -> u64 {
         (self.0 >> 64) as u64
     }
 
     /// Helper function to obtain the 64 least significant bits of the balance.
-    pub fn lower_half(self) -> u64 {
+    pub const fn lower_half(self) -> u64 {
```
deuszx, line 695: These are not related to your PR.

afck: I had to make some of them `const` for `MAX_FEES`, so I thought I'd add it to all where it's possible.

linera-faucet/server/src/lib.rs
```diff
@@ -25,6 +25,9 @@ use serde::Deserialize;
 use tower_http::cors::CorsLayer;
 use tracing::info;
 
+/// A rough estimate of the maximum fee for a block.
+const MAX_FEE: Amount = Amount::from_millis(100);
```
deuszx, line 29: This won't be necessary after https://github.com/linera-io/linera-protocol/pull/3850 ?

afck: There will still be some fees left, because we are making an operation (rather than processing a message with a grant).

deuszx: Ah, OK. I thought it's the fee for block creation only

linera-faucet/server/src/lib.rs
```diff
@@ -47,9 +50,11 @@ pub struct QueryRoot<C> {
 
 /// The root GraphQL mutation type.
 pub struct MutationRoot<C> {
-    chain_id: ChainId,
+    main_chain_id: ChainId,
+    tmp_chain_id: Arc<Mutex<Option<ChainId>>>,
```
deuszx, line 54: IMHO these two fields should have appropriate comments. `tmp_` and `main_` do not say anything.

linera-faucet/server/src/lib.rs
```diff
@@ -118,17 +123,57 @@ where
...
+        let tmp_chain_id = match maybe_tmp_chain_id {
+            Some(tmp_chain_id) => tmp_chain_id,
+            None => {
+                let key_pair = main_client.key_pair().await?;
+                let balance = self
+                    .amount
+                    .try_add(MAX_FEE)?
+                    .try_mul(u128::from(self.max_claims_per_chain))?
+                    .try_add(MAX_FEE)? // One more block fee for closing the chain.
+                    .min(main_balance.try_sub(MAX_FEE)?);
```
deuszx, line 142: This could be extracted to a method (`chain_init_balance`) or even the field on the struct. We don't use `max_claims_per_chain` except for calculating this. It could be an argument to the constructor but the constructor could do the calculation on struct instantiation.

linera-faucet/server/src/lib.rs
```diff
@@ -145,20 +190,29 @@ where
...
-        self.context.lock().await.update_wallet(&client).await?;
-        let (message_id, certificate) = match result? {
-            ClientOutcome::Committed(result) => result,
-            ClientOutcome::WaitForTimeout(timeout) => {
-                return Err(Error::new(format!(
-                    "This faucet is using a multi-owner chain and is not the leader right now. \
-                    Try again at {}",
-                    timeout.timestamp,
-                )));
+        self.context.lock().await.update_wallet(&tmp_client).await?;
```
deuszx, line 196: Why are you updating the wallet even before checking if the chain was opened properly?

linera-faucet/server/src/lib.rs
```diff
@@ -118,17 +123,57 @@ where
...
     async fn do_claim(&self, owner: AccountOwner) -> Result<ClaimOutcome, Error> {
-        let client = self.context.lock().await.make_chain_client(self.chain_id)?;
+        let main_client = self
+            .context
+            .lock()
+            .await
+            .make_chain_client(self.main_chain_id)?;
+        let maybe_tmp_chain_id = *self.tmp_chain_id.lock().await;
+        let main_balance = main_client.local_balance().await?;
+        let tmp_chain_id = match maybe_tmp_chain_id {
```
deuszx, line 133: I think renaming to `faucet_chain_id` will make things clearer. Right now the `tmp_` prefix makes it look like it's just created and soon will be killed, which is confusing even more b/c down below we try to close it.

linera-faucet/server/src/lib.rs
```diff
@@ -118,17 +123,57 @@ where
...
+                    .lock()
+                    .await
+                    .update_wallet_for_new_chain(
+                        chain_id,
+                        Some(key_pair),
+                        certificate.block().header.timestamp,
+                    )
+                    .await?;
+                *self.tmp_chain_id.lock().await = Some(chain_id);
+                chain_id
```
deuszx, line 160: I'd extract this `None => { ... }` clause to a method - `fn open_new_faucet_chain`.

linera-faucet/server/src/lib.rs
```diff
@@ -118,17 +123,57 @@ where
...
+            }
+        };
+        let tmp_client = self.context.lock().await.make_chain_client(tmp_chain_id)?;
```
deuszx, line 163: If you rename `tmp_chain_id` to `faucet_chain_id` then this should be renamed to `faucet_client`.

linera-service/tests/linera_net_tests.rs
```diff
@@ -2988,6 +2991,88 @@ async fn test_end_to_end_faucet_with_long_chains(config: impl LineraNetConfig) -
...
+#[cfg_attr(feature = "scylladb", test_case(LocalNetConfig::new_test(Database::ScyllaDb, Network::Grpc) ; "scylladb_grpc"))]
+#[cfg_attr(feature = "dynamodb", test_case(LocalNetConfig::new_test(Database::DynamoDb, Network::Grpc) ; "aws_grpc"))]
+#[cfg_attr(feature = "kubernetes", test_case(SharedLocalKubernetesNetTestingConfig::new(Network::Grpc, BuildArg::Build) ; "kubernetes_grpc"))]
+#[cfg_attr(feature = "remote-net", test_case(RemoteNetTestingConfig::new(None) ; "remote_net_grpc"))]
+#[test_log::test(tokio::test)]
+async fn test_end_to_end_faucet_chain_limit(config: impl LineraNetConfig) -> Result<()> {
+    let _guard = INTEGRATION_TEST_GUARD.lock().await;
+    tracing::info!("Starting test {}", test_name!());
+
+    let chain_count = 3;
```
deuszx, line 3005: This isn't a _chain count_ thought but _max claims per chain_

linera-faucet/server/src/tests.rs
```diff
@@ -76,16 +80,20 @@ async fn test_faucet_rate_limiting() {
...
     let context = ClientContext {
-        client,
+        clients: BTreeMap::from_iter([(client.chain_id(), client), (client2.chain_id(), client2)]),
         update_calls: 0,
     };
     let context = Arc::new(Mutex::new(context));
     let root = MutationRoot {
-        chain_id,
+        main_chain_id,
+        tmp_chain_id: Arc::new(Mutex::new(Some(chain_id))),
```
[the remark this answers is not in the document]
afck: Yes, but then I'd need to implement a lot more functionality in the mock `ClientContext`, and it isn't strictly necessary to test the rate limiting arithmetic.

linera-faucet/server/src/lib.rs
afck: Good point! That precaution was probably unnecessary in the first place, and now it's even less important because we aren't even using the main chain here.

(Originally my thinking had been: If there is an error that doesn't _necessarily_ mean that no block was created, so better update the wallet, just to be safe. But the faucet is supposed to be the only owner anyway.)

deuszx (approved):

PR #3899, opened 2025-05-07
Address two issues with EVM

MathieuDutSik:
## Motivation

We identified two problems with the EVM, which we correct now. Also, some typos from the closed PR #3883 are put there.

## Proposal

The `gas_limit` is added to the BlobkEnv variable.

A more serious problem is the constructor argument for the contract being run in `execute_message`. It was incorrect when running a non-trivial constructor on another chain. The serialization was missing.

## Test Plan

A non-trivial constructor that does not do anything, just making a check, was introduced to `evm_example_execute_message.sol`. It triggers the previous problem.

## Release Plan

## Links

### DIFF
--- a/linera-execution/src/evm/database.rs
+++ b/linera-execution/src/evm/database.rs
@@ -9,6 +9,7 @@ use std::{
     sync::{Arc, Mutex},
 };
 
+use linera_base::vm::VmRuntime;
 use linera_views::common::from_bytes_option;
 use revm::{
     db::AccountState,
@@ -17,7 +18,13 @@ use revm::{
 };
 use revm_primitives::{address, Address, BlobExcessGasAndPrice, BlockEnv, EvmState, B256, U256};
 
-use crate::{BaseRuntime, Batch, ContractRuntime, ExecutionError, ViewError};
+use crate::{BaseRuntime, Batch, ContractRuntime, ExecutionError, ServiceRuntime, ViewError};
+
+// The runtime costs are not available in service operations.
+// We need to set a limit to gas usage in order to avoid blocking
+// the validator.
+// We set up the limit similarly to Infura to 20 million.
+pub const EVM_SERVICE_GAS_LIMIT: u64 = 20_000_000;
 
 /// The cost of loading from storage.
 const SLOAD_COST: u64 = 2100;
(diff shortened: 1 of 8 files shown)


linera-execution/src/evm/database.rs
```diff
@@ -378,4 +379,35 @@ where
             blob_excess_gas_and_price,
         })
     }
+
+    pub fn constructor_argument(&self) -> Result<Vec<u8>, ExecutionError> {
+        let mut runtime = self.runtime.lock().expect("The lock should be possible");
+        let constructor_argument = runtime.application_parameters()?;
+        Ok(serde_json::from_slice::<Vec<u8>>(&constructor_argument)?)
```
deuszx, line 386: Why do we use `serde_json` here? I imagine that in order for this to work, we'd have to serialize with serde as well.

MathieuDutSik: Because the application parameters are serialized with `serde_json`.

linera-service/tests/fixtures/evm_example_execute_message.sol
```diff
@@ -8,6 +8,10 @@ contract ExampleExecuteMessage {
     uint64 value;
     Linera.MessageId last_message_id;
 
+    constructor(uint64 test_value) {
+       require(test_value == 42);
```
deuszx, line 12: Can you modify the constructor so that it takes the `test_value` and sets one of its field to it and later check, by calling contract's method, whether it was set properly?

It's not obvious that in this test the constructor is even called.

linera-execution/src/evm/database.rs
```diff
@@ -378,4 +379,35 @@ where
             blob_excess_gas_and_price,
         })
     }
+
+    pub fn constructor_argument(&self) -> Result<Vec<u8>, ExecutionError> {
+        let mut runtime = self.runtime.lock().expect("The lock should be possible");
+        let constructor_argument = runtime.application_parameters()?;
```
deuszx, line 385: So application _parameters_ are treated differently between EVM and Linera VM? i.e. for the latter they are application-specific parameters that are different from a constructor?

linera-service/tests/fixtures/evm_example_execute_message.sol
MathieuDutSik: Well, you can run with 41 and see that the test does not pass.

linera-execution/src/evm/database.rs
MathieuDutSik: The application parameters are treated the same way in Wasm and Evm:
* They are serialized with `serde_json`.
* For the EVM, this is somewhat cumbersome since the input is `Vec<u8>`, but that is how it is.

deuszx (approved):

PR #3905, opened 2025-05-07
Set default fallback duration to `MAX`.

afck:
## Motivation

The validators are not currently expected to start client processes automatically. So chains that enter fallback mode are effectively bricked.

## Proposal

Set the default fallback timeout to `MAX`, so validators will never sign a fallback timeout certificate.

## Test Plan

(Only a default chain config change.)

## Release Plan

## Links

### DIFF
--- a/linera-core/src/unit_tests/worker_tests.rs
+++ b/linera-core/src/unit_tests/worker_tests.rs
@@ -185,11 +185,21 @@ where
         owner: AccountOwner,
         balance: Amount,
     ) -> ChainDescription {
+        self.add_root_chain_with_ownership(index, balance, ChainOwnership::single(owner))
+            .await
+    }
+
+    async fn add_root_chain_with_ownership(
+        &mut self,
+        index: u32,
+        balance: Amount,
+        ownership: ChainOwnership,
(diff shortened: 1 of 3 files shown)


linera-core/src/unit_tests/worker_tests.rs
```diff
@@ -3560,7 +3570,11 @@ where
     let public_key = signer.generate_new();
     let mut env = TestEnvironment::new(storage, false, false).await;
     let balance = Amount::from_tokens(5);
-    let chain_1_desc = env.add_root_chain(1, public_key.into(), balance).await;
+    let mut ownership = ChainOwnership::single(public_key.into());
+    ownership.timeout_config.fallback_duration = TimeDelta::from_secs(5);
```
deuszx, line 3574: It'd be worth adding a comment why we set the fallback manually and why to 5 seconds.

afck: Done in https://github.com/linera-io/linera-protocol/pull/3905/commits/17661636e0526aa53a83cea225f20c2eecd65272.

PR #3906, opened 2025-05-07
`linera-core`: more sophisticated blob download algorithm

Twey:
## Motivation

Our blob download can sometimes be very slow when we have broken or slow validators.

## Proposal

Use an algorithm that hedges the download with other validators.

## Test Plan

I hope that CI will suffice for correctness; for performance I will try against the testnet from the Web.

## Release Plan

- These changes will be released in a new minor version of `@linera/client`, but otherwise follow the normal release process.

## Links

### DIFF
--- a/linera-core/src/remote_node.rs
+++ b/linera-core/src/remote_node.rs
@@ -349,15 +349,101 @@ impl<N: ValidatorNode> RemoteNode<N> {
         validators: &[Self],
         timeout: Duration,
     ) -> Option<Vec<Blob>> {
-        let mut stream = blob_ids
-            .iter()
-            .map(|blob_id| Self::download_blob(validators, *blob_id, timeout))
-            .collect::<FuturesUnordered<_>>();
-        let mut blobs = Vec::new();
-        while let Some(maybe_blob) = stream.next().await {
-            blobs.push(maybe_blob?);
+        use futures::FutureExt as _;
+        use std::collections::{HashMap, VecDeque};
+
+        let validators: HashMap<_, _> = validators.into_iter().map(|v| (&v.public_key, v)).collect();
+
+        // randomize the blobs
+        // assign slots to validators
+        // sort the validators by slots
+        // for each blob, ready a timeout event with no validator
+        // for each event:
+        // - if okay, amend the completed blobs and add a slot to the validator (and cancel)
+        // - if cancel, add a slot to the validator
+        // - if timeout or error, put the task back on the queue (but don't increase slots)
+        // - if validator slots and tasks in queue, assign task to validator and start new task and timeout
+        // if incomplete, return None, else return all blobs
+
+        enum Event {
+            Complete {
+                node: ValidatorPublicKey,
+                // `None` if the task was canceled (another task already completed the blob).
+                blob: Option<Blob>,
+            },
+            Incomplete {
+                blob_id: BlobId,
+            },
         }
-        Some(blobs)
+
+        let mut events = FuturesUnordered::default();
+        for blob_id in blob_ids {
+            let blob_id = *blob_id;
+            events.push(async move {
+                Event::Incomplete { blob_id }
+            }.boxed_local());
+        }
+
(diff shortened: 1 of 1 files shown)


linera-core/src/remote_node.rs
```diff
@@ -349,15 +349,101 @@ impl<N: ValidatorNode> RemoteNode<N> {
...
+        }
+
+        if !remaining_blobs.is_empty() {
+            None
+        } else {
+            let mut blobs = Vec::new();
+            for blob_id in blob_ids {
+                blobs.push(completed_blobs.remove(blob_id).expect("missing blob"));
+            }
+            Some(blobs)
```
deuszx, line 445: Can't we return `completed_blobs.values()`? Or is it just additional safety measure to make sure all blobs are fetched?

linera-core/src/remote_node.rs
```diff
@@ -349,15 +349,101 @@ impl<N: ValidatorNode> RemoteNode<N> {
...
+        let mut events = FuturesUnordered::default();
+        for blob_id in blob_ids {
+            let blob_id = *blob_id;
+            events.push(async move {
+                Event::Incomplete { blob_id }
+            }.boxed_local());
+        }
+
+        let num_slots = blob_ids.len().div_ceil(validators.len());
+        let mut slots: HashMap<_, _> = validators.keys().map(|v| (**v, num_slots)).collect();
```
afck, line 388: Nit: I think this will leave the last validator(s) (by hash) with almost no work, if the ceiling is much higher than the quotient. E.g. if there are five validators and six blobs, the first three validators will each get two jobs, and the last two none.

afck (commented):
The algorithm looks good to me. Should definitely improve performance. :+1:

linera-core/src/remote_node.rs
bart-linera: Instead of selecting the first validator with a free slot, we could select the first among those with the highest number of slots. Then in your example, one validator would get two jobs, and the rest would get one each.

linera-core/src/remote_node.rs
Twey: Since it's a `Vec` at the time I thought that the return `Vec` was intended to match the order of the `blob_ids` `Vec`, but on reading the consumer code a bit more I see that's not the case, so this is redundant.

linera-core/src/remote_node.rs
deuszx: I don't think prioritizing by weight is a good idea. Validators should be equally important, regardless of their stake. Their stake should only influence how often they have the chance to "earn" but not how important they are for security.

afck (commented):
Actually… doesn't this still allow a single slow validator to create a long delay?

A blob ID is not in `remaining_blobs` as long as any validator is busy with it, and we don't make another validator try that blob until the busy one either returns the blob or an error, do we?

deuszx:
> Actually… doesn't this still allow a single slow validator to create a long delay?
> 
> A blob ID is not in `remaining_blobs` as long as any validator is busy with it, and we don't make another validator try that blob until the busy one either returns the blob or an error, do we?

I think the timeout would kick in and the blob would be fetched from someone else?

afck:
But we don't want to ever wait for the full gRPC timeout just because _one_ validator is slow.

linera-core/src/remote_node.rs
Twey: I think that in the Linera protocol weight is exactly how important they are for security.  But it wouldn't be a security issue here.  The idea is just that validators with more stake should be willing to serve more traffic, since they benefit from it more.

afck:
Ah, right, you're just creating a whole separate task for that! :+1: 

Then we should probably, when handling `Event::Incomplete`, and maybe also in the loop before creating new events, check whether the blob is in `completed_blobs` by now. Because currently a blob can be in both collections.

Twey:
That's true!  I forgot :sweat_smile: Thanks!

PR #3908, opened 2025-05-07
Integrate the Web client library from `linera-web`

Twey:
## Motivation

We'd like to be able to test changes to this repo using a Web client.

## Proposal

Merge the Rust code from https://github.com/linera-io/linera-web/ into this repository under `linera-web/`, making the JavaScript `@linera/client` library available for the Web.

## Test Plan

CI.

## Release Plan

## Links


linera-web/Cargo.toml
```diff
@@ -0,0 +1,67 @@
...
+futures = "0.3.30"
+js-sys = "0.3"
+log = "0.4.21"
+nonzero_lit = "0.1.2"
+serde = "=1.0.205"
+serde-wasm-bindgen = "0.6.5"
+tokio-util.workspace = true
+tracing = "0.1.40"
+wasm-bindgen = "0.2.95"
+wasm-bindgen-futures = "0.4"
```
ndr-ds, line 27: Any reason why we're not grabbing these from the workspace? Or adding it there if it isn't 🤔
That might be what's generating the huge `Cargo.lock` file

ma2bd (approved):

Twey:
Manually merged as https://github.com/linera-io/linera-protocol/commit/43e630976f987eb6ea88ae4ea0c9641647dfeb3a (and previous commits: the link between them, and the original `linera-web` commits, was destroyed because we don't allow merge commits on `main`)

linera-web/Cargo.toml
Twey: Addressing in a follow-up PR since there might be changes needed to be compatible with the right versions.

linera-web/rust-toolchain.toml
```diff
@@ -0,0 +1 @@
+../linera-protocol/toolchains/nightly/rust-toolchain.toml
```
ndr-ds, line 1: Were these symlinks merged? The path seems wrong, no? Or am I missing something? 🤔
Seeing docker builds fail from main saying these files don't exist

Twey: Yes, sorry!  I thought the Docker build here was sufficient to ensure the Docker build off main succeeds.  Should be fixed in https://github.com/linera-io/linera-protocol/pull/3955.

Issue #3399, opened 2025-02-23
Remove 2 deprecated CLI options of `linera wallet init`


ma2bd:
Remove 2 CLI options of `linera wallet init`:
* `--with-new-chain` (people should use `linera wallet request-chain`)
* `--with-other-chains` (people should use `linera wallet follow-chain`)

Also remove the `FaucetOptions` in `cli_wrapper`.

This should be easy but
* e2e tests need to be upgraded (mostly simplified)
* We need to decide what to do with `print_peg_certificate_hash`

ownwift:
good

afck:
> We need to decide what to do with `print_peg_certificate_hash`

Maybe `linera wallet init` should just always synchronize the admin chain, and print its latest hash.

afck:
I realized we already don't sync automatically on wallet initialization anymore.

I'll remove the info message for now and created https://github.com/linera-io/linera-protocol/issues/3909.

Issue #3909, opened 2025-05-08
Print an info message when using faucet or old local data.


afck:
When initializing a wallet using the genesis config provided by a remote node (e.g. the faucet), the client basically trusts the faucet that it will connect it to the right network.

Similarly, if the client has been offline very long and the latest locally known committee is already expired, syncing based on that information means trusting that this committee has not been compromised. (Long-Range Attack)

In both cases we should print a message after synchronization, so that the user can compare the downloaded data to some trusted source (block explorer, other users, etc.). Specifically, we should print the latest admin chain block hash. If the admin chain is empty and the genesis timestamp is still recent enough, we should print the genesis config hash.

afck:
We could make that warning less irritating by having a configurable list of trusted sources: Endpoints that can provide the latest committee and/or admin chain block hashes, and that the user can trust to some extent even though they aren't necessarily staked validators.

In any situation where the client can't trust the latest committee anymore or doesn't have any data yet, it would query _all_ of these sources. If they all agree, it would trust the committee they provide again, and maybe only print a less alarming INFO-level message. If any of them disagrees it should print a very clear error message and probably terminate.

PR #3910, opened 2025-05-08
Remove 2 deprecated CLI options of `linera wallet init`.

afck:
## Motivation

`--with-new-chain` and `--with-other-chains` are not necessary anymore; `request-chain` and `follow-chain` should be used instead.

## Proposal

Remove them, and the `FaucetOption` enum.

Also don't print the latest hash anymore: This was meant to avoid trusting the faucet or expired committees, but we already don't sync on `wallet init` anymore, so it's in the wrong place. I created #3909 to revisit this once we address the rest of the long-range attack issue.

## Test Plan

The tests were updated to use `request-chain` instead.

## Release Plan

- Before the next testnet, the Linera Manual needs to be updated.

## Links

- Closes https://github.com/linera-io/linera-protocol/issues/3399.

### DIFF
--- a/linera-service/src/linera/main.rs
+++ b/linera-service/src/linera/main.rs
@@ -6,7 +6,7 @@
 #![deny(clippy::large_futures)]
 
 use std::{
-    collections::{BTreeMap, BTreeSet, HashMap},
+    collections::{BTreeMap, BTreeSet},
     env,
     ops::Deref,
     path::PathBuf,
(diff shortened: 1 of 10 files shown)


linera-service/src/linera/main.rs
```diff
@@ -1235,48 +1235,6 @@ impl Runnable for Job {
                 );
             }
 
-            Wallet(WalletCommand::Init {
```
deuszx, line 1238: Why is it safe to remove handling of this command here now?

afck: We `match` on the command in two places:
* In the `run()` function, where storage is not yet initialized. `Wallet::Init` is handled there, because as part of wallet initialization we also initialize storage.
* In the `<Job as Runnable>::run<S>(self, storage: S)` method, where storage is already available. We used to assign and sync the new chain there, but now that's not necessary anymore.

christos-h (approved):
LGTM :)

PR #3912, opened 2025-05-08
Fix error msg enrichement for Bytecode::load_from_file

deuszx:
## Motivation

When we fail to load the `Bytecode` the errors message doesn't contain the path:
```
2025-05-08T11:07:28.399520Z ERROR linera: Error is persistence error: I/O error: No such file or directory (os error 2)
```

even though the code tries to add it. The issue seems to be due to a mismatch between the error type returned `load_from_file` function and the expectations of `.with_context` from the `anyhow` crate.

`load_from_file` returns `std::io::Result<Self>`, which means the error type is `std::io::Error`. However, `with_context` is only available on `Result<T, anyhow::Error>`, and it only adds context to errors that are already converted to `anyhow::Error`.

## Proposal

Enrich the log manually on the callsite.

## Test Plan

Tested locally and the result is now
```
2025-05-08T11:20:04.498782Z ERROR linera: Error is persistence error: failed to load contract bytecode from "./target/wasm32-unknown-unknown/release/my-counter_contract.wasm". error: No such file or directory (os error 2)
```

## Release Plan

## Links

### DIFF
--- a/linera-client/src/client_context.rs
+++ b/linera-client/src/client_context.rs
@@ -640,12 +640,24 @@ where
         vm_runtime: VmRuntime,
     ) -> Result<ModuleId, Error> {
         info!("Loading bytecode files");
-        let contract_bytecode = Bytecode::load_from_file(&contract)
-            .await
-            .with_context(|| format!("failed to load contract bytecode from {:?}", &contract))?;
-        let service_bytecode = Bytecode::load_from_file(&service)
-            .await
-            .with_context(|| format!("failed to load service bytecode from {:?}", &service))?;
+        let contract_bytecode = Bytecode::load_from_file(&contract).await.map_err(|e| {
+            error::Inner::Persistence(
+                format!(
+                    "failed to load contract bytecode from {:?}. error: {}",
+                    &contract, e
+                )
+                .into(),
+            )
+        })?;
+        let service_bytecode = Bytecode::load_from_file(&service).await.map_err(|e| {
+            error::Inner::Persistence(
+                format!(
+                    "failed to load service bytecode from {:?}. error: {}",
+                    &service, e
+                )
+                .into(),
+            )
+        })?;
 
         info!("Publishing module");
         let (blobs, module_id) =


linera-client/src/client_context.rs
```diff
@@ -640,12 +640,24 @@ where
         vm_runtime: VmRuntime,
     ) -> Result<ModuleId, Error> {
         info!("Loading bytecode files");
-        let contract_bytecode = Bytecode::load_from_file(&contract)
-            .await
-            .with_context(|| format!("failed to load contract bytecode from {:?}", &contract))?;
-        let service_bytecode = Bytecode::load_from_file(&service)
-            .await
-            .with_context(|| format!("failed to load service bytecode from {:?}", &service))?;
+        let contract_bytecode = Bytecode::load_from_file(&contract).await.map_err(|e| {
```
afck, line 643: (Would be great to do this inside `load_from_file` directly. But that would probably mean we'd have yet another error type?)

afck (approved):
Looks good to me!

I guess the `Debug` impl of `PathBuf` prints the lossy UTF-8 conversion?

afck:
(And is there some `impl From<io::Error>` impl and error variant that could now be removed?)

linera-client/src/client_context.rs
deuszx: That method returns `io::Error` and I didn't want to change that to the custom `Inner` error type (which has some macro magic enriched that could make it more complicated than necessary).

deuszx:
> (And is there some `impl From<io::Error>` impl and error variant that could now be removed?)

Good question - looks like [there is](https://github.com/linera-io/linera-protocol/blob/main/linera-client/src/persistent/file.rs#L19-L20).

Maybe @Twey can share his depthless wisdom here.

Twey:
A better and smaller solution here would probably be to replace `std::fs` with [`fs_err`](https://docs.rs/fs-err/latest/fs_err/) in `linera-base`.

deuszx:
> A better and smaller solution here would probably be to replace the import of `std::fs` with [`fs_err`](https://docs.rs/fs-err/latest/fs_err/) in `linera-base`. We try to use that elsewhere but we seem to have overlooked `linera-base`.

I can give it a try but - in principle - I don't like adding a whole dependency for something that requires 2 lines of code

afck:
We're already using `fs-err`, at least in linera-version.

Twey:
It doesn't require 2 lines of code to do the equivalent thing, this approach requires 2 lines of code × the number of uses of `std::fs` in `linera-base` × the number of callers :)

A bigger problem here, though, is that the conversion from `linera_client::Error` to `anyhow::Error` drops the context.  It might be good (separately) to add

```rust
impl Error {
    fn to_anyhow(self) -> anyhow::Error {
        let contexts: Vec<_> = self.all(vec![]).0.into_iter().cloned().collect();
        let mut error = anyhow::Error::from(self);
        for context in contexts {
            error = error.context(context);
        }
        error
    }
}
```

and make sure to `.map_err(linera_client::Error::to_anyhow)` in `linera_service` when we cast to `anyhow::Error`.

deuszx:
I didn't add anyhow though - I actually removed it.

Twey:
Hence ‘separately’ — it would be great to do the more general thing in a separate PR :)

Twey:
> I actually removed it.

Sorry, I see the confusion — the `Context` trait here that provides `context()` is not `anyhow::Context` but `thiserror_context::Context`.  The reason this code doesn't work as expected is that converting from `linera_client::Error` to `anyhow::Error` (with `impl<E: std::error::Error> From<E> for anyhow::Error`) drops the context from `thiserror_context::Context` because it goes via `std::error::Error`, which has no notion of context.  In this specific case the context shouldn't have been necessary in the first place, because `fs-err` already does this for us.  But in general there is other context we'd like to preserve, so we should avoid that conversion and instead provide an implementation that preserves the context.  (The implementation I gave above can be generalized to all implementations of `thiserror_context::Context` with a few more type variables, and then we can similarly fix the problem for any other crates that might be experiencing it.)

ma2bd:
@deuszx @Twey what do we want to do with this PR?

deuszx:
I'm gonna close this since I don't really want to spend time on the approach that @Twey suggested - higher priority issues at hand.

PR #3913, opened 2025-05-08
Improve download certificate performance.

bart-linera:
Continuation of #3862 

## Motivation

Currently, slow or inactive validators can slow down the process of downloading certificates. We can prevent this by downloading them from multiple validators in parallel.

## Proposal

Download certificates from 1/3 of the validators in parallel instead of trying the validators one by one.

## Test Plan

CI will make sure that no functionality is broken.
Manual benchmarks against the testnet will be used to determine if the performance improvement is noticeable.

## Release Plan

    - be released in a new SDK,

## Links

- [Previous PR](https://github.com/linera-io/linera-protocol/pull/3862)

### DIFF
--- a/linera-client/src/client_context.rs
+++ b/linera-client/src/client_context.rs
@@ -410,41 +410,34 @@ where
         S: Storage + Clone + Send + Sync + 'static,
     {
         let node_provider = self.make_node_provider();
-        let client = self.client.clone_with(
-            node_provider.clone(),
-            "Temporary client for fetching the parent chain",
-            vec![message_id.chain_id, chain_id],
-            false,
-        );
+        self.client.track_chain(chain_id);
 
         // Take the latest committee we know of.
         let admin_chain_id = self.wallet.genesis_admin_chain();
         let query = ChainInfoQuery::new(admin_chain_id).with_committees();
-        let nodes: Vec<_> = if let Some(validators) = validators {
-            node_provider
-                .make_nodes_from_list(validators)?
-                .map(|(public_key, node)| RemoteNode { public_key, node })
-                .collect()
-        } else {
-            let info = client.local_node().handle_chain_info_query(query).await?;
-            let committee = info
-                .latest_committee()
-                .ok_or(error::Inner::ChainInfoResponseMissingCommittee)?;
-            node_provider
-                .make_nodes(committee)?
-                .map(|(public_key, node)| RemoteNode { public_key, node })
-                .collect()
-        };
+        let info = self
+            .client
+            .local_node()
+            .handle_chain_info_query(query)
+            .await?;
+        let committee = info
+            .latest_committee()
+            .ok_or(error::Inner::ChainInfoResponseMissingCommittee)?;
+        let nodes: Vec<_> = node_provider
+            .make_nodes(committee)?
+            .map(|(public_key, node)| RemoteNode { public_key, node })
+            .collect();
 
         // Download the parent chain.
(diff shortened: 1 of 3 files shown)


linera-client/src/client_context.rs
```diff
@@ -404,47 +404,45 @@ where
...
         let target_height = message_id.height.try_add_one()?;
-        client
-            .download_certificates(&nodes, message_id.chain_id, target_height)
+        self.client
+            .download_certificates(
+                &nodes,
+                message_id.chain_id,
+                target_height,
+                committee.clone(),
+            )
```
afck, line 439: Can we just use `synchronize_from_validators` now, so we don't have to bother with the committee here? (Not sure if that's a bigger change or just works out of the box.)

…and use this new algorithm in `synchronize_chain_state`, too?

linera-client/src/client_context.rs
```diff
@@ -404,47 +404,45 @@ where
...
         S: Storage + Clone + Send + Sync + 'static,
     {
         let node_provider = self.make_node_provider();
-        let client = self.client.clone_with(
-            node_provider.clone(),
-            "Temporary client for fetching the parent chain",
-            vec![message_id.chain_id, chain_id],
-            false,
-        );
+        self.client.track_chain(chain_id);
```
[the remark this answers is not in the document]
bart-linera: I think that just got ported from `main` by accident, TBH :sweat_smile:

linera-client/src/client_context.rs
```diff
@@ -414,14 +414,15 @@ where
...
+            .handle_chain_info_query(query)
+            .await?;
         let committee = info
-            .epoch
-            .and_then(|epoch| {
-                info.requested_committees
-                    .as_ref()
-                    .and_then(|committees| committees.get(&epoch))
-            })
+            .latest_committee()
```
afck, line 425: I was about to remove that in https://github.com/linera-io/linera-protocol/pull/3929. :innocent: 

Anyway, I think most of this code is just the `Client::epoch_and_committee` that I'm introducing there.

bart-linera: Thanks for the heads up! It doesn't affect this PR, but it would affect porting it to `main`, if we decide to do so :+1:

linera-service/src/linera/main.rs
```diff
@@ -1134,18 +1134,12 @@ impl Runnable for Job {
                     .await?;
                 let faucet = cli_wrappers::Faucet::new(faucet_url);
                 let outcome = faucet.claim(&owner).await?;
-                let validators = faucet.current_validators().await?;
```
afck, line 1137: I guess for the existing testnet that's fine, but let's not do that on main: When initializing the wallet, we _have_ to get the _current_ validators from somewhere. Our local storage will only know the genesis validators and they might be long gone.

bart-linera: `assign_chain_to_key` synchronizes the admin chain and uses the latest validators from there, so that should be fine, I think?

afck: But whom does it sync the admin chain from? It has to start with the genesis validators, and they might not be online anymore.

There's no alterantive to getting the current validators from another (somewhat trusted—see https://github.com/linera-io/linera-protocol/issues/3909) source on initialization.

bart-linera: Hmm, good point :thinking:

bart-linera:
Closing, as it's no longer relevant, I believe.

PR #3915, opened 2025-05-08
Move cross-chain error log to after the loop.

afck:
## Motivation

The error log `Dropping cross-chain query` is printed on every retry attempt.

## Proposal

Print it only after the last attempt.

## Test Plan

CI

## Release Plan

## Links


christos-h (approved):
Makes sense!

PR #3916, opened 2025-05-09
Remove admin chain ID from genesis config

bart-linera:
## Motivation

The `admin_id` field in `GenesisConfig` is now reduntant, because the ID of the admin chain is implied by the rest of the config. It can only cause trouble if a wrong ID is supplied.

## Proposal

Remove `admin_id` from `GenesisConfig`. Make `GenesisConfig` compute the ID based on the rest of the input. Make chain initialization more consistent in other places, too.

## Test Plan

## Release Plan

## Links


linera-client/src/config.rs
```diff
@@ -267,21 +266,49 @@ impl BcsSignable<'_> for GenesisConfig {}
...
+            epoch: Epoch::ZERO,
+            ownership: ChainOwnership::single(public_key.into()),
+        };
+        let description = ChainDescription::new(origin, config, self.timestamp);
+        self.chains.push(description.clone());
+        description
+    }
+
+    pub fn admin_chain_description(&self) -> &ChainDescription {
+        &self.chains[0]
```
ma2bd, line 305: If the admin chain is always the first root chain of the genesis config, this needs to be documented, and the constructor needs to be changed to ensure that `chains` is never empty.

linera-client/src/config.rs
```diff
@@ -267,21 +266,49 @@ impl BcsSignable<'_> for GenesisConfig {}
...
             committee,
-            admin_id,
             timestamp,
             chains: Vec::new(),
             policy,
             network_name,
         }
     }
 
+    pub fn add_chain(&mut self, public_key: AccountPublicKey, balance: Amount) -> ChainDescription {
```
ma2bd, line 282: `add_(simple_)root_chain` ?

linera-client/src/config.rs
```diff
@@ -267,21 +266,49 @@ impl BcsSignable<'_> for GenesisConfig {}
 impl GenesisConfig {
     pub fn new(
```
ma2bd, line 267: This constructor creates an invalid value (see below).

linera-service/src/linera/main.rs
```diff
@@ -1872,58 +1868,29 @@ async fn run(options: &ClientOptions) -> Result<i32, Error> {
...
             )?;
-            let mut chains = vec![];
-            for i in 0..=*num_other_initial_chains {
+            let admin_chain_description =
+                genesis_config.add_chain(admin_public_key, *initial_funding);
+            let mut chains = vec![UserChain::make_initial(
+                admin_public_key.into(),
+                admin_chain_description,
+                timestamp,
+            )];
```
ma2bd, line 1888: Can we move this to the constructor? (see other comments)

linera-client/src/config.rs
bart-linera: TBH, I wasn't sure if we want that - but is there a use case for the admin chain to not be the first root chain? I think we always use Root(0) as the admin chain, anyway. But I can try to figure out something to keep this configurable if we need it.
Good point on changing the constructor.

linera-client/src/config.rs
```diff
@@ -254,34 +254,92 @@ impl<S: Deref<Target = InMemorySigner>> SignerState<S> {
...
+) -> ChainDescription {
+    let committees: BTreeMap<_, _> = [(
+        Epoch::ZERO,
+        bcs::to_bytes(committee).expect("serializing a committee should not fail"),
+    )]
+    .into_iter()
+    .collect();
+    let origin = ChainOrigin::Root(index);
+    let config = InitialChainConfig {
+        admin_id,
```
afck, line 281: Do we need that in `InitialChainConfig` at all?

afck (approved):

linera-execution/src/committee.rs
```diff
@@ -243,8 +243,8 @@ impl Committee {
                 (
                     validator_key,
                     ValidatorState {
-                        network_address: validator_key.to_string(),
-                        votes: 1,
+                        network_address: "Tcp:localhost:8080".to_string(),
```
ma2bd, line 246: This seems a bit fragile. Can't we use a format string with proper Rust values?

linera-client/src/config.rs
ma2bd: @bart-linera I still don't see any documentation about the new convention.

PR #3920, opened 2025-05-12
Fix authenticated signer when re-proposing a fast block.

afck:
## Motivation

When re-proposing a validated block from an earlier round there is a validated block certificate, signed by a quorum of validators, attesting the correctness of the outcome and the validity of the oracle responses—_and_ that the `authenticated_signer` is actually the one who originally submitted the block proposal to the validators!

This last part I missed in the special case where we re-propose an earlier proposal by a super owner from the fast round! In that case, there is no validated block certificate, which is why we disallow oracles. But we erroneously compare the `authenticated_signer` to the owner _re-proposing_ the block, rather than the original super owner.

Of course comparing it to the super owner is not enough: The regular owner must not be able to do something in the super owner's name without permission. So we need to also verify the super owner's signature (and super ownership!) again.

Super owners by design take greater responsibility for a chain's liveness than regular owners, and in the case of re-proposing a _fast_ block, the super owner's signature needs to play a similar role to the validators' signatures when re-proposing a _regular_ block.

## Proposal

Properly distinguish the _three_ cases of a block proposal:
* The current proposer is the one who originally created this block. They are the authenticated signer.
* This is a retry of an earlier proposal by a super owner in the _fast_ round: The super owner is the authenticated signer, but their signature must be included in the proposal and verified again.
* This is a retry of an earlier proposal in regular round: The original proposer is the authenticated signer, but we don't need to verify their signature again; the validators' signatures of the earlier round's certificate already proves that the proposal is valid (in addition to proving that the included oracle responses are, too).

## Test Plan

`test_re_propose_fast_block` was added.

## Release Plan

## Links

### DIFF
--- a/linera-chain/src/data_types.rs
+++ b/linera-chain/src/data_types.rs
@@ -204,6 +204,21 @@ pub struct MessageBundle {
     pub messages: Vec<PostedMessage>,
 }
 
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[cfg_attr(with_testing, derive(Eq, PartialEq))]
+/// An earlier proposal that is being retried.
+pub enum OriginalProposal {
+    /// A proposal in the fast round.
+    Fast {
+        public_key: AccountPublicKey,
+        signature: AccountSignature,
+    },
+    /// A validated block certificate from an earlier round.
+    Regular {
+        certificate: LiteCertificate<'static>,
+    },
+}
+
 /// An authenticated proposal for a new block.
 // TODO(#456): the signature of the block owner is currently lost but it would be useful
 // to have it for auditing purposes.
@@ -214,7 +229,7 @@ pub struct BlockProposal {
     pub public_key: AccountPublicKey,
     pub signature: AccountSignature,
     #[debug(skip_if = Option::is_none)]
-    pub validated_block_certificate: Option<LiteCertificate<'static>>,
+    pub original_proposal: Option<OriginalProposal>,
 }
 
 /// A message together with kind, authentication and grant information.
@@ -502,17 +517,42 @@ impl BlockProposal {
             content,
             public_key,
             signature,
-            validated_block_certificate: None,
+            original_proposal: None,
         })
     }
 
-    pub async fn new_retry(
+    pub async fn new_retry_fast(
+        owner: AccountOwner,
+        round: Round,
+        old_proposal: BlockProposal,
+        signer: &(impl Signer + ?Sized),
+    ) -> Result<Self, Box<dyn Error>> {
+        let content = ProposalContent {
+            round,
+            block: old_proposal.content.block,
+            outcome: None,
+        };
+        let signature = signer.sign(&owner, &CryptoHash::new(&content)).await?;
+        let public_key = signer.get_public_key(&owner).await?;
+
+        Ok(Self {
+            content,
+            public_key,
+            signature,
+            original_proposal: Some(OriginalProposal::Fast {
+                public_key: old_proposal.public_key,
+                signature: old_proposal.signature,
+            }),
+        })
+    }
+
+    pub async fn new_retry_regular(
         owner: AccountOwner,
         round: Round,
         validated_block_certificate: ValidatedBlockCertificate,
         signer: &(impl Signer + ?Sized),
     ) -> Result<Self, Box<dyn Error>> {
-        let lite_cert = validated_block_certificate.lite_certificate().cloned();
+        let certificate = validated_block_certificate.lite_certificate().cloned();
         let block = validated_block_certificate.into_inner().into_inner();
         let (block, outcome) = block.into_proposal();
         let content = ProposalContent {
@@ -527,7 +567,7 @@ impl BlockProposal {
             content,
             public_key,
             signature,
-            validated_block_certificate: Some(lite_cert),
+            original_proposal: Some(OriginalProposal::Regular { certificate }),
         })
     }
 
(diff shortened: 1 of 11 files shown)


linera-chain/src/manager.rs
```diff
@@ -452,21 +455,39 @@ where
...
-                .as_ref()
-                .is_none_or(|locking| locking.round() < lite_cert.round)
-            {
-                let value = ValidatedBlock::new(block.clone());
-                if let Some(certificate) = lite_cert.clone().with_value(value) {
-                    self.update_locking(LockingBlock::Regular(certificate), blobs.clone())?;
+        match &proposal.original_proposal {
+            Some(OriginalProposal::Regular { certificate }) => {
+                if self
+                    .locking_block
```
deuszx, line 461: Not from this PR but why is it _locking_ block and not _locked_?

linera-chain/src/data_types.rs
```diff
@@ -502,17 +517,42 @@ impl BlockProposal {
...
             signature,
-            validated_block_certificate: None,
+            original_proposal: None,
+        })
+    }
+
+    pub async fn new_retry_fast(
+        owner: AccountOwner,
+        round: Round,
+        old_proposal: BlockProposal,
```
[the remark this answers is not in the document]
afck: The fast round comes before all others, so it's possible to re-propose a block from the fast round in later rounds, but not vice-versa.

The constructors need different inputs: If the original proposal was a fast block, we only have the proposal itself, with the super owner's signature. If it was a regular block, we have the validators' signatures, i.e. the certificate.

linera-chain/src/manager.rs
afck: The term "locked" is specifically relative to a validator: It's the highest block have voted to confirm, and they will refuse to accept any _different_ proposals in later rounds unless they are presented with a higher validated block certificate.

What's contained in the proposal is the latter: the thing that can unlock a validator, and make them sign a new confirmed vote and therefore consider a different block "locked". So the locking block is a proof that makes the validator change their locked block.
(I'm always open to ideas for better names. :sweat_smile:)

linera-chain/src/manager.rs
```diff
@@ -452,21 +455,39 @@ where
...
+                    .as_ref()
+                    .is_none_or(|locking| locking.round() < certificate.round)
+                {
+                    let value = ValidatedBlock::new(block.clone());
+                    if let Some(certificate) = certificate.clone().with_value(value) {
+                        self.update_locking(LockingBlock::Regular(certificate), blobs.clone())?;
+                    }
+                }
+            }
+            Some(OriginalProposal::Fast {
```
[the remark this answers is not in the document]
afck: I'll update it.

Now that I think about it, this match arm is actually unnecessary, since we're not immediately voting to confirm.

Maybe it's still better to leave it in: That way, a client synchronizing from us will also be informed about the fast block proposal, and will re-propose it, making it less likely that their round fails.

linera-core/src/unit_tests/client_tests.rs
```diff
@@ -2250,6 +2252,117 @@ where
     Ok(())
 }
 
+#[test_case(MemoryStorageBuilder::default(); "memory")]
+#[cfg_attr(feature = "storage-service", test_case(ServiceStorageBuilder::new().await; "storage_service"))]
+#[cfg_attr(feature = "rocksdb", test_case(RocksDbStorageBuilder::new().await; "rocks_db"))]
+#[cfg_attr(feature = "dynamodb", test_case(DynamoDbStorageBuilder::default(); "dynamo_db"))]
+#[cfg_attr(feature = "scylladb", test_case(ScyllaDbStorageBuilder::default(); "scylla_db"))]
+#[test_log::test(tokio::test)]
+async fn test_re_propose_fast_block<B>(storage_builder: B) -> anyhow::Result<()>
```
deuszx, line 2261: To be honest, I don't understand why this test has to be so complicated 🤔 . I thought that we want to test that re-proposing a fast block (originally proposed by superowner) by a regular owner should succeed. It shouldn't require so many steps. Wont' this be enough:?
- we set 3/4 of validators to faulty first
- superowner tries to propose a block in the `Fast` round. It should fail b/c we don't have a quorum of validators
- we set all validators to honest again
- regular owner tries to re-propose it again - it succeeds

linera-chain/src/manager.rs
bart-linera: I'm still confused, TBH 😅 `self.locking_block` is a block stored in the manager that we're going to confirm, isn't it? So it sounds like something that matches your description of a _locked_ block, unless I'm missing something 😅

(It sounds to me as if `self.locking_block` was something that makes the _validator_ "locked", as in: having decided on a block they can confirm.)

linera-core/src/unit_tests/client_tests.rs
afck: There is no code path currently where the regular owner would try to finish _another_ owner's proposal unless they also want to make a proposal themselves.

The step where the regular owner first makes an unsuccessful proposal is maybe unnecessary, but the other test (where we re-propose a _regular_ proposal) does it, and I thought it doesn't hurt to make sure in this case, too, that the existence of such a proposal doesn't make the scenario fail.

bart-linera (approved):

deuszx (approved):

linera-chain/src/manager.rs
afck: It's similar, but it isn't _always_ the validator's actual locked block, because there are some cases where we set it without actually signing a confirmed vote. The invariant is that it's in the same or a later round as the locked block. But in general it's advantageous for clients to learn about validated block certificates in as high a round as possible from the validators, which is why we keep them in the chain manager.

PR #3925, opened 2025-05-13
`linera-faucet`: return a whole `ChainDescription`

Twey:
## Motivation

Currently we work with the faucet by first claiming a chain, which creates that chain's description in the network, then downloading the chain description from the validators into the client storage.  This adds a roundtrip, but more importantly requires that we already have a client before getting a chain from the faucet, making it hard to work with the wallet before the client is initialized, which is necessary for e.g. initializing signers.

## Proposal

Instead, return the whole chain description directly from the faucet, and remove the `ClaimOutcome` type (and thus the `linera-faucet` crate, which currently exists just to hold it).

## Test Plan

CI.

## Release Plan

## Links


linera-client/src/client_context.rs
```diff
@@ -404,21 +404,15 @@ where
         }
     }
 
-    pub async fn assign_new_chain_to_key(
+    pub async fn chain_description(
```
afck, line 407: Let's add a comment here to clarify that this might actually _download_ the chain description blob, and without verifying that it has been published.

linera-faucet/server/src/lib.rs
```diff
@@ -145,46 +145,48 @@ where
...
         self.context.lock().await.update_wallet(&client).await?;
-        let (chain_id, certificate) = match result? {
-            ClientOutcome::Committed(result) => result,
+        let chain_id = match result? {
+            ClientOutcome::Committed((chain_id, _certificate)) => chain_id,
             ClientOutcome::WaitForTimeout(timeout) => {
                 return Err(Error::new(format!(
                     "This faucet is using a multi-owner chain and is not the leader right now. \
-                    Try again at {}",
+                      Try again at {}",
```
afck, line 169: (I always wonder about that; is that the correct way to indent multi-line string literals? :sweat_smile:)

linera-client/src/client_context.rs
deuszx, line 407: This function is exactly(?) like the one in `chain_listener.rs` 👀

Twey: Oops!  I thought I'd deleted this.  Thanks :)

linera-service/src/linera/main.rs
```diff
@@ -1251,17 +1251,17 @@ impl Runnable for Job {
                     {faucet_url}",
                 );
                 let faucet = cli_wrappers::Faucet::new(faucet_url);
-                let outcome = faucet.claim(&owner).await?;
-                println!("{}", outcome.chain_id);
-                println!("{}", outcome.certificate_hash);
-                println!("{}", owner);
+                let description = faucet.claim(&owner).await?;
+                println!("{}", description.id());
+                serde_json::to_writer(std::io::stdout(), &description)?;
```
afck, line 1256: Maybe that doesn't need to be on `stdout`, and should be a debug or info log instead?

afck (approved):
No blockers from my side! :+1: 

But it would be good if @bart-linera could also take a look. And we need to make sure that we keep track of all the places where we put a blob in storage without verifying it, so we can fix them later!

linera-client/src/client_context.rs
Twey: I don't believe that's true — I think `ensure_has_chain_description` downloads the blob safely?

linera-faucet/server/src/lib.rs
Twey: Oh this was just some combination of my editor and `cargo fmt` :sweat_smile: Reverting.

deuszx:
I can see that this PR improves our lifes by removing the `linera-faucet` but I do not see an improvement on a client side (neither in the test, nor in the CI) 👀 can you point it out to me?

linera-service/src/linera/main.rs
Twey: Oh I thought I was going to need it, but I don't really.  I'll remove it entirely.

Twey:
@deuszx Currently it doesn't, but it will improve the Web client API in a bit :)

linera-client/src/client_context.rs
afck: It currently downloads it _without_ verifying.

We don't need to add a comment here if @bart-linera will address that _inside_ `ensure_has_chain_description`, or if he'll revisit all the call sites anyway. Let's just make sure we don't forget about this!

Twey: Verifying here means looking for a block that published the blob?

afck: Yes, exactly.

bart-linera: I'm pretty sure I will address the issue within `ensure_has_chain_description`, but I'd opt for taking a note of it, anyway (could be a `TODO (#2351)`).

Twey: So the only new callsite is in `linera-faucet-server`, which will always hit the local path.  But I'll add a comment to the existing callsite just in case.

PR #3929, opened 2025-05-13
Make `epoch` not optional in `ChainInfo` and state.

afck:
## Motivation

Every initialized chain has a current epoch, and we are not using the `epoch` field to check whether a chain is initialized.

## Proposal

Make `epoch` not optional.
Also deduplicate some related chain client code.

## Test Plan

CI

## Release Plan

## Links


ma2bd (approved):
thanks!

PR #3930, opened 2025-05-13
Remove unused code.

afck:
## Motivation

Using [cargo-workspace-unused-pub](https://crates.io/crates/cargo-workspace-unused-pub) I found some unused code. Of course it is often legitimate to have public code that is not used within the workspace, but I believe most of the items it found are actually not needed.

## Proposal

Remove some unused code.

## Test Plan

CI—and a careful review so that I don't remove anything that is meant for other users of the crates.

## Release Plan

## Links

### DIFF
--- a/linera-core/src/local_node.rs
+++ b/linera-core/src/local_node.rs
@@ -15,7 +15,7 @@ use linera_base::{
 };
 use linera_chain::{
     data_types::{BlockProposal, ProposedBlock},
-    types::{Block, ConfirmedBlockCertificate, GenericCertificate, LiteCertificate},
+    types::{Block, GenericCertificate, LiteCertificate},
     ChainStateView,
 };
 use linera_execution::{Query, QueryOutcome};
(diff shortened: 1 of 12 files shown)


ma2bd (approved):

linera-core/src/local_node.rs
```diff
@@ -294,35 +276,6 @@ where
         Ok(response)
     }
 
-    /// Obtains the certificate containing the specified message.
-    #[instrument(level = "trace", skip(self))]
-    pub async fn certificate_for_block(
```
deuszx, line 299: I can imagine this coming in handy - especially when improving the performance of syncing parent chain. We want to avoid downloading everything so I was thinking that we would download certificates for "latest" block and go from there.

afck: That's true. We can always bring it back if needed. But I'm not a fan of keeping around unused code in case we need it.

PR #3937, opened 2025-05-14
Enable access to the Ethereum Logs

### LINKED ISSUE #3758: Extracting Ethereum events from Linera events
We are redirecting the Ethereum Log to the Linera events.
However, we are not yet extracting them from the log which needs to be done.

MathieuDutSik:
## Motivation

The Ethereum logs are an essential feature of the EVM, and we need to support it. Since the events are available as blobs it is relatively easy to support them.

Fixes #3758 

## Proposal

The following is done:
* The function `fn read_events_from_index` is added to the storage.
* Since the way the serialization is done influences the function, adequate tests are added.
* The required traits are added to the `StreamId`, `GenericApplicationId`, `StreamName`, and the needed tests are added so that `FromStr` is the reverse of `Display`.
* The `IndexAndEvent` type is added, as this is how the events appear from downloading them.
* GraphQL types are implemented via `SimpleObject` and `InputObject` with `input_name` being used for consistency..
* The function `events_from_index` is added to the `NodeService`.
* The truncation length of 200 was not enough for debugging and was extended to 500. It might be a good value overall, as 500 characters will not flood the display.
* Several `.clone()` on storage that were not needed have been removed.
* The block height is added to the emitted events.

## Test Plan

The test for the feature has been added.

## Release Plan

## Links

None.

### DIFF
--- a/CLI.md
+++ b/CLI.md
@@ -29,6 +29,7 @@ This document contains the help content for the `linera` command-line program.
 * [`linera service`↴](#linera-service)
 * [`linera faucet`↴](#linera-faucet)
 * [`linera publish-module`↴](#linera-publish-module)
+* [`linera list-events-from-index`↴](#linera-list-events-from-index)
 * [`linera publish-data-blob`↴](#linera-publish-data-blob)
 * [`linera read-data-blob`↴](#linera-read-data-blob)
 * [`linera create-application`↴](#linera-create-application)
@@ -92,6 +93,7 @@ A Byzantine-fault tolerant sidechain with low-latency finality and high throughp
 * `service` — Run a GraphQL service to explore and extend the chains of the wallet
 * `faucet` — Run a GraphQL service that exposes a faucet where users can claim tokens. This gives away the chain's tokens, and is mainly intended for testing
 * `publish-module` — Publish module
+* `list-events-from-index` — Reading the events from a starting index
 * `publish-data-blob` — Publish a data blob of binary data
 * `read-data-blob` — Verify that a data blob is readable
 * `create-application` — Create an application
@@ -648,6 +650,25 @@ Publish module
 * `--vm-runtime <VM_RUNTIME>` — The virtual machine runtime to use
 
   Default value: `wasm`
+
+
+
+## `linera list-events-from-index`
+
+Reading the events from a starting index
+
+**Usage:** `linera list-events-from-index [OPTIONS] --stream-id <STREAM_ID> [CHAIN_ID]`
+
+###### **Arguments:**
+
+* `<CHAIN_ID>` — The chain to query. If omitted, query the default chain of the wallet
+
+###### **Options:**
+
+* `--stream-id <STREAM_ID>` — The stream being considered
+* `--start-index <START_INDEX>` — Index of the message to start with
+
+  Default value: `0`
 
 
 
--- a/CLI.md
+++ b/CLI.md
@@ -93,7 +93,7 @@ A Byzantine-fault tolerant sidechain with low-latency finality and high throughp
 * `service` — Run a GraphQL service to explore and extend the chains of the wallet
 * `faucet` — Run a GraphQL service that exposes a faucet where users can claim tokens. This gives away the chain's tokens, and is mainly intended for testing
 * `publish-module` — Publish module
-* `list-events-from-index` — Reading the events from a starting index
+* `list-events-from-index` — Print events from a specific chain and stream from a specified index
 * `publish-data-blob` — Publish a data blob of binary data
 * `read-data-blob` — Verify that a data blob is readable
 * `create-application` — Create an application
@@ -655,7 +655,7 @@ Publish module
 
 ## `linera list-events-from-index`
 
-Reading the events from a starting index
+Print events from a specific chain and stream from a specified index
 
 **Usage:** `linera list-events-from-index [OPTIONS] --stream-id <STREAM_ID> [CHAIN_ID]`
 
--- a/linera-base/src/identifiers.rs
+++ b/linera-base/src/identifiers.rs
@@ -497,7 +497,7 @@ impl StreamId {
     }
 }
 
-/// The result of an events from index.
+/// The result of an `events_from_index`.
 #[derive(
     Debug,
     Eq,
@@ -514,7 +514,7 @@ impl StreamId {
     SimpleObject,
 )]
 pub struct IndexAndEvent {
-    /// The index of the found index.
+    /// The index of the found event.
     pub index: u32,
     /// The event being returned.
     pub event: Vec<u8>,
--- a/linera-base/src/identifiers.rs
+++ b/linera-base/src/identifiers.rs
@@ -4,7 +4,7 @@
 //! Core identifiers used by the Linera protocol.
 
 use std::{
-    fmt::self,
+    fmt,
     hash::{Hash, Hasher},
     marker::PhantomData,
 };
@@ -331,7 +331,7 @@ impl fmt::Display for GenericApplicationId {
             GenericApplicationId::User(application_id) => {
                 Display::fmt("User", f)?;
                 Display::fmt(&application_id, f)
-            },
+            }
         }
     }
 }
@@ -520,7 +520,6 @@ pub struct IndexAndEvent {
 
 scalar!(IndexAndEvent);
 
-
 impl fmt::Display for StreamId {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         Display::fmt(&self.application_id, f)?;
@@ -535,9 +534,13 @@ impl std::str::FromStr for StreamId {
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         let parts = s.split(':').collect::<Vec<_>>();
         if parts.len() == 2 {
-            let application_id = GenericApplicationId::from_str(parts[0]).context("Invalid GenericApplicationId!")?;
+            let application_id = GenericApplicationId::from_str(parts[0])
+                .context("Invalid GenericApplicationId!")?;
             let stream_name = StreamName::from_str(parts[1]).context("Invalid StreamName!")?;
-            Ok(StreamId { application_id, stream_name })
+            Ok(StreamId {
+                application_id,
+                stream_name,
+            })
         } else {
             Err(anyhow!("Invalid blob ID: {}", s))
         }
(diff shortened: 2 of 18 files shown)


linera-service/src/linera/main.rs
```diff
@@ -1039,12 +1039,32 @@ impl Runnable for Job {
...
+                    options.inner.clone(),
+                    wallet,
+                    Box::new(signer.into_value()),
+                );
+                let start_time = Instant::now();
+                let chain_id = chain_id.unwrap_or_else(|| context.default_chain());
+                let index_events = storage
+                    .list_events_from_index(&chain_id, &stream_id, start_index)
+                    .await?;
+                println!("{:?}", index_events);
```
afck, line 1058: We could pretty print this:

```suggestion
                println!("{:#?}", index_events);
```

…or even output it as a table with one event per line?

MathieuDutSik: Ok.

linera-storage/src/db_storage.rs
```diff
@@ -840,6 +880,36 @@ where
         Ok(exists)
     }
 
+    async fn list_events_from_index(
```
afck, line 883: Maybe `read_` instead of `list_`? (For `list` I would expect it to return only the indices, not the events, I think.)

linera-storage/src/db_storage.rs
```diff
@@ -840,6 +880,36 @@ where
...
+        chain_id: &ChainId,
+        stream_id: &StreamId,
+        start_index: u32,
+    ) -> Result<Vec<IndexAndEvent>, ViewError> {
+        let mut prefix = vec![INDEX_EVENT_ID];
+        prefix.extend(bcs::to_bytes(chain_id).unwrap());
+        prefix.extend(bcs::to_bytes(stream_id).unwrap());
+        let mut keys = Vec::new();
+        let mut indices = Vec::new();
+        for short_key in self.store.find_keys_by_prefix(&prefix).await?.iterator() {
```
afck, line 894: Why not `find_key_values_by_prefix` right away?

linera-storage/src/db_storage.rs
```diff
@@ -328,22 +328,31 @@ enum BaseKey {
...
+        identifiers::{
+            ApplicationId, BlobId, BlobType, ChainId, EventId, GenericApplicationId, StreamId,
+            StreamName,
+        },
     };
 
     use crate::db_storage::{
-        BaseKey, BLOB_ID_LENGTH, CHAIN_ID_LENGTH, INDEX_BLOB_ID, INDEX_CHAIN_ID,
+        BaseKey, BLOB_ID_LENGTH, CHAIN_ID_LENGTH, INDEX_BLOB_ID, INDEX_CHAIN_ID, INDEX_EVENT_ID,
     };
```
afck, line 347:
```suggestion
    };

```

linera-storage/src/db_storage.rs
```diff
@@ -328,22 +328,31 @@ enum BaseKey {
...
+    // Several functionalities of the storage rely on the way that the serialization
+    // is done. Thus we need to check that the serialization work in the way that
```
afck, line 349:
```suggestion
    // is done. Thus we need to check that the serialization works in the way that
```

linera-service/tests/linera_net_tests.rs
```diff
@@ -418,6 +418,123 @@ async fn test_evm_end_to_end_counter(config: impl LineraNetConfig) -> Result<()>
...
+        application_id,
+        stream_name,
+    };
+
+    let mut start_index = 0;
+    let indices_and_events = node_service
+        .events_from_index(&chain, &stream_id, start_index)
+        .await?;
+    let index_and_event = indices_and_events[0].clone();
+    tracing::info!("1: index_and_event={index_and_event:?}");
```
afck, line 493: (What does the `1:` mean?)

linera-service/src/cli_wrappers/wallet.rs
```diff
@@ -1204,6 +1206,19 @@ impl NodeService {
         Ok(serde_json::from_value(committees)?)
     }
 
+    pub async fn events_from_index(
+        &self,
+        chain_id: &ChainId,
+        stream_id: &StreamId,
+        start_index: u32,
+    ) -> Result<Vec<IndexAndEvent>> {
+        let query = format!("query {{ eventsFromIndex(chainId: \"{chain_id}\", streamId: {}, startIndex: {start_index}) {{ index event }} }}", stream_id.to_value());
```
afck, line 1215: (Maybe break the string into multiple lines and format it a bit.)

linera-base/src/identifiers.rs
```diff
@@ -324,6 +324,33 @@ pub enum GenericApplicationId {
     User(ApplicationId),
 }
 
+impl fmt::Display for GenericApplicationId {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            GenericApplicationId::System => Display::fmt("System", f),
+            GenericApplicationId::User(application_id) => {
+                Display::fmt("User", f)?;
```
afck, line 332: Should we add a separator (colon?) after `System` and `User`? Looks weird if the word is not separated from the hex string at all.

linera-base/src/identifiers.rs
```diff
@@ -453,6 +497,56 @@ impl StreamId {
     }
 }
 
+/// The result of an events from index.
```
afck, line 500: If you refer to a function, let's use backticks and the correct function name: `` `events_from_index` ``

linera-base/src/identifiers.rs
```diff
@@ -453,6 +497,56 @@ impl StreamId {
...
+    Hash,
+    Serialize,
+    Deserialize,
+    WitLoad,
+    WitStore,
+    WitType,
+    SimpleObject,
+)]
+pub struct IndexAndEvent {
+    /// The index of the found index.
```
afck, line 517:
```suggestion
    /// The index of the found event.
```

linera-base/src/identifiers.rs
```diff
@@ -453,6 +497,56 @@ impl StreamId {
...
+        Display::fmt(":", f)?;
+        Display::fmt(&self.stream_name, f)
+    }
+}
+
+impl std::str::FromStr for StreamId {
+    type Err = anyhow::Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let parts = s.split(':').collect::<Vec<_>>();
```
afck, line 535: I see; I guess if we use the colon in the generic application ID as well, we have to use `rsplit_once` here.

CLI.md
```diff
@@ -652,6 +654,25 @@ Publish module
 
 
 
+## `linera list-events-from-index`
+
+Reading the events from a starting index
```
deuszx, line 659: This description isn't helpful as I have no idea what will happen when I run this command.

Should it say "prints events for <chain_id>, using "some" format. Optionally caller can specify a stream/starting index" ?


Also, what will happen when there are thousands of such events? Why do we need this command in the CLI?

linera-core/src/client/mod.rs
```diff
@@ -3352,6 +3352,19 @@ impl<Env: Environment> ChainClient<Env> {
         }
     }
 
+    /// Returns the indices and events from the storage
+    pub async fn events_from_index(
```
deuszx, line 3356: This can OOM, right? There's no backpressure, paging etc

linera-execution/src/evm/revm.rs
```diff
@@ -782,10 +782,11 @@ where
         // TODO(#3758): Extracting Ethereum events from the Linera events.
         if !logs.is_empty() {
             let mut runtime = self.db.runtime.lock().expect("The lock should be possible");
+            let block_height = runtime.block_height()?;
             let stream_name = bcs::to_bytes("ethereum_event")?;
             let stream_name = StreamName(stream_name);
             for log in &logs {
-                let value = bcs::to_bytes(&(origin, log))?;
+                let value = bcs::to_bytes(&(origin, block_height.0, log))?;
```
deuszx, line 789: Just a note: if EVM events won't be ABI-encoded (or more practically - if they can't be deserialized/decoded using typical EVM tools) it will be make integrations much more difficult.

deuszx (approved):

linera-storage/src/db_storage.rs
MathieuDutSik: Ok.

linera-storage/src/db_storage.rs
MathieuDutSik: Ok.

linera-service/tests/linera_net_tests.rs
MathieuDutSik: Oversight, should have been removed.
For the record, it correspond to the first calls to `events_from_index`.

linera-base/src/identifiers.rs
MathieuDutSik: Thanks.

linera-core/src/client/mod.rs
MathieuDutSik: Yes, this can OOM. There are ways to get GraphQL pagination, but we have not implemented that thus far.

linera-service/src/cli_wrappers/wallet.rs
MathieuDutSik: Ok, done.

CLI.md
MathieuDutSik: Ok, it has been renamed.
This is needed for the node service.

linera-base/src/identifiers.rs
MathieuDutSik: Ok, done. Thanks.

linera-base/src/identifiers.rs
```diff
@@ -201,14 +201,14 @@ impl BlobId {
...
-impl Display for BlobId {
+impl fmt::Display for BlobId {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(f, "{}:{}", self.blob_type, self.hash)?;
         Ok(())
     }
 }
 
-impl FromStr for BlobId {
+impl std::str::FromStr for BlobId {
```
ma2bd, line 211: What was the problem with `FromStr` ?

linera-service/src/linera/command.rs
```diff
@@ -658,6 +658,20 @@ pub enum ClientCommand {
         publisher: Option<ChainId>,
     },
 
+    /// Print events from a specific chain and stream from a specified index.
+    ListEventsFromIndex {
```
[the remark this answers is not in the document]
ma2bd: Why is there a `start` and not an `end` index?

linera-service-graphql-client/gql/service_schema.graphql
```diff
@@ -578,6 +578,20 @@ type IncomingBundle {
...
+"""
+The result of an `events_from_index`.
+"""
+type IndexAndEvent {
+	"""
+	The index of the found event.
+	"""
+	index: Int!
+	"""
+	The event being returned.
```
ma2bd, line 590: Are those bytes?

Issue #3936, opened 2025-05-13
Use one small sender queue per pair of (sender, receiver) chains


ma2bd:
Instead of a single sender queue for all receivers, we should have one pending message per receiver. When a new message is scheduled, it should replace any previously pending message.

This should work because scheduled messages for a given receiver contain all the previous unconfirmed requests.

PR #3938, opened 2025-05-14
Create one task for each sender-request-type triple of cross-chain requests.

afck:
## Motivation

Currently a new task is spawned for every outgoing cross-chain request. For failing requests, these tasks keep retrying for a long time, occupying one spot that counts towards the task limit. If more messages to the same destination are created, these are likely to fail in the same way, and create _additional_ tasks.

## Proposal

Create at most one task for each sender-request-type triple (where the type is either "update" or "confirmation"). Later messages with the same triple supersede the earlier ones, but are handled by the same task.

Remove the task limit.

## Test Plan

CI

## Release Plan

## Links

- Closes https://github.com/linera-io/linera-protocol/issues/3936.


linera-rpc/src/grpc/server.rs
```diff
@@ -399,75 +400,133 @@ where
...
-
-        receiver
-            .for_each_concurrent(max_concurrent_tasks, |(cross_chain_request, shard_id)| {
-                let shard = network.shard(shard_id);
-                let remote_address = shard.http_address();
-
-                let pool = pool.clone();
-                let nickname = nickname.clone();
-
+        let mut join_set = JoinSet::new();
```
Twey, line 406: I don't think this affects correctness, so just a note, but I think that `tokio::task::JoinSet` won't work on the Web.  We have an abstraction `linera_core::join_set_ext::JoinSet` for this purpose.

That said this codepath is just for the server, so I'm not sure it matters.  Maybe a good practice?

afck: Ah, right, I hadn't even realized I was using the tokio one!

linera-rpc/src/grpc/server.rs
```diff
@@ -399,75 +401,133 @@ where
...
+        // An entry means that a task is running for the given sender, recipient and request kind.
+        // If it is Some, that means the task is still working on another request, and will pick
+        // up the new one when it is done. Since later requests always supersede earlier ones,
+        // we can always overwrite the entry if a new one comes in.
+        let cross_chain_tasks = Arc::new(Mutex::new(HashMap::new()));
+        let semaphore = Arc::new(tokio::sync::Semaphore::new(
+            cross_chain_max_concurrent_tasks,
```
ma2bd, line 414: Maybe we don't need `cross_chain_max_concurrent_tasks` any more

linera-rpc/src/grpc/server.rs
```diff
@@ -399,75 +400,133 @@ where
...
-            .for_each_concurrent(max_concurrent_tasks, |(cross_chain_request, shard_id)| {
-                let shard = network.shard(shard_id);
-                let remote_address = shard.http_address();
-
-                let pool = pool.clone();
-                let nickname = nickname.clone();
-
+        let mut join_set = JoinSet::new();
+        // An entry means that a task is running for the given sender, recipient and request kind.
+        // If it is Some, that means the task is still working on another request, and will pick
```
Twey, line 409: As a rule of thumb, if I have to use comments to attach meaning to `Option` I usually take it as an indication that I should create a custom enum for the task.

linera-rpc/src/grpc/server.rs
```diff
@@ -399,75 +400,133 @@ where
...
-                                );
+                    let result = || async {
+                        let cross_chain_request = cross_chain_request.clone().try_into()?;
+                        let request = Request::new(cross_chain_request);
+                        let mut client =
+                            ValidatorWorkerClient::new(pool.channel(remote_address.clone())?)
+                                .max_encoding_message_size(GRPC_MAX_MESSAGE_SIZE)
+                                .max_decoding_message_size(GRPC_MAX_MESSAGE_SIZE);
+                        let response = client.handle_cross_chain_request(request).await?;
+                        Ok::<_, anyhow::Error>(response)
```
Twey, line 468: Nit: `anyhow::Result::Ok(response)` avoids the turbofish :)

linera-rpc/src/grpc/server.rs
afck: I'm not sure: Since receiver chains don't need to exist, users can create an unlimited number of tasks that way.

On the other hand that's kind of already the case, since we don't count sleeping tasks towards the limit.

linera-rpc/src/grpc/server.rs
afck: Nice, good idea! (I didn't add this line, though; only moved it.)

linera-rpc/src/grpc/server.rs
```diff
@@ -399,75 +401,133 @@ where
...
+                                %error,
+                                i,
+                                from_shard = this_shard,
+                                to_shard = shard_id,
+                                "Failed to send cross-chain query",
+                            );
+                            i += 1;
+
+                            let mut guard = cross_chain_tasks.lock().unwrap();
+                            match guard.entry(cross_chain_request.sender_recipient_update()) {
```
Twey, line 483: Nit: since you don't bind the vacant case you can just use `if let Some(entry) = guard.get_mut(…)`.

linera-rpc/src/grpc/server.rs
```diff
@@ -399,75 +401,133 @@ where
...
+                        let cross_chain_request = cross_chain_request.clone().try_into()?;
+                        let request = Request::new(cross_chain_request);
+                        let mut client =
+                            ValidatorWorkerClient::new(pool.channel(remote_address.clone())?)
+                                .max_encoding_message_size(GRPC_MAX_MESSAGE_SIZE)
+                                .max_decoding_message_size(GRPC_MAX_MESSAGE_SIZE);
+                        let response = client.handle_cross_chain_request(request).await?;
+                        Ok::<_, anyhow::Error>(response)
+                    };
+                    match result().await {
```
Twey, line 470: I think if you await `result` above then swap the order of the matches here you can avoid some duplication?

linera-rpc/src/grpc/server.rs
afck: Surprisingly, that doesn't work, and it still says:
```
468 |                         anyhow::Result::Ok(response)
    |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ cannot infer type for type parameter `E` declared on the enum `Result`
```

Twey: Oh I didn't realize they'd done that!  There's an optional type parameter for the error.  Yes you'll have to `anyhow::Result::<_>::Ok`.  No escaping the turbofish I suppose.

linera-rpc/src/grpc/server.rs
afck: Do you have an idea for names?
Something like:

```rust
enum CrossChainTaskQueue {
    Empty,
    Full(CrossChainRequest),
}
```

afck: (Not really a "queue", if it's only one task…)

linera-rpc/src/grpc/server.rs
ma2bd: Let's keep discussing this

linera-rpc/src/grpc/server.rs
Twey: Like this: https://github.com/Twey/linera-protocol/commit/379221680027e6d82cb227d9100e0df2d58afe1d

afck: Thanks! Feel free to push!

linera-rpc/src/grpc/server.rs
Twey: Maybe it's just a ‘pending task’?

Alternatively, since empty is never a valid state for the task, we could remove this option and use the empty state for what's currently `None`?

linera-rpc/src/grpc/server.rs
```diff
@@ -397,77 +398,113 @@ where
...
+                    return;
+                }
 
-        receiver
-            .for_each_concurrent(max_concurrent_tasks, |(cross_chain_request, shard_id)| {
-                let shard = network.shard(shard_id);
-                let remote_address = shard.http_address();
+                let mut i = 0;
+                while i < cross_chain_max_retries {
+                    linera_base::time::timer::sleep(cross_chain_sender_delay).await;
```
Twey, line 444: So we always delay, even on the first attempt?

linera-rpc/src/grpc/server.rs
```diff
@@ -397,77 +398,113 @@ where
...
+            let nickname = nickname.clone();
+            let cross_chain_tasks = cross_chain_tasks.clone();
+            join_set.spawn(async move {
+                // Send the cross-chain query and retry if needed.
+                if cross_chain_sender_failure_rate > 0.0
+                    && rand::thread_rng().gen::<f32>() < cross_chain_sender_failure_rate
+                {
+                    warn!("Dropped 1 cross-chain message intentionally.");
+                    return;
+                }
```
Twey, line 440: We should be careful to drop the entry here or no new messages will ever be delivered through this channel.

afck: Done in https://github.com/linera-io/linera-protocol/pull/3938/commits/9985c0bc8a2b377e75f7d7c44b03d4085627bf3a.

linera-core/src/data_types.rs
```diff
@@ -237,6 +237,10 @@ pub enum CrossChainRequest {
     },
 }
 
+/// An discriminant for message queues: messages with the same queue ID will be delivered
+/// in order.
+pub type CrossChainQueueId = (ChainId, ChainId, bool);
```
ma2bd, line 242: `pub type` aliases are a bit meh, so are tuples. Can we have a struct? Does it need to be public?

linera-rpc/src/grpc/server.rs
```diff
@@ -397,77 +406,148 @@ where
...
+        enum Action {
+            Proceed { id: usize },
+            Retry,
+        }
+
+        #[derive(Clone)]
+        struct Task {
+            shard_id: ShardId,
+            request: linera_core::data_types::CrossChainRequest,
+        }
```
ma2bd, line 423: nit: Any reason to have this in the method?

linera-rpc/src/grpc/server.rs
```diff
@@ -397,77 +406,148 @@ where
...
+                pool.channel(network.shard(task.shard_id).http_address())?,
+            )
+            .max_encoding_message_size(GRPC_MAX_MESSAGE_SIZE)
+            .max_decoding_message_size(GRPC_MAX_MESSAGE_SIZE);
+            let response = client.handle_cross_chain_request(request).await?;
+            anyhow::Result::<_>::Ok(response)
+        };
+
+        #[derive(Clone)]
+        struct Job {
```
ma2bd, line 442: same

ma2bd (commented):
It would be nice to do the same for simple network

afck:
> It would be nice to do the same for simple network

I'd prefer to do that in a separate PR, after finishing the gRPC one; that way I don't have to address each review comment twice.

linera-rpc/src/simple/server.rs
```diff
@@ -83,62 +85,38 @@ where
...
-                    }
-                    _ => {
-                        debug!(
-                            from_shard = this_shard,
-                            to_shard = shard_id,
-                            "Sent cross-chain query",
-                        );
-                        continue 'receive_loop;
-                    }
+        let pool = Arc::new(Mutex::new(
```
afck, line 90: Alternatively we could make the pool clonable:
* In the UDP case that would simply mean putting the mutex _inside_ the pool struct.
* In the TCP case we could use a `DashMap`, just like we do for gRPC.

linera-rpc/src/cross_chain_message_queue.rs
```diff
@@ -0,0 +1,199 @@
...
+#[derive(Copy, Clone, PartialEq, Eq, Hash)]
+struct QueueId {
+    sender: ChainId,
+    recipient: ChainId,
+    is_update: bool,
+}
+
+impl QueueId {
+    /// Returns a discriminant for the message's queue.
+    fn new(request: &CrossChainRequest) -> Self {
```
deuszx, line 165: nit: Personally I think `from` would be more appropriate.

linera-rpc/src/cross_chain_message_queue.rs
```diff
@@ -0,0 +1,199 @@
...
+                match job_states.entry(queue) {
+                    Entry::Vacant(entry) => futures.push(run_action.clone()(
+                        Action::Proceed { id: 0 },
+                        queue,
+                        entry.insert(Job {
+                            id: 0,
+                            retries: 0,
+                            nickname: nickname.clone(),
+                            task,
+                        }).clone(),
```
deuszx, line 135: nit: I think it would be cleaner if you first insert the new job into the job state and then pass the clone to `run_action`.

linera-rpc/src/cross_chain_message_queue.rs
```diff
@@ -0,0 +1,199 @@
...
+
+#[derive(Clone)]
+struct Task {
+    pub shard_id: ShardId,
+    pub request: linera_core::data_types::CrossChainRequest,
+}
+
+#[derive(Clone)]
+struct Job {
+    pub id: usize,
```
deuszx, line 195: It'd be nice to have some comment(s) here. It took me a while to understand how these ids are assigned:
- if there are no jobs for this tasks's queue - it's `0`
- otherwise - increment last job's `id`. 
 
Meaning that `job.id` is basically a tracker how many tasks have arrived for this queue while it was busy handling previous request.

linera-rpc/src/cross_chain_message_queue.rs
```diff
@@ -0,0 +1,199 @@
...
+    };
+
+    loop {
+        tokio::select! {
+            Some((queue, action)) = futures.next() => {
+                let Entry::Occupied(mut state) = job_states.entry(queue) else {
+                    panic!("running job without state");
+                };
+
+                let remove = matches!(action, Action::Proceed { id } if state.get().id < id)
```
deuszx, line 95: TBH it wasn't easy - at least for me - to decipher why this particular `matches!` means we can remove the job from the queue. IIUC we remove it when this task's is the last one to arrive on this queue and we successfuly handled it. Maybe add a function on `state` (which is a `Job` 🙃 ) which would make it clear (even from the function name) why this queue was processed successfully.

linera-rpc/src/cross_chain_message_queue.rs
```diff
@@ -0,0 +1,199 @@
...
+                    state.get_mut().retries += 1
+                }
+
+                futures.push(run_action.clone()(action, queue, state.get().clone()));
+            }
+
+            request = receiver.next() => {
+                let Some((request, shard_id)) = request else { break };
+                #[cfg(with_metrics)]
+                CROSS_CHAIN_MESSAGE_TASKS.set(job_states.len() as i64);
```
deuszx, line 112: Why not track `futures.len()` instead? It's the one that tracks how many cross-chain messages are in-flight.

linera-rpc/src/simple/server.rs
Twey: I'm not fussed either way.  In both cases the changes are limited to `linera_rpc::simple` and have no wider impact.  Probably as a general rule I prefer to leave the mutex outside an opaque struct so that consumers can choose to opt in/out.

linera-rpc/src/cross_chain_message_queue.rs
```diff
@@ -0,0 +1,199 @@
...
+    cross_chain_sender_delay: Duration,
+    cross_chain_sender_failure_rate: f32,
+    this_shard: ShardId,
+    mut receiver: mpsc::Receiver<(CrossChainRequest, ShardId)>,
+    handle_request: F,
+) where
+    F: Fn(ShardId, CrossChainRequest) -> G + Send + Clone + 'static,
+    G: Future<Output = anyhow::Result<()>>,
+{
+    let mut futures = futures::stream::FuturesUnordered::new();
```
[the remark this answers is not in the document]
Twey: It's all the futures run by the function :)  Specifically, currently (but not necessarily) both the timeout futures and the in-flight requests.

linera-rpc/src/cross_chain_message_queue.rs
Twey: For me I think I usually expect `from` to be an injection.

linera-rpc/src/cross_chain_message_queue.rs
```diff
@@ -0,0 +1,199 @@
...
+    let mut job_states: HashMap<QueueId, Job> = HashMap::new();
+
+    let run_task = |task: Task| async move { handle_request(task.shard_id, task.request).await };
+
+    let run_action = |action, queue, state: Job| async move {
```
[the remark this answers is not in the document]
Twey: It's the state of the job.  But your complaint is valid, the connection isn't necessarily clear.  Probably the correct solution is to rename `Job` → `JobState`, which is a distinction I don't usually like to make (_ceci n'est pas une pipe_) but maybe relevant here.

linera-rpc/src/cross_chain_message_queue.rs
Twey: I don't have strong feelings about it, but I'm curious why you think so.  I usually introduce a variable only if it adds something:
- the name will be used more than once,
- the variable needs a type annotation (sometimes the turbofish suffices for this, but because the turbofish gives input types rather than outputs it's often clearer to give the output — which is what the rest of the program cares about — and let the compiler do the work of figuring out the inputs), or
- the expression that gives rise to the value doesn't make it clear what the result is, so I can add some documentation in the form of the variable name.

Neither is true here, so (IMO) adding a new variable just increases the number of names the reader has to keep in mind without giving anything back for their money.

linera-rpc/src/cross_chain_message_queue.rs
Twey: I think that's maybe a bit too low-level for a comment — comments usually want to be about the ‘why’ not the ‘what’ (which is written in the code itself).  I think the important insight here is that the IDs are monotonically increasing so long as there's a job in the queue.

I was tempted to name it ‘height’, which is a concept readers should be comfortable with from the rest of the codebase — do you think that would have been clearer on your first reading?

linera-rpc/src/cross_chain_message_queue.rs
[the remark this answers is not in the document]
Twey: Maybe `steps` is a better name?  Since it's specifically functions that will yield some next action to take.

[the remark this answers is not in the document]
deuszx: But it's not _any_ future. It's not even _any_ timeout b/c it's always per-queue future.

linera-rpc/src/cross_chain_message_queue.rs
deuszx: Here it's literally mapping from `CrossChainRequest` to a `QueueId`.

linera-rpc/src/cross_chain_message_queue.rs
deuszx: In that case we should update the metric when the job is successfully removed from the `job_states`?

linera-rpc/src/cross_chain_message_queue.rs
deuszx: Yeah, I think `height` will be clearer.

linera-rpc/src/cross_chain_message_queue.rs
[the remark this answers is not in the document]
Twey: That's fair, maybe `jobs` is a good name? (apropos the discussion about `JobState` above)

linera-rpc/src/cross_chain_message_queue.rs
Twey: Right, which is definitely not an injection :)

linera-rpc/src/cross_chain_message_queue.rs
Twey: That seems sensible, we could just move this outside the loop?

linera-rpc/src/simple/server.rs
afck: It's just that with the `DashMap`, multiple requests could be handled at the same time. Keeping the mutex outside makes that impossible.

linera-rpc/src/cross_chain_message_queue.rs
deuszx: That's why I didn't suggest implementing `From`/`Into` but rename `new` into `from` which makes it clear that `QueueId` is constructed _from_ a `CrossChainRequest`.

deuszx: Not gonna insist though :)

linera-rpc/src/simple/server.rs
afck: Seems complicated, though; I think we'd even need mutexes in the `DashMap`'s entries. I'll leave that for another PR, if we decide to do it.

linera-rpc/src/cross_chain_message_queue.rs
afck: Not sure, I'd prefer to avoid overloading the term "height".

PR #3941, opened 2025-05-14
Move committee getting logic into extension trait.

afck:
## Motivation

The `client` module contains a lot of logic to get committee information out of a `ChainInfo` from the local node.

## Proposal

Move that logic into an extension trait in `local_node`.

Also rename `chain_info` to `fetch_chain_info`, because it can download it from the validators.

## Test Plan

CI

## Release Plan

## Links

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -1888,12 +1888,15 @@ impl<Env: Environment> ChainClient<Env> {
         #[cfg(with_metrics)]
         let _latency = metrics::SYNCHRONIZE_CHAIN_STATE_LATENCY.measure_latency();
 
-        let info = self.client.chain_info_with_committees(chain_id).await?;
-        let committee = info.current_committee()?;
-        let validators = self.client.make_nodes(committee)?;
+        let committee = self
+            .client
+            .chain_info_with_committees(chain_id)
+            .await?
+            .into_current_committee()?;
+        let validators = self.client.make_nodes(&committee)?;
         communicate_with_quorum(
             &validators,
-            committee,
+            &committee,
             |_: &()| (),
             |remote_node| {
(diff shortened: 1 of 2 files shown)


linera-core/src/client/mod.rs
```diff
@@ -1861,18 +1854,8 @@ impl<Env: Environment> ChainClient<Env> {
...
-            .handle_chain_info_query(query)
-            .await?
-            .info;
-        let committee = info
-            .requested_committees
-            .ok_or(LocalNodeError::InvalidChainInfoResponse)?
-            .remove(&info.epoch)
-            .ok_or(LocalNodeError::InactiveChain(chain_id))?;
+        let info = self.chain_info_with_committees().await?;
+        let committee = info.current_committee()?;
```
deuszx, line 1858: I think I'd just chain `current_comittee()` call after `.await?`- w/o the intermediate `info` 🤔 

Here and below.

deuszx (approved):

linera-core/src/client/mod.rs
deuszx: I just think if `info` not used anywhere, these two are now so short and self-describing I wouldn't add more variables.

afck: I'll change the other call site where it's possible, but at this one here it isn't: We're using the info in the next two lines, too.

linera-core/src/local_node.rs
```diff
@@ -329,3 +329,38 @@ where
         Ok(())
     }
 }
+
+/// Extension trait for [`ChainInfo`]s from our local node. These should always be valid and
+/// contain the requested information.
+pub trait LocalChainInfoExt {
```
ma2bd, line 335: nice

Issue #3942, opened 2025-05-14
Move logic from `ChainClient` to `Client`.


afck:
`Client` already owns the local node, and also manages the chain client state map for _all_ chains.

`ChainClient` should just be a wrapper for convenience, to avoid repeatedly passing in the chain ID as an argument.

PR #3943, opened 2025-05-14
Move some `ChainClient` functions to `Client`.

afck:
## Motivation

We want to move more logic to `Client` (see #3942). Specifically `update_from_info` currently only updates the chain client state correctly if the chain info happens to be the one for the `ChainClient`'s own chain. It should _always_ update the chain client state, however.

## Proposal

Move `update_from_info` to `Client`.
(Also `process_certificate`. More to follow.)

## Test Plan

CI

## Release Plan

## Links

- Part of #3942.

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -496,6 +496,25 @@ impl<Env: Environment> Client<Env> {
         self.local_node.storage_client().write_blob(&blob).await?;
         Ok(blob)
     }
+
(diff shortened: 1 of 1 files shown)


MathieuDutSik (approved):
The change seems very sensible.

deuszx (approved):

bart-linera (approved):
I might be moving over even more stuff in the future, unless you beat me to it 😅

PR #3946, opened 2025-05-15
Move admin_id from ChainDescription into NetworkDescription

bart-linera:
## Motivation

There is no use case for different chains having different admin chain IDs, so admin ID should be a network-wide value, not a per-chain value.

## Proposal

Move the admin chain ID from `ChainDescription` into `NetworkDescription`.

## Test Plan

## Release Plan

## Links


afck (approved):

afck:
I fixed the tests; closing this in favor of #3961.

Issue #285, opened 2023-01-06
Protection against long-range attacks


ma2bd:

afck:
We will introduce (as part of the genesis config) a minimum delay `d` between the introduction of a new committee and the withdrawal of the stake (e.g. an epoch could last one day, and that duration could be two weeks).
So all certificates signed by a committee that is no older than `d` are still backed by validator stake.

And if a certificate has any descendants or message recipients that can be verified that way, i.e. they have been re-certified, they also can trusted indirectly.

So we should allow uploading a vector of certificates to a validator in a request. The validator checks that:
* the certificates form a chain, i.e. link to each other by parent hash,
* the highest certificate is signed by a committee no older than `d`,
* and all certificates are valid,

and then executes them. This may require several requests for very long chains; if the client starts supplying the highest certificates first, the validator knows that they are legitimate.

Finally, we need to rethink the acceptance criterion for cross-chain messages:
* A block with an incoming message should be accepted if a quorum of validators can verify that the message was actually sent, i.e. verified the sending block. This does not depend on which epoch the receiving chain is in.
* Clients—especially when making fast blocks—will want to make absolutely sure that the validators will still accept the received message, so they should only include messages which are sent by a block signed by a committee no older than e.g. `d / 2`, so have a sufficient margin of error.

afck:
> a quorum of validators can verify that the message was actually sent

And I'm wondering whether this needs to go all the way back to genesis, after all. If we don't require validators to have all sender chains recursively before they accept a certificate, a quorum of the current validators can conspire to make up a sender chain that never existed.

If we require that validators only sign block proposals where they have the complete set of dependencies back to genesis, at least messages cannot be created out of nothing.

(Even then, validators can conspire with past validators to extend some long-abandoned chain, but maybe that's okay. And maybe we can prevent that by automatically—using fallback validators—closing very old chains automatically.)

PR #3947, opened 2025-05-15
Use admin chain instead of `known_committees`.

afck:
## Motivation

We want to move more functionality into `Client` (https://github.com/linera-io/linera-protocol/issues/3942) and in general simplify the logic.

`known_committees` currently returns the union of all committees known to the `ChainClient`'s local chain and the admin chain. This prevents moving its call sites to `Client`, and also doesn't make sense when we move towards addressing https://github.com/linera-io/linera-protocol/issues/285: The currently trusted committees cannot depend on which `ChainClient` instance caused the synchronization.

## Proposal

In `prepare_chain` make sure that the admin chain is always at least as up-to-date as the local chain.

Replace `known_committees` with `admin_committees`.

Update `test_change_voting_rights`: It makes sense that the test would fail now if we revoked the old committee _before_ migrating the other chain!

## Test Plan

CI

## Release Plan

## Links

- Part of https://github.com/linera-io/linera-protocol/issues/3942.

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -438,23 +438,12 @@ impl<Env: Environment> Client<Env> {
 
     /// Obtains all the committees trusted by any of the given chains. Also returns the highest
     /// of their epochs.
-    // TODO(#285): This should probably return _all_ currently trusted committees, independent of
-    // specific chains.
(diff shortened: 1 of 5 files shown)


christos-h (approved):
Makes sense :).

PR #3896, opened 2025-05-06
Validator destination

usagi32:
## Motivation

Essential work to close #3666.
Also closes #3667.

## Proposal
- Caching for the recently used blocks and other things from the storage.
- Dependency resolution with a topological sort for the whole state.
- Parallelization of the exporters across threads.
- Refactoring of the binary crate.
- Validator destination.

## Test Plan
CI

## Release Plan

- Nothing to do / These changes follow the usual release cycle.

PR #3949, opened 2025-05-15
Indexer destination

### LINKED ISSUE #3669: Implement the `Indexer` destination
This is similar to `Validator` but using a smaller GRPC API

usagi32:
## Motivation

Generic indexer destination for the block exporter.
To be rebased on top of #3896 .

## Proposal

- Integration with the local net.
- Global indexing for the blobs for each destination kind.
- Also Closes #3669 

## Test Plan

Unit + integration tests

## Release Plan


linera-client/src/config.rs
```diff
@@ -243,17 +243,34 @@ pub struct Destination {
...
+    pub fn indexer_address(&self) -> String {
         let tls = match self.tls {
             TlsConfig::ClearText => "http",
             TlsConfig::Tls => "https",
         };
 
         format!("{}://{}:{}", tls, self.endpoint, self.port)
     }
+
+    pub fn validator_address(&self) -> String {
```
deuszx: Add panic!s here that will fail when we call `validator_address` on `Indexer` variant (and vice versa)

linera-client/src/config.rs
```diff
@@ -292,3 +309,14 @@ impl Default for LimitsConfig {
         }
     }
 }
+
+impl Destination {
+    pub fn address(&self) -> String {
```
deuszx, line 302: This repeats the `validator_address` and `indexer_address` methods above. It's also incorrect if destination kind is indexer - b/c that uses gRPC protocol.

linera-service/src/cli/net_up_utils.rs
```diff
@@ -183,7 +183,7 @@ pub async fn handle_net_up_service(
     faucet_chain: Option<u32>,
     faucet_port: NonZeroU16,
     faucet_amount: Amount,
-    num_block_exporters: u32,
+    _num_block_exporters: u32,
```
deuszx: Why this change? If it's really unused, why not remove it?

linera-service/src/exporter/common.rs
```diff
@@ -58,6 +72,21 @@ impl From<BadNotificationKind> for ExporterError {
     }
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub(crate) struct CanonicalBlock {
```
deuszx, line 79: Again - what is _canonical_ in this context?

linera-service/src/exporter/proto/indexer.proto
```diff
@@ -0,0 +1,46 @@
...
+        Block block = 2;
+    }
+}
+
+message CryptoHash {
+    fixed64 part1 = 1;
+    fixed64 part2 = 2;
+    fixed64 part3 = 3;
+    fixed64 part4 = 4;
+}
```
deuszx: In `rpc.proto` we already define `CryptoHash` as
```
message CryptoHash {
  bytes bytes = 1;
}
```
Try importing `rpc.proto` here and reusing those definitions.

linera-service/src/exporter/proto/indexer.proto
```diff
@@ -0,0 +1,46 @@
...
+
+enum BlobType {
```
deuszx: Do we need to be able to discriminate the blob type at the gRPC protocol level?

linera-service/tests/local_net_tests.rs
```diff
@@ -1067,16 +1071,51 @@ async fn test_linera_exporter(database: Database, network: Network) -> Result<()
...
+        .await?;
+
+    tokio::time::sleep(Duration::from_secs(4)).await;
+
+    let validator_client = net.validator_client(1).await?;
+    let chain_info = validator_client
+        .handle_chain_info_query(ChainInfoQuery::new(chain))
+        .await?;
+
+    assert!(chain_info.info.next_block_height == 1.into());
```
deuszx, line 1204: Can you modify the test such that first we verify the new validator hasn't seen any blocks, then we start an exporter pointing at it, and now it sees the block?

linera-service/src/exporter/storage.rs
```diff
@@ -396,6 +460,12 @@ impl Weighter<CryptoHash, Arc<ConfirmedBlockCertificate>> for BlockCacheWeighter
     }
 }
 
+impl Weighter<usize, CanonicalBlock> for CanonicalStateCacheWeighter {
+    fn weight(&self, _key: &usize, val: &CanonicalBlock) -> u64 {
+        (size_of::<usize>() + bcs::serialized_size(val).unwrap()) as u64
```
deuszx: Similarly to previous PR - we shouldn't be serializing the value. Let's assume some "sensible default". We have a limit for blobs per block - let's add that to the block size and make the value here static.

linera-service/src/exporter/common.rs
```diff
@@ -103,8 +136,13 @@ impl From<BlockId> for LiteBlockId {
 
 #[macro_export]
 macro_rules! dispatch {
+    ($func:expr, log_batch = $log_value:expr $(, $args:expr)* ) => {{
```
deuszx: To be honest - the way `dispatch!` is used is very unreadable and is actually making things worse. Let's replace that with a normal function.

linera-service/src/exporter/runloops/block_processor/mod.rs
```diff
@@ -485,4 +489,35 @@ mod test {
             chain.iter().map(|block| block.inner().hash()).collect(),
         )
     }
+
+    #[tokio::test]
+    async fn test_topological_sort_5() -> anyhow::Result<()> {
```
deuszx, line 496: I have the same problem here as before - what is this test testing? its name is not helping, there are no comments and at first reading I don't understand how it differs from the previous ones. 

Please:
- rename to something meaningful
- add comments
- make sure this is not re-testing what is already done.

linera-service/src/exporter/runloops/block_processor/walker.rs
```diff
@@ -54,9 +59,29 @@ where
...
+            let mut blobs_to_send = Vec::new();
+            let mut blobs_to_index_block_with = Vec::new();
+            for id in node_visitor.node.blobs {
+                if !self.is_blob_indexed(id).await? {
+                    blobs_to_index_block_with.push(id);
+                    if !node_visitor.node.native_blobs.contains(&id) {
+                        blobs_to_send.push(id);
+                    }
+                }
+            }
```
deuszx: Why did we not index blobs for "validator destination" as well?

linera-service/src/exporter/runloops/block_processor/walker.rs
```diff
@@ -112,6 +140,8 @@ impl NodeVisitor {
 #[derive(Debug)]
 struct ProcessedBlock {
     block: BlockId,
+    blobs: Vec<BlobId>,
+    native_blobs: Vec<BlobId>,
```
deuszx: What's the difference between blobs and native blobs? Change to more helpful names.

linera-service/src/exporter/runloops/mod.rs
```diff
@@ -128,18 +135,20 @@ async fn start_exporter_tasks<S, F>(
...
+            destination,
+            storage.clone().unwrap(),
+            arced_node_provider.clone(),
         );
-        let _handle = set.spawn(exporter_task.run_with_shutdown(shutdown_signal.clone()));
     }
 
     while let Some(res) = set.join_next().await {
-        res.unwrap().unwrap();
+        let _res = res.unwrap();
```
deuszx, line 151: We're possibly ignoring an `Err(_)` here. Why is that safe?

linera-service/src/exporter/runloops/validator_exporter.rs
```diff
@@ -52,7 +56,12 @@ where
         self,
         shutdown_signal: F,
     ) -> anyhow::Result<()> {
-        let address = self.destination_config.address();
+        ensure!(
+            DestinationKind::Validator == self.destination_config.kind,
+            ExporterError::DestinationError
+        );
```
deuszx, line 61: Yes, if you move this check to `fn validator_address()` it will be more useful since you won't have to think about this on the callsite.

linera-service/src/exporter/runloops/block_processor/walker.rs
```diff
@@ -139,6 +169,8 @@ impl ProcessedBlock {
         Self {
             dependencies,
             block: block_id,
+            blobs: block.required_blob_ids().into_iter().collect(),
+            native_blobs: block.block().created_blob_ids().into_iter().collect(),
```
deuszx: `required_blob_ids()` already includes `created_blob_ids()`. Why do you keep them separately and duplicate?

linera-service/src/exporter/runloops/indexer/client.rs
```diff
@@ -0,0 +1,86 @@
...
+        })
+    }
+
+    pub(super) async fn synchronize_long_lived_stream(
+        &mut self,
+        queue_size: usize,
+    ) -> Result<(Sender<Element>, Streaming<()>), ExporterError> {
+        let mut retry_count = 0;
+        let streams = loop {
+            let (stream, sink) = tokio::sync::mpsc::channel(queue_size);
```
deuszx: Please rename to sender/receiver or tx/rx rather than introducing thrid naming scheme.

linera-service/src/exporter/runloops/indexer/client.rs
```diff
@@ -0,0 +1,86 @@
...
+            let request = Request::new(ReceiverStream::new(sink));
+            match self.client.index_batch(request).await {
+                Ok(res) => break (stream, res.into_inner()),
+                Err(e) => {
+                    if retry_count > self.max_retries {
+                        return Err(ExporterError::SynchronizationFailed(e.into()));
+                    }
+                    println!("{:?}", e);
```
deuszx: Change to `error!`

linera-service/src/exporter/runloops/indexer/indexer_exporter.rs
```diff
@@ -0,0 +1,221 @@
...
+                        tracing::error!("unexpected error: {e}, re-trying to establish a stream");
+                        sleep(Duration::from_secs(1)).await;
+                    }
+                },
+
+                res = acknowledgement_task.run() => {
+                    if let Err(e) = res {
+                        tracing::error!("unexpected error: {e}, re-trying to establish a stream");
+                        sleep(Duration::from_secs(1)).await;
+                    }
```
deuszx: You're already sleeping and logging in `acknowledgement_task.run()` when it errors. Remove one of them - either here or in `fn run()` (I'd suggest the latter).

linera-service/src/exporter/storage.rs
```diff
@@ -115,8 +117,14 @@ where
         }
     }
 
-    async fn push_block(&mut self, block_hash: CryptoHash) -> Result<(), ExporterError> {
-        self.shared_canonical_state.push(block_hash).await
+    async fn get_blobs(&self, blobs: &[BlobId]) -> Result<Vec<Arc<Blob>>, ExporterError> {
+        let tasks = blobs.iter().map(|id| self.get_blob(*id));
```
deuszx, line 124: I'm not so sure all these optimization are better than simply reading blobs in batch from the underlying storage - rather than having to do multiple request-response roundtrips to the underlying database for every blob separately.

linera-service/src/exporter/runloops/indexer/client.rs
deuszx: Why not create an unbounded channel here, with throttling, and just keep sending new blocks to the `sink` here to push them to the indexer?

linera-service/src/exporter/runloops/indexer/indexer_exporter.rs
```diff
@@ -0,0 +1,221 @@
...
+            ExporterError::DestinationError
+        );
+
+        let furure = signal.into_future();
+        let mut pinned = Box::pin(furure);
+
+        let address = self.destination_config.indexer_address();
+        let mut client = IndexerClient::new(&address, self.options)?;
+
+        loop {
```
deuszx, line 67: Isn't this loop dangerous? It will keep on recreating the connection with the indexer destination regardless of errors. Maybe at least add some retry limit to it.

linera-service/src/exporter/runloops/indexer/client.rs
```diff
@@ -0,0 +1,86 @@
...
+            .max_decoding_message_size(GRPC_MAX_MESSAGE_SIZE);
+
+        Ok(Self {
+            client,
+            retry_delay: options.retry_delay,
+            max_retries: options.max_retries,
+        })
+    }
+
+    pub(super) async fn synchronize_long_lived_stream(
```
deuszx: The name isn't very helpful in understanding what's happening - add a short comment. And maybe rename to `setup_indexer_client` - since it tries to establish a streaming connection with the destination.

linera-service/src/exporter/runloops/indexer/client.rs
```diff
@@ -0,0 +1,86 @@
...
+    ) -> Result<(Sender<Element>, Streaming<()>), ExporterError> {
+        let mut retry_count = 0;
+        let streams = loop {
+            let (stream, sink) = tokio::sync::mpsc::channel(queue_size);
+            let request = Request::new(ReceiverStream::new(sink));
+            match self.client.index_batch(request).await {
+                Ok(res) => break (stream, res.into_inner()),
+                Err(e) => {
+                    if retry_count > self.max_retries {
+                        return Err(ExporterError::SynchronizationFailed(e.into()));
```
deuszx, line 53: I think it's more like an connection initialization error rather than synchronization 🤔 The only time when we can handle the error from call to `client.index_batch` is before we return from this loop so this can't happen while we're already connected?

linera-service/src/exporter/runloops/indexer/client.rs
```diff
@@ -0,0 +1,86 @@
...
+    pub(super) async fn synchronize_long_lived_stream(
+        &mut self,
+        queue_size: usize,
+    ) -> Result<(Sender<Element>, Streaming<()>), ExporterError> {
+        let mut retry_count = 0;
+        let streams = loop {
+            let (stream, sink) = tokio::sync::mpsc::channel(queue_size);
+            let request = Request::new(ReceiverStream::new(sink));
+            match self.client.index_batch(request).await {
+                Ok(res) => break (stream, res.into_inner()),
```
deuszx: Instead of assigning this `loop` to a value and returning, maybe just `return` from here. I think it is more readable.

linera-service/src/exporter/runloops/indexer/indexer_exporter.rs
```diff
@@ -0,0 +1,221 @@
...
+        Self {
+            incoming,
+            destination_id,
+            storage,
+        }
+    }
+
+    async fn run(&mut self) -> anyhow::Result<()> {
+        while self.incoming.message().await?.is_some() {
+            self.storage.increment_destination(self.destination_id);
```
deuszx, line 140: This assumes that the indexer destination processes blocks in order in which they were sent - which doesn't have to be the case. On top of that, the `Element` we're sending can be either block or a blob and there's no contract when the recipient should send back the "ack" - we might as well be incrementing the destination when recipient acks the blob, not a block.

Maybe we should change the `indexer.proto` so that the type return from the service is something that can be mapped to a `BlockHeight` and change `increment_destination` to `set_processed_block(block_height)` and make sure we increment it monotonically only. I.e. no skipped blocks.

linera-service/src/exporter/runloops/indexer/indexer_exporter.rs
```diff
@@ -0,0 +1,221 @@
...
+        }
+
+        tracing::error!("stream closed unexpectedly, retrying to establish a stream");
```
deuszx: There is no "retrying to establish a stream" in this function. If the caller does that, that should be logged there - not here. Please remove. Also the sleep is unnecessary - the caller should decide what to do in case of premature closing.

linera-service/src/exporter/runloops/indexer/indexer_exporter.rs
```diff
@@ -0,0 +1,221 @@
...
+            start_height,
+            storage,
+            buffer: sender,
+        }
+    }
+
+    async fn run(&mut self) -> anyhow::Result<()> {
+        let mut index = self.start_height;
+        let mut futures = FuturesOrdered::new();
+        while futures.len() != self.queue_size {
```
deuszx: This is dangerous because `100 != 5`. Let's make sure we push to `futures` only if there is still capacity.

linera-service/src/exporter/runloops/validator_exporter.rs
```diff
@@ -102,27 +111,35 @@ where
 
     async fn run(
         &self,
-        mut receiver: Receiver<Arc<ConfirmedBlockCertificate>>,
+        mut receiver: Receiver<(Arc<ConfirmedBlockCertificate>, Vec<BlobId>)>,
```
deuszx, line 113: A block already has `required_blob_ids()` which returns all blobs (ids) that are used/created by that block. Why do we need to send it separately?

linera-service/src/exporter/test_utils.rs
```diff
@@ -0,0 +1,497 @@
...
+}
+
+#[derive(Clone, Default)]
+pub(crate) struct DummyValidator {
+    pub(crate) fault_guard: Arc<AtomicBool>,
+    pub(crate) blobs: Arc<DashSet<BlobId>>,
+    pub(crate) state: Arc<DashSet<CryptoHash>>,
+}
+
+impl DummyValidator {
```
deuszx, line 142: Nit: for `DummyIndexer` you first have `impl Indexer for DummyIndexer` followed by `impl DummyIndexer` and here it's the other way around. Can you make consistent?

linera-service/src/exporter/test_utils.rs
```diff
@@ -0,0 +1,497 @@
...
+    }
+
+    async fn handle_lite_certificate(
+        &self,
+        _request: Request<linera_rpc::grpc::api::LiteCertificate>,
+    ) -> Result<Response<linera_rpc::grpc::api::ChainInfoResult>, Status> {
+        unimplemented!()
+    }
+
+    async fn handle_confirmed_certificate(
```
deuszx, line 170: nit: let's put the only implemented method in the top of `impl` block` and all `unimplemented!()` ones to follow.

linera-service/src/exporter/test_utils.rs
```diff
@@ -0,0 +1,497 @@
...
+        DestinationKind::Validator
+    }
+
+    fn blobs(&self) -> &DashSet<BlobId> {
+        self.blobs.as_ref()
+    }
+
+    fn fault_guard(&self) -> &AtomicBool {
+        self.fault_guard.as_ref()
+    }
```
deuszx: nit: let's replace this with `fn set_faulty()` and `fn unset_faulty()`.

linera-service/src/exporter/storage.rs
```diff
@@ -138,16 +146,37 @@ where
...
+        let block = self
             .shared_storage
             .shared_canonical_state
             .get(index)
             .await?;
-        self.shared_storage.get_block(hash).await
+
+        Ok((
+            self.shared_storage.get_block(block.block_hash).await?,
+            block.blobs.into(),
```
deuszx, line 164: Again - why not use `block.required_blob_ids()` ?

linera-service/src/cli/net_up_utils.rs
usagi32: It's remaining part from the CLI. it can be used later if required

linera-service/src/exporter/common.rs
usagi32: The order these blocks are seen by this validator. Just the same naming scheme used in previous PR for same the field. In case we want to name it something different, what's on your mind?

linera-service/src/exporter/common.rs
usagi32: Let me try

linera-service/src/exporter/proto/indexer.proto
usagi32: But would it be better if we type cast it explicitly? 
`message CryptoHash {
  bytes bytes = 1;
}`
can be any length, of course this would have been checked at the implementation level.

linera-service/src/exporter/proto/indexer.proto
usagi32: so like `rpc.proto`, just simple bytes?

linera-service/src/exporter/runloops/block_processor/walker.rs
usagi32: Indexing for blobs was not that necessary for the validator destination as for the other destinations nor that was planned that well at that time, so this could have been broken and be a part of separate PR. The validator destination already gives a gRPC response for missing blobs and and those are sent in a round trip then the block is sent again. But this was stateless, we can now avoid sending those blobs again if we have done so before

linera-service/src/exporter/runloops/block_processor/walker.rs
usagi32: I think I'll change it to `created_blobs`

linera-service/src/exporter/runloops/block_processor/walker.rs
usagi32: we want all the blobs that are `required_blob_ids - created_blob_ids`. These will be made part of the `CanonicalState` or `CanonicalBlock`

linera-service/src/exporter/runloops/indexer/client.rs
usagi32: Any examples? I mean bounded channels are always better unless I am missing something?

linera-service/src/exporter/runloops/indexer/client.rs
usagi32: This can happen happen after new connections. It's mainly for the gRPC stream

linera-service/src/exporter/runloops/indexer/indexer_exporter.rs
usagi32: Retrying is handled in `synchronize_long_lived_stream`

linera-service/src/exporter/runloops/indexer/indexer_exporter.rs
usagi32: in `acknowledgement_task.run`, sleeping and logging occurs on an `Ok` variant when the stream is closed normally but since this is an infinite it's not supposed to happen. Here this is on an error variant. So both are exclusive

linera-service/src/exporter/runloops/indexer/indexer_exporter.rs
usagi32: Yes, gRPC streams does maintain the order of messages. So there is a monotonic guarantee. And we only sent "ack" back for the blocks and not for the blobs. `DummyIndexer` has a reference implementation of this trait

linera-service/src/exporter/runloops/indexer/indexer_exporter.rs
usagi32: we started from 0

linera-service/src/exporter/runloops/mod.rs
usagi32: oh right, this is a `JoinError`, let me see what we can do with it

linera-service/src/exporter/runloops/validator_exporter.rs
usagi32: This is for `Blob`s that are not contained by that block and not already sent before

linera-service/src/cli/net_up_utils.rs
deuszx: If it's unneeded please remove.

linera-service/src/exporter/proto/indexer.proto
deuszx: Yes. If you import `rpc.proto` (as I suggested in another comment) you won't even have to write it down.

linera-service/src/exporter/runloops/block_processor/walker.rs
deuszx: My point is - why do you need to discriminate? When receiving validator sees a certificate, it checks whether it has all of `certificiate.required_blob_ids()` in its DB - if not, responds with `BlobNotFound` error. This means both types of blobs (created and used/required) are treated the same way. Why do you treat them differently here?

linera-service/src/exporter/runloops/validator_exporter.rs
deuszx: What's the difference though? a block (contained within `ConfirmedBlockCertificate`) already referes to all blobs that it needs to be fully verified. Why can't we use that?

linera-service/src/exporter/runloops/mod.rs
usagi32: this is a workaround for now, all this (all the detached tasks and threads, their handles and joining) is handled properly in the committee exporter PR using dedicated types and methods

linera-service/src/exporter/storage.rs
usagi32: I wouldn't say multiple roundtrips, these requests will happen in parallel not in serial also caching and locking in the cache can be parallelized

linera-service/tests/local_net_tests.rs
usagi32: it can be difficult, if we start the exporter after the block is proposed, the exporter will not have the notification for that block. I had manually verified this by not starting the exporter and asserting that the new validator has not seen any blocks as this new validator is detached from the current committee, so the client will not propose blocks to it

linera-service/src/exporter/proto/indexer.proto
```diff
@@ -0,0 +1,23 @@
+syntax = "proto3";
+package indexer.linera_indexer;
+
+import "google/protobuf/empty.proto";
+
+service Indexer {
+    rpc IndexBatch(stream Element) returns (stream google.protobuf.Empty) {}
```
deuszx, line 12: If you won't change this interface to return a stream of block heights (so that it's clear that we're _ack-ing_ **blocks**) then at least add a very clear comment on this interface for the implementors.

linera-client/src/config.rs
deuszx: Seems unaddressed

linera-service/src/exporter/runloops/block_processor/walker.rs
deuszx: You didn't answer my question

linera-service/src/exporter/runloops/validator_exporter.rs
deuszx: Same here - I don't see you answering my question.

linera-service/src/exporter/runloops/validator_exporter.rs
deuszx: Unaddressed.

linera-service/src/exporter/runloops/indexer/indexer_exporter.rs
deuszx: Re-establishing of connection with the indexer destination is done here - when you call `client.setup_indexer_client`. In `setup_indexer_client` you retry "within" a single loop iteration but when one of the tasks _here_ returns an error you log that and loop again - effectively recreating the connection with the destination. And THAT has infinite loop.

linera-service/src/exporter/runloops/validator_exporter.rs
usagi32: I have removed the `validator_address()` function, replaced it with a normal `fn address()` on `Destination` that matches correctly based on the `DestinationKind`

linera-service/src/exporter/runloops/indexer/indexer_exporter.rs
deuszx: gRPC stream maintains the order but nothing prevents the client from acking the blob, instead of block height. They are also free to process the incoming messages in parallel so they _could_ ACK blocks out of order.

Anyway, that's probably a too big of a change for now so lets keep it simple.

linera-service/src/exporter/runloops/validator_exporter.rs
deuszx: [It returns](https://github.com/linera-io/linera-protocol/pull/3949/files#diff-4ef7ab1957e36d4be3dc56a442e6c972eb3a3c606357e993ebdc9d417300a2d2R304-R308) http(s) for indexer destination - I thought it uses gRPC protocol?

usagi32: gRPC itself uses http as the address, it's different for the validator as the validator client/node in the linera-rpc internally parses "grpc" and "grpcs" into "http" and "https" respectively

deuszx: Ah, in config files we use `grpc`/`grpcs` but - like you say - internally we map it to TlsConfig 👍

linera-service/src/exporter/runloops/indexer/indexer_exporter.rs
deuszx:
```suggestion
        while futures.len() <= self.queue_size {
```

is safer and uses the same amount of characters

linera-service/src/exporter/runloops/validator_exporter.rs
deuszx: This is validator exporter and the destination recipient already tells you which blobs it's missing via `BlobsNotFound`. You can then read those missing blobs from storage and upload. This will simplify the API here a lot.

usagi32: we have to check this on our end too as a malicious validator can also send faulty responses, this will make sure we only send among those blobs that are meant to be sent, incase another error even after this we then stop

deuszx: That can be checked via `block.required_blob_ids().contains(_)`.

linera-service/src/exporter/runloops/indexer/client.rs
deuszx: That's what I meant - this error happens only when we try to establish a connection.

It's a bit weird that there is no separate method for that - i.e. simply establishing a connection - and we have to call `index_batch` with an empty message I suppose(?)

linera-service/src/exporter/runloops/validator_exporter.rs
usagi32: Let's say a block has 100 required blobs, out of those we had already sent 98 of them from some other chains/blocks
Now this block will be pushed in form of `CanonicalBlock` with its hash and only those 2 remaining blobs that are yet not seen by that destination. And same here we will check the response `BlobsNotFound` among those 2 remaining blocks, if we do it with `required_blob_ids()` we could be sending all of those 100

linera-service/src/exporter/runloops/indexer/client.rs
usagi32: Ah no, not with an empty message. That's handled by gRPC for creating a stream, we only need to supply the channel half

linera-service/src/exporter/runloops/validator_exporter.rs
deuszx: `BlobsNotFound(blob_ids)` will include only the 2 missing blob ids, not all 100. Then you compare the `blob_ids` against `block.required_blob_ids()` to make sure the validator is not malicious.

linera-service/src/exporter/runloops/block_processor/walker.rs
deuszx: You haven't done it - rename and add the comment these are blobs contained by the `Block` itself.

linera-service/src/exporter/runloops/block_processor/walker.rs
```diff
@@ -139,6 +169,8 @@ impl ProcessedBlock {
         Self {
             dependencies,
             block: block_id,
+            required_blobs: block.required_blob_ids().into_iter().collect(),
```
deuszx, line 177: created blobs are also required blobs. Why not remove all created blob (ids) from this collection right away? they will be sent together with the `Block` itself and couldn't have been sent before.

linera-service/src/exporter/runloops/block_processor/walker.rs
```diff
@@ -54,9 +59,29 @@ where
                 continue;
             }
 
+            // all the block dependecies have been resolved for this block
+            // now just resolve the blobs
+            let mut blobs_to_send = Vec::new();
+            let mut blobs_to_index_block_with = Vec::new();
+            for id in node_visitor.node.required_blobs {
```
deuszx, line 66: Right now this collection has `created_blobs` as well (b/c `block.required_blob_ids()` includes them) and we have already sent the created blobs (together with the block) earlier.

linera-service/src/exporter/runloops/validator_exporter.rs
usagi32: Yes, I was talking about this part that can be malicious, `BlobsNotFound(blob_ids)` could show a different number than 2 for e.g. all the 100 blob ids, if we check it `required_blob_ids()` then all of them would have to be sent

deuszx: if that's the case then our protocol has the same problem - i.e. our clients are sending all the blobs that the validator is supposedly missing. 

What if the destination had a bug and lost some of the blobs and needs to request them again?

linera-service/src/exporter/runloops/block_processor/walker.rs
usagi32: But we have to index them in the blob's SetView if they are seen for the first time, also they could have been sent before as blobs are content addressed in the core protocol, so any two blocks can create the same blob

deuszx: They are seen for the first time if they're part of the `block.created_blob` field. They are created as part of the block execution so couldn't have been seen earlier. They won't be included in that set if someone already publushed them before.

linera-service/src/exporter/runloops/validator_exporter.rs
usagi32: Yes, I agree. Block exporter can also be subject to abuse this way as it's a automated service unlike the clients also they work on a quorum of consensus. 

What should we do then?

linera-service/src/exporter/runloops/block_processor/walker.rs
usagi32: So that means two block cannot publish the same `Data` or `ContractBytecode` \ `ServiceBytecode` blob?

We can see the code above in `walk()` fn that these created blobs are only used for filtering out what we need to index and finally send

deuszx: `created_blobs` != `published blob`. Contract and service bytecode are never part of the `created_blobs`.

usagi32: maybe you are right, so should we not index the created blobs in the SetView? In the code to resolve blobs, created_blobs are only used to index the blobs in the SetView and also just to exclude them from the `CanonicalBlock` that contains the blobs to actually send separately

linera-service/src/exporter/runloops/block_processor/walker.rs
usagi32: it's excluded in the `blobs_to_send` that's why I we were taking the `created_blobs` field separately, just to exclude them (and also index them in the SetView)

PR #3950, opened 2025-05-15
Creating cli module

ndr-ds:
## Motivation

This will be useful to reuse the code here in other places

## Proposal

Create module

## Test Plan

CI

## Release Plan

- Nothing to do / These changes follow the usual release cycle.

PR #3952, opened 2025-05-15
Introduce linera multi-benchmark

ndr-ds:
## Motivation

To reproduce big numbers in benchmarks, we need to use multiple processes to exhaust the CPU on big VMs.
Right now, the only way to do that is manually starting one wallet per process, then manually starting all the benchmarks as well.
To reach big numbers like 1M TPS, I was using 4 different benchy instances, and spawning 12 processes on each. That means that I had to manually wallet init different wallets 48 times, then manually start the benchmark process 48 times... not ideal.

## Proposal

Let's automate this whole process with `linera multi-benchmark`. You can specify how many processes you want, and it'll gradually start them for you, so that TPS is gradually ramped up. The same arguments you pass to `linera benchmark` you can pass to this one as well, and it'll create all the processes with those arguments.

## Test Plan

Ran this with multiple processes against a local network, it works :)

## Release Plan

### DIFF
--- a/linera-service/src/cli/command.rs
+++ b/linera-service/src/cli/command.rs
@@ -444,6 +444,31 @@ pub enum ClientCommand {
     #[cfg(feature = "benchmark")]
     Benchmark(BenchmarkCommand),
 
+    /// Runs multiple `linera benchmark` processes in parallel.
+    #[cfg(feature = "benchmark")]
+    MultiBenchmark {
+        /// The number of `linera benchmark` processes to run in parallel.
+        #[arg(long = "processes", default_value = "1")]
+        processes: usize,
+
+        /// The faucet (which implicitly defines the network)
+        #[arg(long = "faucet")]
+        faucet: String,
+
+        /// If running on local SSD, specify the directory to store the storage and wallet.
+        /// If not specified, a temporary directory will be used for each client.
+        #[arg(long = "ssd-dir")]
+        ssd_dir: Option<String>,
+
+        /// The benchmark command to run.
+        #[clap(flatten)]
+        command: BenchmarkCommand,
+
+        /// The delay between starting the benchmark processes, in seconds.
+        #[arg(long, default_value = "10")]
+        delay_between_processes: u64,
+    },
+
     /// Create genesis configuration for a Linera deployment.
     /// Create initial user chains and print information to be used for initialization of validator setup.
     /// This will also create an initial wallet for the owner of the initial "root" chains.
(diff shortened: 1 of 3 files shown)


ndr-ds:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/3952?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#3952** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/3952?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/3952?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3951** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/3951?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3950** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/3950?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

linera-service/src/linera/main.rs
```diff
@@ -863,6 +870,172 @@ impl Runnable for Job {
                     .await?;
             }
 
+            #[cfg(feature = "benchmark")]
+            MultiBenchmark {
```
afck, line 874: Does this even use `storage`? Maybe this should be in the standalone `run` function and just return when it's done, rather than in `Job::run`?

linera-service/src/linera/main.rs
```diff
@@ -863,6 +870,172 @@ impl Runnable for Job {
...
+                let mut rx_handles = Vec::new();
+                for ((child, stdout_handle, stderr_handle), rx) in results {
+                    children.push(child);
+                    stdout_handles.push(stdout_handle);
+                    stderr_handles.push(stderr_handle);
+                    rx_handles.push(rx);
+                }
+
+                if confirm_before_start {
+                    info!("Waiting until all children processes are ready...");
```
afck, line 956:
```suggestion
                    info!("Waiting until all child processes are ready...");
```

afck (approved):
We should revisit this in the future and find out why a single process with many chains is slower than multiple processes with fewer chains.

ndr-ds:
Yeah, the design for `linera benchmark` I think ends up with too much thread contention, unfortunately. I have some ideas on how to improve it, but you're right, one process (if well designed), should in theory be able to max out the CPU, regardless of the VM size.
I'll add that to my TODO list :)

linera-service/src/linera/main.rs
ndr-ds: Good point!

ndr-ds:
### Merge activity

* **May 19, 4:04 PM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/linera-io/linera-protocol/3952).

PR #3954, opened 2025-05-16
windows build

usagi32:
## Motivation
Some minor issues that fails the build for windows.
Also fails the integration tests for the linera-service crate.

## Proposal
- Change the default stack size for windows from 1 megabytes to 8 megabytes which is the default for linux.
- Windows binary paths don't recognize a binary without the ".exe" extension.
- Some other minor workarounds around the paths and directories and toml naming.
## Test Plan
Nothing to test other than the CI, these are just simple workarounds.
Nevertheless, had included some tests later.
## Release Plan

### DIFF
--- a/linera-service/src/storage.rs
+++ b/linera-service/src/storage.rs
@@ -220,45 +220,54 @@ example service:tcp:127.0.0.1:7878:table_do_my_test"
                     "For RocksDB, the formatting has to be rocksdb:directory or rocksdb:directory:spawn_mode:namespace");
             }
 
-            // windows has directory naming scheme like "C://xyz", "//??D:xyz/" etc.
-            let owned = if cfg!(windows) {
-                s.replacen(':', ";", 1)
-            } else {
-                s.to_string()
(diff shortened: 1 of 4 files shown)


linera-service/src/storage.rs
```diff
@@ -219,9 +219,18 @@ example service:tcp:127.0.0.1:7878:table_do_my_test"
                 bail!(
                     "For RocksDB, the formatting has to be rocksdb:directory or rocksdb:directory:spawn_mode:namespace");
             }
-            let parts = s.split(':').collect::<Vec<_>>();
+
+            // windows has directory naming scheme like "C://xyz", "//??D:xyz/" etc.
+            let owned = if cfg!(windows) {
+                s.replacen(':', ";", 1)
```
Twey, line 225: Won't this replace the first instance, leaving you with `rocksdb;C:\Windows\…:blocking:namespace`?

Also how does this interact with relative paths?

usagi32: Actually this won't, rocksdb is already striped above before.

Though relative paths are not used by the commands in the workspace, we should certainly cover those case. I have updated it and also included some tests.

PR #3956, opened 2025-05-16
Chain description duplication

Twey:
## Motivation

The `chain_description` function is duplicated between the `ClientContext` struct and the `ClientContext` trait due to `Sync`ness issues with `Persist`.  As requested by @deuszx in https://github.com/linera-io/linera-protocol/pull/3956, deduplicate it, which entails several proximal cleanups.

## Proposal

- Further our goal of making `ChainClient` a syntactic convenience for calling `Client` with a chain ID by removing the `ensure_has_chain_description` call from `make_chain_client` and correspondingly making `make_chain_client` sync.  This removes the need for the wallet to be `Sync`, even when we `tokio::spawn`.
- Merge `LocalPersist` and `Persist` using conditional compilation, as we do elsewhere.
- Split the provided functions in `ClientContext` into `ClientContextExt` to allow dropping `async_trait`.
- Use the `ClientContextExt` version of `chain_description()` in `client_context.rs`.

## Test Plan

CI.

## Release Plan

## Links

### DIFF
--- a/linera-client/src/chain_listener.rs
+++ b/linera-client/src/chain_listener.rs
@@ -7,7 +7,6 @@ use std::{
     time::Duration,
 };
 
-use async_trait::async_trait;
 use futures::{
     future::{join_all, select_all},
     lock::Mutex,
@@ -61,16 +60,16 @@ pub struct ChainListenerConfig {
 
 type ContextChainClient<C> = ChainClient<<C as ClientContext>::Environment>;
 
-#[cfg_attr(not(web), async_trait, trait_variant::make(Send))]
-#[cfg_attr(web, async_trait(?Send))]
-pub trait ClientContext: 'static {
+#[cfg_attr(not(web), trait_variant::make(Send))]
+#[allow(async_fn_in_trait)]
+pub trait ClientContext {
     type Environment: linera_core::Environment;
 
     fn wallet(&self) -> &Wallet;
 
     fn storage(&self) -> &<Self::Environment as linera_core::Environment>::Storage;
 
-    async fn make_chain_client(&self, chain_id: ChainId)
+    fn make_chain_client(&self, chain_id: ChainId)
         -> Result<ContextChainClient<Self>, Error>;
 
     fn client(&self) -> &linera_core::client::Client<Self::Environment>;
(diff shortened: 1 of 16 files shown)


linera-client/src/chain_listener.rs
```diff
@@ -61,17 +60,16 @@ pub struct ChainListenerConfig {
 
 type ContextChainClient<C> = ChainClient<<C as ClientContext>::Environment>;
 
-#[cfg_attr(not(web), async_trait, trait_variant::make(Send))]
-#[cfg_attr(web, async_trait(?Send))]
-pub trait ClientContext: 'static {
+#[cfg_attr(not(web), trait_variant::make(Send))]
+#[allow(async_fn_in_trait)]
+pub trait ClientContext {
```
deuszx, line 65: If we could extract `update_wallet_for_new_chain` and `update_wallet` - this method wouldn't have any async fns anymore.

linera-client/src/persistent/mod.rs
```diff
@@ -29,8 +29,8 @@ pub use memory::Memory;
 /// the value in memory.
 ///
 /// `LocalPersist` is a non-`Send` version.
-#[trait_variant::make(Persist: Send)]
-pub trait LocalPersist: Deref {
+#[cfg_attr(not(web), trait_variant::make(Send))]
```
deuszx, line 32: The comment above is now outdated.

linera-core/src/client/mod.rs
```diff
@@ -273,9 +273,7 @@ impl<Env: Environment> Client<Env> {
             ));
         }
 
-        let _ = self.ensure_has_chain_description(chain_id).await?;
```
deuszx, line 276: Why do we no longer have to check that?

Twey: I believe that any code path that wants a chain description will retrieve it by calling `ensure_has_chain_description`.  So we should still check it, just lazily.

linera-client/src/chain_listener.rs
Twey: I plan to completely revamp this trait in a short while after moving the wallet into the client, so I'm not keen to do it here.

deuszx (approved):
This addresses my original comment but not in a way that I was expecting - i.e. simply removes the code paths that made the call to the duplicated method.

Twey:
> This addresses my original comment but not in a way that I was expecting - i.e. simply removes the code paths that made the call to the duplicated method.

I don't think that's correct?  There are still two calls to `chain_description`, but they now both point to the trait method rather than the inherent function, which is removed.

deuszx:
> > This addresses my original comment but not in a way that I was expecting - i.e. simply removes the code paths that made the call to the duplicated method.
> 
> I don't think that's correct? There are still two calls to `chain_description`, but they now both point to the trait method rather than the inherent function, which is removed.

You're right. I didn't notice that the `trait` and `impl` were collapsed.

PR #3958, opened 2025-05-17
fix symlinks in linera-web

ma2bd:
## Motivation

fix symlinks inbroduced by #3908

## Proposal

remove the extra path name

## Test Plan

not sure


ndr-ds (approved):

ma2bd:
Files are deleted in #3955 anyway

PR #3959, opened 2025-05-18
Add `input_name` entry to some RFQ types.

MathieuDutSik:
## Motivation

For a given type T, we can apply the derive statement for `SimpleObject` and `InputObject`. Those will implement the `Object` and `InputObject` GraphQL trait. For a given type, we should implement at most one of those. 

If both are introduced, then we should have an `input_name` statement associated to it so that the Input type has a different name to the output type.

This is generally satisfied in the code, with the exception of 4 types in RFQ.

## Proposal

The solution is to add the `input_name` statement.
It might be a better idea to remove some traits definition altogether. 

## Test Plan

The CI. Currently, the RFQ passes only via the README.md test

## Release Plan

## Links

None

### DIFF
--- a/examples/rfq/src/state.rs
+++ b/examples/rfq/src/state.rs
@@ -10,12 +10,14 @@ use rfq::{RequestId, TokenPair, Tokens};
 use serde::{Deserialize, Serialize};
 
 #[derive(Clone, Debug, Serialize, Deserialize, SimpleObject, InputObject)]
+#[graphql(input_name = "QuoteRequestedInput")]
 pub struct QuoteRequested {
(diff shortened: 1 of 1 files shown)


afck (approved):
Good catch!

deuszx (approved):
I would expect the derivation to fail if this isn's satisfied 😬

afck:
Yes, it's a bit weird that this is a _runtime_ error. [But it is documented.](https://async-graphql.github.io/async-graphql/en/define_simple_object.html#used-for-both-input-and-output)

PR #3960, opened 2025-05-19
Implement events for EVM based applications.

### LINKED ISSUE #3785: Implement `process_streams` for EVM.
#3784 introduces `process_streams`, which is called in Wasm contracts whenever streams they subscribe to have new events.

We should do the same for EVM contracts, and also expose `subscribe_to_events` and `unsubscribe_from_events`.

(Related to #365.)

### LINKED ISSUE #3966: Possibly change the API of `fn process_streams`
The code of PR #3960 uses a deserialization in its first line. We could send data in the right type by using RLP and `alloy-sol-types`.

MathieuDutSik:
## Motivation

A major feature of Linera is the streams. This PR implements them for the EVM.

Fixes #3785 
Fixed #3966 

## Proposal

The following was done:
* The `LineraRuntimePrecompile` was changed so that the input is put there. So the awkward conversion with `vec` has been changed in `revm.rs`.
* A lot of conversion-to-string errors have been eliminated, and the code is much simplified.
* Correspondingly, the serialization has been expanded for `LineraRuntimePrecompile`. The `LineraTypes.sol` is thus expanded.
* The corresponding file `LineraTypes.yaml` has been added to the code.
* The `fn process_streams` has been added for the REVM.rs and a test `test_evm_process_streams_end_to_end_counters` has been added.
* The `Bytes::copy_from_slice(&resul)` have been replaced by `Bytes::from(result)` 
* The `transact_commit` / `transact` are now taking a `Vec<u8>`, which reduces the copy operations.
* Many errors `ExecutionError::EvmQuery` are replaced by `into()` and `?`.
* The `execute_message` and `process_streams` are now checking for the presence of the selector which avoids having neater errors.
* We now check that the `Success` does return `Return` which allows better error handling.

## Test Plan

A test has been added for `process_streams` in the CI.

## Release Plan

## Links

None.

### DIFF
--- a/linera-execution/solidity/Linera.sol
+++ b/linera-execution/solidity/Linera.sol
@@ -7,74 +7,69 @@ import "./LineraTypes.sol";
 // This library provides Linera functionalities to EVM contracts
 // It should not be modified.
 
-// The Precompile keys below correspond to the BCS serialization of
-// the `PrecompileTag` in `linera-execution/src/evm/revm.rs`.
-// (0,0): chain_id
-// (0,1): application_creator_chain_id
-// (0,2): chain_ownership
-// (0,3): read data blob
-// (0,4): assert data blob exists
-// (1,0): try_call_application
-// (1,1): validation round
-// (1,2): send_message
-// (1,3): message_id
-// (1,4): message_is_bouncing
-// (2,0): try_query_application
 library Linera {
 
-    function inner_chain_id(uint8 val) internal returns (LineraTypes.ChainId memory) {
-        address precompile = address(0x0b);
-        bytes memory input1 = new bytes(2);
-        input1[0] = bytes1(uint8(0));
-        input1[1] = bytes1(val);
-        (bool success, bytes memory output1) = precompile.call(input1);
-        require(success);
-        return LineraTypes.bcs_deserialize_ChainId(output1);
-    }
-
     function chain_id() internal returns (LineraTypes.ChainId memory) {
-        return inner_chain_id(uint8(0));
+        address precompile = address(0x0b);
+        LineraTypes.BaseRuntimePrecompile memory base = LineraTypes.BaseRuntimePrecompile_case_chain_id();
+        LineraTypes.RuntimePrecompile memory input1 = LineraTypes.RuntimePrecompile_case_base(base);
+        bytes memory input2 = LineraTypes.bcs_serialize_RuntimePrecompile(input1);
+        (bool success, bytes memory output) = precompile.call(input2);
+        require(success);
+        return LineraTypes.bcs_deserialize_ChainId(output);
     }
 
     function application_creator_chain_id() internal returns (LineraTypes.ChainId memory) {
-        return inner_chain_id(uint8(1));
+        address precompile = address(0x0b);
+        LineraTypes.BaseRuntimePrecompile memory base = LineraTypes.BaseRuntimePrecompile_case_application_creator_chain_id();
+        LineraTypes.RuntimePrecompile memory input1 = LineraTypes.RuntimePrecompile_case_base(base);
+        bytes memory input2 = LineraTypes.bcs_serialize_RuntimePrecompile(input1);
+        (bool success, bytes memory output) = precompile.call(input2);
+        require(success);
+        return LineraTypes.bcs_deserialize_ChainId(output);
     }
 
     function chain_ownership() internal returns (LineraTypes.ChainOwnership memory) {
         address precompile = address(0x0b);
-        bytes memory input1 = new bytes(2);
-        input1[0] = bytes1(uint8(0));
-        input1[1] = bytes1(uint8(2));
-        (bool success, bytes memory output1) = precompile.call(input1);
+        LineraTypes.BaseRuntimePrecompile memory base = LineraTypes.BaseRuntimePrecompile_case_chain_ownership();
+        LineraTypes.RuntimePrecompile memory input1 = LineraTypes.RuntimePrecompile_case_base(base);
+        bytes memory input2 = LineraTypes.bcs_serialize_RuntimePrecompile(input1);
+        (bool success, bytes memory output) = precompile.call(input2);
         require(success);
-        return LineraTypes.bcs_deserialize_ChainOwnership(output1);
+        return LineraTypes.bcs_deserialize_ChainOwnership(output);
     }
 
     function read_data_blob(bytes32 hash) internal returns (bytes memory) {
         address precompile = address(0x0b);
-        bytes memory input1 = new bytes(2);
-        input1[0] = bytes1(uint8(0));
-        input1[1] = bytes1(uint8(3));
-        bytes memory input2 = abi.encodePacked(input1, hash);
-        (bool success, bytes memory output1) = precompile.call(input2);
+        LineraTypes.CryptoHash memory hash2 = LineraTypes.CryptoHash(hash);
+        LineraTypes.BaseRuntimePrecompile_ReadDataBlob memory read_data_blob_ = LineraTypes.BaseRuntimePrecompile_ReadDataBlob(hash2);
+        LineraTypes.BaseRuntimePrecompile memory base = LineraTypes.BaseRuntimePrecompile_case_read_data_blob(read_data_blob_);
+        LineraTypes.RuntimePrecompile memory input1 = LineraTypes.RuntimePrecompile_case_base(base);
+        bytes memory input2 = LineraTypes.bcs_serialize_RuntimePrecompile(input1);
+        (bool success, bytes memory output) = precompile.call(input2);
         require(success);
-        return output1;
+        return output;
     }
 
     function assert_data_blob_exists(bytes32 hash) internal {
         address precompile = address(0x0b);
-        bytes memory input1 = new bytes(2);
-        input1[0] = bytes1(uint8(0));
-        input1[1] = bytes1(uint8(4));
-        bytes memory input2 = abi.encodePacked(input1, hash);
-        (bool success, bytes memory output1) = precompile.call(input2);
+        LineraTypes.CryptoHash memory hash2 = LineraTypes.CryptoHash(hash);
+        LineraTypes.BaseRuntimePrecompile_AssertDataBlobExists memory assert_data_blob_exists_ = LineraTypes.BaseRuntimePrecompile_AssertDataBlobExists(hash2);
+        LineraTypes.BaseRuntimePrecompile memory base = LineraTypes.BaseRuntimePrecompile_case_assert_data_blob_exists(assert_data_blob_exists_);
+        LineraTypes.RuntimePrecompile memory input1 = LineraTypes.RuntimePrecompile_case_base(base);
+        bytes memory input2 = LineraTypes.bcs_serialize_RuntimePrecompile(input1);
+        (bool success, bytes memory output) = precompile.call(input2);
         require(success);
-        assert(output1.length == 0);
+        assert(output.length == 0);
     }
 
     function try_call_application(bytes32 universal_address, bytes memory operation) internal returns (bytes memory) {
         address precompile = address(0x0b);
-        bytes memory input2 = abi.encodePacked(uint8(1), uint8(0), universal_address, operation);
+        LineraTypes.ApplicationId memory target = LineraTypes.ApplicationId(LineraTypes.CryptoHash(universal_address));
+        LineraTypes.ContractRuntimePrecompile_TryCallApplication memory try_call_application_ = LineraTypes.ContractRuntimePrecompile_TryCallApplication(target, operation);
+        LineraTypes.ContractRuntimePrecompile memory contract_ = LineraTypes.ContractRuntimePrecompile_case_try_call_application(try_call_application_);
+        LineraTypes.RuntimePrecompile memory input1 = LineraTypes.RuntimePrecompile_case_contract(contract_);
+        bytes memory input2 = LineraTypes.bcs_serialize_RuntimePrecompile(input1);
         (bool success, bytes memory output) = precompile.call(input2);
         require(success);
         return output;
@@ -82,18 +77,22 @@ library Linera {
 
     function validation_round() internal returns (LineraTypes.opt_uint32 memory) {
         address precompile = address(0x0b);
-        bytes memory input1 = new bytes(2);
-        input1[0] = bytes1(uint8(1));
-        input1[1] = bytes1(uint8(1));
-        (bool success, bytes memory output1) = precompile.call(input1);
+        LineraTypes.ContractRuntimePrecompile memory contract_ = LineraTypes.ContractRuntimePrecompile_case_validation_round();
+        LineraTypes.RuntimePrecompile memory input1 = LineraTypes.RuntimePrecompile_case_contract(contract_);
+        bytes memory input2 = LineraTypes.bcs_serialize_RuntimePrecompile(input1);
+        (bool success, bytes memory output) = precompile.call(input2);
         require(success);
-        LineraTypes.opt_uint32 memory val = LineraTypes.bcs_deserialize_opt_uint32(output1);
-        return val;
+        return LineraTypes.bcs_deserialize_opt_uint32(output);
     }
 
-    function send_message(bytes32 input_chain_id, bytes memory message) internal {
+    function send_message(bytes32 chain_id1, bytes memory message) internal {
         address precompile = address(0x0b);
-        bytes memory input2 = abi.encodePacked(uint8(1), uint8(2), input_chain_id, message);
+        LineraTypes.ChainId memory chain_id2 = LineraTypes.ChainId(LineraTypes.CryptoHash(chain_id1));
+        LineraTypes.ContractRuntimePrecompile_TryCallApplication memory try_call_application_;
+        LineraTypes.ContractRuntimePrecompile_SendMessage memory send_message_ = LineraTypes.ContractRuntimePrecompile_SendMessage(chain_id2, message);
+        LineraTypes.ContractRuntimePrecompile memory contract_ = LineraTypes.ContractRuntimePrecompile_case_send_message(send_message_);
+        LineraTypes.RuntimePrecompile memory input1 = LineraTypes.RuntimePrecompile_case_contract(contract_);
+        bytes memory input2 = LineraTypes.bcs_serialize_RuntimePrecompile(input1);
         (bool success, bytes memory output) = precompile.call(input2);
         require(success);
         require(output.length == 0);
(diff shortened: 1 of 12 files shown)


deuszx (commented):
How is `LineraTypes.yaml` being generated? Should it be added to the CI like the other `*.yaml` file genrations (schemas for indexer etc.)

linera-execution/solidity/Linera.sol
```diff
@@ -7,123 +7,177 @@ import "./LineraTypes.sol";
...
+        require(success);
+        return LineraTypes.bcs_deserialize_MessageIsBouncing(output);
+    }
+
+    function linera_emit(bytes memory stream_name, bytes memory value) internal returns (uint32) {
+        address precompile = address(0x0b);
+        LineraTypes.StreamName memory stream_name2 = LineraTypes.StreamName(stream_name);
+        LineraTypes.ContractRuntimePrecompile_Emit memory emit_ = LineraTypes.ContractRuntimePrecompile_Emit(stream_name2, value);
+        LineraTypes.ContractRuntimePrecompile memory contract_ = LineraTypes.ContractRuntimePrecompile_case_emit(emit_);
+        LineraTypes.RuntimePrecompile memory input1 = LineraTypes.RuntimePrecompile_case_contract(contract_);
```
deuszx, line 126: Good these are hidden behind a better API b/c I'd have no idea what's going on in here.

linera-execution/solidity/LineraTypes.sol
deuszx, line 1: Is this file re-generated only?

linera-service/tests/fixtures/evm_example_process_streams.sol
```diff
@@ -0,0 +1,42 @@
...
+
+    function increment_value(uint64 increment) external {
+        total_value = total_value + increment;
+        bytes memory stream_name = abi.encodePacked(uint8(42));
+        bytes memory value = abi.encode(total_value);
+        Linera.linera_emit(stream_name, value);
+    }
+
+    function process_streams(bytes memory input) external {
+        LineraTypes.StreamUpdates memory updates = LineraTypes.bcs_deserialize_StreamUpdates(input);
```
deuszx, line 25: Why does the contract need to deserialize that? Can't we pass in the proper type right away?

linera-service/tests/fixtures/evm_example_process_streams.sol
```diff
@@ -0,0 +1,42 @@
...
+        bytes memory stream_name = abi.encodePacked(uint8(42));
```
deuszx, line 26: Extract to a [constant](https://solidity-by-example.org/constants/)

linera-service/tests/linera_net_tests.rs
```diff
@@ -832,6 +831,124 @@ async fn test_evm_execute_message_end_to_end_counter(config: impl LineraNetConfi
...
+    let mutation = increment_valueCall { increment };
+    let mutation = mutation.abi_encode();
+    let mutation = EvmQuery::Mutation(mutation);
+    application1.run_json_query(mutation).await?;
+
+    // Third: process the inbox on chain2
+
+    node_service2.process_inbox(&chain2).await?;
+
+    // Fourth: getting the balance
```
deuszx, line 930: It's not a balance.

linera-execution/src/evm/revm.rs
```diff
@@ -274,55 +280,87 @@ fn address_to_user_application_id(address: Address) -> ApplicationId {
...
+enum BaseRuntimePrecompile {
     /// Key prefix for `chain_id`
     ChainId,
     /// Key prefix for `application_creator_chain_id`
     ApplicationCreatorChainId,
     /// Key prefix for `chain_ownership`
     ChainOwnership,
     /// Key prefix for `read_data_blob`
-    ReadDataBlob,
+    ReadDataBlob { hash: CryptoHash },
```
deuszx, line 291: This isn't a prefix anymore, is it? same for `AssertDataBlobExists`

linera-execution/solidity/LineraTypes.sol
MathieuDutSik: Yes.

linera-service/tests/fixtures/evm_example_process_streams.sol
MathieuDutSik: I thought about doing that.
This is possible but requires a little bit more work. So, in the end I decided for a streamlined solution. But this can change.

deuszx: Maybe create an issue for this with possible refactor.

linera-execution/src/evm/revm.rs
MathieuDutSik: Yes, thank you find finding the oversight.

linera-service/tests/fixtures/evm_example_process_streams.sol
MathieuDutSik: You are right about that.
However, we cannot put a constant value that is a bytes. So, the `stream_name` cannot be put as such a constant.

linera-execution/solidity/Linera.sol
MathieuDutSik: Yes, that handwritten code was looking very error-prone.
Better to use some computer generated code.

linera-execution/solidity/Linera.sol
```diff
@@ -7,123 +7,177 @@ import "./LineraTypes.sol";
...
 
     function read_data_blob(bytes32 hash) internal returns (bytes memory) {
         address precompile = address(0x0b);
-        bytes memory input1 = new bytes(2);
-        input1[0] = bytes1(uint8(0));
-        input1[1] = bytes1(uint8(3));
-        bytes memory input2 = abi.encodePacked(input1, hash);
-        (bool success, bytes memory output1) = precompile.call(input2);
+        LineraTypes.CryptoHash memory hash2 = LineraTypes.CryptoHash(hash);
+	LineraTypes.BaseRuntimePrecompile_ReadDataBlob memory read_data_blob_ = LineraTypes.BaseRuntimePrecompile_ReadDataBlob(hash2);
```
afck, line 45: Formatting: Some of these are tabs, some are spaces.

linera-service/tests/linera_net_tests.rs
```diff
@@ -832,6 +831,124 @@ async fn test_evm_execute_message_end_to_end_counter(config: impl LineraNetConfi
...
+#[cfg_attr(feature = "kubernetes", test_case(SharedLocalKubernetesNetTestingConfig::new(Network::Grpc, BuildArg::Build) ; "kubernetes_grpc"))]
+#[cfg_attr(feature = "remote-net", test_case(RemoteNetTestingConfig::new(None) ; "remote_net_grpc"))]
+#[test_log::test(tokio::test)]
+async fn test_evm_process_streams_end_to_end_counters(config: impl LineraNetConfig) -> Result<()> {
+    use alloy_primitives::B256;
+    use alloy_sol_types::{sol, SolCall};
+    use linera_base::vm::EvmQuery;
+    use linera_execution::test_utils::solidity::{get_evm_contract_path, read_evm_u64_entry};
+    use linera_sdk::abis::evm::EvmAbi;
+    let _guard = INTEGRATION_TEST_GUARD.lock().await;
```
afck, line 847: Usually there's a blank line after the imports.

```suggestion

    let _guard = INTEGRATION_TEST_GUARD.lock().await;
```

linera-execution/solidity/Linera.sol
MathieuDutSik: Yes, there was one. Just one.
Maybe we should detect tabs in CI.

linera-service/tests/linera_net_tests.rs
MathieuDutSik: Ok, I reorganized a bit the code of the file with respect to spaces.

linera-service/tests/fixtures/evm_example_process_streams.sol
ma2bd: Yes, the caller should always follow the convention of the callee. No exception

ma2bd: bonus: `StreamUpdate` didn't quite belong to the RPC formats in the first place

linera-service/tests/fixtures/evm_example_process_streams.sol
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) Zefchain Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.0;
+
+import "./Linera.sol";
+import "./LineraTypes.sol";
+
+contract ExampleExecuteMessage {
```
afck:
```suggestion
contract ExampleProcessStreams {
```

(Name is also wrong in `evm_example_empty_instantiate.sol`.

linera-execution/src/evm/revm.rs
```diff
@@ -1006,27 +1054,39 @@ where
...
+        vec: Vec<u8>,
     ) -> Result<ExecutionResultSuccess, ExecutionError> {
         // An application can be instantiated in Linera sense, but not in EVM sense,
         // that is the contract entries corresponding to the deployed contract may
         // be missing.
         if !self.db.is_initialized()? {
             self.initialize_contract()?;
         }
-        self.transact_commit(ch, vec)
+        self.transact_commit(Choice::Call, vec)
```
afck: (Not this PR, but:) We could rename `Choice` to e.g. `EvmTxKind` or `EvmCallTo`.

linera-execution/src/evm/mod.rs
```diff
@@ -33,11 +34,21 @@ pub enum EvmExecutionError {
...
+    #[error("Precompile error: {0}")]
+    PrecompileError(String),
     #[error("The operation was reverted with {gas_used} gas used and output {output:?}")]
     Revert {
         gas_used: u64,
         output: revm_primitives::Bytes,
     },
     #[error("The operation was halted with {gas_used} gas used due to {reason:?}")]
     Halt { gas_used: u64, reason: HaltReason },
+    #[error("The interpreter did not return, reason={reason:?}, gas_used={gas_used}, gas_refunded={gas_refunded}, logs={logs:?}, output={output:?}")]
```
afck: Line too long.

linera-execution/solidity/LineraTypes.sol
```diff
@@ -44,292 +43,313 @@ library LineraTypes {
...
         return (0,0);
     }
 
-    struct TimeoutConfig {
-        opt_TimeDelta fast_round_duration;
-        TimeDelta base_timeout;
-        TimeDelta timeout_increment;
-        TimeDelta fallback_duration;
+    struct AccountOwner {
+        uint8 choice;
```
afck, line 55: Maybe `tag`, `discriminant`, `account_type` or `owner_kind`?

And should this be an enum instead?

linera-execution/solidity/LineraTypes.sol
```diff
@@ -44,292 +43,313 @@ library LineraTypes {
...
+        // choice=0 corresponds to Reserved
+        uint8 reserved;
+        // choice=1 corresponds to Address32
+        CryptoHash address32;
+        // choice=2 corresponds to Address20
+        bytes20 address20;
```
afck, line 61: Would it be better to only have a single `bytes32` field and truncate that based on the enum discriminant?

linera-execution/solidity/LineraTypes.sol
```diff
@@ -44,292 +43,313 @@ library LineraTypes {
...
+        (new_pos, value) = bcs_deserialize_offset_ApplicationId(0, input);
         require(new_pos == input.length, "incomplete deserialization");
         return value;
     }
 
-    struct key_values_AccountOwner_uint64 {
-        AccountOwner key;
-        uint64 value;
+    struct BaseRuntimePrecompile {
+        uint8 choice;
```
afck, line 183: Same as above. I'd avoid the name "choice" for enums; in a sense, every enum is a choice.

linera-execution/solidity/LineraTypes.sol
```diff
@@ -44,292 +43,313 @@ library LineraTypes {
...
+        // choice=0 corresponds to ChainId
+        // choice=1 corresponds to ApplicationCreatorChainId
+        // choice=2 corresponds to ChainOwnership
+        // choice=3 corresponds to ReadDataBlob
+        BaseRuntimePrecompile_ReadDataBlob read_data_blob;
+        // choice=4 corresponds to AssertDataBlobExists
+        BaseRuntimePrecompile_AssertDataBlobExists assert_data_blob_exists;
```
afck, line 190: Those have the same field; would it make sense to just have a single `CryptoHash` field instead, so we don't waste memory?

linera-execution/solidity/LineraTypes.sol
```diff
@@ -370,264 +389,1360 @@ library LineraTypes {
...
+        // choice=3 corresponds to MessageId
+        // choice=4 corresponds to MessageIsBouncing
+        // choice=5 corresponds to Emit
+        ContractRuntimePrecompile_Emit emit_;
+        // choice=6 corresponds to ReadEvent
+        ContractRuntimePrecompile_ReadEvent read_event;
+        // choice=7 corresponds to SubscribeToEvents
+        ContractRuntimePrecompile_SubscribeToEvents subscribe_to_events;
+        // choice=8 corresponds to UnsubscribeFromEvents
+        ContractRuntimePrecompile_UnsubscribeFromEvents unsubscribe_from_events;
```
afck, line 494: These also have identical fields, and probably overlap with `read_event`…
In general, would it make sense to store everything apart from the discriminator as bytes in large tagged unions like this one, to avoid wasting memory?

linera-execution/solidity/LineraTypes.sol
```diff
@@ -370,264 +389,1360 @@ library LineraTypes {
...
+        internal
+        pure
+        returns (ContractRuntimePrecompile memory)
+    {
+        ContractRuntimePrecompile_TryCallApplication memory try_call_application;
+        ContractRuntimePrecompile_SendMessage memory send_message;
+        ContractRuntimePrecompile_Emit memory emit_;
+        ContractRuntimePrecompile_SubscribeToEvents memory subscribe_to_events;
+        ContractRuntimePrecompile_UnsubscribeFromEvents memory unsubscribe_from_events;
+        return ContractRuntimePrecompile(uint8(6), try_call_application, send_message, emit_, read_event, subscribe_to_events, unsubscribe_from_events);
```
afck, line 588: Instead of listing all the fields explicitly, would something like this work?

```sol
ContractRuntimePrecompile memory precompile;
precompile.choice = 6;
precompile.read_event = read_event;
return precompile;
```

linera-execution/solidity/LineraTypes.sol
afck, line 1: Wouldn't it actually be easier at this point to write a general BCS-for-EVM implementation?

linera-execution/solidity/LineraTypes.sol
MathieuDutSik: The only kind of enum existing in solidity are trivial ones. We are forced to use structs because of that.

linera-execution/solidity/LineraTypes.sol
MathieuDutSik: That is possible, however, this is automatically generated code, and so this is kind of complex to do.

linera-execution/solidity/LineraTypes.sol
MathieuDutSik: Yes, but it is computer-generated code, so yes, every enum is a choice and that is why it is used.

linera-execution/solidity/LineraTypes.sol
MathieuDutSik: Yes, but that would have to be coded in `serde-reflection` and the coding of the API would have to be done.

linera-execution/solidity/LineraTypes.sol
MathieuDutSik: Yes, this optimization can be put relatively easily in `serde-reflection`.

linera-execution/solidity/LineraTypes.sol
MathieuDutSik: I am not sure what you mean.
Yes, this has been done in `serde-reflection`. Do you mean writing EVM assembly?

linera-execution/solidity/LineraTypes.sol
MathieuDutSik: Yes, there are possibilities of improvement. It is not trivial to implement it by using `serde-reflection`.

linera-execution/solidity/LineraTypes.sol
afck: I know: I mean just instead of `uint8 choice` we could use `AccountOwnerTag tag` where `AccountOwnerTag` is an enum.

linera-execution/solidity/LineraTypes.sol
afck: Ah, sorry, I missed that!

linera-execution/solidity/LineraTypes.sol
afck: No, I hadn't realized this was already generated code.

linera-service/tests/fixtures/evm_example_process_streams.sol
MathieuDutSik: Thank you and sorry for this oversight.

linera-execution/solidity/LineraTypes.sol
afck: Where is the code that generates `LineraTypes.sol`?
Do we check in CI that it still matches?

linera-execution/src/evm/revm.rs
MathieuDutSik: We can rename, no problem with that.

linera-execution/src/evm/mod.rs
MathieuDutSik: Ok, reduced.

linera-service/tests/fixtures/evm_example_process_streams.sol
MathieuDutSik: Ok, done.

linera-execution/solidity/LineraTypes.sol
MathieuDutSik: No, we do not.
For that, we would need to have a release of `serde-reflection` that matches what is being used.
For sure, this can be done, but there are 100 more important priorities in my opinion.

linera-execution/solidity/LineraTypes.sol
MathieuDutSik: That makes sense, yes.
I will have to think about the right way to do it in `serde-reflection`, but yes, it can be done.

linera-execution/solidity/LineraTypes.sol
afck: I think we should do that. Let's discuss with the team.

IMO any auto-generated code that we commit should:
* be clearly marked as such in a comment at the beginning of the file,
* be generated with code that is easy to find, and
* be checked in CI.

PR #3962, opened 2025-05-19
Faster key value store view

### LINKED ISSUE #3933: Implement a `KeyValueStoreView` that uses only one key.
The `KeyValueStoreView` uses one key/value in the underlying disk storage for each key/value of the smart contract state.

This is a scalable storage method. However, this does not mean that it is efficient. For example, if the total storage used were less than 500K, then storing the smart contract key/values as a `BTreeMap<Vec<u8>,Vec<u8>>` being serialized would be faster.

This applies both to Wasm smart contracts and to EVM smart contracts (for which both keys and values are of length 32).

Typical key/value usage in Ethereum is about 50-500 key/values for DeFi apps, and 1000 for complex apps, so 64K maximum. Solana use cases appear smaller.

It would make sense to have a scalable solution that is also fast and would start by using a `BTreeMap<Vec<u8>,Vec<u8>>`. When passing a specific threshold, we would split the keys into individual values.
Notes:
* If the usage of the smart contract decreases below the threshold, it is not possible to simplify the key into one single key.
* There are alternatives like preloading the 20 most-used keys. However, that does not cover the fact that the use case that in 99% of the cases, we are using less than 500K and this is what we want to support very well.

MathieuDutSik:
## Motivation

The `KeyValueStoreView` is a scalable key-value store that is scalable. This means that if a smart contract has
4G of data in many small keys, then the computation of the hash will work nicely.

However, for many smart contracts, this is overkill: The total data being used is simply too small. This is especially
the case for EVM smart contracts where storage comes at a premium. Typically, they will use at most 1000 slots,
which makes 64K. Also, the fully scalable system is relatively slow.

So, for a small number of keys we want to have a `BTreeMap<Vec<u8>,Vec<u8>>` which becomes a scalable
system when the number of keys is large.

Fixes #3933 

## Proposal

The following is done:
* The existing `KeyValueStoreView` is renamed as `BigKeyValueStoreView`.
* A `SmallKeyValueStoreView` is introduced.
* A combining system `KeyValueStoreView` is introduced.

I considered introducing a trait for the specific functions of the `KeyValueStoreView`. I decided against because for the small variant, we can make most functions not async and this may be preferable.

## Test Plan

The CI should do it.

## Release Plan

## Links

None.


linera-views/src/views/key_value_store_view.rs
```diff
@@ -1164,45 +1115,1304 @@ where
...
+            None => true,
+            Some(SizeAndState {
+                total_size: _,
+                state,
+            }) => state.is_some(),
+        };
+        Ok(match is_small {
+            true => Self::Small(SmallKeyValueStoreView::post_load(context, values)?),
+            false => Self::Big(BigKeyValueStoreView::post_load(context, values)?),
+        })
```
ma2bd, line 1828: Please merge the two `match`s and remove `is_small`.

linera-views/src/views/key_value_store_view.rs
```diff
@@ -1164,45 +1115,1304 @@ where
...
+                    batch.put_key_value(key_total_size, &size_and_state)?;
+                    let mut big =
+                        BigKeyValueStoreView::post_load(small.context().clone(), &[None])?;
+                    big.stored_total_size = small.total_size;
+                    big.total_size = small.total_size;
+                    *self = Self::Big(big);
+                    Ok(true)
+                }
+            }
+            Self::Big(big) => big.flush(batch),
```
ma2bd, line 1882: Any chance we could ever go back to being small?

linera-views/src/views/key_value_store_view.rs
MathieuDutSik: Ok.

ma2bd:
Moving this to draft as we discussed just using `SmallKeyValueStore` standalone

MathieuDutSik:
Killed as it has been decided that the "big" design is not great:
* We have no scheme for reverting to "small" when that is possible.
- Making it possible would change the behavior of existing APIs, which might have undesirable effects. That could be done in the future, but right now it is not clear how the API could be changed intelligently.
- On the other hand, we do not know whether an application switching from small to big will revert to small.
* The change from "small" to "big" establishes a discontinuity of behavior that might be considered inadequate.
* The computation of the hash requires accessing the whole data structure. So, even if we use the "big" scheme, then we need to access to the whole set of keys.
- A counterargument to that is that there might be a more efficient way to compute the hash by computing a historical hash. That hash would be `sha256(prev_hash, batch)`. But does that hold with the existing protocol?
- Another idea is to reorganize the hash computation. We could compute the hash by prefixes like computing the hash of the keys `[x0, v]` with `x0` between 0 and 255. 

But those ideas are for the future, and so the PR is killed.

Issue #3969, opened 2025-05-20
Don't fully download all the sender chains.


afck:
We have the `previous_message_blocks` field in the block body now, so clients can verify that they have all messages from any given sender chain without downloading the whole chain.

afck:
I will add a `loose_blocks` map to the chain state that on the server will always be empty, and on the client contains verified (i.e. checked committee signatures) confirmed blocks that aren't necessarily contiguous and haven't been processed (executed and added to `confirmed_log`) yet.

We currently have `execute_block`, which only computes the outcome and modifies the _execution_ state, and `apply_confirmed_block`, which updates the tip state, outboxes and other chain state data.

A new `apply_loose_block` function will add a loose block and _only_ update the outboxes _where applicable_, without touching the tip state execution state or anything else.

Since it is now now guaranteed anymore that all outboxes have been updated up to the same height, we need an additional check: Only update the outbox if `previous_message_blocks` contains the hash of the block that `OutboxStateView::next_height_to_schedule` points to—the latter can be looked up in either the `confirmed_log` or the `loose_blocks`, depending on the height.

This also makes it necessary to _not_ remove empty outboxes anymore: We _always_ need to remember `next_height_to_schedule`.

afck:
To reduce the number of round trips I might also add a new endpoint for the proxy:
`DownloadSendingCertificates(CryptoHash, BlockHeight, ChainId)`
It returns the requested certificate and all those of its ancestors that sent something to the specified chain ID, down to the specified block height.

PR #3972, opened 2025-05-21
`metrics` submodules; other cleanups.

afck:
## Motivation

The Prometheus metrics are behind a feature flag. In some files there are dozens of conditionally compiled items, and the metrics-only imports are often far away from the places where they are used. This makes it hard to add imports in the right places and keep track of interactions between different feature flags.

We also call `process_outgoing_messages` separately for each transaction which could cause some redundant loading of outboxes. Probably not a big issue since views stay in memory, but still unnecessary, and I will need to change this anyway soon for https://github.com/linera-io/linera-protocol/issues/3969.

Finally, there are a few redundantly separated `impl` blocks, and we're inconsistent whether we fully qualify or import `try_join_all`.

## Proposal

* Create `metrics` submodules in several places, so only a single `#[cfg(with_metrics)]` attribute is needed to cover their items and imports. This reduces the total number of that attribute from 250 to 165.
* Pass the whole block into `process_outgoing_messages`: That function only needs all the recipients of messages sent by that block; it's irrelevant what transaction they originate from.
* Merge `impl` blocks.
* Fully qualify `try_join_all`.

## Test Plan

CI

## Release Plan

## Links

### DIFF
--- a/linera-chain/src/chain.rs
+++ b/linera-chain/src/chain.rs
@@ -1,8 +1,6 @@
 // Copyright (c) Zefchain Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-#[cfg(with_metrics)]
-use std::sync::LazyLock;
 use std::{
     collections::{BTreeMap, BTreeSet, HashMap, HashSet},
     sync::Arc,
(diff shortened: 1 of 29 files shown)


linera-chain/src/chain.rs
```diff
@@ -977,18 +977,9 @@ where
...
+        self.process_outgoing_messages(block).await?;
 
-        let recipients = block
-            .body
-            .messages
-            .iter()
-            .flatten()
-            .map(|message| message.destination)
-            .collect::<BTreeSet<_>>();
+        let recipients = block.recipients();
```
deuszx, line 974: This will now call `recipients` on the `block` twice:
- once in `process_outgoing_messages`
- second time here

vs previously only once. Not sure how important that is for performance 🤷 probably not that much but I wanted to point it out.

deuszx (approved):

linera-chain/src/chain.rs
afck: Done in https://github.com/linera-io/linera-protocol/pull/3972/commits/3502199f2be73719649e61f83b1e2a8fd0b5ceac.

PR #3973, opened 2025-05-21
Revm alloy version bump

MathieuDutSik:
## Motivation

We want to use the latest version of "alloy" and "revm" in Linera. Upgrading to the most recent proved
far more difficult than expected.

## Proposal

* Both alloy and revm are upgraded simultaneously since there is some issues with the package `c-kzg`.
* The `linera-ethereum` code is simplified by removing the `provider` entry that was not needed.
* The REVM is changed to the most recent API which is quite different from the one of 19.7.0

## Test Plan

The CI should do the job.

## Release Plan

## Links

None.


linera-ethereum/src/test_utils/mod.rs
```diff
@@ -30,12 +43,115 @@ sol!(
     "./contracts/EventNumerics.json"
 );
 
+#[allow(clippy::type_complexity)]
+#[derive(Clone)]
+pub struct EthereumClient {
+    pub provider: FillProvider<
```
deuszx, line 49: nit: Maybe extract to `pub type EthereumClientProvider = ` and reuse `EthereumClientProvider` here?

linera-ethereum/src/test_utils/mod.rs
```diff
@@ -132,12 +246,18 @@ pub struct EventNumericsContractFunction {
 impl EventNumericsContractFunction {
     pub async fn new(anvil_test: AnvilTest) -> anyhow::Result<Self> {
         // Deploying the event numerics contract
+        println!("EventNumericsContractFunction, new, step 1");
```
deuszx, line 249: leftovers

linera-ethereum/src/client.rs
```diff
@@ -3,13 +3,11 @@
...
-    primitives::{Address, Bytes, U256, U64},
-    rpc::types::eth::{
-        request::{TransactionInput, TransactionRequest},
-        BlockId, BlockNumberOrTag, Filter, Log,
-    },
+use alloy::rpc::types::eth::{
+    request::{TransactionInput, TransactionRequest},
+    BlockId, BlockNumberOrTag, Filter, Log,
 };
+use alloy_primitives::{Address, Bytes, U256, U64};
```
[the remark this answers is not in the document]
MathieuDutSik: You have two ways of accessing the code, either via `alloy::primitives` from `alloy` crate or via `alloy_primitives` from the `alloy-primitives` crate.

Two crates are widely used: `alloy-primitives` and `alloy-sol-types` (for creating complex type input). Those exist as standalone. The other stuff is in `alloy`.

The `alloy` crate contains the following functionality:
* The anvil test framework.
* Working with contracts and interacting with them.
* Tracking keys, nonce, gas filling, and other stuff.

linera-ethereum/src/test_utils/mod.rs
MathieuDutSik: The problem is that this prevents implementing a new operator. Because we would be implementing function for a function defined out of crate.

linera-ethereum/src/test_utils/mod.rs
MathieuDutSik: Yes, cleaned out.

deuszx (approved):

Issue #3974, opened 2025-05-21
Replace `persistent` with `linera-views`


Twey:
Currently we are using the `persistent` library to store client configuration to disk on the filesystem or in an IndexedDB database on the Web.  `persistent` is lightweight and supports writing directly to a JSON file, preserving backwards compatibility with the existing code that wrote a JSON file manually, but writing a JSON file on every state change scales poorly, and the functionality of `persistent` is a subset of that of `linera-views`.

Instead, we should switch to a database file (SQLite? RocksDB?) for filesystem storage, and use `linera-views` for both that and IndexedDB.

PR #3978, opened 2025-05-21
Move the wallet into `linera-core`

Twey:
## Motivation

Currently we store the wallet (information about the last-seen chain state) in two different places: in-memory in `linera-core` and then duplicated by the chain listener onto disk whenever something happens.  This means that we duplicate logic and also that the two copies of the state can get out of sync.

## Proposal

Move the wallet handling directly into `linera-core`, using dependency inversion (over the `Persist` trait, though we should [replace it with `linera-views`](https://github.com/linera-io/linera-protocol/issues/3974) in the future) to keep `linera-core` pure and testable.  That way the wallet can become the source of truth for (its relevant portions of) chain state.

## Test Plan

CI.

## Release Plan

## Links


Twey:
Superseded by https://github.com/linera-io/linera-protocol/pull/5080.

PR #3988, opened 2025-05-22
Change the behavior of LogView read function

MathieuDutSik:
## Motivation

The function `read` of the `LogView` is restricting the range on input. That behavior is at odds with the one of other containers where we return an error on incorrect input. Moreover, there is no use case for handling incorrect range on input.

## Proposal

An incorrect range error is returned when the input range is incorrect.

## Test Plan

The CI. The tests have to be addressed for the change of API.

## Release Plan

## Links

Help.

### DIFF
--- a/linera-service/src/proxy/grpc.rs
+++ b/linera-service/src/proxy/grpc.rs
@@ -333,7 +333,7 @@ where
     /// Returns the appropriate gRPC status for the given [`ViewError`].
     fn error_to_status(err: ViewError) -> Status {
         let mut status = match &err {
-            ViewError::TooLargeValue | ViewError::BcsError(_) => {
+            ViewError::TooLargeValue | ViewError::BcsError(_) | ViewError::IncorrectRange => {
                 Status::invalid_argument(err.to_string())
             }
             ViewError::StoreError { .. }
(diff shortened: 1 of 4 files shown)


linera-service/src/proxy/grpc.rs
```diff
@@ -333,7 +333,7 @@ where
     /// Returns the appropriate gRPC status for the given [`ViewError`].
     fn error_to_status(err: ViewError) -> Status {
         let mut status = match &err {
-            ViewError::TooLargeValue | ViewError::BcsError(_) => {
+            ViewError::TooLargeValue | ViewError::BcsError(_) | ViewError::IncorrectRange => {
                 Status::invalid_argument(err.to_string())
             }
             ViewError::StoreError { .. }
```
ma2bd, line 339: Afaict the GRPC status INVALID_ARGUMENT doesn't apply to any `ViewError`: https://grpc.io/docs/guides/status-codes/

ma2bd: In fact all `ViewError`s are INTERNAL_ERROR except maybe those that will be moved to `StorageError` and capture a NOT_FOUND concept

ma2bd (changes requested):
I don't really see why the current API is problematic. When ranges can be open, the underlying semantic is necessarily the intersection of existing indices with the range.

MathieuDutSik:
This PR has not reached consensus.
When querying more values than available, it would have made sense to report an error or null entries.
But that is not the consensus here.
So, the PR is killed.

Issue #3270, opened 2025-02-06
Leave out the execution step in sender chains.


afck:
When downloading sender chains, omit the execution step. Only verify and update the outboxes.

Sonioskoyi:
Good luck

PR #3989, opened 2025-05-22
Don't execute sender chains

afck:
## Motivation

When the client verifies incoming messages to its own chains, it doesn't need the sender chain's execution state, so computing the block execution there is unnecessary work.

Eventually we want to go even further and avoid even downloading some of the sender chain's blocks (see https://github.com/linera-io/linera-protocol/issues/3969).

## Proposal

As a preparation for #3969, we introduce an `unexecuted_blocks` map to the chain state: It maps block heights that are higher than `next_block_height`, i.e. haven't been executed yet, to block hashes.

We don't purge empty outboxes anymore. When updating the outboxes, we make sure that we haven't skipped any messages and compare to the next height to schedule: That allows us to update some outboxes even for loose blocks, ahead of execution.

For now we still download all blocks of the sender chain. Skipping any will be done in a future PR.
However, we now add them as loose blocks and don't execute them.

## Test Plan

CI

## Release Plan

## Links

- Closes https://github.com/linera-io/linera-protocol/issues/3270.

### DIFF
--- a/linera-base/src/data_types.rs
+++ b/linera-base/src/data_types.rs
@@ -13,6 +13,7 @@ use std::{
     hash::Hash,
     io, iter,
     num::ParseIntError,
+    ops::{Bound, RangeBounds},
     path::Path,
     str::FromStr,
 };
@@ -490,6 +491,32 @@ impl TryFrom<BlockHeight> for usize {
 impl From<u64> for BlockHeight {
     fn from(value: u64) -> Self {
         Self(value)
+    }
+}
+
+/// Allows converting [`BlockHeight`] ranges to inclusive tuples of bounds.
+pub trait RangeExt {
+    /// Returns the range as a tuple of inclusive bounds.
+    /// If the range is empty, returns `None`.
+    fn to_inclusive(&self) -> Option<(BlockHeight, BlockHeight)>;
+}
+
+impl<T: RangeBounds<BlockHeight>> RangeExt for T {
+    fn to_inclusive(&self) -> Option<(BlockHeight, BlockHeight)> {
+        let start = match self.start_bound() {
+            Bound::Included(height) => *height,
+            Bound::Excluded(height) => height.try_add_one().ok()?,
+            Bound::Unbounded => BlockHeight(0),
+        };
+        let end = match self.end_bound() {
+            Bound::Included(height) => *height,
+            Bound::Excluded(height) => height.try_sub_one().ok()?,
+            Bound::Unbounded => BlockHeight::MAX,
+        };
+        if start > end {
+            return None;
+        }
+        Some((start, end))
     }
 }
 
--- a/linera-base/src/data_types.rs
+++ b/linera-base/src/data_types.rs
@@ -488,13 +488,13 @@ impl TryFrom<BlockHeight> for usize {
 }
 
 /// Allows converting [`BlockHeight`] ranges to inclusive tuples of bounds.
-pub trait RangeExt {
+pub trait BlockHeightRangeBounds {
     /// Returns the range as a tuple of inclusive bounds.
     /// If the range is empty, returns `None`.
     fn to_inclusive(&self) -> Option<(BlockHeight, BlockHeight)>;
 }
 
-impl<T: RangeBounds<BlockHeight>> RangeExt for T {
+impl<T: RangeBounds<BlockHeight>> BlockHeightRangeBounds for T {
     fn to_inclusive(&self) -> Option<(BlockHeight, BlockHeight)> {
         let start = match self.start_bound() {
             Bound::Included(height) => *height,
--- a/linera-chain/src/chain.rs
+++ b/linera-chain/src/chain.rs
@@ -280,7 +280,7 @@ where
     pub outbox_counters: RegisterView<C, BTreeMap<BlockHeight, u32>>,
 
     /// Blocks that have been verified but not executed yet, and that may not be contiguous.
-    pub unexecuted_blocks: MapView<C, BlockHeight, CryptoHash>,
+    pub other_blocks: MapView<C, BlockHeight, CryptoHash>,
 }
 
 /// Block-chaining state.
@@ -943,12 +943,12 @@ where
             &block.body.messages,
         )?;
         self.confirmed_log.push(hash);
-        self.unexecuted_blocks.remove(&block.header.height)?;
+        self.other_blocks.remove(&block.header.height)?;
         Ok(())
     }
 
     /// Applies a block without executing it. This only updates the outboxes.
-    pub async fn apply_unexecuted_block(
+    pub async fn process_block_without_executing(
         &mut self,
         block: &ConfirmedBlock,
     ) -> Result<(), ChainError> {
@@ -958,7 +958,7 @@ where
             return Ok(());
         }
         self.process_outgoing_messages(block).await?;
-        self.unexecuted_blocks.insert(&block.header.height, hash)?;
+        self.other_blocks.insert(&block.header.height, hash)?;
         Ok(())
     }
 
@@ -1104,7 +1104,7 @@ where
         };
         for height in start.max(next_height).0..=end.0 {
             hashes.push(
-                self.unexecuted_blocks
+                self.other_blocks
                     .get(&BlockHeight(height))
                     .await?
                     .ok_or_else(|| {
@@ -1160,7 +1160,7 @@ where
                         })?)
                     }
                     Some(height) => {
-                        Some(self.unexecuted_blocks.get(&height).await?.ok_or_else(|| {
+                        Some(self.other_blocks.get(&height).await?.ok_or_else(|| {
                             ChainError::InternalError("missing entry in unexecuted_blocks".into())
                         })?)
                     }
--- a/linera-chain/src/chain.rs
+++ b/linera-chain/src/chain.rs
@@ -280,7 +280,7 @@ where
     pub outbox_counters: RegisterView<C, BTreeMap<BlockHeight, u32>>,
 
     /// Blocks that have been verified but not executed yet, and that may not be contiguous.
-    pub other_blocks: MapView<C, BlockHeight, CryptoHash>,
+    pub preprocessed_blocks: MapView<C, BlockHeight, CryptoHash>,
 }
 
 /// Block-chaining state.
@@ -943,22 +943,20 @@ where
             &block.body.messages,
         )?;
         self.confirmed_log.push(hash);
-        self.other_blocks.remove(&block.header.height)?;
+        self.preprocessed_blocks.remove(&block.header.height)?;
         Ok(())
     }
 
-    /// Applies a block without executing it. This only updates the outboxes.
-    pub async fn process_block_without_executing(
-        &mut self,
-        block: &ConfirmedBlock,
-    ) -> Result<(), ChainError> {
+    /// Adds a block to `preprocessed_blocks`, and updates the outboxes where possible.
+    pub async fn preprocess_block(&mut self, block: &ConfirmedBlock) -> Result<(), ChainError> {
         let hash = block.inner().hash();
         let block = block.inner().inner();
-        if block.header.height < self.tip_state.get().next_block_height {
+        let height = block.header.height;
+        if height < self.tip_state.get().next_block_height {
             return Ok(());
         }
         self.process_outgoing_messages(block).await?;
-        self.other_blocks.insert(&block.header.height, hash)?;
+        self.preprocessed_blocks.insert(&height, hash)?;
         Ok(())
     }
 
@@ -1104,11 +1102,11 @@ where
         };
         for height in start.max(next_height).0..=end.0 {
             hashes.push(
-                self.other_blocks
+                self.preprocessed_blocks
                     .get(&BlockHeight(height))
                     .await?
                     .ok_or_else(|| {
-                        ChainError::InternalError("missing entry in unexecuted_blocks".into())
+                        ChainError::InternalError("missing entry in preprocessed_blocks".into())
                     })?,
             );
         }
@@ -1159,11 +1157,9 @@ where
                             ChainError::InternalError("missing entry in confirmed_log".into())
                         })?)
                     }
-                    Some(height) => {
-                        Some(self.other_blocks.get(&height).await?.ok_or_else(|| {
-                            ChainError::InternalError("missing entry in unexecuted_blocks".into())
-                        })?)
-                    }
+                    Some(height) => Some(self.preprocessed_blocks.get(&height).await?.ok_or_else(
+                        || ChainError::InternalError("missing entry in preprocessed_blocks".into()),
+                    )?),
                     None => None,
                 };
                 // Schedule only if no block is missing that sent something to the same recipient.
--- a/linera-chain/src/chain.rs
+++ b/linera-chain/src/chain.rs
@@ -277,6 +277,9 @@ where
     /// Number of outgoing messages in flight for each block height.
     /// We use a `RegisterView` to prioritize speed for small maps.
     pub outbox_counters: RegisterView<C, BTreeMap<BlockHeight, u32>>,
+
+    /// Blocks that have been verified but not processed yet, and that may not be contiguous.
+    pub loose_blocks: MapView<C, BlockHeight, CryptoHash>,
 }
 
 /// Block-chaining state.
@@ -418,9 +421,6 @@ where
                 // Important for the test in `all_messages_delivered_up_to`.
                 self.outbox_counters.get_mut().remove(&update);
             }
-        }
-        if outbox.queue.count() == 0 {
-            self.outboxes.remove_entry(target)?;
         }
         #[cfg(with_metrics)]
         metrics::NUM_OUTBOXES
@@ -942,6 +942,7 @@ where
             &block.body.messages,
         )?;
         self.confirmed_log.push(hash);
+        self.loose_blocks.remove(&block.header.height)?;
         Ok(())
     }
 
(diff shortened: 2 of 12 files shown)


linera-core/src/client/mod.rs
```diff
@@ -1971,7 +2073,10 @@ impl<Env: Environment> ChainClient<Env> {
                 for certificate in certificates.into_values() {
                     let hash = certificate.hash();
                     let mode = ReceiveCertificateMode::AlreadyChecked;
-                    if let Err(e) = client.receive_certificate(certificate, mode, None).await {
+                    if let Err(e) = client
+                        .receive_loose_certificate(certificate, mode, None)
+                        .await
+                    {
                         warn!("Received invalid certificate {hash}: {e}");
```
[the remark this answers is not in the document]
afck: Good point; since the signatures have already been checked this should really only fail if a quorum of validators is faulty. I'll make it an `error!`.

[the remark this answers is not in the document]
afck: Although… we haven't checked if they're from the right chain ID. We probably should. Will add that…

linera-chain/src/chain.rs
```diff
@@ -942,6 +943,22 @@ where
             &block.body.messages,
         )?;
         self.confirmed_log.push(hash);
+        self.unexecuted_blocks.remove(&block.header.height)?;
+        Ok(())
+    }
+
+    /// Applies a loose block without executing it. This only updates the outboxes.
```
ma2bd: There are a few instance of `loose` left in the comments here and below.

linera-chain/src/chain.rs
```diff
@@ -277,6 +278,9 @@ where
     /// Number of outgoing messages in flight for each block height.
     /// We use a `RegisterView` to prioritize speed for small maps.
     pub outbox_counters: RegisterView<C, BTreeMap<BlockHeight, u32>>,
+
+    /// Blocks that have been verified but not executed yet, and that may not be contiguous.
```
ma2bd, line 282: Looks like we also execute the outgoing messages for those blocks. Right?

ma2bd: It seems to me that the blocks are "partially executed" instead of "unexecuted".

linera-core/src/chain_worker/state/attempted_changes.rs
```diff
@@ -415,6 +415,53 @@ where
         Ok((info, actions))
     }
 
+    /// Processes a loose block without executing it.
```
ma2bd: Partially execute a confirmed block?

linera-core/src/client/mod.rs
```diff
@@ -2860,18 +2976,33 @@ impl<Env: Environment> ChainClient<Env> {
...
+            .await
+    }
+
+    /// Processes loose certificate for which this chain is a recipient.
+    #[instrument(
+        level = "trace",
+        skip(certificate),
+        fields(certificate_hash = ?certificate.hash()),
+    )]
+    async fn receive_unexecuted_certificate(
```
ma2bd: It's not "unexecuted" just yet here, it is?
Maybe `receive_sender_certificate` or `receive_certificate_for_partial_execution` ?

linera-core/src/local_node.rs
```diff
@@ -308,8 +321,12 @@ where
         let futures = chain_ids
             .into_iter()
             .map(|chain_id| async move {
-                let local_info = self.chain_info(*chain_id).await?;
-                Ok::<_, LocalNodeError>((*chain_id, local_info.next_block_height))
+                let chain = self.chain_state_view(*chain_id).await?;
+                let mut next_height = chain.tip_state.get().next_block_height;
+                while chain.unexecuted_blocks.contains_key(&next_height).await? {
+                    next_height.try_add_assign_one()?;
+                }
```
ma2bd: This is a bit scary in terms of performance but maybe ok for now

linera-chain/src/chain.rs
```diff
@@ -1094,12 +1140,35 @@ where
...
+                    }
+                    Some(height) => {
+                        Some(self.unexecuted_blocks.get(&height).await?.ok_or_else(|| {
+                            ChainError::InternalError("missing entry in unexecuted_blocks".into())
+                        })?)
+                    }
+                    None => None,
+                };
+                // Schedule only if no block is missing that sent something to the same recipient.
+                if prev_hash.as_ref() != block.body.previous_message_blocks.get(target) {
```
ma2bd, line 1166: We cannot mark the block as partially executed then, do we?

linera-chain/src/chain.rs
```diff
@@ -1113,6 +1182,31 @@ where
     }
 }
 
+trait RangeExt {
+    /// Returns the range as a tuple of inclusive bounds.
+    /// If the range is empty, returns `None`.
+    fn to_inclusive(&self) -> Option<(BlockHeight, BlockHeight)>;
+}
+
+impl<T: RangeBounds<BlockHeight>> RangeExt for T {
```
ma2bd: maybe this belong to `linera-base`?

linera-chain/src/chain.rs
afck: I'd say "executing" a block means updating the chain's execution state. After https://github.com/linera-io/linera-protocol/pull/3695 that's also the only thing `ChainStateView::execute_block` actually does.

The separate `apply_confirmed_block` updates outboxes, state hash and chain manager, which are outside of the execution state and don't have to be in sync with it.

Maybe we need a different word for `apply`, but I'd say these blocks are all unexecuted.

linera-core/src/local_node.rs
afck: I think we'll need to add something like a `MapView::indices_in_range` method that returns all the block heights in a given range.

linera-chain/src/chain.rs
afck: Originally I had called this collection `loose_blocks` (see renaming discussion), and it was meant to list the blocks in storage that we haven't applied to the execution state yet.

What `process_outgoing_messages` is doing now is: update all outboxes where possible, given the set of blocks we do have. (Currently that will always be all of them, but when the blocks are actually sparse, it won't.)
For some recipients (like the one that caused us to download the sender chain) we will update the outbox because we have all the necessary ancestor blocks. For others we won't, and possibly never will.

Anyway, this collection was never meant to mark anything as "partially executed". It only points to sparse blocks that are later than the current execution state, and which _may_ have partially applied to the outboxes. `process_outgoing_messages` is also idempotent now, so there's no need to store other information about the progress.

linera-core/src/client/mod.rs
afck: I renamed it to `receive_sender_certificate`… although that's not correct for all cases, since we'll also download certificates to verify blobs.

linera-chain/src/chain.rs
ma2bd: If a block is skipped here between of missing sparse-ancestors, then we provide the missing blocks later, do we need to "apply" the initial block again?

afck: Yes, we do.
So if we have two tracked chains receiving messages from the same sender chain, we'll apply the blocks that send to both twice.

linera-chain/src/chain.rs
afck: Done in https://github.com/linera-io/linera-protocol/pull/3989/commits/bb4f788166ee59a12f9be8ac72dc78ac3b00a7bf.

linera-core/src/local_node.rs
Twey: Maybe a TODO in case we forget?

linera-chain/src/chain.rs
```diff
@@ -942,6 +943,22 @@ where
             &block.body.messages,
         )?;
         self.confirmed_log.push(hash);
+        self.unexecuted_blocks.remove(&block.header.height)?;
```
deuszx: I'm bit confused - this collection is called _unexecuted blocks_ and we remove a block from it but it's still NOT executed.

linera-chain/src/chain.rs
```diff
@@ -942,6 +943,22 @@ where
...
+        Ok(())
+    }
+
+    /// Applies an unexecuted block without executing it. This only updates the outboxes.
```
deuszx: Since _apply_ is a bit overloaded (and undefined) term maybe we can just leave the second sentence of the comment?

"Processes outgoing messages and adds the block to the unexecuted_blocks collection"

Also, it would be nice to explain here why this method even exist (i.e. when processing outgoing messages w/o executing the block is useful) but I don't know yet how to put that into words nicely.

linera-chain/src/chain.rs
```diff
@@ -1067,6 +1084,35 @@ where
         Ok(())
     }
 
+    pub async fn block_hashes(
```
deuszx, line 1087:
```suggestion
    /// Returns a vector of continuous and inclusive (on both ends) unexecuted block hashes for the given block height range.
    /// Returns an error if we're missing block for any of the block heights.
    pub async fn unexecuted_block_hashes(
```
?

linera-chain/src/chain.rs
afck: We only call `apply_confirmed_block` after executing it; maybe this function should now be called `apply_executed_block`.

linera-base/src/data_types.rs
```diff
@@ -486,6 +487,32 @@ impl TryFrom<BlockHeight> for usize {
     }
 }
 
+/// Allows converting [`BlockHeight`] ranges to inclusive tuples of bounds.
+pub trait RangeExt {
```
ma2bd: nit: Either `RangeExt<T>: RangeBounds<T>` then replace `BlockHeight` by `T` OR rename `RangeExt` into `BlockHeightRangeExt` or `BlockHeightRangeBounds`.

linera-chain/src/chain.rs
ma2bd: `process` ?

linera-chain/src/chain.rs
ma2bd: The negation in `unexecuted` is not ideal. Better call them `applied_blocks` or `processed_blocks` IMHO so the invariant is clear (e.g. "blocks that have been applied")

linera-core/src/client/mod.rs
ma2bd: I still see `receive_unexecuted_certificate` here which is not good.

linera-core/src/worker.rs
```diff
@@ -880,6 +890,45 @@ where
...
 
+    /// Applies an unexecuted block without executing it. This does not update the execution state, but
+    /// can create cross-chain messages and store blobs. It also does _not_ check the signatures;
+    /// the caller is responsible for checking them using the correct committee.
+    #[instrument(skip_all, fields(
+        nick = self.nickname,
+        chain_id = format!("{:.8}", certificate.block().header.chain_id),
+        height = %certificate.block().header.height,
+    ))]
+    pub async fn fully_process_unexecuted_certificate_with_notifications(
```
ma2bd: The "unexecuted" concept is not ideal. This is visible in the sentence: "Applies an unexecuted block without executing it."

What's happening is:
* we have a (normal confirmed) block
* we choose not to execute it for some reason (e.g. we only care about messages and events created by this block)
* instead we only apply/process/XXX it
* it becomes an applied/processed/XXXed block

linera-base/src/data_types.rs
afck: Unfortunately the standard library has no trait for types with a minimum and maximum value, so we can't make it generic. ~Maybe just `BlockHeightRange` then?~ (Edit: name collision!) I'll call it `BlockHeightRangeBounds`. :+1:

linera-core/src/worker.rs
afck: Yes, but once the chain is actually sparse, the block won't even be "fully applied/processed/XXXed": We might do that again (_still_ without executing it) to add more messages from the same block to the outbox later. And since the only things we actually do with it—adding messages to the outbox, and storing blobs—are idempotent, it's not that important that we remember that it was partially processed.

It matters a bit more that these blocks have not been executed yet, but on the other hand, that's obvious from the `next_block_height`. So this map view really just serves as an index of other stored blocks (apart from the ones in `confirmed_log`, which are exactly the executed ones).

Maybe just `other_blocks` then? And I'll change the method names and comments (I had just use find-and-replace to go from "loose block" to "unexecuted block", but that doesn't make sense in some places.)

linera-chain/src/chain.rs
afck: It returns both executed and unexecuted ones. I'll update the comment.

linera-core/src/local_node.rs
afck: I don't know yet what exactly we'll need instead of this, but I added a TODO pointing to #3969 anyway.

linera-chain/src/chain.rs
afck: I called them `other_blocks` for now, but we can discuss alternatives. If you prefer `applied_blocks` or `processed_blocks` I can rename it, but in a sense they're only _partially_ applied/processed, at least when it comes to the outbox.

linera-chain/src/chain.rs
```diff
@@ -1067,6 +1084,37 @@ where
...
+            let usize_start = usize::try_from(start)?;
+            let usize_end = usize::try_from(end.min(last_height))?;
+            self.confirmed_log.read(usize_start..=usize_end).await?
+        } else {
+            Vec::new()
+        };
+        for height in start.max(next_height).0..=end.0 {
+            hashes.push(
+                self.other_blocks
+                    .get(&BlockHeight(height))
```
deuszx, line 1106: In theory, `height` here could have already pushed a block hash to `hashes` in the previous iteration - from `self.confirmed_log` - right?

linera-chain/src/chain.rs
```diff
@@ -1094,12 +1142,35 @@ where
...
         let recipients = block.recipients();
         let block_height = block.header.height;
+        let next_height = self.tip_state.get().next_block_height;
 
         // Update the outboxes.
         let outbox_counters = self.outbox_counters.get_mut();
         let targets = recipients.into_iter().collect::<Vec<_>>();
         let outboxes = self.outboxes.try_load_entries_mut(&targets).await?;
-        for mut outbox in outboxes {
+        for (mut outbox, target) in outboxes.into_iter().zip(&targets) {
```
deuszx, line 1149: Just a note from the reviewer's perspective:

You added couple dozens of non-trivial code here to a function, without any explanation or "introduction". To be honest I have no idea what to expect - what I'm reading. I have to read it, reverse-engineer its purpose and then try to understand/come up WHY this is here specifically.

It'd be much easier if this loop was preceded with a short comment.

linera-chain/src/chain.rs
```diff
@@ -1094,12 +1142,35 @@ where
...
+                            ChainError::InternalError("missing entry in unexecuted_blocks".into())
+                        })?)
+                    }
+                    None => None,
+                };
+                // Schedule only if no block is missing that sent something to the same recipient.
+                if prev_hash.as_ref() != block.body.previous_message_blocks.get(target) {
+                    continue;
+                }
+            }
```
deuszx, line 1169: This whole loop is like a guard, checking if we should execute the 2 lines below it. Can we extract to a function? Then it could be as simple as:
```rust
outboxes.into_iter().zip(&targets).filter(<guard>).for_each(<schedule>)
```

linera-core/src/chain_worker/state/attempted_changes.rs
```diff
@@ -414,6 +414,53 @@ where
...
+
+        let required_blob_ids = block.required_blob_ids();
+        let created_blobs: BTreeMap<_, _> = block.iter_created_blobs().collect();
+        let blobs_result = self
+            .state
+            .get_required_blobs(block.required_blob_ids(), &created_blobs)
+            .await
+            .map(|blobs| blobs.into_values().collect::<Vec<_>>());
+
+        if let Ok(blobs) = &blobs_result {
```
deuszx, line 439: What hapens if `blob_result` is `Err`? i.e. that we have some blobs missing. You don't seem to report that to the caller.

linera-core/src/chain_worker/state/attempted_changes.rs
```diff
@@ -414,6 +414,53 @@ where
...
+        if tip.next_block_height > block.header.height {
+            // We already processed this block.
+            return self.state.create_network_actions().await;
+        }
+
+        let required_blob_ids = block.required_blob_ids();
+        let created_blobs: BTreeMap<_, _> = block.iter_created_blobs().collect();
+        let blobs_result = self
+            .state
+            .get_required_blobs(block.required_blob_ids(), &created_blobs)
```
deuszx, line 435: You already have the `required_blob_ids` above 👀

afck:
As discussed I renamed it to `preprocessed_blocks`.

linera-core/src/client/mod.rs
```diff
@@ -336,6 +336,22 @@ impl<Env: Environment> Client<Env> {
         stop: BlockHeight,
     ) -> Result<Option<Box<ChainInfo>>, ChainClientError> {
         let mut last_info = None;
+        // First load any blocks from local storage, if available.
+        let mut hashes = Vec::new();
+        {
+            let chain = self.local_node.chain_state_view(chain_id).await?;
+            while start < stop {
```
bart-linera: Should we update `start` here? To something like max(start, current_chain_tip)? (After all, if `start` is lower than the tip, we should have everything we need in between, and the getting of preprocessed blocks below will fail, because the block at `start` is already processed, not just preprocessed.)

afck: Hmm… we actually don't even need to pass in the next block height, since we always want to start from the current tip.

~And we shouldn't return a chain info, because the caller is interested in the highest _downloaded_ block, not the `next_block_height`. I'll fix that.~ (Edit: No, in this function we _do_ fully process certificates.)

deuszx (commented):
I had an in-progress review that I didn't finish

linera-core/src/chain_worker/state/attempted_changes.rs
afck: We do `blobs_result?` below. We just need to make sure we store the blob states first, so we accept the blobs if we receive them.

linera-chain/src/chain.rs
afck: I tried, but the filter function would not only have to be `async`, but also use `&self` as well as `&outbox`. The problem is that `self` is already borrowed mutably, so that doesn't work.

deuszx: Thanks for checking

linera-chain/src/chain.rs
afck: No, `usize_end` is at most `next_height - 1`, so the two ranges are disjoint.

PR #3992, opened 2025-05-22
Fix `dynamodb` tests by disabling `aws_lc` crypto

Twey:
## Motivation

Recent package upgrades led to both the `ring` and `aws_lc_rs` features being enabled for `rustls`, which therefore fails to pick a sensible default.

## Proposal

Disable the features in the AWS packages that cause `rustls/aws_lc_rs` to be enabled, so that `rustls` defaults to `ring`.

## Test Plan

CI.

## Release Plan

## Links

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -83,8 +83,6 @@ async-graphql-derive = "=7.0.2"
 async-lock = "3.3.0"
 async-trait = "0.1.77"
 async-tungstenite = { version = "0.22", features = ["tokio-runtime"] }
-aws-config = "1.6.3"
-aws-sdk-dynamodb = "1.76.0"
 aws-smithy-types = "1.3.1"
 axum = "0.7.4"
 base64 = "0.22.0"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -317,6 +315,15 @@ native-fungible = { path = "./examples/native-fungible" }
...
 social = { path = "./examples/social" }
 
+[workspace.dependencies.aws-config]
+default-features = false
+version = "1.6.3"
+
+[workspace.dependencies.aws-sdk-dynamodb]
+default-features = false
+features = ["rt-tokio", "rustls"]
+version = "1.76.0"
```
deuszx, line 325: Why this way vs the usual inline?

deuszx (approved):
no blockers

MathieuDutSik (approved):

Cargo.toml
Twey: I had a lot more options here at some point and the inline was pretty unreadable.

`cargo publish` normalizes out the inlines anyway :)

PR #3994, opened 2025-05-22
Faucet metrics

christos-h:
## Motivation

The current operation of the faucet is opaque which it makes it hard to debug and understand it's behaviour.

## Proposal

Give the faucet the capability to export prometheus metrics.

## Test Plan

Tested manually.


linera-faucet/server/build.rs
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) Zefchain Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+fn main() {
+    cfg_aliases::cfg_aliases! {
+        with_metrics: { all(not(target_arch = "wasm32"), feature = "metrics") },
+    };
+}
```
Twey, line 8: This doesn't seem necessary — this crate should never be built for `wasm32`.

linera-service/Cargo.toml
```diff
@@ -87,6 +92,7 @@ linera-core.workspace = true
 linera-execution = { workspace = true, features = ["fs"] }
 linera-faucet-client.workspace = true
 linera-faucet-server.workspace = true
+linera-metrics.workspace = true
```
afck: Shouldn't this be an optional dependency?

linera-faucet/server/Cargo.toml
```diff
@@ -11,6 +11,16 @@ homepage.workspace = true
 license.workspace = true
 edition.workspace = true
 
+[package.metadata.cargo-machete]
+ignored = ["prometheus"]
+
+
```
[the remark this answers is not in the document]
afck: (Still two lines?)

linera-faucet/server/src/lib.rs
```diff
@@ -211,36 +217,43 @@ where
...
+            config: config.chain_listener_config,
             storage,
-            port,
-            amount,
-            end_timestamp,
+            port: config.port,
+            #[cfg(feature = "metrics")]
+            metrics_port: config.metrics_port,
+            amount: config.amount,
+            end_timestamp: config.end_timestamp,
```
afck, line 256: Would it make sense for the faucet to just contain a single field `config` then?

linera-faucet/server/Cargo.toml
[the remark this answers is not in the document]
christos-h: Sorry!  :man_facepalming:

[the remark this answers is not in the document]
christos-h: Done.

ndr-ds (changes requested):
Overall lgtm, I agree with the recent comments from @afck though, so back to you for those

linera-faucet/server/src/lib.rs
christos-h: I don't think so? There are fields like `context`, `storage`, `start_timestamp`  and `start_balance` of which some are generic.

I could put the config object as a field in the Faucet struct along with the other fields but then you have another layer of indirection (`self.config.chain_id` for example).

ndr-ds: I personally think the extra layer of indirection is fine 🤔 we have that same thing in other places of the code, and it makes `FaucetService` look cleaner IMHO

afck: Either way is fine with me. Not a blocker from my side!

PR #3995, opened 2025-05-22
Account for published blobs in the resource at the end of transaction execution rather than upfront.

deuszx:
## Motivation

We are charging for all published blobs before we execute any operation in the block. This is wrong b/c we want the incoming grant to be able to cover the fees. The grants are processed during operation processing only so they are not taken into account at the beginning of the block processing.

## Proposal

Move the published blobs' resources tracking to the end of processing a transaction that published it. At this point any incoming grant is already processed. In order to do so, we track blob IDs referenced during transaction execution via `TransactionTracker.published_blobs` and consume it in `BlockExecutionTracker::process_txn_outcome`.

Also, charge for publishing user-created blobs only. `BlobType::Committee` is not accounted for.

## Test Plan

CI.

## Release Plan

## Links

### DIFF
--- a/linera-base/src/data_types.rs
+++ b/linera-base/src/data_types.rs
@@ -1375,9 +1375,9 @@ impl Blob {
         Ok(Self::new_data(fs::read(path)?))
     }
 
-    /// Returns whether the blob is a "user blob".
-    pub fn is_user_blob(&self) -> bool {
-        self.content().blob_type().is_user_blob()
+    /// Returns whether the blob is of [`BlobType::Committee`] variant.
+    pub fn is_committee_blob(&self) -> bool {
+        self.content().blob_type().is_committee_blob()
     }
 }
 
--- a/linera-base/src/data_types.rs
+++ b/linera-base/src/data_types.rs
@@ -1374,6 +1374,11 @@ impl Blob {
     pub async fn load_data_blob_from_file(path: impl AsRef<Path>) -> io::Result<Self> {
         Ok(Self::new_data(fs::read(path)?))
     }
+
+    /// Returns whether the blob is a "user blob".
+    pub fn is_user_blob(&self) -> bool {
+        self.content().blob_type().is_user_blob()
+    }
 }
 
 impl Serialize for Blob {
--- a/linera-chain/src/block_tracker.rs
+++ b/linera-chain/src/block_tracker.rs
@@ -164,8 +164,6 @@ impl<'resources, 'blobs> BlockExecutionTracker<'resources, 'blobs> {
         // Account for blobs published indirectly but referenced by the transaction.
         for blob_id in &txn_outcome.blobs_published {
             if let Some(blob) = self.published_blobs.get(blob_id) {
-                // We don't check its sizes again, it was check at the beginning
-                // of the block processing.
                 resource_controller
                     .track_blob_published(blob)
                     .with_execution_context(context)?;
--- a/linera-chain/src/block_tracker.rs
+++ b/linera-chain/src/block_tracker.rs
@@ -157,14 +157,8 @@ impl<'resources, 'blobs> BlockExecutionTracker<'resources, 'blobs> {
         // Account for blobs published by this transaction directly.
         for blob in &txn_outcome.blobs {
             resource_controller
-                .policy()
-                .check_blob_size(blob.content())
+                .track_blob_published(blob)
                 .with_execution_context(context)?;
-            if blob.content().blob_type().is_user_blob() {
-                resource_controller
-                    .track_blob_published(blob.content())
-                    .with_execution_context(context)?;
-            }
         }
 
         // Account for blobs published indirectly but referenced by the transaction.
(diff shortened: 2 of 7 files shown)


linera-chain/src/block_tracker.rs
```diff
@@ -149,14 +154,36 @@ impl<'resources> BlockExecutionTracker<'resources> {
...
+        // Account for blobs published by this transaction directly.
         for blob in &txn_outcome.blobs {
-            if blob.content().blob_type() == BlobType::Data {
+            resource_controller
+                .policy()
+                .check_blob_size(blob.content())
+                .with_execution_context(context)?;
+            if blob.content().blob_type().is_user_blob() {
                 resource_controller
                     .track_blob_published(blob.content())
```
ma2bd, line 165: How about calling `track_blob_published(&blob)` and letting the resource controller decide the fee based on the blob type? Also it would probably make sense to charge most blobs in the same way very soon.

linera-chain/src/block_tracker.rs
```diff
@@ -149,14 +154,36 @@ impl<'resources> BlockExecutionTracker<'resources> {
...
+            if let Some(blob) = self.published_blobs.get(blob_id) {
+                // We don't check its sizes again, it was check at the beginning
+                // of the block processing.
+                if blob.content().blob_type().is_user_blob() {
+                    resource_controller
+                        .track_blob_published(blob.content())
+                        .with_execution_context(context)?;
+                }
+            } else {
+                return Err(ChainError::InternalError(format!(
```
afck, line 181: (This is "internal error" because we already check that we have all published blobs in an earlier step, is it?)

linera-execution/src/transaction_tracker.rs
```diff
@@ -34,11 +37,17 @@ pub struct TransactionTracker {
     /// Events recorded by contracts' `emit` calls.
     events: Vec<Event>,
     /// Blobs created by contracts.
+    ///
+    /// As of right now, blobs created by the contracts are one of the two:
+    /// - OpenChain
+    /// - CreateApplication
```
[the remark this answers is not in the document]
afck: (If we do keep it, let's use backticks.)

linera-execution/src/transaction_tracker.rs
```diff
@@ -34,11 +37,17 @@ pub struct TransactionTracker {
...
     blobs: BTreeMap<BlobId, Blob>,
     /// Operation result.
     operation_result: Option<Vec<u8>>,
     /// Streams that have been updated but not yet processed during this transaction.
     streams_to_process: BTreeMap<ApplicationId, AppStreamUpdates>,
+    /// Published blobs this transaction refers to by BlobId.
```
afck, line 49:
```suggestion
    /// Published blobs this transaction refers to by [`BlobId`].
```

linera-chain/src/block_tracker.rs
deuszx: Yes, exactly.

linera-chain/src/block_tracker.rs
deuszx: I think we don't charge only for the `BlobType::Committee` which is a system-type blob required for correctness.

linera-execution/src/transaction_tracker.rs
deuszx: Done in [Fix typos](https://github.com/linera-io/linera-protocol/pull/3995/commits/c9563c96589bf5511e29f652dcc6ebfddcf5bbb6).

linera-execution/src/transaction_tracker.rs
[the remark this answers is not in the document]
deuszx: Done in [Fix typos](https://github.com/linera-io/linera-protocol/pull/3995/commits/c9563c96589bf5511e29f652dcc6ebfddcf5bbb6).

linera-chain/src/block_tracker.rs
```diff
@@ -44,17 +44,21 @@ pub struct BlockExecutionTracker<'resources> {
     // Index of the currently executed transaction in a block.
     transaction_index: u32,
 
+    // Blobs publised in the block.
```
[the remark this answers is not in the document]
deuszx: Done in [Fix typos](https://github.com/linera-io/linera-protocol/pull/3995/commits/c9563c96589bf5511e29f652dcc6ebfddcf5bbb6).

linera-base/src/data_types.rs
```diff
@@ -1374,6 +1374,11 @@ impl Blob {
     pub async fn load_data_blob_from_file(path: impl AsRef<Path>) -> io::Result<Self> {
         Ok(Self::new_data(fs::read(path)?))
     }
+
+    /// Returns whether the blob is a "user blob".
+    pub fn is_user_blob(&self) -> bool {
```
afck, line 1379: Maybe we should do the opposite, `is_committee_blob`? That's more specific… and I don't like using quotes in comments—that sounds like it's not the correct technical term.

afck (commented):
No blockers from my side!

afck (approved):

linera-base/src/data_types.rs
deuszx: Done in [Rename is_user_blob to is_commitee_blob](https://github.com/linera-io/linera-protocol/pull/3995/commits/671e4d1bea6603d31ae08eeb7624d4e1dc69f074)

PR #4003, opened 2025-05-23
Make REVM address coincide with the one of Linera

MathieuDutSik:
## Motivation

The smart contract addresses that we are using in the EVM are internally all the same: `Address::ZERO.create(0)`.
This creates consistency problems between Ethereum and Linera.

This internal address is available in solidity-based EVM smart contracts as `address(this)`.

## Proposal

We use the `CreateScheme::Custom { address }` that was previously introduced in REVM and merged in REVM-23.1.0.
We also switched to version 24.0.0 of REVM.

Next steps:
* Add the create/create2 functionality to the code.
* Add the transfer functionalities to the code and unify them with Linera.

## Test Plan

* A CI test of the `address(this)` feature is being introduced.
* Also, additional checks are done for the contract address in `fn check_contract_initialization`.

## Release Plan

## Links

None.

### DIFF
--- a/linera-execution/src/evm/database.rs
+++ b/linera-execution/src/evm/database.rs
@@ -66,9 +66,13 @@ impl StorageStats {
 }
 
 pub(crate) struct DatabaseRuntime<Runtime> {
+    /// This is the storage statistics.
     storage_stats: Arc<Mutex<StorageStats>>,
+    /// This is the EVM address of the contract
     pub contract_address: Address,
+    /// The runtime of the contract.
     pub runtime: Arc<Mutex<Runtime>>,
+    /// The uncommited changes to the contract.
     pub changes: EvmState,
 }
 
@@ -114,15 +118,11 @@ impl<Runtime: BaseRuntime> DatabaseRuntime<Runtime> {
     }
 
     /// Returns the tag associated to the contract.
-    fn get_contract_address_key(
-        &self,
-        address: &Address,
-        contract_address: &Address,
-    ) -> Option<u8> {
+    fn get_contract_address_key(&self, address: &Address) -> Option<u8> {
         if address == &Address::ZERO {
             return Some(KeyTag::NullAddress as u8);
         }
-        if address == contract_address {
+        if address == &self.contract_address {
             return Some(KeyTag::ContractAddress as u8);
         }
         None
@@ -200,7 +200,7 @@ where
             return Ok(Some(account.info.clone()));
         }
         let mut runtime = self.runtime.lock().expect("The lock should be possible");
-        let val = self.get_contract_address_key(&address, &self.contract_address);
+        let val = self.get_contract_address_key(&address);
         let Some(val) = val else {
             return Ok(Some(AccountInfo::default()));
         };
@@ -223,7 +223,7 @@ where
                 Some(slot) => slot.present_value(),
             });
         }
-        let val = self.get_contract_address_key(&address, &self.contract_address);
+        let val = self.get_contract_address_key(&address);
         let Some(val) = val else {
             panic!("There is no storage associated to externally owned account");
         };
@@ -265,7 +265,7 @@ where
             if !account.is_touched() {
                 continue;
             }
-            let val = self.get_contract_address_key(address, &self.contract_address);
+            let val = self.get_contract_address_key(address);
             if let Some(val) = val {
                 let key_prefix = vec![val, KeyCategory::Storage as u8];
                 let key_info = vec![val, KeyCategory::AccountInfo as u8];
--- a/linera-execution/src/evm/database.rs
+++ b/linera-execution/src/evm/database.rs
@@ -41,6 +41,8 @@ const SSTORE_COST_RESET: u64 = 2900;
 /// The refund from releasing data.
 const SSTORE_REFUND_RELEASE: u64 = 4800;
 
+/// The number of key writes, reads, release, and no change in EVM has to be accounted for.
+/// Then we remove those costs from the final bill.
 #[derive(Clone, Default)]
 pub(crate) struct StorageStats {
     key_no_operation: u64,
@@ -65,10 +67,12 @@ impl StorageStats {
     }
 }
 
+/// This is the encapsulation of the `Runtime` corresponding to the contract.
 pub(crate) struct DatabaseRuntime<Runtime> {
-    /// This is the storage statistics.
+    /// This is the storage statistics of the read/write in order to adjust gas costs.
     storage_stats: Arc<Mutex<StorageStats>>,
-    /// This is the EVM address of the contract
+    /// This is the EVM address of the contract.
+    /// At the creation, is is set to `Address::ZERO` and then later set to the correct value.
     pub contract_address: Address,
     /// The runtime of the contract.
     pub runtime: Arc<Mutex<Runtime>>,
--- a/linera-execution/src/evm/database.rs
+++ b/linera-execution/src/evm/database.rs
@@ -18,7 +18,7 @@ use revm_database::{AccountState, DBErrorMarker};
 use revm_primitives::{address, Address, B256, U256};
 use revm_state::{AccountInfo, Bytecode, EvmState};
 
-use crate::{BaseRuntime, Batch, ContractRuntime, ExecutionError, ServiceRuntime};
+use crate::{ApplicationId, BaseRuntime, Batch, ContractRuntime, ExecutionError, ServiceRuntime, ViewError};
 
 // The runtime costs are not available in service operations.
 // We need to set a limit to gas usage in order to avoid blocking
@@ -67,6 +67,7 @@ impl StorageStats {
 
 pub(crate) struct DatabaseRuntime<Runtime> {
     storage_stats: Arc<Mutex<StorageStats>>,
+    pub contract_address: Address,
     pub runtime: Arc<Mutex<Runtime>>,
     pub changes: EvmState,
 }
@@ -75,6 +76,7 @@ impl<Runtime> Clone for DatabaseRuntime<Runtime> {
     fn clone(&self) -> Self {
         Self {
             storage_stats: self.storage_stats.clone(),
+            contract_address: self.contract_address,
             runtime: self.runtime.clone(),
             changes: self.changes.clone(),
         }
@@ -96,7 +98,13 @@ pub enum KeyCategory {
     Storage,
 }
 
-impl<Runtime> DatabaseRuntime<Runtime> {
+fn application_id_to_address(application_id: ApplicationId) -> Address {
+    let application_id: [u64; 4] = <[u64; 4]>::from(application_id.application_description_hash);
+    let application_id: [u8; 32] = linera_base::crypto::u64_array_to_be_bytes(application_id);
+    Address::from_slice(&application_id[0..20])
+}
+
+impl<Runtime: BaseRuntime> DatabaseRuntime<Runtime> {
     /// Encode the `index` of the EVM storage associated to the smart contract
     /// in a linera key.
     fn get_linera_key(val: u8, index: U256) -> Result<Vec<u8>, ExecutionError> {
@@ -106,11 +114,15 @@ impl<Runtime> DatabaseRuntime<Runtime> {
     }
 
     /// Returns the tag associated to the contract.
-    fn get_contract_address_key(&self, address: &Address) -> Option<u8> {
+    fn get_contract_address_key(
+        &self,
+        address: &Address,
+        contract_address: &Address,
+    ) -> Option<u8> {
         if address == &Address::ZERO {
             return Some(KeyTag::NullAddress as u8);
         }
-        if address == &Address::ZERO.create(0) {
+        if address == contract_address {
             return Some(KeyTag::ContractAddress as u8);
         }
         None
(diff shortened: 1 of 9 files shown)


MathieuDutSik:
> > The smart contract addresses that we are using are internally all the same
> 
> @MathieuDutSik Can you explain more? I think you mean in the EVM or something?

Yes, I mean in the EVM.

ma2bd:
> > > The smart contract addresses that we are using are internally all the same
> > 
> > 
> > @MathieuDutSik Can you explain more? I think you mean in the EVM or something?
> 
> Yes, I mean in the EVM.

So is this for the `msg.signer` and `tx.origin`values?

MathieuDutSik:
> > > > The smart contract addresses that we are using are internally all the same
> > > 
> > > 
> > > @MathieuDutSik Can you explain more? I think you mean in the EVM or something?
> > 
> > 
> > Yes, I mean in the EVM.
> 
> So is this for the `msg.signer` and `tx.origin`values?

Neither, it is about `address(this)`.

ma2bd:
> Neither, it is about `address(this)`.

Then please explain in the summary that you're fixing the contract's own address `address(this)`

linera-execution/src/evm/revm.rs
```diff
@@ -723,17 +737,30 @@ impl<Runtime: ContractRuntime> CallInterceptorContract<Runtime> {
 
 struct CallInterceptorService<Runtime> {
     db: DatabaseRuntime<Runtime>,
+    contract_address: Address,
```
deuszx, line 740: Is this address of a contract _being called_ or _callee_? Rename or add a comment

linera-execution/src/evm/revm.rs
```diff
@@ -856,6 +897,7 @@ where
     }
 
     fn execute_message(&mut self, message: Vec<u8>) -> Result<(), ExecutionError> {
+        self.db.set_contract_address()?;
```
deuszx, line 900: Why do we call `set_contract_address` in every function here?

deuszx (commented):
I have no idea whether this PR is correct or not. Its description is unclear to me, there are no regression tests that would showcase why the change is necessary (I'm not sure it's possible to write one) function names are unhelpful and code has no comments. 

Please improve.

linera-execution/src/evm/revm.rs
MathieuDutSik: This is explained in `linera-execution/src/evm/database.rs`:
```rust
    pub fn new(runtime: Runtime) -> Self {
        let storage_stats = StorageStats::default();
        // We cannot acquire a lock on runtime here.
        // So, we set the contract_address to a default value
        // and update it later.
        Self {
            storage_stats: Arc::new(Mutex::new(storage_stats)),
            contract_address: Address::ZERO,
            runtime: Arc::new(Mutex::new(runtime)),
            changes: HashMap::new(),
	}
    }
```

MathieuDutSik:
> I have no idea whether this PR is correct or not. Its description is unclear to me, there are no regression tests that would showcase why the change is necessary (I'm not sure it's possible to write one) function names are unhelpful and code has no comments.
> 
> Please improve.

I added an end-to-end test for the `address(this)` functionality.

linera-execution/src/evm/database.rs
```diff
@@ -96,7 +100,13 @@ pub enum KeyCategory {
     Storage,
 }
 
-impl<Runtime> DatabaseRuntime<Runtime> {
+fn application_id_to_address(application_id: ApplicationId) -> Address {
+    let application_id: [u64; 4] = <[u64; 4]>::from(application_id.application_description_hash);
+    let application_id: [u8; 32] = linera_base::crypto::u64_array_to_be_bytes(application_id);
+    Address::from_slice(&application_id[0..20])
+}
```
deuszx, line 107: Why not a `From<ApplicationId> for Address` in the `evm` core file?

MathieuDutSik: In short, because we need to drop the last 12 bytes.

The alternatives are not too appealing:
* Creating an error and using `TryFrom` could solve the problem, but in the case in question, we know that the application is an EVM one, so it just introduces an error that never occurs.
* Putting the `From` and leaving the 12 bytes leaves the false impression that `ApplicationId` can be converted to `Address`.

linera-execution/src/evm/revm.rs
MathieuDutSik: That is the logical design of the scheme.
We don't want a single file.

linera-execution/src/evm/database.rs
```diff
@@ -67,6 +67,7 @@ impl StorageStats {
 
 pub(crate) struct DatabaseRuntime<Runtime> {
     storage_stats: Arc<Mutex<StorageStats>>,
+    pub contract_address: Address,
```
deuszx, line 70: so this is an address of a contract that is currently being executed?

linera-execution/src/evm/database.rs
```diff
@@ -106,11 +116,15 @@ impl<Runtime> DatabaseRuntime<Runtime> {
     }
 
     /// Returns the tag associated to the contract.
-    fn get_contract_address_key(&self, address: &Address) -> Option<u8> {
+    fn get_contract_address_key(
+        &self,
+        address: &Address,
+        contract_address: &Address,
```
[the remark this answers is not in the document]
deuszx: It still doesn't explain why we need two addresses. You alerady have `&self` - which has `self.contract_address` and this method accept two more arguments of type `Address`. So my question is still the same - why do we need THREE contract addresses while there can be only one contract in the current context.

linera-execution/src/evm/database.rs
MathieuDutSik: No to the rename:
* `current_address` makes it look like it will change. It will not.
* `current_address` hides the fact that this is the address of the contract

Instead, I added comments on those points.

linera-execution/src/evm/database.rs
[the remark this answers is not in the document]
MathieuDutSik: I am not 100% sure, but Address zero corresponds to the Address of the user who sent the request.
The address of the smart contract is `self.contract_address`.

Where do you see THREE contract addresses?

I removed the argument `contract_address` and replaced `contract_address` by `self.contract_address` since indeed the address is available by the `self`.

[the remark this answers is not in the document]
deuszx: > Where do you see THREE contract addresses?

`self` has one and the two as arguments.

linera-execution/src/evm/database.rs
MathieuDutSik: It is because you cannot access to the runtime when building the new. Otherwise I would have implemented it that way.

When putting the `Address::ZERO`, we are not putting a valid address to the contract, we put something because we have to. The correct value is set to by the call of `set_contract_address`.

Therefore, it is incorrect to say that it that the contract address changes, it is being set.

deuszx: Current comment are useless. A variable `storage_stats` got a comment "This is the storage statistics.". A variable `contract_address` got a comment "This is the EVM address of the contract". What address? Current? prevous? caller? Callee? Yours, mine? Write commens such that they help the reader - that is the rule #2 of writing comments (rule #1 being don't comment what can be expressed by types or variable names).

There is no comment on `DatabaseRuntime` so reader has no idea what's the context.

The `contract_address` here is the address of contract being currently executed/on the stack. It can change - because it starts as `Address.ZERO`.

linera-execution/src/evm/revm.rs
```diff
@@ -644,12 +654,17 @@ fn get_interpreter_result(
 
 struct CallInterceptorContract<Runtime> {
     db: DatabaseRuntime<Runtime>,
+    // This is the caller contract address
+    contract_address: Address,
```
deuszx:
```suggestion
    caller_contract: Address,
```
The type makes it clear it's an address

linera-execution/src/evm/revm.rs
```diff
@@ -706,11 +732,19 @@ impl<Runtime: ContractRuntime> CallInterceptorContract<Runtime> {
         context: &mut Ctx<'_, Runtime>,
         inputs: &mut CallInputs,
     ) -> Result<Option<CallOutcome>, ExecutionError> {
-        let contract_address = Address::ZERO.create(0);
-        if inputs.target_address == PRECOMPILE_ADDRESS || inputs.target_address == contract_address
+        // Every call to a contract passes by this function.
+        // Three kinds:
+        // --- Call to the PRECOMPILE smart contract.
+        // --- Call to the EVM smart contract itself
```
deuszx, line 745: Do you mean here that when a contract calls itself it goes through `call_or_fail`?

linera-execution/src/evm/revm.rs
```diff
@@ -706,11 +732,19 @@ impl<Runtime: ContractRuntime> CallInterceptorContract<Runtime> {
...
+        // --- Call to other EVM smart contract
+        if self.precompile_addresses.contains(&inputs.target_address)
+            || inputs.target_address == self.contract_address
         {
+            // Precompile calls are handled by the precompile code.
+            // The EVM smart contract is being called
```
deuszx, line 751: If precompile calls are handled separately, an `inputs.target_address` is then unexpected here and should never happen?

linera-execution/src/evm/revm.rs
```diff
@@ -728,17 +762,33 @@ impl<Runtime: ContractRuntime> CallInterceptorContract<Runtime> {
 
 struct CallInterceptorService<Runtime> {
     db: DatabaseRuntime<Runtime>,
+    // This is the caller contract address
+    contract_address: Address,
```
deuszx:
```suggestion
    caller_contract: Address,
```

linera-execution/src/evm/revm.rs
deuszx: That's not what I was asking about. I asked for the explanation here (or on `set_contract_address` function) explaining why we have to call it in the first place. Do not explain it here in the GH comments but leave that in the code for the future readers.

linera-execution/src/evm/revm.rs
```diff
@@ -728,17 +762,33 @@ impl<Runtime: ContractRuntime> CallInterceptorContract<Runtime> {
...
         Self {
             db: self.db.clone(),
+            contract_address: self.contract_address,
+            precompile_addresses: self.precompile_addresses.clone(),
         }
     }
 }
 
 impl<'a, Runtime: ServiceRuntime> Inspector<Ctx<'a, Runtime>> for CallInterceptorService<Runtime> {
+    fn create(
```
deuszx, line 788: This `impl` has the same body for `ServiceRuntime` and `ContractRuntime`. Can't it be shared via an impl for `Runtime: BaseRuntime` (i.e. a Runtime that implements `BaseRuntime` - a common trait for both `Service*` and `Contract*` runtimes.

linera-execution/src/evm/revm.rs
```diff
@@ -706,11 +732,19 @@ impl<Runtime: ContractRuntime> CallInterceptorContract<Runtime> {
...
+            || inputs.target_address == self.contract_address
         {
+            // Precompile calls are handled by the precompile code.
+            // The EVM smart contract is being called
             return Ok(None);
         }
+        // Other smart contracts calls are handled by the runtime
         let target = address_to_user_application_id(inputs.target_address);
         let argument = get_call_argument(context, &inputs.input);
         let authenticated = true;
```
deuszx, line 757: Not a change from this PR but:

why is `authenticated` set to `true` here? In Wasm contracts it's an argument of a call and not hardcoded.

deuszx (commented):
No blockers - couple of points to improve the code.

A more general one: almost all of the code for `CallInterceptorContract` and `CallInterceptorService` is duplicated - apart from a couple of places where it differs. That's hundreds of lines of code that could be removed.

linera-execution/src/evm/revm.rs
MathieuDutSik: The name of the variable is correct.
The `self.contract_address` is used in the `CreateScheme::Custom { address }` to create the contract.
Therefore, the terminology `caller_contract` is inadequate since it implies somehow that the contract exists when in fact, it may not.

linera-execution/src/evm/revm.rs
MathieuDutSik: Yes.
The function `fn call` of the `Inspector` intercepts all the calls to a function. The first such call is indeed the initial call to the contract, which may be to deploy it or to call a specific function.

linera-execution/src/evm/revm.rs
MathieuDutSik: The function `fn call` of the inspector can intercept any calls. Calls to precompile, to the contract itself, calls to other contracts.

So, if we want the precompile calls to be rerouted to their correct destination, we have to return `Ok(None)` in order for the precompile contract to be called.

linera-execution/src/evm/revm.rs
MathieuDutSik: So, the comment was changed.

linera-execution/src/evm/revm.rs
MathieuDutSik: Same as before, the comments was changed.

linera-execution/src/evm/revm.rs
MathieuDutSik: Yes, we can do that.

linera-execution/src/evm/revm.rs
MathieuDutSik: It is a difference between Linera and the EVM.

MathieuDutSik:
> No blockers - couple of points to improve the code.
> 
> A more general one: almost all of the code for `CallInterceptorContract` and `CallInterceptorService` is duplicated - apart from a couple of places where it differs. That's hundreds of lines of code that could be removed.

Ok, thank you for the approval.
Yes, there is some code duplication. Note that the PR https://github.com/linera-io/linera-protocol/pull/3960 does address part of those code duplications.

linera-execution/src/evm/database.rs
MathieuDutSik: The `Runtime` is associated with a smart contract, therefore, I consider the question of what the `contract_address` means to be null and void.

And yes, it starts with `Address::ZERO` and is later updated.

PR #4005, opened 2025-05-23
Unbox the `Signer`

Twey:
## Motivation

As a stopgap measure, when @deuszx first implemented the `Signer` trait, he passed it as a trait object, boxing all its occurrences and associated types.  Now that we have https://github.com/linera-io/linera-protocol/pull/3828, this is no longer necessary, and will make implementing it for non-`Send` types harder, as each boxing site will need to decide whether the resulting type should be `Box<dyn Signer>` or `Box<dyn Signer + Send>`.

## Proposal

Unbox it, passing it in as a type parameter through the `linera_core::Environment` trait.

## Test Plan

CI.

## Release Plan

## Links


ma2bd (approved):

deuszx (commented):
The idea behind using `Box<dyn Signer>` was that this way the concrete type for `Signer` is transparent to the code/client and left entirely to the caller. By introducing the `Signer` type into `Environment` it's no longer the case and the client has to know what kind of Signer is going to be used. In some sense that's a regression w.r.t. to flexibility.

linera-client/src/unit_tests/chain_listener.rs
```diff
@@ -104,7 +104,7 @@ async fn test_chain_listener() -> anyhow::Result<()> {
     let config = ChainListenerConfig::default();
     let storage_builder = MemoryStorageBuilder::default();
     let clock = storage_builder.clock().clone();
-    let mut builder = TestBuilder::new(storage_builder, 4, 1, &mut signer).await?;
+    let mut builder = TestBuilder::new(storage_builder, 4, 1, signer.clone()).await?;
```
deuszx, line 107: Ah, cool. I think I added the `Arc<RwLock<Inner>>` to the `InMemorySigner` and forgot to remove this unnecessay `&mut` argument passing.

linera-core/src/unit_tests/client_tests.rs
```diff
@@ -92,8 +92,8 @@ async fn test_initiating_valid_transfer_with_notifications<B>(
 where
     B: StorageBuilder,
 {
-    let mut signer = InMemorySigner::new(None);
-    let mut builder = TestBuilder::new(storage_builder, 4, 1, &mut signer)
+    let signer = InMemorySigner::new(None);
```
deuszx, line 95: I wonder if it's something that `clippy` could help with - i.e. recognize that `InMemorySigner` is a thin wrapper for `Arc<RwLock` and `mut` is not necessary b/c it can still do internal mutation.

linera-core/src/unit_tests/test_utils.rs
```diff
@@ -724,17 +724,18 @@ impl GenesisStorageBuilder {
     }
 }
 
-pub type ChainClient<S> = crate::client::ChainClient<crate::environment::Impl<S, NodeProvider<S>>>;
+pub type ChainClient<S> =
+    crate::client::ChainClient<crate::environment::Impl<S, NodeProvider<S>, InMemorySigner>>;
```
deuszx, line 728: Looks like `ChainClient` could use a `with_signer(self)` (maybe not now, but as some point) - that replaces the `Signer` instance while keeping everything else.

linera-service/src/schema_export.rs
```diff
@@ -163,20 +163,20 @@ impl ValidatorNodeProvider for DummyValidatorNodeProvider {
 )]
 struct Options {}
 
-struct DummyContext<P, S> {
```
deuszx, line 166: What was `P` even for 🤔

deuszx (approved):
Apart from the other comment - looks good.

Maheshverat:
unsubscribe

Maheshverat:
Unsubscribe

Twey:
> The idea behind using `Box<dyn Signer>` was that this way the concrete type for `Signer` is transparent to the code/client and left entirely to the caller. By introducing the `Signer` type into `Environment` it's no longer the case and the client has to know what kind of `Signer` is going to be used. In some sense that's a regression w.r.t. to flexibility.

The client is still agnostic to the `Signer` (nothing in `Environment` should be made concrete in `linera-core` or below), but the compiler no longer has to be, which frees us up to use associated types, importantly associated future types for `async fn`, and rely on auto traits.

linera-core/src/unit_tests/test_utils.rs
Twey: Feel free to add it as part of the JavaScript signer :) I'm not sure how much sense it makes to change signer after the client has started but it is tempting to be able to set the signer on a per-chain basis.

ma2bd:
> Maybe it's technically agnostic but it's not transparent anymore. `ChainClient<EnvA>` is not the same as `ChainClient<EnvB>` if `Env`s differ in the `Signer` they use.

(If there are no associated types) I suppose one could re-introduce dynamic dispatch later with `ChainClient<Box<dyn Env>>`

PR #4006, opened 2025-05-24
Update link in Cargo.toml

strmfos:
https://docs.rs/linera-indexer-graphql-client/latest/linera_indexer_graphql_client/ - dead
https://docs.rs/linera-service-graphql-client/latest/linera_service_graphql_client/ - new


ma2bd:
I don't think this is correct: the crate simply doesn't have documentation.

ma2bd (changes requested):
please remove the link

strmfos:
did it, thanks

ma2bd (approved):
thanks!

Issue #2905, opened 2024-11-15
Charge for system API calls


jvff:
# Motivation

Execution across the Wasm interface between the host node and the guest application costs resources in terms of execution time, but it is currently not charged for any fees.

# Proposal

Update the `TransactionTracker` to track system API calls and charge a fee (even if zero) on each and every one of them.

ma2bd:
I suspect the only design decision is whether the fees are specified as "dense" map (like the existing fees) or as a spare one.

Issue #1532, opened 2024-01-22
Charge fees for the different types of system calls from an application


ma2bd:
Some system calls (beyond storage) may be expensive and require accounting.

PR #4007, opened 2025-05-24
Correct a bug in the gas accounting of REVM

MathieuDutSik:
## Motivation

There was an error in the gas accounting 

## Proposal

The following is done:
* The `Gas::new(0)` is replaced by a `Gas::new(gas_limit)`.
* The `EvmExecutionError` errors are now more precise.

## Test Plan

The CI.

## Release Plan

## Links

None

### DIFF
--- a/linera-execution/src/evm/revm.rs
+++ b/linera-execution/src/evm/revm.rs
@@ -324,12 +324,15 @@ enum PrecompileTag {
     Service(ServicePrecompileTag),
 }
 
-fn get_precompile_output(output: Vec<u8>) -> Result<Option<InterpreterResult>, String> {
+fn get_precompile_output(
+    output: Vec<u8>,
+    gas_limit: u64,
+) -> Result<Option<InterpreterResult>, String> {
     // The gas usage is set to zero since the proper accounting is done
     // by the called application
     let output = Bytes::copy_from_slice(&output);
     let result = InstructionResult::default();
-    let gas = Gas::new(0);
+    let gas = Gas::new(gas_limit);
     Ok(Some(InterpreterResult {
         result,
         output,
@@ -415,7 +418,7 @@ impl<'a, Runtime: ContractRuntime> PrecompileProvider<Ctx<'a, Runtime>> for Cont
         if address == &PRECOMPILE_ADDRESS {
             let input = get_precompile_argument(context, &inputs.input);
             let output = Self::call_or_fail(&input, gas_limit, context)?;
-            return get_precompile_output(output);
+            return get_precompile_output(output, gas_limit);
         }
         self.inner
             .run(context, address, inputs, is_static, gas_limit)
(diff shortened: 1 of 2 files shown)


linera-execution/src/evm/revm.rs
```diff
@@ -415,7 +418,7 @@ impl<'a, Runtime: ContractRuntime> PrecompileProvider<Ctx<'a, Runtime>> for Cont
         if address == &PRECOMPILE_ADDRESS {
             let input = get_precompile_argument(context, &inputs.input);
             let output = Self::call_or_fail(&input, gas_limit, context)?;
-            return get_precompile_output(output);
+            return get_precompile_output(output, gas_limit);
```
deuszx, line 421: Just from the quick look at this: 

the `gas_limit` value is passed into `Self::call_or_fail` - if the `gas_limit` is supposed to be like "gas left" then `call_or_fail` will decrease it and that - decreased - gas limit should be passed into the `get_precompile_output`. Right now the same gas limit applies to every call here - was that intended?

MathieuDutSik: The gas quantity is the one being passed to the program. It decreases during the computation. So, the gas limit being passed is always different given the context.

The structure is well visible in [github/revm](https://github.com/bluealloy/revm/blob/main/crates/handler/src/precompile_provider.rs):
* We first create a gas record with `Gas::new(gas_limit)`.
* The costs are recorded with `gas.record_cost(gas_used)`.
* It is then returned as a `Option<InterpreterResult>`.

In our implementation we must thus start with `Gas::new(gas_limit)`. This is where we were wrong with `Gas::new(0)`. We do not record the costs since they are accounted separately in Linera.

MathieuDutSik: The precompile calls are about the following:
* `ChainId`, `ApplicationCreatorChainId`, `ChainOwnership` are constant call functions. Currently, they are not accounted for, but they should be after https://github.com/linera-io/linera-protocol/issues/2905 and https://github.com/linera-io/linera-protocol/issues/1532.
* Same for in my opinion for `ValidationRound`, `MessageId`, `MessageIsBouncing`.
* `ReadDataBlob` has an entry `blob_read` in `ResourceControlPolicy`.
* `AssertDataBlobExists` does not seem accounted for.
* `TryCallApplication` is accounted by other calls.
* `TryQueryApplication` does not appear to be accounted for.
* The quadruple `Emit`, `ReadEvent`, `SubscribeToEvent`, `UnsubscribeFromEvent` does not appear to be accounted for.

The point is the following: All those costs should NOT be accounted for in the gas accounting which is EVM specific. In Linera the costs are separated and then finally grouped for the total costs.

That `gas_limit` argument is not used in the code. The PR https://github.com/linera-io/linera-protocol/pull/3960 makes this clear by removing the `gas_limit` argument to the `call_or_fail` function. I am not sure how you come with the `2 * gas_limit` expression.

linera-execution/src/evm/revm.rs
```diff
@@ -324,12 +324,15 @@ enum PrecompileTag {
...
+fn get_precompile_output(
+    output: Vec<u8>,
+    gas_limit: u64,
+) -> Result<Option<InterpreterResult>, String> {
     // The gas usage is set to zero since the proper accounting is done
     // by the called application
     let output = Bytes::copy_from_slice(&output);
     let result = InstructionResult::default();
-    let gas = Gas::new(0);
+    let gas = Gas::new(gas_limit);
```
deuszx, line 335: Comment above is outdated - it says gas is set to 0. Update it - but not remove - so that it matches the current behavior.

MathieuDutSik: Thank you. Corrected.

deuszx (commented):
So basically before this PR a test that calls a precompile would fail? How did you find out that `Gas::new(0)` doesn't work.

PR #4017, opened 2025-05-27
Add Wasm-compatible adapters for implementing Signer trait that work in web browser. Add sample integrations.

deuszx:
## Motivation

We want to be able to sign block proposals using EVM-compatible wallets in the web browser.

## Proposal

Add `linera-web/signer` directory with sample implementations of `Signer` trait for in-memory private key (`in_memory.ts`) and MetaMask integration (`metamask.ts`).

Pre-existing `Wallet` implementations in `linera-web` was adapted to work with the new abstractions.

## Test Plan

CI/Manual testing.

## Release Plan


linera-web/src/signer.rs
```diff
@@ -0,0 +1,172 @@
+// Copyright (c) Zefchain Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! This module contains various implementation of the [`Signer`] trait usable in the browser.
+
+pub mod in_memory {
```
Twey, line 6: As per our discussion yesterday — let's remove this module and the `InMemory` prefixes apropos our discussion yesterday about having the signer logic in JavaScript.

linera-web/src/signer.rs
```diff
@@ -0,0 +1,172 @@
...
+    use linera_base::{
+        crypto::{AccountPublicKey, AccountSignature, CryptoHash, Signer},
+        identifiers::AccountOwner,
+    };
+    use wasm_bindgen::prelude::*;
+    use wasm_bindgen_futures::JsFuture;
+    use web_sys::wasm_bindgen;
+
+    #[repr(u8)]
+    #[wasm_bindgen(js_name = "SignerError")]
```
Twey, line 68: If you use `thiserror` you can throw this — it'll be convertible to a `JsError`.

linera-web/src/signer.rs
deuszx: Done

linera-web/src/lib.rs
```diff
@@ -26,21 +25,20 @@ use linera_core::{
     node::{ValidatorNode as _, ValidatorNodeProvider as _},
 };
 use linera_faucet_client::Faucet;
-use linera_persistent::{self as persistent, Persist as _};
+use linera_persistent::{self as persistent};
```
Twey: Nit:
```suggestion
use linera_persistent as persistent;
```

linera-web/signer/src/metamask_signer.ts
```diff
@@ -0,0 +1,97 @@
+import { ethers } from "ethers";
+import { IJsSigner } from "@linera/client";
```
Twey: We use the prefix `Js` in Rust to refer to JavaScript wrappers over Rust objects (which I'm not sure I don't regret), but we should definitely never use it in JavaScript :)  Also the Hungarian `I` prefix [is frowned upon](https://ts.dev/style/#identifiers) in TypeScript.

linera-web/signer/src/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from "./in_memory";
+export * from "./metamask_signer";
```
Twey: I think I'd like to keep the MetaMask work for a follow-up PR.

linera-web/signer/src/in_memory.ts
```diff
@@ -0,0 +1,55 @@
...
+      typeof owner !== "string" ||
+      !ethers.isAddress(owner) ||
+      this.wallet.address.toLowerCase() !== owner.toLowerCase()
+    ) {
+      throw new Error("Invalid owner address");
+    }
+    return this.wallet.signingKey.publicKey;
+  }
+
+  async contains_key(owner: string): Promise<boolean> {
```
Twey: JavaScript prefers `camelCase` for function identifiers.

.gitignore
```diff
@@ -25,3 +25,4 @@
 # Helm
 kubernetes/linera-validator/working/*
 node_modules
+**dist/
```
Twey:
```suggestion
**/dist/
```

I think this one will exclude anything ending in the string `dist`, and we probably only want to exclude directories called `dist`.

linera-base/src/crypto/secp256k1/evm.rs
```diff
@@ -195,6 +200,12 @@ impl FromStr for EvmPublicKey {
     type Err = CryptoError;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
+        // If the string starts with "0x", we remove it to decode the hex string.
+        let s = if let Some(stripped) = s.strip_prefix("0x") {
+            stripped
+        } else {
+            s
+        };
         hex::decode(s)?.as_slice().try_into()
```
Twey:
```suggestion
        hex::decode(s.strip_prefix("0x").unwrap_or(s))?.as_slice().try_into()
```

linera-web/signer/package-lock.json
Twey, line 1: Is this lock necessary now we have a `pnpm` workspace lock?

linera-web/pnpm-workspace.yaml
```diff
@@ -0,0 +1,3 @@
+packages:
+  - "signer"
+  - "client"
```
Twey, line 3: Nit:
```suggestion
packages:
- signer
- client
```
is sufficient :)

linera-web/signer/src/in_memory.ts
```diff
@@ -0,0 +1,55 @@
+import { Wallet, ethers } from "ethers";
```
Twey: Nit: `kebab-case` is more common for TS file naming.

linera-web/src/lib.rs
```diff
@@ -54,9 +52,12 @@ async fn get_storage() -> Result<WebStorage, <linera_views::memory::MemoryStore
     .await
 }
 
-type PersistentWallet = persistent::Memory<Wallet>;
-type PersistentSigner = persistent::Memory<InMemorySigner>;
-type ClientContext = linera_client::client_context::ClientContext<WebEnvironment, PersistentWallet>;
+/// A wallet that stores the user's chains and keys in memory.
+#[wasm_bindgen(js_name = "Wallet")]
+pub struct InMemoryWallet(persistent::Memory<Wallet>);
```
Twey: The name of this struct is purposely more general than its (opaque) implementation because it shouldn't remain an in-memory wallet.

linera-web/src/signer.rs
```diff
@@ -0,0 +1,158 @@
...
+#[repr(u8)]
+#[wasm_bindgen(js_name = "SignerError")]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum JsSignerError {
+    MissingKey = 0,
+    SigningError = 1,
+    PublicKeyParseError = 2,
+    JsConversionError = 3,
+    UnexpectedSignatureFormat = 4,
+    UnknownError = 9,
```
Twey: There's no need for the `Error` suffix on every (or some) variants of an error enum — it's clear that they represent error conditions from the context.

linera-web/src/signer.rs
```diff
@@ -0,0 +1,158 @@
...
+            JsSignerError::SigningError => write!(f, "Error signing the value"),
+            JsSignerError::PublicKeyParseError => write!(f, "Error parsing the public key"),
+            JsSignerError::JsConversionError => write!(f, "Error converting JS value"),
+            JsSignerError::UnexpectedSignatureFormat => {
+                write!(f, "Unexpected signature format received from JS")
+            }
+            JsSignerError::UnknownError => write!(f, "An unknown error occurred"),
+        }
+    }
+}
```
Twey: I think we talked about this, but can you remind me why we can't use `thiserror` here?

linera-web/src/signer.rs
```diff
@@ -0,0 +1,158 @@
...
+  get_public_key(owner: string): Promise<string>;
+
+  /**
+   * Checks whether the instance holds a key whose associated address matches the given EVM address.
+   *
+   * @param owner - The EVM address to check for.
+   * @returns A promise that resolves to `true` if the key exists and matches the given address, otherwise `false`.
+   */
+  contains_key(owner: string): Promise<boolean>;
+}"#;
```
Twey: I'd really like this to be in a separate (`.d.ts`) file.  Maybe we can `include_str`?

linera-web/src/signer.rs
```diff
@@ -0,0 +1,158 @@
...
+impl From<JsValue> for JsSignerError {
+    fn from(value: JsValue) -> Self {
+        if let Some(fnum) = value.as_f64() {
+            match fnum as u8 {
+                0 => JsSignerError::MissingKey,
+                1 => JsSignerError::SigningError,
+                2 => JsSignerError::PublicKeyParseError,
+                3 => JsSignerError::JsConversionError,
+                4 => JsSignerError::UnexpectedSignatureFormat,
+                _ => JsSignerError::UnknownError,
```
Twey: So do we expect JavaScript code to throw numbers?  I think I'd rather just bubble up a `JsValue` for whatever exception types the implementer chooses to throw.

linera-web/src/signer.rs
```diff
@@ -0,0 +1,158 @@
...
+            JsSignerError::UnknownError
+        }
+    }
+}
+
+impl std::error::Error for JsSignerError {}
+
+// An interface that will be compiled to TypeScript and exported for use in the browser.
+#[wasm_bindgen(typescript_custom_section)]
+const JS_SIGNER_INTERFACE: &'static str = r#"
```
Twey: It's not necessary to prefix everything in a `signer` module with `(Js)Signer` :)

(this is actually [a Clippy lint](https://rust-lang.github.io/rust-clippy/master/index.html#module_name_repetitions) that [~is~ was](https://github.com/rust-lang/rust-clippy/pull/13541) enabled for this crate under the `clippy::pedantic` umbrella — bypassed here on the technicality that the module is named `signer` not `js_signer`)

linera-web/src/signer.rs
deuszx: But it will still end up in the same `*.d.ts` file - you just want to extract this interface from this particular Rust file?

linera-web/src/signer.rs
deuszx: Signer's `type Error` member is required to be `Sync`.

linera-web/src/lib.rs
deuszx: It is in memory though.

linera-web/pnpm-workspace.yaml
deuszx: I think that's [prettier's](https://prettier.io/) doing 🤔 I'll check

.gitignore
deuszx: Yeah, this is a typo.

linera-web/src/signer.rs
deuszx: ~~I was actually meant to do that and forgot. good idea.~~

deuszx: No, actually this won't work... Let me explain:

the `type Error` on `Signer` has to be `Send` etc. `JsValue` is not. That's why I introduced a new error which is `Send`. We have to return `Self::Error` from these methods hence bubbling up the `JsValue` is not going to work 
b/c `JsValue` is not implementing `StdError`.

linera-web/signer/package-lock.json
deuszx: Removed.

linera-web/pnpm-workspace.yaml
deuszx: Yeah, that's prettier formatting it.

linera-web/src/signer.rs
Twey: Right, I don't have an issue with the built structure but I don't love embedding large chunks of TypeScript into Rust source files.

linera-web/src/lib.rs
Twey: Currently it is, but I would like it to not be. It will be stored in IndexedDB soon.

deuszx: OK, I can rename back to `PersistentWallet`.

deuszx: Done.

linera-web/src/signer.rs
deuszx: Done in [Extract Signer typescript interface to a file.](https://github.com/linera-io/linera-protocol/pull/4017/commits/775985228af7f97636549b106382d878980f4121)

linera-web/src/signer.rs
Twey: I'd suggest that we either: 
- make the error not require `Send` on the Web, or
- use `gloo_utils::errors::JsError` to bring the error contents into Rust-land.

deuszx: I checked and `gloo_utils` havne't been updated in almost a year and has ton of issues.

As for the first - out of scope for this PR I'd say. I can do it in a follow-up.

linera-web/src/signer.rs
Twey: `thiserror` doesn't prevent that, so long as you don't store the `JsValue` inside it — and we're throwing it away here anyway.

deuszx: So you want me to bring in a whole new dependency to `linera-web` only to write the `Display` messages inlined, in a macro, rather than like this? If it's only about not having to implement `Display` manually I'd rather to leave it as is.

linera-web/signer/src/index.ts
deuszx: Done

linera-web/src/signer.rs
deuszx: I tried using `gloo`'s `JsError` but it's not really improving much I think. It has only `TryFrom<JsValue> for JsError` so we can't easily map our `JsValue` errors to `JsError` - b/c there's always a possibility it will not map properly.

Twey: I'm working on removing the `Send` and `Sync` bounds from `Signer::Error` — I'll resolve this for now and we can remove this code after I'm done with that.

linera-web/src/signer.rs
Twey: `linera-web` already has this dependency (transitively) :)  `thiserror` is pervasive in the Rust ecosystem, and it would be nice to use it here as is idiomatic in the rest of our codebase.

But you're right, it's probably fine for now because the more important features (preserving backtraces, for example) would rely on having additional information that we're currently throwing away here.  I'll revisit this after we can put `JsError`s in the signer error type.

linera-web/src/signer.rs
```diff
@@ -0,0 +1,125 @@
...
+extern "C" {
+    // We refer to the interface defined above.
+    #[wasm_bindgen(typescript_type = "Signer")]
+    pub type JsSigner;
+
+    #[wasm_bindgen(catch, method)]
+    async fn sign(this: &JsSigner, owner: JsValue, value: Vec<u8>) -> Result<JsValue, JsValue>;
+
+    #[wasm_bindgen(catch, method)]
+    async fn get_public_key(this: &JsSigner, owner: JsValue) -> Result<JsValue, JsValue>;
```
Twey: Does this not require a rename (`js_name = getPublicKey`) to match the interface? (and below)

deuszx: Not really b/c the JavaScript interfaces are not used in the Rust code at all 🤔 I think what Rust sees vs what JavaScript sees are two different things. I.e. the remapping of TypeScript interfaces is not affecting the Rust here and the `extern "C"` things are not visible in the JavaScript land.

linera-web/src/signer.rs
deuszx: I think we won't be able to use `gloo`'s `JsError` anyway. Or rather it won't be that useful.

linera-web/src/signer.rs
Twey: Right, but if the user passes something like `{ getPublicKey: () => … }` and then we call `get_public_key` on it (which is what this code does) we'll get an error (‘`get_public_key` is undefined’).  `js_name = getPublicKey` means that when we call `get_public_key` on the Rust side we'll call the JavaScript function `getPublicKey`, which is what our TypeScript interface claims should exist.

That is to say, this `extern` block and the TypeScript `interface` block must match, even though it won't break the build, or we'll get runtime errors if the caller is following the TypeScript interface.

Twey (approved):
Looks good to me.  Great work!

PR #4018, opened 2025-05-27
Change the API of read_blob_state to return options.

MathieuDutSik:
## Motivation

Following PR https://github.com/linera-io/linera-protocol/pull/4014 we look at the `read_blob_state` and change their API as options. The same design considerations apply here.

## Proposal

The `read_blob_state(s)` were returning a ViewError that contained a `NotFound`. This is inadequate and in any
case the `NotFound` is never processed.

In all use cases of those functions, we had some `BlobsNotFound` variant in the relevant types, the error type was thus put there. It could change the behavior of some function. However, this is better than having some error buried in `ViewError`.

The `NotFound` cannot yet be changed because it is used by the `read_certificate(s)`. 

Other changes:
* After PR 4104, some of the metrics were placed before the storage calls. This has been corrected.
* In some of the accounting of the calls, the plural calls were sometimes counting several calls as one, and in other cases as the number of entries in the call. This has been corrected.

## Test Plan

The CI.

## Release Plan

## Links

None.

### DIFF
--- a/linera-core/src/unit_tests/test_utils.rs
+++ b/linera-core/src/unit_tests/test_utils.rs
@@ -588,13 +588,19 @@ where
         sender: oneshot::Sender<Result<CryptoHash, NodeError>>,
     ) -> Result<(), Result<CryptoHash, NodeError>> {
         let validator = self.client.lock().await;
-        let certificate_hash = validator
+        let blob_state = validator
             .state
             .storage_client()
             .read_blob_state(blob_id)
             .await
-            .map(|blob_state| blob_state.last_used_by)
             .map_err(Into::into);
+        let certificate_hash = match blob_state {
+            Err(err) => Err(err),
+            Ok(blob_state) => match blob_state {
+                None => Err(NodeError::BlobsNotFound(vec![blob_id])),
+                Some(blob_state) => Ok(blob_state.last_used_by),
+            },
+        };
 
         sender.send(certificate_hash)
     }
--- a/linera-service/src/proxy/main.rs
+++ b/linera-service/src/proxy/main.rs
@@ -346,9 +346,13 @@ where
                 let certificates = self.storage.read_certificates(hashes).await?;
                 Ok(Some(RpcMessage::DownloadCertificatesResponse(certificates)))
             }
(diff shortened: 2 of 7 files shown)


linera-core/src/unit_tests/test_utils.rs
```diff
@@ -588,13 +588,19 @@ where
...
-        let certificate_hash = validator
+        let blob_state = validator
             .state
             .storage_client()
             .read_blob_state(blob_id)
             .await
-            .map(|blob_state| blob_state.last_used_by)
             .map_err(Into::into);
+        let certificate_hash = match blob_state {
+            Err(err) => Err(err),
```
deuszx, line 598: Since you're returning the error, we can just propagate it with `?` above directly. i.e. `.await?`. Then you shoudnt even need the `map_err(Into::into)` b/c `?` transforms the error automatically IIRC 🤔

deuszx (approved):

linera-service/src/proxy/main.rs
```diff
@@ -346,9 +346,13 @@ where
                 let certificates = self.storage.read_certificates(hashes).await?;
                 Ok(Some(RpcMessage::DownloadCertificatesResponse(certificates)))
             }
-            BlobLastUsedBy(blob_id) => Ok(Some(RpcMessage::BlobLastUsedByResponse(Box::new(
-                self.storage.read_blob_state(*blob_id).await?.last_used_by,
-            )))),
+            BlobLastUsedBy(blob_id) => {
+                let blob_state = self.storage.read_blob_state(*blob_id).await?;
+                let blob_state = blob_state.ok_or(anyhow!("Blob not found {}", blob_id))?;
```
afck, line 351: I'd use `ok_or_else` here, too.

linera-storage/src/db_storage.rs
```diff
@@ -855,7 +848,7 @@ where
             #[cfg(with_metrics)]
             metrics::READ_CERTIFICATES_COUNTER
                 .with_label_values(&[])
-                .inc();
+                .inc_by(hashes.len() as u64);
```
afck, line 851: (So now it's the number of certificates, rather than the number of `read_certificates` calls?)

Issue #2652, opened 2024-10-19
The client does not always update the wallet.


afck:
In `linera/main.rs` and `linera/client_context.rs`, we only update the wallet for the chains we expect to have been modified, i.e. the chain the command was executed on. But e.g. when synchronizing received certificates we may end up updating other (multi-owner) chains in our wallet, too. We should always update the wallet for all chains.

Issue #2569, opened 2024-10-04
Revisit `ChainClientState` locking.


afck:
https://github.com/linera-io/linera-protocol/pull/2133 allowed calling all `ChainClient` methods concurrently from different tasks, which introduced some issues that we addressed in https://github.com/linera-io/linera-protocol/pull/2567.

We should verify that all remaining `ChainClient` methods are really safe to run concurrently, and consider finding a better way to ensure that those that aren't don't.

afck:
Fixed in #5666.

PR #4024, opened 2025-05-28
Remove hash, time and height from `ChainClientState`.

afck:
## Motivation

The next block height, most recent block hash, and most recent block timestamp are currently read from the wallet and put into `ChainClientState`, which is then updated frequently (but not always: https://github.com/linera-io/linera-protocol/issues/2652) based on the local node's chain state.

So the fields in the `ChainClientState` are meant to first equal the values from the wallet, and then the values from the chain state view.

## Proposal

Keep only the values from the wallet in `ChainClient` and never update them. Whenever the _current_ values are needed, read them from the chain state directly.

## Test Plan

CI

## Release Plan

## Links

- Closes https://github.com/linera-io/linera-protocol/issues/2652.
- _Partly_ addresses https://github.com/linera-io/linera-protocol/issues/2569.

### DIFF
--- a/linera-client/src/wallet.rs
+++ b/linera-client/src/wallet.rs
@@ -9,11 +9,7 @@ use linera_base::{
     ensure,
     identifiers::{AccountOwner, ChainId},
 };
-use linera_core::{
-    client::{ChainClient, PendingProposal},
-    data_types::ChainInfo,
-    Environment,
-};
+use linera_core::{client::PendingProposal, data_types::ChainInfo};
 use serde::{Deserialize, Serialize};
 
 use crate::{config::GenesisConfig, error, Error};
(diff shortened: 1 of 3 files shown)


linera-client/src/wallet.rs
```diff
@@ -132,17 +132,21 @@ impl Wallet {
         Ok(())
     }
 
-    pub fn update_from_state<Env: Environment>(&mut self, chain_client: &ChainClient<Env>) {
+    pub async fn update_from_state<Env: Environment>(
```
Twey, line 135: And after I just got rid of this `async`!  Is it not possible to just get the current chain info here without making a query?  I think the point of calling this function is that the chain info in the `ChainClient` has just been updated — so it seems a bit silly to update it again here.

afck: What do you mean by "local chain info"? To get the chain info we have to query the database, so it's async.

The point of this PR is exactly to remove keeping a copy of some of the chain info's fields in memory all the time, which we already fail to always keep in sync.

linera-client/src/wallet.rs
```diff
@@ -132,17 +132,21 @@ impl Wallet {
...
+        &mut self,
+        chain_client: &ChainClient<Env>,
+    ) -> Result<(), Error> {
         let client_owner = chain_client.preferred_owner();
-        let state = chain_client.state();
+        let info = chain_client.chain_info().await?;
```
deuszx, line 140: Can't we pass in the `info` itself into this method, rather than `chain_client`? This way the method is non-async.

afck: I'll change it—but we still have to pass in the chain client, because we take the preferred owner and the pending block from it.

afck: Done in https://github.com/linera-io/linera-protocol/pull/4024/commits/6fce2cb6752d6c4b6ad86ac245ec8aaf6ff51df1.

linera-client/src/wallet.rs
Twey: Obviated by the thread at https://github.com/linera-io/linera-protocol/pull/4024#discussion_r2111991430.

Twey (approved):
Looks good to me!

Issue #4025, opened 2025-05-28
Check whether sender chain certificates are actually relevant.


afck:
In `synchronize_received_certificates_from_validator`, we should check that the downloaded certificates are actually relevant, i.e.:
For each chain, the highest downloaded certificate must have sent a message to our chain.

PR #4026, opened 2025-05-28
Check that downloaded sender chain blocks are relevant.

afck:
## Motivation

When trying to find "received certificates", faulty validators might send us certificates that didn't actually send anything to the chain in question, causing the client to unnecessarily process valid but irrelevant blocks.

## Proposal

Check that they actually sent something.

## Test Plan

Since this only defends against some spam, isn't critical for security, and testing it would require considerably extending the test validators, I suggest not adding a test for this. CI will check if it _breaks_ anything (i.e. yields false positives).

## Release Plan

## Links

- Closes https://github.com/linera-io/linera-protocol/issues/4025.

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -807,7 +807,7 @@ impl<Env: Environment> Client<Env> {
         let remote_certificates = remote_node
             .download_certificates(certificate_hashes)
             .await?;
-        let mut certificates = BTreeMap::new();
+        let mut certificates_by_height_by_chain = BTreeMap::new();
 
         // Check the signatures and keep only the ones that are valid.
         for confirmed_block_certificate in remote_certificates {
@@ -832,7 +832,7 @@ impl<Env: Environment> Client<Env> {
                     warn!("Skipping received certificate from past epoch {epoch:?}");
                 }
                 CheckCertificateResult::New => {
-                    certificates
+                    certificates_by_height_by_chain
                         .entry(sender_chain_id)
                         .or_insert_with(BTreeMap::new)
                         .insert(height, confirmed_block_certificate.clone());
@@ -842,7 +842,7 @@ impl<Env: Environment> Client<Env> {
 
         // Increase the tracker up to the first position we haven't downloaded.
         for entry in remote_log {
-            if certificates
+            if certificates_by_height_by_chain
                 .get(&entry.chain_id)
                 .is_some_and(|certs| certs.contains_key(&entry.height))
             {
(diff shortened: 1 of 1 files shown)


linera-core/src/client/mod.rs
```diff
@@ -807,9 +807,9 @@ impl<Env: Environment> Client<Env> {
         let remote_certificates = remote_node
             .download_certificates(certificate_hashes)
             .await?;
+        let mut certificates = BTreeMap::new();
```
MathieuDutSik, line 810: The type here is `BTreeMap<ChainId, BTreeMap<BlockHeight,GenericCertificate<ConfirmedBlock>>>`.
That is somewhat complex and the name does not give a hint that it would be that.

What about `certificates_by_height_by_chain` or something similar.

linera-core/src/client/mod.rs
```diff
@@ -832,31 +832,47 @@ impl<Env: Environment> Client<Env> {
...
+        }
+
         Ok(ReceivedCertificatesFromValidator {
             public_key: remote_node.public_key,
             tracker,
-            certificates,
+            certificates: certificates
+                .into_values()
+                .flat_map(BTreeMap::into_values)
+                .collect(),
```
MathieuDutSik, line 875: So, for a given chain, we can take certificates of different heights, not just the last one.

MathieuDutSik (approved):
Looks sensible.

linera-core/src/client/mod.rs
afck: The last one should be the one that sent the latest message to our chain.

_Currently_ we download all its descendants, but soon (#3969) we will download only the ones that sent messages.

linera-core/src/client/mod.rs
afck: Done in https://github.com/linera-io/linera-protocol/pull/4026/commits/a2ee3d9a4bb37472012910e5aafe97dc60f15457.

linera-core/src/client/mod.rs
```diff
@@ -832,31 +832,47 @@ impl<Env: Environment> Client<Env> {
...
+            if !certs
+                .values()
+                .last()
+                .is_some_and(|cert| cert.block().recipients().contains(&chain_id))
+            {
+                warn!(
+                    "Skipping received certificates from chain {sender_chain_id:.8}:
+                    No messages for {chain_id:.8}."
+                );
+                certs.clear();
```
deuszx, line 865: This would clear the `certs` for the sender chain but it wouldn't decrease the tracker. Isn't that a problem?

afck:
> they do not cover all cases where this change might introduce a regression

Do you have specific test cases in mind that we should add?

deuszx (approved):

afck:
As discussed, I ran the tests locally with a `panic!()` in the code to verify that this check is passed in all of them. That's expected: It can only be triggered by a very specific fault in the validators, which would be complicated to implement.

PR #4029, opened 2025-05-29
`linera-views`: refactor errors

Twey:
## Motivation

We consistently require that the store errors be convertible to `ViewError`, but we don't encode that bound in `Context`, so we duplicate `ViewError: From<Context::Error>` bounds throughout the codebase.

## Proposal

Move the bound into the definition of the `Context` trait instead.

While we're at it, use a boxed error for the store errors in `ViewError` instead of converting to string, so that the backtrace and other error metadata isn't lost.

## Test Plan

CI.

## Release Plan

## Links


linera-views/src/error.rs
```diff
@@ -0,0 +1,72 @@
...
+    #[error("the value is too large for the client")]
+    TooLargeValue,
+}
+
+impl ViewError {
+    /// Creates a `NotFound` error with the given message and key.
+    pub fn not_found<T: std::fmt::Debug>(msg: &str, key: T) -> Self {
+        ViewError::NotFound(format!("{msg} {key:?}"))
+    }
+}
```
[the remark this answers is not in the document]
Twey: That will be nice!  I think we currently only explicitly import `ViewError` in several places to use `ViewError::not_found`.

linera-views/src/context.rs
```diff
@@ -80,7 +80,11 @@ impl BaseKey {
 
 /// The context in which a view is operated. Typically, this includes the client to
 /// connect to the database and the address of the current entry.
-pub trait Context: Clone {
+#[cfg_attr(not(web), trait_variant::make(Send + Sync))]
+pub trait Context: Clone
+where
+    crate::ViewError: From<Self::Error>,
```
ma2bd, line 86: Interesting, I thought I had tried that and it failed then.

linera-views/src/error.rs
```diff
@@ -0,0 +1,72 @@
...
+    #[error("missing database entries")]
+    MissingEntries,
+
+    /// The values are incoherent.
+    #[error("post load values error")]
+    PostLoadValuesError,
+
+    /// The value is too large for the client
+    #[error("the value is too large for the client")]
+    TooLargeValue,
```
ma2bd, line 64: English nit: Should this be `ValueTooLarge` or something else?

Twey: `TooLargeValue` is consistent with `MissingEntries`, for example.

PR #4032, opened 2025-05-29
Small changes to speed up and trim down CI costs.

deuszx:
## Motivation

Speed up and cut down CI costs.

## Proposal

`docker-compose` was ran on every push which seems unnecessary. It checks that our docker setup (that is expected to mimic validators' usage) is working. Since we care that it works on `main` and possibly release branches it doesn't have to run on every commit. **Now it runs on merge queue (before merge to `main`) and on release branches**

`test-readme-scripts` tests backwards compatibility. This is important but 95% of our PRs do not break them. We can detect it on `main` (and release branches) and fix right away. **Now it runs on merge queue (before merge to `main`) and on release branches**.

`lint-check-all-features` **runs on merge queue (before merge to `main`) and on release branches**.

`default-features-and-witty-integration-test` now runs on `ubuntu-latest-8-cores` : reason being this task already lasted 40+ minutes and is `cargo test` which should parallelize nicely.

`check-wit-files` runs on `ubuntu-latest` (rather than `*-16-cores`).

## Test Plan

CI testing CI.

## Release Plan

## Links

### DIFF
--- a/.github/workflows/docker-compose.yml
+++ b/.github/workflows/docker-compose.yml
@@ -3,14 +3,6 @@ name: Docker Compose
 on:
   push:
     branches: [ main, 'devnet_*', 'testnet_*' ]
-  pull_request:
-    branches:
-      - "**"
-    paths:
-      - 'docker/**'
-      - 'Cargo.toml'
-      - '.github/workflows/docker-compose.yml'
-      - 'toolchains/**'
   workflow_dispatch:
 
 # This allows a subsequently queued workflow run to interrupt previous runs on pull-requests
--- a/.github/workflows/docker-compose.yml
+++ b/.github/workflows/docker-compose.yml
@@ -2,7 +2,8 @@ name: Docker Compose
 
 on:
   push:
-    branches: [ main, 'devnet_*', 'testnet_*' ]
+    branches: [ 'devnet_*', 'testnet_*' ]
+    merge_group:
   workflow_dispatch:
 
 # This allows a subsequently queued workflow run to interrupt previous runs on pull-requests
--- a/.github/workflows/lint-check-all-features.yml
+++ b/.github/workflows/lint-check-all-features.yml
@@ -6,7 +6,7 @@ on:
     merge_group:
   workflow_dispatch:
 
-# This allows a subsequently queued workflow run to interrupt previous runs on pull-requests
+# This allows a subsequently queued workflow run to interrupt previous runs on pull requests
 concurrency:
   group: '${{ github.workflow }} @ ${{ github.event.pull_request.head.label || github.head_ref || github.run_id }}'
   cancel-in-progress: true
--- a/.github/workflows/lint-check-all-features.yml
+++ b/.github/workflows/lint-check-all-features.yml
@@ -1,11 +1,10 @@
-name: Run README tests
+name: Lint check all features
 
 on:
   push:
     branches: [ 'devnet_*', 'testnet_*' ]
     merge_group:
   workflow_dispatch:
-  
 
 # This allows a subsequently queued workflow run to interrupt previous runs on pull-requests
 concurrency:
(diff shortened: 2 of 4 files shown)


Twey:
> We can detect it on `main` (and release branches) and fix right away.

> Both can be easily moved to a github merge queue.

I'm a bit confused about this — the merge queue will be tested _before_ merging to main, right?

Twey (approved):
Looks sensible to me, but as usual with CI I guess we won't know until we try it in anger :)

.github/workflows/test-readmes.yml
```diff
@@ -0,0 +1,60 @@
...
+      - 'docker_scylla/**'
+      - 'configuration/**'
+      - 'kubernetes/**'
+  workflow_dispatch:
+  
+
+# This allows a subsequently queued workflow run to interrupt previous runs on pull-requests
+concurrency:
+  group: '${{ github.workflow }} @ ${{ github.event.pull_request.head.label || github.head_ref || github.run_id }}'
+  cancel-in-progress: true
```
Twey, line 13: I don't think we should `cancel-in-progress` on the main branch — we should test every (mainline) commit.

deuszx: If you push a fix to `main` it makes sense to cancel the previous-possibly-broken commit. You can't fix it without new commit (assuming we don't force push to `main`) so it'll be red anyway.

Twey: The problem is that GitHub keeps records of which commits failed and why, and it would be nice to have that for all the commits on `main`.  Also you run into the problem where if we merge commits to `main` too fast, older commits will effectively not have their tests run, and then if the tests fail it's hard to see which commit caused it.

deuszx: It's now moved to a merge queue.

.github/workflows/docker-compose.yml
```diff
@@ -3,14 +3,6 @@ name: Docker Compose
...
   push:
     branches: [ main, 'devnet_*', 'testnet_*' ]
-  pull_request:
-    branches:
-      - "**"
-    paths:
-      - 'docker/**'
-      - 'Cargo.toml'
-      - '.github/workflows/docker-compose.yml'
-      - 'toolchains/**'
```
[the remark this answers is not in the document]
afck: But running this seems reasonable to me; are we really modifying these files that often?

.github/workflows/lint-check-all-features.yml
```diff
@@ -0,0 +1,49 @@
+name: Lint check all features
+
+on:
+  push:
+    branches: [ 'devnet_*', 'testnet_*' ]
+    merge_group:
+  workflow_dispatch:
+
+# This allows a subsequently queued workflow run to interrupt previous runs on pull-requests
```
afck:
```suggestion
# This allows a subsequently queued workflow run to interrupt previous runs on pull requests
```

.github/workflows/test-readmes.yml
```diff
@@ -0,0 +1,54 @@
+name: Run README tests
+
+on:
+  push:
+    branches: [ 'devnet_*', 'testnet_*' ]
+    merge_group:
+  workflow_dispatch:
+  
+
+# This allows a subsequently queued workflow run to interrupt previous runs on pull-requests
```
afck:
```suggestion
# This allows a subsequently queued workflow run to interrupt previous runs on pull requests
```

.github/workflows/docker-compose.yml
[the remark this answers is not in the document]
deuszx: Do you mean I should bring back these guards so that we run the job even less frequently? Right now it's moved to the merge queue (so it makes sure `main` is never broken) and release branches. I thought we should make absolutely sure that "releasable" software works under docker.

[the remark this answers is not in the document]
afck: Ah, sorry! Yes, makes sense then. No need to bring them back.

PR #4034, opened 2025-05-29
Restructure the error types of `linera-views`.

MathieuDutSik:
## Motivation

Several entries in the `ViewError` and `DynamoDbStoreInternalError` can be removed.

## Proposal

The following is done:
* The `TooLargeValue` is removed from `IndexedDbError` as it is unused.
* The `TooLargeValue` is removed from `MemoryStoreError`. This requires changing one test, which ought to be fine.
* The `TooLargeValue` is removed from `ViewError`.
* The unused errors for `DynamoDbStoreInternalError` are removed.
* No other unused errors were found.

## Test Plan

The CI.

## Release Plan

## Links

None.

### DIFF
--- a/linera-views/src/error.rs
+++ b/linera-views/src/error.rs
@@ -20,9 +20,9 @@ pub enum ViewError {
     #[error(transparent)]
     ArithmeticError(#[from] linera_base::data_types::ArithmeticError),
 
-    /// An failed to lock a reentrant collection entry since it currently being accessed
+    /// Failed to lock a reentrant collection entry since it is currently being accessed
     #[error(
-        "failed to lock a reentrant collection entry since it currently being accessed: {0:?}"
+        "failed to lock a reentrant collection entry since it is currently being accessed: {0:?}"
     )]
     TryLockError(Vec<u8>),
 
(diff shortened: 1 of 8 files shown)


linera-views/src/error.rs
```diff
@@ -20,8 +20,10 @@ pub enum ViewError {
     #[error(transparent)]
     ArithmeticError(#[from] linera_base::data_types::ArithmeticError),
 
-    /// An error happened while trying to lock.
-    #[error("failed to lock collection entry: {0:?}")]
+    /// An failed to lock a reentrant collection entry since it currently being accessed
+    #[error(
+        "failed to lock a reentrant collection entry since it currently being accessed: {0:?}"
```
Twey:
```suggestion
        "failed to lock a reentrant collection entry since it is currently being accessed: {0:?}"
```

Twey (approved):

afck (approved):

linera-views/src/error.rs
MathieuDutSik: Thanks.

Issue #4037, opened 2025-05-30
ChainListener panics when it's run without any user chains but has GenesisConfig


deuszx:
ChainListener.run throws the panic at https://github.com/linera-io/linera-protocol/blob/main/linera-client/src/chain_listener.rs#L385 but it should add an admin chain to the listening chains as well. 

Solution:

When reating the `ChainListener` from the `GenesisConfig` we should find the admin chain and add it to the `listening` map by default.

PR #4044, opened 2025-06-02
Always add admin chain to listener.

afck:
## Motivation

The chain listener currently panics if it has an empty list of chains.

Also, it doesn't even listen to the admin chain if no chain is in the wallet.

## Proposal

Always listen to the admin chain.

## Test Plan

A unit test for this was added.

## Release Plan

## Links

- Closes https://github.com/linera-io/linera-protocol/issues/4037.

### DIFF
--- a/linera-client/src/unit_tests/chain_listener.rs
+++ b/linera-client/src/unit_tests/chain_listener.rs
@@ -20,6 +20,7 @@ use linera_core::{
     DEFAULT_GRACE_PERIOD,
 };
 use linera_execution::system::Recipient;
+use linera_storage::Storage;
 use tokio_util::sync::CancellationToken;
 
 use super::util::make_genesis_config;
@@ -53,7 +54,8 @@ impl chain_listener::ClientContext for ClientContext {
         let chain = self
             .wallet
             .get(chain_id)
-            .unwrap_or_else(|| panic!("Unknown chain: {}", chain_id));
+            .cloned()
+            .unwrap_or_else(|| UserChain::make_other(chain_id, Timestamp::from(0)));
         self.client.create_chain_client(
             chain_id,
             chain.block_hash,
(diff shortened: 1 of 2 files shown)


bart-linera (approved):
Looks reasonable!

linera-client/src/unit_tests/chain_listener.rs
```diff
@@ -53,7 +54,8 @@ impl chain_listener::ClientContext for ClientContext {
         let chain = self
             .wallet
             .get(chain_id)
-            .unwrap_or_else(|| panic!("Unknown chain: {}", chain_id));
+            .cloned()
+            .unwrap_or_else(|| UserChain::make_other(chain_id, Timestamp::from(0)));
```
deuszx, line 58: Shouldn't we also add it to `self.wallet` if we create it?

afck: We do the same thing in production. (I'll deduplicate this code, possibly in another PR.)

I'm not sure we should. At least in the tests it doesn't matter.

deuszx: OK, I figured that if we expect the chain to be in the wallet and it's not, we should add it if we're creating a new one.

linera-client/src/unit_tests/chain_listener.rs
```diff
@@ -191,3 +193,73 @@ async fn test_chain_listener() -> anyhow::Result<()> {
 
     Ok(())
 }
+
+/// Tests that the chain listener always listens to the admin chain.
+#[test_log::test(tokio::test)]
+async fn test_chain_listener_admin_chain() -> anyhow::Result<()> {
+    // Create two chains.
```
deuszx: Why do we need the second chain? The minimal example is that we have an admin chain and a new client that requests some tokens from the faucet will start tracking it.

linera-client/src/unit_tests/chain_listener.rs
afck: Sure, but I'm not sure we should necessarily have every chain in the wallet that we create a `ChainClient` for.

linera-client/src/unit_tests/chain_listener.rs
afck: Sorry, that comment is outdated. We're actually only creating one chain. And that _is_ the admin chain, because it's root chain 0, and the test builder automatically configures that as the admin chain.

linera-client/src/unit_tests/chain_listener.rs
```diff
@@ -191,3 +193,72 @@ async fn test_chain_listener() -> anyhow::Result<()> {
...
+    let storage_builder = MemoryStorageBuilder::default();
+    let mut builder = TestBuilder::new(storage_builder, 4, 1, signer.clone()).await?;
+    let client0 = builder.add_root_chain(0, Amount::ONE).await?;
+    let genesis_config = make_genesis_config(&builder);
+    let admin_id = genesis_config.admin_id();
+    let storage = builder.make_storage().await?;
+    let delivery = CrossChainMessageDelivery::NonBlocking;
+
+    println!("Client 0 chain ID: {}", client0.chain_id());
+    println!("Admin chain ID: {}", admin_id);
```
deuszx: leftovers

PR #4046, opened 2025-06-02
Close the faucet chain if `open_chain` fails.

afck:
## Motivation

If `open_chain` fails, e.g. because the faucet chain is out of funds (and, due to some other failure, that was not handled in the previous request), the request fails and the same thing will happen again for all future requests.

## Proposal

If `open_chain` fails, switch to a new faucet chain.

## Test Plan

We don't know exactly yet what failed on Testnet Babbage, so we can't reproduce it. Deploying this should handle errors more gracefully.

## Release Plan

## Links

### DIFF
--- a/linera-faucet/server/src/lib.rs
+++ b/linera-faucet/server/src/lib.rs
@@ -171,10 +171,9 @@ where
         }
 
         let ownership = ChainOwnership::single(owner);
-        let (message_id, certificate) = faucet_client
+        let result = faucet_client
             .open_chain(ownership, ApplicationPermissions::default(), self.amount)
-            .await?
-            .try_unwrap()?;
+            .await;
         self.context
             .lock()
             .await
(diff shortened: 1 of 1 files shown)


linera-faucet/server/src/lib.rs
```diff
@@ -171,18 +171,19 @@ where
...
+            .await;
         self.context
             .lock()
             .await
             .update_wallet(&faucet_client)
             .await?;
 
         // Only keep using this chain if there will still be enough balance to close it.
-        if faucet_client.local_balance().await? < self.amount.try_add(MAX_FEE.try_mul(2)?)? {
+        if result.is_err()
```
deuszx, line 184: It's a bit confusing to read the comment and the code b/c the comment is like a negation of the code. I.e. it explains what we will do if this `if` statement is false.

Maybe rewrite the comment so that it explains the `if` block - "Close this faucet if OpenChain request errored or its balance is too low".

afck: Done in https://github.com/linera-io/linera-protocol/pull/4046/commits/eb40abe86ecb8970d30f61bcff993cadbec05153.

PR #4047, opened 2025-06-02
Optimize ScyllaDB's batch writes

ndr-ds:
## Motivation

So it is a known thing that batches are [not token aware](https://rust-driver.docs.scylladb.com/stable/statements/batch.html#performance) in the ScyllaDB’s Rust driver. Batches will be sent to a random node, which will then forward things to the proper nodes, which makes it not be “token aware”. Which also means there’s an extra network hop for most batch requests. This is what the default load balancing policies currently do.
So currently if someone using the Rust driver needs atomicity, they can use batches, but they’ll get a bit of a performance hit as the batch won’t be token aware.
So, for us to have the best performance on batches, maintaining the per partition atomicity that it guarantees, we would need shard aware batching, but that’s not yet supported in the Rust driver.
There is [some work](https://github.com/scylladb/scylla-rust-driver/pull/738) being attempted for “shard aware batching”, but one of the reviewers is arguing that there are ways of solving this problem that [don’t involve user code](https://github.com/scylladb/scylla-rust-driver/pull/738#issuecomment-2203328104). These ways involve creating a custom Load Balancing Policy, which is what I'm doing in this PR.

## Proposal

Build a custom "Sticky" Load Balancing Policy. This policy will be specific to a given partition: given the partition, it will remember what are the `(node, shard)` pairs for all the replicas containing this partition's data. Then for every batch that we try to send, send them to one of the replicas, in a round robin fashion to spread load across the replicas.

We'll have an LRU cache keyed on the partition key, that contains either a `Ready` value or a `NotReady` value.
The reason for this is that there are some cases where you try to get the endpoints information for a token, and the Rust driver hasn't updated it's metadata yet about the table, so that information isn't filled yet. If we have a `Ready` value, we have the actual "sticky" policy already, with the `(node, shard)` endpoints, and we're good to go. If you have a `NotReady` value, you'll have a timestamp of when we last attempted to get the endpoints. We always wait at least 2 seconds before trying again, to give the driver time to update itself, and not overload it with these endpoint requests. Until then we use the default policy and take a bit of a performance hit, but should be for very limited time.

The `NotReady` state can also contain the `Token` already for that partition, in case we managed to calculate it in the last attempt. The `Token` is calculated by doing a `Murmur3` hash of the tables specs and partition key. If the table doesn't change that `Token` will never change for this partition key. Since there's hashing involved, we cache it to not do that repeated work.

If we ever decide to auto scale our ScyllaDB deployment based on load, we'll need to add a mechanism here to invalidate these cache entries when that happens.

## Test Plan

CI + I won't merge before I benchmark this code together with the new key space partitioning PR, to make sure the performance is what we expect.

## Release Plan

### DIFF
--- a/linera-views/src/backends/scylla_db.rs
+++ b/linera-views/src/backends/scylla_db.rs
@@ -10,7 +10,11 @@
 use std::{
     collections::{BTreeSet, HashMap},
     ops::Deref,
-    sync::Arc,
+    sync::{
+        atomic::{AtomicUsize, Ordering},
+        Arc,
+    },
+    time::Duration,
 };
 
 use async_lock::{Semaphore, SemaphoreGuard};
@@ -23,17 +27,24 @@ use scylla::{
         session::Session,
         session_builder::SessionBuilder,
     },
+    cluster::{ClusterState, Node, NodeRef},
     deserialize::{DeserializationError, TypeCheckError},
     errors::{
-        DbError, ExecutionError, IntoRowsResultError, NewSessionError, NextPageError, NextRowError,
-        PagerExecutionError, PrepareError, RequestAttemptError, RequestError, RowsError,
+        ClusterStateTokenError, DbError, ExecutionError, IntoRowsResultError, MetadataError,
+        NewSessionError, NextPageError, NextRowError, PagerExecutionError, PrepareError,
+        RequestAttemptError, RequestError, RowsError,
     },
     policies::{
-        load_balancing::{DefaultPolicy, LoadBalancingPolicy},
+        load_balancing::{DefaultPolicy, FallbackPlan, LoadBalancingPolicy, RoutingInfo},
         retry::DefaultRetryPolicy,
     },
     response::PagingState,
-    statement::{batch::BatchType, prepared::PreparedStatement, Consistency},
+    routing::{Shard, Token},
+    statement::{
+        batch::{Batch, BatchType},
+        prepared::PreparedStatement,
+        Consistency,
+    },
 };
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
@@ -225,15 +236,17 @@ impl ScyllaDbClient {
     async fn build_default_session(uri: &str) -> Result<Session, ScyllaDbStoreInternalError> {
         // This explicitly sets a lot of default parameters for clarity and for making future changes
         // easier.
-        SessionBuilder::new()
+        let session = SessionBuilder::new()
             .known_node(uri)
+            .cluster_metadata_refresh_interval(Duration::from_secs(10))
             .default_execution_profile_handle(Self::build_default_execution_profile_handle(
                 Self::build_default_policy(),
             ))
             .build()
             .boxed_sync()
-            .await
-            .map_err(Into::into)
+            .await?;
+        session.refresh_metadata().await?;
+        Ok(session)
     }
 
     async fn get_multi_key_values_statement(
@@ -243,6 +256,7 @@ impl ScyllaDbClient {
         if let Some(prepared_statement) = self.multi_key_values.get(&num_markers) {
             return Ok(prepared_statement.clone());
         }
+
         let markers = std::iter::repeat_n("?", num_markers)
             .collect::<Vec<_>>()
             .join(",");
@@ -265,6 +279,7 @@ impl ScyllaDbClient {
         if let Some(prepared_statement) = self.multi_keys.get(&num_markers) {
             return Ok(prepared_statement.clone());
         };
+
         let markers = std::iter::repeat_n("?", num_markers)
             .collect::<Vec<_>>()
             .join(",");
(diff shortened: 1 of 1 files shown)


ndr-ds:
* **#4198** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4198?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4049** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4049?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4047** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4047?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4047?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#4066** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4066?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

linera-views/src/backends/scylla_db.rs
```diff
@@ -242,48 +292,46 @@ impl ScyllaDbClient {
...
+            .join(",");
+        let prepared_statement = self
+            .session
+            .prepare(format!(
+                "SELECT k FROM {}.{} WHERE root_key = ? AND k IN ({})",
+                KEYSPACE, self.namespace, markers
+            ))
+            .await?;
+        self.multi_keys
+            .insert(num_markers, prepared_statement.clone());
```
MathieuDutSik: Why switch back to the construction with two searches, `get` and then `insert` ?

ndr-ds: `Dashmap`'s `entry` will actually deadlock if you call it and there's anyone else using the map :)
So it's not a good idea here, and was actually a bug to use it. We got "lucky" that CI didn't trigger it much in the PR that was merged, but it will be triggered increasingly more with these PRs, specially the new partitions one.
But CI in this PR was already deadlocking sometimes.

afck:
Do we need this if we never have batches that affect multiple shards?

Can we enforce that each view is on a single shard, and then just not batch anything else?

ndr-ds:
> Do we need this if we never have batches that affect multiple shards?

Yes, even if all queries in the batch are for the same partition key, the Rust driver will still just send the batch to a random node by default.

> Can we enforce that each view is on a single shard, and then just not batch anything else?

Yes, but that's a much bigger change I think 😅 both the enforcing that each view is on a single shard, and not batching.
I want to do a few follow ups here where we do that enforcement, and stop doing single query batches and do just the queries directly instead. It's still useful to keep this even in that case, because now batches should still be reasonably performant even in the case where people don't care about atomicity, but want to use batches just to minimize network calls to the DB.

linera-views/src/backends/scylla_db.rs
```diff
@@ -98,13 +113,38 @@ const MAX_BATCH_SIZE: usize = 5000;
 /// The keyspace to use for the ScyllaDB database.
 const KEYSPACE: &str = "kv";
 
+/// The default size of the cache for the load balancing policies.
+const DEFAULT_LOAD_BALANCING_POLICY_CACHE_SIZE: usize = 50_000;
```
afck: Does this work now, as of Rust 1.83.0?

```suggestion
const DEFAULT_LOAD_BALANCING_POLICY_CACHE_SIZE: NonZeroUsize = NonZeroUsize::new(50_000).unwrap();
```

linera-views/src/backends/scylla_db.rs
```diff
@@ -524,6 +653,76 @@ impl ScyllaDbClient {
...
+        "StickyShardPolicy".to_string()
+    }
+
+    // Always try first to route to the sticky shard.
+    fn pick<'a>(
+        &'a self,
+        _request: &'a RoutingInfo<'a>,
+        _cluster: &'a ClusterState,
+    ) -> Option<(NodeRef<'a>, Option<Shard>)> {
+        // fetch_add will wrap around on overflow, so we should be ok just incrementing forever here.
```
afck, line 609: Why not use `fetch_update` instead, so it always contains the actual value?

linera-views/src/backends/scylla_db.rs
```diff
@@ -524,6 +653,76 @@ impl ScyllaDbClient {
...
+impl LoadBalancingPolicy for StickyShardPolicy {
+    fn name(&self) -> String {
+        "StickyShardPolicy".to_string()
+    }
+
+    // Always try first to route to the sticky shard.
+    fn pick<'a>(
+        &'a self,
+        _request: &'a RoutingInfo<'a>,
+        _cluster: &'a ClusterState,
```
afck: (So the reason we can't just use `request.token` and `cluster.get_token_endpoints(…)` here is that for batch statements, the former will be `None`?)

linera-views/src/backends/scylla_db.rs
```diff
@@ -412,46 +460,127 @@ impl ScyllaDbClient {
...
+            Ok(policy) => {
+                let policy = Arc::new(policy);
+                cache.push(
+                    partition_key.to_vec(),
+                    LoadBalancingPolicyCacheEntry::Ready(policy.clone()),
+                );
+                policy
+            }
+            Err(error) => {
+                // Cache that the policy creation failed, so we don't try again too soon, and don't
```
afck: Is this expected? Should we log if that happens?

linera-views/src/backends/scylla_db.rs
```diff
@@ -412,46 +460,127 @@ impl ScyllaDbClient {
...
+        let policy = {
+            let mut cache = self
+                .batch_load_balancing_policies
+                .lock()
+                .map_err(|_| ScyllaDbStoreInternalError::PoisonedMutex)?;
+            if let Some(policy) = cache.get(partition_key) {
+                match policy {
+                    LoadBalancingPolicyCacheEntry::Ready(policy) => policy.clone(),
+                    LoadBalancingPolicyCacheEntry::NotReady(timestamp, token) => {
+                        if Timestamp::now().delta_since(*timestamp)
```
afck: Maybe we should use `Instance` here instead of `Timestamp`? Our `linera_base` timestamp type is just to define and serialize timestamps in the protocol as `u64`s. Since this here is only used locally I'd go with the standard library. Also, this expression here could then be `timestamp.elapsed()`.

linera-views/src/backends/scylla_db.rs
ndr-ds: `fetch_update` AFAIU is actually implemented as basically a `compare_exchange` busy loop, and performance will degrade proportional to contention. `fetch_add` is much cheaper, specially with `Ordering::Relaxed`, won't have performance degradation based on contention, and does the job we want also. So made more sense to use it IMHO

linera-views/src/backends/scylla_db.rs
ndr-ds: According to the rust driver's [documentation](https://rust-driver.docs.scylladb.com/stable/statements/batch.html#performance), yes. It will just send to a random node.

linera-views/src/backends/scylla_db.rs
ndr-ds: I can do a `WARN` here I think, but it shouldn't happen too many times AFAIU

linera-views/src/backends/scylla_db.rs
ndr-ds: Seems like it does!

linera-views/src/backends/scylla_db.rs
afck: Sorry, I meant `Instant`.

linera-views/src/backends/scylla_db.rs
MathieuDutSik: Ok, but then I would add a comment on that.
I think the most appropriate is "We are not using the `dashmap::DashMap::entry` in order to reduce contention and creating deadlocks".

linera-views/src/backends/scylla_db.rs
```diff
@@ -98,13 +112,38 @@ const MAX_BATCH_SIZE: usize = 5000;
...
+    // The timestamp of the last time the policy creation was attempted.
+    NotReady(Instant, Option<Token>),
+}
+
+/// The configuration of the ScyllaDB client.
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct ScyllaDbClientConfig {
+    /// The delay before the sticky load balancing policy creation is retried.
+    pub delay_before_sticky_load_balancing_policy_retry_ms: u64,
+}
```
MathieuDutSik: Now that we have a `ScyllaDbClientConfig`, why not put the `DEFAULT_LOAD_BALANCING_POLICY_CACHE_SIZE` in it?

MathieuDutSik: Also, the replication_factor could have its place here. It does not have to be in this PR, though.

linera-views/src/backends/scylla_db.rs
afck: I guess it's generally never safe to hold `DashMap` references over an `await` point. (Or is it?)

linera-views/src/backends/scylla_db.rs
```diff
@@ -98,13 +112,38 @@ const MAX_BATCH_SIZE: usize = 5000;
...
 const KEYSPACE: &str = "kv";
 
+/// The default size of the cache for the load balancing policies.
+const DEFAULT_LOAD_BALANCING_POLICY_CACHE_SIZE: NonZeroUsize = NonZeroUsize::new(50_000).unwrap();
+
+enum LoadBalancingPolicyCacheEntry {
+    Ready(Arc<dyn LoadBalancingPolicy>),
+    // The timestamp of the last time the policy creation was attempted.
+    NotReady(Instant, Option<Token>),
+}
```
MathieuDutSik: I have some maybe stupid concern, but I think that the sharding policy in `ScyllaDb` is dynamic.
So, I wonder if this affects the caching.

linera-views/src/backends/scylla_db.rs
ndr-ds: @MathieuDutSik The `Dashmap` documentation is pretty clear about the deadlock risk, I think adding another comment here is redundant.

@afck Exactly, it's not safe. Which is why it's best to avoid things that hold a reference to the map for too long, like using `entry` for example. It is designed for concurrent access though, there are just these nuances to be careful about

linera-views/src/backends/scylla_db.rs
ndr-ds: I could, sure! And on the `replication_factor`, yeah, that's in my TODO list 😅 I noticed a while back it doesn't belong on common config as it's actually specific to `ScyllaDb`

linera-views/src/backends/scylla_db.rs
ndr-ds: Good question. For a given `partition_key`, the `Token` is always unique, as it's always calculated by the same hash function. So the `Token` won't change. As far as the endpoints, and AFAIU, we'll only reshuffle token ranges (causing the nodes that hold a different token to change) when we scale ScyllaDB up or down, as in add or remove VMs in the NodePool, or run `ALTER TABLE` commands, stuff like that.
We currently don't do that, and it would complicate the code a bit to add support for it now, so I would rather leave it for when it's needed

MathieuDutSik (approved):
No obstacle on my side, but I would like to see benchmarks confirming the improvements before merging.

Benchmarks can be added to the PR description.

linera-views/src/backends/scylla_db.rs
```diff
@@ -412,46 +462,128 @@ impl ScyllaDbClient {
...
+        // Getting the sticky shard policy does some serializing and hashing under the hood, so
+        // we cache the policy to avoid that extra work.
+        let policy = {
+            let mut cache = self
+                .batch_load_balancing_policies
+                .lock()
+                .map_err(|_| ScyllaDbStoreInternalError::PoisonedMutex)?;
+            if let Some(policy) = cache.get(partition_key) {
+                match policy {
+                    LoadBalancingPolicyCacheEntry::Ready(policy) => policy.clone(),
```
ma2bd: No need for nesting `if let` and `match`:
```
match cache.get(partition_key) {
    Some(LoadBalancingPolicyCacheEntry::Ready(policy)) => ...
    ...
    None => ...
}
```

linera-views/src/backends/scylla_db.rs
ma2bd: Wait, we don't support changes in the sharding assignment on the ScyllaDb side?

ndr-ds: This sticky load balancing policy part currently doesn't, but we can add support for that in a follow up PR, I think :)

ndr-ds:
I can add support for that on this PR, or in a following one, to not inflate the size of this one even more (I won't merge this without that, will merge the full stack at once, probably). I was thinking on creating a stack that we know works, then merging the full stack once we agree it gets things to a good state (instead of doing a separate branch).
Happy to do a separate branch as well though, if we prefer that 😅

ndr-ds:
Closing this, as we went with a simpler approach

PR #4052, opened 2025-06-03
Deduplicate certificate downloading code.

afck:
## Motivation

The functionality of `try_query_certificates_from` is basically duplicated in `try_synchronize_chain_state_from`. Also it's a bit arbitrary which errors we ignore and which we pass on, and the prefix `try_` doesn't really add anything.

## Proposal

Deduplicate code, remove the prefix, only ignore errors where necessary: e.g. when iterating over validators, when we don't want one validator's fault to stop the iteration.

Also, do the `chain_id` check already in `RemoteNode::query_certificates_from`: If a remote node returns certificates that don't match the requested chain, that node is faulty.

## Test Plan

CI

## Release Plan

## Links

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -257,7 +257,7 @@ impl<Env: Environment> Client<Env> {
     }
 
     /// Fetches the chain description blob if needed, and returns the chain info.
-    pub async fn fetch_chain_info(
+    async fn fetch_chain_info(
         &self,
         chain_id: ChainId,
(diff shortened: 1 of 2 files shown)


ma2bd:
In this context, didn't "try_" mean that silently failing is ok?

afck:
> In this context, didn't "try_" mean that silently failing is ok?

Yes, but we weren't consistent in terms of which errors we ignored. I moved that logic up to the loop where we go through all the validators instead.

deuszx (approved):

PR #4053, opened 2025-06-03
Simplify storage of BlobStates

ma2bd:
## Motivation

* Improve code readability
* Fix a particular failure path for clients

## Proposal

* Remove a few local variables
* Do not record last_used_by in the blob state if the certificate is not yet in storage

## Test Plan

CI

## Release Plan

### DIFF
--- a/linera-chain/src/block.rs
+++ b/linera-chain/src/block.rs
@@ -139,12 +139,12 @@ impl ConfirmedBlock {
     }
 
     /// Returns a blob state that applies to all blobs used by this block.
-    pub fn to_blob_state(&self) -> BlobState {
+    pub fn to_blob_state(&self, is_stored_block: bool) -> BlobState {
         BlobState {
-            last_used_by: self.0.hash(),
+            last_used_by: is_stored_block.then_some(self.0.hash()),
             chain_id: self.chain_id(),
             block_height: self.height(),
-            epoch: self.epoch(),
+            epoch: is_stored_block.then_some(self.epoch()),
         }
     }
 }
--- a/linera-storage/src/db_storage.rs
+++ b/linera-storage/src/db_storage.rs
@@ -702,19 +702,20 @@ where
         blob_state: BlobState,
     ) -> Result<Epoch, ViewError> {
         let current_blob_state = self.read_blob_state(blob_id).await?;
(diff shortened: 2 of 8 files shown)


linera-chain/src/block.rs
```diff
@@ -139,12 +139,12 @@ impl ConfirmedBlock {
...
     /// Returns a blob state that applies to all blobs used by this block.
-    pub fn to_blob_state(&self) -> BlobState {
+    pub fn to_blob_state(&self, is_stored_block: bool) -> BlobState {
         BlobState {
-            last_used_by: self.0.hash(),
+            last_used_by: is_stored_block.then_some(self.0.hash()),
             chain_id: self.chain_id(),
             block_height: self.height(),
-            epoch: self.epoch(),
+            epoch: is_stored_block.then_some(self.epoch()),
```
afck, line 147: (Why specifically `last_used_by` and `epoch`, but not `chain_id` and `block_height`?)

linera-storage/src/db_storage.rs
```diff
@@ -734,36 +712,15 @@ where
...
         for (maybe_blob_state, blob_id) in maybe_blob_states.iter().zip(blob_ids) {
-            let (should_write, latest_epoch) = match maybe_blob_state {
-                None => (true, blob_state.epoch),
-                Some(current_blob_state) => (
-                    overwrite && current_blob_state.epoch < blob_state.epoch,
-                    current_blob_state.epoch.max(blob_state.epoch),
-                ),
-            };
-            if should_write {
+            if maybe_blob_state.as_ref().and_then(|state| state.epoch) < blob_state.epoch {
```
afck: (`None < Some(_)`, is it?)

linera-chain/src/block.rs
ma2bd: * `last_used_by` to prevent the client-side code from downloading the blob.
* `epoch` for the ordering.
* The rest stays for debugging purpose?

linera-storage/src/db_storage.rs
ma2bd: Oh I messed this up

ma2bd: This was failing to write the temporary blob state when `maybe_blob_state` is `None` and so is `blob_state.epoch`.

PR #4056, opened 2025-06-04
Remove Rust jobs from running on main branch.

deuszx:
## Motivation

Rust workflow already runs on every commit and also on merge queue right now. There's no point in running it on `main` as well.

## Proposal

Remove `main` from branches for which `rust.yml` is ran.

## Test Plan

CI

## Release Plan

- Nothing to do 

## Links

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -2,7 +2,8 @@ name: Rust
 
 on:
   push:
-    branches: [ main, 'devnet_*', 'testnet_*' ]
+    branches: [ 'devnet_*', 'testnet_*' ]
+  merge_group:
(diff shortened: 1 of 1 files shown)


.github/workflows/rust.yml
```diff
@@ -2,7 +2,8 @@ name: Rust
 
 on:
   push:
-    branches: [ main, 'devnet_*', 'testnet_*' ]
+    branches: [ 'devnet_*', 'testnet_*' ]
```
Twey, line 5: Does this mean that we'll still run these tests on every commit in PRs to devnet/testnet branches?

Twey (approved):

PR #4057, opened 2025-06-04
Replacement of `KeyValueStoreView` by a direct version.

MathieuDutSik:
## Motivation

The existing `KeyValueStoreView` is a scalable design for implementing a Key-Value store, which is needed by the EVM smart contracts and the Wasm smart contracts. However, it is a slow design that is overkill for 99% of the use cases.

Replacing it with an `enum Big/Small` might be a solution but it presents an inconvenient that we cannot easily transition from a Big sdesign to a small design if the storage use decreases. Also:
* It creates a strong discontinuity in the usage performance near the chosen threshold.
* Anyway, when we compute the hash, we have to access the whole set of key/values of the `KeyValueStoreView`.

## Proposal

The following is done:
* The whole Key/value store is replaced by a single `BTreeMap<Vec<u8>, vec<u8>>`.
* Therefore, the API becomes a `RegisterView`, and this container is chosen accordingly. Note that this doubles the memory usage, but this was already the case for the small container in the now-closed PR.
* We cannot use the derive macro in `linera-views`. Therefore, the file `key_value_store_view.rs` is moved to `linera-execution`.
* The `ViewContainer` is eliminated and with it the corresponding tests.

## Test Plan

The CI was adequate for identifying problems with the design.

## Release Plan

## Links

None.


ma2bd:
> The existing KeyValueStoreView is a scalable design for implementing a Key-Value store

Except that hashing is going to force reading everything anyway

linera-execution/src/key_value_store_view.rs
```diff
@@ -0,0 +1,399 @@
+// Copyright (c) Zefchain Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! We implement a `KeyValueStoreView` implements View and the function of `KeyValueStore`.
```
ma2bd, line 4: This doesn't belong here, does it? Especially if we want to keep the unit tests.

ma2bd: Also it could be called `SmallKeyValueStoreView` if we want to keep the old implementation around for performance comparison (which are not done in this PR)

ma2bd (changes requested):
As discussed, we'd like to
* keep the old DS in place
* have the new "small" data-structure in `linera-view` as well.
* make the change later when we have benchmarks

MathieuDutSik:
I created another PR that encapsulated the ideas of the request for change.

PR #4060, opened 2025-06-04
`linera-views`: make `Sync` a bound on `View`, except on the Web

Twey:
## Motivation

Currently we have a lot of implementations in `linera-views` that don't work on the Web because they require `Sync` bounds.  For example, `MapView` and `QueueView` are not usable on the Web.

## Proposal

Conditionally use `trait-variant` to add bounds to the traits themselves rather than the implementations.  [Futures should ‘always’ be `Sync`](https://internals.rust-lang.org/t/what-shall-sync-mean-across-an-await/12020/18) anyway.

`linera_storage_service::client` contains an implementation of `KeyValueStore` that is not `Sync` even though it isn't on the Web, due to a trait object inside `tonic` that [has](https://github.com/hyperium/tonic/issues/81) [caused](https://github.com/hyperium/tonic/pull/82) [some](https://github.com/hyperium/tonic/pull/84) [contention](https://github.com/hyperium/tonic/issues/117).  For that implementation, follow the solution [given there](https://github.com/hyperium/tonic/issues/117#issuecomment-748009137) and use [`sync_wrapper::SyncFuture`](https://docs.rs/sync_wrapper/latest/sync_wrapper/struct.SyncFuture.html) to make the futures `Sync` based on the fact that they can only be used through `Pin<&mut Self>` anyway.

Ditto `aws-smithy-async`.

## Test Plan

CI.

## Release Plan

## Links

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5636,7 +5636,6 @@ dependencies = [
  "serde",
  "serde-reflection",
  "similar-asserts",
- "sync_wrapper 1.0.2",
  "test-strategy",
  "thiserror 1.0.69",
  "tokio",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5636,6 +5636,7 @@ dependencies = [
  "serde",
  "serde-reflection",
  "similar-asserts",
+ "sync_wrapper 1.0.2",
  "test-strategy",
  "thiserror 1.0.69",
  "tokio",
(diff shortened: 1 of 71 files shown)


Twey:
If https://github.com/hyperium/tonic/pull/2293 gets merged then we can revert these changes to `linera-storage-service`.

PR #4062, opened 2025-06-04
Restructure the calls to `read_certificate(s)`.

MathieuDutSik:
## Motivation

The functions `read_certificate(s)` are introducing an entry `ViewError::NotFound` in the `ViewError`, which is not great.
Here we eliminate this.

## Proposal

What we want to do is separate errors that are of storage access of pure bugs from the errors that are more normal in operations.

There are 4 schemes of error in `deserialize_certificate`:
* Two `NotFound` patterns.
* One assertion that the value of the hash coincides.
* One check of consistency of the hashes.

Now they are grouped in two error patterns: `ReadCertificateError` and `ReadCertificatesError`.
The resulting error type is something complicated of the form: `Result<Result<Certificate, ReadCertificateError>, ViewError>`. The corresponding reading is thus `??`.

This is the logic of what we do: First deal with the error of bad access to the storage, and then with the error of the reading itself that was not correct. 

The corresponding `ReadCertificatesError` are added to the `WorkerError` and `ChainClientError`.

This, of course, does not resolve the errors themselves. But exposing them directly is better than having them burried in `ViewError`.

There are still patterns of errors `NotFound` left to eliminate.

## Test Plan

The CI.

## Release Plan

## Links

None.


linera-storage/src/db_storage.rs
```diff
@@ -937,20 +970,22 @@ where
...
-            .ok_or_else(|| ViewError::not_found("certificate bytes for hash", hash))?;
-        let value_bytes = pair[1]
-            .as_ref()
-            .ok_or_else(|| ViewError::not_found("value bytes for hash", hash))?;
+    ) -> Result<ResultReadCertificate, ViewError> {
+        let Some(cert_bytes) = pair[0].as_ref() else {
+            return Ok(Err(ReadCertificateError::MissingLiteCertificate(hash)));
+        };
+        let Some(value_bytes) = pair[1].as_ref() else {
+            return Ok(Err(ReadCertificateError::MissingLiteCertificate(hash)));
```
```suggestion
            return Ok(Err(ReadCertificateError::MissingConfirmedBlock(hash)));
```

linera-storage/src/db_storage.rs
```diff
@@ -937,20 +970,22 @@ where
...
+        };
         let cert = bcs::from_bytes::<LiteCertificate>(cert_bytes)?;
         let value = bcs::from_bytes::<ConfirmedBlock>(value_bytes)?;
-        assert_eq!(value.hash(), hash);
-        let certificate = cert
+        let value_hash = value.hash();
+        if value_hash != hash {
```
afck: But this is something `linera-storage` itself guarantees, isn't it? Same for `with_value` below succeeding.

The other error can be caused by the user, when trying to read data that was never written.

So I'm not sure they belong in the same error type.

afck (commented):
I'm not sure if that's useful: I'd rather return a `Result<Option<_>, _>`. Return `None` if the lite certificate or block wasn't found, and return an error in all other cases.

(Just my opinion. I'll let others chime in.)

MathieuDutSik:
> I'm not sure if that's useful: I'd rather return a `Result<Option<_>, _>`. Return `None` if the lite certificate or block wasn't found, and return an error in all other cases.
> 
> (Just my opinion. I'll let others chime in.)

Ok, we have to see the design that we want to achieve:
* I think it is clear that we want `ViewError` to be only about the errors of the storage and the views.
* If the `linera-storage` hash inconsistencies are possible bugs in the code, then an `assert` statement might be the best design for both of them. That is the `ViewError::InconsistentEntries` would become an `assert`.
* And then yes if either of them is missing, then return `None`.

That way the error would become a `CertificatesNotFound(Vec![CryptoHash])`. That makes sense to me. Is it what you had in mind?

But yes, let's see what others have to say about it.

deuszx:
My opinion is to return `None` for missing entry. `View`s are collections and it's rather typical to return `None` when an entry is missing.

Also, turning this into an error variant makes it easier to miss, handle and propagate up the stack. We have this problem with `BlobNotFound` error variant which was incorrectly bubbled up without handling locally.

linera-storage/src/lib.rs
```diff
@@ -332,6 +330,88 @@ pub struct ChainRuntimeContext<S> {
     user_services: Arc<DashMap<ApplicationId, UserServiceCode>>,
 }
 
+#[derive(Debug, Error)]
+pub enum ReadCertificateError {
```
ma2bd: This seems like parts of a `enum StorageError`

linera-storage/src/lib.rs
```diff
@@ -332,6 +330,88 @@ pub struct ChainRuntimeContext<S> {
...
+impl std::fmt::Display for ReadCertificateError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{:?}", self)
+    }
+}
+
+type ResultReadCertificate = Result<ConfirmedBlockCertificate, ReadCertificateError>;
+
+#[derive(Debug, Error)]
+pub struct ReadCertificatesError {
```
ma2bd: This looks complicated. The motivation for this PR is to simplify things.

MathieuDutSik: It would be inexact to say that the motivation of this PR is to simplify things.

The motivation of this PR is to clean up things, to expose the right level of abstraction. You want simple? Put one error in `ViewError::NotFound` and the other in `ViewError::InconsistentEntries`. That is the current state of affairs, and we are not happy about it.

Is it possible to do another design? Surely:
* For example, in the query of certificates, we can take only the first incorrect one if it happens.
* Or, we can hide the reason for the hash failure behind an `IncorrectHashes(Vec<CryptoHash>)`.

Those choices are reasonable because we know where the approximation occurs. Also, it would show up directly in the error variant `IncorrectHash` or `IncorrectHashes`, expressing that something is wrong without exposing the exact reason for the error that needs to be dug.

That is an honest simplification. A dishonest simplification is to hide the error behind a generic `ViewError` or a variant of the same idea.

linera-storage/src/lib.rs
MathieuDutSik: First of all, I did try to implement a `StorageError` in my initial refactoring that led to PR #4014. If I ended up doing something different, it is because it resulted in an unwieldly code that would hardly be called a simplification.

The explanation of the PR are I think clear and apply just as well to this case: It is simply not realistic to have a generic `StorageError` type for `linera-storage`. Now, what could make sense is the following:
```rust
enum ReadCertificateError {
  ViewError(#[from] ViewError),
  IncorrectHashes(Vec<CryptoHash>),
}
```

This is about equivalent to returning a vector of options. I understand that having a
```rust
Result<Result<Certificate, ReadCertificateError>, ViewError>
```
is not something we want to develop in the code even if that leads to in my opinion neat construction `??`.

But what I am sure is that returning an `enum StorageError` is not adequate since it requires handling a number of error scenarios that simply cannot occur in some code.

linera-client/src/unit_tests/chain_listener.rs
```diff
@@ -221,8 +221,10 @@ async fn test_chain_listener_admin_chain() -> anyhow::Result<()> {
     for i in 0.. {
         linera_base::time::timer::sleep(Duration::from_secs(i)).await;
         let result = storage.read_certificate(certificate.hash()).await;
-        if result.ok().as_ref() == Some(&certificate) {
-            break;
+        if let Ok(result) = result.as_ref() {
```
afck: Actually, we should probably just `unwrap` here: I don't think we expect any other errors in this test?

linera-storage/src/db_storage.rs
```diff
@@ -937,20 +937,20 @@ where
...
         let cert = bcs::from_bytes::<LiteCertificate>(cert_bytes)?;
         let value = bcs::from_bytes::<ConfirmedBlock>(value_bytes)?;
-        assert_eq!(value.hash(), hash);
-        let certificate = cert
-            .with_value(value)
-            .ok_or(ViewError::InconsistentEntries)?;
-        Ok(certificate)
+        let value_hash = value.hash();
+        if value_hash != hash {
+            return Ok(None);
```
afck: This one should actually be an error: Storage doesn't guarantee that we have every certificate, but it _does_ guarantee that we store all blocks and certificates under the correct hash.

linera-service/src/exporter/common.rs
```diff
@@ -17,9 +17,12 @@ pub(crate) enum ExporterError {
     #[error("received an invalid notification.")]
     BadNotification(BadNotificationKind),
 
-    #[error("unable to load the exporter state")]
+    #[error("unable to load the exporter state: {0}")]
     StateError(ViewError),
 
+    #[error("Missing or invalid certificate: {0}")]
```
afck: Let's make sure storage only returns `Ok(None)` if the certificate can't be found.

```suggestion
    #[error("Missing certificate: {0}")]
```

linera-core/src/worker.rs
```diff
@@ -152,6 +152,9 @@ pub enum WorkerError {
     #[error(transparent)]
     ViewError(#[from] ViewError),
 
+    #[error("The certificates are missing or inconsistent: {0:?}")]
```
afck:
```suggestion
    #[error("Certificates are in confirmed_log but not in storage: {0:?}")]
```

linera-core/src/chain_worker/state/mod.rs
```diff
@@ -501,7 +502,17 @@ where
                     })?,
             );
         }
-        let certificates = self.storage.read_certificates(hashes).await?;
+        let certificates = self.storage.read_certificates(hashes.clone()).await?;
+        let (certificates, invalid_certificates) = certificates
+            .into_iter()
+            .zip(hashes)
+            .partition_map::<Vec<_>, Vec<_>, _, _, _>(|(certificate, hash)| match certificate {
```
afck: This is done in eight places now. Let's deduplicate that. Maybe there should be a helper function in storage for this after all?

linera-service/src/exporter/common.rs
MathieuDutSik: I agree that the error scenario of missing certificates is the most common one.
However, there are other failure scenarios that can result in a `None`.
I understand that putting the whole set of possibilities is suboptimal, as was my first idea.

But I think we should be clear that missing is simply one failure scenario.
Therefore, I prefer that the error message reflect that.

linera-core/src/chain_worker/state/mod.rs
MathieuDutSik: Yes, that was the reason for having the previous design.
But the unbundling of the `try_join_all` did create a use case where returning a `Vec<Option<_>>` is the right way.

So, yes, for a helper function.

linera-service/src/exporter/common.rs
afck: I think we should rather make every other kind of error a `ViewError` or `StorageError`. `None` should _only_ mean that it wasn't found.

MathieuDutSik: Well, we are coming around to my previous design:
* Putting errors in `ViewError` which are not at the level of `ViewError` is precisely the problem that this PR seeks to address (partially, there are other items).
* Creating a `StorageError` for all kinds of possible errors relevant to `linera-storage` has the problem that was explained in the previous PRs: We have to process the errors.

For me, having a `None` if the retrieval failed for some reason was fine. However, if that is not so in your opinion, then there are only a few solutions possible:
* Have an `assert` in the code. This is currently the case for one of the error scenario.
* Have an error case that is more complex than the `Option`. However, that quickly ends up being something like an error, and so the `Result<Result<_,_>, _>` of my previous design.

afck: My vote would be to add a `ViewError` variant for it for now, and then later move towards a `StorageError`. With that, I wouldn't return `Result<Result<_, _>, _>`, but only a `Result<_, StorageError>` and a `StorageError::View` variant.

But maybe it _does_ need more discussion with the team, after all.

MathieuDutSik: We are already having an `InconsistentEntries` in `ViewError` so there is nothing to add. Unless you want to add another variant corresponding to the `assert_eq!(value.hash(), hash);` of course.

The problem of `StorageError` is that we put many different errors together that have no place in being there.

If the `Result<Result<_,_>,_>` is inadequate, then we can put a
```rust
enum ReadCertificatesError {
  ViewError(ViewError),
  NotFound(CryptoHash),
  InconsistentEntries(CryptoHash),
}

async fn read_certificates(....) -> Result<Vec<_>, ReadCertificatesError>
```

The problem is that those different error cases have to be handled separately. And so we cannot use the `?` operator.

afck: Then I'd just return `ViewError::InconsistentEntries` in that case for now, and keep the `None` only for the case where the entry (or one of them) doesn't exist.

MathieuDutSik: Ok.
That is indeed the only choice possible with the constraints imposed to improve things a little.

linera-storage/src/db_storage.rs

linera-service/src/exporter/common.rs
MathieuDutSik: Ok, implemented accordingly.

.github/workflows/scylladb.yml
```diff
@@ -36,7 +36,7 @@ jobs:
 
   test:
     runs-on: ubuntu-latest-8-cores
-    timeout-minutes: 40
+    timeout-minutes: 50
```
afck, line 39: (Did this change make it slower, or is it just generally flaky because it takes ~40 minutes?)

linera-core/src/node.rs
```diff
@@ -194,6 +194,9 @@ pub enum NodeError {
     #[error("Worker error: {error}")]
     WorkerError { error: String },
 
+    #[error("Missing certificates: {error}")]
+    ReadCertificatesError { error: String },
```
afck: Let's not introduce error variants that are only created in tests. You could just use `MissingCertificateValue` in the test.

.github/workflows/scylladb.yml
MathieuDutSik: Generally flaky.

linera-core/src/node.rs
MathieuDutSik: Ok.
I introduce a panic statement.

linera-client/src/unit_tests/chain_listener.rs
MathieuDutSik: I used `?` since the test function returns `anyhow::Result<()>`.

linera-client/src/unit_tests/chain_listener.rs
```diff
@@ -220,8 +220,8 @@ async fn test_chain_listener_admin_chain() -> anyhow::Result<()> {
         .unwrap();
     for i in 0.. {
         linera_base::time::timer::sleep(Duration::from_secs(i)).await;
-        let result = storage.read_certificate(certificate.hash()).await;
-        if result.ok().as_ref() == Some(&certificate) {
+        let result = storage.read_certificate(certificate.hash()).await?;
+        if result == Some(certificate.clone()) {
```
afck:
```suggestion
        if result.as_ref() == Some(&certificate) {
```

(Or `is_some_and`, if you prefer.)

afck (approved):

linera-client/src/unit_tests/chain_listener.rs
MathieuDutSik: Ok, thanks.

PR #4068, opened 2025-06-05
fix spelling issue  lib.rs

davidjsonn:
Hey team! Fixed error in linera-execution/src/lib.rs
`implementor` - `implementer`


ma2bd:
both spelling are valid

PR #4073, opened 2025-06-06
Removed unused method, ignore error when compiled for docker

deuszx:
## Motivation

CI started to fail on `compose` jobs, complaining that methods are unused.

## Proposal

Indeed, `receive_and_preprocess_checked_certificate` was unused - so that one is removed. `receive_sender_certificate` is used but when built for `docker` job, it complains about it regardless. Just ignore the error for now with `#[allow(dead_code)]` for this method.

## Test Plan

CI

## Release Plan

- Nothing to do
## Links


linera-core/src/client/mod.rs
```diff
@@ -721,6 +721,7 @@ impl<Env: Environment> Client<Env> {
 
     /// Processes the confirmed block in the local node without executing it.
     #[instrument(level = "trace", skip_all)]
+    #[allow(dead_code)] // Otherwise CI fails when built for docker.
```
ma2bd, line 724: why would docker not need this function?

ma2bd: (or why would the `dead_code` linter have false positives?)

PR #4075, opened 2025-06-06
Make sure that blobs downloaded by the client are validated

### LINKED ISSUE #2351: Don't set the committee before checking it in `process_confirmed_block`.

bart-linera:
## Motivation

#3787 left a potential security issue: when a client was downloading a missing `ChainDescription` blob, it would just download a blob, but wouldn't make sure that it was legitimately created on another chain.

## Proposal

Whenever a `ChainDescription` is fetched, fetch the certificate for the block that created it and validate it against committees known from the admin chain.

## Test Plan

CI should catch regressions.

## Release Plan

## Links

- Closes #2351 

## Notes

The `TODO` in `attempted_changes.rs` has been removed after verifying that the safety issue is no longer present there: the committees are taken from a blob, but that blob can only exist on the validator if it has been legitimately created on another chain.

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -270,10 +270,12 @@ impl<Env: Environment> Client<Env> {
                 // This should be a single blob: the ChainDescription of the chain we're
                 // fetching the info for.
                 assert_eq!(blob_ids.len(), 1);
-                let chain_desc_blob = self.fetch_blob(blob_ids[0], validators).await?;
+                let chain_desc_blob = self
+                    .update_local_node_with_blobs_from(blob_ids, validators)
+                    .await?;
                 self.local_node
                     .storage_client()
-                    .write_blobs(&[chain_desc_blob])
+                    .write_blobs(&chain_desc_blob)
                     .await?;
                 self.local_node.chain_info(chain_id).await
             }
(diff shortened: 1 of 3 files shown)


linera-core/src/client/mod.rs
```diff
@@ -448,6 +453,42 @@ impl<Env: Environment> Client<Env> {
...
+        // replaced with something better
+        for node in nodes {
+            let blob_certificate = match node.download_certificate_for_blob(blob_id).await {
+                Ok(certificate) => certificate,
+                Err(_) => continue,
+            };
+            let Some(blob) = blob_certificate
+                .value()
+                .block()
+                .created_blobs()
```
afck: Since this function doesn't assume it's a chain description blob, we shouldn't assume it's in `created_blobs`. Could just be a _required_ blob.

linera-core/src/client/mod.rs
```diff
@@ -448,6 +453,42 @@ impl<Env: Environment> Client<Env> {
             .collect())
     }
 
+    async fn fetch_blob(
```
afck: Isn't this doing the same as `update_local_node_with_blobs_from`, but serially instead of in parallel?

bart-linera: Indeed! :man_facepalming: Well, it also returns the blob, but it can just be read afterwards if needed.

bart-linera: Ah, no, wait, there is a difference: `update_local_node_with_blobs_from` works for multiple blobs and a single validator, while `fetch_blob` works for a single blob and multiple validators. I guess ideally we'd just have a function for multiple blobs and multiple validators, but that's what the TODO is about, after all :grimacing:

afck: Yes, I'd prefer generalizing over duplicating. And we can make the other function return the blob, too?

lerlar:
required blob

afck:
Closing in favor of #4082.

Issue #3902, opened 2025-05-07
Optimize `synchronize_chain_state`/`download_certificates`.


afck:
These two `ChainClient` functions do almost the same thing, and both not in an optimal way:
* `synchronize_chain_state` downloads the whole chain from _all_ validators in parallel (wasteful) until a quorum of tasks have completed.
* `download_certificates` goes through the validators sequentially in random order (bad if the first one is slow) and attempts to download the certificates only from that one.

This code should be deduplicated and optimized, so that the client is neither slowed down by the slowest validator, nor do we redundantly download everything from everyone.

## Sketch of a Design

Create a task for each validator:
* First, get the chain info to learn that validator's `next_block_height`.
* Then repeatedly download a range of certificates. (See below.)

In another task, process the certificates in order, whenever they become available.

Stop downloading but finish processing the downloaded certificates as soon as there is a height H such that:
* We have downloaded all certificates up to (excluding) H,
* a quorum of validators has returned a chain info, and
* Fewer than ⅓ of validators by weight reports a `next_block_height` > H.

To select a range to download for a validator, try to optimize, for example:
* Always use a range that that validator actually claims to have (based on their `next_block_height`).
* If available, choose a range that has not been assigned to any validator yet.
* Otherwise choose a range that has been assigned to the lowest number of validators.

All of this requires a lot of coordination between the tasks; we should probably prefer channels over shared state if feasible.

PR #4082, opened 2025-06-09
Make sure that blobs downloaded by the client are validated

### LINKED ISSUE #2351: Don't set the committee before checking it in `process_confirmed_block`.

afck:
This is a completion of @bart-linera's #4075:

## Motivation

#3787 left a potential security issue: when a client was downloading a missing `ChainDescription` blob, it would just download a blob, but wouldn't make sure that it was legitimately created on another chain.

## Proposal

Whenever a `ChainDescription` is fetched, fetch the certificate for the block that created it and validate it against committees known from the admin chain.

## Test Plan

CI should catch regressions.

## Release Plan

## Links

- Closes #2351 

## Notes

The `TODO` in `attempted_changes.rs` has been removed after verifying that the safety issue is no longer present there: the committees are taken from a blob, but that blob can only exist on the validator if it has been legitimately created on another chain.

### DIFF
--- a/linera-client/src/client_context.rs
+++ b/linera-client/src/client_context.rs
@@ -56,7 +56,7 @@ use {
 };
 
 use crate::{
-    chain_listener::{self, ClientContext as _, ClientContextExt as _},
+    chain_listener::{self, ClientContext as _},
     client_options::{ChainOwnershipConfig, ClientContextOptions},
     error, util,
     wallet::{UserChain, Wallet},
@@ -346,7 +346,8 @@ impl<Env: Environment, W: Persist<Target = Wallet>> ClientContext<Env, W> {
         owner: AccountOwner,
     ) -> Result<(), Error> {
         self.client.track_chain(chain_id);
-        let chain_description = self.chain_description(chain_id).await?;
+        let client = self.make_chain_client(chain_id);
+        let chain_description = client.ensure_has_chain_description().await?;
         let config = chain_description.config();
 
         if !config.ownership.verify_owner(&owner) {
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -270,10 +270,12 @@ impl<Env: Environment> Client<Env> {
                 // This should be a single blob: the ChainDescription of the chain we're
                 // fetching the info for.
                 assert_eq!(blob_ids.len(), 1);
-                let chain_desc_blob = self.fetch_blob(blob_ids[0], validators).await?;
+                let chain_desc_blob = self
+                    .update_local_node_with_blobs_from(blob_ids, validators)
+                    .await?;
                 self.local_node
                     .storage_client()
-                    .write_blobs(&[chain_desc_blob])
+                    .write_blobs(&chain_desc_blob)
                     .await?;
                 self.local_node.chain_info(chain_id).await
             }
@@ -453,42 +455,6 @@ impl<Env: Environment> Client<Env> {
             .collect())
     }
 
-    async fn fetch_blob(
-        &self,
-        blob_id: BlobId,
-        nodes: &[RemoteNode<impl ValidatorNode>],
-    ) -> Result<Blob, LocalNodeError> {
-        // TODO(#3844): a very naive algorithm, just trying the validator nodes one by one - to be
-        // replaced with something better
-        for node in nodes {
-            let blob_certificate = match node.download_certificate_for_blob(blob_id).await {
-                Ok(certificate) => certificate,
-                Err(_) => continue,
-            };
-            let Some(blob) = blob_certificate
-                .value()
-                .block()
-                .created_blobs()
-                .get(&blob_id)
-                .cloned()
-            else {
-                continue;
-            };
-            if self
-                .receive_sender_certificate(
-                    blob_certificate,
-                    ReceiveCertificateMode::NeedsCheck,
-                    None,
-                )
-                .await
-                .is_ok()
-            {
-                return Ok(blob);
-            }
-        }
-        Err(LocalNodeError::BlobsNotFound(vec![blob_id]))
-    }
-
     /// Ensures that the client has the `ChainDescription` blob corresponding to this
     /// client's `ChainId`.
     pub async fn ensure_has_chain_description(
@@ -507,7 +473,11 @@ impl<Env: Environment> Client<Env> {
         };
         // Recover history from the current validators, according to the admin chain.
         let nodes = self.validator_nodes().await?;
-        Ok(self.fetch_blob(chain_desc_id, &nodes).await?)
+        Ok(self
+            .update_local_node_with_blobs_from(vec![chain_desc_id], &nodes)
+            .await?
+            .pop()
+            .unwrap())
     }
 
     /// Updates the latest block and next block height and round information from the chain info.
@@ -1093,8 +1063,11 @@ impl<Env: Environment> Client<Env> {
                         }
                     }
                     if let LocalNodeError::BlobsNotFound(blob_ids) = &err {
-                        self.update_local_node_with_blobs_from(blob_ids.clone(), remote_node)
-                            .await?;
+                        self.update_local_node_with_blobs_from(
+                            blob_ids.clone(),
+                            &[remote_node.clone()],
+                        )
+                        .await?;
                         // We found the missing blobs: retry.
                         if let Err(new_err) = self
                             .local_node
@@ -1148,17 +1121,36 @@ impl<Env: Environment> Client<Env> {
     async fn update_local_node_with_blobs_from(
         &self,
         blob_ids: Vec<BlobId>,
-        remote_node: &RemoteNode<Env::ValidatorNode>,
-    ) -> Result<(), ChainClientError> {
+        remote_nodes: &[RemoteNode<impl ValidatorNode>],
+    ) -> Result<Vec<Blob>, LocalNodeError> {
         future::try_join_all(blob_ids.into_iter().map(|blob_id| async move {
-            let certificate = remote_node.download_certificate_for_blob(blob_id).await?;
-            // This will download all ancestors of the certificate and process all of them locally.
-            self.receive_sender_certificate(certificate, ReceiveCertificateMode::NeedsCheck, None)
-                .await
+            for remote_node in remote_nodes {
+                let certificate = match remote_node.download_certificate_for_blob(blob_id).await {
+                    Ok(certificate) => certificate,
+                    Err(_) => continue,
+                };
+                // This will download all ancestors of the certificate and process all of them locally.
+                if self
+                    .receive_sender_certificate(
+                        certificate,
+                        ReceiveCertificateMode::NeedsCheck,
+                        None,
+                    )
+                    .await
+                    .is_ok()
+                {
+                    let blob = self
+                        .local_node
+                        .storage_client()
+                        .read_blob(blob_id)
+                        .await?
+                        .ok_or_else(|| LocalNodeError::BlobsNotFound(vec![blob_id]))?;
+                    return Ok(blob);
+                }
+            }
+            Err(LocalNodeError::BlobsNotFound(vec![blob_id]))
         }))
-        .await?;
-
-        Ok(())
+        .await
     }
 
     /// Downloads and processes confirmed block certificates that use the given blobs.
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -267,15 +267,14 @@ impl<Env: Environment> Client<Env> {
             Err(LocalNodeError::BlobsNotFound(blob_ids)) => {
                 // If the chain is missing then the error is a WorkerError
                 // and so a BlobsNotFound
-                // TODO(#2351): make sure the blobs are legitimate!
-                let blobs = RemoteNode::download_blobs(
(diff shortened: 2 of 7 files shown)


linera-client/src/client_context.rs
```diff
@@ -346,7 +346,8 @@ impl<Env: Environment, W: Persist<Target = Wallet>> ClientContext<Env, W> {
         owner: AccountOwner,
     ) -> Result<(), Error> {
         self.client.track_chain(chain_id);
-        let chain_description = self.chain_description(chain_id).await?;
+        let client = self.make_chain_client(chain_id);
+        let chain_description = client.ensure_has_chain_description().await?;
```
deuszx: It's bit weird that `ensure` returns anything. Maybe rename to `get_chain_description`?

linera-core/src/client/mod.rs
```diff
@@ -462,17 +464,16 @@ impl<Env: Environment> Client<Env> {
...
-        // TODO(#2351): make sure that the blob is legitimately created!
         let nodes = self.validator_nodes().await?;
-        let blob =
-            RemoteNode::download_blob(&nodes, chain_desc_id, self.options.blob_download_timeout)
-                .await
-                .ok_or(LocalNodeError::BlobsNotFound(vec![chain_desc_id]))?;
-        self.local_node.storage_client().write_blob(&blob).await?;
-        Ok(blob)
+        let blob = self
+            .update_local_node_with_blobs_from(vec![chain_desc_id], &nodes)
```
deuszx, line 474: What if the blob here was created when a different `node` were in the committee?

linera-client/src/client_context.rs
afck: …or `fetch_chain_description`? (Even though it only fetches it if we don't have it, of course.)

linera-core/src/client/mod.rs
```diff
@@ -462,17 +464,16 @@ impl<Env: Environment> Client<Env> {
...
+            .await?
+            .pop()
+            .unwrap();
```
deuszx: `unwrap()`? 👀

linera-core/src/client/mod.rs
```diff
@@ -267,15 +267,11 @@ impl<Env: Environment> Client<Env> {
...
-                    &blob_ids,
-                    validators,
-                    self.options.blob_download_timeout,
-                )
-                .await
-                .ok_or(LocalNodeError::BlobsNotFound(blob_ids))?;
-                self.local_node.storage_client().write_blobs(&blobs).await?;
+                // This should be a single blob: the ChainDescription of the chain we're
+                // fetching the info for.
+                assert_eq!(blob_ids.len(), 1);
```
christos-h: Do we want to use  `ensure!` here instead of  `assert_eq`?

linera-core/src/client/mod.rs
```diff
@@ -381,11 +377,17 @@ impl<Env: Environment> Client<Env> {
             let mut result = self.handle_certificate(certificate.clone()).await;
 
             if let Err(LocalNodeError::BlobsNotFound(blob_ids)) = &result {
-                if let Some(blobs) = remote_node.try_download_blobs(blob_ids).await {
-                    // TODO(#2351): Don't store downloaded blobs without certificate.
-                    let _ = self.local_node.store_blobs(&blobs).await;
-                    result = self.handle_certificate(certificate.clone()).await;
+                for blob_id in blob_ids {
```
christos-h: This looks trivially parallelizable?

linera-client/src/client_context.rs
deuszx: Personally, I am not a fan of verb _fetch_ - I think it tries to describe something more than `get` but still is too ambiguous.

linera-core/src/client/mod.rs
afck: I guess we could go through all validators of all past committees… but that goes for sender chains etc. as well. I created https://github.com/linera-io/linera-protocol/issues/4083 to revisit this.

linera-core/src/client/mod.rs
afck: I'll add a comment…

linera-core/src/client/mod.rs
```diff
@@ -1120,17 +1117,36 @@ impl<Env: Environment> Client<Env> {
...
-        remote_node: &RemoteNode<Env::ValidatorNode>,
-    ) -> Result<(), ChainClientError> {
+        remote_nodes: &[RemoteNode<impl ValidatorNode>],
+    ) -> Result<Vec<Blob>, LocalNodeError> {
         future::try_join_all(blob_ids.into_iter().map(|blob_id| async move {
-            let certificate = remote_node.download_certificate_for_blob(blob_id).await?;
-            // This will download all ancestors of the certificate and process all of them locally.
-            self.receive_sender_certificate(certificate, ReceiveCertificateMode::NeedsCheck, None)
-                .await
+            for remote_node in remote_nodes {
```
christos-h: This as well? (Trivially parallelizable).

linera-core/src/client/mod.rs
```diff
@@ -3433,9 +3456,9 @@ impl<Env: Environment> ChainClient<Env> {
         &self,
         chain_id: ChainId,
         local_node: &mut LocalNodeClient<Env::Storage>,
-    ) -> Option<BlockHeight> {
-        let info = self.local_chain_info(chain_id, local_node).await?;
-        Some(info.next_block_height)
+    ) -> BlockHeight {
```
christos-h, line 3465: Nice catch.

linera-core/src/client/mod.rs
deuszx: Or we should be able to assume that if we receive the certificate from the _current_ committee then it had been validated?

linera-core/src/client/mod.rs
afck: Maybe. It _should_ be unreachable, but I can make it an `InternalError`.

linera-core/src/client/mod.rs
afck: That definitely, yes. But in general we should support synchronizing something that only a past committee has, even if none of the current validators have it. (Separate issue, though; that's why I created #4083.)

linera-core/src/client/mod.rs
afck: Or I'll just remove it. Not worth adding another error variant.

linera-core/src/client/mod.rs
afck: Right, this is the tradeoff again between redundantly downloading it from every validator at the same time and waiting for a single slow validator, where we'd really need a more sophisticated algorithm: https://github.com/linera-io/linera-protocol/issues/3902
I'll try to come up with a temporary solution that's at least a bit better than waiting for a single validator again…

christos-h: Ah right my bad. Missed the return condition.

linera-core/src/client/mod.rs
```diff
@@ -260,33 +260,28 @@ impl<Env: Environment> Client<Env> {
...
         chain_id: ChainId,
-        validators: &[RemoteNode<impl ValidatorNode>],
-    ) -> Result<Box<ChainInfo>, LocalNodeError> {
+        validators: &[RemoteNode<Env::ValidatorNode>],
+    ) -> Result<Box<ChainInfo>, ChainClientError> {
         match self.local_node.chain_info(chain_id).await {
             Ok(info) => Ok(info),
             Err(LocalNodeError::BlobsNotFound(blob_ids)) => {
+                // Make sure the admin chain is up to date.
+                self.synchronize_chain_state(self.admin_id).await?;
```
deuszx, line 269: This is potentially a performance regression. Is there a "quick" way to verify if we're up-to-date? Can we optimistically try to skip synchronization with admin chain and somehow detect it?

deuszx, line 269: This is potentially a performance regression. Is there a "quick" way to verify if we're up-to-date? Can we optimistically try to skip synchronization with admin chain and somehow detect it?

linera-core/src/client/mod.rs
```diff
@@ -462,17 +465,17 @@ impl<Env: Environment> Client<Env> {
             .await?;
         if let Some(blob) = blob {
             // We have the blob - return it.
-            return Ok(blob);
+            return Ok(bcs::from_bytes(blob.bytes())?);
         };
         // Recover history from the current validators, according to the admin chain.
-        // TODO(#2351): make sure that the blob is legitimately created!
+        self.synchronize_chain_state(self.admin_id).await?;
```
deuszx, line 471: And again we're synchronizing the admin chain - is this one of the reasons you want to move everything to a single place so that we have a better control over what has been done or what needs to be done?

linera-core/src/client/mod.rs
```diff
@@ -1120,17 +1119,40 @@ impl<Env: Environment> Client<Env> {
...
+    ) -> Result<Vec<Blob>, ChainClientError> {
+        let timeout = self.options.blob_download_timeout;
         future::try_join_all(blob_ids.into_iter().map(|blob_id| async move {
-            let certificate = remote_node.download_certificate_for_blob(blob_id).await?;
-            // This will download all ancestors of the certificate and process all of them locally.
-            self.receive_sender_certificate(certificate, ReceiveCertificateMode::NeedsCheck, None)
-                .await
+            let mut stream = remote_nodes
+                .iter()
+                .zip(0..)
```
deuszx, line 1128: Isn't this the same as `.enumerate()`?

linera-core/src/client/mod.rs
afck: Yes, exactly!
I don't know how to get around these otherwise.

E.g. we shouldn't do this at all if we're in "connected" mode, i.e. a chain listener is running. I'm thinking about writing a design doc for a "local node synchronizer" that does all this properly. I'd rather not try to add more ad-hoc case distinctions here for now.

linera-core/src/client/mod.rs
afck: (See below.)

linera-core/src/client/mod.rs
afck: Almost, but I want `u32`s instead of `usize`s.

linera-core/src/client/mod.rs
```diff
@@ -1120,17 +1119,40 @@ impl<Env: Environment> Client<Env> {
...
+                if let Ok(blob) = maybe_blob {
+                    return Ok(blob);
+                }
+            }
+            Err(LocalNodeError::BlobsNotFound(vec![blob_id]).into())
         }))
-        .await?;
-
-        Ok(())
+        .await
```
deuszx, line 1155: Do I read this correctly - we try to download all blobs in parallel and within a "single blob context" we will try to download it from different validators sequentially, by scheduling it to start after an ever-growing timeout?

Would it be more readable if you extract the `async move { }` closure to a function that tries a new node only if the previous one failed? That way the retries can be "tighter" - i.e. we won't wait a `timeout * i * i` but can start immediately (if previous node didn't have the blob).

linera-core/src/client/mod.rs
deuszx: You can always downcast `usize` to `u32` 🤔

linera-core/src/client/mod.rs
deuszx: Definitely.

linera-core/src/client/mod.rs
afck: Yes, but that's more code, and introduces an unnecessary `u32::try_from` that can fail—why not use `u32` right away?

deuszx: I thought `u32` to `usize` cannot fail 👍 

Anyway, already resolved.

linera-core/src/client/mod.rs
afck: > only if the previous one failed

But that's what we had before, and what is causing the client performance issues: If the first validator we ask takes a very long time to respond or fail, we just keep waiting.

But you're right that we could try the next one _immediately_ after the earlier ones failed, and skip the timeout.

Anyway, this affects not only this particular place in the code, but all of them. Again: This is why I think we need some sort of local node synchronizer that has all that logic in one place.

deuszx: >  If the first validator we ask takes a very long time to respond or fail, we just keep waiting.

So we have a timeout. We try the next if one of the two happens
1. Validator responds with `BlobsNotFound`
2. Request times out

deuszx (approved):
The last comment is more like an idea for improvement - not a blocker. You can decide what to do with it.

linera-core/src/client/mod.rs
afck: Let's postpone this. I have so many questions and comments about this, it'll end up more complex than this whole PR.

Issue #4077, opened 2025-06-09
Remove Signer::get_public_key


deuszx:
This method is a blocker for integrating more EVM wallest - like MetaMask - which do not expose the full public key to the clients.

PR #4088, opened 2025-06-09
Remove get_public_key method from Signer.

deuszx:
## Motivation

The `get_public_key` method in the `Signer` API is a blocker for implementing that trait for EVM wallets - like MetaMask - which do not expose the public key to the users.

## Proposal

Remove `Signer::get_public_key`. 

There were few palces where we used `get_public_key()` via `ChainClient` only to map that into `AccountOwner` - these were updated to call `client.identity()` directly instead.

The biggest change is in the `BlockProposal` struct, which also carried `AccountPublicKey` used for verifying the signature later on. Here the public keys were "inlined" into the `AccountSignature`: for `Ed25519 and `Secp256k1` cases this meant simply adding the key to those variants:
```rust
pub enum AccountSignature {
    Ed25519 {
        signature: ed25519::Ed25519Signature,
        public_key: ed25519::Ed25519PublicKey,
    },
    Secp256k1 {
        signature: secp256k1::Secp256k1Signature,
        public_key: secp256k1::Secp256k1PublicKey,
    },
```
and use the `public_key`s to verify the signatures later.

But for the EVM variant we wouldn't be able to do that (for the same reason described in the referrenced issue). Here we inline the `Address` (EVM address):
```rust
    EvmSecp256k1 {
        signature: secp256k1::evm::EvmSignature,
        address: Address,
    },
```

Inlining the `Address` shouldn't be necessary as EVM signatures allow for recovering the address but for that we need `signature` and `value` (for which that signature was produced) but we don't always have that - `BlockProposal`'s `OriginalProposal` is `AccountSignature` only so there's no way to recover the `Address`.

## Test Plan

CI

## Release Plan

## Links

Closes https://github.com/linera-io/linera-protocol/issues/4077


linera-base/src/crypto/mod.rs
```diff
@@ -197,42 +233,20 @@ impl AccountPublicKey {
...
+        match self {
+            AccountSignature::Ed25519 {
+                signature,
+                public_key,
+            } => signature.check(value, public_key),
+            AccountSignature::Secp256k1 {
+                signature,
+                public_key,
+            } => signature.check(value, public_key),
+            AccountSignature::EvmSecp256k1 { signature, .. } => signature.check_with_recover(value),
```
ma2bd: The fact that we're not using the address is suspicious

linera-base/src/crypto/secp256k1/evm.rs
```diff
@@ -399,6 +409,15 @@ impl EvmSignature {
         self.verify_inner::<T>(prehash, author)
     }
 
+    /// Checks a signature against a recovered public key.
+    pub fn check_with_recover<'de, T>(&self, value: &T) -> Result<(), CryptoError>
```
ma2bd: This should take an expected `address` as argument and verify that the reconstructed `public_key` matches the address.

linera-base/src/crypto/mod.rs
```diff
@@ -197,42 +233,20 @@ impl AccountPublicKey {
 
 impl AccountSignature {
     /// Verifies the signature for the `value` using the provided `public_key`.
-    pub fn verify<'de, T>(&self, value: &T, author: AccountPublicKey) -> Result<(), CryptoError>
```
ma2bd, line 200: Would it make sense to take the expected address as argument? Otherwise, there will always be an external step where you verify that the public key matches the expected address.

linera-base/src/crypto/mod.rs
deuszx: I agree, I was on the fence about this but if others think so as well - I'll do it. It should address the other comment as well.

linera-base/src/crypto/mod.rs
ma2bd: Ok never mind. I see that you just extract the owner from the signature instead of storing it elsewhere.

ma2bd (changes requested):
back to your queue for the earlier comment

linera-base/src/crypto/ed25519.rs
```diff
@@ -310,7 +310,7 @@ impl Ed25519Signature {
     fn check_internal<'de, T>(
         &self,
         value: &T,
-        author: Ed25519PublicKey,
+        author: &Ed25519PublicKey,
```
ma2bd: Are those not `Copy` ?

deuszx: Hmm, they are. I'd have thought clippy will complain

linera-base/src/crypto/mod.rs
deuszx: Done in [Verify that recovered address matches the sender](https://github.com/linera-io/linera-protocol/pull/4088/commits/4de775be41e8c589716e6c6ac0b7742e38ef9336)

linera-base/src/crypto/secp256k1/evm.rs
deuszx: Done in [Verify that recovered address matches the sender](https://github.com/linera-io/linera-protocol/pull/4088/commits/4de775be41e8c589716e6c6ac0b7742e38ef9336)

linera-base/src/crypto/ed25519.rs
deuszx: Done in [Pass public keys by value](https://github.com/linera-io/linera-protocol/pull/4088/commits/9877a2424361f87dbc185b84d253c8acb76c088c)

linera-base/src/crypto/mod.rs
```diff
@@ -86,11 +88,27 @@ pub enum AccountSecretKey {
...
+        public_key: secp256k1::Secp256k1PublicKey,
+    },
     /// EVM secp256k1 signature.
-    EvmSecp256k1(secp256k1::evm::EvmSignature),
+    // No public key b/c it can be recovered from the signature.
+    EvmSecp256k1 {
+        /// Signature of the value.
+        signature: secp256k1::evm::EvmSignature,
+        /// The EVM address of the signer.
+        address: [u8; 20],
```
afck, line 110: Do we need that? I thought in this case it's computable from `signature` alone.

linera-base/src/crypto/secp256k1/evm.rs
```diff
@@ -391,14 +401,33 @@ impl EvmSignature {
...
+        let public_key = EvmPublicKey(self.0.recover_from_msg(prehash).unwrap());
+        if public_key.address().0 != sender_address {
+            return Err(CryptoError::InvalidSignature {
+                error: "Recovered public key does not match sender address".to_string(),
+                type_name: T::type_name().to_string(),
+            });
+        }
+        self.verify_inner::<T>(prehash, public_key)
+    }
     /// Verifies a batch of signatures.
```
afck, line 460:
```suggestion

    /// Verifies a batch of signatures.
```

linera-chain/src/data_types.rs
```diff
@@ -779,7 +774,10 @@ mod signing {
 
         // personal_sign of the `proposal_hash` done via MetaMask.
         // Wrap with proper variant so that bytes match (include the enum variant tag).
-        let metamask_signature = AccountSignature::EvmSecp256k1(EvmSignature::from_str("f2d8afcd51d0f947f5c5e31ac1db73ec5306163af7949b3bb265ba53d03374b04b1e909007b555caf098da1aded29c600bee391c6ee8b4d0962a29044555796d1b").unwrap());
+        let metamask_signature = AccountSignature::EvmSecp256k1 {
+            signature: EvmSignature::from_str("f2d8afcd51d0f947f5c5e31ac1db73ec5306163af7949b3bb265ba53d03374b04b1e909007b555caf098da1aded29c600bee391c6ee8b4d0962a29044555796d1b").unwrap(),
```
afck: (I'd break that string literal over multiple lines.)

linera-base/src/crypto/mod.rs
deuszx: It's recoverable from the signature AND the message it signed and sometimes we don't have the original message and just the signature.

afck: Where do we have only the signature and not the message? I'd imagine that everywhere where we care about the signer we also have the message.

I'm just wondering if we can save those 20 bytes.

It's also not great that we're adding 20 or even 32 bytes to every validator signature in every certificate now, even though we definitely don't need them there.

linera-base/src/crypto/secp256k1/evm.rs
deuszx: Not necessary with [Recover address for Evm signatures](https://github.com/linera-io/linera-protocol/pull/4088/commits/0f0732603c3d239a7c03e3eee1f68a25c99c3b9a)

linera-base/src/crypto/mod.rs
deuszx: Done in [Recover address for Evm signatures](https://github.com/linera-io/linera-protocol/pull/4088/commits/0f0732603c3d239a7c03e3eee1f68a25c99c3b9a)

linera-core/src/chain_worker/state/temporary_changes.rs
```diff
@@ -189,7 +188,16 @@ where
...
-                let super_owner = signature.owner();
+                let old_proposal = BlockProposal {
+                    content: ProposalContent {
+                        block: proposal.content.block.clone(),
+                        round: Round::Fast,
+                        outcome: None,
+                    },
+                    signature: *signature,
+                    original_proposal: None,
+                };
```
afck, line 199: ~It looks like we only need `content`, not the whole `BlockProposal`?~ (Ah, no, makes sense.)

linera-base/src/crypto/mod.rs
```diff
@@ -86,10 +88,21 @@ pub enum AccountSecretKey {
...
     /// secp256k1 signature.
-    Secp256k1(secp256k1::Secp256k1Signature),
+    Secp256k1 {
+        /// Signature of the value.
+        signature: secp256k1::Secp256k1Signature,
+        /// Public key of the signer.
+        public_key: secp256k1::Secp256k1PublicKey,
+    },
     /// EVM secp256k1 signature.
+    // No public key b/c it can be recovered from the signature.
```
ma2bd: I kinda liked having the address here to make sure the `owner()` method below is fast (and not using a EC 
crypto library). This could be useful for indexing blocks for instance.

linera-base/src/crypto/mod.rs
ma2bd: ok so this is where I don't necessarily agree with @afck apparently :)

afck: Sorry, I hadn't realized that the _validator_ signatures don't even use this type, so we won't waste any bytes there!

I don't have a strong opinion about this. I was thinking we'd not only save the 20 bytes but also make it less likely that we forget one of the two comparisons: the expected address with the `address` field here, and the `address` field with the recovered key. But I'm fine with reverting the change, too.

linera-base/src/crypto/secp256k1/evm.rs
```diff
@@ -114,6 +114,34 @@ impl EvmPublicKey {
...
+    where
+        T: BcsSignable<'de>,
+    {
+        let message = CryptoHash::new(value).as_bytes().0;
+        let public_key =
+            signature
+                .0
+                .recover_from_msg(message)
+                .map_err(|_| CryptoError::InvalidSignature {
+                    error: "Failed to recover public key from signature".to_string(),
```
ma2bd, line 139: nit: Don't we use enums to avoid strings in the middle of an error value?

linera-base/src/crypto/secp256k1/evm.rs
```diff
@@ -391,14 +424,23 @@ impl EvmSignature {
...
     }
 
+    /// Checks a signature against a recovered public key.
+    pub fn check_with_recover<'de, T>(&self, value: &T) -> Result<(), CryptoError>
+    where
+        T: BcsSignable<'de> + fmt::Debug,
+    {
+        let prehash = CryptoHash::new(value).as_bytes().0;
+        let public_key = EvmPublicKey(self.0.recover_from_msg(prehash).unwrap());
+        self.verify_inner::<T>(prehash, public_key)
```
ma2bd: Please check a textbook about ECDSA but I believe the `.unwrap()` will fail if the signature is invalid. And conversely if the recovery succeeds, `verify_inner` always succeeds. Therefore it should be omitted.

linera-base/src/crypto/mod.rs
deuszx: Agreed to bring back the inlined address.

linera-base/src/crypto/secp256k1/evm.rs
deuszx: We do, I just didn't want to add a new variant for every kind of error if there is a more generic one that can capture this already (and `InvalidSignature` variant was here before my time already)

linera-base/src/crypto/secp256k1/evm.rs
deuszx: Done in [Include EVM address in EvmSignature explicitly](https://github.com/linera-io/linera-protocol/pull/4088/commits/e0240d1af37378bdf380aa53c4b5ce204d0ce088)

linera-base/src/crypto/mod.rs
deuszx: Done in [Include EVM address in EvmSignature explicitly](https://github.com/linera-io/linera-protocol/pull/4088/commits/e0240d1af37378bdf380aa53c4b5ce204d0ce088)

linera-base/src/crypto/secp256k1/evm.rs
deuszx: Done in [Include EVM address in EvmSignature explicitly](https://github.com/linera-io/linera-protocol/pull/4088/commits/e0240d1af37378bdf380aa53c4b5ce204d0ce088)

linera-base/src/crypto/mod.rs
```diff
@@ -197,41 +232,25 @@ impl AccountPublicKey {
...
+            } => signature.check(value, *public_key),
+            AccountSignature::Secp256k1 {
+                signature,
+                public_key,
+            } => signature.check(value, *public_key),
+            AccountSignature::EvmSecp256k1 {
+                signature,
+                address: sender_address,
+            } => {
+                let _ = signature.check_with_recover(value, *sender_address);
```
ma2bd:
```suggestion
                signature.check_with_recover(value, *sender_address)?;
```

deuszx: Good catch 😬 I totally missed that.

deuszx: Fixed [Propagate the error](https://github.com/linera-io/linera-protocol/pull/4088/commits/79f1abab191014996d6a5c197213ad856f0982aa)

linera-base/src/crypto/secp256k1/evm.rs
```diff
@@ -114,6 +114,34 @@ impl EvmPublicKey {
             }
         }
     }
+
+    /// Returns an EVM address for the public key.
+    pub fn address(&self) -> alloy_primitives::Address {
+        alloy_primitives::Address::from_public_key(&self.0)
+    }
+
+    /// Recover the public key from the signature and the value.
```
afck:
```suggestion
    /// Recovers the public key from the signature and the value.
```

linera-base/src/crypto/secp256k1/evm.rs
```diff
@@ -114,6 +114,34 @@ impl EvmPublicKey {
...
+        T: BcsSignable<'de>,
+    {
+        let message = CryptoHash::new(value).as_bytes().0;
+        let public_key =
+            signature
+                .0
+                .recover_from_msg(message)
+                .map_err(|_| CryptoError::InvalidSignature {
+                    error: "Failed to recover public key from signature".to_string(),
+                    type_name: Self::type_name().to_string(),
```
afck, line 140: (Unrelated, but: Could this be a `&'static str` instead of a `String`?)

linera-base/src/crypto/secp256k1/evm.rs
```diff
@@ -657,4 +721,22 @@ mod tests {
...
+            secp256k1::evm::{EvmKeyPair, EvmPublicKey, EvmSignature},
+            CryptoHash, TestString,
+        };
+        let key_pair = EvmKeyPair::generate();
+        let address = key_pair.public_key.address();
+        let msg = TestString("hello".into());
+        let prehash = CryptoHash::new(&msg);
+        let sig = EvmSignature::new(prehash, &key_pair.secret_key);
+
+        assert!(sig.check_with_recover(&msg, address.0 .0).is_ok());
```
afck: I think this would print the error if it fails:

```suggestion
        sig.check_with_recover(&msg, address.0 .0).unwrap();
```

linera-base/src/identifiers.rs
```diff
@@ -51,6 +51,16 @@ impl AccountOwner {
     }
 }
 
+impl std::fmt::Debug for AccountOwner {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            AccountOwner::Reserved(value) => write!(f, "Reserved({})", value),
+            AccountOwner::Address32(address) => write!(f, "Address32({})", address),
+            AccountOwner::Address20(address) => write!(f, "Address20({})", hex::encode(address)),
```
afck: Could we keep the derive, `use custom_debug_derive::Debug`, and annotate the field with `#[debug(with = "hex_debug")]`?

linera-chain/src/data_types.rs
```diff
@@ -779,9 +778,22 @@ mod signing {
 
         // personal_sign of the `proposal_hash` done via MetaMask.
         // Wrap with proper variant so that bytes match (include the enum variant tag).
-        let metamask_signature = AccountSignature::EvmSecp256k1(EvmSignature::from_str("f2d8afcd51d0f947f5c5e31ac1db73ec5306163af7949b3bb265ba53d03374b04b1e909007b555caf098da1aded29c600bee391c6ee8b4d0962a29044555796d1b").unwrap());
+        let signature = EvmSignature::from_str("f2d8afcd51d0f947f5c5e31ac1db73ec5306163af7949b3bb265ba53d03374b04b1e909007b555caf098da1aded29c600bee391c6ee8b4d0962a29044555796d1b").unwrap();
```
afck: (Line too long.)

linera-chain/src/data_types.rs
```diff
@@ -779,9 +778,22 @@ mod signing {
...
+        let metamask_signature = AccountSignature::EvmSecp256k1 {
+            signature,
+            address: address.0 .0,
+        };
 
         let signature = signer.sign(&proposal);
         assert_eq!(signature, metamask_signature);
+
+        assert_eq!(signature.owner(), public_key.into(),);
```
afck:
```suggestion
        assert_eq!(signature.owner(), public_key.into());
```

linera-chain/src/manager.rs
```diff
@@ -606,33 +602,47 @@ where
...
+        tracing::info!(
+            "Chain's super owners: {:?}, proposal_owner: {:?}",
+            self.ownership.get().super_owners,
+            proposal_owner
+        );
+        tracing::info!(
+            "Chain's owners: {:?}, proposal_owner: {:?}",
+            self.ownership.get().owners,
+            proposal_owner
+        );
```
afck:
```suggestion
```

(Or make it `debug` or `tracing`, and just print the `ownership`; but I'd just remove them.)

linera-base/src/identifiers.rs
deuszx: Yes, we can. I actually think this is already done as I did it for a different enum later on.

linera-chain/src/manager.rs
deuszx: Definitely, this shouldn't get in.

linera-base/src/crypto/secp256k1/evm.rs
Twey: This is what [`thiserror-context`](https://docs.rs/thiserror-context/latest/thiserror_context/) is for: it allows adding string (or other `Display`) context to any error variant.  I think the usage I've settled on is that you have one enum variant for each thing goes wrong (i.e. things the receiver of the error will want to pattern-match on) and use `thiserror-context` to explain what you were doing when the error happened (for debugging).  Alternatively we could use [`tracing_error`](https://docs.rs/tracing-error/latest/tracing_error/) to capture the tracing context we already have, but I worry that there's some context we'd want to capture _only_ on error, and I can't see a way to do that with `tracing`.  I'm not sure which approach is better for us overall, but we are already using `thiserror-context` in the codebase.

linera-base/src/crypto/secp256k1/evm.rs
Twey: A good candidate for context of one kind or another.  The receiver can't meaningfully pattern-match on either of these so there's no point having them in the variant.

linera-base/src/crypto/secp256k1/evm.rs
Twey: Or `assert_matches!(Ok(_), sig.check_with_recover(&msg, address.0 .0))`

deuszx: ~~I think all options are equally reasonable and it's a matter of test. All of this would result in a meaningful error message if the test fails.~~

linera-base/src/crypto/secp256k1/evm.rs
deuszx: I am not changing the behavior - this was already here. If we want to change this - I'd suggest doing this in a different PR. This one isn't about it.

linera-base/src/crypto/secp256k1/evm.rs
deuszx: Actually, they're not equivalent - in my version it won't print what was the `Err(_)` variant.

I'll update.

linera-base/src/crypto/secp256k1/evm.rs
Twey: Maybe the answer to that problem is that these spans are (a subset of) `DEBUG`-level spans, and the facility for displaying errors to the users (be it printing or converting to `anyhow::Error`) should make sure to capture spans at `DEBUG` level from the `tracing_error::SpanTrace` regardless of the current subscriber level.

deuszx: I think we don't expect clients to automatically recover from `CryptoError`s. Right now, when integrating with `linera-web`, I am getting the full error message in the browser console.

linera-base/src/crypto/secp256k1/evm.rs
```diff
@@ -65,16 +65,15 @@ pub struct EvmKeyPair {
...
-pub struct EvmSignature(pub(crate) Signature);
+pub struct EvmSignature(pub Signature);
 
 impl FromStr for EvmSignature {
     type Err = CryptoError;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         // If the string starts with "0x", we remove it before decoding.
         let bytes = hex::decode(s.strip_prefix("0x").unwrap_or(s))?;
-        let sig = Signature::from_erc2098(&bytes);
```
afck, line 76: Or would it be possible to do the same thing as [EIP-2](https://eips.ethereum.org/EIPS/eip-2) (point 2), and generate only signatures with the first s-value bit 0?

deuszx: IIUC this scheme was modified again in https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md

deuszx: We're relying on alloy to handle all the possible cases. The problem here was that the input bytes were interpreted incorrectly - i.e. they assumed the format was different, that the parity bit (recovery) was encoded into `s` while it's appended `[s, r, v]` (32bytes, 32bytes, 1byte).

afck (approved):

PR #4090, opened 2025-06-10
Fail earlier if re-proposal round is wrong.

afck:
## Motivation

We never explicitly check that a re-proposal's round is actually greater than the original proposal's. I couldn't find a way to actually exploit it because the checks inside the chain manager seem to implicitly cover all the cases, but it would be better to fail earlier, and it's a very cheap check.

Thanks, @deuszx, for pointing this out!

## Proposal

Add this to `BlockProposal::check_invariants`.

## Test Plan

CI

## Release Plan

## Links

### DIFF
--- a/linera-chain/src/data_types.rs
+++ b/linera-chain/src/data_types.rs
@@ -594,11 +594,14 @@ impl BlockProposal {
         )
     }
 
-    /// Checks that the public key matches the owner and that the optional certificate matches
-    /// the outcome.
+    /// Checks that the original proposal, if present, matches the new one and has a higher round.
     pub fn check_invariants(&self) -> Result<(), &'static str> {
         match (&self.original_proposal, &self.content.outcome) {
-            (None, None) | (Some(OriginalProposal::Fast { .. }), None) => {}
+            (None, None) => {}
+            (Some(OriginalProposal::Fast { .. }), None) => ensure!(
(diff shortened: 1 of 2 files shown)


linera-chain/src/data_types.rs
```diff
@@ -594,18 +594,25 @@ impl BlockProposal {
...
 
-    /// Checks that the public key matches the owner and that the optional certificate matches
-    /// the outcome.
+    /// Checks that the original proposal, if present, matches the new one and has a higher round.
     pub fn check_invariants(&self) -> Result<(), &'static str> {
         match (&self.original_proposal, &self.content.outcome) {
-            (None, None) | (Some(OriginalProposal::Fast { .. }), None) => {}
+            (None, None) => {}
+            (Some(OriginalProposal::Fast { .. }), None) => ensure!(
+                self.content.round > Round::Fast,
```
deuszx, line 602: This relies on the derived `PartialEq` and `Eq` for `Round` enum, right? That's probably fine but it's easy to miss when refactoring. Can you add a short unit test to `Round` where we encode the expected equalities?

afck: I'll add one for `Ord`; I don't see where it uses `Eq` here? Also, derived `Eq` is pretty trivial and doesn't change if e.g. we change the order of the variants.

PR #4102, opened 2025-06-13
Provide access to consumed resources in tests

deuszx:
## Motivation

Messages may be sent with resource grants so that the message receiver doesn't have to pay any fees to receive the message. However, figuring out how much resources to include in the grant is not trivial.

## Proposal

Add support for writing integration tests asserting whether a block consumed less resources than those that would be provided by a grant.

## Test Plan

TODO: this needs a test.

## Release Plan

- These changes follow the usual release cycle, because this adds a new breaking change to the SDK.

## Links


deuszx:
Superseeded by https://github.com/linera-io/linera-protocol/pull/4179

PR #4105, opened 2025-06-14
TypeScript signer is a library/SDK 

deuszx:
## Motivation

When building `linera-web` examples it sometimes happens that the `package-lock.json` of `/signer` is incompatible/out of date. Both signer and web client are used as SDK/libraries so should't have their lock files committed.

`linera-web/build.bash` builds the client for the web but it didn't prepare the TypeScript signers which required downstream client to manually trigger that.

## Proposal

Signer and linera web client are a library/SDK and as such should not have `package-lock.json` committed. This PR removes that file. 

Add a line to `build.bash` that prepare signers (build javascript and typescript files).

## Test Plan

Manually in `linera-web`.

## Release Plan

## Links


linera-web/pnpm-workspace.yaml
```diff
@@ -1,3 +1,2 @@
 packages:
-  - "signer"
```
Twey, line 2: This one confuses me.  So `signer` is no longer in the workspace, i.e. it gets its own separate `node_modules` and `package-lock.json`?

linera-web/build.bash
```diff
@@ -32,3 +32,5 @@ wasm-bindgen \
     --typescript \
     --target web \
     --split-linked-modules
+
+cd signer && pnpm prepare
```
Twey, line 36: I think this should be `build` — as I understand the lifecycle `prepare` is called by the installer after downloading the package.

linera-web/pnpm-workspace.yaml
deuszx: OK, now signer is back in the workspace but its lock files are not tracked.

Similarly removed the lock file for `linera-web` and added them to .gitignore.

linera-web/signer/src/private-key.ts
```diff
@@ -8,7 +8,7 @@ import { Signer } from "@linera/client";
  * It stores the private key directly in memory, which makes it unsuitable for
  * production environments due to security risks.
  * 
- * The `PrivateKeySigner` uses an in-memory `ethers.Wallet` to sign messages following
+ * The `PrivateKey` uses an in-memory `ethers.Wallet` to sign messages following
```
Twey: Nit:

```suggestion
 * The `PrivateKey` signer uses an in-memory `ethers.Wallet` to sign messages following
```

.gitignore
```diff
@@ -26,3 +26,11 @@
...
 node_modules
 **/dist/
+
+# Web client and signer are supposed to be used
+# as a library, so we don't want to
+# include their lock files.
+linera-web/package-lock.json
+linera-web/pnpm-lock.json
+linera-web/signer/package-lock.json
+linera-web/signer/pnpm-lock.json
```
Twey: Nit:

```suggestion
linera-web/**/*-lock.json
```

seems sensible

Twey (approved):

PR #4110, opened 2025-06-15
Fix typos in Grafana dashboard and update test function in linera_net_tests.rs

zeevick10:
Description:  
This pull request addresses the following changes:
- Corrects the typo "avarage" to "average" in the description fields of the Grafana dashboard JSON for Scylla Alternator.
- Updates the test in linera_net_tests.rs by renaming the method call from set_preferred_owner to set_preferred_owner (fixing the function name to match the correct one).

### DIFF
--- a/linera-service/tests/linera_net_tests.rs
+++ b/linera-service/tests/linera_net_tests.rs
@@ -3540,7 +3540,7 @@ async fn test_end_to_end_change_ownership(config: impl LineraNetConfig) -> Resul
 
     // Make owner2 the only (super) owner.
     client.change_ownership(chain, vec![owner2], vec![]).await?;
-    client.set_preffered_owner(chain, Some(owner2)).await?;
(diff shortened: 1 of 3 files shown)


linera-service/tests/linera_net_tests.rs
```diff
@@ -3540,7 +3540,7 @@ async fn test_end_to_end_change_ownership(config: impl LineraNetConfig) -> Resul
 
     // Make owner2 the only (super) owner.
     client.change_ownership(chain, vec![owner2], vec![]).await?;
-    client.set_preffered_owner(chain, Some(owner2)).await?;
+    client.set_preferred_owner(chain, Some(owner2)).await?;
```
afck, line 3543: Good catch!
But it's also in `wallet.rs`, and needs to be fixed in both places, otherwise it doesn't compile.

afck (approved):

PR #4111, opened 2025-06-15
Remove `linera-server initialize` + simplify CLI options

ma2bd:
## Motivation

* `linera-server initialize` is redundant with `linera storage initialize`
* CLI arguments are duplicated in many places.
* `StorageConfigNamespace` is not a great name for the main storage configuration string.

## Proposal

* Remove `linera-server initialize`
* Create `CommonStorageOptions` and used `#[Clap::flatten]` to share code between the different binaries
* Rename `StorageConfig` into `InnerStorageConfig`
* Rename `StorageConfigNamespace` into `StorageConfig`

`InnerStorageConfig` is still visible and used directly in a few places for now, but we'd like `StorageConfig` to be the real public. 

## Test Plan

CI

## Release Plan


linera-service/src/cli/main.rs
```diff
@@ -1222,33 +1335,17 @@ impl Runnable for Job {
     about = "A Byzantine-fault tolerant sidechain with low-latency finality and high throughput",
 )]
 struct ClientOptions {
+    /// Common options.
+    #[command(flatten)]
+    context: ClientContextOptions,
```
afck: Should we name it `context_options`? (Or call the other field `common_storage`?)

afck (approved):

linera-service/src/cli/main.rs
ma2bd: fair

ndr-ds (approved):

PR #4112, opened 2025-06-15
Move basic EVM test from end-to-end test to `linera-execution`.

MathieuDutSik:
## Motivation

Some of the EVM tests are in `linera-service` as end-to-end test when they can be usefully moved to `linera-execution`.

## Proposal

This corresponds to the `test_evm_end_to_end_basic_check`.

All the other remaining tests in `linera-service` for REVM require the use of end-to-end tests and cannot be moved to `linera-execution`.

## Test Plan

The CI.

## Release Plan

## Links

None.


ma2bd (approved):
Nice!

PR #4115, opened 2025-06-15
Remove CommonStoreInternalConfig and CommonStoreConfig

ma2bd:
## Motivation

Storage backends should only receive configuration options that they actually use.

## Proposal

* Remove `CommonStoreInternalConfig` and `CommonStoreConfig` step by step
* Add missing CLI options to  `linera-storage-service`

Nothing changes from the user point of view.

This temporary creates a little more code duplication between `linera-service` and the binaries in `linera-indexer` and `linera-storage-service` but I think that's fine and should be fixed over time. 

Future work:
* We may re-evaluate whether `max_stream_queries` is best implemented as a method of `ReadableKeyValueStore`.
* I'm still conflicted about `max_concurrent_queries` and would very much like to remove it.
* It's not clear what happens if we pass the wrong replication factor for ScyllaDb.
* We may re-evaluate whether `linera-indexer` and `linera-storage-service` belong to separate crates.

Follows #4111

## Test Plan

CI

### DIFF
--- a/linera-indexer/lib/src/rocks_db.rs
+++ b/linera-indexer/lib/src/rocks_db.rs
@@ -6,8 +6,11 @@ use std::path::PathBuf;
 use clap::Parser as _;
 use linera_views::{
     lru_caching::StorageCacheConfig,
-    rocks_db::{PathWithGuard, RocksDbSpawnMode, RocksDbStore, RocksDbStoreConfig},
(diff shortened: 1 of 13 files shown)


ndr-ds (approved):
Even with the code duplication, I think this is much cleaner :)

PR #4121, opened 2025-06-16
Move transaction execution to `BlockExecutionTracker`

deuszx:
## Motivation

As part of the bigger work we want to make sure that the execution of incoming messages can be paid for by the grant they carry. This is difficult to test now.

## Proposal

Extract transaction execution to `BlockExecutionTracker` to make testing simpler.

## Test Plan

CI. This doesn't introduce new features - just a refactor.

## Release Plan

## Links

### DIFF
--- a/linera-chain/src/block_tracker.rs
+++ b/linera-chain/src/block_tracker.rs
@@ -4,19 +4,27 @@
 use std::collections::{BTreeMap, BTreeSet};
 
 use custom_debug_derive::Debug;
+#[cfg(with_metrics)]
+use linera_base::prometheus_util::MeasureLatency;
 use linera_base::{
-    data_types::{Blob, Event, OracleResponse, Timestamp},
-    identifiers::{AccountOwner, BlobId, ChainId},
+    data_types::{Amount, Blob, BlockHeight, Event, OracleResponse, Timestamp},
+    ensure,
+    identifiers::{AccountOwner, BlobId, ChainId, MessageId},
 };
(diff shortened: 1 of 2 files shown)


linera-chain/src/block_tracker.rs
```diff
@@ -85,8 +101,79 @@ impl<'resources, 'blobs> BlockExecutionTracker<'resources, 'blobs> {
         })
     }
 
+    /// Executes a transaction in the context of the block.
+    pub async fn execute_transaction<C>(
+        &mut self,
+        transaction: Transaction<'_>,
+        round: Option<u32>,
```
afck, line 108: Maybe the round should also be a field in the `BlockExecutionTracker`. It's the same for every transaction.

(And possibly even the `&mut ExecutionStateView<C>`? Not sure.)

deuszx: Yes, I think it _could_ be moved there. I thought that MAYBE the same instance of `BlockExecutionTracker` can be applied on top of other things that change:
- round
- chain view

I.e. when the block is reproposed in a different round, on a different chain state, its transactions or replayed oracle responses don't change?

afck: If it's re-proposed on a different chain state, then the block height must also have been changed.

A different round—maybe. But I think we create the tracker once per block execution and then drop it. And the round can't change during execution.

(Anyway, not a blocker! Can be added later.)

PR #4125, opened 2025-06-17
Implement the accounting of runtime fees.

### LINKED ISSUE #2905: Charge for system API calls

MathieuDutSik:
## Motivation

The access to runtime entries like `ChainId` is not expensive, but it has to be accounted for.

Fixes #2905 

## Proposal

The costs imposed on the users have to be proportional to what it costs the validators. Also,
they have to be easy to understand by the developer.

The following design is then done:
* The user is charged a cheap cost proportional to the number of bytes that they obtain.
* For `ChainId`, `ApplicationId`, and `Timestamp`, that is easy. However, for a structure like the `ChainOwnership`, this is more complicated since the size depends on the implementation. So, instead, we compute a size that is easy to understand by the developer.
* For the parameters, this is easy to do.

Two clone operations are removed in other parts of the code.

## Test Plan

A test has been added for the size of constant-size types.
The CI should handle other costs.

## Release Plan

## Links

None

### DIFF
--- a/linera-execution/src/resources.rs
+++ b/linera-execution/src/resources.rs
@@ -68,7 +68,9 @@ pub const RUNTIME_TIMESTAMP_SIZE: u32 = 8;
 pub const RUNTIME_OWNER_WEIGHT_SIZE: u32 = 8;
 
 /// The runtime constant part size of the `ChainOwnership`.
-/// It does not correspond exactly to the memory size of the `ChainOwnership`.
+/// It consists of one `u32` and four `TimeDelta` which are the constant part of
+/// the `ChainOwnership`. The way we do it is not optimal:
+/// TODO(#4164): Implement a procedure for computing naive sizes.
 pub const RUNTIME_CONSTANT_CHAIN_OWNERSHIP_SIZE: u32 = 4 + 4 * 8;
 
 #[cfg(test)]


linera-execution/tests/fee_consumption.rs
```diff
@@ -307,8 +311,10 @@ async fn test_fee_consumption(
         None => {
             let (expected_chain_balance, expected_owner_balance) = if chain_balance >= consumed_fees
             {
+                println!("Case A");
```
ma2bd: Please remove the `println!`

linera-base/src/identifiers.rs
```diff
@@ -50,6 +50,15 @@ impl AccountOwner {
     pub fn is_chain(&self) -> bool {
         self == &AccountOwner::CHAIN
     }
+
+    /// The size of the `AccountOwner`.
+    pub fn size(&self) -> u32 {
+        match self {
+            AccountOwner::Reserved(_) => 1,
+            AccountOwner::Address32(_) => 32,
+            AccountOwner::Address20(_) => 20,
```
ma2bd, line 59: technically + 1 for the tag

linera-execution/src/execution_state_actor.rs
```diff
@@ -74,10 +74,7 @@ where
         let _latency = metrics::LOAD_CONTRACT_LATENCY.measure_latency();
         let blob_id = id.description_blob_id();
         let description = match txn_tracker.created_blobs().get(&blob_id) {
-            Some(description) => {
-                let blob = description.clone();
```
ma2bd, line 78: So this was a useless `Clone` ?

linera-execution/src/policy.rs
```diff
@@ -33,6 +33,8 @@ pub struct ResourceControlPolicy {
     pub read_operation: Amount,
     /// The price of one write operation.
     pub write_operation: Amount,
+    /// The price of one byte from runtime operation.
```
ma2bd: Maybe better?
```
The price of accessing one byte from the runtime.
```

linera-base/src/identifiers.rs
ma2bd: maybe we should just count `32` to simplify, though

ma2bd (commented):
This looks reasonable to me overall. Curious what others think in the team

linera-execution/tests/fee_consumption.rs
MathieuDutSik: Yes of course. So sorry..

linera-base/src/identifiers.rs
MathieuDutSik: Using 32 was indeed my guess.
I thought that putting 32 was an error since we may have addresses larger than 32 with this post-quantum stuff.

About the tag, yes, it is there. But we want something easy to understand by programmers.

linera-execution/src/execution_state_actor.rs
MathieuDutSik: Yes.

linera-execution/src/policy.rs
MathieuDutSik: Thank you.

linera-service/src/cli/command.rs
```diff
@@ -330,6 +330,10 @@ pub enum ClientCommand {
         #[arg(long)]
         write_operation: Option<Amount>,
 
+        /// Set the price per byte obtained from runtime.
```
afck:
```suggestion
        /// Set the price per byte read from runtime.
```

(Or "obtained" below, but I think they should be the same.)

linera-execution/src/resources.rs
```diff
@@ -268,6 +317,91 @@ where
...
+    pub(crate) fn track_runtime_chain_ownership(
+        &mut self,
+        chain_ownership: &ChainOwnership,
+    ) -> Result<(), ExecutionError> {
+        let mut size = 0;
+        for account_owner in &chain_ownership.super_owners {
+            size += account_owner.size();
+        }
+        for account_owner in chain_ownership.owners.keys() {
+            size += account_owner.size() + RUNTIME_BLOCK_HEIGHT_SIZE;
```
afck: Maybe `RUNTIME_OWNER_WEIGHT_SIZE` should be a separate constant, even if it's the same value.

linera-execution/src/resources.rs
```diff
@@ -48,6 +49,52 @@ impl<Account, Tracker> ResourceController<Account, Tracker> {
...
+
+/// The runtime size of a `ChainId`.
+pub const RUNTIME_CHAIN_ID_SIZE: u32 = 32;
+
+/// The runtime size of a `Timestamp`.
+pub const RUNTIME_TIMESTAMP_SIZE: u32 = 8;
+
+/// The runtime constant part size of the `ChainOwnership`.
+/// It does not correspond exactly to the memory size of the `ChainOwnership`.
+pub const RUNTIME_CONSTANT_CHAIN_OWNERSHIP_SIZE: u32 = 4 + 4 * 8;
```
afck, line 74: Would still be good to have some test for this.

Also, why _doesn't_ it correspond to the stack size of `ChainOwnership`?

MathieuDutSik: Here is the problem.

If we compute the size of an empty `ChainOwnership` by using `std::mem::size_of`, then we get 96. This is far away from `4 + 4 * 8 = 36`.

This discrepancy is due to the design of Rust, the use of 1 byte for encoding 1 boolean, and several other factors. I do not completely understand the discrepancy.

From this, here is what I conclude:
* The discrepancy is impossible to explain simply to basic programmers. It is a complex thing, and there is no straightforward way for it.
* Another problem is that we want this to be constant with respect to Rust versioning. It is not clear to me that the size of 96 will remain in the future.
* Basic programmers need simple stuff. So, we need to simplify.
* For example, having an `Option<T>` be of the same size as `T` because the difference is a boolean, we are counting the number of bytes, and a boolean is one eighth of a byte. An empty BTreeMap should have size 0 and so on.
* Doing the computation with those approximations gets us something that is not very far from the real memory use, and so I think it is the right way to do things.

So that is the set of considerations that has led to this decision.

For testing purposes, the ideal way would probably be to write a `std::mem::naive_size_of`.

linera-service/src/cli/command.rs
MathieuDutSik: Ok.

linera-execution/src/resources.rs
MathieuDutSik: Good catch!
I thought it was some block height.

linera-execution/src/resources.rs
afck: Maybe I'm _too_ basic, but I don't even understand the `4 + 4 * 8`. :sweat_smile: 
Is it "one `u32` and four `TimeDelta`s"?

MathieuDutSik: Yes.

afck: OK. I'm not sure it's that intuitive that e.g. a `bool` or an empty map necessarily counts as size 0. I also don't think that users need to understand how the price is justified by the data structure. I'd rather consider it more important that it matches the actual cost, which is probably closer to `size_of`.

Anyway, not a blocker from my side, but maybe something we should discuss with the team.

deuszx: Maybe we should use a macro (there are a couple of crates that do that) to derive a heap size for all things we want to account for, rather than implement it by hand

MathieuDutSik: Yes, this would be my preferred solution. Which crate would you recommend for that?

deuszx (commented):
[This comment](https://github.com/linera-io/linera-protocol/blob/main/linera-base/src/data_types.rs#L292) has a TODO that we should account for calling system calls - is this PR solving it?

linera-execution/src/resources.rs
deuszx: No, I don't have any preferred one.

MathieuDutSik:
> [This comment](https://github.com/linera-io/linera-protocol/blob/main/linera-base/src/data_types.rs#L292) has a TODO that we should account for calling system calls - is this PR solving it?
> 
> Also, this only charges for the size of things returned but not the time spent retrieving it.

Yes, those are the system calls from a smart contract. Anything else you had in mind?

About the time spent, there are several problems:
* It varies from validator to validator. Not great for the consensus.
* We need something simple to understand by the programmer of smart contracts. Time is not one of them.

deuszx:
> Yes, those are the system calls from a smart contract. Anything else you had in mind?

Nothing in particular - I was asking you since you were working on it.

> About the time spent, there are several problems:
> 
> * It varies from validator to validator. Not great for the consensus.
> * We need something simple to understand by the programmer of smart contracts. Time is not one of them.

Yes, that's why we should have a reference machine on which we run the tests to measure time spent. The results are used to choose the temporal costs.

It doesn't matter it varies from validator-to-validator (b/c they have different machine). They (validators) are suggested to use the reference hardware (or sth close to it). It has no influence on the consensus b/c the costs are deterministic, it's the actual time that varies (unless they use reference hardware) but that's not a problem. The point of gas accounting is to protect from DoS and refund validators for their "machines".

PR #4130, opened 2025-06-17
Leave gaps between sender blocks.

afck:
## Motivation

It's unnecessary to fully download all chains that sent a message to our own chain(s).

## Proposal

Only download the relevant blocks.

## Test Plan

A new client test verifies that we are leaving gaps.

## Release Plan

## Links

- Closes https://github.com/linera-io/linera-protocol/issues/3969.

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -768,8 +768,8 @@ impl<Env: Environment> Client<Env> {
         Ok(())
     }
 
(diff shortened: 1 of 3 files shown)


afck:
Emphasis on "fully": We don't need to _fully_ download the sender chains; so we just download them partially now.

The main part of the change is after `fetch_sent_certificate_hashes`: We now take only the hashes we need instead of all of them.

PR #4131, opened 2025-06-17
Remove committees from the chain description

bart-linera:
## Motivation

`ChainDescription` contains committees that were active at the time of creation of the chain. This unnecessarily duplicates information that is already available in other places, like the committee blobs.

## Proposal

Replace the `committees` field in the `ChainDescription` that maps epochs to committees by just the set of active epochs.

## Test Plan

## Release Plan

## Links


linera-chain/src/unit_tests/chain_tests.rs
```diff
@@ -155,26 +147,27 @@ async fn test_block_size_limit() -> anyhow::Result<()> {
...
-                ValidatorState {
-                    network_address: ValidatorPublicKey::test_key(1).to_string(),
-                    votes: 1,
-                    account_public_key: AccountPublicKey::test_key(1),
-                },
-            )]),
-            ResourceControlPolicy {
-                maximum_block_size,
-                ..ResourceControlPolicy::default()
+    let committee = Committee::new(
```
afck, line 142: This committee is created and stored three times in this file. Let's deduplicate that.

Or should it even be part of `make_child_chain_description_with_config`?

linera-core/src/unit_tests/test_utils.rs
```diff
@@ -172,21 +172,22 @@ where
...
-                .await
-                .state
-                .storage_client()
-                .read_network_description()
-                .await
-                .unwrap()
-                .unwrap()
-                .admin_chain_id,
+            genesis_committee_blob_hash: storage_network_description.genesis_committee_blob_hash,
+            admin_chain_id: storage_network_description.admin_chain_id,
```
afck: Why don't we just return the exact network description we read from storage?
Feels like the tests should pass with that; and if they don't, they should be adjusted.

linera-execution/src/lib.rs
```diff
@@ -1082,10 +1094,30 @@ impl ExecutionRuntimeContext for TestExecutionRuntimeContext {
...
+            .find(|item| item.key().blob_type == BlobType::Committee)
+            .ok_or_else(|| ViewError::NotFound("committee not found".to_owned()))?
+            .value()
+            .bytes()
+            .to_vec();
+        let committee: Committee = bcs::from_bytes(&committee_blob_bytes)?;
+        Ok(epochs
+            .into_iter()
+            .map(|epoch| (epoch, committee.clone()))
+            .collect())
```
afck, line 1121: That may be fine for all the tests we currently have, but it's really just assigning the first committee blob it found to all epochs. Let's at least add a `TODO` and create an issue to fix this.

linera-execution/tests/test_execution.rs
```diff
@@ -1159,7 +1159,9 @@ async fn test_open_chain() -> anyhow::Result<()> {
         ValidatorPublicKey::test_key(0),
         AccountPublicKey::test_key(0),
     )]);
-    let committees = BTreeMap::from([(Epoch::ZERO, bcs::to_bytes(&committee)?)]);
+    let committee_blob = Blob::new_committee(
+        bcs::to_bytes(&committee).expect("serializing a committee should succeed"),
```
afck, line 1163: (BTW in tests it's also fine to just `unwrap()`.)

linera-storage/src/lib.rs
```diff
@@ -173,6 +177,54 @@ pub trait Storage: Sized {
...
+    async fn committees_for(
+        &self,
+        mut epochs: BTreeSet<Epoch>,
+    ) -> Result<BTreeMap<Epoch, Committee>, ViewError> {
+        let read_committee = async |committee_hash| -> Result<Committee, ViewError> {
+            let blob_id = BlobId::new(committee_hash, BlobType::Committee);
+            let committee_blob = self
+                .read_blob(blob_id)
+                .await?
+                .ok_or_else(|| ViewError::NotFound(format!("Blob not found: {}", blob_id)))?;
```
afck: `NotFound` already prints `entry does not exist in memory`, so this could just say:

```suggestion
                .ok_or_else(|| ViewError::NotFound(format!("blob {}", blob_id)))?;
```

Actually, why "in memory"? Should probably be "in storage"?

linera-storage/src/lib.rs
```diff
@@ -173,6 +177,54 @@ pub trait Storage: Sized {
...
+            Ok(bcs::from_bytes(committee_blob.bytes())?)
+        };
+
+        let network_description = self
+            .read_network_description()
+            .await?
+            .ok_or_else(|| ViewError::NotFound("NetworkDescription not found".to_owned()))?;
+        let admin_chain_id = network_description.admin_chain_id;
+        let epoch_creation_events = self
+            .read_events_from_index(&admin_chain_id, &StreamId::system(EPOCH_STREAM_NAME), 0)
```
afck: Maybe the starting index should only be the minimum of the epochs we're interested in.

And we should also allow specifying an _upper_ bound for the event index at some point. (Not urgent, though; in practice it'll usually be the most recent events that are interesting.)

linera-storage/src/lib.rs
```diff
@@ -173,6 +177,54 @@ pub trait Storage: Sized {
...
+        }
+        for index_and_event in epoch_creation_events {
+            let epoch = Epoch::from(index_and_event.index);
+            if !epochs.remove(&epoch) {
+                continue;
+            }
+            let blob_hash = bcs::from_bytes::<CryptoHash>(&index_and_event.event)?;
+            let committee = read_committee(blob_hash).await?;
+            result.insert(epoch, committee);
+        }
```
afck: Maybe these should be joined and run concurrently?

Cargo.toml
```diff
@@ -53,7 +53,6 @@ default-members = [
     "linera-summary",
     "linera-views",
     "linera-views-derive",
-    # "linera-web", # not this one
```
ma2bd: This might have been intended by @Twey

linera-core/src/unit_tests/test_utils.rs
bart-linera: TBH, I have no idea :sweat_smile: This might be an artifact of some earlier form of the tests. I'll try returning just the one from storage and we'll see how it goes :+1:

linera-execution/tests/test_execution.rs
bart-linera: I tend to prefer `expect` to have some kind of a message associated with a potential failure... but then again, tests print the line where the failing unwrap was, don't they? So maybe it is overkill after all.

afck: Yes, they do. Anyway, your call!

linera-core/src/unit_tests/test_utils.rs
bart-linera: Confirmed: tests _are_ passing with just the storage network description :tada:

linera-base/src/data_types.rs
```diff
@@ -834,8 +834,8 @@ pub struct InitialChainConfig {
     pub ownership: ChainOwnership,
     /// The epoch in which the chain is created.
     pub epoch: Epoch,
-    /// Serialized committees corresponding to epochs.
-    pub committees: BTreeMap<Epoch, Vec<u8>>,
+    /// Set of epochs active at the time of creation of the chain.
+    pub active_epochs: BTreeSet<Epoch>,
```
deuszx, line 838: To make this even smaller - we include the lowest & highest active epoch.

PR #4134, opened 2025-06-17
Implement `msg.sender` in Linera support of the EVM.

### LINKED ISSUE #4129: Implement `msg.sender` in the EVM support of Linera.
The `caller` entry in `TxEnv` is set to `Address::ZERO` in the current implementation of the EVM in Linera.
This severely limits what can be implemented. For example, we cannot implement ERC20 without it.

MathieuDutSik:
## Motivation

When calling the EVM smart contract we currently set up the `caller` entry to `Address::ZERO`.
This is quite inadequate as the `msg.sender` construct is widely used in solidity code, for example
with ERC20.

Fixes #4129 

## Proposal

The `msg.sender` is first the Externally Owned Account when calling a contract. When the contract references another contract, the address becomes that of the EVM contract.

The following is done:
* First of all, we need to set the user's address to an `Address20` entry. This is done in a somewhat hackish way by modifying the `generate_from` in `crypto/mod.rs` according to the `revm` feature.
* For `process_streams` there are no `authenticated_*` set so we set it to `Address::ZERO`.
* Then the `caller` entry is computed from the context in question. That is the `authenticated_*` entries. For the application id, we require it to be of one of an EVM address. For a user address, we require it to be an `Address20`.
* The function `get_msg_address` first tries to access the application_id, and then if nothing is available, the `authenticated_caller_if`. This is to implement the `msg.sender` semantic.
* A wasm contract calling an EVM contract has the address set to `Address::ZERO`.
* A user in `Address32` calling an EVM smart contract will have `Address::ZERO`.
* For service calls, the `caller` is set to `SERVICE_ADDRESS`, a constant address that is distinct from `Address::ZERO`.
* The database is reorganized in order to reflect the fact that now we have multiple users and multiple nonces to store.

Other notes:
* The next step is to handle the balances so that effective transfers can be done.
* The solidity smart contracts used for tests are returning a uint64. This is artificial, as it seems that there is a bug in REVM.

## Test Plan

A test in CI has been added.

## Release Plan

## Links

None.

### DIFF
--- a/linera-execution/src/evm/revm.rs
+++ b/linera-execution/src/evm/revm.rs
@@ -465,10 +465,12 @@ type Ctx<'a, Runtime> = MainnetContext<WrapDatabaseRef<&'a mut DatabaseRuntime<R
 // functionalities accessed from the EVM.
 const PRECOMPILE_ADDRESS: Address = address!("000000000000000000000000000000000000000b");
 
-// This is the zero address used for service calls
+// This is the zero address used when no address can be obtained from `authenticated_signer`
+// and `authenticated_caller_id`. This scenario does not occur if an Address20 user calls or
+// if an EVM contract calls another EVM contract.
 const ZERO_ADDRESS: Address = address!("0000000000000000000000000000000000000000");
 
-// This is the address being used for service calls
+// This is the address being used for service calls.
 const SERVICE_ADDRESS: Address = address!("0000000000000000000000000000000000002000");
 
 fn address_to_user_application_id(address: Address) -> ApplicationId {
@@ -1111,7 +1113,7 @@ where
         ensure_message_length(operation.len(), 4)?;
         let caller = self.get_msg_address()?;
         let (gas_final, output, logs) = if &operation[..4] == INTERPRETER_RESULT_SELECTOR {
-            ensure_message_length(operation.len(), 28)?;
+            ensure_message_length(operation.len(), 8)?;
             forbid_execute_operation_origin(&operation[4..8])?;
             let result = self.init_transact_commit(operation[4..].to_vec(), caller)?;
             result.interpreter_result_and_logs()?
@@ -1247,6 +1249,18 @@ where
         self.write_logs(result.logs, "deploy")
     }
 
+    /// Computes the address used in the `msg.sender` variable.
+    /// It is computed in the following way:
+    /// * If a Wasm contract calls an EVM contract then it is `Address::ZERO`.
+    /// * If an EVM contract calls an EVM contract it is the address of the contract.
+    /// * If a user having an `AccountOwner::Address32` address calls an EVM contract
+    ///   then it is `Address::ZERO`.
+    /// * If a user having an `AccountOwner::Address20` address calls an EVM contract
+    ///   then it is this address.
+    ///
+    /// By doing this we ensure that EVM smart contracts works in the same way as
+    /// on the EVM and that users and contracts outside of that realm can still
+    /// call EVM smart contracts.
     fn get_msg_address(&self) -> Result<Address, ExecutionError> {
         let mut runtime = self.db.runtime.lock().expect("The lock should be possible");
         let application_id = runtime.authenticated_caller_id()?;


linera-execution/src/evm/database.rs
```diff
@@ -115,21 +107,17 @@ fn application_id_to_address(application_id: ApplicationId) -> Address {
 impl<Runtime: BaseRuntime> DatabaseRuntime<Runtime> {
     /// Encode the `index` of the EVM storage associated to the smart contract
     /// in a linera key.
-    fn get_linera_key(val: u8, index: U256) -> Result<Vec<u8>, ExecutionError> {
-        let mut key = vec![val, KeyCategory::Storage as u8];
+    fn get_linera_key(key_prefix: &[u8], index: U256) -> Result<Vec<u8>, ExecutionError> {
+        let mut key = key_prefix.to_vec();
         bcs::serialize_into(&mut key, &index)?;
         Ok(key)
```
deuszx, line 113: How can you be sure this is safe? `serialize_into` will write into `mut key` and potentially overwrite content with `&index` value.

linera-execution/src/evm/revm.rs
```diff
@@ -1215,32 +1221,49 @@ where
...
+        let result = self.transact_commit(EvmTxKind::Create, vec_init, caller)?;
         result
             .check_contract_initialization(self.db.contract_address)
             .map_err(EvmExecutionError::IncorrectContractCreation)?;
         self.write_logs(result.logs, "deploy")
     }
 
+    fn get_msg_address(&self) -> Result<Address, ExecutionError> {
+        let mut runtime = self.db.runtime.lock().expect("The lock should be possible");
+        let application_id = runtime.authenticated_caller_id()?;
```
deuszx, line 1266: This is the last application on the callstack - it doesn't include the _current_ application, does it?

linera-execution/src/evm/revm.rs
```diff
@@ -1215,32 +1221,49 @@ where
...
+        if let Some(application_id) = application_id {
+            if application_id.is_evm() {
+                return Ok(application_id.evm_address());
+            }
+        };
+        let account_owner = runtime.authenticated_signer()?;
```
deuszx, line 1274: Shouldn't we check if the callstack is of length <=1 and if so return the `authenticated_signer`? i.e. return the signer of the block if it's the first contract call in the stack

linera-execution/src/evm/revm.rs
```diff
@@ -1401,14 +1425,14 @@ where
...
-
         let block_env = self.db.get_service_block_env()?;
         let inspector = CallInterceptorService {
             db: self.db.clone(),
             contract_address: self.db.contract_address,
             precompile_addresses: precompile_addresses(),
         };
-        let nonce = self.db.get_nonce(&ZERO_ADDRESS)?;
+        let caller = ZERO_ADDRESS;
+        let nonce = self.db.get_nonce(&caller)?;
```
deuszx, line 1454: what's the point of getting a nonce for "zero address"? There should be none. i.e. zero address cannot be used as an author of anything.

linera-execution/src/evm/database.rs
MathieuDutSik: The function `serialize_into` will append the content of the serialization into `key`. So, I am not sure where any overwriting will occur.

deuszx: It depends on the implementation of `std::io::Write` for this type
```rust
/// Same as `to_bytes` but write directly into an `std::io::Write` object.
pub fn serialize_into<W, T>(write: &mut W, value: &T) -> Result<()>
where
    W: ?Sized + std::io::Write,
    T: ?Sized + Serialize,
{
    let serializer = Serializer::new(write, crate::MAX_CONTAINER_DEPTH);
    value.serialize(serializer)
}
```

ma2bd:
> This is artificial, as it seems that there is a bug in REVM.

Wait wat? was this reported?

linera-base/src/crypto/mod.rs
```diff
@@ -195,11 +195,17 @@ impl AccountSecretKey {
...
-    /// Generates a new key pair from the given RNG. Use with care.
+    #[cfg(all(with_getrandom, not(feature = "revm")))]
+    /// Generates a new Ed25519 key pair from the given RNG. Use with care.
     pub fn generate_from<R: CryptoRng>(rng: &mut R) -> Self {
         AccountSecretKey::Ed25519(Ed25519SecretKey::generate_from(rng))
     }
+
+    #[cfg(all(with_getrandom, feature = "revm"))]
+    /// Generates a new Evm Secp256k1 key pair from the given RNG. Use with care.
+    pub fn generate_from<R: CryptoRng>(rng: &mut R) -> Self {
```
ma2bd, line 206: Would it possible to use two different functions instead of introducing feature-gated code branches?

linera-base/src/identifiers.rs
```diff
@@ -937,6 +937,18 @@ impl<A> ApplicationId<A> {
     pub fn bytes32(&self) -> B256 {
         *self.application_description_hash.as_bytes()
     }
+
+    /// Returns whether the `ApplicationId` is the one of an EVM application.
+    pub fn is_evm(&self) -> bool {
```
ma2bd, line 942: I'm not completely fond of this. Don't we have the application description at hand?

linera-execution/src/evm/database.rs
ma2bd: I believe the proposed code is correct (for this implementation of `Write` and likely in general)

linera-execution/src/evm/database.rs
```diff
@@ -115,21 +107,17 @@ fn application_id_to_address(application_id: ApplicationId) -> Address {
...
     /// Returns the tag associated to the contract.
-    fn get_contract_address_key(&self, address: &Address) -> Option<u8> {
-        if address == &Address::ZERO {
-            return Some(KeyTag::NullAddress as u8);
-        }
-        if address == &self.contract_address {
-            return Some(KeyTag::ContractAddress as u8);
-        }
-        None
+    fn get_address_key(&self, prefix: u8, address: &Address) -> Vec<u8> {
```
ma2bd: nit: taking the address by value is more precise here.

linera-execution/src/evm/database.rs
```diff
@@ -264,74 +245,55 @@ where
...
-                let key_prefix = vec![val, KeyCategory::Storage as u8];
-                let key_info = vec![val, KeyCategory::AccountInfo as u8];
-                let key_state = vec![val, KeyCategory::AccountState as u8];
-                if account.is_selfdestructed() {
-                    batch.delete_key_prefix(key_prefix);
-                    batch.put_key_value(key_info, &AccountInfo::default())?;
-                    batch.put_key_value(key_state, &AccountState::NotExisting)?;
+            let key_prefix = self.get_address_key(KeyCategory::Storage as u8, address);
+            let key_info = self.get_address_key(KeyCategory::AccountInfo as u8, address);
+            let key_state = self.get_address_key(KeyCategory::AccountState as u8, address);
```
ma2bd: Would it make sense to eventually use a collections of `View`s here to avoid the boilerplate?

linera-execution/src/evm/database.rs
deuszx: I am not sure this is typical though. The first argument is a target for writing a serialized value. I will try to find an example tomorrow.

ma2bd: Intuitively:
* The signature of this method from `bcs` could and should probably be `serialize_into<W, T>(write: W, ...)` then it would be clear that we can only append to the `write` object.
* The blanket implementation of `Write` for `&mut W` is most certainly just forwarding the method calls.

deuszx: No, you were right - seems like all the implementations keep track of the first free position to write at

linera-execution/src/evm/database.rs
```diff
@@ -264,74 +245,55 @@ where
...
+            if account.is_selfdestructed() {
+                batch.delete_key_prefix(key_prefix);
+                batch.put_key_value(key_info, &AccountInfo::default())?;
+                batch.put_key_value(key_state, &AccountState::NotExisting)?;
```
deuszx, line 258: Do we need to write these "zero values" after deleting the account? It'd be good if we didn't have to store anything for non-existent entries.

linera-execution/src/evm/revm.rs
MathieuDutSik: No. The current application is not there. And if you want to get it, you have it in `BaseRuntime`.

linera-execution/src/evm/revm.rs
MathieuDutSik: We do not have access to the call stack, so I am unsure what you mean.

linera-execution/src/evm/database.rs
```diff
@@ -362,15 +324,12 @@ where
...
-        let mut keys = Vec::new();
-        for key_tag in [KeyTag::NullAddress, KeyTag::ContractAddress] {
-            let key = vec![key_tag as u8, KeyCategory::AccountInfo as u8];
-            keys.push(key);
-        }
         let mut runtime = self.runtime.lock().expect("The lock should be possible");
-        let promise = runtime.contains_keys_new(keys)?;
-        let result = runtime.contains_keys_wait(&promise)?;
-        Ok(result[0] && result[1])
+        let evm_address = runtime.application_id()?.evm_address();
```
deuszx, line 328: nit: the `evm_address()` forces any application address into the EVM address which seems wrong. We should never try to make a Linera address into EVM address.

Fortunately nothing wrong can happen _here_ since such application probably won't exist in the storage and we return `false`.

linera-service/tests/linera_net_tests.rs
```diff
@@ -1153,6 +1153,104 @@ async fn test_evm_process_streams_end_to_end_counters(config: impl LineraNetConf
...
+    let mutation = mutation.abi_encode();
+    let mutation = EvmQuery::Mutation(mutation);
+    application_inner.run_json_query(mutation).await?;
+
+    let mutation = remote_checkCall {
+        remote_address: evm_contract_inner,
+    };
+    let mutation = mutation.abi_encode();
+    let mutation = EvmQuery::Mutation(mutation);
+    application_outer.run_json_query(mutation).await?;
```
deuszx, line 1255: This should return 49, right? If so - can we verify that here?

deuszx (changes requested):
I think we need the tests for the following cases:
- when there's only 1 contract on the stack (i.e. we're calling the contract in the operation) - the `msg.sender` is `authenticated_signer` (i.e. that it's the block proposer)
- if we're calling from Linera to Solidity, the `msg.sender` is `Address::ZERO` (although maybe can be done in a separate PR but please ticket).

I'm also not entirely sure that the choice of `Address::ZERO` for the default address is safe. It has gotten a lot of different meanings in the EVM/Solidity domain and many contracts have special handling for it.

MathieuDutSik:
> > This is artificial, as it seems that there is a bug in REVM.
> 
> Wait wat? was this reported?

I could not reproduce it separately when running in the way that REVM is supposed to be used.

linera-execution/src/evm/revm.rs
MathieuDutSik: It is a fact that for service calls, there is no `authenticated_signer` or `authenticated_caller_id`.
Yet, when calling an EVM smart contract, we need to have an address.

What I proposed instead is to replace `ZERO_ADDRESS` with a `SERVICE_ADDRESS`. It is fixed but different from `ZERO_ADDRESS`. It is a little bit clearer.

And yes, we need nonces. Because those contract calls are called several times.

linera-base/src/crypto/mod.rs
MathieuDutSik: Maybe.
But this would be a bigger PR and force to change the API.
This can be done but I think this should be done in another PR.

linera-execution/src/evm/revm.rs
deuszx: I think this is possibly a way into bunch of bugs in the future (i.e. using zero address and it having nonces which basically makes it look like it's an actual account that can post transactions).

linera-base/src/identifiers.rs
MathieuDutSik: I am not sure how to access it.
The context in which we are using this is when building the address.
What we have is the `authenticated_signer` and `authenticated_called_id`. Those give you the `ApplicationId` but not the application description.
Maybe we could access the application description by accessing the storage, but wouldn't that be slower?

linera-execution/src/evm/revm.rs
deuszx: We (literally _here_) don't but runtime has.

linera-execution/src/evm/database.rs
MathieuDutSik: Yes, we have to delete the entries after deleting the account.
Zeroing values has a precise meaning in solidity: It means the values are absent.

deuszx: I understand we have to delete the entry - I was wondering if we also have to put `AccountInfo::default` and `::NotExisting` into storage - seems like a waste of CPU and storage.

linera-execution/src/evm/database.rs
MathieuDutSik: Nothing can happen here because the application in question is an EVM application. Therefore, it does have an `ApplicationId` and an `Address`.

linera-execution/src/evm/revm.rs
MathieuDutSik: Service calls are never written into the storage, so what can go wrong?

linera-execution/src/evm/database.rs
MathieuDutSik: I do not think so.
The abstraction of having a database that is updated as the execution goes on is managed by REVM. So, the view will not help in that.
The section of the code in question is in the commit part at the end of the execution, where stuff gets written.

linera-execution/src/evm/revm.rs
deuszx: If service calls into contract - and it can do that - then the contract will see `msg.sender == zero address` and its nonce.

Can't we pass in the "caller address" instead?

MathieuDutSik: Explain me how a service can call a contract.

linera-execution/src/evm/database.rs
MathieuDutSik: I follow the convention of REVM here.

linera-execution/src/evm/revm.rs
deuszx: A service can query app's state, meaning it can call its methods. If those methods have a logic that checks the `msg.sender` then it will be incorrect.

MathieuDutSik: Give the method in question.
It would have to be in `ServiceRuntime`.

deuszx: Sorry, you're right. It can read contract's state but it doesn't have access to its methods. If needs to reimplement them (if it wants to).

linera-execution/src/evm/database.rs
MathieuDutSik: Yes, we can pass `address: Address`.
Done.

linera-service/tests/linera_net_tests.rs
MathieuDutSik: No.
Since the mutation does not return a value. For that you need a json query.

MathieuDutSik:
> I think we need the tests for the following cases:
> 
> * when there's only 1 contract on the stack (i.e. we're calling the contract in the operation) - the `msg.sender` is `authenticated_signer` (i.e. that it's the block proposer)
> * if we're calling from Linera to Solidity, the `msg.sender` is `Address::ZERO` (although maybe can be done in a separate PR but please ticket).
> 
> I'm also not entirely sure that the choice of `Address::ZERO` for the default address is safe. It has gotten a lot of different meanings in the EVM/Solidity domain and many contracts have special handling for it.

The first part is tested in `test_evm_msg_sender`.

The second part is now tested in `test_wasm_call_evm_end_to_end_counter`.

linera-service/tests/linera_net_tests.rs
```diff
@@ -1153,6 +1164,104 @@ async fn test_evm_process_streams_end_to_end_counters(config: impl LineraNetConf
...
+    let application_outer = node_service
+        .make_application(&chain, &application_id_outer)
+        .await?;
+
+    let mutation = check_msg_senderCall {
+        remote_address: owner,
+    };
+    let mutation = mutation.abi_encode();
+    let mutation = EvmQuery::Mutation(mutation);
+    application_inner.run_json_query(mutation).await?;
```
deuszx, line 1248: It's very difficult to see that `mutation` query here will have the client address as `msg.sender`. That's why I missed this 👍

deuszx:
Do you plan on adding tests for when Solidity contracts are called by non-EVM account/contract?

linera-execution/src/evm/revm.rs
```diff
@@ -465,9 +465,12 @@ type Ctx<'a, Runtime> = MainnetContext<WrapDatabaseRef<&'a mut DatabaseRuntime<R
 // functionalities accessed from the EVM.
 const PRECOMPILE_ADDRESS: Address = address!("000000000000000000000000000000000000000b");
 
-// This is the zero address of the contract
+// This is the zero address used for service calls
 const ZERO_ADDRESS: Address = address!("0000000000000000000000000000000000000000");
 
+// This is the address being used for service calls
```
deuszx: The `ZERO_ADDRESS` above has the same comment.

deuszx:
> For service calls, the caller is set to ZERO.

This is no longer true https://github.com/linera-io/linera-protocol/pull/4134/files#diff-fae76550d59d4a6fea19d86196abdbba4e2b1a49ed0dfde92141a6df58090665R1439

linera-execution/src/evm/revm.rs
```diff
@@ -1401,14 +1430,14 @@ where
...
         let data = Bytes::from(input);
-
         let block_env = self.db.get_service_block_env()?;
         let inspector = CallInterceptorService {
             db: self.db.clone(),
             contract_address: self.db.contract_address,
             precompile_addresses: precompile_addresses(),
         };
-        let nonce = self.db.get_nonce(&ZERO_ADDRESS)?;
+        let caller = SERVICE_ADDRESS;
```
deuszx, line 1453: Why can't you require that calls to `*Service::transact` also pass in the caller? Just like the `*Contract` does now.

linera-execution/src/evm/revm.rs
```diff
@@ -1105,15 +1109,16 @@ where
     fn execute_operation(&mut self, operation: Vec<u8>) -> Result<Vec<u8>, ExecutionError> {
         self.db.set_contract_address()?;
         ensure_message_length(operation.len(), 4)?;
+        let caller = self.get_msg_address()?;
         let (gas_final, output, logs) = if &operation[..4] == INTERPRETER_RESULT_SELECTOR {
-            ensure_message_length(operation.len(), 8)?;
+            ensure_message_length(operation.len(), 28)?;
```
deuszx: Are these ensures necessary? Where do they come from/How did you calculate them?

linera-execution/src/evm/revm.rs
```diff
@@ -1215,32 +1224,51 @@ where
...
         vec_init.extend_from_slice(&constructor_argument);
-        let result = self.transact_commit(EvmTxKind::Create, vec_init)?;
+        let result = self.transact_commit(EvmTxKind::Create, vec_init, caller)?;
         result
             .check_contract_initialization(self.db.contract_address)
             .map_err(EvmExecutionError::IncorrectContractCreation)?;
         self.write_logs(result.logs, "deploy")
     }
 
+    fn get_msg_address(&self) -> Result<Address, ExecutionError> {
```
deuszx, line 1264: Please add a comment here explaining that this is a deliberate design choice to return `Address::ZERO` if the caller is either a non-Secp256k1 account or non-EVM application.

examples/call-evm-counter/src/contract.rs
```diff
@@ -38,18 +49,23 @@ impl Contract for CallCounterContract {
...
+            function call_from_wasm(address remote_address);
+        }
+        match operation {
+            CallCounterOperation::Increment(increment) => {
+                let operation = incrementCall { input: increment };
+                let operation = operation.abi_encode();
+                self.process_operation(operation)
+            }
+            CallCounterOperation::TestCallAddress => {
+                let remote_address = address!("0000000000000000000000000000000000000000");
```
deuszx, line 63: I think it's quite non-intuitive that `msg.sender` can be so many different values:
1. if EVM contract/account makes a contract call - we set it to the correct address
2. if we make a service call then it's `0x000...20`
3. if Linera contract/account makes the call - we set it to `0x000..00`.

linera-execution/src/evm/revm.rs
MathieuDutSik: Ok, corrected.

MathieuDutSik:
> > For service calls, the caller is set to ZERO.
> 
> This is no longer true https://github.com/linera-io/linera-protocol/pull/4134/files#diff-fae76550d59d4a6fea19d86196abdbba4e2b1a49ed0dfde92141a6df58090665R1439

Corrected.

linera-execution/src/evm/revm.rs
MathieuDutSik: Because there is no caller associated to services.
The `ContractRuntime` has `authenticated_signer` / `authenticated_caller_id`. The service has none of that.

linera-execution/src/evm/revm.rs
MathieuDutSik: That one was an oversight.
Thank you for finding it.

linera-execution/src/evm/revm.rs
MathieuDutSik: Ok, done.

examples/call-evm-counter/src/contract.rs
MathieuDutSik: Yes, the service is quite separate from the other calls.

linera-execution/src/evm/revm.rs
deuszx: but why it didn't fail? the method returns an error if the length is invalid.

MathieuDutSik: By "length is invalid" in that case, we mean too small input.

It suffices to have a call that is nontrivial for it to have a sufficient length.

deuszx: Does this have to be part of every execution or can it be a somethine like test-only/debug assert?

MathieuDutSik: Checking the length is really fast to do, so yes it makes sense to have it in every operation.

I think it is better to make the cheap tests early.

deuszx: I am not worried about speed - rather that there are some valid operations that don't meet this expectation

PR #4136, opened 2025-06-18
Only listen to faucet chain; not other chains in wallet.

afck:
## Motivation

In the testnet we sometimes use the faucet with a wallet with lots of chains. The chain listener then synchronizes them all, which is wasteful.

## Proposal

Allow specifying a different set of chains for the chain listener, so that it doesn't use all chains in the wallet.

In the faucet, start the chain listener with only the admin chain and the faucet's main chain.

## Test Plan

We should now be able to start the testnet faucet with the original wallet again, without it spending a lot of time downloading blocks.

## Release Plan

## Links

### DIFF
--- a/linera-client/src/chain_listener.rs
+++ b/linera-client/src/chain_listener.rs
@@ -124,8 +124,12 @@ impl<C: ClientContext> ChainListener<C> {
     }
 
     /// Runs the chain listener.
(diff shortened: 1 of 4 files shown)


ndr-ds (approved):
Nice, thank you!

PR #4138, opened 2025-06-18
Don't hold `DashMap` lock over await points.

afck:
## Motivation

The ScyllaDB tests currently hang on main because we hold a `DashMap` lock across an `await` point.

## Proposal

Don't.

## Test Plan

This fixed the ScyllaDB deadlock for me locally.

## Release Plan

## Links


afck:
I created #4139: That'll be a bit more work, but it should prevent these issues in the future.

ndr-ds:
@MathieuDutSik It was being fixed on https://github.com/linera-io/linera-protocol/pull/4047, but we haven't merged that yet

PR #4143, opened 2025-06-18
Eliminate another pattern of Dashmap potential block.

MathieuDutSik:
## Motivation

We should not have a Dashmap access to an `Entry` followed by some awaity stuff and then a release as this introduces contention problem.

## Proposal

An examination of the source code revealed a problem in `linera-storage` that is thus addressed.

## Test Plan

The CI.

## Release Plan

## Links

None.

### DIFF
--- a/linera-storage/src/lib.rs
+++ b/linera-storage/src/lib.rs
@@ -10,7 +10,7 @@ mod db_storage;
 use std::sync::Arc;
 
 use async_trait::async_trait;
-use dashmap::{mapref::entry::Entry, DashMap};
+use dashmap::DashMap;
 use linera_base::{
     crypto::CryptoHash,
     data_types::{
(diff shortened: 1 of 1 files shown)


linera-storage/src/lib.rs
```diff
@@ -358,28 +358,26 @@ where
...
-            Entry::Vacant(entry) => {
-                let contract = self.storage.load_contract(description).await?;
-                entry.insert(contract.clone());
-                Ok(contract)
-            }
+        let application_id = description.into();
+        if let Some(contract) = self.user_contracts.get(&application_id) {
+            return Ok(contract.clone());
         }
+        let contract = self.storage.load_contract(description).await?;
```
ma2bd, line 463: In case of race condition, we would load the contract twice and drop the first instance. Not sure if that's a problem. In any case, we may want to add a comment to help future readers.

cc @afck @MathieuDutSik

MathieuDutSik: Yes, of course.
I can add the relevant text. Should I add the same text for the ScyllaDb backend code where the same pattern occurs?

afck: In both cases, it should be idempotent, right? So I think it wouldn't be a problem.

deuszx:
Can you have a look at [this](https://github.com/linera-io/linera-protocol/blob/main/linera-service/src/exporter/storage.rs#L95) file as well? I didn't notice it when reviewing

MathieuDutSik:
> Can you have a look at [this](https://github.com/linera-io/linera-protocol/blob/main/linera-service/src/exporter/storage.rs#L95) file as well? I didn't notice it when reviewing

I could not determine whether this pattern was fine or not. The function are "Atomic async operations" but I do not know what that means.

Besides, looking at the code, I find the following:
```rust
                let blob = self.storage.read_blob(blob_id).await?.unwrap();
```

This is likely a problem if the blob is actually missing.

PR #4147, opened 2025-06-19
Detect early if a validator sent the wrong certificates.

afck:
## Motivation

We want to optimize how we download certificates and distribute work among the validators in the future. To cancel requests to other validators for the same certificate, we need to be sure that we got the certificate we requested.

## Proposal

Compare the downloaded certificate's hash with the requested hash immediately.

Also, remove a redundant argument from `download_certificates`: We always use it with `validator_nodes()`.

## Test Plan

CI

## Release Plan

## Links

### DIFF
--- a/linera-core/src/remote_node.rs
+++ b/linera-core/src/remote_node.rs
@@ -291,7 +291,19 @@ impl<N: ValidatorNode> RemoteNode<N> {
         if hashes.is_empty() {
             return Ok(Vec::new());
         }
-        self.node.download_certificates(hashes).await
+        let certificates = self.node.download_certificates(hashes.clone()).await?;
+        let returned = certificates
+            .iter()
+            .map(ConfirmedBlockCertificate::hash)
+            .collect();
+        ensure!(
+            returned == hashes,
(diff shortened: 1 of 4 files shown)


linera-core/src/remote_node.rs
```diff
@@ -291,7 +291,19 @@ impl<N: ValidatorNode> RemoteNode<N> {
...
             return Ok(Vec::new());
         }
-        self.node.download_certificates(hashes).await
+        let certificates = self.node.download_certificates(hashes.clone()).await?;
+        let returned = certificates
+            .iter()
+            .map(ConfirmedBlockCertificate::hash)
+            .collect();
+        ensure!(
+            returned == hashes,
```
deuszx, line 300: This expects the order to be the same as well, right? We probably maintain it but if it's different we'd fail here even if we actually got the same certs (just in different order).

afck: Yes. We do maintain the order, and I think it needs to be enforced. Callers of this method should be able to rely on the fact that they can e.g. `zip` their list of hashes and the certificates they got.

PR #4123, opened 2025-06-16
Calculate block time p99 in benchmark client

ndr-ds:
## Motivation

Block time is an important metric for us to know during benchmarks

## Proposal

Measure the block time p99, and print it every 5 seconds

## Test Plan

Deployed a network, ran a benchmark against it, saw the p99 being reported

## Release Plan

- Nothing to do / These changes follow the usual release cycle.

PR #4122, opened 2025-06-16
Make benchmark client scale better

ndr-ds:
## Motivation

We were needlessly using `spawn_blocking` for a lot of things, with the `cross_beam` channels... was a bit too complicated.

## Proposal

Simplify the code and make it scale better by using an `AtomicUsize` and tokio's `Notify`. This should make us be able to use up all the vCPUs even from a big benchy instance.

## Test Plan

Will deploy a network and benchmark against it using this

## Release Plan

- Nothing to do / These changes follow the usual release cycle.

PR #4150, opened 2025-06-19
Include pending message bundles in benchmark block proposals

ndr-ds:
## Motivation

We were not processing the inboxes during the benchmarks, which caused the inboxes to just grow indefinitely.

## Proposal

Include the pending messages with the block proposals. This will cause each block to do a bit more work, which will likely force us to use smaller block sizes in the benchmarks to control latency, but will prevent the server's memory from growing indefinitely. 

## Test Plan

Will deploy a network and make sure memory isn't linearly increasing anymore, as well as track the inbox sizes.

## Release Plan

### DIFF
--- a/linera-client/src/benchmark.rs
+++ b/linera-client/src/benchmark.rs
@@ -12,7 +12,7 @@ use std::{
 
 use hdrhistogram::Histogram;
 use linera_base::{
-    data_types::{Amount, Epoch},
+    data_types::Amount,
     identifiers::{AccountOwner, ApplicationId, ChainId},
     listen_for_shutdown_signals,
     time::Instant,
@@ -27,7 +27,7 @@ use linera_sdk::abis::fungible;
 use num_format::{Locale, ToFormattedString};
 use prometheus_parse::{HistogramCount, Scrape, Value};
 use tokio::{
-    sync::{mpsc, Barrier},
+    sync::{mpsc, Barrier, Notify},
     task, time,
 };
 use tokio_util::sync::CancellationToken;
@@ -38,8 +38,6 @@ const LATENCY_METRIC_PREFIX: &str = "linera_proxy_request_latency";
 
 #[derive(Debug, thiserror::Error)]
 pub enum BenchmarkError {
-    #[error("Failed to send message: {0}")]
-    CrossbeamSendError(#[from] crossbeam_channel::SendError<()>),
     #[error("Failed to join task: {0}")]
     JoinError(#[from] task::JoinError),
     #[error("Chain client error: {0}")]
@@ -66,13 +64,13 @@ pub enum BenchmarkError {
     NoDataYetForP99Calculation,
     #[error("Unexpected empty bucket")]
     UnexpectedEmptyBucket,
-    #[error("Failed to send message: {0}")]
+    #[error("Failed to send unit message: {0}")]
     TokioSendUnitError(#[from] mpsc::error::SendError<()>),
     #[error("Failed to create histogram: {0}")]
     HistogramCreationError(#[from] hdrhistogram::CreationError),
     #[error("Failed to record histogram: {0}")]
     HistogramRecordError(#[from] hdrhistogram::RecordError),
-    #[error("Failed to send message: {0}")]
+    #[error("Failed to send u64 message: {0}")]
     TokioSendU64Error(#[from] mpsc::error::SendError<u64>),
 }
 
@@ -94,13 +92,13 @@ impl<Env: Environment> Benchmark<Env> {
         transactions_per_block: usize,
         bps: usize,
         chain_clients: Vec<Vec<ChainClient<Env>>>,
-        epoch: Epoch,
         blocks_infos: Vec<Vec<(Vec<Operation>, AccountOwner)>>,
         committee: Committee,
         health_check_endpoints: Option<String>,
         runtime_in_seconds: Option<u64>,
     ) -> Result<(), BenchmarkError> {
         let bps_count = Arc::new(AtomicUsize::new(0));
+        let notifier = Arc::new(Notify::new());
         let barrier = Arc::new(Barrier::new(num_chain_groups + 1));
 
         let shutdown_notifier = CancellationToken::new();
@@ -110,6 +108,7 @@ impl<Env: Environment> Benchmark<Env> {
             &barrier,
             &shutdown_notifier,
             &bps_count,
+            &notifier,
             transactions_per_block,
             bps,
         );
@@ -131,12 +130,13 @@ impl<Env: Environment> Benchmark<Env> {
             let barrier_clone = barrier.clone();
             let block_time_quantiles_sender = block_time_quantiles_sender.clone();
             let bps_count_clone = bps_count.clone();
+            let notifier_clone = notifier.clone();
             let runtime_control_sender_clone = runtime_control_sender.clone();
             join_set.spawn(
                 async move {
                     Box::pin(Self::run_benchmark_internal(
                         chain_group_index,
-                        epoch,
+                        bps,
                         chain_group,
                         chain_clients,
                         shutdown_notifier_clone,
@@ -144,16 +144,16 @@ impl<Env: Environment> Benchmark<Env> {
                         committee,
                         block_time_quantiles_sender,
                         barrier_clone,
+                        notifier_clone,
                         runtime_control_sender_clone,
                     ))
                     .await?;
 
                     Ok(())
                 }
-                .instrument(tracing::info_span!(
-                    "benchmark_chain_group",
-                    chain_group_index = ?chain_group_index
-                )),
+                .instrument(
+                    tracing::info_span!("chain_group", chain_group_index = ?chain_group_index),
+                ),
             );
         }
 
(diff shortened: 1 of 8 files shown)


ndr-ds:
* **#4177** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4177?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4176** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4176?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4170** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4170?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4171** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4171?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4168** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4168?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4163** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4163?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4150** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4150?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4150?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#4158** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4158?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4151** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4151?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4123** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4123?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4122** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4122?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

ma2bd:
> which will likely force us to use smaller block sizes in the benchmarks to control latency,

not necessarily because receiving a transfer bundle is cheaper (it doesn't require a cross-chain request + confirmation).

ndr-ds:
I see 🤔 I did see a performance degradation on the validator's side when running locally, but I'll try on a deployed network and see, hopefully I'm wrong!

ma2bd:
> I see 🤔 I did see a performance degradation on the validator's side when running locally, but I'll try on a deployed network and see, hopefully I'm wrong!

Well, actual data win but I'd be surprised if you had to reduce the number of operations per block by 50%. (But maybe it would be good news because it would mean something else can be optimized, like inboxes for instance)

linera-client/src/benchmark.rs
```diff
@@ -495,22 +504,43 @@ impl<Env: Environment> Benchmark<Env> {
...
+
+        let mut first_block_proposal = true;
         loop {
             if shutdown_notifier.is_cancelled() {
                 info!("Shutdown signal received, stopping benchmark");
                 break;
             }
 
             let start = Instant::now();
+            let incoming_bundles = chain_client.pending_message_bundles().await?;
```
ma2bd, line 585: Are we tracking the size of this?

ma2bd: Also note that this function depends on the option `max_pending_message_bundles`.

ndr-ds: ~True, I'll make sure we always grab everything.~
As far as tracking it, since this on the client side we can't atm have Prometheus metrics for it.
We could print it (in a sampled way to not affect performance too much), but not sure that's what we want 🤔

ndr-ds: Actually, since the chains run in parallel, and now we're processing the messages, to prevent certain blocks from taking longer than others and making the chains get out of sync and create "out of order" messages, we probably actually need to set this to a number proportional to the number of transactions in the block

ndr-ds: Come to think of it, now that we're processing the messages, this design became a bit too racey 😅
I'm gonna put the PR to Draft and will tweak the design here a bit

linera-core/src/client/mod.rs
```diff
@@ -1784,18 +1785,37 @@ impl<Env: Environment> ChainClient<Env> {
...
+    ) -> Result<Vec<IncomingBundle>, ChainClientError> {
+        self.pending_message_bundles_internal(None).await
+    }
+
+    /// Obtains up to `self.options.max_pending_message_bundles` pending message bundles for the
+    /// local chain.
+    #[instrument(level = "trace")]
+    async fn pending_message_bundles(&self) -> Result<Vec<IncomingBundle>, ChainClientError> {
+        self.pending_message_bundles_internal(Some(self.options.max_pending_message_bundles))
+            .await
```
afck: Do we really need that? Why not just run the benchmark with `max_pending_message_bundles == usize::MAX`?

ndr-ds: Because I think that since the benchmarks can be long running, forgetting to set this option could cause a "memory leak". So IMHO it needs to be enforced, not optional 😅 If someone forgets, they get an issue and the benchmarks stop working as intended, also

afck: I'd just really _really_ like to avoid adding any more complications and indirections to the `client` module if it isn't absolutely necessary. Maybe we can print a warning or even panic if `max_pending_message_bundles` is too low?

linera-client/src/benchmark.rs
ma2bd: Did you?

ndr-ds: Yes, here https://github.com/linera-io/linera-protocol/pull/4158

linera-client/src/benchmark.rs
```diff
@@ -190,40 +185,43 @@ impl<Env: Environment> Benchmark<Env> {
...
+                        );
+                    } else {
+                        warn!(
+                            "Failed to achieve {} BPS/{} TPS, only achieved {} BPS/{} TPS",
+                            formatted_bps_goal,
+                            formatted_tps_goal,
+                            formatted_current_bps,
+                            formatted_current_tps,
+                        );
+                    }
```
deuszx, line 226: I know you're formatting only now but this method has the comment
```
The bps control task will control the BPS from the threads.
```
but it seems like the only thing it's doing is printing BPS/TPS?

linera-service/src/cli/main.rs
```diff
@@ -798,6 +798,14 @@ impl Runnable for Job {
                     confirm_before_start,
                     runtime_in_seconds,
                 } = benchmark_config;
+                assert!(
+                    options.context_options.max_pending_message_bundles >= transactions_per_block,
+                    "max_pending_message_bundles must be set to at least the same as the number of transactions per block ({transactions_per_block}) for benchmarking",
```
afck: (line length)

linera-service/src/cli_wrappers/wallet.rs
```diff
@@ -193,18 +193,42 @@ impl ClientWrapper {
...
+
+    #[cfg(feature = "benchmark")]
+    async fn command_with_arguments(
+        &self,
+        arguments: impl IntoIterator<Item = Cow<'_, str>>,
+    ) -> Result<Command> {
+        self.command_with_envs_and_arguments(
+            &[(
+                "RUST_LOG",
+                &std::env::var("RUST_LOG").unwrap_or(String::from("linera=debug")),
```
afck, line 225: Does this change the behavior of all the tests?

linera-service/src/cli_wrappers/wallet.rs
```diff
@@ -213,17 +237,14 @@ impl ClientWrapper {
...
+    fn required_command_arguments(&self) -> impl Iterator<Item = Cow<'_, str>> + '_ {
         [
             "--wallet".into(),
             self.wallet.as_str().into(),
             "--keystore".into(),
             self.keystore.as_str().into(),
             "--storage".into(),
             self.storage.as_str().into(),
-            "--max-pending-message-bundles".into(),
-            self.max_pending_message_bundles.to_string().into(),
```
afck, line 226: Are we even setting that in any tests?

This selection of arguments looks a bit arbitrary to me.

afck (approved):

ndr-ds:
@deuszx yes, my bad, I was gonna split the code cleanups into another PR, but it started getting reviews even as a draft, so I thought since people had already spent the time looking at it that splitting after might just make it confusing 😅 I'll split earlier next time

linera-service/src/cli_wrappers/wallet.rs
ndr-ds: All `ClientWrapper`s seem to set this to 10 thousand

linera-client/src/benchmark.rs
ndr-ds: Sorry, I temporarily removed the code that actually controls the BPS as a temporary test, just added it back

linera-service/src/cli_wrappers/wallet.rs
ndr-ds: This is just replicating the behavior of `command` below. And this `command_with_arguments` function is only used in benchmark tests

PR #4153, opened 2025-06-20
Don't require ancestor chains in validators.

### LINKED ISSUE #4085: Don't require ancestor chains in validators.
When a client uploads the first block of a chain, the validator should look up the committee (based on the admin chain, or the events it created), and verify the signatures. It should then, based on that, accept the chain description blob, even if it doesn't have the parent chain.

bart-linera:
## Motivation

Currently a validator cannot process a child chain without syncing all the ancestor chains up to the point where the child chain was created.

## Proposal

When a client uploads the first block of a chain, the validator should look up the committee and verify the signatures. It should then, based on that, accept the chain description blob, even if it doesn't have the parent chain.

## Test Plan

CI

## Release Plan

## Links

- closes #4085

### DIFF
--- a/linera-core/src/chain_worker/state/attempted_changes.rs
+++ b/linera-core/src/chain_worker/state/attempted_changes.rs
@@ -303,12 +303,28 @@ where
             let info = ChainInfoResponse::new(&self.state.chain, self.state.config.key_pair());
             return Ok((info, actions));
         }
-        let local_time = self.state.storage.clock().current_time();
-        self.state.ensure_is_active().await?;
-        // Verify the certificate.
-        let (epoch, committee) = self.state.chain.current_committee()?;
-        check_block_epoch(epoch, chain_id, block.header.epoch)?;
-        certificate.check(committee)?;
+
+        if height == BlockHeight::ZERO {
+            // For the initial proposal on a chain, we read the committee directly from
+            // storage. If the certificate is good for this committee, we will accept it -
+            // which will later allow us to also accept the chain description blob.
+            let epoch = block.header.epoch;
+            let committees = self
+                .state
+                .storage
+                .committees_for([epoch].into_iter().collect())
+                .await?;
+            let committee = committees
+                .get(&epoch)
+                .ok_or(WorkerError::UnknownEpoch { chain_id, epoch })?;
+            certificate.check(committee)?;
+        } else {
+            self.state.ensure_is_active().await?;
+            // Verify the certificate.
+            let (epoch, committee) = self.state.chain.current_committee()?;
+            check_block_epoch(epoch, chain_id, block.header.epoch)?;
+            certificate.check(committee)?;
+        }
         // This should always be true for valid certificates.
         ensure!(
             tip.block_hash == block.header.previous_block_hash,
@@ -354,7 +370,19 @@ where
             .filter_map(|blob_id| blobs.remove(blob_id))
             .collect::<Vec<_>>();
 
+        // If height is zero, we haven't initialized the chain state or verified the epoch before -
+        // do it now.
+        // This will fail if the chain description blob is still missing - but that's alright,
+        // because we already wrote the blob state above, so the client can now upload the
+        // blob, which will get accepted, and retry.
+        if height == BlockHeight::ZERO {
+            self.state.ensure_is_active().await?;
+            let (epoch, _) = self.state.chain.current_committee()?;
+            check_block_epoch(epoch, chain_id, block.header.epoch)?;
+        }
+
         // Execute the block and update inboxes.
+        let local_time = self.state.storage.clock().current_time();
         let chain = &mut self.state.chain;
         chain
             .remove_bundles_from_inboxes(block.header.timestamp, &block.body.incoming_bundles)
--- a/linera-core/src/node.rs
+++ b/linera-core/src/node.rs
@@ -9,7 +9,7 @@ use futures::stream::LocalBoxStream as BoxStream;
 use futures::stream::Stream;
 use linera_base::{
     crypto::{CryptoError, CryptoHash, ValidatorPublicKey},
-    data_types::{ArithmeticError, BlobContent, BlockHeight, NetworkDescription},
+    data_types::{ArithmeticError, Blob, BlobContent, BlockHeight, NetworkDescription},
     identifiers::{BlobId, ChainId},
 };
 use linera_chain::{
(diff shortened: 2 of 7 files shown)


linera-core/src/node.rs
```diff
@@ -103,12 +103,15 @@ pub trait ValidatorNode {
     async fn upload_blob(&self, content: BlobContent) -> Result<BlobId, NodeError>;
 
     /// Uploads the blobs to the validator.
-    fn upload_blobs(&self, blobs: Vec<Blob>) -> impl futures::Future<Output = Result<(), NodeError>>
-    where
-        Self: Sync,
-    {
-        async {
-            let tasks = blobs.into_iter().map(|blob| self.upload_blob(blob.into()));
+    fn upload_blobs(
```
ma2bd, line 109: Does someone know why we're not using `async` here? (Without a comment, it's a bit hard to see)

fizyk20: It doesn't compile with `async` - something about default implementations in a trait doesn't work with async functions. I'll add a comment.

bart-linera: It doesn't compile with `async` - something about default implementations in a trait doesn't work with async functions. I'll add a comment.

linera-core/src/chain_worker/state/attempted_changes.rs
```diff
@@ -354,7 +370,19 @@ where
             .filter_map(|blob_id| blobs.remove(blob_id))
             .collect::<Vec<_>>();
 
+        // If height is zero, we haven't initialized the chain state or verified the epoch before -
+        // do it now.
+        // This will fail if the chain description blob is still missing - but that's alright,
+        // because we already wrote the blob state above, so the client can now upload the
+        // blob, which will get accepted, and retry.
```
ma2bd, line 377: Can clients upload the chain description optimistically with the block proposal? We generally want to avoid predictable round trips between clients and servers (especially for some blobs).

bart-linera: I think this is what we used to do, but we explicitly changed it: #3108, #3153, #3204

ma2bd: Right, what we need is a bit more flexibility to re-optimize networking for small blobs

afck: That's true. I added https://github.com/linera-io/linera-protocol/issues/4159 as a sub-issue to https://github.com/linera-io/linera-protocol/issues/3048.

linera-core/src/node.rs
```diff
@@ -102,6 +102,24 @@ pub trait ValidatorNode {
...
+        blobs: Vec<Blob>,
+    ) -> impl futures::Future<Output = Result<(), NodeError>> {
+        let tasks: Vec<_> = blobs
+            .into_iter()
+            .map(|blob| self.upload_blob(blob.into()))
+            .collect();
+        async move {
+            futures::future::try_join_all(tasks).await?;
+            Ok(())
+        }
```
afck: Isn't this equivalent:

```suggestion
        futures::future::try_join_all(tasks)
```

linera-core/src/remote_node.rs
```diff
@@ -207,16 +207,6 @@ impl<N: ValidatorNode> RemoteNode<N> {
...
 
-    /// Uploads the blobs to the validator.
-    #[instrument(level = "trace")]
-    pub(crate) async fn upload_blobs(&self, blobs: Vec<Blob>) -> Result<(), NodeError> {
-        let tasks = blobs
-            .into_iter()
-            .map(|blob| self.node.upload_blob(blob.into()));
-        try_join_all(tasks).await?;
-        Ok(())
-    }
```
afck, line 218: Do we actually need that in the trait, rather than just in `RemoteNode`?

linera-service/tests/local_net_tests.rs
```diff
@@ -915,6 +915,96 @@ async fn test_sync_validator(config: LocalNetConfig) -> Result<()> {
...
+/// Tests if a validator can process blocks on a child chain without syncing the parent
+/// chain.
+// #[cfg_attr(feature = "scylladb", test_case(LocalNetConfig::new_test(Database::ScyllaDb, Network::Udp) ; "scylladb_udp"))]
+#[cfg_attr(feature = "scylladb", test_case(LocalNetConfig::new_test(Database::ScyllaDb, Network::Grpc) ; "scylladb_grpc"))]
+#[cfg_attr(feature = "storage-service", test_case(LocalNetConfig::new_test(Database::Service, Network::Grpc) ; "storage_service_grpc"))]
+// #[cfg_attr(feature = "storage-service", test_case(LocalNetConfig::new_test(Database::Service, Network::Tcp) ; "storage_service_tcp"))]
+#[cfg_attr(feature = "dynamodb", test_case(LocalNetConfig::new_test(Database::DynamoDb, Network::Grpc) ; "aws_grpc"))]
+// #[cfg_attr(feature = "scylladb", test_case(LocalNetConfig::new_test(Database::ScyllaDb, Network::Tcp) ; "scylladb_tcp"))]
+// #[cfg_attr(feature = "dynamodb", test_case(LocalNetConfig::new_test(Database::DynamoDb, Network::Tcp) ; "aws_tcp"))]
+// #[cfg_attr(feature = "dynamodb", test_case(LocalNetConfig::new_test(Database::DynamoDb, Network::Udp) ; "aws_udp"))]
```
afck, line 927: Are those commented-out cases all failing?

linera-core/src/node.rs
bart-linera: Not quite, because this will return a `Result<Vec<BlobId>, NodeError>` instead of `Result<(), NodeError>`, but it doesn't hurt and is cleaner. Nice catch, thanks!

linera-core/src/remote_node.rs
bart-linera: I'm using this in `sync_validator` where I don't have a `RemoteNode`, only an `Env::ValidatorNode` which has a `ValidatorNode` trait bound. It feels like it fits the trait, too, TBH - as a helper that calls `upload_blob` for multiple blobs.

linera-service/tests/local_net_tests.rs
bart-linera: TBH, that's just a copy-paste thing on my part :sweat_smile:  But I expect that since they were commented out on the test I've copied this from (`test_sync_validator`), they would be failing on this one, too, as the tests are very similar.

linera-core/src/node.rs
Twey: https://github.com/rust-lang/impl-trait-utils/issues/17

Twey: We have several comments about this across the codebase; I'm not sure if it makes sense to write it every time we want to use a default async function.

PR #4151, opened 2025-06-20
Remove non continuous version from linera benchmark

ndr-ds:
## Motivation

This version is used only in the tests, and it complicates the code a bit.

## Proposal

Remove it, and add a functionality to run the continuous mode only for a limited amout of time, which will also be useful in the future.

## Test Plan

Tests were adapted, so CI

## Release Plan

- Nothing to do / These changes follow the usual release cycle.

PR #4158, opened 2025-06-21
Make benchmark client design less racey

ndr-ds:
## Motivation

Now that we're gonna start processing all the messages as well, the current design was very prone to race conditions: blocks for different chains get executed in parallel, but the blocks actually depend on the blocks being executed by other chains/tasks, and the messages those blocks are sending.
So it was very easy for race conditions to happen where a block tries to execute a message that depends on another message that wasn't received yet.

## Proposal

Tweak the design a bit:
Instead of having one tokio task per chain, now each tokio task will instead execute a chain group: several chains, that only send messages to each other. We always execute one block for one of the chains, then move on to the next chain in the group, serially. This should prevent race conditions, as blocks being executed by different tokio tasks don't depend on each other anymore.

Now you can just specify the bps you want with `--bps` and the size of your blocks with `--transactions-per-block`, and we'll set the number of chain groups to be the number of CPUs you have, by default. This seems to be enough to saturate CPU (at least locally, have to check on GCP) as you ramp up the BPS amount.

## Test Plan

Deploy a local kind network, run the benchmark against it. I'll also run this against a GCP network

## Release Plan


ndr-ds:
* **#4177** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4177?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4176** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4176?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4170** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4170?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4171** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4171?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4168** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4168?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4163** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4163?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4150** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4150?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4158** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4158?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4158?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#4151** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4151?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4123** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4123?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4122** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4122?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

linera-client/src/benchmark.rs
```diff
@@ -620,33 +610,41 @@ impl<Env: Environment> Benchmark<Env> {
...
 
-    /// Generates information related to one block per chain, up to `num_chains` blocks.
+    /// Generates information related to one block per chain.
     pub fn make_benchmark_block_info(
-        keys: HashMap<ChainId, AccountOwner>,
+        benchmark_chains: Vec<Vec<(ChainId, AccountOwner)>>,
         transactions_per_block: usize,
         fungible_application_id: Option<ApplicationId>,
-    ) -> Vec<(ChainId, Vec<Operation>, AccountOwner)> {
+    ) -> Vec<Vec<(Vec<Operation>, AccountOwner)>> {
```
afck, line 618: Do we really gain anything by putting together this huge collection, rather than creating the operations directly in the chain (group) tasks, where we need them?

linera-service/src/cli/command.rs
```diff
@@ -38,9 +38,14 @@ const DEFAULT_BPS: usize = 10;
...
 pub struct BenchmarkCommand {
-    /// How many chains to use for the benchmark
-    #[arg(long, default_value_t = DEFAULT_NUM_CHAINS)]
-    pub num_chains: usize,
+    /// How many chains to use in each of the chain groups. One tokio task will be created per chain
+    /// group, for sending block proposals.
+    #[arg(long, default_value_t = DEFAULT_NUM_CHAINS_PER_CHAIN_GROUP)]
+    pub num_chains_per_chain_group: usize,
+
+    /// How many chain groups to use. If not provided, the number of CPUs will be used.
```
afck, line 44: (But each chain group task will spend a lot of time waiting for validators' responses, won't it? I guess we'll usually specify this and not use the default anyway…)

linera-service/src/cli/command.rs
```diff
@@ -23,7 +23,7 @@ use linera_rpc::config::CrossChainConfig;
 use serde::Serialize;
 
 #[cfg(feature = "benchmark")]
-const DEFAULT_NUM_CHAINS: usize = 10;
+const DEFAULT_NUM_CHAINS_PER_CHAIN_GROUP: usize = 2;
```
afck: Yes… I was wondering if there is any reason to change this to something other than 2 at all.

afck (approved):
Not blocking this, but:

I'm not sure this is the ideal solution. It depends on what scenario we're trying to measure, but there's no real reason to not have one task per chain. The problem with the implementation is just that:
* The clients all share the same storage.
* We use `pending_message_bundles`, which reads the messages from local storage.
* If one chain's thread has already processed the block locally but not finished uploading the certificate, the next chain's thread may read the incoming messages before the validators know about them.

One alternative approach would be to keep one task per chain, but instead of `pending_message_bundles`, we use channels that notify the next chain's task about the messages that we've already informed the validators about.

And with either approach we should probably check that most messages are received in the end.

ndr-ds:
Wouldn't that be altering the protocol behavior though, arguably making it faster? And therefore a bit of "cheating"? 😅
I wanted to keep things as close to real production behavior as possible 🤔 so isolating the chains that send messages to each other seemed like the most reasonable approach

linera-client/src/benchmark.rs
ndr-ds: Ideally the benchmark's runtime is as light as possible for performance, so pre-calculating the Operations, since they'll always just be repeated, seems reasonable to me 🤔

linera-service/src/cli/command.rs
ndr-ds: Yes, and unfortunately, for large enough machines, our bottleneck here starts being more the fact that they'll all be trying to talk to the same RocksDB instance than anything else (even network I/O). So we'll need to use `multi-benchmark` after all, as it's easier than trying to just make this use multiple rocksdb instances, as `multi-benchmark` already does that by being multi process

linera-service/src/cli/command.rs
ndr-ds: Probably not 🤔 1 if we want to test without cross chain messages, and 2 if we want to test with cross chain messages. Not sure there's much benefit with values greater than one. Maybe this can actually be just a boolean 🤔

afck:
I think the purpose of the benchmark is mainly to test _server_-side performance anyway?

Otherwise we're already cheating by having all clients use the same storage.

I don't see how my proposal would make things any worse.

PR #4161, opened 2025-06-23
Eliminate the `NotFound` for the `read_confirmed_block*`.

MathieuDutSik:
## Motivation

The `ViewError::NotFound` was used in the query. We want to eliminate it from the `ViewError`. So, this case can be ironed out.

## Proposal

The following is done:
* The `ViewError::not_found` is removed.
* The `read_confirmed_block` returns an option.
* The `read_confirmed_blocks_downward` is eliminated from `linera-storage` and when needed it is inlined (two contexts, one being a test).
* The variant is added to the `ChainClientError`. There are no more functions in `ChainClient` whose error is `ViewError`. 

## Test Plan

The CI.

## Release Plan

## Links

None.

### DIFF
--- a/linera-storage/src/db_storage.rs
+++ b/linera-storage/src/db_storage.rs
@@ -597,14 +597,13 @@ where
         Ok(test)
     }
 
-    async fn read_confirmed_block(&self, hash: CryptoHash) -> Result<ConfirmedBlock, ViewError> {
+    async fn read_confirmed_block(&self, hash: CryptoHash) -> Result<Option<ConfirmedBlock>, ViewError> {
         let block_key = bcs::to_bytes(&BaseKey::ConfirmedBlock(hash))?;
-        let maybe_value = self.store.read_value::<ConfirmedBlock>(&block_key).await?;
+        let value = self.store.read_value(&block_key).await?;
         #[cfg(with_metrics)]
         metrics::READ_CONFIRMED_BLOCK_COUNTER
             .with_label_values(&[])
             .inc();
-        let value = maybe_value.ok_or_else(|| ViewError::not_found("value for hash", hash))?;
         Ok(value)
     }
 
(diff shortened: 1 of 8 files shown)


linera-storage/src/db_storage.rs
```diff
@@ -675,18 +677,21 @@ where
...
         let mut hash = Some(from);
         let mut values = Vec::new();
         for _ in 0..limit {
             let Some(next_hash) = hash else {
                 break;
             };
             let value = self.read_confirmed_block(next_hash).await?;
+            let Some(value) = value else {
+                return Ok(None);
+            };
```
ma2bd: This one feels like a hack where we're just using `None` as an error value.

afck: Well, with sparse chains it's not an error for this block to not be available.

But we could do the same as elsewhere, of course, and return a `Vec<Option<_>>` instead of an `Option<Vec<_>>`.

MathieuDutSik: The use of `None` as an error value to indicate the missing entry is not a new thing. It is consistent with `linera-views` and other parts of `linera-storage`.

Maybe, it becomes an issue that the `None` does not indicate which hash was missed. This could be changed by having a return value that is an enum. Another way to improve the output is to add a variant `MissingConfirmedBlockDownward` next to a `MissingConfirmedBlock`.

Note that while not ideal, I consider this design to be better than what we have before:
* The error is clearly visible and not buried in a `ViewError`.
* The hash in question is available for further work.

ma2bd: This function is used only once (and is not particularly optimized), so I suppose we could also avoid the discussion and inline it.

MathieuDutSik: Ok, the code has been restructured accordingly.

afck (approved):

ma2bd (approved):

PR #4166, opened 2025-06-24
Committee destination for the block exporter

### LINKED ISSUE #3668: Implement the `Committee` destination
Similar to `Validator` but all the validators in the current committee are updated.

The current set of validators can be obtained by sending a GRPC requests `ChainInfoQuery` to the `ValidatorWorker` in charge of the admin chain.

See `ValidatorWorkerClient` in `linera_rp/src/grpc/server.rs`

usagi32:
## Motivation
closes #3668 
to be rebased on top of  #3949
## Proposal
- Collected all the detached tasks and thread join handles properly. Now we can join them safely rather than dropping them
- Implemented the committee destination for the block exporter. This will target all the members of the current committee
- To get the information about the current committee, the `BlockProcessor` during its normal scan of blocks will look for committee change transactions in the block. If found, we serialize the blob for the current committee, take their gRPC endpoints and start new validator exporters based on them.
- The committee destination validators are separately handled than the other destinations, hence the previous logic about about the other destination kinds is unchanged
## Test Plan

unit test + CI

## Release Plan

### DIFF
--- a/linera-client/src/config.rs
+++ b/linera-client/src/config.rs
@@ -232,6 +232,9 @@ pub struct BlockExporterConfig {
 pub struct DestinationConfig {
     /// The destination URIs to export to.
     pub destinations: Vec<Destination>,
+    /// Export blocks to the current committee.
+    #[serde(default)]
+    pub committee_destination: bool,
 }
 
 // Each destination has an ID and a configuration.
--- a/linera-service/src/exporter/exporter_service.rs
+++ b/linera-service/src/exporter/exporter_service.rs
@@ -1,8 +1,6 @@
 // Copyright (c) Zefchain Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-use std::thread::JoinHandle;
-
 use async_trait::async_trait;
 use linera_core::worker::Reason;
 use linera_rpc::grpc::api::{
@@ -46,10 +44,9 @@ impl ExporterService {
         self,
         cancellation_token: CancellationToken,
         port: u16,
-        block_processor_handle: Option<JoinHandle<Result<(), ExporterError>>>,
     ) -> core::result::Result<(), ExporterError> {
         info!("Linera exporter is running.");
-        self.start_notification_server(port, cancellation_token, block_processor_handle)
+        self.start_notification_server(port, cancellation_token)
             .await
     }
 
@@ -57,7 +54,6 @@ impl ExporterService {
         self,
         port: u16,
         cancellation_token: CancellationToken,
-        block_processor_handle: Option<JoinHandle<Result<(), ExporterError>>>,
     ) -> core::result::Result<(), ExporterError> {
         let endpoint = get_address(port);
         info!(
@@ -76,10 +72,6 @@ impl ExporterService {
             .serve_with_shutdown(endpoint, cancellation_token.cancelled_owned())
             .await
             .expect("a running notification server");
-
-        if let Some(some_handle) = block_processor_handle {
-            some_handle.join().unwrap()?;
-        }
 
         Ok(())
     }
@@ -119,7 +111,7 @@ mod test {
         let cancellation_token = CancellationToken::new();
         let (tx, mut rx) = unbounded_channel();
         let server = ExporterService::new(tx);
-        let server_handle = tokio::spawn(server.run(cancellation_token.clone(), port, None));
+        let server_handle = tokio::spawn(server.run(cancellation_token.clone(), port));
         LocalNet::ensure_grpc_server_has_started("test server", port as usize, "http").await?;
 
         let mut client = NotifierServiceClient::connect(format!("http://{endpoint}")).await?;
--- a/linera-service/src/exporter/exporter_service.rs
+++ b/linera-service/src/exporter/exporter_service.rs
@@ -1,6 +1,8 @@
 // Copyright (c) Zefchain Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
+use std::thread::JoinHandle;
+
 use async_trait::async_trait;
 use linera_core::worker::Reason;
 use linera_rpc::grpc::api::{
@@ -44,9 +46,10 @@ impl ExporterService {
         self,
         cancellation_token: CancellationToken,
         port: u16,
+        block_processor_handle: JoinHandle<Result<(), ExporterError>>,
     ) -> core::result::Result<(), ExporterError> {
         info!("Linera exporter is running.");
-        self.start_notification_server(port, cancellation_token)
+        self.start_notification_server(port, cancellation_token, block_processor_handle)
             .await
     }
 
@@ -54,6 +57,7 @@ impl ExporterService {
         self,
         port: u16,
         cancellation_token: CancellationToken,
+        block_processor_handle: JoinHandle<Result<(), ExporterError>>,
     ) -> core::result::Result<(), ExporterError> {
         let endpoint = get_address(port);
         info!(
@@ -72,6 +76,8 @@ impl ExporterService {
             .serve_with_shutdown(endpoint, cancellation_token.cancelled_owned())
             .await
             .expect("a running notification server");
+
+        block_processor_handle.join().unwrap()?;
 
         Ok(())
     }
@@ -97,6 +103,8 @@ fn parse_notification(notification: Notification) -> core::result::Result<BlockI
 #[cfg(test)]
 mod test {
 
+    use std::thread;
+
     use linera_base::{crypto::CryptoHash, identifiers::ChainId, port::get_free_port};
     use linera_core::worker::Notification;
     use linera_rpc::grpc::api::notifier_service_client::NotifierServiceClient;
@@ -109,10 +117,12 @@ mod test {
     async fn test_notification_server() -> anyhow::Result<()> {
         let port = get_free_port().await?;
         let endpoint = format!("127.0.0.1:{port}");
-        let cancellation_token = CancellationToken::new();
+        let cancellation_token: CancellationToken = CancellationToken::new();
         let (tx, mut rx) = unbounded_channel();
+        let dummy_thread = thread::spawn(|| Ok(()));
         let server = ExporterService::new(tx);
-        let server_handle = tokio::spawn(server.run(cancellation_token.clone(), port));
+        let server_handle =
+            tokio::spawn(server.run(cancellation_token.clone(), port, dummy_thread));
         LocalNet::ensure_grpc_server_has_started("test server", port as usize, "http").await?;
 
         let mut client = NotifierServiceClient::connect(format!("http://{endpoint}")).await?;
--- a/linera-service/src/exporter/exporter_service.rs
+++ b/linera-service/src/exporter/exporter_service.rs
@@ -46,7 +46,7 @@ impl ExporterService {
         self,
         cancellation_token: CancellationToken,
         port: u16,
-        block_processor_handle: JoinHandle<Result<(), ExporterError>>,
+        block_processor_handle: Option<JoinHandle<Result<(), ExporterError>>>,
     ) -> core::result::Result<(), ExporterError> {
         info!("Linera exporter is running.");
         self.start_notification_server(port, cancellation_token, block_processor_handle)
@@ -57,7 +57,7 @@ impl ExporterService {
         self,
         port: u16,
         cancellation_token: CancellationToken,
-        block_processor_handle: JoinHandle<Result<(), ExporterError>>,
+        block_processor_handle: Option<JoinHandle<Result<(), ExporterError>>>,
     ) -> core::result::Result<(), ExporterError> {
         let endpoint = get_address(port);
         info!(
@@ -77,7 +77,9 @@ impl ExporterService {
             .await
             .expect("a running notification server");
 
-        block_processor_handle.join().unwrap()?;
+        if let Some(some_handle) = block_processor_handle {
+            some_handle.join().unwrap()?;
+        }
 
         Ok(())
     }
@@ -102,9 +104,6 @@ fn parse_notification(notification: Notification) -> core::result::Result<BlockI
 
 #[cfg(test)]
 mod test {
-
-    use std::thread;
-
     use linera_base::{crypto::CryptoHash, identifiers::ChainId, port::get_free_port};
     use linera_core::worker::Notification;
     use linera_rpc::grpc::api::notifier_service_client::NotifierServiceClient;
@@ -117,12 +116,10 @@ mod test {
     async fn test_notification_server() -> anyhow::Result<()> {
         let port = get_free_port().await?;
         let endpoint = format!("127.0.0.1:{port}");
-        let cancellation_token: CancellationToken = CancellationToken::new();
+        let cancellation_token = CancellationToken::new();
         let (tx, mut rx) = unbounded_channel();
-        let dummy_thread = thread::spawn(|| Ok(()));
         let server = ExporterService::new(tx);
-        let server_handle =
-            tokio::spawn(server.run(cancellation_token.clone(), port, dummy_thread));
+        let server_handle = tokio::spawn(server.run(cancellation_token.clone(), port, None));
         LocalNet::ensure_grpc_server_has_started("test server", port as usize, "http").await?;
 
         let mut client = NotifierServiceClient::connect(format!("http://{endpoint}")).await?;
--- a/linera-service/src/exporter/runloops/block_processor/mod.rs
+++ b/linera-service/src/exporter/runloops/block_processor/mod.rs
@@ -1,8 +1,12 @@
 // Copyright (c) Zefchain Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-use std::{future::IntoFuture, time::Duration};
+use std::{
+    future::{Future, IntoFuture},
+    time::Duration,
+};
 
+use linera_execution::committee::Committee;
 use linera_storage::Storage;
 use tokio::{
     sync::mpsc::{UnboundedReceiver, UnboundedSender},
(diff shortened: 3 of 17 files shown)


deuszx:
@usagi32 please, clean up the git history here to include only the commits relevant to the PR.

linera-service/src/exporter/runloops/block_processor/walker.rs
```diff
@@ -171,9 +176,24 @@ impl ProcessedBlock {
...
         dependencies.extend(message_senders);
 
+        let new_committee = block.block().body.operations.iter().find_map(|m| {
+            if let Operation::System(boxed) = m {
+                if let SystemOperation::Admin(AdminOperation::CreateCommittee {
+                    blob_hash, ..
+                }) = &**boxed
+                {
+                    let committee_blob = BlobId::new(*blob_hash, BlobType::Committee);
+                    return Some(committee_blob);
```
deuszx, line 186: Is this `return` necessary?

linera-service/src/exporter/state.rs
```diff
@@ -195,3 +223,111 @@ impl DestinationStates {
             .load(Ordering::Acquire)
     }
 }
+
+#[derive(Debug, Clone, Default)]
+pub(crate) struct CommitteeState {
+    states: Arc<ArcSwap<Box<[CommitteeMember]>>>,
```
ma2bd: What's this complex type for?

linera-service/src/exporter/exporter_service.rs
```diff
@@ -109,10 +117,12 @@ mod test {
     async fn test_notification_server() -> anyhow::Result<()> {
         let port = get_free_port().await?;
         let endpoint = format!("127.0.0.1:{port}");
-        let cancellation_token = CancellationToken::new();
+        let cancellation_token: CancellationToken = CancellationToken::new();
```
ma2bd: The type annotation is not needed

linera-service/src/exporter/exporter_service.rs
```diff
@@ -109,10 +117,12 @@ mod test {
...
         let (tx, mut rx) = unbounded_channel();
+        let dummy_thread = thread::spawn(|| Ok(()));
         let server = ExporterService::new(tx);
-        let server_handle = tokio::spawn(server.run(cancellation_token.clone(), port));
+        let server_handle =
+            tokio::spawn(server.run(cancellation_token.clone(), port, dummy_thread));
```
ma2bd: Can we use an option type pass `None` instead of `dummy_thread`?

linera-client/src/config.rs
```diff
@@ -232,6 +232,9 @@ pub struct BlockExporterConfig {
 pub struct DestinationConfig {
     /// The destination URIs to export to.
     pub destinations: Vec<Destination>,
+    /// Export blocks to the current committee.
+    #[serde(default)]
+    pub committee_destination: bool,
```
ma2bd, line 237: This `bool` is a little weird. Seems like we should instead have an `enum Destination` and a `struct GrpcEndpoint`
```
enum Destination {
    Indexer(GrpcEndpoint),
    Validator(GrpcEndpoint),
    Committee,
}

struct GrpcEndpoint {
    /// The gRPC network protocol.
    pub tls: TlsConfig,
    /// The host name of the target destination (IP or hostname).
    pub endpoint: String,
    /// The port number of the target destination.
    pub port: u16,

}
```

ma2bd: Did you avoid this because of the limitation of the YAML format? I'm seeing `#[serde(untagged)]` being suggested on forums: https://serde.rs/enum-representations.html#untagged

linera-service/src/exporter/runloops/mod.rs
```diff
@@ -170,72 +92,282 @@ where
...
 
     Ok(())
 }
 
-#[allow(clippy::too_many_arguments)]
-fn spawn_exporter_task_on_set<F, S>(
-    set: &mut JoinSet<Result<(), anyhow::Error>>,
-    signal: F,
-    id: DestinationId,
+fn start_theadpool(number_of_threads: usize) -> Vec<(JoinHandle<()>, Handle)> {
```
ma2bd: typo: `theadpool`

linera-service/src/exporter/runloops/mod.rs
```diff
@@ -170,72 +92,282 @@ where
...
+        handles.push((thread_handle, runtime_handle));
+    }
+
+    handles
+}
+
+enum DestinationContext {
+    Committee(usize),
+    NonCommittee(Vec<Destination>),
+}
```
ma2bd: This is a bit odd

linera-service/src/exporter/state.rs
usagi32: To atomically discard the current array of committee validators and replace it at the same time with another. Ensures all threads will see the same thing safely

linera-client/src/config.rs
usagi32: The Committee would otherwise had to be made part of the vector which seemed problematic unless we have reasons to do so, therefore I took it out to the parent struct. This made it type-safe statically saving some hassles

linera-service/src/exporter/runloops/mod.rs
usagi32: Just a helper type

linera-service/src/exporter/state.rs
deuszx: So it's Arc<RwLock<Arc<T>> ?

Reading the documentation of arc-swap - why not use a CoW type instead?

deuszx: I think this is a perfect example of premature optimization. The docs of `arc-swap` crate talk about CPU-contention when using `RwLock<Arc>` but our use case here is not that. The committees change very rarely and right now don't change at all. If they do change then we don't need a millisecond precision - the lock can be taken for 100ms and nothing will happen.

Refactor to remove the crate.

linera-service/src/exporter/runloops/mod.rs
```diff
@@ -170,72 +92,282 @@ where
...
+    fn from_config(config: DestinationConfig) -> Self {
+        Self::NonCommittee(config.destinations.clone())
+    }
+
+    fn from_committee_addresses(addresses: &[String]) -> Self {
+        Self::Committee(addresses.len())
+    }
+}
+
+struct ThreadPoolState<F, S>
```
deuszx: Can you explain the purpose of this struct? It mixes a domain-related logic (it knows about committe destination vs non-committee ones), thread management (spawns and joins threads). It can mess up the whole exporter if is mismanages the ids.

To start with:
- extract to a separate file
- add comments
- add tests

usagi32:
> what happens if we have a "validator destination" that is in the committee and then falls out of it (or vice versa)

I think we can exclude those validators from the current committee by iterating and only start the remaining ones. I'll try to make this change

deuszx:
> > what happens if we have a "validator destination" that is in the committee and then falls out of it (or vice versa)
> 
> I think we can exclude those validators from the current committee by iterating and only start the remaining ones. I'll try to make this change

That's why I asked for the description of the algo first - these are design decisions that should be clear to review.

For example, I think that if you did the change I asked on the first PR - that the destination ID is derived from the destination config, then (requests for) updates would be idempotent. A validator that is part of the committee has its destination ID derived from the host, protocol, etc and that will match the destination ID for that validator if it was in the configuration as "validator destination".

linera-service/src/exporter/state.rs
usagi32: done

linera-service/src/exporter/runloops/mod.rs
usagi32: removed btw

linera-service/src/exporter/runloops/block_processor/walker.rs
usagi32: yes

linera-service/src/exporter/runloops/mod.rs
usagi32: done

linera-service/src/exporter/exporter_service.rs
```diff
@@ -44,16 +46,18 @@ impl ExporterService {
         self,
         cancellation_token: CancellationToken,
         port: u16,
+        block_processor_handle: Option<JoinHandle<Result<(), ExporterError>>>,
```
deuszx: I think this is unnecessary. IIUC it's used as a way for controlling when the `start_notification_server` yields - it'll wait for the `block_processor_handle` before shutting down. Seems like a way to synchronize the completion of all futures. I think we should use a `JoinSet` or [TaskTracker](https://docs.rs/tokio-util/0.7.15/tokio_util/task/task_tracker/struct.TaskTracker.html) instead.

linera-service/src/exporter/runloops/block_processor/mod.rs
```diff
@@ -68,7 +81,31 @@ where
...
                     match walker.walk(next_block_notification).await {
-                        Ok(_) => {},
+                        Ok(maybe_new_committee) => {
+                            if let Some(blob_id) = maybe_new_committee {
+                                let blob = match self.storage.get_blob(blob_id).await {
+                                    Ok(blob) => blob,
+                                    Err(error) => {
+                                        tracing::error!("unable to get the committee blob: {:?} from storage, , received error: {:?}", blob_id, error);
+                                        continue;
+                                    },
```
deuszx: This should be a fatal error. We're trying to load a blob with the new committee but it's missing from the (validator) storage. This breaks our basic assumption that the storage has all the data necessary (blocks and blobs). We don't prune the state so we shouldn't be missing anything. Unless the storage we're connecting to (validator) is not up-to itself in which case the block exporter won't be able to do its job.

linera-service/src/exporter/runloops/block_processor/mod.rs
```diff
@@ -68,7 +81,31 @@ where
...
+                                        tracing::error!("unable to serialize the committee blob: {:?}, received error: {:?}", blob_id, e);
+                                        continue;
+                                    }
+                                };
+
+                                let addresses = committee.validator_addresses().map(|(_, address)| address.to_owned()).collect::<Vec<String>>();
+                                self.pool_state.shutdown_current_committee().await;
+
+                                self.storage.new_committee(addresses.clone());
+                                self.pool_state.start_committee_exporters(addresses.as_ref());
```
deuszx, line 109: Alternative design would be to get rid of `pool_state` from the block processor and have a channel on which we send the new committee. The receiver would interpret that however they like.

linera-service/src/exporter/state.rs
```diff
@@ -118,11 +130,26 @@ where
...
+
+    pub fn replace_current_committee(&self, addresses: Vec<String>) {
+        let state = addresses
+            .into_iter()
+            .map(CommitteeMember::new)
+            .collect::<Vec<_>>()
+            .into_boxed_slice();
+
+        let mut guard = self.committee_state.get().states.lock().unwrap();
+        *guard = state;
```
deuszx, line 146: If new committee includes members of the old one (and almost always that's the case) then we'd overwrite their "destination state" with 0 unncessarily.

linera-service/src/exporter/runloops/validator_exporter.rs
```diff
@@ -55,17 +68,36 @@ where
...
+                self.storage
+                    .get_committee_member_address(self.destination_id),
+                self.storage
+                    .load_committee_member_export_state(self.destination_id),
+            ),
+            ValidatorDestinationKind::NonCommitteeMember(destination) => (
+                destination.address(),
+                self.storage.load_destination_state(self.destination_id),
+            ),
+        };
```
deuszx, line 90: This would be unnecessary if we had a unified way of storing "destination state" for committee and non-committee validators. I think we should be able to do that as - from the perspective of the storage - that difference is irrelevant. It's the higher-level logic components that need to know the difference in order to decide which exporters should be replaced when a new committee is formed.

linera-service/src/exporter/runloops/validator_exporter.rs
```diff
@@ -98,13 +130,13 @@ where
 {
     fn new(
         node: GrpcClient,
-        destination_id: DestinationId,
         storage: &'a ExporterStorage<S>,
+        destination_state: Arc<AtomicU64>,
```
deuszx, line 134: I think I liked the previous version more b/c it didn't require having `Arc<[Arc<AtomicU64>]>`

linera-service/src/exporter/state.rs
```diff
@@ -176,22 +203,117 @@ impl DestinationStates {
...
+        self.states
+            .get(id as usize)
+            .expect("DestinationId should correspond")
+            .clone()
+    }
+}
+
+#[derive(Debug, Clone, Default)]
+pub(crate) struct CommitteeState {
+    states: Arc<Mutex<Box<[CommitteeMember]>>>,
```
deuszx, line 222: It'd require a moderate rewrite but I think a `CommitteeMember` should be a `DestinationId` instead and `DestinationId` should be derived from the endpoint (address) to which we will be pushing the updates to. This would simplify a couple of things:
- we no longer need to differentiate between committee and non-committee destinations in the storage b/c `load_destination_state(destinationId)` will work the same for both
- when replacing a committee we'd be really replacing the addresses to which we push, their (destination) states are tracked separately

linera-service/src/exporter/runloops/mod.rs
```diff
@@ -170,72 +92,88 @@ where
...
-                ValidatorExporter::new(node_provider, id, storage, destination, work_queue_size);
+    for _ in 0..number_of_threads {
+        let runtime = tokio::runtime::Builder::new_current_thread()
+            .enable_all()
+            .build()
+            .unwrap();
+        let runtime_handle = runtime.handle().clone();
+        let thread_handle = thread::spawn(move || {
+            runtime.block_on(std::future::pending::<()>());
+        });
```
deuszx, line 151: This assumes that the workload per each runtime/thread will be (more or less) uniform which we cannot guarantee. Some destinations will be indexers, some validators (among which the size of gaps will differ). If we are sure that a thread-per-cpu runtime is the best - we should use a dedicated runtime for that like [monoio](https://github.com/bytedance/monoio) but unless we have hard data confirming that, we shouldn't optimize prematurely.

Replacing this with a multithreaded runtime with `number_of_threads` workers will simplify the code.

linera-client/src/config.rs
deuszx, line 237: What is this field for? it doesn't do anything. It's read only in one place only - in `cli_wrappers/LocalNet#generate_block_exporter_config` when writing a toml file.

It seems like we export to committee always anyway.

linera-service/src/exporter/runloops/mod.rs
```diff
@@ -393,18 +333,106 @@ mod test {
...
+            NodeOptions {
+                send_timeout: Duration::from_millis(4000),
+                recv_timeout: Duration::from_millis(4000),
+                retry_delay: Duration::from_millis(1000),
+                max_retries: 10,
+            },
+            0,
+            1,
+            DestinationConfig {
+                committee_destination: true,
```
deuszx, line 384: To prove that this field doesn't do antything, I set it to `false` and the test passes.

linera-service/src/exporter/runloops/task_manager.rs
```diff
@@ -0,0 +1,315 @@
...
+        let startup_addresses = self
+            .startup_destinations
+            .iter()
+            .map(|destination| destination.address())
+            .collect::<Vec<String>>();
+        for address in committee_addresses {
+            if !startup_addresses.contains(address) {
+                self.spawn_committee();
+            }
+        }
```
deuszx, line 79: This filters out the "startup destinations" (like validators) out from the committee - that's good, we won't duplicate the work.

linera-service/src/exporter/runloops/task_manager.rs
```diff
@@ -0,0 +1,315 @@
...
+        let id = self.next_spawn_id;
+        self.next_spawn_id += 1;
+        id
+    }
+
+    fn next_committee_spawn_id(&mut self) -> u16 {
+        let id = self.next_committee_spawn_id;
+        self.next_committee_spawn_id += 1;
+        id
+    }
```
deuszx, line 119: This is dangerous - both `next_spawn_id` and `next_committee_spawn_id` start from 0 meaning they will get assigned the same ID values as `destination_id` which are used to query the storage, track their state of sync w.r.t. to the canonical chain. This isn't happening b/c we store committee's members state separately form the rest but it's easy to miss and there are no tests/guardrails to prevent from this.

linera-service/src/exporter/runloops/task_manager.rs
```diff
@@ -0,0 +1,315 @@
...
+        <F as IntoFuture>::IntoFuture: Future<Output = ()> + Send + Sync + 'static,
+    {
+        let startup_addresses = self
+            .startup_destinations
+            .iter()
+            .map(|destination| destination.address())
+            .collect::<Vec<String>>();
+        for address in committee_addresses {
+            if !startup_addresses.contains(address) {
+                self.spawn_committee();
```
deuszx, line 77: This will start a new exporter, starting from `destination_state = 0`, for every committee member every time _someone_ in a committee changes. That's unnecessary - a validator that was in old committee and is also in a new one doesn't need to start from scratch.

linera-service/src/exporter/runloops/task_manager.rs
```diff
@@ -0,0 +1,315 @@
...
+        let number_of_threads = self.threads.len();
+        let handle = self
+            .threads
+            .get_mut(self.next_thread_index)
+            .expect("handle_index is manually checked below");
+
+        self.next_thread_index += 1;
+        if self.next_thread_index == number_of_threads {
+            self.next_thread_index = 0;
+        }
```
deuszx, line 131: This can be simplified to `self.next_thread_index = next_thread_index % number_of_threads`

linera-service/src/exporter/runloops/task_manager.rs
```diff
@@ -0,0 +1,315 @@
...
+            self.node_provider.clone(),
+            id,
+            self.storage.clone().unwrap(),
+            destination,
+            self.work_queue_size,
+        );
+        let handle = self
+            .runtime_handle
+            .spawn(exporter_task.run_with_shutdown(self.shutdown_signal.clone()));
+        let _abort_handle = self
```
deuszx, line 229: Ideally we'd collect abort handles as well.

linera-service/src/exporter/runloops/task_manager.rs
```diff
@@ -0,0 +1,315 @@
...
+            .committee_task_handles
+            .get_or_insert_default()
+            .spawn(handle);
```
deuszx, line 232: Why do we spawn an already spawned task? 🤔

linera-service/src/exporter/runloops/task_manager.rs
```diff
@@ -0,0 +1,315 @@
...
+    runtime::Handle,
+    task::{JoinError, JoinSet},
+};
+
+use crate::storage::ExporterStorage;
+
+/// This type manages tasks like spawning different exporters on the different
+/// threads, discarding the committees and joining every thread properly at the
+/// end for graceful shutdown of the process.
+pub(super) struct ThreadPoolState<F, S>
```
deuszx, line 23: I am not able to confirm this is working as intended so I am going to refactor it to shared multithreaded tokio Runtime.

linera-service/src/exporter/exporter_service.rs
deuszx: Partially done in [Do not pass BlockProcessor handle to exporter service. We will manage it differently](https://github.com/linera-io/linera-protocol/pull/4200/commits/53d9f57e883bcfe79b02c3cf04c6e52755071eae)

linera-service/src/exporter/runloops/block_processor/mod.rs
deuszx: Done in [Raise an error if an expected blob is missing from the validator storage](https://github.com/linera-io/linera-protocol/pull/4200/commits/db3b344dba931b39edfdf191516002c7a693cc6b)

linera-service/src/exporter/runloops/mod.rs
deuszx: Done in [Respect the committee_update flag in the block exporter](https://github.com/linera-io/linera-protocol/pull/4200/commits/52aad562486b8f3cf74028f86f01cb1c3a82ed99)

linera-client/src/config.rs
deuszx: Fixed in [Respect the committee_update flag in the block exporter](https://github.com/linera-io/linera-protocol/pull/4200/commits/52aad562486b8f3cf74028f86f01cb1c3a82ed99)

linera-service/src/exporter/state.rs
deuszx: When working on this I extended the `test_committee_destination` test in a way that we try to update the committee again - that test fails https://github.com/linera-io/linera-protocol/pull/4166/commits/e196f8e683b04978cfe79e0ee41d783e2b65e0ec

deuszx:
Superseeded by #4200

PR #4177, opened 2025-06-25
Add process ID to multi-benchmark in case we want to run several multi-benchmark processes in the same machine

ndr-ds:
## Motivation

Using the same benchy instance with local SSD to benchmark different networks at the same time is currently not possible because it assumes one runner per machine (which is the most common use case). However, I have needed to benchmark different networks at the same time to save time.

## Proposal

Create a PID directory for each instance of the `multi-benchmark` that is running, within the local ssd directory, to save the client state (wallet and DB)

## Test Plan

Using this to benchmark several networks at the same time

## Release Plan

### DIFF
--- a/linera-service/src/cli/main.rs
+++ b/linera-service/src/cli/main.rs
@@ -2126,7 +2126,7 @@ Make sure to use a Linera client compatible with this network.
         ClientCommand::MultiBenchmark {
             processes,
             faucet,
-            ssd_dir,
+            client_state_dir,
             command,
             delay_between_processes,
         } => {
@@ -2139,7 +2139,14 @@ Make sure to use a Linera client compatible with this network.
 
             let clients = (0..*processes)
                 .map(|n| {
-                    let path_provider = PathProvider::new(ssd_dir)?;
+                    let path_provider = if let Some(client_state_dir) = client_state_dir {
+                        let pid = std::process::id();
+                        let subdir = PathBuf::from(client_state_dir).join(pid.to_string());
+                        std::fs::create_dir_all(&subdir)?;
+                        PathProvider::new(&Some(subdir.to_string_lossy().to_string()))?
+                    } else {
+                        PathProvider::new(client_state_dir)?
+                    };
                     Ok(Arc::new(ClientWrapper::new_with_extra_args(
                         path_provider,
                         Network::Grpc,
--- a/linera-service/src/cli/command.rs
+++ b/linera-service/src/cli/command.rs
@@ -474,17 +474,18 @@ pub enum ClientCommand {
     #[cfg(feature = "benchmark")]
     MultiBenchmark {
         /// The number of `linera benchmark` processes to run in parallel.
-        #[arg(long = "processes", default_value = "1")]
+        #[arg(long, default_value = "1")]
         processes: usize,
 
         /// The faucet (which implicitly defines the network)
-        #[arg(long = "faucet")]
+        #[arg(long)]
         faucet: String,
 
-        /// If running on local SSD, specify the directory to store the storage and wallet.
+        /// If specified, a directory named after the process ID will be created in this directory,
+        /// and the client state will be stored there.
         /// If not specified, a temporary directory will be used for each client.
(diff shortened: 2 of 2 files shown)


ndr-ds:
* **#4177** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4177?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4177?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

linera-service/src/cli/main.rs
```diff
@@ -2139,7 +2139,14 @@ Make sure to use a Linera client compatible with this network.
 
             let clients = (0..*processes)
                 .map(|n| {
-                    let path_provider = PathProvider::new(ssd_dir)?;
+                    let path_provider = if let Some(client_state_dir) = client_state_dir {
+                        let subdir =
+                            PathBuf::from(client_state_dir).join(rand::random::<u16>().to_string());
```
ma2bd: `u16` is too short. You'll start getting collisions around 256 entries.

There exist APIs to create unique temporary directories (including in a given folder)

ndr-ds: I don't think I'll be running hundreds of these 🤔 I currently need to start each manually. This is `linera multi-benchmark` not `linera benchmark` 🤔
But true, using a temp dir thing might be better in general and more readable, I'll do that

linera-service/src/cli/main.rs
```diff
@@ -2137,7 +2137,16 @@ Make sure to use a Linera client compatible with this network.
 
             let clients = (0..*processes)
                 .map(|n| {
-                    let path_provider = PathProvider::new(ssd_dir)?;
+                    let path_provider = if let Some(client_state_dir) = client_state_dir {
+                        PathProvider::new(&Some(
```
ma2bd: Would it make sense to use `PathProvider::TemporaryDirectory` ?

linera-service/src/cli/main.rs
```diff
@@ -2137,7 +2137,16 @@ Make sure to use a Linera client compatible with this network.
...
+                            tempfile::tempdir_in(client_state_dir)?
+                                .keep()
+                                .display()
+                                .to_string(),
+                        ))?
+                    } else {
+                        PathProvider::new(client_state_dir)?
```
ma2bd: nit: This constructor `new` taking an option is a little surprising / specific. Perhaps `from_cli_option` would be a better name.

linera-service/src/cli/main.rs
ndr-ds: No, because `TemporaryDirectory` just creates a random directory (probably under `/tmp`). We want to be able to tell it where to create the directory, because we want this stuff to run on local SSD, which is mounted in a specific directory

linera-service/src/cli/main.rs
ndr-ds: `new` taking an `Option` here is pre-existing behavior 🤔 I can try to think of a name for it, maybe `from_path_option`? It doesn't necessarily come from cli every time

ma2bd: `from_path_option` works for me. thanks!

linera-service/src/cli/main.rs
ma2bd: No, `TemporaryDirectory` is just a variant holding a tempdir

ndr-ds: Yes. And that tempdir is created using `tempfile::tempdir()`. Which creates it under `/tmp`.
I however want a temporary directory created in a specific directory. In this case it'll need to be the directory where the local ssd is mounted. Which is not `/tmp`. So just using `TemporaryDirectory` won't work.
Unless I modify it to take what directory I want the temporary directory to be created in? But since there's no other use case for that, I thought complicating the API there was unnecessary 🤔

Twey (approved):

afck (approved):

PR #4179, opened 2025-06-27
Return ResourceTracker from `execute_block` method

deuszx:
## Motivation

Messages may be sent with resource grants so that the message receiver doesn't have to pay any fees to receive the message. However, figuring out how much resources to include in the grant is not trivial.

## Proposal

Do that - i.e. return `ResourceTracker` back to the caller of `execute_block`.

As part of that work some refactor of `BlockExecutionTracker` was done. Including removal of `EMPTY_BLOCK_FEE` entirely - that was effectively a no-op.

## Test Plan

CI

## Release Plan

## Links

### DIFF
--- a/linera-chain/src/block_tracker.rs
+++ b/linera-chain/src/block_tracker.rs
@@ -393,6 +393,7 @@ impl<'blobs> BlockExecutionTracker<'blobs> {
             self.events,
             self.blobs,
             self.operation_results,
+            self.resource_controller.tracker,
         )
     }
 }
@@ -403,4 +404,5 @@ pub(crate) type FinalizeExecutionResult = (
     Vec<Vec<Event>>,
     Vec<Vec<Blob>>,
     Vec<OperationResult>,
+    ResourceTracker,
 );
--- a/linera-chain/src/block_tracker.rs
+++ b/linera-chain/src/block_tracker.rs
@@ -115,7 +115,7 @@ impl<'blobs> BlockExecutionTracker<'blobs> {
                     .track_block_size_of(&incoming_bundle)
                     .with_execution_context(chain_execution_context)?;
                 for (message_id, posted_message) in incoming_bundle.messages_and_ids() {
-                    Box::pin(self.execute_message_in_block(
+                    Box::pin(self.execute_message(
                         chain,
                         message_id,
                         posted_message,
@@ -130,8 +130,7 @@ impl<'blobs> BlockExecutionTracker<'blobs> {
                 self.resource_controller_mut()
                     .track_block_size_of(&operation)
                     .with_execution_context(chain_execution_context)?;
-                Box::pin(self.execute_operation_in_block(chain, operation, &mut txn_tracker))
-                    .await?;
+                Box::pin(self.execute_operation(chain, operation, &mut txn_tracker)).await?;
             }
         }
 
@@ -156,7 +155,7 @@ impl<'blobs> BlockExecutionTracker<'blobs> {
     }
 
     /// Executes a message as part of an incoming bundle in a block.
-    async fn execute_message_in_block<C>(
+    async fn execute_message<C>(
         &mut self,
         chain: &mut ExecutionStateView<C>,
         message_id: MessageId,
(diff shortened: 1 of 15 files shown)


linera-chain/src/block_tracker.rs
```diff
@@ -261,6 +229,44 @@ impl<'resources, 'blobs> BlockExecutionTracker<'resources, 'blobs> {
...
+    where
+        C: Context + Clone + Send + Sync + 'static,
+        C::Extra: ExecutionRuntimeContext,
+    {
+        let execution_context = ChainExecutionContext::Operation(txn_tracker.transaction_index());
+        self.resource_controller_mut()
+            .with_state(&mut chain.system)
+            .await?
+            .track_operation(operation)
+            .with_execution_context(execution_context)?;
```
deuszx, line 247: This piece was previously done after we execute the operation but doing it early could detect cases where the operation couldn't have been paid for.

linera-chain/src/block_tracker.rs
```diff
@@ -58,32 +56,28 @@ pub struct BlockExecutionTracker<'resources, 'blobs> {
 
     // Blobs published in the block.
     published_blobs: BTreeMap<BlobId, &'blobs Blob>,
```
afck, line 58: (Do the fields above still need to be public?)

linera-chain/src/block_tracker.rs
```diff
@@ -58,32 +56,28 @@ pub struct BlockExecutionTracker<'resources, 'blobs> {
...
-
-    // We expect the number of outcomes to be equal to the number of transactions in the block.
-    expected_outcomes_count: usize,
 }
 
-impl<'resources, 'blobs> BlockExecutionTracker<'resources, 'blobs> {
+impl<'blobs> BlockExecutionTracker<'blobs> {
     /// Creates a new BlockExecutionTracker.
+    #[expect(clippy::too_many_arguments)]
```
afck, line 63: Maybe `new` should just take a `&BlockHeader` argument, instead of all the individual arguments?

linera-chain/src/block_tracker.rs
afck: (Technically the operation could increase the balance and pay for itself. Not sure if we should allow that.)

linera-chain/src/chain.rs
```diff
@@ -798,18 +799,20 @@ where
             chain.crypto_hash().await?
         };
 
-        let (messages, oracle_responses, events, blobs, operation_results) =
-            block_execution_tracker.finalize();
+        let (messages, oracle_responses, events, blobs, operation_results, resource_tracker) =
+            block_execution_tracker.finalize(block.incoming_bundles.len() + block.operations.len());
 
-        Ok(BlockExecutionOutcome {
+        let execution_outcome = BlockExecutionOutcome {
```
afck, line 805: Maybe `finalize` should return the `BlockExecutionOutcome` instead of its fields.

linera-chain/src/unit_tests/chain_tests.rs
```diff
@@ -677,7 +683,10 @@ async fn test_service_http_response_size_limit(
 
     application.expect_call(ExpectedCall::default_finalize());
 
-    chain.execute_block(&block, time, None, &[], None).await
+    chain
+        .execute_block(&block, time, None, &[], None)
+        .await
+        .map(|(outcome, _resources)| outcome)
```
afck, line 689: Or just change the return type. The test cases only match `Ok(_)` anyway.

(Also in the other tests above, where possible.)

linera-core/src/unit_tests/client_tests.rs
```diff
@@ -1270,21 +1270,34 @@ where
...
-    let sender = builder.add_root_chain(1, Amount::from_tokens(3)).await?;
+    let balance = Amount::from_tokens(3);
+    let sender = builder.add_root_chain(1, balance).await?;
+
+    let recipient = Recipient::Account(Account::new(
+        ChainId(CryptoHash::test_hash("some_chain")),
+        sender.identity().await.unwrap(),
+    ));
 
+    // We can transfer more than we have.
```
afck, line 1287:
```suggestion
    // We cannot transfer more than we have.
```

afck (approved):

linera-chain/src/block_tracker.rs
deuszx: Probably not 👍

ma2bd:
@deuszx What's the status of this PR? (move to draft?)

PR #4180, opened 2025-06-27
End-to-end test of block proposal execution with grant only.

deuszx:
## Motivation

While working on exposing this feature from the `BlockExecutionTracker` I wanted to test whether a grant can cover execution of the message. This test uncovered a bug - we were charging a "base" fee for outgoing messages (and operation execution) which made that test fail. 

## Proposal

This base fee charging was removed (both for operation and a message). After that, we never use that cost/policy entry anywhere anymore so they are removed.

## Test Plan

CI

## Release Plan

## Links

Closes https://github.com/linera-io/linera-protocol/issues/4173

### DIFF
--- a/linera-chain/src/block_tracker.rs
+++ b/linera-chain/src/block_tracker.rs
@@ -413,3 +413,134 @@ pub(crate) type FinalizeExecutionResult = (
     Vec<OperationResult>,
     ResourceTracker,
 );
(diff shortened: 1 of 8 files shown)


ma2bd:
@deuszx Whats the status of this PR? (move to draft?)

Issue #4064, opened 2025-06-04
Remove empty outboxes.


afck:
In https://github.com/linera-io/linera-protocol/pull/3989 I had to remove the code that deletes empty outboxes. We should find a way to do that again, despite the outboxes being at different heights now.

afck:
https://github.com/linera-io/linera-protocol/pull/3989 actually slowed down the `same_chain_native_token_transfers` benchmark by 25%. Removing the empty outboxes should improve this again.

afck:
As discussed, we should at least avoid having to load all the empty outboxes when creating network actions. We should either:
* Delete empty outboxes, and store `next_height_to_schedule` separately, or
* have a separate list of nonempty outboxes.

PR #4181, opened 2025-06-27
Leave gaps in sender chains on the validator side

### LINKED ISSUE #4160: Update validators about sender chains, even if they have gaps.
After https://github.com/linera-io/linera-protocol/pull/4130, if I left gaps in a sender chain and try to make a proposal that receives a message, and a validator doesn't have that message in the inbox, the client just gives up instead of updating the validator about the sender chain.

Instead, the client needs to download the rest of the sender chain, so that it can update the validator about it. This might mean that the updater needs to own a `Client` rather than just a `LocalNode`.

The solution should also address the same issue for blobs: The client may have only the certificate proving a blob itself, without its ancestors. It will have to download the ancestors before it can update the validator.

bart-linera:
## Motivation

If a client creates a block receiving messages from a chain with gaps, and a validator doesn't have that chain yet, the client currently has no way of convincing the validator that the proposal is correct, because the validator expects the whole sender chain as proof of the sent messages.

## Proposal

Make validators accept confirmed certificates for blocks on chains with gaps. If there is a gap, blocks should only be preprocessed, instead of fully processed - like on the clients.

## Test Plan

CI with an added test exercising the problematic scenario

## Release Plan

## Links

- closes #4160

### DIFF
--- a/linera-chain/src/chain.rs
+++ b/linera-chain/src/chain.rs
@@ -426,6 +426,10 @@ where
         }
         if outbox.queue.count() == 0 {
             self.nonempty_outboxes.get_mut().remove(target);
+            // If the outbox is empty and not ahead of the executed blocks, remove it.
+            if *outbox.next_height_to_schedule.get() <= self.tip_state.get().next_block_height {
+                self.outboxes.remove_entry(target)?;
+            }
         }
         #[cfg(with_metrics)]
         metrics::NUM_OUTBOXES
@@ -784,7 +788,7 @@ where
                     .ok_or_else(|| {
                         ChainError::InternalError("missing entry in confirmed_log".into())
                     })?;
-                previous_message_blocks.insert(recipient, hash);
+                previous_message_blocks.insert(recipient, (hash, height));
             }
         }
 
@@ -965,8 +969,7 @@ where
         Ok(())
     }
 
-    /// Returns the hashes of all blocks in the given range. Returns an error if we are missing
-    /// any of those blocks.
+    /// Returns the hashes of all blocks we have in the given range.
     pub async fn block_hashes(
         &self,
         range: impl RangeBounds<BlockHeight>,
@@ -1031,7 +1034,8 @@ where
             if block_height > next_height {
                 // There may be a gap in the chain before this block. We can only add it to this
                 // outbox if the previous message to the same recipient has already been added.
-                let prev_hash = match outbox.next_height_to_schedule.get().try_sub_one().ok() {
+                let maybe_prev_hash = match outbox.next_height_to_schedule.get().try_sub_one().ok()
+                {
                     // The block with the last added message has already been executed; look up its
                     // hash in the confirmed_log.
                     Some(height) if height < next_height => {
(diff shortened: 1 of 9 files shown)


linera-chain/src/chain.rs
```diff
@@ -426,10 +426,6 @@ where
         }
         if outbox.queue.count() == 0 {
             self.nonempty_outboxes.get_mut().remove(target);
-            // If the outbox is empty and not ahead of the executed blocks, remove it.
-            if *outbox.next_height_to_schedule.get() <= self.tip_state.get().next_block_height {
-                self.outboxes.remove_entry(target)?;
-            }
```
afck: Then let's reopen https://github.com/linera-io/linera-protocol/issues/4064 and see if we can find another solution.

linera-chain/src/chain.rs
```diff
@@ -990,14 +986,9 @@ where
...
-                self.preprocessed_blocks
-                    .get(&BlockHeight(height))
-                    .await?
-                    .ok_or_else(|| {
-                        ChainError::InternalError("missing entry in preprocessed_blocks".into())
-                    })?,
-            );
+            if let Some(hash) = self.preprocessed_blocks.get(&BlockHeight(height)).await? {
+                hashes.push(hash);
+            }
```
afck, line 994: The method's doc comment is now outdated.

linera-chain/src/chain.rs
bart-linera: As discussed on Slack, I implemented a height-based check in `process_outgoing_messages`.

linera-core/src/chain_worker/state/attempted_changes.rs
```diff
@@ -325,6 +321,15 @@ where
             check_block_epoch(epoch, chain_id, block.header.epoch)?;
             certificate.check(committee)?;
         }
+
+        if tip.next_block_height < height {
```
deuszx, line 327: It'd be nice to have some short comment here explaining why we check and what we do.

deuszx (approved):
LGTM

linera-core/src/chain_worker/state/attempted_changes.rs
bart-linera: Added in 9046618

linera-core/src/chain_worker/state/attempted_changes.rs
```diff
@@ -290,11 +290,6 @@ where
 
         // Check that the chain is active and ready for this confirmation.
         let tip = self.state.chain.tip_state.get().clone();
-        if tip.next_block_height < height {
```
ma2bd, line 293: @bart-linera  Do we need to modify `handle_block_proposal` to require that all gaps are filled before a new block is proposed?

bart-linera: I think we don't have to change anything - handling a proposal involves checking that the proposal's height equals `tip.next_block_height`, and that is only updated when a block is fully processed, not pre-processed. So accepting a proposal should already fail if the gaps aren't filled.

deuszx: Can we add a test just to make sure (if there's none yet)? The test you added here looks fairly clean, extending it shouldn't be difficult.

bart-linera: I can try adding an attempt to propose a block on the sender chain, that should fail with the validator having gaps... unless the client will automatically attempt to fill them in :thinking: But I think if I don't sync the chain explicitly, that shouldn't happen. I'll try and see.  :+1:

bart-linera: Hm, no, actually submitting a proposal involves sending all potentially missing info :grimacing: And I see in the logs that this is what happened: the client actually updated the validator with confirmed certificates for the block in the gap and the preprocessed block. But I don't think I'll be able to test for that in an end-to-end test... this might have to be done in a worker test.

deuszx: Thanks for adding the test. It looks good.

bart-linera: Added a worker test in cfc9fe5

PR #4184, opened 2025-06-27
Replace active_epochs by min and max active epoch

bart-linera:
## Motivation

`ChainDescription`s contain a set of active epochs. Since we expect the set of active epochs to be contiguous, we can optimize this further.

## Proposal

Replace the set of active epochs by the minimum and maximum active epoch.

## Test Plan

CI

## Release Plan

## Links

### DIFF
--- a/linera-execution/src/system.rs
+++ b/linera-execution/src/system.rs
@@ -781,7 +781,7 @@ where
         let committees = self
             .context()
             .extra()
-            .committees_for(min_active_epoch, max_active_epoch)
+            .committees_for(min_active_epoch..=max_active_epoch)
             .await?;
         self.committees.set(committees);
         let admin_id = self
--- a/linera-storage/src/lib.rs
+++ b/linera-storage/src/lib.rs
@@ -7,7 +7,7 @@
 
 mod db_storage;
 
-use std::{collections::BTreeMap, sync::Arc};
+use std::{collections::BTreeMap, ops::RangeInclusive, sync::Arc};
 
 use async_trait::async_trait;
 use dashmap::DashMap;
@@ -174,9 +174,14 @@ pub trait Storage: Sized {
     /// Returns a map of the committees for the given epochs.
     async fn committees_for(
         &self,
-        min_epoch: Epoch,
-        max_epoch: Epoch,
+        epoch_range: RangeInclusive<Epoch>,
     ) -> Result<BTreeMap<Epoch, Committee>, ViewError> {
+        // Short-circuit for an empty input range.
+        if epoch_range.is_empty() {
+            return Ok(BTreeMap::new());
+        }
+        let min_epoch = epoch_range.start();
+        let max_epoch = epoch_range.end();
         let read_committee = async |committee_hash| -> Result<Committee, ViewError> {
             let blob_id = BlobId::new(committee_hash, BlobType::Committee);
             let committee_blob = self
@@ -193,7 +198,7 @@ pub trait Storage: Sized {
         let admin_chain_id = network_description.admin_chain_id;
         let mut result = BTreeMap::new();
         // special case: the genesis epoch is stored in the NetworkDescription
-        if min_epoch == Epoch::ZERO {
+        if *min_epoch == Epoch::ZERO {
             let genesis_committee =
                 read_committee(network_description.genesis_committee_blob_hash).await?;
             result.insert(Epoch::ZERO, genesis_committee);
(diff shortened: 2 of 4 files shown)


linera-execution/src/system.rs
```diff
@@ -824,7 +831,18 @@ where
...
+                .keys()
+                .min()
+                .copied()
+                .unwrap_or(Epoch::ZERO),
+            self.committees
+                .get()
+                .keys()
+                .max()
+                .copied()
+                .unwrap_or(Epoch::ZERO),
```
deuszx, line 845: Previous version could return an empty set while new one will always return at least `Epoch::ZERO`.

That's probably fine in practice though.

linera-storage/src/lib.rs
```diff
@@ -177,7 +174,8 @@ pub trait Storage: Sized {
     /// Returns a map of the committees for the given epochs.
     async fn committees_for(
         &self,
-        mut epochs: BTreeSet<Epoch>,
+        min_epoch: Epoch,
+        max_epoch: Epoch,
```
deuszx: There's a degenerative case when we pass in (by mistake) a `min_epoch > max_epoch`. Maybe it's worth panicing in such case?

linera-execution/src/system.rs
bart-linera: Yes, I thought about making the upper bound exclusive, but I decided that it makes no sense: there should always be at least one active epoch. If there isn't, we have bigger problems :sweat_smile:

linera-storage/src/lib.rs
```diff
@@ -195,13 +193,13 @@ pub trait Storage: Sized {
...
         // special case: the genesis epoch is stored in the NetworkDescription
-        if epochs.remove(&Epoch::ZERO) {
+        if min_epoch == Epoch::ZERO {
             let genesis_committee =
                 read_committee(network_description.genesis_committee_blob_hash).await?;
             result.insert(Epoch::ZERO, genesis_committee);
         }
 
-        let start_index = epochs.first().unwrap_or(&Epoch::ZERO).0;
+        let start_index = min_epoch.0.max(1);
```
deuszx, line 207: Looks like the old version had a small (but harmless) bug – we removed `Epoch::ZERO` from `epochs` and yet we would `unwrap_or(Epoch::ZERO)` here. The new version starts at at least `Epoch(1)`.

linera-storage/src/lib.rs
```diff
@@ -210,34 +208,22 @@ pub trait Storage: Sized {
...
-
-        for (epoch, committee) in epochs_and_committees {
-            epochs.remove(&epoch);
-            result.insert(epoch, committee);
-        }
+        result.extend(
+            futures::future::try_join_all(
+                epoch_creation_events
+                    .into_iter()
+                    .take_while(|index_and_event| index_and_event.index <= max_epoch.0)
```
deuszx, line 220: I am always unsure whether the order of returning from the collection of futures is in the same order in which they were created 🤔 otherwise we might stop processing here prematurely (i.e. some future that resolves later on should still be returned).

ma2bd:
> Since we expect the set of active epochs to be contiguous, we can optimize this further.

That's actually an interesting question whether we should support "yanking" specific epochs but I agree that we should probably keep it simple. (@afck wdyt?)

linera-storage/src/lib.rs
ma2bd: I believe this iterator creates the futures

deuszx: Ah, of course 🤦 I got confused by `read_events_from_index` which is an async function that returns a `Vec<_>` (not a `Vec<impl Future<...>>`).

linera-storage/src/lib.rs
ma2bd: I was wondering about this. Perhaps we could use `epoch_range: Range<Epoch>`.

afck:
I like the rule that the active epochs are contiguous. :+1:

linera-storage/src/lib.rs
bart-linera: I guess we could use ranges, but you can still construct a range with an upper bound less than lower bound - only it's an empty range in such a case. I guess here we could treat it similarly: just return an empty map if max is less than min?

ma2bd: I guess the debate is more about a struct with two named fields vs two epoch parameters (that could be swapped by mistake).

ma2bd (approved):

Issue #4178, opened 2025-06-26
`MissingEntries` in `QueueView`


afck:
Commit https://github.com/linera-io/linera-protocol/commit/a86687bc936254c764348167fdd34c165f0b13f8 in https://github.com/linera-io/linera-protocol/pull/4150 revealed an issue that causes a `MissingEntries` error in `QueueView`.

## To Reproduce:

```
git checkout a86687bc936254c764348167fdd34c165f0b13f8
cargo build --release -p linera-storage-service -p linera-service
cargo run --release --bin linera -- net up --validators 1 --shards 1 --with-faucet
```
Then, in another tab, set the wallet env variables to empty directories (or delete `~/.config/linera`) and:
```
cargo run --release --bin linera --features benchmark -- wallet init --faucet http://localhost:8080
cargo run --release --bin linera --features benchmark -- wallet request-chain --faucet http://localhost:8080
```
You can then run the benchmarks as often as needed:
```
cargo run --release --bin linera --features benchmark -- --max-pending-message-bundles 300 benchmark --bps 150 --transactions-per-block 300 --close-chains
```

## Expected Outcome

The benchmark should run without errors. (Depending on the hardware it might not reach the configured BPS.)

## Actual Outcome

There are several `missing database entries` errors.

afck:
## Where things go wrong in the queue view

I created a [branch with added logging](https://github.com/afck/linera-protocol/tree/debug-missing-entries). As far as I can tell:
- During `handle_confirmed_certificate`:
  - `QueueView` is loaded with `stored_indices = 0..300`. (LRU cache `0..300`, storage `0..300`. ✔️)
  - `QueueView::flush` gets called; adds `stored_indices = 0..0` to the batch. Should overwrite the `0..300`!
  - `QueueView` gets dropped (`stored_indices = 0..0`).
- 0.1s later, during `handle_cross_chain_request`, `QueueView` is loaded with `stored_indices = 0..300`. LRU cache `0..300` ❌, storage `0..0` ✔️: LRU cache value is wrong even though storage is correct!

## Some changes make the error disappear:

- Using storage service (`LINERA_STORAGE=service:tcp:127.0.0.1:1235:client`) instead of RocksDB for the wallet.
- Running in `debug` mode instead of `release` mode.
- Adding too many logs or assertions.
- Disabling the LRU cache: `--storage-max-cache-size 0`
[...]

PR #2330, opened 2024-07-30
Remove chain guards

jvff:
## Motivation

Chain guards were originally used to synchronize access to chain states in storage. With the chain workers (#2135), that is no longer needed.

However, without the chain guards, there's a future risk of concurrent chain access, which could in the worst case lead to data corruption.

## Proposal

Remove the chain guards, and add a CI check for chain load operations to mitigate the future risk of code changes introducing data races.

## Test Plan

Ran the CI script manually, and checked that it succeeds if everything is as expected and fails if there are any new unexpected chain load operations.

## Release Plan

Nothing needed, because this is an internal refactor/cleanup.

## Links

- [reviewer checklist](https://github.com/linera-io/linera-protocol/blob/main/CONTRIBUTING.md#reviewer-checklist)

PR #4186, opened 2025-06-28
Reintroduce the `chain_guard` in Linera.

MathieuDutSik:
## Motivation

It has recently appeared that some bug are occurring in Linera. This PR addresses the bug, but it is not a definitive solution.

## Proposal

The bug is described in https://github.com/linera-io/linera-protocol/issues/4178

Further analysis revealed the following:
* There were two threads working on the same chain. This is visible not just from the logs, but also from having a Mutex based database in the code that keeps track of what is opened.
* By reintroducing the ChainGuards that were removed in https://github.com/linera-io/linera-protocol/pull/2330 we make the bug disappear.

## Test Plan

The CI.
The bug is not visible in the current CI. The effect can be visible by following the instructions in the GitHub Issue 4178.

## Release Plan

It is a little unclear how the solution should be implemented.

## Links

None

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5668,6 +5668,7 @@ dependencies = [
  "linera-views",
  "prometheus",
  "serde",
+ "tokio",
 ]
 
 [[package]]
--- a/examples/Cargo.lock
+++ b/examples/Cargo.lock
@@ -3893,6 +3893,7 @@ dependencies = [
  "linera-views",
  "prometheus",
  "serde",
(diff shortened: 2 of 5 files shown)


MathieuDutSik:
This reimplementation of the `ChainGuards` is a reversion that could have been adequate for the purpose of avoiding the bug.

However, it introduces slowdowns that make the CI fail. This is reason enough for refusing to consider it to be the right path.

PR #4193, opened 2025-07-01
Use NodeConfig to setup Scylla

ndr-ds:
## Motivation

There's this `NodeConfig` that we can use to setup the disks properly for Scylla, and do a few other optimizations to the setup.

## Proposal

Use the `NodeConfig` for Scylla, but continue using the existing hook script to setup local SSD for the shards, when using `DualStore`. Make sure the Scylla stuff only runs in the Scylla VMs by using placements.

## Test Plan

Deployed a network with these changes both using `DualStore` and using just `ScyllaDB`, both work.

## Release Plan

### DIFF
--- a/kubernetes/linera-validator/helmfile.yaml
+++ b/kubernetes/linera-validator/helmfile.yaml
@@ -3,10 +3,10 @@ environments:
     values:
       - writeToGrafanaCloud: {{ env "LINERA_WRITE_TO_GRAFANA_CLOUD" | default "false" }}
         validatorLabel: {{ env "LINERA_VALIDATOR_LABEL" | default (printf "local-%s" (env "USER")) }}
-        usingLocalSsd: {{ env "LINERA_HELMFILE_SET_USING_LOCAL_SSD" | default "false" }}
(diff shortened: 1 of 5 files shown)


ndr-ds:
* **#4193** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4193?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4193?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

eldios:
side comment:
don't particularly like the use of RAID0 as it's as brittle as it can get, but it's also true that we're a VM so theoretically the disks we use are a virtual abstraction that should already have some level of redundancy below.

Also we getting more and more entrenched/intertwined with Google Cloud, not sure if that's good/intentional or not.

Don't have enough knowledge about our infra to say if the two above are acceptable trade-offs or not, so will let other review and chime in as well.

eldios (approved):
approve with caveat about RAID0 but it's a requirement from ScyllaDB

PR #4194, opened 2025-07-02
Test the generation of the code in the CI.

MathieuDutSik:
## Motivation

The file `LineraTypes.sol` is generated from the `LineraTypes.yaml` but this is not tested in the CI.

## Proposal

The `serdegen` is installed by using `cargo`. Then the file is generated and then tested for equality.

## Test Plan

The CI is precisely the point.

## Release Plan

## Links

None.


ma2bd (approved):
LGTM

PR #4195, opened 2025-07-02
Unify confirmed certificate processing between clients and validators

bart-linera:
## Motivation

Now that validators allow for gaps in chains, we can unify the way confirmed certificate are processed on the client and on the validator.

## Proposal

Remove the `PreprocessCertificate` worker request and use `ProcessConfirmedBlock` instead (which can still call `preprocess_certificate` under the hood if there are gaps).

Moreover, always use the committee from storage for certificate validation. We used to get the current committee from the chain state for nonzero-height blocks, but this can be incorrect if there are gaps, because the committee might have changed within the gap, or we might not even have a committee in the chain state. Thus, we should always use the block epoch to get the committee from storage, but if the block is going to be fully processed, we should also verify that the epoch in it matches the current epoch on the chain.

## Test Plan

CI

## Release Plan

## Links

### DIFF
--- a/linera-core/src/chain_worker/state/attempted_changes.rs
+++ b/linera-core/src/chain_worker/state/attempted_changes.rs
@@ -449,7 +449,7 @@ where
 
     /// Stores a block's blobs, and adds its messages to the outbox where possible.
     /// Does not execute the block.
-    pub(super) async fn preprocess_certificate(
+    async fn preprocess_certificate(
         &mut self,
         certificate: ConfirmedBlockCertificate,
     ) -> Result<NetworkActions, WorkerError> {
--- a/linera-core/src/chain_worker/state/attempted_changes.rs
+++ b/linera-core/src/chain_worker/state/attempted_changes.rs
@@ -300,27 +300,17 @@ where
         }
 
         // We haven't processed the block - verify the certificate first
-        if height == BlockHeight::ZERO {
-            // For the initial proposal on a chain, we read the committee directly from
-            // storage. If the certificate is good for this committee, we will accept it -
-            // which will later allow us to also accept the chain description blob.
-            let epoch = block.header.epoch;
-            let committees = self
-                .state
-                .storage
-                .committees_for([epoch].into_iter().collect())
-                .await?;
-            let committee = committees
-                .get(&epoch)
-                .ok_or(WorkerError::UnknownEpoch { chain_id, epoch })?;
-            certificate.check(committee)?;
-        } else {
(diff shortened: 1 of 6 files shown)


linera-core/src/chain_worker/state/attempted_changes.rs
```diff
@@ -300,27 +300,17 @@ where
...
+        let epoch = block.header.epoch;
+        // Get the committee for the block's epoch from storage.
+        let committees = self
+            .state
+            .storage
+            .committees_for([epoch].into_iter().collect())
+            .await?;
+        let committee = committees
+            .get(&epoch)
+            .ok_or(WorkerError::UnknownEpoch { chain_id, epoch })?;
```
deuszx, line 308: Can it happen that this error is returned when we try to preprocess a block, on a chain with gaps, and the committee changed in that gap? if so, the response of the client to this error should be that it tries to fill in the gaps of the validator - right? I don't see it happening right now.

bart-linera: As long as the validator (or client) has the correct committee from the admin chain, it shouldn't. If the committee changed within the gap, the block will have the new epoch in the header, and so this code will look up the new epoch. This was an issue _before_ this change, because then the code would look up the "current" epoch on the chain (which would be the pre-gap epoch), and it could be a different epoch than the post-gap one.

linera-core/src/chain_worker/state/attempted_changes.rs
```diff
@@ -363,6 +356,14 @@ where
             .storage
             .maybe_write_blob_states(&blob_ids, blob_state)
             .await?;
```
afck, line 342: Everything up to here could probably also be deduplicated.

I think if we check `tip.next_block_height < height` in just two places (or even just one?), we can remove `preprocess_certificate` entirely.

bart-linera: Good point!
I deduplicated a bit more, per your suggestion. There is still _some_ duplicated code, which could technically be deduplicated by wrapping a big part in `if tip.next_block_height == height` instead of returning early, but to me personally this is slightly cleaner than the alternative.

afck (approved):

linera-core/src/chain_worker/state/attempted_changes.rs
deuszx: It wasn't only that. Here, if we miss the certificate, we return a generic `ViewError` to the client which I don't think is automatically recovered?

PR #4201, opened 2025-07-04
chore: fix reference 

dizer-ti:
Hey devs, fixed reference

https://github.com/linera-io/linera-protocol/linera-web/signer/#readme -> dead reference 
https://github.com/linera-io/linera-protocol/tree/main/linera-web/signer -> new reference


Twey:
Thanks!

PR #4202, opened 2025-07-04
Remove dependency to genawaiter

ma2bd:
## Motivation

* `genawaiter` is not maintained
* Use idiomatic Rust code

## Proposal

* Use the good old `Write` trait.
* Use a trait `FileContentGenerator` with a generic method

## Test Plan

CI

## Release Plan

    - be released in a new SDK,
    - be released in a new SDK,
      SDK soon.

### DIFF
--- a/linera-execution/src/bin/wit_generator.rs
+++ b/linera-execution/src/bin/wit_generator.rs
@@ -12,13 +12,15 @@ use std::{
     path::{Path, PathBuf},
 };
 
-use anyhow::{ensure, Context, Result};
+use anyhow::{Context, Result};
 use clap::Parser as _;
 use linera_execution::{
     BaseRuntimeApi, ContractEntrypoints, ContractRuntimeApi, ContractSyncRuntimeHandle,
     RuntimeApiData, ServiceEntrypoints, ServiceRuntimeApi, ServiceSyncRuntimeHandle,
 };
-use linera_witty::wit_generation::{StubInstance, WitInterfaceWriter, WitWorldWriter};
+use linera_witty::wit_generation::{
+    FileContentGenerator, StubInstance, WitInterfaceWriter, WitWorldWriter,
+};
 
 /// Command line parameters for the WIT generator.
 #[derive(Debug, clap::Parser)]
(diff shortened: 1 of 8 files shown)


linera-execution/src/bin/wit_generator.rs
```diff
@@ -142,34 +128,51 @@ impl Operation for WriteToFile {
...
+            file: BufReader::new(file),
+        }
+    }
+}
+
+impl Write for FileComparer {
+    fn write(&mut self, part: &[u8]) -> std::io::Result<usize> {
+        self.buffer.resize(part.len(), 0);
+        self.file.read_exact(&mut self.buffer)?;
+        if self.buffer != part {
```
deuszx, line 149: I wonder if we could give better diffs here. I known the argument type is byte slice but in practice we pass the full file content (or am I wrong) so we should be able to compare human-readable content.

linera-execution/src/bin/wit_generator.rs
```diff
@@ -142,34 +128,51 @@ impl Operation for WriteToFile {
...
-        &mut self,
-        path: &Path,
-        contents: impl Iterator<Item = &'c str>,
-    ) -> Result<()> {
-        let mut buffer = Vec::new();
-        let mut file = BufReader::new(
-            File::open(path)
-                .with_context(|| format!("Failed to open file at {}", path.display()))?,
-        );
+struct FileComparer {
```
deuszx: Nit: rename to `FileComparator`?

deuszx (approved):

linera-execution/src/bin/wit_generator.rs
ma2bd: Yes that's fair. "Comparer" is not even a word apparently.

linera-execution/src/bin/wit_generator.rs
ma2bd: I honestly don't completely understand why we need this "check" mode in the first place. For CI, we can always overwrite files and run `git diff`.

deuszx: Fair enough. I thought we could be interested in _what_ exactly changed rather than blindly overwrite things.

Twey: This applies to a lot of our CI checks and is something I've been thinking about doing for other reasons (e.g. documentation).

PR #4204, opened 2025-07-08
Implement ERC20 in a contract that also implement transfers

### LINKED ISSUE #3931: Support ERC20 on Linera and extend Fungible Linera contracts.
In the Ethereum world, ERC-20 is a standard with a specific API. This API has several features that are absent from the `Fungible` application in Linera.

There is a reference implementation [Open Zeppelin ERC-20](https://docs.openzeppelin.com/contracts/4.x/erc20) which is extremely widely used and therefore it makes sense to take it as a starting point for Linera. 

Notes:
* This basic ERC-20 implementation could be extended by adding multichain capabilities to it. The extended version could be called `ERC20Multichain`.
* The `totalSupply` function is constant for ERC-20 basic OpenZeppelin (there are variants `ERC20Burnable` / `ERC20Mintable`). For a true multichain, it is almost impossible to have a function `totalSupply` that queries all chains in which a token could be instantiated.
* An application satisfying the ERC-20 could be used by DeFi EVM applications running on Linera. If the multichain effort is done by `ERC20Multichain` then this could lighten the efforts of Ethereum developers moving to Linera.

The ERC-20 API:
```solidity
interface IERC20 {
    // Returns the total number of tokens in existence
    function totalSupply() external view returns (uint256);

    // Returns the amount of tokens owned by `account`
    function balanceOf(address account) external view returns (uint256);

    // Transfers `amount` tokens to `recipient`
    function transfer(address recipient, uint256 amount) external returns (bool);

    // Returns the remaining number of tokens that `spender` is allowed to spend on behalf of `owner`
    function allowance(address owner, address spender) external view returns (uint256);

    // Sets `amount` as the allowance of `spender` over the caller’s tokens
    function approve(address spender, uint256 amount) external returns (bool);

    // Moves `amount` tokens from `sender` to `recipient` using the allowance mechanism
    function transferFrom(address sender, address recipient, uint256 amount) external returns (bool);

    // Emitted when `value` tokens are moved from one account (`from`) to another (`to`)
    event Transfer(address indexed from, address indexed to, uint256 value);

    // Emitted when the allowance of a `spender` for an `owner` is set by a call to `approve`
    event Approval(address indexed owner, address indexed spender, uint256 value);
}
```

MathieuDutSik:
## Motivation

The ERC20 is a major fungible token smart contract in the Ethereum world.
They form a basis of Decentralized Finance applications such as Morpho, Uniswap, or others.

Fixes #3931

## Proposal

The reference implementation [OpenZeppelin ERC20](https://docs.openzeppelin.com/contracts/4.x/erc20) is the one that is used in most cases. So, we take it as the basis for our implementation.

In our implementation, we have a constant value. There is no minting or burning. That implements the `totalSupply` very easily: We simply return the value at the constructor level.

The transfer from one chain to another is done with one `_burn` (in the sending chain) and one `_mint` (in the receiving chain).

However, since we have a multi-chain, and we take the total chain, we have to change the code so that the `totalSupply` would work. One design that would not have worked is to have a child contract since the events are emitted with the contract address and the `msg.sender` is used to identify the sender.
 
The ERC20 contract depends on the `OpenZeppelin`, which is published via`npm`. Therefore, we have added this to the compilation in `solidity.rs`

## Test Plan

The CI tests the functionalities.

## Release Plan

## Links

None.

### DIFF
--- a/linera-service/tests/fixtures/erc20_shared.sol
+++ b/linera-service/tests/fixtures/erc20_shared.sol
@@ -72,16 +72,22 @@ contract ERC20_shared is Context, IERC20, IERC20Errors {
      * the user with the tokens.
      */
     function execute_message(bytes memory input) external {
-        (address user, uint256 amount) = abi.decode(input, (address, uint256));
-        _mint(user, amount);
+        (address source, address destination, uint256 amount) = abi.decode(input, (address, address, uint256));
+        LineraTypes.OptionBool result = Linera.message_is_bouncing().value;
+        require(result != LineraTypes.OptionBool.None);
+        if (result == LineraTypes.OptionBool.True) {
+            _mint(source, amount);
+        } else {
+            _mint(destination, amount);
+        }
     }
 
     /**
      * This is the transfer to another chain under the same username.
      */
-    function transferToChain(bytes32 chain_id, uint256 amount) external {
+    function transferToChain(bytes32 chain_id, address destination, uint256 amount) external {
         _burn(msg.sender, amount);
-        bytes memory message = abi.encode(msg.sender, amount);
+        bytes memory message = abi.encode(msg.sender, destination, amount);
         Linera.send_message(chain_id, message);
     }
 
--- a/linera-service/tests/linera_net_tests.rs
+++ b/linera-service/tests/linera_net_tests.rs
@@ -1442,7 +1442,7 @@ async fn test_evm_erc20_shared(config: impl LineraNetConfig) -> Result<()> {
         function totalSupply();
         function transfer(address to, uint256 value);
         function balanceOf(address account);
-        function transferToChain(bytes32 chain_id, uint256 value);
+        function transferToChain(bytes32 chain_id, address destination, uint256 value);
     }
 
     let the_supply = U256::from(1000000000);
@@ -1514,6 +1514,7 @@ async fn test_evm_erc20_shared(config: impl LineraNetConfig) -> Result<()> {
     let chain_id: B256 = chain_id.into();
     let mutation = transferToChainCall {
         chain_id,
+        destination: address2,
         value: transfer2,
     };
     let mutation = EvmQuery::Mutation(mutation.abi_encode());
@@ -1531,6 +1532,8 @@ async fn test_evm_erc20_shared(config: impl LineraNetConfig) -> Result<()> {
         the_supply - transfer1 - transfer2
     );
 
+    let query = balanceOfCall { account: address2 };
+    let query = EvmQuery::Query(query.abi_encode());
     let result = application2.run_json_query(query).await?;
     assert_eq!(read_evm_u256_entry(result), transfer2);
 


linera-base/src/identifiers.rs
```diff
@@ -59,6 +59,15 @@ impl AccountOwner {
             AccountOwner::Address20(_) => 20,
         }
     }
+
+    /// Gets the EVM address if possible
+    #[cfg(with_revm)]
+    pub fn to_address(&self) -> Option<Address> {
```
ma2bd: nit: `to_evm_address` is more future-proof as we may rename `AccountOwner` in the future.

linera-service/tests/linera_net_tests.rs
```diff
@@ -1408,6 +1408,142 @@ async fn test_wasm_end_to_end_counter(config: impl LineraNetConfig) -> Result<()
...
+    let query = EvmQuery::Query(query.abi_encode());
+    let result = application1.run_json_query(query).await?;
+    assert_eq!(read_evm_u256_entry(result), the_supply - transfer1);
+
+    let query = balanceOfCall { account: address2 };
+    let query = EvmQuery::Query(query.abi_encode());
+    let result = application1.run_json_query(query).await?;
+    assert_eq!(read_evm_u256_entry(result), transfer1);
+
+    // Transfering to another chain and checking the balances.
```
ma2bd: "transferring" ?

linera-base/src/identifiers.rs
MathieuDutSik: Yes, that is a better name.

linera-service/tests/linera_net_tests.rs
MathieuDutSik: You are correct. Done.

deuszx:
> In our implementation, we have a constant value. There is no minting or burning. That implements the totalSupply very easily: We simply return the value at the constructor level.

I don't understand these decisions, can you explain?

linera-service/tests/fixtures/erc20_shared.sol
```diff
@@ -0,0 +1,291 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts (last updated v5.3.0) (token/ERC20/ERC20.sol)
```
deuszx: If this contract is modified OZ contract we should note that in the header - also what was modified. Right now it looks as if it's taken _as-is_ without any modifications but that's not the case.

I'd suggest putting a long introduction and explanation here, written for Solidity developers, explaining the differences.

linera-service/tests/fixtures/erc20_shared.sol
```diff
@@ -0,0 +1,291 @@
...
+    function instantiate(bytes memory input) external {
+        uint256 input_value = abi.decode(input, (uint256));
+        require(input_value == summed_total_supply);
+        _mint(msg.sender, input_value);
+    }
+
+    function execute_message(bytes memory input) external {
+        (address user, uint256 amount) = abi.decode(input, (address, uint256));
+        _mint(user, amount);
+    }
```
deuszx: I don't understand this - how can you assume that all `execute_message` calls are for `_mint`-ing

linera-service/tests/fixtures/erc20_shared.sol
```diff
@@ -0,0 +1,291 @@
...
+ * conventional and does not conflict with the expectations of ERC-20
+ * applications.
+ */
+contract ERC20_shared is Context, IERC20, IERC20Errors {
+    mapping(address account => uint256) private _balances;
+
+    mapping(address account => mapping(address spender => uint256)) private _allowances;
+
+    uint256 private summed_total_supply; // total supply of all the chains.
+    uint256 private _totalSupply; // total supply of this chain
```
deuszx, line 51: What about name and ticker fields?

linera-service/tests/fixtures/erc20_shared.sol
```diff
@@ -0,0 +1,291 @@
...
+        require(input_value == summed_total_supply);
+        _mint(msg.sender, input_value);
+    }
+
+    function execute_message(bytes memory input) external {
+        (address user, uint256 amount) = abi.decode(input, (address, uint256));
+        _mint(user, amount);
+    }
+
+    function transferToChain(bytes32 chain_id, uint256 amount) external {
```
deuszx: IIUC we want to move away from chains holding a balance, in which case this should transfer to an account on chain.

linera-service/tests/fixtures/erc20_shared.sol
deuszx: OK, I can see that `transferToChain` sends a message but that mesage is sent to a `chain`, not user, while here you're using `user` as the recipient of the transfer. So what happens if you transfer to a different chain and that chain becomes the recipient of the tokens (here `user`) - how is the _chain_ now supposed to sign a transaction that moves the tokens?

linera-service/tests/linera_net_tests.rs
```diff
@@ -1408,6 +1408,142 @@ async fn test_wasm_end_to_end_counter(config: impl LineraNetConfig) -> Result<()
...
+
+    // Checking the balances on both chains.
+
+    let query = balanceOfCall { account: address1 };
+    let query = EvmQuery::Query(query.abi_encode());
+    let result = application1.run_json_query(query.clone()).await?;
+    assert_eq!(
+        read_evm_u256_entry(result),
+        the_supply - transfer1 - transfer2
+    );
```
deuszx, line 1533: Can you show how we can transfer the tokens sent to a (different) `chain` as recipient?

MathieuDutSik:
> > In our implementation, we have a constant value. There is no minting or burning. That implements the totalSupply very easily: We simply return the value at the constructor level.
> 
> I don't understand these decisions, can you explain?

First of all if we implement minting and burning, we have to make some decisions about how this is done.  This is non-trivial and that is not the purpose of this implementation.

Secondly, the implementation of tokens whose supply is varying are more complicated than the other ones. This is because the `total_supply` function is a global function. And so we would need to have events in order to keep track of mints/burns.

linera-service/tests/fixtures/erc20_shared.sol
MathieuDutSik: Ok, done.

linera-service/tests/fixtures/erc20_shared.sol
MathieuDutSik: The chain is not the same as the user, so I am not sure what you mean.

linera-service/tests/fixtures/erc20_shared.sol
MathieuDutSik: I did not saw them as important, they are not part of the ERC20 standard, just some extension and so I removed them.
I can add them back if needed.

linera-service/tests/linera_net_tests.rs
MathieuDutSik: This is what is being done here, so I am not sure what you mean.

linera-service/tests/fixtures/erc20_shared.sol
MathieuDutSik: The tokens are transferred to the other chain under the same user name.

deuszx: Are you assuming that this contract lives on a user chain and only chain owner can transfer the tokens? If so, I think that's wrong. There will be chains where tokens belong to accounts other than chain owners and those accounts should also be allowed to transfer tokens.

linera-service/tests/fixtures/erc20_shared.sol
deuszx: This is related to the convo on `function transferToChain`.

linera-service/tests/fixtures/erc20_shared.sol
```diff
@@ -0,0 +1,317 @@
...
+            }
+        } else {
+            unchecked {
+                // Overflow not possible: balance + value is at most totalSupply, which we know fits into a uint256.
+                _balances[to] += value;
+            }
+        }
+
+        emit Transfer(from, to, value);
+    }
```
deuszx, line 222: Imagine the case where:
- I transfer tokens with cross-chain transfer to a new chain (`chainA`)
- I want to now _move_ (transfer) tokens on `chainA` using some account (let's call it `accountC`)

Where is the check that verifies that `accountC` is allowed to move these tokens? The tokens aren't stored under `_balances[accountC]`, they're under `_balances[chainA]`.

linera-service/tests/linera_net_tests.rs
deuszx: Commented on [here](https://github.com/linera-io/linera-protocol/pull/4204#discussion_r2201467389)

linera-service/tests/fixtures/erc20_shared.sol
deuszx: I think this is all connected (see also [this](https://github.com/linera-io/linera-protocol/pull/4204#discussion_r2201461386))

linera-execution/src/test_utils/solidity.rs
```diff
@@ -91,6 +98,16 @@ pub fn get_bytecode(source_code: &str, contract_name: &str) -> anyhow::Result<Ve
             writeln!(test_code_file, "{}", literal_path)?;
         }
     }
+    if source_code.contains("@openzeppelin") {
+        let _output = Command::new("npm")
```
ma2bd, line 102: `let _output =` is not needed. Also, there must be a better method than `Command::output` if you don't need the output.

linera-service/tests/fixtures/erc20_shared.sol
```diff
@@ -0,0 +1,317 @@
...
+// It is not possible to nest an ERC20 contract into another ERC20
+// contract and so a modification of the code could not be avoided.
+//
+// This implementation has been adapted to our need in the following
+// ways:
+// * The totalSupply() of this contract is constant and fixed at
+//   construction to be the total one. It returns the sum of the supply
+//   over all the chains.
+// * The localSupply() has been introduced to return the supply of one
+//   particular chain.
```
ma2bd, line 14: Do we need this? I would not add anything unless provably useful.

linera-service/tests/fixtures/erc20_shared.sol
ma2bd: The "total supply" of a single chain is most likely useless.

linera-service/tests/fixtures/erc20_shared.sol
ma2bd: @deuszx Are you familiar with the fact that cross-chain messages are internal to an application?

deuszx: ~~Yes, I don't see how it's relevant here?~~

I see now, I misread how the message is encoded.

linera-service/tests/fixtures/erc20_shared.sol
deuszx: Indeed, I forgot that ticker and symbol are part of the metadata interface.

linera-service/tests/fixtures/erc20_shared.sol
deuszx: Nevermind, I misread the code.

linera-service/tests/fixtures/erc20_shared.sol
```diff
@@ -0,0 +1,317 @@
...
+        (address user, uint256 amount) = abi.decode(input, (address, uint256));
+        _mint(user, amount);
+    }
+
+    /**
+     * This is the transfer to another chain under the same username.
+     */
+    function transferToChain(bytes32 chain_id, uint256 amount) external {
+        _burn(msg.sender, amount);
+        bytes memory message = abi.encode(msg.sender, amount);
```
deuszx: Why not support sending to a different chain but also a different account on the recipient chain? I understand we can simulate that by sending to the `msg.sender` on a different chain and then moving tokens to a new user within that chain but that's two transactions and the second one being on a chain where we might not be an owner even

deuszx (approved):
Sorry for the confusing comments - I misread the `chain_id` on `execute_message` as part of the message itself, rather than the recipient of the message.

I have only one question:

Why not allow to transfer tokens to a different user on a different chain as well? Rather than supporting transfer under `msg.sender` only.

linera-service/tests/fixtures/erc20_shared.sol
```diff
@@ -0,0 +1,323 @@
...
+contract ERC20_shared is Context, IERC20, IERC20Errors {
+    mapping(address account => uint256) private _balances;
+
+    mapping(address account => mapping(address spender => uint256)) private _allowances;
+
+    uint256 private summed_total_supply; // total supply of all the chains.
+    uint256 private _totalSupply; // total supply of this chain
+
+    /**
+     * @dev Sets the values for {name} and {symbol}.
```
deuszx, line 54: This comment is invalid.

linera-service/tests/fixtures/erc20_shared.sol
```diff
@@ -0,0 +1,323 @@
...
+     *
+     */
+    constructor(uint256 the_supply) {
+        summed_total_supply = the_supply;
+    }
```
deuszx, line 59: As discussed during the offsite - "this code is bad" - this will initialize this particular token contract with `the_supply` on all chains (that request the app). It's basically a free `mint` .

PR #4205, opened 2025-07-08
Keep worker lock until chain state is dropped.

afck:
## Motivation

The root cause of https://github.com/linera-io/linera-protocol/issues/4178 seems to be that worker tasks in some cases are still handling requests after being evicted. If at that point another request for the same chain comes in, a new worker is created and there are two loaded `ChainStateView`s for the same chain at the same time.

## Proposal

Remove the worker tasks that handle requests in a loop. Instead, maintain a `Arc<RwLock<BTreeMap<ChainId, Arc<Mutex<ChainWorkerActor<StorageClient>>>>>>` where a lock for each actor is kept whenever any request is being handled, and also when it is removed from the map.

We could now consider removing the channel and the `ChainWorkerRequest` enum, and do some other simplifications, but I would like to separate them from the fix itself.

## Test Plan

With this change I cannot reproduce https://github.com/linera-io/linera-protocol/issues/4178 anymore.

## Release Plan

## Links

- Closes https://github.com/linera-io/linera-protocol/issues/4178.

### DIFF
--- a/linera-core/src/worker.rs
+++ b/linera-core/src/worker.rs
@@ -5,18 +5,20 @@
 use std::{
     collections::{BTreeMap, HashMap, HashSet, VecDeque},
     num::NonZeroUsize,
-    sync::{Arc, Mutex, RwLock},
+    sync::{self, Arc},
     time::Duration,
 };
 
-use futures::future::Either;
+use futures::{
+    future::Either,
+    stream::{FuturesUnordered, StreamExt},
+};
 use linera_base::{
     crypto::{CryptoError, CryptoHash, ValidatorPublicKey, ValidatorSecretKey},
     data_types::{ApplicationDescription, ArithmeticError, Blob, BlockHeight, Epoch, Round},
     doc_scalar,
     hashed::Hashed,
     identifiers::{AccountOwner, ApplicationId, BlobId, ChainId},
-    time::timer::{sleep, timeout},
 };
 #[cfg(with_testing)]
 use linera_chain::ChainExecutionContext;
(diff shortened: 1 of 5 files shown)


linera-core/src/worker.rs
```diff
@@ -682,120 +679,85 @@ where
...
-
-            let (sender, receiver) = mpsc::unbounded_channel();
-            chain_workers.push(chain_id, sender.clone());
-
-            Some((sender, Some(receiver)))
+                    .map(|(&chain_id, worker)| {
+                        let worker = Arc::clone(worker);
+                        async move { (chain_id, worker.lock_owned().await) }
+                    })
+                    .collect::<Vec<_>>(),
```
Twey, line 748: Nit: I think you can `collect()` straight into a `FuturesUnordered`.

linera-core/src/worker.rs
```diff
@@ -682,120 +679,85 @@ where
...
-    )> {
-        let mut chain_workers = self.chain_workers.lock().unwrap();
-
-        if let Some(endpoint) = chain_workers.get(&chain_id) {
-            Some((endpoint.clone(), None))
-        } else {
-            if chain_workers.len() >= usize::from(chain_workers.cap()) {
-                let (chain_to_evict, _) = chain_workers
+            )
+            .await
```
Twey, line 736: This is a little bit subtle and significantly the reason for the complication in the previous initialization code: holding the guard to the map over the await point here leaves us open to deadlocks if the caller doesn't remember to drive this task along with any other task that may access the chain workers map.  It's important that whenever we must hold a lock across an await point, we make sure that any concurrent calls that would try to take the lock also drive the task that is currently holding the lock.

afck:
Closing this, since it still needs substantial changes:
* @Twey's [comment](https://github.com/linera-io/linera-protocol/pull/4205#discussion_r2192338043) needs to be addressed.
* We currently return `Arc<Mutex<ChainWorkerActor>>`s while freeing the lock of the map, so they can still get removed from the map and then used.

PR #4208, opened 2025-07-09
Committee destination fixes

deuszx:
## Motivation

Fix http://github.com/linera-io/linera-protocol/pull/4207

The PR on which we build this is not passing tests. There were a couple of things wrong:
1. It incorrecty assigned _destination IDs_ to the spawned tasks:
    - it increments an ID when a new exporter needs to run and then uses that ID as an index in a vector. That vector is not initialized properly and - when loading a destination state - it panics due to missing indice
2. It created a new exporter, with empty destination state, for the same validator destination if it appeared in the new committee (meaning we would duplicate many block exports)
3. It created multiple single-threaded tokio Runtimes - we simplify it here with the use of single multi-threaded tokio Runtime with `max_worker_threads` limit.

## Proposal

Instead of manually assigning destination IDs we derive it as a tuple of `(address, destination kind)`. This simplifies the process of identifying the destinations for removal/update/restart. Using new ID we filter out old committee members and no longer shut them down if they are also in the new committee.

## Links

### DIFF
--- a/linera-service/src/exporter/runloops/mod.rs
+++ b/linera-service/src/exporter/runloops/mod.rs
@@ -219,7 +219,7 @@ mod test {
     {
         let mut destinations = Vec::new();
         let cancellation_token = CancellationToken::new();
-        let indexer = spawn_dummy_indexer(&mut destinations, &cancellation_token).await?;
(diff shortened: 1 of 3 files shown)


Twey (approved):
I think there are some improvements still on the table — but let's get this merged for now :)

PR #4209, opened 2025-07-09
Replace chain worker limit with TTL.

### LINKED ISSUE #4178: `MissingEntries` in `QueueView`

afck:
## Motivation

The root cause of https://github.com/linera-io/linera-protocol/issues/4178 seems to be that worker tasks in some cases are still handling requests after being evicted. If at that point another request for the same chain comes in, a new worker is created and there are two loaded `ChainStateView`s for the same chain at the same time.

## Proposal

Remove the chain worker limit: chain workers are never evicted.

Instead, introduce a chain worker state TTL: After that duration without any requests for a particular chain, the chain worker state is dropped. But the task itself keeps listening and will load the state again whenever a new request comes in.

## Test Plan

With this, I cannot reproduce #4178 anymore.

## Release Plan

## Links

- Closes #4178.


linera-core/src/chain_worker/state/mod.rs
```diff
@@ -99,6 +100,118 @@ where
         self.chain.chain_id()
     }
 
+    /// Handles a request and applies it to the chain state.
+    #[instrument(skip(self, request))]
```
Twey: Nit: `skip_all` (but we should trace the request! maybe inside as `tracing::debug!(?request, "handling request")`?)

Twey (approved):

linera-core/src/chain_worker/state/mod.rs
```diff
@@ -235,6 +348,19 @@ where
...
+    /// Clears the shared chain view, and acquires and drops its write lock.
+    ///
+    /// This is the only place a write lock is acquired, and read locks are acquired in
+    /// the `chain_state_view` method, which has a `&mut self` receiver like this one.
+    /// That means that when this function returns, no readers will be waiting to acquire
+    /// the lock and it is safe to write the chain state to storage without any readers
+    /// having a stale view of it.
+    pub(super) async fn clear_shared_chain_view(&mut self) {
+        if let Some(shared_chain_view) = self.shared_chain_view.take() {
+            let _ = shared_chain_view.write().await;
```
ma2bd: I always feel that `let _ =` are dangerous. Maybe add a type: `let _ : Guard = ...` ?

ndr-ds:
Nice! 🙌🏻

PR #4210, opened 2025-07-10
Insert a `DelegatedFungible` smart contract.

MathieuDutSik:
## Motivation

Both Ethereum with the ERC20 and Solana with the SPL have informal standards about the minimal set of capabilities of a fungible contract.

In those standards, a user A can allow another user B to access his funds up to some extent.
Then user B can spend tokens of user A up to the allowance.
This is missing from the fungible app.

## Proposal

Currently, we have `fungible` and `native-fungible` that implement the same API. A new `delegated-fungible` contract has been introduced. This is because both for Ethereum and Solana, the fungible token are informal standard that are not part of the core protocol.

In order to test the smart contract we are forced by the API to have a multi-owner chain.
This is the first time that a wasm application is deployed on a multi-owner chain.
This leads to several interesting things:
* The last argument of `publish_and_create` is `Some(chain2)` as opposed to `None` in all other tests.
* The lack of functionality to change the default chain of the client has appeared.
* The `process_inbox` command is used on the `node_service` even if only one chain is being used. This is because the first thing that `process_inbox` does is `synchronize_from_validators`. The synchronization is available in the client, but not in the `node_service`.

The `fungible` was taken as a model, and some improvements made in `delegated-fungible` could be backported to `fungible`:
* The `fn instantiate` does not check for emptiness but for the total sum of tokens. I think this is better.
* In the case of the absence of tokens, the `fn instantiate` is panicking, which I think is the right course of action as opposed to assigning arbitrary values.
* The access to the `MapView` is done via `get_mut_or_default`, which is better than `get`. This is not always possible since in the case of balance 0 we need to erase the key.
* The `execute_operation` is returning a non-trivial response. This is a little bit at odds with the use of `GraphQL` API for accessing the storage.

## Test Plan

The CI. A test has been added.

## Release Plan

## Links

None.


examples/delegated-fungible/README.md
```diff
@@ -0,0 +1,12 @@
...
+This example application implements a delegated fungible. It is the equivalent for Linera
+of Ethereum ERC20 contract and Solana SPL. That is the difference with `fungible` is that
+allowances are supported.
+
+## How It Works
+
+The API `approve` and `transfer_from` have been added to the functionality, following the ERC20
+contract.
+
+## Usage
```
deuszx: This section is missing - is that on purpose? If it's supposed to stay empty I'd say we should simply remove it.

examples/delegated-fungible/src/contract.rs
```diff
@@ -0,0 +1,187 @@
...
+        }
+    }
+
+    async fn store(mut self) {
+        self.state.save().await.expect("Failed to save state");
+    }
+}
+
+impl DelegatedFungibleTokenContract {
+    async fn claim(&mut self, source_account: Account, amount: Amount, target_account: Account) {
```
deuszx: A comment would be nice.

deuszx (commented):
I find this difficult to follow as different transfer-like operations (Credit, Debit, Transfer(From)) are mixed and work on either local (same chain) or remote (different chain) recipients. Can we deliminate them more clearly?

One the same note - actions on `self.state` and `self.runtime` and mixed up, depending on who's the source/recipient. It's not clear whether code is correct b/c of that. This is meant to serve as an example/reference so this has to be easier to understand and replicate

examples/delegated-fungible/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+        owner: AccountOwner,
+        /// The spender account
+        spender: AccountOwner,
+        /// Maximum amount to be transferred
+        allowance: Amount,
+    },
+    /// Transfers tokens from a (locally owned) account to a (possibly remote) account.
+    Transfer {
+        /// Owner to transfer from
+        owner: AccountOwner,
```
deuszx: Maybe we should rename `AccountOwner` to address 🤔 I find this a bit confusing that `AccountOwner` is local account while `Account` is both `AccountOwner` and chain.

examples/delegated-fungible/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+        /// Amount to be transferred
+        amount: Amount,
+        /// Target account to transfer the amount to
+        target_account: Account,
+    },
+    /// Transfers tokens from a (locally owned) account to a (possibly remote) account by using the allowance.
+    TransferFrom {
```
deuszx: Original ([from ERC20](https://docs.openzeppelin.com/contracts/4.x/api/token/erc20#IERC20-transferFrom-address-address-uint256-)) `transferFrom` has 3 arguments:
- from
- to
- amount

Why does our version need to have 4?

examples/delegated-fungible/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+        spender: AccountOwner,
+        /// Amount to be transferred
+        amount: Amount,
+        /// Target account to transfer the amount to
+        target_account: Account,
+    },
+    /// Same as `Transfer` but the source account may be remote. Depending on its
+    /// configuration, the target chain may take time or refuse to process
+    /// the message.
+    Claim {
```
deuszx: So if `source_account` is local chain then it's synchronous but if it's remote then it's an operation? And if it's local then this is no different from `transferFrom`?

examples/delegated-fungible/src/lib.rs
MathieuDutSik: In the ERC20, they use `msg.sender`, which is then the spender.

examples/delegated-fungible/README.md
MathieuDutSik: Ok.

examples/delegated-fungible/src/contract.rs
MathieuDutSik: Ok, done.

examples/delegated-fungible/src/lib.rs
MathieuDutSik: I think the whole issue of renaming is for another discussion, PR and so on.

MathieuDutSik:
> I find this difficult to follow as different transfer-like operations (Credit, Debit, Transfer(From)) are mixed and work on either local (same chain) or remote (different chain) recipients. Can we deliminate them more clearly?
> 
> One the same note - actions on `self.state` and `self.runtime` and mixed up, depending on who's the source/recipient. It's not clear whether code is correct b/c of that. This is meant to serve as an example/reference so this has to be easier to understand and replicate

All of this is part of the code `fungible` and is I think an argument for merging the fungible with delegated-fungible.

examples/delegated-fungible/src/lib.rs
deuszx: I agree, I just wanted to share this observation when looking at these constructs in a new context.

examples/delegated-fungible/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+This example application implements a delegated fungible. It is the equivalent for Linera
+of Ethereum ERC20 contract and Solana SPL. That is the difference with `fungible` is that
+allowances are supported.
+
+## How It Works
+
+The API `approve` and `transfer_from` have been added to the functionality, following the ERC20
+contract.
+
```
deuszx: I think it's missing the crucial description of the multi-chain aspect.

examples/delegated-fungible/src/contract.rs
```diff
@@ -0,0 +1,188 @@
...
+    async fn execute_operation(&mut self, operation: Self::Operation) -> Self::Response {
+        match operation {
+            Operation::Approve {
+                owner,
+                spender,
+                allowance,
+            } => {
+                self.runtime
+                    .check_account_permission(owner)
+                    .expect("Permission for Transfer operation");
```
deuszx: `check_account_permission` returns true if `owner` is equivalent to either `authenticated_caller_id` or `authenticated_signer`. I'm wondering:
1. does this work for all combinations of callers:
  - Solidity contract, 
  - Wasm contract, 
  - EVM account
  - Linera account
2. Do we even need to call it manually? in EVM world, the `owner` of the `Approve` call will be `msg.sender` (direct caller). Maybe we can pull that information from the runtime rather than requiring in the operation? I.e. it should be `authenticated_caller_id` (and for direct calls, that should be equivalent to `authenticated_signer`?)

examples/delegated-fungible/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+        /// Owner to transfer from
+        owner: AccountOwner,
+        /// The spender of the amount.
+        spender: AccountOwner,
+        /// Amount to be transferred
+        amount: Amount,
+        /// Target account to transfer the amount to
+        target_account: Account,
+    },
+    /// Same as `Transfer` but the source account may be remote. Depending on its
```
deuszx: It's more like `TransferFrom` if we can use a different source for the transfer.

Maybe we should have one construct that covers both - `TransferFrom` would use `Account` type for `spender` - i.e. the source from where we transfer tokens from can be either local chain or remote one.

examples/fungible/src/contract.rs
```diff
@@ -83,6 +92,23 @@ impl Contract for FungibleTokenContract {
...
+                amount,
+                target_account,
+            } => {
+                self.runtime
+                    .check_account_permission(spender)
+                    .expect("Permission for Transfer operation");
+                self.state
+                    .debit_for_transfer_from(owner, spender, amount)
+                    .await;
+                self.finish_transfer_to_account(amount, target_account, owner)
```
deuszx, line 107: Observation:

We should see if we can clean this up. This is a very simple contract and in these 3 lines we reach into three different things: `self.runtime`, `self.state` and `self.<method>`.

examples/fungible/src/lib.rs
```diff
@@ -41,6 +47,92 @@ pub enum Message {
     },
 }
 
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct OwnerSpender {
+    /// Account to withdraw from
+    pub owner: AccountOwner,
+    /// Account to do the withdrawing
+    pub spender: AccountOwner,
+}
```
deuszx, line 55: Do we need a wrapper struct? I think it might be adding unnecessary complexity/bloat.

examples/fungible/src/lib.rs
```diff
@@ -41,6 +47,92 @@ pub enum Message {
...
+
+scalar!(OwnerSpender);
+
+impl OwnerSpender {
+    pub fn new(owner: AccountOwner, spender: AccountOwner) -> Self {
+        if owner == spender {
+            panic!("owner should be different from spender");
+        }
```
deuszx, line 63: I don't think this is a hard requirement.

deuszx (approved):

PR #4211, opened 2025-07-10
Kebab cased config-path exporter arg.

christos-h:
## Motivation

Keep consistent casing in binary arguments.

## Proposal

Minor fix.

## Test Plan

CI should cach regressions.

### DIFF
--- a/linera-service/src/exporter/main.rs
+++ b/linera-service/src/exporter/main.rs
@@ -36,7 +36,7 @@ mod test_utils;
 )]
 struct ExporterOptions {
     /// Path to the TOML file describing the configuration for the block exporter.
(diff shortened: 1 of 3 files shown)


deuszx (approved):

linera-service/src/exporter/main.rs
```diff
@@ -36,7 +36,7 @@ mod test_utils;
 )]
 struct ExporterOptions {
     /// Path to the TOML file describing the configuration for the block exporter.
-    #[arg(long = "config_path")]
+    #[arg(long = "config-path")]
```
ma2bd: Just use `#[arg(long)]`

PR #4012, opened 2025-05-26
migrate to `async-graphql` 7.0.17

ma2bd:
## Motivation

Remove temporary workaround introduced in #3940

## Proposal

migrate to async-graph 7.0.17

## Test Plan

CI

## Release Plan

These changes should be backported to the latest `testnet` branch, then be released in a new SDK,

PR #4212, opened 2025-07-10
`testnet_babbage`: pinned GraphiQL

Twey:
## Motivation

Backport of #4012.


ma2bd (approved):
~~Please improve the title of the PR before merging. (I believe we're removing the pinned version)~~

Twey:
No, we are updating the version of `async-graphql` to 7.0.17, which pins the version of GraphiQL.  In 7.0.16 and before the GraphiQL version is unpinned, which is what leads to the behaviour changing underneath us.

Edit: I see the confusion — this

PR #4213, opened 2025-07-11
Provide access to further linera functionalities in the EVM.

MathieuDutSik:
## Motivation

We want the EVM applications to be able to fully access the Linera functionalities.
Here we extend it by adding access to the Linera balances, authenticated signer
as well as other features.

## Proposal

The work simply adds the following functionalities to EVM applications:
* chain_timestamp
* chain_balance
* owner_balance
* owner_balances
* balance_owners
* authenticated_signer
* authenticated_caller_id
* query_service

As it was it looked simple but several problems showed up:
* New types have to be added, but since we do not have a type `uint256` in YAML file, we have to use a `[u8; 32]`.
* This forces a conversion to `uint256` to do in the code. The problem is that this externalize some types, so we have an `AccountOwnerBalance` and `AccountOwnerBalanceInner`.
* In order to force having all the types in `LineraTypes` and so in order to have a `LineraTypes.AccountOwnerBalance` and not a `Linera.AccountOwnerBalance`, we patch the generated file. This is added to the CI.
* The serialization of the U256 is incorrect and returns 33 bytes instead of 32. In order to deal with that, a type `AmountU256` is introduced to encapsulate them.

## Test Plan

The CI and the additional tests.

## Release Plan

## Links

None.

### DIFF
--- a/linera-execution/solidity/Linera.sol
+++ b/linera-execution/solidity/Linera.sol
@@ -13,41 +13,381 @@ import "./LineraTypes.sol";
 
 library Linera {
 
-    function chain_id() internal returns (LineraTypes.ChainId memory) {
+    // Exported types
+
+    struct ChainId {
+        bytes32 value;
+    }
+
+    function chainid_from(LineraTypes.ChainId memory entry)
+        internal
+        pure
+        returns (ChainId memory)
+    {
+        return ChainId(entry.value.value);
+    }
+
+    struct AccountOwner {
+        uint8 choice;
+        // choice=0 corresponds to Reserved
+        uint8 reserved;
+        // choice=1 corresponds to Address32
+        bytes32 address32;
+        // choice=2 corresponds to Address20
+        bytes20 address20;
+    }
+
+    function accountowner_from(LineraTypes.AccountOwner memory owner)
+        internal
+        pure
+        returns (AccountOwner memory)
+    {
+        return AccountOwner(owner.choice, owner.reserved, owner.address32.value, owner.address20);
+    }
+
+    function accountowner_to(Linera.AccountOwner memory owner)
+        internal
+        pure
+        returns (LineraTypes.AccountOwner memory)
+    {
+        LineraTypes.CryptoHash memory hash = LineraTypes.CryptoHash(owner.address32);
+        return LineraTypes.AccountOwner(owner.choice, owner.reserved, hash, owner.address20);
+    }
+
+    struct AccountOwnerBalance {
+        Linera.AccountOwner account_owner;
+        uint256 balance;
+    }
+
+    function accountownerbalance_from(LineraTypes.AccountOwnerBalance memory entry)
+        internal
+        pure
+        returns (AccountOwnerBalance memory)
+    {
+        return AccountOwnerBalance(accountowner_from(entry.account_owner), entry.balance);
+    }
+
+    struct TimeDelta {
+        uint64 value;
+    }
+
+    function timedelta_from(LineraTypes.TimeDelta memory entry)
+        internal
+        pure
+        returns (TimeDelta memory)
+    {
+        return TimeDelta(entry.value);
+    }
+
+    struct opt_TimeDelta {
+        bool has_value;
+        uint64 value;
+    }
+
+    function opt_timedelta_from(LineraTypes.opt_TimeDelta memory entry)
+        internal
+        pure
+        returns (opt_TimeDelta memory)
+    {
+        return opt_TimeDelta(entry.has_value, entry.value.value);
+    }
+
+    struct TimeoutConfig {
+        opt_TimeDelta fast_round_duration;
+        TimeDelta base_timeout;
+        TimeDelta timeout_increment;
+        TimeDelta fallback_duration;
+    }
+
+    function timeoutconfig_from(LineraTypes.TimeoutConfig memory entry)
+        internal
+        pure
+        returns (TimeoutConfig memory)
+    {
+        return TimeoutConfig(opt_timedelta_from(entry.fast_round_duration),
+                             timedelta_from(entry.base_timeout),
+                             timedelta_from(entry.timeout_increment),
+                             timedelta_from(entry.fallback_duration));
+    }
+
+    struct AccountOwnerWeight {
+        Linera.AccountOwner account_owner;
+        uint64 weight;
+    }
+
+    function accountownerweight_from(LineraTypes.key_values_AccountOwner_uint64 memory entry)
+        internal
+        pure
+        returns (AccountOwnerWeight memory)
+    {
+        return AccountOwnerWeight(accountowner_from(entry.key), entry.value);
+    }
+
+    struct ChainOwnership {
+        AccountOwner[] super_owners;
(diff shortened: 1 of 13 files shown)


linera-execution/src/evm/revm.rs
```diff
@@ -484,35 +487,49 @@ fn address_to_user_application_id(address: Address) -> ApplicationId {
...
 enum ContractRuntimePrecompile {
-    /// Calling `try_call_application` of `ContractRuntime`
-    TryCallApplication {
-        target: ApplicationId,
-        argument: Vec<u8>,
-    },
-    /// Calling `validation_round` of `ContractRuntime`
-    ValidationRound,
+    /// Calling `authenticated_signer` of `ContractRuntime`
+    AuthenticatedSigner,
```
deuszx, line 507: Isn't this the same as `tx.origin`?

linera-execution/src/evm/revm.rs
```diff
@@ -484,35 +487,49 @@ fn address_to_user_application_id(address: Address) -> ApplicationId {
...
+    /// Calling `message_id` of `ContractRuntime`
+    MessageId,
+    /// Calling `message_is_bouncing` of `ContractRuntime`
+    MessageIsBouncing,
+    /// Calling `authenticated_caller_id` of `ContractRuntime`
+    AuthenticatedCallerId,
```
deuszx, line 513: Isn't this `msg.sender`?

linera-execution/solidity/LineraTypes.sol.patch
```diff
@@ -0,0 +1,14 @@
...
+ library LineraTypes {
+ 
++    struct AccountOwnerBalance {
++        LineraTypes.AccountOwner account_owner;
++        uint256 balance;
++    }
++
+     function bcs_serialize_len(uint256 x)
+         internal
+         pure
```
deuszx: Why do we need to check in the patch, rather than regenerate `LineraTypes.sol` ?

linera-execution/src/evm/revm.rs
MathieuDutSik: Not if you have a Wasm application that calls the EVM contract.
And in any case, the `tx.origin` is not set right now, but will be done later.

linera-execution/solidity/Linera.sol
```diff
@@ -23,6 +25,17 @@ library Linera {
         return LineraTypes.bcs_deserialize_ChainId(output);
     }
 
+    function block_height() internal returns (uint64) {
```
deuszx, line 286: Is there a plan to find the usage cost of these new `Linera`-specific methods?

linera-execution/src/evm/revm.rs
MathieuDutSik: Not if a Wasm application calls the contract.

linera-execution/solidity/LineraTypes.sol.patch
MathieuDutSik: The type `uint256` is not part of what can be generated by `serde-reflection`. Therefore, this type cannot be generated by `serdegen`.

linera-execution/solidity/Linera.sol
MathieuDutSik: There is better than a plan, there is an implementation:
```rust
    fn block_height(&mut self) -> Result<BlockHeight, ExecutionError> {
        let mut this = self.inner();
        let height = this.height;
        this.resource_controller.track_runtime_block_height()?;
        Ok(height)
    }
```

linera-execution/src/evm/revm.rs
deuszx: so, if I write a Solidity app then I have to check both `msg.sender` and `AuthenticatedCallerId` ? b/c I don't know if I'm being called by:
- Wasm contract
- Solidity contract
- Ethereum account
- Linera account
?

linera-execution/solidity/Linera.sol
deuszx: Sure, but we don't really know what's the cost of accesing these endpoints from Wasm vs Solidity. There is different "glue" between the host and contracts that will take different resources

MathieuDutSik: What makes you think so?

deuszx: Surely the overhead of Linera SDK, Wasm (de)serialisation, VM implementation is different from Solidity & EVM stack?

linera-execution/solidity/LineraTypes.sol.patch
deuszx: Ok, I understand it doesn't exactly fit. It's weird though that we have to patch it. It means there will never be a static file in our repo that can be used as single.source of truth.

linera-execution/src/evm/revm.rs
MathieuDutSik: If you have a normal solidity application that you move to Linera then what happens is that if the address is not Ethereum one then the `msg.sender` is Address::ZERO. That Address is typically dismissed by solidity smart contracts.

If the solidity smart contracts want to support the same functionality as a classic Wasm smart contract, then it has the same programming complexity as that Wasm Linera application.

deuszx: But if this was Wasm contract, we don't need to check in four different places but two? Because the enpoints (authenticated signer/caller) return enums?

linera-execution/solidity/Linera.sol
MathieuDutSik: And?

deuszx: Nevermind

linera-execution/solidity/LineraTypes.sol.patch
ma2bd: @MathieuDutSik At this point, I would suggest writing a binary tool based on `serde-generate` as a library. The library exposes a number of advanced codegen options that are not available in CLI.

linera-execution/src/evm/revm.rs
deuszx: Or to put differently - I understand why it is this way, I'm just trying to point out that this will require different type of defensive-programming depending in which language (Solidity vs Linera SDK) the contract is written. This might be difficult to explain.

It'd be best if the experience was as similar as possible.

linera-execution/solidity/LineraTypes.sol.patch
MathieuDutSik: Ok.
I have some alternative proposal:
* Write the types that are returned in `Linera.sol`. This is a little bit of boilerplate code to write.
* This allows removing the `LineraTypes.XX` types from the output, which I think is a big win.
* No need to patch the code.

linera-execution/src/evm/revm.rs
MathieuDutSik: We come back to some tracks that we covered before: The conversion of address between `Address32` and `Address20`.

We decided so far to keep both kinds of addresses. And I think this is the correct decision. There is simply a difference in kind of addresses, and having a clear separation is quite good in my opinion.

PR #4226, opened 2025-07-15
`linera-web`: support getting balance

Twey:
## Motivation

A partner wants to see the chain balance.

## Proposal

Add a function to get the current chain balance.

## Test Plan

CI.

## Release Plan

    - be released in a new SDK,

## Links

### DIFF
--- a/linera-web/src/lib.rs
+++ b/linera-web/src/lib.rs
@@ -321,6 +321,19 @@ impl Client {
         Ok(())
     }
 
+    /// Gets the balance of the default chain.
+    ///
+    /// # Errors
+    /// If the chain couldn't be established.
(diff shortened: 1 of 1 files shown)


linera-web/src/lib.rs
```diff
@@ -321,6 +321,19 @@ impl Client {
         Ok(())
     }
 
+    /// Gets the balance of the default chain.
+    ///
+    /// # Errors
+    /// If the chain couldn't be established.
+    pub async fn balance(&self) -> JsResult<String> {
```
deuszx, line 328: It should be fairly easy to get the balance of any chain I think 🤔 Actually, since all chains have the same state encoding for the native balance we should be able to query it w/o even downloading that chain

PR #4232, opened 2025-07-17
Replace unnecessary clone() with as_ref() in error handling

dizer-ti:
## Motivation
Replace app_permissions.execute_operations.clone().unwrap() with app_permissions.execute_operations.as_ref().unwrap() to avoid unnecessary cloning when formatting error messages


afck:
Doesn't build; `clone` is necessary.

PR #4233, opened 2025-07-18
Add development environment in Docker

Twey:
## Motivation

We'd like people who like Docker to be able to use Docker to build our software.

## Proposal

Build a composite Rust toolchain using [`nixup`](https://github.com/Twey/nixup) that combines our stable and nightly Rust toolchains in a way that's compatible with rustup so we can combine both `devShell`s into one `devShell`.  Export a build output `packages.<arch>.container` that builds a Docker image containing the default development shell.

As an aside, clean up a couple of things in the Nix:
- fix macOS building
- filter the source with `nix-gitignore` to reduce the data copied to the Nix store

## Test Plan

This has been tested locally on my machine (`x86_64-linux`) and @deuszx' macOS machine (`aarch64-arm-darwin`).

## Release Plan

## Links

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -7,6 +7,7 @@
     # Rust
     rust-overlay.url = "github:oxalica/rust-overlay";
     crane.url = "github:ipetkov/crane";
+    nixup.url = "github:Twey/nixup";
 
     # Dev tools
     treefmt-nix.url = "github:numtide/treefmt-nix";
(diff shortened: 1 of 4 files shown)


flake.nix
```diff
@@ -7,6 +7,7 @@
     # Rust
     rust-overlay.url = "github:oxalica/rust-overlay";
     crane.url = "github:ipetkov/crane";
+    nixup.url = "github:Twey/nixup";
```
deuszx, line 10: 👀  pour qui?

Twey: Pour nous :) (see the description)

Twey: The previous (Nix-flow) approach to having two toolchains was to use a `devShell` for each.  This means we can have a single `devShell` that includes both, which is helpful since we can only put the one `devShell` in the Docker image.

As a bonus it brings the Nix flow closer to the rustup flow since they both can now use the `cargo +nightly` syntax to access Rust nightly, though the Nix flow sets up both toolchains automatically while rustup requires a manual step to add the nightly toolchain (something like `cd toolchains/nightly && rustup show`).

deuszx (approved):

PR #4239, opened 2025-07-18
Explicitly set max_concurrent_streams to u32::MAX-1

bart-linera:
Backport of #4237 to the testnet branch

### DIFF
--- a/linera-service/src/proxy/grpc.rs
+++ b/linera-service/src/proxy/grpc.rs
@@ -261,9 +261,8 @@ where
             .build_v1()?;
         let public_server = join_set.spawn_task(
             self.public_server()?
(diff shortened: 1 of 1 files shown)


linera-service/src/proxy/grpc.rs
```diff
@@ -261,6 +261,9 @@ where
             .build_v1()?;
         let public_server = join_set.spawn_task(
             self.public_server()?
+                .max_concurrent_streams(Some(u32::MAX - 1)) // we subtract one to make sure
+                // that the value is not
+                // interpreted as "not set"
```
Twey: The formatting is a little odd here — could we move the comment above the call?

bart-linera: I guess we should probably do that on the other two branches as well...

bart-linera:
Closing as outdated.

PR #4240, opened 2025-07-18
Replace `ReadBlob` with `VerifyBlob`

### LINKED ISSUE #2490: Remove ReadBlob
It was introduced to be test-only but it is very risky to change the block data format depending on the `test` feature.

We could remove `ReadBlob` altogether or replace it with a proper instruction `VerifyBlob` that doesn't the same but (hopefully) without downloading the blob from storage.

afck:
## Motivation

The `ReadBlob` operation was added for testing; it reads a blob but then drops the result.

## Proposal

Rename it to `VerifyBlob` and only assert the blob's existence.

## Test Plan

CI

## Release Plan

## Links

- Closes #2490

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -2224,7 +2224,7 @@ impl<Env: Environment> ChainClient<Env> {
             hash,
             blob_type: BlobType::Data,
         };
-        self.execute_operation(SystemOperation::ReadBlob { blob_id })
+        self.execute_operation(SystemOperation::VerifyBlob { blob_id })
             .await
     }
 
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -53,7 +53,6 @@ default-members = [
     "linera-summary",
     "linera-views",
     "linera-views-derive",
-    # "linera-web", # not this one
```
bart-linera: I think this is supposed to serve some purpose, but I'm not sure what exactly, TBH :sweat_smile:

bart-linera (approved):

PR #4246, opened 2025-07-19
Remove associated types Keys and KeyValues in KV store traits

ma2bd:
## Motivation

* Simplify code.

As pointed out by teammates, this optimization is only used for `DynamoDb` and doesn't apply to the databases that we use in production.

## Proposal

* Remove associated types Keys and KeyValues
* Remove the traits

I stop short of removing the iterator objects in DynamoDb. We could probably simplify the code some more but it's non-trivial and could be slightly slower. 

## Test Plan

CI

### DIFF
--- a/linera-views/src/backends/dynamo_db.rs
+++ b/linera-views/src/backends/dynamo_db.rs
@@ -45,10 +45,7 @@ use crate::{
     common::get_uleb128_size,
     journaling::{DirectWritableKeyValueStore, JournalConsistencyError, JournalingKeyValueStore},
     lru_caching::{LruCachingConfig, LruCachingStore},
-    store::{
-        AdminKeyValueStore, KeyIterable, KeyValueIterable, KeyValueStoreError,
-        ReadableKeyValueStore, WithError,
-    },
+    store::{AdminKeyValueStore, KeyValueStoreError, ReadableKeyValueStore, WithError},
     value_splitting::{ValueSplittingError, ValueSplittingStore},
     FutureSyncExt as _,
 };
@@ -260,16 +257,6 @@ fn extract_key_value(
     Ok((key, value))
 }
 
-/// Extracts the `(key, value)` pair attributes from an item (returned by value).
-fn extract_key_value_owned(
-    prefix_len: usize,
-    attributes: &mut HashMap<String, AttributeValue>,
-) -> Result<(Vec<u8>, Vec<u8>), DynamoDbStoreInternalError> {
-    let key = extract_key(prefix_len, attributes)?.to_vec();
-    let value = extract_value_owned(attributes)?;
-    Ok((key, value))
-}
-
 struct TransactionBuilder {
     start_key: Vec<u8>,
     transactions: Vec<TransactWriteItem>,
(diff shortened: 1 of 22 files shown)


linera-views/src/backends/dynamo_db.rs
```diff
@@ -697,73 +676,40 @@ struct QueryResponses {
...
         }
     }
 }
 
-/// A set of `(key, value)` returned by a search query on DynamoDB.
-pub struct DynamoDbKeyValues {
-    result_queries: QueryResponses,
-}
-
+// Inspired by https://depth-first.com/articles/2020/06/22/returning-rust-iterators/
```
afck: Why not use `impl Iterator`? I don't think either of the arguments against it apply here.

```rust
self.responses
    .iter()
    .flat_map(|response| &response.items)
    .map(move |item| extract_key(self.prefix_len, item))
```

(Edit: Ah, you say it would be slightly slower? But I don't see why.)

linera-views/src/backends/dynamo_db.rs
```diff
@@ -926,21 +834,30 @@ impl ReadableKeyValueStore for DynamoDbStoreInternal {
...
+    ) -> Result<Vec<Vec<u8>>, DynamoDbStoreInternalError> {
         let result_queries = self
             .get_list_responses(KEY_ATTRIBUTE, &self.start_key, key_prefix)
             .await?;
-        Ok(DynamoDbKeys { result_queries })
+        let mut keys = Vec::new();
+        for key in result_queries.iterate_keys() {
+            keys.push(key?.to_vec());
+        }
+        Ok(keys)
```
afck: Would probably allocate the right vec size right away (also below):

```suggestion
        result_queries.iterate_keys()
            .map(|key| key.map(|k| k.to_vec()))
            .collect()
```

afck (approved):

linera-views/src/backends/dynamo_db.rs
ma2bd: Ok let me see if I can simplify the code further. I can also look at what @MathieuDutSik initially wrote since we're basically reverting to that

ma2bd: Piece of cake for claude

linera-views/src/backends/value_splitting.rs
```diff
@@ -202,12 +199,11 @@ where
...
         key_prefix: &[u8],
-    ) -> Result<Self::KeyValues, Self::Error> {
+    ) -> Result<Vec<(Vec<u8>, Vec<u8>)>, Self::Error> {
         let small_key_values = self.store.find_key_values_by_prefix(key_prefix).await?;
-        let mut small_kv_iterator = small_key_values.into_iterator_owned();
+        let mut small_kv_iterator = small_key_values.into_iter();
         let mut key_values = Vec::new();
-        while let Some(result) = small_kv_iterator.next() {
-            let (mut big_key, value) = result?;
+        while let Some((mut big_key, value)) = small_kv_iterator.next() {
```
MathieuDutSik, line 206: Perhaps we can eliminate the need for the `small_kv_iterator` and the while loop.

ma2bd: I don't see how because of the `small_kv_iterator` below.

Issue #4022, opened 2025-05-28
Pay for publishing blobs before uploading data.


afck:
The operation that publishes data or bytecode blobs should not depend on the actual data. Instead, the operation should specify the hash and size of the blob, and cost the appropriate fees for that size. Each node that has processed this operation in a _confirmed_ block should store the information in the blob state: That a blob of this size with this hash has been paid for.

Whenever a user now uploads that blob, the size and hash are checked and it is stored. After that, it can be used in any block.

With these new semantics:
* Validators never have to accept any blob data before the payment for the data is finalized.
* We can remove `ChainStateView::pending_proposed_blobs`.

PR #4247, opened 2025-07-20
Allow contract to create blobs

### LINKED ISSUE #3707: Allow applications to create data blobs.
Applications should be able to programmatically create and publish data blobs.

MathieuDutSik:
## Motivation

In order for the EVM to be able to create contracts, it needs to be able to create the corresponding data blobs.
So, that the function has been introduced.

Fixes #3707 

## Proposal

The code has been implemented as expected, except for a specific point:
* What we need to create is a blob of the type `EvmBytecode`.
* Therefore, we need to introduce a `create_blob` instead of a `create_data_blob`.
* The `read_data_blob` / `assert_data_blob` could be extended to support all the blobs. But this has not been done in this PR.

## Test Plan

The CI, and 3 tests that have been introduced.

## Release Plan

## Links

None


afck (changes requested):
We don't want to allow contracts to create arbitrary blobs, in particular not _committee_ blobs, or arbitrary application description blobs (with the wrong chain ID), or chain descriptions (with the wrong parent chain).

Instead, maybe we should add a `publish_module` function, that specifically publishes bytecode blobs.

MathieuDutSik:
> We don't want to allow contracts to create arbitrary blobs, in particular not _committee_ blobs, or arbitrary application description blobs (with the wrong chain ID), or chain descriptions (with the wrong parent chain).
> 
> Instead, maybe we should add a `publish_module` function, that specifically publishes bytecode blobs.

Thank you. That is indeed the right way to do.
PR has been updated.

linera-sdk/src/contract/runtime.rs
```diff
@@ -378,6 +380,22 @@ where
...
     }
 
+    /// Creates a new data blob and returns its hash.
+    pub fn create_data_blob(&mut self, bytes: Vec<u8>) -> DataBlobHash {
+        let blob_id = contract_wit::create_data_blob(&bytes);
+        DataBlobHash(blob_id.hash.into())
+    }
+
+    /// Publishes a module with contract and service bytecode and returns the module ID.
+    pub fn publish_module(
```
afck, line 390: (I wonder if it would be better to have a `publish_wasm_module` with two and a `publish_evm_module` with only one argument? Maybe also on the CLI and in the node service. Let's discuss with the team.)

linera-base/src/data_types.rs
```diff
@@ -1118,6 +1118,23 @@ impl Bytecode {
 
         CompressedBytecode { compressed_bytes }
     }
+
+    /// Compresses the [`Bytecode`] into a [`CompressedBytecode`].
+    #[cfg(target_arch = "wasm32")]
+    pub fn compress(&self) -> CompressedBytecode {
+        use ruzstd::encoding::{CompressionLevel, FrameCompressor};
```
afck, line 1125: This is used during contract execution now, is it?
I assume this is deterministic; but we still need to make sure we never accidentally upgrade this crate after mainnet, if the upgrade changes the outcome at all.

linera-execution/src/runtime.rs
```diff
@@ -1590,6 +1590,27 @@ impl ContractRuntime for ContractSyncRuntimeHandle {
...
+        &mut self,
+        contract: Bytecode,
+        service: Bytecode,
+        vm_runtime: VmRuntime,
+    ) -> Result<ModuleId, ExecutionError> {
+        let (blobs, module_id) =
+            crate::runtime::create_bytecode_blobs_sync(contract, service, vm_runtime);
+        for blob in blobs {
+            self.inner().transaction_tracker.add_created_blob(blob);
+        }
```
afck, line 1610: Remember to charge the fees.

linera-sdk/src/contract/runtime.rs
MathieuDutSik: Maybe (and I am not enthusiastic about this), but I would create an issue for this.
That seems fairly fine right now.

linera-base/src/data_types.rs
MathieuDutSik: Yes, I was concerned with a different issue, which is the compatibility of `ruzst` and `zstd` but since it was already used, I thought it would be ok.

But yes, this is a concern. What we could do is add tests that check the correctness of the compression/decompression. It would not be 100% catching, of course, but that could be better than nothing.

afck: And maybe also compare some compression outcome's hash with a hard-coded hash, so we notice if some update would change the compression outcome.

ma2bd: Is this only computed for block proposals or also certificates?

afck: Also certificates: It's run during execution, if I understand correctly.

linera-execution/src/runtime.rs
MathieuDutSik: Good question.
I am not sure we need, but if so, then it also applies to `open_chain` and `create_application`.
So, let me check.

MathieuDutSik: My conclusion is that the tracking is done by one single function, which is `track_blob_published`.
That function is called from the `process_txn_outcome` in `linera-core`.
Therefore, I believe that the blobs created here are processed in the same manner as other blobs generated in the block, and no additional processing code is required.

MathieuDutSik: Maybe you mean issue #4022, which I agree is a problem, but should be done in another PR.

linera-execution/src/runtime.rs
```diff
@@ -1828,3 +1849,32 @@ impl From<&MessageContext> for ExecutingMessage {
         }
     }
 }
+
+/// Creates a compressed Contract, Service and bytecode synchronously.
```
afck:
```suggestion
/// Creates a compressed contract and service bytecode synchronously.
```

linera-execution/tests/contract_runtime_apis.rs
```diff
@@ -950,3 +950,269 @@ async fn test_perform_http_request(authorized_apps: Option<Vec<()>>) -> Result<(
 
     Ok(())
 }
+
+/// Tests the contract system API to create data blobs.
```
afck: Do we need this test? Feels redundant, given the two tests below.

linera-execution/tests/contract_runtime_apis.rs
```diff
@@ -950,3 +950,269 @@ async fn test_perform_http_request(authorized_apps: Option<Vec<()>>) -> Result<(
...
+
+    let created_blob1 = created_blobs.get(&expected_blob_id1).unwrap();
+    let created_blob2 = created_blobs.get(&expected_blob_id2).unwrap();
+    assert_eq!(created_blob1.bytes(), test_data1);
+    assert_eq!(created_blob2.bytes(), test_data2);
+
+    Ok(())
+}
+
+/// Tests creating an empty data blob.
```
afck: …and we could just make one of the two blobs above empty.

linera-execution/tests/contract_runtime_apis.rs
```diff
@@ -950,3 +950,269 @@ async fn test_perform_http_request(authorized_apps: Option<Vec<()>>) -> Result<(
...
+    // Verify the empty blob was created and tracked
+    let created_blobs = transaction_tracker.created_blobs();
+    assert!(created_blobs.contains_key(&expected_blob_id));
+    let created_blob = created_blobs.get(&expected_blob_id).unwrap();
+    assert!(created_blob.bytes().is_empty());
+
+    Ok(())
+}
+
+/// Tests the contract system API to publish a module with WASM runtime.
```
afck: Same here. It feels like `test_publish_module_different_bytecode` covers everything this test does.

afck (approved):

linera-execution/tests/contract_runtime_apis.rs
MathieuDutSik: Correct.

linera-execution/tests/contract_runtime_apis.rs
MathieuDutSik: Correct, done.

linera-execution/tests/contract_runtime_apis.rs
MathieuDutSik: Yes, changed. Just one test suffices.

linera-execution/src/runtime.rs
MathieuDutSik: Ok, thanks.

Issue #4231, opened 2025-07-17
Make fewer queries when executing blocks


Twey:
Currently we observe that an increment of the `counter` example makes 48 requests to validators.  This seems excessive, especially for a pre-initialized single-owner chain, which theoretically doesn't require any new information from the validators.

afck:
@deuszx: I think you made several improvements—that number might be much lower now? Should this issue be closed?

deuszx:
> [@deuszx](https://github.com/deuszx): I think you made several improvements—that number might be much lower now? Should this issue be closed?

Yes, currently counter increment makes `3 * num_ validators` queries.

Issue #3927, opened 2025-05-13
Remove `InactiveChain`/`BlobsNotFound` ambiguity.


afck:
In some places if a chain has not been initialized yet, we return `BlobsNotFound` with the chain description blob, in other places `InactiveChain`. This should be unified.

MathieuDutSik:
There is indeed an ambiguity here.

Here is the issues that I identified:
* The `InactiveChain` in `ExecutionError` gets converted to `NodeError::ExecutionError` while it could very well be converted to a `NodeError::InactiveChain` or a `NodeError::BlobsNotFound`. 
* The only pattern in which an `InactiveChain` is created in `linera-execution` is for a missing admin chain. Maybe we could have instead a `InactiveAdminChain`? 
* The `BlobsNotFound` by its name sounds like an error. However, the `InactiveChain` does not carry that meaning. So, maybe that should be the error being returned when a chain is missing.

afck:
Right, so maybe we could remove `InactiveChain` entirely and always use `BlobsNotFound`?

MathieuDutSik:
That makes sense. And then the result has to be interpreted, whether it is an error by the user or by the protocol.

Issue #3519, opened 2025-03-10
Make `next_replayed_oracle_response` safer to use.


afck:
The caller is currently expected to match against the expected `OracleResponse` variant, return a specific error if it doesn't match, use the actual oracle if not in replay mode, and in either case call `add_oracle_response` afterwards.

We could at least pass in the pattern matching closure to extract the data from the variant.

We could also consider calling `add_oracle_response` automatically if it was found, so the caller would only have to call it when using the oracle.

Maybe it's even feasible to pass in the closure that uses the oracle.

PR #4248, opened 2025-07-21
Avoid redundant chain info requests; fix event oracles.

afck:
## Motivation

The client makes a lot of chain info queries, causing unnecessary round trips. One example is `ValidatorUpdater::send_chain_information`, where we always first request the chain info from the validator, to learn which blocks they are missing.

When reading events we are not always using the oracle mechanism. that can cause confirmed blocks to fail even though the proposal succeeded.

## Proposal

* In `send_chain_information`, always send the highest block right away. Often that is enough, and the response will confirm that.
* In `send_chain_information`, only send the timeout certificate if it is relevant for the validator.
* On `NodeError::InactiveChain`, send blobs, not certificates. (This may be unreachable anyway; needs a closer look: https://github.com/linera-io/linera-protocol/issues/3927)
* Use the oracle mechanism where appropriate.
* Add chain info queries to the worker metrics.

## Test Plan

CI

Locally, it reduces the number of chain info queries handled by the validators in the `test_open_chain_then_close_it` test from 75 to 72.

## Release Plan

## Links

- Should partially address #4231, but needs verification.

### DIFF
--- a/linera-core/src/worker.rs
+++ b/linera-core/src/worker.rs
@@ -88,6 +88,14 @@ mod metrics {
             &["validator_name"],
         )
     });
+
+    pub static CHAIN_INFO_QUERIES: LazyLock<IntCounterVec> = LazyLock::new(|| {
+        register_int_counter_vec(
+            "chain_info_queries",
+            "Number of chain info queries processed",
+            &[],
+        )
+    });
 }
 
 /// Instruct the networking layer to send cross-chain requests and/or push notifications.
@@ -917,6 +925,8 @@ where
         query: ChainInfoQuery,
     ) -> Result<(ChainInfoResponse, NetworkActions), WorkerError> {
         trace!("{} <-- {:?}", self.nickname, query);
+        #[cfg(with_metrics)]
+        metrics::CHAIN_INFO_QUERIES.with_label_values(&[]).inc();
         let result = self
             .query_chain_worker(query.chain_id, move |callback| {
                 ChainWorkerRequest::HandleChainInfoQuery { query, callback }
--- a/linera-execution/src/execution_state_actor.rs
+++ b/linera-execution/src/execution_state_actor.rs
@@ -12,7 +12,8 @@ use futures::{channel::mpsc, StreamExt as _};
 use linera_base::prometheus_util::MeasureLatency as _;
 use linera_base::{
     data_types::{
-        Amount, ApplicationPermissions, ArithmeticError, BlobContent, BlockHeight, Timestamp,
+        Amount, ApplicationPermissions, ArithmeticError, BlobContent, BlockHeight, OracleResponse,
+        Timestamp,
     },
     ensure, hex_debug, hex_vec_debug, http,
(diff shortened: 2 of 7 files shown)


afck:
It looks like this depends on https://github.com/linera-io/linera-protocol/issues/365, specifically sending certificates for missing events when making a block proposal.

deuszx:
There's noticable improvement in number of queries we make with this PR (compared to `main`):

1. `main`
image
(9 queries)


2. this PR
image
(7 queries; `HandleLiteCertificate` transfer less data than `HandleChainInfoQuery` as well)

linera-core/src/worker.rs
```diff
@@ -88,6 +88,14 @@ mod metrics {
             &["validator_name"],
         )
     });
+
+    pub static CHAIN_INFO_QUERIES: LazyLock<IntCounterVec> = LazyLock::new(|| {
+        register_int_counter_vec(
```
deuszx, line 93: if we had `register_int_counter` (no `_vec`) then we wouldn't need the `.with_label_values(&[])`

linera-execution/src/system.rs
```diff
@@ -577,14 +577,23 @@ where
...
+                            {
+                                missing_events.push(event_id);
+                                continue;
+                            }
+                        }
+                        Some(OracleResponse::EventExists(recorded_event_id))
+                            if recorded_event_id == event_id => {}
+                        Some(_) => return Err(ExecutionError::OracleResponseMismatch),
                     }
+                    txn_tracker.add_oracle_response(OracleResponse::EventExists(event_id));
```
deuszx, line 596: seems like we add it only for the case where
```rust
OracleResponse::EventExists(recorded_event_id))
                            if recorded_event_id == event_id
```

Maybe move that there?

linera-execution/src/execution_state_actor.rs
```diff
@@ -445,10 +446,25 @@ where
...
             }
 
-            ReadEvent { event_id, callback } => {
-                let event = self.context().extra().get_event(event_id.clone()).await?;
-                let event = event.ok_or(ExecutionError::EventsNotFound(vec![event_id]))?;
-                callback.respond(event);
+            ReadEvent {
+                event_id,
+                callback,
+                mut txn_tracker,
```
deuszx, line 452: TransactionTracker is quite big - is it OK to send it over a channel? Can't we verify it on the caller side?

linera-core/src/worker.rs
deuszx: Done that locally - will push soon.

linera-execution/src/system.rs
afck: Good point; I have an issue to clean that up a bit anyway: https://github.com/linera-io/linera-protocol/issues/3519

afck: But it's not true, is it? We also need it in the `None` case? (I.e. when we're not in replay mode at all: Then we have to actually assert the event exists, and then record that.)

linera-execution/src/execution_state_actor.rs
afck: Done in #4288.

PR #4251, opened 2025-07-21
Remove the `next_message_index` as it is not used.

MathieuDutSik:
## Motivation

An examination of the source code reveals that the variable `next_message_index` is never used.

## Proposal

Eliminate it completely.

## Test Plan

The CI.

## Release Plan

## Links

None.


afck (approved):
Good catch! :+1: 
Yes, `next_application_index` and `next_chain_index` have replaced this.

PR #4253, opened 2025-07-22
Send missing events to validators.

afck:
## Motivation

The client will only send a block proposal to the validators if it has all events that were read during execution. Validators that don't have any of the events return an error.

## Proposal

* Unify all missing event errors: `EventsNotFound(Vec<EventId>)`.
* In the `ValidatorUpdater`, when making a proposal, send the publisher chains and retry.

## Test Plan

A client test using the social example was extended. The new test fails without this functionality.

## Release Plan

## Links

- Closes https://github.com/linera-io/linera-protocol/issues/4254.

### DIFF
--- a/linera-core/src/unit_tests/wasm_client_tests.rs
+++ b/linera-core/src/unit_tests/wasm_client_tests.rs
@@ -48,6 +48,7 @@ use crate::{
         ChainClientError,
     },
     local_node::LocalNodeError,
+    test_utils::FaultType,
     worker::WorkerError,
 };
 
@@ -726,9 +727,11 @@ where
 {
     let keys = InMemorySigner::new(None);
     let vm_runtime = VmRuntime::Wasm;
-    let mut builder = TestBuilder::new(storage_builder, 4, 1, keys)
+    let mut builder = TestBuilder::new(storage_builder, 4, 0, keys)
         .await?
         .with_policy(ResourceControlPolicy::all_categories());
+    builder.set_fault_type([3], FaultType::Offline).await;
+
     let sender = builder.add_root_chain(0, Amount::ONE).await?;
     let receiver = builder.add_root_chain(1, Amount::ONE).await?;
 
@@ -757,19 +760,11 @@ where
     let request_subscribe = social::Operation::Subscribe {
         chain_id: sender.chain_id(),
     };
-    let cert = receiver
+    receiver
         .execute_operation(Operation::user(application_id, &request_subscribe)?)
         .await
         .unwrap()
         .unwrap();
-
-    // Subscribe the receiver. This also registers the application.
-    sender.synchronize_from_validators().await.unwrap();
-    sender
-        .receive_certificate_and_update_validators(cert)
-        .await
-        .unwrap();
-    let _certs = sender.process_inbox().await.unwrap();
 
     // Make a post.
     let text = "Please like and comment!".to_string();
@@ -787,6 +782,10 @@ where
         .receive_certificate_and_update_validators(cert.clone())
         .await
         .unwrap();
+
+    builder.set_fault_type([3], FaultType::Honest).await;
+    builder.set_fault_type([2], FaultType::Offline).await;
+
     let certs = receiver.process_inbox().await.unwrap().0;
     assert_eq!(certs.len(), 1);
 
--- a/linera-core/src/updater.rs
+++ b/linera-core/src/updater.rs
@@ -303,7 +303,7 @@ where
                         .iter()
                         .map(|event_id| event_id.chain_id)
                         .filter(|chain_id| !publisher_chain_ids_sent.contains(chain_id))
-                        .collect::<Vec<_>>();
+                        .collect::<BTreeSet<_>>();
                     ensure!(
                         !new_chain_ids.is_empty(),
                         NodeError::EventsNotFound(event_ids)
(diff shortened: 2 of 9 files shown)


linera-execution/src/system.rs
```diff
@@ -576,14 +577,19 @@ where
...
-                        self.context()
-                            .extra()
-                            .contains_event(event_id.clone())
-                            .await?,
-                        ExecutionError::EventNotFound(event_id)
-                    );
+                    if !self
+                        .context()
+                        .extra()
+                        .contains_event(event_id.clone())
```
MathieuDutSik, line 583: It is not related to this PR, but the fact that we have to clone the `event_id` is due to the construction `bcs::to_bytes(&BaseKey::Event(event_id))?;` that requires a clone, while in spirit we should pass a reference.

linera-core/src/worker.rs
```diff
@@ -221,6 +222,8 @@ pub enum WorkerError {
     UnexpectedBlob,
     #[error("Number of published blobs per block must not exceed {0}")]
     TooManyPublishedBlobs(u64),
+    #[error("Missing network description")]
+    MissingNetworkDescription,
```
MathieuDutSik, line 226: Maybe we could indicate the epoch that is missing.

linera-core/src/updater.rs
```diff
@@ -295,6 +297,28 @@ where
                     // synchronize them now and retry.
                     self.send_chain_information_for_senders(chain_id).await?;
                 }
+                Err(NodeError::EventsNotFound(event_ids)) => {
```
MathieuDutSik, line 300: We come here with the same pattern that we have seen before.
Recoverable errors and unrecoverable errors.
Yet, in order to organize the code better, we should have a nested of error like `Result<Result<_,_>,_>`.
But we do not want that. Or maybe we could have a triple
```rust
enum OurResult {
  ok: T,
  recoverable: K,
  unrecoverable: L
}
```

I think we can gain a little by being clearer about this aspect of the code.

linera-core/src/unit_tests/wasm_client_tests.rs
```diff
@@ -757,20 +760,12 @@ where
     let request_subscribe = social::Operation::Subscribe {
         chain_id: sender.chain_id(),
     };
-    let cert = receiver
+    receiver
         .execute_operation(Operation::user(application_id, &request_subscribe)?)
         .await
         .unwrap()
         .unwrap();
```
MathieuDutSik, line 767: The double unwrap is a little unclear here since we are no longer using the returned value.
Therefore, maybe it is better to write
```rust
assert!(matches!(outcome, ClientOutcome::Committed(_)));
```

Maybe several other instances here.

linera-core/src/updater.rs
```diff
@@ -295,6 +297,28 @@ where
...
+                    let mut publisher_heights = BTreeMap::new();
+                    let new_chain_ids = event_ids
+                        .iter()
+                        .map(|event_id| event_id.chain_id)
+                        .filter(|chain_id| !publisher_chain_ids_sent.contains(chain_id))
+                        .collect::<Vec<_>>();
```
MathieuDutSik: Is it guaranteed that the `new_chain_ids` has no duplication?
If not then we could collect the `ChainId` into a `BTreeSet` and avoid repetition in the `local_node.chain_info(chain_id).await` below.

MathieuDutSik (approved):
This looks ok to me.

linera-core/src/worker.rs
afck: The network description doesn't change; it's not per-epoch. If it doesn't exist that basically means storage hasn't been initialized, so this error should be unreachable.

linera-core/src/updater.rs
afck: Yes, but reorganizing our errors will be a huge change, much bigger than this PR.

linera-execution/src/system.rs
afck: I agree, but I'd do that for all `BaseKey` variants and methods, and in a separate PR.

Not sure if `BaseKey` needs `Deserialize` at all? If not, it can be references. If yes, we'd have to use `Cow`s.

linera-core/src/unit_tests/wasm_client_tests.rs
afck: How do you feel about an `unwrap_ok_committed` method for an extension trait of `Result<ClientOutcome<_>>`? That would turn the double `unwrap`s into a single, somewhat more readable line everywhere, and would be more concise than `assert!(matches!(outcome, ClientOutcome::Committed(_)))`.

linera-core/src/updater.rs
afck: Good point! Done in https://github.com/linera-io/linera-protocol/pull/4253/commits/7552a1a582d1e2f43cb15f39b5ffb64c59084beb.

linera-core/src/unit_tests/wasm_client_tests.rs
afck: Done, but I'll make a separate PR for this, too because it's going to be more changed lines than this one.

linera-core/src/updater.rs
MathieuDutSik: Ok, course. Yes, not in this PR

linera-core/src/unit_tests/wasm_client_tests.rs
MathieuDutSik: As you prefer.
I do not like the `fn unwrap()` given that it creates some ambiguity.

PR #4262, opened 2025-07-22
Replace double `unwrap` with `unwrap_ok_committed`.

afck:
## Motivation

Many client methods return `Result<ClientOutcome<_>, _>`. In the tests, we often `unwrap` both the result and the outcome.

## Proposal

Introduce an extension trait and an `unwrap_ok_committed` method to unwrap both at once.

## Test Plan

CI

## Release Plan

## Links

### DIFF
--- a/linera-core/src/data_types.rs
+++ b/linera-core/src/data_types.rs
@@ -363,15 +363,3 @@ impl<T> ClientOutcome<T> {
         }
     }
 }
-
-#[cfg(with_testing)]
-pub trait ClientOutcomeResultExt<T, E> {
-    fn unwrap_ok_committed(self) -> T;
-}
-
-#[cfg(with_testing)]
-impl<T, E: std::fmt::Debug> ClientOutcomeResultExt<T, E> for Result<ClientOutcome<T>, E> {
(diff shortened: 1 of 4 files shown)


deuszx (approved):

linera-core/src/data_types.rs
```diff
@@ -363,3 +363,15 @@ impl<T> ClientOutcome<T> {
...
+pub trait ClientOutcomeResultExt<T, E> {
+    fn unwrap_ok_committed(self) -> T;
+}
+
+#[cfg(with_testing)]
+impl<T, E: std::fmt::Debug> ClientOutcomeResultExt<T, E> for Result<ClientOutcome<T>, E> {
+    fn unwrap_ok_committed(self) -> T {
+        self.unwrap().unwrap()
+    }
+}
```
Twey: Maybe we could write this directly in the `linera-core` tests where it's used?

afck: Moved it to `test_utils` in https://github.com/linera-io/linera-protocol/pull/4262/commits/68a079a72ac8160405756b7bf9c39cc93fdf4d2f.

PR #4263, opened 2025-07-22
Add event skip lists

bart-linera:
## Motivation

We want to be able to leave gaps in chains publishing events, just like we do with sender chains.

## Proposal

We need skip lists for events in blocks, analogous to the ones we have for messages: for each emitted event, have the hash and height of the last block that emitted an event to the same stream.

## Test Plan

CI

## Release Plan

## Links


linera-chain/src/block_tracker.rs
```diff
@@ -351,6 +351,15 @@ impl<'resources, 'blobs> BlockExecutionTracker<'resources, 'blobs> {
             .collect()
     }
 
+    /// Returns stream ids for events published in the block.
```
```suggestion
    /// Returns stream IDs for events published in the block.
```

(Should also be "IDs" in the other comments.)

afck (approved):

PR #4264, opened 2025-07-22
Move metrics to ClientContext

ndr-ds:
## Motivation

Until we have a Control Pane cluster that has a Prometheus Server that we can use to export client metrics to, we need another way to track timings on the client

## Proposal

This PR moves the functionality of the existing timings specific to `benchmark` into `ClientContext`, so we can keep track of more stuff moving forward. It also makes the timings optional. We can add more to this, then make them into Prometheus metrics in the future.

## Test Plan

Ran the client locally, saw the timings printed

## Release Plan

- Nothing to do / These changes follow the usual release cycle.

PR #4219, opened 2025-07-14
Send transactions to every other chain in the chain group

ndr-ds:
## Motivation

Right now every chain sends a cross chain message to one other chain exclusively. So each chain needs to receive just one gRPC message from 1 other chain.

## Proposal

To make the benchmarks more realistic, we need to make it so that all chains send requests to all other chains, generating many more gRPC messages and network traffic.

## Test Plan

Ran locally and saw that things work as expected

## Release Plan

- Nothing to do / These changes follow the usual release cycle.

PR #4265, opened 2025-07-23
Use full client flow on benchmarks

ndr-ds:
## Motivation

We're changing things to use the full client flow from benchmarks. This is part of a whole refactor to make the benchmarks more "realistic".

## Proposal

By using the full flow, we also had to rewrite how we do the client logging of timing metrics. We'll have a timing tokio task running that will manage the histograms, and whoever wants to log the time of something sends a message to this tokio task.
Then every some number of seconds we print that information to the screen for the user.

## Test Plan

Ran locally, saw the metrics being correctly printed

## Release Plan

### DIFF
--- a/linera-client/src/benchmark.rs
+++ b/linera-client/src/benchmark.rs
@@ -13,12 +13,10 @@ use std::{
 use linera_base::{
     data_types::Amount,
     identifiers::{AccountOwner, ApplicationId, ChainId},
-    listen_for_shutdown_signals,
     time::Instant,
 };
 use linera_core::{client::ChainClient, Environment};
 use linera_execution::{
-    committee::Committee,
     system::{Recipient, SystemOperation},
     Operation,
 };
@@ -32,9 +30,7 @@ use tokio::{
 use tokio_util::sync::CancellationToken;
 use tracing::{debug, error, info, warn, Instrument as _};
 
-use crate::client_metrics::{
-    BlockTimeTimings, BlockTimings, ClientMetricsError, SubmitFastBlockProposalTimings,
-};
+use crate::chain_listener::{ChainListener, ClientContext};
 
 const PROXY_LATENCY_P99_THRESHOLD: f64 = 400.0;
 const LATENCY_METRIC_PREFIX: &str = "linera_proxy_request_latency";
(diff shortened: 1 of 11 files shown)


ndr-ds:
* **#4269** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4269?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4267** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4267?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4219** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4219?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4265** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4265?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4265?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#4264** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4264?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4266** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4266?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

linera-client/src/benchmark.rs
```diff
@@ -557,53 +540,11 @@ impl<Env: Environment> Benchmark<Env> {
...
-                .submit_fast_block_proposal(&committee, &operations, &incoming_bundles, chain_owner)
-                .await
-                .map_err(BenchmarkError::ChainClient)?;
-            let submit_fast_block_proposal_ms =
-                submit_fast_block_proposal_start.elapsed().as_millis() as u64;
-            let communicate_chain_updates_start = Instant::now();
-            // We assume the committee will not change during the benchmark.
             chain_client
-                .communicate_chain_updates(&committee)
+                .execute_operations(operations, vec![])
```
afck, line 555: This might not receive any incoming messages, because we are neither regularly synchronizing from the validators, nor listening for notifications.

Explicitly synchronizing is expensive, so I think the right thing to do is to run a chain listener for each client. (Or, even higher-level: run a whole node service, and communicate with it via GraphQL?)

ndr-ds: I see, let me see if I can run a chain listener then

ma2bd: Or we could "relieve" the mail once in a while, instead of listening continuously. We probably want to group the incoming messages any way

ndr-ds: I changed things to run a chain listener, but with `skip_process_inbox` being true, because then the new blocks from the full flow will just pick up the pending bundles, instead of creating a whole other block to process the messages with `process_inbox`. I think that should work reasonably well, right? Then we can iterate and optimize later, if needed 🤔

ma2bd: yes

ma2bd (approved):
LGTM

PR #4268, opened 2025-07-23
Make `ContractRuntime::transfer` synchronous for local transfers.

### LINKED ISSUE #3486: Feature Discussion: Process transfer on the same chain instantly
**Motivation**
Currently when transferring native tokens to another application of the same chain in an operation, the recipient won't receive transferred funds in the same block. It'll be received in next block. If we would like to consume the funds in the target application, we have to create a new message for that. I think it's worth to support process transfer of same chain instantly then we can call target application to consume the funds directly.

**Screenshot**
Image
The image describe what we've tested for that. In the commented part we would like to consume the liquidity funds then create pool directly but funds is not ready. So we need to create another message for that.

MathieuDutSik:
## Motivation

During the implementation of `transfer` operations in the EVM, we face the problem that while the 
`ContractRuntime::transfer` immediately debits the source, and the destination receives the amount
only on the next block.

While this makes complete sense if the destination chain is different from the source chain, there is a use
case for transfer operations within the same chain that are done immediately without emitting messages.

Fixes #3486 

## Proposal

The code of `SystemExecutionStateView::transfer` has been changed so that it does not emit a message if the source chain is the same as the destination chain.

The operations that have been corrected are:
* `Claim`
* `Transfer`
* `Withdraw`

## Test Plan

The CI.

Several tests in `worker_test.rs` had to be changed because the emission of messages had changed.

The end-to-end tests are also unified because now `fungible` and the native token have the same behavior for the emission of messages.

More tests for `Claim`, `Withdraw`, and `Credit` would be useful to do in subsequent PRs.

## Release Plan

## Links

None.

### DIFF
--- a/linera-core/src/unit_tests/worker_tests.rs
+++ b/linera-core/src/unit_tests/worker_tests.rs
@@ -392,15 +392,19 @@ where
         .with_transfer(source, recipient, amount);
         match recipient {
             Recipient::Account(account) => {
-                messages.push(vec![direct_outgoing_message(
-                    account.chain_id,
-                    MessageKind::Tracked,
-                    SystemMessage::Credit {
-                        source,
-                        target: account.owner,
-                        amount,
-                    },
-                )]);
+                if chain_id != account.chain_id {
+                    messages.push(vec![direct_outgoing_message(
+                        account.chain_id,
+                        MessageKind::Tracked,
+                        SystemMessage::Credit {
+                            source,
+                            target: account.owner,
+                            amount,
+                        },
+                    )]);
+                } else {
+                    messages.push(Vec::new());
+                }
             }
             Recipient::Burn => messages.push(Vec::new()),
         }
@@ -884,15 +888,18 @@ where
     let sender_public_key = signer.generate_new();
     let sender_owner = sender_public_key.into();
     let mut env = TestEnvironment::new(storage_builder.build().await?, false, false).await;
-    let chain_desc = env
+    let chain_1_desc = env
         .add_root_chain(1, sender_owner, Amount::from_tokens(5))
         .await;
-    let chain_1 = chain_desc.id();
-    let chain_2 = env.add_root_chain(2, sender_owner, Amount::ZERO).await.id();
+    let chain_1 = chain_1_desc.id();
+    let chain_2_desc = env.add_root_chain(2, sender_owner, Amount::ZERO).await;
+    let chain_2 = chain_2_desc.id();
+    let chain_3_desc = env.add_root_chain(3, sender_owner, Amount::ZERO).await;
+    let chain_3 = chain_3_desc.id();
 
     let certificate0 = env
         .make_simple_transfer_certificate(
-            chain_desc.clone(),
+            chain_1_desc.clone(),
             sender_public_key,
             chain_2,
             Amount::ONE,
@@ -904,9 +911,9 @@ where
 
     let certificate1 = env
         .make_simple_transfer_certificate(
-            chain_desc.clone(),
+            chain_1_desc.clone(),
             sender_public_key,
-            chain_1,
+            chain_3,
             Amount::ONE,
             Vec::new(),
             Amount::from_tokens(3),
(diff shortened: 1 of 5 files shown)


ma2bd:
We could decide that the usual transfer operation (and system call) don’t create a message when the transfer is local.
This should be a simple modification of `fn transfer` in `system.rs`

linera-execution/src/system.rs
```diff
@@ -638,14 +646,31 @@ where
...
             Recipient::Account(account) => {
-                let message = SystemMessage::Credit {
-                    amount,
-                    source,
-                    target: account.owner,
-                };
-                Ok(Some(
-                    OutgoingMessage::new(account.chain_id, message).with_kind(MessageKind::Tracked),
-                ))
+                // Check if destination is on the same chain
```
ma2bd: nit: This comment would be barely needed if we used a more specific name for `chain_id`, like `local_chain_id`, `source_chain_id`, `this_chain_id`.

linera-execution/src/system.rs
```diff
@@ -638,14 +646,31 @@ where
...
+                if account.chain_id == chain_id {
+                    // Handle same-chain transfer locally
```
ma2bd: nit: "Execute" (a transfer) ? (x2)

ma2bd: Note that all the comments in the file end with a `.`

linera-execution/src/system.rs
ma2bd: We could also have `assert_eq!(chain_id, self.context().extra().chain_id);`

MathieuDutSik: Ok.

linera-core/src/unit_tests/worker_tests.rs
```diff
@@ -1846,12 +1856,72 @@ where
         .await?;
     let chain = env.worker().chain_state_view(chain_1).await?;
     assert!(chain.is_active());
-    assert_eq!(Amount::ZERO, *chain.execution_state.system.balance.get());
-    let inbox = chain
+    assert_eq!(Amount::ONE, *chain.execution_state.system.balance.get());
+
+    // With the new optimization, transfers to the same chain should not create messages
+    // but may still have the inbox entry since the transfer happened
```
ma2bd: * Remove "With the new optimization". 
* I don't understand the second part of this comment (remove?). I see `assert!(inbox.is_none());` below.
* Finish sentences with a `.`

MathieuDutSik: Corrected.

linera-execution/src/system.rs
```diff
@@ -647,44 +648,96 @@ where
...
+                        if target == AccountOwner::CHAIN {
+                            let new_balance = self.balance.get().saturating_add(amount);
+                            self.balance.set(new_balance);
+                        } else {
+                            let balance = self.balances.get_mut_or_default(&target).await?;
+                            *balance = balance.saturating_add(amount);
+                        }
+                    } else {
+                        // Recipient is on a different chain, create outgoing Credit message
+                        let message = SystemMessage::Credit {
```
afck: ~Either we shouldn't allow this, or we should merge the `transfer` and `claim` functions (and operations?) into one.~

linera-execution/src/system.rs
```diff
@@ -749,15 +795,23 @@ where
...
+                            // Handle cross-chain operation with message.
+                            let message = SystemMessage::Credit {
+                                amount,
+                                source: owner,
+                                target: account.owner,
+                            };
+                            outcome.push(
+                                OutgoingMessage::new(account.chain_id, message)
+                                    .with_kind(MessageKind::Tracked),
+                            );
```
afck: That logic still seems duplicated, both in `claim` and `transfer`.

MathieuDutSik: Yes, thank you. Refactored.

afck (approved):

afck:
(@ma2bd's status is still "requesting changes". Were all his concerns addressed?)

MathieuDutSik:
> (@ma2bd's status is still "requesting changes". Were all his concerns addressed?)

Yes, as far as I can tell.
There are next steps for this work: test of claims, elimination of Recipient, allowances, etc.

PR #4272, opened 2025-07-24
Merge transactions into a single collection in blocks.

### LINKED ISSUE #1200: Put transactions (operations and messages) in a single collection.
A block should contain a single `Vec<Transaction>` where

```rust
enum Transaction {
    Operation(Operation),
    IncomingMessage(IncomingMessage),
}
```

which can be in any order.

afck:
## Motivation

Blocks contain operations and incoming message bundles as separate collections. It's impossible to process e.g. a message in between two operations. Also, transactions are not properly exposed in GraphQL.

## Proposal

Merge the two collections, expose more information via GraphQL.

(Note: This was almost entirely written using Claude Code.)

## Test Plan

CI; tests have been updated.

## Release Plan

## Links

- Closes #1200 

### DIFF
--- a/linera-service-graphql-client/gql/service_schema.graphql
+++ b/linera-service-graphql-client/gql/service_schema.graphql
@@ -550,25 +550,6 @@ A unique identifier for a user application or for the system application
 """
 scalar GenericApplicationId
 
-type GraphQLOperation {
-	"""
-	The type of operation: "System" or "User"
-	"""
-	operationType: String!
-	"""
-	For user operations, the application ID
-	"""
-	applicationId: ApplicationId
-	"""
-	For user operations, the serialized bytes (as a hex string for GraphQL)
-	"""
-	userBytesHex: String
-	"""
-	For system operations, the serialized bytes (as a hex string for GraphQL)
-	"""
-	systemBytesHex: String
-}
-
 """
 The state of an inbox.
 * An inbox is used to track bundles received and executed locally.
@@ -884,6 +865,25 @@ type MutationRoot {
 Notify that a chain has a new certified block or a new message
 """
 scalar Notification
+
+type Operation {
+	"""
+	The type of operation: "System" or "User"
+	"""
+	operationType: String!
+	"""
+	For user operations, the application ID
+	"""
+	applicationId: ApplicationId
+	"""
+	For user operations, the serialized bytes (as a hex string for GraphQL)
+	"""
+	userBytesHex: String
+	"""
+	For system operations, the serialized bytes (as a hex string for GraphQL)
+	"""
+	systemBytesHex: String
+}
 
 """
 The execution result of a single operation.
@@ -1263,7 +1263,7 @@ type TransactionMetadata {
 	"""
 	The operation, if this is an ExecuteOperation transaction
 	"""
-	operation: GraphQLOperation
+	operation: Operation
 }
 
 scalar VersionInfo
--- a/linera-chain/src/data_types.rs
+++ b/linera-chain/src/data_types.rs
@@ -156,7 +156,8 @@ pub enum Transaction {
 impl BcsHashable<'_> for Transaction {}
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize, SimpleObject)]
-pub struct GraphQLOperation {
+#[graphql(name = "Operation")]
+pub struct OperationMetadata {
     /// The type of operation: "System" or "User"
     pub operation_type: String,
     /// For user operations, the application ID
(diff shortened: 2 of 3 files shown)


linera-chain/src/block.rs
```diff
@@ -364,17 +360,45 @@ pub struct BlockBody {
     pub operation_results: Vec<OperationResult>,
 }
 
+impl BlockBody {
+    /// Returns all operations in this block body.
+    pub fn operations(&self) -> impl Iterator<Item = &Operation> {
+        self.transactions.iter().filter_map(|tx| match tx {
+            Transaction::ExecuteOperation(operation) => Some(operation),
+            Transaction::ReceiveMessages(_) => None,
```
ma2bd, line 372: Should this be called `ReceiveBundle` now?

afck: Maybe; not sure. It's kind of correct since it is already a plural, and maybe "messages" makes it more obvious than "bundle"?

linera-chain/src/chain.rs
```diff
@@ -650,28 +664,29 @@ where
...
         }
-        let origins = bundles_by_origin.keys().copied();
-        let inboxes = self.inboxes.try_load_entries_mut(origins).await?;
+        let origins = bundles_by_origin.keys().copied().collect::<Vec<_>>();
+        let inboxes = self.inboxes.try_load_entries_mut(&origins).await?;
         let mut removed_unskippable = HashSet::new();
         for ((origin, bundles), mut inbox) in bundles_by_origin.into_iter().zip(inboxes) {
             tracing::trace!(
-                "Removing {:?} from {chain_id:.8}'s inbox for {origin:}",
+                "Removing [{}] from {chain_id:.8}'s inbox for {origin:}",
```
deuszx, line 675: I personally prefer if arguments of the log (things that can change) are not inlined to the log line:
```rust
tracing::trace!(
  bundle_heights: bundles.iter()...,
  origin_chain = ?origin,
  "Removing bundles from chain's inbox ....
)
```

not something that has to be done in this PR ofc.

linera-core/src/client/mod.rs
```diff
@@ -1981,11 +1990,19 @@ impl<Env: Environment> ChainClient<Env> {
...
         let info = self.chain_info().await?;
         let timestamp = self.next_timestamp(incoming_bundles, info.timestamp);
+        let transactions = incoming_bundles
+            .iter()
+            .map(|bundle| Transaction::ReceiveMessages(bundle.clone()))
+            .chain(
+                operations
+                    .iter()
+                    .map(|operation| Transaction::ExecuteOperation(operation.clone())),
+            )
```
deuszx, line 2000: Why not do that with a `From<_>` impl and do sth like `incoming_bundles.iter().map(Into::into).chain(...)`.

deuszx: Ah, you have to clone it anyway so it wouldn't be that much more concise.

linera-indexer/lib/src/db/sqlite/mod.rs
```diff
@@ -176,8 +176,8 @@ impl SqliteDatabase {
         })?;
 
         // Count aggregated data
-        let operation_count = block.body.operations.len();
-        let incoming_bundle_count = block.body.incoming_bundles.len();
+        let operation_count = block.body.operations().count();
+        let incoming_bundle_count = block.body.incoming_bundles().count();
```
deuszx, line 180: Hm, but `operations` and `incoming_bundles` no longer exists so how come you're still working with it?

linera-service-graphql-client/gql/service_schema.graphql
```diff
@@ -559,6 +550,25 @@ A unique identifier for a user application or for the system application
 """
 scalar GenericApplicationId
 
+type GraphQLOperation {
```
deuszx: This is the only type that has `GraphQL` prefix – can we get away with it?

linera-indexer/lib/src/db/sqlite/mod.rs
afck: Yes, I didn't make very deep changes to the indexer yet. I actually don't fully understand why it puts the incoming bundles in a separate table but not the operations. (Maybe because the latter need to be indexed by recipients, too?)

deuszx: Maybe it should put the operations in a separate table as well. It's just that messages are grouped into bundles and bundles are what is referenced by the block so there's an additional layer of indirection.

afck: Yes, maybe. But I'd try to avoid redesigning the whole indexer in this PR.

linera-service-graphql-client/gql/service_schema.graphql
afck: I'd go with:

```rust
#[graphql(name = "Operation")]
pub struct OperationMetadata {
```

(Because it's only used in `TransactionMetadata`.)

afck: Done in https://github.com/linera-io/linera-protocol/pull/4272/commits/99399313677fcc9b68d64ab127b89a6fb44b924b.

ma2bd (approved):

PR #4273, opened 2025-07-24
Allow EVM smart contract to create new EVM contracts.

### LINKED ISSUE #3757: Implement EVM create operations.
The `fn create` and `fn create2` are functions of Ethereum smart contracts that allow the creation of new smart contracts.

This is possible because the Inspector trait allows catching create operations, which can then be sent to the Linera runtime.

Now, the creation function in the ContractRuntime is
```rust
    fn create_application(
        &mut self,
        module_id: ModuleId,
        parameters: Vec<u8>,
        argument: Vec<u8>,
        required_application_ids: Vec<ApplicationId>,
    ) -> Result<ApplicationId, ExecutionError>;
```
So, in Linera the application needs first to be created with the blobs for contract/service and then the `create_application` can create an `ApplicationId` blob.

While the creation in the REVM is done with
```rust
    fn create(&mut self, context: &mut CTX, inputs: &mut CreateInputs) -> Option<CreateOutcome> {
```
The `inputs.init_code` contains the bytes of the bytecode and the constructor argument.

Therefore, for the implementation of the creation of contracts, we need to
* Be able from a blob from a contract. This is issue #3707.
* Be able to determine the size of the bytecode. What we have access to is the bytecode concatenated with the constructor argument, and it is unclear how to get the bytecode length.

MathieuDutSik:
## Motivation

It is very frequent for EVM contracts to create another contract. Thus, we need to support this feature.

Fixes #3757 

## Proposal

The implementation is surprisingly straightforward:
* Create the blobs with `publish_module`.
* Create the application with `create_application`.
* Return the result according to the format of the EVM.

The last step requires accessing the bytecode of a smart contract. This is implemented with a direct call.

## Test Plan

In the CI, the test `test_evm_end_to_end_child_subcontract` has been added and shows the generation of the contract and access to them.

## Release Plan

## Links

None.

### DIFF
--- a/linera-execution/src/evm/database.rs
+++ b/linera-execution/src/evm/database.rs
@@ -147,10 +147,7 @@ impl<Runtime: BaseRuntime> DatabaseRuntime<Runtime> {
 
     /// Returns the current storage states and clears it to default.
     pub fn take_storage_stats(&self) -> StorageStats {
-        let mut storage_stats_read = self
-            .storage_stats
-            .lock()
-            .expect("The lock should be possible");
+        let mut storage_stats_read = self.storage_stats.lock().unwrap();
         let storage_stats = storage_stats_read.clone();
         *storage_stats_read = StorageStats::default();
         storage_stats
@@ -158,13 +155,13 @@ impl<Runtime: BaseRuntime> DatabaseRuntime<Runtime> {
 
     /// Insert error into the database
     pub fn insert_error(&self, exec_error: ExecutionError) {
-        let mut error = self.error.lock().expect("The lock should be possible");
+        let mut error = self.error.lock().unwrap();
         *error = Some(format!("Runtime error {:?}", exec_error));
     }
 
     /// Process the error.
     pub fn process_any_error(&self) -> Result<(), EvmExecutionError> {
-        let error = self.error.lock().expect("The lock should be possible");
+        let error = self.error.lock().unwrap();
         if let Some(error) = error.clone() {
             return Err(EvmExecutionError::RuntimeError(error.clone()));
         }
@@ -217,7 +214,7 @@ where
             let account = self.changes.get(&address).unwrap();
             return Ok(Some(account.info.clone()));
         }
-        let mut runtime = self.runtime.lock().expect("The lock should be possible");
+        let mut runtime = self.runtime.lock().unwrap();
         let key_info = self.get_address_key(KeyCategory::AccountInfo as u8, address);
         let promise = runtime.read_value_bytes_new(key_info)?;
         let result = runtime.read_value_bytes_wait(&promise)?;
@@ -240,14 +237,11 @@ where
         let key_prefix = self.get_address_key(KeyCategory::Storage as u8, address);
         let key = Self::get_linera_key(&key_prefix, index)?;
         {
-            let mut storage_stats = self
-                .storage_stats
-                .lock()
-                .expect("The lock should be possible");
+            let mut storage_stats = self.storage_stats.lock().unwrap();
             storage_stats.key_read += 1;
         }
         let result = {
-            let mut runtime = self.runtime.lock().expect("The lock should be possible");
+            let mut runtime = self.runtime.lock().unwrap();
             let promise = runtime.read_value_bytes_new(key)?;
             runtime.read_value_bytes_wait(&promise)
         }?;
@@ -265,11 +259,8 @@ where
 {
     /// Effectively commits changes to storage.
     pub fn commit_changes(&mut self) -> Result<(), ExecutionError> {
-        let mut storage_stats = self
-            .storage_stats
-            .lock()
-            .expect("The lock should be possible");
-        let mut runtime = self.runtime.lock().expect("The lock should be possible");
+        let mut storage_stats = self.storage_stats.lock().unwrap();
+        let mut runtime = self.runtime.lock().unwrap();
         let mut batch = Batch::new();
         for (address, account) in &self.changes {
             if !account.is_touched() {
@@ -354,7 +345,7 @@ where
     /// Sets the EVM contract address from the value Address::ZERO.
     /// The value is set from the `ApplicationId`.
     pub fn set_contract_address(&mut self) -> Result<(), ExecutionError> {
-        let mut runtime = self.runtime.lock().expect("The lock should be possible");
+        let mut runtime = self.runtime.lock().unwrap();
         let application_id = runtime.application_id()?;
         self.contract_address = application_id_to_address(application_id);
         Ok(())
@@ -367,7 +358,7 @@ where
     /// We determine whether the contract is already initialized, sets the
     /// `is_revm_initialized` and then returns the result.
     pub fn set_is_initialized(&mut self) -> Result<bool, ExecutionError> {
-        let mut runtime = self.runtime.lock().expect("The lock should be possible");
+        let mut runtime = self.runtime.lock().unwrap();
         let evm_address = runtime.application_id()?.evm_address();
         let key_info = self.get_address_key(KeyCategory::AccountInfo as u8, evm_address);
         let promise = runtime.contains_key_new(key_info)?;
@@ -377,7 +368,7 @@ where
     }
 
     pub fn get_block_env(&self) -> Result<BlockEnv, ExecutionError> {
-        let mut runtime = self.runtime.lock().expect("The lock should be possible");
+        let mut runtime = self.runtime.lock().unwrap();
         // The block height being used
         let block_height_linera = runtime.block_height()?;
         let block_height_evm = block_height_linera.0;
@@ -420,7 +411,7 @@ where
     }
 
     pub fn constructor_argument(&self) -> Result<Vec<u8>, ExecutionError> {
-        let mut runtime = self.runtime.lock().expect("The lock should be possible");
+        let mut runtime = self.runtime.lock().unwrap();
         let constructor_argument = runtime.application_parameters()?;
         Ok(serde_json::from_slice::<Vec<u8>>(&constructor_argument)?)
     }
@@ -432,7 +423,7 @@ where
 {
     pub fn get_contract_block_env(&self) -> Result<BlockEnv, ExecutionError> {
         let mut block_env = self.get_block_env()?;
-        let mut runtime = self.runtime.lock().expect("The lock should be possible");
+        let mut runtime = self.runtime.lock().unwrap();
         // We use the gas_limit from the runtime
         let gas_limit = runtime.maximum_fuel_per_block(VmRuntime::Evm)?;
         block_env.gas_limit = gas_limit;


deuszx:
CI fails due to unused imports - it doesn't run the tests you mention.

linera-execution/src/evm/revm.rs
```diff
@@ -51,6 +51,9 @@ const PROCESS_STREAMS_SELECTOR: &[u8] = &[227, 9, 189, 153];
 /// only when creating a new instance of a shared contract
 const INSTANTIATE_SELECTOR: &[u8] = &[156, 163, 60, 158];
 
+/// Returns the deployed bytecode of the contract.
```
deuszx: Can you expain more in-depth why the sequence of these bytes is valid for getting the deployed bytecode?

MathieuDutSik: The calls in Ethereum are based on selectors.

Here we choose an arbitrary selector (it is part of Fibonacci sequence) and ask the contract to return the deployed bytecode.

This he does as is apparent at the start of the execute_operation.

linera-execution/src/evm/database.rs
```diff
@@ -78,6 +81,8 @@ pub(crate) struct DatabaseRuntime<Runtime> {
     pub runtime: Arc<Mutex<Runtime>>,
     /// The uncommitted changes to the contract.
     pub changes: EvmState,
+    /// Whether the contract has been instantiated in REVM.
```
ma2bd, line 84: Do you mean the application? As opposed to instantiated in a block?

linera-execution/src/evm/revm.rs
```diff
@@ -1055,6 +1130,24 @@ impl<'a, Runtime: ServiceRuntime> Inspector<Ctx<'a, Runtime>> for CallIntercepto
...
+        &mut self,
+        _context: &mut Ctx<'_, Runtime>,
+        inputs: &mut CreateInputs,
+    ) -> Result<Option<CreateOutcome>, ExecutionError> {
+        if !self.db.is_revm_instantiated {
+            self.db.is_revm_instantiated = true;
+            inputs.scheme = CreateScheme::Custom {
+                address: self.contract_address,
+            };
+            Ok(None)
```
ma2bd, line 1222: how come this doesn't return an error?

linera-execution/src/evm/revm.rs
```diff
@@ -54,6 +54,17 @@ const PROCESS_STREAMS_SELECTOR: &[u8] = &[254, 72, 102, 28];
...
+/// The selector when calling for `InterpreterResult`. This is a fictional
+/// selector that does not correspond to a real function.
+const INTERPRETER_RESULT_SELECTOR: &[u8] = &[1, 2, 3, 4];
+
+/// The selector when accessing for the deployed bytecode. This is a fictional
+/// selector that does not correspond to a real function.
+const GET_DEPLOYED_BYTECODE_SELECTOR: &[u8] = &[21, 34, 55, 89];
+
+/// The serde serialization of a trivial vector.
+const SERDE_EMPTY_VECTOR: &[u8] = &[91, 93];
```
ma2bd: Do you mean BCS?

linera-execution/src/evm/revm.rs
```diff
@@ -884,6 +891,31 @@ impl<'a, Runtime: ServiceRuntime> PrecompileProvider<Ctx<'a, Runtime>> for Servi
...
 }
 
+fn map_result_create_outcome(
+    result: Result<Option<CreateOutcome>, ExecutionError>,
+) -> Option<CreateOutcome> {
+    match result {
+        Err(_error) => {
+            // An alternative way would be to return None, which would induce
+            // Revm to call the smart contract in its database, where it is
+            // non-existent.
```
ma2bd: If we're going to have a comment, maybe it's better explain what we're actually doing and why?

afck:
> The solution is possibly to add access to the array `created_blobs` in `read_data_blob` and `assert_data_blob`. However, this does not suffice, and one has to add the `txn_tracker` to the `create_application` and `check_bytecode_blobs`. Where exactly is the limit here? This is to be investigated.

I'm not sure I understand; didn't you recently make that work for all blobs? If not, I'd say it should.

deuszx:
Honestly, after reading the description I am not sure this fully implements the described feature:

In the **Proposal** section you write:

> However, this does not suffice, and one has to add the txn_tracker to the create_application and check_bytecode_blobs. Where exactly is the limit here? This is to be investigated.

And in the **Test Plan**:

> This needs to be corrected and much remains to be done.

-----------
So, is it implemented or not?

linera-execution/src/evm/revm.rs
MathieuDutSik: No.
If you have a `Vec::new()` and you serialize it with `serde_json` then the serialization is going to be "[]" whose sequence is `[91, 93]` because 91 is the ascii character of "[" and 93 the ascii character of "]".

MathieuDutSik:
> > The solution is possibly to add access to the array `created_blobs` in `read_data_blob` and `assert_data_blob`. However, this does not suffice, and one has to add the `txn_tracker` to the `create_application` and `check_bytecode_blobs`. Where exactly is the limit here? This is to be investigated.
> 
> I'm not sure I understand; didn't you recently make that work for all blobs? If not, I'd say it should.

No, I limited myself to the features I needed for my purpose.

MathieuDutSik:
> > This needs to be corrected and much remains to be done.
> 
> So, is it implemented or not?

In my opinion, it is since I have updated my PR but forgot to update the PR description.

linera-execution/src/evm/revm.rs
MathieuDutSik: I know.
In the other PR for the balance I implement a better scheme that allows to propagate the errors.

Let me put it here.

linera-execution/src/evm/revm.rs
MathieuDutSik: This is the way that EVM contracts are created in Linera:
* This is the `Inspector` trait. If we return `None`, then it means that REVM has to do the job. If returning `Some(_)` then we have done the initialization as we want. So, by returning `None` we let REVM do the creation of the contract.
* But there is another way to change the way the intialization is done: By altering the `CreateInputs`.
* So, we set the address to the one of the contracts as computed from the `ApplicationDescription`.

Now, this is services:
* The first initialization is the one where the contract itself is created. So, we need it.
* But for the other calls to creation, it means that another contract is being created. So, this is not a service call.

linera-execution/src/evm/revm.rs
MathieuDutSik: Ok, done.

linera-execution/src/evm/database.rs
MathieuDutSik: We have the following notions:
* In the EVM, when a contract is being created, you submit the code and the constructor arguments.
* The creation itself leads to some keys in storage and the deployed code being created. Those are in `changes: EvmState`.
* The creation of a contract is needed for contract and service calls. If you call a service on a chain in which it is not instantiated, then the creation is done via the REVM. But since it is not in contract and so cannot be written in storage, it stays in storage. But that is fine. We can respond to the service call.
* This PR is about the creation of contracts. But the `fn create` will catch the creation of the application itself and the creation of other contracts. And so we need to catch the creation of a contract with a boolean `is_revm_instantiated`.

linera-service/tests/linera_net_tests.rs
```diff
@@ -511,6 +511,120 @@ async fn test_evm_end_to_end_counter(config: impl LineraNetConfig) -> Result<()>
...
+        index: U256::from(1),
+    };
+    let query1 = EvmQuery::Query(query1.abi_encode());
+    let address1 = application.run_json_query(query1).await?;
+    let address1 = read_evm_address_entry(address1);
+    assert_ne!(address0, address1);
+
+    // Creating the applications
+
+    let application0: ApplicationId = address0.into();
```
ma2bd: nit: `let application0 = ApplicationId::from(address0);`

linera-service/tests/linera_net_tests.rs
```diff
@@ -511,6 +511,120 @@ async fn test_evm_end_to_end_counter(config: impl LineraNetConfig) -> Result<()>
...
+    let application0: ApplicationId<EvmAbi> = application0.with_abi();
```
ma2bd: or better now: `let application0 = ApplicationId::from(address0).with_abi::<EvmAbi>();`

linera-service/tests/linera_net_tests.rs
MathieuDutSik: Ok.

linera-service/tests/linera_net_tests.rs
MathieuDutSik: Thanks.

linera-execution/src/evm/revm.rs
```diff
@@ -977,11 +1007,58 @@ impl<'a, Runtime: ContractRuntime> Inspector<Ctx<'a, Runtime>>
...
+            Ok(None)
+        } else {
+            let contract = linera_base::data_types::Bytecode::new(inputs.init_code.to_vec());
+            let service = linera_base::data_types::Bytecode::new(vec![]);
+            let mut runtime = context
+                .db()
+                .0
+                .runtime
+                .lock()
+                .expect("The lock should be possible");
```
ma2bd: `unwrap()`

linera-execution/src/evm/database.rs
```diff
@@ -323,12 +362,13 @@ where
 
     /// Checks if the contract is already initialized. It is possible
     /// that the constructor has not yet been called.
-    pub fn is_initialized(&self) -> Result<bool, ExecutionError> {
+    pub fn is_initialized(&mut self) -> Result<bool, ExecutionError> {
```
MathieuDutSik: Change the name of the function, as it actually mutates the state.

ma2bd (approved):

linera-execution/src/evm/revm.rs
ma2bd: Then, how about `JSON_EMPTY_VECTOR` ?

MathieuDutSik: Ok.

MathieuDutSik:
> > The solution is possibly to add access to the array `created_blobs` in `read_data_blob` and `assert_data_blob`. However, this does not suffice, and one has to add the `txn_tracker` to the `create_application` and `check_bytecode_blobs`. Where exactly is the limit here? This is to be investigated.
> 
> I'm not sure I understand; didn't you recently make that work for all blobs? If not, I'd say it should.

Yes, it has been done. I updated the description.

linera-execution/src/evm/revm.rs
MathieuDutSik: Ok, done all over the code of `revm.rs` and `database.rs`.

linera-execution/src/evm/database.rs
MathieuDutSik: Done.

linera-execution/src/evm/revm.rs
```diff
@@ -977,11 +992,130 @@ impl<'a, Runtime: ContractRuntime> Inspector<Ctx<'a, Runtime>>
...
+    ) -> Result<Option<CreateOutcome>, ExecutionError> {
+        if !self.db.is_revm_instantiated {
+            self.db.is_revm_instantiated = true;
+            inputs.scheme = CreateScheme::Custom {
+                address: self.contract_address,
+            };
+            Ok(None)
+        } else {
+            let contract = linera_base::data_types::Bytecode::new(inputs.init_code.to_vec());
+            let service = linera_base::data_types::Bytecode::new(vec![]);
```
afck, line 1090: We should probably make `service` an `Option` in `publish_module`.

MathieuDutSik: Here, it really does not matter.
The `publish_module` is called with `VmRuntime::Evm`.
Therefore, in the `create_bytecode_blobs_sync`, only the contract bytecode gets compressed.
I presume that the compiler is able to see that and not build the `service` bytecode.

PR #4277, opened 2025-07-25
Separate the traits AdminKeyValueStore and KeyValueStore

### LINKED ISSUE #4275: Separate the trait AdminKeyValueStore and KeyValueStore
After the refactoring `AdminKeyValueStore` (to be renamed `KeyValueDatabase`) will no longer extend `KeyValueStore`, forcing user to effectively pass a partition key (aka "root key") and choose between the "exclusive" and the "shared" mode.

ma2bd:
## Motivation

fixes #4275 

## Proposal

* Rename `AdminKeyValueStore` into `KeyValueDatabase` and make it a separate trait
* Keep the errors shared
* Do not rename more things for now

## Test Plan

CI

## Release Plan


linera-storage-service/src/client.rs
```diff
@@ -491,6 +500,10 @@ impl AdminKeyValueStore for StorageServiceStoreInternal {
         })
     }
 
+    fn open_shared(&self, root_key: &[u8]) -> Result<Self::Store, Self::Error> {
+        self.open_exclusive(root_key)
+    }
```
MathieuDutSik: That looks like a bug.
If not, then a rename of functions is needed.

linera-storage-service/src/server.rs
```diff
@@ -661,10 +667,11 @@ async fn main() {
                 inner_config,
                 storage_cache_config,
             };
-            let store = RocksDbStore::maybe_create_and_connect(&config, &namespace)
+            let database = RocksDbDatabase::maybe_create_and_connect(&config, &namespace)
                 .await
                 .expect("store");
-            let store = LocalStore::RocksDb(store);
+            let store_impl = database.open_shared(&[]).expect("Failed to open store");
```
MathieuDutSik: Usually, we do not use stuff like `store_impl` where the `_impl` does not add anything.

linera-storage-service/tests/store_test.rs
```diff
@@ -4,10 +4,10 @@
 #![cfg(feature = "storage-service")]
 
 use anyhow::Result;
-use linera_storage_service::client::{StorageServiceStore, StorageServiceStoreInternal};
+use linera_storage_service::client::StorageServiceDatabaseInternal;
 use linera_views::{
     batch::Batch,
-    store::TestKeyValueStore as _,
+    store::TestKeyValueDatabase,
```
MathieuDutSik: What is wrong about letting the `as _` which is a little helpful.

linera-storage/src/db_storage.rs
```diff
@@ -895,16 +913,17 @@ where
...
+        let store = self.database.open_exclusive(&root_key)?;
         Ok(ViewContext::create_root_context(store, block_exporter_id).await?)
     }
 }
 
-impl<Store, C> DbStorage<Store, C>
+impl<Database, C> DbStorage<Database, C>
 where
-    Store: KeyValueStore + Clone + Send + Sync + 'static,
+    Database: KeyValueDatabase + Clone + Send + Sync + 'static,
```
MathieuDutSik, line 923: Is there a single instance where a `KeyValueDatabase` does not have a corresponding `::Store: KeyValueStore`?

linera-views/src/backends/dual.rs
```diff
@@ -7,15 +7,26 @@ use serde::{Deserialize, Serialize};
...
+/// A dual database.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct DualDatabase<D1, D2, A> {
+    /// The first database.
+    pub first_database: D1,
+    /// The second database.
+    pub second_database: D2,
+    /// Marker for the static root key assignment.
+    _marker: std::marker::PhantomData<A>,
+}
```
MathieuDutSik, line 28: So no rename `first`-> `shared` and `second`-> `exclusive`.

linera-views/src/backends/dual.rs
```diff
@@ -226,59 +220,79 @@ where
...
+            }
+            StoreInUse::Second => {
+                let store = self
+                    .second_database
+                    .open_exclusive(root_key)
+                    .map_err(DualStoreError::Second)?;
+                Ok(DualStore::Second(store))
+            }
+        }
+    }
```
MathieuDutSik: So, now the dual store notion has become a notion orthogonal to the notion of shared/exclusive.
This is a generalization, but my question is why?
Usually, we want to introduce a new notion for some reason.

linera-views/src/backends/dynamo_db.rs
```diff
@@ -343,32 +353,23 @@ impl AdminKeyValueStore for DynamoDbStoreInternal {
...
-            root_key_written: Arc::new(AtomicBool::new(false)),
-        })
+    fn open_exclusive(&self, root_key: &[u8]) -> Result<Self::Store, DynamoDbStoreInternalError> {
+        let mut start_key = EMPTY_ROOT_KEY.to_vec();
+        start_key.extend(root_key);
+        self.open_internal(start_key)
+    }
+
+    fn open_shared(&self, root_key: &[u8]) -> Result<Self::Store, DynamoDbStoreInternalError> {
+        self.open_exclusive(root_key)
```
MathieuDutSik: If the opening of a shared one is the same as opening an exclusive one, then we should have an `open_store` and then both `open_shared/open_exclusive` that refer to it.

linera-views/src/backends/scylla_db.rs
```diff
@@ -743,21 +759,32 @@ impl AdminKeyValueStore for ScyllaDbStoreInternal {
...
+        Ok(ScyllaDbStoreInternal {
+            store,
+            semaphore,
+            max_stream_queries,
             root_key,
         })
     }
 
-    fn open_exclusive(&self, root_key: &[u8]) -> Result<Self, ScyllaDbStoreInternalError> {
+    fn open_shared(&self, root_key: &[u8]) -> Result<Self::Store, ScyllaDbStoreInternalError> {
```
MathieuDutSik: Here the implementation of `open_shared` is the same as `open_exclusive`.
That's a little unfortunate to have code paste.

linera-views/src/backends/value_splitting.rs
```diff
@@ -454,22 +475,20 @@ impl WritableKeyValueStore for LimitedTestMemoryStore {
...
 #[cfg(with_testing)]
 impl LimitedTestMemoryStore {
     /// Creates a `LimitedTestMemoryStore`
     pub fn new() -> Self {
-        let namespace = generate_test_namespace();
-        let store =
-            MemoryStore::new_for_testing(TEST_MEMORY_MAX_STREAM_QUERIES, &namespace).unwrap();
-        LimitedTestMemoryStore { store }
+        let inner = MemoryStore::new_for_testing();
+        LimitedTestMemoryStore { inner }
```
MathieuDutSik, line 491: What is a little puzzling here is that the maximal size is not showing up.
The tests show that the splitting is still working correctly. But then is the splitting till occurring for the `MemoryStore`.

linera-views/src/test_utils/performance.rs
```diff
@@ -220,8 +238,13 @@ where
...
 /// Benchmarks the `write_batch` operation.
-pub async fn write_batch<S: TestKeyValueStore>(iterations: u64) -> Duration {
-    let store = S::new_test_store().await.unwrap();
+pub async fn write_batch<D>(iterations: u64) -> Duration
+where
+    D: TestKeyValueDatabase,
+    D::Store: KeyValueStore,
+{
+    let namespace = D::connect_test_namespace().await.unwrap();
+    let store = namespace.open_shared(&[]).unwrap();
```
MathieuDutSik: I think we could have `D::new_tesz_store(&[]).await.unwrap()` per the trait definition.

linera-views/src/test_utils/performance.rs
```diff
@@ -5,7 +5,7 @@ use std::time::{Duration, Instant};
 
 use crate::{
     batch::Batch,
-    store::{KeyValueStore, TestKeyValueStore},
+    store::{KeyValueStore, ReadableKeyValueStore, TestKeyValueDatabase, WritableKeyValueStore},
```
MathieuDutSik: Maybe add `as _` when possible?

linera-web/src/lib.rs
```diff
@@ -35,18 +35,19 @@ use crate::signer::JsSigner;
...
 
 type WebEnvironment =
     linera_core::environment::Impl<WebStorage, linera_rpc::node_provider::NodeProvider, JsSigner>;
 
 type JsResult<T> = Result<T, JsError>;
 
-async fn get_storage() -> Result<WebStorage, <linera_views::memory::MemoryStore as WithError>::Error>
-{
+async fn get_storage(
+) -> Result<WebStorage, <linera_views::memory::MemoryDatabase as WithError>::Error> {
```
MathieuDutSik, line 46: Here the code is not a generic one. So, why use a complex definition and not `MemoryStoreError`?

linera-base/src/crypto/signer.rs
```diff
@@ -80,6 +80,13 @@ mod in_mem {
     #[derive(Clone)]
     pub struct InMemorySigner(Arc<RwLock<InMemSignerInner>>);
 
+    #[cfg(not(with_getrandom))]
+    impl Default for InMemorySigner {
```
MathieuDutSik, line 84: Why exactly the default trait should be implemented only in that case for the `InMemorySigner`?
The restriction `with_getrandom` seems arbitrary.

linera-storage-service/src/client.rs
ma2bd: There is just no difference in the implementation in this case. What we could do is switch the roles, though. (Define `open_exclusive` as a function of `open_shared`)

linera-storage-service/src/server.rs
ma2bd: thx. I missed that one

linera-views/src/backends/dual.rs
ma2bd: In a next PR. It wouldn't be a just rename

linera-views/src/backends/dual.rs
ma2bd: This was the case before this PR.

linera-storage/src/db_storage.rs
ma2bd: `DynamoDbStoreInternal` and `ScyllaDbStoreInternal` implement `DirectKeyValueStore`

linera-storage-service/tests/store_test.rs
ma2bd: Let me check

linera-views/src/backends/scylla_db.rs
ma2bd: yes we can change that

linera-views/src/test_utils/performance.rs
ma2bd: good catch

linera-views/src/backends/value_splitting.rs
ma2bd: That's `WritableKeyValueStore::MAX_VALUE_SIZE` (not changed in the PR)

linera-web/src/lib.rs
ma2bd: I didn't write that but maybe @Twey thought it's easier to maintain that way.

linera-base/src/crypto/signer.rs
ma2bd: This is solving a clippy warning. In the other case, `new` has an extra argument!

linera-views/src/backends/lru_caching.rs
```diff
@@ -467,113 +485,106 @@ pub struct LruCachingConfig<C> {
...
+        );
+        Ok(store)
     }
 
-    fn open_exclusive(&self, root_key: &[u8]) -> Result<Self, Self::Error> {
-        let store = self.store.open_exclusive(root_key)?;
-        let store = LruCachingStore::new(store, self.storage_cache_config());
-        store.enable_exclusive_access();
+    fn open_shared(&self, root_key: &[u8]) -> Result<Self::Store, Self::Error> {
+        let store = self.database.open_exclusive(root_key)?;
```
ma2bd, line 519: oops

linera-views/src/backends/metering.rs
```diff
@@ -427,37 +443,45 @@ where
...
+    fn open_exclusive(&self, root_key: &[u8]) -> Result<Self::Store, Self::Error> {
+        let _latency = self.counter.open_exclusive_latency.measure_latency();
+        let store = self.database.open_exclusive(root_key)?;
+        let counter = self.counter.clone();
+        Ok(MeteredStore { counter, store })
     }
 
-    fn open_exclusive(&self, root_key: &[u8]) -> Result<Self, Self::Error> {
+    fn open_shared(&self, root_key: &[u8]) -> Result<Self::Store, Self::Error> {
         let _latency = self.counter.open_exclusive_latency.measure_latency();
```
ma2bd, line 480: oops

linera-storage-service/src/client.rs
ma2bd: I swapped the definitions

linera-storage-service/tests/store_test.rs
ma2bd: it worked

linera-views/src/backends/dynamo_db.rs
ma2bd: see other comment

linera-views/src/test_utils/performance.rs
ma2bd: done

PR #4278, opened 2025-07-26
Provide the access to `created_blobs` for the contract calls.

MathieuDutSik:
## Motivation

In the EVM framework of doing operation, we need to be able to do the operations
```rust
let module_id = runtime.publish_module(....
let application_id = runtime.create_application(module_id, ....)
runtime.call_application(application_id, ....)
```

in the same block. This is currently not possible, and this PR addresses that problem from the protocol viewpoint. A Wasm test is provided, and the EVM should follow thereafter.

## Proposal

* When the `create_application` is made, the blobs are checked for existence. Given that those blobs currently exist only in `created_blobs`, we have to provide access to them.
* When the `runtime.call_application` is being made, the corresponding contract may exist only in `created_blobs`, so again we provide access to it.

We also provide this for the service, since a contract can call its service during its instantiation (one example is `ethereum-tracker`). However, the service will see the state before the instantiation, not in the block that is currently being constructed.

## Test Plan

The CI.

A new test, `create-and-call`, has been introduced that demonstrates those capabilities.

## Release Plan

## Links

None.

### DIFF
--- a/examples/create-and-call/README.md
+++ b/examples/create-and-call/README.md
@@ -6,8 +6,7 @@ in the `test_create_and_call_end_to_end` end-to-end test.
 
 ## Note
 
-In the block in which `publish_module`, `create_application` and `call_application` are done,
-we cannot call the `query_application` cannot be called since the service are about the previous
-block.
+In the operation in which `publish_module`, `create_application` and `call_application` are done,
+we can call `query_application` but it uses default values.
 
 ## Usage


examples/create-and-call/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+This example demonstrates an example where a contract creates another contract and calls it.
+The example application being created is the counter-no-graphql. The deployment can be seen
+in the `test_create_and_call_end_to_end` end-to-end test.
+
+## Note
+
+In the block in which `publish_module`, `create_application` and `call_application` are done,
+we cannot call the `query_application` cannot be called since the service are about the previous
+block.
```
bart-linera: TBH, I'm not sure what you mean here, could you clarify?

examples/create-and-call/src/contract.rs
```diff
@@ -0,0 +1,94 @@
...
+        let CreateAndCallOperation::CreateAndCall(
+            contract_bytes,
+            service_bytes,
+            initialization_value,
+            increment_value,
+        ) = operation;
+
+        // Step 1: Convert Vec<u8> to Bytecode and publish module with Wasm runtime
+        let contract_bytecode = Bytecode::new(contract_bytes);
+        let service_bytecode = Bytecode::new(service_bytes);
```
bart-linera, line 59: It could be a good idea to publish the bytecodes separately, like with a regular application, and only pass the ModuleId here - although I guess it is nice to know that this way works, too.

examples/create-and-call/src/contract.rs
```diff
@@ -0,0 +1,94 @@
...
+        // Step 3: Call the service. It should return the value before
+        // the initialization of this contract and thus zero.
+        let counter_request = CounterRequest::Query;
+        let value = self.runtime.query_service(application_id, counter_request);
+        assert_eq!(value, 0);
+
+        // Step 4: Call the contract with counter increment operation
+        let counter_operation = CounterOperation::Increment(increment_value);
+        self.runtime
+            .call_application(true, application_id, &counter_operation)
```
bart-linera, line 84: ~~Maybe we could try querying again and asserting that the value increased?~~ 

EDIT: Nevermind, I see you wrote this in the PR description:

> However, the service will see the state before the instantiation, not in the block that is currently being constructed.

So I guess we can't do that. (But maybe it would make sense to add such a comment in the contract code, too?)

examples/create-and-call/README.md
MathieuDutSik: Ok, done.

examples/create-and-call/src/contract.rs
MathieuDutSik: The scenario of this test is what is needed for the EVM application.
It is a more challenging test, but I think a good thing to do.

examples/create-and-call/src/contract.rs
MathieuDutSik: This has been added.

PR #4283, opened 2025-07-28
Update validators less eagerly - fewer queries when proposing a block

deuszx:
## Motivation

We've noticed that in our app demos we make more than optimal # of requests when proposing a block. Obviously we should try to do as little work as possible.

## Test Plan

CI

## Release Plan

## Links

### DIFF
--- a/linera-base/src/prometheus_util.rs
+++ b/linera-base/src/prometheus_util.rs
@@ -5,8 +5,8 @@
 
 use prometheus::{
     exponential_buckets, histogram_opts, linear_buckets, register_histogram,
-    register_histogram_vec, register_int_counter_vec, register_int_gauge_vec, Histogram,
-    HistogramVec, IntCounterVec, IntGaugeVec, Opts,
+    register_histogram_vec, register_int_counter, register_int_counter_vec, register_int_gauge_vec,
+    Histogram, HistogramVec, IntCounter, IntCounterVec, IntGaugeVec, Opts,
 };
 
 use crate::time::Instant;
@@ -21,6 +21,12 @@ pub fn register_int_counter_vec(
 ) -> IntCounterVec {
     let counter_opts = Opts::new(name, description).namespace(LINERA_NAMESPACE);
     register_int_counter_vec!(counter_opts, label_names).expect("IntCounter can be created")
+}
+
+/// Wrapper around Prometheus register_int_counter_vec! macro which also sets the `linera` namespace
+pub fn register_int_counter(name: &str, description: &str) -> IntCounter {
+    let counter_opts = Opts::new(name, description).namespace(LINERA_NAMESPACE);
+    register_int_counter!(counter_opts).expect("IntCounter can be created")
 }
 
 /// Wrapper around Prometheus `register_histogram_vec!` macro which also sets the `linera` namespace
--- a/linera-base/src/prometheus_util.rs
+++ b/linera-base/src/prometheus_util.rs
@@ -13,7 +13,7 @@ use crate::time::Instant;
 
 const LINERA_NAMESPACE: &str = "linera";
 
-/// Wrapper around Prometheus register_int_counter_vec! macro which also sets the `linera` namespace
+/// Wrapper around Prometheus `register_int_counter_vec!` macro which also sets the `linera` namespace
 pub fn register_int_counter_vec(
     name: &str,
     description: &str,
@@ -23,7 +23,7 @@ pub fn register_int_counter_vec(
     register_int_counter_vec!(counter_opts, label_names).expect("IntCounter can be created")
 }
 
-/// Wrapper around Prometheus register_int_counter_vec! macro which also sets the `linera` namespace
+/// Wrapper around Prometheus `register_int_counter!` macro which also sets the `linera` namespace
 pub fn register_int_counter(name: &str, description: &str) -> IntCounter {
     let counter_opts = Opts::new(name, description).namespace(LINERA_NAMESPACE);
     register_int_counter!(counter_opts).expect("IntCounter can be created")
--- a/linera-client/src/chain_listener.rs
+++ b/linera-client/src/chain_listener.rs
@@ -218,7 +218,6 @@ impl<C: ClientContext> ChainListener<C> {
             }
             Reason::NewRound { .. } => self.update_validators(&notification).await?,
             Reason::NewBlock { hash, .. } => {
-                // self.update_validators(&notification).await?;
                 self.update_wallet(notification.chain_id).await?;
                 self.add_new_chains(*hash).await?;
                 let publishers = self
--- a/linera-client/src/chain_listener.rs
+++ b/linera-client/src/chain_listener.rs
@@ -218,7 +218,7 @@ impl<C: ClientContext> ChainListener<C> {
             }
             Reason::NewRound { .. } => self.update_validators(&notification).await?,
             Reason::NewBlock { hash, .. } => {
-                self.update_validators(&notification).await?;
+                // self.update_validators(&notification).await?;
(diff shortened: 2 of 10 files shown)


linera-core/src/client/mod.rs
```diff
@@ -2825,9 +2826,12 @@ impl<Env: Environment> ChainClient<Env> {
         // Send the query to validators.
         let certificate = if round.is_fast() {
             let hashed_value = ConfirmedBlock::new(block);
-            self.client
+            let certificate = self
+                .client
                 .submit_block_proposal(&committee, proposal, hashed_value)
-                .await?
+                .await?;
+            self.update_validators(Some(&committee)).await?;
```
deuszx, line 2829: For non-fast rounds we don't need to update validators explicitly since we already call `finalize_block(confirmed_certificate)`.

linera-core/src/updater.rs
```diff
@@ -495,24 +495,55 @@ where
                 height, chain_id, ..
             } => (*height, *chain_id),
         };
-        // Update the validator with missing information, if needed.
-        let delivery = CrossChainMessageDelivery::NonBlocking;
-        self.send_chain_information(chain_id, target_block_height, delivery)
-            .await?;
```
deuszx, line 501: Rather then updating validators eagerly, we catch errors and - if they signal that validators is missing some data - only then we update them.

linera-client/src/chain_listener.rs
```diff
@@ -218,7 +218,7 @@ impl<C: ClientContext> ChainListener<C> {
             }
             Reason::NewRound { .. } => self.update_validators(&notification).await?,
             Reason::NewBlock { hash, .. } => {
-                self.update_validators(&notification).await?;
+                // self.update_validators(&notification).await?;
```
deuszx: We don't need to update validators about the new block here. The notification we're processing IS coming from the validators themselves.

linera-core/src/node.rs
```diff
@@ -279,6 +279,21 @@ pub enum NodeError {
...
+    pub fn is_missing_data(&self) -> bool {
+        // TODO: Should we include other errors here?
+        // It is unclear why `ChainError { error: "Round number should be Fast" }` is returned in some cases.
+        matches!(
+            self,
+            NodeError::ChainError { .. }
+                | NodeError::MissingCrossChainUpdate { .. }
+                | NodeError::MissingCertificateValue
+                | NodeError::BlobsNotFound(_)
+                | NodeError::EventsNotFound(_)
```
deuszx: to fix: These are actually not needed as `ValidatorUpdater::send_block_proposal` is recovering from those errors already.

deuszx: we should figure out how to return `ChainError { error: "Round number should be Fast" }` error more directly such that it's clear the validator needs to receive updates about the chain

linera-base/src/prometheus_util.rs
```diff
@@ -23,6 +23,12 @@ pub fn register_int_counter_vec(
     register_int_counter_vec!(counter_opts, label_names).expect("IntCounter can be created")
 }
 
+/// Wrapper around Prometheus register_int_counter_vec! macro which also sets the `linera` namespace
```
afck:
```suggestion
/// Wrapper around Prometheus `register_int_counter_vec!` macro which also sets the `linera` namespace
```

linera-client/src/chain_listener.rs
```diff
@@ -218,7 +218,6 @@ impl<C: ClientContext> ChainListener<C> {
             }
             Reason::NewRound { .. } => self.update_validators(&notification).await?,
             Reason::NewBlock { hash, .. } => {
-                self.update_validators(&notification).await?;
```
afck, line 221: We are processing notifications from the local node here. ~This could be about a certificate we have just put together locally, from the validator signatures (i.e. _no_ validator has it yet); and even~ if we received it from a validator, we only know that at least _one_ of them has it.

~I can't easily come up with a test scenario where this would break, off the top of my head, but I also don't see a reason why this call would always be unnecessary.~

linera-core/src/client/mod.rs
```diff
@@ -2337,6 +2337,7 @@ impl<Env: Environment> ChainClient<Env> {
 
         let mutex = self.state().client_mutex();
         let _guard = mutex.lock_owned().await;
+        // TOOD: We shouldn't need to call this explicitly.
```
afck, line 2336: ~This was to guard against the case where this is called even if there already is a pending block in progress. Not sure if it's still necessary.~

Edit: It's safe, since `set_pending_proposal` will refuse to overwrite a pending proposal. We might hit that error instead of finishing the existing proposal, though.

linera-core/src/client/mod.rs
afck: Good catch! :+1:

linera-core/src/updater.rs
```diff
@@ -482,7 +506,7 @@ where
         &mut self,
         action: CommunicateAction,
     ) -> Result<LiteVote, ChainClientError> {
-        let (target_block_height, chain_id) = match &action {
+        let (_target_block_height, chain_id) = match &action {
```
afck: Let's remove that from all the cases then; this line should just be: `let chain_id = action.chain_id()`.

afck (commented):
Most of this looks great! :+1: 
Not sure if all of the changes are safe, though.

afck (approved):
On second thought, the added error handling when making a proposal should make all those changes safe.

We could discuss with the team whether it's okay behavior that a client that receives a notification from one validator doesn't always update the others.

deuszx:
> On second thought, the added error handling when making a proposal should make all those changes safe.

I plan on going through the `ChainError`s once more (a separate PR though) and see whether it can be split into more fine-grained types of errors, like:
- critical
- recoverable (need retry)
- recipient needs updating
- etc.

This is already happening to some extent in `From<ChainError> for NodeError` but we might have missed something.

linera-client/src/chain_listener.rs
deuszx: We sent `ConfirmedBlock` to all validators already though (or will send) so they know about it. I'm not sure if we need to notify them _again_ every time one of validators accepts the new block.

afck: Did we? Even if we only processed it because _one_ validator notified us about the new block?

deuszx: This flow is initiated in `WorkerState::handle_confirmed_certificate`so only after quorum of validators have confirmed the block proposal.

afck: They may each have _signed_ the certificate, but they may not have the actual certificate with a quorum of signatures yet.

E.g. if a faulty client sent the final certificate (with a quorum of signatures) only to _one_ validator, the others don't have it, even though they signed it. Our own client may now receive the notification from that one validator, download the certificate from it and handle it locally. That would trigger this notification.

deuszx: The only way we could have constructed the `ConfirmedBlockCertificate` is if the quorum of validators responded to our `ConfirmedBlock`. It could have happened if we
- sent the fast block, or
- went through the `FinalizeBlock` logic

In both cases we communicate with all validators to send them the confirmed certificate. If they are online and responsive they will get it. If they're not, I don't see how calling it again here would be making a difference (we had tried a moment ago to do it already).

afck: But a faulty co-owner might not have gone through the same logic, or their connection failed etc. So it can happen that not all validators have the certificate.

afck: (Anyway, this is not a blocker from my side!)

Issue #4261, opened 2025-07-22
Make notifications more fine-grained, so that client know whether a relevant event was created.


afck:

afck:
If we make that distinction, we could also introduce different ways to subscribe to a chain, to distinguish between:
* chain followers, who only need `NewBlock` notifications, and
* chain owners, who also need `NewRound` and `NewIncomingBundle`.

PR #4284, opened 2025-07-29
Add information about published events to the NewBlock notification

bart-linera:
## Motivation

We want to include the information about published events in the notifications so that clients can choose whether they are interested in the particular blocks or not.

## Proposal

Add an `event_streams` field to the `NewBlock` notification, containing the stream IDs of streams to which the block published any events.

## Test Plan

CI

## Release Plan

## Links

- #4261 (this is just the validator-side part, the client-side part of actually using the event information in the notifications will be done in a follow-up PR)


Twey:
I was assuming the event notifications would have the event values directly in them.  Does getting the value of an event not require synchronizing the chain, otherwise?

bart-linera:
@Twey Currently it does, but with sparse chains for events it will only require getting the block with the event, and potentially other missing blocks with events in the same stream.

PR #4289, opened 2025-08-04
Remove message ID.

### LINKED ISSUE #3477: Serialize `MessageId` as a string if `is_human_readable`. Revisit other `bcs_scalar!` types.
If one uses `MessageId` as a field in a struct returned from a GraphQL query, that field will be a string with the hexadecimal representation of the message ID's BCS serialization. Using `serde_json` to deserialize the struct will fail, because it will expect `MessageId` as a struct.

We should probably revisit all types where we are using `bcs_scalar!`, and change them so that the serde serialization implementation also uses a string if [`is_human_readable()`](https://docs.rs/serde/latest/serde/trait.Serializer.html#method.is_human_readable).

afck:
## Motivation

The `MessageId` type isn't used anymore.

## Proposal

Remove it.

(This was almost entirely written by Claude Code, albeit with lots of back and forth. It also came up with the replacement `message_origin_chain_id` by itself.)

## Test Plan

CI

## Release Plan

## Links

- Closes #3477.


deuszx (commented):
Looks good although I am not able to verify Solidity-related changes. Were they generated or written manually?

afck:
> Were they generated or written manually?

As I said, this was all written by Claude.
So we should definitely wait for someone to carefully review the Solidity changes before merging.

afck:
(`LineraTypes.sol` is generated.)

linera-service/tests/fixtures/evm_example_execute_message.sol
```diff
@@ -20,9 +20,7 @@ contract ExampleExecuteMessage {
     function execute_message(bytes memory input) external {
         uint64 increment = abi.decode(input, (uint64));
         value = value + increment;
-        Linera.opt_MessageId memory opt_message_id = Linera.message_id();
-        require(opt_message_id.has_value);
-        last_message_id = opt_message_id.value;
+        last_message_origin_chain_id = Linera.message_origin_chain_id();
```
MathieuDutSik, line 23: We could maybe check that the origin of the message is the one we expect.
I can help write this code this weekend if that is a problem.

examples/rfq/src/contract.rs
```diff
@@ -314,9 +314,8 @@ impl Contract for RfqContract {
 impl RfqContract {
     fn get_message_creation_chain_id(&mut self) -> ChainId {
         self.runtime
-            .message_id()
-            .expect("Getting message id should not fail")
-            .chain_id
+            .message_origin_chain_id()
+            .expect("Getting message origin chain ID should not fail")
     }
```
MathieuDutSik: This code section should be at the beginning of the `fn execute_message`.

examples/amm/src/contract.rs
```diff
@@ -446,9 +446,8 @@ impl AmmContract {
 
     fn get_message_creation_chain_id(&mut self) -> ChainId {
         self.runtime
-            .message_id()
-            .expect("Getting message id should not fail")
-            .chain_id
+            .message_origin_chain_id()
+            .expect("Getting message origin chain ID should not fail")
     }
```
MathieuDutSik: That code section should be at the beginning of the `execute_message`.

linera-execution/solidity/Linera.sol
```diff
@@ -196,6 +196,11 @@ library Linera {
         AccountOwner value;
     }
 
+    struct opt_ChainId {
```
ma2bd, line 199: @MathieuDutSik Where these names `opt_ChainId` come from? Can we use correctly formed names like `ChainIdOption` or `OptionalChainId` instead?

linera-execution/solidity/Linera.sol
```diff
@@ -204,6 +209,22 @@ library Linera {
...
+
+    function opt_chainid_none()
+        internal
+        pure
+        returns (opt_ChainId memory)
+    {
+        return opt_ChainId(false, ChainId(bytes32(0)));
+    }
+
     enum OptionBool { None, True, False }
```
ma2bd, line 228: `OptionalBool` ?

ma2bd (approved):

linera-execution/solidity/Linera.sol
MathieuDutSik: The term `opt_ChainId` comes from the code generation in `serde-generate`.
See line 687 of the file `serde-generate/src/solidity.rs`.

linera-execution/solidity/Linera.sol
MathieuDutSik: This name is coming from line 694 of `serde-generate/src/solidity.rs`

bart-linera (approved):

linera-service/tests/fixtures/evm_example_execute_message.sol
afck: How do we know which one to expect here?

examples/rfq/src/contract.rs
afck: Done in https://github.com/linera-io/linera-protocol/pull/4289/commits/dafcaf52e91e62111d2b95a26a5fcc5a7475e19c.

examples/amm/src/contract.rs
afck: Done in https://github.com/linera-io/linera-protocol/pull/4289/commits/dafcaf52e91e62111d2b95a26a5fcc5a7475e19c.

PR #4293, opened 2025-08-05
Fix sanity check for incoming bundle notifications.

afck:
## Motivation

The sanity check was still assuming we would be _executing_ the sender chain.

## Proposal

Check the inboxes instead.

## Test Plan

The client test with sparse sender chains was modified to use the notification logic instead: With the fix, this doesn't print the spurious error anymore.

## Release Plan

## Links

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -3492,34 +3492,40 @@ impl<Env: Environment> ChainClient<Env> {
         maybe_info.map_or(BlockHeight::ZERO, |info| info.next_block_height)
     }
 
+    /// Returns the next height we expect to receive from the given sender chain, according to the
+    /// local inbox.
+    #[instrument(level = "trace")]
+    async fn local_next_height_to_receive(
+        &self,
+        origin: ChainId,
+    ) -> Result<BlockHeight, ChainClientError> {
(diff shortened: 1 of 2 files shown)


linera-core/src/client/mod.rs
```diff
@@ -3492,34 +3492,40 @@ impl<Env: Environment> ChainClient<Env> {
...
-                        chain_id = %self.chain_id,
-                        "NewIncomingBundle: Fail to process notification: {error}"
-                    );
-                    return;
-                }
-                if self.local_next_block_height(origin, &mut local_node).await <= height {
-                    error!(
+                self.find_received_certificates_from_validator(remote_node)
+                    .await?;
+                if self.local_next_height_to_receive(origin).await? <= height {
```
deuszx, line 3527: I think the `if` here is unnecessary. We just checked above if it's `> height` so the alternative it's `<=`.

afck: But the point of `find_received_certificates_from_validator` is to put things in our inbox and thereby _increase_ `local_next_height_to_receive`. Up to at least the height that the validator notified us about.

deuszx: Ah, I didn't think that it might change it.

Issue #4154, opened 2025-06-20
Provide a map from account → chains in the faucet


Twey:
# Problem

We have partners who would like Linera to behave more like Ethereum.  One big difference is that Linera has multiple chains where Ethereum has only one.  This means that in Linera to run a client one needs not only the private keys but also an up-to-date list of all the user's chains.

# Suggested solution

As a stop-gap measure, since the faucet is already involved in wallet creation, have it also provide a list of all chains claimed (on that faucet) by a given account.  This allows users to recreate the full wallet config from just the private key by talking to the faucet.

afck:
Do we still need this, or is #4301 sufficient for now?

Twey:
Fixed in #4301.

PR #4301, opened 2025-08-07
Make faucet claims idempotent.

afck:
## Motivation

In the future we plan to add a map of default user chains to the system itself (https://github.com/linera-io/linera-protocol/issues/4154), but this will involve deeper changes.

## Proposal

For now, make the faucet remember which chains it has created so far. If the same user (i.e. same public key) requests a chain twice, it now returns the same chain ID again.

Written mostly by Claude.

## Test Plan

`test_faucet_persistence` was added.

## Release Plan

## Links

- Quick temporary solution for https://github.com/linera-io/linera-protocol/issues/4154.

### DIFF
--- a/linera-faucet/server/src/tests.rs
+++ b/linera-faucet/server/src/tests.rs
@@ -3,7 +3,7 @@
 
 #![allow(clippy::large_futures)]
 
-use std::sync::Arc;
+use std::{fs, path::PathBuf, sync::Arc};
 
 use futures::lock::Mutex;
 use linera_base::{
@@ -18,7 +18,7 @@ use linera_core::{
     test_utils::{FaultType, MemoryStorageBuilder, StorageBuilder as _, TestBuilder},
 };
 
-use super::MutationRoot;
+use super::{FaucetStorage, MutationRoot};
 
 struct ClientContext {
     client: ChainClient<environment::Test>,
@@ -87,6 +87,8 @@ async fn test_faucet_rate_limiting() {
         update_calls: 0,
     };
     let context = Arc::new(Mutex::new(context));
+    let faucet_storage = Arc::new(Mutex::new(FaucetStorage::default()));
+    let storage_path = PathBuf::from("/tmp/test_faucet_storage.json");
     let root = MutationRoot {
         chain_id,
         context: context.clone(),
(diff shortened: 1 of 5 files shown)


linera-faucet/server/src/tests.rs
```diff
@@ -134,6 +138,61 @@ async fn test_faucet_rate_limiting() {
...
+        .unwrap();
+    let chain_id = client.chain_id();
+    let context = ClientContext {
+        client,
+        update_calls: 0,
+    };
+    let context = Arc::new(Mutex::new(context));
+
+    let faucet_storage = Arc::new(Mutex::new(FaucetStorage::default()));
+    let storage_path = PathBuf::from("/tmp/test_faucet_persistence.json");
```
deuszx: We have `tempfile` crate that manages these temeporary files/dirs for testing.

Twey (approved):

linera-faucet/server/src/tests.rs
afck: Done in https://github.com/linera-io/linera-protocol/pull/4301/commits/06710e573f88e6ef240aa46adcfd62f1c4e59b85.

afck:
One of the `remote-net` tests now loops indefinitely:
`
2025-08-07T14:36:33.802954Z DEBUG linera::main: linera_rpc::grpc::client: sending gRPC request handler="handle_block_proposal" request=BlockProposal { content: ProposalContent { block: ProposedBlock { chain_id: 21400120f052bbc1, epoch: Epoch(2), operations: [System(OpenChain(OpenChainConfig { ownership: ChainOwnership { owners: {Address32(e4ed9dd5d27c9083): 100}, multi_leader_rounds: 2, open_multi_leader_rounds: false, timeout_config: TimeoutConfig { base_timeout: TimeDelta(10000000), timeout_increment: TimeDelta(1000000), fallback_duration: TimeDelta(18446744073709551615) } }, balance: Amount(10000000000000000000), application_permissions: ApplicationPermissions }))], height: BlockHeight(0), timestamp: Timestamp(1754577372776486), authenticated_signer: Some(Address32(de166237331a2966)), previous_block_hash: None }, round: MultiLeader(0) }, signature: Ed25519 { signature: 44fb574808712c84, public_key: fcf518d56455283a } }
`
Need to investigate…

afck:
Right, this causes the `remote-net` tests to reuse chains. (Looks like different tests use the same client key?)
Probably `request-chain` should be modified, so that it also synchronizes the chain.

afck:
Or, even better: Finally properly handle the error where the validators have blocks we don't have, even if it's a chain with no other owner.

deuszx (approved):

PR #4303, opened 2025-08-07
Fix the LRU caching bug

### LINKED ISSUE #4299: LRU cache out of sync with storage.
The modified benchmark in #4269 is failing with `message not received yet` errors. To reproduce:
* Check out https://github.com/linera-io/linera-protocol/commit/bbfe1354791e1ffe196b6c01602e91bb5b564df1.
* Start a local network with `cargo build -p linera-storage-service -p linera-service && cargo run --bin linera -- net up --validators 1 --shards 1 --with-faucet`
* Run multi-benchmark with: `RUST_BACKTRACE=1 cargo run --bin linera --features benchmark -- multi-benchmark --processes 2 --faucet http://localhost:8080/ --bps 2 --transactions-per-block 100 --num-chains 2 --close-chains --cross-wallet-transfers`

My [debug-not-received](https://github.com/afck/linera-protocol/tree/debug-not-received) branch contains a few added fixes, `main` merges, logs and, crucially, assertions that the LRU cache agrees with storage. On this branch we actually see those assertions fail.

bart-linera:
## Motivation

We could be dropping incoming messages in some cases due to how different kinds of caching in our system interact with one another.

## Proposal

Make sure that restoring the execution state from a cache does not restore the storage cache as well. This is done by simply making the `execution_state` not be the first field in `ChainStateView`.

## Test Plan

The scenario reproducing the issue has been run manually, which confirmed that the change fixes the issue.

## Release Plan

## Links

- closes #4299

### DIFF
--- a/linera-chain/src/chain.rs
+++ b/linera-chain/src/chain.rs
@@ -242,12 +242,12 @@ where
     /// Hash of the execution state.
     pub execution_state_hash: RegisterView<C, Option<CryptoHash>>,
     /// Execution state, including system and user applications.
-    /// NOTE: This has to **not** be the first field in `ChainStateView` in order for execution
-    /// state caching to work. This is because the derivation of the `RootView` trait delegates
-    /// `context()` calls to the first field. If we cache the field, we also save the storage LRU
-    /// cache (if enabled), which is a part of the context, and restoring its value from the cache
-    /// results in storage using an outdated cache, which may result in incorrect values being
-    /// returned by read operations.
+    // NOTE: This has to **not** be the first field in `ChainStateView` in order for execution
+    // state caching to work. This is because the derivation of the `RootView` trait delegates
+    // `context()` calls to the first field. If we cache the field, we also save the storage LRU
+    // cache (if enabled), which is a part of the context, and restoring its value from the cache
+    // results in storage using an outdated cache, which may result in incorrect values being
+    // returned by read operations.
     pub execution_state: ExecutionStateView<C>,
 
     /// Block-chaining state.


ndr-ds (approved):

linera-chain/src/chain.rs
```diff
@@ -239,10 +239,16 @@ pub struct ChainStateView<C>
...
-    pub execution_state: ExecutionStateView<C>,
     /// Hash of the execution state.
     pub execution_state_hash: RegisterView<C, Option<CryptoHash>>,
+    /// Execution state, including system and user applications.
+    /// NOTE: This has to **not** be the first field in `ChainStateView` in order for execution
+    /// state caching to work. This is because the derivation of the `RootView` trait delegates
+    /// `context()` calls to the first field. If we cache the field, we also save the storage LRU
+    /// cache (if enabled), which is a part of the context, and restoring its value from the cache
+    /// results in storage using an outdated cache, which may result in incorrect values being
+    /// returned by read operations.
```
afck: I think it would be better if this didn't show up in GraphQL. Also, it doesn't affect any user of this type; it's only relevant if one wants to modify the type.

It would also be good to have a ticket for a proper solution, and link to it with a `TODO` here.

```suggestion
    // NOTE: This has to **not** be the first field in `ChainStateView` in order for execution
    // state caching to work. This is because the derivation of the `RootView` trait delegates
    // `context()` calls to the first field. If we cache the field, we also save the storage LRU
    // cache (if enabled), which is a part of the context, and restoring its value from the cache
    // results in storage using an outdated cache, which may result in incorrect values being
    // returned by read operations.
```

afck (approved):

ma2bd:
This feels like a hack. Let's go for #4304 directly instead?

bart-linera:
Closing in favor of #4304 , then :+1:

PR #4304, opened 2025-08-08
Fix the LRU caching bug - alternative solution

### LINKED ISSUE #4299: LRU cache out of sync with storage.

bart-linera:
## Motivation

We could be dropping incoming messages in some cases due to how different kinds of caching in our system interact with one another.

## Proposal

Remove the context from cached ExecutionStateViews and restore it upon restoring from cache. This way the storage caches aren't saved, and the correct cache is used by all the views even when restored from execution state cache.

## Test Plan

Tested manually that it fixes the issue + CI for regressions

## Release Plan

## Links

- Alternative to #4303 
- Closes #4299

### DIFF
--- a/linera-execution/src/system.rs
+++ b/linera-execution/src/system.rs
@@ -8,6 +8,7 @@ mod tests;
 
 use std::{
     collections::{BTreeMap, BTreeSet, HashSet},
+    fmt::{Display, Formatter},
     mem,
 };
 
@@ -293,6 +294,15 @@ impl Recipient {
     }
 }
 
+impl Display for Recipient {
+    fn fmt(&self, f: &mut Formatter) -> std::fmt::Result {
+        match self {
+            Recipient::Burn => write!(f, "burn"),
+            Recipient::Account(account) => write!(f, "{}", account),
+        }
+    }
+}
+
 impl From<ChainId> for Recipient {
     fn from(chain_id: ChainId) -> Self {
         Recipient::chain(chain_id)
@@ -422,7 +432,7 @@ where
                 let maybe_message = self
                     .transfer(context.authenticated_signer, None, owner, recipient, amount)
                     .await?;
-                txn_tracker.add_outgoing_messages(maybe_message)?;
+                txn_tracker.add_outgoing_messages(maybe_message);
             }
             Claim {
                 owner,
@@ -440,7 +450,7 @@ where
                         amount,
                     )
                     .await?;
-                txn_tracker.add_outgoing_message(message)?;
+                txn_tracker.add_outgoing_message(message);
             }
             Admin(admin_operation) => {
                 ensure!(
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -181,7 +181,7 @@ jobs:
     needs: changed-files
     if: needs.changed-files.outputs.should-run == 'true'
     runs-on: ubuntu-latest
-    timeout-minutes: 30
+    timeout-minutes: 40
 
     steps:
     - uses: actions/checkout@v4
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5412,6 +5412,7 @@ dependencies = [
  "linera-base",
  "linera-chain",
  "linera-core",
+ "linera-execution",
  "linera-service-graphql-client",
  "linera-version",
  "linera-views",
--- a/linera-chain/src/certificate/lite.rs
+++ b/linera-chain/src/certificate/lite.rs
@@ -45,22 +45,26 @@ impl LiteCertificate<'_> {
         }
     }
 
-    /// Creates a [`LiteCertificate`] from a list of votes, without cryptographically checking the
+    /// Creates a [`LiteCertificate`] from a list of votes with their validator public keys, without cryptographically checking the
     /// signatures. Returns `None` if the votes are empty or don't have matching values and rounds.
-    pub fn try_from_votes(votes: impl IntoIterator<Item = LiteVote>) -> Option<Self> {
+    pub fn try_from_votes(
+        votes: impl IntoIterator<Item = (ValidatorPublicKey, LiteVote)>,
+    ) -> Option<Self> {
         let mut votes = votes.into_iter();
-        let LiteVote {
-            value,
-            round,
+        let (
             public_key,
-            signature,
-        } = votes.next()?;
+            LiteVote {
+                value,
+                round,
+                signature,
+            },
+        ) = votes.next()?;
         let mut signatures = vec![(public_key, signature)];
-        for vote in votes {
+        for (validator_key, vote) in votes {
             if vote.value.value_hash != value.value_hash || vote.round != round {
                 return None;
             }
-            signatures.push((vote.public_key, vote.signature));
+            signatures.push((validator_key, vote.signature));
         }
         Some(LiteCertificate::new(value, round, signatures))
     }
--- a/linera-chain/src/data_types.rs
+++ b/linera-chain/src/data_types.rs
@@ -338,7 +338,6 @@ struct VoteValue(CryptoHash, Round, CertificateKind);
 pub struct Vote<T> {
     pub value: T,
     pub round: Round,
-    pub public_key: ValidatorPublicKey,
     pub signature: ValidatorSignature,
 }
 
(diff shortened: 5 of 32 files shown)


linera-views/src/store.rs
```diff
@@ -255,3 +255,91 @@ pub trait TestKeyValueDatabase: KeyValueDatabase {
...
+
+    impl KeyValueStoreError for DummyStoreError {
+        const BACKEND: &'static str = "dummy";
+    }
+
+    impl WithError for DummyStore {
+        type Error = DummyStoreError;
+    }
+
+    impl ReadableKeyValueStore for DummyStore {
```
deuszx: Would it be correct to `panic!` here? to detect calls to something that shouldn't be called? Or maybe even use the `!` (unreachable) type?

bart-linera: Probably, as we don't expect these functions to actually be called. I'll try it out :+1:

linera-execution/src/system.rs
```diff
@@ -93,6 +93,27 @@ pub struct SystemExecutionStateView<C> {
     pub event_subscriptions: MapView<C, (ChainId, StreamId), EventSubscriptions>,
 }
 
+impl<C: Context, C2: Context> ReplaceContext<C2> for SystemExecutionStateView<C> {
```
afck, line 97: We should probably provide a `derive` macro for this at some point. (Not in this PR, of course.)

linera-core/src/chain_worker/state/temporary_changes.rs
```diff
@@ -334,7 +337,12 @@ where
             .await?;
         self.0.execution_state_cache.insert_owned(
             &outcome.state_hash,
-            self.0.chain.execution_state.clone_unchecked()?,
+            self.0
+                .chain
+                .execution_state
+                .clone_unchecked()?
```
afck: Is this necessary? Doesn't `with_context` clone everything anyway?

(Maybe it shouldn't, and take `self` instead? Or should it be called `clone_with_context`?)

bart-linera: Ah, right - I initially removed it, but restored it after getting some test failures. It wasn't the reason, though, so it can be removed :+1:

(As for taking `self`, I tried that, but there is a view for which it is impossible to implement with taking ownership because of some `Arc`s in the structure.)

linera-execution/src/system.rs
bart-linera: It's problematic, because it's not straightforward how the `Result` type should be automatically defined. But if possible, then it should definitely be desirable :+1:

linera-views/src/views/mod.rs
```diff
@@ -96,6 +96,15 @@ pub trait View: Sized {
     }
 }
 
+/// A view which can have its context replaced.
+pub trait ReplaceContext<C: crate::context::Context>: View {
+    /// The type returned after replacing the context.
+    type Result: View<Context = C>;
```
ma2bd: nit: `Result` is a bit overloaded. I believe `Target` is idiomatic

linera-execution/src/system.rs
ma2bd: IIRC the current macro is already introspecting the container declaration to detect the context parameter.

linera-views/src/views/mod.rs
```diff
@@ -96,6 +96,15 @@ pub trait View: Sized {
...
+
+    /// Returns a view with a replaced context.
+    async fn with_context(&self, ctx: impl FnOnce(&Self::Context) -> C + Clone) -> Self::Result;
```
ma2bd: What's the reasoning behind using a clonable `FnOnce` function? Why not `Fn`?

ma2bd: Also we may want to consume `self` to simplify the code (even if this means calling `clone_unchecked` first).

`clone_unchecked` takes `&mut self` by the way. Maybe something to double-check because there could be a reason (EDIT: actually there is! I just remembered)

bart-linera: The reasoning was that we don't call this more than once, but we sometimes need to pass it to multiple calls of `with_context` on child views, so we need to clone it.

As for consuming self, as far as I can tell, it's impossible to implement it this way on `ReentrantByteCollectionView` due to the `Arc<RwLock>`s inside it - even if we own the `Arc`s, it doesn't guarantee exclusive ownership of whatever is inside, and there might be a non-cloneable view inside on which we need to call `with_context` as well.

linera-execution/src/system.rs
bart-linera: Ah, I see, it just seems to assume that context is the first generic parameter. If we make that assumption anyway, then maybe it wouldn't be that hard to implement the macro after all :thinking:

deuszx: We should watch out for these assumptions in macro.

linera-views/src/views/mod.rs
ma2bd: Then we probably want to take `&mut self` because saving the cloned view later can modify the storage. (And &-references on views are not allowed to mutate the DB)

ma2bd: (And &-references on views are not allowed to mutate the DB)

linera-execution/src/system.rs
ma2bd: We do!

linera-views/src/store.rs
```diff
@@ -255,3 +255,91 @@ pub trait TestKeyValueDatabase: KeyValueDatabase {
         database.open_shared(&[])
     }
 }
+
+/// A module containing a dummy store used for caching views.
+pub mod inactive_store {
+    use super::*;
+
+    /// A store which is not actually stored anything - used for caching views.
```
afck:
```suggestion
    /// A store which does not actually store anything - used for caching views.
```

afck (approved):

afck (approved):

linera-views/src/views/mod.rs
```diff
@@ -96,6 +96,16 @@ pub trait View: Sized {
     }
 }
 
+/// A view which can have its context replaced.
+pub trait ReplaceContext<C: crate::context::Context>: View {
```
Twey, line 100: I'm not sure I understand the reason for this parameter.  It's anyway fixed by `Target`.

linera-views/src/views/mod.rs
```diff
@@ -96,6 +96,16 @@ pub trait View: Sized {
...
+    /// The type returned after replacing the context.
+    type Target: View<Context = C>;
+
+    /// Returns a view with a replaced context.
+    async fn with_context(&mut self, ctx: impl FnOnce(&Self::Context) -> C + Clone)
+        -> Self::Target;
```
Twey, line 106: `impl FnOnce + Clone` should generally be `impl Fn`.

linera-views/src/views/mod.rs
bart-linera: Without it, we'd have to do separate impls for, say, `ExecutionStateView<ViewContext>` and `ExecutionStateView<InactiveContext>`. Also, it makes sense for a single type to have the ability to have its context replaced by many different context types, not just a single one.

linera-views/src/views/mod.rs
bart-linera: That wouldn't work with multiple calls of `self.subfield.with_context(ctx)` within an implementation of `SomeView::with_context`. It would result in a "value used after move". `Fn` doesn't imply `Clone`, and we need the `Clone`.

PR #4305, opened 2025-08-08
Display more details.

deuszx:
## Motivation

Currently only a few fields (from the header, body or messages) are stored in a denormalized way in a database – ie they are stored in dedicated fields – while the whole block is stored as a blob data. We want to show more details to user, this could be done in one of the two ways:
1. load the blob, deserialize in frontend
2. store in denormalized way and load directly in frontend

## Proposal

Implement the approach `#2` – add more fields to database tables and store block's details directly there. Also, use the new data to show on the frontend of the explorer.

Some refactoring was done as part of the PR:
- add a dedicated type `ExpandableSection` that shows only the metadata of the section and expands on click to show more
- add `CopyableHash` to present a text that can be copied.

## Test Plan

Manual

## Release Plan

## Links

### DIFF
--- a/linera-explorer-new/server/database.js
+++ b/linera-explorer-new/server/database.js
@@ -60,14 +60,103 @@ export class BlockchainDatabase {
     return stmt.all(blockHash);
   }
 
-  // Get posted messages for a bundle
+  // Get posted messages for a bundle with SystemMessage fields
   getPostedMessages(bundleId) {
     const stmt = this.db.prepare(`
-      SELECT * FROM posted_messages 
+      SELECT *, system_target, system_amount, system_source, system_owner, system_recipient,
+             message_type, application_id, system_message_type
+      FROM posted_messages 
       WHERE bundle_id = ? 
       ORDER BY message_index ASC
     `);
     return stmt.all(bundleId);
+  }
+
+  // Get block with all bundles and their messages in a single query
+  getBlockWithBundlesAndMessages(blockHash) {
+    const stmt = this.db.prepare(`
+      SELECT 
+        ib.id as bundle_id,
+        ib.bundle_index,
+        ib.origin_chain_id,
+        ib.action,
+        ib.source_height,
+        ib.source_timestamp,
+        ib.source_cert_hash,
+        ib.transaction_index as bundle_transaction_index,
+        ib.created_at as bundle_created_at,
+        pm.id as message_id,
+        pm.message_index,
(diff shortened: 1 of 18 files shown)


linera-explorer-new/server/database.js
```diff
@@ -60,16 +60,105 @@ export class BlockchainDatabase {
...
+    
+    const rows = stmt.all(blockHash);
+    
+    // Group results by bundle
+    const bundlesMap = new Map();
+    
+    for (const row of rows) {
+      const bundleId = row.bundle_id;
+      
+      if (!bundlesMap.has(bundleId)) {
```
deuszx, line 118: There should never be a duplicate entry with the same `bundleId`.

linera-explorer-new/server/index.js
```diff
@@ -75,6 +75,30 @@ app.get('/api/blocks/:hash/bundles', (req, res) => {
   }
 });
 
+// Get bundles with messages - optimized single query
+app.get('/api/blocks/:hash/bundles-with-messages', (req, res) => {
```
deuszx, line 79: This should be a default way for getting a block by hash.

linera-explorer-new/src/hooks/useDatabase.ts
```diff
@@ -139,13 +149,37 @@ export const useBlock = (hash: string) => {
...
+          eventsResult,
+          oracleResponsesResult
+        ] = await Promise.all([
           api.getBlockByHash(hash),
-          api.getIncomingBundles(hash)
+          api.getBundlesWithMessages(hash), // Use optimized query
+          api.getOperations(hash),
+          api.getMessages(hash),
+          api.getEvents(hash),
+          api.getOracleResponses(hash)
```
deuszx, line 165: All these data should be queried for with a single query with JOINs.

linera-explorer-new/src/types/blockchain.ts
```diff
@@ -44,13 +44,81 @@ export interface PostedMessage {
...
+  system_source?: string;
+  system_owner?: string;
+  system_recipient?: string;
   message_data: Uint8Array;
   created_at: string;
 }
 
+export interface IncomingBundleWithMessages extends IncomingBundle {
+  messages: PostedMessage[];
+}
```
deuszx, line 61: This should be called `IncomingBundle` since having incoming messages is the default.

PR #4269, opened 2025-07-23
Make chains transfer to each other across benchmark processes

ndr-ds:
## Motivation

Right now the benchmark client processes only do transfer amongst chains in their wallets. We should have a way of these processes be able to send transfers to chains in other wallets.

## Proposal

Pass a file to the benchmark client processes with all the chain IDs (from across wallets) that we can send transfers to. Enable that via `multi-benchmark` with an option

## Test Plan

Ran this with the new option locally against a local network, seems to do what it's supposed to.

## Release Plan

- Nothing to do / These changes follow the usual release cycle.

PR #4312, opened 2025-08-08
Run benchmark e2e tests for a bit longer

ndr-ds:
## Motivation

Right now the tests start and run just for one second, which since we're doing a bunch of extra stuff now, it's not enough time. From the logs it seems like we're trying to kill things before some messages sent with `wait_for_outgoing_messages` (not coming from the benchmark specific code) actually get delivered. So it seems like we just wait forever in that case.

## Proposal

Run the tests for slightly longer (5 seconds instead of 1), and that seems to prevent us from killing things too early. Might be a good follow up to see if blocking cross chain messages should timeout?

## Test Plan

CI should pass now

## Release Plan

### DIFF
--- a/linera-service/tests/local_net_tests.rs
+++ b/linera-service/tests/local_net_tests.rs
@@ -801,8 +801,8 @@ async fn test_end_to_end_benchmark(mut config: LocalNetConfig) -> Result<()> {
         .benchmark(BenchmarkCommand {
             num_chains: 2,
             transactions_per_block: 10,
-            bps: 1,
-            runtime_in_seconds: Some(1),
+            bps: 2,
+            runtime_in_seconds: Some(5),
             close_chains: true,
             ..Default::default()
         })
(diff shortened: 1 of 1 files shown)


ndr-ds:
* **#4269** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4269?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4312** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4312?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4312?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ndr-ds:
Ideally I 100% agree, but since this test just runs a long running thing, we have to just run it with a time limit, unless we want to design a whole thing just for tests (which is what happened before). So given the circumstances, this seems reasonable 🤔

PR #4316, opened 2025-08-11
Add the `application-creator-chain-id` to the ServiceRuntime.

### LINKED ISSUE #3606: Is it possible to also export `application_creator_chain_id` to service runtime in `linera-sdk` ?
**Motivation**
Sometimes we may need to query application creator chain in frontend. For example, we're trying to build a `blob gateway` application which will let users register their blobs to the gateway application when publishing. But it mustn't access the blob index on every microchain. Just make it could be accessed all over the world from its creator chain will be enough. I believe it'll work if we hard code creator chain id in frontend. But if it's possible to get that from application query then it'll be much robust method for application developer.

**Suggestion**
Export `application_creator_chain_id` to service runtime then we can query it directly.

MathieuDutSik:
## Motivation

The `application-creator-chain-id` was not in `ServiceRuntime` which was an oversight.

Fixes #3606 

## Proposal

The implementation is straightforward.

## Test Plan

The CI.
The test has been added to the `counter-no-graphql`.

## Release Plan

## Links

None

### DIFF
--- a/examples/counter-no-graphql/src/service.rs
+++ b/examples/counter-no-graphql/src/service.rs
@@ -63,7 +63,12 @@ mod tests {
     use std::sync::Arc;
 
     use futures::FutureExt as _;
-    use linera_sdk::{util::BlockingWait, views::View, Service, ServiceRuntime};
+    use linera_sdk::{
+        linera_base_types::{ChainId, CryptoHash},
+        util::BlockingWait,
+        views::View,
+        Service, ServiceRuntime,
+    };
 
     use super::{CounterRequest, CounterService, CounterState};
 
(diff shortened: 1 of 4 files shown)


afck (approved):
I wonder whether we should expose the whole `ApplicationDescription`. (In both contract and service.) But that's a separate discussion and PR, of course.

MathieuDutSik:
> I wonder whether we should expose the whole `ApplicationDescription`. (In both contract and service.) But that's a separate discussion and PR, of course.

I think no:
* The `ApplicationDescription` is a relatively complex data structure. If we expose it, then we bind ourselves as it becomes a public API.
* Some of the entries like `application_index` are of no importance to the user.

PR #4318, opened 2025-08-11
Remove public key from `Vote`.

afck:
## Motivation

The public key is known from the context everywhere, and having it explicitly in `Vote` and `LiteVote` just duplicates data and creates more error cases.

## Proposal

Remove the redundant field (with Claude).

## Test Plan

CI

## Release Plan

## Links

### DIFF
--- a/linera-chain/src/certificate/lite.rs
+++ b/linera-chain/src/certificate/lite.rs
@@ -45,22 +45,26 @@ impl LiteCertificate<'_> {
         }
     }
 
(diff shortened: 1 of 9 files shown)


deuszx (approved):
My understanding was that the public key is there to be able to verify the vote in _isolation_ – without the context.

Previously we were signing with Ed25519 which doesn't have public key recovery feature (given signature and hash signed) but now – with secp256k1 – we can do that so this can still be achieved.

afck:
Right, that's true. But even so: In the client code we always know the key from context anyway.

PR #4327, opened 2025-08-12
Improve "missing vote" errors.

afck:
## Motivation

We are seeing "missing vote" errors in some scenarios, although it's not clear it can currently be reproduced on `main`.

## Proposal

Include in the error message whether it was a missing vote due to a proposed block, validated block or requested timeout.

Make validators return errors instead of just skipping in more cases. Now we only skip signing and return the chain info if we either aren't a validator (don't have signing keys) or we already have signed the requested vote.

## Test Plan

CI should catch regressions.
Otherwise these changes should both make "missing vote" errors less likely (impossible, with non-faulty validators) and easier to debug.

## Release Plan

## Links

### DIFF
--- a/linera-chain/src/manager.rs
+++ b/linera-chain/src/manager.rs
@@ -360,7 +360,7 @@ where
 
     /// Checks if the current round has timed out, and signs a `Timeout`. Returns `true` if the
     /// chain manager's state has changed.
-    pub fn vote_timeout(
+    pub fn create_timeout_vote(
         &mut self,
         chain_id: ChainId,
         height: BlockHeight,
--- a/linera-chain/src/manager.rs
+++ b/linera-chain/src/manager.rs
@@ -358,34 +358,40 @@ where
         Ok(Outcome::Accept)
     }
 
-    /// Checks if the current round has timed out, and signs a `Timeout`.
+    /// Checks if the current round has timed out, and signs a `Timeout`. Returns `true` if the
+    /// chain manager's state has changed.
     pub fn vote_timeout(
         &mut self,
         chain_id: ChainId,
         height: BlockHeight,
+        round: Round,
         epoch: Epoch,
         key_pair: Option<&ValidatorSecretKey>,
         local_time: Timestamp,
-    ) -> bool {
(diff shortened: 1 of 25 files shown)


linera-chain/src/manager.rs
```diff
@@ -358,34 +358,40 @@ where
         Ok(Outcome::Accept)
     }
 
-    /// Checks if the current round has timed out, and signs a `Timeout`.
+    /// Checks if the current round has timed out, and signs a `Timeout`. Returns `true` if the
+    /// chain manager's state has changed.
```
MathieuDutSik, line 362: I find the name of this function to be misleading.
With the existing API, the name should be `has_manager_state_changed` with the error pattern being whether it has timeout or not. But the error pattern is usually something that is in the background and not in the name of the function.

afck: Good point, but `has_manager_state_changed` doesn't convey that it _actively_ changes the state.

Maybe `create_timeout_vote`?

afck: Done in https://github.com/linera-io/linera-protocol/pull/4327/commits/22de1e47995b6b0dd11b2a9d5c5b2680a807d1d0.

linera-chain/src/manager.rs
```diff
@@ -431,11 +437,6 @@ where
         }
 
         if let Some(locking) = self.locking_block.get() {
-            if let LockingBlock::Regular(locking_cert) = locking {
-                if locking_cert.hash() == certificate.hash() && locking.round() == new_round {
-                    return Ok(Outcome::Skip); // We already handled this certificate.
-                }
-            }
```
MathieuDutSik, line 438: Why is that code removed?

afck: The motivation for this PR is that we don't want to `Skip` and return a chain info without a vote. This `Skip` here was redundant and potentially wrong: If the reason for the locking block is that we already signed to confirm it, we would return in line 430 above already. If the reason is different, and we saw it only e.g. because it was included in a new proposal, then we shouldn't return `Skip`, and continue with the remaining checks instead.

PR #4329, opened 2025-08-12
Remove unused error variants; make some modules private.

afck:
## Motivation

Public items make it hard to detect unused code.

## Proposal

I let Claude find and remove unused error variants. I also made some modules in `linera-core` private.

## Test Plan

CI

## Release Plan

## Links

### DIFF
--- a/linera-chain/src/lib.rs
+++ b/linera-chain/src/lib.rs
@@ -26,7 +26,7 @@ pub use chain::ChainStateView;
 use data_types::{MessageBundle, PostedMessage};
 use linera_base::{
     bcs,
-    crypto::{CryptoError, CryptoHash},
+    crypto::CryptoError,
     data_types::{ArithmeticError, BlockHeight, Round, Timestamp},
     identifiers::{ApplicationId, ChainId},
 };
@@ -123,8 +123,7 @@ pub enum ChainError {
         "Block timestamp {new} must not be earlier than the parent block's timestamp {parent}"
     )]
     InvalidBlockTimestamp { parent: Timestamp, new: Timestamp },
-    #[error("Cannot initiate a new block while the previous one is still pending confirmation")]
-    PreviousBlockMustBeConfirmedFirst,
+
     #[error("Round number should be at least {0:?}")]
     InsufficientRound(Round),
     #[error("Round number should be greater than {0:?}")]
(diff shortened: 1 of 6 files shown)


linera-execution/src/lib.rs
```diff
@@ -320,16 +314,6 @@ pub enum ExecutionError {
     InvalidCommitteeEpoch { expected: Epoch, provided: Epoch },
     #[error("Failed to remove committee")]
     InvalidCommitteeRemoval,
-    #[error("Amount overflow")]
-    AmountOverflow,
```
MathieuDutSik, line 324: Is it because we are using `saturating_add`?
An overflow would have made sense,

linera-execution/src/lib.rs
```diff
@@ -221,8 +221,6 @@ pub enum ExecutionError {
         caller_id: Box<ApplicationId>,
         callee_id: Box<ApplicationId>,
     },
-    #[error("Attempt to write to storage from a contract")]
-    ServiceWriteAttempt,
```
MathieuDutSik, line 225: Good catch.

linera-execution/src/lib.rs
```diff
@@ -320,16 +314,6 @@ pub enum ExecutionError {
...
-    #[error("Amount overflow")]
-    AmountOverflow,
-    #[error("Amount underflow")]
-    AmountUnderflow,
-    #[error("Chain balance overflow")]
-    BalanceOverflow,
-    #[error("Chain balance underflow")]
-    BalanceUnderflow,
-    #[error("Application {0:?} is not registered by the chain")]
-    UnknownApplicationId(Box<ApplicationId>),
```
MathieuDutSik, line 332: Indeed, the error is obsolete.

linera-chain/src/lib.rs
```diff
@@ -161,15 +155,8 @@ pub enum ChainError {
...
     MissingMandatoryApplications(Vec<ApplicationId>),
-    #[error("Can't use grant across different broadcast messages")]
-    GrantUseOnBroadcast,
     #[error("Executed block contains fewer oracle responses than requests")]
     MissingOracleResponseList,
-    #[error("Unexpected hash for CertificateValue! Expected: {expected:?}, Actual: {actual:?}")]
-    CertificateValueHashMismatch {
-        expected: CryptoHash,
-        actual: CryptoHash,
-    },
```
MathieuDutSik, line 172: We have an error `ViewError::InconsistentEntries` for hash inconsistencies in the storage. Not sure what this one was about.

linera-execution/src/lib.rs
```diff
@@ -295,8 +291,6 @@ pub enum ExecutionError {
 
     #[error("No NetworkDescription found in storage")]
     NoNetworkDescriptionFound,
-    #[error("Invalid committees")]
-    InvalidCommittees,
```
MathieuDutSik, line 299: I think replaced by committees.

linera-execution/src/lib.rs
```diff
@@ -245,8 +243,6 @@ pub enum ExecutionError {
     HttpResponseSizeLimitExceeded { limit: u64, size: u64 },
     #[error("Runtime failed to respond to application")]
     MissingRuntimeResponse,
-    #[error("Module ID {0:?} is invalid")]
-    InvalidModuleId(ModuleId),
```
MathieuDutSik, line 249: Indeed, the errors would appear somewhere else in `ExecutionError`.

MathieuDutSik (approved):
I think all the cases that have been found are indeed because the error case is now obsolete.
I did not find a case where the error being removed is because of uncovered error scenario.

linera-execution/src/lib.rs
afck: I think we just wrap `ArithmeticError` instead.

PR #4331, opened 2025-08-12
Add single destination per block option to benchmark client

ndr-ds:
## Motivation

Right now the default is for every chain to always send transfers to every other chain, in a way that in each block we cycle through the chains.

## Proposal

Add an option so that we continue sending chains to every other chain, but every block contains transactions to just one of the other chains, and we cycle through chains per block instead of per transfer.

## Test Plan

Deploy a network and test this mode against it.

## Release Plan

### DIFF
--- a/linera-client/src/benchmark.rs
+++ b/linera-client/src/benchmark.rs
@@ -121,6 +121,7 @@ impl<Env: Environment> Benchmark<Env> {
         delay_between_chains_ms: Option<u64>,
         chain_listener: ChainListener<C>,
         shutdown_notifier: &CancellationToken,
+        single_destination_per_block: bool,
     ) -> Result<(), BenchmarkError> {
         let num_chains = chain_clients.len();
         let bps_counts = (0..num_chains)
@@ -183,6 +184,7 @@ impl<Env: Environment> Benchmark<Env> {
                         notifier_clone,
                         runtime_control_sender_clone,
                         delay_between_chains_ms,
+                        single_destination_per_block,
                     ))
                     .await?;
 
@@ -566,6 +568,7 @@ impl<Env: Environment> Benchmark<Env> {
         notifier: Arc<Notify>,
         runtime_control_sender: Option<mpsc::Sender<()>>,
         delay_between_chains_ms: Option<u64>,
+        single_destination_per_block: bool,
     ) -> Result<(), BenchmarkError> {
         barrier.wait().await;
         if let Some(delay_between_chains_ms) = delay_between_chains_ms {
@@ -600,6 +603,7 @@ impl<Env: Environment> Benchmark<Env> {
                         transactions_per_block,
                         fungible_application_id,
                         &mut destination_manager,
+                        single_destination_per_block,
                     ),
                     vec![]
                 ) => {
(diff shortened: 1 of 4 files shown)


ndr-ds:
* **#4331** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4331?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4331?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#4345** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4345?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4346** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4346?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

linera-client/src/benchmark.rs
```diff
@@ -40,6 +40,8 @@ use crate::chain_listener::{ChainListener, ClientContext};
 const PROXY_LATENCY_P99_THRESHOLD: f64 = 400.0;
 const LATENCY_METRIC_PREFIX: &str = "linera_proxy_request_latency";
 
+pub type BlockOperations = Vec<Operation>;
```
ma2bd: nit: can this be not public?

linera-client/src/benchmark.rs
```diff
@@ -643,18 +645,19 @@ impl<Env: Environment> Benchmark<Env> {
         Ok(all_chains)
     }
 
+    // This function returns a vector, one position for each of the sending chains.
+    // Each of those positions contains has a vector of block operations for that chain.
```
ma2bd: typo: contains has

linera-client/src/benchmark.rs
```diff
@@ -643,18 +645,19 @@ impl<Env: Environment> Benchmark<Env> {
...
     pub fn make_benchmark_block_info(
         benchmark_chains: Vec<(ChainId, AccountOwner)>,
         transactions_per_block: usize,
         fungible_application_id: Option<ApplicationId>,
```
ma2bd: Can we just take a number of payments per block and remove the complicated adjustments?

linera-client/src/benchmark.rs
```diff
@@ -685,30 +688,42 @@ impl<Env: Environment> Benchmark<Env> {
...
+                    owner: AccountOwner::CHAIN,
+                    recipient: Recipient::chain(recipient_chain_id),
+                    amount,
+                }),
+            };
+
+            let mut block_operations = Vec::new();
+            if single_destination_per_block {
+                // Generate one block per destination chain
+                for recipient_chain_id in other_chains {
```
ma2bd: Can we pick `recipient_chain_id` at random? (once)

linera-client/src/benchmark.rs
```diff
@@ -685,30 +688,42 @@ impl<Env: Environment> Benchmark<Env> {
...
+                    let operations: Vec<Operation> = (0..adjusted_transactions_per_block)
```
ma2bd: nit: type annotation is probably not needed (here and below)

linera-client/src/benchmark.rs
```diff
@@ -685,30 +688,42 @@ impl<Env: Environment> Benchmark<Env> {
...
+                        .map(|_| create_operation(recipient_chain_id))
+                        .collect();
+                    block_operations.push(operations);
+                }
+            } else {
+                // Generate one block with transactions cycling through destinations
```
ma2bd: We don't really want cycling. Who knows how it impacts the system. Can we instead pick `recipient_chain_id` at random? (N times where N is the desired number of payments)

linera-client/src/benchmark.rs
ma2bd: This will also simplify the code -- which currently has to create many blocks ahead of time to ensure distribution.

linera-client/src/benchmark.rs
```diff
@@ -643,18 +645,19 @@ impl<Env: Environment> Benchmark<Env> {
...
+    ) -> Result<Vec<Vec<BlockOperations>>, BenchmarkError> {
         let mut blocks_infos = Vec::new();
         let amount = Amount::from_attos(1);
 
         for (current_chain_id, owner) in benchmark_chains.iter() {
-            let mut operations = Vec::new();
-
             let mut other_chains: Vec<_> = if all_chains.len() == 1 {
                 // If there's only one chain, just have it send to itself.
                 all_chains.clone()
```
ndr-ds: <p class='graphite__hidden'><i>[Re: line +674]</i></p>
For every chain, I randomize the destinations here. Do you mean you want a sequence that is randomized everytime? Instead of cycling through an initially randomized sequence?
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4331?utm_source=unchanged-line-comment">Graphite</a>.</p>

ma2bd: Yes I would prefer not cycling through lists of recipients (although it won't probably change much).

afck (approved):

PR #4332, opened 2025-08-13
Make some variables private and other corrections.

MathieuDutSik:
## Motivation

Some variables are public without reason, and some inexact comments.

## Proposal

Straightforward corrections.

## Test Plan

The CI.

## Release Plan

## Links

None

### DIFF
--- a/linera-chain/src/block.rs
+++ b/linera-chain/src/block.rs
@@ -159,7 +159,7 @@ pub struct Timeout(Hashed<TimeoutInner>);
 
 #[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
 #[serde(rename = "Timeout")]
-pub struct TimeoutInner {
+pub(crate) struct TimeoutInner {
     chain_id: ChainId,
     height: BlockHeight,
     epoch: Epoch,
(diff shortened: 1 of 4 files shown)


linera-chain/src/block.rs
```diff
@@ -160,9 +160,9 @@ pub struct Timeout(Hashed<TimeoutInner>);
 #[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
 #[serde(rename = "Timeout")]
 pub struct TimeoutInner {
```
afck: Good catch! Probably the whole type can be private?

```suggestion
struct TimeoutInner {
```

linera-chain/src/block_tracker.rs
```diff
@@ -380,7 +380,7 @@ impl<'resources, 'blobs> BlockExecutionTracker<'resources, 'blobs> {
     /// Finalizes the execution and returns the collected results.
     ///
     /// This method should be called after all transactions have been processed.
-    /// Panics if the number of outcomes does match the expected count.
+    /// Panics if the counts does not match the expected counts.
```
afck: That explains less than the original. I guess "outcomes" is not clear either; we could just say "the number of lists of oracle responses, outgoing messages, events, or blobs"?

linera-chain/src/block.rs
MathieuDutSik: Not possible, but we can put it as `pub(crate)`.

linera-chain/src/block_tracker.rs
MathieuDutSik: Ok.

afck:
Not sure about the PR title. Did you mean "Make"?

deuszx (approved):

PR #4336, opened 2025-08-13
Put multiple OpenChain operations in a faucet block.

### LINKED ISSUE #4334: Faucet: Allow multiple `OpenChain` per block.
The faucet could be a lot faster if it put more than one `OpenChain` operations in each block, whenever more than one requests are incoming.

afck:
## Motivation

The rate at which the faucet can open new user chains is currently very limited because it puts only one `OpenChain` operation in each block.

## Proposal

Batch the operations. (Written mostly by Claude.)

## Test Plan

New tests were added.

## Release Plan

## Links

- Closes #4334.


linera-service/src/cli/command.rs
```diff
@@ -750,6 +750,10 @@ pub enum ClientCommand {
         /// Path to the persistent storage file for faucet mappings.
         #[arg(long)]
         storage_path: Option<PathBuf>,
+
+        /// Maximum number of operations to include in a single block (default: 100).
+        #[arg(long, default_value = "100")]
+        max_batch_size: usize,
```
deuszx, line 746: Maybe it's safer to do it just for the faucet at first and be it a client-side feature?

afck: I don't understand: This argument _is_ for the client, isn't it?

linera-service/tests/linera_net_tests.rs
```diff
@@ -4327,6 +4327,95 @@ async fn test_end_to_end_faucet_with_long_chains(config: impl LineraNetConfig) -
...
+    // Verify that fewer than NUM_REQUESTS blocks were created, i.e. some of them were batched.
+    let port = get_node_port().await;
+    let service = client1.run_node_service(port, ProcessInbox::Skip).await?;
+    let query =
+        format!("query {{ chain(chainId:\"{chain1}\") {{ tipState {{ nextBlockHeight }} }} }}");
+    let response = service.query_node(query).await?;
+    let height = response["chain"]["tipState"]["nextBlockHeight"]
+        .as_u64()
+        .unwrap();
+    assert!(height < NUM_REQUESTS as u64);
```
deuszx, line 4411: This doesn't really verify that we can batch up tp `max_batch_size` operations per block. Is it too low-level of the test to write?

deuszx: I see it's a different test. So this test is a bit hand-wave'y. Do you think we need it given you have other tests added?

afck: Yes, in the end-to-end test we can't really guarantee that they all arrive "simultaneously enough" that they all end up in a single block. I'm not sure how to make that deterministic.

afck: True, we could remove the test. Not sure what's best.

linera-faucet/server/src/lib.rs
```diff
@@ -182,72 +211,249 @@ where
...
+            let config = OpenChainConfig {
+                ownership: ChainOwnership::single(request.owner),
+                balance: self.config.amount,
+                application_permissions: ApplicationPermissions::default(),
+            };
+            operations.push(Operation::system(SystemOperation::OpenChain(config)));
+        }
+
+        // Execute all operations in a single block
+        let result = client.execute_operations(operations, vec![]).await?;
```
deuszx: It looks that if we fails here due to an "out of gas" error – b/c we constructed a batch too big – then we never retry and recover?

linera-faucet/server/src/lib.rs
```diff
@@ -182,72 +211,249 @@ where
...
+where
+    C: ClientContext + 'static,
+{
+    /// Creates a new batch processor
+    fn new(
+        config: BatchProcessorConfig,
+        context: Arc<Mutex<C>>,
+        faucet_storage: Arc<Mutex<FaucetStorage>>,
+        pending_requests: Arc<Mutex<VecDeque<PendingRequest>>>,
+        request_notifier: Arc<Notify>,
```
deuszx, line 259: Why not implement it as a bounded channel with a bound of `max_batch_size`, spawn the processor to a different tokio thread where it polls and when it reaches some "low" timer or a `max_batch_size` - constructs a block. You wouldn't need the `Arc<Notify>` and the `Mutex<VecDeque` synchronization

linera-faucet/server/src/lib.rs
afck: Yes, but I don't think that's realistically going to happen on the faucet with 100 `OpenChain` messages. If it runs out of gas then it's also out of funds for new chains anyway.

linera-faucet/server/src/lib.rs
afck: > when it reaches some "low" timer

I think we should really optimize latency of the most common case: That there is only one request at a time.

deuszx: I meant that we can run a ticker at 100-200ms and we propose a new block every 200ms regardless of the queue length but we cap at `max_batch_size`.

linera-faucet/server/src/lib.rs
deuszx: What I mean by `OutOfGas` is that operations for the single block proposal try to take more gas than is allowed.

linera-faucet/server/src/lib.rs
afck: That adds 100ms latency to the single request case. Not sure if it would really be an improvement in any other case.

deuszx: So what is being batched if we notify on every incoming request and we propose immediately

linera-faucet/server/src/lib.rs
afck: But that's just a matter of configuring the faucet so that the configured number of operations fits in the budget. And in practice, I think 100 will always be fine?

linera-faucet/server/src/lib.rs
afck: If while the first block is getting finalized multiple requests come in, those do end up being batched in the second block.

linera-faucet/server/src/lib.rs
deuszx: I don't know. Can't we catch the error, remove some requests and retry?

afck: OK, I'll try.
But if that happens it should probably change the `max_batch_size` parameter, otherwise it's guaranteed to happen every time.

deuszx: Yes, if that was the batch size.

afck: I went through the `ExecutionError` variants and matched the ones related to limits: https://github.com/linera-io/linera-protocol/pull/4336/commits/ef1d6d874eaa29cd1b51c97d837af33de2ef45c8

Since all `OpenChain` operations should incur the same costs, this reduces the `max_batch_size` to the index of the failed operation.

deuszx: 🧠

linera-faucet/server/src/lib.rs
```diff
@@ -182,72 +212,271 @@ where
...
+                ) =>
+            {
+                tracing::error!(%exec_err, "Execution of operation {i} failed; reducing batch size");
+                self.config.max_batch_size = i as usize;
+                // Put the valid requests back into the queue.
+                let mut pending_requests = self.pending_requests.lock().await;
+                for request in valid_requests.into_iter().rev() {
+                    pending_requests.push_front(request);
+                }
+                return Ok(()); // Don't return an error, so we retry.
```
deuszx: Nit: I don't think this should be part of `execute_batch` method. I think execution should be independent from `BatchProcessor`'s config manipulation. Not a blocker though.

linera-faucet/server/src/lib.rs
afck: Turns out this is the wrong error variant. Will push a fix in a few minutes.

linera-faucet/server/src/tests.rs
```diff
@@ -88,105 +88,458 @@ async fn test_faucet_rate_limiting() {
...
+    let result7 = root.do_claim(AccountPublicKey::test_key(6).into()).await;
+    assert!(
+        result7.is_err(),
+        "Claim should fail when validators are offline"
+    );
+
+    // Verify update_wallet calls (includes successful operations and final error case)
+    let update_calls = context.lock().await.update_calls;
+    assert!(
+        update_calls >= 3,
```
deuszx: Are we not hitting 4 consistently anymore? I thought we're doing claims one-by-one here.

linera-faucet/server/src/lib.rs
afck: I can give it a try next week, but for now it's difficult because this uses `anyhow` errors. We'd have to create a new error type or try to typecast the cause, which makes the error matching even more cumbersome than it already is.

Maybe we should take a look at our `linera-core` error types in the first place, to make this simpler. I wonder if `LocalNodeError` was a mistake, and it should be merged back into `ChainClientError`.

linera-faucet/server/src/tests.rs
```diff
@@ -88,105 +88,458 @@ async fn test_faucet_rate_limiting() {
...
+    let batch_config = super::BatchProcessorConfig {
         chain_id,
-        context: context.clone(),
         amount: Amount::from_tokens(1),
         end_timestamp: Timestamp::from(6000),
         start_timestamp: Timestamp::from(0),
         start_balance: Amount::from_tokens(6),
-        faucet_storage,
         storage_path,
+        max_batch_size: 10,
```
deuszx: If we want to test the same behavior as previously setting this to 1 is more appropriate.

linera-faucet/server/src/tests.rs
afck: We call it only once per block, not once per claim. But you're right, we could just set the batch size to 1 in this test, then it should be 4 again.

linera-faucet/server/src/tests.rs
```diff
@@ -88,105 +88,458 @@ async fn test_faucet_rate_limiting() {
...
+    }
+
+    assert_eq!(processed, vec![1, 2, 3]);
+
+    // Queue should be empty now
+    {
+        let requests = pending_requests.lock().await;
+        assert_eq!(requests.len(), 0);
+    }
+}
```
deuszx: This tests `VecDequeu` implementation - I'm not sure it's needed.

linera-faucet/server/src/tests.rs
```diff
@@ -88,105 +88,458 @@ async fn test_faucet_rate_limiting() {
...
+    use std::path::PathBuf;
+
+    use linera_base::data_types::{Amount, Timestamp};
+
+    // Test that we can create amounts and compare them
+    let small_amount = Amount::from_tokens(100);
+    let large_amount = Amount::from_tokens(10000);
+
+    assert!(large_amount > small_amount);
+    assert_eq!(small_amount, Amount::from_tokens(100));
```
deuszx: necessary?

linera-faucet/server/src/tests.rs
```diff
@@ -88,105 +88,458 @@ async fn test_faucet_rate_limiting() {
...
+
+    // Test path creation
+    let path = PathBuf::from("/tmp/test_faucet.json");
+    assert_eq!(path.extension().unwrap(), "json");
+
+    // Test batch size validation logic
+    let max_batch_size = 50;
+    assert!(max_batch_size > 0);
+    assert!(max_batch_size <= 100); // reasonable upper bound
+}
```
deuszx: Whole test looks unnecessary.

linera-faucet/server/src/tests.rs
```diff
@@ -88,105 +88,458 @@ async fn test_faucet_rate_limiting() {
...
+    tokio::time::sleep(Duration::from_millis(10)).await;
+
+    // Send notification
+    notifier.notify_one();
+
+    // Task should complete
+    let result = tokio::time::timeout(Duration::from_millis(100), handle).await;
+    assert!(result.is_ok());
+    assert_eq!(result.unwrap().unwrap(), "notified");
+}
```
deuszx, line 188: This tests `std::sync::Notify`.

linera-faucet/server/src/tests.rs
```diff
@@ -88,105 +88,458 @@ async fn test_faucet_rate_limiting() {
...
+    for owner in &queued_owners {
+        assert!(
+            loaded_storage.get_chain(owner).is_some(),
+            "Loaded storage should contain owner"
+        );
+    }
+}
+
+#[tokio::test]
+async fn test_batch_processor_error_handling() {
```
deuszx: This test is unnecessary as well - it doesn't test anything. Also, it claims to test error handling but doesn't do anything with it.

linera-faucet/server/src/lib.rs
ma2bd: What's the problem with `LocalNodeError` ?

afck: It's just that we have `ChainClientError::ChainError` but also `ChainClientError::LocalNodeError(WorkerError(ChainError))`, and I matched the wrong one at first. Maybe it's okay, but it's cumbersome to match the right thing here (also because of the `Box`.)

linera-faucet/server/src/tests.rs
```diff
@@ -88,114 +89,208 @@ async fn test_faucet_rate_limiting() {
...
+    let update_calls = context.lock().await.update_calls;
+    assert_eq!(update_calls, 3);
+
+    // Clean up
+    cancellation_token.cancel();
+    let _ = processor_task.await;
+}
+
+#[tokio::test]
+async fn test_faucet_storage_basic() {
```
deuszx: This test still seems totally unneccessary.

linera-faucet/server/src/tests.rs
```diff
@@ -88,114 +89,208 @@ async fn test_faucet_rate_limiting() {
...
+    // Test: too many claims should eventually fail
+    let result6 = root.do_claim(AccountPublicKey::test_key(5).into()).await;
+    assert!(
+        result6.is_err(),
+        "Fifth claim should fail due to rate limit"
+    );
+
+    // Verify update_wallet calls (includes successful operations and final error case)
+    let update_calls = context.lock().await.update_calls;
+    assert_eq!(update_calls, 3);
```
deuszx, line 172: So we no longer register a call for errors?

linera-faucet/server/src/tests.rs
```diff
@@ -88,114 +89,208 @@ async fn test_faucet_rate_limiting() {
...
+    storage.save(&storage_path).await.unwrap();
+    assert!(storage_path.exists());
+
+    // Test loading saved empty storage
+    let loaded_storage = super::FaucetStorage::load(&storage_path).await.unwrap();
+    assert!(loaded_storage.owner_to_chain.is_empty());
+}
+
+#[test]
+fn test_multiply() {
```
deuszx, line 180: I wonder if there's a way to rewrite this to a prop test? maybe turn both numbers to u256 and multiply then compare with the result of `super::multiply`?

linera-faucet/server/src/tests.rs
afck: No need to update the wallet if no block was produced, no.

linera-faucet/server/src/tests.rs
```diff
@@ -88,114 +89,208 @@ async fn test_faucet_rate_limiting() {
...
+        Arc::clone(&request_notifier),
+    );
 
-    // Second claim with the same owner should return the same chain
-    let second_result = root.do_claim(test_owner).await;
-    assert!(second_result.is_ok());
-    let second_description = second_result.unwrap();
+    // Create 3 different owners for batch processing
+    let owners = [
+        AccountOwner::from_str("0x1111111111111111111111111111111111111111").unwrap(),
```
deuszx: nit: I prefer `CryptoHash::new(&TestString("owner1")).into()`. It's shorter and more descriptive.

linera-faucet/server/src/tests.rs
```diff
@@ -88,114 +89,187 @@ async fn test_faucet_rate_limiting() {
...
+    }
 
-#[test]
-fn test_multiply() {
-    use super::multiply;
+    // Execute the batch - this triggers BlockTooLarge error
+    batch_processor
+        .process_batch()
+        .await
+        .expect("Batch processing should succeed");
```
deuszx, line 269: What is the reason this batch is too large? `maximum_block_size`?

linera-faucet/server/src/tests.rs
afck: Maybe, yes.

But this test wasn't even modified in this PR, only moved. I can move it back, but I do think it makes sense to have it in the `tests` module.

linera-faucet/server/src/tests.rs
afck: Yes. I couldn't come up with a better way to cause one of those error types.

linera-faucet/server/src/lib.rs
deuszx: OK, I'd put that into the **Proposal** then that this is the intended design.

linera-faucet/server/src/lib.rs
```diff
@@ -78,6 +89,33 @@ pub struct Validator {
...
+    owner: AccountOwner,
+    responder: oneshot::Sender<Result<ChainDescription, Error>>,
+}
+
+/// Configuration for the batch processor
+struct BatchProcessorConfig {
+    chain_id: ChainId,
+    amount: Amount,
+    end_timestamp: Timestamp,
+    start_timestamp: Timestamp,
```
deuszx, line 105: Maybe it's my OCD but I'd put start _before_ end 🤔

linera-faucet/server/src/lib.rs
```diff
@@ -182,72 +214,282 @@ where
...
+            valid_requests.push(request);
+        }
 
-        if self.start_timestamp < self.end_timestamp {
+        if valid_requests.is_empty() {
+            return Ok(());
+        }
+
+        // Rate limiting check for the batch
+        let client = self
```
deuszx: Do we need to recreate the client every time we want to propose a block? It's always for the same chain.

linera-faucet/server/src/lib.rs
```diff
@@ -182,72 +214,282 @@ where
...
+            .context
+            .lock()
+            .await
+            .make_chain_client(self.config.chain_id);
+
+        if self.config.start_timestamp < self.config.end_timestamp {
```
deuszx, line 346: nit:
Since we never update these values it seems like the check could have been made once at the startup only. 

Also - we never check if `local_time > start_timestamp` - seems like an oversight.

linera-faucet/server/src/lib.rs
```diff
@@ -182,72 +214,282 @@ where
...
+                            ) =>
+                    {
+                        tracing::error!(%exec_err, "Execution of operation {i} failed; reducing batch size");
+                        self.config.max_batch_size = i as usize;
+                        // Put the valid requests back into the queue.
+                        let mut pending_requests = self.pending_requests.lock().await;
+                        for request in valid_requests.into_iter().rev() {
+                            pending_requests.push_front(request);
+                        }
+                        return Ok(()); // Don't return an error, so we retry.
```
deuszx, line 430: Is it OK to call `update_wallet` (we do that earlier) for the case when we end up retrying? if we didn't propose and there's no timeout, there's no need to update it.

linera-faucet/server/src/lib.rs
```diff
@@ -182,72 +214,282 @@ where
...
+        if chain_descriptions.len() != valid_requests.len() {
+            let error_msg = format!(
+                "Mismatch between operations ({}) and results ({})",
+                valid_requests.len(),
+                chain_descriptions.len()
+            );
+            for request in valid_requests {
+                let _ = request.responder.send(Err(Error::new(error_msg.clone())));
+            }
+            return Err(anyhow::anyhow!(error_msg));
```
deuszx, line 472: What if we had multiple requests to open the same chain - we'll probably end up opening only one (with the introduction of the recent feature that stores the owner->chain_id mapping) so this will trigger the error.

afck: No, we'd end up creating multiple chains, since the key is not in the map yet.

I don't think that's a problem, but we could deduplicate requests in each batch to address it. (Maybe in a separate PR?)

linera-faucet/server/src/lib.rs
afck: First of all there's no harm in doing so, apart from a bit of additional disk I/O.

But more importantly, if we made a proposal to the validators, we should _always_ save the wallet, even if the proposal failed. We might have a pending block now that we need to remember.

linera-faucet/server/src/lib.rs
afck: It's not necessary: The inequality below will fail in that case, since `remaining_duration` will equal `full_duration`.

linera-faucet/server/src/lib.rs
deuszx: Maybe in general case but we're talking about a very specific one here.

linera-faucet/server/src/lib.rs
deuszx: Yes, but it's useless check to make on every proposal.

linera-faucet/server/src/lib.rs
afck: Done in https://github.com/linera-io/linera-protocol/pull/4336/commits/166e8cb6ee11a2e28d8bf395f5a910c2241a84b8.

linera-faucet/server/src/lib.rs
```diff
@@ -182,72 +215,282 @@ where
...
+                    .amount
+                    .saturating_mul(valid_requests.len() as u128);
+                let Ok(remaining_balance) = balance.try_sub(total_amount) else {
+                    // Not enough balance - reject all requests
+                    for request in valid_requests {
+                        let _ = request
+                            .responder
+                            .send(Err(Error::new("The faucet is empty.")));
+                    }
+                    return Ok(());
```
deuszx, line 372: Shouldn't we try to handle as many requests as possible rather than failing the whole batch? Maybe we can handle `n-1` requests and the last one fails the batch.

afck: Is that really worth the added complexity?
The period between start and end time will be on the order of months, and the block time on the order of seconds or less.

linera-faucet/server/src/lib.rs
afck: Right, that's true. But this case is going to happen at most _once_ during the entire uptime of a faucet process, and even that only if it was misconfigured. I'd say it's not worth adding more lines of code just for that.

linera-faucet/server/src/lib.rs
afck: But an extremely cheap one. Where else would we make the check? Wouldn't we end up just checking a boolean here instead? Is that even faster, or are both a single CPU cycle?

deuszx: I just think it's unnecessary _mental_ complexity to the reader. I like to simplify code not only for the machine. It's not a blocker though ofc, especially that it was like this before this PR. 

Gonna resolve.

linera-faucet/server/src/lib.rs
```diff
@@ -1,10 +1,17 @@
 // Copyright (c) Zefchain Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
+#![recursion_limit = "256"]
```
bart-linera, line 4: Is this necessary?

afck: Yes. I haven't investigated further, but when I tried to remove it it didn't build anymore.

bart-linera: That's interesting. It's usually necessary if there are some deeply nested types in the code, IIRC, but I don't see that happening here :thinking:

PR #4337, opened 2025-08-13
Restrict the scope of some variables and functions.

MathieuDutSik:
## Motivation

An examination of the code indicates that some variables and functions should have a restricted scope.

## Proposal

The changed files are self-explanatory.

## Test Plan

CI.

## Release Plan

## Links

None

### DIFF
--- a/linera-views/src/common.rs
+++ b/linera-views/src/common.rs
@@ -16,7 +16,7 @@ use serde::de::DeserializeOwned;
 use crate::ViewError;
 
 #[doc(hidden)]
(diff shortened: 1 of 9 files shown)


linera-views/src/common.rs
```diff
@@ -16,7 +16,7 @@ use serde::de::DeserializeOwned;
 use crate::ViewError;
 
 #[doc(hidden)]
```
ma2bd: We can remove doc hidden then.

ma2bd (approved):

linera-views/src/common.rs
MathieuDutSik: Ok, thanks.

PR #4341, opened 2025-08-14
Elide unused lifetime.

MathieuDutSik:
## Motivation

Curiously, it occurred on clippy locally, but not on CI.

## Proposal

Do, the needed correction.

## Test Plan

The CI.

## Release Plan

## Links

None


afck (approved):
I saw that today, too. But we haven't changed this part of the code, and neither did we change the Rust version. :raised_eyebrow:

PR #4342, opened 2025-08-14
Test the existence of an application

MathieuDutSik:
## Motivation

For the complete support of EVM balances, we need to be able to test the existence of an application.

## Proposal

The implementation was relatively straightforward:
* We added the check to the `system.rs` code.
* Premature termination occurs if the `created_blob()` contains the blob.

## Test Plan

The tests have been added to the `create-and-call` function.

## Release Plan

## Links

None.

### DIFF
--- a/linera-execution/src/runtime.rs
+++ b/linera-execution/src/runtime.rs
@@ -972,7 +972,11 @@ where
     ) -> Result<bool, ExecutionError> {
         let this = self.inner();
         let blob_id = application_id.description_blob_id();
-        if this.transaction_tracker.created_blobs().contains_key(&blob_id) {
+        if this
+            .transaction_tracker
+            .created_blobs()
+            .contains_key(&blob_id)
+        {
             return Ok(true);
         }
         let exists = this


linera-execution/src/execution_state_actor.rs
```diff
@@ -490,6 +490,23 @@ where
...
+                callback,
+            } => {
+                let exists = match self
+                    .system
+                    .describe_application(application_id, &mut TransactionTracker::default())
+                    .await
+                {
+                    Ok(_) => true,
+                    Err(ExecutionError::BlobsNotFound(_)) => false,
+                    Err(ExecutionError::ViewError(_)) => false,
```
afck, line 505: `SystemExecutionStateView::describe_application` only records an oracle response if it does find the application description. That means if it returns an error, that block must fail.

Do we really need to programmatically check _whether_ an application exists?
If that's required, we need to implement that differently, and record the `true` or `false` outcome as an oracle response.

Also, aren't there other `ViewErrors` (actual storage problems) that shouldn't just be interpreted as "does not exist"?

MathieuDutSik: For the support of the transfer with the `fallback` function, yes we need to check whether an application exists. But this is an edge case. Not absolutely critical.

So, I will rethink and implement it as an oracle response.

About the `ViewError`:
* Any one of those can end up the system.
* The only errors that is processed later is maybe `NotFound`. The `BlobsNotFound` / `EventsNotFound` are elsewhere.
* The absence of key results in a `None`. Errors should never lead to the conclusion that a key is absent.

ma2bd:
@MathieuDutSik Would it be sufficient for the EVM use case to test if an application has been used on the local chain at least once already? This would be simpler to implement and would work with fast blocks.

MathieuDutSik:
> @MathieuDutSik Would it be sufficient for the EVM use case to test if an application has been used on the local chain at least once already? This would be simpler to implement and would work with fast blocks.

I don't think so. Better to do it right. 

That being said, this is not an essential requirement. So, if we do it, let's just do it right.

ma2bd:
We're discussing what "right" means here.

MathieuDutSik:
This does not represent a way that we want to continue on.

Still, it could make sense to have an Oracle for testing the existence of Blobs.

PR #4343, opened 2025-08-15
Eliminate the `Recipient` type.

### LINKED ISSUE #3580: Work on removal of `Recipient::Burn`
We can probably kill this `Burn` thing by the way and simplify (away?) `Recipient` now. People will be able to send to address `0x0` (or anything really) if they really want to burn tokens.

_Originally posted by @ma2bd in https://github.com/linera-io/linera-protocol/pull/3573#discussion_r1997674781_

MathieuDutSik:
## Motivation

The Recipient type variant `Burn` is used only for tests. So, we remove it. This let the `Recipient` having just a variant `Account`. And so the `Recipient` type is eliminated.

Fixes #3580 

## Proposal

Straightforward from the description.

## Test Plan

The CI.
The tests were restructured accordingly. For some tests the balance is now unchanged. For some tests we introduced another chain.

## Release Plan

## Links

None.

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -2745,10 +2745,8 @@ impl<Env: Environment> ChainClient<Env> {
         amount: Amount,
         account: Account,
     ) -> Result<ClientOutcome<ConfirmedBlockCertificate>, ChainClientError> {
-        self.transfer(from, amount, account)
-            .await
+        self.transfer(from, amount, account).await
     }
-
 
     /// Attempts to synchronize chains that have sent us messages and populate our local
     /// inbox.
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -2749,15 +2749,6 @@ impl<Env: Environment> ChainClient<Env> {
             .await
     }
 
-    /// Burns tokens.
-    #[instrument(level = "trace")]
-    pub async fn burn(
-        &self,
-        owner: AccountOwner,
-        amount: Amount,
-    ) -> Result<ClientOutcome<ConfirmedBlockCertificate>, ChainClientError> {
-        self.transfer(owner, amount, Recipient::Burn).await
-    }
 
     /// Attempts to synchronize chains that have sent us messages and populate our local
     /// inbox.
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -2748,6 +2748,18 @@ impl<Env: Environment> ChainClient<Env> {
         self.transfer(from, amount, account).await
     }
 
+    /// Burns tokens (transfer to a special address).
+    #[cfg(with_testing)]
+    #[instrument(level = "trace")]
+    pub async fn burn(
+        &self,
+        owner: AccountOwner,
+        amount: Amount,
+    ) -> Result<ClientOutcome<ConfirmedBlockCertificate>, ChainClientError> {
+        let recipient = Account::burn_address(self.chain_id);
+        self.transfer(owner, amount, recipient).await
+    }
+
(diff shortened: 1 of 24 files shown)


ma2bd:
Good idea. Thanks!

linera-execution/tests/test_system_execution.rs
```diff
@@ -62,7 +61,7 @@ async fn test_simple_system_operation() -> anyhow::Result<()> {
     )
     .await
     .unwrap();
-    assert_eq!(view.system.balance.get(), &Amount::ZERO);
+    assert_eq!(view.system.balance.get(), &Amount::from_tokens(4));
```
afck: But now this doesn't actually check whether the transfer has happened or not. I'd rather transfer to address zero or something, which effectively still burns it.

linera-core/src/unit_tests/client_tests.rs
```diff
@@ -421,7 +432,13 @@ where
 
     // We need at least three validators for making an operation.
     builder.set_fault_type([0, 1], FaultType::Offline).await;
-    let result = client.burn(AccountOwner::CHAIN, Amount::ONE).await;
```
afck: We could still keep the `burn` function, and make it transfer to address zero instead? (I.e. to the zero account on the same chain.)

linera-core/src/client/mod.rs
```diff
@@ -3522,7 +3511,7 @@ impl<Env: Environment> ChainClient<Env> {
     ) -> Result<ClientOutcome<ConfirmedBlockCertificate>, ChainClientError> {
         self.execute_operation(SystemOperation::Transfer {
             owner,
-            recipient: Recipient::Account(account),
+            recipient: account,
```
afck: Let's just call the parameter `recipient`, too, then.

MathieuDutSik: Ok.

linera-execution/tests/test_system_execution.rs
MathieuDutSik: Ok, done that way.

linera-core/src/unit_tests/client_tests.rs
MathieuDutSik: Ok, done that way.

linera-core/src/unit_tests/client_tests.rs
```diff
@@ -1503,7 +1501,11 @@ where
         *info2_b.manager.requested_locking.unwrap()
     );
     let bt_certificate = client_2b
-        .burn(AccountOwner::CHAIN, Amount::from_tokens(1))
+        .transfer_to_account(
```
afck: `burn`?

linera-core/src/unit_tests/client_tests.rs
```diff
@@ -1636,7 +1638,11 @@ where
 
     client2_b.prepare_chain().await.unwrap();
     let bt_certificate = client2_b
-        .burn(AccountOwner::CHAIN, Amount::from_tokens(1))
+        .transfer_to_account(
```
afck: `burn`?

afck (approved):

linera-core/src/unit_tests/client_tests.rs
MathieuDutSik: It is a little difficult since the `.burn` is pointing to a special but unknown address. But we need it to match it with the `SystemOperation::Transfer`.

So, instead I used a `recipient = Account::burn_address(chain_id)` and go from there.

linera-core/src/unit_tests/client_tests.rs
MathieuDutSik: Same as above.

PR #4345, opened 2025-08-15
Generate operations dynamically

ndr-ds:
## Motivation

Right now we always precalculate the operations for each block. However, the performance gain we get from it should be minimal, and it limits us on what we can do.

## Proposal

Generate the operations dynamically instead, giving more flexibility on what we can do.

## Test Plan

Tested against a local network, works as expected

## Release Plan


ndr-ds:
* **#4331** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4331?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4345** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4345?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4345?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#4346** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4346?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

linera-client/src/benchmark.rs
```diff
@@ -737,3 +728,38 @@ impl<Env: Environment> Benchmark<Env> {
...
+            num_destinations: destination_chains.len(),
+            destination_chains,
+            rng,
+        })
+    }
+
+    fn get_next_destination(&mut self) -> ChainId {
+        // Check if we've gone through all destinations
+        if self.destination_index >= self.num_destinations {
+            // Reshuffle the destinations for the next cycle
```
afck, line 753: ~Do we need that? Otherwise the whole destination manager could be replaced by an iterator using `cycle`.~

(I see; I saw the comment in the other PR.)

linera-client/src/benchmark.rs
```diff
@@ -737,3 +728,38 @@ impl<Env: Environment> Benchmark<Env> {
...
+}
+
+impl ChainDestinationManager {
+    fn new(mut destination_chains: Vec<ChainId>) -> Result<Self, BenchmarkError> {
+        let mut rng = SmallRng::from_rng(thread_rng())?;
+        destination_chains.shuffle(&mut rng);
+
+        Ok(Self {
+            destination_index: 0,
+            num_destinations: destination_chains.len(),
```
afck: That seems unnecessary.

linera-client/src/benchmark.rs
```diff
@@ -37,6 +37,16 @@ use crate::chain_listener::{ChainListener, ClientContext};
 const PROXY_LATENCY_P99_THRESHOLD: f64 = 400.0;
 const LATENCY_METRIC_PREFIX: &str = "linera_proxy_request_latency";
 
+/// Parameters for generating operations on the fly during benchmark execution
+pub struct BenchmarkOperationParams<Env: Environment> {
+    /// The chain client for this benchmark task
+    pub chain_client: ChainClient<Env>,
+    /// The owner of the source chain
+    pub owner: AccountOwner,
```
afck: Isn't that just `chain_client.identity().await?`?

linera-client/src/benchmark.rs
```diff
@@ -37,6 +37,16 @@ use crate::chain_listener::{ChainListener, ClientContext};
 const PROXY_LATENCY_P99_THRESHOLD: f64 = 400.0;
 const LATENCY_METRIC_PREFIX: &str = "linera_proxy_request_latency";
 
+/// Parameters for generating operations on the fly during benchmark execution
+pub struct BenchmarkOperationParams<Env: Environment> {
+    /// The chain client for this benchmark task
+    pub chain_client: ChainClient<Env>,
```
afck: This isn't really a "parameter"; maybe `BenchmarkOperationParams` isn't really worth it as a type?

linera-client/src/benchmark.rs
ndr-ds: I guess even though `identity()` does a local chain info query, it might be fine 🤔 since the View will likely be in memory

linera-client/src/benchmark.rs
```diff
@@ -601,6 +628,47 @@ impl<Env: Environment> Benchmark<Env> {
...
 
+    fn create_operation(
+        fungible_application_id: Option<ApplicationId>,
+        recipient_chain_id: ChainId,
+        owner: AccountOwner,
+        amount: Amount,
+    ) -> Operation {
+        match fungible_application_id {
+            Some(application_id) => {
+                Self::fungible_transfer(application_id, recipient_chain_id, owner, owner, amount)
```
ma2bd, line 630: The recipient seems to be the same account owner as the sender. This may create large ledgers (map views) and slow down the chains. Also, the received money won't be usable by the owner of the receiver chain. This being said, I don't have an easy fix unless we support the chain account `0x0` in fungible.

linera-service/src/cli/main.rs
```diff
@@ -863,7 +862,7 @@ impl Runnable for Job {
                             info!("Benchmark cancelled by user");
                             context
                                 .wrap_up_benchmark(
-                                    chain_clients,
+                                    chain_clients.clone(),
```
afck: Is this `clone()` necessary? We're `return`ing right after this.

linera-client/src/client_context.rs
```diff
@@ -736,14 +735,23 @@ where
...
+                    all_chains.clone()
+                } else {
+                    all_chains
+                        .iter()
+                        .filter(|&destination_chain_id| *destination_chain_id != *source_chain_id)
+                        .copied()
+                        .collect()
+                }
+            })
+            .collect();
```
afck: Why pass around this huge list of lists with _n (n - 1)_ entries instead of just the `all_chains` list with _n_ entries? We can randomly select the actual destinations later, when we actually create the operations.

linera-service/src/cli/main.rs
ndr-ds: Good catch! Was probably left over from a previous attempt

linera-client/src/client_context.rs
ndr-ds: True 😅 The tasks will have to each own the list of all chains regardless, so we can't avoid the clone, but at least we don't have to pass the huge thing around for no reason

PR #4347, opened 2025-08-16
Change the API of `execute_message`

### LINKED ISSUE #4271: Change the API of `execute_message`
Currently, the `execute_message` is taking just an `operation: Vec<u8>`.

There are other functions related to messages:
* `ContractRuntime::message_id` that returns an `Option<MessageId>`.
* `ContractRuntime::message_is_bouncing` that returns an `Option<bool>`.

The behavior is as follows:
* When the execution context is an `execute_message`, the results are some `Some(...)`.
* In other contexts, it is a `None`.

This is rather cumbersome, and it would probably be better to have the following API:
```rust
fn execute_message(&mut self, is_bouncing: bool, message_id: MessageId, message: Vec<u8>) -> Result<(), ExecutionError>
```

and remove the functions `message_is_bouncing` and `message_id` from the `ContractRuntime`.

MathieuDutSik:
## Motivation

The function `execute_message` takes a single argument. However, two additional arguments are implicit
and can be accessed by the runtime:
* is_bouncing
* message_origin_chain_id.

Their return types `Option<bool>` and `Option<ChainId>` reflect this. The solution is to have a single
function `execute_message` that takes 3 arguments.

Fixes #4271 

## Proposal

The solution was relatively straightforward to implement and yields a net simplification.

For the source of the message, several names are possible:
* `source_chain_id`.
* `message_origin_chain_id`.
* `chain_id`.
* `origin`.
I chose `origin` since it is the terminology used in `MessageContext` and is the shortest one.

The code of the smart contracts was not corrected. Ignored argument `is_bouncing` are quite apparent.
This should be corrected so that the smart contract handles the bouncing case correctly. But this is for further
work.

## Test Plan

The CI.

## Release Plan

## Links

None.

### DIFF
--- a/examples/amm/src/contract.rs
+++ b/examples/amm/src/contract.rs
@@ -55,7 +55,7 @@ impl Contract for AmmContract {
         }
     }
 
-    async fn execute_message(&mut self, message: Self::Message) {
+    async fn execute_message(&mut self, _is_bouncing: bool, origin: ChainId, message: Self::Message) {
         assert_eq!(
             self.runtime.chain_id(),
             self.runtime.application_creator_chain_id(),
@@ -93,7 +93,7 @@ impl Contract for AmmContract {
                 self.transfer(owner, input_amount, amm_account, input_token_idx);
 
                 let amm_app_owner = self.get_amm_app_owner();
-                let message_origin_account = self.get_message_origin_account(owner);
+                let message_origin_account = self.get_message_origin_account(origin, owner);
                 self.transfer(
                     amm_app_owner,
                     output_amount,
@@ -180,7 +180,7 @@ impl Contract for AmmContract {
                 }
 
                 let amm_account = self.get_amm_account();
-                let message_origin_account = self.get_message_origin_account(owner);
+                let message_origin_account = self.get_message_origin_account(origin, owner);
                 // See if we'll need to send refunds
                 if token0_amount < max_token0_amount {
                     self.transfer(
@@ -272,7 +272,7 @@ impl Contract for AmmContract {
                     self.get_shares(other_amount, token_to_remove_amount, &balance0_bigint)
                 };
 
-                let message_origin_account = self.get_message_origin_account(owner);
+                let message_origin_account = self.get_message_origin_account(origin, owner);
                 let current_shares = self
                     .current_shares_or_default(&message_origin_account)
                     .await;
@@ -295,7 +295,7 @@ impl Contract for AmmContract {
                     .check_account_permission(owner)
                     .expect("Permission for RemoveAllAddedLiquidity message");
 
-                let message_origin_account = self.get_message_origin_account(owner);
+                let message_origin_account = self.get_message_origin_account(origin, owner);
                 let current_shares = self
                     .current_shares_or_default(&message_origin_account)
                     .await;
@@ -444,12 +444,9 @@ impl AmmContract {
         }
     }
 
-    fn get_message_origin_account(&mut self, owner: AccountOwner) -> Account {
+    fn get_message_origin_account(&mut self, origin: ChainId, owner: AccountOwner) -> Account {
         Account {
-            chain_id: self
-                .runtime
-                .message_origin_chain_id()
-                .expect("Getting message origin chain ID should not fail"),
+            chain_id: origin,
             owner,
         }
     }
--- a/linera-execution/src/lib.rs
+++ b/linera-execution/src/lib.rs
@@ -333,7 +333,12 @@ pub trait UserContract {
     fn execute_operation(&mut self, operation: Vec<u8>) -> Result<Vec<u8>, ExecutionError>;
 
     /// Applies a message originating from a cross-chain message.
-    fn execute_message(&mut self, is_bouncing: bool, origin: ChainId, message: Vec<u8>) -> Result<(), ExecutionError>;
+    fn execute_message(
+        &mut self,
+        is_bouncing: bool,
+        origin: ChainId,
+        message: Vec<u8>,
+    ) -> Result<(), ExecutionError>;
 
     /// Reacts to new events on streams this application subscribes to.
     fn process_streams(&mut self, updates: Vec<StreamUpdate>) -> Result<(), ExecutionError>;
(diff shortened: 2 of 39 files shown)


deuszx:
I'm not convinced this is a necessary change. Just like origin/sender etc are accessed via runtime similarly the origin chain id (and is_bouncing) can be there. In the API we have what's necessary and nothing more. Not every contract will use these new arguments and so they will have to ignore it.

ma2bd:
I like the argument `is_bouncing` because it is likely going to be used immediately and people should know about bouncing messages. For the other one, we need to look at the resulting changes in examples.

MathieuDutSik:
> I like the argument `is_bouncing` because it is likely going to be used immediately and people should know about bouncing messages. For the other one, we need to look at the resulting changes in examples.

Yes, for `is_bouncing` that seems clear.

Now, regarding the `source_chain_id`:
* There are 6 contracts that do not use it and 4 that use it. Therefore, I think there are valid cases where it matters and valid cases where it does not matter.
* The `source_chain_id` can be put in the message itself. It is also possible to put other information: original block timestamp, height, etc.

Therefore, my proposal is:
* Add the `ChainId` in the message of the smart contract that uses it.
* Remove the `origin: ChainId` from the `execute_message` prototype.
* Keep the `is_bouncing` in the prototype.

ma2bd:
> Add the ChainId in the message of the smart contract that uses it.

Definitely not. This is a waste of storage and network (because the information is already present)

examples/amm/src/contract.rs
```diff
@@ -444,12 +449,9 @@ impl AmmContract {
         }
     }
 
-    fn get_message_origin_account(&mut self, owner: AccountOwner) -> Account {
+    fn get_message_origin_account(&mut self, origin: ChainId, owner: AccountOwner) -> Account {
```
ma2bd: inline this function

linera-execution/src/lib.rs
```diff
@@ -693,13 +698,6 @@ pub trait ContractRuntime: BaseRuntime {
...
-    /// If the current message (if there is one) was rejected by its destination and is now
-    /// bouncing back.
-    fn message_is_bouncing(&mut self) -> Result<Option<bool>, ExecutionError>;
-
-    /// The chain ID where the current message originated from, if there is one.
-    fn message_origin_chain_id(&mut self) -> Result<Option<ChainId>, ExecutionError>;
-
     /// The optional authenticated caller application ID, if it was provided and if there is one
     /// based on the execution context.
     fn authenticated_caller_id(&mut self) -> Result<Option<ApplicationId>, ExecutionError>;
```
ma2bd, line 703: FYI these functions use to be in a `Context` structure and we moved them out because it felt cleaner (than passing around a context).

For this PR to go through, we need to agree that message-specific context values are an exception AND we don't plan to add more context values in the future for messages.

examples/amm/src/contract.rs
MathieuDutSik: Done.

linera-execution/src/lib.rs
ma2bd: Incidentally, we need to clarify in the doc what the origin of a bouncing message is.

examples/amm/src/contract.rs
```diff
@@ -55,7 +55,12 @@ impl Contract for AmmContract {
...
     }
 
-    async fn execute_message(&mut self, message: Self::Message) {
+    async fn execute_message(
+        &mut self,
+        _is_bouncing: bool,
+        origin: ChainId,
+        message: Self::Message,
+    ) {
         assert_eq!(
```
ma2bd, line 64: Note that this assert would be wrong on bouncing messages. So to be consistent, we might also add `assert!(!is_bouncing, "This application doesn't use bouncing messages.")`

linera-execution/src/lib.rs
MathieuDutSik: That seems indeed a big ask.

Because I can very well conceive that we pass more arguments to the function `execute_message`.
Potentially, all the entries of the `MessageContext` could be passed, and there is more than `origin` and `is_bouncing`. So, yes, we could add more context values.

Possible options:
* Have a `fn execute_message(context: Context, message: Vec<u8>)` which leaves room for extending the values passed.
* Just add the `is_bouncing` entry.

examples/amm/src/contract.rs
MathieuDutSik: But, here is the problem: Shouldn't the assert statement be `assert!(!is_bouncing, "Bouncing messages are not yet supported")`?

Not processing bouncing messages is fine for `Message::RemoveLiquidity` and `Message::RemoveAllAddedLiquidity` since no operation of consequence is done before that needs to be reverted.

However, for `Message::AddLiquidity` and `Message::Swap`, there is a transfer operation before the message emission. So, what about that operation?

In all transfers, the transfers are done to the creator chain of the AMM. So, the creator chain of the AMM receives two kinds of messages:
* The ones of transfer
* The ones of the AMM.

I think the only reasonable design is that all messages are bounced or none of them (is that the case @afck?). Therefore, the transfers are reverted by the bouncing messages, and so it should be ok. And so the correct handling of bouncing messages for the AMM should be, I think
```rust
if is_bouncing {
    return;
}
```

Then the assert statement would work fine.

ma2bd: An app cannot receive "bouncing messages" if it doesn't create "tracked messages".

If `is_bouncing` is confusing, then that's an argument in favor of hiding it until people need it.

MathieuDutSik: Thank you for correcting me.
I would argue that if it forces clarification on the `is_bouncing` then it is a desirable outcome.
But of course, we can abort the PR as well.

linera-execution/src/lib.rs
```diff
@@ -693,13 +698,6 @@ pub trait ContractRuntime: BaseRuntime {
     /// The authenticated signer for this execution, if there is one.
     fn authenticated_signer(&mut self) -> Result<Option<AccountOwner>, ExecutionError>;
 
-    /// If the current message (if there is one) was rejected by its destination and is now
-    /// bouncing back.
-    fn message_is_bouncing(&mut self) -> Result<Option<bool>, ExecutionError>;
-
-    /// The chain ID where the current message originated from, if there is one.
-    fn message_origin_chain_id(&mut self) -> Result<Option<ChainId>, ExecutionError>;
```
ma2bd, line 701: Alternatively, we could just remove the options here: calling those APIs outside of a message would return an error.

ma2bd: @MathieuDutSik What do you think? It seems like it would simplify the Solidity code a but already.

ma2bd: Never mind, I see that the `Result` type doesn't exist in the SDK. It's really just an option.

MathieuDutSik:
That direction of change did not reach consensus and so the PR is killed.

PR #4354, opened 2025-08-18
Merge "attempted" and "temporary" changes into a single `ChainWorkerGuard` type.

afck:
## Motivation

The distinction between `ChainWorkerStateWithAttemptedChanges` and `ChainWorkerStateWithTemporaryChanges` is unnecessary: The `succeeded` field might as well be omitted, because calling `rollback` after saving is a no-op. It's also a bit misleading, because if one calls `save` in `ChainWorkerStateWithTemporaryChanges`, the changes _will_ be permanent.

## Proposal

Remove `succeeded`. Replace the two types with a single `ChainWorkerGuard`.

This will enable other cleanups, but for clarity I'd do those in a separate PR. In this one, the methods have just been moved together into a single type.

## Test Plan

CI

## Release Plan

## Links


linera-core/src/chain_worker/state/guard.rs
```diff
@@ -704,30 +687,319 @@ where
...
+        Ok(outcome)
+    }
+
     /// Stores the chain state in persistent storage.
     ///
     /// Waits until the [`ChainStateView`] is no longer shared before persisting the changes.
     async fn save(&mut self) -> Result<(), WorkerError> {
-        self.state.clear_shared_chain_view().await;
-        self.state.chain.save().await?;
-        self.succeeded = true;
```
ma2bd, line 718: Also, any accidental change after calling `save()` would not be rollback because `succeeded` was now true.

afck: Yes, that was actually a bit dangerous.

PR #4355, opened 2025-08-18
Simplify worker state implementation.

afck:
## Motivation

https://github.com/linera-io/linera-protocol/pull/4354 unified the chain worker guard. With the original distinction removed, some functions can be merged now.

## Proposal

Simplify the implementation.

## Test Plan

CI (only refactoring)

## Release Plan

## Links

### DIFF
--- a/linera-core/src/chain_worker/state/guard.rs
+++ b/linera-core/src/chain_worker/state/guard.rs
@@ -832,6 +832,8 @@ where
             .update_counters(&block.transactions, &outcome.messages)?;
         // Verify that the resulting chain would have no unconfirmed incoming messages.
         chain.validate_incoming_bundles().await?;
+        // Don't save the changes since the block is not confirmed yet.
+        chain.rollback();
 
         // Create the vote and store it in the chain state.
         let block = outcome.with(proposal.content.block.clone());


linera-core/src/chain_worker/state/guard.rs
```diff
@@ -883,15 +832,43 @@ where
             .update_counters(&block.transactions, &outcome.messages)?;
         // Verify that the resulting chain would have no unconfirmed incoming messages.
         chain.validate_incoming_bundles().await?;
-        Ok((outcome, local_time))
+        // Don't save the changes since the block is not confirmed yet.
+        chain.rollback();
```
deuszx, line 836: This is something we didn't have to do when these functions weren't merged.

MathieuDutSik: Yes, but as indicated in previous PR, if the state has not changed then the rollback is a noop.

afck: Yes, but is it really better to have two calls just so that the state is implicitly rolled back, rather than an explicit `rollback`?

deuszx: I meant that _here_ it had to be called b/c otherwise we'd have saved changes at a wrong moment.

PR #4356, opened 2025-08-18
Resolve the blob persistence problem

### LINKED ISSUE #4279: Provide access to the the existing blobs in `runtime.read_data_blob`
We face the problem that in the code of the contract, the following does not work
```rust
let blob_id = runtime.create_data_blob(...)
runtime.read_data_blob(....)
```

because the `read_data_blob` provides access to the data in the previous block, not to the one being built.

However, the solution is not to first check the existence in `created_blobs`. This is because for services, the blobs to be searched are only in the published one, not in the block that is under creation.

A possible solution to that could be to add a function `read_data_blob` to the `ContractRuntime` and `ServiceRuntime` but not to the `BaseRuntime`.

----

The changes of PR https://github.com/linera-io/linera-protocol/pull/4278 lead us to the following situation:

* Within the same operation: we have access to the just created blobs.
* Within the same block, but not the same operation, we do not have access to the created blobs.
* Within different blocks, we have access to previously created blobs.

This is not a very natural situation.

### LINKED ISSUE #4287: Make created blobs available immediately after creation.
This was addressed in #4278, but:
* only for bytecode and application description blobs, and
* only within the same transaction, but not later transaction in the same block.

MathieuDutSik:
## Motivation

Following PR #4247, PR #4278, and PR #3382, we face the scenario where a contract can create a blob
but not have it available right away. This concerns the access to Blob Bytecode, Blob ApplicationDescription, and
Blob Data.

This PR addresses those problems.

Fixes #4287 
Fixes #4279 

## Proposal

This PR addresses the problem with the data blob. But it also addresses the issues for the ApplicationDescription blob
and the contract blobs.

There are three scenarios to cover:
* The blob is created in one block and read in another block.
* The blob is created in one operation and read in the same operation.
* The blob is created in two different operations of the same block.

The first scenario is already working correctly.

The second scenario is addressed with the addition of the test of `created_blobs` in `ReadBlobContent`. This is done
in the same way as for access to the ApplicationDescription and the Contract/Service.

The third scenario is addressed by the change in how the TransactionTracker is built in `block_tracker`.

## Test Plan

The CI and the test in `test_wasm_end_to_end_publish_read_data_blob`, which considers all 3 scenarios.

## Release Plan

## Links

None

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -313,6 +313,7 @@ matching-engine = { path = "./examples/matching-engine" }
 meta-counter = { path = "./examples/meta-counter" }
 native-fungible = { path = "./examples/native-fungible" }
 non-fungible = { path = "./examples/non-fungible" }
+publish-read-data-blob = { path = "./examples/publish-read-data-blob" }
 social = { path = "./examples/social" }
 
 [workspace.dependencies.aws-config]
--- a/examples/publish-read-data-blob/README.md
+++ b/examples/publish-read-data-blob/README.md
@@ -1,0 +1,11 @@
+# Publish and read a data blob
+
+This example shows how to create a blob and how to read it.
+This example should be read in conjunction with the end-to-end
+test `test_wasm_end_to_end_publish_read_data_blob`.
+
+It shows 3 scenarios:
+* Publishing and reading blobs with the publishing and reading in different
+blocks.
+* Publishing and reading blobs in the same transaction.
+* Publishing and reqding blobs in the same block but not the same transaction
--- a/examples/publish-read-data-blob/src/lib.rs
+++ b/examples/publish-read-data-blob/src/lib.rs
@@ -4,8 +4,7 @@
 /*! ABI of the Publish-Read Data Blob Example Application */
 
 use linera_sdk::{
-    linera_base_types::{ContractAbi, ServiceAbi},
-    DataBlobHash,
+    linera_base_types::{ContractAbi, DataBlobHash, ServiceAbi},
 };
 use serde::{Deserialize, Serialize};
 
--- a/examples/publish-read-data-blob/src/lib.rs
+++ b/examples/publish-read-data-blob/src/lib.rs
@@ -1,0 +1,38 @@
+// Copyright (c) Zefchain Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+/*! ABI of the Publish-Read Data Blob Example Application */
+
+use linera_sdk::{
+    linera_base_types::{ContractAbi, ServiceAbi},
+    DataBlobHash,
+};
+use serde::{Deserialize, Serialize};
+
+pub struct PublishReadDataBlobAbi;
+
+#[derive(Debug, Deserialize, Serialize)]
+pub enum Operation {
+    CreateDataBlob(Vec<u8>),
+    ReadDataBlob(DataBlobHash, Vec<u8>),
+    CreateAndReadDataBlob(Vec<u8>),
+}
+
+#[derive(Debug, Deserialize, Serialize)]
+pub enum ServiceQuery {
+    GetHash,
+    PublishDataBlob(Vec<u8>),
+    ReadDataBlob(DataBlobHash, Vec<u8>),
+    PublishAndCreateOneOperation(Vec<u8>),
+    PublishAndCreateTwoOperations(Vec<u8>),
+}
+
+impl ContractAbi for PublishReadDataBlobAbi {
+    type Operation = Operation;
+    type Response = ();
+}
+
+impl ServiceAbi for PublishReadDataBlobAbi {
+    type Query = ServiceQuery;
+    type QueryResponse = Vec<u8>;
+}
--- a/examples/publish-read-data-blob/src/lib.rs
+++ b/examples/publish-read-data-blob/src/lib.rs
@@ -19,7 +19,6 @@ pub enum Operation {
 
 #[derive(Debug, Deserialize, Serialize)]
 pub enum ServiceQuery {
-    GetHash,
     PublishDataBlob(Vec<u8>),
     ReadDataBlob(DataBlobHash, Vec<u8>),
     PublishAndCreateOneOperation(Vec<u8>),
--- a/examples/publish-read-data-blob/src/lib.rs
+++ b/examples/publish-read-data-blob/src/lib.rs
@@ -3,9 +3,7 @@
 
 /*! ABI of the Publish-Read Data Blob Example Application */
 
-use linera_sdk::{
-    linera_base_types::{ContractAbi, DataBlobHash, ServiceAbi},
-};
+use linera_sdk::linera_base_types::{ContractAbi, DataBlobHash, ServiceAbi};
 use serde::{Deserialize, Serialize};
 
 pub struct PublishReadDataBlobAbi;
--- a/linera-chain/src/block_tracker.rs
+++ b/linera-chain/src/block_tracker.rs
@@ -176,6 +176,7 @@ impl<'resources, 'blobs> BlockExecutionTracker<'resources, 'blobs> {
             self.next_application_index,
             self.next_chain_index,
             self.oracle_responses()?,
+            &self.blobs,
         ))
     }
 
--- a/linera-execution/src/execution_state_actor.rs
+++ b/linera-execution/src/execution_state_actor.rs
@@ -70,14 +70,14 @@ where
         #[cfg(with_metrics)]
         let _latency = metrics::LOAD_CONTRACT_LATENCY.measure_latency();
         let blob_id = id.description_blob_id();
-        let description = match txn_tracker.created_blobs().get(&blob_id) {
+        let description = match txn_tracker.get_blob_content(&blob_id) {
             Some(blob) => bcs::from_bytes(blob.bytes())?,
             None => self.system.describe_application(id, txn_tracker).await?,
         };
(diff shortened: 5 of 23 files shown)


Cargo.toml
```diff
@@ -313,6 +313,7 @@ matching-engine = { path = "./examples/matching-engine" }
 meta-counter = { path = "./examples/meta-counter" }
 native-fungible = { path = "./examples/native-fungible" }
 non-fungible = { path = "./examples/non-fungible" }
+publish-read-data-blob = { path = "./examples/publish-read-data-blob" }
```
afck, line 316: Is a new example for this really necessary?
We could publish a data blob using the system operation, and then try to mint an NFT using that blob using `non-fungible`, in a separate operation in the same block.

linera-chain/src/block_tracker.rs
```diff
@@ -171,13 +171,21 @@ impl<'resources, 'blobs> BlockExecutionTracker<'resources, 'blobs> {
...
             self.oracle_responses()?,
-        ))
+        );
+
+        let mut previously_created_blobs = BTreeMap::new();
+        for tx_blobs in &self.blobs {
+            for blob in tx_blobs {
+                previously_created_blobs.insert(blob.id(), blob.content().clone());
+            }
+        }
```
afck: Why not do this at the beginning of the function, and add it as an argument to `TransactionTracker::new`?

Also, now that we're cloning all created blobs for every transaction, I really think we should prioritize turning the data in the blob content from a `Box<[u8]>` into an `Arc<[u8]>`.

linera-execution/src/execution_state_actor.rs
```diff
@@ -392,19 +392,26 @@ where
...
 
-            ReadBlobContent { blob_id, callback } => {
-                let blob = self.system.read_blob_content(blob_id).await?;
+            ReadBlobContent {
+                blob_id,
+                callback,
+                txn_tracker,
+            } => {
+                let content = match txn_tracker.get_blob_content(&blob_id) {
+                    Some(content) => content,
```
afck: Strictly speaking, if we find it in the transaction tracker, we don't need to create an oracle response entry. ~But that's currently happening at the call site. Anyway, probably an optimization for later.~

Actually: Why pass the txn tracker in and back, if we could just check whether it has the blob content at the call site instead?

Cargo.toml
MathieuDutSik: Here are 3 reasons for introducing this test:
* The contract `non-fungible` is using GraphQL, which introduces an undirection, which is unwelcome.
* In this example, the `publish-read-data-blob`, the reading is done by an operation. The standard way is to use a service. This is done to create an operation. 
* By writing it that way, we have a very clear view of the 3 scenarios in the test.

In the end, I decided to create that contract in order to be able to create two `schedule_operation` in sequence so that I had a block with two operations. As I understand it, when we do a call via `application.query_json`, then the block has only 1 operation.

linera-chain/src/block_tracker.rs
MathieuDutSik: You are right. I looked at the other contexts in which we have `TransactionTracker::new`, but they are all testing related so that is fine.

Yes, I agree that the switch to `Arc<[u8]>` is really important now.

linera-execution/src/execution_state_actor.rs
MathieuDutSik: Actually, the processing of `ExecutionRequest` does a little bit more than just returning the value: It also returns the boolean `is_new`.

And there are scenarios where we have `is_new=true` / `false` with the blob being created in the same transaction:
* The `create_data_blob` is not changing the `used_blob`.
* The `read_data_blob` would have `is_new=true` the first time and `is_new=false` the second time.

afck: Ah, great! I believe we can return `false` for any blob that was created on this chain, i.e. in particular if we find it in `created_blobs`.

MathieuDutSik: Ok, I changed it according to that.
Note that we no longer have a cost associated to `read_blob_content`. In other words, if you read a blob within the same block that has created it, then there is no cost associated it.

Cargo.toml
afck: OK, maybe it's cleaner.

> reading is done by an operation. The standard way is to use a service.

Not sure what you mean by this.

What I was suggesting was one block with two operations:
1. `SystemOperation::PublishDataBlob`.
2. A `non-fungible` operation to mint an NFT using that data blob.

examples/publish-read-data-blob/README.md
```diff
@@ -0,0 +1,11 @@
+# Publish and read a data blob
+
+This example shows how to create a blob and how to read it.
+This example should be read in conjuction with the end to end
```
afck:
```suggestion
This example should be read in conjunction with the end-to-end
```

examples/publish-read-data-blob/src/lib.rs
```diff
@@ -0,0 +1,38 @@
...
+
+#[derive(Debug, Deserialize, Serialize)]
+pub enum Operation {
+    CreateDataBlob(Vec<u8>),
+    ReadDataBlob(DataBlobHash, Vec<u8>),
+    CreateAndReadDataBlob(Vec<u8>),
+}
+
+#[derive(Debug, Deserialize, Serialize)]
+pub enum ServiceOperation {
```
afck: Maybe `ServiceQuery`?

linera-execution/src/transaction_tracker.rs
```diff
@@ -45,6 +45,8 @@ pub struct TransactionTracker {
     /// - [`ApplicationDescription`]
     /// - [`ChainDescription`]
     blobs: BTreeMap<BlobId, Blob>,
```
afck: (This could also be `BlobContent` instead of `Blob`, I guess.)

linera-service/tests/linera_net_tests.rs
```diff
@@ -4734,3 +4734,80 @@ async fn test_end_to_end_repeated_transfers(config: impl LineraNetConfig) -> Res
...
+    // Method 1: Publishing and reading in different blocks.
+
+    let test_data = b"This is test data for method 1.".to_vec();
+
+    // publishing the data.
+    let mutation = ServiceOperation::PublishDataBlob(test_data.clone());
+    application.run_json_query(&mutation).await?;
+
+    // getting the blob_id.
+    let query = ServiceOperation::GetHash;
```
afck: We could also remove that variant and just compute the blob hash directly.

linera-execution/src/execution_state_actor.rs
```diff
@@ -90,14 +90,14 @@ where
         #[cfg(with_metrics)]
         let _latency = metrics::LOAD_SERVICE_LATENCY.measure_latency();
         let blob_id = id.description_blob_id();
-        let description = match txn_tracker.created_blobs().get(&blob_id) {
+        let description = match txn_tracker.get_blob_content(&blob_id) {
             Some(blob) => bcs::from_bytes(blob.bytes())?,
             None => self.system.describe_application(id, txn_tracker).await?,
```
afck, line 95: So here we didn't actually need to clone the blob at all. Maybe `get_blob_content` should return a reference then?

afck (approved):

Cargo.toml
MathieuDutSik: For the service issue:
* The variant case `Operation::ReadDataBlob` could be covered in a service call.
* But it is there so that we can have a `schedule_operation` followed by another `schedule_operation`.

Now, about "one block with two operations", I do not know how to do it from the end-to-end test except in the way I found, which was to call a service that calls two `schedule_operation` in sequence.
If I call from the end-to-end test a `node_service.publish_data_blob`, then it gets executed immediately, and I get one block. And then if I call the NFT mint, I have another block. So, defeat of purpose.

Maybe you meant changing the `non-fungible` contract. But that is unwelcome since it is supposed to be an illustration of the functionality. Not a test stuff.

This leads me to two ideas:

(A) It is true that we are starting to see some entries in "examples" that should be better called "examples_test". Those could be separated, and the `linera_net_tests.rs` could be further split so that they are run only with `storage-service`, which is the fastest locally.

(B) In the node_service, we can submit one mutation. Could it be possible to submit several mutations at once? That could be welcomed by users so that they could explicitly create a block with several transactions.

afck: Right, good point.

I agree with A.
And B, too, but maybe we will solve that in a different way in the long run (e.g. automatically batch the user's requests into blocks).

examples/publish-read-data-blob/README.md
MathieuDutSik: Ok.

examples/publish-read-data-blob/src/lib.rs
MathieuDutSik: Yes.

linera-execution/src/transaction_tracker.rs
MathieuDutSik: Correct

linera-service/tests/linera_net_tests.rs
MathieuDutSik: Yes, significant simplification.

linera-execution/src/execution_state_actor.rs
MathieuDutSik: Right, we could avoid a few clone operations.

PR #4361, opened 2025-08-19
Synchronize chain state after notification stream is established.

afck:
## Motivation

`test_wasm_end_to_end_social_event_streams::remote_net_grpc` is failing on `devnet-2025-08-15`. The problem seems to be that the relevant block on the publisher chain arrives _after_ we synchronize that chain explicitly but _before_ the notification streams from the validators are established, so we miss that block.

It's difficult to reproduce this issue in another way; locally, the tests all pass.

## Proposal

Synchronize after establishing the notification stream, for each validator.

## Test Plan

For me, this fixes the test.

## Release Plan

## Links

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -1854,10 +1854,16 @@ impl<Env: Environment> ChainClient<Env> {
 
     /// Obtains the committee for the current epoch of the local chain.
     #[instrument(level = "trace")]
-    pub async fn local_committee(&self) -> Result<Committee, LocalNodeError> {
-        self.chain_info_with_committees()
-            .await?
-            .into_current_committee()
+    pub async fn local_committee(&self) -> Result<Committee, ChainClientError> {
+        let info = match self.chain_info_with_committees().await {
+            Ok(info) => info,
+            Err(LocalNodeError::BlobsNotFound(_)) => {
+                self.synchronize_chain_state(self.chain_id).await?;
+                self.chain_info_with_committees().await?
+            }
+            Err(err) => return Err(err.into()),
+        };
+        Ok(info.into_current_committee()?)
     }
 
     /// Obtains the committee for the latest epoch on the admin chain.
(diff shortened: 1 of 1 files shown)


afck:
Good point. This might cause one additional chain info request per validator. I guess we could call `synchronize_chain_state` at the beginning of the function only if `local_committee` fails. I'll try that.

afck:
Done in https://github.com/linera-io/linera-protocol/pull/4361/commits/23d2694005659285c88f4b8adde7a7764cbd9476.

linera-core/src/client/mod.rs
```diff
@@ -3766,14 +3758,21 @@ impl<Env: Environment> ChainClient<Env> {
...
-        let (chain_id, nodes, local_node) = {
-            let committee = self.local_committee().await?;
+        let (nodes, local_node) = {
+            let committee = match self.local_committee().await {
+                Ok(committee) => committee,
+                Err(LocalNodeError::BlobsNotFound(_)) => {
+                    self.synchronize_chain_state(self.chain_id).await?;
+                    self.local_committee().await?
+                }
+                Err(err) => return Err(err.into()),
```
deuszx: Would it make sense to have this fallback in `self.local_committee()`? Or is this the only case

afck: Good idea! Done in https://github.com/linera-io/linera-protocol/pull/4361/commits/9ad5f2d8ffb5e7603380df6077fff622856e678f.

ma2bd (approved):

linera-core/src/client/mod.rs
```diff
@@ -3788,9 +3786,19 @@ impl<Env: Environment> ChainClient<Env> {
...
                 continue;
             };
+            let this = self.clone();
             let stream = stream::once({
                 let node = node.clone();
-                async move { node.subscribe(vec![chain_id]).await }
+                async move {
+                    let stream = node.subscribe(vec![this.chain_id]).await?;
+                    // Only now the notification stream is established. We may have missed
+                    // notifications since the last time we synchronized.
```
deuszx, line 3795: I'd have thought that it should be OK to miss past notifications – i.e. from the time before we subscribed. That's usually how event streams work.

afck: When you start the node service, the expectation is that it first synchronizes all the relevant chains and then follows along using the validator notifications, so the chains are fully synchronized at all times.

If we missed a notification from the validators about block 5, it could be that our local chain remains at height 4. At least until we get a notification about block 6, which may be late or never.

Or take the example of the reconfiguration test:
* If I start the node service _after_ an epoch change, the client will automatically migrate my chain to the new epoch (because it synchronizes).
* If I start it _before_ an epoch change, the client will (this time due to notifications from the validators) migrate my chain to the new epoch.
* If I start it at the same time of an epoch change, we should also expect this to work.

deuszx: Right, and doing it another way (first syncing and then starting the notification service) may lead to missing some notifications again.

Thanks.

PR #4366, opened 2025-08-19
Address a bug with message during instantiation.

### LINKED ISSUE #3539: Should we ignore `block.messages()[0].is_empty()` assertion in test `ActiveChain` if we'll create message in instantiation ?
**Description**
If we create application which will create new message when instantiating in integration tests, it'll be asserted by `assert!(block.message()[0].is_empty)`.

**Suggestion**
Should we ignore `block.messages()[0].is_empty()` assertion in test `ActiveChain` if we'll create message in instantiation ?

**Screenshot**
Image

MathieuDutSik:
## Motivation

The scenario of a smart contract emitting a message during instantiation breaks the `TestValidator` framework.
We address the problem here.

Fixes #3539 

## Proposal

The correction pointed out in the issue is exactly what is needed.
A smart contract has been created that demonstrates that the correction works and also test
in the end-to-end test, which is nice to check.

## Test Plan

CI.

## Release Plan

## Links

None.

### DIFF
--- a/linera-sdk/src/test/chain.rs
+++ b/linera-sdk/src/test/chain.rs
@@ -560,7 +560,6 @@ impl ActiveChain {
 
         let block = creation_certificate.inner().block();
         assert_eq!(block.messages().len(), 1);
-        assert!(block.messages()[0].is_empty());
 
         let description = ApplicationDescription {
             module_id: module_id.forget_abi(),
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5750,6 +5750,7 @@ dependencies = [
  "tower-http 0.6.6",
  "tracing",
  "tracing-subscriber 0.3.19",
+ "track-instantiation",
 ]
 
 [[package]]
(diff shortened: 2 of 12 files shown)


linera-sdk/src/test/chain.rs
```diff
@@ -556,7 +556,6 @@ impl ActiveChain {
 
         let block = creation_certificate.inner().block();
         assert_eq!(block.messages().len(), 1);
-        assert!(block.messages()[0].is_empty());
```
ma2bd, line 563: Is this the fix?

ma2bd: Do we need the other assert?

MathieuDutSik: Yes it is the fix.

Yes, we need the other assert. When we instantiate, we have only one block of messages being emitted.

afck (commented):
Good catch! I agree with the fix. But do we really need a slow and expensive end-to-end regression test for that one removed line?

MathieuDutSik:
> Good catch! I agree with the fix. But do we really need a slow and expensive end-to-end regression test for that one removed line?

No, we do not need it.
It was useful for checking that things work correctly, but I would not defend it as I would in other cases (`create-and-call` and `publish-read-data-blob`).

MathieuDutSik:
> Good catch! I agree with the fix. But do we really need a slow and expensive end-to-end regression test for that one removed line?

So, I removed the end-to-end test. But I kept the integration test.
I tried to use the `meta-counter` as an example, but without success, as it is a more complex test case.

ma2bd (approved):

PR #4367, opened 2025-08-19
Move to the use of DataBlobHash.

MathieuDutSik:
## Motivation

The `BaseRuntime` and `ContractRuntime` are a little bit disorganised with their input and output:
* `read_data_blob` / `assert_data_blob_exists` are taking a `CryptoHash` as input.
* `create_data_blob` is returning a `BlobId`.

## Proposal

The natural type is the `DataBlobHash` that is the natural type for this.

Previous to this PR, it is in the `linera-sdk`. We are forced to move it to `linera-base`.

## Test Plan

The CI.

## Release Plan

## Links

None.

### DIFF
--- a/linera-execution/src/runtime.rs
+++ b/linera-execution/src/runtime.rs
@@ -16,8 +16,7 @@ use linera_base::{
     },
     ensure, http,
     identifiers::{
-        Account, AccountOwner, BlobId, BlobType, ChainId, EventId, GenericApplicationId, StreamId,
-        StreamName,
+        Account, AccountOwner, ChainId, EventId, GenericApplicationId, StreamId, StreamName,
     },
     ownership::ChainOwnership,
     time::Instant,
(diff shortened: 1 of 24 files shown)


linera-execution/src/runtime.rs
```diff
@@ -938,9 +938,9 @@ where
         Ok(())
     }
 
-    fn read_data_blob(&mut self, hash: &CryptoHash) -> Result<Vec<u8>, ExecutionError> {
+    fn read_data_blob(&mut self, hash: DataBlobHash) -> Result<Vec<u8>, ExecutionError> {
         let mut this = self.inner();
-        let blob_id = BlobId::new(*hash, BlobType::Data);
+        let blob_id = BlobId::new(hash.0, BlobType::Data);
```
afck: We could also `impl From<DataBlobHash> for BlobId`.

MathieuDutSik: Good idea.

afck (approved):

PR #4371, opened 2025-08-20
Eliminate `authenticated_caller_id` from the `OperationContext`

MathieuDutSik:
## Motivation

It is not used and not set to any other value than `None`.

## Proposal

Straight removal.
I am not 100% sure, but I think that it is not an error that it is not used.
The `authenticated_caller_id` is instead available via the call_stack.

## Test Plan

The CI.

## Release Plan

## Links

None.


afck (approved):
Good catch!

PR #4373, opened 2025-08-20
Improve syncing on incoming message

deuszx:
## Motivation

When inspecting the performance of our web demos (namely hosted-fungible) we noticed that an incoming transfer triggers 7 network requests while only 3 (last ones) are about accepting an incoming transfer (proposing a block that includes the incoming message).

After closer research, it was discovered that we make:
- `HandleChainInfoQuery` - first request to learn about the sender chain tip
- `HandleChainInfoQuery` - second time we request via `fetch_sent_certificate_hashes` with a specific `BlockRange` (after checking the `info.requested_received_log`)
- `DownloadCertificates` - to download the certififcates (by hash)
- `DownloadCertificates` - unnecessary query with empty payload (this was a bug).
- and finally three requests to propose the block.

## Proposal

This PR proposes an improvement of the situation – decreasing # of requests from 7 to 5:
- `HandleChainInfoQuery` – to learn about the missing certificates for block heights that we might be missing
- `DownloadCertificatesByHeights` – new endpoint that we use to download certificates at specific block heights
- and finally three requests to propose a block.

I don't think we can get it any lower with the way `ChainClient` is currently structured.

This also improves all calls to `query_certificates_from` which are now making 1 network query instead of two.

## Test Plan

CI

## Release Plan

- Nothing to do / These changes follow the usual release cycle. **OR**
    - be released in a new SDK,

## Links


linera-core/src/client/mod.rs
```diff
@@ -804,19 +802,13 @@ impl<Env: Environment> Client<Env> {
...
-                let height1 = *remote_heights.last()?;
-
                 // Find the hashes of the blocks we need.
-                let range = BlockHeightRange::multi(height0, height1.0 + 1 - height0.0);
+                let limit = remote_heights.last()?.0 + 1 - height0.0;
                 Some(async move {
-                    let hashes = remote_node
-                        .fetch_sent_certificate_hashes(sender_chain_id, range)
+                    let certificates = remote_node
+                        .download_certificates_range(sender_chain_id, height0, Some(limit))
```
afck: This will always download the whole range, even if we don't need it, so we'd lose a lot of the "sparse client" advantages again.

I think the endpoint really has to take a list of heights, and then return only the certificate for those heights, rather than a range.

afck: (The list here would be all the `remote_heights` that are greater than the local height, I think.)

afck: But even then, withe removal of the code below, and since we're already downloading the certificates here, this will mean that we download all the relevant blocks of the sender chain again and again, each time we receive a new message.

I think we actually need to look in our local sender chain outbox to find out which messages to the receiver chain we've already handled.

linera-core/src/data_types.rs
```diff
@@ -58,6 +58,20 @@ impl BlockHeightRange {
     }
 }
 
+impl Display for BlockHeightRange {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        if let Some(limit) = self.limit {
+            if limit == 0 && limit == 1 {
+                write!(f, "{}", self.start)
```
afck: For `0` I'd write "empty" or something.

linera-core/src/client/mod.rs
deuszx: We compare `remote_heights` with `local_next` so we shouldn't re-request already known blocks.

afck: But `local_next` points to the next block to _execute_. We should really be looking into our outboxes here.

deuszx: I don't see where I am changing the approach w.r.t. to sparcity:

Previously we did:
1. `HandleChainInfoQuery.with_range(range)` -> returns certificate hashes for blocks in `range`.
2. Filter out certs we already have locally
3. Request the rest

What we do now is we request all those certs in the `range` right away.

This doesn't change the sparcity story - at the end we should have the same blocks locally as before

afck: But we may have certificates locally that are higher than `local_next`. Checking in storage whether we have those does catch them and avoid requesting them again. We really do need to use the local outbox height here and not `local_next`.

linera-core/src/data_types.rs
deuszx: `0` should be disallowed but I didn't want to change the API of `BlockHeightRange`.

linera-core/src/client/mod.rs
afck: (Sorry, I totally missed that this was already coming from our own outbox. :facepalm: )

linera-core/src/client/mod.rs
```diff
@@ -817,6 +814,11 @@ impl<Env: Environment> Client<Env> {
         .flatten()
         .collect::<Vec<_>>();
 
+        info!(
+            "Received {} certificates from validator {:?} for chain {chain_id:.8}",
+            certificates.len(),
+            remote_node.public_key
+        );
```
deuszx: temporary

linera-rpc/src/grpc/client.rs
```diff
@@ -406,6 +406,9 @@ impl ValidatorNode for GrpcClient {
         let mut missing_hashes = hashes;
         let mut certs_collected = Vec::with_capacity(missing_hashes.len());
         loop {
+            if missing_hashes.is_empty() {
+                break;
+            }
```
deuszx: This fixes the issue with unnecessary call to `DownloadCertificates`.

linera-core/src/chain_worker/state/guard.rs
```diff
@@ -913,6 +913,18 @@ where
 
             info.requested_pending_message_bundles = messages;
         }
+        if !query.request_sent_certificate_hashes_by_heights.is_empty() {
+            let mut hashes = Vec::new();
+            for height in query.request_sent_certificate_hashes_by_heights {
+                let hash = chain
+                    .confirmed_log
```
afck: If `height > chain.tip_state.get().next_block_height`, we need to look it up in `preprocessed_blocks` rather than `confirmed_log`. Alternatively: `chain.block_hashes(height..=height)`.

linera-core/src/data_types.rs
```diff
@@ -130,6 +134,13 @@ impl ChainInfoQuery {
 
     pub fn with_sent_certificate_hashes_in_range(mut self, range: BlockHeightRange) -> Self {
```
afck: Do we still need this one then?

linera-core/src/remote_node.rs
```diff
@@ -169,14 +169,10 @@ impl<N: ValidatorNode> RemoteNode<N> {
...
     ) -> Result<Vec<ConfirmedBlockCertificate>, NodeError> {
         tracing::debug!(name = ?self.public_key, ?chain_id, ?start, ?limit, "Querying certificates");
-        let range = BlockHeightRange {
-            start,
-            limit: Some(limit),
-        };
-        let query = ChainInfoQuery::new(chain_id).with_sent_certificate_hashes_in_range(range);
-        let info = self.handle_chain_info_query(query).await?;
-        self.node
-            .download_certificates(info.requested_sent_certificate_hashes)
```
afck, line 179: I think this downloaded at most 1000 at a time (see `download_certificates::download_certificates`). We should do something similar when downloading by heights.

linera-core/src/remote_node.rs
```diff
@@ -246,54 +242,16 @@ impl<N: ValidatorNode> RemoteNode<N> {
         }
     }
 
-    /// Returns the list of certificate hashes on the given chain in the given range of heights.
-    /// Returns an error if the number of hashes does not match the size of the range.
```
afck, line 250: We could still check in this function that we got the certificates with the correct heights and chain ID.

linera-core/src/unit_tests/test_utils.rs
```diff
@@ -596,6 +607,52 @@ where
...
+            Err(NodeError::ClientIoError {
+                error: "Failed to receive chain info response".to_string(),
+            })
+        })?;
+
+        let hashes = match chain_info_response {
+            Ok(response) => response.info.requested_sent_certificate_hashes,
+            Err(e) => {
+                sender.send(Err(e)).ok();
+                return Ok(());
```
afck:
```suggestion
                return sender.send(Err(e));
```

linera-rpc/src/client.rs
```diff
@@ -245,6 +245,28 @@ impl ValidatorNode for Client {
         })
     }
 
+    async fn download_certificates_by_heights(
+        &self,
+        chain_id: ChainId,
+        mut heights: Vec<BlockHeight>,
+    ) -> Result<Vec<ConfirmedBlockCertificate>, NodeError> {
+        heights.sort();
```
afck, line 253: We could sort already in `remote_node`. That would make it easier to immediately check that we got the certificates we requested.

linera-core/src/chain_worker/state/guard.rs
```diff
@@ -913,6 +913,18 @@ where
 
             info.requested_pending_message_bundles = messages;
         }
+        if !query.request_sent_certificate_hashes_by_heights.is_empty() {
```
Twey: I don't think the `if` adds anything here (it's just duplicating the first check of the `for`).

(Empty vectors don't allocate.)

linera-rpc/src/grpc/client.rs
```diff
@@ -434,6 +437,26 @@ impl ValidatorNode for GrpcClient {
         Ok(certs_collected)
     }
 
+    #[instrument(target = "grpc_client", skip(self), err(level = Level::WARN), fields(address = self.address))]
+    async fn download_certificates_by_heights(
+        &self,
+        chain_id: ChainId,
+        heights: Vec<BlockHeight>,
+    ) -> Result<Vec<ConfirmedBlockCertificate>, NodeError> {
```
afck, line 442: Do we need the same `is_empty` check here, too?

linera-rpc/src/grpc/client.rs
```diff
@@ -434,6 +437,26 @@ impl ValidatorNode for GrpcClient {
...
+        let request = CertificatesByHeightRequest { chain_id, heights };
+        let received: Vec<ConfirmedBlockCertificate> = Vec::<Certificate>::try_from(
+            client_delegate!(self, download_certificates_by_heights, request)?,
+        )?
+        .into_iter()
+        .map(|cert| {
+            ConfirmedBlockCertificate::try_from(cert)
```
afck: Can't we use the `ConfirmedBlockCertificate` type directly in the protocol, so that this would be a serialization error?

linera-core/src/data_types.rs
```diff
@@ -310,6 +321,13 @@ impl ChainInfoResponse {
 
 impl BcsSignable<'_> for ChainInfo {}
 
+/// Request for downloading certificates by height range.
+#[derive(Debug, Clone)]
+pub struct CertificatesByHeightRequest {
+    pub chain_id: ChainId,
+    pub heights: Vec<BlockHeight>,
+}
+
```
Twey: I wonder if it would be better to have a single ‘get certificates’ endpoint with a sophisticated request type with various optional filters for the certificates.

linera-core/src/remote_node.rs
```diff
@@ -246,54 +242,16 @@ impl<N: ValidatorNode> RemoteNode<N> {
...
-        ensure!(
-            returned == hashes,
-            NodeError::UnexpectedCertificates {
-                returned,
-                requested: hashes
-            }
-        );
-        Ok(certificates)
+        self.node
+            .download_certificates_by_heights(chain_id, heights)
```
Twey, line 248: Is there a reason we can't put the functionality right here instead of adding the layer of indirection?

linera-rpc/src/grpc/client.rs
Twey:
```suggestion
         while !missing_hashes.is_empty() {
```

linera-rpc/tests/snapshots/format__format.yaml.snap
```diff
@@ -733,45 +736,53 @@ NodeError:
           SEQ:
             TYPENAME: CryptoHash
     13:
+      MissingCertificatesByHeights:
+        STRUCT:
+          - chain_id:
+              TYPENAME: ChainId
+          - heights:
+              SEQ:
+                TYPENAME: BlockHeight
```
Twey: The reason Protobuf has field IDs and always-optional fields is so that you can put new fields at the end and avoid breaking compatibility.

linera-core/src/chain_worker/state/guard.rs
deuszx: Ah, true. I just wanted to make it explicit.

linera-rpc/tests/snapshots/format__format.yaml.snap
deuszx: Right, I know. I don't think we worry about breaking compatibility right now. It doesn't mean i shouldn't follow best practices.

linera-core/src/remote_node.rs
deuszx: What indirection? That we do that in the proxy?

afck: This code is in the client. (It's the client's handle of a remote node connection.) So that would create a lot of client-proxy round-trips in addition to the proxy-server round-trips which should be much faster.

linera-rpc/tests/snapshots/format__format.yaml.snap
afck: But maybe for now, on `main`, having fields in an order that conceptually makes sense is more important that backwards compatibility?

linera-core/src/remote_node.rs
deuszx: Yes. More importantly – if that wasn't clear – is that we make ONE query to a validator vs multiple queries. Now the "multiple queries" happens between validator's proxy and worker.

linera-core/src/data_types.rs
deuszx: Probably not.

linera-rpc/src/client.rs
deuszx: We kindof already do b/c we construct it from the starting height and the limit. This is just a failover for the cases where it's used from a different flow.

linera-rpc/src/grpc/client.rs
deuszx: I don't understand – we use it in the protocol

afck: I mean shouldn't this part of some `conversions.rs` rather than here? (Not important; not a blocker!)

deuszx: The issue here is that the response is a `Certificate` enum – either validatedo, confirmed or timeout.

linera-core/src/data_types.rs
deuszx: Done in [Remove unused ChainInfoQuery field](https://github.com/linera-io/linera-protocol/pull/4373/commits/1b535eadfbd13f2b20e50ea9ec9081afa0e44615)

linera-rpc/src/grpc/client.rs
deuszx: Done in [Add loop and early exit to download_certs_at_heights](https://github.com/linera-io/linera-protocol/pull/4373/commits/19e2145238a9094d1819a9fb3777d5ba00106573)

linera-rpc/src/grpc/client.rs
deuszx: Done in [Add loop and early exit to download_certs_at_heights](https://github.com/linera-io/linera-protocol/pull/4373/commits/19e2145238a9094d1819a9fb3777d5ba00106573)

linera-core/src/remote_node.rs
deuszx: Done in https://github.com/linera-io/linera-protocol/pull/4373/commits/be9e058f513102bad986662d1d7d0adcec24b559

linera-core/src/chain_worker/state/guard.rs
deuszx: Done

linera-rpc/src/grpc/client.rs
afck: Right, so that even wastes a byte in the protocol.

linera-core/src/chain_worker/state/guard.rs
deuszx: I see how `chain.block_hashes(height..=height)` would also work but it seems like an unnecessary complexity? I'd rather first try `confirmed_log` and - if `None` - try `preprocessed_blocks`.

linera-core/src/remote_node.rs
deuszx: Done in [Add loop and early exit to download_certs_at_heights](https://github.com/linera-io/linera-protocol/pull/4373/commits/19e2145238a9094d1819a9fb3777d5ba00106573)

linera-core/src/chain_worker/state/guard.rs
afck: That's one unnecessary storage lookup, though. If we compare with `next_block_height` or use `block_hashes`, we always look it up in the right place.

linera-rpc/src/grpc/client.rs
deuszx: We'd have to change the APIs to special case the `ConfirmedBlockCertificate` type.

linera-core/src/chain_worker/state/guard.rs
deuszx: Done https://github.com/linera-io/linera-protocol/pull/4373/commits/e0d7044142d5788a4aae718a9945f78708b4ba9e

linera-rpc/src/grpc/client.rs
afck: True; not in this PR at least then. :sweat_smile:

linera-core/src/remote_node.rs
```diff
@@ -246,51 +245,34 @@ impl<N: ValidatorNode> RemoteNode<N> {
...
         ensure!(
-            returned == hashes,
-            NodeError::UnexpectedCertificates {
-                returned,
-                requested: hashes
+            expected_heights.is_empty(),
+            NodeError::MissingCertificatesByHeights {
+                chain_id,
+                heights: expected_heights.into_iter().collect(),
             }
```
afck, line 278: Isn't it important for the caller in our case that the certificates are in the correct order, by increasing height? We are not checking that.

linera-service/src/proxy/grpc.rs
```diff
@@ -625,6 +629,52 @@ where
         )?))
     }
 
+    #[instrument(skip_all, err(Display))]
+    async fn download_certificates_by_heights(
```
afck, line 633: I'm confused: Isn't this a new private function that isn't used in this module?

linera-rpc/tests/snapshots/format__format.yaml.snap
Twey: I'd agree if we weren't planning to backport this to the testnet.

linera-service/src/proxy/grpc.rs
Twey: This is a new trait method.

linera-rpc/tests/snapshots/format__format.yaml.snap
deuszx: Ahh... you're right.

linera-core/src/remote_node.rs
deuszx: I don't know, but I can add that check.

linera-rpc/tests/snapshots/format__format.yaml.snap
deuszx: Fixed in [Move errors to the end of the enum](https://github.com/linera-io/linera-protocol/pull/4373/commits/e6c4e56c740629cdde6fccd2f67287eaebdebae5).

Also - [Fix backwards incompatibility in protobuf def](https://github.com/linera-io/linera-protocol/pull/4373/commits/e84faccb360a92d5be1a27f78baed6e86478b868)

linera-core/src/remote_node.rs
deuszx: Done in [Ensure that certificates are returned in the expected order](https://github.com/linera-io/linera-protocol/pull/4373/commits/67f2ea22c0b19a0ee4237508b5777c69ba1e45da)

linera-core/src/remote_node.rs
```diff
@@ -169,14 +172,10 @@ impl<N: ValidatorNode> RemoteNode<N> {
...
         tracing::debug!(name = ?self.public_key, ?chain_id, ?start, ?limit, "Querying certificates");
-        let range = BlockHeightRange {
-            start,
-            limit: Some(limit),
-        };
-        let query = ChainInfoQuery::new(chain_id).with_sent_certificate_hashes_in_range(range);
-        let info = self.handle_chain_info_query(query).await?;
-        self.node
-            .download_certificates(info.requested_sent_certificate_hashes)
+        let heights = (start.0..start.0 + limit)
```
deuszx, line 175: This is wrong, I think, by 1 too many 🤔

linera-core/src/remote_node.rs
```diff
@@ -246,51 +245,46 @@ impl<N: ValidatorNode> RemoteNode<N> {
...
-            warn!(
-                ?range,
-                received_num = hashes.len(),
-                "Validator sent invalid number of certificate hashes."
-            );
-            return Err(NodeError::InvalidChainInfoResponse);
+        if certificates.len() > expected_heights.len() {
+            return Err(NodeError::TooManyCertificatesReturned {
+                chain_id,
+                remote_node: Box::new(self.public_key),
```
deuszx, line 254: clippy started complaining about enum too big (if w/o the `Box`)

linera-core/src/remote_node.rs
afck: I don't think so? Because you're using `..`, not `..=`.
`1..(1 + 2)` contains two numbers.

linera-rpc/tests/snapshots/format__format.yaml.snap
```diff
@@ -350,9 +344,9 @@ ChainInfoQuery:
         TYPENAME: AccountOwner
     - request_committees: BOOL
     - request_pending_message_bundles: BOOL
-    - request_sent_certificate_hashes_in_range:
-        OPTION:
-          TYPENAME: BlockHeightRange
+    - request_sent_certificate_hashes_by_heights:
+        SEQ:
+          TYPENAME: BlockHeight
```
deuszx: I think this change is not backwards compatible and there's no way to handle that like protobuf does.

Thoughts @Twey ?

linera-core/src/remote_node.rs
deuszx: Ah, right.. We convert to `Range` type first that is exclusive on the right-hand side.

linera-rpc/tests/snapshots/format__format.yaml.snap
ma2bd: Indeed, hotfixing won't work

linera-rpc/tests/snapshots/format__format.yaml.snap
```diff
@@ -242,12 +242,6 @@ BlockHeader:
           TYPENAME: AccountOwner
 BlockHeight:
   NEWTYPESTRUCT: U64
-BlockHeightRange:
```
deuszx: I should revert these removals for backwards compatibility purposes.

linera-rpc/proto/rpc.proto
```diff
@@ -185,9 +196,6 @@ message ChainInfoQuery {
   // Query the received messages that are waiting to be picked in the next block.
   bool request_pending_message_bundles = 4;
 
-  // Query a range of certificates hashes sent from the chain.
-  optional bytes request_sent_certificate_hashes_in_range = 5;
```
ma2bd, line 189: @deuszx Can we have three separate PRs? (or at least two)
* Implementing the new endpoints on the validator side
* Using the new endpoints in the client (could be merged with the first)
* Removing the old endpoints (must be a separate because it won't be release to testnet #3)

deuszx: I don't like separation into "add thing" and "use the thing" b/c reviewing the first one is more difficult w/o the context in which it's going to be used.

Regarding removal of the new endpoints – I am not doing that, am I? I only retired the unused field in the protobuf but in a backwards-compatible way.

ma2bd: > I don't like separation into "add thing" and "use the thing"
ok

> "unused field"
Not if people are running old clients. But ok if we are able to create the next testnet SDK (0.15.0 I think) directly with only new clients it's fine.

linera-core/src/data_types.rs
```diff
@@ -76,6 +76,8 @@ pub struct ChainInfoQuery {
     #[debug(skip_if = Not::not)]
     pub request_pending_message_bundles: bool,
     /// Query a range of certificate hashes sent from the chain.
+    //  dev: this field is left and unused to maintain backwards compatibility
```
afck, line 79: (Should we create an issue to remove this, so we don't forget?)

linera-core/src/data_types.rs
```diff
@@ -76,6 +76,8 @@ pub struct ChainInfoQuery {
...
+    //  after hotfixing testnet conway.
```
afck:
```suggestion
    //  after hotfixing Testnet Conway.
```

(Or "testnet Conway". But at least "Conway" is a proper noun.)

linera-core/src/node.rs
```diff
@@ -145,6 +145,13 @@ pub trait ValidatorNode {
         hashes: Vec<CryptoHash>,
     ) -> Result<Vec<ConfirmedBlockCertificate>, NodeError>;
 
+    /// Requests a batch of certificates from a specific chain by height range.
```
afck:
```suggestion
    /// Requests a batch of certificates from a specific chain by heights.
```

linera-core/src/remote_node.rs
```diff
@@ -246,51 +235,46 @@ impl<N: ValidatorNode> RemoteNode<N> {
         }
     }
 
-    /// Returns the list of certificate hashes on the given chain in the given range of heights.
-    /// Returns an error if the number of hashes does not match the size of the range.
+    /// Downloads a range of certificates from the given chain.
```
afck:
```suggestion
    /// Downloads a list of certificates from the given chain.
```

linera-core/src/unit_tests/wasm_client_tests.rs
```diff
@@ -914,7 +914,7 @@ where
     let owner_a = keys.generate_new().into();
     let owner_b = keys.generate_new().into();
     let clock = storage_builder.clock().clone();
-    let mut builder = TestBuilder::new(storage_builder, 2, 0, keys).await?;
+    let mut builder = TestBuilder::new(storage_builder, 4, 0, keys).await?;
```
afck, line 917: Thanks! Let's see if that fixes the flakiness.

My thought yesterday was that 2 is a bad number because then 1 is the validity threshold, so if a test has two `communicate_with_quorum` instances that are expected to fail, then in some cases it might be the same validator in both instances and in others it might be two different ones.

But now I'm not so sure anymore, because in this particular test there is only _one_ instance where `communicate_with_quorum` is expected to fail.

linera-core/src/data_types.rs
deuszx: Yes. Will do that.

deuszx: https://github.com/linera-io/linera-protocol/issues/4402

PR #4374, opened 2025-08-20
Correct the `authenticated_signer` / `authenticated_caller_id` in contract calls.

MathieuDutSik:
## Motivation

The `authenticated_signer` / `authenticated_caller_id` are set according to the following way:
* In a contract call from another contract the caller_id is set. That works.
* When calling another contract authenticated, both are passed. That was not correct.
* When calling another contract unauthenticated, both are set to `None`. That was not correct.
* The system has access to the `authenticated_signer` / `authenticated_caller_id` even if unauthenticated.

## Proposal

Two errors were found:

The first error is about the `fn authenticated_signer` which was set to the one of the block. This
is not the same as the one accessible from the contract. So, the change was done so that the
values are set from the call stack.

The second error was that the `authenticated_signer` / `authenticated_caller_id` were cached.

After those two corrections, it passes.

Next Work: Separate those end-to-end tests from the other classical tests.

## Test Plan

The CI.

The end-to-end test is `test_end_to_end_contract_transfer`. It is a contrived example that creates another
smart contract with the same module id. This allows having a non-trivial `authenticated_caller_id`.

Then the application allows checking that the values are the ones expected.

## Release Plan

## Links

None.


linera-sdk/src/contract/runtime.rs
```diff
@@ -33,11 +33,9 @@ where
     application_id: Option<ApplicationId<Application::Abi>>,
```
ma2bd, line 35: How about `test_runtime.rs`?

linera-execution/src/runtime.rs
```diff
@@ -1169,7 +1169,8 @@ impl ContractSyncRuntimeHandle {
 
 impl ContractRuntime for ContractSyncRuntimeHandle {
     fn authenticated_signer(&mut self) -> Result<Option<AccountOwner>, ExecutionError> {
-        Ok(self.inner().authenticated_signer)
+        let this = self.inner();
+        Ok(this.current_application().signer)
```
ma2bd: I moved this fix to #4392 because we need to deploy it immediately and I wanted a unit test in `linera-execution` only.

linera-sdk/src/contract/runtime.rs
ma2bd: Ok never mind. There is no caching there

linera-execution/src/runtime.rs
MathieuDutSik: So, the caching bug is not an issue?
`linera-core` and `linera-execution` do not depend on `linera-sdk` so the bug could not appear in those parts of the code.
But it could be made to appear in an integration test.

linera-sdk/src/contract/runtime.rs
MathieuDutSik: There is caching in the integration tests.
I could reproduce the bug there.

linera-execution/src/runtime.rs
ma2bd: It is definitely an issue but with a different urgency level because it's SDK-related (not a protocol change).
Therefore, grouping the two fixes in the same PR was not working

linera-sdk/src/contract/runtime.rs
ma2bd: Yes, this makes sense. The integration tests are using the real Wasm VM. `test_runtime.rs` is for unit tests

examples/contract-call/tests/integration.rs
```diff
@@ -0,0 +1,82 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+//! Integration tests for the Contract Call application
+
+#![cfg(not(target_arch = "wasm32"))]
+
+use contract_call::{ContractTransferAbi, Parameters};
+use linera_sdk::{
+    linera_base_types::{Account, AccountOwner, Amount},
+    test::TestValidator,
```
ma2bd, line 11: Thanks for the integration test. Now that testing the fix doesn't require `linera-service`, would it be possible to move the "fixture" app and this test to `linera-sdk/tests`?

ma2bd: If that's too much work, this could be in another PR.

ma2bd (approved):

examples/contract-call/tests/integration.rs
MathieuDutSik: I looked at this problem of moving the tests to `linera-sdk/tests`, but it does not seem easy.
We would have to change `linera-sdk` to be a virtual package and that does not seem a good idea.

What I propose is instead to create a `linera-sdk-tests`. That would have the same status as `examples`.
That would already gain something since the wasm code in the `create-and-call` / `publish-read-data-blob`
is compiled 4 times for wasm32 but executes 0 times in web assembly.

PR #4382, opened 2025-08-20
Introduce minimal tests

MathieuDutSik:
## Motivation

It has recently emerged that some interesting end-to-end tests are different from the usual ones (fungible, crowd-funding, matching engine, etc.):
* They use esoteric features like create applications.
* They test inner protocol features like accessing blobs from two transactions in the same block.
* They are simpler and may not have any storage.
* They are not using GraphQL API.
* Their minimal nature makes them less dependent on the underlying storage and the number of validators/shards.

It makes sense to separate those tests from the other tests. The end-to-end tests are expensive.

## Proposal

The idea is to call those tests "minimal tests" and introduce the file `minimal_tests.rs`, and test them
with just one validator, one shard, and the storage service.

We cannot call them "unit tests" since "unit end-to-end test" is not great.

Also, those tests cannot be unit tests, since the features being tested exercise all aspects of the code, but do not do so intensively. If however, that is not the case, then those tests can be extended very easily. We can switch from "minimal_test" to "new_test".

## Test Plan

The CI.

## Release Plan

## Links

None.


ma2bd:
The e2e tests in linera-service were always meant to test the protocol features, not the applications.

IMHO what we're missing is integration tests, defined in the directory of the example apps.

afck:
And tests about the execution semantics could be in `wasm_client_tests` in `linera-core`. Those should be much faster than end-to-end tests in `linera-service`.

MathieuDutSik:
> And tests about the execution semantics could be in `wasm_client_tests` in `linera-core`. Those should be much faster than end-to-end tests in `linera-service`.

Speed is relative. For example, the message in instantiation test is 6.26s in integration test and 12.82s in end-to-end test. Not that much faster.

MathieuDutSik:
This idea could still make sense.
However, as a first, we would need some tests for which the integration test, the linera-core test or the linera-execution test are not sufficient.

PR #4383, opened 2025-08-20
Move Web examples into `linera-protocol`

Twey:
## Motivation

We'd like to develop and deploy the Web examples along with `linera-protocol`.

## Proposal

Copy the files from `linera-web` commit c92f521b4a2ada81242560c327eb457e9593eb85.

It would be nice to merge them using `git subtree` or similar, but since we forbid merge commits in `main` there's no alternative but to destroy the history.

## Release Plan

    - be released in a new SDK,
    - be released in a new SDK,
      SDK soon.

## Links


examples/fungible/README.md
```diff
@@ -104,15 +104,15 @@ for the account.
...
         \"$OWNER_1\": \"100.\"
     } }" \
     --json-parameters "{ \"ticker_symbol\": \"FUN\" }" \
 )
 ```
 
-This will store the application ID in a new variable `APP_ID`.
+This will store the application ID in a new variable `LINERA_APPLICATION_ID`.
 
 ### Using the Token Application
```
ma2bd: We should be able to split this section to clearly separate the Web frontend (for end users in the Web) and the CLI tool (for developers, AI agents, services, etc).

Twey: I'm not too sure what you mean there — all documentation to do with the Web frontend is in the section at the bottom.  The variables need to be renamed only because Vite wants a prefix for all build variables, so rather than introducing new variables at the end I renamed the existing ones, but there's nothing Web-specific here.

Are you saying you'd like the documentation for the Web frontend to be in a separate file?

ma2bd: (Concretely, starting the node service --just below this comment-- makes no sense for Web users)

Twey: We could restructure the document into two top-level sections, one on the backend (application deployment, GraphQL) and one on the frontend (frontend building)?

ma2bd: Yes, that's what I meant

Twey: Then, perhaps three — backend deployment, accessing using the node service, frontend?

Twey: Done in https://github.com/linera-io/linera-protocol/pull/4383/commits/c02691fb64422ec4bcfcb39d17bf5dae73db7e3c https://github.com/linera-io/linera-protocol/pull/4383/commits/070a94c11e17e3ddf2988ab8aeb9b0e032927b2e — let me know what you think.  I've moved the node service section below the Web frontend section to clarify that it's not necessary to run the frontend, and added some ‘alternatively,’ text to it to (hopefully) clarify that the Web frontend isn't necessary for the node service, either.

ma2bd: This looks great! Thanks!

linera-web/build.bash
```diff
@@ -7,12 +7,12 @@ cd $(dirname -- "${BASH_SOURCE[0]}")
...
 
-wasm_bindgen_cargo_version=$(cargo metadata --format-version 1 | jq -r '.packages[] | select(.name == "wasm-bindgen").version')
-target_dir=$(cargo metadata --format-version 1 | jq -r .target_directory)
+wasm_bindgen_cargo_version=$(cargo +nightly metadata --format-version 1 | jq -r '.packages[] | select(.name == "wasm-bindgen").version')
+target_dir=$(cargo +nightly metadata --format-version 1 | jq -r .target_directory)
 
 if [[ "$wasm_bindgen_cargo_version" != "$wasm_bindgen_cli_version" ]]
 then
-    cargo update --package wasm-bindgen --precise "$wasm_bindgen_cli_version"
+    cargo +nightly update --package wasm-bindgen --precise "$wasm_bindgen_cli_version"
```
ma2bd: Do we need to pin the version of nightly? Or use the rust toolchain?

ma2bd (approved):

linera-web/build.bash
Twey: With `rustup`, it'll be using the version from `rust-toolchain.toml`. This should be a no-op there, it's just that the rustup-style toolchain dispatcher I wrote for Nix doesn't read the toolchain file, but in that case it's pinned by Nix — so in both cases it's pinned already.

Twey: (this was not correct, `rustup` won't match this to the available nightly version but will just pull the latest nightly)

PR #4385, opened 2025-08-21
Fix "skipped proposal" scenario; take note of proposals even if they fail.

### LINKED ISSUE #2971: Take note of block proposals in the client without handling them.

afck:
## Motivation

Proposals by other chain owners that the client learns about from validators can fail to execute locally, e.g. due to different oracle values. But in multi-leader rounds, clients need to take not of such failed proposals anyway, so they know that they shouldn't also make a conflicting proposal in the same round.

## Proposal

Add another `signed_proposed_block` field to the chain manager to keep track of this.

## Test Plan

A regression test was added; and it failed without this fix.

## Release Plan

## Links

- Closes #2971.

### DIFF
--- a/linera-chain/src/manager.rs
+++ b/linera-chain/src/manager.rs
@@ -154,6 +154,12 @@ where
     /// The probability distribution for choosing a fallback round leader.
     #[cfg_attr(with_graphql, graphql(skip))] // Derived from validator weights.
     pub fallback_distribution: RegisterView<C, Option<WeightedAliasIndex<u64>>>,
+    /// Highest-round authenticated block that we have received, but not necessarily check yet.
+    /// If there are multiple proposals in the same round, this contains only the first one.
+    /// This can even contain proposals that did not execute successfully, to determine which round
+    /// to propose in.
+    #[cfg_attr(with_graphql, graphql(skip))]
+    pub signed_proposal: RegisterView<C, Option<BlockProposal>>,
     /// Highest-round authenticated block that we have received and checked. If there are multiple
     /// proposals in the same round, this contains only the first one.
     #[cfg_attr(with_graphql, graphql(skip))]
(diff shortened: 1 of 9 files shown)


linera-chain/src/manager.rs
```diff
@@ -154,6 +154,12 @@ where
     /// The probability distribution for choosing a fallback round leader.
     #[cfg_attr(with_graphql, graphql(skip))] // Derived from validator weights.
     pub fallback_distribution: RegisterView<C, Option<WeightedAliasIndex<u64>>>,
+    /// Highest-round authenticated block that we have received, but not necessarily check yet.
```
bart-linera:
```suggestion
    /// Highest-round authenticated block that we have received, but not necessarily checked yet.
```

bart-linera (approved):

PR #4387, opened 2025-08-21
Introduce another `KeyValueStoreView`

MathieuDutSik:
## Motivation

The `KeyValueStoreView` is a scalable solution. But it may be faster to use a `BTreeMap` based solution that is faster for small cases.

## Proposal

We introduce the `SmallKeyValueStoreView` that uses a `RegisterView` on a `BTreeMap`.
However, we cannot use the derive, so we write the `SmallKeyValueStoreView` directly.

The switch is accessible by using the feature `smallkeyvaluestoreview` so as to select the small
key-value-store view.

## Test Plan

The CI.

More benchmarks are needed. This is only a draft PR so far.

## Release Plan

## Links

None.

### DIFF
--- a/linera-execution/Cargo.toml
+++ b/linera-execution/Cargo.toml
@@ -15,6 +15,7 @@ version.workspace = true
 workspace = true
 
 [features]
+smallkeyvaluestoreview = []
 test = ["tokio/macros", "linera-base/test", "linera-views/test", "proptest"]
 revm = [
     "dep:revm",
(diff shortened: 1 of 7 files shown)


MathieuDutSik:
The reached consensus is that if a user wants this kind of functionality, it can be implemented onchain by using `RegisterView<BTreeMap<Vec<u8>,Vec<u8>>>`.

So, the PR is closed.

PR #4388, opened 2025-08-21
Remove the chain worker state guard; `rollback` explicitly.

afck:
## Motivation

The chain worker state guard rolls back the chain state view in its `Drop` impl. But `handle_request` is already a natural point to call `rollback`.

## Proposal

Remove the guard; call `rollback` in `handle_request`.

## Test Plan

CI

## Release Plan

## Links

### DIFF
--- a/linera-core/src/chain_worker/state.rs
+++ b/linera-core/src/chain_worker/state.rs
@@ -201,6 +201,9 @@ where
             warn!("Callback for `ChainWorkerActor` was dropped before a response was sent");
         }
 
+        // Roll back any unsaved changes to the chain state: If there was an error while trying
+        // to handle the request, the chain state might contain unsaved and potentially invalid
+        // changes. The next request needs to be applied to the chain state as it is in storage.
         self.chain.rollback();
     }
 


linera-core/src/chain_worker/state.rs
```diff
@@ -21,40 +27,444 @@ use linera_chain::{
...
+                        .await,
+                )
+                .is_ok(),
+        };
+
+        if !responded {
+            warn!("Callback for `ChainWorkerActor` was dropped before a response was sent");
+        }
+
+        self.chain.rollback();
```
afck, line 207: This is the new explicit `rollback` call.

Twey (approved):
You only live once!

linera-core/src/chain_worker/state.rs
Twey: I'd love a big scary comment here.

afck: Done in https://github.com/linera-io/linera-protocol/pull/4388/commits/f41fdf1284534e45845df8e679d85c1f144a4d5c.

deuszx: Someone, sometime, in the future, will come and add this to a `Drop` impl so that we don't have to think about it :P

afck: I actually think it's more likely to forget to use the wrapper when adding a new request variant than to introduce an early return in `handle_request` that doesn't even have `Result` as a return type.
Either way there's no 100% failsafe way to prevent future bugs.

deuszx: You wouldn't be able to forget if the API required existed only on the type.

afck: But if there's a new request type you'd implement the required API, and that's just as easy (or easier) directly on the chain worker state as it is on the guard.

PR #4392, opened 2025-08-21
Fix authenticated signer syscall and remove the offending field

ma2bd:
## Motivation

Fix authenticated signer.

## Proposal

As pointed out in #4374, we were using the transaction instead of the last application on the calling stack, effectively ignoring whether calls were "authenticated" or not.

* Fix the bug
* Remove the field that caused the confusion in the first place

## Test Plan

CI + new unit test

I verified that the new unit test fails before the fix.

## Release Plan

    - be released in a new SDK,

### DIFF
--- a/linera-execution/src/runtime.rs
+++ b/linera-execution/src/runtime.rs
@@ -87,9 +87,6 @@ pub struct SyncRuntimeInternal<UserInstance: WithContext> {
     height: BlockHeight,
     /// The current consensus round. Only available during block validation in multi-leader rounds.
     round: Option<u32>,
-    /// The authenticated signer of the operation or message, if any.
-    #[debug(skip_if = Option::is_none)]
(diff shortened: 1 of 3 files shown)


linera-execution/src/runtime.rs
```diff
@@ -1085,7 +1079,6 @@ impl ContractSyncRuntimeHandle {
 
         {
             let runtime = self.inner();
-            assert_eq!(runtime.authenticated_signer, action.signer());
             assert_eq!(runtime.chain_id, chain_id);
```
ndr-ds, line 1082: Do you have to add this here instead now?
```suggestion
            assert_eq!(runtime.current_application().signer, action.signer());
            assert_eq!(runtime.chain_id, chain_id);
```

ma2bd: There is no current application yet.

PR #4394, opened 2025-08-22
Move the `create-and-call` test case from end-to-end to integration test.

MathieuDutSik:
## Motivation

The end-to-end tests are relatively expensive in CI. So, it is advantageous to move them to integration tests.

## Proposal

We do the following:
* We remove the `&self` of `ActiveChain` that is not needed.
* We make the function `ActiveChain` public to be used in the operational test.
* We introduce the function to return `Bytecode` since to instantiate, we cannot use the compressed bytecode.
* We introduce the test that exercises the `create-and-call` smart contract.

## Test Plan

The CI.

The PR that introduces the `create-and-call` also addresses the persistence of the contract and service blobs
that have been created. We tested that this integration test fails if the correction is removed.

## Release Plan

## Links

None.

### DIFF
--- a/examples/create-and-call/tests/test_create_and_call.rs
+++ b/examples/create-and-call/tests/test_create_and_call.rs
@@ -13,7 +13,7 @@ use linera_sdk::test::{ActiveChain, TestValidator};
 /// application to dynamically create an instance of the counter and increment it.
 /// The test verifies that the counter is correctly initialized and incremented.
 #[tokio::test(flavor = "multi_thread")]
-async fn test_create_and_call_end_to_end() {
+async fn test_create_and_call() {
     let (validator, create_call_module_id) =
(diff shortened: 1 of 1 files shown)


ma2bd (approved):
Awesome, thanks!
Then, maybe we could move everything to `linera-sdk/tests` ?

MathieuDutSik:
> Awesome, thanks! Then, maybe we could move everything to `linera-sdk/tests`?

I looked at this problem of moving the tests to `linera-sdk/tests`, but it does not seem easy.
We would have to change `linera-sdk` to be a virtual package and that does not seem a good idea.

What I propose is instead to create a `linera-sdk-tests`. That would have the same status as `examples`.
That would already gain something since the wasm code in the `create-and-call` / `publish-read-data-blob`
is compiled 4 times for wasm32 but executes 0 times in web assembly.

PR #4396, opened 2025-08-22
Add a script to stress-test faucets

Twey:
## Motivation

On Testnet Babbage, we experienced a faucet failure (rather, DoS) when many people tried to create wallets and chains at the same time.  To make sure it doesn't happen again, we'd like to stress-test the faucet with artificial data.

## Proposal

Add a small bash script that executes `linera wallet init; linera wallet request-chain` against a faucet many times in parallel, printing timing information on stdout in JSON for analysis.

## Test Plan

Run against a local faucet.

## Release Plan

## Links

### DIFF
--- a/scripts/faucet-stress-test.bash
+++ b/scripts/faucet-stress-test.bash
@@ -1,0 +1,49 @@
+#!/usr/bin/env bash
+
+faucet_url=${1-http://localhost:8080}; shift
+count=${1-100}; shift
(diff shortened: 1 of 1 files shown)


deuszx (commented):
Can you expand this with the statistics-gathering Python script and use that by defualt? I think it gives much better picture about the distribution, not only blind stress-testing.

Twey:
I don't want to conflate the stress-testing script with a statistical analysis tool, though if you'd like to commit the analysis script we can pipe this and other things (e.g. benchmarks) into it?

PR #4399, opened 2025-08-24
Switch the `Blob` from `Box<[u8]>` to `Arc<Box<[u8]>>`.

### LINKED ISSUE #2389: Change blob bytes from `Vec<u8>` to `Arc<[u8]>`
This will help us prevent unnecessary copying of the blob data

MathieuDutSik:
## Motivation

We are doing some cloning operations for blobs. This occurs in the `TransactionTracker` with the `previously_created_blobs`.

Fixes #2389

## Proposal

The implementation is done with the `serde_with` for the serialization.

## Test Plan

The CI.
A test has been introduced for this serialization and hash.

## Release Plan

## Links

None

### DIFF
--- a/linera-base/src/data_types.rs
+++ b/linera-base/src/data_types.rs
@@ -1320,28 +1320,10 @@ impl BlobContent {
         &self.bytes
     }
 
-    /// Gets the inner blob's bytes, consuming the blob.
-    pub fn into_bytes(self) -> Arc<Box<[u8]>> {
-        self.bytes
-    }
-
-    /// Gets the inner blob's bytes, consuming the blob.
-    /// Returns Some(Vec<u8>) if there's only one reference to the data,
-    /// None if there are multiple references.
-    ///
-    /// Note: Due to Rust's type system limitations with Arc<[u8]> (unsized type),
-    /// this method cannot avoid cloning entirely, but it only succeeds when
-    /// there's a single reference, making the clone unnecessary in most practical cases.
-    pub fn into_vec(self) -> Option<Vec<u8>> {
-        // Check if we're the sole owner (no other strong or weak references)
-        if Arc::strong_count(&self.bytes) == 1 && Arc::weak_count(&self.bytes) == 0 {
-            // We're the only owner, so extract the data
-            // Unfortunately, Arc<[u8]> doesn't support try_unwrap due to being unsized,
-            // so we must use this approach
-            Some((*self.bytes).to_vec())
-        } else {
-            None
-        }
+    /// Convert a BlobContent into `Vec<u8>` without cloning if possible.
+    pub fn into_vec_or_clone(self) -> Vec<u8> {
+        let bytes = Arc::unwrap_or_clone(self.bytes);
+        bytes.into_vec()
     }
 
     /// Returns the type of data represented by this blob's bytes.
@@ -1457,11 +1439,6 @@ impl Blob {
     /// Gets a reference to the inner blob's bytes.
     pub fn bytes(&self) -> &[u8] {
         self.content.bytes()
-    }
-
-    /// Gets the inner blob's bytes.
-    pub fn into_bytes(self) -> Arc<Box<[u8]>> {
-        self.content.into_bytes()
     }
 
     /// Loads data blob from a file.
--- a/linera-base/src/data_types.rs
+++ b/linera-base/src/data_types.rs
@@ -1265,7 +1265,10 @@ impl BlobContent {
     /// Creates a new [`BlobContent`] from the provided bytes and [`BlobId`].
     pub fn new(blob_type: BlobType, bytes: impl Into<Box<[u8]>>) -> Self {
         let bytes = bytes.into();
-        BlobContent { blob_type, bytes: Arc::new(bytes) }
+        BlobContent {
+            blob_type,
+            bytes: Arc::new(bytes),
+        }
     }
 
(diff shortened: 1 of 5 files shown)


linera-rpc/tests/snapshots/format__format.yaml.snap
```diff
@@ -114,7 +114,8 @@ BlobContent:
   STRUCT:
     - blob_type:
         TYPENAME: BlobType
-    - bytes: BYTES
+    - bytes:
+        SEQ: U8
```
ma2bd: `BYTES` is better

linera-base/src/data_types.rs
```diff
@@ -1244,28 +1245,112 @@ impl CompressedBytecode {
     }
 }
 
+impl Serialize for BlobContent {
```
ma2bd: Do we need this boilerplate?

ma2bd: Did you miss this ?
https://serde.rs/feature-flags.html#rc

MathieuDutSik: Yes, thank you.

linera-rpc/tests/snapshots/format__format.yaml.snap
MathieuDutSik: Yes, but how to do it?
This type is written by `serde-reflection::Tracer`, and the configuration of the tracer does not appear to leave room for choosing `BYTES`.

afck: Right, we'd need https://github.com/serde-rs/bytes/issues/43 first. But the last comment on that issue says `serde_with::Bytes` is an alternative.

MathieuDutSik: Thank you for pointing this out.

However, the `serde_as` works for `Arc<Box<[u8]>>` but not for `Arc<[u8]>`.

afck: ~That's weird! `Arc<Box<[u8]>>` seems like an unnecessary redirection.~ (Edit: I get it now!)

Anyway, we could still use `#[serde(with = "module")]` directly, and do the conversion ourselves.

linera-base/src/data_types.rs
```diff
@@ -1247,25 +1248,24 @@ impl CompressedBytecode {
 impl BcsHashable<'_> for BlobContent {}
 
 /// A blob of binary data.
-#[derive(Hash, Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
```
bart-linera, line 1250: Nit: unnecessary change of the order of derives (probably introduced by removing those and re-adding them)

MathieuDutSik: Ok, thanks.

linera-rpc/tests/snapshots/format__format.yaml.snap
MathieuDutSik: Ok, found a way to do that.

linera-execution/src/runtime.rs
```diff
@@ -946,7 +946,7 @@ where
             this.transaction_tracker
                 .replay_oracle_response(OracleResponse::Blob(blob_id))?;
         }
-        Ok(content.into_bytes().into_vec())
+        Ok(content.into_bytes().to_vec())
```
afck: Now we are always cloning the data here. :confused: 
So maybe it actually makes sense to use `Arc<Box<[u8]>>` after all? Then we could use `Arc::unwrap_or_clone` here.

But maybe _this_ method should return `Arc<[u8]>` (or `Arc<Box<[u8]>>`) anyway? There's no reason to move out of the `Arc` yet.

MathieuDutSik: Yes, I agree. That clone is horrible, since we know it is not needed.
So, `Arc<Box<[u8]>>` is maybe the solution.

linera-base/src/data_types.rs
```diff
@@ -1313,9 +1320,10 @@ impl BlobContent {
         &self.bytes
     }
 
-    /// Gets the inner blob's bytes, consuming the blob.
-    pub fn into_bytes(self) -> Box<[u8]> {
-        self.bytes
+    /// Convert a BlobContent into `Vec<u8>` without cloning if possible.
+    pub fn into_vec_or_clone(self) -> Vec<u8> {
```
afck, line 1324: Actually, why do we _ever_ want to convert blob bytes from `Box<[u8]>` to `Vec<u8>`? We never _modify_ them, so I'd just use `Box<[u8]>` or `Arc<Box<[u8]>>` everywhere.

(It might turn out that `Arc<[u8]>` is good enough after all—sorry! Because if we never need to convert it to a `Vec`, do we even ever need to convert it to `Arc`? Or can we simply always use `Arc<[u8]>`?)

MathieuDutSik: You are right that the data being read is never modifed so having some storage of the kind `Box<[u8]>` or `Arc<Box<[u8]>>` for them makes them.

But somehow the implications are wide-ranging:
* The `CompressedBytecode` function.
* The reading of data blob with `read_data_blob`. Which has to pass the wit boundary.

So, let's think before getting into that.

afck (approved):
No blockers from my side, but as a next step we should consider using `Arc<[u8]>` or `Arc<Box<[u8]>>` everywhere, including `CompressedBytecode` etc., to avoid even more cloning.

PR #4400, opened 2025-08-24
Make the handling of GraphQL in the counter example in line with other contracts.

### LINKED ISSUE #2854: Use the generated `async_graphql::SimpleObject` in the counter example
The counter example doesn't demonstrate how `async_graphql::SimpleObject` should be used.

This concerts the code and the "service" page of the dev manual.

MathieuDutSik:
## Motivation

The `counter` example was processing GraphQL queries quite differently from other examples.
We put it in line with other examples.

Fixes #2854 

## Proposal

The work is fairly standard, but forces the introduction of a `CounterOperation` since the input
must satisfy the `GraphQLMutationRoot` trait.

## Test Plan

The CI.

## Release Plan

## Links

None.

### DIFF
--- a/examples/counter/README.md
+++ b/examples/counter/README.md
@@ -88,7 +88,7 @@ query {
 - To increase the value of the counter by 3, perform the `increment` operation.
 ```gql,uri=http://localhost:8080/chains/$CHAIN_1/applications/$APPLICATION_ID
 mutation Increment {
-  increment(value: 3)
+  increment(field0: 3)
 }
 ```
 - Running the query again would yield `4`.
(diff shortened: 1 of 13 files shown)


examples/counter/README.md
```diff
@@ -88,7 +88,7 @@ query {
 - To increase the value of the counter by 3, perform the `increment` operation.
 ```gql,uri=http://localhost:8080/chains/$CHAIN_1/applications/$APPLICATION_ID
 mutation Increment {
-  increment(value: 3)
+  increment(field0: 3)
```
ma2bd, line 91: my bad I shouldn't have let this go through

PR #4403, opened 2025-08-25
Remove the `Account` duplication.

### LINKED ISSUE #2496: Merge `fungible::Account` and `identifiers::Account` type
# Motivation

There are two `Account` types in the SDK, which can be confusing for new developers.

# Proposal

Merge the `fungible` type into the generic `Account` type. This might require making the `fungible` application support "chain accounts", i.e., accounts on a chain where any owner can use the tokens.

### LINKED ISSUE #3462: GraphQL cannot parse `Account` structure from its format string

MathieuDutSik:
## Motivation

We have the `Account` in two contexts:
* The `linera-sdk/src/abis/fungible.rs`
* The `linera-base/src/identifiers.rs`
which is not great. This PR unifies both types.

Fixes #2496 
Fixes #3462 

## Proposal

It is a given fact that in Linera, we will have a type `Account` that contains both the `ChainId` and the `AccountOwner`. An additional complication is that GraphQL forces to have one type in input and another type in output.

The current design is to use the `fungible::Account` for the GraphQL with the following (simplified) design:
```rust
#[derive(SimpleObject, InputObject)]
#[graphql(input_name = "FungibleAccount")]
struct Account { .... }
```

with this design the input type is `FungibleAccount` and the output type is `Account`. This is inadequate since an account could have many usage besides `FungibleAccount`. This is the error visible in issue #3462. Instead the types should reflect the nature of the type; things like `AccountInput`, `AccountOutput` and so on.

Generally, we assign a specific type to the input with something like `TypeInput`. That is not great in our context. The type `Account` will mostly be on input, and rarely on output. Therefore, we choose to use `Account` on input and `AccountOutput` on output. This is done via the following:
```rust
#[graphql(name = "AccountOutput", input_name = "Account")]
```

With this change, we can simplify the `FungibleAccount` into `Account` for the input. The output type `AccountOutput` concerns only the refund grant to entry for cross-chain messages.

## Test Plan

The CI addresses it.
However, the change from `FungibleAccount` to `Account` means that some code in other crates may need to be updated as well.

## Release Plan

## Links

None.

### DIFF
--- a/linera-service-graphql-client/src/service.rs
+++ b/linera-service-graphql-client/src/service.rs
@@ -5,7 +5,7 @@ use graphql_client::GraphQLQuery;
 use linera_base::{
     crypto::CryptoHash,
     data_types::{Amount, Blob, BlockHeight, ChainDescription, OracleResponse, Round, Timestamp},
-    identifiers::{Account, AccountOwner, BlobId, ChainId, GenericApplicationId, StreamName},
+    identifiers::{AccountOwner, BlobId, ChainId, GenericApplicationId, StreamName},
 };
 use thiserror::Error;
 
@@ -138,7 +138,10 @@ pub enum ConversionError {
 
 #[cfg(not(target_arch = "wasm32"))]
 mod from {
-    use linera_base::{data_types::Event, identifiers::{Account, StreamId}};
+    use linera_base::{
+        data_types::Event,
+        identifiers::{Account, StreamId},
+    };
     use linera_chain::{
         block::{Block, BlockBody, BlockHeader},
         types::ConfirmedBlock,


MathieuDutSik:
> I don't understand why we need separate input type for the `Account` which is the exact same type as `type AccountOutput` – but maybe that's the limitation of GraphQL itself?
> 
> Otherwise looks good!

Yes, limitation of GraphQL. You have types in input and types in output.

PR #4405, opened 2025-08-25
Implement composed operations

### LINKED ISSUE #2037: Support composed operations (aka scripts)

MathieuDutSik:
## Motivation

Due to the control of the order of the operations by the signer of the blocks, the signer can choose the return of one operation to be the input of the next one.

Fixes #2037 

## Proposal

The idea is simply to introduce a
```rust
enum OperationInput {
  Composed,
  Direct(Vec<u8>),
}
```

with `Direct` being the previous mechanism. Since the `BlockTracker` has all the needed data, this works fine.

Where some problems occur is with the typed operations of Wasm smart contracts. We could introduce an `OperationInput` taking a typed input. But this introduces a number of type complications, and the method instead is to have two functions:
```rust
schedule_operation(A::Oeration)
schedule_composed_operation()
```

And so we get a relatively simple 

## Test Plan

The CI.
One test uses the `Counter` example which was modified to return a `CounterOperation` and so to be composed.

## Release Plan

## Links

None.

### DIFF
--- a/examples/counter/src/contract.rs
+++ b/examples/counter/src/contract.rs
@@ -45,11 +45,11 @@ impl Contract for CounterContract {
         self.state.value.set(value);
     }
 
-    async fn execute_operation(&mut self, operation: CounterOperation) -> u64 {
+    async fn execute_operation(&mut self, operation: CounterOperation) -> CounterOperation {
         let CounterOperation::Increment { value } = operation;
         let new_value = self.state.value.get() + value;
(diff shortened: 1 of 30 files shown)


ma2bd (commented):
This is related to "scripts" or Sui's [PTB](https://docs.sui.io/concepts/transactions/prog-txn-blocks). Great direction but we're going to need some solid design first.

PR #4406, opened 2025-08-25
Move some tests to a separate directory.

MathieuDutSik:
## Motivation

Several tests have emerged that are unit test in nature but require a full application to be built.
Currently, the source code of those tests is in `examples` which makes them sit alongside more
functional applications that are of more general interest. 

## Proposal

The tests are moved to the directory `linera-sdk/tests/fixtures`. They are not part of the strusture of the
cargo crate.

The advantage is that those smart contracts are no longer compiled again and again (ScyllaDb, DynamoDb,
StorageService, RemoteNet), and so we gain on CI here. The cost of compiling for tests in `--release` is about
2-3 minutes per such instance.

The cost of running `linera-sdk/tests/fixtures` is 7 minutes, so overall, we have a speed gain from this work.

## Test Plan

The CI.
The same tests are being tested. The same tests are being applied, just differently ordered.

## Release Plan

## Links

None.

### DIFF
--- a/linera-service/tests/readme_test.rs
+++ b/linera-service/tests/readme_test.rs
@@ -31,7 +31,7 @@ use tokio::process::Command;
 #[test_case::test_case("../examples/native-fungible" ; "native-fungible")]
 #[test_case::test_case("../examples/non-fungible" ; "non-fungible")]
 #[test_case::test_case("../examples/matching-engine" ; "matching engine")]
-#[test_case::test_case("../examples/meta-counter" ; "meta counter")]
+#[test_case::test_case("../linera-sdk/tests/fixtures/meta-counter" ; "meta counter")]
 #[test_case::test_case("../examples/rfq" ; "requests for quotes")]
 #[test_case::test_case("../examples/social" ; "social")]
 #[test_log::test(tokio::test)]


.github/workflows/rust.yml
```diff
@@ -130,6 +130,10 @@ jobs:
       run: |
         cd examples
         cargo test --locked
+    - name: Run linera-sdk tests
+      run: |
+        cd linera-sdk-tests
```
ma2bd: Can this directory be `linera-sdk/tests/fixtures` or `linera-sdk/tests/examples` or something like that?

MathieuDutSik: Yes. I restructured the PR in that way.

examples/Cargo.toml
```diff
@@ -18,10 +16,8 @@ members = [
     "meta-counter",
     "native-fungible",
     "non-fungible",
-    "publish-read-data-blob",
     "rfq",
     "social",
-    "track-instantiation",
```
afck, line 24: Not sure if they should be removed from the workspace. At least we want to know if they fail to build due to some change in their dependencies.

linera-execution/src/wasm/mod.rs
```diff
@@ -364,14 +363,27 @@ pub mod test {
...
 
+    fn build_applications() -> Result<(), std::io::Error> {
+        for dir in ["../examples", "../linera-sdk/tests/fixtures"] {
+            build_applications_in_directory(dir)?;
+        }
+        Ok(())
+    }
+
     pub fn get_example_bytecode_paths(name: &str) -> Result<(String, String), std::io::Error> {
+        use std::path::Path;
```
afck: Should be at the top of the file.

linera-sdk/tests/fixtures/contract-call/res
```diff
@@ -0,0 +1,11 @@
+error: failed to load manifest for workspace member `/Users/mathieudutoursikiric/GITlineraOpen/GITlinera_sdk_tests/linera-protocol_second/linera-sdk-tests/create-and-call`
```
afck: That file was probably added by mistake?

examples/Cargo.toml
MathieuDutSik: But, isn't it the code tested in
```bash
  linera-sdk-tests-fixtures:
    needs: changed-files
    runs-on: ubuntu-latest
    timeout-minutes: 10

    steps:
    - uses: actions/checkout@v4
    - name: Install Protoc
      uses: arduino/setup-protoc@v3
      with:
        repo-token: ${{ secrets.GITHUB_TOKEN }}
    - name: Run linera-sdk fixtures
      run: |
        cd linera-sdk/tests/fixtures
        cargo test --locked
    - name: Run linera-sdk tests application lints
      run: |
        cd linera-sdk/tests/fixtures
        cargo fmt -- --check
        cargo clippy --all-targets --all-features --target wasm32-unknown-unknown --locked
```

So, I think it is still being tested.

linera-execution/src/wasm/mod.rs
MathieuDutSik: Ok, done.

linera-sdk/tests/fixtures/contract-call/res
MathieuDutSik: So, sorry.

examples/Cargo.toml
afck: Yes, in CI. I'm just saying if you run `cargo check` or Clippy locally, it won't apply to those crates anymore. Probably not a problem.

MathieuDutSik: Yes, but hasn't this been the situation forever?
We need to run `cargo clippy` in `linera-protocol` and `linera-protocol/examples`.
So, now there is another place where we need to run this. I agree it is not optimal.

examples/Cargo.toml
```diff
@@ -18,10 +16,8 @@ members = [
     "meta-counter",
```
ma2bd: How about `meta-counter` as well?

linera-execution/src/wasm/mod.rs
ma2bd: did you?

linera-sdk/tests/fixtures/contract-call/res
ma2bd: still there

MathieuDutSik: Thanks, forgotten to commit and push.

linera-execution/src/wasm/mod.rs
MathieuDutSik: No, I had forgotten to commit and push.

examples/Cargo.toml
MathieuDutSik: Yes, I moved it as well.

examples/Cargo.toml
ma2bd: How many workspaces (or crates without a workspace) do we have after this PR?

MathieuDutSik: Currently, we have two workspaces: `linera-protocol` and `linera-protocol/examples`.
After this PR, we have 3.

ma2bd: Makes sense

ma2bd (approved):
Thanks for the cleanup PR. Much appreciated

PR #4411, opened 2025-08-26
Extend the support of GraphQLMutationRoot to structs

### LINKED ISSUE #2534: Support `struct`s in `GraphQLMutationRoot` derive macro
# Motivation

Currently the `GraphQLMutationRoot` derive macro only supports `enum`s. That is a minor nuisance for applications that want to support a single operation.

# Proposal

Treat `struct`s like `enum`s with a single variant.

MathieuDutSik:
## Motivation

The `GraphQLMutationRoot` works only for enums. There are other inputs where it uses a struct.

Fixes #2534 

## Proposal

The code is coded via Claude.

## Test Plan

The CI.
The `Operation` input type is corrected for `counter`, `counter-no-graphql`, and `create-and-call`.

## Release Plan

## Links

None

### DIFF
--- a/examples/counter-no-graphql/src/contract.rs
+++ b/examples/counter-no-graphql/src/contract.rs
@@ -47,8 +47,7 @@ impl Contract for CounterContract {
 
     async fn execute_operation(&mut self, operation: CounterOperation) -> u64 {
         let previous_value = self.state.value.get();
-        let CounterOperation::Increment(value) = operation;
-        let new_value = previous_value + value;
+        let new_value = previous_value + operation.increment;
         self.state.value.set(new_value);
         new_value
     }
@@ -75,7 +74,7 @@ mod tests {
         let mut counter = create_and_instantiate_counter(initial_value);
 
         let increment = 42_308_u64;
-        let operation = CounterOperation::Increment(increment);
+        let operation = CounterOperation { increment };
 
         let response = counter
             .execute_operation(operation)
@@ -106,7 +105,7 @@ mod tests {
         let mut counter = create_and_instantiate_counter(initial_value);
 
         let increment = 8_u64;
-        let operation = CounterOperation::Increment(increment);
+        let operation = CounterOperation { increment };
 
         let response = counter
             .execute_operation(operation)
--- a/examples/counter-no-graphql/src/lib.rs
+++ b/examples/counter-no-graphql/src/lib.rs
@@ -25,6 +25,6 @@ pub enum CounterRequest {
 }
 
 #[derive(Debug, Serialize, Deserialize)]
-pub enum CounterOperation {
-    Increment(u64),
+pub struct CounterOperation {
+    pub increment: u64,
 }
--- a/examples/counter-no-graphql/src/service.rs
+++ b/examples/counter-no-graphql/src/service.rs
@@ -49,8 +49,8 @@ impl Service for CounterService {
 
         match request {
             CounterRequest::Query => *self.state.value.get(),
-            CounterRequest::Increment(value) => {
-                let operation = CounterOperation::Increment(value);
+            CounterRequest::Increment(increment) => {
+                let operation = CounterOperation { increment };
                 self.runtime.schedule_operation(&operation);
                 0
             }
--- a/examples/counter/README.md
+++ b/examples/counter/README.md
@@ -87,9 +87,7 @@ query {
 ```
 - To increase the value of the counter by 3, perform the `increment` operation.
 ```gql,uri=http://localhost:8080/chains/$CHAIN_1/applications/$APPLICATION_ID
-mutation Increment {
-  increment(field0: 3)
-}
+mutation { counterOperation(increment: 3) }
 ```
 - Running the query again would yield `4`.
 
--- a/examples/counter/src/contract.rs
+++ b/examples/counter/src/contract.rs
@@ -46,8 +46,7 @@ impl Contract for CounterContract {
     }
 
     async fn execute_operation(&mut self, operation: CounterOperation) -> u64 {
-        let CounterOperation::Increment(operation) = operation;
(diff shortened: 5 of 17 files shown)


MathieuDutSik:
> I actually thing it's a good design that `Operation` has to be an enum. It's enforcing some structure on dApp developers and that's a good thing. It's easier to learn.

TLDR: I do agree that restricting the allowed types on input makes sense. However, I do not see why forbidding structs makes sense.

Investigating the other blockchains using smart contracts based in Rust I found:
* For Solana, both structs and enums are fine.
* For CosmWasm, both structs and enums are fine (JSON is being used).
* NEAR (near-sdk-rs): both structs and enums are fine (JSON is being used)
* Substrate/ink! (SCALE): both structs and enums are fine.

So, I do not know any Blockchain languages where structs are explicitly forbidden on input. I also do not see why enums are easier to learn. But maybe you can explain.

There are also language where you must use structs:
* The EVM in Ethereum, Avalanche, Polygon, etc. does not have non-trivial enums.
* Move (Aptos/Sui) does not have directly enums, but with some design patterns, we can get the same.

There are other recommendations that I think make sense:
* No anonymous entries in enum. So, do not allow for `Pay(AccountOwner, Amount)` but instead do `Pay(recipient: AccountOwner, amount: Amount)`. This is currently allowed in Linera and it leads to "increment(field0: 3)" which is not great.
* Forbid anonymouns, types on input. For example, in an enum variant "Operation::Transfer(string,string,u64)" should be forbidden in favor of "Operation::Transfer(Account,Account,Amount)". And if those types have to nest a single entry like "Account(string)" then so be it.
* Introduce specific types like "NonZeroAmount" for forbidding ZERO "Amount" in input.
* Forbid specific types like `BTreeMap` on input.
* Specify a maximal size on input.

Conclusion:
* It does make sense to have some recommendation on the types allowed on the input of `execute_operation`.
* Those constraints should be tested at compilation time. For example by the macro `GraphQLMutationRoot`.
* I do not see why structs are problematic.
* Anonymous input ARE problematic (and are currently allowed)

ma2bd:
> For Solana, both structs and enums are fine.

Are we talking about the same thing?

What I'm understanding of this discussion so far is that we support:
```
enum Operation { Command1(Arg1), .. Commandn(ArgN) }   //  or Command1 { ... } etc
```
and in the (rare) case, when an app has a single type of command:
```
enum Operation { Command1(Arg1) }
```
you're advocating for supporting
```
struct Command(Arg)   //  or Command { ... }
```
instead.

1) First, do we agree that this is always going to be generating a single GraphQL mutation API?
2) Personally, I'm not opposed to the idea (assuming we agree on 1) but it is really not urgent. We can add the feature any time we want.

ma2bd (changes requested):
I don't necessarily mind the feature but to @deuszx' point, I believe using a `struct` in our `Counter` example is a very bad idea because it puts people on the wrong track.

* First, people need to understand how `enum Operation { Command1(Arg1) ... OperationN(ArgN) }` works. 
* Then when `N=1`, they may choose to write `struct Operation(Arg)` instead of an enum.

(p.s. All examples above are written with tuples but I also mean braces like `Command1 { ... }`)

MathieuDutSik:
> I don't necessarily mind _supporting_ the feature but to @deuszx' point, I believe using it in our `Counter` example is a very bad idea because it puts people on the wrong track.
> 
> * First, people need to understand how `enum Operation { Command1(Arg1) ... OperationN(ArgN) }` works.
> * Then when `N=1`, they may choose to write `struct Operation(Arg)` instead of an enum.
> 
> It one starts with `struct Operation(Arg)` then wants to add a command, they may end up with `Arg` being an enum, which will still generate exactly one GraphQL mutation API.
> 
> p.s. All examples above are written with tuples but I also mean braces like `Command1 { ... }`.

Ok, I don't mind not allowing struct. It is just a little strange.

But if I want to target a lower-hanging fruit, what about
```rust
pub enum CounterOperation {
    /// Increment the counter by the given value
    Increment(u64),
}
```

It leads to
```GraphQL
mutation Increment {
  increment(field0: 3)
}
```

Can we agree that it is not great?

The first item `field0` is puzzling. And it can be detected at compile time, so we can prevent its use.

Another thing that is problematic is `CounterOperation`. It appears in the types, but does not show up in the mutation.

ma2bd:
Just define
```
pub enum CounterOperation {
    Increment { value: u64) }
}
```
then (based on the README of fungible) I assume this query should work:
```
mutation {
  increment(value: 3)
}
```

deuszx:
I don't know all the intricate details of all other blockchains but on Substrate – IIRC – the endpoints/messages are indeed defined as structs but the macro `ink::message` compiles it down to an enum eventually. If we ever get to having a DSL/macro library for writing contracts on Linera, we might end up defining them as standalone structs/enums but in the end they become a single type – enum.

MathieuDutSik:
> Just define
> 
> ```
> 
> pub enum CounterOperation {
> 
>     Increment { value: u64) }
> 
> }
> 
> ```
> 
> then (based on the README of fungible) I assume this query should work:
> 
> ```
> 
> mutation {
> 
>   increment(value: 3)
> 
> }
> 
> ```

Yes, of course.
But that is not the point of my proposal.

We can prevent the field0 construction at the level of the compilation in the macro code of GraphQLmutationRoot.

ma2bd:
This shows that we don't need to address #2534. So let's just define `pub enum CounterOperation { Increment { value: u64) } }` and move on.

MathieuDutSik:
> This shows that we don't need to address #2534. So let's just define `pub enum CounterOperation { Increment { value: u64) } }` and move on.

Yes, but closing 2534 does not end the story:
* We should report an explicit error when compiling with a struct.
* We should forbid anonymous enums in the input.

PR #4413, opened 2025-08-26
Make execution state actor borrow txn tracker & resource controller.

afck:
## Motivation

The transaction tracker is a field in the runtime but also passed into some execution state methods. In some cases, we take the tracker from the runtime, send it to the execution state actor, send it back and put it back into the runtime field. This is confusing and error-prone.

## Proposal

Make the `ExecutionStateActor` a struct that borrows the state, tracker and resource controller.

## Test Plan

CI

## Release Plan

## Links

### DIFF
--- a/linera-execution/src/runtime.rs
+++ b/linera-execution/src/runtime.rs
@@ -12,7 +12,7 @@ use custom_debug_derive::Debug;
 use linera_base::{
     data_types::{
         Amount, ApplicationPermissions, ArithmeticError, Blob, BlockHeight, Bytecode,
-        OracleResponse, SendMessageRequest, Timestamp,
+        SendMessageRequest, Timestamp,
     },
     ensure, http,
     identifiers::{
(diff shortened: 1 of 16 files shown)


linera-execution/src/runtime.rs
```diff
@@ -1732,14 +1639,20 @@ impl ServiceSyncRuntime {
                 callback,
             } = request;
 
-            self.prepare_for_query(context);
+            if let Err(error) = self.prepare_for_query(context) {
+                let _ = callback.send(Err(error));
```
ma2bd: Please let's not use `let _ =`
(Ok, I can see it was there before, but still bad.)

afck:
~Closing this, because it doesn't work with the service oracle:~

~The problem is that service queries use the same execution state sender. So we must make sure that we never modify the execution state or the transaction tracker based on something that happens in a service oracle query! So we can't move the tracker into the actor.~

linera-execution/src/runtime.rs
afck: Done in https://github.com/linera-io/linera-protocol/pull/4413/commits/5ba317eb1d60d3214454059285a9c52a30e64e0b.

(There are still plenty of `let _ =` left in other places in the code, though, which we should probably address in a separate PR.)

linera-execution/src/runtime.rs
```diff
@@ -313,7 +310,6 @@ impl<UserInstance: WithContext> SyncRuntimeInternal<UserInstance> {
         deadline: Option<Instant>,
         refund_grant_to: Option<Account>,
         resource_controller: ResourceController,
```
bart-linera, line 312: Is it okay that the runtime still has its own resource controller, completely separate from the ExecutionStateActor's resource controller?

linera-execution/src/system.rs
```diff
@@ -886,15 +877,15 @@ where
         module_id: ModuleId,
         parameters: Vec<u8>,
         required_application_ids: Vec<ApplicationId>,
-        mut txn_tracker: TransactionTracker,
+        txn_tracker: &mut TransactionTracker,
```
bart-linera, line 880: :tada:

linera-execution/src/runtime.rs
afck: I think we already had two before this PR, and merged them afterwards. I wonder whether that can be changed, too, but it might be a bigger issue.

bart-linera (approved):
The duplicated ResourceControllers were my only issue, but since we had them before already, it should be fine to leave them these way for now :+1:

PR #4418, opened 2025-08-27
Use of the `Arc<Box<[u8]>>` for the `CompressedBytecode`.

MathieuDutSik:
## Motivation

We are using the `Vec<u8>` in the `CompressedBytecode`. But the vector is never modified so this makes
a copy that is not needed.

## Proposal

The compressed bytecode is now a `Arc<Box<[u8]>>`. An alternative could have been to use `BlobContent`.
However, we would need to track the `BlobType`, which would be hard.

## Test Plan

The CI.

## Release Plan

## Links

None.

### DIFF
--- a/linera-base/src/data_types.rs
+++ b/linera-base/src/data_types.rs
@@ -23,6 +23,7 @@ use async_graphql::{InputObject, SimpleObject};
 use custom_debug_derive::Debug;
 use linera_witty::{WitLoad, WitStore, WitType};
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use serde_with::{serde_as, Bytes};
 use thiserror::Error;
 
 #[cfg(with_metrics)]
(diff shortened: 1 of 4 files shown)


deuszx (commented):
When you create an instance of `CompressedBytes` it's the only copy that holds the compressed `Vec<u8>` . When we create an instance of service bytecode, the `CompressedBytecode` is consumed so there's no copy either.

MathieuDutSik:
> When you create an instance of `CompressedBytes`, it's the only copy that holds the compressed `Vec<u8>` . When we create an instance of service bytecode, the `CompressedBytecode` is consumed, so there's no copy either.

Examining statements by statements:
* "When you create an instance of `CompressedBytes`, it's the only copy that holds the compressed `Vec<u8>`".
  - Examining the code, we see that in `Storage::load_contract`, we have that if the blob is already in the list of created blobs by this transaction or previous transactions, then we have a `content.clone()` which is indeed there ok because an `Arc` is copied, which is almost free.
  - However, this is immediately followed by a `.into_vec_or_clone()`, which if the data is obtained from the first case of the `match` is a clone operation. Therefore, in that case, the `Vec<u8>` is in two places.
  - After the `CompressedBytecode` has been created, it is then decompressed.

* "When we create an instance of service bytecode, the `CompressedBytecode` is consumed, so there's no copy either"
  - If by "create an instance of service bytecode", you mean decompress it, then you are wrong as the API of the decompression is "fn decompress(&self)". Therefore, it is not being consumed.

Summarizing, the net effect of this PR is to avoid one clone operation of the compressed bytecode which is not needed.

linera-base/src/data_types.rs
```diff
@@ -1326,6 +1333,11 @@ impl BlobContent {
         bytes.into_vec()
     }
```
afck, line 1334: Do we still need `into_vec_or_clone`?

afck (approved):

linera-base/src/data_types.rs
MathieuDutSik: Yes, in `linera-execution/src/runtime.rs`.
This is the use case that will be harder to make right.

PR #4419, opened 2025-08-27
Update the way that keys are build in views obtained by `linera-views-derive`.

### LINKED ISSUE #2391: Use smaller keys in views used for views obtained by `linera-views-derive`
In a view such as
```
struct MyView {
  a: RegisterView<u32, C>,
  b: MapView<u32, u32, C>,
}
```

the indices are computed as serialization of an `usize` integer. This leads to the use of 8 bytes while just 1 suffices.
The solution is to have better macros. This does not require limiting oneself to structs having at most 256 entries.
On the negative, this likely yields only very limited improvements.

MathieuDutSik:
## Motivation

It has recently been indicated in PR #4347 that the use `ChainId` in a message would be a waste of space.
But there are other wastage of key space that can be reduced.

Fixes #2391

## Proposal

We check for the number of entries in views obtained by `linera-views-derive`. If lower than 256, we convert to `u8`.
It is likely that we need to remove as well the `linera_views::views::MIN_VIEW_TAG`.

## Test Plan

CI.

## Release Plan

It is definitely breaking the testnet.

## Links

None

### DIFF
--- a/linera-views-derive/src/lib.rs
+++ b/linera-views-derive/src/lib.rs
@@ -95,10 +95,10 @@ fn generate_view_code(input: ItemStruct, root: bool) -> TokenStream2 {
         num_init_keys_quotes.push(quote! { #g :: NUM_INIT_KEYS });
 
         let derive_key_logic = if num_fields < 256 {
-            let idx_byte = idx as u8;
+            let idx_u8 = idx as u8;
             quote! {
-                let __linera_reserved_index_byte = #idx_byte;
-                let __linera_reserved_base_key = context.base_key().derive_tag_key(linera_views::views::MIN_VIEW_TAG, &__linera_reserved_index_byte)?;
+                let __linera_reserved_index = #idx_u8;
+                let __linera_reserved_base_key = context.base_key().derive_tag_key(linera_views::views::MIN_VIEW_TAG, &__linera_reserved_index)?;
             }
         } else {
             assert!(num_fields < 65536);
--- a/linera-views-derive/src/lib.rs
+++ b/linera-views-derive/src/lib.rs
@@ -77,10 +77,10 @@ fn generate_view_code(input: ItemStruct, root: bool) -> TokenStream2 {
     let mut num_init_keys_quotes = Vec::new();
     let mut pre_load_keys_quotes = Vec::new();
     let mut post_load_keys_quotes = Vec::new();
+    let num_fields = input.fields.len();
     for (idx, e) in input.fields.iter().enumerate() {
         let name = e.ident.clone().unwrap();
         let test_flush_ident = format_ident!("deleted{}", idx);
-        let idx_lit = syn::LitInt::new(&idx.to_string(), Span::call_site());
         let g = get_extended_entry(e.ty.clone());
         name_quotes.push(quote! { #name });
         rollback_quotes.push(quote! { self.#name.rollback(); });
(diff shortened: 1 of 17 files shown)


linera-views-derive/src/lib.rs
```diff
@@ -98,25 +97,27 @@ fn generate_view_code(input: ItemStruct, root: bool) -> TokenStream2 {
         let derive_key_logic = if num_fields < 256 {
             let idx_byte = idx as u8;
             quote! {
-                let index_byte = #idx_byte;
-                let base_key = context.base_key().derive_tag_key(linera_views::views::MIN_VIEW_TAG, &index_byte)?;
+                let __internal_index_byte = #idx_byte;
```
ma2bd: Thanks @MathieuDutSik for the PR. If we want to be super paranoid, perhaps `__linera_reserved_xxx` is even more clear.

linera-views-derive/src/snapshots/linera_views_derive__tests__test_generate_view_code_C.snap
```diff
@@ -23,60 +23,72 @@ where
...
     type Context = C;
     fn context(&self) -> &C {
-        use linera_views::views::View;
+        use linera_views::views::View as _;
         self.register.context()
     }
     fn pre_load(context: &C) -> Result<Vec<Vec<u8>>, linera_views::ViewError> {
         use linera_views::context::Context as _;
-        let mut keys = Vec::new();
-        let index = 0;
```
ma2bd: Geez, what was even the type of `index` here? `derive_tag_key` is generic so it could be anything afaict.

linera-views-derive/src/lib.rs
```diff
@@ -93,17 +93,31 @@ fn generate_view_code(input: ItemStruct, root: bool) -> TokenStream2 {
             }
         });
         num_init_keys_quotes.push(quote! { #g :: NUM_INIT_KEYS });
+
+        let derive_key_logic = if num_fields < 256 {
+            let idx_byte = idx as u8;
```
ma2bd: This part where we inspect `num_fields` to optimize key sizes is a breaking change that is going beyond the initial bug report of #4450 (EDIT: and indeed it's addressing #2391 as you noted).

We want to release a fix to #4450 on testnet branch, so can you make two PRs? (starting with the fix)

linera-views-derive/src/snapshots/linera_views_derive__tests__test_generate_view_code_C.snap
ma2bd: Ok it looks like it was the default Rust integer type `i32`. At least it's not `usize`.

ma2bd: See previous comment on separating the PRs.

MathieuDutSik: Yes, correct, the keys being generated have size 4.

linera-views-derive/src/lib.rs
MathieuDutSik: Ok, done in another PR.

MathieuDutSik:
> As said above, this should be two PRs so that we can update the SDK 0.15.

Created another PR https://github.com/linera-io/linera-protocol/pull/4452 which does the addition of the prefix `__linera_reserved_`.

linera-views-derive/src/lib.rs
```diff
@@ -93,15 +94,27 @@ fn generate_view_code(input: ItemStruct, root: bool) -> TokenStream2 {
...
+            quote! {
+                let index_byte = #idx_byte;
+                let base_key = context.base_key().derive_tag_key(linera_views::views::MIN_VIEW_TAG, &index_byte)?;
+            }
+        } else {
+            quote! {
+                let index = #idx_lit;
+                let base_key = context.base_key().derive_tag_key(linera_views::views::MIN_VIEW_TAG, &index)?;
```
ma2bd: Can we make sure that the integer type is (locally) visible when we use `derive_tag_key` on integers? Those types are part of the protocol and any change would be breaking.

linera-views-derive/src/snapshots/linera_views_derive__tests__test_generate_view_code_C.snap
```diff
@@ -55,22 +55,22 @@ where
     ) -> Result<Self, linera_views::ViewError> {
         use linera_views::context::Context as _;
         let mut pos = 0;
-        let index = 0;
-        let pos_next = pos + RegisterView::<C, usize>::NUM_INIT_KEYS;
+        let index_byte = 0u8;
```
ma2bd: nit: Here and in other places. We don't really need `_byte`. (We need `0u8`)

linera-views-derive/src/lib.rs
ma2bd: (ok it looks like this is maybe the last place where it's ambiguous)

MathieuDutSik: You are very right. I added a bound of 65536 on the number of entries.
So, the key is either `u8` or `u16`.

linera-views-derive/src/snapshots/linera_views_derive__tests__test_generate_view_code_C.snap
MathieuDutSik: I removed the "_byte" and added "_u8" / "_u16" when adequate.
Now the generated code does not contain the "_byte", but it does contain "0u8" / "0u16".

PR #4420, opened 2025-08-27
Fetch certificate last used the blob

deuszx:
## Motivation

Continuation of the work on improving performance of clients. Trying to decrease # of roundtrips between clients and validators.

## Proposal

It was noted that we often make two requests to download a certificate for blob:
1. `BlobLastUsedBy` – returning a certificate hash
2. `DownloadCertificate` – returning the certificate.

Here we introduce a new endpoint – `BlobLastUsedByCertificate(blob_id) -> ConfirmedCertifiate` – which does both things in one query.

Note that the new endpoint is introduced in proxy only.

This decreases number of necessary network roundtrips from 9 to 6:

Screenshot 2025-08-27 at 12 07 08

## Test Plan

CI

## Release Plan

- Nothing to do / These changes follow the usual release cycle. **OR**
    - be released in a new SDK,
    - be released in a new SDK,

## Links

### DIFF
--- a/linera-core/src/node.rs
+++ b/linera-core/src/node.rs
@@ -155,6 +155,12 @@ pub trait ValidatorNode {
     /// Returns the hash of the `Certificate` that last used a blob.
     async fn blob_last_used_by(&self, blob_id: BlobId) -> Result<CryptoHash, NodeError>;
 
+    /// Returns the certificate that last used the blob.
+    async fn blob_last_used_by_certificate(
+        &self,
+        blob_id: BlobId,
+    ) -> Result<ConfirmedBlockCertificate, NodeError>;
+
     /// Returns the missing `Blob`s by their IDs.
     async fn missing_blob_ids(&self, blob_ids: Vec<BlobId>) -> Result<Vec<BlobId>, NodeError>;
 }
--- a/linera-rpc/src/message.rs
+++ b/linera-rpc/src/message.rs
@@ -62,6 +62,9 @@ pub enum RpcMessage {
 
     // Internal to a validator
     CrossChainRequest(Box<CrossChainRequest>),
+
+    BlobLastUsedByCertificate(Box<BlobId>),
+    BlobLastUsedByCertificateResponse(Box<ConfirmedBlockCertificate>),
 }
 
 impl RpcMessage {
@@ -100,6 +103,8 @@ impl RpcMessage {
             | DownloadCertificates(_)
             | BlobLastUsedBy(_)
             | BlobLastUsedByResponse(_)
+            | BlobLastUsedByCertificate(_)
+            | BlobLastUsedByCertificateResponse(_)
             | MissingBlobIds(_)
             | MissingBlobIdsResponse(_)
             | DownloadCertificatesResponse(_) => {
@@ -122,6 +127,7 @@ impl RpcMessage {
             | DownloadBlob(_)
             | DownloadConfirmedBlock(_)
             | BlobLastUsedBy(_)
+            | BlobLastUsedByCertificate(_)
             | MissingBlobIds(_)
             | DownloadCertificates(_)
             | DownloadCertificatesByHeights(_, _) => true,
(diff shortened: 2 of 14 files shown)


linera-rpc/src/simple/client.rs
```diff
@@ -246,6 +246,20 @@ impl ValidatorNode for SimpleClient {
             .await
     }
 
+    async fn blob_last_used_by_certificate(
+        &self,
+        blob_id: BlobId,
+    ) -> Result<ConfirmedBlockCertificate, NodeError> {
+        Ok(self
+            .query::<Vec<ConfirmedBlockCertificate>>(RpcMessage::BlobLastUsedByCertificate(
```
afck: Shouldn't this be `query::<ConfirmedBlockCertificate>` then?

linera-rpc/src/message.rs
```diff
@@ -202,6 +209,7 @@ impl TryFrom<RpcMessage> for Vec<ConfirmedBlockCertificate> {
         match message {
             RpcMessage::DownloadCertificatesResponse(certificates) => Ok(certificates),
             RpcMessage::DownloadCertificatesByHeightsResponse(certificates) => Ok(certificates),
+            RpcMessage::BlobLastUsedByCertificateResponse(certificate) => Ok(vec![*certificate]),
```
afck: Why not put this in `impl TryFrom<RpcMessage> for ConfirmedBlock` instead?

linera-core/src/node.rs
```diff
@@ -155,6 +155,12 @@ pub trait ValidatorNode {
     /// Returns the hash of the `Certificate` that last used a blob.
     async fn blob_last_used_by(&self, blob_id: BlobId) -> Result<CryptoHash, NodeError>;
```
afck, line 156: Do we still need this one then?

linera-rpc/src/simple/client.rs
deuszx: If I add `impl TryFrom<RpcMessage> for ConfirmedBlockCertificate` then yes – I was probably being lazy.

linera-core/src/node.rs
deuszx: Now I am afraid of removing anything. the protobuf/gRPC protocol will survive these changes but the simple one, based on `RpcMessage`, won't so I prefer to not remove anything.

linera-rpc/src/message.rs
deuszx: Oh wow... I looked for it and didn't find 🙈

deuszx: because `ConfirmedBlock` and `ConfirmedBlockCertficiate` are differnet things and we need the latter.

afck: Sorry, I mean: Why not `impl TryFrom<RpcMessage> for ConfirmedBlockCertificate`? Why the `Vec`?

deuszx: B/c it didn't exist and I was too lazy to add it. I'll do that.

deuszx: Done in [Add new TryFrom<RpcMessage> for ConfirmedBlockCertificate](https://github.com/linera-io/linera-protocol/pull/4420/commits/6b92b0464b812fb8a568edb3fed9d73239e4a8f6)

linera-rpc/src/simple/client.rs
deuszx: [Add new TryFrom<RpcMessage> for ConfirmedBlockCertificate](https://github.com/linera-io/linera-protocol/pull/4420/commits/6b92b0464b812fb8a568edb3fed9d73239e4a8f6)

.github/workflows/rust.yml
```diff
@@ -229,7 +229,7 @@ jobs:
         restore-keys: solc-
     - name: Get Solc
       if: ${{ steps.cache-solc.outputs.cache-hit != 'true' }}
-      uses: pontem-network/get-solc@master
+      uses: ./.github/actions/get-solc
```
ma2bd: These changes in `.github` seem to belong to another PR.

deuszx: I will update the description, was testing this solution. 

Yes, the action we use for fetching solc is broken. I decided to move it into the repo with the fixes.

deuszx: Opened a separate PR just with this fix https://github.com/linera-io/linera-protocol/pull/4426 

Will rebase once it's merged to `main`.

PR #4421, opened 2025-08-27
Improvement to the `linera-views-derive` code

### LINKED ISSUE #744: linera-views-derive: Do not panic on tuple-structs
The following does not work:
```rust
pub struct StateView<C>(ReentrantCollectionView<C, u8, RegisterView<C, u32>>);
```
We have the unwrapping of an Option that is None and so panics the macro. Maybe, it is easy to resolve.

### LINKED ISSUE #2391: Use smaller keys in views used for views obtained by `linera-views-derive`

MathieuDutSik:
## Motivation

The code in `lib.rs` has several failure scenarios that result in crashes, panics, or other errors.

Fixes #744 

## Proposal

It is completely possible to implement views for unnamed structs, that is:
```rust
pub struct CounterState(pub RegisterView<u64>);
```

However, we choose to make this illegal.  The reason is that this kind of structs would not be used in the protocol code and it is also not allowed by `async_graphql::SimpleObject`.

What this PR does is to add several error patterns without panic to the macro code:
* Unnamed structs lead to explicit error.
* Track the error produced by `RegisterView<u64>` which can be produced by missing Context or by missing view macro.
* structs must have at least one entry.

## Test Plan

The CI.

## Release Plan

## Links

None.

### DIFF
--- a/linera-views-derive/src/lib.rs
+++ b/linera-views-derive/src/lib.rs
@@ -66,21 +66,17 @@ fn generate_view_code(input: ItemStruct, root: bool) -> Result<TokenStream2, Err
 
     let attrs: StructAttrs = deluxe::parse_attributes(&input)
         .map_err(|e| Error::new_spanned(&input, format!("Failed to parse attributes: {}", e)))?;
-    let context = match attrs.context {
-        Some(ctx) => ctx,
-        None => match input.generics.type_params().next() {
-            Some(param) => {
-                let ident = &param.ident;
-                parse_quote! { #ident }
-            }
-            None => {
-                return Err(Error::new_spanned(
-                        &input,
-                        "Missing context: either add a generic type parameter or specify the context with #[view(context = YourContextType)]"
-                    ));
-            }
-        },
-    };
+    let context = attrs.context.or_else(|| {
+        input.generics.type_params().next().map(|param| {
+            let ident = &param.ident;
+            parse_quote! { #ident }
+        })
+    }).ok_or_else(|| {
+        Error::new_spanned(
+            &input,
+            "Missing context: either add a generic type parameter or specify the context with #[view(context = YourContextType)]"
+        )
+    })?;
 
     let struct_name = &input.ident;
     let field_types: Vec<_> = input.fields.iter().map(|field| &field.ty).collect();
--- a/linera-views-derive/src/lib.rs
+++ b/linera-views-derive/src/lib.rs
@@ -6,7 +6,7 @@
 use proc_macro::TokenStream;
 use proc_macro2::{Span, TokenStream as TokenStream2};
 use quote::{format_ident, quote};
-use syn::{parse_macro_input, parse_quote, ItemStruct, Type};
+use syn::{parse_macro_input, parse_quote, Error, ItemStruct, Type};
 
 #[derive(Debug, deluxe::ParseAttributes)]
 #[deluxe(attributes(view))]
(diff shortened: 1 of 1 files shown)


linera-views-derive/src/lib.rs
```diff
@@ -37,33 +37,50 @@ impl<'a> Constraints<'a> {
...
+        Some(segment) => segment,
+        None => return Err(Error::new_spanned(&typepath.path, "Path has no segments")),
+    };
+    let ident = &path_segment.ident;
+    let arguments = &path_segment.arguments;
+    Ok(quote! { #ident :: #arguments })
 }
 
-fn generate_view_code(input: ItemStruct, root: bool) -> TokenStream2 {
+fn generate_view_code(input: ItemStruct, root: bool) -> Result<TokenStream2, Error> {
```
Twey, line 53: Love this!

ma2bd:
This looks great. Maybe @Twey can give the final `accept`

linera-views-derive/src/lib.rs
```diff
@@ -37,33 +37,50 @@ impl<'a> Constraints<'a> {
...
-            .expect("no `context` given and no type parameters")
-            .ident;
-        parse_quote! { #ident }
-    });
+    let attrs: StructAttrs = deluxe::parse_attributes(&input)
+        .map_err(|e| Error::new_spanned(&input, format!("Failed to parse attributes: {}", e)))?;
+    let context = match attrs.context {
+        Some(ctx) => ctx,
+        None => match input.generics.type_params().next() {
+            Some(param) => {
```
Twey: I think this could be more clearly expressed using [`or_else`](https://doc.rust-lang.org/std/option/enum.Option.html#method.or_else).

linera-views-derive/src/lib.rs
```diff
@@ -121,9 +138,15 @@ fn generate_view_code(input: ItemStruct, root: bool) -> TokenStream2 {
         });
     }
 
-    let first_name_quote = name_quotes
-        .first()
-        .expect("list of names should be non-empty");
+    let first_name_quote = match name_quotes.first() {
+        Some(name) => name,
+        None => {
```
Twey:
```suggestion
    let first_name_quote = name_quotes.first().ok_or(Error::new_spanned(…))?;
```

linera-views-derive/src/lib.rs
```diff
@@ -369,64 +406,131 @@ fn generate_clonable_view_code(input: ItemStruct) -> TokenStream2 {
...
 #[proc_macro_derive(HashableView, attributes(view))]
 pub fn derive_hash_view(input: TokenStream) -> TokenStream {
     let input = parse_macro_input!(input as ItemStruct);
-    let mut stream = generate_view_code(input.clone(), false);
-    stream.extend(generate_hash_view_code(input));
+
+    let mut stream = match generate_view_code(input.clone(), false) {
+        Ok(tokens) => tokens,
+        Err(err) => return err.to_compile_error().into(),
+    };
```
Twey:
```suggestion
    let mut stream = generate_view_code(input.clone(), false)
        .map_err(|err| err.to_compile_error().into())?;
```

linera-views-derive/src/lib.rs
```diff
@@ -369,64 +406,131 @@ fn generate_clonable_view_code(input: ItemStruct) -> TokenStream2 {
...
 pub fn derive_hashable_root_view(input: TokenStream) -> TokenStream {
     let input = parse_macro_input!(input as ItemStruct);
-    let mut stream = generate_view_code(input.clone(), true);
-    stream.extend(generate_root_view_code(input.clone()));
-    stream.extend(generate_hash_view_code(input));
+
+    let mut stream = match generate_view_code(input.clone(), true) {
+        Ok(tokens) => tokens,
+        Err(err) => return err.to_compile_error().into(),
+    };
```
Twey: Since this is repeated a lot, it would be nicer to have an implementation function that uses `?` and then convert it just once per top-level function.

linera-views-derive/src/lib.rs
MathieuDutSik: Thanks.

PR #4426, opened 2025-08-27
fix solc github action

deuszx:
## Motivation

The GH Action we use for downloading Solidity compiler (https://github.com/actions-marketplace-validations/pontem-network_get-solc) is broken.

## Proposal

Bring the action in and fix.

## Test Plan

CI

## Release Plan

- Nothing to do

## Links


MathieuDutSik (approved):
LGTM as it solves the problem.

PR #4432, opened 2025-08-28
Fix spurious error logs in `is_retryable`.

afck:
## Motivation

`is_retryable` was originally used only for gRPC _streams_, but is now used in other requests, too. The error-level logs in it are not appropriate anymore.

In particular, `NotFound` is expected in some cases. And in general, the logs should never be error-level, because they can be caused by a single faulty validator.

## Proposal

Fix logs and log levels.

## Test Plan

CI

## Release Plan

    - be released in a new SDK.

## Links

### DIFF
--- a/linera-rpc/src/grpc/client.rs
+++ b/linera-rpc/src/grpc/client.rs
@@ -78,7 +78,7 @@ impl GrpcClient {
                 info!("Unexpected gRPC status: {}; retrying", status);
                 true
             }
-            Code::NotFound => false,
+            Code::NotFound => false, // This code is used if e.g. the validator is missing blobs.
             Code::InvalidArgument
             | Code::AlreadyExists
             | Code::PermissionDenied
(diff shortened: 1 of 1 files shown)


linera-rpc/src/grpc/client.rs
```diff
@@ -75,11 +75,11 @@ impl GrpcClient {
                 true
             }
             Code::Ok | Code::Cancelled | Code::ResourceExhausted => {
-                error!("Unexpected gRPC status: {}; retrying", status);
+                info!("Unexpected gRPC status: {}; retrying", status);
                 true
             }
+            Code::NotFound => false,
```
MathieuDutSik: Why not indicate what this scenario means?

afck: Done in https://github.com/linera-io/linera-protocol/pull/4432/commits/fd4f2fa63d3510388748703c71ed0a50a6b2168f.

PR #4436, opened 2025-08-28
Two-phase chain listener

Twey:
## Motivation

Currently we don't synchronize the chains when starting up the client.  Now that we also don't pessimistically synchronize chains before performing operations, we sometimes run into the situation where we want to do something with a chain whose blocks haven't been seen yet (e.g. because it came from the faucet), which leads to an error.

## Proposal

Ideally when we find a block from the future we should synchronize the offending chain.  But for now it should be sufficient to make sure to synchronize the initial chains at startup.

Split `ChainListener::run` into two futures: the latter is the full chain listener loop, as before, but first there's an initialization step that synchronizes all the initial chains, and the caller can wait on this in serial to ensure the chains are synchronized before performing operations on the client.

## Test Plan

CI should check for regressions; it would be nice to add a test for this case, perhaps after testnet release.

## Release Plan

    - be released in a new SDK,

## Links

### DIFF
--- a/linera-client/src/benchmark.rs
+++ b/linera-client/src/benchmark.rs
@@ -130,14 +130,10 @@ impl<Env: Environment> Benchmark<Env> {
         let notifier = Arc::new(Notify::new());
         let barrier = Arc::new(Barrier::new(num_chains + 1));
 
-        let chain_listener_handle = tokio::spawn(
-            async move {
-                if let Err(e) = chain_listener.run().await {
-                    warn!("Chain listener error: {}", e);
-                }
-            }
-            .instrument(tracing::info_span!("chain_listener")),
-        );
+        let chain_listener_result = chain_listener.run().await;
+
+        let chain_listener_handle =
+            tokio::spawn(async move { chain_listener_result?.await }.in_current_span());
 
         let bps_control_task = Self::bps_control_task(
             &barrier,
@@ -216,7 +212,10 @@ impl<Env: Environment> Benchmark<Env> {
         if let Some(runtime_control_task) = runtime_control_task {
             runtime_control_task.await?;
         }
-        chain_listener_handle.await?;
+
+        if let Err(e) = chain_listener_handle.await? {
+            tracing::error!("chain listener error: {e}");
+        }
 
         Ok(())
     }
--- a/linera-client/src/chain_listener.rs
+++ b/linera-client/src/chain_listener.rs
@@ -10,7 +10,7 @@ use std::{
 use futures::{
     future::{join_all, select_all},
     lock::Mutex,
-    FutureExt as _, StreamExt,
+    Future, FutureExt as _, StreamExt,
 };
 use linera_base::{
     crypto::{CryptoHash, Signer},
@@ -202,25 +202,27 @@ impl<C: ClientContext> ChainListener<C> {
 
     /// Runs the chain listener.
     #[instrument(skip(self))]
-    pub async fn run(mut self) -> Result<(), Error> {
+    pub async fn run(mut self) -> Result<impl Future<Output = Result<(), Error>>, Error> {
         let chain_ids = {
             let guard = self.context.lock().await;
             let mut chain_ids = BTreeSet::from_iter(guard.wallet().chain_ids());
             chain_ids.insert(guard.wallet().genesis_admin_chain());
             chain_ids
         };
-        self.listen_recursively(chain_ids).await?;
-        loop {
-            match self.next_action().await? {
-                Action::ProcessInbox(chain_id) => self.maybe_process_inbox(chain_id).await?,
-                Action::Notification(notification) => {
-                    self.process_notification(notification).await?
+        Ok(async {
+            self.listen_recursively(chain_ids).await?;
+            loop {
+                match self.next_action().await? {
+                    Action::ProcessInbox(chain_id) => self.maybe_process_inbox(chain_id).await?,
+                    Action::Notification(notification) => {
+                        self.process_notification(notification).await?
+                    }
+                    Action::Stop => break,
                 }
-                Action::Stop => break,
             }
-        }
-        join_all(self.listening.into_values().map(|client| client.stop())).await;
-        Ok(())
+            join_all(self.listening.into_values().map(|client| client.stop())).await;
+            Ok(())
+        })
     }
 
     /// Processes a notification, updating local chains and validators as needed.
--- a/linera-client/src/chain_listener.rs
+++ b/linera-client/src/chain_listener.rs
@@ -205,10 +205,20 @@ impl<C: ClientContext> ChainListener<C> {
     pub async fn run(mut self) -> Result<impl Future<Output = Result<(), Error>>, Error> {
         let chain_ids = {
             let guard = self.context.lock().await;
-            let mut chain_ids = BTreeSet::from_iter(guard.wallet().chain_ids());
-            chain_ids.insert(guard.wallet().genesis_admin_chain());
-            chain_ids
+            let admin_chain_id = guard.wallet().genesis_admin_chain();
+            guard
+                .make_chain_client(admin_chain_id)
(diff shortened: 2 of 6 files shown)


linera-client/src/chain_listener.rs
```diff
@@ -314,9 +318,27 @@ impl<C: ClientContext> ChainListener<C> {
...
         }
+
+        futures::future::try_join_all(all_chain_ids.iter().copied().map(|chain_id| {
+            let this = &*self;
+            async move {
+                this.listening
+                    .get(&chain_id)
+                    .expect("listen() should have inserted this above")
+                    .client
+                    .synchronize_from_validators()
```
afck: Don't we already synchronize in `ChainClient::listen`?

I think we should simply make sure that the admin chain's `listen` is started first, and gets past the point of initial synchronization, before we `listen` to the other chains in parallel.

Twey: I don't think we do any more, no.  We synchronize only when processing the inbox now.

afck: We do, but _inside_ the stream: https://github.com/linera-io/linera-protocol/blob/2a5c3984098fbc57ae2e704a1b08afe5dedb4ac0/linera-core/src/client/mod.rs#L3770

We'd have to take this out of the stream so we know when the syncing is done. :confused:

Twey: be20d7bf14867ae0911404cd03e6df96ec030095

Twey: Ah I was looking for `synchronize_from_validators` :) Nevertheless I think https://github.com/linera-io/linera-protocol/commit/be20d7bf14867ae0911404cd03e6df96ec030095 obviates it?

Now to track down this recursion depth issue that doesn't occur locally…

linera-client/src/chain_listener.rs
```diff
@@ -202,25 +202,31 @@ impl<C: ClientContext> ChainListener<C> {
...
-    pub async fn run(mut self) -> Result<(), Error> {
+    pub async fn run(mut self) -> Result<impl Future<Output = Result<(), Error>>, Error> {
+        let admin_chain_id;
         let chain_ids = {
             let guard = self.context.lock().await;
-            let mut chain_ids = BTreeSet::from_iter(guard.wallet().chain_ids());
-            chain_ids.insert(guard.wallet().genesis_admin_chain());
-            chain_ids
+            admin_chain_id = guard.wallet().genesis_admin_chain();
+            BTreeSet::from_iter(guard.wallet().chain_ids())
```
deuszx: We could return a tuple here and assign to `let (chain_ids, admin_chain_id) = ` instead of this.

linera-client/src/chain_listener.rs
```diff
@@ -202,25 +202,31 @@ impl<C: ClientContext> ChainListener<C> {
...
         };
-        self.listen_recursively(chain_ids).await?;
-        loop {
-            match self.next_action().await? {
-                Action::ProcessInbox(chain_id) => self.maybe_process_inbox(chain_id).await?,
-                Action::Notification(notification) => {
-                    self.process_notification(notification).await?
+
+        self.listen_recursively(std::iter::once(admin_chain_id).collect())
+            .await?;
```
deuszx: This doesn't guarantee we update the admin chain:

if you follow the code, it calls `process_inbox_without_prepare` which doesn't sync the chain. It does call `collect_epoch_changes` but that checks _local_ node only. Also, since there is no incoming message on the admin chain – we don't process anything. It'd be surprised if we ever did anyway - we're not the owner on that chain.

linera-client/src/chain_listener.rs
Twey: We could! I prefer names to remembering which position is which, though, especially as the block gets larger.

linera-client/src/chain_listener.rs
Twey: Right! I did add an explicit synchronize in `listen_recursively` but then I removed it. Hopefully that's enough to fix the problem — let me just synchronize here explicitly.

linera-client/src/chain_listener.rs
deuszx: the dangling, uninitialized `let admin_chain_id` is a torn in the eye :)

Twey: Thankfully in Rust (unlike some languages :eyes:) it's perfectly safe!

Twey: (removed anyway — I moved the synchronization inside this block since in needs the guard now)

linera-client/src/chain_listener.rs
Twey: Hopefully fixed in 0b7c63d43be44e1861e3a26669f723e2b4071d7b.

linera-client/src/benchmark.rs
```diff
@@ -130,14 +130,10 @@ impl<Env: Environment> Benchmark<Env> {
...
 
-        let chain_listener_handle = tokio::spawn(
-            async move {
-                if let Err(e) = chain_listener.run().await {
-                    warn!("Chain listener error: {}", e);
-                }
-            }
-            .instrument(tracing::info_span!("chain_listener")),
-        );
+        let chain_listener_result = chain_listener.run().await;
```
afck, line 133: Why not `?` here? If it failed to start, we probably want to exit?

linera-client/src/chain_listener.rs
```diff
@@ -202,25 +202,37 @@ impl<C: ClientContext> ChainListener<C> {
...
         let chain_ids = {
             let guard = self.context.lock().await;
-            let mut chain_ids = BTreeSet::from_iter(guard.wallet().chain_ids());
-            chain_ids.insert(guard.wallet().genesis_admin_chain());
-            chain_ids
+            let admin_chain_id = guard.wallet().genesis_admin_chain();
+            guard
+                .make_chain_client(admin_chain_id)
+                .synchronize_from_validators()
+                .await?;
```
afck, line 212: I'd rather call `self.listen_recursively([admin_chain_id])`, and make sure that, too, first syncs, and then returns a future: That way we don't sync twice, redundantly.

afck (approved):
No blockers from my side!

linera-client/src/chain_listener.rs
deuszx: Maybe we should think about what functions syncs what – b/c `listen_recursively` does a lot of different things that _may_ lead to duplicating the work of the `synchronize_from_validators`.

afck: I think it's correct that `ChainClient::listen` synchronizes at the beginning, so we can't miss any notifications. That's why ideally we wouldn't have to call `sychronize_from_validators` again here.

linera-client/src/benchmark.rs
Twey: I'd have to refactor the error types to do that, so I opted to maintain the current behaviour — maybe @ndr-ds can explain why the chain listener failing isn't fatal for the benchmarks.

linera-client/src/chain_listener.rs
Twey: @deuszx suggests refactoring `synchronize_from_validators` since we don't need to block on _all_ its functionality here as well.  But I'm marking that down as a future optimization.

Twey: Also in this category @deuszx notes that this single extra synchronization costs us about 30 round-trips at startup.

linera-client/src/benchmark.rs
ndr-ds: It currently isn't but it honestly probably should be 🤔 we should ensure it's running correctly to ensure the benchmarks are being run in a "realistic" way. I can do a follow up PR fixing it

ndr-ds: https://github.com/linera-io/linera-protocol/pull/4449

PR #4437, opened 2025-08-28
Extend reconfig test; sync on `wallet init`; fix blob fetching.

afck:
## Motivation

We want to make sure the network works fine even after many reconfigurations, and even if none of the genesis validators are left.

## Proposal

At the end of the reconfiguration end-to-end test, initialize a new client, request a chain and publish a blob. This test actually failed! The following two changes make it pass:

* In `wallet init`, synchronize the admin chain, trusting the validators provided by the faucet (part of https://github.com/linera-io/linera-protocol/issues/4434).
* Fix a bug in `process_certificates`: Since we are handling a _certificate_, not a block proposal, we don't need additional proof for the missing blobs: The certificate itself is proof for them. So we should only download the blobs themselves, and not try to get another certificate for them.

## Test Plan

The reconfiguration test was extended.

## Release Plan

    - be released in a new SDK.

## Links

- _Partly_ addresses https://github.com/linera-io/linera-protocol/issues/4434.

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -376,18 +376,11 @@ impl<Env: Environment> Client<Env> {
             let mut result = self.handle_certificate(certificate.clone()).await;
 
             if let Err(LocalNodeError::BlobsNotFound(blob_ids)) = &result {
-                future::try_join_all(blob_ids.iter().map(|blob_id| async move {
-                    let blob_certificate =
-                        remote_node.download_certificate_for_blob(*blob_id).await?;
-                    self.receive_sender_certificate(
-                        blob_certificate,
-                        ReceiveCertificateMode::NeedsCheck,
-                        None,
-                    )
-                    .await?;
-                    Result::<(), ChainClientError>::Ok(())
+                let blobs = future::join_all(blob_ids.iter().map(|blob_id| async move {
+                    remote_node.try_download_blob(*blob_id).await.unwrap()
                 }))
-                .await?;
+                .await;
+                self.local_node.store_blobs(&blobs).await?;
                 result = self.handle_certificate(certificate.clone()).await;
             }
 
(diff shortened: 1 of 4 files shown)


linera-core/src/client/mod.rs
```diff
@@ -2297,6 +2302,18 @@ impl<Env: Environment> ChainClient<Env> {
         self.client.synchronize_chain_state(chain_id).await
     }
 
+    /// Downloads and processes any certificates we are missing for the given chain, from the given
```
deuszx:
```suggestion
    /// Downloads and processes any certificates we are missing for the current chain, from the given
```

linera-core/src/client/mod.rs
```diff
@@ -376,18 +376,11 @@ impl<Env: Environment> Client<Env> {
...
-                        remote_node.download_certificate_for_blob(*blob_id).await?;
-                    self.receive_sender_certificate(
-                        blob_certificate,
-                        ReceiveCertificateMode::NeedsCheck,
-                        None,
-                    )
-                    .await?;
-                    Result::<(), ChainClientError>::Ok(())
+                let blobs = future::join_all(blob_ids.iter().map(|blob_id| async move {
+                    remote_node.try_download_blob(*blob_id).await.unwrap()
```
deuszx, line 380: Right, b/c the blobs in `blob_ids` can only come from the `certificate` (otherwise it would have been the "unexpected blob" error)?

linera-service/src/cli/main.rs
```diff
@@ -1610,6 +1610,43 @@ impl Runnable for Job {
...
+                                account_public_key: AccountSecretKey::generate().public(),
+                            };
+                            (pub_key, state)
+                        })
+                        .collect(),
+                    Default::default(), // unused
+                );
+                chain_client
+                    .synchronize_chain_state_from_committee(committee)
+                    .await?;
```
deuszx, line 1647: This is for the CLI only, what about other clients? Are they all supposed to do that manually?

afck: Yes, also on initialization.

afck: (But this is still a hack; needs a proper solution, but I hope this can already unblock us.)

PR #4446, opened 2025-08-29
Bump versions of `tonic` dependencies.

deuszx:
## Motivation

Originally the motivation was to use latest `tonic-web-wasm-client` (version `0.8.0`) which introduced the timeout on web clients using JS `fetch` method (https://github.com/devashishdxt/tonic-web-wasm-client/pull/84). In order to do so, update of all tonic dependencies was required.

## Proposal

Update `tonic`.

Unfortunately, we don't use the `tonic-web-wasm-client` latest feature as it also cancels the long-living streaming requests (like our `Subscribe` endpoint) which triggers the re-subscription (repeatedly). This slows down the clients but also could lead to potentially loosing some events emitted while it was un-subscribed.

## Test Plan

CI.

## Release Plan

- This change follows usual release cycle.

## Links

### DIFF
--- a/linera-indexer/lib/Cargo.toml
+++ b/linera-indexer/lib/Cargo.toml
@@ -46,7 +46,7 @@ sqlx = { version = "0.8", features = ["runtime-tokio-rustls", "sqlite"] }
 thiserror.workspace = true
 tokio = { workspace = true, features = ["rt-multi-thread", "macros"] }
 tokio-stream = "0.1"
-tonic = { workspace = true, features = ["server", "codegen"] }
+tonic = { workspace = true, features = ["router", "server", "codegen"] }
 tonic-prost.workspace = true
 tower-http = { workspace = true, features = ["cors"] }
 tracing.workspace = true


linera-rpc/src/grpc/transport.rs
```diff
@@ -20,11 +20,16 @@ impl From<&'_ NodeOptions> for Options {
...
-            // TODO(#1817): this should respect `options`
-            Ok(tonic_web_wasm_client::Client::new(address))
+        pub fn create_channel(address: String, options: &Options) -> Result<Channel, Error> {
+            if let Some(timeout) = options.timeout {
+                let mut options = FetchOptions::new();
+                options.timeout(timeout);
+                Ok(tonic_web_wasm_client::Client::new_with_options(address, options))
+            } else {
+                Ok(tonic_web_wasm_client::Client::new(address))
+            }
```
Twey:
```suggestion
            Ok(tonic_web_wasm_client::Client::new_with_options(
                address,
                options.timeout
                    .map(|timeout| FetchOptions::new().timeout(timeout))
                    .unwrap_or_default(),
            ))
```

deuszx: Yeah, I didn't even want to construct `FetchOptions` if there's no timeout set.

Twey: I'm not sure that's the kind of optimization I'd write extra code for without some heavy profiling data :sweat_smile: but nevertheless:

```suggestion
            let mut client = tonic_web_wasm_client::Client::new(address);
            if let Some(timeout) = options.timeout {
                client.with_options(FetchOptions::new().timeout(timeout));
            }
            Ok(client)
```

deuszx: Convinced :)

linera-rpc/src/grpc/transport.rs
```diff
@@ -20,9 +20,9 @@ impl From<&'_ NodeOptions> for Options {
 
 cfg_if::cfg_if! {
     if #[cfg(web)] {
-        pub use tonic_web_wasm_client::{Client as Channel, Error};
+        pub use tonic_web_wasm_client::{Client as Channel, Error, options::FetchOptions};
 
-        pub fn create_channel(address: String, _options: &Options) -> Result<Channel, Error> {
+        pub fn create_channel(address: String, options: &Options) -> Result<Channel, Error> {
```
Twey: Won't this warn?

deuszx: 🤦 

I had a githook and silenced it b/c I wanted to push it quickly

linera-rpc/src/grpc/transport.rs
```diff
@@ -20,7 +20,7 @@ impl From<&'_ NodeOptions> for Options {
 
 cfg_if::cfg_if! {
     if #[cfg(web)] {
-        pub use tonic_web_wasm_client::{Client as Channel, Error};
+        pub use tonic_web_wasm_client::{Client as Channel, Error, options::FetchOptions};
```
Twey: This is unused as well

deuszx: Done.

PR #4447, opened 2025-08-29
`devnet_2025_08_21`: backport #4436 (Two-phase chain listener)

Twey:
Backport #4436.


eldios:
Close in favor of cherry-picking

PR #4453, opened 2025-09-01
Fix `Address20` debug format.

afck:
## Motivation

`AccountOwner::Address20`'s `Debug` impl prints something like:

```
Address20([10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10])
```

because the `hex_debug` attribute is in the wrong place.

## Proposal

Put the attribute directly on the field, not the enum variant. The `Debug` impl now says:

```
Address20(0a0a0a0a0a0a0a0a..)
```

## Test Plan

Tried it locally.

## Release Plan

- Should be backported to testnet and devnet.

## Links

### DIFF
--- a/linera-base/src/identifiers.rs
+++ b/linera-base/src/identifiers.rs
@@ -30,7 +30,7 @@ use crate::{
 };
 
 /// An account owner.
-#[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd, WitLoad, WitStore, WitType)]
+#[derive(Clone, Copy, Eq, Hash, Ord, PartialEq, PartialOrd, WitLoad, WitStore, WitType)]
 #[cfg_attr(with_testing, derive(test_strategy::Arbitrary))]
 pub enum AccountOwner {
     /// Short addresses reserved for the protocol.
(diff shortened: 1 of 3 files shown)


linera-base/src/identifiers.rs
```diff
@@ -38,8 +38,7 @@ pub enum AccountOwner {
     /// 32-byte account address.
     Address32(CryptoHash),
     /// 20-byte account EVM-compatible address.
-    #[debug(with = "hex_debug")]
-    Address20([u8; 20]),
+    Address20(#[debug(with = "hex_debug")] [u8; 20]),
```
MathieuDutSik: I would add a test for `Address20` as well as `Address32` (which I assume is correct right now, just unsure because of a lack of test).

afck: Good idea! Turns out:
* For `Address32` we only print 8 bytes because of the `Debug` impl for `CryptoHash`. I guess I should change that, too, and print the whole hash.
* For `Reserved`, we print it in decimal. I guess that's okay, although I'd prefer hexadecimal there, too.

MathieuDutSik: For `Reserved`, we currently have only 0 as a possibility.
This corresponds to `CHAIN`, so I guess that we can do something here as well.

afck: Done in https://github.com/linera-io/linera-protocol/pull/4453/commits/c0f846965167a0d1cfff35217ca28460eaaa5960; but now I'm a bit worried that this will make other log lines _more_ spammy, because now all hashes are four times as long.

(I'll leave changing the representation of `Reserved(0)` for a later PR; not sure about that.)

MathieuDutSik (approved):

linera-base/src/identifiers.rs
afck: As discussed, let's print only 8 bytes for both variants instead.

afck: Done in https://github.com/linera-io/linera-protocol/pull/4453/commits/e0abfd04f29e6986313b9add6f246434721453c1.

Issue #4422, opened 2025-08-27
`test_memory_skipping_proposal::wasmer` is flaky.


afck:
Example: https://github.com/linera-io/linera-protocol/actions/runs/17254879458/job/48964674297?pr=4415

```
ERROR linera_execution::wasm::runtime_api: panicked at hex-game/src/contract.rs:127:9:
assertion `left == right` failed: Victory can only be claimed by the player whose turn it is not.
  left: Some(Address32(ffa6ce7684115bb6))
 right: Some(Address32(8558d631bc35d0c9))
```

afck:
More recent example, with more logging: https://github.com/linera-io/linera-protocol/actions/runs/17329885156/job/49203202252?pr=4446#step:6:414

PR #4455, opened 2025-09-01
Make `fault_type` part of `LocalValidatorClient`.

afck:
## Motivation

The `linera-core` unit tests don't run a whole network, and instead wrap a worker in a mutex inside a clonable `LocalValidatorClient` struct. To simulate faults, a `FaultType` enum value is stored alongside the worker inside the mutex.

Since `ChainClient` sometimes cancels tasks that wait for a response from the validator, and we don't want to cancel the actual worker code in random places, we `spawn` tasks with the calls to the worker. (Maybe with the chain state actors that isn't necessary anymore? But that's a bigger discussion, and such a change could cause new flakiness if we're missing something.) These spawned tasks have a handle on the mutex with the worker state and the fault type.

That means the spawned tasks can outlive calls to `ChainClient` functions in the tests. A `set_fault_type` call after such a function has returned can still affect how the spawned tasks behave.

I strongly suspect this is what causes the very rare `test_memory_skipping_proposal` failure in CI, where we observe that a validator that should have had fault type `DontProcessValidated` ends up processing a validated block certificate anyway, because _after_ the client call that had spawned that task, we call `set_fault_type(… Honest)`.

## Proposal

Put the fault type in the `LocalValidatorClient`, so it is copied before the task is spawned.

## Test Plan

CI

I will close https://github.com/linera-io/linera-protocol/issues/4422; let's reopen it if it happens again anyway.

## Release Plan

- These changes should be backported to the testnet and devnet.

## Links

- Fixes https://github.com/linera-io/linera-protocol/issues/4422.

### DIFF
--- a/linera-core/src/unit_tests/test_utils.rs
+++ b/linera-core/src/unit_tests/test_utils.rs
@@ -55,7 +55,7 @@ use crate::{
         ValidatorNodeProvider,
     },
     notifier::ChannelNotifier,
-    worker::{NetworkActions, Notification, ProcessableCertificate, WorkerState},
+    worker::{Notification, ProcessableCertificate, WorkerState},
 };
 
 #[derive(Debug, PartialEq, Clone, Copy)]
@@ -80,7 +80,6 @@ where
     S: Storage,
 {
     state: WorkerState<S>,
-    fault_type: FaultType,
     notifier: Arc<ChannelNotifier<Notification>>,
 }
 
@@ -91,6 +90,7 @@ where
 {
     public_key: ValidatorPublicKey,
     client: Arc<Mutex<LocalValidator<S>>>,
+    fault_type: FaultType,
 }
 
 impl<S> ValidatorNode for LocalValidatorClient<S>
@@ -280,13 +280,13 @@ where
 {
     fn new(public_key: ValidatorPublicKey, state: WorkerState<S>) -> Self {
         let client = LocalValidator {
-            fault_type: FaultType::Honest,
             state,
             notifier: Arc::new(ChannelNotifier::default()),
         };
         Self {
             public_key,
             client: Arc::new(Mutex::new(client)),
+            fault_type: FaultType::Honest,
         }
     }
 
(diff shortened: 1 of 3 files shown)


linera-core/src/unit_tests/test_utils.rs
```diff
@@ -335,69 +335,34 @@ where
...
         sender: oneshot::Sender<Result<ChainInfoResponse, NodeError>>,
     ) -> Result<(), Result<ChainInfoResponse, NodeError>> {
-        let mut validator = self.client.lock().await;
-        let handle_block_proposal_result =
-            Self::handle_block_proposal(proposal, &mut validator).await;
-        let result = match handle_block_proposal_result {
-            Some(Err(NodeError::BlobsNotFound(_))) => {
-                handle_block_proposal_result.expect("handle_block_proposal_result should be Some")
-            }
-            _ => match validator.fault_type {
```
deuszx, line 345: I don't understand why you couldn't do the same thing w/o moving the `FaultType` to the new type? If we moved the match on `validator.fault_type` to the beginning of this function (and the other one), and only call to `handle_block_proposal` for appropriate "faults" – wouldn't that fix the problem as well?

I am not sure a *client is a good place to store the validator's fault.

afck: Since this is already inside the spawned task I think that wouldn't be enough.

Maybe `LocalValidatorClient` isn't a great name anyway: It represents the handle of a connection to the validator, on the client side. But in these tests, since we don't actually have a network, that distinction isn't clear anyway. And I'd say it doesn't matter whether we store the fault type inside that handle or inside `LocalValidator`.

afck:
> we don't know whether we're testing the test-only validator code or production one

We do: In the `client_tests` and `wasm_client_tests` modules we always use the test validators, since these are meant to test the `Client` and `ChainClient` implementations, not the validator implementations.

deuszx:
> > we don't know whether we're testing the test-only validator code or production one
> 
> We do: In the `client_tests` and `wasm_client_tests` modules we always use the test validators, since these are meant to test the `Client` and `ChainClient` implementations, not the validator implementations.

Here's we're using `LocalValidatorCLient` which has its own implementation of `ValidatorNode` with a custom logic. So we're not testing production validator.

afck:
Yes, that's what I'm saying: We do _know_ what we're testing. We're testing the _client_ implementation, using _test_ validators.

PR #4457, opened 2025-09-01
Add `currentCommittee` to the faucet; use it in `wallet init`.

afck:
## Motivation

In https://github.com/linera-io/linera-protocol/pull/4437 I used the existing `currentValidators` faucet query in `wallet init` and then made the client synchronize the initial admin chain from a committee where all these validators have the same weight.

## Proposal

Add a `currentCommittee` query that returns the actual current committee, and use that to synchronize.

## Test Plan

The reconfiguration test exercises this.

## Release Plan

- These changes should be backported to devnet and testnet, published in an SDK and deployed in the faucet.

## Links

- Closes https://github.com/linera-io/linera-protocol/issues/4434.

### DIFF
--- a/linera-faucet/client/src/lib.rs
+++ b/linera-faucet/client/src/lib.rs
@@ -5,8 +5,14 @@
 
 // TODO(#3362): generate this code
 
-use linera_base::{crypto::ValidatorPublicKey, data_types::ChainDescription};
+use std::collections::{BTreeMap, BTreeSet};
+
+use linera_base::{
+    crypto::ValidatorPublicKey,
+    data_types::{Amount, ChainDescription},
+};
 use linera_client::config::GenesisConfig;
+use linera_execution::{committee::ValidatorState, Committee, ResourceControlPolicy};
 use linera_version::VersionInfo;
 use thiserror_context::Context;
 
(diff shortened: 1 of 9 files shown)


linera-faucet/client/src/lib.rs
```diff
@@ -137,4 +143,140 @@ impl Faucet {
...
+            maximum_http_response_bytes: u64,
+            http_request_timeout_ms: u64,
+            http_request_allow_list: BTreeSet<String>,
+        }
+
+        #[derive(serde::Deserialize)]
+        #[serde(rename_all = "camelCase")]
+        struct CommitteeResponse {
+            validators: BTreeMap<ValidatorPublicKey, ValidatorState>,
+            policy: PolicyResponse,
```
deuszx: Do we need the `policy` or is it just for completeness?

afck: It's just for completeness.

In general I think it's surprising if a field is just missing in GraphQL, and it's dangerous to construct "fake committees" anywhere in the code.

linera-faucet/client/src/lib.rs
```diff
@@ -137,4 +143,140 @@ impl Faucet {
...
+            maximum_blob_size: u64,
+            maximum_published_blobs: u64,
+            maximum_block_proposal_size: u64,
+            maximum_bytes_read_per_block: u64,
+            maximum_bytes_written_per_block: u64,
+            maximum_oracle_response_bytes: u64,
+            maximum_http_response_bytes: u64,
+            http_request_timeout_ms: u64,
+            http_request_allow_list: BTreeSet<String>,
+        }
```
deuszx: Can't we use `ResourceControlPolicy` here? It derives `Deserialize`

afck: `linera-faucet-client` is implemented in a way that deserializes the GraphQL responses as JSON. If we used the original `ResourceControlPolicy` the fields would be snake case. Or, if we changed them to camel case, they would be camel case for other serialization formats, too.

afck: Discussed this in the meeting: I will make at least the resource policy a scalar instead.

afck: Done in https://github.com/linera-io/linera-protocol/pull/4457/commits/581ec6392519deeb100937b758a6f9f62ff04620.

PR #4458, opened 2025-09-01
Add `epoch` to the `UserChain` and `linera wallet show`

### LINKED ISSUE #2958: Include epoch in `wallet show`.
It is important for users, and also for us when debugging, to see which epoch their chains are in. And the epoch number is usually small, so it doesn't take much screen space.

Let's include it in `wallet show`.
(It could even be shown in the left column, so we don't have to add more rows to the output?)

MathieuDutSik:
## Motivation

The `linera wallet show` presents the data on the chain being tracked. Other information can be added easily.

Fixes #2958 

## Proposal

The `epoch` is always present in the output of `chain_info_query`. So, it can be added to the `UserChain`
easily.

Possible extension:
* Adding if a chain is closed or not (it suffices to add a boolean to `ChainInfo`).

## Test Plan

The readme test case fungible which is the only one with `linera wallet show` has been 

## Release Plan

## Links

None.

### DIFF
--- a/linera-client/src/client_context.rs
+++ b/linera-client/src/client_context.rs
@@ -6,7 +6,7 @@ use std::sync::Arc;
 use futures::Future;
 use linera_base::{
     crypto::{CryptoHash, ValidatorPublicKey},
-    data_types::{BlockHeight, Timestamp},
+    data_types::{BlockHeight, Epoch, Timestamp},
     identifiers::{Account, AccountOwner, ChainId},
     ownership::ChainOwnership,
     time::{Duration, Instant},
@@ -107,8 +107,9 @@ where
         chain_id: ChainId,
         owner: Option<AccountOwner>,
         timestamp: Timestamp,
+        epoch: Epoch,
     ) -> Result<(), Error> {
-        self.update_wallet_for_new_chain(chain_id, owner, timestamp)
+        self.update_wallet_for_new_chain(chain_id, owner, timestamp, epoch)
             .await
     }
 
@@ -306,6 +307,7 @@ impl<Env: Environment, W: Persist<Target = Wallet>> ClientContext<Env, W> {
         chain_id: ChainId,
         owner: Option<AccountOwner>,
         timestamp: Timestamp,
+        epoch: Epoch,
     ) -> Result<(), Error> {
         if self.wallet.get(chain_id).is_none() {
             self.mutate_wallet(|w| {
@@ -316,6 +318,7 @@ impl<Env: Environment, W: Persist<Target = Wallet>> ClientContext<Env, W> {
                     timestamp,
                     next_block_height: BlockHeight::ZERO,
                     pending_proposal: None,
+                    epoch: Some(epoch),
                 })
             })
             .await?;
@@ -388,7 +391,14 @@ impl<Env: Environment, W: Persist<Target = Wallet>> ClientContext<Env, W> {
         }
 
         self.wallet_mut()
-            .mutate(|w| w.assign_new_chain_to_owner(owner, chain_id, chain_description.timestamp()))
+            .mutate(|w| {
+                w.assign_new_chain_to_owner(
+                    owner,
+                    chain_id,
+                    chain_description.timestamp(),
+                    chain_description.config().epoch,
+                )
+            })
             .await
             .map_err(|e| error::Inner::Persistence(Box::new(e)))?
             .context("assigning new chain")?;
--- a/linera-client/src/wallet.rs
+++ b/linera-client/src/wallet.rs
@@ -5,7 +5,7 @@ use std::{collections::BTreeMap, iter::IntoIterator};
 
 use linera_base::{
     crypto::CryptoHash,
-    data_types::{BlockHeight, ChainDescription, Timestamp},
+    data_types::{BlockHeight, ChainDescription, Epoch, Timestamp},
     ensure,
     identifiers::{AccountOwner, ChainId},
 };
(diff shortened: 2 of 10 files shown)


linera-client/src/wallet.rs
```diff
@@ -165,6 +167,7 @@ pub struct UserChain {
     pub timestamp: Timestamp,
     pub next_block_height: BlockHeight,
     pub pending_proposal: Option<PendingProposal>,
+    pub epoch: Option<Epoch>,
```
afck, line 171: ~Maybe this could be non-optional: For root chains, it starts at 0, for child chains it starts at the same epoch as the parent chain. Basically we should always get the epoch from the same source as we get the timestamp.~

linera-client/src/wallet.rs
```diff
@@ -194,6 +198,7 @@ impl UserChain {
             timestamp,
             next_block_height: BlockHeight::ZERO,
             pending_proposal: None,
+            epoch: Epoch::ZERO,
```
afck: We pass in a timestamp here; do we also have the correct epoch at the call site?

linera-client/src/wallet.rs
```diff
@@ -207,6 +212,7 @@ impl UserChain {
...
```
afck: (Same here.)

linera-client/src/wallet.rs
MathieuDutSik: Yes, we could correct here.

linera-client/src/wallet.rs
MathieuDutSik: At that, I preferred to change the usage of `make_other` to set timestamp to 0 and epoch to 0.
By putting the timestamp at genesis or earlier, we guarantee that the epoch is zero.

afck: I don't think that's correct, though.

Whenever we add a _root_ chain to the wallet with height 0, it's correct to set the epoch to zero. But if we add a _child_ chain, we should ideally never do that, and get the epoch from the parent chain instead. Most child chains will _never_ be in epoch zero, not even if they don't have any blocks yet.

MathieuDutSik: Ok, but the problem is how to get the epoch in question.

We have two contexts where `make_other` is needed:
* The `make_chain_client` has access to the `Wallet`. The Wallet does not have any epoch available. Sure, we have the committee in the genesis config, but not the epoch. At the same time, that does not appear to matter since the `UserChain` is used by the `create_chain_client` but does not use the epoch. So, it somehow doesn't matter.
* We have a call from `FollowChain`. But it does not tell the parent chain.

So, I think a reasonable strategy would be to set it to `None` in those two contexts. Which are of importance, since they are not test codes.

afck: Right, then I guess it has to be an `Option` after all. Sorry for the detour!

MathieuDutSik: Not at all.
We set up to `Some(_)` in a larger set of contexts than when creating the PR.

linera-client/src/wallet.rs
```diff
@@ -194,11 +200,12 @@ impl UserChain {
...
             next_block_height: BlockHeight::ZERO,
             pending_proposal: None,
+            epoch: Some(epoch),
         }
     }
 
-    /// Creates an entry for a chain that we don't own. The timestamp must be the genesis
-    /// timestamp or earlier.
+    /// Creates an entry for a chain that we don't own. The timestamp is the genesis
+    /// timestamp or earlier. The Epoch is 0.
```
afck: Comment needs to be updated.

linera-client/src/client_context.rs
```diff
@@ -388,7 +391,14 @@ impl<Env: Environment, W: Persist<Target = Wallet>> ClientContext<Env, W> {
         }
 
         self.wallet_mut()
-            .mutate(|w| w.assign_new_chain_to_owner(owner, chain_id, chain_description.timestamp()))
+            .mutate(|w| {
+                w.assign_new_chain_to_owner(
+                    owner,
+                    chain_id,
+                    chain_description.timestamp(),
+                    chain_description.config().epoch,
```
afck, line 399: Or should we just pass in (a reference to) the chain description?

linera-client/src/wallet.rs
MathieuDutSik: Ok, done.

linera-client/src/wallet.rs
MathieuDutSik: Ok, done.

linera-client/src/wallet.rs
```diff
@@ -194,11 +200,12 @@ impl UserChain {
             timestamp,
             next_block_height: BlockHeight::ZERO,
             pending_proposal: None,
+            epoch: Some(epoch),
         }
     }
 
     /// Creates an entry for a chain that we don't own. The timestamp must be the genesis
-    /// timestamp or earlier.
+    /// timestamp or earlier. The Epoch is unknwon.
```
afck:
```suggestion
    /// timestamp or earlier. The epoch is unknown.
```

(Or "`None`"; or we could just remove that sentence again.)

linera-client/src/wallet.rs
```diff
@@ -194,11 +200,12 @@ impl UserChain {
             timestamp,
             next_block_height: BlockHeight::ZERO,
             pending_proposal: None,
+            epoch: Some(epoch),
```
afck, line 203:
```suggestion
            epoch: Some(description.config.epoch),
```

linera-client/src/wallet.rs
MathieuDutSik: Ok, for `None`.

linera-client/src/wallet.rs
MathieuDutSik: Thank you.

PR #4462, opened 2025-09-02
Replace fault type `Malicious` with `NoChains`.

### LINKED ISSUE #3858: Rethink faulty validators in tests
https://github.com/linera-io/linera-protocol/pull/3787#discussion_r2065778559
Currently, with the initial chain balance being a part of the ChainDescription, a different balance means the chain will have a different ID. Find a way of simulating faulty validators that disagree on the balance, but agree on the chain ID.

afck:
## Motivation

We simulate a range of validator faults in the `linera-core` tests. One of them is called `Malicious`, but it really just initializes the storage with wrong chains and returns `ArithmeticError`s in some cases. After https://github.com/linera-io/linera-protocol/pull/3787#discussion_r2065778559, the wrong initialization even causes the chains to have different chain ID, not just a wrong balance, so they will often return `InactiveChain` errors.

Even before that chain, the storage initialization fault didn't play well with `set_fault_type`: Even if we switch them to `Honest`, they will still be faulty due to their storage contents.

## Proposal

Replace `Malicious` with `NoChains`: Storage _does_ get initialized, but these validators return `InactiveChain` anyway, when handling block proposals, certificates or chain info queries.

## Test Plan

CI

## Release Plan

- (But might as well be backported: It's test only and not problematic.)

## Links

- Closes #3858.

### DIFF
--- a/linera-core/src/unit_tests/client_tests.rs
+++ b/linera-core/src/unit_tests/client_tests.rs
@@ -803,7 +803,7 @@ where
 {
     let signer = InMemorySigner::new(None);
     let mut builder = TestBuilder::new(storage_builder, 4, 0, signer).await?;
-    builder.set_fault_type([0, 1], FaultType::Malicious);
+    builder.set_fault_type([0, 1], FaultType::NoChains);
     let chain_1 = builder.add_root_chain(1, Amount::from_tokens(4)).await?;
     let chain_2 = builder.add_root_chain(2, Amount::from_tokens(4)).await?;
     let result = chain_1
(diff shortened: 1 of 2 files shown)


deuszx (approved):

bart-linera (approved):
The idea used to be that malicious nodes have the same chains as honest ones, only a different idea about the chain balances - but that became impossible once the initial balance started affecting the chain ID with the changing of `ChainDescription`s into blobs.

The change makes sense :+1:

PR #4464, opened 2025-09-02
Fix and re-enable `test_finalize_validated`.

### LINKED ISSUE #3860: Fix `test_finalize_validated`
The test is currently sometimes failing if the faulty validators respond to client0 before the correct ones. Make sure that we don't hit this edge case, as handling of it is not the point of the test.

afck:
## Motivation

This test is `#[ignore]`d because it's flaky: If the faulty validators respond before the other ones start processing the proposal, neither of them will have the proposal in their chain manager.

Also, `test_sparse_sender_chain` is flaky: If the randomly chosen validator (the first in the committee, i.e. by public key) is faulty, processing the notification will fail.

## Proposal

Explicitly handle the proposal in one of the validators to make sure the test is set up for the intended scenario.

In `test_sparse_sender_chain`, use two validators with no faulty one, so the selected validator is guaranteed to have the block.

## Test Plan

I ran it 20 times locally and it always passed.

## Release Plan

- (Backporting would be no problem.)

## Links

- Closes #3860.

### DIFF
--- a/linera-core/src/unit_tests/client_tests.rs
+++ b/linera-core/src/unit_tests/client_tests.rs
@@ -1271,7 +1271,7 @@ where
     B: StorageBuilder,
 {
     let signer = InMemorySigner::new(None);
-    let mut builder = TestBuilder::new(storage_builder, 4, 1, signer).await?;
+    let mut builder = TestBuilder::new(storage_builder, 2, 0, signer).await?;
     let sender = builder.add_root_chain(1, Amount::from_tokens(4)).await?;
     let receiver = builder.add_root_chain(2, Amount::ZERO).await?;
     let receiver_id = receiver.chain_id();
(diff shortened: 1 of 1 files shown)


bart-linera (approved):
:tada:

bart-linera:
Just as a note for the future: `test_sparse_sender_chains` needed fixing because it used one faulty validator, and after #4462 such validators have no chains. Before that PR, they used to have "fake" chains with zero initial balances, and because the test uses a chain that has a zero initial balance anyway, the faulty validator would still have the correct chain and the test would work even if it was chosen as the validator to use when processing the notification. Now, not having the chain, a faulty validator can't process the notification correctly, which is why the test became flaky and setting up the network without faulty validators fixes it.

PR #4465, opened 2025-09-02
Make transaction tracker oracle methods harder to misuse.

### LINKED ISSUE #3519: Make `next_replayed_oracle_response` safer to use.

afck:
## Motivation

In the execution code, oracles must always be used as follows:
* In replay mode, we consume an oracle response from the recorded ones.
* If _not_ in replay mode, we query the actual oracle to obtain the value.
* In both cases, we have to add it to the oracle responses in the execution outcome.

It's currently easy to forget one of those steps.

## Proposal

Refactor the transaction tracker's methods to make it a bit harder to use them the wrong way.

Also remove a few unused execution requests. (We could make these crate-private, but in a separate PR.)

## Test Plan

CI

## Release Plan

- (But could easily be backported.)

## Links

- Closes #3519.

### DIFF
--- a/linera-execution/src/execution_state_actor.rs
+++ b/linera-execution/src/execution_state_actor.rs
@@ -387,13 +387,8 @@ where
                 http_responses_are_oracle_responses,
                 callback,
             } => {
-                let response = if let Some(response) =
-                    self.txn_tracker.next_replayed_oracle_response()?
-                {
-                    match response {
-                        OracleResponse::Http(response) => response,
-                        _ => return Err(ExecutionError::OracleResponseMismatch),
-                    }
+                let maybe_response = if self.txn_tracker.is_replaying() {
+                    None
                 } else {
                     let headers = request
                         .headers
@@ -439,13 +434,17 @@ where
                             .min(committee.policy().maximum_oracle_response_bytes);
                     }
 
-                    self.receive_http_response(response, response_size_limit)
-                        .await?
+                    Some(OracleResponse::Http(
+                        self.receive_http_response(response, response_size_limit)
+                            .await?,
+                    ))
                 };
 
                 // Record the oracle response
-                self.txn_tracker
-                    .add_oracle_response(OracleResponse::Http(response.clone()));
+                let response = match self.txn_tracker.add_oracle_response(maybe_response)? {
+                    OracleResponse::Http(response) => response.clone(),
+                    _ => return Err(ExecutionError::OracleResponseMismatch),
+                };
 
                 callback.respond(response);
             }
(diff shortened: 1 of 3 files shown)


linera-execution/src/execution_state_actor.rs
```diff
@@ -387,13 +387,8 @@ where
...
                 callback,
             } => {
-                let response = if let Some(response) =
-                    self.txn_tracker.next_replayed_oracle_response()?
-                {
-                    match response {
-                        OracleResponse::Http(response) => response,
-                        _ => return Err(ExecutionError::OracleResponseMismatch),
-                    }
+                let maybe_response = if self.txn_tracker.is_replaying() {
```
deuszx, line 390: shouldn't this be a negation?
```suggestion
                let maybe_response = if !self.txn_tracker.is_replaying() {
```

Previously, the `next_replayed_oracle_response` would return:
> If not in replay mode, `None` is returned,

linera-execution/src/transaction_tracker.rs
```diff
@@ -241,11 +265,12 @@ impl TransactionTracker {
                 recorded_response == oracle_response,
                 ExecutionError::OracleResponseMismatch
             );
+            self.oracle_responses.push(recorded_response);
             true
         } else {
+            self.oracle_responses.push(oracle_response);
```
deuszx: Why this split? Wasn't the previous version correct already?

linera-execution/src/execution_state_actor.rs
afck: In the `None` case, i.e. in non-replay mode, we queried the actual oracle.

That's still the same: In non-replay mode we query the oracle (the `else` clause), but in replay mode we set this to `None` because we will obtain the actual value from the replaying responses below.

This PR just merges both calls so you can't forget one, i.e. we don't obtain the `Some` value in replay mode _yet_.

linera-execution/src/transaction_tracker.rs
afck: Good catch! Done in https://github.com/linera-io/linera-protocol/pull/4465/commits/160aae7ff5e3f788cd12b45bc96adea2ff1791a6.

linera-execution/src/execution_state_actor.rs
deuszx: This is more confusing than before IMO. 

Previously, when `next_replayed_oracle_response` returned `Some(Http(response))` we'd simply add that to `add_oracle_response`.

Now the `add_oracle_response` has new logic included that interprets `None` as "we should use a next entry from the oracle responses.

Maybe it'd be bit clearer if you didn't try to have single `let response` here and special case in the `if { ... }` and `else { ... }` clauses? Maybe 
```rust
if self.txn_tracker.is_replaying() {
  self.txn_tracker.replay(callback)
}
```

afck: Yes, maybe it's not much of an improvement after all. Let's do the removal part first: https://github.com/linera-io/linera-protocol/pull/4467

It's not in all call sites as simple as putting the response in the callback, so I think what you're suggesting would basically be the approach from before.

The case I'm worried about is that in one of the many places where we use oracles we forget to add the response in one of the two cases (or both).

deuszx: Maybe if we mark the `callback` as `must_use` and then make the code so that the only thing you can do with it is push via `self.transaction_tracker`?

afck: Right, so `replay` either returns an oracle response _and_ adds it to the outcomes, _or_ it returns a `must_use` value that can only be used by later adding the response from the actual oracle to the outcomes.

Maybe I'll try that approach tomorrow.

afck:
Closing in favor of https://github.com/linera-io/linera-protocol/pull/4471.

Issue #4255, opened 2025-07-22
Automatically download missing events from validators.


afck:

afck:
Done for the events expected to be available. Won't do for e.g. pre-checkpoint events.

PR #4466, opened 2025-09-02
Filter blocks in chains publishing events, introduce sparse chains wrt events

### LINKED ISSUE #4261: Make notifications more fine-grained, so that client know whether a relevant event was created.

bart-linera:
## Motivation

If we only listen to a chain because of the events it publishes, we might not want to download and process all the blocks.

## Proposal

Notify about the events in blocks separately from the notifications about new blocks. Filter those events client side, preprocess blocks with events if there are gaps.

(There could still be scenarios in which a client is missing events, and won't download them, effectively stalling the processing of events on a chain it owns. This will be taken care of as part of #4255 )

## Test Plan

CI - `test_wasm_end_to_end_social_event_streams` was extended to make sure that a sparse event publisher chain is created on a client in an appropriate scenario.

## Release Plan

## Links

- Closes #4261

### DIFF
--- a/linera-core/src/chain_worker/state.rs
+++ b/linera-core/src/chain_worker/state.rs
@@ -254,7 +254,7 @@ where
         }
         let response = self.prepare_chain_info_response(query).await?;
         // Trigger any outgoing cross-chain messages that haven't been confirmed yet.
-        let actions = self.create_network_actions().await?;
+        let actions = self.create_network_actions(None).await?;
         Ok((response, actions))
     }
 
@@ -356,8 +356,11 @@ where
         }
     }
 
-    /// Loads pending cross-chain requests.
-    async fn create_network_actions(&self) -> Result<NetworkActions, WorkerError> {
+    /// Loads pending cross-chain requests, and adds `NewRound` notifications where appropriate.
+    async fn create_network_actions(
+        &self,
+        old_round: Option<Round>,
+    ) -> Result<NetworkActions, WorkerError> {
         let mut heights_by_recipient = BTreeMap::<_, Vec<_>>::new();
         let mut targets = self.chain.nonempty_outbox_chain_ids();
         if let Some(tracked_chains) = self.tracked_chains.as_ref() {
@@ -371,13 +374,30 @@ where
             let heights = outbox.queue.elements().await?;
             heights_by_recipient.insert(target, heights);
         }
-        self.create_cross_chain_requests(heights_by_recipient).await
+        let cross_chain_requests = self
+            .create_cross_chain_requests(heights_by_recipient)
+            .await?;
+        let mut notifications = Vec::new();
+        if let Some(old_round) = old_round {
+            let round = self.chain.manager.current_round();
+            if round > old_round {
+                let height = self.chain.tip_state.get().next_block_height;
+                notifications.push(Notification {
+                    chain_id: self.chain_id(),
+                    reason: Reason::NewRound { height, round },
+                });
+            }
+        }
+        Ok(NetworkActions {
+            cross_chain_requests,
+            notifications,
+        })
     }
 
     async fn create_cross_chain_requests(
         &self,
         heights_by_recipient: BTreeMap<ChainId, Vec<BlockHeight>>,
-    ) -> Result<NetworkActions, WorkerError> {
+    ) -> Result<Vec<CrossChainRequest>, WorkerError> {
         // Load all the certificates we will need, regardless of the medium.
         let heights = BTreeSet::from_iter(heights_by_recipient.values().flatten().copied());
         let next_block_height = self.chain.tip_state.get().next_block_height;
@@ -424,7 +444,7 @@ where
             .zip(certificates)
             .collect::<HashMap<_, _>>();
         // For each medium, select the relevant messages.
-        let mut actions = NetworkActions::default();
+        let mut cross_chain_requests = Vec::new();
         for (recipient, heights) in heights_by_recipient {
             let mut bundles = Vec::new();
             for height in heights {
@@ -438,9 +458,9 @@ where
                 recipient,
                 bundles,
             };
-            actions.cross_chain_requests.push(request);
+            cross_chain_requests.push(request);
         }
-        Ok(actions)
+        Ok(cross_chain_requests)
     }
 
     /// Returns true if there are no more outgoing messages in flight up to the given
@@ -488,32 +508,20 @@ where
             }
         );
         certificate.check(committee)?;
-        let mut actions = NetworkActions::default();
         if self
             .chain
             .tip_state
             .get()
             .already_validated_block(certificate.inner().height())?
         {
-            return Ok((self.chain_info_response(), actions));
+            return Ok((self.chain_info_response(), NetworkActions::default()));
         }
         let old_round = self.chain.manager.current_round();
-        let timeout_chain_id = certificate.inner().chain_id();
-        let timeout_height = certificate.inner().height();
         self.chain
             .manager
             .handle_timeout_certificate(certificate, self.storage.clock().current_time());
-        let round = self.chain.manager.current_round();
-        if round > old_round {
-            actions.notifications.push(Notification {
-                chain_id: timeout_chain_id,
-                reason: Reason::NewRound {
-                    height: timeout_height,
-                    round,
-                },
-            })
-        }
         self.save().await?;
+        let actions = self.create_network_actions(Some(old_round)).await?;
         Ok((self.chain_info_response(), actions))
     }
 
@@ -580,7 +588,6 @@ where
         let (epoch, committee) = self.chain.current_committee()?;
         check_block_epoch(epoch, header.chain_id, header.epoch)?;
         certificate.check(committee)?;
-        let mut actions = NetworkActions::default();
         let already_committed_block = self.chain.tip_state.get().already_validated_block(height)?;
         let should_skip_validated_block = || {
             self.chain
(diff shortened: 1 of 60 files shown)


linera-chain/src/chain.rs
```diff
@@ -277,6 +277,9 @@ where
     pub previous_event_blocks: MapView<C, StreamId, BlockHeight>,
     /// Mailboxes used to send messages, indexed by their target.
     pub outboxes: ReentrantCollectionView<C, ChainId, OutboxStateView<C>>,
+    /// The indices of latest events emitted per stream (could be ahead of the last
+    /// executed block in sparse chains).
+    pub last_events: MapView<C, StreamId, u32>,
```
afck: (I wonder whether we should make `stream_event_counts` also track the last emitted event, rather than the index of the next one, for consistency.)

linera-chain/src/chain.rs
```diff
@@ -1133,6 +1141,34 @@ where
...
+                if index.checked_sub(1) == current_max_index {
+                    updated_streams.insert(stream_id.clone());
+                    current_max_index = Some(index);
+                }
+            }
+            if let Some(max_index) = current_max_index {
+                self.last_events.insert(&stream_id, max_index)?;
+            }
+        }
+        Ok(updated_streams)
```
afck, line 1186: (So the `last_events` entries are the index of the last event of the contiguous initial segment? _Not_ the actual highest index we've seen? Then maybe it would be easier to describe if it was the lowest index we _haven't_ seen yet, just like in `stream_event_counts`?)

linera-client/src/chain_listener.rs
```diff
@@ -210,12 +214,14 @@ impl<C: ClientContext> ChainListener<C> {
                 .make_chain_client(admin_chain_id)
                 .synchronize_from_validators()
                 .await?;
-            BTreeSet::from_iter(
+            // TODO: do we want to only listen to events on the admin chain?
```
afck: Let's not merge `TODO`s without an associated GitHub issue number.

But I think you're right, yes: Most clients only care about the events, not other admin chain blocks. :+1:

linera-service/src/cli_wrappers/wallet.rs
```diff
@@ -1250,6 +1250,31 @@ impl NodeService {
...
+        amount: Amount,
+    ) -> Result<()> {
+        let json_owner = owner.to_value();
+        let json_recipient = recipient.to_value();
+        let query = format!(
+            "mutation {{ transfer(\
+                 chainId: \"{chain_id}\",
+                 owner: {json_owner}, \
+                 recipient: {json_recipient}, \
+                 amount: \"{amount}\")\
```
afck:
```suggestion
            "mutation {{ transfer(\
                 chainId: \"{chain_id}\", \
                 owner: {json_owner}, \
                 recipient: {json_recipient}, \
                 amount: \"{amount}\") \
```

linera-service/src/cli_wrappers/wallet.rs
```diff
@@ -1250,6 +1250,31 @@ impl NodeService {
...
+             }}"
+        );
+        let data = self.query_node(query).await?;
+        let _transfer = data["transfer"]
+            .as_str()
+            .context("missing transfer string in response")?
+            .trim();
+        Ok(())
```
afck: Might as well return the block hash, like e.g. below in `publish_data_blob`.

linera-core/src/chain_worker/state.rs
```diff
@@ -815,14 +829,16 @@ where
             .flatten()
             .map(|event| event.stream_id.clone())
             .collect();
```
afck: That could duplicate notifications, though: We should probably only notify about those where the `last_events` entry has actually changed. Some of them may already have been sent because the block was already preprocessed.

linera-core/src/client/mod.rs
```diff
@@ -3625,16 +3695,81 @@ impl<Env: Environment> ChainClient<Env> {
...
+                        {
+                            error!("NewBlock: Fail to synchronize new block after notification");
+                        }
+                        trace!(
+                            chain_id = %self.chain_id,
+                            %height,
+                            "NewBlock: processed notification",
+                        );
+                    }
+                    ListeningMode::EventsOnly(_) => {
```
afck, line 3698: In the `EventsOnly` case we also want to ignore `NewRound` notifications.

linera-core/src/client/mod.rs
```diff
@@ -3625,16 +3695,81 @@ impl<Env: Environment> ChainClient<Env> {
...
+                height,
+                hash,
+                event_streams,
+            } => {
                 if self
-                    .local_next_block_height(chain_id, &mut local_node)
+                    .local_next_block_height(notification.chain_id, &mut local_node)
                     .await?
-                    <= height
+                    > height
```
afck, line 3715: I wonder whether instead of (or in addition to) the block height the notification should contain the event indices. Then we could compare it to the `last_events` in our local chain state, which would tell us more accurately whether we need to (pre)process that block now.

linera-core/src/client/mod.rs
```diff
@@ -3625,16 +3695,81 @@ impl<Env: Environment> ChainClient<Env> {
...
+                    debug!(
+                        chain_id = %self.chain_id,
+                        "Accepting redundant notification for new block"
+                    );
+                    return Ok(());
                 }
+                let should_process = match listening_mode {
+                    ListeningMode::FullChain => true,
+                    ListeningMode::EventsOnly(relevant_events) => {
+                        relevant_events.intersection(&event_streams).count() != 0
```
afck: Maybe `next().is_some()`, to avoid iterating over all the events in the intersection. Also in the chain listener.

linera-core/src/client/mod.rs
```diff
@@ -3625,16 +3695,81 @@ impl<Env: Environment> ChainClient<Env> {
...
+                    );
+                    return Ok(());
+                }
+                trace!(
+                    chain_id = %self.chain_id,
+                    %height,
+                    "NewEvents: processing notification"
+                );
+                let mut certificates = remote_node.node.download_certificates(vec![hash]).await?;
+                // download_certificates ensure that we will get exactly one
```
```suggestion
                // download_certificates ensures that we will get exactly one
```

linera-chain/src/chain.rs
```diff
@@ -937,16 +941,20 @@ where
     }
 
     /// Adds a block to `preprocessed_blocks`, and updates the outboxes where possible.
-    pub async fn preprocess_block(&mut self, block: &ConfirmedBlock) -> Result<(), ChainError> {
+    pub async fn preprocess_block(
```
deuszx, line 946: Please add a comment what's the set of `StreamId`s returned exectly.

linera-chain/src/chain.rs
bart-linera: I think my main reason for doing it this way was that the check for continuity is easier - I don't need special handling for the case of the stream not being present in `last_events`. But I guess I can also do `unwrap_or(0)` after `self.last_events_get()`, which is just as simple, and will be more consistent :+1: (only the field name will probably need changing, too)

linera-client/src/chain_listener.rs
bart-linera: I'll just set it to listen to only epoch events by default, then :+1:

linera-core/src/client/mod.rs
bart-linera: I guess it would make sense, but isn't that potentially a significant amount of data for a notification? There could be multiple streams with multiple events per stream in a block.

linera-chain/src/chain.rs
bart-linera: I guess I'll go with the other suggestion, that is: changing `last_events` to be consistent with this one.

linera-chain/src/chain.rs
```diff
@@ -1156,15 +1161,20 @@ where
...
+                .next_expected_events
+                .get(&stream_id)
+                .await?
+                .unwrap_or(0);
             for index in indices {
-                if index.checked_sub(1) == current_max_index {
+                if index == current_expected_index {
                     updated_streams.insert(stream_id.clone());
-                    current_max_index = Some(index);
+                    current_expected_index = index.saturating_add(1);
```
afck, line 1178: (Right, maybe your variant was better, and both collections should use that. :sweat_smile: Anyway, in practice either is fine.)

linera-service/src/cli_wrappers/wallet.rs
```diff
@@ -1256,23 +1256,20 @@ impl NodeService {
...
              }}"
         );
         let data = self.query_node(query).await?;
-        let _transfer = data["transfer"]
-            .as_str()
-            .context("missing transfer string in response")?
-            .trim();
-        Ok(())
+        serde_json::from_value(data["transfer"].clone())
+            .context("missing publishDataBlob field in response")
```
afck:
```suggestion
            .context("missing transfer field in response")
```

bart-linera: Risks of copy-paste :sweat_smile:

Cargo.toml
```diff
@@ -55,7 +55,6 @@ default-members = [
     "linera-summary",
     "linera-views",
     "linera-views-derive",
-    # "linera-web", # not this one
```
afck: Or was that deliberate, to make sure we don't accidentally add it?
(If it is, we should add a better comment.)

bart-linera: Argh, this gets removed automatically by the script updating schemas. I usually remember to not commit that change, but not this time :weary:

afck (approved):

bart-linera:
Listening to just the events on the admin chain turned out to be too complicated for this PR, because then the committees in the execution state of the admin chain don't get updated, and a lot of places in the code depend on those being up to date.

PR #4468, opened 2025-09-02
Raw read cert

deuszx:


linera-rpc/proto/rpc.proto
```diff
@@ -116,6 +116,15 @@ message CertificatesBatchResponse {
   repeated Certificate certificates = 1;
 }
 
+message CertificatesBatch {
+  repeated CertificateRaw certificates = 1;
+}
+
+message CertificateRaw {
```
ma2bd, line 123: nit: naming -> `RawCertificate`?

It's quite possible that we don't need the other `Certificate` at all. I looked at `GrpcProxyable` and it's not using the specific fields.

deuszx: > It's quite possible that we don't need the other `Certificate` at all. I looked at `GrpcProxyable` and it's not using the specific fields.

We need it for backwards compatiblity for `DownloadCertificates` endpoint and others 🤔

deuszx: > nit: naming -> `RawCertificate`?

I prefer to start with the noun. Also, the _certificate_ part is more important (it's a domain object) and is something I'd probably use when searching for the symbol.

deuszx:
Closing in favor of more targetable PRs.

PR #4473, opened 2025-09-03
Send `NewRound` notifications after proposals.

### LINKED ISSUE #4472: Block proposals should trigger `NewRound` notifications.
If in multi-leader mode, all owners' clients should be informed whenever the round number changes, so they don't propose a conflicting block in the same round.

afck:
## Motivation

`NewRound` notifications are mainly needed for single-leader mode, where they are triggered by timeout certificates.

However, in multi-leader mode, in case of failed proposals, they are also useful, so that the other owners can avoid making a conflicting proposal in the same round.

## Proposal

Send `NewRound` notifications in `handle_block_proposal`, too.

## Test Plan

An assertion was added to a worker test.

## Release Plan

## Links

- Closes #4472.

### DIFF
--- a/linera-core/src/chain_worker/state.rs
+++ b/linera-core/src/chain_worker/state.rs
@@ -254,7 +254,7 @@ where
         }
         let response = self.prepare_chain_info_response(query).await?;
         // Trigger any outgoing cross-chain messages that haven't been confirmed yet.
-        let actions = self.create_network_actions().await?;
+        let actions = self.create_network_actions(None).await?;
         Ok((response, actions))
     }
 
@@ -356,8 +356,11 @@ where
         }
     }
 
-    /// Loads pending cross-chain requests.
-    async fn create_network_actions(&self) -> Result<NetworkActions, WorkerError> {
+    /// Loads pending cross-chain requests, and adds `NewRound` notifications where appropriate.
+    async fn create_network_actions(
+        &self,
+        old_round: Option<Round>,
+    ) -> Result<NetworkActions, WorkerError> {
         let mut heights_by_recipient = BTreeMap::<_, Vec<_>>::new();
         let mut targets = self.chain.nonempty_outbox_chain_ids();
         if let Some(tracked_chains) = self.tracked_chains.as_ref() {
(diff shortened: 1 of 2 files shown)


linera-core/src/chain_worker/state.rs
```diff
@@ -1356,6 +1341,18 @@ where
         self.chain.save().await?;
         Ok(())
     }
+
+    /// Adds a `NewRound` notification to the network actions if the round has changed.
+    fn add_new_round_notification(&self, old_round: Round, actions: &mut NetworkActions) {
```
deuszx: Why not call this in `create_network_actions`? So that we don't have to remember to call it explicitly.

afck: Then we'd have to pass in the old round there, too, and we only need it in three of the six places where we call `create_network_actions`. I guess we could pass in an option? Not sure if that's better.

deuszx: At least you won't forget to call it (like with the oracles)

afck: Actually, we also don't always use `create_network_actions` when we need to `add_new_round_notification`: In `process_timeout` we only need the latter, since that never triggers messages.

deuszx: I don't think it matters, timeout notification is one of the network actions. If `create_network_actions` doesn't add anything itself, the new round notification will be there.

afck: I guess then I should do the same for the `NewBlock` notification. That should also be added inside `create_network_actions`.

But that means I have to pass in not only the old round but also whether this is a new height.

deuszx: You can do it in a separate PR if you prefer – to gather everything in one place.

afck: I did everything except `NewBlock` in https://github.com/linera-io/linera-protocol/pull/4473/commits/790e3775cea1c2e26d6709fac14c9de306425096. Not sure how to do that last one cleanly yet. Yes, let's leave that for another PR.

PR #4475, opened 2025-09-04
feat(docker): automatic host setup for ScyllaDB with optional XFS support

eldios:
## Summary
- Added automatic host system configuration for ScyllaDB via a setup container
- Fixed critical AIO (Asynchronous I/O) limits that were causing ScyllaDB to fail
- XFS filesystem is now completely optional - provides performance benefits but not required
- All system parameters are dynamically calculated based on available resources

## Key Issue Resolved
ScyllaDB was failing with `io_setup: Resource temporarily unavailable` due to insufficient AIO slots. The real issue was not XFS but the host's AIO limits being too low (65536 vs required 88208+).

### 🚀 Major Features
- **New `scylla-setup` container** that runs before ScyllaDB to configure host system:
  - Uses Ubuntu 24.04 with bash for reliability
  - Runs in privileged mode to modify host kernel parameters
  - Dynamically calculates all values based on CPU cores and memory
  - Always completes successfully (warnings for failed settings)

- **Dynamic parameter calculation**:
  - AIO limits: `cores × shards × 65536 + 50% buffer`
  - Network settings: Scaled with CPU cores (somaxconn, syn_backlog)
  - Memory settings: Based on available RAM (vm.max_map_count)
  - File descriptors: 200K per core (minimum 1M)
  - Network buffers: 1% of RAM (16MB-128MB range)

- **XFS is now optional**:
  - Removed `--skip-xfs-check` flag (unnecessary for optional feature)
  - XFS path validation provides information only, never blocks deployment
  - Script continues with Docker volumes if XFS not configured
  - Clear messaging that XFS is optional, not required

### 🔧 System Optimizations
The setup container configures:
- ✅ AIO max-nr (critical for ScyllaDB)
- ✅ Network settings (somaxconn, TCP buffers, SYN cookies)
- ✅ Memory settings (swappiness, dirty ratios, max_map_count)
- ✅ Transparent Huge Pages disabled
- ✅ CPU governor to performance mode
- ✅ NUMA balancing disabled (ScyllaDB manages it)
- ✅ File descriptor limits

### 📚 Documentation
- Updated help text to show XFS as optional
- Added clear instructions for XFS setup using LABEL/UUID
- Removed confusing `--skip-xfs-check` option
- Log output improvements to stderr for clean command substitution

## Test Plan
- [x] Verify scylla-setup container runs and configures host
- [x] Test ScyllaDB starts successfully without XFS
- [x] Confirm AIO limits are properly set
- [x] Validate dynamic calculation based on system resources
- [x] Test with XFS path shows informational messages only
- [x] Ensure deployment never blocks on XFS issues
- [x] Verify backward compatibility

## Context
The deployment was failing because:
1. ScyllaDB requires high AIO limits (async I/O operations)
2. Default Linux systems have low AIO limits (65536)
3. ScyllaDB with `--developer-mode 0` enforces production requirements

The solution automatically configures the host system for optimal ScyllaDB performance while making XFS completely optional.

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/scripts/deploy-validator.sh
+++ b/scripts/deploy-validator.sh
@@ -720,12 +720,6 @@ services:
       SCYLLA_DIRECT_IO_MODE: "true"
       # Set appropriate cache size (adjust based on available RAM)
       SCYLLA_CACHE_SIZE: "4G"
-
-# Remove the default local volume since we're using bind mount
-volumes:
-  linera-scylla-data:
-    external: true
-    name: unused-scylla-volume
 EOF
 
 	log INFO "Docker Compose override generated at: ${compose_override}"
--- a/scripts/deploy-validator.sh
+++ b/scripts/deploy-validator.sh
@@ -15,6 +15,8 @@
 #   --skip-genesis      - Skip downloading genesis configuration
 #   --force-genesis     - Force re-download of genesis configuration
 #   --custom-tag TAG    - Use custom image tag (for testing, no _release suffix)
+#   --xfs-path PATH     - XFS partition mount path for ScyllaDB (required)
+#   --skip-xfs-check    - Skip XFS validation (not recommended for production)
 #   --help, -h          - Show help message
 #   --dry-run           - Show what would be done without executing
 #   --verbose, -v       - Enable verbose output
@@ -31,12 +33,14 @@
 #   PORT                - Internal validator port (default: 19100)
 #   METRICS_PORT        - Metrics collection port (default: 21100)
 #   NUM_SHARDS          - Number of validator shards (default: 4)
+#   SCYLLA_XFS_PATH     - XFS partition mount path for ScyllaDB data (required)
 #
 # Requirements:
 #   - Docker
 #   - Docker Compose (as plugin)
 #   - Git (for branch detection)
 #   - wget (for genesis download)
+#   - XFS formatted partition for ScyllaDB (production deployments)
 #
 # Author: Linera Team
 # Date: $(date +%Y-%m-%d)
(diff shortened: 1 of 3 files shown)


scripts/deploy-validator.sh
```diff
@@ -583,6 +591,147 @@ validate_email() {
 	return 0
 }
 
+# Check AIO configuration
+check_aio_configuration() {
+	local current_aio
+	local required_aio=1048576
```
ndr-ds, line 603: Our kubernetes deployments seem to use 4082080 btw ha but maybe 1048576 is enough

scripts/deploy-validator.sh
```diff
@@ -583,6 +591,147 @@ validate_email() {
...
+    volumes:
+      # Override default volume with XFS bind mount
+      - ${scylla_data_dir}:/var/lib/scylla
+    # Additional ScyllaDB optimizations for XFS
+    environment:
+      SCYLLA_AUTO_CONF: 1
+      # Enable direct I/O on XFS
+      SCYLLA_DIRECT_IO_MODE: "true"
+      # Set appropriate cache size (adjust based on available RAM)
+      SCYLLA_CACHE_SIZE: "4G"
```
ndr-ds: Should this be a parameter with 5GB as the default since it should be adjusted?

scripts/deploy-validator.sh
eldios: that number comes from here:
```
Setting up a Cluster
Before starting the cluster, make sure the [aio-max-nr](https://www.kernel.org/doc/Documentation/sysctl/fs.txt) value is high enough (1048576 or more).
```
https://university.scylladb.com/setup-a-scylla-cluster/

We can change it if we feel it's not enough in our tests.

scripts/deploy-validator.sh
eldios: I don't remember where I got that number from.. we can change it to whatever, why 5GB in particular over 4GB?

ndr-ds: That was a typo, I meant 4G. What I'm saying is since it's supposed to be adjusted, it could be an option of the shell script. Like an optional `--scylla-cache-size` or something

eldios: added

ndr-ds (approved):

PR #4478, opened 2025-09-04
Add new DownloadRawCertificatesByHeight endpoint and use it.

deuszx:
## Motivation

During our performance tests it was found that proxies' bottleneck is (de)serialization of certificates when responding to `DownloadCertificatesByHeights` request.

## Proposal

It is a waste of CPU cycles to deserialize data (after loading from the DB) only to serialize it for transporting over gRPC protocol.

This PR:
- adds a new method on the storage `read_certificates_raw` which returns raw bytes of the requested certificates
- adds a new endpoint to `rpc.proto/ValidatorNode` – `DownloadRawCertificatesByHeight` that responds with (bcs) bytes of the requested certificates
- updates proxy code to serve the new endpoint
- the gRPC _client_  does NOT get a new method, instead the old `download_certificates_by_height` is modified to use the new endpoint. **This is safe:** b/c we will release this as a new SDK version meaning old clients still use old code paths while next version uses new.

## Test Plan

CI

## Release Plan

    - be released in a new SDK,
    - be released in a new SDK,

## Links

### DIFF
--- a/linera-service/src/proxy/grpc.rs
+++ b/linera-service/src/proxy/grpc.rs
@@ -35,7 +35,8 @@ use linera_rpc::{
             BlobContent, BlobId, BlobIds, BlockProposal, Certificate, CertificatesBatchRequest,
             CertificatesBatchResponse, ChainInfoResult, CryptoHash, HandlePendingBlobRequest,
             LiteCertificate, NetworkDescription, Notification, PendingBlobRequest,
-            PendingBlobResult, SubscriptionRequest, VersionInfo,
+            PendingBlobResult, RawCertificate, RawCertificatesBatch, SubscriptionRequest,
+            VersionInfo,
         },
         pool::GrpcConnectionPool,
         GrpcProtoConversionError, GrpcProxyable, GRPC_CHUNKED_MESSAGE_FILL_LIMIT,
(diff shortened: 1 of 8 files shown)


linera-service/src/proxy/grpc.rs
```diff
@@ -770,6 +839,14 @@ impl<T> GrpcMessageLimiter<T> {
...
         Ok(true)
     }
+
+    /// Adds the given number of bytes to the remaining capacity.
+    ///
+    /// Returns the new remaining capacity.
+    fn add(&mut self, bytes_len: usize) -> usize {
+        self.remaining = self.remaining.saturating_sub(bytes_len);
+        self.remaining
+    }
```
afck, line 849: Might be better to mimic `fits`, and call this `fits_raw`: return `true` if `bytes_len` is at most `remaining`, even if `remaining` is `0` afterwards.

And then `fits` could probably use `fits_raw` internally.

linera-storage/src/db_storage.rs
```diff
@@ -824,6 +824,39 @@ where
...
+        &self,
+        hashes: I,
+    ) -> Result<Vec<(Vec<u8>, Vec<u8>)>, ViewError> {
+        let hashes = hashes.into_iter().collect::<Vec<_>>();
+        if hashes.is_empty() {
+            return Ok(Vec::new());
+        }
+        let keys = Self::get_keys_for_certificates(&hashes)?;
+        let store = self.database.open_shared(&[])?;
+        let values = store.read_multi_values_bytes(keys).await;
```
afck: Why not `?` here? We only record the metrics if `is_ok` anyway, so we might as well return here on error.

linera-service/src/proxy/grpc.rs
deuszx: [Changes to fits/fit_raw](https://github.com/linera-io/linera-protocol/pull/4478/commits/4d132de19de8593db22e107674893cffdd2ff365)

linera-storage/src/db_storage.rs
deuszx: [Return error early from read_certificates_raw](https://github.com/linera-io/linera-protocol/pull/4478/commits/aba6439c2c91f5a5367b87aaba0538ff75fc7108)

linera-service/src/proxy/grpc.rs
```diff
@@ -833,19 +834,18 @@ impl<T> GrpcMessageLimiter<T> {
...
     /// Adds the given number of bytes to the remaining capacity.
     ///
-    /// Returns the new remaining capacity.
-    fn add(&mut self, bytes_len: usize) -> usize {
+    /// Returns whether we managed to fit the element.
+    fn fits_raw(&mut self, bytes_len: usize) -> bool {
+        if self.remaining < bytes_len {
+            return false;
+        }
         self.remaining = self.remaining.saturating_sub(bytes_len);
```
afck, line 847: (Might as well use `-` now.)

afck (approved):

PR #4480, opened 2025-09-04
Add bool to ChainInfoQuery to trigger create_network_actions

deuszx:


linera-core/src/updater.rs
```diff
@@ -431,11 +431,11 @@ where
...
-                let query = ChainInfoQuery::new(chain_id);
+                let query = ChainInfoQuery::new(chain_id).no_network_actions();
                 self.remote_node.handle_chain_info_query(query).await?
             }
         } else {
-            let query = ChainInfoQuery::new(chain_id);
+            let query = ChainInfoQuery::new(chain_id).no_network_actions();
             self.remote_node.handle_chain_info_query(query).await?
         };
         let initial_block_height = remote_info.next_block_height;
```
ma2bd, line 441: You may have to add an argument to `send_chain_info_up_to_heights` etc so that `send_chain_information_for_senders` has the same behavior as before. (I believe that's the only place.)

Alternatively @afck suggested we could send the highest lite certificate that we know instead of the chain info query. (This seems like a good in general in the updater. It will return a minimal chain info response in case of failure)

deuszx: Let's continue the discussion in https://github.com/linera-io/linera-protocol/pull/4518

I think this is not necessary b/c it's already happening: whenever `send_chain_info_up_to_heights` leads to `send_chain_information` sending`ChainInfoQuery` to a remote node, it will end up sending a confirmed certificate as well and those trigger creation of network actions. And we have to send a `ChainInfoQuery` only once, when we really need to know only remote node's knowledge about that particular chain.

PR #4481, opened 2025-09-04
Replace `dashmap` with `papaya`.

afck:
## Motivation

`dashmap` can cause deadlocks if guards are held across `await` points, but the compiler does not warn about it.

## Proposal

Use the lock-free `papaya` instead.

## Test Plan

CI (no new functionality)

## Release Plan

- (But could be backported.)

## Links


linera-core/src/client/chain_client_state.rs
```diff
@@ -32,6 +32,15 @@ impl ChainClientState {
         }
     }
 
+    /// Clones the state. This must only be used to update the state, and one of the two clones
+    /// must be dropped.
+    pub fn clone_for_update(&self) -> ChainClientState {
```
ma2bd: Can we make this a private method?

ma2bd: maybe `clone_for_update_unchecked` so that people read the comment?

linera-core/src/client/mod.rs
```diff
@@ -1582,62 +1579,44 @@ impl ChainClientError {
...
+    /// Updates the chain's state using a closure.
+    #[instrument(level = "trace", skip(self, f))]
+    fn update_state<F>(&self, f: F)
+    where
+        F: Fn(&mut ChainClientState),
+    {
+        let chains = self.client.chains.pin();
+        chains
+            .update(self.chain_id, |state| {
+                let mut state = state.clone_for_update();
```
ma2bd: The comment says one of the two clones must be dropped but that's not the case here.

linera-core/src/client/chain_client_state.rs
ma2bd: TBH I don't quite yet understand the comment or the reason behind it (see other question)

linera-core/src/client/mod.rs
afck: `update` needs a side-effect-free `Fn` (_not_ `FnOnce`); in the case of a conflict it may get executed more than once. So we can't move or mutate the state here; we have to clone it. However, after the `update` call, only one clone will remain in the map.

linera-core/src/client/chain_client_state.rs
afck: Done in https://github.com/linera-io/linera-protocol/pull/4481/commits/667d27f6452a2423651618416940327e80251f9b.

linera-core/src/client/mod.rs
```diff
@@ -1582,62 +1579,44 @@ impl ChainClientError {
...
+    /// Updates the chain's state using a closure.
+    #[instrument(level = "trace", skip(self, f))]
+    fn update_state<F>(&self, f: F)
+    where
+        F: Fn(&mut ChainClientState),
+    {
+        let chains = self.client.chains.pin();
+        chains
+            .update(self.chain_id, |state| {
+                let mut state = state.clone_for_update_unchecked();
```
ma2bd, line 1615: Afaict we always update the `pending_proposal` here and never the `client_mutex`. Maybe that's ok.

afck: We only do that in `clear_pending_proposal` and `new_pending_block`.
* It looks like the former is only called in tests now! I'll make it test-only…
* The latter is only called in `execute_block`, where we already hold a mutex.

afck: Done in https://github.com/linera-io/linera-protocol/pull/4481/commits/0045340d68eab68e78e1de6b971631085a50b816.

PR #4483, opened 2025-09-04
[testnet] Add new DownloadRawCertificatesByHeight endpoint and use it.

deuszx:
## Motivation

During our performance tests it was found that proxies' bottleneck is (de)serialization of certificates when responding to `DownloadCertificatesByHeights` request.

## Proposal

It is a waste of CPU cycles to deserialize data (after loading from the DB) only to serialize it for transporting over gRPC protocol.

This PR:
- adds a new method on the storage `read_certificates_raw` which returns raw bytes of the requested certificates
- adds a new endpoint to `rpc.proto/ValidatorNode` – `DownloadRawCertificatesByHeight` that responds with (bcs) bytes of the requested certificates
- updates proxy code to serve the new endpoint
- the gRPC _client_  does NOT get a new method, instead the old `download_certificates_by_height` is modified to use the new endpoint. **This is safe:** b/c we will release this as a new SDK version meaning old clients still use old code paths while next version uses new.

## Test Plan

CI

## Release Plan

    - be released in a new SDK,
    - be released in a new SDK,

## Links

Testnet backport of https://github.com/linera-io/linera-protocol/pull/4478

### DIFF
--- a/linera-service/src/proxy/grpc.rs
+++ b/linera-service/src/proxy/grpc.rs
@@ -35,7 +35,8 @@ use linera_rpc::{
             BlobContent, BlobId, BlobIds, BlockProposal, Certificate, CertificatesBatchRequest,
             CertificatesBatchResponse, ChainInfoResult, CryptoHash, HandlePendingBlobRequest,
             LiteCertificate, NetworkDescription, Notification, PendingBlobRequest,
-            PendingBlobResult, SubscriptionRequest, VersionInfo,
+            PendingBlobResult, RawCertificate, RawCertificatesBatch, SubscriptionRequest,
+            VersionInfo,
         },
         pool::GrpcConnectionPool,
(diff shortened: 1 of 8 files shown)


linera-service/src/proxy/grpc.rs
```diff
@@ -754,11 +824,18 @@ impl<T> GrpcMessageLimiter<T> {
...
+    /// Adds the given number of bytes to the remaining capacity.
+    ///
+    /// Returns whether we managed to fit the element.
+    fn fits_raw(&mut self, bytes_len: usize) -> bool {
+        if self.remaining < bytes_len {
+            return false;
         }
-        self.remaining -= required;
-        Ok(true)
+        self.remaining = self.remaining.saturating_sub(bytes_len);
```
afck, line 837: (Might as well use `-`.)

deuszx:
> (Assuming this is the same, and there were no major merge conflicts.)

I think there were no merge conflicts at all. The +/- is exactly the same.

PR #4469, opened 2025-09-02
eBPF Pyroscope

ndr-ds:
## Motivation

We're adding Pyroscope support back, but using eBPF this time

## Proposal

* Add Pyroscope to our helm chart
* Add a Grafana dashboard with the profiling info

## Test Plan

Deployed a network and saw the Pyroscope profiling data both on the Pyroscope UI and on Grafana, as well as logs from Pyroscope on Loki:

## Release Plan

- Nothing to do / These changes follow the usual release cycle.

PR #4487, opened 2025-09-04
Optimize RocksDB prefix searches

ndr-ds:
## Release Plan

    - be released in a new SDK,
    - be released in a new SDK,
      SDK soon.

## Links


ndr-ds:
* **#4487** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4487?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#4498](https://github.com/linera-io/linera-protocol/pull/4498) <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4498?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#4571](https://github.com/linera-io/linera-protocol/pull/4571) <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4571?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4487?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#4469** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4469?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

linera-views/src/backends/rocks_db.rs
```diff
@@ -198,20 +217,36 @@ impl RocksDbStoreExecutor {
...
+        // Enable async I/O for better concurrency
+        read_opts.set_async_io(true);
+
+        let mut upper_bound = prefix.clone();
+        if let Some(last_byte) = upper_bound.last_mut() {
+            if *last_byte < 255 {
+                *last_byte += 1;
+                read_opts.set_iterate_upper_bound(upper_bound);
+            }
+        }
```
ma2bd: should we factorize this code?

deuszx:
Thanks for the PR @ndr-ds . I can follow the motivation and the code but I am missing a proof that it indeed works. Can you share the results? Did you run the test after making all of the changes or do you have piece-wise as well (ie bench after making just one of the changes).

MathieuDutSik (commented):
I have the following issues with the PR:
* The construction of the `upper_bound` is reminicent of `get_upper_bound_option` but in a wrong way. For a key of the form `[0, 2, 255]` it does not give any upper bound while actually you have one, which is `[0, 3]`. So, reuse that code.
* The use of the `read_opts.set_iterate_upper_bound` is a nice idea. But if doing that, then it means that the comparison with `key.starts_with(&prefix)` is no longer needed since it is done in RocksDB.
* The use of `set_async_io` looks nice but what exactly is changed?
* We now have two tests for the iterator: `iter.valid()` and `iter.key()`. Could we reduce that and have just a single `while Some(key) = key.next()` or similar.

So, I think it is a good idea to use `read_opts`, the PR requires further work.

ndr-ds:
`set_async_io` does prefetching, so it improves performance of sequential reads

linera-views/src/backends/rocks_db.rs
```diff
@@ -166,25 +166,38 @@ impl RocksDbStoreExecutor {
...
+
         let mut prefix = self.start_key.clone();
         prefix.extend(key_prefix);
         let len = prefix.len();
-        let mut iter = self.db.raw_iterator();
-        let mut keys = Vec::new();
+
+        let read_opts = self.get_find_prefix_read_opts(&prefix);
+        let mut iter = self.db.raw_iterator_opt(read_opts);
         iter.seek(&prefix);
```
MathieuDutSik: It seems that this part of the code is repeated between `find_keys_by_prefix_internal` and `find_key_values_by_prefix_internal`.

Therefore, I think more code can be moved to `get_find_prefix_read_opts`, which can be renamed as `get_initial_iterator`.

linera-views/src/backends/rocks_db.rs
```diff
@@ -198,20 +211,18 @@ impl RocksDbStoreExecutor {
...
-        let mut next_key = iter.key();
-        while let Some(key) = next_key {
-            if !key.starts_with(&prefix) {
-                break;
-            }
+
+        let mut key_values = Vec::new();
+        while let Some(key) = iter.key() {
             if let Some(value) = iter.value() {
```
MathieuDutSik: Here we have a suboptimal code.
We process the `.key()` to determine if we are done or not and then get the value. Then we use the `.value()`. But actually, if it is something, then this one should be some.

I think the right way is to use `.item()`, which will get you the key/value at once or none at all. So, this would simplify the code.

MathieuDutSik (commented):
If that easy and sensible, I would take the opportunity to switch to version 0.24.0.

ndr-ds:
I can do a version upgrade in a separate PR, don't wanna mix too many things here

MathieuDutSik (approved):
Thanks for the upgrade!

PR #4493, opened 2025-09-05
Clients update validators about missing committees.

### LINKED ISSUE #4490: Clients should update validators about missing committees.
In practice this should never happen, because validators should always have the latest admin chain (see #4488), but clients should still need to handle this case:

If a validator doesn't have the latest committee and I send it a block signed by that committee, it will try to read the committee blob and fail. But the committee blob does not (and must not!) count as a "required blob" of that block, since the block is not enough to re-confirm it (because it is already signed by _that_ committee).

So if a validator responds with `BlobsNotFound`, we need to handle both cases, not just the first:
* blobs that are required and re-certified by the certificate we sent, and
* the committee blob!

In the first case, we already send the missing blobs, and that's enough. In the second case, we need to update the validator about the admin chain.

afck:
## Motivation

If a validator doesn't have the latest committee and receives a block signed by that committee, it will fail with `EventsNotFound`.

Also, the validator updater is sending the wrong blocks in some cases for sparse event publisher chains.

## Proposal

Handle that in the client and update the validator about the admin chain. Use the correct block heights.

Almost all the credit for this goes to @bart-linera!

## Test Plan

A test was added that reproduced the issue.

## Release Plan

    - be released in a new SDK.
    - be released in a new SDK.

## Links

- Closes #4490.

### DIFF
--- a/linera-core/src/updater.rs
+++ b/linera-core/src/updater.rs
@@ -446,15 +446,19 @@ where
         };
         // Figure out which certificates this validator is missing. In many cases, it's just the
         // last one, so we optimistically send that one right away.
-        let hash = {
-            let chain = self.local_node.chain_state_view(chain_id).await?;
-            let hashes = chain.block_hashes(height..=height).await?;
-            hashes.into_iter().next().ok_or_else(|| {
+        let hash = self
+            .local_node
+            .chain_state_view(chain_id)
+            .await?
+            .block_hashes(height..=height)
+            .await?
+            .into_iter()
+            .next()
+            .ok_or_else(|| {
                 ChainClientError::InternalError(
                     "send_chain_information called with invalid target_block_height",
                 )
-            })?
-        };
+            })?;
         let certificate = self
             .local_node
             .storage_client()


linera-core/src/updater.rs
```diff
@@ -221,25 +222,39 @@ where
...
-                let maybe_blobs = self.local_node.read_blobs_from_storage(blob_ids).await?;
-                let blobs = maybe_blobs.ok_or_else(|| original_err.clone())?;
-                self.remote_node.node.upload_blobs(blobs.clone()).await?;
-                self.remote_node
-                    .handle_confirmed_certificate(certificate, delivery)
-                    .await
-            }
-            _ => result,
-        }?)
+        for _ in [0, 1] {
```
ma2bd: nit: perhaps `0..2` is more idiomatic?

linera-core/src/updater.rs
```diff
@@ -221,25 +222,39 @@ where
...
+            result = match result {
+                Err(NodeError::EventsNotFound(event_ids))
+                    if event_ids.iter().all(|event_id| {
+                        event_id.stream_id == StreamId::system(EPOCH_STREAM_NAME)
+                    }) && certificate.inner().chain_id() != self.admin_chain_id().await? =>
```
ma2bd: What's special about the admin chain that requires this while other (event-producing) chains don't need it?
Or should generalize this logic to every chain?

linera-core/src/updater.rs
```diff
@@ -436,7 +480,36 @@ where
...
         } else {
             let query = ChainInfoQuery::new(chain_id);
-            self.remote_node.handle_chain_info_query(query).await?
+            match self
+                .remote_node
+                .handle_chain_info_query(query.clone())
+                .await
+            {
+                Ok(info) => info,
+                Err(ref original_err @ NodeError::BlobsNotFound(ref blob_ids)) => {
```
ma2bd: In which scenario(s) does a chain-info-query return a "blob not found" error?

afck: If it doesn't have the chain description blob.

linera-core/src/updater.rs
afck: When handling a confirmed certificate, the validator doesn't actually need to have the received events, and doesn't fail in that case.

The only thing that can cause an `EventsNotFound` is that the validator is missing the committee that signed the certificate.

afck: The extra check is so that the client doesn't get a stack overflow because a validator keeps responding with `EventsNotFound` even for blocks on the admin chain.

Which is what I think I observed, so I need to look into that, too.

linera-core/src/updater.rs
afck: I made the loop more conservative, and check explicitly that the same error is not handled more than once.

linera-core/src/updater.rs
afck: I found another bug that caused us to send the wrong event publisher blocks and that was likely the cause of this.

linera-core/src/updater.rs
afck: This was also only necessary due to the wrong event publisher heights. Fixed now.

linera-chain/src/chain.rs
```diff
@@ -537,6 +537,13 @@ where
         }
     }
 
+    pub async fn next_height_to_preprocess(&self) -> Result<BlockHeight, ChainError> {
```
deuszx, line 543: Can you add a comment with the explanation why do it this way? I've found myself wondering what many of our pub methods do.

linera-core/src/updater.rs
```diff
@@ -329,8 +352,13 @@ where
...
                     for chain_id in new_chain_ids {
-                        let info = self.local_node.chain_info(chain_id).await?;
-                        publisher_heights.insert(chain_id, info.next_block_height);
+                        let height = self
+                            .local_node
+                            .chain_state_view(chain_id)
+                            .await?
+                            .next_height_to_preprocess()
+                            .await?;
+                        publisher_heights.insert(chain_id, height);
```
deuszx, line 365: Ideally, this method wouldn't have to be a `loop` and keep rediscovering new `publisher_chain_ids` that need to be sent. Is the reasoning here that validator will keep telling us about new chains it needs to update before accepting the block proposal?

linera-core/src/updater.rs
```diff
@@ -390,6 +418,29 @@ where
...
+        let local_admin_info = self.local_node.chain_info(admin_id).await?;
+        Box::pin(self.send_chain_information(
+            admin_id,
+            local_admin_info.next_block_height,
+            CrossChainMessageDelivery::NonBlocking,
+        ))
+        .await
+    }
+
+    async fn admin_chain_id(&self) -> Result<ChainId, ChainClientError> {
```
deuszx: Can we cache this? I know it's local but still - it doesn't change.

linera-core/src/updater.rs
afck: I think each match arm is only supposed to happen in at most one of the loop's iterations: It's just different things that could be missing on the validator's side. We could do something similar to what I added in `send_confirmed_certificate`, and verify that.

linera-core/src/updater.rs
```diff
@@ -433,14 +486,13 @@ where
                 );
                 let query = ChainInfoQuery::new(chain_id);
                 self.remote_node.handle_chain_info_query(query).await?
```
deuszx, line 477: This is a follow-up to our discussion – we don't have the block at the specified height so how can we update the remote node about this chain. Maybe we should just leave the logging here? If we don't have the block, we won't be able to send certificates down below (b/c we don't have `target_block_height`, we probably don't have `target_block_height-1` either.

afck: I agree, but IIRC this made a test fail, and it might be orthogonal to this PR, so I'd rather do it in a follow-up.

linera-core/src/updater.rs
```diff
@@ -411,11 +462,13 @@ where
                     .read_certificate(*hash)
                     .await?
                     .ok_or_else(|| ChainClientError::MissingConfirmedBlock(*hash))?;
+                let admin_id = self.admin_chain_id().await?;
                 match self.send_confirmed_certificate(certificate, delivery).await {
                     Err(ChainClientError::RemoteNodeError(NodeError::EventsNotFound(
                         event_ids,
                     ))) if event_ids.iter().all(|event_id| {
-                        event_id.stream_id.application_id == GenericApplicationId::System
+                        event_id.stream_id == StreamId::system(EPOCH_STREAM_NAME)
```
deuszx: Why do you use `.all` and not `.any`?

afck: The only way processing a confirmed block can fail is due to missing committees. (It's like with messages: A _confirmed_ block doesn't fail, even if its received messages weren't seen yet.)

ma2bd: or missing blobs in general?

afck: Sorry, I meant: The only way a confirmed block can fail with `EventsNotFound` is if the events are the committee creation events.

linera-core/src/updater.rs
afck: Done in https://github.com/linera-io/linera-protocol/pull/4493/commits/0c36e38dc7a27537541d9a2e40281b3152acd323.

linera-chain/src/chain.rs
afck: Done in https://github.com/linera-io/linera-protocol/pull/4493/commits/0c36e38dc7a27537541d9a2e40281b3152acd323.

linera-core/src/updater.rs
```diff
@@ -221,25 +222,46 @@ where
...
-            }
-            _ => result,
-        }?)
+        let mut sent_admin_chain = false;
+        let mut sent_blobs = false;
+        loop {
+            result = match result {
+                Err(NodeError::EventsNotFound(event_ids))
+                    if !sent_admin_chain
+                        && certificate.inner().chain_id() != self.admin_id
```
deuszx, line 241: I wonder: is _this_ check necessary? Why do you need to filter out cases where the certificate already-sent was for an admin chain?

linera-core/src/updater.rs
```diff
@@ -221,25 +222,46 @@ where
...
+                    // The validator doesn't have the committee that signed the certificate.
+                    sent_admin_chain = true;
+                    self.update_admin_chain().await?;
+                    self.remote_node
+                        .handle_confirmed_certificate(certificate.clone(), delivery)
+                        .await
+                }
+                Err(NodeError::BlobsNotFound(blob_ids)) if !sent_blobs => {
+                    // The validator is missing the blobs required by the certificate.
+                    sent_blobs = false;
```
deuszx: Did you mean to set `sent_blobs` to true here?

linera-core/src/updater.rs
afck: I just want to defend against a faulty node: If it keeps sending these for _admin_ chain certificates, we'd get a stack overflow!

linera-core/src/updater.rs
afck: Yes!! Good catch! (I had sent this to `false` for debugging. :facepalm:)

afck: Fixed in https://github.com/linera-io/linera-protocol/pull/4493/commits/57b6cc67f2b954b3c006f8cc4d8f8e37ac2a3561.

linera-core/src/updater.rs
deuszx: OK, in that case – can we catch that with a separate match arm and log appropriate error/warning?

linera-core/src/updater.rs
```diff
@@ -221,25 +222,46 @@ where
...
+                    // The validator doesn't have the committee that signed the certificate.
+                    sent_admin_chain = true;
+                    self.update_admin_chain().await?;
+                    self.remote_node
+                        .handle_confirmed_certificate(certificate.clone(), delivery)
+                        .await
+                }
+                Err(NodeError::BlobsNotFound(blob_ids)) if !sent_blobs => {
+                    // The validator is missing the blobs required by the certificate.
+                    sent_blobs = true;
```
deuszx: Maybe it doesn't matter in practice but I'd feel safer if we set it to `true` _after_ we actually upload the blobs. Similarly for the `sent_admin_chain` above.

linera-core/src/updater.rs
afck: But the general error case is meant to catch all other node errors, including all those where the validator is faulty.

afck: I.e. why should we add more match arms for all conceivable validator misbehaviors?

deuszx: Not a general arm but an arm where you know the validator is faulty for this specific reason. B/c in the higher-level logic (i.e. the caller) this context/information is lost.

deuszx: > I.e. why should we add more match arms for all conceivable validator misbehaviors?

so why are you defending from this specific error if there are probably more that you're not detecting?

afck: No, we _are_ "detecting" all other misbehaviors by returning an error from this function.

We introduced this match arm because we realized that _some_ `EventsNotFound` errors actually are _not_ a misbehavior. But that is only the case if the events are of that particular type.

linera-core/src/updater.rs
afck: My thinking was that I wanted to keep it close to where we check its value; but if you prefer I can move it to the bottom.

deuszx: I meant just couple of lines below – after you call `.upload_blobs`.

linera-core/src/updater.rs
deuszx: I don't think we are detecting all other misbehaviors (just the ones that we had thought of) but ok.

My reasoning was that this context – validator is misbehaving b/c he's sending `EventsNotFound` for a chain he should have been _just_ updated about – is not clear when you see the error enum in the logs.

If you're convinced it is, fine.

afck: > I don't think we are detecting all other misbehaviors (just the ones that we had thought of) but ok.

By that I simply meant that for almost all kinds of node errors we simply return an error here and don't keep trying to update that particular validator.

The existing match arms are for those specific errors where we know the node might only be missing information, so we send and retry.

So I consider this a whitelist: If we're missing anything then either the existing match arms should be even stricter (return `Err` in more cases) or there should be more match arms where we send and retry. I don't see the point of adding additional match arms here for errors where we _don't_ try to recover. There's hundreds of conceivable misbehaviors, and for all of them we want to return `Err`.

deuszx: Logs are for when you need to debug and error codes are not enough to understand what is going on.

afck: OK: But then I'd add a log to the general error case here, rather than add lots of different match arms. If you agree, I'll do that.

deuszx: Yes, I think it makes sense to do it here.

linera-core/src/updater.rs
afck: Done in https://github.com/linera-io/linera-protocol/pull/4493/commits/9cc4ed103d2932df9efc454eee22d722dbb59fb9; I also added `instrument` to `send_confirmed_certificate` to log when we return an error.

linera-core/src/updater.rs
afck: Done in https://github.com/linera-io/linera-protocol/commit/9cc4ed103d2932df9efc454eee22d722dbb59fb9 (using `instrument(…err…)`).

linera-core/src/updater.rs
```diff
@@ -433,14 +476,13 @@ where
                 );
                 let query = ChainInfoQuery::new(chain_id);
                 self.remote_node.handle_chain_info_query(query).await?
-            }
+            };
+            (info.next_block_height, info.manager.current_round)
         } else {
-            let query = ChainInfoQuery::new(chain_id);
-            self.remote_node.handle_chain_info_query(query).await?
+            (BlockHeight::ZERO, Round::Fast)
```
deuszx: Just for completeness: we return here if `target_block_height == 0` (b/c `target_block_height.try_sub_one()` returned `None`). We assign `BlockHeight::ZERO` to `remote_height` and then down below create a range: `0..0`.

If the chain is empty, the only thing to do is to make the remote node handle the timeout - if there is one. Why not do it here and return early?

linera-core/src/updater.rs
deuszx: I was wondering what if this fails and there are missing events for other chains than the admin chain (but one of them is for the admin chain). Would it be incorrect if we update the node anyway? If it's not incorrect, why not simplify the code?

linera-core/src/updater.rs
afck: Not sure: That would duplicate a few lines around the `handle_timeout_certificate` call.

linera-core/src/updater.rs
afck: A node that returns any `EventsNotFound` containing any event that isn't from the admin chain here is faulty.

(It's legitimate to return other `EventsNotFound` errors when handling a _proposal_, but not when handling a confirmed block.)

linera-core/src/updater.rs
deuszx: Personally, I'd prefer the duplication than the confusion.

afck: ~Done in https://github.com/linera-io/linera-protocol/pull/4493/commits/c3fbfd65e3be7fa56b1c14f88a519082eff0a142.~

linera-core/src/updater.rs
```diff
@@ -390,57 +422,63 @@ where
...
         // Figure out which certificates this validator is missing. In many cases, it's just the
         // last one, so we optimistically send that one right away.
-        let remote_info = if let Ok(height) = target_block_height.try_sub_one() {
-            let chain = self.local_node.chain_state_view(chain_id).await?;
-            if let Some(hash) = chain
-                .block_hashes(height..target_block_height)
-                .await?
-                .first()
+        let chain = self.local_node.chain_state_view(chain_id).await?;
+        let hashes = chain.block_hashes(height..target_block_height).await?;
```
deuszx: Since we know that `height` can only be `== target_block_height - 1`, maybe `height..=height` would be clearer that there can be only 1 element.

deuszx: And since we never use `hashes` later on, maybe just append the `.into_iter().next()...` to the result of `.await?` and keep `let hash = ` only.

afck: Done in https://github.com/linera-io/linera-protocol/pull/4493/commits/a427941dcc20650138e81c402492734f62e1bec6.

linera-core/src/updater.rs
afck: ~Reverted: It would actually require _a lot_ of duplication, I think.~

afck: ~Done in https://github.com/linera-io/linera-protocol/pull/4493/commits/3cd2d0fd34fcf794a31265e53df9b4142197160f.~

afck: As discussed, I removed this for now, because it makes a test fail. I pushed the branch with the commit here: https://github.com/afck/linera-protocol/tree/issue-4490-with-early-return. Also see https://github.com/linera-io/linera-protocol/issues/4513 (the description is wrong; it fails with both variants).

deuszx (approved):

PR #4495, opened 2025-09-05
Don't set base timeout to u64::MAX on test_end_to_end_listen_for_new_rounds

ndr-ds:
## Motivation

We're setting the base timeout for the chain we're opening to `u64::MAX`. However, since this test is racey by design, if we happen to get a `RoundTimeout` in proposals in both of the chains, we'll effectively wait forever and the test will never return.

## Proposal

Set the base timeout to something more reasonable. Setting to 10 seconds, as it's what the rest of the file is using.

## Test Plan

Had a consistent repro of this happening and the test hanging forever. After this change, the test passes.

## Release Plan

### DIFF
--- a/linera-service/tests/linera_net_tests.rs
+++ b/linera-service/tests/linera_net_tests.rs
@@ -4577,7 +4577,7 @@ async fn test_end_to_end_listen_for_new_rounds(config: impl LineraNetConfig) -> 
             vec![100, 100],
             0,
             Amount::from_tokens(9),
-            u64::MAX,
+            10_000,
         )
         .await?;
     client1.assign(owner1, chain2).await?;


ndr-ds:
* **#4495** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4495?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4495?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ma2bd (approved):

afck (changes requested):
In this scenario there should never be a round timeout, that's the point of this test: both chain owners are live, so one of them should always immediately propose a block.

This test used to work reliably. We need to find the commit that broke it.

ndr-ds:
Fix for the notifications is this: https://github.com/linera-io/linera-protocol/pull/4516
Closing this PR

PR #4496, opened 2025-09-05
Port PR #3682 to main

ma2bd:
## Motivation

Fix dev manual

## Proposal

Port forgotten PR #3682

## Test Plan

CI

## Release Plan

    - be released in a new SDK,
    - be released in a new SDK,

### DIFF
--- a/examples/counter/src/service.rs
+++ b/examples/counter/src/service.rs
@@ -7,7 +7,7 @@ mod state;
 
 use std::sync::Arc;
 
-use async_graphql::{EmptySubscription, Request, Response, Schema, Object};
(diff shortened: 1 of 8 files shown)


examples/counter/src/service.rs
```diff
@@ -49,7 +56,37 @@ impl Service for CounterService {
...
+
+#[Object]
+impl MutationRoot {
+    async fn increment(&self, value: u64) -> [u8; 0] {
+        self.runtime.schedule_operation(&value);
+        []
+    }
+}
+// ANCHOR_END: mutation
+*/
```
afck, line 76: We could actually use this in the `counter` example, instead of the derive.

afck (approved):

PR #4497, opened 2025-09-06
[testnet] Port PR #3682 (mdbook anchors) to the testnet branch

ma2bd:
## Motivation

Fix documentation

## Proposal

Port PR #3682 to the new testnet branch

## Test Plan

CI

## Release Plan

This time, I'm also going to make a PR to main: #4496

### DIFF
--- a/examples/counter/src/contract.rs
+++ b/examples/counter/src/contract.rs
@@ -14,16 +14,20 @@ use linera_sdk::{
 
 use self::state::CounterState;
 
+// ANCHOR: contract_struct
+linera_sdk::contract!(CounterContract);
+
 pub struct CounterContract {
     state: CounterState,
     runtime: ContractRuntime<Self>,
 }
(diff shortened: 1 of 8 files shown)


examples/counter/src/contract.rs
```diff
@@ -14,50 +14,62 @@ use linera_sdk::{
 
 use self::state::CounterState;
 
+// ANCHOR: contract_struct
```
afck, line 17: We could add a comment explaining these, so that a reader new to Linera doesn't think this needs to be in all apps or something.

ma2bd: Where should we have the comment though? (It cannot be on every file)

afck: Or maybe just a short `// Anchors to include code in linera.dev.` on the first anchor in every file? Not sure; maybe it's better without.

PR #4505, opened 2025-09-07
Change the shuffling to take into account the weight.

### LINKED ISSUE #2616: When trying validators in random order, take weights into account.
If a validator has five times less weight its misbehavior should do five times less harm. So whenever we try to obtain something sequentially from the different validators, in order, we should not just `shuffle` the vector: the validators should be picked with a probability proportional to their weights. That way, a validator with very little stake can only cause very few delays.

MathieuDutSik:
## Motivation

Not all validators are equal. Taking into account their weight during the shuffling makes the validators with more weight more responsible for accessing the

Fixes #2616 

## Proposal

We simply count the weights when doing the random shuffling.

## Test Plan

The CI.

## Release Plan

## Links

None.

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -62,7 +62,7 @@ use linera_execution::{
 };
 use linera_storage::{Clock as _, ResultReadCertificates, Storage as _};
 use linera_views::ViewError;
-use rand::prelude::SliceRandom as _;
+use rand::Rng;
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
 use tokio::sync::{mpsc, OwnedRwLockReadGuard};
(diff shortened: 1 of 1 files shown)


linera-core/src/client/mod.rs
```diff
@@ -317,18 +317,63 @@ impl<Env: Environment> Client<Env> {
...
+        for (index, weight) in weights.iter().enumerate() {
+            if let Some(weight) = weight {
+                position += weight;
+                if position >= target {
+                    return (index, *weight);
+                }
+            }
+        }
+        panic!("target could not be reached");
+    }
```
afck: I think this reimplements the functionality of `rand_distr::weighted`?

linera-core/src/client/mod.rs
```diff
@@ -317,18 +317,63 @@ impl<Env: Environment> Client<Env> {
...
+            })
+            .collect::<Vec<_>>();
+        let mut remaining_weight = weights.iter().map(|s| s.unwrap()).sum();
+        let mut weighted_validators = Vec::new();
+        for _ in 0..number_validator {
+            let target = rng.gen_range(0..remaining_weight);
+            let (index, weight) = Self::get_validator_number(&weights, target);
+            remaining_weight -= weight;
+            weights[index] = None;
+            weighted_validators.push(validators[index].clone());
```
afck: Why not make the weights non-optional, and _remove_ the selected weight and validator from the collections. Then we also wouldn't have to clone.

linera-core/src/client/mod.rs
```diff
@@ -317,18 +317,63 @@ impl<Env: Environment> Client<Env> {
...
-        let mut info = self.fetch_chain_info(chain_id, &validators).await?;
-        for remote_node in validators {
+        let (_, committee) = self.admin_committee().await?;
+        let validators = self.make_nodes(&committee)?;
+        // Shuffle validators using their weights
+        let weighted_validators = self.weighted_shuffle_validators(&validators, &committee);
+        let mut info = self
+            .fetch_chain_info(chain_id, &weighted_validators)
+            .await?;
+        for remote_node in weighted_validators {
```
afck: We could avoid shuffling all the validators at once, and instead return and remove a random one in each iteration. Because in most cases only one iteration will be necessary anyway.

MathieuDutSik: That is a good idea.
Unfortunately, it does not work since the random number generators cannot be shared over threads safely, and no RNG of the rand crate can be shared safely.

linera-core/src/client/mod.rs
MathieuDutSik: Yes, I thought this was better memory-wise, but a bad idea it seems.

linera-core/src/client/mod.rs
MathieuDutSik: Yes.
I liked the mathematics of the generation code, and so I coded it.

linera-core/src/client/mod.rs
afck: Instead of `weighted_shuffle`, can't you just do something like `weighted_select`, which removes and returns a validator from the list, and returns `None` once the list is empty? Then you'd replace the loop with:
`while let Some(remote_node) in weighted_select(&mut validators, &mut rng) {`
No need to share the RNG?

MathieuDutSik: Well, yes, of course it would have to be shared.
But, if you really want to make it shareable, then a lock has to be introduced. Which should be doable.

afck: We could get the `thread_rng` inside `weighted_select`, then it wouldn't need to be held across an `await` point. Anyway, not a blocker!

MathieuDutSik: Switch to the `StdRng` which has better shareability possibilities.

afck (approved):

PR #4510, opened 2025-09-08
Restore old behavior: Don't record oracle if blob known from same block.

afck:
## Motivation

#4413 introduced a bug where we record redundant oracle responses:
* `blob_used` itself already records the response, so there is no need to call `replay_oracle_response`.
* If the blob was published or created by this block, there is no reason to record an oracle response either.

## Proposal

Restore the old behavior.

## Test Plan

This caused the remote tests against Testnet Conway to break, and with this fix on the `testnet_conway` branch, they work again.

## Release Plan

## Links

### DIFF
--- a/examples/publish-read-data-blob/README.md
+++ b/examples/publish-read-data-blob/README.md
@@ -1,8 +1,8 @@
 # Publish and read a data blob
 
 This example shows how to create a blob and how to read it.
-This example should be read in conjunction with the end-to-end
-test `test_wasm_end_to_end_publish_read_data_blob`.
+This example should be read in conjunction with the client
+test `run_test_publish_read_data_blob`.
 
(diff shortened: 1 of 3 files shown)


ma2bd:
@afck How hard is it to cover this behavior with a unit test?

afck:
I'll try to extend a test to check for this regression.

I just noticed that the test that originally uncovered the problem was removed in https://github.com/linera-io/linera-protocol/pull/4397, and replaced by a client test (which we wouldn't run against the network). (My bad: I suggested replacing more tests with client tests.)

afck:
Extended the test in https://github.com/linera-io/linera-protocol/pull/4510/commits/3407f6d58da82fc0c25dcf99f7b53eb9c8d72150.

PR #4512, opened 2025-09-08
`linera-web`: improve `package.json`s

Twey:
## Motivation

The package metadata currently points to `linera-web`.

Also, `npm publish --public` doesn't work for new packages, and could be missed by mistake.

## Proposal

Add some stub documentation here in `linera-protocol`, and update the `package.json`s to point to it.

Also, add the public publish option to the config.

## Test Plan

Published it.

## Release Plan

## Links

### DIFF
--- a/linera-web/clippy.toml
+++ b/linera-web/clippy.toml
@@ -1,0 +1,1 @@
+doc-valid-idents = ["MetaMask", ".."]


linera-web/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+# `linera-web`
+
+This module defines the client API for the Web extension.
+
+<!-- cargo-rdme end -->
+
+## Building
+
+Make sure your `cargo` invokes a nightly `rustc`. Then, you can build
```
ma2bd, line 18: Is nightly Rust still needed? I don't think I used it to compile locally last time.

linera-web/README.md
```diff
@@ -0,0 +1,20 @@
+<!-- cargo-rdme start -->
+
+# `linera-web`
+
+This module defines the client API for the Web extension.
```
ma2bd: (and web pages hosting a Linera client)

ma2bd (approved):

linera-web/README.md
ma2bd: Also how about the `signer` package.

ma2bd: Perhaps mention the package names `@linera/client` and `@linera/signer`.

linera-web/README.md
Twey: It is, there's no threading on the Web without nightly.

Twey: (You usually don't need to explicitly select a nightly because there's a rustup configuration file in the directory.)

PR #4515, opened 2025-09-08
[testnet] Clients update validators about missing committees. (#4493)

### LINKED ISSUE #4490: Clients should update validators about missing committees.

deuszx:
## Motivation

If a validator doesn't have the latest committee and receives a block signed by that committee, it will fail with `EventsNotFound`.

Also, the validator updater is sending the wrong blocks in some cases for sparse event publisher chains.

## Proposal

Handle that in the client and update the validator about the admin chain. Use the correct block heights.

Almost all the credit for this goes to @bart-linera!

## Test Plan

A test was added that reproduced the issue.

## Release Plan

    - be released in a new SDK.

## Links

- Closes #4490.

### DIFF
--- a/linera-rpc/tests/snapshots/format__format.yaml.snap
+++ b/linera-rpc/tests/snapshots/format__format.yaml.snap
@@ -365,6 +365,7 @@ ChainInfoQuery:
     - request_sent_certificate_hashes_by_heights:
         SEQ:
           TYPENAME: BlockHeight
+    - create_network_actions: BOOL
 ChainInfoResponse:
   STRUCT:
     - info:
--- a/linera-core/src/chain_worker/state.rs
+++ b/linera-core/src/chain_worker/state.rs
@@ -246,6 +246,7 @@ where
         &mut self,
         query: ChainInfoQuery,
     ) -> Result<(ChainInfoResponse, NetworkActions), WorkerError> {
+        let create_network_actions = query.create_network_actions;
         if let Some((height, round)) = query.request_leader_timeout {
             self.vote_for_leader_timeout(height, round).await?;
         }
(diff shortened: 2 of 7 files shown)


linera-rpc/tests/snapshots/format__format.yaml.snap
```diff
@@ -852,7 +852,7 @@ OracleResponse:
       Event:
         TUPLE:
           - TYPENAME: EventId
-          - SEQ: U8
```
afck, line 855: Right, this changes the format; we could omit the `serde_bytes` part and only keep the `hex_debug` one?

afck: (…or upgrade all validators, of course.)

deuszx: good eye!

deuszx: @ma2bd do you have strong opinion here? The remote tests passed so this is a change that is compatible with current testnet 🤔 I'm not really sure why they're green since this is changing the format.

ma2bd: This is fine. BYTES and SEQ[u8] have the same BCS and bincode encodings.

PR #4518, opened 2025-09-08
[testnet] Control creation of network actions when handling `ChainInfoQuery`.

deuszx:
## Motivation

Creating network actions requires reading certificates – this takes a lot of CPU time (reading from storage, deserializing certificates, etc.) but it's not always needed. Network actions are created in couple of places but one surprising one was when handling a `ChainInfoQuery`.

## Proposal

Add a boolean field to `ChainInfoQuery` struct that controls whether the caller wants to create network actions. By default it is set to `true` to maintain backwards compatibility but clients can call `no_network_actions` to set it to false.

## Test Plan

CI.

## Release Plan

This is already a testnet backport.

## Links

### DIFF
--- a/linera-core/src/data_types.rs
+++ b/linera-core/src/data_types.rs
@@ -167,6 +167,11 @@ impl ChainInfoQuery {
         self.request_fallback = true;
         self
     }
+
+    pub fn with_network_actions(mut self) -> Self {
+        self.create_network_actions = true;
+        self
+    }
 }
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
--- a/linera-core/src/local_node.rs
+++ b/linera-core/src/local_node.rs
@@ -257,7 +257,7 @@ where
         let (_response, actions) = self
             .node
             .state
-            .handle_chain_info_query(ChainInfoQuery::new(sender_chain))
+            .handle_chain_info_query(ChainInfoQuery::new(sender_chain).with_network_actions())
             .await?;
         let mut requests = VecDeque::from_iter(actions.cross_chain_requests);
         while let Some(request) = requests.pop_front() {


linera-rpc/src/grpc/conversions.rs
```diff
@@ -614,6 +614,7 @@ impl TryFrom<api::ChainInfoQuery> for ChainInfoQuery {
             request_fallback: chain_info_query.request_fallback,
             request_sent_certificate_hashes_by_heights,
             request_sent_certificate_hashes_in_range: None,
+            create_network_actions: chain_info_query.create_network_actions.unwrap_or(true),
```
afck, line 617: This is what maintains backwards compatibility, isn't it?

The default value in the _code_ might as well be `false`? That would save a lot of `no_network_actions` calls, and replace them with only one or two `with_network_actions` calls?

afck:
I'm okay with this approach for now, but in the long term I'd prefer something that doesn't need a new field: It weird that the client has to tell the validator whether to retry some internal messages. An alternative might be to cache in memory in the validator whether all messages have been delivered and/or when they have been sent.

linera-rpc/src/grpc/conversions.rs
deuszx: Yes, I am actually in the process of doing that.

deuszx:
> I'm okay with this approach for now, but in the long term I'd prefer something that doesn't need a new field: It weird that the client has to tell the validator whether to retry some internal messages. An alternative might be to cache in memory in the validator whether all messages have been delivered and/or when they have been sent.

Yes, I agree that approach would be cleaner. It would probably be slightly slower AND – in some cases – also didn't achieve the same result b/c there might be some cross-chain messages to send when handling _pure_ `ChainInfoQuery` - i.e. one that doesn't do that with the current approach.

afck:
On the other hand with the in-memory information it may not have to create network actions in some cases where the current approach _would_ request to create them, because the server knows best which messages are still pending.

linera-rpc/src/grpc/conversions.rs
deuszx: Done in https://github.com/linera-io/linera-protocol/pull/4518/commits/db3b12c89820ab90f8507febbb5f8d70cd243c04

afck (commented):
Now I can't find the place where you _do_ set `create_network_actions` to true on the client side. :sweat_smile:

afck (approved):

PR #4521, opened 2025-09-09
Move `large_futures` config to `Cargo.toml`; deny in more crates.

afck:
## Motivation

Large futures increase the risk of stack overflows, which is why we `#[deny(clippy::large_futures)]` in several places. However, this covers only a small fraction of the code.

## Proposal

Move the lint config to `Cargo.toml` and use workspace configuration in other crates. Fix large future warnings in some places.

## Test Plan

CI

## Release Plan

## Links

### DIFF
--- a/linera-chain/Cargo.toml
+++ b/linera-chain/Cargo.toml
@@ -11,6 +11,9 @@ license.workspace = true
 repository.workspace = true
 version.workspace = true
 
+[lints]
+workspace = true
+
 [features]
 metrics = ["prometheus", "linera-views/metrics", "linera-execution/metrics"]
 test = [
--- a/linera-chain/src/lib.rs
+++ b/linera-chain/src/lib.rs
@@ -3,8 +3,6 @@
 
 //! This module manages the state of a Linera chain, including cross-chain communication.
 
-#![deny(clippy::large_futures)]
-
 pub mod block;
 mod certificate;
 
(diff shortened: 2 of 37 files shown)


linera-views/tests/views_tests.rs
```diff
@@ -310,305 +310,308 @@ async fn test_store<S>(
...
-        }
-        if config.with_queue {
-            assert_eq!(view.queue.read_front(10).await?, vec![7]);
-        }
-        if config.with_map {
-            assert_eq!(view.map.get("Hello").await?, Some(5));
-        }
-        if config.with_set {
-            assert!(view.set.contains(&42).await?);
-        }
```
deuszx, line 386: Did you remove those?

afck: ~Wow, good catch! Looks like I did, accidentally!~ I only meant to `Box::pin` everything!

afck: No, wait, they are still there, aren't they? It's only that the diff is displayed in a weird way.

deuszx: Yes, very weird. Difficult to check if everything is kept

afck: I double-checked using meld: No lines were removed, only wrapped in `Box::pin`.

PR #4522, opened 2025-09-09
Rm serde bytes from Event variant of OracleResponse

deuszx:
## Motivation

https://github.com/linera-io/linera-protocol/pull/4518 changed the format for `Event` variant of the `OracleResponse`.

## Proposal

Even though the remote tests against testnet conway passed, we can err on the side of caution and revert back that change.

## Test Plan

CI

## Release Plan

- Nothing to do / These changes follow the usual release cycle (we're already on testnet branch)

### DIFF
--- a/linera-base/src/data_types.rs
+++ b/linera-base/src/data_types.rs
@@ -1044,12 +1044,7 @@ pub enum OracleResponse {
     /// The block's validation round.
     Round(Option<u32>),
     /// An event was read.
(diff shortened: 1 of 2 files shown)


afck (approved):

ma2bd:
This should not be needed but up to you guys

deuszx:
Closing since the network is already updated with this.

PR #4523, opened 2025-09-09
Control creation of network actions when handling `ChainInfoQuery`. 

deuszx:
## Motivation

Creating network actions requires reading certificates – this takes a lot of CPU time (reading from storage, deserializing certificates, etc.) but it's not always needed. Network actions are created in couple of places but one surprising one was when handling a `ChainInfoQuery`.

## Proposal

Add a boolean field to `ChainInfoQuery` struct that controls whether the caller wants to create network actions. By default it is set to `true` to maintain backwards compatibility but clients can call `no_network_actions` to set it to false.

## Test Plan

CI.

## Release Plan

This is already backported to testnet (https://github.com/linera-io/linera-protocol/pull/4518)

## Links

### DIFF
--- a/linera-core/src/chain_worker/state.rs
+++ b/linera-core/src/chain_worker/state.rs
@@ -246,6 +246,7 @@ where
         &mut self,
         query: ChainInfoQuery,
     ) -> Result<(ChainInfoResponse, NetworkActions), WorkerError> {
+        let create_network_actions = query.create_network_actions;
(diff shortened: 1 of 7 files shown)


afck:
Let's link to https://github.com/linera-io/linera-protocol/pull/4518 in the description.

afck (approved):

ma2bd:
@deuszx I would recommend always using `false` as default value for booleans. (Here this means changing the field name)

PR #4530, opened 2025-09-10
Implement a function `has_empty_storage` in `BaseRuntime`.

MathieuDutSik:
## Motivation

For the purpose of testing if a contract exists in EVM, we introduce the functionality of testing if a contract
has empty storage.

## Proposal

The following is done:
* The functionality is added to the `BaseRuntime`.
* It uses the `TotalStorageSize` in the `execution_state_actor.rs`.
* It uses the `view.total_size()` of the `KeyValueStoreView`.
* The functionality is provided to the Wasm smart contracts.

Several points have to be noted:
* The Wasm smart contracts are persistent. Only at the end with the `self.state.save().await` does their state become observable. This point is exemplified in the example test.
* So, if a contract is writing its state only at the `fn finalize` step, then we cannot see that its state is non-trivial only at the next block.
* The designed use case for this feature is EVM smart contracts. Fortunately, for those contracts, we are writing the state directly to the storage and we are not waiting for `fn finalize`. So, that is fine here.

## Test Plan

The CI.

For testing the `has_empty_storage`, we implemented separately some end-to-end tests.
They have not been put in the PR, since you need two contracts:
* one that gets instantiated
* and one that does the `has_empty_storage`.

## Release Plan

This does not appear to be TestNet breaking.

## Links

The test is implemented in [test_has_empty_state](https://github.com/MathieuDutSik/linera_end_to_end_tests/tree/main/test_has_empty_state)

### DIFF
--- a/linera-execution/src/execution_state_actor.rs
+++ b/linera-execution/src/execution_state_actor.rs
@@ -681,7 +681,6 @@ where
                     }
                     None => (0, 0),
                 };
-                tracing::info!("TotalStorageSize, result={result:?}");
                 callback.respond(result);
             }
         }
--- a/linera-execution/src/runtime.rs
+++ b/linera-execution/src/runtime.rs
@@ -898,7 +898,7 @@ where
             .recv_response()
     }
 
-    fn has_trivial_storage(&mut self, application: ApplicationId) -> Result<bool, ExecutionError> {
+    fn has_empty_storage(&mut self, application: ApplicationId) -> Result<bool, ExecutionError> {
         let this = self.inner();
         let (key_size, value_size) = this
             .execution_state_sender
--- a/linera-execution/src/lib.rs
+++ b/linera-execution/src/lib.rs
@@ -736,7 +736,7 @@ pub trait BaseRuntime {
     fn assert_data_blob_exists(&mut self, hash: DataBlobHash) -> Result<(), ExecutionError>;
 
     /// Returns true if the corresponding contract uses a zero amount of storage.
-    fn has_trivial_storage(&mut self, application: ApplicationId) -> Result<bool, ExecutionError>;
+    fn has_empty_storage(&mut self, application: ApplicationId) -> Result<bool, ExecutionError>;
 }
 
 pub trait ServiceRuntime: BaseRuntime {
--- a/linera-execution/src/wasm/runtime_api.rs
+++ b/linera-execution/src/wasm/runtime_api.rs
@@ -671,14 +671,14 @@ where
     }
 
     /// Returns true if the corresponding contract uses a zero amount of storage.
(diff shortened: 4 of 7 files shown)


linera-execution/src/runtime.rs
```diff
@@ -897,6 +897,18 @@ where
             .send_request(|callback| ExecutionRequest::AssertBlobExists { blob_id, callback })?
             .recv_response()
     }
+
+    fn has_trivial_storage(&mut self, application: ApplicationId) -> Result<bool, ExecutionError> {
```
deuszx: "trivial" is very ambiguous. Aren't you checking if its storage is empty/zero? Maybe "is_storage_initialized"?

MathieuDutSik: "zero" storage, that is having a key of value zero value would get you a non-zero size. However, if you read a view with the keys missing, then it would be as if it was zero. So, "is_storage_zero" does not fit.

Some smart contracts will have no storage (e.g. native-fungible). The `fn load` may modify the state. Or not. So, the initialization will create some data or not. So, the "is_storage_initialized" does not fit.

Saying that the storage is empty would be quite adequate. So, would you prefer "has_empty_storage" ?

linera-execution/src/execution_state_actor.rs
```diff
@@ -668,6 +668,22 @@ where
...
+            } => {
+                let view = self.state.users.try_load_entry(&application).await?;
+                let result = match view {
+                    Some(view) => {
+                        let total_size = view.total_size();
+                        (total_size.key, total_size.value)
+                    }
+                    None => (0, 0),
+                };
+                tracing::info!("TotalStorageSize, result={result:?}");
```
deuszx: leftover?

linera-execution/src/runtime.rs
deuszx: Yes, I think `has_empty_storage` conveys the purpose better.

linera-execution/src/execution_state_actor.rs
MathieuDutSik: Yes, sorry.

linera-execution/src/runtime.rs
MathieuDutSik: Ok, renamed.

deuszx: Can you update the PR description and title? They go into commit message after merge.

ma2bd (approved):
LGTM. Should we test `has_empty_storage` in CI?

MathieuDutSik:
> LGTM. Should we test `has_empty_storage` in CI?

It is a little expensive, since we have to insert two smart contracts.
Better to let me check from another repo.

PR #4531, opened 2025-09-10
Fix some pedantic Clippy warnings.

afck:
## Motivation

Many of the non-default Clippy lints in the "pedantic" category are actually useful, especially the "don't re-implement the standard library" ones.

## Proposal

Fix many pedantic warnings. Deny some of the lints.

## Test Plan

CI

## Release Plan

- (But could be backported.)

## Links


linera-base/src/data_types.rs
```diff
@@ -1459,7 +1457,7 @@ impl Blob {
     }
 
     /// Loads data blob from a file.
-    pub async fn load_data_blob_from_file(path: impl AsRef<Path>) -> io::Result<Self> {
+    pub fn load_data_blob_from_file(path: impl AsRef<Path>) -> io::Result<Self> {
```
afck, line 1460: I wonder whether we should use async I/O functions here instead.

afck:
> Do you know how close we are to being able to enable blanket pedantic?

I suspect we're somewhat close to that, but we'd still need to add a handful of `allow` attributes. We could give it a try now.

PR #4534, opened 2025-09-10
Change the way keys are written down in `keystore.json`.

### LINKED ISSUE #4489: `keystore.json` stores keys as lists of decimal numbers.
Maybe these should be hexadecimal.

MathieuDutSik:
## Motivation

Currently, the `keystore.json` encode the key as a vector of decimal values.

Fixes #4489 

## Proposal

We simply change the ways that the encoding is done.
We go from
```
{
  "keys": [
    [
      "0x39b7028692e9a61d2e2ad945bf8e83f25cdfc3cbec6470db6f1ad6fbaa1600ca",
     "7b2245643235353139223a2261326332363133343864366266663965303362376633303638643031616563656363326538343864343661316335353761353030653264376530333264326365227d"
    ],
```

to
```
{
  "keys": [
    [
      "0x39b7028692e9a61d2e2ad945bf8e83f25cdfc3cbec6470db6f1ad6fbaa1600ca",
      "7b2245643235353139223a2261326332363133343864366266663965303362376633303638643031616563656363326538343864343661316335353761353030653264376530333264326365227d"
    ],
```

## Test Plan

The CI.
The remote end-to-end test will test this feature.

## Release Plan

This definitely breaks the existing wallets.

## Links

None.

### DIFF
--- a/linera-base/src/crypto/signer.rs
+++ b/linera-base/src/crypto/signer.rs
@@ -120,7 +120,7 @@ mod in_mem {
         }
 
         /// Returns the public key corresponding to the given `owner`.
-        pub fn keys(&self) -> Vec<(AccountOwner, Vec<u8>)> {
+        pub fn keys(&self) -> Vec<(AccountOwner, String)> {
             let inner = self.0.read().unwrap();
             inner.keys()
         }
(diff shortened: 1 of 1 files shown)


ma2bd (approved):
Thanks. Definitely useful

ma2bd:
> Can we keep the API to work with `Vec<u8>` and only store as hex?

Same question

MathieuDutSik:
> Can we keep the API to work with `Vec<u8>` and only store as hex?

Resolved.

MathieuDutSik:
> > Can we keep the API to work with `Vec<u8>` and only store as hex?
> 
> Same question

Resolved.

PR #4536, opened 2025-09-10
`examples/counter`: keep the name `value` for the operation argument

Twey:
## Motivation

PR #4400 made the counter example's input an enum instead of a struct, which is closer to what people are likely to want to use when extending the example.  Unfortunately, as a side effect, the name of the value to increment by in GraphQL changed from `value` to `field0`.

## Proposal

Use a struct variant for the enum instead of a tuple variant, so that the argument name is reverted to `value`.

## Test Plan

CI should catch regressions.

## Release Plan

    - be released in a new SDK

## Links


Twey:
Obviated by #4538.

PR #4538, opened 2025-09-10
Streamline the use of GraphQLMutationRoot.

MathieuDutSik:
## Motivation

The `GraphQLMutationRoot` is used for some smart contracts like `non-fungible` when it is not needed.

## Proposal

There are two ways to implement the trait `MutationRoot`:
* One is to use the function `mutation_root` of the trait `GraphQLMutationRoot` implemented by the derive macro. For example, `counter`.
* Implement the trait explicitly. For example, `non-fungible`.

Therefore, two changes are made:
* When the second strategy is used, we should not implement the `GraphQLMutationRoot` (3 cases removed).
* When the trait is being used, we formulate it as `GraphQLMutationRoot as _` which is clearer (and examples are supposed to help).

## Test Plan

The CI.

## Release Plan

It does not break anything.

## Links

None.

### DIFF
--- a/examples/meta-counter/src/contract.rs
+++ b/examples/meta-counter/src/contract.rs
@@ -95,7 +95,7 @@ impl Contract for MetaCounterContract {
             Message::Increment(value) => {
                 let counter_id = self.counter_id();
                 log::trace!("executing {} via {:?}", value, counter_id);
-                let operation = counter::CounterOperation::Increment(value);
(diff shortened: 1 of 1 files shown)


ma2bd:
For counter, I would avoid the macro

MathieuDutSik:
> For counter, I would avoid the macro

Ok, updated accordingly.

Twey (approved):

PR #4539, opened 2025-09-10
Add a test demonstrating how to pass multiple mutations in the `node_service`.

MathieuDutSik:
## Motivation

It is quite reasonable for users to want to do several operations in just one block.

## Proposal

The solution is simply to formulate the multiple mutations by using a construction of the GraphQL queries.

## Test Plan

In order to have multiple operations, an operation `repeated_transfer` was added to the `NativeFungibleApp`.
The test `test_wasm_end_to_end_same_wallet_fungible` was modified accordingly.

## Release Plan

The change is fully compatible with the existing codebase.

## Links

None.

### DIFF
--- a/linera-service/tests/linera_net_tests.rs
+++ b/linera-service/tests/linera_net_tests.rs
@@ -160,7 +160,7 @@ impl NativeFungibleApp {
         account_owner: &AccountOwner,
         amount_transfer: Amount,
         destination: Account,
-        num_operation: usize,
+        num_operations: usize,
     ) -> Value {
         let mutation = format!(
             "transfer(owner: {}, amount: \"{}\", targetAccount: {})",
@@ -168,10 +168,7 @@ impl NativeFungibleApp {
             amount_transfer,
             destination.to_value(),
         );
-        let mut mutations = Vec::new();
-        for _ in 0..num_operation {
-            mutations.push(mutation.clone());
-        }
+        let mutations = vec![mutation; num_operations];
         self.0.multiple_mutate(&mutations).await.unwrap()
     }
 


linera-service/tests/linera_net_tests.rs
```diff
@@ -155,6 +155,26 @@ impl NativeFungibleApp {
...
+            "transfer(owner: {}, amount: \"{}\", targetAccount: {})",
+            account_owner.to_value(),
+            amount_transfer,
+            destination.to_value(),
+        );
+        let mut mutations = Vec::new();
+        for _ in 0..num_operation {
+            mutations.push(mutation.clone());
+        }
+        self.0.multiple_mutate(&mutations).await.unwrap()
```
ma2bd, line 172: I believe `vec![mutation; num_operations]` does the job

linera-service/tests/linera_net_tests.rs
```diff
@@ -155,6 +155,26 @@ impl NativeFungibleApp {
         self.0.mutate(mutation).await.unwrap()
     }
 
+    async fn repeated_transfer(
+        &self,
+        account_owner: &AccountOwner,
+        amount_transfer: Amount,
+        destination: Account,
+        num_operation: usize,
```
ma2bd: `num_operations`

linera-service/tests/linera_net_tests.rs
MathieuDutSik: Thanks, done.

linera-service/tests/linera_net_tests.rs
MathieuDutSik: Thanks, done.

linera-service/src/cli_wrappers/wallet.rs
```diff
@@ -1652,6 +1652,15 @@ impl<A> ApplicationWrapper<A> {
         self.run_graphql_query(&format!("mutation {{ {mutation} }}"))
             .await
     }
+
+    pub async fn multiple_mutate(&self, mutations: &[String]) -> Result<Value> {
+        let mut out = String::from("mutation {\n");
+        for (index, mutation) in mutations.iter().enumerate() {
+            out = format!("{}  u{}: {}\n", out, index, mutation);
```
afck, line 1659:
```suggestion
            write!(&mut out, "  u{index}: {mutation}\n")?;
```
(Or `unwrap` or `expect`, since we write to a string. Or `push_str`.)

(What does the "u" stand for?)

afck:
So this was already working, right? Maybe change the title, since only a test was added.

PR #4540, opened 2025-09-11
Bump versions to 0.15.2

deuszx:
## Motivation

Prepare for the release of new SDK.

## Proposal

Bump versions across all toml files to 0.15.2

## Test Plan

CI

## Release Plan

- Release
- 
## Links

### DIFF
--- a/examples/Cargo.lock
+++ b/examples/Cargo.lock
@@ -3694,7 +3694,7 @@ checksum = "4ec2a862134d2a7d32d7983ddcdd1c4923530833c9f2ea1a44fc5fa473989058"
 
 [[package]]
 name = "linera-base"
-version = "0.15.1"
+version = "0.15.2"
 dependencies = [
  "alloy-primitives",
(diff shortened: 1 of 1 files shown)


linera-web/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "linera-web"
 description = "The Linera Web client"
-version = "0.15.1"
+version = "0.15.2"
```
afck: Shouldn't we use `version.workspace = true` here, too?

deuszx: We can do that. In the future though I'd imagine that our clients and backends can be versioned separately.

deuszx: Done in [linera-web uses workspace version](https://github.com/linera-io/linera-protocol/pull/4540/commits/1826dce5db356daca575d32363b48124f4570ae4)

PR #4498, opened 2025-09-06
Continuous memory profiling

ndr-ds:
## Motivation

`jemalloc` is a more performant allocator (we do see a small performance improvement in benchmarks from it), and if we use it it's easier for us to have memory profiles.

## Proposal

This adds support for using `jemalloc` as an allocator, as well as expose a couple endpoints:
* One that generates a pprof and returns it once the endpoint is queried
* One that generates a flamegraph and returns it once the endpoint is queried

## Test Plan

Deployed a network with this and the `linera-infra` changes, saw the memory flamegraphs on Pyroscope, saw the memory profile Grafana dashboard, and managed to curl the flamegraph endpoint and get a memory profile svg returned to me.

## Release Plan

- Nothing to do / These changes follow the usual release cycle.

PR #4544, opened 2025-09-11
Decrease block cache size, increase execution state cache size

ndr-ds:
## Motivation

After a while, while benchmarking, some of our shards start OOMing. After inspecting memory profiles, it seems that the culprit is the `ValueCache` for blocks specifically.

## Proposal

Lower the size of the cache so that it fills up and start evicting entries before making the VM run out of memory.

## Test Plan

Deploy a network with this code, benchmark it for a while and make sure that memory goes flat

## Release Plan

### DIFF
--- a/linera-core/src/unit_tests/value_cache_tests.rs
+++ b/linera-core/src/unit_tests/value_cache_tests.rs
@@ -11,12 +11,15 @@ use linera_base::{
 };
 use linera_chain::types::Timeout;
 
-use super::{ValueCache, DEFAULT_VALUE_CACHE_SIZE};
+use super::ValueCache;
+
+/// Test cache size for unit tests.
+const TEST_CACHE_SIZE: usize = 10;
 
 /// Tests attempt to retrieve non-existent value.
 #[test]
 fn test_retrieve_missing_value() {
-    let cache = ValueCache::<CryptoHash, Hashed<Timeout>>::default();
+    let cache = ValueCache::<CryptoHash, Hashed<Timeout>>::new(TEST_CACHE_SIZE);
     let hash = CryptoHash::test_hash("Missing value");
 
     assert!(cache.get(&hash).is_none());
@@ -26,7 +29,7 @@ fn test_retrieve_missing_value() {
 /// Tests inserting a certificate value in the cache.
 #[test]
 fn test_insert_single_certificate_value() {
-    let cache = ValueCache::<CryptoHash, Hashed<Timeout>>::default();
+    let cache = ValueCache::<CryptoHash, Hashed<Timeout>>::new(TEST_CACHE_SIZE);
     let value = create_dummy_certificate_value(0);
     let hash = value.hash();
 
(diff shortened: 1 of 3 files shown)


ndr-ds:
* **#4559** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4559?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4555** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4555?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4556** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4556?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4554** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4554?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4544** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4544?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4544?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#4573** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4573?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4572** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4572?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4571** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4571?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4487** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4487?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#4498](https://github.com/linera-io/linera-protocol/pull/4498) <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4498?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#4469** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4469?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

linera-core/src/worker.rs
```diff
@@ -43,6 +43,9 @@ use crate::{
     value_cache::ValueCache,
 };
 
+const BLOCK_CACHE_SIZE: usize = 5_000;
```
afck, line 46: If it really OOMed with 10_000 I'd do more than just halve it. I'd go down to 1_000 or even 100 maybe?

linera-core/src/unit_tests/value_cache_tests.rs
```diff
@@ -11,12 +11,15 @@ use linera_base::{
 };
 use linera_chain::types::Timeout;
 
-use super::{ValueCache, DEFAULT_VALUE_CACHE_SIZE};
+use super::ValueCache;
+
+/// Test cache size for unit tests.
+const TEST_CACHE_SIZE: usize = 1000;
```
afck: Why not 10? Does it really make a difference in any of the tests? Otherwise it will just make the tests slower.

deuszx:
We already have size-based cache in the repository: [BlockCache](https://github.com/linera-io/linera-protocol/blob/main/linera-service/src/exporter/storage.rs#L46) in the `linera-service/exporter`. It uses `quick_cache` [Cache](https://docs.rs/quick_cache/0.6.16/quick_cache/sync/struct.Cache.html) with a custom weighter [BlockCacheWeighter](https://github.com/linera-io/linera-protocol/blob/main/linera-service/src/exporter/storage.rs#L513-L520). 

Maybe as a follow-up though.

linera-core/src/worker.rs
ndr-ds: In principle I agree, but the more we lower this the more we seem to be penalized in performance... So 5k seems like a good compromise

linera-core/src/unit_tests/value_cache_tests.rs
ndr-ds: True, 10 should be fine. Keep in mind that the slowness we would get from this is from the initial memory allocation only, because if it's some long running test it would make the test faster, actually, as more things would be cached.
For these tests that trade-off doesn't seem to make sense though, so 10 indeed makes more sense.

PR #4547, opened 2025-09-11
Return error if we try to initialize storage with new genesis

deuszx:
## Motivation

https://github.com/linera-io/linera-protocol/pull/4543 started treating storage initialization as idempotent but it is still useful to detect if we try to initialize it for a network that has a different Genesis config.

## Proposal

Return (and log) an error if the new genesis config is different from the one our storage is initialized with.

## Test Plan

CI

## Release Plan

    - be released in a new SDK,

## Links
https://github.com/linera-io/linera-protocol/pull/4543

### DIFF
--- a/linera-client/src/config.rs
+++ b/linera-client/src/config.rs
@@ -169,6 +169,15 @@ impl GenesisConfig {
             .await
             .map_err(linera_chain::ChainError::from)?
         {
+            if description != self.network_description() {
+                // We can't initialize storage with a different network description.
+                tracing::error!(
+                    current_network=?description,
+                    new_network=?self.network_description(),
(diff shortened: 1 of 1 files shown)


linera-client/src/config.rs
```diff
@@ -30,6 +30,8 @@ pub enum Error {
     Chain(#[from] linera_chain::ChainError),
     #[error("persistence error: {0}")]
     Persistence(Box<dyn std::error::Error + Send + Sync>),
+    #[error("storage is already initialized: {0:?}")]
```
ma2bd, line 33: maybe add `for another network: {1}` ?

PR #4555, opened 2025-09-12
Add more instrumentation

ndr-ds:
## Motivation

Now that we have distributed tracing (after https://github.com/linera-io/linera-protocol/pull/4556), we need more instrumentation so we have data about more functions in the breakdowns.

## Proposal

Instrument more functions with `telemetry_only` so that we don't get spammed in our logs, but the spans still get sent to Tempo.

## Test Plan

Tested this with https://github.com/linera-io/linera-protocol/pull/4556, saw the spans properly show in the breakdowns.

## Release Plan

### DIFF
--- a/linera-chain/src/chain.rs
+++ b/linera-chain/src/chain.rs
@@ -35,6 +35,7 @@ use linera_views::{
     views::{ClonableView, CryptoHashView, RootView, View},
 };
 use serde::{Deserialize, Serialize};
+use tracing::instrument;
 
 use crate::{
     block::{Block, ConfirmedBlock},
@@ -391,6 +392,9 @@ where
         self.context().extra().chain_id()
     }
 
+    #[instrument(target = "telemetry_only", skip_all, fields(
+        chain_id = %self.chain_id(),
+    ))]
     pub async fn query_application(
         &mut self,
         local_time: Timestamp,
@@ -408,6 +412,10 @@ where
             .with_execution_context(ChainExecutionContext::Query)
     }
 
+    #[instrument(target = "telemetry_only", skip_all, fields(
+        chain_id = %self.chain_id(),
+        application_id = %application_id
+    ))]
     pub async fn describe_application(
         &mut self,
         application_id: ApplicationId,
@@ -419,6 +427,11 @@ where
             .with_execution_context(ChainExecutionContext::DescribeApplication)
     }
 
+    #[instrument(target = "telemetry_only", skip_all, fields(
+        chain_id = %self.chain_id(),
+        target = %target,
+        height = %height
+    ))]
     pub async fn mark_messages_as_received(
         &mut self,
         target: &ChainId,
@@ -506,6 +519,9 @@ where
 
     /// Verifies that this chain is up-to-date and all the messages executed ahead of time
     /// have been properly received by now.
+    #[instrument(target = "telemetry_only", skip_all, fields(
+        chain_id = %self.chain_id()
+    ))]
     pub async fn validate_incoming_bundles(&self) -> Result<(), ChainError> {
         let chain_id = self.chain_id();
         let pairs = self.inboxes.try_load_all_entries().await?;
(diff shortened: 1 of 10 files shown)


ndr-ds:
* **#4554** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4554?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4559** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4559?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4555** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4555?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4555?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#4556** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4556?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

linera-chain/src/chain.rs
```diff
@@ -391,6 +392,10 @@ where
         self.context().extra().chain_id()
     }
 
+    #[instrument(target = "telemetry_only", skip_all, fields(
+        chain_id = %self.chain_id(),
+        next_block_height = %self.tip_state.get().next_block_height
```
deuszx: I think the block height is unnecessary detail, it will obscure the trace.

linera-chain/src/chain.rs
```diff
@@ -419,6 +428,11 @@ where
             .with_execution_context(ChainExecutionContext::DescribeApplication)
     }
 
+    #[instrument(target = "telemetry_only", skip_all, fields(
+        chain_id = %self.chain_id(),
+        target = %target,
```
deuszx, line 432: We should really decide how we label things. I've seen things like _nickname_ , _target_, etc. and the are really NOT helping.

linera-chain/src/chain.rs
```diff
@@ -661,6 +683,10 @@ where
     }
 
     /// Removes the incoming message bundles in the block from the inboxes.
+    #[instrument(target = "telemetry_only", skip_all, fields(
+        chain_id = %self.chain_id(),
+        timestamp = %timestamp
```
deuszx: I'd drop `timestamp` from there.

linera-core/src/chain_worker/actor.rs
```diff
@@ -243,8 +243,9 @@ where
 
     /// Runs the worker until there are no more incoming requests.
     #[instrument(
+        target = "telemetry_only",
```
deuszx: So we won't see this in the normal logs anymore?

linera-core/src/chain_worker/actor.rs
```diff
@@ -287,7 +291,7 @@ where
                         let Some((request, span)) = maybe_request else {
                             break; // Request sender was dropped.
                         };
-                        Box::pin(worker.handle_request(request).instrument(span)).await;
+                        Box::pin(worker.handle_request(request)).instrument(span).await;
```
deuszx, line 293: Small change but are you sure this is what we want? @Twey has more experience with these spans.

linera-core/src/chain_worker/state.rs
```diff
@@ -103,7 +106,7 @@ where
     }
 
     /// Handles a request and applies it to the chain state.
-    #[instrument(skip_all)]
+    #[instrument(target = "telemetry_only", skip_all, fields(chain_id = %self.chain_id()))]
```
deuszx: Same – this won't show up in the normal logs anymore?

linera-core/src/chain_worker/state.rs
```diff
@@ -264,6 +267,10 @@ where
     }
 
     /// Returns the requested blob, if it belongs to the current locking block or pending proposal.
+    #[instrument(target = "telemetry_only", skip_all, fields(
+        chain_id = %self.chain_id(),
+        blob_id = %blob_id
```
deuszx, line 275: `blob_id` is IMHO unnecessary.

linera-core/src/chain_worker/actor.rs
ndr-ds: Will remove

linera-core/src/chain_worker/actor.rs
ndr-ds: If we instrument inside the pin, we reach the compilation recursion limit

ndr-ds:
Ah, you're right! I'll do a pass, and if it was already `INFO` level before, the span should be sent already even without the `telemetry_only` thing. So I'll make sure that behavior is maintained for those before I merge

Issue #4535, opened 2025-09-10
Implement a `SyncValidator` on multiple chains at once.


MathieuDutSik:
The function `fn sync_validator` accesses a single chain in a validator. However, if we look at the `SyncValidator` we can accesses the validator on several chains.

However, when we access several chains in a validator, we can diminish the number of operations in a radical way. The problem is that this forces the introduction of several functions:
* First, having a `.handle_chain_info_queries` for multiple chains.
* Accessing `missing_certificate_hashes` over multiple `chain_state_view`.
* The reading of certificates is already parallelized, so no work is needed there.
* The `handle_confirmed_certificate` is currently a very serial algorithm. It is unclear whether it can be parallelized in the `sync_validator` of a single chain. It could be parallelized over the chains.

Before getting into that major work, it would be good to know whether this is important and which part of the processing takes the most time.

PR #4558, opened 2025-09-12
Implement the syncing with all validators functionality.

### LINKED ISSUE #4488: A command to update _all_ validators about chains.
When adding a new validator, we should send at least the full admin chain to that validator, and maybe also the faucet chain.
With the current code, the client will only wait for a quorum and then exit; if the chains are long, this won't wait for the new validator to fully catch up.

We should add a command that does the same as `sync-validator`, but for _all_ validators.
And maybe the client should do that by default whenever it does something on the _admin_ chain.

MathieuDutSik:
## Motivation

In operations, it is useful to be able to sync with all validators in one single operation.

Fixes #4488 

## Proposal

The implementation is straightforward.

## Test Plan

The CI.
I did test with README.md kind of files and added a `linera sync-all-validators` function. And it does access to the validators without error.
I did not test that the syncing does happen, though.

I created issue #4535 that shows some idea for improving the syncing with validators.

## Release Plan

Fully compatible with existing DevNet / TestNet.

## Links

None.

### DIFF
--- a/linera-service/src/cli/main.rs
+++ b/linera-service/src/cli/main.rs
@@ -586,9 +586,7 @@ impl Runnable for Job {
                 let committee = context.wallet().genesis_config().committee.clone();
 
                 for (_validator_name, network_address) in committee.validator_addresses() {
-                    let validator = context
-                        .make_node_provider()
-                        .make_node(network_address)?;
+                    let validator = context.make_node_provider().make_node(network_address)?;
 
                     for chain_id in &chains {
                         let chain = context.make_chain_client(*chain_id);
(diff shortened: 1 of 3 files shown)


linera-service/src/cli/main.rs
```diff
@@ -562,6 +563,42 @@ impl Runnable for Job {
...
+
+                for (_validator_name, network_address) in committee.validator_addresses() {
+                    let validator = context.make_node_provider().make_node(network_address)?;
+
+                    for chain_id in &chains {
+                        let chain = context.make_chain_client(*chain_id);
+
+                        Box::pin(chain.sync_validator(validator.clone())).await?;
+                    }
+                }
```
deuszx: Why not in parallel? We could parallelize across validators or even `validator x chain`.

MathieuDutSik: Ok, done.

PR #4568, opened 2025-09-15
support wallet home directory

ma2bd:
## Motivation

Allow using `--with-wallet NAME` without setting 3 environment variables first!

## Proposal

* Introduce a notion of home directory (e.g. set to a temporary directory for testing)
* Maintain backward compatibility 

## Test Plan

CI

## Release Plan

This first PR is meant to be backward compatibly so that it can be backported on testnet_conway


Twey:
Instead of a home directory could we have `--with-wallet` just take a directory path?

Otherwise we probably want a full suite of commands to do management — create, list, update, delete available wallets.  But I'd much rather we just use a directory with a normal directory path, and the user can manage wallets with their normal filesystem tools.

PR #4569, opened 2025-09-15
fix default storage namespace

ma2bd:
## Motivation

I found this inconsistency after copying a default wallet to a different directory. Then, I couldn't use the copy.

## Proposal

* Use the same default value for all configuration modes
* Use the value that is less likely to break user wallets

## Test Plan

CI + testnet manually

## Release Plan

TBD -- This may break validator deployments if they used the default namespace value.

### DIFF
--- a/linera-service/src/cli/main.rs
+++ b/linera-service/src/cli/main.rs
@@ -1844,7 +1844,7 @@ impl ClientOptions {
                     path: Self::config_path()?.join("wallet.db"),
                     spawn_mode,
                 };
-                let namespace = "default".to_string();
+                let namespace = linera_storage::DEFAULT_NAMESPACE.to_string();
                 Ok(StorageConfig {
                     inner_storage_config,
                     namespace,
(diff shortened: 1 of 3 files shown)


deuszx (approved):
Modulo CI fix.

ma2bd:
@ndr-ds How do we update the image used to run kubernetes tests?

ndr-ds:
It should rebuild every time 🤔 Let me take a look

ndr-ds:
I believe since `default` is a CQL reserved keyword https://docs.datastax.com/en/cql/hcd/reference/cql-syntax.html?#keywords, if you wanna use it you need to surround the identifier with double quotes, which we currently don't do in the ScyllaDB storage code 😅
So either we go and and add double quotes everywhere that the namespace/table name is referred, or we just use a different name

ma2bd:
> I believe since `default` is a CQL reserved keyword https://docs.datastax.com/en/cql/hcd/reference/cql-syntax.html?#keywords, if you wanna use it you need to surround the identifier with double quotes, which we currently don't do in the ScyllaDB storage code 😅 So either we go and and add double quotes everywhere that the namespace/table name is referred, or we just use a different name

Good catch! We should double-quote!

ndr-ds:
This should do it: https://github.com/linera-io/linera-protocol/pull/4745
If you rebase this on top of #4745 I think it should work

PR #4570, opened 2025-09-15
Add some useful client commands

bart-linera:
## Motivation

It is hard to find out the ID of the admin chain, or to check the contents of blocks using the CLI client.

## Proposal

This PR adds some useful commands to the client: `show-network-description` (which contains the admin chain ID), `chain show-chain-description` and `chain show-block`.

## Test Plan

Tested manually.

## Release Plan

TBD (could be backported to the testnet and devnet branches, but it's not strictly necessary).

## Links

### DIFF
--- a/linera-service/src/cli/command.rs
+++ b/linera-service/src/cli/command.rs
@@ -6,7 +6,7 @@ use std::{borrow::Cow, num::NonZeroU16, path::PathBuf};
 use chrono::{DateTime, Utc};
 use linera_base::{
     crypto::{AccountPublicKey, CryptoHash, ValidatorPublicKey},
-    data_types::{Amount, Epoch},
+    data_types::{Amount, BlockHeight, Epoch},
     identifiers::{Account, AccountOwner, ApplicationId, ChainId, ModuleId, StreamId},
     time::Duration,
     vm::VmRuntime,
@@ -269,6 +269,9 @@ pub enum ClientCommand {
         /// Chain ID (must be one of our chains)
         chain_id: ChainId,
     },
+
+    /// Print out the network description.
+    ShowNetworkDescription,
 
     /// Read the current native-token balance of the given account directly from the local
     /// state.
@@ -917,6 +920,10 @@ pub enum ClientCommand {
     #[command(subcommand)]
     Wallet(WalletCommand),
 
+    /// Show the contents of the wallet.
+    #[command(subcommand)]
+    Chain(ChainCommand),
+
     /// Manage Linera projects.
     #[command(subcommand)]
     Project(ProjectCommand),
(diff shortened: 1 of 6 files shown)


linera-service/src/cli/main.rs
```diff
@@ -1654,6 +1659,33 @@ impl Runnable for Job {
...
+            Chain(ChainCommand::ShowChainDescription { chain_id }) => {
+                let context = ClientContext::new(
+                    storage,
+                    options.context_options.clone(),
+                    wallet,
+                    signer.into_value(),
+                );
+                let chain_id = chain_id.unwrap_or_else(|| context.default_chain());
+                let chain_client = context.make_chain_client(chain_id);
+                let description = chain_client.get_chain_description().await?;
```
deuszx: I'd just make sure that in the case of missing chain description (and the block above) we print some readable message and not - possibly obscure - output of the error variant.

linera-service/src/cli/command.rs
```diff
@@ -1206,6 +1215,23 @@ pub enum WalletCommand {
     ForgetChain { chain_id: ChainId },
 }
 
+#[derive(Clone, clap::Subcommand)]
+pub enum ChainCommand {
+    /// Show the contents of a block.
+    ShowBlock {
+        /// The chain to show the block.
+        chain_id: Option<ChainId>,
```
MathieuDutSik: Specify here and below that if `None` the default chain is shown.

MathieuDutSik (approved):
I would add the use of the commands in one (or more) of the README.md in the `examples` directory.

linera-service/src/cli/main.rs
bart-linera: See 42546ef, I'm not sure if that's what you had in mind 😅

linera-service/src/cli/main.rs
```diff
@@ -1654,6 +1662,51 @@ impl Runnable for Job {
...
+                    .context("Failed to load chain")?;
+                let block_hash = chain_state_view
+                    .block_hashes(height..=height)
+                    .await
+                    .context("Failed to find a block hash for the given height")?[0];
+                let block = context
+                    .storage()
+                    .read_confirmed_block(block_hash)
+                    .await
+                    .context("Failed to find the given block in storage")?;
```
deuszx, line 1686: Personally I'd replace with something simpler and definitive:

"block <hash> not found in storage"

"chain description for <chain_id> not found in storage"

etc. _failed_ sounds a bit like an intermediate failure.

bart-linera: I was mostly following the pattern of the other messages in other subcommands, but I can change these if you feel strongly about this :sweat_smile:

deuszx: I just think "Failed to do ..." may also mean this is an internal problem during the process.

"X is not found in storage" is non-ambiguous.

bart-linera: I think that would actually be a drawback here - because `read_confirmed_block` can fail in more ways than just not finding the block in storage. For example, finding it and failing to deserialize it. This only adds context to the potential error, not replaces it (and printing a separate message for each error would require a huge `match`, which would also duplicate a lot of `Display` implementations for the error variants, I think).

PR #4576, opened 2025-09-15
Implement the caching of the `find_{keys,key_value}_by_prefix` operations.

MathieuDutSik:
## Motivation

The `find_keys_by_prefix` has emerged as a choke point for some benchmarks.
Considering that it is the only part that is not cached, it is interesting to consider this.

## Proposal

The implementation is done in the following way:
* The `find_{keys,key_values}_by_prefix` are cached.
* The `delete_prefix` leads to a `find_key_values_by_prefix` entry in the cache.
* All the information about the operation is used to avoid making some calls to the storage.

## Test Plan

The CI.
Some unit tests have been added to the `lru_caching.rs` file.

## Release Plan

The input file would change because some additional entries have been added.
However, this can still be ported to existing DevNet / TestNet.

## Links

None.

### DIFF
--- a/linera-views/src/lru_prefix_cache.rs
+++ b/linera-views/src/lru_prefix_cache.rs
@@ -299,7 +299,7 @@ impl LruPrefixCache {
         assert!(self.queue.insert(cache_key, size).is_none());
     }
 
-    /// If the FindKeys map contain a prefix that is a prefix of key in argument,
+    /// If the FindKeys map contains a prefix that is a prefix of key in argument,
     /// then returns it and the corresponding FindKeys. Otherwise `None`.
     ///
     /// The algorithm of this function is tested in `test_lower_bounds`.
@@ -307,34 +307,37 @@ impl LruPrefixCache {
     /// What makes this functionality work is that the set of keys of the `find_keys_map`
     /// is prefix free.
     ///
-    /// For one of the map in question, let us define `set` to be the `BTreeSet` of the keys
-    /// of the map.
+    /// Claim: `self.get_existing_find_keys_entry(key)` finds a prefix of key in `self.find_keys_map.keys()` iff such prefix exists.
+    ///
+    /// Note that when it exists, such a prefix is unique because `self.find_keys_map.keys()` is prefix-free.
+    ///
+    /// Proof: For the map `find_keys_map` in question, let us define `set` to be the `BTreeSet` of
+    /// the keys of the map.
     /// Then define `S = { s in set | s <= key }` for the lexicographic ordering.
     /// First of all the expression `self.find_keys_map.range(..=key.to_vec())` corresponds
     /// to S and `self.find_keys_map.range(..=key.to_vec()).next_back()` is
     /// * None if S is empty.
     /// * Some(M) with M the maximum of S if S is non-empty.
     ///
-    /// If there is a prefix `p` of `key` in `S` then we have `p <= key` in the lexicographic
-    /// ordering. So, S is non-empty. That prefix `p` is the highest element of `S`. If that
-    /// were not the case, we would have for another `q in S` the inequality `p < q <= key`.
-    /// This forces `q` to be a prefix of `key` and `p` to be a prefix of `q`. This breaks
-    /// the prefix-free property.
+    /// First, if `self.get_existing_find_keys_entry(key) == Some(stored_key)` then given the code,
+    /// clearly `stored_key` is a prefix of key.
     ///
-    /// If there is no prefix `p` of `key` in `S` then if `S` is empty the code is correct.
-    /// If it is not empty, then it has a maximum element M. That highest element cannot
-    /// be a prefix of `key` and that is checked in the code (named `key_store` there).
-    /// If there were an element `h` of `S` that would be a prefix, then the inequality
-    /// `h <= M <= key` would force `M` to also be a prefix of `key`.
+    /// Conversely, let us assume that `self.find_keys_map.keys()` contains a certain prefix `p` of key.
+    /// Because in particular `p <= key`, we have `self.find_keys_map.range(..=key.to_vec()).next_back() == Some((stored_key, _))`
+    /// for some value `stored_key` such that `p <= stored_key <= key`.
     ///
-    /// Therefore the code does return an `Some(p)` if and only if there is a prefix `p`
-    /// in the map of find-keys prefixes.
+    /// Next, let us prove that `p` is a prefix of `stored_key`. (This will entail `stored_key == p` due to
+    /// the prefix-free property of `self.find_keys_map`, therefore the algorithm's answer is correct.)
+    ///
+    /// By contradiction. Let `w` be the longest common prefix between `p` and `stored_key`. If `p = w p2` and
+    /// `stored_key = w s2` with `p2` non-empty, then `p <= stored_key` implies `s2 > p2`. But `p` is also a
+    /// prefix of `key` therefore `stored_key > key`, contradiction.
     fn get_existing_find_keys_entry(&self, key: &[u8]) -> Option<(&Vec<u8>, &FindKeysEntry)> {
         match self.find_keys_map.range(..=key.to_vec()).next_back() {
             None => None,
-            Some((key_store, value)) => {
-                if key.starts_with(key_store) {
-                    Some((key_store, value))
+            Some((stored_key, value)) => {
+                if key.starts_with(stored_key) {
+                    Some((stored_key, value))
                 } else {
                     None
                 }
@@ -349,9 +352,9 @@ impl LruPrefixCache {
     ) -> Option<(&Vec<u8>, &mut FindKeysEntry)> {
         match self.find_keys_map.range_mut(..=key.to_vec()).next_back() {
             None => None,
-            Some((key_store, value)) => {
-                if key.starts_with(key_store) {
-                    Some((key_store, value))
+            Some((stored_key, value)) => {
+                if key.starts_with(stored_key) {
+                    Some((stored_key, value))
                 } else {
                     None
                 }
@@ -367,9 +370,9 @@ impl LruPrefixCache {
     ) -> Option<(&Vec<u8>, &FindKeyValuesEntry)> {
         match self.find_key_values_map.range(..=key.to_vec()).next_back() {
             None => None,
-            Some((key_store, value)) => {
-                if key.starts_with(key_store) {
-                    Some((key_store, value))
+            Some((stored_key, value)) => {
+                if key.starts_with(stored_key) {
+                    Some((stored_key, value))
                 } else {
                     None
                 }
@@ -388,9 +391,9 @@ impl LruPrefixCache {
             .next_back()
         {
             None => None,
-            Some((key_store, value)) => {
-                if key.starts_with(key_store) {
-                    Some((key_store, value))
+            Some((stored_key, value)) => {
+                if key.starts_with(stored_key) {
+                    Some((stored_key, value))
                 } else {
                     None
                 }
@@ -525,16 +528,16 @@ impl LruPrefixCache {
         if self.has_exclusive_access {
             let lower_bound = self.get_existing_keys_entry_mut(key);
             if let Some((lower_bound, cache_entry)) = lower_bound {
-                let key_red = &key[lower_bound.len()..];
-                cache_entry.update_entry(key_red, true);
+                let reduced_key = &key[lower_bound.len()..];
+                cache_entry.update_entry(reduced_key, true);
                 let cache_key = CacheKey::FindKeys(lower_bound.to_vec());
                 let new_size = lower_bound.len() + cache_entry.size();
                 self.update_cache_key_sizes(&cache_key, new_size);
             }
             match self.get_existing_find_key_values_entry_mut(key) {
                 Some((lower_bound, cache_entry)) => {
-                    let key_red = &key[lower_bound.len()..];
-                    cache_entry.update_entry(key_red, Some(value.to_vec()));
+                    let reduced_key = &key[lower_bound.len()..];
+                    cache_entry.update_entry(reduced_key, Some(value.to_vec()));
                     let cache_key = CacheKey::FindKeyValues(lower_bound.to_vec());
                     let new_size = lower_bound.len() + cache_entry.size();
                     self.update_cache_key_sizes(&cache_key, new_size);
@@ -558,8 +561,8 @@ impl LruPrefixCache {
             let lower_bound = self.get_existing_keys_entry_mut(key);
             let mut matching = false; // If matching, no need to insert in the value cache
             if let Some((lower_bound, cache_entry)) = lower_bound {
-                let key_red = &key[lower_bound.len()..];
-                cache_entry.update_entry(key_red, false);
+                let reduced_key = &key[lower_bound.len()..];
+                cache_entry.update_entry(reduced_key, false);
                 let cache_key = CacheKey::FindKeys(lower_bound.to_vec());
                 let new_size = lower_bound.len() + cache_entry.size();
                 self.update_cache_key_sizes(&cache_key, new_size);
@@ -567,8 +570,8 @@ impl LruPrefixCache {
             }
             let lower_bound = self.get_existing_find_key_values_entry_mut(key);
             if let Some((lower_bound, cache_entry)) = lower_bound {
-                let key_red = &key[lower_bound.len()..];
-                cache_entry.update_entry(key_red, None);
+                let reduced_key = &key[lower_bound.len()..];
+                cache_entry.update_entry(reduced_key, None);
                 let cache_key = CacheKey::FindKeyValues(lower_bound.to_vec());
                 let new_size = lower_bound.len() + cache_entry.size();
                 self.update_cache_key_sizes(&cache_key, new_size);
@@ -839,8 +842,8 @@ impl LruPrefixCache {
             // Now trying the FindKeyValues map.
             let lower_bound = self.get_existing_find_key_values_entry(key);
             let (lower_bound, result) = if let Some((lower_bound, find_entry)) = lower_bound {
-                let key_red = &key[lower_bound.len()..];
-                (lower_bound, find_entry.get_read_value(key_red))
+                let reduced_key = &key[lower_bound.len()..];
+                (lower_bound, find_entry.get_read_value(reduced_key))
             } else {
                 return None;
             };
@@ -869,8 +872,8 @@ impl LruPrefixCache {
             // Now trying the FindKeys map.
             let lower_bound = self.get_existing_find_keys_entry(key);
             let result = if let Some((lower_bound, find_entry)) = lower_bound {
-                let key_red = &key[lower_bound.len()..];
-                Some((lower_bound, find_entry.contains_key(key_red)))
+                let reduced_key = &key[lower_bound.len()..];
+                Some((lower_bound, find_entry.contains_key(reduced_key)))
             } else {
                 None
             };
@@ -882,8 +885,8 @@ impl LruPrefixCache {
             // Now trying the FindKeyValues map.
             let lower_bound = self.get_existing_find_key_values_entry(key);
             let (lower_bound, result) = if let Some((lower_bound, find_entry)) = lower_bound {
-                let key_red = &key[lower_bound.len()..];
-                (lower_bound, find_entry.contains_key(key_red))
+                let reduced_key = &key[lower_bound.len()..];
+                (lower_bound, find_entry.contains_key(reduced_key))
             } else {
                 return None;
             };


deuszx:
The **Proposal** section should explain the high-level explanation what is being done. Here, it basically repeats the title of the PR: "cache find_keys_by_prefix; don't make unnecessary calls to the storage".

linera-indexer/lib/src/rocks_db.rs
```diff
@@ -47,6 +47,18 @@ pub struct RocksDbConfig {
     /// The maximal number of entries in the storage cache.
     #[arg(long, default_value = "1000")]
     pub max_cache_entries: usize,
+
+    /// The maximal memory used in the value cache.
```
deuszx: In megabytes? kilobytes?

linera-indexer/lib/src/rocks_db.rs
```diff
@@ -47,6 +47,18 @@ pub struct RocksDbConfig {
...
+    #[arg(long, default_value = "10000000")]
+    pub max_cache_value_size: usize,
+
+    /// The maximal memory used in the find_keys_by_prefix cache.
```
deuszx: Do we need to configure each cache separately?

linera-views/src/backends/lru_caching.rs
```diff
@@ -74,6 +113,12 @@ pub struct StorageCacheConfig {
     pub max_entry_size: usize,
     /// The maximum number of entries in the cache.
     pub max_cache_entries: usize,
+    /// The maximum size of cached values
```
deuszx: In config you called it _memory_ but here just _size_.

linera-views/src/backends/lru_caching.rs
```diff
@@ -83,32 +128,149 @@ pub const DEFAULT_STORAGE_CACHE_CONFIG: StorageCacheConfig = StorageCacheConfig
...
     fn size(&self) -> usize {
         match self {
-            CacheEntry::Value(vec) => vec.len(),
+            ValueCacheEntry::Value(vec) => vec.len(),
             _ => 0,
         }
     }
 }
 
+struct FindKeysEntry(BTreeSet<Vec<u8>>);
```
deuszx: Maybe Merkle Patricia Trie could be better-suited for this since it's designed to store data that share a prefix.

linera-views/src/backends/lru_caching.rs
```diff
@@ -83,32 +128,149 @@ pub const DEFAULT_STORAGE_CACHE_CONFIG: StorageCacheConfig = StorageCacheConfig
...
+    fn get_find_keys(&self, key_prefix: &[u8]) -> Vec<Vec<u8>> {
+        let key_prefix = key_prefix.to_vec();
+        let delta = key_prefix.len();
+        self.0
+            .range(get_interval(key_prefix))
+            .map(|key| key[delta..].to_vec())
+            .collect()
+    }
+
+    fn get_contains_key(&self, key: &[u8]) -> bool {
```
deuszx: What do you need the `get_` prefix for?

linera-views/src/backends/lru_caching.rs
```diff
@@ -83,32 +128,149 @@ pub const DEFAULT_STORAGE_CACHE_CONFIG: StorageCacheConfig = StorageCacheConfig
...
 }
 
+struct FindKeysEntry(BTreeSet<Vec<u8>>);
+
+impl FindKeysEntry {
+    fn size(&self) -> usize {
+        self.0.iter().map(|key| key.len()).sum()
+    }
+
+    fn get_find_keys(&self, key_prefix: &[u8]) -> Vec<Vec<u8>> {
```
deuszx: `get_keys_for_prefix`?

linera-views/src/backends/lru_caching.rs
```diff
@@ -83,32 +128,149 @@ pub const DEFAULT_STORAGE_CACHE_CONFIG: StorageCacheConfig = StorageCacheConfig
...
+            .range(get_interval(key_prefix))
+            .map(|key| key[delta..].to_vec())
+            .collect()
+    }
+
+    fn get_contains_key(&self, key: &[u8]) -> bool {
+        self.0.contains(key)
+    }
+
+    fn update_cache_entry(&mut self, key: &[u8], new_value: Option<Vec<u8>>) {
```
deuszx: I don't think you need the `_cache_` part – it's implicit in this context.

linera-views/src/backends/lru_caching.rs
```diff
@@ -83,32 +128,149 @@ pub const DEFAULT_STORAGE_CACHE_CONFIG: StorageCacheConfig = StorageCacheConfig
...
+    }
+
+    fn get_find_key_values(&self, key_prefix: &[u8]) -> Vec<(Vec<u8>, Vec<u8>)> {
+        let key_prefix = key_prefix.to_vec();
+        let delta = key_prefix.len();
+        self.0
+            .range(get_interval(key_prefix))
+            .map(|(key, value)| (key[delta..].to_vec(), value.to_vec()))
+            .collect()
+    }
```
deuszx, line 227: If you returned `impl Iterator<Item = (Vec<u8>, Vec<u8>)>` then the caller could decide whether they want just the key, just value, or both. Both methods with one function.

Alternatively, it could be used internally – by these two – and expose the final API.

deuszx: If you returned `impl Iterator<Item = (Vec<u8>, Vec<u8>)>` then the caller could decide whether they want just the key, just value, or both. Both methods with one function.

Alternatively, it could be used internally – by these two – and expose the final API.

linera-views/src/backends/lru_caching.rs
```diff
@@ -83,32 +128,149 @@ pub const DEFAULT_STORAGE_CACHE_CONFIG: StorageCacheConfig = StorageCacheConfig
...
+
+    fn get_contains_key(&self, key: &[u8]) -> bool {
+        self.0.contains_key(key)
+    }
+
+    fn get_read_value(&self, key: &[u8]) -> Option<Vec<u8>> {
```
deuszx: What's the deal with `_read_`? What does it refer to? Get and Read are interchangeable.

linera-views/src/backends/lru_caching.rs
```diff
@@ -83,32 +128,149 @@ pub const DEFAULT_STORAGE_CACHE_CONFIG: StorageCacheConfig = StorageCacheConfig
...
+        let keys = self
+            .0
+            .range(get_interval(key_prefix.to_vec()))
+            .map(|(key, _)| key.clone())
+            .collect::<Vec<_>>();
+        for key in keys {
+            self.0.remove(&key);
+        }
+    }
+}
```
deuszx: Many methods between `FindKeyValuesEntry` and `FindKeysEntry` could be extracted to a common trait with default implementations.

linera-views/src/backends/lru_caching.rs
```diff
@@ -117,135 +279,619 @@ impl LruPrefixCache {
...
 
+    /// A used key needs to be put on top
+    fn put_cache_key_on_top(&mut self, cache_key: CacheKey) {
+        let size = self.queue.remove(&cache_key).unwrap();
+        self.queue.insert(cache_key, size);
+    }
+
+    /// Remove an entry from the queue
+    fn remove_cache_key(&mut self, cache_key: &CacheKey) {
+        let size = self.queue.remove(cache_key).unwrap();
```
deuszx: Why are those `unwrap()`s safe? Replace with `expect(<msg>)` where message explain why it was never supposed to happen (i.e. the panic)

linera-views/src/backends/lru_caching.rs
```diff
@@ -117,135 +279,619 @@ impl LruPrefixCache {
...
+                self.total_find_keys_size += new_size;
+            }
+            CacheKey::FindKeyValues(_) => {
+                self.total_find_key_values_size -= old_size;
+                self.total_find_key_values_size += new_size;
+            }
+        }
+    }
+
+    /// Insert a cache entry into the key
```
deuszx: Is this comment correct? It doesn't explain what's happening in `insert_cache_key`.

linera-views/src/backends/lru_caching.rs
```diff
@@ -117,135 +279,619 @@ impl LruPrefixCache {
...
+                if key.starts_with(key_store) {
+                    Some((key_store, value))
+                } else {
+                    None
+                }
+            }
+        }
+    }
+
+    fn get_find_keys_lower_bound_update(
```
deuszx: I think that a "Rust way" is to append a `_mut` to the function name when you intend to return a mutable pointer. So `get_find_keys_lower_bound_mut`.

deuszx: Same for `get_find_key_values_lower_bound_update`.

linera-views/src/backends/lru_caching.rs
```diff
@@ -117,135 +279,619 @@ impl LruPrefixCache {
...
+            }
+        }
+    }
+
+    /// Trim value cache so that it fits within bounds
+    fn trim_value_cache(&mut self) {
+        let mut keys = Vec::new();
+        let mut total_size = self.total_value_size;
+        let mut iter = self.queue.iter();
+        loop {
```
deuszx: a `while total_size > self.config.max_cache_value_size` loop will be more readable.

linera-views/src/backends/lru_caching.rs
MathieuDutSik: We had that before for the `KeyValueStore` with the `type Keys: KeyITerable` / `type KeyValues: KeyValueIterable`. 
But we removed that.

linera-views/src/backends/lru_caching.rs
```diff
@@ -117,135 +279,619 @@ impl LruPrefixCache {
...
+            .range_mut(..=key.to_vec())
+            .next_back()
+        {
+            None => None,
+            Some((key_store, value)) => {
+                if key.starts_with(key_store) {
+                    Some((key_store, value))
+                } else {
+                    None
+                }
```
deuszx: If `range_mut(_=key)` returns `Some(_)`, when does it happen that `key` would **not** start with the `key_store` prefix? I thought it's the whole point of the range queries.

linera-indexer/lib/src/rocks_db.rs
MathieuDutSik: in bytes. Added.

linera-indexer/lib/src/rocks_db.rs
MathieuDutSik: Right now, yes it is done in that way.

linera-views/src/backends/lru_caching.rs
MathieuDutSik: Yes, because the cache is in memory.

linera-views/src/backends/lru_caching.rs
MathieuDutSik: It is a general idea. That could be applied to many other parts of the code.
But yes, good idea.
I created an issue.

linera-views/src/backends/lru_caching.rs
MathieuDutSik: Ok, can eliminate.

linera-views/src/backends/lru_caching.rs
MathieuDutSik: I choose `get_keys_by_prefix`.

linera-views/src/backends/lru_caching.rs
MathieuDutSik: Correct.

linera-views/src/backends/lru_caching.rs
MathieuDutSik: It is overkill, I think.

linera-views/src/backends/lru_caching.rs
MathieuDutSik: Well, because the key was present. We can of course add the expect, though the benefit is limited.

linera-views/src/backends/lru_caching.rs
MathieuDutSik: Not correct indeed. Corrected.

linera-views/src/backends/lru_caching.rs
MathieuDutSik: Yes, good point.

linera-views/src/backends/lru_caching.rs
MathieuDutSik: If there are more than 1 way to exit the loop, then I find that better.

linera-views/src/backends/lru_caching.rs
MathieuDutSik: It is about the ordering of the keys.

linera-views/src/backends/lru_caching.rs
```diff
@@ -55,60 +55,222 @@ mod metrics {
...
+    /// The maximum size of an entry size, in bytes.
     pub max_entry_size: usize,
     /// The maximum number of entries in the cache.
     pub max_cache_entries: usize,
+    /// The maximum size of cached values.
+    pub max_cache_value_size: usize,
+    /// The maximum size of cached find_keys_by_prefix.
+    pub max_cache_find_keys_size: usize,
+    /// The maximum size of cached find_key_values_by_prefix.
+    pub max_cache_find_key_values_size: usize,
```
ma2bd: You create the terms "value cache", "find-keys cache", "find-key-values cache" in other places. Therefore, those names should be
`max_value_cache_size`
`max_find_keys_cache_size`
`max_find_key_values_cache_size`

linera-views/src/backends/lru_caching.rs
```diff
@@ -117,141 +279,634 @@ impl LruPrefixCache {
...
+                let cache_key = CacheKey::Value(key);
+                self.insert_cache_key(cache_key, size);
             }
         }
+        self.trim_value_cache();
         self.trim_cache();
     }
 
+    /// Invalidates corresponding find entries.
+    pub fn correct_find_entry(&mut self, key: &[u8], new_value: Option<Vec<u8>>) {
```
ma2bd:
```
// Updates the find-keys (and find-key-values) cache entries
pub fn update_find_key_entries(...) { ... }
``

File: linera-views/src/backends/lru_caching.rs [hunk 18]
```diff
@@ -117,141 +279,634 @@ impl LruPrefixCache {
...
+            // Now trying the find maps.
+            let lower_bound = self.get_find_key_values_lower_bound(key);
+            let (lower_bound, result) = if let Some((lower_bound, find_entry)) = lower_bound {
+                let key_red = &key[lower_bound.len()..];
+                (lower_bound, find_entry.get_read_value(key_red))
+            } else {
+                return None;
+            };
+            let cache_key = CacheKey::FindKeyValues(lower_bound.clone());
+            self.put_cache_key_on_top(cache_key);
```
ma2bd: Promoting the entire find query because we hit a value inside is questionable.

linera-views/src/backends/lru_caching.rs
```diff
@@ -117,141 +279,634 @@ impl LruPrefixCache {
...
-            self.queue.insert(key.to_vec(), key_value_size);
+            let cache_key = CacheKey::Value(key.to_vec());
+            self.put_cache_key_on_top(cache_key);
+            return result;
+        }
+        if self.has_exclusive_access {
+            // Now trying the find maps.
+            let lower_bound = self.get_find_key_values_lower_bound(key);
+            let (lower_bound, result) = if let Some((lower_bound, find_entry)) = lower_bound {
+                let key_red = &key[lower_bound.len()..];
```
ma2bd: Here, IIUC it looks like we're trying our luck in previously cached `find-key-values` entries to see if one of them would contain the key we're looking for.

This approach seems to have a few downsides:
* This slows down the most important API `query_read_value`
* We may store information on some values twice in memory.
* We cache find-keys queries and find-key-values queries separately, creating code duplication (and missed hits in rare cases?)

Instead, would it possible to remove the notion of cached find-key-values entirely?
In order to cache find-key-values, we would see them as a combination of a cached find-keys query and many read-value queries.

This seems to solve the problems above:
* `query_read_value` can ignore `find-keys` entries
* Values are stored in one place
* `FindKeyValuesEntry` and all the corresponding code is no longer needed.

I suppose one drawback is that this introduces the possibility of partial cache hits for a `find-key-values`, meaning that we could hit a similar query `find-keys` but having forgotten some of the values. I believe that's fine: the missing values can be retrieve from storage in one multi-read. (If needed, we could also make this case impossible by evicting the find-keys entry itself if any of the values is evicted, but I don't see why.)

MathieuDutSik: Here is my answer to the raised issues:
* The idea of the cache is that access to the storage is so very expensive that caching it is a good idea.
* We have to look at what time scale we are considering here. The timescale of accessing a `BTreeMap` is about a dozen nanoseconds. So, checking if the value is in the `find-key-values-by-queries` cache is quite sensible. The access to storage is going to be in milliseconds in the best scenario.
* I do not think that information is stored twice in memory. The `find_key_values_map` is prefix-free. When an entry is inserted into the `find_key_values_map`, then the corresponding value caches are cleared.
* The separation of the code between `find-keys-by-prefix` and `find-key-values-by-prefix` is, I think, a welcome change in this PR. Before I tried to have it all in, and this created some problems.
* If we want to disable the caching of the `find-key-values-by-prefix`, then we can set the maximal size to `0`. But maybe the better proposition is to introduce a `max_cache_size` for the `value`, `find-keys`, `find-key-values`.

linera-views/src/backends/lru_caching.rs
```diff
@@ -55,60 +55,222 @@ mod metrics {
...
-    /// The maximum size of an entry size, in bytes
+    /// The maximum size of an entry size, in bytes.
     pub max_entry_size: usize,
     /// The maximum number of entries in the cache.
     pub max_cache_entries: usize,
+    /// The maximum size of cached values.
+    pub max_cache_value_size: usize,
+    /// The maximum size of cached find_keys_by_prefix.
+    pub max_cache_find_keys_size: usize,
+    /// The maximum size of cached find_key_values_by_prefix.
```
afck:
```suggestion
    /// The maximum size of cached `find_keys_by_prefix` results.
    pub max_cache_find_keys_size: usize,
    /// The maximum size of cached `find_key_values_by_prefix` results.
```

linera-views/src/backends/lru_caching.rs
```diff
@@ -55,60 +55,222 @@ mod metrics {
...
+
+    fn delete_prefix(&mut self, key_prefix: &[u8]) {
+        let keys = self
+            .0
+            .range(get_interval(key_prefix.to_vec()))
+            .map(|(key, _)| key.clone())
+            .collect::<Vec<_>>();
+        for key in keys {
+            self.0.remove(&key);
+        }
```
deuszx: if you iterate over the result of the `self.0.range.map` you won't have to allocate the vector

MathieuDutSik: How could that work?
If you want to remove entries from a map, you need `&mut self`access. But precisely, the `.range` previously has `&self` access.

deuszx: Ah, this is removal. Sorry, I missed that you need exclusive access.

linera-views/src/backends/lru_caching.rs
ma2bd: Even if values are not duplicated, the set of values is still stored in two places, and the absence of duplication would be a nontrivial invariant.

linera-views/src/backends/lru_caching.rs
MathieuDutSik: Thank you.

linera-views/src/backends/lru_caching.rs
MathieuDutSik: The "nontrivial invariant" in question is an issue that can be addressed in a number of ways.

One is to use formal verification in Rust. The formalisms being used for Rust are Kani and VeriFast. It is a little bit unclear how much we can verify for `BTreeMap`. But that track is realistic even though success is not guaranteed.

Looking at this by hand, we find that the code whose coherence is in question is `LruPrefixCache`. It is a single type. So, in order to better understand the "nontrivial invariant" in question, it is better to put it into a separate module. This way, we can identify the functions that are private from the ones that are not. So, a `lru_caching_cache.rs` is created.

The code of the function is mostly some relatively simple code. But there are some parts that are more challenging. I am thinking of the "*_lower_bound" functions authored by @afck. They provide some ways to improve the `DeletionSet` and `SuffixClosedSetIterator`. So, maybe this should be done in a separate PR.

The fact that the `FindKeyValues` entries contain information that are possibly duplicated from the value cache is a fact. However, the `FindKeys` entries also contain some duplicate information, since they can resolve the `contains_key` queries. Therefore, that "nontrivial invariant" argument applies just as well here.

linera-views/src/lru_prefix_cache.rs
```diff
@@ -0,0 +1,1289 @@
+// Copyright (c) Zefchain Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! A prefix based cache for the key/value store.
```
ma2bd: nit: AFAIK "prefix cache" or "prefix-based cache" is not really a thing.
I think you just mean "An LRU cache that supports prefix-search APIs"

linera-views/src/lru_prefix_cache.rs
```diff
@@ -0,0 +1,1289 @@
...
+            self.0.remove(&key);
+        }
+    }
+}
+
+/// Stores the data for simple `read_values` queries.
+///
+/// This data structure is inspired by the crate `lru-cache` but was modified to support
+/// range deletions.
+pub(crate) struct LruPrefixCache {
```
ma2bd, line 157: nit: Can we move this definition near the top of the file?

linera-views/src/lru_prefix_cache.rs
```diff
@@ -0,0 +1,1289 @@
...
+    find_keys_map: BTreeMap<Vec<u8>, FindKeysEntry>,
+    find_key_values_map: BTreeMap<Vec<u8>, FindKeyValuesEntry>,
+    queue: LinkedHashMap<CacheKey, usize, RandomState>,
+    config: StorageCacheConfig,
+    total_size: usize,
+    total_value_size: usize,
+    total_find_keys_size: usize,
+    total_find_key_values_size: usize,
+    /// Whether we have exclusive R/W access to the keys under the root key of the store.
+    pub has_exclusive_access: bool,
```
ma2bd: This could be private if we add `fn has_exclusive_access(&self) -> bool { ... }`

linera-views/src/lru_prefix_cache.rs
```diff
@@ -0,0 +1,1289 @@
...
+}
+
+impl LruPrefixCache {
+    /// Creates an `LruPrefixCache`.
+    pub fn new(config: StorageCacheConfig, has_exclusive_access: bool) -> Self {
```
ma2bd: Should all the `pub` methods below be `pub(crate)` logically speaking?

linera-views/src/lru_prefix_cache.rs
```diff
@@ -0,0 +1,1289 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+//! A prefix based cache for the key/value store.
+
+use std::collections::{btree_map::Entry, hash_map::RandomState, BTreeMap, BTreeSet};
+
+use linked_hash_map::LinkedHashMap;
+use serde::{Deserialize, Serialize};
+
+use crate::common::get_interval;
```
ma2bd: The name of this function doesn't say much. It should be called `get_key_range_for_prefix` or something like that.

linera-views/src/lru_prefix_cache.rs
```diff
@@ -0,0 +1,1289 @@
...
+}
+
+struct FindKeysEntry(BTreeSet<Vec<u8>>);
+
+impl FindKeysEntry {
+    fn size(&self) -> usize {
+        self.0.iter().map(|key| key.len()).sum()
+    }
+
+    fn get_keys_by_prefix(&self, key_prefix: &[u8]) -> Vec<Vec<u8>> {
```
ma2bd: This function should take a `Vec<u8>`.

linera-views/src/lru_prefix_cache.rs
```diff
@@ -0,0 +1,1289 @@
...
+            _ => 0,
+        }
+    }
+}
+
+struct FindKeysEntry(BTreeSet<Vec<u8>>);
+
+impl FindKeysEntry {
+    fn size(&self) -> usize {
+        self.0.iter().map(|key| key.len()).sum()
```
ma2bd: `.map(Vec::len).sum()` ?

linera-views/src/lru_prefix_cache.rs
```diff
@@ -0,0 +1,1289 @@
...
+    }
+
+    fn get_keys_by_prefix(&self, key_prefix: &[u8]) -> Vec<Vec<u8>> {
+        let key_prefix = key_prefix.to_vec();
+        let delta = key_prefix.len();
```
ma2bd: "delta" is weird. In doubt, just go for the least surprising: `key_prefix_len` or `length`.

linera-views/src/lru_prefix_cache.rs
```diff
@@ -0,0 +1,1289 @@
...
+}
+
+#[derive(Eq, Hash, PartialEq, Debug)]
+enum CacheKey {
+    Value(Vec<u8>),
+    FindKeys(Vec<u8>),
+    FindKeyValues(Vec<u8>),
+}
+
+pub(crate) enum ValueCacheEntry {
```
ma2bd: `ValueEntry` would be move consistent with the other entries -- unless you rename them (`FindKeysCacheEntry` etc).

linera-views/src/lru_prefix_cache.rs
```diff
@@ -0,0 +1,1289 @@
...
+            total_size: 0,
+            total_value_size: 0,
+            total_find_keys_size: 0,
+            total_find_key_values_size: 0,
+            has_exclusive_access,
+        }
+    }
+
+    /// A used key needs to be put on top.
+    fn put_cache_key_on_top(&mut self, cache_key: CacheKey) {
```
ma2bd: nit: `move_cache_key_on_top` ?

linera-views/src/lru_prefix_cache.rs
```diff
@@ -0,0 +1,1289 @@
...
+        let old_size = *size;
+        *size = new_size;
+        self.decrease_sizes(cache_key, old_size);
+        self.increase_sizes(cache_key, new_size);
+    }
+
+    /// Insert a cache_key into the queue and update sizes.
+    fn insert_cache_key(&mut self, cache_key: CacheKey, size: usize) {
+        self.increase_sizes(&cache_key, size);
+        self.queue.insert(cache_key, size);
```
ma2bd: Could we assert that the `cache_key` was absent? (otherwise sizes may be silently miscalculated)

linera-views/src/backends/lru_caching.rs
```diff
@@ -568,6 +472,22 @@ impl<S> LruCachingStore<S> {
...
+        };
+        if has_exclusive_access {
+            Some(cache)
+        } else {
+            None
+        }
+    }
 }
 
 /// A memory darabase with caching.
```
ma2bd: typo: darabase

linera-views/src/lru_prefix_cache.rs
```diff
@@ -0,0 +1,1289 @@
...
+            None => None,
+            Some((key_store, value)) => {
+                if key.starts_with(key_store) {
+                    Some((key_store, value))
+                } else {
+                    None
+                }
+            }
+        }
+    }
```
ma2bd: Is this correct? If `self.find_keys_map` is `{ "a" -> E1, "ab" -> E2}` and `key` is "ac", then `self.find_keys_map.range(..=key.to_vec()).next_back()` returns `("ab", E2)`, which is not useful, but `("a", E1)` was.

ma2bd: Sadly, you may have to try `key`, then `key[0..len-1]`, etc.

linera-views/src/lru_prefix_cache.rs
```diff
@@ -0,0 +1,1289 @@
...
+
+    /// Same as above but returns a mutable reference.
+    fn get_find_keys_lower_bound_mut(
+        &mut self,
+        key: &[u8],
+    ) -> Option<(&Vec<u8>, &mut FindKeysEntry)> {
+        match self.find_keys_map.range_mut(..=key.to_vec()).next_back() {
```
ma2bd, line 353: same

linera-views/src/lru_prefix_cache.rs
ma2bd: Can we have a unit test for this function?

linera-views/src/lru_prefix_cache.rs
```diff
@@ -0,0 +1,1289 @@
...
+                self.total_find_keys_size -= size;
+            }
+            CacheKey::FindKeyValues(_) => {
+                self.total_find_key_values_size -= size;
+            }
+        }
+    }
+
+    /// Increase the sizes of the keys.
+    fn increase_sizes(&mut self, cache_key: &CacheKey, size: usize) {
```
ma2bd, line 243: Should we merge these two functions into `fn update_sizes(&mut self, cache_key: &CacheKey, old_size: usize, new_size: usize)` ?

linera-views/src/lru_prefix_cache.rs
```diff
@@ -0,0 +1,1289 @@
...
+    /// Update the cache size to the new size without changing position.
+    fn update_cache_key_size(&mut self, cache_key: &CacheKey, new_size: usize) {
+        let size = self
+            .queue
+            .get_mut(cache_key)
+            .expect("cache_key should be present");
+        let old_size = *size;
+        *size = new_size;
+        self.decrease_sizes(cache_key, old_size);
+        self.increase_sizes(cache_key, new_size);
```
ma2bd: See previous comment. This would become: `self.update_sizes(cache_key, old_size, new_size)`

linera-views/src/lru_prefix_cache.rs
```diff
@@ -0,0 +1,1289 @@
...
+    }
+
+    /// Insert a cache_key into the queue and update sizes.
+    fn insert_cache_key(&mut self, cache_key: CacheKey, size: usize) {
+        self.increase_sizes(&cache_key, size);
```
ma2bd, line 298: See previous comment. This would become: `self.update_sizes(cache_key, 0, new_size)`

linera-views/src/lru_prefix_cache.rs
```diff
@@ -0,0 +1,1289 @@
...
+        };
+        self.insert_value(key, cache_entry)
+    }
+
+    /// Inserts a contains_key result into the cache.
+    pub fn insert_contains_key(&mut self, key: &[u8], result: bool) {
+        let cache_entry = match result {
+            false => ValueCacheEntry::DoesNotExist,
+            true => ValueCacheEntry::Exists,
+        };
```
ma2bd: nit: use `if-then-else` ?

linera-views/src/lru_prefix_cache.rs
```diff
@@ -0,0 +1,1289 @@
...
+    fn remove_cache_key(&mut self, cache_key: &CacheKey) {
+        let size = self
+            .queue
+            .remove(cache_key)
+            .expect("cache_key should be present");
+        self.decrease_sizes(cache_key, size);
+    }
+
+    /// Update the cache size to the new size without changing position.
+    fn update_cache_key_size(&mut self, cache_key: &CacheKey, new_size: usize) {
```
ma2bd: `_sizes` (plural) ?

linera-views/src/lru_prefix_cache.rs
```diff
@@ -0,0 +1,1289 @@
...
+        loop {
+            let value = iter.next();
+            let Some((cache_key, size)) = value else {
+                break;
+            };
+            if total_size < self.config.max_cache_value_size {
+                break;
+            }
+            if let CacheKey::Value(key) = cache_key {
+                total_size -= size;
```
ma2bd: Looks like we're double subtracting from `total_size` (here and in `remove_cache_key`)

linera-views/src/lru_prefix_cache.rs
```diff
@@ -0,0 +1,1289 @@
...
+        loop {
+            let value = iter.next();
+            let Some((cache_key, size)) = value else {
+                break;
+            };
+            if total_size < self.config.max_cache_find_keys_size {
+                break;
+            }
+            if let CacheKey::FindKeys(prefix) = cache_key {
+                total_size -= size;
```
ma2bd: same

linera-views/src/lru_prefix_cache.rs
```diff
@@ -0,0 +1,1289 @@
...
+        loop {
+            let value = iter.next();
+            let Some((cache_key, size)) = value else {
+                break;
+            };
+            if total_size < self.config.max_cache_find_key_values_size {
+                break;
+            }
+            if let CacheKey::FindKeyValues(prefix) = cache_key {
+                total_size -= size;
```
ma2bd: same

linera-storage-service/src/server.rs
```diff
@@ -282,12 +282,27 @@ enum StorageServerOptions {
         /// The maximum size of the cache, in bytes (keys size + value sizes)
         #[arg(long, default_value = "10000000")]
         max_cache_size: usize,
-        /// The maximum size of an entry size, in bytes
+        /// The maximum size of a value entry size, in bytes
         #[arg(long, default_value = "1000000")]
-        max_entry_size: usize,
+        max_value_entry_size: usize,
+        /// The maximum size of a findkeys entry size, in bytes
```
ma2bd:
```suggestion
        /// The maximum size of a find-keys entry, in bytes
```

linera-storage-service/src/server.rs
```diff
@@ -282,12 +282,27 @@ enum StorageServerOptions {
...
+        #[arg(long, default_value = "1000000")]
+        max_findkeys_entry_size: usize,
+        /// The maximum size of a findkeyvalues entry size, in bytes
```
ma2bd:
```suggestion
        /// The maximum size of a find-key-values entry, in bytes
```

linera-storage-service/src/server.rs
```diff
@@ -282,12 +282,27 @@ enum StorageServerOptions {
...
         #[arg(long, default_value = "10000000")]
         max_cache_size: usize,
-        /// The maximum size of an entry size, in bytes
+        /// The maximum size of a value entry size, in bytes
         #[arg(long, default_value = "1000000")]
-        max_entry_size: usize,
+        max_value_entry_size: usize,
+        /// The maximum size of a findkeys entry size, in bytes
+        #[arg(long, default_value = "1000000")]
+        max_findkeys_entry_size: usize,
```
ma2bd: `max_find_keys_entry_size`

linera-storage-service/src/server.rs
```diff
@@ -282,12 +282,27 @@ enum StorageServerOptions {
...
+        /// The maximum size of a findkeyvalues entry size, in bytes
+        #[arg(long, default_value = "1000000")]
+        max_findkeyvalues_entry_size: usize,
```
ma2bd: max_find_key_values_entry_size

linera-views/src/lru_prefix_cache.rs
```diff
@@ -0,0 +1,1289 @@
...
+
+    /// Insert a cache_key into the queue and update sizes.
+    fn insert_cache_key(&mut self, cache_key: CacheKey, size: usize) {
+        self.increase_sizes(&cache_key, size);
+        self.queue.insert(cache_key, size);
+    }
+
+    /// If the FindKeys map contain a prefix that is a prefix of key in argument,
+    /// then returns it and the corresponding FindKeys. Otherwise `None`.
+    fn get_find_keys_lower_bound(&self, key: &[u8]) -> Option<(&Vec<u8>, &FindKeysEntry)> {
```
ma2bd: `get_find_keys_lower_bound` --> `get_existing_find_keys_entry`

linera-views/src/lru_prefix_cache.rs
```diff
@@ -0,0 +1,1289 @@
...
+}
+
+/// Stores the data for simple `read_values` queries.
+///
+/// This data structure is inspired by the crate `lru-cache` but was modified to support
+/// range deletions.
+pub(crate) struct LruPrefixCache {
+    value_map: BTreeMap<Vec<u8>, ValueCacheEntry>,
+    find_keys_map: BTreeMap<Vec<u8>, FindKeysEntry>,
+    find_key_values_map: BTreeMap<Vec<u8>, FindKeyValuesEntry>,
```
ma2bd: Can the comments include a description of the invariants of this structure?
For instance: are the keys in `find_keys_map` are never prefixes of each other?

linera-views/src/lru_prefix_cache.rs
```diff
@@ -0,0 +1,1289 @@
...
+        self.insert_value(key, cache_entry)
+    }
+
+    /// Inserts the result of `find_keys_by_prefix` in the cache.
+    pub fn insert_find_keys(&mut self, key_prefix: Vec<u8>, keys: &[Vec<u8>]) {
+        if self.config.max_findkeys_entry_size == 0 {
+            // zero max size, exit from the start
+            return;
+        }
+        let size = key_prefix.len() + keys.iter().map(|k| k.len()).sum::<usize>();
```
ma2bd: nit: remove `<usize>`. Consider `.map(Vec::len)`

linera-views/src/lru_prefix_cache.rs
```diff
@@ -0,0 +1,1289 @@
...
+        }
+    }
+
+    /// Inserts a read_value result into the cache.
+    pub fn insert_read_value(&mut self, key: &[u8], value: &Option<Vec<u8>>) {
+        let cache_entry = match value {
+            None => ValueCacheEntry::DoesNotExist,
+            Some(vec) => ValueCacheEntry::Value(vec.to_vec()),
+        };
+        self.insert_value(key, cache_entry)
```
ma2bd, line 601: We don't call `get_find_keys_lower_bound_mut` here. Is this intended? (add a comment?)

linera-views/src/lru_prefix_cache.rs
```diff
@@ -0,0 +1,1289 @@
...
+            .range(get_interval(key_prefix))
+            .map(|key| key[delta..].to_vec())
+            .collect()
+    }
+
+    fn contains_key(&self, key: &[u8]) -> bool {
+        self.0.contains(key)
+    }
+
+    fn update_entry(&mut self, key: &[u8], new_value: Option<Vec<u8>>) {
```
ma2bd: `new_value: bool` would avoid a `Clone` in `update_find_entries`.

linera-views/src/lru_prefix_cache.rs
```diff
@@ -0,0 +1,1289 @@
...
+            .range(get_interval(key_prefix.clone()))
+            .map(|(x, _)| x.clone())
+            .collect::<Vec<_>>();
+        for key in keys {
+            self.find_keys_map.remove(&key);
+            let cache_key = CacheKey::FindKeys(key);
+            self.remove_cache_key(&cache_key);
+        }
+        // Clearing up the value entries as they are covered by the new FindKeys.
+        // That is the exists or not. The Value entries are not covered by FindKeys.
```
ma2bd:
```
that is, the `Exists` and `DoesNotExist`
```

linera-views/src/lru_prefix_cache.rs
```diff
@@ -0,0 +1,1289 @@
...
+            }
+            // No find keys for shared access
+        }
+    }
+
+    /// Returns the cached value, or `Some(None)` if the entry does not exist in the
+    /// database. If `None` is returned, the entry might exist in the database but is
+    /// not in the cache.
+    pub fn query_read_value(&mut self, key: &[u8]) -> Option<Option<Vec<u8>>> {
+        // First querying the value_map
```
ma2bd:
```suggestion
        // First, query the value map
```

linera-views/src/lru_prefix_cache.rs
ma2bd: Even better, can we create a function that asserts all these global invariants, and use it in unit tests?

linera-views/src/lru_prefix_cache.rs
```diff
@@ -0,0 +1,1289 @@
...
+                // the deleted prefix as a FindKeyValues.
+                let size = key_prefix.len();
+                let cache_key = CacheKey::FindKeyValues(key_prefix.to_vec());
+                let find_entry = FindKeyValuesEntry(BTreeMap::new());
+                self.find_key_values_map
+                    .insert(key_prefix.to_vec(), find_entry);
+                self.insert_cache_key(cache_key, size);
+            }
+        } else {
+            // Just forget about the entries.
```
ma2bd:
```suggestion
            // Just forget about the entries in the value map.
```

linera-views/src/lru_prefix_cache.rs
```diff
@@ -0,0 +1,1289 @@
...
+            }
+        };
+        let cache_key = CacheKey::FindKeyValues(lower_bound.to_vec());
+        self.put_cache_key_on_top(cache_key);
+        Some(result)
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
ma2bd, line 951: Would it be possible to evaluate the test coverage? We want to make sure that all paths are covered here. See also other comment.

ma2bd (commented):
Nice work. The most serious comment is a possible issue with `get_find_key(_value)s_lower_bound`.

I would also like to make the tests more strict (asserting all invariants) and ideally reach 100% test coverage.

linera-views/src/lru_prefix_cache.rs
MathieuDutSik: Ok.

linera-views/src/lru_prefix_cache.rs
MathieuDutSik: Yes, that is correct.

linera-views/src/lru_prefix_cache.rs
MathieuDutSik: Yes agreed. Also, remove the `pub(crate)`.

linera-views/src/lru_prefix_cache.rs
MathieuDutSik: Ok.

linera-views/src/backends/lru_caching.rs
MathieuDutSik: Ok.

linera-views/src/lru_prefix_cache.rs
MathieuDutSik: Ok. Used `prefix_len` as in other parts of the code.

linera-views/src/lru_prefix_cache.rs
MathieuDutSik: Ok.

linera-views/src/lru_prefix_cache.rs
MathieuDutSik: Ok.

linera-views/src/lru_prefix_cache.rs
MathieuDutSik: Ok.

linera-views/src/lru_prefix_cache.rs
MathieuDutSik: Yes.

linera-views/src/lru_prefix_cache.rs
MathieuDutSik: Yes, we can use `Vec::len`.
However, the type annotation for `sum` is needed as the compiler cannot deduce the type otherwise.

linera-views/src/lru_prefix_cache.rs
MathieuDutSik: Ok.
And done for other functions.

linera-storage-service/src/server.rs
MathieuDutSik: Ok, done.

linera-storage-service/src/server.rs
MathieuDutSik: Ok.

linera-views/src/lru_prefix_cache.rs
MathieuDutSik: Yes, we can this assert statements.

linera-views/src/lru_prefix_cache.rs
MathieuDutSik: Yes, of course.
I thought that we wanted to avoid getter functions.

linera-views/src/lru_prefix_cache.rs
MathieuDutSik: Yes, we can, but why? And if yes, what should be the logic of ordering the types.

The way I order code is inspired by C++. A type is defined before it is used. So, first the type, then the `impl`. Then another code. So, the config type is on top, the `ValueEntry`, `FindKeysEntry`, `FindKeyValuesEntry` are then. 

Then, at the end of the file, we have the tests.

I am, of course, well aware that Rust is not C++ and that the ordering can be any one we like. But if not C++ like ordering, then what logic should we use for the ordering of the code?

linera-views/src/lru_prefix_cache.rs
MathieuDutSik: Ok, done.

linera-views/src/lru_prefix_cache.rs
MathieuDutSik: Ok.

linera-views/src/lru_prefix_cache.rs
MathieuDutSik: Ok.

linera-views/src/lru_prefix_cache.rs
MathieuDutSik: Ok.

linera-views/src/lru_prefix_cache.rs
MathieuDutSik: No.
We are copying the `total_value_size`, not taking a reference. Then we determine which `CacheKey::Value` needs to be removed. But that does not modify the `self.total_value_size`. The keys to be removed are then determined.

Then, in the second loop, we are dropping the keys. That is where the sizes are decreased.

To avoid confusion, the `total_size` is renamed as `control_size`.

linera-views/src/lru_prefix_cache.rs
MathieuDutSik: Same answer.

linera-views/src/lru_prefix_cache.rs
MathieuDutSik: Same answer.

linera-views/src/lru_prefix_cache.rs
MathieuDutSik: Thanks.
Very good find.

linera-storage-service/src/server.rs
MathieuDutSik: Ok.

linera-storage-service/src/server.rs
MathieuDutSik: Ok.

linera-views/src/lru_prefix_cache.rs
MathieuDutSik: Ok, done.

linera-views/src/lru_prefix_cache.rs
MathieuDutSik: I did use the `update_sizes` but nest it in the `increase_sizes`.

linera-views/src/lru_prefix_cache.rs
MathieuDutSik: We can do that. This makes the code slightly more efficient.

linera-views/src/lru_prefix_cache.rs
MathieuDutSik: I added a test.
However, not a unit test, which I think cannot really test the whole possibility of scenarios.
Instead, I added a random test named `test_lower_bounds` that generates prefix-free sets
at random and does the checks repeatedly.

linera-views/src/lru_prefix_cache.rs
MathieuDutSik: That is exactly the same algorithm.
So, what is the point of adding more checks?

linera-views/src/lru_prefix_cache.rs
ma2bd: no big deal

linera-views/src/lru_prefix_cache.rs
ma2bd: It's a longer discussion

linera-views/src/lru_prefix_cache.rs
ma2bd: Do we agree that this code is broken?

linera-views/src/lru_prefix_cache.rs
ma2bd: My point is to fix the code

linera-storage-service/src/server.rs
ma2bd: Nope

linera-views/src/lru_prefix_cache.rs
ma2bd: where?

linera-views/src/lru_prefix_cache.rs
MathieuDutSik: Yes, it is possible.
The data is relatively hard to read, but it can be obtained.
Method is
```rust
cargo install cargo-llvm-cov
cargo llvm-cov --html
```

It indicates 91% coverage for `lru_prefix_cache.rs`.
Requiring 100% coverage is a tall order on CI. What I have seen in other groups is that the merged Pull Request should increase the coverage.

Of course, I think adding test coverage should be in another PR.

MathieuDutSik: So, I added more tests, and now we are at 95% coverage.
I also resolved some bugs on the way.

linera-views/src/lru_prefix_cache.rs
MathieuDutSik: Yes, did that. It is called `check_coherence`.

linera-views/src/lru_prefix_cache.rs
MathieuDutSik: The example that you provide with keys "a" and "ab" is not prefix-free.

ma2bd: I see

linera-views/src/lru_prefix_cache.rs
MathieuDutSik: If useful had been available in the FindKeys / FindKeyValues, then it would have matched before, and so there would not have been any need to insert in the cache. Comments have been added.

linera-views/src/lru_prefix_cache.rs
MathieuDutSik: Now, at 98.25% but actually every single line of the code passes by the coverage.
So, we have to see how to continue further.

linera-views/src/lru_prefix_cache.rs
```diff
@@ -0,0 +1,2590 @@
...
+        self.update_sizes(cache_key, old_size, new_size);
+    }
+
+    /// Insert a cache_key into the queue and update sizes.
+    fn insert_cache_key(&mut self, cache_key: CacheKey, size: usize) {
+        self.increase_sizes(&cache_key, size);
+        assert!(self.queue.insert(cache_key, size).is_none());
+    }
+
+    /// If the FindKeys map contain a prefix that is a prefix of key in argument,
```
ma2bd: typo: contains

linera-views/src/lru_prefix_cache.rs
```diff
@@ -0,0 +1,2590 @@
...
+    /// First of all the expression `self.find_keys_map.range(..=key.to_vec())` corresponds
+    /// to S and `self.find_keys_map.range(..=key.to_vec()).next_back()` is
+    /// * None if S is empty.
+    /// * Some(M) with M the maximum of S if S is non-empty.
+    ///
+    /// If there is a prefix `p` of `key` in `S` then we have `p <= key` in the lexicographic
+    /// ordering. So, S is non-empty. That prefix `p` is the highest element of `S`. If that
+    /// were not the case, we would have for another `q in S` the inequality `p < q <= key`.
+    /// This forces `q` to be a prefix of `key` and `p` to be a prefix of `q`. This breaks
+    /// the prefix-free property.
```
ma2bd: "of S."

linera-views/src/lru_prefix_cache.rs
```diff
@@ -0,0 +1,2617 @@
...
+    ///
+    /// If there is no prefix `p` of `key` in `S` then if `S` is empty the code is correct.
+    /// If it is not empty, then it has a maximum element M. That highest element cannot
+    /// be a prefix of `key` and that is checked in the code (named `key_store` there).
+    /// If there were an element `h` of `S` that would be a prefix, then the inequality
+    /// `h <= M <= key` would force `M` to also be a prefix of `key`.
+    ///
+    /// Therefore the code does return an `Some(p)` if and only if there is a prefix `p`
+    /// in the map of find-keys prefixes.
```
ma2bd: 1) nit: `key_store` is not an amazing name. I propose to rename it `stored_key`.

2) Thanks for proposing a proof. FWIW I put together another one which I find a little easier to follow (but I'm obviously biased):

> Claim: `self.get_existing_find_keys_entry(key)` finds a prefix of `key` in `self.find_keys_map.keys()` iff such prefix exists.
>
> Note that when it exists, such a prefix is unique because `self.find_keys_map.keys()` is prefix-free.
> 
> Proof: First, if `self.get_existing_find_keys_entry(key) == Some(stored_key)` then given the code, clearly `stored_key` is a prefix of `key`.
> 
> Conversely, let's assume that `self.find_keys_map.keys()` contains a certain prefix `p` of `key`. Because in particular `p <= key`, we have `self.find_keys_map.range(..=key.to_vec()).next_back() == Some((stored_key, _))` for some value `stored_key` such that `p <= stored_key <= key`.
> 
> Next, let's prove that `p` is a prefix of `stored_key`. (This will entail `stored_key == p` due to the prefix-free property of `self.find_keys_map`, therefore the algorithm's answer is correct.)
> 
> By contradiction. Let `w` be the longest common prefix between `p` and `stored_key`. If `p = w p2` and `stored_key = w s2` with `p2` non-empty, then `p <= stored_key` implies `s2 > p2`. But `p` is also a prefix of `key` therefore `stored_key > key`, contradiction.

linera-views/src/lru_prefix_cache.rs
```diff
@@ -0,0 +1,2617 @@
...
+        self.trim_value_cache();
+        self.trim_cache();
+    }
+
+    /// Puts a key/value in the cache.
+    pub(crate) fn put_key_value(&mut self, key: &[u8], value: &[u8]) {
+        if self.has_exclusive_access {
+            let lower_bound = self.get_existing_keys_entry_mut(key);
+            if let Some((lower_bound, cache_entry)) = lower_bound {
+                let key_red = &key[lower_bound.len()..];
```
ma2bd: nit: what does `key_red` mean? Did you mean `reduced_key` or just `suffix`?

linera-views/src/lru_prefix_cache.rs
```diff
@@ -0,0 +1,2617 @@
...
+                cache_entry.update_entry(key_red, true);
+                let cache_key = CacheKey::FindKeys(lower_bound.to_vec());
```
ma2bd, line 533: Did you mean `cached_key` ?

linera-views/src/lru_prefix_cache.rs
```diff
@@ -0,0 +1,2617 @@
...
+            let find_key_values_map_set = self
+                .find_key_values_map
+                .keys()
+                .cloned()
+                .collect::<BTreeSet<Vec<u8>>>();
+            // Checking prefix-free ness of the find-keys-map / find-key-values-map
+            check_prefix_free(find_keys_map_set.clone());
+            check_prefix_free(find_key_values_map_set.clone());
+            // Building the data set and sizes.
+            // Also checking that the sizes are coherent.
```
ma2bd, line 1003: This is great because size miscalculations could easily cause a panic due to a `usize`-underflow.

linera-views/src/lru_prefix_cache.rs
MathieuDutSik: Ok.

linera-views/src/lru_prefix_cache.rs
MathieuDutSik: Ok.

linera-views/src/lru_prefix_cache.rs
MathieuDutSik: Changed to `reduced_key`.

linera-views/src/lru_prefix_cache.rs
MathieuDutSik: No.
First, the type is `CacheKey`. Secondly, it is really a key in the `queue` and the various maps. Thirdly, what is cached may be a key, but it can be a prefix.

linera-views/src/lru_prefix_cache.rs
MathieuDutSik: Ok, I adapted it.

PR #4577, opened 2025-09-15
Restructure `linera-web` directory

Twey:
## Motivation

We are unhappy with the directory structure of `linera-web`, which looks too much like an SDK crate.

## Proposal

Rename it to plain old `web` to indicate that it's not (published as) a crate, with the npm package structure (`@linera/client` and `@linera/signer`) reproduced inside.

## Test Plan

CI.

## Release Plan

## Links


ma2bd (approved):
Thanks!

PR #4581, opened 2025-09-16
Add support to multiple proxies on net up --kubernetes

ndr-ds:
## Motivation

Currently `net up --kubernetes` always runs with just one proxy.

## Proposal

Add support for multiple proxies.

## Test Plan

Ran a local network with more than 1 proxy, everything seems to work as expected.

## Release Plan

- Nothing to do / These changes follow the usual release cycle.

PR #4410, opened 2025-08-26
Bring back the kind remote net tests

ndr-ds:
## Motivation

We removed this a while back, and never added it back. This is useful to test the changes against a kubernetes network, running ScyllaDB as well.

## Proposal

Test against a kind network again.

## Test Plan

CI

## Release Plan

- Nothing to do / These changes follow the usual release cycle.

PR #4582, opened 2025-09-16
Add block explorer to net up --kubernetes

ndr-ds:
## Motivation

We're working on adding the Indexer/Explorer/Exporter into our deployments.

## Proposal

This adds them to `linera net up --kubernetes` as a starting point. Next, adding to the actual GCP deployment.

## Test Plan

Started a local k8s network, sent some transactions to it, and managed to see the information in the block explorer:

## Release Plan


ndr-ds:
* **#4582** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4582?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4582?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#4581** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4581?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

linera-service/src/cli_wrappers/local_kubernetes_net.rs
```diff
@@ -129,6 +135,9 @@ impl SharedLocalKubernetesNetTestingConfig {
             docker_image_name: String::from("linera:latest"),
             build_mode: BuildMode::Release,
             policy_config: ResourceControlPolicyConfig::Testnet,
+            num_block_exporters: 0,
```
deuszx, line 138: So we run  the indexer, explorer but not exporter? it won't work.

linera-service/src/cli/net_up_utils.rs
```diff
@@ -149,6 +153,12 @@ pub async fn handle_net_up_kubernetes(
     let shutdown_notifier = CancellationToken::new();
     tokio::spawn(listen_for_shutdown_signals(shutdown_notifier.clone()));
 
+    let num_block_exporters = if with_block_exporter {
+        num_block_exporters
```
deuszx, line 161: We could assert here (or earlier) that `num_block_exporters` is non-zero.

linera-service/src/cli_wrappers/local_kubernetes_net.rs
ndr-ds: `SharedLocalKubernetesNetTestingConfig` is some unused thing that will be removed once I'm done with https://github.com/linera-io/linera-protocol/pull/4410. Maybe the name here is not good, because if `num_block_exporters` is zero, we don't run any of them.

kubernetes/linera-validator/templates/indexer.yaml
```diff
@@ -0,0 +1,83 @@
...
+      containers:
+        - name: linera-indexer
+          image: {{ .Values.indexer.image }}
+          imagePullPolicy: {{ .Values.indexer.imagePullPolicy }}
+          command:
+            - "./linera-indexer-grpc"
+            - "--port"
+            - "{{ .Values.indexer.port }}"
+            - "--database-path"
+            - "{{ .Values.indexer.databasePath }}"
```
deuszx, line 46: for the future: Can we make it so that the SQLite file is replicated automatically?

kubernetes/linera-validator/exporter-config.toml.tpl
```diff
@@ -0,0 +1,29 @@
...
+
+[[destination_config.destinations]]
+file_name = "/data/linera-exporter.log"
+kind = "Logging"
+
+[[destination_config.destinations]]
+kind = "Indexer"
+tls = "ClearText"
+port = {{ .Values.indexer.port }}
+endpoint = "linera-indexer"
```
deuszx, line 20: This works only in docker-compose. In real deployment this will have to be an address of an actual indexer endpoint.

kubernetes/linera-validator/templates/indexer.yaml
ndr-ds: We could! Keep in mind that regular PVCs are zonal deployments, so only if the entire zone goes down for some reason we would lose the data. It's very unlikely. If we want extra protection, we could do regional PVCs, that replicate the data across different zones of the same region.
But yeah, I agree it's more of a follow up for the future

kubernetes/linera-validator/exporter-config.toml.tpl
ndr-ds: Actually on kubernetes this will be the actual internal hostname of the indexer endpoint :) so this works

deuszx (approved):

kubernetes/linera-validator/templates/indexer.yaml
deuszx: To complete the question: we have sqlite db which is file-based. We should be set up some syncing/fallbacks in case the machine is wiped or lost.

ndr-ds: Like I said, it's currently in a PVC. PVCs are over the network disks. They already don't live in the actual indexer VM. So if the machine is moved or something (like in a maintenance event), we won't lose the data.
I think we should be ok here for now, losing this data should be a very low probability thing (our ScyllaDB for example uses the same over the network disks for testnet). Of course there's more we can do in the future to add more safeguards for it like I mentioned (replicated PVC, backups, etc)

deuszx: ok cool. Do we have metrics for latency there?

PR #4592, opened 2025-09-17
Implement the fees for the events.

### LINKED ISSUE #4259: Add separate fees categories for emitting and reading events.

MathieuDutSik:
## Motivation

Events costs need to be accounted for.

Fixes #4259

## Proposal

Since events are implemented as blobs, it makes sense to have their costs identical to those of blobs.
In that way, we avoid having some economic imbalance.

## Test Plan

The CI.

## Release Plan

This does change the costs for validators and so it is a breaking change.

## Links

None.

### DIFF
--- a/linera-execution/src/execution_state_actor.rs
+++ b/linera-execution/src/execution_state_actor.rs
@@ -493,6 +493,10 @@ where
                     .await?;
                 let index = *count;
                 *count = count.checked_add(1).ok_or(ArithmeticError::Overflow)?;
+                self.resource_controller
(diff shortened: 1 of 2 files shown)


deuszx:
> Since events are implemented as blobs, it makes sense to have their costs identical to those of blobs.
In that way, we avoid having some economic imbalance.

I remember some data is stored per shards vs on the worker nodes (for example blobs are available to everyone while blocks are on shard). How does it look like with events?

deuszx (approved):
Looks good. Ideally we'd have some tests as well for this 🤔

Issue #4562, opened 2025-09-13
fungible app only supports 20-byte account owners


ma2bd:
This doesn't work for me in the UI:

```
0x00@836f13b3815abdb4df2fcb47dca2dc4a7595c2557587c359d3dee97380fa09fa
0xc271f3fe7ce54619f5eb427bb77948097e76fcbfe7b0f82d9d88697d88accd19@836f13b3815abdb4df2fcb47dca2dc4a7595c2557587c359d3dee97380fa09fa
```

deuszx:
I'm gonna take it.

PR #4599, opened 2025-09-18
Native fungible demo accepts 32-byte long address

deuszx:
## Motivation

Fungible demo frontend accepted only 20-byte accounts.

## Proposal

Modify the regex to also accept 32-bytes (64 chars).

## Test Plan

Manual.

## Release Plan

This is a testnet branch. Updates fungible demo can be released from it.

## Links
Solves https://github.com/linera-io/linera-protocol/issues/4562

### DIFF
--- a/examples/native-fungible/index.html
+++ b/examples/native-fungible/index.html
@@ -171,8 +171,12 @@
 
           let errors = [];
           try {
-              const match = recipient.match(/^(0x[0-9a-f]{40})@([0-9a-f]{64})$/);
+              // I failed to make it work with a single regex,
+              // the regex with group alternatives for the owner produces different `match` array
+              // with different group indices so we can't extract it deterministicly when parsing as `recipient` down below.
+              const match = recipient.match(/^(0x[0-9a-f]{40})@([0-9a-f]{64})$/) || recipient.match(/^(0x[0-9a-f]{64})@([0-9a-f]{64})$/);
               if (!match) throw new Error('Invalid recipient address: expected `owner@chain_id`');
+              console.log('Match: ', match);
 
               const query = gql(`mutation(
                   $donor: AccountOwner!,
--- a/examples/native-fungible/index.html
+++ b/examples/native-fungible/index.html
@@ -171,10 +171,7 @@
 
           let errors = [];
           try {
-              // I failed to make it work with a single regex,
-              // the regex with group alternatives for the owner produces different `match` array
-              // with different group indices so we can't extract it deterministicly when parsing as `recipient` down below.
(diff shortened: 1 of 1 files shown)


examples/native-fungible/index.html
```diff
@@ -171,8 +171,12 @@ <h2>Chain history for <code id="chain-id" class="hex">requesting a new microchai
 
           let errors = [];
           try {
-              const match = recipient.match(/^(0x[0-9a-f]{40})@([0-9a-f]{64})$/);
+              // I failed to make it work with a single regex,
+              // the regex with group alternatives for the owner produces different `match` array
+              // with different group indices so we can't extract it deterministicly when parsing as `recipient` down below.
+              const match = recipient.match(/^(0x[0-9a-f]{40})@([0-9a-f]{64})$/) || recipient.match(/^(0x[0-9a-f]{64})@([0-9a-f]{64})$/);
```
Twey: I think the way to do it with one regex is

```suggestion
              const match = recipient.match(/^(0x[0-9a-f]{40}|0x[0-9a-f]{64})@([0-9a-f]{64})$/);
```

No big deal though.

examples/native-fungible/index.html
```diff
@@ -171,8 +171,12 @@ <h2>Chain history for <code id="chain-id" class="hex">requesting a new microchai
...
               if (!match) throw new Error('Invalid recipient address: expected `owner@chain_id`');
+              console.log('Match: ', match);
```
Twey: Vestigial debugging line?

Twey: (if you think this line is likely to be useful for debugging problems in production, please use `console.debug`; otherwise we can just remove it)

examples/native-fungible/index.html
deuszx: Yes, I tried this. It then returned 5 matched groups, but not always (depending on whether it was 32-byte or 20-byte address). That's the comment above.

examples/native-fungible/index.html
deuszx: No, leftover.

deuszx: Removed.

examples/native-fungible/index.html
Twey: I suspect you put the capture group inside the alternative — I tested this one and it works okay.

Twey (approved):

examples/native-fungible/index.html
deuszx: Ah, that might be - yes. I'll redo.

Twey: Oh, we also support uppercase (and mixes, for checksum reasons) so it might be worth adding the `/i` flag here.

PR #4603, opened 2025-09-18
Small fixes after #4570

bart-linera:
## Motivation

Changes from #4570 left a few small oversights.

## Proposal

Fix an incorrect doc string, restore a deleted line from Cargo.toml, change the Debug implementation for CryptoHash (so that the commands printing out chain IDs are actually useful).

## Test Plan

CI

## Release Plan

## Links

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,7 +56,6 @@ default-members = [
     "linera-summary",
     "linera-views",
     "linera-views-derive",
-    # "linera-web", # not this one
     "linera-witty",
     "linera-witty-macros",
     "linera-witty/test-modules",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,6 +56,7 @@ default-members = [
     "linera-summary",
     "linera-views",
     "linera-views-derive",
+    # "linera-web", # not this one
     "linera-witty",
     "linera-witty-macros",
     "linera-witty/test-modules",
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -56,6 +56,7 @@ default-members = [
     "linera-summary",
     "linera-views",
     "linera-views-derive",
+    # "linera-web", # not this one
```
ma2bd: It's going to be `web` anyway, right?

bart-linera: To be honest, I have no idea why that line is being left here, but we are always taking care not to remove it. I think it's useful to @Twey for some reason.

linera-base/src/unit_tests.rs
```diff
@@ -160,9 +160,9 @@ fn chain_ownership_test_case() -> ChainOwnership {
 fn account_owner_debug_format() {
     assert_eq!(&format!("{:?}", AccountOwner::Reserved(10)), "Reserved(10)");
     let addr32 = AccountOwner::Address32(CryptoHash::from([10u8; 32]));
-    let debug32 = "Address32(0a0a0a0a0a0a0a0a..)";
+    let debug32 = "Address32(0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a..)";
```
deuszx: This is full address so no `..` should be necessary at the end 🤔

bart-linera: Ah, fair point!

Cargo.toml
ma2bd: No you can remove

bart-linera: I'll leave it in for now, but if we can confirm that it doesn't serve any purpose (I seem to recall that someone explained the purpose to me when I accidentally removed it in some PR, but I don't remember the details), I can open another PR removing it.

ma2bd: `linera-web` doesn't exist any more after #4577

bart-linera: Ohh! I see, I didn't understand at first. :man_facepalming: Okay, I'll remove it, then, thanks for the clarification!

PR #4604, opened 2025-09-18
Much faster memleak_translate.sh

ndr-ds:
## Motivation

The current version of the script uses `addr2line`, which is incredibly slow.

## Proposal

Use `llvm-symbolizer` instead, which is much faster, and cache the translations so we don't translate the same address twice.

## Test Plan

Tested this while debugging the GoL backend memory leak, and it's way way faster

## Release Plan


ndr-ds:
* **#4604** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4604?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4604?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

eldios (approved):

scripts/memleak_translate.sh
```diff
@@ -2,42 +2,125 @@
...
 [[ -n $BASE ]] || { echo "could not find base address for $exe" >&2; exit 1; }
 
-# helper: translate <$addr> using addr2line, compensating for PIE base
-translate () {
-    local addr=$1
-    addr=${addr//[<>[\]]}
-    [[ $addr != 0x* ]] && addr="0x${addr}"
-    local rel=$(printf "0x%x" $(( addr - BASE )))
-    addr2line -e "$exe" -f -C -p "$rel" 2>/dev/null || echo "$addr"
-}
```
Twey, line 33: The problem here is just that we're invoking `addr2line` for every line.  The `addr2line` startup time is quite slow, but the line processing time is fast — so we should use a long-running `addr2line` process.  The same probably applies to `llvm-symbolizer`.

ndr-ds: Yup, that's what I'm doing now :) and `llvm-symbolizer` is supposed to be faster than `addr2line` I think

PR #4606, opened 2025-09-19
Add macOS support for memory profiling

ndr-ds:
## Motivation

We're using the `unprefixed_malloc_on_supported_platforms` feature to be able to use `malloc_conf` instead of the prefixed `_rjem_malloc_conf`, as it's recommended, but MacOS is not one of the supported platforms.

## Proposal

Use the prefixed version for MacOS

## Test Plan

Ran `linera net up` with memory profiling enabled from my MacOS, it works now

## Release Plan

- Nothing to do / These changes follow the usual release cycle.

PR #4607, opened 2025-09-19
Add monitoring to node service

ndr-ds:
## Motivation

Currently, if we start a node service we get no metrics or memory profiling.

## Proposal

Add support for metrics and memory profiling from node service. Also fix the fact that it doesn't terminate with Ctrl+C (the faucet had the same thing, so I fixed it there too).

## Test Plan

Start a node service locally, see that memory profile is enabled and working, and that killing it with Ctrl+C actually kills it

## Release Plan

### DIFF
--- a/linera-metrics/src/monitoring_server.rs
+++ b/linera-metrics/src/monitoring_server.rs
@@ -15,8 +15,6 @@ pub fn start_metrics(
     address: impl ToSocketAddrs + Debug + Send + 'static,
     shutdown_signal: CancellationToken,
 ) {
-    info!("Starting to serve metrics on {:?}", address);
-
     #[cfg(feature = "memory-profiling")]
     let app = {
         // Try to add memory profiling endpoint
@@ -42,7 +40,13 @@ pub fn start_metrics(
     let app = Router::new().route("/metrics", get(serve_metrics));
 
     tokio::spawn(async move {
-        if let Err(e) = axum::serve(tokio::net::TcpListener::bind(address).await.unwrap(), app)
+        let listener = tokio::net::TcpListener::bind(address)
+            .await
+            .expect("Failed to bind to address");
+        let address = listener.local_addr().expect("Failed to get local address");
+
+        info!("Starting to serve metrics on {:?}", address);
+        if let Err(e) = axum::serve(listener, app)
             .with_graceful_shutdown(shutdown_signal.cancelled_owned())
             .await
         {
--- a/linera-faucet/server/src/lib.rs
+++ b/linera-faucet/server/src/lib.rs
@@ -702,7 +702,9 @@ where
         let batch_processor_task = batch_processor.run(cancellation_token.clone());
         let tcp_listener =
             tokio::net::TcpListener::bind(SocketAddr::from(([0, 0, 0, 0], port))).await?;
-        let server = axum::serve(tcp_listener, app).into_future();
+        let server = axum::serve(tcp_listener, app)
+            .with_graceful_shutdown(cancellation_token.cancelled_owned())
+            .into_future();
         futures::select! {
             result = Box::pin(chain_listener).fuse() => result?,
             _ = Box::pin(batch_processor_task).fuse() => {},
--- a/linera-service/src/cli/command.rs
+++ b/linera-service/src/cli/command.rs
@@ -742,6 +742,11 @@ pub enum ClientCommand {
         /// The port on which to run the server
         #[arg(long)]
         port: NonZeroU16,
+
+        /// The port to expose metrics on.
+        #[cfg(with_metrics)]
+        #[arg(long)]
+        metrics_port: NonZeroU16,
     },
 
     /// Run a GraphQL service that exposes a faucet where users can claim tokens.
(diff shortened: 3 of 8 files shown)


ndr-ds:
* **#4607** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4607?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4607?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#4606** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4606?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

ma2bd (approved):

linera-metrics/src/monitoring_server.rs
```diff
@@ -42,7 +40,13 @@ pub fn start_metrics(
...
 
     tokio::spawn(async move {
-        if let Err(e) = axum::serve(tokio::net::TcpListener::bind(address).await.unwrap(), app)
+        let listener = tokio::net::TcpListener::bind(address)
+            .await
+            .expect("Failed to bind to address");
+        let address = listener.local_addr().expect("Failed to get local address");
+
+        info!("Starting to serve metrics on {:?}", address);
+        if let Err(e) = axum::serve(listener, app)
```
deuszx, line 49: This wasn't needed - you have the `address` above, it's passed in.

ndr-ds: Yes, it was! We pass port 0 on the client, if we print it above, we don’t know what port it actually binded to, we just see 0

deuszx: but you have `metrics_port` argument that you're threading through. Why are you not using it here?

ndr-ds: `metrics_port` will be passed in as part of the `address`. And it can be zero. Port zero just tells the OS to bind to some available port, so we won’t know the port in advance, only after binding.

deuszx: I know. My point is that either:
1. you pass in a non-zero port that you MUST bind to (b/c your infra expects metrics to be ther) or,
2. you don't care, in which case you pass _nothing_ and you let the app decide.

In both cases _here_ we should already know what port it is. We have helper somewhere that finds first available port for such case.

ndr-ds: Any “helper” that doesn’t just do port 0 (most don’t) are racey 🙂
They’ll give you an available port, but in between getting the port and actually binding to it, something else can bind to it before you do. Using port 0 is the most reliable and standard way of doing this.
And I do care what port I bind to, because I need it to query the endpoint for metrics/flamegraphs/etc. So I need to make sure the correct port is printed.

deuszx: Then why not tell the app the specific port you expect it to bind, rather than extract it from the logs. We were doing it in a couple of places (with chain and account addresses) and it breaks things in most unexpected ways.

ndr-ds: Sure, we could do that. But I think the user should be able to just specify port 0 if he wants, which means we need to print the address down here anyways. And I honestly don’t really see what’s the harm of it 😅

deuszx: I think it's simply better engineering. You need to know the port, so why not specify it. Parsing logs to extract non-deterministic addresses is a bad pattern. Also, how will you do it in the cloud? will you be parsing logs from GCP machines?

ndr-ds: Port 0 is of course meant for when running the client locally. If you don’t care what port it is, and want to avoid specifying a port already in use accidentally, you can use port 0 and get it from the logs. 
For example, if I’m running 15 benchmark client instances locally, you want me to manually specify 15 ports? How is that better engineering exactly? Specially if I’m using multi benchmark, I just want to use port 0

deuszx: I said it at the beginning – if you don't care, don't pass in ANYTHING. Just don't pass 0.

ndr-ds: To reach 1M TPS with the realistic benchmarks, we’ll potentially have to run hundreds of clients

ndr-ds: And use a racey “find free port” helper? Like I said, in that case we might as well use port 0. Not sure why you’re so against it. Port 0 is widely used everywhere for this type of scenario

deuszx: the CLI arguments are the ultimate way to _override_ configs. The default behavior is that program finds a random port (`0`). If you pass in `0` it's not overwriting anything, I think it's an anti-pattern.

ndr-ds: I see where the confusion is. I meant that we make `metrics_port` always have to be provided. Then the user can willingly specify port 0, if he so chooses, to leave the port selecting to the OS

PR #4609, opened 2025-09-19
Allow multi-read in `CollectionView`

### LINKED ISSUE #4588: `linera-views`: nested `CollectionView` GraphQL queries are not supported
A nested query on `CollectionView`s like

```graphql
query {
    reportedSolutions {
        entries(input:{}) {
            key
            value {
                entries(input: {}) {
                    key
                    value
                }
            }
        }
    }
}
```

where `reported_solutions: CollectionView<_, _, CollectionView<_, _, _>>` always yields a reëntrancy error

```rust
    /// We failed to acquire an entry in a `CollectionView`.
    #[error("trying to access a collection view while some entries are still being accessed")]
    CannotAcquireCollectionEntry,
```

MathieuDutSik:
## Motivation

The reading of multiple keys in `CollectionView` does not work. This is unfortunate since the underlying
container of `CollectionView` is `async_lock::RwLock` which allows multiple read or one write. We `CollectionView`
should have the same semantics.

Fixes #4588 

## Proposal

The following steps are made:
* The `ReadGuardedView` is now an enum. The `Loaded` entry is when it is present in updates. The `NotLoaded` when accessed from storage.
* The `NotLoaded` contains some lock so as to avoid a write to be created and lead to incoherent state.
* As a consequence, when doing the `try_load_entry` of a view on storage, the updates is not changed. But that is fine, since it is anyway in the LRU cache.
* A `try_load_entries` has been introduced since we can now read multiple keys at once.
* The code of `fn entries` has been simplifed to use those functions.
* The `try_load_entries` no longer take values since only references are needed.
* The `Clone` requirement have been alleviated in `CollectionView` / `ReentrantCollectionView`.

## Test Plan

The CI.
The unit tests in the code should be sufficient for the functionality.

## Release Plan

It addresses an issue that showed up in the GameOfLife. So, moving it to TestNet seems like.a good idea.

## Links

None.

### DIFF
--- a/linera-views/src/views/collection_view.rs
+++ b/linera-views/src/views/collection_view.rs
@@ -356,7 +356,7 @@ impl<W: View> ByteCollectionView<W::Context, W> {
                     Update::Removed => {
                         results.push(None);
                     }
-                    _ => {
+                    Update::Set(_) => {
                         let updates = self.updates.read().await;
                         results.push(Some(ReadGuardedView::Loaded {
                             updates,
@@ -1640,14 +1640,11 @@ mod graphql {
             };
 
             let values = self.try_load_entries(&keys).await?;
-            values
+            Ok(values
                 .into_iter()
                 .zip(keys)
-                .map(|(value, key)| match value {
-                    None => Err(missing_key_error(&key)),
-                    Some(value) => Ok(Entry { value, key }),
-                })
-                .collect()
+                .filter_map(|(value, key)| value.map(|value| Entry { value, key }))
+                .collect())
         }
     }
 
@@ -1703,14 +1700,11 @@ mod graphql {
             };
 
             let values = self.try_load_entries(&keys).await?;
-            values
+            Ok(values
                 .into_iter()
                 .zip(keys)
-                .map(|(value, key)| match value {
-                    None => Err(missing_key_error(&key)),
-                    Some(value) => Ok(Entry { value, key }),
-                })
-                .collect()
+                .filter_map(|(value, key)| value.map(|value| Entry { value, key }))
+                .collect())
         }
     }
 }
--- a/linera-views/src/views/collection_view.rs
+++ b/linera-views/src/views/collection_view.rs
@@ -80,9 +80,7 @@ impl<W> std::ops::Deref for ReadGuardedView<'_, W> {
                 };
                 view
             }
-            ReadGuardedView::NotLoaded { _updates, view } => {
-                view
-            }
+            ReadGuardedView::NotLoaded { _updates, view } => view,
         }
     }
 }
@@ -289,22 +287,15 @@ impl<W: View> ByteCollectionView<W::Context, W> {
         &self,
         short_key: &[u8],
     ) -> Result<Option<ReadGuardedView<W>>, ViewError> {
-        let updates = self
-            .updates
-            .read()
-            .await;
+        let updates = self.updates.read().await;
         match updates.get(short_key) {
-            Some(update) => {
-                match update {
-                    Update::Removed => Ok(None),
-                    _ => {
-                        Ok(Some(ReadGuardedView::Loaded {
-                            updates,
-                            short_key: short_key.to_vec(),
-                        }))
-                    }
-                }
-            }
+            Some(update) => match update {
+                Update::Removed => Ok(None),
+                _ => Ok(Some(ReadGuardedView::Loaded {
+                    updates,
+                    short_key: short_key.to_vec(),
+                })),
+            },
             None => {
                 let key_index = self
                     .context
@@ -361,20 +352,18 @@ impl<W: View> ByteCollectionView<W::Context, W> {
 
         for (position, short_key) in short_keys.into_iter().enumerate() {
             match updates.get(&short_key) {
-                Some(update) => {
-                    match update {
-                        Update::Removed => {
-                            results.push(None);
-                        }
-                        _ => {
-                            let updates = self.updates.read().await;
-                            results.push(Some(ReadGuardedView::Loaded {
-                                updates,
-                                short_key: short_key.clone(),
-                            }));
-                        }
+                Some(update) => match update {
+                    Update::Removed => {
+                        results.push(None);
                     }
-                }
+                    _ => {
+                        let updates = self.updates.read().await;
+                        results.push(Some(ReadGuardedView::Loaded {
+                            updates,
+                            short_key: short_key.clone(),
+                        }));
+                    }
+                },
                 None => {
                     results.push(None); // Placeholder, may be updated later
                     if !self.delete_storage_first {
@@ -403,7 +392,7 @@ impl<W: View> ByteCollectionView<W::Context, W> {
         }
         let values = self
             .context
-	    .store()
+            .store()
             .read_multi_values_bytes(keys_to_load)
             .await?;
 
@@ -1657,12 +1646,10 @@ mod graphql {
             values
                 .into_iter()
                 .zip(keys)
-                .map(|(value, key)|
-                     match value {
-                         None => Err(missing_key_error(&key)),
-                         Some(value) => Ok(Entry { value, key })
-                     }
-                )
+                .map(|(value, key)| match value {
+                    None => Err(missing_key_error(&key)),
+                    Some(value) => Ok(Entry { value, key }),
+                })
                 .collect()
         }
     }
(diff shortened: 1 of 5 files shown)


linera-views/tests/views_tests.rs
```diff
@@ -518,7 +518,7 @@ where
             if config.with_collection {
                 let subview = view.collection2.load_entry_mut("ciao").await?;
                 let subsubview = subview.try_load_entry("!").await?.unwrap();
-                assert!(subview.try_load_entry("!").await.is_err());
+                let _subsubview = subview.try_load_entry("!").await?.unwrap();
```
ma2bd: Just remove `let _subsubview =`

linera-views/src/views/collection_view.rs
```diff
@@ -301,11 +310,9 @@ impl<W: View> ByteCollectionView<W::Context, W> {
...
                     let context = self.context.clone_with_base_key(key);
                     let view = W::load(context).await?;
-                    entry.insert(Update::Set(view));
-                    let guard = RwLockWriteGuard::downgrade(updates);
-                    Ok(Some(ReadGuardedView {
-                        guard,
-                        short_key: short_key.to_vec(),
+                    Ok(Some(ReadGuardedView::NotLoaded {
+                        _updates: updates,
+                        view,
```
ma2bd, line 315: very nice

linera-views/src/views/collection_view.rs
```diff
@@ -269,25 +287,16 @@ impl<W: View> ByteCollectionView<W::Context, W> {
         &self,
         short_key: &[u8],
     ) -> Result<Option<ReadGuardedView<W>>, ViewError> {
-        let mut updates = self
-            .updates
-            .try_write()
-            .ok_or(ViewError::CannotAcquireCollectionEntry)?;
```
ma2bd, line 275: It's great to finally remove this! We can remove the comment "May fail if one subview is already being visited."

linera-views/src/views/collection_view.rs
```diff
@@ -1477,16 +1641,15 @@ mod graphql {
...
-            for key in keys {
-                let value = self
-                    .try_load_entry(&key)
-                    .await?
-                    .ok_or_else(|| missing_key_error(&key))?;
-                values.push(Entry { value, key })
-            }
-
-            Ok(values)
+            let values = self.try_load_entries(&keys).await?;
```
ma2bd, line 1642: nice!

linera-views/src/views/collection_view.rs
ma2bd: Let's make sure to do another (non-backportable) PR after this one to remove `CannotAcquireCollectionEntry`.

ma2bd (commented):
Looks great overall. `try_load_entries` could use another pair of eyes

linera-views/tests/views_tests.rs
MathieuDutSik: Ok.

linera-views/src/views/collection_view.rs
MathieuDutSik: Thanks.

linera-views/src/views/collection_view.rs
```diff
@@ -269,25 +287,16 @@ impl<W: View> ByteCollectionView<W::Context, W> {
...
-                    }
-                    Update::Removed => Ok(None),
-                }
-            }
-            btree_map::Entry::Vacant(entry) => {
+        let updates = self.updates.read().await;
+        match updates.get(short_key) {
+            Some(update) => match update {
+                Update::Removed => Ok(None),
+                _ => Ok(Some(ReadGuardedView::Loaded {
```
deuszx: using `_` in match is an anti-pattern: when you add new variants to an enum compiler won't warn you about mishandling.

linera-views/src/views/collection_view.rs
```diff
@@ -1477,16 +1641,15 @@ mod graphql {
...
-                values.push(Entry { value, key })
-            }
-
-            Ok(values)
+            let values = self.try_load_entries(&keys).await?;
+            values
+                .into_iter()
+                .zip(keys)
+                .map(|(value, key)| match value {
+                    None => Err(missing_key_error(&key)),
```
deuszx: I think it should be OK if one of the requested entries is not found, while others are. We shouldn't fail the whole query.

deuszx: Imagine if we query for an ERC20-like app with 100 addreses – 99 are present but 1 is missing. Does the whole query fail? It'd be bad.

linera-views/src/views/collection_view.rs
MathieuDutSik: Ok.

linera-views/src/views/collection_view.rs
MathieuDutSik: That is a change of API.
It should be done in another PR, I think.
Also, right now, we do not have support for GraphQL for Solidity contracts.

deuszx: It's not about Solidity, Game of Life does the same thing. Portal could query all users' scores (or batch it) instead of doing 12k separate queries – each of a new user.

MathieuDutSik: Yes, and?
The Game Of Life is returning an error if 1 of the keys being queried is missing. Changing that would be a change of API.

deuszx: What "yes and"? The current behavior is wrong. You don't fail the WHOLE query if one of the 100s keys is missing.

deuszx: It's not a breaking change b/c we don't support reading multiple values right now. This is a new behavior.

linera-views/src/views/collection_view.rs
deuszx: There are more occurences of the same pattern – fix all of them.

linera-views/src/views/collection_view.rs
MathieuDutSik: The right way in that case is to return a `null`.
I propose to do this in another PR.

deuszx: Why in another PR? This isn't a breaking change b/c we don't support querying multiple keys at once.

linera-views/src/views/collection_view.rs
ma2bd: I would generally agree. Here the enum is defined in the same file and unlikely to be changed without checking all the occurrences, but at the same time it is also very easy to replace the few entries.

linera-views/src/views/collection_view.rs
MathieuDutSik: It is a breaking change:
* Currently, if one key is missing in the `ReentrantCollectionView`, then the whole query fails.
* After this PR we have the same behavior.

ma2bd: The label `filters` inside `MapInput` suggests that we return the existing entries at the intersection of all filters. The list of keys being a "filter", it is a bug to return an error if some keys don't exist.

ma2bd: > if one key is missing in the ReentrantCollectionView, then the whole query fails.

Yes but that's a bug (when we use `MapInput`)

MathieuDutSik: The structure of the queries is somewhat odd:
* The input to the `fn entries` is an `Option<MapInput>`.
* The `MapInput` encapsulates an `Option<MapFilters>`.
* The `MapFilters` encapsulates a `Vec<K>`.

So, you have 3 levels of `Option` but that does not matter as it gets collapsed for `MapView`, `ReentrantCollectionView`, `CollectionView` as one single `Option`. All 3 handle the `entries` in essentially the same way.

It is a little unclear to me why this special structure is needed. There are few calls to the `entries` function in the code, the examples and related code using Linera. And those calls ask for the whole of `entries`. In fact all the calls to entries ask for all the keys (so `None`, wherever that is put in the hierarchy).

There is no notion of filter in `GraphQL` as a search on the reference https://spec.graphql.org/October2021/ does not show up any matching for `filter`. So, whatever meaning there is to filter is one that we decide by ourselves.

The `SimpleObject` macro when applied to a struct, will provide implementations to a `BTreeMap`, which in case of a missing key will return `null`. But of course, we are not forced to follow this convention, and if we decide that a `filters` changes thing, then why not.

ma2bd: We discussed this internally on slack just one or two days ago.

PR #4616, opened 2025-09-20
Add counter to collection and map views

ma2bd:
## Motivation

Add some missing APIs + improve GraphQL code for collections

## Proposal

* support count() for collections
* make SetView similar to MapView (**GraphQL API breaking change**)
* (nit) remove unnecessary Clone bounds

## Test Plan

CI + tested manually

## Release Plan

These changes should be backported to the latest `testnet` branch, then be released in a new SDK.

(Note that the SDK will be breaking certain queries but that's ok)

### DIFF
--- a/linera-views/src/views/set_view.rs
+++ b/linera-views/src/views/set_view.rs
@@ -967,12 +967,10 @@ mod graphql {
     {
         async fn elements(&self, count: Option<usize>) -> Result<Vec<I>, async_graphql::Error> {
             let mut indices = self.indices().await?;
-            Ok(if let Some(count) = count {
+            if let Some(count) = count {
                 indices.truncate(count);
-                indices
-            } else {
-                indices
-            })
+            }
+            Ok(indices)
         }
 
         #[graphql(derived(name = "count"))]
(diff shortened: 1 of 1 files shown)


ma2bd:
The code for `SetView` is now similar to `MapView`. I will test on the GOL app some more before landing.

linera-views/src/views/set_view.rs
```diff
@@ -939,70 +939,78 @@ pub type HashedCustomSetView<C, I> =
...
+        I: Send + Sync + Serialize + DeserializeOwned + async_graphql::OutputType,
+    {
+        async fn elements(&self, count: Option<usize>) -> Result<Vec<I>, async_graphql::Error> {
+            let mut indices = self.indices().await?;
+            Ok(if let Some(count) = count {
+                indices.truncate(count);
+                indices
+            } else {
+                indices
+            })
```
afck: (Maybe more concise; also below.)

```suggestion
            let mut indices = self.indices().await?;
            if let Some(count) = count {
                indices.truncate(count);
            }
            Ok(indices)
```

afck (approved):

linera-views/src/views/set_view.rs
ma2bd: definitely

PR #4617, opened 2025-09-21
reduce visibility of chain actor internals

ma2bd:
## Motivation

code cleanup

## Proposal

reduce visibility of chain actor internals

## Test Plan

CI

## Release Plan

These changes should be backported to the latest `testnet` branch


MathieuDutSik (approved):
That seems straightforward enough. Surprising that it was not caught by previous cleaning PRs.

PR #4621, opened 2025-09-21
Implement `BLOCK_CACHE_SIZE` / `EXECUTION_STATE_CACHE_SIZE` as input parameters

MathieuDutSik:
## Motivation

Those two variables are constants in the program while varying values could help.

## Proposal

Add it as parameter inputs.
A creation function has been introduced, which leads to a simplification of the code.

## Test Plan

The CI.

## Release Plan

The goal is to backport it to testnet conway. But let's first merge in in the main.

## Links

None.

### DIFF
--- a/linera-service/src/server.rs
+++ b/linera-service/src/server.rs
@@ -68,6 +68,8 @@ struct ServerContext {
     shard: Option<usize>,
     grace_period: Duration,
     chain_worker_ttl: Duration,
+    block_cache_size: usize,
+    execution_state_cache_size: usize,
 }
 
 impl ServerContext {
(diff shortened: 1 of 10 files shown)


linera-service/src/server.rs
```diff
@@ -270,6 +274,18 @@ struct ServerOptions {
     /// The number of Tokio blocking threads to use.
     #[arg(long, env = "LINERA_SERVER_TOKIO_BLOCKING_THREADS")]
     tokio_blocking_threads: Option<usize>,
+
+    /// Size of the block cache (default: 5000)
+    #[arg(long, env = "LINERA_BLOCK_CACHE_SIZE", default_value = "5000")]
```
ma2bd, line 279: Thanks for the env variable!

ma2bd (approved):
LGTM. Just need to fix the lint, CLI, and compilation on web

PR #4623, opened 2025-09-22
remove one chain state view + nits

ma2bd:
## Motivation

Shared state views should only be used for GraphQL / service queries.

## Proposal

Introduce a proper enum and return value (we were halfway there)

## Test Plan

CI

## Release Plan

    - be released in a new SDK,

### DIFF
--- a/linera-core/src/worker.rs
+++ b/linera-core/src/worker.rs
@@ -857,7 +857,7 @@ where
 
         #[cfg(with_metrics)]
         {
-            if !matches!(_outcome, BlockOutcome::Skipped) {
+            if matches!(_outcome, BlockOutcome::Processed) {
                 let (
                     certificate_log_str,
                     round_type,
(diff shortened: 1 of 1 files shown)


linera-core/src/worker.rs
```diff
@@ -837,37 +839,25 @@ where
...
 
-        let result = self
+        let (info, actions, _outcome) = self
             .process_confirmed_block(certificate, notify_when_messages_are_delivered)
             .await?;
 
         #[cfg(with_metrics)]
         {
-            if let Some(metrics_data) = metrics_data {
+            if !matches!(_outcome, BlockOutcome::Skipped) {
```
afck: But we probably only want to record the metrics if it was processed, so that each block is recorded only once?

A block can get preprocessed multiple times, so that might be misleading.

ma2bd: yes indeed

Issue #4624, opened 2025-09-22
Improve account → chains map performance.


afck:
https://github.com/linera-io/linera-protocol/pull/4301 is a temporary solution for https://github.com/linera-io/linera-protocol/issues/4154, but is now deployed on Testnet Conway. Since we store the whole map in a single file, performance is degrading as the file grows.

We should:
* [ ] ~Use `papaya` instead of an `Arc<Mutex<_>>`, so that adding and looking up entries doesn't require a lock.~
* [ ] ~Write the file atomically, with `NamedTempFile::persist`, so that stopping the process while writing doesn't result in an incomplete file.~
* [ ] ~Return a response to clients without waiting for the write to complete.~
* [ ] ~Make sure only one write is in progress at a time, and avoid redundant writes.~
* [x] Store only chain IDs, not whole chain descriptions.
* [ ] ~Shard the file, e.g. by the first one or two bytes of the public key. That way, we don't have to write everything every time.~
* [x] Extend the tests, to make sure chain claims are idempotent even after a faucet restart.
* [x] **Use SQLite instead of a JSON file.**

PR #4625, opened 2025-09-22
Improve faucet performance; store map in the background.

afck:
## Motivation

Requesting a new chain from the faucet now takes 3 seconds because the JSON file with the "account owner → chain description" map is getting very large.

## Proposal

Send the response _before_ storing the file; overwrite the file atomically. Use `papaya` so that we don't need to lock anything to look up map entries.

## Test Plan

Tests were updated.

## Release Plan

    - used to restart the faucet. Before that, we should manually backup the faucet storage file and make sure that it is complete, i.e. ends with the expected closing `}`.

## Links

- Partly addresses https://github.com/linera-io/linera-protocol/issues/4624.

### DIFF
--- a/linera-faucet/server/src/lib.rs
+++ b/linera-faucet/server/src/lib.rs
@@ -128,6 +128,8 @@ struct FaucetStorage {
     owner_to_chain: Arc<papaya::HashMap<AccountOwner, ChainDescription>>,
     sender: Arc<watch::Sender<()>>,
     _save_join_handle: Arc<Mutex<Option<JoinHandle<()>>>>,
+    storage_path: PathBuf,
+    receiver: Arc<Mutex<watch::Receiver<()>>>,
 }
 
 impl FaucetStorage {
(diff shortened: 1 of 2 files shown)


linera-faucet/server/src/lib.rs
```diff
@@ -113,58 +116,126 @@ struct BatchProcessor<C: ClientContext> {
...
     client: ChainClient<C::Environment>,
-    faucet_storage: Arc<Mutex<FaucetStorage>>,
+    faucet_storage: Arc<FaucetStorage>,
     pending_requests: Arc<Mutex<VecDeque<PendingRequest>>>,
     request_notifier: Arc<Notify>,
 }
 
 /// Persistent mapping of account owners to chain descriptions
-#[derive(Debug, Clone, Serialize, Deserialize, Default)]
+#[derive(Debug, Clone)]
```
deuszx, line 125: Does the whole thing need to be clone-able? I'm worried about a case where you do create to copies of it and they both try to update the file.

linera-faucet/server/src/lib.rs
```diff
@@ -113,58 +116,126 @@ struct BatchProcessor<C: ClientContext> {
...
 struct FaucetStorage {
     /// Maps account owners to their corresponding chain descriptions
-    owner_to_chain: HashMap<AccountOwner, ChainDescription>,
+    owner_to_chain: Arc<papaya::HashMap<AccountOwner, ChainDescription>>,
+    sender: Arc<watch::Sender<()>>,
+    _save_join_handle: Arc<Mutex<Option<JoinHandle<()>>>>,
```
ndr-ds, line 130: Why is the underscore needed? 🤔
```suggestion
    save_join_handle: Arc<Mutex<Option<JoinHandle<()>>>>,
```

afck:
Closing in favor of https://github.com/linera-io/linera-protocol/pull/4631.

Issue #4618, opened 2025-09-21
chain actor TTL on clients should distinguish tracked chains from others


ma2bd:
Client are different from validator nodes:
* Generally, they have a few tracked chains (including owned ones) and messages come from random sender chains.
* However, clients are **not** sharded: they will replay cross-chain messages from every sender chains locally.
* Loading chains from storage on clients is cheap.

As a result, chain actor TTL should be very high for tracked chains and very low for sender chains.

PR #4630, opened 2025-09-23
Separate chain worker TTL for sender chains.

deuszx:
## Motivation

We don't need to keep chain workers for sender chains in memory for as long as the _tracked_ chains. Usually they are needed only for a single query.

## Proposal

Add a separate config to control this.

New config parameter , for the client, is `sender-chain-worker-ttl-ms`. Defaults to 1 second.

I also added a way to set it via an env var: `LINERA_SENDER_CHAIN_WORKER_TTL_MS`. 
While I was at it, I added env var control to the "other" (tracked chains ttl): `LINERA_CHAIN_WORKER_TTL_MS`.
## Test Plan

CI

## Release Plan
    - be released in a new SDK,

## Links
Closes https://github.com/linera-io/linera-protocol/issues/4618

### DIFF
--- a/linera-core/src/chain_worker/config.rs
+++ b/linera-core/src/chain_worker/config.rs
@@ -24,7 +24,7 @@ pub struct ChainWorkerConfig {
     pub grace_period: Duration,
     /// Idle chain workers free their memory after that duration without requests.
     pub ttl: Duration,
-    /// Ttl for sender chains.
+    /// TTL for sender chains.
     // We don't want them to keep in memory forever since usually they're short-lived.
     pub sender_chain_ttl: Duration,
 }


linera-core/src/chain_worker/config.rs
```diff
@@ -24,6 +24,9 @@ pub struct ChainWorkerConfig {
     pub grace_period: Duration,
     /// Idle chain workers free their memory after that duration without requests.
     pub ttl: Duration,
+    /// Ttl for sender chains.
```
afck:
```suggestion
    /// TTL for sender chains.
```

linera-web/src/lib.rs
```diff
@@ -86,6 +86,7 @@ pub const OPTIONS: ClientContextOptions = ClientContextOptions {
     long_lived_services: false,
     blob_download_timeout: linera_base::time::Duration::from_millis(1000),
     chain_worker_ttl: Duration::from_secs(30),
+    sender_chain_worker_ttl: Duration::from_secs(1),
```
ma2bd: How about 100ms?

deuszx: We don't want to go too low so that we don't unload the chain actor before handling the query. Also, 1s shouldn't be a problem – how many different chains can be loaded in 1s? If a block has 50 incoming messages at most, it's 50 different chain actors. We have latency of ~500ms to that's 100 different chain actors.

deuszx: Ah, this is web-only and it can't be overwritten :/

deuszx: Decreased to 200ms

linera-core/src/chain_worker/config.rs
deuszx: Done

PR #4631, opened 2025-09-23
Use SQLite for the faucet chain map.

afck:
## Motivation

Requesting a new chain from the faucet now takes 3 seconds because the JSON file with the "account owner → chain description" map is getting very large.

## Proposal

Store the map in an SQLite database, with only chain IDs rather than whole descriptions. `--storage-path` is mandatory now.

On startup, traverse the chain back to the last block whose created chains are already in the database. Then traverse back up and update the database.

(Mostly written with Claude.)

## Test Plan

Tests were updated.

## Release Plan

    - used to restart the faucet with `--storage-path` changed accordingly. Before that, we should manually backup the faucet storage file and make sure that it is complete, i.e. ends with the expected closing `}`.

## Links

- Closes https://github.com/linera-io/linera-protocol/issues/4624.

### DIFF
--- a/linera-faucet/server/src/lib.rs
+++ b/linera-faucet/server/src/lib.rs
@@ -521,7 +521,11 @@ where
             .store_chains_batch(chains_to_store)
             .await
         {
-            tracing::warn!("Failed to save chains to database: {}", e);
+            let error_msg = format!("Failed to save chains to database: {}", e);
+            for request in valid_requests {
+                let _ = request.responder.send(Err(Error::new(error_msg.clone())));
+            }
+            anyhow::bail!(error_msg);
         }
 
         // Respond to requests.


linera-faucet/server/src/lib.rs
```diff
@@ -484,23 +508,25 @@ where
...
+                (request.owner, description.id())
+            })
+            .collect::<Vec<_>>();
 
-            if let Err(e) = storage.save(&self.config.storage_path).await {
-                tracing::warn!("Failed to save faucet storage: {}", e);
-            }
+        if let Err(e) = self
+            .faucet_storage
+            .store_chains_batch(chains_to_store)
```
deuszx, line 521: So we store _after_ we send the description to the caller after all?

linera-faucet/server/src/lib.rs
```diff
@@ -593,22 +619,29 @@ where
...
             .await
-            .context("Failed to load faucet storage")?;
-        let faucet_storage = Arc::new(Mutex::new(faucet_storage));
+            .context("Failed to initialize faucet database")?;
+
+        // Try to migrate from legacy JSON file if it exists
+        let json_path = storage_path.with_extension("json");
+        if let Err(e) = faucet_storage.migrate_from_json(&json_path).await {
+            tracing::warn!("Failed to migrate from JSON storage: {}", e);
+        }
```
deuszx: We should remove the JSON after successful migration so that we don't migrate again.

deuszx: Ah, ok. It's renamed.

linera-faucet/server/src/database.rs
```diff
@@ -0,0 +1,189 @@
...
+    ) -> anyhow::Result<()> {
+        let mut tx = self.pool.begin().await?;
+
+        for (owner, chain_id) in chains {
+            let owner_str = serde_json::to_string(&owner)?;
+            let chain_id_str = chain_id.to_string();
+
+            sqlx::query(
+                r#"
+                INSERT OR REPLACE INTO chains (owner, chain_id)
```
deuszx, line 270: This could overwrite already created chain – we probably shouldn't do that. I think `INSERT OR IGNORE` is more suitable.

ndr-ds:
Just out of curiosity: are we using SQLite here because it's supposed to be simpler? Why not use the existing RocksDB instance that the faucet already runs on?

afck:
(As discussed, we want the faucet file to be easy to backup separately, and easy to inspect. But I don't have a strong opinion on the choice of backend.)

linera-faucet/server/src/lib.rs
afck: Good point, this should be fast now, so we can make sure we store first.

linera-faucet/server/src/database.rs
afck: Right, that should never happen, so we should fail if it exists.

afck: Now that method is also called when syncing; and I think we should handle the case gracefully where the existing chain created multiple new chains for the same account, so I'd rather leave it like this.

linera-faucet/server/src/lib.rs
afck: Done in https://github.com/linera-io/linera-protocol/pull/4631/commits/e4f251abe219f2155809b13423044057ba9ccb34.
But if writing to the DB fails we still only print a warning. Not sure what's the write thing to do in that case, since the chains have already been created at that point.

linera-faucet/server/src/database.rs
```diff
@@ -0,0 +1,283 @@
...
+
+    /// Stores multiple chain mappings in a single transaction
+    pub async fn store_chains_batch(
+        &self,
+        chains: Vec<(AccountOwner, ChainId)>,
+    ) -> anyhow::Result<()> {
+        let mut tx = self.pool.begin().await?;
+
+        for (owner, chain_id) in chains {
+            let owner_str = serde_json::to_string(&owner)?;
```
ma2bd: @afck Why `serde_json`? This adds double quote around the address.

ma2bd: I'm assuming it's just a random AI preference

linera-faucet/server/src/database.rs
ma2bd: TBH I think that's fine. If we call "store", it means we want to store.

linera-faucet/server/src/lib.rs
ma2bd: How about failing the GraphQL query so the users can retry later?

ma2bd: That's what we do with block proposals for instance.

linera-faucet/server/src/tests.rs
```diff
@@ -272,3 +277,415 @@ async fn test_batch_size_reduction_on_limit_errors() {
...
+    cancellation_token.cancel();
+    let _ = processor_task.await;
+    drop(root);
+    drop(faucet_storage);
+
+    // === PHASE 2: Delete the database file (simulate data loss) ===
+    std::fs::remove_file(&storage_path).expect("Should be able to delete database file");
+
+    // === PHASE 3: Create new faucet instance (should sync from blockchain) ===
+    let faucet_storage_2 = Arc::new(FaucetDatabase::new(&storage_path).await.unwrap());
```
ma2bd, line 588: nice!

linera-faucet/server/src/database.rs
ma2bd: fixed in 489b96f2fa9651f7ce79efafece70e39ec9ae3ef

linera-faucet/server/src/lib.rs
ma2bd: fixed in 0d0952e0488dd0333513f563c4894f58e6098e19

ma2bd: Longer discussion: counting on reconstruction to fix the missing DB entry would be tempting, but reconstruction only works for the last blocks.

linera-service/src/cli_wrappers/wallet.rs
```diff
@@ -544,12 +544,19 @@ impl ClientWrapper {
         amount: Amount,
     ) -> Result<FaucetService> {
         let port = port.into().unwrap_or(8080);
+        let temp_dir = tempfile::tempdir()
+            .context("Failed to create temporary directory for faucet storage")?;
+        let storage_path = temp_dir.path().join("faucet_storage.sqlite");
```
deuszx, line 549: Why temp directory again? This will loose data all again.

linera-faucet/server/src/database.rs
deuszx: Even if we decide to overwrite, this is still wrong. It's non-deterministic and depends on the order in which storage was populated. If you traversed chain from bottom to tip – you'll have the last one, if from tip to bottom: the first one.

linera-faucet/server/src/database.rs
```diff
@@ -0,0 +1,283 @@
...
+                );
+                self.store_chains_batch(chains_to_store).await?;
+            }
+        }
+
+        Ok(())
+    }
+
+    /// Extracts OpenChain operations from a certificate and returns (owner, chain_id) pairs.
+    async fn extract_opened_chains(
```
deuszx, line 198: from a purely software engineering perspective: I don't think this should be a method on a database instance. Maybe there should be a module `migration` and the result of this function would be fed into the `db.store_chains_batch`.

linera-faucet/server/src/database.rs
```diff
@@ -0,0 +1,283 @@
...
+            }
+
+            // Add to our height->hash map
+            height_to_hash.insert(current_height, hash);
+
+            // Move to the previous block
+            current_hash = certificate.block().header.previous_block_hash;
+        }
+
+        Ok(height_to_hash)
```
deuszx, line 158: Pretty dangerous to store ALL blocks faucet ever produced in a single in memory map. Good it worked out.

linera-faucet/server/src/lib.rs
```diff
@@ -203,27 +150,36 @@ where
...
+            .get_chain_id(&owner)
+            .await
+            .map_err(|e| Error::new(e.to_string()))?
         {
-            let storage = self.faucet_storage.lock().await;
-            if let Some(existing_description) = storage.get_chain(&owner) {
-                return Ok(existing_description.clone());
-            }
+            // Retrieve the chain description from local storage
+            return get_chain_description_from_storage(&self.storage, existing_chain_id).await;
```
deuszx, line 176: I'd put chain descriptions into sqlite if we're at it. It'd be even faster.

linera-faucet/server/src/lib.rs
```diff
@@ -322,32 +329,47 @@ where
...
+                Err(err) => {
+                    tracing::error!("Database error: {err}");
+                    Err(Error::new(err.to_string()))
+                }
+            };
+            if let Err(response) = request.responder.send(response) {
+                tracing::error!(
+                    "Receiver dropped while sending response {response:?} for request from {}",
+                    request.owner
+                );
```
deuszx, line 364: Why not put this inside the `Ok(Some(existing_chain_id))` ? It'd be clearer that this snippet is executed only for already-existing chains.

linera-faucet/server/src/database.rs
afck: Yes, good catch! I missed that.

linera-faucet/server/src/tests.rs
```diff
@@ -272,3 +277,415 @@ async fn test_batch_size_reduction_on_limit_errors() {
...
+    // Start the batch processor
+    let cancellation_token = CancellationToken::new();
+    let processor_task = {
+        let mut batch_processor = batch_processor;
+        let token = cancellation_token.clone();
+        tokio::spawn(async move { batch_processor.run(token).await })
+    };
+
+    // Set time to allow claims
+    clock.set(Timestamp::from(1000));
```
deuszx, line 346: but `pending_requests` vec is empty

linera-service/src/cli_wrappers/wallet.rs
afck: This function is only used in tests, and in `linera net up`. Since the latter also tears down the network on shutdown, I think it's fine to use a temporary file for the faucet storage, too?

linera-faucet/server/src/tests.rs
```diff
@@ -272,3 +277,415 @@ async fn test_batch_size_reduction_on_limit_errors() {
...
+    assert_eq!(
+        chain_3.id(),
+        chain_3_again.id(),
+        "New chain should be persistent after sync"
+    );
+
+    // Clean up
+    cancellation_token_2.cancel();
+    let _ = processor_task_2.await;
+}
```
deuszx, line 691: nice! what was the prompt that you used to ask CC to write these tests?

linera-faucet/server/src/database.rs
afck: Seems to be ~20 MB for the current faucet chain.

linera-service/src/cli_wrappers/wallet.rs
deuszx: For wallet/signer in CLI we generate the files ourselves and tell the program (via arguments) where they are. Most often we use ones in `/tmp` but that's a delibrate choice and not a hidden feature.

linera-faucet/server/src/lib.rs
afck: I just changed it from description to chain ID, to save space and because it's redundant. Not sure if we should duplicate the descriptions here just to save one DB round trip. They are a lot larger than chain IDs.

linera-faucet/server/src/tests.rs
afck: It was very high-level. Something like: "Write a test that claims a chain, then restarts the faucet, and claims a chain with the same account again, to verify that even after a restart the mapping is preserved."

linera-service/src/cli_wrappers/wallet.rs
afck: So you'd add a mandatory `--faucet-storage-path` argument to `net up`?

linera-faucet/server/src/lib.rs
deuszx: Would be good to know the time diff at least. Right now seems like claims take ~700ms.

linera-service/src/cli_wrappers/wallet.rs
deuszx: Yes. I'd just follow the pattern that we are already using.

linera-faucet/server/src/lib.rs
afck: Is it plausible that a chain description lookup in the local database is a significant fraction of that? The most common case will be that someone requests a _new_ chain anyway, and then there is no lookup?

deuszx: Ah right, I was measuring new chains only.

deuszx: Resolving.

linera-faucet/server/src/lib.rs
afck: I wanted to deduplicate the `send` call and `error!` log: We send a response in both the `Ok(Some(_))` and `Err(_)` cases.

linera-faucet/server/src/database.rs
afck: It also duplicates some code from `lib.rs`, and its logic is questionable: It shouldn't bother going through the transactions at all, and just use the descriptions. I'll work on a follow-up PR…

linera-service/src/cli_wrappers/wallet.rs
afck: I created https://github.com/linera-io/linera-protocol/issues/4635 for this; let's discuss with the team.

PR #4639, opened 2025-09-24
Add a limit for the number of received_log entries in a chain info response

bart-linera:
## Motivation

The huge amount of entries in the `received_log` seems to be making it impossible to sync the old GoL chain on Testnet Conway.

## Proposal

Limit the number of returned entries in a single response. Client should request more entries with an increased offset until the number of returned entries is less than the limit (indicating that there are no more entries).

## Test Plan

CI; Manual testing on the testnet will show whether this fixes the syncing issue

## Release Plan

    - be released in a new SDK,

## Links

### DIFF
--- a/linera-core/src/chain_worker/state.rs
+++ b/linera-core/src/chain_worker/state.rs
@@ -48,6 +48,10 @@ use crate::{
     worker::{NetworkActions, Notification, Reason, WorkerError},
 };
 
+/// The maximum number of entries in a received_log included in a ChainInfo response.
+// TODO(#4638): Revisit the number.
+pub const CHAIN_INFO_MAX_RECEIVED_LOG_ENTRIES: usize = 1000;
+
 /// The state of the chain worker.
 pub(crate) struct ChainWorkerState<StorageClient>
 where
@@ -1411,7 +1415,9 @@ where
         info.requested_sent_certificate_hashes = hashes;
         if let Some(start) = query.request_received_log_excluding_first_n {
             let start = usize::try_from(start).map_err(|_| ArithmeticError::Overflow)?;
-            info.requested_received_log = chain.received_log.read(start..).await?;
+            let end = (start.saturating_add(CHAIN_INFO_MAX_RECEIVED_LOG_ENTRIES))
+                .min(chain.received_log.count());
+            info.requested_received_log = chain.received_log.read(start..end).await?;
         }
         if query.request_manager_values {
             info.manager.add_values(&chain.manager);
--- a/linera-core/src/chain_worker/state.rs
+++ b/linera-core/src/chain_worker/state.rs
@@ -50,7 +50,7 @@ use crate::{
 
 /// The maximum number of entries in a `received_log` included in a `ChainInfo` response.
 // TODO(#4638): Revisit the number.
-pub const CHAIN_INFO_MAX_RECEIVED_LOG_ENTRIES: usize = 1000;
+pub const CHAIN_INFO_MAX_RECEIVED_LOG_ENTRIES: usize = 50_000;
 
 /// The state of the chain worker.
 pub(crate) struct ChainWorkerState<StorageClient>
--- a/linera-core/src/chain_worker/state.rs
+++ b/linera-core/src/chain_worker/state.rs
@@ -48,7 +48,7 @@ use crate::{
     worker::{NetworkActions, Notification, Reason, WorkerError},
 };
 
-/// The maximum number of entries in a received_log included in a ChainInfo response.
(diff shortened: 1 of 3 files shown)


linera-core/src/chain_worker/state.rs
```diff
@@ -48,6 +48,10 @@ use crate::{
     worker::{NetworkActions, Notification, Reason, WorkerError},
 };
 
+/// The maximum number of entries in a received_log included in a ChainInfo response.
```
afck:
```suggestion
/// The maximum number of entries in a `received_log` included in a `ChainInfo` response.
```

(Or "chain info response"?)

linera-core/src/chain_worker/state.rs
```diff
@@ -1411,7 +1415,9 @@ where
         info.requested_sent_certificate_hashes = hashes;
         if let Some(start) = query.request_received_log_excluding_first_n {
             let start = usize::try_from(start).map_err(|_| ArithmeticError::Overflow)?;
-            info.requested_received_log = chain.received_log.read(start..).await?;
+            let end = (start.saturating_add(CHAIN_INFO_MAX_RECEIVED_LOG_ENTRIES))
+                .min(chain.received_log.count());
```
afck:
```suggestion
                .min(chain.received_log.count().saturating_sub(start));
```

bart-linera: Should we be doing this subtraction? It's not the length after all, it's the end index.

afck: Ah, sorry, you're right! :facepalm:

afck (approved):

linera-core/src/chain_worker/state.rs
```diff
@@ -48,6 +48,10 @@ use crate::{
     worker::{NetworkActions, Notification, Reason, WorkerError},
 };
 
+/// The maximum number of entries in a `received_log` included in a `ChainInfo` response.
+// TODO(#4638): Revisit the number.
+pub const CHAIN_INFO_MAX_RECEIVED_LOG_ENTRIES: usize = 1000;
```
deuszx: Can you make it configurable? If you don't want to push the arg through all the layers you could add reading from an env var with fallback to 1000 as the default

bart-linera: Yes, with one caveat: then I can't depend on comparing the number of returned entries to the maximum in the client, because the client's value might be different from the validator's, and so I'll have to keep downloading until zero entries are returned. It's not a huge problem, but it will require at least two round trips even for the smallest chains.

afck: Ah, right, I forgot about that! Not sure what's best then.

bart-linera: It's supposed to be a temporary solution, anyway, so I'd leave it as is for now and just implement a better one when revisiting the issue for future networks.

afck (approved):

linera-core/src/chain_worker/state.rs
deuszx: Just for the record:

This is a validator-side thing. We will be hotfixing ALL validators (including external ones, at some point). Re-deploying the network is dangerous process (possibly involving downtimes) so we don't want to do it unless strictly necessary. 

You did the math and this number should be OK _memory wise_ but what if 1000 is too much to request at one time anyway? We don't have a way to change and retry w/o redeployment.

Hence, adding an env var is required IMO. Especially it's as simple as `env::var("LINERA_xxx").unwrap_or_else(1000)`

linera-core/src/chain_worker/state.rs
```diff
@@ -48,6 +48,10 @@ use crate::{
     worker::{NetworkActions, Notification, Reason, WorkerError},
 };
 
+/// The maximum number of entries in a `received_log` included in a `ChainInfo` response.
+// TODO(#4638): Revisit the number.
+pub const CHAIN_INFO_MAX_RECEIVED_LOG_ENTRIES: usize = 50_000;
```
afck: I thought we wanted to test it with a lower value, more like 1000 or 2000, to be sure that this definitely can't be the problem anymore?

Once we've established that, we can in a later update increase the number again to improve performance, but then we have time to thoroughly test that.

bart-linera: I'm also a bit afraid of making the sync process too slow if we set this too low, and since the sync process worked when there were more than 80000 certificates to sync (I was printing the numbers while syncing back when it still worked and only used lots of memory), I'm reasonably certain that this value will be okay, while not requiring hundreds of round trips for syncing the chain.

PR #4640, opened 2025-09-24
Faucet: some deduplication and added sanity check.

afck:
## Motivation

https://github.com/linera-io/linera-protocol/pull/4631 made the faucet use SQLite for the chains map, but it introduced some duplicated logic.

## Proposal

Deduplicate getting chain descriptions from a block; extract some utility functions; add a sanity check that created chains' owners are as expected; fix a few typos.

## Test Plan

CI

## Release Plan

- _Maybe_ backport to `testnet_conway` to keep the diff small.

## Links

- Original PR: https://github.com/linera-io/linera-protocol/pull/4631

### DIFF
--- a/linera-faucet/server/src/lib.rs
+++ b/linera-faucet/server/src/lib.rs
@@ -483,12 +483,10 @@ where
         }
 
         // Store results.
-        let chains_to_store = requests
+        let chains_to_store = chain_descriptions
             .iter()
-            .zip(&chain_descriptions)
-            .map(|(request, (_owner, description))| (request.owner, description.id()))
+            .map(|(owner, description)| (*owner, description.id()))
             .collect();
-
         if let Err(e) = self
             .faucet_storage
             .store_chains_batch(chains_to_store)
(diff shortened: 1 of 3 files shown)


linera-faucet/server/src/lib.rs
```diff
@@ -750,3 +734,35 @@ where
         schema.execute(request.into_inner()).await.into()
     }
 }
+
+/// Returns all `(AccountOwner, ChainDescription)` pairs for single-owner chains created in this
+/// block.
+fn extract_opened_single_owner_chains(
+    certificate: &ConfirmedBlockCertificate,
+) -> anyhow::Result<Vec<(AccountOwner, ChainDescription)>> {
```
deuszx, line 742: Or make this function more generic, return an iterator over `(AccountOwner, ChainDescription)` and filter out on the call-site all chains that have more than 1 owner?

afck: But then we'd have to do the same kind of filtering in all three call sites.

deuszx (approved):

PR #4641, opened 2025-09-24
Track the admin chain by default.

afck:
## Motivation

It's useful to see the admin chain in the wallet, even for non-admin users.

## Proposal

Add the admin chain on `wallet init`.

Do not make chains the default unless we own them.

## Test Plan

CI
I also ran the README tests locally.

## Release Plan

- These changes _could_ be backported to the latest `testnet` branch, then
    - be released in a new SDK.

## Links

### DIFF
--- a/linera-client/src/wallet.rs
+++ b/linera-client/src/wallet.rs
@@ -45,7 +45,7 @@ impl Wallet {
     }
 
     pub fn insert(&mut self, chain: UserChain) {
-        if self.default.is_none() {
+        if self.default.is_none() && chain.owner.is_some() {
             self.default = Some(chain.chain_id);
         }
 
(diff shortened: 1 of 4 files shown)


MathieuDutSik (approved):
I can confirm that running the test
```bash
cargo test test_script_in_readme::fungible
```

there are now 3 chains being shown, while previously, there was just 2.

linera-client/src/wallet.rs
```diff
@@ -45,7 +45,7 @@ impl Wallet {
     }
 
     pub fn insert(&mut self, chain: UserChain) {
-        if self.default.is_none() {
+        if self.default.is_none() && chain.owner.is_some() {
```
deuszx, line 48: nice catch

PR #4541, opened 2025-09-11
Track admin chain; improve `wallet show`.

afck:
## Motivation

The admin chain is not in the wallet by default, and even if it is, it's impossible to tell from `wallet show` which one it is.

## Proposal

Add the admin chain to the wallet by default.

Improve the `wallet show` output:

```
WALLET CHAINS (4 total)

Admin Chain [DEFAULT]
  Chain ID:     8fd4233c5d03554f87d47a711cf70619727ca3d148353446cab81fb56922c9b7
  Owner:        No owner key
  Timestamp:    2025-08-22 18:41:31.339336
  Blocks:       0

Root Chain 1
  Chain ID:     01fc3cf7935540bb48c6e4436aba62791a4f4b10635af6bdfed6e9bb2b499676
  Owner:        No owner key
  Timestamp:    2025-09-09 16:27:58.875446
  Blocks:       1520
  Latest Block: fe3e32a3573b975f0a4560687e04f374063a357d4a79f562e1b62599ce64a292

Root Chain 2
  Chain ID:     5fdfea6db921421723b2ed1e4f4f0f8c53559fd8b90120b464fc4232bfa34379
  Owner:        No owner key
  Timestamp:    2025-09-09 16:46:32.738908
  Blocks:       0

Child Chain
  Chain ID:     da164b2deaff1577c0ccd854762d4e59c98dc611f89be1dca8bd2da4698aaa6b
  Owner:        0xd18e5ee3ddfabd091cc5dfe40f914c28a407df170339b2503cf1e91fc22e2f5f
  Timestamp:    2025-09-11 09:40:01.297139
  Blocks:       0
```

## Test Plan

Discuss the new layout!

## Release Plan

- These changes should be backported to the latest `main` branch, and
- released in a new SDK.

## Links

- [reviewer checklist](https://github.com/linera-io/linera-protocol/blob/main/CONTRIBUTING.md#reviewer-checklist)

PR #4642, opened 2025-09-24
Improve `wallet show`.

afck:
## Motivation

It's hard to tell from `wallet show` which chain is the admin chain. The table layout uses a lot of horizontal space.

## Proposal

Improve the `wallet show` output:

```
WALLET (4 chains in total)

Child Chain [DEFAULT]
  Chain ID:     a08244c18c828b7b89289f7164428e0b861fb462042416f6c2cfb1f4b53c8621
  Owner:        0x86843df3f162b0a820724d37612ecde13c9a022d0cc96b765be127d34cec2d19
  Timestamp:    2025-09-24 12:54:53.051239
  Blocks:       0
  Epoch:        0

Admin Chain
  Chain ID:     82fc0ff002b793c66db0c57e657c05a30a98cd49d335f8125d0c3bdc4997ffa1
  Owner:        No owner key
  Timestamp:    2025-09-24 12:53:29.841442
  Blocks:       0
  Epoch:        0

Root Chain 1
  Chain ID:     0b67f224ab915776d170eeb7d497e8d1ed831126f22eaf9138de48e3c47c8639
  Owner:        No owner key
  Timestamp:    2025-09-24 12:55:04.030389
  Blocks:       2
  Epoch:        0
  Latest Block: f67e7ee881b6f9235f1a4cc6f2515d9ce4afccb1bd799f8780987901a581ac03

Child Chain
  Chain ID:     87c4ef0bfc8d6581529dc32c98d05ba99be9e9202f8f788f9c0c2a872c02fdf6
  Owner:        0x9010b58faf7b4908de3069aa7aa042ba4f2eb0de13724059960aaa5be33aa6a6
  Timestamp:    2025-09-24 12:55:04.030389
  Blocks:       0
  Epoch:        0
```

## Test Plan

* Discuss the new layout!
* CI
* I ran the README tests locally.

## Release Plan

- These changes _could_ be backported to `testnet_conway`, and
- released in a new SDK.

## Links

- Original attempt: https://github.com/linera-io/linera-protocol/pull/4541

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2277,17 +2277,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "comfy-table"
-version = "7.1.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4a65ebfec4fb190b6f90e944a817d60499ee0744e582530e2c9900a22e591d9a"
-dependencies = [
- "crossterm",
- "unicode-segmentation",
- "unicode-width 0.2.1",
-]
-
-[[package]]
 name = "concurrent-queue"
 version = "2.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2749,28 +2738,6 @@ name = "crossbeam-utils"
 version = "0.8.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d0a5c400df2834b80a4c3327b3aad3a4c4cd4de0629063962b03235697506a28"
-
-[[package]]
-name = "crossterm"
-version = "0.28.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "829d955a0bb380ef178a640b91779e3987da38c9aea133b20614cfed8cdea9c6"
-dependencies = [
- "bitflags 2.9.1",
- "crossterm_winapi",
- "parking_lot",
- "rustix 0.38.44",
- "winapi",
-]
-
-[[package]]
-name = "crossterm_winapi"
-version = "0.9.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "acdd7c62a3665c7f6830a51635d9ac9b23ed385797f70a83bb8bafe9c572ab2b"
-dependencies = [
- "winapi",
-]
 
 [[package]]
 name = "crowd-funding"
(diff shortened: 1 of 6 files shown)


MathieuDutSik (approved):
The preceding format was more esthetically pleasing. But functionality and parsability must be the priority.

I confirm that the output of
```bash
cargo test test_script_in_readme::fungible
```

is just 2 chains.

But there are some ways that could be be used to improve the presentation while still being parsable:
* "Child Chain": Who is the parent?
* Output is "WALLET CHAINS (2 total)" I propose a rewrite "WALLET (2 chains in total)". I presume we have never 1 chain and so the "s" is fine.
* I must say that I do not understand the "owner" entry. Apparently, a wallet can I have more than 1 owner, and then the corresponding owner is shown. Or if there are none then "No owner key".
* The "Latest Block" entry is sometimes present, sometimes not. It seems explained by "Blocks" being 0 or not. If no block, then maybe we could write "Latest Block: (none since no block)".
* Any other info that is compact. It is no longer a table, so use that possibility.

None of this is a blocker.

afck:
> "Child Chain": Who is the parent?

Let's discuss and add that in a separate PR if desired. It will require loading the chain description.

> Output is "WALLET CHAINS (2 total)" I propose a rewrite "WALLET (2 chains in total)".

Done in https://github.com/linera-io/linera-protocol/pull/4642/commits/1c7084252f4c8e15f6c7a31820cc1e25ef097ab0.

> I presume we have never 1 chain and so the "s" is fine.

After `wallet init` it's only one chain (the admin chain). I fixed it in https://github.com/linera-io/linera-protocol/pull/4642/commits/1c7084252f4c8e15f6c7a31820cc1e25ef097ab0.

> I must say that I do not understand the "owner" entry. Apparently, a wallet can I have more than 1 owner, and then the corresponding owner is shown. Or if there are none then "No owner key".

This is the account that we use to sign block proposals for this chain.

> The "Latest Block" entry is sometimes present, sometimes not. It seems explained by "Blocks" being 0 or not. If no block, then maybe we could write "Latest Block: (none since no block)".

Isn't it self-explanatory if it says that the chain has 0 blocks?

> Any other info that is compact. It is no longer a table, so use that possibility.

Yes, I agree, but (just like in the previous point) I'd prefer to avoid showing redundant information. E.g. instead of always printing `Latest Block`, I'd rather omit `Epoch` if it is the same as the admin chain's.

Anyway, I'd definitely like to improve all of this further! This is just a start.

ma2bd:
I'm not sure I like all of the changes here. In particular, "Child chain" is weird

afck:
Would "regular chain" be a better wording? Or "non-root chain"?

What are the other changes you don't like?

PR #4643, opened 2025-09-24
[Backport] Add a limit for the number of received_log entries in a chain info response

bart-linera:
## Motivation

The huge amount of entries in the `received_log` seems to be making it impossible to sync the old GoL chain on Testnet Conway.

## Proposal

Limit the number of returned entries in a single response. Client should request more entries with an increased offset until the number of returned entries is less than the limit (indicating that there are no more entries).

## Test Plan

CI; Manual testing on the testnet will show whether this fixes the syncing issue

## Release Plan

- These changes should be:
    - released in a new SDK,
    - released in a validator hotfix.

## Links

### DIFF
--- a/linera-core/src/chain_worker/config.rs
+++ b/linera-core/src/chain_worker/config.rs
@@ -7,8 +7,10 @@ use std::sync::Arc;
 
 use linera_base::{crypto::ValidatorSecretKey, time::Duration};
 
+use crate::CHAIN_INFO_MAX_RECEIVED_LOG_ENTRIES;
+
 /// Configuration parameters for the [`ChainWorkerState`][`super::state::ChainWorkerState`].
-#[derive(Clone, Default)]
+#[derive(Clone)]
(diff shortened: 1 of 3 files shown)


linera-core/src/chain_worker/config.rs
```diff
@@ -50,3 +52,18 @@ impl ChainWorkerConfig {
...
+    fn default() -> Self {
+        Self {
+            key_pair: None,
+            allow_inactive_chains: false,
+            allow_messages_from_deprecated_epochs: false,
+            long_lived_services: false,
+            grace_period: Default::default(),
+            ttl: Default::default(),
+            sender_chain_ttl: Default::default(),
+            chain_info_max_received_log_entries: CHAIN_INFO_MAX_RECEIVED_LOG_ENTRIES,
```
ma2bd, line 66: Thanks I was thinking about that

PR #4646, opened 2025-09-24
Rename `ensure_is_active` to make it clear that it modifies the chain.

afck:
## Motivation

A function shouldn't be called `ensure_` if it modifies the state.

## Proposal

Rename them.

In a follow-up PR, we should investigate whether we really need to `save()` in all these cases.

## Test Plan

(Only renaming.)

## Release Plan

- These changes could be backported to `testnet_conway`.

## Links

### DIFF
--- a/linera-chain/src/chain.rs
+++ b/linera-chain/src/chain.rs
@@ -491,10 +491,7 @@ where
     }
 
     /// Initializes the chain if it is not active yet.
-    pub async fn initialize_if_inactive(
-        &mut self,
-        local_time: Timestamp,
-    ) -> Result<(), ChainError> {
+    pub async fn initialize_if_needed(&mut self, local_time: Timestamp) -> Result<(), ChainError> {
         // Initialize ourselves.
         if self
             .execution_state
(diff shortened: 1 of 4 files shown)


linera-chain/src/chain.rs
```diff
@@ -490,8 +490,11 @@ where
         self.execution_state.system.is_active()
     }
 
-    /// Invariant for the states of active chains.
-    pub async fn ensure_is_active(&mut self, local_time: Timestamp) -> Result<(), ChainError> {
+    /// Initializes the chain if it is not active yet.
+    pub async fn initialize_if_inactive(
```
deuszx: Maybe just _initialize_ ? and a comment that this is idempotent and if it's already initialized it returns early.

ma2bd: `maybe_initialize` or `initialize_if_needed` comes to mind but the new name also works

afck: Done in https://github.com/linera-io/linera-protocol/pull/4646/commits/5136ec8f7ca195e560c05faeef4abb038398622b.

deuszx: I don't like `maybe_` in method names. `initialize` is idempotent function anyway and the caller doesn't care if it's the first or second (or any other) call to it. It just cares it's initialized.

Anyway, not a blocker.

afck: I used `_if_needed`.

PR #4647, opened 2025-09-24
Use `thread_rng` instead of `from_entropy`.

afck:
## Motivation

[`from_entropy` uses `getrandom` and is slower than `thread_rng`](https://docs.rs/rand/0.8.5/rand/trait.SeedableRng.html#method.from_entropy).

## Proposal

Use `thread_rng` to select validators: This does not need to be high-quality randomness.

## Test Plan

CI

## Release Plan

## Links

- Original PR: #4505


MathieuDutSik (approved):
LGTM

PR #4650, opened 2025-09-24
Counter collection views testnet (Backport of #4616)

ma2bd:
## Motivation

Backport #4616

## Proposal

See #4616

## Test Plan

CI

### DIFF
--- a/linera-views/src/views/set_view.rs
+++ b/linera-views/src/views/set_view.rs
@@ -967,12 +967,10 @@ mod graphql {
     {
         async fn elements(&self, count: Option<usize>) -> Result<Vec<I>, async_graphql::Error> {
             let mut indices = self.indices().await?;
-            Ok(if let Some(count) = count {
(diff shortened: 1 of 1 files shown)


linera-views/src/views/set_view.rs
```diff
@@ -939,70 +939,78 @@ pub type HashedCustomSetView<C, I> =
...
+        I: Send + Sync + Serialize + DeserializeOwned + async_graphql::OutputType,
+    {
+        async fn elements(&self, count: Option<usize>) -> Result<Vec<I>, async_graphql::Error> {
+            let mut indices = self.indices().await?;
+            Ok(if let Some(count) = count {
+                indices.truncate(count);
+                indices
+            } else {
+                indices
+            })
```
afck: (Also below; not sure if this was done on main?)

```suggestion
            if let Some(count) = count {
                indices.truncate(count);
            }
            Ok(indices)
```

PR #4652, opened 2025-09-24
Track the admin chain by default. (#4641)

afck:
Backport of https://github.com/linera-io/linera-protocol/pull/4641.

## Motivation

It's useful to see the admin chain in the wallet, even for non-admin users.

## Proposal

Add the admin chain on `wallet init`.

Do not make chains the default unless we own them.

## Test Plan

CI
I also ran the README tests locally.

## Release Plan

- These changes _could_ be backported to the latest `testnet` branch, then
    - be released in a new SDK.

## Links

- PR on main: https://github.com/linera-io/linera-protocol/pull/4641

### DIFF
--- a/linera-service/src/cli/main.rs
+++ b/linera-service/src/cli/main.rs
@@ -1672,7 +1672,7 @@ impl Runnable for Job {
                 let Some(network_description) = storage.read_network_description().await? else {
                     anyhow::bail!("Missing network description");
                 };
-                let context = ClientContext::new(
+                let mut context = ClientContext::new(
                     storage,
                     options.context_options.clone(),
                     wallet,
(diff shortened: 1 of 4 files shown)


linera-service/src/cli/main.rs
```diff
@@ -1684,6 +1684,7 @@ impl Runnable for Job {
                 chain_client
                     .synchronize_chain_state_from_committee(committee)
                     .await?;
+                context.update_wallet_from_client(&chain_client).await?;
```
ma2bd, line 1687: This feels a little bit like a hack but ok

ma2bd (approved):

linera-service/src/cli/main.rs
afck: What would you do instead?

PR #4653, opened 2025-09-24
Invert validators' weights

deuszx:


linera-core/src/client/mod.rs
```diff
@@ -293,7 +293,7 @@ impl<Env: Environment> Client<Env> {
             .iter()
             .map(|validator| {
                 let validator_state = committee.validators.get(&validator.public_key).unwrap();
-                validator_state.votes
+                100u64 - validator_state.votes
```
afck, line 296: That should probably be `101u64.saturating_sub(validator_state.votes).max(1)`.

deuszx: It was just a temporary change to test the theory.

PR #4657, opened 2025-09-24
Forbid anonymous enum variants in `GraphQLMutationRoot`.

MathieuDutSik:
## Motivation

The `MutationRoot` implementation can be difficult to write down (but we have 4 examples doing that in the `examples` directory). Therefore, we have a trait `GraphQLMutationRoot` and a derive macro that allows implementing the `MutationRoot` more easily.

That macro represents the choices that we think are the right ones for smart contract writers. For example, we do not allow structs by design. It has recently emerged that `field0` is not a favored design. Therefore, it makes full sense that the macro should not allow it.

Note that we do not force programmers to use the `GraphQLMutationRoot`. We also do not force them to use `GraphQL`. However, the `GraphQLMutationRoot` that we provide should represent the choices that we think are the right ones.

## Proposal

Explicitly, forbid that check.

## Test Plan

The CI. 
Added a test for testing that this scenario is not allowed.

## Release Plan

Safe to backport to the TestNet / DevNet.

## Links

None.

### DIFF
--- a/linera-sdk-derive/src/lib.rs
+++ b/linera-sdk-derive/src/lib.rs
@@ -57,26 +57,12 @@ fn generate_mutation_root_code(input: ItemEnum, crate_root: &str) -> TokenStream
                     }
                 });
             }
-            Fields::Unnamed(unnamed) => {
-                let mut fields = vec![];
-                let mut field_names = vec![];
-                for (i, field) in unnamed.unnamed.iter().enumerate() {
-                    let name = concat(&syn::parse_str::<Ident>("field").unwrap(), &i.to_string());
-                    let ty = &field.ty;
-                    fields.push(quote! {#name: #ty});
-                    field_names.push(name);
-                }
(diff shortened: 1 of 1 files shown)


ma2bd:
We don't really need to disallow this. The macro works

MathieuDutSik:
> We don't really need to disallow this. The macro works

That does not make sense to me.

We have two realistic scenarios:
* Supporting as much as possible in the macro.
* Providing support for the features that we recommend and only those.

The support for structs was proposed in a previous PR, but dismissed. Also, the `field0` was an issue in recent works. So, why not prevent the use of anonymous variants?

Note also that the documentation is in dire contradiction with those stated best practices. In the documentation https://linera.dev/developers/backend/abi.html, the Operation is not a struct, not an enum, but a `u64`.

ma2bd:
closing this. We have more important things to discuss

PR #4666, opened 2025-09-25
Add an option to limit how many tasks are joined concurrently.

afck:
## Motivation

Joining an unbounded number of concurrent tasks can cause spikes in memory usage.

## Proposal

Replace all the `[try_]join_all` calls in the client module with `buffer_unordered`, with a configurable limit.

## Test Plan

CI

## Release Plan

    - be released in a new SDK.

## Links


afck:
~That actually doesn't work due to cancellation safety.~

I changed it to avoid cancelling.

ma2bd (approved):

PR #4668, opened 2025-09-25
Limit remote log conway (with CLI option)

ma2bd:
## Motivation

variation on #4643 (feel free to cherry-pick commits)

## Proposal

* Do not parse env variables from linera-core
* Make the value not configurable from clients
* Add an option for validators to experiment
* Set the value to 20000

## Test Plan

CI

### DIFF
--- a/linera-core/src/chain_worker/config.rs
+++ b/linera-core/src/chain_worker/config.rs
@@ -27,6 +27,8 @@ pub struct ChainWorkerConfig {
     /// TTL for sender chains.
     // We don't want them to keep in memory forever since usually they're short-lived.
     pub sender_chain_ttl: Duration,
+    /// Override the default size to truncate receive log entries in chain info responses.
+    pub override_chain_info_max_received_log_entries: Option<usize>,
 }
 
 impl ChainWorkerConfig {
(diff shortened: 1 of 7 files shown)


linera-core/src/chain_worker/config.rs
```diff
@@ -27,6 +27,8 @@ pub struct ChainWorkerConfig {
     /// TTL for sender chains.
     // We don't want them to keep in memory forever since usually they're short-lived.
     pub sender_chain_ttl: Duration,
+    /// Override the default size to truncate receive log entries in chain info responses.
+    pub override_chain_info_max_received_log_entries: Option<usize>,
```
afck, line 31: Why not make it non-optional, and make the default value visible in the CLI help?

linera-core/src/lib.rs
```diff
@@ -29,3 +29,7 @@ pub use crate::join_set_ext::{JoinSetExt, TaskHandle};
 
 pub mod environment;
 pub use environment::Environment;
+
+/// The maximum number of entries in a `received_log` included in a `ChainInfo` response.
+// TODO(#4638): Revisit the number.
+pub const CHAIN_INFO_MAX_RECEIVED_LOG_ENTRIES: usize = 100_000;
```
afck, line 35: If this is the minimum, I would choose a much lower value, unless we are very confident that the problem cannot occur with 100000.

afck (approved):
No blockers from my side.

PR #4672, opened 2025-09-26
Make connection and process_inbox logs less noisy.

afck:
## Motivation

Some `INFO` log messages are rather noisy if there are lots of chains in a wallet.

## Proposal

Make them `DEBUG`-level instead.

## Test Plan

(Only logging changes.)

## Release Plan

    - be released in a new SDK.

## Links

### DIFF
--- a/linera-client/src/chain_listener.rs
+++ b/linera-client/src/chain_listener.rs
@@ -26,7 +26,7 @@ use linera_core::{
 };
 use linera_storage::{Clock as _, Storage as _};
 use tokio_util::sync::CancellationToken;
-use tracing::{debug, instrument, warn, Instrument as _};
+use tracing::{debug, info, instrument, warn, Instrument as _};
 
 use crate::{
     wallet::{UserChain, Wallet},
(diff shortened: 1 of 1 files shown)


linera-core/src/client/mod.rs
```diff
@@ -3975,7 +3975,7 @@ impl<Env: Environment> ChainClient<Env> {
                 if let Err(error) = &result {
                     warn!(?error, "Could not connect to validator {public_key}");
                 } else {
-                    info!("Connected to validator {public_key}");
+                    debug!("Connected to validator {public_key}");
```
bart-linera, line 4014: Isn't this one printed just once per validator? In which case I wouldn't consider it very noisy, and I at least found it reassuring when I could see the node connecting to various validators.

afck: I think it's printed once per validator _per chain ID_.

linera-client/src/chain_listener.rs
```diff
@@ -541,9 +541,9 @@ impl<C: ClientContext> ChainListener<C> {
                 debug!(%chain_id, "Cannot find key for chain");
             }
             Err(error) => warn!(%error, "Failed to process inbox."),
-            Ok((certs, None)) => info!("Done processing inbox. {} blocks created.", certs.len()),
+            Ok((certs, None)) => debug!("Done processing inbox. {} blocks created.", certs.len()),
             Ok((certs, Some(new_timeout))) => {
-                info!(
+                debug!(
                     "{} blocks created. Will try processing the inbox later based \
                      on the given round timeout: {new_timeout:?}",
```
ma2bd: I kinda like those :)

linera-client/src/chain_listener.rs
```diff
@@ -541,9 +541,9 @@ impl<C: ClientContext> ChainListener<C> {
                 debug!(%chain_id, "Cannot find key for chain");
             }
             Err(error) => warn!(%error, "Failed to process inbox."),
-            Ok((certs, None)) => info!("Done processing inbox. {} blocks created.", certs.len()),
+            Ok((certs, None)) => debug!("Done processing inbox. {} blocks created.", certs.len()),
             Ok((certs, Some(new_timeout))) => {
-                info!(
+                debug!(
```
ma2bd: This one I kinda like

afck:
@ma2bd: I reverted those two, and instead added the chain ID to the log line, so it's more clear why it sometimes prints several of them (one per chain).

ma2bd (approved):

PR #4683, opened 2025-09-29
Historical hash POC

ma2bd:
## Motivation

Proof of concept for historical hashing

## Proposal

Introduce a new combinator and a new trait

## Test Plan

CI

## Release Plan

This won't be trivial to release in testnet but maybe still doable.


linera-views/src/views/historical_hash_wrapper.rs
```diff
@@ -0,0 +1,528 @@
...
+#[repr(u8)]
+enum KeyTag {
+    /// Prefix for the indices of the view.
+    Inner = MIN_VIEW_TAG,
+    /// Prefix for the hash.
+    Hash,
+}
+
+impl<C, W> HistoricallyHashableView<C, W> {
+    fn make_hash(&self, batch: &Batch) -> Result<HasherOutput, ViewError> {
```
afck:
```suggestion
    /// Computes a new hash based on the stored hash and the batch.
    fn make_hash(&self, batch: &Batch) -> Result<HasherOutput, ViewError> {
```

linera-views/src/views/historical_hash_wrapper.rs
```diff
@@ -0,0 +1,528 @@
...
+            hash,
+            inner,
+        })
+    }
+
+    async fn load(context: Self::Context) -> Result<Self, ViewError> {
+        let keys = Self::pre_load(&context)?;
+        let values = context.store().read_multi_values_bytes(keys).await?;
+        Self::post_load(context, &values)
+    }
```
afck, line 140: (I'm surprised this isn't the default implementation. Isn't it the same for almost all views?)

linera-views/src/views/historical_hash_wrapper.rs
```diff
@@ -0,0 +1,528 @@
...
+
+    async fn has_pending_changes(&self) -> bool {
+        self.inner.has_pending_changes().await
+    }
+
+    fn flush(&mut self, batch: &mut Batch) -> Result<bool, ViewError> {
+        let mut inner_batch = Batch::new();
+        self.inner.flush(&mut inner_batch)?;
+        let hash = self.make_hash(&inner_batch)?;
+        batch.operations.extend(inner_batch.operations);
```
afck, line 155: (This would feel safer if there were a `Batch::extend`, so it's obvious to the reader that we aren't missing any other fields.)

linera-views/src/views/historical_hash_wrapper.rs
```diff
@@ -0,0 +1,528 @@
...
+            _phantom: PhantomData,
+            stored_hash: self.stored_hash,
+            hash: self.hash,
+            inner: self.inner.clone_unchecked()?,
+        })
+    }
+}
+
+impl<W: ClonableView> HistoricallyHashableView<W::Context, W> {
+    /// Obtain a hash of the history of the changes in the view.
```
afck:
```suggestion
    /// Obtains a hash of the history of the changes in the view.
```

linera-views/src/views/historical_hash_wrapper.rs
```diff
@@ -0,0 +1,528 @@
...
+    pub async fn historical_hash(&mut self) -> Result<HasherOutput, ViewError> {
+        if let Some(hash) = self.hash {
+            return Ok(hash);
+        }
+        let mut batch = Batch::new();
+        if self.inner.has_pending_changes().await {
+            let mut inner = self.inner.clone_unchecked()?;
+            inner.flush(&mut batch)?;
```
afck, line 198: Doesn't this modify `inner`?
If we call `historical_hash` and then `flush`, doesn't that store a different state than just calling `flush`? Maybe this could be checked in `test_historically_hashable_view_hash_changes_with_modifications`.

linera-views/src/views/historical_hash_wrapper.rs
MathieuDutSik: Yes, `almost` all.
Will create a PR to address that.

linera-views/src/views/historical_hash_wrapper.rs
```diff
@@ -0,0 +1,528 @@
...
+        let mut batch = Batch::new();
+        view.flush(&mut batch)?;
+        assert!(batch.is_empty());
+        context.store().write_batch(batch).await?;
+
+        let hash_after = view.historical_hash().await?;
+        assert_eq!(hash_before, hash_after);
+
+        Ok(())
+    }
```
MathieuDutSik, line 541: One test should be added that illustrates the "historical" moniker.
* Set to 3, compute hash. set to 49, compute hash as hash_a.
* Set directly to 49, compute hash as hash_b.

We should have hash_a != hash_b, because this hash depends on the history of the hash and not on its state.

linera-views/src/views/mod.rs
```diff
@@ -112,19 +115,15 @@ pub trait HashableView: View {
...
-    async fn hash_mut(&mut self) -> Result<<Self::Hasher as Hasher>::Output, ViewError>;
-
     /// Computes the hash of the values.
     ///
     /// Implementations do not need to include a type tag. However, the usual precautions
     /// to enforce collision resistance must be applied (e.g. including the length of a
     /// collection of values).
     async fn hash(&self) -> Result<<Self::Hasher as Hasher>::Output, ViewError>;
+
+    /// Same as `hash` but guaranteed to be lock-free.
```
MathieuDutSik: There is no such guarantee.
There is less locking, but that is all that one can say.

linera-views/src/views/mod.rs
```diff
@@ -166,24 +165,22 @@ pub trait RootView: View {
 /// A [`View`] that also supports crypto hash
 #[cfg_attr(not(web), trait_variant::make(Send))]
 pub trait CryptoHashView: HashableView {
-    /// Computing the hash and attributing the type to it.
+    /// Computing the hash and attributing the type to it. May require locking.
     async fn crypto_hash(&self) -> Result<CryptoHash, ViewError>;
 
-    /// Computing the hash and attributing the type to it.
+    /// Same as `crypto_hash` but guaranteed to be lock-free.
```
MathieuDutSik: Same.

linera-views/src/views/mod.rs
```diff
@@ -166,24 +165,22 @@ pub trait RootView: View {
...
-///
-/// Sharing is unsafe because by having two view instances for the same data, they may have invalid
-/// state if both are used for writing.
+/// A view that can be shared (unsafely) by cloning it.
 ///
-/// Sharing the view is guaranteed to not cause data races if only one of the shared view instances
-/// is used for writing at any given point in time.
+/// Note: Calling `flush` on any of the shared views will break the other views. Therefore,
+/// cloning views is only safe if `flush` only ever happens after all the copies but one
+/// have been dropped.
```
MathieuDutSik, line 183: Is it the only failure scenario?

linera-views/src/views/historical_hash_wrapper.rs
ma2bd: Perhaps we should make that a static method actually

linera-views/src/views/historical_hash_wrapper.rs
ma2bd: That's why `inner` is a deep-copy of `self.inner`. Am I misunderstanding the question?

linera-views/src/views/mod.rs
ma2bd: I could write wait-free is that's more clear

linera-views/src/views/mod.rs
ma2bd: What is?

linera-views/src/views/historical_hash_wrapper.rs
ma2bd: was fixed in a separate PR

linera-views/src/views/historical_hash_wrapper.rs
ma2bd: done

linera-views/src/views/historical_hash_wrapper.rs
ma2bd: done

linera-views/src/views/historical_hash_wrapper.rs
afck: Ah, of course, makes sense! :facepalm:

linera-views/src/views/historical_hash_wrapper.rs
afck: (Actually, not done yet. Maybe you haven't pushed yet?)

ma2bd: Yes sorry I lost wifi in the plane, then forgot :)

afck (approved):

linera-views/src/views/mod.rs
MathieuDutSik: "wait-free" is an even stronger statement than "lock-free" which is definitely not satisfied by that code.

ma2bd: let me put a PR

PR #4684, opened 2025-09-29
Restructure the `RemoteNetTestConfig` type.

MathieuDutSik:
## Motivation

We have some errors "ERROR linera: Error is Failed to close chain" in the running of `RemoteNet` tests, either against a local set of validators or the TestNet. Those errors are not critical, but are annoying.

## Proposal

The following has been done:
* The `RemoteNetTestingConfig` takes some argument `faucet_url: Option<String>`. However, that argument is never used. Instead, the faucet URL is set from the environment variable `LINERA_FAUCET_URL`. This is the case for remote-net tests and for the test against an existing testnet.
* Therefore, the `faucet_url` argument can be dropped.
* Instead the argument `close_chain` is added. If false, the chain leaks.

The net effect of this PR is to remove the error from the running of the remote-net tests.
What it does not:
* It does not change the close-chain feature, whether we want to continue it or not.
* It leaves the chains for all other kinds of tests.
* It does not address the permission problem that prevents the closure for some chains in `remote-net`.
* If a chain were to fail closing, no error is reported.

## Test Plan

The CI.

## Release Plan

This PR could be backported to the `testnet_conway` branch.

## Links

None.

### DIFF
--- a/linera-service/tests/linera_net_tests.rs
+++ b/linera-service/tests/linera_net_tests.rs
@@ -41,8 +41,6 @@ use linera_sdk::{
     feature = "storage-service",
 ))]
 use linera_service::cli_wrappers::local_net::{Database, LocalNetConfig};
-#[cfg(feature = "remote-net")]
-use linera_service::cli_wrappers::remote_net::RemoteNetTestingConfig;
 #[cfg(any(
     feature = "dynamodb",
     feature = "scylladb",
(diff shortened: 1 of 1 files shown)


linera-service/src/cli_wrappers/remote_net.rs
```diff
@@ -71,6 +70,7 @@ pub struct RemoteNet {
     testing_prng_seed: Option<u64>,
     next_client_id: usize,
     tmp_dir: Arc<TempDir>,
+    close_chain: bool,
```
afck: (We could make this type `OnClientDrop` instead of `bool`? Not sure.)

afck:
```suggestion
    close_chains: bool,
```

afck (approved):

linera-service/src/cli_wrappers/remote_net.rs
MathieuDutSik: Yes, done.

PR #4687, opened 2025-09-29
Implement `Allocative` for the protocol views.

MathieuDutSik:
## Motivation

Tracking the size of the objects being used is important for considering the memory usage and resolving the OOM.

## Proposal

The `Allocative` trait is implemented for the views, and the code is followed from there.

Several design decisions were made:
* For the Weighted distribution, the entries are dropped.
* In the views, the context is not taken into account of memory usage.
* For the keys of Ed25519, Secp256k1, EvmSecp256k1, the sizes are written explicitly.

## Test Plan

The CI.

Adding in `process_confirmed_block` the following 
```rust
tracing::info!("process_confirmed_block, size_of_unique={}", allocative::size_of_unique(&self.chain));
```

works as expected. How to implement more systematically is open to discussion.

## Release Plan

The goal is to add it to the TestNet Conway.

## Links

None.

### DIFF
--- a/linera-base/src/crypto/mod.rs
+++ b/linera-base/src/crypto/mod.rs
@@ -19,8 +19,7 @@ pub use hash::*;
 use linera_witty::{WitLoad, WitStore, WitType};
 pub use secp256k1::{
     evm::{EvmPublicKey, EvmSecretKey, EvmSignature},
-    Secp256k1PublicKey, Secp256k1SecretKey, Secp256k1Signature, SECP256K1_PUBLIC_KEY_SIZE,
-    SECP256K1_SIGNATURE_SIZE,
+    Secp256k1PublicKey, Secp256k1SecretKey, Secp256k1Signature,
 };
 use serde::{Deserialize, Serialize};
 pub use signer::*;
--- a/linera-base/src/crypto/secp256k1/mod.rs
+++ b/linera-base/src/crypto/secp256k1/mod.rs
@@ -32,10 +32,10 @@ use crate::doc_scalar;
 const SECP256K1_SCHEME_LABEL: &str = "secp256k1";
 
 /// Length of secp256k1 compressed public key.
-pub const SECP256K1_PUBLIC_KEY_SIZE: usize = 33;
+const SECP256K1_PUBLIC_KEY_SIZE: usize = 33;
 
 /// Length of secp256k1 signature.
-pub const SECP256K1_SIGNATURE_SIZE: usize = 64;
+const SECP256K1_SIGNATURE_SIZE: usize = 64;
 
 /// A secp256k1 secret key.
 #[derive(Eq, PartialEq)]
--- a/linera-chain/src/certificate/generic.rs
+++ b/linera-chain/src/certificate/generic.rs
@@ -5,10 +5,7 @@
 use allocative::{Allocative, Key, Visitor};
 use custom_debug_derive::Debug;
 use linera_base::{
-    crypto::{
-        CryptoHash, ValidatorPublicKey, ValidatorSignature, SECP256K1_PUBLIC_KEY_SIZE,
-        SECP256K1_SIGNATURE_SIZE,
-    },
+    crypto::{CryptoHash, ValidatorPublicKey, ValidatorSignature},
     data_types::Round,
 };
 use linera_execution::committee::Committee;
@@ -28,9 +25,9 @@ impl<T: Allocative + CertificateValue> Allocative for GenericCertificate<T> {
     fn visit<'a, 'b: 'a>(&self, visitor: &'a mut Visitor<'b>) {
         visitor.visit_field(Key::new("GenericCertificate_value"), &self.value);
         visitor.visit_field(Key::new("GenericCertificate_round"), &self.round);
-        for (_, _) in &self.signatures {
-            visitor.visit_simple(Key::new("ValidatorPublicKey"), SECP256K1_PUBLIC_KEY_SIZE);
-            visitor.visit_simple(Key::new("ValidatorSignature"), SECP256K1_SIGNATURE_SIZE);
+        for (public_key, signature) in &self.signatures {
+            visitor.visit_field(Key::new("ValidatorPublicKey"), public_key);
+            visitor.visit_field(Key::new("ValidatorSignature"), signature);
         }
     }
 }


linera-chain/src/certificate/generic.rs
```diff
@@ -20,6 +21,17 @@ pub struct GenericCertificate<T: CertificateValue> {
     signatures: Vec<(ValidatorPublicKey, ValidatorSignature)>,
 }
 
+impl<T: Allocative + CertificateValue> Allocative for GenericCertificate<T> {
+    fn visit<'a, 'b: 'a>(&self, visitor: &'a mut Visitor<'b>) {
+        visitor.visit_field(Key::new("GenericCertificate_value"), &self.value);
+        visitor.visit_field(Key::new("GenericCertificate_round"), &self.round);
+        for (public_key, _) in &self.signatures {
+            visitor.visit_field(Key::new("ValidatorPublicKey"), public_key);
```
deuszx, line 29: public keys also have a fixed size, don't they? Why assume the size for signatures but not public keys?

MathieuDutSik: You are right. Corrected.

linera-base/src/crypto/mod.rs
```diff
@@ -73,6 +92,22 @@ pub enum AccountPublicKey {
...
+            }
+            AccountPublicKey::Secp256k1(_) => {
+                visitor.visit_simple(Key::new("Secp256k1PublicKey"), SECP256K1_PUBLIC_KEY_SIZE);
+            }
+            AccountPublicKey::EvmSecp256k1(_) => {
+                visitor.visit_simple(Key::new("EvmPublicKey"), EVM_PUBLIC_KEY_SIZE);
+            }
+        }
+    }
+}
```
afck, line 76: Would something like this work?

```rust
#[derive(Allocative)]
pub enum AccountPublicKey {
    Ed25519(#[allocative(visit = visit_allocative_simple)] ed25519::Ed25519PublicKey),
}

fn visit_allocative_simple<'a, 'b, T>(_: &T, visitor: &'a mut Visitor<'b>) {
    visitor.visit_simple_sized::<T>();
}
```

linera-base/src/crypto/mod.rs
```diff
@@ -111,6 +146,40 @@ pub enum AccountSignature {
     },
 }
 
+impl Allocative for AccountSignature {
```
afck: This could also be derived, if we
* implement `Allocative` for the signature types, _or_
* use `#[allocative(visit = "visit_allocative_simple")]` on all fields.

linera-views/src/views/collection_view.rs
```diff
@@ -50,6 +52,20 @@ pub struct ByteCollectionView<C, W> {
...
 
+impl<C, W: Allocative> Allocative for ByteCollectionView<C, W> {
+    fn visit<'a, 'b: 'a>(&self, visitor: &'a mut Visitor<'b>) {
+        let name = Key::new("ByteCollectionView");
+        let size = mem::size_of::<Self>();
+        let mut visitor = visitor.enter(name, size);
+        let updates = self
+            .updates
+            .try_read()
+            .expect("acquire a read on the updates");
```
afck: Why can we expect that? Doesn't this fail whenever any task holds a write lock?

linera-views/src/views/collection_view.rs
```diff
@@ -1166,9 +1184,11 @@ where
 }
 
 /// A map view that serializes the indices.
-#[derive(Debug)]
+#[derive(Debug, Allocative)]
+#[allocative(bound = "W: Allocative, C, I")]
```
afck: (Just for consistency with `CollectionView`.)

```suggestion
#[allocative(bound = "C, I, W: Allocative")]
```

linera-views/src/views/hashable_wrapper.rs
```diff
@@ -26,6 +27,12 @@ pub struct WrappedHashableContainerView<C, W, O> {
     inner: W,
 }
 
+impl<C, W: Allocative, O> Allocative for WrappedHashableContainerView<C, W, O> {
+    fn visit<'a, 'b: 'a>(&self, visitor: &'a mut Visitor<'b>) {
+        self.inner.visit(visitor);
```
afck: Doesn't that mean we can derive it, and `skip` the other fields?
(But actually, we should probably not skip `stored_hash`, at least.)

linera-views/src/views/key_value_store_view.rs
```diff
@@ -198,15 +199,19 @@ impl SizeData {
...
+#[allocative(bound = "C")]
 pub struct KeyValueStoreView<C> {
+    #[allocative(skip)]
     context: C,
     deletion_set: DeletionSet,
     updates: BTreeMap<Vec<u8>, Update<Vec<u8>>>,
     stored_total_size: SizeData,
     total_size: SizeData,
     sizes: ByteMapView<C, u32>,
+    #[allocative(skip)]
```
afck: We could probably use `visit_allocative_simple` for this. (See above.)

linera-views/src/views/queue_view.rs
```diff
@@ -47,9 +48,12 @@ enum KeyTag {
 }
 
 /// A view that supports a FIFO queue for values of type `T`.
-#[derive(Debug)]
+#[derive(Debug, Allocative)]
+#[allocative(bound = "C, T: Allocative")]
 pub struct QueueView<C, T> {
+    #[allocative(skip)]
     context: C,
+    #[allocative(skip)]
```
afck: Same here.

linera-views/src/views/reentrant_collection_view.rs
```diff
@@ -83,6 +85,34 @@ pub struct ReentrantByteCollectionView<C, W> {
...
+impl<C, W: Allocative> Allocative for ReentrantByteCollectionView<C, W> {
+    fn visit<'a, 'b: 'a>(&self, visitor: &'a mut Visitor<'b>) {
+        let name = Key::new("ReentrantByteCollectionView");
+        let size = mem::size_of::<Self>();
+        let mut visitor = visitor.enter(name, size);
+
+        for (k, v) in &self.updates {
+            let key_name = Key::new("key");
+            visitor.visit_field(key_name, k);
+            //
```
afck:
```suggestion

```

linera-views/src/views/reentrant_collection_view.rs
```diff
@@ -83,6 +85,34 @@ pub struct ReentrantByteCollectionView<C, W> {
...
+            match v {
+                Update::Removed => {
+                    let key = Key::new("update_removed");
+                    visitor.visit_field(key, &());
+                }
+                Update::Set(v) => {
+                    let key = Key::new("update_set");
+                    let v = v
+                        .try_read()
+                        .expect("the view of ReentrantByteCollectionView");
```
afck: Can't this panic?

linera-views/src/views/collection_view.rs
MathieuDutSik: Ok, done.

linera-views/src/views/reentrant_collection_view.rs
MathieuDutSik: Ok.

linera-views/src/views/reentrant_collection_view.rs
MathieuDutSik: Yes.
Considering that we recently changed this behavior in `clone_unchecked`, we should avoid that.
But we do not have the option of throwing an error in that case. So, we visit only if we have access to `Some(v)`.

linera-base/src/crypto/mod.rs
MathieuDutSik: Thanks, that works!
How do you find this?

linera-views/src/views/collection_view.rs
MathieuDutSik: Yes.
There are two alternatives:
* One is to panic.
* The other is not to account for the memory usage. I switched to that.

linera-views/src/views/key_value_store_view.rs
MathieuDutSik: Yes, we can do that.
Thanks.

linera-views/src/views/queue_view.rs
MathieuDutSik: For the `context` I think it is better to skip. We want to keep track of the size of the view, and the context is I think out of it.

For the `stored_indices`, I indeed switch to `visit_allocative_simple`.

linera-views/src/views/hashable_wrapper.rs
MathieuDutSik: Ok, changed.

linera-base/src/crypto/mod.rs
MathieuDutSik: Yes, it turned out to be possible to do that.

linera-base/src/crypto/mod.rs
```diff
@@ -35,6 +36,11 @@ pub type ValidatorSignature = secp256k1::Secp256k1Signature;
 /// The key pair of a validator.
 pub type ValidatorKeypair = secp256k1::Secp256k1KeyPair;
 
+/// The SECP256k1 public key sizes also used for validators.
+pub const SECP256K1_PUBLIC_KEY_SIZE: usize = std::mem::size_of::<secp256k1::Secp256k1PublicKey>();
+/// The SECP256k1 signature sizes also used for validators.
```
afck:
```suggestion
/// The secp256k1 public key sizes also used for validators.
pub const SECP256K1_PUBLIC_KEY_SIZE: usize = std::mem::size_of::<secp256k1::Secp256k1PublicKey>();
/// The secp256k1 signature sizes also used for validators.
```

(I think that's how we spell it elsewhere.)

linera-base/src/crypto/mod.rs
```diff
@@ -35,6 +36,11 @@ pub type ValidatorSignature = secp256k1::Secp256k1Signature;
...
+pub const SECP256K1_SIGNATURE_SIZE: usize = std::mem::size_of::<secp256k1::Secp256k1Signature>();
+
```
afck: Actually, do we still need those at all? Aren't those "simple", too?

(And even if: If it's only used in one place I'd rather use `size_of` there directly?)

linera-chain/src/certificate/lite.rs
```diff
@@ -29,6 +30,18 @@ pub struct LiteCertificate<'a> {
...
 }
 
+impl Allocative for LiteCertificate<'_> {
+    fn visit<'a, 'b: 'a>(&self, visitor: &'a mut Visitor<'b>) {
+        visitor.visit_field(Key::new("LiteCertificate_value"), &self.value);
+        visitor.visit_field(Key::new("LiteCertificate_round"), &self.round);
+        let signatures = self.signatures.deref();
+        for (public_key, _) in signatures {
+            visitor.visit_field(Key::new("ValidatorPublicKey"), public_key);
+            visitor.visit_simple(Key::new("ValidatorSignature"), 64);
```
afck: Can we use `size_of` here instead of a number?

linera-chain/src/certificate/lite.rs
```diff
@@ -29,6 +30,18 @@ pub struct LiteCertificate<'a> {
     pub signatures: Cow<'a, [(ValidatorPublicKey, ValidatorSignature)]>,
 }
 
+impl Allocative for LiteCertificate<'_> {
```
afck, line 33: Is this necessary because `Allocative` doesn't support `Cow`?

linera-chain/src/certificate/lite.rs
```diff
@@ -29,6 +30,18 @@ pub struct LiteCertificate<'a> {
...
+    fn visit<'a, 'b: 'a>(&self, visitor: &'a mut Visitor<'b>) {
+        visitor.visit_field(Key::new("LiteCertificate_value"), &self.value);
+        visitor.visit_field(Key::new("LiteCertificate_round"), &self.round);
+        let signatures = self.signatures.deref();
```
afck: I wonder if we should only count it if its owned?

linera-chain/src/data_types/mod.rs
```diff
@@ -431,6 +442,14 @@ pub struct Vote<T> {
     pub signature: ValidatorSignature,
 }
 
+impl<T: Allocative> Allocative for Vote<T> {
```
afck: Looks like we should be able to derive this, and either use `visit_allocative_simple` or derive `Allocative` for `ValidatorSignature`, too?

linera-views/src/views/hashable_wrapper.rs
```diff
@@ -18,10 +20,14 @@ use crate::{
...
 /// A hash for `ContainerView` and storing of the hash for memoization purposes
-#[derive(Debug)]
+#[derive(Debug, Allocative)]
+#[allocative(bound = "C, O, W: Allocative")]
 pub struct WrappedHashableContainerView<C, W, O> {
+    #[allocative(skip)]
     _phantom: PhantomData<C>,
+    #[allocative(visit = visit_allocative_simple)]
     stored_hash: Option<O>,
+    #[allocative(visit = visit_allocative_simple)]
```
afck, line 30: (That will only count the stack size of the mutex, rather than its contents, will it? It would be more correct to add `size_of::<Option<O>>()` to that. Since it's just a hash, maybe it doesn't matter much, but I'd at least add a short comment.

To be clear: In either case I agree that we do _not_ need to acquire the mutex lock here!)

linera-views/src/views/key_value_store_view.rs
```diff
@@ -198,15 +199,19 @@ impl SizeData {
...
     context: C,
     deletion_set: DeletionSet,
     updates: BTreeMap<Vec<u8>, Update<Vec<u8>>>,
     stored_total_size: SizeData,
     total_size: SizeData,
     sizes: ByteMapView<C, u32>,
+    #[allocative(visit = visit_allocative_simple)]
     stored_hash: Option<HasherOutput>,
+    #[allocative(visit = visit_allocative_simple)]
```
afck, line 217: (Same here.)

linera-base/src/crypto/mod.rs
MathieuDutSik: Ok, thanks.

linera-base/src/crypto/mod.rs
MathieuDutSik: Yes, those constants are not needed since there are some with the same name in `src/crypto/secp256k1/mod.rs`.

linera-chain/src/certificate/lite.rs
MathieuDutSik: Yes.

linera-chain/src/certificate/lite.rs
MathieuDutSik: I would like that. However, the `.is_owned()` is available only in experimental.

afck: You can use `matches!(self.signatures, Cow::Owned(_))`?

linera-chain/src/data_types/mod.rs
MathieuDutSik: You are right. I thought it was not possible because I thought that `Secp256k1Sgnature` was defined outside of the crate. But no this is possible because it is a nesting:
```rust
pub struct Secp256k1Signature(pub Signature);
```

linera-views/src/views/hashable_wrapper.rs
MathieuDutSik: Thank you for your understanding. I added a note.

linera-views/src/views/key_value_store_view.rs
MathieuDutSik: Same note.

linera-chain/src/certificate/lite.rs
MathieuDutSik: You are, of course, right!
Wonder why adding some `is_owned()` is marked as experimental when it seems to be a basic feature.

linera-chain/src/certificate/lite.rs
MathieuDutSik: Since, we have implemented `Allocative` for `ValidatorSignature`, we can avoid that pesky constant.

linera-base/src/crypto/mod.rs
afck: In the `allocative` docs: https://docs.rs/allocative/latest/allocative/trait.Allocative.html#allocativevisit--

linera-base/src/crypto/secp256k1/mod.rs
```diff
@@ -31,10 +32,10 @@ use crate::doc_scalar;
 const SECP256K1_SCHEME_LABEL: &str = "secp256k1";
 
 /// Length of secp256k1 compressed public key.
-const SECP256K1_PUBLIC_KEY_SIZE: usize = 33;
+pub const SECP256K1_PUBLIC_KEY_SIZE: usize = 33;
 
 /// Length of secp256k1 signature.
-const SECP256K1_SIGNATURE_SIZE: usize = 64;
+pub const SECP256K1_SIGNATURE_SIZE: usize = 64;
```
afck: Can we just use `mem::size_of` instead? (Because in this case it _is_ about the in-memory size.)

linera-views/src/views/hashable_wrapper.rs
```diff
@@ -17,10 +19,15 @@ use crate::{
...
 /// Wrapping a view to memoize its hash.
-#[derive(Debug)]
+#[derive(Debug, Allocative)]
+#[allocative(bound = "C, O, W: Allocative")]
 pub struct WrappedHashableContainerView<C, W, O> {
+    #[allocative(skip)]
     _phantom: PhantomData<C>,
+    #[allocative(visit = visit_allocative_simple)]
     stored_hash: Option<O>,
+    // We are only keeping track of the size of the Mutex here.
```
afck:
```suggestion
    // We are only keeping track of the size of the mutex here.
```
(or `` the `Mutex` ``)

linera-views/src/views/key_value_store_view.rs
```diff
@@ -200,15 +201,20 @@ impl SizeData {
...
+    #[allocative(skip)]
     context: C,
     deletion_set: DeletionSet,
     updates: BTreeMap<Vec<u8>, Update<Vec<u8>>>,
     stored_total_size: SizeData,
     total_size: SizeData,
     sizes: ByteMapView<C, u32>,
+    #[allocative(visit = visit_allocative_simple)]
     stored_hash: Option<HasherOutput>,
+    // We are only keeping track of the size of the Mutex here.
```
afck: Here, too.

linera-chain/src/certificate/lite.rs
```diff
@@ -29,6 +30,19 @@ pub struct LiteCertificate<'a> {
     pub signatures: Cow<'a, [(ValidatorPublicKey, ValidatorSignature)]>,
 }
 
+impl Allocative for LiteCertificate<'_> {
```
afck, line 33: I wonder whether we could even use `derive` for `LiteCertificate`, and, similar to `visit_allocative_simple`, write a `visit_allocative_cow`. (Not a blocker; just a thought!)

I wonder if `allocative` would even accept pull requests for things like `Cow`.

linera-base/src/crypto/secp256k1/mod.rs
afck: Actually, are these constants even still needed, now that we use the `visit` annotation?

linera-views/src/views/hashable_wrapper.rs
MathieuDutSik: Ok, thanks.

linera-views/src/views/key_value_store_view.rs
MathieuDutSik: Yes.

linera-base/src/crypto/secp256k1/mod.rs
MathieuDutSik: Yes correct. No longer needed.

linera-chain/src/certificate/lite.rs
MathieuDutSik: Ok, I can work on that.

Issue #4686, opened 2025-09-29
Transition from multi- to single-user rounds can fail.


afck:
In [the consensus protocol](https://github.com/linera-io/linera-protocol/blob/main/linera-chain/src/manager.rs), the current round number is determined based on timeout certificates, the locking block, and block proposals. The latter are necessary, because all multi-leader rounds can be skipped without a timeout, just by making a proposal in a higher round.

Since due to oracles, block proposals accepted by one node won't necessarily also be accepted by another node, the `signed_proposal` field was introduced, to keep track of those proposals even if we reject them, because they mean that that round has started.

This is particularly important for the first single-leader round, because that one cannot be skipped, so all validators need to be convinced that it has started, in order for it to time out. There are a few bugs preventing this currently:
* The client doesn't update validators when they return `WrongRound` to a timeout request.
* The validators only update `signed_proposal` if its round is _strictly_ less than `SingleLeader(0)`.

We need to fix this, so that the following scenario is handled correctly:
* Half of the validators know about a proposal in `SingleLeader(0)`, but that proposal cannot be validated anymore (e.g. because of `assert_before`).
* The other validators are still in a `MultiLeader` round.
* A client wants to make a new proposal.

PR #4688, opened 2025-09-29
Fix consensus issues related to timeouts, rounds, ownership.

afck:
## Motivation

On `testnet_conway`, we ran into a situation where the client couldn't commit a block anymore on a chain where some validators were in round `SingleLeader(0)`, because they knew about a proposal in that round, and other validators were in `MultiLeader(0)` or `MultiLeader(1)`. The client didn't automatically update the latter ones. While debugging this, we also found a few other consensus issues.

## Proposal

* The current round calculation should take the `signed_proposal` into account.
* We also need to update `signed_proposal` for a proposal in round `SingleLeader(0)`.
* Some noisy warnings should be DEBUG-level.
* We should update our chain even if we are the only owner but a _regular_ owner, because in that case we might use the same secret key on another device.
* `request_leader_timeout_if_needed` should not fail if we couldn't get a timeout certificate: In some cases a proposal in the current round might still be successful.
* When requesting a timeout certificate, the `ValidatorUpdater` should handle `WrongRound` and `UnexpectedBlockHeight` correctly, and call `send_chain_information`.
* In `send_chain_information`, in addition to timeout certificates, also update the remote node about proposals and the locking block, to make sure they are in the same round.
* Recompute the current round in the chain manager when the `signed_proposal` changes.
* Added a temporary workaround to `can_propose` for https://github.com/linera-io/linera-protocol/issues/4694.
* In the client, don't return `WaitForTimeout` if the current round isn't at least `SingleLeader(0)`.
* Remove the `test_end_to_end_retry_pending_block` test: Since `prepare_chain` now fails, even if we are the only regular owner, it's difficult to make a pending block without simulating faulty-but-online validators. But we have several tests in `linera-core` (among them the new one) that already test `process_pending_block`.
* Fix `can_propose` (https://github.com/linera-io/linera-protocol/pull/4695).

We should soon try to deduplicate some of this code: https://github.com/linera-io/linera-protocol/issues/4690

## Test Plan

Tested on `testnet_conway`, where blocks could not be committed anymore on one chain. With these changes, it worked again.

~I created https://github.com/linera-io/linera-protocol/issues/4689 to add a regression test later.~ (Edit: Added a regression test, with Claude's help.)

## Release Plan

    - be released in a new SDK,

## Links

- Closes https://github.com/linera-io/linera-protocol/issues/4686.
- Closes https://github.com/linera-io/linera-protocol/issues/4689.
- Closes https://github.com/linera-io/linera-protocol/issues/4694.


linera-chain/src/manager.rs
```diff
@@ -578,6 +578,12 @@ where
                     .iter()
                     .map(|proposal| proposal.content.round),
             )
+            .chain(
+                self.signed_proposal
+                    .get()
+                    .iter()
+                    .map(|proposal| proposal.content.round),
+            )
```
deuszx: Please add some comment(s) explaining the approach here. maybe not even necessarily only this particular addition but the collection of all rounds in general.

linera-chain/src/manager.rs
```diff
@@ -685,7 +691,7 @@ where
     /// Sets the signed proposal, if it is newer than the known one, and not from a single-leader
     /// round. Returns whether it was updated.
     pub fn update_signed_proposal(&mut self, proposal: &BlockProposal) -> bool {
-        if proposal.content.round > Round::MultiLeader(u32::MAX) {
+        if proposal.content.round > Round::SingleLeader(0) {
```
deuszx, line 697: The comment says that the signed proposal canot be from the single-leader round. Can you explain why it can be now?

linera-core/src/client/mod.rs
```diff
@@ -1033,7 +1033,7 @@ impl<Env: Environment> Client<Env> {
 
         let validators = self.make_nodes(&committee)?;
         Box::pin(self.fetch_chain_info(chain_id, &validators)).await?;
-        communicate_with_quorum(
+        if let Err(e) = communicate_with_quorum(
```
deuszx: I hoped we can do this change in a separate PR.

linera-core/src/client/mod.rs
```diff
@@ -4129,11 +4133,14 @@ impl<Env: Environment> ChainClient<Env> {
...
-            .all_owners()
-            .any(|owner| Some(owner) != self.preferred_owner.as_ref())
+    /// Returns whether the given ownership includes anyone whose secret key we don't have, or
+    /// any regular (not super) owners. Returns `false` only if we are the only owner and a super.
+    fn has_other_or_regular_owners(&self, ownership: &ChainOwnership) -> bool {
+        !ownership.owners.is_empty()
+            || ownership
+                .super_owners
+                .iter()
+                .any(|owner| Some(owner) != self.preferred_owner.as_ref())
```
deuszx: Maybe this method would be simpler if it was on `ChainOwnership` and `self.preferred_owner` was the argument? Something like `is_singular_and_super_owner(self.preferred_owner)`? Right now it reads kinda ... backwards.

linera-core/src/updater.rs
```diff
@@ -300,6 +301,63 @@ where
         }?)
     }
 
+    async fn request_timeout(
```
deuszx, line 308: It would be good to get a comment here. Maybe Mr Anthropic can help

linera-core/src/updater.rs
```diff
@@ -300,6 +301,63 @@ where
...
+        round: Round,
+        height: BlockHeight,
+    ) -> Result<Box<ChainInfo>, ChainClientError> {
+        let query = ChainInfoQuery::new(chain_id).with_timeout(height, round);
+        let info = match self
+            .remote_node
+            .handle_chain_info_query(query.clone())
+            .await
+        {
+            Ok(info) => info,
```
deuszx: Can you short-circuit here with `handle_chain_info_query.await?` and not even handle the `Ok(info)` variant?

linera-core/src/updater.rs
```diff
@@ -300,6 +301,63 @@ where
...
+            Err(NodeError::WrongRound(validator_round)) => {
+                tracing::info!(
+                    "Failed to request timeout from validator {} because it sees \
+                            chain {} at round {} instead of {}.",
+                    self.remote_node.public_key,
+                    chain_id,
+                    validator_round,
+                    round
+                );
```
deuszx, line 325: Personally, I like to write these logs in such a way that there is no string interpolation:
```rust
tracing::info!(
  validator=?self.remote_node.public_key,
  chain_id,
  validator_round,
  round,
 "..."
);
```
It makes working with distributed tracing dashboards easier b/c every one of these arguments becomes a key in the json that can be filtered.

linera-core/src/updater.rs
```diff
@@ -300,6 +301,63 @@ where
...
+                    chain_id,
+                    found_block_height,
+                    CrossChainMessageDelivery::NonBlocking,
+                )
+                .await?;
+                self.remote_node.handle_chain_info_query(query).await?
+            }
+            Err(err) => Err(err)?,
+        };
+        Ok(info)
```
deuszx: nit: if you return easly for all other errors we can have one call to `handle_chain_info_query` after handling the two variants. The intention might be clearer that way:
```rust
  error => return error,
};
```

linera-core/src/updater.rs
```diff
@@ -438,82 +496,138 @@ where
...
+                        }
+                    };
+                for certificate in certificates {
+                    self.send_confirmed_certificate(certificate, delivery)
+                        .await?;
+                }
             }
-        }
+            info
+        } else {
```
deuszx, line 548: IIUC, this `else` branch is entered when `target_block_height.try_sub_one()` returns `None`. Which should mean that validator doesn't know about this chain yet.

I'm wondering why we want it to be handled differently from the `Some(_)` case (modulo updating validator about the chain). What I am thinking about is this approach

0. if validator doesn't know about the chain, send the required blobs.
1. Optimistically, send the latest certificate
2. proceed as before.

If the chain is at some `target_block_height` that is higher than 0, it's not enough to update it with the chain description (and all blobs' state) for it to handle the (chain info) query properly and get updated w/ the chain info.

I'm assuming that the purpose of `send_chain_info` is to update the validator about the `chain_id` up to `target_block_height`.

linera-core/src/updater.rs
```diff
@@ -438,82 +496,138 @@ where
...
+            }
+            Some(LockingBlock::Fast(proposal)) => {
+                if let Err(err) =
+                    Box::pin(self.send_block_proposal(Box::new(proposal), vec![])).await
+                {
+                    tracing::info!("Failed to send block proposal: {err}");
+                }
+            }
+            None => {}
+        }
```
deuszx, line 617: This method is doing now A LOT of requests. We should probably take another pass over it to see if we're not doing any unnecessarily.

linera-core/src/updater.rs
afck: I can't use `?`, because we do want to handle some errors.

But I could return `Ok(info)` early, if that's what you mean?

linera-core/src/client/mod.rs
```diff
@@ -2289,34 +2292,32 @@ impl<Env: Environment> ChainClient<Env> {
...
         let nodes = self.client.make_nodes(&committee)?;
         // Proceed to downloading received certificates.
+        let received_certificate_batches = Arc::new(std::sync::Mutex::new(Vec::new()));
         let result = communicate_with_quorum(
             &nodes,
             &committee,
             |_| (),
             |remote_node| {
                 let client = &self.client;
+                let received_certificate_batches = Arc::clone(&received_certificate_batches);
```
deuszx: Maybe a set will be better? to avoid duplicates.

linera-core/src/updater.rs
deuszx: Ah right, yes

linera-core/src/updater.rs
afck: Not sure I understand; isn't that exactly what I'm doing?

> if validator doesn't know about the chain, send the required blobs.

Yes, if the target block height is `0`, we send the chain description blob. That's exactly the else branch.

> Optimistically, send the latest certificate

That's all in the `if` branch.

> proceed as before.

Exactly: After the `if` and `else` we do the stuff we need both at height 0 and later, namely the current chain manager values.

deuszx: I meant that you don't do the same work in the `else` branch that you do in `if`.

linera-core/src/client/mod.rs
afck: This is not even the certificates but the _batches_ of certificates. We merge and deduplicate them in `receive_certificates_from_validators`.

linera-core/src/updater.rs
deuszx: For example, if validator is missing both the chain and hasn't seen the latest epoch change you won't update it properly b/c only in `if` branch you update it about the new committee.

linera-core/src/updater.rs
afck: Will add a comment; I don't trust Claude with this yet. :sweat_smile:

linera-core/src/updater.rs
afck: I agree: We should think about whether we really only need these for the round calculation (I think that's the case), and then send at most one of these values (the highest-round one).
I could add that tomorrow to this PR, or do it in a separate one.

linera-core/src/updater.rs
afck: Right… I need to think about this some more. I almost think it's not feasible without fully addressing https://github.com/linera-io/linera-protocol/issues/4690.

linera-core/src/client/mod.rs
afck: Anyway, I reverted this, as you requested; we can do this in a separate PR.

linera-chain/src/manager.rs
```diff
@@ -682,10 +700,13 @@ where
         self.ownership.get().super_owners.contains(owner)
     }
 
-    /// Sets the signed proposal, if it is newer than the known one, and not from a single-leader
-    /// round. Returns whether it was updated.
+    /// Sets the signed proposal, if it is newer than the known one, at most from the first
+    /// single-leader round. Returns whether it was updated.
+    ///
+    /// We don't update the signed proposal for any rounds later than `SingleLeader(0)`,
+    /// because single-leader rounds cannot be skipped without a timeout certificate.
```
deuszx, line 691: Why can we skip/update/timeout SingleLeader(0) without the timeout (certificate)?

afck: We can't! But a proposal _in_ `SingleLeader(0)` only makes that round the _current_ one, so we don't skip it. We only skipped all multi-leader rounds.

deuszx: Ah, I see. So the reason we don't return early `false` here for `SingleLeader(0)` isn't b/c it can't be skipped but because it's the first one for which we can update the `self.singed_proposal`?

In such case, I think the `>` condition is not the best test for it, it's not explained by the comment on this function.

afck: It's the _last_ one for which we should update `signed_proposal`, yes. For all later rounds, we _need_ a timeout certificate anyway, so there's no point.

E.g. we can't enter `SingleLeader(1)` without a _timeout certificate_ for `SingleLeader(0)` (or a validated block certificate for `SingleLeader(1)`), because `SingleLeader(0)` can't be skipped.

But we _can_ enter `SingleLeader(0)` due to a proposal in `SingleLeader(0)`, because all lower rounds (except `Fast`) can be skipped.

I don't think it would even be dangerous to remove this condition entirely, because we wouldn't accept that proposal in the first place if it skipped anything that it shouldn't. But returning `false` here means we can avoid one unnecessary `chain.save()`.

deuszx: so maybe simply:
```rust
if matches!(proposal.round, Round::SingleLeader(_)) { return false; }
```
?

afck: No, because for `SingleLeader(0)` we want to return `true`, and for later rounds like `Validator` we also want to return `false`.

deuszx: OK, I just think this one little `if` is hiding lots of things:
1. It special cases `SingleLeader(0)` b/c as you write it _enters_ the first "skippable" (w/ timeout) round
2. It uses the derived ordering between different round variants.

while the rest of this function is concerned about only the latter. 

But OK, let's leave it as is.

linera-base/src/ownership.rs
```diff
@@ -178,6 +178,11 @@ impl ChainOwnership {
         };
         Some(next_round)
     }
+
+    /// Returns whether the given owner is the only owner, and a super owner.
+    pub fn is_single_super_owner(&self, owner: &AccountOwner) -> bool {
+        self.owners.is_empty() && self.super_owners.len() == 1 && self.super_owners.contains(owner)
```
ma2bd, line 184: It doesn't matter if we have several super owners. They have to coordinate and behave as one anyway.

afck: But this is used for _synchronization_. Even if they coordinate via other channels and make sure that one proposes blocks only on Mondays, and the other on Wednesdays, they'd still need to download each other's blocks?

afck:
~Updating our chain as the only regular owner made prepare_chain fail in `apply_client_command`, which meant that `test_end_to_end_retry_pending_block` would fail because it wouldn't even set the pending block. For now I changed `apply_client_command` to just warn and continue anyway in that case, but it is maybe something that needs discussion.~

~The test also takes longer now, because we retry gRPC requests on error code `Unavailable`, which is what we get if the validator is simply down. And we retry 10 times by default.~

Edit: I removed that test.

linera-core/src/updater.rs
afck: Done in https://github.com/linera-io/linera-protocol/pull/4688/commits/2c212344a7bd1df8016ed1a88a635bc7c0245c1f.

afck:
I had to add the last commit to avoid calling `send_block_proposal` recursively, which can cause stack overflows. Calling `handle_block_proposal` is enough, because that will set the `signed_proposal` on the remote node, bringing it up to the current round. However, that only works if we also update the validators, because it relies on the fact that they will now set `signed_proposal` even if the proposal is in round `SingleLeader(0)`.

linera-client/src/client_context.rs
```diff
@@ -436,7 +436,9 @@ impl<Env: Environment, W: Persist<Target = Wallet>> ClientContext<Env, W> {
         Fut: Future<Output = Result<ClientOutcome<T>, E>>,
         Error: From<E>,
     {
-        client.prepare_chain().await?;
+        if let Err(err) = client.prepare_chain().await {
+            tracing::warn!("Failed to synchronize the chain: {err}");
+        }
```
deuszx: Why is it OK to continue if we didn't synchronize the chain properly? I thought the idea is that we're always synchronized before making the proposal.

linera-core/src/updater.rs
deuszx: I think this is my last outstanding  comment.

linera-client/src/client_context.rs
afck: Yes, see [this comment](https://github.com/linera-io/linera-protocol/pull/4688#issuecomment-3350895080) and what I wrote on Slack: I agree this isn't ideal, and we should probably move the "retry pending block" test to `client_tests` (we might already have one!) and remove the e2e test?

linera-core/src/updater.rs
afck: But this wasn't made _worse_ by this PR, was it? The same problem was already there with the early return?

afck: Also: Wouldn't it indirectly update the validator about the latest epoch? It would send a certificate for the chain description blob, and if the validator returned an error for that certificate, it would send the admin chain to the validator?

deuszx: Yeah, maybe. I don't know 😬 

> But this wasn't made _worse_ by this PR, was it? The same problem was already there with the early return?

Possibly, yeah. OK, let's see if it doesn't break later on. Right now it's client-side only so we can always patch.

linera-base/src/ownership.rs
ma2bd: Mmm we're both right. If I have a single user in two wallets and use one wallet on Mondays and the other on Wednesdays, I still to download blocks. BUT maybe those power users should know they need to call `linera sync`. It also depends whether benchmarks are ok without this heuristics at all.

ma2bd: Anyways, this is probably fine for now.

afck: True; I'm fine with either. I can change it in a follow-up PR; probably not worth cancelling this one again?

PR #4696, opened 2025-09-30
Rename `authenticated_signer` as `authenticated_owner`.

MathieuDutSik:
## Motivation

The term "authenticated signer" is not great, since when a message is transmitted, the term is used while who is going to sign is going to be another user/owner.

## Proposal

Rename the "authenticated signer" as "authenticated_owner".

## Test Plan

The CI seems to be the right way.

## Release Plan

It is likely to be a breaking change.

## Links

None.


ma2bd:
I believe what we agreed on is "authenticated_owner"

MathieuDutSik:
> I believe what we agreed on is "authenticated_owner"

This is not what I heard, and I can do the change.
I also prefer the "authenticated_owner".

ma2bd (approved):

PR #4699, opened 2025-09-30
Don't sync automatically if only super owners.

afck:
## Motivation

As discussed [here](https://github.com/linera-io/linera-protocol/pull/4688#discussion_r2389522578), we don't need to automatically sync even if there are multiple super owners.

## Proposal

Change `is_single_super_owner` accordingly.

## Test Plan

CI

## Release Plan

    - be released in a new SDK.

## Links

- Discussion: https://github.com/linera-io/linera-protocol/pull/4688#discussion_r2389522578

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -1847,7 +1847,7 @@ impl<Env: Environment> ChainClient<Env> {
                 .ownership
                 .is_super_owner_no_regular_owners(&owner)
         }) {
-            // We are the only owner, so we should be up to date.
+            // There are only super owners; they are expected to sync manually.
             ensure!(
                 info.next_block_height >= self.initial_next_block_height,
                 ChainClientError::WalletSynchronizationError
@@ -2012,7 +2012,7 @@ impl<Env: Environment> ChainClient<Env> {
                 .ownership
                 .is_super_owner_no_regular_owners(&owner)
         }) {
-            // If we are not a single super owner, we could be missing recent
+            // If we are not a super owner or there are regular owners, we could be missing recent
             // certificates created by other clients. Further synchronize blocks from the network.
             // This is a best-effort that depends on network conditions.
             info = self.client.synchronize_chain_state(self.chain_id).await?;


linera-core/src/client/mod.rs
```diff
@@ -1842,10 +1842,11 @@ impl<Env: Environment> ChainClient<Env> {
...
-        if self
-            .preferred_owner
-            .is_some_and(|owner| info.manager.ownership.is_single_super_owner(&owner))
-        {
+        if self.preferred_owner.is_some_and(|owner| {
+            info.manager
+                .ownership
+                .is_super_owner_no_regular_owners(&owner)
+        }) {
             // We are the only owner, so we should be up to date.
```
bart-linera: The comment is outdated after the change, isn't it?

Also, wouldn't it be true that we should be up to date if we're the only owner in general (as in, not necessarily a super owner)?

linera-core/src/client/mod.rs
```diff
@@ -2005,10 +2006,12 @@ impl<Env: Environment> ChainClient<Env> {
...
-            .preferred_owner
-            .is_none_or(|owner| !info.manager.ownership.is_single_super_owner(&owner))
-        {
+        if self.preferred_owner.is_none_or(|owner| {
+            !info
+                .manager
+                .ownership
+                .is_super_owner_no_regular_owners(&owner)
+        }) {
             // If we are not a single super owner, we could be missing recent
```
bart-linera: Same here.

linera-core/src/client/mod.rs
afck: Good catch!

The idea is that it's legitimate to use the same owner key on multiple devices (even though it can temporarily hurt liveness). Only super owners are supposed to be extra careful, and know when to sync manually.

bart-linera: Ah, makes sense :+1:

bart-linera (approved):
TBH, I feel like we're creating a footgun for super owners with this one, but approving since it does agree with what has been discussed in the comments linked in the description.

afck:
Yes, but super owners are a somewhat dangerous feature (for a chain's liveness) in the first place, and their whole point is optimizing for latency. So it makes sense to give them full control and avoid any unnecessary network requests.

Issue #4235, opened 2025-07-18
Fix Firefox support


Twey:
[Mozilla have recently changed the behaviour in Firefox of `new WebAssembly.Module` to forbid compiling from a `SharedArrayBuffer`](https://bugzilla.mozilla.org/show_bug.cgi?id=1965217).

Wasmer (with the web backend) compiles WebAssembly by passing a [view](https://rustwasm.github.io/wasm-bindgen/api/js_sys/struct.Uint8Array.html#method.view) of the underlying memory.  In our case, the underlying memory is a `SharedArrayBuffer`, so compilation now fails on Firefox.

We need to fix it by patching Wasmer to copy the data instead of using a view when the underlying memory is a `SharedArrayBuffer`, but we are so far behind Wasmer mainstream (2 major versions) that it seems worthwhile to try to upstream our changes and the patch (#4203) then upgrade rather than piling more patches onto our fork.

PR #4704, opened 2025-10-01
Web tests using Vitest and Playwright

Twey:
## Motivation

Currently all our testing for the Web client is manual.  We'd like to know when PRs break our example apps.

## Proposal

This PR sets up conventions for our Web tests that enable us to test in a variety of different environments.  Going forward we can add more and more in-depth tests to cover more scenarios, and if we backport this PR we can also use these tests as an automated healthcheck for the deployed Web apps.

We unify the Web tests using [Vitest](https://vitest.dev/) with a [Playwright](https://playwright.dev/) backend to run tests in real browsers (currently just Chromium, but we should also test at least on WebKit and Firefox once https://github.com/linera-io/linera-protocol/issues/4235 is fixed).

Add a couple of basic tests for `@linera/client`: getting a wallet from the faucet and initializing the client.

## Test Plan

This is the test plan :)

## Release Plan

    - be released in a new SDK.

## Links


web/@linera/client/tests/PrivateKeySigner.test.ts
```diff
@@ -22,7 +23,7 @@ test("signs message correctly", async () => {
   const cryptoHash =
     "c520e2b24b05e70c39c36d4aa98e9129ac0079ea002d4c382e6996ea11946d1e";
   const owner = signer.address().toLowerCase();
-  const signature = await signer.sign(owner, Buffer.from(cryptoHash, "hex"));
+  const signature = await signer.sign(owner, Uint8Array.from(cryptoHash.match(/.{1,2}/g).map((byte) => parseInt(byte, 16))));
```
Twey, line 26: `Buffer` exists only in Node.js, not in the browser.

.github/workflows/web.yml
```diff
@@ -80,12 +81,41 @@ jobs:
     - name: Install pnpm
       uses: pnpm/action-setup@v4
       with:
-        version: 10
+        version: 10.17
 
+    # Setup fixtures
+    - name: Put nightly toolchain file in place
+      run: |
+        ln -sf toolchains/nightly/rust-toolchain.toml
```
ndr-ds: Don't we already do this on L67?

web/@linera/client/docs/media/CONTRIBUTING.md
```diff
@@ -0,0 +1,268 @@
...
+* If you send a bug report, please ensure that your description is clear and has sufficient
+instructions to be able to reproduce the issue.
+
+
+## Pull Requests (PRs)
+
+To make a contribution to the code after discussing it in a GitHub issue,
+
+1. Fork the repo and create your branch from `main`.
+2. If you've added code that should be tested, add tests.
```
ndr-ds, line 19: What type of code shouldn't be tested? 🤔

ndr-ds (approved):

web/@linera/client/docs/media/CONTRIBUTING.md
Twey: I don't know, but you should take it up with the root `CONTRIBUTING.md` 😅 It's just copied here as part of doc generation.

.github/workflows/web.yml
Twey: Right you are 🤦 Good spot, thanks!

PR #4706, opened 2025-10-01
Don't do heavy syncing of sender chains unless requested.

### LINKED ISSUE #4705: Don't sync all received certificates on service startup.
In some cases this can take a long time, and it's not necessary to produce new blocks.

Full synchronization can happen in the background once the service is running.

afck:
## Motivation

`find_received_certificates_from_validator` and `find_received_certificates` can take a lot of time (and memory) if there are a lot of incoming messages.

## Proposal

Only do them as part of `synchronize_from_validators`. On notification, `prepare_chain` and when we're about to propose a block, only download what's necessary.

Download and process sender certificates in batches, to avoid the high memory usage.

Download them in a background task in the node service, with some sleep time in between, to limit interference with the other node service tasks.

(Mostly written with Claude Sonnet 4.5.)

## Test Plan

CI
A new test was added for the updated `prepare_chain`. `test_sparse_sender_chain` already exists, and tests notification handling with a sparse chain.

## Release Plan

    - be released in a new SDK.

## Links

- Closes #4705.

### DIFF
--- a/linera-client/src/chain_listener.rs
+++ b/linera-client/src/chain_listener.rs
@@ -422,6 +422,7 @@ impl<C: ClientContext + 'static> ChainListener<C> {
         cancellation_token: CancellationToken,
     ) -> Result<(), Error> {
         info!("Starting background certificate sync for chain {chain_id}");
+        let client = context.lock().await.make_chain_client(chain_id);
 
         loop {
             // Check if we should stop.
@@ -434,7 +435,6 @@ impl<C: ClientContext + 'static> ChainListener<C> {
             Self::sleep(sync_sleep_ms).await;
 
             // Sync one batch.
-            let client = context.lock().await.make_chain_client(chain_id);
             match client.sync_received_certificates_batch().await {
                 Ok(has_more) => {
                     if !has_more {
--- a/linera-client/src/chain_listener.rs
+++ b/linera-client/src/chain_listener.rs
@@ -213,7 +213,7 @@ impl<C: ClientContext> ChainListener<C> {
             let admin_chain_id = guard.wallet().genesis_admin_chain();
             guard
                 .make_chain_client(admin_chain_id)
-                .synchronize_from_validators()
+                .synchronize_chain_state(admin_chain_id)
                 .await?;
             BTreeMap::from_iter(
                 guard
--- a/linera-client/src/chain_listener.rs
+++ b/linera-client/src/chain_listener.rs
@@ -58,14 +58,6 @@ pub struct ChainListenerConfig {
         env = "LINERA_LISTENER_DELAY_AFTER"
     )]
     pub delay_after_ms: u64,
-
-    /// Number of certificates to fetch per batch during background synchronization.
-    #[arg(
-        long = "listener-sync-batch-size",
-        default_value = "100",
-        env = "LINERA_LISTENER_SYNC_BATCH_SIZE"
-    )]
-    pub sync_batch_size: usize,
 }
 
 type ContextChainClient<C> = ChainClient<<C as ClientContext>::Environment>;
@@ -240,17 +232,14 @@ impl<C: ClientContext + 'static> ChainListener<C> {
         // if enabled (sync_sleep_ms is Some).
         if let Some(sync_sleep_ms) = sync_sleep_ms {
             let context = Arc::clone(&self.context);
-            let config = Arc::clone(&self.config);
             let cancellation_token = self.cancellation_token.clone();
             for chain_id in chain_ids.keys() {
                 let context = Arc::clone(&context);
-                let config = Arc::clone(&config);
                 let cancellation_token = cancellation_token.clone();
                 let chain_id = *chain_id;
                 drop(linera_base::task::spawn(async move {
                     if let Err(e) = Self::background_sync_received_certificates(
                         context,
-                        config,
                         sync_sleep_ms,
                         chain_id,
                         cancellation_token,
@@ -425,10 +414,9 @@ impl<C: ClientContext + 'static> ChainListener<C> {
 
     /// Background task that syncs received certificates in small batches.
     /// This discovers unacknowledged sender blocks gradually without overwhelming the system.
-    #[instrument(skip(context, config, cancellation_token))]
+    #[instrument(skip(context, cancellation_token))]
     async fn background_sync_received_certificates(
         context: Arc<Mutex<C>>,
-        config: Arc<ChainListenerConfig>,
         sync_sleep_ms: u64,
         chain_id: ChainId,
         cancellation_token: CancellationToken,
@@ -447,10 +435,7 @@ impl<C: ClientContext + 'static> ChainListener<C> {
 
             // Sync one batch.
             let client = context.lock().await.make_chain_client(chain_id);
-            match client
-                .sync_received_certificates_batch(config.sync_batch_size)
-                .await
-            {
+            match client.sync_received_certificates_batch().await {
                 Ok(has_more) => {
                     if !has_more {
                         info!("Background sync completed for chain {chain_id}");
--- a/linera-client/src/chain_listener.rs
+++ b/linera-client/src/chain_listener.rs
@@ -58,6 +58,14 @@ pub struct ChainListenerConfig {
         env = "LINERA_LISTENER_DELAY_AFTER"
     )]
     pub delay_after_ms: u64,
+
+    /// Number of certificates to fetch per batch during background synchronization.
+    #[arg(
+        long = "listener-sync-batch-size",
+        default_value = "100",
+        env = "LINERA_LISTENER_SYNC_BATCH_SIZE"
+    )]
+    pub sync_batch_size: usize,
 }
 
 type ContextChainClient<C> = ChainClient<<C as ClientContext>::Environment>;
@@ -207,7 +215,10 @@ impl<C: ClientContext + 'static> ChainListener<C> {
 
     /// Runs the chain listener.
     #[instrument(skip(self))]
-    pub async fn run(mut self) -> Result<impl Future<Output = Result<(), Error>>, Error> {
+    pub async fn run(
+        mut self,
+        sync_sleep_ms: Option<u64>,
+    ) -> Result<impl Future<Output = Result<(), Error>>, Error> {
         let chain_ids = {
             let guard = self.context.lock().await;
             let admin_chain_id = guard.wallet().genesis_admin_chain();
@@ -225,24 +236,31 @@ impl<C: ClientContext + 'static> ChainListener<C> {
             )
         };
 
-        // Start background tasks to sync received certificates for each chain.
-        let context = Arc::clone(&self.context);
-        let cancellation_token = self.cancellation_token.clone();
-        for chain_id in chain_ids.keys() {
-            let context = Arc::clone(&context);
-            let cancellation_token = cancellation_token.clone();
-            let chain_id = *chain_id;
-            tokio::spawn(async move {
-                if let Err(e) = Self::background_sync_received_certificates(
-                    context,
-                    chain_id,
-                    cancellation_token,
-                )
-                .await
-                {
-                    warn!("Background sync failed for chain {}: {}", chain_id, e);
-                }
-            });
+        // Start background tasks to sync received certificates for each chain,
+        // if enabled (sync_sleep_ms is Some).
+        if let Some(sync_sleep_ms) = sync_sleep_ms {
+            let context = Arc::clone(&self.context);
+            let config = Arc::clone(&self.config);
+            let cancellation_token = self.cancellation_token.clone();
+            for chain_id in chain_ids.keys() {
+                let context = Arc::clone(&context);
+                let config = Arc::clone(&config);
+                let cancellation_token = cancellation_token.clone();
+                let chain_id = *chain_id;
+                drop(linera_base::task::spawn(async move {
+                    if let Err(e) = Self::background_sync_received_certificates(
+                        context,
+                        config,
+                        sync_sleep_ms,
+                        chain_id,
+                        cancellation_token,
+                    )
+                    .await
+                    {
+                        warn!("Background sync failed for chain {}: {}", chain_id, e);
+                    }
+                }));
+            }
         }
 
         Ok(async {
@@ -407,15 +425,14 @@ impl<C: ClientContext + 'static> ChainListener<C> {
 
     /// Background task that syncs received certificates in small batches.
     /// This discovers unacknowledged sender blocks gradually without overwhelming the system.
-    #[instrument(skip(context, cancellation_token))]
+    #[instrument(skip(context, config, cancellation_token))]
     async fn background_sync_received_certificates(
         context: Arc<Mutex<C>>,
+        config: Arc<ChainListenerConfig>,
+        sync_sleep_ms: u64,
         chain_id: ChainId,
         cancellation_token: CancellationToken,
     ) -> Result<(), Error> {
-        const BATCH_SIZE: usize = 100;
-        const SLEEP_BETWEEN_BATCHES_MS: u64 = 500;
-
         info!(
             "Starting background certificate sync for chain {}",
             chain_id
@@ -429,11 +446,14 @@ impl<C: ClientContext + 'static> ChainListener<C> {
             }
 
             // Sleep to avoid overwhelming the system.
-            Self::sleep(SLEEP_BETWEEN_BATCHES_MS).await;
+            Self::sleep(sync_sleep_ms).await;
 
             // Sync one batch.
             let client = context.lock().await.make_chain_client(chain_id);
-            match client.sync_received_certificates_batch(BATCH_SIZE).await {
+            match client
+                .sync_received_certificates_batch(config.sync_batch_size)
+                .await
+            {
                 Ok(has_more) => {
                     if !has_more {
                         info!("Background sync completed for chain {}", chain_id);
--- a/linera-client/src/chain_listener.rs
+++ b/linera-client/src/chain_listener.rs
@@ -257,7 +257,7 @@ impl<C: ClientContext + 'static> ChainListener<C> {
                     )
                     .await
                     {
-                        warn!("Background sync failed for chain {}: {}", chain_id, e);
+                        warn!("Background sync failed for chain {chain_id}: {e}");
                     }
                 }));
             }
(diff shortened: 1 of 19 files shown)


linera-core/src/client/mod.rs
```diff
@@ -2331,6 +2341,76 @@ impl<Env: Environment> ChainClient<Env> {
...
+        if missing_blocks.is_empty() {
+            return Ok(());
+        }
+
+        let (_, committee) = self.admin_committee().await?;
+        let nodes = self.client.make_nodes(&committee)?;
+        let (max_epoch, committees) = self.client.admin_committees().await?;
+
+        // Download certificates for each sender chain at the specific heights.
+        let certificates = stream::iter(missing_blocks.into_iter())
```
deuszx, line 2321: IIUC, the issue was here – correct me if I'm wrong @bart-linera  – that this `certificates` collection was slow to download.

linera-core/src/client/mod.rs
```diff
@@ -2331,6 +2341,76 @@ impl<Env: Environment> ChainClient<Env> {
...
+        // Receive and process the certificates.
+        for certificate in valid_certificates {
+            self.client
+                .receive_sender_certificate(
+                    certificate,
+                    ReceiveCertificateMode::AlreadyChecked,
+                    Some(nodes.clone()),
+                )
+                .await?;
+        }
```
deuszx, line 2368: If you extract this part then this function is no longer specific to sender chains and can be reused anywhere where we expect to download lots of certs.

linera-client/src/chain_listener.rs
```diff
@@ -213,7 +213,7 @@ impl<C: ClientContext> ChainListener<C> {
             let admin_chain_id = guard.wallet().genesis_admin_chain();
             guard
                 .make_chain_client(admin_chain_id)
-                .synchronize_from_validators()
+                .synchronize_chain_state(admin_chain_id)
```
deuszx, line 219: Weird that we already have the client for `admin_chain_id` but the method needs this argument anyway. Can't it be loaded from the context of the chain client?

linera-client/src/chain_listener.rs
```diff
@@ -225,6 +225,26 @@ impl<C: ClientContext> ChainListener<C> {
...
+            let chain_id = *chain_id;
+            tokio::spawn(async move {
+                if let Err(e) = Self::background_sync_received_certificates(
+                    context,
+                    chain_id,
+                    cancellation_token,
+                )
+                .await
+                {
+                    warn!("Background sync failed for chain {}: {}", chain_id, e);
```
deuszx: The same comment about structured logging as last time: instead of interpolating, use it as keys for the log.

linera-core/src/client/mod.rs
afck: The issue was also that at this point we tried to download all sender blocks that the validators know about.

Now `missing_blocks` will only contain the "removed" entries from the inbox, but not all the sender blocks of messages that have not been received in a confirmed block yet, so that should be much less.

linera-client/src/chain_listener.rs
```diff
@@ -385,6 +405,54 @@ impl<C: ClientContext> ChainListener<C> {
...
+            if cancellation_token.is_cancelled() {
+                info!("Background sync cancelled for chain {}", chain_id);
+                break;
+            }
+
+            // Sleep to avoid overwhelming the system.
+            Self::sleep(SLEEP_BETWEEN_BATCHES_MS).await;
+
+            // Sync one batch.
+            let client = context.lock().await.make_chain_client(chain_id);
```
deuszx: You shouldn't need to recreate the client in each iteration of the loop?

linera-core/src/client/mod.rs
afck: Do you have another call site in mind, where this is currently duplicated?

linera-client/src/chain_listener.rs
```diff
@@ -225,24 +236,31 @@ impl<C: ClientContext + 'static> ChainListener<C> {
...
+        if let Some(sync_sleep_ms) = sync_sleep_ms {
+            let context = Arc::clone(&self.context);
+            let config = Arc::clone(&self.config);
+            let cancellation_token = self.cancellation_token.clone();
+            for chain_id in chain_ids.keys() {
+                let context = Arc::clone(&context);
+                let config = Arc::clone(&config);
+                let cancellation_token = cancellation_token.clone();
+                let chain_id = *chain_id;
+                drop(linera_base::task::spawn(async move {
```
deuszx, line 240: Is this `drop` here safe?

linera-client/src/chain_listener.rs
afck: Yes, we could call it on `Client` directly, rather than `ChainClient`. But I'd prefer to leave that for another PR. Currently we don't really expose `Client` itself.

deuszx: We already have `synchronize_chain_state_from_committee` which does the same thing

linera-client/src/chain_listener.rs
afck: I asked @Twey in the meeting today, and he confirmed.

linera-client/src/chain_listener.rs
afck: I think I already did that?

linera-core/src/client/mod.rs
deuszx: I think we have too many places where logic is duplicated for this reason: we look at the funciton name without inspecting its content and the name is very specific (like here).

It's hard for me to believe that we had never needed what this method does.

linera-client/src/chain_listener.rs
deuszx: Yes, it's outdated.

linera-core/src/unit_tests/client_tests.rs
```diff
@@ -2781,3 +2781,78 @@ where
...
+            receiver.chain_id(),
+            receiver_info.block_hash,
+            receiver_info.next_block_height,
+        )
+        .await?;
+
+    // Test that prepare_chain downloads the missing sender blocks.
+    // The new client has the inbox state showing that a message was processed,
+    // but needs to download the sender block.
+    let info = receiver2.prepare_chain().await?;
```
deuszx, line 2852: It'd be nice to test the other change as well – that we don't download senders chains for which we haven't yet received (and included) the messages.

linera-core/src/client/mod.rs
```diff
@@ -3745,7 +3936,7 @@ impl<Env: Environment> ChainClient<Env> {
                     );
                     return Ok(());
                 }
```
deuszx, line 3865: Do we need the `_with_dependencies`? Does it ever make sense to download the block w/o them?

linera-core/src/client/mod.rs
```diff
@@ -2292,11 +2300,29 @@ impl<Env: Environment> ChainClient<Env> {
         #[cfg(with_metrics)]
         let _latency = metrics::FIND_RECEIVED_CERTIFICATES_LATENCY.measure_latency();
 
+        // Sync in batches until all received certificates are downloaded.
+        loop {
+            let has_more = self.sync_received_certificates_batch().await?;
+            if !has_more {
+                break;
+            }
+        }
```
deuszx: Isn't this equivalent to:
```rust
while self.sync_received_certificates_batch().await? {}
```

linera-client/src/chain_listener.rs
afck: Yes; both call `Client::synchronize_chain_state_from_committee` in the end.

linera-core/src/client/mod.rs
afck: Well, the "dependencies" here are relative to the receiver chain: the ancestor blocks that sent something to the same chain.

If that sender chain already sent something to another one of our tracked chains, we might have downloaded the same block but with a different set of ancestors.

linera-client/src/chain_listener.rs
deuszx: I meant: why not call `synchronize_chain_state_from_committee()` here instead?

linera-core/src/client/mod.rs
afck: You're probably right, but glancing over the code I didn't find any obvious candidates just now. So maybe also something for a follow-up PR?

deuszx: In this file there are a couple that try to do the same – `download_certificates`, `download_certificates_from`,

linera-core/src/client/mod.rs
```diff
@@ -820,14 +820,14 @@ impl<Env: Environment> Client<Env> {
         Ok(())
     }
 
-    /// Downloads and preprocesses all confirmed block certificates that sent any message to this
-    /// chain.
+    /// Downloads a limited batch of received certificates from a validator.
```
deuszx, line 823: I think "received certificates" isn't a good label for what we're trying to do here. It's a too big of a simplification and it's adding more confusion. It's a certificate for blocks where we receive an incoming message.

afck: We use the term "received certificates" a lot, though, and already before this PR. We need a term for it, because we can't write a full explanation in the 11 places where it currently occurs.

Maybe "sender certificates" is a better term?

linera-core/src/client/mod.rs
deuszx: There's a lot of duplication between this function and `synchronize_received_certificates_batch_from_validator`

linera-core/src/unit_tests/client_tests.rs
afck: Done in https://github.com/linera-io/linera-protocol/pull/4706/commits/2ca14d0fd2f35ec39141398e351a0239b091411d.

linera-core/src/client/mod.rs
afck: Done in https://github.com/linera-io/linera-protocol/pull/4706/commits/1eb7b2c7740815873d9688a5447265b8fadccff3.

linera-core/src/client/mod.rs
deuszx: "sender certificate" doesn't imply that the message as already received.

> We use the term "received certificates" a lot, though, and already before this PR. We need a term for it, because we can't write a full explanation in the 11 places where it currently occurs.

Maybe, but it's not an explanation for spreading the confusion. I won't fight for it though, I know I'll loose.

linera-client/src/chain_listener.rs
afck: Because then we have to explicitly get the committee, whereas `synchronize_chain_state` does that automatically.

linera-core/src/client/mod.rs
afck: I'm not fighting, I just don't know what to write instead?

linera-client/src/chain_listener.rs
deuszx: Sorry but this is dumb. We have:
- `synchronize_chain_state(chain_id)` which internally loads the committee and then synchronizes the state
- `synchronize_chain_state_from_committee(committee)` which requires committee but internally knows which chain to synchronize
- `synchronize_chain_state_from_committee` (on the client) that expects both

but we don't have `fn synchronize_chain_state` which hides both. Even though everything is available in the `ChainClient`: 
- which chain it's for
- how to get the committee

linera-core/src/unit_tests/client_tests.rs
```diff
@@ -2781,3 +2781,108 @@ where
...
+    sender2
+        .transfer_to_account(
+            AccountOwner::CHAIN,
+            Amount::from_tokens(1),
+            Account::chain(receiver.chain_id()),
+        )
+        .await
+        .unwrap_ok_committed();
+
+    // It does not synchronize the second message.
```
deuszx: Isn't it the case that it synchronizes the message (i.e. local client knows about it) but it doesn't _receive_ it (include in the block proposal)?

linera-core/src/unit_tests/client_tests.rs
```diff
@@ -2781,3 +2781,108 @@ where
...
+    receiver.synchronize_from_validators().await?;
+
+    // Verify the balance increased.
+    assert_eq!(
+        receiver.local_balance().await.unwrap(),
+        Amount::from_tokens(4)
```
deuszx: It'd be bit user to see that it's working as expected if you extracted `receive_start_balance` to a variable and here just tested that the `local_balance` equals `starting_balance + transfer_amount`.

afck: Or just set the starting balance to 0: https://github.com/linera-io/linera-protocol/pull/4706/commits/3a487b8aef656ddd600cd5abe537aa7bba65b107

linera-core/src/client/mod.rs
deuszx: I meant that every time we download a block at some height, we should download everything that's needed to process it locally – that means we need to download all of its dependencies. 

Unless it's a block for a sparse chain but that one is different, the default behavior should be that we download everything.

afck: Yes, but this one is for sparse chains, so maybe "dependencies" is the wrong word.

linera-core/src/client/mod.rs
```diff
@@ -2305,29 +2326,190 @@ impl<Env: Environment> ChainClient<Env> {
...
+        // Recursively collect all certificates we need, following
+        // the chain of previous_message_blocks back to next_outbox_height.
+        let mut certificates = BTreeMap::new();
+        let mut current_height = height;
+
+        // Stop if we've reached the height we've already processed.
+        while current_height >= next_outbox_height {
+            // Download the certificate for this height.
+            let downloaded = remote_node
+                .download_certificates_by_heights(sender_chain_id, vec![current_height])
```
deuszx, line 2404: Why not download the whole range between `next_outbox_height` and `height` with one request?

deuszx: Ah, b/c we might not need all of the blocks in the range – only those that link back (via `block.body.previous_message_block`) to our chain.

afck: Because we don't need all of them, only the ones that sent a message to our chain. If we downloaded all of them, we wouldn't make use of the "sparse sender chains" feature anymore.

afck: But since this happens based on a notification, it will usually not even be necessary to download more than one block. (Only if the previous sender block's notification was missed somehow.)

linera-core/src/client/mod.rs
```diff
@@ -2305,29 +2326,190 @@ impl<Env: Environment> ChainClient<Env> {
...
+            } else {
+                // No more dependencies.
+                break;
+            }
+        }
+
+        // Process certificates in ascending block height order (BTreeMap keeps them sorted).
+        for certificate in certificates.into_values() {
+            self.client
+                .receive_sender_certificate(
```
deuszx, line 2440: BTW, comment on `receive_sender_certificate` says:

> Processes the confirmed block in the local node **without executing it.**

but if you follow the code, in `ChainWorkerState::process_confirmed_block` we ARE executing it.

afck: Right, that's a bit outdated: We are executing it _if_ we have executed its parent. Otherwise we only preprocess it.

deuszx: Then the logic for "receiving" a sender certificate and non-sender one could be extracted & reused. 

Not in this PR but IMHO this duplication is affecting our ability to understand the code.

bart-linera: Not necessarily - only if it is the next block on the chain. If there is a gap, it won't be executed.

deuszx: Yes, but the logic could be shared among both sparse and full chains. it's that for full chains we expect it to always execute vs not always on sparse.

linera-core/src/client/mod.rs
```diff
@@ -2021,13 +2027,15 @@ impl<Env: Environment> ChainClient<Env> {
...
+        let missing_blocks = self
             .chain_state_view()
             .await?
-            .validate_incoming_bundles()
-            .await;
-        if matches!(result, Err(ChainError::MissingCrossChainUpdate { .. })) {
-            self.find_received_certificates().await?;
+            .collect_missing_sender_blocks()
+            .await?;
+        if !missing_blocks.is_empty() {
```
deuszx: You have the same test in `download_missing_sender_blocks` as well.

linera-core/src/client/mod.rs
afck: The term is also used in the `received_log` field in the chain state and chain info, so I'd keep using it with exactly that meaning for now: that it has been put in the inbox.

Changing that terminology is something I would prefer to do in a separate PR, and it probably needs some discussion first.

linera-client/src/chain_listener.rs
afck: I created https://github.com/linera-io/linera-protocol/issues/4715 for this. (It was not introduced in this PR.)

linera-client/src/chain_listener.rs
Twey: Yes totally safe.  On UNIX this is a `tokio::task::JoinHandle` and on the Web it's a `oneshot::Receiver`, and we ignore errors from sending to it.

linera-service/tests/linera_net_tests.rs
```diff
@@ -2085,19 +2112,22 @@ async fn test_wasm_end_to_end_allowances_fungible(config: impl LineraNetConfig)
...
         .await;
 
     app1.assert_allowance(&owner1, &owner2, Amount::from_tokens(93))
         .await;
 
-    // Call process inbox in order to synchronize from validators
-    node_service2.process_inbox(&chain2).await?;
-    app2.assert_allowance(&owner1, &owner2, Amount::from_tokens(93))
-        .await;
+    // Call process inbox in order to synchronize from validators.
```
deuszx: We don't process the inbox anymore and `get_allowance` doesn't mutate anything. How does it work now?

afck: Good catch! Removing that.

The `eventually` below means we just wait until client 2's node service has processed the `NewBlock` message and synchronized the chain. `process_inbox` was only used to synchronize here, not to actually process any messages.

linera-service/tests/linera_net_tests.rs
```diff
@@ -2505,7 +2546,10 @@ async fn test_wasm_end_to_end_non_fungible(config: impl LineraNetConfig) -> Resu
     )
     .await;
 
-    assert_eq!(node_service2.process_inbox(&chain2).await?.len(), 1);
+    assert!(
+        eventually(|| async { node_service2.process_inbox(&chain2).await.unwrap().len() == 1 })
+            .await
+    );
```
MathieuDutSik, line 2552: Those repetitions are annoying.

afck: I created https://github.com/linera-io/linera-protocol/issues/4718 to address this, but it will be a lot of work.

linera-core/src/client/mod.rs
afck: As discussed, I agree that there is a lot of potential for deduplication in this module, but I made a few attempts and they all became quite large and nontrivial, so I would prefer postponing that, too.

linera-core/src/client/mod.rs
afck: I verschlimmbessered the name in https://github.com/linera-io/linera-protocol/pull/4706.

bart-linera (approved):

PR #4708, opened 2025-10-01
Improve the `find_received_certificates` process

bart-linera:
## Motivation

`find_received_certificates` currently downloads all of the sender certificates for a chain from all the validators, before processing them. This is very inefficient with chains receiving lots of messages.

## Proposal

Collect the certificates to be downloaded from a quorum, and then try to download each certificate once. Process the certificates every time a download for a chain finishes.

## Test Plan

CI, manual testing if syncing with Testnet Conway is improved

## Release Plan

    - be released in a new SDK,

## Links

### DIFF
--- a/linera-core/src/client/validator_trackers.rs
+++ b/linera-core/src/client/validator_trackers.rs
@@ -127,10 +127,14 @@ impl ValidatorTracker {
 
 #[cfg(test)]
 mod test {
-    use linera_base::{crypto::CryptoHash, identifiers::ChainId};
+    use linera_base::{
+        crypto::{CryptoHash, ValidatorKeypair},
+        data_types::BlockHeight,
+        identifiers::ChainId,
+    };
     use linera_chain::data_types::ChainAndHeight;
 
-    use super::ValidatorTracker;
+    use super::{super::received_log::ReceivedLogs, ValidatorTracker, ValidatorTrackers};
 
     #[test]
     fn test_validator_tracker() {
@@ -164,5 +168,51 @@ mod test {
             height: 0.into(),
         });
         assert_eq!(tracker.current_tracker_value, 3);
+    }
+
+    #[test]
+    fn test_filter_out_already_known() {
+        let chain1 = ChainId(CryptoHash::test_hash("chain1"));
+        let chain2 = ChainId(CryptoHash::test_hash("chain2"));
+
+        let validator = ValidatorKeypair::generate().public_key;
+
+        let log: Vec<_> = vec![
+            (chain1, 0),
+            (chain2, 0),
+            (chain1, 1),
+            (chain1, 2),
+            (chain2, 1),
+            (chain1, 3),
+            (chain2, 2),
+        ]
+        .into_iter()
+        .map(|(chain_id, height)| ChainAndHeight {
+            chain_id,
+            height: height.into(),
+        })
+        .collect();
+
+        let mut received_log = ReceivedLogs::from_received_result(vec![(validator, log.clone())]);
+
+        assert_eq!(received_log.num_chains(), 2);
+        assert_eq!(received_log.num_certs(), 7);
+
+        let mut tracker = ValidatorTrackers::new(
+            vec![(validator, log)],
+            &vec![(validator, 0)].into_iter().collect(),
+        );
+
+        let local_heights = vec![(chain1, BlockHeight(3)), (chain2, BlockHeight(3))]
+            .into_iter()
+            .collect();
+
+        tracker.filter_out_already_known(&mut received_log, local_heights);
+
+        assert_eq!(received_log.num_chains(), 2); // we do not remove empty chains
+        assert_eq!(received_log.num_certs(), 1);
+
+        // tracker should have shifted to point to (chain1, 3)
+        assert_eq!(tracker.0[&validator].current_tracker_value, 5);
     }
 }


linera-core/src/client/mod.rs
```diff
@@ -825,149 +827,96 @@ impl<Env: Environment> Client<Env> {
...
+        remote_heights: Vec<BlockHeight>,
+        sender: mpsc::UnboundedSender<ChainAndHeight>,
+    ) {
+        let certificates = loop {
+            let Some(remote_node) = nodes.pop() else {
+                error!("could not download certificates for {}", sender_chain_id);
+                return;
+            };
+            if let Ok(certificates) = remote_node
+                .download_certificates_by_heights(sender_chain_id, remote_heights.clone())
```
afck: Isn't that introducing yet another place where we go through the validators sequentially and are slowed down a lot by the slowest one?

bart-linera: Yes, but I don't think we have a better solution right now :disappointed:  And if there are lots of sender chains, downloading them in parallel will mean that even if some of them are slowed down, the others will be progressing in the meantime, reducing the impact.

afck: I'd still prefer something like we do in `update_local_node_with_blobs_from`: Retry more validators in parallel, the longer it takes. Might be a better compromise?

afck: (And I think we need to properly handle both cases: the one we're seeing currently, with lots of sender chains, but also the one with a single extremely chatty sender chain.)

linera-core/src/client/mod.rs
```diff
@@ -825,149 +827,96 @@ impl<Env: Environment> Client<Env> {
...
+        for certificate in certificates {
+            let hash = certificate.hash();
+            let chain_id = certificate.block().header.chain_id;
+            let height = certificate.block().header.height;
+            let mode = ReceiveCertificateMode::NeedsCheck;
+            if let Err(err) = self
+                .receive_sender_certificate(certificate, mode, None)
+                .await
+            {
+                error!("Received invalid certificate {hash}: {err}");
```
afck: If the first node we try sends us too few or even invalid certificates, it looks like we'd simply give up and in the end consider synchronization "complete", instead of trying another node?

linera-core/src/client/mod.rs
afck: `remote_heights` could theoretically also be way too many, could it? I mean, even the list of heights could be too much.

In the current code we kind of have automatic batching for this, too, because we only ask for as many heights at once as we got in a single chain info response.

bart-linera: Good point. Yes, I think retrying validators in parallel should work.

I also think we should handle both cases properly, but this should at least not be worse than before with a chatty sender chain, should it? (Well, except for what you noted below, that `remote_heights` itself could be huge - good point, I'll re-add batching in such a case.)

linera-core/src/client/mod.rs
bart-linera: Hmm, that's right. How about collecting the heights that failed, and just looping until this collection is empty or we run out of nodes?

linera-core/src/client/mod.rs
deuszx: What if we combine both approaches:
- based on the chain info response decide how big of a batch we can request
- split the `remote_heights` into `n` batches
- select `n` validators from the set and make `n` requests in parallel
- collect the responses and, after some timeout, retry the remaining batches with the validators that had already responded
?

linera-core/src/client/mod.rs
afck: Yes, that sounds good.

afck: You're also already collecting the information about which validator _claims_ to have which heights. We could use that, and avoid asking validators for heights that we already know they don't have.

linera-core/src/client/mod.rs
```diff
@@ -825,149 +827,96 @@ impl<Env: Environment> Client<Env> {
...
+        };
 
-        // We keep track of the height we've successfully downloaded and checked, per chain.
-        let mut downloaded_heights = BTreeMap::new();
-        // And we make a list of chains we already fully have locally. We need to make sure to
-        // put all their sent messages into the inbox.
-        let mut other_sender_chains = Vec::new();
+        if !certificates.is_empty() {
+            trace!(
+                "got {} certificates for chain {}",
```
deuszx: instead of interpolating - use proper structured logging:
```
trace!(
chain=%sender_chain_id,
certificate_count=%certificates.len(),
"received certificates"
)
```

deuszx: here and everywhere else where you have logging.

linera-core/src/client/mod.rs
```diff
@@ -825,149 +827,96 @@ impl<Env: Environment> Client<Env> {
...
-                    // the skipped certificate again.
-                    warn!("Skipping received certificate from past epoch {epoch:?}");
-                }
-                CheckCertificateResult::New => {
-                    certificates_by_height_by_chain
-                        .entry(sender_chain_id)
-                        .or_insert_with(BTreeMap::new)
-                        .insert(height, confirmed_block_certificate);
-                }
+        for certificate in certificates {
```
deuszx: Certificates have to be processed in ascending order – let's sort them first before receiving.

linera-core/src/client/mod.rs
```diff
@@ -4297,18 +4361,123 @@ impl Drop for AbortOnDrop {
...
+            tracker.downloaded_cert(chain_and_height);
+        }
+    }
+
+    fn into_map(self) -> BTreeMap<ValidatorPublicKey, u64> {
+        self.0
+            .into_iter()
+            .map(|(validator, tracker)| (validator, tracker.current_tracker_value))
+            .collect()
+    }
```
deuszx: I'd love to see the new code put into its respective files to simplify the review.

linera-core/src/client/mod.rs
```diff
@@ -4297,18 +4361,123 @@ impl Drop for AbortOnDrop {
...
+    fn num_chains(&self) -> usize {
+        self.0.len()
+    }
+
+    fn num_certs(&self) -> usize {
+        self.0.values().map(|heights| heights.len()).sum()
+    }
+}
+
+struct ValidatorTracker {
```
deuszx: Some comments please.

linera-core/src/client/mod.rs
```diff
@@ -4297,18 +4361,123 @@ impl Drop for AbortOnDrop {
...
+            already_downloaded: BTreeSet::new(),
+        }
+    }
+
+    fn downloaded_cert(&mut self, chain_and_height: ChainAndHeight) {
+        self.already_downloaded.insert(chain_and_height);
+        self.maximize_tracker();
+    }
+
+    fn maximize_tracker(&mut self) {
```
deuszx: What does it even mean to _maximize a tracker_? Now everyone reading this code have to reverse engineer your thought process.

Add comment

linera-core/src/client/mod.rs
```diff
@@ -825,149 +827,96 @@ impl<Env: Environment> Client<Env> {
...
-        for entry in remote_log {
-            if certificates_by_height_by_chain
-                .get(&entry.chain_id)
-                .is_some_and(|certs| certs.contains_key(&entry.height))
-            {
-                tracker += 1;
-            } else {
+    /// Downloads the log of received messages for a chain from a validator.
+    #[instrument(level = "trace", skip(self))]
+    async fn get_received_log_from_validator(
```
deuszx, line 966: pub methods (`find_received_certificates`) should come before private ones.

linera-core/src/client/mod.rs
```diff
@@ -2237,73 +2093,281 @@ impl<Env: Environment> ChainClient<Env> {
...
+        received_logs: Vec<(ValidatorPublicKey, Vec<ChainAndHeight>)>,
+        trackers: HashMap<ValidatorPublicKey, u64>,
+        nodes: Vec<RemoteNode<Env::ValidatorNode>>,
+        cancellation_token: Option<CancellationToken>,
+    ) -> Result<(), ChainClientError> {
+        let (received_logs, mut validator_trackers) = {
+            (
+                ReceivedLogs::from_received_result(received_logs.clone()),
+                ValidatorTrackers::new(received_logs, &trackers),
+            )
```
deuszx: Why not expect those types by the function right away? Instead of making the mapping at the very beginning

linera-core/src/client/mod.rs
```diff
@@ -2237,73 +2093,281 @@ impl<Env: Environment> ChainClient<Env> {
...
+            .await?;
+
+        info!("find_received_certificates finished");
+
+        Ok(())
+    }
+
+    /// Computes the heights of the blocks in the sender chains we need to download the
+    /// certificates for.
+    async fn prepare_remote_heights(
```
deuszx: I keep jumping up and down reading this file b/c you decided to define the method before using it: I read the definition, read two more definitions, then see its usages, then go back up to see how the definition looks like.

deuszx: A `get_heights_to_download` is a better name I think – "prepare" is ambiguous and here we just find out which (remote) heights are higher than our local ones.

linera-core/src/client/mod.rs
```diff
@@ -2237,73 +2093,281 @@ impl<Env: Environment> ChainClient<Env> {
...
+        self.download_sender_certificates(received_logs, trackers, nodes, cancellation_token)
+            .await?;
+
+        info!("find_received_certificates finished");
+
+        Ok(())
+    }
+
+    /// Computes the heights of the blocks in the sender chains we need to download the
+    /// certificates for.
```
deuszx: "Compares validators' received logs of sender chains with local node information and returns per chain block heights that sent us a message which we didn't see yet."

linera-core/src/client/mod.rs
```diff
@@ -4297,18 +4361,123 @@ impl Drop for AbortOnDrop {
...
-    public_key: ValidatorPublicKey,
-    /// The new tracker value for that validator.
-    tracker: u64,
-    /// The downloaded certificates. The signatures were already checked and they are ready
-    /// to be processed.
-    certificates: Vec<ConfirmedBlockCertificate>,
-    /// Sender chains that were already up to date locally. We need to ensure their messages
-    /// are delivered.
-    other_sender_chains: Vec<ChainId>,
+struct ReceivedLogs(BTreeMap<ChainId, BTreeMap<BlockHeight, Vec<(ValidatorPublicKey, usize)>>>);
```
deuszx: What's the `usize` for? I can't see it being used anywhere

linera-core/src/client/mod.rs
bart-linera: It's a bit more complicated than it seems, because we cannot simply naively split everything into batches, as then we might end up with stuff in one batch depending on stuff in another batch. But I guess we could sort per chain and height first, and then split _that_ :thinking:

linera-core/src/client/mod.rs
bart-linera: They should be sorted already, as `download_certificates_by_heights` is supposed to return the certificates in the same order it was supplied heights, and that would be already sorted because it eventually comes from a `BTreeSet`.

linera-core/src/client/mod.rs
bart-linera: Good point. I think I intended to use it to figure out how the trackers can be updated, but eventually settled on using a different solution.

linera-core/src/client/mod.rs
deuszx: But downloading is independent. Processing might not be but we should be able to simply download certs, by height, however we want.

bart-linera: Yes, but if we decouple downloading from processing completely, we could be downloading stuff that we might not be able to process for a long time, and unnecessarily increase memory usage. Doing this per chain is a simple way to avoid that, and it shouldn't hurt efficiency significantly.

deuszx: OK, I forgot that you're now pipelining.

bart-linera: Fixed in [ce703a9](https://github.com/linera-io/linera-protocol/pull/4708/commits/ce703a97c151edcd2c4e9dc3f59935fe2872870b)

linera-core/src/client/mod.rs
bart-linera: Fixed in [d4b8d84](https://github.com/linera-io/linera-protocol/pull/4708/commits/d4b8d8447d38307363bb63ca1b56deb37a6fa11e) and [6d08031](https://github.com/linera-io/linera-protocol/pull/4708/commits/6d08031f39f57f417bed59605a03b375eee6613d)

linera-core/src/client/mod.rs
bart-linera: Fixed in [53cb992](https://github.com/linera-io/linera-protocol/pull/4708/commits/53cb9925673c7b7fc3d19a77137318250b6117b4)

linera-core/src/client/mod.rs
bart-linera: Fixed in [c5f5672](https://github.com/linera-io/linera-protocol/pull/4708/commits/c5f567251be27420d0672aec4d4d22df10cfc266)

linera-core/src/client/mod.rs
bart-linera: Fixed in [bc049a6](https://github.com/linera-io/linera-protocol/pull/4708/commits/bc049a6265e7713459d4cd54019deef5e7370636)

linera-core/src/client/mod.rs
bart-linera: Fixed in [bc049a6](https://github.com/linera-io/linera-protocol/pull/4708/commits/bc049a6265e7713459d4cd54019deef5e7370636)

linera-core/src/client/mod.rs
bart-linera: Fixed in [bc049a6](https://github.com/linera-io/linera-protocol/pull/4708/commits/bc049a6265e7713459d4cd54019deef5e7370636)

linera-core/src/client/mod.rs
bart-linera: (Hopefully) fixed in [9f5b3a1](https://github.com/linera-io/linera-protocol/pull/4708/commits/9f5b3a19556b0d1ee874495bcbe150898cbb0e97)

linera-core/src/client/mod.rs
bart-linera: (Hopefully) fixed in [9f5b3a1](https://github.com/linera-io/linera-protocol/pull/4708/commits/9f5b3a19556b0d1ee874495bcbe150898cbb0e97)

bart-linera: Fixed in [9f5b3a1](https://github.com/linera-io/linera-protocol/pull/4708/commits/9f5b3a19556b0d1ee874495bcbe150898cbb0e97)

linera-core/src/client/mod.rs
bart-linera: Fixed in [9f5b3a1](https://github.com/linera-io/linera-protocol/pull/4708/commits/9f5b3a19556b0d1ee874495bcbe150898cbb0e97)

linera-core/src/client/mod.rs
bart-linera: Fixed in [bc049a6](https://github.com/linera-io/linera-protocol/pull/4708/commits/bc049a6265e7713459d4cd54019deef5e7370636)

linera-core/src/client/mod.rs
bart-linera: Fixed in [53cb992](https://github.com/linera-io/linera-protocol/pull/4708/commits/53cb9925673c7b7fc3d19a77137318250b6117b4)

linera-core/src/client/received_log.rs
```diff
@@ -0,0 +1,114 @@
...
+            })
+    }
+}
+
+/// Iterator adapter lazily yielding batches of size `self.batch_size` from
+/// `self.iterator`.
+pub(super) struct LazyBatch<I: Iterator<Item = (ChainAndHeight, ValidatorPublicKey)>> {
+    iterator: I,
+    batch_size: usize,
+}
```
deuszx: Isn't this equivalent to `iterator.chunks(batch_size)` ?

bart-linera: It should be - but when I tried using `Itertools::chunks`, borrow checker would complain for some reason :man_shrugging:

linera-core/src/client/mod.rs
```diff
@@ -4271,6 +4368,40 @@ impl<Env: Environment> ChainClient<Env> {
     }
 }
 
+/// Performs `f` in parallel on multiple nodes, starting with a quadratically increasing delay on
+/// each subsequent node. Returns error `err` is all of the nodes fail.
+async fn communicate_in_parallel<'a, A, E1, E2, F, G, R, V>(
```
ndr-ds: One small nit here is that `FuturesUnordered` will poll the futures concurrently in the current task. It doesn't actually introduce any real parallelism. For true parallelism you need to actually spawn tokio tasks with a `JoinSet` or something like that, but since the futures are doing network calls, I don't think actual parallelism will improve anything here.
So this is mostly a nit on the naming/wording used here 😅

linera-core/src/client/received_log.rs
```diff
@@ -0,0 +1,114 @@
...
+impl<I: Iterator<Item = (ChainAndHeight, ValidatorPublicKey)>> Iterator for LazyBatch<I> {
+    type Item = ReceivedLogs;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.iterator.next().map(|next_item| {
+            ReceivedLogs::from_iterator(
+                std::iter::once(next_item)
+                    .chain((&mut self.iterator).take(self.batch_size.saturating_sub(1))),
+            )
+        })
```
afck: ~So that is a batch of the required size, but with one item per (validator, chain ID, height) _triple_, is it? So it could be one giant batch for one validator, or many small batches for different validators?~

I see, it will usually be the latter because the validators are the inner loop?

linera-core/src/client/mod.rs
```diff
@@ -825,149 +831,135 @@ impl<Env: Environment> Client<Env> {
...
-        let info = remote_node.handle_chain_info_query(query).await?;
-        let received_entries = info.requested_received_log.len();
-        let has_more = received_entries >= CHAIN_INFO_MAX_RECEIVED_LOG_ENTRIES;
-
-        let remote_log = info.requested_received_log;
-        let remote_heights = Self::heights_per_chain(&remote_log);
+        sender_chain_id: ChainId,
+        nodes: &[RemoteNode<Env::ValidatorNode>],
+        received_log: &ReceivedLogs,
+        mut remote_heights: Vec<BlockHeight>,
```
afck, line 841: Can't those be computed from `received_log`?

linera-core/src/client/mod.rs
```diff
@@ -825,149 +831,135 @@ impl<Env: Environment> Client<Env> {
...
+        sender: mpsc::UnboundedSender<ChainAndHeight>,
+    ) {
+        while !remote_heights.is_empty() {
+            let mut remote_heights_clone = remote_heights.clone();
+            let certificates = match communicate_in_parallel(
+                nodes,
+                async move |remote_node| {
+                    // No need trying to download certificates the validator didn't have in their
+                    // log - we'll retry downloading the remaining ones next time we loop.
+                    remote_heights_clone.retain(|height| {
```
afck: Isn't this moving one `remote_heights` clone into the closure, that will get filtered repeatedly, rather than cloning it again in each iteration and filtering using the validator for _that_ iteration only?

linera-core/src/client/mod.rs
```diff
@@ -825,149 +831,135 @@ impl<Env: Environment> Client<Env> {
...
+                let hash = certificate.hash();
+                let chain_id = certificate.block().header.chain_id;
+                let height = certificate.block().header.height;
+                let mode = ReceiveCertificateMode::NeedsCheck;
+                if let Err(err) = self
+                    .receive_sender_certificate(certificate, mode, None)
+                    .await
+                {
+                    error!(
+                        error = %err,
```
afck: Why not call the variable `error`?
```suggestion
                        %error,
```

linera-core/src/client/mod.rs
```diff
@@ -825,149 +831,135 @@ impl<Env: Environment> Client<Env> {
...
+            for certificate in certificates {
+                let hash = certificate.hash();
+                let chain_id = certificate.block().header.chain_id;
+                let height = certificate.block().header.height;
+                let mode = ReceiveCertificateMode::NeedsCheck;
+                if let Err(err) = self
+                    .receive_sender_certificate(certificate, mode, None)
+                    .await
+                {
+                    error!(
```
afck: I'd make that `info` at most: We don't want error-level logs that can be caused by a single bad validator.

linera-core/src/client/mod.rs
```diff
@@ -825,149 +831,135 @@ impl<Env: Environment> Client<Env> {
...
-
-        let remote_log = info.requested_received_log;
-        let remote_heights = Self::heights_per_chain(&remote_log);
+        sender_chain_id: ChainId,
+        nodes: &[RemoteNode<Env::ValidatorNode>],
+        received_log: &ReceivedLogs,
+        mut remote_heights: Vec<BlockHeight>,
+        sender: mpsc::UnboundedSender<ChainAndHeight>,
+    ) {
+        while !remote_heights.is_empty() {
```
afck, line 853: Can't a single bad validator make us loop indefinitely now? They give us some height that nobody has yet, and then never deliver it?

linera-core/src/client/mod.rs
```diff
@@ -1216,12 +1191,10 @@ impl<Env: Environment> Client<Env> {
...
         // Deduplicate IDs.
         let blob_ids = blob_ids.into_iter().collect::<BTreeSet<_>>();
-        stream::iter(blob_ids.into_iter().map(|blob_id| async move {
-            let mut stream = remote_nodes
-                .iter()
-                .zip(0..)
-                .map(|(remote_node, i)| async move {
-                    linera_base::time::timer::sleep(timeout * i * i).await;
+        stream::iter(blob_ids.into_iter().map(|blob_id| {
+            communicate_in_parallel(
```
afck: :tada:

linera-core/src/client/mod.rs
```diff
@@ -2261,73 +2133,298 @@ impl<Env: Environment> ChainClient<Env> {
...
         .await;
 
-        let ((), batches_with_progress) = match result {
-            Ok(result) => result,
-            Err(CommunicationError::Trusted(NodeError::InactiveChain(id))) if id == chain_id => {
-                // The chain is visibly not active (yet or any more) so there is no need
-                // to synchronize received certificates.
-                return Ok(false);
+        trace!("got result from communicate_with_quorum");
+        trace!(received_log_batches_len=%received_log_batches.lock().unwrap().len());
```
```suggestion
        trace!(received_log_batches_len = %received_log_batches.lock().unwrap().len());
```

afck: Should that really be a separate `trace!` invocation?

linera-core/src/client/mod.rs
```diff
@@ -2261,73 +2133,298 @@ impl<Env: Environment> ChainClient<Env> {
...
+            num_certs = %received_logs.num_certs(),
+            "download_sender_certificates: number of chains and certificates to sync",
+        );
+
+        let remote_heights = self
+            .get_heights_to_download(&received_logs, &mut validator_trackers)
+            .await?;
+
+        debug!(
+            remaining_total_certificates=%remote_heights.values().map(|h| h.len()).sum::<usize>(),
```
```suggestion
            remaining_total_certificates = %remote_heights.values().map(|h| h.len()).sum::<usize>(),
```

linera-core/src/client/received_log.rs
bart-linera: It depends how many chains there are - because the batch will then be handed over to tasks, with one task per chain, and each task will then take the set of validators, shuffle them and use `communicate in parallel` on the shuffled list.

linera-core/src/client/mod.rs
bart-linera: Kind of, but there is also some stuff that uses external objects in this computation (like, we update the validator trackers based on our local chain heights).

linera-core/src/client/mod.rs
bart-linera: Those aren't really iterations, though - more like parallel tasks, only some of them delayed. So yes, we clone the same list of remaining blocks to be downloaded into each task, and then filter using the validator the task is spawned for. Then whichever task returns first will result in the blocks that will be filtered out of the `remote_heights`, that will then get passed here on the next pass through the `while` loop (if there is a next pass - ideally, there shouldn't be, because the validator that responds should give us all the blocks we need).

linera-core/src/client/mod.rs
bart-linera: No, because if no validators return any results, we land in the `Err` branch of the `match communicate_in_parallel` and return.

linera-core/src/client/mod.rs
bart-linera: Not sure why I did that :sweat_smile:

linera-core/src/client/mod.rs
afck: Still, it seems there is no reason for not cloning `remote_heights` in every task separately, and filtering out only the ones that _that_ validator doesn't have?

linera-core/src/client/mod.rs
afck: But could they give us a height nobody has yet, and then repeatedly deliver an invalid (wrongly signed) certificate for that height? Then we would only end up in the `Err` branch below, with the log. Wouldn't we just keep trying to download that height?

linera-core/src/client/mod.rs
bart-linera: Oh, I see what you mean now. Yes, that's what I intended to do, actually - but you're right, it doesn't seem like this is what I am doing :sweat_smile: I'll fix it :+1:

linera-core/src/client/mod.rs
bart-linera: Hmm, good point. We'd have to filter out the wrong certificates at the closure level to avoid that scenario, and return an error if no correct certificates remain. Do you think that would be a reasonable approach?

afck: Yes… but keep in mind that one wrong certificate at height 5 doesn't mean there aren't any right certificates at height 5.

linera-core/src/client/received_log.rs
```diff
@@ -0,0 +1,114 @@
+// Copyright (c) Zefchain Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+use std::collections::{BTreeMap, BTreeSet};
+
+use linera_base::{crypto::ValidatorPublicKey, data_types::BlockHeight, identifiers::ChainId};
+use linera_chain::data_types::ChainAndHeight;
+
+/// Struct keeping track of the blocks sending messages to a particular chain.
```
deuszx: I think the comment and the struct are incomplete. It says that it's tracking which blocks sent a message to a particular chain – but it doesn't mention blocks on which sender chain. Is the key `ChainId` a sender or a receiving chain?

linera-core/src/client/received_log.rs
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    /// Returns the total number of certificates recorded in this log.
+    pub(super) fn num_certs(&self) -> usize {
+        self.0.values().map(|heights| heights.len()).sum()
+    }
+
+    /// Splits this `ReceivedLogs` into batches of size `batch_size`. Batches are sorted
+    /// by chain ID and height.
+    pub(super) fn into_batches(self, batch_size: usize) -> impl Iterator<Item = ReceivedLogs> {
```
deuszx: Can you write a couple unit tests asserting that the resulting batched iterator behaves as expected?

linera-core/src/client/validator_trackers.rs
```diff
@@ -0,0 +1,94 @@
...
+/// Manages a "tracker"  of a single validator.
+/// The received log is the list of chains and heights of blocks sending messages to a
+/// particular receiver chain. The tracker is the index of the first entry in that log
+/// that corresponds to a block that hasn't been processed yet.
+/// In order to keep the tracker value up to date, we keep the part of the log starting
+/// with the first entry corresponding to a not-yet-processed block, and a record of which
+/// certificates we have already downloaded. Whenever we download a certificate corresponding to
+/// the first block in the log, we increase the tracker and pop the blocks off the log,
+/// until we hit one we haven't downloaded yet.
+pub(super) struct ValidatorTracker {
```
deuszx: Can you add the receiver's chain id into the struct? You may not need it but I think it's useful to _complete_ the struct. Otherwise, on its own, this struct is indistinguishable for another chain.

linera-core/src/client/mod.rs
bart-linera: I thought this wouldn't be a problem, because then we wouldn't filter it out and we'd get to it in the next iteration... but if this is the _only_ certificate left, and we receive the wrong one first, we'll bail :weary: I'm not sure how to solve it - I guess we'd have to have some maximum number of retries for a single certificate? So that if we receive a wrong certificate, we'll let other validators send us another one, but we won't get stuck receiving a single wrong certificate, either...

linera-core/src/client/validator_trackers.rs
bart-linera: Hmm, TBH, I don't see a point, if it's not going to be used anywhere. Like, which receiving chain the struct is for is defined by the usage, similar to, say, block height, which is also indistinguishable for different chains.

linera-core/src/client/received_log.rs
```diff
@@ -0,0 +1,115 @@
+// Copyright (c) Zefchain Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+use std::collections::{BTreeMap, BTreeSet};
+
+use linera_base::{crypto::ValidatorPublicKey, data_types::BlockHeight, identifiers::ChainId};
+use linera_chain::data_types::ChainAndHeight;
+
+/// Struct keeping track of the blocks sending messages to some chain, from chains identified by
```
deuszx, line 9: "to some chain" – which chain?

linera-core/src/client/mod.rs
afck: Ideally if any validator sends us any wrong certificate, we should remove that validator entirely. And we should remove all heights that we know about from only faulty validators.

linera-core/src/client/received_log.rs
```diff
@@ -0,0 +1,115 @@
...
+/// the keys in the map.
+pub(super) struct ReceivedLogs(
+    BTreeMap<ChainId, BTreeMap<BlockHeight, BTreeSet<ValidatorPublicKey>>>,
+);
+
+impl ReceivedLogs {
+    /// Creates a new instance of `ReceivedLogs`.
+    pub(super) fn new() -> Self {
```
deuszx: This doens't have to be `pub(super)`, it's used only down below in the `fold` as the starting point.

Maybe remove the method entirely and inline the `ReceivedLogs(BTreeMap::new())` there?

linera-core/src/client/received_log.rs
```diff
@@ -0,0 +1,115 @@
...
+                    .or_default()
+                    .insert(validator);
+                acc
+            })
+    }
+}
+
+/// Iterator adapter lazily yielding batches of size `self.batch_size` from
+/// `self.iterator`.
+pub(super) struct LazyBatch<I: Iterator<Item = (ChainAndHeight, ValidatorPublicKey)>> {
```
deuszx: `LazyBatch` is local type as well – doesn't need `pub(super)`.

linera-core/src/client/validator_trackers.rs
```diff
@@ -0,0 +1,94 @@
...
+    current_tracker_value: u64,
+    to_be_downloaded: VecDeque<ChainAndHeight>,
+    already_downloaded: BTreeSet<ChainAndHeight>,
+}
+
+impl ValidatorTracker {
+    /// Creates a new `ValidatorTracker`.
+    pub(super) fn new(tracker: u64, validator_log: Vec<ChainAndHeight>) -> Self {
```
deuszx: Doesn't need `pub(super)`.

deuszx: None of the methods on `ValidatorTracker` (singular) need it

linera-core/src/client/validator_trackers.rs
deuszx: This is local type – doesn't need `pub(super)`.

linera-core/src/client/received_log.rs
bart-linera: The chain for which the struct is constructed. Kind of like "a block height is the height of a block in a chain". I can elaborate on that in the comment, but I am a bit concerned about potentially making it too verbose.

linera-core/src/client/validator_trackers.rs
bart-linera: Fixed in https://github.com/linera-io/linera-protocol/pull/4708/commits/303bf097719a7271cb62680d8df6e168b80afc7e

linera-core/src/client/validator_trackers.rs
bart-linera: Fixed in https://github.com/linera-io/linera-protocol/pull/4708/commits/303bf097719a7271cb62680d8df6e168b80afc7e

linera-core/src/client/received_log.rs
bart-linera: Fixed in https://github.com/linera-io/linera-protocol/pull/4708/commits/303bf097719a7271cb62680d8df6e168b80afc7e

linera-core/src/client/received_log.rs
bart-linera: Fixed in https://github.com/linera-io/linera-protocol/pull/4708/commits/303bf097719a7271cb62680d8df6e168b80afc7e

linera-core/src/client/received_log.rs
bart-linera: Added a test in https://github.com/linera-io/linera-protocol/pull/4708/commits/dec776c5c5f73939fd30e5a2e0199b13c873a8e8 - let me know if you think I should cover some more test cases.

linera-core/src/client/mod.rs
bart-linera: Hopefully https://github.com/linera-io/linera-protocol/pull/4708/commits/6929b65045233166eb3afb0b80a69aac63fe1f29 addresses all the concerns :sweat_smile:

linera-core/src/client/mod.rs
bart-linera: Should be addressed in https://github.com/linera-io/linera-protocol/pull/4708/commits/ba759ba7552415fc56ed4decb7a863489f169af4

linera-core/src/client/received_log.rs
deuszx: This is testing low-level implementation details but not how you're using the API.

deuszx: You should test that a `ReceivedLogs::into_batches(batch_size)` returns expected iterator. There should be no usage of `LazyBatch` in the test – that's an implementation detail.

linera-core/src/client/mod.rs
```diff
@@ -4271,6 +4399,42 @@ impl<Env: Environment> ChainClient<Env> {
...
+            async move {
+                linera_base::time::timer::sleep(timeout * i * i).await;
+                fun(node).await.map_err(|err| (remote_node.public_key, err))
+            }
+        })
+        .collect::<FuturesUnordered<_>>();
+    let mut errors = vec![];
+    while let Some(maybe_result) = stream.next().await {
+        match maybe_result {
+            Ok(result) => return Ok(result),
```
deuszx, line 4422: Will it cancel the rest of the spawned futures?

bart-linera: I would expect so, since they will just be dropped, and they aren't spawned on separate threads or anything like that, so as far as I understand, nothing will be executing them anymore.

linera-core/src/client/mod.rs
```diff
@@ -2261,73 +2162,300 @@ impl<Env: Environment> ChainClient<Env> {
...
 
-        Ok(has_more)
+        debug!("download_sender_certificates: finished processing other_sender_chains");
+
+        Ok(validator_trackers)
+    }
+
+    /// Compares validators' received logs of sender chains with local node information and returns
+    /// a per-chain list of block heights that sent us messages we didn't see yet.
+    async fn get_heights_to_download(
```
deuszx: IMHO this could be a method on `ValidatorTrackers` where `remote_heights` and `local_next_heights` are arguments to it: it updates `validator_trackers` and returns (remote) heights we haven't downloaded yet. 

Although, I'm not sure why we need to return it - I thought `ValidatorTracker.to_be_downloaded` is tracking it already.

linera-core/src/client/received_log.rs
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    /// Returns the total number of certificates recorded in this log.
+    pub(super) fn num_certs(&self) -> usize {
+        self.0.values().map(|heights| heights.len()).sum()
+    }
+
+    /// Splits this `ReceivedLogs` into batches of size `batch_size`. Batches are sorted
+    /// by chain ID and height.
+    pub(super) fn into_batches(self, batch_size: usize) -> impl Iterator<Item = ReceivedLogs> {
```
afck: Alternatively, we could just return a non-batched iterator, and at the call site use `Itertools::chunks` or `Itertools::next_array`, until `Iterator::next_chunk` becomes stable.

linera-core/src/client/mod.rs
afck: I agree, should be something like `communicate_concurrently`.

linera-core/src/client/mod.rs
afck: Good catch! The problem wasn't introduced in this PR, of course, but let's add that to the documentation of this function, and then double-check at the call sites that this is fine.

linera-core/src/client/mod.rs
```diff
@@ -2261,73 +2162,300 @@ impl<Env: Environment> ChainClient<Env> {
...
+    ) {
+        let stream = FuturesUnordered::from_iter(other_sender_chains.into_iter().map(|chain_id| {
+            let local_node = self.client.local_node.clone();
+            async move {
+                if let Err(error) = match local_node
+                    .retry_pending_cross_chain_requests(chain_id)
+                    .await
+                {
+                    Ok(()) => Ok(()),
+                    Err(LocalNodeError::BlobsNotFound(blob_ids)) => {
```
afck, line 2422: I think I only now understand why we need this:

We may have downloaded a few blocks from a sender chain, but for that we didn't need the chain description blob. Now we want to retry the outgoing messages, and we put the request for doing that into the chain info request, and for handling a chain info request, the node needs the chain description blob.

We should (not in this PR!) add a way to tell the local node to wait for the outgoing messages without needing the chain description. If this whole retrying thing is even needed at all, in the local node.

linera-core/src/client/mod.rs
afck: And a function called `get_` probably shouldn't mutate anything.

linera-core/src/client/mod.rs
```diff
@@ -825,149 +831,164 @@ impl<Env: Environment> Client<Env> {
...
+            let remote_heights_ref = &remote_heights;
+            nodes.shuffle(&mut rand::thread_rng());
+            let certificates = match communicate_in_parallel(
+                &nodes,
+                async move |remote_node| {
+                    let mut remote_heights = remote_heights_ref.clone();
+                    // No need trying to download certificates the validator didn't have in their
+                    // log - we'll retry downloading the remaining ones next time we loop.
+                    remote_heights.retain(|height| {
+                        received_log.validator_has_block(
```
afck, line 863: (It still feels like something could be simplified: First we got them from the received log, and now we filter them using the received log. Not a blocker, but maybe something to look into later.)

linera-core/src/client/mod.rs
```diff
@@ -825,149 +831,164 @@ impl<Env: Environment> Client<Env> {
...
+                        // It makes no sense to return `Ok(_)` if we aren't going to try downloading
+                        // anything from the validator - let the function try the other validators
+                        Err(())
+                    } else {
+                        let mut certificates = remote_node
+                            .download_certificates_by_heights(sender_chain_id, remote_heights)
+                            .await
+                            .map_err(|_| ())?;
+                        certificates.retain(|cert| {
+                            Self::check_certificate(max_epoch, committees_ref, cert)
```
afck: If any `check_certificate` call returns an _error_ (invalid signatures!) we should immediately break and return an error here: that validator is broken.

The filtering should only be applied based on which committees we might be missing.

linera-core/src/client/mod.rs
```diff
@@ -825,149 +831,164 @@ impl<Env: Environment> Client<Env> {
...
+                        received_log.validator_has_block(
+                            &remote_node.public_key,
+                            sender_chain_id,
+                            *height,
+                        )
+                    });
+                    if remote_heights.is_empty() {
+                        // It makes no sense to return `Ok(_)` if we aren't going to try downloading
+                        // anything from the validator - let the function try the other validators
+                        Err(())
```
afck: You could `return Err(())` here and reduce the indentation below.

linera-core/src/client/mod.rs
afck: Better, but there's still a regression with this PR: Previously, we would stop once we're done syncing from the fastest quorum of validators. Now, we will continue until the slowest non-faulty validator has given us everything they have.

An attacker could make `linera sync` arbitrarily slow for us:
* Create lots of chains, and send lots of messages to our chain from a single block on each of them.
* Send the confirmed certificates for those blocks only to a single validator that the attacker controls.
* Report the correct `received_log` to us, with all those messages. No other validator will know about them!
* Veeery slowly send us the blocks we're asking for, just below our timeout.

I don't think that's a blocker for now, but at least we should create an issue for it.

It's also not a problem for the node service, where this is running in the background anyway, and may even be the preferred behavior!

linera-core/src/client/mod.rs
```diff
@@ -825,149 +831,164 @@ impl<Env: Environment> Client<Env> {
...
+                let hash = certificate.hash();
+                let chain_id = certificate.block().header.chain_id;
+                let height = certificate.block().header.height;
+                // We checked the certificates right after downloading them.
+                let mode = ReceiveCertificateMode::AlreadyChecked;
+                if let Err(error) = self
+                    .receive_sender_certificate(certificate, mode, None)
+                    .await
+                {
+                    error!(%error, %hash, "Received invalid certificate");
```
afck: Could a faulty validator trigger this by refusing to supply a block published by that certificate? (See the scenario above: It's not guaranteed that any honest validator has this certificate.)

linera-core/src/client/mod.rs
```diff
@@ -825,149 +831,164 @@ impl<Env: Environment> Client<Env> {
...
+                } else {
+                    processed_correctly.insert(height);
+                    if let Err(err) = sender.send(ChainAndHeight { chain_id, height }) {
+                        error!(
+                            %chain_id,
+                            %height,
+                            error = %err,
```
afck: (Or just call the variable `error`?)

```suggestion
                            %error,
```

linera-core/src/client/mod.rs
```diff
@@ -825,149 +831,164 @@ impl<Env: Environment> Client<Env> {
...
+                            "failed to send chain and height over the channel",
+                        );
+                    }
                 }
             }
+
+            remote_heights.retain(|height| !processed_correctly.contains(height));
```
afck: What about the certificates we filtered out because we don't have the committee that signed them. Will we try downloading them again in the next iteration?

linera-core/src/client/mod.rs
```diff
@@ -2261,73 +2162,300 @@ impl<Env: Environment> ChainClient<Env> {
...
-                // to synchronize received certificates.
-                return Ok(false);
+        trace!(
+            received_log_batches_len = %received_log_batches.lock().unwrap().len(),
+            "got result from communicate_with_quorum"
+        );
+
+        if let Err(e) = result {
+            warn!(
+                error = %e,
```
afck: (Or just `%error`?)

linera-core/src/client/mod.rs
```diff
@@ -2261,73 +2162,300 @@ impl<Env: Environment> ChainClient<Env> {
...
-                // The chain is visibly not active (yet or any more) so there is no need
-                // to synchronize received certificates.
-                return Ok(false);
+        trace!(
+            received_log_batches_len = %received_log_batches.lock().unwrap().len(),
+            "got result from communicate_with_quorum"
+        );
+
+        if let Err(e) = result {
+            warn!(
```
afck: That one could actually be error-level, could it? We can still continue processing what we've got, but it makes sense to log an error if we can't reach a quorum.

linera-core/src/client/mod.rs
```diff
@@ -2261,73 +2162,300 @@ impl<Env: Environment> ChainClient<Env> {
...
+            .local_node
+            .chain_state_view(chain_id)
+            .await?
+            .received_certificate_trackers
+            .get()
+            .clone();
+
+        trace!("find_received_certificates: read trackers");
+
+        let received_log_batches = Arc::new(std::sync::Mutex::new(Vec::new()));
```
afck, line 2201: We could also modify `communicate_with_quorum` to return all the results it has so far (`Ok`s and `Err`s). Then we wouldn't need a mutex here.

linera-core/src/client/mod.rs
```diff
@@ -2261,73 +2162,300 @@ impl<Env: Environment> ChainClient<Env> {
...
+        }
+
+        let received_logs: Vec<_> = {
+            let mut received_log_batches = received_log_batches.lock().unwrap();
+            std::mem::take(received_log_batches.as_mut())
+        };
+
+        debug!(
+            received_logs_len=%received_logs.len(),
+            received_logs_total=%received_logs.iter().map(|x| x.1.len()).sum::<usize>(),
```
```suggestion
            received_logs_len = %received_logs.len(),
            received_logs_total = %received_logs.iter().map(|x| x.1.len()).sum::<usize>(),
```

linera-core/src/client/mod.rs
```diff
@@ -2261,73 +2162,300 @@ impl<Env: Environment> ChainClient<Env> {
...
+            "find_received_certificates: received_logs length"
```
afck: Maybe:

```suggestion
            "Collected received logs"
```

linera-core/src/client/mod.rs
afck: For now, it could also be an `mpsc` channel, then we wouldn't need any locking.

linera-core/src/client/mod.rs
```diff
@@ -2261,73 +2162,300 @@ impl<Env: Environment> ChainClient<Env> {
...
+        );
+
+        let (received_logs, mut validator_trackers) = {
+            (
+                ReceivedLogs::from_received_result(received_logs.clone()),
+                ValidatorTrackers::new(received_logs, &trackers),
+            )
+        };
+
+        for received_log in received_logs.into_batches(CHAIN_INFO_MAX_RECEIVED_LOG_ENTRIES) {
```
afck: We wanted to avoid constants, and make everything configurable via CLI and/or env variables.

linera-core/src/client/mod.rs
```diff
@@ -2261,73 +2162,300 @@ impl<Env: Environment> ChainClient<Env> {
...
+            validator_trackers = self
+                .download_sender_certificates(
```
afck: Let's call this `receive_` instead of `download_`, since it also processes them? (See naming discussion on Slack.)

linera-core/src/client/mod.rs
```diff
@@ -2261,73 +2162,300 @@ impl<Env: Environment> ChainClient<Env> {
...
-            Err(error) => {
-                return Err(error.into());
+        }
+
+        info!("find_received_certificates finished");
+
+        Ok(())
+    }
+
+    /// Downloads the certificates for blocks sending messages to this chain that we are
```
```suggestion
    /// Downloads and processes the certificates for blocks sending messages to this chain that we are
```

Or `processes or preprocesses`? Or `(pre)processes`?

linera-core/src/client/mod.rs
```diff
@@ -2261,73 +2162,300 @@ impl<Env: Environment> ChainClient<Env> {
...
+
+        debug!(
+            remaining_total_certificates = %remote_heights.values().map(|h| h.len()).sum::<usize>(),
+            "download_sender_certificates: computed remote_heights"
+        );
+
+        let mut other_sender_chains = Vec::new();
+        let (sender, mut receiver) = mpsc::unbounded_channel::<ChainAndHeight>();
+
+        let cert_futures =
```
afck: So these can also get cancelled. It's probably fine, but we need to make sure they are cancel-safe.

linera-core/src/client/mod.rs
```diff
@@ -2261,73 +2162,300 @@ impl<Env: Environment> ChainClient<Env> {
...
-            .collect();
+        drop(sender);
+
+        let validator_trackers = update_trackers.await.map_err(|_| {
+            ChainClientError::InternalError("could not join the update_trackers task")
+        })?;
 
-        self.receive_certificates_from_validators(received_certificate_batches)
+        debug!(
+            num_other_chains=%other_sender_chains.len(),
```
```suggestion
            num_other_chains = %other_sender_chains.len(),
```

linera-core/src/client/mod.rs
bart-linera: It's all kind of the same data, but organized differently. `ReceivedLogs` is organized per-chain: for each chain, we have the set of heights to download, based on the logs we received from the validators. `ValidatorTrackers` contains those logs split by-validator, and in the order in which the validators have them, so that we know how much of each validator's log we have processed. After we read local heights, we can use it to get filtered lists of heights per chain, so that we don't re-download stuff, and at the same time use it to update the trackers.

So `ValidatorTrackers` kind of has that information, but not organized in an easily accessible way. I guess I can make this a method on the trackers, though.

linera-core/src/client/mod.rs
bart-linera: You mean a blob? I guess :weary: Because of course just because everyone signed the certificate, doesn't mean that they actually have the full certificate, because it's up to the client to send it to them :weary: I don't have a good idea how to solve it right now... Maybe I'll just decrease the log level to warn for now, because this is just a thing that can happen, I guess.

linera-core/src/client/mod.rs
bart-linera: I think so :grimacing:

linera-core/src/client/validator_trackers.rs
```diff
@@ -91,4 +91,33 @@ impl ValidatorTrackers {
...
+
+    /// Compares validators' received logs of sender chains with local node information and returns
+    /// a per-chain list of block heights that sent us messages we didn't see yet. Updates
+    /// the trackers accordingly.
+    pub(super) fn filter_heights_to_download_and_update_trackers(
+        &mut self,
+        mut remote_heights: BTreeMap<ChainId, BTreeSet<BlockHeight>>,
+        local_next_heights: BTreeMap<ChainId, BlockHeight>,
+    ) -> BTreeMap<ChainId, BTreeSet<BlockHeight>> {
+        for (sender_chain_id, remote_heights) in remote_heights.iter_mut() {
```
afck: Or `&mut remote_heights`. (I think that will become a Clippy lint at some point. Or maybe it's a "pedantic" one.)

linera-core/src/client/received_log.rs
```diff
@@ -0,0 +1,313 @@
...
+                    .or_default()
+                    .insert(validator);
+                acc
+            },
+        )
+    }
+}
+
+/// Iterator adapter lazily yielding batches of size `self.batch_size` from
+/// `self.iterator`.
```
deuszx: Comment seems outdated – there's no `self.iterator` anymore.

linera-core/src/client/received_log.rs
```diff
@@ -0,0 +1,313 @@
...
+        let keys = heights.keys().copied().collect();
+        Self {
+            keys,
+            heights,
+            batch_size,
+            current_chain: 0,
+        }
+    }
+
+    fn next_chain_and_height(&mut self) -> Option<(ChainAndHeight, BTreeSet<ValidatorPublicKey>)> {
```
deuszx, line 138: Wouldn't it be simpler if we sorted the `heights` in the constructor and stored them in a `Vec<_>` and then, when the next batch is needed, we take from it?

deuszx: In the `impl Iterator` for `BatchingHelper` you are coping the `validator` for every `ChainAndHeight` value

linera-core/src/client/received_log.rs
```diff
@@ -0,0 +1,313 @@
...
+            let chain_a = ChainId(CryptoHash::test_hash("chain_a"));
+            let chain_b = ChainId(CryptoHash::test_hash("chain_b"));
+            if chain_a < chain_b {
+                (chain_a, chain_b)
+            } else {
+                (chain_b, chain_a)
+            }
+        };
+        let validator = ValidatorKeypair::generate().public_key;
+        let test_log = ReceivedLogs::from_iterator(
```
deuszx: `from_iterator` is an implementation details. The API/constructor for `ReceivedLogs` is `from_received_result`

linera-core/src/client/received_log.rs
```diff
@@ -0,0 +1,313 @@
...
+            // make sure that chain1 is lexicographically earlier than chain2
+            let chain_a = ChainId(CryptoHash::test_hash("chain_a"));
+            let chain_b = ChainId(CryptoHash::test_hash("chain_b"));
+            if chain_a < chain_b {
+                (chain_a, chain_b)
+            } else {
+                (chain_b, chain_a)
+            }
+        };
+        let validator = ValidatorKeypair::generate().public_key;
```
deuszx, line 224: We should test with more than one validator as well.

linera-core/src/client/received_log.rs
```diff
@@ -0,0 +1,313 @@
...
+            .collect::<Vec<_>>();
+
+        assert_eq!(
+            chains_heights,
+            vec![
+                vec![(chain1, 1), (chain2, 1)],
+                vec![(chain1, 2), (chain2, 2)],
+                vec![(chain1, 3), (chain2, 3)],
+                vec![(chain1, 4), (chain2, 4)],
+                vec![(chain1, 5)]
```
deuszx, line 277: Wouldn't it be better to return `chain1` first, in its entirety, and then move to `chain2`?

linera-core/src/client/received_log.rs
bart-linera: Not sure if that makes a big difference, but it's definitely an option :+1:

linera-core/src/client/validator_trackers.rs
```diff
@@ -0,0 +1,123 @@
...
+    to_be_downloaded: VecDeque<ChainAndHeight>,
+    already_downloaded: BTreeSet<ChainAndHeight>,
+}
+
+impl ValidatorTracker {
+    /// Creates a new `ValidatorTracker`.
+    fn new(tracker: u64, validator_log: Vec<ChainAndHeight>) -> Self {
+        Self {
+            current_tracker_value: tracker,
+            to_be_downloaded: validator_log.into_iter().collect(),
```
deuszx: We should make sure this is sorted in ascending order.

linera-core/src/client/validator_trackers.rs
```diff
@@ -0,0 +1,123 @@
...
+/// that corresponds to a block that hasn't been processed yet.
+/// In order to keep the tracker value up to date, we keep the part of the log starting
+/// with the first entry corresponding to a not-yet-processed block, and a record of which
+/// certificates we have already downloaded. Whenever we download a certificate corresponding to
+/// the first block in the log, we increase the tracker and pop the blocks off the log,
+/// until we hit one we haven't downloaded yet.
+struct ValidatorTracker {
+    current_tracker_value: u64,
+    to_be_downloaded: VecDeque<ChainAndHeight>,
+    already_downloaded: BTreeSet<ChainAndHeight>,
```
deuszx: I'd make it `BTreeMap<Chain, BlockHeight>` and store the highest block downloaded. Then the tests for "is already downloaded" are more flexible and faster.

linera-core/src/client/validator_trackers.rs
```diff
@@ -0,0 +1,123 @@
...
+        {
+            let first_cert = self.to_be_downloaded.pop_front().unwrap();
+            self.already_downloaded.remove(&first_cert);
+            self.current_tracker_value += 1;
+        }
+    }
+}
+
+/// Keeps multiple `ValidatorTracker`s for multiple validators.
+pub(super) struct ValidatorTrackers(BTreeMap<ValidatorPublicKey, ValidatorTracker>);
```
deuszx, line 12: Since `ValidatorTracker` (singular) is never used outside of the file, this `ValidatorTrackers`, should be at the top of the file. In a sense `ValidatorTracker` is an implementation detail.

linera-core/src/client/received_log.rs
bart-linera: That's what the code used to do, but I deliberately changed it. That's because when we're actually processing a batch, we create one task per chain: so if we spread this across multiple chains, we're getting a boost from concurrent processing.

But there's one improvement I will make here: instead of taking just one block from each chain at a time, we could take some larger number, say 100 by default. Then we can also take advantage of downloading many certificates for a single chain at a time.

What I want to avoid is creating one task that attempts to download and process 20000 certificates at once, but having 20000 tasks, each downloading and processing 1 certificate is also suboptimal, I guess.

linera-core/src/client/validator_trackers.rs
bart-linera: This is being passed in the order received from the validator, and it's actually important for this order to be preserved, because the trackers in the ChainStateView are indices into the lists of entries in this order.

linera-core/src/client/received_log.rs
deuszx: I think the code would be definitely simpler and you wouldn't "waste" CPU cycles every time a new batch is requested.

linera-core/src/client/received_log.rs
deuszx: The issue with the current approach is that we will be spawning multiple chain actors which are subject to killing (TTL, cache size, etc.).

linera-core/src/client/validator_trackers.rs
bart-linera: I guess I was worried about the possibility of some certificates being downloaded out of order... but we should make sure that doesn't happen, anyway (and I think I actually introduced a place where it can happen - I should fix it).

linera-core/src/client/mod.rs
```diff
@@ -825,149 +831,173 @@ impl<Env: Environment> Client<Env> {
...
-                if remote_heights.is_empty() {
-                    // Our highest, locally executed block is higher than any block height
-                    // from the current batch. Skip this batch, but remember to wait for
-                    // the messages to be delivered to the inboxes.
-                    other_sender_chains.push(sender_chain_id);
-                    return None;
-                };
-                Some(async move {
+        sender_chain_id: ChainId,
+        nodes: &[RemoteNode<Env::ValidatorNode>],
```
deuszx, line 839: Since we call `self.admin_committees` we can construct the latest vaidators from `committees` field (and their `address` field). It will simplify the method signature and is probably the correct thing to do anyway.

linera-core/src/client/mod.rs
```diff
@@ -825,149 +831,173 @@ impl<Env: Environment> Client<Env> {
...
+        received_log: &ReceivedLogs,
+        mut remote_heights: Vec<BlockHeight>,
+        sender: mpsc::UnboundedSender<ChainAndHeight>,
+    ) {
+        // Verify the certificate before doing any expensive networking.
```
deuszx: Is this comment still valid?

linera-core/src/client/mod.rs
```diff
@@ -2261,73 +2176,276 @@ impl<Env: Environment> ChainClient<Env> {
...
+            received_log_batches_len = %received_log_batches.lock().unwrap().len(),
+            "got result from communicate_with_quorum"
+        );
+
+        if let Err(error) = result {
+            error!(
+                %error,
+                "Failed to synchronize received_logs from at least a quorum of validators",
+            );
+        }
```
deuszx, line 2229: Do we need the quorum for this though? We are updating our view of each of the validators' received log. It has nothing to do with finding a quorum for one specific thing.

linera-core/src/client/mod.rs
```diff
@@ -2261,73 +2176,276 @@ impl<Env: Environment> ChainClient<Env> {
...
+                let received_log_batches = Arc::clone(&received_log_batches);
                 Box::pin(async move {
-                    client
-                        .synchronize_received_certificates_batch_from_validator(
-                            chain_id,
-                            &remote_node,
-                        )
-                        .await
+                    let batch = client
+                        .get_received_log_from_validator(&trackers_clone, chain_id, &remote_node)
```
deuszx: `get_received_log_from_validator` needs the tracker to extract the received certificates count, we could pass the number here right away and simplify the function.

linera-core/src/client/mod.rs
```diff
@@ -2261,73 +2176,276 @@ impl<Env: Environment> ChainClient<Env> {
...
-        let ((), batches_with_progress) = match result {
-            Ok(result) => result,
-            Err(CommunicationError::Trusted(NodeError::InactiveChain(id))) if id == chain_id => {
-                // The chain is visibly not active (yet or any more) so there is no need
-                // to synchronize received certificates.
-                return Ok(false);
+        trace!(
+            received_log_batches_len = %received_log_batches.lock().unwrap().len(),
+            "got result from communicate_with_quorum"
+        );
```
deuszx: This and the `debug!` below are duplicates – let's remove one.

linera-core/src/client/mod.rs
```diff
@@ -825,149 +831,173 @@ impl<Env: Environment> Client<Env> {
...
+                    let mut certificates_with_check_results = vec![];
+                    for cert in certificates {
+                        if let Ok(check_result) =
+                            Self::check_certificate(max_epoch, committees_ref, &cert)
+                        {
+                            certificates_with_check_results
+                                .push((cert, check_result.into_result().is_ok()));
+                        } else {
+                            // Invalid signature - the validator is faulty
+                            return Err(());
```
deuszx, line 887: There are actually 3 ways in which the `check_certificate` can fail:
- cert is from the future
- cert is from a retired committee
- cert fails the signature check

So the first two are not necessarily malicious.

linera-core/src/client/mod.rs
```diff
@@ -2261,73 +2176,276 @@ impl<Env: Environment> ChainClient<Env> {
...
+                    nodes.clone(),
+                    cancellation_token.clone(),
+                )
+                .await?;
+
+            let new_trackers = validator_trackers.to_map();
+            trace!(
+                ?new_trackers,
+                "find_received_certificates: new tracker values"
+            );
```
deuszx: This isn't really purely _new_ tracker values – you start with non-empty state of `ValidatorTrackers`, you process the (first) batch, overwrite `validator_trackers` with the result of `filter_heights_to_download_and_update_trackers` and then print it. `validator_trackers.to_map` would return non-empty result even without processing any of the batches.

linera-core/src/client/validator_trackers.rs
```diff
@@ -0,0 +1,123 @@
...
+            .map(|(validator, tracker)| (*validator, tracker.current_tracker_value))
+            .collect()
+    }
+
+    /// Compares validators' received logs of sender chains with local node information and returns
+    /// a per-chain list of block heights that sent us messages we didn't see yet. Updates
+    /// the trackers accordingly.
+    pub(super) fn filter_heights_to_download_and_update_trackers(
+        &mut self,
+        mut remote_heights: BTreeMap<ChainId, BTreeSet<BlockHeight>>,
```
deuszx: I believe that if you change this to 
```rust
remote_heights:  &mut BTreeMap<ChainId, BTreeSet<BlockHeight>>,
```
then this won't need to return anything, the `remote_heights` from the caller will be mutated in place, and no cloning will be necessary.

I also think this will make the function clearer to undertand – conceptually it would `remote_heights` that we need to download using `local_next_heights` (and update the `ValidatorsTracker` in the meantime).

linera-core/src/client/mod.rs
```diff
@@ -2261,73 +2176,276 @@ impl<Env: Environment> ChainClient<Env> {
...
+
+        let mut other_sender_chains = Vec::new();
+        let (sender, mut receiver) = mpsc::unbounded_channel::<ChainAndHeight>();
+
+        let cert_futures =
+            remote_heights
+                .into_iter()
+                .filter_map(|(sender_chain_id, remote_heights)| {
+                    let remote_heights = remote_heights.into_iter().collect::<Vec<_>>();
+                    if remote_heights.is_empty() {
```
deuszx: You can make this check before line 2332.

linera-core/src/client/mod.rs
bart-linera: We could, but I wanted to skip calling `make_nodes` in each per-chain task, when there can be lots of them, and the list of the nodes would have been created just moments before, anyway.

linera-core/src/client/mod.rs
bart-linera: We don't, which is why we don't fail here, but a situation in which we can't get a quorum is an abnormal one.

linera-core/src/client/mod.rs
deuszx: I'll be removing it anyway in my PR.

linera-core/src/client/mod.rs
deuszx: My point is that maybe `communicate_in_parallel` would be a better choice. `communicate_with_quorum` is a "heavier" method.

linera-core/src/client/mod.rs
bart-linera: Yes, and they are handled by the other branch of the `if`. `check_certificate` returns an `Ok` if the signature is valid, but then the result inside could indicate a missing committee. That's covered by the `into_result().is_ok()` on that value, but in this case we do not treat the validator as malicious.

linera-core/src/client/mod.rs
bart-linera: So maybe not `new`, but `updated`?

linera-core/src/client/mod.rs
deuszx: Ah, right.. I totally missed that the other "errors" are returned as Ok(_) part.

linera-core/src/client/mod.rs
bart-linera: The other one is not a good choice, because it starts subsequent queries with a delay, and returns the first one that gets a response - but the idea here is to collect the logs from as many validators as possible and combine them (because different validators can have different entries in the logs).

deuszx: Right, and I think this is the problem with `communicate_in_parallel` – it's not really parallel if different tasks are starting with an increasing delay 😅

bart-linera: True :sweat_smile: The name has been changed already, BTW (to `communicate_concurrently`), but it still doesn't really reflect the delayed nature of subsequent queries :sweat_smile:

linera-core/src/client/validator_trackers.rs
```diff
@@ -0,0 +1,165 @@
...
+                    // validators' trackers accordingly
+                    self.downloaded_cert(ChainAndHeight {
+                        chain_id: *sender_chain_id,
+                        height: *height,
+                    });
+                }
+            }
+            remote_heights.retain(|h| *h >= local_next);
+        }
+    }
```
deuszx: I have locally changed this to:
```rust
pub(super) fn filter_out_already_known(
        &mut self,
        remote_heights: &mut ReceivedLogs,
        local_next_heights: BTreeMap<ChainId, BlockHeight>,
    ) {
        for (sender_chain_id, local_highest) in &local_next_heights {
            if let Some(remote_heights) = remote_heights.get_chain_mut(sender_chain_id) {
                remote_heights.retain(|height, _| {
                    if height < local_highest {
                        // we consider all of the heights below our local next height
                        // to have been already downloaded, so we will increase the
                        // validators' trackers accordingly
                        self.downloaded_cert(ChainAndHeight {
                            chain_id: *sender_chain_id,
                            height: *height,
                        });
                        false
                    } else {
                        true
                    }
                });
            }
        }
    }
```

I think it's simpler, not sure if you agree.

This has the added benefit that you don't need to clone the `ReceivedLogs`' inner map before passing it in.

linera-core/src/client/validator_trackers.rs
```diff
@@ -0,0 +1,165 @@
...
+            height: 1.into(),
+        });
+        assert_eq!(tracker.current_tracker_value, 1);
+
+        tracker.downloaded_cert(ChainAndHeight {
+            chain_id: chain2,
+            height: 0.into(),
+        });
+        assert_eq!(tracker.current_tracker_value, 3);
+    }
```
deuszx, line 171: We should also test `filter_heights_to_download_and_update_trackers`. Looks like it will be fairly simple to add.

linera-core/src/client/received_log.rs
```diff
@@ -0,0 +1,511 @@
...
+        }
+
+        maybe_heights
+            .and_then(|heights| heights.pop_first())
+            .map(|(height, validators)| (ChainAndHeight { chain_id, height }, validators))
+    }
+}
+
+impl Iterator for BatchingHelper {
+    type Item = ReceivedLogs;
```
deuszx, line 185: TBH I find it weird that an iterator for `ReceivedLogs` returns batches of type `ReceivedLogs` which in turn return an iterator with `Item = ReceivedLogs`, etc...

bart-linera: I think there aren't that many layers of indirection in there :sweat_smile:

deuszx: I meant that you can create the iterators "indeifintely"

PR #4712, opened 2025-10-01
Add support for chrome trace exporter

ndr-ds:
## Motivation

Right now for tracing information we can only visualize it using Tempo, which doesn't really work for local client runs.

## Proposal

Add support for `ChromeTraceExporter`, which generates a JSON file that can be ingested by the Perfetto UI for visualization.

## Test Plan

Ran the client locally with the `--chrome-trace-exporter`, got the JSON file and managed to see the tracing data in the Perfetto UI
https://ui.perfetto.dev/#!/viewer?local_cache_key=00000000-0000-0000-f5e0-334428275111

By default it will save the profile JSON in the current directory, but that can be changed using the `--otel-trace-file` option, or the `LINERA_OTEL_TRACE_FILE` env variable.

## Release Plan

### DIFF
--- a/linera-base/src/tracing.rs
+++ b/linera-base/src/tracing.rs
@@ -12,8 +12,6 @@ use std::{
 
 use is_terminal::IsTerminal as _;
 use tracing::Subscriber;
-#[cfg(all(not(target_arch = "wasm32"), feature = "tempo"))]
-use tracing_subscriber::filter::{filter_fn, FilterExt as _};
 use tracing_subscriber::{
     fmt::{
         self,
@@ -24,17 +22,51 @@ use tracing_subscriber::{
     layer::{Layer, SubscriberExt as _},
     registry::LookupSpan,
     util::SubscriberInitExt,
+    EnvFilter,
 };
-#[cfg(all(not(target_arch = "wasm32"), feature = "tempo"))]
-use {
-    opentelemetry::{global, trace::TracerProvider},
-    opentelemetry_otlp::{SpanExporter, WithExportConfig},
-    opentelemetry_sdk::{
-        trace::{self as sdktrace, SdkTracerProvider},
-        Resource,
-    },
-    tracing_opentelemetry::OpenTelemetryLayer,
+
+#[cfg(not(target_arch = "wasm32"))]
+pub use crate::tracing_otel::{
+    init_with_chrome_trace_exporter, init_with_opentelemetry, ChromeTraceGuard,
 };
+
+pub(crate) struct EnvConfig {
+    pub(crate) env_filter: EnvFilter,
+    span_events: FmtSpan,
+    format: Option<String>,
+    color_output: bool,
+    log_name: String,
+}
+
+impl EnvConfig {
+    pub(crate) fn stderr_layer<S>(&self) -> Box<dyn Layer<S> + Send + Sync>
+    where
+        S: Subscriber + for<'span> LookupSpan<'span>,
+    {
+        prepare_formatted_layer(
+            self.format.as_deref(),
+            fmt::layer()
+                .with_span_events(self.span_events.clone())
+                .with_writer(std::io::stderr)
+                .with_ansi(self.color_output),
+        )
+    }
+
+    pub(crate) fn maybe_log_file_layer<S>(&self) -> Option<Box<dyn Layer<S> + Send + Sync>>
+    where
+        S: Subscriber + for<'span> LookupSpan<'span>,
+    {
+        open_log_file(&self.log_name).map(|file_writer| {
+            prepare_formatted_layer(
+                self.format.as_deref(),
+                fmt::layer()
+                    .with_span_events(self.span_events.clone())
+                    .with_writer(Arc::new(file_writer))
+                    .with_ansi(false),
+            )
+        })
+    }
+}
 
 /// Initializes tracing in a standard way.
 ///
@@ -46,65 +78,19 @@ use {
 /// store log files. If it is set, a file named `log_name` with the `log` extension is
 /// created in the directory.
 pub fn init(log_name: &str) {
-    init_internal(log_name, false);
+    let config = get_env_config(log_name);
+    let maybe_log_file_layer = config.maybe_log_file_layer();
+    let stderr_layer = config.stderr_layer();
+
+    tracing_subscriber::registry()
+        .with(config.env_filter)
+        .with(maybe_log_file_layer)
+        .with(stderr_layer)
+        .init();
 }
 
-/// Initializes tracing with full OpenTelemetry support.
-///
-/// **IMPORTANT**: This function must be called from within a Tokio runtime context
-/// as it initializes OpenTelemetry background tasks for span batching and export.
-///
-/// This sets up complete tracing with OpenTelemetry integration, including the
-/// OpenTelemetry layer in the subscriber to export spans to Tempo.
-///
-/// ## Span Filtering for Performance
-///
-/// By default, spans created by `#[instrument]` are logged to console AND sent
-/// to OpenTelemetry. In order to not spam stderr, you can set a low level, and use the
-/// `telemetry_only` target:
-///
-/// ```rust
-/// use tracing::{instrument, Level};
-///
-/// // Always sent to telemetry; console output controlled by level
-/// #[instrument(level = "trace", target = "telemetry_only")]
-/// fn my_called_too_frequently_function() {
-///     // Will be sent to OpenTelemetry regardless of RUST_LOG level
-///     // Will only appear in console if RUST_LOG includes trace level
-/// }
-///
-/// // Higher level - more likely to appear in console
-/// #[instrument(level = "info", target = "telemetry_only")]
-/// fn my_important_function() {
-///     // Will be sent to OpenTelemetry regardless of RUST_LOG level
-///     // Will appear in console if RUST_LOG includes info level or higher
-/// }
-/// ```
-///
-/// **Key behaviors:**
-/// - If span level >= RUST_LOG level: span goes to BOTH telemetry AND console (regardless of target)
-/// - If span level < RUST_LOG level AND target = "telemetry_only": span goes to telemetry ONLY
-/// - If span level < RUST_LOG level AND target != "telemetry_only": span is filtered out completely
-/// - Default level for `telemetry_only` should be `trace` for minimal console noise
-/// - All explicit log calls (tracing::info!(), etc.) are always printed regardless of span filtering
-pub async fn init_with_opentelemetry(log_name: &str) {
-    #[cfg(feature = "tempo")]
-    {
-        init_internal(log_name, true);
-    }
-
-    #[cfg(not(feature = "tempo"))]
-    {
-        tracing::warn!(
-            "OpenTelemetry initialization requested but 'tempo' feature is not enabled. \
-             Initializing standard tracing without OpenTelemetry support."
-        );
-        init_internal(log_name, false);
-    }
-}
-
-fn init_internal(log_name: &str, with_opentelemetry: bool) {
-    let env_filter = tracing_subscriber::EnvFilter::builder()
+pub(crate) fn get_env_config(log_name: &str) -> EnvConfig {
+    let env_filter = EnvFilter::builder()
         .with_default_directive(tracing_subscriber::filter::LevelFilter::INFO.into())
         .from_env_lossy();
 
@@ -116,87 +102,12 @@ fn init_internal(log_name: &str, with_opentelemetry: bool) {
     let color_output =
         !std::env::var("NO_COLOR").is_ok_and(|x| !x.is_empty()) && std::io::stderr().is_terminal();
 
-    let stderr_layer = prepare_formatted_layer(
-        format.as_deref(),
-        fmt::layer()
-            .with_span_events(span_events.clone())
-            .with_writer(std::io::stderr)
-            .with_ansi(color_output),
-    );
-
-    let maybe_log_file_layer = open_log_file(log_name).map(|file_writer| {
-        prepare_formatted_layer(
-            format.as_deref(),
-            fmt::layer()
-                .with_span_events(span_events)
-                .with_writer(Arc::new(file_writer))
-                .with_ansi(false),
-        )
-    });
-
-    #[cfg(any(target_arch = "wasm32", not(feature = "tempo")))]
-    {
-        let _ = with_opentelemetry;
-        tracing_subscriber::registry()
-            .with(env_filter)
-            .with(maybe_log_file_layer)
-            .with(stderr_layer)
-            .init();
-    }
-
-    #[cfg(all(not(target_arch = "wasm32"), feature = "tempo"))]
-    {
-        if with_opentelemetry {
-            // Initialize OpenTelemetry within async context
-            let otlp_endpoint = std::env::var("OTEL_EXPORTER_OTLP_ENDPOINT")
-                .unwrap_or_else(|_| "http://tempo.tempo.svc.cluster.local:4317".to_string());
-
-            let exporter = SpanExporter::builder()
-                .with_tonic()
-                .with_endpoint(otlp_endpoint)
-                .build()
-                .expect("Failed to create OTLP exporter");
-
-            let resource = Resource::builder()
-                .with_service_name(log_name.to_string())
-                .build();
-
-            let tracer_provider = SdkTracerProvider::builder()
-                .with_resource(resource)
-                .with_batch_exporter(exporter)
-                .with_sampler(sdktrace::Sampler::AlwaysOn)
-                .build();
-
-            // Set the global tracer provider
-            global::set_tracer_provider(tracer_provider.clone());
-
-            let tracer = tracer_provider.tracer("linera");
-
-            let telemetry_only_filter =
-                filter_fn(|metadata| metadata.is_span() && metadata.target() == "telemetry_only");
-
-            let otel_env_filter = tracing_subscriber::EnvFilter::builder()
-                .with_default_directive(tracing_subscriber::filter::LevelFilter::INFO.into())
-                .from_env_lossy();
-
-            let opentelemetry_filter = otel_env_filter.or(telemetry_only_filter);
-
-            let opentelemetry_layer =
-                OpenTelemetryLayer::new(tracer).with_filter(opentelemetry_filter);
-
-            tracing_subscriber::registry()
-                .with(env_filter)
-                .with(maybe_log_file_layer)
-                .with(stderr_layer)
-                .with(opentelemetry_layer)
-                .init();
-        } else {
-            tracing_subscriber::registry()
-                .with(env_filter)
-                .with(maybe_log_file_layer)
-                .with(stderr_layer)
-                .init();
-        }
+    EnvConfig {
+        env_filter,
+        span_events,
+        format,
+        color_output,
+        log_name: log_name.to_string(),
     }
 }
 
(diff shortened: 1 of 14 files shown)


ndr-ds:
* **#4712** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4712?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4712?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

deuszx:
~~There's a crate for this apparently https://docs.rs/tracing-chrometrace/latest/tracing_chrometrace/~~

Ahh, so old that maybe outdated.

linera-service/src/cli/main.rs
```diff
@@ -1992,7 +1992,7 @@ async fn kill_all_processes(pids: &[u32]) {
 }
 
 fn should_init_opentelemetry(command: &ClientCommand) -> bool {
-    matches!(command, ClientCommand::Faucet { .. })
+    matches!(command, ClientCommand::Faucet { .. }) || env::var_os("OTEL_EXPORTER_TYPE").is_some()
```
ma2bd: Can we use proper CLI options using `clap` with the `env` option?

ma2bd: See e.g. `LINERA_CHAIN_WORKER_TTL_MS`

linera-base/src/chrome_trace_exporter.rs
```diff
@@ -0,0 +1,296 @@
...
+};
+
+use opentelemetry_sdk::{
+    error::OTelSdkError,
+    trace::{SpanData, SpanExporter},
+};
+use serde::Serialize;
+
+#[derive(Debug, Serialize)]
+struct ChromeTraceEvent {
```
deuszx: Is this format defined by some spec? If so, can you link it here as the comment to the struct?

linera-base/src/chrome_trace_exporter.rs
```diff
@@ -0,0 +1,296 @@
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    id: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    args: Option<serde_json::Value>,
+}
+
+struct ChromeTraceFile {
+    file: std::fs::File,
+    is_first_event: bool,
+}
```
deuszx: I think we should separate collecting the trace and writing it to the file. I suggest we first have a `ChromeTrace` struct that is jus a wrapper around vector of events. And then, have a function to write that trace to the file.

If you do that, it should be enough to derive serde on the struct to get the write/read work out of the box.

linera-base/src/chrome_trace_exporter.rs
```diff
@@ -0,0 +1,296 @@
...
+    fn drop(&mut self) {
+        let _ = self.finalize();
+    }
+}
+
+/// OpenTelemetry span exporter that writes spans to a Chrome trace JSON file.
+///
+/// The exported file can be opened in `chrome://tracing` or the Perfetto UI
+/// for visualization and analysis of distributed traces.
+pub struct ChromeTraceExporter {
```
deuszx: In general I think pub structs should be at the top of the file and private ones down below. Same goes for ordering functions: first pub, later private.

linera-base/src/chrome_trace_exporter.rs
```diff
@@ -0,0 +1,296 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_json_format() {
+        let path = PathBuf::from("/tmp/test_chrome_trace.json");
```
deuszx: Maybe use `tempfile` crate that we already have

linera-base/src/chrome_trace_exporter.rs
```diff
@@ -0,0 +1,296 @@
...
+
+                let args_value = if args.is_empty() {
+                    None
+                } else {
+                    Some(serde_json::Value::Object(args))
+                };
+
+                let name = span.name.to_string();
+                let cat = Some(span.instrumentation_scope.name().to_string());
+                let pid = 1;
```
deuszx: `std::process::id()` returns the actual PID.

linera-base/src/chrome_trace_exporter.rs
```diff
@@ -0,0 +1,296 @@
...
+                let tid = thread_id.unwrap_or(1);
```
deuszx: Does it have to be numerical? If it doesn't, and we can use `String`, then I think `n/a` will be better. Using the same fake number for everything can cause problems.

linera-base/src/tracing.rs
```diff
@@ -147,29 +163,59 @@ fn init_internal(log_name: &str, with_opentelemetry: bool) {
...
                 .with_service_name(log_name.to_string())
                 .build();
 
-            let tracer_provider = SdkTracerProvider::builder()
-                .with_resource(resource)
-                .with_batch_exporter(exporter)
-                .with_sampler(sdktrace::Sampler::AlwaysOn)
-                .build();
+            let tracer_provider = match exporter_type.as_str() {
+                "chrome_json" => {
```
deuszx: We shouldn't require `tempo` feature flag for this variant to work.

linera-base/src/tracing.rs
```diff
@@ -49,6 +53,18 @@ pub fn init(log_name: &str) {
     init_internal(log_name, false);
 }
 
+/// Shuts down the OpenTelemetry tracer provider.
+///
+/// This function should be called before the application exits to ensure all spans are flushed.
+pub fn shutdown_opentelemetry() {
+    #[cfg(all(not(target_arch = "wasm32"), feature = "tempo"))]
+    {
+        if let Some(provider) = TRACER_PROVIDER.get() {
```
deuszx: There's [`global::tracer_provider`](https://docs.rs/opentelemetry/0.31.0/opentelemetry/global/fn.tracer_provider.html) method that returns the current, global tracer provider. You can use that to shut it down and get rid of the static.

deuszx: Hmm, or not 🤔 it looks like you're working with `SdkTracerProvider` and not the `GlobalTracerProvider`.

linera-base/src/chrome_trace_exporter.rs
ndr-ds: Claude did this and I missed it 😓

linera-base/src/tracing.rs
ndr-ds: Unfortunately we need this `OnceLock`

linera-base/src/chrome_trace_exporter.rs
```diff
@@ -0,0 +1,369 @@
...
+    ts: u64,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    dur: Option<u64>,
+    pid: u32,
+    tid: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    id: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    args: Option<serde_json::Value>,
+}
```
deuszx: Those variable names are crazy :P

nit: we could use some useful name and use serde rename so the format is compatible with the spec.

linera-base/src/chrome_trace_exporter.rs
```diff
@@ -0,0 +1,369 @@
...
+
+/// A collection of Chrome trace events.
+///
+/// This struct wraps a vector of trace events and can be serialized to the
+/// Chrome Trace Event Format JSON structure.
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct ChromeTrace {
```
deuszx: Are you using it anywhere?

linera-client/src/client_options.rs
```diff
@@ -92,6 +92,13 @@ pub struct ClientContextOptions {
     #[arg(long, default_value = "10")]
     pub max_retries: u32,
 
+    /// OpenTelemetry exporter type for distributed tracing.
+    /// Valid values: "chrome_json", "otlp" (requires tempo feature and
+    /// will be used as the default).
+    /// Can also be set via OTEL_EXPORTER_TYPE environment variable.
+    #[arg(long, env = "OTEL_EXPORTER_TYPE")]
```
deuszx: We should be preppending all env vars with `LINERA_` and we've already started doing that in a couple of places.

linera-client/src/client_options.rs
```diff
@@ -92,6 +92,13 @@ pub struct ClientContextOptions {
     #[arg(long, default_value = "10")]
     pub max_retries: u32,
 
+    /// OpenTelemetry exporter type for distributed tracing.
+    /// Valid values: "chrome_json", "otlp" (requires tempo feature and
```
deuszx: For this I think you can use `value_names` from the `clap`.

linera-base/src/tracing.rs
```diff
@@ -134,46 +139,84 @@ fn init_internal(log_name: &str, with_opentelemetry: bool) {
...
                 .build();
 
-            let tracer_provider = SdkTracerProvider::builder()
-                .with_resource(resource)
-                .with_batch_exporter(exporter)
-                .with_sampler(sdktrace::Sampler::AlwaysOn)
-                .build();
+            let tracer_provider = match exporter_type.as_str() {
+                "chrome_json" | "chrome" => {
+                    let trace_file = std::env::var("OTEL_TRACES_FILE")
```
deuszx: Please use `LINERA_OTEL_TRACE_FILE` (note that I dropped the -s from traces since it's always a singular trace file IIUC).

linera-base/src/tracing.rs
```diff
@@ -134,46 +139,84 @@ fn init_internal(log_name: &str, with_opentelemetry: bool) {
...
+
+                    SdkTracerProvider::builder()
+                        .with_resource(resource)
+                        .with_span_processor(processor)
+                        .with_sampler(sdktrace::Sampler::AlwaysOn)
+                        .build()
+                }
+                #[cfg(feature = "tempo")]
+                "otlp" => {
+                    let otlp_endpoint = std::env::var("OTEL_EXPORTER_OTLP_ENDPOINT")
```
deuszx: LINERA_ in env var.

linera-base/src/chrome_trace_exporter.rs
```diff
@@ -0,0 +1,369 @@
...
+        Self::new()
+    }
+}
+
+/// OpenTelemetry span exporter that writes spans to a Chrome trace JSON file.
+///
+/// The exported file can be opened in `chrome://tracing` or the Perfetto UI
+/// for visualization and analysis of distributed traces.
+pub struct ChromeTraceExporter {
+    trace_file: Arc<Mutex<ChromeTraceFile>>,
```
deuszx: If you use `ChromeTrace` here instead, you will not need to manually construct the output JSON file (as you do now down below).

linera-base/src/chrome_trace_exporter.rs
```diff
@@ -0,0 +1,369 @@
...
+    fn add_event(&mut self, event: &ChromeTraceEvent) -> std::io::Result<()> {
+        if !self.is_first_event {
+            writeln!(self.file, ",")?;
+        }
+        self.is_first_event = false;
+
+        let json = serde_json::to_string(event)?;
+        write!(self.file, "    {}", json)?;
+        Ok(())
+    }
```
deuszx: The intermediate state of the file is not valid Chrome Trace IIUC (it's not valid JSON even, b/c it doesn't have the closing brackets etc).

linera-base/src/chrome_trace_exporter.rs
ndr-ds: Sorry, I failed to properly review some code Claude wrote, will publish the PR back once I've done that

linera-base/src/chrome_trace_exporter.rs
ndr-ds: Crazy indeed 😅

linera-base/src/chrome_trace_exporter.rs
ndr-ds: Should be fixed once we start properly using `ChromeTrace`

linera-base/src/chrome_trace_exporter.rs
Twey, line 1: Is there a reason we're rolling our own here instead of using [`tracing-chrome`](https://docs.rs/tracing-chrome/latest/tracing_chrome/)?

Twey:
Can we reuse the existing `RUST_LOG_FORMAT` environment variable for log formats?

linera-base/src/chrome_trace_exporter.rs
ndr-ds: `tracing-chrome` has not been updated in over a year. Is it actively maintained? I thought it wasn't 🤔

linera-base/src/tracing.rs
Twey: Can we reuse the existing `RUST_LOG_FORMAT` environment variable for log formats and output to stderr as per [12factor](https://12factor.net/logs)?

ndr-ds: But this is not a log file 🤔 it's a JSON file in a specific Chrome trace format

linera-base/src/chrome_trace_exporter.rs
ndr-ds: If it is, then happy to use it! Should simplify the code quite a bit

ndr-ds: Maybe I'll just use it 😅 even if outdated it surely is better than my code

Twey: The format hasn't changed in the last year as I know — should still be good :)

linera-base/src/tracing.rs
Twey: In general I don't like (and 12factor advises against) writing to files, or baking other log sinks into the application, because it presupposes things about the environment, such that there's a filesystem (which there isn't, usually, in the case of Wasm). It would be really nice to be able to use these from the Web, but so long as it's hard-coded to write to a file we probably can't, at least without jumping through hoops to simulate a filesystem.

Maybe we could take a `Write`r here as an input?

(Tracing is a superset of logging — this is why we use the `tracing` crate to log output 🙂)

ndr-ds: Good point! I guess I misunderstood because from the link it seems that tracing would need to be a subset of logging, not a superset, for it to apply 🤔
But anyways. I can make it more generic.

linera-base/src/tracing_otel.rs
```diff
@@ -0,0 +1,109 @@
...
+        .with_resource(resource)
+        .with_batch_exporter(exporter)
+        .with_sampler(opentelemetry_sdk::trace::Sampler::AlwaysOn)
+        .build();
+
+    global::set_tracer_provider(tracer_provider.clone());
+    let tracer = tracer_provider.tracer("linera");
+
+    let telemetry_only_filter =
+        filter_fn(|metadata| metadata.is_span() && metadata.target() == "telemetry_only");
```
Twey, line 47: I'm a bit unsure about this use of `target`.  Targets in `tracing` typically refer to a chunk of code, e.g. a module, and they work on a prefix basis.  It sounds a bit like we're trying to introduce a custom log level that's even more verbose than `TRACE`, but `TRACE` definitionally can be arbitrarily verbose — is there a reason we can't put these events in `TRACE`?

linera-base/src/tracing_otel.rs
```diff
@@ -0,0 +1,109 @@
...
+
+/// Initializes tracing with Chrome Trace JSON exporter.
+///
+/// Returns a guard that must be kept alive for the duration of the program.
+/// When the guard is dropped, the trace data is flushed and completed.
+///
+/// Exports traces to Chrome Trace JSON format which can be visualized in:
+/// - Chrome: `chrome://tracing`
+/// - Perfetto UI: <https://ui.perfetto.dev>
+pub fn init_with_chrome_trace_exporter<W>(log_name: &str, writer: W) -> ChromeTraceGuard
```
Twey, line 91: It's a bit weird for this function to live in a file called `tracing_otel.rs` since it's not related to OpenTelemetry.  It would be nice to move it to `tracing.rs` with the other generic tracing infrastructure.

(I think we should actually move what's currently into `tracing.rs` to `linera_service::cli` and `tracing_web.rs` into `web/@linera/client`, but that's a different matter that's probably out of scope here.)

Twey, line 91: Do we need `W` to be `Send`?  This also probably rules out using it on the Web (since whatever writer we pass in is likely to refer to a JavaScript object, which are not `Send`).

linera-service/src/server.rs
```diff
@@ -506,7 +506,7 @@ fn log_file_name_for(command: &ServerCommand) -> Cow<'static, str> {
 }
 
 async fn run(options: ServerOptions) {
-    linera_base::tracing::init_with_opentelemetry(&log_file_name_for(&options.command)).await;
+    linera_base::tracing::init_with_opentelemetry(&log_file_name_for(&options.command), None);
```
Twey, line 511: This function seems misnamed — the name passed here is not a filename.

linera-base/src/tracing_otel.rs
ndr-ds: My thinking was that I wanted a solution where I could just send stuff to Tempo, without actually printing anything to the screen, unless explicitly stated otherwise. So for example, if we're running in production with log level `INFO`, that means that the spans that are below `INFO` will not get sent to Tempo. So if we just add a bunch of instrumentation under the `TRACE` log level, that means we need super spammy logs in order to get the tracing information in Tempo to analyse.
So basically `telemetry_only` guarantees that it will be sent to Tempo. Otherwise, it only gets sent if the log level is satisfied.
So that's why I did this `telemetry_only` thing 🤔 Of course there's a very real chance that I've missed something and there's a better solution here, so open to alternatives :)

linera-base/src/tracing_otel.rs
ndr-ds: `tracing_otel` might be a bad name here. Basically a lot of these OTel related crates are not compatible with `wasm32`. So I wanted to isolate the non `wasm32` compatible thing in it's own module/file, so I created this one. Maybe I should rename it?

linera-service/src/server.rs
ndr-ds: It is a filename 🤔 we just also use it as the service name for convenience

linera-base/src/tracing_otel.rs
ndr-ds: Unfortunately we do :/ https://docs.rs/tracing-chrome/0.7.2/tracing_chrome/struct.ChromeLayerBuilder.html#method.writer

Twey: I guess we can concoct something with a channel 🤔

Twey: I think `tracing.rs` is used only on UNIX. We don't use it for on-chain bytecode, and on the Web we use `tracing_web.rs`.

linera-service/src/server.rs
Twey: Here's it's not the name of a file — so not a filename :) The function actually returns an identifier for the log stream, which is later used (in the other code path) to construct a filename.

linera-service/src/cli/main.rs
```diff
@@ -2001,17 +2001,47 @@ async fn kill_all_processes(pids: &[u32]) {
...
 
-fn should_init_opentelemetry(command: &ClientCommand) -> bool {
-    matches!(command, ClientCommand::Faucet { .. })
-}
-
-fn main() -> anyhow::Result<()> {
-    let options = ClientOptions::init();
-
-    if !should_init_opentelemetry(&options.command) {
+#[cfg(not(target_arch = "wasm32"))]
```
ma2bd, line 2004: remove this?

linera-service/src/server.rs
ma2bd: @ndr-ds Let's fix this as well. Did you mean `log_name_for` ?

PR #4721, opened 2025-10-02
Fix sender block sync.

deuszx:
# Motivation

When synchronising a chain, we were downloading only the sender block – without it's ancestors.

# Proposal

Download the block with all the ancestors.

# Testing strategy.

CI.

 # Notes

Needs backport to testnet branch.

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -1579,12 +1579,10 @@ pub enum ChainClientError {
     #[error("Epoch is already revoked")]
     EpochAlreadyRevoked,
 
-    #[error(
-        "Failed to download missing sender blocks from chain {chain_id} at heights {heights:?}"
-    )]
-    CannotDownloadMissingSenderBlocks {
+    #[error("Failed to download missing sender blocks from chain {chain_id} at height {height}")]
+    CannotDownloadMissingSenderBlock {
         chain_id: ChainId,
-        heights: Vec<BlockHeight>,
+        height: BlockHeight,
     },
 }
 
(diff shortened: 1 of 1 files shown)


linera-core/src/client/mod.rs
```diff
@@ -2320,58 +2318,48 @@ impl<Env: Environment> ChainClient<Env> {
...
-        let certificates = stream::iter(missing_blocks.into_iter())
+        stream::iter(missing_blocks.into_iter())
             .map(|(sender_chain_id, heights)| {
+                let height = heights.into_iter().max();
                 let nodes = nodes.clone();
+                let this = self.clone();
                 async move {
+                    let Some(height) = height else {
+                        return Ok(());
+                    };
```
deuszx, line 2331: So we don't download the _highest_ missing sender block?

linera-core/src/client/mod.rs
```diff
@@ -2320,58 +2318,48 @@ impl<Env: Environment> ChainClient<Env> {
...
                     // Try to download from any node.
                     for node in &nodes {
-                        if let Ok(certs) = node
-                            .download_certificates_by_heights(sender_chain_id, heights.clone())
+                        if let Err(err) = this
+                            .download_sender_block_with_sending_ancestors(
+                                sender_chain_id,
+                                height,
+                                node,
+                            )
```
deuszx, line 2339: I think it might create duplicate requests 🤔 If we're missing blocks 3,4,5 and 5 references 4, and 4 references 3 then
- request for 5th one will download 5,4,3
- for the 4th one will download 4,3
- etc

PR #4723, opened 2025-10-02
Fix sender block sync. (#4721)

afck:
Backport of #4721.

# Motivation

When synchronising a chain, we were downloading only the sender block – without it's ancestors.

# Proposal

Download the block with all the ancestors.

# Testing strategy.

CI.

 # Notes

PR to main: #4721.


afck:
Duplicate of #4722.

PR #4727, opened 2025-10-03
Address the case of W::NUM_INIT_KEYS being 0.

MathieuDutSik:
## Motivation

It turns out that `cunks_exact(N)` requires N to be non-zero. This is a problem since some legitimate structures have `N=0`.

## Proposal

In the case of size 0, we put specific code for that. A more comprehensive solution was possible by checking ahead for size 0, but that is not needed.

## Test Plan

The CI.

## Release Plan

This is a bug correction and it should be backported to Testnet Conway.
This is because the bug applies to some structures like
```rust
CollectionView<String, MapView<String, u64>>
```

which are used for the Game of Life.

## Links

None.

### DIFF
--- a/linera-views/src/views/collection_view.rs
+++ b/linera-views/src/views/collection_view.rs
@@ -397,7 +397,11 @@ impl<W: View> ByteCollectionView<W::Context, W> {
             .await?;
 
         for (loaded_values, (position, context)) in
-            values.chunks_exact(W::NUM_INIT_KEYS).zip(entries_to_load)
+            if W::NUM_INIT_KEYS == 0 {
+                Vec::new().into_iter()
+            } else {
+                values.chunks_exact(W::NUM_INIT_KEYS).collect::<Vec<_>>().into_iter()
+            }.zip(entries_to_load)
         {
             let view = W::post_load(context, loaded_values)?;
             let updates = self.updates.read().await;
--- a/linera-views/src/views/reentrant_collection_view.rs
+++ b/linera-views/src/views/reentrant_collection_view.rs
@@ -539,9 +539,11 @@ impl<W: View> ReentrantByteCollectionView<W::Context, W> {
             }
         }
         let values = self.context.store().read_multi_values_bytes(keys).await?;
-        for (loaded_values, short_key) in values
-            .chunks_exact(W::NUM_INIT_KEYS)
-            .zip(short_keys_to_load)
+        for (loaded_values, short_key) in if W::NUM_INIT_KEYS == 0 {
+                Vec::new().into_iter()
+            } else {
+                values.chunks_exact(W::NUM_INIT_KEYS).collect::<Vec<_>>().into_iter()
+            }.zip(short_keys_to_load)
         {
             let key = self
                 .context
@@ -638,7 +640,11 @@ impl<W: View> ReentrantByteCollectionView<W::Context, W> {
                 .read_multi_values_bytes(keys_to_load)
                 .await?;
             for (loaded_values, (position, short_key, context)) in
-                values.chunks_exact(W::NUM_INIT_KEYS).zip(entries_to_load)
+                if W::NUM_INIT_KEYS == 0 {
+                    Vec::new().into_iter()
+                } else {
+                    values.chunks_exact(W::NUM_INIT_KEYS).collect::<Vec<_>>().into_iter()
+                }.zip(entries_to_load)
             {
                 let view = W::post_load(context, loaded_values)?;
                 let wrapped_view = Arc::new(RwLock::new(view));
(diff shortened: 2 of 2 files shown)


ma2bd:
what's the impact of the bug?

MathieuDutSik:
> what's the impact of the bug?

The bug makes `try_load_entries` fail for the `CollectionView<T,W>` with `W::NUM_INIT_KEYS = 0`.
This is, for example, the case of `gol-challenge`, which has
```rust
pub reported_solutions: CollectionView<AccountOwner, MapView<DataBlobHash, Timestamp>>,
```

The function `try_load_entries` is used by `fn entries` of the GraphQL function of `CollectionView`. Therefore, some GraphQL queries of `gol-challenge` will fail (though with an explicit error message).

The correction to `ReentrantCollectionView` is somewhat less important. This container is not used in the smart contracts. It is used in the protocol, but the 4 instances `KeyValueStoreView`, `InboxStateView`, `OutboxStateView`, and `PendingBlobsView` all have `W::NUM_INIT_KEYS > 0`. So, the correction, while useful, does not have a direct impact.

linera-views/src/views/collection_view.rs
```diff
@@ -386,25 +386,36 @@ impl<W: View> ByteCollectionView<W::Context, W> {
...
-            });
+        if W::NUM_INIT_KEYS == 0 {
+            for (position, context) in entries_to_load {
+                let view = W::post_load(context, &[])?;
+                let updates = self.updates.read().await;
+                results[position] = Some(ReadGuardedView::NotLoaded {
+                    _updates: updates,
+                    view,
+                });
+            }
```
ma2bd, line 397: Here and below, do we have to duplicate so much code? It seems that the previous version of the PR was almost there.

linera-views/src/views/collection_view.rs
```diff
@@ -397,7 +397,11 @@ impl<W: View> ByteCollectionView<W::Context, W> {
             .await?;
 
         for (loaded_values, (position, context)) in
-            values.chunks_exact(W::NUM_INIT_KEYS).zip(entries_to_load)
+            if W::NUM_INIT_KEYS == 0 {
+                Vec::new().into_iter()
+            } else {
+                values.chunks_exact(W::NUM_INIT_KEYS).collect::<Vec<_>>().into_iter()
+            }.zip(entries_to_load)
```
ma2bd: Isn't it an obvious case for an intermediate variable?
```
let chunks = if W::NUM_INIT_KEYS == 0 {
    std::iter::empty()   // <- better ?
} else {
    values.chunks_exact(W::NUM_INIT_KEYS).collect::<Vec<_>>().into_iter() // <- why collect().into_iter() ???
};
for ... in chunks.zip(entries_to_load) {
    ...
}
```

ma2bd (changes requested):
Good catch. Let's try to keep the code reasonably concise

linera-views/src/views/collection_view.rs
MathieuDutSik: This does not compile
```rust
399 |           let chunks = if W::NUM_INIT_KEYS == 0 {
    |  ______________________-
400 | |             std::iter::empty()
    | |             ------------------ expected because of this
401 | |         } else {
402 | |             values.chunks_exact(W::NUM_INIT_KEYS)
    | |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `Empty<_>`, found `ChunksExact<'_, Option<Vec<u8>>>`
403 | |         };
    | |_________- `if` and `else` have incompatible types
    |
    = note: expected struct `std::iter::Empty<_>`
               found struct `std::slice::ChunksExact<'_, std::option::Option<std::vec::Vec<u8>>>`
```

linera-views/src/views/collection_view.rs
MathieuDutSik: But it was wrong. It was not working. It was not crashing, but the results were incorrect.

To avoid duplication, what we can do is avoid entirely the use of the `chunks_exact` and instead use explicit for loops and chunks obtained via `&[pos*W::NUM_INIT_KEYS..(pos+1)*W::NUM_INIT_KEYS]`.

It is perfectly fine code, and it is the one that was in the first version before being changed by Janito.

linera-views/src/views/collection_view.rs
ma2bd: Indeed, one has to use `itertools::Either`:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=104d68519c1b915603ed998c75dff5ad

https://gist.github.com/rust-play/104d68519c1b915603ed998c75dff5ad

linera-views/src/views/collection_view.rs
ma2bd: I would really avoiding code duplication. See also answer below (if this helps?)

MathieuDutSik: Yes, of course. I also do not like this.
I wrote explicitly that we can avoid code duplication with a for loop. But that seems forbidden. At least you did not comment on that.

So, code duplication seems unavoidable.

linera-views/src/views/collection_view.rs
MathieuDutSik: That does compile.
But it is not correct. We do not read the entries correctly.

This is because if you had `vec1.iter().zip(vec2)` and vec1 and vec2 are not of the same length, then the returned result is of length the minimum of the two.

MathieuDutSik: If you want to experiment, I added a `test_nested_collection_map_view` that exercises all those critical points.

ma2bd: Here is my experiment: #4731

MathieuDutSik:
The correction has been done by PR #4731

PR #4728, opened 2025-10-03
Make chain manager recover from `current_round` bug.

afck:
## Motivation

https://github.com/linera-io/linera-protocol/pull/4688 fixed a bug where the chain manager would set its `signed_proposal` but not update the current round accordingly. However, some existing chains may already be in a bad state.

## Proposal

Detect the bad state and recompute the current round.

## Test Plan

Backport and see if it makes the affected chains recover. Otherwise we should never see the warning log.

## Release Plan

    - be released in a new SDK,

## Links

- Original bug fix: https://github.com/linera-io/linera-protocol/pull/4688

### DIFF
--- a/linera-chain/src/manager.rs
+++ b/linera-chain/src/manager.rs
@@ -699,6 +699,16 @@ where
         }
         if let Some(old_proposal) = self.signed_proposal.get() {
             if old_proposal.content.round >= proposal.content.round {
+                if *self.current_round.get() < old_proposal.content.round {
+                    tracing::warn!(
+                        chain_id = %proposal.content.block.chain_id,
+                        current_round = ?self.current_round.get(),
+                        proposal_round = ?old_proposal.content.round,
+                        "Proposal round is greater than current round. Updating."
+                    );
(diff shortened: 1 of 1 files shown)


linera-chain/src/manager.rs
```diff
@@ -699,6 +699,16 @@ where
...
+                if *self.current_round.get() < old_proposal.content.round {
+                    tracing::warn!(
+                        chain_id = %proposal.content.block.chain_id,
+                        current_round = ?self.current_round.get(),
+                        proposal_round = ?old_proposal.content.round,
+                        "Proposal round is greater than current round. Updating."
+                    );
+                    self.update_current_round(local_time);
+                    return true;
+                }
```
bart-linera, line 711: This shouldn't be possible for new validators to run into, right? Only for the validators that were running the old version of the code and got stuck in a bad state? If yes, I would only merge it to `testnet_conway`, and not to `main`, because the code should be useless on future networks.

afck: Exactly; that's why I made it WARN-level.

PR #4733, opened 2025-10-05
Add command to show ownership configuration of a chain

ma2bd:
## Motivation

`change-ownership` is pretty hard to use (and unsafe)

## Proposal

* Add a command show-ownership
* Refuse to change ownerwhip to 0 owners
* Use JSON outputs consistently

## Test Plan

CI + tested manually

## Release Plan

    - be released in a new SDK,

### DIFF
--- a/linera-service/src/cli/main.rs
+++ b/linera-service/src/cli/main.rs
@@ -265,6 +265,13 @@ impl Runnable for Job {
                 chain_id,
                 ownership_config,
             } => {
+                ensure!(
+                    !ownership_config.super_owners.is_empty()
+                        || !ownership_config.owners.is_empty(),
+                    "This command requires at least one owner or super owner to be set. \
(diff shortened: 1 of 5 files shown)


linera-service/src/cli/main.rs
```diff
@@ -253,10 +253,25 @@ impl Runnable for Job {
...
+
             ChangeOwnership {
                 chain_id,
                 ownership_config,
             } => {
+                ensure!(
+                    !ownership_config.super_owners.is_empty()
+                        || !ownership_config.owners.is_empty(),
+                    "This command requires at least one owner or super owner to be set. \
+                     To close a chain, use `close-chain`. To show the current config, use `show-ownership`."
```
afck, line 272: (We could consider disallowing it even as part of the protocol. One should always use `close-chain` instead.)

afck (approved):

PR #4734, opened 2025-10-06
Improve a few error and log message strings.

afck:
## Motivation

The pending block clearing message is INFO-level, but it's spammy because it also shows for each successfully committed block.

The `UnexpectedBlockHeight` error is confusing, because it's not clear which height is which.

## Proposal

Make the pending block message DEBUG.

Change the message of `UnexpectedBlockHeight`.

## Test Plan

CI

## Release Plan

- We could backport it to `testnet_conway`, then:
    - release a new SDK,
    - do a validator hotfix.

## Links

### DIFF
--- a/linera-chain/src/lib.rs
+++ b/linera-chain/src/lib.rs
@@ -106,10 +106,13 @@ pub enum ChainError {
     },
     #[error("The signature was not created by a valid entity")]
     InvalidSigner,
-    #[error("Chain is at height {chain_tip_height} but block is at {block_height}")]
+    #[error(
+        "Chain is expecting a next block at height {expected_block_height} but the given block \
+        is at height {found_block_height} instead"
+    )]
     UnexpectedBlockHeight {
-        chain_tip_height: BlockHeight,
-        block_height: BlockHeight,
+        expected_block_height: BlockHeight,
+        found_block_height: BlockHeight,
     },
     #[error("The previous block hash of a new block should match the last block of the chain")]
     UnexpectedPreviousBlockHash,
--- a/linera-core/src/chain_worker/state.rs
+++ b/linera-core/src/chain_worker/state.rs
@@ -1038,8 +1038,8 @@ where
         ensure!(
             height == chain.tip_state.get().next_block_height,
             WorkerError::UnexpectedBlockHeight {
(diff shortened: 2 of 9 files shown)


linera-chain/src/lib.rs
```diff
@@ -106,12 +106,10 @@ pub enum ChainError {
     },
     #[error("The signature was not created by a valid entity")]
     InvalidSigner,
-    #[error(
-        "Was expecting block height {expected_block_height} but found {found_block_height} instead"
-    )]
+    #[error("Chain is at height {chain_tip_height} but block is at {block_height}")]
```
ma2bd: How about keeping the current fields and adding one for the chain:
`Chain {chain_id} is expecting a next block at height {expected_block_height} but the given block is at height {found_block_height} instead`

ma2bd: Mmm this won't be backportable though. Also I wonder if we should have a separate error from block proposals and one for confirmed blocks.

afck: Actually: I'll revert the field renaming, and _not_ add the chain ID yet. That way we can cleanly backport it to Conway.

afck: Separate error also wouldn't be backportable. Let's add the chain ID and separate the errors later.

afck: Reverted in https://github.com/linera-io/linera-protocol/pull/4734/commits/6de9883c74f3d6ef048d1f5cb5ec721d69afa5ab.

linera-core/src/chain_worker/state.rs
```diff
@@ -1379,11 +1379,12 @@ where
                 .await?;
         }
         if let Some(next_block_height) = query.test_next_block_height {
+            // If not, send the same error as if a block with next_block_height was proposed.
             ensure!(
                 chain.tip_state.get().next_block_height == next_block_height,
                 WorkerError::UnexpectedBlockHeight {
```
ma2bd, line 1385: Oh, this 3rd scenario for the error was indeed confusing because the block is not sent yet (IIRC).

afck: Yes, but as long as we can't add new error variants I think it's better to treat the chain info request as "the block" rather than the actual chain state on the validator.

PR #4738, opened 2025-10-06
Add more info to `query-validator[s]`; show only diff in `query-validators`.

afck:
Port of https://github.com/linera-io/linera-protocol/pull/4746.

## Motivation

It helps with debugging consensus issues to get a quick overview of which round the validators are in, and if there are any other discrepancies compared to the local node.

## Proposal

Add more information to `query-validator[s]`. In `query-validators`, show any data that differs from the local node, for every validator.

Example `query-validators` output:

```
Querying validators about chain 3e6bdd095d2e4e30f12e8da38ea1409f2442696b01badbda4226577df09479ff.

Local Node:
Linera protocol: v0.15.4
RPC API hash: dYdv0oSjieFam5sNCUoabG3O2HkATGwTHFKiudkk94o
GraphQL API hash: WkChYtBRBRRD6+FZP+CxcpytMknWIi3dgV3nEBcUkck
WIT API hash: vwRMPCb+RzZbBCJ3PmC6AprU5ns+1ivKT6xQTMxrs58E
Source code: https://github.com/linera-io/linera-protocol/tree/919d8b98eb (dirty)
Genesis config hash: 5c7d9d274b85444d8fe9824c300b4bc2d8fc3c0fdb937f80180f8cf63b095589
Block hash: f7a8dd540e60a9c7822f3746ed41cf5b5c29186c2b692368ccea2fdc20bb85f1
Next height: 45
Timestamp: 2025-10-06 08:55:57.418186
Epoch: 13
Round: single-leader round 40

Public key: 021b2bb25d442272865e26569c6c3a61dae538a844a1405cfe1fc83c1106d70d6c
Address: grpcs:linera-test.artifact.systems:443
Weight: 1
Source code: https://github.com/linera-io/linera-protocol/tree/b3604f3fa0b4214027e12cb599ad110cd8c7745a
Block hash: d2b34b803f9f5ea84c6feb4b852dcac129cef4d61867149fd661f4b59cc428e5
Next height: 46
Timestamp: 2025-10-06 11:35:30.451316
Round: single-leader round 0
```

## Release Plan

## Links

- PR to main: https://github.com/linera-io/linera-protocol/pull/4746

### DIFF
--- a/linera-client/src/client_context.rs
+++ b/linera-client/src/client_context.rs
@@ -14,7 +14,7 @@ use linera_base::{
 use linera_chain::types::ConfirmedBlockCertificate;
 use linera_core::{
     client::{ChainClient, Client},
-    data_types::{ChainInfoQuery, ClientOutcome},
+    data_types::{ChainInfo, ChainInfoQuery, ClientOutcome},
     join_set_ext::JoinSet,
     node::ValidatorNode,
     Environment, JoinSetExt as _,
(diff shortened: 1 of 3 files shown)


linera-client/src/client_context.rs
```diff
@@ -63,6 +63,95 @@ use crate::{
...
+            Ok(version_info)
+                if ref_version.is_some_and(|ref_version| ref_version == version_info) => {}
+            Ok(version_info) => {
+                print!("{version_info}");
+            }
+            Err(err) => println!("Error getting version info: {err}"),
+        }
+
+        let ref_info = reference.and_then(|ref_results| ref_results.chain_info.as_ref().ok());
+        match &self.chain_info {
```
deuszx, line 157: It'd be useful to also print the chain ID. Without it we don't really know what chain the data below refers to 👀

afck: Although I think that's only necessary if the user didn't specify a chain? I updated it, and the example output in the description.

deuszx: I'd simply print it always, regardless of the usage. So that the data set shown is always the same.

deuszx (approved):

linera-client/src/client_context.rs
afck: Done in https://github.com/linera-io/linera-protocol/pull/4738/commits/c75553ac45d37f90066bc19cea026523c3ad5e18.

PR #4739, opened 2025-10-06
Backport 4406 / 4397 / 4394 / 4366 to TestNet Conway of end-to-end CI changes.

MathieuDutSik:
Backports #4406, #4397, #4394, #4366

## Motivation

Several changes to the CI are welcomed and make working easier.
Fewer end-to-end tests for what is not viewed as essential.

## Proposal

Those 4 PRs are backported to TestNet Conway.

## Test Plan

The tests are tested in the same way as in the main.

## Release Plan

Backports to TestNet Conway.

## Links

None.


linera-sdk/tests/fixtures/.gitignore
```diff
@@ -0,0 +1 @@
+target/
```
afck, line 1: Would it be better to add that to the `.gitignore` in the root directory instead?

afck (approved):
Approving, assuming there weren't any major merge conflicts. (I didn't re-review all the code.)

linera-sdk/tests/fixtures/.gitignore
MathieuDutSik: I was thinking about doing that. But there is already an `examples/.gitignore`. So, coherence requires somehow having one for `linera-sdk/tests/fixtures`.

In total, we have 15 `.gitignore`. Maybe we could reduce that. But I would that this is outside of the scope ofn this work.

PR #4748, opened 2025-10-07
Send only the relevant sender chain on `MissingCrossChainUpdate`.

afck:
## Motivation

`send_chain_information_for_senders` can cause many requests if there are lots of sender chains. The `MissingCrossChainUpdate` code was written with user chains in mind, that would only receive messages from a handful of other chains.

## Proposal

Send only the requested missing update.

That can mean that there will be many iterations of that loop, if the validator is very far behind, but it should be more lightweight than `send_chain_information_for_senders`.

## Test Plan

CI

## Release Plan

- These changes should be backported to `testnet_conway` and released in a new SDK.

## Links

### DIFF
--- a/linera-core/src/updater.rs
+++ b/linera-core/src/updater.rs
@@ -393,8 +393,9 @@ where
                     // Some received certificates may be missing for this validator
                     // (e.g. to create the chain or make the balance sufficient) so we are going to
                     // synchronize them now and retry.
-                    self.send_chain_info_up_to_heights(
-                        vec![(origin, height.try_add_one()?)],
+                    self.send_chain_information(
+                        origin,
+                        height.try_add_one()?,
                         CrossChainMessageDelivery::Blocking,
                     )
                     .await?;


linera-core/src/updater.rs
```diff
@@ -380,12 +380,24 @@ where
...
                 }
-                Err(NodeError::MissingCrossChainUpdate { .. }) if !sent_cross_chain_updates => {
+                Err(NodeError::MissingCrossChainUpdate {
+                    chain_id,
+                    origin,
+                    height,
+                }) if chain_id == proposal.content.block.chain_id
+                    && sent_cross_chain_updates
+                        .get(&origin)
+                        .is_none_or(|h| *h < height) =>
```
deuszx, line 390: If `height < h` (i.e. there's a block on `origin` at height `height` is sending a message to `chain_id` that this validator hasn't received yet) – shouldn't we send those blocks anyway?

afck: `send_chain_information` sends all blocks from the sender chain that we have, up to the given height. So if `h > height` here, that means we have already sent them everything we have, up to a higher height. No need to try again for now.

deuszx: Yes, I realized that after I wrote it but you were faster.

PR #4752, opened 2025-10-08
Introduce RequestsScheduler to track (and deduplicate) all requests and validators' scores

deuszx:
## Motivation

  The Linera client needs to interact with multiple validator nodes efficiently. Previously, the
  client would make individual requests to validators without:

  1. Performance tracking: No mechanism to prefer faster, more reliable validators
  2. Request deduplication: Concurrent requests for the same data would all hit the network, wasting
  bandwidth and validator resources
  3. Response caching: Repeated requests for the same data would always go to validators
  4. Load balancing: No rate limiting per validator, risking overload
  5. Resilience: No fallback mechanism when a validator is slow or unresponsive

  This led to:
  - Unnecessary network traffic and validator load
  - Poor user experience with redundant waiting
  - No optimization based on validator performance
  - Risk of overwhelming validators with too many concurrent requests
  - No recovery mechanism when validators are slow

##  Proposal

  This PR introduces `RequestsScheduler`, a sophisticated request orchestration layer that provides
  intelligent peer selection, request deduplication, caching, and performance-based routing.

  Key Features

  1. Performance Tracking with Exponential Moving Averages (EMA)
  - Tracks latency, success rate, and current load for each validator
  - Uses configurable weights to compute a composite performance score
  - Intelligently selects the best available validator for each request
  - Weighted random selection from top performers to avoid hotspots

  2. Request Deduplication
  - Exact matching: Multiple concurrent requests for identical data are deduplicated
  - Subsumption-based matching: Smaller requests are satisfied by larger in-flight requests that
  contain the needed data (e.g., a request for blocks 10-12 can be satisfied by an in-flight request
  for blocks 10-20)
  - Broadcast mechanism ensures all waiting requesters receive the result when the request completes
  - Timeout handling: Stale in-flight requests (>200ms) are not deduplicated against, allowing fresh
  attempts

  3. Response Caching
  - Successfully completed requests are cached with configurable TTL (default: 2 seconds)
  - LRU eviction when cache reaches maximum size (default: 1000 entries)
  - Works with both exact and subsumption matching
  - Only successful results are cached

  4. Slot-Based Rate Limiting
  - Each validator has a maximum concurrent request limit (default: 100)
  - Async await mechanism: requests wait for available slots without polling
  - Prevents overloading individual validators
  - Automatic slot release on request completion

  5. Alternative Peer Handling
  - When multiple callers request the same data, they register as "alternative peers"
  - If the original request times out (>200ms), any alternative peer can complete the request
  - The result is broadcast to all waiting requesters
  - Provides resilience against slow validators

  6. Modular Architecture

  Created a new ` requests_scheduler` module with clear separation of concerns:
```
 requests_scheduler/
  ├── mod.rs              - Module exports and constants
  ├── scheduler.rs          - RequestsScheduler orchestration logic
  ├── in_flight_tracker.rs - In-flight request tracking and deduplication
  ├── node_info.rs        - Per-validator performance tracking
  ├── request.rs          - Request types and result extraction
  └── scoring.rs          - Configurable scoring weights
```
  API

  High-level APIs:
```rust
  // Execute with best available validator
  scheduler.with_best(request_key, |peer| async {
      peer.download_certificates(chain_id, start, limit).await
  }).await

  // Execute with specific validator
  scheduler.with_peer(request_key, peer, |peer| async {
      peer.download_blob(blob_id).await
  }).await

  Configuration:
  let manager = RequestsScheduler::with_config(
      validator_nodes,
      max_requests_per_node: 100,
      weights: ScoringWeights { latency: 0.4, success: 0.4, load: 0.2 },
      alpha: 0.1,              // EMA smoothing factor
      max_expected_latency_ms: 5000.0,
      cache_ttl: Duration::from_secs(2),
      max_cache_size: 1000,
  );
```

  Benefits

  - Reduced network load: Deduplication and caching eliminate redundant requests
  - Better performance: Intelligent peer selection routes to fastest validators
  - Improved reliability: Alternative peer mechanism provides resilience
  - Protection for validators: Rate limiting prevents overload
  - Efficient resource usage: EMA-based scoring optimizes validator selection
  - Clean architecture: Modular design makes code maintainable and testable

  Metrics

  In production usage, this should significantly reduce:
  - Network traffic between clients and validators
  - Validator CPU/memory usage from redundant requests
  - Client request latency through caching and smart routing
  - Failed requests through performance tracking and rate limiting

The following metrics have been added to Prometheus (with compiled with `--features metrics`):
- `requests_scheduler_response_time_ms` - Response time for requests to validators in milliseconds
- `requests_scheduler_request_total` - Total number of requests made to each validator
- `requests_scheduler_request_success` - Number of successful requests to each validator (`(requests_scheduler_request_total - requests_scheduler_request_success) / requests_scheduler_request_total` is an error rate)
- `requests_scheduler_request_deduplication_total` - Number of requests that were deduplicated by joining an in-flight request
- `requests_scheduler_request_cache_hit_total` - Number of requests that were served from cache

## Test Plan

Existing CI makes sure we maintain backwards compatibility. Some tests have been added to the new modules.

## Release Plan

## Links

### DIFF
--- a/linera-core/src/client/requests_scheduler/scheduler.rs
+++ b/linera-core/src/client/requests_scheduler/scheduler.rs
@@ -358,8 +358,7 @@ impl<Env: Environment> RequestsScheduler<Env> {
             },
             peer.clone(),
             |peer| async move {
-                peer.download_certificates_by_heights(chain_id, heights)
-                    .await
+                Box::pin(peer.download_certificates_by_heights(chain_id, heights)).await
             },
         )
         .await


linera-core/src/client/mod.rs
```diff
@@ -347,8 +355,21 @@ impl<Env: Environment> Client<Env> {
...
+                    RequestKey::Certificates {
+                        chain_id,
+                        start: next_height,
+                        limit,
+                    },
+                    remote_node.clone(),
+                    async move |peer| {
+                        peer.download_certificates_from(chain_id, next_height, limit)
+                            .await
+                    },
```
bart-linera: We seem to be passing the same info twice: once in the `RequestKey`, once in the operation to be executed. Maybe it would be possible to make the `RequestKey` determine the operation?

deuszx: I know, but I couldn't find a way to get rid of this duplication. Caller chooses the operation (the closure) but he code that executes the closure (`track_request`) and caching logic (`deduplicated_request`) are separate functions. What I could do is merge those two but then testing will be much more difficult.

bart-linera: What I was thinking was that maybe the operation could be chosen by matching on the `RequestKey` inside the `ValidatorManager` (or some method on the `RequestKey`, even?), instead of by passing a closure. But maybe that wouldn't work with some use cases? I haven't reviewed most of the code yet, those are loose ideas I'm getting while reading the PR.

deuszx: Sure. Initially `ValidatorManager` had the matching API to `RemoteNode` – i.e. it had methods like `download_certificate`, `download_blob`, `download_certificate_for_blob` but with time I moved away from this and worked with the closure (although there are still methods for downloading blobs). I could revert to that approach but I thought that it's more limiting.

linera-core/src/client/validator_manager/request.rs
```diff
@@ -0,0 +1,235 @@
...
+
+    /// Converts certificate-related requests to a common representation of (chain_id, sorted heights).
+    ///
+    /// This helper method normalizes both `Certificates` and `CertificatesByHeights` variants
+    /// into a uniform format for easier comparison and overlap detection.
+    ///
+    /// # Returns
+    /// - `Some((chain_id, heights))` for certificate requests, where heights are sorted
+    /// - `None` for non-certificate requests (Blob, PendingBlob, CertificateForBlob)
+    fn height_range(&self) -> Option<Vec<BlockHeight>> {
```
afck: We could return a `BTreeSet` instead, to make `contains` more efficient in the `subsumes` implementation. Or, if they are always ordered anyway, we could avoid calling `contains` below at all, and just iterate over both?

linera-core/src/client/requests_scheduler/request.rs
```diff
@@ -0,0 +1,235 @@
...
+    fn try_extract_result(&self, from: &Self, result: &RequestResult) -> Option<RequestResult> {
+        // Only certificate results can be extracted
+        let certificates = match result {
+            RequestResult::Certificates(certs) => certs,
+            _ => return None,
+        };
+
+        if self.chain_id().is_none() || from.chain_id().is_none() {
+            return None;
+        }
```
afck, line 173: Instead of all this, shouldn't we just `if !self.subsumes(result) { return None; }`?

Otherwise it isn't guaranteed that the `filtered` results below are complete.

linera-core/src/client/validator_manager/mod.rs
```diff
@@ -0,0 +1,24 @@
...
+pub use request::RequestKey;
+pub use scoring::ScoringWeights;
+
+// Module constants
+pub(super) const MAX_IN_FLIGHT_REQUESTS: usize = 100;
+pub(super) const MAX_ACCEPTED_LATENCY_MS: f64 = 5000.0;
+pub(super) const CACHE_TTL_SEC: u64 = 2;
+pub(super) const CACHE_MAX_SIZE: usize = 1000;
+// Maximum latency for an in-flight request before we stop deduplicating it.
+pub(super) const MAX_REQUEST_TTL_MS: u64 = 200;
```
afck: Should be CLI options and/or env variables, I guess.

linera-core/src/client/validator_manager/node_info.rs
```diff
@@ -0,0 +1,189 @@
...
+    /// Adapts quickly to changes in response time
+    pub(super) ema_latency_ms: f64,
+
+    /// Exponential Moving Average of success rate (0.0 to 1.0)
+    /// Tracks recent success/failure patterns
+    pub(super) ema_success_rate: f64,
+
+    /// Thread-safe counter tracking the number of concurrent requests currently in flight
+    /// Used to prevent overwhelming individual nodes with too many parallel requests
+    pub(super) in_flight_requests: Arc<tokio::sync::RwLock<usize>>,
```
afck: Or `AtomicUsize`?

linera-core/src/client/validator_manager/node_info.rs
```diff
@@ -0,0 +1,189 @@
...
+    ) -> Self {
+        Self {
+            node,
+            ema_latency_ms: 100.0, // Start with reasonable latency expectation
+            ema_success_rate: 1.0, // Start optimistically with 100% success
+            in_flight_requests: Arc::new(tokio::sync::RwLock::new(0)),
+            slot_available: Arc::new(tokio::sync::Notify::new()),
+            total_requests: 0,
+            weights,
+            alpha: alpha.clamp(0.01, 0.5), // Ensure alpha is in reasonable range
```
afck: Maybe log a warning or error if it isn't.

linera-core/src/client/requests_scheduler/node_info.rs
```diff
@@ -0,0 +1,189 @@
...
+    /// - **Latency score**: Inversely proportional to EMA latency
+    /// - **Success score**: Directly proportional to EMA success rate
+    /// - **Load score**: Inversely proportional to current load
+    ///
+    /// Returns a score from 0.0 to 1.0, where higher values indicate better performance.
+    pub(super) async fn calculate_score(&self) -> f64 {
+        // 1. Normalize Latency (lower is better, so we invert)
+        let latency_score = 1.0
+            - (self.ema_latency_ms.min(self.max_expected_latency_ms)
+                / self.max_expected_latency_ms);
```
afck, line 87: If the max latency is 5s, that means the latency scores of a node that returned after 100 ms will be very similar to that of a node that returned with 200 ms.

I think that's fine for now, especially since the requests can have very differently-sized responses anyway, but maybe something to tweak in the future.

linera-core/src/client/requests_scheduler/node_info.rs
```diff
@@ -0,0 +1,189 @@
...
+
+        // 4. Apply cold-start penalty for nodes with very few requests
+        let confidence_factor = (self.total_requests as f64 / 10.0).min(1.0);
+
+        // 5. Combine with weights
+        let raw_score = (self.weights.latency * latency_score)
+            + (self.weights.success * success_score)
+            + (self.weights.load * load_score);
+
+        // Apply confidence factor to penalize nodes with too few samples
```
afck, line 106: Is that necessary? If a node has too few samples, this makes us avoid getting more samples, doesn't it?

linera-core/src/client/validator_manager/node_info.rs
```diff
@@ -0,0 +1,189 @@
...
+        let current = *self.in_flight_requests.read().await;
+        current < max_requests
+    }
+
+    /// Atomically attempts to reserve a request slot for this node.
+    ///
+    /// Returns `true` if a slot was successfully acquired, `false` if the node
+    /// is already at maximum capacity. This prevents overwhelming individual
+    /// nodes with too many concurrent requests.
+    pub(super) async fn acquire_request_slot(&self, max_requests: usize) -> bool {
```
afck: Should `in_flight_requests` actually be a semaphore then?

linera-core/src/client/validator_manager/in_flight_tracker.rs
```diff
@@ -0,0 +1,218 @@
...
+    /// - `Some(InFlightMatch::Subsuming { key, outcome })`: Subsuming request found
+    pub(super) async fn try_subscribe(&self, key: &RequestKey) -> Option<InFlightMatch> {
+        let in_flight = self.entries.write().await;
+
+        // First check for exact match
+        if let Some(entry) = in_flight.get(key) {
+            let elapsed = Instant::now().duration_since(entry.started_at);
+
+            let outcome = if elapsed > self.timeout {
+                SubscribeOutcome::TimedOut(elapsed)
```
afck: But there might still be a more recent, subsuming (non-exact) one?

linera-core/src/client/validator_manager/in_flight_tracker.rs
```diff
@@ -0,0 +1,218 @@
...
+
+        if let Some(entry) = in_flight.remove(key) {
+            let waiter_count = entry.sender.receiver_count();
+            tracing::trace!(
+                key = ?key,
+                waiters = waiter_count,
+                "request completed; broadcasting result to waiters",
+            );
+            if waiter_count != 0 {
+                let _ = entry.sender.send(result);
```
afck: Maybe log something if that fails. (In general, we are trying to avoid `let _ = `.)

linera-core/src/client/requests_scheduler/cache.rs
```diff
@@ -0,0 +1,183 @@
...
+                    Some(key.clone())
+                } else {
+                    None
+                }
+            })
+            .collect();
+
+        for key in &expired_keys {
+            cache.remove(key);
+        }
```
afck, line 151: (Maybe simpler: `HashMap::retain`?)

linera-core/src/client/validator_manager/manager.rs
```diff
@@ -0,0 +1,1268 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
```
afck: (Do we still have to add this line to new files?)

linera-core/src/client/validator_manager/manager.rs
```diff
@@ -0,0 +1,1268 @@
...
+    pub static VALIDATOR_REQUEST_SUCCESS: LazyLock<IntCounterVec> = LazyLock::new(|| {
+        register_int_counter_vec(
+            "validator_manager_request_success",
+            "Number of successful requests to each validator",
+            &["validator"],
+        )
+    });
+
+    /// Counter for requests that were deduplicated (joined an in-flight request)
+    pub static REQUEST_CACHE_DEDUPLICATION: LazyLock<IntCounter> = LazyLock::new(|| {
```
afck: Isn't this always the same as `REQUEST_CACHE_HIT`?

linera-core/src/client/validator_manager/manager.rs
```diff
@@ -0,0 +1,1268 @@
...
+    pub static VALIDATOR_REQUEST_TOTAL: LazyLock<IntCounterVec> = LazyLock::new(|| {
+        register_int_counter_vec(
+            "validator_manager_request_total",
+            "Total number of requests made to each validator",
+            &["validator"],
+        )
+    });
+
+    /// Counter of successful requests per validator
+    pub static VALIDATOR_REQUEST_SUCCESS: LazyLock<IntCounterVec> = LazyLock::new(|| {
```
afck: We could make those metrics `pub(super)` that are only used in this file.

linera-core/src/client/validator_manager/manager.rs
```diff
@@ -0,0 +1,1268 @@
...
+    /// Thread-safe map of validator nodes indexed by their public keys.
+    /// Each node is wrapped with EMA-based performance tracking information.
+    nodes: Arc<tokio::sync::RwLock<BTreeMap<ValidatorPublicKey, NodeInfo<Env>>>>,
+    /// Maximum number of concurrent requests allowed per node.
+    /// Prevents overwhelming individual validators with too many parallel requests.
+    max_requests_per_node: usize,
+    /// Default scoring weights applied to new nodes.
+    default_weights: ScoringWeights,
+    /// Default EMA smoothing factor for new nodes.
+    default_alpha: f64,
```
afck: (We are always using the same smoothing factor for all nodes in any given `ValidatorManager`, are we? I'd omit the `default_` here, then, because it suggests that it can be changed for individual nodes. Same for the weights and latency.)

linera-core/src/client/validator_manager/manager.rs
```diff
@@ -0,0 +1,1268 @@
...
+    /// Creates a new `ValidatorManager` with default configuration.
+    pub fn new(nodes: impl IntoIterator<Item = RemoteNode<Env::ValidatorNode>>) -> Self {
+        Self::with_config(
+            nodes,
+            MAX_IN_FLIGHT_REQUESTS,
+            ScoringWeights::default(),
+            0.1,
+            MAX_ACCEPTED_LATENCY_MS,
+            Duration::from_secs(CACHE_TTL_SEC), // 60 second cache TTL
+            CACHE_MAX_SIZE,                     // cache up to 100 entries
```
afck: Those comments will become outdated if we change the default values. If I did, I wouldn't think of looking into this function.

linera-core/src/client/requests_scheduler/scheduler.rs
```diff
@@ -0,0 +1,1268 @@
...
+    pub fn with_config(
+        nodes: impl IntoIterator<Item = RemoteNode<Env::ValidatorNode>>,
+        max_requests_per_node: usize,
+        weights: ScoringWeights,
+        alpha: f64,
+        max_expected_latency_ms: f64,
+        cache_ttl: Duration,
+        max_cache_size: usize,
+    ) -> Self {
+        assert!(alpha > 0.0 && alpha < 1.0, "Alpha must be in (0, 1) range");
```
afck, line 183: Aren't we `clamp`ing it to a narrower range elsewhere? Why not assert that range then?

linera-core/src/client/validator_manager/request.rs
deuszx: I was meant to change the impl to use the ordering and I forgot 🙈

linera-core/src/client/validator_manager/manager.rs
```diff
@@ -0,0 +1,1268 @@
...
+    ) -> Self {
+        assert!(alpha > 0.0 && alpha < 1.0, "Alpha must be in (0, 1) range");
+        Self {
+            nodes: Arc::new(tokio::sync::RwLock::new(
+                nodes
+                    .into_iter()
+                    .map(|node| {
+                        let public_key = node.public_key;
+                        (
+                            public_key,
```
afck: Nit: I think this will format more nicely?

```suggestion
                        (
                            node.public_key,
```

Alternatively I'd `let node_info = NodeInfo::with_config(…)` and then return `(public_key, node_info)`.

linera-core/src/client/validator_manager/mod.rs
deuszx: Yes, that's one of the reasons it's still a draft – I have to move these to the CLI args.

linera-core/src/client/validator_manager/node_info.rs
deuszx: I blame Claude 😬

linera-core/src/client/requests_scheduler/node_info.rs
deuszx: We can decrease the `max_expected_latency_ms` to something like 500ms and tweak the parameters to favor lower latency even more.

linera-core/src/client/validator_manager/node_info.rs
deuszx: 🧠 

This might simplify the code a lot.

linera-core/src/client/validator_manager/manager.rs
```diff
@@ -0,0 +1,1268 @@
...
+            in_flight_tracker: InFlightTracker::new(Duration::from_millis(MAX_REQUEST_TTL_MS)),
+            cache: RequestsCache::new(cache_ttl, max_cache_size),
+        }
+    }
+
+    /// Executes an operation with an automatically selected peer, handling deduplication,
+    /// tracking, and peer selection.
+    ///
+    /// This method provides a high-level API for executing operations against remote nodes
+    /// while leveraging the ValidatorManager's intelligent peer selection, performance tracking,
```
afck:
```suggestion
    /// while leveraging the `ValidatorManager`'s intelligent peer selection, performance tracking,
```

linera-core/src/client/validator_manager/manager.rs
```diff
@@ -0,0 +1,1268 @@
...
+                    success = %is_success,
+                    response_time_ms = %response_time_ms,
+                    score = %score,
+                    total_requests = %info.total_requests,
+                    "Request completed"
+                );
+            }
+        }
+
+        // Record prometheus metrics
```
afck:
```suggestion
        // Record Prometheus metrics
```

linera-core/src/client/validator_manager/manager.rs
```diff
@@ -0,0 +1,1268 @@
...
+                    .inc();
+            }
+        }
+
+        result
+    }
+
+    /// Deduplicates concurrent requests for the same data.
+    ///
+    /// If a request for the same key is already in-flight, this method waits for
```
afck:
```suggestion
    /// If a request for the same key is already in flight, this method waits for
```

(Hyphen only before nouns, I think: https://www.oxfordlearnersdictionaries.com/us/definition/english/in-flight)

linera-core/src/client/validator_manager/manager.rs
```diff
@@ -0,0 +1,1268 @@
...
+    /// the existing request to complete and returns its result. Otherwise, it
+    /// executes the operation and broadcasts the result to all waiting callers.
+    ///
+    /// This method also performs **subsumption-based deduplication**: if a larger
+    /// request that contains all the data needed by this request is already cached
+    /// or in-flight, we can extract the subset result instead of making a new request.
```
afck:
```suggestion
    /// or in flight, we can extract the subset result instead of making a new request.
```

linera-core/src/client/validator_manager/manager.rs
```diff
@@ -0,0 +1,1268 @@
...
+        // Check cache for exact or subsuming match
+        if let Some(result) = self.cache.get(&key).await {
+            return Ok(result);
+        }
+
+        // Check if there's an in-flight request (exact or subsuming)
+        if let Some(in_flight_match) = self.in_flight_tracker.try_subscribe(&key).await {
+            match in_flight_match {
+                InFlightMatch::Exact(outcome) => match outcome {
+                    SubscribeOutcome::Subscribed(mut receiver) => {
```
afck: Should we add the `peer` to that request as an alternative peer?

linera-core/src/client/requests_scheduler/scheduler.rs
```diff
@@ -0,0 +1,1268 @@
...
+        // Create weights based on normalized scores
+        // Add small epsilon to prevent zero weights
+        let weights: Vec<f64> = top_nodes.iter().map(|(score, _)| score.max(0.01)).collect();
+
+        if let Ok(dist) = WeightedIndex::new(&weights) {
+            let mut rng = rand::thread_rng();
+            let index = dist.sample(&mut rng);
+            Some(top_nodes[index].1.clone())
+        } else {
+            // Fallback to the best node if weights are invalid
```
afck, line 744: This should be unreachable? I'd at least log an error.

linera-core/src/client/validator_manager/manager.rs
```diff
@@ -0,0 +1,1268 @@
...
+        // hold it beyond the lock scope.
+        let notify = {
+            let nodes = self.nodes.read().await;
+            nodes.get(peer_key).map(|info| info.slot_available.clone())
+        };
+
+        // If peer doesn't exist, return immediately
+        let notify = match notify {
+            Some(n) => n,
+            None => return,
```
afck: Should this be unreachable? Then I'd at least log an error.

Or would it make sense to insert the peer instead?

linera-core/src/client/validator_manager/manager.rs
```diff
@@ -0,0 +1,1268 @@
...
+                    execution_count_clone.fetch_add(1, Ordering::SeqCst);
+                    // Wait for signal before completing
+                    rx.await.unwrap();
+                    test_result_ok()
+                })
+                .await
+        });
+
+        // Give the first request time to register as in-flight
+        tokio::time::sleep(Duration::from_millis(10)).await;
```
afck: Then why use `spawn` at all? In 10 ms the request will probably have completed?

linera-core/src/client/requests_scheduler/scheduler.rs
```diff
@@ -0,0 +1,1268 @@
...
+            first_request.await.unwrap();
+        let result2: Result<Vec<ConfirmedBlockCertificate>, NodeError> =
+            second_request.await.unwrap();
+
+        assert!(result1.is_ok());
+        assert!(result2.is_ok());
+        assert_eq!(result1, result2);
+
+        // Operation should only have been executed once (deduplication worked)
+        assert_eq!(execution_count.load(Ordering::SeqCst), 1);
```
afck, line 905: What does this test add compared to `test_cache_hit_returns_cached_result` (and vice versa)?

linera-core/src/client/requests_scheduler/scheduler.rs
```diff
@@ -0,0 +1,1268 @@
...
+        assert!(result.is_ok());
+
+        // All subscriber requests should also complete successfully
+        for handle in handles {
+            assert_eq!(handle.await.unwrap(), result);
+        }
+
+        // Operation should only have been executed once (all requests were deduplicated)
+        assert_eq!(execution_count.load(Ordering::SeqCst), 1);
+    }
```
afck, line 965: Doesn't this just extend the first two tests, i.e. make them obsolete?

linera-core/src/client/requests_scheduler/scheduler.rs
```diff
@@ -0,0 +1,1268 @@
...
+                    test_result_ok()
+                })
+                .await
+        });
+
+        // Give the first request time to register as in-flight
+        tokio::time::sleep(Duration::from_millis(10)).await;
+
+        // Wait for the timeout to elapse
+        tokio::time::sleep(Duration::from_millis(MAX_REQUEST_TTL_MS + 1)).await;
```
afck, line 995: We're trying to avoid `sleep` in tests. I guess it's difficult in this case, but maybe there's a way to use the fake clock?

linera-core/src/client/requests_scheduler/scheduler.rs
```diff
@@ -0,0 +1,1268 @@
...
+    #[tokio::test]
+    async fn test_slot_limiting_blocks_excess_requests() {
+        // Tests the slot limiting mechanism:
+        // - Creates a ValidatorManager with max_requests_per_node = 2
+        // - Starts two slow requests that acquire both available slots
+        // - Starts a third request and verifies it's blocked waiting for a slot (execution count stays at 2)
+        // - Completes the first request to release a slot
+        // - Verifies the third request now acquires the freed slot and executes (execution count becomes 3)
+        // - Confirms all requests complete successfully
+        use linera_base::identifiers::BlobType;
```
afck, line 1032: Should probably be at the start of the file.

linera-core/src/client/validator_manager/node_info.rs
afck: Yes, I think you (or Claude) pretty much re-implemented a semaphore.

linera-core/src/client/mod.rs
```diff
@@ -1248,7 +1314,14 @@ impl<Env: Environment> Client<Env> {
             communicate_concurrently(
                 remote_nodes,
                 async move |remote_node| {
```
afck, line 1256: If this task gets canceled, does that cause any issues, e.g. with the in-flight tracker? (Maybe fixed if we use a Tokio semaphore anyway.)

linera-core/src/client/validator_manager/node_info.rs
deuszx: Refactored to use Semaphore.

linera-core/src/client/validator_manager/node_info.rs
deuszx: Refactored to use Semaphore.

linera-core/src/client/validator_manager/manager.rs
```diff
@@ -393,19 +393,23 @@ impl<Env: Environment> ValidatorManager<Env> {
...
+        let semaphore = node.in_flight_semaphore.clone();
+        let _permit = semaphore.acquire().await.unwrap();
+        drop(nodes);
 
         // Execute the operation
         let result = operation(peer).await;
 
         // Update metrics and release slot
         let response_time_ms = start_time.elapsed().as_millis() as u64;
+        drop(_permit); // Explicitly drop the permit to release the slot
```
afck: (Then the variable doesn't need the underscore.)

linera-core/src/client/validator_manager/manager.rs
```diff
@@ -604,7 +608,7 @@ impl<Env: Environment> ValidatorManager<Env> {
         // Filter nodes that can accept requests and calculate their scores
         let mut scored_nodes = Vec::new();
         for info in nodes.values() {
-            if info.can_accept_request(self.max_requests_per_node).await {
```
afck: Should we already try to acquire permits here? Otherwise there's no guarantee they can still accept requests below.

linera-core/src/client/requests_scheduler/request.rs
deuszx: Good point.

linera-core/src/client/validator_manager/manager.rs
deuszx: not always we follow with the request after this call so I didn't want to do it. Maybe this method should be changed – it's not guaranteed that returned peers will be able to accept a request when the time comes (since the slots can be used in the meantime) – and instead we'd calculate score for everyone and the peers with higher current load will be scored lower due to lack of availble slots:
```
let current_load =
            (self.max_in_flight as f64) - (self.in_flight_semaphore.available_permits() as f64);
let load_score =
            1.0 - (current_load.min(self.max_in_flight as f64) / self.max_in_flight as f64);
```
ie the `load_score` part would be 0 for them.

linera-core/src/client/requests_scheduler/cache.rs
deuszx: How is it simpler? 🤔

linera-core/src/client/validator_manager/manager.rs
deuszx: Removed, will push soon.

linera-core/src/client/requests_scheduler/cache.rs
afck: Right, maybe it isn't, since you have to keep track of the expired keys anyway.

linera-core/src/client/requests_scheduler/node_info.rs
deuszx: The idea was to not prioritise nodes that haven't received enough requests to trust the score.

linera-core/src/client/validator_manager/mod.rs
deuszx: Done  in [Move all ValidatorManager config consts to ClientContextOptions](https://github.com/linera-io/linera-protocol/pull/4752/commits/59e2df164f7cf0c5d0cfaa3d4ae978feffff1de5)

linera-core/src/client/validator_manager/request.rs
deuszx: I think I addressed this in https://github.com/linera-io/linera-protocol/pull/4752/commits/a9c836a67fb06c1ad52c37023d2660771ce34835

linera-core/src/client/requests_scheduler/request.rs
deuszx: I think I addressed this in https://github.com/linera-io/linera-protocol/pull/4752/commits/a9c836a67fb06c1ad52c37023d2660771ce34835

linera-core/src/client/requests_scheduler/node_info.rs
deuszx: Since it's now configurable I'm gonna leave it as-is. We can adjust it with different args.

linera-core/src/client/validator_manager/manager.rs
deuszx: Not really – `REQUESTS_CACHE_HIT` is incremented when the request (key) is found in the response cache and `REQUESTS_CACHE_DEDUPLICATION` is incremented when the request is deduplicated with another one in-flight.

linera-core/src/client/validator_manager/manager.rs
deuszx: Done in [Move all ValidatorManager config consts to ClientContextOptions](https://github.com/linera-io/linera-protocol/pull/4752/commits/59e2df164f7cf0c5d0cfaa3d4ae978feffff1de5)

linera-core/src/client/validator_manager/manager.rs
deuszx: Done in [Remove can_accept_request](https://github.com/linera-io/linera-protocol/pull/4752/commits/d4c0f2c1276282a1bd2bac30b654a91460453f86)

linera-core/src/client/validator_manager/manager.rs
deuszx: This is removed now with the use of Semaphore.

linera-core/src/client/validator_manager/manager.rs
deuszx: This is used as adjective here, I think 🤔 A request is _cached_ or _in-flight_ (in progress).

linera-core/src/client/requests_scheduler/scheduler.rs
deuszx: I'm gonna remove the clamping altogether and just assert it's between 0 and 1.

linera-core/src/client/requests_scheduler/scheduler.rs
deuszx: It's used in this single test 🤔

linera-core/src/client/mod.rs
deuszx: Yes, hopefully when the task with the permit is dropped the semaphore permit is released.

deuszx: but good point!

linera-core/src/client/validator_manager/manager.rs
deuszx: Done in [Typos and small corrections](https://github.com/linera-io/linera-protocol/pull/4752/commits/22acd1673824df1153592f5cd48d1c3ed359dac4).

linera-core/src/client/validator_manager/manager.rs
deuszx: Done in [Typos and small corrections](https://github.com/linera-io/linera-protocol/pull/4752/commits/22acd1673824df1153592f5cd48d1c3ed359dac4).

linera-core/src/client/validator_manager/manager.rs
deuszx: Done in [Typos and small corrections](https://github.com/linera-io/linera-protocol/pull/4752/commits/22acd1673824df1153592f5cd48d1c3ed359dac4).

linera-core/src/client/validator_manager/manager.rs
deuszx: Done in [Typos and small corrections](https://github.com/linera-io/linera-protocol/pull/4752/commits/22acd1673824df1153592f5cd48d1c3ed359dac4).

linera-core/src/client/validator_manager/manager.rs
deuszx: Done in [Typos and small corrections](https://github.com/linera-io/linera-protocol/pull/4752/commits/22acd1673824df1153592f5cd48d1c3ed359dac4).

linera-core/src/client/requests_scheduler/scheduler.rs
deuszx: Done in [Move alpha smoothing factor to client options](https://github.com/linera-io/linera-protocol/pull/4752/commits/8e098765f391f662d3bf17d5d3bbb9f234468d3c).

linera-core/src/client/validator_manager/node_info.rs
deuszx: Done in [Move alpha smoothing factor to client options](https://github.com/linera-io/linera-protocol/pull/4752/commits/8e098765f391f662d3bf17d5d3bbb9f234468d3c).

linera-core/src/client/validator_manager/in_flight_tracker.rs
deuszx: You're right.

deuszx: Done in https://github.com/linera-io/linera-protocol/pull/4752/commits/bbbd690afe280af26c3f672fd2c7e7fcbf702d08

linera-core/src/client/validator_manager/in_flight_tracker.rs
deuszx: Done in https://github.com/linera-io/linera-protocol/pull/4752/commits/aab98390cd97248e8106a8cb5b8976be03d1258b

linera-core/src/client/requests_scheduler/scheduler.rs
deuszx: Done in https://github.com/linera-io/linera-protocol/pull/4752/commits/aab98390cd97248e8106a8cb5b8976be03d1258b

linera-core/src/client/validator_manager/manager.rs
deuszx: It should already be added in `with_peer`. If the request is in flight then it will be added.

linera-core/src/client/validator_manager/manager.rs
deuszx: It won't be completed b/c it's waiting for the message in `rx.await.unwrap();`.

linera-core/src/client/requests_scheduler/cache.rs
deuszx: I did rewrite it to use `retain` and keep track of # keys being removed. I think I misunderstood your comment at first.

linera-core/src/client/requests_scheduler/scheduler.rs
deuszx: The first one tests that we return cached result. The second tests that we deduplicate requests.

linera-core/src/client/requests_scheduler/scheduler.rs
deuszx: Not really. First tests servng results from the cache, second request deduplication, third - similar a bit to the second - but makes sure ALL subscribers are notified

linera-core/src/client/requests_scheduler/scheduler.rs
deuszx: it's 50ms+1, I think it's fine.

linera-core/src/client/mod.rs
deuszx: Done in [Add RemoteNode API to ValidatorManager; move away from closures](https://github.com/linera-io/linera-protocol/pull/4752/commits/ba357d3585636369ab8a995e48371751f9f8046b)

CLI.md
```diff
@@ -184,6 +184,24 @@ Client implementation and command-line tool for the Linera blockchain
 * `--max-joined-tasks <MAX_JOINED_TASKS>` — Maximum number of tasks that can are joined concurrently in the client
 
   Default value: `100`
+* `--max-in-flight-requests <MAX_IN_FLIGHT_REQUESTS>` — Maximum concurrent requests per validator node
+
+  Default value: `100`
+* `--max-accepted-latency-ms <MAX_ACCEPTED_LATENCY_MS>` — Maximum expected latency in milliseconds for score normalization
+
+  Default value: `5000`
+* `--cache-ttl-sec <CACHE_TTL_SEC>` — Time-to-live for cached responses in seconds
```
afck: Should we make that milliseconds, too? We use ms for almost all timing options.

linera-client/src/client_options.rs
```diff
@@ -186,6 +186,58 @@ pub struct ClientContextOptions {
...
+        default_value_t = linera_core::client::validator_manager::MAX_REQUEST_TTL_MS,
+        env = "LINERA_VALIDATOR_MANAGER_MAX_REQUEST_TTL_MS"
+    )]
+    pub max_request_ttl_ms: u64,
+
+    /// Smoothing factor for Exponential Moving Averages (0 < alpha < 1)
+    /// Higher values give more weight to recent observations
+    /// Typical values are between 0.01 and 0.5
+    /// A value of 0.1 means that 10% of the new observation is considered
+    /// and 90% of the previous average is retained
```
afck:
```suggestion
    /// Smoothing factor for Exponential Moving Averages (0 < alpha < 1).
    /// Higher values give more weight to recent observations.
    /// Typical values are between 0.01 and 0.5.
    /// A value of 0.1 means that 10% of the new observation is considered
    /// and 90% of the previous average is retained.
```

linera-core/src/client/mod.rs
```diff
@@ -362,17 +372,20 @@ impl<Env: Environment> Client<Env> {
...
-                .await,
+                &self
+                    .validator_manager
+                    .download_blobs(remote_nodes, blob_ids, self.options.blob_download_timeout)
+                    .await?
+                    .ok_or_else(|| {
+                        ChainClientError::RemoteNodeError(NodeError::BlobsNotFound(
+                            blob_ids.to_vec(),
+                        ))
+                    })?,
```
afck: I'd put that in a local variable before we call `store_blobs`.

linera-core/src/client/validator_manager/request.rs
```diff
@@ -0,0 +1,576 @@
...
+    fn from(cert: ConfirmedBlockCertificate) -> Self {
+        RequestResult::Certificate(Box::new(cert))
+    }
+}
+
+impl From<RequestResult> for Option<Blob> {
+    fn from(result: RequestResult) -> Self {
+        match result {
+            RequestResult::Blob(blob) => blob,
+            _ => panic!("Cannot convert RequestResult to Option<Blob>"),
```
afck: Why not use `TryFrom` then?

linera-core/src/client/validator_manager/request.rs
```diff
@@ -0,0 +1,576 @@
...
+        if in_flight_req_heights.is_empty() {
+            return false; // An empty set cannot subsume non-empty
+        }
+
+        in_flight_req_heights
+            .first()
+            .expect("heights1 is not empty")
+            <= new_req_heights.first().expect("heights2 is not empty")
+            && in_flight_req_heights.last().expect("heights1 is not empty")
+                >= new_req_heights.last().expect("heights2 is not empty")
```
afck: That assumes that the vectors of heights are contiguous?

linera-core/src/client/validator_manager/request.rs
```diff
@@ -0,0 +1,576 @@
...
+            return Some(RequestResult::Certificates(vec![])); // Nothing requested
+        }
+        let requested_start = requested_heights.first().unwrap();
+        let requested_end = requested_heights.last().unwrap();
+
+        // Filter certificates to only those at the requested heights
+        let filtered: Vec<_> = certificates
+            .iter()
+            .filter(|cert| {
+                &cert.value().height() >= requested_start && &cert.value().height() <= requested_end
```
afck: That also assumes contiguous ranges?

linera-core/src/client/requests_scheduler/request.rs
```diff
@@ -0,0 +1,576 @@
...
+        let chain_id = ChainId(CryptoHash::test_hash("chain1"));
+        let req1 = RequestKey::Certificates {
+            chain_id,
+            heights: vec![BlockHeight(12), BlockHeight(13)],
+        };
+        let req2 = RequestKey::Certificates {
+            chain_id,
+            heights: vec![BlockHeight(12), BlockHeight(14)],
+        };
+        assert!(!req1.subsumes(&req2));
```
afck, line 227: And `assert!(!req2.subsumes(&req1))`?

linera-core/src/client/requests_scheduler/in_flight_tracker.rs
```diff
@@ -0,0 +1,209 @@
...
+        if let Some(entry) = in_flight.get(key) {
+            let elapsed = Instant::now().duration_since(entry.started_at);
+
+            if elapsed <= self.timeout {
+                return Some(InFlightMatch::Exact(Subscribed(entry.sender.subscribe())));
+            }
+        }
+
+        // Sometimes a request key may not have the exact match but may be subsumed by a larger one.
+        for (in_flight_key, entry) in in_flight.iter() {
```
afck, line 64: Not sure how many in-flight requests we are expecting, but if we had the map by chain ID and/or type of request, we wouldn't have to iterate over all of them.

linera-core/src/client/validator_manager/in_flight_tracker.rs
```diff
@@ -0,0 +1,209 @@
...
+    ///
+    /// # Arguments
+    /// - `key`: The request key to look up
+    ///
+    /// # Returns
+    /// - `Vec<N>`: List of alternative peers (empty if no entry exists)
+    pub(super) async fn get_alternative_peers(&self, key: &RequestKey) -> Option<Vec<N>> {
+        let in_flight = self.entries.read().await;
+
+        if let Some(entry) = in_flight.get(key) {
```
afck: (`?` works for `Option`s, too.)

linera-core/src/client/mod.rs
deuszx: OK, but that was already here – I just moved it around.

linera-core/src/client/validator_manager/request.rs
deuszx: Does it? It assumes they are ordered but not contiguous 🤔

CLI.md
deuszx: I thought that we'd never need anything less than 1 second but I agree with you we should be consistent. Will change.

linera-core/src/client/validator_manager/cache.rs
```diff
@@ -0,0 +1,387 @@
...
+    cached_at: Instant,
+}
+
+/// Cache for request results with TTL-based expiration and LRU eviction.
+///
+/// This cache supports:
+/// - Exact match lookups
+/// - Subsumption-based lookups (larger requests can satisfy smaller ones)
+/// - TTL-based expiration
+/// - Size-based LRU eviction
```
afck: It's not size-based, is it? Below we just use `cache.len() < self.max_cache_size`.

linera-core/src/client/validator_manager/request.rs
deuszx: My reasoning was that this is more of a "programmer error" – i.e. we should never return blob when certificates were requested (etc.) so I didn't want to handle that as part of the `Err(_)` variant on the caller side.

linera-core/src/client/requests_scheduler/scheduler.rs
```diff
@@ -0,0 +1,1260 @@
...
+                    key = ?key,
+                    subsumed_by = ?subsuming_key,
+                        "deduplicating request (subsumption) - joining larger in-flight request"
+                    );
+                    #[cfg(with_metrics)]
+                    metrics::REQUEST_CACHE_DEDUPLICATION.inc();
+                    // Wait for result from the subsuming request
+                    match receiver.recv().await {
+                        Ok(result) => {
+                            match result.as_ref() {
```
afck, line 620: Why `as_ref()`? Can't we merge the two `match`es and do `Ok(Ok(res))` and `Ok(Err(e))`?

linera-core/src/client/validator_manager/request.rs
afck: But `subsumes` should only return `true` if one request is a superset of the other, e.g. `[3, 4, 5, 6]` subsumes `[3, 5]`. But `[3, 4, 6]` does not.

linera-core/src/client/validator_manager/request.rs
afck: I feel like panics in a `From` impl are a footgun. If I saw that `impl From<Fruit> for Apple` I wouldn't think twice about using it, and then the durian would explode in my face…

deuszx: Here the panic would mean that we're mapping incorrectly, that the respective `From` that goes the other direction is incorrect and/or that the logic in the node is invalid and returns blob for a certificate request. Each of those error variants will require different handling. Who is responsible for this? Should I return to the caller information that "node return blob when you requested certificate" or that "no certificates were available"?

linera-core/src/client/validator_manager/cache.rs
deuszx: "length-based"?

linera-core/src/client/requests_scheduler/scheduler.rs
deuszx: it's an `Arc` and I need to extract the `RequestResult` that's inside.

deuszx: Also, I wanted to explicitly have the branches separate b/c one is about sender dropping the receiver and the other is about the request/result error. Totally different control flows.

linera-core/src/client/validator_manager/cache.rs
afck: Or just "LRU eviction"?

linera-core/src/client/validator_manager/request.rs
afck: I'd still prefer an `expect` or `unwrap` at the call site, rather than a `From` impl that panics.

linera-core/src/client/requests_scheduler/in_flight_tracker.rs
deuszx: I'll see how much work is that and do it.

linera-core/src/client/validator_manager/request.rs
deuszx: I'm doing the change 👍

linera-core/src/client/validator_manager/request.rs
deuszx: You're right 😬

linera-core/src/client/requests_scheduler/in_flight_tracker.rs
afck: (Doesn't have to be in this PR! Not a blocker.)

linera-core/src/client/validator_manager/request.rs
deuszx: Done in [Fix subsumes implementation](https://github.com/linera-io/linera-protocol/pull/4752/commits/6c8f9fc430479753f76c747a3cc9a49bfda54392).

linera-core/src/client/validator_manager/request.rs
deuszx: Done in [Fix subsumes implementation](https://github.com/linera-io/linera-protocol/pull/4752/commits/6c8f9fc430479753f76c747a3cc9a49bfda54392).

linera-core/src/client/mod.rs
deuszx: Done in [Put blobs in the local variable before storing](https://github.com/linera-io/linera-protocol/pull/4752/commits/d9d39903a58a6714207e71ca6b65daa218bf226b).

linera-core/src/client/validator_manager/request.rs
deuszx: Done in [Rewrite conversion from RequestResult to use TryFrom](https://github.com/linera-io/linera-protocol/pull/4752/commits/105c18519509c0f2c2d156d5d3bda2cc851270d6)

linera-core/src/client/requests_scheduler/in_flight_tracker.rs
deuszx: OK, I'd do it in a separate PR – this one is already quite big.

linera-core/src/client/requests_scheduler/request.rs
```diff
@@ -139,20 +147,9 @@ impl SubsumingKey<RequestResult> for super::request::RequestKey {
...
-
-        in_flight_req_heights
-            .first()
-            .expect("heights1 is not empty")
-            <= new_req_heights.first().expect("heights2 is not empty")
-            && in_flight_req_heights.last().expect("heights1 is not empty")
-                >= new_req_heights.last().expect("heights2 is not empty")
+        new_req_heights
+            .iter()
+            .all(|height| in_flight_req_heights.contains(height))
```
afck: (Not a blocker, but as I said before: This could be $`O(n)`$ if we made use of the fact that they are both ordered.)

linera-core/src/client/requests_scheduler/request.rs
```diff
@@ -170,32 +167,24 @@ impl SubsumingKey<RequestResult> for super::request::RequestKey {
...
-        }
-        if filtered.first().unwrap().value().height() != *requested_start
-            || filtered.last().unwrap().value().height() != *requested_end
-        {
-            return None; // Missing some requested heights
+        let mut collected = vec![];
+        while let Some(height) = requested_heights.first() {
+            if let Some(pos) = certificates
+                .iter()
+                .position(|cert| &cert.value().height() == height)
```
afck: (Same here.)

afck (approved):

linera-core/src/client/requests_scheduler/request.rs
deuszx: We kindof already are doing that, aren't we - at least in terms of length of `requested_heights`? `first` is `O(1)` and `certificates` are ordered as well so if the requested height exists in `certificates` then it will be the found earlier than others.

afck: But we go through all of `requested_heights`, so the later ones will iterate through most of `certificates`. The algorithm is currently $`O(n^2)`$ and it could be $`O(n)`$.

It would be $`O(n)`$ if we removed everything up to `pos` from `certificates`, but that's only borrowed. So I'd rather use an iterator. I can make a PR later.

deuszx: I hope I did that in [Rewrite try_extract_result and subsumes to O(n) complexity.](https://github.com/linera-io/linera-protocol/pull/4752/commits/8809c8409772695cf8c892abd2fa55e9e064c1ff).

afck: Great, thanks! :+1:

linera-core/src/client/requests_scheduler/in_flight_tracker.rs
ma2bd: Is this a blocker for the backport?

afck: I think @deuszx measured that this PR already improves performance a lot on the testnet, so it seems like the CPU overhead is small compared to the bandwidth gains?

deuszx: I think the iteration overhead, compared to the subsumption test is negligible - we are iterating, yes, but we quickly throw away pairs that cannot possible subsume each other (b/c the chain ID is empty or request type doesn't match).

PR #4755, opened 2025-10-08
`linera_core::client`: batch downloading of missing blobs

Twey:
## Motivation

Currently, when we synchronize a chain, even though we receive the certificates in a batch to `process_certificates`, we handle them one by one at the local node level, and if a certificate is missing blobs we stop, download the blobs for that certificate, then retry, making the download of the blobs sequential.  This makes startup time for the client linear in the number of certificates-with-blobs present in its initial chains (notably, the admin chain).

## Proposal

We already have an ahead-of-time indicator of which blobs will be required by the certificates, so there's no need to download them one at a time.  If we don't have some blobs marked as required by the certificate batch, try to download them (concurrently) before proceeding to process the batch.  ~Thereafter, `BlobsNotFound` when processing the batch is a hard error.~ `required_blob_ids()` is conservative, so we still need to download blobs and retry if we get a `BlobsNotFound` error thereafter.

## Test Plan

CI.

## Release Plan

    - be released in a new SDK.

## Links

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -420,7 +420,7 @@ impl<Env: Environment> Client<Env> {
             }
         };
         for certificate in certificates {
-            last_info = Some(self.handle_certificate(Box::new(certificate)).await?.info);
+            last_info = Some(self.handle_certificate(certificate).await?.info);
         }
         // Now download the rest in batches from the remote node.
         while next_height < stop {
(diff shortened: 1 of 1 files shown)


linera-core/src/client/mod.rs
```diff
@@ -451,23 +451,27 @@ impl<Env: Environment> Client<Env> {
...
+            .iter()
+            .flat_map(|certificate| certificate.value().required_blob_ids())
+            .collect();
 
-            if let Err(LocalNodeError::BlobsNotFound(blob_ids)) = &result {
-                let blobs =
-                    futures::stream::iter(blob_ids.iter().copied().map(|blob_id| async move {
+        match self.local_node.read_blob_states_from_storage(&required_blob_ids).await {
+            Ok(_) => (),
+            Err(LocalNodeError::BlobsNotFound(blob_ids)) => {
```
afck, line 482: ~I'd prefer to make sure `self.handle_certificate` below returns exactly this error. Why does it currently return only a part of the blob IDs? We should have this check in the chain worker, when processing a confirmed certificate.~

Ah, I see, it's for all certificates together!

afck (approved):

linera-core/src/client/mod.rs
Twey: Right, it's to extend the plurality of `process_certificate*s*` (which works on a batch received from the same validator) to the blob downloads.  Maybe we could make `LocalNode::handle_certificate*s*` do this instead, but it would be a new function — and it would still have to retry sequentially anyway, since `required_blob_ids()` aren't necessarily _sufficient_ blob IDs :)

deuszx:
[We already had](https://github.com/linera-io/linera-protocol/blob/testnet_conway/linera-core/src/remote_node.rs#L313) `download_blobs` (plural) that parallelized this – why duplicate the logic?

Twey:
Ah, the logic was already duplicated here — I just moved it into a function.  Probably we should integrate the improvements to this logic directly into `RemoteNode::download_blobs` instead.

PR #4763, opened 2025-10-09
Add `try_load_all_entries` to `CollectionView`.

MathieuDutSik:
## Motivation

Just like for `ReentrantCollectionView`, we can load all entries at once in `CollectionView`.

## Proposal

The code was added in a straightforward way.

## Test Plan

The test has been to the CI. The consequence is that the `try_load_all_entries` has been tested for `ReentrantCollectionView`. This led to the removal of the `Q` types that was not used.

## Release Plan

## Links

None.

### DIFF
--- a/linera-views/src/views/collection_view.rs
+++ b/linera-views/src/views/collection_view.rs
@@ -435,9 +435,9 @@ impl<W: View> ByteCollectionView<W::Context, W> {
     pub async fn try_load_all_entries(
         &self,
     ) -> Result<Vec<(Vec<u8>, ReadGuardedView<W>)>, ViewError> {
+        let updates = self.updates.read().await; // Acquire the read lock to prevent writes.
         let short_keys = self.keys().await?;
         let mut results = Vec::with_capacity(short_keys.len());
-        let updates = self.updates.read().await;
 
         let mut keys_to_load = Vec::new();
         let mut keys_to_load_metadata = Vec::new();
@@ -546,7 +546,7 @@ impl<W: View> ByteCollectionView<W::Context, W> {
     /// # })
     /// ```
     pub async fn contains_key(&self, short_key: &[u8]) -> Result<bool, ViewError> {
-        let updates = self.updates.write().await;
+        let updates = self.updates.read().await;
         Ok(match updates.get(short_key) {
             Some(entry) => match entry {
                 Update::Set(_view) => true,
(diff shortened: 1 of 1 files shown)


linera-views/src/views/collection_view.rs
```diff
@@ -415,6 +415,84 @@ impl<W: View> ByteCollectionView<W::Context, W> {
...
+                    };
+                    let updates = self.updates.read().await;
+                    let view = ReadGuardedView::Loaded {
+                        updates,
+                        short_key: short_key.clone(),
+                    };
+                    results.push((short_key.clone(), Some(view)));
+                }
+                None => {
+                    assert!(!self.delete_storage_first); // otherwise it would not be in keys.
```
ma2bd: I don't quite get the comment. Can you rephrase?

MathieuDutSik: If we are in the `None` case, then this implies that we need to access the storage.
We know that the key exists because otherwise, it would not appear in the `self.keys()`.
But if that is so, then we are in `!self.delete_storage_first`.

So, here we write down a simple invariant, but I agree this is something that we could remove.

ma2bd: Still rephrase the comment. What is "it" in the sentence? Did you mean `short_keys`?

MathieuDutSik: Yes, rephrased and detailed.

linera-views/src/views/collection_view.rs
```diff
@@ -415,6 +415,87 @@ impl<W: View> ByteCollectionView<W::Context, W> {
...
+        let mut results = Vec::with_capacity(short_keys.len());
+        let updates = self.updates.read().await;
+
+        let mut keys_to_load = Vec::new();
+        let mut keys_to_load_metadata = Vec::new();
+        for (position, short_key) in short_keys.iter().enumerate() {
+            match updates.get(short_key) {
+                Some(update) => {
+                    let Update::Set(_) = update else {
+                        unreachable!();
```
ma2bd, line 448: For the records, the fact that `short_keys` and `updates` are consistent is true but not so straightforward: I believe `self.keys()` cannot be modified before we call `self.updates.read()` only because we hold a `&self` reference and we are careful about interior mutability in every other method.

ma2bd (approved):
I'm assuming we're following the same code patterns as before.

linera-views/src/views/collection_view.rs
MathieuDutSik: That was indeed an insightful comment.

In principle, in the nanosecond between a `self.keys()` and an `updates.read()`, we could have a write that acquire it. The solution is to do the `let updates = self.updates.read().await` upfront.

But doing that creates deadlocks. This was because we had some `updates.write()` where a `read()` was adequate.

PR #4764, opened 2025-10-09
Show locked; filter by weight; fetch all missing senders.

afck:
## Motivation

It's sometimes useful to see the current locked block of each validator, e.g. when debugging why a chain isn't making progress.

It's also helpful to focus on only the validators with a lot of voting weight, since they matter most in that case.

For proposals we learn about from validators, we only fetch one missing sender block.

## Proposal

Show some information about the locking block, if present. Add a `--min-votes` option. Fetch _all_ missing sender blocks.

## Test Plan

I'm using this for debugging on `testnet_conway`. With the fix for fetching missing senders I managed to get the locked block into my local node's chain manager.

## Release Plan

- These changes should be ported to `testnet_conway` and released in a new SDK.

## Links

### DIFF
--- a/linera-client/src/client_context.rs
+++ b/linera-client/src/client_context.rs
@@ -785,7 +785,7 @@ impl<Env: Environment, W: Persist<Target = Wallet>> ClientContext<Env, W> {
             .genesis_config_hash);
         let chain_info = self
             .make_chain_client(chain_id)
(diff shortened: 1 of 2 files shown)


afck:
This is just one chain info query to the local node, isn't it?

PR #4767, opened 2025-10-09
Add support for "null" in case of missing values and add integration test for `GraphQL`

MathieuDutSik:
## Motivation

GraphQL is the canonical way to access Wasm application state on Linera. Therefore, it is of paramount
importance of having excellent support for GraphQL in the `linera-views`.

The canonical way of handling missing values in a query is to return `null`. Ignoring it and pretending that it was
not queried is not the recommended way. This is for example visible on [GraphQL Search and Filter – How to search and filter results with GraphQL](https://www.apollographql.com/blog/how-to-search-and-filter-results-with-graphql).
In this exposition, we see that the GraphQL schema types are
```GraphQL
type Query {
  album(id: ID!): Album
  albums(genre: Genre, ids: [ID!]): [Album]! 
}
```

The meaning of that schema is that the function `album` returns something that may be null, and that `albums` returns a list of entries that may be `null`. This is because by default in GraphQL a variable can be nulled. If not it has a postfix of "!".

The GraphQL standard does not say that if a request is failing to return null. However, if returning null is not allowed, then in case of a missing key, you should return null at a higher level. See [GraphQL, Section 7.1.2](https://spec.graphql.org/June2018/#sec-Errors). Therefore for our case, it is better to return a null value in case of missing key.

## Proposal

The `Entry` is changed in order to have a value that is an `Option<V>`. The change go surprisingly smoothly.

## Test Plan

The CI.

Due to the importance of GraphQL in Linera applications, a specific smart contract has been created for which we can test the 

## Release Plan

This changes are also relevant to the Game Of Life test.

## Links

See above.

### DIFF
--- a/linera-sdk/tests/fixtures/Cargo.toml
+++ b/linera-sdk/tests/fixtures/Cargo.toml
@@ -6,6 +6,7 @@ members = [
     "create-and-call",
     "event-emitter",
     "event-subscriber",
+    "graph-ql-queries",
     "meta-counter",
     "publish-read-data-blob",
     "time-expiry",
@@ -25,6 +26,7 @@ counter = { path = "../../../examples/counter" }
 counter-no-graphql = { path = "../../../examples/counter-no-graphql" }
 create-and-call = { path = "./create-and-call" }
 event-emitter = { path = "./event-emitter" }
+graph-ql-queries = { path = "./graph-ql-queries" }
 meta-counter = { path = "./meta-counter" }
 publish-read-data-blob = { path = "./publish-read-data-blob" }
 time-expiry = { path = "./time-expiry" }
--- a/linera-sdk/tests/fixtures/graph-ql-queries/tests/graph_ql_operation.rs
+++ b/linera-sdk/tests/fixtures/graph-ql-queries/tests/graph_ql_operation.rs
@@ -1,0 +1,95 @@
+// Copyright (c) Zefchain Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Integration tests for the Matching Engine application
+
+#![cfg(not(target_arch = "wasm32"))]
+
+use graph_ql_queries::{GraphQlQueriesAbi, GraphQlQueriesOperation};
+use linera_sdk::test::{QueryOutcome, TestValidator};
+
+#[tokio::test]
+async fn test_queries() {
+    let (validator, module_id) =
+        TestValidator::with_current_module::<GraphQlQueriesAbi, (), ()>().await;
+
+    let mut chain = validator.new_chain().await;
+
+    let application_id = chain.create_application(module_id, (), (), vec![]).await;
+
+    let operation1 = GraphQlQueriesOperation::SetRegister { value: 124 };
+    let operation2 = GraphQlQueriesOperation::InsertMapString {
+        key: "a".into(),
+        value: 91,
+    };
+    let operation3 = GraphQlQueriesOperation::InsertCollString {
+        key: "a".into(),
+        value: 91,
+    };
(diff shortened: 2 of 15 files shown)


linera-sdk/tests/fixtures/Cargo.toml
```diff
@@ -20,6 +21,7 @@ contract-call = { path = "./create-and-call" }
 counter = { path = "../../../examples/counter" }
 counter-no-graphql = { path = "../../../examples/counter-no-graphql" }
 create-and-call = { path = "./create-and-call" }
+graph-ql-queries = { path = "./graph-ql-queries" }
```
Twey, line 29: Nit: let's be consistent with the spelling `graphql`.

linera-sdk/tests/fixtures/graph-ql-queries/tests/graph_ql_operation.rs
```diff
@@ -0,0 +1,95 @@
...
+                .with_operation(application_id, operation1)
+                .with_operation(application_id, operation2)
+                .with_operation(application_id, operation3)
+                .with_operation(application_id, operation4);
+        })
+        .await;
+
+    // READ1
+
+    for (query, expected_response) in [
```
Twey, line 47: It would be nice to use `insta` here (with the snapshot identifier being the query).

linera-views/src/graphql.rs
```diff
@@ -36,7 +36,7 @@ pub struct Entry<
     V: async_graphql::OutputType + Send + Sync,
 > {
     pub key: K,
-    pub value: V,
+    pub value: Option<V>,
```
Twey, line 39: I'm not really convinced that map entry values should _always_ be optional.  E.G. when iterating over sets of entries (perhaps with `entries` with a filter other than `keys`) we _should_ just skip missing entries, so the value should be non-optional.

MathieuDutSik: Right now, the object `filters` does not implement any significant filter. You have only two options: either `None` with all keys returned or the keys being selected.

Maybe, the `filters` are meant to be extended to a larger feature. If so, I have not seen a description of it as a GitHub issue. But then, the real question is what actually would be possible as `filters`:
* What we are considering is the `CollectionView` / `ReentrantCollectionView` function. The input type I and view type V are generic. So, how exactly can we input a meaningful filter?
* The page [GraphQL search and filter](https://www.apollographql.com/blog/how-to-search-and-filter-results-with-graphql) gives some examples like `(a) => a.genre == genre` with `a` of the output type. How could we implement a filtering function on the view? 
* It does not appear that the `async-graphql` library has any functionality for filtering. So, I do not know if we could implement filtering in a generic way without some major engineering effort.

I have three other reasons for returning `null` in case of missing values:
* The change is putting the behavior of `CollectionView` / `ReentransCollectionView` in line with `MapView`. Right now the `MapView` is returning `null` for missing keys. I think it is important to have `MapView<K,V>` as near to `CollectionView<K,RegisterView<V>>` as possible.
* The change of returning `null` on missing values is extending the functionality. It is not breaking anything. All of the tests that were passing are still passing.
* In the page [GraphQL search and filter](https://www.apollographql.com/blog/how-to-search-and-filter-results-with-graphql) the schema for the filtering is "albums(input: AlbumsInput): [Album]! " and so it does return `null`on missing keys.

All that being said, I do not see this work as a final work on GraphQL for `linera-views`. For example, cursors, intervals, or similar are definitely needed. It is possible to add those functionalities since they operate in a generic way. But we have to be clear about what is realistic to implement in `linera-views`.

ma2bd:
converting stale PR to draft

PR #4769, opened 2025-10-09
Handle dropped actors without crashing the RPC thread

ma2bd:
## Motivation

Crashing the RPC thread was a bad idea because it doesn't crash the process itself. Let's try another approach

## Proposal

* Fail the RPC query
* Remove the dropped actor from the actor cache

## Test Plan

CI + TBD?

## Release Plan

    - be released in a new SDK,

### DIFF
--- a/linera-core/src/worker.rs
+++ b/linera-core/src/worker.rs
@@ -239,9 +239,15 @@ pub enum WorkerError {
     #[error("Missing network description")]
     MissingNetworkDescription,
     #[error("ChainWorkerActor for chain {chain_id} stopped executing unexpectedly: {error}")]
-    ChainActorSendError { chain_id: ChainId, error: String },
+    ChainActorSendError {
+        chain_id: ChainId,
+        error: Box<dyn std::error::Error + Send + Sync>,
+    },
     #[error("ChainWorkerActor for chain {chain_id} stopped executing without responding: {error}")]
-    ChainActorRecvError { chain_id: ChainId, error: String },
+    ChainActorRecvError {
+        chain_id: ChainId,
+        error: Box<dyn std::error::Error + Send + Sync>,
+    },
 }
 
 impl From<ChainError> for WorkerError {
@@ -808,7 +814,7 @@ where
                 // The actor endpoint was dropped. Better luck next time!
                 Err(WorkerError::ChainActorRecvError {
                     chain_id,
-                    error: e.to_string(),
+                    error: Box::new(e),
                 })
             }
             Ok(response) => response,
@@ -844,7 +850,7 @@ where
             // The actor was dropped. Give up without (re-)inserting the endpoint in the cache.
             return Err(WorkerError::ChainActorSendError {
                 chain_id,
-                error: e.to_string(),
+                error: Box::new(e),
             });
         }
 


linera-core/src/worker.rs
```diff
@@ -764,13 +768,26 @@ where
         let chain_actor = self.get_chain_worker_endpoint(chain_id).await?;
         let (callback, response) = oneshot::channel();
 
-        chain_actor
-            .send((request_builder(callback), tracing::Span::current()))
-            .expect("`ChainWorkerActor` stopped executing unexpectedly");
+        if let Err(e) = chain_actor.send((request_builder(callback), tracing::Span::current())) {
+            // The actor endpoint was dropped. Let's clear it from the cache.
+            self.clear_chain_worker_endpoint(chain_id);
```
afck: It may be possible at this point that some future somewhere still has a reference of the shared chain state view.

It's maybe not a blocker for now, but we should make sure soon that the shared views are only used for GraphQL.

linera-core/src/worker.rs
```diff
@@ -858,6 +875,16 @@ where
         }
     }
 
+    /// Clear any endpoint to a [`ChainWorkerActor`] from the cache for the given chain.
+    #[instrument(level = "trace", target = "telemetry_only", skip(self), fields(
```
Twey: Why should this be `telemetry_only`?

I think the default configuration for `instrument` should just be `level = "debug"`.  Remember these don't cause new log lines unless execution tracing is explicitly enabled with `RUST_LOG_SPAN_EVENTS`, but add potentially helpful debug information to events that occur within the span.

linera-core/src/worker.rs
```diff
@@ -764,13 +768,26 @@ where
...
-        response
-            .await
-            .expect("`ChainWorkerActor` stopped executing without responding")
+        match response.await {
+            Err(e) => {
+                // The actor endpoint was dropped. Let's clear it from the cache.
+                self.clear_chain_worker_endpoint(chain_id);
+                Err(WorkerError::ChainActorRecvError {
+                    chain_id,
+                    error: e.to_string(),
```
Twey: Please never convert errors to strings as it loses the entire backtrace, and then debugging gets hard :)  If you need a dynamic error, which I don't think you do here, you can use `Box<dyn std::error::Error + Send>`.

linera-core/src/worker.rs
deuszx: As @ndr-ds axplained to me:

> Basically if you have `telemetry_only` it just means that you’re gonna send the traces to Tempo for sure. Even if you’re below the log level. But if you’re in or above the log level, it will still also print to the screen

So we can have both `debug` and `telemtry_only`.

linera-core/src/worker.rs
```diff
@@ -764,13 +768,26 @@ where
...
+                })
+            }
+            Ok(response) => response,
```
Twey: Should we also drop it from the cache if `response.is_err()`?

Twey: (no, probably not — those errors are recoverable)

linera-core/src/worker.rs
ndr-ds: Yes, sorry, I need to change the name of this to `telemetry` or something like that, the `_only` was supposed to have been removed and I missed it 🤦🏻‍♂️

Twey: Gotcha :)  Shouldn't we just always send all `TRACE`s (and above) to Tempo though?

linera-core/src/worker.rs
ma2bd: I had an issue with a generic type. Let me try again

linera-core/src/worker.rs
ndr-ds: There were some of these `instrument` annotations with `TRACE` for example. Since we run with log level `INFO` by default, if I sent to Tempo just based on the log level, it wouldn't send those spans. And we technically want all the spans 😅 so we have as much information as possible.
So I needed a way to disassociate the log level with it being sent to Tempo, and that's why I did the `target` thing, and added `telemetry_only` (soon to be `tempo` maybe) everywhere. To guarantee that even if we don't match the log level, we still send the spans to Tempo.
Hopefully the explanation makes sense 😅

linera-core/src/worker.rs
```diff
@@ -761,39 +763,34 @@ where
...
-        let chain_actor = self.get_chain_worker_endpoint(chain_id).await?;
-        let (callback, response) = oneshot::channel();
-
-        chain_actor
-            .send((request_builder(callback), tracing::Span::current()))
-            .expect("`ChainWorkerActor` stopped executing unexpectedly");
+        let (response, receiver) = {
+            let mut chain_workers = self.chain_workers.lock().unwrap();
+
+            let (sender, new_receiver) = if let Some(endpoint) = chain_workers.remove(&chain_id) {
```
Twey: The problem is that this will be the case both when the chain worker has crashed (and needs to be restarted) _and_ when someone else is using the chain worker to process a query, in which case we must not start a new one.

Maybe rather than an `Option` this should be an `async::Mutex<Option>`, and we treat `None` the same as not present (i.e. start a new one)?

linera-core/src/worker.rs
ndr-ds: OTOH I could just always send to `Tempo` regardless of level 😅 but I wanted a way for people to opt out of sending to `Tempo`, in case some spans are just too noisy (for example spans in the microseconds that are very frequent and pollute the Tempo visualization)

linera-core/src/worker.rs
Twey: I wouldn't recommend making it generic — I think you can either use the concrete type from `oneshot` or else use a trait object.

linera-core/src/worker.rs
ma2bd: Doesn't the `self.chain_workers.lock()` around this line ensure that we atomically use & test the entrypoint ?

Twey: Yes, the test+use here is fine, but then we drop (free) the lock at the end of the block, and _another_ request can come in and find the endpoint missing (and decide to spawn another one).  This task then won't see that until it goes to put the endpoint back and finds there's already one there (and/or gets its storage corrupted…)

linera-core/src/worker.rs
Twey: I think I'd rather have a ‘don't send to Tempo’ annotation than a ‘do send to Tempo’ annotation :thinking:

Twey: I did start the convention of `instrument(level = "trace")`, but on reflection I regret it: I think they should basically always be `DEBUG`.  The only reason to make them `TRACE` is so that they get compiled out when we use `tracing/release_max_level_debug`, but after your experimentation it seems that there's no point doing that, and in exchange we lose potentially useful context from our `DEBUG` logs.

linera-core/src/worker.rs
ma2bd: I see. Good point

linera-core/src/worker.rs
ndr-ds: Here's my attempt at making this less confusing https://github.com/linera-io/linera-protocol/pull/4771

linera-core/src/worker.rs
ndr-ds: Don't we always lock the entire map? Then what you're describing doesn't seem possible 🤔 but maybe I'm missing something

PR #4771, opened 2025-10-09
Make sending traces to Tempo less confusing

ndr-ds:
## Motivation

Right now we have a confusing mix of features and naming around telemetry export. The code references both `otel` and `telemetry_only`, and it's unclear how to opt in or out of sending trace information to Tempo/OpenTelemetry backends. Additionally, there are a few bugs preventing OpenTelemetry from working correctly in production deployments.

## Proposal

Instead of `telemetry_only`, use `opentelemetry.skip` as a span field to opt out of sending specific spans to Tempo/OpenTelemetry.

By default, all spans are sent to OpenTelemetry/Tempo backends (when the `opentelemetry` feature is enabled), regardless of log level. To exclude a span from OpenTelemetry export while keeping it in local traces (Chrome trace, logs), add `opentelemetry.skip` to the span fields:

```rust
// This span will NOT be sent to Tempo
#[instrument(fields(opentelemetry.skip = true))]
fn internal_helper() { }

// This span WILL be sent to Tempo (default behavior)  
#[instrument]
fn public_api() { }
```

**Key Changes:**

**Naming & Clarity:**
- Renamed `otel` feature to `opentelemetry` for clarity
- Renamed `--otel-exporter-otlp-endpoint` CLI flag to `--otlp-exporter-endpoint`
- Renamed `--otel-trace-file` CLI flag to `--chrome-trace-file` (more accurate)
- Standardized all environment variables to use `LINERA_OTLP_EXPORTER_ENDPOINT`
- Chrome trace export includes ALL spans (useful for local debugging)
- OpenTelemetry/Tempo export respects the `opentelemetry.skip` filter

**Bug Fixes:**
- **Fixed empty endpoint validation**: Previously, passing an empty string as the OTLP endpoint would cause the exporter to fail with "Failed to parse endpoint". Now properly validates that the endpoint is non-empty before attempting to create the exporter.
- **Fixed faucet crash with OpenTelemetry**: The OTLP exporter was being initialized before the Tokio runtime was created, causing a panic: "there is no reactor running, must be called from the context of a Tokio 1.x runtime". Moved tracing initialization to after runtime creation.

**Test Improvements:**
- Added proper feature guards (`with_testing`) for test-only functions
- CI tests verify that spans with `opentelemetry.skip` are filtered from OpenTelemetry export but still appear in Chrome traces

## Test Plan

- CI tests verify that spans with `opentelemetry.skip` are filtered from OpenTelemetry export but still appear in Chrome traces
- All existing tests pass with the renamed feature and CLI flags
- Manual testing confirms OTLP traces now successfully export to Tempo without crashes

## Release Plan

Might be good to backport just so if we want to add more instrumentation on the testnet, things are not as confusing :)


ndr-ds:
* **#4771** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4771?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4771?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

linera-base/src/tracing_otel.rs
```diff
@@ -59,15 +56,9 @@ pub fn init_with_opentelemetry(log_name: &str, otlp_endpoint: Option<&str>) {
     global::set_tracer_provider(tracer_provider.clone());
     let tracer = tracer_provider.tracer("linera");
 
-    let telemetry_only_filter =
-        filter_fn(|metadata| metadata.is_span() && metadata.target() == "telemetry_only");
+    let tempo_filter = filter_fn(|metadata| metadata.is_span() && metadata.target() != "no_tempo");
```
Twey: The interface here is OpenTelemetry; Tempo is our OpenTelemetry _consumer_, provided by the infra, so it shouldn't be named in the codebase.  Some validators might prefer to use a different OpenTelemetry aggregator for example.

Twey: The `target` is useful to have in events.  Rather than a custom `target`, could we just standardize a field (e.g. `"opentelemetry/disable": true`)?

ndr-ds: Good point! I guess this is not Tempo specific at all

linera-base/src/tracing_otel.rs
```diff
@@ -81,11 +81,11 @@ pub fn init_with_opentelemetry(log_name: &str, otlp_endpoint: Option<&str>) {
         .init();
 }
 
-/// Fallback when tempo feature is not enabled.
-#[cfg(not(feature = "tempo"))]
+/// Fallback when otel feature is not enabled.
+#[cfg(not(feature = "otel"))]
```
Twey: Nit: I don't love the `otel` abbreviation.  We mostly try to avoid abbreviations as a style policy.

Twey (approved):
Looks good to me :) (modulo formatting!)

linera-base/src/tracing_otel.rs
```diff
@@ -59,15 +57,17 @@ pub fn init_with_opentelemetry(log_name: &str, otlp_endpoint: Option<&str>) {
...
-    let telemetry_only_filter =
-        filter_fn(|metadata| metadata.is_span() && metadata.target() == "telemetry_only");
+    let otel_filter = filter_fn(|metadata| {
+        if !metadata.is_span() {
+            return false;
+        }
+        metadata
+            .fields()
+            .field("otel.skip")
+            .is_none()
```
Twey: Nit: it would be nice to require the value be true in order to disable; this makes it easier to do more complex conditionals, like
```rust
#[tracing::instrument(fields(otel.skip = !otel))]
fn foo(otel: bool) {
  …
}
```

Twey: Nit: to avoid the double negative, we could alternatively name it `otel.emit`, and skip only when it's `false` (i.e. the default value if not provided is considered to be `true`).

PR #4788, opened 2025-10-13
Use `multi_get` for preprocessed_blocks.

MathieuDutSik:
## Motivation

The function `block_hashes` has emerged as a limiting factor for the running of the system.
An iteration of `get` operations is ripe for optimization.

## Proposal

We use a `multi_get`. Since what is downloaded is just a hash, there is no risk of memory problems.

## Test Plan

The CI.

## Release Plan

This kind of acceleration can be backported to TestNet Conway easily.

## Links

None.

### DIFF
--- a/linera-chain/src/chain.rs
+++ b/linera-chain/src/chain.rs
@@ -1116,10 +1116,17 @@ where
             Vec::new()
         };
         // Everything after (including) next_height in preprocessed_blocks if we have it.
-        for height in start.max(next_height).0..=end.0 {
-            if let Some(hash) = self.preprocessed_blocks.get(&BlockHeight(height)).await? {
-                hashes.push(hash);
-            }
+        let block_heights = (start.max(next_height).0..=end.0)
+            .map(BlockHeight)
+            .collect::<Vec<_>>();
+        for hash in self
+            .preprocessed_blocks
+            .multi_get(&block_heights)
+            .await?
+            .into_iter()
+            .flatten()
+        {
+            hashes.push(hash);
         }
         Ok(hashes)
     }


linera-chain/src/chain.rs
```diff
@@ -1116,10 +1116,17 @@ where
...
-            }
+        let block_heights = (start.max(next_height).0..=end.0)
+            .map(BlockHeight)
+            .collect::<Vec<_>>();
+        for hash in self
+            .preprocessed_blocks
+            .multi_get(&block_heights)
+            .await?
+            .into_iter()
+            .flatten()
```
afck, line 1127: It would be great to have a `range_get` in the future, that doesn't go through all the numbers. It's totally plausible to have a sender chain of length 10000 with only one sending block in it, and a `range_get` would be the only way to make that performant.

MathieuDutSik: Thanks for the suggestion.
I think we need some alternative to `MapView` for `preprocessed_blocks`. Maybe `LogView`.

afck: But a `LogView` can't be sparse, can it?

MathieuDutSik: Well, there are a number of points to consider. First `MapView`:
* Accessing an interval of keys `[K1, K2)` in the storage can be done natively. The only exception is `DynamoDb`, where we would have to access a lot of keys and discard the ones that do not match.
* A `range_get` can be implemented accordingly for the `CustomMapView`. Issues created.

Then `LogView`:
* The entries of the `LogView` could be made mutable. Right now, they are not.
* Yes, the `LogView` is not sparse. But entries like blockheight are almost continuous, so having values being `None` in the missing case makes sense.

PR #4789, opened 2025-10-13
Use `multi_get` in `execute_block_inner`

MathieuDutSik:
## Motivation

Some operations in `execute_block_inner` use for loops with each entry of the loop being an await operation.
This is slow.

## Proposal

Use the recently introduced `multi_get` operations.

## Test Plan

The CI.

## Release Plan

These kinds of changes can be easily backported to TestNet Conway.

## Links

None.

### DIFF
--- a/linera-chain/src/chain.rs
+++ b/linera-chain/src/chain.rs
@@ -864,7 +864,6 @@ where
         }
 
         let recipients = block_execution_tracker.recipients();
-        let mut previous_message_blocks = BTreeMap::new();
         let heights = previous_message_blocks_view.multi_get(&recipients).await?;
         let mut recipient_heights = Vec::new();
         let mut indices = Vec::new();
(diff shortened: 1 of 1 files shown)


linera-chain/src/chain.rs
```diff
@@ -864,32 +864,44 @@ where
...
             }
         }
+        let hashes = confirmed_log.multi_get(indices).await?;
+        let mut previous_message_blocks = BTreeMap::new();
+        for (hash, (recipient, height)) in hashes.into_iter().zip(recipient_heights) {
+            let hash = hash.ok_or_else(|| {
+                ChainError::InternalError("missing entry in confirmed_log".into())
+            })?;
+            previous_message_blocks.insert(recipient, (hash, height));
+        }
```
afck, line 884: (We could simplify this a lot by storing the hashes in `previous_message_blocks_view`, too, instead of looking them all up. Can't be backported, but something to consider for main in a separate PR.)

linera-chain/src/chain.rs
```diff
@@ -864,32 +864,44 @@ where
...
+        let mut recipient_heights = Vec::new();
+        let mut indices = Vec::new();
+        for (height, recipient) in heights.into_iter().zip(recipients) {
+            if let Some(height) = height {
+                let index = usize::try_from(height.0).map_err(|_| ArithmeticError::Overflow)?;
+                indices.push(index);
+                recipient_heights.push((recipient, height));
             }
         }
+        let hashes = confirmed_log.multi_get(indices).await?;
```
deuszx, line 877: This is super low-level code that has to understand the implementation details of the `MapView`. It should never be here - in a high-level logic concerning block execution.

It should be moved to `MapView` that can recover the indices from the key and value types.

PR #4791, opened 2025-10-13
Relax the inbox check.

### LINKED ISSUE #413: Relax safety criterion for inboxes during block proposals
As noted by @afck we can tolerate anticipated messages when validating a block proposal, as long as they are not caused by the block.

This will be important when we archive chains: we don't want to have to run them in every future validator just to generate messages.

Next step for sustainabilty: do not even maintain the inbox for chains that are provably retired. (The difference between a chain that is now archived and one that was not created yet can be check by querying other workers recursively. Of course this information can be cached as well.)

afck:
## Motivation

Currently a block proposal is always rejected (by a validator, or even by the local node) if any of the incoming messages included on this chain up to and including the proposal itself have not been received from the sender chain yet.

For example, if:
- block A0 on chain A has included a message from block B0 on chain B,
- and now a proposal for block A1 is made that includes a message from block C0 on chain C,
- and a validator has block C0 but not B0,

it would currently reject that proposal.

That is unnecessarily restrictive, because block A0 has already been signed by a quorum of validators, and it causes more round-trips and download work for clients.

## Proposal

Relax the inbox check, so that only incoming messages in the proposal itself have to be received yet. The validator in the above example would accept the proposal.

In other words, don't reject a new block proposal just because you have an old, already confirmed block in that chain with an incoming message whose sender block you have not seen yet.

## Test Plan

CI; some tests have been updated.

## Release Plan

    - be released in a new SDK,
    - be released in a validator hotfix, ideally at the same time.

## Links

- Closes #413.

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -2115,7 +2115,7 @@ impl<Env: Environment> ChainClient<Env> {
     }
 
     /// Prepares the chain for the next operation, i.e. makes sure we have synchronized it up to
-    /// its current height and are not missing any received messages from the inbox.
+    /// its current height.
     #[instrument(level = "trace")]
     pub async fn prepare_chain(&self) -> Result<Box<ChainInfo>, ChainClientError> {
         #[cfg(with_metrics)]


linera-core/src/client/mod.rs
```diff
@@ -2203,16 +2141,6 @@ impl<Env: Environment> ChainClient<Env> {
                 .await?;
```
ma2bd, line 2141: Update the description of the function?

linera-core/src/client/mod.rs
```diff
@@ -2203,16 +2141,6 @@ impl<Env: Environment> ChainClient<Env> {
...
 
-        // Check if we're missing any sender blocks for cross-chain messages.
-        let missing_blocks = self
-            .chain_state_view()
-            .await?
-            .collect_missing_sender_blocks()
-            .await?;
-        // Download any sender blocks we're missing.
-        self.client
-            .download_missing_sender_blocks(self.chain_id, missing_blocks)
```
ma2bd, line 2214: Are we going to do the downloading later when we propose?

afck: We don't need them for proposing anymore.

They will be downloaded as part of `find_received_certificates`, but now we don't need to distinguish those that were already included in a block.

PR #4793, opened 2025-10-14
[testnet] Relax the inbox check. (#4791)

afck:
Backport of #4791.

## Motivation

Currently a block proposal is always rejected (by a validator, or even by the local node) if any of the incoming messages included on this chain up to and including the proposal itself have not been received from the sender chain yet.

For example, if:
- block A0 on chain A has included a message from block B0 on chain B,
- and now a proposal for block A1 is made that includes a message from block C0 on chain C,
- and a validator has block C0 but not B0,

it would currently reject that proposal.

That is unnecessarily restrictive, because block A0 has already been signed by a quorum of validators, and it causes more round-trips and download work for clients.

## Proposal

Relax the inbox check, so that only incoming messages in the proposal itself have to be received yet. The validator in the above example would accept the proposal.

In other words, don't reject a new block proposal just because you have an old, already confirmed block in that chain with an incoming message whose sender block you have not seen yet.

## Test Plan

CI; some tests have been updated.

## Release Plan

- These changes should:
    - be released in a new SDK,
    - be released in a validator hotfix, ideally at the same time.

## Links

- PR to main: #4791

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -1990,7 +1990,7 @@ impl<Env: Environment> ChainClient<Env> {
     }
 
     /// Prepares the chain for the next operation, i.e. makes sure we have synchronized it up to
-    /// its current height and are not missing any received messages from the inbox.
+    /// its current height.
     #[instrument(level = "trace")]
     pub async fn prepare_chain(&self) -> Result<Box<ChainInfo>, ChainClientError> {
         #[cfg(with_metrics)]


linera-core/src/unit_tests/client_tests.rs
```diff
@@ -2844,19 +2844,15 @@ where
...
     let info = receiver2.prepare_chain().await?;
     assert_eq!(info.next_block_height, BlockHeight::from(1));
 
-    // Verify that sender's block WAS downloaded.
     let local_node = &receiver2.client.local_node;
     let sender_info = local_node.chain_info(sender.chain_id()).await?;
     assert_eq!(
         sender_info.next_block_height,
-        BlockHeight::from(1),
+        BlockHeight::ZERO,
```
deuszx, line 2855: This changed b/c we no longer have to download the sender chain when proposing a block with the incoming message (from it)?

afck: Yes, exactly.

linera-chain/src/chain.rs
```diff
@@ -712,6 +658,7 @@ where
     pub async fn remove_bundles_from_inboxes(
         &mut self,
         timestamp: Timestamp,
+        must_be_present: bool,
```
deuszx, line 661: It'd be nice to have a comment explaining when this value is supposed to be `true` or `false`. People really won't know this in a month from now.

afck: Done in #4798.

PR #4797, opened 2025-10-14
Avoid using shared chain state views except for GraphQL.

afck:
## Motivation

With the exception of a GraphQL query, the chain state view should only be accessed by the corresponding chain worker. GraphQL is why the worker can return a shared chain state view behind a lock, but we are using that in many other places, too. After https://github.com/linera-io/linera-protocol/pull/4769, this is even more dangerous, as an actor could be dropped and restarted while shared views are still being used.

## Proposal

Replace several `chain_state_view()` calls with chain info or new chain worker requests.

Also, extend a comment. (See https://github.com/linera-io/linera-protocol/pull/4793#discussion_r2428602856.)

## Test Plan

CI

## Release Plan

    - be released in a new SDK.

## Links

### DIFF
--- a/linera-core/src/chain_worker/actor.rs
+++ b/linera-core/src/chain_worker/actor.rs
@@ -157,6 +157,36 @@ where
         new_trackers: BTreeMap<ValidatorPublicKey, u64>,
         callback: oneshot::Sender<Result<(), WorkerError>>,
     },
+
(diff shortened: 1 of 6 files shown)


afck:
Currently `fn chain_state_view` is in `linera-core`, but used in `linera-service`. Not sure what's the best way to avoid this in the future.

PR #4799, opened 2025-10-14
Process epoch changes and events every time a block is executed

bart-linera:
## Motivation

We would only process epoch changes and event streams when explicitly processing the inbox, which means it was possible eg. for users to unintentionally skip epoch updates while keeping proposing blocks on their chains.

## Proposal

Process epoch changes (and any other relevant event streams) every time a block is proposed.

## Test Plan

CI

## Release Plan

    - be released in a new SDK,

## Links

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -402,19 +402,12 @@ impl<Env: Environment> Client<Env> {
     ) -> Result<Option<Box<ChainInfo>>, ChainClientError> {
         let mut last_info = None;
         // First load any blocks from local storage, if available.
-        let mut hashes = Vec::new();
-        let mut next_height = BlockHeight::ZERO;
-        {
-            let chain = self.local_node.chain_state_view(chain_id).await?;
-            next_height = next_height.max(chain.tip_state.get().next_block_height);
-            while next_height < stop {
-                let Some(hash) = chain.preprocessed_blocks.get(&next_height).await? else {
-                    break;
-                };
-                hashes.push(hash);
-                next_height = next_height.try_add_one()?;
-            }
-        }
+        let chain_info = self.local_node.chain_info(chain_id).await?;
+        let mut next_height = chain_info.next_block_height;
+        let hashes = self
+            .local_node
+            .get_preprocessed_block_hashes(chain_id, next_height, stop)
+            .await?;
         let certificates = self
             .storage_client()
             .read_certificates(hashes.clone())
@@ -628,7 +621,7 @@ impl<Env: Environment> Client<Env> {
         let certificate = self
             .communicate_chain_action(committee, finalize_action, hashed_value)
             .await?;
-        self.receive_certificate(certificate.clone(), ReceiveCertificateMode::AlreadyChecked)
+        self.receive_certificate_with_checked_signatures(certificate.clone())
             .await?;
         Ok(certificate)
     }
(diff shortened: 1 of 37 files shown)


linera-core/src/client/mod.rs
```diff
@@ -2164,75 +2164,6 @@ impl<Env: Environment> ChainClient<Env> {
         Ok(info)
     }
 
-    /// Submits a fast block proposal to the validators.
-    ///
-    /// This must only be used with valid epoch and super owner.
-    #[instrument(level = "trace", skip(committee, operations))]
-    pub async fn submit_fast_block_proposal(
```
afck, line 2129: Good catch!

linera-core/src/client/mod.rs
```diff
@@ -2790,17 +2727,46 @@ impl<Env: Environment> ChainClient<Env> {
         }
     }
 
+    /// Creates a vector of transactions which, in addition to the provided operations,
+    /// also contains epoch changes, receiving message bundles and event stream updates
+    /// (if there are any to be processed).
+    #[instrument(level = "trace", skip(operations))]
+    async fn prepend_epochs_messages_and_events(
```
deuszx, line 2702: Would we ever want to process operations (propose blocks) without prepending the epoch messages and events? If not then maybe there's no need to give it this special name. It makes it look like there's an alternative when we don't do that.

Maybe `prepare_transactions` and appropraite comment is more readable.

linera-core/src/client/mod.rs
deuszx, line 2129: This was unused?

bart-linera: Yes, but not caught by the linters, because it was `pub`.

linera-core/src/client/mod.rs
bart-linera: We probably wouldn't, but I actually think this makes for a proper descriptive name - it tells you exactly what the function does, so I'd prefer to leave it as is.

deuszx: Then you should call it `prepend_epochs_messages_and_events_and_map_to_transactions` to really _really_ tell me exactly what the function does.

linera-core/src/client/mod.rs
```diff
@@ -2765,16 +2716,10 @@ impl<Env: Environment> ChainClient<Env> {
             ClientOutcome::Committed(None) => {}
         }
 
-        let incoming_bundles = self.pending_message_bundles().await?;
-        let identity = self.identity().await?;
-        let confirmed_value = self
-            .new_pending_block(incoming_bundles, operations, blobs, identity)
-            .await?;
+        let value = self.new_pending_block(transactions, blobs).await?;
```
deuszx: Why not call it `block`?

bart-linera: I guess I wanted to minimize the changes, but I can change it to `block`.

afck (approved):

linera-core/src/client/mod.rs
```diff
@@ -2790,17 +2735,46 @@ impl<Env: Environment> ChainClient<Env> {
         }
     }
 
+    /// Creates a vector of transactions which, in addition to the provided operations,
+    /// also contains epoch changes, receiving message bundles and event stream updates
+    /// (if there are any to be processed).
```
deuszx, line 2698: More important than saying what it does is explaining why it does it and when we should call it.

deuszx (approved):

PR #4801, opened 2025-10-14
Show system operations in block explorer.

afck:
## Motivation

`linera-explorer` doesn't show the contents of system operations and outgoing messages.

Here's what I tried:
```
cargo run --bin linera net up
```
Then I pasted the exports, made a transfer from the admin chain to another chain in my wallet, and started the service:
```
export LINERA_WALLET="/tmp/.tmpyTcUrm/wallet_0.json"
export LINERA_KEYSTORE="/tmp/.tmpyTcUrm/keystore_0.json"
export LINERA_STORAGE="rocksdb:/tmp/.tmpyTcUrm/client_0.db"
cargo run --bin linera -- wallet show
cargo run --bin linera -- transfer --from 9f2047e18cd35c7509fc118c1c1df6a88d936d450ac0c2ab4752b6cf51f81f1c --to 03479ab664451306a9bad6428f9b451527009c489fe21ade5e977411931d6942 0.1
cargo run --bin linera -- service --port 8080
```
In another tab the indexer:
```
cargo run --bin linera-indexer -- run-graph-ql
```
And in a fourth tab the explorer:
```
cd linera-explorer
npm run full
npm run serve
```
If I then open localhost:3000 in Chromium, I can't see who transferred how much.

## Proposal

Include structured information in the transaction metadata in GraphQL.

## Test Plan

Some tests were added, and I tried a transfer locally. The transaction now looks like this:

image

## Release Plan

    - be released in a new SDK.

## Links

### DIFF
--- a/linera-chain/src/data_types/metadata.rs
+++ b/linera-chain/src/data_types/metadata.rs
@@ -11,9 +11,7 @@ use linera_base::{
     identifiers::{Account, AccountOwner, ApplicationId, ChainId},
     ownership::{ChainOwnership, TimeoutConfig},
 };
-use linera_execution::{
-    system::AdminOperation, Message, Operation, SystemMessage, SystemOperation,
-};
+use linera_execution::{system::AdminOperation, Message, SystemMessage, SystemOperation};
 use serde::{Deserialize, Serialize};
 
 /// Timeout configuration metadata for GraphQL.
@@ -25,7 +23,8 @@ pub struct TimeoutConfigMetadata {
     pub base_timeout_ms: String,
     /// The duration by which the timeout increases after each single-leader round in milliseconds.
     pub timeout_increment_ms: String,
-    /// The age of an incoming tracked or protected message after which validators start transitioning to fallback mode, in milliseconds.
+    /// The age of an incoming tracked or protected message after which validators start
+    /// transitioning to fallback mode, in milliseconds.
     pub fallback_duration_ms: String,
 }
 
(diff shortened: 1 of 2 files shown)


linera-chain/src/data_types/core.rs
Twey, line 1: Let's put ‘core’ things directly in `mod.rs` rather than introducing the `core` name (which is nothing to do with `linera-core`)?

linera-chain/src/data_types/mod.rs
```diff
@@ -164,8 +168,8 @@ pub struct OperationMetadata {
     pub application_id: Option<ApplicationId>,
     /// For user operations, the serialized bytes (as a hex string for GraphQL)
     pub user_bytes_hex: Option<String>,
-    /// For system operations, the serialized bytes (as a hex string for GraphQL)
-    pub system_bytes_hex: Option<String>,
+    /// For system operations, structured representation
+    pub system_operation: Option<SystemOperationMetadata>,
```
deuszx, line 172: This is a breaking change, isn't it?

deuszx: Meaning it can't be backported to testnet.

afck: It changes the GraphQL schema, yes.

afck: Why can't it be backported? It just has to be used with the updated block explorer.

deuszx: Ah, sorry. I thought it's used in more places than just block explorer.

If it's not, I'd say its place isn't in here but in explorer-specific crate.

linera-chain/src/data_types/metadata.rs
```diff
@@ -0,0 +1,484 @@
...
+    /// The age of an incoming tracked or protected message after which validators start transitioning to fallback mode, in milliseconds.
+    pub fallback_duration_ms: String,
+}
+
+impl From<&TimeoutConfig> for TimeoutConfigMetadata {
+    fn from(config: &TimeoutConfig) -> Self {
+        TimeoutConfigMetadata {
+            fast_round_ms: config
+                .fast_round_duration
+                .map(|d| (d.as_micros() / 1000).to_string()),
```
afck, line 36: Not sure about this: We lose some precision here, and converting it back wouldn't necessarily result in the same timeout config.

On the other hand, we also use milliseconds in all other user-visible places, so most timeout configs will be whole milliseconds.

PR #4803, opened 2025-10-15
Use the BTreeMap for the `multi_get` in `LogView`.

MathieuDutSik:
## Motivation

It has emerged from #4789 that a `merge_get` is being done on some indices that can overlap.
This makes the `multi_get` a little bit slower than it should be.

## Proposal

Use a `BTreeMap` for accessing the keys.

## Test Plan

The CI.

## Release Plan

This change could be backported to the TestNet Conway.

## Links

None

### DIFF
--- a/linera-views/src/views/log_view.rs
+++ b/linera-views/src/views/log_view.rs
@@ -269,8 +269,11 @@ where
             }
             let values = self.context.store().read_multi_values(keys).await?;
             for (positions, value) in vec_positions.into_iter().zip(values) {
-                for position in positions {
-                    *result.get_mut(position).unwrap() = value.clone();
+                if let Some((&last, rest)) = positions.split_last() {
+                    for &position in rest {
+                        *result.get_mut(position).unwrap() = value.clone();
+                    }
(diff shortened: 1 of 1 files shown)


linera-views/src/views/log_view.rs
```diff
@@ -245,24 +248,30 @@ where
...
+                keys.push(key);
+                vec_positions.push(positions);
+            }
             let values = self.context.store().read_multi_values(keys).await?;
-            for (pos, value) in positions.into_iter().zip(values) {
-                *result.get_mut(pos).unwrap() = value;
+            for (positions, value) in vec_positions.into_iter().zip(values) {
+                for position in positions {
+                    *result.get_mut(position).unwrap() = value.clone();
+                }
```
afck, line 277: (Not sure if it's worth it, but we could avoid the cloning for the last position.)

MathieuDutSik: Yes, it is worth it. Thank you.
Found the way to do it with `split_last()`.

PR #4804, opened 2025-10-15
[testnet] backport `multi_get` changes to `MapView` and protocol code (#4789, #4788, #4634)

MathieuDutSik:
Backport #4789, #4788, #4634

## Motivation

The `multi_get` was introduced in `MapView`, which provides some improvement to the protocol code.
The most affected is maybe `block_hashes`, which showed up as a slow part in the benchmarks.

## Proposal

The backport is clear.

## Test Plan

The CI.

## Release Plan

The backport to Conway is the point.

## Links

None

### DIFF
--- a/linera-chain/src/chain.rs
+++ b/linera-chain/src/chain.rs
@@ -1091,7 +1091,7 @@ where
             .flatten()
         {
             hashes.push(hash);
-	}
+        }
         Ok(hashes)
     }
 
--- a/linera-chain/src/chain.rs
+++ b/linera-chain/src/chain.rs
@@ -822,31 +822,43 @@ where
         }
 
         let recipients = block_execution_tracker.recipients();
-        let mut previous_message_blocks = BTreeMap::new();
-        for recipient in recipients {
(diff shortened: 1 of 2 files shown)


linera-chain/src/chain.rs
```diff
@@ -1068,10 +1080,17 @@ where
             Vec::new()
         };
         // Everything after (including) next_height in preprocessed_blocks if we have it.
-        for height in start.max(next_height).0..=end.0 {
-            if let Some(hash) = self.preprocessed_blocks.get(&BlockHeight(height)).await? {
-                hashes.push(hash);
-            }
+        let block_heights = (start.max(next_height).0..=end.0)
```
deuszx, line 1083: It'd be nice to implement the `Range` for `BlockHeight`.

deuszx (approved):
My [comment](https://github.com/linera-io/linera-protocol/pull/4789/files#r2428615564) from the original PR still stands: 

this is a low-level code that was added to a high-level protocol piece. It shouldn't be like this.

Approving since it's a backport.

PR #4805, opened 2025-10-15
Remove `receive_certificate_and_update_validators`; fix flaky test.

afck:
## Motivation

`test_memory_cross_chain_message` is flaky; `receive_certificate_and_update_validators` sometimes fails. But this method is not expected to work anymore since sender chains are now sparse and it tries to get the local chain info for the sender chain. Also, it is used only in tests.

## Proposal

Remove it. Use `synchronize_from_validators` or `synchronize_chain_state` instead.

## Test Plan

CI

## Release Plan

- These changes should be backported to `testnet_conway`.

## Links

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -621,7 +621,7 @@ impl<Env: Environment> Client<Env> {
         let certificate = self
             .communicate_chain_action(committee, finalize_action, hashed_value)
             .await?;
-        self.receive_certificate_no_check(certificate.clone())
+        self.receive_certificate_with_checked_signatures(certificate.clone())
             .await?;
         Ok(certificate)
     }
@@ -738,7 +738,7 @@ impl<Env: Environment> Client<Env> {
     /// Processes the confirmed block certificate in the local node without checking signatures.
     /// Also downloads and processes all ancestors that are still missing.
     #[instrument(level = "trace", skip_all)]
-    async fn receive_certificate_no_check(
+    async fn receive_certificate_with_checked_signatures(
         &self,
         certificate: ConfirmedBlockCertificate,
     ) -> Result<(), ChainClientError> {


linera-core/src/client/mod.rs
```diff
@@ -735,51 +735,16 @@ impl<Env: Environment> Client<Env> {
         Ok(certificate)
     }
 
-    /// Processes the confirmed block certificate and its ancestors in the local node, then
-    /// updates the validators up to that certificate.
+    /// Processes the confirmed block certificate in the local node without checking signatures.
+    /// Also downloads and processes all ancestors that are still missing.
     #[instrument(level = "trace", skip_all)]
-    async fn receive_certificate_and_update_validators(
+    async fn receive_certificate_no_check(
```
MathieuDutSik: I want to put a hard stop on functions having a postfix `_no_check` for the following reasons:
* Why add it there and not in so many other functions?
* A check might be added later, then rename the function? Forget doing it?
* Some checks might be added by features. This is how I would do it: Add some feature "sanity-check" that makes the protocol 3-4 times slower but does all kinds of consistency checks. Not there right now, but I would really like that in the future.

linera-core/benches/client_benchmarks.rs
```diff
@@ -54,39 +54,30 @@ pub async fn run_claim_bench<B>(
...
         .transfer_to_account(AccountOwner::CHAIN, amt, account)
         .await
         .unwrap()
         .unwrap();
 
-    chain2
-        .receive_certificate_and_update_validators(cert)
-        .await
-        .unwrap();
+    chain2.synchronize_from_validators().await.unwrap();
```
MathieuDutSik, line 63: Unrelated to this PR, but is there a reason why this function does not return a `anyhow::Result<()>` instead of all thses unwrappings?

afck: Last time we talked about going one way or the other (results everywhere vs. `unwrap` everywhere) it was controversial and we didn't end up making a decision.

linera-core/src/client/mod.rs
afck: OK, maybe it's a bad name. This is not about some arbitrary sanity check, but to clarify that this should only be called with a certificate whose signatures have already been verified.

linera-core/benches/client_benchmarks.rs
MathieuDutSik: It was about other functions that were already returning `anyhow::Result<()>`. Anyway, not related.

afck: Yes, I agree: Some do and some don't, and we should do the same for all functions. But we first need to discuss with the team which way to go.

afck:
The explicit `receive_certificate_…` call was only simulating what is already happening automatically inside `synchronize_from_validators`. Doing it afterwards would be a no-op.

linera-core/src/client/mod.rs
afck: I renamed it to `receive_certificate_with_checked_signatures`.

MathieuDutSik:
> The explicit `receive_certificate_…` call was only simulating what is already happening automatically inside `synchronize_from_validators`. Doing it afterwards would be a no-op.

Ok, I understand that `receive_certificate_and_update_validators` is replaced by `synchronize_from_validators`.
However, the previous version was taking a certificate as an argument.

Examining other parts of the test code, we see that there are some
```rust
    assert_eq!(
        builder
            .check_that_validators_have_certificate(client1.chain_id, BlockHeight::ZERO, 3)
            .await
            .unwrap(),
        certificate
    );
```

Is it possible to insert in the test some checks of the certificate?

I mean, yes, it can be a no-op. But we are dealing with test code, so why not test what can be tested?

PR #4811, opened 2025-10-16
Check the block height before the epoch in `process_timeout`

bart-linera:
## Motivation

When processing a timeout, we check whether the epoch of the certificate is correct before checking whether the height is correct - this may lead to spurious `InvalidEpoch` errors.

## Proposal

Validate the height first.

## Test Plan

CI

## Release Plan

    - be released in a new SDK,

## Links


deuszx (commented):
And a timeout certificate _can_ be created for a previous epoch while still being correct b/c we're timing out the round in the last epoch?

bart-linera:
Basically the committee that is current on the chain is the one that is expected to sign anything regarding the next block height, including timeouts. So if the last block on the chain uses an older committee, both the next block and the timeout certificates for the next height will be expected to be signed by that older committee.

PR #4812, opened 2025-10-16
Lower Scylla commitlog size

ndr-ds:
## Motivation

The current commitlog size value works better for read heavy workloads. That's not the case for the testnet, and it damages the reads even more because it generates more SSTables before triggering a compaction.

## Proposal

Lower the commitlog size just on the testnet for now

## Test Plan

Will deploy this. No new docker image is necessary, as this is just a config change.


ma2bd (approved):
Can we tune this differently for certain partitions?

ndr-ds:
Unfortunately not :(

PR #4814, opened 2025-10-17
Use a new schema for the data storage in Linera.

### LINKED ISSUE #4807: All blobs live on the same Scylla partition
We're seeing read latency spikes (very high read latency p95, up to the ~300 ms range) in our Scylla backend.

These seem to be correlated to blob read operations like `contains_blob`.

We suspect that what's happening is that:
- blob queries are relatively slow, because blobs are big
- all blobs are assigned the same `root_key`, which we use as a Scylla partition key
- being on the same partition means that all blob queries are handled on one CPU, where they block all other queries to that CPU
- this results in the queries at the tail of the query queue being serviced very late

As a solution, we suggest setting the partition key of blobs to be just the blob ID.

This is a different behaviour from that expected by `linera-views` (which correlates things under the same `root_key`, e.g. chain ID, for fast multi-key queries on related objects), so we'd like to pull out the ‘key/value store abstraction’ layer (`linera_views::backend`) from the ‘persistent views’ layer of `linera-views` and use that directly for blobs.

MathieuDutSik:
## Motivation

Databases commonly use a `partition_key` which corresponds to `root_key` in our code.
The partition key is used for hashing purposes in order to spread the workload over nodes.

An unfortunate feature of the existing schema is that Blobs, BlobStates, Events, and certificates
are all in the same partition (the one corresponding to `&[]`). This causes some performance
problems. A common recommendation for schema design is that the partition key should be 
spread out so that one bin does not receive too much data.

Fixes #4807 

## Proposal

The following proposal is implemented:
* For all base keys except Event, the root key is determined from the serialization.
* For the Events, we want to access several events at once. So, the base key is serialized by taking only the `ChainId`, `StreamId`. This led to the introduction of a `fn root_key(&self)` for the `BaseKey` type. The function does not return errors since for the types in question, BlobId, CryptoHash, ChainId, returning an error is impossible.

Why this change is the right one:
* There is no limit to the number of partition keys in databases. On the other hand, there is a limit to the size of data for a specified partition key. So concentrating all data on one partition key creates potential problems above 100M and may fail completely for 2G.
* We are already having a root-key formed from the `ChainId` for the application states. So, we already accept that we can have many many partition keys.

The `Batch` of `linera-storage` is replaced by a `MultiPartitionBatch`. It is unfortunate that we had the collision with the `Batch` of `linera-views`.

This PR does the requested job of changing only the `linera-storage`. However, there are some losses of parallelization for the `read_multi_values/contains_keys` operation. This is not irremediable:
* We can add some functions `read_multi_root_values(_, root_keys: Vec<Vec<u8>>, key: Vec<u8>)` to the `KeyValueDatabase`. It is possible to implement this feature efficiently in `ScyllaDb`, which is our main database target.
* We can add some `write_multi_partition_batch` to the `KeyValueDatabase`. Note that the existing `write_batch` in `db_storage.rs` is creating many futures, but the right solution is likely to group the entries. Of course, batch size is an issue, but it has to be addressed by measuring it not spreading over all.
* It is a little bit problematic to see how those features could be implemented in the combinators like LruCaching, ValueSplitting, and so on.

## Test Plan

The CI.

## Release Plan

Hopefully, to put it into the main.

It is possible to write a migration tool that takes the existing storage of TestNet Conway and converts it to the new schema. But that is only if we really want to do that.

Before that, it would be good to see if the scalability works as expected for ScyllaDb runs.

## Links

None.

### DIFF
--- a/linera-storage/src/db_storage.rs
+++ b/linera-storage/src/db_storage.rs
@@ -1034,7 +1034,7 @@ where
         Ok(())
     }
 
-    #[instrument(skip_all, fields(batch_size = batch.key_value_bytes.len()))]
+    #[instrument(skip_all, fields(batch_size = batch.keys_value_bytes.len()))]
     async fn write_batch(&self, batch: MultiPartitionBatch) -> Result<(), ViewError> {
         if batch.keys_value_bytes.is_empty() {
             return Ok(());


linera-storage/src/db_storage.rs
```diff
@@ -224,37 +225,46 @@ pub mod metrics {
...
 }
 
+/// The default key used when the root_key contains the information.
+const DEFAULT_KEY: &[u8] = &[0];
+
 #[derive(Default)]
-struct Batch {
-    key_value_bytes: Vec<(Vec<u8>, Vec<u8>)>,
+struct MultiPartitionBatch {
+    keys_value_bytes: Vec<(Vec<u8>, Vec<u8>, Vec<u8>)>,
```
ma2bd, line 240: We may want to group the entries by partition: `Vec<(Vec<u8>, SinglePartitionBatch)>`

linera-storage/src/db_storage.rs
```diff
@@ -224,37 +225,46 @@ pub mod metrics {
...
 
-impl Batch {
+impl MultiPartitionBatch {
     fn new() -> Self {
         Self::default()
     }
 
-    fn put_key_value_bytes(&mut self, key: Vec<u8>, value: Vec<u8>) {
-        self.key_value_bytes.push((key, value));
+    fn put_key_value_bytes(&mut self, root_key: Vec<u8>, key: Vec<u8>, value: Vec<u8>) {
```
ma2bd, line 248: "root key" is a little weird outside of `View`s. Can we just say "partition" or "partition_key"?

linera-storage/src/db_storage.rs
```diff
@@ -319,9 +330,77 @@ enum BaseKey {
     NetworkDescription,
 }
 
+impl BaseKey {
+    fn root_key(&self) -> Vec<u8> {
```
ma2bd: `fn partition` ?

linera-storage/src/db_storage.rs
```diff
@@ -224,37 +225,46 @@ pub mod metrics {
...
+        self.keys_value_bytes.push((root_key, key, bytes));
         Ok(())
     }
 
     fn add_blob(&mut self, blob: &Blob) -> Result<(), ViewError> {
         #[cfg(with_metrics)]
         metrics::WRITE_BLOB_COUNTER.with_label_values(&[]).inc();
-        let blob_key = bcs::to_bytes(&BaseKey::Blob(blob.id()))?;
-        self.put_key_value_bytes(blob_key.to_vec(), blob.bytes().to_vec());
+        let root_key = BaseKey::Blob(blob.id()).root_key();
```
ma2bd: It seems like we don't need the `enum BaseKey` at all.

linera-storage/src/db_storage.rs
```diff
@@ -319,9 +330,77 @@ enum BaseKey {
...
+            }
+            BaseKey::Certificate(hash) => {
+                let mut key = vec![INDEX_CERTIFICATE];
+                key.extend_from_slice(hash.as_bytes().as_slice());
+                key
+            }
+            BaseKey::ConfirmedBlock(hash) => {
+                let mut key = vec![INDEX_CONFIRMED_BLOCK];
+                key.extend_from_slice(hash.as_bytes().as_slice());
+                key
```
ma2bd: These should have the same partition actually

linera-storage/src/db_storage.rs
```diff
@@ -319,9 +330,77 @@ enum BaseKey {
...
+                key.push(blob_id.blob_type as u8);
+                key.extend_from_slice(blob_id.hash.as_bytes().as_slice());
+                key
+            }
+            BaseKey::BlobState(blob_id) => {
+                let mut key = vec![INDEX_BLOB_STATE];
+                key.push(blob_id.blob_type as u8);
+                key.extend_from_slice(blob_id.hash.as_bytes().as_slice());
+                key
+            }
```
ma2bd: same

linera-storage/src/db_storage.rs
ma2bd: We could define `PartitionKey` and serialize it but it seems like you prefer do things by hand.

linera-storage/src/db_storage.rs
```diff
@@ -319,9 +330,77 @@ enum BaseKey {
...
+}
+
+fn is_chain_state(root_key: &[u8]) -> bool {
+    if root_key.is_empty() {
+        return false;
+    }
+    root_key[0] == INDEX_CHAIN_ID
+}
+
 const INDEX_CHAIN_ID: u8 = 0;
```
ma2bd: nit: The expected name would be `CHAIN_ID_TAG`.

linera-storage/src/db_storage.rs
```diff
@@ -319,9 +330,77 @@ enum BaseKey {
...
+const INDEX_CERTIFICATE: u8 = 1;
+const INDEX_CONFIRMED_BLOCK: u8 = 2;
```
ma2bd: `BLOCK_HASH_TAG`

linera-storage/src/db_storage.rs
```diff
@@ -319,9 +330,77 @@ enum BaseKey {
...
 const INDEX_BLOB_ID: u8 = 3;
```
ma2bd: `BLOB_ID_TAG`

ma2bd:
I think we can also start thinking about have two modules co-exist on the testnet branch: the current partitioning and the new one. Then, we will auto-migrate data on DB startup. This will require an extra version number.

linera-storage/src/db_storage.rs
MathieuDutSik: I very much have an argument against that.

What we would like ideally is to be able to read `(root_key_i, key_i)` with the root_keys all different and the keys all different as well. But we cannot do that with ScyllaDb, which is our target system.

What we can read efficiently are two orthogonal kinds of read_keys:
* Reading with the same root_key but varying key. This is what we have now with `read_multi_values_bytes`.
* Reading with the same key but varying root_key. These are the functions I would like to introduce in the `KeyValueStore` trait.

If we put the certificates and the confirmed block on the same partition, then we are no longer able to use the second class of functions for `read_certificates`.

MathieuDutSik:
> I think we can also start thinking about have two modules co-exist on the testnet branch: the current partitioning and the new one. Then, we will auto-migrate data on DB startup. This will require an extra version number.

No problem. 

To have the two schemas coexist on TestNet, we first need to have them coexist on main. So, my question is, how do we make the storage aware of in which storage it is running?

Two ideas that come to mind:
* Passing an argument
* Reading keys that would be present, for example, the `NetworkDescription`, and concluding from that read whether we are in the old schema or the new one.

linera-storage/src/db_storage.rs
ma2bd: > What we would like ideally is to be able to read (root_key_i, key_i) with the root_keys all different and the keys all different as well. 

Why?

ma2bd: See other comment https://github.com/linera-io/linera-protocol/pull/4814#discussion_r2441369859

MathieuDutSik: According to the required change that started this discussion, we would like to have the `Certificate` and `ConfirmedBlock` on the same partition. The keys would then be, for example, `[0]` and `[1]`.

Let's examine what the code of `read_certificates` would request the database for `hashes` being `hash1, hash2, hash3`. The root_keys would then be `root_key1,root_key2,root_key3`. The needed pairs of `(root_key, key)` needed would be `(root_key1,[0])`, `(root_key2,[0])`, `(root_key3,[0])`, `(root_key1,[1])`, `(root_key2,[1])`, `(root_key3,[1])`,

That is the answer to the "Why?":  Ideally, we would need to access a sequence of `(root_key_i, key_i)` so as to be able to read in one single operation the data needed for the `read_certificates`.

It is a general problem: If we do not group database access operations into single operations, then we face the problem that we have to send many different requests, which clog the system. This PR does make many operations as simple loops, but the goal would be in the end to have global read operations.

Therefore, when designing a new schema, we need to keep in mind what is possible with the storage. It is especially important to consider ScyllaDb, which is the system that we are using extensively right now. Migrating databases from one schema to another is a slow and expensive endeavor, and so it is good to have a good schema from the start.

So, what is possible for `ScyllaDb`? ScyllaDb uses the Cassandra Query Language, and so what can be done depends on that:
(A) It is possible to access key pairs that share the same `root_key`. The request in question is
```
SELECT k,v FROM kv.pairs WHERE root_key = ? AND k IN (?, ....., ?)
```
and this is what we are using in `linera-views/src/backends/scylla_db.rs`.

(B) It is possible to access key pairs that share the same `key`. The request in question is
```
SELECT root_key,v FROM kv.pairs WHERE root_key IN (?, ...., ?) AND k = ?
```
that works correctly.

(C) It is NOT possible to access `(root_key_i,key_i)`. The request in question is
```
SELECT root_key,k,v FROM kv.pairs WHERE (root_key,k) IN ((?,?),(?,?),.....,(?,?))
```
The error is "The query is syntactically correct but invalid, Error message: Multi-column relations can only be applied to clustering columns but was applied to: root_key"

(D) It is possible to access key pairs that form a direct product. The request in question is
```
SELECT root_key,k,v FROM kv.pairs WHERE root_key IN (?,?,.....,?) and k IN (?,?,....,?)
```
That does work!

-------

The consequences of the decision on the schema are the following:
* If we compute the `root_key` for the certificate/confirmed blocks according to the PR, then if we want to have global operations, we need to implement type B operations.
* If we put the certificate/confirmed block on the same partition, and we want to have global operations. Then we need type D operations. Otherwise, we would need two operations of type B.

linera-storage/src/db_storage.rs

linera-storage/src/db_storage.rs

linera-storage/src/db_storage.rs

linera-storage/src/db_storage.rs
MathieuDutSik: That is kind of difficult since it would require introducing a `BTreeMap`.

linera-storage/src/db_storage.rs
MathieuDutSik: The nomenclature is the following:
```rust
fn open_shared(&self, root_key: &[u8]) -> Result<Self::Store, Self::Error>;
```

and the argument `root_key` is exactly what is passed to `open_shared`. So, of course we can change that, but then the changes would be big and concern a lot of the codebase. So, I guess, this would be for a different PR.

linera-storage/src/db_storage.rs
MathieuDutSik: Again, this is a renaming that should be done accordingly all over the codebase.
And so likely in a different PR.

linera-storage/src/db_storage.rs
MathieuDutSik: Yes, in that case, it does pass and we can merge in the same partition.

linera-storage/src/db_storage.rs
```diff
@@ -308,20 +328,72 @@ pub struct DbStorage<Database, Clock = WallClock> {
 }
 
 #[derive(Debug, Serialize, Deserialize)]
-enum BaseKey {
+enum RootKey {
```
ma2bd, line 331: This enum is still rather useless

linera-storage/src/db_storage.rs
```diff
@@ -1019,20 +1068,20 @@ where
...
+        if batch.keys_value_bytes.is_empty() {
             return Ok(());
         }
         let mut futures = Vec::new();
-        for (key, bytes) in batch.key_value_bytes {
-            let store = self.database.open_shared(&[])?;
+        for (root_key, key, bytes) in batch.keys_value_bytes {
+            let store = self.database.open_shared(&root_key)?;
             futures.push(async move { Self::write_entry(&store, key, bytes).await });
         }
```
ma2bd, line 1046: Can we make sure to open partitions at most once and do a multi-key write?

linera-storage/src/db_storage.rs
ma2bd: Here we're getting the worst of both worlds:
* we're not using BCS (any more) to reduce boilerplate
* we're stack-allocating an object to destroy it immediately

linera-storage/src/db_storage.rs
ma2bd: I believe this works but it's a little weird because in the rest of the codebase, we're rather consistent about using multi-key operations when we can

MathieuDutSik: Well, the code prior to this PR is:
```rust
    async fn write_batch(&self, batch: Batch) -> Result<(), ViewError> {
        if batch.key_value_bytes.is_empty() {
            return Ok(());
        }
        let mut futures = Vec::new();
        for (key, bytes) in batch.key_value_bytes {
            let store = self.database.open_shared(&[])?;
            futures.push(async move { Self::write_entry(&store, key, bytes).await });
        }
        futures::future::try_join_all(futures).await?;
        Ok(())
    }
```

So, I would say that the PR is fairly neutral in that respect. It does not make things worse. For other parts of the code, yes this PR does make operations not grouped.

The solution to that is to write down a multi-partition batch. The issue is #4815

linera-storage/src/db_storage.rs
MathieuDutSik: Ok, replaced with
```rust
impl RootKey {
    fn bytes(&self) -> Vec<u8> {
        bcs::to_bytes(self).unwrap()
    }
}
```

So, avoid the boilerplate.

linera-storage/src/db_storage.rs
```diff
@@ -559,34 +590,30 @@ where
...
-        for (blob_id, result) in blob_ids.iter().zip(results) {
-            if !result {
+        for blob_id in blob_ids {
+            let root_key = RootKey::Blob(*blob_id).bytes();
+            let store = self.database.open_shared(&root_key)?;
+            let test = store.contains_key(DEFAULT_KEY).await?;
+            if !test {
                 missing_blobs.push(*blob_id);
             }
         }
```
ndr-ds, line 619: Here we're increasing the number of DB queries. Couldn't we create another `PreparedStatement` on `ScyllaDbClient` with a `root_key IN (?, ?)` type query?
That would allow us to keep this a single network request to Scylla as before.

linera-storage/src/db_storage.rs
```diff
@@ -637,31 +664,22 @@ where
...
-        let blob_keys = blob_ids
-            .iter()
-            .map(|blob_id| bcs::to_bytes(&BaseKey::Blob(*blob_id)))
-            .collect::<Result<Vec<_>, _>>()?;
-        let store = self.database.open_shared(&[])?;
-        let maybe_blob_bytes = store.read_multi_values_bytes(blob_keys).await?;
+        let mut blobs = Vec::new();
+        for blob_id in blob_ids {
+            blobs.push(self.read_blob(*blob_id).await?);
+        }
```
ndr-ds, line 670: Same here as in `missing_blobs`

linera-storage/src/db_storage.rs
```diff
@@ -677,14 +695,10 @@ where
...
-            .map(|blob_id| bcs::to_bytes(&BaseKey::BlobState(*blob_id)))
-            .collect::<Result<_, _>>()?;
-        let store = self.database.open_shared(&[])?;
-        let blob_states = store
-            .read_multi_values::<BlobState>(blob_state_keys)
-            .await?;
+        let mut blob_states = Vec::new();
+        for blob_id in blob_ids {
+            blob_states.push(self.read_blob_state(*blob_id).await?);
+        }
```
ndr-ds, line 701: Same here as in `missing_blobs`

linera-storage/src/db_storage.rs
```diff
@@ -709,15 +723,14 @@ where
...
-            .read_multi_values::<BlobState>(blob_state_keys)
-            .await?;
-        let mut batch = Batch::new();
+        let mut maybe_blob_states = Vec::new();
+        for blob_id in blob_ids {
+            let root_key = RootKey::Blob(*blob_id).bytes();
+            let store = self.database.open_shared(&root_key)?;
+            let maybe_blob_state = store.read_value::<BlobState>(ONE_KEY).await?;
+            maybe_blob_states.push(maybe_blob_state);
+        }
```
ndr-ds, line 732: Same here as in `missing_blobs`

linera-storage/src/db_storage.rs
```diff
@@ -742,15 +755,14 @@ where
...
+        let mut blob_states = Vec::new();
+        for blob in blobs {
+            let root_key = RootKey::Blob(blob.id()).bytes();
+            let store = self.database.open_shared(&root_key)?;
+            let has_state = store.contains_key(ONE_KEY).await?;
+            blob_states.push(has_state);
+            if has_state {
                 batch.add_blob(blob)?;
             }
         }
```
ndr-ds, line 768: Same here as in `missing_blobs`

PR #4818, opened 2025-10-19
Use crypto_hash_mut whenever possible

ma2bd:
## Motivation

`hash_mut` has fewer locks + I'm looking to simplify the implementation of the historical hash

## Proposal

Use `crypto_hash_mut` whenever possible 

I'm tempted to remove the non-mut version but I suppose that's a different discussion.

## Test Plan

CI

## Release Plan

    - be released in a new SDK,

### DIFF
--- a/linera-chain/src/chain.rs
+++ b/linera-chain/src/chain.rs
@@ -502,7 +502,7 @@ where
             return Ok(());
         }
         // Recompute the state hash.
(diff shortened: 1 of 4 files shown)


ma2bd:
> LGTM
> 
> That being said, it is incorrect to say that `hash_mut` is lock-free. It has just fewer locks.

right

PR #4819, opened 2025-10-19
Make clone_unchecked return a Result

ma2bd:
## Motivation

Remove the potential panic in reentrant_collection_view.rs

## Proposal

* Make clone_unchecked return a Result
* Remove dead code

## Test Plan

CI

## Release Plan

### DIFF
--- a/linera-views-derive/src/snapshots/linera_views_derive__tests__generate_clonable_view_code-3.snap
+++ b/linera-views-derive/src/snapshots/linera_views_derive__tests__generate_clonable_view_code-3.snap
@@ -1,6 +1,6 @@
 ---
 source: linera-views-derive/src/lib.rs
-expression: pretty(generate_clonable_view_code(input))
+expression: pretty(generate_clonable_view_code(input).unwrap())
 ---
 impl linera_views::views::ClonableView for TestView
 where
@@ -12,10 +12,10 @@ where
     >: ClonableView,
     Self: linera_views::views::View,
 {
-    fn clone_unchecked(&mut self) -> Self {
-        Self {
-            register: self.register.clone_unchecked(),
-            collection: self.collection.clone_unchecked(),
-        }
+    fn clone_unchecked(&mut self) -> Result<Self, linera_views::ViewError> {
+        Ok(Self {
+            register: self.register.clone_unchecked()?,
+            collection: self.collection.clone_unchecked()?,
+        })
     }
 }
--- a/linera-views-derive/src/snapshots/linera_views_derive__tests__generate_clonable_view_code-2.snap
+++ b/linera-views-derive/src/snapshots/linera_views_derive__tests__generate_clonable_view_code-2.snap
@@ -1,6 +1,6 @@
 ---
 source: linera-views-derive/src/lib.rs
-expression: pretty(generate_clonable_view_code(input))
+expression: pretty(generate_clonable_view_code(input).unwrap())
 ---
 impl<C, MyParam> linera_views::views::ClonableView for TestView<C, MyParam>
 where
@@ -9,10 +9,10 @@ where
     CollectionView<C, usize, RegisterView<C, usize>>: ClonableView,
     Self: linera_views::views::View,
 {
-    fn clone_unchecked(&mut self) -> Self {
-        Self {
-            register: self.register.clone_unchecked(),
-            collection: self.collection.clone_unchecked(),
-        }
+    fn clone_unchecked(&mut self) -> Result<Self, linera_views::ViewError> {
+        Ok(Self {
+            register: self.register.clone_unchecked()?,
+            collection: self.collection.clone_unchecked()?,
+        })
     }
 }
--- a/linera-views-derive/src/snapshots/linera_views_derive__tests__generate_clonable_view_code-4.snap
+++ b/linera-views-derive/src/snapshots/linera_views_derive__tests__generate_clonable_view_code-4.snap
@@ -1,6 +1,6 @@
 ---
 source: linera-views-derive/src/lib.rs
-expression: pretty(generate_clonable_view_code(input))
+expression: pretty(generate_clonable_view_code(input).unwrap())
 ---
 impl<MyParam> linera_views::views::ClonableView for TestView<MyParam>
 where
@@ -13,10 +13,10 @@ where
     >: ClonableView,
     Self: linera_views::views::View,
 {
-    fn clone_unchecked(&mut self) -> Self {
-        Self {
-            register: self.register.clone_unchecked(),
-            collection: self.collection.clone_unchecked(),
-        }
+    fn clone_unchecked(&mut self) -> Result<Self, linera_views::ViewError> {
+        Ok(Self {
+            register: self.register.clone_unchecked()?,
+            collection: self.collection.clone_unchecked()?,
+        })
     }
 }
--- a/linera-views-derive/src/snapshots/linera_views_derive__tests__generate_clonable_view_code-5.snap
+++ b/linera-views-derive/src/snapshots/linera_views_derive__tests__generate_clonable_view_code-5.snap
@@ -1,6 +1,6 @@
 ---
 source: linera-views-derive/src/lib.rs
-expression: pretty(generate_clonable_view_code(input))
+expression: pretty(generate_clonable_view_code(input).unwrap())
 ---
 impl linera_views::views::ClonableView for TestView
 where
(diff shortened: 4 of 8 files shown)


linera-service/src/exporter/storage.rs
```diff
@@ -142,14 +142,14 @@ where
         self.shared_canonical_state.push(block)
     }
 
-    fn clone(&mut self) -> Self {
-        Self {
+    fn clone(&mut self) -> Result<Self, ExporterError> {
```
ma2bd, line 145: @deuszx Note sure why we need this method in the first place

linera-service/src/exporter/storage.rs
```diff
@@ -422,13 +422,13 @@ where
         }
     }
 
-    fn clone(&mut self) -> Self {
-        Self {
+    fn clone(&mut self) -> Result<Self, ExporterError> {
```
MathieuDutSik, line 425: Typically, a `clone` operation
* Does not return an error.
* Take `&self` as argument instead of `&mut self`.

So maybe a renaming of the function should be done.

linera-views-derive/src/snapshots/linera_views_derive__tests__generate_clonable_view_code-3.snap
```diff
@@ -12,10 +12,10 @@ where
     >: ClonableView,
     Self: linera_views::views::View,
 {
-    fn clone_unchecked(&mut self) -> Self {
-        Self {
-            register: self.register.clone_unchecked(),
-            collection: self.collection.clone_unchecked(),
-        }
+    fn clone_unchecked(&mut self) -> Result<Self, linera_views::ViewError> {
```
MathieuDutSik, line 15: The name `clone_unchecked` is quite unclear about what it means.

linera-views/src/views/reentrant_collection_view.rs
```diff
@@ -219,20 +219,19 @@ impl<W: ClonableView> ClonableView for ReentrantByteCollectionView<W::Context, W
...
                             .try_write()
-                            .expect("Unable to acquire write lock during clone_unchecked");
-
-                        Update::Set(Arc::new(RwLock::new(view.clone_unchecked())))
+                            .ok_or_else(|| ViewError::TryLockError(key.clone()))?;
+                        Update::Set(Arc::new(RwLock::new(view.clone_unchecked()?)))
                     }
                 };
-                (key.clone(), cloned_value)
+                Ok::<_, ViewError>((key.clone(), cloned_value))
```
MathieuDutSik, line 226: We know from the state prior to #4362 that the type annotation is not needed.

linera-views/src/error.rs
```diff
@@ -8,10 +8,6 @@ pub enum ViewError {
     #[error(transparent)]
     BcsError(#[from] bcs::Error),
 
-    /// We failed to acquire an entry in a `CollectionView`.
-    #[error("trying to access a collection view while some entries are still being accessed")]
-    CannotAcquireCollectionEntry,
-
```
MathieuDutSik, line 14: I think we kept this for compatibility reasons with Testnet Conway.

MathieuDutSik (commented):
I have some problems with this PR, which effectively reverses #4362 / #4368.

The very name of `clone_unchecked` and the documentation around the trait in `linera-views/src/view/mod.rs` makes it clear that the safety of views obtained from `clone_unchecked` depends on the care of the programmer and that Rust cannot help you here.

At the same time, when we return a `TryLockError` in `ReentrantCollectionView`, we have no way of processing that error. So, the functional change of throwing of `TryLockError` appears null.

There are two other issues with the function `clone_unchecked` in my opinion:
* The `&mut self` argument. That is for one single reason of `hash: Mutex::new(*self.hash.get_mut().unwrap())` in `hashable_wrapper` and `key_value_store_view`.
* Why not make the function `clone_unchecked` async? This would reliably address the case of concurrent access to the view and avoid the error.

linera-views/src/error.rs
ma2bd: We don't serialize `ViewError` over the network

linera-views/src/views/reentrant_collection_view.rs
ma2bd: I'm seeing `.collect::<Result<_, ViewError>>()?;` in at least one place

linera-views-derive/src/snapshots/linera_views_derive__tests__generate_clonable_view_code-3.snap
ma2bd: Not sure I see the problem but that's why we have hopefully clarifying comments on the trait itself.

ma2bd:
> I have some problems with this PR, which effectively reverses #4362 / #4368.

I missed the fact that #4362 introduced a nontrivial `.expect`. Otherwise, I would not have accepted it. (Or at least I hope so)

> The very name of `clone_unchecked` and the documentation around the trait in `linera-views/src/view/mod.rs` makes it clear that the safety of views obtained from `clone_unchecked` depends on the care of the programmer and that Rust cannot help you here.

I guess there is "uncheck" and "uncheck". Panicking is still surprising. In this codebase, we only panic for a handful unrecoverable errors like OOMs. `TryLockError` is not of them.

> At the same time, when we return a `TryLockError` in `ReentrantCollectionView`, we have no way of processing that error. 

In other languages, it's fairly common to have runtime errors when mutating and iterating on a collection.

> There are two other issues with the function `clone_unchecked` in my opinion:
> 
> * The `&mut self` argument. That is for one single reason of `hash: Mutex::new(*self.hash.get_mut().unwrap())` in `hashable_wrapper` and `key_value_store_view`.

No, that is not the reason. The goal is to guarantee that one cannot mutate the database by mistake with a shared reference on a view.

> * Why not make the function `clone_unchecked` async? This would reliably address the case of concurrent access to the view and avoid the error.

I believe returning `TryLockError` is consistent with the other functions of `ReentrantCollectionView` taking `&mut self` where we expect all the sub-locks to be immediately available. Waiting would create the possibility of complex deadlocks (or livelocks due async threads being canceled).

linera-views/src/views/reentrant_collection_view.rs
MathieuDutSik: Ok, minor point anyway.

MathieuDutSik:
> > I have some problems with this PR, which effectively reverses #4362 / #4368.
> 
> I missed the fact that #4362 introduced a nontrivial `.expect`. Otherwise, I would not have accepted it. (Or at least I hope so)

Ok.

> > The very name of `clone_unchecked` and the documentation around the trait in `linera-views/src/view/mod.rs` makes it clear that the safety of views obtained from `clone_unchecked` depends on the care of the programmer and that Rust cannot help you here.
> 
> I guess there is "uncheck" and "uncheck". Panicking is still surprising. In this codebase, we only panic for a handful unrecoverable errors like OOMs. `TryLockError` is not of them.

Ok, but if we do not process the `TryLockError` then this is equivalent as panicking.

By "processing", I mean the stuff we do for `BlobsNotFound`. Catch the error and do something about it that is different from stopping completely the application.

> > At the same time, when we return a `TryLockError` in `ReentrantCollectionView`, we have no way of processing that error.
> 
> In other languages, it's fairly common to have runtime errors when mutating and iterating on a collection.

We had a runtime error recently when calling `fn entries` of `CollectionView`.
Sure, the error was a `CannotAcquireCollectionEntry`. But that was anyway terminating the run of the application.

> > There are two other issues with the function `clone_unchecked` in my opinion:
> > 
> > * The `&mut self` argument. That is for one single reason of `hash: Mutex::new(*self.hash.get_mut().unwrap())` in `hashable_wrapper` and `key_value_store_view`.
> 
> No, that is not the reason. The goal is to guarantee that one cannot mutate the database by mistake with a shared reference on a view.

Ok, thank you.

> > * Why not make the function `clone_unchecked` async? This would reliably address the case of concurrent access to the view and avoid the error.
> 
> I believe returning `TryLockError` is consistent with the other functions of `ReentrantCollectionView` taking `&mut self` where we expect all the sub-locks to be immediately available. Waiting would create the possibility of complex deadlocks (or livelocks due async threads being canceled).

Ok, I agree.
That was one way to avoid the `TryLockError`. But yes, it introduces its own complications.

linera-views/src/error.rs
MathieuDutSik: Ok. Then I would have removed it when eliminating the possibility of this error in `CollectionView`.

PR #4823, opened 2025-10-20
Introduce reading of keys by blocks for DynamoDb.

MathieuDutSik:
## Motivation

For `DynamoDB`, we are executing `read_multi_values` queries by creating `N` queries for `N` access to the storage. One for each key. But it is possible to group the queries so that we have fewer queries.

## Proposal

We use the `BatchGetItem` functionality of `DynamoDB`.
We cannot group all reads in one single query. But instead of sending `N` queries, we send about `N/40`, which is a significant gain.

## Test Plan

The CI.

## Release Plan

It does not appear important to put it on TestNet Conway since DynamoDB is not used for it.

## Links

None.

### DIFF
--- a/linera-views/src/backends/dynamo_db.rs
+++ b/linera-views/src/backends/dynamo_db.rs
@@ -158,7 +158,6 @@ const MAX_TRANSACT_WRITE_ITEM_SIZE: usize = 100;
 /// Since the maximum size of a value is 400K, this gets us 40 as upper limit
 const MAX_BATCH_GET_ITEM_SIZE: usize = 40;
 
-
 /// Builds the key attributes for a table item.
 ///
 /// The key is composed of two attributes that are both binary blobs. The first attribute is a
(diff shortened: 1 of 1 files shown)


ma2bd:
> For DynamoDb we were doing read_multi_values with handles.

What does this mean? It seems like the new code has some form of query batching. Is that the reason for this PR??

MathieuDutSik:
> > For DynamoDb we were doing read_multi_values with handles.
> 
> What does this mean? It seems like the new code has some form of query batching. Is that the reason for this PR??

Yes, we cannot avoid having some batching. We have that as well in ScyllaDb.

In the code in the main for N keys, we create N green threads instance. This is inefficient. In the code of this PR we create about N/40 green threads.

ma2bd:
> > > For DynamoDb we were doing read_multi_values with handles.
> > 
> > 
> > What does this mean? It seems like the new code has some form of query batching. Is that the reason for this PR??
> 
> Yes, we cannot avoid having some batching. We have that as well in ScyllaDb.
> 
> In the code in the main for N keys, we create N green threads instance. This is inefficient. In the code of this PR we create about N/40 green threads.

Please fix the summary!

ma2bd (approved):

PR #4824, opened 2025-10-20
Replace execution state hash with zeroes in high epochs

bart-linera:
## Motivation

Hashing gets slow when the execution state grows to large sizes.

## Proposal

Temporarily - for testnet only - skip the computation of the execution state hash and return zeroes in high epochs (>20 in this PR).

## Test Plan

CI

## Release Plan

- Release a new SDK
- Update all validators
- Move to epoch 20 once that's done

## Links

### DIFF
--- a/linera-execution/src/lib.rs
+++ b/linera-execution/src/lib.rs
@@ -89,7 +89,7 @@ const MAX_STREAM_NAME_LEN: usize = 64;
 /// current committee, causes the execution state not to be hashed, and instead the hash
 /// returned to be all zeros.
 // Note: testnet-only! This should not survive to mainnet.
-const FLAG_ZERO_HASH: &str = "FLAG_ZERO_HASH.linera.network";
+pub const FLAG_ZERO_HASH: &str = "FLAG_ZERO_HASH.linera.network";
 
 /// An implementation of [`UserContractModule`].
 #[derive(Clone)]
--- a/linera-service/tests/local_net_tests.rs
+++ b/linera-service/tests/local_net_tests.rs
@@ -22,6 +22,7 @@ use linera_base::{
     vm::VmRuntime,
 };
 use linera_core::{data_types::ChainInfoQuery, node::ValidatorNode};
+use linera_execution::FLAG_ZERO_HASH;
 use linera_sdk::linera_base_types::AccountSecretKey;
 use linera_service::{
     cli_wrappers::{
@@ -87,7 +88,7 @@ async fn test_end_to_end_reconfiguration(config: LocalNetConfig) -> Result<()> {
         .open_and_assign(&client_2, Amount::from_tokens(3))
         .await?;
     let port = get_node_port().await;
-    let node_service_2 = match network {
+    let mut node_service_2 = match network {
         Network::Grpc | Network::Grpcs => {
             Some(client_2.run_node_service(port, ProcessInbox::Skip).await?)
         }
@@ -237,7 +238,7 @@ async fn test_end_to_end_reconfiguration(config: LocalNetConfig) -> Result<()> {
         )
         .await?;
 
-    if let Some(mut service) = node_service_2 {
+    if let Some(service) = &mut node_service_2 {
         assert!(
             eventually(|| async { !service.process_inbox(&chain_2).await.unwrap().is_empty() })
                 .await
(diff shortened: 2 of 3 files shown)


linera-execution/src/execution.rs
```diff
@@ -47,6 +49,51 @@ pub struct ExecutionStateView<C> {
...
+                .await
+                .map(|hash| CryptoHash::new(&ExecutionStateHash(hash.into())))
+        }
+    }
+
+    async fn crypto_hash_mut(&mut self) -> Result<CryptoHash, ViewError> {
+        if self
+            .system
+            .current_committee()
+            .is_some_and(|(epoch, _)| epoch > EPOCH_STOP_HASHING)
```
deuszx: greater or greater-or-equal?

bart-linera: Yeah, it depends on when exactly we want to stop hashing.

linera-execution/src/execution.rs
```diff
@@ -47,6 +49,51 @@ pub struct ExecutionStateView<C> {
...
+        } else {
+            #[derive(Serialize, Deserialize)]
+            struct ExecutionStateHash([u8; 32]);
+            impl BcsHashable<'_> for ExecutionStateHash {}
+            self.hash_mut()
+                .await
+                .map(|hash| CryptoHash::new(&ExecutionStateHash(hash.into())))
+        }
+    }
+}
```
deuszx, line 78: Ideally we'd have a unit test for this as well. Do you think it's doable?

bart-linera: I can try to add one :+1:

linera-execution/src/execution.rs
ma2bd: `>=` makes sense to me. We stop hashing at `EPOCH_STOP_HASHING`, right?

linera-execution/src/execution.rs
```diff
@@ -47,6 +49,51 @@ pub struct ExecutionStateView<C> {
     pub stream_event_counts: MapView<C, StreamId, u32>,
 }
 
+// The epoch after which, instead of hashing the view, we just return zeroes for
+// performance.
+// Note: testnet-only! This should not survive to mainnet.
+const EPOCH_STOP_HASHING: Epoch = Epoch(20);
+
+impl<C> CryptoHashView for ExecutionStateView<C>
```
ma2bd: Let me backport #4818, then we just need `crypto_hash_mut` -- avoiding code duplication.

afck:
Actually, I think that means we can't use the execution state cache anymore? It stores the state by hash.

bart-linera:
Hmm, good point. So I guess the options are:
- Disable the cache
- Give up on this change
- Look for some other identifier for the caching purposes. Proposal hash, maybe? A proposal should never be too large, and being for a specific chain ID and height, it should also determine the final execution state uniquely.

afck:
Good idea!

> it should also determine the final execution state uniquely

…up to oracles… but I think that's fine in this case? It just means we won't needlessly consult the oracles again.

Alternatively, can't we just use the block hash? Whenever we have an execution state, we probably compute the resulting block hash for after the execution shortly after.

bart-linera:
Let me have a look into how we use the cache, because I don't know if we have the block hash when we want to read from the cache.

EDIT: We do, so that might be the way to go :+1: I'll have a go at implementing it.

linera-execution/src/execution.rs
bart-linera: Done in [317579c](https://github.com/linera-io/linera-protocol/pull/4824/commits/317579c8983c4e1e7a458c0b7e33108fe01d8c17)

linera-execution/src/execution.rs
bart-linera: It is backported now, but we still have to implement `crypto_hash`, as it still exists in the trait, don't we?

linera-execution/src/lib.rs
```diff
@@ -85,6 +85,11 @@ pub const LINERA_TYPES_SOL: &str = include_str!("../solidity/LineraTypes.sol");
 /// The maximum length of a stream name.
 const MAX_STREAM_NAME_LEN: usize = 64;
 
+// The epoch after which, instead of hashing the view, we just return zeroes for
+// performance.
```
afck:
```suggestion
/// The epoch after which, instead of hashing the view, we just return zeroes for
/// performance.
```

linera-execution/src/unit_tests/system_tests.rs
```diff
@@ -129,3 +132,28 @@ async fn empty_accounts_are_removed() -> anyhow::Result<()> {
...
+    view.system.committees.set(committees.clone());
+    view.system.epoch.set(epoch);
+    // The hash should still be nonzero before the threshold epoch.
+    assert_ne!(view.crypto_hash_mut().await?, zero_hash);
+
+    committees.insert(EPOCH_STOP_HASHING, Committee::default());
+    view.system.committees.set(committees);
+    view.system.epoch.set(EPOCH_STOP_HASHING);
+    // Starting from this epoch, the hash should be all zeros.
+    assert_eq!(view.crypto_hash_mut().await?, zero_hash);
```
afck, line 160: You could also check `crypto_hash` wherever you check `crypto_hash_mut`.

afck:
> Nothing to do

A lot to do! We need to release a new SDK, update all validators, and only once that's done move to epoch 20.

linera-execution/src/execution.rs
ma2bd: You don't have to implement the trait

linera-execution/src/execution.rs
```diff
@@ -47,6 +49,46 @@ pub struct ExecutionStateView<C> {
...
+                .await
+                .map(|hash| CryptoHash::new(&ExecutionStateViewHash(hash.into())))
+        }
+    }
+
+    async fn crypto_hash_mut(&mut self) -> Result<CryptoHash, ViewError> {
+        if self
+            .system
+            .current_committee()
+            .is_some_and(|(epoch, _)| epoch >= EPOCH_STOP_HASHING)
```
ma2bd: I've been thinking: instead of using a fixed constant `EPOCH_STOP_HASHING` (raising some questions for local test networks), we could instead repurpose one of the fields of `ResourceControlPolicy` to be used as a switch.

ma2bd:
@bart-linera Can we have an e2e test with reconfigurations?

bart-linera:
Note: it's not straightforward to check the execution state hash in an end-to-end test in code, but I ran it locally and saw that the execution state hash does indeed become all zeros after the last reconfiguration.

ma2bd:
> Note: it's not straightforward to check the execution state hash in an end-to-end test in code, but I ran it locally and saw that the execution state hash does indeed become all zeros after the last reconfiguration.

What we want is test liveness after the reconfiguration.

bart-linera:
Yes, the change I made executes a transfer after the reconfiguration and asserts it went through, so liveness is being tested.

linera-service/tests/local_net_tests.rs
```diff
@@ -271,6 +272,54 @@ async fn test_end_to_end_reconfiguration(config: LocalNetConfig) -> Result<()> {
...
+        let epochs = committees.into_keys().collect::<Vec<_>>();
+        assert_eq!(&epochs, &[Epoch(3), Epoch(4)]);
+
+        service.ensure_is_running()?;
+    } else {
+        client_2.sync(chain_2).await?;
+        client_2.process_inbox(chain_2).await?;
+    }
+
+    // Check that we can still produce blocks
```
ma2bd, line 293: nice!

Issue #4802, opened 2025-10-14
Introduce `multi_get`, `contains_key`, `try_load_entries` variants taking a lambda function.


MathieuDutSik:
We have several functions where we access a list of values from a list of keys, and the result is passed into a for loop with `zip`.

Maybe we could instead pass a lambda function that process the pairs `(key,value)` that are obtained.

afck:
Why pass in a function? Why not return `impl Iterator<Item = (Key, Value)>`?

MathieuDutSik:
Closed by #4827

PR #4827, opened 2025-10-20
Introduce `multi_get_pairs` operations and group another access to `preprocessed_blocks`.

MathieuDutSik:
## Motivation

There are more uses of `multi_get` that are possible to group.
We also introduced some `multi_get_pairs` functions in order to avoid some `.zip` constructions.

#4802 

## Proposal

The `multi_get_pairs` is introduced for `*MapView` and `LogView`.
For the `*CollectionView`, we introduce similar functions.

The pattern in `state.rs` was grouped. When possible the `.zip` were eliminated.

## Test Plan

The CI.

## Release Plan

## Links

None.

### DIFF
--- a/linera-views/src/views/collection_view.rs
+++ b/linera-views/src/views/collection_view.rs
@@ -415,7 +415,7 @@ impl<W: View> ByteCollectionView<W::Context, W> {
         Ok(results)
     }
 
-    /// Load multiple entries for reading at once with their keys.
+    /// Loads multiple entries for reading at once with their keys.
     /// ```rust
     /// # tokio_test::block_on(async {
     /// # use linera_views::context::MemoryContext;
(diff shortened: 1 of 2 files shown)


linera-views/src/views/collection_view.rs
```diff
@@ -415,6 +415,37 @@ impl<W: View> ByteCollectionView<W::Context, W> {
         Ok(results)
     }
 
+    /// Load multiple entries for reading at once with their keys.
```
afck:
```suggestion
    /// Loads multiple entries for reading at once with their keys.
```

(Also below.)

afck (approved):

linera-views/src/views/collection_view.rs
MathieuDutSik: Ok, thank you.

PR #4833, opened 2025-10-21
`change-validators` command.

afck:
## Motivation

`set-validator` and `remove-validator` each create a new epoch just to add or remove a single validator. We sometimes want to add or remove multiple validators, and creating lots of new epochs increases the protocol overhead when syncing the admin chain.

## Proposal

Add `change-validators` to add and remove multiple validators in one epoch.

Also, remove the outdated `process_inbox` call and log message.

## Test Plan

The reconfiguration test was updated to use this command, too.

## Release Plan

- These changes should be backported `testnet_conway`, then
    - be released in a new SDK.

## Links

### DIFF
--- a/linera-service/src/cli/main.rs
+++ b/linera-service/src/cli/main.rs
@@ -696,7 +696,7 @@ impl Runnable for Job {
                                 }
                                 RemoveValidator { public_key } => {
                                     if validators.remove(&public_key).is_none() {
-                                        warn!("Validator {public_key} does not exist; aborting.");
+                                        error!("Validator {public_key} does not exist; aborting.");
                                         return Ok(ClientOutcome::Committed(None));
                                     }
                                 }
@@ -709,7 +709,7 @@ impl Runnable for Job {
                                     // Validate that all validators to add do not already exist.
                                     for validator in &add_validators {
                                         if validators.contains_key(&validator.public_key) {
-                                            warn!(
+                                            error!(
                                                 "Cannot add existing validator: {}. Aborting operation.",
                                                 validator.public_key
                                             );
@@ -720,7 +720,7 @@ impl Runnable for Job {
                                     for validator in &modify_validators {
                                         match validators.get(&validator.public_key) {
                                             None => {
-                                                warn!(
+                                                error!(
                                                     "Cannot modify nonexistent validator: {}. Aborting operation.",
                                                     validator.public_key
                                                 );
@@ -732,7 +732,7 @@ impl Runnable for Job {
                                                     && existing.account_public_key == validator.account_key
                                                     && existing.votes == validator.votes
                                                 {
-                                                    warn!(
+                                                    error!(
                                                         "Validator {} is not being modified. Aborting operation.",
                                                         validator.public_key
                                                     );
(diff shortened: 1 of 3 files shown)


linera-service/src/cli/main.rs
```diff
@@ -684,10 +695,40 @@ impl Runnable for Job {
...
+                                    for public_key in &remove_validators {
+                                        if !validators.contains_key(public_key) {
+                                            warn!(
+                                                "Cannot remove nonexistent validator: {public_key}. Aborting operation."
+                                            );
+                                            return Ok(ClientOutcome::Committed(None));
+                                        }
+                                    }
+                                    // Add validators
+                                    for validator in add_validators {
```
deuszx, line 754: Do we want to support adding a validator that already is in the validator set?

afck: Yes, just like in `set-validator` we might want to update their account key or weight etc.

afck: Do you think we should change `--add` to `--set` then, for consistency? :thinking:

deuszx: It is bit transparent. For removal you exit when caller tries to remove a non-existent validator even though it'd be a harmless no-op. Here it might be done by a mistake and change the weight.

Personally, I would require the implicit `set` call.

afck: You mean you'd have `--add`, `--set` and `--remove`, and in the first case require that the validator isn't in the committee yet, and in the other two cases that it is?

afck: (Or `--modify` rather than `--set`? Or should one just combine `--remove` and `--add` to do that?)

deuszx: `--add`, `--remove`, `--set` (or modify).

ndr-ds: IMHO either we have `--add`, `--remove` and `--modify`, or we have `--set` (that adds and modifies) and `--remove`. The latter is consistent with the current code's behavior. If that behavior's what we want or not is up for discussion :)

afck: Good point; done in https://github.com/linera-io/linera-protocol/pull/4833/commits/fda3d90234f7fcb21163c9f498c864a85d6336d1.

linera-service/src/cli/main.rs
```diff
@@ -684,10 +696,87 @@ impl Runnable for Job {
                                 }
                                 RemoveValidator { public_key } => {
                                     if validators.remove(&public_key).is_none() {
-                                        warn!("Skipping removal of nonexistent validator");
+                                        warn!("Validator {public_key} does not exist; aborting.");
```
eldios: I'd remove `aborting.` since that assumes execution stops.

if we actually interrupt execution, this should be an error, IMO

linera-service/src/cli/main.rs
```diff
@@ -684,10 +696,87 @@ impl Runnable for Job {
...
+                                    add_validators,
+                                    modify_validators,
+                                    remove_validators,
+                                    skip_online_check: _,
+                                } => {
+                                    // Validate that all validators to add do not already exist.
+                                    for validator in &add_validators {
+                                        if validators.contains_key(&validator.public_key) {
+                                            warn!(
+                                                "Cannot add existing validator: {}. Aborting operation.",
```
eldios, line 713: same as above, if we abort the command, this should be an error, otherwise just leave the warning without the aborting part.

linera-service/src/cli/main.rs
afck: These all do interrupt execution. Turning them into errors…

afck: Done in cbf2ecedf8d2cd4b0d2143a4edd81dc24ed4c91c.

PR #4834, opened 2025-10-21
Add `NotificationsExt` trait to await specific notifications.

### LINKED ISSUE #4718: Use notifications in the tests instead of `sync` and `eventually`.
The `linera-service` tests that use the node service explicitly synchronize from validators or use `eventually` to wait for certain conditions in lots of places. Neither should be necessary, if instead we figure out exactly what notifications we expect, and then wait for those.

afck:
## Motivation

The test-only `eventually` function waits in a loop with increasing delays until a condition is satisfied. But usually these conditions are satisfied exactly when the node processes a certain block and outputs the corresponding notification.

## Proposal

Add a test-only `NotificationsExt` trait to simplify waiting for a particular notification. Use it instead of several `eventually` and `sync` calls in the tests.

## Test Plan

Most tests were changed to use this instead of `eventually`. Only the ones waiting for the faucet weren't, because it currently doesn't support GraphQL subscriptions for notifications.

## Release Plan

- These changes _could_ be backported to `testnet_conway`, to keep the diff small.

## Links

- Closes #4718.

### DIFF
--- a/linera-service/src/cli_wrappers/wallet.rs
+++ b/linera-service/src/cli_wrappers/wallet.rs
@@ -1047,6 +1047,35 @@ impl ClientWrapper {
         Ok(())
     }
 
+    pub async fn change_validators(
+        &self,
+        add_validators: &[(String, String, usize, usize)], // (public_key, account_key, port, votes)
+        modify_validators: &[(String, String, usize, usize)], // (public_key, account_key, port, votes)
+        remove_validators: &[String],
+    ) -> Result<()> {
+        let mut command = self.command().await?;
+        command.arg("change-validators");
+
+        for (public_key, account_key, port, votes) in add_validators {
+            let address = format!("{}:127.0.0.1:{}", self.network.short(), port);
+            let validator_spec = format!("{public_key},{account_key},{address},{votes}");
+            command.args(["--add", &validator_spec]);
+        }
+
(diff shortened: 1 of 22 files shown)


linera-service/src/cli_wrappers/wallet.rs
```diff
@@ -1698,3 +1714,112 @@ impl<A> From<String> for ApplicationWrapper<A> {
...
+        loop {
+            let notification = futures::select! {
+                () = timeout => bail!("Timeout waiting for notification"),
+                notification = self.next().fuse() => notification.context("Stream closed")??,
+            };
+            if let Reason::NewIncomingBundle { height, origin } = notification.reason {
+                if expected_height.is_none_or(|h| h == height) && origin == expected_origin {
+                    return Ok(());
+                }
+            }
```
deuszx, line 1842: All three functions have almost the same implementation except for:
- `Reason` variant they await for
- the subsequent check

nit: maybe there's a clean way to extract it to a function and just pass the closure that takes `notification` and returns `Ok(T)`.

afck: Done in https://github.com/linera-io/linera-protocol/pull/4834/commits/fad81d6347dd90964319acc2867a156899e1af12.

PR #4835, opened 2025-10-22
Implement migration on existing old storage.

MathieuDutSik:
## Motivation

We are working with several storage. One being the storage of the old schema (where blobs are all on the same partition) and the new schema (where blobs are spread out).

## Proposal

When creating the storage instance, check the version by accessing the `SchemaDescription`. If incorrect, then upgrade.

Note:
* It would be advantageous to disable cache since during migration, we know it is not 

## Test Plan

Tests are clearly missing in this PR and need to be added.

## Release Plan

The goal is to put this code in TestNet Conway in order to deal with ScyllaDb performance.

## Links

None.

### DIFF
--- a/linera-base/src/identifiers.rs
+++ b/linera-base/src/identifiers.rs
@@ -604,6 +604,8 @@ impl std::str::FromStr for StreamId {
     Debug,
     PartialEq,
     Eq,
+    Ord,
+    PartialOrd,
     Hash,
     Clone,
     Serialize,
--- a/linera-base/src/identifiers.rs
+++ b/linera-base/src/identifiers.rs
@@ -604,8 +604,6 @@ impl std::str::FromStr for StreamId {
     Debug,
     PartialEq,
     Eq,
-    Ord,
-    PartialOrd,
     Hash,
     Clone,
     Serialize,
--- a/linera-storage/src/db_storage.rs
+++ b/linera-storage/src/db_storage.rs
@@ -296,7 +296,7 @@ impl MultiPartitionBatch {
     fn add_event(&mut self, event_id: EventId, value: Vec<u8>) -> Result<(), ViewError> {
         #[cfg(with_metrics)]
         metrics::WRITE_EVENT_COUNTER.with_label_values(&[]).inc();
-        let key = event_key(&event_id);
+        let key = to_event_key(&event_id);
         let root_key = RootKey::Event(event_id.chain_id).bytes();
         self.put_key_value_bytes(root_key, key, value);
         Ok(())
@@ -343,10 +343,18 @@ impl RootKey {
     }
 }
 
-pub(crate) fn event_key(event_id: &EventId) -> Vec<u8> {
-    let mut key = bcs::to_bytes(&event_id.stream_id).unwrap();
-    key.extend(bcs::to_bytes(&event_id.index).unwrap());
-    key
+#[derive(Debug, Serialize, Deserialize)]
+pub(crate) struct RestrictedEventId {
+    pub stream_id: StreamId,
+    pub index: u32,
+}
+
+pub(crate) fn to_event_key(event_id: &EventId) -> Vec<u8> {
+    let restricted_event_id = RestrictedEventId {
+        stream_id: event_id.stream_id.clone(),
+        index: event_id.index,
+    };
+    bcs::to_bytes(&restricted_event_id).unwrap()
 }
 
 fn is_chain_state(root_key: &[u8]) -> bool {
@@ -372,7 +380,7 @@ mod tests {
     };
 
     use crate::db_storage::{
-        event_key, RootKey, BLOB_ID_LENGTH, BLOB_ID_TAG, CHAIN_ID_LENGTH, CHAIN_ID_TAG,
+        to_event_key, RootKey, BLOB_ID_LENGTH, BLOB_ID_TAG, CHAIN_ID_LENGTH, CHAIN_ID_TAG,
     };
 
     // Several functionalities of the storage rely on the way that the serialization
@@ -424,7 +432,7 @@ mod tests {
             stream_id,
             index,
         };
-        let key = event_key(&event_id);
+        let key = to_event_key(&event_id);
         assert!(key.starts_with(&prefix));
     }
 }
@@ -887,7 +895,7 @@ where
 
     #[instrument(skip_all, fields(event_id = ?event_id))]
     async fn read_event(&self, event_id: EventId) -> Result<Option<Vec<u8>>, ViewError> {
-        let event_key = event_key(&event_id);
+        let event_key = to_event_key(&event_id);
         let root_key = RootKey::Event(event_id.chain_id).bytes();
         let store = self.database.open_shared(&root_key)?;
         let event = store.read_value_bytes(&event_key).await?;
@@ -898,7 +906,7 @@ where
 
     #[instrument(skip_all, fields(event_id = ?event_id))]
     async fn contains_event(&self, event_id: EventId) -> Result<bool, ViewError> {
-        let event_key = event_key(&event_id);
+        let event_key = to_event_key(&event_id);
         let root_key = RootKey::Event(event_id.chain_id).bytes();
         let store = self.database.open_shared(&root_key)?;
         let exists = store.contains_key(&event_key).await?;
--- a/linera-storage/src/db_storage.rs
+++ b/linera-storage/src/db_storage.rs
@@ -226,30 +226,30 @@ pub mod metrics {
 }
 
 /// The default key used when the root_key contains the information.
-const DEFAULT_KEY: &[u8] = &[0];
+pub(crate) const DEFAULT_KEY: &[u8] = &[0];
 
 /// The second key used when the root_key contains the information.
-const ONE_KEY: &[u8] = &[1];
+pub(crate) const ONE_KEY: &[u8] = &[1];
 
 fn get_01_keys() -> Vec<Vec<u8>> {
     vec![vec![0], vec![1]]
 }
 
 #[derive(Default)]
-struct MultiPartitionBatch {
+pub(crate) struct MultiPartitionBatch {
     keys_value_bytes: Vec<(Vec<u8>, Vec<u8>, Vec<u8>)>,
 }
 
 impl MultiPartitionBatch {
-    fn new() -> Self {
+    pub(crate) fn new() -> Self {
         Self::default()
     }
 
-    fn put_key_value_bytes(&mut self, root_key: Vec<u8>, key: Vec<u8>, value: Vec<u8>) {
+    pub(crate) fn put_key_value_bytes(&mut self, root_key: Vec<u8>, key: Vec<u8>, value: Vec<u8>) {
         self.keys_value_bytes.push((root_key, key, value));
     }
 
-    fn put_key_value<T: Serialize>(
+    pub(crate) fn put_key_value<T: Serialize>(
         &mut self,
         root_key: Vec<u8>,
         value: &T,
@@ -319,7 +319,7 @@ impl MultiPartitionBatch {
 /// Main implementation of the [`Storage`] trait.
 #[derive(Clone)]
 pub struct DbStorage<Database, Clock = WallClock> {
-    database: Arc<Database>,
+    pub(crate) database: Arc<Database>,
     clock: Clock,
     wasm_runtime: Option<WasmRuntime>,
     user_contracts: Arc<papaya::HashMap<ApplicationId, UserContractCode>>,
@@ -328,7 +328,7 @@ pub struct DbStorage<Database, Clock = WallClock> {
 }
 
 #[derive(Debug, Serialize, Deserialize)]
-enum RootKey {
+pub(crate) enum RootKey {
     ChainState(ChainId),
     CryptoHash(CryptoHash),
     Blob(BlobId),
@@ -338,12 +338,12 @@ enum RootKey {
 }
 
 impl RootKey {
-    fn bytes(&self) -> Vec<u8> {
+    pub(crate) fn bytes(&self) -> Vec<u8> {
         bcs::to_bytes(self).unwrap()
     }
 }
 
-fn event_key(event_id: &EventId) -> Vec<u8> {
+pub(crate) fn event_key(event_id: &EventId) -> Vec<u8> {
     let mut key = bcs::to_bytes(&event_id.stream_id).unwrap();
     key.extend(bcs::to_bytes(&event_id.index).unwrap());
     key
@@ -403,7 +403,7 @@ mod tests {
     }
 
     // The listing of the events in `read_events_from_index` depends on the
-    // serialization of `BaseKey::Event`.
+    // serialization of `RootKey::Event`.
     #[test]
     fn test_root_key_event_serialization() {
         let hash = CryptoHash::test_hash("49");
@@ -1035,7 +1035,7 @@ where
     }
 
     #[instrument(skip_all, fields(batch_size = batch.keys_value_bytes.len()))]
-    async fn write_batch(&self, batch: MultiPartitionBatch) -> Result<(), ViewError> {
+    pub(crate) async fn write_batch(&self, batch: MultiPartitionBatch) -> Result<(), ViewError> {
         if batch.keys_value_bytes.is_empty() {
             return Ok(());
         }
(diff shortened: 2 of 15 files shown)


MathieuDutSik:
The migration code is currently broken since it assumes that only one process is doing the migration.

Note: Now addressed.

linera-storage/src/db_storage.rs
```diff
@@ -1049,7 +1049,13 @@ where
     }
 }
 
-impl<Database, C> DbStorage<Database, C> {
+impl<Database, C> DbStorage<Database, C>
+where
+    Database: KeyValueDatabase + Clone + Send + Sync + 'static,
+    Database::Error: Send + Sync,
+    Database::Store: KeyValueStore + Clone + Send + Sync + 'static,
+    C: Clock + Clone + Send + Sync + 'static,
```
ma2bd, line 1066: This should not be needed

linera-storage/src/db_storage.rs
```diff
@@ -1157,3 +1163,168 @@ where
...
     }
 }
+
+#[derive(Default, Debug, Eq, PartialEq, Serialize, Deserialize)]
+#[repr(u8)]
+enum SchemaDescription {
+    /// Version 0, all the blobs, certificates, confirmed blocks, events and network description on the same partition.
+    /// This is marked as default since it does not exist in the old scheme and would be obtained from unwrap_or_default
+    #[default]
+    Version0SingleBlobPartition,
```
ma2bd: nit: `Version0`, `Version1`, etc feels cleaner. We have the comments (and the code) to read if we want the details.

linera-storage/src/db_storage.rs
```diff
@@ -1157,3 +1163,168 @@ where
...
+    /// Version 1, spreading by ChainId, CryptoHash, and BlobId.
+    Version1MultiBlobPartition,
+}
+
+/// The key containing the schema of the storage.
+const SCHEMA_ROOT_KEY: &[u8] = &[0, 1, 1, 2, 3, 5, 8, 13, 21, 34, 55, 89, 144, 233];
```
ma2bd: What are these bytes made of? Please use either:
* a unit test to keep the bytes consistent with the formula.
* `once_cell::sync::Lazy` to compute the bytes from the formula (once) at runtime.

linera-storage/src/db_storage.rs
```diff
@@ -1157,3 +1163,168 @@ where
...
+    Blob(BlobId),
+    BlobState(BlobId),
+    Event(EventId),
+    BlockExporterState(u32),
+    NetworkDescription,
+}
+
+const EMPTY_KEY: &[u8] = &[];
+// The keys corresponding to `ChainState` and `BlockExporterState` are not moved.
+const MOVABLE_KEYS_0_1: &[u8] = &[1, 2, 3, 4, 5, 7];
```
ma2bd: same comment

ma2bd: The explanation is fairly cryptic. It seems like you're referring to the variant tags of `BaseKey`

linera-storage/src/db_storage.rs
```diff
@@ -1157,3 +1163,168 @@ where
...
+
+/// The total number of keys being migrated in a block.
+/// we use chunks to avoid OOM
+const BLOCK_KEY_SIZE: usize = 90;
```
ma2bd: Can we have a CLI option (with an environment variable `LINERA_xx` to set this value)?

ma2bd: Also the name is very vague. Can you change to something like "migration chunk size" or "migration batch size"?

linera-storage/src/db_storage.rs
```diff
@@ -1157,3 +1163,168 @@ where
...
+    C: Clock + Clone + Send + Sync + 'static,
+    Database::Error: Send + Sync,
+{
+    async fn migrate_key_set(
+        &self,
+        first_byte: &u8,
+        base_keys: Vec<Vec<u8>>,
+    ) -> Result<(), ViewError> {
+        tracing::info!(
+            "migrate_key_set with first_byte={first_byte} for |base_keys|={}",
```
ma2bd: INFO logs should be a little more human-friendly than this

linera-storage/src/db_storage.rs
```diff
@@ -1157,3 +1163,168 @@ where
...
+}
+
+impl<Database, C> DbStorage<Database, C>
+where
+    Database: KeyValueDatabase + Clone + Send + Sync + 'static,
+    Database::Store: KeyValueStore + Clone + Send + Sync + 'static,
+    C: Clock + Clone + Send + Sync + 'static,
+    Database::Error: Send + Sync,
+{
+    async fn migrate_key_set(
```
ma2bd: What's a "key set"? Can we use existing concepts instead? E.g. `migrate_shared_root_keys` or `migrate_shared_partitions` come to mind.

linera-storage/src/db_storage.rs
```diff
@@ -1157,3 +1163,168 @@ where
...
+    Version0SingleBlobPartition,
+    /// Version 1, spreading by ChainId, CryptoHash, and BlobId.
+    Version1MultiBlobPartition,
+}
+
+/// The key containing the schema of the storage.
+const SCHEMA_ROOT_KEY: &[u8] = &[0, 1, 1, 2, 3, 5, 8, 13, 21, 34, 55, 89, 144, 233];
+
+#[derive(Debug, Serialize, Deserialize)]
+enum BaseKey {
```
ma2bd: I assume this is the V0 BaseKey.
Can you define a module `Version0` and move the V0 definitions (and migration code?) into this module for namespacing purpose?

linera-storage/src/db_storage.rs
```diff
@@ -1157,3 +1163,168 @@ where
...
+            let mut batch = Batch::new();
+            for key in chunk_base_keys {
+                batch.delete_key(key.to_vec());
+            }
+            store.write_batch(batch).await?;
+        }
+        Ok(())
+    }
+
+    async fn migrate_0_to_1(&self) -> Result<(), ViewError> {
```
ma2bd: migrate_v0_to_v1

linera-storage/src/db_storage.rs
```diff
@@ -1157,3 +1163,168 @@ where
...
+            let base_keys = keys
+                .into_iter()
+                .map(|key| {
+                    let mut base_key = vec![*first_byte];
+                    base_key.extend(key);
+                    base_key
+                })
+                .collect::<Vec<Vec<u8>>>();
+            self.migrate_key_set(first_byte, base_keys).await?;
+            // Some keys can be left in the value-splitting.
```
ma2bd: Then why do we even try to remove keys inside `migrate_key_set`?

linera-storage/src/db_storage.rs
```diff
@@ -1157,3 +1163,168 @@ where
...
+            store.write_batch(batch).await?;
+        }
+        Ok(())
+    }
+
+    async fn migrate_0_to_1(&self) -> Result<(), ViewError> {
+        for first_byte in MOVABLE_KEYS_0_1 {
+            let store = self.database.open_shared(&[])?;
+            let keys = store.find_keys_by_prefix(&[*first_byte]).await?;
+            let base_keys = keys
```
ma2bd: Why are we computing `base_keys` outside `migrate_key_set` if we're going to pass `first_byte` anyway?

linera-storage/src/db_storage.rs
```diff
@@ -1157,3 +1163,168 @@ where
         Ok(Self::new(database, wasm_runtime, clock))
     }
 }
+
+#[derive(Default, Debug, Eq, PartialEq, Serialize, Deserialize)]
+#[repr(u8)]
+enum SchemaDescription {
```
ma2bd: nit: maybe `StorageSchemaVersion` or just `SchemaVersion`?

linera-storage/src/db_storage.rs
```diff
@@ -1157,3 +1163,168 @@ where
...
+                "index={index} processing chunk of size {}",
+                chunk_base_keys.len()
+            );
+            let store = self.database.open_shared(&[])?;
+            let values = store
+                .read_multi_values_bytes(chunk_base_keys.to_vec())
+                .await?;
+            let values = values
+                .into_iter()
+                .map(|value| value.ok_or(ViewError::MissingEntries))
```
ma2bd: we could also crash

ma2bd: The `collect` is useless but in fact, it would be even more memory-efficient to unwrap the values in the loop below.

linera-storage/src/db_storage.rs
```diff
@@ -1157,3 +1163,168 @@ where
...
+            }
+            store.write_batch(batch).await?;
+        }
+        Ok(())
+    }
+
+    async fn migrate_0_to_1(&self) -> Result<(), ViewError> {
+        for first_byte in MOVABLE_KEYS_0_1 {
+            let store = self.database.open_shared(&[])?;
+            let keys = store.find_keys_by_prefix(&[*first_byte]).await?;
```
ma2bd: Could this be an issue for Rocksdb on testnet without https://github.com/linera-io/linera-protocol/pull/4859 ?

linera-storage/src/db_storage.rs
ma2bd: Alternatively (better?), we could have a file `db_storage_v0.rs`

linera-storage/src/db_storage.rs
```diff
@@ -1157,3 +1163,168 @@ where
...
+            self.migrate_key_set(first_byte, base_keys).await?;
+            // Some keys can be left in the value-splitting.
+            let mut batch = Batch::new();
+            batch.delete_key_prefix(vec![*first_byte]);
+            store.write_batch(batch).await?;
+        }
+        Ok(())
+    }
+
+    async fn get_database_schema(&self) -> Result<SchemaDescription, ViewError> {
```
ma2bd: Can we use this for a version check as well? We should not accept to read a database in the wrong format.

linera-storage/src/db_storage.rs
MathieuDutSik: So, that if it gets interrupted in the middle, when we restart, we do not start from 0.

linera-storage/src/db_storage.rs
MathieuDutSik: What do you mean by version check?
I thought we agreed that the format was decided by the existence or not of the schema key.

linera-storage/src/db_storage.rs
MathieuDutSik: No.

linera-storage/src/db_storage.rs

linera-storage/src/db_storage.rs
MathieuDutSik: ok.

linera-storage/src/db_storage.rs
MathieuDutSik: The Fibonacci numbers up to the value fitting into an u8.

linera-storage/src/db_storage.rs
MathieuDutSik: Ok, for shared_partition.

linera-storage/src/db_storage.rs
MathieuDutSik: Yes, we can. But is it a good idea?
The basic size that we can download in ScyllaDb is 99. For DynamoDB, it is 40. So, it is a reasonable value.

Note that if we read 90 keys, what goes down the chain of operations can be more than 90 keys because of the value splitting of the keys.

linera-storage/src/db_storage.rs

linera-storage/src/db_storage.rs
MathieuDutSik: Why?

linera-storage/src/db_storage.rs

linera-storage/src/migration.rs
```diff
@@ -0,0 +1,296 @@
...
+use crate::{
+    Clock,
+    db_storage::{DbStorage, DEFAULT_KEY, event_key, MultiPartitionBatch, ONE_KEY, RootKey},
+};
+
+#[derive(Default, Debug, Eq, PartialEq, Serialize, Deserialize)]
+#[repr(u8)]
+enum SchemaDescription {
+    /// Version 0, all the blobs, certificates, confirmed blocks, events and network description on the same partition.
+    /// This is marked as default since it does not exist in the old scheme and would be obtained from unwrap_or_default
```
afck, line 24:
```suggestion
    /// This is marked as default since it does not exist in the old scheme and would be obtained from `unwrap_or_default`.
```

linera-storage/src/migration.rs
```diff
@@ -0,0 +1,296 @@
...
+    #[default]
+    Version0,
+    /// Version 1, spreading by ChainId, CryptoHash, and BlobId.
```
afck, line 27:
```suggestion
    /// Version 1, spreading by chain ID, crypto hash, and blob ID.
```

(Or `` `ChainId`, `CryptoHash`, and `BlobId` ``.)

linera-storage/src/migration.rs
```diff
@@ -0,0 +1,296 @@
...
+    Version1,
+}
+
+/// The key containing the schema of the storage.
+const SCHEMA_ROOT_KEY: &[u8] = &[0, 1, 1, 2, 3, 5, 8, 13, 21, 34, 55, 89, 144, 233];
```
afck: Why not `[0; 14]`?

linera-storage/src/db_storage.rs
```diff
@@ -328,7 +328,7 @@ pub struct DbStorage<Database, Clock = WallClock> {
 }
 
 #[derive(Debug, Serialize, Deserialize)]
-enum RootKey {
+pub(crate) enum RootKey {
     ChainState(ChainId),
     CryptoHash(CryptoHash),
```
afck, line 333: (Sorry I'm late to the party, but:)
This variant should probably be called `Block` or `Certificate`? If I understand correctly it will contain two entries: a confirmed block, and the certificate (validator signatures) for that block?

linera-storage/src/migration.rs
```diff
@@ -0,0 +1,296 @@
...
+const UNUSED_EMPTY_KEY: &[u8] = &[];
+// The keys corresponding to `ChainState` are not moved.
+// The keys in `BlockExporterState` are moved, but belong to separate root keys.
+const MOVABLE_KEYS_0_1: &[u8] = &[1, 2, 3, 4, 5, 7];
+
+/// The total number of keys being migrated in a block.
+/// we use chunks to avoid OOM
+const BLOCK_KEY_SIZE: usize = 90;
+
+fn map_base_key(base_key: &[u8]) -> Result<(Vec<u8>, Vec<u8>), ViewError> {
```
afck, line 52: So given a base key, this returns the root (aka partition) key and a key for the specific entry inside that partition? (Does the latter have a name?) Would be good to document that.

linera-storage/src/db_storage.rs
ma2bd: It compiled before without. So, why are you adding these lines?

linera-storage/src/db_storage.rs
```diff
@@ -338,12 +338,12 @@ enum RootKey {
 }
```
ma2bd, line 339: Is there a way we could add the `SchemaVersion` cleanly here?
If we were using the first tag (`0`) with an empty key, it would not collide with chain states.

linera-storage/src/migration.rs
ma2bd: or just `[0;1]`?

linera-base/src/identifiers.rs
```diff
@@ -604,6 +604,8 @@ impl std::str::FromStr for StreamId {
     Debug,
     PartialEq,
     Eq,
+    Ord,
+    PartialOrd,
```
ma2bd: `cfg(with_testing)` or use `HashMap` instead of `BTreeMap`?

linera-storage/src/db_storage.rs
```diff
@@ -328,25 +328,34 @@ pub struct DbStorage<Database, Clock = WallClock> {
...
 
 #[derive(Debug, Serialize, Deserialize)]
-enum RootKey {
+pub(crate) enum RootKey {
     ChainState(ChainId),
     CryptoHash(CryptoHash),
     Blob(BlobId),
     Event(ChainId),
-    BlockExporterState(u32),
+    SchemaDescription,
```
ndr-ds, line 336: Shouldn't we add new things to the end of the enum here? AFAIU the DB will have stuff in there where the bcs serialized enum discriminant for `BlockExporterState` will now be read as `SchemaDescription`

linera-storage/src/migration.rs
MathieuDutSik: Eliminated.

linera-storage/src/db_storage.rs
MathieuDutSik: Yes, we can have a Certificate or a confirmed block. So, taking the intersection as being the CryptoHash made things easier in my opinion.

linera-storage/src/db_storage.rs
MathieuDutSik: Yes, we can. Done.

linera-base/src/identifiers.rs
MathieuDutSik: Yes, I could use `HashMap`.

linera-storage/src/db_storage.rs
MathieuDutSik: In the new PR, I changed so that the ChainState and BlockExporter remain at the same position.

linera-storage/src/db_storage.rs
MathieuDutSik: I instead created a file `migration.rs`.

linera-storage/src/db_storage.rs
MathieuDutSik: Added more details.

linera-storage/src/db_storage.rs
MathieuDutSik: It came from an experience of making new async returning an error and doing the migration. Eliminated.

linera-storage/src/migration.rs
MathieuDutSik: Added some documentation. Yes, the "key" itself does not have a fancy name, it is just the "key".

linera-storage/src/migration.rs
MathieuDutSik: Ok.

linera-storage/src/migration.rs
MathieuDutSik: Done.

linera-storage/src/db_storage.rs
MathieuDutSik: Ok for SchemaVersion.

linera-storage/src/db_storage.rs

MathieuDutSik:
Kill this PR, since the migration code must be in the `testnet_conway` branch. So another PR was created.
The points raised in this PR were implemented in the other PR.

PR #4842, opened 2025-10-22
Key the execution state cache with the block hash

bart-linera:
## Motivation

If a confirmed certificate was submitted to a validator with a correct execution state hash, but with an incorrect execution outcome, if the validator already had the execution state hash in its cache, it wouldn't re-execute the certificate and notice anything wrong.

## Proposal

Use the block hash instead of the execution state hash as the key in `execution_state_cache`. This way if the outcomes are wrong, the validator will re-execute the block and notice mismatching outcomes.

## Test Plan

CI

## Release Plan

## Links

- #4824 - a related PR on the `testnet_conway` branch that exposed the issue.

### DIFF
--- a/linera-core/src/chain_worker/state.rs
+++ b/linera-core/src/chain_worker/state.rs
@@ -1301,7 +1301,7 @@ where
         let (_, committee) = self.chain.current_committee()?;
         block.check_proposal_size(committee.policy().maximum_block_proposal_size)?;
 
-        let outcome = self
+        let executed_block = self
             .execute_block(&block, local_time, round, published_blobs)
             .await?;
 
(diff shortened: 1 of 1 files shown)


linera-core/src/chain_worker/state.rs
```diff
@@ -1560,8 +1560,10 @@ where
                     .execute_block(block, local_time, round, published_blobs, None),
             )
             .await?;
+        let block = Block::new(block.clone(), outcome.clone());
```
afck: We should probably return the full block from here, then, otherwise the caller will assemble it again? Then we won't need to clone the outcome here.

PR #4846, opened 2025-10-23
exporter: Add retry logging to indexer connection

ndr-ds:
## Motivation

When block exporters fail to connect to the indexer, they retry with exponential backoff but provide no visibility into the retry process. Operators cannot determine whether connection issues are transient (retrying successfully) or persistent (about to fail). This lack of observability makes it difficult to diagnose network connectivity issues and indexer availability problems during deployments and operations.

## Proposal

Add comprehensive logging to the indexer connection retry logic:

- **WARNING level** on each retry attempt, including:
  - Current retry count and max retries remaining
  - Calculated backoff delay in milliseconds
  - Error that triggered the retry
- **ERROR level** when all retries are exhausted, clearly indicating:
  - Final retry count
  - Max retries configured
  - Error that caused final failure
  - Explicit statement that exporter task will exit

This provides operators with actionable telemetry to distinguish transient network issues from persistent failures.

## Test Plan

1. Deploy network with intentionally unreachable indexer endpoint
2. Observe exporter logs show WARNING messages with increasing retry counts and backoff delays
3. Wait for max retries to be exhausted
4. Verify ERROR message appears with clear failure indication
5. Confirm logs provide enough information to diagnose the connection issue

## Release Plan

- These changes should be backported to the latest `testnet` branch, then
    - be released in a validator hotfix.

PR #4845, opened 2025-10-23
Make all heap profile logs trace

ndr-ds:
## Motivation

In a previous change, we lowered the log level for heap profile operations from INFO to TRACE to reduce log noise from frequent monitoring polls. However, we missed one log statement in the heap profile HTTP endpoint that still uses INFO level.

This inconsistency means that while most heap profile operations log at TRACE level, the HTTP endpoint still creates excessive log noise with `"Serving heap profile via /debug/pprof"` messages appearing in default logs every time the endpoint is polled.

## Proposal

Change the remaining log statement in `linera-metrics/src/memory_profiler.rs` from `info!` to `trace!` for the heap profile endpoint message, completing the consistency improvement started in the earlier change.

## Test Plan

1. Deploy network with memory profiling enabled
2. Verify heap profile endpoint still works (returns pprof data)
3. Observe that default logs no longer contain repetitive "Serving heap profile" messages
4. Enable TRACE logging and verify messages appear as expected
5. Confirm consistency with other heap profile logging

## Release Plan

- These changes should be backported to the latest `testnet` branch, then
    - be released in a validator hotfix.

PR #4847, opened 2025-10-23
exporter: Monitor task failures and propagate errors for restart

ndr-ds:
## Motivation

Block exporter tasks that fail to connect to the indexer (after exhausting retries from the previous PR) log ERROR messages but the main exporter process never detects these failures and continues running indefinitely. This occurs because `service.run()` blocks waiting for a shutdown signal, so when a spawned exporter task fails in the background, `handle.join()` never executes to propagate the error.

This creates silent failure scenarios where:
- Exporters appear healthy (pod Running, no restarts)
- Logs show connection errors hours ago
- No blocks are being exported
- Kubernetes never restarts the pod because the process hasn't exited

Operators discover these failures only by manually reviewing logs, not through automated monitoring or pod health checks.

## Proposal

Replace sequential execution with concurrent monitoring using `tokio::select!` to race the service against background tasks:

**Main entry point changes:**
- Wrap `handle.join()` in `tokio::task::spawn_blocking` to get an awaitable future
- Use `tokio::select!` to race `service.run()` against the block processor thread
- If block processor fails first, propagate the error immediately and cancel the shutdown notifier
- If service completes first (shutdown signal), wait for block processor to finish gracefully

**Block processor event loop changes:**
- Add new branch to main `select!` loop that monitors `join_set.join_next()`
- When any exporter task completes or fails, log appropriately and propagate errors
- Convert task panics to `ExporterError::GenericError` for consistent error handling

**Task tracking refactor:**
- Replace `HashMap<JoinHandle>` with `JoinSet<(DestinationId, anyhow::Result<()>)>`
- Add separate `HashMap<DestinationId, AbortHandle>` for selective task cancellation
- Store destination ID with each task result for detailed error reporting

This ensures errors propagate immediately through the normal error chain, triggering pod restarts via Kubernetes and enabling automated recovery.

## Test Plan

1. Deploy network with intentionally unreachable indexer
2. Wait for exporter connection retries to exhaust (logs will show ERROR from previous PR)
3. Verify exporter process exits with error code (not silent failure)
4. Confirm Kubernetes restarts the pod automatically
5. Verify new pod instance retries connection attempts

Compare against old behavior where pod would run indefinitely with logged errors but no restart.

## Release Plan

### DIFF
--- a/linera-service/src/exporter/main.rs
+++ b/linera-service/src/exporter/main.rs
@@ -116,10 +116,26 @@ impl Runnable for ExporterContext {
         );
 
         let service = ExporterService::new(sender);
-        service
-            .run(shutdown_notifier, self.config.service_config.port)
-            .await?;
-        handle.join().unwrap()
+
+        // Monitor the block processor thread in a blocking task
+        let mut handle_monitor = tokio::task::spawn_blocking(move || handle.join().unwrap());
+
+        // Run service and monitor block processor thread concurrently
+        // If the block processor fails, we need to detect it and propagate the error
+        tokio::select! {
+            service_result = service.run(shutdown_notifier.clone(), self.config.service_config.port) => {
+                // Service completed first (shutdown signal received)
+                service_result?;
+                // Now wait for the block processor to finish
+                handle_monitor.await.expect("Failed to join block processor thread")
+            }
+            processor_result = &mut handle_monitor => {
+                // Block processor thread completed first - likely an error
+                let result = processor_result.expect("Failed to join block processor thread");
+                shutdown_notifier.cancel();
+                result
+            }
+        }
     }
 }
 
--- a/linera-service/src/exporter/runloops/block_processor/mod.rs
+++ b/linera-service/src/exporter/runloops/block_processor/mod.rs
@@ -78,6 +78,30 @@ where
 
                 _ = interval.tick() => self.storage.save().await?,
 
+                Some(result) = self.exporters_tracker.join_set.join_next() => {
+                    // A task completed - check if it failed
+                    match result {
+                        Ok((id, Ok(()))) => {
+                            tracing::info!(destination_id=?id, "Exporter task completed successfully during execution");
+                        }
+                        Ok((id, Err(e))) => {
+                            tracing::error!(
+                                destination_id=?id,
+                                error=%e,
+                                "Exporter task failed"
+                            );
+                            return Err(ExporterError::GenericError(e.into()));
+                        }
+                        Err(join_err) => {
+                            tracing::error!(
+                                error=%join_err,
+                                "Exporter task panicked"
+                            );
+                            return Err(ExporterError::GenericError(Box::new(join_err)));
+                        }
+                    }
+                },
+
                 Some(next_block_notification) = self.new_block_queue.recv() => {
                     let walker = Walker::new(&mut self.storage);
                     match walker.walk(next_block_notification).await {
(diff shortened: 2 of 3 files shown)


ndr-ds:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4847?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#4850** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4850?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4849** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4849?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4848** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4848?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4847** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4847?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4847?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#4846** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4846?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4845** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4845?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

deuszx:
I think that an exporter that fails to export to one of the targets (due to misconfiguration or that destination's problems) shouldn't abort the whole process. It should log an error but continue operating.

ndr-ds:
Even if that target is the indexer? Because that's what was happening here. The indexer wasn't reachable, and it would just silently fail (before https://github.com/linera-io/linera-protocol/pull/4846, now it prints logs before failing), and just be a process that sits there and does nothing as it has no tasks running 😅
As it's only target task exited without actually accomplishing it's role, since we retried the max number of times...
So either we keep retrying here forever, which is not ideal, or we crash the process. Crashing the process is generally more common practice at least in the k8s world, AFAIK, because then k8s will restart the process and try again. And the crash loop also gives another layer of signal to whoever's looking at it

deuszx:
It's entirely possible that single exporter will have 10 different destinations to handle. One of them being unavailable shouldn't cause the whole process to crash.

I agree that not even logging any warnings was a mistake, but now that's fixed.

ndr-ds:
We need some signal that the process is unhealthy though. Maybe we can keep track of the destinations, and if they all fail, we crash the process? Because in this case the indexer was the only one, and it was failing

deuszx:
We can add a metric, or detect a log, but crashing the whole process is wrong here in my opinion. It was a single destination _in this particular test_ but it won't be the case in production deployments.

Let's take a different example: exporting to a committee. We will have N (20 let's say) validators that we export to – if one of them is unavailable, will you exit the whole process?

ndr-ds:
Yes, I get it that crashing for one destination failing is bad, I understand that. I guess what I'm proposing is only crashing if all the destinations fail, because then in that case we would have a process that literally does nothing, if we keep it running 😅
So it definitely makes sense to crash in that particular case, I think. I can work on the changes for that and update the PR.
To be clear, I'll update the PR making it so that if we fail a destination, we log an error, but won't necessarily crash, we only crash if this failing destination is the last one in the process

deuszx:
I don't think it will work – we are adding destinations dynamically as committee changes.

ndr-ds:
I see! I didn't realize we did that. Then indeed this doesn't make sense. Will close

PR #4849, opened 2025-10-23
Fix a few explorer frontend bugs

ndr-ds:
## Motivation

The explorer frontend hardcodes API URLs to `http://localhost:3002`, which only works in local development when both frontend and backend run on localhost. This breaks in production deployments where:

- Frontend is served through Kubernetes ingress
- Backend API is accessed via the same ingress with `/api` path prefix
- Browsers block mixed content (HTTPS frontend → HTTP localhost backend)
- CORS issues arise from cross-origin requests

The hardcoded URLs prevent the explorer from working in deployed environments.

## Proposal

Replace hardcoded absolute URLs with relative paths that work in all environments. Add Vite dev server proxy configuration to maintain local development workflow while using relative paths.

This approach works in all environments:
- **Local dev**: Vite proxy forwards `/api` → `http://localhost:3002`
- **Production**: Kubernetes ingress routes `/api` → backend service

## Test Plan

1. Test production deployment:
   - Deploy explorer with Kubernetes ingress configured
   - Verify frontend loads and connects to backend
   - Confirm no CORS or mixed content errors
   - Verify all API calls work through ingress

## Release Plan

- These changes should be backported to the latest `testnet` branch, then
    - be released in a validator hotfix.

PR #4848, opened 2025-10-23
Fix notification forwarding to avoid duplicate messages and handle lag

ndr-ds:
## Motivation

The notification forwarding system has two bugs:

1. **Duplicate messages to exporters**: When validators have multiple proxies, the servers will have a task per proxy forwarding notifications to all exporters. This means exporters receive duplicate notifications (N copies for N proxies), causing redundant processing and potential data inconsistencies.

2. **Receiver lag crashes forwarding**: The broadcast receiver for notifications can lag if the consumer is slower than the producer. When lag occurs, the code treats `RecvError::Lagged` as a fatal error and exits the forwarding loop, stopping all notification forwarding permanently even though the channel is still functional.

Both issues reduce system reliability and create operational problems that are difficult to diagnose.

## Proposal

**Fix duplicate forwarding:**
- Add `exporter_forwarded` flag to track whether exporters have been included
- Only the first proxy forwards to exporters, subsequent proxies forward to an empty exporter list
- Each proxy still forwards to its own validator endpoint

**Fix lag handling:**
- Replace `while let Ok()` pattern with explicit `match` on receiver result
- Handle `RecvError::Lagged(skipped_count)` by logging a warning and continuing
- Handle `RecvError::Closed` by logging and breaking (legitimate shutdown)
- Only actually lagged/skipped messages trigger the warning, not every message

This ensures exporters receive exactly one copy of each notification and forwarding continues even when temporary lag occurs.

## Test Plan

1. Deploy network with multiple proxies (2+) and exporters
2. Verify exporters receive exactly one copy of each notification (not N copies)
3. Generate high notification load to induce receiver lag
4. Verify warning logs appear for lagged messages
5. Confirm notification forwarding continues after lag events
6. Verify no duplicate processing in exporter/indexer

## Release Plan

- These changes should be backported to the latest `testnet` branch, then
    - be released in a validator hotfix.

PR #4850, opened 2025-10-23
Add support for exporter/indexer/explorer stack deployment

ndr-ds:
## Motivation

Deploying the exporter/indexer/explorer stack requires careful dependency ordering and runtime checks:

1. **Exporters need database to exist** before starting (ScyllaDB must be initialized)
2. **Exporters need indexer to be ready** before attempting connections
3. **Indexer image needs linera CLI** to perform database checks
4. **Configuration needs flexibility** to support both internal cluster DNS and external endpoints with TLS

Without these capabilities, deployments fail with race conditions where exporters start before dependencies are ready, causing crash loops and manual intervention.

## Proposal

Add comprehensive deployment support through init containers that wait for dependencies, include linera CLI in the indexer image for database checks, and make indexer endpoints configurable to support both internal and external deployments with automatic TLS handling.

## Test Plan

1. Deploy fresh network with exporter/indexer/explorer stack:
   - Verify ScyllaDB init container waits for database
   - Verify indexer readiness init container waits for indexer
   - Confirm exporters start only after both checks pass
   
2. Test with external indexer endpoint:
   - Configure `blockExporter.indexerEndpoint` with external URL
   - Verify TLS configuration is applied
   - Confirm connectivity works

3. Test failure scenarios:
   - Deploy without indexer, verify exporter init container retries indefinitely
   - Kill indexer, verify exporters restart and wait for it to return

## Release Plan

### DIFF
--- a/docker/Dockerfile.indexer
+++ b/docker/Dockerfile.indexer
@@ -73,9 +73,9 @@ ENV GIT_COMMIT=${git_commit}
 ENV RUSTFLAGS=${rustflags}
 
 
-# Build block exporter with scylladb and metrics features
+# Build linera CLI and block exporter with scylladb and metrics features
 RUN cargo build ${build_flag:+"$build_flag"} \
-    -p linera-service \
+    --bin linera \
     --bin linera-exporter \
     --features $build_features
 
@@ -86,6 +86,7 @@ RUN cargo build ${build_flag:+"$build_flag"} \
 
 # Move binaries to avoid directory conflicts and clean up to save space
 RUN mv \
+    target/"$build_folder"/linera \
     target/"$build_folder"/linera-exporter \
     target/"$build_folder"/linera-indexer-grpc \
     ./
@@ -94,6 +95,7 @@ RUN mv \
 FROM scratch AS builder_copy
 ARG binaries
 COPY \
+    "$binaries"/linera \
     "$binaries"/linera-exporter \
     "$binaries"/linera-indexer-grpc \
     ./
@@ -121,8 +123,9 @@ RUN update-ca-certificates
 
 ARG target
 
-# Copy only indexer and exporter binaries
+# Copy linera CLI, indexer and exporter binaries
 COPY --from=binaries \
+    linera \
     linera-indexer-grpc  \
     linera-exporter \
     ./
--- a/kubernetes/linera-validator/templates/block-exporter.yaml
+++ b/kubernetes/linera-validator/templates/block-exporter.yaml
@@ -49,6 +49,44 @@ spec:
         effect: NoSchedule
       {{- end }}
       initContainers:
+        - name: linera-exporter-initializer
+          image: {{ .Values.indexer.image }}
+          imagePullPolicy: {{ .Values.indexer.imagePullPolicy }}
+          securityContext:
+            runAsNonRoot: true
+            runAsUser: 65534
+            allowPrivilegeEscalation: false
+            capabilities:
+              drop:
+                - ALL
+          resources:
+            requests:
+              memory: "64Mi"
+              cpu: "50m"
+            limits:
+              memory: "128Mi"
+              cpu: "100m"
+          command:
+            - sh
+            - -c
+            - |
+              set -euo pipefail
+              while true; do
+                output=$(./linera storage check-existence --storage "{{ .Values.storage }}" 2>&1)
+                status=$?
+                if [ "$status" -eq 0 ]; then
+                  echo "Database already exists, no need to initialize."
+                  exit 0
+                else
+                  if [ "$status" -eq 1 ]; then
+                    echo "Database does not exist, retrying in {{ .Values.blockExporter.initRetryIntervalSeconds | default 5 }} seconds..."
+                  else
+                    echo "An unexpected error occurred (status: $status): $output"
+                    echo "Retrying in {{ .Values.blockExporter.initRetryIntervalSeconds | default 5 }} seconds..."
+                  fi
+                  sleep {{ .Values.blockExporter.initRetryIntervalSeconds | default 5 }}
+                fi
+              done
         - name: config-selector
           image: busybox
           command:
@@ -62,6 +100,65 @@ spec:
               mountPath: /configmap
             - name: config
               mountPath: /config
+        - name: indexer-readiness-check
+          image: curlimages/curl:latest
+          securityContext:
+            runAsNonRoot: true
+            runAsUser: 65534
+            allowPrivilegeEscalation: false
+            readOnlyRootFilesystem: true
+            capabilities:
+              drop:
+                - ALL
+          resources:
+            requests:
+              memory: "32Mi"
+              cpu: "25m"
+            limits:
+              memory: "64Mi"
+              cpu: "50m"
+          env:
+            - name: INDEXER_ENDPOINT
+              value: {{ if .Values.blockExporter.indexerEndpoint }}"{{ .Values.blockExporter.indexerEndpoint }}"{{ else }}"linera-indexer-{{ .Values.networkName }}.linera-indexer.svc.cluster.local"{{ end }}
+            - name: INDEXER_PORT
+              value: {{ if .Values.blockExporter.indexerEndpoint }}"443"{{ else }}"{{ .Values.blockExporter.indexerPort }}"{{ end }}
+            - name: INDEXER_TLS
+              value: {{ if .Values.blockExporter.indexerEndpoint }}"true"{{ else }}"false"{{ end }}
+          command:
+            - sh
+            - -c
+            - |
+              set -euo pipefail
+              # Validate extracted values
+              if [ -z "$INDEXER_ENDPOINT" ] || [ -z "$INDEXER_PORT" ]; then
+                echo "Error: INDEXER_ENDPOINT or INDEXER_PORT is empty"
+                exit 1
+              fi
+
+              if [ "$INDEXER_TLS" = "true" ]; then
+                INDEXER_URL="https://${INDEXER_ENDPOINT}:${INDEXER_PORT}"
+              else
+                INDEXER_URL="http://${INDEXER_ENDPOINT}:${INDEXER_PORT}"
+              fi
+
+              echo "Checking indexer readiness at ${INDEXER_URL}..."
+
+              while true; do
+                # Try to connect to the indexer (just check if it's listening)
+                # Using --insecure for init container readiness check only
+                # This is acceptable because:
+                # 1. Only checking if service is reachable, not exchanging sensitive data
+                # 2. External endpoints may use self-signed certs in dev/staging
+                # 3. Actual exporter traffic uses proper TLS validation
+                if curl --max-time 5 --connect-timeout 5 --silent --fail --insecure "${INDEXER_URL}" > /dev/null 2>&1 || \
+                   curl --max-time 5 --connect-timeout 5 --silent --head --insecure "${INDEXER_URL}" > /dev/null 2>&1; then
+                  echo "Indexer is reachable, exporter can start."
+                  exit 0
+                else
+                  echo "Indexer not ready yet, retrying in {{ .Values.blockExporter.initRetryIntervalSeconds | default 5 }} seconds..."
+                  sleep {{ .Values.blockExporter.initRetryIntervalSeconds | default 5 }}
+                fi
+              done
       containers:
         - name: linera-block-exporter
           image: {{ .Values.indexer.image }}
--- a/kubernetes/linera-validator/templates/proxy.yaml
+++ b/kubernetes/linera-validator/templates/proxy.yaml
@@ -16,7 +16,7 @@ spec:
       name: metrics
   selector:
     app: proxy
-  clusterIp: None
+  clusterIP: None
 
 ---
 apiVersion: v1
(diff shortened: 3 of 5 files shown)


ndr-ds:
* **#4855** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4855?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4850** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4850?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4850?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#4849** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4849?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4848** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4848?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

docker/Dockerfile.indexer
```diff
@@ -73,9 +73,9 @@ ENV GIT_COMMIT=${git_commit}
 ENV RUSTFLAGS=${rustflags}
 
 
-# Build block exporter with scylladb and metrics features
+# Build linera CLI and block exporter with scylladb and metrics features
 RUN cargo build ${build_flag:+"$build_flag"} \
-    -p linera-service \
+    --bin linera \
```
deuszx, line 78: This file was supposed to be used for local testing and not production deployment so the image size was intentionally kept as small as possible.

ndr-ds: Right, but we don't always need the `linera-exporter` binary, so it doesn't make sense to include it in the main `Dockerfile`, and now because we check for the DB's existence as an init container to the exporter (since it also depends on it), we need the `linera` binary here

deuszx: > because we check for the DB's existence as an init container to the exporter

We shouldn't be doing that. Exporter and indexer (and its DB) are separate concepts – they should not depend on one another at the dockerfile level.

ndr-ds: The exporter uses `ScyllaDB`, and it runs with the validator, potentially. So it needs to make sure that `ScyllaDB` is properly initialized and running before it starts, just like the shards. I'm not talking about the indexer DB here

deuszx: OK, got it.

kubernetes/linera-validator/templates/block-exporter.yaml
```diff
@@ -49,6 +49,28 @@ spec:
...
+            - sh
+            - -c
+            - |
+              while true; do
+                ./linera storage check-existence --storage "{{ .Values.storage }}"
+                status=$?
+                if [ "$status" -eq 0 ]; then
+                  echo "Database already exists, no need to initialize."
+                  exit 0
+                else
```
eldios, line 80: **Issue:** The error handling doesn't capture the actual error details, making debugging difficult when unexpected errors occur.

**Recommendation:** Capture and log the actual error output:
```bash
while true; do
  output=$(./linera storage check-existence --storage "{{ .Values.storage }}" 2>&1)
  status=$?
  if [ "$status" -eq 0 ]; then
    echo "Database already exists, no need to initialize."
    exit 0
  else
    if [ "$status" -eq 1 ]; then
      echo "Database does not exist, retrying in 5 seconds..."
    else
      echo "An unexpected error occurred (status: $status): $output"
      echo "Retrying in 5 seconds..."
    fi
    sleep 5
  fi
done
```

This will help diagnose issues faster when the database check fails unexpectedly.

kubernetes/linera-validator/templates/block-exporter.yaml
```diff
@@ -62,6 +84,40 @@ spec:
...
             - name: config
               mountPath: /config
+        - name: indexer-readiness-check
+          image: curlimages/curl:latest
+          command:
+            - sh
+            - -c
+            - |
+              # Extract indexer endpoint from config
+              INDEXER_ENDPOINT=$(grep "^endpoint" /config/exporter-config.toml | sed 's/.*"\(.*\)".*/\1/')
```
eldios: **Critical Issue:** This TOML parsing using grep/sed is fragile and error-prone:

1. **No validation** that extracted values are non-empty
2. **Format assumptions** - grep `"^endpoint"` assumes specific TOML formatting
3. **Complex regex** - `'s/.*"\(.*\)".*/\1/'` can fail with edge cases
4. **Case-sensitive TLS check** - comparing to literal `"Tls"` string

**Recommended Solutions:**

**Option 1 - Use a proper TOML parser:**
```yaml
- name: indexer-readiness-check
  image: alpine:latest
  command:
    - sh
    - -c
    - |
      apk add --no-cache curl yq
      INDEXER_ENDPOINT=$(yq eval '.destination_config.destinations[] | select(.kind == "Indexer") | .endpoint' /config/exporter-config.toml)
      INDEXER_PORT=$(yq eval '.destination_config.destinations[] | select(.kind == "Indexer") | .port' /config/exporter-config.toml)
      INDEXER_TLS=$(yq eval '.destination_config.destinations[] | select(.kind == "Indexer") | .tls' /config/exporter-config.toml)
      
      # Validate extracted values
      if [ -z "$INDEXER_ENDPOINT" ] || [ -z "$INDEXER_PORT" ]; then
        echo "Error: Failed to extract indexer configuration"
        exit 1
      fi
      
      # Rest of script...
```

**Option 2 - Simplify by using Helm values directly:**
Since the config is templated, pass the values as env vars instead of parsing TOML:
```yaml
env:
  - name: INDEXER_ENDPOINT
    value: {{ if .Values.blockExporter.indexerEndpoint }}"{{ .Values.blockExporter.indexerEndpoint }}"{{ else }}"linera-indexer-{{ .Values.networkName }}.linera-indexer.svc.cluster.local"{{ end }}
  - name: INDEXER_PORT
    value: {{ if .Values.blockExporter.indexerEndpoint }}"443"{{ else }}"{{ .Values.blockExporter.indexerPort }}"{{ end }}
  - name: INDEXER_TLS
    value: {{ if .Values.blockExporter.indexerEndpoint }}"true"{{ else }}"false"{{ end }}
```

Option 2 is cleaner and avoids TOML parsing entirely.

kubernetes/linera-validator/templates/block-exporter.yaml
```diff
@@ -49,6 +49,28 @@ spec:
...
+                if [ "$status" -eq 0 ]; then
+                  echo "Database already exists, no need to initialize."
+                  exit 0
+                else
+                  if [ "$status" -eq 1 ]; then
+                    echo "Database does not exist, retrying in 5 seconds..."
+                  else
+                    echo "An unexpected error occurred (status: $status), retrying in 5 seconds..."
+                  fi
+                  sleep 5
```
eldios: **Suggestion:** The 5-second retry interval is hardcoded in both init containers. Consider making this configurable:

In `values.yaml`:
```yaml
blockExporter:
  initRetryIntervalSeconds: 5
```

Then use:
```bash
sleep {{ .Values.blockExporter.initRetryIntervalSeconds | default 5 }}
```

This allows operators to adjust retry behavior without modifying templates.

kubernetes/linera-validator/templates/proxy.yaml
```diff
@@ -16,7 +16,7 @@ spec:
       name: metrics
   selector:
     app: proxy
-  clusterIp: None
+  clusterIP: None
```
eldios, line 19: Good catch fixing the typo! However, this change is unrelated to the PR's main purpose (exporter/indexer deployment support).

**Recommendation:** Consider splitting this into a separate minor PR or at minimum, mention it in the PR description as an incidental fix discovered during development.

eldios (changes requested):
**Issues to Address:**

### Critical
**Fragile TOML parsing** (line 94): The shell-based parsing using grep/sed is brittle and lacks validation. Recommend either:
1. Using a proper TOML parser (yq/jq)
2. Passing values directly as env vars from Helm (cleaner solution)

### Important
**Error handling gap** (line 65): The database check doesn't capture actual error output, making debugging harder. Should log the error details when unexpected statuses occur.

### Nice to Have
- Make retry intervals configurable via values.yaml
- Split the proxy.yaml typo fix into a separate PR (scope creep)

The critical TOML parsing issue should be addressed before merge to ensure reliability in production deployments. The rest are recommended improvements that would enhance maintainability.

Once the parsing is made robust, this will be a solid addition to the deployment tooling.

kubernetes/linera-validator/templates/proxy.yaml
eldios: actually this is needed otherwise it fails, I'm having this same problem in my tests

kubernetes/linera-validator/templates/block-exporter.yaml
```diff
@@ -49,6 +49,29 @@ spec:
         effect: NoSchedule
       {{- end }}
       initContainers:
+        - name: linera-exporter-initializer
```
eldios, line 52: Missing security context and resource limits for init container

kubernetes/linera-validator/templates/block-exporter.yaml
```diff
@@ -49,6 +49,29 @@ spec:
...
+          image: {{ .Values.indexer.image }}
+          imagePullPolicy: {{ .Values.indexer.imagePullPolicy }}
+          command:
+            - sh
+            - -c
+            - |
```
eldios, line 72: Shell script should start with `set -euo pipefail` for safety - exits on undefined vars, pipe failures, and errors.

kubernetes/linera-validator/templates/block-exporter.yaml
```diff
@@ -62,6 +85,44 @@ spec:
               mountPath: /configmap
             - name: config
               mountPath: /config
+        - name: indexer-readiness-check
```
eldios, line 103: Missing security context and resource limits - same as first init container.

kubernetes/linera-validator/templates/block-exporter.yaml
```diff
@@ -62,6 +85,44 @@ spec:
...
+                INDEXER_URL="https://${INDEXER_ENDPOINT}:${INDEXER_PORT}"
+              else
+                INDEXER_URL="http://${INDEXER_ENDPOINT}:${INDEXER_PORT}"
+              fi
+
+              echo "Checking indexer readiness at ${INDEXER_URL}..."
+
+              while true; do
+                # Try to connect to the indexer (just check if it's listening)
+                if curl --max-time 5 --connect-timeout 5 --silent --fail --insecure "${INDEXER_URL}" > /dev/null 2>&1 || \
```
eldios, line 153: `--insecure` flag bypasses TLS certificate validation - security risk when connecting to external endpoints. Consider using proper CA bundle or at minimum add a comment explaining why this is acceptable.

eldios: this can also be done in future as a hardening step, resolving for now

kubernetes/linera-validator/templates/block-exporter.yaml
eldios: we should always set resources and limits... we saw what can happen today with the faucet (ie: the node dies when all memory is used).

eldios:
**Implementation examples for review comments:**

**1. Security context + resource limits for init containers:**

```yaml
- name: linera-exporter-initializer
  image: {{ .Values.indexer.image }}
  imagePullPolicy: {{ .Values.indexer.imagePullPolicy }}
  securityContext:
    runAsNonRoot: true
    runAsUser: 65534
    allowPrivilegeEscalation: false
    readOnlyRootFilesystem: true
    capabilities:
      drop:
        - ALL
  resources:
    requests:
      memory: "64Mi"
      cpu: "50m"
    limits:
      memory: "128Mi"
      cpu: "100m"
  command:
    - sh
    - -c
    - |
      set -euo pipefail
      while true; do
        # ... rest of script
```

**2. Safe shell script with set -euo pipefail:**

```yaml
command:
  - sh
  - -c
  - |
    set -euo pipefail
    while true; do
      output=$(./linera storage check-existence --storage "{{ .Values.storage }}" 2>&1)
      status=$?
      # ... rest of logic
```

**3. Same security context for indexer-readiness-check:**

```yaml
- name: indexer-readiness-check
  image: curlimages/curl:latest
  securityContext:
    runAsNonRoot: true
    runAsUser: 65534
    allowPrivilegeEscalation: false
    readOnlyRootFilesystem: true
    capabilities:
      drop:
        - ALL
  resources:
    requests:
      memory: "32Mi"
      cpu: "25m"
    limits:
      memory: "64Mi"
      cpu: "50m"
  env:
    # ... existing env vars
```

**4. Handling TLS certificate validation - two options:**

Option A - Add comment explaining why insecure is acceptable:
```yaml
# Using --insecure for init container readiness check only
# This is acceptable because:
# 1. Only checking if service is reachable, not exchanging sensitive data
# 2. External endpoints may use self-signed certs in dev/staging
# 3. Actual exporter traffic uses proper TLS validation
if curl --max-time 5 --connect-timeout 5 --silent --fail --insecure "${INDEXER_URL}" > /dev/null 2>&1; then
```

Option B - Use proper CA bundle (production):
```yaml
if curl --max-time 5 --connect-timeout 5 --silent --fail --cacert /etc/ssl/certs/ca-certificates.crt "${INDEXER_URL}" > /dev/null 2>&1; then
```

Issue #4839, opened 2025-10-22
Deadlocks for `test_end_to_end_listen_for_new_rounds`


MathieuDutSik:
The test `test_end_to_end_listen_for_new_rounds` of the CI is doing 8 transfers from two concurrent threads.

However, the number of transfers can be changed to a variable number `n`. This reveals some deadlocks:
* For $n \leq 7$ we have deadlock
* For $8\leq n \leq 29$ the test passes.
* For $n\geq 30$ we have deadlocks.

PR #4851, opened 2025-10-23
Fix `test_end_to_end_listen_for_new_rounds`.

afck:
## Motivation

`test_end_to_end_listen_for_new_rounds` was meant to test two active single-round leaders transferring tokens until they run out of money. It is more complicated than it needs to be and has deadlock issues because it uses a synchronous barrier.

## Proposal

Don't use a barrier at all. Instead, return the client wrappers from the spawned tasks so they don't get dropped early.

Set the chain balance to 8.9 so exactly 8 transfers can be made.

Add more comments and assertions.

## Test Plan

Works for me locally; let's try it in CI!

## Release Plan

- These changes could be backported to `testnet_conway`. (Low priority)

## Links

- Closes https://github.com/linera-io/linera-protocol/issues/4839. (The test is only meant to pass with exactly 8 now.

### DIFF
--- a/linera-explorer-new/src/hooks/useDatabase.ts
+++ b/linera-explorer-new/src/hooks/useDatabase.ts
@@ -9,10 +9,10 @@ export const useAPI = () => {
   const [error, setError] = useState<string | null>(null);
 
   useEffect(() => {
-    // Test API connection
+    // Test API connection using relative path
     const testConnection = async () => {
       try {
-        const response = await fetch('http://localhost:3002/api/health');
+        const response = await fetch('/api/health');
         if (response.ok) {
           setIsConnected(true);
         } else {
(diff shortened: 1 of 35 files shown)


afck:
I think that's just because the channel only has capacity 8, so `send` blocks at the 9th? I added a comment.

MathieuDutSik (approved):
At least now I understand why it deadlocks.

If there are more cash in the chain than 8, then the transfers are done but will deadlock.

It is not perfect, but at least now the deadlock scenario is much clearer.

afck:
OK, I have an idea: Let's use `try_send` instead of `send`, and panic if the buffer is full.

afck:
Done. It now crashes with too low or too high balance, and doesn't block.

ma2bd:
@afck Is this ready to land?

afck:
~@ma2bd: It got removed from the merge queue because the `remote-net` tests timed out, so I'll need to look into that.~

(Done: Needed to drop clients with `spawn_blocking`.)

PR #4852, opened 2025-10-23
Remove UNIQUE constraints – those should be tracked by the node anyway

deuszx:
## Motivation

We saw UNIQUE constraints violations for the `operations` table when testing w/ the exporter.

## Proposal

Remove all constraints – those should be maintained by the node anyway and add additional work for the SQlite (that has to check them).

## Test Plan

manual.

## Release Plan

- Nothing to do / These changes follow the usual release cycle.## Links


Twey:
Does it add additional work or does it enable optimizations?

PR #4856, opened 2025-10-24
Introduce RequestsScheduler to track (and deduplicate) all requests and validators' scores

deuszx:
Backport of #4752 

## Motivation

  The Linera client needs to interact with multiple validator nodes efficiently. Previously, the
  client would make individual requests to validators without:

  1. Performance tracking: No mechanism to prefer faster, more reliable validators
  2. Request deduplication: Concurrent requests for the same data would all hit the network, wasting
  bandwidth and validator resources
  3. Response caching: Repeated requests for the same data would always go to validators
  4. Load balancing: No rate limiting per validator, risking overload
  5. Resilience: No fallback mechanism when a validator is slow or unresponsive

  This led to:
  - Unnecessary network traffic and validator load
  - Poor user experience with redundant waiting
  - No optimization based on validator performance
  - Risk of overwhelming validators with too many concurrent requests
  - No recovery mechanism when validators are slow

##  Proposal

  This PR introduces `RequestsScheduler`, a sophisticated request orchestration layer that provides
  intelligent peer selection, request deduplication, caching, and performance-based routing.

  Key Features

  1. Performance Tracking with Exponential Moving Averages (EMA)
  - Tracks latency, success rate, and current load for each validator
  - Uses configurable weights to compute a composite performance score
  - Intelligently selects the best available validator for each request
  - Weighted random selection from top performers to avoid hotspots

  2. Request Deduplication
  - Exact matching: Multiple concurrent requests for identical data are deduplicated
  - Subsumption-based matching: Smaller requests are satisfied by larger in-flight requests that
  contain the needed data (e.g., a request for blocks 10-12 can be satisfied by an in-flight request
  for blocks 10-20)
  - Broadcast mechanism ensures all waiting requesters receive the result when the request completes
  - Timeout handling: Stale in-flight requests (>200ms) are not deduplicated against, allowing fresh
  attempts

  3. Response Caching
  - Successfully completed requests are cached with configurable TTL (default: 2 seconds)
  - LRU eviction when cache reaches maximum size (default: 1000 entries)
  - Works with both exact and subsumption matching
  - Only successful results are cached

  4. Slot-Based Rate Limiting
  - Each validator has a maximum concurrent request limit (default: 100)
  - Async await mechanism: requests wait for available slots without polling
  - Prevents overloading individual validators
  - Automatic slot release on request completion

  5. Alternative Peer Handling
  - When multiple callers request the same data, they register as "alternative peers"
  - If the original request times out (>200ms), any alternative peer can complete the request
  - The result is broadcast to all waiting requesters
  - Provides resilience against slow validators

  6. Modular Architecture

  Created a new ` requests_scheduler` module with clear separation of concerns:
```
 requests_scheduler/
  ├── mod.rs              - Module exports and constants
  ├── scheduler.rs          - RequestsScheduler orchestration logic
  ├── in_flight_tracker.rs - In-flight request tracking and deduplication
  ├── node_info.rs        - Per-validator performance tracking
  ├── request.rs          - Request types and result extraction
  └── scoring.rs          - Configurable scoring weights
```
  API

  High-level APIs:
```rust
  // Execute with best available validator
  scheduler.with_best(request_key, |peer| async {
      peer.download_certificates(chain_id, start, limit).await
  }).await

  // Execute with specific validator
  scheduler.with_peer(request_key, peer, |peer| async {
      peer.download_blob(blob_id).await
  }).await

  Configuration:
  let manager = RequestsScheduler::with_config(
      validator_nodes,
      max_requests_per_node: 100,
      weights: ScoringWeights { latency: 0.4, success: 0.4, load: 0.2 },
      alpha: 0.1,              // EMA smoothing factor
      max_expected_latency_ms: 5000.0,
      cache_ttl: Duration::from_secs(2),
      max_cache_size: 1000,
  );
```

  Benefits

  - Reduced network load: Deduplication and caching eliminate redundant requests
  - Better performance: Intelligent peer selection routes to fastest validators
  - Improved reliability: Alternative peer mechanism provides resilience
  - Protection for validators: Rate limiting prevents overload
  - Efficient resource usage: EMA-based scoring optimizes validator selection
  - Clean architecture: Modular design makes code maintainable and testable

  Metrics

  In production usage, this should significantly reduce:
  - Network traffic between clients and validators
  - Validator CPU/memory usage from redundant requests
  - Client request latency through caching and smart routing
  - Failed requests through performance tracking and rate limiting

The following metrics have been added to Prometheus (with compiled with `--features metrics`):
- `requests_scheduler_response_time_ms` - Response time for requests to validators in milliseconds
- `requests_scheduler_request_total` - Total number of requests made to each validator
- `requests_scheduler_request_success` - Number of successful requests to each validator (`(requests_scheduler_request_total - requests_scheduler_request_success) / requests_scheduler_request_total` is an error rate)
- `requests_scheduler_request_deduplication_total` - Number of requests that were deduplicated by joining an in-flight request
- `requests_scheduler_request_cache_hit_total` - Number of requests that were served from cache

## Test Plan

Existing CI makes sure we maintain backwards compatibility. Some tests have been added to the new modules.

## Release Plan

## Links

### DIFF
--- a/linera-core/src/client/requests_scheduler/scheduler.rs
+++ b/linera-core/src/client/requests_scheduler/scheduler.rs
@@ -165,7 +165,6 @@ impl<Env: Environment> RequestsScheduler<Env> {
     /// - `cache_ttl`: Time-to-live for cached responses
     /// - `max_cache_size`: Maximum number of entries in the cache
     /// - `max_request_ttl`: Maximum latency for an in-flight request before we stop deduplicating it
-    #[expect(clippy::too_many_arguments)]
     pub fn with_config(
         nodes: impl IntoIterator<Item = RemoteNode<Env::ValidatorNode>>,
         weights: ScoringWeights,
--- a/linera-core/src/remote_node.rs
+++ b/linera-core/src/remote_node.rs
@@ -208,7 +208,8 @@ impl<N: ValidatorNode> RemoteNode<N> {
             Err(NodeError::BlobsNotFound(_error)) => {
                 tracing::debug!(
                     ?blob_id,
-                    validator=?self.public_key,
+                    validator = ?self.public_key,
+                    address = ?self.node.address(),
                     "validator is missing the blob",
                 );
                 Ok(None)


linera-core/src/client/requests_scheduler/cache.rs
```diff
@@ -0,0 +1,387 @@
...
+                    tracing::trace!(
+                        key = ?key,
+                        "cache hit (subsumption) - extracted result from larger cached request"
+                    );
+                    #[cfg(with_metrics)]
+                    metrics::REQUEST_CACHE_HIT.inc();
+                    return T::try_from(extracted).ok();
+                }
+            }
+        }
```
bart-linera, line 93: We might want to figure out some smarter way to store cache entries, as the way it is now, we need to iterate through the whole cache in order to declare a cache miss. (Probably not a big issue, but could become one if we allow large numbers of entries.)

linera-core/src/client/requests_scheduler/scheduler.rs
```diff
@@ -0,0 +1,1291 @@
...
+    {
+        let start_time = Instant::now();
+        let public_key = peer.public_key;
+
+        // Acquire request slot
+        let nodes = self.nodes.read().await;
+        let node = nodes.get(&public_key).expect("Node must exist");
+        let semaphore = node.in_flight_semaphore.clone();
+        let permit = semaphore.acquire().await.unwrap();
+        drop(nodes);
```
bart-linera: Probably not very important, but wouldn't it be better to `drop(nodes)` before acquiring the semaphore?

deuszx: I removed the Semaphore entirely.

deuszx:
@afck @ma2bd I removed Semaphore from the PR as requested.

linera-rpc/src/grpc/client.rs
```diff
@@ -71,11 +71,11 @@ impl GrpcClient {
     fn is_retryable(status: &Status) -> bool {
         match status.code() {
             Code::DeadlineExceeded | Code::Aborted | Code::Unavailable | Code::Unknown => {
-                info!("gRPC request interrupted: {status:?}; retrying");
+                trace!("gRPC request interrupted: {status:?}; retrying");
                 true
             }
             Code::Ok | Code::Cancelled | Code::ResourceExhausted => {
-                info!("Unexpected gRPC status: {status:?}; retrying");
+                trace!("Unexpected gRPC status: {status:?}; retrying");
```
afck, line 78: Might still be good to have a DEBUG-log when we give up after a number of retries, and maybe below, where we don't retry?

linera-rpc/src/client.rs
```diff
@@ -44,6 +44,14 @@ impl From<SimpleClient> for Client {
 impl ValidatorNode for Client {
     type NotificationStream = NotificationStream;
 
+    fn address(&self) -> String {
+        match self {
+            Client::Grpc(grpc_client) => grpc_client.address().to_string(),
```
afck, line 49:
```suggestion
            Client::Grpc(grpc_client) => grpc_client.address(),
```

linera-core/src/remote_node.rs
```diff
@@ -219,18 +200,20 @@ impl<N: ValidatorNode> RemoteNode<N> {
...
+                    Ok(Some(blob))
                 }
             }
-            Err(error) => {
+            Err(NodeError::BlobsNotFound(_error)) => {
                 tracing::debug!(
-                    "Failed to fetch blob {blob_id} from validator {}: {error}",
-                    self.public_key
+                    ?blob_id,
+                    validator=?self.public_key,
```
afck:
```suggestion
                    validator = ?self.public_key,
```

afck: But we could now print `self.node.address` instead (or in addition), which is probably more useful. (Not necessarily in this PR.)

linera-client/src/client_options.rs
```diff
@@ -186,6 +186,58 @@ pub struct ClientContextOptions {
     /// Maximum number of tasks that can are joined concurrently in the client.
     #[arg(long, default_value = "100")]
     pub max_joined_tasks: usize,
+
+    /// Maximum concurrent requests per validator node.
```
afck: It's now more like a scaling factor than a maximum.

deuszx: I'm gonna remove it entirely.

linera-rpc/src/grpc/client.rs
deuszx: If it's no longer retry-able we return an error:
```
Err(NodeError::GrpcError {
                        error: format!("remote request [{handler}] failed with status: {s:?}"),
                    });
```

deuszx: As for the requests types where we don't retry I don't think it's going to add anything to the observability. We don't retry not because something went wrong but rather b/c we don't use those gRPC errors (and so we don't expect them).

linera-rpc/src/client.rs
deuszx: Surprised clippy didn't complain.

deuszx: Yeah, actually it was correct - different types. `grpc_client.address()` returns `&str` and we want `String`.

deuszx:
> Approving, assuming nothing fundamental has changed compared to the Conway version.

Yes, should be pretty much the same now.

PR #4858, opened 2025-10-25
Address a bug with the `root_keys` in `linera-storage-service`.

MathieuDutSik:
## Motivation

There is a bug in the handling of the root_keys in the `linera-storage-service`. The `root_key` is directly appended to the `start_key`. The result is an object which is not prefix-free.

## Proposal

Replace
```rust
start_key.extend(root_key)
```

by
```rust
start_key.extend(bcs::to_bytes(&root_key)?)
```

by doing so, we no longer have ambiguity in the construction of `start_key`.

## Test Plan

The CI.
Added a test that fails on main on that precise problem.

## Release Plan

Since, the `linera-storage-service` is not used in TestNet Conway, no backport is needed.

## Links

None.

### DIFF
--- a/linera-storage-service/tests/store_test.rs
+++ b/linera-storage-service/tests/store_test.rs
@@ -4,12 +4,13 @@
 #![cfg(feature = "storage-service")]
 
 use anyhow::Result;
-use linera_storage_service::{
-    client::{StorageServiceDatabase, StorageServiceDatabaseInternal},
-};
+use linera_storage_service::client::{StorageServiceDatabase, StorageServiceDatabaseInternal};
 use linera_views::{
     batch::Batch,
-    store::{TestKeyValueDatabase as _, KeyValueDatabase as _, ReadableKeyValueStore as _, WritableKeyValueStore as _},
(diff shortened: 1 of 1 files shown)


ma2bd:
> Since, the linera-storage-service is not used in TestNet Conway, no backport is needed.

`linera-storage-service` is used locally by developers, so we do want to backport

linera-storage-service/src/client.rs
```diff
@@ -304,14 +304,14 @@ impl WritableKeyValueStore for StorageServiceStoreInternal {
             chunk_size += self.start_key.len();
         }
 
-        let root_key_len = self.start_key.len() - self.prefix_len;
+        let bcs_root_key_len = self.start_key.len() - self.prefix_len;
```
ma2bd, line 307: Did you mean to adjust this formula?

MathieuDutSik: No.
The formula is correct. I only wanted to write that the length in question is not that of `root_key` but of its serialization.

PR #4859, opened 2025-10-26
Correct a bug in `root_key` handling in `RocksDb`.

MathieuDutSik:
## Motivation

The bug addressed in PR #4858 is also present in `RocksDb`.

## Proposal

The same solution is applied. The test code inserted in `linera-storage-service` is added to `root_key_admin_test` so that all storage are now tested on this possible issue.

## Test Plan

The CI.

## Release Plan

This PR should NOT be merged into TestNet Conway.

The reason is that while the bug being addressed is real, correcting it means that we lose access to the
RocksDB storage, through the standard API.

Backport in TestNet Conway after main.

## Links

None.

### DIFF
--- a/linera-views/src/test_utils/mod.rs
+++ b/linera-views/src/test_utils/mod.rs
@@ -835,7 +835,8 @@ where
     }
     assert_eq!(keys, read_keys);
 
-    // Checking prefix freeness of the
+    // Checking prefix freeness of the (root_key, key). This is a
+    // common problem that needs to be tested.
     let database = D::connect_test_namespace().await.expect("database");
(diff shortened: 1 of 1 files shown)


linera-views/src/test_utils/mod.rs
```diff
@@ -833,6 +833,20 @@ where
         }
     }
     assert_eq!(keys, read_keys);
+
+    // Checking prefix freeness of the
```
ma2bd: nit: finish sentence

MathieuDutSik: Ok.

PR #4860, opened 2025-10-26
Change the API of `list_root_keys`.

MathieuDutSik:
## Motivation

The API of `list_root_keys` is taking the config and namespace. This is a little bit illogical, since the function should be a function of the database itself.

## Proposal

Restructure the API accordingly.

## Test Plan

The CI should catch the problems.

## Release Plan

This is an internal change, which is useful for debugging and consistency.
It can be backported to TestNet Conway but that does not appear to be essential.

## Links

None.

### DIFF
--- a/linera-storage-service/src/client.rs
+++ b/linera-storage-service/src/client.rs
@@ -519,7 +519,9 @@ impl KeyValueDatabase for StorageServiceDatabaseInternal {
     }
 
     async fn list_root_keys(&self) -> Result<Vec<Vec<u8>>, StorageServiceStoreError> {
-        let query = RequestListRootKeys { namespace: self.namespace.clone() };
+        let query = RequestListRootKeys {
+            namespace: self.namespace.clone(),
+        };
         let request = tonic::Request::new(query);
         let mut client = StorageServiceClient::new(self.channel.clone());
         let response = client.process_list_root_keys(request).make_sync().await?;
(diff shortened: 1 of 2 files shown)


linera-views/src/store.rs
```diff
@@ -165,10 +165,7 @@ pub trait KeyValueDatabase: WithError + Sized {
 
     /// Lists the root keys of the namespace.
     /// It is possible that some root keys have no keys.
```
ma2bd, line 167: > It is possible that some root keys have no keys.

I don't understand this comment. Did you mean that a root key can be empty?

linera-storage-service/src/client.rs
```diff
@@ -518,16 +518,12 @@ impl KeyValueDatabase for StorageServiceDatabaseInternal {
         Ok(namespaces)
     }
 
-    async fn list_root_keys(
-        config: &Self::Config,
-        namespace: &str,
-    ) -> Result<Vec<Vec<u8>>, StorageServiceStoreError> {
-        let namespace = bcs::to_bytes(namespace)?;
```
ma2bd, line 525: did we remove this intentionally?

ma2bd: never mind

linera-views/src/store.rs
MathieuDutSik: Yes, in some databases, though not all.

ma2bd: Ok let's fix the cryptic comment

PR #4864, opened 2025-10-27
`linera-service`: print faucet URL even if no faucet chain is given

Twey:
## Motivation

The faucet URL is accidentally not printed if `--faucet-chain` is not passed explicitly.

## Proposal

Print it whenever `--with-faucet` is used.

## Test Plan

Tested locally, but we don't rely on this behaviour anywhere.

## Release Plan

## Links


afck:
(This could also make sense to backport.)

PR #4865, opened 2025-10-27
[testnet] introduce new database schema and migration tooling.

MathieuDutSik:
## Motivation

The existing database schema is flawed because it consolidates all blobs, blob states, confirmed blocks, and events under a single root key (the shared one `&[]`).

We change this here.

## Proposal

The new schema is chosen with the following goal in mind:
* The existing root_keys of `ChainState` and `BlockExplorer` remain in the same place.
* The partition keys are short: At most 33 bytes.
* The grouping is via CryptoHash, BlobId, ChainId, or index.
* The `RootKey` type is only used for getting the `root_key` in contrast to `BaseKey`, which serves two purposes: root_key for some variants, key for some others.

Of utmost importance is to have a viable migration path. This is done according to the following points:
* For the client, the migration is done automatically at startup. The same applies to indexers, exporters, and so on. This is possible because those applications do not share a storage.
* There is a special key indicating the working schema. For old schema, its absence indicates the state of the database.
* For servers and proxies, there is no way to migrate startup. So a migration tool `linera-migration` is introduced.
* At startup, for proxies and servers, we check whether the storage has been migrated. If not we stop right away.
* However, when running from a blank state, it works. We determine whether we are 

For testing, a test `test_storage_migration` is introduced. It does the following:
* It creates a storage according to the old schema.
* It does the migration
* It reads the storage state and then checks for equality.
* This is done for `RocksDB`, `ScyllaDB`, and `memory`. This is not done for `DynamoDB` because of a problem with this database that will soon be addressed.

Of relevance is the annoying fact that root keys of the form `&[]` are not handled correctly for `RocksDB` and `linera-storage-service`. This turns out not to be a problem:
* Yes, doing a `find_keys_by_prefix(&[])` on the shared partition would get you keys from other root keys. However, that use case does not occur.
* Instead, what we have is `find_keys_by_prefix(&[x])` for `x` a byte that is NOT the first byte of other root keys. So it works.

## Test Plan

The CI.

## Release Plan

TestNet Conway is the target.

## Links

None.


linera-indexer/lib/src/rocks_db.rs
```diff
@@ -69,12 +70,22 @@ impl RocksDbRunner {
...
         };
         let namespace = config.client.namespace.clone();
-        let database = RocksDbDatabase::maybe_create_and_connect(&store_config, &namespace).await?;
+        let store_config = StoreConfig::RocksDb {
+            config: rocksdb_store_config.clone(),
+            namespace: namespace.clone(),
+        };
+        store_config
+            .clone()
+            .run_with_store(StorageMigration)
```
afck, line 84: I think there was also the idea yesterday to make this part of `InitializeStorageJob`?

linera-storage/src/db_storage.rs
```diff
@@ -224,37 +225,55 @@ pub mod metrics {
     });
 }
 
+/// The default key used when the root_key contains the information.
```
```suggestion
/// The default key used when the root key contains all the information.
```

linera-storage/src/db_storage.rs
```diff
@@ -224,37 +225,55 @@ pub mod metrics {
...
+pub(crate) const DEFAULT_KEY: &[u8] = &[0];
+
+/// The second key used when the root_key contains the information.
```
```suggestion
/// The second key used when the root key contains all the information.
```

linera-storage/src/db_storage.rs
```diff
@@ -308,20 +328,45 @@ pub struct DbStorage<Database, Clock = WallClock> {
 }
 
 #[derive(Debug, Serialize, Deserialize)]
-enum BaseKey {
+pub(crate) enum RootKey {
     ChainState(ChainId),
-    Certificate(CryptoHash),
-    ConfirmedBlock(CryptoHash),
+    CryptoHash(CryptoHash),
```
afck: Should this be called `Block` or `Certificate` instead of `CryptoHash`?

linera-storage/src/db_storage.rs
```diff
@@ -308,20 +328,45 @@ pub struct DbStorage<Database, Clock = WallClock> {
...
+        return false;
+    }
+    root_key[0] == CHAIN_ID_TAG
 }
 
-const INDEX_CHAIN_ID: u8 = 0;
-const INDEX_BLOB_ID: u8 = 3;
-const INDEX_EVENT_ID: u8 = 5;
+const CHAIN_ID_TAG: u8 = 0;
+const BLOB_ID_TAG: u8 = 2;
```
afck: If we really keep the hybrid approach with both `enum RootKey` and explicit tag constants, I'd put the constants directly next to the `enum` definition.

linera-storage/src/db_storage.rs
```diff
@@ -545,7 +585,7 @@ where
         &self.clock
     }
 
-    #[instrument(level = "trace", target = "telemetry_only", skip_all, fields(chain_id = %chain_id))]
+    #[instrument(level = "trace", skip_all, fields(chain_id = %chain_id))]
```
afck: (Not your fault, but I think `fields(%chain_id)` also works.)

linera-storage/src/db_storage.rs
```diff
@@ -559,34 +599,30 @@ where
...
-        }
-        let results = store.contains_keys(keys).await?;
         let mut missing_blobs = Vec::new();
-        for (blob_id, result) in blob_ids.iter().zip(results) {
-            if !result {
+        for blob_id in blob_ids {
+            let root_key = RootKey::Blob(*blob_id).bytes();
+            let store = self.database.open_shared(&root_key)?;
+            let test = store.contains_key(DEFAULT_KEY).await?;
+            if !test {
```
```suggestion
            if !store.contains_key(DEFAULT_KEY).await? {
```

linera-storage/src/db_storage.rs
```diff
@@ -873,39 +894,40 @@ where
...
+        let root_key = RootKey::Event(event_id.chain_id).bytes();
+        let store = self.database.open_shared(&root_key)?;
         let exists = store.contains_key(&event_key).await?;
         #[cfg(with_metrics)]
         metrics::CONTAINS_EVENT_COUNTER.with_label_values(&[]).inc();
         Ok(exists)
     }
 
-    #[instrument(target = "telemetry_only", skip_all, fields(chain_id = %chain_id, stream_id = %stream_id, start_index = %start_index))]
+    #[instrument(skip_all, fields(chain_id = %chain_id, stream_id = %stream_id, start_index = %start_index))]
```
```suggestion
    #[instrument(skip_all, fields(%chain_id, %stream_id, %start_index))]
```

linera-storage/src/db_storage.rs
```diff
@@ -1013,35 +1031,41 @@ where
...
+    #[instrument(skip_all, fields(batch_size = batch.keys_value_bytes.len()))]
+    pub(crate) async fn write_batch(&self, batch: MultiPartitionBatch) -> Result<(), ViewError> {
+        if batch.keys_value_bytes.is_empty() {
             return Ok(());
         }
         let mut futures = Vec::new();
-        for (key, bytes) in batch.key_value_bytes {
-            let store = self.database.open_shared(&[])?;
+        for (root_key, key, bytes) in batch.keys_value_bytes {
+            let store = self.database.open_shared(&root_key)?;
```
afck, line 1066: `open_shared` at least involves a lot of cloning in some cases. Would it significantly reduce the overhead if we batched them by `root_key` here? (I think we talked about this yesterday: you'd need to put them in a map, keyed by root key?)

linera-storage/src/db_storage.rs
```diff
@@ -224,37 +225,55 @@ pub mod metrics {
     });
 }
 
+/// The default key used when the root_key contains the information.
+pub(crate) const DEFAULT_KEY: &[u8] = &[0];
+
+/// The second key used when the root_key contains the information.
+pub(crate) const ONE_KEY: &[u8] = &[1];
```
afck: Maybe it would be cleaner to create separate constants, even if some of them have the same value: `BLOCK_KEY`, `LITE_CERTIFICATE_KEY`, `BLOB_KEY`, `BLOB_STATE_KEY`, etc.

linera-storage/src/migration.rs
```diff
@@ -0,0 +1,572 @@
...
+    /// Version 0, all the blobs, certificates, confirmed blocks, events and network description on the same partition.
+    /// This is marked as default since it does not exist in the old scheme and would be obtained from `unwrap_or_default`.
+    #[default]
+    Version0,
+    /// Version 1, spreading by chain ID, crypto hash, and blob ID.
+    Version1,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+enum BaseKey {
```
afck: So this is the version 0 one? Please add a comment to make that clear.

linera-storage/src/migration.rs
```diff
@@ -0,0 +1,572 @@
...
+// This implies that data on those root keys do not need to be moved.
+// For other tag variants, that are on the shared partition, we need to move
+// the data into their own partitions.
+const MOVABLE_KEYS_0_1: &[u8] = &[1, 2, 3, 4, 5, 7];
+
+/// The total number of keys being migrated in a block.
+/// we use chunks to avoid OOM
+const BLOCK_KEY_SIZE: usize = 90;
+
+// We map a `BaseKey` in the shared partition to a `(RootKey,key)` in the
```
afck: Would be good to find a name for the non-root part of the key. "Main key"? Or "entry key"?
Then I'd write something like:
`` We map a base key in the shared partition to a root key-main key pair in the new schema. ``
Or:
`` We map a serialized `BaseKey` in the shared partition to a pair: the serialized `RootKey`, and the main key. ``

linera-storage/src/migration.rs
```diff
@@ -0,0 +1,572 @@
...
+    C: Clock + Clone + Send + Sync + 'static,
+    Database::Error: From<bcs::Error> + Send + Sync,
+{
+    async fn migrate_shared_partition(
+        &self,
+        first_byte: &u8,
+        keys: Vec<Vec<u8>>,
+    ) -> Result<(), ViewError> {
+        tracing::info!(
+            "migrate_shared_partition for {} keys starting with with {first_byte}",
```
afck: INFO-level logs are meant for the end user, so let's make them a bit easier to understand without knowledge of the code: explain that this is migrating the database schema from version 0 to 1, etc. But in this case, maybe this should be DEBUG? The user doesn't care about the first byte.

Or if we keep the INFO log here, maybe we can print the actual data type instead of the first byte? I.e. print that this is now migrating the blobs, blob states, etc.?

linera-storage/src/migration.rs
```diff
@@ -0,0 +1,572 @@
...
+                })
+                .collect::<Vec<Vec<u8>>>();
+            let store = self.database.open_shared(&[])?;
+            let values = store
+                .read_multi_values_bytes(chunk_base_keys.to_vec())
+                .await?;
+            let mut batch = MultiPartitionBatch::new();
+            for (base_key, value) in chunk_base_keys.iter().zip(values) {
+                let value = value.ok_or(ViewError::MissingEntries)?;
+                tracing::info!("base_key={base_key:?} value={value:?}");
```
afck: Would it be more readable to print them using `hex::encode`? And do we really want to print the full value? And probably not at INFO-level?

linera-storage/src/migration.rs
```diff
@@ -0,0 +1,572 @@
...
+            }
+            store.write_batch(batch).await?;
+        }
+        Ok(())
+    }
+
+    async fn migrate_v0_to_v1(&self) -> Result<(), ViewError> {
+        for first_byte in MOVABLE_KEYS_0_1 {
+            let store = self.database.open_shared(&[])?;
+            let keys = store.find_keys_by_prefix(&[*first_byte]).await?;
```
afck, line 151: (This will load all keys of each type into memory at the same time; that should still be fine, should it, even for the testnet validators? I.e. loading _all_ block hashes, or loading _all_ blob IDs?)

linera-storage/src/migration.rs
```diff
@@ -0,0 +1,572 @@
...
+        let schema = self.get_database_schema().await?;
+        if schema == SchemaVersion::Version0 {
+            self.migrate_v0_to_v1().await?;
+        }
+        self.write_database_schema(&SchemaVersion::Version1).await?;
+        Ok(())
+    }
+
+    pub async fn assert_is_migrated_database(&self) -> Result<(), ViewError> {
+        let root_key = RootKey::SchemaVersion.bytes();
```
afck: Did you mean `NetworkDescription`?

linera-storage/src/migration.rs
```diff
@@ -0,0 +1,572 @@
...
+        network_description: Option<Vec<u8>>,
+    }
+
+    fn get_vector(rng: &mut impl Rng, len: usize) -> Vec<u8> {
+        let mut v = Vec::new();
+        for _ in 0..len {
+            let value = rng.gen::<u8>();
+            v.push(value);
+        }
+        v
```
afck: Maybe more concise:

```suggestion
        rng.sample_iter(distributions::Standard).take(len).collect()
```

linera-storage/src/migration.rs
```diff
@@ -0,0 +1,572 @@
...
+        certificates: BTreeMap<CryptoHash, Vec<u8>>,
+        confirmed_blocks: BTreeMap<CryptoHash, Vec<u8>>,
+        blobs: BTreeMap<BlobId, Vec<u8>>,
+        blob_states: BTreeMap<BlobId, Vec<u8>>,
+        events: HashMap<EventId, Vec<u8>>,
+        block_exporter_states: BTreeMap<u32, Vec<(Vec<u8>, Vec<u8>)>>,
+        network_description: Option<Vec<u8>>,
+    }
+
+    fn get_vector(rng: &mut impl Rng, len: usize) -> Vec<u8> {
```
afck: Maybe `random_vector`? Or `create_vector`? `get_` is usually used for functions that return some already existing value. (Same for the other functions below.)

linera-storage/src/migration.rs
```diff
@@ -0,0 +1,572 @@
...
+        CryptoHash::test_hash(format!("rnd_val={rnd_val}"))
+    }
+
+    fn get_stream_id(rng: &mut impl Rng) -> StreamId {
+        let application_id = GenericApplicationId::System;
+        let stream_name = StreamName(get_vector(rng, 10));
+        StreamId {
+            application_id,
+            stream_name,
+        }
```
afck:
```suggestion
        let stream_name = StreamName(get_vector(rng, 10));
        StreamId::system(stream_name)
```

linera-storage/src/migration.rs
```diff
@@ -0,0 +1,572 @@
...
+        let stream_id = get_stream_id(rng);
+        let index = rng.gen::<u32>();
+        EventId {
+            chain_id,
+            stream_id,
+            index,
+        }
+    }
+
+    fn reorder_key_values(key_values: Vec<(Vec<u8>, Vec<u8>)>) -> Vec<(Vec<u8>, Vec<u8>)> {
```
afck: Why not just use `key_values.sort_unstable()`?

linera-storage/src/migration.rs
```diff
@@ -0,0 +1,572 @@
...
+            .collect::<BTreeMap<Vec<u8>, Vec<u8>>>();
+        map.into_iter().collect::<Vec<(Vec<u8>, Vec<u8>)>>()
+    }
+
+    fn get_storage_state() -> StorageState {
+        let mut rng = make_deterministic_rng();
+        let key_size = 5;
+        let value_size = 10;
+        // 0: the chain states.
+        let n_chain_id = 2;
```
afck: I think `chain_id_count` makes it clearer than `n_`. (Same for the others below.)

linera-storage/src/migration.rs
```diff
@@ -0,0 +1,572 @@
...
+            let value = get_vector(&mut rng, value_size);
+            certificates.insert(hash, value);
+        }
+        // 2: the confirmed blocks
+        let n_blocks = 2;
+        let mut confirmed_blocks = BTreeMap::new();
+        for _i_block in 0..n_blocks {
+            let hash = get_hash(&mut rng);
+            let value = get_vector(&mut rng, value_size);
+            confirmed_blocks.insert(hash, value);
```
afck, line 327: (Would be more realistic to use the same hashes as for the certificates. Not sure if it matters for the test.)

linera-storage/src/migration.rs
```diff
@@ -0,0 +1,572 @@
...
+        }
+        // 5: the events
+        let n_events = 2;
+        let mut events = HashMap::new();
+        for _i_event in 0..n_events {
+            let event_id = get_event_id(&mut rng);
+            let value = get_vector(&mut rng, value_size);
+            events.insert(event_id, value);
+        }
+        // 6: the block exports
```
afck:
```suggestion
        // 6: the block exporters
```

(Also below.)

linera-storage/src/migration.rs
```diff
@@ -0,0 +1,572 @@
...
+        Ok(())
+    }
+
+    fn is_valid_root_key(root_key: &[u8]) -> bool {
+        if root_key.is_empty() {
+            // It corresponds to the &[]
+            return false;
+        }
+        if root_key == [4] {
+            // It corresponds to the key of the schema database.
```
afck:
```suggestion
            // It corresponds to the key of the database schema.
```

linera-storage/src/migration.rs
```diff
@@ -0,0 +1,572 @@
...
+                            let event_id = EventId {
+                                chain_id,
+                                stream_id: restricted_event_id.stream_id,
+                                index: restricted_event_id.index,
+                            };
+                            events.insert(event_id, value);
+                        }
+                    }
+                    RootKey::SchemaVersion => {
+                        // No operation needed.
```
afck: I guess we could assert here that it's version 1 now.

linera-storage/src/db_storage.rs

linera-storage/src/db_storage.rs

linera-storage/src/db_storage.rs
MathieuDutSik: I have a problem with that in the sense that in the `BaseKey`, we have `Certificate` and `ConfirmedBlock` and they look very different. That is why I chose `CryptoHash`, which was neutral enough.

I take you on authority for what is the best name for the variant. Would it be `Block`?

linera-storage/src/db_storage.rs
MathieuDutSik: ok.

MathieuDutSik: The tag constants are needed because we use the serialization details in `list_blob_ids` and `list_chain_ids`.
We cannot treat the serialization purely as a black-box, unfortunately.

linera-storage/src/db_storage.rs
MathieuDutSik: Thanks.

linera-storage/src/db_storage.rs
afck: A certificate is just a block with validator signatures. (`LiteCertificate` is basically the list of signatures.)

Maybe `ConfirmedBlock`?

linera-storage/src/db_storage.rs
afck: Yes, but I think yesterday the idea was brought up to not use an enum at all, and always add the tag explicitly. (Not sure; we can discuss in the meeting today.)

linera-storage/src/migration.rs
MathieuDutSik: Thanks.

linera-storage/src/migration.rs
MathieuDutSik: Thanks.

linera-storage/src/migration.rs
MathieuDutSik: Thanks. I wanted to have the test pass yesterday, and writing the code was faster than searching for it.

linera-storage/src/migration.rs
MathieuDutSik: Thanks.

linera-storage/src/migration.rs
MathieuDutSik: Thanks.

linera-storage/src/migration.rs
MathieuDutSik: Oh, yes. Thank you very much.

linera-storage/src/migration.rs
MathieuDutSik: In our discussion yesterday, we had 16G as the total size on the validators.
We download only the keys. So, my guess is that it is ok.
Otherwise, we can do an iteration for the second key from 0 to 255, which I guess would be safer but also slower.

linera-storage/src/migration.rs
MathieuDutSik: Ok for `create_vector`.

linera-storage/src/migration.rs
MathieuDutSik: Ok.

linera-storage/src/migration.rs
MathieuDutSik: I do that for the blocks, but I kept the certificate loops.

linera-storage/src/db_storage.rs
MathieuDutSik: Ok, done.

linera-storage/src/db_storage.rs
MathieuDutSik: Ok, done.

linera-storage/src/db_storage.rs
```diff
@@ -225,14 +225,25 @@ pub mod metrics {
     });
 }
 
-/// The default key used when the root_key contains the information.
-pub(crate) const DEFAULT_KEY: &[u8] = &[0];
+/// The key used for blobs when the root key contains all the information.
```
afck: (For blobs, blob states, certificates and blocks, the root key now always contains all the information.)

```suggestion
/// The key used for blobs. The blob ID itself is contained in the root key.
```

linera-storage/src/db_storage.rs
```diff
@@ -338,6 +343,11 @@ pub(crate) enum RootKey {
     BlockExporterState(u32),
 }
 
+const CHAIN_ID_TAG: u8 = 0;
+const BLOB_ID_TAG: u8 = 2;
+const CHAIN_ID_LENGTH: usize = std::mem::size_of::<ChainId>();
+const BLOB_ID_LENGTH: usize = std::mem::size_of::<BlobId>();
```
afck: Maybe we should assert in a test that this is correct. This is the in-memory size, after all, and the relevant part is probably the BCS-encoded size?

(Edit: This is the last potential blocker from my side.)

linera-storage/src/migration.rs
```diff
@@ -40,6 +40,9 @@ enum BaseKey {
     NetworkDescription,
 }
 
+/// The key used the the database schema
```
afck:
```suggestion
/// The key used for the database schema version.
```

linera-storage/src/db_storage.rs
MathieuDutSik: Yes, right. Took me some to implement this idea, but you are completely correct.

linera-storage/src/migration.rs
MathieuDutSik: Thanks.

linera-storage/src/migration.rs
```diff
@@ -248,6 +254,36 @@ mod tests {
         network_description: Option<Vec<u8>>,
     }
 
+    impl StorageState {
+        fn append_storage_state(&mut self, storage_state: StorageState) {
+            for (chain_id, key_values) in storage_state.chain_ids_key_values {
+                self.chain_ids_key_values.insert(chain_id, key_values);
+            }
```
afck: Maybe shorter (also below):

```suggestion
            self.chain_ids_key_values.extend(storage_state.chain_ids_key_values);
```

linera-storage/src/db_storage.rs

linera-storage/src/db_storage.rs
MathieuDutSik: This is already tested in `test_root_key_chainstate_serialization` and `test_root_key_blob_serialization`.

linera-storage/src/migration.rs
MathieuDutSik: Thanks.

linera-storage/src/migration.rs
MathieuDutSik: Yes, not needed actually.

linera-storage/src/db_storage.rs

linera-storage/src/db_storage.rs
MathieuDutSik: I think it was a misunderstanding.
So far we are using `bcs` whenever possible and use different schemes when forced.

linera-storage/src/db_storage.rs
MathieuDutSik: Done and also for other entries.

linera-storage/src/migration.rs
MathieuDutSik: Yes, done.

linera-storage/src/migration.rs
MathieuDutSik: Ok, switched to "debug".

linera-storage/src/migration.rs
MathieuDutSik: I do not know a notion of "main key" in databases that I know.
The use of the word "key" is non-controversial due to code like that
```rust
    async fn read_value_bytes(&self, key: &[u8]) -> Result<Option<Vec<u8>>, Self::Error>;
```

linera-storage/src/migration.rs
MathieuDutSik: Switched to the code `db_storage_v0.rs`.

linera-storage/src/db_storage.rs
afck: But isn't that accidental? The in-memory size can depend on the target architecture and even compiler version. ~If this is about BCS, so we should use `bcs::serialized_size`. (Probably needs to be inside a `LazyLock` then.)~

afck: Actually, do we even need these?

If I understand correctly, we determine whether it's a chain ID by the first byte, then try to deserialize it. We can just detect when serialization fails, instead of first comparing the length.

That's much safer, because even if we change the types in a way that they aren't constant-size anymore (e.g. add a variant `BlobType::Foo(u8)`, with an additional byte) or we `Box` something, this would still work.

linera-storage/src/db_storage.rs
MathieuDutSik: We can put two keys together with that strategy and no map are needed.
This is a very limited improvement, but it is available.

afck: (Not a blocker from my side, but might make sense as a small optimization.)

MathieuDutSik: I implemented it.

linera-storage/src/db_storage.rs
MathieuDutSik: Well maybe.
But now, we have a somewhat rigid framework. And I like that. No escape.
Or if you have something really better, we could do it in another PR.

linera-storage/src/migration.rs
afck: Then maybe:

```suggestion
// We map a serialized `BaseKey` in the shared partition to a serialized `RootKey` and key in the
```

afck (commented):
Looks great! The only remaining potential blocker for me is the `mem::size_of` issue.

linera-storage/src/db_storage.rs
afck: But this code simply can't be expected to work on all platforms or with different Rust versions or anything. `mem::size_of` has very little to do with the BCS size. I'd rather use explicit numbers here then.

deuszx: I think @afck is right: if those sizes are supposed to be "BCS sizes" then we should mark that explicitly and not rely on the accidental equality between `mem::size_of` and `BCS`. As Andreas points out: `usize` is platform-dependent, BCS is not – so here's the fist place where this equality can break.

linera-storage/src/migration.rs
```diff
@@ -0,0 +1,600 @@
...
+const DATABASE_SCHEMA_KEY: &[u8] = &[0];
+
+const UNUSED_EMPTY_KEY: &[u8] = &[];
+// We choose the ordering of the variants in `BaseKey` and `RootKey` so that
+// the root keys corresponding to `ChainState` and `BlockExporter` remain the
+// same in their serialization.
+// This implies that data on those root keys do not need to be moved.
+// For other tag variants, that are on the shared partition, we need to move
+// the data into their own partitions.
+const MOVABLE_KEYS_0_1: &[u8] = &[1, 2, 3, 4, 5, 7];
```
bart-linera, line 41: It would be good to point out that 0 and 6 correspond to `ChainState` and `BlockExporter`, respectively (if I understand what this is correctly... it's not obvious even with the comment, I'm afraid :grimacing: )

linera-storage/src/db_storage.rs
afck: This is what I had in mind: https://github.com/afck/linera-protocol/commit/6de301bfe71496ce2deedb3021d3cd608a066198

I don't understand why the current code would silently ignore an entry with a blob or chain ID tag but unexpected length, but would return an error if it has the right length and deserialization fails. My commit would always error, because I think we expect that every root key starting with that tag is of that form? (Or is that not true?)

linera-storage/src/migration.rs
MathieuDutSik: Ok.

linera-storage/src/db_storage.rs
MathieuDutSik: @afck: Ok, I used your proposition.
@deuszx: Whether the comparison for equality is done with usize of size 32 or 64 does not matter.

linera-storage/src/db_storage.rs
```diff
@@ -1126,8 +1125,8 @@ where
         let root_keys = database.list_root_keys().await?;
         let mut blob_ids = Vec::new();
         for root_key in root_keys {
-            if root_key.len() == 1 + BLOB_ID_LENGTH && root_key[0] == BLOB_ID_TAG {
-                let root_key_red = &root_key[1..=BLOB_ID_LENGTH];
+            if !root_key.is_empty() && root_key[0] == BLOB_ID_TAG {
```
afck, line 1128: Good catch! :+1:

afck (approved):

PR #4867, opened 2025-10-28
New BlockExecuted notification & support querying app state at intermediate execution states.

deuszx:
## Motivation

Some applications may decide that it is safe to update their UIs whenever the block proposal is handled locally (i.e. it's executed, set as pending proposal and sent to validators) instead of waiting for the confirmed block notification.

## Proposal

Add new notification `Reason::BlockExecuted { height, hash }` which is emitted after staging the block locally (i.e. it's set as pending and executed). The `hash` can be used to query the applications state even before it's committed. To support this the `query_application` functionality was extended to support querying at intermediate, cached, execution state views. When querying, no hash is provided, we query the committed state. Similarly, if the provided hash is no longer in the cache (b/c it was already committted in the meantime or some other proposal superseeded this one), the committed state is queried.

## Test Plan

A test with notifications was extended to verify that the event is emitted.

## Release Plan

- These changes should
    - be released in a new SDK,
    - backported to Testnet.

## Links

### DIFF
--- a/linera-core/src/chain_worker/state.rs
+++ b/linera-core/src/chain_worker/state.rs
@@ -1274,9 +1274,18 @@ where
         let local_time = self.storage.clock().current_time();
         if let Some(requested_block) = block_hash {
             if let Some(mut state) = self.execution_state_cache.remove(&requested_block) {
+                // We try to use a cached execution state for the requested block.
+                // We want to pretend that this block is committed, so we set the next block height.
+                let next_block_height = self
+                    .chain
+                    .tip_state
+                    .get()
+                    .next_block_height
+                    .try_add_one()
+                    .expect("block height to not overflow");
                 let context = QueryContext {
                     chain_id: self.chain_id(),
-                    next_block_height: self.chain.tip_state.get().next_block_height,
+                    next_block_height,
                     local_time,
                 };
                 let outcome = state
--- a/linera-service/src/node_service.rs
+++ b/linera-service/src/node_service.rs
@@ -1027,8 +1027,8 @@ where
         let application_id: ApplicationId = application_id.parse()?;
 
         debug!(
-            chain_id = %chain_id,
-            application_id = %application_id,
+            %chain_id,
+            %application_id,
             "processing request for application:\n{:?}",
             &request
         );


linera-core/src/chain_worker/state.rs
```diff
@@ -1261,14 +1265,49 @@ where
...
-        let outcome = self
-            .chain
-            .query_application(local_time, query, self.service_runtime_endpoint.as_mut())
-            .await?;
-        Ok(outcome)
+        if let Some(requested_block) = block_hash {
+            if let Some(mut state) = self.execution_state_cache.remove(&requested_block) {
+                let context = QueryContext {
+                    chain_id: self.chain_id(),
+                    next_block_height: self.chain.tip_state.get().next_block_height,
```
afck: That should really be the one that belongs to the block, not our current chain state?

linera-core/src/chain_worker/state.rs
```diff
@@ -1261,14 +1265,49 @@ where
...
+                    })
+                    .await
+                    .query_application(context, query, self.service_runtime_endpoint.as_mut())
+                    .await
+                    .with_execution_context(ChainExecutionContext::Query)?;
+                self.execution_state_cache
+                    .insert_owned(&requested_block, state);
+                Ok(outcome)
+            } else {
+                tracing::debug!(requested_block = %requested_block, "requested block hash not found in cache, querying committed state");
```
afck, line 1306: We could check whether the most recent block hash (in the chain's tip state) matches, and otherwise maybe it should be an error?

linera-core/src/chain_worker/state.rs
```diff
@@ -1261,14 +1265,49 @@ where
...
+        block_hash: Option<CryptoHash>,
+    ) -> Result<QueryOutcome, WorkerError> {
         self.initialize_and_save_if_needed().await?;
         let local_time = self.storage.clock().current_time();
-        let outcome = self
-            .chain
-            .query_application(local_time, query, self.service_runtime_endpoint.as_mut())
-            .await?;
-        Ok(outcome)
+        if let Some(requested_block) = block_hash {
```
afck, line 1275: You could make this a `let … else` and return early in the simple case (currently `else`).

linera-core/src/chain_worker/state.rs
deuszx: If we were asking the committed state then we'd use the `next_block_height` which is the next, to-be, committed state. Why wouldn't we use the same for this query? the `tip_state.next_block_height` is the `next_block_height` as well for us here.

afck: If this is a pending block at height 5. If it were committed, `next_block_height` would be 6. It's a bit weird that if it's not committed yet, we query the same state, but with `next_block_height` 5. I think the query should pretend it's committed. Not a blocker from my side, though.

linera-core/src/chain_worker/state.rs
deuszx: We could do that. I thought that returning an error would make the API more difficult to use – the caller would have to handle the error and try the committed state. Usually the UI (caller) is interested in knowing what's the current state (the fact that we use a specific block hash is secondary, I think). It's not checking if some block hash (proposal) has been included in the state. If we allowed for at most one intermediate state in `execution_state_cache` I'd use that – but we allow for multiple – so we need to identify it.

linera-service/src/node_service.rs
```diff
@@ -1025,12 +1027,14 @@ where
         let application_id: ApplicationId = application_id.parse()?;
 
         debug!(
-            "Processing request for application {application_id} on chain {chain_id}:\n{:?}",
+            chain_id = %chain_id,
+            application_id = %application_id,
```
afck:
```suggestion
            %chain_id,
            %application_id,
```

linera-core/src/chain_worker/state.rs
deuszx: >  I think the query should pretend it's committed.

I think you're right 🤔

linera-service/src/node_service.rs
```diff
@@ -946,12 +946,13 @@ where
         application_id: ApplicationId,
         request: Vec<u8>,
         chain_id: ChainId,
+        state_hash: Option<CryptoHash>,
```
afck: Or `block_hash`?

deuszx: Ah, missed that one.

deuszx: Done.

linera-service/src/node_service.rs
deuszx: Done in [Address review comments](https://github.com/linera-io/linera-protocol/pull/4867/commits/67e20574598f556da76ea5f853fa28647d6a285e).

linera-core/src/chain_worker/state.rs
deuszx: Done in [Address review comments](https://github.com/linera-io/linera-protocol/pull/4867/commits/67e20574598f556da76ea5f853fa28647d6a285e).

linera-core/src/chain_worker/state.rs
deuszx: Like you said in the other comment – we should pretend it's using the committed state.

linera-core/src/chain_worker/state.rs
```diff
@@ -1291,10 +1291,10 @@ where
                     .await
                     .with_execution_context(ChainExecutionContext::Query)?;
                 self.execution_state_cache
-                    .insert_owned(&requested_state, state);
+                    .insert_owned(&requested_block, state);
```
afck, line 1303: This is still a state, though, and not a block, is it? It's just _keyed_ by block hash?

linera-core/src/chain_worker/state.rs
```diff
@@ -1274,9 +1274,18 @@ where
...
             if let Some(mut state) = self.execution_state_cache.remove(&requested_block) {
+                // We try to use a cached execution state for the requested block.
+                // We want to pretend that this block is committed, so we set the next block height.
+                let next_block_height = self
+                    .chain
+                    .tip_state
+                    .get()
+                    .next_block_height
+                    .try_add_one()
+                    .expect("block height to not overflow");
```
afck, line 1285: (Are we clearing the cache whenever we commit a block? We probably should.)

deuszx: We do, I didn't touch that part.

linera-core/src/chain_worker/state.rs
deuszx: I don't understand 🤔 I'm putting the state back to the cache so that it can be removed when it's (block) actually being committed/confirmed.

afck: Yes, I'm saying it should be called `requested_state`, not `requested_block`, should it?

linera-core/src/chain_worker/state.rs
```diff
@@ -1261,14 +1265,58 @@ where
...
+    ) -> Result<QueryOutcome, WorkerError> {
         self.initialize_and_save_if_needed().await?;
         let local_time = self.storage.clock().current_time();
-        let outcome = self
-            .chain
-            .query_application(local_time, query, self.service_runtime_endpoint.as_mut())
-            .await?;
-        Ok(outcome)
+        if let Some(requested_block) = block_hash {
+            if let Some(mut state) = self.execution_state_cache.remove(&requested_block) {
```
afck, line 1276: Actually, would it be better to clone it rather than remove it? Because while it's removed from the cache, we might end up needlessly re-executing it.

linera-core/src/chain_worker/state.rs
deuszx: Why? the key is still the block (hash).

linera-core/src/chain_worker/state.rs
deuszx: We don't have the `: Clone` boundary so that's why it's being removed and then put back in.

linera-core/src/chain_worker/state.rs
afck: But `requested_block` is an execution state, not a block, is it?

linera-core/src/chain_worker/state.rs
afck: But doesn't it have a `clone_unchecked`? And wasn't this how it was created in the first place?

linera-core/src/chain_worker/state.rs
deuszx: `requested_block` is a block hash, `state` is the execution state.

linera-core/src/chain_worker/state.rs
deuszx: I don't see any `clone_` method neither on `ValueCache` or `ExecutionStateView`. And IIRC, `clone_unchecked` was being removed

linera-core/src/chain_worker/state.rs
afck: Ah, of course! :facepalm:

PR #4868, opened 2025-10-28
Support querying state at cached views; New BlockExecuted notification

deuszx:
Backport of #4867 

## Motivation

Some applications may decide that it is safe to update their UIs whenever the block proposal is handled locally (i.e. it's executed, set as pending proposal and sent to validators) instead of waiting for the confirmed block notification.

## Proposal

Add new notification `Reason::BlockExecuted { height, hash }` which is emitted after staging the block locally (i.e. it's set as pending and executed). The `hash` can be used to query the applications state even before it's committed. To support this the `query_application` functionality was extended to support querying at intermediate, cached, execution state views. When querying, no hash is provided, we query the committed state. Similarly, if the provided hash is no longer in the cache (b/c it was already committted in the meantime or some other proposal superseeded this one), the committed state is queried.

## Test Plan

A test with notifications was extended to verify that the event is emitted.

## Release Plan

- These changes should
    - backported to Testnet.

## Links

### DIFF
--- a/examples/counter/index.html
+++ b/examples/counter/index.html
@@ -84,22 +84,26 @@
               logs.insertBefore(entry, logs.firstChild);
           }
 
-          async function updateCount() {
-              const response = await counter.query('{ "query": "query { value }" }');
+          async function updateCount(block_hash) {
+              const response = await counter.query('{ "query": "query { value }" }', block_hash || '');
               document.getElementById('count').innerText
                   = JSON.parse(response).data.value;
           }
 
-          updateCount();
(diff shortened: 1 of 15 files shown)


deuszx:
> Looks good. (Assuming no major merge conflicts?)

The only merge conflicts (for all backports rn) come from the fact that on `main` the `client/mod.rs` was split into `chain_client` and some methods now live in two files.

afck:
Good point. I wonder if we should backport that refactoring. (@Twey)

PR #4869, opened 2025-10-28
Remove semaphore from RequestsScheduler

deuszx:
## Motivation

Semaphore might lead to deadlocks if buggy usage and we don't want know yet if its usage here is improvement.

## Proposal

Remove Semaphore, track current in-flight requests number only.

## Test Plan

CI

## Release Plan

## Links

### DIFF
--- a/CLI.md
+++ b/CLI.md
@@ -186,9 +186,6 @@ Client implementation and command-line tool for the Linera blockchain
 * `--max-joined-tasks <MAX_JOINED_TASKS>` — Maximum number of tasks that can are joined concurrently in the client
 
   Default value: `100`
-* `--max-in-flight-requests <MAX_IN_FLIGHT_REQUESTS>` — Maximum concurrent requests per validator node
-
-  Default value: `100`
 * `--max-accepted-latency-ms <MAX_ACCEPTED_LATENCY_MS>` — Maximum expected latency in milliseconds for score normalization
 
   Default value: `5000`


linera-core/src/client/requests_scheduler/scheduler.rs
```diff
@@ -475,18 +475,22 @@ impl<Env: Environment> RequestsScheduler<Env> {
         let public_key = peer.public_key;
 
         // Acquire request slot
```
afck: Comment is now outdated.

linera-core/src/client/requests_scheduler/scheduler.rs
```diff
@@ -475,18 +475,22 @@ impl<Env: Environment> RequestsScheduler<Env> {
...
         // Execute the operation
         let result = operation(peer).await;
 
         // Update metrics and release slot
         let response_time_ms = start_time.elapsed().as_millis() as u64;
-        drop(permit); // Explicitly drop the permit to release the slot
+        {
+            let mut nodes = self.nodes.write().await;
+            let node = nodes.get_mut(&public_key).expect("Node must exist");
+            node.current_load = node.current_load.saturating_sub(1);
```
afck: I wonder if we still need some kind of guard with a `Drop` implementation: the future could get canceled while `operation(peer)` is being `await`ed?

deuszx: Ah, true 😬 crazy.

I'm just gonna remove this field entirely. We can add it later if someone decides it's needed again.

deuszx: Removed in https://github.com/linera-io/linera-protocol/pull/4869/commits/0f130be909d9445c964cda901cfa095cb7018de9.

linera-core/src/client/requests_scheduler/scheduler.rs
deuszx: Removed in https://github.com/linera-io/linera-protocol/pull/4869/commits/0f130be909d9445c964cda901cfa095cb7018de9.

linera-core/src/client/requests_scheduler/scheduler.rs
afck: But if we don't take the current load into account, won't that send all requests to the same node (the one with the best score)?

deuszx: No, we don't use that feature yet.

afck: You could use an `AtomicUsize` as the counter, and wrap it with a `Drop` impl.

Or, probably that's pretty much the same as cloning an `Arc<()>` and using `strong_count`.

deuszx: `AtomicUsize` won't work. I'd need `Mutex<_>` but `NodeInfo` is within the `RwLock` anyway.

afck (approved):

linera-core/src/client/requests_scheduler/scheduler.rs
afck: Sure, it would need to be `Arc<AtomicUsize>`.

But a simple `Arc<()>` and using `strong_count()` as a counter would also work, would it? (Not a blocker, just curious!)

deuszx: Sounds like a neat trick indeed :)

deuszx:
Merging this since the compatiblity tests seem to be non-det. We need to look into it separately.

PR #4872, opened 2025-10-29
Add more tombstone data to scylla dashboards

ndr-ds:
## Motivation

The Scylla monitoring dashboards lacked visibility into tombstone-related metrics, which are critical for diagnosing performance issues and understanding compaction behavior.

## Proposal

Enhanced the Scylla dashboards with additional tombstone metrics and visualizations to help operators identify when tombstones are causing performance degradation.

## Test Plan

- Deployed updated dashboards to test environment
- Verified new metrics display correctly
- Validated dashboard functionality

## Release Plan

- These changes should be backported to the latest testnet branch, then
    - be released in a validator hotfix.

PR #4871, opened 2025-10-28
Make sure that the Scylla config ConfigMap exists when Scylla gets created

ndr-ds:
## Motivation

We're setting Scylla settings using this ConfigMap, but there's actually no guarantee that it will exist by the time Scylla starts. I've seen networks deployed that just ignore the ConfigMap and use default settings, for example, because the ConfigMap wasn't created yet when Scylla started.

## Proposal

Enforce that the ConfigMap will be created before Scylla starts.

## Test Plan

Deployed a network, it uses the ConfigMap options more reliably now

## Release Plan

- These changes should be backported to the latest `testnet` branch, then
    - be released in a validator hotfix.

PR #4870, opened 2025-10-28
Fix small bug on exporter config

ndr-ds:
## Motivation

`sh` doesn't support `set -euo pipefail`, I missed this in a previous PR

## Proposal

Use `bash` when available, otherwise use `set -eu` instead.

## Test Plan

Deployed a network, exporter starts as expected now.

## Release Plan

- These changes should be backported to the latest `testnet` branch, then
    - be released in a validator hotfix.

PR #4873, opened 2025-10-29
Stop prepending all root_keys with a 0x00 byte

ndr-ds:
## Motivation

During high-volume benchmark testing with ScyllaDB, we observed severe performance degradation caused by a partition key distribution issue. Shard 0 was receiving 4x more writes than other shards (226k vs 53-78k writes), leading to:

- 800+ MB compaction backlog on shard 0 while other shards had 20-30 MB
- 355 L0 SSTables accumulated (target: 4)
- Write latency spikes under load
- Benchmark failures due to timeout errors

Root cause analysis revealed that all root keys were being prepended with an extra 0x00 byte when opening ScyllaDB shared connections, causing hash clustering on a single shard.

## Proposal

Modified get_big_root_key() in linera-views/src/backends/scylla_db.rs to only prepend 0x00 for empty root keys (to satisfy ScyllaDB's non-empty key requirement), while using non-empty keys as-is for natural hash distribution.

When millions of keys start with identical 0x00 0x00 prefix:

1. ScyllaDB's Murmur3 hash function processes inputs in 4-byte chunks
2. Despite Murmur3's excellent avalanche properties, small statistical biases amplify at scale
3. The biased-token-round-robin algorithm assigns disproportionate tokens to shard 0

The fix ensures non-empty keys maintain their natural hash distribution across all shards.

## Test Plan

Benchmark testing on GKE cluster andre-test-tizpyj:

Before fix (andre-test-w9guzl):

- Shard 0: 226,000 writes (4x average) ⚠️
- Compaction backlog: 800+ MB on shard 0, 20-30 MB on others
- L0 SSTables: 355 (target: 4)
- Key format: 0x0000... (double prefix)

After fix (andre-test-tizpyj):

- Write distribution:
Shard 0: 132,596 writes (20.9%)
Shard 1: 79,454 writes (12.5%)
Shard 2: 155,011 writes (24.4%)
Shard 3: 54,178 writes (8.5%)
Shard 4: 130,532 writes (20.6%)
Shard 5: 54,327 writes (8.5%)
Shard 6: 28,727 writes (4.5%)
- Max deviation: 1.7x average (vs 4x before) ✅
- Compaction backlog: 0 MB across all shards ✅
- L0 SSTables: Reduced from 355 → 16 ✅
- ScyllaDB write batch p99 latency: Consistent 10ms ✅
- Key format: 0x00XX... (single tag byte) ✅

Verification:
Sampled 1.1M keys from kv.default table - all keys verified with correct single-byte tag format, no keys found with double 0x00 0x00 prefix.

## Release Plan

- These changes should be backported to the latest testnet branch, then

### DIFF
--- a/linera-views/src/backends/scylla_db.rs
+++ b/linera-views/src/backends/scylla_db.rs
@@ -607,6 +607,10 @@ pub enum ScyllaDbStoreInternalError {
     /// A next row error in ScyllaDB
     #[error(transparent)]
     NextRowError(#[from] NextRowError),
+
+    /// Empty root keys are not supported
+    #[error("Empty root keys are not supported with ScyllaDB")]
+    EmptyRootKey,
 }
 
 impl KeyValueStoreError for ScyllaDbStoreInternalError {
@@ -724,10 +728,12 @@ impl DirectWritableKeyValueStore for ScyllaDbStoreInternal {
 }
 
 // ScyllaDB requires that the keys are non-empty.
-fn get_big_root_key(root_key: &[u8]) -> Vec<u8> {
-    let mut big_key = vec![0];
-    big_key.extend(root_key);
-    big_key
+// Empty root keys are not allowed.
+fn get_big_root_key(root_key: &[u8]) -> Result<Vec<u8>, ScyllaDbStoreInternalError> {
+    if root_key.is_empty() {
+        return Err(ScyllaDbStoreInternalError::EmptyRootKey);
+    }
+    Ok(root_key.to_vec())
 }
 
 /// The type for building a new ScyllaDB Key Value Store
@@ -774,7 +780,7 @@ impl KeyValueDatabase for ScyllaDbDatabaseInternal {
         let store = self.store.clone();
         let semaphore = self.semaphore.clone();
         let max_stream_queries = self.max_stream_queries;
-        let root_key = get_big_root_key(root_key);
+        let root_key = get_big_root_key(root_key)?;
         Ok(ScyllaDbStoreInternal {
             store,
             semaphore,
@@ -837,7 +843,6 @@ impl KeyValueDatabase for ScyllaDbDatabaseInternal {
         let mut root_keys = BTreeSet::new();
         while let Some(row) = rows.next().await {
             let (root_key,) = row?;
-            let root_key = root_key[1..].to_vec();
             root_keys.insert(root_key);
         }
         Ok(root_keys.into_iter().collect::<Vec<_>>())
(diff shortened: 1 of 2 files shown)


ndr-ds:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4873?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#4873** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4873?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4873?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#4872** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4872?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4871** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4871?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4870** <a href="https://app.graphite.dev/github/pr/linera-io/linera-protocol/4870?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

linera-views/src/backends/scylla_db.rs
```diff
@@ -725,9 +725,14 @@ impl DirectWritableKeyValueStore for ScyllaDbStoreInternal {
 
 // ScyllaDB requires that the keys are non-empty.
 fn get_big_root_key(root_key: &[u8]) -> Vec<u8> {
-    let mut big_key = vec![0];
-    big_key.extend(root_key);
-    big_key
+    if root_key.is_empty() {
+        // Only prepend 0x00 for empty keys to satisfy ScyllaDB's non-empty requirement
+        vec![0]
```
ma2bd: Doesn't it create a collision between `vec![0]` and `vec![]`?

ma2bd: Seems like we should "just" disallow empty root keys -- which is now possible with the new DB schema.

deuszx: Yes, if we can get rid of this special casing even better.

MathieuDutSik: Yes, since we are no longer using the `&[]`, I think we can require the `root_key` to be non-empty.
This nukes the idea of using `v0` / `v1` for the schema, but that is fine.
Also, the correction can be done in DynamoDB, where the prepending is done as well.

Note that I am quite skeptical of the explanation for the concentration of the keys. But fine, now we can drop the initial 0.

Issue #4874, opened 2025-10-29
When calling `linera query-balance`, it got stuck


kikakkz:
**Description**
When query balance in a chain of the wallet, it'll run infinitely with error log `2025-10-29T08:34:35.123499Z  INFO linera::main: linera_core::client: Message failed to execute locally and will be rejected. error=Required execution fees exceeded the total funding available. Fees 0.00001, available balance: 0. origin=c9a7a945036d9200e73fe49eedc30049b6789965c556f5ebbfa4cfd77709bf14 chain_id=4e2b6066a4d2cbd4158e64a86408e4d543605397a719ee98c8962478625cebe6`. So I transfer some amount from another chain to this chain but it won't recover.

**Steps**
1 Download the debug wallet `https://drive.google.com/file/d/1BJDQuqKNC6R9NqI-jHgBOQjTjGFyMY8h/view?usp=drive_link`
2 Decompress the wallet `tar xvvf wallet.tar.gz`
3 Compile `linera` binary with `testnet_conway` branch
4 `linera --wallet wallet/wallet.json --keystore wallet/keystore.json --storage rocksdb:wallet/client.db query-balance 4e2b6066a4d2cbd4158e64a86408e4d543605397a719ee98c8962478625cebe6`

**Expected result**
The chain balance should be returned.

**Actual result**
1 It will start with incoming messages
```
2025-10-29T08:34:44.630524Z  INFO linera::main: linera: Evaluating the local balance of 4e2b6066a4d2cbd4158e64a86408e4d543605397a719ee98c8962478625cebe6:0x00 by staging execution of known incoming messages                       
[...]

PR #4875, opened 2025-10-29
No fees for rejecting

afck:
## Motivation

We charge fees for all outgoing messages, including bounced ones. That means that a chain with very low balance can become unable to reject tracked messages.

The client doesn't expect this, and keeps looping and setting the failing message to `Reject`, even if it already is.

## Proposal

Don't charge for bouncing messages.

Also, charge only for the length of the operation outcome payload, rather than the length of its serialized size, so that an empty outcome is free, too.

Finally, we now break out of the loop in the client if a _rejected_ message fails.

## Test Plan

A new test was added that reproduced the issue.

## Release Plan

- I will prepare a separate PR for a migration of `testnet_conway` to this new execution semantics. We can decide later whether we want to do that.

## Links

- Found in: https://github.com/linera-io/linera-protocol/issues/4874

### DIFF
--- a/linera-chain/src/block_tracker.rs
+++ b/linera-chain/src/block_tracker.rs
@@ -325,7 +325,7 @@ impl<'resources, 'blobs> BlockExecutionTracker<'resources, 'blobs> {
         }
 
         self.resource_controller
-            .track_block_size(txn_outcome.operation_result.len())
+            .track_block_size_of(&(&txn_outcome.operation_result))
             .with_execution_context(context)?;
 
         self.next_application_index = txn_outcome.next_application_index;


linera-chain/src/block_tracker.rs
```diff
@@ -322,7 +325,7 @@ impl<'resources, 'blobs> BlockExecutionTracker<'resources, 'blobs> {
         }
 
         self.resource_controller
-            .track_block_size_of(&(&txn_outcome.operation_result))
+            .track_block_size(txn_outcome.operation_result.len())
```
deuszx: if empty outcome is non-empty, when serialized, it shouldn't be free

afck: We already moved away from charging _exactly_ for the serialized size of a block, and are only tracking its contents and not most of its header fields.

Would you prefer to keep including the whole serialized size, but making another exception here if the transaction was a rejected incoming bundle?

afck: Ah, wait, this doesn't even matter: We're not charging anything here, we just check the block size limit! I'll revert this.

afck: Done in https://github.com/linera-io/linera-protocol/pull/4875/commits/c8f4c6059ab93b70653680c19446476c670742b6.

PR #4877, opened 2025-10-29
Extend MessagePolicy filtering to events

bart-linera:
## Motivation

We might want to process the inbox of a chain selectively in some cases (eg. if a chain has a lot of incoming messages in the inbox and lacks the funds to process them, we could send a transfer and only process the messages from the sender of the transfer first).

## Proposal

~~Add a `--senders` argument to `process-inbox`, specifying the chains from which we want to process messages. If this argument is not present, process all incoming messages as before.~~

Turns out the feature already existed - only extend filtering by chain to events. Also, test message filtering by chain.

## Test Plan

CI

## Release Plan

    - be released in a new SDK

## Links

### DIFF
--- a/linera-core/src/unit_tests/wasm_client_tests.rs
+++ b/linera-core/src/unit_tests/wasm_client_tests.rs
@@ -53,7 +53,7 @@ use crate::{
     client::{
         chain_client::{self, ChainClient},
         client_tests::{MemoryStorageBuilder, StorageBuilder, TestBuilder},
-        ClientOutcome,
+        BlanketMessagePolicy, ClientOutcome, MessagePolicy,
     },
     local_node::LocalNodeError,
     test_utils::{ClientOutcomeResultExt as _, FaultType},
@@ -709,7 +709,15 @@ where
     builder.set_fault_type([3], FaultType::Offline);
 
     let sender = builder.add_root_chain(0, Amount::ONE).await?;
-    let receiver = builder.add_root_chain(1, Amount::ONE).await?;
+    let sender2 = builder.add_root_chain(1, Amount::ONE).await?;
+    // Make sure that sender's chain ID is less than sender2's - important for the final
+    // query check
+    let (sender, sender2) = if sender.chain_id() < sender2.chain_id() {
+        (sender, sender2)
+    } else {
+        (sender2, sender)
+    };
+    let mut receiver = builder.add_root_chain(2, Amount::ONE).await?;
 
     let module_id = receiver.publish_wasm_example("social").await?;
     let module_id = module_id.with_abi::<social::SocialAbi, (), ()>();
@@ -719,12 +727,21 @@ where
         .await
         .unwrap_ok_committed();
 
-    // Request to subscribe to the sender.
+    // Request to subscribe to the senders.
     let request_subscribe = social::Operation::Subscribe {
         chain_id: sender.chain_id(),
     };
+    let request_subscribe2 = social::Operation::Subscribe {
+        chain_id: sender2.chain_id(),
+    };
     receiver
-        .execute_operation(Operation::user(application_id, &request_subscribe)?)
+        .execute_operations(
+            vec![
+                Operation::user(application_id, &request_subscribe)?,
+                Operation::user(application_id, &request_subscribe2)?,
+            ],
+            vec![],
+        )
         .await
         .unwrap_ok_committed();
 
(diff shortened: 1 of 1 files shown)


linera-core/src/client/chain_client/mod.rs
```diff
@@ -1276,26 +1282,27 @@ impl<Env: Environment> ChainClient<Env> {
...
         operations: Vec<Operation>,
+        senders: &BTreeSet<ChainId>,
     ) -> Result<Vec<Transaction>, Error> {
-        let incoming_bundles = self.pending_message_bundles().await?;
-        let stream_updates = self.collect_stream_updates().await?;
+        let incoming_bundles = self
+            .pending_message_bundles()
+            .await?
+            .into_iter()
+            .filter(|bundle| senders.is_empty() || senders.contains(&bundle.origin));
```
afck: I'd pass this into `pending_message_bundles` instead:
1. Maybe _loading_ the messages in the first place can now be faster, because we don't even iterate over all inboxes.
2. `pending_message_bundles` already truncates the list at the end to the configured maximum number of messages. The selected messages could all be from other senders, so we would drop them, and miss the messages from the selected senders.

bart-linera: Hmm, good point!

Although for loading to be faster I'd have to change how `ChainInfoQuery` works, and that would affect not just the client side, but the validators as well, so maybe in a separate PR?

afck: Makes sense! But at least we should filter before truncating.

linera-core/src/client/chain_client/mod.rs
```diff
@@ -522,10 +525,9 @@ impl<Env: Environment> ChainClient<Env> {
             .requested_pending_message_bundles
             .into_iter()
             .filter_map(|mut bundle| {
-                self.options
-                    .message_policy
-                    .must_handle(&mut bundle)
-                    .then_some(bundle)
+                ((senders.is_empty() || senders.contains(&bundle.origin))
+                    && self.options.message_policy.must_handle(&mut bundle))
```
afck: Wait, `MessagePolicy` already contains a `restrict_chain_ids_to`. We already have that feature, do we? :facepalm:

linera-client/src/client_context.rs
```diff
@@ -493,12 +493,23 @@ impl<Env: Environment, W: Persist<Target = Wallet>> ClientContext<Env, W> {
...
         chain_client: &ChainClient<Env>,
+    ) -> Result<Vec<ConfirmedBlockCertificate>, Error> {
+        self.process_inbox_from_senders(chain_client, BTreeSet::new())
+            .await
+    }
+
+    pub async fn process_inbox_from_senders(
+        &mut self,
+        chain_client: &ChainClient<Env>,
+        senders: BTreeSet<ChainId>,
```
ma2bd: This should be an `Option<BTreeSet<ChainId>>`. In this kind of "allow list", the empty set means we don't accept anything.

linera-core/src/client/chain_client/mod.rs
bart-linera: You're right :man_facepalming: The only difference here is that I'm applying this to events as well - not sure if we want that with `MessagePolicy` or not? If yes, I can change the PR to just apply the `MessagePolicy` to events, too.

afck: Right, good catch! I think it makes sense to apply it to events, too, yes.

linera-core/src/unit_tests/client_tests.rs
```diff
@@ -2593,6 +2595,44 @@ where
...
+    assert_eq!(receiver.local_balance().await.unwrap(), Amount::ONE);
+
+    // Let's accept the other one, too.
+    receiver.options_mut().message_policy = MessagePolicy::new(BlanketMessagePolicy::Accept, None);
+    let certs = receiver.process_inbox().await?.0;
+    assert_eq!(certs.len(), 1);
+    assert_eq!(
+        receiver.local_balance().await.unwrap(),
+        Amount::from_tokens(2)
+    );
```
afck, line 2634: We should now test what's actually new, too: filtering of events. (Maybe that can easily be done as part of one of the `social` example tests?)

linera-client/src/client_context.rs
bart-linera: Turns out we already had filtering by chain in the `MessagePolicy`, so I removed this completely (and `MessagePolicy` has an `Option<HashSet<ChainId>>`, so it's correct).

afck (commented):
PR title is outdated.

linera-core/src/unit_tests/client_tests.rs
bart-linera: Added in d246af2

linera-core/src/unit_tests/wasm_client_tests.rs
```diff
@@ -780,6 +797,73 @@ where
...
+        Some([sender.chain_id()].into_iter().collect()),
+    );
+
+    // Receiver should only process the event from sender now.
+    let certs = receiver.process_inbox().await.unwrap().0;
+    assert_eq!(certs.len(), 1);
+
+    // There should be an UpdateStreams operation due to the new post.
+    let operations = certs[0].block().body.operations().collect::<Vec<_>>();
+    let [Operation::System(operation)] = &*operations else {
```
afck, line 834: But even two streams would still be processed in a single operation, wouldn't they? Does this fail if you don't restrict the chain IDs?

bart-linera: I think that would be a single operation, but there is a vector of updated streams inside the operations, and a line below asserts that it only contains sender's stream.

PR #4880, opened 2025-10-30
Turn block heights list into ranges to load more at once from storage.

afck:
## Motivation

When handling a chain info query asking for sender certificate hashes, we currently load those hashes from the chain state view one by one: `chain.block_hashes(height..=height).await?`. In most cases, the block heights are actually contiguous.

## Proposal

Turn the list into an equivalent list of ranges, and load each range at once.

## Test Plan

CI should catch any regressions. We should find a way to try out if this improves performance, or even addresses the problems on the testnet.

## Release Plan

    - be released in a new SDK,

## Links

### DIFF
--- a/linera-core/src/chain_worker/state.rs
+++ b/linera-core/src/chain_worker/state.rs
@@ -1524,7 +1524,7 @@ where
             info.requested_pending_message_bundles = bundles;
         }
         let mut hashes = Vec::new();
-        let height_ranges = Self::into_ranges(query.request_sent_certificate_hashes_by_heights);
+        let height_ranges = into_ranges(query.request_sent_certificate_hashes_by_heights);
         for height_range in height_ranges {
             hashes.extend(chain.block_hashes(height_range).await?);
         }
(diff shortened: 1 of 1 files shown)


linera-core/src/chain_worker/state.rs
```diff
@@ -1523,8 +1524,9 @@ where
             info.requested_pending_message_bundles = bundles;
         }
         let mut hashes = Vec::new();
-        for height in query.request_sent_certificate_hashes_by_heights {
-            hashes.extend(chain.block_hashes(height..=height).await?);
+        let height_ranges = Self::into_ranges(query.request_sent_certificate_hashes_by_heights);
+        for height_range in height_ranges {
+            hashes.extend(chain.block_hashes(height_range).await?);
         }
```
deuszx, line 1530: Alternatively we can create a single range and then filter out here. Fewer roundtrips to the DB.

afck: Yes, but very bad if we just requested two block with a 1000-block gap in between.

deuszx: Very unlikely.

afck: Can totally happen, if a very busy chain only occasionally sends messages to a client. E.g. some DEX where I made a trade yesterday and another today, but in between thousands of other people did.

deuszx (approved):

PR #4881, opened 2025-10-30
Turn block heights list into ranges to load more at once from storage.

afck:
Backport of #4880.

## Motivation

When handling a chain info query asking for sender certificate hashes, we currently load those hashes from the chain state view one by one: `chain.block_hashes(height..=height).await?`. In most cases, the block heights are actually contiguous.

## Proposal

Turn the list into an equivalent list of ranges, and load each range at once.

## Test Plan

CI should catch any regressions. We should find a way to try out if this improves performance, or even addresses the problems on the testnet.

## Release Plan

- These changes should
    - be released in a new SDK,

## Links

- PR to main: #4880

### DIFF
--- a/linera-core/src/chain_worker/state.rs
+++ b/linera-core/src/chain_worker/state.rs
@@ -1518,7 +1518,7 @@ where
             info.requested_pending_message_bundles = bundles;
         }
         let mut hashes = Vec::new();
-        let height_ranges = Self::into_ranges(query.request_sent_certificate_hashes_by_heights);
+        let height_ranges = into_ranges(query.request_sent_certificate_hashes_by_heights);
         for height_range in height_ranges {
             hashes.extend(chain.block_hashes(height_range).await?);
         }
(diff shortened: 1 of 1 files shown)


linera-core/src/chain_worker/state.rs
```diff
@@ -1594,6 +1596,21 @@ where
...
+        let mut values_iter = values.into_iter().peekable();
+        iter::from_fn(move || {
+            let start = values_iter.next()?;
+            let mut end = start;
+            while values_iter.peek() == end.try_add_one().ok().as_ref() {
+                end = values_iter.next()?;
+            }
+            Some(start..=end)
+        })
+    }
```
deuszx: Didn't ask on the original PR but can you create a small unit test? Just to be sure

afck: Working on it… I'll add it to the other PR first.

PR #4882, opened 2025-10-31
[testnet] Pin several large futures. (#4883)

afck:
## Motivation

```
export LINERA_FAUCET_URL=https://faucet.testnet-conway.linera.net
cargo test test_wasm_end_to_end_fungible::remote_net_grpc --features remote-net
```
fails with a stack overflow.

## Proposal

`Box::pin` a lot of futures larger than 3 kB (using Claude).

## Test Plan

CI

This change fixed it for me locally.

## Release Plan

- These changes should be ported to `main`.
- These changes should be:
    - be released in a new SDK,

## Links

- PR to `main`: #4883

### DIFF
--- a/linera-core/src/chain_worker/state.rs
+++ b/linera-core/src/chain_worker/state.rs
@@ -1340,9 +1340,8 @@ where
         let (_, committee) = self.chain.current_committee()?;
         block.check_proposal_size(committee.policy().maximum_block_proposal_size)?;
 
-        let outcome = self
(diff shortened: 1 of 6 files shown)


afck:
> I have no idea how

My view is that stack overflows are caused by at least one of two problems:
* Too many stack frames.
* Too large stack frames.

The fact that we had lots of futures > 3 kB points to me that the second point is at least part of the problem. That it fixed the stack overflow indicates that we probably don't have unreasonably high call stacks anywhere.

PR #4885, opened 2025-11-01
Make the migration more granular.

MathieuDutSik:
## Motivation

We had the `linera storage check-existence` triggering the migration, which is wrong on more than one level:
* It is a read operation, it is very counter-intuitive that it triggers a migration.
* The `check-existence` is about a namespace. So whether it was migrated or not does not matter.

## Proposal

The following is done:
* For the initialization of storage, the migration is automatic.
* For `DatabaseToolCommand` now has a function `need_migration` and `need_assert_migration`.
* When running the storage operation, we do the migration if needed and assert if they are migrated.

## Test Plan

CI would help. But we need some additional tests.

## Release Plan

On TestNet Conway.

## Links

None.

### DIFF
--- a/CLI.md
+++ b/CLI.md
@@ -1305,7 +1305,7 @@ Operation on the storage
 * `list-namespaces` — List the namespaces in the database
 * `list-blob-ids` — List the blob IDs in the database
 * `list-chain-ids` — List the chain IDs in the database
-* `migrate` — Do the migration of the storage
+* `migrate` — Migrate storage from schema version 0 to version 1
 
 
 
@@ -1371,7 +1371,7 @@ List the chain IDs in the database
 
 ## `linera storage migrate`
 
-Do the migration of the storage
+Migrate storage from schema version 0 to version 1
 
 **Usage:** `linera storage migrate`
 


linera-service/src/cli/command.rs
```diff
@@ -1108,6 +1108,58 @@ pub enum DatabaseToolCommand {
 
     /// List the chain IDs in the database
     ListChainIds,
+
+    /// Do the migration of the storage
+    Migration,
```
afck:
```suggestion
    /// Migrate storage from schema version 0 to version 1.
    Migrate,
```

linera-service/src/storage.rs
```diff
@@ -758,6 +758,32 @@ impl RunnableWithStore for StorageMigration {
     }
 }
 
+pub struct IsStorageMigrated;
```
afck: Or just `StorageV1`?

afck:
> For the initialization of storage, the migration is automatic.

Is it?
Shouldn't migration _only_ happen when we explicitly call `migrate`, and otherwise only the assertions (V0 or V1 or neither) happen?

(If I read the code correctly, that's actually how you implemented it?)

MathieuDutSik:
> > For the initialization of storage, the migration is automatic.
> 
> Is it? Shouldn't migration _only_ happen when we explicitly call `migrate`, and otherwise only the assertions (V0 or V1 or neither) happen?
> 
> (If I read the code correctly, that's actually how you implemented it?)

That was not how things were asked on slack. The idea was that "linera storage initialize" would do the migration.
It made sense to me because there is one single call to the initialization, not concurrent calls.

linera-service/src/storage.rs
MathieuDutSik: Ok, replaced by "AssertStorageV1".

linera-service/src/cli/command.rs
MathieuDutSik: Ok, done.

PR #4886, opened 2025-11-01
Improve the `list_blob_ids` / `list_chain_ids`

MathieuDutSik:
## Motivation

Several changes suggested by the work on the storage are integrated here.

## Proposal

The following changes have been made:
* The `BLOB_ID_LENGTH` / `CHAIN_ID_LENGTH` are eliminated from the code just as in TestNet Conway.
* The `list_blob_ids` /  `list_chain_ids` are part of `DbStorage`.
* The functions are then integrated into the storage trait.

## Test Plan

The CI.

## Release Plan

## Links

None.

### DIFF
--- a/linera-storage/src/db_storage.rs
+++ b/linera-storage/src/db_storage.rs
@@ -337,6 +337,10 @@ enum RootKey {
     NetworkDescription,
 }
 
+const CHAIN_ID_TAG: u8 = 0;
+const BLOB_ID_TAG: u8 = 2;
+const EVENT_ID_TAG: u8 = 3;
+
 impl RootKey {
     fn bytes(&self) -> Vec<u8> {
         bcs::to_bytes(self).unwrap()
@@ -364,10 +368,6 @@ fn is_chain_state(root_key: &[u8]) -> bool {
     root_key[0] == CHAIN_ID_TAG
 }
 
-const CHAIN_ID_TAG: u8 = 0;
-const BLOB_ID_TAG: u8 = 2;
-const EVENT_ID_TAG: u8 = 3;
-
 #[cfg(test)]
 mod tests {
     use linera_base::{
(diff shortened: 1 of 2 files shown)


linera-storage/src/db_storage.rs
```diff
@@ -985,6 +984,34 @@ where
         let store = self.database.open_exclusive(&root_key)?;
         Ok(ViewContext::create_root_context(store, block_exporter_id).await?)
     }
+
+    async fn list_blob_ids(&self) -> Result<Vec<BlobId>, ViewError> {
+        let database = self.database.deref();
```
ma2bd: Do need this?

ma2bd (commented):
Can we also list event_ids while we're at it? :)

linera-storage/src/db_storage.rs
MathieuDutSik: For `list_chain_ids`, `list_blob_ids` indeed we can avoid it.
But for `list_event_ids`, we do need it.

MathieuDutSik:
> Can we also list event_ids while we're at it? :)

Yes.
This was requested after I created the PR.

linera-storage/src/db_storage.rs
```diff
@@ -358,8 +366,7 @@ fn is_chain_state(root_key: &[u8]) -> bool {
 
 const CHAIN_ID_TAG: u8 = 0;
 const BLOB_ID_TAG: u8 = 2;
-const CHAIN_ID_LENGTH: usize = std::mem::size_of::<ChainId>();
-const BLOB_ID_LENGTH: usize = std::mem::size_of::<BlobId>();
+const EVENT_ID_TAG: u8 = 3;
```
afck: I'd put these directly next to the `enum` that determines them.

linera-storage/src/lib.rs
```diff
@@ -400,6 +400,15 @@ pub trait Storage: Sized {
         &self,
         block_exporter_id: u32,
     ) -> Result<Self::BlockExporterContext, ViewError>;
+
+    /// Lists the blob IDs of the storage.
```
afck: (Maybe "in storage"?)

afck (approved):

linera-storage/src/lib.rs
MathieuDutSik: Ok.

linera-storage/src/db_storage.rs
MathieuDutSik: Ok, done.

PR #4890, opened 2025-11-02
Break out of the proposal staging loop if rejected message fails.

afck:
## Motivation

The staging loop loops forever if a rejected message fails. (It sets it to `Rejected`, which it already is, and tries again!)

After the `FLAG_FREE_REJECT` activation, this shouldn't be reachable anymore, but it makes sense to detect this case anyway.

## Proposal

Break out of the loop if a rejected message fails.

## Test Plan

I tried it locally.

## Release Plan

- Release in a new SDK. (Not urgent.)

## Links

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -1314,6 +1314,9 @@ impl<Env: Environment> Client<Env> {
                             "Protected incoming message failed to execute locally"
                         )
                     );
+                    if message.action == MessageAction::Reject {
+                        return result;
+                    }
                     // Reject the faulty message from the block and continue.
                     // TODO(#1420): This is potentially a bit heavy-handed for
                     // retryable errors.


linera-core/src/client/mod.rs
```diff
@@ -1314,6 +1314,9 @@ impl<Env: Environment> Client<Env> {
                             "Protected incoming message failed to execute locally"
                         )
                     );
+                    if message.action == MessageAction::Reject {
```
ma2bd, line 1317: Rejecting a message should never fail. So an `ensure!` would be consistent.

ma2bd: Or is this temporary until we roll out the other fix?

afck: `ensure!` is just a bit awkward in this case: we'd want to return the error that we know the `result` is. So I'd prefer just returning that over having `result.unwrap_err()` inside `ensure!`.

Or do you think it should be a different error?

I'd prefer to leave it like this on `testnet_conway`, because as far as the code goes, it's not unreachable. It's only unexpected once all chains have migrated to an epoch with `FLAG_FREE_REJECT`.

PR #4891, opened 2025-11-02
Make DB migration officially concurrent

ma2bd:
## Motivation

Simplify the deployment process by making migration tasks officially concurrent

## Proposal

We're already tested this (involuntarily) except the process was aborted.

## Test Plan

to be tested manually

### DIFF
--- a/linera-storage/src/migration.rs
+++ b/linera-storage/src/migration.rs
@@ -23,6 +23,8 @@ use crate::{
 
 #[derive(Debug)]
 enum SchemaVersion {
+    /// No schema version detected.
+    Uninitialized,
     /// Version 0. All the blobs, certificates, confirmed blocks, events and network
     /// description are on the same partition.
     Version0,
@@ -30,7 +32,8 @@ enum SchemaVersion {
     Version1,
 }
 
-/// How long we should wait before retrying when we detect another migration in progress.
+/// How long we should wait (in minutes) before retrying when we detect another migration
+/// in progress.
 const MIGRATION_WAIT_BEFORE_RETRY_MIN: u64 = 3;
 
 const UNUSED_EMPTY_KEY: &[u8] = &[];
@@ -142,7 +145,6 @@ where
             for key in chunk_base_keys {
                 batch.delete_key(key.to_vec());
             }
-            // Migrate chunk.
             store.write_batch(batch).await?;
         }
         Ok(())
@@ -161,7 +163,7 @@ where
         loop {
             if matches!(
                 self.get_storage_state().await?,
-                None | Some(SchemaVersion::Version1)
+                SchemaVersion::Uninitialized | SchemaVersion::Version1
             ) {
                 // Nothing to do.
                 return Ok(());
@@ -181,26 +183,29 @@ where
         }
     }
 
-    async fn get_storage_state(&self) -> Result<Option<SchemaVersion>, ViewError> {
+    async fn get_storage_state(&self) -> Result<SchemaVersion, ViewError> {
         let store = self.database.open_shared(&[])?;
         let key = bcs::to_bytes(&BaseKey::NetworkDescription).unwrap();
         if store.contains_key(&key).await? {
-            return Ok(Some(SchemaVersion::Version0));
+            return Ok(SchemaVersion::Version0);
         }
 
         let root_key = RootKey::NetworkDescription.bytes();
         let store = self.database.open_shared(&root_key)?;
         if store.contains_key(NETWORK_DESCRIPTION_KEY).await? {
-            return Ok(Some(SchemaVersion::Version1));
+            return Ok(SchemaVersion::Version1);
         }
 
-        Ok(None)
+        Ok(SchemaVersion::Uninitialized)
     }
 
     /// Assert that the storage is at the last version (or not yet initialized).
     pub async fn assert_is_migrated_storage(&self) -> Result<(), ViewError> {
         let state = self.get_storage_state().await?;
-        assert!(matches!(state, None | Some(SchemaVersion::Version1)));
+        assert!(matches!(
+            state,
+            SchemaVersion::Uninitialized | SchemaVersion::Version1
+        ));
         Ok(())
     }
 }


linera-storage/src/migration.rs
```diff
@@ -155,40 +158,49 @@ where
...
     }
 
+    /// Assert that the storage is at the last version (or not yet initialized).
     pub async fn assert_is_migrated_storage(&self) -> Result<(), ViewError> {
-        let (is_initialized, schema) = self.get_storage_state().await?;
-        if is_initialized {
-            assert_eq!(schema, SchemaVersion::Version1);
-        }
+        let state = self.get_storage_state().await?;
+        assert!(matches!(state, None | Some(SchemaVersion::Version1)));
```
MathieuDutSik: That `None | Some(SchemaVersion::Version1)` is not great.

linera-storage/src/migration.rs
```diff
@@ -155,40 +158,49 @@ where
...
+            }
+            let result = self.migrate_v0_to_v1().await;
+            if let Err(ViewError::MissingEntries) = result {
+                tracing::warn!(
+                    "It looks like a migration is already in progress on this database. \
+                     I will wait for {:?} minutes and retry.",
+                    MIGRATION_WAIT_BEFORE_RETRY_MIN
+                );
+                // Duration::from_mins is not yet stable for tokio 1.36.
+                tokio::time::sleep(Duration::from_secs(MIGRATION_WAIT_BEFORE_RETRY_MIN * 60)).await;
```
MathieuDutSik, line 179: Unfortunately, the `from_mins` seems to be part of unstable (how hard could it be?).

linera-storage/src/migration.rs
```diff
@@ -11,6 +11,7 @@ use linera_views::{
     ViewError,
 };
 use serde::{Deserialize, Serialize};
+use tokio::time::Duration;
```
MathieuDutSik, line 14: We have a `linera_base::time::Duration`. Should we use it?

linera-storage/src/migration.rs
```diff
@@ -20,17 +21,18 @@ use crate::{
...
+    /// Version 0. All the blobs, certificates, confirmed blocks, events and network
+    /// description are on the same partition.
     Version0,
-    /// Version 1, spreading by chain ID, crypto hash, and blob ID.
+    /// Version 1. New partitions are assigned by chain ID, crypto hash, and blob ID.
     Version1,
 }
 
+/// How long we should wait before retrying when we detect another migration in progress.
+const MIGRATION_WAIT_BEFORE_RETRY_MIN: u64 = 3;
```
MathieuDutSik: In the description, I would specify the time unit used and use seconds instead of minutes, as that seems more suitable for the task.

linera-storage/src/migration.rs
```diff
@@ -140,6 +142,7 @@ where
             for key in chunk_base_keys {
                 batch.delete_key(key.to_vec());
             }
+            // Migrate chunk.
```
MathieuDutSik: Maybe both batches could be build in the same loop.

linera-storage/src/migration.rs
```diff
@@ -155,40 +158,49 @@ where
...
+        loop {
+            if matches!(
+                self.get_storage_state().await?,
+                None | Some(SchemaVersion::Version1)
+            ) {
+                // Nothing to do.
+                return Ok(());
+            }
+            let result = self.migrate_v0_to_v1().await;
+            if let Err(ViewError::MissingEntries) = result {
```
MathieuDutSik, line 172: I think there is a potential loophole.
Process 1 reads the keys and starts migrating. It deletes 50 keys out of the 90 keys to delete, but it is not instantaneous.
Process 2 reads the keys minus the 50 deleted ones. It also migrates. While doing so, some blocks may be partially migrated. This is because if the migration of a chunk is interrupted, then that chunk will not be moved.

Is it likely to occur? No. We already have in the Kubernetes scripts the following vulnerability
```bash
if !linera storage check-existence {
    linera storage intialize
}
```

But never encountered a problem in practice.

----

Are we feeling lucky?
If not, there is a way to address that:
* Change the API of `fn migrate_v0_to_v1(&self) -> Result<usize, ViewError>` with the number of keys being migrated.
* End the loop when the number of keys being migrated is 0.

This would be safer, and if the migration is done the `find_keys_by_prefix` would be super fast.

afck: Aren't we first writing to the new partition, then deleting the entry from the old?

So what would actually go wrong? Isn't the only danger that some new partitions get written twice?

MathieuDutSik: I do not think that the deletion is atomic.
So, I could imagine a scenario where the deletion batch is implemented partially, and so one the other process is not reading with the same block.

Maybe it is far-fetched and will never ever occur in practice, just like `linera storage check-existence`/ `linera storage initialize`.

I proposed a schema to avoid that and increase confidence in the software. But if you think it is overkill, you can approve the PR.

ma2bd: > I do not think that the deletion is atomic.

Why not? `write_batch` is supposed to be atomic. (We don't have journaling for shared partitions, but AFAICT this just means we could get `ScyllaDbStoreInternalError::BatchTooLong` etc.)

linera-storage/src/migration.rs
ma2bd: why? I think that's pretty clear.

linera-storage/src/migration.rs
ma2bd: yes, that's what I wrote in the comment

linera-storage/src/migration.rs
ma2bd: no, it's just a web thing

linera-storage/src/migration.rs
MathieuDutSik: Ok.

linera-storage/src/migration.rs
ma2bd: Oh actually I misread the code and thought that was already the case.

linera-storage/src/migration.rs
MathieuDutSik: Yes, sorry for that.

linera-storage/src/migration.rs
MathieuDutSik: The multi-partition batch and the single-partition batch. Now, they are both named `batch` but it does not have to be that way.

linera-storage/src/migration.rs
afck: No, I don't think it's overkill if there is a potential issue! I just don't understand the issue! :sweat_smile: 

Even if the write only partially completes one of the new partitions, won't the next pass finish it? We are just moving individual key-value pairs. And if the deletion is partial, it will just write that key-value pair again?

ma2bd: Also, I don't see why that could be a problem (yet?). We only ever delete entries that have already been migrated

linera-storage/src/migration.rs
ma2bd: it's fine. I'll undo my change

linera-storage/src/migration.rs
ma2bd: ok

linera-storage/src/migration.rs
ma2bd: I could replace None by SchemaVersion::Uninitialized if you prefer

linera-storage/src/migration.rs
MathieuDutSik: Yes, but we delete keys that have already been migrated. That is correct.

However, if we reach the `ViewError::MissingEntries`, then the chunk is NOT migrated. And then we wait. The other process might write the `NetworkDescription`, and so the chunk will not be considered. So, I think we can have a scenario where some keys have not been migrated because of that.

Therefore, I think it makes sense to keep track of the size of the chunk being migrated and to terminate when no error occurs, and the size being moved is 0.

afck: My understanding is probably wrong, but it was something like this:

If process 2 wrote the network description, that means it successfully moved every key that it found. I.e. every key that wasn't deleted by process 1 was moved by process 2.

The keys that _were_ deleted by process 1 were moved by process 1.

So the network description can only be written once all keys have been moved?

MathieuDutSik: Ok, that looks correct.

afck (approved):

linera-storage/src/migration.rs
ma2bd: done

PR #4893, opened 2025-11-03
Allow zero key length in key and prefix for `DynamoDb`.

MathieuDutSik:
## Motivation

Recently, the operation `find_keys_by_prefix(&[])` has emerged as an operation on stores. This operation
does not work on `DynamoDb` since the prefix has to be non-empty as well as the key.

## Proposal

Add a `[1]` as prefix.

## Test Plan

The CI should catch the bugs. The `run_reads` has been updated for that.

## Release Plan

It is not that important since DynamoDB is not used on TestNet Conway.

## Links

None.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -340,7 +340,7 @@ version = "1.6.3"
 [workspace.dependencies.aws-sdk-dynamodb]
 default-features = false
 features = ["rt-tokio", "rustls"]
-version = "1.76.0"
+version = "1.88.0"
 
 [profile.release]
 lto = "thin"


ma2bd (commented):
So in general, value-keys can be empty but not partition-keys?

MathieuDutSik:
> So in general, value-keys can be empty but not partition-keys?

The way to deal with empty vs non-empty keys is by prefixing by `[1]` the key or root_key.

It used to be that we had a `root_key` that was empty: the infamous `&[]`. We no longer. So, requiring root_keys to be non-empty is a sensible decision. 

There are now contexts where we need to list all the keys of a store and the command is `find_keys_by_prefix(&[])`. Therefore, it is sensible to allow empty keys and empty prefixes.

afck:
Would an alternative be to simply use a different query (without the `AND k >= ?` part) if the specified prefix is empty?

MathieuDutSik:
> Would an alternative be to simply use a different query (without the `AND k >= ?` part) if the specified prefix is empty?

I tried that some years ago and as I remember it was not working. But I can retry.

PR #4897, opened 2025-11-03
Simplify and generalize `block_hashes`.

afck:
## Motivation

After introducing sparse chains, we use `block_hashes` not just for ranges, but for sparse lists of blocks, too.

## Proposal

Use `LogView::multi_get` (list of heights) instead of `read` (range). Make `block_hashes` accept a list of heights instead of a range.

## Test Plan

CI should catch regressions; no new functionality was added.

## Release Plan

    - be released in a new SDK, (low priority!)
    - be released in a validator hotfix. (low priority!)

## Links

- Related: #4880 

### DIFF
--- a/linera-chain/src/chain.rs
+++ b/linera-chain/src/chain.rs
@@ -1072,7 +1072,7 @@ where
         chain_id = %self.chain_id(),
         next_block_height = %self.tip_state.get().next_block_height,
     ))]
-    pub async fn block_hashes_by_height(
+    pub async fn block_hashes(
         &self,
         heights: impl IntoIterator<Item = BlockHeight>,
     ) -> Result<Vec<CryptoHash>, ChainError> {
--- a/linera-core/src/chain_worker/state.rs
+++ b/linera-core/src/chain_worker/state.rs
@@ -1570,7 +1570,7 @@ where
             info.requested_pending_message_bundles = bundles;
         }
         let hashes = chain
-            .block_hashes_by_height(query.request_sent_certificate_hashes_by_heights)
+            .block_hashes(query.request_sent_certificate_hashes_by_heights)
             .await?;
         info.requested_sent_certificate_hashes = hashes;
         if let Some(start) = query.request_received_log_excluding_first_n {
--- a/linera-core/src/updater.rs
+++ b/linera-core/src/updater.rs
@@ -522,7 +522,7 @@ where
                 .local_node
                 .chain_state_view(chain_id)
                 .await?
-                .block_hashes_by_height([height])
+                .block_hashes([height])
                 .await?
                 .into_iter()
                 .next()
@@ -561,7 +561,7 @@ where
                 .local_node
                 .chain_state_view(chain_id)
                 .await?
-                .block_hashes_by_height(heights)
+                .block_hashes(heights)
                 .await?;
             if !validator_missing_hashes.is_empty() {
                 // Send the requested certificates in order.
--- a/linera-service/src/cli/main.rs
+++ b/linera-service/src/cli/main.rs
@@ -1776,7 +1776,7 @@ impl Runnable for Job {
                     .await
                     .context("Failed to load chain")?;
                 let block_hash = chain_state_view
-                    .block_hashes_by_height([height])
+                    .block_hashes([height])
                     .await
                     .context("Failed to find a block hash for the given height")?[0];
                 let block = context


linera-chain/src/chain.rs
```diff
@@ -1069,40 +1068,31 @@ where
...
-        };
         // Everything up to (excluding) next_height is in confirmed_log.
-        let mut hashes = if let Ok(last_height) = next_height.try_sub_one() {
-            let usize_start = usize::try_from(start)?;
-            let usize_end = usize::try_from(end.min(last_height))?;
-            self.confirmed_log.read(usize_start..=usize_end).await?
-        } else {
-            Vec::new()
-        };
+        let (confirmed_heights, unconfirmed_heights): (Vec<_>, Vec<_>) = heights
```
MathieuDutSik: Not important at all, but don't we want to avoid the type annotation?

linera-core/src/chain_worker/state.rs
```diff
@@ -1570,11 +1569,9 @@ where
...
             info.requested_pending_message_bundles = bundles;
         }
-        let mut hashes = Vec::new();
-        let height_ranges = into_ranges(query.request_sent_certificate_hashes_by_heights);
-        for height_range in height_ranges {
-            hashes.extend(chain.block_hashes(height_range).await?);
-        }
+        let hashes = chain
+            .block_hashes(query.request_sent_certificate_hashes_by_heights)
+            .await?;
```
MathieuDutSik, line 1574: A clear win. Bravo.

linera-chain/src/chain.rs
```diff
@@ -1069,40 +1068,31 @@ where
     #[instrument(skip_all, fields(
         chain_id = %self.chain_id(),
         next_block_height = %self.tip_state.get().next_block_height,
-        start_height = ?range.start_bound(),
-        end_height = ?range.end_bound()
     ))]
     pub async fn block_hashes(
         &self,
-        range: impl RangeBounds<BlockHeight>,
+        heights: impl IntoIterator<Item = BlockHeight>,
```
MathieuDutSik, line 1077: When I see that, I have the following considerations:
* Apparently, the heights are not ordered. It might be the effective use case, but that is not written explicitly nor tested. Maybe rename as `unordered_heights`.
* The code is called from a `request_sent_certificate_hashes_by_heights`, so I would rename the function to `block_hashes_by_height`.

So, some nomenclature improvement could help here.

linera-chain/src/chain.rs
```diff
@@ -1069,40 +1068,31 @@ where
...
-        for hash in self
+        let unconfirmed_hashes = self
             .preprocessed_blocks
-            .multi_get(&block_heights)
-            .await?
+            .multi_get(&unconfirmed_heights)
+            .await?;
+        Ok(confirmed_hashes
             .into_iter()
+            .chain(unconfirmed_hashes)
```
MathieuDutSik, line 1096: The way the code is written as `confirmed_hashes.into_iter().chain(unconfirmed_hashes)` seems to imply that we have first the confirmed and then the unconfirmed hashes.

But for the data to be structured that way, we should have the heights being ordered. That might be so in the use case considered. But then I would recommend renaming the input as `ordered_heights`.

linera-core/src/updater.rs
```diff
@@ -556,12 +556,14 @@ where
                 Ok(info) => info,
             };
             // Obtain the missing blocks and the manager state from the local node.
-            let range = info.next_block_height..target_block_height;
+            let heights = (info.next_block_height.0..target_block_height.0)
+                .map(BlockHeight)
+                .collect::<Vec<_>>();
```
ma2bd: Do we need to collect?

linera-chain/src/chain.rs
afck: In all call sites, the heights _are_ ordered, and the hashes will be in the same order as the input heights. But I believe we are not relying on that fact anywhere, so maybe I should just document that otherwise the output will not be in the same order?

linera-core/src/updater.rs
afck: No, good catch! Removed.

linera-chain/src/chain.rs
MathieuDutSik: Yes, I have no objection to the code assuming the assumptions are right, but it seems that what is expected from it is not clear:
* That the vector of CryptoHash could be in the same order as the heights should be mentioned.
* If the heights are actually ordered on input, then maybe it should be mentioned. Not sure though.
* If the ordering of the cryptohashes is not the same as the heights, then I guess `block_hashes` is fine as a name.

afck: OK, I renamed it back and explained the ordering in the comment.

linera-chain/src/chain.rs
afck: Done.

PR #4902, opened 2025-11-04
Crash exporter if block processor task crashes

ndr-ds:
## Motivation

Forward port of #4900:
Right now, if the block processor returns an error, since we await the service first, we'll never see that error unless the service receives some shutdown signal or itself finishes for some reason, which in most cases doesn't happen.
If the block processor returns an error, we should see that error and propagate it, even if the service is still running and receiving notifications.

## Proposal

poll both futures at once with `select!` so that if the block processor returns an error, we can panic as we should

## Test Plan

Deploy on `conway-test` then on `testnet-conway`

## Release Plan

- These changes are already backported to the latest `testnet` branch

### DIFF
--- a/linera-service/src/exporter/main.rs
+++ b/linera-service/src/exporter/main.rs
@@ -116,10 +116,17 @@ impl Runnable for ExporterContext {
         );
 
         let service = ExporterService::new(sender);
-        service
-            .run(shutdown_notifier, self.config.service_config.port)
-            .await?;
-        handle.join().unwrap()
+
+        let mut block_processor_task = tokio::task::spawn_blocking(move || handle.join().unwrap());
+        tokio::select! {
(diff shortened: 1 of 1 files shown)


ndr-ds:
* **#4902** <a href="https://app.graphite.com/github/pr/linera-io/linera-protocol/4902?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/linera-io/linera-protocol/4902?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ndr-ds:
This actually made the exporters be able to make progress, so I think it's a net positive change we should merge into `main` as well

PR #4903, opened 2025-11-04
Retry certificate reads for up to a second before erroring.

deuszx:
## Motivation

In Testnet deployment we see that block processor fails to find the certificate in the DB. 

## Proposal

Scylla is eventually consistent database and maybe processor tries to read the cert before Scylla fully commits.

Retry certificate read for up to a second since the first try/failure.

## Test Plan

CI/manual.

## Release Plan

- These changes should be released to testnet deployment of exporter, and
- If successful backported to `main`.
- 
## Links

### DIFF
--- a/linera-service/src/exporter/runloops/block_processor/mod.rs
+++ b/linera-service/src/exporter/runloops/block_processor/mod.rs
@@ -31,7 +31,7 @@ where
     committee_destination_update: bool,
     // Temporary solution.
     // Tracks certificates that failed to be read from storage
-    // along with the time of the failure to avoid retrying too often.
+    // along with the time of the failure to avoid retrying for too long.
     retried_certs: HashMap<CryptoHash, (u8, Instant)>,
 }
 


linera-service/src/exporter/runloops/block_processor/mod.rs
```diff
@@ -120,6 +127,30 @@ where
...
                         },
 
+                        Err(ExporterError::ReadCertificateError(hash)) => {
+                            match self.retried_certs.remove(&hash) {
+                                // We retry only if the time elapsed since the first attempt is
+                                // less than 1 second. The assumption is that Scylla cannot
+                                // be eventually consistent for too long.
+                                Some((retries, first_attempt)) => {
+                                    let elapsed = Instant::now().duration_since(first_attempt);
+                                    if elapsed < Duration::from_secs(1) {
```
afck: I'd rather make this depend on the number of retries than the duration? Because what if the queue is so long that it takes more than one second to go through it: Then already at the first would-be retry, we'd give up?

deuszx: Yes. On the other hand – if we go through retries too fast we might do 10 tries in 100ms and still fail to read it.

ndr-ds: I just DM'ed you saying exactly this @deuszx haha
I would say exponential backoff, if at least the period of the backoff passed, try again, up to X times

deuszx: There's currently no mechanism to delay block/notificatin processing. This place here that I'm modifying isn't a spawned task but a global scheduler – if I sleep here it will affect processing of ALL notifications.

afck: But we _are_ inside one huge `select` statement. We could put the retry entries aside somewhere and e.g. only add them back into the queue above in the `tick` branch.

deuszx: This is supposed to be a temporary code, I'd prefer not to complicate it too much. The root issue is Scylla/DB - not the processor.

afck: Then I'd still make sure we retry at least once or twice?
Maybe:
```suggestion
                                    if retries < 3 || elapsed < Duration::from_secs(1) {
```

ndr-ds: If this is a race condition, retrying too fast kind of defeats the purpose, no? That's why I thought retrying with an exponential back-off would be ideal. But that would not be a simple change like this, as we would likely have to change what we store in the `new_block_queue` maybe

linera-service/src/exporter/runloops/block_processor/mod.rs
```diff
@@ -120,6 +127,30 @@ where
                             self.new_block_queue.push_back(next_block_notification);
                         },
 
+                        Err(ExporterError::ReadCertificateError(hash)) => {
+                            match self.retried_certs.remove(&hash) {
+                                // We retry only if the time elapsed since the first attempt is
+                                // less than 1 second. The assumption is that Scylla cannot
+                                // be eventually consistent for too long.
```
ndr-ds:
```suggestion
                                // be inconsistent for too long.
```

linera-service/src/exporter/runloops/block_processor/mod.rs
```diff
@@ -27,6 +29,10 @@ where
     storage: BlockProcessorStorage<T>,
     new_block_queue: NewBlockQueue,
     committee_destination_update: bool,
+    // Temporary solution.
+    // Tracks certificates that failed to be read from storage
+    // along with the time of the failure to avoid retrying too often.
```
ndr-ds: We don't do anything to avoid retrying to often, do we?
```suggestion
    // along with the time of the failure to avoid retrying for too long.
```

linera-service/src/exporter/runloops/block_processor/mod.rs
deuszx: Truth is, if we can't find certificate at height N (due to Scylla's (in)consistency) we won't find N+1 either (we shouldn't at least) so, in theory, sleeping here and retrying shouldn't affect the final goal. It would mean that we're changing the behavior of the processor's queue though and I'd like to avoid that if possible.

PR #4907, opened 2025-11-04
Revert #4642; show "admin"/"default" in `wallet show`.

afck:
## Motivation

#4642 was controversial, but we want to make it easy to see in `wallet show` which one is the admin chain.

## Proposal

Print "admin" and/or "default" in the chain ID cell in `wallet show` (implemented by Claude):

image

## Test Plan

I tried it locally. (See screenshot.)

## Release Plan

    - be released in a new SDK.

## Links

- Reverting #4642.


ma2bd:
How about this: no columns (maybe just separating lines `-----`), every chain is printed as a record of several "KEY: VALUE" entries, DEFAULT and ADMIN are displayed under the key "tags" ?

afck:
Closing in favor of #4927.

PR #4908, opened 2025-11-04
Add gRPC keepalive configuration to detect stale connections

ndr-ds:
## Motivation

Seems that `connect_lazy` doesn't reconnect automatically unless we specify a keep alive https://github.com/hyperium/tonic/issues/1254
There's some evidence we might be seeing that on the testnet on the proxy, as the user from the issue is seeing the same errors we're seeing.

## Proposal

Add keep alives so that they can expire and `connect_lazy` can reconnect automatically.

## Test Plan

We can reproduce this on the testnet. We can deploy to `conway-test` just to make sure it doesn't break anything (run e2e tests, for example), and then deploy to `testnet-conway`.

### DIFF
--- a/linera-rpc/src/grpc/transport.rs
+++ b/linera-rpc/src/grpc/transport.rs
@@ -34,7 +34,11 @@ cfg_if::cfg_if! {
             options: &Options,
         ) -> Result<Channel, Error> {
             let mut endpoint = tonic::transport::Endpoint::from_shared(address)?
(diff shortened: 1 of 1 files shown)


eldios (approved):

eldios:
shouldn't we port this to `main` and then backport? 🤔 we're having the same issues there right?

ma2bd:
@eldios It makes sense to me to test and fix the testnet first

PR #4910, opened 2025-11-05
[testnet] Sync if we learn about a new round while requesting a timeout.

### LINKED ISSUE #4909: Client should update itself on "wrong round" errors.
There are cases where the client requests a timeout certificate, and the validators respond with an error because they are already in a _later_ round. In that case, the client should download the proof for the later round, but it looks like it doesn't.

afck:
## Motivation

If a client requests a timeout certificate and the validators return a `WrongRound` error with a _higher_ round, the client currently fails to update its local node about the new round, and thus cannot propose a block.

## Proposal

Synchronize the chain state if we get a `WrongRound` error with a round higher than the one we see locally. (Almost all done by Claude!)

## Test Plan

A regression test was added.

## Release Plan

- These changes should be ported to `main`.

## Links

- Fixes #4909.

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -3158,58 +3158,37 @@ impl<Env: Environment> ChainClient<Env> {
     /// Requests a leader timeout certificate if the current round has timed out. Returns the
     /// chain info for the (possibly new) current round.
     async fn request_leader_timeout_if_needed(&self) -> Result<Box<ChainInfo>, ChainClientError> {
-        let mut info = self.chain_info_with_manager_values().await?;
-        // If the current round has timed out, we request a timeout certificate and retry in
-        // the next round.
-        if let Some(round_timeout) = info.manager.round_timeout {
-            if round_timeout <= self.storage_client().clock().current_time() {
-                match self.request_leader_timeout().await {
-                    Ok(_) => {
-                        info = self.chain_info_with_manager_values().await?;
-                    }
-                    Err(ChainClientError::CommunicationError(ref comm_error)) => {
-                        // Check if validators are in a different (higher) round.
-                        let validator_round = match comm_error {
-                            CommunicationError::Trusted(NodeError::WrongRound(round)) => {
-                                Some(*round)
-                            }
-                            CommunicationError::Sample(errors) => {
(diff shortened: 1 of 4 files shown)


linera-core/src/client/mod.rs
```diff
@@ -3158,19 +3158,37 @@ impl<Env: Environment> ChainClient<Env> {
...
+        let Some(round_timeout) = info.manager.round_timeout else {
+            return Ok(info); // The current round does not time out.
+        };
+        if round_timeout > self.storage_client().clock().current_time() {
+            return Ok(info); // The current round has not timed out yet.
+        };
+        match self.request_leader_timeout().await {
+            Ok(_) => {}
+            Err(ChainClientError::CommunicationError(comm_error))
+                if comm_error.errors().any(|err| {
```
deuszx, line 3171: Wow, why is it part of `communication error`? Can we make it more direct/explicit for `main`?

afck: I don't understand; you mean you'd like to not introduce `CommunicationError::errors`?

afck: One alternative I considered was to do it inside the updater: I.e. as soon as we learn from _a specific_ validator that they claim to be in a later round, we ask _only that validator_ for the proof. It's more added code, but maybe it's better?

deuszx: Yes, I like that approach better b/c we have the `WrongRound` (IIRC) error given to us directly.

afck: I think I have to replace the `LocalNodeClient` in the `ValidatorUpdater` with a `Client` first, so I can call `synchronize_chain_state_from`. I'll work on that on `main`, but it might be a large change…

bart-linera: It sounds like a better approach to me, too, but I'm fine with the one in this PR as well, if the other one turns out to be too complex.

afck: Here's the alternative solution (to `main`): https://github.com/linera-io/linera-protocol/pull/4912

afck:
Closing in favor of https://github.com/linera-io/linera-protocol/pull/4914.

PR #4913, opened 2025-11-05
When first request fails, start subsequent ones in parallel with increasing delay.

deuszx:
## Motivation

We observed that when the first request failed, the failure was broadcasted to all waiting peers. This slowed down the process and didn't use alternative peers we expect to have the data.

## Proposal

Detect if request fails and if so, try all alternative peers before erroring. We spawn a retry operation for every alternative peer with ever-increasing delay, delayed by `75ms` by default.

## Test Plan

CI (a test was added for this case).

## Release Plan

    - be released in a new SDK,

## Links

### DIFF
--- a/CLI.md
+++ b/CLI.md
@@ -203,7 +203,7 @@ Client implementation and command-line tool for the Linera blockchain
   Default value: `0.1`
 * `--alternative-peers-retry-delay-ms <ALTERNATIVE_PEERS_RETRY_DELAY_MS>` — Delay in milliseconds between starting requests to different peers. This helps to stagger requests and avoid overwhelming the network
 
-  Default value: `75`
+  Default value: `150`
 * `--storage <STORAGE_CONFIG>` — Storage configuration for the blockchain history
 * `--storage-max-concurrent-queries <STORAGE_MAX_CONCURRENT_QUERIES>` — The maximal number of simultaneous queries to the database
 * `--storage-max-stream-queries <STORAGE_MAX_STREAM_QUERIES>` — The maximal number of simultaneous stream queries to the database
--- a/linera-core/src/client/requests_scheduler/mod.rs
+++ b/linera-core/src/client/requests_scheduler/mod.rs
@@ -21,7 +21,7 @@ pub const CACHE_TTL_MS: u64 = 2000;
 pub const CACHE_MAX_SIZE: usize = 1000;
 pub const MAX_REQUEST_TTL_MS: u64 = 200;
 pub const ALPHA_SMOOTHING_FACTOR: f64 = 0.1;
-pub const STAGGERED_DELAY_MS: u64 = 75;
+pub const STAGGERED_DELAY_MS: u64 = 150;
 
 /// Configuration for the `RequestsScheduler`.
 #[derive(Debug, Clone)]


linera-core/src/client/requests_scheduler/scheduler.rs
```diff
@@ -646,9 +656,42 @@ impl<Env: Environment> RequestsScheduler<Env> {
...
+        // Remove the peer we're about to use from alternatives (it shouldn't retry with itself)
+        self.in_flight_tracker
+            .remove_alternative_peer(&key, &peer)
+            .await;
+
         // Execute the actual request
-        tracing::trace!(key = ?key, "executing new request");
-        let result = operation(peer).await;
+        tracing::trace!(key = ?key, peer = ?peer, "executing new request");
+        let result = operation(peer.clone()).await;
```
afck, line 666: Maybe for a future PR, but: would it make sense to also set a timeout here after which we retry with another peer, even if this operation hasn't returned an error yet?

linera-core/src/client/requests_scheduler/scheduler.rs
```diff
@@ -665,6 +708,84 @@ impl<Env: Environment> RequestsScheduler<Env> {
...
+                }
+            }
+        }
+
+        // All attempts failed
+        tracing::debug!(
+            key = ?key,
+            "all staggered parallel retry attempts failed"
+        );
+        Err(last_error.unwrap_or(NodeError::UnexpectedMessage))
```
afck, line 786: (This `unwrap_or` error case is actually unreachable, is it?)

linera-core/src/client/requests_scheduler/scheduler.rs
```diff
@@ -665,6 +708,84 @@ impl<Env: Environment> RequestsScheduler<Env> {
...
+
+        while let Some((peer_idx, result)) = futures.next().await {
+            match result {
+                Ok(value) => {
+                    tracing::info!(
+                        key = ?key,
+                        peer_index = peer_idx,
+                        "staggered parallel retry succeeded"
+                    );
+                    return Ok(value);
```
afck, line 767: (We need to be sure that those futures are all cancel-safe. They probably are?)

linera-core/src/client/requests_scheduler/scheduler.rs
```diff
@@ -665,6 +708,84 @@ impl<Env: Environment> RequestsScheduler<Env> {
...
+        Fut: Future<Output = Result<T, NodeError>>,
+    {
+        use futures::stream::{FuturesUnordered, StreamExt};
+        use linera_base::time::{timer::sleep, Duration};
+
+        let mut futures = FuturesUnordered::new();
+
+        // Create futures with staggered delays
+        for (idx, peer) in peers.into_iter().enumerate() {
+            let delay = Duration::from_millis(staggered_delay_ms * idx as u64);
```
afck, line 743: I wonder if the duration should increase quadratically: The more peers we are already trying the more likely it is that the request is simply a very slow one, and it's not the peers' fault.

linera-core/src/client/requests_scheduler/scheduler.rs
```diff
@@ -1105,4 +1303,137 @@ mod tests {
...
+                staggered_delay_ms,
+                delay
+            );
+        }
+
+        // Total time should be significantly less than sequential.
+        // With staggered parallel: node0 fails immediately, node1 starts at 10ms (and takes 20ms),
+        // node2 starts at 20ms and succeeds at 25ms total
+        let total_time = Instant::now().duration_since(start_time).as_millis();
+        assert!(total_time >= staggered_delay_ms && total_time < 50,);
```
afck, line 1437: (I hope that test doesn't turn flaky. But I guess it's not feasible to use the fake clock here?)

linera-core/src/client/requests_scheduler/scheduler.rs
deuszx: I guess we could put _all_ alternative peers under the `staggered*` function right away 🤔 a bit like we do in the test.

linera-core/src/client/requests_scheduler/scheduler.rs
deuszx: hopefully, yes. I can rewrite the code to get rid of this arm.

linera-core/src/client/requests_scheduler/scheduler.rs
deuszx: They should all be requests to download some data. Even if we don't cancel them, the result is data being dropped at worst.

linera-core/src/client/requests_scheduler/scheduler.rs
deuszx: I haven't tried. I can ticket it for the future task.

deuszx: https://github.com/linera-io/linera-protocol/issues/4931

linera-core/src/client/requests_scheduler/scheduler.rs
deuszx: Actually, I'm gonna ticket this as there are some edge cases that I think will need more thought.

deuszx: Anyway, done here [Set timeout for the first request](https://github.com/linera-io/linera-protocol/pull/4913/commits/5704fbf6c81d1ef670bfdee5cf08db0770414721)

linera-core/src/client/requests_scheduler/scheduler.rs
deuszx: Done in [Set timeout for the first request](https://github.com/linera-io/linera-protocol/pull/4913/commits/5704fbf6c81d1ef670bfdee5cf08db0770414721)

linera-core/src/client/requests_scheduler/scheduler.rs
deuszx: Ah, it conflicts with the other timeout feature (which starts a new request if the currently in flight one is delayed). I'm gonna revert this for now and do it in a separate PR.

linera-core/src/client/requests_scheduler/scheduler.rs
```diff
@@ -663,7 +666,15 @@ impl<Env: Environment> RequestsScheduler<Env> {
 
         // Execute the actual request
         tracing::trace!(key = ?key, peer = ?peer, "executing new request");
-        let result = operation(peer.clone()).await;
+        let result = futures::select! {
+            _timeout = linera_base::time::timer::sleep(self.retry_delay).fuse() => {
+                tracing::trace!(key = ?key, "retry delay elapsed, proceeding with request");
+                Err(NodeError::WorkerError { error: "timeout".to_string() }) // Placeholder error to trigger retries
```
afck: But we don't want to cancel the ongoing operation in that case? We just want to add it to the unordered futures, basically?

deuszx: Yes, maybe you're right. I didn't add it to the method b/c it changes the other rule: `peer` node is the first one we request data from. To maintain both I'd have to rewrite code quite a lot.

This is reverted now anyway.

PR #4916, opened 2025-11-05
Add the `root_key` to the `MissingEntries` error.

MathieuDutSik:
## Motivation

We have some error of `MissingEntries` occurring in the logs. We do not know what is the relevant root key.
So, we add it here.

## Proposal

The following has been done:
* The `fn root_key(&self) -> Result<Vec<u8>, Self::Error>` is added to the trait definition. We need a `Result` because of the deserialization of the keys in RocksDB and StorageService.
* For RocksDB, ScyllaDB, DynamoDB, and StorageService, this did not add any new data.
* The key is not added to the error since this would force some clone operation.

## Test Plan

The CI.

## Release Plan

The change could be backported to TestNet Conway.

## Links

None.

### DIFF
--- a/linera-views/src/backends/journaling.rs
+++ b/linera-views/src/backends/journaling.rs
@@ -123,7 +123,6 @@ where
         self.store.max_stream_queries()
     }
 
-    /// The `root_key` does not change
     fn root_key(&self) -> Result<Vec<u8>, Self::Error> {
         self.store.root_key()
     }
--- a/linera-views/src/backends/memory.rs
+++ b/linera-views/src/backends/memory.rs
@@ -112,7 +112,7 @@ static MEMORY_DATABASES: LazyLock<Mutex<MemoryDatabases>> =
 pub struct MemoryStore {
     /// The map used for storing the data.
     map: Arc<RwLock<MemoryStoreMap>>,
-    /// The `root_key`
+    /// The root key.
     root_key: Vec<u8>,
     /// The maximum number of queries used for a stream.
     max_stream_queries: usize,
--- a/linera-views/src/error.rs
+++ b/linera-views/src/error.rs
@@ -1,6 +1,8 @@
 // Copyright (c) Zefchain Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
+use linera_base::hex;
+
 /// Main error type for the crate.
 #[derive(thiserror::Error, Debug)]
 pub enum ViewError {
@@ -50,7 +52,7 @@ pub enum ViewError {
     InconsistentEntries,
 
     /// The database is corrupt: Some entries are missing
-    #[error("missing database entries")]
+    #[error("missing database entrie for the root key {}", hex::encode(.0))]
     MissingEntries(Vec<u8>),
 
     /// The values are incoherent.
--- a/linera-views/src/store.rs
+++ b/linera-views/src/store.rs
@@ -47,7 +47,7 @@ pub trait ReadableKeyValueStore: WithError {
     /// Retrieve the number of stream queries.
     fn max_stream_queries(&self) -> usize;
 
-    /// Gets the root_key of the store.
+    /// Gets the root key of the store.
     fn root_key(&self) -> Result<Vec<u8>, Self::Error>;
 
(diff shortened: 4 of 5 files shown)


linera-views/src/views/queue_view.rs
```diff
@@ -306,7 +306,8 @@ where
         for entry in self.context.store().read_multi_values(keys).await? {
             match entry {
                 None => {
-                    return Err(ViewError::MissingEntries);
+                    let root_key = self.context.store().root_key()?;
+                    return Err(ViewError::MissingEntries(root_key));
```
afck, line 310: Should we add the (non-root) entry key, too?

linera-views/src/backends/journaling.rs
```diff
@@ -123,6 +123,11 @@ where
         self.store.max_stream_queries()
     }
 
+    /// The `root_key` does not change
```
afck:
```suggestion
    /// Returns the view's root key?
```

linera-views/src/backends/memory.rs
```diff
@@ -111,6 +112,8 @@ static MEMORY_DATABASES: LazyLock<Mutex<MemoryDatabases>> =
 pub struct MemoryStore {
     /// The map used for storing the data.
     map: Arc<RwLock<MemoryStoreMap>>,
+    /// The `root_key`
```
afck:
```suggestion
    /// The root key
```

(Either way not a very useful comment, though.

linera-views/src/error.rs
```diff
@@ -51,7 +51,7 @@ pub enum ViewError {
 
     /// The database is corrupt: Some entries are missing
     #[error("missing database entries")]
```
afck: Let's add the root key to the message string, too. (Maybe in hexadecimal?)

linera-views/src/store.rs
```diff
@@ -47,6 +47,9 @@ pub trait ReadableKeyValueStore: WithError {
     /// Retrieve the number of stream queries.
     fn max_stream_queries(&self) -> usize;
 
+    /// Gets the root_key of the store.
```
afck:
```suggestion
    /// Gets the root key of the store.
```

linera-views/src/views/queue_view.rs
MathieuDutSik: We can, but with the existing code, this forces a clone.

Can we avoid that clone? Yes, we are currently taking a `Vec<Vec<u8>>`. I think we can replace by a `&[Vec<u8>]`.

If you really want the key, I can do the clone and put it.

afck: The clone would only happen in the error case anyway, would it?
(Not a blocker from my side! Not sure how useful it would be for debugging.)

linera-views/src/backends/journaling.rs
MathieuDutSik: I prefer to remove it altogether.

linera-views/src/backends/memory.rs
MathieuDutSik: Yes, but it has to be documented.

linera-views/src/store.rs
MathieuDutSik: Ok, thanks.

linera-views/src/error.rs
MathieuDutSik: Ok, that is the key point indeed.
Rewritten as
```rust
    #[error("missing database entrie for the root key {}", hex::encode(.0))]
    MissingEntries(Vec<u8>),
```

Hope that is ok.

linera-views/src/views/queue_view.rs
MathieuDutSik: Not so easily, unfortunately.

linera-views/src/error.rs
```diff
@@ -50,8 +52,8 @@ pub enum ViewError {
     InconsistentEntries,
 
     /// The database is corrupt: Some entries are missing
-    #[error("missing database entries")]
-    MissingEntries,
+    #[error("missing database entrie for the root key {}", hex::encode(.0))]
```
ma2bd, line 55: @MathieuDutSik @afck typo: "database entrie"

PR #4920, opened 2025-11-06
Improve a few log messages: timeouts, validator addresses.

afck:
## Motivation

Some of our logs are less useful than they could be:
* Many show a validator's public key instead of address.
* One prints the `Debug` format of a `RoundTimeout`, with the time as a number of microseconds since the epoch.
* We mix structured logging and printing values inline in the log message.

## Proposal

* Print the validator address.
* `impl Display for RoundTimeout`, and show the formatted date and time.
* Prefer structured logging.

## Test Plan

(Only minor logging changes.)

## Release Plan

    - be released in a new SDK, (not urgent)
    - be released in a validator hotfix. (not urgent)

## Links

### DIFF
--- a/linera-client/src/chain_listener.rs
+++ b/linera-client/src/chain_listener.rs
@@ -594,14 +594,14 @@ impl<C: ClientContext + 'static> ChainListener<C> {
             Ok((certs, None)) => info!(
                 %chain_id,
                 created_block_count = %certs.len(),
-                "Done processing inbox.",
+                "done processing inbox",
             ),
             Ok((certs, Some(new_timeout))) => {
                 info!(
                     %chain_id,
                     created_block_count = %certs.len(),
                     timeout = %new_timeout,
-                    "Waiting for round timeout before continuing to process the inbox.",
+                    "waiting for round timeout before continuing to process the inbox",
                 );
                 listening_client.timeout = new_timeout.timestamp;
             }
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -437,9 +437,10 @@ impl<Env: Environment> Client<Env> {
                 .download_certificates_from(&remote_node, chain_id, target_next_block_height)
                 .await
             {
-                Err(err) => info!(
-                    "Failed to download certificates from validator {:?}: {err}",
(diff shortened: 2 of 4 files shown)


bart-linera (approved):
Simple but very helpful :clap:

linera-client/src/chain_listener.rs
```diff
@@ -598,7 +598,7 @@ impl<C: ClientContext + 'static> ChainListener<C> {
             Ok((certs, Some(new_timeout))) => {
                 info!(
                     "{} blocks created on chain {chain_id}. Will try processing the inbox later \
-                    based on the round timeout: {new_timeout:?}",
+                    based on the round timeout: {new_timeout}",
```
deuszx: While we're at it: can you rewrite this log to:
```rust
info!(
  ?chain_id,
  block_count=?certs.len(),
  ?new_timeout,
  "new blocks created. delay processing inbox to later round timeout"
);
```
?

afck: Or maybe without "new blocks created". (The block count might be 0.)

deuszx: That's fine. I just would like for us to move ALL logs to this approach. When we start grepping through logs it will be easier of we can look for the whole string (rn we can't b/c of the interpolation).

afck: How do you feel about https://github.com/linera-io/linera-protocol/pull/4920/commits/d2fcfc9c3cc0002c3c821ac534dd0971c1dedb1e?

deuszx: Looks fine. IIRC we should start with a lowercase and _not_ end the log with a period? but that's probably something up for discussion.

afck: Yes, we should agree on one or the other. I think we have lots of logs with uppercase and periods, so let's have the discussion first before we change that.

afck: After some research, I think I agree with you (at least about the period). Updated in https://github.com/linera-io/linera-protocol/pull/4920/commits/6464d60044e45a2065be4597d729cccc0411cfdf.

PR #4926, opened 2025-11-06
[testnet] Backport the "fn root_key" and the MissingEntries change (#4916)

MathieuDutSik:
Backport https://github.com/linera-io/linera-protocol/pull/4916.

## Motivation

The "fn root_key(&self)" function and the "MissingEntries(Vec<u8>)" are of interest to TestNet usage.

## Proposal

Backport the change.
It is not completely straightforward, since the handling of `root_key` has changed in main for RocksDB and StorageService (see https://github.com/linera-io/linera-protocol/pull/4859 and https://github.com/linera-io/linera-protocol/pull/4858).

## Test Plan

The same test function has been added.

## Release Plan

To TestNet Conway.

## Links

None.

### DIFF
--- a/linera-views/src/error.rs
+++ b/linera-views/src/error.rs
@@ -1,6 +1,8 @@
 // Copyright (c) Zefchain Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
+use linera_base::hex;
+
 /// Main error type for the crate.
 #[derive(thiserror::Error, Debug)]
 pub enum ViewError {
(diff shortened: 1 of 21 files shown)


linera-views/src/error.rs
```diff
@@ -54,8 +56,8 @@ pub enum ViewError {
     InconsistentEntries,
 
     /// The database is corrupt: Some entries are missing
-    #[error("missing database entries")]
-    MissingEntries,
+    #[error("missing database entrie for the root key {}", hex::encode(.0))]
```
afck:
```suggestion
    #[error("missing database entries for the root key {}", hex::encode(.0))]
```

afck (approved):

PR #4929, opened 2025-11-07
Don't pass empty string, it's optional

deuszx:
## Motivation

We are seeing errors in the console, coming from the Linera web client, about CryptoHash expecting 32-bytes but getting 0.

## Proposal

This happens b/c frontends pass empty string (`''`) to a method (`query`) that expects `Option<CryptoHash>` and assuming that if it's `Some(_)` it must be correct slice.

After fixing `query` in #4904 I forgot to update this.

## Test Plan

Manual. Verified that with this fix we no longer get the errors.

## Release Plan

- Release web demos to the frontend
- Backport fix to `main`

## Links

### DIFF
--- a/examples/native-fungible/index.html
+++ b/examples/native-fungible/index.html
@@ -166,7 +166,7 @@
       }
 
       async function updateBalance(application, owner, blockHash) {
-          const response = JSON.parse(await application.query(gql(`query { tickerSymbol, accounts { entry(key: "${owner}") { value } } }`), blockHash || ''));
+          const response = JSON.parse(await application.query(gql(`query { tickerSymbol, accounts { entry(key: "${owner}") { value } } }`)));
           console.debug('application response:', response);
           document.querySelector('#ticker-symbol').textContent = response.data.tickerSymbol;
           document.querySelector('#balance').textContent = (+(response?.data?.accounts?.entry.value || 0)).toFixed(2);
--- a/examples/counter/index.html
+++ b/examples/counter/index.html
@@ -85,7 +85,7 @@
           }
 
           async function updateCount(block_hash) {
-              const response = await counter.query('{ "query": "query { value }" }', block_hash || '');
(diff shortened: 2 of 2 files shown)


examples/native-fungible/index.html
```diff
@@ -166,7 +166,7 @@ <h2>Chain history for <code id="chain-id" class="hex">requesting a new microchai
       }
 
       async function updateBalance(application, owner, blockHash) {
-          const response = JSON.parse(await application.query(gql(`query { tickerSymbol, accounts { entry(key: "${owner}") { value } } }`), blockHash || ''));
+          const response = JSON.parse(await application.query(gql(`query { tickerSymbol, accounts { entry(key: "${owner}") { value } } }`)));
```
bart-linera, line 169: Did you mean to remove `blockHash` here completely? If yes, maybe it should no longer be an argument to the function, as it doesn't seem to be used anywhere else? (But I suspect you meant the same thing as above: `blockHash` instead of `blockHash || ''`?)

deuszx: It's still used (see line 248) it's just sometimes it's `null` which will translate to `None` when crossing TS->Wasm boundary.

bart-linera: But not within `function updateBalance`, which ends on line 173 (and it's still an argument to that function)? Or am I misunderstanding something?

deuszx: Ah, you're right 😬 I fixed `counter` and here removed too much. Let me push the update where I _do_ use it here.

bart-linera: (In case it's not clear: I'm pointing out that after the change, `blockHash` is not being passed to `application.query` in `updateBalance` at all. So even if it's not `null`, `application.query` will get a `null`.)

(Ninja'd :sweat_smile: )

deuszx: Pushed directly to testnet_conway 🙈

PR #4934, opened 2025-11-08
Change the API of "contains_keys" and "read_multi_values_bytes" to take a reference.

MathieuDutSik:
## Motivation

The functions `contains_keys` and `read_multi_values_bytes` take a `Vec<Vec<u8>>` which is in opposition with other functions that take just a reference.

## Proposal

Do the change in a straightforward way.

## Test Plan

The CI.

## Release Plan

## Links

None.


MathieuDutSik:
> Does this potentially break contracts in a backwards-incompatible way? i.e. if a contract uses `contains_keys` (or `read_multi_values_bytes`).

No.
Because the `contains_keys` and others API are defined in `BaseRuntime`, `ContractRuntime`.

ma2bd (approved):

PR #4935, opened 2025-11-08
Restructure the CI to diminish load.

MathieuDutSik:
## Motivation

We have high CI expenses that should be diminished as much as possible.

## Proposal

The following steps are done:
* Move the Ethereum tests to a separate file that is triggered only on `linera-execution`, `linera-ethereum` and other relevant directory. Just like for DynamoDB, ScyllaDB.
* Harmonize the dynamodb.yml, scylladb.yml
* Remove a `DOCKER_COMPOSE_WAIT` that is wrongly put.
* Put the `remote-net-test`, `remote-kubernetes-net-test`, and `benchmark-test` as merge group just like docker-compose and other 3 tests.
* Add the `concurrency` group to the `long_faucet_chain_test.yml`.

## Test Plan

The CI.

## Release Plan

Some of the changes can be backported to the TestNet Conway.

## Links

None.

### DIFF
--- a/.github/workflows/benchmark-test.yml
+++ b/.github/workflows/benchmark-test.yml
@@ -1,7 +1,11 @@
 name: Benchmark Test
 
 on:
+  push:
+    branches: [ 'devnet_*', 'testnet_*' ]
+  pull_request:
   merge_group:
+  workflow_dispatch:
 
 # This allows a subsequently queued workflow run to interrupt previous runs on pull requests
 concurrency:
@@ -30,6 +34,7 @@ permissions:
 
 jobs:
   benchmark-test:
+    if: ${{ github.event_name == 'merge_group' }}
     runs-on: ubuntu-latest
     timeout-minutes: 40
 
--- a/.github/workflows/remote-kubernetes-net-test.yml
+++ b/.github/workflows/remote-kubernetes-net-test.yml
@@ -1,7 +1,11 @@
 name: Remote Kubernetes Net Test
 
 on:
+  push:
+    branches: [ 'devnet_*', 'testnet_*' ]
+  pull_request:
   merge_group:
+  workflow_dispatch:
 
 # This allows a subsequently queued workflow run to interrupt previous runs on pull requests
 concurrency:
@@ -30,6 +34,7 @@ permissions:
 
 jobs:
   remote-kubernetes-net-test:
+    if: ${{ github.event_name == 'merge_group' }}
     runs-on: ubuntu-latest-16-cores
     timeout-minutes: 90
 
--- a/.github/workflows/remote-net-test.yml
+++ b/.github/workflows/remote-net-test.yml
@@ -1,7 +1,11 @@
 name: Remote Net Test
 
 on:
+  push:
+    branches: [ 'devnet_*', 'testnet_*' ]
+  pull_request:
   merge_group:
+  workflow_dispatch:
 
 # This allows a subsequently queued workflow run to interrupt previous runs on pull requests
 concurrency:
@@ -30,6 +34,7 @@ permissions:
 
 jobs:
   remote-net-test:
+    if: ${{ github.event_name == 'merge_group' }}
     runs-on: ubuntu-latest-8-cores
     timeout-minutes: 40
 


deuszx:
Looks good, I'd make `remote-net-test` runnable on workflow dispatch as well.

What time savings have you observed?

MathieuDutSik:
> Looks good, I'd make `remote-net-test` runnable on workflow dispatch as well.
> 
> What time savings have you observed?

The time savings are very visible: The tasks are not run.

deuszx:
I meant by how much it cuts the CI run on per commit basis. I found my answer.

deuszx:
> Looks good, I'd make `remote-net-test` runnable on workflow dispatch as well.

Waiting for this

MathieuDutSik:
> > Looks good, I'd make `remote-net-test` runnable on workflow dispatch as well.
> 
> Waiting for this

It is already the case.
```bash
  remote-net-test:
    if: ${{ github.event_name == 'merge_group' }}
```

deuszx:
Workflow dispatch means devs (with appropriate permissions) can spawn a CI job from a github UI, not that it runs on a merge group. These are two different triggers.

.github/workflows/rust.yml
```diff
@@ -71,8 +71,7 @@ jobs:
               - '!CONTRIBUTING.md'
               - '!INSTALL.md'
   remote-net-test:
-    needs: changed-files
-    if: needs.changed-files.outputs.should-run == 'true'
+    if: ${{ github.event_name == 'merge_group' }}
```
deuszx: If it runs only as part of the merge group, extract it to a separate file. This and other jobs in this file that are merge group only

.github/workflows/benchmark-test.yml
```diff
@@ -0,0 +1,59 @@
+name: Benchmark Test
+
+on:
+  push:
+    branches: [ 'devnet_*', 'testnet_*' ]
+  pull_request:
```
deuszx, line 6: This means it will run on every PR as well. I thought we want to get rid of it?

MathieuDutSik: Currently, we run on every PR. But also for every single commit.
With the change, we run it only when merging into main. That is the difference.

deuszx: `on: pull_request` means it will run on every commit on an already-open PR: https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request

> Runs your workflow when activity on a pull request in the workflow's repository occurs. For example, if no activity types are specified, the workflow runs when a pull request is opened or reopened or when the head branch of the pull request is updated.

MathieuDutSik: This is clearly not true, since we can see in this PR:
```bach
1 in progress, 7 skipped, 26 successful checks
```

7 skipped is exactly what we want from this PR: 3 more skipped checks.

deuszx: Having both `pull_request` and `merge_group` is a waste as well. We merge to `main` only via merge group and `merge_group` (should) first merge `main` into the PR and then run the CI. This is the same as running the CI after merging to `main`.

MathieuDutSik: I disagree with what you say: "merge_group (should) first merge main into the PR and then run the CI"
We should first run the CI on those skipped tests and then merge (if the tests pass of course).

The code that I wrote is exactly the same as the one for the other entries like "long_faucet_chain_test.yml" and "lint-check-all-features.yml". So, there is no change of behavior here.

deuszx: When a PR is merged via merge_group it (github) will create a temporary branch, merge target branch into the PR branch, and THEN run the CI on it. This will run CI on the same state of a repo as CI ran on `main`.

This isn't what I think, this is how it works.

MathieuDutSik: Explain how the behavior of the created files is different from that of the existing files "lint-check-all-features.yml", "long_faucet_chain_test.yml", "test-readmes.yml", and "docker-compose.yml".

deuszx: Does it matter? Maybe the others are wrong and they need to be fixed.

MathieuDutSik: Well, we could merge the PR and then address the problem later.

I think you are the author of the original design. So, you are most qualified to do the correction.

deuszx: Are you going to address all of them then in a follow-up?

PR #4937, opened 2025-11-09
Add option to trigger network actions during linera query-validator(s)

### LINKED ISSUE #4936: query-validators --min-votes breaks the final count
2025-11-09T08:03:40.573556Z  INFO linera: 18/18 validators are OK.

ma2bd:
## Motivation

Needed for troubleshooting

fixes #4936 

## Proposal

Add option --create-network-actions

## Test Plan

CI

## Release Plan

    - be released in a new SDK,

### DIFF
--- a/linera-client/src/client_context.rs
+++ b/linera-client/src/client_context.rs
@@ -721,8 +721,10 @@ impl<Env: Environment, W: Persist<Target = Wallet>> ClientContext<Env, W> {
         address: &str,
         node: &impl ValidatorNode,
         chain_id: ChainId,
+        create_network_actions: bool,
     ) -> Result<ChainInfo, Error> {
-        let query = ChainInfoQuery::new(chain_id).with_manager_values();
+        let mut query = ChainInfoQuery::new(chain_id).with_manager_values();
+        query.create_network_actions = create_network_actions;
         match node.handle_chain_info_query(query).await {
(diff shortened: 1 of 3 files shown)


afck (approved):
(Good catch with the OK count!)

linera-client/src/client_context.rs
```diff
@@ -149,7 +149,7 @@ impl ValidatorQueryResults {
             reference.and_then(|ref_results| ref_results.genesis_config_hash.as_ref().ok());
         match &self.genesis_config_hash {
             Ok(hash) if ref_genesis_hash.is_some_and(|ref_hash| ref_hash == hash) => {}
-            Ok(hash) => println!("Genesis config hash: {hash}"),
+            Ok(hash) => println!("Genesis config hash differs: {hash}"),
```
ma2bd, line 152: We're apparently parsing this text but that's not even the value it used to be any more

afck: Where? In CI? I can't find where we parse it.

ma2bd: I believe you wrote the line :)
https://github.com/linera-io/linera-protocol/blob/main/linera-service/src/cli_wrappers/wallet.rs#L515

ma2bd:
This code has been refactored since then

PR #4938, opened 2025-11-10
Group operations for `maybe_get_required_blobs`

MathieuDutSik:
## Motivation

The `pending_proposed_blobs` needs to be simplified. But before that, we can restructure the access code.

## Proposal

Group the operations in the querying of the data in `maybe_get_required_blobs`.

## Test Plan

The CI.

## Release Plan

## Links

None.

### DIFF
--- a/linera-core/src/chain_worker/state.rs
+++ b/linera-core/src/chain_worker/state.rs
@@ -376,15 +376,16 @@ where
 
         let (missing_indices, missing_blob_ids) = missing_indices_blob_ids(&maybe_blobs);
         if !missing_indices.is_empty() {
-            let pending_blobs = self
+            let all_entries_pending_blobs = self
                 .chain
                 .pending_proposed_blobs
                 .try_load_all_entries()
                 .await?;
             for (index, blob_id) in missing_indices.into_iter().zip(missing_blob_ids) {
-                for (_, pending_blob) in &pending_blobs {
-                    if let Some(blob) = pending_blob.get(&blob_id).await? {
+                for (_, pending_blobs) in &all_entries_pending_blobs {
+                    if let Some(blob) = pending_blobs.get(&blob_id).await? {
                         maybe_blobs[index].1 = Some(blob);
+                        break;
                     }
                 }
             }


linera-core/src/chain_worker/state.rs
```diff
@@ -341,43 +336,64 @@ where
...
-                    .chain
-                    .pending_proposed_blobs
-                    .try_load_all_entries()
-                    .await?
-                {
-                    if let Some(blob) = pending_blobs.get(blob_id).await? {
-                        *maybe_blob = Some(blob);
-                        break;
-                    }
+            for (_, pending_blobs) in self
```
afck: You probably want to `continue` here if `maybe_blob.is_some()`?

MathieuDutSik: Yes, you are very much correct.
I changed that part of the code significantly, though.

linera-core/src/chain_worker/state.rs
```diff
@@ -374,16 +374,18 @@ where
...
-                .await?
-            {
-                if let Some(blob) = pending_blobs.get(blob_id).await? {
-                    *maybe_blob = Some(blob);
-                    break;
+                .await?;
+            for (index, blob_id) in missing_indices.into_iter().zip(missing_blob_ids) {
+                for (_, pending_blob) in &pending_blobs {
+                    if let Some(blob) = pending_blob.get(&blob_id).await? {
+                        maybe_blobs[index].1 = Some(blob);
```
afck, line 387: And `break;`?

linera-core/src/chain_worker/state.rs
```diff
@@ -374,16 +374,18 @@ where
...
                 .pending_proposed_blobs
                 .try_load_all_entries()
-                .await?
-            {
-                if let Some(blob) = pending_blobs.get(blob_id).await? {
-                    *maybe_blob = Some(blob);
-                    break;
+                .await?;
+            for (index, blob_id) in missing_indices.into_iter().zip(missing_blob_ids) {
+                for (_, pending_blob) in &pending_blobs {
```
afck: (Strictly speaking, it's not a single `pending_blob`. It's still a collection of pending blobs that belong to a single block.)

linera-core/src/chain_worker/state.rs
MathieuDutSik: Yes. Sorry.

linera-core/src/chain_worker/state.rs
MathieuDutSik: Ok. renamed "pending_blobs" as "all_entries_pending_blobs" and "pending_blob" as "pending_blobs".

PR #4939, opened 2025-11-10
Add a notion of lease in journaling

ma2bd:
## Motivation

Prevent DB corruption

## Proposal

* Add a lease management API to DirectWritableKeyValueStore
* Use it in journaling.rs

TODO:
* tests
* add journaling to the storage-service?
* deduplicate code
* make the TTL configurable
* in scylladb we could also use `USING TTL`

## Test Plan

CI + TBD

## Release Plan

### DIFF
--- a/linera-views/src/backends/journaling.rs
+++ b/linera-views/src/backends/journaling.rs
@@ -19,15 +19,20 @@
 //! time the data in a block are written, the journal header is updated in the same
 //! transaction to mark the block as processed.
 
+use std::sync::{Arc, Mutex};
+
+use rand::Rng as _;
 use serde::{Deserialize, Serialize};
 use static_assertions as sa;
 use thiserror::Error;
+use tokio::{select, task::JoinHandle};
+use tokio_util::sync::CancellationToken;
 
 use crate::{
     batch::{Batch, BatchValueWriter, DeletePrefixExpander, SimplifiedBatch},
     store::{
-        DirectKeyValueStore, KeyValueDatabase, ReadableKeyValueStore, WithError,
-        WritableKeyValueStore,
+        DirectKeyValueStore, DirectWritableKeyValueStore, KeyValueDatabase, LeaseConflict,
+        ReadableKeyValueStore, WithError, WritableKeyValueStore,
     },
     views::MIN_VIEW_TAG,
 };
@@ -44,7 +49,25 @@ pub struct JournalingKeyValueStore<S> {
     /// The inner store.
     store: S,
     /// Whether we have exclusive R/W access to the keys under root key.
-    has_exclusive_access: bool,
+    /// All subviews in a root view should use the same lease.
+    has_exclusive_access: Option<Arc<Lease>>,
+}
+
+/// A journaling key-value store.
+pub struct Lease {
+    /// The background task that maintains the lease.
+    _task: JoinHandle<()>,
+    /// The cancellation token.
+    token: CancellationToken,
+    /// The latest conflict that we had, if any. This is a best-effort only used for error
+    /// reporting.
+    conflict: Arc<Mutex<Option<LeaseConflict>>>,
+}
+
+impl Drop for Lease {
+    fn drop(&mut self) {
+        self.token.cancel();
+    }
 }
 
 /// Data type indicating that the database is not consistent
@@ -56,6 +79,9 @@ pub enum JournalConsistencyError {
 
     #[error("Refusing to use the journal without exclusive database access to the root object.")]
     JournalRequiresExclusiveAccess,
+
+    #[error("Failed to acquire lease {0:?}")]
+    LeaseConflict(LeaseConflict),
 }
 
 /// The tag used for the journal stuff.
@@ -70,6 +96,8 @@ enum KeyTag {
     Journal = 1,
     /// Prefix for the block entry.
     Entry,
+    /// Prefix for the lease.
+    Lease,
 }
 
 fn get_journaling_key(tag: u8, pos: u32) -> Result<Vec<u8>, bcs::Error> {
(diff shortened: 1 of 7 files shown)


linera-views/src/backends/journaling.rs
```diff
@@ -178,16 +208,91 @@ where
...
+            loop {
+                select! {
+                        _ = cloned_token.cancelled() => {
+                            tracing::trace!("Discarding lease for UUID {uuid} and root key: {root_key_owned:?}");
+                            if let Err(err) = store
+                            .discard_lease(lease_key.clone(), uuid)
+                            .await
+                            {
+                                tracing::error!(
+                                "Failed to renew lease for UUID {uuid} and root key {root_key_owned:?}: {err}"
```
ma2bd, line 255: typo: discard lease

linera-views/src/backends/scylla_db.rs
```diff
@@ -710,6 +750,129 @@ impl DirectWritableKeyValueStore for ScyllaDbStoreInternal {
...
+
+        // Check if the LWT was applied
+        let rows = result.into_rows_result()?;
+        let mut rows = rows.rows::<(bool,)>()?;
+
+        if let Some(row) = rows.next() {
+            let (applied,) = row?;
+            if !applied {
+                // Delete failed - UUID doesn't match or row doesn't exist
+                tracing::warn!("Failed to discard lease");
```
ma2bd, line 869: TODO: better message

linera-views/src/backends/scylla_db.rs
```diff
@@ -710,6 +750,129 @@ impl DirectWritableKeyValueStore for ScyllaDbStoreInternal {
...
+            .execute_single_page(
+                &store.lease_obtain_lwt,
+                &values,
+                scylla::response::PagingState::start(),
+            )
+            .await?;
+
+        // Check if the LWT was applied
+        let rows = result.into_rows_result()?;
+        let mut rows = rows.rows::<(bool,)>()?;
```
ma2bd, line 800: not sure about this

MathieuDutSik (commented):
Comment:
* To make the examples compile under wasm32, the best is to make journaling, lru and the like not available in wasm32, where anyway they are not used.
* After those corrections, the DynamoDB end-to-end test fail randomly and the ScyllaDB test fail due to syntax error in the request.
* The `obtain_lease` is returning a `Result<Result<_,_>,_>`. Such constructions were previously proposed and refused for `linera-storage` in favor of returning a `Result<Option<_>,_>`. I am find with doing such a construction, but better be clear about that.
* To put the data in the journaling seems preposterous. Why not instead create a `LeaseDatabase`? That seems cleaner.
* To have the `tokio::spawn` in `open_exclusive` which is not async is a break with previous practice. Could we instead have a boolean like `need_lease` in the store and whenever we need a database access we set it up. That is the lease is obtained at the first time we do an access.
* Independent of this PR, we have in journaling.rs the entry `Journal = 1,` which is contrary to our practices.
* In journaling, the database error (pattern "Err(err)" is simply eaten instead of being propagated upstream with ?.
* The uuid in journaling.rs is a random number. Why not instead put the Thread ID? Isn't what we are really after.
* Both for DynamoDB and ScyllaDB, we are going to implement the lease with a key/value. Whjy not use the existing system. Adding more attributes (in DynamoDB) and keys (in ScyllaDB) seems like a waste. If it were up to me, I would write a single key `[255]` and the value would be the thread-id and the end-time.
* The `open_exclusive` is doing the job of spanning the thread. But the threads that does the cancellation token is done in the same thread that creates it.

Why not create a special `(key, value)`, never deletes it.
* If absent, create it with (thread_id, duration).
* If present, if same thread_id, go ahead, if not, check the duration.

Do that check at the first access of an async operation in a `LeaseDatabase` connector.

ma2bd:
> * To make the examples compile under wasm32, the best is to make journaling, lru and the like not available in wasm32, where anyway they are not used.

yes

> * After those corrections, the DynamoDB end-to-end test fail randomly and the ScyllaDB test fail due to syntax error in the request.

yeah claude must have hallucinated some of the DB queries

> * The `obtain_lease` is returning a `Result<Result<_,_>,_>`. Such constructions were previously proposed and refused for `linera-storage` in favor of returning a `Result<Option<_>,_>`. I am find with doing such a construction, but better be clear about that.

I'm not seeing which `Result<Option<_>,_>` you mean in particular. It looks like in many cases `Ok(None)` means "the value was missing". If we change `obtain_lease` to use `Result<Option<_>,_>` instead of `Result<Result<(), _>>`, we will have that `Ok(None)` means "everything's fine" and `Ok(Some(LeaseConflict))` means "there was an error" which is quite confusing.

> * To put the data in the journaling seems preposterous. Why not instead create a `LeaseDatabase`? That seems cleaner.

We can discuss but reusing `DirectWritableKeyValueStore` made the PR much smaller. Also, in practice, journals are used for the shared remote database that we want to protect with a lease. 

> * To have the `tokio::spawn` in `open_exclusive` which is not async is a break with previous practice. Could we instead have a boolean like `need_lease` in the store and whenever we need a database access we set it up. That is the lease is obtained at the first time we do an access.

`tokio::spawn` is not async so it works. We want the lease to be managed immediately after `open_exclusive` not after the write DB write. We could also make `open_exclusive` async.

> * Independent of this PR, we have in journaling.rs the entry `Journal = 1,` which is contrary to our practices.

unrelated and not changeable without migration

> * In journaling, the database error (pattern "Err(err)" is simply eaten instead of being propagated upstream with ?.

probably intended

> * The uuid in journaling.rs is a random number. Why not instead put the Thread ID? Isn't what we are really after.

The thread ID is not stable for a given root key. We could certainly add more information in addition to the UUID.

> * Both for DynamoDB and ScyllaDB, we are going to implement the lease with a key/value. Why not use the existing system. Adding more attributes (in DynamoDB) and keys (in ScyllaDB) seems like a waste. If it were up to me, I would write a single key `[255]` and the value would be the thread-id and the end-time.

This won't allow the atomic test-and-update, which is what this PR is about.

> * The `open_exclusive` is doing the job of spanning the thread. But the threads that does the cancellation token is done in the same thread that creates it.

so what? define "thread"
 
> Why not create a special `(key, value)`, never deletes it.
> 
> * If absent, create it with (thread_id, duration).
> * If present, if same thread_id, go ahead, if not, check the duration.
> 
> Do that check at the first access of an async operation in a `LeaseDatabase` connector.

not sure where you're going with this

PR #4941, opened 2025-11-10
[testnet] Specify the context of errors in `MissingEntries`.

MathieuDutSik:
## Motivation

Specify the context of `MissingEntries` errors.

## Proposal

Add a String to specify that context.

## Test Plan

There is not much way to check it up. 

## Release Plan

To put in TestNet Conway.

## Links

None.

### DIFF
--- a/linera-storage/src/migration.rs
+++ b/linera-storage/src/migration.rs
@@ -135,7 +135,7 @@ where
                 .await?;
             let mut batch = MultiPartitionBatch::new();
             for (base_key, value) in chunk_base_keys.iter().zip(values) {
-                let value = value.ok_or(ViewError::MissingEntries("migration".into()))?;
+                let value = value.ok_or_else(|| ViewError::MissingEntries("migration".into()))?;
                 let (root_key, key) = map_base_key(base_key)?;
                 batch.put_key_value(root_key, key, value);
             }


linera-views/src/views/bucket_queue_view.rs
```diff
@@ -457,8 +457,7 @@ impl<C: Context, T: DeserializeOwned + Clone, const N: usize> BucketQueueView<C,
                         let data = match data {
                             Some(value) => value,
                             None => {
-                                let root_key = self.context.store().root_key()?;
-                                return Err(ViewError::MissingEntries(root_key));
+                                return Err(ViewError::MissingEntries("BucketQueueView A".into()));
```
afck: Maybe:

```suggestion
                                return Err(ViewError::MissingEntries("BucketQueueView::delete_front".into()));
```

(Also below.)

MathieuDutSik: Ok, thanks. Done.

linera-storage/src/migration.rs
```diff
@@ -135,7 +135,7 @@ where
                 .await?;
             let mut batch = MultiPartitionBatch::new();
             for (base_key, value) in chunk_base_keys.iter().zip(values) {
-                let value = value.ok_or(ViewError::MissingEntries(Vec::new()))?;
+                let value = value.ok_or(ViewError::MissingEntries("migration".into()))?;
```
afck: If we're allocating a `String` I'd use `ok_or_else`.

Alternatively I guess we could use `MissingEntries(&'static str)` instead of `String`.

MathieuDutSik: Correct, done.

ma2bd (approved):

PR #4945, opened 2025-11-11
[testnet] Log validator addresses, not pub keys.

afck:
## Motivation

It's hard to tell from the public keys in the logs which validator they refer to.

## Proposal

Print the address instead.

## Test Plan

(Only logs have changed.)

## Release Plan

- These changes should be ported to `main` and
    - be released in a new SDK.

## Links

### DIFF
--- a/linera-core/src/remote_node.rs
+++ b/linera-core/src/remote_node.rs
@@ -110,7 +110,7 @@ impl<N: ValidatorNode> RemoteNode<N> {
             match result {
                 Err(NodeError::MissingCertificateValue) => {
                     debug!(
-                        validator = %self.address(),
+                        address = self.address(),
                         certificate_hash = %certificate.hash(),
                         "validator forgot a validated block value that they signed before",
                     );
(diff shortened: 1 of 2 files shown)


linera-core/src/client/mod.rs
```diff
@@ -4046,13 +4047,16 @@ impl<Env: Environment> ChainClient<Env> {
...
-            .filter_map(move |result| async move {
-                if let Err(error) = &result {
-                    info!(?error, "Could not connect to validator {public_key}");
-                } else {
-                    debug!("Connected to validator {public_key}");
+            .filter_map(move |result| {
+                let address = address.clone();
+                async move {
+                    if let Err(error) = &result {
+                        info!(?error, "Could not connect to validator {address}");
```
deuszx:
```
info!(?error, ?address, "could not connect to validator")
```
:) and below

deuszx (approved):

linera-core/src/client/mod.rs
afck: Done in https://github.com/linera-io/linera-protocol/pull/4945/commits/3319240a60836a9438006c472a74d343cca7e0d2.

PR #4946, opened 2025-11-11
Log validator addresses, not pub keys.

afck:
## Motivation

It's hard to tell from the public keys in the logs which validator they refer to.

## Proposal

Print the address instead.

Also, fix an alignment issue in `wallet show`. (Already fixed on `testnet_conway`.)

## Test Plan

(Only logs have changed.)

## Release Plan

- These changes should be ported to `main` and
    - be released in a new SDK.

## Links

- PR to `testnet_conway`: #4945

### DIFF
--- a/linera-core/src/client/chain_client/mod.rs
+++ b/linera-core/src/client/chain_client/mod.rs
@@ -2773,7 +2773,7 @@ impl<Env: Environment> ChainClient<Env> {
                     {
                         tracing::info!(
                             chain_id = %this.chain_id,
-                            validator = remote_node.address(),
+                            address = remote_node.address(),
                             ?notification,
                             %error,
                             "failed to process notification",
--- a/linera-core/src/remote_node.rs
+++ b/linera-core/src/remote_node.rs
@@ -110,7 +110,7 @@ impl<N: ValidatorNode> RemoteNode<N> {
             match result {
                 Err(NodeError::MissingCertificateValue) => {
                     debug!(
-                        validator = %self.address(),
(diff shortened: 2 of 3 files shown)


linera-core/src/client/chain_client/mod.rs
```diff
@@ -2769,9 +2773,10 @@ impl<Env: Environment> ChainClient<Env> {
                     {
                         tracing::info!(
                             chain_id = %this.chain_id,
-                            validator_public_key = ?remote_node.public_key,
+                            validator = remote_node.address(),
```
deuszx: we should decide whether we log it under `validator`, `address` , `validator_address` or sth else.

afck: Yes… what do you prefer? Maybe `validator_address` is the most clear? On the other hand I like concise logs…

deuszx: maybe `address` then? It's clear from the context it's the validator's address.

afck: ~Done in https://github.com/linera-io/linera-protocol/pull/4946/commits/9c3cc7fab39116162d6578e9d87f64099b7b43d4.~
Done in https://github.com/linera-io/linera-protocol/pull/4946/commits/c8b0b43b6b3dd0c119efdb6acde63165dc30e999.

(Sorry, somehow VSCode search got confused and didn't find all instances…)

deuszx (approved):

PR #4949, opened 2025-11-11
Call `sync_if_needed` for validated block certificates, too.

afck:
## Motivation

In `finalize_locking_block` we currently treat every `CommunicationError` as a timeout, and we don't synchronize (either the client or the validator) if there is a round or block height mismatch.

## Proposal

Call `sync_if_needed` in the "validated block" case, too. Don't return round timeouts, because that never makes sense when submitting a `ValidatedBlock` certificate.

## Test Plan

CI for regressions

We already test `sync_if_needed` in other test cases.

## Release Plan

    - be released in a new SDK.

## Links

### DIFF
--- a/linera-core/src/client/chain_client/mod.rs
+++ b/linera-core/src/client/chain_client/mod.rs
@@ -1834,17 +1834,12 @@ impl<Env: Environment> ChainClient<Env> {
             "Finalizing locking block"
         );
         let committee = self.local_committee().await?;
-        match self
+        let certificate = self
             .client
             .finalize_block(&committee, certificate.clone())
-            .await
-        {
-            Ok(certificate) => {
-                self.update_validators(Some(&committee)).await?;
-                Ok(ClientOutcome::Committed(Some(certificate)))
-            }
-            Err(error) => Err(error),
-        }
+            .await?;
+        self.update_validators(Some(&committee)).await?;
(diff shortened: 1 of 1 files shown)


linera-core/src/client/chain_client/mod.rs
```diff
@@ -1843,18 +1843,6 @@ impl<Env: Environment> ChainClient<Env> {
...
-            {
-                // Communication errors in this case often mean that someone else already
-                // finalized the block or started another round.
-                let timestamp = info.manager.round_timeout.ok_or(error)?;
-                Ok(ClientOutcome::WaitForTimeout(RoundTimeout {
-                    timestamp,
-                    current_round: info.manager.current_round,
-                    next_block_height: info.next_block_height,
-                }))
-            }
```
bart-linera, line 1857: Maybe instead of this `match` we can now have something like this?
```rust
let certificate = self.client.finalize_block(&committee, certificate).await?;
self.update_validators(Some(&committee)).await?;
Ok(ClientOutcome::Committed(Some(certificate)))
```

afck: Good catch!
Done in https://github.com/linera-io/linera-protocol/pull/4949/commits/1f2ea36cf7257fdf9cbc1d01e19ac1abb876e70c.

PR #4951, opened 2025-11-11
First request should be subject to a delayed alternative.

deuszx:
## Motivation

`RequestsScheduler` sends a request to a first peer and, if it fails or time out, enter `try_staggerred_parallel` function which will try new requests after ever-increasing delay (but before the requests time out or fail). This is inconsistent and vulnerable when first request is sent to a slow/unavailable/malicious peer.

## Proposal

Make the logic consistent: first request is subject to the same retry-after-delay logic as the rest. If delay fires first, a new request (to an alternative peer) is started and delay updated. If any request fails w/ an error – a new request is started.

## Test Plan

A unit test already exists for this.

## Release Plan

- These changes should be deployed to testnet and then backported to `main`.

## Links

### DIFF
--- a/linera-core/src/client/requests_scheduler/in_flight_tracker.rs
+++ b/linera-core/src/client/requests_scheduler/in_flight_tracker.rs
@@ -192,7 +192,7 @@ impl<N: Clone> InFlightTracker<N> {
     /// Pops and returns the newest alternative peer from the list.
     ///
     /// Removes and returns the last peer from the alternative peers list (LIFO - newest first).
-    /// Returns None if the entry doesn't exist or the list is empty.
+    /// Returns `None` if the entry doesn't exist or the list is empty.
     ///
     /// # Arguments
     /// - `key`: The request key to look up
--- a/linera-core/src/client/requests_scheduler/scheduler.rs
+++ b/linera-core/src/client/requests_scheduler/scheduler.rs
@@ -1314,7 +1314,7 @@ mod tests {
         // Give time for alternative peers to register
         tokio::time::sleep(Duration::from_millis(100)).await;
 
-        // With the new implementation, alternatives are being popped as staggered parallel runs.
+        // Alternatives are being popped as staggered parallel runs.
         // The first request is blocked waiting for the signal, so staggered parallel has started
         // and may have already popped one or both alternatives. We just verify that at least
         // one alternative was registered (before being popped).
@@ -1457,7 +1457,7 @@ mod tests {
         );
 
         // Second call should start immediately after first fails (aggressive retry)
-        // With new implementation, when node 0 fails immediately, we immediately start node 1
+        // When node 0 fails immediately, we immediately start node 1
         if times.len() > 1 {
             let delay = times[1].1.as_millis();
             assert!(


linera-core/src/client/requests_scheduler/in_flight_tracker.rs
```diff
@@ -188,6 +188,26 @@ impl<N: Clone> InFlightTracker<N> {
             alt_peers.retain(|p| p != peer);
         }
     }
+
+    /// Pops and returns the newest alternative peer from the list.
+    ///
+    /// Removes and returns the last peer from the alternative peers list (LIFO - newest first).
+    /// Returns None if the entry doesn't exist or the list is empty.
```
afck:
```suggestion
    /// Returns `None` if the entry doesn't exist or the list is empty.
```

linera-core/src/client/requests_scheduler/in_flight_tracker.rs
```diff
@@ -188,6 +188,26 @@ impl<N: Clone> InFlightTracker<N> {
...
+    /// # Returns
+    /// - `Some(N)`: The newest alternative peer
+    /// - `None`: No entry exists or alternatives list is empty
+    pub(super) async fn pop_alternative_peer(&self, key: &RequestKey) -> Option<N> {
+        if let Some(entry) = self.entries.read().await.get(key) {
+            let mut alt_peers = entry.alternative_peers.write().await;
+            alt_peers.pop()
+        } else {
+            None
+        }
```
afck, line 209: (Maybe more concise:)

```suggestion
        let entry = self.entries.read().await.get(key)?;
        let mut alt_peers = entry.alternative_peers.write().await;
        alt_peers.pop()
```

linera-core/src/client/requests_scheduler/scheduler.rs
```diff
@@ -28,16 +35,25 @@ use super::{
...
+// Conditional Send trait bound for futures based on target architecture
+#[cfg(target_arch = "wasm32")]
+trait MaybeSendFuture: Future {}
+#[cfg(target_arch = "wasm32")]
+impl<T: Future> MaybeSendFuture for T {}
+
+#[cfg(not(target_arch = "wasm32"))]
+trait MaybeSendFuture: Future + Send {}
+#[cfg(not(target_arch = "wasm32"))]
+impl<T: Future + Send> MaybeSendFuture for T {}
```
afck: (That looks like it could be useful in other parts of the code, too. Maybe something for another PR.)

linera-core/src/client/requests_scheduler/scheduler.rs
```diff
@@ -710,70 +707,151 @@ impl<Env: Environment> RequestsScheduler<Env> {
...
+        let peer_index = AtomicU32::new(0);
+
+        let push_future = |futures: &mut FuturesUnordered<IndexedFuture<T>>, fut: Fut| {
+            #[cfg(target_arch = "wasm32")]
+            #[allow(clippy::redundant_async_block)]
+            futures.push(async move { fut.await }.boxed_local());
+            #[cfg(not(target_arch = "wasm32"))]
+            // if removed types don't match.
+            #[allow(clippy::redundant_async_block)]
+            futures.push(async move { fut.await }.boxed());
```
afck: Does this work, too?

```suggestion
            futures.push(IndexedFuture::new(async move { fut.await }));
```

linera-core/src/client/requests_scheduler/scheduler.rs
```diff
@@ -710,70 +707,151 @@ impl<Env: Environment> RequestsScheduler<Env> {
...
+        let mut next_delay = Box::pin(sleep(staggered_delay * peer_index.load(Ordering::SeqCst)));
 
-            let delayed_fut = async move {
-                if delay.as_millis() > 0 {
-                    sleep(delay).await;
+        // Phase 1: Race between futures completion and delays (while alternatives might exist)
+        loop {
+            // Exit condition: no futures running and can't start any more
+            if futures.is_empty() {
+                if let Some(peer) = self.in_flight_tracker.pop_alternative_peer(key).await {
```
afck, line 741: Should we check here in case `first_peer` was already registered as an alternative peer, or can't that happen anyway?

linera-core/src/client/requests_scheduler/scheduler.rs
```diff
@@ -710,70 +707,151 @@ impl<Env: Environment> RequestsScheduler<Env> {
...
+                // A request completed
+                Either::Left((Some(result), delay_fut)) => {
+                    // Keep the delay future for next iteration
+                    next_delay = delay_fut;
 
-        while let Some((peer_idx, result)) = futures.next().await {
+                    match result {
+                        Ok(value) => {
+                            tracing::trace!(
+                                key = ?key,
```
afck:
```suggestion
                                ?key,
```

linera-core/src/client/requests_scheduler/scheduler.rs
```diff
@@ -710,70 +707,151 @@ impl<Env: Environment> RequestsScheduler<Env> {
...
+                            );
+                            return Ok(value);
+                        }
+                        Err(e) => {
+                            tracing::debug!(
+                                key = ?key,
+                                error = %e,
+                                "staggered parallel request attempt failed"
+                            );
+                            last_error = e;
```
afck: (Just a suggestion; feel free to ignore!)

```suggestion
                        Err(error) => {
                            tracing::debug!(
                                ?key,
                                %error,
                                "staggered parallel request attempt failed"
                            );
                            last_error = error;
```

linera-core/src/client/requests_scheduler/scheduler.rs
```diff
@@ -710,70 +707,151 @@ impl<Env: Environment> RequestsScheduler<Env> {
...
+            }
+        }
+
+        // Phase 2: No more alternatives, just wait for remaining futures to complete
+        while let Some(result) = futures.next().await {
             match result {
                 Ok(value) => {
-                    tracing::info!(
+                    tracing::trace!(
                         key = ?key,
```
afck:
```suggestion
                        ?key,
```

linera-core/src/client/requests_scheduler/scheduler.rs
```diff
@@ -710,70 +707,151 @@ impl<Env: Environment> RequestsScheduler<Env> {
...
                 Err(e) => {
                     tracing::debug!(
                         key = ?key,
-                        peer_index = peer_idx,
                         error = %e,
-                        "staggered parallel retry attempt failed"
+                        "staggered parallel request attempt failed"
                     );
-                    last_error = Some(e);
+                    last_error = e;
```
afck: (Same as above.)

linera-core/src/client/requests_scheduler/scheduler.rs
```diff
@@ -1272,19 +1353,11 @@ mod tests {
...
-            .expect("in-flight entry")
-            .into_iter()
-            .map(|p| p.public_key)
-            .collect::<Vec<_>>();
-        assert_eq!(
-            alt_peers,
-            vec![nodes[1].public_key, nodes[2].public_key],
-            "expected nodes 2 and 3 to be registered as alternative peers"
-        );
+        // With the new implementation, alternatives are being popped as staggered parallel runs.
```
afck:
```suggestion
        // Alternatives are being popped as staggered parallel runs.
```

linera-core/src/client/requests_scheduler/scheduler.rs
```diff
@@ -1419,21 +1495,26 @@ mod tests {
             times[0].1.as_millis()
         );
 
-        // Second call should start after a stagger delay (~10ms)
+        // Second call should start immediately after first fails (aggressive retry)
+        // With new implementation, when node 0 fails immediately, we immediately start node 1
```
afck:
```suggestion
        // When node 0 fails immediately, we immediately start node 1
```

afck (approved):

linera-core/src/client/requests_scheduler/scheduler.rs
deuszx: I thought so too, that's why I asked @Twey for a review :)

linera-core/src/client/requests_scheduler/scheduler.rs
deuszx: Without box-ing that is architecture-speific? I can check.

linera-core/src/client/requests_scheduler/scheduler.rs
deuszx: Will do.

linera-core/src/client/requests_scheduler/scheduler.rs
deuszx: Will update all logs in this file like this.

linera-core/src/client/requests_scheduler/in_flight_tracker.rs
deuszx: Unfortunately, rustc complains about borrowing a value that is dropped at the end of the statement (i.e. at the end of line 204, `entry` is being dropped so we can't refer to it in line 205).

linera-core/src/client/requests_scheduler/scheduler.rs
deuszx: Removed this and `IndexedFuture` below.

linera-core/src/client/requests_scheduler/scheduler.rs
deuszx: Removed.

linera-core/src/client/requests_scheduler/scheduler.rs
deuszx: This and other occurences of the pattern are now udpated.

linera-core/src/client/requests_scheduler/scheduler.rs
deuszx: Ah, I meant to write this code so that it never happens – but I think it might 😬

deuszx: Ah, no – before we call `try_staggered*` we remove `first_peer` from the list of alternatives.

PR #4943, opened 2025-11-10
[fix] Split "flush" into two parts: "pre_save" and "post_save"

ma2bd:
## Motivation

The current code assumes that the DB operation `write_batch` after `flush` always succeed.
If it fails, the view may become inconsistent with the DB storage.

## Proposal

* Split `flush` into two parts: `pre_save` and `post_save`
* Optimize the historical hash to use `pre_save` directly. (However, we now use a mutex to avoid calling `self.inner.pre_save` from `post_save`.)

TODO: review the AI generated `pre_save` / `post_save` functions carefully

## Test Plan

CI

## Release Plan

- These changes should be backported to the latest `testnet` branch, then
    - be released in a new SDK,
    - be released in a validator hotfix.

PR #4953, opened 2025-11-11
Split `flush` into `pre_save` & `post_save`

MathieuDutSik:
## Motivation

Continuing the previous PR #4943 

## Proposal

The issues noted in the PR are addressed.

## Test Plan

The CI.

## Release Plan

It should be backported to Testnet Conway.

## Links

None.

### DIFF
--- a/linera-views/src/views/reentrant_collection_view.rs
+++ b/linera-views/src/views/reentrant_collection_view.rs
@@ -226,22 +226,10 @@ impl<W: View> View for ReentrantByteCollectionView<W::Context, W> {
         Ok(delete_view)
     }
 
-    #[cfg(not(any(target_arch = "wasm32", feature = "web")))]
     fn post_save(&mut self) {
         for (_index, update) in mem::take(&mut self.updates) {
             if let Update::Set(view) = update {
-                let mut view = view.write_arc_blocking();
-                view.post_save();
-            }
-        }
-        self.delete_storage_first = false;
-    }
-
-    #[cfg(any(target_arch = "wasm32", feature = "web"))]
-    fn post_save(&mut self) {
-        for (_index, update) in mem::take(&mut self.updates) {
-            if let Update::Set(view) = update {
-                let mut view = view.try_write().unwrap();
+                let mut view = view.try_write().expect("pre_save was called before");
                 view.post_save();
             }
         }


linera-views/src/views/key_value_store_view.rs
```diff
@@ -305,61 +312,71 @@ impl<C: Context> View for KeyValueStoreView<C> {
...
+        }
+        self.deletion_set.delete_storage_first = false;
+        self.deletion_set.deleted_prefixes.clear();
+        self.updates.clear();
+        self.sizes.post_save()?;
+        let hash = *self.hash.lock().unwrap();
+        self.stored_hash = hash;
+        self.stored_total_size = self.total_size;
+        Ok(())
+    }
```
ma2bd: This still looks wrong. (The lines `if self.deletion_set.delete_storage_first { .. }` do nothing)

ma2bd: Perhaps the original code is wrong (and by that I mean: has dead code)

linera-views/src/views/reentrant_collection_view.rs
```diff
@@ -189,44 +189,56 @@ impl<W: View> View for ReentrantByteCollectionView<W::Context, W> {
...
-        self.delete_storage_first = false;
         Ok(delete_view)
     }
 
+    fn post_save(&mut self) -> Result<(), ViewError> {
+        for (index, update) in mem::take(&mut self.updates) {
+            if let Update::Set(view) = update {
+                let mut view = view
+                    .try_write()
+                    .ok_or_else(|| ViewError::TryLockError(index.clone()))?;
```
ma2bd: `post_save` must not return an error. I believe we should panic here. It will never happen because `pre_save` is called earlier and was able to acquire all the locks without failing -- all happening while the context calling `pre_save` and `post_save` has an exclusive reference on the view, which prevent new locks from being acquired.

linera-views/src/views/mod.rs
```diff
@@ -94,12 +94,17 @@ pub trait View: Sized {
...
     /// The returned boolean indicates whether the operation removes the view or not.
-    fn flush(&mut self, batch: &mut Batch) -> Result<bool, ViewError>;
+    fn pre_save(&self, batch: &mut Batch) -> Result<bool, ViewError>;
+
+    /// Updates the view state after the batch has been executed in the database.
+    /// This should be called after `pre_save` and after the batch has been successfully written to storage.
+    /// This leaves the view in a clean state with no pending changes.
+    ///
+    /// May panic if `pre_save` was not called right before on `self`.
+    fn post_save(&mut self) -> Result<(), ViewError>;
```
ma2bd: See other comment. Please do not return a `Result`. Otherwise we get a potential DB corruption again.

linera-views/src/views/reentrant_collection_view.rs
MathieuDutSik: We recently had that struggle with the `clone_unchecked` where we chose in the end not to panic. So, I thought it was best to do the same.

Otherwise, yes, we can `.unwrap()`. It is somewhat equivalent to throwing an error that is never caught.

linera-views/src/views/mod.rs
MathieuDutSik: Ok, I eliminated the `Result<_,_>` entry.
I put code specific to wasm32 / web which does unwrap() and other parts where this is not needed.

MathieuDutSik: Reverted that to now use the `.try_write().expect("After pre_save, we should have access")`.

linera-views/src/views/reentrant_collection_view.rs
ma2bd: > We recently had that struggle with the clone_unchecked where we chose in the end not to panic. So, I thought it was best to do the same.

It is different. Here we have a justification that the panic will never happen under mild assumptions (certain internal APIs are not called)

PR #4955, opened 2025-11-11
Improve the native fungible token ABI

ma2bd:
## Motivation

A step towards making the native fungible token usable in CLOB etc.

## Proposal

* Remove NativeFungibleTokenAbi and use FungibleTokenAbi instead
* Accept that some allowance operations are not available (yet).

This PR does NOT address #4416, though.

## Test Plan

CI

## Release Plan

    - be released in a new SDK

### DIFF
--- a/examples/native-fungible/src/contract.rs
+++ b/examples/native-fungible/src/contract.rs
@@ -5,7 +5,7 @@
 
 use linera_sdk::{
     abis::fungible::{
-        FungibleResponse, InitialState, NativeFungibleOperation, NativeFungibleTokenAbi, Parameters,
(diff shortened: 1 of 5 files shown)


MathieuDutSik:
I was thinking about eating the bullet and putting an array "allowances" in `SystemExecutionStateChain`. When unused, it is just the cost of a pointer.

About the permissions, I do not know a solution.

ma2bd:
> I was thinking about eating the bullet and putting an array "allowances" in `SystemExecutionStateChain`. When unused, it is just the cost of a pointer.
> 
> About the permissions, I do not know a solution.

I can't say whether allowances will be useful or not but I don't mind them being also implemented in `SystemExecutionStateChain` if this helps.

PR #4957, opened 2025-11-12
Backport of staggerred parallel feature from testnet_conway branch #4913 #4951

deuszx:
## Motivation

Backport of #4913 #4951 (no conflicts).

## Proposal

Requests scheduler will now attach a timer that races with a spawned request. Whenever timer comes back, before the request resolves, it will repeat the same request using an alternative peer from the list (if any). This way we won't be blocked by slow or unavailable peers.

## Test Plan

Unit test added.

## Release Plan

## Links

#4913 #4951 

### DIFF
--- a/linera-core/src/client/requests_scheduler/scheduler.rs
+++ b/linera-core/src/client/requests_scheduler/scheduler.rs
@@ -1,7 +1,14 @@
 // Copyright (c) Zefchain Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-use std::{cmp::Ordering, collections::BTreeMap, future::Future, sync::Arc};
+use std::{
+    collections::BTreeMap,
+    future::Future,
+    sync::{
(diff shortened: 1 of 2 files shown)


linera-core/src/client/requests_scheduler/scheduler.rs
```diff
@@ -602,41 +619,38 @@ impl<Env: Environment> RequestsScheduler<Env> {
...
                                         );
                                     }
                                 }
-                                Err(e) => {
+                                Err(error) => {
                                     tracing::trace!(
-                                        key = ?key,
-                                        error = %e,
+                                        ?key,
+                                        ?error,
```
afck, line 645:
```suggestion
                                        %error,
```

(At least we use `Display` elsewhere. Sorry, I missed that on `main`.)

deuszx: Damn github copilot.

Issue #2809, opened 2024-11-04
`Post` trait doesn't correctly codify which types can be posted across threads


Twey:
Currently the `linera_base::task::Post` trait allows posting any object that can be cast to/from a `JsValue`.  However, some JavaScript types cannot be posted across threads: the only objects that can be safely posted are [those supported by the `structuredClone` algorithm](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm#supported_types).  Other types will fail either at send or deserialize time (some objects can be posted without error, but will arrive on the other end as an empty object `{}`).

Likewise, there are objects that are [transferable](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Transferable_objects) but not directly supported by `structuredClone` that we should be able to post, but currently cannot due to the structure of the trait.  

Maybe a better solution would be to have a trait for _transferable_ objects, of which structured-cloneable objects are a trivial case, and use that in `linera_base::task`.  This would correctly capture the set of objects we can safely transfer.

Twey:
On nightly (which we require for the Web anyway) we can use the [`auto_traits`](https://doc.rust-lang.org/beta/unstable-book/language-features/auto-traits.html) unstable feature to define an auto trait for objects that can be sent through `postMessage`.

Issue #3651, opened 2025-03-26
OOM on Web due to ephemeral threads


Twey:
Currently the way we implement threads in Wasm means that every new thread grows the single shared Wasm memory.  Since Wasm provides no way to shrink the memory, this results in the memory usage growing with every new thread (i.e. every block executed) until it hits a browser-imposed limit and crashes.

Some solutions to this problem that I think should work, in ascending order of difficulty:
- change `linera-protocol` to use a fixed-size pool of threads that are re-used when executing bytecode
- change `wasm-bindgen` to use multiple memories, so that the thread's memory can be added to a fresh memory that is freed when the thread ends
- change our fork of `wasm-thread` to track a thread's used memory and release it when the thread exits

Issue #2927, opened 2024-11-20
Support dynamic loading of modules on the Web


Twey:
This is currently disabled since it would require an async call to receive the module via `postMessage`, which [we don't currently support](https://github.com/linera-io/linera-protocol/issues/2552).

deuszx:
This is worked around via https://github.com/linera-io/linera-protocol/pull/6444/changes/fa95da4f4c5ecd2af2a7db4abb369665ce3eadd3

PR #4959, opened 2025-11-13
Replace `wasm_thread` with `web-thread`

Twey:
## Motivation

Our hacked-up version of `wasm_thread` isn't a great fit for our use cases.

Firstly, `wasm_thread` aims to mimic the API of `std::thread`.  This was never sufficient for us as we need to send non-`Send` messages across threads actively using `postMessage`, which `std::thread` has no equivalent for.  We patched `wasm_thread` but it was at odds with the point of the package.

Then, in order to use `wasm_thread` without threading a shim script URL through the code down into `linera-base`, we dynamically got the URL of the `wasm-bindgen` shim script using `import.meta.url`.  This is poorly (and inconsistently) supported by bundlers, which want to know what the URL will be used for; but since the `Worker` construction in `wasm_thread` is done in Rust code they cannot analyze it.  This resulted in our `@linera/client` library being rather fragile under different bundlers, with workarounds required to ensure they don't interfere with the paths of the various files.

## Proposal

Introduce the [`web-thread`](https://docs.rs/web-thread/latest/web_thread/) library, a simplified equivalent to `wasm_thread` that has a Web-first API, including a trait for types that can be passed by `postMessage` (including [transferable objects](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Transferable_objects)), and whose worker logic is primarily implemented in JavaScript, making it much more comprehensible to bundlers.

This allows us:
- to communicate both `Send` and `Post` objects with long-running threads, as is required by https://github.com/linera-io/linera-protocol/issues/3651 and eventually https://github.com/linera-io/linera-protocol/issues/2927
- to remove the bundling caveats from Linera documentation
- to support WebPack
- to delete `linera_base::task` and the conditional compilation therein

## Test Plan

The `web-thread` library itself has been tested with Vite as well as WebPack (via Next.js with the `--webpack` flag), and includes example applications to that effect.  

I have manually tested our two Web applications in this repository.

I defer to CI for full tests of the native pathways.

## Release Plan

## Links

- Fixes https://github.com/linera-io/linera-protocol/issues/2809.
- First step towards https://github.com/linera-io/linera-protocol/issues/3651.
- [`web-thread` docs](https://docs.rs/web-thread/latest/web_thread/)

### Turbopack

Turbopack is currently not supported due to an open issue when processing cyclic dependencies.  See the following discussions for more information:

* https://github.com/vercel/next.js/issues/85119
* https://github.com/vercel/next.js/discussions/77102
* https://github.com/emscripten-core/emscripten/issues/20580

### DIFF
--- a/linera-client/src/chain_listener.rs
+++ b/linera-client/src/chain_listener.rs
@@ -170,7 +170,7 @@ impl<C: ClientContext> ListeningClient<C> {
     }
 
     async fn stop(self) {
-        // TODO(TODO): this is unnecessary: the join handle now also acts as an abort handle
+        // TODO(#4965): this is unnecessary: the join handle now also acts as an abort handle
         drop(self.abort_handle);
         self.join_handle.await;
     }


linera-storage/src/lib.rs
```diff
@@ -353,12 +351,11 @@ pub trait Storage: Sized {
             compressed_bytes: content.into_arc_bytes(),
         };
         #[cfg_attr(not(any(with_wasm_runtime, with_revm)), allow(unused_variables))]
-        let service_bytecode = linera_base::task::Blocking::<linera_base::task::NoInput, _>::spawn(
-            move |_| async move { compressed_service_bytecode.decompress() },
-        )
-        .await
-        .join()
-        .await?;
+        let service_bytecode = web_thread::Thread::new()
```
deuszx, line 354: What I liked about previous code was that it didn't mention "web" (or no web) – it was hidden in `linera_base::task`. Now I feel like this piece (and one above) are meant to run only on web (b/c `web_thread`) but – IIUC – the web_thread actually makes the choice which version to run based on the target architecture?

linera-execution/src/lib.rs
```diff
@@ -272,10 +276,8 @@ pub enum ExecutionError {
     UnauthorizedHttpRequest(reqwest::Url),
     #[error("Attempt to perform an HTTP request to an invalid URL")]
     InvalidUrlForHttpRequest(#[from] url::ParseError),
-    #[error("Failed to send contract code to worker thread: {0:?}")]
-    ContractModuleSend(#[from] linera_base::task::SendError<UserContractCode>),
-    #[error("Failed to send service code to worker thread: {0:?}")]
-    ServiceModuleSend(#[from] linera_base::task::SendError<UserServiceCode>),
+    #[error("Worker thread failure: {0:?}")]
+    Thread(#[from] web_thread::Error),
```
deuszx, line 281: Worker thread are not target-specific, right? i.e. we use them both on web and native.

linera-storage/src/lib.rs
Twey: That's right (or in fact `web-thread-select`, which is an API wrapper crate).

There's unfortunately no getting around mentioning the Web here.  The existing abstraction in `linera-base` was actually misnamed: as I think you pointed out at some point, it always looked a bit funny because the name didn't mention the Web, but the API made no sense on UNIX, which doesn't really have a concept of explicitly sending things between threads.  The API of this crate really is designed for the Web, and the native equivalent truly is just a shim.

The only way we could sensibly avoid mentioning the Web in the name of this abstraction layer (wherever we put it) is if we generalized it to `explicit-send-thread` or something.  But at some point there is such a thing as too much abstraction, especially since we don't even have a non-Web implementation (not any plans to make one) :)

linera-execution/src/lib.rs
Twey: That's right.  What we call `web_thread` is actually `web-thread-select`, which can automatically shim the `web-thread` API on UNIX (using a combination of `wasm32` arch + `web` feature flag since [there's no target triple for the Web](https://users.rust-lang.org/t/fixing-rusts-webassembly-targets/88947)).

linera-storage/src/lib.rs
Twey: We could alternatively keep the entrypoint to this code in `linera_base::task`, but it would just be re-exporting `web-thread-select`.  We re-export some things (e.g. crypto) from `linera-base` in order to cement our choices into the Linera API/ABI, but for internal things like this I don't think it makes much sense.

I'd also (zooming out a bit) quite like to remove a bunch of stuff from `linera-base`: it's become a bit of a catch-all for utilities used throughout the codebase, even those that don't work on all our supported targets (e.g. on-chain contracts), which I think was the original point of `linera-base`.

Twey:
> It's understood but I'd like to "say" it out loud anyway: with the new crate living in external repository it makes (bug)fixes slower and harder to detect.

On the other hand it means that `web-thread` can be much more self-contained.  Since its scope is well-defined I hope we won't have to make too many changes to it.

deuszx:
> > It's understood but I'd like to "say" it out loud anyway: with the new crate living in external repository it makes (bug)fixes slower and harder to detect.
> 
> On the other hand it means that `web-thread` can be much more self-contained. Since its scope is well-defined I hope we won't have to make too many changes to it.

it could live in this project's repositroy but have separate release cycle.

PR #4961, opened 2025-11-13
[nits] improve readability of BucketQueueView

ma2bd:
## Motivation

I forced myself to read this code by trying to improve it

## Proposal

Lots of small changes

## Test Plan

CI

## Release Plan

This could be backported

### DIFF
--- a/linera-views/src/views/bucket_queue_view.rs
+++ b/linera-views/src/views/bucket_queue_view.rs
@@ -37,60 +37,58 @@ mod metrics {
 }
 
 /// Key tags to create the sub-keys of a [`BucketQueueView`] on top of the base key.
-/// * The Front is special and downloaded at the view loading.
-/// * The Store is where the structure of the buckets is stored.
-/// * The Index is for storing the specific buckets.
 #[repr(u8)]
 enum KeyTag {
-    /// Prefix for the front of the view
+    /// Key tag for the front bucket (index 0).
     Front = MIN_VIEW_TAG,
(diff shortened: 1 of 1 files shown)


linera-views/src/views/bucket_queue_view.rs
```diff
@@ -37,91 +37,98 @@ mod metrics {
...
-    /// prefix `KeyTag::Index`.
-    indices: Vec<(usize, usize)>,
-    /// The position of the front in the first index.
-    position: usize,
+struct BucketStore {
+    /// The descriptions of all stored buckets. The first description is expected to start
+    /// with index 0 (front bucket) and will be ignored.
+    descriptions: Vec<BucketDescription>,
+    /// The position of the front value in the front bucket.
+    front_position: usize,
```
MathieuDutSik, line 57: I know it is minor, but since the PR is about readability, could we order the code as
* Most inner struct/enum, followed by "impl .... {" statements.
* Then the next most inner
* And in the end the BucketQueueView.

So, here the ordering would be KeyTag, BucketDescription, BucketStore, Cursor, State<T>, Bucket<T>. It is mostly like that now. Just switch BucketDescription/BucketStore and put the definition of Bucket<T> before its implementation.

ma2bd: I like that you have a consistent rule but the problem is that Rust engineers are typically used to doing the opposite (top-down ordering, starting with the most important containers).

PR #4962, opened 2025-11-13
Decrease certificate and block proposal stack size.

afck:
## Motivation

We have seen stack overflows before that we fixed by `Box`ing some large futures: https://github.com/linera-io/linera-protocol/pull/4882

The root cause for some of these is that we often put large types like certificates and block proposals on the stack.

## Proposal

Box the `value` in certificates: this usually contains a whole block.

Box the `BlockProposal` fields.

Both mostly done by Claude. Its summary:

  - `GenericCertificate<ConfirmedBlock>`: 664 bytes → 40 bytes (94% reduction)
  - `GenericCertificate<ValidatedBlock>`: 664 bytes → 40 bytes (94% reduction)
  - `GenericCertificate<Timeout>`: 112 bytes → 40 bytes (64% reduction)
  - `BlockProposal`: 688 bytes → 24 bytes (96.5% reduction)
  - `LockingBlock` enum: 688 bytes → 40 bytes (94% reduction)

A possible alternative: Accept that some types are very large; avoid passing them by value and holding them on the stack across an `await` point. (I.e. in practice keep them in a `Box` most of the time.)

## Test Plan

CI

## Release Plan

- These changes _could_ be backported to `testnet_conway`.

## Links

### DIFF
--- a/linera-chain/src/data_types/mod.rs
+++ b/linera-chain/src/data_types/mod.rs
@@ -313,10 +313,10 @@ pub enum OriginalProposal {
 #[derive(Clone, Debug, Serialize, Deserialize, Allocative)]
 #[cfg_attr(with_testing, derive(Eq, PartialEq))]
 pub struct BlockProposal {
-    pub content: ProposalContent,
-    pub signature: AccountSignature,
+    pub content: Box<ProposalContent>,
+    pub signature: Box<AccountSignature>,
     #[debug(skip_if = Option::is_none)]
-    pub original_proposal: Option<OriginalProposal>,
+    pub original_proposal: Option<Box<OriginalProposal>>,
 }
 
 /// A message together with kind, authentication and grant information.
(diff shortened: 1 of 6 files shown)


linera-chain/src/data_types/mod.rs
```diff
@@ -313,10 +313,10 @@ pub enum OriginalProposal {
 #[derive(Clone, Debug, Serialize, Deserialize, Allocative)]
 #[cfg_attr(with_testing, derive(Eq, PartialEq))]
 pub struct BlockProposal {
-    pub content: ProposalContent,
-    pub signature: AccountSignature,
+    pub content: Box<ProposalContent>,
+    pub signature: Box<AccountSignature>,
```
deuszx, line 317: Is signature really that big? I thought it's couple hundred bytes at most.

afck: I think it was 128; it contains a public key, too, does it? Sure, we could remove that box again.

deuszx: We don't need to box 128 bytes surely.

bart-linera: When single kilobytes are a problem, 128 bytes is not that insignificant IMO. Not saying we definitely have to box it, but it doesn't sound that unreasonable to me.

PR #4963, opened 2025-11-13
Thread latest certficiate for chain updates

deuszx:
## Motivation

When debugging Faucet performance we noticed that `communicate_with_quorum` takes significantly longer when sending chain updates (compared to submitting block proposals and validated certificates). Comparing the work being done between the calls, `send_chain_information` additionally has to load the latest hash from the `local_node.chain_state_view` (which is two async calls) and then read a certificate for it from the storage (another async call).

## Proposal

Thread through the latest certificate if it's known and don't read it from the storage in `send_chain_information`.

## Test Plan

Tests are updated.

## Release Plan

- Should be deployed to testnet Faucet
- Backport to `main`.

## Links

### DIFF
--- a/linera-client/src/chain_listener.rs
+++ b/linera-client/src/chain_listener.rs
@@ -459,7 +459,16 @@ impl<C: ClientContext + 'static> ChainListener<C> {
     async fn update_validators(&self, notification: &Notification) -> Result<(), Error> {
         let chain_id = notification.chain_id;
         let listening_client = self.listening.get(&chain_id).expect("missing client");
-        if let Err(error) = listening_client.client.update_validators(None, None).await {
+        let latest_block = if let Reason::NewBlock { hash, .. } = &notification.reason {
+            listening_client.client.read_certificate(*hash).await.ok()
+        } else {
+            None
+        };
+        if let Err(error) = listening_client
+            .client
+            .update_validators(None, latest_block)
+            .await
+        {
             warn!(
                 "Failed to update validators about the local chain after \
                  receiving {notification:?} with error: {error:?}"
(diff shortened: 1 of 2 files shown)


linera-core/src/updater.rs
```diff
@@ -621,31 +628,35 @@ where
         chain_id: ChainId,
         target_block_height: BlockHeight,
         delivery: CrossChainMessageDelivery,
+        latest_certificate: Option<GenericCertificate<ConfirmedBlock>>,
```
afck, line 631: This is basically an alternative to `target_block_height`, is it? I wonder whether we should use `Either`, and specify exactly one of them.

linera-client/src/chain_listener.rs
```diff
@@ -459,7 +459,7 @@ impl<C: ClientContext + 'static> ChainListener<C> {
     async fn update_validators(&self, notification: &Notification) -> Result<(), Error> {
         let chain_id = notification.chain_id;
         let listening_client = self.listening.get(&chain_id).expect("missing client");
-        if let Err(error) = listening_client.client.update_validators(None).await {
+        if let Err(error) = listening_client.client.update_validators(None, None).await {
```
afck: I wonder if, in the case of a `NewBlock` notification, it would be faster to load the certificate here, rather than in each of the futures later.

deuszx: Yes, good point. Especially that we did discuss briefly sending whole certificate as part of `NewBlock` notification (if it's small).

deuszx: Done in [Read certificate and thread through for NewBlock notifications](https://github.com/linera-io/linera-protocol/pull/4963/commits/4774206c6fbff4e5b6f991871c0e9d1d3617dbf8).

PR #4972, opened 2025-11-15
rename Amount::saturating_div into saturating_ratio

ma2bd:
## Motivation

The `Amount` can be multiplied by `u128` using `saturating_mul` but we don't have an inverse operation. The existing `saturating_div` compute a `u128` ratio between two amounts.

## Proposal

* rename `Amount::saturating_div` into `saturating_ratio`
* create a new `Amount::saturating_div` that takes a u128 similar to `saturating_mul`

## Test Plan

CI

## Release Plan

    - be released in a new SDK,

### DIFF
--- a/linera-base/src/data_types.rs
+++ b/linera-base/src/data_types.rs
@@ -463,7 +463,7 @@ macro_rules! impl_wrapped_number {
                 Ok(())
             }
 
-            /// Saturating multiplication.
+            /// Saturating division.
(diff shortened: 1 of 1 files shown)


afck (approved):

linera-base/src/data_types.rs
```diff
@@ -463,6 +463,11 @@ macro_rules! impl_wrapped_number {
                 Ok(())
             }
 
+            /// Saturating multiplication.
```
```suggestion
            /// Saturating division.
```

PR #4973, opened 2025-11-15
Support fractional prices in the matching engine

### LINKED ISSUE #841: Implement Euclidean division in the MatchingEngine
At the present time, the matching engine is done only with integer prices.
However, having fractional prices or fixed fractions (like 5 decimals) would
be nice.

But the tokens are not divisible, so purely using fractions is not working and
so we need to use the Euclidean division to be able to handle more general
prices. This causes a number of complications:
* Suppose you want to buy token1 using token0 and we have 9 token0 on hand.
* Then at a price of 2.5 you may fill 2 token1 for a price of 5 token0. And that's it at a price of 2.5
* But then the price of 3 you can get another token1 and that may be something that you want.

There are other issues with serialization and the ordering but that is secondary.

ma2bd:
## Motivation

* Fixes #841
* Demonstrate how to embed parameters in the "extra" field of `View`s to help with code structure. (I mentioned this possibility to @kikakkz once but we had no example to show.) 

## Proposal

Allow creators of a market to specify a fixed decimal precision for prices.
* The amount to be paid for an order is: `amount = quantity * price/10^precision`.
* If `quantity` (expressed in "attos", the lowest precision for `Amount`) is not a multiple of 10^precision, then the order is simply invalid.

We modify the code in several steps:
* add application parameters to storage context
* nit: rename `next_order_number` into `next_order_id`
* rename "amount" into "quantity" + add integration tests
* reorganize the code (moving methods closer to their main object)
* finally, add the notion of "price decimals"

Discussion:
* AFAICT it is very standard for market prices to have fixed precision. Eventually, `Amount`s should probably have a configurable (reduced) precision as well.
* Parameters being immutable, copying them in the `extra` field is pretty harmless. One could also embed the entire runtime using `Arc<Mutex<ContractRuntime<MyContractState>>` but this is arguably a tradeoff (e.g. any view can now schedule outgoing messages).

stacked with #4971 and #4972

## Test Plan

CI

## Release Plan

    - be released in a new SDK,

### DIFF
--- a/examples/amm/src/lib.rs
+++ b/examples/amm/src/lib.rs
@@ -5,10 +5,9 @@
 
 use async_graphql::{scalar, Request, Response};
 use linera_sdk::{
-    graphql::GraphQLMutationRoot,
-    linera_base_types::{AccountOwner, Amount, ContractAbi, ServiceAbi},
     abis::fungible::FungibleTokenAbi,
-    linera_base_types::ApplicationId,
+    graphql::GraphQLMutationRoot,
+    linera_base_types::{AccountOwner, Amount, ApplicationId, ContractAbi, ServiceAbi},
 };
 use serde::{Deserialize, Serialize};
 


examples/matching-engine/src/contract.rs
```diff
@@ -57,17 +57,20 @@ impl Contract for MatchingEngineContract {
     type Parameters = Parameters;
     type EventValue = ();
 
-    async fn load(runtime: ContractRuntime<Self>) -> Self {
-        let state = MatchingEngineState::load(runtime.root_view_storage_context())
+    async fn load(mut runtime: ContractRuntime<Self>) -> Self {
+        let parameters = runtime.application_parameters();
+        let context = linera_views::context::ViewContext::new_unchecked(
```
ma2bd, line 37: (`new_unsafe` in the current SDK)

examples/matching-engine/src/contract.rs
```diff
@@ -57,17 +57,20 @@ impl Contract for MatchingEngineContract {
...
+            runtime.key_value_store(),
+            Vec::new(),
+            parameters,
```
ma2bd, line 40: This could also be `Arc::new(Mutex::new(runtime))` if we wanted (also I haven't tried)

linera-base/src/data_types.rs
```diff
@@ -463,6 +463,11 @@ macro_rules! impl_wrapped_number {
                 Ok(())
             }
 
+            /// Saturating multiplication.
```
```suggestion
            /// Saturating division.
```

examples/matching-engine/tests/basic_operations.rs
```diff
@@ -0,0 +1,371 @@
...
+        .add_block(|block| {
+            block.with_messages_from(&certificate);
+        })
+        .await;
+
+    user_chain_a.handle_received_messages().await;
+    matching_chain.handle_received_messages().await;
+
+    // The order should be created successfully (no panics)
+}
```
afck: (I'm not a fan of having tests that are basically subsets of another test. But I guess these ones are cheap.)

examples/matching-engine/tests/basic_operations.rs
```diff
@@ -0,0 +1,371 @@
...
+
+    matching_chain
+        .add_block(|block| {
+            block.with_messages_from(&cert2);
+        })
+        .await;
+
+    user_chain_a.handle_received_messages().await;
+
+    // The order should be modified successfully
```
afck: Shouldn't we check at this point that five tokens have been refunded and the order now has an amount of 95?

examples/matching-engine/tests/order_matching.rs
```diff
@@ -0,0 +1,443 @@
...
+        })
+        .await;
+
+    user_chain_a.handle_received_messages().await;
+    user_chain_b.handle_received_messages().await;
+    matching_chain.handle_received_messages().await;
+
+    // Both orders should be fully matched
+    // User A should have received 10 tokens B
+    // User B should have received 10 tokens A * price
```
afck: Yes, but we don't check that?

examples/matching-engine/tests/order_matching.rs
```diff
@@ -0,0 +1,443 @@
...
+        .add_block(|block| {
+            block.with_messages_from(&cert_ask);
+        })
+        .await;
+
+    user_chain_a.handle_received_messages().await;
+    user_chain_b.handle_received_messages().await;
+    matching_chain.handle_received_messages().await;
+
+    // Orders should match - liquidity provider gets the better price
```
afck: But we don't check?

examples/matching-engine/src/lib.rs
```diff
@@ -28,27 +28,25 @@ impl ServiceAbi for MatchingEngineAbi {
...
-/// account[0] -= price * quantity;
+/// account[0] -= price * quantity * 10^-price_decimals;
 /// account[1] += quantity;
 /// ```
-/// Thus the quantity (also called _count_) is an `Amount`.
+/// where `price_decimals` is a parameter set when the market is created.
+///
+/// The `quantity` (also called _count_) is of type `Amount` as well as the balance of the
+/// accounts. Therefore, the number of decimals used by quantities in a valid order must
+/// be not exceed `Amount::DECIMAL_PLACES - price_decimals`.
```
afck:
```suggestion
/// not exceed `Amount::DECIMAL_PLACES - price_decimals`.
```

examples/matching-engine/src/lib.rs
```diff
@@ -93,6 +92,7 @@ impl CustomSerialize for PriceAsk {
 #[derive(Clone, Copy, Debug, SimpleObject, InputObject)]
 #[graphql(input_name = "PriceBidInput")]
 pub struct PriceBid {
+    /// A price multiplied by 10^price_decimals.
```
afck: You write "multiplied" here and "divided" above…?

In both cases, isn't `price` the actual price, _multiplied_ by `10^price_decimals`? I.e. to get the actual price, you need to _divide_ `price`?

examples/matching-engine/src/lib.rs
```diff
@@ -176,13 +207,23 @@ impl Order {
...
 impl Parameters {
+    pub fn check_precision(&self, order: &Order) {
+        order
+            .check_precision(self.price_decimals)
+            .expect("Invalid precision in order");
+    }
+
     pub fn product_price_amount(&self, price: Price, quantity: Amount) -> Amount {
-        quantity
+        let count = quantity.saturating_div(10u128.pow(self.price_decimals as u32));
```
afck, line 225: Now that I think about it, a `checked_div` would maybe be more appropriate.

I mean, fungible tokens aren't meant to have a supply `> Amount::MAX`, but if they do, maybe it's better to fail rather than saturate? (Not necessarily in this PR.)

ma2bd: Note that this app cannot mint tokens. But ok let's think about the case of a zero price

ma2bd:
Good point. Let me review the tests

examples/matching-engine/src/lib.rs
afck: (Ah, right, this divides only by the power of ten, and the multiplication below already panics. 👍)

examples/matching-engine/tests/basic_operations.rs
ma2bd: No actually these tests are surprisingly slow (perhaps because of app linking time?)

PR #4974, opened 2025-11-16
[matching engine] Track pending orders on user chains

ma2bd:
## Motivation

Allow users to display their pending orders without following the entire CLOB appchain

## Proposal

* Add a new type of message and a field of user chains
* Replace `Cancel` orders by `Modify` with new quantity equal to `0`
* Simplify `ModifyQuantity` to avoid code duplication

## Test Plan

CI

## Release Plan

### DIFF
--- a/examples/matching-engine/src/contract.rs
+++ b/examples/matching-engine/src/contract.rs
@@ -11,7 +11,7 @@ use linera_sdk::{
     Contract, ContractRuntime,
 };
 use matching_engine::{
-    MatchingEngineAbi, Message, Operation, Order, OrderNature, Parameters, Price,
+    MatchingEngineAbi, Message, Operation, Order, OrderNature, Parameters, PendingOrderInfo, Price,
 };
 use state::{MatchingEngineState, ModifyQuantity, Transfer};
 
@@ -89,15 +89,16 @@ impl Contract for MatchingEngineContract {
         }
     }
 
-    /// Execution of the order on the creation chain
+    /// Execution of messages
     async fn execute_message(&mut self, message: Message) {
-        assert_eq!(
-            self.runtime.chain_id(),
-            self.runtime.application_creator_chain_id(),
-            "Action can only be executed on the chain that created the matching engine"
-        );
         match message {
             Message::ExecuteOrder { order } => {
+                // This message can only be executed on the matching engine chain
+                assert_eq!(
+                    self.runtime.chain_id(),
+                    self.runtime.application_creator_chain_id(),
+                    "ExecuteOrder can only be executed on the chain that created the matching engine"
+                );
                 let owner = order.owner();
                 let origin_chain_id = self.runtime.message_origin_chain_id().expect(
                     "Incoming message origin chain ID has to be available when executing a message",
@@ -106,6 +107,53 @@ impl Contract for MatchingEngineContract {
                     .check_account_permission(owner)
                     .expect("Permission for ExecuteOrder message");
                 self.execute_order_local(order, origin_chain_id).await;
+            }
+            Message::OrderPending {
+                owner,
+                order_id,
+                order_info,
+            } => {
+                // This message is received on the sender chain from the matching engine
+                // Order was inserted and is pending
+                let pending_orders = self
+                    .state
+                    .pending_orders
+                    .get_mut_or_default(&owner)
+                    .await
+                    .expect("Failed to load pending orders");
+                pending_orders.insert(order_id, order_info);
+            }
+            Message::OrderUpdated {
+                owner,
+                order_id,
+                new_quantity,
+            } => {
+                // This message is received on the sender chain from the matching engine
+                let pending_orders = self
+                    .state
+                    .pending_orders
+                    .get_mut(&owner)
+                    .await
+                    .expect("Failed to load pending orders")
+                    .expect("Account should have pending orders");
+
+                // Update the pending order
+                if let Some(order) = pending_orders.get_mut(&order_id) {
+                    order.quantity = new_quantity;
+                }
+            }
+            Message::OrderRemoved { owner, order_id } => {
+                // This message is received on the sender chain from the matching engine
+                let pending_orders = self
+                    .state
+                    .pending_orders
+                    .get_mut(&owner)
+                    .await
+                    .expect("Failed to load pending orders")
+                    .expect("Account should have pending orders");
+
+                // Remove the pending order
+                pending_orders.remove(&order_id);
             }
         }
     }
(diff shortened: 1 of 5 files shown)


MathieuDutSik (commented):
This design of keeping track of the order structure inside of the Matching Engine is not the standard way that this is done.

What exchanges are publishing are the list of operations. See for example [Nasdaq: Computer-To-Computer-Interface](https://www.sec.gov/files/rules/other/nasdaqllcf1a4_5/e_nasdaqctci.pdf).
Then it is the job of the receivers of the information to keep track of he orders and maybe have a nice `MapView<AccountOwner, BTreeMap<OrderId, ....>>`.

If we follow this design, then what we would do is instead emit events corresponding to the transactions. Then another contract would keep track of those events and of the relevant balances, pending orders, and so on if he so wishes. We have all the functionalities in Linera for that.

examples/matching-engine/src/state.rs
```diff
@@ -546,24 +578,20 @@ impl LevelView {
...
-            ModifyQuantity::Partial(reduce_quantity) => state_order
-                .quantity
-                .try_sub(reduce_quantity)
-                .expect("Attempt to cancel a larger quantity than available"),
-        };
-        let corr_reduce_quantity = state_order.quantity.try_sub(new_quantity).unwrap();
+        let cancel_quantity = state_order
+            .quantity
+            .try_sub(new_quantity)
+            .expect("Attempt to increase quantity");
```
afck, line 592: Feels like we should support that, too.

Instead of distinguishing between deleting, reducing and adding an order, why not just have a single operation that sets the quantity for a given price? (Just a thought; not something for this PR!)

examples/matching-engine/src/contract.rs
```diff
@@ -178,35 +218,105 @@ impl MatchingEngineContract {
...
-                self.send_to(transfer);
+
+                // Send removal notifications for filled orders
+                let matching_engine_chain = self.runtime.chain_id();
+                for (filled_chain_id, filled_owner, filled_order_id) in filled_orders {
+                    if filled_chain_id != matching_engine_chain {
+                        let removal_message = Message::OrderUpdated {
+                            owner: filled_owner,
+                            order_id: filled_order_id,
+                            new_quantity: Amount::ZERO,
```
afck, line 236: What about partially filled orders?

examples/matching-engine/src/lib.rs
```diff
@@ -136,23 +136,19 @@ scalar!(OrderNature);
...
         nature: OrderNature,
         price: Price,
     },
-    /// Cancelling of an order
-    Cancel {
-        owner: AccountOwner,
-        order_id: OrderId,
-    },
-    /// Modifying order (only decreasing is allowed)
+    /// Modify an order. The quantity can only be decreased. If nul, the order is
```
afck:
```suggestion
    /// Modify an order. The quantity can only be decreased. If zero, the order is
```
(Or "null"?)

examples/matching-engine/src/lib.rs
```diff
@@ -307,19 +320,23 @@ mod tests {
...
             owner,
             order_id: 1,
-            reduce_quantity: Amount::from_attos(199),
+            new_quantity: Amount::from_attos(199),
         };
         assert!(modify_invalid.check_precision(price_decimals).is_err());
 
-        // Test Cancel order (should always succeed)
-        let cancel_order = Order::Cancel { owner, order_id: 1 };
+        // Test order cancellation.
```
afck, line 334: (I still think this whole test is not really carrying its weight: It just tests different methods that all delegate to the same, simple, precision check, regardless of the order type.)

examples/matching-engine/src/state.rs
```diff
@@ -72,6 +77,11 @@ pub struct LevelView {
 #[derive(RootView, SimpleObject)]
 #[view(context = Context)]
 pub struct MatchingEngineState {
+    /// Pending orders tracked on user chains (not used on matching engine chain).
+    /// Maps owner to a map of order_id to pending order details.
```
afck:
```suggestion
    /// Maps owner to a map of order ID to pending order details.
```

examples/matching-engine/src/state.rs
```diff
@@ -86,7 +96,7 @@ pub struct MatchingEngineState {
     /// fundamental information on the order (price, nature, account)
     pub orders: MapView<OrderId, KeyBook>,
     /// The map giving for each account owner the set of order_id
-    /// owned by that owner.
+    /// owned by that owner (used on the matching engine chain).
```
afck, line 99:
```suggestion
    /// The map giving for each account owner the set of order IDs
    /// owned by that owner.
```

examples/matching-engine/src/state.rs
```diff
@@ -240,20 +250,29 @@ impl MatchingEngineState {
     /// * That clearing creates a number of transfer orders.
     /// * If after the level clearing the order is completely filled then it is not
     ///   inserted. Otherwise, it became a liquidity order in the matching engine
+    ///
+    /// Returns: (transfers, order_id, remaining_amount, filled_orders)
+    /// where filled_orders contains (chain_id, owner, order_id) for each filled order
```
afck:
```suggestion
    /// Returns: `(transfers, order_id, remaining_amount, filled_orders)`
    /// where `filled_orders` contains `(chain_id, owner, order_id)` for each filled order
```

examples/matching-engine/src/state.rs
```diff
@@ -284,6 +303,17 @@ impl MatchingEngineState {
                             .remove_entry(&price_ask)
                             .expect("Failed to remove ask level");
                     }
+                    // Collect filled orders with chain_id information
```
afck:
```suggestion
                    // Collect filled orders with chain ID information
```

examples/matching-engine/src/state.rs
```diff
@@ -330,6 +360,17 @@ impl MatchingEngineState {
                             .remove_entry(&price_bid)
                             .expect("Failed to remove bid level");
                     }
+                    // Collect filled orders with chain_id information
```
afck:
```suggestion
                    // Collect filled orders with chain ID information
```

afck (commented):
@MathieuDutSik: I think what you're suggesting would create a lot more overhead: Each user would have to read _all_ events, rather than get notified about only the messages that concern their own orders.

MathieuDutSik:
> @MathieuDutSik: I think what you're suggesting would create a lot more overhead: Each user would have to read _all_ events, rather than get notified about only the messages that concern their own orders.

Well,
It depends on what we want to achieve:
* Typically, when doing transactions, there aren't really that many transactions. The 1M achievement is just that, an achievement that has little relation to real usage. There are about 60K visa card transactions per second. So, the 1M is a little irrelevant. What is more important for payment is the latency.
* So publishing all the transactions and reading them does not strike me as a bad idea.
* The inconvenience of accessing `pending_orders` is that we have to parse the view and the `BTreeMap`. But yes, the managing of the orders is done by the contract.

afck:
Sure, I'm not against _also_ publishing all transactions in an event stream. I agree that could be useful, and I didn't mean to suggest that the overhead for _the network_ would be large.

I just mean that for an individual user making an order it makes sense to me that there should be a low-overhead way to know their order status. This implementation provides that using a message. Subscribing to and filtering the event stream would be a lot more work for the client.

examples/matching-engine/src/state.rs
ma2bd: that's what we do in this PR!

examples/matching-engine/src/contract.rs
ma2bd: line 253 below

ma2bd: Do you mean orders that were already in the book? Apparently we do not support this?

ma2bd:
> This design of keeping track of the order structure inside of the Matching Engine is not the standard way that this is done.

I don't understand this comment
1) We're only modifying the state of the application in user chains (acting as a client)
2) We're not tracking new information

examples/matching-engine/src/state.rs
afck: Are we? I can see that it's being unified with _canceling_ an order, but not with _creating_ an order?

examples/matching-engine/src/contract.rs
afck: I think `insert_and_uncross_market` _does_ partially fill orders. E.g. if there's a single bid in the book with an amount of 10, and I make an ask with an amount of 5 and the same price, my ask will be filled, and the bid will be reduced by 5 but remain in the book.

In that case we should probably send a message to the bidder as well, to notify them that their order was partially filled, i.e. modified?

examples/matching-engine/src/state.rs
ma2bd: Order creation is different because you're adding liquidity (instead of removing it). Concretely,
* you need to bring in funds (or tokens)
* you may need to "uncross" the market

examples/matching-engine/src/contract.rs
ma2bd: yes you're correct. We're missing some logic here

PR #4975, opened 2025-11-16
Implement asynchronous iterators for reading data in the storage.

### LINKED ISSUE #2742: try_load_all_entries should return iterator over batches of entries
[`try_load_all_entries`](https://github.com/linera-io/linera-protocol/blob/main/linera-views/src/views/reentrant_collection_view.rs#L684) returns _all_ entries at once. In extreme cases this can lead to an OOM. And even in less extereme situations, when amount of entries returned is too many (to fit into a block for example, or loaded into a smart contract context) can lead to suboptimal performance.

`try_load_all_entries` should return an iterator over batches of reasonable size (TBD what this size is, probably context-dependent) and have the caller decide whether more entries needs loading.

### LINKED ISSUE #4831: Implement `read_multi_values_iter` in `ReadableKeyValueStore`.
We have the following `get` loop in the code
```rust
    async fn get_preprocessed_block_hashes(
        &self,
        start: BlockHeight,
	end: BlockHeight,
    ) -> Result<Vec<CryptoHash>, WorkerError> {
        let mut hashes = Vec::new();
        let mut height = start;
        while height < end {
            match self.chain.preprocessed_blocks.get(&height).await? {
                Some(hash) => hashes.push(hash),
                None => break,
            }
            height = height.try_add_one()?;
        }
        Ok(hashes)
    }
```

A possible improvement would be something like:
```rust
let mut hashes = Vec::new();
for hash in self.chain.preprocessed_blocks.multi_get(&start..end).await? {
  match hash {
    Some(hash) => hashes.push(hash),
    None => break,
  }
}
Ok(hashes)
```

However, if the `None` is found on the first entry, then this is not better. What we need is an async iterator that accesses the entries one by one. The construction would then be:
```rust
   let mut hashes = Vec::new();
   let values = self.chain.preprocessed_blocs.multi_get_iter(&start..end)?;
   while let Some(hash) = values.next().await {
      match hash? {
        Some(hash) => hashes.push(hash),
        None => break,
      }
   }
   Ok(hashes)
```

However, in order to implement such a feature, we need to have access to the keys iteratively. That would require implementing something like
```rust
  fn read_multi_values_bytes(&self, keys: Vec<Vec<u8>>) -> Result<Self::ValueIterator, Self::Error>
```

Where `Self::ValueItewrator` is the async iterator implementing `async fn next(&self) -> Option<Option<Vec<u8>>`.

Note that the implementation in `ScyllaDb` and `DynamoDb` is already using many green threads that can be organized in a stream and so provide the needed functionality.

MathieuDutSik:
## Motivation

Accessing data at once can be suboptimal when reading data.
The solution to that is to implement asynchronous iterators.

Fixes https://github.com/linera-io/linera-protocol/issues/4831
Fixes https://github.com/linera-io/linera-protocol/issues/2742

## Proposal

The implementation takes the keys by value (that is `Vec<Vec<u8>>`).
We cannot use references because when we create specific keys in
lru-caching and value-splitting, we would need to have a self-reference
which is not allowed.

The implementation by container is then as follows:
* For DynamoDB and ScyllaDB, the downloading of data is done by block of 99 or 40 keys. So, the implementation of the iterator is straightforward to do.
* For RocksDB / LineraStorageService, we download by blocks of 50 keys.This is somewhat artificial. But better than downloading all the data.
* For lru-caching, value-splitting, we predetermine the needed keys. But when needed, sometimes we need to access keys (keys that were removed from the cache or additional segments) and so we access to the storage.

The implementation in the views is done for two cases where we have a use case in the code (the two issues in question):
* Access to multiple keys in `MapView`.
* Access to multiple views in `try_load_all_entries`.

For those two use cases, the loop can end prematurely, justifying the use of iterators.
It is unsure whether we can really reduce memory usage 

## Test Plan

The CI.

## Release Plan

Those changes could be backported to TestNet Conway.

## Links

None.

### DIFF
--- a/linera-core/src/chain_worker/state.rs
+++ b/linera-core/src/chain_worker/state.rs
@@ -394,17 +394,28 @@ where
 
         let (missing_indices, missing_blob_ids) = missing_indices_blob_ids(&maybe_blobs);
         if !missing_indices.is_empty() {
-            let all_entries_pending_blobs = self
+            let mut remaining = missing_indices
+                .into_iter()
+                .zip(missing_blob_ids)
+                .collect::<Vec<_>>();
+            let mut iter = self
                 .chain
                 .pending_proposed_blobs
-                .try_load_all_entries()
+                .try_load_all_entries_iter()
                 .await?;
-            for (index, blob_id) in missing_indices.into_iter().zip(missing_blob_ids) {
-                for (_, pending_blobs) in &all_entries_pending_blobs {
-                    if let Some(blob) = pending_blobs.get(&blob_id).await? {
-                        maybe_blobs[index].1 = Some(blob);
-                        break;
+            while let Some((_, pending_blobs)) = iter.next().await? {
+                let mut i = 0;
+                while i < remaining.len() {
+                    let (index, blob_id) = &remaining[i];
+                    if let Some(blob) = pending_blobs.get(blob_id).await? {
+                        maybe_blobs[*index].1 = Some(blob);
+                        remaining.swap_remove(i);
+                    } else {
+                        i += 1;
                     }
+                }
+                if remaining.is_empty() {
+                    break;
                 }
             }
         }
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -40,7 +40,7 @@ use linera_chain::{
     ChainError,
 };
 use linera_execution::{committee::Committee, ExecutionError};
-use linera_storage::{Clock as _, Storage as _};
+use linera_storage::{Clock as _, ResultReadCertificates, Storage as _};
 use rand::seq::SliceRandom;
 use received_log::ReceivedLogs;
 use serde::{Deserialize, Serialize};
@@ -559,24 +559,21 @@ impl<Env: Environment> Client<Env> {
             .local_node
             .get_preprocessed_block_hashes(chain_id, next_height, end)
             .await?;
-        last_info = Box::pin(async {
-            let certificates = self.storage_client().read_certificates(&hashes).await?;
-            let mut info = last_info;
-            // Missing certificates are skipped; they will be downloaded from the remote node below.
-            for certificate in certificates.into_iter().flatten() {
-                if let Some(until) = until_block_time {
-                    if certificate.value().block().header.timestamp >= until {
-                        break;
-                    }
-                }
-                info = self
-                    .handle_certificate(Arc::unwrap_or_clone(certificate))
-                    .await?
-                    .info;
+        let certificates = self.storage_client().read_certificates(&hashes).await?;
+        let certificates = match ResultReadCertificates::new(certificates, hashes) {
+            ResultReadCertificates::Certificates(certificates) => certificates,
+            ResultReadCertificates::InvalidHashes(hashes) => {
+                return Err(chain_client::Error::ReadCertificatesError(hashes))
             }
-            Ok::<_, chain_client::Error>(info)
-        })
-        .await?;
+        };
+        for certificate in certificates {
+            if let Some(until) = until_block_time {
+                if certificate.value().block().header.timestamp >= until {
+                    break;
+                }
+            }
+            last_info = self.handle_certificate(certificate).await?.info;
+        }
         Ok(last_info)
     }
 
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -40,7 +40,7 @@ use linera_chain::{
     ChainError,
 };
 use linera_execution::{committee::Committee, ExecutionError};
-use linera_storage::{Clock as _, ResultReadCertificates, Storage as _};
+use linera_storage::{Clock as _, Storage as _};
 use rand::seq::SliceRandom;
 use received_log::ReceivedLogs;
 use serde::{Deserialize, Serialize};
@@ -365,10 +365,6 @@ impl<Env: Environment> Client<Env> {
         self.environment.network()
     }
 
-    pub(crate) fn options(&self) -> &chain_client::Options {
-        &self.options
-    }
-
     /// Handles any pending local cross-chain requests, notifying subscribers.
     pub async fn retry_pending_cross_chain_requests(
         &self,
@@ -563,21 +559,24 @@ impl<Env: Environment> Client<Env> {
             .local_node
             .get_preprocessed_block_hashes(chain_id, next_height, end)
             .await?;
-        let certificates = self.storage_client().read_certificates(&hashes).await?;
-        let certificates = match ResultReadCertificates::new(certificates, hashes) {
-            ResultReadCertificates::Certificates(certificates) => certificates,
-            ResultReadCertificates::InvalidHashes(hashes) => {
-                return Err(chain_client::Error::ReadCertificatesError(hashes))
-            }
-        };
-        for certificate in certificates {
-            if let Some(until) = until_block_time {
-                if certificate.value().block().header.timestamp >= until {
-                    break;
+        last_info = Box::pin(async {
+            let certificates = self.storage_client().read_certificates(&hashes).await?;
+            let mut info = last_info;
+            // Missing certificates are skipped; they will be downloaded from the remote node below.
+            for certificate in certificates.into_iter().flatten() {
+                if let Some(until) = until_block_time {
+                    if certificate.value().block().header.timestamp >= until {
+                        break;
+                    }
                 }
+                info = self
+                    .handle_certificate(Arc::unwrap_or_clone(certificate))
+                    .await?
+                    .info;
             }
-            last_info = self.handle_certificate(certificate).await?.info;
-        }
+            Ok::<_, chain_client::Error>(info)
+        })
+        .await?;
         Ok(last_info)
     }
 
--- a/linera-core/src/updater.rs
+++ b/linera-core/src/updater.rs
@@ -770,17 +770,17 @@ where
             return Ok(info);
         }
 
-        let batch_size = self.client.options().certificate_upload_batch_size as usize;
-        for chunk in heights.chunks(batch_size) {
-            let certificates = self
-                .read_certificates_for_heights(chain_id, chunk.to_vec())
-                .await?;
-
-            for certificate in certificates {
-                self.send_confirmed_certificate(&certificate, delivery)
+        // Send any additional missing certificates in order
+        Box::pin(async {
+            let storage = self.client.local_node.storage_client();
+            let mut stream = storage.read_certificates_by_heights_iter(chain_id, heights);
+            while let Some(certificate) = stream.next().await {
+                self.send_confirmed_certificate(&certificate?, delivery)
                     .await?;
             }
-        }
+            Ok::<_, chain_client::Error>(())
+        })
+        .await?;
 
         Ok(info)
     }
@@ -947,28 +947,18 @@ where
             .into_iter()
             .map(|(chain_id, heights)| {
                 let mut updater = self.clone();
-                async move {
-                    // Get all block hashes for this chain at the specified heights in one call
-                    let heights_vec: Vec<_> = heights.into_iter().collect();
-                    let certificates = updater
-                        .client
-                        .local_node
-                        .storage_client()
-                        .read_certificates_by_heights(chain_id, &heights_vec)
-                        .await?
-                        .into_iter()
-                        .flatten()
-                        .collect::<Vec<_>>();
-
-                    // Send each certificate
-                    for certificate in certificates {
+                Box::pin(async move {
+                    let heights_vec = heights.into_iter().collect::<Vec<_>>();
+                    let storage = updater.client.local_node.storage_client();
+                    let mut stream =
+                        storage.read_certificates_by_heights_iter(chain_id, heights_vec);
+                    while let Some(certificate) = stream.next().await {
                         updater
-                            .send_confirmed_certificate(&certificate, delivery)
+                            .send_confirmed_certificate(&certificate?, delivery)
                             .await?;
                     }
-
                     Ok::<_, chain_client::Error>(())
-                }
+                })
             })
             .collect::<FuturesUnordered<_>>()
             .try_collect::<Vec<_>>()
--- a/linera-core/src/updater.rs
+++ b/linera-core/src/updater.rs
@@ -778,11 +778,9 @@ where
                 self.send_confirmed_certificate(&certificate?, delivery)
                     .await?;
             }
-            Ok::<_, chain_client::Error>(())
+            Ok::<_, chain_client::Error>(info)
         })
-        .await?;
-
-        Ok(info)
+        .await
     }
 
     /// Reads certificates for the given heights from storage.
--- a/linera-storage-service/src/client.rs
+++ b/linera-storage-service/src/client.rs
@@ -19,7 +19,9 @@ use linera_views::store::TestKeyValueDatabase;
 use linera_views::{
     batch::{Batch, WriteOperation},
     lru_caching::LruCachingDatabase,
-    store::{KeyValueDatabase, ReadableKeyValueStore, WithError, WritableKeyValueStore},
+    store::{
+        KeyValueDatabase, ReadValueStream, ReadableKeyValueStore, WithError, WritableKeyValueStore,
+    },
 };
 use serde::de::DeserializeOwned;
 use tonic::transport::{Channel, Endpoint};
@@ -44,6 +46,9 @@ use crate::{
 
 // The maximum key size is set to 1M rather arbitrarily.
 const MAX_KEY_SIZE: usize = 1000000;
+
+// The batch size in `read_multi_values_bytes_iter`.
+const BATCH_SIZE: usize = 50;
 
 // The shared store client.
 // * Interior mutability is required for the client because
@@ -206,6 +211,19 @@ impl ReadableKeyValueStore for StorageServiceStoreInternal {
         } else {
             self.read_entries(message_index, num_chunks).await
         }
+    }
+
+    fn read_multi_values_bytes_iter(&self, keys: Vec<Vec<u8>>) -> ReadValueStream<'_, Self::Error> {
+        let store = self.clone();
+
+        Box::pin(async_stream::try_stream! {
+            for chunk in keys.chunks(BATCH_SIZE) {
+                let values = store.read_multi_values_bytes(chunk).await?;
+                for value in values {
+                    yield value;
+                }
+            }
+        })
     }
 
     async fn find_keys_by_prefix(
--- a/linera-storage/src/db_storage.rs
+++ b/linera-storage/src/db_storage.rs
@@ -8,6 +8,7 @@ use std::{
 };
 
 use async_trait::async_trait;
+use futures::stream::StreamExt;
 #[cfg(with_metrics)]
 use linera_base::prometheus_util::MeasureLatency as _;
 use linera_base::{
@@ -43,7 +44,7 @@ use {
     std::cmp::Reverse,
 };
 
-use crate::{ChainRuntimeContext, Clock, Storage};
+use crate::{ChainRuntimeContext, Clock, Storage, StorageStream};
 
 #[cfg(with_metrics)]
 pub mod metrics {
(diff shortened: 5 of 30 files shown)


linera-views/src/store.rs
```diff
@@ -38,12 +38,25 @@ pub trait WithError {
...
 }
 
+/// Trait for async iterator over multiple values.
+#[cfg_attr(not(web), trait_variant::make(Send))]
+pub trait ReadMultiIterator<E: KeyValueStoreError> {
+    /// Returns the next value from the iterator.
+    /// Returns `Ok(None)` when iteration is complete.
+    /// Returns `Ok(Some(None))` when a key doesn't exist.
+    /// Returns `Ok(Some(Some(value)))` when a key exists with a value.
+    async fn next(&mut self) -> Result<Option<Option<Vec<u8>>>, E>;
```
afck: Why not use `Stream`?

linera-views/src/store.rs
```diff
@@ -38,12 +38,25 @@ pub trait WithError {
     type Error: KeyValueStoreError;
 }
 
+/// Trait for async iterator over multiple values.
+#[cfg_attr(not(web), trait_variant::make(Send))]
+pub trait ReadMultiIterator<E: KeyValueStoreError> {
+    /// Returns the next value from the iterator.
+    /// Returns `Ok(None)` when iteration is complete.
+    /// Returns `Ok(Some(None))` when a key doesn't exist.
```
deuszx: Why not the reverse? i.e. when key doesn't exist: 
- return `Ok(None)` when key doesn't exist. This is clearer b/c key doesn't exist form the first element. while `Ok(Some(None))` could be returned for the subsequent calls to the iterator. 
- `Ok(Some(None))` when there are no more elements in the stream.

linera-views/src/lru_prefix_cache.rs
```diff
@@ -820,6 +820,28 @@ impl LruPrefixCache {
...
 
+    /// Returns whether the cache can decide if a key is present or not.
+    pub(crate) fn test_key_presence(&self, key: &[u8]) -> bool {
+        // First, query the value map
+        let result = match self.value_map.get(key) {
+            None => false,
+            Some(entry) => match entry {
+                ValueEntry::DoesNotExist => false,
+                ValueEntry::Exists => false,
+                ValueEntry::Value(_) => true,
```
deuszx: maybe just return early here `return true`, other variants no-op. You don't need the intermediate `result` value.

linera-views/src/lru_prefix_cache.rs
```diff
@@ -820,6 +820,28 @@ impl LruPrefixCache {
...
+            },
+        };
+        if result {
+            return true;
+        }
+        if self.has_exclusive_access {
```
deuszx, line 829: Why for `self.has_exclusive_access == false` we claim that key is not present?

linera-views/src/store.rs
MathieuDutSik: No.
The function `fn next` returns a `Option<Item>` with `Item` being what is being produced. Here what is being produced is `Option<Vec<u8>>`.

linera-views/src/lru_prefix_cache.rs
MathieuDutSik: If `!self.has_exclusive_access` then that array is empty so there is no point in checking in it.

linera-views/src/store.rs
deuszx: if so, then why this signature isn't `Option<Item>` ? Where `Item` is a type member to be defined by the implementation.

linera-views/src/lru_prefix_cache.rs
afck: The whole function is basically:

```rust
matches!(self.value_map.get(key), Some(ValueEntry::Value(_)))
    || (self.has_exclusive_access && self.get_existing_find_key_values_entry(key).is_some())
```

(But I'm not against a more verbose version, if you find it more readable.)

MathieuDutSik: Yes, you are completely right.
Corrected.

linera-views/src/store.rs
MathieuDutSik: Here `Item = Option<Vec<u8>>` which gets you `Option<Option<Vec<u8>>>`.

deuszx: I understand. I mean that from `Option<Option<Vec<u8>>` you can't tell why you chose the semantics described in the comment wheras with `Option<Item>` it is.

~~Actually, even with `Option<Item>` and `type Item = Option<Vec<u8>>` it still makes more sense for `None` to mean the key doesn't exist, rather than `Some(None)` that you're proposing.~~

linera-views/src/store.rs
afck: I.e. should `ReadMultiIterator<E: KeyValueStoreError>` just have no methods, and instead have `impl`s and bounds so that it is equivalent to `Stream<Item = Result<Option<Vec<u8>>, E> + MaybeSend` where `MaybeSend` is `Send` iff `not(web)`?

linera-views/tests/store_tests.rs
```diff
@@ -36,7 +36,7 @@ async fn test_read_multi_values_memory() {
 async fn test_read_multi_values_dynamo_db() {
     use linera_views::dynamo_db::DynamoDbDatabase;
     let config = DynamoDbDatabase::new_test_config().await.unwrap();
-    big_read_multi_values::<DynamoDbDatabase>(config, 22000000, 1000).await;
+    big_read_multi_values::<DynamoDbDatabase>(config, 22000000, 200).await;
```
afck, line 39: Why? Is there a regression with 1000?

linera-views/src/views/key_value_store_view.rs
```diff
@@ -1283,6 +1283,19 @@ impl<C: Context> ReadableKeyValueStore for ViewContainer<C> {
         Ok(view.multi_get(keys).await?)
     }
 
+    fn read_multi_values_bytes_iter(
+        &self,
+        keys: Vec<Vec<u8>>,
+    ) -> impl futures::stream::Stream<Item = Result<Option<Vec<u8>>, ViewContainerError>> {
+        let store = self.clone();
+        async_stream::stream! {
+            let values = store.read_multi_values_bytes(&keys).await?;
```
afck: Why not `store.read_multi_value_bytes_iter(&keys)`?

linera-views/src/test_utils/mod.rs
```diff
@@ -235,6 +236,17 @@ pub async fn run_reads<S: KeyValueStore>(store: S, key_values: Vec<(Vec<u8>, Vec
...
+        // Test read_multi_values_bytes_iter
+        let mut iter = Box::pin(store.read_multi_values_bytes_iter(keys));
+        let mut values_from_iter = Vec::new();
+        loop {
+            match iter.next().await {
+                None => break,
+                Some(Ok(value)) => values_from_iter.push(value),
+                Some(Err(e)) => panic!("Error reading from iterator: {:?}", e),
+            }
+        }
```
afck: Maybe more concise:

```suggestion
        let values_from_iter = iter
            .collect::<Result<Vec<_>>>()
            .await
            .expect("reading from iterator"));
```

linera-views/src/backends/value_splitting.rs
```diff
@@ -202,6 +203,84 @@ where
         Ok(big_values)
     }
 
+    fn read_multi_values_bytes_iter(
+        &self,
+        keys: Vec<Vec<u8>>,
```
afck: I wonder if it would make sense (maybe not in this PR) to also take a stream as an input. (Or just a regular iterator? Needs discussion…) That way, e.g. we wouldn't even put together the `big_key` for the 10000th element if the caller cancels the stream after the 5000th.

linera-views/src/backends/value_splitting.rs
```diff
@@ -202,6 +203,84 @@ where
...
+        let store = self.store.clone();
+        let mut first_segments_stream = Box::pin(self.store.read_multi_values_bytes_iter(big_keys));
+
+        async_stream::stream! {
+            for key in keys {
+                // Get the next value from the first segments stream
+                let value_result = first_segments_stream.next().await
+                    .expect("keys and first_segments should have same length");
+
+                let value = match value_result {
```
afck: We should use `try_stream!` instead of `stream!`. Then this line could probably be `let value = value_result?;`, or `let value = value_result.map_err(ValueSplittingError::InnerStoreError)?` or something like that.

linera-views/src/backends/scylla_db.rs
```diff
@@ -673,6 +673,32 @@ impl ReadableKeyValueStore for ScyllaDbStoreInternal {
...
     }
 
+    fn read_multi_values_bytes_iter(
+        &self,
+        keys: Vec<Vec<u8>>,
+    ) -> impl Stream<Item = Result<Option<Vec<u8>>, Self::Error>> {
+        let batches = keys
+            .chunks(MAX_MULTI_KEYS)
+            .map(|chunk| chunk.to_vec())
+            .collect::<Vec<_>>();
```
afck, line 665:
```suggestion
            .map(|chunk| chunk.to_vec());
```

linera-views/src/backends/rocks_db.rs
```diff
@@ -491,6 +495,39 @@ impl ReadableKeyValueStore for RocksDbStoreInternal {
...
+        let spawn_mode = self.spawn_mode;
+
+        async_stream::stream! {
+            let mut current_position = 0;
+            while current_position < keys.len() {
+                // Calculate the end position for this batch
+                let end_position = std::cmp::min(current_position + BATCH_SIZE, keys.len());
+
+                // Extract the next batch of keys
+                let batch_keys = keys[current_position..end_position].to_vec();
```
afck: Why not use `chunks` here, too, like in the ScyllaDB case?

linera-views/src/backends/lru_caching.rs
```diff
@@ -293,6 +294,72 @@ where
...
+                            let mut cache = cache.lock().unwrap();
+                            cache.query_read_value(key)
+                        };
+                        if let Some(value) = cached_value {
+                            #[cfg(with_metrics)]
+                            metrics::READ_VALUE_CACHE_HIT_COUNT.with_label_values(&[]).inc();
+                            value
+                        } else {
+                            // The key has been evicted. Should be rare.
+                            store.read_value_bytes(key).await?
```
afck: (Ah, that works? Then I'm not sure I understand the difference between `stream` and `try_stream`. Maybe just that in the latter case you write `yield a` instead of `yield Ok(a)`?)

linera-views/src/views/key_value_store_view.rs
MathieuDutSik: Well, we could, but anyway, `ViewContainer` is used only for tests so that does not matter very much I think.

linera-views/src/test_utils/mod.rs
MathieuDutSik: Yes, we can replace by
```rust
    let values_from_iter = store
        .read_multi_values_bytes_iter(keys)
        .try_collect::<Vec<_>>()
        .await
        .expect("Error reading from iterator");
```

linera-views/src/backends/value_splitting.rs
MathieuDutSik: We potentially could have those everywhere. But I want to have a limited change here.

linera-views/src/backends/scylla_db.rs
MathieuDutSik: We need to collect the entries in order to be able to iterate on them. Removing the collect makes the code not compile.

afck: That's just a borrow checker issue. This compiles, because it makes the stream own `keys`, and it avoids the allocation:

```diff
-        let batches = keys
-            .chunks(MAX_MULTI_KEYS)
-            .map(|chunk| chunk.to_vec())
-            .collect::<Vec<_>>();
         let store = self.clone();
 
         async_stream::stream! {
-            for batch_keys in batches {
+            for chunk in keys.chunks(MAX_MULTI_KEYS) {
                 let vals = {
+                    let batch_keys = chunk.to_vec();
```

linera-views/src/backends/rocks_db.rs
MathieuDutSik: Yes, we can use the terminolology chunk systematically.

linera-views/src/backends/value_splitting.rs
MathieuDutSik: Yes, we can do that.
But no, we do not need the `try_stream`.

linera-views/tests/store_tests.rs
MathieuDutSik: With the additional test, it becomes slower.
So, I reduce it in order to get a meaningful test that can be run locally.
I do not think it matters since the DynamoDb is not very much used.

linera-views/src/backends/lru_caching.rs
MathieuDutSik: Yes, that works. I do not need `try_stream`.

linera-views/src/backends/rocks_db.rs
afck: I meant the `chunks` method, so that you don't have to keep track of the positions manually here. (Or maybe there's a reason for it?)

linera-views/src/store.rs
Twey: If you write `-> impl Stream<…>` then `trait_variant::make` will do the `MaybeSend` part for you.

ma2bd:
converting stale PR to draft

linera-views/src/store.rs
```diff
@@ -65,6 +66,24 @@ pub trait ReadableKeyValueStore: WithError {
...
 
+    /// Returns a stream for reading multiple values using the provided `keys`.
+    /// The stream yields `Result<Option<Vec<u8>>, Self::Error>` where:
+    /// - `Ok(None)` when a key doesn't exist
+    /// - `Ok(Some(value))` when a key exists with a value
+    /// - `Err(e)` on error
+    /// The default implementation fetches all values at once and yields them one by one.
+    fn read_multi_values_bytes_iter(
+        &self,
+        keys: Vec<Vec<u8>>,
```
afck: Should `keys` also be a stream?

(Ah, you commented earlier about keeping the change small… but this is a newly added method anyway?)

linera-core/src/chain_worker/state.rs
```diff
@@ -293,18 +293,26 @@ where
 
         let (missing_indices, missing_blob_ids) = missing_indices_blob_ids(&maybe_blobs);
         if !missing_indices.is_empty() {
-            let all_entries_pending_blobs = self
+            let mut remaining = missing_indices.into_iter().zip(missing_blob_ids).collect::<Vec<_>>();
```
afck: I think it would be way easier to make this a map whose keys are the blob IDs and values are sets (or vectors) of indices.

linera-core/src/chain_worker/state.rs
```diff
@@ -1019,14 +1027,24 @@ where
...
+        let mut heights = Vec::new();
         let mut height = start;
         while height < end {
-            match self.chain.preprocessed_blocks.get(&height).await? {
+            heights.push(height);
+            height = height.try_add_one()?;
+        }
+
+        // Build the multi_get_iter
+        let mut iter = self.chain.preprocessed_blocks.multi_get_iter(&heights)?;
```
afck: The old version of this function would often read just 0 or 1 blocks. Now it would read a whole batch that we often wouldn't use, because we only need the contiguous initial segment. I don't think this function is a good use case for the streams.

(And admittedly, we should rename it to `get_contiguous_preprocessed_block_hashes` or something, and clarify that in the comment.)

linera-views/src/store.rs
MathieuDutSik: Well of course we could have such a design.
But the question is: Do we have a good use case for that feature?
If yes, then yes that makes sense.

linera-core/src/chain_worker/state.rs
MathieuDutSik: Ok, let us revert the change to this function.

linera-core/src/updater.rs
```diff
@@ -945,22 +947,20 @@ where
...
-                    let certificates = updater
-                        .client
-                        .local_node
-                        .storage_client()
-                        .read_certificates_by_heights(chain_id, &heights_vec)
-                        .await?
-                        .into_iter()
-                        .flatten()
-                        .collect::<Vec<_>>();
+                    Box::pin(async {
```
afck: Is this necessary? Can we pin the outer `async` block instead, three lines above?

linera-storage-service/src/client.rs
```diff
@@ -45,6 +47,9 @@ use crate::{
 // The maximum key size is set to 1M rather arbitrarily.
 const MAX_KEY_SIZE: usize = 1000000;
 
+// The batch size in `read_multi_values_bytes_iter`.
+const BATCH_SIZE: usize = 50;
```
afck, line 51: In general we're trying to avoid hard-coded constants; maybe this could be a CLI parameter?

linera-storage-service/src/client.rs
```diff
@@ -208,6 +213,26 @@ impl ReadableKeyValueStore for StorageServiceStoreInternal {
...
+        let store = self.clone();
+
+        Box::pin(async_stream::stream! {
+            let mut current_position = 0;
+            while current_position < keys.len() {
+                // Calculate the end position for this batch
+                let end_position = std::cmp::min(current_position + BATCH_SIZE, keys.len());
+
+                // Extract the next batch of keys
+                let chunk = &keys[current_position..end_position];
```
afck: Isn't this just reimplementing `slice::chunks`?

linera-storage/src/db_storage.rs
```diff
@@ -915,6 +916,37 @@ where
             .collect()
     }
 
+    fn read_certificates_iter(
+        &self,
+        hashes: Vec<CryptoHash>,
+    ) -> StorageStream<'_, Result<Option<ConfirmedBlockCertificate>, ViewError>> {
+        Box::pin(async_stream::stream! {
+            let block_keys = get_block_keys();
+            for hash in &hashes {
```
afck: Should this be done concurrently for a few in advance? Otherwise this is just completely sequential, isn't it?

linera-storage-service/src/client.rs
MathieuDutSik: Correct.

linera-storage/src/db_storage.rs
MathieuDutSik: Correct, it does not give benefit over a completely sequential approach.
Remove it.

linera-storage-service/src/client.rs
MathieuDutSik: Yes, we can.
But is that worth it? After all, the `linera-storage-service` is only used for tests.

linera-core/src/updater.rs
MathieuDutSik: Yes, we can. Done.

linera-storage-service/src/client.rs
afck: (That's true; fine with me, not a blocker.)

linera-core/src/client/mod.rs
```diff
@@ -563,21 +560,24 @@ impl<Env: Environment> Client<Env> {
...
+                    .handle_certificate(Arc::unwrap_or_clone(certificate))
+                    .await?
+                    .info;
             }
-            last_info = self.handle_certificate(certificate).await?.info;
-        }
+            Ok::<_, chain_client::Error>(info)
+        })
+        .await?;
         Ok(last_info)
```
afck, line 577: Why not just return `Box::pin(…).await` directly?

linera-core/src/updater.rs
```diff
@@ -770,17 +770,17 @@ where
...
+            while let Some(certificate) = stream.next().await {
+                self.send_confirmed_certificate(&certificate?, delivery)
                     .await?;
             }
-        }
+            Ok::<_, chain_client::Error>(())
+        })
+        .await?;
 
         Ok(info)
```
afck: Same here.

linera-core/src/client/mod.rs
MathieuDutSik: That part of the code is actually a leftover since the change does not uses the _iter construction (no benefit so it was removed).
So reverted.

linera-core/src/updater.rs
MathieuDutSik: Yes, we can simplify it.

PR #4977, opened 2025-11-17
Use historical hashing for `ExecutionStateView`

bart-linera:
## Motivation

Hashing of all the data in `ExecutionStateView`s became a bottleneck in large chain states.

## Proposal

Replace the current hashing mechanism of `ExecutionStateView`s with historical hashing.

## Test Plan

CI will catch regressions

## Release Plan

## Links

### DIFF
--- a/linera-execution/src/execution.rs
+++ b/linera-execution/src/execution.rs
@@ -1,7 +1,11 @@
 // Copyright (c) Zefchain Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-use std::{collections::BTreeMap, vec};
+use std::{
+    collections::BTreeMap,
+    ops::{Deref, DerefMut},
+    vec,
+};
 
 use allocative::Allocative;
 use futures::{FutureExt, StreamExt};
@@ -15,8 +19,8 @@ use linera_views::{
     context::Context,
     historical_hash_wrapper::HistoricallyHashableView,
     key_value_store_view::KeyValueStoreView,
-    reentrant_collection_view::HistoricallyHashedReentrantCollectionView,
-    views::{ClonableView, Hasher as _, ReplaceContext, View},
+    reentrant_collection_view::ReentrantCollectionView,
+    views::{ClonableView, ReplaceContext, View},
     ViewError,
 };
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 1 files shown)


linera-execution/src/execution.rs
```diff
@@ -30,20 +32,37 @@ use {
...
 };
 
 /// A view accessing the execution state of a chain.
-#[derive(Debug, ClonableView, CryptoHashView, Allocative)]
+#[derive(Debug, ClonableView, View, Allocative)]
 #[allocative(bound = "C")]
 pub struct ExecutionStateView<C> {
     /// System application.
-    pub system: SystemExecutionStateView<C>,
+    pub system: HistoricallyHashableView<C, SystemExecutionStateView<C>>,
```
MathieuDutSik: That is one way of doing it.
But alternatively, we could nest the whole `ExecutionStateView` with a
```rust
HistoricallyHashableView<C,  ExecutionStateView<C>>
```

That would require some renaming along the way.

linera-execution/src/execution.rs
```diff
@@ -30,20 +32,37 @@ use {
...
+    C: Context + Clone + Send + Sync + 'static,
+    C::Extra: ExecutionRuntimeContext,
+{
+    pub async fn crypto_hash_mut(&mut self) -> Result<CryptoHash, ViewError> {
+        #[derive(Serialize, Deserialize)]
+        struct ExecutionStateViewHash([u8; 32]);
+        impl BcsHashable<'_> for ExecutionStateViewHash {}
+        let mut hasher = linera_views::sha3::Sha3_256::default();
+        hasher.update_with_bytes(&self.system.historical_hash().await?)?;
+        hasher.update_with_bytes(&self.users.historical_hash().await?)?;
```
afck: Actually, can we `let ExecutionStateView { … } = self`, so the compiler would complain if we added a field, and we can't forget it here?

linera-execution/src/execution.rs
bart-linera: I thought about it (renaming `ExecutionStateView` to something like `ExecutionStateViewInner` and wrapping that in a `HistoricallyHashableView`), but eventually decided this way was simpler. I could still try doing it the other way if we decide it makes more sense.

ma2bd: I was thinking `HistoricallyHashableView<C,  ExecutionStateView<C>>` as well TBH

ma2bd: so that crypto_hash_mut is trivial

bart-linera: It seems that the change as it is now might have exposed a bug in `linera_views`, so I'll try to get to the bottom of that before I reimplement this the other way.

ma2bd: Right. I may have mixed up a key in `pre_load` / `post_load`, creating a BCS-deserialization error when we try to deserialize an option value

bart-linera: Done in 27b629d

PR #4978, opened 2025-11-17
Simplify and optimize quorum calculation; fix blob bug.

afck:
## Motivation

The current `Committee` implementation is a bit complicated, and sometimes sets the quorum threshold higher than necessary.

## Proposal

The assumption is that $$N \geq 3 f + 1$$, where $$f$$ is the fault tolerance, i.e. the maximum total votes of the faulty validators. The _validity threshold_ is $$f + 1$$. Given $$N$$, the highest possible value for $$f$$ is therefore $$\lceil N / 3 \rceil - 1$$.

The _quorum threshold_ $$q$$ is minimal such that any two quorum intersect in at least a validity threshold (i.e. have at least one honest validator in common), i.e. $$2 q - N \geq f + 1$$. Thus $$q = \lceil (N + f + 1) / 2 \rceil$$.

In particular, if e.g. $$N = 3$$, then $$f = 0$$ and $$q = 2$$.

This change revealed an issue with the certificate handling logic in the worker: `committees_for` returns `ViewErrors` instead of `BlobsNotFound`/`EventsNotFound`, so the client fails to send the validators the admin chain if needed.

This was fixed; I also kept minor cleanups I made while debugging this. I got a few stack overflows, so I doubled the client's Tokio thread stack size.

## Test Plan

CI

## Release Plan

- The fixes (but maybe not the quorum change) should be backported to `testnet_conway` and released in a new SDK and validator hotfix.

## Links

### DIFF
--- a/linera-execution/src/system.rs
+++ b/linera-execution/src/system.rs
@@ -782,57 +782,18 @@ where
         self.description.set(Some(description));
         self.epoch.set(epoch);
 
-        let net_description = self
+        let committees = self
+            .context()
+            .extra()
+            .get_committees(min_active_epoch..=max_active_epoch)
+            .await?;
+        let admin_chain_id = self
             .context()
             .extra()
             .get_network_description()
             .await?
-            .ok_or(ExecutionError::NoNetworkDescriptionFound)?;
-        let admin_chain_id = net_description.admin_chain_id;
-        let mut committees = BTreeMap::new();
-
-        // Special case: the genesis epoch is stored in the NetworkDescription.
-        if min_active_epoch == Epoch::ZERO {
-            let blob_id = BlobId::new(
-                net_description.genesis_committee_blob_hash,
-                BlobType::Committee,
-            );
-            let committee_blob = self
-                .context()
-                .extra()
-                .get_blob(blob_id)
-                .await?
-                .ok_or(ExecutionError::BlobsNotFound(vec![blob_id]))?;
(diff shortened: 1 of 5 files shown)


linera-execution/src/system.rs
```diff
@@ -781,20 +781,61 @@ where
...
         self.epoch.set(epoch);
-        let committees = self
-            .context()
-            .extra()
-            .committees_for(min_active_epoch..=max_active_epoch)
-            .await?;
-        self.committees.set(committees);
-        let admin_id = self
+
+        let net_description = self
```
bart-linera: I feel like it would still be a good idea to have this committee reading logic extracted into a function somewhere, so that we don't have to duplicate all this code wherever we need to read committees... But I guess it's good enough for now, with just two places where we do that.

afck: Good point, I'll give it a try and deduplicate them.

afck: Done in https://github.com/linera-io/linera-protocol/pull/4978/commits/66c677391c5afe6e87fa10fe6b042efe1e021f89. I also addressed @deuszx's comments on #4986 in this commit.

It's a bit painful to convert the errors, and to collect all the missing event IDs and blob IDs, but I guess it makes sense.

bart-linera: Nice, I like your approach!

PR #4979, opened 2025-11-17
Add Postgres DB backend for the indexer.

deuszx:
## Motivation

Turns out our SQLite file for Testnet is quite big already – over 30 GB. We also learned that it might be useful to have a read-only access to the DB remotely, which is not possible with SQLite.

## Proposal

Add support for Postgres backend. Explorer frontend was also updated. So were docker scripts.

## Test Plan

Tested locally with all 3 options:
1. `--memory`
2. `--sqlite <path_to_file>`
3. `--postgres <local_instance>`

Integration tests were added for postgres.

## Release Plan

## Links

### DIFF
--- a/linera-service/src/exporter/runloops/validator_exporter.rs
+++ b/linera-service/src/exporter/runloops/validator_exporter.rs
@@ -118,17 +118,7 @@ where
                 .with_label_values(&[self.node.address()])
                 .set(receiver.len() as i64);
             match self.dispatch_block((*block).clone()).await {
-                Ok(_) => {
-                    if let Err(error) = self.upload_blobs(blobs_ids).await {
-                        tracing::error!(
-                            %error,
-                            block=?block.hash(),
-                            chain_id=?block.inner().chain_id(),
-                            "error when uploading blobs after dispatching block"
-                        );
-                        // We ignore the error and continue to the next block.
-                    }
-                }
+                Ok(_) => {}
 
                 Err(NodeError::BlobsNotFound(blobs_to_maybe_send)) => {
                     let blobs = blobs_ids


docker/Dockerfile.indexer
```diff
@@ -73,9 +73,12 @@ ENV GIT_COMMIT=${git_commit}
 ENV RUSTFLAGS=${rustflags}
 
 
-# Build linera CLI and block exporter with scylladb and metrics features
+# Build linera CLI, storage server, server, proxy, and block exporter with scylladb and metrics features
 RUN cargo build ${build_flag:+"$build_flag"} \
     --bin linera \
+    --bin linera-storage-server \
+    --bin linera-server \
+    --bin linera-proxy \
```
ndr-ds: Why are we building this in the indexer image? We have the main `Dockerfile` with this already.

linera-indexer/example/src/grpc_main.rs
```diff
@@ -29,15 +43,36 @@ async fn main() -> Result<(), IndexerError> {
...
-        .map_err(IndexerError::Other)?;
+    // Start server with the selected database backend
+    if args.memory {
+        tracing::info!("Starting indexer with in-memory SQLite database");
+        let database = SqliteDatabase::new("sqlite::memory:").await?;
+        let grpc_server = IndexerGrpcServer::new(database);
+        grpc_server
+            .serve(args.port)
+            .await
+            .map_err(IndexerError::Other)?;
```
ndr-ds, line 54: Could we call this just once in the bottom?

docker/explorer-entrypoint.sh
```diff
@@ -3,26 +3,26 @@ set -e
...
-echo "Waiting for indexer database..."
-while [ ! -f "${DB_PATH:-/data/indexer.db}" ]; do
-    echo "Database not found at ${DB_PATH:-/data/indexer.db}, waiting..."
-    sleep 2
-done
+# Log which database type is being used
+if [ -n "$DATABASE_URL" ]; then
+    echo "Using PostgreSQL database"
+else
+    echo "Using SQLite database at ${DB_PATH:-/data/indexer.db}"
```
ndr-ds, line 10: Don't we still need to wait for the file to exist here?

.github/workflows/indexer.yml
```diff
@@ -0,0 +1,65 @@
...
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+    - name: Check indexer library compiles
+      run: |
+        cargo check --locked -p linera-indexer
+    - name: Check indexer example compiles
+      run: |
+        cargo check --locked -p linera-indexer-example
+    - name: Run indexer unit tests
+      run: |
+        cargo test --locked -p linera-indexer --lib -- --skip db::postgres --skip db::sqlite
```
ndr-ds: Could we specify to include just db::memory here instead?

ndr-ds (changes requested):
Looks good overall. Back to you, as I have a few questions that could be blocking.  
Just FYI, the file is actually currently over 18GB already 😅

.github/workflows/indexer.yml
deuszx: This is about _skipping_ those tests (ie these are not arguments which DB use).

docker/Dockerfile.indexer
deuszx: I think it started differently – this image was slightly smaller than `Dockerfile` 🤔 I'll see if I can do it again.

deuszx: Actually, `Dockerfile` doesn't have `linera-storage-server` and `linera-exporter` – so half of the binaries I need here. Since it doesn't need them (docker-based validators don't run storage-service and that's needed for the exporter) I would prefer to NOT add it there. This makes these changes necessary.

docker/explorer-entrypoint.sh
deuszx: This is moved to `docker-compose`:
1. Explorer depends on indexer
2. indexer's healthcheck is availability of the gRPC endpoint and we can't have healthy gRPC service w/o the database.

.github/workflows/indexer.yml
deuszx: But we don't need to skip `db::sqlite` as they're fairly fast since they run in memory.

linera-indexer/example/src/grpc_main.rs
deuszx: Actually, we can't. `IndexerGrpcServer::new` expects `D` type – it doesn't expect `D: DatabaseTrait` – so all of these if branches return different type (`IndexerDatabase<Sqlite>`, `IndexerDatabase<Postgres>`).

.github/workflows/indexer.yml
deuszx: Done in [Fix indexer.yml](https://github.com/linera-io/linera-protocol/pull/4979/commits/419864456e42bb9be821e16c5f70b6c70bc9d255).

ndr-ds: I know it's skipping... I'm saying that being explicit about what you're including is generally better than specifying what you're excluding. Because a test can be added, and you'll just start running it automatically, even if it doesn't need to be run, if the person forgets to skip it here.
Can't you do something like:
```
cargo test --locked -p linera-indexer --lib -- db::memory
```

?

docker/Dockerfile.indexer
ndr-ds: Then add just the `storage-server` one here? The server and proxy binaries are huge. Please let's not add it here

linera-indexer/example/src/grpc_main.rs
ndr-ds: Maybe a helper function then? Something like this:
```
async fn serve_with_db<D: IndexerDatabase + 'static>(
    database: D, 
    port: u16
) -> Result<(), IndexerError> 
where D::Error: Into<ProcessingError> {
    let grpc_server = IndexerGrpcServer::new(database);
    grpc_server.serve(port).await.map_err(IndexerError::Other)
}
```

ndr-ds (changes requested):
Back to you for my replies

docker/Dockerfile.indexer
deuszx: That's why I don't want to add them there – they're not needed in the `Dockerfile` that is used externally and ARE needed here for testing the exporter-indexer stack. If you're using it in other deployments then there should exist a different Dockerfile – this one is for testing locally, not deployments.

linera-indexer/example/src/grpc_main.rs
deuszx: I prefer to leave as is.

.github/workflows/indexer.yml
deuszx: This version is better for the future as I don't want to add explicitly what needs to run every time I add new tests. I just know I don't want to run postgres-based ones.

docker/Dockerfile.indexer
ndr-ds: All I'm saying is to delete `linera-proxy` and `linera-server` from this `Dockerfile.indexer` file. Can't you just get those binaries from the `Dockerfile` instead? And what do you mean `Dockerfile.indexer` isn't supposed to be used in deployments? It currently is.

linera-indexer/example/src/grpc_main.rs
ndr-ds: Why? Generally we don't want to duplicate code, if we can

docker/Dockerfile.indexer
deuszx: >  And what do you mean Dockerfile.indexer isn't supposed to be used in deployments? It currently is.

I said it the first time you used it that way and I'm just repeating it again – it wasn't meant to be used during deployments. It's for testing the stack locally.

ndr-ds: If you don't test the stack locally with the same image, you have inconsistent behavior in your local tests

deuszx: Image is just a binary with some arguments, it shouldn't affect the correctness.

ndr-ds: No, the image defines the Docker container that things will run on, which will contain binaries, things installed, etc. Ideally the test environment should be as similar to production as possible, otherwise your local tests aren't really testing if production will be ok or not, which is the goal of all tests 😅
But ok, if you wanna keep these binaries with the argument that this is test only, then I would rename this dockerfile back to having a `-test` suffix, and create another one without the `linera-proxy` and `linera-server` binaries

deuszx: This image also has the indexer – surely you don't need it when deploying an exporter.

The reason all those binaries are bundled up in the same image is b/c our intra-crates dependencies are bit of a mess now and "everything needs everything" during build stage so separating it out into different Docker images doesn't speed things up.

deuszx: The goal of these docker files isn't to mimic the production deployment – b/c we don't know (and don't want to) whether it will be deployed as a docker or not at this stage. We just want an easy to reproduce setup to test the whole stack.

ndr-ds: I know for a fact that building `linera-proxy` and `linera-server` for this image (which is used by both the exporter and the indexer) is not needed, as that's the current state in production. So since this is a "test" image, we need a production image that doesn't unnecessarily links these binaries and pushes them to the Google's Artifact registry, for example.

ndr-ds: Yes I get it that you want this to be a test image with everything. I'm just asking you to create a production one then, in this case

deuszx: I don't know what's needed in production. If you tell me exactly how it should be configured I can do it.

deuszx: > I know for a fact that building `linera-proxy` and `linera-server` for this image (which is used by both the exporter and the indexer) is not needed, as that's the current state in production. So since this is a "test" image, we need a production image that doesn't unnecessarily links these binaries and pushes them to the Google's Artifact registry, for example.

Maybe it's not running but it's needed when _building_ – otherwise cargo complains that it's missing dependencies.

ndr-ds: It's just whatever this file had before your changes, with the web directory fix, I guess

deuszx: Done in [Split Dockerfiles](https://github.com/linera-io/linera-protocol/pull/4979/commits/8a19e56ae06e9f514404b8552195ef8287638d38)

linera-service/src/exporter/runloops/validator_exporter.rs
```diff
@@ -118,7 +118,17 @@ where
...
+                    if let Err(error) = self.upload_blobs(blobs_ids).await {
+                        tracing::error!(
+                            %error,
+                            block=?block.hash(),
+                            chain_id=?block.inner().chain_id(),
+                            "error when uploading blobs after dispatching block"
+                        );
+                        // We ignore the error and continue to the next block.
+                    }
+                }
```
ndr-ds: I'm assuming this is some bugfix that is intended to be included here

deuszx: Ah, sorry. I forgot to remove it. It's not needed (the `blobs_ids` are the ones from _inside_ the block) – it's not breaking anything but definitely not needed.

deuszx: Reverted these lines in [Don't send blobs that are already in the block](https://github.com/linera-io/linera-protocol/pull/4979/commits/21ff1fcecd6c55fb13fc27a8c265fe95b1d14500)

PR #4981, opened 2025-11-18
deployment: add support for the LRU cache

eldios:
## Motivation

This addition/backport of the LRU settings addition.
This is the original PR : https://github.com/linera-io/linera-protocol/pull/4964

## Test Plan

Changes were tested on an ephemeral temporary network successfully.

## Release Plan

- These changes ARE the backport and will follow the usual release cycle.

## Links

Original PR: https://github.com/linera-io/linera-protocol/pull/4964

### DIFF
--- a/kubernetes/linera-validator/templates/proxy.yaml
+++ b/kubernetes/linera-validator/templates/proxy.yaml
@@ -100,13 +100,7 @@ spec:
                 --storage scylladb:tcp:scylla-client.scylla.svc.cluster.local:9042 \
                 --storage-replication-factor {{ .Values.storageReplicationFactor | quote }} \
                 --storage-max-cache-size {{ .Values.storageCacheConfig.maxCacheSize | int }} \
-                --storage-max-value-entry-size {{ .Values.storageCacheConfig.maxValueEntrySize | int }} \
-                --storage-max-find-keys-entry-size {{ .Values.storageCacheConfig.maxFindKeysEntrySize | int }} \
-                --storage-max-find-key-values-entry-size {{ .Values.storageCacheConfig.maxFindKeyValuesEntrySize | int }} \
                 --storage-max-cache-entries {{ .Values.storageCacheConfig.maxCacheEntries | int }} \
-                --storage-max-cache-value-size {{ .Values.storageCacheConfig.maxCacheValueSize | int }} \
-                --storage-max-cache-find-keys-size {{ .Values.storageCacheConfig.maxCacheFindKeysSize | int }} \
-                --storage-max-cache-find-key-values-size {{ .Values.storageCacheConfig.maxCacheFindKeyValuesSize | int }} \
                 --id "$ORDINAL" \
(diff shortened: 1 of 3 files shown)


docker/docker-compose.indexer-test.yml
```diff
@@ -25,7 +25,17 @@ services:
...
+      --port ${INDEXER_PORT:-8081}
+      --database-path ${INDEXER_DATABASE_PATH:-/data/indexer.db}
+      --max-cache-size 10000000
+      --max-value-entry-size 1000000
+      --max-find-keys-entry-size 1000000
+      --max-find-key-values-entry-size 1000000
+      --max-cache-entries 1000
+      --max-cache-value-size 10000000
+      --max-cache-find-keys-size 10000000
+      --max-cache-find-key-values-size 10000000"
```
MathieuDutSik: Has this been tested?
The file is `linera-indexer/example/src/grpc_main.rs`.
And those options are not available there.

docker/docker-compose.yml
```diff
@@ -53,6 +53,22 @@ services:
...
+      - --storage-max-find-key-values-entry-size
+      - "1000000"
+      - --storage-max-cache-entries
+      - "1000"
+      - --storage-max-cache-value-size
+      - "10000000"
+      - --storage-max-cache-find-keys-size
+      - "10000000"
+      - --storage-max-cache-find-key-values-size
+      - "10000000"
```
MathieuDutSik: What is interesting is that those options already have default values.
However, the values are here. But I guess that makes sense to put those values in the script. Are those different from the default values?

Maybe put also input values for `send_timeout`, `recv_timeout` and others.

docker/docker-compose.yml
```diff
@@ -77,6 +93,22 @@ services:
...
```
MathieuDutSik: Same remarks obviously.

But, I have another point: For Kubernetes, we can store things in a variable in `storageCacheConfig`.
Isn't there something equivalent for docker-compose file.

MathieuDutSik (commented):
Thanks for inserting the data in the input files.

docker/docker-compose.indexer-test.yml
eldios: I tested everything... but I'll test again and double-check THIS section again.

Thanks for catching this.

docker/docker-compose.yml
eldios: no those values are the same as the default ones, it was mostly about deciding the values. If we're cool with the default values I'll just remove this for cleanliness.

docker/docker-compose.yml
eldios: using a proper k8s approach (like configmap or secrets) is a good idea but we're not doing that now, though. It's a good improvement for the future 👍🏻 

docker-compose mostly uses ENV usually OR can use a mapped file OR a volume .. so we can figure out something similar.

PR #4983, opened 2025-11-18
Return contexts with correct base keys from views

bart-linera:
## Motivation

The derived implementation of `View::context` currently returns just the context of the first sub-view of the view. This means that if a view A has a sub-view B, and view B has a sub-view C, when A calls `context()` on B and assumes that its own base key is eg. a byte shorter than what it gets as a result, it will in fact get a context with C's base key, which may lead to wrong results if B also appends some bytes to the base key before passing the context on to C (which happens in hashable wrappers, for example!).

## Proposal

Make the implementation of `context()` in all views correctly trim the base key contained in the context before returning the value.

This unfortunately introduces the necessity of cloning the context and returning it by value instead of by reference.

An alternative (but somewhat more complicated) solution would be to require _all_ views to store their inner contexts, and make the `View` derive macro require a `context` field in all structs for which `View` is derived.

## Test Plan

CI will catch regressions.

This was also manually confirmed to fix a bug that surfaced in #4977.

## Release Plan

- This bug may be causing some issues on the testnet, but the fix cannot be safely backported without a complex migration.

## Links

### DIFF
--- a/linera-views-derive/src/snapshots/linera_views_derive__tests__test_generate_view_code_C.snap
+++ b/linera-views-derive/src/snapshots/linera_views_derive__tests__test_generate_view_code_C.snap
@@ -22,8 +22,11 @@ where
             RegisterView<C, usize>,
         > as linera_views::views::View>::NUM_INIT_KEYS;
     type Context = C;
-    fn context(&self) -> &C {
-        self.register.context()
+    fn context(&self) -> C {
+        use linera_views::context::Context as _;
+        let __bytes_to_trim = 2;
+        let context = self.register.context();
+        context.clone_with_trimmed_key(__bytes_to_trim)
     }
     fn pre_load(context: &C) -> Result<Vec<Vec<u8>>, linera_views::ViewError> {
         use linera_views::context::Context as _;
@@ -106,7 +109,7 @@ where
             Self::post_load(context, &[])
         } else {
             let keys = Self::pre_load(&context)?;
-            let values = context.store().read_multi_values_bytes(keys).await?;
+            let values = context.store().read_multi_values_bytes(&keys).await?;
             Self::post_load(context, &values)
         }
     }
--- a/linera-views-derive/src/snapshots/linera_views_derive__tests__test_generate_view_code_C_with_where.snap
+++ b/linera-views-derive/src/snapshots/linera_views_derive__tests__test_generate_view_code_C_with_where.snap
@@ -23,8 +23,11 @@ where
             RegisterView<C, usize>,
         > as linera_views::views::View>::NUM_INIT_KEYS;
     type Context = C;
-    fn context(&self) -> &C {
-        self.register.context()
+    fn context(&self) -> C {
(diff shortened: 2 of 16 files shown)


linera-views/src/context.rs
```diff
@@ -65,6 +65,18 @@ impl BaseKey {
         Ok(key)
     }
 
+    /// Returns this key with a number of final bytes trimmed.
+    pub fn trimmed_key(&self, n: usize) -> Result<Vec<u8>, bcs::Error> {
+        if self.bytes.len() < n {
+            return Err(bcs::Error::Custom(format!(
+                "attempted to trim {} bytes from key of lenght {}",
```
afck:
```suggestion
                "attempted to trim {} bytes from key of length {}",
```

linera-views/src/context.rs
```diff
@@ -65,6 +65,18 @@ impl BaseKey {
         Ok(key)
     }
 
+    /// Returns this key with a number of final bytes trimmed.
+    pub fn trimmed_key(&self, n: usize) -> Result<Vec<u8>, bcs::Error> {
```
deuszx: This doesn't have to be `pub`, right?

linera-views-derive/src/lib.rs
```diff
@@ -133,6 +133,16 @@ fn generate_view_code(input: ItemStruct, root: bool) -> Result<TokenStream2, Err
         });
     }
 
+    let trim_key_logic = if num_fields < 256 {
```
deuszx, line 140: Can you leave a comment in the code explaining where these 3 magic numbers come from?

linera-execution/tests/revm.rs
```diff
@@ -277,13 +281,15 @@ async fn test_terminate_query_by_lack_of_fuel() -> anyhow::Result<()> {
         module: module.clone(),
     };
     {
-        let pinned = view.context().extra().user_contracts().pin();
+        let context = view.context();
+        let pinned = context.extra().user_contracts().pin();
```
deuszx, line 285: Why is it not correct to inline this anymore?

deuszx:
Ideally we'd also have a test proving that this was necessary but maybe that's too complicated to automate?

linera-execution/tests/revm.rs
bart-linera: Because `context()` now creates a temporary value, which then gets used by reference by `extra()` - but if it's not bound anywhere, the compiler complains that it's instantly dropped and the reference outlives the value it is referring to.

bart-linera:
> Ideally we'd also have a test proving that this was necessary but maybe that's too complicated to automate?

I can try. It should be relatively easy to create a dummy view type in a test module that would exhibit this issue.

linera-views-derive/src/lib.rs
bart-linera: Added in 6e14761

bart-linera:
@deuszx I added a test and verified that it was failing before the changes from this PR.

PR #4986, opened 2025-11-19
[testnet] Fix missing committee blobs handling.

afck:
## Motivation

https://github.com/linera-io/linera-protocol/pull/4978 revealed that the client fails to send the admin chain to validators that are missing a committee in some cases.

## Proposal

Remove `committees_for` from storage. Return `BlobsNotFound` and `EventsNotFound` instead of `ViewError`s.

Also remove some redundant code, and check height before epoch when handling a validated block.

I got stack overflows in some cases, so I also doubled the client's Tokio thread stack size.

## Test Plan

CI

https://github.com/linera-io/linera-protocol/pull/4978 failed without this.

## Release Plan

- These changes should
    - be released in a new SDK,
## Links

- PR to main that revealed this issue and also includes a quorum calculation change: https://github.com/linera-io/linera-protocol/pull/4978

### DIFF
--- a/linera-core/src/updater.rs
+++ b/linera-core/src/updater.rs
@@ -629,23 +629,23 @@ where
         let info = if let Ok(height) = target_block_height.try_sub_one() {
             // Figure out which certificates this validator is missing. In many cases, it's just the
             // last one, so we optimistically send that one right away.
-            let hash = self
-                .client
-                .local_node
-                .chain_state_view(chain_id)
-                .await?
-                .block_hashes([height])
-                .await?
-                .into_iter()
-                .next()
-                .ok_or_else(|| {
-                    ChainClientError::InternalError(
-                        "send_chain_information called with invalid target_block_height",
-                    )
-                })?;
-            let certificate = if let Some(certificate) = latest_certificate {
-                certificate
+            let certificate = if let Some(cert) = latest_certificate {
+                cert
             } else {
+                let hash = self
+                    .client
+                    .local_node
+                    .chain_state_view(chain_id)
+                    .await?
+                    .block_hashes([height])
+                    .await?
+                    .into_iter()
+                    .next()
+                    .ok_or_else(|| {
+                        ChainClientError::InternalError(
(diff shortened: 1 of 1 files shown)


linera-core/src/updater.rs
```diff
@@ -633,23 +629,23 @@ where
...
+                .await?
+                .block_hashes([height])
+                .await?
+                .into_iter()
+                .next()
+                .ok_or_else(|| {
+                    ChainClientError::InternalError(
+                        "send_chain_information called with invalid target_block_height",
+                    )
+                })?;
```
deuszx: No no, we DON'T want to do this if `latest_certificate` exists. Otherwise we're wasting time again - that was the whole point of the change. This adds 100ms+ to every call.

linera-service/src/cli/main.rs
```diff
@@ -2234,6 +2234,7 @@ fn main() -> anyhow::Result<process::ExitCode> {
         builder
     };
 
+    runtime.thread_stack_size(4 << 20);
```
deuszx, line 2237: Maybe a comment?

linera-execution/src/system.rs
```diff
@@ -778,20 +778,61 @@ where
...
+                .ok_or(ExecutionError::EventsNotFound(vec![event_id]))?;
+
+            let committee_hash: CryptoHash = bcs::from_bytes(&event)?;
+            let blob_id = BlobId::new(committee_hash, BlobType::Committee);
+            let committee_blob = self
+                .context()
+                .extra()
+                .get_blob(blob_id)
+                .await?
+                .ok_or(ExecutionError::BlobsNotFound(vec![blob_id]))?;
```
deuszx, line 829: Probably not introduced in this PR but maybe we can return all missing committee blobs at once? Then the client can choose whether to udpate one-by-one or in a single message.

afck:
Closing this for now. I will apply your comments on https://github.com/linera-io/linera-protocol/pull/4978 and backport once it's merged.

PR #4988, opened 2025-11-19
Indexer stores blobs published in blobs table.

deuszx:
## Motivation

In the original code I missed blobs that were part of the block – only blobs sent outside of the block were stored.

## Proposal

Extract blobs from the block and store.

Small refactoring of the API was done for better testability.

## Test Plan

CI.

## Release Plan

## Links

### DIFF
--- a/.github/workflows/indexer.yml
+++ b/.github/workflows/indexer.yml
@@ -44,7 +44,7 @@ jobs:
         cargo check --locked -p linera-indexer-example
     - name: Run indexer unit tests
       run: |
-        cargo test --locked -p linera-indexer --lib -- --skip db::postgres
+        cargo test --locked -p linera-indexer --lib
 
   indexer-integration-tests:
     if: github.event_name == 'merge_group' || github.event_name == 'workflow_dispatch'


linera-indexer/lib/src/db/mod.rs
```diff
@@ -26,27 +29,75 @@ pub trait IndexerDatabase: Send + Sync {
...
-        for (blob_id, blob_data) in blobs {
-            self.insert_blob_tx(&mut tx, blob_id, blob_data).await?;
+        // Add blobs from the block body (with transaction index)
+        for (txn_index, transaction_blobs) in block.body.blobs.iter().enumerate() {
+            for blob in transaction_blobs {
+                let blob_id = blob.id();
+                let blob_data = bincode::serialize(blob)?;
+                all_blobs.push((blob_id, blob_data, Some(txn_index as u32)));
+            }
         }
```
deuszx: Here's the first part of the change: store block's blobs.

linera-indexer/lib/src/db/mod.rs
```diff
@@ -26,27 +29,75 @@ pub trait IndexerDatabase: Send + Sync {
...
+            } else {
+                None
+            };
+            self.insert_blob_tx(
+                &mut tx,
+                blob_id,
+                blob_data,
+                block_hash_for_blob,
+                *transaction_index,
+            )
```
deuszx: And here's the second one: 

We now also store blob's txn index and block hash (if it has any).

linera-indexer/lib/src/db/postgres/mod.rs
```diff
@@ -114,16 +120,22 @@ impl PostgresDatabase {
...
         let blob_id_str = blob_id.hash.to_string();
         let blob_type = format!("{:?}", blob_id.blob_type);
+        let block_hash_str = block_hash.as_ref().map(|h| h.to_string());
+        let transaction_index_i64 = transaction_index.map(|i| i as i64);
 
-        // For now, we don't have block_hash and application_id context here
-        // These could be passed as optional parameters in the future
-        sqlx::query("INSERT INTO blobs (hash, blob_type, data) VALUES ($1, $2, $3) ON CONFLICT (hash) DO NOTHING")
+        sqlx::query(
+            "INSERT INTO blobs (hash, blob_type, data, block_hash, transaction_index) VALUES ($1, $2, $3, $4, $5) ON CONFLICT (hash) DO NOTHING"
```
afck: (long line)

linera-indexer/lib/src/db/postgres/tests.rs
```diff
@@ -250,6 +247,134 @@ async fn test_incoming_bundles_storage_and_query() {
...
+
+        // Verify block was stored
+        let retrieved_block = db.get_block(&block_hash).await.unwrap();
+        assert_eq!(block_data, retrieved_block);
+
+        // Verify all blobs were stored
+        assert!(db.get_blob(&standalone_blob.id()).await.is_ok());
+        assert!(db.get_blob(&blob1.id()).await.is_ok());
+        assert!(db.get_blob(&blob2.id()).await.is_ok());
+        assert!(db.get_blob(&blob3.id()).await.is_ok());
```
afck, line 301: `unwrap()` would print the error, `assert!(…is_ok())` would only see the value `false`.

linera-indexer/lib/src/db/postgres/tests.rs
```diff
@@ -250,6 +247,134 @@ async fn test_incoming_bundles_storage_and_query() {
...
+
+        // Query the database directly to verify metadata
+        use sqlx::Row;
```
afck: I'd move that to the top.

linera-indexer/lib/src/db/postgres/tests.rs
```diff
@@ -250,6 +247,134 @@ async fn test_incoming_bundles_storage_and_query() {
...
+        let pool = &db.pool;
+
+        // Check standalone blob has no block_hash or transaction_index
```
afck: (I see, so the block hash in the blobs table is only for blobs that are included in a block's _body_. If a block publishes a data blob that is separate, there's no entry for the publishing block hash.)

linera-indexer/lib/src/grpc/tests.rs
```diff
@@ -288,22 +287,19 @@ async fn test_process_stream_end_to_end_mixed_elements() {
...
-    // - 1 blob stored (all blobs have same content/BlobId, so they overwrite each other)
-    // - 2 blocks stored (valid_block1 and valid_block2 with different chain IDs)
-    // - invalid_block should NOT be stored (deserialization failed)
-
-    assert_eq!(
-        database.blob_count(),
-        1,
-        "Should have 1 unique blob stored (all blobs have same content/BlobId)"
+    // Verify data was stored successfully by querying the database
+    // All 3 blobs have the same content, so they should have the same BlobId
```
afck:
```suggestion
    // All 3 blobs have the same content, so they should have the same blob ID
```

linera-indexer/lib/src/grpc/tests.rs
```diff
@@ -288,22 +287,19 @@ async fn test_process_stream_end_to_end_mixed_elements() {
...
 
-    assert_eq!(
-        database.block_count(),
-        2,
-        "Should have 2 valid blocks stored (invalid block should not be stored)"
-    );
+    // Verify both valid blocks were stored
+    // We can't easily check block count without adding a helper method,
+    // so we verify the blocks exist by attempting to query them
+    // (The test passes if no errors occur during storage)
```
afck: Why can't we try to read the two blocks?

deuszx: Hm, I thought I pushed a change that removed this comment.

linera-indexer/lib/src/db/postgres/tests.rs
deuszx: Yes. Does that make sense?

afck: Sure.
I mean, every blob was published by one or more blocks, and some of the blobs were included in one or more blocks' bodies. Not sure which hash is most useful to put in the table. :shrug:

deuszx: Can a blob be published in more than one block?

afck: Of course! They are content-addressed, and which publishing block was "first" is not objective, so it's valid for both to publish the same blob.

deuszx: Right, the subsequent inserts will be no-ops but each block can be "equally" first trying to publish them.

Maybe it makes more sene to _not_ store it in the DB after all. It creates an illusion that those blocks were first.

linera-indexer/lib/src/grpc/tests.rs
deuszx: Done.

linera-indexer/lib/src/db/postgres/tests.rs
afck: Maybe, yes. Or, _if_ we store a block hash for each blob, I'd always make it the latest block _using_ that blob: that's the most useful, because it can be used for the longest time to prove that blob's existence.

deuszx: So you're saying I should overwrite rather than ignore (on duplicates)?

Also, _latest_ might be hard to define precisely as independent chains may have no order between them.

afck: It's just about the epoch. In the worker code itself, I think, we overwrite if and only if the epoch is strictly greater.

deuszx: but here the data comes from the `Block` itself. Are you saying I should be comparing epochs?

linera-indexer/lib/src/grpc/tests.rs
deuszx: Fixed in [Various small changes from the review](https://github.com/linera-io/linera-protocol/pull/4988/commits/a75c4c2e0242c276a99b893f1724bde1ade06a70)

linera-indexer/lib/src/db/postgres/tests.rs
deuszx: Fixed in [Various small changes from the review](https://github.com/linera-io/linera-protocol/pull/4988/commits/a75c4c2e0242c276a99b893f1724bde1ade06a70)

linera-indexer/lib/src/db/postgres/mod.rs
deuszx: Fixed in [Various small changes from the review](https://github.com/linera-io/linera-protocol/pull/4988/commits/a75c4c2e0242c276a99b893f1724bde1ade06a70)

linera-indexer/lib/src/db/postgres/tests.rs
afck: That's what I'd do, yes: With each blob, always associate the epoch and hash of the latest block we've seen use it.

afck (approved):
Looks good to me, but the PR title reads like a crash blossom.

deuszx:
> Looks good to me, but the PR title reads like a crash blossom.

Updated :)

PR #4989, opened 2025-11-19
Allow filtering incoming messages by application.

afck:
## Motivation

It's useful to filter out spam messages on a node service that e.g. is running a chain specifically for one application.

## Proposal

Allow filtering in two ways:
* All bundles that contain _at least one_ message for a whitelisted app.
* All bundles that contain _only_ messages for whitelisted apps.

The latter is stricter, but the former can be useful e.g. for apps that create other apps and then call them.

## Test Plan

A test for the new filtering options was added. (Claude)

## Release Plan

    - be released in a new SDK.

## Links

- Closes https://github.com/linera-io/linera-protocol/issues/4987.

### DIFF
--- a/CLI.md
+++ b/CLI.md
@@ -170,7 +170,7 @@ Client implementation and command-line tool for the Linera blockchain
 
 * `--restrict-chain-ids-to <RESTRICT_CHAIN_IDS_TO>` — A set of chains to restrict incoming messages from. By default, messages from all chains are accepted. To reject messages from all chains, specify an empty string
 * `--reject-message-bundles-without-application-ids <REJECT_MESSAGE_BUNDLES_WITHOUT_APPLICATION_IDS>` — A set of application IDs. If specified, only bundles with at least one message from one of these applications will be accepted
-* `--reject-messages-with-other-application-ids <REJECT_MESSAGES_WITH_OTHER_APPLICATION_IDS>` — A set of application IDs. If specified, only bundles where all messages are from one of these applications will be accepted
+* `--reject-message-bundles-with-other-application-ids <REJECT_MESSAGE_BUNDLES_WITH_OTHER_APPLICATION_IDS>` — A set of application IDs. If specified, only bundles where all messages are from one of these applications will be accepted
 * `--timings` — Enable timing reports during operations
 * `--timing-interval <TIMING_INTERVAL>` — Interval in seconds between timing reports (defaults to 5)
 
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -144,7 +144,7 @@ pub struct MessagePolicy {
     reject_message_bundles_without_application_ids: Option<HashSet<GenericApplicationId>>,
     /// A collection of applications: If `Some`, only bundles all of whose messages are by these
     /// applications will be accepted.
-    reject_messages_with_other_application_ids: Option<HashSet<GenericApplicationId>>,
+    reject_message_bundles_with_other_application_ids: Option<HashSet<GenericApplicationId>>,
 }
 
 #[derive(Copy, Clone, Debug, clap::ValueEnum)]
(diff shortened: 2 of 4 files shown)


CLI.md
```diff
@@ -169,6 +169,8 @@ Client implementation and command-line tool for the Linera blockchain
     Don't include any messages in blocks, and don't make any decision whether to accept or reject
 
 * `--restrict-chain-ids-to <RESTRICT_CHAIN_IDS_TO>` — A set of chains to restrict incoming messages from. By default, messages from all chains are accepted. To reject messages from all chains, specify an empty string
+* `--accept-messages-with-application-ids <ACCEPT_MESSAGES_WITH_APPLICATION_IDS>` — A set of application IDs. If specified, only bundles with at least one message from one of these applications will be accepted
+* `--reject-messages-with-other-application-ids <REJECT_MESSAGES_WITH_OTHER_APPLICATION_IDS>` — A set of application IDs. If specified, only bundles where all messages are from one of these applications will be accepted
```
deuszx: and if the same application ID is in both sets? Then the `REJECT_MESSAGES_WITH_OTHER_APPLICATION_IDS` takes precedence?

afck: Yes, exactly: both filters are applied, and the stricter one wins.

linera-core/src/client/mod.rs
```diff
@@ -139,6 +139,10 @@ pub struct MessagePolicy {
     /// accepted. `Option::None` means that messages from all chains are accepted. An empty
     /// `HashSet` denotes that messages from no chains are accepted.
     restrict_chain_ids_to: Option<HashSet<ChainId>>,
+    /// A collection of applications: If `Some`, only bundles with at least one message by any
+    /// of these applications will be accepted.
+    accept_messages_with_application_ids: Option<HashSet<GenericApplicationId>>,
+    reject_messages_with_other_application_ids: Option<HashSet<GenericApplicationId>>,
```
bart-linera: This should probably be documented as well, even if just for completeness.

afck: Done in https://github.com/linera-io/linera-protocol/pull/4989/commits/4e5fc485860385b00b149aacef6518dfb30fab0a.

ma2bd: How about renaming `accept_messages_with_application_ids` into `reject_message_bundles_without_application_ids` ?

afck: Done in https://github.com/linera-io/linera-protocol/pull/4989/commits/9d53a2d717544dd444edb0bb15fe28fe2a7a6313.

deuszx: in other places we used _required_ application IDs.

afck: What names would you suggest for the two fields then?

afck: I guess we could use `restrict_to_application_ids` and `required_application_ids`?

CLI.md
```diff
@@ -169,6 +169,8 @@ Client implementation and command-line tool for the Linera blockchain
     Don't include any messages in blocks, and don't make any decision whether to accept or reject
 
 * `--restrict-chain-ids-to <RESTRICT_CHAIN_IDS_TO>` — A set of chains to restrict incoming messages from. By default, messages from all chains are accepted. To reject messages from all chains, specify an empty string
+* `--reject-message-bundles-without-application-ids <REJECT_MESSAGE_BUNDLES_WITHOUT_APPLICATION_IDS>` — A set of application IDs. If specified, only bundles with at least one message from one of these applications will be accepted
+* `--reject-messages-with-other-application-ids <REJECT_MESSAGES_WITH_OTHER_APPLICATION_IDS>` — A set of application IDs. If specified, only bundles where all messages are from one of these applications will be accepted
```
ma2bd: sorry but should it be also "message-bundles" here then?

linera-core/src/client/mod.rs
deuszx: Funny thing is we already use _exactly_ `required_application_ids` in `ApplicationDescription`:
```rust
    /// Required dependencies.
    pub required_application_ids: Vec<ApplicationId>,
```

afck: And we already use `restrict_chain_ids_to`.

afck: @ma2bd: How do you feel about `restrict_to_application_ids` and `required_application_ids`?

(These are client options, so it's only clear from the help text, not from the name, that these refer to incoming message bundles. Alternatively, we could also rename `restrict_chain_ids_to`, to include "messages" or "message bundles".)

PR #4993, opened 2025-11-19
Fix missing committee blob handling. (#4978)

afck:
Partial backport of #4978.

## Motivation

The current `Committee` implementation is a bit complicated, and sometimes sets the quorum threshold higher than necessary.

## Proposal

#4978 revealed an issue with the certificate handling logic in the worker: `committees_for` returns `ViewErrors` instead of `BlobsNotFound`/`EventsNotFound`, so the client fails to send the validators the admin chain if needed.

This was fixed; I also kept minor cleanups I made while debugging this. I got a few stack overflows, so I doubled the client's Tokio thread stack size.

## Test Plan

CI

## Release Plan

- The fixes (but maybe not the quorum change) should be
    - released in a new SDK and
    - released in a validator hotfix.

## Links

- PR to main: #4978

### DIFF
--- a/linera-service/src/cli/main.rs
+++ b/linera-service/src/cli/main.rs
@@ -2234,6 +2234,8 @@ fn main() -> anyhow::Result<process::ExitCode> {
         builder
     };
 
+    // The default stack size 2 MiB causes some stack overflows in ValidatorUpdater methods.
+    runtime.thread_stack_size(4 << 20);
     if let Some(blocking_threads) = options.tokio_blocking_threads {
         runtime.max_blocking_threads(blocking_threads);
     }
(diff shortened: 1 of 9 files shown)


linera-service/src/cli/main.rs
```diff
@@ -2234,6 +2234,8 @@ fn main() -> anyhow::Result<process::ExitCode> {
         builder
     };
 
+    // The default stack size 2 MiB causes some stack overflows in ValidatorUpdater methods.
+    runtime.thread_stack_size(4 << 20);
```
ma2bd, line 2238: does this belong to this PR?

ma2bd: Mm I guess it does

ma2bd: But did you mean 2 or 4 MiB?

afck: Yes, something in this PR made it necessary, but it looks like we were already very close to the stack size.

PR #5001, opened 2025-11-20
feat: Add validator command group with batch operations

eldios:
## Summary

Refactors validator management commands from a flat structure into a hierarchical `validator` command group with 7 subcommands. Adds batch operations for managing validators from JSON files, replacing the need for external scripts.

## Motivation

- **Pain Point 1**: 50+ flat commands made CLI discovery difficult
- **Pain Point 2**: No batch operations - required external ZSH scripts for managing multiple validators
- **Pain Point 3**: Inconsistent naming patterns across validator commands
- **Solution**: Domain-based command grouping with JSON batch support

### New `linera validator` Command Group

**Query Operations:**
- `query` - Query single validator status
- `query-batch` - Query multiple validators from JSON file
- `list` - List all validators with optional filtering

**Management Operations (Admin):**
- `add` - Add single validator with online checks
- `remove` - Remove validator by public key
- `batch-update` - Add/modify/remove validators from JSON file with dry-run support
- `sync` - Synchronize validator with local chain state

### JSON Format Support

Two flexible formats for batch operations:

**Simple List** (adds all):
```json
{
  "validators": [
    {
      "publicKey": "ed25519:...",
      "accountKey": "ed25519:...",
      "networkAddress": "192.168.1.100:8080",
      "votes": 100
    }
  ]
}
```

**Operations** (granular control):
```json
{
  "add": [{...}],
  "modify": [{...}],
  "remove": ["pubkey1", "pubkey2"]
}
```

### Backward Compatibility

All old commands remain functional with deprecation messages in help text:
- `query-validator` → `validator query`
- `query-validators` → `validator list`
- `set-validator` → `validator add`
- `remove-validator` → `validator remove`
- `change-validators` → `validator batch-update`
- `sync-validator` → `validator sync`

**No compile-time warnings** - deprecation guidance appears only in `--help` output.

## Testing

✅ Build completes with 0 deprecation warnings
✅ All 7 new subcommands working correctly
✅ JSON parsing validated for both formats
✅ Backward compatibility confirmed - old commands still accessible
✅ CLI documentation regenerated with `linera help-markdown`

## Files Changed

- `linera-service/src/cli/command.rs` (+163 lines): Command definitions and JSON structures
- `linera-service/src/cli/main.rs` (+518 lines): Command handlers for all 7 subcommands
- `CLI.md`: Auto-regenerated documentation

## Example Usage

```bash
# Query single validator
linera validator query 192.168.1.100:8080

# Batch query from file
linera validator query-batch validators.json

# List validators with minimum votes
linera validator list --min-votes 100

# Add validator with online check
linera validator add --public-key ed25519:... --account-key ed25519:... --address 192.168.1.100:8080 --votes 100

# Batch update with dry-run preview
linera validator batch-update operations.json --dry-run

# Apply batch changes
linera validator batch-update operations.json
```

## Impact

- **Developer Experience**: Clearer command organization, better discoverability
- **Operations**: Batch operations eliminate need for external scripts
- **Migration**: Zero breaking changes - smooth transition path via deprecation messages

### DIFF
--- a/CLI.md
+++ b/CLI.md
@@ -54,6 +54,14 @@ This document contains the help content for the `linera` command-line program.
 * [`linera chain`↴](#linera-chain)
 * [`linera chain show-block`↴](#linera-chain-show-block)
 * [`linera chain show-chain-description`↴](#linera-chain-show-chain-description)
+* [`linera validator`↴](#linera-validator)
+* [`linera validator query`↴](#linera-validator-query)
+* [`linera validator query-batch`↴](#linera-validator-query-batch)
+* [`linera validator list`↴](#linera-validator-list)
+* [`linera validator add`↴](#linera-validator-add)
+* [`linera validator remove`↴](#linera-validator-remove)
+* [`linera validator batch-update`↴](#linera-validator-batch-update)
+* [`linera validator sync`↴](#linera-validator-sync)
 * [`linera project`↴](#linera-project)
 * [`linera project new`↴](#linera-project-new)
 * [`linera project test`↴](#linera-project-test)
@@ -92,12 +100,12 @@ Client implementation and command-line tool for the Linera blockchain
 * `sync-balance` — (DEPRECATED) Synchronize the local state of the chain with a quorum validators, then query the local balance
 * `sync` — Synchronize the local state of the chain with a quorum validators
 * `process-inbox` — Process all pending incoming messages from the inbox of the given chain by creating as many blocks as needed to execute all (non-failing) messages. Failing messages will be marked as rejected and may bounce to their sender depending on their configuration
-* `query-validator` — Show the version and genesis config hash of a new validator, and print a warning if it is incompatible. Also print some information about the given chain while we are at it
-* `query-validators` — Show the current set of validators for a chain. Also print some information about the given chain while we are at it
-* `sync-validator` — Synchronizes a validator with the local state of chains
-* `set-validator` — Add or modify a validator (admin only)
-* `remove-validator` — Remove a validator (admin only)
-* `change-validators` — Add, modify, and/or remove multiple validators in a single epoch (admin only)
+* `query-validator` — (DEPRECATED) Show the version and genesis config hash of a new validator
+* `query-validators` — (DEPRECATED) Show the current set of validators for a chain
+* `sync-validator` — (DEPRECATED) Synchronizes a validator with the local state of chains
+* `set-validator` — (DEPRECATED) Add or modify a validator (admin only)
+* `remove-validator` — (DEPRECATED) Remove a validator (admin only)
+* `change-validators` — (DEPRECATED) Add, modify, and/or remove multiple validators in a single epoch (admin only)
 * `revoke-epochs` — Deprecates all committees up to and including the specified one
 * `resource-control-policy` — View or update the resource control policy
 * `benchmark` — Run benchmarks to test network performance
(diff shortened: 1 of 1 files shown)


.gcloudignore
```diff
@@ -4,6 +4,9 @@
 # ---------------------------------------------------------------------------------*
 node_modules
 npm-debug.log
+# Ignore dangling symlinks to node_modules in web examples
```
ma2bd: @Twey ?

CLI.md
```diff
@@ -547,9 +562,9 @@ Deprecated: Use change-validators instead, which allows adding, changing and rem
 
 ## `linera remove-validator`
 
-Remove a validator (admin only)
+(DEPRECATED) Remove a validator (admin only)
 
-Deprecated: Use change-validators instead, which allows adding, changing and removing any number of validators in a single operation.
```
Twey, line 552: It's probably time to remove this entirely — it was already deprecated and now it's double-deprecated :sweat_smile:  Do we have anything relying on it?

linera-service/src/cli/main.rs
```diff
@@ -606,6 +606,524 @@ impl Runnable for Job {
...
+
+                        for chain_id in chains {
+                            let chain = context.make_chain_client(chain_id);
+
+                            Box::pin(chain.sync_validator(validator.clone())).await?;
+                        }
+                    }
+                }
+            }
+
```
deuszx, line 1126: if we start the work on splitting – can you move it to a dedicated file as well?

linera-service/src/cli/command.rs
```diff
@@ -194,6 +194,135 @@ use crate::util::{
...
+        /// The public key of the validator.
+        #[arg(long)]
+        public_key: ValidatorPublicKey,
+    },
+
+    /// Add, modify, and/or remove validators from a JSON file (admin only)
+    ///
+    /// JSON format supports two styles:
+    /// 1. Simple list: {"validators": [{...}]} - adds all validators
+    /// 2. Operations: {"add": [{...}], "modify": [{...}], "remove": ["pubkey"]}
```
Twey, line 301: This is dangerous because the keys of the object are unordered so the commands may be executed in an unexpected order. I'd prefer JSON-lines:
```json
{ "add": … }
{ "modify": … }
{ "add": … }
…

Comment (Twey) [reply linera-service/src/cli/command.rs:301]: I'd like the elided parts to be documented here, as well (it's not obvious, at least to me, what should go there).

Comment (Twey) [reply linera-service/src/cli/command.rs:301]: Rather than baking in file-reading, I think it would be simpler and more flexible to read the operations from stdin. File functionality can still be provided by the shell (`linera validator batch-update <changes.json`).

Comment (Twey) [reply linera-service/src/cli/command.rs:301]: Alternatively, rather than introducing a new format here we could take a stream of lines of `update` commands in the same format as the individual commands. `--batch` can then be a flag to the normal update command to tell it to read a stream of argument lines.

File: linera-service/src/cli/command.rs [hunk 5]
```diff
@@ -194,6 +194,135 @@ use crate::util::{
...
+            public_key: spec.public_key,
+            account_key: spec.account_key,
+            address: spec.network_address,
+            votes: spec.votes,
+        }
+    }
+}
+
+#[derive(Clone, clap::Subcommand)]
+pub enum ValidatorCommand {
```
Twey, line 233: Nit: I'd prefer this in a submodule rather than growing this already quite large file.

Twey:
Typically we merge (and squash) to `main` first, then backport. I'm not the biggest fan of that approach, but it's probably worth keeping consistent for now.

linera-service/src/cli/command.rs
deuszx: Why not a shell script? I haven't seen this approach anywhere before.

Twey: I think my preferred approach would indeed just be a shell script since we're not really doing anything more clever than executing the commands one by one.  But if we do want a batch command then the above applies.  There are some examples of this sort of thing, e.g. gpg2, but indeed it's not as common as just writing a shell script.

eldios:
Close in favor of https://github.com/linera-io/linera-protocol/pull/5009

PR #5005, opened 2025-11-21
Forward cross chain messages for different recipients actually in parallel

ndr-ds:
## Motivation

Right now we're using `FuturesUnordered`, which will poll the futures concurrently in the same task.

## Proposal

Use `JoinSet` so we actually can poll these in parallel, using different cores

## Test Plan

Benchmarked a network without this, and with this, and with this our cross chain message queue wait time improved by about 10%. This was on a network with 4 worker threads per shard.

## Release Plan

### DIFF
--- a/linera-rpc/src/cross_chain_message_queue.rs
+++ b/linera-rpc/src/cross_chain_message_queue.rs
@@ -58,9 +58,9 @@ pub(crate) async fn forward_cross_chain_queries<F, G>(
     handle_request: F,
 ) where
     F: Fn(ShardId, CrossChainRequest) -> G + Send + Clone + 'static,
-    G: Future<Output = anyhow::Result<()>>,
+    G: Future<Output = anyhow::Result<()>> + Send + 'static,
 {
-    let mut steps = futures::stream::FuturesUnordered::new();
+    let mut steps = tokio::task::JoinSet::new();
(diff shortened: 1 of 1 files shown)


ndr-ds:
* **#5005** <a href="https://app.graphite.com/github/pr/linera-io/linera-protocol/5005?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/linera-io/linera-protocol/5005?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ndr-ds:
Wether we initially spawn a task or not is controlled by `job_states`, which is a `HashMap` from `QueueId` to `JobState`. So stuff that has the same `sender` and `recipient` will use the same task.

ma2bd:
Interesting. What is the CPU-heavy operation taking advantage of multiple cores?

ndr-ds:
Like I mentioned in the description, this wasn't really a bottleneck. Was just a place that I thought could be one. But the improvements were small, so we're probably not even exhausting the 1 core with this, as it just gets slightly faster in parallel.
I'm also open to closing this, if people think it's not worth it :)
I personally think it doesn't hurt 🤷🏻‍♂️ but up to you guys

PR #5009, opened 2025-11-21
feat: refactor validator commands to hierarchical structure

eldios:
## Motivation

**Before:** Validator commands scattered across 7 top-level commands with inconsistent naming and no grouping:
```bash
linera --help | grep -i '\-validator'
  query-validator                 Show the version and genesis config...
  query-validators                Show the current set of validators...
  sync-validator                  Synchronizes a validator...
  sync-all-validators             Synchronizes all validators...
  set-validator                   Add or modify a validator (admin only)
  remove-validator                Remove a validator (admin only)
  change-validators               Add, modify, and/or remove multiple validators...
```

**After:** Clean, hierarchical structure under `linera validator` with alphabetical ordering:
```bash
linera validator --help
Manage validators in the committee

Usage: linera validator [OPTIONS] <COMMAND>

Commands:
  add           Add a validator to the committee
  batch-query   Query multiple validators using a JSON specification file
  batch-update  Apply multiple validator changes from JSON input
  list          List all validators in the committee
  query         Query a single validator's state and connectivity
  remove        Remove a validator from the committee
  sync          Synchronize chain state to a validator
  help          Print this message or the help of the given subcommand(s)
```

**QoL Impact:** Massive improvement for DevOps workflows - all validator operations now discoverable under one command, dramatically speeding up daily operations.

### Breaking Changes
- **Removed:** `query-validator`, `query-validators`, `sync-validator`, `sync-all-validators`, `set-validator`, `remove-validator`, `change-validators`
- **Added:** `linera validator {add, batch-query, batch-update, list, query, remove, sync}`

### Key Improvements

**1. Simplified JSON Format (No Verbs)**
```json
{
  "validator_pubkey_1": {
    "accountKey": "ed25519:...",
    "address": "grpcs://validator1.example.com:443",
    "votes": 100
  },
  "validator_pubkey_2": null,
  "validator_pubkey_3": {}
}
```
- Public key as map key = add or modify validator
- `null` or `{}` = remove validator
- Missing key = unchanged validator

**2. Flexible Input Methods**
```bash
# File
linera validator batch-update changes.json

# Stdin pipe
cat changes.json | linera validator batch-update

# Shell redirect
linera validator batch-update <changes.json
```

**3. Safety Features**
- Detailed recap before applying changes with formatted output (box drawing)
- Confirmation prompt requiring uppercase "YES"
- `--yes` / `-y` flag for automation/CI
- `--dry-run` for preview without execution

**4. CLI Best Practices**
- ✅ Alphabetical command ordering
- ✅ Consistent, action-oriented help messages
- ✅ Multi-line documentation with examples
- ✅ Short flags for common operations (`-y`)
- ✅ Clear parameter descriptions with format hints
- ✅ Visual hierarchy in output

### Example: batch-update Recap
```
╔══════════════════════════════════════════════════════════════════════════════╗
║                        VALIDATOR BATCH UPDATE RECAP                          ║
╚══════════════════════════════════════════════════════════════════════════════╝

Summary:
  • 2 validator(s) to add
  • 1 validator(s) to modify
  • 1 validator(s) to remove

Validators to ADD:
  + ed25519:abc123...
    Address:     grpcs://validator1.example.com:443
    Account Key: ed25519:def456...
    Votes:       100

═════════════════════════════════════════════════════════════════════════════
⚠️  WARNING: This operation will modify the validator committee.
             Changes are permanent and will be broadcast to the network.
═════════════════════════════════════════════════════════════════════════════

Do you want to proceed? Type 'YES' (uppercase) to confirm:
```

## Implementation

**Architecture:**
- All validator logic encapsulated in `validator.rs` (~1100 lines)
- Zero code duplication with main.rs
- Type aliases for complex ClientContext types
- Zero `unwrap()` calls - proper `Result<T, E>` error handling throughout

**Code Quality:**
- 12 comprehensive unit tests (100% passing)
- Zero clippy warnings: `cargo clippy -- -D warnings`
- Proper validation at all input boundaries

## Test Plan

**Unit Tests:**
```bash
cargo test --package linera-service --lib cli::validator::tests
running 12 tests
test result: ok. 12 passed; 0 failed
```

**Manual Verification:**
- ✅ All 7 subcommands display in `--help`
- ✅ Stdin input works (pipe, redirect, explicit `-`)
- ✅ Confirmation prompt validates "YES" correctly
- ✅ `--yes` flag bypasses confirmation
- ✅ `--dry-run` shows recap without applying
- ✅ JSON validation catches malformed input

**Files Modified:**
- `linera-service/src/cli/validator.rs` - NEW (~1100 lines)
- `linera-service/src/cli/command.rs` - Updated command enum
- `linera-service/src/cli/main.rs` - Minimal delegation
- `linera-service/src/cli/mod.rs` - Added validator module

## Release Plan

- These changes follow the usual release cycle.

**Migration Note:** Old commands map directly to new subcommands. Update scripts by replacing command names:
- `query-validator` → `validator query`
- `query-validators` → `validator list`
- `set-validator` → `validator add`
- `remove-validator` → `validator remove`
- `change-validators` → `validator batch-update`
- `sync-validator` / `sync-all-validators` → `validator sync`

## Links

- [Reviewer checklist](https://github.com/linera-io/linera-protocol/blob/main/CONTRIBUTING.md#reviewer-checklist)

### DIFF
--- a/linera-service/src/cli_wrappers/wallet.rs
+++ b/linera-service/src/cli_wrappers/wallet.rs
@@ -1058,33 +1058,41 @@ impl ClientWrapper {
         modify_validators: &[(String, String, usize, usize)], // (public_key, account_key, port, votes)
         remove_validators: &[String],
     ) -> Result<()> {
-        use std::collections::HashMap;
+        use std::str::FromStr;
 
-        use serde_json::Value;
+        use linera_base::crypto::{AccountPublicKey, ValidatorPublicKey};
 
-        // Build JSON object for validator changes
-        let mut changes = HashMap::new();
+        // Build a map that will be serialized to JSON
+        // Use the exact types that deserialization expects
+        let mut changes = std::collections::HashMap::new();
 
         // Add/modify validators
-        for (public_key, account_key, port, votes) in
+        for (public_key_str, account_key_str, port, votes) in
             add_validators.iter().chain(modify_validators.iter())
         {
+            let public_key = ValidatorPublicKey::from_str(public_key_str)
+                .with_context(|| format!("Invalid validator public key: {}", public_key_str))?;
+
+            let account_key = AccountPublicKey::from_str(account_key_str)
+                .with_context(|| format!("Invalid account public key: {}", account_key_str))?;
+
             let address = format!("{}:127.0.0.1:{}", self.network.short(), port);
-            changes.insert(
-                public_key.clone(),
-                serde_json::json!({
-                    "accountKey": {
-                        "Secp256k1": account_key
-                    },
-                    "address": address,
-                    "votes": votes,
-                }),
-            );
+
+            // Create ValidatorChange struct
+            let change = crate::cli::validator::ValidatorChange {
+                account_key,
+                network_address: address,
+                votes: std::num::NonZero::new(*votes as u64).context("Votes must be non-zero")?,
+            };
+
+            changes.insert(public_key, Some(change));
         }
 
-        // Remove validators (set to null)
-        for validator_key in remove_validators {
-            changes.insert(validator_key.clone(), Value::Null);
+        // Remove validators (set to None)
+        for validator_key_str in remove_validators {
+            let public_key = ValidatorPublicKey::from_str(validator_key_str)
+                .with_context(|| format!("Invalid validator public key: {}", validator_key_str))?;
+            changes.insert(public_key, None);
         }
 
         // Create temporary file with JSON


linera-service/src/cli/validator.rs
```diff
@@ -0,0 +1,1260 @@
...
+use linera_persistent::Persist;
+use linera_rpc::node_provider::NodeProvider;
+use linera_storage::Storage;
+use serde::{Deserialize, Serialize};
+use tokio::sync::Mutex;
+use tracing::{error, info, warn};
+
+/// Type alias for the complex ClientContext type used in validator handlers.
+type MutexedContext<S, W, Si> =
+    Arc<Mutex<ClientContext<linera_core::environment::Impl<S, NodeProvider, Si>, W>>>;
```
Twey: You can make it less complex by abstracting over `Environment` itself rather than all its parameters — and then you probably don't gain much from aliasing it.

You probably took this from `client_context.rs`, but it's used there in the concrete only because we need to construct one.  Here it's used only in a negative position, so there's no need to be so prescriptive.

linera-service/src/cli/validator.rs
```diff
@@ -0,0 +1,1260 @@
...
+    /// Check if this spec represents an empty/remove operation.
+    fn is_empty(&self) -> bool {
+        self.account_key.is_none() && self.network_address.is_none() && self.votes.is_none()
+    }
+
+    /// Validate that all required fields are present for add/modify operations.
+    fn validate(&self) -> Result<()> {
+        if self.is_empty() {
+            return Ok(()); // Empty is valid (represents removal)
+        }
```
Twey: Is there a reason we need two different ways to represent removal?

linera-service/src/cli/validator.rs
```diff
@@ -0,0 +1,1260 @@
...
+            bail!("address is required for add/modify operations");
+        }
+        let address = self.network_address.as_ref().unwrap();
+        if address.is_empty() {
+            bail!("Validator network address cannot be empty");
+        }
+        let votes = self.votes.unwrap_or(1);
+        if votes == 0 {
+            bail!("Validator votes must be greater than 0");
+        }
```
Twey: If all of these fields are required it would be nice to have them not be `Option`al in the first place.

linera-service/src/cli/validator.rs
```diff
@@ -0,0 +1,1260 @@
...
+        /// Preview changes without applying them
+        #[arg(long)]
+        dry_run: bool,
+        /// Skip confirmation prompt (use with caution)
+        #[arg(long, short = 'y')]
+        yes: bool,
+        /// Skip online connectivity checks for validators being added or modified
+        #[arg(long)]
+        skip_online_check: bool,
+    },
```
Twey: I think we can probably just call this `update`.  It subsumes both batch and individual operations.

linera-service/src/cli/validator.rs
```diff
@@ -0,0 +1,1260 @@
...
+                .read_to_string(&mut buffer)
+                .context("Failed to read from stdin")?;
+            buffer
+        }
+        Some(path_str) => {
+            // Read from file
+            std::fs::read_to_string(path_str)
+                .with_context(|| format!("Failed to read batch file: {}", path_str))?
+        }
+    };
```
Twey: [clio](https://crates.io/crates/clio) will do this in a generic way (failing serde parsing if the file/stream is not readable).

linera-service/src/cli/validator.rs
Twey: I guess this should be a `NonZero<u64>` then… though I'm not certain votes can't be zero?

linera-service/src/cli/validator.rs
eldios: okay, simplified/cleaned up a bit. Please check 🙏🏻

linera-service/src/cli/validator.rs
eldios: you mentioned setting to `null` in our conversation, I prefer `{}`.. I guess both are valid, I added both

linera-service/src/cli/validator.rs
eldios: added.

linera-service/src/cli/validator.rs
eldios: done

linera-service/src/cli/validator.rs
eldios: so I tried adding back without the alias but `clippy` got REALLY mad about `complex_type` issues... so I think having a bit of a proxy alias wouldn't hurt too much.

So unless you have big concerns about leaving it there, I'll just pass.

Twey: I don't mind the alias as much as I mind the concrete reference to `Impl`, which should really be an implementation detail. (I know Matt has strong feelings about type aliases, though.) The purpose of the `Environment` trait is to allow us to update the set of things that are considered dependencies for the client without changing every point along the codepath where the client flows, and every point where we have to mention `Impl` by name is a point that has to be updated every time we touch the `Environment` trait.

linera-service/src/cli/validator.rs
Twey: Right, but I chose `null` for a reason:

For a merge the patch `{ "foo": "bar" }` applied to `{ "foo": "baz", "quux": "arthur" }` results in `{ "foo": "baz", "quux": "arthur" }`.  So in general fields that are omitted from the merge mean to leave those fields alone.  By extension, an object with _all_ fields omitted means to leave _all_ fields on the object alone.  Making it instead suddenly mean ‘delete the object’ results in a weird discontinuity where an object with 0 fields behaves in a way that's incompatible with an object with 1 field, an object with 2 fields, et cetera.  Meanwhile `null` doesn't have an obvious merge semantics with objects, so repurposing it to mean ‘delete’ isn't too incongruous with the rest of the semantics.

This is also a pretty standard semantics for merges that need to support delete, e.g. [Redis](https://redis.io/docs/latest/commands/json.merge/) or the JSON Merge Patch standard ([RFC 7386](https://datatracker.ietf.org/doc/html/rfc7386#section-1)).

linera-service/src/cli/validator.rs
```diff
@@ -0,0 +1,1218 @@
...
+    pub votes: NonZero<u64>,
+}
+
+impl ValidatorSpec {
+    /// Validate the validator specification.
+    fn validate(&self) -> Result<()> {
+        if self.network_address.is_empty() {
+            bail!("Validator network address cannot be empty");
+        }
+        Ok(())
```
Twey, line 50: If `network_address` is an [`Url`](https://docs.rs/url/latest/url/struct.Url.html) this will be done at argument parse time and there's no need for a separate validation step.

In general it's good to push (static) validation into the parsing routines as much as possible, since then you get a standard failure experience for bad arguments.

linera-service/src/cli/validator.rs
eldios: I see, well specifying an empty object goes against our initial policy of 'only update what is specified' so there's no reason to specify a key with an empty value.. but I see how that may be confusing, and probably using `null` is semantically more significative of 'I just want to nullify/delete this.
So I'll just move to use `null` as you hinted 👍🏻

Twey: To be sure, there's no reason for a human to write an empty object there.  But if the input is generated programmatically (e.g. from a list of properties to be changed) or mechanically (e.g. using `sed`, or multiple cursors in an editor, or just by a user who's a bit undercaffeinated) it requires a bit more logic from the caller to handle the special case and make sure to remove the object to be unchanged rather than just leave it empty.

linera-service/src/cli/validator.rs
eldios: okay, I think I get what you're saying.. can you check again the latest changes? if they achieve what you had in mind, let's close this as done

linera-service/src/cli/validator.rs
eldios: sorry, true, fixed 👍🏻

Twey: Still looks like a string to me :thinking:

linera-service/src/cli/validator.rs
```diff
@@ -0,0 +1,1163 @@
...
+}
+
+/// Main entry point for handling validator commands.
+pub async fn handle_command<S, W, Si>(
+    context_options: ClientContextOptions,
+    storage: S,
+    wallet: W,
+    signer: Si,
+    block_cache_size: usize,
+    execution_state_cache_size: usize,
```
Twey: You always need a `ClientContext` here.  Why not construct it outside the function(s)?

```rust
impl ValidatorCommand {
    async fn run(self, context: impl ClientContext) -> Result<()> {
        …
    }
}
```

linera-service/src/cli/validator.rs
```diff
@@ -0,0 +1,1163 @@
...
+        .with_context(|| format!("Failed to parse query batch file: {}", path.display()))?;
+    Ok(batch)
+}
+
+/// Main entry point for handling validator commands.
+pub async fn handle_command<S, W, Si>(
+    context_options: ClientContextOptions,
+    storage: S,
+    wallet: W,
+    signer: Si,
```
Twey: The point of the `Environment` trait is to not have to unbundle all the elements of the environment.  You can either take an environment here:

```rust
pub async fn handle_command(
    environment: impl Environment,
    …
)```

or just take a client context directly (since you always construct one anyway):

```rust
pub async fn handle_command<E: Environment>(
    context: ClientContext<E>,
    …
)
```

or (if you aren't depending on any implementation details of the `ClientContext` struct) even abstract away the implementation using the `ClientContext` trait (`linera_client::chain_listener::ClientContext`):

```rust
pub async fn handle_command(context: impl ClientContext, …)
```

linera-service/src/cli/validator.rs
```diff
@@ -0,0 +1,1163 @@
...
+fn parse_query_batch_file(path: &Path) -> Result<ValidatorQueryBatch> {
+    let contents = std::fs::read_to_string(path)
+        .with_context(|| format!("Failed to read query batch file: {}", path.display()))?;
+    let batch: ValidatorQueryBatch = serde_json::from_str(&contents)
+        .with_context(|| format!("Failed to parse query batch file: {}", path.display()))?;
+    Ok(batch)
+}
+
+/// Main entry point for handling validator commands.
+pub async fn handle_command<S, W, Si>(
```
Twey: Nit: in general functions of the form `verb_object` are better written as inherent or trait methods `Object::verb`:

```rust
impl ValidatorCommand {
    pub async fn handle(…) {
        …
    }
}
```

(though I would prefer the verb ‘run’ because it's really just ‘handling itself’)

linera-service/src/cli/validator.rs
```diff
@@ -0,0 +1,1163 @@
...
+            min_votes,
+        } => {
+            let context = ClientContext::new(
+                storage,
+                context_options,
+                wallet,
+                signer,
+                block_cache_size,
+                execution_state_cache_size,
+            );
```
Twey: This seems to be duplicated for every branch.  It would probably be better to move it out of the `match` (or, better, out of the function entirely).

linera-service/src/cli/validator.rs
```diff
@@ -0,0 +1,1163 @@
...
+        return Ok(());
+    }
+
+    // Confirmation prompt (unless --yes flag is set)
+    if !yes {
+        println!("═════════════════════════════════════════════════════════════════════════════");
+        println!("⚠️  WARNING: This operation will modify the validator committee.");
+        println!("             Changes are permanent and will be broadcast to the network.");
+        println!("═════════════════════════════════════════════════════════════════════════════\n");
+        println!("Do you want to proceed? Type 'YES' (uppercase) to confirm: ");
```
Twey: We have a convention (that could be improved with proper output format flags, but should probably be left for now) of using `stdout` for machine-readable output and `stderr` for human-readable output.

eldios: fixed

eldios:
Okay I'm sticking with the implementation that works for now, I spent WAY too much time on this.. we can refine another time if/when needed.

PR #5018, opened 2025-11-24
Add a `first_leader` setting to chain ownership.

### LINKED ISSUE #5017: Chain option to make single-leader rounds always start with a particular leader.
An option for `ChainOwnership` that explicitly sets the first round leader.

This is useful for pools of backends for chains that want one node to be responsible for each chain, but in case of a node going offline, want another node to take over.

afck:
## Motivation

It can be useful to explicitly configure the leader of the first `SingleLeader` round.

## Proposal

Add a `first_leader` field to `ChainOwnership` for this.

The tests uncovered that the `compose` test actually used downloaded images instead of the newly built ones. (Claude found and fixed the issue.)

## Test Plan

CI; one test was updated to use `Some(owner)`.

## Release Plan

## Links

- Closes #5017.

### DIFF
--- a/.github/workflows/docker-compose.yml
+++ b/.github/workflows/docker-compose.yml
@@ -30,7 +30,7 @@ permissions:
 
 jobs:
   compose:
-    if: ${{ github.event_name == 'merge_group' }}
+    if: ${{ github.event_name == 'merge_group' || github.event_name == 'workflow_dispatch' }}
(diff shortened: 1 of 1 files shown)


ma2bd:
This makes sense to me. Letting others review as well

afck:
I tried the fixed `compose` test here: https://github.com/linera-io/linera-protocol/actions/runs/19734768216

PR #5019, opened 2025-11-24
Restructure the `MultiPartitionBatch` and the keys.

MathieuDutSik:
## Motivation

Due to the effort to make the database migration as good as possible, the database schema of `testnet_conway` ends up being better than the one of `main`.

## Proposal

Backport the changes from the `testnet_conway` to `main` (reverse of the usual way).
This covers two changes:
* The `MultiPartitionBatch` can support several `key/values` on the sme `root_key`.
* The keys are now no longer `[0]` and `[1]` which is a little better.

## Test Plan

The CI.

## Release Plan

Unify the `testnet_conway` and `main` is the end result of the PR.

## Links

None.

### DIFF
--- a/linera-storage/src/db_storage.rs
+++ b/linera-storage/src/db_storage.rs
@@ -1127,7 +1127,7 @@ mod tests {
         },
     };
 
-    use crate::db_storage::{to_event_key, RootKey, BLOB_ID_TAG, CHAIN_ID_TAG};
+    use crate::db_storage::{to_event_key, RootKey, BLOB_ID_TAG, CHAIN_ID_TAG, EVENT_ID_TAG};
 
     // Several functionalities of the storage rely on the way that the serialization
     // is done. Thus we need to check that the serialization works in the way that
@@ -1181,6 +1181,8 @@ mod tests {
             stream_id,
             index,
         };
+        let root_key = RootKey::Event(chain_id).bytes();
+        assert_eq!(root_key[0], EVENT_ID_TAG);
         let key = to_event_key(&event_id);
         assert!(key.starts_with(&prefix));
     }


linera-storage/src/db_storage.rs
```diff
@@ -225,55 +225,58 @@ pub mod metrics {
...
-fn get_01_keys() -> Vec<Vec<u8>> {
-    vec![vec![0], vec![1]]
+/// The key used for lite certificates. The cryptohash itself is contained in the root key.
+const LITE_CERTIFICATE_KEY: &[u8] = &[91];
+
+/// The key used for confirmed blocks. The cryptohash itself is contained in the root key.
+const BLOCK_KEY: &[u8] = &[221];
+
+/// The key used for the network description.
+const NETWORK_DESCRIPTION_KEY: &[u8] = &[119];
```
afck: Where do all those numbers come from?

linera-storage/src/db_storage.rs
```diff
@@ -225,55 +225,58 @@ pub mod metrics {
...
+
+fn get_block_keys() -> Vec<Vec<u8>> {
+    vec![LITE_CERTIFICATE_KEY.to_vec(), BLOCK_KEY.to_vec()]
 }
 
 #[derive(Default)]
+#[allow(clippy::type_complexity)]
 struct MultiPartitionBatch {
-    keys_value_bytes: Vec<(Vec<u8>, Vec<u8>, Vec<u8>)>,
+    keys_value_bytes: Vec<(Vec<u8>, Vec<(Vec<u8>, Vec<u8>)>)>,
```
afck: Sorry if I've asked this before, but: Why aren't we using a map here? Wouldn't it make sense to sort the key-value pairs by root key?

linera-storage/src/db_storage.rs
```diff
@@ -330,11 +340,12 @@ pub struct DbStorage<Database, Clock = WallClock> {
 #[derive(Debug, Serialize, Deserialize)]
 enum RootKey {
     ChainState(ChainId),
-    CryptoHash(CryptoHash),
+    ConfirmedBlock(CryptoHash),
     Blob(BlobId),
     Event(ChainId),
-    BlockExporterState(u32),
+    Placeholder,
```
afck: Is this to make it match the testnet somehow?

linera-storage/src/db_storage.rs
```diff
@@ -330,11 +340,12 @@ pub struct DbStorage<Database, Clock = WallClock> {
...
     NetworkDescription,
+    BlockExporterState(u32),
 }
```
ma2bd, line 349: I would actually base the RootKey on whatever data is available for partitioning:
```
enum RootKey {
    NetworkDescription,
    BlockExporterIndex(u32),
    ChainId(ChainId),
    BlockHash(CryptoHash),     // updated
    BlobId(BlobId),
}
```

afck: But can we call `CryptoHash` `BlockHash`?
(We might end up adding something else that's partitioned by hash. Also, `ChainId` also just wraps a `CryptoHash`. Or would it make sense to introduce a `BlockHash` type at some point…?)

ma2bd: You're right. BlockHash is a better than CryptoHash

linera-storage/src/db_storage.rs
MathieuDutSik: They have been chosen somewhat at random.

linera-storage/src/db_storage.rs
MathieuDutSik: Yes.

linera-storage/src/db_storage.rs
MathieuDutSik: The only case where this makes sense is the `write_events`, but ok, we can do it.

linera-storage/src/db_storage.rs
afck: Is there a reason to not just use 0, 1, 2, …?

linera-storage/src/db_storage.rs
afck: Is that necessary?
Should we at least create a ticket and TODO to remove it before testnet 4?

linera-storage/src/db_storage.rs
MathieuDutSik: The problem with making `ChainId` cover both events and ChainState is that for ChainState, we rely on having exclusive access.

linera-storage/src/db_storage.rs
MathieuDutSik: Since other reviewers are fine with changing the orders of the entries, I guess we can remove the `PlaceHolder`.

linera-storage/src/db_storage.rs
MathieuDutSik: Yes, changed to that.
It was somewhat of a good idea when replacing the `DEFAULT_KEY= [0]` / `ONE_KEY=[1]`.
I still think, it is better since those numbers are easy to search in logs.

linera-storage/src/db_storage.rs
ma2bd: oh yes of course! sorry

linera-storage/src/db_storage.rs
afck: (OK, I'm fine with either. If you use magic numbers for the logs, that should be clearly documented. But I imagine searching numbers <256 will always return many false positives anyway.)

PR #5020, opened 2025-11-24
Remove the `InactiveChain` from the code.

### LINKED ISSUE #3927: Remove `InactiveChain`/`BlobsNotFound` ambiguity.

MathieuDutSik:
## Motivation

The `InactiveChain` and `BlobsNotFound` are used somewhat interchangeably in the code. This has to be addressed.

Fixes #3927 

## Proposal

The `BlobsNotFound` carries an error with it, while `InactiveChain` does not. In truth, it is impossible to determine whether a chain is active or not in the code, since it is unclear whether this is an error or not.

Potentially, we could add processing of the `BlobsNotFound` errors in the `client`.

## Test Plan

The CI.

## Release Plan

The change of errors means that it is difficult to see whether we could backport it to `testnet_conway`.

## Links

None.

### DIFF
--- a/linera-core/src/client/chain_client/mod.rs
+++ b/linera-core/src/client/chain_client/mod.rs
@@ -626,7 +626,7 @@ impl<Env: Environment> ChainClient<Env> {
         let manager = self.chain_info().await?.manager;
         ensure!(
             manager.ownership.is_active(),
-            LocalNodeError::InactiveChain(self.chain_id)
+            LocalNodeError::BlobsNotFound(vec![BlobId::chain(self.chain_id)])
         );
 
         let is_owner = manager
@@ -1950,7 +1950,7 @@ impl<Env: Environment> ChainClient<Env> {
             let ownership = self.prepare_chain().await?.manager.ownership;
             ensure!(
                 ownership.is_active(),
-                ChainError::InactiveChain(self.chain_id)
+                ChainError::BlobsNotFound(vec![BlobId::chain(self.chain_id)])
             );
             let mut owners = ownership.owners.into_iter().collect::<Vec<_>>();
             owners.extend(ownership.super_owners.into_iter().zip(iter::repeat(100)));
@@ -2429,7 +2429,7 @@ impl<Env: Environment> ChainClient<Env> {
                 self.client.update_from_info(&info);
                 Ok(Some(info))
             }
-            Err(LocalNodeError::BlobsNotFound(_) | LocalNodeError::InactiveChain(_)) => Ok(None),
+            Err(LocalNodeError::BlobsNotFound(_)) => Ok(None),
             Err(err) => Err(err.into()),
         }
     }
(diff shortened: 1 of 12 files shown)


linera-core/src/client/chain_client/mod.rs
```diff
@@ -1950,7 +1950,7 @@ impl<Env: Environment> ChainClient<Env> {
             let ownership = self.prepare_chain().await?.manager.ownership;
             ensure!(
                 ownership.is_active(),
-                ChainError::InactiveChain(self.chain_id)
+                ChainError::BlobsNotFound(vec![BlobId::chain(self.chain_id)])
```
ma2bd, line 1953: It looks very weird to me.

ma2bd (changes requested):
I don't understand why we would generate BlobsNotFound if nobody asked for a blob.

linera-core/src/client/chain_client/mod.rs
MathieuDutSik: In the end, it is what it is: A chain being active is the same as a blob existing. So, that seems reasonable to me.

Furthermore, that logic is exposed to the user directly because if he does a `client3.follow_chain(wrong_id, false).await` then he gets from the command line:
```
2025-11-25T06:48:25.494142Z ERROR linera: Error is Remote node operation failed: Blobs not found: [BlobId { blob_type: ChainDescription, hash: e7e3a7314598ba7afe6fa2c281c4a1b0a0eb75d6db43f2d2616acc60b3cba5b2 }]

Caused by:
    Blobs not found: [BlobId { blob_type: ChainDescription, hash: e7e3a7314598ba7afe6fa2c281c4a1b0a0eb75d6db43f2d2616acc60b3cba5b2 }]
```

So, the PR simply exposes to the programmer what has long been known to the user. I know it is weird, but if chains are encoded as Blobs, then that is what it is.

afck: Right, this particular instance should probably be a new error: Here we have a chain with no owners, which is really a different _kind_ of not being "active". Here we aren't missing the blob.

linera-core/src/local_node.rs
```diff
@@ -357,14 +354,18 @@ impl LocalChainInfoExt for ChainInfo {
         self.requested_committees
             .ok_or(LocalNodeError::InvalidChainInfoResponse)?
             .remove(&self.epoch)
-            .ok_or(LocalNodeError::InactiveChain(self.chain_id))
+            .ok_or(LocalNodeError::BlobsNotFound(vec![BlobId::chain(
```
afck, line 357: Would the server return a chain info at all, if it doesn't have the chain description?

MathieuDutSik:
This was not considered a correct solution to the problem.

PR #5021, opened 2025-11-24
Make the port of the validators in e2e tests configurable via environment variable

### LINKED ISSUE #4114: Remove hardcoded ports for database tests
Port values should be taken from environment variables.

MathieuDutSik:
## Motivation

The port tests are being hardcoded, which creates some trouble if one wants to run two end-to-end tests on the same computer.

Fixes #4114 

## Proposal

Add an environment variable `TEST_OFFSET_PORT` as an environment variable.

## Test Plan

The CI.

## Release Plan

It could be backported to `testnet_conway`.

## Links

None.

### DIFF
--- a/linera-service/src/cli_wrappers/local_net.rs
+++ b/linera-service/src/cli_wrappers/local_net.rs
@@ -49,8 +49,6 @@ use crate::{
 /// Maximum allowed number of shards over all validators.
 const MAX_NUMBER_SHARDS: usize = 1000;
 
-pub const FIRST_PUBLIC_PORT: usize = 13000;
-
 pub enum ProcessInbox {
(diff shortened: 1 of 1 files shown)


deuszx (commented):
There's a `get_free_port` in `linera_base/src/port.rs`. Can you use that instead?

MathieuDutSik:
> There's a `get_free_port` in `linera_base/src/port.rs`. Can you use that instead?

No.
The ports have to be coherent between all the validators/proxies. So, randomness is not possible.

The random port is used for the node services.

ma2bd (approved):
LGTM but I couldn't help fixing the title

PR #5024, opened 2025-11-26
Handling cross-chain messages in batches.

afck:
## Motivation

The current implementation processes cross-chain updates and their confirmations one at a time, even though they often affect different inboxes or outboxes, that could be updated concurrently.

## Proposal

This PR (With the help of Claude Opus 4.5) introduces batching and fair scheduling for cross-chain message processing:

* Separate queues with round-robin scheduling: Split the chain worker actor's single request queue into three separate queues: regular requests, cross-chain updates, and cross-chain confirmations. The actor rotates through all three queues, ensuring fair processing and preventing any queue from starving the others.
* Batch processing: Cross-chain updates and confirmations are batched together (up to a configurable limit) and processed in a single call to the chain worker, with a single database write transaction.
* Delivery notification fix: Never send delivery notifications for heights `next_block_height` and above.

## Test Plan

CI

We should run the benchmark with this to see if this improves performance.

## Release Plan

## Links

### DIFF
--- a/linera-core/src/worker.rs
+++ b/linera-core/src/worker.rs
@@ -18,7 +18,7 @@ use linera_base::{
     hashed::Hashed,
     identifiers::{AccountOwner, ApplicationId, BlobId, ChainId, EventId, StreamId},
     time::Instant,
-    util::traits::DynError,
+    util::traits::{AutoTraits, DynError},
 };
 #[cfg(with_testing)]
 use linera_chain::ChainExecutionContext;
@@ -913,7 +913,7 @@ where
         nickname = %self.nickname,
         chain_id = %chain_id
     ))]
-    fn send_chain_worker_request<T: Send + Sync + 'static>(
+    fn send_chain_worker_request<T: AutoTraits>(
         &self,
         chain_id: ChainId,
         send_fn: impl FnOnce(
--- a/linera-core/src/chain_worker/actor.rs
+++ b/linera-core/src/chain_worker/actor.rs
@@ -270,7 +270,7 @@ enum RequestType {
 /// The receiver endpoints for a [`ChainWorkerActor`].
 pub(crate) struct ChainActorReceivers<Context>
 where
-    Context: linera_views::context::Context + Clone + Send + Sync + 'static,
+    Context: linera_views::context::Context + Clone + 'static,
 {
     /// Receiver for regular chain worker requests.
     pub requests: mpsc::UnboundedReceiver<(ChainWorkerRequest<Context>, tracing::Span, Instant)>,


linera-chain/src/chain.rs
```diff
@@ -426,62 +426,87 @@ where
...
 
-    #[instrument(skip_all, fields(
-        chain_id = %self.chain_id(),
-        target = %target,
-        height = %height
-    ))]
+    /// Marks messages as received for multiple recipients.
+    /// Takes a map from recipient to the maximum confirmed height.
+    ///
+    /// Returns `true` if any messages were marked as received.
```
deuszx, line 388: Is this safe? Shouldn't we return `true` when ALL messages are received?

linera-chain/src/chain.rs
```diff
@@ -426,62 +426,87 @@ where
...
-            *counter = counter.checked_sub(1).ok_or(ArithmeticError::Underflow)?;
-            if *counter == 0 {
-                // Important for the test in `all_messages_delivered_up_to`.
-                self.outbox_counters.get_mut().remove(&update);
+        // Load all outboxes at once.
+        let recipients = max_height_by_recipient
+            .keys()
+            .copied()
+            .collect::<Vec<ChainId>>();
+        let loaded_outboxes = self.outboxes.try_load_entries_pairs_mut(recipients).await?;
```
deuszx, line 402: IDK if that's the problem in reality but this is a pattern that had slowed us down in the past: loading a lot of data at once from the storage

deuszx: This is chain IDs so it shouldn't be too high.

linera-chain/src/chain.rs
afck: This is used to tell the caller whether any change was made at all.

That way, if it returns `false`, the caller can avoid unnecessarily computing whether a delivery notifier needs to be triggered.

The cross-chain message handshake protocol is designed to tolerate losing or duplicating messages between the workers. If all processed cross-chain confirmations were duplicates of earlier ones, this is a no-op and returns `false`.

linera-core/src/worker.rs
```diff
@@ -823,100 +875,162 @@ where
...
+        if let Some(receivers) = receivers {
+            drop(chain_workers);
+            self.spawn_chain_worker_actor(chain_id, receivers);
+        }
+
+        Ok(())
+    }
+
+    /// Spawns a new chain worker actor.
+    fn spawn_chain_worker_actor(
```
deuszx, line 956: note: I haven't yet seen the whole PR but is it OK to spawn a new chain worker actor if we already have one for this chain_id? Multiple chain worker actors will all be trying to access the same state (most probably) and it can lead to lock issues.

deuszx: Ah, you check if there's already a chain_worker before calling this.

linera-core/src/worker.rs
```diff
@@ -823,100 +875,162 @@ where
...
+            })?
     }
 
-    /// Find an endpoint and call it. Create the endpoint if necessary.
-    #[instrument(level = "trace", skip(self), fields(
+    /// Sends a request to a chain worker, creating the worker if necessary.
+    ///
+    /// This method holds the lock on the chain workers map while sending to ensure that
+    /// if the send fails (because the actor was dropped), no other thread can get a stale
+    /// endpoint. The endpoint is removed before sending and only re-inserted on success.
```
deuszx, line 911: Due to the lock on `chain_workers` we can still process requests sequentially only. You write that this is done so in order to not give requests stale endpoints but this should be happening rarely (actor dropping) so maybe we shouldn't be slowing everything down b/c a "bad thing may happen" and we should code for the happy path – workers are not dropped.

afck: We aren't holding the lock for the whole time the request is being processed! We only hold it until we know that `send` succeeded, which should be really fast and doesn't involve I/O.

This was already how it worked before this PR, and we did it to make sure that we don't accidentally drop the wrong actor, which may still be processing other requests. That could lead to doubly-loaded chain state views again.

linera-chain/src/chain.rs
```diff
@@ -426,62 +426,87 @@ where
...
+                    // Important for the test in `max_height_with_all_messages_delivered`.
+                    self.outbox_counters.get_mut().remove(update);
+                }
+            }
+            if outbox.queue.count() == 0 {
+                self.nonempty_outboxes.get_mut().remove(recipient);
+                // If the outbox is empty and not ahead of the executed blocks, remove it.
+                if *outbox.next_height_to_schedule.get() <= self.tip_state.get().next_block_height {
+                    self.outboxes.remove_entry(recipient)?;
+                }
```
deuszx, line 446: I think this part is changed from before: in the previous version if `updates` was empty we'd short-circuit and return `false` early. Here we update the `outbox` regardless. Is it a no-op?

afck: Done in https://github.com/linera-io/linera-protocol/pull/5024/commits/7f9da1ae010c13242185238fcf3e46a119bc5b9a.

linera-core/src/chain_worker/state.rs
```diff
@@ -986,80 +981,104 @@ where
...
-    #[instrument(skip_all, fields(
-        chain_id = %self.chain_id(),
-        recipient = %recipient,
-        latest_height = %latest_height
-    ))]
-    async fn confirm_updated_recipient(
+    /// Handles a batch of cross-chain confirmation requests efficiently with a single save operation.
+    /// Takes a map from recipient to the maximum confirmed height.
+    #[instrument(level = "trace", skip(self, confirmations))]
+    pub(super) async fn confirm_updated_recipient(
```
deuszx: s/confirm_updated_recipient/confirm_updated_recipients
?

ndr-ds:
This does seem to improve latency a bit, and decreases CPU usage as well, so it's a net positive.
I will play with the batch sizes to see if I can make it scale our message processing as well, and will report back here.

afck:
Closing this for now: The benchmarks didn't show a clear improvement.

PR #5031, opened 2025-11-27
Omit the decimal point if amount is an integer.

afck:
## Motivation

An amount 5 tokens is currently displayed as "5.", even though the dot is not needed for integers.

## Proposal

Omit the dot.

## Test Plan

Test was updated.

## Release Plan

    - be released in a new SDK.

## Links

### DIFF
--- a/linera-base/src/data_types.rs
+++ b/linera-base/src/data_types.rs
@@ -558,10 +558,11 @@ impl Display for Amount {
         // For now, we never trim non-zero digits so we don't lose any precision.
         let precision = f.precision().unwrap_or(0).max(fractional_part.len());
         let sign = if f.sign_plus() && self.0 > 0 { "+" } else { "" };
+        let dot = if precision == 0 { "" } else { "." };
         // The amount of padding: desired width minus sign, point and number of digits.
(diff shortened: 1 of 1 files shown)


ma2bd:
@afck I didn't get what the problem was with the dot, did you?

afck:
No problem, just surprising for users maybe? We can keep it, if you prefer. Or as @deuszx suggests always print at least one digit after it?

ma2bd:
> No problem, just surprising for users maybe? We can keep it, if you prefer. Or as @deuszx suggests always print at least one digit after it?

I feel that we need to understand what applications need a little better, and perhaps solve the issue on the Javascript side of things.

deuszx:
> > No problem, just surprising for users maybe? We can keep it, if you prefer. Or as @deuszx suggests always print at least one digit after it?
> 
> I feel that we need to understand what applications need a little better, and perhaps solve the issue on the Javascript side of things.

I don't think it's "JS side" of things. JS uses Wasm code and here it's a `Display` impl for those types. And indeed `5.` looks weird. I'd prefer if it was `5.0`.

PR #5033, opened 2025-11-27
Cleanups and minor fixes to rounds, ownership and leader calculation.

afck:
## Motivation

Much of the code related to rounds and chain owners is hard to read and sometimes inconsistent.

* The logic for computing the round leader is currently repeated in three places.
* The client would fail to recognize that fallback owners are owners unless they are the current leader.
* The client could fail in some edge cases because `round_for_new_proposal` is used to determine the validation round oracle value.

## Proposal

* Deduplicate round leader calculation.
* Extract some code into `ChainOwnership` methods; rename `verify_owner` to `is_owner`.
* Super owners _can_ propose in all except validator rounds, because in fallback mode the validators are supposed to override them and receive mandatory messages. They _should_ avoid proposing in other owners' single-leader rounds, however, so that will be the client's default behavior.
* Fix the fallback ownership check in the client.
* Fix the validation round oracle value computation: Even if an error is returned because the current round is `Fast` and we should not propose in any later round, set the value to `None` and continue.

## Test Plan

A regression test for the validation round oracle issue was added.

## Release Plan

    - be released in a new SDK,

## Links

### DIFF
--- a/linera-chain/src/manager.rs
+++ b/linera-chain/src/manager.rs
@@ -625,8 +625,9 @@ where
     /// proposal's round.
     ///
     /// Super owners can always propose, except in `Validator` rounds, but it is recommended that
-    /// they only propose in `Fast` rounds. In multi-leader rounds, any regular owner can propose
-    /// (or anyone, if `open_multi_leader_rounds`) and in other rounds there is only one leader.
+    /// they don't interfere with single-leader rounds. In multi-leader rounds, any owner can
+    /// propose (or anyone, if `open_multi_leader_rounds`) and in other rounds there is only
+    /// one leader.
     pub fn can_propose(&self, owner: &AccountOwner, round: Round) -> bool {
         let ownership = self.ownership.get();
         if ownership.super_owners.contains(owner) {
@@ -842,8 +843,10 @@ impl ChainManagerInfo {
             .map(|vote| Box::new(vote.value.clone()));
     }
 
-    /// Returns whether the `identity` is allowed to propose a block in `round`, except that it
-    /// returns `false` for super owners in single-leader rounds unless they are the leader.
+    /// Returns whether the `identity` is allowed to propose a block in `round`.
+    ///
+    /// **Exception:** In single-leader rounds, a **super owner** should only propose
+    /// if they are the designated **leader** for that round.
     pub fn should_propose(
         &self,
         identity: &AccountOwner,
@@ -898,8 +901,8 @@ fn calculate_distribution<'a, T: 'a>(
     }
 }
 
-/// Returns the leader for single-leader or validator rounds, or `None` if there is no designated
-/// leader (i.e. in fast or multi-leader rounds).
+/// Returns the designated leader for single-leader or validator rounds.
+/// Returns `None` for fast or multi-leader rounds.
 fn compute_round_leader<'a>(
     round: Round,
     seed: u64,


linera-base/src/ownership.rs
```diff
@@ -156,8 +156,17 @@ impl ChainOwnership {
...
-        self.super_owners.contains(owner) || self.owners.contains_key(owner)
+    pub fn is_owner(&self, owner: &AccountOwner) -> bool {
+        self.super_owners.contains(owner)
+            || self.owners.contains_key(owner)
+            || self.first_leader.as_ref().is_some_and(|fl| fl == owner)
+    }
+
+    /// Returns `true` if this is an owner or if `open_multi_leader_rounds`.
+    pub fn is_multi_leader_owner(&self, owner: &AccountOwner) -> bool {
+        self.open_multi_leader_rounds
```
deuszx, line 167: This will return `true` even if `owner` isn't even't in `self.owners` or `self.super_owners` collections.

deuszx: Same for the function above.

linera-chain/src/manager.rs
```diff
@@ -881,9 +842,9 @@ impl ChainManagerInfo {
             .map(|vote| Box::new(vote.value.clone()));
     }
 
-    /// Returns whether the `identity` is allowed to propose a block in `round`.
-    /// This is dependent on the type of round and whether `identity` is a validator or (super)owner.
-    pub fn can_propose(
+    /// Returns whether the `identity` is allowed to propose a block in `round`, except that it
+    /// returns `false` for super owners in single-leader rounds unless they are the leader.
```
deuszx: I find this comment quite confusing, I think it includes a double negation.

Alternative
```
/// Returns whether the `identity` is **allowed to propose a block** in `round`.
/// **Exception:** In single-leader rounds, a **super owner** is only allowed to propose if they are the designated **leader** for that round.
```

linera-chain/src/manager.rs
```diff
@@ -948,6 +898,35 @@ fn calculate_distribution<'a, T: 'a>(
     }
 }
 
+/// Returns the leader for single-leader or validator rounds, or `None` if there is no designated
+/// leader (i.e. in fast or multi-leader rounds).
```
deuszx:
```suggestion
/// Returns the designated leader for single-leader or validator rounds.
/// Returns `None` for fast or multi-leader rounds.
```

linera-client/src/chain_listener.rs
```diff
@@ -344,8 +345,7 @@ impl<C: ClientContext + 'static> ChainListener<C> {
                 if blob_id.blob_type == BlobType::ChainDescription {
                     let chain_desc: ChainDescription = bcs::from_bytes(blob.content().bytes())
                         .expect("ChainDescription should deserialize correctly");
-                    let owners = chain_desc.config().ownership.all_owners().cloned();
-                    Some((ChainId(blob_id.hash), owners.collect::<Vec<_>>()))
+                    Some((ChainId(blob_id.hash), chain_desc))
```
deuszx, line 348: Were these changes necessary?

afck: No, but it avoids some unnecessary cloning.

linera-base/src/ownership.rs
afck: Yes, but I guess the `first_leader` has to be considered an owner? They can propose blocks, after all.

afck: Maybe this method should be `can_propose_in_multi_leader_round`?

deuszx (approved):
I have to say that it's difficult to find exactly where the bug was. PR includes more than just the bug fix and even when reviewing commit-by-commit it's unclear where the root of the fix is. Next time, when you create bugfix PRs, please try to do more targeted ones.

afck:
Sorry! Originally I meant to just deduplicate the leader selection, then I noticed a bunch of naming issues and that bug. :sweat_smile:

Issue #5035, opened 2025-11-27
Add more Web tests


Twey:
We have a basic testing framework for the Web now, but we should have some more real-world testing scenarios, including:

- application execution
- cross-application calls
- event subscriptions

PR #5036, opened 2025-11-27
`linera-execution`: preload required applications

Twey:
## Motivation

It turns out that we never preloaded the required applications for an application and relied on dynamic application loading to load the dependencies just-in-time.  Unfortunately, dynamic application loading doesn't work on the Web, which means we can't currently make cross-application calls there even in the static case where the dependencies are known ahead of time.

## Proposal

Recursively load the tree of application dependencies before executing the application call, both in service and contract.

## Test Plan

It would be nice to have a test for this, but I've postponed it as https://github.com/linera-io/linera-protocol/issues/5035.

## Release Plan

    - be released in a new SDK,

(Hopefully, the execution with preloading is equivalent to the execution with dynamic loading.)

## Links

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11466,9 +11466,9 @@ dependencies = [
 
 [[package]]
 name = "web-thread"
-version = "0.2.1"
+version = "0.2.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


ma2bd:
Superseded by #5042

PR #5039, opened 2025-11-28
Ignore missing cert errors if it's expected. Load it from env

deuszx:
## Motivation

We know that on testnet some certs are missing so it doesn't make sense to raise critical errors on them which crash the exporter.

## Proposal

Load, from an env var `LINERA_KNOWN_MISSING_CERTS`, list of hashes that are expected to be missing. If we fail to read a certificate that is on that list, we don't treat it as an error.

## Release Plan

- Testnet only.
-
## Links

### DIFF
--- a/linera-service/src/exporter/runloops/block_processor/mod.rs
+++ b/linera-service/src/exporter/runloops/block_processor/mod.rs
@@ -148,7 +148,11 @@ where
 
                         Err(ExporterError::ReadCertificateError(hash)) => {
                             if self.known_missing_certs.contains(&hash) {
-                                tracing::debug!(?hash, "certificate known to be missing, skipping retry");
+                                tracing::debug!(
+                                    block_id=?next_block_notification,
+                                    missing_cert=?hash,
+                                    "certificate known to be missing, skipping retry of this notification"
+                                );
                                 continue;
                             }
                             match self.retried_certs.remove(&hash) {


linera-service/src/exporter/runloops/block_processor/mod.rs
```diff
@@ -138,6 +147,10 @@ where
                         },
 
                         Err(ExporterError::ReadCertificateError(hash)) => {
+                            if self.known_missing_certs.contains(&hash) {
+                                tracing::debug!(?hash, "certificate known to be missing, skipping retry");
+                                continue;
```
afck, line 156: (That means all descendants will also fail, does it?)

deuszx: RIght now it does: 

when we receive a notification we start "walking" its dependencies in a DFT and try to first export all missing ones (which we haven't yet marked as "visited" in the exporter's storage). So, if there's some missing cert at height H, and we receive a notificiation for H+1 that is in the storage, we will try to export H first. If that cert is missing from the validator's storage we'd raise the error here and eventually crash. We won't crash anymore but we will skip all notifications that lead to this error. 

I will log the `block_id` that lead to this.

deuszx: Done in [Improve the log](https://github.com/linera-io/linera-protocol/pull/5039/commits/9600fc2de6b6148cefcf5c696b9f59f24c4655b7)

PR #5040, opened 2025-11-28
Modify tests for historical hashing

bart-linera:
## Motivation

After the `ExecutionStateView` switches to historical hashing, it will be very hard to manually compute a hash of a given execution state, since it will depend not on the state itself, but on its history. That will break a lot of the chain worker tests.

## Proposal

Make the chain worker tests create certificates (and thus compute hashes inside) by executing the corresponding blocks on an auxiliary worker, then assert that the execution outcomes are as expected.

Two tests are being removed:
- `test_handle_certificate_with_open_chain`, as it has only been testing sending a transfer to a child chain,
- `test_handle_certificate_wrong_owner`, as it tested a very artificial scenario of a wrong owner being used to initialize a state used to generate a certificate, and the state hash was the only part of the certificate being affected, which we can no longer test for.

## Test Plan

CI

## Release Plan

## Links

- prerequisite for #4977 

### DIFF
--- a/linera-chain/src/block.rs
+++ b/linera-chain/src/block.rs
@@ -588,6 +588,7 @@ impl Block {
     }
 
     /// Returns whether the outcomes of the block's execution match the passed values.
+    #[cfg(with_testing)]
     #[expect(clippy::too_many_arguments)]
     pub fn outcome_matches(
         &self,


linera-chain/src/block.rs
```diff
@@ -587,6 +587,27 @@ impl Block {
             && *previous_block_hash == self.header.previous_block_hash
     }
 
+    /// Returns whether the outcomes of the block's execution match the passed values.
+    #[expect(clippy::too_many_arguments)]
+    pub fn outcome_matches(
```
deuszx, line 593: Why not derive `Eq` and use that instead?

linera-core/src/unit_tests/worker_tests.rs
```diff
@@ -127,43 +127,55 @@ where
...
+                5_000,
+                10_000,
+            )
+            .with_allow_inactive_chains(is_client)
+            .with_allow_messages_from_deprecated_epochs(is_client)
+            .with_long_lived_services(has_long_lived_services)
+            .with_grace_period(Duration::from_micros(TEST_GRACE_PERIOD_MICROS))
+        };
+
+        let worker = make_worker(ValidatorKeypair::generate()).await;
```
deuszx, line 174: The worker that we use just to execute a block (btw, I'd change the naming – worker vs executing worker is confusing, I thought the `executing_worker` will be the one that just do the block execution) shouldn't need to be a validator. I think you can just pass `None` here?

linera-core/src/unit_tests/worker_tests.rs
```diff
@@ -84,6 +79,7 @@ const TEST_GRACE_PERIOD_MICROS: u64 = 500_000;
 struct TestEnvironment<S: Storage> {
     committee: Committee,
     worker: WorkerState<S>,
+    executing_worker: WorkerState<S>,
```
deuszx, line 88: This is THE PLACE that fixes the tests. Please add a comment why we need two workers now.

linera-core/src/unit_tests/worker_tests.rs
```diff
@@ -473,6 +549,22 @@ where
             ..SystemExecutionState::new(description.clone())
         }
     }
+
+    async fn execute_proposal(
```
deuszx, line 558: It'd be good to have a comment here reminding why we're using different worker.

deuszx (approved):
How does the new "execution worker" invalidates the changes to its state? I'm thinking about a situation when validator rejects the proposal – not committing the changes – hance the proposal will not affect future state hashes.

linera-chain/src/block.rs
```diff
@@ -587,6 +587,27 @@ impl Block {
...
+        &self,
+        messages: Vec<Vec<OutgoingMessage>>,
+        previous_message_blocks: BTreeMap<ChainId, (CryptoHash, BlockHeight)>,
+        previous_event_blocks: BTreeMap<StreamId, (CryptoHash, BlockHeight)>,
+        oracle_responses: Vec<Vec<OracleResponse>>,
+        events: Vec<Vec<Event>>,
+        blobs: Vec<Vec<Blob>>,
+        operation_results: Vec<OperationResult>,
+    ) -> bool {
+        self.body.messages == messages
```
afck: We could `let BlockBody { … } = &self.body` so the compiler warns us if we forget any fields, and it's obvious to the reader if any fields are deliberately ignored.

linera-chain/src/block.rs
afck, line 593: Should this be in `BlockTestExt` instead?

linera-core/src/unit_tests/wasm_worker_tests.rs
```diff
@@ -195,48 +183,29 @@ where
...
     let application_description_blob = Blob::new_application_description(&application_description);
-    let application_description_blob_id = application_description_blob.id();
     let application_id = From::from(&application_description);
     let create_block = make_first_block(creator_chain.id())
         .with_timestamp(2)
         .with_operation(create_operation);
-    creator_system_state.timestamp = Timestamp::from(2);
-    let mut creator_state = creator_system_state.into_view().await;
-    creator_state
-        .simulate_instantiation(
```
afck, line 206: Can we remove this function now?

linera-core/src/unit_tests/worker_tests.rs
```diff
@@ -333,6 +378,37 @@ where
     /// the `previous_confirmed_block` also did.
```
afck: This is probably actually fixed now, is it?

linera-core/src/unit_tests/worker_tests.rs
```diff
@@ -333,6 +378,37 @@ where
...
+
+    /// Creates a certificate with a transfer.
+    ///
+    /// This does not work for blocks with ancestors that sent a message to the same recipient, unless
+    /// the `previous_confirmed_block` also did.
+    /// It also does not work as a certificate that can be processed - it doesn't do
+    /// proper hashing of the execution state, so such a certificate will be rejected by
+    /// the worker.
+    #[expect(clippy::too_many_arguments)]
+    async fn make_transfer_certificate_for_epoch_unprocessable(
```
afck, line 414: Do we still need this?

linera-core/src/unit_tests/worker_tests.rs
afck: And to make it clear _that_ we're using a different worker. From the method name I'd have expected `worker` to execute the proposal.

linera-chain/src/block.rs
bart-linera: This checks the equality of only a subset of the fields. We could introduce a new structure just for tests, and then compare that, but I'm not sure if it's worth it.

linera-core/src/unit_tests/worker_tests.rs
bart-linera: This worker still needs a keypair, the argument is not an `Option`. It's not a validator, though - it's not included in the committee. Only the `executing_worker`'s public key is in the committee.

As for the naming: I also considered a name like `certificate_worker` or `certificate_creator`, because for some of the tests it is only used for creating certificates, but some tests also assert the changes of state in a worker that is assumed to be a validator's worker, so it has to also be the one that is used to make the assertions. Now that I think about it... how about `validator_worker`, since only this worker actually has some voting power?

linera-core/src/unit_tests/worker_tests.rs
bart-linera: Yes, some tests exercise scenarios with certificates containing invalid data, and those can't be created by executing a proposal on a worker.

linera-chain/src/block.rs
bart-linera: Makes sense :+1:

linera-core/src/unit_tests/worker_tests.rs
deuszx: > This worker still needs a keypair

Why? it shouldn't be signing anything.

bart-linera: Oh, nevermind, I was blind. The argument to `WorkerState::new` _is_ optional :man_facepalming: I think you're right, then :+1:

linera-core/src/unit_tests/worker_tests.rs
bart-linera: Right, it should be fixed :+1:

linera-core/src/unit_tests/worker_tests.rs
bart-linera: There is one test in which the non-validator worker having a key is actually useful: `test_handle_block_proposal_with_chaining`.

In that test, we check that we don't handle a child proposal until we have confirmed its parent. To do that, we need the child proposal, and to have a child proposal, we need the certificate for the parent: which means that we need to process the parent on the validator worker just to prepare the test values, so it's no longer suitable for testing behavior dependent on not having processed the parent block. But we still want to check whether a worker processes stuff correctly, including creation of validated and confirmed votes - this requires keys, and the easiest solution to test this is to just initialize the other worker with a keypair as well.

linera-chain/src/block.rs
bart-linera: Whoops, actually, it's not such a good idea :grimacing: We only implement `BlockTestExt` for `ProposedBlock` at the moment, which doesn't have outcomes, and we don't implement it for a `Block`, which does. It would have to be in a separate trait for actual `Block`s - what do you think?

linera-core/src/unit_tests/wasm_worker_tests.rs
bart-linera: We still use it in REVM tests.

linera-chain/src/block.rs
afck: Ah, sorry, I mixed those up!

Then maybe just make it `cfg(with_testing)`?

linera-core/src/unit_tests/worker_tests.rs
bart-linera: Added in [3f43436](https://github.com/linera-io/linera-protocol/pull/5040/commits/3f434363b6b4a25a6431984a28b9561241fbce95)

linera-core/src/unit_tests/worker_tests.rs
bart-linera: Added a comment in [3f43436](https://github.com/linera-io/linera-protocol/pull/5040/commits/3f434363b6b4a25a6431984a28b9561241fbce95)

linera-chain/src/block.rs
bart-linera: Applied in [3f43436](https://github.com/linera-io/linera-protocol/pull/5040/commits/3f434363b6b4a25a6431984a28b9561241fbce95)

bart-linera:
> How does the new "execution worker" invalidates the changes to its state? I'm thinking about a situation when validator rejects the proposal – not committing the changes – hance the proposal will not affect future state hashes.

I'm not sure what you're asking about. If the validator rejects a proposal and it's not expected, the test will fail at that point, because either `execute_proposal` or `handle_block_proposal` or something similar will return an error. And if we expect an error, we can assert that, too.

PR #5041, opened 2025-11-29
[testnet] Preload required applications

ma2bd:
## Motivation

Unbreak composed applications on web

Based on @Twey's PR #5036 + extra fixes

## Proposal

* preload application dependencies
* do not crash if `finalize` is called but the loaded app was not used
* preloading an app doesn't actually load it from storage

## Test Plan

CI

### DIFF
--- a/linera-execution/src/lib.rs
+++ b/linera-execution/src/lib.rs
@@ -162,18 +162,6 @@ impl UserContractCode {
 
 pub struct JsVec<T>(pub Vec<T>);
 
-impl<T> From<Vec<T>> for JsVec<T> {
-    fn from(values: Vec<T>) -> Self {
-        Self(values)
-    }
-}
-
-impl<T> From<JsVec<T>> for Vec<T> {
-    fn from(values: JsVec<T>) -> Self {
-        values.0
-    }
-}
-
 #[cfg(web)]
 const _: () = {
     // TODO(#2775): add a vtable pointer into the JsValue rather than assuming the
--- a/linera-execution/src/execution.rs
+++ b/linera-execution/src/execution.rs
@@ -296,11 +296,11 @@ where
         let (codes, descriptions) = actor.service_and_dependencies(application_id).await?;
 
         let thread = web_thread::Thread::new();
-        let service_runtime_task = thread.run_send(JsVec::from(codes), move |codes| async move {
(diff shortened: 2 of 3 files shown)


linera-execution/src/lib.rs
```diff
@@ -202,6 +216,36 @@ const _: () = {
...
+        }
+    }
+
+    impl<T: web_thread::Post> web_thread::Post for JsVec<T> {
+        fn transferables(&self) -> js_sys::Array {
+            let mut array = js_sys::Array::new();
+            for x in &self.0 {
+                array = array.concat(&x.transferables());
+            }
+            array
```
Twey, line 234: Just

```suggestion
            self.0.iter().map(Post::transferables).collect::<Result<js_sys::Array, _>()?.into()
```

should work, no?

linera-execution/src/lib.rs
```diff
@@ -160,6 +160,20 @@ impl UserContractCode {
...
+        Self(values)
+    }
+}
+
+impl<T> From<JsVec<T>> for Vec<T> {
+    fn from(values: JsVec<T>) -> Self {
+        values.0
+    }
+}
+
```
Twey: Unclear on the purpose of these impls when the field is public anyway

linera-execution/src/runtime.rs
```diff
@@ -43,19 +43,23 @@ mod tests;
 
 pub trait WithContext {
     type UserContext;
+    type Code;
```
Twey, line 46: Love this

linera-execution/src/lib.rs
ma2bd: Idk. I took your code from web-thread :)

linera-execution/src/lib.rs
ma2bd: Yes, I suppose we could remove this

linera-execution/src/lib.rs
Twey: That's fair. I learnt about the `collect` impl later 😅

Twey: Oh, it needs to be flattened anyway. Never mind, we wouldn't gain much simplicity from it I suppose.

PR #5044, opened 2025-11-30
[backport] Forward cross chain messages for different recipients actually in parallel

ma2bd:
## Motivation

backport #5005

## Proposal

Forward cross chain messages for different recipients actually in parallel

## Test Plan

CI

## Release Plan

This is the backport


ndr-ds:
Looks like there's other changes as well? Other than #5005

PR #5056, opened 2025-12-01
Node service plugins

afck:
## Motivation

On a private repository Mathieu started working on an alternative node service with a plugin system for external processes that want to interact with apps and commit operations.

## Proposal

Add the plugin system to the node service itself. (All done by Claude so far.)

## Test Plan

CI
(Claude added a test; to extend that we'd have to write an example app that requests actions from a plugin.)

## Release Plan

    - be released in a new SDK.

## Links

### DIFF
--- a/linera-service/tests/local_net_tests.rs
+++ b/linera-service/tests/local_net_tests.rs
@@ -1289,8 +1289,7 @@ async fn test_node_service_with_task_processor() -> Result<()> {
     let mut notifications = Box::pin(node_service.notifications(chain).await?);
 
     // Query the initial task count (should be 0).
-    let app = node_service
-        .make_application::<TaskProcessorAbi>(&chain, &app_id.with_abi::<TaskProcessorAbi>())?;
+    let app = node_service.make_application(&chain, &app_id.with_abi::<TaskProcessorAbi>())?;
     let task_count: u64 = app.query_json("taskCount").await?;
     assert_eq!(task_count, 0);
 
@@ -1318,10 +1317,7 @@ async fn test_node_service_with_task_processor() -> Result<()> {
 
     // Check that the task was processed (task count should be 1).
     let task_count: u64 = app.query_json("taskCount").await?;
-    assert_eq!(
-        task_count, 1,
-        "Expected task count to be 1 after processing"
-    );
+    assert_eq!(task_count, 1);
 
     net.ensure_is_running().await?;
     net.terminate().await?;


linera-service/tests/local_net_tests.rs
```diff
@@ -1232,3 +1232,58 @@ impl EthereumTrackerApp {
...
+    let app_id: ApplicationId = app_id_str.trim().parse()?;
+
+    // Create a simple echo operator script.
+    let tmp_dir = tempfile::tempdir()?;
+    let operator_path = tmp_dir.path().join("echo-operator");
+    {
+        let mut file = std::fs::File::create(&operator_path)?;
+        // Script that reads stdin and writes it to stdout.
+        writeln!(file, "#!/bin/sh")?;
+        writeln!(file, "cat")?;
```
ma2bd, line 1275: this is cool

linera-service/tests/local_net_tests.rs
```diff
@@ -1232,3 +1232,58 @@ impl EthereumTrackerApp {
...
+        .await?;
+
+    node_service.ensure_is_running()?;
+
+    // The counter app doesn't implement the task processor interface (nextActions, etc.),
+    // so the task processor will fail to query it. But we've verified that:
+    // 1. The CLI arguments are accepted.
+    // 2. The node service starts successfully.
+    // 3. The task processor is initialized (even if it can't query the app).
+
```
ma2bd: meh :)

linera-service/src/task_processor.rs
```diff
@@ -0,0 +1,323 @@
+// Copyright (c) Zefchain Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Task processor for executing off-chain operators on behalf of on-chain applications.
+//!
+//! The task processor watches specified applications for requests to execute off-chain tasks,
+//! runs external operator binaries, and submits the results back to the chain.
```
deuszx, line 7: This design adds an integration flow that goes onchain app -> offchain logic – this is a second one alongside the oracle calls that we have.

I think this is wrong as it's creating hard-to-change flows: if we want to add a new trigger, we'll need to update the app. Which is more difficult in "Linera world" as all instances of the app would have to be migrated as well.

Why not use oracle calls? Or events? One uses events to recreate onchain state of the app and that's all you need to trigger the offchain action.

linera-service/src/cli/main.rs
```diff
@@ -1112,20 +1113,42 @@ impl Runnable for Job {
...
+                let chain_id =
+                    default_chain.expect("Service requires a default chain in the wallet");
+
+                let cancellation_token = CancellationToken::new();
+                tokio::spawn(listen_for_shutdown_signals(cancellation_token.clone()));
+
+                // Start the task processor if operator applications are specified.
+                if !operator_application_ids.is_empty() {
+                    let operators = Arc::new(operators.into_iter().collect());
+                    info!("Supported operators: {:?}", operators);
```
deuszx:
```suggestion
                    info!(?operators, "supported operators");
```

linera-service/src/task_processor.rs
afck: I'm not sure how oracle calls could replace this flow: The goal of this is to also support data sources that aren't available (e.g. aren't HTTP-whitelisted) for the validators, and are provided by whoever is running the app backend.

Events could probably be used: that would add slightly more data to he block body that could be considered redundant, since the only consumer of the event is the backend that is running the full chain anyway, and they are not meant to be subscribed to by any other chains.

The current approach with the service query also has the advantage that the service doesn't have to remember anything after a restart: it can just make another query to the app to find out which actions are upcoming. If we were using events, we would have to keep track of which ones were already processed?

Anyway, I don't have a strong opinion; I simply used the same approach as the original PR.

linera-service/src/cli/main.rs
afck: Not sure: INFO-level logs are user-visible. Maybe we shouldn't even use `{:?}`, and `join` the `operators` instead?

linera-service/tests/local_net_tests.rs
afck: I (or rather Claude) extended the test.

linera-service/src/cli/main.rs
afck: Done in https://github.com/linera-io/linera-protocol/pull/5056/commits/6fcbd8d1b7ee0ccdaef95d25603de16381354615.

deuszx: I'm not convinced. We (I think) agreed to some format of the log and we shouldn't change it. Now you create a precedence for the author to decide how the log line should look like and we will end up discussing them again. This is a node service code, not a 3rd party (external) app.

linera-service/src/task_processor.rs
deuszx: > I'm not sure how oracle calls could replace this flow:

I didn't mean _this_ particular flow but in general - oracles are for onchain logic to request validators to do sth.

> Events could probably be used: that would add slightly more data to he block body that could be considered redundant, since the only consumer of the event is the backend that is running the full chain anyway, and they are not meant to be subscribed to by any other chains.

Events are used to publish state changes. Offchain logic can reconstruct the state and act on it. This should suffice here. 

> if we were using events, we would have to keep track of which ones were already processed?

In a sense, yes, by the virtue of reconstructed state. But a "node service" for the app has to follow the chain/app anyway in order to propose blocks on it so it would have the events anyway.

bart-linera: As far as I understand, this does not request the _validators_ to do sth, it requests the _proposer_ to do sth. So it seems that using a different way is quite appropriate that way.

An alternative I can see is to do all this in an external process, separate from the node service, that monitors some chains and then does stuff like proposing blocks / following chains based on what it sees, but the node service is essentially the part of the system that is supposed to react to blocks, so making it extensible like this makes sense to me.

deuszx: I am not advocating against the system of extensions. What I'm simply saying is that we shouldn't be adding new mechanism for onchain apps to signal to offchain processes to do sth. We already have two:
- oracles
- events

afck: But it's not really a _new_ mechanism, is it? It's just a third one in your list, that we already have anyway:
 * service queries

bart-linera:
Closing as superseded by #5072

PR #5058, opened 2025-12-02
Add Postgres DB backend for the indexer.

deuszx:
Backport of #4979 

## Motivation

Turns out our SQLite file for Testnet is quite big already – over 30 GB. We also learned that it might be useful to have a read-only access to the DB remotely, which is not possible with SQLite.

## Proposal

Add support for Postgres backend. Explorer frontend was also updated. So were docker scripts.

Tested locally with all 3 options:
1. `--memory`
2. `--sqlite <path_to_file>`
3. `--postgres <local_instance>`

## Test Plan

Integration tests were added for postgres.

## Release Plan

- Nothing to do

## Links

### DIFF
--- a/docker/Dockerfile.explorer
+++ b/docker/Dockerfile.explorer
@@ -2,6 +2,11 @@
 # Stage 1: Build the frontend
 FROM node:18-alpine AS frontend-builder
 
+# Build argument for API base URL (injected at build time)
+ARG VITE_API_BASE_URL=http://localhost:3002/api
+
+RUN apk add --no-cache python3 make g++
+
 WORKDIR /app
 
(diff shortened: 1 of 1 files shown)


linera-explorer-new/package.json
```diff
@@ -11,20 +11,25 @@
     "start": "concurrently \"npm run server\" \"npm run dev\""
   },
   "dependencies": {
+    "better-sqlite3": "^12.4.1",
+    "cors": "^2.8.5",
+    "express": "^4.18.2",
+    "lucide-react": "^0.263.1",
```
deuszx, line 17: duplicates: remove

PR #5059, opened 2025-12-02
Fix/backup validator volume

eldios:
## Motivation

Fix the Scylla Backup script

## Proposal

There's a couple minor fixes proposed by @conanxyz

## Test Plan

Manually tested.

## Release Plan

will backport to testnet_conway after this is merged.

### DIFF
--- a/linera-chain/src/block.rs
+++ b/linera-chain/src/block.rs
@@ -587,6 +587,38 @@ impl Block {
             && *previous_block_hash == self.header.previous_block_hash
     }
 
+    /// Returns whether the outcomes of the block's execution match the passed values.
+    #[cfg(with_testing)]
+    #[expect(clippy::too_many_arguments)]
+    pub fn outcome_matches(
+        &self,
+        expected_messages: Vec<Vec<OutgoingMessage>>,
+        expected_previous_message_blocks: BTreeMap<ChainId, (CryptoHash, BlockHeight)>,
+        expected_previous_event_blocks: BTreeMap<StreamId, (CryptoHash, BlockHeight)>,
+        expected_oracle_responses: Vec<Vec<OracleResponse>>,
+        expected_events: Vec<Vec<Event>>,
+        expected_blobs: Vec<Vec<Blob>>,
+        expected_operation_results: Vec<OperationResult>,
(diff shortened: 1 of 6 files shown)


ma2bd (approved):

scripts/backup-validator-keys.sh
```diff
@@ -60,8 +60,14 @@ done
 # Backup ScyllaDB data volume info (for reference)
 echo ""
 echo "→ Getting volume information..."
-docker volume inspect linera-scylla-data 2>/dev/null >scylla-volume-info.json ||
-	echo "  ⚠️  ScyllaDB volume not found"
+
```
eldios, line 63: Minor: prefer double brackets for bash scripts:\n```suggestion\nif [[ -n \"$SCYLLA_VOLUME_NAME\" ]]; then\n```

scripts/backup-validator-keys.sh
```diff
@@ -113,7 +119,7 @@ Look for these files in your backup:
 ## Emergency Recovery
```
eldios, line 119: Bug: `$SCYLLA_VOLUME_NAME` won't be expanded here because the heredoc on line 87 uses `<<'EOF'` (quoted), which prevents all variable/command substitution.\n\nThe README will literally contain `$SCYLLA_VOLUME_NAME` instead of the actual volume name. Same issue affects `$(date)` and `$(hostname)` at the bottom.\n\n**Fix**: Change line 87 from `<<'EOF'` to `<<EOF` (unquoted) to enable variable expansion.

eldios (commented):
Found a bug with the heredoc variable expansion. The `<<'EOF'` syntax prevents `$SCYLLA_VOLUME_NAME`, `$(date)`, and `$(hostname)` from being expanded - the README will contain literal strings instead of values.

**Required fix:**
- Line 87: Change `<<'EOF'` to `<<EOF`

**Optional improvement:**
- Line 63: Use `[[ ]]` instead of `[ ]` for bash

eldios:
okay I guess it's easier to just refactor into a new PR since `main` has diverged quite a bit.

superseded by https://github.com/linera-io/linera-protocol/pull/5310

PR #5061, opened 2025-12-03
Remove the memory limit from Ethereum.

MathieuDutSik:
## Motivation

Ethereum has a strict limitation to 24576 bytes which forces the use of multiple contracts for effective running of Ethereum functionalities.

## Proposal

The limit is set to the `maximum_blob_size`.

This requires adding the functionality to the `BaseRuntime`. We only added the needed function and did not add the corresponding Wasm access.
We also did not add full access to `ResourceControlPolicy`. This would expose the internals and lead to an abstraction leakage.

For testnet, this sets the limit to 1M for the contract, which is a radical change coming from Ethereum.

## Test Plan

The CI is the normal way.
Locally, we had proof that these changes do address the limit.

## Release Plan

Whenever Ethereum is deemed acceptable for TestNet, we should broadcast this change.

## Links

None.

### DIFF
--- a/linera-execution/src/lib.rs
+++ b/linera-execution/src/lib.rs
@@ -845,7 +845,7 @@ pub trait BaseRuntime {
     /// Returns true if the corresponding contract uses a zero amount of storage.
     fn has_empty_storage(&mut self, application: ApplicationId) -> Result<bool, ExecutionError>;
 
-    /// Returns the maximum blob size from the ResourceControlPolicy.
+    /// Returns the maximum blob size from the `ResourceControlPolicy`.
     fn maximum_blob_size(&mut self) -> Result<u64, ExecutionError>;
 }
 


linera-execution/src/evm/revm.rs
```diff
@@ -66,6 +66,10 @@ const GET_DEPLOYED_BYTECODE_SELECTOR: &[u8] = &[21, 34, 55, 89];
 /// The json serialization of a trivial vector.
 const JSON_EMPTY_VECTOR: &[u8] = &[91, 93];
 
+/// The maximum size of a deployed EVM contract
+/// Note: The limit of Etherem contract is 24576.
+const MAX_SIZE_EVM_CONTRACT: usize = 10000000;
```
afck:
```suggestion
const MAX_SIZE_EVM_CONTRACT: usize = 10_000_000;
```

(Or `10 << 20`, if we want to make it 10 MiB?)

afck (commented):
Maybe this should be configurable as part of the `ResourceControlPolicy`?

MathieuDutSik:
> Maybe this should be configurable as part of the `ResourceControlPolicy`?

You are right. Maybe we could set it up equal to `maximum_blob_size`. It is 1_000_000 for testnet, so that seems more reasonable indeed.

That being said, it is not totally trivial, since we need to provide access to `maximum_blob_size` to the runtime. We were previously against exposing the `ResourceControlPolicy` because of the abstraction leakage it represented.

So, there are two alternatives:
* Add a function `fn maximum_blob_size(&mut self)` to the `BaseRuntime`.
* Add a function `fn policy_value(&mut self, policy_request: PolicyRequest)` where at first `PolicyRequest` would be an enum with only one variant `MaximumBlobSize`.

afck:
I think it's fine to expose the resource control policy. There's nothing wrong with contracts checking the current fee structure and limits?

I agree, just using `maximum_blob_size` here makes a lot of sense!

MathieuDutSik:
> I think it's fine to expose the resource control policy. There's nothing wrong with contracts checking the current fee structure and limits?

I created a PR some time ago which exposed the `ResourceControlPolicy` and it got yanked.

The reason, and it makes sense to me, is that if we expose the `ResourceControlPolicy` to the contract then we cannot change the `ResourceControlPolicy` in the future. So, exposing API that are not used is not a good idea.

> I agree, just using `maximum_blob_size` here makes a lot of sense!

Thanks.

afck:
> we cannot change the `ResourceControlPolicy` in the future

You mean we can't add fields? That kind of change, after mainnet launch, would require a soft fork anyway, and, yes, it would then also change the SDK and require a new "VM version". But it's not impossible.

MathieuDutSik:
> > we cannot change the `ResourceControlPolicy` in the future
> 
> You mean we can't add fields? That kind of change, after mainnet launch, would require a soft fork anyway, and, yes, it would then also change the SDK and require a new "VM version". But it's not impossible.

Well, yes, nothing is impossible. But the point is that exposing APIs that are not used and create problems for us in the future is not great. At least that was reason enough to refuse my PR in the past, and I have no wish to reproduce the experience.

afck:
Of course, let's definitely not do it in this PR! (I just think it's something to reconsider for the future.) I'd just use the one value for the blob size here.

linera-execution/src/lib.rs
```diff
@@ -844,6 +844,9 @@ pub trait BaseRuntime {
 
     /// Returns true if the corresponding contract uses a zero amount of storage.
     fn has_empty_storage(&mut self, application: ApplicationId) -> Result<bool, ExecutionError>;
+
+    /// Returns the maximum blob size from the ResourceControlPolicy.
```
afck:
```suggestion
    /// Returns the maximum blob size from the `ResourceControlPolicy`.
```

(Or "resource control policy".)

MathieuDutSik: Ok, thanks.

PR #5062, opened 2025-12-03
Introduce clap complete for auto completion in the client.

### LINKED ISSUE #4732: Support bash autocompletion in the linera CLI
https://docs.rs/clap_complete/latest/clap_complete/

MathieuDutSik:
## Motivation

The linera tool has a complex API and efforts have been put to restructure it.
Practitioners need all the help they can get to use the linera client.

Fixes #4732 

## Proposal

The completion has been added only for the client. But it could be added to the proxy and server.

It is used in the following way for bash
```bash
linera completion bash > linera.bash
source linera.bash
```
For zsh it goes
```zsh
linera completion zsh > linera.zsh
source linera.zsh
```

Other supported shells: fish, elvish and powershell.

## Test Plan

CI.
The bash and zsh completions have been tested locally.

## Release Plan

It can be backported to TestNet Conway.

## Links

None.

### DIFF
--- a/linera-service/src/cli/main.rs
+++ b/linera-service/src/cli/main.rs
@@ -2015,7 +2015,12 @@ async fn run(options: &ClientOptions) -> Result<i32, Error> {
 
         ClientCommand::Completion { shell } => {
             let mut cmd = <ClientOptions as clap::CommandFactory>::command();
-            generate(*shell, &mut cmd, "linera", &mut std::io::stdout());
+            generate(
+                *shell,
+                &mut cmd,
+                env!("CARGO_BIN_NAME"),
+                &mut std::io::stdout(),
+            );
             Ok(0)
         }
 


linera-service/src/cli/main.rs
```diff
@@ -2011,6 +2013,12 @@ async fn run(options: &ClientOptions) -> Result<i32, Error> {
             Ok(0)
         }
 
+        ClientCommand::Completion { shell } => {
+            let mut cmd = <ClientOptions as clap::CommandFactory>::command();
+            generate(*shell, &mut cmd, "linera", &mut std::io::stdout());
```
afck: I'd use `env!("CARGO_BIN_NAME")` rather than `"linera"` here. Just one less place where the binary name is duplicated, and makes it easier in the (very unlikely) case that we'd rename the binary.

deuszx:
Nice!

We should also add `linera completion` call to the docs so that users know it's available.

PR #5071, opened 2025-12-04
Move `previous_…_blocks` into the execution state.

### LINKED ISSUE #5068: Move `previous_message_blocks`, `previous_event_blocks` to the execution state.
This makes checkpointing easier. Also, these are conceptually part of the execution state.

afck:
## Motivation

`previous_message_blocks` and `previous_event_blocks` are really part of the execution state conceptually: they get updated only when blocks are executed. Specifically for implementing checkpointing it's inconvenient if the are outside of it, in the chain state view.

## Proposal

Move the fields into the execution state view.

## Test Plan

CI

## Release Plan

## Links

- Closes #5068.

### DIFF
--- a/linera-chain/src/chain.rs
+++ b/linera-chain/src/chain.rs
@@ -271,10 +271,6 @@ where
         BucketQueueView<C, TimestampedBundleInInbox, TIMESTAMPBUNDLE_BUCKET_SIZE>,
     /// Unskippable bundles that have been removed but are still in the queue.
     pub removed_unskippable_bundles: SetView<C, BundleInInbox>,
-    /// The heights of previous blocks that sent messages to the same recipients.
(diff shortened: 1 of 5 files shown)


deuszx (approved):

linera-chain/src/chain.rs
```diff
@@ -970,11 +962,13 @@ where
         let recipients = self.process_outgoing_messages(block).await?;
 
         for recipient in recipients {
-            self.previous_message_blocks
+            self.execution_state
```
bart-linera, line 965: The comment on the method still says "This does not touch the execution state itself" :sweat_smile:

afck: Ah, good point! We probably _should_ update these fields in `execute_block_inner` now instead! Will fix in a follow-up.

PR #5072, opened 2025-12-04
Node service plugins

bart-linera:
## Motivation

On a private repository Mathieu started working on an alternative node service with a plugin system for external processes that want to interact with apps and commit operations.

## Proposal

Add the plugin system to the node service itself. (The task processor part done by Claude, but reviewed.)

This PR is an extension of #5056 

## Test Plan

CI
(Claude added a test; to extend that we'd have to write an example app that requests actions from a plugin.)

## Release Plan

    - be released in a new SDK,

## Links

### DIFF
--- a/linera-service/src/cli/main.rs
+++ b/linera-service/src/cli/main.rs
@@ -37,7 +37,7 @@ use async_trait::async_trait;
 use chrono::Utc;
 use clap_complete::generate;
 use colored::Colorize;
-use futures::{FutureExt as _, StreamExt as _};
+use futures::{lock::Mutex, FutureExt as _, StreamExt as _};
 use linera_base::{
     crypto::{InMemorySigner, Signer},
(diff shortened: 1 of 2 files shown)


CLI.md
```diff
@@ -736,6 +736,9 @@ Run a GraphQL service to explore and extend the chains of the wallet
 
   Default value: `0`
 * `--port <PORT>` — The port on which to run the server
+* `--operator-application-ids <OPERATOR_APPLICATION_IDS>` — Application IDs of operator applications to watch. When specified, a task processor is started alongside the node service
```
ma2bd, line 741: "operator application" is not quite right because some of these applications may not use an "operator" at all. Let's do a round of brainstorming

ma2bd (approved):
@bart-linera Thanks for the PR. This looks awesome

PR #2194, opened 2024-06-26
Generalize client storage from the filesystem

Twey:
## Motivation

The browser doesn't have access to a filesystem.  In order to store and load wallet configuration, we must use a different backend, such as [localStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage).  Currently, though, the `linera-service` crate assumes filesystem access, and stores configuration as JSON on the filesystem.

## Proposal

Generalize the `WalletState` struct with a new trait that represents any data that can be persisted.  Move the previous behaviour into an implementation of that trait specialized to filesystem storage.

## Test Plan

CI.

## Release Plan

None required.

## Links

- [reviewer checklist](https://github.com/linera-io/linera-protocol/blob/main/CONTRIBUTING.md#reviewer-checklist)

PR #5080, opened 2025-12-04
Move the wallet into `linera-core`

Twey:
## Motivation

In #2194 I tried to abstract over the wallet by introducing a trait for values that could be persisted.  However, this still exposed too much information about the wallet: the wallet data structure used by `linera-client` (then `linera-service`) contains additional information that is not useful to the core client.  This made incremental porting hard, as to move the wallet into `linera-core` implied moving all that data (which includes TCP/gRPC-specific address information) into `linera-core` as well, where it does not belong.  Much of this information (signing keys) has been separated out into the `keystore.json` in #3696, but some still remains (the default chain and the genesis config).

Now, in order to support wallet autosigning, we would like to support implementing wallets for more complicated JavaScript structures as well, which may not map neatly to the CLI `wallet.json` file structure.

## Proposal

Add a `Wallet` trait to `linera-core` (as a component of `Environment`) that only exposes the part of the interface related to managing chains, as trait methods.  Currently the trait expects shared-ref `&self` methods as the client will duplicate the wallet all over the place; after future client refactorings we can probably simplify it to support `&mut` methods.

Generalize `linera-service` to maintain the `Environment` abstraction a little better, though there are still places that fix various component types.

## Test Plan

CI should be sufficient.

## Release Plan

    - be released in a new SDK,

## Links

### DIFF
--- a/linera-service/src/cli/main.rs
+++ b/linera-service/src/cli/main.rs
@@ -1512,6 +1512,13 @@ impl Runnable for Job {
                     wallet.set_default_chain(description.id())?;
                 }
 
+                let context = options
+                    .create_client_context(storage, wallet, signer.into_value())
+                    .await?;
+                let chain_client = context.make_chain_client(description.id()).await?;
+                chain_client.synchronize_from_validators().await?;
+                context.update_wallet_from_client(&chain_client).await?;
+
                 // print this only after adding the chain to the wallet
                 println!("{}", description.id());
                 println!("{owner}");
@@ -1545,6 +1552,7 @@ impl Runnable for Job {
                     .synchronize_chain_state_from_committee(committee)
                     .await?;
                 context.update_wallet_from_client(&chain_client).await?;
+                dbg!(context.wallet().chain_ids());
             }
 
             Wallet(WalletCommand::FollowChain { chain_id, sync }) => {
--- a/linera-service/src/wallet.rs
+++ b/linera-service/src/wallet.rs
@@ -131,7 +131,7 @@ impl linera_core::Wallet for Wallet {
         id: ChainId,
         chain: wallet::Chain,
     ) -> Result<Option<wallet::Chain>, Self::Error> {
-        let chain = self.try_insert(id, chain);
+        let chain = self.try_insert(id, chain)?;
         self.save()?;
         Ok(chain)
     }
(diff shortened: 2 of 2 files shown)


linera-service/src/cli/validator.rs
```diff
@@ -3,1006 +3,798 @@
```
ma2bd, line 3: Did the PR need to refactor this file?

Twey: I didn't have to, but I got frustrated making a bunch of changes to it every time I called `ClientContext::new`, so I refactored it to speed up my development.

(It could probably be done in a separate PR — but I'd already changed the APIs it calls and it was 3am :sweat_smile:)

linera-service/src/cli/validator.rs
```diff
@@ -3,1006 +3,798 @@
...
     }
+}
 
-    #[test]
-    fn test_validate_validator_change_empty_address() {
-        let spec = ValidatorChange {
-            account_key: AccountPublicKey::test_key(0),
-            network_address: String::new(),
-            votes: NonZero::new(100).unwrap(),
+impl Sync {
```
ma2bd, line 739: Not sure we needed the structs Sync/Query/etc but ok

ma2bd (approved):
LGTM!

linera-service/src/cli/validator.rs
Twey: Separating them out just lets us skip some boilerplate:  When we had them all as variants we wrote:

```rust
enum Command {
    Sync {
        sync_arg_1: SyncArg1,
        sync_arg_2: SyncArg2,
        sync_arg_3: SyncArg3,
    }
}

fn handle_command(command: Command) {
    match command {
        Sync { sync_arg_1, sync_arg_2, sync_arg_3 } => handle_sync(sync_arg_1, sync_arg_2, sync_arg_3),
    }
}

fn handle_sync(sync_arg_1: SyncArg1, sync_arg_2: SyncArg2, sync_arg_3: SyncArg3) {
    …
}
```

It's not nice to have to repeat that, but more importantly if we want to introduce a new argument we have to change four places (at least — since there's no way to name the bundle it's common for bits of it to end up being sent to other places) instead of the two we'd like to (definition and use). For minimal syntactic overhead keeping our sums and products separate lets us pass that bundle around without deconstructing and reconstructing it needlessly.

Twey: It also means we can keep the definition next to the use if we decide to move the logic around.

PR #5091, opened 2025-12-08
Collect pending messages after acquiring client lock.

afck:
## Motivation

We currently collect the pending messages before acquiring the client lock. That can lead to race conditions where two futures do that at the same time, the first one commits a block with the messages, and then the other one tries to add another block with the same messages, resulting in a "messages out of order" error.

## Proposal

Acquire the lock first.

A side effect of this is that `process_inbox` will try to finalize the pending block even if there are no pending messages. That made one test fail, but the call in that test was not needed anyway.

## Test Plan

CI

## Release Plan

    - be released in a new SDK.

## Links

### DIFF
--- a/linera-core/src/client/chain_client/mod.rs
+++ b/linera-core/src/client/chain_client/mod.rs
@@ -1236,7 +1236,7 @@ impl<Env: Environment> ChainClient<Env> {
 
         let mutex = self.client_mutex();
         let _guard = mutex.lock_owned().await;
-        // TODO: We shouldn't need to call this explicitly.
+        // TODO(#5092): We shouldn't need to call this explicitly.
         match self.process_pending_block_without_prepare().await? {
             ClientOutcome::Committed(Some(certificate)) => {
                 return Ok(ExecuteBlockOutcome::Conflict(certificate))


linera-core/src/client/chain_client/mod.rs
```diff
@@ -1230,31 +1230,13 @@ impl<Env: Environment> ChainClient<Env> {
...
-        transactions: Vec<Transaction>,
-        blobs: Vec<Blob>,
     ) -> Result<ExecuteBlockOutcome, Error> {
         #[cfg(with_metrics)]
         let _latency = super::metrics::EXECUTE_BLOCK_LATENCY.measure_latency();
 
         let mutex = self.client_mutex();
         let _guard = mutex.lock_owned().await;
-        // TOOD: We shouldn't need to call this explicitly.
+        // TODO: We shouldn't need to call this explicitly.
```
ma2bd: While we are at it, do we need the mysterious TODO? If yes, can we create a task?

afck: I was added in https://github.com/linera-io/linera-protocol/commit/6ba7fff8f2c8fe732a7fda8e47c6136fe3911c65
@deuszx: not sure what you meant by it?
I guess we should have a discussion about how to handle pending blocks in general, when to retry them and what to do if the user wants to do something else but there is a pending block?

deuszx: It was meant for us to discuss how to handle the pending block – here we're trying to propose the _previous_ one before we propose the one that was _just_ triggerred. i.e. we might end up proposing two blocks – asking for two signatures – for 1 user request.

afck:
```suggestion
        // TODO(#5092): We shouldn't need to call this explicitly.
```

afck: I created #5092 and linked to it.

PR #5095, opened 2025-12-08
Don't fail w/ an error on destinations mismatch

deuszx:
## Motivation

We have the case of failing exporters on startup due to the mismatch. Config files didn't change so there might be a problem with the DB state.

## Proposal

Don't crash on destination mismatch.

## Test Plan

Manual.

## Release Plan
 -These changes follow the usual release cycle.
- These changes should be backported 

## Links

### DIFF
--- a/linera-service/src/exporter/state.rs
+++ b/linera-service/src/exporter/state.rs
@@ -53,7 +53,7 @@ where
             .await
             .map_err(ExporterError::StateError)?;
 
-        let stored_destinations: Vec<DestinationId> = {
+        let stored_destinations = {
             let pinned = view.destination_states.get().states.pin();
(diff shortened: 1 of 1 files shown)


linera-service/src/exporter/state.rs
```diff
@@ -53,20 +53,22 @@ where
             .await
             .map_err(ExporterError::StateError)?;
 
-        let destinations_len = destinations.len();
+        let stored_destinations: Vec<DestinationId> = {
```
MathieuDutSik: Do we need to indicate the type with `Vec<DestinationId>`?

deuszx: We don't, you're right.

PR #5096, opened 2025-12-09
Use `QueueView` instead of `BucketQueueView`.

afck:
## Motivation

In the benchmarks we are seeing that a lot of time is spent in `BucketQueueView::pre_save`. My theory is that it's serializing the list of buckets, which takes O(N) time depending on the number N _of total entries in the queue_! (_Not_ only on _newly added_ entries!)

```
  | Queue Type      | Time for 100 incremental saves (10,000 existing elements) |
  |-----------------|-----------------------------------------------------------|
  | QueueView       | 33.6 - 34.3 µs                                            |
  | BucketQueueView | 171.5 - 179.3 µs                                          |
```

(And that's with optimally-filled 100-item buckets for the first 10000 items!)

## Proposal

Use `QueueView` instead, so it scales better. We can switch back once we fix the scalability issues of `BucketQueueView`.

## Test Plan

CI

A benchmark was added. We should also try this out with the large-scale benchmark in a real network.

## Release Plan

## Links

### DIFF
--- a/linera-views/benches/queue_view.rs
+++ b/linera-views/benches/queue_view.rs
@@ -212,5 +212,126 @@ fn bench_bucket_queue_view(criterion: &mut Criterion) {
     });
 }
 
-criterion_group!(benches, bench_queue_view, bench_bucket_queue_view);
+/// Benchmark that specifically tests the overhead of saving a queue after adding one element,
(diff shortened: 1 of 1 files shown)


bart-linera (approved):
LGTM

MathieuDutSik (approved):
Thanks.

PR #5099, opened 2025-12-09
Support disabling of contract logging.

deuszx:
## Motivation

We have a `BaseRuntimeApi::log` endpoint that when called from a contract will log to the stdout. This leads to massive amounts of logs on validators.

## Proposal

Support opt-in enabling of the feature.

This is achieved by adding a boolean, config flag to `ExecutionRuntimeConfig` which is then accessed from within the `BaseRuntimeApi`.

## Test Plan

Manual

```
➜  linera-protocol git:(testnet_conway) ✗ ./target/debug/linera --contract-logs wallet follow-chain 6dbf54bd5dfb29560c7911f3e6763bb8c792da46f604f3d2b6b52a91521a9edf --sync
2025-12-09T11:55:17.268236Z  WARN handle_chain_info_query{address="https://linera-testnet.runtime-client-rpc.com:443"}: grpc_client: error=Blobs not found: [BlobId { blob_type: ChainDescription, hash: 6dbf54bd5dfb29560c7911f3e6763bb8c792da46f604f3d2b6b52a91521a9edf }]
2025-12-09T11:55:17.311762Z  WARN handle_chain_info_query{address="https://linera.everstake.one:443"}: grpc_client: error=Blobs not found: [BlobId { blob_type: ChainDescription, hash: 6dbf54bd5dfb29560c7911f3e6763bb8c792da46f604f3d2b6b52a91521a9edf }]
2025-12-09T11:55:18.480625Z  INFO linera_execution::wasm::runtime_api: Fund success transfer_id 1985 fund type Swap on chain 6dbf54bd5dfb29560c7911f3e6763bb8c792da46f604f3d2b6b52a91521a9edf
^C
➜  linera-protocol git:(testnet_conway) ✗ ./target/debug/linera  wallet follow-chain 6dbf54bd5dfb29560c7911f3e6763bb8c792da46f604f3d2b6b52a91521a9edf --sync
2025-12-09T11:55:34.824419Z  WARN handle_chain_info_query{address="https://linera-testnet.runtime-client-rpc.com:443"}: grpc_client: error=Blobs not found: [BlobId { blob_type: ChainDescription, hash: 6dbf54bd5dfb29560c7911f3e6763bb8c792da46f604f3d2b6b52a91521a9edf }]
2025-12-09T11:55:34.862661Z  WARN handle_chain_info_query{address="https://linera.everstake.one:443"}: grpc_client: error=Blobs not found: [BlobId { blob_type: ChainDescription, hash: 6dbf54bd5dfb29560c7911f3e6763bb8c792da46f604f3d2b6b52a91521a9edf }]
2025-12-09T11:55:41.903767Z  WARN handle_chain_info_query{address="https://linera-testnet-validator.contributiondao.com:443"}: grpc_client: error=Blobs not found: [BlobId { blob_type: ChainDescription, hash: 6dbf54bd5dfb29560c7911f3e6763bb8c792da46f604f3d2b6b52a91521a9edf }]
```

## Release Plan
- These changes follow the usual release cycle.
    - be released in a new SDK,

## Links

### DIFF
--- a/linera-execution/src/runtime.rs
+++ b/linera-execution/src/runtime.rs
@@ -1613,7 +1613,7 @@ impl ServiceSyncRuntime {
     ) -> Self {
         // Query the allow_application_logs setting from the execution state.
         let allow_application_logs = execution_state_sender
-            .send_request(|callback| ExecutionRequest::AllowContractLogs { callback })
+            .send_request(|callback| ExecutionRequest::AllowApplicationLogs { callback })
             .ok()
             .and_then(|receiver| receiver.recv_response().ok())
             .unwrap_or(false);
--- a/linera-execution/src/execution_state_actor.rs
+++ b/linera-execution/src/execution_state_actor.rs
@@ -689,7 +689,7 @@ where
                 callback.respond(result);
             }
 
-            AllowContractLogs { callback } => {
+            AllowApplicationLogs { callback } => {
                 let allow = self
                     .state
                     .context()
@@ -1320,7 +1320,7 @@ pub enum ExecutionRequest {
         callback: Sender<(u32, u32)>,
     },
 
-    AllowContractLogs {
+    AllowApplicationLogs {
         #[debug(skip)]
         callback: Sender<bool>,
     },


linera-execution/src/wasm/runtime_api.rs
```diff
@@ -232,14 +232,23 @@ where
...
-    fn log(_caller: &mut Caller, message: String, level: log::Level) {
-        match level {
-            log::Level::Trace => tracing::trace!("{message}"),
-            log::Level::Debug => tracing::debug!("{message}"),
-            log::Level::Info => tracing::info!("{message}"),
-            log::Level::Warn => tracing::warn!("{message}"),
-            log::Level::Error => tracing::error!("{message}"),
+    fn log(caller: &mut Caller, message: String, level: log::Level) -> Result<(), RuntimeError> {
+        let allowed = caller
+            .user_data_mut()
```
afck, line 237: Do we need `_mut`?

linera-execution/src/execution.rs
```diff
@@ -269,7 +269,8 @@ where
                 application_id,
                 bytes,
             } => {
-                let ExecutionRuntimeConfig {} = self.context().extra().execution_runtime_config();
+                let ExecutionRuntimeConfig { .. } =
+                    self.context().extra().execution_runtime_config();
```
afck: (I don't even understand what this line is for.)

linera-service/src/cli/main.rs
```diff
@@ -1723,6 +1723,10 @@ struct ClientOptions {
     #[arg(long)]
     wasm_runtime: Option<WasmRuntime>,
 
+    /// Output log messages from contract execution.
+    #[arg(long)]
+    contract_logs: bool,
```
afck: I wonder if this should be `true` for clients by default (and `false` for validators, of course).

linera-execution/src/runtime.rs
```diff
@@ -944,6 +944,15 @@ where
...
         Ok(self.inner().resource_controller.policy().maximum_blob_size)
     }
+
+    fn allow_contract_logs(&mut self) -> Result<bool, ExecutionError> {
+        let this = self.inner();
+        let allowed = this
+            .execution_state_sender
+            .send_request(|callback| ExecutionRequest::ContractLogsAllowed { callback })?
+            .recv_response()?;
+        Ok(allowed)
```
afck: Does this really have to be a request to the execution state actor? We could just initialize the runtime with that flag, could we?

deuszx: What is "the runtime" really? Our abstractions are pretty weird around this – they're mostly empty (phantom data) types with some method on the instance that use the `BaseRuntimeApi` (and other) traits.

linera-execution/src/execution.rs
deuszx: Me neither.

linera-execution/src/wasm/runtime_api.rs
deuszx: Maybe not, I just followed the pattern in this file. Other reads also use `_mut`.

linera-execution/src/runtime.rs
afck: Probably `SyncRuntimeInternal` would be a candidate to put the field in?

deuszx: It didn't seem like a good place for it for a couple of reason:
1. It's about "current" state of the execution state – height, chain, applications on the stack, fuel used, etc.
2. it's not available from `BaseRuntimeApi`

I think the current place makes more sense. I don't think our abstractions around the access to the Contract API or storage are the best but that's a different issue.

deuszx: Although, this approach is pretty bad performance-wise: for every `log` call we have to make an async call to decide whether it's available. I'll try to come up with something better.

linera-execution/src/execution.rs
ma2bd: it's tech debt that wasn't removed

linera-execution/src/wasm/runtime_api.rs
deuszx: Somehow the non-`_mut` version is not even visible to the compiler.

linera-execution/src/execution.rs
deuszx: Removed. This and the other place.

linera-execution/src/runtime.rs
deuszx: See [Put the flag in SyncRuntimeInternal](https://github.com/linera-io/linera-protocol/pull/5099/commits/38a041d07af4ec7851b4dd5b5ad5d867b56a6735)

linera-execution/src/runtime.rs
```diff
@@ -1612,6 +1622,7 @@ impl ServiceSyncRuntime {
                 None,
                 ResourceController::default(),
                 (),
+                false, // Services don't output contract logs
```
ma2bd: Services should be able to output logs. Or did you mean that the option doesn't apply?

linera-service/src/cli/main.rs
ma2bd: Maybe let's take a minute and discuss offline the behaviors that we want contact/service/client/validators/proxy.

linera-execution/src/runtime.rs
deuszx: Changed the approach. Please re-review.

linera-execution/src/runtime.rs
deuszx: Now services are controlled the same way the contracts.

linera-execution/src/runtime.rs
```diff
@@ -1601,6 +1611,13 @@ impl ServiceSyncRuntime {
         context: QueryContext,
         deadline: Option<Instant>,
     ) -> Self {
+        // Query the allow_contract_logs setting from the execution state.
+        let allow_contract_logs = execution_state_sender
+            .send_request(|callback| ExecutionRequest::AllowContractLogs { callback })
+            .ok()
+            .and_then(|receiver| receiver.recv_response().ok())
+            .unwrap_or(false);
```
afck, line 1619: (Can't it be passed into `new_with_deadline`, too? When I suggested a field in the runtime I was hoping to get rid of the execution state actor request altogether. If we can't get rid of that, maybe it's cleaner to always use it and not have the boolean field at all?)

CLI.md
```diff
@@ -241,6 +241,7 @@ Client implementation and command-line tool for the Linera blockchain
 
   Default value: `1`
 * `--wasm-runtime <WASM_RUNTIME>` — The WebAssembly runtime to use
+* `--contract-logs` — Output log messages from contract execution
```
afck: Should it be `--application-logs` or `--with-application-logs`, since it affects both contracts and services?

linera-execution/src/runtime.rs
deuszx: No, it's not possible. There's a one call-site where service runtime actor is spawned NOT from the context of the execution runtime (a long-living service/endpoint of some kind) and there we don't have access to it.

deuszx: > If we can't get rid of that, maybe it's cleaner to always use it and not have the boolean field at all?)

This approach will be slightly slower as to check if we should log or not we have to go through all the piping of message sending and receiving.

linera-execution/src/wasm/runtime_api.rs
deuszx: We need mutable version fo `DerefMut` used somewhere down the line.

CLI.md
deuszx: Done in [Rename flag to (with)application_logs](https://github.com/linera-io/linera-protocol/pull/5099/commits/935d8474bb872e96f7c49c0a78b4186b9b5a3ed9).

linera-execution/src/runtime.rs
```diff
@@ -1611,8 +1611,8 @@ impl ServiceSyncRuntime {
         context: QueryContext,
         deadline: Option<Instant>,
     ) -> Self {
-        // Query the allow_contract_logs setting from the execution state.
-        let allow_contract_logs = execution_state_sender
+        // Query the allow_application_logs setting from the execution state.
+        let allow_application_logs = execution_state_sender
             .send_request(|callback| ExecutionRequest::AllowContractLogs { callback })
```
afck: The request should probably also be `AllowApplicationLogs` now.

deuszx: Ah, done in [One more rename](https://github.com/linera-io/linera-protocol/pull/5099/commits/a2da76d312a3f9c092f2fde86412a593a7103524).

PR #5101, opened 2025-12-09
Follow-only chains

### LINKED ISSUE #5098: "Follow-only" chains
If we follow a chain that we don't own or don't intend to propose new blocks for, we are not interested in incoming messages and events, so it's unnecessary overhead to download sender and publisher blocks.

We should add a mode in which we only follow the chain itself.

afck:
## Motivation

All chains in the wallet are currently fully synchronized, including their sender and publisher chains. But in use cases where the user only wants to _read_ from that chain and not _write_ to it (i.e. never make a block proposal, e.g. because they don't even own it), sender and publisher chains are not needed, and neither are the chain manager values.

## Proposal

Add a new listening mode and flag that marks chains as "follow-only". For those chains, only their own confirmed blocks are synchronized.

The `follow-chain` and `forget-keys` commands automatically set that flag, and `assign` unsets it.

## Test Plan

A test was added.

## Release Plan

- These changes could possibly be backported to `testnet_conway`, then
    - be released in a new SDK.

## Links

- Closes #5098.

### DIFF
--- a/linera-client/src/unit_tests/chain_listener.rs
+++ b/linera-client/src/unit_tests/chain_listener.rs
@@ -318,6 +318,41 @@ async fn test_chain_listener_follow_only() -> anyhow::Result<()> {
         "Follow-only chain A should not have had its inbox processed"
     );
 
+    // Verify that the listener's wallet still shows chain A at height 0.
+    let wallet_chain_a = context.lock().await.wallet().get(chain_a_id).unwrap();
+    assert_eq!(
+        wallet_chain_a.next_block_height,
+        BlockHeight::ZERO,
+        "Wallet should show chain A at height 0"
+    );
+
+    // Now have the original chain_a client process its inbox, creating a block.
+    chain_a.process_inbox().await?;
+
+    // Wait for the chain listener to see the NewBlock notification and update its wallet.
+    // This verifies that follow-only mode DOES process NewBlock notifications.
+    for i in 0.. {
+        tokio::task::yield_now().await;
+
+        let wallet_chain_a = context.lock().await.wallet().get(chain_a_id).unwrap();
+        if wallet_chain_a.next_block_height >= BlockHeight::from(1) {
+            break;
+        }
+        if i >= 50 {
+            panic!(
+                "Wallet not updated after chain A created a block. Expected height >= 1, got {}",
+                wallet_chain_a.next_block_height
+            );
+        }
+    }
+
+    // Verify the wallet was updated and follow_only is preserved.
+    let wallet_chain_a = context.lock().await.wallet().get(chain_a_id).unwrap();
+    assert!(
+        wallet_chain_a.follow_only,
+        "follow_only flag should be preserved in wallet"
+    );
+
     cancellation_token.cancel();
     handle.await;
 


linera-core/src/client/mod.rs
```diff
@@ -281,8 +281,10 @@ impl ListeningMode {
         match (reason, self) {
             // FullChain processes everything.
             (_, ListeningMode::FullChain) => true,
-            // FollowChain only processes new blocks on the chain itself.
+            // FollowChain processes new blocks on the chain itself, including blocks that
+            // produced events.
             (Reason::NewBlock { .. }, ListeningMode::FollowChain) => true,
+            (Reason::NewEvents { .. }, ListeningMode::FollowChain) => true,
```
deuszx, line 287: We should get `NewBlock` anyway, right? I.e. there can't be a `NewEvents` notification w/o `NewBlock`.

afck: That's true, but we still need to _handle_ `NewEvents` as well: This chain might be not only followed explicitly by us, but also produce events for another chain. If that other chain is `FullChain`, we need to handle the event.

linera-core/src/client/chain_client/mod.rs
```diff
@@ -302,6 +302,11 @@ impl<Env: Environment> ChainClient<Env> {
         }
     }
 
+    /// Returns whether this chain is in follow-only mode.
+    fn follow_only(&self) -> bool {
```
deuszx: Rename to `is_follow_only`.

linera-core/src/client/chain_client/mod.rs
```diff
@@ -1624,8 +1629,14 @@ impl<Env: Environment> ChainClient<Env> {
...
     /// To create a block that actually executes the messages in the inbox,
     /// `process_inbox` must be called separately.
+    ///
+    /// If the chain is in follow-only mode, this only downloads blocks for this chain without
+    /// fetching manager values or sender/publisher chains.
     #[instrument(level = "trace")]
     pub async fn synchronize_from_validators(&self) -> Result<Box<ChainInfo>, Error> {
+        if self.follow_only() {
+            return self.client.synchronize_chain_state(self.chain_id).await;
+        }
```
deuszx: I don't like this addition. What we're adding is an option that _does less_ so if anything we should be removing code, not adding it. This adds complexity and we already have too much of it.

linera-core/src/client/chain_client/state.rs
```diff
@@ -38,9 +43,20 @@ impl State {
         State {
             pending_proposal: self.pending_proposal.clone(),
             client_mutex: Arc::clone(&self.client_mutex),
+            follow_only: self.follow_only,
         }
     }
 
+    /// Returns whether this chain is in follow-only mode.
+    pub fn follow_only(&self) -> bool {
```
deuszx: rename to `is_follow_only`.

linera-core/src/client/mod.rs
```diff
@@ -1243,29 +1301,46 @@ impl<Env: Environment> Client<Env> {
...
             .download_certificates_from(remote_node, chain_id, remote_info.next_block_height)
-            .await?
-        {
-            local_info = new_info;
-        };
+            .await?;
+
+        if !with_manager_values {
+            return Ok(());
+        }
```
deuszx, line 1376: For the case when the chain is in the follow-only mode, we return early here. But we didn't do any meaningful work before it – do we even eed to handle those two cases? What does the code above really do if the client is in the follow-only mode?

linera-core/src/wallet/mod.rs
```diff
@@ -21,6 +21,11 @@ pub struct Chain {
     pub timestamp: Timestamp,
     pub pending_proposal: Option<PendingProposal>,
     pub epoch: Option<Epoch>,
+    /// If true, we only follow this chain's blocks without downloading sender chain blocks
+    /// or participating in consensus rounds. Use this for chains we're interested in observing
+    /// but don't intend to propose blocks for.
+    #[serde(default)]
+    pub follow_only: bool,
```
deuszx, line 28: Don't we need to also store the `EventsOnly` variant?

linera-core/src/unit_tests/client_tests.rs
```diff
@@ -3191,3 +3192,70 @@ where
...
+        .chain_info(sender.chain_id())
+        .await?;
+    assert_eq!(
+        sender_info.next_block_height,
+        BlockHeight::ZERO,
+        "Follower should not have downloaded the sender's blocks"
+    );
+
+    Ok(())
+}
```
deuszx, line 3261: This doesn't test the `ChainListener`, right? There were changes to its code as well that we should test.

linera-core/src/wallet/mod.rs
afck: Those chains aren't in the wallet: they are only listened to because a non-follow-only chain in the wallet is subscribed to them.

linera-core/src/client/chain_client/mod.rs
afck: What do you suggest instead?

`synchronize_from_validators` is called in the `sync` command, and I think the right thing to do for it is: if it's a full chain, synchronize everything, otherwise only download the confirmed blocks for the chain.

linera-core/src/client/mod.rs
afck: It downloaded and processed all confirmed blocks. (We really need a better name for `download_certificates_from`.)

deuszx: OK, I forgot – I thought it's only _downloading_ (like the func name suggests).

linera-core/src/unit_tests/client_tests.rs
afck: Added a test in https://github.com/linera-io/linera-protocol/pull/5101/commits/01b7e2f91c1bcc03d2f36de90aace62990c0892c and https://github.com/linera-io/linera-protocol/pull/5101/commits/856a71253cc6def45260e3340265c3ef07e35821.

linera-client/src/unit_tests/chain_listener.rs
```diff
@@ -187,6 +189,141 @@ async fn test_chain_listener() -> anyhow::Result<()> {
...
+    }
+
+    // Now verify that chain A's inbox was NOT processed (follow-only ignores NewIncomingBundle).
+    chain_a.synchronize_from_validators().await?;
+    let chain_a_info = chain_a.chain_info().await?;
+    assert_eq!(
+        chain_a_info.next_block_height,
+        BlockHeight::ZERO,
+        "Follow-only chain A should not have had its inbox processed"
+    );
```
deuszx, line 319: Can we extend the test to verify that `ChainListener` does follow blocks though?

afck: Done in https://github.com/linera-io/linera-protocol/pull/5101/commits/c89acd6ff489f587793f5608e147580c36ea8e5e.

deuszx: Who processes the notification if the chain listener's background sync is set to `false`? Doesn't it work b/c everything in the test uses the same storage?

afck: Background sync is different from listening for notifications. The chain listener always listens.

Background sync means: On startup, we begin synchronizing all certificates from sender and publisher chains, because we might have missed notifications while we were offline.

PR #5102, opened 2025-12-09
Disable fallback mode.

afck:
## Motivation

Keeping track of all unskippable messages turns out to be expensive, and we are not using fallback mode on the testnet anyway.

## Proposal

Stop using `unskippable_bundles` and `remove_unskippable_bundles`. We can't remove these fields so storage remains compatible, but we can stop reading to and writing from them.

## Test Plan

CI

The fallback test was removed.

## Release Plan

## Links


linera-chain/src/chain.rs
```diff
@@ -597,9 +588,7 @@ where
         metrics::NUM_INBOXES
             .with_label_values(&[])
             .observe(self.inboxes.count().await? as f64);
-        let entry = BundleInInbox::new(*origin, &bundle);
-        let skippable = bundle.is_skippable();
-        let newly_added = inbox
+        let _newly_added = inbox
```
ma2bd: nit: remove `let _newly_added =` ?

linera-core/src/chain_worker/state.rs
```diff
@@ -1251,21 +1252,16 @@ where
...
-                let height = chain.tip_state.get().next_block_height;
-                let key_pair = self.config.key_pair();
-                if chain
-                    .manager
-                    .vote_fallback(chain_id, height, *epoch, key_pair)
-                {
-                    self.save().await?;
-                }
+        let epoch = chain.execution_state.system.epoch.get();
+        if chain.ownership().timeout_config.fallback_duration == TimeDelta::ZERO {
```
ma2bd: I see. In this case, this means that fallback owners are always authorized to request a fallback round? (as opposed to conditionally)

afck: Yes… but we don't use it on the testnet anyway, so I'll just disable it entirely.

PR #5105, opened 2025-12-10
Enable fallback mode only based on epochs; don't track unskippable messages.

### LINKED ISSUE #5103: Make fallback mode use only epochs, not messages.
Keeping track of all unskippable bundles adds a lot of overhead. Let's instead tie fallback mode only to epochs: If a chain is in epoch e, and epoch e+1 is already older than the configured `fallback_delay`, the validators vote to switch to fallback mode.

afck:
## Motivation

Keeping track of the `unskippable_bundles` collection seems to significantly increase the overhead for cross-chain messaging, as observed in the benchmarks.

## Proposal

Remove it.
Enable fallback mode based on epochs only, not based on tracked or protected messages.

## Test Plan

The fallback test was updated.

## Release Plan

## Links

- Closes #5103.

### DIFF
--- a/linera-chain/src/chain.rs
+++ b/linera-chain/src/chain.rs
@@ -26,10 +26,8 @@ use linera_views::{
     context::Context,
     log_view::LogView,
     map_view::MapView,
-    queue_view::QueueView,
     reentrant_collection_view::{ReadGuardedView, ReentrantCollectionView},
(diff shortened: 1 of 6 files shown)


ma2bd (approved):
LGTM. Thanks!

PR #5107, opened 2025-12-10
Add functionality to the QueueView / SetView.

MathieuDutSik:
## Motivation

The `QueueView` has `front` / `delete_front` while we may want to have a `pop`.
The `SetView` also has the `contains` / `remove`, and we may want to do both at once.

## Proposal

Implement a `.pop()` for `QueueView`. Implement a `remove_if_present` for `SetView`.

Unfortunately, the `front` is not systematically followed by a `delete_front`. So, we cannot use it in the code.
For the `SetView` change, another PR indicates a different direction: Altogether removal of the unskippable entries.

## Test Plan

CI. Unit tests have been added.

## Release Plan

Not that important. Could be 

## Links

None.

### DIFF
--- a/linera-views/src/views/queue_view.rs
+++ b/linera-views/src/views/queue_view.rs
@@ -276,6 +276,27 @@ where
         }
     }
 
+    /// Pops the front value, if any.
+    /// ```rust
+    /// # tokio_test::block_on(async {
+    /// # use linera_views::context::MemoryContext;
+    /// # use linera_views::queue_view::QueueView;
+    /// # use linera_views::views::View;
+    /// # let context = MemoryContext::new_for_testing(());
+    /// let mut queue = QueueView::load(context).await.unwrap();
+    /// queue.push_back(42);
+    /// queue.push_back(34);
+    /// assert_eq!(queue.pop().await.unwrap(), Some(42));
+    /// assert_eq!(queue.pop().await.unwrap(), Some(34));
+    /// assert_eq!(queue.pop().await.unwrap(), None);
+    /// # })
+    /// ```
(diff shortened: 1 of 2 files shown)


linera-views/src/views/queue_view.rs
```diff
@@ -276,6 +276,27 @@ where
...
+    /// queue.push_back(34);
+    /// assert_eq!(queue.pop().await.unwrap(), Some(42));
+    /// assert_eq!(queue.pop().await.unwrap(), Some(34));
+    /// assert_eq!(queue.pop().await.unwrap(), None);
+    /// # })
+    /// ```
+    pub async fn pop(&mut self) -> Result<Option<T>, ViewError> {
+        let value = self.front().await?;
+        self.delete_front();
+        Ok(value)
```
ma2bd, line 297: Do we need this API?

MathieuDutSik: It turns out no. That being said it is the standard usage of queues.

bart-linera:
Instead of adding `remove_if_present`, wouldn't it make more sense for `remove` to return a boolean, indicating whether the value was present? That's how `BTreeSet` and `HashSet` work, at least.

MathieuDutSik:
> Instead of adding `remove_if_present`, wouldn't it make more sense for `remove` to return a boolean, indicating whether the value was present? That's how `BTreeSet` and `HashSet` work, at least.

We could of course do that. But `fn remove` is not async and does not even return an error. So, it is quite a useful idea to still have both forms even if their name could change from the ones we have now.

ma2bd:
converting stale PR to draft

MathieuDutSik:
This PR does not appear to be useful. So, better to terminate it.

Issue #5023, opened 2025-11-26
Web clients on Windows sometimes get `InvalidTimestamp` errors


Twey:
A.K.A.

```
The block timestamp is in the future.
```

This can come from the client's worker (where it can be safely ignored) or the validator's worker (where it can't).

We think this results from a failure in timezone conversion, but we haven't narrowed down the precise requirements for reproduction.  [`web-time`](https://docs.rs/web-time/latest/web_time/) is a likely candidate.

Twey:
Maybe:
- `web-time` reads the wrong time from the system clock on Windows (not taking timezone into account)
- the types from `web-time` are not serialized correctly when sending to the validator

PR #5108, opened 2025-12-10
[testnet] Retry and warn if validators return InvalidTimestamp. (#5090)

afck:
Backport of #5090.

## Motivation

Validators return `InvalidTimestamp` errors if a proposed block's timestamp is too far in the future. (For a short grace period, 500 ms by default, they wait.)

Some users are getting this error, possibly because their local clock is not in sync with the validators'.

## Proposal

Whenever we get `InvalidTimestamp` from a validator, we wait and retry.

If the local clock is the problem and not the validators', this can cause long delays, of course. So we print a warning log message as soon as a validity threshold has returned this error.

To backport this to `testnet_conway` in a backwards-compatible way, instead of adding a variant to `NodeError` (which would cause deserialization errors in clients that weren't updated), I changed the error string and parse it. That way, even if the validator is updated but the client isn't, the error is still legible.

An alternative would be to add the variant, which fails to parse in clients that are not updated. (I.e. the error will just say "failed to unmarshal error".) But the code would be cleaner, and the same as on `main`.

## Test Plan

A test was added.

## Release Plan

- This should be released in a new SDK _first_, then the validators should be updated.

## Links

- Related to https://github.com/linera-io/linera-protocol/issues/5023.

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -576,7 +576,7 @@ impl<Env: Environment> Client<Env> {
                 chain_id = %proposal.content.block.chain_id,
                 %block_timestamp,
                 %local_time,
-                "Block timestamp is in the future; waiting for validators",
+                "Block timestamp is in the future; waiting until it can be proposed",
             );
         }
 


linera-core/src/client/mod.rs
```diff
@@ -567,13 +567,60 @@ impl<Env: Environment> Client<Env> {
...
+
+        // Check if the block timestamp is in the future and log INFO.
+        let block_timestamp = proposal.content.block.timestamp;
+        let local_time = self.local_node.storage_client().clock().current_time();
+        if block_timestamp > local_time {
+            info!(
+                chain_id = %proposal.content.block.chain_id,
+                %block_timestamp,
+                %local_time,
+                "Block timestamp is in the future; waiting for validators",
```
ma2bd: what do you mean by "waiting for validators"?

afck: Good point, it's not the validators' fault that the timestamp is in the future. Will fix…

linera-core/src/node.rs
```diff
@@ -318,6 +320,40 @@ pub enum NodeError {
...
+    /// The validator's local time when it rejected the block.
+    pub validator_local_time: Timestamp,
+}
+
+impl NodeError {
+    /// If this error is an `InvalidTimestamp` error (wrapped in `WorkerError`), parses and
+    /// returns the timestamps. Returns `None` for other error types.
+    ///
+    /// The error string format is expected to contain `[us:{block_timestamp}:{local_time}]`
+    /// where both values are microseconds since epoch.
```
ma2bd: We're parsing strings on the testnet branch only, I suppose?

linera-core/src/client/mod.rs
```diff
@@ -567,13 +567,60 @@ impl<Env: Environment> Client<Env> {
...
+                    min_skew = min_skew.min(clock_skew);
+                    max_skew = max_skew.max(clock_skew);
+                    if skew_weight >= validity_threshold {
+                        warn!(
+                            skew_weight,
+                            validity_threshold,
+                            min_skew_ms = min_skew.as_micros() / 1000,
+                            max_skew_ms = max_skew.as_micros() / 1000,
+                            "A validity threshold of validators reported clock skew; \
+                             consider checking your system clock",
```
ma2bd, line 610: Note that realistically, no user is going to read this.

linera-core/src/updater.rs
```diff
@@ -452,17 +458,43 @@ where
...
+                        %chain_id,
+                        block_timestamp = %invalid_ts.block_timestamp,
+                        ?clock_skew,
+                        "validator's clock is behind; waiting and retrying",
+                    );
+                    // Report the clock skew before sleeping so the caller can aggregate.
+                    let _ = clock_skew_sender.send((self.remote_node.public_key, clock_skew));
+                    storage
+                        .clock()
+                        .sleep_until(invalid_ts.block_timestamp.saturating_add(clock_skew))
```
ma2bd, line 495: How long can we wait in the worst case scenario?

afck: There's no limit, _but_: We cancel this task if we get a quorum of signature (or a validity threshold of errors) from other validators.

linera-core/src/client/mod.rs
afck: What else can we do? :shrug: 
(Ah, you mean keep track of the skew and take it into account for future block proposals… but that's beyond the scope of this PR.)

PR #5113, opened 2025-12-10
Thread pool

### LINKED ISSUE #3651: OOM on Web due to ephemeral threads

Twey:
## Motivation

https://github.com/linera-io/linera-protocol/issues/3651

## Proposal

Introduce a thread pool on top of `web-thread` and reüse a fixed pool of threads (here hardcoded to 20, without any particular rationale; we should make it configurable as future work).

## Test Plan

CI.  The pool logic is shared between the native and Web cases, so should be adequately tested by our usual end-to-end tests.

## Release Plan

    - be released in a new SDK,

## Links

- fixes #3651

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11611,7 +11611,8 @@ dependencies = [
 [[package]]
 name = "web-thread"
 version = "0.2.2"
-source = "git+https://github.com/linera-io/web-thread?branch=pool#d733fd6b9562ea43e32902d35f73dadd7c9c57f3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "060cfc12637ebaf48b344331cea64fe6d13fed9a1381fb6cce10c3c7495778b4"
 dependencies = [
  "futures",
  "pin-project-lite",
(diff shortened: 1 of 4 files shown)


linera-core/src/environment/mod.rs
```diff
@@ -20,9 +22,7 @@ pub trait Environment: AutoTraits {
     type ValidatorNode: crate::node::ValidatorNode + AutoTraits + Clone;
     // TODO(#5064): we shouldn't hard-code `Send` + `Sync` here
```
ma2bd, line 23: Seems like you solved this TODO?

Twey: I did but saw this too late :sweat_smile: Removed in #5116.

linera-storage/src/db_storage.rs
```diff
@@ -1081,6 +1087,9 @@ impl<Database, C> DbStorage<Database, C> {
         Self {
             database: Arc::new(database),
             clock,
+            // The `Arc` here is required on native but useless on the Web.
+            #[cfg_attr(web, expect(clippy::arc_with_non_send_sync))]
+            thread_pool: Arc::new(linera_execution::ThreadPool::new(20)),
```
ndr-ds, line 1092: Just curious: what's the rationale behind the 20 here?

PR #5114, opened 2025-12-10
[testnet] Follow-only chains (#5101)

afck:
Backport of #5101, with backwards-compatible wallet.

## Motivation

All chains in the wallet are currently fully synchronized, including their sender and publisher chains. But in use cases where the user only wants to _read_ from that chain and not _write_ to it (i.e. never make a block proposal, e.g. because they don't even own it), sender and publisher chains are not needed, and neither are the chain manager values.

## Proposal

Add a new listening mode and flag that marks chains as "follow-only". For those chains, only their own confirmed blocks are synchronized.

The `follow-chain` and `forget-keys` commands automatically set that flag, and `assign` unsets it.

In this backport, we also added code to read the old wallet format and initialize the flag based on whether a chain has a keypair assigned.

## Test Plan

A test was added.

## Release Plan

- Should be released in a new SDK.

## Links

- PR to main: #5101

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -287,8 +287,8 @@ impl<Env: Environment> Client<Env> {
 
     /// Returns whether the given chain is in follow-only mode (no owner key in the wallet).
     ///
-    /// If the chain is not in the wallet, returns `true` since we cannot participate in
-    /// consensus without an owner key.
+    /// If the chain is not in the wallet, returns `true` since we don't have an owner key
+    /// for it.
     async fn is_chain_follow_only(&self, chain_id: ChainId) -> bool {
         match self.wallet().get(chain_id).await {
             Ok(Some(chain)) => chain.owner.is_none(),
--- a/linera-core/src/unit_tests/client_tests.rs
+++ b/linera-core/src/unit_tests/client_tests.rs
@@ -48,6 +48,7 @@ use crate::{
         BlanketMessagePolicy, ChainClient, ChainClientError, ChainClientOptions, ClientOutcome,
         ListeningMode, MessageAction, MessagePolicy,
     },
+    environment::wallet::Chain,
     local_node::LocalNodeError,
     node::{
         NodeError::{self, ClientIoError},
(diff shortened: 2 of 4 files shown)


linera-core/src/wallet/mod.rs
```diff
@@ -24,10 +25,44 @@ pub struct Chain {
     /// If true, we only follow this chain's blocks without downloading sender chain blocks
     /// or participating in consensus rounds. Use this for chains we're interested in observing
     /// but don't intend to propose blocks for.
-    #[serde(default)]
     pub follow_only: bool,
```
Twey: Can we not just set this to `Option<bool>` and use an accessor that encodes the defaulting logic?

afck: Done in https://github.com/linera-io/linera-protocol/pull/5114/commits/67c98c196ae7d2bcd1768844202401eb88a5a9e6. Now I'm wondering if we should just leave it `None` in most cases.

afck:
Re-running that test. I can't reproduce the failure: `test_open_chain_node_service::remote_net_grpc` passed for me, against the testnet.

linera-core/src/client/mod.rs
```diff
@@ -146,15 +146,54 @@ mod metrics {
...
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        use std::cmp::Ordering;
+        match (self, other) {
+            (ListeningMode::FullChain, ListeningMode::FullChain) => Some(Ordering::Equal),
+            (ListeningMode::FullChain, _) => Some(Ordering::Greater),
+            (_, ListeningMode::FullChain) => Some(Ordering::Less),
+            (ListeningMode::FollowChain, ListeningMode::FollowChain) => Some(Ordering::Equal),
+        }
+    }
 }
```
ma2bd, line 173: @afck Do we need to write this?

afck: Right, I guess on the testnet branch we could as well derive it, and make sure the variants are in the right order. But it's maybe more explicit this way, and on main we have to write it anyway because of the event sets?

ma2bd: ok I was just curious

PR #5116, opened 2025-12-11
Autosigning on the Web

Twey:
## Motivation

Signing blocks on the Web can get tiresome when using wallets like MetaMask or Dynamic, since we aren't integrated with their autosigning features, and working with chains with hundreds of incoming messages typically means accepting hundreds of signatures.

## Proposal

First connect to the ‘real’ wallet, and use it to sign a block that adds a new local keypair as an additional owner of the chain.  Then, replace the original owner of the chain with this new temporary owner in the wallet so that future blocks will be signed with that key instead.  We currently keep the wallet in memory on the Web so needn't worry about overriding the real owner.

This is a slightly clumsy approach (and the application must be careful to send any assets back to the original owner!) but should work as a first step.

Implement a little supporting API to the Web client, then demo an autosigning flow in the MetaMask counter demo.

## Test Plan

Tested by hand.

## Release Plan

    - be released in a new SDK,

## Links

### DIFF
--- a/examples/counter/metamask/index.html
+++ b/examples/counter/metamask/index.html
@@ -64,10 +64,14 @@
           const chain = await faucet.claimChain(wallet, owner);
           document.getElementById('owner').innerText = owner;
           document.getElementById('chain-id').innerText = chain;
+
+          // add a new local wallet to the chain that can autosign blocks
+          // without prompting every time
           const autosigner = linera.signer.PrivateKey.createRandom();
           const client = await new linera.Client(wallet, new linera.signer.Composite(autosigner, signer));
           await client.addOwner(autosigner.address());
           wallet.setOwner(chain, autosigner.address());
+
           const counter = await client.application(import.meta.env.LINERA_APPLICATION_ID);
           const logs = document.getElementById('logs');
           const incrementButton = document.getElementById('increment-btn');


web/@linera/client/src/signer/mod.rs
```diff
@@ -24,67 +25,64 @@ pub enum JsSignerError {
...
 
-impl std::error::Error for JsSignerError {}
+impl std::error::Error for Error {}
 
 #[wasm_bindgen]
 extern "C" {
     // We refer to the interface defined above.
     #[wasm_bindgen(typescript_type = "Signer")]
-    pub type JsSigner;
+    pub type Signer;
```
deuszx, line 69: The old name was made on purpose to not conflict with the `Signer` trait. Is this rename necessary?

web/@linera/client/src/lib.rs
```diff
@@ -366,6 +305,20 @@ impl Client {
         )?)
     }
 
+    /// Adds a new owner to the default chain.
+    ///
+    /// # Errors
+    ///
+    /// If the owner is in the wrong format, or the chain client can't be instantiated.
+    #[wasm_bindgen(js_name = addOwner)]
+    pub async fn add_owner(&self, owner: JsValue) -> JsResult<()> {
```
deuszx, line 314: We will soon have to support API where (web) clients can manage multiple chains – why not prepare this method for that right away?

examples/counter/metamask/index.html
```diff
@@ -69,9 +61,13 @@ <h2>Chain history for <code id="chain-id" class="hex">requesting chain…</code>
...
           const owner = await signer.address();
+          const chain = await faucet.claimChain(wallet, owner);
           document.getElementById('owner').innerText = owner;
-          document.getElementById('chain-id').innerText = await faucet.claimChain(wallet, owner);
-          const client = await new linera.Client(wallet, signer);
+          document.getElementById('chain-id').innerText = chain;
+          const autosigner = linera.signer.PrivateKey.createRandom();
+          const client = await new linera.Client(wallet, new linera.signer.Composite(autosigner, signer));
+          await client.addOwner(autosigner.address());
+          wallet.setOwner(chain, autosigner.address());
```
deuszx, line 73: Please, add comments here explaining what we're doing and what for – I link devs/builders to our examples often to explain things.

deuszx (approved):
LGTM :) this was a surprisingly small change

web/@linera/client/src/lib.rs
Twey: It is prepared :) This will become a chain method.

examples/counter/metamask/index.html
Twey: Good shout, thank you :)

web/@linera/client/src/signer/mod.rs
Twey: Probably not, but I was in there anyway :)

Everything in this crate is JavaScript-related, so the `Js` prefix should be unnecessary. I started it back when this crate was not so well-defined, but it's outlived its usefulness. 

I struggle a bit with the trait/implementation naming in Rust, but as a general rule I think you should sacrifice the names of imports (here `SignerTrait`) for the good of exports (here `Signer`), since then at least the badness is localized.

PR #5118, opened 2025-12-11
feat(helm): Use separate image for block-exporter

eldios:
## Summary
- Add dedicated `image` and `imagePullPolicy` configuration for `blockExporter` in the Helm chart
- Block exporter now uses its own image instead of reusing the indexer image
- This allows independent versioning of the block exporter container

## Changes
- `kubernetes/linera-validator/values.yaml`: Added `blockExporter.image` and `blockExporter.imagePullPolicy`
- `kubernetes/linera-validator/templates/block-exporter.yaml`: Changed from `indexer.image` to `blockExporter.image`

## Testing
- Helm template renders correctly with new values
- Requires corresponding lineractl changes (linera-infra PR)

### DIFF
--- a/kubernetes/linera-validator/templates/block-exporter.yaml
+++ b/kubernetes/linera-validator/templates/block-exporter.yaml
@@ -49,6 +49,44 @@ spec:
         effect: NoSchedule
       {{- end }}
       initContainers:
+        - name: linera-exporter-initializer
+          image: {{ .Values.lineraImage }}
+          imagePullPolicy: {{ .Values.lineraImagePullPolicy }}
+          securityContext:
+            runAsNonRoot: true
+            runAsUser: 65534
+            allowPrivilegeEscalation: false
+            capabilities:
+              drop:
+                - ALL
+          resources:
+            requests:
+              memory: "64Mi"
+              cpu: "50m"
+            limits:
+              memory: "128Mi"
+              cpu: "100m"
+          command:
+            - bash
+            - -c
+            - |
+              set -euo pipefail
+              while true; do
+                output=$(./linera storage check-existence --storage "{{ .Values.storage }}" 2>&1)
+                status=$?
+                if [ "$status" -eq 0 ]; then
+                  echo "Database already exists, no need to initialize."
+                  exit 0
+                else
+                  if [ "$status" -eq 1 ]; then
+                    echo "Database does not exist, retrying in {{ .Values.blockExporter.initRetryIntervalSeconds | default 5 }} seconds..."
+                  else
+                    echo "An unexpected error occurred (status: $status): $output"
+                    echo "Retrying in {{ .Values.blockExporter.initRetryIntervalSeconds | default 5 }} seconds..."
+                  fi
+                  sleep {{ .Values.blockExporter.initRetryIntervalSeconds | default 5 }}
+                fi
+              done
         - name: config-selector
           image: busybox
           command:


kubernetes/linera-validator/templates/block-exporter.yaml
```diff
@@ -49,44 +49,6 @@ spec:
         effect: NoSchedule
       {{- end }}
       initContainers:
-        - name: linera-exporter-initializer
```
deuszx, line 52: Are you replacing this with anyhting? It's still important to either keep restarting the exporter until scylla is initialized OR (as here) wait w/ starting it.

eldios: that's what kubernetes does.. if the main process dies (which is what I understood the linera-exporter will do if it can't connect to the DB) it will restart the container it belongs to indefinitely until it works

deuszx: Sounds good. I just wanted to make sure it will be taken care of.

ndr-ds: I don't think I agree with this. This is what init containers are for, isn't it? Making sure some state is true before initializing the main container?  
Now instead of just waiting on an init container, we'll basically crash loop until the db initializes 😅  
Seems worse IMHO. I think we should keep the init container.

ndr-ds: We should crash loop when something is wrong. Nothing will be necessarily wrong here.

deuszx: OK, you're right @ndr-ds . The "problem" here is that exporter's image no longer contains `linera` binary – which is a good thing. But we can use another docker image as the init step.

eldios:
@ndr-ds I'll answer here cause I can't answer in the comment apparently.

The new block-exporter image doesn't have the `linera` binary that would do the check, and since kubernetes restarts anyway, it's okay to "fail" as eventually the DB will be up and then be able to connect.. this is also only happening at the very beginning cause once the DB is up, it shouldn't happen.

I asked @deuszx and he confirmed it's okay and completely safe for the exporter to fail to connect and restart.

nit: init containers were originally meant for initializing the pod state (thus the name), the fact that they're used as pre-check containers is a sad story about how we never got a dedicated feature about it in kubernetes

ndr-ds:
Can't we just use a docker image that has the `linera` binary there then instead? Like @deuszx suggested

kubernetes/linera-validator/templates/block-exporter.yaml
```diff
@@ -50,8 +50,8 @@ spec:
       {{- end }}
       initContainers:
         - name: linera-exporter-initializer
-          image: {{ .Values.indexer.image }}
-          imagePullPolicy: {{ .Values.indexer.imagePullPolicy }}
+          image: {{ .Values.lineraImage }}
+          imagePullPolicy: {{ .Values.lineraImagePullPolicy }}
```
deuszx, line 54: Is this a type? Shouldn't it be `.Values.lineraImage.ImagePullPolicy` ?

ndr-ds (approved):
lgtm

kubernetes/linera-validator/templates/block-exporter.yaml
```diff
@@ -50,8 +50,8 @@ spec:
       {{- end }}
       initContainers:
         - name: linera-exporter-initializer
-          image: {{ .Values.indexer.image }}
-          imagePullPolicy: {{ .Values.indexer.imagePullPolicy }}
+          image: {{ .Values.lineraImage }}
```
deuszx, line 53: And here – it seems we uesd to have the pattern of `.Values.<name>.image` but not here anymore

kubernetes/linera-validator/templates/block-exporter.yaml
eldios: sadly .. no 😅 we'll have to clean up those options, but it's a task for another day

kubernetes/linera-validator/templates/block-exporter.yaml
eldios: same as in the other comment.. this is all old (non) standard code. We'll normalize and clean up all these options, but not today.

PR #5126, opened 2025-12-11
Autosigning improvements

Twey:
## Motivation

@afck pointed out that having a lot of owners with non-zero weights would slow the chain down in single-leader rounds.

@ma2bd pointed out that we need to make mutations using the original owner.

## Proposal

Add a `weight` option to `addOwner` in an options object, and default it to zero.

Add an optional `owner` option to `query` in an options object, and use it to sign mutations.  Also, document the above inline.

## Test Plan

Tested locally.  Also, CI should test some of the API (but sadly not yet as much as we'd like it to).

## Release Plan

    - be released in a new SDK,

## Links

### DIFF
--- a/web/@linera/client/src/lib.rs
+++ b/web/@linera/client/src/lib.rs
@@ -134,6 +134,13 @@ struct TransferParams {
     recipient: linera_base::identifiers::Account,
 }
 
+#[derive(Default, serde::Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct QueryOptions {
+    block_hash: Option<String>,
+    owner: Option<AccountOwner>,
+}
+
 #[wasm_bindgen]
 impl Client {
     /// Creates a new client and connects to the network.
@@ -314,11 +321,21 @@ impl Client {
     ///
     /// If the owner is in the wrong format, or the chain client can't be instantiated.
     #[wasm_bindgen(js_name = addOwner)]
-    pub async fn add_owner(&self, owner: JsValue) -> JsResult<()> {
+    pub async fn add_owner(&self, owner: JsValue, options: JsValue) -> JsResult<()> {
+        #[derive(Default, serde::Deserialize)]
+        struct Options {
+            #[serde(default)]
+            weight: u64,
+        }
+
         let owner = serde_wasm_bindgen::from_value(owner)?;
+        let Options { weight } =
+            serde_wasm_bindgen::from_value::<Option<_>>(options)?.unwrap_or_default();
         let chain_client = self.default_chain_client().await?;
-        self.apply_client_command(&chain_client, || chain_client.share_ownership(owner, 100))
-            .await??;
+        self.apply_client_command(&chain_client, || {
+            chain_client.share_ownership(owner, weight)
+        })
+        .await??;
         Ok(())
     }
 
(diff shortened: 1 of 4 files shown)


web/@linera/client/src/lib.rs
```diff
@@ -408,9 +425,14 @@ impl Application {
...
-    pub async fn query(&self, query: &str, block_hash: Option<String>) -> JsResult<String> {
+    pub async fn query(&self, query: &str, options: JsValue) -> JsResult<String> {
         tracing::debug!("querying application: {query}");
-        let chain_client = self.client.default_chain_client().await?;
+        let QueryOptions { block_hash, owner } =
+            serde_wasm_bindgen::from_value::<Option<_>>(options)?.unwrap_or_default();
+        let mut chain_client = self.client.default_chain_client().await?;
+        if let Some(owner) = owner {
+            chain_client.set_preferred_owner(owner);
+        }
```
deuszx, line 435: With this, if client receives an epoch-changing event it'd still ask user to sign manually (with the preferred owner). Instead, the `options.owner` should be set as preferrable only for the duration of the `query` and reset it back to the previous one at the end of it. Otherwise it's nullifying the potential autosigner.

Twey: > With this, if client receives an epoch-changing event it'd still ask user to sign manually (with the preferred owner).

The owner is a parameter of `query`, and epoch-changing events are processed during `process_inbox`, not during `query_application`.  So they'll use the (wallet's) preferred owner, which is the autosigning owner.

> Instead, the `options.owner` should be set as preferrable only for the duration of the query and reset it back to the previous one at the end of it.

Indeed, that's how this works: the chain client (which is created only for the duration of this query, then destroyed) is updated, not the wallet (which sets the default owner for all future chain clients created).

deuszx: `set_preferred_owner` calls `update_wallet_from_client` which updates the wallet.

deuszx: Ah nvm... There are two methods with the same name – `set_preferred_owner` – one on `ChainClient` and another on `ClientContext` 🤦 and the former (as is the case here) only changes for the instance as you say..

web/@linera/client/src/lib.rs
```diff
@@ -314,11 +321,21 @@ impl Client {
     ///
     /// If the owner is in the wrong format, or the chain client can't be instantiated.
     #[wasm_bindgen(js_name = addOwner)]
-    pub async fn add_owner(&self, owner: JsValue) -> JsResult<()> {
+    pub async fn add_owner(&self, owner: JsValue, options: JsValue) -> JsResult<()> {
+        #[derive(Default, serde::Deserialize)]
+        struct Options {
+            #[serde(default)]
+            weight: u64,
```
deuszx, line 328: Can you show the example how to construct this on the frontend side? As an example you can just pass instance with `weight: 0` here.

Twey: There's a bigger piece of work to make sure all the TypeScript types are correct. (I think right now wherever we have `JsValue` here we get `any`). I'll probably introduce `tsify-next` in my next PR.

PR #5127, opened 2025-12-11
Autosigning on the Web (#5116)

Twey:
## Motivation

Signing blocks on the Web can get tiresome when using wallets like MetaMask or Dynamic, since we aren't integrated with their autosigning features, and working with chains with hundreds of incoming messages typically means accepting hundreds of signatures.

## Proposal

First connect to the ‘real’ wallet, and use it to sign a block that adds a new local keypair as an additional owner of the chain. Then, replace the original owner of the chain with this new temporary owner in the wallet so that future blocks will be signed with that key instead. We currently keep the wallet in memory on the Web so needn't worry about overriding the real owner.

This is a slightly clumsy approach (and the application must be careful to send any assets back to the original owner!) but should work as a first step.

Implement a little supporting API to the Web client, then demo an autosigning flow in the MetaMask counter demo.

## Test Plan

Tested by hand.

## Release Plan

    - be released in a new SDK,

## Links


examples/counter/metamask/index.html
```diff
@@ -51,14 +51,6 @@ <h2>Chain history for <code id="chain-id" class="hex">requesting chain…</code>
       </div>
     </div>
 
-    <script type="importmap">
-      {
-          "imports": {
-              "@linera/client": "./js/@linera/client/dist/index.js"
-          }
-      }
-    </script>
```
deuszx, line 60: Why is this importmap not necessary anymore?

web/@linera/client/src/lib.rs
```diff
@@ -365,6 +304,20 @@ impl Client {
...
+    /// Adds a new owner to the default chain.
+    ///
+    /// # Errors
+    ///
+    /// If the owner is in the wrong format, or the chain client can't be instantiated.
+    #[wasm_bindgen(js_name = addOwner)]
+    pub async fn add_owner(&self, owner: JsValue) -> JsResult<()> {
+        let owner = serde_wasm_bindgen::from_value(owner)?;
+        let chain_client = self.default_chain_client().await?;
+        self.apply_client_command(&chain_client, || chain_client.share_ownership(owner, 100))
```
deuszx, line 316: Sharing ownership turns all superowners into regular owners as a side effect.

examples/counter/metamask/index.html
Twey: With the introduction of the `web-thread` library bundling no longer breaks `@linera/client`, so we can now just use Vite's normal resolution functionality.

Issue #4740, opened 2025-10-06
`@linera/client`: add options object


Twey:
We have the `ClientOptions` object for configuring the Linera client, but currently only an ad-hoc subset of it is exposed in the Web client API.

We should expose (a JavaScript marshalling of) the whole object in the `linera.Client` constructor.

Twey:
Fixed in https://github.com/linera-io/linera-protocol/pull/5143.

PR #5143, opened 2025-12-13
Multichain API for Web

Twey:
## Motivation

We have several internal and external use cases for dealing with more than one chain at once.

## Proposal

Add chain-specific API, and have `Client` only initialize the client rather than creating a default chain.  This provides a place to put the (currently hardcoded) client options, which I do, closing https://github.com/linera-io/linera-protocol/issues/4740.

## Test Plan

Tested locally on all our Web examples.

## Release Plan

    - be released in a new SDK,

## Links

### DIFF
--- a/linera-client/src/chain_listener.rs
+++ b/linera-client/src/chain_listener.rs
@@ -35,7 +35,9 @@ use tracing::{debug, error, info, instrument, warn, Instrument as _};
 
 use crate::error::{self, Error};
 
-#[derive(Default, Debug, Clone, clap::Args, serde::Serialize, serde::Deserialize, tsify_next::Tsify)]
+#[derive(
+    Default, Debug, Clone, clap::Args, serde::Serialize, serde::Deserialize, tsify_next::Tsify,
+)]
 #[serde(rename_all = "camelCase")]
 pub struct ChainListenerConfig {
     /// Do not create blocks automatically to receive incoming messages. Instead, wait for
--- a/web/@linera/client/src/signer/mod.rs
+++ b/web/@linera/client/src/signer/mod.rs
@@ -62,7 +62,6 @@ impl From<JsValue> for Error {
 
 impl std::error::Error for Error {}
 
-
 #[wasm_bindgen(typescript_custom_section)]
 const _: &str = r#"import type { Signer } from '../signer/index.js';"#;
 
@@ -72,7 +71,11 @@ extern "C" {
     pub type Signer;
 
     #[wasm_bindgen(catch, method)]
-    async fn sign(this: &Signer, owner: AccountOwner, value: Vec<u8>) -> Result<js_sys::JsString, JsValue>;
+    async fn sign(
+        this: &Signer,
+        owner: AccountOwner,
+        value: Vec<u8>,
+    ) -> Result<js_sys::JsString, JsValue>;
 
     #[wasm_bindgen(catch, method, js_name = "containsKey")]
     async fn contains_key(this: &Signer, owner: AccountOwner) -> Result<JsValue, JsValue>;
@@ -91,15 +94,19 @@ impl linera_base::crypto::Signer for Signer {
         value: &CryptoHash,
     ) -> Result<AccountSignature, Self::Error> {
         Ok(AccountSignature::EvmSecp256k1 {
-            signature: String::from(self
-                // Pass CryptoHash without serializing as that adds bytes
-                // to the serialized value which we don't want for signing.
-                .sign(*owner, value.as_bytes().0.to_vec())
-                .await?).parse().map_err(|_| Error::UnexpectedSignatureFormat)?,
+            signature: String::from(
+                self
+                    // Pass CryptoHash without serializing as that adds bytes
+                    // to the serialized value which we don't want for signing.
+                    .sign(*owner, value.as_bytes().0.to_vec())
+                    .await?,
+            )
+            .parse()
+            .map_err(|_| Error::UnexpectedSignatureFormat)?,
             address: if let AccountOwner::Address20(address) = owner {
                 *address
             } else {
-                return Err(Error::InvalidAccountOwnerType)
+                return Err(Error::InvalidAccountOwnerType);
             },
         })
     }


web/@linera/client/src/chain/mod.rs
```diff
@@ -0,0 +1,192 @@
...
+
+pub mod application;
+pub use application::Application;
+
+#[wasm_bindgen]
+pub struct Chain {
+    // TODO(TODO) required for now for wallet updates, but we should move that
+    // functionality into `linera-core`
+    pub(crate) client: Arc<AsyncMutex<ClientContext<Environment>>>,
+    pub(crate) chain_client: ChainClient<Environment>,
```
deuszx: IMO this shows that our API has to be refactored – client and chain_client shouldn't be independent structs.

web/@linera/client/src/lib.rs
```diff
@@ -124,14 +53,38 @@ pub struct Client {
     // This use of `futures::lock::Mutex` is safe because we only
     // expose concurrency to the browser, which must always run all
     // futures on the global task queue.
-    client_context: Arc<AsyncMutex<ClientContext>>,
+    // It does nothing here in this single-threaded context, but is
+    // hard-coded by `ChainListener`.
+    client: Arc<AsyncMutex<linera_client::ClientContext<Environment>>>,
```
deuszx: rename to `client_context`.

web/@linera/client/src/chain/application.rs
```diff
@@ -0,0 +1,109 @@
...
+                })
+                .await?;
+        }
+
+        Ok(String::from_utf8(response)?)
+    }
+}
+
+#[wasm_bindgen(start)]
+pub fn main() {
```
deuszx: Do you think with these we would get the logs back? 🤔 w/o the need for the fix I made?

web/@linera/client/src/chain/mod.rs
Twey: For sure.  The functionality available on the `ClientContext` should be available on the `Client`, which is accessible through the `ChainClient`.  It's a refactor that's a little out of scope here though.

web/@linera/client/src/lib.rs
Twey: `ClientContext` is misnamed now, it's really just a client.  I can rename it if you like?

web/@linera/client/src/chain/application.rs
Twey: Oh this shouldn't be here, it should be in main :) Draft!

But no, this has the same behaviour (is the same code) as before this PR.  The diff is just a little messy right now.

web/@linera/client/src/lib.rs
deuszx: I'm not convinced that renaming `ClientContext` to client is a good idea – `ClientContext` already has `Client` (returns it from a method). And `ChainClient` has a `Client` as well.. Too many clients – that will be confusing.

Twey: We have lots of clients — faucet clients, HTTP clients, gRPC clients, Ethereum clients… :)  The two clients you mention are the same client, though.  Currently we have five or six (Linera) client types, all of which are clients for the Linera protocol differing only in the external interface they present:
- the core client (`linera_core::Client`) is the internal client implementation that exposes details of the protocol, (like other things in `linera-core`) designed to be used as a building block
- `linera_client::ClientContext` is a client that exposes our stable (user-facing) Rust client API; all the other client façades should follow this API
- `linera_web::Client` is the client exposed as a JavaScript API
- the node service (`linera_service::node_service`) is the client exposed as a GraphQL API
- the CLI (`linera_service::cli`) is the client exposed as a shell command

Notionally there is only one client object, with its logic implemented in `linera-core`, but it exposes several different façades depending on which APIs you want to use to access it.  That's why they all are (or maybe should) be called the same thing.  `ClientContext` is (now) simply misnamed for historical reasons: it isn't a context for the client (that's provided by the `linera_core::Environment` trait and its various implementations) but actually a client itself, plus some additional data (that are now redundant, actually: we could remove them, but that's out of scope of this PR).

The original split between `linera-core` and `linera-client` is that `linera-client` (originally, `linera-service`) is supposed to expose higher-level functionality on top of `linera-core` that makes the API easier to use — things like retry loops, dependency injection mechanisms, or type erasure that are convenient defaults for most users but maybe not appropriate for all occasions (with a particularly important case being testing).  I think it's also an opportunity for us to expose our stable client API that we want to commit to long-term.  As such there's no room for confusion: the only (Rust) client that the end-user will be exposed to is the one from `linera-client`.

Personally if you have an object that mediates access to or stores details about a ‘foo’, I prefer just to name it ‘Foo’ — not `FooClient`, `FooId`, `FooAddress`, `FooData`, et cetera.  These are implementation details of how the object represents the foo: I'd much rather just name it for the ‘foo’ object itself.  But we do have a history of naming things ‘client’ that I don't really want to tackle in this PR.  Last time I tried it we struggled to settle on exactly what object our `Client`s represent.

deuszx: I don't agree the argument with multiple clients is a valid explanation for naming one more thing the same way. It's fine to duplicate the name if it's used in different contexts. The problem we have here is that they coexist literally next to each other. We can't have two things called the same, in the same struct.

deuszx: I didn't ask to rename the type - I asked to rename the value.

Twey: > The problem we have here is that they coexist literally next to each other. We can't have two things called the same, in the same struct.

`linera_core::Client` and `linera_client::Client` are not the same name, and also are (thankfully) unlikely to live next to each other, as they belong to different levels of abstraction.  `linera_core::Client` and `linera_core::ChainClient` might live next to each other, but are also not the same name :) 

> I didn't ask to rename the type - I asked to rename the value.

I'm not really convinced that it's a good idea to misname the value to match an incorrect type name.  I think I'd rather improve the situation where I find it.  But I think I can find a good compromise for now…

Twey: (resolving because it no longer applies, but feel free to re-open if necessary)

deuszx: I was meaning to resolve it myself tomorrow.

examples/native-fungible/index.html
```diff
@@ -226,43 +218,48 @@ <h2>Chain history for <code id="chain-id" class="hex">requesting a new microchai
...
           const owner = signer.address();
           const chainId = await faucet.claimChain(wallet, owner);
-          const client = await new linera.Client(wallet, signer, false, true);
+          console.warn(6);
+          const client = await new linera.Client(wallet, signer);
+          console.warn(7);
           document.querySelector('#chain-id').innerText = chainId;
           document.querySelector('#owner').innerText = owner;
 
-          console.log('using application id:', import.meta.env.LINERA_APPLICATION_ID);
```
deuszx, line 240: Why are you removing this? I added it on purpose – very often we want to check what's the applicaton ID we're working with and it's nowhere to be found.

At the same time – you left lots of debugging console.warn-s

examples/native-fungible/index.html
```diff
@@ -226,43 +218,48 @@ <h2>Chain history for <code id="chain-id" class="hex">requesting a new microchai
               navigator.clipboard.writeText(document.querySelector('#account-id').textContent);
           });
 
+          console.warn(1);
```
ma2bd: is this going to be removed?

web/@linera/client/src/chain/application.rs
```diff
@@ -0,0 +1,84 @@
...
+    ///
+    /// # Errors
+    /// If the application ID is invalid, the query is incorrect, or
+    /// the response isn't valid UTF-8.
+    ///
+    /// # Panics
+    /// On internal protocol errors.
+    #[wasm_bindgen]
+    // TODO(#14) allow passing bytes here rather than just strings
+    // TODO(#15) a lot of this logic is shared with `linera_service::node_service`
```
ma2bd: Do you to reference task from another repository?

examples/native-fungible/index.html
Twey: Ah, I meant to move this into the client library itself so we get it for all applications — good catch, thank you :)

examples/native-fungible/index.html
Twey: It should have been!

web/@linera/client/src/chain/application.rs
Twey: Not newly, but yes.  @deuszx kindly ported them all over to this repository; I'll fix up the references.

examples/native-fungible/index.html
Twey: https://github.com/linera-io/linera-protocol/pull/5143/commits/32acdae33470c34d00830f59bdc5aae7671ec3bd

examples/native-fungible/index.html
Twey: https://github.com/linera-io/linera-protocol/pull/5143/commits/32acdae33470c34d00830f59bdc5aae7671ec3bd

web/@linera/client/src/chain/application.rs
Twey: https://github.com/linera-io/linera-protocol/pull/5143/commits/141ab700ebf368282546c406b4ecfbf9c580a101

PR #5146, opened 2025-12-14
Autosigning improvements (#5126)

Twey:
## Motivation

@afck pointed out that having a lot of owners with non-zero weights would slow the chain down in single-leader rounds.

@ma2bd pointed out that we need to make mutations using the original owner.

## Proposal

Add a `weight` option to `addOwner` in an options object, and default it to zero.

Add an optional `owner` option to `query` in an options object, and use it to sign mutations. Also, document the above inline.

## Test Plan

Tested locally. Also, CI should test some of the API (but sadly not yet as much as we'd like it to).

## Release Plan

    - be released in a new SDK,

## Links

### DIFF
--- a/examples/counter/index.html
+++ b/examples/counter/index.html
@@ -82,19 +82,19 @@
               logs.insertBefore(entry, logs.firstChild);
           }
 
-          async function updateCount(block_hash) {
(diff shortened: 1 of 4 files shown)


Twey:
(YOLO'ing because we need it for PM and the changes are already accepted for `main`)

PR #5151, opened 2025-12-15
Attempt at resolving the issue.

MathieuDutSik:
## Motivation

Some "web" tests are not passing, and we need to correct them.

## Proposal

Do the synchronization only if we have some chains.

## Release Plan

To be merged in Testnet Conway.

## Links

None.


MathieuDutSik:
Not a useful PR.

PR #5156, opened 2025-12-15
Keep chains sparse after a service restart

bart-linera:
## Motivation

If a client listens only to events from a chain, and builds a sparse event chain this way, then restarts the service, the restarted service will attempt to synchronize the whole chain, filling in all the gaps and defeating the purpose of a sparse chain.

## Proposal

Do not synchronize chains we listen to in `EventOnly` mode. Instead, when a new event arrives, check that the block that has emitted the previous event is present. If not, synchronize that block and repeat.

This has the disadvantage that we won't synchronize missing events until a new one appears, so a better solution might be needed.

Written partially by Claude.

## Test Plan

CI

An existing test has been modified to showcase the issue. It is failing without the fix, and passing with the fix.

## Release Plan

## Links


linera-core/src/client/chain_client/mod.rs
```diff
@@ -2711,16 +2714,21 @@ impl<Env: Environment> ChainClient<Env> {
...
+                    // For EventsOnly mode, skip full sync - we'll lazily download
+                    // event-relevant blocks when we receive notifications.
                     let remote_node = RemoteNode { public_key, node };
-                    this.client
-                        .synchronize_chain_state_from(&remote_node, this.chain_id)
-                        .await?;
+                    if !matches!(listening_mode_for_sync, ListeningMode::EventsOnly(_)) {
+                        this.client
+                            .synchronize_chain_state_from(&remote_node, this.chain_id)
+                            .await?;
```
deuszx: An idea for the (future) chain/client refactor: a `ChainClient` instance contains the `ListeningMode` and each variant is a separate `ChainClient` type. Then, there's a few generic functions – like `synchronize` which depending on the client mode does the work. We'd get away with all these `if`s and early returns.

deuszx (approved):
I think that for what it's trying to do, this looks correct.

We need to discuss and decide whether we want to merge this approach or figure out something else to fix the issue mentioned in the description.

linera-core/src/client/mod.rs
```diff
@@ -1014,6 +1014,68 @@ impl<Env: Environment> Client<Env> {
         Ok(())
     }
 
+    /// Downloads any missing event blocks that the given certificate references in its
+    /// `previous_event_blocks` field, for the specified streams.
+    ///
+    /// This is used for lazy synchronization of EventsOnly chains: instead of downloading
```
```suggestion
    /// This is used for lazy synchronization of `EventsOnly` chains: instead of downloading
```
(Or "events-only chains"?)

linera-core/src/client/chain_client/mod.rs
```diff
@@ -2711,16 +2714,21 @@ impl<Env: Environment> ChainClient<Env> {
...
                     let stream = node.subscribe(vec![this.chain_id]).await?;
                     // Only now the notification stream is established. We may have missed
                     // notifications since the last time we synchronized.
+                    // For EventsOnly mode, skip full sync - we'll lazily download
+                    // event-relevant blocks when we receive notifications.
                     let remote_node = RemoteNode { public_key, node };
-                    this.client
-                        .synchronize_chain_state_from(&remote_node, this.chain_id)
-                        .await?;
+                    if !matches!(listening_mode_for_sync, ListeningMode::EventsOnly(_)) {
```
afck: Maybe this should be `if listening_mode_for_sync.should_sync_chain_state()`, and in that method we should have a `match` without a catch-all arm, to make sure that whenever we add or modify listening modes, we don't miss this.

linera-core/src/client/chain_client/mod.rs
```diff
@@ -2548,18 +2548,21 @@ impl<Env: Environment> ChainClient<Env> {
                     "NewEvents: processing notification"
                 );
                 let mut certificates = remote_node.node.download_certificates(vec![hash]).await?;
-                // download_certificates ensures that we will get exactly one
-                // certificate in the result.
                 let certificate = certificates
                     .pop()
                     .expect("download_certificates should have returned one certificate");
                 self.client
                     .receive_sender_certificate(
```
afck: Should we do this _after_ downloading the predecessors? Maybe we don't handle events that way right now, but it would maximize the number of messages that can be put in the outbox.

linera-core/src/client/mod.rs
```diff
@@ -1014,6 +1014,68 @@ impl<Env: Environment> Client<Env> {
...
+        self.collect_missing_event_block_predecessors(certificate, relevant_streams, &mut pending)
+            .await?;
+
+        while let Some((_, hash)) = pending.pop_last() {
+            let mut certificates = remote_node.node.download_certificates(vec![hash]).await?;
+            let prev_cert = certificates
+                .pop()
+                .ok_or_else(|| NodeError::MissingCertificates(vec![hash]))?;
+
+            self.handle_certificate(prev_cert.clone()).await?;
```
afck: (I see, here we also handle them in reverse order. I guess it's fine for events, but as above, for messages it would be better in chronological order.)

linera-core/src/client/mod.rs
```diff
@@ -1014,6 +1014,68 @@ impl<Env: Environment> Client<Env> {
...
+        for stream_id in relevant_streams {
+            if let Some((prev_hash, prev_height)) = certificate
+                .block()
+                .body
+                .previous_event_blocks
+                .get(stream_id)
+            {
+                if !self
+                    .storage_client()
+                    .contains_certificate(*prev_hash)
```
afck: That means if there were events for a stream of interest in blocks 2, 4 and 6, and we already have block 4 for a _different_ reason (e.g. it sent a message), then we would fail to download block 2?

I guess that's fine for now…? Otherwise we'd need another map in the chain state view to keep track of the lowest event we're still missing in each stream emitted by this chain.

linera-core/src/client/chain_client/mod.rs
bart-linera: Yes, that would be better conceptually, but I'm a bit concerned what happens if there are _lots_ of predecessors missing. We already seemed to have issues stemming from putting too much in memory before writing to disk in places...

linera-core/src/client/mod.rs
bart-linera: Good point. I was trying to avoid issues like that, but they still can happen with this code :grimacing: I think you're right about needing another map, but I think it will be best to fix this issue in this PR already.

linera-core/src/client/chain_client/mod.rs
```diff
@@ -2627,17 +2638,20 @@ impl<Env: Environment> ChainClient<Env> {
...
                     "NewEvents: processing notification"
                 );
-                let mut certificates = remote_node.node.download_certificates(vec![hash]).await?;
-                // download_certificates ensures that we will get exactly one
-                // certificate in the result.
-                let certificate = certificates
-                    .pop()
-                    .expect("download_certificates should have returned one certificate");
+                // Use the subscribed streams from EventsOnly mode, or fall back
+                // to the streams from the notification itself.
```
afck: Wait, if it isn't `EventsOnly`, we shouldn't do anything, should we? We'll react to `NewBlock` in that case, and can ignore `EventsOnly`?

bart-linera: Ah, good point!

bart-linera: I realized we also should only fetch the blocks for streams we actually care about. That should be fixed in 642aa0e

bart-linera: Actually, I believe there's one more thing that should be changed, which I did in 1fd64aa

linera-core/src/client/chain_client/mod.rs
```diff
@@ -2638,22 +2638,28 @@ impl<Env: Environment> ChainClient<Env> {
...
-                    _ => event_streams,
+                // Use the subscribed streams from EventsOnly mode to figure out which
+                // streams we are interested in.
+                // If the listening mode is not EventsOnly, we will process the
+                // certificate when we get the NewBlock notification about this block.
+                let relevant_streams = match self.listening_mode() {
+                    Some(ListeningMode::EventsOnly(subscribed)) => {
+                        subscribed.intersection(&event_streams).cloned().collect()
+                    }
+                    _ => BTreeSet::new(),
```
afck: But then we might as well `return`. And we should make `is_relevant` return `false` then in the first place.

bart-linera: Makes sense. And regarding `is_relevant`, that is actually the case. Maybe we should also make `NewEvents` not relevant to `FullChain` and `FollowChain`, since they will process the `NewBlock` notifications? (so then the `if !matches!` in 1fd64aa would also become unnecessary)

afck: Yes, that's what I mean. But I'm currently trying that on `testnet_conway` and running into issues, so feel free to omit that change for now. We can always forward-port it separately.

PR #5163, opened 2025-12-16
Background sync of received certificates fix

deuszx:
## Motivation

We are synchronising sender chains that we shouldn't be b/c they're sending messages to a chain we follow in `FollowOnly` mode.

## Proposal

Filter out chains which we don't follow in `FullChain` mode.

## Test Plan

manual.

## Release Plan

- These changes should be
    - be released in a new SDK,
    - backported to `main`

## Links


ma2bd (approved):
makes sense

PR #5183, opened 2025-12-17
chore: bump checkout to v6

Dahka2321:
Updated actions/checkout from v4 to v6 in ci.yml workflow


ma2bd:
We should update the github action consistently but also this is not super important

PR #5184, opened 2025-12-18
Fix docker compose validator

ndr-ds:
## Motivation

The Docker Compose validator setup has a bug: all 4 shard containers run with --shard 0 instead of each having a unique shard number (0, 1, 2, 3). This causes inter-shard communication to fail because all shards are configured identically rather than as distinct shards of the validator.

Additionally, the proxy health check was checking port 19100, which is an internal port. Checking on port 21100 instead, which is for metrics.

## Proposal

1. Fix shard configuration: Replace the single shard service using replicas: 4 with 4 separate services (shard-0, shard-1, shard-2, shard-3), each with the correct --shard argument. Docker Compose's replicas feature doesn't support per-replica command arguments, so separate service definitions are required.
2. Fix proxy health check: Change the health check port from 19100 to 21100 (the metrics port that the proxy actually listens on).
3. Update alloy override: Update docker-compose.alloy.yml dependencies to reference the new shard service names.

## Test Plan

- Deployed the fix to the OVH docker compose validator (ovh-or1-compose-validator-01)
- Verified all 4 shards start with correct shard numbers (0, 1, 2, 3)
- Verified inter-shard connectivity works
- Verified proxy health check passes

## Release Plan

### DIFF
--- a/docker/alloy-config.river
+++ b/docker/alloy-config.river
@@ -19,17 +19,36 @@ prometheus.scrape "proxy_metrics" {
   scrape_timeout  = "10s"
 }
 
-// Scrape metrics from shard services (all 4 replicas)
+// Scrape metrics from all shard services
 prometheus.scrape "shard_metrics" {
   targets = [
     {
-      __address__ = "shard:21100",
+      __address__ = "docker-shard-1:21100",
       job         = "linera-shard",
+      shard       = "0",
+      instance    = env("HOSTNAME"),
+    },
+    {
+      __address__ = "docker-shard-2:21100",
+      job         = "linera-shard",
+      shard       = "1",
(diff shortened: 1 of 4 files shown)


ndr-ds:
* **#5184** <a href="https://app.graphite.com/github/pr/linera-io/linera-protocol/5184?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/linera-io/linera-protocol/5184?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ma2bd (approved):
omg thanks!

PR #5188, opened 2025-12-18
Make sure background sync is correctly started for chains being listened to

bart-linera:
## Motivation

Node service plugins introduced an interface in the `ChainListener` that allows contracts to request listening to chains. However, the background sync task wasn't started for chains being listened to due to such requests.

## Proposal

Make sure background sync is started if there is a request to listen in the `FullChain` mode.

## Test Plan

CI will catch regressions

## Release Plan

    - be released in a new SDK,

## Links

- Also contains the change from #5163 

### DIFF
--- a/linera-client/src/chain_listener.rs
+++ b/linera-client/src/chain_listener.rs
@@ -302,35 +302,6 @@ impl<C: ClientContext + 'static> ChainListener<C> {
         })
     }
 
-    async fn start_background_sync(
-        &mut self,
-        chain_id: ChainId,
-        mode: ListeningMode,
-    ) -> Option<CancellationToken> {
-        if !self.enable_background_sync {
-            return None;
-        }
-        if mode != ListeningMode::FullChain {
-            return None;
-        }
-        let context = Arc::clone(&self.context);
-        let cancellation_token = CancellationToken::new();
-        let cancellation_token_clone = cancellation_token.clone();
-        linera_base::task::spawn(async move {
-            if let Err(e) = Self::background_sync_received_certificates(
-                context,
-                chain_id,
-                cancellation_token_clone,
-            )
-            .await
-            {
-                warn!("Background sync failed for chain {chain_id}: {e}");
-            }
-        })
-        .forget();
-        Some(cancellation_token)
-    }
-
     /// Processes a notification, updating local chains and validators as needed.
     async fn process_notification(&mut self, notification: Notification) -> Result<(), Error> {
         Self::sleep(self.config.delay_before_ms).await;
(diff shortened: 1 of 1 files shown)


bart-linera:
Hmm, fair point. I mean - if we change to `FullChain` from any other mode, this will work correctly, because it will just start the task. But if we stop listening, the background task won't be stopped :thinking:  Maybe we should keep a handle, or a separate cancellation token, in the `ListeningClient`, so that we can abort it when the `ListeningClient` is stopped?

BTW, it's currently not possible to switch listening to a "lower" mode - the assumption being that if there are two separate reasons for listening, and one implies a "higher" mode than the other, we should just be listening in the "higher" mode of the two. More sophisticated management of the listening tasks would require a lot more work, I'm afraid :grimacing:

linera-client/src/chain_listener.rs
```diff
@@ -232,15 +242,13 @@ impl<C: ClientContext + 'static> ChainListener<C> {
...
         }
     }
 
     /// Runs the chain listener.
     #[instrument(skip(self))]
-    pub async fn run(
-        mut self,
-        enable_background_sync: bool,
-    ) -> Result<impl Future<Output = Result<(), Error>>, Error> {
+    pub async fn run(mut self) -> Result<impl Future<Output = Result<(), Error>>, Error> {
```
deuszx, line 251: Was this change necessary?

linera-client/src/chain_listener.rs
```diff
@@ -510,13 +495,43 @@ impl<C: ClientContext + 'static> ChainListener<C> {
...
+        &mut self,
+        chain_id: ChainId,
+        mode: ListeningMode,
+    ) -> Option<CancellationToken> {
+        if !self.enable_background_sync {
+            return None;
+        }
+        if mode != ListeningMode::FullChain {
+            return None;
+        }
```
deuszx, line 516: We could add some logging for the less obvious cases:
1. `enable_background_sync == true` but `mode != ListeningMode::FullChain`
2. `enable_background_sync == false` but `mode == ListeningMode::FullChain`

🤔 

Actually, how could we end up with `FullChain` listening mode but disabled background sync?

linera-client/src/chain_listener.rs
bart-linera: Yes, because background sync for a chain can now be started at a point different than when the whole Chain Listener is started - so we need to remember the setting in the struct itself, to know whether we want to start background sync for chains or not.

linera-client/src/chain_listener.rs
bart-linera: I think we enable background sync in all production scenarios, anyway. This setting only exists to make it possible to disable starting background sync tasks in unit tests, AFAIK.

linera-client/src/chain_listener.rs
deuszx: This sounds wrong – i.e. we decide to _spawn_ the background sync at a different point in time than when we decide whether we (will) want to run it. Shouldn't those two be done at the same time?

bart-linera: But background sync is per chain, and we may want to start listening to a chain at any point while the listener (as in, the component) is running. So we initialize the listener at startup, but then we might want to start to listen to some chain much later, and if background sync is enabled, we will start the background sync for that chain then.

linera-client/src/chain_listener.rs
deuszx: Can we add some logging regardless?

bart-linera: I'm not sure if it's that useful - because both cases will be pretty common. The first one will pop up in production when listening to sparse event chains, or with followed chains. The second one will be everywhere in unit tests, where we disable background sync for convenience. Neither case is unusual in any way (but I can of course add some `debug` or maybe even `trace` logs).

PR #5189, opened 2025-12-18
Returns generated block of processed messages in handle_received_messages

kikakkz:
## Motivation

Right now if we create a chain in message which is processed in `handle_received_messages` , we cannot get the `ChainDescription` of the created chain.

## Proposal

Returns the generated block from `handle_received_messages` then unit tests can get `ChainDescription` from the block.

## Test Plan

CI

## Release Plan

These changes follow the usual release cycle

### DIFF
--- a/linera-sdk/src/test/chain.rs
+++ b/linera-sdk/src/test/chain.rs
@@ -295,7 +295,7 @@ impl ActiveChain {
     /// Adds a block to this microchain that receives all queued messages in the microchains
     /// inboxes.
     ///
-    /// Returns the generate block of processed messages.
+    /// Returns the certificate of the latest block added to the chain, if any.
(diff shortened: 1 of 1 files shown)


linera-sdk/src/test/chain.rs
```diff
@@ -294,7 +294,9 @@ impl ActiveChain {
     ///
     /// Adds a block to this microchain that receives all queued messages in the microchains
     /// inboxes.
-    pub async fn handle_received_messages(&self) {
+    ///
+    /// Returns the generate block of processed messages.
```
deuszx: I'd reword this to:

_Returns the certificate of the latest block added to the chain, if any._

afck (approved):

PR #5196, opened 2025-12-24
Add option `--public` to the node service

### LINKED ISSUE #5195: add an option to make the node service expose only public (i.e. safe) APIs
This means disallowing the main GraphQL mutation root and preventing application services from scheduling operations.

The name of the option could be something like `--public`.

ma2bd:
## Motivation

fixes #5195

## Proposal

Add option `--public` to the node service

## Test Plan

TBD

## Release Plan

`main` and `testnet_conway`


ma2bd:
superseded by #5201

PR #5197, opened 2025-12-26
fix missing logs in tests

ma2bd:
## Motivation

#5142 only solved the issue for one type of tests.

## Proposal

Only CLI users want logs disabled by default.

## Test Plan

tested manually

## Release Plan

will be backported to `main`

### DIFF
--- a/linera-execution/src/lib.rs
+++ b/linera-execution/src/lib.rs
@@ -463,11 +463,19 @@ pub trait UserService {
 }
 
 /// Configuration options for the execution runtime available to applications.
-#[derive(Clone, Copy, Default)]
+#[derive(Clone, Copy)]
 pub struct ExecutionRuntimeConfig {
     /// Whether contract log messages should be output.
     /// This is typically enabled for clients but disabled for validators.
     pub allow_application_logs: bool,
+}
+
+impl Default for ExecutionRuntimeConfig {
+    fn default() -> Self {
+        Self {
+            allow_application_logs: true,
+        }
+    }
 }
 
 /// Requirements for the `extra` field in our state views (and notably the


linera-execution/src/lib.rs
```diff
@@ -463,13 +463,21 @@ pub trait UserService {
...
     pub allow_application_logs: bool,
 }
 
+impl Default for ExecutionRuntimeConfig {
+    fn default() -> Self {
+        Self {
+            allow_application_logs: true,
+        }
+    }
+}
```
MathieuDutSik, line 479: If we would like to enable the logs in tests, then setting the defaults like that is, of course, addresses it. But that seems overkill. We want logs.

Then having a `Default` implementation depending on whether we are in tests or not is the next idea. But that is an anti design and it merely indicates that the Default means different things to different people (for me, it means no logs).

So, I propose the following:
* Removing the `Default` trait implementation for `ExecutionRuntimeConfig`.
* In `DbStorage::new` replace `ExecutionRuntimeConfig::default` by `ExecutionRuntimeConfig::no_logs()`.
* In `TestExecutionRuntimeConfig::execution_runtime_config` return the `ExecutionRuntimeConfig::with_logs`. We do set up the `execution_runtime_config` to default in all use cases it seems. So setting up the logs for tests seems fine.

MathieuDutSik (commented):
I do not like the use of default. But if that is a blocker for you, I am ready to approve it.

linera-execution/src/lib.rs
ma2bd: The names `no_logs` and `with_logs` are not future proof if `ExecutionRuntimeConfig` grows.

ma2bd:
> I do not like the use of default. But if that is a blocker for you, I am ready to approve it.

Yes this is a high priority fix, not a refactoring

PR #5199, opened 2025-12-29
Correction to the remote EVM contract call.

MathieuDutSik:
## Motivation

In previous work, the ability to create an EVM contract was introduced.
We already had the functionality of calling an EVM contract from another EVM contract.
So, it was reasonable to think that the created contract B from contract A could be called from contract A. That is not so. The contract had to be registered.

## Proposal

Add the registration of the created contract to the journal (this was found by extensive
interaction with Claude).

## Test Plan

Add the calls of contract B from contract A. This has been combined with calls directly to contract B.
Then the values are checked by service queries.

## Release Plan

I would support enabling EVM in TestNet Conway.

## Links

None.

### DIFF
--- a/linera-execution/src/evm/revm.rs
+++ b/linera-execution/src/evm/revm.rs
@@ -985,43 +985,49 @@ impl<Runtime: ContractRuntime> CallInterceptorContract<Runtime> {
                 )?;
             }
             let module_id = Self::publish_create_inputs(context, inputs)?;
-            let mut runtime = context.db().0.runtime.lock().unwrap();
-            let chain_id = runtime.chain_id()?;
-            let application_id = runtime.application_id()?;
-            let expected_application_id =
-                Self::get_expected_application_id(&mut runtime, module_id)?;
-            if inputs.value != U256::ZERO {
-                let amount = Amount::try_from(inputs.value).map_err(EvmExecutionError::from)?;
-                let destination = Account {
-                    chain_id,
-                    owner: expected_application_id.into(),
+            let (deployed_bytecode, address) = {
+                let mut runtime = context.db().0.runtime.lock().unwrap();
+                let chain_id = runtime.chain_id()?;
+                let application_id = runtime.application_id()?;
+                let expected_application_id =
+                    Self::get_expected_application_id(&mut runtime, module_id)?;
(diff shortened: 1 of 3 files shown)


linera-execution/src/evm/revm.rs
afck, line 972: If we `return`ed early here, the whole rest of the function would have less indentation.

linera-execution/src/evm/revm.rs
```diff
@@ -985,43 +985,49 @@ impl<Runtime: ContractRuntime> CallInterceptorContract<Runtime> {
...
-                argument,
-                required_application_ids,
-            )?;
-            assert_eq!(expected_application_id, created_application_id);
-            let argument = GET_DEPLOYED_BYTECODE_SELECTOR.to_vec();
-            let deployed_bytecode: Vec<u8> =
-                runtime.try_call_application(false, created_application_id, argument)?;
+            let bytecode = revm_state::Bytecode::new_legacy(deployed_bytecode.clone().into());
+            context.journal().load_account(address)?;
+            context.journal().set_code(address, bytecode);
```
afck, line 1025: What does that do?

If I understand correctly, from Linera's perspective, `create_application` above created a whole new application ID with a separate state, containing the created contract's bytecode.

Are we duplicating the bytecode here, in this app's state?

MathieuDutSik: It inserts the bytecode in the database of REVM.

I think I will kill that PR because I have a new branch that does not have to deal with that and instead runs everything in REVM and implements reentrancy that we want for some EVM applications such as Morpho.

ma2bd: @MathieuDutSik I'd rather land this one first if that was a useful correction.

MathieuDutSik: So, should I merge it now?

ma2bd: Yes (after addressing @afck's comment above, if possible)

MathieuDutSik: That's kind of the problem. There is indeed duplication in the code.
That is what the new PR is supposed to address among other things.

PR #5201, opened 2026-01-05
Add option `--read-only` to the node servicePublic apis

afck:
## Motivation

We want an option to run the node service so that it can be publicly used for queries, but not for mutations.

## Proposal

Add a `--read-only` option.

(Based on #5196; I added a test with Claude's help.)

## Test Plan

A test was added.

## Release Plan

- These changes should be ported to `main`, and
- released in a new SDK.

## Links

- Fixes https://github.com/linera-io/linera-protocol/issues/5195.
- Closes https://github.com/linera-io/linera-protocol/pull/5196. (Superseded)

### DIFF
--- a/.github/workflows/docker-compose.yml
+++ b/.github/workflows/docker-compose.yml
@@ -53,7 +53,7 @@ jobs:
         # run: echo 1048576 > /proc/sys/fs/aio-max-nr
       - name: Build client binary
         run: |
-          cargo install --path linera-service --bin linera --bin linera-server --debug
+          cargo install --path linera-service --bin linera --bin linera-server --debug --locked
       # TODO(#2709): Remove this step once this workflow runs in a custom runner
       - name: Patch Docker compose file to run ScyllaDB in developer mode
         run: |
--- a/linera-service/template/Cargo.toml.template
+++ b/linera-service/template/Cargo.toml.template
@@ -5,6 +5,7 @@ edition = "2021"
 
 [dependencies]
 async-graphql = {{ version = "=7.0.17", default-features = false }}
+async-graphql-value = {{ version = "=7.0.17" }}
 {linera_sdk_dep}
 futures = {{ version = "0.3 "}}
 serde = {{ version = "1.0", features = ["derive"] }}


linera-service/src/cli/command.rs
```diff
@@ -726,6 +726,11 @@ pub enum ClientCommand {
         /// Example: `--operators my-operator=/path/to/binary`
         #[arg(long = "operators", value_parser = parse_operator)]
         operators: Vec<(String, PathBuf)>,
+
+        /// Run in public mode: disallow mutations and prevent queries from scheduling
+        /// operations. Use this when exposing the service to untrusted clients.
+        #[arg(long)]
+        public: bool,
```
deuszx: I'm not sure I agree that `public` automatically means "don't allow mutations". Maybe the reverse like _restricted_ or _untrusted_ ?

linera-service/tests/local_net_tests.rs
```diff
@@ -1387,3 +1388,44 @@ async fn test_node_service_with_task_processor() -> Result<()> {
...
+    let result = node_service
+        .transfer(chain, owner, recipient, Amount::from_tokens(1))
+        .await;
+    assert_matches!(result, Err(_));
+
+    net.ensure_is_running().await?;
+    net.terminate().await?;
+
+    Ok(())
+}
```
deuszx, line 1453: Can you add a variant here that asserts `public = false` allows for mutations?

I understand we probably have other tests that implicitly do this already but testing the feature in a small, self-contained test will make debugging easier in the future.

afck: In general I agree, but end-to-end tests are pretty slow and expensive. Do you think extending this test is better than adding a comment that points to another test where mutations are used?

linera-service/src/cli/command.rs
afck: That's true, maybe `--restricted`, `--untrusted` or `--no-mutations` would be better. What do you think, @ma2bd?

linera-service/tests/local_net_tests.rs
deuszx: Can we extend this particular test? not spin up everything from the start just shut down the first client and start a new one in the same test/network?

linera-service/src/cli/command.rs
afck: As discussed, I will rename it to `--read-only`.

ma2bd: `--read-only` ?

linera-service/tests/local_net_tests.rs
afck: Done in https://github.com/linera-io/linera-protocol/pull/5201/commits/106215fb39a175d2794ce349d9564df8f2503a9f.

PR #5013, opened 2025-11-21
Check block cache on create_cross_chain_requests

ndr-ds:
## Motivation

Under high load, create_network_actions was observed taking ~70% of CPU time. Profiling showed the bottleneck was in the call chain:

`create_network_actions`​ -> `create_cross_chain_requests`​ -> `read_certificates`​ -> `deserialize_certificate`​

Certificate deserialization alone was consuming ~60% of CPU. Since these are certificates for blocks on the current chain, many of them should already be available in the block_values cache.

## Proposal

- Check the `block_values`​ cache before reading from storage in `create_cross_chain_requests`​
- Only fetch certificates from storage for blocks not found in the cache
- Insert fetched blocks into the cache so subsequent lookups can benefit
- Work directly with `Hashed<Block>`​ instead of certificates, since only the block data is needed for constructing message bundles

This avoids repeated deserialization of certificates that are already cached.

## Test Plan

- CI passes
- Deployed to a test network and verified the optimization reduces CPU usage during sustained load

## Release Plan

- Nothing to do / These changes follow the usual release cycle.

PR #5203, opened 2026-01-05
Check block cache on create_cross_chain_requests (#5013)

ndr-ds:
## Motivation

This is a backport of #5013.

Under high load, create_network_actions was observed taking ~70% of CPU
time. Profiling showed the bottleneck was in the call chain:

`create_network_actions`​ -> `create_cross_chain_requests`​ ->
`read_certificates`​ -> `deserialize_certificate`​

Certificate deserialization alone was consuming ~60% of CPU. Since these
are certificates for blocks on the current chain, many of them should
already be available in the block_values cache.

## Proposal

- Check the `block_values`​ cache before reading from storage in
`create_cross_chain_requests`​
- Only fetch certificates from storage for blocks not found in the cache
- Insert fetched blocks into the cache so subsequent lookups can benefit
- Work directly with `Hashed<Block>`​ instead of certificates, since
only the block data is needed for constructing message bundles

This avoids repeated deserialization of certificates that are already
cached.

## Test Plan

- CI passes
- Deployed to a test network and verified the optimization reduces CPU
usage during sustained load

### DIFF
--- a/linera-core/src/chain_worker/state.rs
+++ b/linera-core/src/chain_worker/state.rs
@@ -514,26 +514,50 @@ where
                     })?,
             );
         }
-        let certificates = self.storage.read_certificates(hashes.clone()).await?;
-        let certificates = match ResultReadCertificates::new(certificates, hashes) {
-            ResultReadCertificates::Certificates(certificates) => certificates,
-            ResultReadCertificates::InvalidHashes(hashes) => {
-                return Err(WorkerError::ReadCertificatesError(hashes))
+
+        let mut uncached_hashes = Vec::new();
+        let mut height_to_blocks: HashMap<BlockHeight, Hashed<Block>> = HashMap::new();
+
+        for hash in hashes {
+            if let Some(hashed_block) = self.block_values.get(&hash) {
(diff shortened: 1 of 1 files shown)


linera-core/src/chain_worker/state.rs
```diff
@@ -514,26 +514,50 @@ where
...
-            .into_iter()
-            .zip(certificates)
-            .collect::<HashMap<_, _>>();
-        // For each medium, select the relevant messages.
+        }
+
+        if !uncached_hashes.is_empty() {
+            let certificates = self
+                .storage
+                .read_certificates(uncached_hashes.clone())
```
afck, line 532: (Maybe we should also have a `read_blocks` method: Those are stored separately from the signatures already, and `read_certificates` unnecessarily reads both.)

ndr-ds: Yeah, that's a good point

ndr-ds: Follow up: https://github.com/linera-io/linera-protocol/pull/5207

PR #5205, opened 2026-01-05
Improve `send_chain_information`

deuszx:
## Motivation

  The `send_chain_information` method in `linera-core/src/updater.rs` had grown into a complex, monolithic function that was difficult to understand and maintain. The method handled multiple distinct responsibilities (height synchronization, round synchronization, chain initialization) in a single 150+ line function with limited documentation.

  ## Proposal

  This PR refactors the `send_chain_information` method to improve code clarity and maintainability through:

  1. **Comprehensive documentation**: Added detailed doc comments explaining the two-phase synchronization strategy (height sync followed by round sync)

  2. **Method decomposition**: Split the monolithic function into four focused methods:
     - `sync_chain_height`: Synchronizes a validator to a specific block height using an optimistic approach (send last certificate first, fill gaps if needed)
     - `initialize_new_chain_on_validator`: Handles the special case of validators that don't know about a chain yet
     - `sync_consensus_round`: Synchronizes consensus round state at a given height
     - `send_chain_info_for_blobs`: Sends chain information for chains referenced by blobs (deduplicates previously repeated logic)

  3. **Improved error handling**: Made round synchronization more resilient by treating failures as non-fatal (logs debug messages instead of propagating errors), since height sync is the primary objective

  4. **Better logging**: Added more descriptive debug messages throughout the synchronization process to aid debugging

  This is a pure refactoring with no functional changes to the synchronization protocol or behavior.

  ## Test Plan

  - Existing integration tests pass unchanged, demonstrating no behavioral changes
  - Manual testing on testnet_conway to verify validator synchronization continues to work correctly

  ## Release Plan

  - Should be backported to `main`.

  ## Links

### DIFF
--- a/linera-core/src/updater.rs
+++ b/linera-core/src/updater.rs
@@ -636,18 +636,18 @@ where
     /// Sends chain information to bring a validator up to date with a specific chain.
     ///
     /// This method performs a two-phase synchronization:
-    /// 1. **Height synchronization**: Ensures the validator has all blocks up to `target_block_height`
+    /// 1. **Height synchronization**: Ensures the validator has all blocks up to `target_block_height`.
     /// 2. **Round synchronization**: If heights match, ensures the validator has proposals/certificates
-    ///    for the current consensus round
+    ///    for the current consensus round.
     ///
     /// # Height Sync Strategy
     /// - For existing chains (target_block_height > 0):
-    ///   * Optimistically sends the last certificate first (often that's all that's missing)
-    ///   * Falls back to full chain query if the validator needs more context
-    ///   * Sends any additional missing certificates in order
+    ///   * Optimistically sends the last certificate first (often that's all that's missing).
+    ///   * Falls back to full chain query if the validator needs more context.
+    ///   * Sends any additional missing certificates in order.
     /// - For new chains (target_block_height == 0):
-    ///   * Sends the chain description and dependencies first
-    ///   * Then queries the validator's state
+    ///   * Sends the chain description and dependencies first.
+    ///   * Then queries the validator's state.
     ///
     /// # Round Sync Strategy
     /// Once heights match, if the local node is at a higher round, sends the evidence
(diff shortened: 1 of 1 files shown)


linera-core/src/updater.rs
```diff
@@ -647,158 +633,296 @@ where
         .await
     }
 
+    /// Sends chain information to bring a validator up to date with a specific chain.
+    ///
+    /// This method performs a two-phase synchronization:
+    /// 1. **Height synchronization**: Ensures the validator has all blocks up to `target_block_height`
```
afck: (I'd add periods after full sentences. Also below.)

```suggestion
    /// 1. **Height synchronization**: Ensures the validator has all blocks up to `target_block_height`.
```

linera-core/src/updater.rs
```diff
@@ -647,158 +633,296 @@ where
...
+            self.send_confirmed_certificate(certificate, delivery)
+                .await?;
+        }
+
+        Ok(info)
+    }
+
+    /// Initializes a new chain on the validator by sending the chain description and dependencies.
+    ///
+    /// This is called when the validator doesn't know about the chain yet (target_block_height == 0).
```
afck: (I'd avoid referring to `target_block_height` in a comment of a function that doesn't have that parameter.)

linera-core/src/updater.rs
```diff
@@ -647,158 +633,296 @@ where
...
+    ///
+    /// Returns the ChainInfo from the validator after initialization.
```
afck:
```suggestion
    /// Returns the [`ChainInfo`] from the validator after initialization.
```

linera-core/src/updater.rs
```diff
@@ -647,158 +633,296 @@ where
...
+        for blob_state in blob_states {
+            let block_chain_id = blob_state.chain_id;
+            let block_height = blob_state.block_height.try_add_one()?;
+            chain_heights
+                .entry(block_chain_id)
+                .and_modify(|h| *h = block_height.max(*h))
+                .or_insert(block_height);
+        }
+
+        self.send_chain_info_up_to_heights(chain_heights, delivery)
```
afck, line 922: We should actually send _only_ those heights now that we have sparse chains, shouldn't we? Not the whole chains _up to_ the heights?

(Not in this PR, but maybe something we should fix soon…)

afck (approved):
Looks great! ❤️ 
No blockers.

linera-core/src/updater.rs
deuszx: Done in https://github.com/linera-io/linera-protocol/pull/5206

PR #5206, opened 2026-01-05
Send only blocks at missing heights

deuszx:
## Motivation

  When syncing chain information for blobs between validators, we were sending more data than necessary. The previous implementation would send **all blocks up to the height** where a blob was published, even when only the specific block containing that blob was needed.

  For sparse chains (chains with gaps in block heights), this meant sending many unnecessary blocks, wasting network bandwidth and validator processing time.

Addresses https://github.com/linera-io/linera-protocol/pull/5205#discussion_r2662207184

  ## Proposal

  This PR changes the blob synchronization strategy from "send all blocks up to height N" to "send only blocks at specific heights":

  1. **Modified `send_chain_info_for_blobs`**: Instead of tracking the maximum height per chain, now collects a `BTreeSet` of all specific heights where blobs exist
  2. **New method `send_chain_info_at_heights`**: Sends chain information only for the exact block heights specified, not all blocks leading up to those heights
  3. **Updated documentation**: Clarifies that this optimization specifically benefits sparse chains

  ### Example Impact

  Previously, if blobs existed at heights [5, 10, 15] on a chain:
  - Sent blocks 0-5, 0-10, 0-15 (up to 30 blocks sent, with significant duplication)

  Now:
  - Sends only blocks at heights 5, 10, 15 (exactly 3 blocks)

  ## Test Plan

  - CI

  ## Release Plan

  - Backport to `main`.

  ## Links

### DIFF
--- a/linera-core/src/updater.rs
+++ b/linera-core/src/updater.rs
@@ -911,13 +911,13 @@ where
             .read_blob_states_from_storage(blob_ids)
             .await?;
 
-        let mut chain_heights = BTreeMap::new();
+        let mut chain_heights: BTreeMap<ChainId, BTreeSet<BlockHeight>> = BTreeMap::new();
         for blob_state in blob_states {
             let block_chain_id = blob_state.chain_id;
             let block_height = blob_state.block_height;
             chain_heights
                 .entry(block_chain_id)
-                .or_insert_with(BTreeSet::new)
+                .or_default()
                 .insert(block_height);
         }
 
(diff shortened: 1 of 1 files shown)


bart-linera (approved):
LGTM :+1:

linera-core/src/updater.rs
```diff
@@ -912,17 +914,69 @@ where
         let mut chain_heights = BTreeMap::new();
         for blob_state in blob_states {
             let block_chain_id = blob_state.chain_id;
-            let block_height = blob_state.block_height.try_add_one()?;
+            let block_height = blob_state.block_height;
             chain_heights
                 .entry(block_chain_id)
-                .and_modify(|h| *h = block_height.max(*h))
-                .or_insert(block_height);
+                .or_insert_with(BTreeSet::new)
```
afck:
```suggestion
                .or_default()
```

linera-core/src/updater.rs
```diff
@@ -912,17 +914,69 @@ where
...
+            .collect();
+
+        FuturesUnordered::from_iter(tasks.into_iter().map(|(chain_id, height)| {
+            let mut updater = self.clone();
+            async move {
+                // Get the certificate at this specific height and send only that
+                let hash = updater
+                    .client
+                    .local_node
+                    .get_block_hashes(chain_id, vec![height])
```
afck: It would be much more efficient to do this only once per chain, for all heights: Each call needs exclusive access to the chain worker.

linera-core/src/updater.rs
```diff
@@ -912,17 +914,69 @@ where
...
+                    .ok_or_else(|| {
+                        ChainClientError::InternalError(
+                            "send_chain_info_at_heights called with invalid height",
+                        )
+                    })?;
+                let certificate = updater
+                    .client
+                    .local_node
+                    .storage_client()
+                    .read_certificate(hash)
```
afck: `read_certificates` for all the hashes at once might be more efficient.

linera-core/src/updater.rs
deuszx: I thought there can be only one height per chain but if we have multiple blobs, uploaded at different heights, on the same chain then that won't be true.

linera-core/src/updater.rs
deuszx: Done in [Address review comments](https://github.com/linera-io/linera-protocol/pull/5206/commits/fdb769ce3f316f95a3a07d0714a3b4b547f6e9b4)

linera-core/src/updater.rs
deuszx: Done in [Address review comments](https://github.com/linera-io/linera-protocol/pull/5206/commits/fdb769ce3f316f95a3a07d0714a3b4b547f6e9b4)

linera-core/src/updater.rs
deuszx: Done in [Address review comments](https://github.com/linera-io/linera-protocol/pull/5206/commits/fdb769ce3f316f95a3a07d0714a3b4b547f6e9b4)

PR #5214, opened 2026-01-07
Add `change_ownership` to contract runtime.

afck:
## Motivation

There are use cases where applications want to change the chain ownership.

## Proposal

Add a method for that, and an option to give applications permission to change the ownership. (Mostly Claude.)

## Test Plan

A test was added (also for `change_application_permissions`, which was missing).

## Release Plan

- Nothing to do / These changes follow the usual release cycle. (Can't easily be backported.)

## Links

- Closes #2371.
- [reviewer checklist](https://github.com/linera-io/linera-protocol/blob/main/CONTRIBUTING.md#reviewer-checklist)

PR #5215, opened 2026-01-07
[testnet] System call to change app permission

ma2bd:
## Motivation

Support "controlled" chains in the context of backend services

## Proposal

* Add the missing system API
* Re-use the existing permission `can_close_chain`

## Test Plan

CI + manual tests TBD


ma2bd:
> Looks good! We could adapt the test from #5214.

Yes a test would be nice

Issue #4128, opened 2025-06-17
Implement reentrancy and `tx.origin` for EVM.


MathieuDutSik:
Currently, the implementation of the EVM support in Linera uses REVM but in a special way:
* The running of individual smart contracts is done in separate instances. So, if one contract calls another, we use a redirection by using the precompiles.
* The state of the contract is stored with multiple keys in the storage, which are of small size (32 + 32 bytes for one key/value).
* In the way that REVM is implemented, the different smart contracts are all implemented in one single instance, which is faster.

This design has the following inconvenience:
* The cross contract calls EVM -> EVM are quite slow.
* The use of many small keys is anti-design for the key/value store.
* Reentrancy cannot be implemented. It is not only because reentrancy is not supported in Linera, but also because this would require changing the way that REVM is designed completely.
* The `tx.origin` cannot be set. Even if it is not recommended to use this pattern, it would be good to support 

There is a way to address those points:
* Run several EVM smart contracts calling each other by using REVM as designed.
* Store all K/V of an EVM instance into a single `BTreeMap<B256,B256>`. This would work even for the largest smart contracts like bridges, which have millions of K/V but not billions. If that becomes a problem, we can split it into smaller pieces.
* The `execute_operation` would be changed so that it would have 3 options:
 - Getting all keys.
 - Writing a bunch of keys.
 - Normal execution of the contract.
* So, when an external EVM smart contract is called, the database would first access the keys and then execute the contract. When execution is finished, the data will be written back to the relevant storages.
* The reentrancy would work and the `tx.origin` would work as expected.
* Wasm smart contract would not be reentrant. The reentrancy would not extend to Wasm calling EVM and then calling Wasm.

By doing this, we would address one key limitation of our design.

MathieuDutSik:
Solved by #5216

Issue #3932, opened 2025-05-13
Support different call method in EVM


MathieuDutSik:
EVM has different kinds of calls: `call`, `staticcall,` and `delegatecall`.
All should be adequately supported.

MathieuDutSik:
Solved by #5216

PR #5216, opened 2026-01-07
Address issues found by implementing Morpho

MathieuDutSik:
## Motivation

The existing implementation of EVM using the REVM does not allow for EVM reentrancy to work.
This PR addresses the issues and many other issues in the code.

## Proposal

The main idea is that if an EVM contract is calling another EVM contract then the execution is being done in the same
REVM instance. At a stroke this allows to implement different function calls, reentrancy and tx.origin.

Where things are different is for the storing of data. Each contract stores its deployed bytecode and storage on its own storage. This is done so that each contract can be called independently on its own.

So, the `DatabaseRuntime` is eliminated and becomes a more complex software. There is now a `ContractDatabase` / `ServiceDatabase`. This is because when we access to another contract, we need to do a specific `try_call_application` / `handle_query` to access the storage of that contract.

The externally owned accounts are not saved in contract storage. There is no longer any redundancy in the contract storage.

If a contract A creates some contract B and C, then during the time of the execution of this contract the contracts are generated. At the end of the execution, the contracts are written to storage:
* The `peek_application_index` returns the number of contracts created before the EVM contract gets started.
* The number of indices is stored and then when the contracts are written, they are created and need to be reordered for the `create_application` to work correctly.

The storage costs tracking has been eliminated since it does not match what is being done in EVM. For example, the following code:
```solidity
uint256 slot = 0;
for (uint i = 0; i < 100; i++) {
    storage[slot] = i;
}
storage[slot] = 0;
```

will have 99 reset costs associated to it. But in the array changes, we see only the final value. Since REVM does
not separate the storage costs from other costs, we cannot do that separation.

## Test Plan

Additional CI tests were added for the needed features encountered while implementing the morpho system.

9 tests were created for Morpho that exercise both the non-reentrant and reentrant functionality.

## Release Plan

It depends on how it is decided, but could be ported to Testnet Conway branch.

## Links

https://github.com/linera-io/linera-protocol/issues/3932
https://github.com/linera-io/linera-protocol/issues/4128

See the directory "test_evm_smart_contracts" in
https://github.com/MathieuDutSik/linera_end_to_end_tests

### DIFF
--- a/linera-execution/src/evm/database.rs
+++ b/linera-execution/src/evm/database.rs
@@ -753,11 +753,9 @@ where
             }
         }
         for entry in contracts_to_create {
-            if let Some((address, account, module_id)) = entry {
-                self.create_new_contract(address, account, module_id)?;
-            } else {
-                unreachable!();
-            };
+            let (address, account, module_id) =
+                entry.expect("An entry since all have been matched above");
+            self.create_new_contract(address, account, module_id)?;
         }
         for (address, revm_balance) in balances {
             self.check_balance(address, revm_balance)?;


linera-service/src/cli_wrappers/local_net.rs
```diff
@@ -324,7 +324,7 @@ impl Validator {
 #[cfg(with_testing)]
 impl LocalNetConfig {
     pub fn new_test(database: Database, network: Network) -> Self {
-        let num_shards = 4;
+        let num_shards = 1;
```
ma2bd: Did you mean to change this?

MathieuDutSik: No, it was used for testing, since it runs faster locally, but I forgot to revert it.
I think I will introduce a function so that we can parametrize this with environment variables.

linera-base/src/vm.rs
```diff
@@ -64,6 +65,10 @@ pub struct InvalidVmRuntime(String);
 /// The possible types of queries for an EVM contract
 #[derive(Clone, Debug, Deserialize, Serialize)]
 pub enum EvmQuery {
+    /// Account state
+    AccountInfo,
+    /// Account state
```
afck: "Storage query"?

linera-service/tests/linera_net_tests.rs
```diff
@@ -876,18 +919,22 @@ async fn test_evm_end_to_end_balance_and_transfer(config: impl LineraNetConfig)
...
     assert_contract_balance(&app_b, address1, Amount::ZERO).await?;
     assert_contract_balance(&app_b, address2, Amount::ZERO).await?;
     assert_contract_balance(&app_b, address_app, Amount::ZERO).await?;
+    tracing::info!("test_evm_end_to_end_balance_and_transfer, step 10");
 
     // Creating app_b via service calls and checking balances.
     assert_contract_balance(&app_c, address1, Amount::ZERO).await?;
     assert_contract_balance(&app_c, address2, Amount::ZERO).await?;
     assert_contract_balance(&app_c, address_app, Amount::ZERO).await?;
+    tracing::info!("test_evm_end_to_end_balance_and_transfer, step 11");
```
afck: Did you keep these logs deliberately? (If so, maybe they should state more clearly what happened in each step?)

linera-execution/src/evm/inputs.rs
```diff
@@ -264,4 +264,10 @@ mod tests {
         let selector = &keccak256("instantiate(bytes)".as_bytes())[..4];
         assert_eq!(selector, INSTANTIATE_SELECTOR);
     }
+
+    #[test]
+    fn check_set_up_part_a2() {
+        let selector = &keccak256("set_up_part_a2()".as_bytes())[..4];
+        assert_eq!(selector, &[100, 8, 176, 177]);
```
afck: What's this for? Why compare to a hard-coded selector?

linera-execution/src/evm/revm.rs
```diff
@@ -961,78 +950,37 @@ impl<Runtime: ContractRuntime> CallInterceptorContract<Runtime> {
...
         inputs: &mut CreateInputs,
     ) -> Result<Option<CreateOutcome>, ExecutionError> {
-        if !self.db.is_revm_instantiated {
-            self.db.is_revm_instantiated = true;
+        if !self.db.0.is_revm_instantiated {
+            self.db.0.is_revm_instantiated = true;
             inputs.scheme = CreateScheme::Custom {
                 address: self.contract_address,
             };
             Ok(None)
```
afck: We could return early here, instead of having a huge `else` block.

linera-execution/src/evm/database.rs
```diff
@@ -106,37 +78,31 @@ impl<Runtime> Clone for DatabaseRuntime<Runtime> {
     }
 }
 
-#[repr(u8)]
-pub enum KeyCategory {
-    AccountInfo,
-    AccountState,
-    Storage,
+/// Encodes the `index` of the EVM storage associated to the smart contract
```
afck:
```suggestion
/// Encodes the `index` of the EVM storage associated with the smart contract
```

linera-execution/src/evm/database.rs
```diff
@@ -106,37 +78,31 @@ impl<Runtime> Clone for DatabaseRuntime<Runtime> {
...
-        key
-    }
-
-    /// Returns the tag associated to the contract.
-    fn get_address_key(prefix: u8, address: Address) -> Vec<u8> {
-        let mut key = vec![prefix];
-        key.extend(address);
-        key
-    }
+/// Returns the tag associated to the contract.
```
afck:
```suggestion
/// Returns the tag associated with the contract.
```

linera-execution/src/evm/database.rs
```diff
@@ -169,56 +127,35 @@ impl<Runtime: BaseRuntime> DatabaseRuntime<Runtime> {
...
+            .unwrap_or_default();
+        if !is_newly_created {
+            account_info.balance = self.get_start_balance(address)?;
+        }
+        // We return an account as there is no difference between
+        // a default account and the absence of account.
+        Ok(Some(account_info))
+    }
+
+    /// Reading the state from local contract storage.
```
afck:
```suggestion
    /// Reads the state from local contract storage.
```

linera-execution/src/evm/database.rs
afck, line 275: (Sorry if I asked this before, but:)
Why can't this happen during a contract execution (operation/message), with a contract that has pending changes?

linera-execution/src/evm/database.rs
```diff
@@ -228,20 +165,46 @@ where
...
+        let key_info = get_address_key(KeyCategory::AccountInfo as u8, address);
+        let promise = runtime.read_value_bytes_new(key_info)?;
+        let result = runtime.read_value_bytes_wait(&promise)?;
+        match result {
+            None => Ok(None),
+            Some(bytes) => Ok(Some(bcs::from_bytes::<AccountInfo>(&bytes)?)),
+        }
+    }
+
+    /// Reading the state from local contract storage.
```
afck:
```suggestion
    /// Reads the state from local contract storage.
```

afck: This comment is the same as above. What's the difference between the two methods?

Also, please explain what `f` is used for.

linera-execution/src/evm/database.rs
```diff
@@ -228,20 +165,46 @@ where
...
+    }
+
+    /// Returns whether the address has empty storage.
+    fn has_empty_storage(&self, address: Address) -> Result<bool, ExecutionError> {
+        let application_id = address_to_user_application_id(address);
+        let mut runtime = self.runtime.lock().unwrap();
+        runtime.has_empty_storage(application_id)
+    }
+
+    /// Reading the starting balance
```
afck:
```suggestion
    /// Reads the starting balance
```

But this, too, could use a more detailed explanation.

linera-execution/src/evm/database.rs
```diff
@@ -263,143 +226,44 @@ where
...
+    pub fn get_account_info(&self) -> Result<AccountInfo, ExecutionError> {
+        let address = self.contract_address;
+        let account_info = self.account_info_from_local_storage(address)?;
+        let mut account_info = account_info.ok_or(EvmExecutionError::MissingAccountInfo)?;
+        account_info.balance = self.get_start_balance(address)?;
+        Ok(account_info)
     }
 
-    fn storage_ref(&self, address: Address, index: U256) -> Result<U256, ExecutionError> {
+    /// Reading the storage
```
afck: (Same here, and for other functions below.)

linera-execution/src/evm/database.rs
```diff
@@ -477,17 +341,36 @@ where
...
-        // We use the gas_limit from the runtime
-        let gas_limit = runtime.maximum_fuel_per_block(VmRuntime::Evm)?;
-        block_env.gas_limit = gas_limit;
-        Ok(block_env)
+        let account_info = runtime.try_call_application(false, application_id, argument)?;
+        let account_info = bcs::from_bytes(&account_info)?;
+        Ok(Some(account_info))
+    }
+
+    /// Getting the storage value of another contract.
```
afck: ~The storage value? Doesn't this _call_ the other application?~

I see, it calls "get contract storage".

linera-execution/src/evm/database.rs
```diff
@@ -512,13 +395,411 @@ where
...
+    Runtime: ContractRuntime,
+{
+    pub fn new(runtime: Runtime) -> Self {
+        Self(
+            InnerDatabase::new(runtime),
+            Arc::new(Mutex::new(HashMap::new())),
+        )
+    }
+
+    /// Effectively commits changes to storage.
```
afck: Does it?

MathieuDutSik:
> Sorry, but I'll need a lot more documentation and comments in the code to be able to review this.
> 
> Also, remember that function doc comments should start with a 3rd person singular verb (e.g. `gets`, not `getting`).
> 
> Would it make sense to split some of these changes (renaming `Ctx`) into a separate PR?

Ok, I will do my best to write the documentation in the files.

Regarding doing separate PRs, I do not think so. In a sense, this PR is hopefully the final one for EVM support in `linera-execution`. That is why I wanted to do it. The previous attempts at clearing were solving problem by problem, but still leaving the existing infrastructure, which was painful to think about. Here, we address all the problems at once and simplify the code.

linera-execution/src/evm/inputs.rs
MathieuDutSik: Sorry for this leftover. It came from my various experiments when debugging.
To eliminate.

linera-service/tests/linera_net_tests.rs
MathieuDutSik: Not deliberately, sorry.

linera-base/src/vm.rs
MathieuDutSik: Yes.

linera-execution/src/evm/revm.rs
MathieuDutSik: Ok.

linera-execution/src/evm/database.rs
MathieuDutSik: Ok.

linera-execution/src/evm/database.rs
MathieuDutSik: Ok.

linera-execution/src/evm/database.rs
MathieuDutSik: Ok.

linera-execution/src/evm/database.rs
MathieuDutSik: This is because the `changes` variable is assigned only when the changes have been committed.

That commit occurs at
```rust
    fn commit(&mut self, changes: EvmState) {
        self.0.changes = changes;
    }
```

That is called at the end of the execution.

linera-execution/src/evm/revm.rs
```diff
@@ -55,16 +56,24 @@ use crate::{
...
+
+/// The selector when writing info to storage. This is a fictional
+/// selector that does not correspond to a real EVM function.
+pub const COMMIT_CONTRACT_CHANGES_SELECTOR: &[u8] = &[5, 15, 52, 203];
+
+/// The selector when creating a contract from an already full account.
+/// This does not correspond to a real EVM function.
+pub const ALREADY_CREATED_CONTRACT_SELECTOR: &[u8] = &[23, 47, 106, 235];
 
 /// The json serialization of a trivial vector.
```
afck:
```suggestion
/// The JSON serialization of a trivial vector.
```

linera-execution/src/evm/revm.rs
```diff
@@ -491,10 +475,10 @@ impl<'a, Runtime: ContractRuntime> PrecompileProvider<Ctx<'a, Runtime>> for Cont
 }
 
 fn get_evm_destination<Runtime: ContractRuntime>(
-    context: &mut Ctx<'_, Runtime>,
+    context: &mut ContractCtx<'_, Runtime>,
     account: Account,
 ) -> Result<Option<Address>, ExecutionError> {
-    let mut runtime = context.db().0.runtime.lock().unwrap();
+    let mut runtime = context.db().0 .0.runtime.lock().unwrap();
```
afck: This is repeated a lot. Should we have a method for this?


```suggestion
    let mut runtime = context.db().0.lock_runtime();
```

Or alternatively: Would it make sense to move more of this logic into `ContractDatabase`? E.g. forward the whole `contract_runtime_call` below with the `ContractRuntimePrecompile` argument into a new `ContractDatabase` method, which does the `match` internally?

(Not a blocker, just wondering if this could be simplified.)

linera-execution/src/evm/revm.rs
```diff
@@ -814,39 +801,52 @@ impl<Runtime> Clone for CallInterceptorContract<Runtime> {
...
     }
 }
 
 impl<Runtime: ContractRuntime> CallInterceptorContract<Runtime> {
-    /// Gets the expected `ApplicationId`. We need to transfer
-    /// native tokens before the application is created (see below).
-    /// Therefore, we need to pre-compute the obtained application ID.
+    /// Gets the expected `ApplicationId` corresponding to the
+    /// EVM contract being created.
+    /// module_id is the module being created.
```
afck:
```suggestion
    /// `module_id` is the module being created.
```

linera-execution/src/evm/revm.rs
```diff
@@ -814,39 +801,52 @@ impl<Runtime> Clone for CallInterceptorContract<Runtime> {
...
+    ///
+    /// The index num_apps is the index of the application being
```
afck:
```suggestion
    /// The index `num_apps` is the index of the application being
```

linera-execution/src/evm/revm.rs
```diff
@@ -814,39 +801,52 @@ impl<Runtime> Clone for CallInterceptorContract<Runtime> {
...
+    /// created. This is needed because the calls to `create_applications`
+    /// are done at the end of the execution of the contract.
+    /// So, the peek_application_index always returns the same index
```
afck:
```suggestion
    /// So, `peek_application_index` always returns the same index
```

linera-execution/src/evm/revm.rs
```diff
@@ -814,39 +801,52 @@ impl<Runtime> Clone for CallInterceptorContract<Runtime> {
...
+    /// during the execution.
+    ///
+    /// The parameters are empty because there is no argument
+    /// to the creation of the contract. In fact the .init_code
```
afck:
```suggestion
    /// to the creation of the contract. In fact the `.init_code`
```

linera-execution/src/evm/revm.rs
```diff
@@ -860,14 +860,15 @@ impl<Runtime: ContractRuntime> CallInterceptorContract<Runtime> {
         Ok(ApplicationId::from(&application_description))
     }
 
-    /// Publishes the `inputs`.
+    /// Publishes the init_code as a `ModuleId`.
```
afck:
```suggestion
    /// Publishes `inputs.init_code` as a `ModuleId`.
```

linera-execution/src/evm/revm.rs
afck:
```suggestion
    /// The function `create` of the `Inspector` trait is called
```

afck:
```suggestion
    /// * It can return some specific `CreateInputs` to be used.
```

(I don't think the type `CreateInput`—singular—exists?)

afck:
```suggestion
    /// * `inputs.init_code` is formed by the concatenation
```

afck, line 978:
```suggestion
   ///
   ///   ```
   ///   constructor(uint a, uint b) {
   ///       value = a + b
   ///   }
   ///   ```
   ///
   ///   or
   ///
   ///   ```
   ///   constructor(uint b) {
   ///       value = 3 + b
   ///   }
   ///   ```
   ///
```

afck:
```suggestion
    ///   This is stored in the `AccountInfo` entry. It is
```

afck:
```suggestion
    ///   have already put in the `init_code`.
```

linera-execution/src/evm/revm.rs
```diff
@@ -938,102 +939,56 @@ impl<Runtime: ContractRuntime> CallInterceptorContract<Runtime> {
...
-    ///   a new contract and so a call to `fn create_or_fail` in
-    ///   another instance of Revm.
-    /// * When returning the `CreateOutcome`, we need to have the
-    ///   deployed bytecode. This is implemented through a special
-    ///   call to `GET_DEPLOYED_BYTECODE_SELECTOR`. This is done
-    ///   with an `execute_operation`.
-    /// * Data is put together as a `Some(...)` which tells Revm
-    ///   that it does not need to execute the bytecode since the
-    ///   output is given to it.
+    /// * The application_id is being computed and used to adjust
```
afck:
```suggestion
    /// * The application ID is being computed and used to adjust
```

linera-execution/src/evm/revm.rs
```diff
@@ -938,102 +939,56 @@ impl<Runtime: ContractRuntime> CallInterceptorContract<Runtime> {
...
+    ///   the contract creation.
+    /// * The balance is adjusted so that when it gets created in
+    ///   Linera it will have the correct balance.
+    /// * But that is the only change being done. We return Ok(None)
```
afck:
```suggestion
    /// * But that is the only change being done. We return `Ok(None)`
```

linera-execution/src/evm/revm.rs
```diff
@@ -938,102 +939,56 @@ impl<Runtime: ContractRuntime> CallInterceptorContract<Runtime> {
...
+    ///   which means that the contract creation is done as usual
+    ///   with REVM.
```
afck:
```suggestion
    ///   with Revm.
```

linera-execution/src/evm/revm.rs
```diff
@@ -938,102 +939,56 @@ impl<Runtime: ContractRuntime> CallInterceptorContract<Runtime> {
...
-    ///   `deposit_funds` because we would not have the right
-    ///   to make the transfer.
-    /// * Therefore, we have to do the transfer before
-    ///   `create_application`. This forces us to know the
-    ///   `application_id` before it is created. This is done
-    ///   by building the `ApplicationDescription` and its hash.
+    /// The instantiation in Linera of the separate contracts
+    /// is done at the end of the contract execution. It goes
+    /// in the following way.
+    /// * The HashMap<ApplicationId, (ModuleId, u32)> contains
```
afck:
```suggestion
    /// * The `HashMap<ApplicationId, (ModuleId, u32)>` contains
```

linera-execution/src/evm/revm.rs
```diff
@@ -938,102 +939,56 @@ impl<Runtime: ContractRuntime> CallInterceptorContract<Runtime> {
...
+    ///   the list of application to create and their index.
+    /// * The index and ModuleId allow the `create_application`
```
afck:
```suggestion
    /// * The index and module ID allow `create_application`
```

linera-execution/src/evm/revm.rs
```diff
@@ -938,102 +939,56 @@ impl<Runtime: ContractRuntime> CallInterceptorContract<Runtime> {
...
+    ///   to create a contract with the right application_id.
```
afck:
```suggestion
    ///   to create a contract with the right application ID.
```

linera-execution/src/evm/revm.rs
```diff
@@ -938,102 +939,56 @@ impl<Runtime: ContractRuntime> CallInterceptorContract<Runtime> {
...
+    /// * The `create_application` is in that case just a
```
afck:
```suggestion
    /// * `create_application` is in that case just a
```

linera-execution/src/evm/database.rs
```diff
@@ -512,13 +430,444 @@ where
...
+
+#[repr(u8)]
+pub enum KeyCategory {
+    AccountInfo,
+    AccountState,
+    Storage,
+}
+
+// The Database for contracts
+
```
afck:
```suggestion
/// The Database for contracts
```

linera-execution/src/evm/database.rs
```diff
@@ -512,13 +430,444 @@ where
...
+pub(crate) struct ContractDatabase<Runtime>(
+    pub InnerDatabase<Runtime>,
+    pub Arc<Mutex<HashMap<ApplicationId, (ModuleId, u32)>>>,
```
afck: We access these fields a lot below. It would be more readable if they were named, e.g. `db` and `modules` or `inner` and `apps`, …?

linera-execution/src/evm/revm.rs
afck: Is that valid Markdown for lists?

afck:
```suggestion
    /// to produce `Some(_)` as output. Note that in the EVM
```

linera-execution/src/evm/revm.rs
```diff
@@ -1310,9 +1158,19 @@ where
     Runtime: ContractRuntime,
 {
     fn instantiate(&mut self, argument: Vec<u8>) -> Result<(), ExecutionError> {
-        self.db.set_contract_address()?;
+        self.db.0.set_contract_address()?;
         let caller = self.get_msg_address()?;
         let instantiation_argument = serde_json::from_slice::<EvmInstantiation>(&argument)?;
+        // This is the case of a contract created by REVM by another contract. We only
```
afck:
```suggestion
        // This is the case of a contract created by Revm by another contract. We only
```

linera-execution/src/evm/revm.rs
```diff
@@ -1322,32 +1180,53 @@ where
         Ok(())
     }
 
+    /// Execute the operation.
```
afck:
```suggestion
    /// Executes the operation.
    ///
```

linera-execution/src/evm/revm.rs
```diff
@@ -1322,32 +1180,53 @@ where
...
+    /// The first 3 possibilities are internal calls
+    /// from another REVM instance:
```
afck:
```suggestion
    /// from another Revm instance:
```

linera-execution/src/evm/revm.rs
```diff
@@ -1322,32 +1180,53 @@ where
...
+    /// * The `GET_ACCOUNT_INFO_SELECTOR` retrieves the
+    ///   AccountInfo of that EVM contract.
```
afck:
```suggestion
    ///   `AccountInfo` of that EVM contract.
```

linera-execution/src/evm/revm.rs
```diff
@@ -1322,32 +1180,53 @@ where
...
+    /// * The `GET_CONTRACT_STORAGE_SELECTOR` is about
+    ///   individual storage entries.
+    /// * The `COMMIT_CONTRACT_CHANGES_SELECTOR` is about
+    ///   committing the state
+    ///
+    /// If not in those cases, then the execution proceeds
+    /// normally and creates an REVM instance.
```
afck:
```suggestion
    /// normally and creates an Revm instance.
```

linera-execution/src/evm/revm.rs
MathieuDutSik: Thanks.

linera-execution/src/evm/revm.rs
MathieuDutSik: Ok.

linera-execution/src/evm/revm.rs
MathieuDutSik: Ok.

linera-execution/src/evm/revm.rs
MathieuDutSik: Ok.

linera-execution/src/evm/revm.rs
MathieuDutSik: ok.

linera-execution/src/evm/revm.rs
MathieuDutSik: Ok.

linera-execution/src/evm/revm.rs
MathieuDutSik: Ok.

linera-execution/src/evm/revm.rs
MathieuDutSik: Ok.

linera-execution/src/evm/revm.rs
MathieuDutSik: Ok.

linera-execution/src/evm/revm.rs
MathieuDutSik: Ok.

linera-execution/src/evm/revm.rs
MathieuDutSik: Ok.

linera-execution/src/evm/database.rs
MathieuDutSik: Ok.

linera-execution/src/evm/revm.rs
MathieuDutSik: Ok.

linera-execution/src/evm/revm.rs
MathieuDutSik: Ok.

linera-execution/src/evm/revm.rs
MathieuDutSik: Ok.

linera-execution/src/evm/revm.rs
MathieuDutSik: Ok.

linera-execution/src/evm/revm.rs
MathieuDutSik: Ok.

linera-execution/src/evm/revm.rs
MathieuDutSik: Ok.

linera-execution/src/evm/revm.rs
MathieuDutSik: Ok.

linera-execution/src/evm/revm.rs
MathieuDutSik: Ok.

linera-execution/src/evm/database.rs
MathieuDutSik: Ok, done. Changed to `inner` / `modules`.
The `db` is a problem because the entry `db` already occurs for other types and so this is confusing.

linera-execution/src/evm/revm.rs
MathieuDutSik: Ok, I changed it to standard comment.

MathieuDutSik: Ok.

MathieuDutSik: Ok

MathieuDutSik: Ok.

MathieuDutSik: Ok.

MathieuDutSik: The type is plural.

MathieuDutSik: Ok.

linera-execution/src/evm/database.rs
```diff
@@ -169,56 +128,66 @@ impl<Runtime: BaseRuntime> DatabaseRuntime<Runtime> {
...
-impl<Runtime> DatabaseRef for DatabaseRuntime<Runtime>
-where
-    Runtime: BaseRuntime,
-{
-    type Error = ExecutionError;
-
-    /// The `basic_ref` is the function for reading the state of the application.
-    fn basic_ref(&self, address: Address) -> Result<Option<AccountInfo>, ExecutionError> {
+    /// Reads the account info from the storage for a contract A
+    /// of address contract_address.
```
afck: I was confused at first what this was referring to. Maybe:

```suggestion
    /// at address `Self::contract_address`.
```

linera-execution/src/evm/database.rs
```diff
@@ -169,56 +128,66 @@ impl<Runtime: BaseRuntime> DatabaseRuntime<Runtime> {
...
+    /// * The function `f` is accessing the state for an account
+    ///   whose address is different from contract_address (e.g.
```
afck:
```suggestion
    ///   whose address is different from `contract_address` (e.g.
```

linera-execution/src/evm/database.rs
```diff
@@ -169,56 +128,66 @@ impl<Runtime: BaseRuntime> DatabaseRuntime<Runtime> {
...
+    ///   a contract created in the contract A or accessed by A,
+    ///   e.g. ERC20).
+    ///   For `ContractRuntime` and `ServiceRuntime` the method
+    ///   to access is different. So, the function has to be
+    ///   provided as argument.
+    /// * The address in question being read.
```
afck:
```suggestion
    /// * `address`: The address being read.
```

linera-execution/src/evm/database.rs
```diff
@@ -169,56 +128,66 @@ impl<Runtime: BaseRuntime> DatabaseRuntime<Runtime> {
...
+    /// * Whether the contract is newly created or not. For newly
```
afck:
```suggestion
    /// * `is_newly_created`: Whether the contract is newly created or not. For newly
```

linera-execution/src/evm/database.rs
```diff
@@ -169,56 +128,66 @@ impl<Runtime: BaseRuntime> DatabaseRuntime<Runtime> {
...
+    ///   created contract, the balance is the one of REVM. For
```
afck:
```suggestion
    ///   created contract, the balance is the one of Revm. For
```

linera-execution/src/evm/database.rs
```diff
@@ -169,56 +128,66 @@ impl<Runtime: BaseRuntime> DatabaseRuntime<Runtime> {
...
+        let mut runtime = self.runtime.lock().unwrap();
+        let key_info = get_address_key(KeyCategory::AccountInfo);
+        let promise = runtime.read_value_bytes_new(key_info)?;
+        let result = runtime.read_value_bytes_wait(&promise)?;
+        Ok(from_bytes_option::<AccountInfo>(&result)?)
+    }
+
+    /// Reads the state from the inner database.
+    ///
+    /// If changes is not empty, then it means that
```
afck:
```suggestion
    /// If `changes` is not empty, then it means that
```

linera-execution/src/evm/database.rs
```diff
@@ -169,56 +128,66 @@ impl<Runtime: BaseRuntime> DatabaseRuntime<Runtime> {
...
+    /// the contract has been instantiated for a
+    /// service query. In that case we do not have
+    /// any storage possible to access, just changes.
+    ///
+    /// In the other case, we access the storage directly
```
afck:
```suggestion
    /// In the other case, we access the storage directly.
```

linera-execution/src/evm/database.rs
```diff
@@ -228,20 +197,50 @@ where
...
+            Ok(None)
+        } else {
+            // This matches other EVM contracts.
+            f(self, address)
+        }
+    }
+
+    /// Returns whether the address has empty storage.
+    /// An address has an empty storage if and only if it is
+    /// an externally owned account(EOA).
```
afck:
```suggestion
    /// an externally owned account (EOA).
```

linera-execution/src/evm/database.rs
afck:
```suggestion
        //   Another in the transfer by Revm.
```

linera-execution/src/evm/database.rs
```diff
@@ -263,143 +262,52 @@ where
...
-        account_info.balance = start_balance;
-        // We return an account as there is no difference between
-        // a default account and the absence of account.
-        Ok(Some(account_info))
+        })
     }
 
-    fn code_by_hash_ref(&self, _code_hash: B256) -> Result<Bytecode, ExecutionError> {
-        panic!("Functionality code_by_hash_ref not implemented");
+    pub fn get_account_info(&self) -> Result<AccountInfo, ExecutionError> {
```
afck, line 357: This sets the balance to the start balance… so this is meant to be called only during initialization, not at a later point while handling a transaction, is it?

linera-execution/src/evm/database.rs
```diff
@@ -477,17 +384,36 @@ where
...
 }
 
-impl<Runtime> DatabaseRuntime<Runtime>
+impl<Runtime> InnerDatabase<Runtime>
 where
     Runtime: ContractRuntime,
 {
-    pub fn get_contract_block_env(&self) -> Result<BlockEnv, ExecutionError> {
-        let mut block_env = self.get_block_env()?;
+    /// Getting the smart contract code if existing.
```
afck:
```suggestion
    /// Gets the smart contract code if existing.
```

linera-execution/src/evm/database.rs
```diff
@@ -477,17 +384,36 @@ where
...
-        // We use the gas_limit from the runtime
-        let gas_limit = runtime.maximum_fuel_per_block(VmRuntime::Evm)?;
-        block_env.gas_limit = gas_limit;
-        Ok(block_env)
+        let account_info = runtime.try_call_application(false, application_id, argument)?;
+        let account_info = bcs::from_bytes(&account_info)?;
+        Ok(Some(account_info))
+    }
+
+    /// Getting the storage value of another contract.
```
afck:
```suggestion
    /// Gets the storage value of another contract.
```

linera-execution/src/evm/database.rs
```diff
@@ -512,13 +438,456 @@ where
...
+        let key_prefix = get_address_key(KeyCategory::Storage);
+        let key_info = get_address_key(KeyCategory::AccountInfo);
+        let key_state = get_address_key(KeyCategory::AccountState);
+        if account.is_selfdestructed() {
+            batch.delete_key_prefix(key_prefix);
+            batch.put_key_value(key_info, &AccountInfo::default())?;
+            batch.put_key_value(key_state, &AccountState::NotExisting)?;
+        } else {
+            let is_newly_created = account.is_created();
+            // We write here the state of the user in question. But that does not matter
```
afck: I don't understand…

linera-execution/src/evm/database.rs
```diff
@@ -512,13 +438,456 @@ where
...
+                    AccountState::StorageCleared
+                } else {
+                    AccountState::Touched
+                }
+            };
+            batch.put_key_value(key_state, &account_state)?;
+            for (index, value) in &account.storage {
+                if value.present_value() != value.original_value() {
+                    let key = get_storage_key(*index);
+                    if value.original_value() == U256::ZERO {
```
afck: Is this case necessary? Isn't the only thing that matters whether `present_value()` is zero?

linera-execution/src/evm/database.rs
```diff
@@ -512,13 +438,456 @@ where
...
+            return false;
+        }
+        if !account.is_touched() {
+            // Not modified accounts do not need to be written down.
+            return false;
+        }
+        let code_hash = account.info.code_hash;
+        // User accounts are not written. This is fine since the balance
+        // is accessed from Linera and the nonce are not accessible in
+        // solidity smart contracts.
```
afck:
```suggestion
        // Solidity smart contracts.
```

afck: Or should it rather be:

```suggestion
        // EVM smart contracts.
```

(Because it doesn't matter what it was compiled _from_?)

linera-execution/src/evm/database.rs
```diff
@@ -512,13 +438,456 @@ where
...
+        let code_empty = code_hash == KECCAK_EMPTY || code_hash.is_zero();
+        !code_empty
+    }
+
+    /// Creates a new contract. The account contains
```
afck:
```suggestion
    /// Creates a new contract. The `account` contains
```

linera-execution/src/evm/database.rs
```diff
@@ -512,13 +438,456 @@ where
...
+        let mut argument = COMMIT_CONTRACT_CHANGES_SELECTOR.to_vec();
+        argument.extend(bcs::to_bytes(&account)?);
+        let mut runtime = self.inner.runtime.lock().unwrap();
+        runtime.try_call_application(false, application_id, argument)?;
+        Ok(())
+    }
+
+    /// Effectively commits changes to storage.
+    /// This is done in the following way:
+    /// * Identify the balances that needs to be checked
```
afck:
```suggestion
    /// * Identify the balances that need to be checked
```

linera-execution/src/evm/database.rs
```diff
@@ -512,13 +438,456 @@ where
...
+    /// * Write down the state of the contract for contract_address locally.
```
afck:
```suggestion
    /// * Write down the state of the contract for `contract_address` locally.
```

linera-execution/src/evm/database.rs
```diff
@@ -512,13 +438,456 @@ where
...
+    /// * For the other contracts, if it already created, commit it.
+    ///
+    /// If not insert them into the map.
+    /// * Iterates over the entry of the map and creates the contract in the
```
afck:
```suggestion
    /// * Iterates over the entries of the map and creates the contracts in the
```

linera-execution/src/evm/database.rs
```diff
@@ -512,13 +438,456 @@ where
...
+                    }
+                }
+                balances.push((address, revm_balance));
+            }
+        }
+        for (address, account, module_id) in map_creation.into_values() {
+            self.create_new_contract(address, account, module_id)?;
+        }
+        for (address, revm_balance) in balances {
+            self.check_balance(address, revm_balance)?;
```
afck, line 761: (This is a sanity check that should never fail, is it? Would it fail for the non-writable accounts?)

linera-execution/src/evm/database.rs
```diff
@@ -512,13 +438,456 @@ where
...
+            map.contains_key(&application_id)
+        };
+        self.inner.read_basic_ref(
+            InnerDatabase::<Runtime>::get_contract_account_info,
+            address,
+            is_newly_created,
+        )
+    }
+
+    /// There are two ways to implements the trait:
```
afck:
```suggestion
    /// There are two ways to implement the trait:
```

linera-execution/src/evm/database.rs
```diff
@@ -512,13 +438,456 @@ where
...
+    /// * Returns entries with "code: Some(...)"
+    /// * Returns entries with "code: None".
+    ///
+    /// Since we choose the first design, the "code_by_hash_ref" is not needed. There
```
afck:
```suggestion
    /// Since we choose the first design, `code_by_hash_ref` is not needed. There
```

linera-execution/src/evm/database.rs
```diff
@@ -512,13 +438,456 @@ where
...
+    /// is an example in revm source code of this kind.
```
afck:
```suggestion
    /// is an example in the Revm source code of this kind.
```

(Should we link to that example?)

linera-execution/src/evm/database.rs
```diff
@@ -512,13 +438,456 @@ where
...
+    fn code_by_hash_ref(&self, _code_hash: B256) -> Result<Bytecode, ExecutionError> {
+        panic!("Returned AccountInfo should have code: Some(...) and so code_by_hash_ref should never be called");
+    }
+
+    /// Access the storage by the relevant remote access function.
```
afck:
```suggestion
    /// Accesses the storage by the relevant remote access function.
```

linera-execution/src/evm/database.rs
```diff
@@ -512,13 +438,456 @@ where
...
+            address,
+            is_newly_created,
+        )
+    }
+
+    /// There are two ways to implements the trait:
+    /// * Returns entries with "code: Some(...)"
+    /// * Returns entries with "code: None".
+    ///
+    /// Since we choose the first design, the "code_by_hash_ref" is not needed. There
```
afck:
```suggestion
    /// Since we choose the first design, `code_by_hash_ref` is not needed. There
```

linera-execution/src/evm/database.rs
```diff
@@ -512,13 +438,456 @@ where
...
+    /// is an example in revm source code of this kind.
```
afck:
```suggestion
    /// is an example in Revm source code of this kind.
```

linera-execution/src/evm/revm.rs
```diff
@@ -1390,9 +1282,10 @@ fn process_execution_result(
             logs,
             output,
         } => {
-            let mut gas_final = gas_used;
-            gas_final -= storage_stats.storage_costs();
-            assert_eq!(gas_refunded, storage_stats.storage_refund());
+            // Apply EIP-3529 refund cap (London fork)
+            let max_refund = gas_used / 5;
+            let actual_refund = gas_refunded.min(max_refund);
+            let gas_final = gas_used - actual_refund;
```
afck, line 1348: Do we need to apply this in Linera?

Anyway, it looks like this computation can move to the `else` branch.

linera-execution/src/evm/revm.rs
MathieuDutSik: I introduced a `.lock_runtime()` function.

linera-execution/src/evm/database.rs
MathieuDutSik: Yes, and this was actually not clear. I changed the explanation.

linera-execution/src/evm/revm.rs
MathieuDutSik: It is the standard way to do in the EVM. So, I would think it is a nice thing to do as well in Linera.

linera-execution/src/evm/database.rs
MathieuDutSik: Ok.

linera-execution/src/evm/database.rs
MathieuDutSik: Ok.

linera-execution/src/evm/database.rs
MathieuDutSik: Ok.

linera-execution/src/evm/database.rs
MathieuDutSik: Exactly.

linera-execution/src/evm/database.rs
MathieuDutSik: Ok.

linera-execution/src/evm/database.rs
MathieuDutSik: Yes, source code origin does not matter.

linera-execution/src/evm/database.rs
MathieuDutSik: Ok.

linera-execution/src/evm/database.rs
MathieuDutSik: Ok.

linera-execution/src/evm/database.rs
MathieuDutSik: Ok.

linera-execution/src/evm/database.rs
MathieuDutSik: Ok.

linera-execution/src/evm/database.rs
MathieuDutSik: Ok.

linera-execution/src/evm/database.rs
MathieuDutSik: Ok.

linera-execution/src/evm/database.rs
MathieuDutSik: Ok.

linera-execution/src/evm/database.rs
MathieuDutSik: Ok.

linera-execution/src/evm/database.rs
MathieuDutSik: Ok.

linera-execution/src/evm/database.rs
MathieuDutSik: Ok.

linera-execution/src/evm/database.rs
MathieuDutSik: Ok.

linera-execution/src/evm/database.rs
MathieuDutSik: Ok.

linera-execution/src/evm/database.rs
MathieuDutSik: Ok.

linera-execution/src/evm/database.rs
MathieuDutSik: Ok.

linera-execution/src/evm/database.rs
MathieuDutSik: Ok.

linera-execution/src/evm/database.rs
MathieuDutSik: Ok.

linera-execution/src/evm/database.rs
MathieuDutSik: Yes, it should never fail.
But yes, we can extend the check to other accounts.

MathieuDutSik: Added.

linera-execution/src/evm/database.rs
MathieuDutSik: Yes, I have to add this feature. I forgot about that.

linera-execution/src/evm/database.rs
MathieuDutSik: You are completely correct.

linera-execution/src/evm/database.rs
```diff
@@ -45,25 +45,65 @@ pub const EVM_SERVICE_GAS_LIMIT: u64 = 20_000_000;
...
+    ///
+    /// Initially set to `Address::ZERO` and updated to the actual contract address
+    /// derived from the Linera `ApplicationId` once the runtime becomes available.
     pub contract_address: Address,
-    /// The caller to the smart contract.
+
+    /// The address of the entity calling this smart contract.
+    ///
+    /// Corresponds to `msg.sender` in Solidity. May be an EOA address, another
+    /// contract address, or `Address::ZERO` for non-EVM callers.
```
afck: Is it always zero if not EVM? What if the caller is a Wasm contract?

linera-execution/src/evm/database.rs
```diff
@@ -45,25 +45,65 @@ pub const EVM_SERVICE_GAS_LIMIT: u64 = 20_000_000;
...
+    /// Wrapped in `Arc<Mutex<>>` to allow shared access across database clones
+    /// while maintaining safe mutation.
     pub runtime: Arc<Mutex<Runtime>>,
-    /// The uncommitted changes to the contract. It is used only for service calls.
+
+    /// Uncommitted state changes from EVM execution.
+    ///
+    /// For contract operations, accumulated during execution and committed at the end.
+    /// For service queries on uninitialized contracts, holds the temporary state
+    /// since persistent storage is not yet available.
```
afck: Or "is not available" (no "yet")?

linera-execution/src/evm/database.rs
```diff
@@ -89,20 +129,22 @@ fn get_storage_key(index: U256) -> Vec<u8> {
...
 where
     Runtime: BaseRuntime,
 {
-    /// Creates a new `DatabaseRuntime`.
+    /// Creates a new database instance with default values.
+    ///
+    /// The contract address is initially set to `Address::ZERO` and must be updated
+    /// later via `set_contract_address()` once the runtime can be safely accessed.
+    /// This deferred initialization is necessary because locking the runtime during
+    /// construction could cause issues.
```
afck: (What issues? Is there a more concrete reason?)

linera-execution/src/evm/database.rs
```diff
@@ -134,19 +197,30 @@ where
...
+    ///   one of Revm. For existing contract, the balance has to
+    ///   be accessed from Linera.
+    ///
+    /// For Externally Owned Accounts, the function is indeed
+    /// called, but it does not access the storage. Instead it
+    /// creates a default `AccountInfo` whose balance is computed
+    /// from the one in Linera. This is the case both for the faucet
+    /// and for other accounts.
+    ///
+    /// For the contract for which address == contract_address we
```
afck:
```suggestion
    /// For the contract for which `address == contract_address` we
```

linera-execution/src/evm/database.rs
```diff
@@ -628,36 +739,42 @@ where
...
-                balances.push((address, revm_balance));
             }
         }
-        for (address, account, module_id) in map_creation.into_values() {
-            self.create_new_contract(address, account, module_id)?;
+        for entry in contracts_to_create {
+            if let Some((address, account, module_id)) = entry {
+                self.create_new_contract(address, account, module_id)?;
+            } else {
+                unreachable!();
```
afck: Is it? We initialized this with all `None`s and then only set it to `Some` under certain conditions…

linera-execution/src/evm/database.rs
MathieuDutSik: It is a fact that in EVM you have access to the user calling you via `msg.sender`.

In Linera, as specified by ContractRuntime (Wasm is not essential to the story) a call can be not-authenticated. In that case, we have that `authenticated_caller_id` / `authenticated_owner` are both `None`.

There are two ways we could handle that:
* One is to forbid this and return an error.
* Another is to pass the `Address::ZERO`.

We could of course have chose the first option. But it forbids some possible use case. If the solidity programmer wants to avoid that, it can do it very well by writing
```solidity
require(msg.sender != address(0), "Only authenticated calls")
```

So, I think this is the right choice here.

afck: Sure, makes sense to allow this.

But I'd imagine that _if_ the caller makes an authenticated call, then we'd set `msg.sender` to the (converted) value of `authenticated_caller_id`, even if the caller is Wasm? (If we do that, we should update this comment.)

linera-execution/src/evm/database.rs
MathieuDutSik: Yes, "not available" is more accurate. Though the contract could be intialized later by a contract call.

linera-execution/src/evm/database.rs
MathieuDutSik: Ok.
Right now when InnerDatabase::new is called, it is in the following use case:
```rust
    pub fn prepare(module: Vec<u8>, runtime: Runtime) -> Self {
        let db = ContractDatabase::new(runtime);
        Self { module, db }
    }
```

we cannot call the `runtime.application_id` in that context. We get a locking crash. There could be a correction to that, but it would require changing the trait `UserContractModule`, which maybe I should do, but in another PR.

Now, of course, there could be contextes where the `RevmContractInstance::prepare` where we can do this, but the only case where we call this function we have the locking problem.

linera-execution/src/evm/database.rs
MathieuDutSik: Ok.

linera-execution/src/evm/database.rs
MathieuDutSik: Ok.
We have that `modules` is a map from `ApplicationId` to `(ModuleId, u32)` with the index being the position in the list of contracts being created. 

If we create N applications, then the map has size N. Otherwise, we would have hash collision and we do not think we have this. The indices being created are between 0 and N-1 (inclusive).

Every single one of the applications that are registered in the modules map, is also created by Revm and enters into the "changes: EvmState" entry. Of course, execution could fail prematurely for one reason of another. But then the whole execution fails, not just one entry is missing.

So, when in the loop
```rust
for (address, account) in changes
```

those created contracts do show up in the `account`. They have to pass two tests:
* One pass the `is_account_writable`. But yes it does because the contract address is not the faucet, the account is for sure touched since it is new and the code is non-empty.
* Another pass the `address == self.inner.contract_address`. But yes, it does since the newly created contract has an address different from `contract_address`. If not, we would have, again, a hash collision.

So, yes at the end of the loop, all the `contracts_to_create` entries are `Some(...)`. Moreover, they are in the right order, and so we can officialise it in Linera.

linera-execution/src/evm/database.rs
MathieuDutSik: Ok, I updated the contract.

linera-execution/src/evm/revm.rs
ma2bd: This feature is interesting but it is very weird to bury this kind of policies inside the `revm` module disregarding what the rest of Linera is doing.

linera-execution/src/evm/database.rs
ma2bd: nit: this should be using `.expect("short sentence")` instead of `if let Some(..) =`

linera-execution/src/evm/revm.rs
MathieuDutSik: With regard to "what the rest of Linera is doing", I should note that we still have issue #1536 about refunding storage. So, I am not sure twhat we should compare with.

Second, the gas usage is well understood by EVM programmers so having the same scheme is maybe a good idea.

linera-execution/src/evm/database.rs
MathieuDutSik: Ok, done.

afck (approved):

PR #5217, opened 2026-01-07
Add `change_ownership` to contract runtime; unify permissions.

### LINKED ISSUE #2371: Allow applications to change application permissions and ownership.
Add setters and getters for the current chain's application permissions and ownership to the contract runtime.

To keep the permissions config simple, let's replace `close_chain` with `config_chain`: These applications will be able to close the chain as well as change the chain config.

afck:
## Motivation

There are use cases where applications want to change the chain ownership.

## Proposal

Add a method for that. Unify permissions to close the chain, change ownership and change app permissions. (Mostly Claude.)

## Test Plan

A test was added.

## Release Plan

- Nothing to do / These changes follow the usual release cycle. (Can't easily be backported.)

## Links

- Closes #2371.
- Based on #5215, but with unified permission setting.

### DIFF
--- a/examples/amm/README.md
+++ b/examples/amm/README.md
@@ -259,7 +259,7 @@ linera --wait-for-outgoing-messages change-ownership \
 linera --wait-for-outgoing-messages change-application-permissions \
     --execute-operations $AMM_APPLICATION_ID \
     --mandatory-applications $AMM_APPLICATION_ID \
-    --close-chain $AMM_APPLICATION_ID
+    --manage-chain $AMM_APPLICATION_ID
 
 linera service --port $PORT &
 ```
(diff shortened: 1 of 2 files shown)


linera-base/src/ownership.rs
```diff
@@ -243,6 +243,14 @@ pub enum CloseChainError {
     NotPermitted,
 }
 
+/// Errors that can happen when attempting to change the chain ownership.
+#[derive(Clone, Copy, Debug, Error, WitStore, WitType)]
+pub enum ChangeOwnershipError {
```
ma2bd: I wonder if we should unify the errors as well (or not)

ma2bd (approved):
nice!

linera-base/src/ownership.rs
afck: Done in https://github.com/linera-io/linera-protocol/pull/5217/commits/522a9bddcaa05905f311757d14ac6985be88fdb9.

PR #5223, opened 2026-01-08
Treat CLI params as deltas instead of the desired value

bart-linera:
## Motivation

Currently, changing the application permissions from the CLI requires specifying the whole final state of the permissions. So changing a single field requires specifying all of them, otherwise they will be replaced by default values.

## Proposal

Treat the parameters passed to `change-application-permissions` as specifying changes instead of the target state.

The arguments now also have to be specified as JSON, because it is necessary to be able to differentiate between no change, a `None` or an empty list for some of the fields. Thus:

- an absence of a parameter means no change to the corresponding field
- a "null" means a `None` value (where applicable)
- a "[]" means an empty list.

## Test Plan

Manually checked that the parameters are correctly interpreted.

CI will catch regressions.

## Release Plan

    - be released in a new SDK,

## Links

### DIFF
--- a/CLI.md
+++ b/CLI.md
@@ -321,11 +321,11 @@ Open (i.e. activate) a new multi-owner chain deriving the UID from an existing o
 * `--fallback-duration-ms <FALLBACK_DURATION>` — The age of an incoming tracked or protected message after which the validators start transitioning the chain to fallback mode, in milliseconds
 
   Default value: `86400000`
-* `--execute-operations <EXECUTE_OPERATIONS>` — If present, only operations from the specified applications are allowed, and no system operations. Otherwise all operations are allowed
-* `--mandatory-applications <MANDATORY_APPLICATIONS>` — At least one operation or incoming message from each of these applications must occur in every block
-* `--manage-chain <MANAGE_CHAIN>` — These applications are allowed to manage the chain: close it, change application permissions, and change ownership
-* `--call-service-as-oracle <CALL_SERVICE_AS_ORACLE>` — These applications are allowed to call services as oracles on the current chain using the system API
-* `--make-http-requests <MAKE_HTTP_REQUESTS>` — These applications are allowed to make HTTP requests on the current chain using the system API
+* `--execute-operations <EXECUTE_OPERATIONS>` — A JSON list of applications allowed to execute operations on this chain. If set to null, all operations will be allowed. Otherwise, only operations from the specified applications are allowed, and no system operations. Absence of the argument leaves current permissions unchanged
+* `--mandatory-applications <MANDATORY_APPLICATIONS>` — A JSON list of applications, such that at least one operation or incoming message from each of these applications must occur in every block. Absence of the argument leaves current mandatory applications unchanged
+* `--manage-chain <MANAGE_CHAIN>` — A JSON list of applications allowed to manage the chain: close it, change application permissions, and change ownership. Absence of the argument leaves current managing applications unchanged
+* `--call-service-as-oracle <CALL_SERVICE_AS_ORACLE>` — A JSON list of applications that are allowed to call services as oracles on the current chain using the system API. If set to null, all applications will be able to do so. Absence of the argument leaves the current value of the setting unchanged
+* `--make-http-requests <MAKE_HTTP_REQUESTS>` — A JSON list of applications that are allowed to make HTTP requests on the current chain using the system API. If set to null, all applications will be able to do so. Absence of the argument leaves the current value of the setting unchanged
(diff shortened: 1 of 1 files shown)


linera-client/src/client_options.rs
```diff
@@ -399,34 +399,41 @@ impl TryFrom<ChainOwnershipConfig> for ChainOwnership {
...
     /// every block.
-    #[arg(long)]
-    pub mandatory_applications: Option<Vec<ApplicationId>>,
-    /// These applications are allowed to manage the chain: close it, change application
-    /// permissions, and change ownership.
-    #[arg(long)]
-    pub manage_chain: Option<Vec<ApplicationId>>,
+    #[arg(long, value_parser = util::parse_json_app_vec)]
+    pub mandatory_applications: Option<std::vec::Vec<ApplicationId>>,
+    /// These applications are allowed to close the current chain using the system API.
```
afck: I think you reverted the comment.

linera-client/src/client_options.rs
```diff
@@ -399,34 +399,41 @@ impl TryFrom<ChainOwnershipConfig> for ChainOwnership {
 pub struct ApplicationPermissionsConfig {
     /// If present, only operations from the specified applications are allowed, and
     /// no system operations. Otherwise all operations are allowed.
-    #[arg(long)]
-    pub execute_operations: Option<Vec<ApplicationId>>,
+    #[arg(long, value_parser = util::parse_json_optional_app_vec)]
+    pub execute_operations: Option<std::option::Option<Vec<ApplicationId>>>,
```
afck, line 410: Do we really need `std::option::`? (Also `vec` below.)

linera-client/src/client_options.rs
bart-linera: Whoops, good catch!

linera-client/src/client_options.rs
bart-linera: Unfortunately yes :grimacing: Without it, the `derive(Arg)` macro misidentifies the type it should expect from the parsing function (it has special handling for `Option<Option<_>>` and `Option<Vec<_>>`). This is the only way I'm aware of that can throw it off and make it expect the full type.

I was going to add a comment about that, but I forgot - I'll add one.

bart-linera:
> Do we have any test or example in a README to see how this looks?

If we have, they will be outdated - good point, I should make sure that we have something and that it is up to date :+1:

afck (approved):

linera-client/src/client_options.rs
```diff
@@ -397,36 +397,55 @@ impl TryFrom<ChainOwnershipConfig> for ChainOwnership {
...
-impl From<ApplicationPermissionsConfig> for ApplicationPermissions {
-    fn from(config: ApplicationPermissionsConfig) -> ApplicationPermissions {
-        ApplicationPermissions {
-            execute_operations: config.execute_operations,
-            mandatory_applications: config.mandatory_applications.unwrap_or_default(),
-            manage_chain: config.manage_chain.unwrap_or_default(),
-            call_service_as_oracle: config.call_service_as_oracle,
-            make_http_requests: config.make_http_requests,
+impl ApplicationPermissionsConfig {
+    pub fn update(self, application_permissions: &mut ApplicationPermissions) {
```
deuszx, line 434: The title of the PR says _"treat as deltas, not desired value"_ but the logic here looks like the latter: we're not updating (adding or removing) – we're replacing.

ma2bd (approved):

ma2bd:
Can we modify `change-ownership` as well along the same lines? (here or in a different PR?)

linera-client/src/client_options.rs
bart-linera: Right, it may be a bit misleading :grimacing: The idea is that you will only specify the fields you're changing after this PR, while now you need to specify _all_ the fields you want to have non-default values. So if you want to add an app to, say, `execute_operations`, but you already have some apps in `mandatory_applications` and don't want to change that, you will need to specify both, while with this PR it's enough to specify `execute_operations`.

I guess it would be possible to make this a true delta, and have options like, say, `--add-mandatory-application` and `--remove-mandatory-application`, but it would be a somewhat bigger change.

bart-linera:
@ma2bd I can merge this one for now, and prepare a follow-up PR to this effect :+1:

linera-client/src/client_options.rs
deuszx: Right, I'd call it "selective update". Or actually not call it anything special at all. It's just now each field of `ApplicationPermissionConfig` can be updated independently.

PR #5227, opened 2026-01-09
Improve the CLI UX for change-ownership

bart-linera:
## Motivation

#5223 improved the CLI UX for `change-application-permissions`, but `change-ownership` still requires specifying all the relevant fields.

## Proposal

Make analogous changes to `change-ownership`.

## Test Plan

Matching engine and AMM readme tests run the `change-ownership` command, they were ran locally to check the correctness of the changes.

CI will catch other regressions.

## Release Plan

    - be released in a new SDK,

## Links

### DIFF
--- a/linera-service/src/cli_wrappers/wallet.rs
+++ b/linera-service/src/cli_wrappers/wallet.rs
@@ -902,12 +902,12 @@ impl ClientWrapper {
             .arg("open-multi-owner-chain")
             .args(["--from", &from.to_string()])
             .arg("--owners")
-            .args(owners.iter().map(AccountOwner::to_string))
+            .arg(serde_json::to_string(&owners)?)
             .args(["--base-timeout-ms", &base_timeout_ms.to_string()]);
         if !weights.is_empty() {
             command
                 .arg("--owner-weights")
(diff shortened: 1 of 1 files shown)


linera-client/src/client_options.rs
```diff
@@ -289,71 +289,83 @@ impl Options {
...
+    // from the parser. Without it, it infers the inner type (so either ApplicationId or
+    // Vec<ApplicationId>), which is not what we want here - we want the parsers to return the full
+    // expected types.
+    #[arg(long, value_parser = util::parse_json::<Vec<AccountOwner>>)]
+    pub super_owners: Option<std::vec::Vec<AccountOwner>>,
 
-    /// The new regular owners.
-    #[arg(long, num_args(0..))]
-    pub owners: Vec<AccountOwner>,
+    /// A JSON list of the new owners. Absence of the argument leaves the current list of
```
afck, line 302: (Now that we're using JSON anyway, we might as well combine `owners` and `owner_weights` into a single JSON map. But we can decide that later, of course.)

ma2bd: yes I was going to say the same!

PR #5233, opened 2026-01-12
Feature/certificate lookup by height

deuszx:
## Motivation

Currently, in order to read certificates from the storage we have to go through a corresponding chain actor. This affects all queries that concern that chain as actors process requests sequentially. It has the biggest, observable slowdown when syncing outdated validators – we look up many certificate hashes which blocks the corresponding chain actor for the duration of the query.

  This PR adds efficient height-based certificate indexing to support these operations.

##  Proposal

  This PR introduces a reverse index from `(ChainId, BlockHeight)` to certificate hash, enabling height-based certificate lookups:

  1. Adds `RootKey::BlockByHeight(ChainId, BlockHeight)` to create a reverse index mapping heights to certificate hashes
  2. When certificates are added via `add_certificate()`, the storage layer automatically writes the `(chain_id, height) -> hash` mapping alongside the certificate
  3. Adds `read_certificates_by_heights()` to the `Storage` trait

##  Test Plan
New tests added:
  - Multi-chain isolation: Verifies that height lookups correctly isolate certificates across different chains
  - Duplicate height requests: Ensures all duplicate positions are correctly filled (fixed bug where only last index was populated)
  - Consistency validation: Confirms that certificates retrieved by height match those retrieved by hash
  - Edge cases: Empty requests, missing heights, and cross-chain queries

  All tests pass in CI. Manual testing verified that the index is correctly maintained during certificate addition and that batch retrieval preserves ordering.

##  Release Plan

##  Links

  - https://github.com/linera-io/linera-protocol/blob/main/CONTRIBUTING.md#reviewer-checklist

### DIFF
--- a/linera-core/src/updater.rs
+++ b/linera-core/src/updater.rs
@@ -805,14 +805,15 @@ where
 
         let certificates = storage.read_certificates(hashes.clone()).await?;
 
-        // Write back the height->hash indices we learned from the fallback
-        let indices: Vec<_> = heights.into_iter().zip(hashes.clone()).collect();
-        storage
-            .write_certificate_height_indices(chain_id, &indices)
-            .await?;
-
-        match ResultReadCertificates::new(certificates, hashes) {
-            ResultReadCertificates::Certificates(certs) => Ok(certs),
+        match ResultReadCertificates::new(certificates, hashes.clone()) {
+            ResultReadCertificates::Certificates(certs) => {
+                // Write back the height->hash indices we learned from the fallback
+                let indices: Vec<_> = heights.into_iter().zip(hashes.clone()).collect();
+                storage
+                    .write_certificate_height_indices(chain_id, &indices)
+                    .await?;
+                Ok(certs)
+            }
             ResultReadCertificates::InvalidHashes(hashes) => {
                 Err(ChainClientError::ReadCertificatesError(hashes))
             }


linera-storage/src/db_storage.rs
```diff
@@ -284,6 +284,14 @@ impl MultiPartitionBatch {
         Ok(())
     }
 
+    /// Adds a certificate to the batch.
+    ///
+    /// Writes both the certificate data (indexed by hash) and a height index
+    /// (mapping chain_id + height to hash).
+    ///
+    /// Note: If called multiple times with the same (chain_id, height), the height
```
```suggestion
    /// Note: If called multiple times with the same `(chain_id, height)`, the height
```

linera-storage/src/lib.rs
```diff
@@ -152,6 +152,15 @@ pub trait Storage: linera_base::util::traits::AutoTraits + Sized {
         hashes: I,
     ) -> Result<Vec<(Vec<u8>, Vec<u8>)>, ViewError>;
 
+    /// Reads certificates by heights for a given chain.
+    /// Returns a vector where each element corresponds to the input height.
+    /// Elements are None if no certificate exists at that height.
```
afck:
```suggestion
    /// Elements are `None` if no certificate exists at that height.
```

linera-storage/src/db_storage.rs
```diff
@@ -450,6 +469,492 @@ mod tests {
...
+            data_types::{BlockHeight, Epoch, Round, Timestamp},
+        };
+        use linera_chain::{
+            block::{Block, BlockBody, BlockHeader, ConfirmedBlock},
+            types::ConfirmedBlockCertificate,
+        };
+        use linera_views::{
+            memory::MemoryDatabase,
+            store::{KeyValueDatabase, ReadableKeyValueStore as _},
+        };
```
afck: We have a `with_testing` `use` at the beginning of the file, too.

linera-storage/src/db_storage.rs
```diff
@@ -450,6 +469,492 @@ mod tests {
...
+                    chain_id,
+                    epoch: Epoch::ZERO,
+                    height: BlockHeight(height),
+                    timestamp: Timestamp::from(0),
+                    state_hash: CryptoHash::new(&TestString::new(format!("state_hash_{}", height))),
+                    previous_block_hash: None,
+                    authenticated_signer: None,
+                    transactions_hash: CryptoHash::new(&TestString::new(format!(
+                        "tx_hash_{}",
+                        height
```
afck: Maybe more concise and more common (and also a pedantic Clippy lint, I believe):

```suggestion
                        "tx_hash_{height}"
```

linera-storage/src/db_storage.rs
```diff
@@ -450,6 +469,492 @@ mod tests {
...
+        let result = storage
+            .read_certificates_by_heights(chain_a, &[BlockHeight(20)])
+            .await
+            .unwrap();
+        assert!(result[0].is_none());
+    }
+
+    #[cfg(with_testing)]
+    #[tokio::test]
+    async fn test_read_certificates_by_heights_duplicates() {
```
afck: Can't that be part of `test_read_certificates_by_heights`? Would be just a few more lines instead of having to do all that setup again.

linera-storage/src/db_storage.rs
```diff
@@ -940,6 +1445,60 @@ where
...
+
+            let hash_option = value_bytes
+                .map(|bytes| bcs::from_bytes::<CryptoHash>(&bytes))
+                .transpose()?;
+            hash_options.push(hash_option);
+        }
+
+        // Step 2: Collect valid hashes and build index mapping
+        let mut hash_to_indices: std::collections::HashMap<CryptoHash, Vec<usize>> =
+            std::collections::HashMap::new();
```
afck: Why not import `HashMap`?

linera-core/src/updater.rs
```diff
@@ -792,6 +768,47 @@ where
...
+
+        if all_found {
+            return Ok(certificates_by_height.into_iter().flatten().collect());
+        }
+
+        // Fallback to the traditional approach
+        let hashes = self
+            .client
+            .local_node
+            .get_block_hashes(chain_id, heights)
```
afck: We could get only the _missing_ heights here.

(If that's complicated because of the positions, we can probably just make sure the list is always ordered by height. It's a private method anyway.)

deuszx: I don't understand. This method is called with missing heights only.

afck: `heights` is still the same vec that was passed into this method, isn't it? So we haven't removed those heights for which `certificates_by_height` does contain a `Some` entry?

deuszx: Ah, that one. I don't think it will actually matter – either we have the reverse index for the range or not, I don't think there will be gaps in it.

The new code will write the index for all new blocks. The migration will migrate everything.

bart-linera (approved):
Some minor things already pointed out by Andreas aside, looks good to me :+1:

linera-storage/src/db_storage.rs
deuszx: Done in [Address review comments](https://github.com/linera-io/linera-protocol/pull/5233/commits/d055a448c3390df1a900d6f785edcd322fa5c7ae)

linera-storage/src/db_storage.rs
deuszx: Done in [Address review comments](https://github.com/linera-io/linera-protocol/pull/5233/commits/d055a448c3390df1a900d6f785edcd322fa5c7ae)

linera-storage/src/db_storage.rs
deuszx: Done in [Address review comments](https://github.com/linera-io/linera-protocol/pull/5233/commits/d055a448c3390df1a900d6f785edcd322fa5c7ae)

linera-storage/src/db_storage.rs
deuszx: Done in [Address review comments](https://github.com/linera-io/linera-protocol/pull/5233/commits/d055a448c3390df1a900d6f785edcd322fa5c7ae)

linera-storage/src/lib.rs
deuszx: Done in [Address review comments](https://github.com/linera-io/linera-protocol/pull/5233/commits/d055a448c3390df1a900d6f785edcd322fa5c7ae)

linera-storage/src/db_storage.rs
deuszx: Done in [Address review comments](https://github.com/linera-io/linera-protocol/pull/5233/commits/d055a448c3390df1a900d6f785edcd322fa5c7ae)

linera-storage/src/db_storage.rs
```diff
@@ -351,6 +369,7 @@ pub(crate) enum RootKey {
     Placeholder,
     NetworkDescription,
     BlockExporterState(u32),
+    BlockByHeight(ChainId, BlockHeight),
```
MathieuDutSik: So, yes I think it would be preferable to group all the entries with `ChainId` together with something like
`ChainId(ChainId)`.

Then, we would have an internal
```rust
enum ChainIdCases {
  ChainState,
  Event,
  BlockByHeight(BlockHeight),
}
```

Because all of the entries of the `ChainId` should be on the same partition.

deuszx: Done in https://github.com/linera-io/linera-protocol/pull/5233/commits/271a28ad74d11b2d500fbb683f85b5d29353abe5

linera-storage/src/db_storage.rs
```diff
@@ -313,12 +313,13 @@ impl MultiPartitionBatch {
         key_values.push((key, value));
         self.put_key_values(root_key, key_values);
 
-        // Write height index: (chain_id, height) -> hash
+        // Write height index: `chain_id`` -> `height `-> hash
```
afck: Duplicate backtick. But this isn't rustdoc anyway, so it doesn't have to be Markdown.

```suggestion
        // Write height index: chain_id -> height -> hash
```

linera-storage/src/db_storage.rs
MathieuDutSik: No, this is not done in this commit.
The `ChainState`, `Event` and `BlockByHeight` will be in different shards of the database.

deuszx: This is a bigger refactor that does not concern changes specifically done in this PR. It could have been done before when the `Event` variant was added.

afck: I think we do want to have them in separate partitions. (Or at least the chain state view should remain on its own partition.)

We just want all the by-height lookup entries for the same chain in the same partition.

Issue #4948, opened 2025-11-11
Don't retry operations at the next height by default.


afck:
`ChainClient::execute_operations` currently retries the operations at the next block height if a conflicting block was committed. This isn't always desirable and shouldn't be the default behavior.

afck:
With this change it's questionable whether the client should try to finalize anything other than its own proposal at all, implicitly. Up to now, this was done specifically so that the client would then be in a position to make its new proposal at the next height.

Basically, when trying to commit our own operation, we can end up in different situations, e.g.:
* We already have a conflicting local pending block (i.e. our own previous operation that we hadn't managed to commit).
* There is a conflicting locked block: In this case, the _only_ thing that can move the chain forward is to try and get that committed, even if it's someone else's. (But maybe we want to leave even that to a different owner?)
* There's a validated block: The best chance to make the chain progress quickly is to try and finalize that, but we _could_ still introduce our own conflicting proposal, which would possibly make the next round fail and slow down the chain.
* We're in single-leader mode and it's not our turn: We have to wait for the round timeout, get a timeout certificate—and then see whether there's a locked block or we can propose our own.
* Everything's free and we can commit our block.

We need to decide what the client should do in each case, and also how CLI commands, node service mutations etc. should behave.

An argument to keep the behavior where we try to finalize whatever is waiting, even if it is another owner's block, is that it keeps the number of possible outcomes of client operations small: It's `WaitForTimeout`, `Conflict` or `Committed`. Otherwise we need to add a new variant for a conflicting pending (but not yet committed) locked block etc.

PR #5234, opened 2026-01-12
Don't automatically retry operations on conflict.

afck:
## Motivation

The chain client currently automatically retries the same operations at the next block height if a conflicting block got committed at the current height. But the same operations are not necessarily desired at the new chain state anymore.

## Proposal

Remove the automatic retry behavior, and return a `ClientOutcome::Conflict` instead, to let the caller decide.

## Test Plan

The tests were updated.

## Release Plan

    - be released in a new SDK.

## Links

- Closes https://github.com/linera-io/linera-protocol/issues/4948.


linera-core/src/data_types.rs
```diff
@@ -397,6 +403,7 @@ impl<T> ClientOutcome<T> {
         match self {
             ClientOutcome::Committed(t) => ClientOutcome::Committed(f(t)),
             ClientOutcome::WaitForTimeout(timeout) => ClientOutcome::WaitForTimeout(timeout),
+            ClientOutcome::Conflict(certificate) => ClientOutcome::Conflict(certificate),
```
bart-linera, line 406: Would it make sense to just do `other_outcome => other_outcome`, or do we want to explicitly handle each variant just not to miss another one requiring special handling in case we ever add one? (Also below.)

afck: No, it's a different type because of the `f(t)` above.

Issue #4742, opened 2025-10-06
Applications don't load with Rust 1.87 or later


Twey:
Starting from toolchain 1.87 upwards the resulting Wasm is not compatible with the network. It throws the following error when trying to create the instance of the contract:

```
2025-10-06T15:52:25.971095Z  INFO linera::main: linera_client::client_context: Module published successfully!
2025-10-06T15:52:25.971144Z  INFO linera::main: linera_client::client_context: Synchronizing client and processing inbox
2025-10-06T15:52:26.244197Z  INFO linera::main: linera_core::client: Received 0 certificates from 8 validator(s).
2025-10-06T15:52:26.614261Z ERROR linera: Error is Failed to create application

Caused by:
    chain client error: Local node operation failed: Worker operation failed: Execution error: Invalid Wasm module: Unknown opcode 252 during Operation(0)
```

You can see it publishes just fine and fails later – which is not ideal.To fix this, the rust toolchain used has to be 1.86 at most. Anything after that seems to produce a Wasm code that contains the disallowed "memory bulk" operation.

afck:
Once this is done, we can revert #4894.

deuszx:
https://github.com/linera-io/linera-protocol/pull/5646

ndr-ds:
What's left to do here before we can close this?

ma2bd:
@ndr-ds backport?

Issue #4203, opened 2025-07-07
Upgrade to Wasmer 5


Twey:

deuszx:
https://github.com/wasmerio/wasmer/pull/6252

PR #5236, opened 2026-01-12
Fix some upcoming rustc warnings

Twey:
## Motivation

Some combination of Wasm target and [RPITIT](https://rustc-dev-guide.rust-lang.org/return-position-impl-trait-in-trait.html) causes the compiler to ICE sometimes when invoked by `rust-analyzer`.

## Proposal

I wanted to bump the nightly and fix some new warnings.

(Note that this does not interact with https://github.com/linera-io/linera-protocol/issues/4742 because we only use the stable compiler for building applications.)

Unfortunately, our Wasmer version also has a bug that makes it incompatible with newer Rust.  We need to fix that first before actually bumping the Rust compiler.  There's nothing to stop us pre-emptively fixing the warnings though.

## Test Plan

Tested locally.

## Release Plan

## Links

### DIFF
--- a/toolchains/nightly/rust-toolchain.toml
+++ b/toolchains/nightly/rust-toolchain.toml
@@ -1,5 +1,5 @@
 [toolchain]
-channel = "nightly-2026-01-05"
+channel = "nightly-2025-03-21"
 components = [ "clippy", "rustfmt", "rust-src" ]
 targets = [ "wasm32-unknown-unknown" ]
 profile = "minimal"


Twey:
This can't work until we [upgrade Wasmer](https://github.com/linera-io/linera-protocol/issues/4203), since we hit Wasmer bug https://github.com/wasmerio/wasmer/issues/5610.  I'll keep the warning fixes but remove the actual version bump.

PR #5237, opened 2026-01-12
Backport: Merge tracked chains with listening mode map. (#5232)

bart-linera:
## Motivation

`tracked_chains` and `ListeningMode`s both express the client's relation to a chain, and affect how e.g. cross-chain messages to these chains are treated.

## Proposal

Merge them to have a single source of truth.

This is a backport of @afck 's #5232 

## Test Plan

This is mostly a refactoring, so existing tests should suffice.

## Release Plan

## Links

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -196,6 +196,18 @@ impl ListeningMode {
             (ListeningMode::FollowChain, _) => (),
         }
     }
+
+    /// Returns whether this mode implies follow-only behavior (i.e., not participating in
+    /// consensus rounds).
+    pub fn is_follow_only(&self) -> bool {
+        !matches!(self, ListeningMode::FullChain)
+    }
+
+    /// Returns whether this is a full chain mode (synchronizing sender chains and updating
+    /// inboxes).
+    pub fn is_full(&self) -> bool {
+        matches!(self, ListeningMode::FullChain)
+    }
 }
 
 /// A builder that creates [`ChainClient`]s which share the cache and notifiers.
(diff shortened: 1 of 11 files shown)


linera-core/src/client/mod.rs
```diff
@@ -297,13 +309,35 @@ impl<Env: Environment> Client<Env> {
...
+    pub fn chain_mode(&self, chain_id: ChainId) -> Option<ListeningMode> {
+        self.chain_modes
+            .read()
+            .expect("Panics should not happen while holding a lock to `chain_modes`")
+            .get(&chain_id)
+            .cloned()
+    }
+
+    /// Returns whether a chain is fully tracked by the local node.
+    pub fn is_tracked(&self, chain_id: ChainId) -> bool {
```
deuszx, line 335: s/is_tracked/is_fully_tracked ?

afck: Right, or just `is_full`? (We could remove the term "tracked" entirely.)

(If we change it, we should also change it on main.)

deuszx: `is_full` is not non-ambiguous from the call-site

afck: …or `is_full_chain`?

On the other hand, I don't know if `ListeningMode` and `FullChain` etc. are the best terms anymore. Maybe it should be `TrackingMode`… not sure.

bart-linera: We could change it, but then it would be good to change it on `main` as well, I think.

EDIT: Ah, I wrote this response only seeing the first comment, GitHub doesn't refresh threads automatically :man_facepalming:

ma2bd: I'm just going to land this PR to unblock my stuff. Feel free to iterate in other PRs

PR #5238, opened 2026-01-12
Don't automatically make every child chain a full chain.

afck:
## Motivation

The chain worker used to add every chain that was opened as a child of a tracked chain to the tracked chains, too. That was so that _if_ the current user owned the child chain, and the chain listener added it, it would be able to receive messages.

But now tracked chains have been merged with the listening modes (https://github.com/linera-io/linera-protocol/pull/5232), so it suffices to add these in the chain listener.

## Proposal

Move that functionality from the chain worker to the client/listener.

## Test Plan

CI

## Release Plan

    - be released in a new SDK.

## Links

- Follow-up to #5232.

### DIFF
--- a/linera-client/src/chain_listener.rs
+++ b/linera-client/src/chain_listener.rs
@@ -363,6 +363,7 @@ impl<C: ClientContext + 'static> ChainListener<C> {
             .await?
             .ok_or(chain_client::Error::MissingConfirmedBlock(hash))?
             .into_block();
+        let parent_chain_id = block.header.chain_id;
         let blobs = block.created_blobs().into_iter();
         let new_chains = blobs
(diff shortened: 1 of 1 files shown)


deuszx (commented):
Do we already have a test for this?

afck:
@deuszx: I added a test in https://github.com/linera-io/linera-protocol/pull/5238/commits/20d4ad0be72ebade7f779bcfcbb7d232ea7bcd0e.

afck:
Right, for use cases like the RFQ app (see failed README test) we _do_ need to start following all newly created chains that we own, even if we weren't the ones who created them. Will fix…

ma2bd (approved):

PR #5239, opened 2026-01-12
Set RUST_LOG=linera=info for remote compatibility and rust tests

deuszx:
## Motivation

Our tests' logging is very noisy.

## Proposal

Set `linera=info` for remote compatibility tests and `rust.yml`.

## Test Plan

CI

## Release Plan

## Links

### DIFF
--- a/.github/workflows/remote_compatibility.yml
+++ b/.github/workflows/remote_compatibility.yml
@@ -33,7 +33,7 @@ env:
   RUSTDOCFLAGS: -A rustdoc::redundant_explicit_links -D warnings
   RUSTFLAGS: -D warnings
   RUSTUP_MAX_RETRIES: 10
-  RUST_LOG: linera=debug
+  RUST_LOG: linera=info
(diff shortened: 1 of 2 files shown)


.github/workflows/remote_compatibility.yml
```diff
@@ -33,7 +33,7 @@ env:
   RUSTDOCFLAGS: -A rustdoc::redundant_explicit_links -D warnings
   RUSTFLAGS: -D warnings
   RUSTUP_MAX_RETRIES: 10
-  RUST_LOG: linera=debug
+  RUST_LOG: linera=info
```
ma2bd, line 36: I believe removing the line would set the INFO level by default. Are you worried about non-linera INFO logs?

deuszx: I just wanted to make it easier to re-run with `debug` in case some PR fails and needs more detailed logging.

PR #5241, opened 2026-01-13
[testnet] Tentatively fix the controller logic

ma2bd:
## Motivation

Fix the controller logic

## Proposal

* remove buggy `return` statement
* add logs to help with debugging
* make sure that we assign a default owner to chain in "full" tracking mode

## Test Plan

CI + tested manually

## Release Plan

to be backported on `main`

### DIFF
--- a/linera-client/src/chain_listener.rs
+++ b/linera-client/src/chain_listener.rs
@@ -611,6 +611,7 @@ impl<C: ClientContext + 'static> ChainListener<C> {
                                         chain.timestamp = chain_description.timestamp();
                                         chain.epoch = Some(chain_description.config().epoch);
                                         // Insert (overwrites if already exists)
+                                        // TODO(#5247): Do not reset preexisting epoch and timestamp.
                                         context_guard
                                             .wallet()
                                             .insert(chain_id, chain)
--- a/linera-client/src/client_context.rs
+++ b/linera-client/src/client_context.rs
@@ -490,6 +490,7 @@ impl<Env: Environment> ClientContext<Env> {
             return Err(error::Inner::ChainOwnership.into());
         }
 
+        // TODO(#5247): Do not reset preexisting epoch and timestamp.
         self.wallet()
             .insert(
                 chain_id,


linera-client/src/chain_listener.rs
```diff
@@ -580,7 +581,49 @@ impl<C: ClientContext + 'static> ChainListener<C> {
                     match command {
                         ListenerCommand::Listen(new_chains) => {
                             debug!(?new_chains, "received command to listen to new chains");
-                            self.listen_recursively(new_chains).await?;
+                            let mut chains = BTreeMap::new();
+                            let context_guard = self.context.lock().await;
```
ma2bd, line 585: Is it ok to take a lock here?

linera-client/src/chain_listener.rs
```diff
@@ -580,7 +581,49 @@ impl<C: ClientContext + 'static> ChainListener<C> {
...
+                                        let mut chain = context_guard
+                                            .wallet()
+                                            .get(chain_id)
+                                            .await
+                                            .map_err(error::Inner::wallet)?
+                                            .unwrap_or_default();
+                                        // Update owner and chain metadata
+                                        chain.owner = Some(owner);
+                                        chain.timestamp = chain_description.timestamp();
+                                        chain.epoch = Some(chain_description.config().epoch);
```
ma2bd, line 612: This overrides any preexisting timestamp and epoch which is not a ideal. This being said, `assign_new_chain_to_key` seems to also have this problem.

examples/controller/src/contract.rs
```diff
@@ -89,6 +93,7 @@ impl Contract for ControllerContract {
                     self.execute_worker_command_locally(owner, command, creator_chain_id)
                         .await;
                 } else {
+                    log::info!("Sending worker command from {} to {creator_chain_id} for remote execution: {owner} {command:?}", self.runtime.chain_id());
```
afck:
```suggestion
                    log::info!(
                        "Sending worker command from {} to {creator_chain_id} for remote \
                        execution: {owner} {command:?}",
                        self.runtime.chain_id()
                    );
```

(Also below.)

examples/controller/src/contract.rs
```diff
@@ -223,19 +234,20 @@ impl ControllerContract {
...
     ) {
-        let Some(admins) = self.state.admins.get() else {
+        log::info!("Executing controller command locally: {admin} {command:?}");
+        if let Some(admins) = self.state.admins.get() {
+            assert!(
+                admins.contains(&admin),
+                "Controller command can only be executed by an authorized admin account. Got {admin}"
+            );
+        } else {
             // No admin list was set yet. Hence, everyone is an owner. However, messages
```
afck, line 252: (In other words, if it's not set, every _chain owner_ is an admin?)

linera-client/src/chain_listener.rs
afck: (At a glance I'd say yes.)

linera-client/src/chain_listener.rs
afck: Yes, we could instead take the `max` here, or only insert if the entry doesn't exist yet. But it'll get updated with the next block anyway.

linera-client/src/chain_listener.rs
```diff
@@ -580,7 +581,49 @@ impl<C: ClientContext + 'static> ChainListener<C> {
...
+                                        context_guard
+                                            .wallet()
+                                            .insert(chain_id, chain)
+                                            .await
+                                            .map_err(error::Inner::wallet)?;
+                                        chains.insert(chain_id, ListeningMode::FullChain);
+                                        continue;
+                                    }
+                                }
+                                chains.insert(chain_id, ListeningMode::FollowChain);
```
afck, line 624: When will `owner` be `None`?

afck (approved):
No blockers from my side, but maybe @bart-linera can also take a look?

bart-linera (approved):
Yup, looks good to me, too :+1:

linera-client/src/chain_listener.rs
ma2bd: Yes, I'm assuming that we don't need to prepare the wallet in this case.

examples/controller/src/contract.rs
ma2bd: correct, but we don't accept messages in this case (so you need to be an owner of the chain that created the app)

ma2bd: I found that this was a useful design pattern and end up using it across all the recent apps

linera-client/src/chain_listener.rs
ma2bd: I created a task

ma2bd: (It's fixed on main by the way)

PR #5243, opened 2026-01-13
Delay proposal handling instead of `sleep`.

afck:
## Motivation

When handling a block proposal with a timestamp slightly in the future, the chain worker actor currently `sleep`s until that time. During that period, it cannot handle any other requests.

## Proposal

Delay handling such proposals until the timestamp. In the meantime, continue processing requests.

## Test Plan

A worker test was added.

## Release Plan

## Links

### DIFF
--- a/linera-core/src/chain_worker/actor.rs
+++ b/linera-core/src/chain_worker/actor.rs
@@ -31,7 +31,7 @@ use linera_execution::{
 use linera_storage::{Clock as _, Storage};
 use linera_views::context::{Context, InactiveContext};
 use tokio::sync::{mpsc, oneshot, OwnedRwLockReadGuard};
-use tracing::{instrument, trace, Instrument as _};
+use tracing::{debug, instrument, trace, Instrument as _};
 
 use super::{config::ChainWorkerConfig, state::ChainWorkerState, DeliveryNotifier};
 use crate::{


linera-core/src/chain_worker/actor.rs
```diff
@@ -337,17 +390,69 @@ where
...
-        self.storage.clock().sleep_until(timeout).await
+        now.saturating_add(ttl)
+    }
+
+    /// Checks if a block proposal should be delayed because its timestamp is in the future.
+    ///
+    /// Returns `Some(timestamp)` if the proposal should be delayed until that timestamp,
+    /// or `None` if it should be processed immediately (either because the timestamp is
+    /// not in the future, or because it's beyond the grace period and should error).
+    fn should_delay_proposal(&self, proposal: &BlockProposal) -> Option<Timestamp> {
```
deuszx: Minor: rename to `delay_until`

linera-core/src/chain_worker/actor.rs
```diff
@@ -337,17 +390,69 @@ where
...
+        // Only validators need to wait for the timestamp.
```
deuszx: Is this really necessary here? i.e. if a non-validator node returns `None` here, what will happen?

afck: They handle the proposal locally immediately, which is usually preferable to another delay.

We just want to make sure that no proposal gets _signed_ earlier than its timestamp.

linera-core/src/chain_worker/actor.rs
```diff
@@ -396,13 +506,69 @@ where
             .instrument(span.clone())
             .await?;
 
-            Box::pin(worker.handle_request(request))
-                .instrument(span)
-                .await;
+            // Check if the first request should be delayed.
+            if let ChainWorkerRequest::HandleBlockProposal { ref proposal, .. } = request {
+                if let Some(timestamp) = self.should_delay_proposal(proposal) {
+                    debug!("Delaying block proposal until timestamp {}", timestamp);
```
deuszx:
```rust
debug!(delayed_until=?timestamp, "delaying block proposal")
```

linera-core/src/chain_worker/actor.rs
```diff
@@ -396,13 +506,69 @@ where
...
+                    delayed_proposals.push(Reverse(DelayedProposal {
+                        timestamp,
+                        sequence: next_sequence,
+                        request,
+                        span,
+                        _queued_at,
+                    }));
+                    next_sequence += 1;
```
deuszx: I don't like that this leaks here. Do we need it? Can't you resolve the conflcits by `request` ? If not, can we encapsulate the `next_sequence` tracking into `DelayedProposals` struct?

afck: Maybe we don't need it at all: I wanted to make sure that _other_ requests are still handled in the order in which they are received. But I guess it's fine for delayed proposals to be handled in any order.

afck: Done in https://github.com/linera-io/linera-protocol/pull/5243/commits/dcc5e0865841427b75891d7362f44dc7c3e97930.

linera-core/src/chain_worker/actor.rs
deuszx: what about nodes that aren't validators but also not clients

afck: I'd say for them it doesn't really matter either: They can't _sign_ the proposal, so they might as well execute it right away?

ma2bd:
> I like the idea but I think the code increases the complexity of the already-complicated code. I'd really like to clean it up a bit.

@afck Would it simplify things if we created an async task that puts back the query in the actor queue after a delay?

afck:
> Would it simplify things if we created an async task that puts back the query in the actor queue after a delay?

A bit, maybe. But I imagine it would be more overhead, and lots of tasks that we need to make sure keep getting polled (also on the web), etc.

ma2bd:
> > Would it simplify things if we created an async task that puts back the query in the actor queue after a delay?
> 
> A bit, maybe. But I imagine it would be more overhead, and lots of tasks that we need to make sure keep getting polled (also on the web), etc.

@Twey what do you think?

linera-core/src/chain_worker/actor.rs
```diff
@@ -396,13 +499,67 @@ where
...
+                                    let queue_wait_time_ms =
+                                        ready._queued_at.elapsed().as_secs_f64() * 1000.0;
+                                    metrics::CHAIN_WORKER_REQUEST_QUEUE_WAIT_TIME
+                                        .observe(queue_wait_time_ms);
+                                }
+                                Box::pin(worker.handle_request(ready.request))
+                                    .instrument(ready.span)
+                                    .await;
+                            }
+                        } else if wakeup_time >= ttl_timeout && delayed_proposals.is_empty() {
```
bart-linera: Is it possible for `wakeup_time` to be `< ttl_timeout`?

afck: Yes, if a block proposal's timestamp is in the future, but not further than the chain worker TTL.

bart-linera: But `wakeup_time` itself is calculated as the minimum of the two timestamps, isn't it?

bart-linera: Oh, wait, I had it backwards :man_facepalming: Nevermind!

bart-linera: Then it shouldn't be possible for the `wakeup_time` to be strictly _greater_ than `ttl_timeout`, is it? And the comparison could technically test for equality?

afck:
I replaced the binary heap with the spawn-tasks-that-sleep-and-reinsert approach.

linera-core/src/chain_worker/actor.rs
```diff
@@ -401,17 +443,37 @@ where
...
+                        break;
+                    }
                     maybe_request = incoming_requests.recv().fuse() => {
-                        let Some((request, span, _queued_at)) = maybe_request else {
+                        let Some((request, span, queued_at)) = maybe_request else {
                             break; // Request sender was dropped.
                         };
 
-                        // Record how long the request waited in queue (in milliseconds)
+                        // Check if this request should be delayed.
```
ma2bd: why does it appear twice?

afck: Right, I should deduplicate that.

(It appears twice because of how `handle_requests` is organized: We listen for incoming requests in the outer loop, and any request causes us to load the chain state. Then we process requests in the inner loop until we are idle for a while. Then we unload the chain state again and proceed with the outer loop.)

linera-core/src/chain_worker/actor.rs
deuszx: but they'll execute it on an invalid block height potentially? Wouldn't it be "more correct" (and simpler) if we delayed unconditionally? For validators, non-validators and local clients?

afck: No, this has nothing to do with the block height: The proposal is already for a specific height, and of course the worker will reject it if it's for the wrong height.

Sure, we could remove that condition. If all participants have accurate clocks, all of this should rarely be relevant anyway.

(But _if_ it is relevant, all it does is create an unnecessary delay in the client, so I'd vote for keeping the condition.)

deuszx: what client? Local? but local client will use the same clock for creating the block proposal and then here – so it should never decide to delay it.

deuszx: I think this is relevant only when proposing client is ahead of the validator. In which case, I think, we should delay for both active and non-active validators. Then we can remove this line and simplify the code (it won't need the `self` anymore).

afck: Yes, in the client this could only trigger if they take part in a multi-owner chain and their clock is _lagging behind_: They could learn from a validator about another owner's pending proposal (that the validator, with an accurate clock, has happily processed), and decide that the proposal's timestamp is "in the future" (because of the local lagging clock).

afck: Removed it in https://github.com/linera-io/linera-protocol/pull/5243/commits/59721ebcd2766a5311c58beb7894b16006555bd4.

afck: Actually, there may be another reason to not do this on the client:

@Twey: On the web, is it okay to `drop` the handle returned from `spawn`?

Twey: Both on the Web and otherwise dropping the handle will result in the task being cancelled.  (It's a [`RemoteHandle`](https://docs.rs/futures/latest/futures/future/struct.RemoteHandle.html).)  If you want to keep it alive you can use [`forget()`](https://docs.rs/futures/latest/futures/future/struct.RemoteHandle.html#method.forget).

afck: Right! Then I'm surprised the test passes! Anyway, I will fix this.

afck: Done in https://github.com/linera-io/linera-protocol/pull/5243/commits/c624d5055dc18507d71ffd0b3d3dbb9d143561e4.

PR #5256, opened 2026-01-14
Handle response w/ gaps in the collection

deuszx:
## Motivation

Now that we have sparse chains, validator may respond with gaps in the certificates.

## Proposal

Don't assume response contains contiguous, non-empty sequence of requested certificates.

## Test Plan

CI

## Release Plan

- These changes should be
    - be released in a new SDK

## Links

### DIFF
--- a/linera-core/src/node.rs
+++ b/linera-core/src/node.rs
@@ -150,6 +150,10 @@ pub trait ValidatorNode {
     ) -> Result<Vec<ConfirmedBlockCertificate>, NodeError>;
 
     /// Requests a batch of certificates from a specific chain by heights.
+    ///
+    /// Returns certificates in ascending order by height. This method does not guarantee
+    /// that all requested heights will be returned; if some certificates are missing,
+    /// the caller must handle that.
     async fn download_certificates_by_heights(
         &self,
         chain_id: ChainId,


linera-rpc/src/grpc/client.rs
```diff
@@ -476,11 +476,13 @@ impl ValidatorNode for GrpcClient {
...
-            // Honest validator should return certificates in the same order as the requested hashes.
-            missing = missing[received.len()..].to_vec();
+            // Remove only the heights we actually received from missing set.
+            for cert in &received {
+                missing.remove(&cert.inner().height());
+            }
             certs_collected.append(&mut received);
         }
-
+        certs_collected.sort_by_key(|cert| cert.inner().height());
```
afck, line 485: We should add to the method documentation that they are returned sorted by height, rather than in the order of the input vec. (Alternatively, the method could accept a _set_ as an input, then that's clear anyway?)

deuszx: Everywhere in the code w assume that the returned value maintains the order of the input 😬

afck: Yes, we could make this function do the same, and re-order it to match the input. Not sure if that's better.

deuszx: I don't understand – the code maintains the order of the returned value (as long as there are things to return). It just doesn't mention that in the docs (what I meant is that we have this implicit assumption everywhere were the input is `Vec<some_ids>` and return value is `Vec<_>`).

afck: I'm confused: This line sorts by height. So if the input `heights` are `[5, 3, 4]`, doesn't it return `[block 3, block 4, block 5]` instead of `[block 5, block 3, block 4]`?

deuszx: Ah, I see what you mean – yes. This method assume the input is in the ascending order (as all our code does actually). I'll change the input argument to a set then 😬

deuszx: The changes to the code are too far-reaching. I will update the docs on this method (the assumptions about the increasing order of block heights should be safe as we do that implicitly everywhere).

deuszx: Done in https://github.com/linera-io/linera-protocol/pull/5256/changes/33c1bc7fb59081bda793472467db03de96f7e91c

PR #5258, opened 2026-01-14
Use more efficient storage lookup by height in proxy

deuszx:
## Motivation

https://github.com/linera-io/linera-protocol/pull/5233 adds a reverse index tracking `(ChainId, BlockHeight) -> Certificate Hash` to the `Storage`. This means we now have a more efficient way to load certificates by (chain, height) tuple. #5233 used the new API on the client side but we also support `DownloadCertificateByHeight` queries on proxy.

## Proposal

This PR refactors `download_certificates_by_heights` and `download_raw_certificates_by_heights` to use the new `Storage` API. Both use the approach:
- check the storage, using the new reverse index, to read certificates be height.
- if we don't get all of the certificates from the reverse index, we use the fallback logic (old logic)
  - write back the mapping to the reverse index

## Test Plan

CI

## Release Plan

- These changes should be backported to the latest `main` branch
- These changes should be

## Links

### DIFF
--- a/linera-storage/src/db_storage.rs
+++ b/linera-storage/src/db_storage.rs
@@ -678,7 +678,7 @@ mod tests {
             .read_certificates_by_heights(chain_id, &heights)
             .await
             .unwrap();
-        assert_eq!(result.len(), 3); // BlockHeight(3) was duplicated.
+        assert_eq!(result.len(), 4); // BlockHeight(3) was duplicated.
         assert!(result[0].is_some());
         assert!(result[1].is_none()); // Height 2 doesn't exist
         assert!(result[2].is_some());
@@ -1469,26 +1469,29 @@ where
             .read_certificate_hashes_by_heights(chain_id, heights)
             .await?;
 
-        // Map from hash to index in the heights array
-        let mut indices = HashMap::new();
+        // Map from hash to all indices in the heights array (handles duplicates)
+        let mut indices: HashMap<CryptoHash, Vec<usize>> = HashMap::new();
         for (index, maybe_hash) in hashes.iter().enumerate() {
             if let Some(hash) = maybe_hash {
-                indices.insert(*hash, index);
+                indices.entry(*hash).or_default().push(index);
             }
         }
 
-        let existing_hashes = hashes.into_iter().flatten().collect::<Vec<_>>();
+        // Deduplicate hashes for the storage query
+        let unique_hashes = indices.keys().copied().collect::<Vec<_>>();
 
         let mut result = vec![None; heights.len()];
 
         for (raw_cert, hash) in self
-            .read_certificates_raw(&existing_hashes)
+            .read_certificates_raw(&unique_hashes)
             .await?
             .into_iter()
-            .zip(existing_hashes)
+            .zip(unique_hashes)
         {
-            if let Some(index) = indices.get(&hash) {
-                result[*index] = raw_cert;
+            if let Some(idx_list) = indices.get(&hash) {
+                for &index in idx_list {
+                    result[index] = raw_cert.clone();
+                }
             } else {
                 // This should not happen, but log a warning if it does.
                 tracing::warn!(


linera-storage/src/db_storage.rs
```diff
@@ -1468,33 +1466,58 @@ where
...
+        heights: &[BlockHeight],
+    ) -> Result<Vec<Option<ConfirmedBlockCertificate>>, ViewError> {
+        let valid_hashes: Vec<Option<CryptoHash>> = self
+            .read_certificate_hashes_by_heights(chain_id, heights)
+            .await?;
+
+        let mut indices = HashMap::new();
+        for (index, maybe_hash) in valid_hashes.iter().enumerate() {
+            if let Some(hash) = maybe_hash {
+                indices.insert(*hash, index);
```
afck: Shouldn't be an issue in most use cases, but: What if the same height occurs multiple times, so that the same hash has more than one index?

linera-service/src/proxy/grpc.rs
```diff
@@ -374,6 +374,56 @@ where
         status.set_source(Arc::new(err));
         status
     }
+
+    /// Fallback method to get certificate hashes by heights using ChainInfoQuery.
```
afck:
```suggestion
    /// Fallback method to get certificate hashes by heights using `ChainInfoQuery`.
```

linera-service/src/proxy/grpc.rs
afck, line 743: Would it be easier if this called `download_raw_certificates_by_heights` and then deserialized the certificates? Might be less duplication?

deuszx: Might be, haven't thought of that. Will look into this.

deuszx: This is not so easy to do due to different types being put in the response (and the limiter working on different serialization formats – protobuf vs bsc) but I have deduplicated code in other places (`db_storage.rs` and some methods here). PTAL again.

linera-storage/src/db_storage.rs
deuszx: That's exactly the test failure. I think it is OK to _not_ return duplicates anymore:

Clients should be prepared that not all requested hashes/heights will be returned and so they [clients] should process the response "by key" (by hash or by height of the returned certificates) rather than as an opaque collection. This means, on their side, processing of the same height/hash would be a no-op.

linera-storage/src/db_storage.rs
afck, line 1422: (Not in this PR, but: Should this be done concurrently?)

linera-storage/src/db_storage.rs
afck: Would it be better to return a `BTreeSet` then, instead of a `Vec`?

linera-storage/src/db_storage.rs
```diff
@@ -1468,36 +1456,74 @@ where
...
+        for (raw_cert, hash) in self
+            .read_certificates_raw(&existing_hashes)
+            .await?
+            .into_iter()
+            .zip(existing_hashes)
+        {
+            if let Some(index) = indices.get(&hash) {
+                result[*index] = raw_cert;
+            } else {
+                // This should not happen, but log a warning if it does.
```
afck, line 1496: Or even an error? This can only happen due to a bug or database corruption, can it?

afck (approved):
No blockers from my side!

linera-storage/src/db_storage.rs
deuszx: And how for will we push it? The gRPC (and other client-facing APIs) work w/ vectors.

linera-storage/src/db_storage.rs
deuszx: A bug only I think b/c both `indices` and `hashes` are created and manipulated in this function. In theory this isn't possible – we shouldn't get a `hash` that we didn't prevously put in `indices`.

afck (approved):

PR #5259, opened 2026-01-14
[testnet]  Delay proposal handling instead of sleep. (#5243)

afck:
Backport of #5243.

## Motivation

When handling a block proposal with a timestamp slightly in the future, the chain worker actor currently `sleep`s until that time. During that period, it cannot handle any other requests.

## Proposal

Delay handling such proposals until the timestamp. In the meantime, continue processing requests.

## Test Plan

A worker test was added.

## Release Plan

## Links

- PR to main: #5243

### DIFF
--- a/linera-core/src/chain_worker/actor.rs
+++ b/linera-core/src/chain_worker/actor.rs
@@ -406,7 +406,7 @@ where
                     clock.sleep_until(delay_until).await;
                     // Re-insert the request into the queue. If the channel is closed,
                     // the actor is shutting down, so we can ignore the error.
-                    let _ = sender.send((request, span, queued_at));
+                    sender.send((request, span, queued_at)).ok();
                 })
                 .forget();
                 return None;


linera-core/src/chain_worker/actor.rs
```diff
@@ -338,22 +343,84 @@ where
...
+        if let ChainWorkerRequest::HandleBlockProposal { ref proposal, .. } = request {
+            if let Some(delay_until) = self.delay_until(proposal) {
+                debug!(%delay_until, "delaying block proposal");
+                let sender = request_sender.clone();
+                let clock = self.storage.clock().clone();
+                task::spawn(async move {
+                    clock.sleep_until(delay_until).await;
+                    // Re-insert the request into the queue. If the channel is closed,
+                    // the actor is shutting down, so we can ignore the error.
+                    let _ = sender.send((request, span, queued_at));
```
ma2bd: nits (not urgent): FWIW I still dislike `let _ =` because we don't know if you're ignoring `.await` or additional layers of `?`.

ma2bd: Maybe we should put this in the coding guidelines and have Claude follow it

ma2bd: I believe `.ok()` or `.unwrap_or(())` is correct. I also found this https://crates.io/crates/ignore-result

afck: I like `.ok()`. I'd prefer that over adding another dependency, I think.

Anyway, I'm experimenting with this here for the other occurrences of `let _`: https://github.com/linera-io/linera-protocol/pull/5261

PR #5260, opened 2026-01-15
Allow tracking of Resource uses.

### LINKED ISSUE #1615: Allow measuring resource usage in application tests
# Motivation

The resources consumed by an application to execute an operation or a message can only be known from the fees spent when staging a block, for example in a client. This makes it hard to verify that applications are using resources in an expected way and also makes it hard to estimate how much resources are required for operations and messages.

# Proposal

Allow application tests (unit and/or integration tests) to obtain a summary of resources consumed during execution. These can then be used to check resource consumption behavior and to obtain better visibility for fee estimates.

MathieuDutSik:
## Motivation

There are resource costs associated with producing a block. This is in particular important
here for integration tests where we would like to know what was the wasm costs among other
things.

Fixes #1615 

## Proposal

The resources expended are returned by the `add_block` function.
This requires adding it all along the execution path. I think this is fine since it is a small
structure and it is anyway useful to have it directly on output.

The data 

## Test Plan

The tests remain the same, but we can have some print statements for the test if one wishes.
The source code looks like that:
```rust
    let (_, resources) = sender_chain
         .add_block(.....
    print_resources("Transfer block", &resources);
```

The output of `cargo test -- --nocapture` is then
```bash
=== Sender transfer block ===
  Wasm fuel: 69687
  Read operations: 1
  Write operations: 1
  Bytes read: 16
  Bytes written: 51
  Bytes stored: 0
=== Transfer block ===
  Wasm fuel: 69687
  Read operations: 1
  Write operations: 1
  Bytes read: 16
  Bytes written: 51
  Bytes stored: 0
=== Receiver reject block ===
  Wasm fuel: 0
  Read operations: 0
  Write operations: 0
  Bytes read: 0
  Bytes written: 0
  Bytes stored: 0
```

for the fungible test.

##  Release Plan

We need to backport to TestNet Conway to have the metrics for the PM.

## Links

None.

### DIFF
--- a/examples/fungible/tests/cross_chain.rs
+++ b/examples/fungible/tests/cross_chain.rs
@@ -9,19 +9,8 @@ use fungible::{FungibleTokenAbi, InitialState, InitialStateBuilder, Parameters};
 use linera_sdk::{
     abis::fungible::FungibleOperation,
     linera_base_types::{Account, AccountOwner, Amount},
-    test::{MessageAction, ResourceTracker, TestValidator},
+    test::{MessageAction, TestValidator},
 };
-
-/// Prints the resource usage from a block execution.
-fn print_resources(label: &str, resources: &ResourceTracker) {
-    println!("=== {label} ===");
-    println!("  Wasm fuel: {}", resources.wasm_fuel);
-    println!("  Read operations: {}", resources.read_operations);
-    println!("  Write operations: {}", resources.write_operations);
-    println!("  Bytes read: {}", resources.bytes_read);
-    println!("  Bytes written: {}", resources.bytes_written);
-    println!("  Bytes stored: {}", resources.bytes_stored);
-}
 
 /// Test transferring tokens across microchains.
 ///
@@ -64,7 +53,7 @@ async fn test_cross_chain_transfer() {
             );
         })
         .await;
-    print_resources("Transfer block", &resources);
+    println!("Transfer block: {resources}");
 
     assert_eq!(
         fungible::query_account(application_id, &sender_chain, sender_account).await,
@@ -119,7 +108,7 @@ async fn test_bouncing_tokens() {
             );
         })
         .await;
-    print_resources("Sender transfer block", &resources);
+    println!("Sender transfer block: {resources}");
 
     assert_eq!(
         fungible::query_account(application_id, &sender_chain, sender_account).await,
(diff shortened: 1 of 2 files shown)


examples/fungible/tests/cross_chain.rs
```diff
@@ -9,9 +9,20 @@ use fungible::{FungibleTokenAbi, InitialState, InitialStateBuilder, Parameters};
 use linera_sdk::{
     abis::fungible::FungibleOperation,
     linera_base_types::{Account, AccountOwner, Amount},
-    test::{MessageAction, TestValidator},
+    test::{MessageAction, ResourceTracker, TestValidator},
 };
 
+/// Prints the resource usage from a block execution.
+fn print_resources(label: &str, resources: &ResourceTracker) {
```
afck: Should we `impl Display for ResourceTracker` in general?

examples/fungible/tests/cross_chain.rs
```diff
@@ -53,6 +64,7 @@ async fn test_cross_chain_transfer() {
             );
         })
         .await;
+    print_resources("Transfer block", &resources);
```
afck: This is the only place where we use it for now, is it?

I wonder if this should be a benchmark instead: I think `criterion` supports benchmarking other measures than time, and it prints the relative change etc. Maybe this is how our examples should suggest using the resource tracker: To benchmark the resource consumption of the app?

examples/fungible/tests/cross_chain.rs
MathieuDutSik: We definitely could.
But the question is how to do it. The `ResourceTracker` has 29 entries. Printing all of them is out of the question.
Maybe printing only the non-zero entries?

afck: Yes, only the non-zero ones makes sense to me.

Maybe some of them could also be merged into a single line, like reads and writes of the same type etc.

examples/fungible/tests/cross_chain.rs
MathieuDutSik: Yes, benchmarking the resource consumption is quite important, which is why I am pushing for this PR.

For example, the `Operation::Transfer` of `fungible` on `main` costs 4% more than on `testnet_conway`.

afck (approved):

Issue #5224, opened 2026-01-08
Allow `examples/controller` to change the client policy of its node-service workers


ma2bd:
Right now we can pass a (uniform) client policy on the command line but this won't work for a dynamic set of applications (and chains).

For the controller's point of view, we probably want one policy per `local_chain` and `local_service`.

PR #5263, opened 2026-01-15
Allow the controller to manage the message policies of the workers

bart-linera:
## Motivation

The controller may need to set the message policies of the workers it manages (#5224 )

## Proposal

Add a `ListenerCommand` that can tell the `ChainListener` to adjust the message policies per-chain-client. Make the controller module in the node service read the message policies from the controller state when processing state changes and emit an appropriate command.

## Test Plan

CI will catch regressions.

## Release Plan

    - be released in a new SDK,

## Links

### DIFF
--- a/linera-service/src/controller.rs
+++ b/linera-service/src/controller.rs
@@ -193,22 +193,22 @@ where
         // These are local_chains that don't have services (not in active_chains)
         self.listened_local_chains = local_chains.difference(&active_chains).cloned().collect();
 
-        if let Err(err) = self.command_sender.send(ListenerCommand::SetMessagePolicy(
+        if let Err(error) = self.command_sender.send(ListenerCommand::SetMessagePolicy(
             state.local_message_policy,
         )) {
-            error!(%err, "error sending a command to chain listener");
+            error!(%error, "error sending a command to chain listener");
         }
-        if let Err(err) = self
+        if let Err(error) = self
             .command_sender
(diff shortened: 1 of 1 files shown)


linera-base/Cargo.toml
```diff
@@ -40,6 +40,7 @@ async-trait.workspace = true
 bcs.workspace = true
 cfg-if.workspace = true
 chrono.workspace = true
+clap.workspace = true
```
ma2bd: FWIW `clap` should be in `[target.'cfg(not(target_arch = "wasm32"))'.dependencies]`

linera-service/src/controller.rs
```diff
@@ -193,6 +193,11 @@ where
         // These are local_chains that don't have services (not in active_chains)
         self.listened_local_chains = local_chains.difference(&active_chains).cloned().collect();
 
+        if let Err(err) = self.command_sender.send(ListenerCommand::SetMessagePolicy(
+            state.local_message_policy,
+        )) {
+            error!(%err, "error sending a command to chain listener");
```
afck: Maybe "error": I think that's what we use elsewhere in the logs.

```suggestion
            error!(%error, "error sending a command to chain listener");
```

bart-linera:
It has to be available in both smart contracts and the client now, and as far as I could tell, it has to be in `linera_base` for that. But correct me if I'm wrong!

Issue #5064, opened 2025-12-03
Persistent Web storage


Twey:
We have a storage backend for the Web based on IndexedDB.  However, it isn't usable because much of our storage code assumes that the storage backend is `Send + Sync`, which is not true of the JavaScript objects held by the IndexedDB backend.  In order to have persistent storage working on the Web we need to remove this assumption, or at least make it conditionally compiled.

Issue #4637, opened 2025-09-24
Make web wallet persistent


deuszx:
Currently, due to some API mismatches, the client here doesn't use the IndexedDB persistence layer.

Originally in linera-web: https://github.com/linera-io/linera-web/issues/12

PR #5265, opened 2026-01-15
Web storage

Twey:
## Motivation

We'd like to not recompute chains on page load, especially for upcoming apps where chains can become quite long.

## Proposal

Store the ephemeral chain data in IndexedDB on the page.

I thought this would be relatively straightforward, as we already had an IndexedDB storage implementation that was tested in CI.  Unfortunately, it turns out that there were some problems not caught by those tests:
- the `create` implementation for the `linera_views::KeyValueDatabase` trait was nonsensical, as it could only be executed inside a `versionchange` transaction, i.e. when connecting to the database
- the library we were using for IndexedDB, `indexed_db_futures`, has [a bug](https://github.com/Alorel/rust-indexed-db/issues/33) that manifests in multithreaded `wasm-bindgen-futures` executors; since we use multithreading for chain execution, this bug is activated

In practice, this means this PR:
- relaxes some `Send` and `Sync` bounds throughout the codebase
- rethinks the way that `linera_views::backends::indexed_db` does namespacing, using the database name for namespacing instead of the object store
- removes the IndexedDB `linera-persistent` implementation, since we aren't using it right now and if we do start using it it will need to be rewritten anyway
- replaces `indexed_db_futures` in `linera-views` with [`indexed-db`](https://docs.rs/indexed-db/)
- adds some functionality to the counter demo to reüse private keys by providing a mnemonic in the URL parameters (for testing purposes only!)

## Test Plan

CI tests should cover basic store functionality.  Additionally, I tested the functionality locally with the counter app (which I extended with functionality to read the wallet mnemonic from the URL query parameters).

## Release Plan

    - be released in a new SDK,

## Links

- fixes https://github.com/linera-io/linera-protocol/issues/5064

### DIFF
--- a/linera-views/src/backends/indexed_db.rs
+++ b/linera-views/src/backends/indexed_db.rs
@@ -310,7 +310,7 @@ impl KeyValueDatabase for IndexedDbDatabase {
     }
 
     async fn create(config: &Self::Config, namespace: &str) -> Result<()> {
-        Self::connect(config, namespace).await?.database;
+        let Self { .. } = Self::connect(config, namespace).await?;
         Ok(())
     }
 


Twey:
The `create` function in `linera_views::backends::indexed_db` makes no sense, as IndexedDB database schemata can only be updated on version change.  Instead of using object stores as namespaces, we should use databases as namespaces.

Unfortunately, https://github.com/Alorel/rust-indexed-db/issues/33 makes this crate hard to use for our use case.  Since we use atomics for our own multithreading, the `wasm-bindgen` multithreaded executor is used by default, triggering this bug.

.gitignore
```diff
@@ -41,6 +41,8 @@ linera-web/**/*-lock.json
 CLAUDE.md
 .serena
 docs/
+
+# Zed
```
ma2bd, line 45: lol, why?

web/@linera/client/src/storage.rs
```diff
@@ -2,18 +2,19 @@
 // SPDX-License-Identifier: Apache-2.0
 
 // TODO(#4637): convert to IndexedDbStore once we refactor Context
```
ma2bd: remove TODO? link this task to #4637

linera-views/src/backends/indexed_db.rs
```diff
@@ -266,71 +259,63 @@ impl KeyValueDatabase for IndexedDbDatabase {
...
 
-    async fn exists(config: &Self::Config, namespace: &str) -> Result<bool, IndexedDbStoreError> {
-        Ok(Self::connect(config, "")
-            .await?
-            .database
-            .object_store_names()
-            .any(|x| x == namespace))
+    async fn exists(_config: &Self::Config, _namespace: &str) -> Result<bool> {
+        // IndexedDB will create the database if it doesn't exist, so let's pretend it
+        // always exists.
```
ma2bd, line 308: how about the namespace though?

Twey: The namespace and the database are equivalent here.  ‘Namespace’ is the `linera-views` concept that used to be implemented in IndexedDB as an object store name, but is now implemented using a database name.  I tagged some of them so that when we get IndexedDB v3 support we can pick them out of the full set of available databases.  It's unfortunate that this means that `exists` doesn't imply the namespace is listed by `list_all`.

I used to use object stores (IndexedDB ‘tables’) as namespace, since they can be listed, and a constant database name.  But object stores aren't versioned, and that messes up versioning for the database (basically we have to use a constantly-incrementing version as a hack to support dynamically creating object stores, since the schema gets disconnected from the version, and then we can't make use of the version for… actual versioning).

.gitignore
Twey: Zed likes to keep per-project config together in `.zed`.  We `.gitignore` tool-specific files in general.

Twey: Oh I see!  This happened while I was merging.  But it fits the general structure of our `.gitignore`.

web/@linera/client/src/storage.rs
Twey: Actually this is the wrong task anyway!  Should have been https://github.com/linera-io/linera-protocol/issues/5064.

.gitignore
ma2bd: (I was asking about the added comment)

Twey: Right, I didn't realize that the content had been merged away. 😀

Twey: Right, I didn't realize that the content had been merged away. 😀

PR #5272, opened 2026-01-15
[testnet] Remove `let _ =`. (#5261)

afck:
Backport of #5261.

## Motivation

`let _ =` can hide unawaited futures and discarded errors.

## Proposal

Add to `CONTRIBUTING.md` that we want to avoid this.

Fix the occurrences. Some were unnecessary in the first place; some were missing `unwrap` or `?`s in tests.

In some cases I added `.ok()`, replacing the `Result` with an `Option`. This is slightly better because I believe the compiler would still complain if it e.g. were a _nested_ result.

## Test Plan

CI

## Release Plan

- Nothing to do.

## Links

- PR to main: #5261

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -67,7 +67,7 @@ Contributions should generally follow the [Rust API guidelines](https://rust-lan
 * Re-exports (`pub use foo = bar::foo`) should be limited to definitions that would be private otherwise.
 
 * Avoid `let _ = ...` to discard values, because it can hide important information like
-  unawaited futures or unhandled errors.
+  unawaited futures or unhandled errors. `let _x =` should only be used for RAII guards.
 
 
 ## Formatting and linting
--- a/examples/fungible/src/contract.rs
+++ b/examples/fungible/src/contract.rs
@@ -43,7 +43,7 @@ impl Contract for FungibleTokenContract {
 
     async fn instantiate(&mut self, state: Self::InstantiationArgument) {
         // Validate that the application parameters were configured correctly.
-        let _parameters = self.runtime.application_parameters();
+        self.runtime.application_parameters();
 
         let mut total_supply = Amount::ZERO;
         for value in state.accounts.values() {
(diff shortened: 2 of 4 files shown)


deuszx (approved):

examples/fungible/src/contract.rs
```diff
@@ -43,7 +43,7 @@ impl Contract for FungibleTokenContract {
 
     async fn instantiate(&mut self, state: Self::InstantiationArgument) {
         // Validate that the application parameters were configured correctly.
-        let _ = self.runtime.application_parameters();
+        let _parameters = self.runtime.application_parameters();
```
ma2bd: We also don't want useless `let _x = `. Those actually impact stack allocation.

linera-client/src/unit_tests/chain_listener.rs
```diff
@@ -57,7 +57,8 @@ impl chain_listener::ClientContext for ClientContext {
         timestamp: Timestamp,
         epoch: Epoch,
     ) -> Result<(), Error> {
-        let _ = self
+        // Ignore if chain already exists in wallet; test mock doesn't care.
+        let _insert_result = self
```
ma2bd: why is this better?

CONTRIBUTING.md
```diff
@@ -66,6 +66,9 @@ Contributions should generally follow the [Rust API guidelines](https://rust-lan
 
 * Re-exports (`pub use foo = bar::foo`) should be limited to definitions that would be private otherwise.
 
+* Avoid `let _ = ...` to discard values, because it can hide important information like
```
ma2bd, line 69: Also `let _x =` should only be used for RAII guards

examples/fungible/src/contract.rs
deuszx: Yes, I had the same thought initially – it's not really providing any more type checks but I think the point here is that it improves readability

afck: Right, it's not even necessary; I removed it, and added to `CONTRIBUTING.md`.

PR #5273, opened 2026-01-16
Backport multiple Storage API changes

deuszx:
Backports #5233 #5256 #5258 #5267

## Motivation

Backport new storage API changes.

## Proposal

Backport all PRs.

## Test Plan

CI

## Release Plan

## Links

 #5233 #5256 #5258 #5267

### DIFF
--- a/linera-core/src/updater.rs
+++ b/linera-core/src/updater.rs
@@ -769,16 +769,7 @@ where
             .read_certificates_by_heights(chain_id, &heights)
             .await?;
 
-        // Verify we got all certificates
-        let certificates: Vec<_> = certificates_by_height.into_iter().flatten().collect();
-
-        if certificates.len() != heights.len() {
-            return Err(chain_client::Error::InternalError(
-                "failed to read all certificates for heights",
-            ));
-        }
-
-        Ok(certificates)
+        Ok(certificates_by_height.into_iter().flatten().collect())
     }
 
     /// Initializes a new chain on the validator by sending the chain description and dependencies.


linera-core/src/updater.rs
```diff
@@ -711,24 +711,15 @@ where
...
-                .get_block_hashes(chain_id, vec![height])
+            self.read_certificates_for_heights(chain_id, vec![height])
                 .await?
                 .into_iter()
                 .next()
                 .ok_or_else(|| {
                     chain_client::Error::InternalError(
-                        "send_chain_information called with invalid target_block_height",
+                        "failed to read latest certificate for height sync",
                     )
```
MathieuDutSik, line 721: I wonder if that error could ever occur.
When reding a vector of 1 element that it returns a vector of 0 elements. So, unwrap() might be adequate. Or a `read_certificate_for_height`.

linera-core/src/updater.rs
```diff
@@ -781,6 +757,58 @@ where
         Ok(info)
     }
 
+    /// Reads certificates for the given heights from storage.
+    ///
+    /// First attempts to use the optimized `read_certificates_by_heights` method.
+    /// Falls back to the traditional `get_block_hashes` + `read_certificates` approach
+    /// if the direct height lookup doesn't return all certificates.
```
MathieuDutSik: I understand why this is done that way: The storage of the Testnet Conway is not migrated. Therefore, we need to deal with a system that has missing blocks. That is an interesting design decision that should be documented in the PR description. The statement "Backport all PRs" makes it look like it is a straightforward backport. It is definitely not the case.

Also, the use of "CI" for testing the functionality appears inadequate to me. If you work with CI, you will not test the fallback functionality when some storage is missing.

linera-rpc/src/grpc/client.rs
```diff
@@ -444,12 +444,12 @@ impl ValidatorNode for GrpcClient {
         chain_id: ChainId,
         heights: Vec<BlockHeight>,
     ) -> Result<Vec<ConfirmedBlockCertificate>, NodeError> {
-        let mut missing = heights;
+        let mut missing: BTreeSet<BlockHeight> = heights.into_iter().collect();
```
MathieuDutSik: We have a convention of avoiding type annotation. Here you can work with
```rust
let mut missing = heights.into_iter().collect::<BTreeSet<_>>();
```

linera-service/src/proxy/grpc.rs
```diff
@@ -376,6 +377,100 @@ where
...
+        let chain_info_request =
+            ChainInfoQuery::new(chain_id).with_sent_certificate_hashes_by_heights(heights.clone());
+
+        let chain_info_response = self
+            .handle_chain_info_query(Request::new(chain_info_request.try_into()?))
+            .await?;
+
+        let chain_info_result = chain_info_response.into_inner();
+
+        let hashes: Vec<linera_base::crypto::CryptoHash> = match chain_info_result.inner {
```
MathieuDutSik: Type annotation should be avoided.

linera-service/src/proxy/grpc.rs
```diff
@@ -376,6 +377,100 @@ where
...
+                )));
+            }
+            None => {
+                return Err(Status::internal("Empty chain info result"));
+            }
+        };
+
+        // Write back the height->hash indices we learned from the fallback
+        let indices: Vec<(BlockHeight, linera_base::crypto::CryptoHash)> =
+            heights.into_iter().zip(hashes.iter().copied()).collect();
```
MathieuDutSik: Type annotation can be avoided.

linera-service/src/proxy/grpc.rs
```diff
@@ -376,6 +377,100 @@ where
...
+        self.0
+            .storage
+            .write_certificate_height_indices(chain_id, &indices)
+            .await
+            .map_err(Self::view_error_to_status)?;
```
MathieuDutSik: If we had to use the fallback, we try to avoid it next time. Nice.

linera-service/src/proxy/grpc.rs
```diff
@@ -376,6 +377,100 @@ where
...
+    async fn collect_raw_certificates_by_hashes(
+        &self,
+        hashes: Vec<linera_base::crypto::CryptoHash>,
+    ) -> Result<Vec<RawCertificate>, Status> {
+        let mut limiter: GrpcMessageLimiter<linera_chain::types::Certificate> =
+            GrpcMessageLimiter::new(GRPC_CHUNKED_MESSAGE_FILL_LIMIT);
+
+        let mut result = vec![];
+
+        for batch in hashes.chunks(100) {
```
MathieuDutSik: It is a magic constant.
It should be put somewhere, not hidden in the code.

linera-service/src/proxy/grpc.rs
```diff
@@ -376,6 +377,100 @@ where
...
+            let certificates: Vec<(Vec<u8>, Vec<u8>)> = self
```
MathieuDutSik: Avoid type annotation (this case is easy).

linera-service/src/proxy/grpc.rs
```diff
@@ -634,37 +729,28 @@ where
...
-            };
-            for certificate in certificates {
-                if grpc_message_limiter.fits::<Certificate>(certificate.clone().into())? {
-                    returned_certificates.push(linera_chain::types::Certificate::from(certificate));
-                } else {
-                    break 'outer;
-                }
-            }
-        }
+        let certificates: Vec<linera_chain::types::Certificate> = raw_certificates
```
MathieuDutSik: Avoid type annotation.
Also avoid type such as `linera_chain::types::Certificate` and instead put a use statement on top of the file.

linera-service/src/proxy/grpc.rs
```diff
@@ -727,70 +816,48 @@ where
...
+        // First, try the direct height-based lookup
+        let raw_certificates_by_height = self
+            .0
+            .storage
+            .read_certificates_by_heights_raw(chain_id, &heights)
+            .await
+            .map_err(Self::view_error_to_status)?
+            .into_iter()
+            .flatten()
+            .collect::<Vec<(Vec<u8>, Vec<u8>)>>();
```
MathieuDutSik: We have almost the same code section duplicated.
One time with `read_certificates_by_height`, one time with a `_raw` version. Would it be possible to avoid this code duplication?

MathieuDutSik (commented):
The PR does more than its description says.
The use of `_fallback` schemes exposes new failure scenario that should be tested locally before merging the PR.

The code that might be of interest for that is https://github.com/MathieuDutSik/linera_end_to_end_tests/tree/main/test_migration_database_schema

It is for the migration, but it could be adapted to the case here:
* Create a non-trivial contract with the code of `testnet_conway` (so without those modifications)
* Run operations with the code of this PR and so exercise the `_fallback` code.

deuszx:
> The PR does more than its description says. The use of `_fallback` schemes exposes new failure scenario that should be tested locally before merging the PR.
> 
> The code that might be of interest for that is https://github.com/MathieuDutSik/linera_end_to_end_tests/tree/main/test_migration_database_schema
> 
> It is for the migration, but it could be adapted to the case here:
> 
> * Create a non-trivial contract with the code of `testnet_conway` (so without those modifications)
> * Run operations with the code of this PR and so exercise the `_fallback` code.

Thanks for the review. You make an important point – there's no need for maintaining the `_fallback` scenarios for `main` branch! I will simplify the PR.

linera-core/src/updater.rs
deuszx: This is now removed from this PR.

linera-service/src/proxy/grpc.rs
deuszx: It is but it also was here before this PR.

linera-service/src/proxy/grpc.rs
deuszx: I tried to deduplicate in the original PR but it's not trivial b/c we use different return types and different logic to check whether the type "fits" in the gRPC message.

linera-core/src/updater.rs
deuszx: `.unwrap()` won't give a helpful message and there's no `read_certificate_for_height`. This way the single-element query fails the same way as multiple-elements one.

linera-storage/src/db_storage.rs
```diff
@@ -898,14 +917,106 @@ where
...
+            .await?
+            .into_iter()
+            .zip(unique_hashes)
+        {
+            if let Some(idx_list) = indices.get(&hash) {
+                for &index in idx_list {
+                    result[index] = raw_cert.clone();
+                }
+            } else {
+                // This should not happen, but log a warning if it does.
```
afck, line 986: Or an error?

linera-storage/src/db_storage.rs
```diff
@@ -898,14 +917,106 @@ where
...
+                tracing::warn!(
+                    hash=?hash,
```
```suggestion
                    %hash,
```

afck (approved):

linera-rpc/src/grpc/client.rs
deuszx: Done (and others)

linera-storage/src/db_storage.rs
```diff
@@ -898,14 +917,103 @@ where
...
+            .into_iter()
+            .zip(unique_hashes)
+        {
+            if let Some(idx_list) = indices.get(&hash) {
+                for &index in idx_list {
+                    result[index] = raw_cert.clone();
+                }
+            } else {
+                // This should not happen, but log a warning if it does.
+                tracing::error!(?hash, "certificate hash not found in indices map",);
```
afck, line 987:
```suggestion
                tracing::error!(?hash, "certificate hash not found in indices map");
```

deuszx: if `cargo fmt` didn't fix it, I am reluctant as well :D I don't want to be paying attention to formatting errors like this.

afck: (I think it doesn't for macros, because it doesn't know whether a trailing comma matters for a specific macro. Anyway, not important!)

linera-core/src/updater.rs
MathieuDutSik: Ok, It is just annoying to have this function called on a single element and we have to handle the totally not realistic case where it returns a vector of length 0 (but length 2 would have been fine).

I think the right way would be to have a function `read_certificate_for_height`.

MathieuDutSik:
> Thanks for the review. You make an important point – there's no need for maintaining the `_fallback` scenarios for `main` branch! I will simplify the PR.

Oh, I see. The backport is in the other direction. Yes, we do not need the `_fallback`.

linera-service/src/proxy/grpc.rs
MathieuDutSik: Yeah. Unfortunately.

linera-core/src/updater.rs
deuszx: We don't have a function like this on the `storage` either – we'd have to add a special one on `Storage` and it'd duplicate the logic (or simply piggyback) on the one that requests multiple certificates.

What I'm getting at is that this duplication/weirdness would have to happen at some level - either here or in the storage. By doing it here we have fewer lines of code (we don't need to add `read_certificate_for_height` (singular) to all the layers below).

PR #5275, opened 2026-01-16
fix(explorer): use relative API URL by default in Dockerfile

eldios:
## Summary

- Change `VITE_API_BASE_URL` default from `http://localhost:3002/api` to `/api`
- Fixes CORS errors in production deployments where frontend and backend are served from the same domain via ingress

## Problem

The explorer frontend was trying to fetch from `http://localhost:3002/api` in production, causing CORS errors:
```
Access to fetch at 'http://localhost:3002/api/blocks?limit=50&offset=0' from origin 'https://explorer.testnet-conway.linera.net' has been blocked by CORS policy
```

This happened because the `localhost` URL was baked into the frontend bundle at Docker build time.

## Solution

Default to relative URL `/api` which nginx/ingress proxies correctly to the backend. For local development, the original URL can still be passed via:
```
--build-arg VITE_API_BASE_URL=http://localhost:3002/api
```

## Test plan

- [ ] Rebuild explorer image without passing `VITE_API_BASE_URL` build arg
- [ ] Deploy to testnet-conway
- [ ] Verify explorer loads blocks without CORS errors

### DIFF
--- a/docker/Dockerfile.explorer
+++ b/docker/Dockerfile.explorer
@@ -3,7 +3,9 @@
 FROM node:18-alpine AS frontend-builder
 
 # Build argument for API base URL (injected at build time)
-ARG VITE_API_BASE_URL=http://localhost:3002/api
+# Default to relative URL for production deployments (proxied via ingress)
+# For local dev, override with: --build-arg VITE_API_BASE_URL=http://localhost:3002/api
+ARG VITE_API_BASE_URL=/api
 
 RUN apk add --no-cache python3 make g++
 


docker/Dockerfile.explorer
```diff
@@ -3,7 +3,9 @@
 FROM node:18-alpine AS frontend-builder
 
 # Build argument for API base URL (injected at build time)
-ARG VITE_API_BASE_URL=http://localhost:3002/api
+# Default to relative URL for production deployments (proxied via ingress)
+# For local dev, override with: --build-arg VITE_API_BASE_URL=http://localhost:3002/api
+ARG VITE_API_BASE_URL=/api
```
deuszx, line 8: I think the correct fix will be:
```
ARG VITE_API_BASE_URL=http://localhost:3002/api
ENV VITE_API_BASE_URL=$VITE_API_BASE_URL
```
 
And then you can overwrite the `VITE_API_BASE_URL` during build if necessary:
```
docker run -e VITE_API_BASE_URL=/api
```

THe current proposal won't work when running this docker-compose stack locally as it will build with `VITE_API_BASE_URL=/api`

eldios: no unfortunately that won't work because Vite bakes ENV vars into the JS bundle at build time. 
By runtime, the value is already hardcoded in the JavaScript - so setting container env vars has no effect

basically when building with docker OR docker compose you just have to add the build-arg as specified in the comment, and that's it.

eldios: local kubernetes clusters like kind would also work btw, I think it's just straight docker or docker compose that would fail

deuszx: > no unfortunately that won't work because Vite bakes ENV vars into the JS bundle at build time.
By runtime, the value is already hardcoded in the JavaScript - so setting container env vars has no effect

Yes, that's why I wrote you should overwrite it during build-time.

> I think it's just straight docker or docker compose that would fail

And that's a problem – this docker compose setup is meant to be used locally as well.

I can do it for local deployments to unblock you.

deuszx (approved):

PR #5278, opened 2026-01-17
make serialization of wallet stable

ma2bd:
## Motivation

Allow checking `wallet.json` in git repositories

## Proposal

* Overwrite `serde::Serialize`
* Add tests to ensure consistent and stable (de)serialization

## Test Plan

CI

## Release Plan

    - be released in a new SDK,

### DIFF
--- a/linera-core/src/environment/wallet/memory.rs
+++ b/linera-core/src/environment/wallet/memory.rs
@@ -18,11 +18,12 @@ pub struct Memory(papaya::HashMap<ChainId, Chain>);
 /// Custom Serialize implementation that ensures stable ordering by sorting entries by ChainId.
 impl Serialize for Memory {
     fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
-        let mut items: Vec<_> = self.0.pin().iter().map(|(k, v)| (*k, v.clone())).collect();
+        let guard = self.0.pin();
+        let mut items: Vec<_> = guard.iter().collect();
         items.sort_by_key(|(k, _)| *k);
         let mut map = serializer.serialize_map(Some(items.len()))?;
         for (k, v) in items {
-            map.serialize_entry(&k, &v)?;
+            map.serialize_entry(k, v)?;
         }
         map.end()
     }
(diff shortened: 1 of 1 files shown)


linera-core/src/environment/wallet/memory.rs
```diff
@@ -11,9 +12,22 @@ use super::{Chain, Wallet};
...
 /// This can be used as-is as an ephemeral wallet for testing or ephemeral clients, or as
 /// a building block for more complex wallets that layer persistence on top of it.
-#[derive(Default, Clone, serde::Serialize, serde::Deserialize)]
+#[derive(Default, Clone, serde::Deserialize)]
 pub struct Memory(papaya::HashMap<ChainId, Chain>);
 
+/// Custom Serialize implementation that ensures stable ordering by sorting entries by ChainId.
+impl Serialize for Memory {
+    fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        let mut items: Vec<_> = self.0.pin().iter().map(|(k, v)| (*k, v.clone())).collect();
```
afck: (Nit: If we keep the `pin()` around, it shouldn't be necessary to clone the keys and values. We could sort and serialize the references.)

linera-core/src/environment/wallet/memory.rs
```diff
@@ -116,3 +130,68 @@ impl Wallet for Memory {
...
+
+        // Insert in non-sorted order
+        memory.insert(id3, make_chain(3));
+        memory.insert(id1, make_chain(1));
+        memory.insert(id2, make_chain(2));
+
+        // Serialize multiple times and verify output is identical
+        let json1 = serde_json::to_string(&memory).unwrap();
+        let json2 = serde_json::to_string(&memory).unwrap();
+        let json3 = serde_json::to_string(&memory).unwrap();
```
afck: I imagine `papaya::HashMap` behaves like the standard `HashMap`? And there iteration order _is_ consistent within the same process, isn't it?

Maybe we should rather assert that the output is sorted?

linera-core/src/environment/wallet/memory.rs
ma2bd: right. Let me have a look

linera-core/src/environment/wallet/memory.rs
ma2bd: done

linera-core/src/environment/wallet/memory.rs
ma2bd: done

PR #5283, opened 2026-01-19
Rewrite explorer backend to Rust

deuszx:
## Motivation

TypeScript (JavaScript) isn't the best technology to use when we want to support multiple concurrent queries.

## Proposal

Rewrite the explorer server to Rust.

## Test Plan

Manual

## Release Plan

- Backport to `main`.

## Links


deuszx:
> The code looks reasonable to me. Seems like we're keeping the TS code around?

I wanted to keep it before we test the Rust version fully. Once we confirm it works (for that I need to deploy it), I'll create the PR with removal.

PR #5285, opened 2026-01-19
Chain aliases

### LINKED ISSUE #5280: Consider giving chains a local name in wallet storage
Automatically generated names could be `admin` and `user-{0,1,2}` but `linera wallet {follow-chain,request-chain}` could take an extra name argument. We would need a way to rename chains. If we restrict names to be shorter than chain ID, commands could accept a name instead of a chain ID as an alias.

afck:
## Motivation

It's inconvenient to copy and paste chain IDs.

## Proposal

Give aliases to chains in the wallet, and support using these on the CLI instead of chain IDs (also as part of accounts). They can also be used in node service URLs.

Aliases are at most 63 characters long, i.e. strictly shorter than a chain ID, so that no chain ID is a valid alias.

The admin chain is `admin` by default, other chains are `user-0`, `user-1`, etc., unless specified. The root chains in the genesis wallet are `root-0`, `root-1`, etc.

## Test Plan

A test was added. The `counter` example README was changed to use aliases in GraphQL URLs.

## Release Plan

    - be released in a new SDK.

## Links

- Closes #5280.

### DIFF
--- a/linera-core/src/environment/wallet/mod.rs
+++ b/linera-core/src/environment/wallet/mod.rs
@@ -15,6 +15,40 @@ use crate::{client::PendingProposal, data_types::ChainInfo};
 mod memory;
 pub use memory::Memory;
 
+/// The maximum length of a chain name. Chain names must be shorter than a chain ID
+/// (which is 64 hex characters) to avoid ambiguity.
+pub const MAX_CHAIN_NAME_LENGTH: usize = 63;
+
+/// Validates a chain name, returning an error if it's invalid.
+///
+/// Chain names must:
+/// - Not be empty
+/// - Be at most [`MAX_CHAIN_NAME_LENGTH`] characters
+/// - Not contain colons (used as separator in account strings)
+pub fn validate_chain_name(name: &str) -> Result<(), ChainNameError> {
+    if name.is_empty() {
+        return Err(ChainNameError::Empty);
+    }
+    if name.len() > MAX_CHAIN_NAME_LENGTH {
+        return Err(ChainNameError::TooLong(name.len()));
+    }
+    if name.contains(':') {
+        return Err(ChainNameError::ContainsColon);
+    }
+    Ok(())
+}
+
+/// Error returned when a chain name is invalid.
+#[derive(Clone, Debug, thiserror::Error)]
+pub enum ChainNameError {
+    #[error("chain name cannot be empty")]
+    Empty,
+    #[error("chain name is too long ({0} characters, maximum is {MAX_CHAIN_NAME_LENGTH})")]
+    TooLong(usize),
+    #[error("chain name cannot contain colons (used as separator in account strings)")]
+    ContainsColon,
+}
+
 #[derive(Clone, serde::Serialize, serde::Deserialize)]
 pub struct Chain {
     /// The name of this chain in the wallet.
--- a/linera-service/src/wallet.rs
+++ b/linera-service/src/wallet.rs
@@ -14,12 +14,9 @@ use linera_base::{
 use linera_client::config::GenesisConfig;
 use linera_core::wallet;
 use linera_persistent as persistent;
+pub use wallet::MAX_CHAIN_NAME_LENGTH;
 
 use crate::cli::command::ChainIdOrName;
-
-/// The maximum length of a chain name. Chain names must be shorter than a chain ID
-/// (which is 64 hex characters) to avoid ambiguity.
-pub const MAX_CHAIN_NAME_LENGTH: usize = 63;
 
 #[derive(serde::Serialize, serde::Deserialize)]
 struct Data {
@@ -299,19 +296,15 @@ impl Wallet {
         chain_id: ChainId,
         name: Option<String>,
     ) -> anyhow::Result<()> {
-        // Determine the new name.
+        // Determine the new name. User-provided names are validated; auto-generated names
+        // are trusted.
         let new_name = match name {
-            Some(name) => name,
+            Some(name) => {
+                wallet::validate_chain_name(&name)?;
+                name
+            }
             None => self.next_default_name(chain_id).await?,
         };
-        // Validate the name.
-        anyhow::ensure!(
-            new_name.len() <= MAX_CHAIN_NAME_LENGTH,
-            "chain name is too long ({} characters, maximum is {})",
-            new_name.len(),
-            MAX_CHAIN_NAME_LENGTH
-        );
-        anyhow::ensure!(!new_name.is_empty(), "chain name cannot be empty");
         // Check if the name is already used by another chain.
         if let Some(existing_chain_id) = self.resolve_chain_name(&new_name) {
             anyhow::ensure!(
--- a/linera-client/src/client_context.rs
+++ b/linera-client/src/client_context.rs
@@ -466,8 +466,12 @@ impl<Env: Environment> ClientContext<Env> {
         timestamp: Timestamp,
         epoch: Epoch,
     ) -> Result<(), Error> {
+        // User-provided names are validated; auto-generated names are trusted.
         let chain_name = match name {
-            Some(provided_name) => provided_name,
+            Some(provided_name) => {
+                wallet::validate_chain_name(&provided_name)?;
+                provided_name
+            }
             None => self.default_chain_name(chain_id).await?,
         };
         let chain = wallet::Chain::new(chain_name, owner, epoch, timestamp);
--- a/linera-client/src/error.rs
+++ b/linera-client/src/error.rs
@@ -4,7 +4,7 @@
 use linera_base::{
     crypto::ValidatorPublicKey, data_types::NetworkDescription, identifiers::ChainId,
 };
-use linera_core::node::NodeError;
+use linera_core::{node::NodeError, wallet::ChainNameError};
 use linera_version::VersionInfo;
 use thiserror_context::Context;
 
@@ -37,6 +37,8 @@ pub(crate) enum Inner {
     Arithmetic(#[from] linera_base::data_types::ArithmeticError),
     #[error("incorrect chain ownership")]
     ChainOwnership,
+    #[error("invalid chain name: {0}")]
+    ChainName(#[from] ChainNameError),
     #[cfg(not(web))]
     #[error("Benchmark error: {0}")]
     Benchmark(#[from] BenchmarkError),
--- a/linera-service/src/cli/command.rs
+++ b/linera-service/src/cli/command.rs
@@ -18,6 +18,7 @@ use linera_client::{
     },
     util,
 };
+use linera_core::wallet;
 use linera_rpc::config::CrossChainConfig;
 
 use crate::{cli::validator, task_processor::parse_operator};
@@ -98,13 +99,8 @@ impl FromStr for ChainIdOrName {
         if let Ok(chain_id) = s.parse::<ChainId>() {
             return Ok(ChainIdOrName::ChainId(chain_id));
         }
-        // Otherwise treat as a name, but validate the length.
-        anyhow::ensure!(
-            s.len() <= crate::wallet::MAX_CHAIN_NAME_LENGTH,
-            "chain name is too long ({} characters, maximum is {})",
-            s.len(),
-            crate::wallet::MAX_CHAIN_NAME_LENGTH
-        );
+        // Otherwise treat as a name, but validate it.
+        wallet::validate_chain_name(s)?;
         Ok(ChainIdOrName::Name(s.to_string()))
     }
 }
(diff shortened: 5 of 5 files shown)


afck:
I have some suggestions for cleanups, but they involve bigger refactorings, so I'll merge this and propose them as follow-ups.

afck:
Converting back to draft: We need to also support `Account`s using aliases.

linera-core/src/environment/wallet/mod.rs
```diff
@@ -15,51 +15,62 @@ use crate::{client::PendingProposal, data_types::ChainInfo};
...
-            timestamp: info.timestamp,
-            pending_proposal: None,
-            epoch: Some(info.epoch),
-        }
+/// Validates a chain name, returning an error if it's invalid.
+///
+/// Chain names must:
+/// - Not be empty
+/// - Be at most [`MAX_CHAIN_NAME_LENGTH`] characters
+/// - Not contain colons (used as separator in account strings)
```
bart-linera, line 27: I wonder if it wouldn't be a better idea to have a whitelisted set of characters instead (like: alphanumeric + hyphen, underscore and maybe a few more).

Although as long as the feature is purely client-side, maybe it's okay.

bart-linera (approved):
I like this feature a lot! Maybe we could introduce user-defined aliases for applications as well?

CLI.md
```diff
@@ -333,7 +336,7 @@ Display who owns the chain, and how the owners work together proposing blocks
 
 ###### **Options:**
 
-* `--chain-id <CHAIN_ID>` — The ID of the chain whose owners will be changed
+* `--chain-id <CHAIN_ID>` — The ID or name of the chain whose owners will be shown
```
Twey, line 339: If it's no longer just a chain ID it's incorrect to call the flag `--chain-id` and the exemplar `CHAIN_ID`.  Maybe `--chain <CHAIN>`?  (and below)

linera-core/src/environment/wallet/mod.rs
afck: Good idea, I'll use a whitelist instead. :+1:

CLI.md
```diff
@@ -436,7 +439,7 @@ NOTE: The balance does not reflect messages that have not been synchronized from
 
 ###### **Arguments:**
 
-* `<ACCOUNT>` — The account to query, written as `CHAIN-ID:OWNER` or simply `CHAIN-ID` for the chain balance. By default, we read the chain balance of the default chain in the wallet
+* `<ACCOUNT>` — The account to query, written as `CHAIN-ID:OWNER`, `CHAIN-NAME:OWNER`, or simply `CHAIN-ID`/`CHAIN-NAME` for the chain balance. By default, we read the chain balance of the default chain in the wallet
```
Twey, line 442: Let's use the word `CHAIN` to talk about chain references, and explain above that it may be a chain ID or name.

examples/counter/README.md
```diff
@@ -98,16 +95,15 @@ First, a node service for the current wallet has to be started:
...
-Clicking the printed URL will take you to a
-[GraphiQL](https://www.gatsbyjs.com/docs/how-to/querying-data/running-queries-with-graphiql/)
+Navigate to <http://localhost:8080/chains/my-chain/applications/$LINERA_APPLICATION_ID>
+to access a [GraphiQL](https://www.gatsbyjs.com/docs/how-to/querying-data/running-queries-with-graphiql/)
 environment connected to your app.
 
 To get the current value of `counter`, you can run the query:
 
-```gql,uri=http://localhost:8080/chains/$CHAIN/applications/$LINERA_APPLICATION_ID
+```gql,uri=http://localhost:8080/chains/my-chain/applications/$LINERA_APPLICATION_ID
```
Twey, line 106: Probably not for this PR, but it seems strange to allow naming chains but not applications.

examples/counter/README.md
```diff
@@ -43,18 +43,15 @@ Enable logs for user applications:
...
 export LINERA_WALLET="$LINERA_TMP_DIR/wallet.json"
 export LINERA_KEYSTORE="$LINERA_TMP_DIR/keystore.json"
 export LINERA_STORAGE="rocksdb:$LINERA_TMP_DIR/client.db"
 
 linera wallet init --faucet $LINERA_FAUCET_URL
-
-INFO=($(linera wallet request-chain --faucet $LINERA_FAUCET_URL))
-CHAIN="${INFO[0]}"
-OWNER="${INFO[1]}"
+linera wallet request-chain --faucet $LINERA_FAUCET_URL --name my-chain
```
Twey, line 54: This is the default chain (named `default`) so it's not really necessary to mention `--name` here.

linera-client/src/chain_listener.rs
```diff
@@ -79,8 +79,8 @@ pub trait ClientContext {
 
     fn client(&self) -> &Arc<linera_core::client::Client<Self::Environment>>;
 
-    fn admin_chain(&self) -> ChainId {
-        self.client().admin_chain()
+    fn admin_id(&self) -> ChainId {
```
Twey, line 82: Nit: I think I like this less than `admin_chain`.  The fact that it identifies the admin chain is more important than the fact that the admin chain is represented by an ID (which is anyway already encoded in the type).  An ‘admin ID’ could (to the uninitiated reader) refer to a bunch of things, such as an administrative user account.

linera-client/src/chain_listener.rs
```diff
@@ -101,29 +101,41 @@ pub trait ClientContext {
...
-            let follow_only = chain.is_follow_only();
+            {
+                None => (None, BlockHeight::ZERO, None, None),
+                Some(chain) => (
+                    chain.block_hash,
+                    chain.next_block_height,
+                    chain.pending_proposal,
+                    chain.owner,
+                ),
+            };
```
Twey, line 118: Nit: `unwrap_or_default()` (`Default` is implemented for quadruples whose components are `Default`)

examples/counter/README.md
afck: Isn't it named `user-0` by default? I wouldn't make the name `default` because the user could change the default chain without changing the name.

We could use `user-0` in the README, of course, but I think that would illustrate the feature less clearly?

linera-client/src/chain_listener.rs
Twey: (and below)

afck: I did this for consistency because we use `admin_id` in other places, but I can change it the other way round. Or should it be `admin_chain_id`?

linera-service/src/wallet.rs
```diff
@@ -239,6 +243,84 @@ impl Wallet {
...
+    ///
+    /// Looks up chains by their stored name.
+    pub fn resolve_chain_name(&self, name: &str) -> Option<ChainId> {
+        for (chain_id, chain) in self.items() {
+            if chain.name == name {
+                return Some(chain_id);
+            }
+        }
+        None
+    }
```
Twey, line 261: This is linear in number of chains!  I think I'd rather double-key chains by an `enum ChainRef { Id(#[from] ChainId), Name(#[from] ChainName) }` (and make sure that chain names aren't stringly typed).

web/@linera/client/src/faucet.rs
```diff
@@ -50,13 +50,12 @@ impl Faucet {
...
-        wallet.chains.insert(
-            chain_id,
-            wallet::Chain {
-                owner: Some(owner),
-                ..description.into()
-            },
-        );
+        let name = wallet::next_default_chain_name(&*wallet.chains)
+            .await
+            .expect("infallible");
```
Twey, line 55: The `expect` string is supposed to explain _why_ it's infallible.

linera-core/src/environment/wallet/mod.rs
Twey: Speaking as someone who's had to deal with OWASP, disallowing characters from values, especially user-supplied values, because they have special meaning in other contexts is a slippery slope.  I'd rather, for example:
- introduce a syntax for names that can't conflict with IDs, for example `@name` à la git (this leads to `account@@name` for the `account@chain` syntax we use on the Web, which is probably fine)
- add escaping syntax to account strings for the special characters that account strings introduce

examples/counter/README.md
Twey: Oh you're right!  Sorry, I misremembered the PR description, which I read before bed last night. :sweat_smile: 

I would really like to have the default chain be nameable as `default`, and I think this example (which uses only one chain anyway) maybe isn't as good as some of the other examples to show off the chain naming feature.

linera-client/src/chain_listener.rs
Twey: My preference is `admin_chain` because I think that the fact that the chain is being represented by a chain ID is an implementation detail.  But I don't mind `admin_chain_id` too much, and it may be a less controversial choice.

examples/counter/README.md
afck: Well, I wanted to use one where I could really get rid of the environment variable containing the chain ID. But we do need that in all examples where we put it into a GraphQL query. (We probably shouldn't automatically replace chain aliases there…)

afck: I guess we could disallow the name `default` and have it always refer to the default chain (in addition to its given name)?

web/@linera/client/src/faucet.rs
Twey: If this is actually `Infallible` you can use: `.unwrap_or_else(|never| match never {})`

linera-client/src/chain_listener.rs
afck: It's just that we already use names like `chain` and `admin_chain` for chain state views, rather than IDs, so maybe it's clearer?

linera-service/src/wallet.rs
afck: How would you represent it in JSON? Still indexed by the ID, with the name as a field (as it is now), I guess?

web/@linera/client/src/faucet.rs
afck: Or just `?`, actually.

linera-core/src/environment/wallet/mod.rs
afck: So which set of characters would you allow? And what syntax for escaping? Do you have something specific in mind where we don't reinvent the wheel?

linera-client/src/chain_listener.rs
Twey: Idealistically speaking I'd have that mean that `admin_chain()` returns a chain view and then the ID comes from `admin_chain().id()` (with `admin_chain()` lazily loaded so that doesn't imply a performance hit).  But to be sure if it helps with disambiguation for now then `admin_chain_id()` is fine by me.

linera-service/src/wallet.rs
Twey: We could keep it as two maps (`ChainId ⇒ Option<Chain>` and `ChainName ⇒ Option<ChainId>`) in the JSON and even the memory wallet, in fact.  I just mean in the `Wallet` trait methods.

linera-client/src/chain_listener.rs
afck: But in most places in the code we don't want to use chain state views (should be confined to the chain actor and tests), and also internally we're not loading the chain state view to obtain its ID; it's the other way round.

linera-core/src/environment/wallet/mod.rs
bart-linera: One potentially significant drawback of a whitelist I can think of is that we already have Chinese users, they might want to use their characters in chain names, and there isn't really any good reason for preventing them from doing that... but whitelisting the whole CJK section of Unicode might be a bit of an issue :sweat_smile:

afck: That's true.

Either way I'm definitely in favor of the `account@chain` syntax as opposed to `chain:account`.

Twey: > So which set of characters would you allow?

All of them! :grin: 

> And what syntax for escaping?

Given the expected usages I'd probably suggest URI escaping, which has `@` become `%40`.

Twey: Yes we've been meaning to replace the colon-based syntax for a bit already.

examples/counter/README.md
Twey: I think rather than introducing multiple names for chains (right now…) we could just have the default name for a chain be `default` if it's about to be made default in the wallet.

linera-service/src/wallet.rs
Twey: Sorry, I thought a little harder about this and I'm not sure (especially apropos our conversation yesterday) I like the idea of pushing this down into `linera-core` at all.  If we do the translation at the `linera-service` level (with an extra map in the concrete JSON, to avoid the linear scan) there's no need to change the `linera-core` interfaces.

afck: Right, removing the `name` field from `Chain` again should be the first step. I can then try to move the second map out of `Memory` again, and out of `linera-core`.

linera-client/src/chain_listener.rs
Twey: The interface doesn't have to reflect the implementation.  E.G. if we were do do that, by ‘lazily loaded’ I mean that whatever we return from `admin_chain()` (which needn't be a `ChainStateView`, just some view of the chain: maybe a `ChainClient` would be a better match, actually) would be a proxy object that would just be populated (at least initially) with the chain ID.  The general pattern is that when you have a genitive in a name you can (and generally should, because it makes the API more powerful) map it to a property or method.

But anyway, I'm not suggesting that for now!  `admin_chain_id` is fine if we're happy with that.

afck:
(Closing this; there's so many conflicts now it would almost be a rewrite.)

PR #5287, opened 2026-01-19
Backport: Make the `--owners` CLI argument to `change-ownership` a map (#5284)

bart-linera:
## Motivation

Right now `change-ownership` takes the owners and their weights in separate arguments. Since we're already using JSON for both, we can unify them into a single map argument.

## Proposal

Parse the value passed to `--owners` as a map.

This also affects `open-multi-owner-chain`.

## Test Plan

CI will catch regressions.

Readme tests were run manually to check that the new format is handled correctly.

## Release Plan

## Links

### DIFF
--- a/CLI.md
+++ b/CLI.md
@@ -295,20 +295,20 @@ Open (i.e. activate) a new multi-owner chain deriving the UID from an existing o
 ###### **Options:**
 
 * `--from <CHAIN_ID>` — Chain ID (must be one of our chains)
-* `--super-owners <SUPER_OWNERS>` — A JSON list of the new super owners. Absence of the argument leaves the current set of super owners unchanged
-* `--owners <OWNERS>` — A JSON map of the new owners to their weights. Absence of the argument leaves the current set of owners unchanged
-* `--multi-leader-rounds <MULTI_LEADER_ROUNDS>` — The number of rounds in which every owner can propose blocks, i.e. the first round number in which only a single designated leader is allowed to propose blocks. "null" is equivalent to 2^32 - 1. Absence of the argument leaves the current setting unchanged
+* `--super-owners <SUPER_OWNERS>` — A JSON list of the new super owners. Absence of the option leaves the current set of super owners unchanged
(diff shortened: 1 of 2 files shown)


CLI.md
```diff
@@ -296,14 +296,7 @@ Open (i.e. activate) a new multi-owner chain deriving the UID from an existing o
 
 * `--from <CHAIN_ID>` — Chain ID (must be one of our chains)
 * `--super-owners <SUPER_OWNERS>` — A JSON list of the new super owners. Absence of the argument leaves the current set of super owners unchanged
```
ma2bd: nit: Perhaps "Absence of the option" is more clear. `<SUPER_OWNERS>` is the argument to me.

examples/amm/README.md
```diff
@@ -254,12 +254,12 @@ with empty blocks.
...
 
 linera --wait-for-outgoing-messages change-ownership \
-    --owners "[\"$OWNER_AMM\",\"$OWNER_2\"]"
+    --owners "{\"$OWNER_AMM\":100,\"$OWNER_2\":100}"
 
 linera --wait-for-outgoing-messages change-application-permissions \
     --execute-operations "[\"$AMM_APPLICATION_ID\"]" \
     --mandatory-applications "[\"$AMM_APPLICATION_ID\"]" \
-    --manage-chain "[\"$AMM_APPLICATION_ID\"]"
+    --close-chain "[\"$AMM_APPLICATION_ID\"]"
```
ma2bd, line 262: was that not failing in CI?

CLI.md
bart-linera: Makes sense, I wasn't 100% sure about the terminology TBH :+1:

examples/amm/README.md
bart-linera: No, because we're not running the readme tests for PRs :grimacing: I thought I ran them manually before merging the PR, but apparently I forgot that time :grimacing: 

I'm not actually sure when they are run automatically. Perhaps only during the nightly tests on `main`.

afck: I think they run as part of the merge queue on `main`, but we're not using the merge queue on `testnet_conway`.

PR #5290, opened 2026-01-20
Make `ListenerCommand`s update the wallet

bart-linera:
## Motivation

Currently, when the `ChainListener` is told to start or stop listening to a chain, the wallet isn't necessarily updated.

## Proposal

Update the wallet explicitly when handling `ListenerCommand`s.

## Test Plan

CI will catch regressions.

## Release Plan

    - be released in a new SDK,

## Links

### DIFF
--- a/linera-client/src/unit_tests/chain_listener.rs
+++ b/linera-client/src/unit_tests/chain_listener.rs
@@ -427,3 +427,96 @@ async fn test_chain_listener_admin_chain() -> anyhow::Result<()> {
 
     Ok(())
 }
+
+/// Tests that the ListenerCommand::Listen actually adds chains to the wallet.
+#[test_log::test(tokio::test)]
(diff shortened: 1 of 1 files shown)


linera-client/src/unit_tests/chain_listener.rs
```diff
@@ -427,3 +427,96 @@ async fn test_chain_listener_admin_chain() -> anyhow::Result<()> {
...
+        wallet_chain.next_block_height,
+        BlockHeight::ZERO,
+        "Chain should be at height 0"
+    );
+
+    cancellation_token.cancel();
+    handle.await;
+
+    Ok(())
+}
```
afck, line 522: I wonder if this could be integrated in (or extended to) a more general `Listen` test, that also verifies that it starts listening immediately. (Not necessarily in this PR.)

Issue #5083, opened 2025-12-04
Remove the genesis configuration from `ClientContext`


Twey:
It's used only in a couple of places for the verification of nodes, a concept that is exported outside of the `ClientContext`, so the genesis config could also be exported outside of the `ClientContext`.  Alternatively and additionally, the genesis config hash, which is sufficient to verify the node, can in theory be retrieved from the local node.

In combination with #5082, this would obviate the `ClientContext` entirely, making it equivalent to `linera_core::Client`.  The `linera-client` crate could be repurposed to provide a stable Rust client API.

afck:
> the fact that the client `storage` is the same as the `LocalNode` `storage`

I'd make that assumption. The local node is part of the client.

Issue #5082, opened 2025-12-04
Move the default chain concept out of `ClientContext` into the various interface layers


Twey:
It doesn't necessarily make sense to have a concept of default chain baked into non-interactive APIs where the same functionality can be achieved with a variable.  It probably makes more sense to move the notion into the CLI and node service directly, and simplifies the `ClientContext` code a lot.

PR #5292, opened 2026-01-20
Move all wallet data into the memory wallet.

afck:
## Motivation

Working on #5285 I found it awkward that the memory wallet contains only the chains (so it doesn't even know e.g. the admin ID), but the persisted wallet also contains the default chain and genesis config.

## Proposal

Move the genesis config to `linera-core` and add all the information so that the memory wallet contains exactly what is being persisted in the wallet file.

## Test Plan

CI

## Release Plan

- These changes should be backported to `testnet_conway`.

## Links

- Related: #5082, #5083

### DIFF
--- a/linera-client/src/unit_tests/chain_listener.rs
+++ b/linera-client/src/unit_tests/chain_listener.rs
@@ -436,8 +436,6 @@ async fn test_chain_listener_listen_command_adds_chains_to_wallet() -> anyhow::R
     let client0 = builder.add_root_chain(0, Amount::ONE).await?;
     let chain_id0 = client0.chain_id();
 
-    let genesis_config = GenesisConfig::new_testing(&builder);
-    let admin_id = genesis_config.admin_id();
     let storage = builder.make_storage().await?;
 
     let context = ClientContext {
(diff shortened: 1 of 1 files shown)


linera-client/src/config.rs
```diff
@@ -78,180 +78,35 @@ impl CommitteeConfig {
...
-        min_active_epoch: Epoch::ZERO,
-        max_active_epoch: Epoch::ZERO,
-        epoch: Epoch::ZERO,
-        ownership: ChainOwnership::single(public_key.into()),
-    };
-    ChainDescription::new(origin, config, timestamp)
+/// Extension trait to create a `GenesisConfig` from a `CommitteeConfig`.
+pub trait GenesisConfigExt {
+    /// Creates a `GenesisConfig` with the first chain being the admin chain.
+    fn new(
```
ma2bd, line 84: nit: `new` feels a little weird for a trait method

Twey: It's usually bad to construct an object as part of a trait because it deprives the provider of the opportunity to operate on the concrete type before type-erasing it for the consumer.  This is the cause of a lot of the complexity in our `linera-views` initialization for example, which then has to try to provide abstractions over all the things a constructor could possibly want to do with the database.

afck:
Closing this for now. It's controversial and complicates the `Wallet` traits. Maybe we can come up with something cleaner.

linera-client/src/config.rs
ma2bd: My comment was mostly about the choice of the name.

ma2bd: > This is the cause of a lot of the complexity in our linera-views initialization for example

I'm not following this. Which trait specifically do you mean?

PR #5293, opened 2026-01-20
Start validator exporters from last committee on startup

deuszx:
## Motivation

When the exporter restarts, it currently has no way to know which validators were in the latest committee. This means committee exporters cannot be automatically restarted for the current set of validators, requiring manual configuration or missing validator destinations until a new committee change is processed.

  ## Proposal

  This PR adds persistence and automatic recovery of the latest committee blob ID in the exporter:

  1. **Persist committee blob ID**: When processing a `CreateCommittee` operation, the exporter now persists the committee blob ID to its state view. This is loaded on subsequent startups.

  2. **Fallback scan for legacy databases**: For databases created before this feature was added, the exporter performs a one-time backwards scan through the canonical blocks to find the most recent `CreateCommittee` operation. Once found, the blob ID is persisted for future restarts.

  3. **Startup committee restoration**: On startup, if `committee_destination` is enabled, the exporter loads the persisted committee blob ID (or finds it via scan), extracts the validator addresses, and automatically starts committee exporters for those destinations.

  4. **Additive configuration**: Configured validator destinations are combined with the persisted committee destinations, allowing operators to add additional validators via config while still using the stored committee.

  Additional fixes included:
  - Fixed a bug in `initiate()` where stored destination states were being discarded instead of returned, breaking restart persistence
  - Normalized exporter metrics names to use consistent `exporter_` prefix

## Test Plan

  - Added `test_committee_blob_persisted_after_processing`: Verifies that after processing a block with `CreateCommittee`, the blob ID is persisted and can be retrieved after reload
  - Added `test_committee_loaded_on_startup`: Verifies that a persisted committee blob ID can be loaded and used to retrieve committee destinations
  - Added `test_fallback_committee_scan_finds_committee`: Verifies that the fallback scan correctly finds committee when no persisted blob ID exists (simulating legacy database)
  - Verified existing `test_restart_persistence_and_faulty_destination_restart` passes (was failing due to the `initiate()` bug)

## Release Plan

- These changes should be backported to `main` (without the fallback code)
- These changes should be deployed to `testnet` exporters

##  Links

  - https://github.com/linera-io/linera-protocol/blob/main/CONTRIBUTING.md#reviewer-checklist

### DIFF
--- a/linera-service/src/exporter/runloops/mod.rs
+++ b/linera-service/src/exporter/runloops/mod.rs
@@ -5,6 +5,7 @@ use std::future::{Future, IntoFuture};
 
 use block_processor::BlockProcessor;
 use indexer::indexer_exporter::Exporter as IndexerExporter;
+use linera_base::identifiers::BlobId;
 use linera_execution::committee::Committee;
 use linera_rpc::NodeOptions;
 use linera_service::config::{DestinationConfig, DestinationId, DestinationKind, LimitsConfig};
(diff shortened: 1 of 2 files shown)


linera-service/src/exporter/metrics.rs
```diff
@@ -8,39 +8,39 @@ use prometheus::{Histogram, HistogramVec, IntCounterVec, IntGauge, IntGaugeVec};
 
 pub(crate) static GET_BLOB_HISTOGRAM: LazyLock<Histogram> = LazyLock::new(|| {
     prometheus_util::register_histogram(
-        "get_blob_histogram_ms",
+        "exporter_get_blob_ms",
```
ndr-ds: This seems like a good use case for subsystems. We should create wrappers on linera-base/src/prometheus_util.rs that accept subsystems, then set them on the Opts together with the namespace.  
Then you could create a `exporter_metrics` module or something that implements these register methods, and sets the subsystem to "exporter"

ndr-ds: I can do the first part, here you go: https://github.com/linera-io/linera-protocol/pull/5299

deuszx: Done.

PR #5294, opened 2026-01-20
Backport: Make `ListenerCommand`s update the wallet (#5290)

bart-linera:
## Motivation

Currently, when the `ChainListener` is told to start or stop listening to a chain, the wallet isn't necessarily updated.

## Proposal

Update the wallet explicitly when handling `ListenerCommand`s.

## Test Plan

CI will catch regressions.

## Release Plan

## Links

### DIFF
--- a/linera-client/src/chain_listener.rs
+++ b/linera-client/src/chain_listener.rs
@@ -557,7 +557,7 @@ impl<C: ClientContext + 'static> ChainListener<C> {
             .event_stream_publishers()
             .await?
             .into_iter()
-            .map(|chain_id| (chain_id, (ListeningMode::FollowChain, None)))
+            .map(|chain_id| (chain_id, (ListeningMode::FullChain, None)))
             .collect();
         for publisher_id in publishing_chains.keys() {
             self.event_subscribers


linera-client/src/chain_listener.rs
```diff
@@ -550,7 +557,7 @@ impl<C: ClientContext + 'static> ChainListener<C> {
             .event_stream_publishers()
             .await?
             .into_iter()
-            .map(|chain_id| (chain_id, ListeningMode::FullChain))
+            .map(|chain_id| (chain_id, (ListeningMode::FullChain, None)))
```
afck, line 560: (Can you remind me why we need `FullChain` here?)

linera-client/src/chain_listener.rs
```diff
@@ -638,6 +622,9 @@ impl<C: ClientContext + 'static> ChainListener<C> {
                                     continue;
                                 };
                                 listening_client.stop().await;
+                                if let Err(error) = self.context.lock().await.wallet().remove(chain_id).await {
```
afck, line 625: Do we still need to _save_ the wallet as an intermediate step, is that unnecessary because the controller should already handle restarts correctly?

afck (approved):

linera-client/src/chain_listener.rs
bart-linera: I think `remove` saves it internally.

linera-client/src/chain_listener.rs
bart-linera: Honestly, not sure, but when I changed it to `FollowChain`, the CI wasn't passing :grimacing: It seems the events aren't processed correctly when a chain is only followed. I'll have to look into it.

linera-client/src/chain_listener.rs
bart-linera: Or if you're asking whether we still need it - technically, we shouldn't need saving either added or removed chains, but because we'll now be saving added chains, we should also be saving removed ones, so that we don't clutter the wallet unnecessarily.

linera-client/src/chain_listener.rs
bart-linera: Ah, of course. We only call `process_new_events` when the `ListeningMode` is `FullChain` :man_facepalming: I'll try to change it in a separate PR.

PR #5295, opened 2026-01-21
Add a `cost_tracking` example code.

MathieuDutSik:
## Motivation

The running of the smart contract has fuel costs associated to it. It is useful to know what they are.

## Proposal

Makes the `remaining_fuel` function available from the `linera-sdk`.

## Test Plan

Inserts the sequence of tests in `cost_tracking`. It is run via
`cargo test -- --nocapture`.

Conclusions:
* `serde_json` is more expensive than `bcs` which is more expensive than `bincode`. Cost for a very simple struct is 1400 wasm fuel.
* The emission of a message costs 2000 gas. The transfer is 1000.

## Release Plan

The information is also meaningful for TestNet Conway and could be backported there.

## Links

None.

### DIFF
--- a/linera-sdk/tests/fixtures/cost-tracking/src/contract.rs
+++ b/linera-sdk/tests/fixtures/cost-tracking/src/contract.rs
@@ -157,8 +157,7 @@ impl CostTrackingContract {
         self.log_entry("after_application_id");
 
         self.log_entry("before_application_parameters");
-        let params = self.runtime.application_parameters();
-        black_box(&params);
+        black_box(&self.runtime.application_parameters());
         self.log_entry("after_application_parameters");
     }
 
@@ -336,7 +335,7 @@ impl Contract for CostTrackingContract {
         self.state.data.set(String::new());
     }
 
-    async fn execute_operation(&mut self, operation: Operation) -> () {
+    async fn execute_operation(&mut self, operation: Operation) {
         match operation {
             Operation::RunAll => {
                 self.run_all().await;


linera-sdk/tests/fixtures/cost-tracking/tests/cost_tracking_test.rs
```diff
@@ -0,0 +1,65 @@
+// Copyright (c) Zefchain Labs, Inc.
```
ma2bd, line 1: Would it make sense to take a snapshot of the results with `cargo insta` ?

MathieuDutSik: I think yes.
Done.

MathieuDutSik: Unfortunately, it fails on CI.
What seems to be the case is that compilation is platform-dependent, and so this yields different Wasm costs.

ma2bd:
The CI failure looks related

linera-sdk/tests/fixtures/cost-tracking/tests/cost_tracking_test.rs
MathieuDutSik: So, I reverted that change.

ma2bd: makes sense

ma2bd (approved):
Thanks for the PR

PR #5304, opened 2026-01-22
Introduce the trait BaseAbi and make the Service use it instead of bcs.

MathieuDutSik:
## Motivation

The trait `ContractAbi` provides explicit functions for serializing and deserializing operations.
This leaves the impression that it suffices to change those functions in order to switch from
one serialization to another.

That is incorrect since the `Service` has the `schedule_operation` implemented as
```rust
    pub fn schedule_operation(&self, operation: &impl Serialize) {
        let bytes = bcs::to_bytes(operation).expect("Failed to serialize application operation");

        service_wit::schedule_operation(&bytes);
    }
```

This is unavoidable since the `schedule_operation` is a Service function and so needs operations.

## Proposal

Introduce a `BaseAbi` trait that implements those functions.

## Test Plan

It works. By replacing `bcs` by `bincode` the wasm test pass.
The `evm_call_wasm` test does not pass since we do not have bincode serialization for solidity
in `serde_generate`.

## Release Plan

The PR is mostly a cleanup because the serialization has not been changed and experiments with `bincode`
indicate that it is not necessarily better.

## Links

None,

### DIFF
--- a/linera-base/src/abi.rs
+++ b/linera-base/src/abi.rs
@@ -21,10 +21,6 @@ impl<T> Abi for T where T: ContractAbi + ServiceAbi {}
 /// A trait that includes all the types exported by a Linera application contract.
 pub trait ContractAbi {
     /// The type of operation executed by the application.
-    ///
-    /// Operations are transactions directly added to a block by the creator (and signer)
-    /// of the block. Users typically use operations to start interacting with an
-    /// application on their own chain.
     type Operation: Serialize + DeserializeOwned + Send + Sync + Debug + 'static;
 
     /// The response type of an application call.
@@ -41,7 +37,6 @@ pub trait ContractAbi {
         bcs::to_bytes(operation)
             .map_err(|e| format!("BCS serialization error {e:?} for operation {operation:?}"))
     }
-
     /// How the `Response` is deserialized
     fn deserialize_response(response: Vec<u8>) -> Result<Self::Response, String> {
         bcs::from_bytes(&response)
@@ -58,7 +53,7 @@ pub trait ContractAbi {
 
 // ANCHOR: service_abi
 /// A trait that includes all the types exported by a Linera application service.
-pub trait ServiceAbi {
+pub trait ServiceAbi: ContractAbi {
     /// The type of a query receivable by the application's service.
     type Query: Serialize + DeserializeOwned + Send + Sync + Debug + 'static;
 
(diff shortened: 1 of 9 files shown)


ma2bd:
Nice catch! How about having `trait ServiceAbi: ContractAbi { .. }` instead of introducing another trait?

linera-base/src/abi.rs
```diff
@@ -18,18 +18,16 @@ pub trait Abi: ContractAbi + ServiceAbi {}
...
+pub trait BaseAbi {
     /// The type of operation executed by the application.
     ///
     /// Operations are transactions directly added to a block by the creator (and signer)
     /// of the block. Users typically use operations to start interacting with an
     /// application on their own chain.
     type Operation: Serialize + DeserializeOwned + Send + Sync + Debug + 'static;
 
-    /// The response type of an application call.
-    type Response: Serialize + DeserializeOwned + Send + Sync + Debug + 'static;
```
ma2bd, line 31: Maybe `schedule_operation` should actually do something about responses (or will in the future).

MathieuDutSik:
> Nice catch! How about having `trait ServiceAbi: ContractAbi { .. }` instead of introducing another trait?

We worked to separate `Contract` from `Service`. That would break that existing design.

linera-base/src/abi.rs
MathieuDutSik: Well, this is a problem of design. The `Service` is for several mostly independent purposes:
* Creating operations as this is done here. I think we should separate it from the service.
* Queries the value of contracts.
* Doing queries to oracles where the wasm cost can vary between validators (but the result should be the same).

But I would say this is for another PR if ever.

ma2bd: It's totally related to this PR because if you move `Response` to `BaseAbi` then it seems to be equal to `ContractAbi`, hence the suggestion `ServiceAbi: ContractAbi`.

I'm not saying `Service: Contract` by the way.

MathieuDutSik: Ok, done.

linera-sdk/src/abis/evm.rs
```diff
@@ -22,7 +22,6 @@ impl ContractAbi for EvmAbi {
     fn serialize_operation(operation: &Self::Operation) -> Result<Vec<u8>, String> {
         Ok(operation.to_vec())
     }
-
```
ma2bd, line 25: nit: keep the newline

PR #5305, opened 2026-01-22
Change account CLI syntax from `chain:owner` to `owner@chain`

afck:
This aligns the CLI account format with the web interface. Chain-only accounts (without an owner) continue to be represented as just the chain ID.

## Motivation

On the CLI we use the syntax `chain:owner` for an account, whereas we use `owner@chain` on the web.

## Proposal

Use `owner@chain` on the CLI, too.

`0x00@chain` and just `chain` are equivalent, but the former is printed by the `Display` impl.

## Test Plan

A unit test for the string representation was added.

## Release Plan

    - be released in a new SDK.

## Links

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,12 +134,12 @@ linera query-balance "$CHAIN1"
 linera query-balance "$CHAIN2"
 
 # Now let's fund the user balances.
-linera transfer 5 --from "$CHAIN1" --to "$CHAIN1:$ACCOUNT1"
-linera transfer 2 --from "$CHAIN1:$ACCOUNT1" --to "$CHAIN2:$ACCOUNT2"
+linera transfer 5 --from "$CHAIN1" --to "$ACCOUNT1@$CHAIN1"
+linera transfer 2 --from "$ACCOUNT1@$CHAIN1" --to "$ACCOUNT2@$CHAIN2"
 
 # Query user balances again.
-linera query-balance "$CHAIN1:$ACCOUNT1"
-linera query-balance "$CHAIN2:$ACCOUNT2"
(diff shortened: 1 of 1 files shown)


linera-base/src/identifiers.rs
```diff
@@ -156,27 +156,26 @@ impl Account {
 
 impl fmt::Display for Account {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "{}:{}", self.chain_id, self.owner)
+        if self.owner.is_chain() {
+            write!(f, "{}", self.chain_id)
```
afck: Or should we print something like `0x00@chain` in that case? (More consistent with the previous behavior.)

afck: Done in https://github.com/linera-io/linera-protocol/commit/9fa15e2fdb37b77f79102e722eb365148b317605.

Twey: Does this completely replace `chain-id` accounts with `0x00@chain-id`?

(I'd like that! the conflation of accounts with chains is largely a historical artefact, afaict)

afck: We still _accept_ just `chain-id`. We just don't print it that way.

If we decide to not accept that anymore, let's maybe also do that in a separate PR? Might require discussion with the team first.

Twey (approved):

linera-base/src/identifiers.rs
Twey: Makes sense to me.  I don't think it's urgent right now, but this seems like a step in the right direction.

PR #5309, opened 2026-01-22
[testnet] Don't automatically retry operations on conflict. (#5234)

afck:
Backport of #5234.

## Motivation

The chain client currently automatically retries the same operations at the next block height if a conflicting block got committed at the current height. But the same operations are not necessarily desired at the new chain state anymore.

## Proposal

Remove the automatic retry behavior, and return a `ClientOutcome::Conflict` instead, to let the caller decide.

## Test Plan

The tests were updated.

## Release Plan

## Links

- PR to main: #5234

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -1768,6 +1768,12 @@ pub enum ChainClientError {
         chain_id: ChainId,
         height: BlockHeight,
     },
+
+    #[error(
+        "A different block was already committed at this height. \
+         The committed certificate hash is {0}"
+    )]
+    Conflict(CryptoHash),
 }
 
 impl From<Infallible> for ChainClientError {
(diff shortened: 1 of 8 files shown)


linera-core/src/client/mod.rs
```diff
@@ -2728,10 +2735,13 @@ impl<Env: Environment> ChainClient<Env> {
         // TODO(#5092): We shouldn't need to call this explicitly.
         match self.process_pending_block_without_prepare().await? {
             ClientOutcome::Committed(Some(certificate)) => {
-                return Ok(ExecuteBlockOutcome::Conflict(certificate))
+                return Ok(ClientOutcome::Conflict(Box::new(certificate)))
```
deuszx, line 2738: Do I understand correctly that now we have two different `ClientOutcome` enums?

afck: I don't think so, no?

deuszx: Ah, sorry. I saw that you changed `Some(certificate)` to `Box::new(certificate)` but I didn't noticed you changed the variant of the enum.

PR #5311, opened 2026-01-23
Fix handle block when open_multi_leader_rounds is true

kikakkz:
## Motivation

In default, we always have owner in super_owners or owners of a microchain. But when open_multi_leader_rounds is true, super_owners and owners may be empty set. This case is not processed correctly right now.

## Proposal

Fix the case of open_multi_leader_rounds is true at wallet and validator side.

## Test Plan

CI

## Release Plan

be released in a new SDK

## Links

<!--
Optional section for related PRs, related issues, and other references.

### DIFF
--- a/linera-base/src/ownership.rs
+++ b/linera-base/src/ownership.rs
@@ -153,6 +153,7 @@ impl ChainOwnership {
         !self.super_owners.is_empty()
             || !self.owners.is_empty()
             || self.timeout_config.fallback_duration == TimeDelta::ZERO
+            || self.open_multi_leader_rounds
     }
 
     /// Returns `true` if this is an owner or super owner.
--- a/linera-client/src/client_context.rs
+++ b/linera-client/src/client_context.rs
@@ -488,9 +488,13 @@ impl<Env: Environment> ClientContext<Env> {
             .extend_chain_mode(chain_id, ListeningMode::FullChain);
         let client = self.make_chain_client(chain_id).await?;
         let chain_description = client.get_chain_description().await?;
-        let config = chain_description.config();
+        // let config = chain_description.config();
 
-        if !config.ownership.is_owner(&owner) {
+        let info = client.chain_info().await?;
+
+        if !info.manager.ownership.is_owner(&owner)
+            && !info.manager.ownership.open_multi_leader_rounds
+        {
             tracing::error!(
                 "The chain with the ID returned by the faucet is not owned by you. \
                 Please make sure you are connecting to a genuine faucet."
--- a/linera-core/src/chain_worker/state.rs
+++ b/linera-core/src/chain_worker/state.rs
@@ -1528,7 +1528,8 @@ where
         block.check_proposal_size(policy.maximum_block_proposal_size)?;
         // Check the authentication of the block.
         ensure!(
-            chain.manager.can_propose(&owner, proposal.content.round),
+            chain.manager.can_propose(&owner, proposal.content.round)
+                || chain.ownership().open_multi_leader_rounds,
             WorkerError::InvalidOwner
         );
         let old_round = self.chain.manager.current_round();
--- a/linera-core/src/client/chain_client/mod.rs
+++ b/linera-core/src/client/chain_client/mod.rs
@@ -653,7 +653,7 @@ impl<Env: Environment> ChainClient<Env> {
         let is_owner = manager.ownership.is_owner(&preferred_owner)
             || fallback_owners.contains(&preferred_owner);
 
-        if !is_owner {
+        if !is_owner && !manager.ownership.open_multi_leader_rounds {
             warn!(
                 chain_id = %self.chain_id,
                 ownership = ?manager.ownership,
@@ -1913,7 +1913,9 @@ impl<Env: Environment> ChainClient<Env> {
             .values()
             .map(|v| (AccountOwner::from(v.account_public_key), v.votes))
             .collect();
-        if manager.should_propose(identity, round, seed, &current_committee) {
+        if manager.should_propose(identity, round, seed, &current_committee)
+            || manager.ownership.open_multi_leader_rounds
+        {
             return Ok(Either::Left(round));
         }
         if let Some(timeout) = info.round_timeout() {


linera-client/src/client_context.rs
```diff
@@ -488,9 +488,13 @@ impl<Env: Environment> ClientContext<Env> {
             .extend_chain_mode(chain_id, ListeningMode::FullChain);
         let client = self.make_chain_client(chain_id).await?;
         let chain_description = client.get_chain_description().await?;
-        let config = chain_description.config();
+        // let config = chain_description.config();
```
ma2bd, line 491: remove extra line?

kikakkz: I'm not sure if we need to put open_multi_leader_rounds expression as a method (or merge to any exists method) of ownership so right now it's just a draft modification which let it work for discussion. After this discussion I'll clean the code~

linera-core/src/chain_worker/state.rs
```diff
@@ -1528,7 +1528,8 @@ where
         block.check_proposal_size(policy.maximum_block_proposal_size)?;
         // Check the authentication of the block.
         ensure!(
-            chain.manager.can_propose(&owner, proposal.content.round),
+            chain.manager.can_propose(&owner, proposal.content.round)
+                || chain.ownership().open_multi_leader_rounds,
```
afck, line 1532: This should be unnecessary? `can_propose` should already return `true` if the current round is a multi-leader round and `open_multi_leader_rounds` is `true`.

linera-base/src/ownership.rs
```diff
@@ -153,6 +153,7 @@ impl ChainOwnership {
         !self.super_owners.is_empty()
             || !self.owners.is_empty()
             || self.timeout_config.fallback_duration == TimeDelta::ZERO
+            || self.open_multi_leader_rounds
```
afck, line 156: Good catch! 👍 
But let's make sure there _are_ any multi-leader rounds, too:


```suggestion
            || (self.open_multi_leader_rounds && self.multi_leader_rounds > 0)
```

It's still an open question what we should do if the multi-leader rounds have been exhausted and there are no other chain owners. Then the chain is effectively bricked. I guess at least the _recommendation_ (Or should it be enforced?) should be that if there are no owners, `multi_leader_rounds` should be `u32::MAX`.

linera-client/src/client_context.rs
afck, line 491: Good catch! 👍 

Originally this method was used only for _new_ chains, where the current state was still identical with the chain description. But now we use it in the `assign` command for any chain.

We should actually go further and not use the chain description at all in this method. Take the timestamp and epoch from the chain info, too.

linera-core/src/client/chain_client/mod.rs
```diff
@@ -1913,7 +1913,9 @@ impl<Env: Environment> ChainClient<Env> {
             .values()
             .map(|v| (AccountOwner::from(v.account_public_key), v.votes))
             .collect();
-        if manager.should_propose(identity, round, seed, &current_committee) {
+        if manager.should_propose(identity, round, seed, &current_committee)
+            || manager.ownership.open_multi_leader_rounds
```
afck, line 1917: This one I'm not sure about:
* Firstly, even if we do want to do e.g. a PoW proposal, we still need to check whether the current round is a multi-leader round.
* Secondly, how does this work with PoW? If we have found a PoW solution, that will only be valid in high enough rounds, will it? The current round selection would choose the first multi-leader round that is still available (has no pending proposal), but in this case we might need a higher one?

linera-core/src/chain_worker/state.rs
kikakkz: Yeah, I checked `can_propose` again and it already processed `open_multi_leader_rounds`.

linera-core/src/client/chain_client/mod.rs
kikakkz: For the first one, yes, we should check it's multi-leader round, too.
For the second one (I'm not sure I understand it fully actually~~), right now the hash is calculated with
```
#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]                                                                              
pub struct MiningBase {                                                                                                                     
    pub nonce: CryptoHash,                                                                                                                  
    pub height: BlockHeight,                                                                                                                
    pub chain_id: ChainId,                                                                                                                  
    pub signer: AccountOwner,                                                                                                               
    pub previous_nonce: CryptoHash,                                                                                                         
}                                                                                                                                           
                                                                                                                                            
impl BcsSignable<'_> for MiningBase {}
```
And after the Fast round and application & pool creation, it'll goto mining (with empty owners set) directly in current implementation. We didn't touch or check the round ATM.

afck: I see. It'll have to depend on the round eventually, though, otherwise two solutions can block each other indefinitely.

kikakkz: > I see. It'll have to depend on the round eventually, though, otherwise two solutions can block each other indefinitely.

`BlockHeight` may be same in different rounds ?

afck: Yes, at each block height there are (potentially) many rounds.

kikakkz: Understood. Will add round in MiningBase.

afck: The important thing for liveness is that higher rounds are harder to mine. It's not important that the round is part of the data that gets hashed for mining; what's important is that the round determines the _difficulty_.

linera-base/src/ownership.rs
kikakkz: E.g. `self.open_multi_leader_rounds && self.multi_leader_rounds == u32::MAX` right ? Or `assert` when constructing or changing the ownership ?

linera-client/src/client_context.rs
kikakkz: Will update this part in formal PR, too.

linera-base/src/ownership.rs
afck: If we decide that it's invalid, we should ensure (i.e. return an error; not panic) it when constructing the ownership, I think, yes. I'll discuss it with the team.

linera-core/src/client/chain_client/mod.rs
kikakkz: Do you mean the miners will shutdown machines if all tokens have been mined.

Currently we design a difficulty adjustment interval to about 3 hours (2160 blocks) right now. So for each 2160 blocks it should take about 3 hours to produce. If it's shorter then difficulty will be raised according to the gap of expected duration and elapsed duration, otherwise decreased. The block reward will be halved about one year.

afck: No, I mean this still causes liveness problems if by chance two clients happen to solve the PoW challenge roughly at the same time. (Maybe not very likely with 3 hours, but still possible.) Then they make both their valid proposals in round 0, so that round fails, but then they can still both make the same proposal in round 1 and conflict again, and so on.

kikakkz: There'll be only one `Mine` operation could be run successfully for one `BlockHeight`. The first one `Mine` operation will update the `mining_height` in application state, then `submit_height <= mining_height` will fail to run. Then the second proposal at the same height with valid solution will be rejected.

kikakkz: Will the BlockHeight be increased even the message is rejected ?

kikakkz: @afck I just read the client code and found even the message is rejected, a new block will be produced then BlockHeight will be increased. Then if someone follow the meme chain with a client which doesn't implement Mine operation (e.g. follow the chain with official client), then the mining will always fail due to the block height is different.

afck: The problem happens before the operation or message could get committed (or rejected): This is about the validators agreeing on what the next block will be in the first place. If there are two conflicting block _proposals_, they may fail to agree on a block in that round, and have to retry in the next round.

If the same two `Mine` operations remain valid at the next round, they can both make the same proposals again.

(Anyway, with a 3 hour target maybe that's unlikely for now. We can improve on that later.)

kikakkz: > @afck I just read the client code and found even the message is rejected, a new block will be produced then BlockHeight will be increased. Then if someone follow the meme chain with a client which doesn't implement Mine operation (e.g. follow the chain with official client), then the mining will always fail due to the block height is different.

And, how to handle this ? Right now I just try to implement the auto maker with miner to run full work flow. But if anyone else follow and assign the chain to their owner with `linera` client, it'll process the inbox directly then get many block proposals with rejected messages then miners will always fail for error height.

afck: You need to configure the chain with `ApplicationPermissions::mandatory_applications` set to that app.

The app should then enforce that each block starts with a valid `Mine` operation, so no block without that is valid.

But actually… that only works for applications without messages, because you're right: A rejected incoming message from that application still satisfies the `mandatory_applications` requirement. I'll bring it up with the team next week.

kikakkz: > You need to configure the chain with `ApplicationPermissions::mandatory_applications` set to that app.
> 
> The app should then enforce that each block starts with a valid `Mine` operation, so no block without that is valid.
> 
> But actually… that only works for applications without messages, because you're right: A rejected incoming message from that application still satisfies the `mandatory_applications` requirement. I'll bring it up with the team next week.

Hi @afck , how is this going ? It's the base for our mining mechanism.

afck: I created a PR for this: #5353
(Not sure if we can backport it to the testnet, though. But that shouldn't stop you from using the `mandatory_applications` setting and testing with it.)

PR #5318, opened 2026-01-23
Implement the sync views for smart contracts

### LINKED ISSUE #3026: Have a version of `linera_views` that is synchronous
This would simplify the Linera SDK which is a win.
And it would not impact the smart contracts since they are not really async. The problem is that the code duplication would be an unfortunate side effect.

MathieuDutSik:
## Motivation

The containers used in the examples are asynchronous, but actually the code is synchronous.

Fixes #3026 

## Proposal

The implementation is large but done in a straightforward way:
* Implement a `SyncView` and `SyncContext` trait.
* Implement a `ViewSyncContext`.
* Implement the equivalent of the classic containers, but only the ones that make sense to use in Wasm smart contracts. This is done in `sync_view`.

Possible next steps:
* Make the trait functions of `load`, `instantiate`, `execute_operation`, etc. synchronous. 
* Possibly replace the `async-graphql` by a `graphql`.

## Test Plan

The CI.
Running the
```bash
cargo test -- --nocapture
```

in the fungible example lead us to a 23% improvement in wasm gas fuel expense.

## Release Plan

It is a little bit unclear how this could be implemented on testnet_conway.

## Links

None.

### DIFF
--- a/examples/amm/src/state.rs
+++ b/examples/amm/src/state.rs
@@ -4,8 +4,8 @@
 use linera_sdk::{
     linera_base_types::{Account, Amount},
     views::{linera_views, MapView, RegisterView, ViewStorageContext},
+    RootView,
 };
-use linera_sdk::RootView;
 
 #[derive(RootView, async_graphql::SimpleObject)]
 #[view(context = ViewStorageContext)]
--- a/examples/fungible/src/state.rs
+++ b/examples/fungible/src/state.rs
@@ -2,10 +2,10 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use fungible::{InitialState, OwnerSpender};
-use linera_sdk::RootView;
 use linera_sdk::{
     linera_base_types::{AccountOwner, Amount},
     views::{linera_views, MapView, ViewStorageContext},
+    RootView,
 };
 
 /// The application state.
--- a/examples/controller/src/state.rs
+++ b/examples/controller/src/state.rs
@@ -7,8 +7,8 @@ use linera_sdk::{
     abis::controller::{ManagedServiceId, Worker},
     linera_base_types::{AccountOwner, ChainId, MessagePolicy},
     views::{linera_views, MapView, RegisterView, SetView, ViewStorageContext},
+    RootView,
 };
-use linera_sdk::RootView;
 
 /// The state of the service controller application.
 #[derive(RootView, async_graphql::SimpleObject)]
--- a/examples/counter-no-graphql/src/state.rs
+++ b/examples/counter-no-graphql/src/state.rs
@@ -1,8 +1,10 @@
 // Copyright (c) Zefchain Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-use linera_sdk::views::{linera_views, RegisterView, ViewStorageContext};
-use linera_sdk::RootView;
+use linera_sdk::{
+    views::{linera_views, RegisterView, ViewStorageContext},
+    RootView,
+};
 
 /// The application state.
 #[derive(RootView)]
--- a/examples/counter/src/state.rs
+++ b/examples/counter/src/state.rs
@@ -1,8 +1,10 @@
 // Copyright (c) Zefchain Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-use linera_sdk::views::{linera_views, RegisterView, ViewStorageContext};
-use linera_sdk::RootView;
+use linera_sdk::{
+    views::{linera_views, RegisterView, ViewStorageContext},
+    RootView,
+};
 
 /// The application state.
 #[derive(RootView, async_graphql::SimpleObject)]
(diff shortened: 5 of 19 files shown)


examples/fungible/src/state.rs
```diff
@@ -4,21 +4,21 @@
...
 };
 
 /// The application state.
-#[derive(RootView)]
+#[derive(SyncView)]
 #[view(context = ViewStorageContext)]
 pub struct FungibleTokenState {
-    pub accounts: MapView<AccountOwner, Amount>,
-    pub allowances: MapView<OwnerSpender, Amount>,
+    pub accounts: SyncMapView<AccountOwner, Amount>,
```
ma2bd: Can we keep `MapView`? This is an SDK alias already

examples/fungible/src/state.rs
```diff
@@ -4,21 +4,21 @@
 use fungible::{InitialState, OwnerSpender};
 use linera_sdk::{
     linera_base_types::{AccountOwner, Amount},
-    views::{linera_views, MapView, RootView, ViewStorageContext},
+    views::{linera_views, SyncMapView, SyncView, ViewStorageContext},
 };
 
 /// The application state.
-#[derive(RootView)]
+#[derive(SyncView)]
```
ma2bd: idem: `RootView` seems preferrable

ma2bd: and otherwise the full name is probably `SyncRootView` ?

examples/fungible/src/state.rs
MathieuDutSik: Ok, this can be done.

MathieuDutSik: Done.

examples/fungible/src/state.rs
MathieuDutSik: Ok, corrected.

examples/amm/src/state.rs
```diff
@@ -3,10 +3,10 @@
 
 use linera_sdk::{
     linera_base_types::{Account, Amount},
-    views::{linera_views, MapView, RegisterView, RootView, ViewStorageContext},
+    views::{linera_views, MapView, RegisterView, SyncRootView, ViewStorageContext},
```
ma2bd: I meant to suggest `SyncRootView` to be the name in `linera_views` but similarly to the other names, we could drop the `Sync` in the `linera-sdk` alias.

linera-views/src/lib.rs
```diff
@@ -83,6 +83,9 @@ pub use error::ViewError;
 /// Elementary data-structures implementing the [`views::View`] trait.
 pub mod views;
 
+/// Synchronous views and containers.
+pub mod sync_view;
```
ma2bd: `sync_views`

linera-views/src/store.rs
```diff
@@ -115,6 +115,107 @@ pub trait WritableKeyValueStore: WithError {
...
+    /// Writes the `batch` in the database.
+    fn write_batch(&self, batch: Batch) -> Result<(), Self::Error>;
+
+    /// Clears any journal entry that may remain.
+    /// The journal is located at the `root_key`.
+    fn clear_journal(&self) -> Result<(), Self::Error>;
+}
+
+/// Synchronous direct write key-value operations with simplified batch.
+pub trait DirectWritableSyncKeyValueStore: WithError {
```
ma2bd, line 186: Here and in other places, it seems like the concept `XYZ` becomes `SyncXYZ` (consistent with your comment here)

linera-views/src/lib.rs
MathieuDutSik: Ok.

linera-views/src/store.rs
MathieuDutSik: Ok, done.

examples/amm/src/state.rs
MathieuDutSik: Ok. Done.

MathieuDutSik:
> The code duplication is a bit terrifying but we knew this might happen. Do you have any other measurement (beside fuel) to justify the effort (bytecode size? latency?) Also the fungible app is a bit too simple for a final assessment. Any luck measuring the matching engine (or some more complex internal app)?

Besides the fuel expense, there is a gain in simplicity of the contract code.

About metrics:
* The matching engine wasm code size decreased by 8% both for contract and service.
* The performance increases by about 19% for the matching engine test.

Twey:
+7,000 LoC :scream: 

I'm a bit leery about the level of code duplication this entails.  If we avoid blocking code-folding optimizations then the compiler for the SDK code should know that there's no async required and fold it away — without making a less-general copy of `linera-views`.  @ma2bd points out that the mutexes are different because we use async mutexes — it could be that we can avoid async mutexes entirely (since they're [loaded with footguns](https://rfd.shared.oxide.computer/rfd/0609)), but if not we could parameterize the view over the async mutex type and for the SDK provide one whose async operations are no-ops.

If we are going to copy/paste the whole code of `linera-views` I'd quite like it to be in a separate library, e.g. `linera-views-sync`, or possibly folded right into `linera-sdk` since that's the only place that will use it.

MathieuDutSik:
Another direction has been chosen.
So, let us close for now this PR.

PR #5319, opened 2026-01-24
Use serde-reflection to track the BCS formats of application data

ma2bd:
## Motivation

Ultimately enable the block explorer with introspection features

## Proposal

* Use serde-reflection and cargo-insta to track the BCS formats
* Normalize the container names so that all formats are equivalent
* Introduce a new trait `BcsApplication`

This should allows using `serde-reflection`'s [json_converter](https://docs.rs/serde-reflection/latest/serde_reflection/json_converter/index.html) module fairly directly to interprete BCS values as JSON.

## Test Plan

CI

### DIFF
--- a/examples/social/src/lib.rs
+++ b/examples/social/src/lib.rs
@@ -99,6 +99,7 @@ pub struct Key {
 
 /// An event emitted by the social app.
 #[derive(Debug, PartialEq, Serialize, Deserialize)]
+#[doc(hidden)]
 pub enum Event {
     /// A new post was created
     Post { post: OwnPost, index: u32 },


examples/time-expiry/tests/snapshots/format__format.snap
```diff
@@ -0,0 +1,18 @@
...
+    ENUM:
+      0:
+        ExpireAfter:
+          NEWTYPE:
+            TYPENAME: TimeDelta
+OPERATION:
+  TYPENAME: TimeExpiryOperation
+RESPONSE: UNIT
+MESSAGE: UNIT
+EVENT_VALUE: UNIT
```
afck, line 18: I'd add formats only to the demo-able examples, not to the ones that are only meant for tests.

linera-sdk/src/formats.rs
```diff
@@ -0,0 +1,33 @@
...
+    /// The format of messages.
+    pub message: Format,
+    /// The format of events.
+    pub event_value: Format,
+}
+
+/// An application using BCS as binary encoding.
+pub trait BcsApplication {
+    /// Link the public Abi of application for good measure.
+    type Abi;
```
afck, line 29:
```suggestion
    type Abi: ContractAbi;
```

examples/social/src/lib.rs
```diff
@@ -96,6 +96,63 @@ pub struct Key {
     pub index: u32,
 }
 
+/// An event emitted by the social app.
+#[derive(Debug, PartialEq, Serialize, Deserialize)]
+pub enum Event {
```
afck, line 103: We were always saying that messages and events are internal implementation details of an application, as opposed to operations and results. I guess that's not entirely true anymore for applications where we want to make them human-readable.

Anyway, we could still consider (case-by-case?) making this private, so that it's readable in the explorer but still not part of the Rust library's interface, i.e. it doesn't give the impression that other crates are supposed to use this type.

examples/rfq/tests/snapshots/format__format.snap
```diff
@@ -0,0 +1,156 @@
...
+            TUPLEARRAY:
+              CONTENT: U8
+              SIZE: 20
+  Amount:
+    NEWTYPESTRUCT: U128
+  ApplicationId:
+    STRUCT:
+      - application_description_hash:
+          TYPENAME: CryptoHash
+  ChainId:
```
afck, line 33: (I guess the explorer might then decide by name how to display these. In the future e.g. it could show types called `ChainId` as links to that chain. We should definitely show types called `Amount` as fixed-point integers with 10<sup>18</sup> decimals.)

examples/gen-nft/tests/snapshots/format__format.snap
```diff
@@ -0,0 +1,105 @@
+---
+source: gen-nft/tests/format.rs
+expression: "GenNftApplication::formats().unwrap()"
+---
+REGISTRY:
+  Account:
```
afck, line 6: It would be great if `serde_reflection` included the comments in the format files.

(But I guess then it would have to be completely rewritten as a procedural macro, rather than something that traces the actual serialization process, which would be a lot more fragile and error-prone. 😕)

examples/amm/src/lib.rs
```diff
@@ -95,3 +95,49 @@ pub enum Message {
...
+        linera_base_types::AccountOwner,
+    };
+    use serde_reflection::{Samples, Tracer, TracerConfig};
+
+    use super::{AmmAbi, Message, Operation};
+
+    /// The AMM application.
+    pub struct AmmApplication;
+
+    impl BcsApplication for AmmApplication {
```
deuszx, line 113: We need to make this derive-able for all contracts. I.e. it shouldn't require manual setup.

linera-sdk/src/abis/controller.rs
```diff
@@ -130,3 +130,84 @@ pub struct LocalWorkerState {
...
+            let (response, _) = tracer.trace_type::<()>(&samples)?;
+            let (message, _) = tracer.trace_type::<Message>(&samples)?;
+            let (event_value, _) = tracer.trace_type::<()>(&samples)?;
+
+            // Trace additional supporting types to populate the registry
+            tracer.trace_type::<WorkerCommand>(&samples)?;
+            tracer.trace_type::<ControllerCommand>(&samples)?;
+            tracer.trace_type::<Worker>(&samples)?;
+            tracer.trace_type::<ManagedServiceId>(&samples)?;
+            tracer.trace_type::<AccountOwner>(&samples)?;
```
deuszx: I find this confusing that we have `Format` type but also have to trace additional supporting types separately. This means it cannot be easily automated (i.e. some detection based on already-existing contract requirements).

ma2bd: This is required for enums. There is no way around it for now. Claude oneshot-ed the whole thing, though.

linera-sdk/src/formats.rs
ma2bd: I don't know. We're not really using this type.

examples/social/src/lib.rs
ma2bd: Yes, good idea, let's make these types private.

examples/time-expiry/tests/snapshots/format__format.snap
ma2bd: It seems to me like the examples only meant for tests should be moved somewhere else in another PR.

examples/rfq/tests/snapshots/format__format.snap
ma2bd: When we have a standard notion of environment (see json_converter docs). We should trim these entries before returning the value of type `Formats`.

examples/amm/src/lib.rs
ma2bd: That's a completely different project.

examples/social/src/lib.rs
ma2bd: Ok so it actually doesn't work because `contact.rs` and `service.rs` need to use these definitions from another compilation unit. What we could do is mark these as `#[doc(hidden)]`. I'll add the commit. Let me know what you think

afck: (I guess they could be public in yet another crate—something like `app-internals`—and then imported but not re-exported here. But I agree it's not worth complicating things further; I'm happy with `#[doc(hidden)]`. Not a blocker anyway.)

afck: `#[doc(hidden)]` should apply to `Event`, too, though, not only `Message`.

examples/gen-nft/tests/snapshots/format__format.snap
ma2bd: Yes that's super hard to do

PR #5322, opened 2026-01-25
Add the `wasm-opt` to the created WebAssembly files.

MathieuDutSik:
## Motivation

The running of WebAssembly programs has some costs associated with it.
We want to minimize as much as possible the Wasm fuels. 

## Proposal

Apply the `wasm-opt` to the created Wasm files. Running the `wasm-opt` is very fast
so this is not a problem for the tests.

## Test Plan

The CI.
Running on the fungible example, we get a 1.6% improvement in performance.

## Release Plan

This can be backported to TestNet Conway.

## Links

None.

### DIFF
--- a/linera-base/src/util/mod.rs
+++ b/linera-base/src/util/mod.rs
@@ -7,3 +7,4 @@ Utilities used throughout the Linera codebase.
 
 pub mod future;
 pub mod traits;
+pub mod wasm;
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -192,6 +192,7 @@ jobs:
     steps:
     - uses: actions/checkout@v4
     - uses: actions-rust-lang/setup-rust-toolchain@v1
+    - uses: jetli/wasm-pack-action@v0.4.0
     - uses: foundry-rs/foundry-toolchain@v1.4.0
     - name: Ensure Solc Directory Exists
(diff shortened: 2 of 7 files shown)


ma2bd:
What's "wasm-opt" ?

MathieuDutSik:
> What's "wasm-opt" ?

A program from "wasm-pack". It does decrease the size of fungible contract by 30% and improve the speed a little.

linera-base/src/util/mod.rs
```diff
@@ -7,3 +7,4 @@ Utilities used throughout the Linera codebase.
 
 pub mod future;
 pub mod traits;
+pub mod wasm;
```
ma2bd, line 10: Does this have to be in `linera-base`?

MathieuDutSik: We have code for compiling the contract in 4 places:
* linera-execution/src/wasm/mod.rs
* linera-sdk/src/test/chain.rs
* linera-service/src/cli_wrappers/wallet.rs
* linera-service/src/project.rs

Now maybe we could move it to `linera-execution`.

MathieuDutSik:
The gain from this use of `wasm_opt` appear modest.
They also appear to break CI. So, I will leave to future Linera programmers to decide what is best.

PR #5330, opened 2026-01-26
Track the changes in bytes stored

MathieuDutSik:
## Motivation

The number of bytes stored is always 0 in the runs of the smart contract, which is inadequate.

Edit #1536

## Proposal

Track the changes.
We limit the reduction in the expense by relasing storage to at most half of the total costs.

## Test Plan

The CI.
An integration test has been added that the storage costs was non-zero for a storage test.

## Release Plan

Backporting to TestNet Conway would be a protocol change.

## Links

None.

### DIFF
--- a/linera-execution/src/execution_state_actor.rs
+++ b/linera-execution/src/execution_state_actor.rs
@@ -306,7 +306,19 @@ where
                 callback,
             } => {
                 let mut view = self.state.users.try_load_entry_mut(&id).await?;
+                let size_before = view.total_size();
                 view.write_batch(batch).await?;
+                let size_after = view.total_size();
+                let total_before = u64::from(size_before.key) + u64::from(size_before.value);
+                let total_after = u64::from(size_after.key) + u64::from(size_after.value);
+                let delta = i64::try_from(total_after)
+                    .and_then(|after| i64::try_from(total_before).map(|before| after - before))
+                    .map_err(|_| ExecutionError::from(ArithmeticError::Overflow))?;
+                if delta != 0 {
+                    let delta = i32::try_from(delta)
+                        .map_err(|_| ExecutionError::from(ArithmeticError::Overflow))?;
+                    self.resource_controller.track_stored_bytes(delta)?;
+                }
                 callback.respond(());
             }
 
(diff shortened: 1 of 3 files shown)


linera-execution/src/execution_state_actor.rs
```diff
@@ -907,10 +920,25 @@ where
...
+            .bytes_stored
+            .checked_sub(initial_bytes_stored)
+            .ok_or(ExecutionError::from(ArithmeticError::Overflow))?;
+        let controller_balance = controller.balance()?;
+        let fees_spent = if controller_balance > initial_balance {
+            Amount::ZERO
+        } else {
+            initial_balance.try_sub(controller_balance)?
+        };
+        let refund_limit = Amount::from_attos(fees_spent.to_attos() / 2);
```
ma2bd, line 939: This `1/2` should be part of the policy (if only for visibility).

ma2bd (commented):
We may want to address the performance questions before committing to direction:
https://github.com/linera-io/linera-protocol/issues/5443

MathieuDutSik:
> We may want to address the performance questions before committing to direction: #5443

Well, I agree. I do not like the tracking of the individual costs of each and every write. That is very expensive.
My proposal would be to replace this by making the costs on the batch that is created.

linera-execution/src/resources.rs
```diff
@@ -425,6 +425,40 @@ where
...
+            self.update_balance(self.policy.bytes_stored_price(count)?)?;
+        } else {
+            let count = u64::try_from(delta.checked_abs().ok_or(ArithmeticError::Overflow)?)
+                .map_err(|_| ArithmeticError::Overflow)?;
+            let refund = self.policy.bytes_stored_price(count)?;
+            let refund = if refund > refund_limit {
+                refund_limit
+            } else {
+                refund
+            };
```
ndr-ds, line 454:
```suggestion
            let refund = refund.min(refund_limit);
```

MathieuDutSik:
There is no reliable way to get good performance when keeping track of the `(key, value)` sizes.

Issue #5331, opened 2026-01-26
Tag the chains of synthetic traffic so the indexer can separate then from real traffic


ma2bd:
Two approaches:
1. Set an artificial owner address with weight 0
2. Use a separate faucet just for tests

PR #5213, opened 2026-01-06
Log method name on more metrics

ndr-ds:
## Motivation

We don't have the `method_name` label for all the proxy/server level metrics, and we should.

## Proposal

Add `method_name` label

## Test Plan

Will deploy a small network and make sure this works as expected.

## Release Plan

- These changes should be backported to the latest `testnet` branch, then
    - be released in a validator hotfix.

PR #5334, opened 2026-01-26
Distinguish synthetic from organic traffic in metrics and traces

ndr-ds:
## Motivation

Enable filtering of benchmark/synthetic traffic from organic/production traffic in observability dashboards. This
allows us to accurately measure production performance without benchmark traffic skewing the metrics.

**Note:** This PR addresses **live request metrics only**. Tagging synthetic traffic generated chains/etc in the database/indexer (so
chains created by benchmarks can be filtered) will be addressed separately in #5331.

## Proposal

Implement OpenTelemetry baggage propagation to tag traffic with a `traffic_type` label (`organic`, `synthetic`, or
`unknown`):

**Core propagation infrastructure**:

- `OtelContextLayer` Tower middleware that extracts OpenTelemetry context from incoming gRPC requests
- Helper functions for injecting/extracting context and traffic type from requests

**Client-side**:

- All outgoing gRPC requests now inject the traffic type baggage
- Checks `LINERA_TRAFFIC_TYPE` env var to determine traffic type

**Proxy**:

- Extracts traffic type from incoming requests for metrics
- Propagates OpenTelemetry context when forwarding requests to workers

**Server**:

- Added `OtelContextLayer` to the tower stack
- All request level metrics now include `traffic_type` label

**Benchmark command**:

- `linera benchmark` automatically sets `LINERA_TRAFFIC_TYPE=synthetic`

**Affected metrics** (all now have `traffic_type` label):

- `server_request_latency`, `server_request_count`, `server_request_success`, `server_request_error`
- `server_request_latency_per_request_type`
- `proxy_request_latency`, `proxy_request_success`, `proxy_request_error`

## Test Plan

- CI
- Temporary network deployed, and benchmarked + "organic" traffic also generated by a client, watch the different traffic types be properly labelled

Can also confirm that the traces between proxy -> shard are properly connected now:

### DIFF
--- a/linera-base/src/tracing_opentelemetry.rs
+++ b/linera-base/src/tracing_opentelemetry.rs
@@ -158,16 +158,6 @@ pub fn init_with_opentelemetry(log_name: &str, otlp_endpoint: Option<&str>) {
         .with_sampler(opentelemetry_sdk::trace::Sampler::AlwaysOn)
         .build();
 
-    // Configure composite propagator for W3C TraceContext + Baggage
-    // This enables:
-    // 1. Trace context propagation across gRPC boundaries (connected traces in Tempo)
-    // 2. Baggage propagation for traffic_type labeling (organic vs synthetic)
-    let propagator = TextMapCompositePropagator::new(vec![
-        Box::new(TraceContextPropagator::new()),
-        Box::new(BaggagePropagator::new()),
-    ]);
-    global::set_text_map_propagator(propagator);
-
     init_with_tracer_provider(log_name, tracer_provider);
 }
 
--- a/linera-rpc/src/propagation.rs
+++ b/linera-rpc/src/propagation.rs
@@ -27,59 +27,84 @@
 //! // Use cx.with_baggage() to access baggage values
 //! ```
 
+use std::task::{Context as TaskContext, Poll};
+
+use futures::{future::BoxFuture, FutureExt};
 use opentelemetry::{
     global,
     propagation::{Extractor, Injector},
     Context,
 };
 use tonic::metadata::{MetadataKey, MetadataMap, MetadataValue};
+use tower::{Layer, Service};
 use tracing::warn;
 
-/// Wrapper for injecting context into tonic metadata.
-struct MetadataInjector<'a>(&'a mut MetadataMap);
+/// Baggage key for traffic type labeling.
+///
+/// Used to distinguish organic traffic from synthetic (benchmark) traffic.
+/// Valid values are "organic" and "synthetic".
+pub const TRAFFIC_TYPE_KEY: &str = "traffic_type";
 
-impl Injector for MetadataInjector<'_> {
-    fn set(&mut self, key: &str, value: String) {
-        match MetadataKey::from_bytes(key.as_bytes()) {
-            Ok(key) => match MetadataValue::try_from(&value) {
-                Ok(value) => {
-                    self.0.insert(key, value);
-                }
-                Err(error) => {
-                    warn!(
-                        value,
-                        error = format!("{error:#}"),
-                        "failed to parse metadata value"
-                    );
-                }
-            },
-            Err(error) => {
-                warn!(
-                    key,
-                    error = format!("{error:#}"),
-                    "failed to parse metadata key"
-                );
-            }
-        }
+/// Traffic type for normal production traffic.
+pub const TRAFFIC_TYPE_ORGANIC: &str = "organic";
+
+/// Traffic type for synthetic benchmark traffic.
+pub const TRAFFIC_TYPE_SYNTHETIC: &str = "synthetic";
+
+/// Traffic type when OpenTelemetry feature is disabled.
+pub const TRAFFIC_TYPE_UNKNOWN: &str = "unknown";
+
+/// Environment variable to override the traffic type.
+///
+/// Set this to "synthetic" to mark all outgoing requests as benchmark traffic.
+/// This is useful for benchmark tools that cannot easily set OpenTelemetry baggage.
+pub const TRAFFIC_TYPE_ENV_VAR: &str = "LINERA_TRAFFIC_TYPE";
+
+/// Tower layer that extracts OpenTelemetry context from incoming gRPC requests.
+///
+/// This layer extracts W3C TraceContext and Baggage headers from the request
+/// metadata and stores the extracted context in the request extensions.
+///
+/// # Usage
+///
+/// ```ignore
+/// use linera_rpc::propagation::OtelContextLayer;
+/// use tower::ServiceBuilder;
+///
+/// let service = ServiceBuilder::new()
+///     .layer(OtelContextLayer)
+///     .service(my_service);
+/// ```
+#[derive(Clone, Copy, Debug, Default)]
+pub struct OtelContextLayer;
+
+/// Service wrapper that extracts OpenTelemetry context from requests.
+#[derive(Clone, Debug)]
+pub struct OtelContextService<S> {
+    inner: S,
+}
+
+/// Extension type to store the extracted OpenTelemetry context.
+#[derive(Clone, Debug)]
+pub struct ExtractedOtelContext(pub Context);
+
+/// Trait for request types that can provide access to extracted OpenTelemetry context.
+///
+/// This trait abstracts over `http::Request` and `tonic::Request` to allow
+/// generic functions that work with either request type.
+pub trait HasOtelContext {
+    fn get_otel_context(&self) -> Option<&ExtractedOtelContext>;
+}
+
+impl<B> HasOtelContext for http::Request<B> {
+    fn get_otel_context(&self) -> Option<&ExtractedOtelContext> {
+        self.extensions().get::<ExtractedOtelContext>()
     }
 }
 
-/// Wrapper for extracting context from tonic metadata.
-struct MetadataExtractor<'a>(&'a MetadataMap);
-
-impl Extractor for MetadataExtractor<'_> {
-    fn get(&self, key: &str) -> Option<&str> {
-        self.0.get(key).and_then(|value| value.to_str().ok())
-    }
-
-    fn keys(&self) -> Vec<&str> {
-        self.0
-            .keys()
-            .filter_map(|key| match key {
-                tonic::metadata::KeyRef::Ascii(key) => Some(key.as_str()),
-                tonic::metadata::KeyRef::Binary(_) => None,
-            })
-            .collect()
+impl<T> HasOtelContext for tonic::Request<T> {
+    fn get_otel_context(&self) -> Option<&ExtractedOtelContext> {
+        self.extensions().get::<ExtractedOtelContext>()
     }
 }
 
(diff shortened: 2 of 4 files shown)


linera-base/src/tracing_opentelemetry.rs
```diff
@@ -145,6 +158,16 @@ pub fn init_with_opentelemetry(log_name: &str, otlp_endpoint: Option<&str>) {
         .with_sampler(opentelemetry_sdk::trace::Sampler::AlwaysOn)
         .build();
 
+    // Configure composite propagator for W3C TraceContext + Baggage
+    // This enables:
+    // 1. Trace context propagation across gRPC boundaries (connected traces in Tempo)
+    // 2. Baggage propagation for traffic_type labeling (organic vs synthetic)
+    let propagator = TextMapCompositePropagator::new(vec![
+        Box::new(TraceContextPropagator::new()),
```
deuszx: So this also enables this cross-machine trace correlation? Something we didn't have before

.gitignore
```diff
@@ -41,6 +41,7 @@ linera-web/**/*-lock.json
 .claude
 CLAUDE.md
 .serena
+docs/
```
deuszx, line 44: What generates this?

linera-service/src/proxy/grpc.rs
```diff
@@ -68,31 +75,41 @@ mod metrics {
...
+            &[TRAFFIC_TYPE_LABEL],
             linear_bucket_interval(1.0, 50.0, 2000.0),
         )
     });
     pub static PROXY_REQUEST_COUNT: LazyLock<IntCounterVec> = LazyLock::new(|| {
-        register_int_counter_vec("proxy_request_count", "Proxy request count", &[])
+        register_int_counter_vec(
+            "proxy_request_count",
+            "Proxy request count",
+            &[TRAFFIC_TYPE_LABEL],
```
deuszx, line 91: Why not count the method name?

linera-service/src/proxy/grpc.rs
```diff
@@ -68,31 +75,41 @@ mod metrics {
...
     pub static PROXY_REQUEST_LATENCY: LazyLock<HistogramVec> = LazyLock::new(|| {
         register_histogram_vec(
             "proxy_request_latency",
             "Proxy request latency",
-            &[],
+            &[TRAFFIC_TYPE_LABEL],
             linear_bucket_interval(1.0, 50.0, 2000.0),
         )
     });
     pub static PROXY_REQUEST_COUNT: LazyLock<IntCounterVec> = LazyLock::new(|| {
```
deuszx, line 87: if `PROXY_REQUEST_COUNT = PROXY_REQUEST_SUCCESS + PROXY_REQUEST_ERROR` then we don't need this metric.

linera-service/src/proxy/grpc.rs
```diff
@@ -126,18 +144,29 @@ where
         self.service.poll_ready(cx)
     }
 
-    fn call(&mut self, request: Req) -> Self::Future {
+    fn call(&mut self, request: http::Request<B>) -> Self::Future {
         #[cfg(with_metrics)]
         let start = linera_base::time::Instant::now();
+
+        // Extract traffic type from request extensions (set by OtelContextLayer).
+        // Falls back to "unknown" if opentelemetry feature is disabled.
```
deuszx, line 147: Can we make sure that we never run benchmarks w/o opentelemetry? If so, then there should never be the case when it's `unknown` .

.gitignore
ndr-ds: Claude generates this, to put design docs in

linera-base/src/tracing_opentelemetry.rs
ndr-ds: It's supposed to, yes. I'm trying to make it work. We technically had it before, but we weren't using it properly, I don't think

linera-service/src/proxy/grpc.rs
ndr-ds: Sure, but not all calls go through `log_and_return_proxy_request_outcome`. We currently log `PROXY_REQUEST_COUNT`on a service level layer, so everything will for sure go through it. But I guess if we count the success/error there we won't need it 🤔

linera-service/src/proxy/grpc.rs
ndr-ds: I have a different PR doing that https://github.com/linera-io/linera-protocol/pull/5213  
I guess I could close it and incorporate that functionality here 🤔

linera-service/src/proxy/grpc.rs
ndr-ds: Our Dockerfile builds with `opentelemetry` by default. And on the client side, if we try to run the benchmark client without it, we bail out.  
We should still leave the `unknown` there, because if something goes wrong, it will signal that to us.

ndr-ds:
That's an interesting thought 🤔 let me try

ndr-ds:
Actually, we can't do this, because we'll wrap the `Err` in a request that still has `grpc-status` as 0 (success). So this simplification won't work

linera-service/src/proxy/grpc.rs
ndr-ds: I'll just leave it as a separate PR, this PR is already too big

linera-rpc/src/propagation.rs
```diff
@@ -0,0 +1,387 @@
...
+
+use opentelemetry::{
+    global,
+    propagation::{Extractor, Injector},
+    Context,
+};
+use tonic::metadata::{MetadataKey, MetadataMap, MetadataValue};
+use tracing::warn;
+
+/// Wrapper for injecting context into tonic metadata.
```
deuszx: I always put pub types at the top. I prefer to see the types I'm most-likely looking for rather than implementation details.

linera-service/src/proxy/grpc.rs
```diff
@@ -483,10 +547,13 @@ where
         &self,
         request: Request<BlockProposal>,
     ) -> Result<Response<ChainInfoResult>, Status> {
+        let traffic_type = Self::extract_traffic_type(&request);
```
deuszx, line 548: is it possible to extract `traffic_type` from the `inner` value?

linera-service/src/proxy/grpc.rs
```diff
@@ -311,6 +345,35 @@ where
         }
     }
 
+    /// Extracts traffic type from request for metrics labeling.
+    /// Returns "organic", "synthetic", or "unknown".
+    #[cfg(feature = "opentelemetry")]
+    fn extract_traffic_type<R>(request: &Request<R>) -> &'static str {
```
deuszx, line 349: do you need this method at all? Why not call `get_traffic_type_from_tonic_request` directly

linera-service/src/proxy/grpc.rs
```diff
@@ -311,6 +345,35 @@ where
...
+        get_traffic_type_from_tonic_request(request)
+    }
+
+    /// Returns "unknown" traffic type when opentelemetry is disabled.
+    #[cfg(not(feature = "opentelemetry"))]
+    fn extract_traffic_type<R>(_request: &Request<R>) -> &'static str {
+        "unknown"
+    }
```
deuszx, line 357: Why not put this feature variant to `get_traffic_type_from_tonic_request` and get rid of these duplicates

linera-service/src/proxy/grpc.rs
deuszx: Ah, b/c `propagation.rs` has no notion of `opentelemtry` feature flag

linera-rpc/src/propagation.rs
```diff
@@ -0,0 +1,387 @@
...
+        .extensions()
+        .get::<ExtractedOtelContext>()
+        .map_or(TRAFFIC_TYPE_ORGANIC, |ext| get_traffic_type(&ext.0))
+}
+
+/// Gets the traffic type from a tonic::Request's extensions.
+///
+/// Returns "organic" if no context was extracted or no traffic type baggage was set.
+/// Tower middleware extensions are preserved in tonic::Request extensions.
+pub fn get_traffic_type_from_tonic_request<T>(request: &tonic::Request<T>) -> &'static str {
```
deuszx: Why do we need both `get_traffic_type_from_tonic_request` and `get_traffic_type_from_request`?

ndr-ds: Good point, I guess I can create a trait to simplify it

linera-service/src/proxy/grpc.rs
ndr-ds: No, inner is just the payload (e.g., BlockProposal).

linera-service/src/proxy/grpc.rs
```diff
@@ -68,31 +70,41 @@ mod metrics {
     };
     use prometheus::{HistogramVec, IntCounterVec};
 
+    /// Label for distinguishing organic vs synthetic (benchmark) traffic.
+    pub const TRAFFIC_TYPE_LABEL: &str = "traffic_type";
```
deuszx, line 74: You have this in `propagation.rs` as well.

linera-rpc/src/propagation.rs
```diff
@@ -0,0 +1,376 @@
...
+
+/// Service wrapper that extracts OpenTelemetry context from requests.
+#[derive(Clone, Debug)]
+pub struct OtelContextService<S> {
+    inner: S,
+}
+
+/// Extension type to store the extracted OpenTelemetry context.
+#[derive(Clone, Debug)]
+pub struct ExtractedOtelContext(pub Context);
```
deuszx, line 89: Could you work with just a `&'a Context`? it seems that the methods return a reference already

linera-rpc/src/propagation.rs
```diff
@@ -0,0 +1,376 @@
...
+    fn get_otel_context(&self) -> Option<&ExtractedOtelContext> {
+        self.extensions().get::<ExtractedOtelContext>()
+    }
+}
+
+/// Injects the OpenTelemetry context into tonic metadata.
+///
+/// This injects both W3C TraceContext (`traceparent`, `tracestate`) and
+/// W3C Baggage (`baggage`) headers into the metadata, enabling distributed
+/// tracing and baggage propagation across gRPC service boundaries.
```
deuszx, line 115: Can we still run non-benchmarked deployments w/o the _additional baggage_? To decrease the size of payloads.

deuszx (approved):
No blockers

linera-rpc/src/propagation.rs
ndr-ds: I don't think so, I think the request extensions require `static. Cloning it should be pretty cheap though

linera-rpc/src/propagation.rs
ndr-ds: Organic traffic should not have anything set 🤔  
And the overhead here is very small

deuszx: Small isn't non-zero. We are trying to optimise everything.

PR #5336, opened 2026-01-27
Fix controller logic - port of #5241

bart-linera:
## Motivation

Some parts of #5241 still haven't been ported to `main`.

## Proposal

Port the missing parts (most importantly: the fix in the `controller` example app logic).

## Test Plan

CI will catch regressions

## Release Plan

## Links

### DIFF
--- a/examples/controller/src/contract.rs
+++ b/examples/controller/src/contract.rs
@@ -95,7 +95,7 @@ impl Contract for ControllerContract {
                 } else {
                     log::info!(
                         "Sending worker command from {} to {creator_chain_id} \
-                                for remote execution: {owner} {command:?}",
+                        for remote execution: {owner} {command:?}",
                         self.runtime.chain_id()
                     );
                     self.runtime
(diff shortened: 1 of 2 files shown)


examples/controller/src/contract.rs
```diff
@@ -89,6 +93,11 @@ impl Contract for ControllerContract {
                     self.execute_worker_command_locally(owner, command, creator_chain_id)
                         .await;
                 } else {
+                    log::info!(
+                        "Sending worker command from {} to {creator_chain_id} \
+                                for remote execution: {owner} {command:?}",
```
afck:
```suggestion
                        for remote execution: {owner} {command:?}",
```

examples/controller/src/contract.rs
```diff
@@ -103,6 +112,11 @@ impl Contract for ControllerContract {
                     self.execute_controller_command_locally(admin, command)
                         .await;
                 } else {
+                    log::info!(
+                        "Sending controller command from {} to {creator_chain_id} \
+                                for remote execution: {admin} {command:?}",
```
afck:
```suggestion
                        for remote execution: {admin} {command:?}",
```

examples/controller/src/contract.rs
bart-linera: I saw your comments in the original PR, and then I made the same mistakes :man_facepalming: :sweat_smile:

PR #5341, opened 2026-01-27
Include the preferred owner in the `Listen` command

bart-linera:
## Motivation

When we tell the `ChainListener` to start listening to a chain, it is also important to tell it which owner's identity we should be using for the chain.

## Proposal

Include the identity in the `Listen` command. This is a port of the changes from #5241 , with some small cleanups.

## Test Plan

CI

## Release Plan

- These changes (or rather, the cleanups) should be backported to the latest `testnet` branch, then
    - be released in a new SDK

## Links


bart-linera:
I think that was the idea, but I also seem to remember that there was some reason making it not that straightforward to implement... unfortunately, I can't remember what that reason was at the moment :grimacing:

PR #5347, opened 2026-01-28
Add a basic test for the controller logic

bart-linera:
## Motivation

We currently have no end-to-end tests for the controller logic.

## Proposal

Add a test exercising worker registration and service assignment (written mostly by Claude).

## Test Plan

CI

## Release Plan

## Links

### DIFF
--- a/linera-service/src/controller.rs
+++ b/linera-service/src/controller.rs
@@ -274,6 +274,8 @@ where
             .await
             .make_chain_client(service_chain_id)
             .await?;
+        // The processor may need to propose blocks with task results - for that, it will
+        // need the chain client to be configured with a preferred owner.
         if let Some(owner) = self.chain_client.preferred_owner() {
             chain_client.set_preferred_owner(owner);
         }


linera-service/tests/linera_net_tests.rs
```diff
@@ -4995,3 +4998,232 @@ async fn test_end_to_end_repeated_transfers(config: impl LineraNetConfig) -> Res
...
+        )
+        .await?;
+
+    node_service1.process_inbox(&worker1_chain).await?;
+
+    let app1 = node_service1.make_application(&worker1_chain, &controller_id)?;
+
+    let mut worker1_registered = false;
+    for _ in 0..30 {
+        linera_base::time::timer::sleep(Duration::from_secs(1)).await;
```
afck: Can't we wait for a `NewBlock` or `NewIncomingBundle` notification instead, like we do in most other tests now? (Also below.)

linera-service/tests/linera_net_tests.rs
```diff
@@ -4995,3 +4998,235 @@ async fn test_end_to_end_repeated_transfers(config: impl LineraNetConfig) -> Res
...
+    assert!(
+        state.local_worker.is_some(),
+        "Worker 2 should be registered after block notification"
+    );
+
+    let admin_app = admin_node_service.make_application(&admin_chain, &controller_id)?;
+    let response = admin_app.query("workers { keys }").await?;
+    let worker_keys: Vec<String> = serde_json::from_value(response["workers"]["keys"].clone())?;
+    assert!(
+        worker_keys.len() >= 2,
```
afck: Aren't we expecting _exactly_ 2?

linera-service/tests/linera_net_tests.rs
```diff
@@ -4995,3 +4998,235 @@ async fn test_end_to_end_repeated_transfers(config: impl LineraNetConfig) -> Res
...
+        "Expected at least 2 workers, got {}",
+        worker_keys.len()
+    );
+
+    assert_ne!(worker1_chain, admin_chain);
+    assert_ne!(worker2_chain, admin_chain);
+    assert_ne!(worker1_chain, worker2_chain);
```
afck, line 5184: (That's basically testing `open_and_assign`?)

linera-service/tests/linera_net_tests.rs
```diff
@@ -4995,3 +4998,235 @@ async fn test_end_to_end_repeated_transfers(config: impl LineraNetConfig) -> Res
...
+        .unwrap_or_else(|_| {
+            panic!("should receive a notification about a block on chain {admin_chain}")
+        });
+
+    // Worker 1 chain block 2: receive service removal
+    notifications1
+        .wait_for_block(BlockHeight::from(2))
+        .await
+        .unwrap_or_else(|_| {
+            panic!("should get notification about a block on chain {worker1_chain}")
```
afck, line 5295: These make the test _very_ verbose. I'd just `unwrap()`, and maybe put this information (which chain, and maybe which block height) directly into the error returned from `wait_for_block`. (Also above.)

linera-service/tests/linera_net_tests.rs
bart-linera: Right, good point :+1:

linera-service/tests/linera_net_tests.rs
bart-linera: I was wondering if I should leave those, but I decided that this also helps make sure that we haven't messed up the test code and we are in fact testing on three separate chains. I can remove those lines if you think that's not really necessary.

linera-service/tests/linera_net_tests.rs
bart-linera: That would be a good idea, but I'm not sure if it's easy to do - the notifications object is just a stream which, AFAICT, doesn't remember which chain it was constructed for, so we'd have to wrap it in an additional struct. It's of course doable, but I'm not sure if it's worth the hassle (but the addition of the chain IDs to the message itself has helped me a lot in debugging the test).

linera-service/src/controller.rs
```diff
@@ -268,12 +268,15 @@ where
...
-        let chain_client = self
+        let mut chain_client = self
             .context
             .lock()
             .await
             .make_chain_client(service_chain_id)
             .await?;
+        if let Some(owner) = self.chain_client.preferred_owner() {
+            chain_client.set_preferred_owner(owner);
+        }
```
afck, line 281: We could add a comment why that's necessary.

PR #5349, opened 2026-01-28
fix(Makefile): extract correct application ID from publish-and-create output

eldios:
## Summary

- Fix deploy-app-counter and deploy-app-fungible targets extracting chain ID instead of application ID
- Remove `2>&1` redirection that merged stderr (containing chain ID in log messages) with stdout
- Change `head -1` to `tail -1` to capture the last 64-char hex string (the actual application ID)

## Root Cause

The `linera publish-and-create` command:
1. Logs `info!("Publishing and creating application on chain {}", publisher)` to **stderr**
2. Prints `println!("{}", application_id)` to **stdout**

Using `2>&1 | grep -oE '[a-f0-9]{64}' | head -1` merged both streams and captured the chain ID (first match) instead of the application ID.

## Test plan

- [x] Run `make deploy-app-counter` and verify COUNTER_APP_ID differs from CHAIN
- [x] Run `make deploy-app-fungible` and verify FUNGIBLE_APP_ID differs from CHAIN
- [x] Verify deployed apps work correctly with the frontend demos

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -137,10 +137,10 @@ init-wallet: setup ## Initialize wallet and request chain from faucet
 	@$(LINERA_BIN) wallet init \
 			--faucet https://faucet.$(NETWORK_NAME).linera.net
 	@printf "$(YELLOW)⛓️  Requesting chain...$(NC)\n"
-	@INFO=$$($(LINERA_BIN) wallet request-chain \
-			--faucet https://faucet.$(NETWORK_NAME).linera.net | tr '\n' ' ') && \
-		CHAIN=$$(echo "$$INFO" | awk '{print $$1}') && \
-		OWNER=$$(echo "$$INFO" | awk '{print $$2}') && \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -186,7 +186,7 @@ deploy-app-counter: build-wasm-counter ## Deploy counter app to blockchain
 		cd $(EXAMPLES_DIR)/counter && \
 		COUNTER_APP_ID=$$(../../$(LINERA_BIN) publish-and-create \
 			../target/wasm32-unknown-unknown/release/counter_{contract,service}.wasm \
-			--json-argument "1" 2>&1 | grep -oE '[a-f0-9]{64}' | head -1) && \
+			--json-argument "1" | grep -oE '[a-f0-9]{64}' | tail -1) && \
```
ma2bd: there is no need for the grep and the tail. See (most of) our README files in `examples/`

eldios: okay cleaned it up. Thanks for catching that.

PR #5351, opened 2026-01-28
`linera-sdk`: remove `yield_once` from the system view API

Twey:
## Motivation

https://github.com/linera-io/linera-protocol/pull/5318 attempts to save gas by removing `async` from our storage mechanisms in the `linera-sdk`, making a copy of `linera-views` that is exactly the same but with `async` removed (since all our syscalls, including storage syscalls, are synchronous on the guest side).  However, this entails a large amount of code duplication.  Instead, the compiler should be able to optimize away the async code when the implementation is known to not yield.

So I went digging for what was stopping the async machinery being folded away in `linera-sdk` and I found [this](https://github.com/linera-io/linera-protocol/blob/main/linera-sdk/src/views/system_api.rs): the reason our SDK futures weren't being optimized away when they're actually sync is that they weren't actually sync, they all contain a yield_once future that forces them to yield up through the async machinery to our blocking_wait loop (which, of course, immediately polled them again).  Without this there's no async in the resulting code after code folding (i.e. no poll function is ever called) even without making any changes to the `linera-views` traits.

## Proposal

Remove `yield_once` from `linera-sdk`.

## Test Plan

Checked locally that no `poll` function is ever called after this change, using `cargo rustc -- --emit=asm`.

## Release Plan

## Links

### DIFF
--- a/linera-sdk/src/views/system_api.rs
+++ b/linera-sdk/src/views/system_api.rs
@@ -21,7 +21,6 @@ use crate::{
         contract_runtime_api::{self, WriteOperation},
     },
     service::wit::base_runtime_api as service_wit,
-    util::yield_once,
 };
 
 /// We need to have a maximum key size that handles all possible underlying
@@ -121,7 +120,6 @@ impl ReadableKeyValueStore for KeyValueStore {
             KeyValueStoreError::KeyTooLong
         );
         let promise = self.wit_api.contains_key_new(key);
-        yield_once().await;
         Ok(self.wit_api.contains_key_wait(promise))
     }
 
@@ -133,7 +131,6 @@ impl ReadableKeyValueStore for KeyValueStore {
             );
         }
         let promise = self.wit_api.contains_keys_new(keys);
-        yield_once().await;
         Ok(self.wit_api.contains_keys_wait(promise))
     }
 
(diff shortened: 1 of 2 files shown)


linera-sdk/src/views/system_api.rs
ma2bd, line 104: Should we let SDK users decide if they want a KeyValueStore with promises?

Twey: You mean in the sense of an async runtime, once that API is available? 

I don't think that `linera-views` is currently equipped to generalize over promises (as opposed to Rust futures). They sound similar at a high level but semantically they're very different; e.g. (in one design) promises have no `poll`, only a primitive `join`. Unifying values with futures is obvious (as we do here), and unifying values with promises is likewise straightforward, but unifying futures with promises is less natural to me, or at least there are several ways to do it and off the top of my head it's not obvious which is best.

ma2bd: I meant choosing between what we have now vs this PR

ma2bd: This begin said, this PR would already be a clear win for the `testnet_conway` branch as it is.

Twey: This PR doesn't change the user-visible API of the SDK.  It just removes a (completely vestigial, as far as I can see) yield.

ma2bd (approved):
unblocking the quick fix

PR #5352, opened 2026-01-29
Filter notifications by listening mode in `process_notification`.

afck:
## Motivation

This check whether a notification is relevant depending on the listening mode was added in https://github.com/linera-io/linera-protocol/pull/5232 but missed in the backport https://github.com/linera-io/linera-protocol/pull/5237.

## Proposal

Add it.

## Test Plan

This should not have caused any failures, only unnecessary downloads.

## Release Plan

## Links

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -4086,6 +4086,7 @@ impl<Env: Environment> ChainClient<Env> {
             debug!(
                 chain_id = %self.chain_id,
                 reason = ?notification.reason,
+                listening_mode = ?self.listening_mode(),
                 "Ignoring notification due to listening mode"
             );
             return Ok(());


linera-core/src/client/mod.rs
```diff
@@ -4079,6 +4079,17 @@ impl<Env: Environment> ChainClient<Env> {
...
         notification: Notification,
     ) -> Result<(), ChainClientError> {
+        let dominated = self
+            .listening_mode()
+            .is_none_or(|mode| !mode.is_relevant(&notification.reason));
+        if dominated {
+            debug!(
+                chain_id = %self.chain_id,
+                reason = ?notification.reason,
+                "Ignoring notification due to listening mode"
```
deuszx, line 4090: Can we log the listening mode as well?

afck: Done in https://github.com/linera-io/linera-protocol/pull/5352/commits/63e7523b3f4e486faa53725d8de25407b1c05d5b.

PR #5356, opened 2026-01-29
Enable HTTP/2 keepalive on the proxy's public gRPC server

afck:
## Motivation

We are missing notifications in some cases.

## Proposal

Configure HTTP/2 and TCP keepalive also on the server. (It's already configured on the client.)

## Test Plan

- Upgrade the validators; see if we still miss notifications.

## Release Plan

## Links


ma2bd (changes requested):
I understand we may not need this

PR #5357, opened 2026-01-29
Log every notification in the client.

afck:
## Motivation

We are missing notifications in some cases.

## Proposal

Log every notification in the client.

We could enable them by adding `linera_core::client=debug` to `RUST_LOG`.

(Also remove unnecessary `tracing::` qualification, since `debug` is already imported.)

## Test Plan

N/A

## Release Plan

- Run a client with debug logging and see if this log shows up for every expected notification.
- Port to `main`.

## Links

- Related: #5356


afck:
I think that's not needed for now.

PR #5358, opened 2026-01-29
PM benchmark client

ndr-ds:
## Motivation

Add support for benchmarking PM (Prediction Market) order submission to measure performance under load. This enables stress testing the prediction market infrastructure with configurable order patterns.

## Proposal

Adds a new `linera benchmark pm` subcommand that:

- Creates benchmark chains and funds them via the standard Linera faucet
- Claims BASE tokens from the PM faucet for each benchmark chain
- Generates PM orders (YES/NO bids) and submits them to the market chain
- Supports configurable order patterns:
- `matching`: Generates complementary YES+NO bid pairs that will match
- `mixed`: Configurable mix of matching and non-matching orders

AMM not supported yet.

Key changes:
- New `BenchmarkMode` enum to distinguish between native transfer, fungible transfer, and PM benchmarks
- `PmOrderGenerator` for generating order book operations with realistic price distributions (does not query and follow the actual market value)
- Refactored `BenchmarkOperationGenerator` (renamed from `ChainDestinationManager`) to handle different operation
types

## Test Plan

- Manual testing against deployed PM infrastructure on andre-test-c77l3j network
- Verified benchmark chains are created and funded correctly
- Confirmed orders are submitted at the target BPS rate

### DIFF
--- a/linera-client/src/benchmark.rs
+++ b/linera-client/src/benchmark.rs
@@ -23,7 +23,7 @@ use linera_execution::{system::SystemOperation, Operation};
 use linera_sdk::abis::fungible::{self, FungibleOperation};
 use num_format::{Locale, ToFormattedString};
 use prometheus_parse::{HistogramCount, Scrape, Value};
-use rand::{seq::SliceRandom, thread_rng};
+use rand::{rngs::SmallRng, seq::SliceRandom, thread_rng, Rng, SeedableRng};
 use serde::{Deserialize, Serialize};
 use tokio::{
     sync::{mpsc, Barrier, Notify},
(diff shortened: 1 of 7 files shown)


linera-client/src/benchmark.rs
```diff
@@ -34,6 +34,228 @@ use tracing::{debug, error, info, warn, Instrument as _};
...
+                market_chain_id, ..
+            } => Some(*market_chain_id),
+            _ => None,
+        }
+    }
+}
+
+/// Pattern for generating PM orders (for CLI parsing).
+///
+/// These types mirror pm-engine but are defined here to avoid a dependency on pm-app.
```
ma2bd, line 75: This is not ideal. We're still changing the APIs

ndr-ds:
Will take a different approach with this, so closing

PR #5359, opened 2026-01-30
Retry on h2 protocol error Internal gRPC error

ndr-ds:
## Motivation

When validators (proxy/shards) restart, long-running gRPC notification streams terminate with an HTTP/2 protocol
error. This error has status code `Internal` with message `"h2 protocol error: error reading a body from connection"`.

Currently, `Internal` errors are not retryable, causing clients (like pm-infra workers) to permanently lose their
notification subscriptions after a validator restart, even if `maxRetries` is set to a high value.

## Proposal

Make HTTP/2 protocol errors retryable by adding a special case in `is_retryable()` for `Code::Internal` errors that
contain `"h2 protocol error"` in their message.

These errors are transient network issues (connection forcibly closed by server restart), not actual internal server
errors, and should be retried like other transient errors (`Unavailable`, `DeadlineExceeded`, etc.).

## Test Plan

1. Deployed a pm-infra worker with the fix against a test network
2. Restarted validator proxy/shards while worker was subscribed to notifications
3. Verified worker reconnected and continued receiving notifications after validator came back up
4. Previously, the same scenario caused the worker to permanently stop receiving notifications

### DIFF
--- a/linera-rpc/src/grpc/client.rs
+++ b/linera-rpc/src/grpc/client.rs
@@ -80,6 +80,13 @@ impl GrpcClient {
                 trace!("Unexpected gRPC status: {status:?}; retrying");
                 true
             }
+            Code::Internal if status.message().contains("h2 protocol error") => {
+                // HTTP/2 connection reset errors are transient network issues, not real
+                // internal errors. This happens when the server restarts and the
+                // connection is forcibly closed.
+                trace!("gRPC connection reset: {status:?}; retrying");
+                true
+            }
             Code::NotFound => false, // This code is used if e.g. the validator is missing blobs.
             Code::InvalidArgument
             | Code::AlreadyExists


linera-rpc/src/grpc/client.rs
```diff
@@ -78,6 +78,13 @@ impl GrpcClient {
                 trace!("Unexpected gRPC status: {status:?}; retrying");
                 true
             }
+            Code::Internal if status.message().contains("h2 protocol error") => {
```
ma2bd: Should we retry all "Internal" errors?

Twey: In theory we should never retry any `Internal` errors.  From [what I can see](https://github.com/grpc/grpc/blob/master/doc/statuscodes.md#status-codes-and-their-use-in-grpc), `Internal` is usually supposed to mean that there's a bug in the protocol implementation on the client or server.  I wonder why we're getting it on a bog-standard connection drop.

Twey: I guess this is a known bug in `tonic`: @LucioFranco wrote [a FIXME comment about it](https://github.com/hyperium/tonic/blob/ed4bafc35af7a284357cf4e77e26b9c3eb0266ca/tonic/src/status.rs#L382-L410) in the Tonic source.

ndr-ds: Yes, what @Twey said. Generally Internal errors are not supposed to be retriable... but we're for some reason getting that specific one on regular temporary connection drops 🤔

Twey: IMHO the problem is that Tonic can report errors from either gRPC or the underlying transport, but the error type that Tonic exposes is tied to the gRPC error codes.  So Tonic has had to force the entire space of possible transport errors into the limited set of gRPC errors, and it somewhat arbitrarily chose `Internal` for connection errors.

I guess this will change when the `transport` module [is removed](https://github.com/hyperium/tonic/issues/891#issuecomment-1092011690), since it will no longer be possible to assume that transport errors are h2 errors.

Twey: tl;dr what @ndr-ds wrote is exactly the right hack. Though we might want to be pickier about the [error strings](https://docs.rs/h2/latest/src/h2/frame/reason.rs.html#65-82).  I think we want to retry:
- [`CONNECT_ERROR`](https://docs.rs/h2/latest/h2/struct.Reason.html#associatedconstant.CONNECT_ERROR) (`"connection established in response to a CONNECT request was reset"`)
- [`SETTINGS_TIMEOUT`](https://docs.rs/h2/latest/h2/struct.Reason.html#associatedconstant.SETTINGS_TIMEOUT) (`"settings ACK not received in timely manner"`)
- and maybe [`ENHANCE_YOUR_CALM`](https://docs.rs/h2/latest/h2/struct.Reason.html#associatedconstant.ENHANCE_YOUR_CALM) (`"detected excessive load generating behavior"`) — but only with an appropriate backoff!

Needless to say none of these strings are protected by semver so we may see changes in our code's behaviour with upgrades to `tonic` or `h2`. :confounded:

ma2bd: > In theory we should never retry any Internal errors.

This is not clear to me at all. Some people seem to retry "500 internal server" errors. The main concern is the risk to re-execute non-idempotent operations but we are not concerned by this.

ma2bd: Do you see any concrete problem with retrying all GRPC "internal" error by default?

ndr-ds: I didn't see anything explicitly saying they're not supposed to be retriable, but the [documentation](https://grpc.github.io/grpc/core/md_doc_statuscodes.html) says it's reserved for "serious errors", and the examples they give don't seem like retryable errors...  
I feel like we shouldn't retry for everything, because it might add significant latency when we hit non-retriable errors, but that's just my 2c.

ma2bd: So when tonic changes the error string, we find ourselves debugging missing notifications again?

ndr-ds: Yeah, that's the tradeoff. I guess we can just retry all `Internal` errors for now, until they fix things

ma2bd: OK I'm ok either way

PR #5365, opened 2026-01-31
Limit number of message bundle drops per staged block.

afck:
## Motivation

If there are many incoming messages from many different senders, and the client's default limit for incoming bundles per block is set too high, the staging loop will run over and over, dropping one sender's message after another, as they keep hitting the block limit.

## Proposal

After three (configurable!) failures due to block limits, drop _all_ remaining messages.

## Test Plan

We should try this with the benchmark that are currently seeing a high latency.

## Release Plan

- Release an SDK.

## Links


afck:
Closing in favor of #5367.

PR #5367, opened 2026-01-31
Cache execution state when staging blocks.

afck:
## Motivation

When we discard a transaction during block staging, we restart block execution from the start, which is wasteful and causes long delays.

## Proposal

Cache the execution state after each transaction and restore it if the next one fails.

## Test Plan

Tests have been added.

## Release Plan

- Release SDK.

## Links

- Closes https://github.com/linera-io/linera-protocol/issues/1401.

### DIFF
--- a/linera-chain/src/chain.rs
+++ b/linera-chain/src/chain.rs
@@ -700,7 +700,7 @@ where
         replaying_oracle_responses: Option<Vec<Vec<OracleResponse>>>,
         exec_policy: BundleExecutionPolicy,
     ) -> Result<(BlockExecutionOutcome, ResourceTracker), ChainError> {
-        // AutoRetry is incompatible with replaying oracle responses because removing or
+        // AutoRetry is incompatible with replaying oracle responses because discarding or
         // rejecting bundles would change which transactions execute.
         if !matches!(exec_policy, BundleExecutionPolicy::Abort) {
             assert!(
@@ -776,7 +776,7 @@ where
 
             // If the transaction executed successfully, we move on to the next one.
             // On transient errors (e.g. missing blobs) we fail, so it can be retried after
-            // syncing. In auto-retry mode, we can remove or reject message bundles that failed
+            // syncing. In auto-retry mode, we can discard or reject message bundles that failed
             // with non-transient errors.
             let (error, context, incoming_bundle, saved_chain, saved_tracker) =
                 match (result, transaction, checkpoint) {
@@ -800,26 +800,26 @@ where
 
             if error.is_limit_error() && i > 0 {
                 failure_count += 1;
-                // If we've exceeded max failures, remove all remaining message bundles.
+                // If we've exceeded max failures, discard all remaining message bundles.
                 let maybe_sender = if failure_count > max_failures {
                     info!(
                         failure_count,
                         max_failures,
-                        "Exceeded max bundle failures, removing all remaining message bundles"
+                        "Exceeded max bundle failures, discarding all remaining message bundles"
                     );
                     None
                 } else {
-                    // Not the first - remove it and same-sender subsequent bundles.
+                    // Not the first - discard it and same-sender subsequent bundles.
                     info!(
                         %error,
                         index = i,
                         origin = %incoming_bundle.origin,
-                        "Message bundle exceeded block limits and will be removed for \
+                        "Message bundle exceeded block limits and will be discarded for \
                         retry in a later block"
                     );
                     Some(incoming_bundle.origin)
                 };
-                Self::remove_remaining_bundles(block, i, maybe_sender);
+                Self::discard_remaining_bundles(block, i, maybe_sender);
                 // Continue without incrementing i (next transaction is now at i).
             } else if incoming_bundle.bundle.is_protected()
                 || incoming_bundle.action == MessageAction::Reject
@@ -840,7 +840,7 @@ where
             }
         }
 
-        // This can only happen if all transactions were incoming bundles that all got removed
+        // This can only happen if all transactions were incoming bundles that all got discarded
         // due to resource limit errors. This is unlikely in practice but theoretically possible.
         ensure!(!block.transactions.is_empty(), ChainError::EmptyBlock);
 
@@ -914,8 +914,8 @@ where
         ))
     }
 
-    /// Removes all bundles from the given origin (or all if `None`), starting at the given index.
-    fn remove_remaining_bundles(
+    /// Discards all bundles from the given origin (or all if `None`), starting at the given index.
+    fn discard_remaining_bundles(
         block: &mut ProposedBlock,
         mut index: usize,
         maybe_origin: Option<ChainId>,
@@ -936,11 +936,11 @@ where
     /// Executes a block with a specified policy for handling bundle failures.
     ///
     /// This method supports automatic retry with checkpointing when bundles fail:
-    /// - For limit errors (block too large, fuel exceeded, etc.): the bundle is removed
+    /// - For limit errors (block too large, fuel exceeded, etc.): the bundle is discarded
     ///   so it can be retried in a later block, unless it's the first transaction
     ///   (which gets rejected as inherently too large).
     /// - For non-limit errors: the bundle is rejected (triggering bounced messages).
-    /// - After `max_failures` failed bundles, all remaining message bundles are removed.
+    /// - After `max_failures` failed bundles, all remaining message bundles are discarded.
     ///
     /// The block may be modified to reflect the actual executed transactions.
     #[instrument(skip_all, fields(
--- a/linera-chain/src/data_types/mod.rs
+++ b/linera-chain/src/data_types/mod.rs
@@ -325,15 +325,15 @@ pub enum BundleExecutionPolicy {
     /// Automatically handle failing bundles with checkpointing and retry.
     ///
     /// This policy is intended for use by clients when preparing proposals. It modifies
-    /// the proposal by removing or rejecting bundles that fail to execute:
+    /// the proposal by discarding or rejecting bundles that fail to execute:
     ///
-    /// - For limit errors (block too large, fuel exceeded, etc.): remove the bundle
+    /// - For limit errors (block too large, fuel exceeded, etc.): discard the bundle
     ///   so it can be retried in a later block, unless it's the first transaction
     ///   (in which case it's inherently too large and gets rejected).
     /// - For non-limit errors: reject the bundle (triggering bounced messages).
-    /// - After `max_failures` removed bundles, remove all remaining message bundles.
+    /// - After `max_failures` discarded bundles, discard all remaining message bundles.
     AutoRetry {
-        /// Maximum number of removed bundles before removing all remaining message bundles.
+        /// Maximum number of discarded bundles before discarding all remaining message bundles.
         max_failures: u32,
     },
 }
--- a/CLI.md
+++ b/CLI.md
@@ -134,7 +134,9 @@ Client implementation and command-line tool for the Linera blockchain
 * `--max-pending-message-bundles <MAX_PENDING_MESSAGE_BUNDLES>` — The maximum number of incoming message bundles to include in a block proposal
 
   Default value: `10`
-* `--max-block-limit-errors <MAX_BLOCK_LIMIT_ERRORS>` — Maximum number of message bundles to remove from a block proposal due to block limit errors before removing all remaining bundles
+* `--max-block-limit-errors <MAX_BLOCK_LIMIT_ERRORS>` — Maximum number of message bundles to discard from a block proposal due to block limit errors before discarding all remaining bundles.
+
+   Discarded bundles can be retried in the next block.
 
   Default value: `3`
 * `--chain-worker-ttl-ms <CHAIN_WORKER_TTL>` — The duration in milliseconds after which an idle chain worker will free its memory
--- a/linera-client/src/client_options.rs
+++ b/linera-client/src/client_options.rs
@@ -55,8 +55,10 @@ pub struct Options {
     #[arg(long, default_value = "10")]
     pub max_pending_message_bundles: usize,
 
-    /// Maximum number of message bundles to remove from a block proposal due to block limit
-    /// errors before removing all remaining bundles.
+    /// Maximum number of message bundles to discard from a block proposal due to block limit
+    /// errors before discarding all remaining bundles.
+    ///
+    /// Discarded bundles can be retried in the next block.
     #[arg(long, default_value = "3")]
     pub max_block_limit_errors: u32,
 
--- a/linera-core/src/client/chain_client/mod.rs
+++ b/linera-core/src/client/chain_client/mod.rs
@@ -86,8 +86,10 @@ use crate::{
 pub struct Options {
     /// Maximum number of pending message bundles processed at a time in a block.
     pub max_pending_message_bundles: usize,
-    /// Maximum number of loaded message bundles to remove from a block proposal due to
-    /// block limit errors before removing all remaining bundles.
+    /// Maximum number of message bundles to discard from a block proposal due to block limit
+    /// errors before discarding all remaining bundles.
+    ///
+    /// Discarded bundles can be retried in the next block.
     pub max_block_limit_errors: u32,
     /// The policy for automatically handling incoming messages.
     pub message_policy: MessagePolicy,


linera-chain/src/chain.rs
```diff
@@ -684,26 +684,29 @@ where
...
+    async fn execute_block_inner_with_policy(
         chain: &mut ExecutionStateView<C>,
         confirmed_log: &LogView<C, CryptoHash>,
-        block: &ProposedBlock,
+        block: &mut ProposedBlock,
         local_time: Timestamp,
         round: Option<u32>,
         published_blobs: &[Blob],
         replaying_oracle_responses: Option<Vec<Vec<OracleResponse>>>,
+        policy: BundleExecutionPolicy,
```
bart-linera: Maybe `bundle_policy`? So that both policies (this one and `committee_policy`) used in the function are named something else than just `policy`.

linera-chain/src/chain.rs
```diff
@@ -739,12 +740,136 @@ where
...
 
-        for transaction in block.transaction_refs() {
-            block_execution_tracker
-                .execute_transaction(transaction, round, chain)
-                .await?;
+        // Extract max_failures from policy.
+        let max_failures = match policy {
+            BundleExecutionPolicy::Abort => 0,
+            BundleExecutionPolicy::AutoRetry { max_failures } => max_failures,
+        };
```
bart-linera: Since `AutoRetry` with `max_failures == 0` is possible, and it will work essentially the same as `Abort`, only with (unnecessarily) saving a checkpoint before processing a bundle, maybe the whole policy should just be collapsed down to something like `max_bundle_retries`? And then below `auto_retry` could just be defined as `max_bundle_retries != 0`.

linera-chain/src/chain.rs
```diff
@@ -684,26 +684,29 @@ where
         optional_vec.ok_or_else(|| ChainError::InternalError("Missing outboxes".into()))
     }
 
-    /// Executes a block: first the incoming messages, then the main operation.
-    /// Does not update chain state other than the execution state.
+    /// Executes a block with a specified policy for handling bundle failures.
+    ///
+    /// Unlike `execute_block_inner`, this method supports automatic retry with checkpointing.
```
bart-linera: Is there an `execute_block_inner` left after this change?

linera-core/src/chain_worker/state.rs
```diff
@@ -1497,6 +1512,54 @@ where
...
+        policy: BundleExecutionPolicy,
+    ) -> Result<(Block, ChainInfoResponse, ResourceTracker), WorkerError> {
+        self.initialize_and_save_if_needed().await?;
+        let local_time = self.storage.clock().current_time();
+        let (_, committee) = self.chain.current_committee()?;
+        block.check_proposal_size(committee.policy().maximum_block_proposal_size)?;
+
+        self.chain
+            .remove_bundles_from_inboxes(block.timestamp, true, block.incoming_bundles())
+            .await?;
```
bart-linera, line 1521: Should this happen _after_ the execution below, so that we only remove the bundles that have actually been executed?

linera-chain/src/chain.rs
afck: No, it's different: `Abort` means we don't modify the block, and just return any error. `AutoRetry { max_failures: 0 }` means: As soon as we hit the first block limit, we remove _all_ remaining incoming bundles from the block. (And not just that particular sender's.)

linera-core/src/chain_worker/state.rs
afck: The worker only uses `Abort` mode: It never modifies proposals.

bart-linera: Not sure if I understand. Do you mean the policy is always `Abort` here? If yes, then why pass it as a parameter to the function at all?

afck: Ah, sorry, you're right.

But when staging we're not persisting the chain state anyway, so this line is only here to check whether we actually _have_ the included bundles. I think it's fine to fail early here if not, even if some of the bundles would be removed from the block again due to block limits.

linera-chain/src/data_types/mod.rs
```diff
@@ -316,6 +316,28 @@ pub enum MessageAction {
...
+    /// This policy is intended for use by clients when preparing proposals. It modifies
+    /// the proposal by removing or rejecting bundles that fail to execute:
+    ///
+    /// - For limit errors (block too large, fuel exceeded, etc.): remove the bundle
+    ///   so it can be retried in a later block, unless it's the first transaction
+    ///   (in which case it's inherently too large and gets rejected).
+    /// - For non-limit errors: reject the bundle (triggering bounced messages).
+    /// - After `max_failures` removed bundles, remove all remaining message bundles.
+    AutoRetry {
+        /// Maximum number of removed bundles before removing all remaining message bundles.
```
deuszx: Remove remaining message bundles from the inbox, right? What could be the cause for the repeated failures?

linera-chain/src/unit_tests/chain_tests.rs
```diff
@@ -55,6 +55,25 @@ impl ChainStateView<MemoryContext<TestExecutionRuntimeContext>> {
             .await
             .expect("Loading from memory should work")
     }
+
+    /// Test helper that calls `execute_block` with default test parameters.
```
deuszx: What are those "test parameters"? Can you put that in the comment? Previously the `None`s were easy to spot

linera-chain/src/block_tracker.rs
```diff
@@ -367,17 +363,65 @@ impl<'resources, 'blobs> BlockExecutionTracker<'resources, 'blobs> {
         self.resource_controller
     }
 
+    /// Saves a checkpoint of the tracker's mutable state.
+    ///
+    /// This captures all state that could be modified during transaction execution,
+    /// allowing restoration if execution fails.
+    pub fn save_checkpoint(&self) -> TrackerCheckpoint {
```
deuszx: It's more like "create checkpoint" – b/c this method doesn't save anything.

linera-chain/src/chain.rs
```diff
@@ -739,12 +747,103 @@ where
...
-        for transaction in block.transaction_refs() {
-            block_execution_tracker
-                .execute_transaction(transaction, round, chain)
-                .await?;
+        // Extract max_failures from exec_policy.
+        let max_failures = match exec_policy {
+            BundleExecutionPolicy::Abort => 0,
+            BundleExecutionPolicy::AutoRetry { max_failures } => max_failures,
+        };
+        let auto_retry = !matches!(exec_policy, BundleExecutionPolicy::Abort);
```
deuszx, line 755: Why not `max_failures > 0` ?

linera-chain/src/chain.rs
```diff
@@ -739,12 +747,103 @@ where
...
         )?;
 
-        for transaction in block.transaction_refs() {
-            block_execution_tracker
-                .execute_transaction(transaction, round, chain)
-                .await?;
+        // Extract max_failures from exec_policy.
+        let max_failures = match exec_policy {
+            BundleExecutionPolicy::Abort => 0,
+            BundleExecutionPolicy::AutoRetry { max_failures } => max_failures,
```
deuszx, line 753: I guess this just shows we don't really need the `Abort` variant?

linera-execution/src/lib.rs
```diff
@@ -441,6 +441,18 @@ impl ExecutionError {
                 | ExecutionError::BlockTooLarge
         )
     }
+
+    /// Returns whether this is a transient error that may resolve after syncing.
+    ///
+    /// Transient errors like missing blobs or events might succeed after the node syncs
+    /// with the network. These errors should fail the block entirely (not reject the message)
+    /// so the block can be retried later.
+    pub fn is_transient_error(&self) -> bool {
```
deuszx, line 450: Doesn't seem like this needs to be part of the PR?

linera-core/src/client/mod.rs
```diff
@@ -1557,94 +1555,69 @@ impl<Env: Environment> Client<Env> {
     /// If a message fails due to block limits being exceeded and it's not the first
     /// transaction, the message is removed (not rejected) so it can be retried in a later
     /// block. All subsequent messages from the same sender are also removed.
-    // TODO(#2806): Measure how failing messages affect the execution times.
+    ///
+    /// Returns the modified block (with rejected/removed bundles) and the execution result.
     #[tracing::instrument(level = "trace", skip(self, block))]
     async fn stage_block_execution_and_discard_failing_messages(
```
deuszx: Do we need this method? It just bypasses to `stage_block_execution_with_policy`

linera-chain/src/data_types/mod.rs
afck: No, from the block. Precisely so they _stay_ in the inbox.

I.e. if we hit 3 (by default) block limit errors (e.g. running out of gas) and already removed three senders' messages from the block, at that point we decide to remove _all_ remaining messages from the block (at and after the current transaction), because the block is probably already too close to the limit.

linera-chain/src/chain.rs
afck: Because `max_failures == 0` is different from `Abort`:
* `Abort` means we never modify the block proposal; every error causes us to cancel execution and return the error.
* `AutoRetry { max_failures: 0 }` means: As soon as we hit any block limit error in an incoming message, we remove that and all remaining incoming messages from the block.

linera-chain/src/chain.rs
afck: `max_failures` is not used in `Abort` mode. Maybe it should be an `Option` but then we'd just have another infallible `unwrap`.

linera-execution/src/lib.rs
afck: It does, we're using it: For transient errors we need to avoid _rejecting_ the message and should fail the block immediately (even in `AutoRetry` mode), so the client can download e.g. the missing blobs and retry the block.

linera-chain/src/chain.rs
```diff
@@ -684,26 +684,36 @@ where
...
         local_time: Timestamp,
         round: Option<u32>,
         published_blobs: &[Blob],
         replaying_oracle_responses: Option<Vec<Vec<OracleResponse>>>,
+        exec_policy: BundleExecutionPolicy,
     ) -> Result<(BlockExecutionOutcome, ResourceTracker), ChainError> {
+        // AutoRetry is incompatible with replaying oracle responses because removing or
+        // rejecting bundles would change which transactions execute.
+        if !matches!(exec_policy, BundleExecutionPolicy::Abort) {
+            assert!(
```
deuszx, line 706: This results in a panic, right?

linera-chain/src/chain.rs
deuszx: Ok, why not `matches!(exec_policy, BundleExecutionPolicy::AutoRetry { .. })` ? Seems clearer

linera-chain/src/chain.rs
```diff
@@ -815,20 +914,48 @@ where
...
+    fn remove_remaining_bundles(
+        block: &mut ProposedBlock,
+        mut index: usize,
+        maybe_origin: Option<ChainId>,
+    ) {
+        while index < block.transactions.len() {
+            if matches!(
+                &block.transactions[index],
+                Transaction::ReceiveMessages(bundle)
+                if maybe_origin.is_none_or(|origin| bundle.origin == origin)
```
deuszx, line 927: This hurts me eyes (and brain) but alternatives are not as concise.

linera-core/src/client/mod.rs
```diff
@@ -1554,97 +1552,48 @@ impl<Env: Environment> Client<Env> {
...
-                    );
-                    incoming_bundle.action = MessageAction::Reject;
-                    continue;
-                }
+            if let Err(LocalNodeError::BlobsNotFound(blob_ids)) = &result {
+                let validators = self.validator_nodes().await?;
+                self.update_local_node_with_blobs_from(blob_ids.clone(), &validators)
+                    .await?;
+                continue; // We found the missing blob: retry.
+            }
```
deuszx, line 1583: Are you changing the behavior here? i.e. we didn't try to recover from `BlobsNotFound` before?

linera-core/src/client/mod.rs
```diff
@@ -1554,97 +1552,48 @@ impl<Env: Environment> Client<Env> {
...
-    // TODO(#2806): Measure how failing messages affect the execution times.
-    #[tracing::instrument(level = "trace", skip(self, block))]
-    async fn stage_block_execution_and_discard_failing_messages(
+    /// Attempts to execute the block locally with a specified policy for handling bundle failures.
+    /// If any attempt to read a blob fails, the blob is downloaded and execution is retried.
+    ///
+    /// Returns the modified block (bundles may be rejected/removed based on the policy)
+    /// and the execution result.
+    #[instrument(level = "trace", skip(self, block))]
+    async fn stage_block_execution_with_policy(
```
deuszx, line 1561: This method has changed compared to `testnet_conway` a lot 👀  we should put it bank in sync

linera-core/src/client/mod.rs
```diff
@@ -1554,97 +1552,48 @@ impl<Env: Environment> Client<Env> {
...
+        block: ProposedBlock,
         round: Option<u32>,
         published_blobs: Vec<Blob>,
+        policy: BundleExecutionPolicy,
     ) -> Result<(Block, ChainInfoResponse), chain_client::Error> {
         loop {
             let result = self
-                .stage_block_execution(block.clone(), round, published_blobs.clone())
+                .local_node
+                .stage_block_execution_with_policy(
```
deuszx, line 1571: This won't reuse the cached execution state when we missed a blob for message at index `n` but all up to `n-1` were fine?

linera-core/src/client/mod.rs
afck: No, we had two nested staging loops:
* The inner would download blobs and retry.
* The outer would reject/remove failed messages.

So we'd always first try to download the blobs.

But now the loops switched places, because we moved the "reject/remove" one into the worker, so we need to make sure we don't reject/remove due to missing blobs/events/etc. and pass that kind of error up to the outer loop.

linera-chain/src/chain.rs
afck: Would you prefer `maybe_origin == None || maybe_origin == Some(origin)`.
(Do you dislike `is_none_or` and `is_some_and` in general?)

linera-chain/src/chain.rs
afck: Yes, it's invalid to call this with oracle responses and `AutoRetry`. I wonder if we should just move the responses into the `Abort` variant, even though they're not really part of the "policy".

linera-core/src/client/mod.rs
afck: Has it? I guess when we backport this we'll replace it anyway.

linera-core/src/client/mod.rs
afck: That's true, that's difficult to do without restructuring the chain state view, I think. Or at least it feels dangerous to pass the cached execution state back from the worker, when the underlying chain state view gets unlocked and could change. But maybe it would be fine, if we make sure that we're still at the same height afterwards?

linera-core/src/client/mod.rs
deuszx: Right, I didn't notice this and it's missing on the diff.

linera-chain/src/chain.rs
deuszx: I think the latter – IMHO this urge to be concise is hurting readability. But leave it, I'll eventually get used to seeing it.

linera-core/src/client/mod.rs
deuszx: Maybe not in this PR? Also, we'd always try to first download all blobs that are _statically_ known to be needed, right? i.e. some explicit dependencies or application calls and it's only the case for those that are dynamically used (like a call to another app during execution)?

afck: No, I think there's still a lot of room for optimization there, too.

linera-chain/src/chain.rs
```diff
@@ -739,12 +747,103 @@ where
...
+                        (error, context, incoming_bundle, saved_chain, saved_tracker)
+                    }
+                    (Err(e), _, _) => return Err(e),
+                };
+
+            // Restore checkpoint.
+            *chain = saved_chain;
+            block_execution_tracker.restore_checkpoint(saved_tracker);
+
+            if error.is_limit_error() && i > 0 {
```
deuszx, line 801: Why do you need `i > 0` ? I understand that for a failed message in the first transaction we want to remove it from the inbox entirely (not just for this block proposal) but I think it'd be better if we had that logic (message rejection/handling) encapsulated in one place

afck: But we do! There's the general case for a failed message, in the `else` branches below:
* If a message fails, _reject_ it (and therefore remove it from the inbox… unless it's protected).

This here is the special case:
* If a message other than the first failed due to a block limit error, maybe we _don't_ need to reject it, and can just remove it from the block, so it can be retried in the next.

linera-chain/src/chain.rs
```diff
@@ -739,12 +747,103 @@ where
...
+                failure_count += 1;
+                // If we've exceeded max failures, remove all remaining message bundles.
+                let maybe_sender = if failure_count > max_failures {
```
deuszx, line 804: Shouldn't it be `if failure_count >= max_failures` ? Or even simply `failure_count == max_failures` (as we always increment by 1 only).

afck: Currently the description says:
> Maximum number of message bundles to remove from a block proposal due to block limit errors before removing all remaining bundles

We've already incremented `failure_count`, so that's the total number of block limit errors we've seen so far.

So I'd read the description as saying: If `failure_count <= max_failures`, then we still want to remove only that sender's remaining messages, rather than all of them.

linera-chain/src/data_types/mod.rs
ma2bd: I agree the difference "removed"<>"rejected" could be clarified.
Maybe "discarded" works better?

afck: Done in https://github.com/linera-io/linera-protocol/pull/5367/commits/f71173e9e37e2326ef5f4b2fb76b8310ca9b178d.

PR #5368, opened 2026-01-31
Handle different epochs in the integration tests.

MathieuDutSik:
## Motivation

Handling different epochs is important for integration test since changing the policy lead to a change of epoch.

## Proposal

Determine the `min_epoch/max_epoch` in order to create the right ChainDescription.

## Test Plan

The test `http_request.rs` adds the creation of a chain that led to the discovery of this problem

## Release Plan

This could be added to the TestNet Conway.

## Links

None.

### DIFF
--- a/examples/how-to/perform-http-requests/tests/http_requests.rs
+++ b/examples/how-to/perform-http-requests/tests/http_requests.rs
@@ -34,7 +34,7 @@ async fn service_query_performs_http_request() -> anyhow::Result<()> {
         .await;
     // Tests that we can create test chains after the change of epoch
     // due to the change of policy.
-    std::hint::black_box(validator.new_chain().await);
+    validator.new_chain().await;
 
     let QueryOutcome { response, .. } = chain
         .graphql_query(application_id, "query { performHttpRequest }")


examples/how-to/perform-http-requests/tests/http_requests.rs
```diff
@@ -32,6 +32,9 @@ async fn service_query_performs_http_request() -> anyhow::Result<()> {
                 .insert("localhost".to_owned());
         })
         .await;
+    // Tests that we can create test chains after the change of epoch
+    // due to the change of policy.
+    let _user_chain = validator.new_chain().await;
```
ma2bd: Because of `let _x =` we can't tell if it's testing anything

examples/how-to/perform-http-requests/tests/http_requests.rs
```diff
@@ -34,7 +34,7 @@ async fn service_query_performs_http_request() -> anyhow::Result<()> {
         .await;
     // Tests that we can create test chains after the change of epoch
     // due to the change of policy.
-    let _user_chain = validator.new_chain().await;
+    std::hint::black_box(validator.new_chain().await);
```
ma2bd: I don't think you need the black_box. In any case, I was just suggesting to remove `let _user_chain =`

examples/how-to/perform-http-requests/tests/http_requests.rs
MathieuDutSik: The error was occurring in the creation of the chain. This was failing before the change.

I replace it with the use of `std::hint::black_box` which should prevent it from being optimized away.

ma2bd: What is the return type?

MathieuDutSik: The return type of `validator.new_chain().await` is `ActiveChain`.

The function `std::hint_black_box` is the identity. But the point is that this prevents it from being optimized away by the compiler.

examples/how-to/perform-http-requests/tests/http_requests.rs
MathieuDutSik: Ok. Updated.

afck (approved):

PR #5369, opened 2026-02-01
[testnet] Initial client-side support for open multi-leader round chains. (#5328)

afck:
Backport of #5328.

## Motivation

Open multi-leader rounds so far are only supported on the server side and not taken into account in the client at all.

## Proposal

Allow the client's preferred owner even if it's not an owner of the chain but open multi-leader rounds are enabled.

Also, `assign_new_chain_to_key` now uses the latest chain info, not the initial chain description to determine ownership, timestamp etc.

Thanks @kikakkz for finding this and writing the initial fix.

## Test Plan

The worker test was moved to `client_tests` and uses the new logic extracted from `assign_new_chain_to_key`.

## Release Plan

- Backport to `testnet_conway`, then
    - release a new SDK.

## Links

- PR to main: #5328

### DIFF
--- a/linera-client/src/client_context.rs
+++ b/linera-client/src/client_context.rs
@@ -474,10 +474,6 @@ impl<Env: Environment> ClientContext<Env> {
         chain_id: ChainId,
         owner: AccountOwner,
     ) -> Result<(), Error> {
-        if !self.client.has_key_for(&owner).await? {
-            tracing::error!("No key pair found for owner {owner}");
-            return Err(error::Inner::ChainOwnership.into());
-        }
         self.client
             .extend_chain_mode(chain_id, ListeningMode::FullChain);
         let client = self.make_chain_client(chain_id).await?;
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -2018,6 +2018,10 @@ impl<Env: Environment> ChainClient<Env> {
         &self,
         owner: AccountOwner,
     ) -> Result<Box<ChainInfo>, ChainClientError> {
+        ensure!(
+            self.client.has_key_for(&owner).await?,
+            ChainClientError::CannotFindKeyForChain(self.chain_id)
+        );
         // Ensure we have the chain description blob.
         self.client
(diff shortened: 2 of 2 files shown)


linera-client/src/client_context.rs
```diff
@@ -477,10 +477,17 @@ impl<Env: Environment> ClientContext<Env> {
...
+        // Ensure we have the chain description blob.
+        client.get_chain_description().await?;
+
+        // Synchronize and get chain info.
+        client.synchronize_from_validators().await?;
+        let info = client.chain_info().await?;
+
+        // Validate that the owner can propose on this chain (either as owner or via
+        // open_multi_leader_rounds).
+        if !info.manager.ownership.is_multi_leader_owner(&owner) {
```
deuszx: Do we check whether wallet has the key for the `owner`?

deuszx (approved):
No hard blockers but maybe in `prepare_for_owner` we should also validate that signer has the private key for the `owner`.

linera-client/src/client_context.rs
afck: I don't think so.

Maybe the cleanest fix would be to add this in `main.rs` instead, before we call `assign_new_chain_to_key`?

```rust
ensure!(signer.contains_key(&owner), "Missing key for owner");
```

afck: Done in https://github.com/linera-io/linera-protocol/pull/5369/commits/3464e73abcbfbf2f4dc9c44261b99e2a1561a2e8; also renamed as you suggested.

deuszx: This works for CLI only, right?

afck: It's only used on the CLI, but I fixed it (and an error message) anyway in https://github.com/linera-io/linera-protocol/pull/5369/commits/7d37de284ba5a0c0c55318b536c51612ecd00d55.

afck: Actually, moved it into `prepare_for_owner`: https://github.com/linera-io/linera-protocol/pull/5369/commits/162e96f557068f655ece219e433c5b17dca761ac

PR #5373, opened 2026-02-02
`@linera/client`: runtime-configurable logging

Twey:
## Motivation

Performance profiling, while useful, causes the JavaScript memory usage to expand greatly (and grow over time!).

## Proposal

Disable profiling and logs below `INFO` by default, and add a couple of different ways of enabling them:

The `initialize` function now takes an `InitializeOptions` that can be used to set two options:

- `log` takes a log filter string à la `RUST_LOG`;
- `profiling` is a boolean that can be used to enable performance reporting via the Web Performance API.

If not set explicitly, and `window.location` is available, defaults for these can also be provided in the URL search parameters as `LINERA_LOG` and `LINERA_PROFILING` respectively.

## Test Plan

Tested locally.

## Release Plan

    - be released in a new SDK,

## Links

### DIFF
--- a/web/@linera/client/src/lib.rs
+++ b/web/@linera/client/src/lib.rs
@@ -57,11 +57,18 @@ pub struct InitializeOptions {
     pub profiling: bool,
 }
 
+static INITIALIZED: std::sync::OnceLock<()> = std::sync::OnceLock::new();
+
 #[wasm_bindgen]
 pub fn initialize(options: Option<InitializeOptions>) {
     use tracing_subscriber::{
         prelude::__tracing_subscriber_SubscriberExt as _, util::SubscriberInitExt as _, EnvFilter,
     };
+
+    if INITIALIZED.set(()).is_err() {
+        tracing::warn!("already initialized, not re-initializing");
+        return;
+    }
 
     let options = options.unwrap_or_default();
 


web/@linera/client/Cargo.toml
```diff
@@ -23,7 +23,7 @@ serde.workspace = true
 serde-wasm-bindgen.workspace = true
 tokio.workspace = true
 tokio-util.workspace = true
-tracing = { workspace = true, features = ["release_max_level_debug"] }
+tracing.workspace = true
```
deuszx, line 26: w/o the feature flag does it pass-through everything ?

web/@linera/client/src/lib.rs
```diff
@@ -44,117 +43,45 @@ pub use wallet::Wallet;
...
-    // futures on the global task queue.
-    // It does nothing here in this single-threaded context, but is
-    // hard-coded by `ChainListener`.
-    Arc<AsyncMutex<linera_client::ClientContext<Environment>>>,
-);
+#[derive(serde::Deserialize, Default, tsify::Tsify)]
+#[tsify(from_wasm_abi)]
+#[serde(default, rename_all = "camelCase")]
+pub struct Options {
+    pub log: String,
```
deuszx, line 55: Can you explain (as a comment here) what structures is the `log` expected to take? Down below you use it in `.parse_lossy(_)` and it's unclear here what it should be.

web/@linera/client/src/index.ts
```diff
@@ -1,4 +1,18 @@
...
+  if (window.location) {
+    const params = new URL(window.location.href).searchParams;
+    const defaults: wasm.Options = {};
+    defaults.profiling = params.get('LINERA_PROFILING') !== null;
+    defaults.log = params.get('LINERA_LOG') || '';
+    options = { ...defaults, ...options };
+  }
+
+  await wasm.default();
+  wasm.initialize(options);
```
deuszx, line 17: smart.

web/@linera/client/src/lib.rs
Twey: Oh this is an old version! I must have forgotten to push. One moment

web/@linera/client/Cargo.toml
Twey: No.  This is configured at runtime now (and the default is `INFO`, less verbose than the `DEBUG` we currently have).

PR #5376, opened 2026-02-03
Include source maps when packaging linera web client.

deuszx:
## Motivation

Without them it's much harder to debug or measure.

## Proposal

Inline sources which provides full debugging capability without shipping a separate `src/` folder.

## Test Plan

Tested manually. The size increase if negligible: it's ~8kB increase for ~27MB client.

## Release Plan

- These changes should be forward-ported to `main`.

## Links


Twey:
Doesn't seem too harmful.  Maybe we should distribute the source tree though :thinking: It would make the package bigger but not affect loading times.

PR #5378, opened 2026-02-03
Preferred chain owners on the Web

Twey:
## Motivation

Previously we added the ability for application queries to be given a particular owner.  However, to do autosigning right we need to be able to set the owner when setting chain owners (with `setOwner`).

## Proposal

Add a notion of chain options that includes the chain owner used for queries to the chain.  Move the `owner` option from `Application.query` into the chain options in `Client.chain`.

## Test Plan

Checked locally with the MetaMask fungible example (which changes are included here).

## Release Plan

    - be released in a new SDK,

## Links

### DIFF
--- a/examples/counter/metamask/index.html
+++ b/examples/counter/metamask/index.html
@@ -70,12 +70,16 @@
      document.getElementById('chain-id').innerText = chainId;
 
      // For autosigning: first we set up a local (in-memory) signer, and provide it to the
-     // client along with the MetaMask signer
+     // client along with the MetaMask signer.
+     // We set it as the default owner of the chain in the wallet.
      const autosigner = linera.signer.PrivateKey.createRandom();
-     const client = await new linera.Client(wallet, new linera.signer.Composite(autosigner, signer));
+     wallet.setOwner(chainId, autosigner.address());
 
-     // Connect to our user chain.
-     const chain = await client.chain(chainId);
+     const client = await new linera.Client(wallet, new linera.signer.Composite(autosigner, signer));
+     const chain = await client.chain(chainId, { owner });
+
+     // For autosigning: we then add the in-memory signer as an owner of the chain to allow it to sign transactions.
+     await chain.addOwner(autosigner.address());
 
      // Connect to the counter application on the chain.
      const counter = await chain.application(import.meta.env.LINERA_APPLICATION_ID);
@@ -87,12 +91,6 @@
          addLogEntry(newBlock);
          updateCount(newBlock.hash);
      });
-
-     // For autosigning: we then add the in-memory signer as an owner of the chain in the
-     // wallet, and set it as the default owner (so it will be used to process messages and
-     // events).
-     await chain.addOwner(autosigner.address());
-     wallet.setOwner(chainId, autosigner.address());
 
      function addLogEntry(block) {
          const entry = logs.getElementsByTagName('template')[0].content.cloneNode(true);
@@ -109,9 +107,7 @@
      updateCount();
 
      incrementButton.addEventListener('click', () => {
-         // For autosigning: when we make user-initiated mutations, we explicitly make them with
-         // the original (MetaMask) owner by providing the `owner` option to the `query` call.
-         counter.query('{ "query": "mutation { increment(value: 1) }" }', { owner });
+         counter.query('{ "query": "mutation { increment(value: 1) }" }');
      });
     </script>
   </body>
--- a/web/@linera/client/src/chain/application.rs
+++ b/web/@linera/client/src/chain/application.rs
@@ -1,7 +1,7 @@
 // Copyright (c) Zefchain Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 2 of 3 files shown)


web/@linera/client/src/chain/application.rs
```diff
@@ -19,10 +19,9 @@ pub struct Application {
 #[serde(rename_all = "camelCase")]
 #[tsify(from_wasm_abi)]
 pub struct QueryOptions {
+    // TODO(#5377): should this be in `ChainOptions`?
```
deuszx, line 22: I don't think so. In general we don't store historical execution states in Linera – this here is only for the temporary changes that the chain actor may maintain while block is proposed but not yet finalised. This is set when querying after receiving `BlockExecuted` notification (issues after block is executed locally) vs `ConfirmedBlock. This makes it really the query-specific value.

examples/counter/metamask/index.html
```diff
@@ -70,12 +70,16 @@ <h2>Chain history for <code id="chain-id" class="hex">requesting chain…</code>
      document.getElementById('chain-id').innerText = chainId;
 
      // For autosigning: first we set up a local (in-memory) signer, and provide it to the
-     // client along with the MetaMask signer
+     // client along with the MetaMask signer.
+     // We set it as the default owner of the chain in the wallet.
      const autosigner = linera.signer.PrivateKey.createRandom();
+     wallet.setOwner(chainId, autosigner.address());
```
deuszx, line 76: I find it confusing now:
- `wallet.setOwner` uses `autosigner` as owner of `chainId` within the `wallet`
- `client.chain` uses `owner`as owner of (the same) `chain` within the `chain`

We end up having different owners for the same chain in two structs.

I liked the previous approach where the owner set set explicitly in the `query`. I think it was cleaner.

Twey: It was certainly more convenient.  Unfortunately it's also wrong :)

web/@linera/client/src/chain/application.rs
Twey: Does that mean that this API is racy?  I.E. if the block finalizes in the background before getting to the `query` then that `block_hash` may not work any more.

examples/counter/metamask/index.html
Twey: We _could_ alternatively remove the owner from the `ChainClient` entirely and add it as a parameter to every `ChainClient` method.  I'm not sure that's cleaner though.

web/@linera/client/src/chain/application.rs
deuszx: if it can't find the block hash it will use the latest state (finalized one). Precisley b/c the block might get finalized - in which case the `BlockExecuted` results are already in the finalized state. 

The assumption is that the client/caller is not interseted in the state at this _exact_ block hash but the latest/post-execution results. Those are either in the cached state or finalized one so it doesn't matter if we use the finalized state (if the block is confirmed in the meantime).

examples/counter/metamask/index.html
deuszx: I think having two (different) owners in wallet and chain client is almost wrong as well.

web/@linera/client/src/chain/application.rs
Twey: So even the results are non-deterministic?  That's interesting… but I still don't see what makes this application query–specific. What stops us from asking for other state from the given block hash, like (native) account balances?

examples/counter/metamask/index.html
Twey: I'm open to other APIs, but probably in a follow-up.  This is just reflecting the `ChainClient` API, so any serious changes should probably start there (and then we can bubble it up to here).

Twey: Basically the structure at the moment is:
- chains themselves can have many owners, but only one owner at a time can submit transactions/make queries
- the wallet includes a default owner for every chain that is used if no owner is explicitly specified 
- the chain client also stores an owner to be used for operations through that chain client, thus allowing overriding the default owner for operations

web/@linera/client/src/chain/application.rs
deuszx: b/c once intermediate state is finalized it's removed so in a case of a long-lived chain instance it doesn't make sense

Twey: Right, I get that.  `block_hash` doesn't really mean ‘query at this block hash’ but ‘query at any block at or above this block hash’.  But it would still make sense to be able to do all sorts of queries at that cached block, not just application queries.

deuszx: I don't agree – the indermediate state is really expected to be VERY short lived (<1s) and it's meant to speed up the UI state updates.

Twey: What if the UI is displaying the chain balance?

deuszx: I don't understand the question.

Twey:
I misspoke: this was tested with the _counter_ example, not the fungible example.

PR #5380, opened 2026-02-03
Fix gRPC subscription retry count to reset on reconnection

ndr-ds:
## Motivation

Long-lived gRPC subscription streams would silently terminate after accumulating `max_retries` errors over time, even
when each individual reconnection was successful. This caused workers to stop receiving notifications and go silent.

## Proposal

Change `retry_count` from a local variable to a shared `Arc<AtomicU32>` that can be reset from within the `unfold`
closure when `client.subscribe()` succeeds.

Previously, `retry_count` only reset when a successful notification was received. Now it also resets when a
reconnection succeeds, so `max_retries` applies to consecutive failed reconnection attempts rather than total errors
over the stream's lifetime.

## Test Plan

- CI
- Manual testing with PM benchmark that was experiencing the issue

### DIFF
--- a/linera-rpc/src/grpc/client.rs
+++ b/linera-rpc/src/grpc/client.rs
@@ -1,7 +1,16 @@
 // Copyright (c) Zefchain Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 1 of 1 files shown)


deuszx (approved):
LGTM

PR #5385, opened 2026-02-04
Lock the wallet on the Web

Twey:
## Motivation

With the introduction of persistent storage on the Web, we create the possibility that different clients (e.g. running in different tabs) may share storage.  This is illegal wrt the Linera data model, which assumes exclusive access to the storage for performance reasons.

## Proposal

Put different wallets (as defined by their default chains) in different storage namespaces.  Where the same wallet is used in different contexts, use the [Web Locks API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Locks_API) to fail immediately on client initialization.

## Test Plan

Tested locally with the `counter` example.

## Release Plan

    - be released in a new SDK,

## Links

### DIFF
--- a/.github/workflows/lint-check-all-features.yml
+++ b/.github/workflows/lint-check-all-features.yml
@@ -19,7 +19,7 @@ env:
   RUST_BACKTRACE: short
   # We allow redundant explicit links because `cargo rdme` doesn't know how to resolve implicit intra-crate links.
   RUSTDOCFLAGS: -A rustdoc::redundant_explicit_links -D warnings
-  RUSTFLAGS: -D warnings
+  RUSTFLAGS: -D warnings --cfg=web_sys_unstable_apis
   RUSTUP_MAX_RETRIES: 10
   RUST_LOG: warn
 


web/@linera/client/src/lock.rs
```diff
@@ -0,0 +1,71 @@
+// Copyright (c) Zefchain Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
```
deuszx, line 2: Can you ask Claude to add a module doc to this?

web/@linera/client/src/lock.rs
```diff
@@ -0,0 +1,122 @@
...
+                );
+
+            callback.forget();
+        }))
+        .await
+        .unwrap_throw();
+
+        serde_wasm_bindgen::from_value::<Option<Self>>(value)
+            .unwrap_throw()
+            .ok_or_else(|| Error::Contended { name: name.into() })
```
deuszx, line 114: what i can't wrap my head around is where do you actully construact an instance of `Lock { lock, release }`?

deuszx: nvm, I see it 🤦

web/@linera/client/src/lock.rs
```diff
@@ -0,0 +1,122 @@
...
+        options.set_if_available(true);
+
+        let value = JsFuture::from(js_sys::Promise::new(&mut |resolve, _reject| {
+            let callback =
+                wasm_bindgen::closure::Closure::once(move |lock: Option<web_sys::Lock>| {
+                    js_sys::Promise::new(&mut |release, _reject| {
+                        let value = if let Some(lock) = &lock {
+                            Some(Self {
+                                lock: lock.clone(),
+                                release,
```
deuszx, line 79: I hate callback hell

deuszx:
A test would be a nice addition :)

Twey:
In the interests of getting it right out — I'll do this as a follow-up tomorrow.

Twey:
In the interests of getting it right out — I'll do this as a follow-up tomorrow.

web/@linera/client/src/lock.rs
Twey: Thankfully — I've done some work here so it doesn't spread to the rest of the codebase :)

PR #5387, opened 2026-02-05
Backport preferred chain owners PR. Add metamask fungible example.

deuszx:
#5378 backport

## Motivation

Previously we added the ability for application queries to be given a particular owner.  However, to do autosigning right we need to be able to set the owner when setting chain owners (with `setOwner`).

## Proposal

Add a notion of chain options that includes the chain owner used for queries to the chain.  Move the `owner` option from `Application.query` into the chain options in `Client.chain`.

## Test Plan

A new demo was added - MetaMask fungible example - to test the behavior. Indeed, the autosigning of icoming messages works.

## Release Plan

## Links

- #5378 


examples/native-fungible/metamask/vite.config.ts
```diff
@@ -0,0 +1,43 @@
+import { defineConfig, loadEnv } from 'vite';
```
Twey, line 1: I typically just use a symlink for these. Thus far we've managed to keep the build for them all roughly identical.

Issue #5386, opened 2026-02-04
Add CLI option to cap the time spent adding and staging incoming bundles to a block proposal


ma2bd:
Perhaps `--max-staging-time-message-bundles` ?

Intuitively we could disable `--max-pending-message-bundles` by default, and set this new option to (say) `500ms`.

afck:
We'd probably pass the parameter into `stage_block_execution_with_policy` (maybe make it part of the policy itself), and measure the time it takes to execute each message bundle. I'd suggest we even include the failed attempts for then rejected/discarded message bundles, otherwise block production could still be delayed much longer than the time limit. Once the limit is exceeded after executing a bundle, we'd discard all remaining bundles from that block—but still execute any remaining operations! (We currently never discard operations in the staging loop, and I don't think we should.)

As a next step we could consider even stopping an individual message execution once the time limit is reached, but I'm not sure right now whether it's safe to cancel that task.

ndr-ds:
Closing this, but lmk if we should leave it open for any follow ups you guys had in mind

PR #5393, opened 2026-02-05
Add --staging-bundles-time-budget-ms for time-budgeted bundle staging

ndr-ds:
## Motivation

Currently, block staging limits incoming message bundles by _count_ (`max_pending_message_bundles`), which doesn't
account for the variable cost of executing different bundles. Expensive bundles can cause high block execution latency
even when the count is low, while cheap bundles underutilize capacity.

This adds a time-based budget that gives operators direct control over block execution latency: the client includes
as many bundles as fit within a wall-clock time budget, producing more balanced blocks automatically.

Related issue: https://github.com/linera-io/linera-protocol/issues/5386

## Proposal

Add `--staging-bundles-time-budget-ms` (`Option<Duration>`) CLI option. When set, instead of including up to
`max_pending_message_bundles` bundles per block, the client includes as many bundles as can be executed within the
time budget.

**Semantics:**

- Only `ReceiveMessages` (bundles) are time-budgeted; `ExecuteOperation` transactions always execute
- After each bundle execution, cumulative bundle time is checked against the budget
- The bundle that pushes cumulative time over the budget IS included (already executed)
- Remaining bundles are selectively discarded (operations still execute)
- When `None`, behavior is unchanged (uses `max_pending_message_bundles` count limit)

**Refactoring:** `BundleExecutionPolicy` was split into `BundleFailurePolicy` (Abort/AutoRetry) and a
`BundleExecutionPolicy` struct that combines the failure policy with the optional time budget.

## Test Plan

- CI
- Manual testing: run a worker with `--staging-bundles-time-budget-ms 500` and verify the slow blocks issue is
resolved

### DIFF
--- a/linera-chain/src/unit_tests/chain_tests.rs
+++ b/linera-chain/src/unit_tests/chain_tests.rs
@@ -57,7 +57,7 @@ impl ChainStateView<MemoryContext<TestExecutionRuntimeContext>> {
     }
 
     /// Test helper that calls `execute_block` with default test parameters:
-    /// `round = None`, `replayed_oracle_responses = None`, `policy = Abort`.
+    /// `round = None`, `replayed_oracle_responses = None`, `on_failure = Abort`.
     #[cfg(with_testing)]
     pub async fn execute_test_block_simple(
         &mut self,
@@ -71,7 +71,7 @@ impl ChainStateView<MemoryContext<TestExecutionRuntimeContext>> {
             None,
             published_blobs,
             None,
-            BundleExecutionPolicy::Abort,
+            BundleExecutionPolicy::committed(),
         )
         .await
     }
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -37,8 +37,8 @@ use linera_base::{
 use linera_base::{data_types::Bytecode, vm::VmRuntime};
 use linera_chain::{
     data_types::{
-        BlockProposal, BundleExecutionPolicy, ChainAndHeight, IncomingBundle, LiteVote,
-        ProposedBlock, Transaction,
+        BlockProposal, BundleExecutionPolicy, BundleFailurePolicy, ChainAndHeight, IncomingBundle,
+        LiteVote, ProposedBlock, Transaction,
     },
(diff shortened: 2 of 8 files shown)


linera-chain/src/unit_tests/chain_tests.rs
```diff
@@ -71,7 +71,10 @@ impl ChainStateView<MemoryContext<TestExecutionRuntimeContext>> {
             None,
             published_blobs,
             None,
-            BundleExecutionPolicy::Abort,
+            BundleExecutionPolicy {
+                on_failure: BundleFailurePolicy::Abort,
+                time_budget: None,
+            },
```
afck: We could call this `BundleExecutionPolicy::committed()`. (For committed blocks this is the only policy.)

linera-core/src/client/mod.rs
```diff
@@ -1588,6 +1588,9 @@ pub struct ChainClientOptions {
     pub max_block_limit_errors: u32,
     /// Maximum number of new stream events processed at a time in a block.
     pub max_new_events_per_block: usize,
+    /// Time budget for staging message bundles. When set, limits bundle execution by time
+    /// rather than by count.
```
afck, line 1592: ~Or rather: In addition to by count? We still only take at most `max_pending_message_bundles`, do we?~

(Edit: I see, we don't. But maybe we should?)

afck (approved):
No blockers, but a few questions.

linera-core/src/client/mod.rs
ndr-ds: I don't think we should 🤔 the whole point of this was not having to come up with a number for this. For fast enough messages, most numbers you pick might be too low.

afck: Then I wonder if we should disallow specifying both parameters, or at least print a warning that the other is ignored.

ma2bd: @ndr-ds I'm not completely fond of overriding `max_pending_message_bundles`. Can we bring it back? (we can also increase the default value to a large number like `250`)

Issue #5333, opened 2026-01-26
review code quality in tracing_opentelemetry.rs


ma2bd:
1) We use `eprintln!` instead of (say) `tracing::warn!`.
2) The implementation of `init_with_opentelemetry` prints a warning even if `otlp_endpoint` is not set.

ma2bd:
```
LINERA_OTLP_EXPORTER_ENDPOINT not set and no endpoint provided. Falling back to standard tracing without OpenTelemetry span export. Baggage propagation is still enabled.
```

PR #5394, opened 2026-02-05
Remove eprintln usage from tracing_opentelemetry

ndr-ds:
## Motivation

Addresses #5333: improve code quality in `tracing_opentelemetry.rs` by replacing `eprintln!` with proper
`tracing::warn!` logging.

## Proposal

Replace `eprintln!` with `tracing::warn!` in both `init_with_opentelemetry` implementations:

1. **With `opentelemetry` feature**: When no OTLP endpoint is configured, initialize tracing first, then log a warning
via `tracing::warn!` before falling back to standard tracing.

2. **Without `opentelemetry` feature**: Similarly, initialize tracing first, then warn that the feature is not
enabled.

This ensures fallback messages go through the proper logging infrastructure rather than directly to stderr, making
them filterable and consistent with the rest of the codebase.

## Test Plan

- CI

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -230,8 +230,8 @@ jobs:
         until nc -z 127.0.0.1 1235; do sleep 1; done
     - name: Run the benchmark test
       run: |
-        cargo build --locked -p linera-service --bin linera-benchmark --features storage-service
-        cargo test --locked -p linera-service --features storage-service benchmark
+        cargo build --locked -p linera-service --bin linera-benchmark --features storage-service,opentelemetry
+        cargo test --locked -p linera-service --features storage-service,opentelemetry benchmark
 
   ethereum-tests:
     needs: changed-files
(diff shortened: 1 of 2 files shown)


ma2bd:
Why is CI failing @ndr-ds ?

ndr-ds:
I think it should be good now

.github/workflows/rust.yml
```diff
@@ -320,7 +320,7 @@ jobs:
         until nc -z 127.0.0.1 1235; do sleep 1; done
     - name: Run the storage-service tests
       run: |
-        cargo test --features storage-service -- storage_service --nocapture
+        cargo test --features storage-service,opentelemetry -- storage_service --nocapture
```
ma2bd, line 323: Are we using the data somehow?

ma2bd: Is this going to trigger a lot of recompilation?

ma2bd (approved):

.github/workflows/rust.yml
ndr-ds: This type of context propagation is an OpenTelemetry feature, that's why we need the feature, so we compile those crates in. We don't really use the data in the sense that we don't use the fact that it's synthetic traffic, but I thought it was a good idea to disallow running the benchmarks without this feature, so that it is always known that the traffic generated by it is synthetic.
As far as recompilation, it's just the opentelemetry sdk stuff, so I don't think it's too bad

Issue #5323, opened 2026-01-26
Linera web client doesn't work on Safari 26.2


ma2bd:
Repro steps:
* attach safari inspector to iphone
* browse [demos.linera.net/fungible](https://demos.linera.net/testnet/fungible/index.html)

<img width="1130" height="540" alt="Image" src="https://github.com/user-attachments/assets/7d2b9520-96c3-4f64-8035-62f15b668005" />

Twey:
This is a [known Safari bug](https://github.com/emscripten-core/emscripten/issues/25905).  There is theoretically [a fix in WebKit](https://github.com/WebKit/WebKit/commit/597e00851c7198c8da1efa0f8f756295d9d098ad) but I'm guessing it hasn't made it downstream into Safari yet.

We can't bypass it by disabling bulk memory features because [we can't link Wasm programs for shared memory without them](https://github.com/llvm/llvm-project/blob/e7f3226e4f70f502cbd60ca5e999a6680850a50e/lld/wasm/Writer.cpp#L636-L640), and we currently rely on threading for correctness.

deuszx:
Should be now resolved via https://github.com/linera-io/linera-protocol/pull/5395

PR #5395, opened 2026-02-06
Safari memory fix

deuszx:
## Motivation

Safari 26.2 doesn't work w/ our demos https://github.com/linera-io/linera-protocol/issues/5323

## Proposal

Detect if the browser is Safari 26.2 we preallocate memory manually while still single-threded. This prevents further `memory.grow` calls. For other browseres we use old approach.

## Test Plan

Manual

## Release Plan

- These changes should be backported to `main`
- Included in the next SDK release

## Links

- https://github.com/linera-io/linera-protocol/issues/5323
- https://github.com/alexcrichton/dlmalloc-rs/issues/61

### DIFF
--- a/web/@linera/client/src/index.ts
+++ b/web/@linera/client/src/index.ts
@@ -25,9 +25,11 @@ export async function initialize(options?: wasm.InitializeOptions) {
   const exports = await wasm.default();
 
   // Safari 26.2 crashes with shared WebAssembly memory during
-  // multi-threaded operation (WebKit #303387). Pre-allocating a large block
-  // while still single-threaded prevents the crash — Rust's wasm32 allocator
-  // (dlmalloc) acquires all memory via memory.grow, so filling its pool here
+  // multi-threaded operation (WebKit #303387). Rust's wasm32 allocator
+  // (dlmalloc) uses via memory.grow to decide how much memory is left
+  // so starting with INITIAL and MAXIMUM values that are too close 
+  // to the actual memory limit causes it to grow memory and crash.
+  // Pre-allocating a large block while still single-threaded prevents the crash — 
   // avoids memory.grow calls once worker threads are running.
   if (isBrokenSafari()) {
     const PREALLOC_BYTES = 768 * 1024 * 1024;


web/@linera/client/src/index.ts
```diff
@@ -4,6 +4,15 @@ export type { Signer } from './signer/Signer.js';
...
 import * as wasm from './wasm/index.js';
 
+function isSafari(): boolean {
+  try {
+    const ua = navigator.userAgent;
+    return /Safari\//.test(ua) && !/Chrome\/|Chromium\//.test(ua);
+  } catch {
+    return false;
+  }
+}
```
Twey: Let's use [Bowser](https://www.npmjs.com/package/bowser): UA string parsing is a minefield today, and with Bowser we can also easily limit this to Safari 26.2.* where the bug happens (hopefully it's fixed by 26.3!).

web/@linera/client/src/index.ts
```diff
@@ -13,6 +22,25 @@ export async function initialize(options?: wasm.InitializeOptions) {
...
   }
 
-  await wasm.default();
+  const exports = await wasm.default();
+
+  // Safari 26-26.2 crashes with shared WebAssembly memory during
+  // multi-threaded operation (WebKit #303387). Pre-allocating a large block
+  // while still single-threaded prevents the crash — Rust's wasm32 allocator
+  // (dlmalloc) acquires all memory via memory.grow, so filling its pool here
+  // avoids memory.grow calls once worker threads are running.
```
Twey: This comment is wrong but we know it :)

deuszx: which parts? b/c at least 2/3 of it I've confirmed empirically :)

Twey: It's not that it acquires all its memory via `memory.grow`, it's that it uses `memory.grow` to even decide how much memory it has.

Also the bug is new in 26.2 (right?) not from 26.

Twey: I've filed https://github.com/alexcrichton/dlmalloc-rs/issues/61 which we could link to.

web/@linera/client/src/index.ts
```diff
@@ -13,6 +22,25 @@ export async function initialize(options?: wasm.InitializeOptions) {
...
+  if (isSafari()) {
+    const PREALLOC_BYTES = 768 * 1024 * 1024;
+    try {
+      const ex = exports as any;
```
Twey: Nit: this shouldn't be necessary :thinking: The type of `init()` includes these exports correctly.

web/@linera/client/src/index.ts
```diff
@@ -13,6 +22,25 @@ export async function initialize(options?: wasm.InitializeOptions) {
...
-  await wasm.default();
+  const exports = await wasm.default();
+
+  // Safari 26-26.2 crashes with shared WebAssembly memory during
+  // multi-threaded operation (WebKit #303387). Pre-allocating a large block
+  // while still single-threaded prevents the crash — Rust's wasm32 allocator
+  // (dlmalloc) acquires all memory via memory.grow, so filling its pool here
+  // avoids memory.grow calls once worker threads are running.
+  if (isSafari()) {
+    const PREALLOC_BYTES = 768 * 1024 * 1024;
```
ndr-ds, line 35: Why 768MB?

web/@linera/client/src/index.ts
deuszx: Done.

web/@linera/client/src/index.ts
deuszx: Update. I hope it's good now.

web/@linera/client/src/index.ts
deuszx: Done.

web/@linera/client/src/index.ts
deuszx: safe middleground I hope.

Twey:
LGTM.  This is a small fix at the end of a really painful deep-dive — great work!

PR #5398, opened 2026-02-06
Add more instrumenting around block execution

ndr-ds:
## Motivation

We're still missing instrumentation around block execution, which was making it difficult to pinpoint where the message execution latency was coming from.

## Proposal

This PR adds `#[instrument]` tracing spans throughout the block execution path:

- **Block tracker** (`linera-chain/src/block_tracker.rs`): Instrumented `execute_transaction`,
`execute_message_in_block`, and `process_txn_outcome` with relevant context fields (chain_id, block_height,
transaction_index, message origin, etc.)

- **Execution state actor** (`linera-execution/src/execution_state_actor.rs`): Added spans to `load_contract`,
`handle_request`, `process_subscriptions`, `contract_and_dependencies`, `run_user_action_with_runtime`,
`execute_operation`, and `execute_message` with fields indicating operation/message types

- **Runtime** (`linera-execution/src/runtime.rs`): Instrumented `load_contract_instance`, `run_action`, `finalize`,
and `execute` methods

- **WASM backends** (both wasmer and wasmtime): Added `#[instrument(skip_all)]` to all contract entrypoint methods
(`instantiate`, `execute_operation`, `execute_message`, `process_streams`, `finalize`)

Additionally, adjusted the `state_hash_computation_latency` histogram bucket upper bound from 500ms to 2000ms to
capture longer-running hash computations (it can easily go over 500ms if we have zero hash disabled, for example).

## Test Plan

- CI
- Deployed a network with this change, and saw the new tracing data on Tempo


ndr-ds:
So `#[instrument]` doesn't create logs, it creates spans. These spans will be sent to tempo, when configured.
Spans only generate logs when you turn on span event logging, via the `RUST_LOG_SPAN_EVENTS` env variable, which we don't really do anywhere currently, AFAIK. The context from these spans might appear in logs when you explicitly call `info!`, etc, but that's it. They won't generate extra logs.
I guess if you have log spam on the web, we need to adjust the log level there? cc @Twey

linera-chain/src/block_tracker.rs
```diff
@@ -421,3 +438,11 @@ pub(crate) type FinalizeExecutionResult = (
...
     ResourceTracker,
 );
+
+/// Returns a human-readable name for the transaction type.
+fn transaction_type_name(transaction: &Transaction) -> &'static str {
+    match transaction {
+        Transaction::ReceiveMessages(_) => "ReceiveMessages",
+        Transaction::ExecuteOperation(_) => "ExecuteOperation",
+    }
+}
```
afck: (We could use `strum::AsRefStr` for this. ~Especially if we want to do the same with other enums, too.~ I see, we do!)

ndr-ds: Ah, didn't know about `strum`! Nice :)

PR #5404, opened 2026-02-06
Fix benchmark close chain race condition (frontport of #5397)

ndr-ds:
## Motivation

The benchmark test is flaky because `close_benchmark_chain` can fail
when there are still in-flight requests from
benchmark tasks. When the shutdown signal is received, pending requests
may still be committed as blocks, causing a
conflict when we try to close the chain. The current code panics on this
conflict instead of handling it gracefully.

Example failure:
https://github.com/linera-io/linera-protocol/actions/runs/21718806404/job/62642487443?pr=5391

## Proposal

Replace the `.expect()` call in `close_benchmark_chain` with a retry
loop that properly handles both `ClientOutcome`
variants:

- **`Conflict`**: Log at info level and retry immediately (another block
was committed, we just need to try again)
- **`WaitForTimeout`**: Log at info level and sleep until the timeout
timestamp expires before retrying (we're not the
round leader)

This follows the same patterns used elsewhere in the codebase (e.g.,
`wait_for_next_round` in
`linera-client/src/util.rs`).

## Test Plan

- CI

### DIFF
--- a/linera-client/src/benchmark.rs
+++ b/linera-client/src/benchmark.rs
@@ -11,12 +11,13 @@ use std::{
 };
 
 use linera_base::{
-    data_types::Amount,
+    data_types::{Amount, Timestamp},
     identifiers::{Account, AccountOwner, ApplicationId, ChainId},
     time::Instant,
 };
 use linera_core::{
     client::chain_client::{self, ChainClient},
+    data_types::ClientOutcome,
     Environment,
 };
 use linera_execution::{system::SystemOperation, Operation};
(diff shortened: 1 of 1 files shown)


ndr-ds:
* **#5404** <a href="https://app.graphite.com/github/pr/linera-io/linera-protocol/5404?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/linera-io/linera-protocol/5404?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #5406, opened 2026-02-09
Add free application IDs to resource control policy

afck:
## Motivation

We want to allow using some apps without spending fees on messages and events.

## Proposal

Add a configurable list of application IDs to ResourceControlPolicy whose message- and event-related fees are waived. When executing messages or processing streams for these apps, all fees (fuel, storage I/O, blobs, events, HTTP requests, oracle queries, message size) are skipped while resource limits and tracking are still enforced. Other operation fees remain unchanged. Outgoing message fees from free apps are also skipped in block tracking.

## Test Plan

Tests were added.

## Release Plan

- Maybe find a way to backport this to `testnet_conway`.
    - then release a new SDK and validator hotfix.

## Links

### DIFF
--- a/linera-chain/src/block_tracker.rs
+++ b/linera-chain/src/block_tracker.rs
@@ -324,6 +324,9 @@ impl<'resources, 'blobs> BlockExecutionTracker<'resources, 'blobs> {
 
         // Account for blobs published by this transaction directly.
         for blob in &txn_outcome.blobs {
+            if txn_outcome.free_blob_ids.contains(&blob.id()) {
+                continue; // Blob publishing fees are waived for free apps.
+            }
             resource_controller
                 .track_blob_published(blob)
                 .with_execution_context(context)?;
--- a/linera-service/src/cli/main.rs
+++ b/linera-service/src/cli/main.rs
@@ -657,13 +657,11 @@ impl Runnable for Job {
                                             .unwrap_or(existing_policy.http_request_allow_list),
                                         free_application_ids: free_application_ids
                                             .map(|ids| {
-                                                ids.into_iter()
-                                                    .map(|s| {
-                                                        serde_json::from_value(
-                                                            serde_json::Value::String(s),
-                                                        )
-                                                    })
-                                                    .collect::<Result<BTreeSet<_>, _>>()
+                                                ids.into_iter().map(|s| s.parse()).collect::<Result<
+                                                    BTreeSet<_>,
+                                                    _,
+                                                >>(
+                                                )
                                             })
                                             .transpose()
                                             .expect("Invalid application ID")
@@ -2083,7 +2081,7 @@ async fn run(options: &Options) -> Result<i32, Error> {
                     .as_ref()
                     .map(|ids| {
                         ids.iter()
-                            .map(|s| serde_json::from_value(serde_json::Value::String(s.clone())))
+                            .map(|s| s.parse())
                             .collect::<Result<BTreeSet<_>, _>>()
                     })
                     .transpose()
--- a/linera-execution/src/execution_state_actor.rs
+++ b/linera-execution/src/execution_state_actor.rs
@@ -705,6 +705,9 @@ where
             }
 
             AddCreatedBlob { blob, callback } => {
+                if self.resource_controller.is_free {
+                    self.txn_tracker.mark_blob_free(blob.id());
+                }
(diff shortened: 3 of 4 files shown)


linera-chain/src/block_tracker.rs
```diff
@@ -286,6 +286,11 @@ impl<'resources, 'blobs> BlockExecutionTracker<'resources, 'blobs> {
             if message_out.kind == MessageKind::Bouncing {
                 continue; // Bouncing messages are free.
             }
+            if let Message::User { application_id, .. } = &message_out.message {
+                if resource_controller.policy().is_free_app(application_id) {
```
deuszx, line 307: what about the `track_blob_published` below? it's also calling `self.update_balance(self.policy.blob_published_price(size)?)?;`

linera-execution/src/policy.rs
```diff
@@ -94,6 +94,8 @@ pub struct ResourceControlPolicy {
     pub http_request_timeout_ms: u64,
     /// The list of hosts that contracts and services can send HTTP requests to.
     pub http_request_allow_list: BTreeSet<String>,
+    /// The list of application IDs for which all message- and event-related fees are waived.
+    pub free_application_ids: BTreeSet<ApplicationId>,
```
deuszx, line 98: So this is not backwards compatible? Or for testnet you imagine putting it in a different place?

MathieuDutSik, line 98: Maybe use `HashSet` since I do not think ordering matters.

afck: Exactly: The HTTP allow list, as with the other flags.

linera-service/src/cli/main.rs
```diff
@@ -2064,6 +2079,16 @@ async fn run(options: &Options) -> Result<i32, Error> {
...
+                free_application_ids: free_application_ids
+                    .as_ref()
+                    .map(|ids| {
+                        ids.iter()
+                            .map(|s| serde_json::from_value(serde_json::Value::String(s.clone())))
+                            .collect::<Result<BTreeSet<_>, _>>()
+                    })
+                    .transpose()
+                    .expect("Invalid application ID")
+                    .unwrap_or(existing_policy.free_application_ids),
```
MathieuDutSik, line 2089: Here we use `as_ref` and then clone later on. That seems inefficient.
Also a clear code duplication with the preceding entry.

linera-execution/tests/fee_consumption.rs
```diff
@@ -423,3 +424,150 @@ impl FeeSpend {
...
+        },
+    ));
+    application.expect_call(ExpectedCall::default_finalize());
+
+    let context = linera_execution::OperationContext {
+        chain_id,
+        height: BlockHeight(0),
+        round: Some(0),
+        authenticated_owner: None,
+        timestamp: Default::default(),
```
MathieuDutSik: Maybe write `Timestamp::default()` instead.

linera-execution/src/resources.rs
```diff
@@ -26,6 +26,8 @@ pub struct ResourceController<Account = Amount, Tracker = ResourceTracker> {
     pub tracker: Tracker,
     /// The account paying for the resource usage.
     pub account: Account,
+    /// When true, balance deductions are skipped (fees waived for free apps).
+    is_free: bool,
```
MathieuDutSik: If it were up to me, I would put `pub is_free` and `pub policy`.
To be clear, using setter/getter is fine and using `pub ...` is also fine. I just do not like the mix.
Not a blocker obviously.

linera-execution/src/policy.rs
afck: But this needs to serialize in a deterministic way, so I think we need something with a defined order.

linera-chain/src/block_tracker.rs
afck: Good catch! Done in https://github.com/linera-io/linera-protocol/pull/5406/commits/661073cf77363346818654bea5e5eb27700aedf0.

linera-service/src/cli/main.rs
afck: We don't actually need to clone: Fixed in https://github.com/linera-io/linera-protocol/pull/5406/commits/661073cf77363346818654bea5e5eb27700aedf0.

(Note that we need `as_ref()`, because the whole struct above is only a reference. `Option::map` consumes the option.)

PR #5408, opened 2026-02-09
Fix client owner wallet overwrite

deuszx:
## Motivation

Currently there's an issue with `Wallet::update_from_client` method overwrites wallet's (preferred) owner. This manifests itself in apps with autosigning set up for incoming bundles: when an autosigning key is overwritten by the chain's (session) key during command execution leading to a behavior where proposals are being signed using chain's key (not autosigner).

## Proposal

Don't overwrite wallet owner from client

## Test Plan

Test was added (fails w/o the fix).

## Release Plan

- These changes should be ported to `main`
- These changes should be included in the next SDK release.

## Links

### DIFF
--- a/linera-client/src/unit_tests/chain_listener.rs
+++ b/linera-client/src/unit_tests/chain_listener.rs
@@ -68,13 +68,13 @@ impl chain_listener::ClientContext for ClientContext {
         client: &ChainClient<environment::Test>,
     ) -> Result<(), Error> {
         let info = client.chain_info().await?;
-        let client_owner = client.preferred_owner();
+        let existing_owner = self.wallet().get(info.chain_id).and_then(|c| c.owner);
         let pending_proposal = client.pending_proposal().clone();
         self.wallet().insert(
             info.chain_id,
             wallet::Chain {
                 pending_proposal,
-                owner: client_owner,
+                owner: existing_owner,
                 ..info.as_ref().into()
             },
         );
--- a/linera-client/src/unit_tests/chain_listener.rs
+++ b/linera-client/src/unit_tests/chain_listener.rs
@@ -506,3 +506,179 @@ async fn test_chain_listener_listen_command_adds_chains_to_wallet() -> anyhow::R
 
     Ok(())
 }
+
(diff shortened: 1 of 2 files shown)


Twey (approved):
LGTM as long as we don't have any existing tests that rely on this behaviour.

linera-client/src/unit_tests/chain_listener.rs
```diff
@@ -68,13 +68,13 @@ impl chain_listener::ClientContext for ClientContext {
         client: &ChainClient<environment::Test>,
     ) -> Result<(), Error> {
         let info = client.chain_info().await?;
-        let client_owner = client.preferred_owner();
+        let existing_owner = self.wallet().get(info.chain_id).and_then(|c| c.owner);
```
afck, line 71: (I think we want to avoid single-letter identifiers.)

afck (approved):
Not sure if that's _always_ the desired behavior: I imagine in some cases we don't want to add the chain to the wallet at all; and maybe there are even other cases where we update the preferred owner in the chain client and _do_ want to set it in the wallet, too?

Anyway, for now I'm fine with this. As discussed, we should probably make all of that more explicit in `linera-core`'s API.

deuszx:
>  I imagine in some cases we don't want to add the chain to the wallet at all; 

That is already happening, even before this PR, isn't it?

>  and maybe there are even other cases where we update the preferred owner in the chain client and do want to set it in the wallet, too?

With the current code it might be quite convoluted. At least on the frontend side we treat chain clients as ephemeral/session-like constructs. I think it makes sense given the multi-owner nature of the chain.

linera-client/src/unit_tests/chain_listener.rs
ma2bd: re: single letters: we could decide that short lambdas are ok

deuszx: I agree - short ones should be ok. Other languages (like Scala) have a nice construct for this: `and_then(_.owner)` :)

PR #5413, opened 2026-02-10
Replace `usize` by `u32` in the view definitions.

### LINKED ISSUE #4970: linera-views: Replace `usize` in serializable data
I'm seeing `usize` in BucketQueueView.

I believe this is not platform-safe. Unless we are sure that Wasm is always usize=u32 and for non-wasm usize=u64 ??

MathieuDutSik:
## Motivation

When constructing views we are using usize which has a specific serialization. This pose problem because
usize is specific to an architecture. Sure, most laptop are 64-bit but WebAssembly is mostly 32-bits.

Fixes #4970 

## Proposal

Switch the definition of view of `BucketQueueView`, `QueueView`, and `LogView` to `u64` from `usize`.
All the other view do not have `usize` in their definition.

## Test Plan

CI.

## Release Plan

A testnet release to testnet_conway is possible since it can safely be assumed that all the validators
are all 64-bits and so usize=u64 in that case.

## Links

None.

### DIFF
--- a/linera-views-derive/src/lib.rs
+++ b/linera-views-derive/src/lib.rs
@@ -108,7 +108,7 @@ fn generate_view_code(input: &ItemStruct, root: bool) -> Result<TokenStream2, Er
         let idx_u32 = u32::try_from(idx).expect("number of fields exceeds u32::MAX");
         let derive_key_logic = quote! {
             let __linera_reserved_index: u32 = #idx_u32;
-            let __linera_reserved_base_key = context.base_key().derive_tag_key(linera_views::views::MIN_VIEW_TAG, __linera_reserved_index)?;
+            let __linera_reserved_base_key = context.base_key().derive_tag_key(linera_views::views::MIN_VIEW_TAG, &__linera_reserved_index)?;
         };
 
         pre_load_keys_quotes.push(quote! {
--- a/linera-views-derive/src/lib.rs
+++ b/linera-views-derive/src/lib.rs
@@ -89,7 +89,6 @@ fn generate_view_code(input: &ItemStruct, root: bool) -> Result<TokenStream2, Er
     let mut num_init_keys_quotes = Vec::new();
     let mut pre_load_keys_quotes = Vec::new();
     let mut post_load_keys_quotes = Vec::new();
-    let num_fields = input.fields.len();
     for (idx, e) in input.fields.iter().enumerate() {
         let name = e.ident.clone().unwrap();
         let delete_view_ident = format_ident!("deleted{}", idx);
@@ -106,19 +105,10 @@ fn generate_view_code(input: &ItemStruct, root: bool) -> Result<TokenStream2, Er
         });
         num_init_keys_quotes.push(quote! { #g :: NUM_INIT_KEYS });
 
-        let derive_key_logic = if num_fields < 256 {
-            let idx_u8 = idx as u8;
-            quote! {
-                let __linera_reserved_index = #idx_u8;
-                let __linera_reserved_base_key = context.base_key().derive_tag_key(linera_views::views::MIN_VIEW_TAG, &__linera_reserved_index)?;
-            }
-        } else {
-            assert!(num_fields < 65536);
-            let idx_u16 = idx as u16;
-            quote! {
-                let __linera_reserved_index = #idx_u16;
-                let __linera_reserved_base_key = context.base_key().derive_tag_key(linera_views::views::MIN_VIEW_TAG, &__linera_reserved_index)?;
-            }
+        let idx_u32 = u32::try_from(idx).expect("number of fields exceeds u32::MAX");
+        let derive_key_logic = quote! {
+            let __linera_reserved_index: u32 = #idx_u32;
+            let __linera_reserved_base_key = context.base_key().derive_tag_key(linera_views::views::MIN_VIEW_TAG, __linera_reserved_index)?;
         };
 
         pre_load_keys_quotes.push(quote! {
@@ -133,18 +123,10 @@ fn generate_view_code(input: &ItemStruct, root: bool) -> Result<TokenStream2, Er
         });
     }
 
-    // derive_key_logic above adds one byte to the key as a tag, and then either one or two more
-    // bytes for field indices, depending on how many fields there are. Thus, we need to trim 2
-    // bytes if there are less than 256 child fields (then the field index fits within one byte),
-    // or 3 bytes if there are more.
-    let trim_key_logic = if num_fields < 256 {
-        quote! {
-            let __bytes_to_trim = 2;
-        }
-    } else {
-        quote! {
-            let __bytes_to_trim = 3;
-        }
+    // derive_key_logic above adds one byte to the key as a tag, plus four bytes for the
+    // `u32`-encoded field index, so we trim 5 bytes total.
+    let trim_key_logic = quote! {
+        let __bytes_to_trim = 5;
     };
 
     let first_name_quote = name_quotes
--- a/linera-views/src/context.rs
+++ b/linera-views/src/context.rs
@@ -49,18 +49,11 @@ impl BaseKey {
         key
     }
 
-    /// Obtains the `Vec<u8>` key from the key by serialization and using the base key.
-    pub fn derive_key(&self, index: u32) -> Result<Vec<u8>, bcs::Error> {
-        let mut key = self.bytes.clone();
-        bcs::serialize_into(&mut key, &index)?;
-        Ok(key)
-    }
-
     /// Obtains the `Vec<u8>` key from the key by serialization and using the `base_key`.
-    pub fn derive_tag_key(&self, tag: u8, index: u32) -> Result<Vec<u8>, bcs::Error> {
+    pub fn derive_tag_key<I: Serialize>(&self, tag: u8, index: &I) -> Result<Vec<u8>, bcs::Error> {
         assert!(tag >= MIN_VIEW_TAG, "tag should be at least MIN_VIEW_TAG");
         let mut key = self.base_tag(tag);
-        bcs::serialize_into(&mut key, &index)?;
+        bcs::serialize_into(&mut key, index)?;
         Ok(key)
     }
 
--- a/linera-views/src/context.rs
+++ b/linera-views/src/context.rs
@@ -50,21 +50,17 @@ impl BaseKey {
     }
 
     /// Obtains the `Vec<u8>` key from the key by serialization and using the base key.
-    pub fn derive_key<I: Serialize>(&self, index: &I) -> Result<Vec<u8>, bcs::Error> {
+    pub fn derive_key(&self, index: u32) -> Result<Vec<u8>, bcs::Error> {
         let mut key = self.bytes.clone();
-        bcs::serialize_into(&mut key, index)?;
-        assert!(
-            key.len() > self.bytes.len(),
-            "Empty indices are not allowed"
-        );
+        bcs::serialize_into(&mut key, &index)?;
         Ok(key)
     }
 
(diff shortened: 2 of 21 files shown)


ma2bd:
@Twey Is this testnet-backportable to testnet with the web storage?

Twey:
I don't believe so, no.

linera-views/src/views/log_view.rs
```diff
@@ -58,7 +58,7 @@ pub struct LogView<C, T> {
     /// Whether to clear storage before applying updates.
     delete_storage_first: bool,
     /// The number of entries persisted in storage.
-    stored_count: usize,
+    stored_count: u64,
```
ma2bd: Should this in fact be a `u32`? So that it always fits in an `usize`.

ma2bd (commented):
I wonder if `u64` is the only choice (see comment)

linera-views/src/views/log_view.rs
ma2bd: Or alternatively the `count()` methods could return a `u64`.

ndr-ds:
Wouldn't u32 be the better choice here? It always fits in usize on both 64-bit and 32-bit/WASM, so all the as usize casts become lossless. 4 billion entries per view is more than enough.

With u64, every u64 → usize cast (there are ~30 in this diff) silently truncates on 32-bit platforms, which is the exact portability problem this PR is trying to fix. With u32, that entire class of issue disappears.

ma2bd:
> Wouldn't u32 be the better choice here? It always fits in usize on both 64-bit and 32-bit/WASM, so all the as usize casts become lossless. 4 billion entries per view is more than enough.
> 
> With u64, every u64 → usize cast (there are ~30 in this diff) silently truncates on 32-bit platforms, which is the exact portability problem this PR is trying to fix. With u32, that entire class of issue disappears.

Yes exactly. E.g. the length of Java vectors is limited to `MAX::u32` for portability reasons.

ndr-ds (changes requested):
Back to you for that then @MathieuDutSik

MathieuDutSik:
> Back to you for that then @MathieuDutSik

Ok, switched to `u32`.

linera-views/src/views/bucket_queue_view.rs
```diff
@@ -322,7 +325,7 @@ where
             self.stored_front_position = 0;
             self.cursor = None;
         } else if let Some(cursor) = self.cursor {
-            for _ in 0..cursor.offset {
+            for _ in 0..(cursor.offset as usize) {
```
afck: No need to convert here?

linera-views/src/views/bucket_queue_view.rs
```diff
@@ -291,10 +291,13 @@ where
                 // This shouldn't happen if stored_count() > 0
                 0
             };
-            let mut start = 0;
-            while start < self.new_back_values.len() {
-                let end = std::cmp::min(start + N, self.new_back_values.len());
-                let value_chunk: Vec<_> = self.new_back_values.range(start..end).collect();
+            let mut start = 0_u32;
+            while start < self.new_back_values.len() as u32 {
+                let end = std::cmp::min(start + N, self.new_back_values.len() as u32);
```
afck: Do we need to guard somewhere against `new_back_values` or `stored_buckets` growing larger than `u32::MAX`? It would be better to return an error in that case rather than wrapping around here. (Maybe not a likely problem in practice.)

linera-views/src/views/bucket_queue_view.rs
```diff
@@ -406,15 +409,16 @@ impl<C: Context, T, const N: usize> BucketQueueView<C, T, N> {
...
         } else {
             let Some(cursor) = self.cursor else {
                 return 0;
             };
-            let mut stored_count = 0;
-            for offset in cursor.offset..self.stored_buckets.len() {
+            let mut stored_count = 0u32;
+            let offset_start = cursor.offset as usize;
+            for offset in offset_start..self.stored_buckets.len() {
                 stored_count += self.stored_buckets[offset].len();
```
afck: (Maybe also a place to check for `u32` overflows.)

linera-views/src/views/bucket_queue_view.rs
```diff
@@ -635,24 +641,25 @@ impl<C: Context, T: DeserializeOwned + Clone, const N: usize> BucketQueueView<C,
...
+            let offset_start = cursor.offset as usize;
+            for offset in offset_start..self.stored_buckets.len() {
                 let bucket = &self.stored_buckets[offset];
                 let size = bucket.len() - position;
                 if !bucket.is_loaded() {
                     let key = self.get_bucket_key(bucket.index)?;
                     keys.push(key);
                 };
-                if size >= count_remain {
+                if size >= count_remain as u32 {
```
afck: Just to be safe…? (Also below.)

```suggestion
                if u32::try_from(count_remain).is_none_or(|cr| size >= cr) {
```

ndr-ds (commented):
Title and description need to be updated now.  
Also, I don't fully understand why we need to cast `usize` to `u32` in so many places? We're potentially narrowing to `u32`, when I think we should actually widen the other side to `usize` when necessary, because that's always safe, and only narrow to `u32` (with `try_from`) when writing back to the views.

afck:
> that's always safe

(Not on 16-bit or 8-bit architectures!! :space_invader: :space_invader: :space_invader:)

linera-views/src/views/bucket_queue_view.rs
```diff
@@ -247,21 +248,21 @@ where
             stored_front_position = 0;
         } else if let Some(cursor) = self.cursor {
             // Delete buckets that are before the cursor
-            for i in 0..cursor.offset {
+            for i in 0..cursor.offset as usize {
                 let bucket = &self.stored_buckets[i];
                 let index = bucket.index;
                 let key = self.get_bucket_key(index)?;
```
afck, line 254: Maybe `get_bucket_key` should also take a `u32`? Then we don't need the `as usize` above.

linera-views/src/views/bucket_queue_view.rs
```diff
@@ -247,21 +248,21 @@ where
...
                 batch.delete_key(key);
             }
             stored_front_position = cursor.position;
             // Build descriptions for remaining buckets
-            let first_index = self.stored_buckets[cursor.offset].index;
+            let first_index = self.stored_buckets[cursor.offset as usize].index;
```
afck, line 259: I was only _half_ joking above: I'd still find it cleaner to either use `usize::try_from` or ensure somehow that `size_of::<usize>() >= size_of::<u32>()`, either in a test, or (better) statically (e.g. with `const _: () = assert!(…)`) or using `target_pointer_width`.

linera-views/src/views/bucket_queue_view.rs
```diff
@@ -291,17 +292,17 @@ where
                 // This shouldn't happen if stored_count() > 0
                 0
             };
-            let mut start = 0;
+            let mut start = 0_usize;
             while start < self.new_back_values.len() {
-                let end = std::cmp::min(start + N, self.new_back_values.len());
+                let end = std::cmp::min(start + N as usize, self.new_back_values.len());
                 let value_chunk: Vec<_> = self.new_back_values.range(start..end).collect();
```
afck, line 298: Isn't this re-implementing `slice::chunks`?

linera-views/src/views/bucket_queue_view.rs
```diff
@@ -291,17 +292,17 @@ where
...
                 let key = self.get_bucket_key(index)?;
                 batch.put_key_value(key, &value_chunk)?;
                 descriptions.push(BucketDescription {
                     index,
-                    length: end - start,
+                    length: (end - start) as u32,
```
afck, line 303: That's just `value_chunk.len()`, is it?

linera-views/src/views/bucket_queue_view.rs
```diff
@@ -413,12 +414,11 @@ impl<C: Context, T, const N: usize> BucketQueueView<C, T, N> {
...
-            let mut stored_count = 0;
-            for offset in cursor.offset..self.stored_buckets.len() {
-                stored_count += self.stored_buckets[offset].len();
+            let mut stored_count = 0_usize;
+            for offset in (cursor.offset as usize)..self.stored_buckets.len() {
+                stored_count += self.stored_buckets[offset].len() as usize;
             }
-            stored_count -= cursor.position;
-            stored_count
+            stored_count - cursor.position as usize
```
afck, line 421:
```suggestion
            let stored_count = self.stored_buckets[cursor.offset as usize..]
                .iter()
                .map(|bucket| bucket.len() as usize)
                .sum()
                .unwrap_or_default()
                - cursor.position as usize;
```

linera-views/src/views/log_view.rs
afck, line 126: Should `derive_tag_key` take a `u32` instead of `usize`?

linera-views/src/views/queue_view.rs
```diff
@@ -405,7 +413,9 @@ where
                     .cloned(),
             );
         } else {
-            let start = self.stored_indices.end + new_back_len - count;
+            let stored_consumed =
+                u32::try_from(count - new_back_len).map_err(|_| ArithmeticError::Overflow)?;
```
afck, line 417: Or `Underflow`?

linera-views/src/views/log_view.rs
MathieuDutSik: Very good point indeed.

linera-views-derive/src/lib.rs
```diff
@@ -106,19 +105,10 @@ fn generate_view_code(input: &ItemStruct, root: bool) -> Result<TokenStream2, Er
...
-                let __linera_reserved_base_key = context.base_key().derive_tag_key(linera_views::views::MIN_VIEW_TAG, &__linera_reserved_index)?;
-            }
-        } else {
-            assert!(num_fields < 65536);
-            let idx_u16 = idx as u16;
-            quote! {
-                let __linera_reserved_index = #idx_u16;
-                let __linera_reserved_base_key = context.base_key().derive_tag_key(linera_views::views::MIN_VIEW_TAG, &__linera_reserved_index)?;
-            }
+        let idx_u32 = u32::try_from(idx).expect("number of fields exceeds u32::MAX");
```
afck, line 108: A bit cleaner would be using `for (e, idx) in input.fields.iter().zip(0u32..)` above; no conversion needed.

linera-views/src/context.rs
```diff
@@ -50,21 +50,17 @@ impl BaseKey {
...
-            key.len() > self.bytes.len(),
-            "Empty indices are not allowed"
-        );
+        bcs::serialize_into(&mut key, &index)?;
         Ok(key)
     }
 
     /// Obtains the `Vec<u8>` key from the key by serialization and using the `base_key`.
-    pub fn derive_tag_key<I: Serialize>(&self, tag: u8, index: &I) -> Result<Vec<u8>, bcs::Error> {
+    pub fn derive_tag_key(&self, tag: u8, index: u32) -> Result<Vec<u8>, bcs::Error> {
```
afck: I see! So this index `n` can mean "the `n`-th field of a derived view" as well as "the `n`-th bucket"? 😩 
I guess in the former case it _is_ nice to have that optimization: most views have fewer than 256 fields, and we only need one byte for it.
I just think that for queue views, we should use `u32` instead of `usize` here.

So maybe I'd leave these functions generic and bring back the `I: Serialize`, and just use `u32` instead of `usize` at the call site in the queue views?

MathieuDutSik: Ok, done.

afck:
Closing in favor of https://github.com/linera-io/linera-protocol/pull/6277.

PR #5415, opened 2026-02-10
Log application logs via tracing crate

deuszx:
## Motivation

Before this PR, application logging on the web were log via ` web_sys::console::*` but that meant they weren't adhering to `tracing`'s filtering rules. Which we can now use in the constructor of web client since https://github.com/linera-io/linera-protocol/pull/5373.

## Proposal

Use `tracing` crate instead. Add a special `target` to the log lines – `user_application_log` so that it can be filtered out (or turned off) like `RUST_LOG=user_application_log=off`.

## Test Plan

Verified manually.

## Release Plan

- These changes should be included in the next SDK release.

## Links

### DIFF
--- a/linera-execution/src/execution_state_actor.rs
+++ b/linera-execution/src/execution_state_actor.rs
@@ -715,18 +715,20 @@ where
             }
 
             #[cfg(web)]
-            Log { message, level } => {
-                // Output directly to browser console with clean formatting
-                let formatted: js_sys::JsString = format!("[CONTRACT {level}] {message}").into();
-                match level {
-                    tracing::log::Level::Trace | tracing::log::Level::Debug => {
-                        web_sys::console::debug_1(&formatted)
-                    }
-                    tracing::log::Level::Info => web_sys::console::log_1(&formatted),
-                    tracing::log::Level::Warn => web_sys::console::warn_1(&formatted),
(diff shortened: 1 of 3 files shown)


linera-execution/src/execution_state_actor.rs
```diff
@@ -715,18 +715,20 @@ where
...
+                    tracing::info!(target: "user_application_log", message = %message);
+                }
+                tracing::log::Level::Warn => {
+                    tracing::warn!(target: "user_application_log", message = %message);
+                }
+                tracing::log::Level::Error => {
+                    tracing::error!(target: "user_application_log", message = %message);
+                }
+            },
         }
```
Twey, line 732: Since we already have a real `tracing::Level`, you can use [`event`](https://docs.rs/tracing/latest/tracing/macro.event.html) to pass the level through rather than matching on each level.

deuszx: This is not so easy: contracts use level from `tracing::log::Level` while `event!` macro expects `tracing::Level`. I'd have to change not only the type of the enum variant but also runtime API.

Twey: I did not realize those were different!  Okay, let's leave it for now.

PR #5419, opened 2026-02-10
`@linera/client`: give search params precedence over application-provided logging spec

Twey:
## Motivation

We'd like to be able to override the log filters for debug purposes even if the application has a preferred default.

## Proposal

Give the search params precedence over the `log` option.

## Test Plan

Manual.

## Release Plan

    - be released in a new SDK,

## Links

### DIFF
--- a/web/@linera/client/src/index.ts
+++ b/web/@linera/client/src/index.ts
@@ -15,11 +15,14 @@ function isBrokenSafari(): boolean {
 
 export async function initialize(options?: wasm.InitializeOptions) {
   if (window.location) {
+    // Allow overriding the application's log filters using the `LINERA_LOG` and
+    // `LINERA_PROFILING` search params, for debugging.
     const params = new URL(window.location.href).searchParams;
-    const defaults: wasm.InitializeOptions = {};
-    defaults.profiling = params.get('LINERA_PROFILING') !== null;
-    defaults.log = params.get('LINERA_LOG') || '';
-    options = { ...defaults, ...options };
+    const log = params.get('LINERA_LOG');
+
+    if (!options) options = {};
+    if (params.get('LINERA_PROFILING')) options.profiling = true;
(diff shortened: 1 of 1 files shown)


web/@linera/client/src/index.ts
```diff
@@ -15,21 +15,24 @@ function isBrokenSafari(): boolean {
...
+    // `LINERA_PROFILING` search params, for debugging.
     const params = new URL(window.location.href).searchParams;
-    const defaults: wasm.InitializeOptions = {};
-    defaults.profiling = params.get('LINERA_PROFILING') !== null;
-    defaults.log = params.get('LINERA_LOG') || '';
-    options = { ...defaults, ...options };
+    const log = params.get('LINERA_LOG');
+
+    if (!options) options = {};
+    if (params.get('LINERA_PROFILING')) options.profiling = true;
```
deuszx, line 24: How do you enable `LINERA_PROFILING` in the address bar? `LINERA_PROFILING=true` ? If so, shouldn't we check if it's disabled?

Twey: Any (or no) value of `LINERA_PROFILING` will do the trick.

deuszx: so `LINERA_PROFILING=false` will enable profiling? Quite surprising. Why not respect the boolean?

Twey: Conventional encoding of booleans in GET params (e.g. as generated by HTML forms) is as presence/absence.  The spelling of the false value as `false` is language-specific.

deuszx: sure but here it represents a log filter which has the form of key-value. I tried finding any information about URL relation w/ GET params and there is none - i.e. it's not in any standard.

deuszx (approved):

Issue #5379, opened 2026-02-03
Gracefully handle failures in process_streams


ma2bd:
Right now, clients will retry and fail to build a block again.

Clients should probably skip the `UpdateStreams` operation that fails and carry on with a block.

Open questions:
* Should the client pause the stream on its own (off-chain) for some time?
* Should it be authorized to unsubscribe for good (on-chain)?

afck:
## Changes we can make in the client

If it's a block limit error, it might help to split the `UpdateStreams` into smaller pieces, and only update one stream at a time, or even one event at a time.

If even that fails, the client could mark that particular stream as stuck and (locally) unsubscribe from it.

## Changing execution semantics

That's not ideal either, though: If a faulty application A subscribes to events from an application B and just always fails on B's events, any chain with both applications will be unable to make A process the streams.

So maybe we should (at least for the next testnet) move away from the model again where we have a single `UpdateStreams` transaction in which all subscribing applications are called, and instead manage streams separately per subscriber app? Maybe it would be best to change `UpdateStreams` to `UpdateStream`, which updates _one_ application about _one_ stream at a time.

afck:
Another option is to not fail the block if `process_streams` fails: Just don't call that application (and, on `main`, make a note in that block that it failed).

afck:
A very simple short-term workaround would be to add a limit, similar to the pending message bundles, for how many events we update apps about in any given block.

afck:
Reopening: still needs a better fix on `main`.

bart-linera:
#5804 provides a more elegant solution, but it's still possible that a single `UpdateStream` operation exceeds the limits by itself (if a stream has lots of events and the subscriber is very much out of date). In such cases, it might make sense to first try to reduce the number of events processed in the operation (by decreasing the `next_index` parameter), and only drop the operation entirely if that doesn't help.

PR #5420, opened 2026-02-10
Add `max_new_events_per_block` setting to limit stream events per block 

afck:
This is for the testnet only. On `main` we might use a different solution. (To be discussed.)

## Motivation

The `UpdateStreams` can be very long and expensive and even exceed block limits if the number of events is large.

## Proposal

Add a configurable limit in the client, just like we have for incoming message bundles.

## Test Plan

A test was added that verifies that with a limit of 3 and 8 new events, `process_inbox` will now produce three blocks but still eventually process all of them.

## Release Plan

- Release a new SDK.

## Links

- Related: #5379


ndr-ds:
We might wanna do event staging time budget at some point as well

PR #5423, opened 2026-02-11
Lock the wallet on the Web (#5385)

Twey:
## Motivation

With the introduction of persistent storage on the Web, we create the possibility that different clients (e.g. running in different tabs) may share storage. This is illegal wrt the Linera data model, which assumes exclusive access to the storage for performance reasons.

## Proposal

Put different wallets (as defined by their default chains) in different storage namespaces. Where the same wallet is used in different contexts, use the [Web Locks
API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Locks_API) to fail immediately on client initialization.

## Test Plan

Tested locally with the `counter` example.

## Release Plan

    - be released in a new SDK,

## Links

### DIFF
--- a/.github/workflows/web.yml
+++ b/.github/workflows/web.yml
@@ -21,7 +21,6 @@ env:
   RUST_BACKTRACE: full
   # We allow redundant explicit links because `cargo rdme` doesn't know how to resolve implicit intra-crate links.
   RUSTDOCFLAGS: -A rustdoc::redundant_explicit_links -D warnings
-  RUSTFLAGS: -D warnings
   RUSTUP_MAX_RETRIES: 10
   RUST_LOG: linera=debug
   RUST_LOG_FORMAT: plain
@@ -70,6 +69,8 @@ jobs:
       run: |
         ln -sf toolchains/nightly/rust-toolchain.toml
     - uses: actions-rust-lang/setup-rust-toolchain@v1
+      with:
+        rustflags: ""
     - uses: arduino/setup-protoc@v3
       with:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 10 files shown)


web/@linera/client/src/wallet.rs
```diff
@@ -37,4 +39,28 @@ impl Wallet {
                 "chain {chain_id} doesn't exist in wallet"
             )))
     }
+
+    #[must_use]
```
ma2bd, line 43: why `must_use` ?

.github/workflows/web.yml
```diff
@@ -70,6 +69,8 @@ jobs:
       run: |
         ln -sf toolchains/nightly/rust-toolchain.toml
     - uses: actions-rust-lang/setup-rust-toolchain@v1
+      with:
+        rustflags: ""
```
ma2bd, line 73: What does this do?

ma2bd (approved):
LGTM

web/@linera/client/src/wallet.rs
deuszx: I guess so that we make sure the wallet names are derived using the consistent/expected approach

Twey: Clippy wanted it (`must_use_candidate`) and I didn't see a reason not to. Extended discussion [here](https://users.rust-lang.org/t/liberal-application-of-must-use/85173/2).

.github/workflows/web.yml
Twey: It stops `setup-rust-toolchain` from overriding our `RUSTFLAGS`, a thing it apparently does by default.

Twey: It stops `setup-rust-toolchain` from overriding our `RUSTFLAGS`, a thing it apparently does by default.

PR #5425, opened 2026-02-11
Add missing app ID; submit task outcomes in the correct order.

afck:
## Motivation

Some apps may expect task outcomes to be submitted in the order in which they were requested.

## Proposal

Order them correctly in the task processor.

Also, add the application ID when rescheduling a query.

## Test Plan

TBD: This is rather difficult to test.

## Release Plan

- Release a new SDK.

## Links

### DIFF
--- a/linera-service/tests/local_net_tests.rs
+++ b/linera-service/tests/local_net_tests.rs
@@ -1360,7 +1360,7 @@ async fn test_node_service_with_task_processor() -> Result<()> {
 #[cfg(feature = "storage-service")]
 #[test_log::test(tokio::test)]
 async fn test_task_processor_outcome_ordering() -> Result<()> {
-    use std::{io::Write, os::unix::fs::PermissionsExt, time::Duration};
+    use std::{io::Write, os::unix::fs::PermissionsExt};
 
     use linera_base::{abi::ContractAbi, identifiers::ApplicationId};
 
@@ -1417,6 +1417,9 @@ async fn test_task_processor_outcome_ordering() -> Result<()> {
 
     node_service.ensure_is_running()?;
 
+    // Subscribe to notifications for the chain.
+    let mut notifications = Box::pin(node_service.notifications(chain).await?);
+
     let app = node_service.make_application(&chain, &app_id.with_abi::<TaskProcessorAbi>())?;
 
     // Submit both tasks in a single block: slow first, then fast.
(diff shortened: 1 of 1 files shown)


linera-service/src/task_processor.rs
```diff
@@ -198,22 +200,49 @@ impl<Env: linera_core::Environment> TaskProcessor<Env> {
...
+                                task.operator,
+                                task.input,
+                                operators,
+                            ))
+                        })
+                        .collect();
+                    let results = future::join_all(handles).await;
+                    // Wait for the previous batch to finish sending outcomes first.
+                    if let Some(previous) = previous {
+                        let _ = previous.await;
```
ma2bd: nit: remove the let to clarify the code

linera-service/src/task_processor.rs
```diff
@@ -82,6 +83,7 @@ impl<Env: linera_core::Environment> TaskProcessor<Env> {
             notifications,
             deadlines: BinaryHeap::new(),
             operators,
+            last_task_handles: BTreeMap::new(),
```
MathieuDutSik, line 87: Put it in the same order as the TaskProcessor.

linera-service/src/task_processor.rs
MathieuDutSik, line 94: That clone can be avoided I think.

linera-service/src/task_processor.rs
```diff
@@ -198,22 +200,49 @@ impl<Env: linera_core::Environment> TaskProcessor<Env> {
...
+                        .map(|task| {
+                            let operators = operators.clone();
+                            tokio::spawn(Self::execute_task(
+                                application_id,
+                                task.operator,
+                                task.input,
+                                operators,
+                            ))
+                        })
+                        .collect();
```
MathieuDutSik, line 221: Use  `collect::<Vec<_>>`.

linera-service/src/task_processor.rs
MathieuDutSik, line 252: The function is no longer executing a task. Instead it is determining the outcome.
So, I would rename it as `determine_outcome`.

MathieuDutSik, line 177: Is it possible to change that to  `process_action` that process a single application_id?
The big loop that is completely factorizable is not very elegant.

afck: No, because `process_actions` takes `&mut self`, so we can't pass in a reference to parts of `self`.

afck: It's still also executing the task. The only thing that changed is that it _returns_ the outcome now, rather than putting it in the channel.

afck: That would introduce a lot of loops at the call sites. And it wasn't introduced in this PR, so maybe let's not change it here?

MathieuDutSik: Correct, thank you.

linera-service/tests/local_net_tests.rs
```diff
@@ -1355,6 +1355,115 @@ async fn test_node_service_with_task_processor() -> Result<()> {
...
+    .await?;
+
+    // Wait until both tasks have been processed.
+    for _ in 0..30 {
+        tokio::time::sleep(Duration::from_secs(1)).await;
+        let count: u64 = app.query_json("taskCount").await?;
+        if count >= 2 {
+            break;
+        }
+    }
```
bart-linera: Should we wait for a notification about a new block instead, like in other tests?

linera-service/tests/local_net_tests.rs
```diff
@@ -1355,6 +1355,115 @@ async fn test_node_service_with_task_processor() -> Result<()> {
...
+
+    let task_count: u64 = app.query_json("taskCount").await?;
+    assert!(
+        task_count >= 2,
+        "Expected at least 2 tasks, got {task_count}"
+    );
```
bart-linera: Maybe assert equality here? We don't ever expect more than 2 tasks here, do we?

linera-service/tests/local_net_tests.rs
```diff
@@ -1355,6 +1355,115 @@ async fn test_node_service_with_task_processor() -> Result<()> {
...
+
+    // Verify the results are in request order (slow first, fast second),
+    // not completion order (which would be fast first).
+    let results: Vec<String> = app.query_json("results").await?;
+    assert!(
+        results.len() >= 2,
+        "Expected at least 2 results, got {results:?}"
+    );
```
bart-linera: Same as above.

PR #5431, opened 2026-02-12
Fix duplicate task generation

ndr-ds:
## Motivation

The task processor can call `process_actions` multiple times while a previous batch's tasks are still in-flight.
When there's a backlog (`schedule.start` far in the past), the deadline timer fires
immediately because the next event is also in the past. The `last_requested_callbacks` guard passes, and
`next_actions` reads stale on-chain state (schedule.start hasn't advanced yet), generating overlapping timestamps.
When both batches are eventually submitted as blocks, batch 2's stale timestamps don't
match the now-advanced `schedule.start`, causing:

```
contract.rs:115: assert_eq!(timed_data.timestamp, next_timestamp, "unexpected timestamp in posted data")
```

## Proposal

Introduce an `in_flight_apps: BTreeSet<ApplicationId>` guard in the task processor. When a batch of tasks is spawned
for an application, the app is marked as in-flight and subsequent `process_actions` calls skip it. A new
`TaskMessage` enum replaces the raw `(ApplicationId, TaskOutcome)` channel, adding a `BatchComplete` variant that the
spawned task sends after all outcomes. On `BatchComplete`, the guard is cleared and `process_actions` is
re-triggered for that app so it reads freshly-updated on-chain state.

## Test Plan

- CI
- I had a stuck worker with a big backlog. After rebuilding pm-app with this fix and restarting the worker with the new package version, I see no more panics, and it seems that the worker is actually making progress through the backlog:

### DIFF
--- a/linera-service/src/task_processor.rs
+++ b/linera-service/src/task_processor.rs
@@ -22,7 +22,7 @@ use linera_base::{
 };
 use linera_core::{client::ChainClient, node::NotificationStream, worker::Reason};
 use serde_json::json;
-use tokio::{io::AsyncWriteExt, process::Command, select, sync::mpsc, task::JoinHandle};
+use tokio::{io::AsyncWriteExt, process::Command, select, sync::mpsc};
 use tokio_util::sync::CancellationToken;
 use tracing::{debug, error, info};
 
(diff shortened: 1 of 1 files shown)


linera-service/src/task_processor.rs
```diff
@@ -243,8 +268,10 @@ impl<Env: linera_core::Environment> TaskProcessor<Env> {
                             }
                         }
                     }
+                    // Signal that this batch is done so the main loop can process
+                    // the next batch for this application.
+                    let _ = sender.send(TaskMessage::BatchComplete { application_id });
```
ma2bd, line 273: can we remove `let _ =` ?

ndr-ds: Yes, my bad https://github.com/linera-io/linera-protocol/pull/5436

linera-service/src/task_processor.rs
```diff
@@ -176,6 +197,10 @@ impl<Env: linera_core::Environment> TaskProcessor<Env> {
 
     async fn process_actions(&mut self, application_ids: Vec<ApplicationId>) {
         for application_id in application_ids {
+            if self.in_flight_apps.contains(&application_id) {
+                debug!("Skipping {application_id}: tasks already in flight");
+                continue;
+            }
```
ma2bd, line 203: This is wrong. When we call `process_actions` after `process_events` , we have already removed the events from the queue. If we skip anything, the queue could become empty, and we will have to wait for something else to happen (like a new block).

ma2bd: Oh but we "re-process immediately to pick up new tasks" when the app stopped being in-flight. So everything's fine.

PR #5435, opened 2026-02-12
Adjust service chain ownership when reassigning services between workers

bart-linera:
## Motivation

If a service is reassign from worker A to worker B by the controller application, worker A should cease to be the owner of the service's chain, and worker B should become an owner.

## Proposal

Upon reassignment, when telling worker A to stop running the service, worker A will add worker B as an owner and notify the main controller chain.

The main controller will then tell worker B to take over, telling it at which height on the service chain it became an owner. Worker B then waits until it synchronizes the service chain until that height, and when it's ready, it starts listening to the chain in `FullChain` mode and removes worker A as an owner.

## Test Plan

`test_controller` was extended to reassign a service from one worker to another, and then check if the second worker properly handles a task request.

## Release Plan

    - be released in a new SDK,

## Links

- #5270 

### DIFF
--- a/examples/controller/README.md
+++ b/examples/controller/README.md
@@ -211,3 +211,7 @@ Messages are sent between chains to coordinate state:
 | `Stop { service_id, new_owners }` | Controller -> Worker | Stop a service, initiating ownership handoff via the service chain |
 | `FollowChain { chain_id }` | Controller -> Worker | Follow a chain |
 | `ForgetChain { chain_id }` | Controller -> Worker | Stop following a chain |
+| `AddOwners { service_id, new_owners }` | Worker -> Service Chain | Add new owners to a service chain during handoff |
+| `RemoveOwners { owners_to_remove }` | Worker -> Service Chain | Remove old owners from a service chain after handoff |
+| `OwnersAdded { service_id, added_at }` | Service Chain -> Worker | Confirm new owners were added at a given block height |
+| `HandoffStarted { service_id, target_block_height }` | Worker -> Controller | Notify controller that handoff phase 1 is complete |


examples/controller/src/contract.rs
```diff
@@ -147,26 +197,126 @@ impl Contract for ControllerContract {
...
                     .await;
             }
+            Message::HandoffStarted {
+                service_id,
+                target_block_height,
+            } => {
+                assert_eq!(
+                    self.runtime.chain_id(),
+                    self.runtime.application_creator_chain_id(),
+                    "HandoffStarted can only be executed on the chain that created the PM engine"
```
ma2bd: I think we mean "that created the controller" (several places)

examples/controller/src/contract.rs
```diff
@@ -47,16 +48,37 @@ pub enum Message {
...
         service_id: ManagedServiceId,
+        new_owners: HashSet<AccountOwner>,
     },
     FollowChain {
         chain_id: ChainId,
     },
     ForgetChain {
         chain_id: ChainId,
     },
+    // -- Messages sent from the worker to a service chain --
```
ma2bd: by "worker", you mean: "the worker's (control/default) chain"

examples/controller/src/contract.rs
```diff
@@ -47,16 +48,37 @@ pub enum Message {
...
+    RemoveOwners {
+        owners_to_remove: HashSet<AccountOwner>,
+    },
+    // -- Messages sent from a service chain to the worker --
+    OwnersAdded {
+        service_id: ManagedServiceId,
+        added_at: BlockHeight,
+    },
+    // -- Messages sent from the workers to the controller chain --
+    HandoffStarted {
```
ma2bd, line 78: Would it be possible to have a (mermaid?) diagram with all the messages? I understand we have 5 chains in 3 locations :)

ma2bd: Actually, does it work if chains have several workers? (I believe the data structures allow it)

bart-linera: Generally I tried to avoid assuming a single worker per service, since as you're saying, data structures allow multiple workers, but I also wasn't trying very hard to make it work in that case because of a comment in the controller that we should only use one worker per service for now. I actually believe this should work even with multiple workers, but I haven't tested it.

ma2bd: I'm asking because in the case of a single worker per chain, it seems like we could remove `HandoffStarted` and have worker A directly send `Start` to worker B. But of course, it's interesting to be future-proof.

ma2bd:
Please next time [link the PR](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue) to the github issue in the proper way

bart-linera:
Ah, right, sorry!

examples/controller/src/contract.rs
bart-linera: I guess even with multiple workers, the controller could include the list of new workers to be notified after the owners are added in the `Stop` message :thinking:

PR #5437, opened 2026-02-13
Change CanonicalState sharing model.

deuszx:
## Motivation

The exporter's `CanonicalState` is cloned via `clone_unchecked()`, giving the exporter a `LogView` with a frozen `stored_count`. After flush persists entries to DB, the exporter's `LogView` still can't read them – `LogView::get()` routes `index >= stored_count` to `new_values` (empty in the clone), never hitting the DB. This means the in-memory buffer can never be cleared and grows without bound.

## Proposal

Share a single `CanonicalState` behind `Arc<tokio::sync::RwLock>` instead of cloning it. Both the block processor and exporter use the same `LogView`, so `stored_count` stays consistent. After flush +`post_save()`, the buffer is safely drained — subsequent reads fall through to the `LogView` which has the correct `stored_count`.

### `CanonicalState`
  - Removed `flushed_count` field (flush now drains the entire buffer)
  - Replaced `Arc<papaya::HashMap<usize, CanonicalBlock>>` with `Vec<CanonicalBlock>` (`RwLock` provides concurrency; entries are sequential)
  - Removed `clone()` (no more `clone_unchecked()`) and `next_index()`
  - `flush()` now uses `buffer.drain(..)` instead of iterating `flushed_count..count`

### `SharedStorage`
  - `shared_canonical_state` field is now `Arc<tokio::sync::RwLock<CanonicalState<C>>>`
  - `clone()` is infallible (Arc clone) — kept `Result` return type to avoid touching 13+ call sites
  - `push_block()` is now async (acquires write lock)

  ### `ExporterStorage` / `BlockProcessorStorage`
  - Read operations (`get_block_with_blob_ids`, `get_block_with_blobs`, `get_latest_index`) acquire a read lock
  - Write operations (`push_block`, `save`) acquire a write lock
  - `clone()` takes `&self` instead of `&mut self`

## Test Plan

- `cargo build -p linera-service` — compiles with no errors
- `cargo test -p linera-service --bin linera-exporter` — all existing tests pass (including `test_committee_destination`)

## Release Plan

  - These changes should be released to testnet, and then
  - Backported to `main`.

## Links

### DIFF
--- a/linera-service/src/exporter/metrics.rs
+++ b/linera-service/src/exporter/metrics.rs
@@ -96,14 +96,6 @@ pub(crate) static NOTIFICATIONS_RECEIVED: LazyLock<IntCounter> = LazyLock::new(|
     )
 });
 
-pub(crate) static BLOCKS_INDEXED: LazyLock<IntCounter> = LazyLock::new(|| {
-    prometheus_util::register_int_counter_with_subsystem(
-        "exporter",
-        "blocks_indexed",
-        "Number of blocks indexed into canonical state",
-    )
-});
-
 pub(crate) static CANONICAL_STATE_HEIGHT: LazyLock<IntGauge> = LazyLock::new(|| {
     prometheus_util::register_int_gauge_with_subsystem(
         "exporter",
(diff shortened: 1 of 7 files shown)


deuszx:
Superseeded by https://github.com/linera-io/linera-protocol/pull/5449

PR #5438, opened 2026-02-13
Generic benchmark operation generator

ndr-ds:
## Proposal

Introduce an `OperationGenerator` trait that decouples operation generation from the benchmark loop, enabling custom
application benchmarks without touching the core benchmark infrastructure.

Key changes:

- **New `OperationGenerator` trait**: defines a `generate_operations` interface that produces operations dynamically
per block instead of pre-generating them. External crates can implement this trait to benchmark arbitrary applications
(e.g., prediction market order benchmarking in `pm-app`).
- **Two built-in generators**: `NativeFungibleTransferGenerator` and `FungibleTransferGenerator`, which replicate the
existing behavior for native and fungible token transfer benchmarks respectively.
- **Backport `--single-destination-per-block` CLI flag**: this already existed on `main` but was missing from the
testnet branch.
- **Per-chain unique owners**: each chain now gets its own unique owner instead of sharing one across a batch,
avoiding conflicts during concurrent benchmarking.
- **Removed `make_benchmark_block_info`**: replaced by dynamic per-block generation through the `OperationGenerator`
trait.
- **Skip wallet lookup when `close_chains` is true**: avoids picking up existing chains from the wallet that would
then be closed after the benchmark, and skips writing the new temporary chains back to the wallet since they won't be
reused.

## Test Plan

- CI
- Manual benchmarking with both native and fungible tokens, with and without `--single-destination-per-block`

### DIFF
--- a/linera-client/src/benchmark.rs
+++ b/linera-client/src/benchmark.rs
@@ -24,7 +24,7 @@ use linera_execution::{system::SystemOperation, Operation};
 use linera_sdk::abis::fungible::{self, FungibleOperation};
 use num_format::{Locale, ToFormattedString};
 use prometheus_parse::{HistogramCount, Scrape, Value};
-use rand::{seq::SliceRandom, thread_rng};
+use rand::{rngs::SmallRng, seq::SliceRandom, thread_rng, SeedableRng};
 use serde::{Deserialize, Serialize};
 use tokio::{
     sync::{mpsc, Barrier, Notify},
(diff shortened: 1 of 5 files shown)


linera-client/src/benchmark.rs
```diff
@@ -35,6 +35,162 @@ use tracing::{debug, error, info, warn, Instrument as _};
...
+        Ok(Self {
+            source_chain_id,
+            destination_chains,
+            destination_index: 0,
+            rng,
+            single_destination_per_block,
+        })
+    }
+
+    fn get_next_destination(&mut self) -> ChainId {
```
ma2bd: nit: remove `get_` (by analogy with iterators)?

PR #5440, opened 2026-02-13
Replace linear retry backoff with full jitter exponential backoff

ndr-ds:
## Motivation

The retry logic in gRPC client requests, gRPC subscription reconnection, and cross-chain message forwarding all used
linear backoff without jitter (`delay * retry_count`). This creates a [thundering herd](https://en.wikipedia.org/wiki/Thundering_herd_problem) risk: when a validator goes down and comes back up, all clients that were retrying wake up at nearly the same time and hit the recovering validator simultaneously, potentially bringing it down again.

This happens because linear backoff is deterministic — every client on the same retry count sleeps the exact same
duration, so their retries synchronize into bursts.

## Proposal

Replace all three retry sites with [Full Jitter exponential backoff](https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/), the industry-standard approach recommended by [AWS](https://docs.aws.amazon.com/sdkref/latest/guide/feature-retry-behavior.html), [Google Cloud](https://cloud.google.com/storage/docs/retry-rategy), and the [gRPC spec](https://github.com/grpc/grpc/blob/master/doc/connection-backoff.md).

The formula is: `sleep = random(0, min(cap, base * 2^attempt))`.

- Exponential growth spaces retries further apart over time
- Full randomization (jitter across the entire range, not just an additive offset) desynchronizes clients so they
spread their retries evenly instead of clustering
- A fixed 30s cap prevents excessive wait times (Google Cloud uses 30s, AWS uses 20s)

## Test Plan

- CI

## Release Plan


ndr-ds:
* **#5440** <a href="https://app.graphite.com/github/pr/linera-io/linera-protocol/5440?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/linera-io/linera-protocol/5440?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

linera-rpc/src/lib.rs
```diff
@@ -58,3 +58,30 @@ pub const FILE_DESCRIPTOR_SET: &[u8] = tonic::include_file_descriptor_set!("file
 pub const CERT_PEM: &str = include_str!(concat!(env!("OUT_DIR"), "/self_signed_cert.pem"));
 #[cfg(not(target_arch = "wasm32"))]
 pub const KEY_PEM: &str = include_str!(concat!(env!("OUT_DIR"), "/private_key.pem"));
+
+/// Maximum bit-shift exponent for u32: `1u32 << 31` is the largest valid shift before overflow.
+const MAX_SHIFT_EXPONENT: u32 = 31;
```
ma2bd: Please no hardcoded constants.

linera-client/src/client_options.rs
```diff
@@ -93,6 +93,14 @@ pub struct Options {
     #[arg(long, default_value = "10")]
     pub max_retries: u32,
 
+    /// Maximum backoff delay for retrying to connect to a validator.
+    #[arg(
+        long = "max-backoff-ms",
+        default_value = "30000",
+        value_parser = util::parse_millis
+    )]
+    pub max_backoff: Duration,
```
deuszx, line 102: s/max_backoff/max_backoff_ms

linera-rpc/src/lib.rs
deuszx: This one seems quite sensible. I can't imagine us changing 👀

deuszx (approved):

linera-rpc/src/lib.rs
```diff
@@ -58,3 +58,23 @@ pub const FILE_DESCRIPTOR_SET: &[u8] = tonic::include_file_descriptor_set!("file
...
+///
+/// Uses the AWS-recommended formula: `sleep = random(0, min(cap, base * 2^attempt))`.
+/// Reference: <https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/>
+pub(crate) fn full_jitter_delay(
+    base_delay: std::time::Duration,
+    attempt: u32,
+    max_backoff: std::time::Duration,
+) -> std::time::Duration {
+    use rand::Rng as _;
+    let exponential_delay = base_delay.saturating_mul(1u32 << attempt.min(MAX_SHIFT_EXPONENT));
```
afck: We could use `1u32.overflowing_shl(attempt).unwrap_or(u32::MAX)`. Then we don't need the constant.

Or even `2u32.saturating_pow(attempt)`? I'm sure the compiler will realize it's a power of 2!

ndr-ds: I had claude create some test code and generate the assembly for it, and surprisingly the compiler (at least on edition 2021) doesn't optimize `2u32.saturating_pow(attempt)` to a
shift, it generates a long branching sequence.  
TIL though about `1u32.checked_shl(attempt).unwrap_or(u32::MAX)`, which literally does the same thing I was doing by hand ha and compiles to the
same 6-instruction sequence as the original shift, so we should be good.

linera-rpc/src/lib.rs
ma2bd: It's a bit more than 32 -1

PR #5441, opened 2026-02-13
Respect ModuleCache max size

ndr-ds:
## Motivation

`ModuleCache::insert()` never incremented `total_size` after inserting a new entry. It only decremented during
eviction in `reduce_size_to()`. This meant:

- `total_size` was permanently 0
- The eviction check `if self.total_size + bytecode_size > self.max_size` never triggered
- The cache grew unboundedly despite the 512 MiB limit
- `reduce_size_to()` would underflow on the subtraction `self.total_size -= bytecode_size`

## Proposal

- Add `self.total_size += bytecode_size` after insertion, only when the key wasn't already present (to avoid
double-counting on re-insertion of the same bytecode)
- Add an early return when a single bytecode exceeds `max_size`, which prevents an underflow in the `self.max_size -
bytecode_size` computation passed to `reduce_size_to()`
- Add 4 unit tests covering: size tracking, eviction triggering, oversized rejection, and duplicate key handling

## Test Plan

- CI + new tests pass

## Release Plan


ndr-ds:
* **#5441** <a href="https://app.graphite.com/github/pr/linera-io/linera-protocol/5441?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/linera-io/linera-protocol/5441?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

linera-execution/src/wasm/module_cache.rs
```diff
@@ -58,11 +69,18 @@ impl<Module: Clone> ModuleCache<Module> {
...
+
         if self.total_size + bytecode_size > self.max_size {
             self.reduce_size_to(self.max_size - bytecode_size);
         }
 
-        self.modules.put(bytecode, module);
+        let already_present = self.modules.put(bytecode, module).is_some();
+        if !already_present {
+            self.total_size += bytecode_size;
+        }
```
deuszx: Should we first check if it's already present BEFORE we `reduce_size_to`?

afck: Right, good catch!

linera-execution/src/wasm/module_cache.rs
```diff
@@ -58,11 +69,21 @@ impl<Module: Clone> ModuleCache<Module> {
...
         let bytecode_size = bytecode.as_ref().len() as u64;
 
+        if bytecode_size > self.max_size {
+            return;
+        }
+
+        if self.modules.contains(&bytecode) {
+            self.modules.put(bytecode, module);
+            return;
+        }
```
afck: Is this equivalent? No need to `put` if the module is already in the cache?

```suggestion
        if self.modules.promote(&bytecode) {
            return;
        }
```

ndr-ds: Good point, I need up upgrade the `lru` crate version for that though. Will do it

PR #5446, opened 2026-02-14
[testnet] Remove the handling of the sizes.

MathieuDutSik:
## Motivation

We are tracking the size of the keys in the smart contract. This is quite an expensive operation. This is
especially bad because we are not using this information.

Potentially, there are two usages:
* For storage costs, the code is missing. But see https://github.com/linera-io/linera-protocol/pull/5330
* For the ¨ has_empty_storage` this is used only for the EVM, but that is not implemented in testnet Conway.

## Proposal

Remove the code that accesses the storages so as to get the speed increase.
However, the entries are kept in the `KeyValueStoreView` and `KeyTag` so as to ensure backward compatibility with already deployed contracts.

## Test Plan

CI.

## Release Plan

There will be other issues to resolve

## Links

### DIFF
--- a/linera-views/src/views/key_value_store_view.rs
+++ b/linera-views/src/views/key_value_store_view.rs
@@ -137,16 +137,11 @@ use {
 };
 
 #[repr(u8)]
-#[allow(dead_code)]
 enum KeyTag {
     /// Prefix for the indices of the view.
     Index = MIN_VIEW_TAG,
-    /// The total stored size
-    TotalSize,
-    /// The prefix where the sizes are being stored
-    Sizes,
-    /// Prefix for the hash.
-    Hash,
+    /// Prefix for the hash. We fix the index for compatibility with existing contracts.
+    Hash = MIN_VIEW_TAG + 3,
 }
 
 /// A pair containing the key and value size.


linera-views/src/views/key_value_store_view.rs
ma2bd: It would be nice to avoid loading the old size maps over and over: we could stop loading them (and remove the fields in the View) or we could load them but clear the data for next time (if non-empty) in `post_save`.

MathieuDutSik: Well, in that case the `post_load` of `ByteMapView is just:
```rust
        Ok(Self {
            context,
            updates: BTreeMap::new(),
            deletion_set: DeletionSet::new(),
        })
```

So good news, nothing to load but bad news, we need a context, so we need a clone of it even if not used.

About deleting the data on storage. This can be done by adding a `DeleteByPrefix` in the code. However, we cannot test if what we delete is non-trivial without doing some expensive `find_keys_by_prefix`. So two realistic possibilities:
* Add a delete_by_prefix that may slow a little every operations.
* Make our mind with it and keep data that we do not need.

I left the code in Option 2, let me know if you want Option 1.

ma2bd: Thanks for verifying. Yes we can leave the data around in storage. Can we delete the unused field from the view in this case?

MathieuDutSik: Yes, I was wrong. No need for those entries.

ma2bd: nit: I wonder if we could remove `#[allow(dead_code)]` and write `Hash = MIN_VIEW_TAG + 3` here

MathieuDutSik: Yes, we can though that does not impact the generated code.

ma2bd: for sure

PR #5417, opened 2026-02-10
Correct the tests that fails.

MathieuDutSik:
## Motivation

After the commit 6f5e8ab5e7adeb79dc59daf87f4273a5e631ee85, the test
```bash
cargo test test_end_to_end_benchmark::scylladb_grpc --features scylladb
```

fails.

## Proposal

Gate the access with `opentelemetry`. Introduce a test for those tests in the CI command.

## Test Plan

CI, that is the point of the PR.

## Release Plan

- Nothing to do / These changes follow the usual release cycle.

## Links

None.

PR #5447, opened 2026-02-15
[testnet] Correct the tests that fails

MathieuDutSik:
## Motivation

The same problem occurs for testnet_conway.
See https://github.com/linera-io/linera-protocol/pull/5417

## Proposal

Same correction

## Test Plan

CI.

## Release Plan

To testnet_conway.

## Links

None


ma2bd (commented):
thanks!

PR #5456, opened 2026-02-17
Linera protocol verification in Solidity.

deuszx:
## Motivation

Bring liquidity into Linera ecosystem.

## Proposal

This is the first PR in a series. We introduce some basic repo structure, CI, etc. The most important part is `LightClient.sol` and `Microchain.sol` contracts. More in-depth explanation can be found in the `README.md` but below I give short introduction:

#### `LightClient.sol`

An admin contract that tracks committee epochs and verifies certificate signatures. It exposes `verifyBlock(bytes)` for other contracts to call, returning the deserialized `Block` on success. It does not store blocks currently.

#### `Microchain.sol`

An abstract contract that tracks blocks for a single Linera microchain. It delegates certificate verification to a `LightClient` instance and enforces chain ID matching and sequential block heights. Concrete subcontracts implement `_onBlock()` to define application-specific logic (e.g., tracking token transfers).

### How we get Solidity types?

In `linera-bridge/tests/format.rs` we create the `serde_reflaction::Tracer` instance which generates a yaml snapshot. Then, in `build.rs` we parse that snapshot and use `serde_generate::solidity` package to generate solidity types – file can be found in `src/solidity/BridgeTypes.sol`.

### Key design decisions

In `LightClient.sol` we have to change a validator set on committee transition. Linera's `ValidatorPublicKey` is in a a compressed variant and it's impossible to derive Ethereum's `Address` type from it. We need uncompressed version, which is passed in as an argument. Contract verifies the additional input validators are matching ones in the committee blob.

Partial parsing of the committee blob: blob referenced in `CreateCommittee` includes `ResourceControlPolicy` which has a couple dozen fields and deserialization will be complicated and expensive.

Read the **Key design decisions** section in the README for more.

## Test Plan

CI.

## Release Plan

- None for now.

## Links

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -225,7 +225,7 @@ jobs:
     - name: Run all tests using the default features (except storage-service)
       run: |
         # TODO(#2764): Actually link this to the default features
-        cargo test --workspace --no-default-features --features fs,macros,wasmer,rocksdb --locked --exclude linera-bridge
+        cargo test --no-default-features --features fs,macros,wasmer,rocksdb --locked
     - name: Run Witty integration tests
       run: |
         cargo test -p linera-witty --features wasmer,wasmtime
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,7 +34,7 @@ members = [
 ]
 default-members = [
     "linera-base",
-    "linera-bridge",
+    # "linera-bridge", # tested separately in CI (requires solc)
     "linera-chain",
     "linera-client",
     "linera-core",


linera-bridge/src/solidity/Microchain.sol
```diff
@@ -0,0 +1,39 @@
...
+    /// into a hash chain. This is safe because `ConfirmedBlockCertificate`
+    /// implies BFT-finalized canonicality — a quorum of validators signed this
+    /// specific block at this height, so no conflicting block can exist.
+    /// If this assumption ever changes at the protocol layer, a
+    /// `previous_block_hash` check should be added here.
+    function addBlock(bytes calldata data) external {
+        BridgeTypes.Block memory blockValue = lightClient.verifyBlock(data);
+
+        require(blockValue.header.chain_id.value.value == chainId, "chain id mismatch");
+        require(blockValue.header.height.value == latestHeight + 1, "block height must be sequential");
```
deuszx, line 30: We might have to reconsider this at some point. Maybe diffs !=1 should be allowed.

linera-bridge/src/solidity/Microchain.sol
```diff
@@ -0,0 +1,39 @@
+// SPDX-License-Identifier: Apache-2.0
```
Twey, line 1: We should probably keep our standard copyright notice on Solidity files too.

linera-bridge/build.rs
```diff
@@ -0,0 +1,34 @@
...
+        .splitn(3, "---")
+        .nth(2)
+        .expect("snapshot file missing insta header");
+
+    let registry: Registry =
+        serde_yaml::from_str(yaml).expect("failed to parse YAML registry from snapshot");
+
+    let out_dir = PathBuf::from("src");
+    let installer = solidity::Installer::new(out_dir);
+    let config = CodeGeneratorConfig::new("BridgeTypes".to_string());
```
ma2bd, line 30: nice to see @MathieuDutSik's work on serdegen being directly applicable

linera-bridge/src/microchain.rs
```diff
@@ -0,0 +1,143 @@
+// Copyright (c) Zefchain Labs, Inc.
```
ma2bd, line 1: nit: we actually don't use the (marketing?) word "microchain" as much in our codebase but just use "chain".

linera-chain/src/data_types/mod.rs
```diff
@@ -490,8 +490,9 @@ impl LiteValue {
     }
 }
 
+//(deuszx): pub is temp.
```
ma2bd, line 493: nit: Create a TODO(#nnn) and a task to remove it then?

linera-bridge/src/microchain.rs
ma2bd: But on second thoughts, maybe that's fine in the context of an EVM host chain

linera-bridge/src/solidity/Microchain.sol
ma2bd: Probably we need to support sparse events

linera-bridge/src/solidity/LightClient.sol
```diff
@@ -0,0 +1,283 @@
...
+    // Per-epoch committee storage
+    struct EpochCommittee {
+        mapping(address => uint64) weights;
+        mapping(address => uint256) indices; // 1-indexed; 0 means not a member
+        uint256 validatorCount;
+        uint64 totalWeight;
+        uint64 quorumThreshold;
+    }
+    mapping(uint32 => EpochCommittee) private committees;
+    uint32 public currentEpoch;
```
ma2bd, line 16: Acceptable (trusted) epochs are typically a range in Linera. The bridge could have a stricter policy but I suspect we can't entirely remove the range.

cc @afck

deuszx: Right, of course. Certificates from multiple epochs are valid (and should be accepted).

The place where this value actually is used is in `addCommittee` and basically it allows rotating committee with a certificate from the `currentEpoch` (and not any ~~later~~ earlier ones).

What needs to be changed is `verifyBlock` that should probably refuse to verify blocks from unbonded epochs.

linera-bridge/src/solidity/LightClient.sol
```diff
@@ -0,0 +1,283 @@
...
+    }
+
+    function _setCommittee(uint32 epoch, address[] memory validators, uint64[] memory weights) internal {
+        require(epoch == currentEpoch + 1 || (epoch == 0 && currentEpoch == 0), "epoch must be sequential");
+        require(validators.length == weights.length, "length mismatch");
+        EpochCommittee storage committee = committees[epoch];
+        uint64 total = 0;
+        for (uint256 i = 0; i < validators.length; i++) {
+            require(committee.weights[validators[i]] == 0, "duplicate validator");
+            committee.weights[validators[i]] = weights[i];
```
ndr-ds, line 168: Should we require that `weights[i] > 0`before this?

linera-bridge/src/solidity/LightClient.sol
```diff
@@ -0,0 +1,283 @@
...
+
+    /// Copies a slice of memory bytes into a new bytes array.
+    function _sliceMemory(bytes memory data, uint256 start, uint256 end) internal pure returns (bytes memory) {
+        uint256 len = end - start;
+        bytes memory result = new bytes(len);
+        for (uint256 i = 0; i < len; i++) {
+            result[i] = data[start + i];
+        }
+        return result;
+    }
```
ndr-ds, line 282: This might not be a big deal, but since this slices a serialized block (which can be up to 1MB), it might be worth copying 32 bytes at a time with assembly instead of byte-by-byte, to save gas. Like is done here https://github.com/GNSPS/solidity-bytes-utils/blob/master/contracts/BytesLib.sol for example. Something like:

```suggestion
    /// Copies a slice of memory bytes into a new bytes array.
    function _sliceMemory(bytes memory data, uint256 start, uint256 end) internal pure returns (bytes memory) {
        uint256 len = end - start;
        bytes memory result = new bytes(len);
        assembly {
            let resultPtr := add(result, 0x20)
            let dataPtr := add(add(data, 0x20), start)
            for { let i := 0 } lt(i, len) { i := add(i, 0x20) } {
                mstore(add(resultPtr, i), mload(add(dataPtr, i)))
            }
        }
        return result;
    }
```

deuszx: Thanks for the resource! I am planning on doing the gas optimisation pass later on. I'll check if this affects the gas cost.

linera-bridge/src/solidity/LightClient.sol
```diff
@@ -0,0 +1,283 @@
...
+            if (txn.choice != 1) continue;
+            BridgeTypes.Operation memory op = txn.execute_operation;
+            // choice=0 is System
+            if (op.choice != 0) continue;
+            BridgeTypes.SystemOperation memory sysOp = op.system;
+            // choice=10 is Admin
+            if (sysOp.choice != 10) continue;
+            BridgeTypes.AdminOperation memory adminOp = sysOp.admin;
+            // choice=1 is CreateCommittee
+            if (adminOp.choice != 1) continue;
```
afck, line 52: Is all of this equivalent with `adminOp.create_committee.epoch.value != 0`?

linera-bridge/src/solidity/LightClient.sol
```diff
@@ -0,0 +1,283 @@
...
+        // Parse blob to extract addresses and weights, verified against caller's keys
+        (address[] memory addrs, uint64[] memory weights) = _parseCommitteeBlob(committeeBlob, validators);
+
+        // Store the new committee
+        _setCommittee(newEpoch, addrs, weights);
+    }
+
+    function verifyBlock(bytes calldata data) external view returns (BridgeTypes.Block memory) {
+        return verifyCertificate(data);
+    }
```
afck, line 81: Is this function necessary then, if it's the same as `verifyCertificate`?

linera-bridge/src/solidity/LightClient.sol
```diff
@@ -0,0 +1,283 @@
...
+            require(uint256(s) <= SECP256K1_N / 2, "non-canonical high-s signature");
+
+            // Try v=27 and v=28 since we don't have the recovery ID
+            address recovered = ecrecover(signedHash, 27, r, s);
+            if (recovered == address(0) || committee.weights[recovered] == 0) {
+                recovered = ecrecover(signedHash, 28, r, s);
+            }
+            require(recovered != address(0), "signature recovery failed");
+            uint64 w = committee.weights[recovered];
+            require(w > 0, "unknown validator");
```
afck, line 147: I think we don't prohibit weight 0 validators at the moment.

linera-bridge/src/solidity/LightClient.sol
```diff
@@ -0,0 +1,283 @@
...
+        uint64 total = 0;
+        for (uint256 i = 0; i < validators.length; i++) {
+            require(committee.weights[validators[i]] == 0, "duplicate validator");
+            committee.weights[validators[i]] = weights[i];
+            committee.indices[validators[i]] = i + 1; // 1-indexed
+            total += weights[i];
+        }
+        committee.validatorCount = validators.length;
+        committee.totalWeight = total;
+        committee.quorumThreshold = 2 * total / 3 + 1;
```
afck, line 174: (When porting this to `main` we need to remember that we slightly tweaked the rounding here.)

linera-bridge/src/solidity/LightClient.sol
```diff
@@ -0,0 +1,283 @@
...
+        uint256 x;
+        uint256 y;
+        assembly {
+            x := calldataload(uncompressed.offset)
+            y := calldataload(add(uncompressed.offset, 32))
+        }
+        uint256 lhs = mulmod(y, y, SECP256K1_P);
+        uint256 x2 = mulmod(x, x, SECP256K1_P);
+        uint256 rhs = addmod(mulmod(x2, x, SECP256K1_P), 7, SECP256K1_P);
+        require(lhs == rhs, "key not on secp256k1 curve");
```
afck, line 262: (And there's no builtin for that?)

deuszx: Apparently erecover doesn't check that. But I'll verify again 👍

linera-bridge/src/solidity/LightClient.sol
deuszx: might be a leftover 👍

deuszx:
Let me merge this for now and address in followup PRs. I have a couple of others waiting in a line.

PR #5464, opened 2026-02-18
Retry in the task processor

afck:
## Motivation

The task processor currently doesn't retry failed tasks.

## Proposal

In each batch, we submit as many task outcomes to the chain as possible, in the correct order.

When a task fails to produce an outcome or cannot be submitted to the chain, it is retried with a configurable delay (default 5 seconds). Exceptions:
* If it fails due to a conflicting block, it is retried immediately.
* If it fails because we are not the round leader on the chain, it is retried when the round times out.

## Test Plan

This is mostly about relatively simple but hard-to-isolate time-based retry logic, so it's difficult to write tests without actual `sleep`s. CI should catch any regressions.

## Release Plan

* Backport to `testnet_conway`.
* Release a new SDK.

## Links

### DIFF
--- a/linera-service/src/task_processor.rs
+++ b/linera-service/src/task_processor.rs
@@ -277,10 +277,15 @@ impl<Env: linera_core::Environment> TaskProcessor<Env> {
                             }
                         }
                     }
-                    let _ = batch_sender.send(BatchResult {
-                        application_id,
-                        retry_at,
-                    });
+                    if batch_sender
+                        .send(BatchResult {
+                            application_id,
+                            retry_at,
+                        })
+                        .is_err()
+                    {
+                        error!("Batch receiver dropped for {application_id}");
+                    }
                 });
             }
         }


CLI.md
```diff
@@ -741,6 +741,9 @@ Run a GraphQL service to explore and extend the chains of the wallet
 * `--operator-application-ids <OPERATOR_APPLICATION_IDS>` — Application IDs of operator applications to watch. When specified, a task processor is started alongside the node service
```
ma2bd, line 741: what would be a better name for this? (something with task?)

linera-service/src/task_processor.rs
```diff
@@ -242,37 +244,43 @@ impl<Env: linera_core::Environment> TaskProcessor<Env> {
...
                     }
-                    // Signal that this batch is done so the main loop can process
-                    // the next batch for this application.
-                    if sender
-                        .send(TaskMessage::BatchComplete { application_id })
-                        .is_err()
-                    {
-                        error!("Outcome receiver dropped for {application_id}");
-                    }
+                    let _ = batch_sender.send(BatchResult {
```
ma2bd: remove the let _

CLI.md
afck: `--task-operator-application-ids` then? :thinking:

afck: Then it should also be `--task-operators` below.

But maybe that's a change for a separate PR?

ma2bd: ok let's review the naming separately

PR #5465, opened 2026-02-18
Add the binary_dir to LocalNet, LocalNetConfig, and the Wallet.

MathieuDutSik:
## Motivation

For the developer experience, it is advantageous to be able to run end-to-end tests. Something similar
to what we have for `local_net_tests.rs`. We have used this as 

## Proposal

Make the directory available in the `LocalNetConfig`, `LocalNet`, and `Wallet`.
If set to `None`, then the usual system is being used.

## Test Plan

The test `test_wasm_counter` of [linera_end_to_end_tests](https://github.com/MathieuDutSik/linera_end_to_end_tests)
demonstrates the functionality.

## Release Plan

It would be interesting to deploy it for `testnet_conway` since the question is more relevant there.

## Links

None

### DIFF
--- a/linera-service/src/cli/main.rs
+++ b/linera-service/src/cli/main.rs
@@ -901,6 +901,7 @@ impl Runnable for Job {
                                         "50".to_string(),
                                         "--timings".to_string(),
                                     ],
+                                    None,
                                 )))
                             })
                             .collect::<Result<Vec<_>, anyhow::Error>>()?;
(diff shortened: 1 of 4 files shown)


MathieuDutSik:
> So the idea is that e.g. another project with a `linera-protocol` subdirectory would pass in something like `env!("CARGO_PKG_NAME").join("linera-protocol")`?

It could be that, but it does not have to. You may want to experiment.

If you want to run some end-to-end tests in some external project that is not in `linera-protocol`, that ought
to be as easy as if you were there (except for putting the directory in question).

PR #5471, opened 2026-02-18
Make the task processor cursor a String.

afck:
## Motivation

`Vec<u8>` is unreadable in the logs.

## Proposal

Make the task processor cursor a `String`. Apps are encouraged to use a readable format, e.g. JSON.

## Test Plan

CI

## Release Plan

- Release in a new SDK.

## Links

### DIFF
--- a/linera-base/src/task_processor.rs
+++ b/linera-base/src/task_processor.rs
@@ -27,7 +27,7 @@ pub struct ProcessorActions {
     pub request_callback: Option<Timestamp>,
     /// An optional cursor for the task processor to store and pass to the application
     /// upon the next query for actions.
-    pub set_cursor: Option<Vec<u8>>,
+    pub set_cursor: Option<String>,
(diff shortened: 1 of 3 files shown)


linera-base/src/task_processor.rs
```diff
@@ -27,7 +27,7 @@ pub struct ProcessorActions {
     pub request_callback: Option<Timestamp>,
     /// An optional cursor for the task processor to store and pass to the application
     /// upon the next query for actions.
-    pub set_cursor: Option<Vec<u8>>,
+    pub set_cursor: Option<String>,
```
deuszx, line 30: or `T: Display`?

afck: Where would `T` come from? The task processor doesn't know specific apps' associated types.

deuszx: Right 🤔 I thought it can be defined at the time of instantiation

PR #5474, opened 2026-02-18
Remove timeout on proxy TCP check in remote-kubernetes-net-test

ndr-ds:
## Motivation

The `remote-kubernetes-net-test` workflow fails on the merge queue because the `timeout
120` on the TCP proxy readiness check is too aggressive. The proxy port-forward may take
longer than 2 minutes to become available after validators report READY, causing exit
code 124 (killed by timeout).

## Proposal

- Remove `timeout 120` from the TCP proxy readiness check, using a bare while loop
instead. The job-level `timeout-minutes: 90` serves as the safeguard against infinite
hangs.
- Add `workflow_dispatch` to the job's `if` condition so the workflow can be manually
triggered for testing without going through the merge queue.

## Test Plan

- Trigger the workflow manually via `workflow_dispatch` to verify it passes before
merging.

## Release Plan


ndr-ds:
* **#5474** <a href="https://app.graphite.com/github/pr/linera-io/linera-protocol/5474?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/linera-io/linera-protocol/5474?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #5480, opened 2026-02-19
Upgrade the `linera-indexer` to more idiomatic code.

MathieuDutSik:
## Motivation

An examination of the code of `linera-indexer` revealed some code duplications that can be eliminated.

## Proposal

Simplify the load and do other simplifications. The use of Claude was iterated to find
simplification of the code.

Only one functional change was done: When the `get_value` failed to get the block, emit some trace::error in the log.

Possible further work: Removing the `.unwrap()` statements and avoiding the code duplication in sqlite / postgresql.

## Test Plan

The CI.

## Release Plan

Can be backported if deemed important.

## Links

None.

### DIFF
--- a/linera-indexer/lib/src/service.rs
+++ b/linera-indexer/lib/src/service.rs
@@ -153,7 +153,7 @@ impl Listener {
                         if let Reason::NewBlock { hash, .. } = data.notifications.reason {
                             match self.service.get_value(chain_id, Some(hash)).await {
                                 Ok(value) => indexer.process(self, &value).await?,
-                                Err(e) => error!("failed to fetch block {:?}: {}", hash, e),
+                                Err(error) => error!("failed to fetch block {hash}: {error}"),
                             }
                         }
                     } else {


linera-indexer/lib/src/indexer.rs
```diff
@@ -180,11 +177,12 @@ where
         plugin: impl Plugin<D> + 'static,
     ) -> Result<(), IndexerError> {
         let name = plugin.name();
-        self.plugins
-            .insert(name.clone(), Box::new(plugin))
-            .map_or_else(|| Ok(()), |_| Err(IndexerError::PluginAlreadyRegistered))?;
+        if self.plugins.insert(name.clone(), Box::new(plugin)).is_some() {
```
Twey: It's strange to update it _and then_ error.  Maybe use the entry API to avoid replacing it?

MathieuDutSik: Yes, corrected, thank you.

linera-indexer/lib/src/service.rs
```diff
@@ -151,8 +151,9 @@ impl Listener {
                 Ok(response) => {
                     if let Some(data) = response.data {
                         if let Reason::NewBlock { hash, .. } = data.notifications.reason {
-                            if let Ok(value) = self.service.get_value(chain_id, Some(hash)).await {
-                                indexer.process(self, &value).await?;
+                            match self.service.get_value(chain_id, Some(hash)).await {
+                                Ok(value) => indexer.process(self, &value).await?,
+                                Err(e) => error!("failed to fetch block {:?}: {}", hash, e),
```
afck:
```suggestion
                                Err(e) => error!("failed to fetch block {hash}: {e}"),
```

Or:


```suggestion
                                Err(e) => error!(%hash, %e, "failed to fetch block"),
```

(But then I'd rename `e` to `error`.)

PR #5484, opened 2026-02-19
Make the controller set appropriate message policies on service chain listeners

### LINKED ISSUE #5224: Allow `examples/controller` to change the client policy of its node-service workers

bart-linera:
## Motivation

Currently, anyone can send any message to any chain managed by the controller, and the workers will automatically receive and process those messages. This creates potential vulnerabilities in the system.

## Proposal

Make workers reject message bundles addressed to service chains that contain no messages for relevant applications and no system messages.

(Draft for now - I will also extend the controller test to check that this works.)

## Test Plan

CI

## Release Plan

    - be released in a new SDK,

## Links

- closes #5224 

### DIFF
--- a/linera-chain/src/data_types/mod.rs
+++ b/linera-chain/src/data_types/mod.rs
@@ -295,12 +295,12 @@ impl IncomingBundle {
                 return false;
             }
         }
-        true
+        !policy.is_reject()
     }
 
     #[instrument(level = "trace", skip(self))]
     pub fn apply_policy(mut self, policy: &MessagePolicy) -> Option<IncomingBundle> {
-        if !self.matches_policy(policy) || policy.is_reject() {
+        if !self.matches_policy(policy) {
             if self.bundle.is_skippable() {
                 return None;
             } else if !self.bundle.is_protected() {


linera-chain/src/data_types/mod.rs
```diff
@@ -273,30 +273,34 @@ impl IncomingBundle {
...
+                return false;
             }
         }
-        if policy.is_reject() {
+        true
+    }
+
+    #[instrument(level = "trace", skip(self))]
+    pub fn apply_policy(mut self, policy: &MessagePolicy) -> Option<IncomingBundle> {
+        if !self.matches_policy(policy) || policy.is_reject() {
```
afck: Should `matches_policy` take into account `is_reject` directly? (Are we using `is_reject` anywhere else at all?)

bart-linera: Hmm, good point, that would make sense :+1:

PR #5468, opened 2026-02-18
Chain actor fairness

afck:
## Motivation

When e.g. running `linera service`, lots of read-only GraphQL queries can clog the chain worker queues and prevent write requests from being processed.

## Proposal

Separate reads and writes; prioritize writes.

(The separate queues should also get us one step closer to #2237.)

## Test Plan

CI

## Release Plan

- Backport to `testnet_conway`.
- Release a new SDK.

## Links

- [reviewer checklist](https://github.com/linera-io/linera-protocol/blob/main/CONTRIBUTING.md#reviewer-checklist)

PR #5486, opened 2026-02-19
Concurrent read-only chain actor read requests.

### LINKED ISSUE #2237: Execute read-only chain operations concurrently
# Motivation

The `ChainWorkerActor` handles one request at a time. However, many of the requests don't change the chain (or rather, rolls back the changes after it has finished). That means that these "read-only" operations could be executed concurrently.

# Proposal

Use `ClonableView::clone_unchecked` to create temporary clones of the `ChainStateView` and spawn tasks for each read-only worker operation.

afck:
## Motivation

Read-only requests to the chain actor could be handled concurrently.

## Proposal

Build on https://github.com/linera-io/linera-protocol/pull/5468 to handle all read requests concurrently.

## Test Plan

CI; we should benchmark this and compare.

## Release Plan

- Release a new SDK.

## Links

- Extends https://github.com/linera-io/linera-protocol/pull/5468.
- Closes #2237.

### DIFF
--- a/linera-core/src/chain_worker/actor.rs
+++ b/linera-core/src/chain_worker/actor.rs
@@ -11,6 +11,7 @@ use std::{
 
 use custom_debug_derive::Debug;
 use futures::{
+    future::Either,
     stream::{FuturesUnordered, StreamExt as _},
     FutureExt,
 };
@@ -51,10 +52,27 @@ pub(crate) type EventSubscriptionsResult = Vec<((ChainId, StreamId), EventSubscr
 /// A request item sent through a channel.
 type RequestItem<Ctx> = (ChainWorkerRequest<Ctx>, tracing::Span, Instant);
 
-/// The endpoint for sending requests to a [`ChainWorkerActor`].
+/// Creates a dual-channel pair for communicating with a [`ChainWorkerActor`].
 ///
 /// Read-only requests and write requests use separate channels so that
 /// the actor can schedule them with a fairness policy.
+pub(crate) fn chain_actor_channel<Ctx: Context + Clone + 'static>(
+) -> (ChainActorEndpoint<Ctx>, ChainActorReceivers<Ctx>) {
+    let (read_sender, read_receiver) = mpsc::unbounded_channel();
+    let (write_sender, write_receiver) = mpsc::unbounded_channel();
+    (
+        ChainActorEndpoint {
+            read_sender,
+            write_sender,
+        },
+        ChainActorReceivers {
+            read_receiver,
+            write_receiver,
+        },
+    )
+}
+
+/// The endpoint for sending requests to a [`ChainWorkerActor`].
 #[derive(Clone)]
 pub(crate) struct ChainActorEndpoint<Ctx: Context + Clone + 'static> {
     read_sender: mpsc::UnboundedSender<RequestItem<Ctx>>,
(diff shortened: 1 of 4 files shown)


linera-core/src/chain_worker/actor.rs
```diff
@@ -496,28 +619,101 @@ where
...
+                            ));
+                        }
+                        for (fut, span) in owned_futures {
+                            read_futures.push(Box::pin(fut.instrument(span)));
+                        }
+                        while read_futures.next().await.is_some() {}
+                        continue; // Back to phase 1 to check for writes.
+                    }
+                }
+
```
Twey, line 685: Even if we don't want to remove the chain actor right away for whatever reason, can't we use a [`tokio::sync::RwLock`](https://docs.rs/tokio/latest/tokio/sync/struct.RwLock.html) internally to do the scheduling here?  (i.e. handle the requests concurrently, with a large-ish buffer e.g. [`FuturesUnordered`](https://docs.rs/futures/latest/futures/stream/struct.FuturesUnordered.html#impl-Stream-for-FuturesUnordered%3CFut%3E), but lock the state in the handler according to the request type)

linera-core/src/chain_worker/actor.rs
```diff
@@ -288,6 +335,53 @@ where
     },
 }
 
+impl<Ctx: Context + Clone + 'static> ChainWorkerRequest<Ctx> {
```
Twey, line 338: `Context` implies `Clone`.

linera-core/src/chain_worker/mod.rs
```diff
@@ -3,18 +3,32 @@
...
+///
+/// On native targets, futures must be `Send` because they are spawned on multi-threaded
+/// runtimes. On wasm, `Send` is not required (single-threaded) and some storage contexts
+/// do not implement `Sync`, making `&self` futures non-`Send`.
+#[cfg(not(web))]
+type BoxedFuture<'a> = Pin<Box<dyn Future<Output = ()> + Send + 'a>>;
+
+/// See non-web variant for documentation.
+#[cfg(web)]
+type BoxedFuture<'a> = Pin<Box<dyn Future<Output = ()> + 'a>>;
```
Twey, line 23: I've been working to try to remove this hack :)  See https://twey.io/rust/boxed-futures/ for some alternative patterns.

linera-core/src/chain_worker/actor.rs
deuszx: If we:
1. Use `RwLock`
2. Priotize writes first
3. The rest that is left are reads that can use the same state concurrently 🤔

afck: Right, because `RwLock` also prioritizes writes… I'll give it a try.

afck: One potential downside is that if we remove the actor, we won't be able to e.g. batch cross-chain messages (see https://github.com/linera-io/linera-protocol/pull/5024) anymore.

Twey: (Not suggesting removing the actor here, but) we can still have a message queue that supports batching without the unstructured concurrency and error boundary that an actor implies. `async` is sufficient to give a call/return API to the caller that may nonetheless be executed out of order.

afck: Let's still see how it turns out if we remove the actor. (I'll put it in a separate draft PR.)

You're right, `handle_cross_chain_request` could just:
* put the request into a queue, and then
* try to acquire the write lock and once it has it process, as a batch, all the requests from the queue.

linera-core/src/chain_worker/actor.rs
```diff
@@ -496,28 +619,101 @@ where
...
+
+            // Process the first request.
             Box::pin(worker.handle_request(request))
                 .instrument(span)
                 .await;
 
+            // Inner loop: write-priority scheduling.
             loop {
+                // Phase 1: Drain all pending writes.
+                while let Ok(item) = receivers.write_receiver.try_recv() {
```
ndr-ds, line 639: Couldn't this starve the reads in a high write throughput scenario, for example? If we go with an `RWLock` like @Twey suggested, they use a FIFO queue internally AFAIU so the starvation is less of an issue

Twey: Right, writing fair mutexes is hard.

afck: Working on it: https://github.com/linera-io/linera-protocol/pull/5502

ma2bd:
Converting to draft as we're working on removing actors altogether: https://github.com/linera-io/linera-protocol/pull/5502

afck:
Closing in favor of #5502.

PR #5487, opened 2026-02-19
Add Gungraun instruction-count benchmarks for linera-views

ndr-ds:
## Motivation

Add deterministic instruction-count benchmarks to detect performance regressions in `linera-views` on every PR. Unlike wall-clock benchmarks, [Gungraun](https://github.com/gungraun/gungraun) uses Valgrind's Callgrind to count CPU instructions — results are deterministic (~0% noise) and require only a single run per benchmark.

## Proposal

21 instruction-count benchmarks covering all major View types:

| View Type | Operations |
| --- | --- |
| MapView | insert, get (hit + miss), indices, contains_key, remove, pre_save |
| QueueView | push_back, front, delete_front, pre_save |
| BucketQueueView | push_back, front, delete_front, pre_save |
| RegisterView | get/set, pre_save |
| CollectionView | load_all_entries (in-memory + from storage), indices |
| Cold Load | MapView::load from storage |

All benchmarks operate on 1000-element views using `MemoryContext`. A CI workflow runs on PRs that touch `linera-views/` or `linera-views-derive/`, compares against the `main` branch baseline, and posts a PR comment with instruction count deltas.

Regression threshold is 1%, with red/green coloring for regressions/improvements.

## Test Plan

- Verified A/A comparison shows "No change" across all 21 benchmarks on deterministic metrics
- Introduced a deliberate regression (unnecessary `.clone()` calls in `MapView::insert`) and confirmed CI correctly detected ~20% instruction count increase

## Release Plan


ndr-ds:
* **#5487** <a href="https://app.graphite.com/github/pr/linera-io/linera-protocol/5487?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/linera-io/linera-protocol/5487?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ndr-ds:
Test comment — will delete shortly.

<details>
<summary>Test table with LaTeX colors</summary>

| Benchmark | Instructions |
|:---|---:|
| `insert_100` | 257,831 (${\color{red}\textbf{+20.5\%}}$) |
| `insert_1000` | 2,955,361 (${\color{green}\textbf{-8.3\%}}$) |
| `get_existing` | 39,320 (+2.1%) |
| `get_missing` | 40,202 (No change) |

</details>

> Regression threshold: **5%** — ${\color{red}\textbf{red}}$ = regression, ${\color{green}\textbf{green}}$ = improvement.

.github/scripts/format-bench-results.py
```diff
@@ -0,0 +1,219 @@
+#!/usr/bin/env python3
```
MathieuDutSik: We currently do not have Python in the codebase.
Instead, we have some bash. It may not be adeqiate and so this needs to be rewritten in rust.

linera-views/benches/gungraun_views.rs
```diff
@@ -0,0 +1,499 @@
+// Copyright (c) Zefchain Labs, Inc.
```
MathieuDutSik, line 1: I find it inadequate to put the instruction count in benchmarks.

Instruction counts are NOT benchmarks. They are related to it, but they are not the same since a core can have many arithmetic units, and we could get better speed with more instructions that use the arithmetic units in a better way.

That does not mean that this metric is useless. It is very stable and deterministic with the same code and compilers.

.github/scripts/format-bench-results.py
ndr-ds: I disagree.  
Python is the right tool for a small CI helper that reads JSON and formats markdown tables. Bash would be way too messy. Rust would be extremely overkill. IMHO, of course :)

linera-views/benches/gungraun_views.rs
ndr-ds: Instruction counts are universally called "benchmarks" in the Rust ecosystem:

- **rustc-perf** (official Rust compiler perf suite) uses instruction counts as its default metric and calls them "benchmarks": https://perf.rust-lang.org/
- **iai-callgrind/gungraun** describes itself as a "benchmarking framework/harness" in its own docs: https://github.com/iai-callgrind/iai-callgrind
- **Criterion's docs** list iai as a "benchmarking alternative": https://bheisler.github.io/criterion.rs/book/iai/iai.html
- **Bencher** categorizes iai alongside Criterion and libtest as "three popular options for benchmarking in Rust": https://bencher.dev/docs/explanation/adapters/
- **cargo bench** treats them via `[[bench]]` sections — the standard benchmark mechanism

The distinction you may be thinking of is that instruction counts measure _work done_ (deterministically) rather than _wall-clock time_ (noisily). Both are benchmarks — they just measure different things. The deterministic property is precisely why instruction counts are preferred for CI regression detection.

ndr-ds:
Ok, script is in bash now

linera-views/benches/gungraun_views.rs
MathieuDutSik: To me, it is more a metric than a benchmark, but you probably do not have other option than to put it in `benches` directory.

ndr-ds:
Total R+W is `Ir + Dr + Dw`: instruction fetches + data reads + data writes. These are Cachegrind's event types defined in [Section 5.2.11 — Cache and Branch Simulation](https://valgrind.org/docs/manual/cg-manual.html#cg-manual.cache-branch-sim).

Instructions is less than Total R+W because like I said above it also has data read and data writes.

Cold Load is just loading a view from storage for the first time.

LL could be renamed to LLC, means Last Level cache. Cachegrind simulates two levels: L1 and LL.  
LL Hits = requests that missed L1 but hit LL.

**I'll** **add** **the** **cache** **sizes** **and** **fix** **CI** **in** **a** **bit**

ndr-ds:
I manually triggered a base save (because we didn't merge yet) a while ago, and that's why it's already catching some potential regression from the old base to now. I can reset the base, but this is good to show us that the warning works :)  
Also, once this is merged the base will get reset to main

PR #5491, opened 2026-02-23
Better error types on the Web (#5483)

Twey:
## Motivation

We'd like to be able to handle different errors differently in the frontends.

## Proposal

Add different error classes for error cases we might want to distinguish (currently just `LockError`, but open for extension) in `error/index.ts`.

Then, instead of using `wasm_bindgen::JsError` everywhere (which becomes an `Error` on the JavaScript side) add our own `Error` type with a custom conversion to `JsValue` that constructs the appropriate error class depending on the enum variant.

It would be nice if `tsify` did this for us.

## Test Plan

Tested locally.

## Release Plan

    - be released in a new SDK,

## Links

### DIFF
--- a/web/@linera/client/src/wallet.rs
+++ b/web/@linera/client/src/wallet.rs
@@ -9,8 +9,7 @@ use linera_core::wallet;
 use wasm_bindgen::prelude::*;
 use web_sys::wasm_bindgen;
 
-use super::JsResult;
-use crate::lock::Lock;
+use crate::{lock::Lock, Error, Result};
 
 /// A wallet that stores the user's chains and keys in memory.
 #[wasm_bindgen]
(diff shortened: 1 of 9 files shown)


web/@linera/client/src/wallet.rs
```diff
@@ -30,14 +29,12 @@ impl Wallet {
...
-    pub async fn set_owner(&self, chain_id: JsValue, owner: JsValue) -> JsResult<()> {
+    pub async fn set_owner(&self, chain_id: JsValue, owner: JsValue) -> Result<()> {
         let chain_id = serde_wasm_bindgen::from_value(chain_id)?;
         let owner = serde_wasm_bindgen::from_value(owner)?;
         self.chains
             .mutate(chain_id, |chain| chain.owner = Some(owner))
-            .ok_or(JsError::new(&format!(
-                "chain {chain_id} doesn't exist in wallet"
-            )))
+            .ok_or(Error::new(&format!("chain {chain_id} doesn't exist in wallet")).into())
```
ndr-ds, line 37: Is the `.into()` necessary here?

Twey: No, and because `RUSTFLAGS` aren't additive CI didn't fail on it :upside_down_face: It'll be fixed in https://github.com/linera-io/linera-protocol/pull/5497.

PR #5494, opened 2026-02-23
Remove some locks held over await statements.

MathieuDutSik:
## Motivation

Locking over an await statement is generally known as not great since it introduces latency and can block the system.

## Proposal

The locks over await cases were identified by CODEX 5.3, with Claude Opus 4.6 used for
checking correctness. The changes were done atomically, with each commit corresponding
to one issue addressed.

## Test Plan

CI.

## Release Plan

Possible backport to testnet_conway.

## Links

None.

### DIFF
--- a/linera-client/src/chain_listener.rs
+++ b/linera-client/src/chain_listener.rs
@@ -22,7 +22,7 @@ use linera_base::{
 use linera_core::{
     client::{
         chain_client::{self, ChainClient},
-        AbortOnDrop, Client, ListeningMode,
+        AbortOnDrop, ListeningMode,
     },
     node::NotificationStream,
     worker::{Notification, Reason},
@@ -225,21 +225,14 @@ pub struct ChainListener<C: ClientContext> {
 }
 
 impl<C: ClientContext + 'static> ChainListener<C> {
-    async fn shared_client_and_timing(
-        context: &Arc<Mutex<C>>,
-    ) -> (
-        Arc<Client<C::Environment>>,
-        Option<tokio::sync::mpsc::UnboundedSender<(u64, linera_core::client::TimingType)>>,
-    ) {
-        let guard = context.lock().await;
-        (guard.client().clone(), guard.timing_sender())
-    }
-
     async fn make_chain_client_from_context(
         context: &Arc<Mutex<C>>,
         chain_id: ChainId,
     ) -> Result<ContextChainClient<C>, Error> {
-        let (client, timing_sender) = Self::shared_client_and_timing(context).await;
+        let (client, timing_sender) = {
+            let guard = context.lock().await;
+            (guard.client().clone(), guard.timing_sender())
+        };
         let chain = client
             .wallet()
             .get(chain_id)
@@ -284,7 +277,10 @@ impl<C: ClientContext + 'static> ChainListener<C> {
     #[instrument(skip(self))]
     pub async fn run(mut self) -> Result<impl Future<Output = Result<(), Error>>, Error> {
         let chain_ids = {
-            let (client, _timing_sender) = Self::shared_client_and_timing(&self.context).await;
+            let client = {
+                let guard = self.context.lock().await;
+                guard.client().clone()
+            };
             let admin_chain_id = client.admin_chain_id();
             Self::make_chain_client_from_context(&self.context, admin_chain_id)
                 .await?
--- a/linera-client/src/chain_listener.rs
+++ b/linera-client/src/chain_listener.rs
@@ -645,8 +645,7 @@ impl<C: ClientContext + 'static> ChainListener<C> {
                                 };
                                 self.remove_event_subscriber(chain_id);
                                 listening_client.stop().await;
-                                let (client, _) = Self::shared_client_and_timing(&self.context).await;
-                                if let Err(error) = client.wallet().remove(chain_id).await {
+                                if let Err(error) = self.context.lock().await.wallet().remove(chain_id).await {
                                     error!(%error, %chain_id, "error removing a chain from the wallet");
                                 }
                             }
--- a/linera-client/src/chain_listener.rs
+++ b/linera-client/src/chain_listener.rs
@@ -22,6 +22,7 @@ use linera_base::{
 use linera_core::{
     client::{
         chain_client::{self, ChainClient},
+        Client,
         AbortOnDrop, ListeningMode,
     },
     node::NotificationStream,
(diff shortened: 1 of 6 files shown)


linera-service/src/cli_wrappers/local_kubernetes_net.rs
```diff
@@ -221,9 +221,13 @@ impl LineraNet for LocalKubernetesNet {
     }
 
     async fn terminate(&mut self) -> Result<()> {
-        let mut kubectl_instance = self.kubectl_instance.lock().await;
+        let mut port_forward_children = {
+            let mut kubectl_instance = self.kubectl_instance.lock().await;
```
afck, line 225: If we change all these mutexes and R/W locks so that their guards are never held across an `await` point, should we replace them by the synchronous standard library ones then rather than the tokio ones?

linera-client/src/chain_listener.rs
```diff
@@ -609,7 +637,8 @@ impl<C: ClientContext + 'static> ChainListener<C> {
                                     continue;
                                 };
                                 listening_client.stop().await;
-                                if let Err(error) = self.context.lock().await.wallet().remove(chain_id).await {
+                                let (client, _) = Self::shared_client_and_timing(&self.context).await;
+                                if let Err(error) = client.wallet().remove(chain_id).await {
```
afck: Is it really necessary to create a new _client_ here? We just need the wallet, after all.

MathieuDutSik: You are right. Let's revert that change.
I look at the code, and I do not see why "fn remove" should be async in the Wallet trait.

linera-service/src/cli_wrappers/local_kubernetes_net.rs
MathieuDutSik: Yes, I was thinking about this.

linera-client/src/chain_listener.rs
Twey: In general, all functions on the `Wallet` trait must be async as they update the underlying storage, which may be async (e.g. in the case of IndexedDB).

linera-service/src/cli_wrappers/local_kubernetes_net.rs
MathieuDutSik: I think yes, we can do this change and do a systematic elimination of the tokio based locks and mutexes with standard library ones.

However, that feels to me to be something for the next PR.

linera-client/src/chain_listener.rs
MathieuDutSik: Thanks for explaining. Indeed, we want to change the `web/@linera/client/src/wallet.rs` in the end to support the wallet with the web storage and that would be async.

linera-service/src/cli/main.rs
```diff
@@ -1264,14 +1264,21 @@ impl Runnable for Job {
...
+                    Some(context.make_chain_client(chain_id).await?)
+                } else {
+                    None
+                };
+
                 let context = Arc::new(Mutex::new(context));
 
                 let (command_sender, command_receiver) = mpsc::unbounded_channel();
 
                 if let Some(controller_id) = controller_application_id {
```
afck, line 1271: (We could merge that with the `if` above. The channel can probably be created before, and the `Arc::new(Mutex::new(context))` duplicated. Personally I'd prefer duplicating one line over an `expect`. Not a blocker, though!)

linera-client/src/chain_listener.rs
afck: Either way it still shouldn't be necessary to create a new client just to access the wallet, should it?

MathieuDutSik: Yes, forgot to push the change yesterday.

linera-client/src/chain_listener.rs
```diff
@@ -225,6 +225,40 @@ pub struct ChainListener<C: ClientContext> {
 }
 
 impl<C: ClientContext + 'static> ChainListener<C> {
+    async fn shared_client_and_timing(
```
deuszx: Not convinced this is necessary. It's actially really used in one place only (`make_chain_client_from_context`). In `fn run` you're ignoring the `_timing_sender` anyway.

linera-service/src/cli_wrappers/local_kubernetes_net.rs
deuszx: I think `parking_lot` mutexes are faster

linera-service/src/cli/main.rs
MathieuDutSik: Ok, reverted that change.

linera-client/src/chain_listener.rs
MathieuDutSik: Ok, removed that function and instead put the code directly in place.

linera-service/src/cli_wrappers/local_kubernetes_net.rs
MathieuDutSik: Ok, but for next PR.

linera-service/src/node_service.rs
```diff
@@ -128,6 +131,40 @@ impl<C> MutationRoot<C>
...
+    async fn shared_client(&self) -> Arc<Client<C::Environment>> {
+        let context = self.context.lock().await;
+        context.client().clone()
+    }
+
+    async fn persist_wallet_from_client(
+        &self,
+        chain_client: &ChainClient<C::Environment>,
+    ) -> Result<(), Error> {
+        let shared_client = self.shared_client().await;
```
afck: Why do we need this new client at all? Doesn't `chain_client` have a handle to the same wallet already?

linera-service/src/cli_wrappers/local_kubernetes_net.rs
ma2bd: The difference is probably not that significant for us.

linera-service/src/node_service.rs
MathieuDutSik: Yes, reverted. We want to avoid concurrent operations on the wallet.

linera-client/src/chain_listener.rs
```diff
@@ -225,6 +225,33 @@ pub struct ChainListener<C: ClientContext> {
...
+        Ok(client.create_chain_client(
+            chain_id,
+            chain.block_hash,
+            chain.next_block_height,
+            chain.pending_proposal,
+            chain.owner,
+            timing_sender,
+            follow_only,
+        ))
+    }
```
afck, line 253: I'm not sure about this: We're basically reimplementing `make_chain_client`. Possible alternatives:
* Maybe the context should not be in a `Mutex` but in an `RwLock`, so we can just take a _read_ lock and call `make_chain_client`.
* Or maybe the client context should be clonable and have an `RwLock` _inside_, so this is all automatic?

MathieuDutSik:
It does not seem to have been a good idea.

PR #5495, opened 2026-02-23
EVM contract detects linera fungible `Credit` operations

deuszx:
## Motivation

Bridge Fungible tokens from Linera to EVM

## Proposal

This PR introduces the following:
1. `FungibleBridge.sol` contract – extends `Microchain.sol`. Upon reception of the new block it:
  - calls its `LightClient.sol` instance (from constructor) to verify the block finalization
  - scans the block's transaction list in search for `FungibleMessage::Credit` operation for the corresponding applicationId
  - if successful: calls the ERC20 token and transfers to the recipient on EVM
2. Generates Solidity (de)serialization and types for the Linera fungible contract.

## Test Plan

CI unit test.

e2e tests upcoming.

## Release Plan

None yet.

## Links

### DIFF
--- a/linera-bridge/README.md
+++ b/linera-bridge/README.md
@@ -22,8 +22,8 @@ The bridge has three layers:
           ┌──────────────┼──────────────┐
           │              │              │
    ┌──────▼──────┐ ┌─────▼──────┐ ┌────▼───────┐
-   │ Microchain  │ │ Microchain │ │ Microchain │  one per chain,
-   │  (USDC)     │ │  (NFT)     │ │  (DEX)     │  enforces chain_id + height
+   │ Fungible-   │ │ Microchain │ │ Microchain │  one per chain,
+   │  Bridge     │ │  (NFT)     │ │  (DEX)     │  enforces chain_id + height
    └─────────────┘ └────────────┘ └────────────┘
 ```
 
@@ -39,7 +39,22 @@ YAML snapshot (tests/snapshots/format__format.yaml.snap)
 BridgeTypes.sol (src/solidity/BridgeTypes.sol)
 ```
 
-The snapshot is checked in and tested via `insta`. This means the generated Solidity stays in sync with the Rust types — if a struct field is added or an enum variant reordered, the snapshot test fails and the developer must update it explicitly.
+Application-specific types follow the same pipeline. For example, `FungibleOperation` from the fungible token application:
+
+```
+Rust types (linera-sdk::abis::fungible)
+    │
+    ▼  serde-reflection (tests/format_fungible.rs)
+YAML snapshot (tests/snapshots/format_fungible__format_fungible.yaml.snap)
+    │
+    ▼  serde-generate via build.rs (shared types declared as external_definitions)
+FungibleTypes.sol (src/solidity/FungibleTypes.sol)
+    imports BridgeTypes.sol for shared types
+```
+
+`FungibleTypes.sol` only contains types and deserializers unique to the fungible application (`FungibleOperation` and its variants). Shared types like `Account`, `AccountOwner`, and `Amount` are reused from `BridgeTypes.sol` via import — `build.rs` passes them as `external_definitions` to `serde-generate`, which emits qualified `BridgeTypes.` references and `import` statements instead of duplicate definitions.
+
+All snapshots are checked in and tested via `insta`. This means the generated Solidity stays in sync with the Rust types — if a struct field is added or an enum variant reordered, the snapshot test fails and the developer must update it explicitly.
 
 ## Certificate verification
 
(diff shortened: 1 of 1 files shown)


linera-bridge/src/solidity/FungibleBridge.sol
```diff
@@ -0,0 +1,59 @@
...
+        bytes32 _applicationId,
+        address _token
+    )
+        Microchain(_lightClient, _chainId, _latestHeight)
+    {
+        applicationId = _applicationId;
+        token = IERC20(_token);
+    }
+
+    function _onBlock(BridgeTypes.Block memory blockValue) internal override {
```
afck, line 32: (At this point we already know that the block is on the correct chain, do we? And this can be called only once per block?)

linera-bridge/src/test_helpers.rs
```diff
@@ -152,6 +156,92 @@ pub fn deploy_microchain(
...
+    initial_supply: alloy_primitives::U256,
+) -> Address {
+    let bytecode = compile_contract(MOCK_ERC20_SOL, "MockERC20.sol", "MockERC20");
+    let constructor_args = (initial_supply,).abi_encode_params();
+    let mut deploy_data = bytecode;
+    deploy_data.extend_from_slice(&constructor_args);
+    deploy_contract(db, deployer, deploy_data)
+}
+
+/// Creates a Transaction::ReceiveMessages containing a fungible Message as a user message.
```
afck, line 216:
```suggestion
/// Creates a `Transaction::ReceiveMessages` containing a fungible `Message` as a user message.
```

linera-bridge/src/lib.rs
```diff
@@ -4,9 +4,17 @@
 /// Solidity source for the BridgeTypes library (generated).
 pub const BRIDGE_TYPES_SOURCE: &str = include_str!("solidity/BridgeTypes.sol");
 
+/// Solidity source for the FungibleTypes library (generated).
```
afck, line 7:
```suggestion
/// Solidity source for the `FungibleTypes` library (generated).
```

linera-bridge/src/lib.rs
```diff
@@ -4,9 +4,17 @@
...
+pub const FUNGIBLE_TYPES_SOURCE: &str = include_str!("solidity/FungibleTypes.sol");
+
+/// Solidity source for the FungibleBridge contract.
```
afck, line 10:
```suggestion
/// Solidity source for the `FungibleBridge` contract.
```

linera-bridge/src/fungible_bridge.rs
```diff
@@ -0,0 +1,278 @@
...
+    mod erc20 {
+        use alloy_sol_types::sol;
+
+        sol! {
+            function balanceOf(address account) external view returns (uint256);
+            function transfer(address to, uint256 amount) external returns (bool);
+        }
+    }
+
+    /// Common test state for FungibleBridge tests.
```
afck, line 34:
```suggestion
    /// Common test state for `FungibleBridge` tests.
```

linera-bridge/src/fungible_bridge.rs
```diff
@@ -0,0 +1,278 @@
...
+            t.query_token_balance(test_target_evm_address()),
+            alloy_primitives::U256::ZERO,
+            "target should start with zero balance"
+        );
+
+        t.submit_credit(test_target(), Amount::from_tokens(100), test_source());
+
+        assert_eq!(
+            t.query_token_balance(test_target_evm_address()),
+            alloy_primitives::U256::from(100_000_000_000_000_000_000u128),
```
afck, line 202: Maybe `let amount = Amount::from_tokens(100)` above and then use `amount.to_attos()` here. Fewer zeros to count. 😅

linera-bridge/src/fungible_bridge.rs
```diff
@@ -0,0 +1,278 @@
...
+            "target should have 100 tokens"
+        );
+    }
+
+    #[test]
+    fn test_fungible_bridge_accumulates_transfers() {
```
afck, line 208: (This really makes the previous test redundant, doesn't it?)

linera-bridge/build.rs
```diff
@@ -25,6 +26,51 @@ fn generate_bridge_types() {
         .expect("failed to generate Solidity code");
 }
 
+/// Generates FungibleTypes.sol from the fungible snapshot.
+/// Primitive types shared with BridgeTypes are declared as external so the generated
```
afck, line 30:
```suggestion
/// Primitive types shared with `BridgeTypes` are declared as external so the generated
```

linera-bridge/src/solidity/FungibleBridge.sol
deuszx: Yes. This is internal method. The only public method in `FungibleBridge` is actually `addBlock` from `Microchain.sol`. `addBlock` calls `_onBlock`.

linera-bridge/src/fungible_bridge.rs
deuszx: This is going to be addressed in another PR that I am preparing. I'm gonna fix it there.

PR #5497, opened 2026-02-23
Web: lock on `claimChain` not `new Client`

Twey:
## Motivation

We'd like to be able to fall back to trying another wallet if the wallet is not available.  While this is technically possible with the lock happening in `new Client`, it is rather cumbersome to have to create the whole client before we find out whether we need to fall back.

## Proposal

Try to lock early, in `Faucet.claimChain`, and error on failure.  If no chain is claimed, the default wallet is still locked when we construct the client.

## Test Plan

Playwright test added.

## Release Plan

    - be released in a new SDK,

## Links


Twey:
> What's the usecase for this?

Mostly, your MetaMask fungible demo.  But it's also just semantically more correct that the `Wallet` should hold the wallet lock than that the `Client` should, i.e. it fixes some rough edges on https://github.com/linera-io/linera-protocol/pull/5483.

web/@linera/client/src/wallet.rs
```diff
@@ -47,17 +61,22 @@ impl Wallet {
 
     /// Lock the wallet, preventing anyone else from using a wallet with this name.
     ///
+    /// If the wallet is already locked, this is a no-op.
```
deuszx, line 64: How does this work? Can I have two clients that both hold the lock on the same wallet? This function won't fail on subsequent `lock` calls so the caller is tricked into thinking that it's locking the wallet for itself 🤔

deuszx: ping @Twey

Twey: Perhaps this comment isn't clear enough, but no, if you have two clients that try to lock the same wallet, the second one will fail, as it has always been.  If _this_ page already holds the lock, i.e. the caller already has the lock, the locking operation will be a no-op.

Issue #3299, opened 2025-02-12
Test if `ChainWorkerActor` gracefully handles storage errors when starting


jvff:
# Motivation

PR #3298 fixes a bug caused by an early-return while initializing the chain actor task. A test that reproduces the issue should be written to ensure that the bug is fixed.

# Proposal

Create a `MockStorage` type that allows injecting faults when loading a chain, and use it check that the chain worker actor task continues running in the presence of initial storage errors.

Issue #2447, opened 2024-09-04
Remove `ChainWorkerActor`


jvff:
# Motivation

Keeping track of the actor's lifetime has led to a few bugs (#2444, #2445), and the main goal of having separate tasks per chain worker can also be achieved by properly splitting up tasks outside `linera-core`.

# Proposal

Keep a map of `ChainWorkerState`s in `WorkerState`.

PR #5502, opened 2026-02-24
Remove chain actors; handle read-only calls concurrently.

### LINKED ISSUE #5536: Remove chain actors
Motivation:
* R/W fairness using a standard `rwlock` implementation
* canceled queries will cancel the work
* overall simplify the code before we add complexity again in a number of ways (pooled cross-chain message saves, chain-state migration)

afck:
## Motivation

The chain actors are complicated and unnecessary, and even read-only requests to them are unnecessarily run only sequentially.

## Proposal

Remove the chain actors, use an `RwLock` instead.

## Test Plan

CI should catch regressions. We should do benchmarks to see if this improves performance.

## Release Plan

- Release a new SDK.

## Links

- Closes #5536, #2447, #2237, #3299

### DIFF
--- a/linera-core/src/chain_worker/handle.rs
+++ b/linera-core/src/chain_worker/handle.rs
@@ -2,6 +2,10 @@
 // SPDX-License-Identifier: Apache-2.0
 
 //! Helpers for managing chain worker state behind an `Arc<RwLock<…>>`.
+//!
+//! Tokio's [`RwLock`] is write-preferring: once a writer is waiting, new readers
+//! queue behind it. This prevents read-only queries from starving write operations
+//! (block proposals, certificate handling, etc.).
 
 use std::{
     ops::{Deref, DerefMut},


linera-core/src/worker.rs
```diff
@@ -33,43 +32,92 @@ use linera_chain::{
...
 
 #[cfg(test)]
 #[path = "unit_tests/worker_tests.rs"]
 mod worker_tests;
 
+/// Acquires a read lock on a chain handle, executes `$body` with the guard named `$guard`,
+/// and returns the result. The entire operation (including chain loading) runs in a spawned
+/// task to isolate the chain state from the caller's future, ensuring `Sync` compatibility
+/// and small future size.
+macro_rules! spawn_chain_read {
```
ma2bd: We used to cancel chain tasks from the RPC handler. I would support doing it again. By now storage has got to be crash safe.

linera-core/src/updater.rs
```diff
@@ -187,9 +187,8 @@ where
                 };
                 let entry = error_scores.entry(err.clone()).or_insert(0);
                 *entry += committee.weight(&name);
-                if *entry >= committee.validity_threshold() {
```
ma2bd: the validity threshold should be `f+1` not `f`

afck: It is!

In https://github.com/linera-io/linera-protocol/pull/4978 I changed it so that if we have 3 validators, a quorum is 2. `f` is still 0, of course, so the validity threshold is 1.

The point here is: We don't necessarily need to stop just because we know that at least one honest validator returned an error (e.g. is missing a blob). We should only stop once it is not possible anymore to reach a quorum.

In the 3 validators case, just because `1 = f + 1` validators returned an error, we should not return. We should wait for at least one more, otherwise we might still get a quorum.

I created https://github.com/linera-io/linera-protocol/pull/5521 to discuss this separately.

linera-core/src/chain_worker/config.rs
```diff
@@ -56,6 +63,7 @@ impl ChainWorkerConfig {
 impl Default for ChainWorkerConfig {
     fn default() -> Self {
         Self {
+            nickname: String::new(),
```
Twey, line 59: This might be a good moment to get rid of this (it really belongs in the `tracing` context).

linera-core/src/chain_worker/config.rs
```diff
@@ -64,6 +72,8 @@ impl Default for ChainWorkerConfig {
             ttl: Default::default(),
             sender_chain_ttl: Duration::from_secs(1),
             chain_info_max_received_log_entries: CHAIN_INFO_MAX_RECEIVED_LOG_ENTRIES,
+            block_cache_size: 5000,
+            execution_state_cache_size: 10_000,
```
Twey, line 69: Where did these come from previously?

linera-core/src/chain_worker/config.rs
afck: I think that's another quite lengthy change that I'd rather not add to this PR.

linera-core/src/chain_worker/config.rs
afck: They were just passed explicitly into `ClientContext::new`, `Client::new`, and `WorkerState::new`. They did (and still do) come from a CLI option, though.

(Edit: Except in tests, where these defaults are used.)

linera-core/src/chain_worker/handle.rs
```diff
@@ -0,0 +1,191 @@
...
+use linera_storage::Storage;
+use tokio::sync::{OwnedRwLockReadGuard, RwLock};
+
+use super::state::ChainWorkerState;
+
+/// A handle to a chain worker's state, protected by an `RwLock`.
+///
+/// Replaces the former channel-based `ChainWorkerActor`. Concurrent reads acquire the
+/// read lock; mutations acquire the write lock. Tokio's `RwLock` is write-preferring,
+/// which guarantees that mutations don't starve if there are many reads.
```
ma2bd: We don't need the code history in the comments.

linera-core/src/chain_worker/handle.rs
```diff
@@ -0,0 +1,191 @@
...
+
+    /// Updates the last-access timestamp.
+    fn touch(&self) {
+        self.last_access
+            .store(current_time_micros(), Ordering::Relaxed);
+    }
+
+    /// Checks whether this handle has been idle beyond its TTL.
+    ///
+    /// A zero TTL means "no expiry" (the handle lives forever).
```
ma2bd: This is a little sloppy. Normally, we would use `NonZero<Option<Duration>>`.

linera-core/src/worker.rs
```diff
@@ -465,16 +505,102 @@ impl WorkerError {
...
+    chain_handles: &ChainHandleMap<S>,
+    config: &ChainWorkerConfig,
+) {
+    let ttl = config.ttl;
+    let sender_ttl = config.sender_chain_ttl;
+    // Sweep at the smaller of the two TTLs, divided by 4 for responsiveness.
+    // If both TTLs are zero, don't sweep at all (handles live forever).
+    let min_ttl = ttl.min(sender_ttl);
+    let interval = if min_ttl.is_zero() {
+        if ttl.is_zero() && sender_ttl.is_zero() {
```
ma2bd: see previous comment

linera-core/src/worker.rs
```diff
@@ -465,16 +505,102 @@ impl WorkerError {
     }
 }
 
+type ChainHandleMap<S> = Arc<Mutex<BTreeMap<ChainId, Arc<ChainHandle<S>>>>>;
```
ma2bd: Should we use a `papaya` map?

linera-core/src/worker.rs
```diff
@@ -465,16 +505,102 @@ impl WorkerError {
...
+                // Acquire write lock — blocks until all outstanding
+                // OwnedRwLock{Read,Write}Guards are released. While we wait,
+                // the handle stays in the map so new requests reuse it.
+                let mut guard = handle.write().await;
+                // Re-lock the map. We hold the write lock, so no one can
+                // acquire new guards. If strong_count == 2 (map + us), no
+                // request is currently referencing this handle.
+                {
+                    let mut handles = chain_handles.lock().unwrap();
+                    if Arc::strong_count(&handle) != 2 || !handle.is_expired(ttl, sender_ttl) {
```
ma2bd: This looks correct but instead of introspecting the `Arc` counters, I wonder if it would be more obviously correct to use `Arc<Mutex<papaya::Map<ChainId, Entry>>>>` with `struct Entry { handle: Weak<ChainHandle<S>, timer: AbortHandle}` and `timer` holding an `Arc` for a duration TTL? (Here `timer` is abortable in case we want to replace it to extend the lifetime. Alternatively, `timer` could be made reusable by receiving `Touch` events on a channel.) Then, this would only require an occasional sweep to remove (small) entries with non-upgradable weak pointers.

linera-core/src/chain_worker/handle.rs
```diff
@@ -0,0 +1,191 @@
...
+                runtime_request_sender,
+            },
+            task: thread_pool
+                .run((), move |()| async move {
+                    ServiceSyncRuntime::new(
+                        execution_state_sender,
+                        QueryContext {
+                            chain_id,
+                            next_block_height: BlockHeight(0),
+                            local_time: Timestamp::from(0),
```
ma2bd, line 87: Wait, is this correct?

ma2bd: If we need to unsupport long-lived services to make this work, I'm ok with that, but perhaps in a separate PR before this one.

linera-core/src/chain_worker/handle.rs
afck: Unfortunately `Duration` is not a `ZeroablePrimitive` for some reason, so I'll use an `Option<Duration>`.

I'll only change it internally, though. If we want to change it on the CLI that needs some discussion first.

linera-core/src/chain_worker/handle.rs
afck: This looks indeed weird! But we were already doing that before this PR: The code just moved. I'll look into it…

afck: It is correct: It's overwritten when a query is made. I added a comment.

linera-core/src/worker.rs
afck: Done. Instead of replacing the task I just update an `AtomicU64` and the task checks it for the last-used timestamp.

linera-core/src/worker.rs
afck: Done.

MathieuDutSik (approved):
This PR addresses issue #2447, which should be mentioned. Also relevant appears to be issue #3299, #5536, #2237. Are all of those addressed?

On its face, having the behavior of a `RwLock` for the chain worker is a significant step forward. But I am concerned about something else:
* We can have concurrent read queries on a chain, which means I think that we can ensure that a chain always has some read queries on it.
* So, that chain, would not be able to process any write queries to it.
* Does that give a way to do a Denial of Service on a chain?
I think the `tokio::sync::RwLock` is designed for avoiding that scenario, but it would be nice to have it written down somewhere.

Otherwise, I find it interesting that `ChainWorkerActor` disappeared, but `ChainWorkerConfig` continues to exist.

ma2bd:
> Otherwise, I find it interesting that `ChainWorkerActor` disappeared, but `ChainWorkerConfig` continues to exist.

We still have `ChainWorkerState` and the general concept of chain worker. It's just not implemented with an actor.

afck:
> Does that give a way to do a Denial of Service on a chain?

As you say, `RwLock` ensures readers cannot starve writers: https://docs.rs/tokio/latest/tokio/sync/struct.RwLock.html
I'll add it to a comment in the code.

linera-core/src/chain_worker/handle.rs
```diff
@@ -0,0 +1,203 @@
...
+                break;
+            }
+            // Touched recently — sleep for the remaining time.
+            let remaining = Duration::from_micros(ttl_micros - idle_micros);
+            linera_base::time::timer::sleep(remaining).await;
+        }
+        // Shut down: acquire write lock to drain outstanding guards, then
+        // clear the endpoint and await the runtime task.
+        let task = {
+            let mut guard = RollbackGuard(state.clone().write_owned().await);
```
ma2bd, line 184: Should we check the `last_access` here again?

afck: We could; it would prevent dropping the state in some cases. ~I'll make a follow-up.~

PR #5506, opened 2026-02-24
Backport: Adjust service chain ownership when reassigning services between workers (#5435)

bart-linera:
## Motivation

If a service is reassign from worker A to worker B by the controller application, worker A should cease to be the owner of the service's chain, and worker B should become an owner.

## Proposal

Upon reassignment, when telling worker A to stop running the service, worker A will add worker B as an owner and notify the main controller chain.

The main controller will then tell worker B to take over, telling it at which height on the service chain it became an owner. Worker B then waits until it synchronizes the service chain until that height, and when it's ready, it starts listening to the chain in FullChain mode and removes worker A as an owner.

## Test Plan

`test_controller` was extended to reassign a service from one worker to another, and then check if the second worker properly handles a task request.

## Release Plan

- These changes should be released in a new SDK

## Links

- Backported PR: #5435 

### DIFF
--- a/examples/controller/tests/snapshots/format__format.snap
+++ b/examples/controller/tests/snapshots/format__format.snap
@@ -18,6 +18,8 @@ REGISTRY:
             TUPLEARRAY:
               CONTENT: U8
               SIZE: 20
+  BlockHeight:
+    NEWTYPESTRUCT: U64
   ChainId:
     NEWTYPESTRUCT:
       TYPENAME: CryptoHash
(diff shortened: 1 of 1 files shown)


linera-sdk/src/abis/controller.rs
afck, line 155: Should the message enum actually be part of the public controller ABI at all?

bart-linera: It's not on `main`, but it got moved to the public ABI on `testnet_conway` - I don't know why, but I expect there was a good reason (maybe related to the PM app?).

PR #5518, opened 2026-02-25
Backport: Make the controller set appropriate message policies on service chain listeners (#5484)

bart-linera:
## Motivation

Currently, anyone can send any message to any chain managed by the controller, and the workers will automatically receive and process those messages. This creates potential vulnerabilities in the system.

## Proposal

Make workers reject message bundles addressed to service chains that contain no messages for relevant applications and no system messages.

## Test Plan

CI

## Release Plan

## Links


ndr-ds (approved):

afck (approved):

ma2bd:
Merged `testnet_conway` in to clear the conflicts (merge commit, no rebase, so your history is untouched). Six files conflicted after 454 commits of drift. Most were mechanical; **three were not, and need your eye** — I could not run the e2e test here, so treat those as proposals rather than resolutions.

### 1. `apply_policy` — I narrowed your change, and this is the one to check

Your branch routes *every* policy mismatch through the skippable/protected logic. `testnet_conway` has since gained `ignore_chain_ids` and `never_reject_application_ids`, and its `test_message_policy` asserts that a bundle failing `restrict_chain_ids_to` is **dropped and stays pending** — which your version turns into a rejection. Merging your version verbatim failed that test (balance 2, expected 3).

Both intents survive if the split is by *kind* of filter:

- **Chain-level** (`restrict_chain_ids_to`, `ignore_chain_ids`) — drop outright, as on `testnet_conway`.
- **Application-level** (`reject_message_bundles_*`, blanket reject) — your `matches_policy`, so an unskippable bundle is rejected rather than stranded.

That keeps the fix your PR actually needs, since the controller filters service chains with `reject_message_bundles_without_application_ids`, and it is what makes the fungible `Credit` bounce back rather than stick. `test_message_policy` passes unchanged with this split. If you meant the chain-level filters to reject too, that test needs updating instead and the call is yours.

The `never_reject_application_ids` short-circuit sits after the chain-level checks, matching its documented "bypasses the other rejection rules (except `restrict_chain_ids_to`)".

### 2. The e2e test — your scenarios kept, re-expressed as polling

`testnet_conway` replaced the height-based waits with polling loops, deliberately: *"Height-based waits were flaky because the exact block at which the registration lands depends on the starting height."* Your `start_h + N` arithmetic would not survive anyway, since your own PR adds blocks.

So I took the polling structure and re-added your scenarios in that idiom: the fungible app setup, `change_application_permissions` on the service chain, the transfer, the handoff to worker 2, `requestTaskOn`, `taskCount == 2`, and the bounce-back assertion — the last as a poll on `get_amount` rather than a one-shot `assert_balances`, since the bounce is asynchronous. **The synchronisation points are my guess at your intent and are unverified**; I have no way to run this test.

### 3. Two things the auto-merge broke silently

Worth knowing about, because neither showed up as a conflict:

- `ClientWrapper::change_application_permissions`, which your test calls, was **deleted from `testnet_conway` by #5539** — the revert of #5506. I restored it. It is a test-only CLI wrapper, so this does not resurrect any of the reverted behaviour.
- `execute_message` kept your `PendingTask { operator, input }` while `testnet_conway` had added an `id` field, so the example stopped compiling for wasm32. That poisoned the `LazyLock` that builds the example modules and took **15 unrelated `::wasmer` tests** down with it, all reporting only "Once instance has previously been poisoned". Fixed by routing through `add_pending_task`, which allocates the id the same way the operation path does.
- The merged `format__format.snap` had gained a `CryptoHash` registry entry that neither branch had. Regenerated.

### Verification

`linera-chain` 52/52, `linera-core` 187/187 (including `test_message_policy` and all 15 `wasm` tests), `task-processor` example tests pass, `cargo check -p linera-service --tests` clean, `cargo fmt --check` clean under the pinned `nightly-2025-04-03`.

Not verified: the `test_controller` e2e test, which needs a running network.

Issue #5517, opened 2026-02-25
Allow deterministically disposing of the client


Twey:
Now the client takes a lock on its wallet, it is important to allow callers to know when the client (especially the wallet with its lock) has been fully disposed (and the lock released).

PR #5519, opened 2026-02-25
Support client disposal

Twey:
## Motivation

Now we have a lock on the wallet in the Web, it's important to know when that lock has been safely released, i.e. the client is destroyed, so that it can be taken again.

## Proposal

Add cancellation to `linera_base::task::NonBlockingFuture` (renamed simply `linera_base::Task`, as the old name is vestigial to contrast with the blocking tasks that used to live there but have been superseded by `web-thread`).

Make `ChainClient` cancel-safe, i.e. dropping its futures (at least those that hold an `Arc` reference to the client) will dispose of their resources.

Unfortunately, because of the lack of async drop, the same treatment cannot be applied to the `ChainListener`, so the above change to `linera_base` is employed to provide feedback for when the async disposal (which is needed as some use of the `Client` is on another task) has completed.

On the JavaScript side, add an `asyncDispose` function to `linera.Client`.  The name corresponds to the name of the [symbol](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/asyncDispose) in the upcoming [ECMAScript Async Explicit Resource Management](https://tc39.es/proposal-async-explicit-resource-management/#table-1) standard, which we can support or polyfill at a later date.

## Test Plan

A test is added for the new behaviour.

## Release Plan

    - be released in a new SDK,

## Links

- Fixes https://github.com/linera-io/linera-protocol/issues/5517.

## Future work

We don't require any async actions during cleanup, except that its subfutures live in different tasks, so we need to asynchronously wait for them to be cancelled.
If the chain listener's concurrency were structured, therefore (i.e. the state of its subtasks were a subset of its own state) we wouldn't need async disposal, as the resources of its subtasks would be cleaned up on drop.  The cancellation token, `asyncDispose`, et cetera could be removed, and cancellation would be possible just by calling `free` (which `wasm-bindgen` provides to JavaScript).

### DIFF
--- a/linera-core/src/client/chain_client/mod.rs
+++ b/linera-core/src/client/chain_client/mod.rs
@@ -988,8 +988,11 @@ impl<Env: Environment> ChainClient<Env> {
         let mut other_sender_chains = Vec::new();
         let (sender, mut receiver) = mpsc::unbounded_channel::<ChainAndHeight>();
 
-        let cert_futures = received_logs.heights_per_chain().into_iter().filter_map(
-            |(sender_chain_id, remote_heights)| {
+        let cert_futures = received_logs.heights_per_chain().into_iter().filter_map({
+            let received_logs = &received_logs;
+            let other_sender_chains = &mut other_sender_chains;
+
+            move |(sender_chain_id, remote_heights)| {
                 if remote_heights.is_empty() {
                     // Our highest, locally executed block is higher than any block height
                     // from the current batch. Skip this batch, but remember to wait for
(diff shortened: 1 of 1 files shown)


linera-core/src/client/chain_client/mod.rs
```diff
@@ -1028,32 +1017,18 @@ impl<Env: Environment> ChainClient<Env> {
...
-
-        select! {
-            _ = stream::iter(cert_futures)
-            .buffer_unordered(self.options.max_joined_tasks)
-            .for_each(future::ready)
-            => (),
-            _ = cancellation_future => ()
-        };
-
         drop(sender);
```
Twey: Perhaps important to note (fao @bart-linera) that this code will no longer be executed on cancellation.  However since we must support dirty cancellation (e.g. SIGKILL) I surmise that it cannot be necessary for safety.

If we want to continue to execute this code on clean cancellation we can instead put back the cancellation token, keep hold of this future, and try to run it to completion.

web/@linera/client/src/lib.rs
```diff
@@ -75,7 +75,7 @@ pub fn initialize(options: Option<InitializeOptions>) {
     // If no log filter is provided, disable the user application log by default, to avoid
     // overwhelming the console with logs from the client library itself.
     let log_filter = if options.log.is_empty() {
-        "user_application_log=off,linera_client=info"
+        "user_application_log=off,linera_web=info"
```
deuszx, line 78: Why this change? I thought we want `linera_client` logs on the web?

Twey: Why do we want `linera-client` logs on the Web?

In practice this mostly meant that all high-level logs from `linera-web` were hidden, so it was a pain to debug or figure out what was going on, but the console instead had spammy logs about what the chain listener is doing in the background.  I assumed it was a typo!

Twey: I think we want _either_ `linera_web=info` _and_ `linera_client=info` logs (and indeed then we might as well use blanket `info`, since `info!` logs are generally for things that are user-facing) or else just `linera_web=info` (or indeed everything at `warn`).  It doesn't make much sense to me for a library halfway down the stack to be more verbose than the top-level application.

linera-core/src/client/chain_client/mod.rs
```diff
@@ -1028,32 +1017,18 @@ impl<Env: Environment> ChainClient<Env> {
             },
         );
 
-        let update_trackers = linera_base::task::spawn(async move {
+        stream::iter(cert_futures)
+            .buffer_unordered(self.options.max_joined_tasks)
+            .collect::<()>()
+            .await;
+
+        let update_trackers = linera_base::Task::spawn(async move {
```
deuszx: Isn't this changing the behavior? Previously, we'd _first_ spawn the `update_trackers` and then process the `cert_futures` stream. Now, we first try to process `cert_futures` and then update trackers. Previously it was parallel (those two tasks), now they're sequential and in different order.

Twey: Right you are.  I think the formatting made it look like these blocks were independent so I moved them around by accident while I was in there :facepalm: but this will be potentially quite a lot slower.  Fixed it up in a9549659faa0395a681d537a7d896eba7ffd8fab.

Twey: (concurrent, not parallel — and actually we immediately await the task anyway so in fact we can just use `join` here and forgo the `Task` entirely)

PR #5521, opened 2026-02-26
Fix premature quorum failure in `communicate_with_quorum`

afck:
## Motivation

The early return on trusted errors used `validity_threshold` (N/3), which could abort when a quorum was still achievable. With 6 weight-100 validators `validity_threshold` is 200 (see https://github.com/linera-io/linera-protocol/pull/4978) but `quorum_threshold` is 400, so 2 errors (weight 200) still left room for 4 successes (weight 400).

This made at least one test flaky, because it could happen that:
* A client requested a chain, and the faucet created it with the help of 4 validators (a quorum).
* Then the client tried to sync the chain, but the first 2 validators (validity threshold!) to respond were the ones that do not know about the new chain yet, so they send back errors.

## Proposal

Use only the `total_votes - quorum_threshold` criterium instead, so we only give up when remaining validators truly cannot form a quorum. Determine whether any error is trusted at the end.

## Test Plan

`test_end_to_end_receipt_of_old_remove_committee_messages` should not be flaky anymore. (Confirmed locally.)

## Release Plan

- (No need to backport: on testnet, the quorum would still be 401 in that scenario.)

## Links

### DIFF
--- a/linera-core/src/updater.rs
+++ b/linera-core/src/updater.rs
@@ -220,7 +220,7 @@ where
         [] => CommunicationError::NoConsensus(committee.quorum_threshold(), scores),
         [(_, score), ..] if *score >= committee.validity_threshold() => {
             // At least one honest validator returned this error.
-            CommunicationError::Trusted(sample.remove(0).0)
+            CommunicationError::Trusted(sample.into_iter().next().unwrap().0)
         }
         // Otherwise no specific error is available to report reliably.}
         _ => CommunicationError::Sample(sample),


linera-core/src/updater.rs
```diff
@@ -187,11 +187,6 @@ where
                 };
                 let entry = error_scores.entry(err.clone()).or_insert(0);
                 *entry += committee.weight(&name);
-                if *entry >= committee.validity_threshold() {
-                    // At least one honest node returned this error.
-                    // No quorum can be reached, so return early.
-                    return Err(CommunicationError::Trusted(err));
-                }
```
bart-linera, line 194: Do we want to remove the early return entirely after all?

afck: Yes: IMO the only criterion to exit the loop should be the one right below: if we can't reach a quorum anymore.

bart-linera: Ah, I missed the check below. That makes sense :+1:

linera-core/src/updater.rs
```diff
@@ -225,11 +220,16 @@ where
         ));
     }
 
-    // No specific error is available to report reliably.
     let mut sample = error_scores.into_iter().collect::<Vec<_>>();
     sample.sort_by_key(|(_, score)| std::cmp::Reverse(*score));
     sample.truncate(4);
-    Err(CommunicationError::Sample(sample))
+    if sample[0].1 >= committee.validity_threshold() {
```
deuszx: if `sample` is empty (in theory shouldn't happen) the `sample[0]` will throw. So I guess `sample.first().map_err(..)?.value()` ?

afck: Then we'd have returned above; see line 216.

linera-core/src/updater.rs
```diff
@@ -225,11 +220,16 @@ where
...
+        // At least one honest validator returned this error.
+        Err(CommunicationError::Trusted(sample.remove(0).0))
+    } else {
+        // No specific error is available to report reliably.
+        Err(CommunicationError::Sample(sample))
+    }
```
Twey: A nit following on from Mateusz' comment: I think it would be clearer (i.e. eliminate the possibility of the misunderstanding he encountered) to sort first then use a match with slice patterns:

```suggestion
    let mut sample = error_scores.into_iter().collect::<Vec<_>>();
    sample.sort_by_key(|(_, score)| std::cmp::Reverse(*score));
    sample.truncate(4);

    Err(match sample.as_slice() {
        [] => CommunicationError::NoConsensus(
            committee.quorum_threshold(),
            scores,
        ),
        
        // At least one honest validator returned this error.
        [(error, score), ..] if score >= committee.validity_threshold() =>
            CommunicationError::Trusted(error.clone()),
            
        // No specific error is available to report reliably.
        _ => CommunicationError::Sample(sample),
    })
```

Even better might be to encapsulate this conversion in a function with a descriptive name:

```rust
impl CommunicationError {
    fn from_node_errors(committee: &Committee, errors: Vec<NodeError>) -> Self {
        …
    }
}
```

then just

```suggestion
    Err(CommunicationError::from_node_errors(committee, error_scores))
```

ma2bd:
~~The validity threshold should not be N/3~~ (EDIT: discussed)

linera-core/src/updater.rs
afck: Done in https://github.com/linera-io/linera-protocol/pull/5521/commits/4c6e98e145333ba0fa1d49638e1b93c41e2577f9.

linera-core/src/updater.rs
```diff
@@ -218,18 +213,18 @@ where
...
-    }
-
-    // No specific error is available to report reliably.
     let mut sample = error_scores.into_iter().collect::<Vec<_>>();
     sample.sort_by_key(|(_, score)| std::cmp::Reverse(*score));
     sample.truncate(4);
-    Err(CommunicationError::Sample(sample))
+    Err(match sample.as_slice() {
+        [] => CommunicationError::NoConsensus(committee.quorum_threshold(), scores),
+        [(_, score)] if *score >= committee.validity_threshold() => {
```
Twey: This (omitting the `..`) has different behaviour from before, i.e. it will now only hit this branch if the validators unanimously agree on the error (where before it would take the highest-scoring error so long as that score exceeded the validity threshold).  Is this intentional?

linera-core/src/updater.rs
```diff
@@ -218,18 +213,18 @@ where
...
+            // At least one honest validator returned this error.
+            CommunicationError::Trusted(sample.remove(0).0)
```
Twey: I actually think it might be cheaper to `error.clone()` than to `sample.remove(0)`, which will cause a full reshuffle of the vector.

afck: Or `sample.into_iter().next()`?

linera-core/src/updater.rs
afck: Ah, no! Good catch!

afck:
As discussed, the validity threshold is indeed exactly $$N / 3$$ in some cases.

E.g. if $$N = 3$$, we have to assume that all validators are honest, i.e. $$f = 0$$. Therefore the threshold is $$1$$. The quorum is now defined as the smallest $$q$$ such that any two sets of weight $$q$$ intersect in a validity threshold (i.e. have at least one honest node in common), so it is $$2$$ in that case.

The change in this PR is that we don't give up after one error from a validator, even though that is a validity threshold, because the remaining two would still form a quorum.

linera-core/src/updater.rs
```diff
@@ -218,18 +213,18 @@ where
...
-    // No specific error is available to report reliably.
     let mut sample = error_scores.into_iter().collect::<Vec<_>>();
     sample.sort_by_key(|(_, score)| std::cmp::Reverse(*score));
     sample.truncate(4);
-    Err(CommunicationError::Sample(sample))
+    Err(match sample.as_slice() {
+        [] => CommunicationError::NoConsensus(committee.quorum_threshold(), scores),
+        [(_, score), ..] if *score >= committee.validity_threshold() => {
+            // At least one honest validator returned this error.
+            CommunicationError::Trusted(sample.into_iter().next().unwrap().0)
```
Twey, line 223: Clever :)

PR #5522, opened 2026-02-26
In `Committee::new`, don't wrap/panic on overflow.

afck:
## Motivation

The `total_votes` fold and `total_votes + validity_threshold` addition in `Committee::new` can overflow in theory.

Thanks to @deuszx for pointing this out!

## Proposal

Use `checked_add` and return `ArithmeticError::Overflow` so callers handle it gracefully.

## Test Plan

CI

## Release Plan

## Links

### DIFF
--- a/linera-core/src/unit_tests/client_tests.rs
+++ b/linera-core/src/unit_tests/client_tests.rs
@@ -1226,8 +1226,7 @@ where
     let user = builder.add_root_chain(1, Amount::ZERO).await?;
     let validators = builder.initial_committee.validators().clone();
 
-    let committee = Committee::new(validators.clone(), ResourceControlPolicy::only_fuel())
-        .expect("test committee votes should not overflow");
+    let committee = Committee::new(validators.clone(), ResourceControlPolicy::only_fuel())?;
     admin.stage_new_committee(committee).await.unwrap();
 
     // Root chain 1 receives the notification about the new epoch.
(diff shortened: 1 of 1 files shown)


linera-core/src/unit_tests/client_tests.rs
```diff
@@ -1226,7 +1226,8 @@ where
     let user = builder.add_root_chain(1, Amount::ZERO).await?;
     let validators = builder.initial_committee.validators().clone();
 
-    let committee = Committee::new(validators.clone(), ResourceControlPolicy::only_fuel());
+    let committee = Committee::new(validators.clone(), ResourceControlPolicy::only_fuel())
+        .expect("test committee votes should not overflow");
```
deuszx: Since tests return a `Result` already, can we simply `?`?

linera-execution/src/committee.rs
```diff
@@ -219,22 +225,30 @@ impl Committee {
     pub fn new(
         validators: BTreeMap<ValidatorPublicKey, ValidatorState>,
         policy: ResourceControlPolicy,
-    ) -> Self {
-        let total_votes = validators.values().fold(0, |sum, state| sum + state.votes);
+    ) -> Result<Self, ArithmeticError> {
```
deuszx, line 228: `ArithmeticError` is bit ambiguous – i.e. w/o the context it's impossible to understand what really failed. At some point this error here needs to be interpreted into something more user-friendly I think

afck: We'd have to have a new client error variant and even a new `ViewError` variant just for this case, which in practice should only be reachable anyway if a faulty node sent us a wrong committee. And it's not like this error is recoverable or the user can do anything about it. Not sure what's the best approach here.

PR #5523, opened 2026-02-26
[testnet] Do not retry tasks if they were unassigned from the worker

### LINKED ISSUE #5503: Quickly kill retry loops (and other pending tasks) when an oracle is unassigned from a worker
Currently, it seems like this can only achieve by restarting the worker.

bart-linera:
## Motivation

Unassigning a service from a worker sometimes doesn't stop the worker from processing the service's tasks.

## Proposal

Only re-process actions from an application after receiving a result from it if it is still assigned to the worker.

## Test Plan

CI will catch regressions

## Release Plan

## Links

- closes #5503

### DIFF
--- a/linera-service/src/task_processor.rs
+++ b/linera-service/src/task_processor.rs
@@ -200,6 +200,10 @@ impl<Env: linera_core::Environment> TaskProcessor<Env> {
 
     async fn process_actions(&mut self, application_ids: Vec<ApplicationId>) {
         for application_id in application_ids {
+            if !self.application_ids.contains(&application_id) {
+                debug!("Skipping {application_id}: it's no longer assigned to this processor");
+                continue;
+            }
             if self.in_flight_apps.contains(&application_id) {
                 debug!("Skipping {application_id}: tasks already in flight");
                 continue;


linera-service/src/task_processor.rs
```diff
@@ -119,14 +119,18 @@ impl<Env: linera_core::Environment> TaskProcessor<Env> {
...
-                        self.deadlines.push(Reverse((
-                            retry_at,
-                            Some(result.application_id),
-                        )));
-                    } else {
-                        // Re-process immediately to pick up new tasks.
-                        self.process_actions(vec![result.application_id]).await;
+                    // The application could have been unassigned from this processor
+                    // in the meantime - do not retry if that is the case.
+                    if self.application_ids.contains(&result.application_id) {
```
afck, line 124: Should we just do that check in `process_actions` instead in general?

bart-linera: We could, but we'd waste some cycles then: we'd schedule a retry, and only after retrying realize that we shouldn't be doing it.

But actually, maybe we should be doing the check in both places, because an application may be unassigned between when the retry is scheduled and when it's actually happening :thinking:

PR #5528, opened 2026-02-26
ci: add release workflow for prebuilt CLI binaries

### LINKED ISSUE #4587: Support binary installation using `cargo-binstall`
We've noticed that on slower machines installing `linera-service` (where the CLI lives) from source can be rather painstaking.

It would be nice to support [`cargo-binstall`](https://github.com/cargo-bins/cargo-binstall), a tool that tries to download binary releases of a package.  In order to do this we would need to:
1) save built GitHub release artefacts from (e.g.) testnet branches for our supported platforms (`x86_64-unknown-linux-gnu` and `aarch64-apple-darwin`)
2) (perhaps) add some metadata to `Cargo.toml`

nick-ruzicka:
## Motivation

Building from source takes 35+ minutes and requires protoc workarounds. Prebuilt binaries reduce CLI install to 30 seconds.

Based on a [measured DX friction report](https://github.com/nick-ruzicka/linera-developer-guide/blob/main/VALIDATION.md) from initial testnet setup — CLI compilation accounted for 71% of total onboarding time.

Fixes #4587

## Proposal

Adds `.github/workflows/release-binaries.yml` that builds `linera` for:
- Linux x86_64 (glibc)
- Linux x86_64 (musl — for Alpine, NixOS, non-glibc distros)
- macOS Intel (x86_64)
- macOS Apple Silicon (ARM)

on every tagged release and attaches binaries to the GitHub Release.

## Test Plan

- [ ] Push a test tag and verify workflow runs
- [ ] Verify binaries are attached to GitHub Release
- [ ] Test install on each platform
- [ ] Verify `cargo-binstall` can discover the artifacts (follow-up: Cargo.toml metadata)

### DIFF
--- a/.github/workflows/release-binaries.yml
+++ b/.github/workflows/release-binaries.yml
@@ -51,6 +51,11 @@ jobs:
             archive: tar.gz
             name_suffix: Linux-x86_64
 
+          - target: x86_64-unknown-linux-musl
+            os: linera-io-self-hosted-ci
+            archive: tar.gz
+            name_suffix: Linux-x86_64-musl
+
           - target: x86_64-apple-darwin
             os: macos-14-large    # Apple Silicon runner (cross-compiles to x86_64)
             archive: tar.gz
@@ -81,6 +86,9 @@ jobs:
         run: |
           sudo apt-get update
           sudo apt-get install -y pkg-config libssl-dev
+          if [[ "${{ matrix.target }}" == *"musl"* ]]; then
+            sudo apt-get install -y musl-tools
+          fi
 
       - name: Build release binaries
         run: |
@@ -212,10 +220,16 @@ jobs:
             sudo mv linera-${{ steps.version.outputs.semver }}-Darwin-x86_64/linera /usr/local/bin/
             ```
 
-            **Linux (x86_64):**
+            **Linux (x86_64, glibc):**
             ```bash
             curl -fsSL https://github.com/linera-io/linera-protocol/releases/download/${{ steps.version.outputs.version }}/linera-${{ steps.version.outputs.semver }}-Linux-x86_64.tar.gz | tar xz
             sudo mv linera-${{ steps.version.outputs.semver }}-Linux-x86_64/linera /usr/local/bin/
+            ```
+
+            **Linux (x86_64, musl — Alpine, NixOS, etc.):**
+            ```bash
+            curl -fsSL https://github.com/linera-io/linera-protocol/releases/download/${{ steps.version.outputs.version }}/linera-${{ steps.version.outputs.semver }}-Linux-x86_64-musl.tar.gz | tar xz
+            sudo mv linera-${{ steps.version.outputs.semver }}-Linux-x86_64-musl/linera /usr/local/bin/
             ```
 
             **Verify:**
--- a/.github/workflows/release-binaries.yml
+++ b/.github/workflows/release-binaries.yml
@@ -52,7 +52,7 @@ jobs:
             name_suffix: Linux-x86_64
 
           - target: x86_64-apple-darwin
-            os: macos-14-large    # Intel Mac runner
+            os: macos-14-large    # Apple Silicon runner (cross-compiles to x86_64)
             archive: tar.gz
             name_suffix: Darwin-x86_64
 
--- a/.github/workflows/release-binaries.yml
+++ b/.github/workflows/release-binaries.yml
@@ -1,0 +1,227 @@
+# .github/workflows/release-binaries.yml
+#
+# Builds and uploads prebuilt Linera CLI binaries on every tagged release.
+# Targets: Linux x86_64, macOS x86_64, macOS ARM (Apple Silicon)
+#
+# Usage:
+#   1. Push a tag:  git tag linera-v0.15.12 && git push origin linera-v0.15.12
+#   2. This workflow creates a GitHub Release with binaries attached
+#   3. Users install with:
+#        curl -fsSL https://github.com/linera-io/linera-protocol/releases/latest/download/linera-<semver>-$(uname -s)-$(uname -m).tar.gz | tar xz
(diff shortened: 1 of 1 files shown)


.github/workflows/release-binaries.yml
```diff
@@ -0,0 +1,227 @@
...
+  # ============================================================
+  build:
+    name: Build ${{ matrix.target }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false     # don't cancel other builds if one fails
+      matrix:
+        include:
+          - target: x86_64-unknown-linux-gnu
+            os: ubuntu-latest-8-cores
```
ndr-ds: We have our own runners, let's use them here because it's cheaper: `linera-io-self-hosted-ci`

.github/workflows/release-binaries.yml
```diff
@@ -0,0 +1,227 @@
...
+            archive: tar.gz
+            name_suffix: Linux-x86_64
+
+          - target: x86_64-apple-darwin
+            os: macos-13          # Intel Mac runner
```
ndr-ds: `macos-13` has been retired: https://github.blog/changelog/2025-09-19-github-actions-macos-13-runner-image-is-closing-down/  
You can use `macos-15-intel`, or `macos-14-large` here. The latter is more expensive, but has more cores, so it might be able to build faster and turn out to actually be cheaper 😅 (they're charged per minute of use)  
But hard to tell

.github/workflows/release-binaries.yml
```diff
@@ -0,0 +1,227 @@
+# .github/workflows/release-binaries.yml
+#
+# Builds and uploads prebuilt Linera CLI binaries on every tagged release.
+# Targets: Linux x86_64, macOS x86_64, macOS ARM (Apple Silicon)
+#
+# Usage:
+#   1. Push a tag:  git tag linera-v0.15.12 && git push origin linera-v0.15.12
+#   2. This workflow creates a GitHub Release with binaries attached
+#   3. Users install with:
+#        curl -fsSL https://github.com/linera-io/linera-protocol/releases/latest/download/linera-$(uname -s)-$(uname -m).tar.gz | tar xz
```
ndr-ds: Don't the archives include the semver?  
Like `linera-<semver>-...tar.gz.`? I don't think this URL will resolve

.github/workflows/release-binaries.yml
```diff
@@ -0,0 +1,227 @@
...
+        id: package
+        run: |
+          STAGING="linera-${{ steps.version.outputs.semver }}-${{ matrix.name_suffix }}"
+          mkdir -p "$STAGING"
+
+          # Copy binaries
+          cp target/${{ matrix.target }}/release/linera "$STAGING/"
+          cp target/${{ matrix.target }}/release/linera-server "$STAGING/" 2>/dev/null || true
+          cp target/${{ matrix.target }}/release/linera-proxy "$STAGING/" 2>/dev/null || true
+          cp target/${{ matrix.target }}/release/linera-storage-service "$STAGING/" 2>/dev/null || true
```
ndr-ds: The `2>/dev/null || true` isn't necessary because the build step runs before this, so if a binary fails to build, the job already fails before reaching this point.  
You can probably just remove that part

ndr-ds (changes requested):
The SDK release tags are pushed to `testnet_conway`, so this PR actually needs to be based off of `testnet_conway`, not `main`, otherwise I don't think it'll ever do anything.  
Another thing to keep in mind is that the mac runners are 5-6x more expensive than their linux counterparts 😅 so we should probably be careful with that, and not overuse it (whatever that means).

nick-ruzicka:
@ndr-ds Addressed all review feedback — rebased onto testnet_conway, switched to self-hosted runners, updated macOS runner to macos-14-large, fixed download URL with semver, removed unnecessary error suppression. Ready for re-review.

.github/workflows/release-binaries.yml
```diff
@@ -0,0 +1,227 @@
...
+      fail-fast: false     # don't cancel other builds if one fails
+      matrix:
+        include:
+          - target: x86_64-unknown-linux-gnu
+            os: linera-io-self-hosted-ci
+            archive: tar.gz
+            name_suffix: Linux-x86_64
+
+          - target: x86_64-apple-darwin
+            os: macos-14-large    # Intel Mac runner
```
ndr-ds: This is actually a Apple Silicon Mac, AFAIK, so we need to update the comment

ndr-ds (approved):

Twey:
Does this fix https://github.com/linera-io/linera-protocol/issues/4587?

.github/workflows/release-binaries.yml
```diff
@@ -0,0 +1,227 @@
...
+  # Build matrix — one job per target platform
+  # ============================================================
+  build:
+    name: Build ${{ matrix.target }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false     # don't cancel other builds if one fails
+      matrix:
+        include:
+          - target: x86_64-unknown-linux-gnu
```
Twey, line 49: It would be nicer to build statically-linked binaries for `x86_64-unknown-linux-musl` so that they work out of the box on non-glibc or non-FHS distros like Alpine or NixOS.

nick-ruzicka: claude code :   Yes — this produces the GitHub Release artifacts that cargo-binstall needs (step 1 from the issue). Step 2 (Cargo.toml metadata for auto-discovery) can be a follow-up once the workflow is merged and we've verified the archive naming convention works with binstall.

nick-ruzicka: Added — new x86_64-unknown-linux-musl target in 53c5bc1. Installs musl-tools on the Linux runner when building the musl target. Install instructions updated with a separate musl download option.

PR #5529, opened 2026-02-26
Extract  from the linera binary into linera-service library

ndr-ds:
## Motivation

The `linera` binary's CLI `Options` struct bundles wallet, keystore, and storage
configuration fields that other tools (e.g. `pm-benchmark`) also need. Currently these
fields and their associated methods live in a private module of the binary
(`linera-service/src/cli/options.rs`), so external consumers must duplicate them.

## Proposal

Extract the shared wallet/keystore/storage CLI options into a new public struct
`CommonClientOptions` in the linera-service library crate
(`linera-service/src/cli/common_options.rs`). The binary's `Options` struct now flattens
`CommonClientOptions` via `#[command(flatten)]` and provides thin delegation methods to
preserve the existing API surface used by `main.rs`.

The CLI behavior is unchanged — all flags, env vars, and defaults remain identical.

## Test Plan

- CI

### DIFF
--- a/linera-service/src/cli/options.rs
+++ b/linera-service/src/cli/options.rs
@@ -1,19 +1,19 @@
 // Copyright (c) Zefchain Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-use std::{env, path::PathBuf};
+use std::path::PathBuf;
 
-use anyhow::{anyhow, bail, Error};
+use anyhow::Error;
 use linera_base::crypto::InMemorySigner;
 use linera_client::{client_context::ClientContext, config::GenesisConfig};
-use linera_execution::{WasmRuntime, WithWasmDefault as _};
+use linera_execution::WithWasmDefault as _;
 use linera_persistent as persistent;
 use linera_service::{
-    cli::command::ClientCommand,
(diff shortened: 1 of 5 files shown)


linera-service/src/cli/options.rs
```diff
@@ -26,20 +26,9 @@ pub struct Options {
...
-
-    /// Given an ASCII alphanumeric parameter `X`, read the wallet state and the wallet
-    /// storage config from the environment variables `LINERA_WALLET_{X}` and
-    /// `LINERA_STORAGE_{X}` instead of `LINERA_WALLET` and
-    /// `LINERA_STORAGE`.
-    #[arg(long, short = 'w', value_parser = crate::util::parse_ascii_alphanumeric_string)]
-    pub with_wallet: Option<String>,
+    /// Common wallet, keystore, and storage options.
+    #[command(flatten)]
+    pub common: CommonClientOptions,
```
ma2bd: "common client options" is a weird name when we already have `client_options`. I think what we mean is options for CLI tools?

Maybe something like `cli_options` or `tool_client_options` or `executable_client_options` or `cli_tool_options`?

PR #5372, opened 2026-02-02
Distinguish synthetic from organic traffic in metrics and traces

ndr-ds:
## Motivation

Forward port of #5334

## Proposal

Forward port the PR

## Test Plan

CI + Tested on the `testnet_conway` branch

PR #5534, opened 2026-02-27
Fix proxy metrics cardinality panic on pending blob handlers

ndr-ds:
## Motivation

PR #5372 was a forward port of #5334 ("Distinguish synthetic from organic traffic in
metrics and traces") from `testnet_conway` to `main`. The original PR on
`testnet_conway` worked correctly, but during the forward port, `handle_pending_blob`
and `download_pending_blob` were left with only 1 label in their `.with_label_values()`
calls, while the metric definitions were updated to require 2 labels
(`METHOD_NAME_LABEL`, `TRAFFIC_TYPE_LABEL`).

This causes a panic on every request to these handlers:

InconsistentCardinality { expect: 2, got: 1 }

The panic kills the tokio task, causing tonic to send `RST_STREAM CANCEL` to the client.
This breaks every test that publishes WASM blobs — the `remote-kubernetes-net-test`
workflow has been 100% failing (72+ consecutive runs) since Feb 4.

## Proposal

- `download_pending_blob`: Add `traffic_type` extraction and pass both labels to
`.with_label_values()`.
- `handle_pending_blob`: Replace inline metrics with the existing
`log_and_return_proxy_request_outcome` helper (which already handles both labels
correctly), matching the pattern used by all other proxy handlers.

## Test Plan

- CI
- Trigger `remote-kubernetes-net-test` workflow, make sure it succeeds now https://github.com/linera-io/linera-protocol/actions/runs/22489571066/job/65147852821

## Release Plan

### DIFF
--- a/linera-service/src/proxy/grpc.rs
+++ b/linera-service/src/proxy/grpc.rs
@@ -22,8 +22,10 @@ use linera_core::{
 };
 #[cfg(with_metrics)]
 use linera_metrics::monitoring_server;
+#[cfg(all(with_metrics, feature = "opentelemetry"))]
+use linera_rpc::propagation::get_traffic_type_from_request;
 #[cfg(feature = "opentelemetry")]
-use linera_rpc::propagation::{get_traffic_type_from_request, OtelContextLayer};
+use linera_rpc::propagation::OtelContextLayer;
 use linera_rpc::{
     config::{ProxyConfig, ShardConfig, TlsConfig, ValidatorInternalNetworkConfig},
     grpc::{
@@ -121,10 +123,10 @@ impl<S> Layer<S> for PrometheusMetricsMiddlewareLayer {
     }
 }
 
-impl<S, B> Service<http::Request<B>> for PrometheusMetricsMiddlewareService<S>
+impl<S, B, ResponseBody> Service<http::Request<B>> for PrometheusMetricsMiddlewareService<S>
 where
+    S: Service<http::Request<B>, Response = http::Response<ResponseBody>> + Send,
     S::Future: Send + 'static,
-    S: Service<http::Request<B>> + std::marker::Send,
     B: Send + 'static,
 {
     type Response = S::Response;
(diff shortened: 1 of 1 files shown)


ndr-ds:
* **#5534** <a href="https://app.graphite.com/github/pr/linera-io/linera-protocol/5534?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/linera-io/linera-protocol/5534?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

.github/workflows/remote-kubernetes-net-test.yml
```diff
@@ -66,10 +71,25 @@ jobs:
...
+          while true; do
+            KIND_CTX=$(kubectl config get-contexts -o name 2>/dev/null | grep '^kind-' | head -1)
+            if [ -n "$KIND_CTX" ]; then
+              kubectl config set-context "$KIND_CTX" --namespace=default
+              echo "Fixed namespace for context $KIND_CTX to 'default'"
+              break
+            fi
+            sleep 1
+          done
+        ) &
```
deuszx, line 87: Was this in the original PR as well?

ndr-ds: I will have to backport some of this to `testnet_conway`

linera-service/src/proxy/grpc.rs
```diff
@@ -668,24 +670,13 @@ where
...
-                    .with_label_values(&["handle_pending_blob"])
-                    .inc();
-                Err(status)
-            }
-        }
+        Self::log_and_return_proxy_request_outcome(
+            client.handle_pending_blob(inner).await,
+            "handle_pending_blob",
+            traffic_type,
+        )
```
afck: Is there a reason we don't use this below e.g. in `download_certificate`?

ndr-ds: Yeah, true, good point. This honestly should be some sort of tonic layer that wraps everything, or something like that. Let me look into that

linera-service/src/proxy/grpc.rs
```diff
@@ -158,6 +164,23 @@ where
                 metrics::PROXY_REQUEST_COUNT
                     .with_label_values(&[traffic_type])
                     .inc();
+
+                let is_error = !response.status().is_success()
+                    || response
+                        .headers()
+                        .get("grpc-status")
+                        .and_then(|v| v.to_str().ok())
+                        .is_some_and(|s| s != "0");
```
afck, line 175: A "0" response with success status is also an error?

ndr-ds: It will be an error if the HTTP status is non-200 OR if the result is 200 and the `grpc-status` is non 0. gRPC always returns 200, but there's a tonic bug now that it can return an HTML response instead with a non-200 status code.  
So I'm trying to catch both here

PR #5535, opened 2026-02-27
EVM to Linera ERC20 transfer

deuszx:
## Motivation

  Enable ERC-20 token transfers from EVM chains (Ethereum, Base L2) to Linera microchains. The existing bridge only supported the Linera→EVM direction (releasing locked ERC-20 tokens on Ethereum when Credit
  messages are processed). This PR adds the reverse path: lock tokens on EVM, prove the deposit on Linera, mint wrapped tokens.

  ## Proposal

  ### Architecture

  The EVM→Linera deposit flow has four stages:

  1. **EVM deposit**: user calls `FungibleBridge.deposit()`, which locks ERC-20 tokens and emits a `DepositInitiated` event
  2. **Proof generation**: an off-chain client fetches the transaction receipt from the EVM RPC, builds a Merkle Patricia Trie (MPT) over all block receipts, and extracts a proof for the target receipt
  3. **On-chain verification**: the `evm-bridge` Linera Wasm app receives a `ProcessDeposit` operation containing the RLP-encoded block header, receipt, and MPT proof nodes. It verifies receipt inclusion against
  the block's `receiptsRoot`, then parses and validates the `DepositInitiated` event.
  4. **Token minting**: the `evm-bridge` app issues a cross-application call to `wrapped-fungible` to mint the corresponding amount to the target account

  ### Solidity changes (`FungibleBridge.sol`)

  - Add `deposit(bytes32 target_chain_id, bytes32 target_application_id, bytes32 target_account_owner, uint256 amount)` function
  - Emit `DepositInitiated` event with 8 fields: `source_chain_id`, `target_chain_id`, `target_application_id`, `target_account_owner`, `depositor` (indexed), `token`, `amount`, `nonce`
  - `_safeTransferFrom` helper handles tokens that don't return a boolean (e.g. USDT)
  - Balance-delta check rejects fee-on-transfer tokens that would cause under-collateralization
  - `require(amount > 0)` prevents zero-value deposit noise
  - Monotonically increasing `depositNonce` for EVM-side replay resistance

  ### MPT proof module (`linera-bridge/src/proof/`)

  Pure-Rust implementation (~1,050 lines) that runs both natively and in Wasm:

  - **`mod.rs`** (on-chain verification): `decode_block_header` (RLP→block hash + receipts root), `verify_receipt_inclusion` (MPT proof walk), `decode_receipt_logs`, `parse_deposit_event`. Feature-gated behind
  `onchain` for Wasm compatibility (no-std-friendly, no allocator-heavy dependencies).
  - **`gen.rs`** (off-chain proof generation): `DepositProofClient` trait with `HttpDepositProofClient` implementation. Queries standard Ethereum JSON-RPC (`eth_getTransactionReceipt`, `eth_getBlockByHash`,
  `eth_getBlockReceipts`), builds the receipts trie locally, generates the MPT proof. Works with any EVM-compatible chain including Base L2.

  Design decisions in the proof module:
  - **DoS bounds**: proof verification caps at 32 nodes / 32 KiB total. Derivation: a block with ~1,400 txs (30M gas / 21K per tx) yields trie depth ~11; 32 nodes is ~3x headroom. Each branch node is ~600 bytes, so
   32 × 600 ≈ 19 KiB; 32 KiB is ~1.7x that.
  - **Emitter binding**: `parse_deposit_event` requires an `expected_emitter: Address` parameter to prevent spoofed events from arbitrary contracts.
  - **Strict topic count**: requires exactly 2 topics (event signature + indexed depositor) to reject logs that don't match the Solidity event definition.
  - **Type byte handling**: type byte `0x00` is permitted (treated as legacy receipt) because the MPT proof already binds the receipt to a specific trie slot—accepting it is harmless and more compatible with edge
  cases.

  ### Fungible token extensions

  - Add `Mint` and `Burn` variants to `FungibleOperation` in `linera-sdk/src/abis/fungible.rs`. Standard fungible rejects these with a panic.
  - New **`wrapped-fungible`** example app: supports `Mint` (authorized by a minter parameter) and `Burn` for bridge-backed tokens. Exposes `evm_token_address` and `evm_source_chain_id` via GraphQL. Reuses
  fungible's state module (made public).

  ### New `evm-bridge` Linera app

  Wasm contract (`examples/evm-bridge/`) that:
  - Accepts `ProcessDeposit { block_header_rlp, receipt_rlp, proof_nodes, tx_index, log_index }`
  - Verifies the block header hash, receipt MPT inclusion, and deposit event fields against its `BridgeParameters` (source chain ID, bridge contract address, token address)
  - Issues a cross-app `Mint` call to `wrapped-fungible`

  ### Code organization

  - `linera-bridge` gains `onchain` / `offchain` feature flags. The `onchain` feature compiles proof verification for Wasm. The `offchain` feature includes the proof generation client and alloy dependencies.
  - `proof::testing` module behind a `testing` feature consolidates test helpers (receipt builders, deposit event builders).
  - Duplicate code removed: `validator_evm_address`, `build_receipt_proof`, `receipt_trie_key`, certificate creation functions, and Solidity source aliases.

  ### Hardening (from review)

  Applied across two review rounds:
  - SafeERC20 pattern + fee-on-transfer rejection + `amount > 0` guard
  - Indexed `depositor` field in the event for easier indexing and analytics
  - Deposit nonce for EVM-side replay resistance
  - DoS bounds on MPT proof verification (32 nodes / 32 KiB)
  - Emitter address binding + exact topic count enforcement
  - 15+ tests covering empty inputs, truncated data, type byte edge cases, boundary overflows, trailing data, wrong tx indices, wrong emitters, and extra topics

  ## Test Plan

  CI

  - **Unit tests**: 70+ tests in `linera-bridge` covering proof construction, verification, event parsing, and edge cases
  - **Integration test**: `anvil_deposit_proof` — deploys MockERC20 + FungibleBridge on a live Anvil node, executes a deposit, generates the proof, verifies it end-to-end (requires `anvil` + `solc`)
  - **Docker e2e tests**: `evm_to_linera_bridge` — full pipeline from EVM deposit through Linera minting with balance verification; `fungible_bridge` — Linera→EVM direction still works
  - **Wasm unit tests**: `evm-bridge/tests/process_deposit` — 3 tests exercising the on-chain contract with valid proofs, wrong chain ID, and wrong bridge address

  ## Release Plan

  TBD

  ## Links

### DIFF
--- a/linera-bridge/tests/e2e/.cargo/config.toml
+++ b/linera-bridge/tests/e2e/.cargo/config.toml
@@ -1,0 +1,5 @@
+[resolver]
+# Prefer crate versions compatible with our rust-version (1.88).
+# This prevents alloy sub-crates from resolving to 1.7+ (MSRV 1.91),
+# which would pull in Rust 1.91 and break linera-wasmer-vm's probestack.
+incompatible-rust-versions = "fallback"
--- a/linera-bridge/tests/e2e/Cargo.lock
+++ b/linera-bridge/tests/e2e/Cargo.lock
@@ -76,9 +76,9 @@ checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
 name = "alloy"
-version = "1.7.3"
+version = "1.6.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4973038846323e4e69a433916522195dce2947770076c03078fc21c80ea0f1c4"
+checksum = "07dc44b606f29348ce7c127e7f872a6d2df3cfeff85b7d6bba62faca75112fdd"
 dependencies = [
  "alloy-consensus",
  "alloy-contract",
@@ -110,9 +110,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-consensus"
-version = "1.7.3"
+version = "1.6.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b0c0dc44157867da82c469c13186015b86abef209bf0e41625e4b68bac61d728"
+checksum = "4e4ff99651d46cef43767b5e8262ea228cd05287409ccb0c947cc25e70a952f9"
 dependencies = [
  "alloy-eips",
  "alloy-primitives",
@@ -137,9 +137,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-consensus-any"
-version = "1.7.3"
+version = "1.6.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ba4cdb42df3871cd6b346d6a938ec2ba69a9a0f49d1f82714bc5c48349268434"
+checksum = "1a0701b0eda8051a2398591113e7862f807ccdd3315d0b441f06c2a0865a379b"
 dependencies = [
  "alloy-consensus",
  "alloy-eips",
@@ -151,9 +151,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-contract"
-version = "1.7.3"
+version = "1.6.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ca63b7125a981415898ffe2a2a696c83696c9c6bdb1671c8a912946bbd8e49e7"
+checksum = "f3c83c7a3c4e1151e8cac383d0a67ddf358f37e5ea51c95a1283d897c9de0a5a"
 dependencies = [
  "alloy-consensus",
  "alloy-dyn-abi",
@@ -252,9 +252,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-eips"
-version = "1.7.3"
+version = "1.6.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b9f7ef09f21bd1e9cb8a686f168cb4a206646804567f0889eadb8dcc4c9288c8"
+checksum = "def1626eea28d48c6cc0a6f16f34d4af0001906e4f889df6c660b39c86fd044d"
 dependencies = [
  "alloy-eip2124",
  "alloy-eip2930",
@@ -288,9 +288,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-json-rpc"
-version = "1.7.3"
+version = "1.6.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ff42cd777eea61f370c0b10f2648a1c81e0b783066cd7269228aa993afd487f7"
+checksum = "e57586581f2008933241d16c3e3f633168b3a5d2738c5c42ea5246ec5e0ef17a"
 dependencies = [
  "alloy-primitives",
  "alloy-sol-types",
@@ -303,9 +303,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-network"
-version = "1.7.3"
+version = "1.6.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8cbca04f9b410fdc51aaaf88433cbac761213905a65fe832058bcf6690585762"
+checksum = "3b36c2a0ed74e48851f78415ca5b465211bd678891ba11e88fee09eac534bab1"
 dependencies = [
  "alloy-consensus",
  "alloy-consensus-any",
@@ -329,9 +329,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-network-primitives"
-version = "1.7.3"
+version = "1.6.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "42d6d15e069a8b11f56bef2eccbad2a873c6dd4d4c81d04dda29710f5ea52f04"
+checksum = "636c8051da58802e757b76c3b65af610b95799f72423dc955737dec73de234fd"
 dependencies = [
  "alloy-consensus",
  "alloy-eips",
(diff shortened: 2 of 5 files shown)


linera-sdk/src/abis/fungible.rs
```diff
@@ -109,6 +109,20 @@ pub enum FungibleOperation {
...
+        /// Amount of tokens to mint
+        amount: Amount,
+    },
+    /// Burns tokens from an account. Only the authorized minter can call this.
+    Burn {
+        /// Account owner whose tokens to burn
+        owner: AccountOwner,
+        /// Amount of tokens to burn
+        amount: Amount,
+    },
```
ma2bd: Let's discuss if we need to change this, or create a new ABI. Especially since this is the testnet branch.

deuszx: Working on moving the Mint and Burn to the wrapped fungible.

afck: (Alternatively we could initialize the wrapped token with a very large initial balance owned by the bridge.)

linera-bridge/src/evm/mod.rs
```diff
@@ -0,0 +1,22 @@
...
+/// Microchain Solidity ABI (block relay on EVM).
+pub mod microchain;
+
+/// Solidity source for the BridgeTypes library (generated).
+pub const BRIDGE_TYPES_SOURCE: &str = include_str!("../solidity/BridgeTypes.sol");
+
+/// Solidity source for the FungibleTypes library (generated).
+pub const FUNGIBLE_TYPES_SOURCE: &str = include_str!("../solidity/FungibleTypes.sol");
+
+/// Solidity source for the FungibleBridge contract.
```
afck, line 21: Is this one also auto-generated?

linera-bridge/src/proof/gen.rs
```diff
@@ -0,0 +1,282 @@
...
+    primitives::B256,
+    providers::{Provider, ProviderBuilder},
+};
+use alloy_rlp::Encodable;
+use anyhow::{bail, Context, Result};
+use async_trait::async_trait;
+
+/// All data needed to submit a `ProcessDeposit` operation to the evm-bridge app.
+#[derive(Debug, Clone)]
+pub struct DepositProof {
```
afck, line 31: This only proves that _if_ a block with that hash was included on the EVM chain, then that particular transaction is included, right?

For the fact that the block hash itself is correct, we'd still trust this client for now, i.e. trust whoever owns the bridge microchain?

But I guess it will be easy to make the contract itself do an HTTP request to each validator's local EVM chain node to verify.
(Edit: Ah, right, that's already explained in the `proof` module. :+1:)

examples/evm-bridge/tests/process_deposit.rs
```diff
@@ -0,0 +1,353 @@
+// Copyright (c) Zefchain Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Integration tests for the EVM bridge app's ProcessDeposit operation.
```
afck, line 4:
```suggestion
//! Integration tests for the EVM bridge app's `ProcessDeposit` operation.
```

examples/evm-bridge/tests/process_deposit.rs
```diff
@@ -0,0 +1,353 @@
...
+        .await;
+
+    // Build a synthetic deposit event and proof
+    let deposit_amount = 1_000_000u64;
+
+    // ChainId(CryptoHash) → [u8; 32] via CryptoHash's From impl
+    let chain_id_bytes: [u8; 32] = chain.id().0.into();
+    let target_chain_b256 = B256::from(chain_id_bytes);
+
+    // The application id as bytes32 — use the forget_abi hash
```
afck, line 98:
```suggestion
    // The application ID as bytes32 — use the forget_abi hash
```

examples/evm-bridge/tests/process_deposit.rs
```diff
@@ -0,0 +1,353 @@
...
+    let app_id_bytes: [u8; 32] = [0; 32]; // placeholder, bridge doesn't check this yet
```
afck, line 99: Do we need a TODO for this?

examples/evm-bridge/tests/process_deposit.rs
```diff
@@ -0,0 +1,353 @@
...
+                    receipt_rlp: receipt,
+                    proof_nodes,
+                    tx_index,
+                    log_index: 0,
+                },
+            );
+        })
+        .await;
+
+    assert!(result.is_err(), "replay deposit should be rejected");
```
afck, line 262: This could be part of the first test, couldn't it? It seems like just an extension at the end.

examples/evm-bridge/tests/process_deposit.rs
```diff
@@ -0,0 +1,353 @@
...
+                    receipt_rlp: receipt,
+                    proof_nodes,
+                    tx_index,
+                    log_index: 0,
+                },
+            );
+        })
+        .await;
+
+    assert!(result.is_err(), "invalid proof should be rejected");
```
afck, line 352: (Personally I'd even put this in the first test, just to avoid the redundant setup steps.)

examples/wrapped-fungible/src/contract.rs
```diff
@@ -0,0 +1,239 @@
...
+    runtime: ContractRuntime<Self>,
+}
+
+linera_sdk::contract!(WrappedFungibleTokenContract);
+
+impl WithContractAbi for WrappedFungibleTokenContract {
+    type Abi = WrappedFungibleTokenAbi;
+}
+
+impl Contract for WrappedFungibleTokenContract {
```
afck, line 27: It's just `fungible` with minting and burning, is it? That can be used for other applications than just wrapping. Maybe `mintable-fungible`?

(Personally I think this would even make sense in `fungible` itself, if it can be disabled at initialization.)

linera-bridge/src/proof/gen.rs
deuszx: Yes, validators will have to do that. It's not yet in this PR though.

examples/wrapped-fungible/src/contract.rs
deuszx: I am in the process of removing Mint and Burn from the `fungible` – it will be in `WrappedFungible` and I will create a separat PR with just that – to minimize the diff for review

deuszx:
Superseeded by #5605

Issue #5537, opened 2026-02-27
Add optional LRU caching to GraphQL service


ma2bd:
* The goal is respond right away instead of locking the chain state view.
* Cache must be maintained per chain and invalidated immediately when a new block comes in.
* This is incompatible with `--long-live-service` but otherwise should be very sound semantically (mutation queries are just queries that have extra outputs to be cached as well -- or that will prevent the query from being cached)

ma2bd:
~~Sadly, we're seeing an issue with data invalidation in production~~
Never mind, it was a UI problem.

PR #5538, opened 2026-02-27
Add LRU response cache for application GraphQL queries

ndr-ds:
## Motivation

Application queries via `/chains/{chain_id}/applications/{app_id}` re-execute the
full pipeline on every request: context lock → chain client creation → view traversal
→ storage reads. There is no caching between the raw KV LRU cache and the browser's
React Query cache.

Fixes https://github.com/linera-io/linera-protocol/issues/5537

## Proposal

  Add a per-chain LRU cache for serialized application query responses, keyed on
  `(chain_id, application_id, request_bytes)`. The cache:

  - **Invalidates on new blocks**: subscribes to the existing `ChannelNotifier` and
    clears the entire per-chain cache when a `NewBlock` notification arrives.
  - **Guards against stale inserts**: each per-chain cache tracks its block height
    (set by the invalidation task). Inserts carry the height at which the query ran;
    if a newer block has since arrived, the insert is silently dropped.
  - **Skips mutation queries**: responses that produce operations are never cached.
  - **Skips historical queries**: only `block_hash: None` (latest state) is cached.
  - **Incompatible with `--long-lived-services`**: long-lived services persist state
    between queries that isn't reflected in blocks, making cached responses stale.
    The CLI panics if both flags are set.
    
### Key changes

  - `linera-core`: Added `Client::subscribe()` to expose the notifier for arbitrary
    chain ID subscriptions.
  - `linera-service`: Added `QueryResponseCache` struct using `papaya::HashMap`
    (lock-free) for per-chain access, with `std::sync::Mutex<PerChainCache>` per chain
    (holds LRU + block height).
  - CLI: Added `--query-cache-size <N>` flag (disabled by default, opt-in).
    Panics at startup if combined with `--long-lived-services`.
  - Prometheus metrics: `query_response_cache_hit`, `query_response_cache_miss`,
    `query_response_cache_invalidation`, `query_response_cache_entries` (gauge).

## Test Plan

- 5 unit tests for `QueryResponseCache`: hit/miss, per-chain isolation, invalidation,
    LRU eviction, stale insert rejection — all passing.
- CI

### DIFF
--- a/linera-service/src/node_service.rs
+++ b/linera-service/src/node_service.rs
@@ -20,7 +20,8 @@ use futures::{lock::Mutex, Future, FutureExt as _, TryStreamExt as _};
 use linera_base::{
     crypto::{CryptoError, CryptoHash},
     data_types::{
-        Amount, ApplicationDescription, ApplicationPermissions, Bytecode, Epoch, TimeDelta,
+        Amount, ApplicationDescription, ApplicationPermissions, BlockHeight, Bytecode, Epoch,
+        TimeDelta,
     },
     identifiers::{
         Account, AccountOwner, ApplicationId, ChainId, IndexAndEvent, ModuleId, StreamId,
@@ -933,14 +934,23 @@ mod query_cache_metrics {
     });
 }
 
-type PerChainLruCache = StdMutex<LruCache<(ApplicationId, Vec<u8>), Vec<u8>>>;
+/// Per-chain cache state: an LRU map plus the block height at which the cache was
+/// last invalidated. Both are behind the same mutex.
+struct PerChainCache {
+    lru: LruCache<(ApplicationId, Vec<u8>), Vec<u8>>,
+    block_height: BlockHeight,
+}
 
 /// An LRU cache for application query responses, keyed per chain.
 ///
 /// Caches serialized response bytes keyed on `(chain_id, application_id, request_bytes)`.
 /// The entire per-chain cache is invalidated when a `NewBlock` notification arrives.
+///
+/// To prevent a race where a slow query inserts stale data *after* an invalidation,
+/// each insert carries the chain's `next_block_height` at the time the query ran.
+/// If a newer block has since been processed, the insert is silently dropped.
 struct QueryResponseCache {
-    chains: papaya::HashMap<ChainId, PerChainLruCache>,
+    chains: papaya::HashMap<ChainId, StdMutex<PerChainCache>>,
     capacity_per_chain: std::num::NonZeroUsize,
 }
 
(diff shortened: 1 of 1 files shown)


ma2bd:
* Please remove `--no-query-cache` and make this feature deactivated by default
* Don't auto-disable `--long-lived-services` just panic if CLI options are conflicting

linera-service/src/node_service.rs
```diff
@@ -890,6 +898,102 @@ where
...
+        }
+        result
+    }
+
+    fn insert(
+        &self,
+        chain_id: ChainId,
+        app_id: ApplicationId,
+        request: Vec<u8>,
+        response: Vec<u8>,
```
ma2bd, line 997: Should we include the current block height in the cached value?

In case there is a race condition between queries finishing to execute at block N and a notification that invalidates block N, we could be at block N+1 with some data from block N still in the cache. If we include the block height in the cache, at least we won't serve this lagging data. (TBH it is also more semantically correct because only a quadruplet `(chain_id, app_id, height, request)` has a deterministic output.)

ma2bd: (blockheight or block hash by the way)

linera-service/src/node_service.rs
```diff
@@ -1050,13 +1210,55 @@ where
...
+                .await
+                .make_chain_client(chain_id)
+                .await?;
+            client
+                .chain_info()
+                .await
+                .map_err(ChainClientError::from)?
+                .next_block_height
+                .try_sub_one()
+                .unwrap_or(BlockHeight::ZERO)
```
ma2bd: move this to the next `if let Some(cache)` ?

ma2bd: never mind

linera-service/src/node_service.rs
```diff
@@ -1050,13 +1210,55 @@ where
...
+            .flatten();
+
+        // When caching is enabled, get the chain's current block height so we can
+        // use it as part of the cache key (ensures stale-height entries miss).
+        let block_height = if cache.is_some() {
+            let client = self
+                .context
+                .lock()
+                .await
+                .make_chain_client(chain_id)
```
ma2bd: is this expensive?

linera-service/src/node_service.rs
```diff
@@ -1050,13 +1210,55 @@ where
...
+                .await?;
+            client
+                .chain_info()
```
ma2bd: now we're making a query to the chain actor, though

linera-service/src/node_service.rs
ndr-ds: Should be pretty cheap

ma2bd:
Looks great!

linera-service/src/node_service.rs
```diff
@@ -1109,6 +1312,13 @@ where
             .await
             .make_chain_client(chain_id)
             .await?;
+        let block_height = client
+            .chain_info()
+            .await
+            .map_err(ChainClientError::from)?
+            .next_block_height
+            .try_sub_one()
+            .unwrap_or(BlockHeight::ZERO);
```
ma2bd, line 1321: @ndr-ds This looks wrong. We should not be losing information. How about returning `next_block_height` as it is and adjusting the rest?

PR #5544, opened 2026-02-28
Use next_block_height consistently in query response cache

ndr-ds:
## Motivation

Post-merge review from ma2bd on #5538:
`next_block_height.try_sub_one().unwrap_or(BlockHeight::ZERO)` is wrong because it loses
information when `next_block_height` is 0 (a chain with no blocks yet). The
height-based staleness tracking in the query cache should use `next_block_height`
directly rather than converting back and forth.

## Proposal

- Remove the `.try_sub_one().unwrap_or(BlockHeight::ZERO)` in `query_user_application`
and return `next_block_height` as-is.
- Convert `Reason::NewBlock { height }` (the committed block's height) to
`next_block_height` via `height.try_add_one()` at the invalidation call site.
- Rename the `block_height` field/parameters to `next_block_height` throughout
`PerChainCache`, `insert`, and `invalidate_chain` so the semantics are unambiguous.
- Document the inherent staleness window between block commit and async cache
invalidation.

## Test Plan

- CI

### DIFF
--- a/linera-service/src/node_service.rs
+++ b/linera-service/src/node_service.rs
@@ -934,11 +934,11 @@ mod query_cache_metrics {
     });
 }
 
-/// Per-chain cache state: an LRU map plus the block height at which the cache was
-/// last invalidated. Both are behind the same mutex.
+/// Per-chain cache state: an LRU map plus the `next_block_height` at which the cache
+/// was last invalidated. Both are behind the same mutex.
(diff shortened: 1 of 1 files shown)


ma2bd:
I used Claude to find a few more issues and fix them here: https://github.com/linera-io/linera-protocol/pull/5545

PR #5545, opened 2026-02-28
Fix graphql query cache after #5538

ma2bd:
## Motivation

* follow up with #5538 (using Claude to find issues) 
* fix error obsered in pm-infra
```
Error: "GraphQL request failed: Object {\"error\": String(\"Local node operation failed: Blobs not found: [BlobId { blob_type: ChainDescription, hash: 55e65476a855b7283bba7ba5d01ec4341776c683105925cc62e75ba94a9eb514 }]\")}"
```

## Proposal

* fix off by error with block heights
* avoid extra chain actor query --> In the future, we could add one field to `QueryOutcome` but this apparently touches a lot of code.
* fix notifications for chains added to the wallet after startup
* add an environment variable to facilitate manual testing

## Test Plan

CI + tested manually

### DIFF
--- a/linera-service/src/node_service.rs
+++ b/linera-service/src/node_service.rs
@@ -1056,16 +1056,22 @@ impl QueryResponseCache {
         if next_block_height < cache.next_block_height {
             return; // A new block arrived since this query started; discard stale response.
         }
+        // If the chain has advanced since the last cache update, also clear stale entries.
+        // Note: This should not happen if notifications are timely. Also, this only
+        // works when we have a cache miss.
         if next_block_height > cache.next_block_height {
-            // The chain has advanced since the last cache update. Clear stale entries.
-            // Note: This should not happen if notifications are timely and only works if
-            // we had a cache miss.
+            debug!(
+                "Unexpected query cache invalidation for chain {chain_id}:\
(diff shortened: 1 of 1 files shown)


linera-service/src/node_service.rs
```diff
@@ -1018,19 +1078,19 @@ impl QueryResponseCache {
...
         let mutex = pinned.get_or_insert_with(*chain_id, || {
             StdMutex::new(PerChainCache {
                 lru: LruCache::new(capacity),
-                block_height,
+                next_block_height,
             })
         });
         let mut per_chain = mutex.lock().expect("LRU mutex poisoned");
-        per_chain.block_height = block_height;
+        per_chain.next_block_height = next_block_height;
```
afck: Maybe we should check that `next_block_height >= per_chain.next_block_height`. (Not sure if that's guaranteed?)

ma2bd: yes this makes sense

ndr-ds:
Good catch on the subscriptions!

ma2bd:
> Good catch on the subscriptions!

Happy I caught it in time because with the controller infrastructure, it would have exploded to our face spectacularly :)

Issue #5264, opened 2026-01-15
Update to recent nightly Rust compiler


Twey:
For a while we've been stuck on an old Rust version.  To bump it to the most recent rustc at time of writing we need to fix https://github.com/linera-io/linera-protocol/issues/4742 and https://github.com/linera-io/linera-protocol/issues/4203 (to incorporate the fix for https://github.com/wasmerio/wasmer/issues/5610) and update our build flags to deal with https://github.com/rust-lang/rust/pull/147225.

PR #5554, opened 2026-03-02
Update Rust tolchains

### LINKED ISSUE #4742: Applications don't load with Rust 1.87 or later

deuszx:
## Motivation

  https://github.com/linera-io/linera-protocol/issues/5264

  The project has been pinned to Rust 1.86.0 stable / nightly-2025-03-21 for a long time. Bumping to a recent toolchain was blocked by three independent issues:
  - **#4742**: Rust 1.87+ emits Wasm bulk-memory opcodes (prefix `0xFC`) that `linera-wasm-instrument` (backed by `parity-wasm`) could not parse, causing applications to fail to load.
  - **#4203**: Rust 1.89+ removed `__rust_probestack` from `compiler-builtins`, breaking the Wasmer VM build.
  - **rust-lang/rust#147225**: Rust 1.92+ made `-Z wasm_c_abi=spec` a hard error, and atomics no longer auto-enable shared memory for the web target.

  ## Proposal

  Bump stable toolchain to **1.93.1** and nightly to **nightly-2026-03-01**.

  ### Dependency fixes (git branches, until crates.io releases are available)

  - **`linera-wasm-instrument`** → `linera-io/wasm-instrument@bulk-memory-support`: enables bulk-memory support in parity-wasm so it can parse opcode `0xFC`. Also enable the `bulk` feature in `linera-execution` so
  the instrumentation pass handles these opcodes. Fixes #4742.
  - **`linera-wasmer{,-compiler,-compiler-singlepass,-vm}`** → `linera-io/wasmer@fix-rust-probestack`: vendors `__rust_probestack` in `linera-wasmer-vm` using `rustversion` to conditionally provide the symbol only
  on Rust 1.89+. Addresses the Wasmer portion of #4203.
  - **`ruzstd`**: unpin from `=0.8.1` to `0.8` (the build failure on 0.8.2 that motivated the pin is fixed in newer Rust).

  ### Custom `wasm32-mvp` target for testnet_conway compatibility

  Rust 1.93/LLVM 18 enables `bulk-memory`, `bulk-memory-opt`, and `nontrapping-fptoint` as **default** features for `wasm32-unknown-unknown`. All use the `0xFC` opcode prefix, which testnet_conway validators
  reject.

  The standard `-C target-feature=-bulk-memory,...` flag is **silently ignored** by LLVM 18 when these features are part of the target's default feature set. A custom target JSON is the only reliable solution:

  - **`linera-service/wasm32-mvp.json`**: identical to `wasm32-unknown-unknown` except `"features": "-bulk-memory,-bulk-memory-opt,-nontrapping-fptoint"`.
  - **`examples/.cargo/config.toml`**: configures `-Zunstable-options`, `build-std = ["std", "panic_abort"]`, and `json-target-spec = true` so `cargo build --target wasm32-mvp` works in the examples workspace.
  - **`.cargo/config.toml`**: sets `RUST_TARGET_PATH` to `linera-service` so the custom target resolves anywhere in the repo.
  - **`CliWrapper::build_application`**: builds contracts using `+nightly-2026-03-01 --target wasm32-mvp.json -Z build-std -Z json-target-spec`, then copies output from `target/wasm32-mvp/release/` to
  `target/wasm32-unknown-unknown/release/` so downstream code is unaffected. Also installs the nightly toolchain with `rust-src` on first run.
  - **`examples/rust-toolchain.toml`**: changed from a symlink to `rust-toolchain.toml` (stable) to a symlink to `toolchains/nightly/rust-toolchain.toml` (nightly), since `-Z build-std` and `-Z json-target-spec`
  require nightly cargo.
  - Example README files updated to use `--target wasm32-mvp` in build commands.

  ### Clippy / lint fixes

  Various Clippy fixes required by the new toolchain:
  - Replace `format!("{}", x)` with `format!("{x}")` (clippy::uninlined_format_args)
  - Replace deprecated `GenericArray::as_slice()` with slice indexing `&digest[..]` in the `llm` example (sha3 API change)
  - Replace `.index()` with `.enumerate()` where flagged
  - Add `cargo-machete` ignores for `half` in `llm` and `gen-nft` (transitive dependency)
  - Reformat `half` workspace dependency for `taplo` compliance

  ### Web target fix

  Remove `-Z wasm_c_abi=spec` (hard error in Rust 1.92+) and add explicit shared-memory linker flags (`-Wl,--shared-memory`, `-Wl,--max-memory=...`) for the web target, since atomics no longer auto-enable shared
  memory.

  ## Test Plan

  CI

  ## Release Plan

  TBD

  ## Links
  https://github.com/linera-io/linera-protocol/issues/5264


.cargo/config.toml
```diff
@@ -1,6 +1,11 @@
 [target.aarch64-unknown-linux-gnu]
 linker = "aarch64-linux-gnu-gcc"
 
+[env]
+# Makes `--target wasm32-mvp` resolve to linera-service/wasm32-mvp.json
```
Twey, line 5: Rather than using `wasm32-mvp`, let's add two targets that we can distinguish:

- `wasm32-linera-chain` for on-chain applications (contract and service)
- `wasm32-unknown-js-web` for the Web browser (`@linera/client`)

Then we can throw away the `web` features that are currently polluting our codebase.

Twey: (I'll suggest one target for both contract and service, even though it might be cleaner to separate them, because it eases compilation of applications for the end-user)

examples/rust-toolchain.toml
```diff
@@ -1 +1 @@
-../rust-toolchain.toml
\ No newline at end of file
+../toolchains/nightly/rust-toolchain.toml
```
Twey, line 1: I wonder if it's time to migrate wholesale to nightly Rust and forget about stable.  Basically all our codepaths now depend on nightly features except for the CLI build, and having just the one Rust version would simplify our build a lot.  (We can still forbid `#![feature]`s in the core codebase.)

linera-service/src/cli_wrappers/wallet.rs
```diff
@@ -1258,23 +1249,73 @@ impl ClientWrapper {
         name: &str,
         is_workspace: bool,
     ) -> Result<(PathBuf, PathBuf)> {
+        // Write the custom target spec to a temporary file. The spec disables
+        // bulk-memory, bulk-memory-opt, and nontrapping-fptoint — opcodes that
+        // use the 0xFC prefix and are not supported by testnet_conway validators.
+        // Using a custom target JSON (rather than -C target-feature flags) is
+        // necessary because Rust 1.93/LLVM 18 ignores the flag for these features
+        // when they are part of the target's default feature set.
```
Twey, line 1257: Can we not just add the target to the application template (`linera-service/template`)?

linera-service/src/cli_wrappers/wallet.rs
```diff
@@ -1258,23 +1249,73 @@ impl ClientWrapper {
...
+        // no-op if the toolchain and component are already up to date.
+        Command::new("rustup")
+            .args([
+                "toolchain",
+                "install",
+                WASM_NIGHTLY_TOOLCHAIN,
+                "--component",
+                "rust-src",
+            ])
+            .spawn_and_wait_for_stdout()
```
Twey, line 1281: Please let's not rely on the presence of rustup and mutating the user's home directory!  We should specify the relevant features in the toolchain file and if rustup is present it should do the right thing.

linera-service/src/cli_wrappers/wallet.rs
```diff
@@ -1258,23 +1249,73 @@ impl ClientWrapper {
...
+            .await?;
         Command::new("cargo")
-            .current_dir(self.path_provider.path())
+            .current_dir(path)
+            .arg(format!("+{WASM_NIGHTLY_TOOLCHAIN}"))
```
Twey, line 1285: As above, this should also be unnecessary if the toolchain file is set up correctly.

Twey:
This PR is muddied a lot by the warning cleanups.  Can we add them to a separate PR?  `is_multiple_of` might require some conditional-compilation finagling (or I guess just that one can stay here — it's small) but the rest I think is backward-compatible.

deuszx:
Being superseeded by #5646 and #5650

PR #5543, opened 2026-02-28
Speed up lint-check-all-features CI with cargo-hack and partitioning

ndr-ds:
## Motivation

The `lint-check-all-features` job regularly takes 47-50 minutes with a 50-minute
timeout, causing random timeouts in the merge queue. It also depends on a custom fork of
`cargo-all-features`
([ma2bd/cargo-all-features](https://github.com/ma2bd/cargo-all-features), frozen at
upstream v1.9.0 since 2023).

## Proposal

Replace `cargo-all-features` with [`cargo-hack`](https://github.com/taiki-e/cargo-hack)
and split the work into 3 parallel CI jobs:

- **Switch to `cargo-hack`**: The de-facto standard for feature-combination testing in
Rust CI. Used by [Cargo
itself](https://doc.rust-lang.org/cargo/guide/continuous-integration.html), tokio,
hyper, and tower. Installed via `taiki-e/install-action` (prebuilt binary, <1s) instead
of `cargo install --git` from a custom fork.

- **Split into 3 partitions**: `cargo hack check --each-feature --workspace
--all-targets --partition M/3` splits the ~150 check invocations across 3 parallel jobs
(~50 each), reducing wall time from ~46 min to ~15-18 min per job.

- **Add `paths:` filter to `pull_request` trigger**: The workflow only triggers on PRs
that modify the workflow file itself. It still runs unconditionally in the merge queue
(`merge_group`) and on devnet/testnet pushes. This replaces the `if: github.event_name
== 'merge_group'` guard.

- **Reduce timeout from 50 to 30 minutes**: Each partition should finish in ~15-18 min,
so 30 min gives comfortable headroom.

- **Remove `[workspace.metadata.cargo-all-features]`** from `Cargo.toml`: No longer
needed. `cargo-hack --each-feature` is equivalent to `max_combination_size = 1`, and
optional deps are excluded by default (equivalent to `skip_optional_dependencies =
true`).

## Test Plan

- This PR modifies the workflow file, so the job will run on the PR itself (validating
the `paths:` filter).
- Verify that all 3 partitions complete successfully within the 30-minute timeout.

## Release Plan

- These changes should be backported to the latest `testnet` branch

PR #5575, opened 2026-03-02
Restore lint-check-all-features skip behavior on PRs

ndr-ds:
## Motivation

PR #5543 changed the `pull_request` trigger to use a `paths` filter, so the workflow
only fires when the workflow file itself is modified. This causes
`lint-check-all-features` to show as permanently "pending" on all other PRs because the
required status check is never reported to branch protection.

## Proposal

Restore the pattern used before #5543: trigger on all PRs but skip the expensive
partition jobs with `if: github.event_name != 'pull_request'`. The summary job accepts
`skipped` as a passing result, so it reports success immediately without using any
self-hosted runner time.

- Remove `paths` filter from `pull_request` trigger
- Add `if: github.event_name != 'pull_request'` on the partition jobs
- Accept `skipped` as a passing result in the summary job

## Test Plan

- Verify on this PR that `lint-check-all-features` reports as passing (not pending)
- Verify the partition jobs are skipped (no self-hosted runner usage on PRs)
- Merge queue still runs the full check

### DIFF
--- a/.github/workflows/lint-check-all-features.yml
+++ b/.github/workflows/lint-check-all-features.yml
@@ -4,8 +4,6 @@ on:
   push:
     branches: [ 'devnet_*', 'testnet_*' ]
   pull_request:
-    paths:
-      - '.github/workflows/lint-check-all-features.yml'
   merge_group:
   workflow_dispatch:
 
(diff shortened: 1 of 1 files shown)


ndr-ds:
* **#5575** <a href="https://app.graphite.com/github/pr/linera-io/linera-protocol/5575?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/linera-io/linera-protocol/5575?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #5576, opened 2026-03-03
Move CI concurrency groups from workflow level to per-job

ndr-ds:
## Motivation

Re-running a failed CI job currently cancels all other jobs in the same workflow that
are still running, because all jobs share a single workflow-level concurrency group with
`cancel-in-progress: true`. This is especially painful in the `rust.yml` workflow which
has 21 independent jobs — a transient failure in a fast 2-minute lint job forces slow
40-minute test jobs to restart from scratch.

## Proposal

Move the `concurrency` block from the workflow level to each individual job, keyed on
the job name + PR identifier. Since none of the jobs within each workflow depend on each
other (except for `changed-files` gate jobs), each job only needs to cancel a previous
run of *itself*.

Before:
```yaml
concurrency:
group: '${{ github.workflow }} @ ${{ ... }}'
cancel-in-progress: true
```

After (per job):
```yaml
jobs:
some-job:
concurrency:
group: 'some-job @ ${{ ... }}'
cancel-in-progress: true
```

Applied consistently to all 15 workflow files that had workflow-level concurrency (43
jobs total). For jobs with generic names like `test` (in dynamodb.yml, scylladb.yml),
the group key is prefixed with the workflow name to avoid cross-workflow collisions.

| Scenario | Before | After |
|----------|--------|-------|
| New push to PR | All old jobs cancelled at once | Each old job cancelled individually
when its new counterpart starts |
| Re-run a failed job | All still-running jobs cancelled | Only that job is affected |

## Test Plan

- CI

## Release Plan

### DIFF
--- a/.github/workflows/benchmark-test.yml
+++ b/.github/workflows/benchmark-test.yml
@@ -7,11 +7,6 @@ on:
   merge_group:
   workflow_dispatch:
 
-# This allows a subsequently queued workflow run to interrupt previous runs on pull requests
-concurrency:
-  group: '${{ github.workflow }} @ ${{ github.event.pull_request.head.label || github.head_ref || github.run_id }}'
-  cancel-in-progress: true
-
 env:
   CARGO_TERM_COLOR: always
   CARGO_INCREMENTAL: 0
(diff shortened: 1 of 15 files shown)


ndr-ds:
* **#5576** <a href="https://app.graphite.com/github/pr/linera-io/linera-protocol/5576?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/linera-io/linera-protocol/5576?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #5586, opened 2026-03-04
Support custom graphql subscriptions

### LINKED ISSUE #5584: Support arbitrary graphql subscriptions in node service
We could allow users to subscribe to the result of a (Wasm) GraphQL query on a given chain as follows:
* a separate task listens to the chain
* at every new block we call query and (if the result is new) notify listeners of the result.
* only a fixed set of queries should be allowed: they are passed on the CLI of the node (say `linera service --allow-user-subscription "query { foo }" `) at startup

ma2bd:
## Motivation

Users want to subscribe to the result of a Wasm GraphQL query on a chain. Currently the node service only supports subscribing to raw `Notification` events, not to query results.

Fixes #5584

## Proposal

Add support for arbitrary GraphQL subscriptions in the node service. The operator declares allowed queries at startup via a repeatable `--allow-subscription` CLI flag:

```
linera service --port 8080 \
  --allow-subscription 'query CounterValue { value }'
```

The operation name (`CounterValue`) is extracted automatically from the query string.

Clients then subscribe via WebSocket using the new `queryResult` GraphQL subscription field, specifying which chain and application to watch:

```graphql
subscription {
  queryResult(name: "CounterValue", chainId: "e476...", applicationId: "e476...")
}
```

The node service lazily spawns a background watcher per unique `(name, chain_id, application_id)` key. The watcher:
1. Subscribes to chain notifications
2. On each `NewBlock`, re-executes the named query against the application
3. Compares result bytes to the previous result and only pushes when the value changes
4. Cleans up automatically when the last subscriber disconnects

## Test Plan

New integration test + CI

## Release Plan

These changes should be backported to the latest `testnet` branch, then be released in a new SDK.

### DIFF
--- a/linera-service/src/query_subscription.rs
+++ b/linera-service/src/query_subscription.rs
@@ -36,10 +36,10 @@ pub fn parse_allowed_subscription(s: &str) -> anyhow::Result<RegisteredQuery> {
     );
     let rest = rest.trim_start();
     // Extract the operation name: sequence of alphanumeric/underscore chars.
-    let name_end = rest
-        .find(|c: char| !c.is_alphanumeric() && c != '_')
-        .unwrap_or(rest.len());
-    let name = &rest[..name_end];
+    let name = rest
+        .split(|c: char| !c.is_alphanumeric() && c != '_')
+        .next()
+        .unwrap_or_default();
     anyhow::ensure!(
         !name.is_empty(),
         "expected an operation name after 'query', e.g. 'query MyQuery {{ ... }}'"


linera-service/src/cli/main.rs
```diff
@@ -1294,6 +1310,8 @@ impl Runnable for Job {
                     Some(chain_id),
                     context,
                     read_only,
+                    query_subscriptions,
+                    cancellation_token.clone(),
```
MathieuDutSik, line 1321: It is a little strange to clone the `cancellation_token` and then pass it as an argument to the run function.

ma2bd: Isn't that standard?

linera-service/src/cli/command.rs
```diff
@@ -758,6 +758,13 @@ pub enum ClientCommand {
         /// operations. Use this when exposing the service to untrusted clients.
         #[arg(long)]
         read_only: bool,
+
+        /// Allow a named GraphQL subscription query.
+        /// The operation name is extracted from the query string.
+        /// Repeatable.
+        /// Example: `--allow-subscription 'query CounterValue { getCounter { value } }'`
+        #[arg(long = "allow-subscription")]
+        allowed_subscriptions: Vec<String>,
```
afck, line 773: (As an extension to this feature, we could later make it possible to explicitly specify a name, so that you can use the same GraphQL `query` in different configurations.)

linera-service/src/node_service.rs
```diff
@@ -940,6 +973,8 @@ where
     context: Arc<Mutex<C>>,
     /// If true, disallow mutations and prevent queries from scheduling operations.
     read_only: bool,
+    query_subscriptions: Option<Arc<crate::query_subscription::QuerySubscriptionManager>>,
```
afck, line 1204: An empty `QuerySubscriptionsManager` seems negligible. Is it really worth wrapping this in an `Option`?

linera-service/src/query_subscription.rs
```diff
@@ -0,0 +1,284 @@
...
+    anyhow::ensure!(
+        rest.starts_with(char::is_whitespace),
+        "expected whitespace after 'query' keyword"
+    );
+    let rest = rest.trim_start();
+    // Extract the operation name: sequence of alphanumeric/underscore chars.
+    let name_end = rest
+        .find(|c: char| !c.is_alphanumeric() && c != '_')
+        .unwrap_or(rest.len());
+    let name = &rest[..name_end];
```
afck: Maybe simpler with `splitn(1, |c| !c.is_alphanumeric() && c != '_')`

linera-service/src/query_subscription.rs
```diff
@@ -0,0 +1,284 @@
...
+        let query_string = self
+            .get_query(&key.name)
+            .ok_or_else(|| {
+                anyhow::anyhow!("no subscription query registered with name '{}'", key.name)
+            })?
+            .to_string();
+
+        let mut watchers = self.watchers.lock().unwrap();
+
+        // If a watcher already exists and still has capacity, reuse it.
```
afck: Capacity? Is that limited?

linera-service/src/query_subscription.rs
```diff
@@ -0,0 +1,284 @@
...
+                return;
+            }
+
+            *last_result = Some(response_bytes.clone());
+
+            // Parse the response as JSON and broadcast.
+            match serde_json::from_slice::<Value>(&response_bytes) {
+                Ok(value) => {
+                    // Ignore send errors (no receivers).
+                    let _ = sender.send(value);
```
afck: We want to avoid `let _ =`.
(This `send` can fail if there are no receivers left.)

linera-service/tests/linera_net_tests.rs
```diff
@@ -1918,6 +1918,85 @@ async fn test_wasm_end_to_end_counter(config: impl LineraNetConfig) -> Result<()
...
+        .next()
+        .await
+        .context("expected initial query result")??;
+    let initial_value: u64 = serde_json::from_value(initial["data"]["value"].clone())?;
+    assert_eq!(initial_value, original_counter_value);
+
+    // Increment the counter.
+    let mutation = format!("increment(value: {increment})");
+    application.mutate(mutation).await?;
+
```
afck, line 1995: We could first add a block that doesn't change the value, and then assert that there is no further notification before the actually changed value.

linera-service/src/cli/command.rs
ma2bd: What do you mean? Here, the name is "CounterValue".

linera-service/src/node_service.rs
ma2bd: I think it's the runtime cost of the notifier?

linera-service/src/cli/command.rs
afck: Yes, but imagine another app that tracks multiple values with names. Would be good to be able to allow following different ones:

```json
query MultiCounterValue { getCounter(name: "A") { value } }
query MultiCounterValue { getCounter(name: "B") { value } }
```

Or e.g. subscribing to balances of different accounts on the same chain.

ma2bd: These should be named `MultiCounterValueA` and `MultiCounterValueB`

linera-service/src/query_subscription.rs
ma2bd: lol

linera-service/src/query_subscription.rs
ma2bd: added a `debug!`

linera-service/tests/linera_net_tests.rs
ma2bd: done

linera-service/src/query_subscription.rs
```diff
@@ -0,0 +1,286 @@
...
+    let result = {
+        let ctx = context.lock().await;
+        match ctx.make_chain_client(key.chain_id).await {
+            Ok(client) => client.query_application(query, None).await,
+            Err(e) => {
+                warn!(name = %key.name, "failed to create chain client: {e}");
+                return;
+            }
+        }
+    };
```
ndr-ds, line 243: Maybe I'm missing something, but won't holding the lock while querying the application effectively serialize all application queries here?  
Is that ok in this context?

PR #5588, opened 2026-03-04
[testnet] Sparse event chains

afck:
Heavily modified backport of #4466.

## Motivation

When a client subscribes to events from a publisher chain, it downloads all blocks from that chain, not only the blocks that actually contain events in the subscribed streams.   

## Proposal

* New notification type `Reason::NewEvents`, emitted even for preprocessed blocks, like on `main`.
* New listening mode `ListeningMode::EventsOnly(BTreeSet<StreamId>)`: only processes `NewEvents` notifications.
* Sparse block download: On `NewEvents`, walks backwards through the `previous_event_blocks` linked list to fetch event-bearing ancestors.
* Event index tracking: A new `MapView<StreamId, u32>` field appended to `ChainStateView` tracks per-stream _contiguous_ event indices, so `NewEvents` is only emitted when all previous events are present. Handles migration from pre-existing data: if the map is empty, it is initialized when we encounter the first block with events.
* Chain listener: `update_event_subscriptions` now maps publishers to `EventsOnly(streams)` instead of `FollowChain`.

## Test Plan

Added `test_memory_sparse_event_chain`, `test_chain_listener_sparse_event_download` and `test_backward_compatible_view_field_append`.

## Release Plan

- Port some of the additional changes back to `main`.
- First publish a new SDK. New clients should work with existing validators.
- Later (not urgent): update the validators.

## Links

- See #4466 on `main`.

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -1229,7 +1229,7 @@ impl<Env: Environment> Client<Env> {
         let (max_epoch, committees) = self.admin_committees().await?;
 
         let mut certificates = BTreeMap::new();
-        let mut heights_to_fetch = BTreeSet::<_>::from([(hash, height)]);
+        let mut blocks_to_fetch = BTreeSet::<_>::from([(height, hash)]);
         let next_expected_events = subscribed_streams
             .iter()
             .zip(
@@ -1244,7 +1244,7 @@ impl<Env: Environment> Client<Env> {
             )
             .collect::<BTreeMap<_, _>>();
 
-        while let Some((current_hash, current_height)) = heights_to_fetch.pop_last() {
+        while let Some((current_height, current_hash)) = blocks_to_fetch.pop_last() {
             if current_height < local_next_block_height {
                 continue; // Already executed locally.
             }
@@ -1262,7 +1262,7 @@ impl<Env: Environment> Client<Env> {
                     .download_certificates(remote_node, sender_chain_id, current_height, 1)
                     .await?;
                 let Some(certificate) = downloaded.into_iter().next() else {
-                    break;
+                    continue;
                 };
 
                 Client::<Env>::check_certificate(max_epoch, &committees, &certificate)?
@@ -1288,7 +1288,7 @@ impl<Env: Environment> Client<Env> {
                         continue;
                     }
                     if !certificates.contains_key(prev_height) {
-                        heights_to_fetch.insert((*prev_hash, *prev_height));
+                        blocks_to_fetch.insert((*prev_height, *prev_hash));
                     }
                 }
             }


linera-chain/src/chain.rs
```diff
@@ -1073,6 +1078,52 @@ where
...
+            } else {
+                0
+            };
+            let mut current_expected_index = self
+                .next_expected_events
+                .get(&stream_id)
+                .await?
+                .unwrap_or(initial_index);
+            for index in indices {
+                if index == current_expected_index || block.header.height <= next_block_height {
```
bart-linera: Is it meant to update the `next_expected_events` in case it wasn't initialized? I'm a bit confused, because the comment on the method says it expects the index to be initialized. Also, won't this cause a stream to always be considered to be updated if we process an old block for some reason?

linera-core/src/chain_worker/state.rs
```diff
@@ -767,6 +767,41 @@ where
         Ok((self.chain_info_response(), actions, BlockOutcome::Processed))
     }
 
+    /// For each stream in `block.body.events` that has no entry in
+    /// `chain.next_expected_events`, walks backwards through the `previous_event_blocks`
+    /// linked list (reading blocks from storage) to collect locally available event
+    /// indices, then initializes `next_expected_events` to the first gap.
+    ///
+    /// Once a block below `next_block_height` is reached, all earlier events for that
+    /// stream are guaranteed present (the block was fully executed), so the walk stops.
```
bart-linera: This comment doesn't seem to be accurate, or am I missing something?

linera-core/src/client/mod.rs
```diff
@@ -1180,6 +1214,76 @@ impl<Env: Environment> Client<Env> {
         Ok(())
     }
 
+    /// Downloads blocks that contain events in the subscribed streams, walking backwards
+    /// through `previous_event_blocks` to fetch event-bearing ancestors that we don't
+    /// already have locally.
+    /// Downloads event-bearing blocks for the given streams by walking the
+    /// `previous_event_blocks` linked list backwards from `height`, stopping at
+    /// `local_next_block_height` (blocks below that are already executed locally).
+    async fn download_event_bearing_blocks(
```
bart-linera, line 1220: Nice! I may be able to just copy it to `main` to fix the issue with my original PR 😅

linera-core/src/client/mod.rs
```diff
@@ -4146,29 +4254,81 @@ impl<Env: Environment> ChainClient<Env> {
...
+                                local_height,
+                                &subscribed,
+                                &remote_node,
+                            )
+                            .await?;
+                    }
+                } else {
+                    self.client
+                        .synchronize_chain_state_from(&remote_node, chain_id)
+                        .await?;
```
bart-linera, line 4318: I wonder if this won't cause us to download everything (including sender chains) even in the `FollowChain` mode? 🤔  (Not introduced in this PR, anyway, so doesn't have to be fixed right now.)

afck: I don't think so: `synchronize_chain_state_from` synchronizes _this_ chain, not its sender chains.

And in `FollowChain` mode, we do want to download the whole chain, right? We just don't want its sender chains.

linera-chain/src/chain.rs
afck: Sorry, I forgot to update the comments for my latest changes! Will fix.

We initialize `next_expected_events` before we execute or preprocess any block containing events.

So if `block.header.height <= next_block_height`, `index` should in fact always equal `current_expected_index`. I'll address that, too.

linera-core/src/client/mod.rs
```diff
@@ -1180,6 +1214,76 @@ impl<Env: Environment> Client<Env> {
...
+            }
+
+            let downloaded = self
+                .requests_scheduler
+                .download_certificates_by_heights(
+                    remote_node,
+                    sender_chain_id,
+                    vec![current_height],
+                )
+                .await?;
```
bart-linera: After thinking some more... won't this download even the blocks that we already have, if they were only pre-processed? Maybe we should check with storage first, and only download if that fails?

And it seems we'll be traversing the whole sparse part of the chain every time we process a notification about new events, won't we? Which may become expensive if it is long 😬

PR #5600, opened 2026-03-05
[frontport] Fix proxy metrics cardinality explosion from bot traffic

ndr-ds:
## Motivation

The tower middleware added in #5534 extracts `method_name` from the URI path using
`rsplit('/')` for per-method success/error metrics. This works correctly for gRPC
requests, but internet-facing proxies also receive bot/scanner traffic (`.env`,
`wp-login.php`, etc.). Each unique probe path creates a permanent new Prometheus time
series, causing unbounded label cardinality.

On a test network, 325+ unique `method_name` label values were observed within hours,
all from bot probes.

This was discovered and fixed during the backport to `testnet_conway` (#5591).

## Proposal

Replace the naive `rsplit('/')` extraction with a check for the gRPC URI pattern.
gRPC paths have the form `/{package}.{Service}/{Method}` — the first segment always
contains a dot. Non-gRPC requests are bucketed as `non_grpc` instead of recording the
raw path, keeping label cardinality bounded to the number of actual gRPC methods.

## Test Plan

- CI
- Verified on a test network (`andrtest-conway`): after deploying, proxy error
metrics collapsed from 325+ unique labels to just 2 (`non_grpc` + one straggler
from before the fix). gRPC success metrics correctly show PascalCase method names
(`HandleBlockProposal`, `HandleLiteCertificate`, etc.).

### DIFF
--- a/linera-service/src/proxy/grpc.rs
+++ b/linera-service/src/proxy/grpc.rs
@@ -141,14 +141,20 @@ where
         #[cfg(with_metrics)]
         let start = linera_base::time::Instant::now();
 
+        // Extract the gRPC method name from the URI path. gRPC paths have the form
+        // `/{package}.{Service}/{Method}` — the first segment always contains a dot.
+        // Non-gRPC requests (bot probes, health checks, etc.) are bucketed as
+        // "non_grpc" to prevent unbounded label cardinality.
         #[cfg(with_metrics)]
-        let method_name = request
-            .uri()
-            .path()
-            .rsplit('/')
-            .next()
-            .unwrap_or("unknown")
-            .to_owned();
+        let method_name = {
+            let path = request.uri().path();
+            let parts: Vec<&str> = path.splitn(3, '/').collect();
+            if parts.len() == 3 && parts[1].contains('.') {
+                parts[2].to_owned()
+            } else {
+                "non_grpc".to_owned()
+            }
+        };
 
         #[cfg(all(with_metrics, feature = "opentelemetry"))]
         let traffic_type: &'static str = get_traffic_type_from_request(&request);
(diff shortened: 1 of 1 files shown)


ndr-ds:
* **#5600** <a href="https://app.graphite.com/github/pr/linera-io/linera-protocol/5600?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/linera-io/linera-protocol/5600?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

deuszx (approved):

linera-service/src/proxy/grpc.rs
```diff
@@ -141,14 +141,20 @@ where
...
-            .uri()
-            .path()
-            .rsplit('/')
-            .next()
-            .unwrap_or("unknown")
-            .to_owned();
+        let method_name = {
+            let path = request.uri().path();
+            let parts: Vec<&str> = path.splitn(3, '/').collect();
+            if parts.len() == 3 && parts[1].contains('.') {
```
afck, line 152: But someone malicious could still make up any number of method names that way, could they?

ndr-ds: Good point! There might be a more ellegant way to implement this, one sec

ndr-ds: If someone tries a gRPC method call with some crafted method name, tonic will actually check it against the registered methods, and if it doesn't match any, it'll return this `UNIMPLEMENTED` status https://github.com/hyperium/tonic/blob/master/tonic-build/src/server.rs#L172  
Which is status 12 https://grpc.github.io/grpc/core/md_doc_statuscodes.html  
So we can check the gRPC status to make sure that only the gRPC methods we have registered get logged as `method_name`

PR #5601, opened 2026-03-05
Make jemalloc memory profiling opt-in via --enable-memory-profiling flag

ndr-ds:
## Motivation

Memory profiling with jemalloc requires `prof:true` in `MALLOC_CONF`, but enabling it at
startup triggers a known livelock (jemalloc/jemalloc#2282). The workaround is to
activate profiling after initialization via `jemalloc_pprof`, but previously this was
only wired up in the node service, not in the proxy, server, exporter, or faucet
binaries.

Additionally, the activation logic was duplicated as ~15-line blocks across every call
site.

## Proposal

- Add `--enable-memory-profiling` CLI flag to all service binaries: `linera`,
`linera-proxy`, `linera-server`, `linera-exporter`, and the faucet.
- Deduplicate the activation logic into `MemoryProfiling::try_activate()` and
`start_metrics_with_profiling()` in `linera-metrics/src/monitoring_server.rs`.
- Each binary's `Options` struct exposes an `enable_memory_profiling()` accessor that
handles the `#[cfg(feature = "jemalloc")]` dance in one place.
- The server binary handles activation once before spawning per-shard metrics servers.

Replaces #5209.

## Test Plan

- CI
- Manual testing: run any binary with `--enable-memory-profiling` and verify the warning
appears when `MALLOC_CONF` is not set, and profiling activates when it is.

## Release Plan

- These changes should be backported to the latest `testnet` branch.

## Links

- Replaces #5209
- jemalloc livelock: jemalloc/jemalloc#2282

PR #5602, opened 2026-03-05
Use libunwind for jemalloc memory profiling stack traces

ndr-ds:
## Motivation

The default libgcc DWARF unwinder (`_Unwind_Find_FDE`) used by jemalloc for profiling
stack traces has a known livelock
([jemalloc/jemalloc#2282](https://github.com/jemalloc/jemalloc/issues/2282)) — one
thread hangs in `_Unwind_Find_FDE` and all others block on futex_wait. We've hit this in
production on validator shards.

## Proposal

Switch jemalloc's stack unwinder to libunwind by building with
`--enable-prof-libunwind`.

- Point `tikv-jemallocator` at
[linera-io/jemallocator](https://github.com/linera-io/jemallocator) fork which adds a
`profiling_libunwind` feature (upstream doesn't support it yet —
[tikv/jemallocator#146](https://github.com/tikv/jemallocator/issues/146))
- Enable `profiling_libunwind` in the `jemalloc` feature
- Add `libunwind-dev` (builder) and `libunwind8` (runtime) to `docker/Dockerfile`

## Test Plan

- CI
- Verify memory profiling works end-to-end in a temporary network deployment

## Release Plan

- These changes shobe backported to `testnet_conway`.

## Links

- Fork PR: https://github.com/linera-io/jemallocator/pull/1
- Upstream issue:
[tikv/jemallocator#146](https://github.com/tikv/jemallocator/issues/146)
- Jemalloc livelock:
[jemalloc/jemalloc#2282](https://github.com/jemalloc/jemalloc/issues/2282)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -152,7 +152,6 @@ indexed-db = "0.4.2"
 insta = "1.36.1"
 is-terminal = "0.4.12"
 itertools = "0.14.0"
-jemalloc_pprof = "0.8.1"
 js-sys = "0.3.70"
 k256 = { version = "0.13.4", default-features = false, features = [
     "ecdsa",
@@ -163,10 +162,13 @@ k256 = { version = "0.13.4", default-features = false, features = [
 ] }
 k8s-openapi = { version = "0.21.1", features = ["v1_28"] }
 kube = "0.88.1"
+linera-jemalloc-ctl = { version = "0.6.1", features = ["use_std"] }
+linera-jemallocator = "0.6.1"
 linera-kywasmtime = "0.1.0"
 linked-hash-map = "0.5.6"
 log = "0.4.21"
 lru = "0.15.0"
+mappings = "0.7.1"
 mini-moka = "0.10.3"
 nonzero_lit = "0.1.2"
 num-bigint = "0.4.3"
@@ -186,6 +188,7 @@ opentelemetry_sdk = { version = "0.30.0", features = ["trace", "rt-tokio"] }
 papaya = "0.2.3"
 pathdiff = "0.2.1"
 port-selector = "0.1.6"
+pprof_util = { version = "0.8.0", features = ["flamegraph", "symbolize"] }
 prettyplease = "0.2.16"
 proc-macro-error = "1.0.4"
 proc-macro2 = "1.0"
@@ -267,7 +270,6 @@ test-log = { version = "0.2.15", default-features = false, features = [
 test-strategy = "0.3.1"
 thiserror = "1.0.65"
 thiserror-context = "0.1.1"
-tikv-jemallocator = "0.6.0"
 tokio = "1.36.0"
 tokio-stream = "0.1.14"
 tokio-test = "0.4.3"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5021,23 +5021,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4a5f13b858c8d314ee3e8f639011f7ccefe71f97f96e50151fb991f267928e2c"
 
 [[package]]
-name = "jemalloc_pprof"
-version = "0.8.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "74ff642505c7ce8d31c0d43ec0e235c6fd4585d9b8172d8f9dd04d36590200b5"
-dependencies = [
- "anyhow",
- "libc",
- "mappings",
- "once_cell",
- "pprof_util",
- "tempfile",
- "tikv-jemalloc-ctl",
- "tokio",
- "tracing",
-]
-
-[[package]]
 name = "jobserver"
 version = "0.1.33"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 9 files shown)


ndr-ds:
* **#5602** <a href="https://app.graphite.com/github/pr/linera-io/linera-protocol/5602?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/linera-io/linera-protocol/5602?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#5601** <a href="https://app.graphite.com/github/pr/linera-io/linera-protocol/5601?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

Cargo.toml
```diff
@@ -257,7 +257,7 @@ test-log = { version = "0.2.15", default-features = false, features = [
 test-strategy = "0.3.1"
 thiserror = "1.0.65"
 thiserror-context = "0.1.1"
-tikv-jemallocator = "0.6.0"
+tikv-jemallocator = { git = "https://github.com/linera-io/jemallocator.git", rev = "16f3e75fb145e12f5ae15ed232d1368d4539fc14" }
```
deuszx: if we use git dependencies we won't be able to publish to crates.

ndr-ds: Ah good point, we might have to publish this fork to crates.io then. I have a PR to hopefully merge this upstream https://github.com/tikv/jemallocator/pull/159, but until that gets merged (if it ever does), we'll need to use our published fork

ndr-ds: Ah nvm, maybe if we just add a `version` it'll work?

Cargo.toml
```diff
@@ -257,7 +257,7 @@ test-log = { version = "0.2.15", default-features = false, features = [
 test-strategy = "0.3.1"
 thiserror = "1.0.65"
 thiserror-context = "0.1.1"
-tikv-jemallocator = "0.6.0"
+tikv-jemallocator = { version = "0.6", git = "https://github.com/linera-io/jemallocator.git", rev = "16f3e75fb145e12f5ae15ed232d1368d4539fc14" }
```
deuszx: Is this not gonna break the crates' publishing step? If there's a version for it, can't we use one from crates?

Cargo.toml
```diff
@@ -409,3 +409,7 @@ ref_option_ref = "deny"
 str_split_at_newline = "deny"
 unnecessary_wraps = "deny"
 used_underscore_binding = "deny"
+
+[patch.crates-io]
+tikv-jemallocator = { git = "https://github.com/linera-io/jemallocator.git", rev = "16f3e75fb145e12f5ae15ed232d1368d4539fc14" }
+tikv-jemalloc-sys = { git = "https://github.com/linera-io/jemallocator.git", rev = "16f3e75fb145e12f5ae15ed232d1368d4539fc14" }
```
deuszx: same question

Cargo.toml
ndr-ds: I don't think so. AFAIK `cargo publish` will strip the git/rev stuff and keep just the version

Cargo.toml
ndr-ds: I believe `cargo publish` will also strip the patches

deuszx: I don't understand then. Either we don't need it or the published crate will contain a different deps than the non-published code?

ndr-ds: So the published crate will use the vanilla jemallocator version without our fork changes. Because SDK users probably don't care about jemalloc and memory profiling? And then our own binaries when we build them will have the fork changes

deuszx: This seems like accidental outcome and not planned 🤔 maybe that's ok.

Cargo.toml
ma2bd: Mmm based on this theory what would happen when someone runs `cargo install`?

Cargo.toml
ma2bd: Can we have a crate `linera-jemallocator`?

Cargo.toml
deuszx: That's where I was getting at – yes :)

ndr-ds: Ah good point, compilation might fail :(   
I might have to just bite the bullet then and publish the fork crate

ndr-ds: Compilation will fail if someone tries to `cargo install` with the `jemalloc` feature, but like I said in the other thread here we could split the profiling into a separate feature, then people could still install binaries with `jemalloc` just not with profiling

Cargo.toml
ndr-ds: So according to Claude's analysis, if we go the published crates route, we need to actually publish 4 crates:  


1. linera-jemalloc-sys — fork of tikv-jemalloc-sys
2. linera-jemalloc-ctl — fork of tikv-jemalloc-ctl, depends on linera-jemalloc-sys
3. linera-jemallocator — fork of tikv-jemallocator, depends on linera-jemalloc-sys
4. linera-jemalloc-pprof — fork of jemalloc_pprof, depends on linera-jemalloc-ctl  
  
If we decide not to publish crates, I could split the profiling into another `jemalloc-profiling` feature or something, and if people try to `cargo install` with profiling, the build will fail. I think that's the tradeoff

ma2bd: sgtm!

ma2bd (approved):

PR #5605, opened 2026-03-05
EVM-to-Linera fungible bridge.

deuszx:
## Motivation

Enable cross-chain asset transfers from EVM-compatible networks to Linera by implementing a trustless EVM→Linera bridge. Users should be able to lock ERC-20 tokens in a Solidity bridge contract and have corresponding wrapped tokens minted on Linera, without relying on a centralized oracle or custodian.

## Proposal

This PR introduces the full EVM→Linera deposit bridge pipeline across four layers:

**Solidity contract (`FungibleBridge.sol`):** Adds a `deposit()` function using `SafeERC20.transferFrom` to lock tokens in the bridge contract. Each deposit emits a `DepositInitiated` event and increments a deposit nonce to ensure ordering. The `IERC20` interface and `MockERC20` test token are extended with `approve`/`transferFrom`/`allowance` support.

**`linera-bridge` crate restructuring:** EVM-related modules are moved under `src/evm/` and the `Cargo.toml` is reorganized with feature gates (`chain`, `offchain`, `testing`, `codegen`, `cli`) so that heavy dependencies are optional depending on the compilation target.

**MPT proof verification and off-chain SDK:** A Wasm-compatible proof verification module is added that performs RLP block header decoding, Merkle Patricia Trie (MPT) inclusion verification, transaction receipt log decoding, and `DepositInitiated` event parsing. An off-chain proof generation SDK with an HTTP client and a new `GenerateDepositProof` CLI command round out the tooling. The `alloy-rlp` and `alloy-trie` crates are added as workspace dependencies.

**`evm-bridge` Wasm application:** A new example Linera application that accepts a deposit proof, verifies the MPT inclusion proof against the EVM chain state, decodes the `DepositInitiated` event, enforces replay protection via nonce tracking, and mints the corresponding wrapped tokens through the existing `wrapped-fungible` app.

## Test Plan

- **Unit tests:** Deposit tests for `FungibleBridge.sol` using the extended `MockERC20` contract; Rust unit tests for MPT proof verification and event decoding.
- **Integration tests (Anvil):** An Anvil-based integration test spins up a local EVM node, deploys the bridge contract, submits a deposit transaction, generates the MPT proof via the off-chain SDK, and verifies it using the on-chain Wasm logic.
- **Docker end-to-end test:** A full EVM→Linera deposit flow test using `docker-compose`, updated to use the Shanghai EVM hardfork and to deploy the `wrapped-fungible` app alongside the bridge. `parse_deployed_address` is extracted into a shared e2e helper to reduce duplication.
- CI runs the full suite on the `evm-bridge` branch.

## Release Plan

    - be released in a new SDK,

## Links

- [[reviewer checklist](https://github.com/linera-io/linera-protocol/blob/main/CONTRIBUTING.md#reviewer-checklist)](https://github.com/linera-io/linera-protocol/blob/main/CONTRIBUTING.md#reviewer-checklist)

### EVM to Linera

```mermaid
sequenceDiagram
      participant User as User (EVM + Linera)
      participant Bridge as FungibleBridge.sol
      participant Relay as Relay Server
      participant BC as Bridge Chain
      participant EvmBridge as evm-bridge App
      participant WF as wrapped-fungible App
      participant UC as User Chain

      User->>Bridge: 1. approve() + deposit(target_chain, app_id, owner, amount)
      Note over Bridge: Locks ERC-20, emits DepositInitiated event

      Relay->>Relay: 2. Generate MPT receipt proof from EVM block
      Note over Relay: block_header_rlp + receipt_rlp + proof_nodes

      Relay->>EvmBridge: 3. ProcessDeposit(proof) on bridge chain
      Note over EvmBridge: Verify block header, MPT proof,<br/>parse DepositInitiated event,<br/>check replay protection

      EvmBridge->>WF: 4. call_application(Mint { target_account, amount })

      alt target_chain == bridge_chain
          Note over WF: Credits target owner locally
      else target_chain != bridge_chain
          WF->>UC: 5. Cross-chain Credit message
          Note over UC: Tokens pending in inbox
          User->>UC: 6. process_inbox()
          Note over UC: Credits target owner
      end

      User->>UC: 7. Query balance
      UC-->>User: 8. Balance reflects minted tokens
```

### Linera to EVM

```mermaid
sequenceDiagram
      participant User as User (Linera)
      participant UC as User Chain
      participant BC as Bridge Chain
      participant Relay as Relay Server
      participant EVM as FungibleBridge.sol

      User->>UC: 1. wrapped-fungible Transfer(Address20, bridge_chain, amount)
      Note over UC: Debits user account

      UC->>BC: 2. Cross-chain Credit message
      Note over BC: Credits Address20 owner

      Note over Relay: NewIncomingBundle notification
      Relay->>BC: 3. process_inbox()
      Relay->>EVM: 4. addBlock(inbox cert)
      Note over EVM: Accepts block (no Burn → no release)

      Relay->>BC: 5. Burn(Address20, amount)
      Note over BC: Debits Address20, tokens destroyed

      Relay->>EVM: 6. addBlock(burn cert)
      Note over EVM: Finds Burn op → releases ERC-20 to Address20

      User->>EVM: 7. ERC-20 balance increased
```

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -109,7 +109,7 @@ jobs:
       run: |
         cargo test --locked -p linera-execution --features wasmtime
 
-  bridge-format-test:
+  bridge-tests:
     needs: changed-files
     if: needs.changed-files.outputs.should-run == 'true'
     runs-on: linera-io-self-hosted-ci


examples/evm-bridge/src/contract.rs
```diff
@@ -0,0 +1,181 @@
...
+            .expect("Failed to load state");
+        EvmBridgeContract { state, runtime }
+    }
+
+    async fn instantiate(&mut self, _argument: ()) {
+        // Validate parameters are present.
+        self.runtime.application_parameters();
+    }
+
+    async fn execute_operation(&mut self, operation: BridgeOperation) -> BridgeResponse {
```
MathieuDutSik: I find the use of response types in smart contracts of limited interest.
This one especially since its only variant is `Ok`. Did you wrote that in analogy with `Fungible` which also has the some non-trivial response types?

examples/evm-bridge/src/contract.rs
```diff
@@ -0,0 +1,181 @@
...
+        match operation {
+            BridgeOperation::ProcessDeposit {
```
MathieuDutSik, line 55: Do you plan to add more operations to the bridge?

examples/evm-bridge/src/contract.rs
```diff
@@ -0,0 +1,181 @@
...
+impl EvmBridgeContract {
+    async fn process_deposit(
+        &mut self,
+        block_header_rlp: &[u8],
+        receipt_rlp: &[u8],
+        proof_nodes: &[Vec<u8>],
+        tx_index: u64,
+        log_index: u64,
+    ) {
+        let params: BridgeParameters = self.runtime.application_parameters();
```
MathieuDutSik: LLM -enerated code has this unfortunate habit of adding type annotations that we typically did not accept in human code.

examples/evm-bridge/src/contract.rs
```diff
@@ -0,0 +1,181 @@
...
+            let mut addr20 = [0u8; 20];
+            addr20.copy_from_slice(&owner_bytes[12..]);
+            AccountOwner::Address20(addr20.into())
+        } else {
+            AccountOwner::Address32(owner_bytes.into())
+        };
+        let amount_u128: u128 = deposit
+            .amount
+            .try_into()
+            .expect("deposit amount exceeds u128");
```
MathieuDutSik: We already have in `linera-base`:
``` rust
impl TryFrom<U256> for Amount {
    type Error = AmountConversionError;
    fn try_from(value: U256) -> Result<Amount, Self::Error> {
        let value = u128::try_from(&value).map_err(|_| AmountConversionError(value))?;
        Ok(Amount(value))
    }
}
```

So, no need to reinvent the wheel.

examples/evm-bridge/src/lib.rs
```diff
@@ -0,0 +1,64 @@
...
+use linera_sdk::linera_base_types::{ApplicationId, ContractAbi, ServiceAbi};
+use serde::{Deserialize, Serialize};
+
+/// Parameters for a deployed bridge instance.
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+pub struct BridgeParameters {
+    /// EVM chain ID of the source chain (e.g. 8453 for Base).
+    pub source_chain_id: u64,
+    /// Address of the FungibleBridge contract on EVM.
+    pub bridge_contract_address: [u8; 20],
```
MathieuDutSik, line 19: I do not like very much the use of `[u8; 20]`. Why not use a type from `allow_primitives` for that? I think `addess` should work.

examples/evm-bridge/src/lib.rs
```diff
@@ -0,0 +1,64 @@
...
+    pub fungible_app_id: ApplicationId,
+    /// ERC-20 token address on the source EVM chain.
+    pub token_address: [u8; 20],
+}
+
+/// Replay-protection key for processed deposits.
+#[derive(Clone, Debug, Eq, Hash, Ord, PartialEq, PartialOrd, Deserialize, Serialize)]
+pub struct DepositKey {
+    pub source_chain_id: u64,
+    pub block_hash: [u8; 32],
```
MathieuDutSik, line 30: Here the `[u8; 32]` has to be replaced by another type. For the preceding `[u8; 20]` there was only one possibility: An Ethereum address. But `[u8; 32]` can stand for different things and so the appropriate type has to be used.

examples/evm-bridge/src/lib.rs
```diff
@@ -0,0 +1,64 @@
...
+/// Operations accepted by the bridge contract.
+#[derive(Debug, Deserialize, Serialize)]
+pub enum BridgeOperation {
+    /// Verify a deposit proof and mint wrapped tokens.
+    ProcessDeposit {
+        block_header_rlp: Vec<u8>,
+        receipt_rlp: Vec<u8>,
+        proof_nodes: Vec<Vec<u8>>,
+        tx_index: u64,
+        log_index: u64,
```
MathieuDutSik, line 44: There are really no better types than that?

examples/evm-bridge/tests/process_deposit.rs
```diff
@@ -0,0 +1,195 @@
...
+                    receipt_rlp: receipt,
+                    proof_nodes,
+                    tx_index,
+                    log_index: 0,
+                },
+            );
+        })
+        .await;
+
+    assert!(result.is_err(), "invalid proof should be rejected");
```
MathieuDutSik, line 240: More failure scenario should be tested.

examples/evm-bridge/src/contract.rs
```diff
@@ -0,0 +1,181 @@
...
+            ChainId::try_from(deposit.target_chain_id.as_slice()).expect("invalid target chain ID");
+        // Detect Address20 vs Address32: if first 12 bytes are zero, treat as Address20.
+        let owner_bytes = deposit.target_account_owner.0;
+        let target_owner = if owner_bytes[..12].iter().all(|&b| b == 0) {
+            let mut addr20 = [0u8; 20];
+            addr20.copy_from_slice(&owner_bytes[12..]);
+            AccountOwner::Address20(addr20.into())
+        } else {
+            AccountOwner::Address32(owner_bytes.into())
+        };
```
MathieuDutSik: We have conversion code in `linera-base` that avoids that.

linera-bridge/src/proof/gen.rs
```diff
@@ -0,0 +1,282 @@
...
+    pub block_header_rlp: Vec<u8>,
+    /// Canonical receipt bytes (EIP-2718 encoded, as stored in the receipts trie).
+    pub receipt_rlp: Vec<u8>,
+    /// MPT proof nodes for the receipt at `tx_index`.
+    pub proof_nodes: Vec<Vec<u8>>,
+    /// Transaction index within the block.
+    pub tx_index: u64,
+    /// Log index within the receipt.
+    pub log_index: u64,
+}
```
MathieuDutSik, line 43: This is the key type. But it is replicated in the other smart contracts, which is unfortunate.

linera-bridge/src/proof/gen.rs
```diff
@@ -0,0 +1,282 @@
...
+            bail!(
+                "receipts root mismatch: computed {receipts_root}, \
+                 header says {}. Receipt encoding may be incorrect.",
+                block.header.inner.receipts_root
+            );
+        }
+
+        Ok(DepositProof {
+            block_header_rlp: header_rlp,
+            receipt_rlp: target_receipt_rlp,
```
MathieuDutSik: Rename your variable upstream

linera-bridge/src/proof/gen.rs
```diff
@@ -0,0 +1,282 @@
...
+            .collect();
+        proof::verify_receipt_inclusion(root, 0, &receipt, &proof_bytes)
+            .expect("proof should verify");
+    }
+
+    #[test]
+    fn test_build_receipt_proof_multiple_receipts() {
+        let receipts: Vec<(u64, Vec<u8>)> = (0..10)
+            .map(|i| {
+                let receipt = build_test_receipt_with_gas(21000 * (i + 1), &[]);
```
MathieuDutSik, line 313: This test uses `build_test_receipt_with_gas` and the preceding `build_test_receipt`. Why so?

linera-bridge/src/proof/gen.rs
```diff
@@ -0,0 +1,282 @@
...
+        let (root, proof_nodes) = proof::build_receipt_proof(&receipts, 2);
+
+        let wrong_receipt = &receipts[3].1;
+        let proof_bytes: Vec<alloy_primitives::Bytes> = proof_nodes
+            .iter()
+            .map(|n| alloy_primitives::Bytes::copy_from_slice(n))
+            .collect();
+        let result = proof::verify_receipt_inclusion(root, 2, wrong_receipt, &proof_bytes);
+        assert!(result.is_err(), "wrong receipt should fail verification");
+    }
```
MathieuDutSik, line 439: More failure scenario should be tested. Here we have just 1.

linera-bridge/tests/anvil_deposit_proof.rs
```diff
@@ -0,0 +1,309 @@
...
+
+    let hex_str = result["contracts"][file_name][contract_name]["evm"]["bytecode"]["object"]
+        .as_str()
+        .expect("bytecode not found in solc output");
+    hex::decode(hex_str).unwrap()
+}
+
+#[tokio::test]
+#[ignore] // Requires `anvil` and `solc`
+async fn test_deposit_proof_generation() {
```
MathieuDutSik: Making the test return `-> Result<()>` would avoid many expect and simplify the code.

linera-bridge/src/proof/mod.rs
```diff
@@ -0,0 +1,1079 @@
...
+
+    let key = receipt_trie_key(tx_index);
+    alloy_trie::proof::verify_proof(receipts_root, key, Some(receipt_rlp.to_vec()), proof_nodes)
+        .map_err(|e| anyhow!("MPT proof verification failed: {e}"))
+}
+
+/// Decodes a receipt's RLP and extracts its logs.
+///
+/// Handles EIP-2718 typed receipts (type byte prefix < 0x80).
+pub fn decode_receipt_logs(receipt_rlp: &[u8]) -> Result<Vec<ReceiptLog>> {
```
MathieuDutSik, line 327: The thing is that the function `decode_receipt_logs` is returning `anyhow::Result` which is only allowed in tests. Fortunately, that is indeed the case: The function is used only in tests.

But that is not apparent here. So, the code used only in tests should be separated from the rest.

linera-bridge/src/solidity/FungibleBridge.sol
```diff
@@ -7,14 +7,31 @@ import "Microchain.sol";
...
 contract FungibleBridge is Microchain {
+    /// Emitted when a user deposits ERC-20 tokens for bridging to Linera.
+    /// The off-chain relayer uses this event (plus an MPT receipt proof) to
+    /// mint the corresponding tokens on the target Linera chain.
+    event DepositInitiated(
+        uint256 source_chain_id,
+        bytes32 target_chain_id,
+        bytes32 target_application_id,
+        bytes32 target_account_owner,
+        address indexed depositor,
```
MathieuDutSik: Maybe the indexed event shoulkd be put in first position?

linera-bridge/src/test_helpers.rs
```diff
@@ -186,6 +173,7 @@ pragma solidity ^0.8.0;
 
 contract MockERC20 {
     mapping(address => uint256) public balanceOf;
+    mapping(address => mapping(address => uint256)) public allowance;
```
MathieuDutSik, line 176: Why put it `public` instead of `private` as done in the implementation by OpenZeppelin?

linera-bridge/tests/e2e/tests/evm_to_linera_bridge.rs
```diff
@@ -0,0 +1,369 @@
...
+        project_name,
+        &compose_file,
+    )
+    .await;
+    let erc20_addr = parse_deployed_address(&erc20_output)?;
+    tracing::info!(%erc20_addr, "MockERC20 deployed");
+
+    let chain_id_bytes32 = format!("0x{chain_id}");
+    let zero_bytes32 =
+        "0x0000000000000000000000000000000000000000000000000000000000000000";
```
MathieuDutSik: Address::ZERO ?

linera-bridge/tests/anvil_deposit_proof.rs
```diff
@@ -0,0 +1,309 @@
...
+    ) external;
+}
+
+/// Compiles a Solidity contract via `solc`, returning deployment bytecode.
+fn compile_contract(source_code: &str, file_name: &str, contract_name: &str) -> Vec<u8> {
+    let dir = tempfile::tempdir().unwrap();
+    let path = dir.path();
+
+    // Write all shared Solidity files so imports resolve
+    for (name, content) in [
```
MathieuDutSik, line 91: We have code for compiling contracts in `linera-execution/src/test_utils/solidity.rs`.
We do not want code duplication.

examples/evm-bridge/src/contract.rs
deuszx: Do you prefer `bool`? or nothing?

examples/evm-bridge/src/contract.rs
deuszx: Maybe, this is evolving, but currently I don't have those plans.

examples/evm-bridge/src/lib.rs
deuszx: Like what? Those are really just bytes (or indices).

linera-bridge/src/proof/gen.rs
deuszx: WDYM? I can see only one `struct DepositProof`

linera-bridge/src/solidity/FungibleBridge.sol
deuszx: it doesn't really matter but I can do it

linera-bridge/src/test_helpers.rs
deuszx: It doesn't matter, it just a test helper.

examples/evm-bridge/tests/process_deposit.rs
deuszx: Any particular ideas?

examples/evm-bridge/src/contract.rs
MathieuDutSik: Nothing.

examples/evm-bridge/src/lib.rs
MathieuDutSik: I want to say that it is preferable to have nesting types rather than raw data like `Vec<u8>`.
Say you could have
```rust
BlockHeaderRlp(Vec<u8>)
ReceiptRlp(Vec<u8>)
Proof(Vec<u8>)
```
which would get us
```rust
ProcessDeposit {
block_header_rlp: BlockHeaderRlp,
receipt_rlp: ReceiptRlp,
proof_nodes: Vec<Proof>,
},
```

Why this is recommended? This is in order to avoid confusion between the types, for example to avoid substitution of `block_header_rlp` and `receipt_rlp`. Such confusions are always a risk in human code, but of course, less so with LLM-generated code.

MathieuDutSik: To give you an idea, in `linera-base` we have a `BlockHeight(u64)` instead of having just `u64` in the code.

linera-bridge/src/proof/gen.rs
MathieuDutSik: What I would have liked is something like
```rust
enum BridgeOperation {
  ProcessDeposit {
    proof: DepositProof,
  }
}
```

Another idea would be to have
```rust
enum BridgeOperation {
  ProcessDeposit {
    block_header_rlp: BlockHeaderRlp,
    receipt_rlp: ReceiptRlp,
    proof_nodes: Vec<Proof>,
    tx_index: u64,
    log_index: u64,
  }
}
```

examples/evm-bridge/tests/process_deposit.rs
MathieuDutSik: Let me list the ways in which the `process_deposit` can fail:
* "invalid block header RLP"
* "receipt inclusion proof failed"
* "log_index {} out of range (receipt has {} logs)"
* "failed to parse DepositInitiated event"
* "source chain ID mismatch"
* "token address mismatch"
* "deposit already processed"
* "failed to insert deposit key"
* "invalid target chain ID"
* "deposit amount exceeds u128"

Should every one of those failure scenaio be tested in the CI? Not necessarily, I am not a big fan of Test Based Development where for every line of code, 3 lines are written. But still only one failure scenario tested does not look like enough.

linera-bridge/src/solidity/FungibleBridge.sol
MathieuDutSik: Ok.

examples/evm-bridge/src/lib.rs
deuszx: The internal linera protocol types and contracts are different things though. It's quite standard for contracts' ABI/API to use the simplest type possible – this gets (de)serialized, integrated from different languages/SDKs and so the simplest possible type is the easiest to represent/integrate.

linera-bridge/src/proof/gen.rs
deuszx: it's just a coinicidence that the input to the proof generator and contract's API use the same set of arguments so I wouldn't be looking for some common abstractions forcefully

examples/evm-bridge/src/lib.rs
deuszx: This has to be encoded when deploying. We don't have `bcs` for alloy types

examples/evm-bridge/src/contract.rs
deuszx: Actually, we don't 🤔 maybe on `main` but not on `testnet_conway`.

Added it.

examples/evm-bridge/src/contract.rs
deuszx: Again – maybe on `main`, not on `testnet_conway`.

Added it.

linera-bridge/src/proof/mod.rs
deuszx: This method is used in e2e tests which are not a test from the Rust pov (i.e. they're not behind the test macro/feature). If I mark it with `#[cfg(test)]` then I won't be able to reuse it in anvil-based test.

linera-bridge/tests/e2e/tests/evm_to_linera_bridge.rs
deuszx: Done in [Address review comments](https://github.com/linera-io/linera-protocol/pull/5605/commits/84ae7714d5f18c0465209fd2a66f53d5602c1e9e)

linera-bridge/src/solidity/FungibleBridge.sol
deuszx: Done in [Address review comments](https://github.com/linera-io/linera-protocol/pull/5605/commits/84ae7714d5f18c0465209fd2a66f53d5602c1e9e)

examples/evm-bridge/src/contract.rs
deuszx: Done in [Address review comments](https://github.com/linera-io/linera-protocol/pull/5605/commits/84ae7714d5f18c0465209fd2a66f53d5602c1e9e)

linera-bridge/src/proof/gen.rs
deuszx: Done in [Address review comments](https://github.com/linera-io/linera-protocol/pull/5605/commits/84ae7714d5f18c0465209fd2a66f53d5602c1e9e)

linera-bridge/src/proof/gen.rs
deuszx: Added a comment to `build_test_receipt_with_gas` in [Address review comments](https://github.com/linera-io/linera-protocol/pull/5605/commits/84ae7714d5f18c0465209fd2a66f53d5602c1e9e). 

Basically, the receipt logs need to carry the gas used as well (when being in the block) so this method just adds "random" gas to them.

linera-bridge/tests/anvil_deposit_proof.rs
deuszx: Done in [Address review comments](https://github.com/linera-io/linera-protocol/pull/5605/commits/84ae7714d5f18c0465209fd2a66f53d5602c1e9e)

examples/evm-bridge/src/contract.rs
deuszx: Done in [Address review comments](https://github.com/linera-io/linera-protocol/pull/5605/commits/84ae7714d5f18c0465209fd2a66f53d5602c1e9e)

examples/evm-bridge/src/contract.rs
deuszx: Done in [Address review comments](https://github.com/linera-io/linera-protocol/pull/5605/commits/84ae7714d5f18c0465209fd2a66f53d5602c1e9e)

examples/evm-bridge/src/contract.rs
deuszx: Done in [Address review comments](https://github.com/linera-io/linera-protocol/pull/5605/commits/84ae7714d5f18c0465209fd2a66f53d5602c1e9e)

linera-bridge/tests/anvil_deposit_proof.rs
deuszx: The solidity compilation utils in `linera-execution/src/test_utils/solidity.rs` are gated behind `#[cfg(with_revm)]` (which requires the revm feature). Using them from `linera-bridge` tests would mean either pulling in the entire `revm` crate as a `dev-dependency`, or restructuring `linera-execution` to move the `solc` helpers out of the `with_revm` gate — since the compilation functions themselves don't actually use `revm`.

MathieuDutSik: Ok, then change the code, move it and adjust to your needs.
I see no justification for having two separate code for doing compilations.

deuszx: In a separate PR. Restructuring `linera-execution` does not fit here.

linera-bridge/src/proof/mod.rs
MathieuDutSik: Well, ok, test features can be painful to deal with.
But one thing you can do fairly easily is to move the test code in a separate file from `linera-bridge/src/proof/mod.rs`.

linera-bridge/tests/anvil_deposit_proof.rs
deuszx: And it's not so easy either: solidity compilation helpers live in `linera_execution::test_utils::solidity`, which requires both `with_revm` (for the solidity submodule) and `with_testing` (for the `test_utils` parent module). Enabling `linera-execution/test` from `linera-bridge` dev-deps triggers workspace feature unification that breaks `linera-storage`: the test feature adds `add_blobs`/`add_events` to the `ExecutionRuntimeContext` trait, and `linera-storage`'s impl of that trait doesn't include them without its own test feature — forcing `linera-bridge` to also pull in `linera-storage = { features = ["test"] }` for something it otherwise has no relationship with.

Deduplicating cleanly would require restructuring how test_utils is feature-gated, which is out of scope here.

MathieuDutSik: Ok.

linera-bridge/src/proof/gen.rs
deuszx: https://github.com/linera-io/linera-protocol/issues/5639

PR #5606, opened 2026-03-05
Implement a random source for smart contracts.

MathieuDutSik:
## Motivation

Some applications execution gain from having a random number generator that is deterministic and can be used for queries.
It is better for 

## Proposal

We add a function to the `BaseRuntime` and to the `linera-sdk`.
```rust
let seed = self.runtime.random_number();
```

The seed is a u64, and we can use it as seed for a `StdRng::seed_from_u64`.

Note:
* There are design choices here: We could have a fully non-deterministic RNG in the service if that helps. However, one of the use of this functionality is for caching. If we are caching entries that are the ones that we want anyway accross validators, then that should be fine.
* The generation uses the ChainId, ApplicationId, block height, and the count. Possibly other parameters could have been put in.

## Test Plan

A test case has been introduced in `linera-sdk/tests/fixtures`.

## Release Plan

It could be backported to `testnet_conway`.

## Links

None.


ma2bd (commented):
So this has been discussed before. I'm not sure we need a new API -- until we have a secure implementation.

MathieuDutSik:
> So this has been discussed before. I'm not sure we need a new API -- until we have a secure implementation.

In my view such an implementation should not provide any cryptographic security. The `u64` is already a give-away for that.
It should simply provide random numbers that can be used for algorithmic purposes.

MathieuDutSik:
Closed since this is not the right design.

PR #5594, opened 2026-03-04
Log method name on more metrics

ndr-ds:
## Motivation

The middleware-level latency and count metrics (`PROXY_REQUEST_LATENCY`,
`PROXY_REQUEST_COUNT`, `SERVER_REQUEST_LATENCY`, `SERVER_REQUEST_COUNT`) only have a
`traffic_type` label but no `method_name`. This makes it impossible to break down
latency/throughput by gRPC method at the middleware level. Meanwhile,
`SERVER_REQUEST_LATENCY_PER_REQUEST_TYPE` exists as a separate metric that
duplicates `SERVER_REQUEST_LATENCY` but with `method_name` added — this is
redundant.

## Proposal

- Add `method_name` label to all 4 middleware-level latency/count metrics in both
the proxy and server.
- Remove the now-redundant `SERVER_REQUEST_LATENCY_PER_REQUEST_TYPE` metric —
`SERVER_REQUEST_LATENCY` now carries both `method_name` and `traffic_type`.
- Simplify `log_request_outcome_and_latency` → `log_request_outcome` in the server
handlers, since latency is tracked at the middleware level. Remove the per-handler
`Instant::now()` calls.
- Use the cardinality-safe method name extraction from #5571 — validates the gRPC
path format and buckets non-gRPC requests as `non_grpc`.

## Test Plan

CI

## Release Plan

- These changes should be backported to the latest `testnet` branch.

## Links

- Supersedes #5213

PR #5609, opened 2026-03-05
Log method name on more metrics

ndr-ds:
## Motivation

The middleware-level latency and count metrics (`PROXY_REQUEST_LATENCY`,
`PROXY_REQUEST_COUNT`, `SERVER_REQUEST_LATENCY`, `SERVER_REQUEST_COUNT`) only have a
`traffic_type` label but no `method_name`. This makes it impossible to break down
latency/throughput by gRPC method at the middleware level. Meanwhile,
`SERVER_REQUEST_LATENCY_PER_REQUEST_TYPE` exists as a separate metric that
duplicates `SERVER_REQUEST_LATENCY` but with `method_name` added — this is
redundant.

## Proposal

- Add `method_name` label to all 4 middleware-level latency/count metrics in both
the proxy and server.
- Remove the now-redundant `SERVER_REQUEST_LATENCY_PER_REQUEST_TYPE` metric —
`SERVER_REQUEST_LATENCY` now carries both `method_name` and `traffic_type`.
- Simplify `log_request_outcome_and_latency` → `log_request_outcome` in the server
handlers, since latency is tracked at the middleware level. Remove the per-handler
`Instant::now()` calls.
- Use the cardinality-safe method name extraction from #5571 — validates the gRPC
path format and buckets non-gRPC requests as `non_grpc`.

## Test Plan

CI

## Release Plan

## Links

- Supersedes #5213

### DIFF
--- a/linera-rpc/src/grpc/server.rs
+++ b/linera-rpc/src/grpc/server.rs
@@ -10,11 +10,9 @@ use std::{
 use futures::{
     channel::mpsc, future::BoxFuture, stream::FuturesUnordered, FutureExt as _, StreamExt as _,
 };
-use linera_base::{
-    data_types::Blob,
-    identifiers::ChainId,
-    time::{Duration, Instant},
-};
+#[cfg(with_metrics)]
+use linera_base::time::Instant;
+use linera_base::{data_types::Blob, identifiers::ChainId, time::Duration};
 use linera_core::{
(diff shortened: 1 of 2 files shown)


ndr-ds:
This was approved on https://github.com/linera-io/linera-protocol/pull/5594, so will merge

PR #5611, opened 2026-03-05
Allow daily claims from the faucet

### LINKED ISSUE #5556: Daily airdrops from the faucet.
Allow claiming another (configurable) amount every 24 hours for any chain created by the faucet.

afck:
## Motivation

Users that claim a chain from the faucet should receive tokens daily, so they can use the testnet indefinitely and pay the fees.

## Proposal

Add daily claims, similar to `pm-faucet`.

## Test Plan

A unit test was added.

## Release Plan

- Update the faucet.

## Links

- Closes #5556.

### DIFF
--- a/linera-faucet/server/src/lib.rs
+++ b/linera-faucet/server/src/lib.rs
@@ -374,6 +374,26 @@ fn current_daily_period(initial_claim_micros: u64, now_micros: u64) -> u64 {
     now_micros.saturating_sub(initial_claim_micros) / DAILY_PERIOD_MICROS
 }
 
+/// Executes a future and records its latency in [`metrics::CLAIM_LATENCY`], labeled by outcome.
+async fn record_claim_latency<T>(
+    future: impl std::future::Future<Output = Result<T, Error>>,
+) -> Result<T, Error> {
+    #[cfg(with_metrics)]
+    let start_time = std::time::Instant::now();
+
+    let result = future.await;
+
+    #[cfg(with_metrics)]
+    {
+        let label = if result.is_ok() { "success" } else { "error" };
+        metrics::CLAIM_LATENCY
+            .with_label_values(&[label])
+            .observe(start_time.elapsed().as_secs_f64() * 1000.0);
+    }
+
+    result
+}
+
 #[async_graphql::Object(cache_control(no_cache))]
 impl<S> MutationRoot<S>
 where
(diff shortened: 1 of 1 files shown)


linera-faucet/server/src/lib.rs
```diff
@@ -310,6 +402,26 @@ where
...
+
+        let result = self.do_daily_claim(owner).await;
+
+        #[cfg(with_metrics)]
+        {
+            let label = if result.is_ok() { "success" } else { "error" };
+            metrics::CLAIM_LATENCY
+                .with_label_values(&[label])
+                .observe(start_time.elapsed().as_secs_f64() * 1000.0);
+        }
```
MathieuDutSik: This pattern is quite unfortunate and we have something better with `ActiveMeasurementGuard`.

linera-faucet/server/src/database.rs
```diff
@@ -19,22 +19,40 @@ use sqlx::{
...
+    pub timestamp: Timestamp,
+}
+
 /// Schema for creating the chains table.
 const CREATE_CHAINS_TABLE: &str = r#"
 CREATE TABLE IF NOT EXISTS chains (
     owner TEXT PRIMARY KEY NOT NULL,
     chain_id TEXT NOT NULL,
-    created_at DATETIME DEFAULT CURRENT_TIMESTAMP
+    created_at INTEGER NOT NULL
```
MathieuDutSik, line 41: Why this loss of accuracy in the accuracy?

linera-faucet/client/src/lib.rs
```diff
@@ -140,6 +163,63 @@ impl Faucet {
...
+    /// Returns the last claim for the given owner, if any.
+    pub async fn last_claim(
+        &self,
+        owner: &linera_base::identifiers::AccountOwner,
+    ) -> Result<Option<LastClaim>, Error> {
+        #[derive(serde::Deserialize)]
+        #[serde(rename_all = "camelCase")]
+        struct Response {
+            last_claim: Option<LastClaim>,
+        }
```
MathieuDutSik, line 194: Could this type definition be put somewhere else than in a function call?

linera-faucet/server/src/lib.rs
```diff
@@ -286,6 +326,58 @@ where
...
+
+    /// Returns the earliest time at which the owner can make a daily claim.
+    /// If the returned timestamp is in the past (or now), the user can claim immediately.
+    /// Returns `None` if the user has not yet completed the initial claim.
+    async fn next_daily_claim(&self, owner: AccountOwner) -> Result<Option<Timestamp>, Error> {
+        let initial_claim = match self
+            .faucet_storage
+            .last_claim(&owner)
+            .await
+            .map_err(|e| Error::new(e.to_string()))?
```
MathieuDutSik: I find those error conversions unfortunate.
I think it cannot be avoided, but could it be put in `database.rs`? 
I think this can reduce the code size since we see this error conversion repeatedly.

linera-faucet/server/src/tests.rs
```diff
@@ -726,3 +734,125 @@ async fn test_blockchain_sync_after_database_deletion() {
...
+
+    let storage_builder = MemoryStorageBuilder::default();
+    let keys = InMemorySigner::new(None);
+    let clock = storage_builder.clock().clone();
+    clock.set(Timestamp::from(0));
+    let mut builder = TestBuilder::new(storage_builder, 4, 1, keys).await.unwrap();
+    let client = builder
+        .add_root_chain(1, Amount::from_tokens(100))
+        .await
+        .unwrap();
```
MathieuDutSik: The `.unwrap()` can be replaced by returning `anyhow::Result`.

linera-faucet/server/src/database.rs
afck: I think it's better if we take the block timestamp rather than the "local write to faucet database" one. And that is in microseconds since the epoch.

linera-faucet/client/src/lib.rs
afck: It's already the pattern in this file, so I'd prefer to not change it in this PR.

(Also, personally I don't think it's that bad: they're really only used in those single functions.)

linera-faucet/server/src/lib.rs
```diff
@@ -206,18 +221,43 @@ pub struct Validator {
...
 #[derive(Debug)]
 struct PendingRequest {
     owner: AccountOwner,
-    responder: oneshot::Sender<Result<ChainDescription, Error>>,
+    /// For daily claims, the existing chain to transfer tokens to.
+    target_chain_id: Option<ChainId>,
+    /// The amount of tokens to send.
+    amount: Amount,
+    /// Whether this is an initial or daily claim.
+    claim_type: ClaimType,
```
bart-linera: Do we need this if `target_chain_id` is only supposed to be `Some` for daily claims? Couldn't we simply distinguish the claims based on whether it is `Some` or `None`?

linera-faucet/server/src/lib.rs
afck: I replaced it in some places but not here, since we use a label that depends on the outcome. I will deduplicate this, at least…

PR #5613, opened 2026-03-05
Fix some comments, style issues, duplication

afck:
## Motivation

Claude found some inaccurate comments, opportunities for deduplication, dead code, typos and style issues.

## Proposal

Fix them.

I didn't remove unused values in the `requests_scheduler`, since I'm not sure if that's still being worked on.

I also didn't fix everything in the chain actor because that's probably getting removed in https://github.com/linera-io/linera-protocol/pull/5502.

## Test Plan

CI

## Release Plan

Some of these could be backported to `testnet_conway`.

## Links

### DIFF
--- a/linera-core/src/chain_worker/state.rs
+++ b/linera-core/src/chain_worker/state.rs
@@ -167,8 +167,14 @@ where
                 callback,
             } => callback
                 .send(
-                    self.stage_block_execution(block, round, &published_blobs)
-                        .await,
+                    self.stage_block_execution_with_policy(
+                        block,
+                        round,
+                        &published_blobs,
+                        BundleExecutionPolicy::Abort,
+                    )
+                    .await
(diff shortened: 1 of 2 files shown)


linera-rpc/src/grpc/conversions.rs
```diff
@@ -613,7 +613,6 @@ impl TryFrom<api::ChainInfoQuery> for ChainInfoQuery {
             request_leader_timeout,
             request_fallback: chain_info_query.request_fallback,
             request_sent_certificate_hashes_by_heights,
-            request_sent_certificate_hashes_in_range: None,
```
deuszx, line 616: We're not using this anymore 🤔 when did it go away? I thought it was some kind of perf imrpovement during sync

afck: I think we just replaced all uses with the one above, where we give a list of heights.

ma2bd (approved):

PR #5618, opened 2026-03-06
Implement the system allowances.

### LINKED ISSUE #4968: Support allowances in system execution

MathieuDutSik:
## Motivation

The native fungible officially supported the allowances but in practice not as the functionality was not present.
This needed to be corrected.

Fixes #4968

## Proposal

Straightforward implementation

## Test Plan

The same framework that supports the allowances for `fungible` also support `native-fungible`.

## Release Plan

Supporting such a functionality in testnet_conway is not exactly impossible but would require some works.

## Links

CI.

### DIFF
--- a/linera-execution/src/system.rs
+++ b/linera-execution/src/system.rs
@@ -762,10 +762,7 @@ where
 
         // Debit from allowance
         let owner_spender = OwnerSpender::new(owner, spender);
-        let allowance = self
-            .allowances
-            .get_mut_or_default(&owner_spender)
-            .await?;
+        let allowance = self.allowances.get_mut_or_default(&owner_spender).await?;
 
         allowance
             .try_sub_assign(amount)
--- a/linera-service/tests/linera_net_tests.rs
+++ b/linera-service/tests/linera_net_tests.rs
@@ -300,6 +300,18 @@ impl FungibleApp {
         amount_transfer: Amount,
         destination: Account,
     ) -> Value {
+        self.try_transfer_from(owner, spender, amount_transfer, destination)
+            .await
+            .unwrap()
+    }
+
+    async fn try_transfer_from(
+        &self,
+        owner: &AccountOwner,
+        spender: &AccountOwner,
+        amount_transfer: Amount,
+        destination: Account,
+    ) -> Result<Value> {
         let mutation = format!(
             "transferFrom(owner: {}, spender: {}, amount: \"{}\", targetAccount: {})",
             owner.to_value(),
@@ -307,7 +319,7 @@ impl FungibleApp {
             amount_transfer,
             destination.to_value(),
         );
-        self.0.mutate(mutation).await.unwrap()
+        self.0.mutate(mutation).await
     }
 }
 
@@ -2551,6 +2563,24 @@ async fn test_wasm_end_to_end_allowances_fungible(
         app2.query_allowance(&owner1, &owner2).await,
         Amount::from_tokens(17)
     );
+
+    let error = app2
+        .try_transfer_from(
+            &owner1,
+            &owner2,
+            Amount::from_tokens(18),
+            Account {
+                chain_id: chain2,
+                owner: owner3,
+            },
+        )
+        .await
+        .unwrap_err();
+    let error = error.to_string();
+    assert!(!error.is_empty());
+    app2.assert_balances(expected_balances).await;
+    app2.assert_allowance(&owner1, &owner2, Amount::from_tokens(17))
+        .await;
 
     // Doing the transfer from owner 1.
     app2.transfer_from(


linera-service/tests/linera_net_tests.rs
```diff
@@ -2394,14 +2401,20 @@ async fn test_wasm_end_to_end_allowances_fungible(config: impl LineraNetConfig)
     let owner3 = client3.keygen().await?;
 
     // Open a chain owned by both clients.
+    // Native fungible needs enough chain balance for the initial account transfers.
+    let initial_balance = if example_name == "native-fungible" {
+        Amount::from_tokens(30)
+    } else {
+        Amount::from_tokens(6)
+    };
```
afck: Why not use 30 in both cases?

examples/native-fungible/src/contract.rs
```diff
@@ -60,8 +60,16 @@ impl Contract for NativeFungibleTokenContract {
...
 
-            FungibleOperation::Approve { .. } => {
-                panic!("Approve operation is not supported by native fungible token")
+            FungibleOperation::Approve {
+                owner,
+                spender,
+                allowance,
+            } => {
+                self.runtime
+                    .check_account_permission(owner)
```
afck, line 69: Is that necessary? It consumes gas and should also be checked by `approve` itself. (Same for `transfer_from` below.)

linera-service/tests/linera_net_tests.rs
MathieuDutSik: Yes.

examples/native-fungible/src/contract.rs
MathieuDutSik: Actually, this is a more complicated problem. See issue https://github.com/linera-io/linera-protocol/issues/4416.

Therefore, I would prefer to leave the check here, if only as a reminder of what is left to resolve.

linera-execution/src/system.rs
```diff
@@ -711,6 +716,81 @@ where
...
+            ExecutionError::UnauthenticatedTransferOwner
+        );
+
+        if amount == Amount::ZERO {
+            return Ok(());
+        }
+
+        let owner_spender = OwnerSpender::new(owner, spender);
+        let total_allowance = self.allowances.get_mut_or_default(&owner_spender).await?;
+        total_allowance.saturating_add_assign(amount);
```
ndr-ds: Are we supposed to follow ERC-20 here? Because if we are, this shouldn't add to it, but replace it: https://eips.ethereum.org/EIPS/eip-20

linera-base/src/identifiers.rs
```diff
@@ -177,6 +177,26 @@ impl std::str::FromStr for Account {
...
+#[derive(Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize, Allocative)]
+pub struct OwnerSpender {
+    /// Account to withdraw from
+    pub owner: AccountOwner,
+    /// Account to do the withdrawing
+    pub spender: AccountOwner,
+}
+
+impl OwnerSpender {
+    /// Creates a new OwnerSpender pair.
```
afck:
```suggestion
    /// Creates a new `OwnerSpender` pair.
```

linera-service/tests/linera_net_tests.rs
```diff
@@ -2561,7 +2567,17 @@ async fn test_wasm_end_to_end_allowances_fungible(config: impl LineraNetConfig)
...
     ];
     app2.assert_balances(expected_balances).await;
-    app2.assert_allowance(&owner1, &owner2, Amount::from_tokens(91))
+    app2.assert_allowance(&owner1, &owner2, Amount::from_tokens(15))
+        .await;
+
+    // Clearing the allowance should remove it entirely.
+    app1.approve(&owner1, &owner2, Amount::ZERO).await;
+    app1.assert_allowance(&owner1, &owner2, Amount::ZERO)
+        .await;
```
afck: Should we test somewhere that trying to `transfer_from` more than the allowance fails?

linera-sdk/src/contract/test_runtime.rs
```diff
@@ -611,6 +611,23 @@ where
...
+    /// allowance.
+    pub fn transfer_from(
+        &mut self,
+        _owner: AccountOwner,
+        _spender: AccountOwner,
+        _destination: Account,
+        _amount: Amount,
+    ) {
+        // No-op in test runtime
+    }
```
afck, line 726: (I guess we could make these work. Might facilitate some tests. But probably the test runtime needs a bit of an overhaul in general at some point.)

linera-execution/src/system.rs
```diff
@@ -711,6 +716,81 @@ where
...
+        ensure!(
+            amount > Amount::ZERO,
+            ExecutionError::IncorrectTransferAmount
+        );
+
+        // Debit from allowance
+        let owner_spender = OwnerSpender::new(owner, spender);
+        let allowance = self
+            .allowances
+            .get_mut(&owner_spender)
```
afck: Maybe `get_mut_or_default`: Then we'd have only one error case below and could save a few lines?

examples/fungible/tests/cross_chain.rs
```diff
@@ -146,3 +146,66 @@ async fn test_bouncing_tokens() {
         Some(initial_amount),
     );
 }
+
+#[tokio::test]
+async fn test_allowance_overwrite() {
```
afck: (Not sure if that deserves a separate test case; could be part of a general "allowances" test?)

linera-execution/src/system.rs
MathieuDutSik: Yes, I changed accordingly.
This is also done that way in Solana.

linera-base/src/identifiers.rs
MathieuDutSik: Ok.

linera-execution/src/system.rs
MathieuDutSik: Yes. Thank you.

linera-sdk/src/contract/test_runtime.rs
MathieuDutSik: Yes, we can. Done.

examples/fungible/tests/cross_chain.rs
MathieuDutSik: Removed.

linera-base/src/identifiers.rs
```diff
@@ -177,6 +177,26 @@ impl std::str::FromStr for Account {
     }
 }
 
+/// A pair of owner and spender accounts for managing allowances.
+#[derive(Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize, Allocative)]
+pub struct OwnerSpender {
```
deuszx, line 182: I disagree about this struct being lifted to an "official" type (from fungible app). Personally, I don't see myself using it. I think it's an unnecessary abstraction that is now "slightly forced" on all implementors.

MathieuDutSik: I am not sure that it is forced on anyone.

As a fact since we have some `OwnerSpender` in `fungible` and `linera-execution`, the type needs to be pretty low in the hierarchy. So, why not in `linera-base`?

deuszx: I meant that it used to be in fungible - which is an "end" app. Now it's in linera-base which effectively becomes part of every code we ship.

PR #5620, opened 2026-03-06
Implement `{Map,Set,Collection}View` variants with count.

MathieuDutSik:
## Motivation

For some applications, we need a variant that keeps track of the count and is fast.

## Proposal

Use Deref/Deref and nest the inner container for that purpose.

## Test Plan

CI.
Tests were added.

## Release Plan

This change could be backported to testnet_conway.

## Links

None.

### DIFF
--- a/linera-views/src/views/count_view.rs
+++ b/linera-views/src/views/count_view.rs
@@ -1,0 +1,557 @@
+// Copyright (c) Zefchain Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Count-tracking variants of map/set/collection views.
+
+use std::{
+    borrow::Borrow,
+    ops::{Deref, DerefMut},
+};
+
+use allocative::Allocative;
+use serde::{de::DeserializeOwned, Serialize};
+
+use crate::{
+    batch::Batch,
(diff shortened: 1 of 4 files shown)


linera-views/src/views/count_view.rs
```diff
@@ -0,0 +1,568 @@
...
+    C: Context,
+    I: Serialize,
+{
+    /// Inserts or resets a value at an index and updates the count in O(1).
+    pub async fn insert<Q>(&mut self, index: &Q, value: V) -> Result<(), ViewError>
+    where
+        I: Borrow<Q>,
+        Q: Serialize + ?Sized,
+    {
+        if !self.map.contains_key(index).await? {
```
ma2bd, line 172: I suspect this can only be acceptable with additional work involving some sort of Blum filter.

MathieuDutSik:
It is quite clear that keeping track of the count entry is very expensive.
A partial solution to that is to use a Bloom filter. However, a bloom filter requires at least two parameters:
* Probability of failure
* Maximal size of the filter.

Those parameters have to be provided and so the usage is quite complex even if some crates implementing the Bloom filter do exist.

So, the branch is closed unless something happens to reconsider this idea.

PR #5621, opened 2026-03-06
Replace selective COPY patterns with COPY . . in all Dockerfiles

ndr-ds:
## Motivation

PR #5592 broke the bridge E2E test because `Dockerfile.bridge` uses a hand-curated list
of COPY commands and stub source files. When `wrapped-fungible` was added as a path
dependency to `linera-bridge`, the Dockerfile wasn't updated, causing `cargo build` to
fail inside Docker with "No such file or directory."

This class of bug is impossible to catch with existing Dockerfile linters (hadolint,
`docker build --check`, conftest — none validate that COPY sources cover the Cargo
dependency graph). And remembering to add/remove things from the Dockerfiles with no
CI to help, is pretty annoying.
The only structural fix is to stop maintaining curated COPY lists entirely.

## Proposal

Replace all selective `COPY <crate> <crate>` patterns (and the ~100 lines of stub
creation in `Dockerfile.bridge`) with a single `COPY . .` in every Dockerfile that
builds from the repo source.

This is safe because:
- All Dockerfiles use multi-stage builds — the final image is `debian:bookworm-slim`
with only the compiled binary. Source code in the builder stage is discarded. Final
image size is unchanged.
- The `.dockerignore` already excludes `.git`, `.github/`, etc. This PR also adds
`**/target` to catch all Cargo build output directories (which can be tens of GB
locally), replacing the previous `scripts/target` and `examples/target` entries.

## Test Plan

CI


deuszx (commented):
Did you measure the build times? Some of those Dockerfiles do certain tricks to speed up and create smaller dockerfiles. Remember those are not only used in CI but also locally. It already takes ~30 minutes to build this stack.

ndr-ds:
It should be basically the same thing, if anything it might be copying a few extra KB of data now (extra code) which is totally fine, and should be very fast

ndr-ds:
> Nice. Can we have the same in pm-app? :)

Yup, I PR’ed them at the same time ha https://github.com/linera-io/pm-app/pull/396

PR #5627, opened 2026-03-09
Try to fix sparse events

ma2bd:
## Motivation

Try to fix sparse events (with a new client and old servers)

## Proposal

This is the result of code exploration with Claude. Claude proposed 3 bug fixes in total but I only kept the 3rd (and most promising) one in the end.

## Test Plan

tested manually with pm-app


afck:
Included in https://github.com/linera-io/linera-protocol/pull/5631.

PR #5638, opened 2026-03-09
Make `read_event` work even if the event is not locally known yet.

afck:
## Motivation

We have a high-priority use case for reading events without having subscribed and seen `UpdateStreams` yet.

## Proposal

Make the client download the publisher chain from height `0` until it finds the missing event, then retry.

## Test Plan

A test was added.
Instead of writing a new example just for this, `UpdateStream` is used manually, without the client having the streams yet.

## Release Plan

- Release a new SDK.

## Links

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -4,7 +4,7 @@
 
 use std::{
     cmp::Ordering,
-    collections::{hash_map, BTreeMap, BTreeSet, HashMap},
+    collections::{hash_map, BTreeMap, BTreeSet, HashMap, HashSet},
     convert::Infallible,
     iter,
     sync::{Arc, RwLock},
@@ -664,7 +664,8 @@ impl<Env: Environment> Client<Env> {
     }
 
     /// Downloads certificates for `chain_id` one at a time, stopping as soon as all
-    /// validators fail to provide the next height (i.e. the chain is exhausted).
+    /// requested events are locally available or all validators fail to provide the
+    /// next height (i.e. the chain is exhausted).
     /// Uses all validators with staggered concurrent requests for each height.
     #[instrument(level = "trace", skip_all, fields(chain_id, num_events = event_ids.len()))]
     async fn download_chain_until_events_found(
@@ -1843,7 +1844,7 @@ impl<Env: Environment> Client<Env> {
         published_blobs: Vec<Blob>,
         policy: BundleExecutionPolicy,
     ) -> Result<(Block, ChainInfoResponse), ChainClientError> {
-        let mut events_downloaded = false;
+        let mut downloaded_events = HashSet::<EventId>::new();
         loop {
             let result = self
                 .local_node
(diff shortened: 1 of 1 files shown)


linera-core/src/client/requests_scheduler/scheduler.rs
```diff
@@ -374,6 +374,43 @@ impl<Env: Environment> RequestsScheduler<Env> {
...
+        let key = RequestKey::Certificates {
+            chain_id,
+            heights: heights.clone(),
+        };
+        let mut peers = peers.to_vec();
+        peers.shuffle(&mut rand::thread_rng());
+        communicate_concurrently(
+            &peers,
+            async move |peer| {
+                self.with_peer(key, peer, move |peer| {
```
deuszx, line 399: There's also `fn with_best` but it hasn't been used yet in prod as there was no real use-case

linera-core/src/client/mod.rs
```diff
@@ -548,6 +639,70 @@ impl<Env: Environment> Client<Env> {
...
+    }
+
+    /// Downloads certificates for `chain_id` one at a time, stopping as soon as all
+    /// validators fail to provide the next height (i.e. the chain is exhausted).
+    /// Uses all validators with staggered concurrent requests for each height.
+    #[instrument(level = "trace", skip_all, fields(chain_id, num_events = event_ids.len()))]
+    async fn download_chain_until_events_found(
+        &self,
+        chain_id: ChainId,
+        event_ids: &[EventId],
```
bart-linera, line 674: `event_ids` seem to be unused at the moment - so we never actually look for them when downloading the chain.

linera-core/src/client/mod.rs
```diff
@@ -548,6 +641,85 @@ impl<Env: Environment> Client<Env> {
...
+            };
+            let Some(batch_info) = self
+                .process_certificates_using_all(&validators, certificates)
+                .await?
+            else {
+                break;
+            };
+            assert!(batch_info.next_block_height > next_height);
+            next_height = batch_info.next_block_height;
+            if self.has_all_events(event_ids).await? {
```
ma2bd, line 703: @bart-linera how about this line?

bart-linera: I could swear I did a search to make sure that `event_ids` doesn't appear anywhere in the function body and that it didn't 😵 I don't know if I was looking at an outdated diff or what happened 😕

linera-core/src/client/mod.rs
bart-linera: ~Nevermind, I was either looking at an old diff or I'm blind 😬~ Or maybe not?

linera-core/src/client/mod.rs
bart-linera: Wait, that's marked as outdated? :thinking: So it looks like _this_ is an old diff :dizzy_face:

linera-core/src/client/mod.rs
```diff
@@ -703,23 +701,10 @@ impl<Env: Environment> Client<Env> {
             };
             assert!(batch_info.next_block_height > next_height);
             next_height = batch_info.next_block_height;
-            if self.has_all_events(event_ids).await? {
-                return Ok(());
-            }
```
ma2bd: ok it's removed here indeed

linera-core/src/client/mod.rs
```diff
@@ -548,6 +639,70 @@ impl<Env: Environment> Client<Env> {
...
+            self.download_chain_until_events_found(chain_id, &chain_event_ids)
+                .await?;
+        }
+        Ok(())
+    }
+
+    /// Downloads certificates for `chain_id` one at a time, stopping as soon as all
+    /// validators fail to provide the next height (i.e. the chain is exhausted).
+    /// Uses all validators with staggered concurrent requests for each height.
+    #[instrument(level = "trace", skip_all, fields(chain_id, num_events = event_ids.len()))]
```
ma2bd, line 670: This code may be silencing the "unused" warnings.

linera-core/src/client/mod.rs
ma2bd: right

linera-core/src/client/mod.rs
ma2bd: I brought it back!

PR #5640, opened 2026-03-09
Move the `fn query_account` to linera-sdk.

### LINKED ISSUE #5345: Make `fungible::query_account` available in `linera-sdk`

MathieuDutSik:
## Motivation

This function is of fairly general interest and so should be moved to `linera-sdk`.

Fixes #5345 

## Proposal

Implement the change directly.

## Test Plan

CI.

## Release Plan

## Links

None.

### DIFF
--- a/linera-sdk/src/test/chain.rs
+++ b/linera-sdk/src/test/chain.rs
@@ -34,7 +34,7 @@ use serde::Serialize;
 use tokio::{fs, sync::Mutex};
 
 use super::{BlockBuilder, TestValidator};
-use crate::{ContractAbi, ServiceAbi};
+use crate::{abis::fungible::FungibleTokenAbi, ContractAbi, ServiceAbi};
 
 /// A reference to a single microchain inside a [`TestValidator`].
 pub struct ActiveChain {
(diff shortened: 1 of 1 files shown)


linera-sdk/src/test/mod.rs
```diff
@@ -37,6 +37,35 @@ pub use self::{
 };
 use crate::{Contract, ContractRuntime, Service, ServiceRuntime};
 
+/// Queries the balance of an account owned by `account_owner` on a specific `chain`.
+#[cfg(with_integration_testing)]
+pub async fn query_account<Abi>(
+    application_id: linera_base::identifiers::ApplicationId<Abi>,
+    chain: &ActiveChain,
```
afck: Would it make sense for this to be a method of `ActiveChain`?

MathieuDutSik: Yes, that makes sense. Thank you.

linera-sdk/src/test/chain.rs
```diff
@@ -762,6 +762,31 @@ impl ActiveChain {
 
         Ok(certificate)
     }
+
+    /// Queries the balance of an account owned by `account_owner` on this chain.
+    pub async fn query_account<Abi>(
+        &self,
+        application_id: ApplicationId<Abi>,
```
afck: (I guess we can't restrict the `Abi`, because we have multiple fungible token apps?)

MathieuDutSik: Correct. We have only `FungibleTokenAbi` now. Let us use this.

PR #5649, opened 2026-03-10
Address authentication issue in `native-fungible`.

### LINKED ISSUE #4416: Native-fungible doesn't work as intended for application owners

MathieuDutSik:
## Motivation

Issue #4416 that is being considered here is about authentication. We have one authentication going on in `native-fungible` contract code:
```rust
                self.runtime
                    .check_account_permission(owner)
                    .expect("Permission for Transfer operation");
```
This one is done with the authenticated_caller_id being equal to the contract that was calling it.

This is followed in the function `fn transfer` of `linera-execution/src/system.rs` by the following:
```rust
            ensure!(
                authenticated_owner == Some(source)
                    || authenticated_application_id.map(AccountOwner::from) == Some(source),
                ExecutionError::UnauthenticatedTransferOwner
            );
```
It looks the same, but it is not: The authenticated caller id in question is this time the one of the `native-fungible` contract.

This means that we cannot expect the `native-fungible` to behave in a way identical to the `fungible` contract despite them having the same API.

Fixes #4416 

## Proposal

It would be very inappropriate to remove the authentication for transfer or to be able to put what should be used.
But we need a degree of choice here. What we have is the `authenticated_owner` and the `authenticated_caller_id`.
The `authenticated_owner` is the first entry in the call stack. The `authenticated_caller_id` is the last entry.

Therefore, it seems appropriate to give the option of choosing the account used for authentication in the call stack.
This is done by having some function `transfer_auth_depth` which specifies the depth at which we do the authentication.
At level 0, this is the `authenticated_caller_id`. At level 1, it is the level before that.

Is this the final design of authentication? Probably not, but it does address the problem.

## Test Plan

CI.
A test has been added in the `crowd-funding` to make it run with the `native-fungible`. The test passes on this branch and fails on `main`.

## Release Plan

We want a token on mainnet, so this could be backported to `testnet_conway`.

## Links

None

### DIFF
--- a/examples/Cargo.lock
+++ b/examples/Cargo.lock
@@ -1807,6 +1807,7 @@ dependencies = [
  "async-graphql",
  "fungible",
  "linera-sdk",
+ "native-fungible",
  "serde",
  "serde_json",
  "tokio",
--- a/examples/crowd-funding/Cargo.toml
+++ b/examples/crowd-funding/Cargo.toml
@@ -14,6 +14,7 @@ serde_json.workspace = true
 [target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]
 fungible = { workspace = true, features = ["test"] }
 linera-sdk = { workspace = true, features = ["test", "wasmer"] }
+native-fungible.workspace = true
 tokio.workspace = true
 
 [[bin]]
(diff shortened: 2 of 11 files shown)


MathieuDutSik:
This approach has the inconvenience of exposing the whole stack for users. This is a major change that we want to avoid.

PR #5650, opened 2026-03-10
Eliminate web feature

deuszx:
## Motivation

  Replace the `web` Cargo feature flag with a custom wasm target (`wasm32-web-unknown`)
  that uses `target_env = "web"` for cfg detection. This eliminates the need to pass
  `features = ["web"]` through the entire dependency tree and makes web-specific code
  selection automatic based on the compilation target.

  ## Proposal

  - Introduce a custom target spec `wasm32-web-unknown.json` that extends
    `wasm32-unknown-unknown` with `"env": "web"`, threading support (`+atomics`,
    `--shared-memory`), and TLS exports.
  - Replace all `cfg(feature = "web")` checks with `cfg(target_env = "web")` via
    `cfg_aliases` in each crate's `build.rs`.
  - Remove the `web` feature from all Cargo.toml files.
  - Use a forked `wasmer-types` with indexmap 2 to resolve a `build-std` dependency
    conflict with custom targets.
  - Add the default `wasm32-unknown-unknown` linker flags (`-z stack-size=1048576`,
    `--stack-first`, `--allow-undefined`, `--no-entry`) to the custom target spec,
    which are not inherited when defining a custom target from scratch.
  - Add COOP/COEP headers to the vitest config for `SharedArrayBuffer` support
    required by the atomics-enabled wasm build.

  **Note on target naming:** The target is named `wasm32-web-unknown` (with `web` in
  the vendor position) as a workaround for `target-lexicon` (used by `wasmer-types`)
  not being able to parse custom OS values. The semantically correct name would be
  `wasm32-unknown-js-web` (arch-vendor-os-env); an upstream Rust PR
  (rust-lang/rust#153686) has been filed to enable this.

  ## Test Plan

  - `cd web && pnpm -r run ci` passes all 5 tests (2 signer + 3 client integration),
    lint, clippy, and typedoc generation.
  - CI web workflow should pass with the existing faucet + validator setup.

  ## Release Plan

  ## Links

  - Upstream Rust PR for proper target naming: rust-lang/rust#153686
  - Wasmer fork with indexmap 2: https://github.com/linera-io/wasmer/tree/wasmer-types-indexmap2

### DIFF
--- a/linera-base/Cargo.toml
+++ b/linera-base/Cargo.toml
@@ -20,17 +20,6 @@ reqwest = ["dep:reqwest"]
 revm = []
 opentelemetry = ["opentelemetry-otlp"]
 test = ["test-strategy", "proptest"]
-web = [
-    "clap",
-    "getrandom/js",
-    "linera-kywasmtime",
-    "rand/getrandom",
-    "rand/std",
-    "rand/std_rng",
-    "tracing-web",
-    "wasm-bindgen-futures",
-    "web-time",
-]
 
 [dependencies]
 allocative.workspace = true
@@ -46,11 +35,9 @@ custom_debug_derive.workspace = true
 derive_more = { workspace = true, features = ["display", "from_str"] }
 ed25519-dalek.workspace = true
 futures.workspace = true
-getrandom = { workspace = true, optional = true }
 hex.workspace = true
 is-terminal.workspace = true
 k256.workspace = true
-linera-kywasmtime = { workspace = true, optional = true }
 linera-witty = { workspace = true, features = ["macros"] }
 prometheus = { workspace = true, optional = true }
 proptest = { workspace = true, optional = true, features = ["alloc"] }
@@ -71,13 +58,18 @@ trait-set.workspace = true
 trait-variant.workspace = true
 tsify.workspace = true
 wasm-bindgen.workspace = true
-wasm-bindgen-futures = { workspace = true, optional = true }
-web-time = { workspace = true, optional = true }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
-clap = { optional = true, workspace = true }
 ruzstd.workspace = true
-tracing-web = { optional = true, workspace = true }
+
+[target.'cfg(target_os = "web")'.dependencies]
+clap.workspace = true
+getrandom = { workspace = true, features = ["js"] }
+linera-kywasmtime.workspace = true
+rand = { workspace = true, features = ["getrandom", "std", "std_rng"] }
+tracing-web.workspace = true
+wasm-bindgen-futures.workspace = true
+web-time.workspace = true
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 chrono.workspace = true
--- a/linera-base/Cargo.toml
+++ b/linera-base/Cargo.toml
@@ -62,7 +62,7 @@ wasm-bindgen.workspace = true
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 ruzstd.workspace = true
 
-[target.'cfg(target_os = "web")'.dependencies]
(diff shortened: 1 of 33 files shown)


linera-base/Cargo.toml
```diff
@@ -71,13 +58,18 @@ trait-set.workspace = true
...
 wasm-bindgen.workspace = true
-wasm-bindgen-futures = { workspace = true, optional = true }
-web-time = { workspace = true, optional = true }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
-clap = { optional = true, workspace = true }
 ruzstd.workspace = true
-tracing-web = { optional = true, workspace = true }
+
+[target.'cfg(target_env = "web")'.dependencies]
```
Twey, line 65: Some of these are Web-specific (e.g. `kywasmtime`) and some are JS-specific (e.g. `getrandom/js`).

linera-service/wasm32-web-unknown.json
Twey, line 1: Shouldn't this be `wasm32-unknown-js-web`?  Target triples are `architecture-vendor-os-env`.

Twey: (Not that JS is exactly an OS — but I reckon it's close enough for the analogy.  Vendors would be for specific browsers I suppose, or maybe specific Wasm interpreters where they're incompatible.)

deuszx: I can't remember now which crate it was but there was an issue where, when using the target triple, the crate failed to compile b/c it didn't recognize `js` as the os. So Claude, after analysing its code (and some trial and error) ended up with this

deuszx: Ah, yes. It was in the commit message:

> - Rename wasm32-unknown-js-web → wasm32-web-unknown so that
    `target-lexicon` (used by wasmer/cranelift) can parse the target name.
    `target-lexicon` has no custom OS support, so "web" goes in the vendor
    position.
> - Use "env": "web" + "os": "unknown" in the target spec instead of
    "os": "web". std has no platform implementation for target_os="web"
    and fails to compile with build-std.

Twey: I think the `env` here is `unknown`.  The _vendor_ in this triple is `web`, not the env.  This is semantic nonsense, though, and Cranelift shouldn't be pulled in on the Web, so I suggest fixing it by working out why Cranelift is aware of this target at all.

For the latter, indeed the `os` has to be `unknown` so as to pull in `dlmalloc` according to [the relevant `Cargo.toml`](https://github.com/wasmerio/wasmer/pull/5999), but there's no need to mess with the vendor instead of keeping the `web` env.

Twey: https://github.com/rust-lang/rust/pull/153686 to allow us to use `wasm32-unknown-js-web`, which I think should otherwise work.

linera-base/Cargo.toml
Twey: Remember we loosely also support (i.e. it's broken right now, but currently isn't too much work to fix) NodeJS, so not every JS environment is a Web environment.

linera-service/wasm32-web-unknown.json
deuszx: Dependencies are:

`linera-web` → enables `wasmer` feature on `linera-execution` → pulls in `linera-wasmer` → depends on `wasmer-types` → depends on `target-lexicon`

So `wasmer-types` unconditionally depends on `target-lexicon` and uses `Triple` for parsing target triples. When compiled for the `web` target, `target-lexicon` tries to parse the target name during compilation.

This can be fixed by either:
1. Changing `wasmer-types` to pull in `target-lexicon` optionally (upstream change)
2. Waiting for your PR to rust-lang
3. keeping the current hack

Issue #5651, opened 2026-03-10
On subscription, notify about recent events.


afck:
When a chain subscribes to an event stream, the client should download the relevant publisher chain blocks and create an `UpdateStreams` operation.

afck:
Done in #5737.

PR #5653, opened 2026-03-10
[testnet] Sparse-sync publisher chains on subscription

afck:
## Motivation

When subscribing to a notification stream, chains should immediately get notified about the most recent events, even if they happened before the subscription.

## Proposal

Use the new `PreviousEventBlocks` endpoint to do a sparse sync for publisher chains on subscription.

## Test Plan

Tests were extended.

## Release Plan

- Port it to `main`. Needs discussion: new endpoint or `ChainInfoQuery` field?
- Hotfix the validators with #5659.
- _Then_ release this in a new SDK.

## Links

- Part of https://github.com/linera-io/linera-protocol/issues/5651, not on `main` yet.


linera-core/src/client/mod.rs
```diff
@@ -2633,31 +2668,76 @@ impl<Env: Environment> ChainClient<Env> {
...
+        }
+        // Always fully sync the admin chain for epoch changes.
+        let admin_chain_id = self.client.admin_chain_id;
+        if admin_chain_id != self.chain_id {
+            self.client.synchronize_chain_state(admin_chain_id).await?;
+        }
+        // For event publisher chains, do a partial sync using previous_event_blocks.
+        let (_, committee) = self.admin_committee().await?;
+        let nodes = self.client.make_nodes(&committee)?;
+        let tasks: Vec<_> = streams_by_chain
```
MathieuDutSik: The type annotations are not great and slowly the code stops following our convention and follows AI conventions.

linera-core/src/client/mod.rs
```diff
@@ -2633,31 +2668,76 @@ impl<Env: Environment> ChainClient<Env> {
...
+    async fn sync_publisher_chain_events(
+        &self,
+        publisher_chain_id: ChainId,
+        stream_ids: BTreeSet<StreamId>,
+        nodes: &[RemoteNode<Env::ValidatorNode>],
+    ) -> Result<(), ChainClientError> {
+        let mut nodes = nodes.to_vec();
+        nodes.shuffle(&mut rand::thread_rng());
+        let result = communicate_concurrently(
+            &nodes,
```
MathieuDutSik: That pattern of converting to vector, shuffling and passing by reference to `communicate_concurrently` gets repeated over and over in the code and would merit a specific function.

linera-core/src/client/mod.rs
```diff
@@ -1508,6 +1510,39 @@ impl<Env: Environment> Client<Env> {
...
+        if previous_blocks.is_empty() {
+            return Ok(());
+        }
+        let initial_blocks = previous_blocks.into_values().collect();
+        let local_height = self
+            .local_node
+            .chain_info(chain_id)
+            .await
+            .map(|info| info.next_block_height)
+            .unwrap_or(BlockHeight::ZERO);
```
MathieuDutSik: Why unwrapping here?

afck: If we don't have a chain info, the local "next block height" is 0.

But you're right, this should only be done for one specific error…

MathieuDutSik: Yes, that was what I meant.
LLM can be problematic with this kind of situations I have noticed.

afck: Done in https://github.com/linera-io/linera-protocol/pull/5653/commits/69ccb1ec1105ea3b7acf6de928e6375f44289a97.

linera-core/src/client/mod.rs
afck: I moved the shuffling inside the function in https://github.com/linera-io/linera-protocol/pull/5653/commits/69ccb1ec1105ea3b7acf6de928e6375f44289a97.

linera-core/src/client/mod.rs
afck: Fixed in https://github.com/linera-io/linera-protocol/pull/5653/commits/69ccb1ec1105ea3b7acf6de928e6375f44289a97.

linera-core/src/client/mod.rs
```diff
@@ -1487,6 +1488,37 @@ impl<Env: Environment> Client<Env> {
...
+            .await?;
+        let initial_blocks = previous_blocks.values().copied().collect();
+        let local_height = match self.local_node.chain_info(chain_id).await {
+            Ok(info) => info.next_block_height,
+            Err(LocalNodeError::InactiveChain(_) | LocalNodeError::BlobsNotFound(_)) => {
+                BlockHeight::ZERO
+            }
+            Err(error) => return Err(error.into()),
+        };
+        self.download_event_bearing_blocks(
```
deuszx, line 1512: Can we perform any kind of _"I already have these events"_ type of filtering?

linera-core/src/client/mod.rs
```diff
@@ -4696,13 +4770,29 @@ impl<Env: Environment> ChainClient<Env> {
...
-                    // downloaded sparsely via NewEvents notifications.
                     if !events_only {
                         let remote_node = RemoteNode { public_key, node };
                         this.client
                             .synchronize_chain_state_from(&remote_node, this.chain_id)
                             .await?;
+                    } else {
+                        // For EventsOnly chains, do a lightweight initial sync:
+                        // query the validator for the latest event-bearing blocks
+                        // for our subscribed streams, then download them sparsely.
```
deuszx, line 4781: This may not be entirely what the clients wants, right? i.e. it's not the case that everyone that listens in `EventsOnly` mode will want to first download all past events as well?

afck: Maybe that needs some discussion. My understanding is that this is the preferred default behavior. (@ma2bd?)

I guess we could make it configurable, but how and on what level? Does the app decide it? Or the user of the client?

deuszx: Let's take the PM-app as an example – when you're a new client that wants to bet, you don't need all past events about old market resolutions and price updates?

afck: That's why the subscriptions are per-generation. Each generation gets a new event stream!

deuszx: OK, but that means app authors will have harder times designing their apps so as to not abuse the default behavior.

afck: So you'd prefer if the default behavior stayed as it is? I.e. you only get notified about events that happened after your client starts listening?

deuszx: Yes. In all frameworks and languages streams work like that – i.e. you only get things after you subscribe.

afck: Hmm… now that we've made it possible in the client to `read_event` even without subscriptions, I guess we could defer the actual downloads anyway.

If I understand @ma2bd correctly, the PM app use case requires that the app gets notified about the _latest_ event index, but that doesn't mean the client has to download all the earlier blocks.

What do you think?

> In all frameworks and languages streams work like that

I'd say the scenario here is a bit different, because there are two points in time where the "subscription happens":
* From the app's point of view, when the subscription is executed as part of a confirmed block. As an app author I'd expect, even based on how other frameworks work, that my app now gets notified about events from now on.
* From the client's point of view: The client(s) making the blocks might shut down and restart, have a network outage or even be down for maintenance for a while. Maybe _that_ shouldn't influence whether the app learns about the latest events?

deuszx: Even if the client restarts it would miss the events that happened while it was down. Althuogh Kafka is a bit different b/c every client maintains the cursor of the last-read event and can resume from there. Maybe that's the semantic that matches more closely? if we have the ability to download events in a pull-based fashion (not push-based) then, as you say, we can simply _download_ what we want (including latest event that had happened before we subscribed).

afck: Makes sense. Yes, I think the crucial thing that's currently missing is that the app isn't even notified, even after the restart, that it missed any events at all. Let's discuss with the team today.

afck: I created #5707 for this.
As discussed, we are going with the full download for now.

ma2bd:
It works!

PR #5659, opened 2026-03-10
[testnet] Add `previous_event_blocks` validator endpoint

afck:
## Motivation

https://github.com/linera-io/linera-protocol/pull/5653 introduces and uses a new endpoint, so it fails the compatibility tests.

## Proposal

Add the new endpoint first, in this separate PR.

## Test Plan

See #5653 for tests that use this endpoint.

## Release Plan

- Hotfix the validators.
- Then merge #5653.

## Links

- Part of #5653 

### DIFF
--- a/linera-base/src/identifiers.rs
+++ b/linera-base/src/identifiers.rs
@@ -502,6 +502,7 @@ pub struct ModuleId<Abi = (), Parameters = (), InstantiationArgument = ()> {
 }
 
 /// The name of an event stream.
+// TODO(#5667): Enforce length limit for stream names.
 #[derive(
     Clone,
     Debug,
--- a/linera-rpc/src/grpc/client.rs
+++ b/linera-rpc/src/grpc/client.rs
@@ -56,6 +56,10 @@ use super::{
     api::{self, validator_node_client::ValidatorNodeClient, SubscriptionRequest},
     transport, GRPC_MAX_MESSAGE_SIZE,
 };
+
+/// Maximum number of stream IDs per `previous_event_blocks` request, to avoid exceeding
+/// the gRPC message size limit in the response. See [`tests::max_stream_ids_fits`].
+pub(crate) const MAX_STREAM_IDS_PER_REQUEST: usize = 10_000;
 #[cfg(feature = "opentelemetry")]
 use crate::propagation::{get_context_with_traffic_type, inject_context};
 use crate::{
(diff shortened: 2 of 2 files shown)


linera-rpc/src/grpc/server.rs
```diff
@@ -1069,6 +1069,63 @@ where
...
+        &self,
+        request: Request<api::PreviousEventBlocksRequest>,
+    ) -> Result<Response<api::PreviousEventBlocksResponse>, Status> {
+        let start = Instant::now();
+        let traffic_type = Self::get_traffic_type(&request);
+        let inner = request.into_inner();
+        let chain_id = inner
+            .chain_id
+            .ok_or_else(|| Status::invalid_argument("missing chain_id"))?
+            .try_into()?;
```
deuszx: I hate that infra/observability code is leaking into domain now :/

linera-rpc/proto/rpc.proto
```diff
@@ -420,3 +426,16 @@ message ChainInfoResponse {
 message BlockHeight {
   uint64 height = 1;
 }
+
+// Request for previous event blocks.
+message PreviousEventBlocksRequest {
+  ChainId chain_id = 1;
+  // bincode-encoded `Vec<StreamId>`
```
deuszx: If you add `message StreamId` then you can write `repeated StreamId = 2`.

Any reason why you took this, low-level representation?

linera-rpc/src/grpc/server.rs
```diff
@@ -1069,6 +1069,63 @@ where
...
+                    true,
+                    "previous_event_blocks",
+                    traffic_type,
+                );
+                let bytes = bincode::serialize(&result).map_err(|err| {
+                    Status::internal(format!("failed to serialize previous_event_blocks: {err}"))
+                })?;
+                Ok(Response::new(api::PreviousEventBlocksResponse {
+                    previous_event_blocks: bytes,
+                }))
```
deuszx: I think we try to write `TryFrom` impls and call its API here rather than manually construct the responses?

linera-rpc/src/grpc/server.rs
afck: What do you mean? Requests have always had a chain ID, so they can be routed to the correct worker, not just for observability.

deuszx: I meant that you have to go through:
1. `traffic_type`
2. `inner`
3. extract `chain_id` from the `inner`.

In an ideal world steps 1 and 2 would be hidden in some kind of a macro. Similarly the `log_outcome_and_return`.

afck: Makes sense, but it's not really something introduced by this PR, is it? Maybe we should address that on `main` first.

deuszx: Definitely. I just thought about it when I saw this code

afck:
> This can potentially return a lot of data – more than a single gRPC message size limit. How do you handle that?

Fixed in https://github.com/linera-io/linera-protocol/pull/5659/commits/9165e63a90e252cb5e091c419ee2d99707bd308f: It now splits it into batches of 10000. (Although subscribing to more than 10000 streams from a single chain probably leads to other issues anyway and maybe isn't a realistic scenario.)

linera-rpc/proto/rpc.proto
afck: Fixed in https://github.com/linera-io/linera-protocol/pull/5659/commits/88cc5f308197d17ee18c80a7afaace02c0d644a7.

linera-rpc/src/grpc/server.rs
afck: Fixed in https://github.com/linera-io/linera-protocol/pull/5659/commits/88cc5f308197d17ee18c80a7afaace02c0d644a7.

PR #5662, opened 2026-03-11
Measure state hash computation latency in microseconds

ndr-ds:
## Motivation

The `state_hash_computation_latency` metric is currently recorded in milliseconds, but
the operation is fast enough that microsecond resolution gives more useful data.

## Proposal

- Switch the histogram from `exponential_bucket_latencies(2000.0)` (ms buckets up to 2s)
to `exponential_bucket_interval(1.0, 2_000_000.0)` (μs buckets from 1μs to 2s)
- Use `measure_latency_us()` instead of `measure_latency()` so values are recorded in
microseconds
- Update the metric description to reflect the new unit

## Test Plan

CI

### DIFF
--- a/linera-chain/src/chain.rs
+++ b/linera-chain/src/chain.rs
@@ -63,8 +63,7 @@ pub(crate) mod metrics {
     use std::sync::LazyLock;
 
     use linera_base::prometheus_util::{
-        exponential_bucket_interval, exponential_bucket_latencies, register_histogram_vec,
-        register_int_counter_vec,
+        exponential_bucket_interval, register_histogram_vec, register_int_counter_vec,
     };
     use linera_execution::ResourceTracker;
     use prometheus::{HistogramVec, IntCounterVec};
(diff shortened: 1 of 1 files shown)


ma2bd:
@ndr-ds We don't hash the state any more on this branch, do we?

ndr-ds:
If the network created from this branch has the proper allow list stuff, then no. But this is just changing the unit to us. If it's enabled, this metric will actually be 2 us or something (because it does nothing). Otherwise, it's in the ms, sometimes hundreds of ms. So it's good to track the full range of possiblities :)

ma2bd:
> If the network created from this branch has the proper allow list stuff, then no. But this is just changing the unit to us. If it's enabled, this metric will actually be 2 us or something (because it does nothing). Otherwise, it's in the ms, sometimes hundreds of ms. So it's good to track the full range of possiblities :)

We should make the "zero hash" policy activated by default in every configuration (production, tests, benchmarks, etc).

deuszx (approved):
I guess this _is_ correct although not going to affect us in practice (i.e. if we know it's 500us vs 1ms).

Issue #5664, opened 2026-03-11
`Unexpected block proposal error` is reachable.


afck:
This can happen if e.g. a notification or background sync updates the current chain while `execute_block` is running.

PR #5666, opened 2026-03-11
Unify client mutex and proposal state.

afck:
## Motivation

`update_from_info` can clear the pending block proposal while `execute_block` is ongoing, because it doesn't lock the `client_mutex`.

## Proposal

Unify both into a single `Arc<Mutex<Option<PendingProposal>>>` per chain. The mutex now both serializes proposals and holds the pending proposal, ensuring all reads/writes are synchronized. Remove `update_from_info` (which only cleared stale proposals), `update_state`, `clone_for_update_unchecked`, and inline `process_certificate` into its call sites.

## Test Plan

CI

This should fix the (rare and hard to reproduce) race condition.

## Release Plan

- Port to `main`.
- Release a new SDK.

## Links

- Fixes https://github.com/linera-io/linera-protocol/issues/5664.

### DIFF
--- a/linera-client/src/client_context.rs
+++ b/linera-client/src/client_context.rs
@@ -317,6 +317,7 @@ where
             name,
             options.chain_worker_ttl,
             options.sender_chain_worker_ttl,
+            options.prioritize_bundles_from.clone().unwrap_or_default(),
             options.to_chain_client_options(),
             options.to_requests_scheduler_config(),
         );
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -277,6 +277,7 @@ impl<Env: Environment> Client<Env> {
         name: impl Into<String>,
         chain_worker_ttl: Duration,
         sender_chain_worker_ttl: Duration,
+        priority_bundle_origins: HashSet<ChainId>,
         options: ChainClientOptions,
         requests_scheduler_config: requests_scheduler::RequestsSchedulerConfig,
     ) -> Self {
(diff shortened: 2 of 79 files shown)


linera-client/src/client_context.rs
```diff
@@ -1006,7 +1006,7 @@ impl<Env: Environment> ClientContext<Env> {
             for chain_client in chain_clients {
                 let info = chain_client.chain_info().await?;
                 let client_owner = chain_client.preferred_owner();
-                let pending_proposal = chain_client.pending_proposal().clone();
+                let pending_proposal = chain_client.pending_proposal().await;
```
MathieuDutSik, line 1012: Here we have client and later chain_client. Maybe unify.

linera-core/src/client/mod.rs
```diff
@@ -3591,21 +3576,22 @@ impl<Env: Environment> ChainClient<Env> {
                     (block, blobs)
                 }
             }
-        } else if let Some(pending_proposal) = pending_proposal {
+        } else if let Some(pending) = proposal_guard.as_ref() {
             // Otherwise we are free to propose our own pending block.
-            let proposed_block = pending_proposal.block;
+            let proposed_block = pending.block.clone();
```
MathieuDutSik, line 3583: Is it possible to simplify this?
We have `proposal_guard` taken as reference, then cloned and finally at the end of the fnction set to `None`. Possibly, we could simplify.

afck: I don't think so: It's crucial that the pending proposal remains until we know we have moved on to the next height. If we removed it here, any error in the rest of the function would leave it empty.

linera-client/src/client_context.rs
afck: Done in https://github.com/linera-io/linera-protocol/pull/5666/commits/f256afaadaaf5ef24ae4338250c110b550ab9a6a.

PR #5668, opened 2026-03-11
Introduce the `save_and_drop` for `RootView`.

### LINKED ISSUE #4976: [nits] Introduce `into_saved(self)` for `RootView`.
So, now we have the pattern in `fn save`.
```rust
let mut batch = Batch::new();
view.pre_save(&mut batch)?;
store.write_batch(batch).await?;
view.post_save();
```

In the functions `fn finalize` of the smart contracts we often have the pattern
```rust
    async fn store(mut self) {
        self.state.save().await.expect("Failed to save state");
    }
```

This means that the `post_save` is done for no reason (but maybe elided away by the compiler). It would make sense to have `into_saved` that does the `pre_save`, `write_batch`, but not the `post_save`.

MathieuDutSik:
## Motivation

The `write_batch` of `fn save()` derive implementation of the `RootView` trait is followed by a `post_save`.
This does not make sense for the smart contract which takes `mut self` as input.

Fixes #4976 

## Proposal

Introduce a `fn save_and_drop(self)` function to the trait and implement it.
It turns out that there is no other use case than the smart contract where we consume a view and write it.

## Test Plan

CI

## Release Plan

It could definitely be backported to `testnet_conway` as it lead to a small gas reduction.

## Links

None.

### DIFF
--- a/examples/amm/src/contract.rs
+++ b/examples/amm/src/contract.rs
@@ -312,8 +312,11 @@ impl Contract for AmmContract {
         }
     }
 
-    async fn store(mut self) {
-        self.state.save().await.expect("Failed to save state");
+    async fn store(self) {
+        self.state
+            .save_and_drop()
+            .await
+            .expect("Failed to save state");
     }
 }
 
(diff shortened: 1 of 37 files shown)


ma2bd:
Definitely interested in a testnet backport!

PR #5669, opened 2026-03-11
Fix benchmark formatting crash and fork PR comment posting

ndr-ds:
## Motivation

The "Instruction Count Benchmarks" workflow fails on every PR that has zero
instruction-count regressions and a baseline to compare against. Additionally, fork PRs
can't post benchmark comments because `GITHUB_TOKEN` is read-only for forks.

Both issues were introduced by #5487 but were masked because that PR had no baseline to
compare against (it was the first run of the workflow).

## Proposal

Two fixes:

1. **Bash empty array bug**: `declare -a REGRESSED_NAMES` doesn't actually initialize
the array. With `set -u`, checking `${#REGRESSED_NAMES[@]}` on an empty
declared-but-uninitialized array triggers "unbound variable". Fixed by using
`REGRESSED_NAMES=()` instead.

2. **Fork PR comment permissions**: Moved comment posting from the main workflow to a
separate `workflow_run`-triggered workflow (`post-benchmark-comment.yml`). The main
workflow now saves the formatted comment as an artifact. The new workflow downloads it
and posts it with write permissions, since `workflow_run` always runs in the context of
the base repository.

## Test Plan

CI

### DIFF
--- a/.github/scripts/format-bench-results.sh
+++ b/.github/scripts/format-bench-results.sh
@@ -172,7 +172,8 @@ fi
 
 # Check for instruction-count regressions (>1% increase)
 if $HAS_CMP; then
-  declare -a REGRESSED_NAMES REGRESSED_PCTS
+  REGRESSED_NAMES=()
+  REGRESSED_PCTS=()
   for ((j = 0; j < COUNT; j++)); do
     pct="${B_IP[j]}"
     if [[ -n "$pct" ]] && awk "BEGIN { exit !($pct > 1) }"; then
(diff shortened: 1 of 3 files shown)


ndr-ds:
* **#5669** <a href="https://app.graphite.com/github/pr/linera-io/linera-protocol/5669?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/linera-io/linera-protocol/5669?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ndr-ds:
We need this for the CI to succeed, so just merging to not block others

PR #5671, opened 2026-03-11
[Testnet conway] clean up async calls after simplification of key-value store view

MathieuDutSik:
## Motivation

Following PR #5446 the `write_batch` function has become sync but we have kept its status as async.
This is incorrect,

## Proposal

Remove the async property of `write_batch`and of other functions that should are in fact sync.

## Test Plan

CI

## Release Plan

On testnet_conway is the point here.

## Links

None.

### DIFF
--- a/linera-core/src/client/requests_scheduler/node_info.rs
+++ b/linera-core/src/client/requests_scheduler/node_info.rs
@@ -66,7 +66,7 @@ impl<Env: Environment> NodeInfo<Env> {
     /// - **Success score**: Directly proportional to EMA success rate
     ///
     /// Returns a score from 0.0 to 1.0, where higher values indicate better performance.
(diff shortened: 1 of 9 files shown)


ma2bd (approved):
nice!

p.s. I fixed the PR title for you

Twey:
I'm a bit confused — how is it possible for `write_batch` (that makes network calls, at least in the case of ScyllaDB) to be sync?

MathieuDutSik:
> I'm a bit confused — how is it possible for `write_batch` (that makes network calls, at least in the case of ScyllaDB) to be sync?

The `write_batch` of `KeyValueStoreView` is indeed sync for the simple reason that the `KeyValueStoreView` does not implement the `WritableKeyValueStore` trait.

So, that is simply a different function from the one you are thinking about.

PR #5672, opened 2026-03-11
Next clippy corrections

### LINKED ISSUE #5573: Enable clippy lints for bugs, perf issues, and API problems
## Enable clippy lints that catch bugs, perf issues, and API problems

These lints catch real issues — potential panics, unnecessary allocations, misleading APIs — but have existing violations that need cleanup first.

### Lints

| Lint | Violations | What it catches |
|------|----------:|----------------|
| `match_on_vec_items` | 1 | Vec indexing in match without bounds check (potential panic) |
| `iter_not_returning_iterator` | 2 | Method named `iter` that doesn't return `Iterator` (API contract violation) |
| `checked_conversions` | 4 | Manual bounds checks that should use `try_from` |
| `needless_collect` | 5 | `.collect::<Vec<_>>()` immediately re-iterated (unnecessary allocation) |
| `unnecessary_wraps` | 6 | Functions that always return `Ok(x)` or `Some(x)` (pointless wrapping) |
| `inefficient_to_string` | 10 | `.to_string()` on `&str` going through Display trait instead of `.to_owned()` |
| `used_underscore_binding` | 13 | `_x` bindings that are actually used (misleading "unused" convention) |
| `from_iter_instead_of_collect` | 24 | `FromIterator::from_iter(x)` instead of idiomatic `.collect()` |
| `or_fun_call` | 26 | `.unwrap_or(String::new())` allocating on the happy path instead of `.unwrap_or_default()` |
| `needless_pass_by_ref_mut` | 61 | `&mut` params that are never mutated (API lies about mutation) |
| `needless_pass_by_value` | 89 | By-value params that are only borrowed (unnecessary cloning at call sites) |

**Total: ~241 violations**

### Why

- `match_on_vec_items`, `checked_conversions`: correctness — prevent panics and incorrect bounds checks
- `needless_collect`, `or_fun_call`, `inefficient_to_string`, `needless_pass_by_value`: performance — avoid unnecessary allocations and copies
- `needless_pass_by_ref_mut`, `unnecessary_wraps`, `iter_not_returning_iterator`: API honesty — signatures should reflect actual behavior
- `used_underscore_binding`, `from_iter_instead_of_collect`: clarity — prevent misleading code patterns

### Work

1. Enable all lints as `warn` in Cargo.toml
2. Fix violations incrementally — most are mechanical and can be fixed with `cargo clippy --fix`. `needless_pass_by_value` and `needless_pass_by_ref_mut` need manual review since they change function signatures. Can be split across 2-3 PRs.
3. Once all violations are resolved and the lints have proven valuable, promote to `deny`

MathieuDutSik:
## Motivation

Clippy is a very useful tool for reducing the complexity of the code and avoiding bad rust code.

Fixes #5573 

## Proposal

Implement a bunch of additional clippy corrections.

## Test Plan

CI

## Release Plan

This could be backported to the testnet_conway if there is demand.

## Links

None.


web/@linera/client/build.bash
```diff
@@ -7,6 +7,7 @@ shopt -s extglob
 cd $(dirname -- "${BASH_SOURCE[0]}")
 
 export RUST_TOOLCHAIN=nightly
+export RUSTFLAGS="-C target-feature=+atomics,+bulk-memory,+mutable-globals,+reference-types --cfg=web_sys_unstable_apis -Z wasm_c_abi=spec${RUSTFLAGS:+ $RUSTFLAGS}"
```
afck: Why is this needed now?

linera-service/src/exporter/storage.rs
```diff
@@ -480,13 +480,15 @@ where
...
     fn flush(&mut self, batch: &mut Batch) -> Result<(), ExporterError> {
-        for (_, value) in self
+        let mut pending_updates = self
             .state_updates_buffer
             .pin()
             .iter()
             .map(|(key, value)| (*key, value.clone()))
-            .collect::<BTreeMap<_, _>>()
-        {
+            .collect::<Vec<_>>();
```
afck: With a `Vec` instead of `BTreeMap`, we don't deduplicate entries anymore. Is that okay here?

linera-execution/src/wasm/wasmer.rs
```diff
@@ -59,9 +59,13 @@ impl WasmContractModule {
...
+        let cached_module = contract_cache
             .get_or_insert_with(contract_bytecode, CachedContractModule::new)
-            .map_err(WasmExecutionError::LoadContractModule)?
+            .map_err(WasmExecutionError::LoadContractModule)?;
+        #[cfg(web)]
+        let (engine, module) = cached_module.create_execution_instance();
+        #[cfg(not(web))]
+        let (engine, module) = cached_module
             .create_execution_instance()
             .map_err(WasmExecutionError::LoadContractModule)?;
```
afck: In that case maybe we should rather silence the warning and keep returning a result even in the `web` version? (Using something like `#[cfg_attr(web, expect(…))]`.)

linera-core/src/client/requests_scheduler/scheduler.rs
```diff
@@ -941,8 +941,8 @@ mod tests {
     }
 
     /// Helper function to create a test result
-    fn test_result_ok() -> Result<Vec<ConfirmedBlockCertificate>, NodeError> {
-        Ok(vec![])
+    fn test_result_ok() -> Vec<ConfirmedBlockCertificate> {
```
afck: Maybe we should just inline this function then?

linera-core/src/client/chain_client/mod.rs
```diff
@@ -1054,41 +1054,44 @@ impl<Env: Environment> ChainClient<Env> {
...
-                            error!(
-                                ?blob_ids,
-                                %error,
-                                "Error while attempting to download blobs during retrying outgoing \
-                                messages"
-                            );
+        let stream = other_sender_chains
+            .into_iter()
+            .map(|chain_id| {
+                let local_node = self.client.local_node.clone();
```
afck: (Sorry, I might have caused a merge conflict here. IIRC we don't need to clone the local node here.)

afck:
(Looks good to me, but the merge conflicts need to be resolved and it would be good if someone else could confirm that the web client and the exporter changes are fine.)

linera-core/src/client/chain_client/mod.rs
MathieuDutSik: Do not worry. It was rebased away.

linera-service/src/exporter/storage.rs
MathieuDutSik: Well, actually, we deduplicate the line below.
But yes, let us keep `BTreeMap`.

linera-core/src/client/requests_scheduler/scheduler.rs
MathieuDutSik: Yes.

web/@linera/client/build.bash
MathieuDutSik: It was a mistake and this was eliminated.

linera-execution/src/wasm/wasmer.rs
MathieuDutSik: Yes, I propose to allow the returning of unneeded_wraps here as it is simpler.

linera-core/src/client/chain_client/mod.rs
```diff
@@ -1032,8 +1032,9 @@ impl<Env: Environment> ChainClient<Env> {
         nodes: &[RemoteNode<Env::ValidatorNode>],
         other_sender_chains: Vec<ChainId>,
     ) {
-        let stream = FuturesUnordered::from_iter(other_sender_chains.into_iter().map(
-            |chain_id| async move {
+        let stream: FuturesUnordered<_> = other_sender_chains
```
afck, line 1035: Should be `collect::<FuturesUnordered<_>>()` instead.

PR #5676, opened 2026-03-12
Implement the support for passing queries in the client.

### LINKED ISSUE #5279: Add a command to send Graphql queries without spinning a node service
Maybe `linera api` or `linera gql`

see also `gh api` from github

MathieuDutSik:
## Motivation

It would be nice to be able to query a chain without opening a node-service.

Fixes #5279 

## Proposal

Add a function to the `linera` for doing exactly that. Add it to the client as well.
Note that the design somehow assumes that the smart contract being accessed is processing a GraphQL query.

## Test Plan

CI.
An additional test is added to `test_wasm_end_to_end_counter`.

## Release Plan

This could be backported to `testnet_conway`.

## Links

None.

### DIFF
--- a/CLI.md
+++ b/CLI.md
@@ -106,7 +106,7 @@ Client implementation and command-line tool for the Linera blockchain
 * `create-genesis-config` — Create genesis configuration for a Linera deployment. Create initial user chains and print information to be used for initialization of validator setup. This will also create an initial wallet for the owner of the initial "root" chains
 * `watch` — Watch the network for notifications
 * `service` — Run a GraphQL service to explore and extend the chains of the wallet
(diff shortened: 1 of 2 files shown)


CLI.md
```diff
@@ -750,6 +752,23 @@ Run a GraphQL service to explore and extend the chains of the wallet
 
 
 
+## `linera query-application`
+
+Query an application without starting the node service. This performs a read-only GraphQL query directly on a smart contract's service, without creating a lock or a persistent HTTP server
```
afck: Not sure if we need the explicit "without starting the node service". That goes for most commands.

MathieuDutSik: Ok.

PR #5677, opened 2026-03-12
Gate Tempo deployment on otel_endpoint being configured

ndr-ds:
## Motivation

Tempo is unconditionally deployed via helmfile even when `otel_endpoint` is not
configured, wasting resources (pod + 100Gi PVC) on clusters that don't use
distributed tracing.

## Proposal

Add an `installed:` condition to the Tempo helmfile release that checks whether
`LINERA_HELMFILE_SET_OTLP_EXPORTER_ENDPOINT` is set. When `otel_endpoint` is
null in network properties, lineractl doesn't set this env var, so Tempo won't
be deployed.

## Test Plan

CI

### DIFF
--- a/kubernetes/linera-validator/helmfile.yaml
+++ b/kubernetes/linera-validator/helmfile.yaml
@@ -181,6 +181,7 @@ releases:
     namespace: tempo
     chart: grafana/tempo
     timeout: 900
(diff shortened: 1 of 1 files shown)


ndr-ds:
This is already deployed on testnet, so merging

PR #5679, opened 2026-03-12
Fix TTL policy for chain workers

ma2bd:
## Motivation

Fix race condition in `spawn_keep_alive` task introduced by #5502. This issue could cause a chain worker to be shut down even though it is actively being used.

## Proposal

- Restructure the keep-alive loop so it always sleeps for exactly the remaining TTL time (instead of the full TTL at the top of every iteration).
- After the TTL expires and the write lock is acquired, re-check `last_access` to confirm the worker is still idle (double-check pattern). If the chain was accessed in the meantime, drop the lock, sleep for the remaining time, and retry.

## Test Plan

Existing CI tests.

## Release Plan

- portable to testnet along with #5502

### DIFF
--- a/linera-core/src/chain_worker/handle.rs
+++ b/linera-core/src/chain_worker/handle.rs
@@ -166,30 +166,50 @@ fn spawn_keep_alive<S: Storage + Clone + 'static>(
 ) {
     linera_base::Task::spawn(async move {
         loop {
-            linera_base::time::timer::sleep(ttl).await;
             let last = last_access.load(Ordering::Relaxed);
             let now = current_time_micros();
             let idle_micros = now.saturating_sub(last);
             let ttl_micros = u64::try_from(ttl.as_micros()).unwrap_or(u64::MAX);
-            if idle_micros >= ttl_micros {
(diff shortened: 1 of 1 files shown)


linera-core/src/chain_worker/handle.rs
```diff
@@ -166,30 +166,50 @@ fn spawn_keep_alive<S: Storage + Clone + 'static>(
...
+
+                // Returned the task to clear the service endpoint.
+                guard.clear_service_runtime()
+            };
+            if let Some(task) = task {
+                if let Err(err) = task.await {
+                    tracing::warn!(%err, "Failed to shut down service runtime");
+                }
             }
+            drop(state);
```
afck, line 210: Before this line, the guard is already dropped, isn't it? Doesn't that mean it's possible that in the meantime another task has used the state in the meantime? (Not a huge problem, but the scenario we were meaning to solve, isn't it?)

ma2bd: I didn't change this. Let me think

PR #5680, opened 2026-03-12
Fix dropped notifications in `retry_pending_cross_chain_requests`

afck:
## Motivation

`retry_pending_cross_chain_requests` was silently dropping `NewIncomingBundle` notifications returned by `handle_cross_chain_request`. This could potentially cause dropped notifications when concurrent paths (e.g. validator notification streams) triggered cross-chain delivery through this method.

## Proposal

Don't drop these.

## Test Plan

CI

It's unclear if this was actually a problem, and related to some of the flaky tests we're seeing.

## Release Plan

## Links

### DIFF
--- a/linera-core/src/client/chain_client/mod.rs
+++ b/linera-core/src/client/chain_client/mod.rs
@@ -1055,11 +1055,11 @@ impl<Env: Environment> ChainClient<Env> {
         nodes: &[RemoteNode<Env::ValidatorNode>],
         other_sender_chains: Vec<ChainId>,
     ) {
-        let stream = FuturesUnordered::from_iter(other_sender_chains.into_iter().map(|chain_id| {
-            let local_node = self.client.local_node.clone();
-            async move {
-                if let Err(error) = match local_node
-                    .retry_pending_cross_chain_requests(chain_id, &self.client.notifier)
+        let stream = FuturesUnordered::from_iter(other_sender_chains.into_iter().map(
+            |chain_id| async move {
+                if let Err(error) = match self
+                    .client
+                    .retry_pending_cross_chain_requests(chain_id)
                     .await
                 {
                     Ok(()) => Ok(()),
@@ -1076,8 +1076,8 @@ impl<Env: Environment> ChainClient<Env> {
                                 messages"
                             );
                         }
-                        local_node
-                            .retry_pending_cross_chain_requests(chain_id, &self.client.notifier)
+                        self.client
+                            .retry_pending_cross_chain_requests(chain_id)
                             .await
                     }
                     err => err,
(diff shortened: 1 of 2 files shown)


linera-core/src/client/mod.rs
```diff
@@ -338,6 +338,16 @@ impl<Env: Environment> Client<Env> {
...
 
+    /// Handles any pending local cross-chain requests, notifying subscribers.
+    pub async fn retry_pending_cross_chain_requests(
+        &self,
+        sender_chain: ChainId,
+    ) -> Result<(), LocalNodeError> {
+        self.local_node
+            .retry_pending_cross_chain_requests(sender_chain, &self.notifier)
+            .await
+    }
```
MathieuDutSik, line 349: Subsequent to this PR, we have 3 functions named `retry_pending_cross_chain_requests` with the following APIs:
```rust
pub async fn retry_pending_cross_chain_requests(&self, sender_chain: ChainId) -> Result<(), ....>
async fn retry_pending_cross_chain_requests(&self, nodes: &[RemoteNode<Env::ValidatorNode>], other_sender_chains: Vec<ChainId>)
pub async fn retry_pending_cross_chain_requests(&self, sender_chain: ChainId, notifier: &impl Notifier) -> Result<(), ...>
```

I do not mind having functions named the same, but here the API seems different, so I would object

linera-core/src/client/chain_client/mod.rs
```diff
@@ -2123,7 +2123,7 @@ impl<Env: Environment> ChainClient<Env> {
                 .extend_chain_mode(description.id(), ListeningMode::FullChain);
             self.client
                 .local_node
-                .retry_pending_cross_chain_requests(self.chain_id)
+                .retry_pending_cross_chain_requests(self.chain_id, &self.client.notifier)
```
MathieuDutSik: I think we can write here
```rust
self.client.retry_pending_cross_chain_requests(self.chain_id)
```

And there are other places where we can simplify.

linera-core/src/client/mod.rs
afck: You're right, I renamed one of them in https://github.com/linera-io/linera-protocol/pull/5680/commits/501ae2df1616d9301c92ad46b7003513d838891b. The other two are doing almost the same, except one of them knows the client's notifier, the other (in the local node) needs it passed in.

linera-core/src/client/chain_client/mod.rs
afck: Done in https://github.com/linera-io/linera-protocol/pull/5680/commits/501ae2df1616d9301c92ad46b7003513d838891b and https://github.com/linera-io/linera-protocol/pull/5680/commits/392b9060b8db0ca5e5a64503d73b16d8f8ef2bb1.

PR #5686, opened 2026-03-13
Avoid a clone operation.

MathieuDutSik:
## Motivation

There is a check of outcome returns that in one branch decays to checking that `outcome == outcome.clone()`.
This is not needed.

## Proposal

Restructure the test to avoid that scenario.

## Test Plan

CI.

## Release Plan

## Links

None.

### DIFF
--- a/linera-rpc/tests/snapshots/format__format.yaml.snap
+++ b/linera-rpc/tests/snapshots/format__format.yaml.snap
@@ -941,14 +941,14 @@ Reason:
         STRUCT:
           - height:
               TYPENAME: BlockHeight
-          - hash:
+          - block_hash:
               TYPENAME: CryptoHash
     1:
       NewEvents:
         STRUCT:
           - height:
               TYPENAME: BlockHeight
-          - hash:
+          - block_hash:
               TYPENAME: CryptoHash
           - event_streams:
               SEQ:
@@ -972,7 +972,7 @@ Reason:
         STRUCT:
           - height:
               TYPENAME: BlockHeight
-          - hash:
+          - block_hash:
               TYPENAME: CryptoHash
 ResourceControlPolicy:
   STRUCT:


linera-core/src/chain_worker/state.rs
```diff
@@ -795,40 +954,37 @@ where
...
+                .with_context(|ctx| {
+                    chain
+                        .execution_state
+                        .context()
+                        .clone_with_base_key(ctx.base_key().bytes.clone())
+                })
+                .await;
+        } else {
+            let (proposed_block, outcome) = block.clone().into_proposal();
+            let oracle_responses = Some(outcome.oracle_responses.clone());
```
Twey, line 831: This is structural so I think we can even avoid the clone here? (by passing by reference)

MathieuDutSik: You are right, I managed to remove that clone operation.

linera-core/src/chain_worker/state.rs
```diff
@@ -856,17 +856,19 @@ where
...
+                hash: block_hash,
+            },
         });
         if !updated_streams.is_empty() {
             actions.notifications.push(Notification {
                 chain_id,
                 reason: Reason::NewEvents {
                     height,
-                    hash,
+                    hash: block_hash,
```
afck: Maybe we should actually rename both `hash` fields to `block_hash`?

linera-core/src/chain_worker/state.rs
```diff
@@ -811,58 +814,61 @@ where
...
+            let (proposed_block, verified, _resource_tracker) = chain
+                .execute_block(
+                    proposed_block,
+                    local_time,
+                    None,
+                    &published_blobs,
+                    oracle_responses,
+                    BundleExecutionPolicy::Abort,
+                )
+                .await?;
```
afck, line 841: (I forgot the reason why `execute_block` doesn't return a `Block` instead. Maybe that type isn't available in `linera-chain`? Or would that involve more cloning again? 🤔)

MathieuDutSik: I will look at that.
It all seems pretty interchangeable to me.

linera-core/src/chain_worker/state.rs
MathieuDutSik: Seems reasonable. Done.

PR #5687, opened 2026-03-13
Fix race conditions with getting and dropping chain workers.

afck:
## Motivation

There are two potential race conditions:
* With very low TTL, concurrent `get_or_create_chain_worker` could both load the chain state from disk. If one finishes, writes a modified state and gets removed from the map again before the other completes, the other puts a stale chain state into the map.
* `spawn_keep_alive` could compare the TTL and decide to drop the state, but before it drops its strong reference, some other task upgrades a weak one.

## Proposal

1. `get_or_create_chain_worker`: Now the map stores `Shared<oneshot::Receiver<Weak<_>>>` and a single atomic compute per iteration handles all cases: return a live worker,
  wait on another loader's `oneshot`, or claim the loading slot. Only the winner loads from storage; others await the result directly.
2. `spawn_keep_alive` simplification (`handle.rs`): Instead of acquiring a write lock to shut down the service runtime, the keep-alive now uses `Arc::try_unwrap` — if it holds the only strong reference, the state is dropped (closing channels shuts down the runtime naturally). If someone else still holds a reference, it gets the `Arc` back and continues.
3. `clear_service_runtime` removal (`state.rs`): No longer needed since the keep-alive doesn't explicitly shut down the runtime. The `service_runtime_task` field gets `#[expect(dead_code)]` since it's held purely for its drop behavior.

## Test Plan

CI
(It would be very hard to produce the race conditions, since they require very precise timing.)

## Release Plan

- Release SDK.
- Hotfix validators. (Not urgent.)

## Links

- Fixes potential regressions from #5502.

### DIFF
--- a/linera-core/src/chain_worker/handle.rs
+++ b/linera-core/src/chain_worker/handle.rs
@@ -101,27 +101,27 @@ impl ServiceRuntimeActor {
         let (execution_state_sender, incoming_execution_requests) =
             futures::channel::mpsc::unbounded();
         let (runtime_request_sender, runtime_request_receiver) = std::sync::mpsc::channel();
-        let task = thread_pool
-            .run((), move |()| async move {
-                // The dummy context is overwritten by `prepare_for_query`
-                // before the first actual query is executed.
-                ServiceSyncRuntime::new(
-                    execution_state_sender,
-                    QueryContext {
-                        chain_id,
-                        next_block_height: BlockHeight(0),
-                        local_time: Timestamp::from(0),
-                    },
-                )
-                .run(&runtime_request_receiver)
-            })
-            .await;
-        ServiceRuntimeActor {
+
+        Self {
             endpoint: ServiceRuntimeEndpoint {
                 incoming_execution_requests,
                 runtime_request_sender,
             },
-            task,
+            task: thread_pool
+                .run((), move |()| async move {
+                    // The dummy context is overwritten by `prepare_for_query`
+                    // before the first actual query is executed.
+                    ServiceSyncRuntime::new(
+                        execution_state_sender,
+                        QueryContext {
+                            chain_id,
+                            next_block_height: BlockHeight(0),
+                            local_time: Timestamp::from(0),
+                        },
+                    )
+                    .run(&runtime_request_receiver)
+                })
+                .await,
         }
     }
 }


linera-core/src/chain_worker/state.rs
```diff
@@ -83,6 +83,8 @@ where
     storage: StorageClient,
     chain: ChainStateView<StorageClient::Context>,
     service_runtime_endpoint: Option<ServiceRuntimeEndpoint>,
+    /// Held so the task is not dropped prematurely; dropped with the state.
+    #[expect(dead_code)]
     service_runtime_task: Option<web_thread_pool::Task<()>>,
```
ma2bd: How does holding or not holding this task handle make any difference?

afck: You're right, we don't need that at all anymore! Just the endpoint!

(The important point is that once we drop the endpoint, nobody can be waiting for the service runtime anymore, because at that point there is nobody with an `Arc` with the chain worker.)

afck: I removed it in https://github.com/linera-io/linera-protocol/pull/5687/commits/fdf0b9902a6cb3ec69c82855442d958378a9d0e8.

linera-core/src/worker.rs
```diff
@@ -750,76 +758,114 @@ where
...
-                service_runtime_endpoint,
-                service_runtime_task,
-            )
-            .await?;
+        let state = crate::chain_worker::state::ChainWorkerState::load(
+            self.chain_worker_config.clone(),
+            self.storage.clone(),
+            self.block_cache.clone(),
+            self.execution_state_cache.clone(),
+            self.chain_modes.clone(),
```
MathieuDutSik, line 852: I wonder if we could gain something by joining those entries together.

linera-core/src/worker.rs
```diff
@@ -750,76 +758,114 @@ where
...
+                            }
+                        },
+                        &guard,
+                    ) {
+                        papaya::Compute::Aborted(Ok(arc), ..) => return Ok(arc),
+                        papaya::Compute::Aborted(Err(wait), ..) => Some(wait),
+                        papaya::Compute::Inserted { .. } | papaya::Compute::Updated { .. } => None,
+                        papaya::Compute::Removed { .. } => unreachable!(),
+                    }
+                    // Guard dropped here.
```
MathieuDutSik: Not a useful remark.

linera-core/src/chain_worker/handle.rs
```diff
@@ -158,38 +147,37 @@ pub(crate) async fn write_lock<S: Storage + Clone + 'static>(
...
 fn spawn_keep_alive<S: Storage + Clone + 'static>(
-    state: Arc<RwLock<ChainWorkerState<S>>>,
+    chain_id: ChainId,
+    mut state: Arc<RwLock<ChainWorkerState<S>>>,
     last_access: Arc<AtomicU64>,
     ttl: Duration,
 ) {
     linera_base::Task::spawn(async move {
+        let ttl_micros = u64::try_from(ttl.as_micros()).unwrap_or(u64::MAX);
+        linera_base::time::timer::sleep(ttl).await;
```
MathieuDutSik: Those conversions to microseconds are a little bit unfortunate since `std::time::Duration` does have `saturating_sub` and so on.

Sure, we have the encapsulation in `linera_base::time::Duration`. But the definition of those functions should be there.

afck: Unfortunately there's no `AtomicSystemTime` or `AtomicDuration`, and I don't want to add a mutex for this, so I added an `AtomicTimestamp` that wraps pretty much the same arithmetic internally. I do think it makes the code a bit cleaner; what do you think?

linera-core/src/chain_worker/handle.rs
```diff
@@ -158,46 +179,33 @@ pub(crate) async fn write_lock<S: Storage + Clone + 'static>(
...
 fn spawn_keep_alive<S: Storage + Clone + 'static>(
-    state: Arc<RwLock<ChainWorkerState<S>>>,
-    last_access: Arc<AtomicU64>,
+    chain_id: ChainId,
+    mut state: Arc<RwLock<ChainWorkerState<S>>>,
+    last_access: Arc<AtomicTimestamp>,
     ttl: Duration,
 ) {
     linera_base::Task::spawn(async move {
+        linera_base::time::timer::sleep(ttl).await;
```
ma2bd: Is this line useful?

afck: No, you're right. I removed it in https://github.com/linera-io/linera-protocol/pull/5687/commits/f0c2c6f133d9a1c51300d91fe56d1b1573dc5dad; also made sure we treat `remaining == 0` the same as `None`.

linera-core/src/chain_worker/state.rs
afck: I brought it back. It looks like dropping a `web_thread_pool::Task` does cancel it.

linera-core/src/worker.rs
afck: Maybe, yes. But since that wasn't introduced with this change I'd prefer to try and discuss that on a separate PR.

linera-core/src/chain_worker/handle.rs
```diff
@@ -54,43 +86,40 @@ impl<S: Storage + Clone + 'static> Drop for RollbackGuard<S> {
...
-                            next_block_height: BlockHeight(0),
-                            local_time: Timestamp::from(0),
-                        },
-                    )
-                    .run(&runtime_request_receiver)
-                })
-                .await,
-        }
+/// Spawns a blocking task to execute the service runtime actor.
+pub(crate) async fn spawn_service_runtime_actor(
```
ma2bd: nit: what was the problem with `impl ServiceRuntimeActor { fn spawn() }` ?

afck: (I don't remember. :thinking: Will revert; possibly in a follow-up PR.)

afck: Done in https://github.com/linera-io/linera-protocol/pull/5687/commits/b8cd288e1cc820952e1abef1cc6bee0cdd0f3ba3.

PR #5688, opened 2026-03-13
move max_pending_message_bundles from ChainClientOptions to ChainWork…

ma2bd:
## Motivation

Preliminary step before optimizing `pending_message_bundles` more agressively

## Proposal

* Move `max_pending_message_bundles` from `ChainClientOptions` to `ChainWorkerConfig`

We started this trend with `priority_bundle_origins`. This is ok because
* The option is static (not changed by the controller)
* `ChainInfoQuery.request_pending_message_bundles` is only used locally by the client.

Validators don't usually serve this query (IIRC clients use the `received_log`) but they could actually benefit from a hard cap.

## Test Plan

CI


linera-core/src/client/mod.rs
ma2bd, line 2375: ah shoot this won't work

linera-core/src/chain_worker/state.rs
afck, line 1746: Maybe we should load the prioritized origins first here, and stop reading from the database once we've reached `max_pending_message_bundles`.

PR #5689, opened 2026-03-13
restructuration of inbox access

MathieuDutSik:
## Motivation

The function `fn process_cross_chain_update` of `linera-core/src/chain_worker/state.rs` is reading rhe inbox of origin three times:
* In next_block_height_to_receive (1 time)
* In last_anticipated_block_height (1 time)
* In receive_message_bundle (as many times as the number of bundles)

Another issue is the `NUM_INBOXES` metric being called in the hot path. The does a `inboxes.count()` which is an expensive operation.

This is inefficient.

## Proposal

Access the inbox directly in `process_cross_chain_update` and pass it around. We could not do it for receive_message_bundle.
Put a single emission of `NUM_INBOXES` when we did a `try_load_all_entries`.

## Test Plan

CI

## Release Plan

Can be backported to `testnet_conway` since this is strictly a change to the ordering of operations.

## Links

None.

### DIFF
--- a/linera-chain/src/chain.rs
+++ b/linera-chain/src/chain.rs
@@ -26,7 +26,7 @@ use linera_views::{
     context::Context,
     log_view::LogView,
     map_view::MapView,
-    reentrant_collection_view::{ReadGuardedView, ReentrantCollectionView, WriteGuardedView},
+    reentrant_collection_view::{ReadGuardedView, ReentrantCollectionView},
     register_view::RegisterView,
     views::{ClonableView, RootView, View},
 };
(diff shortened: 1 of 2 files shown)


linera-chain/src/chain.rs
```diff
@@ -553,10 +539,6 @@ where
 
         // Process the inbox bundle and update the inbox state.
         let mut inbox = self.inboxes.try_load_entry_mut(origin).await?;
-        #[cfg(with_metrics)]
-        metrics::NUM_INBOXES
-            .with_label_values(&[])
-            .observe(self.inboxes.count().await? as f64);
```
ma2bd, line 559: Good catch! It also confirms my fears: people wrongly assume that `count` is O(1). How about we rename this function `iterative_count`?

MathieuDutSik: Yes, will do later that rename.

PR #5695, opened 2026-03-15
Implement notifications with `simple` network.

MathieuDutSik:
## Motivation

The GRPC does support notifications, but `simple` does not. It is true that we use simple very little in comparison with GRPC. However, there is no valid reason for not supporting notifications.
Also, extending the test surfaces should allow detecting more problems.

## Proposal

Straightforward implementation.

## Test Plan

One test was added to the CI for `retry_notification`. But no tests were added for the repeated transfers, which maybe we should have done.

## Release Plan

## Links

None.

### DIFF
--- a/linera-service/src/proxy/main.rs
+++ b/linera-service/src/proxy/main.rs
@@ -319,8 +319,7 @@ where
         let protocol = self.internal_config.protocol;
         let mut select_all = SelectAll::new();
         // Group chains by their owning shard.
-        let mut chains_by_shard =
-            std::collections::HashMap::<usize, Vec<ChainId>>::new();
+        let mut chains_by_shard = std::collections::HashMap::<usize, Vec<ChainId>>::new();
         for chain_id in chains {
             let shard_id = self.internal_config.get_shard_id(chain_id);
             chains_by_shard.entry(shard_id).or_default().push(chain_id);


linera-service/src/proxy/main.rs
```diff
@@ -302,6 +309,50 @@ where
...
     }
 
+    /// Subscribes to notification streams from all shards for the given chains.
+    async fn subscribe_to_shards(
+        &self,
+        chains: Vec<ChainId>,
+    ) -> Option<Pin<Box<dyn Stream<Item = RpcMessage> + Send>>> {
+        let protocol = self.internal_config.protocol;
+        let mut select_all = SelectAll::new();
+        // Connect to each shard and subscribe. We send all chains to every shard;
```
afck: (I guess the proxy would have the knowledge to split it up: It knows which shard does which chains?)

linera-service/src/proxy/main.rs
```diff
@@ -302,6 +309,58 @@ where
...
+        let mut chains_by_shard = std::collections::HashMap::<usize, Vec<ChainId>>::new();
+        for chain_id in chains {
+            let shard_id = self.internal_config.get_shard_id(chain_id);
+            chains_by_shard.entry(shard_id).or_default().push(chain_id);
+        }
+        // Only connect to shards that own at least one requested chain.
+        for (shard_id, shard_chains) in chains_by_shard {
+            let shard = &self.internal_config.shards[shard_id];
+            let address = (shard.host.clone(), shard.port);
+            match protocol.connect(address).await {
```
afck, line 331: I think the gRPC proxy works differently, doesn't it? I imagine the shards just send all notifications to the proxy, and the proxy maintains the set of subscribing clients? Otherwise this will create a new proxy-shard connection for every subscribing client?

I guess it's fine for now since this is only used for testing.

Issue #5700, opened 2026-03-16
Pulling pending message bundles to build blocks gets slower and slower over time


ma2bd:
We're storing cursors and bundles in `InboxStateView`. This is convenient for block execution but to build a new block in the client, we want to access only the non-empty inboxes.

```
$ time scripts/query-assigned-chain.sh engines/engine-v2-22-BTC-1/market
[...]
--- Tip State ---
  Block hash: 3acad4789d61ea599ee12f23b9c4e247f69f0544ace3505088fdd8285c92b5dc
  Next block height: 161770
  Last block time: 2026-03-16 06:58:20 UTC (2min ago)
  Incoming bundles: 587221
  Operations: 14
  Outgoing messages: 849595

--- Chain Balance ---
  Balance: 23954.022360688

--- Pending Inboxes ---
  Total pending bundles: 461 (from 18184 origins)

    - 000aa1cff8e36f90...: 0 bundles
    - 000f745985bd03ae...: 0 bundles
    - 00158dd78443de2d...: 0 bundles
    - 0017bde8b690ae75...: 0 bundles
    - 001959a7770f18e4...: 0 bundles
    - 001e213939b01e1f...: 0 bundles
    - 001f4803623c8ba7...: 0 bundles
    - 0020cad72fd26d6d...: 0 bundles
    - 0035b7c600a85616...: 0 bundles
    - 003a38905eb81dcd...: 0 bundles
    ... and 18174 more origins

=== Query Complete ===

real	0m16.820s
user	0m1.018s
sys	0m0.260s
```

afck:
We're already maintaining a set of `nonempty_outboxes`; I'll do the same for the inboxes. (On `main`; for the backport maybe just an additional `bool` that tells us if that field has been initialized?)

Issue #4795, opened 2025-10-14
Inbox terminology


afck:
If a confirmed block sending a message has been seen in a node (client or validator), and the message put into the recipient chain's inbox, we currently call it "added" or "received". We don't have a good term for when it has been included in a confirmed block as an incoming message on the receiver chain, but in the inbox implementation itself we call it "removed".

After discussing with @deuszx, we'd suggest changing the terminology:
* If the confirmed sender block has been seen and the message added to the inbox, the message is "confirmed".
* If a confirmed block has included it as an incoming message, the message is "received".

We would change the comments, and also e.g.:
* `find_received_certificates` → `find_confirmed_messages`
* `added` and `removed` in the inbox → `confirmed` and `received`

ma2bd:
1) I would keep add/remove in the inbox because this is what we define in the whitepaper (and also it makes sense to me).
2) I'm not fond of overloading the word "confirmed". Right now it means "committed" not "locally verified" or something like that.

PR #5703, opened 2026-03-16
Track non-empty inboxes to avoid loading all inbox entries

afck:
## Motivation

Chains with many interacting origins accumulate `InboxStateView` entries, most of which are empty. This makes processing pending message bundles very inefficient since `try_load_all_entries()` loads every inbox.

## Proposal

Add a `nonempty_inboxes` set (mirroring the existing `nonempty_outboxes` pattern) so that only inboxes with pending added bundles are loaded.

## Test Plan

CI should catch regressions. We could try and do benchmarks first, or just see if it resolves the performance issue (https://github.com/linera-io/linera-protocol/issues/5700) we see on the testnet.

## Release Plan

- Backport to `testnet_conway` with a `bool` that tells us if the collection has been initialized. (So old chain state views can be loaded correctly.)
- Release a new SDK.

## Links

- Part of https://github.com/linera-io/linera-protocol/issues/5700 (but needs backport)

### DIFF
--- a/linera-chain/src/chain.rs
+++ b/linera-chain/src/chain.rs
@@ -655,11 +655,6 @@ where
         Ok(())
     }
 
-    /// Returns the chain IDs of all origins for which a message is waiting in the inbox.
-    pub fn nonempty_inbox_chain_ids(&self) -> impl Iterator<Item = &ChainId> {
-        self.nonempty_inboxes.get().iter()
-    }
-
     /// Returns the chain IDs of all recipients for which a message is waiting in the outbox.
     pub fn nonempty_outbox_chain_ids(&self) -> Vec<ChainId> {
         self.nonempty_outboxes.get().iter().copied().collect()
--- a/linera-core/src/chain_worker/state.rs
+++ b/linera-core/src/chain_worker/state.rs
@@ -1609,7 +1609,7 @@ where
         if query.request_pending_message_bundles {
             let mut bundles = Vec::new();
             let nonempty_origins: Vec<ChainId> =
-                chain.nonempty_inbox_chain_ids().copied().collect();
+                chain.nonempty_inboxes.get().iter().copied().collect();
             #[cfg(with_metrics)]
             metrics::NUM_INBOXES
                 .with_label_values(&[])
@@ -1621,14 +1621,15 @@ where
             };
             let inboxes = chain.inboxes.try_load_entries(&nonempty_origins).await?;
             for (origin, inbox) in nonempty_origins.into_iter().zip(inboxes) {
-                if let Some(inbox) = inbox {
-                    for bundle in inbox.added_bundles.elements().await? {
-                        bundles.push(IncomingBundle {
-                            origin,
-                            bundle,
-                            action,
-                        });
-                    }
+                let inbox = inbox.ok_or_else(|| {
+                    ChainError::InternalError(format!("Missing inbox for origin {origin}"))
+                })?;
+                for bundle in inbox.added_bundles.elements().await? {
+                    bundles.push(IncomingBundle {
+                        origin,
+                        bundle,
+                        action,
+                    });
                 }
             }
             bundles.sort_by_key(|b| b.bundle.timestamp);


linera-chain/src/chain.rs
```diff
@@ -643,10 +648,18 @@ where
                     );
                 }
             }
+            if inbox.added_bundles.count() == 0 {
+                self.nonempty_inboxes.get_mut().remove(&origin);
+            }
```
deuszx, line 653: Is `added_bundles` tracking per block? is it possible that we won't empty the inbox in a single block and then this code here removes the `origin` from `nonempty_inboxes` set even though it's not really empty yet?

afck: No, the `added_bundles` are the bundles that have been added to the inbox but _not_ to any block yet. They are exactly what we load when we try to process the inbox. So if they are 0, the inbox is "empty". (There could still be "removed" entries that have been received in a block but we haven't seen the sender block yet, but those are not relevant for processing the inbox.)

linera-chain/src/chain.rs
```diff
@@ -643,10 +648,18 @@ where
...
         }
         Ok(())
     }
 
+    /// Returns the chain IDs of all origins for which a message is waiting in the inbox.
+    pub fn nonempty_inbox_chain_ids(&self) -> impl Iterator<Item = &ChainId> {
+        self.nonempty_inboxes.get().iter()
+    }
```
MathieuDutSik: This looks very much LLM generated code which loves to insert new functions for limited reasons.

linera-core/src/chain_worker/state.rs
```diff
@@ -1608,23 +1608,27 @@ where
...
+            let inboxes = chain.inboxes.try_load_entries(&nonempty_origins).await?;
+            for (origin, inbox) in nonempty_origins.into_iter().zip(inboxes) {
+                if let Some(inbox) = inbox {
+                    for bundle in inbox.added_bundles.elements().await? {
+                        bundles.push(IncomingBundle {
+                            origin,
+                            bundle,
+                            action,
+                        });
+                    }
```
MathieuDutSik: The case of `inbox = None`  should lead to an explicit error.

linera-chain/src/chain.rs
```diff
@@ -643,10 +648,18 @@ where
                     );
                 }
             }
+            if inbox.added_bundles.count() == 0 {
+                self.nonempty_inboxes.get_mut().remove(&origin);
```
MathieuDutSik, line 652: If that were the case, could we clear the entry in the `ReentrantCollectionView` of the inboxes?

afck: No, it might still contain `removed_bundles`, and even if it doesn't, we need to remember the current cursor position.

linera-chain/src/chain.rs
afck: Removed in https://github.com/linera-io/linera-protocol/pull/5703/commits/a6a41a32aa81092af143a3a2533a358592fd71b5.

linera-core/src/chain_worker/state.rs
afck: Done in https://github.com/linera-io/linera-protocol/pull/5703/commits/a6a41a32aa81092af143a3a2533a358592fd71b5.

linera-chain/src/chain.rs
MathieuDutSik: Then I might ask for renaming `nonempty_inboxes` maybe to `nonempty_added_inboxes`, but that is not really important.

Is there a scenario where a specific `inbox` can be removed from `inboxes`? I do not see it in the code, and I think that is a problem.

Not for this PR of course.

afck: Maybe when we introduce chain checkpoints we can do something like that, yes. For now I think we have to keep them because of the cursors, to prevent replaying incoming messages.

afck: Not sure about the name. It's not like the inbox was added; it's that the inbox doesn't contain any "added messages", which are messages where we've seen the sending block but no receiving block. (Also see https://github.com/linera-io/linera-protocol/issues/4795.)

TBH I can't come up with a name that's better than `nonempty_inboxes` and not exceedingly long.

linera-chain/src/chain.rs
```diff
@@ -235,6 +235,8 @@ where
 
     /// Blocks that have been verified but not executed yet, and that may not be contiguous.
     pub preprocessed_blocks: MapView<C, BlockHeight, CryptoHash>,
+    /// Inboxes with at least one pending added bundle. This allows us to avoid loading all inboxes.
+    pub nonempty_inboxes: RegisterView<C, BTreeSet<ChainId>>,
```
ma2bd, line 239: Note: there is a long-standing limitation of Serde/BCS where `BTreeSet` are serialized as vectors. I think this is fine (and even slightly better here).

ma2bd: (Concretely, if someone reorders the entries in the DB, it will not be rejected..)

afck: Yes… I don't even see why this is a limitation, at least in this use case?

afck: Ah, thanks for the clarification! Yes, that should be fine. This isn't part of the execution state, so it doesn't get hashed or anything.

linera-chain/src/chain.rs
```diff
@@ -643,6 +648,9 @@ where
                     );
                 }
             }
+            if inbox.added_bundles.count() == 0 {
```
ma2bd, line 651: This one is O(1). good!

PR #5705, opened 2026-03-16
Improvements to the metrics.

MathieuDutSik:
## Motivation

The metrics need to be handled carefully:
* They should not be in the hot path.
* Calling the metric should be cheap and not be a big computation.

## Proposal

Move the metrics for inbox/outbox out of the critical path and make them async by computing the
size from the nonempty boxes.

## Test Plan

CI

## Release Plan

## Links

None

### DIFF
--- a/linera-chain/src/chain.rs
+++ b/linera-chain/src/chain.rs
@@ -414,7 +414,7 @@ where
         #[cfg(with_metrics)]
         metrics::NUM_OUTBOXES
             .with_label_values(&[])
-            .observe(self.outboxes.count().await? as f64);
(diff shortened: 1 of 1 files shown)


linera-chain/src/chain.rs
```diff
@@ -412,7 +412,7 @@ where
         #[cfg(with_metrics)]
         metrics::NUM_OUTBOXES
             .with_label_values(&[])
-            .observe(self.outboxes.count().await? as f64);
+            .observe(self.nonempty_outboxes.get().len() as f64);
```
afck, line 417: (Good idea! It slightly changes the meaning of the metric; but I guess it makes more sense this way.)

PR #5706, opened 2026-03-16
Add `request_previous_event_blocks` field to `ChainInfoQuery`

afck:
Port of https://github.com/linera-io/linera-protocol/pull/5659.

## Motivation

When subscribing to a notification stream, chains should immediately get notified about the most recent events, even if they happened before the subscription. For that the client must be able to ask the validators for those latest events.

## Proposal

Expose previous event block lookups through the existing `ChainInfoQuery`/`ChainInfoResponse` mechanism.

This is different from the new endpoint in #5659: Requesting this kind of information via `ChainInfoQuery` is the pattern we currently use everywhere, and on `main` we don't need to be backwards-compatible, so we can add fields to `ChainInfoQuery`.

## Test Plan

A worker test was extended.

(This will be used more in https://github.com/linera-io/linera-protocol/pull/5653, or rather, in its port to `main`.)

## Release Plan

## Links

- Port/variant of #5659.
- Motivated by https://github.com/linera-io/linera-protocol/pull/5653.

### DIFF
--- a/linera-core/src/unit_tests/worker_tests.rs
+++ b/linera-core/src/unit_tests/worker_tests.rs
@@ -3212,8 +3212,8 @@ where
 
     // Query the admin chain for previous_event_blocks.
     let stream_id = StreamId::system(NEW_EPOCH_STREAM_NAME);
-    let query = ChainInfoQuery::new(admin_chain_id)
-        .with_previous_event_blocks(vec![stream_id.clone()]);
+    let query =
+        ChainInfoQuery::new(admin_chain_id).with_previous_event_blocks(vec![stream_id.clone()]);
     let (response, _) = env
         .executing_worker()
         .handle_chain_info_query(query)


linera-core/src/chain_worker/state.rs
```diff
@@ -1645,6 +1645,30 @@ where
...
+                .multi_get(&stream_ids)
+                .await?;
+            let mut indices = Vec::new();
+            let mut streams_with_heights = Vec::new();
+            for (stream_id, height) in stream_ids.into_iter().zip(heights) {
+                if let Some(height) = height {
+                    let index = usize::try_from(height.0).map_err(|_| ArithmeticError::Overflow)?;
+                    indices.push(index);
+                    streams_with_heights.push((stream_id, height));
+                }
```
MathieuDutSik, line 1662: Is `height = None` something that should lead to some error of some kind?

linera-core/src/chain_worker/state.rs
```diff
@@ -1645,6 +1645,30 @@ where
...
+            }
+            let hashes = chain.confirmed_log.multi_get(indices).await?;
+            for (hash, (stream_id, height)) in hashes.into_iter().zip(streams_with_heights) {
+                if let Some(hash) = hash {
```
MathieuDutSik: Same.

afck: Good catch, I made this an error: https://github.com/linera-io/linera-protocol/pull/5706/commits/9895ac0494bfa1529d8ef7b05201ebbe12859313

linera-core/src/chain_worker/state.rs
afck: In this case I don't think it's an error:

The client asked for a set of streams, but it's not clear that all of those streams already have some event. It's perfectly fine for entries to be missing here: no events to be reported.

afck:
I'll add a test…

afck:
Closing this in favor of #5737.

PR #5709, opened 2026-03-16
[testnet] Track non-empty inboxes to avoid loading all inbox entries (#5703)

afck:
Backport of #5703

## Motivation

Chains with many interacting origins accumulate `InboxStateView` entries, most of which are empty. This makes processing pending message bundles very inefficient since `try_load_all_entries()` loads every inbox.

## Proposal

Add a `nonempty_inboxes` set (mirroring the existing `nonempty_outboxes` pattern) so that only inboxes with pending added bundles are loaded.

The set is inside an `Option` in this backport: Chain state view saved before the update will have `None` there, and will lazily initialize the set when needed.

## Test Plan

CI should catch regressions. We could try and do benchmarks first, or just see if it resolves the performance issue (https://github.com/linera-io/linera-protocol/issues/5700) we see on the testnet.

## Release Plan

- Release a new SDK.

## Links

- Fixes https://github.com/linera-io/linera-protocol/issues/5700

### DIFF
--- a/linera-core/src/chain_worker/state.rs
+++ b/linera-core/src/chain_worker/state.rs
@@ -1721,18 +1721,8 @@ where
         query: ChainInfoQuery,
     ) -> Result<ChainInfoResponse, WorkerError> {
         self.initialize_and_save_if_needed().await?;
-        // Lazily initialize the nonempty_inboxes set for chains from older DB versions.
-        if self.chain.nonempty_inboxes.get().is_none() {
-            let pairs = self.chain.inboxes.try_load_all_entries().await?;
-            let nonempty_set: BTreeSet<ChainId> = pairs
-                .into_iter()
-                .filter(|(_, inbox)| inbox.added_bundles.count() > 0)
-                .map(|(origin, _)| origin)
-                .collect();
-            *self.chain.nonempty_inboxes.get_mut() = Some(nonempty_set);
-        }
-        let chain = &self.chain;
-        let mut info = ChainInfo::from(chain);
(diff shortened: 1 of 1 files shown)


linera-chain/src/chain.rs
```diff
@@ -272,6 +272,11 @@ where
     /// The indices of next events we expect to see per stream (could be ahead of the last
     /// executed block in sparse chains).
     pub next_expected_events: MapView<C, StreamId, u32>,
+
+    /// Inboxes with at least one pending added bundle. This allows us to avoid loading all
+    /// inboxes. `None` means the set hasn't been computed yet for this chain (backwards
+    /// compatibility with pre-existing database entries).
+    pub nonempty_inboxes: RegisterView<C, Option<BTreeSet<ChainId>>>,
```
ma2bd, line 279: nice touch with the option!

ma2bd (approved):
LGTM!

linera-core/src/chain_worker/state.rs
```diff
@@ -1747,18 +1747,37 @@ where
...
+            let origins_and_inboxes = origins
+                .into_iter()
+                .zip(inboxes)
+                .filter_map(|(origin, inbox)| Some((origin, inbox?)))
+                .collect::<Vec<_>>();
             #[cfg(with_metrics)]
             metrics::NUM_INBOXES
                 .with_label_values(&[])
-                .observe(pairs.len() as f64);
+                .observe(origins_and_inboxes.len() as f64);
```
ma2bd, line 1774: agreed that this is a more interesting metric anyway

PR #5710, opened 2026-03-16
Make `read_event` work even if the event is not locally known yet.

afck:
Port/rewrite of #5638 and #5656.
## Motivation

We have a use case for reading events without having subscribed and seen `UpdateStreams` yet.

## Proposal

Add an index of all event-publishing blocks.

Make the client download the publishing block whenever it is missing an event, then retry.

## Test Plan

A test was added.
Instead of writing a new example just for this, `UpdateStream` is used manually, without the client having the streams yet.

## Release Plan

- Nothing to do.

## Links

- Testnet version: #5638 
- Follow-up with cleanups: #5656 

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -622,8 +622,7 @@ impl<Env: Environment> Client<Env> {
                         let heights = remote_node
                             .node
                             .event_block_heights(remaining_ref.to_vec())
-                            .await
-                            .map_err(|_| ())?;
+                            .await?;
 
                         // Separate resolved and unresolved events.
                         let mut chain_heights = BTreeMap::<_, BTreeSet<_>>::new();
@@ -645,7 +644,7 @@ impl<Env: Environment> Client<Env> {
                         }
                         if chain_heights.is_empty() {
                             // This validator has no useful information.
-                            return Err(());
+                            return Err(chain_client::Error::from(NodeError::EventsNotFound(remaining_ref.clone())));
                         }
 
                         // Download certificates and verify them.
@@ -659,8 +658,7 @@ impl<Env: Environment> Client<Env> {
                                     chain_id,
                                     heights_vec,
                                 )
-                                .await
-                                .map_err(|_| ())?;
+                                .await?;
                             for cert in &certificates {
                                 // Verify the block contains the expected events.
                                 let block = cert.block();
@@ -673,7 +671,7 @@ impl<Env: Environment> Client<Env> {
                                             %validator_address, ?expected_event_ids, ?block_event_ids,
                                             "validator lied about events in block."
                                         );
-                                        return Err(());
+                                        return Err(NodeError::UnexpectedCertificateValue.into());
                                     }
                                 }
                             }
@@ -684,6 +682,7 @@ impl<Env: Environment> Client<Env> {
                                             %validator_address, %error,
                                             "invalid certificate"
                                         );
+                                        error
                                     })?
                                     .into_result()
                                     .map_err(|error| {
@@ -691,6 +690,7 @@ impl<Env: Environment> Client<Env> {
                                             %validator_address, %error,
                                             "could not check certificate"
                                         );
+                                        error
                                     })?;
                                 checked_certificates.push(cert);
                             }
@@ -1150,7 +1150,7 @@ impl<Env: Environment> Client<Env> {
                     if remote_heights.is_empty() {
                         // It makes no sense to return `Ok(_)` if we aren't going to try downloading
                         // anything from the validator - let the function try the other validators
-                        return Err(());
+                        return Err(NodeError::MissingCertificateValue);
                     }
                     let certificates = self
                         .requests_scheduler
@@ -1159,19 +1159,13 @@ impl<Env: Environment> Client<Env> {
                             sender_chain_id,
                             remote_heights,
                         )
-                        .await
-                        .map_err(|_| ())?;
+                        .await?;
                     let mut certificates_with_check_results = vec![];
                     for cert in certificates {
-                        if let Ok(check_result) =
-                            Self::check_certificate(max_epoch, committees_ref, &cert)
-                        {
-                            certificates_with_check_results
-                                .push((cert, check_result.into_result().is_ok()));
-                        } else {
-                            // Invalid signature - the validator is faulty
-                            return Err(());
-                        }
+                        let check_result =
+                            Self::check_certificate(max_epoch, committees_ref, &cert)?;
+                        certificates_with_check_results
+                            .push((cert, check_result.into_result().is_ok()));
                     }
                     Ok(certificates_with_check_results)
                 },


linera-core/src/client/mod.rs
```diff
@@ -591,6 +591,60 @@ impl<Env: Environment> Client<Env> {
...
+                // Step 1: Query this validator for the block heights.
+                let heights = remote_node
+                    .node
+                    .event_block_heights(event_ids.to_vec())
+                    .await?;
+                // Step 2: Group resolved heights by chain ID.
+                let mut chain_heights = BTreeMap::<_, BTreeSet<_>>::new();
+                for (event_id, maybe_height) in event_ids.iter().zip(heights) {
+                    let height = maybe_height
+                        .ok_or_else(|| NodeError::EventsNotFound(vec![event_id.clone()]))?;
```
bart-linera: Even if multiple events are missing, we'll still only report one as missing this way, won't we? Wouldn't it make more sense to collect all the missing IDs and return that after the loop if it's not empty?

bart-linera: Also, this is inside `communicate_concurrently`... We'll try to download all events from all the validators this way, right? It seems a bit wasteful. And if every validator is missing a different single event, then all events might actually exist, but we'll download none of them, which might be a problem 😬

afck: I added two commits to try and address some of these concerns. We now collect the missing IDs.

The request scheduler should to some extent deduplicate `download_certificates_by_heights` calls, yes.

linera-core/src/client/mod.rs
```diff
@@ -591,6 +591,144 @@ impl<Env: Environment> Client<Env> {
...
+                Ok((certificates, unresolved)) => {
+                    for certificate in certificates {
+                        let mode = ReceiveCertificateMode::AlreadyChecked;
+                        self.receive_sender_certificate(certificate, mode, None)
+                            .await?;
+                    }
+                    if unresolved.len() == remaining_event_ids.len() {
+                        // No progress — all remaining events are still unresolved.
+                        break;
+                    }
```
bart-linera: I'm a bit concerned about this. If one validator cannot resolve some events, returns them as unresolved, and then in the next loop iteration it also happens to be the first one to respond, we might break out of the loop thinking the events are unresolvable, even though it might not be the case. I think what we really want is the intersection of all the events that couldn't be resolved, and stop polling the validators once that hits zero. It's probably not doable with `communicate_concurrently`, we might need another function similar to it that does it differently.

afck: I think this is actually unreachable, because we return `Err(())` above if none of the events could be downloaded.

Also, I believe if we call `validators.retain` below, we always remove all of them, because if `communicate_concurrently` returned an error that means all validators failed.

I'll take another look at this…

afck: I fixed and simplified this in https://github.com/linera-io/linera-protocol/pull/5710/commits/ea3162a3ecc085723fc1b32f5756fbf43e856714 and https://github.com/linera-io/linera-protocol/pull/5710/commits/4562bca69e406f870452f21dcc809cca1653ebc9.

linera-core/src/client/mod.rs
```diff
@@ -591,6 +591,144 @@ impl<Env: Environment> Client<Env> {
...
+                            for cert in &certificates {
+                                // Verify the block contains the expected events.
+                                let block = cert.block();
+                                let block_event_ids = block.event_ids().collect::<HashSet<_>>();
+                                if let Some(expected) =
+                                    expected_events.get(&(chain_id, block.header.height))
+                                {
+                                    if !expected.is_subset(&block_event_ids) {
+                                        // Faulty validator: block doesn't contain expected events.
+                                        return Err(());
```
MathieuDutSik: We probably could have a more precise error here than `Err(())`.

MathieuDutSik: Why not a `chain_client::Error`.

linera-storage/src/db_storage.rs
```diff
@@ -950,6 +967,39 @@ where
...
 
+    async fn read_event_block_heights(
+        &self,
+        event_ids: &[EventId],
+    ) -> Result<Vec<Option<BlockHeight>>, ViewError> {
+        if event_ids.is_empty() {
+            return Ok(Vec::new());
+        }
+        // Group event IDs by chain ID for batch lookups per partition.
+        let mut chain_groups: BTreeMap<ChainId, Vec<(usize, Vec<u8>)>> = BTreeMap::new();
```
MathieuDutSik: This can be simplified.

linera-storage/src/db_storage.rs
```diff
@@ -950,6 +967,39 @@ where
         Ok(hash_options)
     }
 
+    async fn read_event_block_heights(
+        &self,
+        event_ids: &[EventId],
```
MathieuDutSik, line 972: I am quite sure that you must have considered it, but let me ask anyway.
The `EventId` already contains the `chain_id` which is 32 bytes. The block height is 4 bytes.
Wouldn't it have been simpler to add the block_height to the `EventId`?

linera-core/src/client/mod.rs
```diff
@@ -591,6 +591,144 @@ impl<Env: Environment> Client<Env> {
...
+                                            validator = %remote_node.address(), %error,
+                                            "invalid certificate"
+                                        );
+                                    })?
+                                    .into_result()
+                                    .map_err(|error| {
+                                        tracing::debug!(
+                                            validator = %remote_node.address(), %error,
+                                            "could not check certificate"
+                                        );
```
MathieuDutSik, line 692: Here we have a debug warning, but associated to that an error `()`. That seems odd.

linera-core/src/client/mod.rs
afck: This is inside the `communicate_concurrently` closure: That function will return the first `Ok` result and ignore the errors. If all validators return `Err`, it returns `NodeError::EventsNotFound(remaining_event_ids)` (see line 714). So we are not using the errors returned from the closure anyway.

And in general, we don't want a single faulty validator to be able to cause issues in the client, not even an INFO-level log message! I guess I could add a DEBUG log here.

linera-core/src/client/mod.rs
afck: Same argument as above: A single validator sent us an invalid certificate, so we move on to the next validator.

(In the future we should add a bit of bookkeeping to the client, so that it doesn't ask faulty validators over and over again, but we don't have that yet.)

linera-storage/src/db_storage.rs
afck: I guess we could; needs discussion. Back when we designed this we didn't have an endpoint to request blocks by height yet, I think. But I'd rather not make such a change as part of this PR.

linera-storage/src/db_storage.rs
afck: Done in https://github.com/linera-io/linera-protocol/pull/5710/commits/ddcf1553220aa76639aab9f350392ce026c934d2.

linera-core/src/client/mod.rs
afck: Added more logging in https://github.com/linera-io/linera-protocol/pull/5710/commits/b5b4f13e91e39004902ef20e5a3143e257ba6a25.

linera-storage/src/db_storage.rs
MathieuDutSik: Ok.

linera-core/src/client/mod.rs
MathieuDutSik: Ok, I think I understand now.
Still the use of `()` is incredibly confusing and bad style.
Couldn't we have another unit type like `struct FailureAccess;`?
This should be done as well for `download_and_process_sender_chain` and `download_certificates_for_events`.

afck:
@MathieuDutSik: I tried to address your concerns in https://github.com/linera-io/linera-protocol/pull/5710/commits/2db23e2c7f19b5cbea60e87450db3283e33be42c. Please let me know whether you're okay with that.

PR #5711, opened 2026-03-16
[testnet] Allow daily claims from the faucet (#5611)

### LINKED ISSUE #5556: Daily airdrops from the faucet.

afck:
Backport of #5611.

## Motivation

Users that claim a chain from the faucet should receive tokens daily, so they can use the testnet indefinitely and pay the fees.

## Proposal

Add daily claims, similar to `pm-faucet`.

## Test Plan

A unit test was added.

## Release Plan

- Update the faucet.

## Links

- `main` version: #5611
- Closes #5556.

### DIFF
--- a/linera-faucet/server/src/tests.rs
+++ b/linera-faucet/server/src/tests.rs
@@ -23,7 +23,9 @@ use tempfile::TempDir;
 use tokio::sync::{oneshot, Notify};
 use tokio_util::sync::CancellationToken;
 
-use crate::database::FaucetDatabase;
+use crate::{
+    database::FaucetDatabase, BatchProcessor, BatchProcessorConfig, MutationRoot, PendingRequest,
+};
 
 struct ClientContext {
     client: ChainClient<environment::Test>,
@@ -81,12 +83,12 @@ impl chain_listener::ClientContext for ClientContext {
 
 /// Holds all components needed by faucet tests after common setup.
 struct FaucetTestEnv {
-    root: super::MutationRoot<environment::TestStorage>,
+    root: MutationRoot<environment::TestStorage>,
     context: Arc<Mutex<ClientContext>>,
     client: ChainClient<environment::Test>,
     clock: TestClock,
     faucet_storage: Arc<FaucetDatabase>,
-    pending_requests: Arc<Mutex<VecDeque<super::PendingRequest>>>,
+    pending_requests: Arc<Mutex<VecDeque<PendingRequest>>>,
     request_notifier: Arc<Notify>,
     storage_path: PathBuf,
     _temp_dir: TempDir,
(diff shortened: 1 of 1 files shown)


linera-faucet/server/src/lib.rs
```diff
@@ -201,15 +203,30 @@ pub struct MutationRoot<S> {
     pending_requests: Arc<Mutex<VecDeque<PendingRequest>>>,
     request_notifier: Arc<Notify>,
     storage: S,
+    /// Amount for initial claims (chain creation).
+    amount: Amount,
```
deuszx: Maybe then rename to `initial_claim_amount`?

linera-faucet/server/src/lib.rs
```diff
@@ -201,15 +203,30 @@ pub struct MutationRoot<S> {
...
+    /// Amount for daily claims (token transfer).
+    daily_claim_amount: Amount,
 }
 
-/// The result of a successful `claim` mutation.
-#[derive(SimpleObject)]
+/// The result of a successful `claim` or `dailyClaim` mutation.
+#[derive(Clone, Debug, SimpleObject)]
 pub struct ClaimOutcome {
```
deuszx, line 214: This seems to be duplicated at least in two other places – can't it be shared?

linera-faucet/server/src/lib.rs
```diff
@@ -201,15 +203,30 @@ pub struct MutationRoot<S> {
...
-    /// The hash of the parent chain's certificate containing the `OpenChain` operation.
+    /// The hash of the certificate containing the operation.
     pub certificate_hash: CryptoHash,
+    /// The amount of tokens transferred.
+    pub amount: Amount,
+}
+
+/// Information about a previous claim.
+#[derive(Clone, Debug, SimpleObject)]
+pub struct LastClaim {
```
deuszx: Same for this.

linera-faucet/server/src/lib.rs
```diff
@@ -285,25 +336,80 @@ where
...
+        )))
+    }
+}
 
-        #[cfg(with_metrics)]
-        metrics::DATABASE_OPERATION_LATENCY
-            .with_label_values(&["get_chain_id"])
-            .observe(db_start_time.elapsed().as_secs_f64() * 1000.0);
+/// Duration of one daily claim period in microseconds (24 hours).
+const DAILY_PERIOD_MICROS: u64 = 24 * 60 * 60 * 1_000_000;
```
deuszx: Or `Duration::from_hours(24).to_micros()` (I think it exists?). Not important ofc

linera-faucet/server/src/tests.rs
```diff
@@ -724,5 +705,123 @@ async fn test_blockchain_sync_after_database_deletion() {
...
+    );
+
+    let cancellation_token = CancellationToken::new();
+    let processor_task = {
+        let mut batch_processor = batch_processor;
+        let token = cancellation_token.clone();
+        tokio::spawn(async move { batch_processor.run(token).await })
+    };
+
+    let test_owner = AccountPublicKey::test_key(200).into();
```
deuszx, line 682: Up until this point – over 50 lines – is a setup that could be extracted to a helper function

deuszx (approved):

linera-faucet/server/src/lib.rs
afck: But one is in the server and one in the client crate. I don't think it's worth introducing a common crate just for these two structs. Also, the server one only needs `SimpleObject` and the client one only `Deserialize`.

deuszx: Client could depend on a server w/ a feature flag. That's what I'm doing in the bridge/relayer. Anyway, nitpick.

linera-faucet/server/src/lib.rs
afck: Or `TimeDelta::from_secs(24 * 60 * 60).as_micros()`, then we don't have to convert `u128` to `u64`.

ma2bd:
> DO we need to update runbooks/deployment scripts? daily claims amount defaults to 0 so deploying won't have any observable effect.

sure but not in this PR

linera-faucet/server/src/lib.rs
afck: Right, then the server would still need a library target and a separate binary target. Not sure if that's worth it, but if you prefer I can add it.

Issue #5453, opened 2026-02-17
Unified memory management for caches (inspired by Facebook's CacheLib)


ndr-ds:
### Problem

Our in-memory caches use **count-based limits** (number of items) rather than memory-based limits. With current defaults, a validator shard can OOM on a small machine even under normal load, because item sizes vary wildly and are not accounted for.

**Block cache** (`ValueCache<CryptoHash, Hashed<Block>>`): Default 5,000 items per shard. Each block can be up to ~1 MB (`maximum_block_size` in `policy.rs:288`). Worst case: **~5 GB per shard**. With 4 shards on one machine, that's 20 GB in block caches alone.

**Execution state cache** (`ValueCache<CryptoHash, ExecutionStateView<InactiveContext>>`): Default 10,000 items per shard. These are lightweight structural view templates (the `InactiveContext` wraps an `InactiveStore` that panics on access -- no materialized state), so memory impact is modest. Still count-based with no byte tracking.

**Wasm module cache** (`ModuleCache`): Already byte-weighted at 512 MiB -- the one cache that does this correctly.

**LRU prefix caches** (`LruPrefixCache`): Already byte-weighted at 10 MB per instance (`DEFAULT_STORAGE_CACHE_CONFIG` in `lru_caching.rs:106`), but a new instance is created per active chain via `load_chain` -> `open_exclusive` (`db_storage.rs:624`). With many active chains, these accumulate with no aggregate limit.

**The fundamental issue**: each cache is independently configured with no awareness of total available RAM, memory consumed by other caches, or co-location with other processes (proxy, ScyllaDB, other shards). There is no global memory budget. Defaults work on beefy machines but OOM on small ones.

### Proposed solution

Implement a **`MemoryController`** inspired by [Facebook's CacheLib](https://cachelib.org/) -- a battle-tested in-process caching engine used across Meta's infrastructure. CacheLib's design is documented in the [OSDI '20 paper](https://www.usenix.org/system/files/osdi20-berg.pdf), the [pool partitioning guide](https://cachelib.org/docs/Cache_Library_User_Guides/Partition_cache_into_pools/), and the [pool rebalance strategy guide](https://cachelib.org/docs/Cache_Library_User_Guides/pool_rebalance_strategy/).

[...]

PR #5714, opened 2026-03-16
Extract caching into linera-cache crate with concurrent ValueCache

ndr-ds:
## Motivation

`ValueCache` in `linera-core` used `Mutex<LruCache>` which serializes all cache access
across tokio tasks. This is a bottleneck for the block cache under concurrent workloads.
Additionally, `ValueCache` lived in `linera-core`, making it unusable from
`linera-storage` (circular dependency), which blocks adding a blob cache at the storage
layer.

## Proposal

Create a new `linera-cache` crate with two purpose-built cache types:

- **`ValueCache<K, V>`** — backed by `quick_cache::sync::Cache`. Uses sharded locks for
concurrent reads and S3-FIFO eviction for better hit ratios than LRU. Requires `V: Clone`. Used for the block cache (and future blob cache).
- **`UniqueValueCache<K, V>`** — backed by `Mutex<LruCache>`. For values that don't
implement `Clone`, where the access pattern is insert → remove (destructive read) rather
than insert → get (clone). Used for the execution state cache.

`linera-core` now imports directly from `linera_cache` — the old
`linera-core/src/value_cache.rs` re-export file and its test file are deleted. All tests
(14 total: 8 for ValueCache, 6 for UniqueValueCache) live inline in `linera-cache`.  
  
This `linera-cache` crate can become a `linera-memory` crate in the future, once we tackle https://github.com/linera-io/linera-protocol/issues/5453

## Test Plan

- cargo test -p linera-cache                                                    # 14 tests pass
- CI


linera-cache/src/value_cache.rs
```diff
@@ -0,0 +1,302 @@
...
+//! A concurrent cache with efficient eviction.
+
+#[cfg(with_metrics)]
+use std::any::type_name;
+use std::{borrow::Cow, hash::Hash};
+
+use linera_base::{crypto::CryptoHash, hashed::Hashed};
+use quick_cache::sync::Cache;
+
+#[cfg(with_metrics)]
```
deuszx: Personally I like the main structs to be at the top of the file (here the `ValueCache`).

Cargo.toml
```diff
@@ -184,7 +186,9 @@ prometheus = "0.13.3"
 prometheus-parse = "0.2.5"
 proptest = { version = "1.6.0", default-features = false, features = ["alloc"] }
 prost = "^0.14"
-quick_cache = "0.6.13"
+quick_cache = { version = "0.6.13", default-features = false, features = [
```
deuszx: latest version is 0.6.20

linera-cache/src/value_cache.rs
```diff
@@ -0,0 +1,302 @@
...
+
+/// A concurrent cache with efficient eviction.
+///
+/// Backed by `quick_cache` which uses sharded locks for concurrent reads
+/// and S3-FIFO eviction for better hit ratios than LRU.
+pub struct ValueCache<K, V>
+where
+    K: Hash + Eq + Clone + Send + Sync + 'static,
+    V: Clone + Send + Sync + 'static,
+{
```
deuszx: `Cache` struct itself does not require all those trait bounds for `K` and `V` – I think the "good practice" is to move them to `impl` blocks.

linera-cache/src/value_cache.rs
```diff
@@ -0,0 +1,302 @@
...
+}
+
+#[cfg(with_testing)]
+impl<K, V> ValueCache<K, V>
+where
+    K: Hash + Eq + Clone + Send + Sync + 'static,
+    V: Clone + Send + Sync + 'static,
+{
+    /// Returns `true` if the cache contains an entry for the given key.
+    pub fn contains(&self, key: &K) -> bool {
```
deuszx: Seems like a useful function to have – why not promote it to a normal `impl`? i.e. not `with_testing` only

linera-cache/src/value_cache.rs
```diff
@@ -0,0 +1,302 @@
...
+    /// Creates a new `ValueCache` with the given capacity.
+    pub fn new(size: usize) -> Self {
+        ValueCache {
+            cache: Cache::new(size),
+        }
+    }
+
+    /// Inserts a value into the cache if the key is not already present.
+    ///
+    /// Returns `true` if the value was newly inserted.
```
deuszx, line 35: Why `_owned`? There's nothing in the API that suggests this.

linera-cache/src/value_cache.rs
```diff
@@ -0,0 +1,302 @@
...
+    }
+}
+
+impl<T: Clone + Send + Sync + 'static> ValueCache<CryptoHash, Hashed<T>> {
+    /// Inserts a [`Hashed<T>`] into the cache if its hash is not already present.
+    ///
+    /// The `value` is wrapped in a [`Cow`] so that it is only cloned if it needs to be
+    /// inserted in the cache.
+    ///
+    /// Returns `true` if the value was not already present in the cache.
```
deuszx, line 87: `T` doesn't have to be `Clone` if it's inside `Cow<_>`. `into_owned` does not require the type to be cloneable

linera-cache/src/value_cache.rs
ma2bd: yes please! no where-clauses on structs unless you have to (usually because of other where-clauses on structs!!)

linera-cache/src/value_cache.rs
ndr-ds: `Cow::into_owned()` requires `Hashed<T>: Clone → T: Clone`, so I don't think I can do that one

deuszx: Ah, it needs `T: ToOwned` which often is implemented via `Clone`.

linera-cache/src/value_cache.rs
deuszx: Still, I'd move metrics down below. `impl<K, V> ValueCache<K, V>` is more interseting

PR #5715, opened 2026-03-16
[testnet] Fix non-empty inboxes

ma2bd:
## Motivation

Actually enable `nonempty_inboxes` after #5709

## Proposal

Add missing `save()` call

## Test Plan

CI + manually tested


afck:
Ah, good catch! :facepalm:

PR #5717, opened 2026-03-17
Circuit breaker for validator notification streams

### LINKED ISSUE #5699: Wall of logs with `linera service`
See https://github.com/linera-io/pm-app/issues/442

Image

ndr-ds:
## Motivation

Closes #5699.

PR #5619 fixed zombie `AbortHandle` entries by calling `abort()` when a notification
stream
ends naturally, but this creates an infinite re-queue cycle: retries exhaust (~1-2 min
of gRPC
backoff) → abort → entry removed → fresh subscription → retries exhaust → repeat
forever, each
cycle producing a burst of retry logs ("wall of logs").

## Proposal

Implement a per-validator [circuit
breaker](https://martinfowler.com/bliki/CircuitBreaker.html)
in `update_notification_streams`. The pattern follows the classic Closed → Open →
Half-Open
state machine originated by Michael Nygard (_Release It!_, 2007) and documented by
[Martin Fowler](https://martinfowler.com/bliki/CircuitBreaker.html),
[AWS](https://docs.aws.amazon.com/prescriptive-guidance/latest/cloud-design-patterns/cir
cuit-breaker.html),
and [Microsoft
Azure](https://learn.microsoft.com/en-us/azure/architecture/patterns/circuit-breaker).

The probe interval escalates with exponential backoff on repeated failures, matching the
`resetTimeout × exponentialBackoffFactor` refinement used by libraries like
[Arrow's CircuitBreaker](https://arrow-kt.io/learn/resilience/circuitbreaker/).

**Behavior:**

| State | Condition | Action |
| --- | --- | --- |
| Closed | Stream healthy | Normal notification processing |
| → Open | Stream ends (retries exhausted) | One error log; no further subscription attempts until probe time |
| Open | Probe time not reached | Validator skipped entirely (no work, no logs) |
| → Half-Open | Probe time reached | Single subscription attempt |
| Half-Open → Closed | Probe succeeds (stream stays alive) | Info log; normal operation resumes |
| Half-Open → Open | Probe fails (stream ends again) | Warn log; probe interval doubles (5m → 10m → 20m → ... → 1h cap) |

**New CLI arguments** (with sensible defaults):

- `--notification-circuit-breaker-initial-probe-interval-ms` (default: 300000 = 5 min)
- `--notificatiterval-ms` (default: 3600000 = 1 hour)

## Test Plan

CI

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -1908,6 +1908,18 @@ pub struct ChainClientOptions {
     /// Whether to allow creating blocks in the fast round. Fast blocks have lower latency but
     /// must be used carefully so that there are never any conflicting fast block proposals.
     pub allow_fast_blocks: bool,
+    /// Initial probe interval for the notification circuit breaker. When a validator's
+    /// notification stream exhausts retries, the circuit breaker waits this long before
+    /// probing again. Doubles on each failed probe.
+    pub notification_circuit_breaker_initial_probe_interval: Duration,
+    /// Maximum probe interval for the notification circuit breaker. The probe interval
+    /// doubles on each failure but is capped at this value.
+    pub notification_circuit_breaker_max_probe_interval: Duration,
+}
+
+struct CircuitBreakerState {
+    next_probe_at: Instant,
+    probe_interval: Duration,
 }
 
 pub static DEFAULT_CERTIFICATE_DOWNLOAD_BATCH_SIZE: u64 = 500;
@@ -1932,6 +1944,8 @@ impl ChainClientOptions {
             sender_certificate_download_batch_size: DEFAULT_SENDER_CERTIFICATE_DOWNLOAD_BATCH_SIZE,
             max_joined_tasks: 100,
             allow_fast_blocks: false,
+            notification_circuit_breaker_initial_probe_interval: Duration::from_secs(300),
+            notification_circuit_breaker_max_probe_interval: Duration::from_secs(3600),
         }
     }
 }
@@ -4593,7 +4607,8 @@ impl<Env: Environment> ChainClient<Env> {
             }
         }
 
-        let mut senders = HashMap::new(); // Senders to cancel notification streams.
+        let mut senders = HashMap::new();
+        let mut circuit_breakers: HashMap<ValidatorPublicKey, CircuitBreakerState> = HashMap::new();
         let notifications = self.subscribe()?;
         let (abortable_notifications, abort) = stream::abortable(self.subscribe()?);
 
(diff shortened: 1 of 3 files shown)


linera-core/src/client/mod.rs
```diff
@@ -4675,20 +4700,74 @@ impl<Env: Environment> ChainClient<Env> {
                 .collect();
             (nodes, self.client.local_node.clone())
         };
-        // Drop removed validators.
+
+        // Detect circuit breaker state transitions before cleaning up senders.
```
afck, line 4704: Why bother, for the ones that get removed anyway?

Maybe the `senders.retain` loop below should now remove only if `is_aborted && !contains_key`?

linera-core/src/client/mod.rs
```diff
@@ -4650,11 +4669,17 @@ impl<Env: Environment> ChainClient<Env> {
         Ok((update_streams, AbortOnDrop(abort), notifications))
     }
 
-    #[instrument(level = "trace", skip(senders))]
+    #[instrument(level = "trace", skip(senders, circuit_breakers))]
     async fn update_notification_streams(
         &self,
         senders: &mut HashMap<ValidatorPublicKey, AbortHandle>,
+        circuit_breakers: &mut HashMap<ValidatorPublicKey, CircuitBreakerState>,
```
afck, line 4676: (Could also be a single map, with the optional circuit breaker info and the abort handle. Not sure if that's cleaner.)

linera-core/src/client/mod.rs
ndr-ds: Then it would not have a vacant entry further down, which would stop the probing from happening, and I thought adapting the code for that sounded more complex 🤔

linera-core/src/client/mod.rs
ndr-ds: I fee like this would also make the code more complex 🤔

ndr-ds:
They operate at different timescales and aren't redundant:

- The gRPC retry will handle transient per-RPC failures, like connection resets, or other types of brief unavailability due to validator resets, etc. It's the configured number of retries with exponential backoff, capped at 30s. Gets exhausted in a few minutes.
- The Circuit breaker handles longer-term outages. Once gRPC retries exhaust, the stream ends, and without the circuit breaker the next `update_notification_streams` call would immediately create a fresh subscription, which would gRPC retries exhaust again... rinse and repeat. That's what generates the "wall of logs" from #5699.

After the gRPC gives up, instead of giving up completely on the subscription (previous behavior), or just retrying forever aggressively (current behavior), this circuit breaker approach is a good middle ground where we retry, and if we determine the validator is down or with issues, we go into a different timescale of retries, which will make us not lose the subscription (if the validator eventually comes back, we'll eventually reconnect to it), and also not waste resources/pollute logs with aggressive retries.

deuszx:
Still, you shouldn't need the _internal_ loop. I don't undertand how you can distinguish "brief reset" (internal retry) from "linger-term outage" (external). It's only about how long and how frequently the validator is unavailable. Both can be handled by the circuit breaker with proper configuration.

ndr-ds:
Sure, any one loop can probably be tweaked to handle the whole thing, but I think it's better to have this separation of concerns. Gonna merge this for now to clean up the errors, but we can follow up and iterate on the design here if you guys want

PR #5718, opened 2026-03-17
Add global blob cache to DbStorage

ndr-ds:
## Motivation

`load_contract` → `describe_application` → `read_blob` hits the database on every call.
For steady-state workloads with a single app, the same ~3 blobs (description, contract
bytecode, service bytecode) are re-read thousands of times per block. Blobs are
content-addressed and immutable — ideal for caching.

Stacked on #5714.

## Proposal

- Add a `blob_cache: Arc<ValueCache<BlobId, Blob>>` field to `DbStorage`, using the
`ValueCache` from `linera-cache` (PR #5714).
- Intercept `read_blob()` to check the cache before hitting the database, and populate
on miss.
- Add `--blob-cache-size` CLI argument (default 1000) to `CommonStorageOptions`,
threaded through `run_with_storage`, `run_with_store`, and the `RunnableWithStore` trait
so every code path receives the configured value consistently.
- `Blob` clone is cheap (`Arc<Box<[u8]>>` refcount bump), `BlobId` is `Copy` — no
overhead concern.

## Test Plan

- CI

This change seems to make us be able to keep the inboxes empty MUCH better.
On the same network, benchmark with the same exact TPS without this PR and then with this PR (on both the workers and validator), for around the same amount of time, this is how the inbox size looked:

Before:
Screenshot 2026-03-18 at 02 13 50

After:
Screenshot 2026-03-18 at 02 13 20

## Release Plan

### DIFF
--- a/CLI.md
+++ b/CLI.md
@@ -260,6 +260,9 @@ Client implementation and command-line tool for the Linera blockchain
 * `--storage-max-cache-find-key-values-size <STORAGE_MAX_CACHE_FIND_KEY_VALUES_SIZE>` — The maximal memory used in the find_key_values_by_prefix cache
 
   Default value: `10000000`
+* `--blob-cache-size <BLOB_CACHE_SIZE>` — The maximal number of entries in the blob cache
+
+  Default value: `1000`
 * `--storage-replication-factor <STORAGE_REPLICATION_FACTOR>` — The replication factor for the keyspace
 
   Default value: `1`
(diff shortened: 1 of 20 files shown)


deuszx:
> Add `--blob-cache-size` CLI argument (default 1000) 

Maybe we should cap cache by memory, not # of elements.

ndr-ds:
Good point, but I think I would rather keep complexity of this change low, and address this stuff once we do https://github.com/linera-io/linera-protocol/issues/5453

deuszx:
> Good point, but I think I would rather keep complexity of this change low, and address this stuff once we do #5453

Does it have to wait? We've been hit by this a couple of times already, why wait

ndr-ds:
Trying to get this in quickly, we can iterate later

PR #5724, opened 2026-03-17
Simplify faucet and tests.

afck:
Port of #5711.

## Motivation

In the backport #5711 of #5611 some simplifications were made that were not present in #5611.

## Proposal

Port those back to `main` again.

## Test Plan

CI

## Release Plan

## Links

- Follow-up to #5611.
- Port of #5711.

### DIFF
--- a/linera-faucet/server/src/tests.rs
+++ b/linera-faucet/server/src/tests.rs
@@ -3,7 +3,7 @@
 
 #![allow(clippy::large_futures)]
 
-use std::{collections::VecDeque, sync::Arc};
+use std::{collections::VecDeque, path::PathBuf, sync::Arc};
 
 use futures::lock::Mutex;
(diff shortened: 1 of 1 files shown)


linera-faucet/server/src/tests.rs
deuszx, line 1: I'm on a phone so reading big diffs is bit harder but it seems that you are not using the new FaucetTestEnv struct in all the tests? Understandably, some tests need a slightly different setup from others but maybe it's all matter of proper constructor?

afck: Yes, `test_batch_size_reduction_on_limit_errors` isn't really a good fit for the env: It needs a non-spawned `BatchProcessor`, a different resource policy, and its setup is much simpler anyway overall.

PR #5725, opened 2026-03-17
Allow clients to only listen to events from certain apps

### LINKED ISSUE #5708: Allow clients (notably web clients) to only listen to event subscriptions and messages of certain apps
The set of apps can change so unwanted event subscriptions and messages can only be paused (not rejected or skipped)

bart-linera:
## Motivation

There is an issue on the testnet where events from old versions of the application overwhelm clients.

## Proposal

Allow the clients to whitelist applications from which they want to receive events, so that they will only receive ones from the latest version.

## Test Plan

The event streams test was extended to exercise the new setting in the message policy.
CI will catch other regressions.

## Release Plan

## Links

- Closes #5708 

### DIFF
--- a/linera-core/src/unit_tests/wasm_client_tests.rs
+++ b/linera-core/src/unit_tests/wasm_client_tests.rs
@@ -24,8 +24,8 @@ use hex_game::{HexAbi, Operation as HexOperation, Timeouts};
 use linera_base::{
     crypto::{CryptoHash, InMemorySigner},
     data_types::{
-        Amount, BlanketMessagePolicy, BlobContent, BlockHeight, Bytecode, ChainDescription, Event,
-        MessagePolicy, OracleResponse, Round, TimeDelta, Timestamp,
+        Amount, BlanketMessagePolicy, BlobContent, BlockHeight, Bytecode, ChainDescription, Epoch,
+        Event, MessagePolicy, OracleResponse, Round, TimeDelta, Timestamp,
     },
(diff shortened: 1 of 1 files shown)


linera-client/src/client_options.rs
```diff
@@ -168,6 +149,11 @@ pub struct Options {
     #[arg(long, value_parser = util::parse_app_set)]
     pub reject_message_bundles_with_other_application_ids: Option<HashSet<GenericApplicationId>>,
 
+    /// A set of application IDs. If specified, only events coming from streams created by
+    /// applications from this set will be processed.
+    #[arg(long, value_parser = util::parse_app_set)]
+    pub process_events_from_application_ids: Option<HashSet<GenericApplicationId>>,
```
afck, line 155: Not sure if we should allow disabling the system application. Then we don't get epoch changes anymore. (And it's easy to forget!)
What do you think, @ma2bd?

bart-linera: Epoch changes are collected separately, in `collect_epoch_changes`, so they should be safe from this filter. But maybe we should still specifically include system events in the `filter` call, just in case :thinking:

afck: Ah, right, then maybe it's fine as it is!

bart-linera: I'll extend the test to assert that this is indeed working as expected.

PR #5726, opened 2026-03-17
Improve the access to contracts.

MathieuDutSik:
## Motivation

Code examination revealed that there are some inefficiencies in the handling of the contract code:
* The metering is applied even when the application is already in the cache. Clear loss.
* The code is serialized and then deserialized when constructing an instance.

## Proposal

Put the metering inside of the construction of entries in the cache. That part is clear.
Then simplify the construction of the engine.

## Test Plan

CI.

## Release Plan

Can be potentially released in `testnet_conway`.

## Links

None

### DIFF
--- a/linera-execution/src/wasm/wasmer.rs
+++ b/linera-execution/src/wasm/wasmer.rs
@@ -14,9 +14,10 @@ use tokio::sync::Mutex;
 use tracing::instrument;
 
 use super::{
+    add_metering,
     module_cache::ModuleCache,
     runtime_api::{BaseRuntimeApi, ContractRuntimeApi, RuntimeApiData, ServiceRuntimeApi},
-    add_metering, ContractEntrypoints, ServiceEntrypoints, WasmExecutionError,
+    ContractEntrypoints, ServiceEntrypoints, WasmExecutionError,
 };
 use crate::{
     wasm::{WasmContractModule, WasmServiceModule},
@@ -53,8 +54,7 @@ static CONTRACT_ENGINE: LazyLock<wasmer::Engine> = LazyLock::new(|| {
 });
 
 /// A cache of compiled contract modules.
-static CONTRACT_CACHE: LazyLock<Mutex<ModuleCache<wasmer::Module>>> =
-    LazyLock::new(Mutex::default);
+static CONTRACT_CACHE: LazyLock<Mutex<ModuleCache<wasmer::Module>>> = LazyLock::new(Mutex::default);
 
 /// A cache of compiled service modules.
 static SERVICE_CACHE: LazyLock<Mutex<ModuleCache<wasmer::Module>>> = LazyLock::new(Mutex::default);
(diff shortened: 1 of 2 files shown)


linera-execution/src/wasm/wasmer.rs
```diff
@@ -191,50 +212,3 @@ impl From<wasmer::RuntimeError> for ExecutionError {
...
-
-/// Serialized bytes of a compiled contract bytecode.
-// Cloning `Module`s is cheap.
-#[derive(Clone)]
-pub struct CachedContractModule(wasmer::Module);
-
-impl CachedContractModule {
-    /// Creates a new [`CachedContractModule`] by compiling a `contract_bytecode`.
-    pub fn new(contract_bytecode: Bytecode) -> Result<Self, anyhow::Error> {
-        let module = wasmer::Module::new(&Self::create_compilation_engine(), contract_bytecode)?;
```
ma2bd, line 203: Previously we tried very hard to avoid sharing engines between modules but that was not necessary?

MathieuDutSik: The engine is definitely shared for SERVICE. So, I am not sure why we did that.

ma2bd: This was apparently related to native Metering (which we don't use any more), ironically.

https://docs.rs/wasmer-middlewares/4.4.0/wasmer_middlewares/metering/struct.Metering.html#panic

f9c55f62fcf5a94e74d38a2df36c1c5b8be68b15

Twey:
If we're okay with sharing the module now then this looks good to me. The circumlocution here was all just to avoid that IIRC.

ma2bd:
> If we're okay with sharing the module now then this looks good to me. The circumlocution here was all just to avoid that IIRC.

Ok but what was the concern about sharing the module? (And more importantly, does it still apply?)

Twey:
If I remember correctly, modules compiled with one backend must be executed with the same backend. But if that's a problem it will show up in tests. It's not a problem on the Web, where we use the JavaScript WebAssembly API for both.

ma2bd:
> If I remember correctly, modules compiled with one backend must be executed with the same backend. But if that's a problem it will show up in tests. It's not a problem on the Web, where we use the JavaScript WebAssembly API for both.

What do you mean by "backend" here? FWIW the CI is green. (And also on the backport #5728)

Twey:
> What do you mean by "backend" here?

Sorry, I think I mean the `wasmer::Engine` (dredging up old memories :)).  We used to have a separate engine for compilation vs execution, and it's illegal in Wasmer land to take a `Module` created by the former and execute it on the latter, but going through a serialization roundtrip allows the module to work with the new `Engine`.  With this PR we have just the one `Engine` so it should be fine.

> FWIW the CI is green. (And also on the backport https://github.com/linera-io/linera-protocol/pull/5728)

Hence my approval :)  If we had hit this I'm pretty confident the tests would have failed (we'd panic when executing a contract).

PR #5727, opened 2026-03-17
Step 4 of clippy corrections

### LINKED ISSUE #5573: Enable clippy lints for bugs, perf issues, and API problems

MathieuDutSik:
## Motivation

More clippy corrections that should be catch are added to `Cargo.toml`.

Fixes #5573 

## Proposal

Do it in a very straightforward way.

## Test Plan

CI

## Release Plan

We could be added to `testnet_conway`.

## Links

None

### DIFF
--- a/examples/matching-engine/src/state.rs
+++ b/examples/matching-engine/src/state.rs
@@ -393,7 +393,7 @@ impl LevelView {
         let mut remove_order = Vec::new();
         let orders = self
             .queue
-            .iter_mut()
+            .try_iter_mut()
             .await
             .expect("Failed to load iterator over orders");
         for order in orders {
@@ -528,7 +528,7 @@ impl LevelView {
         // If some order has amount zero but is after an order of non-zero amount, then it is left.
         let iter = self
             .queue
-            .iter_mut()
+            .try_iter_mut()
             .await
             .expect("Failed to load iterator over level queue");
         let n_remove = iter
@@ -550,7 +550,7 @@ impl LevelView {
     ) -> Option<(Amount, bool)> {
         let mut iter = self
             .queue
-            .iter_mut()
+            .try_iter_mut()
             .await
             .expect("Failed to load iterator over level queue");
         let state_order = iter.find(|order| order.order_id == order_id)?;


linera-views/src/views/queue_view.rs
```diff
@@ -464,6 +464,7 @@ where
     /// assert_eq!(queue.elements().await.unwrap(), vec![42]);
     /// # })
     /// ```
+    #[allow(clippy::iter_not_returning_iterator)]
```
afck: Should it be `try_iter_mut` then?

linera-views/src/backends/rocks_db.rs
```diff
@@ -715,15 +715,16 @@ pub struct PathWithGuard {
     pub path_buf: PathBuf,
     /// The guard for the directory if one is needed
     #[serde(skip)]
-    _dir: Option<Arc<TempDir>>,
+    #[allow(dead_code)]
```
afck: So on the one hand it _is_ dead code, but on the other hand it would trigger the `used_underscore_binding` lint? 🤨

linera-service/src/storage.rs
```diff
@@ -445,6 +445,7 @@ example service:tcp:127.0.0.1:7878:table_do_my_test"
 }
 
 impl StorageConfig {
+    #[allow(clippy::used_underscore_binding)]
     pub fn maybe_append_shard_path(&mut self, _shard: usize) -> std::io::Result<()> {
         match &mut self.inner_storage_config {
             #[cfg(all(feature = "rocksdb", feature = "scylladb"))]
```
afck, line 451: Maybe at this point it's clearer to just have two `fn maybe_append_shard_path`s: One with RocksDB/ScyllaDB and one without. The latter would then still have `_shard` with underscore and without `allow`.

linera-execution/src/wasm/runtime_api.rs
```diff
@@ -842,7 +842,8 @@ where
     ///
     /// This is called by the metering instrumentation, but the fuel consumed argument is
     /// ignored.
-    fn check_execution_time(caller: &mut Caller, _fuel_consumed: u64) -> Result<(), RuntimeError> {
+    fn check_execution_time(caller: &mut Caller, fuel_consumed: u64) -> Result<(), RuntimeError> {
+        let _ = fuel_consumed;
```
afck: We're trying to avoid `let _ =` in general, too.

linera-views/src/backends/rocks_db.rs
MathieuDutSik: Yes, that is the problem.

But there is a solution. Writing `_dir_guard: dir_guard`.

linera-service/src/storage.rs
MathieuDutSik: I think we have to choose a poison. And here the one I would prefer is the `#[allow(_)]` statement.

linera-execution/src/wasm/runtime_api.rs
MathieuDutSik: Here the poison I would choose is the allow and yes remove the `let _ = ....`.

linera-service/src/cli/net_up_utils.rs
```diff
@@ -34,7 +34,8 @@ struct StorageConfigProvider {
     /// The storage config.
     config: StorageConfig,
     #[cfg(feature = "storage-service")]
-    _service_guard: Option<StorageServiceGuard>,
+    #[allow(dead_code)]
+    service_guard: Option<StorageServiceGuard>,
```
afck: So why does it complain about the underscore here?

MathieuDutSik: It is the same pattern as for the `dir_guard`.

linera-views/src/views/queue_view.rs
MathieuDutSik: Yes, that works.

PR #5731, opened 2026-03-18
Add per-query TTL for subscription invalidation deferral

ma2bd:
## Motivation

When using `--allow-subscription` to register named GraphQL subscription queries, every new block triggers an immediate re-execution of the query. For high-throughput chains, this can be expensive, and some queries may not need to be real-time.

## Proposal

* Add a new repeatable CLI option `--subscription-ttl-secs <Name>=<Secs>` that sets a minimum TTL per registered subscription query.
* When a `NewBlock` notification arrives before the TTL of the last execution has expired, the re-execution is deferred until the remaining TTL elapses.
* Multiple notifications during the TTL window are coalesced into a single deferred re-execution.

Queries without a TTL configured retain the existing behavior (immediate re-execution on every new block).

## Test Plan

- Added `test_wasm_end_to_end_counter_subscription_ttl` integration test
- Existing `test_wasm_end_to_end_counter_subscription`

## Release Plan

This is a testnet PR and it should be ported to `main`.


linera-service/tests/linera_net_tests.rs
```diff
@@ -1616,6 +1617,95 @@ async fn test_wasm_end_to_end_counter_subscription(config: impl LineraNetConfig)
...
+        .await
+        .context("expected initial query result")??;
+    let initial_value: u64 = serde_json::from_value(initial["data"]["value"].clone())?;
+    assert_eq!(initial_value, original_counter_value);
+
+    // Increment the counter. With a TTL, the notification should be deferred.
+    let mutation = format!("increment(value: {increment})");
+    application.mutate(mutation).await?;
+
+    // The update should NOT arrive within 2 seconds (TTL is 4s).
```
afck, line 1685: I'm a bit scared of adding more tests with timing.

But I guess we can revisit this if it turns out to be flaky in CI.

PR #5734, opened 2026-03-18
Make RocksDB config cgroup-aware to prevent OOM in containers

ndr-ds:
## Motivation

pm-infra-worker pods are OOMing on `kubernetes-admin@infra-k8s`. One root cause is that
RocksDB's block cache and background thread configuration uses host machine resources
(`/proc/meminfo` and host CPU count) instead of the cgroup limits imposed by Kubernetes.
On c3-64 nodes (~64 GiB RAM, 64 cores), a pod with a 12 GiB memory limit allocates a 16
GiB block cache (`total_ram / 4`), which alone exceeds the pod's entire memory limit.

## Proposal

- Add `get_available_memory()` that prefers `sys.cgroup_limits().total_memory` with
fallback to `sys.total_memory()` for non-containerized environments.
- Add `get_available_cpus()` using `std::thread::available_parallelism()` which respects
cgroup v2 CPU quotas, replacing `sys.cpus().len()` which returns host CPU count.
- Reduce block cache from `total_ram / 4` to `total_ram / 6`.
- Add `WriteBufferManager` to cap total memtable memory at `total_ram / 4` across all
column families (requires rocksdb 0.22.0).
- Reduce individual memtable size from 256 MiB to 64 MiB for more frequent flushes.
- Switch compaction from Level to Universal.
- Halve background jobs to `(num_cpus / 2).max(4)` instead of `num_cpus`.

With these changes, a pod with a 12 GiB limit will use ~2 GiB for block cache and ~3 GiB
for memtables, leaving headroom for chain state and application logic.

## Test Plan

CI

### DIFF
--- a/linera-views/src/backends/rocks_db.rs
+++ b/linera-views/src/backends/rocks_db.rs
@@ -14,9 +14,9 @@ use std::{
 };
 
 use linera_base::ensure;
-use rocksdb::{BlockBasedOptions, Cache, DBCompactionStyle, SliceTransform};
+use rocksdb::{BlockBasedOptions, Cache, DBCompactionStyle, SliceTransform, WriteBufferManager};
 use serde::{Deserialize, Serialize};
-use sysinfo::{CpuRefreshKind, MemoryRefreshKind, RefreshKind, System};
+use sysinfo::{MemoryRefreshKind, RefreshKind, System};
 use tempfile::TempDir;
 use thiserror::Error;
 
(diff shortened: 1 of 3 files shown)


linera-views/src/backends/rocks_db.rs
```diff
@@ -340,46 +352,52 @@ impl RocksDbStoreInternal {
...
+        let num_cpus = get_available_cpus();
+        let total_ram = get_available_memory(&sys);
+
         let mut options = rocksdb::Options::default();
         options.create_if_missing(true);
         options.create_missing_column_families(true);
-        // Flush in-memory buffer to disk more often
-        options.set_write_buffer_size(WRITE_BUFFER_SIZE);
+
+        let memtable_size = 64 * 1024 * 1024;
```
deuszx: can you move it to a const as well?

linera-views/src/backends/rocks_db.rs
```diff
@@ -340,46 +352,52 @@ impl RocksDbStoreInternal {
...
+        options.set_min_write_buffer_number_to_merge(2);
+        options.set_compression_per_level(&[
+            rocksdb::DBCompressionType::None,
+            rocksdb::DBCompressionType::None,
+            rocksdb::DBCompressionType::None,
+            rocksdb::DBCompressionType::None,
+            rocksdb::DBCompressionType::None,
+            rocksdb::DBCompressionType::None,
+            rocksdb::DBCompressionType::None,
+        ]);
```
deuszx: damn! que pasa? Is there really no other way? If it's all `None`s – isn't that a default?

ndr-ds: Ah, my bad! I think I was playing with it by level at some point, and forgot to clean it up and left it verbose like this 😅

PR #5735, opened 2026-03-18
Cap loaded chain workers with LRU eviction to bound memory

ndr-ds:
## Motivation

pm-infra-worker pods accumulate thousands of `ChainWorkerActor` instances without limit.
Each loaded chain worker holds a `ChainStateView` in memory, and with no cap on the
number of active workers, memory grows unboundedly as more chains are accessed. Combined
with a 30-minute TTL, this leads to OOM kills.

## Proposal

- Replace `BTreeMap<ChainId, ChainActorEndpoint>` with `lru::LruCache` in `WorkerState`,
enforcing a configurable maximum number of loaded chain workers. When the cap is
reached, the least-recently-used endpoint is evicted (sender dropped), causing its actor
to exit and release memory.
- Add `max_loaded_chain_workers` field to `ChainWorkerConfig` (default: 100).
- Add `--max-loaded-chain-workers` CLI arg to both `linera` client and `linera-server`
(env: `LINERA_MAX_LOADED_CHAIN_WORKERS`).
- Add `chain_worker_evictions` Prometheus counter to track eviction rate.

The `lru` crate is already a dependency of `linera-core`. Eviction is transparent to
callers — a new actor is spawned on the next request to an evicted chain, same as today
when a TTL-expired actor exits.

## Test Plan

CI


ndr-ds:
Closing this as we should measure memory usage after the chain actors are removed

PR #5737, opened 2026-03-18
Sparse-sync publisher chains on subscription

afck:
Port of #5653, and adds chain info query field.

## Motivation

When subscribing to a notification stream, chains should immediately get notified about the most recent events, even if they happened before the subscription.

## Proposal

Add a `request_previous_event_blocks` field to `ChainInfoQuery`. Use it to do a sparse sync for publisher chains on subscription.

## Test Plan

Tests were extended.

The social end-to-end test became flaky and had to be fixed by cherry-picking https://github.com/linera-io/linera-protocol/pull/5768 and applying the same logic in one more place.

## Release Plan

- Nothing to do.

## Links

- Fixes https://github.com/linera-io/linera-protocol/issues/5651
- Testnet version: #5653
- Endpoint on testnet (replaced by `ChainInfoQuery` here): https://github.com/linera-io/linera-protocol/pull/5659
- Test fix on the testnet: https://github.com/linera-io/linera-protocol/pull/5768

### DIFF
--- a/linera-service/template/Cargo.toml.template
+++ b/linera-service/template/Cargo.toml.template
@@ -16,6 +16,7 @@ darling = {{ version = ">=0.20, <0.23" }}
 serde_with = {{ version = ">=3, <3.18" }}
 time = {{ version = ">=0.3, <0.3.47" }}
 time-core = {{ version = ">=0.1, <0.1.8" }}
+unicode-segmentation = {{ version = ">=1, <1.13" }}
 
 [dev-dependencies]
 {linera_sdk_dev_dep}


linera-core/src/client/chain_client/mod.rs
```diff
@@ -791,28 +790,73 @@ impl<Env: Environment> ChainClient<Env> {
...
+            .collect::<Result<Vec<_>, _>>()?;
+        Ok(())
+    }
+
+    /// Downloads only event-bearing blocks for the given publisher chain and streams.
+    ///
+    /// Uses `communicate_with_quorum` so that each validator is queried for
+    /// `previous_event_blocks` and the claimed blocks are downloaded from that same
+    /// validator. Waiting for a quorum guarantees that any confirmed event is discovered,
+    /// since at least one honest validator in the quorum must have it.
```
bart-linera, line 839: I feel like this is wasteful, since we'll be trying to download the same things from multiple validators in parallel. Shouldn't this work more like synchronizing sender chains (where we only do communicate with quorum to discover what is there to be downloaded, and then download different batches from different validators)? Or does the RequestScheduler deduplicate requests? 🤔

afck: Yes, it should. (By key, which in this case contains all the height, I think, so it's not perfect. But I'd say if we want to improve on this, we should do that inside the request scheduler.)

MathieuDutSik (approved):
LGTM

PR #5740, opened 2026-03-18
Reduce retry noise when validators are unhealthy

ndr-ds:
## Motivation

When a validator is unhealthy (e.g. during pod restarts), clients produce a wall of
warn-level logs because:

1. Every chain independently retries subscriptions to the same dead validator (N chains
× retries).
2. The old `full_jitter_delay` could return 0ms (uniform random from `[0, cap]`),
allowing instant retries.
3. Per-retry errors were logged at `warn` level, flooding production logs at `info`
threshold.

Ref: #5699

## Proposal

**Jittered exponential backoff with guaranteed minimum delay**
(`linera-base/src/time.rs`):
- New `jittered_backoff_delay` function using the gRPC-recommended ±20% jitter,
guaranteeing at least 80% of the computed backoff. Replaces the old `full_jitter_delay`
which could return 0.

**Shared subscription cooldown** (`linera-rpc/src/grpc/client.rs`, `node_provider.rs`):
- A `papaya::HashMap<String, Instant>` shared across all `GrpcClient` instances from the
same `GrpcNodeProvider` tracks when each validator address last had a subscription
failure.
- Before connecting, `subscribe()` checks the cooldown — if the validator failed
recently (within `max_backoff`, default 30s), it fast-fails instead of opening another
connection.
- On successful reconnection the cooldown is cleared; on failure (initial connect or
retries exhausted) the cooldown is set.
- This reduces the burst from N×retries connections to ~1×retries when a validator goes
down.

**Log level downgrade** (`linera-rpc/src/grpc/client.rs`):
- Per-retry subscription errors downgraded from `warn` to `debug`. The circuit breaker
at the `ChainClient` level already logs meaningful state transitions at higher levels.

**Cross-chain message retry** (`cross_chain_message_queue.rs`):
- Replaced `full_jitter_delay` (exponential with full jitter) with linear backoff capped
at `max_backoff`. Linear backoff is more appropriate for server-to-server cross-chain
messages where thundering herd is not a concern and lower latency matters.

**Clean`linera-rpc/src/lib.rs`):
- Removed the now-unused `full_jitter_delay` function.

## Test Plan

CI

### DIFF
--- a/linera-rpc/src/grpc/client.rs
+++ b/linera-rpc/src/grpc/client.rs
@@ -18,7 +18,7 @@ use linera_base::{
     data_types::{BlobContent, BlockHeight, NetworkDescription},
     ensure,
     identifiers::{BlobId, ChainId, StreamId},
-    time::Duration,
+    time::{Duration, Instant},
 };
 use linera_chain::{
     data_types::{self},
@@ -50,7 +50,7 @@ use linera_core::{
 };
 use linera_version::VersionInfo;
 use tonic::{Code, IntoRequest, Request, Status};
-use tracing::{debug, instrument, trace, warn, Level};
+use tracing::{debug, instrument, trace, Level};
 
 use super::{
     api::{self, validator_node_client::ValidatorNodeClient, SubscriptionRequest},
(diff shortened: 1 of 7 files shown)


linera-rpc/src/grpc/client.rs
```diff
@@ -74,6 +74,10 @@ pub struct GrpcClient {
     retry_delay: Duration,
     max_retries: u32,
     max_backoff: Duration,
+    /// Shared across all `GrpcClient` instances created by the same `GrpcNodeProvider`.
+    /// Tracks when each validator address last had a subscription failure, so that
+    /// other chains don't independently retry the same dead validator.
+    subscription_cooldowns: papaya::HashMap<String, Instant>,
```
deuszx, line 80: We shouldn't be building a subscription request management independently from `RequestsScheduler` - that one was built specifically to address issues like this one.

ndr-ds: Does the `RequestScheduler` deal with subscription streams? 🤔

ndr-ds: I also thought that it was more for request/response type of stuff, not long lived subscription streams

PR #5741, opened 2026-03-19
Aggregate save view latency histograms before computing percentiles

ndr-ds:
## Motivation

The Save View Latency panel on the Views Grafana dashboard shows duplicate
p50/p90/p99 lines with no distinguishing labels. The `histogram_quantile`
call groups by `(le, type)`, which produces a separate percentile set per
view type — currently just `ChainStateView`, but the `byRegexp` color
overrides and `{{type}} pXX` legend format make the panel inconsistent with
every other latency panel in the dashboards.

## Proposal

Aggregate by `(le)` only (dropping the `type` dimension), matching the
pattern used by Load View Latency, Block Execution Latency, and all other
latency panels. Also switch the color overrides from `byRegexp` to exact
`byName` matching.

The live central monitoring dashboard has already been updated to match.

## Test Plan

CI. Verified on central monitoring that the panel now shows exactly 3 clean
lines (p50/p90/p99) with correct green/yellow/red coloring.

### DIFF
--- a/kubernetes/linera-validator/grafana-dashboards/linera/views.json
+++ b/kubernetes/linera-validator/grafana-dashboards/linera/views.json
@@ -207,7 +207,7 @@
         "type": "prometheus",
         "uid": "prometheus"
       },
-      "description": "Time to save a view to storage. p50 is median, p90 is most users' experience, p99 captures tail latency. Broken down by view type.",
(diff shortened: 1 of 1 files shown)


ndr-ds:
Already in central monitoring, so merging

PR #5742, opened 2026-03-19
Cache immutable data in DbStorage

ndr-ds:
## Motivation

All immutable data reads (confirmed blocks, certificates, events) hit the database on
every access, despite being immutable and safe to cache. Additionally, `contains_blob()`
and `missing_blobs()` bypass the existing blob cache entirely. This adds unnecessary DB
pressure on validators during steady-state operation.

## Proposal

Add tiered caching for all immutable data in `DbStorage`, modifying only
`db_storage.rs`:

- **Confirmed blocks**: `ValueCache<CryptoHash, ConfirmedBlock>` — checked before DB in
`read_confirmed_block()` and `read_confirmed_blocks()`.
- **Certificates (L1 deserialized)**: `ValueCache<CryptoHash,
ConfirmedBlockCertificate>` — `read_certificate()` checks L1 first.
- **Certificates (L2 raw bytes)**: `ValueCache<CryptoHash, RawCertificate>` — stores
`(lite_cert_bytes, confirmed_block_bytes)`. Deserialized callers go L1 → L2 → DB, raw
callers go L2 → DB.
- **Events**: `ValueCache<EventId, Vec<u8>>` — checked before DB in `read_event()`.
- **Blob existence**: `contains_blob()` and `missing_blobs()` now check the existing
`blob_cache` before querying DB.

All caches reuse the existing `blob_cache_size` parameter — no constructor signature
changes, no caller changes needed. Batch methods (`read_confirmed_blocks`,
`read_certificates_raw`) are cache-aware: they check the cache per item and only query
DB for misses.

Estimated memory: ~8-21 MB typical for 1000 items across all caches.

## Test Plan

- CI

### DIFF
--- a/linera-storage/src/db_storage.rs
+++ b/linera-storage/src/db_storage.rs
@@ -361,6 +361,9 @@ impl MultiPartitionBatch {
     }
 }
 
+/// Raw certificate bytes: (lite_certificate_bytes, confirmed_block_bytes).
+type RawCertificate = (Vec<u8>, Vec<u8>);
+
 /// Main implementation of the [`Storage`] trait.
 #[derive(Clone)]
 pub struct DbStorage<Database, Clock = WallClock> {
@@ -371,6 +374,10 @@ pub struct DbStorage<Database, Clock = WallClock> {
     user_contracts: Arc<papaya::HashMap<ApplicationId, UserContractCode>>,
     user_services: Arc<papaya::HashMap<ApplicationId, UserServiceCode>>,
     blob_cache: Arc<ValueCache<BlobId, Blob>>,
+    confirmed_block_cache: Arc<ValueCache<CryptoHash, ConfirmedBlock>>,
+    lite_certificate_cache: Arc<ValueCache<CryptoHash, LiteCertificate<'static>>>,
+    certificate_raw_cache: Arc<ValueCache<CryptoHash, RawCertificate>>,
+    event_cache: Arc<ValueCache<EventId, Vec<u8>>>,
     execution_runtime_config: ExecutionRuntimeConfig,
 }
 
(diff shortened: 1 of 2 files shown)


ma2bd:
Should we deactivate `linera-view`'s `lru_caching` entirely in shared mode then?

ndr-ds:
I think we might as well do that 🤔 It currently does nothing other than add overhead 😅

linera-storage/src/db_storage.rs
```diff
@@ -371,6 +374,10 @@ pub struct DbStorage<Database, Clock = WallClock> {
     user_contracts: Arc<papaya::HashMap<ApplicationId, UserContractCode>>,
     user_services: Arc<papaya::HashMap<ApplicationId, UserServiceCode>>,
     blob_cache: Arc<ValueCache<BlobId, Blob>>,
+    confirmed_block_cache: Arc<ValueCache<CryptoHash, ConfirmedBlock>>,
+    certificate_cache: Arc<ValueCache<CryptoHash, ConfirmedBlockCertificate>>,
+    certificate_raw_cache: Arc<ValueCache<CryptoHash, RawCertificate>>,
```
deuszx, line 379: So we store the same thing 3 times?

afck: I guess it does make sense to cache both the deserialized blocks and the serialized certificates. :confused: 

But I wouldn't cache `ConfirmedBlockCertificate`. Instead, this should just store the corresponding `LiteCertificate` and put it together with the corresponding cached block when requested.

linera-storage/src/db_storage.rs
```diff
@@ -1219,13 +1232,19 @@ where
         &self,
         hash: CryptoHash,
     ) -> Result<Option<ConfirmedBlock>, ViewError> {
+        if let Some(block) = self.confirmed_block_cache.get(&hash) {
+            return Ok(Some(block));
+        }
         let root_key = RootKey::ConfirmedBlock(hash).bytes();
         let store = self.database.open_shared(&root_key)?;
-        let value = store.read_value(BLOCK_KEY).await?;
+        let value: Option<ConfirmedBlock> = store.read_value(BLOCK_KEY).await?;
```
afck: I think this is more in line with our guidelines?

```suggestion
        let value = store.read_value::<Option<ConfirmedBlock>>(BLOCK_KEY).await?;
```

PR #5744, opened 2026-03-19
Improve the performance of the access to storage.

MathieuDutSik:
## Motivation

Access to storage is done in the `execution_chain_actor`. It always goes in two
steps of accessing first to the view and then loading doing the operation.
This is repetitive operation, already with the serialization. This is hitting us
quite hard because a contract would have a lot of read operations.

We used to have a cache of views in `ReentrantCollectionView`. But we removed it
because it is not great to put the cache in the views.

## Proposal

Put it in the `ExecutionStateActor`. We unfortunately cannot access by an index because
of long lived services. So, for any query, we have to access via a `BTreeMap<_,_>`.

The views are accessed via `Arc<RwLock<KeyValueStoreView<C>>>`. So, we expose
some of the internals of the `ReentrantCollectionView` to the user. That is a little
unfortunate.

Other approaches:
* Put the `Arc<RwLock<KeyValueStoreView<C>>>` in the `SyncRuntimeContract`. But it is a major change and for a start we do not even have a `Context` in `SyncRuntimeContract`. So `block_on` would be needed to use.
* Other idea is to store the `KeyValueStoreView` as a single `BTreeMap<Vec<u8>,Vec<u8>>` stored as a `RegisterView`. This is a somewhat orthogonal approach but still something to consider.

## Test Plan

CI.

## Release Plan

It can be put in `testnet_conway`.

## Links

None

### DIFF
--- a/linera-execution/src/execution_state_actor.rs
+++ b/linera-execution/src/execution_state_actor.rs
@@ -143,6 +143,19 @@ where
         Ok((code, description))
     }
 
+    /// Pre-warms the user state view for the given application into the cache,
+    /// so that subsequent storage operations don't need to reload from storage.
+    pub(crate) async fn pre_warm_user_state(
+        &mut self,
+        application_id: &ApplicationId,
+    ) -> Result<(), ExecutionError> {
+        self.state
+            .users
+            .try_load_entry_mut(application_id)
+            .await?;
+        Ok(())
+    }
+
     // TODO(#1416): Support concurrent I/O.
     #[instrument(
         skip_all,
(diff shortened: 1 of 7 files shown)


linera-views/src/views/reentrant_collection_view.rs
```diff
@@ -384,6 +390,16 @@ impl<W: View> ReentrantByteCollectionView<W::Context, W> {
...
     }
 
+    /// Loads a subview for the data at the given index in the collection and returns
+    /// the `Arc<RwLock<W>>` directly. If an entry is absent then a default entry is added
+    /// to the collection. This is useful for caching the Arc outside the collection.
+    pub async fn try_load_view_arc(
+        &mut self,
+        short_key: &[u8],
+    ) -> Result<Arc<RwLock<W>>, ViewError> {
+        self.try_load_view_mut(short_key).await
```
afck, line 400: If they are identical, why do we need both `try_load_view_mut` and `try_load_view_arc`?

linera-execution/src/execution_state_actor.rs
```diff
@@ -143,6 +152,20 @@ where
...
+    /// Returns the cached key-value store view for the given application,
+    /// loading it from storage if necessary.
+    async fn get_or_load_key_value_store(
+        &mut self,
+        application_id: &ApplicationId,
+    ) -> Result<Arc<RwLock<KeyValueStoreView<C>>>, ExecutionError> {
+        if let Some(arc) = self.key_value_stores.get(application_id) {
+            return Ok(arc.clone());
+        }
+        let arc = self.state.users.try_load_view_arc(application_id).await?;
```
afck, line 164: I'm probably still misunderstanding the underlying problem, but:

Now that we call `try_load_view_arc`/`try_load_view_mut` instead of just `try_load_view`, we _are_ inserting the child view into the reentrant collection view's internal map, aren't we? So is it even necessary to keep the map in the actor, too? Isn't just replacing `try_load_view` with `try_load_view_mut` enough to make the collection view effectively cache its children?

MathieuDutSik: It is very similar, but not exactly the same since we are instead storing the serialized key in the `ReentrantCollectionView` and the key itself in the `KeyValueStoreView`.

afck: And serializing the key (just an app ID, is it?) is taking that much time that it's worth introducing a separate map?

MathieuDutSik: The problem in my view is that there are a lot of storage calls. So, improving their speed is quite important. But yes, there is another way to address this: use the `ByteReentrantCollectionView` and pass the serialization.

But to be honest, I was hoping for more improvements: Avoiding the access to the entry in the `BTreeMap` altogether. But that proved impossible.

MathieuDutSik:
That PR brings benefit, but they are small. So, I prefer to kill it.

Issue #5722, opened 2026-03-17
`test_end_to_end_repeated_transfers::remote_net_grpc` is flaky.


afck:
Disabled in https://github.com/linera-io/linera-protocol/pull/5693.
We should find the cause of the flakiness and re-enable it.

Example: https://github.com/linera-io/linera-protocol/actions/runs/23151895906/job/67255786447

MathieuDutSik:
It seems to have been addressed by https://github.com/linera-io/linera-protocol/pull/5747

PR #5747, opened 2026-03-19
Reintroduces the `test_end_to_end_repeated_transfers`. 

MathieuDutSik:
## Motivation

This test had some problems of duplicated notifications.

A possible origin of this class of error is the function `compute` which is a CAS function and so can be used several times if a notification A is sent at the same as notification B, the function can be called several times.

This is why the manual [papaya](https://docs.rs/papaya/latest/papaya/struct.HashMap.html#method.compute) recommends using a function that is pure. The function that we used and was emitting notifications is definitely not pure, nor idempotent.

On the other hand, duplicate notifications, do not have to be fatal to the running of an application.

## Proposal

We address the purity of the function by changing the code and emitting the notification out of the `compute` code. We also make the CI test resistant to duplicate notifications.

## Test Plan

CI.
The test `test_end_to_end_repeated_transfer` is reintroduced.

## Release Plan

The change could be backported to `testnet_conway`.

## Links

None

### DIFF
--- a/linera-service/tests/linera_net_tests.rs
+++ b/linera-service/tests/linera_net_tests.rs
@@ -43,13 +43,16 @@ use linera_sdk::{
     feature = "scylladb",
     feature = "storage-service",
 ))]
-use linera_service::cli_wrappers::local_net::{Database, LocalNetConfig};
+use linera_service::cli_wrappers::{
+    local_net::{Database, LocalNetConfig},
+    Network,
+};
 #[cfg(feature = "remote-net")]
 use linera_service::cli_wrappers::{remote_net::RemoteNetTestingConfig, OnClientDrop::*};
 use linera_service::{
     cli_wrappers::{
         local_net::{get_node_port, ProcessInbox},
-        ApplicationWrapper, ClientWrapper, LineraNet, LineraNetConfig, Network, NotificationsExt,
+        ApplicationWrapper, ClientWrapper, LineraNet, LineraNetConfig, NotificationsExt,
     },
     test_name,
 };


linera-core/src/notifier.rs
```diff
@@ -63,22 +63,45 @@ where
...
+        // inside it.
+        let Some(senders) = pinned.get(chain_id).cloned() else {
+            trace!("Chain {chain_id} has no subscribers.");
+            return;
+        };
+
+        // Send notifications (side effect — must happen exactly once).
+        let mut has_dead = false;
+        for sender in &senders {
+            if sender.send(notification.clone()).is_err() {
```
afck, line 80: Actually, this _should_ fix the duplicate notifications, shouldn't it? Now that we're not calling `send` inside `compute` anymore, we don't expect any duplicates anymore? I'd not modify the test then, and see if it's still flaky.

MathieuDutSik: I agree that it should fix the duplicate notification.
I could of course revert the change in the test, but not that some traces would still be left of duplicate notification.

afck: I think as long as we don't know of a place in the code that could definitely cause duplicates, we should make the test fail if there's a duplicate.

MathieuDutSik: Fair enough. Let us try.

afck:
@MathieuDutSik: Feel free to close https://github.com/linera-io/linera-protocol/issues/5722 once this is merged.

MathieuDutSik:
> @MathieuDutSik: Feel free to close #5722 once this is merged.

Ok, done. It passed the CI.
Let us hope that it does not show up later in the future or in a different form.

PR #5755, opened 2026-03-20
Avoid extra JSON-parsing of graphql subscriptions

ma2bd:
## Motivation

Fix performance in large graphql subscriptions

## Proposal

Use the `raw_value` feature of `async_graphql` to avoid extra JSON-parsing.

## Test Plan

CI

### DIFF
--- a/linera-service/src/node_service.rs
+++ b/linera-service/src/node_service.rs
@@ -100,11 +100,9 @@ impl OutputType for RawJson {
         // When the response is serialised to JSON the raw string is emitted
         // verbatim, avoiding any parsing or tree conversion.
         //
-        // See: async-graphql-value/src/value_serde.rs
-        const RAW_VALUE_TOKEN: &str = "$serde_json::private::RawValue";
         Ok(async_graphql::Value::Object(
             std::iter::once((
-                async_graphql::Name::new(RAW_VALUE_TOKEN),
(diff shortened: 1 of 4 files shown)


linera-service/src/node_service.rs
```diff
@@ -60,6 +63,55 @@ use tracing::{debug, error, info, instrument, trace};
...
+        _ctx: &async_graphql::ContextSelectionSet<'_>,
+        _field: &Positioned<async_graphql::parser::types::Field>,
+    ) -> async_graphql::ServerResult<async_graphql::Value> {
+        // Wrap the raw JSON string with the magic token that async-graphql's
+        // ConstValue serializer recognises (with feature `raw_value`).
+        // When the response is serialised to JSON the raw string is emitted
+        // verbatim, avoiding any parsing or tree conversion.
+        //
+        // See: async-graphql-value/src/value_serde.rs
+        const RAW_VALUE_TOKEN: &str = "$serde_json::private::RawValue";
```
ndr-ds: I wonder if we could import `RAW_VALUE_TOKEN` from async graphql directly? Otherwise if they change this string for some reason, this would break 🤔

ndr-ds: Might need to add a dependency to `async-graphql-value` though, so not sure if it's worth it

ma2bd: yes let me check

PR #5760, opened 2026-03-20
Don't add --http-request-allow-list if not set

deuszx:
## Motivation

Bridge e2e task is currently failing - fix it.

## Proposal

Don't add `--http-request-allow-list` in docker compose of the bridge test if it's not set.

Should make the CI green and unblock docker image publishing.

## Test Plan

CI

## Release Plan

None.

## Links


ma2bd:
@deuszx It should be always set in practice, though, unless we want to test an obsolete version of the protocol.

deuszx:
> @deuszx It should be always set in practice, though, unless we want to test an obsolete version of the protocol.

This is just a docker compose for e2e bridge test specifically.

PR #5767, opened 2026-03-20
Expose storage cache CLI flags in the Helm chart

ndr-ds:
## Motivation

The storage cache has per-entry and per-category size limits (`max_value_entry_size`,
`max_find_keys_entry_size`, `max_find_key_values_entry_size`, `max_cache_value_size`,
`max_cache_find_keys_size`, `max_cache_find_key_values_size`) that are already supported
by `CommonStorageOptions` but weren't being passed through the Helm chart. This means
deployed validators couldn't tune these limits without code changes.

Additionally, the existing `maxCacheSize` and `maxCacheEntries` defaults were too low
for
Conway's workload.

## Proposal

- Add 6 new fields to `storageCacheConfig` in `values.yaml` corresponding to the
existing
CLI flags
- Pass them as `--storage-*` flags in both `shards.yaml` and `proxy.yaml` templates
- Bump overall cache defaults: `maxCacheSize` 10MB → 50MB, `maxCacheEntries` 1K → 50K,
`maxCacheValueSize` 10MB → 50MB

Per-entry limits and find-keys/find-key-values cache sizes stay at the Rust defaults.

## Test Plan

CI

### DIFF
--- a/kubernetes/linera-validator/templates/proxy.yaml
+++ b/kubernetes/linera-validator/templates/proxy.yaml
@@ -101,6 +101,12 @@ spec:
                 --storage-replication-factor {{ .Values.storageReplicationFactor | quote }} \
                 --storage-max-cache-size {{ .Values.storageCacheConfig.maxCacheSize | int }} \
                 --storage-max-cache-entries {{ .Values.storageCacheConfig.maxCacheEntries | int }} \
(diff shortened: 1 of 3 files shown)


ndr-ds:
This is already deployed in the validators, so merging

PR #5770, opened 2026-03-20
Make block cache and execution state cache sizes configurable via CLI

ndr-ds:
## Motivation

On testnet-conway, hot shards have 59.4% block cache hit rate (vs 99.6%+ on other
validators), generating ~2,900 extra ScyllaDB reads/s. The block cache size is hardcoded
at 5,000 entries and the execution state cache at 10,000 entries. Making them
configurable lets operators increase them for validators with concentrated hot chains.

## Proposal

- Rename `BLOCK_CACHE_SIZE` and `EXECUTION_STATE_CACHE_SIZE` to
`DEFAULT_BLOCK_CACHE_SIZE` / `DEFAULT_EXECUTION_STATE_CACHE_SIZE` and make them `pub`
for use in tests.
- Add `--block-cache-size` and `--execution-state-cache-size` CLI flags to both
`linera-server` (via `CommonStorageOptions`) and the client (via `Options`).
- Thread the values through `ServerContext` → `WorkerState::new()` for the server path,
and through `Client::new()` → `WorkerState::new_for_client()` for the client path.
- Update the Helm chart to pass both flags to shards with a default of 20,000 entries
(4x the CLI default, appropriate forway's working set).

## Test Plan

CI

### DIFF
--- a/CLI.md
+++ b/CLI.md
@@ -263,6 +263,12 @@ Client implementation and command-line tool for the Linera blockchain
 * `--blob-cache-size <BLOB_CACHE_SIZE>` — The maximal number of entries in the blob cache
 
   Default value: `1000`
+* `--block-cache-size <BLOCK_CACHE_SIZE>` — The number of entries in the block cache
+
+  Default value: `5000`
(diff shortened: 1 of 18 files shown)


ma2bd (approved):
Thanks

PR #5773, opened 2026-03-22
[testnet][web] Add connection timeout to IndexedDB

ma2bd:
## Motivation

Debug and partially solve bricked client connect in production

## Proposal

Failure to cleanup IndexedDB may create stale connections causing future attempts to never return.

Let's use a timer to fail and return an error instead

## Test Plan

CI + tested in pm-app

### DIFF
--- a/web/@linera/client/src/client.rs
+++ b/web/@linera/client/src/client.rs
@@ -49,7 +49,6 @@ impl Client {
     /// On transport or protocol error, if persistent storage is
     /// unavailable, or if `options` is incorrectly structured.
     #[wasm_bindgen(constructor)]
-    #[tracing::instrument(skip_all)]
     pub async fn new(
         mut wallet: Wallet,
         signer: Signer,


web/@linera/client/src/client.rs
```diff
@@ -56,16 +56,35 @@ impl Client {
     ) -> Result<Client> {
         let options = options.unwrap_or_default();
 
+        tracing::debug!("Client::new: acquiring wallet lock...");
```
Twey: We should probably add a parameter to log initialization to support tracing span events (e.g. from `tracing::instrument`) rather than manually constructing the instrumentation like this.

web/@linera/client/src/client.rs
```diff
@@ -56,16 +56,35 @@ impl Client {
...
         wallet.lock().await?;
-        let mut storage = storage::get_storage(&wallet.name()).await?;
+        tracing::debug!("Client::new: wallet lock acquired");
```
Twey: Nit: rather than prefixing all strings in a function, you can use `tracing::instrument` to add a span to the context.

PR #5775, opened 2026-03-23
Address the `test_wasm_end_to_end_social_event_streams` errors.

### LINKED ISSUE #5761: [testnet] flaky test_wasm_end_to_end_social_event_streams::remote_net_grpc
```
thread 'test_wasm_end_to_end_social_event_streams::remote_net_grpc' panicked at linera-service/tests/linera_net_tests.rs:2085:5:
assertion `left == right` failed
  left: Object {"receivedPosts": Object {"keys": Array [Object {"author": String("bc738559188db96764c1db07c28e5816d1c5fed35cfe227837a6c6ebd2fba6bc"), "index": Number(0)}]}}
 right: Object {"receivedPosts": Object {"keys": Array [Object {"author": String("bc738559188db96764c1db07c28e5816d1c5fed35cfe227837a6c6ebd2fba6bc"), "index": Number(1)}, Object {"author": String("bc738559188db96764c1db07c28e5816d1c5fed35cfe227837a6c6ebd2fba6bc"), "index": Number(0)}]}}
```

MathieuDutSik:
## Motivation

The test `test_wasm_end_to_end_social_event_streams` is regularly failing in CI. We propose here a solution to the problem.

Fixes #5761 

## Proposal

In the test, `app1` emits a post that should be received by `app1`. Just after that, a transfer is being done.
The call to `app1` leads to the creation of a stream entry to chain2. But the transfer also leads to another message.

I do not think we have guarantees about the order in which the calls are being processed. Therefore, we can have the next block that has the transfers and the one after that the new post. Or the reverse. So, the test has to be changed.

## Test Plan

CI is the goal.

## Release Plan

## Links

None

### DIFF
--- a/linera-service/tests/linera_net_tests.rs
+++ b/linera-service/tests/linera_net_tests.rs
@@ -2342,7 +2342,9 @@ async fn test_wasm_end_to_end_social_event_streams(config: impl LineraNetConfig)
             ]
         }
     });
-    notifications2.wait_for_block(height2.try_add_one()?).await?;
+    notifications2
+        .wait_for_block(height2.try_add_one()?)
+        .await?;
     assert_eq!(app2.query(query).await?, expected_response1);
 
     let tip_after_first_post = node_service2.chain_tip(chain1).await?;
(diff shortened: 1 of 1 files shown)


afck:
Did https://github.com/linera-io/linera-protocol/pull/5768 not fix it? Maybe we just need to port that to `main`?

MathieuDutSik:
> Did #5768 not fix it? Maybe we just need to port that to `main`?

Yes, right.
Indeed, we should have resolved the same problem on `main`. At least it is the same solution.
Thanks.

afck:
(As discussed, I'm happy to include the fix in https://github.com/linera-io/linera-protocol/pull/5737: I have to update that test there anyway.)

PR #5778, opened 2026-03-23
Access to keys with pagination

MathieuDutSik:
## Motivation

Accessing key-values in `MapView` is an expensive operation that blocks the system.
It would be better to be able to do pagination.

## Proposal

When doing a `find_{keys,key_values}_by_prefix` in ScyllaDB, RocksDB and others, we are accessing the keys with a start and end with the start and end specifically defined. So, there is a very clear way to generalize the code. We also have a very clear way to limit the size of the request. In ScyllaDB, this is done via the " LIMIT X" appended, and for RocksDB, we can stop in the middle.

Now, if we ask for 200 keys and we get 190 then we know that we do not need to continue. If we get exactly 200, then the number of keys may be 200 or 201. But we do not that. So, we report `is_finished=false` in that case. The number of keys being returned may in the end be much less than 200 because the value_splitting forces to split some keys in several smaller keys.

This opens the possibility of reading the key-values of a `MapView` with pagination. This has not been done in this PR.

## Test Plan

CI.
The tests have been expanded.

## Release Plan

This can be backported to `testnet_conway`.

## Links

None.

### DIFF
--- a/linera-execution/src/execution_state_actor.rs
+++ b/linera-execution/src/execution_state_actor.rs
@@ -24,7 +24,7 @@ use linera_base::{
     ownership::ChainOwnership,
     time::Instant,
 };
-use linera_views::{batch::Batch, context::Context, views::View};
+use linera_views::{batch::Batch, context::Context, store::KeyInterval, views::View};
 use oneshot::Sender;
 use reqwest::{header::HeaderMap, Client, Url};
 use tracing::{info_span, instrument, Instrument as _};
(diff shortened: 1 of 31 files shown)


deuszx:
if you get to work on pagination for `try_load_all_entries` – please reference the corresponding issue https://github.com/linera-io/linera-protocol/issues/2742

ma2bd:
This is cool.

PR #5779, opened 2026-03-23
Fix General dashboard error rate queries

ndr-ds:
## Motivation

The General monitoring dashboard has two issues with error rate panels:
- **Proxy Error rate** includes `non_grpc` method noise and shows duplicate lines per
method (no aggregation across instances)
- **Server Error rate** groups by both `method_name` and `instance` (shard), making it
too noisy to be useful

## Proposal

- **Proxy Error rate**: Add `method_name!="non_grpc"` filter and `sum(...) by
(method_name)` aggregation so there's exactly one line per gRPC method
- **Server Error rate**: Drop `instance` from the `by` clause so error rates are
aggregated per method across all shards; update legend from `{{instance}} -
{{method_name}}` to `{{method_name}}`
- Apply equivalent fixes to both the Kubernetes and Docker dashboard JSON files

No changes to Server Latency by Request Type — it was already correct (one series per
method per percentile).

## Test Plan

Verified live on Grafana central monitoring dashboard before committing JSON changes.


ndr-ds:
This is already on central monitoring, so merging

PR #5780, opened 2026-03-23
Add source label to DbStorage metrics for cache vs DB visibility

ndr-ds:
## Motivation

The DbStorage metrics are inconsistent after the application-level `ValueCache`
additions (#5718, #5742):

- **Single-item reads** (`read_blob`, `read_confirmed_block`, `read_certificate`,
`read_event`): metric only fires on cache miss — cache hits are invisible.
- **Batch reads** (`read_confirmed_blocks`, `read_certificates_raw`): metric fires for
all items regardless of cache/DB source.
- **Double counting**: `read_blobs` calls `read_blob()` per item (each increments the
counter), then adds another `inc_by(total)` at batch level. Same bug in
`read_blob_states`.
- **Missing cache check**: `contains_event` goes straight to DB even though
`event_cache` exists.

This makes it impossible to distinguish cache hits from DB reads in Grafana, and the
double-counting inflates batch read metrics.

## Proposal

Add a `source` label with values `"cache"` / `"db"` to all 17 `IntCounterVec` metrics in
DbStorage:

- **Single-item reads with caches**: increment `source="cache"` on cache hit,
`source="db"` on DB path.
- **Batch reads** (`read_confirmed_blocks`, `read_certificates_raw`): split the flat
`inc_by(total)` into separate cache-hit and DB-miss counts.
- **Fix double counting**: remove redundant batch-level `inc_by` from `read_blobs` and
`read_blob_states` — each per-item call already handles its own metric.
- **Add cache check to `contains_event`**: early return on `event_cache.contains()`,
same pattern as `contains_blob`.
- **DB-only methods** (writes, `read_network_description`, `contains_blob_state`, etc.):
use `source="db"`.
- **Remove `READ_BLOB_STATES_COUNTER`**: became dead code after the double-counting fix.

Grafana can now show: `total = sum by (source)`, `cache_rate = cache / (cache + db)`,
`db_pressure = {source="db"}`.

## Test Plan

CI

### DIFF
--- a/linera-storage/src/db_storage.rs
+++ b/linera-storage/src/db_storage.rs
@@ -45,16 +45,24 @@ pub mod metrics {
     use std::sync::LazyLock;
 
     use linera_base::prometheus_util::{
-        exponential_bucket_latencies, register_histogram_vec, register_int_counter_vec,
+        exponential_bucket_latencies, register_histogram_vec, register_int_counter,
+        register_int_counter_vec,
     };
-    use prometheus::{HistogramVec, IntCounterVec};
+    use prometheus::{HistogramVec, IntCounter, IntCounterVec};
+
+    /// Label name for distinguishing cache hits vs DB reads.
+    pub(super) const SOURCE_LABEL: &str = "source";
+    /// Label value for items served from the in-memory cache.
+    pub(super) const CACHE: &str = "cache";
+    /// Label value for items served from the database.
+    pub(super) const DB: &str = "db";
 
     /// The metric counting how often a blob is tested for existence from storage
     pub(super) static CONTAINS_BLOB_COUNTER: LazyLock<IntCounterVec> = LazyLock::new(|| {
         register_int_counter_vec(
             "contains_blob",
             "The metric counting how often a blob is tested for existence from storage",
-            &[],
+            &[SOURCE_LABEL],
         )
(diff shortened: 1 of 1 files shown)


deuszx:
> cache_rate = cache / (cache + db)

If you're meaning to count the cache hit ratio then it would be an inverse of this?

linera-storage/src/db_storage.rs
```diff
@@ -310,7 +309,7 @@ impl MultiPartitionBatch {
     ) -> Result<(), ViewError> {
         #[cfg(with_metrics)]
         metrics::WRITE_CERTIFICATE_COUNTER
-            .with_label_values(&[])
+            .with_label_values(&[metrics::DB])
```
deuszx: Do we ever have `metrics::WRITE_CERTIFICATE_COUNTER` with label `metrics::CACHE`?

ndr-ds:
I don't think so 🤔

linera-storage/src/db_storage.rs
ndr-ds: True, I guess write counters don't need the label at all. Claude added it for consistency I think. Will remove

deuszx: Exactly

PR #5784, opened 2026-03-23
Add ScyllaDB I/O panels and cluster filter to ScyllaDB dashboard

ndr-ds:
## Motivation

The ScyllaDB Grafana dashboard had no visibility into native disk I/O metrics (IOPS,
queue depth, scheduler delay). During recent performance investigations, we needed to
check whether ScyllaDB was I/O bottlenecked and had to query Prometheus manually each
time.

Also, the dashboard lacked a Cluster variable for filtering native ScyllaDB metrics by
validator cluster, and shared crosshair was not enabled (unlike all other Linera
dashboards).

## Proposal

- Add 3 new panels in a "ScyllaDB I/O" row section:
- **IOPS** — read + write ops/sec from `scylla_io_queue_total_{read,write}_ops`
- **I/O Queue Length** — `scylla_io_queue_queue_length` (I/O scheduler backlog)
- **I/O Delay** — `scylla_io_queue_delay` (scheduler dispatch delay in µs)
- Add `$cluster` template variable (populated from `scylla_io_queue_total_read_ops`
cluster label) to filter I/O panels by validator cluster
- Enable shared crosshair (`graphTooltip: 1`) so hovering on one panel he time
cursor on all panels

## Test Plan

Panels verified on the live Grafana dashboard against all 4 validator clusters before
codifying the JSON.

### DIFF
--- a/kubernetes/linera-validator/grafana-dashboards/linera/storage/scylladb.json
+++ b/kubernetes/linera-validator/grafana-dashboards/linera/storage/scylladb.json
@@ -17,7 +17,7 @@
   },
   "editable": true,
   "fiscalYearStartMonth": 0,
-  "graphTooltip": 0,
+  "graphTooltip": 1,
   "links": [],
   "panels": [
     {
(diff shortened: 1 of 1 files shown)


ndr-ds:
Already on central monitoring, so merging

PR #5786, opened 2026-03-24
Add Prometheus metrics to all caches and switch to allocative-based byte weights

ndr-ds:
## Motivation

Our caches had two problems:

1. **No observability for most caches.** `ValueCache` only had type-name-based hit/miss
counters (`value_cache_hit{key_type, value_type}`). `UniqueValueCache` had zero metrics.
The query response cache in `node_service.rs` had its own bespoke metrics module. We
had no visibility into cache byte usage, eviction rates, entry counts, or invalidations
for any cache.

2. **Incorrect or expensive weight functions.** `ValueCache` used item-count eviction
(no weight awareness at all — just `Cache::new(size)`). When we eventually need
weight-based eviction, the existing approach required per-cache weight function closures
— some using `bcs::serialized_size` (full serde traversal cost) and others using
`std::mem::size_of` (wrong for heap-heavy types).

### Unified Prometheus metrics for all caches

Added a `metrics` module to `linera-cache` with six labeled metrics covering all cache
instances:

- `cache_hit{name, t` / `cache_miss{name, type}` — per-cache hit/miss counters
- `cache_entries{name, type}` / `cache_bytes{name, type}` — current entry count and byte
weight gauges
- `cache_eviction{name, type}` — eviction counter (via quick_cache `Lifecycle` hook for
`ValueCache`, manual tracking for `UniqueValueCache`)
- `cache_invalidation{name, type}` — full-cache clear counter

The `type` label is `"value"` or `"unique"`, and `name` is a human-readable label like
`"block"`, `"blob"`, `"execution_state"`, etc. This replaces the old type-name-based
labels (e.g., `key_type="linera_base::crypto::CryptoHash"`) and the bespoke
`query_response_cache_*` metrics in `node_service.rs`.

A global name registry (behind `#[cfg(with_metrics)]`) panics at startup if two caches
are created with the same name, preventing metric label collisions.

### `allocative`-based byte weights

Every cached type already derives `Allocative`. The new `AllocativeWeighter` uses
`allocative::size_of_unique()` which memem::size_of::<T>()` + all uniquely-owned
heap allocations. This replaces the need for per-cache weight function closures.

- `ValueCache::new(name, weight_capacity)` — down from `Cache::new(size)` (item-count)
to weight-based eviction with 2 parameters
- `UniqueValueCache::new(name, capacity)` — keeps item-count LRU eviction but now tracks
byte weight via allocative for the `cache_bytes` gauge

### Byte-based CLI options

All cache size CLI flags are now byte limits (`u64`) instead of entry counts (`usize`):

- `--blob-cache-size`, `--confirmed-block-cache-size`, `--lite-certificate-cache-size`,
`--certificate-raw-cache-size`, `--event-cache-size`, `--block-cache-size` — all default
to 50 MiB (conservative; production deployments tune via Helm)
- `--execution-state-cache-size` — unchanged, stays as item count (it uses
`UniqueValueCache` with LRU eviction)

### Query response cache consolidation

`node_service.rs` had its own `query_cache_metrics` module with 4 bespokeeus
metrics and an `LruCache`-based implementation. Replaced it with `ValueCache` from
`linera-cache`, getting all six standard cache metrics for free. Removed the `lru`
dependency from `linera-service`.

## Test Plan

CI

### DIFF
--- a/linera-bridge/src/main.rs
+++ b/linera-bridge/src/main.rs
@@ -94,25 +94,25 @@ struct ServeOptions {
     #[arg(long, default_value = "3001")]
     port: u16,
 
-    /// The maximal number of entries in the blob cache.
-    #[arg(long, default_value = "1000")]
-    blob_cache_size: usize,
+    /// Maximum bytes in the blob cache.
+    #[arg(long, default_value_t = 52_428_800)]
+    blob_cache_size: u64,
 
-    /// The maximal number of entries in the confirmed block cache.
-    #[arg(long, default_value = "1000")]
-    confirmed_block_cache_size: usize,
+    /// Maximum bytes in the confirmed block cache.
+    #[arg(long, default_value_t = 52_428_800)]
+    confirmed_block_cache_size: u64,
 
-    /// The maximal number of entries in the lite certificate cache.
-    #[arg(long, default_value = "1000")]
-    lite_certificate_cache_size: usize,
+    /// Maximum bytes in the lite certificate cache.
+    #[arg(long, default_value_t = 52_428_800)]
+    lite_certificate_cache_size: u64,
 
-    /// The maximal number of entries in the raw certificate cache.
-    #[arg(long, default_value = "1000")]
-    certificate_raw_cache_size: usize,
+    /// Maximum bytes in the raw certificate cache.
+    #[arg(long, default_value_t = 52_428_800)]
+    certificate_raw_cache_size: u64,
 
-    /// The maximal number of entries in the event cache.
-    #[arg(long, default_value = "1000")]
-    event_cache_size: usize,
+    /// Maximum bytes in the event cache.
+    #[arg(long, default_value_t = 52_428_800)]
+    event_cache_size: u64,
 }
 
 fn main() -> Result<()> {
--- a/linera-cache/src/unique_value_cache.rs
+++ b/linera-cache/src/unique_value_cache.rs
@@ -3,30 +3,55 @@
 
 //! A mutex-guarded LRU cache for move-in/move-out value patterns where `V` is not `Clone`.
 
-use std::{hash::Hash, num::NonZeroUsize, sync::Mutex};
+use std::{
+    hash::Hash,
+    num::NonZeroUsize,
+    sync::{
+        atomic::{AtomicI64, Ordering},
+        Mutex,
+    },
+};
 
+use allocative::Allocative;
 use lru::LruCache;
 
 /// A bounded cache for values that are inserted and then taken out (moved), not cloned.
 ///
 /// Uses `Mutex<LruCache>` internally. Suitable for caching values that don't implement
 /// `Clone`, where the access pattern is insert → remove rather than insert → get.
+///
+/// When built with the `metrics` feature, reports hit, miss, entry-count, byte-weight,
+/// and invalidation counters to Prometheus.
 pub struct UniqueValueCache<K, V>
 where
     K: Hash + Eq,
 {
     cache: Mutex<LruCache<K, V>>,
+    total_bytes: AtomicI64,
+    #[cfg(with_metrics)]
+    name: &'static str,
 }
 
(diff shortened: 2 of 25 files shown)


deuszx:
In the future, can you make functionally indepentent commits? This PR, at the very least, should have 2 commits (metrics and weight-based caching).

linera-bridge/src/main.rs
```diff
@@ -94,25 +94,25 @@ struct ServeOptions {
     #[arg(long, default_value = "3001")]
     port: u16,
 
-    /// The maximal number of entries in the blob cache.
-    #[arg(long, default_value = "1000")]
-    blob_cache_size: usize,
+    /// Maximum bytes in the blob cache.
+    #[arg(long, default_value_t = 52_428_800)]
+    blob_cache_size: u64,
```
deuszx, line 99: If those are CLI args I always include the metric as the suffic: `*_bytes`. See how this one didn't change even though we switched from number of elements to sizes?

linera-cache/src/unique_value_cache.rs
```diff
@@ -3,30 +3,55 @@
...
-    /// Creates a new `UniqueValueCache` with the given capacity.
-    pub fn new(size: usize) -> Self {
-        let size = NonZeroUsize::try_from(size).expect("Cache size must be greater than zero");
+    /// Creates a new `UniqueValueCache`.
+    ///
+    /// * `name` – human-readable label for Prometheus metrics (e.g. `"execution_state"`).
+    /// * `capacity` – maximum number of entries (LRU eviction by item count).
+    ///
+    /// Byte weights are computed automatically via [`allocative::size_of_unique`].
+    #[cfg_attr(not(with_metrics), allow(unused_variables))]
```
deuszx, line 46: We could have two `new` constructors – one for _with_ metrics and one w/o them

linera-cache/src/unique_value_cache.rs
```diff
@@ -39,14 +64,68 @@ where
             cache.promote(key);
             false
         } else {
-            cache.push(*key, value);
+            let added_weight = allocative::size_of_unique(&value) as i64;
```
deuszx, line 67: I'm afraid that the act of measuring thesize of `value` and `evicted_value` itself is affecting the metrics. Maybe we can come up with a scheme that uses some statically-known sizes? A good approximation is "good enough". Or we can use the "maximum-allowed" size for things like blobs and certs.

linera-core/src/worker.rs
```diff
@@ -56,7 +56,7 @@ use crate::{
     CHAIN_INFO_MAX_RECEIVED_LOG_ENTRIES,
 };
 
-pub const DEFAULT_BLOCK_CACHE_SIZE: usize = 5_000;
+pub const DEFAULT_BLOCK_CACHE_SIZE: u64 = 52_428_800;
```
deuszx, line 59: DEFAULT_BLOCK_CACHE_SIZE_BYTES

deuszx, line 59: Where is this number coming from? 52MB?

linera-cache/src/unique_value_cache.rs
```diff
@@ -39,14 +64,68 @@ where
...
+            let evicted = cache.push(*key, value);
+            let mut delta = added_weight;
+            if let Some((_evicted_key, evicted_value)) = evicted {
+                delta -= allocative::size_of_unique(&evicted_value) as i64;
+                #[cfg(with_metrics)]
+                crate::metrics::CACHE_EVICTION
+                    .with_label_values(&[self.name, "unique"])
+                    .inc();
+            }
+            self.total_bytes.fetch_add(delta, Ordering::Relaxed);
```
deuszx, line 77: Why do we have to track the cache's size manually? Isn't the point of using have it done this work?

linera-bridge/src/main.rs
ma2bd: This is a little scary. Do we have the "cache size" ambiguity in other places?

PR #5789, opened 2026-03-24
Update papaya

afck:
## Motivation

In https://github.com/linera-io/linera-protocol/pull/5687 I ran into https://github.com/ibraheemdev/papaya/issues/63. We're not using `retain` directly currently, but I'm not sure if it's used internally by `papaya` anywhere.

## Proposal

Update `papaya` to the latest version, which doesn't have that bug.

## Test Plan

CI

## Release Plan

## Links


afck:
Merging: The test failure is not related to this, and already happening on `testnet_conway`.

PR #5791, opened 2026-03-24
Backport #5601: Make jemalloc memory profiling opt-in via --enable-memory-profiling flag

ndr-ds:
## Motivation

Backport of #5601 to `testnet_conway`. Adds jemalloc as the global allocator and exposes
an `--enable-memory-profiling` CLI flag that activates jemalloc heap profiling and a
`/debug/pprof` endpoint on the metrics server.

Beyond profiling, switching to jemalloc fixes severe memory bloat caused by glibc malloc
fragmentation under heavy async RocksDB workloads. Workers on the default glibc
allocator grow to 50+ GB over a few days, while workers on jemalloc remain stable.

## Proposal

Cherry-pick of `14f2a9f5848` onto `testnet_conway`.

## Test Plan

CI

### DIFF
--- a/linera-metrics/src/memory_profiler.rs
+++ b/linera-metrics/src/memory_profiler.rs
@@ -22,12 +22,33 @@ pub enum MemoryProfilerError {
     ProfCtlNotAvailable,
     #[error("another profiler is already running")]
     AnotherProfilerAlreadyRunning,
+    #[error("failed to activate jemalloc profiling: {0}")]
+    ActivationFailed(String),
 }
 
 /// Memory profiler using safe jemalloc_pprof wrapper (pull model only)
 pub struct MemoryProfiler;
 
 impl MemoryProfiler {
+    /// Activates jemalloc profiling at runtime.
+    ///
+    /// This enables sampling (prof_active) which is off by default to avoid
+    /// the libgcc DWARF unwinder livelock (jemalloc/jemalloc#2282).
+    pub async fn activate() -> Result<(), MemoryProfilerError> {
+        if let Some(prof_ctl) = PROF_CTL.as_ref() {
+            let mut prof_ctl = prof_ctl.lock().await;
+
+            prof_ctl
+                .activate()
+                .map_err(|e| MemoryProfilerError::ActivationFailed(e.to_string()))?;
+
+            info!("jemalloc memory profiling activated");
+            Ok(())
+        } else {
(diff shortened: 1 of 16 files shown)


linera-service/src/cli/main.rs
```diff
@@ -8,21 +8,11 @@
...
-// jemalloc configuration for memory profiling with jemalloc_pprof
-// prof:true,prof_active:true - Enable profiling from start
-// lg_prof_sample:19 - Sample every 512KB for good detail/overhead balance
-
-// Linux/other platforms: use unprefixed malloc (with unprefixed_malloc_on_supported_platforms)
-#[cfg(all(feature = "memory-profiling", not(target_os = "macos")))]
-#[allow(non_upper_case_globals)]
+/// Configure jemalloc profiling infrastructure at startup with sampling disabled.
+/// Profiling is activated at runtime only when `--enable-memory-profiling` is passed.
+#[cfg(feature = "jemalloc")]
```
ma2bd, line 19: Should `jemalloc` be a default feature? Or only for certain targets?

linera-metrics/src/memory_profiler.rs
```diff
@@ -22,12 +22,33 @@ pub enum MemoryProfilerError {
...
 }
 
 /// Memory profiler using safe jemalloc_pprof wrapper (pull model only)
 pub struct MemoryProfiler;
 
 impl MemoryProfiler {
+    /// Activates jemalloc profiling at runtime.
+    ///
+    /// This enables sampling (prof_active) which is off by default to avoid
+    /// the libgcc DWARF unwinder livelock (jemalloc/jemalloc#2282).
```
ma2bd, line 36: Can we dynamically verify which unwind library is present and refuse to enalbe memory profiling if it's the buggy one?

ndr-ds: After https://github.com/linera-io/jemallocator/pull/2, if libunwind is not around it won't default to the buggy one anymore, it'll fail to link, because it'll try to link against it and won't find it. So we should be good now I think

linera-service/src/cli/main.rs
ndr-ds: We could enable it by defailt

ndr-ds: Ah if we enable it by default then CI needs to have libunwind and all that... so maybe not for now (the dockerfiles all have it)

PR #5795, opened 2026-03-25
Add per-method breakdown to proxy and server request metrics

ndr-ds:
## Motivation

The proxy and server middleware track request count and latency, but only with a
`traffic_type` label — there's no method-level breakdown. This makes it impossible to
tell which gRPC methods are driving request volume or latency from the primary
dashboards.

## Proposal

Add a `method_name` label to the four core middleware metrics:
- `linera_proxy_request_latency`
- `linera_proxy_request_count`
- `linera_server_request_latency`
- `linera_server_request_count`

The method name is extracted from the gRPC URI path (`/{package}.{Service}/{Method}` →
`Method`). Non-gRPC requests (health checks, bot probes) are labeled `non_grpc` to
prevent unbounded cardinality.

The extraction logic already existed inline in the proxy for the success/error counters
— it's now pulled into a shared `extract_grpc_method_name()` helper in
`linera_rpc::grpc` alongside the shared label constants.

Since `server_request_latency` now has `method_name`, the separate
`serverst_latency_per_request_type` histogram is redundant and removed, along with
the per-handler `Instant::now()` / latency recording that fed it — latency is already
captured in the middleware layer.

Recording rules are updated to preserve `method_name` in the pre-aggregated series.

Dashboard changes will follow in a separate PR after deployment.

## Test Plan

CI

### DIFF
--- a/linera-rpc/src/grpc/mod.rs
+++ b/linera-rpc/src/grpc/mod.rs
@@ -41,3 +41,75 @@ pub const GRPC_MAX_MESSAGE_SIZE: usize = 16 * MEBIBYTE;
 /// Limit of gRPC message size up to which we will try to populate with data when estimating.
 /// We leave 30% of buffer for the rest of the message and potential underestimation.
 pub const GRPC_CHUNKED_MESSAGE_FILL_LIMIT: usize = GRPC_MAX_MESSAGE_SIZE * 7 / 10;
+
+/// Prometheus label for the gRPC method name.
+pub const METHOD_NAME_LABEL: &str = "method_name";
+
+/// Prometheus label for distinguishing organic vs synthetic (benchmark) traffic.
+pub const TRAFFIC_TYPE_LABEL: &str = "traffic_type";
+
+/// Extracts the gRPC method name from a request URI path.
+///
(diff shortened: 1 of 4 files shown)


deuszx:
I'd swear I had already reviewed PR like this 👀

linera-rpc/src/grpc/mod.rs
```diff
@@ -41,3 +41,23 @@ pub const GRPC_MAX_MESSAGE_SIZE: usize = 16 * MEBIBYTE;
...
+
+/// Prometheus label for distinguishing organic vs synthetic (benchmark) traffic.
+pub const TRAFFIC_TYPE_LABEL: &str = "traffic_type";
+
+/// Extracts the gRPC method name from a request URI path.
+///
+/// gRPC paths have the form `/{package}.{Service}/{Method}` — the first segment
+/// always contains a dot. Non-gRPC requests (health checks, bot probes, etc.) return
+/// `"non_grpc"` to prevent unbounded label cardinality.
+pub fn extract_grpc_method_name(path: &str) -> &str {
```
deuszx, line 56: Some unit tests for this would be welcome (and easy to write)

PR #5796, opened 2026-03-25
Remove notifiers from the list when closed.

MathieuDutSik:
## Motivation

We are exposed to OOM problems.
A small place where we can improve the situation is with respect to dead notifiers: We need to remove them.

## Proposal

We do the following:
* When registering a sender, drop the closed ones.
* The background sweep is now done even when no TTL is set.
* The sweep process stops when there is no notifiers to take care of.

## Test Plan

CI.

## Release Plan

## Links

None

### DIFF
--- a/linera-core/src/chain_worker/delivery_notifier.rs
+++ b/linera-core/src/chain_worker/delivery_notifier.rs
@@ -28,6 +28,7 @@ pub struct DeliveryNotifier {
 
 impl DeliveryNotifier {
     /// Registers a delivery `notifier` for a desired [`BlockHeight`].
+    /// Also prunes any closed oneshot senders to prevent unbounded accumulation.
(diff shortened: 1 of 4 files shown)


afck:
Maybe I'm confused about the notifiers, but: Is that really doing anything? Entries should get removed automatically anyway as soon as the cross-chain messages from the corresponding height have been delivered, shouldn't they? If delivery takes so long that while it's pending we'd run multiple sweeps that seems like a problem in itself.

Or are they accumulating due to some interaction with the set of tracked chains? Then we should probably fix that underlying issue and make sure notifiers are removed as soon as no tracked chains are left that are waiting for message delivery?

MathieuDutSik:
That PR does not seem to help.

PR #5798, opened 2026-03-25
Simplify the `block_cache`.

MathieuDutSik:
## Motivation

The `block_cache: Arc<ValueCache<CryptoHash, Hashed<Block>>>` has a duplication problem.
The `CryptoHash` is present in the key and also in the value. This is a little wasteful.

## Proposal

Replace it with `block_cache: Arc<ValueCache<CryptoHash, Block>>` and introduce a `get_hashed` function.
Investigation of the call sites of `get_hashed` revealed that we always need the hash being produced. So, the
function `get` is never used, but it was kept since it could be useful along the line.

## Test Plan

CI.

## Release Plan

## Links

None.

### DIFF
--- a/linera-core/src/value_cache.rs
+++ b/linera-core/src/value_cache.rs
@@ -124,7 +124,7 @@ impl<T: Clone> ValueCache<CryptoHash, T> {
     ///
     /// The hash used as the cache key is combined with the stored value to
     /// reconstruct the [`Hashed<T>`] without redundant storage.
-    pub fn get_hashed(&self, hash: &CryptoHash) -> Option<Hashed<T>> {
+    pub fn get(&self, hash: &CryptoHash) -> Option<Hashed<T>> {
         let value = Self::track_cache_usage(self.cache.lock().unwrap().get(hash).cloned())?;
         Some(Hashed::unchecked_new(value, *hash))
     }
--- a/linera-core/src/chain_worker/state.rs
+++ b/linera-core/src/chain_worker/state.rs
@@ -415,7 +415,7 @@ where
         let mut height_to_blocks: HashMap<BlockHeight, Hashed<Block>> = HashMap::new();
 
         for hash in hashes {
-            if let Some(hashed_block) = self.block_values.get_hashed(&hash) {
+            if let Some(hashed_block) = self.block_values.get(&hash) {
                 height_to_blocks.insert(hashed_block.inner().header.height, hashed_block);
             } else {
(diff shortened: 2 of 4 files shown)


linera-core/src/value_cache.rs
```diff
@@ -80,6 +80,7 @@ where
     }
 
     /// Returns a `V` from the cache, if present.
+    #[allow(dead_code)]
```
afck: Let's remove it then?

MathieuDutSik: I was thinking of keeping it for possible future use.
But yes, of course, we can remove it.

afck: I'd prefer minimizing the number of `allow(dead_code)` we have. And it should be easy to bring it back if we need it.

MathieuDutSik: Ok, removed.

linera-service/template/Cargo.toml.template
```diff
@@ -16,6 +16,7 @@ darling = {{ version = ">=0.20, <0.23" }}
 serde_with = {{ version = ">=3, <3.18" }}
 time = {{ version = ">=0.3, <0.3.47" }}
 time-core = {{ version = ">=0.1, <0.1.8" }}
+unicode-segmentation = {{ version = ">=1, <1.13" }}
```
afck, line 19: (Is this another dependency that broke the `project new` test? 😬)

linera-core/src/value_cache.rs
```diff
@@ -119,13 +114,22 @@ impl<T: Clone> ValueCache<CryptoHash, Hashed<T>> {
...
             true
         }
     }
 
-    /// Inserts multiple [`HashedCertificateValue`]s into the cache. If they're not
+    /// Retrieves a value from the cache and reconstructs the [`Hashed`] wrapper.
+    ///
+    /// The hash used as the cache key is combined with the stored value to
+    /// reconstruct the [`Hashed<T>`] without redundant storage.
+    pub fn get_hashed(&self, hash: &CryptoHash) -> Option<Hashed<T>> {
```
afck: Should we just call this `get` now?

linera-service/template/Cargo.toml.template
MathieuDutSik: Yes.

linera-core/src/value_cache.rs
MathieuDutSik: Yes, we can.

PR #5804, opened 2026-03-25
Split handling events into multiple operations; drop operations exceeding block limits

bart-linera:
## Motivation

If there are many event streams that have been updated when a block is being proposed, it is possible that handling all the updates would exceed block limits. Currently that would just make the block fail execution, and the event streams would not be able to progress anymore.

## Proposal

Split updating event streams into many operations. If an `UpdateStream` operation fails due to block limits, drop it from the block and continue executing the rest - similar to how messages are handled.

## Test Plan

CI will catch regressions.

## Release Plan

## Links

- Related to #5379, but there are more improvements to be done before closing the issue. 


linera-chain/src/chain.rs
```diff
@@ -744,66 +744,92 @@ where
...
+                (Ok(()), _, _) => {
+                    i += 1;
+                }
+                (
+                    Err(ChainError::ExecutionError(error, _context)),
+                    Transaction::ReceiveMessages(incoming_bundle),
+                    Some(_),
+                ) if !error.is_transient_error() && error.is_limit_error() && i > 0 => {
+                    failure_count += 1;
+                    // If we've exceeded max failures, discard all remaining message bundles.
```
afck, line 806: Should we then also discard all remaining `UpdateStream`s? It's the same block and we must be close to the block limits.

linera-chain/src/chain.rs
```diff
@@ -744,66 +744,92 @@ where
...
+                        info!(
+                            %error,
+                            index = i,
+                            origin = %incoming_bundle.origin,
+                            "Message bundle exceeded block limits and will be discarded for \
+                            retry in a later block"
+                        );
+                        Some(incoming_bundle.origin)
+                    };
+                    Self::discard_remaining_bundles(block, i, maybe_sender);
```
afck, line 827: Don't we need to restore the checkpoint here?

linera-execution/src/system.rs
```diff
@@ -133,14 +133,36 @@ impl<C: Context, C2: Context> ReplaceContext<C2> for SystemExecutionStateView<C>
...
+/// The applications subscribing to a particular stream, and their per-application event indices.
+#[derive(Debug, Clone, Serialize, Deserialize, Allocative)]
 pub struct EventSubscriptions {
-    /// The next event index, i.e. the total number of events in this stream that have already
-    /// been processed by this chain.
-    pub next_index: u32,
-    /// The applications that are subscribed to this stream.
-    pub applications: BTreeSet<ApplicationId>,
+    /// Cached minimum of all per-application `next_index` values. Used for short-circuit
+    /// filtering: if the next available event index is <= this value, no application needs
```
afck, line 143: ~`<` or `<=`?~

~I'd think if it's the minimum of the values below, it's the minimum next index that we _are_ interested in, so "no update needed" means _strictly_ less?~

Ah, I guess by "next available" you mean "lowest unavailable"? 😵 I.e. whatever the "next expected" index is on the publisher chain?

linera-chain/src/chain.rs
bart-linera: Yes! :scream: Good catch!

linera-chain/src/chain.rs
bart-linera: Good idea, I'll make it discard both bundles and updates in both "failures exceeded" scenarios.

linera-chain/src/chain.rs
afck: I guess this shows we don't have enough tests for this. 😬 
Or is there a test that just needs to be extended, to make sure the chain state is as expected afterwards?

linera-core/src/unit_tests/wasm_client_tests.rs
```diff
@@ -1005,13 +1005,19 @@ where
...
-    // The inbox processing should produce UpdateStreams operations for the events.
-    let has_update_streams = certs.iter().any(|cert| {
-        cert.block().body.operations().any(|op| {
-            matches!(op, Operation::System(op) if matches!(**op, SystemOperation::UpdateStreams(_)))
+    // The inbox processing should produce UpdateStream operations for the events.
+    let count_update_streams: usize = certs
+        .iter()
+        .map(|cert| {
+            cert.block().body.operations().filter(|op| {
+            matches!(op, Operation::System(op) if matches!(**op, SystemOperation::UpdateStream { .. }))
```
afck: Maybe we should have `is_update_stream` on `Operation`, too (and then use that to implement the one for `Transaction`)?

linera-chain/src/chain.rs
bart-linera: Yes, we definitely don't :grimacing:  I'm currently trying to figure out the best approach to testing this.

linera-sdk/tests/fixtures/event-emitter/src/lib.rs
afck, line 1: Should these examples live in `linera-sdk/tests/fixtures/` instead, since they are meant for tests only, rather than as documentation?

linera-service/src/cli_wrappers/wallet.rs
```diff
@@ -1274,6 +1274,17 @@ impl ClientWrapper {
         Ok(())
     }
 
+    pub async fn set_maximum_block_size(&self, size: u64) -> Result<()> {
```
afck: Maybe the wallet wrapper should match the CLI, and mirror the `resource-control-policy` command in general?

linera-sdk/tests/fixtures/event-emitter/src/lib.rs
bart-linera: Good point!

linera-service/src/cli_wrappers/wallet.rs
bart-linera: With the number of possible parameters, this will make one horrible function... :grimacing: But it may be more useful to have such a general function rather than many specific ones.

afck: Could it just take a single `MessagePolicy` argument? But internally, it would have to take it apart and use the correct CLI parameters, yes. :grimacing: Not a blocker.

afck (approved):

afck (approved):

PR #5805, opened 2026-03-25
create option --ignored-bundle-origins

ma2bd:
## Motivation

Create a basic option to ignore certain senders when building blocks

## Proposal

Create option `--ignored-bundle-from` doing the opposite of `--prioritize-bundles-from` .

## Test Plan

CI


afck:
The message policy already has a whitelist of sender chains; maybe it should also have a blacklist, and then we should probably use the whitelist or blacklist like in this PR, to already filter in the worker rather than loading all bundles and then dropping them again.

ma2bd:
landing it as discussed with @afck

PR #5809, opened 2026-03-26
[testnet] Fix events-only chain handling

afck:
## Motivation

The `test_wasm_end_to_end_social_event_streams` fails if it's extended with a restarted client.

It turns out events-only chains sometimes get added to the wallet, and after restarting are not considered "events-only" anymore.

## Proposal

Don't add them to the wallet. Also, don't use `local_committee` before the listening mode is set.

## Test Plan

The extended test passes now.

## Release Plan

- Release a new SDK.
- Possibly remove events chains from some wallets in production.

## Links

### DIFF
--- a/linera-client/src/chain_listener.rs
+++ b/linera-client/src/chain_listener.rs
@@ -377,7 +377,9 @@ impl<C: ClientContext + 'static> ChainListener<C> {
                 self.update_validators(&notification).await?;
             }
             Reason::NewBlock { hash, .. } => {
-                self.update_wallet(notification.chain_id).await?;
+                if !matches!(listening_mode, ListeningMode::EventsOnly(_)) {
+                    self.update_wallet(notification.chain_id).await?;
+                }
                 if listening_mode.is_full() {
                     self.add_new_chains(*hash).await?;
                     let publishers = self
--- a/linera-core/src/client/chain_client/mod.rs
+++ b/linera-core/src/client/chain_client/mod.rs
@@ -2951,7 +2951,7 @@ impl<Env: Environment> ChainClient<Env> {
 
         let events_only = self
             .listening_mode()
-            .is_some_and(|m| matches!(m, ListeningMode::EventsOnly(_)));
+            .is_none_or(|m| matches!(m, ListeningMode::EventsOnly(_)));
         let (nodes, local_node) = {
             // For EventsOnly chains, use the admin chain's committee: the chain's own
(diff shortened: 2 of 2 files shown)


linera-client/src/chain_listener.rs
```diff
@@ -377,7 +377,9 @@ impl<C: ClientContext + 'static> ChainListener<C> {
                 self.update_validators(&notification).await?;
             }
             Reason::NewBlock { hash, .. } => {
-                self.update_wallet(notification.chain_id).await?;
+                if !matches!(listening_mode, ListeningMode::EventsOnly(_)) {
+                    self.update_wallet(notification.chain_id).await?;
```
deuszx, line 381: Maybe `update_wallet` shouldn't be adding the chain to the wallet if it's not already there?

linera-core/src/client/chain_client/mod.rs
```diff
@@ -2951,7 +2951,7 @@ impl<Env: Environment> ChainClient<Env> {
 
         let events_only = self
             .listening_mode()
-            .is_some_and(|m| matches!(m, ListeningMode::EventsOnly(_)));
+            .is_none_or(|m| matches!(m, ListeningMode::EventsOnly(_)));
```
deuszx, line 2954: so `None` means `ListeningMode::EventOnly`? I find it confusing that `None` and `Some(EventsOnly)` are equivalent

afck: I think the chain listener always sets the listening mode before starting to listen.

But the CLI commands will also start to listen when trying to make changes in some cases (waiting for new round or something). I'm not sure if they always set a listening mode… but they _do_ sync first anyway if it's their own chain.

linera-client/src/chain_listener.rs
afck: Maybe, yes.

But that sounds like a change that we should carefully plan and do on `main` first, rather than on `testnet_conway`.

deuszx: Do all validators already emit the `NewEvents` notifications? If yes, then this branch can be simplified (we don't have to call `self.process_new_events` here) and just return early if mode is `EventsOnly`.

afck: That's true… and it's enough if a quorum of them does, I guess.

Anyway, maybe also something for a follow-up PR?

PR #5812, opened 2026-03-26
More efficient certificate downloads.

afck:
## Motivation

Downloading certificates is still calling validators in a loop, so a slow one can slow down the client a lot. On `testnet_conway`, this is already faster and uses the requests scheduler.

## Proposal

Use the method from `testnet_conway`.

In addition, deduplicate and fix some code and properly use the height returned from `load_local_certificates` instead of re-downloading the blocks we found locally.

## Test Plan

CI

## Release Plan

- Backport: Some of the fixes may apply to `testnet_conway`, too.

## Links

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -517,17 +517,17 @@ impl<Env: Environment> Client<Env> {
     }
 
     /// Loads and processes certificates from local storage for the given chain, from the
-    /// current local height up to `stop`. Returns the chain info after processing.
+    /// current local height up to `end`. Returns the chain info after processing.
     async fn load_local_certificates(
         &self,
         chain_id: ChainId,
-        stop: BlockHeight,
+        end: BlockHeight,
     ) -> Result<Box<ChainInfo>, chain_client::Error> {
-        let mut info = self.local_node.chain_info(chain_id).await?;
-        let next_height = info.next_block_height;
+        let mut last_info = self.local_node.chain_info(chain_id).await?;
+        let next_height = last_info.next_block_height;
         let hashes = self
             .local_node
-            .get_preprocessed_block_hashes(chain_id, next_height, stop)
+            .get_preprocessed_block_hashes(chain_id, next_height, end)
             .await?;
         let certificates = self.storage_client().read_certificates(&hashes).await?;
         let certificates = match ResultReadCertificates::new(certificates, hashes) {
@@ -537,9 +537,9 @@ impl<Env: Environment> Client<Env> {
             }
         };
         for certificate in certificates {
-            info = self.handle_certificate(certificate).await?.info;
+            last_info = self.handle_certificate(certificate).await?.info;
         }
-        Ok(info)
+        Ok(last_info)
     }
 
     /// Downloads and processes all certificates up to (excluding) the specified height from the


linera-core/src/client/mod.rs
```diff
@@ -467,59 +465,21 @@ impl<Env: Environment> Client<Env> {
...
-                Ok(None) => {}
-            }
+        let validators = self.validator_nodes().await?;
+        let mut info = Box::pin(self.fetch_chain_info(chain_id, &validators)).await?;
+        if target_next_block_height <= info.next_block_height {
+            return Ok(info);
         }
+        info = self
+            .download_certificates_using_all(&validators, chain_id, target_next_block_height)
+            .await?;
```
MathieuDutSik, line 484: I would actually move the code of `download_certificates_using_all` there.

linera-core/src/client/mod.rs
```diff
@@ -569,12 +539,53 @@ impl<Env: Environment> Client<Env> {
...
+        stop: BlockHeight,
+    ) -> Result<Box<ChainInfo>, chain_client::Error> {
+        let mut last_info = self.load_local_certificates(chain_id, stop).await?;
+        let mut next_height = last_info.next_block_height;
+        // Download remaining batches using all validators with staggered fallback.
+        while next_height < stop {
+            let limit = u64::from(stop)
+                .checked_sub(u64::from(next_height))
+                .ok_or(ArithmeticError::Overflow)?
+                .min(self.options.certificate_download_batch_size);
```
MathieuDutSik: Maybe we can avoid the use of the `certificate_download_batch_size` by using async iterators.

linera-core/src/client/requests_scheduler/scheduler.rs
```diff
@@ -369,6 +369,44 @@ impl<Env: Environment> RequestsScheduler<Env> {
...
+        chain_id: ChainId,
+        start: BlockHeight,
+        limit: u64,
+        timeout: Duration,
+    ) -> Result<Vec<ConfirmedBlockCertificate>, NodeError> {
+        if peers.is_empty() {
+            return Err(NodeError::NoValidators);
+        }
+        let heights = (start.0..start.0 + limit)
+            .map(BlockHeight)
```
MathieuDutSik, line 386: Maybe we could introduce the relevant traits for `BlockHeight` to avoid those constructions.

afck: Unfortunately the `Step` trait that would allow us to write `height0..height1` is still experimental and can't be used in stable Rust yet: https://doc.rust-lang.org/std/range/struct.Range.html#impl-IntoIterator-for-Range%3CA%3E

linera-core/src/client/mod.rs
afck: I'm not sure. This way we have exact control over how many we download at a time. Anyway, I'd say we can revisit this later.

linera-core/src/client/mod.rs
afck: Inlined in https://github.com/linera-io/linera-protocol/pull/5812/commits/fe0801ba62ff22c4e8113262c76462268eab793b.

PR #5813, opened 2026-03-26
[testnet] Deduplicate certificate download/processing code

afck:
Backport of the fixes in #5812.

## Motivation

In #5812 some inefficiencies and duplications were fixed.

## Proposal

- Extract `load_local_certificates` helper to share local-storage loading between `download_certificates_from` and `download_certificates_using_all`. It now re-fetches chain info after processing, so callers get the updated `next_block_height` (previously the original height was returned, causing redundant re-downloads).
- Unify `process_certificates_using_all` into `process_certificates` by accepting a slice of nodes instead of a single node reference.
- Both download methods now return non-optional `Box<ChainInfo>`, simplifying their callers.

## Test Plan

CI

## Release Plan

- 
## Links

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -509,17 +509,17 @@ impl<Env: Environment> Client<Env> {
     }
 
     /// Loads and processes certificates from local storage for the given chain, from the
-    /// current local height up to `stop`. Returns the chain info after processing.
+    /// current local height up to `end`. Returns the chain info after processing.
     async fn load_local_certificates(
         &self,
         chain_id: ChainId,
-        stop: BlockHeight,
+        end: BlockHeight,
     ) -> Result<Box<ChainInfo>, chain_client::Error> {
-        let mut info = self.local_node.chain_info(chain_id).await?;
-        let next_height = info.next_block_height;
+        let mut last_info = self.local_node.chain_info(chain_id).await?;
+        let next_height = last_info.next_block_height;
         let hashes = self
             .local_node
-            .get_preprocessed_block_hashes(chain_id, next_height, stop)
+            .get_preprocessed_block_hashes(chain_id, next_height, end)
             .await?;
         let certificates = self.storage_client().read_certificates(&hashes).await?;
         let certificates = match ResultReadCertificates::new(certificates, hashes) {
@@ -529,9 +529,9 @@ impl<Env: Environment> Client<Env> {
             }
         };
         for certificate in certificates {
-            info = self.handle_certificate(certificate).await?.info;
+            last_info = self.handle_certificate(certificate).await?.info;
         }
-        Ok(info)
+        Ok(last_info)
     }
 
     /// Downloads and processes all certificates up to (excluding) the specified height from the


linera-core/src/client/requests_scheduler/scheduler.rs
```diff
@@ -379,6 +379,10 @@ impl<Env: Environment> RequestsScheduler<Env> {
         limit: u64,
         timeout: Duration,
     ) -> Result<Vec<ConfirmedBlockCertificate>, NodeError> {
+        debug_assert!(
```
ma2bd: I generally prefer a plain `assert!` or an error to avoid differences between test and production.

afck: I removed it. (There's no point adding that assertion specifically here: if we really created an epoch with an _empty_ committee, all kinds of things would break! Maybe we should add an assertion somewhere in `committee.rs` instead? Anyway, not really related to this PR.)

linera-core/src/client/mod.rs
```diff
@@ -505,8 +498,22 @@ impl<Env: Environment> Client<Env> {
             }
         };
         for certificate in certificates {
-            last_info = Some(self.handle_certificate(certificate).await?.info);
+            self.handle_certificate(certificate).await?;
         }
+        Ok(self.local_node.chain_info(chain_id).await?)
```
ma2bd: Wasn't the point to avoid calling this function one more time?

afck: Right! Fixed in https://github.com/linera-io/linera-protocol/pull/5813/changes/3ebe9d7f5b68a6d42354af0ab99981540da86293.

linera-core/src/client/mod.rs
```diff
@@ -505,8 +498,22 @@ impl<Env: Environment> Client<Env> {
             }
         };
         for certificate in certificates {
-            last_info = Some(self.handle_certificate(certificate).await?.info);
+            info = self.handle_certificate(certificate).await?.info;
```
MathieuDutSik: Actually here, the `last_info` does make sense.

linera-core/src/client/mod.rs
```diff
@@ -480,19 +477,15 @@ impl<Env: Environment> Client<Env> {
...
-    /// given validator.
-    #[instrument(level = "trace", skip_all)]
-    async fn download_certificates_from(
+    /// Loads and processes certificates from local storage for the given chain, from the
+    /// current local height up to `stop`. Returns the chain info after processing.
+    async fn load_local_certificates(
         &self,
-        remote_node: &RemoteNode<Env::ValidatorNode>,
         chain_id: ChainId,
         stop: BlockHeight,
```
MathieuDutSik: I find the `stop` inexpressive as to the meaning.
It can mean `last_block_height` or `last_batch_height` and the meaning is different for each of them.
Unfortunately, we have both use cases, so we can't simplify that.

afck: It's actually called `end` in lots of other places, so I'll rename `stop` → `end` for now.

PR #5815, opened 2026-03-26
Log IncorrectOutcome as error instead of debug

ndr-ds:
## Motivation

`WorkerError::IncorrectOutcome` fires when a certified block (already signed by a
validator quorum) produces a different execution result on the local node. This
indicates state corruption, non-determinism, or a consensus-breaking bug — all serious
local node issues.

Despite this, it was classified as `is_local() = false`, which means both the gRPC and
simple servers logged it at `debug` level — effectively invisible in production.

## Proposal

Move `IncorrectOutcome` from the `is_local() = false` branch to `is_local() = true` in
`WorkerError::is_local()`. This causes the `log_error` methods in both
`linera-rpc/src/grpc/server.rs` and `linera-rpc/src/simple/server.rs` to emit it at
`error!` level instead of `debug!`.

Audited all other errors in the `false` branch (`CryptoError`, `InvalidOwner`,
`InvalidSigner`, `UnexpectedBlockHeight`, `InvalidEpoch`, `InvalidTimestamp`,
`BlobsNotFound`, etc.) — they are correctly classified as potentially caused by bad peer
messages. Also audited `ChainError::is_local()` and `ExecutionError::is_local()` — no
other misclassifications found.

## Test Plan

CI

### DIFF
--- a/linera-core/src/worker.rs
+++ b/linera-core/src/worker.rs
@@ -413,7 +413,6 @@ impl WorkerError {
             | WorkerError::InvalidEpoch { .. }
             | WorkerError::EventsNotFound(_)
             | WorkerError::InvalidBlockChaining
-            | WorkerError::IncorrectOutcome { .. }
             | WorkerError::InvalidTimestamp { .. }
             | WorkerError::MissingCertificateValue
             | WorkerError::InvalidLiteCertificate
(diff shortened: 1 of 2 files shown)


linera-core/src/worker.rs
```diff
@@ -433,7 +432,8 @@ impl WorkerError {
             | WorkerError::ChainActorSendError { .. }
             | WorkerError::ChainActorRecvError { .. }
             | WorkerError::Thread(_)
-            | WorkerError::ReadCertificatesError(_) => true,
+            | WorkerError::ReadCertificatesError(_)
+            | WorkerError::IncorrectOutcome { .. } => true,
```
ma2bd, line 436: @afck To double-check but indeed this error only happened on already certified blocks.

afck: Yes, it's created only in `process_confirmed_block`.

PR #5818, opened 2026-03-26
Add certificate_upload_batch_size CLI option for ValidatorUpdater

ndr-ds:
## Motivation

The batch size used by `ValidatorUpdater::sync_chain_height` (added in #5800) was
hardcoded to `DEFAULT_CERTIFICATE_DOWNLOAD_BATCH_SIZE`. Reviewers pointed out this is a
conceptually different operation — reading from local storage and uploading to
validators, not downloading — and should have its own configurable option.

## Proposal

- Add `--certificate-upload-batch-size` CLI option (default 500) separate from the
existing `--certificate-download-batch-size`
- Add `certificate_upload_batch_size` field to `chain_client::Options`
- Add `pub(crate) fn options()` accessor on `Client` so `ValidatorUpdater` can read the
configured value
- Wire the new option through the full CLI → `ChainClientOptions` → `ValidatorUpdater`
path

## Test Plan

CI

### DIFF
--- a/linera-core/src/updater.rs
+++ b/linera-core/src/updater.rs
@@ -33,7 +33,7 @@ use tokio::sync::mpsc;
 use tracing::{instrument, Level};
 
 use crate::{
-    client::{chain_client, Client, DEFAULT_CERTIFICATE_DOWNLOAD_BATCH_SIZE},
+    client::{chain_client, Client},
     data_types::{ChainInfo, ChainInfoQuery},
     environment::Environment,
     node::{CrossChainMessageDelivery, NodeError, ValidatorNode},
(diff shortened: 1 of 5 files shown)


linera-core/src/updater.rs
```diff
@@ -778,7 +778,7 @@ where
             return Ok(info);
         }
 
-        let batch_size = DEFAULT_CERTIFICATE_DOWNLOAD_BATCH_SIZE as usize;
+        let batch_size = self.client.options().certificate_download_batch_size as usize;
```
ma2bd: We should use a different option. This one is for downloading then uploading certificates to validators!

afck: Right, good point.

ma2bd: (This is what I was trying to explain during the call)

ma2bd: It is not even downloading by the way, just reading from the local node.

ma2bd: So I would suggest `certificate_upload_batch_size`

PR #5820, opened 2026-03-26
[testnet] Web client: add in-memory storage backend

ma2bd:
## Motivation

The web client currently only supports IndexedDB for storage, which requires a browser environment with persistent storage. In-memory storage was used recently to work around IDB crashes in Chrome.

## Proposal

- Add a `StorageKind` enum (`IndexedDb` / `Memory`) to let callers choose the storage backend when creating a `Client`.
- Introduce `MemStorage` and `MemEnvironment` type aliases alongside the existing `IdbStorage` / `IdbEnvironment`.
- Add `get_mem_storage()` factory for the in-memory backend.
- Refactor `Client`, `Chain`, and `Application` internals to use `ClientContextInner` / `ChainClientInner` enums that dispatch over both backends, with helper macros (`with_chain_client!`, `with_client_and_chain!`) to reduce boilerplate.
- `Client::new` now accepts a `storage_kind` parameter; the rest of the API is unchanged.

## Test Plan

- Existing tests continue to pass (no behavioural change for the IndexedDB path).
- Manual verification that the memory backend initialises and serves queries in a browser environment.

## Release Plan

- Already deployed in `pm-app`
- To be forward-ported to `main`

### DIFF
--- a/web/@linera/client/src/client.rs
+++ b/web/@linera/client/src/client.rs
@@ -276,6 +276,8 @@ impl Client {
             &options,
             default,
             genesis_config,
+            linera_core::worker::DEFAULT_BLOCK_CACHE_SIZE,
+            linera_core::worker::DEFAULT_EXECUTION_STATE_CACHE_SIZE,
         )
         .await?;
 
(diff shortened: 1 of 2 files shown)


deuszx:
Not a fan of added complexity just for the sake of "debugging" storage backend. Maybe if it could be hidden behind already-existing `Storage` API.

ma2bd:
> Not a fan of added complexity just for the sake of "debugging" storage backend. Maybe if it could be hidden behind already-existing `Storage` API.

That's not immediately doable because of types. @Twey has an idea, though (pushing more code to JS)

PR #5792, opened 2026-03-24
Backport #5602: Use libunwind for jemalloc memory profiling stack traces

ndr-ds:
## Motivation

Backport of #5602, stacked on the backport of #5601. Switches jemalloc's stack trace
collection from the default (frame pointers) to libunwind, which produces complete stack
traces in release builds without requiring frame pointers.

## Proposal

Cherry-pick of `9a3f534cbdb` onto the backport of #5601.

## Test Plan

CI

PR #5860, opened 2026-03-27
Add version to tikv-jemallocator git dependency to fix docs CI

ndr-ds:
## Motivation

The `test-crates-and-docrs` CI job has been failing on `testnet_conway` since the
backport of #5602 (#5792). `cargo package` requires all dependencies to have a `version`
field — the backport included the git spec but dropped `version = "0.6"`.

## Proposal

Add `version = "0.6"` to the `tikv-jemallocator` workspace dependency, matching what
#5602 has on `main`. The `[patch.crates-io]` section still ensures the git fork is used
at build time; the version is only needed so `cargo package` knows what to reference
when
stripping git specs for packaging.

Also temporarily adds a `pull_request` trigger to the Documentation workflow so the fix
can be verified on this PR before merging. **This trigger should be removed before or
after merge.**

## Test Plan

- CI — the `test-crates-and-docrs` job will run on this PR via the temporary trigger

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -261,7 +261,7 @@ test-log = { version = "0.2.15", default-features = false, features = [
 test-strategy = "0.3.1"
 thiserror = "1.0.65"
 thiserror-context = "0.1.1"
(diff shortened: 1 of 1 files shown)


ndr-ds:
Merging since this partially breaks the docs tests

PR #5876, opened 2026-03-30
[testnet] Add opt-in recovery from inbox gaps and exec outcome mismatch.

afck:
## Motivation

If a chain confirms the receipt of a cross-chain message and the sender chain updates its outbox accordingly but the recipient chain's update of its inbox somehow fails to be persisted, the recipient chain may never recover because it rejects later cross-chain updates.

## Proposal

Add an `--allow-revert-confirm` config option to the server. If this is enabled, the recipient chain will request that the sender chain resets its outbox and retries the messages.

In addition, add a `--reset-on-incorrect-outcome-mins` option: Servers that encounter an incorrect block execution outcome for a _confirmed_ block (should only be possible due to DB corruption in theory) will re-execute the chain in question. (But only if the last re-execution is longer ago than the specified number of minutes.)

## Test Plan

Tests were added.

## Release Plan

- These changes should be ported to `main`, and
- Affected validators should be restarted with the `--allow-revert-confirm` option.

## Links

### DIFF
--- a/linera-service-graphql-client/gql/service_schema.graphql
+++ b/linera-service-graphql-client/gql/service_schema.graphql
@@ -413,6 +413,12 @@ type ChainStateExtendedView {
 	compatibility with pre-existing database entries).
 	"""
 	nonemptyInboxes: [ChainId!]
+	"""
+	The local wall-clock time when block 0 was last executed. Used to prevent
+	reset-on-incorrect-outcome from looping: if not enough time has elapsed since
+	the last reset, the error is returned instead.
+	"""
+	blockZeroExecutedAt: Timestamp!
 }
 
 """


linera-core/src/chain_worker/state.rs
```diff
@@ -1159,6 +1259,161 @@ where
...
+        let (info, mut actions, outcome) = result;
+        for sender_id in sender_ids {
+            actions
+                .cross_chain_requests
+                .push(CrossChainRequest::RevertConfirm {
+                    sender: sender_id,
+                    recipient: chain_id,
+                    missing_height: BlockHeight::ZERO,
+                });
+        }
```
bart-linera, line 1401: Is it okay to do that _after_ processing the certificates? Or should that be okay because they can be "removed by anticipation" while the certificates are processed?

afck: Why not? So far the actions only contain the _outgoing_ updates (where we are the sender). Their order with respect to the `RevertConfirm` (which concerns us as the recipient) shouldn't matter?

bart-linera: I was thinking that because we reset the whole chain state above, including the inboxes, we'll be missing the messages that the certificates we are processing will be trying to receive. IIUC, adding those messages to the inboxes is the purpose of this `RevertConfirm`, isn't it?

afck: Yes, adding them to `added_bundles` or removing them from `removed_bundles`.

Encountering a received message during execution that is not already in `added_bundles` causes it to get added to `removed_bundles`. Later, when the sender chain sends us the update, we will remove it from `removed_bundles` again.

PR #5877, opened 2026-03-30
Chunk large UpdateRecipient messages to fit gRPC size limits

ma2bd:
## Motivation

Avoid hitting GRPC limits

## Proposal

A single `UpdateRecipient` with bundles spanning a long chain history can easily exceed the 16 MiB gRPC message limit.

* Split the bundle list in `create_cross_chain_requests`: when the estimated serialized size approaches ~11.2 MiB (70% of the limit), flush the current batch and start a new `UpdateRecipient`.
* Also adds `MessageBundle::estimated_size()` and `PostedMessage::estimated_size()` for rough byte-size estimation.

**Important:** We do not send all the chunks at once because the sending queue of `linera-rpc` has size 1.

## Test Plan

CI

### DIFF
--- a/linera-core/src/chain_worker/state.rs
+++ b/linera-core/src/chain_worker/state.rs
@@ -470,6 +470,29 @@ where
     #[instrument(skip_all, fields(
         chain_id = %self.chain_id()
     ))]
+    /// Creates cross-chain requests for a single recipient from its outbox.
+    async fn create_cross_chain_actions_for_recipient(
+        &self,
+        recipient: ChainId,
+    ) -> Result<NetworkActions, WorkerError> {
+        let outbox = self.chain.outboxes.try_load_entry(&recipient).await?;
+        let Some(outbox) = outbox else {
+            return Ok(NetworkActions::default());
+        };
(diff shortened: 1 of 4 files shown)


linera-core/src/chain_worker/state.rs
```diff
@@ -470,6 +470,29 @@ where
     #[instrument(skip_all, fields(
         chain_id = %self.chain_id()
     ))]
+    /// Creates cross-chain requests for a single recipient from its outbox.
```
afck, line 473: (Duplicate doc comment; should be before attributes.)

linera-core/src/updater.rs
```diff
@@ -260,7 +260,7 @@ where
     }
 
     #[instrument(
-        level = "trace", skip_all, err(level = Level::DEBUG),
+        level = "warn", skip_all, err(level = Level::DEBUG),
```
afck, line 263: Is this deliberate?

afck:
Closing this; I merged https://github.com/linera-io/linera-protocol/pull/5879.

PR #5879, opened 2026-03-31
Chunk large UpdateRecipient messages to fit gRPC size limits

afck:
This is https://github.com/linera-io/linera-protocol/pull/5877 with one nit-picky commit added.

## Motivation

Avoid hitting GRPC limits

## Proposal

A single `UpdateRecipient` with bundles spanning a long chain history can easily exceed the 16 MiB gRPC message limit.

* Split the bundle list in `create_cross_chain_requests`: when the estimated serialized size approaches ~11.2 MiB (70% of the limit), flush the current batch and start a new `UpdateRecipient`.
* Also adds `MessageBundle::estimated_size()` and `PostedMessage::estimated_size()` for rough byte-size estimation.

**Important:** We do not send all the chunks at once because the sending queue of `linera-rpc` has size 1.

## Test Plan

CI

### DIFF
--- a/linera-core/src/chain_worker/state.rs
+++ b/linera-core/src/chain_worker/state.rs
@@ -466,11 +466,10 @@ where
         Ok(maybe_blobs)
     }
 
-    /// Loads pending cross-chain requests, and adds `NewRound` notifications where appropriate.
+    /// Creates cross-chain requests for a single recipient from its outbox.
     #[instrument(skip_all, fields(
         chain_id = %self.chain_id()
     ))]
-    /// Creates cross-chain requests for a single recipient from its outbox.
     async fn create_cross_chain_actions_for_recipient(
         &self,
         recipient: ChainId,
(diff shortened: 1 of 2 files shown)


linera-core/src/chain_worker/state.rs
```diff
@@ -597,23 +621,49 @@ where
...
+                        debug!(
+                            "Stopping cross-chain batch for {recipient} at height {height}: \
+                            adding ~{new_size} bytes would exceed chunk limit of {} \
+                            (current batch ~{bundles_size} bytes)",
+                            self.config.cross_chain_message_chunk_limit
+                        );
+                        break;
+                    }
+                }
+                bundles.extend(new_bundles);
```
deuszx, line 656: if `new_bundles` is more than the gRPC message limit then this bundle will never reach the recipient? We'll try to send the `new_bundles`, it will fail at gRPC layer, then we'd go back here for the next batch

afck: I.e. if the "Single block" warning above was printed?

I think the idea is: we might be overestimating the total size (we have this 70% buffer etc.), so it might still fit. And it _should_ still fit: After all, all the outgoing messages did fit in a single confirmed block certificate, which had to also fit entirely in a gRPC message.

PR #5885, opened 2026-04-01
Add `chain_id` to the logs.

MathieuDutSik:
## Motivation

For the purpose of debugging, we want to have more chain_id in the debugging logs.

## Test Plan

CI

## Release Plan

To the `testnet_conway`.

## Links

None

### DIFF
--- a/linera-core/src/chain_worker/actor.rs
+++ b/linera-core/src/chain_worker/actor.rs
@@ -395,7 +395,7 @@ where
             .handle_requests(request_sender, request_receiver)
             .await
         {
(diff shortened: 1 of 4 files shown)


afck:
Not sure if it's better to add the chain ID explicitly to the log messages, or to increase the log level of the `instrument` attributes that already contain the chain ID (like we mentioned yesterday)? Maybe needs discussion.

MathieuDutSik:
That does not seem very useful.

afck:
OK, I made another proposal in https://github.com/linera-io/linera-protocol/pull/6069.

PR #5886, opened 2026-04-01
Fix `handle_revert_confirm`.

afck:
## Motivation

In `handle_revert_confirm` we explicitly add the `missing_height` to the outbox. This was fine originally, when `RevertConfirm` was only used when a genuine gap was detected while handling a confirmed certificate or an `UpdateRecipient` message.

However, when we fully re-execute a chain, we send `RevertConfirm` with `BlockHeight::ZERO` to all senders, even if they didn't send a message at height 0.

Also, we forget the preprocessed blocks when resetting a chain.

## Proposal

~Fix `handle_revert_confirm`.~ (Already fixed in #5890.)

Handle spurious entries in the outbox gracefully. Also handle preprocessed blocks when resetting a chain.

## Test Plan

CI

## Release Plan

- Validator hotfix.

## Links

- Bug was introduced in https://github.com/linera-io/linera-protocol/pull/5876.

### DIFF
--- a/linera-core/src/chain_worker/state.rs
+++ b/linera-core/src/chain_worker/state.rs
@@ -593,7 +593,11 @@ where
             let mut bundles_size = 0;
             for height in heights {
                 let Some(hashed_block) = height_to_blocks.get(&height) else {
-                    tracing::warn!(%height, "ignoring spurious block entry in outbox");
+                    tracing::warn!(
+                        %height,
+                        %recipient,
+                        "spurious entry in outbox; skipping this and higher sender blocks"
+                    );
                     break;
                 };
                 let new_bundles = hashed_block


linera-core/src/chain_worker/state.rs
```diff
@@ -1364,7 +1339,7 @@ where
 
         let mut heights_to_readd = Vec::new();
```
MathieuDutSik: The double `d` is bad style. If it is a typo, correct it.
If not then specify the meaning of the second `d`.

linera-core/src/chain_worker/state.rs
MathieuDutSik: Here we can chain the operations to avoid the drop.

MathieuDutSik: Here we are passing by reference and then cloning the height (admittedly just a `u64`) only because we use `new_heights.len()` later on.

linera-core/src/chain_worker/state.rs
afck: We are adding heights that had been added previously. We are re-adding them.

afck:
> it won't prevent us from adding such entries on RevertConfirm in the future?

It will! I removed the explicit `push(missing_height)` and made the loop take care of this. That way, `missing_height` only gets added if it's actually a sender block.

linera-core/src/chain_worker/state.rs
afck: Done in https://github.com/linera-io/linera-protocol/pull/5886/commits/a89fd020d2b80ae2dd6f1b97ccf89f360b980010.

afck: Done in https://github.com/linera-io/linera-protocol/pull/5886/commits/a89fd020d2b80ae2dd6f1b97ccf89f360b980010.

linera-core/src/chain_worker/state.rs
MathieuDutSik: Ah, ok I read it as "read" + d. Why not write "re_add".

afck: OK, done in https://github.com/linera-io/linera-protocol/pull/5886/changes/812689adbe29d9582df735a0d97ddbd1ccbfa189.

linera-core/src/chain_worker/state.rs
```diff
@@ -1362,10 +1331,10 @@ where
             return Ok(NetworkActions::default());
         };
 
-        let mut heights_to_readd = Vec::new();
+        let mut heights_to_re_add = Vec::new();
         let mut current_height = latest_height;
         while current_height >= missing_height {
-            heights_to_readd.push(current_height);
```
ma2bd, line 1368: Can we document the invariant that `current_height` is a known to be a valid height in `previous_message_blocks` at this point?

afck: Done in https://github.com/linera-io/linera-protocol/pull/5886/commits/29fa0f049e30611e46b1986a83540ca0341aed94.

linera-core/src/chain_worker/state.rs
```diff
@@ -624,9 +592,10 @@ where
             let mut bundles = Vec::new();
             let mut bundles_size = 0;
             for height in heights {
-                let hashed_block = height_to_blocks
-                    .get(&height)
-                    .ok_or_else(|| ChainError::InternalError("missing block".to_string()))?;
+                let Some(hashed_block) = height_to_blocks.get(&height) else {
+                    tracing::warn!(%height, "ignoring spurious block entry in outbox");
+                    break;
```
ma2bd: * "ignoring" is a little understated if we don't continue
* Can we document `break`

afck: I extended the log: https://github.com/linera-io/linera-protocol/pull/5886/commits/d981002f4cbe37ed50ed474a8a877ec93018a670

Or do you mean add a comment in addition to the log?

ma2bd: all good

PR #5875, opened 2026-03-29
Backport bridge to main

deuszx:
## Motivation

Backport the Linera↔EVM bridge relay infrastructure from the `testnet_conway` branch to `main`, enabling active deposit/burn scanning, persistent state tracking, and monitoring capabilities.

## Proposal

This PR brings the bridge relay and supporting infrastructure to `main`, rebased on top of #5888 (structural moves). Key changes:

- **Relay restructuring**: Split monolithic `relay.rs` into `relay/mod.rs`, `relay/evm.rs`, `relay/linera.rs`, `relay/metrics.rs` for better separation of concerns
- **Active scanning**: Continuously scan EVM and Linera chains to detect and retry bridging requests (`monitor/evm.rs`, `monitor/linera.rs`)
- **Persistent state**: Track deposits and burns in SQLite (`monitor/db.rs`) for crash recovery
- **Microchain contract simplification**: Remove sequential block height requirement from `Microchain.sol` — BFT finality from `ConfirmedBlockCertificate` makes it unnecessary
- **EVM finality verification**: Bridge verifies block finality before accepting deposits
- **Bridge relayer improvements**: Docker support, configurable chain owner, health endpoint on exporter
- **`rocksdb` feature on `linera-storage`**: Required by `linera-bridge`

Also removes `test_chain_listener_sparse_event_download` from `chain_listener.rs` tests as it depends on APIs not yet available on `main`.

## Test Plan

- `cargo clippy --locked --all-targets --all-features` passes
- `cargo clippy --locked --no-default-features --features kubernetes` passes
- `cargo clippy --locked --no-default-features` passes
- Bridge e2e tests updated for new Microchain contract (no sequential height)
- New `auto_deposit_scan` e2e test covers active scanning flow

## Release Plan

- Nothing to do / These changes follow the usual release cycle.

## Links

- Depends on #5888 (structural moves extracted from this PR)
- [reviewer checklist](https://github.com/linera-io/linera-protocol/blob/main/CONTRIBUTING.md#reviewer-checklist)

PR #5888, opened 2026-04-01
Structural moves to main

deuszx:
## Motivation

PR #5875 (backport-bridge-to-main) mixes structural code moves with bridge-specific changes,
making it hard to review. This PR extracts the structural moves so they can land independently,
shrinking #5875's diff to just the bridge work.

The reason why `PersistentWallet` and `GenesisConfig` were moved is that I wanted to use both in `linera-bridge` but they were part of `linera-service` which depends on `linera-bridge` – that creates a cyclic dependency. 

## Proposal

Three commits, each independently compilable:

1. **Add `Bytecode::load_from_file_sync`** — synchronous variant of `load_from_file` for
   non-async contexts.

2. **Move `GenesisConfig` from `linera-client` to `linera-core`** — the struct, its `Error`
   type, and `initialize_storage` move to `linera-core::genesis_config`. Re-exported from
   `linera-client::config` for backward compatibility. `CommitteeConfig::into_genesis` replaces
   the old `GenesisConfig::new(CommitteeConfig, ...)` constructor on the client side.

3. **Move `PersistentWallet` from `linera-service` to `linera-core`** — wallet persistence
   logic moves to `linera-core::environment::wallet::persistent`, gated behind a new `fs`
   feature. Display logic (`pretty_print`, `ChainDetails`) stays in `linera-service` behind a
   `WalletExt` trait. Also fixes a `thiserror_context` type ambiguity in `client_context.rs`
   triggered by `linera-persistent` entering the dependency graph.

All public API signatures are preserved — this is a pure structural move.

## Test Plan

- All three commits compile independently and pass `cargo clippy -D warnings` against
   `linera-service` (which pulls in the full dependency graph including feature-gated paths).
- No behavior changes — existing tests cover the moved code.

## Release Plan

## Links

- Unblocks #5875 by extracting structural moves from the bridge backport.

### DIFF
--- a/linera-core/src/environment/wallet/persistent.rs
+++ b/linera-core/src/environment/wallet/persistent.rs
@@ -1,0 +1,205 @@
+// Copyright (c) Zefchain Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+use std::{
+    iter::IntoIterator,
+    sync::{Arc, RwLock},
+};
+
+use futures::{stream, Stream};
+use linera_base::identifiers::{AccountOwner, ChainId};
+use linera_persistent as persistent;
+
+use super::{Chain, Memory, Wallet};
+use crate::GenesisConfig;
+
+#[derive(serde::Serialize, serde::Deserialize)]
+struct Data {
+    pub chains: Memory,
+    default: Arc<RwLock<Option<ChainId>>>,
+    genesis_config: GenesisConfig,
+}
+
+pub struct PersistentWallet(persistent::File<Data>);
+
+// TODO(#5081): `persistent` is no longer necessary here, we can move the locking
+// logic right here
+
+impl Wallet for PersistentWallet {
+    type Error = persistent::file::Error;
+
+    async fn get(&self, id: ChainId) -> Result<Option<Chain>, Self::Error> {
+        Ok(self.get(id))
+    }
+
+    async fn remove(&self, id: ChainId) -> Result<Option<Chain>, Self::Error> {
+        self.remove(id)
+    }
+
+    fn items(&self) -> impl Stream<Item = Result<(ChainId, Chain), Self::Error>> {
+        stream::iter(self.items().into_iter().map(Ok))
+    }
+
+    async fn insert(&self, id: ChainId, chain: Chain) -> Result<Option<Chain>, Self::Error> {
+        self.insert(id, &chain)
+    }
+
+    async fn try_insert(&self, id: ChainId, chain: Chain) -> Result<Option<Chain>, Self::Error> {
+        let chain = self.try_insert(id, chain)?;
+        self.save()?;
+        Ok(chain)
+    }
+
+    async fn modify(
+        &self,
+        id: ChainId,
+        f: impl Fn(&mut Chain) + Send,
+    ) -> Result<Option<()>, Self::Error> {
+        self.mutate(id, f).transpose()
+    }
+}
+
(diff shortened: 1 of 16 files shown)


linera-core/src/environment/wallet/persistent.rs
```diff
@@ -0,0 +1,205 @@
...
+    genesis_config: GenesisConfig,
+}
+
+pub struct PersistentWallet(persistent::File<Data>);
+
+// TODO(#5081): `persistent` is no longer necessary here, we can move the locking
+// logic right here
+
+impl Wallet for PersistentWallet {
+    type Error = persistent::file::Error;
```
Twey, line 29: Let's not hardcode filesystem operations in `linera-core`.  `linera-core` should remain as independent as possible from the system I/O — it's meant to hold only the protocol algorithms themselves.

linera-core/src/genesis_config.rs
```diff
@@ -0,0 +1,175 @@
...
+    ownership::ChainOwnership,
+};
+use linera_execution::committee::Committee;
+use linera_storage::Storage;
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error("I/O error: {0}")]
+    IoError(#[from] std::io::Error),
```
Twey, line 21: Ditto above — this forces the use of specific I/O interfaces in `linera-core`.

Twey:
I think I'm a bit confused about why this is necessary for the bridge.  The bridge depends on the filesystem wallet definition?

deuszx:
> I think I'm a bit confused about why this is necessary for the bridge. The bridge depends on the filesystem wallet definition?

It wants to use the persistent wallet which, before this PR, lived in linera-service.

linera-core/src/environment/wallet/persistent.rs
deuszx: This is behind an `fs` feature flag

deuszx:
Alternatively I can introduce a `linera-wallet` crate where I'd put both of these.

linera-core/src/environment/wallet/persistent.rs
Twey: Nevertheless — let's not use feature flags to inject dependencies.  The `Wallet` trait exists so that this isn't necessary.

Trying to predecide all the possible implementations of an interface and then baking them into the crate that defines the interface (behind feature flags or otherwise) is much more cumbersome than exposing an extension point from the crate with a well-defined and type-checked interface that users can implement as they like.

Twey:
The filesystem wallet in `linera-service` is there because it encodes a specific on-disk wallet implementation that is defined by `linera-service`.  I think it's a bit suspicious to want to manage the CLI tools' wallet format without using the CLI tools.

If we did really want to remove ownership of the format from the CLI tools, we could move it into some kind of `linera-wallet-json-file` crate (n.b. *not* `linera-wallet`: the standardized ‘Linera wallet’ interface is defined by the `Wallet` trait in `linera-core`; this is not that but something much more specific)

linera-core/src/environment/wallet/persistent.rs
deuszx: But the file-based implementation is behind the feature flag only. Not the trait. This Is so that it's possible to depend on this crate without file system

deuszx:
Ok, I will just duplicate the code in Linera bridge.

PR #5891, opened 2026-04-01
Add logging and metrics for journaling and value splitting

ma2bd:
## Motivation

We are investigating intermittent execution state corruption on validators. A potential cause is partial writes during the journaling slow path in the view layer. Adding observability to the journaling and value-splitting backends will help confirm whether these code paths are triggered in production and whether failures occur during journal resolution.

## Proposal

- Add tracing to the journaling backend: warn when slow path is triggered, error when journal resolution fails, info on journal progress, warn when pending journal found on reload.
- Add Prometheus metrics to the journaling backend: journal_fastpath_count, journal_slowpath_count, journal_resolution_failures, journal_pending_on_load, journal_batch_len.
- Add a warn log and value_split_count metric to the value-splitting backend.

## Test Plan

CI

## Release Plan

These changes should be backported to `main`

### DIFF
--- a/linera-views/src/backends/journaling.rs
+++ b/linera-views/src/backends/journaling.rs
@@ -27,7 +27,9 @@ use thiserror::Error;
 mod metrics {
     use std::sync::LazyLock;
 
-    use linera_base::prometheus_util::{register_histogram, register_int_counter};
+    use linera_base::prometheus_util::{
+        exponential_bucket_interval, register_histogram, register_int_counter,
+    };
     use prometheus::{Histogram, IntCounter};
 
     /// Number of write_batch calls that used the fast path (single atomic batch).
@@ -67,7 +69,7 @@ mod metrics {
         register_histogram(
             "journal_batch_len",
             "Number of operations in write_batch calls",
-            None,
+            exponential_bucket_interval(1.0, 10000.0),
         )
     });
 }


linera-views/src/backends/journaling.rs
```diff
@@ -23,6 +23,55 @@ use serde::{Deserialize, Serialize};
...
+    /// Number of write_batch calls that used the fast path (single atomic batch).
+    pub static JOURNAL_FASTPATH_COUNT: LazyLock<IntCounter> = LazyLock::new(|| {
+        register_int_counter(
+            "journal_fastpath_count",
+            "Number of write_batch calls using the fast path",
+        )
+    });
+
+    /// Number of write_batch calls that required journaling.
+    pub static JOURNAL_SLOWPATH_COUNT: LazyLock<IntCounter> = LazyLock::new(|| {
```
ma2bd, line 44: @ndr-ds Should we using labels or something?

afck:
> These changes should be backported to the latest testnet branch

(They should be ported to `main`.)

linera-views/src/backends/value_splitting.rs
```diff
@@ -269,6 +285,13 @@ where
                     let value_ext = if value.len() <= K::MAX_VALUE_SIZE - 4 {
                         Self::get_initial_count_first_chunk(count, &value)?
                     } else {
+                        tracing::warn!(
```
afck, line 288: Is this really unexpected enough to be warn-level?

linera-views/src/backends/journaling.rs
```diff
@@ -23,6 +23,55 @@ use serde::{Deserialize, Serialize};
...
+
+    /// Number of journal resolution failures.
+    pub static JOURNAL_RESOLUTION_FAILURES: LazyLock<IntCounter> = LazyLock::new(|| {
+        register_int_counter(
+            "journal_resolution_failures",
+            "Number of journal resolution failures (potential data inconsistency)",
+        )
+    });
+
+    /// Number of pending journals found during clear_journal (on chain reload).
```
afck:
```suggestion
    /// Number of pending journals found during `clear_journal` (on chain reload).
```

linera-views/src/backends/journaling.rs
```diff
@@ -23,6 +23,55 @@ use serde::{Deserialize, Serialize};
...
+            "Number of pending journals found during chain reload",
+        )
+    });
+
+    /// Histogram of batch sizes (number of operations) for write_batch calls.
+    pub static JOURNAL_BATCH_LEN: LazyLock<Histogram> = LazyLock::new(|| {
+        register_histogram(
+            "journal_batch_len",
+            "Number of operations in write_batch calls",
+            None,
```
afck: (What are the buckets if not specified?)

linera-views/src/backends/value_splitting.rs
ma2bd: The limit is huge for ScyllaDb. We can fix if we ever use DynamoDb again

linera-views/src/backends/journaling.rs
ndr-ds: Depends on what labels you wanna use? I guess we can also correlate the slowpath increase with the warnings on the logs 🤔

linera-views/src/backends/journaling.rs
ndr-ds: Yes we should definitely specify buckets here. I think the default creates buckets until the value 10 or something like that, probably way too low for this

PR #5892, opened 2026-04-01
Db journaling fix

ma2bd:
## Motivation

#5891 added logs but now we would like to prevent a possible DB corruption happening with journaling.

## Proposal

* Distinguish DB errors that may force a reload of the view
* This is the case for certain journaling errors during save().

## Test Plan

CI

## Release Plan

These changes should be backported to the `main` branch

### DIFF
--- a/linera-views/src/backends/journaling.rs
+++ b/linera-views/src/backends/journaling.rs
@@ -109,10 +109,6 @@ pub enum JournalingError<E> {
     #[error(transparent)]
     BcsError(bcs::Error),
 
-    /// The journal block could not be retrieved.
-    #[error("The journal block could not be retrieved, it could be missing or corrupted.")]
-    FailureToRetrieveJournalBlock,
-
     /// Refusing to use the journal without exclusive access.
     #[error("Refusing to use the journal without exclusive database access to the root object.")]
     JournalRequiresExclusiveAccess,
@@ -120,7 +116,23 @@ pub enum JournalingError<E> {
     /// Journal resolution failed; storage may be in an inconsistent state.
     /// The view must be reloaded to complete the pending journal.
     #[error("Journal resolution failed: {0}")]
-    ResolutionFailed(Box<dyn std::error::Error + Send + Sync>),
+    JournalResolutionFailed(JournalingResolutionError<E>),
+}
+
+/// Error type for the journaling key-value store layer.
+#[derive(Error, Debug)]
+pub enum JournalingResolutionError<E> {
+    /// Error from the inner store.
+    #[error(transparent)]
+    Inner(#[from] E),
+
+    /// BCS serialization error.
+    #[error(transparent)]
+    BcsError(bcs::Error),
+
+    /// The journal block could not be retrieved.
+    #[error("The journal block could not be retrieved, it could be missing or corrupted.")]
+    FailureToRetrieveJournalBlock,
 }
 
 impl<E: KeyValueStoreError> From<bcs::Error> for JournalingError<E> {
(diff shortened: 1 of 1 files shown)


linera-views/src/backends/journaling.rs
```diff
@@ -486,7 +525,7 @@ where
         }
         let header = JournalHeader { block_count };
         if block_count > 0 {
-            let value = bcs::to_bytes(&header)?;
+            let value = bcs::to_bytes(&header).map_err(JournalingError::BcsError)?;
```
MathieuDutSik: Those conversions for `BcsError` are annoying and are wholly not needed since the code already introduced:
```rust
impl<E: KeyValueStoreError> From<bcs::Error> for JournalingError<E> {
    fn from(error: bcs::Error) -> Self {
        JournalingError::BcsError(error)
    }
}
```

linera-core/src/chain_worker/state.rs
```diff
@@ -161,8 +164,19 @@ where
     }
 
     /// Handles a request and applies it to the chain state.
+    /// Returns `Err(())` if the worker is poisoned and must be reloaded.
     #[instrument(skip_all, fields(chain_id = %self.chain_id()))]
-    pub async fn handle_request(&mut self, request: ChainWorkerRequest<StorageClient::Context>) {
+    pub async fn handle_request(
+        &mut self,
+        request: ChainWorkerRequest<StorageClient::Context>,
+    ) -> Result<(), ()> {
```
MathieuDutSik: I find it difficult to find a worse return type than `Result<(), ()>`.
At least put an enum for the error with just one variant.
Or return a boolean (but that is worse).

linera-views/src/backends/journaling.rs
```diff
@@ -372,12 +403,12 @@ where
...
                 .read_value::<S::Batch>(&block_key)
                 .await?
-                .ok_or(JournalConsistencyError::FailureToRetrieveJournalBlock)?;
+                .ok_or(JournalingError::FailureToRetrieveJournalBlock)?;
             // Execute the block and delete it from the journal atomically.
             batch.add_delete(block_key);
             header.block_count -= 1;
             if header.block_count > 0 {
-                let value = bcs::to_bytes(&header)?;
+                let value = bcs::to_bytes(&header).map_err(JournalingError::BcsError)?;
```
afck: Wouldn't a BCS error inside this loop still mean that we _partially_ resolved the journal? If we return `BcsError` here, `must_reload_view` would be `false`?

linera-core/src/chain_worker/state.rs
ma2bd: Sure we can have an empty error type.

linera-views/src/backends/journaling.rs
ma2bd: somehow Claude refused to do that. Let me try again

linera-core/src/chain_worker/state.rs
```diff
@@ -2230,12 +2247,22 @@ where
...
     ))]
     async fn save(&mut self) -> Result<(), WorkerError> {
         self.clear_shared_chain_view().await;
-        self.chain.save().await?;
+        if let Err(e) = self.chain.save().await {
+            if e.must_reload_view() {
+                tracing::error!(
+                    chain_id = %self.chain_id(),
+                    "Journal resolution failed; marking worker as poisoned: {e}"
+                );
```
deuszx, line 2265: We try to follow the following pattern when logging:
```rust
tracing::error!(
  error=?e,
  chain_id=%self.chain_id(),
  "Journal resolution failed; marking worker as poisoned"
);
```

linera-core/src/chain_worker/state.rs
ma2bd: > I find it difficult to find a worse return type than Result<(), ()>.

you just did!

linera-views/src/backends/journaling.rs
ma2bd: We wrap the errors at the call site. It's a bit ugly, though

ma2bd: fixed (with more boilerplate!)

linera-core/src/chain_worker/state.rs
ma2bd: done

PR #5894, opened 2026-04-01
Make first_non_admin_chain deterministic by sorting chain IDs

ndr-ds:
## Motivation

`first_non_admin_chain()` returns a non-deterministic result across process restarts.
The wallet's in-memory storage uses `papaya::HashMap`, whose iteration order depends on
hash seeds and memory layout. This caused the testnet-conway faucet to silently switch
between different chains on every restart (confirmed via Loki logs showing 4 different
chains across 4 restarts), breaking user claim tracking each time.

## Proposal

Collect and sort the non-admin chain IDs before picking the first one. `ChainId`
ordering is by hash bytes — arbitrary but deterministic, which is all that's needed.

The serialization path (`Memory::serialize`) already sorts for stable JSON output, but
the in-memory iteration paths (`chain_ids()`, `items()`) do not. This fix targets the
only caller that depends on "first" having a stable meaning.

## Test Plan

CI.

## Release Plan

### DIFF
--- a/linera-client/src/client_context.rs
+++ b/linera-client/src/client_context.rs
@@ -3,7 +3,9 @@
 
 use std::sync::Arc;
 
-use futures::{Future, StreamExt as _, TryStreamExt as _};
+#[cfg(not(web))]
+use futures::StreamExt as _;
+use futures::{Future, TryStreamExt as _};
 use linera_base::{
     crypto::{CryptoHash, ValidatorPublicKey},
     data_types::{ChainDescription, Epoch, Timestamp},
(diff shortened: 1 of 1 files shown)


ndr-ds:
* **#5894** <a href="https://app.graphite.com/github/pr/linera-io/linera-protocol/5894?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/linera-io/linera-protocol/5894?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

linera-client/src/client_context.rs
```diff
@@ -381,14 +381,17 @@ impl<Env: Environment> ClientContext<Env> {
...
-        .await
-        .expect("No non-admin chain specified in wallet with no non-admin chain")
-        .map_err(Error::wallet)
+            .try_filter(|chain_id| futures::future::ready(*chain_id != admin_chain_id))
+            .try_collect()
+            .await
+            .map_err(Error::wallet)?;
+        chain_ids.sort();
+        Ok(*chain_ids
+            .first()
```
afck: Or `.into_iter().min()`.

linera-client/src/client_context.rs
```diff
@@ -381,14 +381,17 @@ impl<Env: Environment> ClientContext<Env> {
 
     pub async fn first_non_admin_chain(&self) -> Result<ChainId, Error> {
         let admin_chain_id = self.admin_chain_id();
-        std::pin::pin!(self
+        let mut chain_ids: Vec<ChainId> = self
```
afck: (Type annotation → `try_collect`)

linera-client/src/client_context.rs
```diff
@@ -3,6 +3,7 @@
 
 use std::sync::Arc;
 
+#[allow(unused_imports)]
```
afck: Why? We should use `cfg` attributes instead.

ndr-ds: My bad, should be fixed now

PR #5895, opened 2026-04-01
[backport] Make first_non_admin_chain deterministic by sorting chain IDs

ndr-ds:
## Motivation

Backport of the `first_non_admin_chain` fix from main. The wallet's in-memory
`papaya::HashMap` returns keys in non-deterministic order, causing the faucet to
silently switch chains on every restart. Confirmed on testnet-conway via Loki logs — 4
different chains across 4 restarts.

## Proposal

Collect and sort non-admin chain IDs before picking the first one, making the selection
deterministic regardless of HashMap iteration order.

## Test Plan

CI.

### DIFF
--- a/linera-client/src/client_context.rs
+++ b/linera-client/src/client_context.rs
@@ -3,7 +3,9 @@
 
 use std::sync::Arc;
 
-use futures::{Future, StreamExt as _, TryStreamExt as _};
+#[cfg(not(web))]
+use futures::StreamExt as _;
+use futures::{Future, TryStreamExt as _};
 use linera_base::{
     crypto::{CryptoHash, ValidatorPublicKey},
     data_types::{ChainDescription, Epoch, Timestamp},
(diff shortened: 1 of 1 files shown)


linera-client/src/client_context.rs
```diff
@@ -377,14 +377,17 @@ impl<Env: Environment> ClientContext<Env> {
 
     pub async fn first_non_admin_chain(&self) -> Result<ChainId, Error> {
         let admin_chain_id = self.admin_chain_id();
-        std::pin::pin!(self
+        let mut chain_ids: Vec<ChainId> = self
```
afck: (Type annotation should be on `try_collect` below.)

linera-client/src/client_context.rs
```diff
@@ -377,14 +377,17 @@ impl<Env: Environment> ClientContext<Env> {
...
-        .await
-        .expect("No non-admin chain specified in wallet with no non-admin chain")
-        .map_err(Error::wallet)
+            .try_filter(|chain_id| futures::future::ready(*chain_id != admin_chain_id))
+            .try_collect()
+            .await
+            .map_err(Error::wallet)?;
+        chain_ids.sort();
+        Ok(*chain_ids
+            .first()
```
afck: Or just `chain_ids.into_iter().min()`? No need to sort just to get the minimum.

linera-client/src/client_context.rs
```diff
@@ -3,6 +3,7 @@
 
 use std::sync::Arc;
 
+#[allow(unused_imports)]
```
afck: Let's avoid `unused_*`.

PR #5896, opened 2026-04-02
Correct a clippy error

### LINKED ISSUE #5893: fix clippy warning
```
warning: `if` chain can be rewritten with `match`
   --> linera-chain/src/inbox.rs:308:17
    |
308 | /                 if front_cursor == cursor {
309 | |                     // We already executed this bundle by anticipation. Remove it from
310 | |                     // the queue.
311 | |                     ensure!(
...   |
339 | |                     );
340 | |                 }
    | |_________________^ help: consider rewriting the `if` chain with `match`: `match front_cursor.cmp(&cursor) {...}`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#comparison_chain
    = note: `#[warn(clippy::comparison_chain)]` on by default
```

MathieuDutSik:
## Motivation

Some clippy correction are left out in CI.

Fixes #5893 

## Proposal

Straightforward correction.

## Test Plan

CI

## Release Plan

This is for `testnet_conway`.

## Links

None.

### DIFF
--- a/linera-chain/src/inbox.rs
+++ b/linera-chain/src/inbox.rs
@@ -1,6 +1,8 @@
 // Copyright (c) Zefchain Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
+use std::cmp::Ordering;
+
 use allocative::Allocative;
 use async_graphql::SimpleObject;
 use linera_base::{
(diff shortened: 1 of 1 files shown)


linera-chain/src/inbox.rs
```diff
@@ -305,38 +307,42 @@ where
...
+                                bundle,
+                            }
+                        );
+                        self.removed_bundles.delete_front();
+                        #[cfg(with_metrics)]
+                        metrics::REMOVED_BUNDLES
+                            .with_label_values(&[])
+                            .observe(self.removed_bundles.count() as f64);
+                    }
+                    Ordering::Less => {
```
afck, line 327: Isn't this the `front_cursor < cursor` case, rather than `front_cursor > cursor`?

MathieuDutSik: Thanks for looking. I prefer the comparison `<` to `Ordering::Less`. But we do not want clippy warnings.

The code matching the `Ordering::Less` corresponds to
```rust
                        return Err(InboxError::GapDetected {
                            expected_height: previous_bundle.height,
                            actual_height: bundle.height,
                        });
```
That code section corresponds in the old code to `cursor > front_cursor` or `front_cursor < cursor`.

And `front_cursor < cursor` is exactly the case `Ordering::Less` of `front_cursor.cmp(&cursor)`.

So, I think the change preserves the behavior.

afck: Ah, right. 🤦

afck (approved):

PR #5899, opened 2026-04-02
Fix `download_certificates_for_events`.

afck:
## Motivation

`communicate_concurrently` returns when the first task returns `Ok`. We've been using it wrong in `download_certificates_for_events`, since the closure returns `Ok` even for validators that didn't have all events.

## Proposal

Check if we have all events in the closure.

Also, don't use `chain_state_view()`.

## Test Plan

CI

## Release Plan

- Release SDK
- Port to `main`.

## Links

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -715,9 +715,9 @@ impl<Env: Environment> Client<Env> {
                     .iter()
                     .filter(|id| {
                         id.chain_id == chain_id
-                            && !next_expected
+                            && next_expected
                                 .get(&id.stream_id)
-                                .is_some_and(|index| *index > id.index)
+                                .is_none_or(|index| *index <= id.index)
                     })
                     .cloned()
                     .collect();


linera-core/src/client/mod.rs
```diff
@@ -661,22 +661,43 @@ impl<Env: Environment> Client<Env> {
...
+                            )
+                            .await?;
+                        if required_ref.iter().all(|(stream_id, &max_index)| {
+                            next_expected
+                                .get(stream_id)
+                                .is_some_and(|index| *index > max_index)
+                        }) {
+                            Ok::<(), chain_client::Error>(())
+                        } else {
+                            Err(NodeError::EventsNotFound(event_ids.to_vec()).into())
```
deuszx: Why is it important for every node in the committee to return all expected events? That's not the reason we're `communicate_concurrently`, right?

afck: No… TBH this error is a placeholder. We've had discussions before where I had used `()` as an error type in `communicate_concurrently` because we'd return a different error from the function anyway. Not sure what's appropriate here.

bart-linera (approved):
Looks good for the happy path, but IIUC `download_certificates_for_events` will return all the events as "not found" even if it manages to download some of them from some of the validators?

afck:
@deuszx, @bart-linera: Thanks for the comments! I tried to clean up the error handling; please take another look.

PR #5900, opened 2026-04-02
More complete `RevertConfirm` requests.

afck:
## Motivation

In some cases we send `RevertConfirm` with the lowest height that we _know_ we are missing.

We should instead request the lowest height we _don't_ know we are _not_ missing, to make it more likely that we really get _all_ events.

## Proposal

Rename the field and request the next expected height.

## Test Plan

CI

## Release Plan

- Hotfix validators
- Port to `main`.

## Links


ma2bd:
Good catch!

PR #5902, opened 2026-04-02
Better use the block value cache.

afck:
## Motivation

There are a few places in the chain worker where we read blocks from storage that might be in the cache.

## Proposal

Consult the cache first. Put loaded blocks into the cache.

## Test Plan

CI

## Release Plan

- Port to main
- Hotfix validators
- Release SDK

## Links


ma2bd (approved):
nice!

PR #5920, opened 2026-04-06
Corrections to the handling of incorrect input in Ethereum.

MathieuDutSik:
## Motivation

Some issues were identified in `linera-ethereum`.

## Proposal

The `to_block/from_block` interval was not handled correctly when the interval was empty.

Also, the parsing of event logs in `parse_log` was incorrect for empty inner types list.

## Test Plan

CI and additional tests were added.

## Release Plan

## Links

None.

### DIFF
--- a/linera-ethereum/src/client.rs
+++ b/linera-ethereum/src/client.rs
@@ -162,6 +162,9 @@ where
         from_block: u64,
         to_block: u64,
     ) -> Result<Vec<EthereumEvent>, Self::Error> {
+        if to_block <= from_block {
+            return Ok(Vec::new());
+        }
         let contract_address = contract_address.parse::<Address>()?;
         let event_name = event_name_from_expanded(event_name_expanded);
         let filter = Filter::new()
--- a/linera-ethereum/src/common.rs
+++ b/linera-ethereum/src/common.rs
@@ -99,12 +99,18 @@ pub fn event_name_from_expanded(event_name_expanded: &str) -> String {
     event_name_expanded.replace(" indexed", "").to_string()
 }
 
-fn parse_uint<T>(entry: B256, parser: impl FnOnce(&BigUint) -> Option<T>) -> Result<T, EthereumServiceError> {
+fn parse_uint<T>(
+    entry: B256,
+    parser: impl FnOnce(&BigUint) -> Option<T>,
+) -> Result<T, EthereumServiceError> {
     let entry = BigUint::from_bytes_be(&entry.0);
     parser(&entry).ok_or(EthereumServiceError::EthereumParsingError)
 }
 
-fn parse_int<T>(entry: B256, parser: impl FnOnce(&BigInt) -> Option<T>) -> Result<T, EthereumServiceError> {
+fn parse_int<T>(
+    entry: B256,
+    parser: impl FnOnce(&BigInt) -> Option<T>,
+) -> Result<T, EthereumServiceError> {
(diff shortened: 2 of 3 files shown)


linera-ethereum/src/client.rs
```diff
@@ -162,6 +162,9 @@ where
         from_block: u64,
         to_block: u64,
     ) -> Result<Vec<EthereumEvent>, Self::Error> {
+        if to_block <= from_block {
+            return Ok(Vec::new());
+        }
```
deuszx, line 167: This should be unnecessary. The Filter below handles this when converting to `Range`.

MathieuDutSik: The code block
```rust
        if to_block <= from_block {
            return Ok(Vec::new());
        }
```
is part of the implementation of `read_events` for clients satisfying the trait `JsonRpcClient`.

That trait `JsonRpcClient` is satisfied by `ServiceEthereumClient` and this used for example for implementing `read_events` for the service of Ethereum tracker.

Therefore, it appears that the stated "Filter" does not apply and that the check is needed.

deuszx: TBH I don't understand your reply.

Anyway, another issue with this is that when if `to_block == from_block` you'll short-circuit and return an empty vector but this isn't consistent with the implementation of the `alloy::Filter` which treats those ranges as inclusive so `to_block == from_block` is valid and should scan that one block for events.

linera-ethereum/src/common.rs
```diff
@@ -200,22 +211,90 @@ pub fn parse_log(
     for ethereum_type in ethereum_types {
         values.push(match ethereum_type.strip_suffix(" indexed") {
             None => {
+                let start = data_index * 32;
+                let end = start + 32;
+                let chunk = log_data
+                    .data
+                    .get(start..end)
+                    .ok_or(EthereumServiceError::EthereumParsingError)?;
                 for (i, val) in vec.iter_mut().enumerate() {
```
deuszx, line 220: I'd start by moving all the `let mut` variables from the main body of the fuction to the for loop – otherwise it looks like they need to be "one level higher" (but they don't).

Second, this logic here is unnecessary complicated IMO. What's happening is you take 32 elements from `log_data.data` and assign it to the `vec`. Why not simply copy that slice (`log_data.data[start..end]`) and assign to the new `vec` in every iteration?

Third – we have `values.push( { body })` where _body_ is a very long pattern match. Extract the pattern match, assign the result to a variable and push _that_ to `values`. The current version is harder to reason about as the order of execution does not match the order of the reading/writing.

When that's done we can discuss the correctness.

linera-ethereum/src/client.rs
MathieuDutSik: Your mention a `alloy::Filter`which is not used in the code, so I am not sure what is your point.

About the reading of events, yes Ethereum does it in one way, and `linera` when accessing the events does it another way. So?

deuszx: It's 10 lines below this one.

MathieuDutSik: This is only a specific call site.
I pointed out to you another call site in `examples/ethereum-tracker/src/service.rs` which does not use the `alloy::Filter` and therefore for which the input needs to be checked.

MathieuDutSik:
This PR did not reach consensus and is therefore closed.

PR #5938, opened 2026-04-08
Revisit bucket queue view

### LINKED ISSUE #4969: Remove redundant BucketDescription in BucketQueueView
* In `post_load`: We should only ignore `bucket_store.descriptions[0]` when `bucket_store.descriptions[0].index == 0`.
* In `pre_save`: We should not write `bucket_store.descriptions[0]` when `bucket_store.descriptions[0].index == 0`.

ma2bd:
## Motivation

* fixes the performance issue in `BucketQueueView` that forced the revert in #5096 (so the outbox can adopt it later without the regression)
* fixes #4969 

## Proposal

Replace the `Vec<BucketDescription>` metadata of `BucketQueueView` (the store struct, renamed `BucketStore` → `BucketLayout`) with three fixed-size fields (`front_position`, `num_buckets`, `first_index`). This is made possible by maintaining the invariant that all middle buckets contain exactly N elements.
    
To enforce this invariant, the back bucket is now always loaded alongside the front (via a new `KeyTag::Back` fixed key, `NUM_INIT_KEYS` = 3). During save, the back bucket is merged with new values and re-chunked into full middle buckets, ensuring only the front and back can be partial.

This PR changes `BucketQueueView` only; the chain outbox stays on `QueueView`. Actually switching the outbox over to `BucketQueueView` is left to a follow-up.

Other improvements:
- `stored_count()` is now O(1) instead of O(number of buckets)
- `back()` no longer needs to load from storage (back is always in memory)
- all bucket-index arithmetic is overflow-checked

## Test Plan

CI, including a targeted unit-test suite for `BucketQueueView`:
- save/reload roundtrip across sizes straddling N (empty, partial front, front+back, several middles)
- the "middle buckets always hold exactly N" invariant, read back from storage
- exact `stored_count()` across the partial-front / partial-back cursor cases
- per-`SaveCase` dispatch, including the combined Patch path (front promotion + surviving middles + back re-chunk in a single save)
- the N = 1 degenerate case, and `rollback` / `clear`-then-`rollback`

Benchmark results for 100 incremental saves with 10,000 existing elements (`cargo bench -p linera-views --bench queue_view -- incremental_save`):

| Queue Type | Before (from PR #5096) | After (O(1) metadata) |
|---|---|---|
| `QueueView` | 33.6 – 34.3 µs | 30.0 – 30.3 µs |
| `BucketQueueView` | 171.5 – 179.3 µs | 58.7 – 59.1 µs |

`BucketQueueView` incremental save is ~3x faster. The remaining gap vs `QueueView` is the consolidation cost (cloning + re-chunking the back bucket), which keeps buckets full to minimize point reads on SST-based stores.

## Release Plan

### DIFF
--- a/linera-chain/src/outbox.rs
+++ b/linera-chain/src/outbox.rs
@@ -8,8 +8,8 @@ use linera_base::data_types::{ArithmeticError, BlockHeight};
 #[cfg(with_testing)]
 use linera_views::context::MemoryContext;
 use linera_views::{
+    bucket_queue_view::BucketQueueView,
     context::Context,
-    queue_view::QueueView,
     register_view::RegisterView,
     views::{ClonableView, View},
     ViewError,
@@ -36,6 +36,12 @@ pub(crate) mod metrics {
     });
 }
 
+// The number of block heights in a bucket
+// The `BlockHeight` has just 8 bytes so the size is constant.
+// This means that by choosing a size of 1000, we have a
+// reasonable size that will not create any memory issues.
+const BLOCK_HEIGHT_BUCKET_SIZE: usize = 1000;
+
 /// The state of an outbox
 /// * An outbox is used to send messages to another chain.
 /// * Internally, this is implemented as a FIFO queue of (increasing) block heights.
@@ -54,7 +60,7 @@ where
     pub next_height_to_schedule: RegisterView<C, BlockHeight>,
     /// Keep sending these certified blocks of ours until they are acknowledged by
     /// receivers.
-    pub queue: QueueView<C, BlockHeight>,
+    pub queue: BucketQueueView<C, BlockHeight, BLOCK_HEIGHT_BUCKET_SIZE>,
 }
 
 impl<C> OutboxStateView<C>
@@ -107,11 +113,11 @@ where
         height: BlockHeight,
     ) -> Result<Vec<BlockHeight>, ViewError> {
         let mut updates = Vec::new();
-        while let Some(h) = self.queue.front().await? {
+        while let Some(h) = self.queue.front().copied() {
             if h > height {
                 break;
             }
-            self.queue.delete_front();
+            self.queue.delete_front().await?;
             updates.push(h);
         }
         #[cfg(with_metrics)]
--- a/linera-chain/src/outbox.rs
+++ b/linera-chain/src/outbox.rs
@@ -36,11 +36,10 @@ pub(crate) mod metrics {
     });
 }
 
-// The number of block heights in a bucket
-// The `BlockHeight` has just 8 bytes so the size is constant.
-// This means that by choosing a size of 1000, we have a
-// reasonable size that will not create any memory issues.
-const BLOCK_HEIGHT_BUCKET_SIZE: usize = 1000;
+/// The number of block heights in a bucket.
+///
+/// The current value keeps buckets comfortably under 1KB.
+const BLOCK_HEIGHT_BUCKET_SIZE: usize = 100;
 
 /// The state of an outbox
 /// * An outbox is used to send messages to another chain.
--- a/linera-chain/src/outbox.rs
+++ b/linera-chain/src/outbox.rs
@@ -8,8 +8,8 @@ use linera_base::data_types::{ArithmeticError, BlockHeight};
 #[cfg(with_testing)]
 use linera_views::context::MemoryContext;
 use linera_views::{
-    bucket_queue_view::BucketQueueView,
     context::Context,
+    queue_view::QueueView,
     register_view::RegisterView,
     views::{ClonableView, View},
     ViewError,
@@ -36,11 +36,6 @@ pub(crate) mod metrics {
     });
 }
 
-/// The number of block heights in a bucket.
-///
-/// The current value keeps buckets comfortably under 1KB.
-const BLOCK_HEIGHT_BUCKET_SIZE: usize = 100;
-
 /// The state of an outbox
 /// * An outbox is used to send messages to another chain.
 /// * Internally, this is implemented as a FIFO queue of (increasing) block heights.
@@ -59,7 +54,7 @@ where
     pub next_height_to_schedule: RegisterView<C, BlockHeight>,
     /// Keep sending these certified blocks of ours until they are acknowledged by
     /// receivers.
-    pub queue: BucketQueueView<C, BlockHeight, BLOCK_HEIGHT_BUCKET_SIZE>,
+    pub queue: QueueView<C, BlockHeight>,
 }
 
 impl<C> OutboxStateView<C>
@@ -112,11 +107,11 @@ where
         height: BlockHeight,
     ) -> Result<Vec<BlockHeight>, ViewError> {
         let mut updates = Vec::new();
-        while let Some(h) = self.queue.front().copied() {
+        while let Some(h) = self.queue.front().await? {
             if h > height {
                 break;
             }
-            self.queue.delete_front().await?;
+            self.queue.delete_front();
             updates.push(h);
         }
         #[cfg(with_metrics)]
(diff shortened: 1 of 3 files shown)


linera-chain/src/outbox.rs
```diff
@@ -36,6 +36,12 @@ pub(crate) mod metrics {
     });
 }
 
+// The number of block heights in a bucket
+// The `BlockHeight` has just 8 bytes so the size is constant.
+// This means that by choosing a size of 1000, we have a
+// reasonable size that will not create any memory issues.
+const BLOCK_HEIGHT_BUCKET_SIZE: usize = 1000;
```
afck: (I guess in practice that means we'll usually load the whole queue; there should almost never be more than 1000 entries in it?)

linera-views/src/views/bucket_queue_view.rs
```diff
@@ -238,127 +228,183 @@ where
...
-                descriptions.push(BucketDescription {
-                    index,
-                    length: u32::try_from(chunk.len()).map_err(|_| ArithmeticError::Overflow)?,
-                });
-                index = index.checked_add(1).ok_or(ArithmeticError::Overflow)?;
+            SaveCase::Rewrite => {
+                if plan.has_storage {
+                    batch.delete_key_prefix(self.context.base_key().bytes.clone());
+                }
+                let mut all_data: Vec<T> = Vec::new();
```
afck:
```suggestion
                let mut all_data = Vec::new();
```

linera-views/src/views/bucket_queue_view.rs
```diff
@@ -238,127 +228,183 @@ where
...
+                let num_buckets = if self.new_back_values.is_empty() {
+                    remaining_count
+                } else {
+                    // Merge old back + new values, re-chunk into full-N middles + new back.
+                    let State::Loaded { data: back_data } =
+                        &self.stored_buckets.back().unwrap().state
+                    else {
+                        unreachable!("back bucket is always loaded");
+                    };
+                    let mut merged: Vec<T> = back_data.clone();
```
afck:
```suggestion
                    let mut merged = back_data.clone();
```

linera-views/src/views/bucket_queue_view.rs
```diff
@@ -238,127 +228,183 @@ where
...
+                    merged.extend(self.new_back_values.iter().cloned());
+                    let chunks: Vec<&[T]> = merged.chunks(N).collect();
```
afck:
```suggestion
                    let chunks = merged.chunks(N).collect::<Vec<_>>();
```

linera-views/src/views/bucket_queue_view.rs
```diff
@@ -384,18 +430,132 @@ where
...
+        remaining_count: u32,
+    },
+}
+
+#[derive(Debug)]
+struct SavePlan {
+    case: SaveCase,
+    /// Offset within `stored_buckets` of the first surviving bucket.
+    remaining_start: usize,
+    /// Position within the front bucket of the cursor.
```
afck: (I'd have said "Position of the cursor in the front bucket".)

linera-views/src/views/bucket_queue_view.rs
```diff
@@ -238,127 +228,183 @@ where
...
-                            index,
+            SaveCase::MetadataOnly => {
+                self.cursor = Some(Cursor {
+                    offset: 0,
+                    position: plan.cursor_position,
+                });
+                self.stored_front_position = plan.cursor_position_u32;
+            }
+            SaveCase::Rewrite => {
+                let mut all_data: Vec<T> = Vec::new();
```
afck:
```suggestion
                let mut all_data = Vec::new();
```

linera-views/src/views/bucket_queue_view.rs
```diff
@@ -238,127 +228,183 @@ where
...
+                }
+                if !self.new_back_values.is_empty() {
+                    let back = self.stored_buckets.pop_back().unwrap();
+                    let State::Loaded { data: back_data } = back.state else {
+                        unreachable!("back bucket is always loaded");
+                    };
+                    let first_index = self.stored_buckets[0].index;
+                    let old_remaining_without_back =
+                        u32::try_from(self.stored_buckets.len()).expect("verified in pre_save");
+                    let mut merged: Vec<T> = back_data;
```
afck:
```suggestion
                    let mut merged = back_data;
```

linera-views/src/views/bucket_queue_view.rs
```diff
@@ -384,18 +430,132 @@ where
...
+    fn write_chunks(
+        &self,
+        batch: &mut Batch,
+        data: &[T],
+        first_index: u32,
+    ) -> Result<u32, ViewError>
+    where
+        T: Serialize + Clone,
+    {
+        let chunks: Vec<&[T]> = data.chunks(N).collect();
```
afck:
```suggestion
        let chunks = data.chunks(N).collect::<Vec<_>>();
```

linera-views/src/views/bucket_queue_view.rs
```diff
@@ -611,26 +783,11 @@ impl<C: Context, T: DeserializeOwned + Clone, const N: usize> BucketQueueView<C,
...
-                self.stored_buckets
-                    .back_mut()
-                    .expect("stored_buckets is non-empty since we just accessed its back element")
-                    .state = State::Loaded { data };
-                Ok(Some(result))
-            }
-        }
+        // Back bucket is always loaded (invariant).
+        let State::Loaded { data } = &bucket.state else {
+            unreachable!("back bucket should always be loaded");
```
afck: If the front and back bucket were not part of the `Vec`, and separate fields instead, we could probably avoid all those `unreachable!`. (Not sure if that would complicate the code further, though.)

linera-views/src/views/bucket_queue_view.rs
```diff
@@ -906,6 +1065,201 @@ mod tests {
...
+            let context = MemoryContext::new_for_testing(());
+            let mut view = BucketQueueView::<_, u32, N>::load(context.clone()).await?;
+            for i in 0..u32::try_from(size).unwrap() {
+                view.push_back(i);
+            }
+            save(&context, &mut view).await?;
+
+            let reloaded = BucketQueueView::<_, u32, N>::load(context).await?;
+            let elements = reloaded.elements().await?;
+            let expected: Vec<u32> = (0..u32::try_from(size).unwrap()).collect();
```
afck:
```suggestion
            let expected = (0..u32::try_from(size).unwrap()).collect::<Vec<_>>();
```

afck (approved):
No blockers. But I'm not convinced using this for the outbox will really perform better than a regular `QueueView`.

ma2bd:
> No blockers. But I'm not convinced using this for the outbox will really perform better than a regular `QueueView`.

How so? We could also use smaller buckets for the outboxes

afck:
I mean, each time we append a height to the outbox queue, we now load the whole bucket, append the height, and save the whole bucket. That's more data… but probably still so little that it doesn't matter.

And maybe together with the cross-chain batching it actually _is_ an improvement over the regular queue view. :thinking:

linera-chain/src/outbox.rs
ma2bd: changed to 100

afck (approved):

afck (approved):

linera-views/src/views/bucket_queue_view.rs
ma2bd: Good question. The back bucket would have to be an option anyway

ma2bd:
> No blockers. But I'm not convinced using this for the outbox will really perform better than a regular `QueueView`.

This PR no longer changes the outbox format.

PR #5947, opened 2026-04-08
Batch cross-chain updates and confirmations.

afck:
## Motivation

In principle, cross-chain updates from different sender chains and cross-chain confirmations from different receivers could all be processed in parallel.

## Proposal

As a first step, batch them and make only one `save()` call at the end of each batch. This should reduce the number of sequential write locks and sequential DB write batches.

Later we could further optimize this and also do the read operations concurrently.

## Test Plan

CI should catch regressions; ideally we should measure whether this improves performance.

## Release Plan

- Release SDK.
- Hotfix validators.

## Links

- Earlier attempt: https://github.com/linera-io/linera-protocol/pull/5024


linera-core/src/worker.rs
```diff
@@ -570,6 +655,8 @@ pub struct WorkerState<StorageClient: Storage> {
     /// manages its own lifetime via a keep-alive task. A background sweep
     /// periodically removes dead entries.
     chain_workers: ChainWorkerMap<StorageClient>,
+    /// Per-chain batch processing state for cross-chain requests.
+    chain_batches: ChainBatchMap,
```
ma2bd, line 701: What prevented chain batches from being managed by their respective chain workers?

afck: We collect the requests while we're waiting to acquire the write lock for the chain worker state, so at least the sender needs to be outside the worker state.

And after putting the request in the channel, we wait for the return value and at the same time drive the shared future that is trying to acquire the write lock and process the batch; so that shared future needs to be accessible without already having the lock, too.

linera-core/src/worker.rs
```diff
@@ -521,6 +527,85 @@ type ChainWorkerFuture<S> = Shared<oneshot::Receiver<ChainWorkerWeak<S>>>;
...
+/// The inner future type for cross-chain batch processing.
+///
+/// Wrapped in `Shared<BatchFuture>` so that all tasks waiting for
+/// cross-chain operations on the same chain can cooperatively poll a
+/// single driver. The driver loops: drain the request channel, acquire the
+/// write lock, process all requests in one batch, repeat.
+#[cfg(not(web))]
+type BatchFuture = pin::Pin<Box<dyn Future<Output = ()> + Send>>;
+#[cfg(web)]
+type BatchFuture = pin::Pin<Box<dyn Future<Output = ()>>>;
```
afck, line 588: @Twey: I think you mentioned a better way to do this at some point? Sorry I forgot the details!

Also below: Not sure if I can get away without `#[allow(clippy::arc_with_non_send_sync)]`?

linera-core/src/worker.rs
```diff
@@ -521,6 +527,85 @@ type ChainWorkerFuture<S> = Shared<oneshot::Receiver<ChainWorkerWeak<S>>>;
...
+    where
+        StorageClient: Storage + Clone + 'static,
+    {
+        let (sender, mut receiver) = mpsc::unbounded_channel();
+        let future: BatchFuture = Box::pin(async move {
+            while let Some(first) = receiver.recv().await {
+                let mut requests = vec![first];
+                while let Ok(req) = receiver.try_recv() {
+                    requests.push(req);
+                }
```
bart-linera, line 578: I wonder if adding a metric here so that we'll be able to see if we actually get batches of more than one request at a time would be useful?

bart-linera (approved):
LGTM

linera-core/src/worker.rs
afck: Done in https://github.com/linera-io/linera-protocol/pull/5947/commits/8e725d1f4241d74ce8c1aed0a9b9a2d68fdb5657.

linera-core/src/chain_worker/handle.rs
```diff
@@ -150,10 +151,11 @@ pub(crate) fn create_chain_worker<S: Storage + Clone + 'static>(
 /// Acquires a read lock, updating the last-access timestamp.
 pub(crate) async fn read_lock<S: Storage + Clone + 'static>(
     state: &Arc<RwLock<ChainWorkerState<S>>>,
-) -> OwnedRwLockReadGuard<ChainWorkerState<S>> {
+) -> Result<OwnedRwLockReadGuard<ChainWorkerState<S>>, WorkerError> {
     let guard = state.clone().read_owned().await;
+    guard.check_not_poisoned()?;
```
ma2bd, line 156: why do we need this here suddenly?

linera-core/src/chain_worker/state.rs
```diff
@@ -2176,6 +2269,14 @@ where
     }
 }
 
+/// Sends a result through a oneshot channel, logging at `debug` level if the
+/// receiver has been dropped.
+pub(crate) fn send_result<T>(sender: oneshot::Sender<T>, value: T) {
+    if sender.send(value).is_err() {
+        tracing::debug!("cannot send cross-chain result; receiver dropped");
```
ma2bd, line 2275: nit: We should probably print the actual error instead of (unnecessarily) assuming that we know what it is.

linera-core/src/worker.rs
```diff
@@ -548,6 +562,87 @@ type ChainWorkerFuture<S> = Shared<oneshot::Receiver<ChainWorkerWeak<S>>>;
...
+/// cross-chain operations on the same chain can cooperatively poll a
+/// single driver. The driver loops: drain the request channel, acquire the
+/// write lock, process all requests in one batch, repeat.
+#[cfg(not(web))]
+type BatchFuture = pin::Pin<Box<dyn Future<Output = ()> + Send>>;
+#[cfg(web)]
+type BatchFuture = pin::Pin<Box<dyn Future<Output = ()>>>;
+
+#[derive(Clone)]
+struct DriverState {
```
ma2bd: What do we mean by "driver"? and "state"?

linera-core/src/worker.rs
```diff
@@ -548,6 +562,87 @@ type ChainWorkerFuture<S> = Shared<oneshot::Receiver<ChainWorkerWeak<S>>>;
...
+    /// Send half of the channel whose receiver lives inside the driver future.
```
ma2bd: "sender half" ?

linera-core/src/worker.rs
```diff
@@ -785,6 +883,11 @@ where
             chain_modes,
             delivery_notifiers: Arc::default(),
             chain_workers,
+            // On wasm, `DriverState` is not `Send`/`Sync` (the inner future
+            // isn't `Send`), but `Arc` is still correct: `WorkerState` clones
+            // share this map and wasm is single-threaded.
+            #[allow(clippy::arc_with_non_send_sync)]
```
ma2bd: nit: perhaps also possible `#[cfg_attr(web, expect(...))]`?

linera-core/src/worker.rs
Twey: It's just to not box the futures. Some alternative patterns: https://twey.io/rust/boxed-futures/

linera-core/src/chain_worker/handle.rs
ma2bd: mmm ok it was in code already

linera-core/src/worker.rs
```diff
@@ -548,6 +562,87 @@ type ChainWorkerFuture<S> = Shared<oneshot::Receiver<ChainWorkerWeak<S>>>;
...
+    sender: mpsc::UnboundedSender<BatchRequest>,
+    /// Weak handle to the shared driver future. Upgraded to `Shared<BatchFuture>`
+    /// by callers who need to poll it.
+    future: WeakShared<BatchFuture>,
+}
+
+impl DriverState {
+    fn new<StorageClient>(
+        state: ChainWorkerArc<StorageClient>,
+    ) -> (DriverState, Shared<BatchFuture>)
```
ma2bd: nit: maybe `create` is more appropriate given that the return type is non-standard.

linera-core/src/worker.rs
ma2bd: Might this be a `Chain(Batch)RequestProcessor` ?

linera-core/src/worker.rs
```diff
@@ -548,6 +562,87 @@ type ChainWorkerFuture<S> = Shared<oneshot::Receiver<ChainWorkerWeak<S>>>;
...
+    where
+        StorageClient: Storage + Clone + 'static,
+    {
+        let (sender, mut receiver) = mpsc::unbounded_channel();
+        let future: BatchFuture = Box::pin(async move {
+            while let Some(first) = receiver.recv().await {
+                let mut requests = vec![first];
+                while let Ok(req) = receiver.try_recv() {
+                    requests.push(req);
+                }
```
ma2bd: Would the `try_recv` loop be more effective _after_ we acquired the `write_lock`?

ma2bd: Claude seems to like the idea of draining the queue both before and after the lock.

```
  You can drop the pre-lock drain entirely (the post-lock one strictly subsumes it) or keep both — the only reason to keep the pre-lock drain is if it lets you skip taking the lock at all in a future optimization where the batch can sometimes be a     
  no-op. With current code there's no such early-out, so a single post-lock drain is enough.
       
  Two small things to note while making the change:                                                                                                                                                                                                         
  - Move the CROSS_CHAIN_BATCH_SIZE observation past the second drain so the histogram reflects what was actually processed.
  - A bursty producer could in principle keep try_recv succeeding indefinitely after we hold the lock; if that's a concern, cap the drain (e.g., for _ in 0..MAX_BATCH) — but the histogram you just added is the right tool to decide whether a cap is     
  needed before adding one.  
```

linera-core/src/worker.rs
afck: I can't figure out how to apply this here.
If I don't box the future, what _is_ the type of the `future` field in `DriverState`? (Renaming that to `ChainBatchRequestProcessor`…)

linera-core/src/chain_worker/state.rs
afck: The "error type" here is just the value we were trying to send. The only error case according to the documentation is that the receiver has been dropped.

linera-core/src/worker.rs
```diff
@@ -548,6 +562,93 @@ type ChainWorkerFuture<S> = Shared<oneshot::Receiver<ChainWorkerWeak<S>>>;
 /// Callers that find a pending entry clone the `Shared` future and await it.
 type ChainWorkerMap<S> = Arc<papaya::HashMap<ChainId, ChainWorkerFuture<S>>>;
 
+/// A cross-chain request waiting to be processed in a batch.
+pub(crate) enum BatchRequest {
```
ma2bd, line 565: nit: Also `BatchRequest` is a little weird. Maybe `Batched(CrossChain)Request`

ma2bd: (but then `ChainBatchRequestProcessor` would also be renamed again, so maybe not)

linera-core/src/worker.rs
```diff
@@ -548,6 +562,93 @@ type ChainWorkerFuture<S> = Shared<oneshot::Receiver<ChainWorkerWeak<S>>>;
...
+    {
+        let (sender, mut receiver) = mpsc::unbounded_channel();
+        let future: BatchFuture = Box::pin(async move {
+            while let Some(first) = receiver.recv().await {
+                let mut requests = vec![first];
+                match handle::write_lock(&state).await {
+                    Ok(mut guard) => {
+                        while requests.len() < batch_size_limit {
+                            match receiver.try_recv() {
+                                Ok(req) => requests.push(req),
```
ma2bd: nit: request

linera-core/src/worker.rs
```diff
@@ -548,6 +562,93 @@ type ChainWorkerFuture<S> = Shared<oneshot::Receiver<ChainWorkerWeak<S>>>;
 /// Callers that find a pending entry clone the `Shared` future and await it.
 type ChainWorkerMap<S> = Arc<papaya::HashMap<ChainId, ChainWorkerFuture<S>>>;
 
+/// A cross-chain request waiting to be processed in a batch.
+pub(crate) enum BatchRequest {
+    Update {
+        origin: ChainId,
+        bundles: Vec<(Epoch, MessageBundle)>,
+        previous_height: Option<BlockHeight>,
+        result_tx: oneshot::Sender<Result<CrossChainUpdateResult, WorkerError>>,
```
ma2bd: `result_tx` -> `result_sender`?

linera-core/src/worker.rs
afck: * Maybe `RequestsBatch` and `RequestsBatchProcessor`?
* Or `CrossChainBatch` and `CrossChainBatchProcessor`?

linera-core/src/worker.rs
Twey: You can make it a type parameter: `struct DriverState<F> { future: WeakShared<F> }` and on the `impl`s add a trait bound `F: Future<Output = ()>` (covers both `Send` and non-`Send` futures, which `Send`ness will propagate to the container).  For the return type of `DriverState::new` you can use RPIT: `-> DriverState<impl Future<Output = ()>>`, though it unfortunately propagates up to the `WorkerState`.  To bound the propagation you can use a trait:

```rust
trait WorkerState {
  // … worker state functions …
}

impl WorkerState for WorkerStateImpl<St: Storage, F: Future<Output = ()>> {
  // … implementations …
}
```

and use an implementation of that trait instead of the concrete `WorkerState` type.

(I wouldn't block the PR on this though!)

afck: That would be an enormous amount of additional code! I think in this case the boxing is actually preferable?

afck: Also, how would I then define e.g. the worker state field in `LocalNode`, or `GrpcServer`…? Would those also have to be generic?

Twey: They're already generic over the storage — it would just involve making them generic over the worker state instead (which bound would include the storage, as well as the internal future type).  In fact, you could probably save on written-down trait bounds overall.  Having to duplicate the signatures into a trait is annoying, but pretty mechanical and limited to this implementation.  (Plus, if you can figure out a minimal set, you can write the others as defaults or extensions.)

afck: But it'll still add probably ~100 lines of code (the trait would have lots of methods, and I don't think many of them could have a default implementation), and the benefit seems small to me?

Twey: The benefit is small because we're already paying the (heavy) cost throughout our codebase:

```
$ rg Send | wc -l
861
```

The proliferation of `Send` bounds (or worse, conditionally-compiled `Send` bounds, which we also have) is a consequence of conditionally-compiling boxed traits, mostly in `linera-execution`.  If we avoid that there should be no need to mention it explicitly at all — since all our types are concrete by the time it matters (`tokio::spawn` in `linera-service`) it should always be inferrable by the compiler.

afck: Right, good point. But maybe that should be addressed everywhere at once then, in a separate PR?

Twey: Unfortunately that PR would be quite involved: there are a fair few things blocking the removal of the trait objects in `linera-execution`.  For now I've just been trying to avoid introducing _more_ cases that will get in the way if we try to fix that.  They become structural pretty quickly!

But as I said,
> (I wouldn't block the PR on this though!)

PR #5954, opened 2026-04-08
Pipeline certificate batch downloads during chain sync

ndr-ds:
## Motivation

Chain synchronization downloads certificate batches sequentially: download batch →
process batch → download next batch. The CPU is idle during downloads and the network is
idle during processing. For a chain with ~13,000 blocks, this results in ~70-89
blocks/sec with each batch cycle taking progressively longer (2.7s → 22s).

## Proposal

Replace the sequential download loop in `download_certificates_from` with a pipelined
sliding window using `FuturesOrdered`. Up to `max_concurrent_batch_downloads` (default: 1) batches are downloaded concurrently, and completed batches are processed sequentially
in order as they arrive.

The number of network requests is unchanged — same batch count as before, just
overlapped with processing instead of serialized.

A new CLI option `--max-concurrent-batch-downloads` controls the concurrency level.

The `RequestsScheduler` is now wrapped in `Arc` to allow sharing across download futures
without requiring `Env: Clone`.

**Future follow-up (needs validator deploy):** Expose a streaming gRPC endpoint where
the client requests a height range and the validator streams blocks back in order. This
would reduce N batch requests to a single streaming request.

## Test Plan

CI.

## Release Plan

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -621,7 +621,7 @@ impl<Env: Environment> Client<Env> {
         let scheduler = self.requests_scheduler.clone();
         let remote = remote_node.clone();
 
-        let _download_task = linera_base::Task::spawn(async move {
+        let download_task = linera_base::Task::spawn(async move {
             let mut download_height = next_height;
             let mut in_flight = FuturesOrdered::<CertificateBatchFuture>::new();
 
@@ -669,6 +669,9 @@ impl<Env: Environment> Client<Env> {
             next_height = info.next_block_height;
             last_info = info;
         }
+        // Await the downloader so any panic inside the spawned task surfaces here
+        // instead of being silently swallowed when the channel closes.
+        download_task.await;
         Ok(last_info)
     }
 
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -606,6 +606,15 @@ impl<Env: Environment> Client<Env> {
         // Download remaining certificates from the remote node using a pipelined
         // sliding window. A background task downloads up to `max_concurrent_batch_downloads`
         // batches concurrently and sends them through a channel for sequential processing.
+        #[cfg(not(web))]
+        type CertificateBatchFuture = std::pin::Pin<
+            Box<dyn Future<Output = Result<Vec<ConfirmedBlockCertificate>, NodeError>> + Send>,
+        >;
+        #[cfg(web)]
+        type CertificateBatchFuture = std::pin::Pin<
+            Box<dyn Future<Output = Result<Vec<ConfirmedBlockCertificate>, NodeError>>>,
+        >;
+
         let max_concurrent = self.options.max_concurrent_batch_downloads;
         let batch_size = self.options.certificate_download_batch_size;
         let (sender, mut receiver) = tokio::sync::mpsc::channel(max_concurrent);
@@ -614,43 +623,36 @@ impl<Env: Environment> Client<Env> {
 
         let _download_task = linera_base::Task::spawn(async move {
             let mut download_height = next_height;
-            let mut in_flight = FuturesOrdered::new();
+            let mut in_flight = FuturesOrdered::<CertificateBatchFuture>::new();
 
-            let enqueue_batch = |in_flight: &mut FuturesOrdered<_>,
-                                 scheduler: &Arc<RequestsScheduler<Env>>,
-                                 remote: &RemoteNode<Env::ValidatorNode>,
-                                 height: BlockHeight,
-                                 limit: u64| {
+            let try_enqueue = |in_flight: &mut FuturesOrdered<CertificateBatchFuture>,
+                               download_height: &mut BlockHeight| {
+                if *download_height >= stop {
+                    return;
+                }
+                let limit = u64::from(stop)
+                    .saturating_sub(u64::from(*download_height))
+                    .min(batch_size);
+                let height = *download_height;
                 let scheduler = scheduler.clone();
                 let remote = remote.clone();
-                in_flight.push_back(async move {
+                in_flight.push_back(Box::pin(async move {
                     scheduler
                         .download_certificates(&remote, chain_id, height, limit)
                         .await
-                });
+                }));
+                *download_height = BlockHeight(u64::from(*download_height) + limit);
             };
 
-            // Fill initial window.
             while in_flight.len() < max_concurrent && download_height < stop {
-                let limit = u64::from(stop)
-                    .saturating_sub(u64::from(download_height))
-                    .min(batch_size);
-                enqueue_batch(&mut in_flight, &scheduler, &remote, download_height, limit);
-                download_height = BlockHeight(u64::from(download_height) + limit);
+                try_enqueue(&mut in_flight, &mut download_height);
             }
 
-            // Drain completed batches and refill the window.
             while let Some(result) = in_flight.next().await {
                 if sender.send(result).await.is_err() {
                     break;
                 }
-                if download_height < stop {
-                    let limit = u64::from(stop)
-                        .saturating_sub(u64::from(download_height))
-                        .min(batch_size);
-                    enqueue_batch(&mut in_flight, &scheduler, &remote, download_height, limit);
-                    download_height = BlockHeight(u64::from(download_height) + limit);
-                }
+                try_enqueue(&mut in_flight, &mut download_height);
             }
         });
 
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -12,7 +12,7 @@ use std::{
 use custom_debug_derive::Debug;
 use futures::{
     future::Future,
-    stream::{self, AbortHandle, FuturesUnordered, StreamExt},
+    stream::{self, AbortHandle, FuturesOrdered, FuturesUnordered, StreamExt},
 };
 #[cfg(with_metrics)]
 use linera_base::prometheus_util::MeasureLatency as _;
@@ -128,6 +128,7 @@ pub static DEFAULT_CERTIFICATE_DOWNLOAD_BATCH_SIZE: u64 = 500;
 pub static DEFAULT_CERTIFICATE_UPLOAD_BATCH_SIZE: u64 = 500;
 pub static DEFAULT_SENDER_CERTIFICATE_DOWNLOAD_BATCH_SIZE: usize = 20_000;
 pub static DEFAULT_MAX_EVENT_STREAM_QUERIES: usize = 1000;
+pub static DEFAULT_MAX_CONCURRENT_BATCH_DOWNLOADS: usize = 1;
 
 #[derive(Debug, Clone, Copy)]
 pub enum TimingType {
@@ -249,7 +250,7 @@ pub struct Client<Env: Environment> {
     /// tracking.
     pub local_node: LocalNodeClient<Env::Storage>,
     /// Manages the requests sent to validator nodes.
-    requests_scheduler: RequestsScheduler<Env>,
+    requests_scheduler: Arc<RequestsScheduler<Env>>,
     /// The admin chain ID.
     admin_chain_id: ChainId,
     /// Chains that should be tracked by the client, along with their listening mode.
@@ -300,7 +301,8 @@ impl<Env: Environment> Client<Env> {
             Some(chain_modes.clone()),
         );
         let local_node = LocalNodeClient::new(state);
-        let requests_scheduler = RequestsScheduler::new(vec![], requests_scheduler_config);
+        let requests_scheduler =
+            Arc::new(RequestsScheduler::new(vec![], requests_scheduler_config));
 
         Self {
             environment,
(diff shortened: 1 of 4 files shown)


ndr-ds:
* **#5954** <a href="https://app.graphite.com/github/pr/linera-io/linera-protocol/5954?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/linera-io/linera-protocol/5954?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

linera-core/src/client/mod.rs
```diff
@@ -125,6 +125,7 @@ mod metrics {
 pub static DEFAULT_CERTIFICATE_DOWNLOAD_BATCH_SIZE: u64 = 500;
 pub static DEFAULT_CERTIFICATE_UPLOAD_BATCH_SIZE: u64 = 500;
 pub static DEFAULT_SENDER_CERTIFICATE_DOWNLOAD_BATCH_SIZE: usize = 20_000;
+pub static DEFAULT_MAX_CONCURRENT_BATCH_DOWNLOADS: usize = 5;
```
ma2bd: Can we start with a very conservative value (`1` ?) so that we don't accidentally roll out expensive experimental features to the web clients? Alternatively `web/@linera/client` could have a different default.

ndr-ds: True, makes total sense, will decrease

linera-core/src/client/mod.rs
```diff
@@ -558,18 +560,47 @@ impl<Env: Environment> Client<Env> {
...
+        type CertificateBatchFuture = std::pin::Pin<
+            Box<dyn Future<Output = Result<Vec<ConfirmedBlockCertificate>, NodeError>> + Send>,
+        >;
+        #[cfg(web)]
+        type CertificateBatchFuture = std::pin::Pin<
+            Box<dyn Future<Output = Result<Vec<ConfirmedBlockCertificate>, NodeError>>>,
+        >;
+
+        let mut download_height = next_height;
+        let mut futures: FuturesOrdered<CertificateBatchFuture> = FuturesOrdered::new();
```
afck: (I think this style is preferred?)

```suggestion
        let mut futures = FuturesOrdered::<CertificateBatchFuture>::new();
```

linera-core/src/client/mod.rs
```diff
@@ -579,6 +610,22 @@ impl<Env: Environment> Client<Env> {
...
+                    .saturating_sub(u64::from(download_height))
+                    .min(batch_size);
+                let height = download_height;
+                let scheduler = self.requests_scheduler.clone();
+                let remote = remote_node.clone();
+                futures.push_back(Box::pin(async move {
+                    scheduler
+                        .download_certificates(&remote, chain_id, height, limit)
+                        .await
+                }));
```
afck: Can this duplication be avoided by using e.g. `StreamExt::for_each_concurrent`?

linera-core/src/client/mod.rs
```diff
@@ -558,18 +560,47 @@ impl<Env: Environment> Client<Env> {
...
 
-            let certificates = self
-                .requests_scheduler
-                .download_certificates(remote_node, chain_id, next_height, limit)
-                .await?;
+        // Process completed batches and refill the download window.
+        while let Some(result) = futures.next().await {
+            let certificates = result?;
             let Some(info) = self
                 .process_certificates(slice::from_ref(remote_node), certificates)
```
afck: Doesn't that mean we're still not doing downloading and processing concurrently? While we're running `process_certificates`, nobody is polling `futures.next()` now.

deuszx: Yes, it definitely is not running in parallel.

ndr-ds: The original implementation used `tokio::spawn`, I think Claude removed it and I didn't see it 🤦🏻‍♂️  
Good catch

linera-core/src/client/mod.rs
```diff
@@ -599,18 +598,67 @@ impl<Env: Environment> Client<Env> {
...
+        // sliding window. A background task downloads up to `max_concurrent_batch_downloads`
+        // batches concurrently and sends them through a channel for sequential processing.
+        #[cfg(not(web))]
+        type CertificateBatchFuture = std::pin::Pin<
+            Box<dyn Future<Output = Result<Vec<ConfirmedBlockCertificate>, NodeError>> + Send>,
+        >;
+        #[cfg(web)]
+        type CertificateBatchFuture = std::pin::Pin<
+            Box<dyn Future<Output = Result<Vec<ConfirmedBlockCertificate>, NodeError>>>,
+        >;
```
afck, line 616: Does this work, too?

```suggestion
        type CertificateBatchFuture = std::pin::Pin<
            Box<dyn Future<Output = Result<Vec<ConfirmedBlockCertificate>, NodeError>> + MaybeSend>,
        >;
```

linera-core/src/client/mod.rs
```diff
@@ -599,18 +598,67 @@ impl<Env: Environment> Client<Env> {
...
+
+        let max_concurrent = self.options.max_concurrent_batch_downloads;
+        let batch_size = self.options.certificate_download_batch_size;
+        let (sender, mut receiver) = tokio::sync::mpsc::channel(max_concurrent);
+        let scheduler = self.requests_scheduler.clone();
+        let remote = remote_node.clone();
+
+        let _download_task = linera_base::Task::spawn(async move {
```
deuszx: Because we spawn and drop the handle, if `download_certificates` panics inside the spawned task, the panic is swallowed; the receiver just sees the channel close and the outer loop exits as if sync completed normally.

deuszx: Also, if the outer `download_certificates_from` future is cancelled (caller drops it), the spawned task keeps running.

linera-core/src/client/mod.rs
```diff
@@ -131,6 +128,7 @@ pub static DEFAULT_CERTIFICATE_DOWNLOAD_BATCH_SIZE: u64 = 500;
 pub static DEFAULT_CERTIFICATE_UPLOAD_BATCH_SIZE: u64 = 500;
 pub static DEFAULT_SENDER_CERTIFICATE_DOWNLOAD_BATCH_SIZE: usize = 20_000;
 pub static DEFAULT_MAX_EVENT_STREAM_QUERIES: usize = 1000;
+pub static DEFAULT_MAX_CONCURRENT_BATCH_DOWNLOADS: usize = 1;
```
deuszx, line 131: PR descriptions states the default is 5.

linera-core/src/client/mod.rs
ndr-ds: I don't think the task will keep running 🤔 AFAIU `linera_base::Task` will make sure it gets aborted.
I've addressed the other comment

linera-core/src/client/mod.rs
```diff
@@ -621,6 +669,9 @@ impl<Env: Environment> Client<Env> {
             next_height = info.next_block_height;
             last_info = info;
         }
+        // Await the downloader so any panic inside the spawned task surfaces here
+        // instead of being silently swallowed when the channel closes.
+        download_task.await;
```
deuszx, line 674: What about this variant, that is more "structurred concurrency"-ish
```rust
// Build the sequence of (height, limit) batches lazily.
let batches = std::iter::from_fn(move || {
    if next_height >= stop {
        return None;
    }
    let limit = u64::from(stop)
        .saturating_sub(u64::from(next_height))
        .min(batch_size);
    let height = next_height;
    next_height = BlockHeight(u64::from(next_height) + limit);
    Some((height, limit))
});

// Map each batch to a download future, run up to `max_concurrent` at once,
// preserving order.
let mut downloads = stream::iter(batches)
    .map(|(height, limit)| {
        let scheduler = scheduler.clone();
        let remote = remote.clone();
        async move {
            scheduler
                .download_certificates(&remote, chain_id, height, limit)
                .await
        }
    })
    .buffered(max_concurrent);

// Process in order; downloads keep flowing while we process.
while let Some(certificates) = downloads.try_next().await? {
    let Some(info) = self
        .process_certificates(slice::from_ref(remote_node), certificates, until_block_time)
        .await?
    else {
        break;
    };
    last_info = info;
}
```

Because we have `.buffered(max_concurrent);` at every `downloads.try_next().await` we will be polling whole batch. Seems simpler: no spawn, no `FuturesUnordered`, no type alias.

linera-core/src/client/mod.rs
ndr-ds: Doesn't seem like this compiles :(

linera-core/src/client/mod.rs
ndr-ds: The buffered version only polls futures when we call `next`. `Task::spawn` polls continuously, so it's faster. Claude tested it and saw that the version with `spawn` is 15-20% faster

deuszx: Were you using the default value for `MAX_CONCURRENT*` ? It's set to 1.

deuszx: ping here @ndr-ds

ndr-ds: Hey, sorry, forgot to reply to this one. So if max_concurrent < n_batches, spawn is always faster. If max_concurrent >= n_batches, the buffered approach ties with spawn. So then we need to figure out if simplifying the code would be worth that perf regression in certain cases. With max_concurrent being one for example, the spawned version can be almost 30% faster.

deuszx: > So if max_concurrent < n_batches, spawn is always faster

I don't think that's the reason spawn is gonna be faster. IIUC, spawn is faster only if there's a CPU-intensive work done (in the spawned task) after it's downloaded (like deserializing, hashing). Or differently: if `CPU_time(processing_step) > ( CPU_time(downloading_step) + Download) )`. If the "fetching (tokio) task" is always finishing before the processing step then yes – spawning should be faster.

ndr-ds: No, it's about polling timing. Both approaches start `max_concurrent` downloads concurrently upfront, that part is the same. The difference shows up as batches complete and new ones are enqueued, i.e. when `max_concurrent < n_batches`.
With buffered, after `try_next()` returns a batch, it pulls the next future from the stream and enqueues it, but doesn't poll it. The newly-pushed future never gets polled until we call `try_next()` again. It sits unpolled while `process_certificates` runs on the same tokio task, and only gets its first poll on the next `try_next().await`, after processing finishes. So each new download is delayed by the full processing time of the batch of certificates, whatever that is.
When `max_concurrent >= n_batches`, all futures are enqueued upfront and polled before any processing starts, so the delay never occurs and both approaches are equivalent, which is exactly what the benchmark shows.

With spawn, the downloader is a separate tokio task whose loop only blocks on `in_flight.next()`. After a batch is sent through the channel, it immediately loops back and polls the newly-pushed future, so the next download starts right after the previous one completes, regardless of how long processing takes.

Also, the benchmark claude wrote for this uses pure `tokio::sleep` with no CPU work and the gap is still there whenever `max_concurrent < n_batches`, which rules out CPU parallelism as the cause.

deuszx: > . So each new download is delayed by the full processing time of the batch of certificates, whatever that is.

Yes, that's exactly what I wrote earlier: if processing time is < download (+deserialization) time then we will be `try_next().await` (polling) more frequently than we can download and hence there should be no difference.

> Also, the benchmark claude wrote for this uses pure tokio::sleep with no CPU work

Then I'd say this benchmark is wrong b/c there is CPU time here.

ndr-ds: I modified the benchmark to vary the processing/download ratio and also added CPU work to the download step to simulate deserialization/hashing. Here are the results with n_batches=20:
```
  download=100ms process=200ms    conc=2  spawn=4159ms  buffered=5082ms  diff=18.2%
  download=100ms process=100ms    conc=2  spawn=2158ms  buffered=3067ms  diff=29.6%
  download=100ms process=50ms     conc=2  spawn=1143ms  buffered=2072ms  diff=44.8%
  download=100ms process=10ms     conc=2  spawn=1051ms  buffered=1261ms  diff=16.7%

  download=100ms+50ms_cpu  process=200ms  conc=2  spawn=4253ms  buffered=6070ms  diff=29.9%
  download=100ms+100ms_cpu process=200ms  conc=2  spawn=4341ms  buffered=7061ms  diff=38.5%
```

The gap doesn't disappear when processing < download, it actually gets larger. With process=50ms the difference is 44.8%. Adding CPU work to the download step also makes spawn more favorable, not less, since the downloader runs on a separate task and can do that CPU work in parallel with process_certificates. Both effects go in spawn's favor.

deuszx: Thanks! This makes sense now – the CPU work (deserialization) in the spawned version is done on the same spawned task while in the bufferred version it happens on the task that is doing the polling so the delay is ever bigger (between consectuive polls for download) since that blocks the task/thread.

ndr-ds: Yup! Exactly :)

deuszx (approved):

PR #5955, opened 2026-04-09
[testnet] Compress block heights as ranges in certificate download debug logs

ndr-ds:
## Motivation

Backport of #5953.

When syncing a chain, each certificate batch download logs a
`CertificatesByHeightRequest` via the `client_delegate!` debug macro. With batch size
500, this produces ~15KB log lines listing every individual `BlockHeight(N)`:

```
request=CertificatesByHeightRequest { chain_id: 01fc3cf7..., heights:
[BlockHeight(14810), BlockHeight(14811), BlockHeight(14812), ... BlockHeight(15309)] }
```

The same problem affects `RequestKey::Certificates` in scheduler trace logs and
`ChainInfoQuery::request_sent_certificate_hashes_by_heights` in chain info query debug
logs.

## Proposal

Add a `CompressedHeights` formatter that displays contiguous block heights as ranges.
The same information, just readable:

```
request=CertificatesByHeightRequest { chain_id: 01fc3cf7..., heights: [14810..15309] }
```

Gaps are preserved: `[1..3, 5, 7..10]`.

Applied to all three structs that contain `Vec<BlockHeight>` fields logged at
debug/trace level:
- `CertificatesByHeightRequest` — custom `Debug` impl
- `RequestKey::Certificates` — custom `Debug` impl
- `ChainInfoQuery::request_sent_certificate_hashes_by_heights` — `#[debug(with =
"...")]`

## Test Plan

CI. Unit tests cover edge cases: empty, single element, contiguous range, gaps, and
all-isolated heights.

### DIFF
--- a/linera-core/src/client/requests_scheduler/request.rs
+++ b/linera-core/src/client/requests_scheduler/request.rs
@@ -1,24 +1,26 @@
 // Copyright (c) Zefchain Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-use std::fmt;
-
+use custom_debug_derive::Debug;
 use linera_base::{
     data_types::{Blob, BlobContent, BlockHeight},
     identifiers::{BlobId, ChainId},
 };
 use linera_chain::types::ConfirmedBlockCertificate;
 
-use crate::{client::requests_scheduler::cache::SubsumingKey, data_types::CompressedHeights};
+use crate::{
+    client::requests_scheduler::cache::SubsumingKey, data_types::debug_compressed_heights,
+};
 
 /// Unique identifier for different types of download requests.
 ///
 /// Used for request deduplication to avoid redundant downloads of the same data.
-#[derive(Clone, PartialEq, Eq, Hash)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub enum RequestKey {
     /// Download certificates by specific heights
     Certificates {
         chain_id: ChainId,
+        #[debug(with = "debug_compressed_heights")]
         heights: Vec<BlockHeight>,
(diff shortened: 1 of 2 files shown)


linera-core/src/client/requests_scheduler/request.rs
```diff
@@ -27,6 +29,27 @@ pub enum RequestKey {
     CertificateForBlob(BlobId),
 }
 
+impl fmt::Debug for RequestKey {
```
afck: We already use `custom_debug_derive` elsewhere, so this should be possible with just a `#[debug(with = …)]` annotation on the `heights` field.

linera-core/src/data_types.rs
```diff
@@ -346,12 +346,62 @@ impl ChainInfoResponse {
...
+        let heights = self.0;
+        write!(f, "[")?;
+        let mut index = 0;
+        while index < heights.len() {
+            if index > 0 {
+                write!(f, ", ")?;
+            }
+            let range_start = u64::from(heights[index]);
+            let mut range_end = range_start;
+            while index + 1 < heights.len() && u64::from(heights[index + 1]) == range_end + 1 {
```
afck: I wonder if we should guard against overflows here.

afck (commented):
Ideally we'd do this not just for logs but even for the requests themselves, i.e. have a datatype that efficiently encodes lists of strictly increasing lists of numbers that often differ by 1.

ndr-ds:
Right 🤔 I can try that on main, but it would probably not be backwards compatible for `testnet-conway`

ma2bd (approved):

PR #5956, opened 2026-04-09
Remove `ValueCache::get` and rename `get_hashed` as `get`.

MathieuDutSik:
## Motivation

Subsequent to PR #5803 we restructured the `block_cache` of `testnet_conway`.
But contrary to PR #5798, we had to keep `get` and `get_hashed` since both were used.
However, that is no longer the case after other changes.

## Proposal

Remove `get` from `value_cache` and rename `get_hashed` as `get`.
This aligns it with `main`.

## Test Plan

CI

## Release Plan

This is to `testnet_conway`.

## Links

None.

### DIFF
--- a/linera-cache/src/value_cache.rs
+++ b/linera-cache/src/value_cache.rs
@@ -51,11 +51,6 @@ where
         Self::track_cache_usage(value)
     }
 
-    /// Returns a clone of the value from the cache, if present.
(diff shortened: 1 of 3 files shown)


MathieuDutSik:
Seems a more radicla rethink is needed.

PR #5959, opened 2026-04-09
Implement the execution of messages together.

### LINKED ISSUE #5698: Use a single `UserAction` to execute an entire bundle of messages
Currently, we run `UserAction::Message` multiple times per bundle. This means every message loads and finalizes its entire stack of applications.

But bundles are transactions, so instead, we should load and finalize each application at most once per bundle.

Pros:
* Saves fuel and Wasm startup time if messages run the same application.
* Unlocks some use cases. E.g. if one message borrows from one app and a next message repays the loan.

Caveat:
* This is a breaking change, so backporting to testnet would require a feature guard (probably not worth it although I have one minor use case).

MathieuDutSik:
## Motivation

When executing the messages of a `IncomingBundle`, we have a bunch of messages to execute.
However, when doing so, if we have say 10 messages to the same application, then what happens
is that 10 times the instance is created and 10 times the `finalize()`.

Fixes #5698 

## Proposal

Call finalize only at the end of the message bundle.

## Test Plan

CI.

## Release Plan

This is a breaking change that cannot be backported to `testnet_conway`.

## Links

None.

### DIFF
--- a/linera-chain/src/block_tracker.rs
+++ b/linera-chain/src/block_tracker.rs
@@ -201,10 +201,8 @@ impl<'resources, 'blobs> BlockExecutionTracker<'resources, 'blobs> {
         C::Extra: ExecutionRuntimeContext,
     {
         Box::pin(async move {
-            // For Reject bundles or bundles with few messages, use the per-message path.
-            if incoming_bundle.action == MessageAction::Reject
-                || !Self::can_batch_bundle(incoming_bundle)
-            {
+            // For Reject bundles, use the per-message path.
+            if incoming_bundle.action == MessageAction::Reject {
                 for posted_message in incoming_bundle.messages() {
                     Box::pin(self.execute_message_in_block(
                         chain,
@@ -225,15 +223,13 @@ impl<'resources, 'blobs> BlockExecutionTracker<'resources, 'blobs> {
 
             let messages: Vec<_> = incoming_bundle.messages().collect();
 
-            // Pre-group user messages by application, preserving order.
+            // Group user messages by application, preserving order.
             let mut app_batches: BTreeMap<ApplicationId, Vec<usize>> = BTreeMap::new();
             for (i, pm) in messages.iter().enumerate() {
                 if let Message::User { application_id, .. } = &pm.message {
                     app_batches.entry(*application_id).or_default().push(i);
                 }
             }
-            // Only apps with ≥2 messages benefit from batching.
-            app_batches.retain(|_, indices| indices.len() >= 2);
 
             let mut executed_apps: BTreeSet<ApplicationId> = BTreeSet::new();
             let mut i = 0;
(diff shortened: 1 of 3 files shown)


linera-chain/src/block_tracker.rs
```diff
@@ -180,6 +182,166 @@ impl<'resources, 'blobs> BlockExecutionTracker<'resources, 'blobs> {
...
     }
 
+    /// Executes all messages in an incoming bundle. Groups consecutive zero-grant
+    /// user messages targeting the same application for batched execution
+    /// (shared runtime, single finalize).
+    ///
+    /// Returns a boxed `dyn Future` so its (deeply nested) internal type does not
+    /// inflate the `execute_transaction` future's `Send`-bound type graph, which
+    /// is walked by callers as deep as `linera-faucet-server`.
+    fn execute_incoming_bundle<'s, C>(
```
ma2bd: nit: we expect `'a` for scope variables

linera-chain/src/block_tracker.rs
```diff
@@ -180,6 +182,166 @@ impl<'resources, 'blobs> BlockExecutionTracker<'resources, 'blobs> {
...
+            }
+
+            Ok(())
+        })
+    }
+
+    /// Returns true if the bundle contains at least two consecutive zero-grant user
+    /// messages targeting the same application — the case where batched execution
+    /// avoids redundant runtime creation and finalize calls.
+    fn can_batch_bundle(incoming_bundle: &IncomingBundle) -> bool {
```
ma2bd: It shouldn't matter whether messages targeting the same application are consecutive or not

linera-chain/src/block_tracker.rs
MathieuDutSik: Yes, but `'a` is already used.

afck: It's probably most common to use `'b` then. (Examples: https://doc.rust-lang.org/std/cmp/trait.PartialEq.html)

linera-chain/src/block_tracker.rs
MathieuDutSik: Ok, this has been addressed. We use the batching systematically.

linera-chain/src/block_tracker.rs
MathieuDutSik: Yes, done.

linera-chain/src/block_tracker.rs
```diff
@@ -29,6 +33,12 @@ use crate::{
     ChainError, ChainExecutionContext, ExecutionResultExt,
 };
 
+/// A boxed future that is `Send` on non-web targets but not required to be `Send` on web/wasm.
+#[cfg(not(web))]
+type BoxedFuture<'a> = Pin<Box<dyn Future<Output = Result<(), ChainError>> + Send + 'a>>;
+#[cfg(web)]
+type BoxedFuture<'a> = Pin<Box<dyn Future<Output = Result<(), ChainError>> + 'a>>;
```
afck, line 41: @Twey: I think you mentioned a better way to do this once, but I don't remember the details?

linera-chain/src/block_tracker.rs
afck: Actually, I think in this case the explicit parameter could even be elided and you could use `'_` below?

linera-chain/src/block_tracker.rs
```diff
@@ -180,6 +182,115 @@ impl<'resources, 'blobs> BlockExecutionTracker<'resources, 'blobs> {
...
+                }
+                return Ok(());
+            }
+
+            // Accept bundle — group all user messages by application_id.
+            let chain_execution_context =
+                ChainExecutionContext::IncomingBundle(txn_tracker.transaction_index());
+            ensure!(!chain.system.closed.get(), ChainError::ClosedChain);
+
+            let messages: Vec<_> = incoming_bundle.messages().collect();
```
```suggestion
            let messages = incoming_bundle.messages().collect::<Vec<_>>();
```

linera-chain/src/block_tracker.rs
```diff
@@ -180,6 +182,115 @@ impl<'resources, 'blobs> BlockExecutionTracker<'resources, 'blobs> {
...
+
+            // Group user messages by application, preserving order.
+            let mut app_batches: BTreeMap<ApplicationId, Vec<usize>> = BTreeMap::new();
```
```suggestion
            let mut app_batches = BTreeMap::<ApplicationId, Vec<usize>>::new();
```

linera-chain/src/block_tracker.rs
```diff
@@ -180,6 +182,115 @@ impl<'resources, 'blobs> BlockExecutionTracker<'resources, 'blobs> {
...
+
+            while i < messages.len() {
+                let posted_message = messages[i];
+
+                // User messages are executed through the batched path.
+                if let Message::User { application_id, .. } = &posted_message.message {
+                    if executed_apps.insert(*application_id) {
+                        let app_id = *application_id;
+                        let indices = &app_batches[&app_id];
+                        let batch: Vec<_> = indices
```
```suggestion
                        let batch = indices
```

afck (commented):
This changes the order: messages are now not executed in the order in which they appear in the bundle. Is this really what we want?

MathieuDutSik:
The existing design is not good since the effective running of the application is difficult and limitation have to be put.

PR #5962, opened 2026-04-09
[testnet] Group operations for maybe_get_required_blobs

MathieuDutSik:
## Motivation

The pending_proposed_blobs needs to be simplified. But before that, we can restructure the access code.

Backport #4938 

## Proposal

Process the different cases by blocks so as to group the operations.

## Test Plan

CI

## Release Plan

This is the backport to `testnet_conway`.

## Links

None.

### DIFF
--- a/linera-core/src/chain_worker/state.rs
+++ b/linera-core/src/chain_worker/state.rs
@@ -2127,7 +2127,7 @@ where
     }
 }
 
-/// Returns the missing indices and corresponding blob_ids.
+/// Returns the missing indices and corresponding blob IDs.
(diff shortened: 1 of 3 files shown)


linera-core/src/chain_worker/state.rs
```diff
@@ -2107,6 +2127,19 @@ where
     }
 }
 
+/// Returns the missing indices and corresponding blob_ids.
```
afck:
```suggestion
/// Returns the missing indices and corresponding blob IDs.
```

afck (approved):

PR #5976, opened 2026-04-10
Add streaming DownloadBlobs RPC endpoint

ndr-ds:
## Motivation

During chain sync, each blob referenced by a batch of certificates is downloaded via a
separate `DownloadBlob(BlobId) -> BlobContent` unary RPC call. Batches with hundreds of
new blobs fire hundreds of sequential round trips before the certificates can be
processed, which becomes the dominant bottleneck: measured sync logs show ~63% of
wall-clock time is spent waiting for blob downloads to complete.

A batch, streaming RPC lets the validator serve multiple blobs on a single connection
and lets the caller process blobs as they arrive instead of waiting for the full batch.

## Proposal

Add a new streaming RPC endpoint on validators. No caller code is switched over in this
PR — the existing singular `DownloadBlob` is kept for backwards compatibility, so
validators can be deployed before clients start using the new endpoint.

- Proto: `rpc DownloadBlobs(BlobIds) returns (stream BlobContent)`
- `ValidatorNode` trait: `async fn download_blobs(&self, blob_ids: Vec<BlobId>) ->
Result<BlobStream, NodeError>` where `BlobStream = BoxStream<'static,
Result<BlobContent, NodeError>>`
- gRPC transport: tonic native server streaming
- Simple transport: one dedicated connection per request, server sends one
`RpcMessage::DownloadBlobResponse` per blob, client filters them into a stream (same
pattern as `subscribe`)
- Server-side `DbStorage::read_blobs` (and `read_blob_states`) are changed from a
sequential loop into `futures::future::try_join_all` of per-blob reads. Each blob lives
under its own `root_key` (its own ScyllaDB partition), so cross-partition reads can't be
coalesced into a single `IN` query — the ScyllaDB-recommended approach is concurrent
per-partition queries via the shard-aware driver, which spreads the load across all
shards and nodes instead of funnelling through a single coordinator. RocksDB also
benefits: tokio can overlap the per-blob cache/SST lookups.

Callers will migrate to the new RPC in a follow-up PR (with retry-on-mid-am-error
logic that replays only the remaining blob ids on another validator).

## Test Plan

CI. The new RPC is additive and no existing caller code is changed, so existing tests
exercise the unchanged `DownloadBlob` path. The new endpoint will be validated
end-to-end in the follow-up PR that switches callers over.

## Release Plan

### DIFF
--- a/linera-core/src/remote_node.rs
+++ b/linera-core/src/remote_node.rs
@@ -224,6 +224,17 @@ impl<N: ValidatorNode> RemoteNode<N> {
         }
     }
 
+    /// Streams a batch of blobs from the validator. Each yielded item is
+    /// a `Result<Blob, NodeError>` — the caller can drive the stream incrementally
+    /// and, on error, track which blob IDs still need to be fetched.
+    #[instrument(level = "trace")]
+    pub async fn download_blobs(
+        &self,
+        blob_ids: Vec<BlobId>,
+    ) -> Result<crate::node::BlobStream, NodeError> {
+        self.node.download_blobs(blob_ids).await
+    }
+
     /// Downloads a list of certificates from the given chain.
     #[instrument(level = "trace")]
     pub async fn download_certificates_by_heights(
--- a/linera-rpc/src/simple/transport.rs
+++ b/linera-rpc/src/simple/transport.rs
@@ -16,7 +16,7 @@ use futures::{
     stream::{self, FuturesUnordered, SplitSink, SplitStream},
     Sink, SinkExt, Stream, StreamExt, TryStreamExt,
 };
-use linera_base::identifiers::ChainId;
+use linera_base::identifiers::{BlobId, ChainId};
 use linera_core::{JoinSetExt as _, TaskHandle};
 use serde::{Deserialize, Serialize};
 use tokio::{
@@ -92,6 +92,16 @@ pub trait MessageHandler: Clone {
     async fn handle_subscribe(
         &mut self,
         _chains: Vec<ChainId>,
+    ) -> Option<Pin<Box<dyn Stream<Item = RpcMessage> + Send>>> {
+        None
+    }
+
+    /// Handle a batch blob download request by streaming one
+    /// `RpcMessage::DownloadBlobResponse` per requested blob ID.
+    /// Returns `None` if not supported.
+    async fn handle_download_blobs(
+        &mut self,
+        _blob_ids: Vec<BlobId>,
     ) -> Option<Pin<Box<dyn Stream<Item = RpcMessage> + Send>>> {
         None
     }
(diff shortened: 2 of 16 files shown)


ndr-ds:
* **#5976** <a href="https://app.graphite.com/github/pr/linera-io/linera-protocol/5976?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/linera-io/linera-protocol/5976?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

linera-core/src/remote_node.rs
```diff
@@ -224,6 +224,17 @@ impl<N: ValidatorNode> RemoteNode<N> {
         }
     }
 
+    /// Streams a batch of blobs from the validator. Each yielded item is
+    /// a `Result<Blob, NodeError>` — the caller can drive the stream incrementally
+    /// and, on error, track which blob ids still need to be fetched.
```
afck, line 229:
```suggestion
    /// and, on error, track which blob IDs still need to be fetched.
```

linera-rpc/src/simple/transport.rs
```diff
@@ -95,6 +95,16 @@ pub trait MessageHandler: Clone {
     ) -> Option<Pin<Box<dyn Stream<Item = RpcMessage> + Send>>> {
         None
     }
+
+    /// Handle a batch blob download request by streaming one
+    /// `RpcMessage::DownloadBlobResponse` per requested blob id.
```
afck, line 100:
```suggestion
    /// `RpcMessage::DownloadBlobResponse` per requested blob ID.
```

linera-rpc/tests/snapshots/format__format.yaml.snap
```diff
@@ -1110,137 +1110,142 @@ RpcMessage:
         NEWTYPE:
           TYPENAME: BlobId
     8:
+      DownloadBlobs:
+        NEWTYPE:
+          SEQ:
+            TYPENAME: BlobId
```
afck, line 1116: (In the backport we'll need to put this at the end.)

linera-service/src/proxy/grpc.rs
```diff
@@ -616,6 +618,30 @@ where
         Ok(Response::new(blob.into_content().try_into()?))
     }
 
+    #[instrument(skip_all, err(Display), fields(method = "download_blobs"))]
+    async fn download_blobs(
+        &self,
+        request: Request<BlobIds>,
+    ) -> Result<Response<Self::DownloadBlobsStream>, Status> {
+        let blob_ids: Vec<linera_base::identifiers::BlobId> = request.into_inner().try_into()?;
```
afck, line 626: I think we prefer to avoid the type hint:

```suggestion
        let blob_ids = linera_base::identifiers::BlobId::try_from(request.into_inner())?;
```

afck (approved):

PR #5981, opened 2026-04-13
Create linera-json-wallet crate

deuszx:
## Motivation

`linera-core` is meant for core protocol logic that can be simulated. `PersistentWallet` wraps filesystem persistence (`linera_persistent::File`) — a concern that doesn't belong in the protocol core. Keystore operations and wallet path resolution were also scattered across `linera-service` CLI code with no clear ownership.

## Proposal

Introduce a new `linera-wallet` crate that owns all wallet and keystore persistence:

1. **Move `PersistentWallet`** from `linera-core::environment::wallet::persistent` to `linera-json-wallet`. Remove the `fs` feature and `linera-persistent` dependency from `linera-core`.

2. **Add `Keystore` type** wrapping `persistent::File<InMemorySigner>` with domain methods (`read`, `create`, `generate_key`, `generate_keys`, `save`, `into_signer`). Implements the `Signer` trait so it can be
passed directly to `ClientContext` — no more leaking `persistent::File` to callers.

3. **Move path resolution** (`config_dir`, `wallet_path`, `keystore_path`) from `linera-service` CLI options to `linera_wallet::paths`.

4. **Move wallet display** (`pretty_print`, `ChainDetails`) from `linera-service::wallet::WalletExt` to `linera_wallet::display`.

5. **Fix bridge relay key generation ordering** — generate keys before creating `ClientContext` instead of mutating shared `Arc` state after construction.

6. **Remove `linera-persistent` dependency from `linera-bridge`** — keystore access now goes through `linera-wallet::Keystore`.

## Test Plan

- `cargo check --workspace --exclude linera-web` compiles cleanly
- `cargo check -p linera-wallet` compiles
- `cargo check -p linera-bridge --features relay` compiles with no unused import warnings
- `cargo check -p linera-service` compiles with no unused import warnings
- Existing tests in `linera-service` and `linera-bridge` continue to pass (no behavioral changes)

## Release Plan

## Links


Twey:
As a high-level comment, it probably makes sense to call this crate something like `linera-json-wallet` — it doesn't define what the wallet is (that's done by the `Wallet` trait in `linera-core`) but a specific implementation of it as a JSON file on the filesystem.

deuszx:
> As a high-level comment, it probably makes sense to call this crate something like `linera-json-wallet` — it doesn't define what the wallet is (that's done by the `Wallet` trait in `linera-core`) but a specific implementation of it as a JSON file on the filesystem.

Done in https://github.com/linera-io/linera-protocol/pull/5981/commits/8d16108275d01a3c0c06ea81be89a775e3fc40e8

linera-service/src/wallet.rs
```diff
@@ -1,113 +1,4 @@
...
-            let chain_id = chain.chain_id;
-            (!chain.is_default, !chain.is_admin, root_id, chain_id)
-        });
-        for chain in chains {
-            chain.print_paragraph();
-        }
-        println!("------------------------");
-    }
-}
+pub use linera_wallet_json::PersistentWallet as Wallet;
```
Twey: I'm not sure this file deserves to exist any more :)

linera-wallet-json/src/lib.rs
Twey: It would be nice to put the wallet and the keystore in separate files.

linera-service/src/wallet.rs
deuszx: Done in [Remove linera-service/src/wallet.rs](https://github.com/linera-io/linera-protocol/pull/5981/commits/341d7883d1950226a633c09068ccb2fd573de5a0)

linera-wallet-json/src/lib.rs
deuszx: Done in [Extract Keystore to a file](https://github.com/linera-io/linera-protocol/pull/5981/commits/d2dbe2f5c33b4d46e10b5e2daeece0f501813d54) and [Extract PersistentWallet to a file](https://github.com/linera-io/linera-protocol/pull/5981/commits/2174957b166a41ed72ba44c5424d0756f9df8666)

linera-bridge/src/relay/mod.rs
```diff
@@ -232,21 +241,17 @@ pub async fn run(
         tracing::info!(%cid, %owner, "Using pre-existing chain");
         (cid, owner)
     } else {
-        // Claim from faucet.
+        // Claim from faucet using pre-generated key.
         let faucet = faucet
             .as_ref()
             .context("--faucet-url is required when --linera-bridge-chain-id is not provided")?;
+        let owner = faucet_owner.expect("faucet owner was generated above");
```
Twey, line 248: Nit: this logic is a little convoluted. Is it possible to merge it with the correlated branch above?

linera-service/src/cli/main.rs
```diff
@@ -62,7 +62,7 @@ use linera_execution::{committee::Committee, Operation};
 use linera_faucet_server::{FaucetConfig, FaucetService};
 #[cfg(with_metrics)]
 use linera_metrics::monitoring_server;
-use linera_persistent::{self as persistent, Persist, PersistExt as _};
+use linera_persistent::{self as persistent, Persist as _};
```
Twey, line 65: I think it would make sense if `linera-persistent` didn't escape from `linera-wallet-json` any more.

Twey (approved):

linera-bridge/src/relay/mod.rs
deuszx: it's because current version of the relayer tries to do two things:
1. Be prepared to be ran w/ pre-existing wallets and chains
2. Start without any inputs and claim everything on the fly (mostly for demo, test purposes)

I'll see if it's possible to simplify the code.

linera-service/src/cli/main.rs
deuszx: We still use it for the `GenesisConfig` in line 2168 which is later `persist()`-ed.

deuszx: I can create another wrapper for `GenesisConfig` but `linera-persistent` will still have to stay as a dependency of `lienra-service` b/c we're using it for `File<ValidatorServerConfig>`

Twey: Makes sense to me. Perhaps a future improvement of some kind. The genesis config is write-once so doesn't really need to be `Persist`, I just used it here to not duplicate the JSON-serialization logic.

PR #5982, opened 2026-04-13
[testnet] Add more clippy corrections.

MathieuDutSik:
## Motivation

We have added more clippy code to `main`. This made the code of the branch `main` and `testnet_conway` diverge.

## Proposal

Add more clippy corrections to `testnet_conway`.

## Test Plan

The CI.

## Release Plan

This is to `testnet_conway`.

## Links

None.


MathieuDutSik:
This does not appear to be really useful.

linera-storage-runtime/src/lib.rs
```diff
@@ -335,16 +329,20 @@ example service:tcp:127.0.0.1:7878:table_do_my_test"
             }
             if parts.len() == 2 || parts.len() == 3 {
                 let path = parts[0].to_string().into();
-                let spawn_mode = match parts[1] {
+                let spawn_mode_str = parts.get(1).expect("length already checked");
+                let spawn_mode = match *spawn_mode_str {
```
afck, line 339: (Which lint was this one? Feels like `parts[1]` should be fine here.)

linera-service/src/task_processor.rs
```diff
@@ -333,6 +333,7 @@ impl<Env: linera_core::Environment> TaskProcessor<Env> {
         Ok(outcome)
     }
 
+    #[allow(clippy::needless_pass_by_ref_mut)]
```
afck: Maybe `expect`?

linera-service/src/controller.rs
```diff
@@ -216,6 +216,7 @@ where
         }
     }
 
+    #[allow(clippy::needless_pass_by_ref_mut)]
```
afck: (Why do we need the `mut` here? Should these be `expect`?)

linera-sdk/src/test/block.rs
```diff
@@ -146,6 +146,7 @@ impl BlockBuilder {
     ///
     /// The operation is serialized using [`bcs`] and added to the block, marked to be executed by
     /// `application`.
+    #[allow(clippy::needless_pass_by_value)]
```
afck: `expect`?

linera-sdk/src/service/test_runtime.rs
```diff
@@ -464,6 +464,7 @@ where
     ///
     /// Cannot be used in fast blocks: A block using this call should be proposed by a regular
     /// owner, not a super owner.
+    #[allow(clippy::needless_pass_by_value)]
```
afck: `expect`?

linera-execution/src/wasm/runtime_api.rs
```diff
@@ -769,7 +770,8 @@ where
     ///
     /// This is called by the metering instrumentation, but the fuel consumed argument is
     /// ignored.
-    fn check_execution_time(caller: &mut Caller, _fuel_consumed: u64) -> Result<(), RuntimeError> {
+    #[allow(unused_variables)]
+    fn check_execution_time(caller: &mut Caller, fuel_consumed: u64) -> Result<(), RuntimeError> {
```
afck, line 773: Why not `_fuel_consumed` then? Or `expect(unused_variables)`?

linera-core/src/local_node.rs
```diff
@@ -298,8 +292,9 @@ where
         chain_ids: impl IntoIterator<Item = &ChainId>,
         receiver_id: ChainId,
     ) -> Result<BTreeMap<ChainId, BlockHeight>, LocalNodeError> {
-        let futures =
-            FuturesUnordered::from_iter(chain_ids.into_iter().map(|chain_id| async move {
+        let futures: FuturesUnordered<_> = chain_ids
```
afck: Type annotation should move down, to `collect`.

afck (commented):
I'm 100% on board with minimizing unnecessary differences between `testnet_conway` and `main`.

Some of the lints (`needless_pass_by_value`, `needless_pass_by_ref_mut`?) seem to require lots of exemptions, and cause tons of changes to the code. We could consider reverting those on `main` instead?

(Whatever we decide, let's apply it to both branches!)

linera-service/src/controller.rs
MathieuDutSik: This is for the `fn query_action` which is in `main`:
```rust
    // Keeping `&mut self` avoids borrowing `TaskProcessor` through `&self` across `.await`,
    // which would make the spawned future require `TaskProcessor: Sync`.
```

linera-core/src/local_node.rs
MathieuDutSik: Ok, done.

linera-service/src/task_processor.rs
MathieuDutSik: Ok. And also in `main`.

linera-service/src/controller.rs
MathieuDutSik: Add the `expect` statement.

linera-sdk/src/test/block.rs
MathieuDutSik: Ok.

linera-sdk/src/service/test_runtime.rs
MathieuDutSik: Ok.

linera-execution/src/wasm/runtime_api.rs
MathieuDutSik: The `_fuel_consumed` gets us an error of unused variable. So, use instead `expect(unused_variables)`.

linera-storage-runtime/src/lib.rs
MathieuDutSik: This fails the `-D clippy::match-on-vec-items`.

afck (approved):

linera-storage-runtime/src/lib.rs
afck: But we could probably:

```rust
match &*parts {
    ["spawn_blocking"] => …
    …
    _ => Err(…),
}
```

PR #5989, opened 2026-04-13
Check local storage before downloading sender certificates from validators

ndr-ds:
## Motivation

Two client functions download sender chain certificates from validators without first
checking local storage (RocksDB). After a worker restart, when multiple managed chains
sync from the same sender, the first chain's sync stores the certificates in RocksDB —
but subsequent chains re-download the same certificates from validators instead of
reading them locally. The same redundancy occurs when notification-driven downloads race
with `find_received_certificates`.

A third function in the same file, `download_event_bearing_blocks`, already follows the
correct pattern: it calls `storage_client().read_certificate(hash)` before downloading.
These changes make the two remaining functions consistent with that pattern.

## Proposal

**`download_sender_block_with_sending_ancestors`** (`linera-core/src/client/mod.rs`):
- Now checks RocksDB before downloading each certificate while walking the
`previous_message_blocks` chain
- Uses hash-based lookup when the hash is ailable from `previous_message_blocks`
(previously discarded as `_prev_hash`), falls back to height-based lookup on the first
iteration
- Introduced a small `try_read_local_certificate` helper shared with future callers

**`download_and_process_sender_chain`** (`linera-core/src/client/mod.rs`):
- Batch-checks RocksDB via `read_certificates_by_heights` at the top of each loop
iteration
- Processes locally-found certificates immediately, only downloads the remaining heights
from validators

## Test Plan

CI

## Release Plan

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -1189,32 +1189,14 @@ impl<Env: Environment> Client<Env> {
                     remote_heights.iter().copied().zip(local_certs.into_iter())
                 {
                     if let Some(certificate) = maybe_cert {
-                        let certificate = Arc::unwrap_or_clone(certificate);
-                        let check_result =
-                            Self::check_certificate(max_epoch, committees_ref, &certificate);
-                        let check_ok = check_result
-                            .map(|r| r.into_result().is_ok())
-                            .unwrap_or(false);
                         let chain_id = certificate.block().header.chain_id;
-                        if check_ok {
-                            let mode = ReceiveCertificateMode::AlreadyChecked;
-                            if self
-                                .receive_sender_certificate(certificate, mode, None)
-                                .await
-                                .is_ok()
-                            {
-                                if let Err(error) = sender.send(ChainAndHeight { chain_id, height })
-                                {
-                                    error!(
-                                        %chain_id, %height, %error,
-                                        "failed to send chain and height over the channel",
-                                    );
-                                }
-                                continue;
-                            }
+                        if let Err(error) = sender.send(ChainAndHeight { chain_id, height })
+                        {
+                            error!(
+                                %chain_id, %height, %error,
+                                "failed to send chain and height over the channel",
+                            );
                         }
-                        // Check or receive failed — try downloading from a validator.
-                        still_needed.push(height);
                     } else {
                         still_needed.push(height);
                     }
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -1185,13 +1185,10 @@ impl<Env: Environment> Client<Env> {
                 .await
             {
                 let mut still_needed = Vec::new();
-                for (height, maybe_cert) in
-                    remote_heights.iter().copied().zip(local_certs.into_iter())
-                {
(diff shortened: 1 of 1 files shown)


ndr-ds:
* **#5989** <a href="https://app.graphite.com/github/pr/linera-io/linera-protocol/5989?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/linera-io/linera-protocol/5989?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

linera-core/src/client/mod.rs
```diff
@@ -1159,6 +1177,54 @@ impl<Env: Environment> Client<Env> {
...
+                {
+                    if let Some(certificate) = maybe_cert {
+                        let certificate = Arc::unwrap_or_clone(certificate);
+                        let check_result =
+                            Self::check_certificate(max_epoch, committees_ref, &certificate);
+                        let check_ok = check_result
+                            .map(|r| r.into_result().is_ok())
+                            .unwrap_or(false);
+                        let chain_id = certificate.block().header.chain_id;
+                        if check_ok {
```
deuszx: I think we should be able to safely assume that if cetificate is in the storage then we don't have to re-validate it.

linera-core/src/client/mod.rs
```diff
@@ -1159,6 +1177,54 @@ impl<Env: Environment> Client<Env> {
...
+                            let mode = ReceiveCertificateMode::AlreadyChecked;
+                            if self
+                                .receive_sender_certificate(certificate, mode, None)
```
deuszx: if we already have it in the local storage - why do we have to receive it (process) again?

linera-core/src/client/mod.rs
```diff
@@ -1159,6 +1177,54 @@ impl<Env: Environment> Client<Env> {
...
+                                {
+                                    error!(
+                                        %chain_id, %height, %error,
+                                        "failed to send chain and height over the channel",
+                                    );
+                                }
+                                continue;
+                            }
+                        }
+                        // Check or receive failed — try downloading from a validator.
```
deuszx: if we get here this should be a fatal(?) error – we had a certificate in the local storage that wasn't correct.

linera-core/src/client/mod.rs
```diff
@@ -1159,6 +1177,36 @@ impl<Env: Environment> Client<Env> {
...
+            // Check local storage first — certificates may already be available from
+            // a prior sync cycle, another receiver chain, or a concurrent notification.
+            if let Ok(local_certs) = self
+                .storage_client()
+                .read_certificates_by_heights(sender_chain_id, &remote_heights)
+                .await
+            {
+                let mut still_needed = Vec::new();
+                for (height, maybe_cert) in
+                    remote_heights.iter().copied().zip(local_certs.into_iter())
```
```suggestion
                    remote_heights.iter().copied().zip(local_certs)
```

afck (approved):
Makes sense.

(In the most common case of a simple sender chain, `download_and_process_sender_chain` should be called with exactly the heights we download, but I agree there are cases where we might already have some of those blocks.)

PR #5991, opened 2026-04-13
Fix inverted chain worker TTL assignment

ndr-ds:
## Motivation

The chain worker TTL logic in `create_chain_worker` assigns `sender_chain_ttl` (short,
default 100ms) to tracked/owned chains and `ttl` (long, default 30min) to untracked
sender chains. This is backwards — tracked chains are the node's own high-value chains
that should stay loaded, while sender chains are transient and should be evicted
quickly.

On PM workers, this means market and event chains (FullChain mode) get evicted after
100ms of idle time, triggering expensive RocksDB reloads (~500ms p99) on every
subsequent request. Meanwhile, random user chains that sent a single order stay cached
for 30 minutes.

Production metrics from PM worker-5 (BTC-1 market chain):
- 6.4 chain state loads/second due to constant eviction of owned chains
- p99 load_chain latency: 509ms
- p99 queue wait: 686ms (requests queueing behind reloads)

On validators this has no effect because `chain_modes` is `None`, so `is_tracked` is
always `false` and all chains get `ttl`.

## Proposal

Flip the condition so tracked chains get the long TTL and sender chains get the short
one.

## Test Plan

CI

## Release Plan

### DIFF
--- a/linera-core/src/chain_worker/handle.rs
+++ b/linera-core/src/chain_worker/handle.rs
@@ -137,9 +137,9 @@ pub(crate) fn create_chain_worker<S: Storage + Clone + 'static>(
     let chain_id = state.chain().chain_id();
     let arc = Arc::new(RwLock::new(state));
     let ttl = if is_tracked {
-        config.sender_chain_ttl
-    } else {
         config.ttl
+    } else {
+        config.sender_chain_ttl
     };
     if let Some(ttl) = ttl {
         spawn_keep_alive(chain_id, Arc::clone(&arc), last_access, ttl);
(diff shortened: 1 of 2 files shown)


ndr-ds:
* **#5991** <a href="https://app.graphite.com/github/pr/linera-io/linera-protocol/5991?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/linera-io/linera-protocol/5991?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ma2bd (approved):

PR #5996, opened 2026-04-14
Introduce the `LazyRegisterView` and make it work with `description`, `application_permissions` and `ownership`.

MathieuDutSik:
## Motivation

The `RegisterView` load function is accessing the value under control when called.

This is a problem if the value in question is big. So, we introduce a `LazyRegisterView` that is loaded
only when appropriate.

## Proposal

Introduce a `LazyRegisterView` that is backward compatible with the `RegisterView`.
We are forced to use interior mutability. Because otherwise the "get()" becomes  `&mut self`.
In contexts of `&self` this forces us to introduce a `read_value` that reads the value without modifying.

The code is changed accordingly in a straightforward way.

## Test Plan

CI
A test that proves backward compatibility of `RegisterView` / `LazyRegisterView` is introduced.

## Release Plan

To `testnet_conway`.

## Links

None

### DIFF
--- a/linera-chain/src/chain.rs
+++ b/linera-chain/src/chain.rs
@@ -439,8 +439,8 @@ where
     }
 
     /// Invariant for the states of active chains.
-    pub async fn is_active(&self) -> bool {
-        self.execution_state.system.is_active().await
+    pub async fn is_active(&self) -> Result<bool, ChainError> {
+        Ok(self.execution_state.system.is_active().await?)
     }
 
     /// Initializes the chain if it is not active yet.
--- a/linera-chain/src/chain.rs
+++ b/linera-chain/src/chain.rs
@@ -439,8 +439,8 @@ where
     }
 
     /// Invariant for the states of active chains.
-    pub fn is_active(&self) -> bool {
-        self.execution_state.system.is_active()
+    pub async fn is_active(&self) -> bool {
+        self.execution_state.system.is_active().await
     }
 
     /// Initializes the chain if it is not active yet.
@@ -463,7 +463,7 @@ where
         let maybe_committee = self.execution_state.system.current_committee().into_iter();
         // Last, reset the consensus state based on the current ownership.
         self.manager.reset(
-            self.execution_state.system.ownership.get().clone(),
+            self.execution_state.system.ownership.get().await?.clone(),
             BlockHeight(0),
             local_time,
             maybe_committee.flat_map(|(_, committee)| committee.account_keys_and_weights()),
@@ -573,8 +573,8 @@ where
             .ok_or_else(|| ChainError::InactiveChain(self.chain_id()))
     }
 
-    pub fn ownership(&self) -> &ChainOwnership {
-        self.execution_state.system.ownership.get()
+    pub async fn ownership(&self) -> Result<&ChainOwnership, ChainError> {
+        Ok(self.execution_state.system.ownership.get().await?)
     }
 
     /// Removes the incoming message bundles in the block from the inboxes.
@@ -983,7 +983,11 @@ where
         }
 
         Self::check_app_permissions(
-            self.execution_state.system.application_permissions.get(),
+            self.execution_state
+                .system
+                .application_permissions
+                .get()
+                .await?,
             &block,
         )?;
 
@@ -1023,7 +1027,8 @@ where
         self.process_outgoing_messages(block).await?;
 
         // Last, reset the consensus state based on the current ownership.
-        self.reset_chain_manager(block.header.height.try_add_one()?, local_time)?;
+        self.reset_chain_manager(block.header.height.try_add_one()?, local_time)
+            .await?;
 
         // Advance to next block height.
         let tip = self.tip_state.get_mut();
@@ -1058,12 +1063,13 @@ where
     }
 
     /// Returns whether this is a child chain.
-    pub fn is_child(&self) -> bool {
-        let Some(description) = self.execution_state.system.description.get() else {
+    pub async fn is_child(&self) -> Result<bool, ChainError> {
+        let description = self.execution_state.system.description.get().await?;
+        let Some(description) = description else {
             // Root chains are always initialized, so this must be a child chain.
-            return true;
+            return Ok(true);
         };
-        description.is_child()
+        Ok(description.is_child())
     }
 
     /// Verifies that the block is valid according to the chain's application permission settings.
@@ -1152,13 +1158,13 @@ where
     }
 
     /// Resets the chain manager for the next block height.
-    fn reset_chain_manager(
+    async fn reset_chain_manager(
         &mut self,
         next_height: BlockHeight,
         local_time: Timestamp,
     ) -> Result<(), ChainError> {
         let maybe_committee = self.execution_state.system.current_committee().into_iter();
-        let ownership = self.execution_state.system.ownership.get().clone();
+        let ownership = self.execution_state.system.ownership.get().await?.clone();
         let fallback_owners =
             maybe_committee.flat_map(|(_, committee)| committee.account_keys_and_weights());
         self.pending_validated_blobs.clear();
--- a/linera-core/src/chain_worker/state.rs
+++ b/linera-core/src/chain_worker/state.rs
@@ -1113,7 +1113,7 @@ where
         }
         inbox.observe_size_metric();
         drop(inbox);
-        if !self.config.allow_inactive_chains && !self.chain.is_active().await {
+        if !self.config.allow_inactive_chains && !self.chain.is_active().await? {
             // Refuse to create a chain state if the chain is still inactive by
             // now. Accordingly, do not send a confirmation, so that the
             // cross-chain update is retried later.
--- a/linera-core/src/chain_worker/state.rs
+++ b/linera-core/src/chain_worker/state.rs
@@ -610,7 +610,7 @@ where
             .get()
             .already_validated_block(certificate.inner().height())?
         {
-            return Ok((self.chain_info_response(), NetworkActions::default()));
+            return Ok((self.chain_info_response().await?, NetworkActions::default()));
         }
         ensure!(
             certificate.inner().epoch() == chain_epoch,
@@ -626,7 +626,7 @@ where
             .handle_timeout_certificate(certificate, self.storage.clock().current_time());
         self.save().await?;
         let actions = self.create_network_actions(Some(old_round)).await?;
-        Ok((self.chain_info_response(), actions))
+        Ok((self.chain_info_response().await?, actions))
     }
 
     /// Tries to load all blobs published in this proposal.
@@ -659,7 +659,7 @@ where
         let missing_blob_ids = missing_blob_ids(&maybe_blobs);
         if !missing_blob_ids.is_empty() {
             let chain = &mut self.chain;
-            if chain.ownership().open_multi_leader_rounds {
+            if chain.ownership().await?.open_multi_leader_rounds {
                 // TODO(#3203): Allow multiple pending proposals on permissionless chains.
                 chain.pending_proposed_blobs.clear();
             }
@@ -717,7 +717,7 @@ where
         if already_committed_block || should_skip_validated_block()? {
             // If we just processed the same pending block, return the chain info unchanged.
             return Ok((
-                self.chain_info_response(),
+                self.chain_info_response().await?,
                 NetworkActions::default(),
                 BlockOutcome::Skipped,
             ));
@@ -750,7 +750,11 @@ where
         )?;
         self.save().await?;
         let actions = self.create_network_actions(Some(old_round)).await?;
-        Ok((self.chain_info_response(), actions, BlockOutcome::Processed))
+        Ok((
+            self.chain_info_response().await?,
+            actions,
+            BlockOutcome::Processed,
+        ))
     }
 
     /// Processes a confirmed block (aka a commit).
@@ -775,7 +779,11 @@ where
             let actions = self.create_network_actions(None).await?;
             self.register_delivery_notifier(height, &actions, notify_when_messages_are_delivered)
                 .await;
-            return Ok((self.chain_info_response(), actions, BlockOutcome::Skipped));
+            return Ok((
+                self.chain_info_response().await?,
+                actions,
+                BlockOutcome::Skipped,
+            ));
         }
 
         // We haven't processed the block - verify the certificate first
@@ -866,7 +874,7 @@ where
             self.register_delivery_notifier(height, &actions, notify_when_messages_are_delivered)
                 .await;
             return Ok((
-                self.chain_info_response(),
+                self.chain_info_response().await?,
                 actions,
                 BlockOutcome::Preprocessed,
             ));
@@ -999,7 +1007,11 @@ where
         self.register_delivery_notifier(height, &actions, notify_when_messages_are_delivered)
             .await;
 
-        Ok((self.chain_info_response(), actions, BlockOutcome::Processed))
+        Ok((
+            self.chain_info_response().await?,
+            actions,
+            BlockOutcome::Processed,
+        ))
     }
 
     /// Schedules a notification for when cross-chain messages are delivered up to the given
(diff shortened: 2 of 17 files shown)


afck:
> We are forced to use interior mutability. Because otherwise the "get()" becomes `&mut self`.

Why not make it async? I think that would be more in line with how other views work. And it doesn't have to have the same API as `RegisterView`, just identical storage formats, I think?

linera-execution/src/system.rs
```diff
@@ -340,18 +341,19 @@ where
...
     }
 
     /// Returns the current committee, if any.
-    pub fn current_committee(&self) -> Option<(Epoch, &Committee)> {
-        let epoch = self.epoch.get();
-        let committee = self.committees.get().get(epoch)?;
-        Some((*epoch, committee))
+    pub async fn current_committee(&self) -> Option<(Epoch, &Committee)> {
+        let epoch = *self.epoch.get();
+        let committees = self.committees.get().await.ok()?;
```
deuszx: Is this change improving anything? We are sitll going to load all committees at once. Sure, it may be lazy but I'd imagine that eventually all paths go through here (ie they need the committee from `SystemExecutionStateView`).

A better solution would be to support getting each key separately.

Also – can we cache the result of this call?

afck: This is only one half of the solution; @ndr-ds is working on the other half: Having a shared in-memory map of all committees per epoch. Then we don't have to look into this register view at all anymore!

(Well, apart from that we somehow need to know the oldest trusted epoch. Details TBD.)

deuszx: But is this new structure, as shown to be used here specifically, is improving performance/memory footprint? We are loading all committees at once. 

The second part you mention – shared, in-memory map of all committees would help even without `LazyRegisterView`?

afck: No, it wouldn't, because the regular `RegisterView` gets loaded even if we don't use its contents!

Also, it will help for some other fields like `ownership`: E.g. in sparse (sender or event publisher) chains those are probably never read, so we can avoid loading them.

I agree that on `main`, in the longer term, we will probably find a much better solution for `committees`. But I think it makes sense to have a `LazyRegisterView` for `ownership`, `application_permissions` and others.

afck: (Much better solution for `committees`: Just store the hash of the committee blob instead of the whole committee.)

deuszx: > No, it wouldn't, because the regular `RegisterView` gets loaded even if we don't use its contents!

If we have all committtees in the shared map then _someone_ will call `current_comittee` eventually which will be better than this.

Also, my second part of the first question was whether it's true that eventually we'd call `current_committee` on every `System*View`.

afck: OK, maybe let's use `LazyRegisterView` for `description`, `application_permissions` and `ownership` on `main`, but not `committees` yet, because the solution for that one will look different on `main`. (What do you think, @MathieuDutSik?)

On `testnet_conway`, we will make `committees` a `LazyRegisterView` so that we don't have to do a migration. @ndr-ds's work will make it possible to not access the field on most chain states anymore.

MathieuDutSik:
> > We are forced to use interior mutability. Because otherwise the "get()" becomes `&mut self`.
> 
> Why not make it async? I think that would be more in line with how other views work. And it doesn't have to have the same API as `RegisterView`, just identical storage formats, I think?

The function `get` becomes async no matter what, this is for sure.

The thing that is unwelcome is that if we do not use interior mutability, then the function `get` is going to be `&mut self` and that is a problem.

linera-execution/src/system.rs
MathieuDutSik: Ok, we can do that.

afck:
> The thing that is unwelcome is that if we do not use interior mutability, then the function get is going to be `&mut self` and that is a problem.

I guess the behavior that would be most in line with how the other views work would be to have an async `get` with `&self` that simply does _not_ keep the read data in memory?
* On the one hand that might be fine, because the read data is in the underlying LRU cache.
* On the other hand the LRU cache contains the _serialized_ data, so the cost of deserialization would be there every time we call `get`.

Not sure what's best. :confused: Personally I'm fine with interior mutability, too.

linera-views/src/views/lazy_register_view.rs
```diff
@@ -0,0 +1,330 @@
...
+    pub async fn get(&self) -> Result<&T, ViewError> {
+        if let Some(value) = &self.update {
+            return Ok(value);
+        }
+        self.ensure_stored_value().await?;
+        let guard = self.stored_value.lock().unwrap();
+        let ptr: *const T = guard.as_ref().unwrap().as_ref();
+        // SAFETY: Once stored, the value behind the Box is never moved or removed
+        // except through &mut self methods. The Box provides a stable heap address.
+        Ok(unsafe { &*ptr })
```
ma2bd: what is this?

ma2bd: We don't want unsafe code

afck: Yes, if we decide we do go with interior mutability, it needs to be safe, of course.

Does `LazyCell` not cover this exact use case?

MathieuDutSik: Yes, sorry for that. Eliminated.

linera-chain/src/chain.rs
```diff
@@ -460,13 +460,16 @@ where
...
-            self.execution_state.system.ownership.get().clone(),
+            self.execution_state.system.ownership.get().await?.clone(),
             BlockHeight(0),
             local_time,
-            maybe_committee.flat_map(|(_, committee)| committee.account_keys_and_weights()),
+            self.execution_state
+                .system
+                .current_committee()
+                .into_iter()
+                .flat_map(|(_, committee)| committee.account_keys_and_weights()),
```
afck: (This change is a no-op now, is it?)

linera-chain/src/chain.rs
```diff
@@ -1058,11 +1066,14 @@ where
     }
 
     /// Returns whether this is a child chain.
-    pub fn is_child(&self) -> bool {
-        let Some(description) = self.execution_state.system.description.get() else {
+    pub async fn is_child(&self) -> bool {
+        let Ok(description) = self.execution_state.system.description.get().await else {
```
afck: We should probably distinguish between actual DB errors and the entry not existing.

afck: Edit: I.e. use `?` here?

linera-chain/src/chain.rs
```diff
@@ -1152,19 +1163,24 @@ where
...
-            .reset(ownership, next_height, local_time, fallback_owners)
+        self.manager.reset(
+            ownership,
+            next_height,
+            local_time,
+            self.execution_state
+                .system
+                .current_committee()
+                .into_iter()
+                .flat_map(|(_, committee)| committee.account_keys_and_weights()),
```
afck: (This part of the change could also be reverted.)

linera-core/src/chain_worker/state.rs
```diff
@@ -2182,10 +2196,12 @@ impl<'a> CrossChainUpdateHelper<'a> {
...
         C: Context + Clone + 'static,
     {
+        let current_epoch = *chain.execution_state.system.epoch.get();
+        let committees = chain.execution_state.system.committees.get();
         CrossChainUpdateHelper {
             allow_messages_from_deprecated_epochs: config.allow_messages_from_deprecated_epochs,
-            current_epoch: *chain.execution_state.system.epoch.get(),
-            committees: chain.execution_state.system.committees.get(),
+            current_epoch,
+            committees,
```
afck: (Also a no-op change.)

linera-execution/src/system.rs
```diff
@@ -340,18 +341,26 @@ where
...
-            && self.ownership.get().is_active()
+    pub async fn is_active(&self) -> bool {
+        let description = match self.description.get().await {
+            Ok(description) => description,
+            Err(_) => return false,
+        };
+        let ownership = match self.ownership.get().await {
+            Ok(ownership) => ownership,
+            Err(_) => return false,
+        };
```
afck: I wonder if it's enough to just load `ownership` here, without `description`. I don't think `current_committee` or `admin_chain_id` can be `Some` without the description? (But please double-check!)

linera-chain/src/chain.rs
MathieuDutSik: Yes, corrected.

linera-core/src/chain_worker/state.rs
MathieuDutSik: Yes.

linera-chain/src/chain.rs
MathieuDutSik: Yes.

linera-chain/src/chain.rs
MathieuDutSik: Ok.
But note that this function "is_child" is not used at all in the codebase.

linera-execution/src/system.rs
MathieuDutSik: Maybe, but let us focus on another problem of this function: It should return an ExecutionError and a storage error should result in an error going upstream, not returning false.

I think that this is more important that this other issue which can be treated independently in another PR.

linera-views/src/views/lazy_register_view.rs
```diff
@@ -0,0 +1,327 @@
...
+    }
+}
+
+impl<C, T> ClonableView for LazyRegisterView<C, T>
+where
+    C: Context,
+    T: Clone + Default + Send + Sync + Serialize + DeserializeOwned,
+{
+    fn clone_unchecked(&mut self) -> Result<Self, ViewError> {
+        let stored_value = match self.stored_value.get_mut() {
```
afck:
```suggestion
        let stored_value = match self.stored_value.get() {
```

linera-views/src/views/lazy_register_view.rs
```diff
@@ -0,0 +1,327 @@
...
+            Some(value) => OnceLock::from(value.clone()),
+            None => OnceLock::new(),
+        };
```
afck: Is this equivalent to `self.stored_value.clone()`? I.e. could we even derive `Clone`?

linera-views/src/views/lazy_register_view.rs
```diff
@@ -0,0 +1,327 @@
...
+    C2: Context,
+    T: Default + Send + Sync + Serialize + DeserializeOwned + Clone,
+{
+    type Target = LazyRegisterView<C2, T>;
+
+    async fn with_context(
+        &mut self,
+        ctx: impl FnOnce(&Self::Context) -> C2 + Clone,
+    ) -> Self::Target {
+        let stored_value = match self.stored_value.get_mut() {
```
afck:
```suggestion
        let stored_value = match self.stored_value.get() {
```

linera-views/src/views/lazy_register_view.rs
```diff
@@ -0,0 +1,327 @@
...
+            Some(value) => OnceLock::from(value.clone()),
+            None => OnceLock::new(),
+        };
```
afck: Or just `self.stored_value.clone()`?

linera-views/src/views/lazy_register_view.rs
```diff
@@ -0,0 +1,327 @@
...
+    T: Default + DeserializeOwned,
+{
+    /// Ensures the stored value is loaded from storage into the `OnceLock` cache.
+    async fn ensure_stored_value(&self) -> Result<(), ViewError> {
+        if self.stored_value.get().is_some() {
+            return Ok(());
+        }
+        let key = self.context.base_key().bytes.clone();
+        let bytes = self.context.store().read_value_bytes(&key).await?;
+        let value: T = from_bytes_option_or_default(&bytes)?;
```
afck:
```suggestion
        let value = from_bytes_option_or_default::<T>(&bytes)?;
```

linera-views/src/views/lazy_register_view.rs
```diff
@@ -0,0 +1,327 @@
...
+    }
+}
+
+impl<C, T> LazyRegisterView<C, T>
+where
+    C: Context,
+    T: Default + DeserializeOwned,
+{
+    /// Ensures the stored value is loaded from storage into the `OnceLock` cache.
+    async fn ensure_stored_value(&self) -> Result<(), ViewError> {
```
afck: I think this should really be inlined in `get`. That would avoid the `unwrap` below. In some of the other places where we call it we actually also want to `get` the value anyway, and in the others I'm sure the compiler will optimize away the unused reference.

linera-views/src/views/lazy_register_view.rs
```diff
@@ -0,0 +1,327 @@
...
+
+    fn compute_hash(&self) -> Result<<sha3::Sha3_256 as Hasher>::Output, ViewError> {
+        #[cfg(with_metrics)]
+        let _hash_latency = metrics::LAZY_REGISTER_VIEW_HASH_RUNTIME.measure_latency();
+        let value: &T = if let Some(value) = &self.update {
+            value
+        } else {
+            self.stored_value
+                .get()
+                .expect("stored value must be loaded before hashing")
```
afck: Is that a panic in other views, too?

linera-views/src/views/lazy_register_view.rs
MathieuDutSik: Yes, we can clone, thank you.

linera-views/src/views/lazy_register_view.rs
MathieuDutSik: Yes.

linera-views/src/views/lazy_register_view.rs
MathieuDutSik: Ok, done.

linera-views/src/views/lazy_register_view.rs
MathieuDutSik: Ok, type annotation was not needed.

linera-views/src/views/lazy_register_view.rs
MathieuDutSik: No, and that was a real bug. Code restructured.

linera-views/src/views/lazy_register_view.rs
MathieuDutSik: Yes, restructured.

linera-views/src/views/lazy_register_view.rs
MathieuDutSik: Yes, done.

linera-views/src/views/lazy_register_view.rs
```diff
@@ -199,7 +180,12 @@ where
...
             return Ok(value);
         }
-        self.ensure_stored_value().await?;
+        if self.stored_value.get().is_none() {
+            let key = self.context.base_key().bytes.clone();
+            let bytes = self.context.store().read_value_bytes(&key).await?;
+            let value = from_bytes_option_or_default(&bytes)?;
+            let _ = self.stored_value.set(Box::new(value));
+        }
         Ok(self.stored_value.get().unwrap())
```
afck: You could avoid the `unwrap` now.

linera-views/src/views/lazy_register_view.rs
```diff
@@ -247,25 +233,22 @@ where
...
-            self.ensure_stored_value().await?;
-            let stored = self.stored_value.get().unwrap();
-            self.update = Some(stored.clone());
+            if self.stored_value.get().is_none() {
+                let key = self.context.base_key().bytes.clone();
+                let bytes = self.context.store().read_value_bytes(&key).await?;
+                let value = from_bytes_option_or_default(&bytes)?;
+                let _ = self.stored_value.set(Box::new(value));
+            }
+            self.update = Some(self.stored_value.get().unwrap().clone());
```
afck:
```suggestion
    let update = self.get().clone();
    self.update = Some(update);
```

linera-views/src/views/lazy_register_view.rs
MathieuDutSik: Ok, corrected.

linera-views/src/views/lazy_register_view.rs
MathieuDutSik: Ok, done.

afck (approved):

PR #5999, opened 2026-04-14
Don't fetch blob when retrying x-chain requests for sender chains (#5998)

afck:
Port of #5998.

## Motivation

`LocalNodeClient::retry_pending_cross_chain_requests` went through `handle_chain_info_query`, which unconditionally calls `initialize_and_save_if_needed`. For a sender chain whose `ChainDescription` blob was never needed locally (e.g. only non-height-0 blocks preprocessed), the init step would fail with `BlobsNotFound` and the caller would download the `ChainDescription` from validators — once per sender per sync.

## Proposal

Add a dedicated worker entry point (`cross_chain_network_actions`) that computes pending cross-chain requests from the outbox without touching execution state, and route the retry through it.

## Test Plan

A test was extended to assert that a sparse sender chain's description wasn't downloaded.

## Release Plan

- Backport, but without the change to the format.

## Links

- `testnet_conway` version: #5998.

### DIFF
--- a/linera-core/src/local_node.rs
+++ b/linera-core/src/local_node.rs
@@ -118,9 +118,7 @@ where
         &self,
         query: ChainInfoQuery,
     ) -> Result<ChainInfoResponse, LocalNodeError> {
-        // In local nodes, cross-chain actions will be handled internally, so we discard them.
-        let (response, _actions) = self.node.state.handle_chain_info_query(query).await?;
-        Ok(response)
+        Ok(self.node.state.handle_chain_info_query(query).await?)
     }
 
     #[instrument(level = "trace", skip_all)]
--- a/linera-core/src/local_node.rs
+++ b/linera-core/src/local_node.rs
@@ -264,16 +264,22 @@ where
     }
 
     /// Handles any pending local cross-chain requests.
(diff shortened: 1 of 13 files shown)


linera-core/src/local_node.rs
```diff
@@ -264,16 +264,22 @@ where
...
     pub async fn retry_pending_cross_chain_requests(
         &self,
         sender_chain: ChainId,
         notifier: &impl Notifier,
     ) -> Result<(), LocalNodeError> {
-        let (_response, actions) = self
+        let actions = self
             .node
             .state
-            .handle_chain_info_query(ChainInfoQuery::new(sender_chain).with_network_actions())
```
deuszx, line 276: It looks like `with_network_actions` is not used anywhere else anymore

deuszx (approved):

linera-core/src/local_node.rs
afck: Removed it (and everything related to it—good catch!)

deuszx: Maybe we should do something similar on testnet_conway but in backwards compatible way?

PR #6009, opened 2026-04-14
Add `--never-reject-application-ids` client option

afck:
## Motivation

In some use cases we never want to reject an application's messages.

## Proposal

Add a new `MessagePolicy` field and matching CLI flag listing application IDs whose messages must never be rejected. Bundles containing any such message bypass the other rejection rules (except `--restrict-chain-ids-to`), and if execution fails they are discarded (along with subsequent bundles from the same sender) for retry in a later block, with a warning logged
instead of triggering a bounce.

## Test Plan

Tests were extended.

## Release Plan

- Release SDK

## Links


linera-base/src/data_types.rs
```diff
@@ -1639,6 +1639,11 @@ pub struct MessagePolicy {
     /// A collection of applications: If `Some`, only event streams from those
     /// applications will be processed.
     pub process_events_from_application_ids: Option<HashSet<GenericApplicationId>>,
+    /// A collection of applications whose messages must never be rejected. Bundles containing
+    /// at least one message from any of these applications bypass the other rejection rules
+    /// (except `restrict_chain_ids_to`), and on execution failure they are discarded for later
+    /// retry instead of being rejected. An empty set disables this feature.
```
ma2bd, line 1645: Now of course this raises the question of how often should these be retried in the future.

afck: Yes, we could blacklist the sender for some time.

linera-chain/src/data_types/mod.rs
```diff
@@ -280,6 +285,13 @@ impl IncomingBundle {
                 return false;
             }
         }
+        if self.messages().any(|posted_msg| {
+            policy
+                .never_reject_application_ids
+                .contains(&posted_msg.message.application_id())
```
deuszx, line 291: I feel like `matches_policy` is such an ambiguous name that it doesnt' really say anything about when/how the method should be used. It tries to match the incoming bundle against _all_ fields in the message policy but what does the `true`/`false` mean then? if it returns `true` we have no idea which policy specifically it matched – was it **"should I reject?"** or **"should I retry?"**

afck: Yes, maybe it should just be called `action`?

bart-linera (commented):
AFAICT, the PR does what it is supposed to do, but I wonder if it is a good idea. I imagine someone could try to leverage this by writing a malicious app that calls such a never-rejected app and sends its own messages as well, thus forcing a client to process messages from a malicious app.

ma2bd:
> AFAICT, the PR does what it is supposed to do, but I wonder if it is a good idea. I imagine someone could try to leverage this by writing a malicious app that calls such a never-rejected app and sends its own messages as well, thus forcing a client to process messages from a malicious app. 

Right, a bundle should be protected from rejection only if all the components are protected

afck:
Changed it so it's only protected if all messages are in the whitelist.

linera-chain/src/chain.rs
ma2bd: Perhaps we should not fail the block in this case any more but treat protected messages as never-rejectable.

ma2bd (approved):

linera-chain/src/chain.rs
afck: Done in https://github.com/linera-io/linera-protocol/pull/6009/commits/879728c04b30d722933195c1d9804acf006dcf51.

PR #6014, opened 2026-04-15
Remove unused functions and types

MathieuDutSik:
## Motivation

Many functions and types which are internal, that is which are not part of `linera-sdk` or `linera-views` are not used and can be removed.

## Proposal

Do a systematic search for such functions. Also mark the functions that are used only in tests as `#[cfg(with_testing)]`.

## Test Plan

CI

## Release Plan

The same analysis can likely be done on `testnet_conway`.

## Links

None


ma2bd (approved):
good catch!

PR #6015, opened 2026-04-15
Fix `BucketQueueView::delete_front` load failure handling

afck:
## Motivation

We see some `entered unreachable code: The front bucket is always loaded.` in the validator 4 logs. We are still investigating the original DB error, but a failed read could cause this:

`BucketQueueView::delete_front` advances `cursor` to the new offset before attempting to load the new front bucket. If the read fails, the view is left with `cursor.offset` pointing at a `NotLoaded` bucket, which is supposed to be `unreachable!`.

## Proposal

Load the new bucket first, and only advance the cursor on success.

## Test Plan

A regression test was added.

## Release Plan

- Port to `main`.
- Release SDK.
- Hotfix validators.

## Links


MathieuDutSik (approved):
LGTM

PR #6021, opened 2026-04-15
[testnet] Introduce the use of LazyRegisterView in the system.

MathieuDutSik:
## Motivation

The RegisterView load function is accessing the value under control when called.

This is a problem if the value in question is big. So, we introduce a LazyRegisterView that is loaded
only when appropriate.

Backport of #5996 

## Proposal

Introduce a LazyRegisterView that is backward compatible with the RegisterView.
We are forced to use interior mutability. Because otherwise the "get()" becomes &mut self.
In contexts of &self this forces us to introduce a read_value that reads the value without modifying.

The code is changed accordingly in a straightforward way.

## Test Plan

CI
A test that proves backward compatibility of RegisterView / LazyRegisterView is introduced.

## Release Plan

This is to `testnet_conway`. So no release needed.

## Links

None.

### DIFF
--- a/linera-core/src/chain_worker/state.rs
+++ b/linera-core/src/chain_worker/state.rs
@@ -1975,8 +1975,7 @@ where
         query: ChainInfoQuery,
     ) -> Result<ChainInfoResponse, WorkerError> {
         self.initialize_and_save_if_needed().await?;
-        let chain = &self.chain;
-        let mut info = ChainInfo::from_chain_view(chain).await?;
+        let mut info = ChainInfo::from_chain_view(&self.chain).await?;
         if query.request_committees {
             info.requested_committees =
                 Some(self.chain.execution_state.system.committees.get().clone());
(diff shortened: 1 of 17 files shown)


linera-core/src/chain_worker/state.rs
```diff
@@ -1962,7 +1975,8 @@ where
         query: ChainInfoQuery,
     ) -> Result<ChainInfoResponse, WorkerError> {
         self.initialize_and_save_if_needed().await?;
-        let mut info = ChainInfo::from(&self.chain);
+        let chain = &self.chain;
```
afck: Is that local variable useful? Otherwise I'd inline it below.

afck (approved):
Looks good! (Assuming `lazy_register_view.rs` and the tests are identical to `main`.)

linera-core/src/chain_worker/state.rs
MathieuDutSik: Yes, done.

Issue #5723, opened 2026-03-17
`test_controller::remote_net_grpc` is flaky.


afck:
Example: https://github.com/linera-io/linera-protocol/actions/runs/23115529989/job/67140313616

afck:
It seems to fail consistently on `testnet_conway` now: https://github.com/linera-io/linera-protocol/actions/runs/23539317702/job/68522512042

PR #6022, opened 2026-04-15
Time-bound the "follow nonexistent chain" check in the test.

afck:
## Motivation

`follow_chain` on a synthetic `ChainId` (`CryptoHash::test_hash("wrong chain ID")`, which is e7e3a7314598ba7afe6fa2c281c4a1b0a0eb75d6db43f2d2616acc60b3cba5b2) eventually errors, but only after the client exhausts its per-validator staggered retries for the missing `ChainDescription` blob. On testnets with many validators that can take minutes, during which the test appears hung.

## Proposal

Wrap the call in a 5s timeout so the test either sees the expected error promptly or kills the child process via `kill_on_drop`, preserving the original intent without paying the full retry budget every run.

Also, make `test_controller` not rely on exact block heights anymore: This is flaky, because it's not deterministic whether e.g. incoming bundles are received in the same or different blocks.

## Test Plan

This makes the tests pass for me locally, rater than hang for a long time.

The compatibility tests should now pass again in CI.

## Release Plan

- Port to `main`.

## Links

- Fixes https://github.com/linera-io/linera-protocol/issues/5723.

### DIFF
--- a/linera-service/tests/linera_net_tests.rs
+++ b/linera-service/tests/linera_net_tests.rs
@@ -5032,8 +5032,7 @@ async fn test_controller(config: impl LineraNetConfig) -> Result<()> {
     let app1 = node_service1.make_application(&worker1_chain, &controller_id)?;
     let worker1 = loop {
         let response = app1.query("localWorkerState").await?;
-        let state: LocalWorkerState =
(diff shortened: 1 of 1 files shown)


ma2bd (approved):
nice!

deuszx (approved):

PR #6024, opened 2026-04-15
Correct the lazy_register_view bug and add some reasons for the unrea…

MathieuDutSik:
## Motivation

The `LazyRegisterView::get_mut` could modify the state and leave it in an incorrect state if an error occurs.

## Proposal

We also add reasons for the unreachable statements.

## Test Plan

CI.

## Release Plan

It can be backported to `testnet_conway`.

## Links

None.

### DIFF
--- a/linera-views/src/views/bucket_queue_view.rs
+++ b/linera-views/src/views/bucket_queue_view.rs
@@ -458,7 +458,7 @@ impl<C: Context, T: DeserializeOwned + Clone, const N: usize> BucketQueueView<C,
             Some(Cursor { offset, position }) => {
                 let bucket = &self.stored_buckets[offset];
                 let State::Loaded { data } = &bucket.state else {
-                    unreachable!();
+                    unreachable!("The front bucket should always be loaded");
                 };
                 Some(&data[position])
             }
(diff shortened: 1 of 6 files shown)


linera-views/src/views/bucket_queue_view.rs
```diff
@@ -458,7 +458,7 @@ impl<C: Context, T: DeserializeOwned + Clone, const N: usize> BucketQueueView<C,
             Some(Cursor { offset, position }) => {
                 let bucket = &self.stored_buckets[offset];
                 let State::Loaded { data } = &bucket.state else {
-                    unreachable!();
+                    unreachable!("The front bucket should always be loaded");
```
afck, line 461: We could make these messages distinct so that _if_ something goes wrong we know exactly where. E.g. by adding the function name:
```suggestion
                    unreachable!("BucketQueueView::front: The front bucket should always be loaded");
```

afck (approved):
Looks good to me.

In the long term, I'd also like to revisit some of these and see if we can remove a few `unreachable!`s by changing the in-memory representation of the views (e.g. `front` could be a separate field rather than just the first entry of the buckets list). Also, I'm not 100% convinced that all of them are unreachable even in the case of database corruption. In general, I think database corruptions should always cause errors, not panics.

afck (approved):

MathieuDutSik:
> Looks good to me.
> 
> In the long term, I'd also like to revisit some of these and see if we can remove a few `unreachable!`s by changing the in-memory representation of the views (e.g. `front` could be a separate field rather than just the first entry of the buckets list). Also, I'm not 100% convinced that all of them are unreachable even in the case of database corruption. In general, I think database corruptions should always cause errors, not panics.

I agree, avoiding the "unreachable" statements is preferable altogether. We should be able to reformulate the Rust code so that it does not happen.

Ans yes, all of them should be unreachable also in case of database corruption. I think they are but I will check some more.

PR #6025, opened 2026-04-15
Fix crate name

deuszx:
## Motivation

An embarassing mistake was made: the name was supposed to be `linera-json-wallet`

## Proposal

Rename `linera-wallet-json` to `linera-json-wallet`.

## Test Plan

CI

## Release Plan

## Links


Twey:
I actually think this name is better 😅 It follows a strict larger to smaller order (a Linera thing that is a wallet implementation by means of JSON). E.G. if we were to introduce an IndexedDB wallet crate we might want to have `linera-wallet-indexeddb` which could come under the same `linera-wallet-*` grouping.

deuszx:
Ok, as you prefer. My thinking was it's a JSON wallet, not wallet JSON

PR #6032, opened 2026-04-15
Implement the execution of the streams, messages and operations in persistent instances over the block.

MathieuDutSik:
## Motivation

When executing a message, an operation, or a stream then we are instantiating, processing, and then finalizing.
So, the finalization is done many times, which is inefficient.

## Proposal

By having persistent instances for contracts, we can address many issues with the execution model: The contract are running, and when we finish the block, we save. That covers many scenario:
* Contract A executes a message, Contract B executes another message which lead to executing A.
* Multiple operations are grouped.

However, that framework fails on the problem of checkpointing:
* We need to be able to checkpoint before executing an `IncomingBundle`. That execution could fail.
* Checkpointing the `ChainStateView` is not enough. One needs to checkpoint the state of the Wasm machine.
* So we use snapshots of the Wasm running instance.

Snapshots:
* We implement them for `wasmer` and `wasmtime`.
* They are small. Something like 1.12 M for matching engine.
* They are better than software solutions like introducing a `save` in the `linera-sdk`.
* They are super fact and do not cost any fuel. Doing a `load` / `save` cost 25k for a `counter` contract.
* They help enforce a strict alignment of the execution for the staging loop.

The execution on the web does not work because of the lack of shared memory (the same problem that forces the preloading of all relevant applications before running). So, we use a different scheme for the web:
* We implement the serialization for the Wasm snapshots.
* We use those snapshots when we execute a transaction.
* After a transaction, we wind down the Wasm execution instance.

With the changes:
* For a given application, `load` is called once inside a block, and `store` is called once. Slight break of the SDK.
* This is breaking the `testnet_conway`.

The flash loan application:
* Contract A calls the bank in order to get tokens.
* Contract B repays the bank in tokens.
* If the `terminate` function of the bank fails to terminate correctly then it means that the loans was not repaid correctly and the block cannot be created.

## Test Plan

CI should cover all scenarios.

A benchmark function for the `matching_engine` is introduced: 50 bids, 50 asks, all sent to the same application. The processing of the inbox is benchmarked. The result is 760ms for the old code and 413ms for the new code.

If accepted, then the documentation has to be updated since the `counter` application is changing.

## Release Plan

It cannot be backported to `testnet_conway` for the following reasons:
* The API of the contract is changed.
* The cost structure is changed.

## Links

None.


deuszx:
This looks promising but it'd be good to have some metric that we can use to confirm/gauge how much of the improvement we're getting.

MathieuDutSik:
> This looks promising but it'd be good to have some metric that we can use to confirm/gauge how much of the improvement we're getting.

During a simulation for the matching engine with one user creating 50 messages, another user 50 messages, and then measuring the `process_inbox` we are achieving a 2x speedup. More precisely 760.286667ms vs 380.791458ms.

However, that implementation over the block does not work for the checkpointing. This is because if we use persistent instances, then when accessing an older state, we have to access the older state. But we would lose the modification to the view.

I leave it here as a draft since it is possible to run without auto_retry with `BundleFailurePolicy::Abort`. So, it could be interesting in some context due to the massive potential benefit.

linera-witty/src/runtime/wasmtime/mod.rs
```diff
@@ -30,12 +32,79 @@ pub struct EntrypointInstance<UserData> {
...
     pub fn new(instance: wasmtime::Instance, store: Store<UserData>) -> Self {
         EntrypointInstance { instance, store }
     }
+
+    /// Creates a snapshot of the Wasm instance's mutable state (memory and globals).
+    pub fn create_snapshot(&mut self) -> WasmInstanceSnapshot {
+        let mut memory_data = Vec::new();
+        let mut globals = Vec::new();
+
+        let exports: Vec<_> = self
```
afck: `Vec<_>` annotation should be on `collect`.

linera-witty/src/runtime/wasmtime/mod.rs
```diff
@@ -30,12 +32,79 @@ pub struct EntrypointInstance<UserData> {
...
+
+        WasmInstanceSnapshot {
+            memory_data,
+            globals,
+        }
+    }
+
+    /// Restores the Wasm instance's mutable state from a snapshot.
+    pub fn restore_snapshot(&mut self, snapshot: &WasmInstanceSnapshot) {
+        let exports: Vec<_> = self
```
afck: (Type annotation)

linera-witty/src/runtime/wasmtime/mod.rs
```diff
@@ -30,12 +32,79 @@ pub struct EntrypointInstance<UserData> {
...
+                    memory_restored = true;
+                }
+                Extern::Global(global) => {
+                    if let Some((_, val)) = snapshot.globals.iter().find(|(n, _)| n == &name) {
+                        global
+                            .set(&mut self.store, *val)
+                            .expect("Failed to restore Wasm global from snapshot");
+                    }
+                }
+                _ => {}
```
afck: Should this be an error? Is it expected that we'd encounter multiple `Extern::Memory` instances?

linera-witty/src/runtime/wasmtime/mod.rs
```diff
@@ -30,12 +32,79 @@ pub struct EntrypointInstance<UserData> {
...
+            .instance
+            .exports(&mut self.store)
+            .map(|export| (export.name().to_string(), export.into_extern()))
+            .collect();
+
+        let mut memory_restored = false;
+        for (name, ext) in exports {
+            match ext {
+                Extern::Memory(mem) if !memory_restored => {
+                    mem.data_mut(&mut self.store)[..snapshot.memory_data.len()]
```
afck: Do we need to check here if the memory data is large enough?

linera-witty/src/runtime/wasmer/mod.rs
```diff
@@ -120,12 +120,77 @@ impl<UserData> AsStoreMut for EntrypointInstance<UserData> {
...
+            .iter()
+            .memories()
+            .next()
+            .map(|(_name, memory)| {
+                memory
+                    .view(&self.store)
+                    .copy_to_vec()
+                    .expect("Failed to copy Wasm memory")
+            })
+            .unwrap_or_default();
```
afck: Should it be an error if we don't have exactly one memory?

linera-witty/src/runtime/wasmtime/mod.rs
```diff
@@ -30,12 +32,79 @@ pub struct EntrypointInstance<UserData> {
...
+                        .copy_from_slice(&snapshot.memory_data);
+                    memory_restored = true;
+                }
+                Extern::Global(global) => {
+                    if let Some((_, val)) = snapshot.globals.iter().find(|(n, _)| n == &name) {
```
afck: Is it expected that they are in the same order as when we were taking the snapshot? If it is, maybe we should do something like `globals_iter.next()` instead of `find` here, and then check that the name matches?

If it's not expected, we could make `snapshot.globals` a map instead, so things can be looked up by name?

linera-witty/src/runtime/wasmer/mod.rs
```diff
@@ -120,12 +120,77 @@ impl<UserData> AsStoreMut for EntrypointInstance<UserData> {
...
+
+        WasmInstanceSnapshot {
+            memory_data,
+            globals,
+        }
+    }
+
+    /// Restores the Wasm instance's mutable state from a snapshot.
+    pub fn restore_snapshot(&mut self, snapshot: &WasmInstanceSnapshot) {
+        if let Some((_name, memory)) = self.instance.exports.iter().memories().next() {
```
afck: Same here: Should we check there's exactly one?

linera-sdk/tests/fixtures/flash-loan/src/contract.rs
```diff
@@ -0,0 +1,140 @@
...
+    /// so we check the accounting tracked in our own state. The actual token transfers
+    /// were performed via the fungible token during `execute_operation`.
+    async fn store(mut self) {
+        let total_borrowed = *self.state.total_borrowed.get();
+        let total_repaid = *self.state.total_repaid.get();
+        if total_borrowed > Amount::ZERO {
+            let params = self.runtime.application_parameters();
+            let interest_millionths = params.interest_millionths as u128;
+            let borrowed_attos = u128::from(total_borrowed);
+            let min_interest = Amount::from_attos(borrowed_attos * interest_millionths / 1_000_000);
```
afck: This could overflow.

linera-sdk/tests/fixtures/flash-loan/src/lib.rs
```diff
@@ -0,0 +1,43 @@
...
+    /// The fungible token application to use.
+    pub fungible_app_id: ApplicationId<fungible::FungibleTokenAbi>,
+    /// Interest in millionths (e.g. 10_000 = 1%).
+    pub interest_millionths: u64,
+}
+
+/// The initial state: how many tokens the pool starts with.
+#[derive(Debug, Deserialize, Serialize)]
+pub struct FlashLoanInitialState {
+    pub pool_balance: Amount,
```
afck: Do we even need this? Isn't it just how much of the underlying fungible token the app happens to have?

linera-service/tests/linera_net_tests.rs
```diff
@@ -1935,6 +1944,145 @@ async fn test_wasm_end_to_end_counter(config: impl LineraNetConfig) -> Result<()
     Ok(())
 }
 
+/// Tests that checkpoint/rollback with rejected bundles produces the same execution
+/// on proposer and validators. This is a regression test for a potential consensus bug
+/// where the proposer drops all Wasm instances after a rollback (causing a reload from
```
afck, line 1953: There isn't really a bug without this feature, is there?

Currently (without this PR), staging and final execution always agree, don't they?

linera-service/tests/linera_net_tests.rs
```diff
@@ -1935,6 +1944,145 @@ async fn test_wasm_end_to_end_counter(config: impl LineraNetConfig) -> Result<()
...
+
+    let (mut net, client1) = config.instantiate().await?;
+
+    let client2 = net.make_client().await;
+    client2.wallet_init(None).await?;
+
+    let chain_creator = client1.load_wallet()?.default_chain().unwrap();
+
+    // Generate keys for both clients and open a multi-owner receiver chain.
+    // Multi-owner avoids fast rounds, so the proposer's outcome is actually
```
afck: Shouldn't we compare it for fast blocks as well?

linera-service/tests/linera_net_tests.rs
```diff
@@ -1935,6 +1944,145 @@ async fn test_wasm_end_to_end_counter(config: impl LineraNetConfig) -> Result<()
...
+    // compared against the validator's re-execution.
+    let owner1 = client1.keygen().await?;
+    let owner2 = client2.keygen().await?;
+    let chain_receiver = client1
+        .open_multi_owner_chain(
+            chain_creator,
+            vec![(owner1, 100), (owner2, 100)].into_iter().collect(),
+            0, // no multi-leader rounds: forces single-leader rounds (no fast round)
```
afck: (Even if this were >0, there wouldn't be fast rounds. Multi-owner rounds are not the same as the fast round.)

linera-service/tests/linera_net_tests.rs
```diff
@@ -3741,6 +3889,427 @@ async fn test_wasm_end_to_end_matching_engine(config: impl LineraNetConfig) -> R
...
+
+    node_service_admin.ensure_is_running()?;
+    node_service_a.ensure_is_running()?;
+    node_service_b.ensure_is_running()?;
+
+    net.ensure_is_running().await?;
+    net.terminate().await?;
+
+    Ok(())
+}
```
afck: What do these two tests have to do with this?

linera-execution/src/lib.rs
```diff
@@ -485,6 +485,16 @@ pub trait UserContract {
 
     /// Finishes execution of the current transaction.
     fn finalize(&mut self) -> Result<(), ExecutionError>;
+
+    /// Creates a snapshot of the Wasm instance's mutable state (memory and globals).
+    ///
+    /// Returns `None` for non-Wasm contract implementations.
```
afck: I think this really shouldn't have a default implementation: It needs to work for all VMs.

(If we want to postpone e.g. the EVM implementation I'd rather add an error and a `TODO` comment in the EVM-specific code.)

linera-execution/src/lib.rs
```diff
@@ -485,6 +485,16 @@ pub trait UserContract {
...
+    fn create_snapshot(&mut self) -> Option<Box<dyn std::any::Any + Send>> {
+        None
+    }
+
+    /// Restores the Wasm instance's mutable state from a snapshot.
+    fn restore_snapshot(&mut self, _snapshot: &(dyn std::any::Any + Send)) {}
```
afck: Same here.

linera-execution/src/lib.rs
```diff
@@ -485,6 +485,16 @@ pub trait UserContract {
 
     /// Finishes execution of the current transaction.
     fn finalize(&mut self) -> Result<(), ExecutionError>;
+
+    /// Creates a snapshot of the Wasm instance's mutable state (memory and globals).
+    ///
+    /// Returns `None` for non-Wasm contract implementations.
+    fn create_snapshot(&mut self) -> Option<Box<dyn std::any::Any + Send>> {
```
afck: And why not make the snapshot an associated type instead of `Any`?

linera-chain/src/block_tracker.rs
```diff
@@ -58,10 +62,20 @@ pub struct BlockExecutionTracker<'resources, 'blobs> {
...
+    pub(crate) command_tx: Option<std::sync::mpsc::Sender<RuntimeCommand>>,
+    /// Receiver for state requests from the block-level contract runtime thread.
+    pub(crate) execution_state_receiver: Option<mpsc::UnboundedReceiver<ExecutionRequest>>,
 }
 
 impl<'resources, 'blobs> BlockExecutionTracker<'resources, 'blobs> {
     /// Creates a new BlockExecutionTracker.
+    ///
+    /// The `runtime_channels` argument is `Some` when a block-level contract runtime thread
+    /// is available (native). On web, WASM modules cannot be shipped to the runtime worker
```
```suggestion
    /// is available (native). On web, Wasm modules cannot be shipped to the runtime worker
```

linera-chain/src/block_tracker.rs
```diff
@@ -58,10 +62,20 @@ pub struct BlockExecutionTracker<'resources, 'blobs> {
...
+    /// via shared memory, so the block-level runtime is not spawned and `runtime_channels`
+    /// is `None` — per-transaction fallback runtimes are used instead.
```
afck: I don't understand; does that mean that in the web client, the staging loop _will_ produce a different outcome than the final block execution?

linera-witty/src/runtime/wasmtime/mod.rs
MathieuDutSik: Eliminated.

linera-witty/src/runtime/wasmtime/mod.rs
MathieuDutSik: Yes, and done systematically.

linera-execution/src/lib.rs
MathieuDutSik: I agree that we should not have a default implementation for the snapshots.

Well execution can fail for a block that contains REVM message. That is true as it is for Wasm.

But the difference is that REVM does not have snapshots. So, we save to storage even if we are not at the end of the block. So, the snapshots functionality becomes a noop.

linera-execution/src/lib.rs
MathieuDutSik: Yes, done similarly.

linera-service/tests/linera_net_tests.rs
MathieuDutSik: They were performance benchmarks.
The goal was to have performance benchmarks that were as near as possible to the `pm-app`.

linera-service/tests/linera_net_tests.rs
MathieuDutSik: Yes, they do.
They also agree on this PR.

linera-execution/src/lib.rs
MathieuDutSik: It is not easy.
The problem is that we have to store the snapshots of many possible execution engines: Wasmer, Wasmtime, REVM (that one is trivial).
We could have a type that encapsulates the `Box<dyn Any+Send>` but I am afraid there is no totally nice solutions.

linera-chain/src/block_tracker.rs
MathieuDutSik: Ok, corrected. 
Actually, the code somehow assumes that the execution engine is Wasm, which is not accurate. But that is already visible with the metrics.

linera-witty/src/runtime/wasmtime/mod.rs
MathieuDutSik: Yes, added a correction.

linera-witty/src/runtime/wasmtime/mod.rs
MathieuDutSik: For Linera, no since we have just one (the heap). However, there are use-cases where we have several, and maybe it would be better not to restruct ourselves initially.

linera-witty/src/runtime/wasmer/mod.rs
MathieuDutSik: I am not sure.
I added error handling to the snapshot operations. However, a priori we could have several memory on Wasm, so this looks like an unnecessary check. There are plenty of other possible failure scenarios.

linera-witty/src/runtime/wasmer/mod.rs
MathieuDutSik: I think no.

linera-witty/src/runtime/wasmtime/mod.rs
MathieuDutSik: Yes, switched to a BTreeMap.

linera-chain/src/block_tracker.rs
MathieuDutSik: That was the big problem. But I think I found a solution.

There are now indeed two execution paths:
* One for native where we have shared memory.
* One for the web.

This is forced because of the lack of shared memory in wasmer. The same problem that forces us to preload the contracts being used. There is a test that those two paths are coherent.

For the Native, we have persistent Wasm instance running over the whole execution of the block. For the web we create the Wasm instance during the scope of the execution. We work with snapshots in order to reinstate the Wasm instance when needed.

So, snapshots are used for both code execution paths, just in different ways.

afck: And using the web approach for all would be less efficient, would it? (Because we'd needlessly snapshot-and-restore.)

MathieuDutSik: Yes.

linera-sdk/tests/fixtures/flash-loan/src/lib.rs
MathieuDutSik: Yes, eliminated.

linera-sdk/tests/fixtures/flash-loan/src/contract.rs
MathieuDutSik: Corrected.

linera-service/tests/linera_net_tests.rs
MathieuDutSik: Ok, added a fast round test.

PR #6037, opened 2026-04-16
Add a command to query a validator for a certificate

bart-linera:
## Motivation

We could use a way of inspecting blocks without downloading whole chains from validators.

## Proposal

Add a command to the client which uses `download_certificates_by_height` to download a single certificate from a given validator and display the block.

## Test Plan

CI will catch possible regressions.
The command has been succesfully used manually.

## Release Plan

- These changes should be ported to the `main` branch

## Links

### DIFF
--- a/linera-service/src/cli/validator.rs
+++ b/linera-service/src/cli/validator.rs
@@ -9,10 +9,15 @@ use anyhow::Context as _;
 use futures::stream::TryStreamExt as _;
 use linera_base::{
     crypto::{AccountPublicKey, ValidatorPublicKey},
+    data_types::BlockHeight,
     identifiers::ChainId,
 };
 use linera_client::{chain_listener::ClientContext as _, client_context::ClientContext};
-use linera_core::{data_types::ClientOutcome, node::ValidatorNodeProvider, Wallet as _};
+use linera_core::{
+    data_types::ClientOutcome,
+    node::{ValidatorNode, ValidatorNodeProvider},
+    Wallet as _,
+};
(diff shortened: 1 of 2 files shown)


linera-service/src/cli/validator.rs
```diff
@@ -685,6 +711,37 @@ impl Query {
...
+        let node = context.make_node_provider().make_node(&self.address)?;
+        let chain_id = self.chain_id.unwrap_or_else(|| context.default_chain());
+        let height = self.height;
+        println!(
+            "Querying validator about the certificate for height {height} on the chain \
+            {chain_id}.\n"
+        );
+
+        let result = node
+            .download_certificates_by_heights(chain_id, vec![height])
```
afck, line 728: (I think that processes/preprocesses the block, does it? It's probably fine, but we could also just download, print and forget it. Not sure what's better.)

linera-service/src/cli/validator.rs
```diff
@@ -685,6 +711,37 @@ impl Query {
...
+            .await;
+
+        match result {
+            Ok(certificates) => {
+                let block = certificates[0].block();
+                println!("{:#?}", block);
```
afck, line 734: Should it be called `query-block` then?
Because this specifically prints only the block and not the signatures or even signatories.

afck (approved):

linera-service/src/cli/validator.rs
bart-linera: I don't think it does, it seems to only call the method on the GRPC client as far as I can tell, but I may be wrong.

PR #6040, opened 2026-04-16
Limit the number of stream IDs per `PreviousEventBlocks` request.

afck:
## Motivation

`PreviousEventBlocks` requests often time out.

## Proposal

Add a configurable `--max-event-stream-queries` option (default 1000) that chunks large `PreviousEventBlocks` queries into multiple smaller requests to avoid timeouts on the server side.

## Test Plan

CI
(The added logic is trivial.)

## Release Plan

- Release SDK.

## Links

### DIFF
--- a/linera-client/src/client_options.rs
+++ b/linera-client/src/client_options.rs
@@ -15,7 +15,8 @@ use linera_base::{
 use linera_core::{
     client::{
         chain_client, DEFAULT_CERTIFICATE_DOWNLOAD_BATCH_SIZE,
-        DEFAULT_CERTIFICATE_UPLOAD_BATCH_SIZE, DEFAULT_SENDER_CERTIFICATE_DOWNLOAD_BATCH_SIZE,
+        DEFAULT_CERTIFICATE_UPLOAD_BATCH_SIZE, DEFAULT_MAX_STREAM_QUERIES,
+        DEFAULT_SENDER_CERTIFICATE_DOWNLOAD_BATCH_SIZE,
     },
     node::CrossChainMessageDelivery,
     DEFAULT_QUORUM_GRACE_PERIOD,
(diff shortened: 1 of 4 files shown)


linera-core/src/client/mod.rs
```diff
@@ -127,6 +127,7 @@ mod metrics {
 pub static DEFAULT_CERTIFICATE_DOWNLOAD_BATCH_SIZE: u64 = 500;
 pub static DEFAULT_CERTIFICATE_UPLOAD_BATCH_SIZE: u64 = 500;
 pub static DEFAULT_SENDER_CERTIFICATE_DOWNLOAD_BATCH_SIZE: usize = 20_000;
+pub static DEFAULT_MAX_STREAM_QUERIES: usize = 1000;
```
deuszx: IDK whether that's a lot or not?

afck: I'd say it's not: It's a lookup for 1000 entries in a `StreamId` → `BlockHeight` map.

linera-client/src/client_options.rs
```diff
@@ -230,6 +231,11 @@ pub struct Options {
     #[arg(long, default_value = "100")]
     pub max_joined_tasks: usize,
 
+    /// Maximum number of event stream IDs to include in a single `PreviousEventBlocks`
+    /// request. Larger sets are split into multiple requests.
+    #[arg(long, default_value_t = DEFAULT_MAX_STREAM_QUERIES)]
+    pub max_stream_queries: usize,
```
ma2bd: The name of the option is confusing: we already have an unrelated `max_stream_queries` option in `linera_views`

ma2bd: (fixed)

PR #6042, opened 2026-04-16
Remove redundant admin chain sync from synchronize_publisher_chains

ndr-ds:
## Motivation

`synchronize_publisher_chains` unconditionally calls `synchronize_chain_state(admin_chain_id)` every time it runs. Its only caller, `synchronize_from_validators`, always runs `prepare_chain` first — which already detects epoch mismatches and syncs the admin chain when needed. The unconditional sync is redundant and causes unnecessary quorum-wide validator queries when multiple chains are processed in sequence.

## Proposal

Remove the unconditional admin chain sync from `synchronize_publisher_chains`. The `admin_chain_id` local variable is kept for the publisher chain filter.

## Test Plan

CI

## Release Plan

## Links

- Related: #5989

### DIFF
--- a/linera-core/src/client/chain_client/mod.rs
+++ b/linera-core/src/client/chain_client/mod.rs
@@ -849,12 +849,8 @@ impl<Env: Environment> ChainClient<Env> {
                     .insert(stream_id.clone());
             }
         }
-        // Always fully sync the admin chain for epoch changes.
-        let admin_chain_id = self.client.admin_chain_id;
-        if admin_chain_id != self.chain_id {
-            self.client.synchronize_chain_state(admin_chain_id).await?;
-        }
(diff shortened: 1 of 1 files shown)


ma2bd:
wow

afck:
Not sure I understand: `prepare_chain` only syncs the admin chain if our _own_ chain is already in a higher epoch. If we haven't accepted the new epoch in our chain yet, we won't sync the admin chain, and then `synchronize_publisher_chains` could encounter a published event from a later epoch and fail.
(Maybe that's okay, though?)

ndr-ds:
@afck Sorry, I sent this PR out and was gonna ask if you guys thought it made sense. Seems like it doesn't 😅  
Will close this then

PR #6050, opened 2026-04-17
Linera bridge relayer runbook

deuszx:
## Motivation

Provision a dedicated docker-compose stack for running a long-lived
`linera-bridge` relayer on a testnet VM, plus the operational runbook
the team needs to deploy, monitor, back up, and troubleshoot it.

## Proposal

- New compose file `docker/docker-compose.bridge-testnet.yml`: single
  `relayer` service with bind-mounted `/var/lib/linera-bridge` data dir,
  `/etc/linera-bridge/.env*` env files, journald logging, loopback-only
  metrics + health, `restart: unless-stopped`, conservative CPU/memory
  limits.
- `docker/bridge-entrypoint.sh`: align the default `--port` (3001) with
  the Dockerfile and the testnet compose port mapping.
- `linera-bridge`: add a `/health` route to the relay HTTP server. The
  testnet compose healthcheck uses a bare TCP probe (only `nc` ships in
  the runtime image); the route is there for downstream collectors that
  prefer HTTP probes.
- `docker/README.testnet.md`: deployment runbook — directory layout,
  env file schema, startup, routine ops, backup, observability table
  matching the metrics emitted by `linera-bridge/src/relay/metrics.rs`,
  Prometheus alert rules (gas balance, permanent failures, throughput
  stall), and a troubleshooting matrix. SQL examples updated for the
  new `pending_*`/`finished_*` schema.

## Test Plan

- CI builds the bridge image and parses the compose file.
- Local: `make demo` still works end-to-end (this PR does not touch
  `setup.sh` or the local docker-compose).
- `docker compose -f docker/docker-compose.bridge-testnet.yml config`
  parses with both the testnet defaults (`/etc/linera-bridge/...`,
  `/var/lib/linera-bridge/...`) and operator-supplied paths via env-var
  overrides where applicable.
- Manual: bring the compose up, verify `nc -z localhost 3001` succeeds,
  `curl -sI http://localhost:3001/health` returns `HTTP/1.1 200 OK`,
  `/metrics` exposes the metrics listed in the Observability table.

## Release Plan

- Nothing to do

## Links

### DIFF
--- a/docker/README.testnet.md
+++ b/docker/README.testnet.md
@@ -3,6 +3,9 @@
 This describes how to provision and operate a `linera-bridge` relayer
 on a single VM, bridging Base Sepolia (EVM) and Testnet Conway (Linera).
 
+For the full design rationale, see [`SPEC-linera-bridge-testnet-deployment.md`](../SPEC-linera-bridge-testnet-deployment.md)
+in the repo root.
+
 ## Prerequisites
 
 - A Linux VM with Docker + docker-compose
@@ -38,7 +41,17 @@ sudo chown root:root /etc/linera-bridge/.env.secret
 
 ### 3. Provision contracts and Linera apps
 
+Run setup.sh as the `linera-bridge` user so the wallet/keystore/storage
+files are owned by the same uid that the container runs as. The
+`--relayer-env` output goes to `/etc/linera-bridge/.env` which is root-
+writable, so allow that one path to the linera-bridge user via sudo or
+write the env file to a temporary location and `sudo cp` it into place.
+The simplest path:
+
 ```bash
+# Make /etc/linera-bridge writable by linera-bridge for this one provision
+sudo chown linera-bridge:linera-bridge /etc/linera-bridge
+sudo -u linera-bridge bash <<'EOF'
 cd examples/bridge-demo
 ./setup.sh \
   --evm-rpc-url https://sepolia.base.org \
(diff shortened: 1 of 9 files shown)


docker/README.testnet.md
```diff
@@ -0,0 +1,218 @@
+# Linera Bridge — Testnet Deployment Runbook
+
+This describes how to provision and operate a `linera-bridge` relayer
+on a single VM, bridging Base Sepolia (EVM) and Testnet Conway (Linera).
+
+For the full design rationale, see [`SPEC-linera-bridge-testnet-deployment.md`](../SPEC-linera-bridge-testnet-deployment.md)
```
ndr-ds, line 6: This is referencing a non existing file

docker/docker-compose.bridge-testnet.yml
```diff
@@ -0,0 +1,36 @@
...
+#   1. Create /var/lib/linera-bridge (data) and /etc/linera-bridge (env)
+#   2. Run examples/bridge-demo/setup.sh in direct mode with
+#      --relayer-env /etc/linera-bridge/.env to populate both directories
+#   3. Drop the EVM signing key in /etc/linera-bridge/.env.secret
+#      (mode 0600, root:root): EVM_PRIVATE_KEY=0x...
+#
+# See docker/README.testnet.md for the full procedure.
+services:
+  relayer:
+    image: linera-bridge:${BRIDGE_TAG:-testnet_conway}
```
ndr-ds, line 13: `testnet_conway` tag is updated every time a PR merges in that branch, right?
We should have a tag by default here that doesn't move. Otherwise, if the pull policy is `Always`, any pod restart will pull the latest image instead of the one that was previously running.
That's why we use `testnet_conway_internal` and `testnet_conway_release`, etc for the validator images, we shouldn't deploy a moving tag ever

docker/docker-compose.bridge-testnet.yml
```diff
@@ -0,0 +1,36 @@
...
+# Prerequisites (operator must do once on the VM):
+#   1. Create /var/lib/linera-bridge (data) and /etc/linera-bridge (env)
+#   2. Run examples/bridge-demo/setup.sh in direct mode with
+#      --relayer-env /etc/linera-bridge/.env to populate both directories
+#   3. Drop the EVM signing key in /etc/linera-bridge/.env.secret
+#      (mode 0600, root:root): EVM_PRIVATE_KEY=0x...
+#
+# See docker/README.testnet.md for the full procedure.
+services:
+  relayer:
```
ndr-ds, line 12: Should we add some reasonable resource limits here?

ndr-ds: And when this is taken down, the default `stop_grace_period` is 10s. If we feel like that's not enough, we could increase it to 30s maybe or something

docker/docker-compose.bridge-testnet.yml
deuszx: Maybe we should start tagging releases w/ versions. Then, here, use the specific version via `BRIDGE_TAG` env.

docker/docker-compose.bridge-testnet.yml
deuszx: Done in https://github.com/linera-io/linera-protocol/pull/6050/changes/a2e1b0e505b2c1f14570ee25f6e5a27e41b2e8a9

docker/README.testnet.md
deuszx: Done in https://github.com/linera-io/linera-protocol/pull/6050/changes/a2e1b0e505b2c1f14570ee25f6e5a27e41b2e8a9

docker/docker-compose.bridge-testnet.yml
ndr-ds: Another thing is that we should use the full image path `us-docker.pkg.dev/linera-io-dev/linera-public-registry/linera-bridge:testnet_conway`. But we also shouldn't deploy the `testnet_conway` image, because that moves on every push. We should create a `testnet_conway_internal` or something here as well that we control. So this should probably default to `us-docker.pkg.dev/linera-io-dev/linera-public-registry/linera-bridge:testnet_conway_internal` or something like that

deuszx: Updated to `us-docker.pkg.dev/linera-io-dev/linera-public-registry/linera-bridge:testnet_conway_internal`

ndr-ds (approved):

PR #6053, opened 2026-04-18
Poison and evict chain worker on any save failure

### LINKED ISSUE #6052: Review impacts of potential error-throwing successful DB writes
* We already identified a scenario with Journaling where DB writes could fail to complete but were already persisted in the journal. This is covered by the state `poisoned` bool.
* #6051 introduced cache invalidation before DB writes: this means that if the DB returns an error, the LRU cache doesn't assume that the DB write didn't occur.

However, this scenario (say timeout after a write is successful) is not fully covered: in case of DB error (except for journaling), the current code calls `rollback()` instead of reloading the view.

ma2bd:
## Motivation

Fixes #6052.

Previously only journal resolution failures poisoned the chain worker. Other write errors (e.g. possible ScyllaDB timeouts) left the worker loaded with potentially stale state.

Stacked on #6051.

## Proposal

- **Poison on any save failure**: `ChainWorkerState::save()` now always sets `poisoned = true` and returns `WorkerError::PoisonedWorker` on any error, triggering immediate eviction and a fresh reload on the next request.

- **Keep `must_reload_view`** for metrics and as way to document that journaling really requires a notion of view poisoning.

## Test Plan

CI

## Links

- Fixes #6052
- Stacked on #6051

### DIFF
--- a/linera-core/src/chain_worker/state.rs
+++ b/linera-core/src/chain_worker/state.rs
@@ -2143,15 +2143,13 @@ where
     ))]
     async fn save(&mut self) -> Result<(), WorkerError> {
         if let Err(error) = self.chain.save().await {
-            if error.must_reload_view() {
-                tracing::error!(
-                    ?error,
-                    chain_id = %self.chain_id(),
-                    "Journal resolution failed; marking worker as poisoned"
-                );
-                self.poisoned = true;
-            }
-            return Err(error.into());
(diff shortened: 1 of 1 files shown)


afck (approved):

ma2bd:
shoot I closed on the wrong branch. Will finish manually

PR #5907, opened 2026-04-02
Add opt-in recovery from inbox gaps and execution outcome mismatch

afck:
Port of https://github.com/linera-io/linera-protocol/pull/5876, https://github.com/linera-io/linera-protocol/pull/5881, https://github.com/linera-io/linera-protocol/pull/5882, https://github.com/linera-io/linera-protocol/pull/5886, https://github.com/linera-io/linera-protocol/pull/5896, https://github.com/linera-io/linera-protocol/pull/5890 and https://github.com/linera-io/linera-protocol/pull/5900.

## Motivation

On the testnet, we added mechanisms to recover from failures to persist inbox updates and from `IncorrectOutcome` errors. These should not be reachable in theory, other than due to database corruptions, but it might still be useful to keep these recovery mechanisms.

## Proposal

Add `RevertConfirm` cross-chain request to recover from state inconsistencies where a recipient chain lost persisted inbox state after a confirmation was sent. When enabled via `--allow-revert-confirm`, the recipient detects inbox gaps and requests the sender to re-add outbox entries and resend bundles.

Also add `--reset-on-incorrect-outcome-mins` to reset and re-execute a chain's entire block history when an `IncorrectOutcome` error is detected (guarded by a minimum cooldown to prevent loops).

## Test Plan

Some tests were added.

## Release Plan

- Nothing to do / These changes follow the usual release cycle.

## Links

- Testnet PRs: https://github.com/linera-io/linera-protocol/pull/5876, https://github.com/linera-io/linera-protocol/pull/5881, https://github.com/linera-io/linera-protocol/pull/5882, https://github.com/linera-io/linera-protocol/pull/5886, https://github.com/linera-io/linera-protocol/pull/5896, https://github.com/linera-io/linera-protocol/pull/5890, https://github.com/linera-io/linera-protocol/pull/5900
- [reviewer checklist](https://github.com/linera-io/linera-protocol/blob/main/CONTRIBUTING.md#reviewer-checklist)

PR #6063, opened 2026-04-20
Reset chains on more errors; restore votes

afck:
## Motivation

In https://github.com/linera-io/linera-protocol/pull/5907 we added options to recover from DB failures, specifically from corrupted chain states that would result in `IncorrectOutcome` errors.

However, corrupted chain states can cause all kinds of errors, most of which are currently the `InternalError` variant.

## Proposal

Split `InternalError`: Use `CorruptedChainState` for those errors that could be caused by such DB issues.

Rename `--reset-on-incorrect-outcome-mins` to `--reset-on-corrupted-chain-state-mins` and, if set, reset the chain state on every `CorruptedChainState` error.

Also, restore the votes after re-executing the chain, to prevent double-signing.

## Test Plan

The test was extended.

But we will have to see if this fixes the remaining issues we see in `testnet_conway`, particularly validator 4.

## Release Plan

- Possibly port to `main`?

## Links

- The first version of the recovery mechanisms: #5907

### DIFF
--- a/linera-core/src/chain_worker/state.rs
+++ b/linera-core/src/chain_worker/state.rs
@@ -969,27 +969,6 @@ where
                 .await?;
             // We should always agree on the messages and state hash.
             if outcome != verified {
-                if let Some(min_duration) = self.config.reset_on_incorrect_outcome {
-                    let block_zero_time = *self.chain.block_zero_executed_at.get();
-                    let elapsed = local_time.duration_since(block_zero_time);
-                    if elapsed >= min_duration {
-                        warn!(
-                            "IncorrectOutcome for block {height} on chain {chain_id}; \
-                            resetting chain state and re-executing"
-                        );
-                        return self
-                            .reset_and_reexecute_chain(
-                                certificate,
-                                notify_when_messages_are_delivered,
-                            )
-                            .await;
-                    }
-                    warn!(
-                        "IncorrectOutcome for block {height} on chain {chain_id}; \
-                        not resetting because only {elapsed:?} elapsed since last \
-                        block 0 execution (minimum: {min_duration:?})"
-                    );
-                }
                 return Err(WorkerError::IncorrectOutcome {
                     submitted: Box::new(outcome),
                     computed: Box::new(verified),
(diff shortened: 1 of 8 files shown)


linera-core/src/worker.rs
```diff
@@ -839,11 +838,51 @@ where
         if let Err(error) = &result {
             if error.must_reload_view() {
                 self.evict_poisoned_worker(chain_id, &state);
+            } else if error.indicates_corrupted_chain_state() {
+                Box::pin(wrap_future(
+                    self.reset_corrupted_chain_state(chain_id, &state),
```
deuszx: Do we need to make sure this task runs to completion without interruption (cancelation)? Otherwise we can end up in a weird state after restore.

afck: Good point!
The state wouldn't be weird, it would just be at a lower height. But we'd fail to restore the votes, so I'll `spawn` this…

deuszx: Should we also mark this chain worker as "unavailable" ? Ideally the task would be blocking, at the chain actor level.

**EDIT:**
OK, I can see you took a different approach – i.e. reusing the previously-existing process of `handle_`-ing the requests. I was wondering whether we won't end up in a weird state where we're re-executing past blocks while still being open to new requests.

linera-core/src/chain_worker/state.rs
```diff
@@ -1278,16 +1259,39 @@ where
...
+        };
+        let chain_id = self.chain_id();
+        let local_time = self.storage.clock().current_time();
+        let block_zero_time = *self.chain.block_zero_executed_at.get();
+        let elapsed = local_time.duration_since(block_zero_time);
+        if elapsed < min_duration {
+            warn!(
+                "Not resetting corrupted chain state for {chain_id} because only \
+                {elapsed:?} elapsed since last block 0 execution (minimum: {min_duration:?})"
+            );
```
deuszx, line 1280: Can you rewrite to the pattern we both know but is too long to describe?

linera-core/src/unit_tests/worker_tests.rs
```diff
@@ -4741,6 +4752,28 @@ where
...
+    // retry because `apply_confirmed_block` for cert_1 would otherwise wipe the
+    // manager during the course of the retry.
+    {
+        let chain = worker_with_recovery.chain_state_view(chain_id).await?;
+        let restored_vote = chain.manager.confirmed_vote.get().as_ref();
+        assert_eq!(
+            restored_vote.map(|v| v.signature),
+            Some(planted_vote.signature),
+            "confirmed_vote should be restored after reset-and-reexecute"
+        );
```
deuszx, line 4774: I can see this test asserts that _previously-casted vote_ is still present but what I don't see from this black-box test is that this vote was ever removed. 

A clearer test would be one where we see that right after the recovery process is spawned no vote is present and then, after the restoration we get the old one. Ideally we'd also try to force the validator (using the "public API"/normal operation) to sign something else in the meantime (while recovery process is running).

linera-core/src/chain_worker/state.rs
deuszx: This and other logs affected in this PR.

linera-core/src/worker.rs
afck: The re-executing task will hold the write lock, yes. Not sure if we should try and do something more than that, like return errors for all incoming requests… they might get handled, in some cases where the chain is short and re-execution is fast.

linera-core/src/unit_tests/worker_tests.rs
afck: But the recovery process has the write lock. Normally nothing else can read or write the chain state while it's in progress, and when it's done, the vote will be in place again.

linera-core/src/worker.rs
afck: Done in https://github.com/linera-io/linera-protocol/pull/6063/commits/3ea8df69cea2b7d5ef7192fb30fdc4554ff4ed85.

linera-core/src/chain_worker/state.rs
afck: Done in https://github.com/linera-io/linera-protocol/pull/6063/commits/915a84cb18b38b7226d2ea7fc5431c8c3f721f4d.

deuszx (approved):

PR #6064, opened 2026-04-20
Storage atomicity and cancel safety fixes. (#6056, #6046, #6053)

afck:
Ports of #6056, #6046 and #6053.

## Motivation

* The LRU cache is not cancel-safe: If the task is canceled, new values could have been written to the store, with the old values still in the cache.
* The block cache uses `Weak` pointers, so inserting and then getting doesn't work: After inserting there is no strong reference so the weak one becomes invalid.
* We currently reload the chain state only on specific journaling errors. But it's not clear that's the only error that can cause the view to become inconsistent.

## Proposal

* `spawn` the task that updates the LRU cache and writes to the underlying storage.
* Use the `Arc` returned from `insert` instead of dropping it.
* Reload the chain state on all `save()` errors.

## Test Plan

CI

## Release Plan

## Links

- `testnet_conway` versions: #6056, #6046, #6053

### DIFF
--- a/linera-core/src/chain_worker/state.rs
+++ b/linera-core/src/chain_worker/state.rs
@@ -2109,15 +2109,13 @@ where
     ))]
     async fn save(&mut self) -> Result<(), WorkerError> {
         if let Err(error) = self.chain.save().await {
-            if error.must_reload_view() {
-                tracing::error!(
-                    ?error,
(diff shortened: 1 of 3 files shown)


MathieuDutSik (approved):
LGTM

ma2bd (approved):

PR #6071, opened 2026-04-20
[testnet dbg] Reset chains on more errors; restore votes (#6063)

afck:
Backport of https://github.com/linera-io/linera-protocol/pull/6063.

## Motivation

In https://github.com/linera-io/linera-protocol/pull/5907 we added options to recover from DB failures, specifically from corrupted chain states that would result in `IncorrectOutcome` errors.

However, corrupted chain states can cause all kinds of errors, most of which are currently the `InternalError` variant.

## Proposal

Split `InternalError`: Use `CorruptedChainState` for those errors that could be caused by such DB issues.

Rename `--reset-on-incorrect-outcome-mins` to `--reset-on-corrupted-chain-state-mins` and, if set, reset the chain state on every `CorruptedChainState` error.

Also, restore the votes after re-executing the chain, to prevent double-signing.

## Test Plan

The test was extended.

But we will have to see if this fixes the remaining issues we see in `testnet_conway`, particularly validator 4.

## Release Plan

- Possibly port to `main`?

## Links

- `testnet_conway` version: https://github.com/linera-io/linera-protocol/pull/6063
- The first version of the recovery mechanisms: #5907

### DIFF
--- a/linera-core/src/worker.rs
+++ b/linera-core/src/worker.rs
@@ -1146,18 +1146,29 @@ where
                     return;
                 }
             };
-            let Some(sender) = &this.outbound_cross_chain_sender else {
-                if !requests.is_empty() {
-                    warn!(
-                        %chain_id, dropped = requests.len(),
-                        "No outbound cross-chain sender installed; \
-                        RevertConfirms after corruption reset are dropped"
-                    );
+            if let Some(sender) = &this.outbound_cross_chain_sender {
+                // Sharded validator path: let the RPC layer route each request to
+                // the shard that owns the target chain.
+                for request in requests {
+                    sender(request);
                 }
-                return;
-            };
-            for request in requests {
-                sender(request);
+            } else {
+                // No routing dispatcher is installed (client node or test), so all
+                // involved chains are co-located on this worker. Dispatch locally
+                // in a loop, following any cascading cross-chain requests.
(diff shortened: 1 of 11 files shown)


linera-core/src/worker.rs
```diff
@@ -1080,7 +1079,67 @@ where
...
+                    error!(
+                        %chain_id, %error,
+                        "Chain worker actor dropped the reset request"
+                    );
+                    return;
+                }
+            };
+            let mut queue = VecDeque::from(requests);
+            while let Some(request) = queue.pop_front() {
+                match this.handle_cross_chain_request(request).await {
```
ma2bd: @afck How does this work? We're not dispatching requests to the network but instead handling them in the same worker, but validators are sharded.

afck: Fixed now… but the fix is somewhat ugly.

deuszx: hmm.. I thought they're sharded by chain so we can assume we're on the worker which has access to the necessary data for this chain.. ah, the sender chain and receiver chain may be on different shards.

afck: These are the `RevertConfirm`s that get sent to all chains that ever sent anything to our reverted chain.

afck:
Closing this, since we moved to `testnet_conway`.

PR #6072, opened 2026-04-20
Sleep before cleaning up in linera net helper; re-enable readme tests

bart-linera:
## Motivation

Readme tests were disabled in the merge queue due to their flakiness.

## Proposal

The flakiness seems to have been due to a race condition between shutting down `linera` processes and a cleanup `rm` call. Delay the call a bit to reduce the chances of this race condition happening.

## Test Plan

Hard to test directly due to inherent non-determinism in test flakiness.
Future CI runs will show if the problem persists.

## Release Plan

- Nothing to do.

## Links

### DIFF
--- a/.github/workflows/test-readmes.yml
+++ b/.github/workflows/test-readmes.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches: [ 'devnet_*', 'testnet_*' ]
   pull_request:
+  merge_group:
   workflow_dispatch:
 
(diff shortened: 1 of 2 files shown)


afck (approved):

afck:
> Nothing to do

I'd maybe even backport it? (We don't use the merge queue on `testnet_conway`, but it's nice to have this fixed in case we want to run the readme tests.)

PR #6076, opened 2026-04-21
Pin the version of `protoc` in CI.

MathieuDutSik:
## Motivation

The `setup-protoc` is often failing in CI.

## Proposal

Pin the version of `protoc` in the CI to the latest one. That should enable the caching.

Also pin the version of `jetli/wasm-bindgen-action` which was pinned to `latest` in one CI check and `0.2.100` in another.

## Test Plan

CI

## Release Plan

## Links

None


deuszx (approved):
Let's give it a shot.

PR #6079, opened 2026-04-21
Address issues with default chains.

### LINKED ISSUE #4091: Set last chain in wallet as default
After removing a chain from the wallet, none of the chains are set to default chain. When there's only one chain left, set that to default automatically.

MathieuDutSik:
## Motivation

When there is just one chain, set it as default.

Fixes https://github.com/linera-io/linera-protocol/issues/4091

## Proposal

Address the issue.
Also resolve a bug when we remove a chain that happens to have been the default.

## Test Plan

CI

## Release Plan

Backport to `testnet_conway` seems reasonable.

## Links

None

### DIFF
--- a/linera-wallet-json/src/wallet.rs
+++ b/linera-wallet-json/src/wallet.rs
@@ -182,28 +182,6 @@ impl PersistentWallet {
             .ok_or_else(|| anyhow::anyhow!("keypair not found for chain `{chain_id}`"))
     }
 
-    pub fn forget_chain(&self, chain_id: ChainId) -> anyhow::Result<Chain> {
-        let chain = self
-            .0
-            .chains
-            .remove(chain_id)
-            .ok_or_else(|| anyhow::anyhow!("nonexistent chain `{chain_id}`"))?;
-        {
-            let mut default = self.0.default.write().unwrap();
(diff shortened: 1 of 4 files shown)


deuszx (commented):
What's the difference between `remove(chain_id)` and `forget(chain_id)`?

MathieuDutSik:
> What's the difference between `remove(chain_id)` and `forget(chain_id)`?

None.
Maybe we could remove it as well.

ma2bd:
I commented in #4091. Should we instead politely decline to remove the default chain? (say unless --force is given)

linera-wallet-json/src/wallet.rs
```diff
@@ -80,6 +80,12 @@ impl PersistentWallet {
             if *default == Some(id) {
                 *default = None;
             }
+            if default.is_none() {
+                let items = self.0.chains.items();
+                if items.len() == 1 {
```
ma2bd, line 85: what happens if we remove the default chain while `len() > 2`

MathieuDutSik: There is no default chain, which is the current behavior.

afck (approved):

PR #6080, opened 2026-04-21
Route corruption-recovery `RevertConfirms` to correct shard

afck:
## Motivation

https://github.com/linera-io/linera-protocol/pull/6063 introduced a bug, that was only found in the review of the `testnet_conway_debug` port https://github.com/linera-io/linera-protocol/pull/6071:

Calling `handle_cross_chain_request` in a loop on the worker only works for single-shard configurations; in sharded validators the target sender chain may live on a different shard. 

## Proposal

Add an `OutboundCrossChainSender` that the RPC server layer (gRPC and simple) installs on `WorkerState`, closing over the shard-routing network config and the cross-chain `mpsc` sender. `spawn_reset_corrupted_chain_state` now dispatches each `RevertConfirm` through that sender so `forward_cross_chain_queries` can route it to the correct shard. When no dispatcher is installed (client nodes, tests) we keep the existing local loop, which is correct because all involved chains are co-located on the same worker.

## Test Plan

CI

## Release Plan

## Links

### DIFF
--- a/linera-core/src/worker.rs
+++ b/linera-core/src/worker.rs
@@ -604,7 +604,17 @@ pub struct WorkerState<StorageClient: Storage> {
     /// manages its own lifetime via a keep-alive task. A background sweep
     /// periodically removes dead entries.
     chain_workers: ChainWorkerMap<StorageClient>,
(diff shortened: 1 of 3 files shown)


deuszx (approved):
elegant

PR #6083, opened 2026-04-21
Some update the the linera-explorer-new for "authenticated-signer" / "authenticated_owner"

MathieuDutSik:
## Motivation

In `linera-explorer-new`, we are using `authenticated-signer` when it is `authenticated-owner` in the rest of the code.

## Proposal

Do the straightforward renaming.

## Test Plan

CI

## Release Plan

Things are different on `testnet_conway`. So, no backport.

## Links

None


deuszx (approved):
Thanks for catching this

PR #6085, opened 2026-04-21
Poison chain worker on cancelled `save`.

afck:
## Motivation

If the caller's future is dropped between `write_batch` landing in storage and `post_save` refreshing the in-memory register views, the views (e.g. `stored_value`) goes stale while storage has advanced.

## Proposal

Set the poison flag up front and clear it only after `save`.

That way, if a request was cancelled while `save` was in progress, the chain worker is reloaded for the next request.

## Test Plan

CI

I will also run a few more tests on https://github.com/linera-io/linera-protocol/pull/6070.

## Release Plan

- Backport to `testnet_conway`.
- Hotfix validators.

## Links

- Fixes #6084
- Draft PR for experimenting with this: #6070 (easier to reproduce in CI)
- [reviewer checklist](https://github.com/linera-io/linera-protocol/blob/main/CONTRIBUTING.md#reviewer-checklist)

PR #6086, opened 2026-04-21
[testnet] Poison chain worker on cancelled save. (#6085)

afck:
Backport of #6085.

## Motivation

If the caller's future is dropped between `write_batch` landing in storage and `post_save` refreshing the in-memory register views, the views (e.g. `stored_value`) goes stale while storage has advanced.

## Proposal

Set the poison flag up front and clear it only after `save`.

That way, if a request was cancelled while `save` was in progress, the chain worker is reloaded for the next request.

## Test Plan

CI

I will also run a few more tests on https://github.com/linera-io/linera-protocol/pull/6070.

## Release Plan

- Hotfix validators.

## Links

- PR to `main`: #6085
- Draft PR for experimenting with this: #6070 (easier to reproduce in CI)

### DIFF
--- a/linera-core/src/chain_worker/state.rs
+++ b/linera-core/src/chain_worker/state.rs
@@ -2151,20 +2151,25 @@ where
 
     /// Stores the chain state in persistent storage.
     ///
-    /// If the save fails, the worker is marked as poisoned and must be reloaded.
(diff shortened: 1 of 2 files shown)


MathieuDutSik (approved):
LGTM

PR #6089, opened 2026-04-21
Shared committee map

afck:
## Motivation

We are using a lot of memory for the `committees` field in the system execution state view, because it is basically duplicated once for every chain worker.

## Proposal

Use a shared map.

The only reason to look into the map is for epoch revocation, which we don't do on this testnet anyway. So we avoid that and always trust old epochs.

We also disallow chain info queries that would force validators to load the map.

## Test Plan

Affected tests were updated.

## Release Plan

- Port some of this to `main`, where applicable. (Even though `main` needs a different solution in the long term.)
- Hotfix validators
- Release SDK

## Links

### DIFF
--- a/linera-chain/src/chain.rs
+++ b/linera-chain/src/chain.rs
@@ -1020,16 +1020,12 @@ where
         }
 
         let mandatory_apps_need_accepted_message = self
-            .execution_state
-            .system
             .current_committee()
             .await?
-            .is_some_and(|(_epoch, committee)| {
-                committee
-                    .policy()
-                    .http_request_allow_list
-                    .contains(FLAG_MANDATORY_APPS_NEED_ACCEPTED_MESSAGE)
-            });
+            .1
+            .policy()
+            .http_request_allow_list
+            .contains(FLAG_MANDATORY_APPS_NEED_ACCEPTED_MESSAGE);
         Self::check_app_permissions(
             self.execution_state
                 .system
--- a/linera-chain/src/chain.rs
+++ b/linera-chain/src/chain.rs
@@ -8,7 +8,7 @@ use std::{
 
 use allocative::Allocative;
 use linera_base::{
-    crypto::{AccountPublicKey, CryptoHash, ValidatorPublicKey},
+    crypto::{CryptoHash, ValidatorPublicKey},
     data_types::{
         ApplicationDescription, ApplicationPermissions, ArithmeticError, Blob, BlockHeight, Epoch,
         OracleResponse, Timestamp,
--- a/linera-chain/src/chain.rs
+++ b/linera-chain/src/chain.rs
@@ -499,18 +499,7 @@ where
         // Recompute the state hash.
         let hash = self.execution_state.crypto_hash_mut().await?;
         self.execution_state_hash.set(Some(hash));
-        let fallback_owners: Vec<(AccountPublicKey, u64)> =
-            match self.execution_state.system.current_committee().await? {
-                Some((_, committee)) => committee.account_keys_and_weights().collect(),
-                None => Vec::new(),
-            };
-        // Last, reset the consensus state based on the current ownership.
-        self.manager.reset(
-            self.execution_state.system.ownership.get().await?.clone(),
-            BlockHeight(0),
-            local_time,
-            fallback_owners.into_iter(),
-        )?;
+        self.reset_chain_manager(BlockHeight(0), local_time).await?;
         Ok(())
     }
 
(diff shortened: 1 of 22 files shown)


linera-core/src/chain_worker/state.rs
```diff
@@ -801,37 +801,25 @@ where
...
-        }
+            .context()
+            .extra()
+            .get_committees(epoch..=epoch)
+            .await
+            .with_execution_context(ChainExecutionContext::Block)?
+            .remove(&epoch)
+            .ok_or_else(|| {
+                ChainError::InternalError(format!(
+                    "missing committee for epoch {epoch}; this is a bug"
```
afck, line 819: ~Actually, if we're just _preprocessing_ this block, this isn't a bug! (Independent of this PR, just highlighting it here so we don't forget.)~

(Sorry, this is correct: `get_committees` above would already return the correct error; this here should be unreachable.)

linera-execution/src/lib.rs
```diff
@@ -631,6 +634,41 @@ pub trait ExecutionRuntimeContext {
...
+    ) -> Result<Option<Arc<Committee>>, ExecutionError> {
+        if let Some(committee) = self.shared_committees().get(epoch) {
+            return Ok(Some(committee));
+        }
+        let Some(net_description) = self.get_network_description().await? else {
+            return Ok(None);
+        };
+        let blob_hash = if epoch.0 == 0 {
+            net_description.genesis_committee_blob_hash
+        } else {
```
ma2bd, line 653: Should we return `Ok(None)` when `epoch` exceeds the local epoch of the chain?

linera-chain/src/chain.rs
```diff
@@ -499,13 +499,21 @@ where
...
         let hash = self.execution_state.crypto_hash_mut().await?;
         self.execution_state_hash.set(Some(hash));
-        let maybe_committee = self.execution_state.system.current_committee().into_iter();
+        let fallback_owners: Vec<(AccountPublicKey, u64)> = match self
+            .execution_state
+            .system
+            .current_committee()
+            .await?
+        {
+            Some((_, committee)) => committee.account_keys_and_weights().collect(),
```
ma2bd, line 508: nit: This `collect()` is a bit of a regression that could be prevented with `Itertools::Either` (or `Box` I suppose).

linera-chain/src/chain.rs
```diff
@@ -1271,14 +1282,20 @@ where
...
         local_time: Timestamp,
     ) -> Result<(), ChainError> {
-        let maybe_committee = self.execution_state.system.current_committee().into_iter();
+        let fallback_owners: Vec<(AccountPublicKey, u64)> = match self
+            .execution_state
+            .system
+            .current_committee()
+            .await?
+        {
+            Some((_, committee)) => committee.account_keys_and_weights().collect(),
```
ma2bd, line 1291: same. Should we have a method `current_account_keys_and_weights` in `SystemExecutionStateView` ?

linera-core/src/chain_worker/state.rs
```diff
@@ -801,37 +801,34 @@ where
...
+                    .extra()
+                    .get_committees(epoch..=epoch)
+                    .await
+                    .map_err(|error| {
+                        ChainError::ExecutionError(Box::new(error), ChainExecutionContext::Block)
+                    })?
+                    .remove(&epoch)
+                    .ok_or_else(|| {
+                        ChainError::InternalError(format!(
+                            "missing committee for epoch {epoch}; this is a bug"
```
ma2bd, line 825: Are we sure it is a bug? (`epoch` comes from the yet-unverified `block.header.epoch`)

ma2bd: Perhaps the classification as InternalError is a bit of a security hazard.

linera-execution/src/lib.rs
```diff
@@ -555,82 +555,59 @@ pub trait ExecutionRuntimeContext {
...
+        Ok(committees)
     }
 
+    /// Returns the committee for `epoch`, consulting the shared cache first. On a miss,
+    /// loads the `NewCommittee` event and the committee blob from storage and memoizes
+    /// the result. Returns `Ok(None)` if the network description, the event, or the
+    /// blob is not available locally.
+    ///
+    /// Determinism during block execution: call sites inside the runtime must only ask
+    /// for epochs up to and including the chain's current epoch (`self.epoch.get()`).
```
ma2bd, line 600: seems like we could just enforce this invariant

ma2bd (approved):
LGTM

linera-execution/src/lib.rs
afck: How? The execution runtime _context_ can't read the view it's the context _of_, can it?

linera-core/src/chain_worker/state.rs
afck: It would be a bug: `get_committees` would have returned the appropriate error if we didn't have the epoch. But if it didn't return an error, the returned map must contain the requested epoch.

linera-execution/src/lib.rs
afck: (Same as below: the context doesn't know the chain's epoch.)

PR #6090, opened 2026-04-21
Make chain worker writes cancel-safe

afck:
## Motivation

Caller cancellation can land between `write_batch` and `post_save`: the storage write completes while in-memory views go unrefreshed, leaving them out of sync.

## Proposal

Run the write work on a detached task so the `RollbackGuard` lives inside the spawn and the lock is held until the DB round-trip and `post_save` have both finished. Other tasks — including a freshly loaded replacement worker — only observe the committed state.

Also:
- Add `linera_base::task::run_detached` abstracting `tokio::task::spawn` on native and `wasm_bindgen_futures::spawn_local` on web.
- Apply the same treatment to `read_lock_initialized`.
- Drop the narrower LRU cache spawn from https://github.com/linera-io/linera-protocol/pull/6056, which is subsumed.
- Bump `linera-faucet-server` recursion_limit to fit the deeper trait  resolution introduced by the new `Send` bounds on `chain_write`.

## Test Plan

CI

The ScyllaDB tests should now not be flaky anymore.

## Release Plan

- Backport to `testnet_conway`
- Release SDK
- Hotfix validators

## Links

### DIFF
--- a/linera-base/src/task.rs
+++ b/linera-base/src/task.rs
@@ -7,6 +7,40 @@ Abstractions over tasks that can be used natively or on the Web.
 
 use futures::{future, Future, FutureExt as _};
 
+/// Spawns `future` on the runtime and awaits its completion.
+///
+/// Unlike [`Task::spawn`], dropping the returned future does *not* cancel the
+/// spawned task — it runs to completion in the background. Use this when the
+/// spawned work (e.g. a storage write paired with its in-memory finalization)
+/// must not be torn apart mid-flight by caller cancellation.
+#[cfg(not(web))]
+pub async fn run_detached<F, R>(future: F) -> R
+where
+    F: Future<Output = R> + Send + 'static,
+    R: Send + 'static,
+{
+    tokio::task::spawn(future)
+        .await
+        .unwrap_or_else(|e| std::panic::resume_unwind(e.into_panic()))
+}
+
+/// Web variant: runs `future` on the event loop via
+/// [`wasm_bindgen_futures::spawn_local`], delivering the output via a oneshot.
+/// The spawned task is inherently detached from any handle on web.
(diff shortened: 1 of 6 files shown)


linera-base/src/task.rs
```diff
@@ -7,6 +7,56 @@ Abstractions over tasks that can be used natively or on the Web.
...
+pub async fn run_detached<F, R>(future: F) -> R
+where
+    F: Future<Output = R> + MaybeSend + 'static,
+    R: MaybeSend + 'static,
+{
+    let (tx, rx) = futures::channel::oneshot::channel();
+    wasm_bindgen_futures::spawn_local(async move {
+        let _ = tx.send(future.await);
+    });
+    rx.await
```
ma2bd, line 56: Do we need this?

ma2bd: @Twey says yes

linera-base/src/task.rs
```diff
@@ -25,34 +25,20 @@ impl<T> MaybeSend for T {}
...
+/// finalization) must not be torn apart mid-flight by caller cancellation.
 pub async fn run_detached<F, R>(future: F) -> R
 where
     F: Future<Output = R> + MaybeSend + 'static,
     R: MaybeSend + 'static,
 {
     let (tx, rx) = futures::channel::oneshot::channel();
-    wasm_bindgen_futures::spawn_local(async move {
+    Task::spawn(async move {
         let _ = tx.send(future.await);
```
ma2bd: Maybe replace `let _` by `if let Err(e) = ... { error!(..) }` ?

ma2bd: Is the oneshot channel desirable in the non-web case?

ma2bd (approved):

PR #6093, opened 2026-04-22
Introduce persistent execution over an `IncomingBundle`

MathieuDutSik:
## Motivation

The execution of an incoming bundle over a specified bundle can be parallelized by having persistent
Wasm instances over the execution of the bundle.

## Proposal

When we execute the bundle, we have to deal with the following scenario:
```
Message to application A.
Message to Application B that executes B
```

So, we need to also change the way that the execution is done.

## Test Plan

CI

## Release Plan

A release to `testnet_conway` would be breaking.

## Links

None

### DIFF
--- a/linera-chain/src/block_tracker.rs
+++ b/linera-chain/src/block_tracker.rs
@@ -4,6 +4,7 @@
 use std::collections::{BTreeMap, BTreeSet};
 
 use custom_debug_derive::Debug;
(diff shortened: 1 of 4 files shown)


MathieuDutSik:
The benefit of this approach appear modest compared with https://github.com/linera-io/linera-protocol/pull/6032
Therefore, I close it for the time being.

PR #6095, opened 2026-04-22
`web`: clone full client instead of just `ClientContext`

Twey:
## Motivation

@deuszx noticed that we don't clone the full client into the subobjects of `@linera/web`, which means that even though the `Application` is still alive sometimes the chain listener can be killed prematurely, resulting in the client not receiving any more notifications.

## Proposal

Clone the whole `Client` into subobjects, not just the `linera_client::ClientContext`.

## Test Plan

Hard to test because it relies on GC behaviour.  Open to suggestions.

## Release Plan

    - be released in a new SDK,

## Links

### DIFF
--- a/web/@linera/client/src/lib.rs
+++ b/web/@linera/client/src/lib.rs
@@ -107,8 +107,3 @@ pub fn initialize(options: Option<InitializeOptions>) {
         )
         .init();
 }
-
-#[wasm_bindgen]
-pub fn panic() {
-    panic!("oh no!");
-}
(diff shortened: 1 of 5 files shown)


deuszx (approved):
Looks good. Worth testing with native-fungible.

web/@linera/client/src/lib.rs
```diff
@@ -107,3 +107,8 @@ pub fn initialize(options: Option<InitializeOptions>) {
         )
         .init();
 }
+
+#[wasm_bindgen]
+pub fn panic() {
+    panic!("oh no!");
```
ma2bd: lol. What is this?

ma2bd: I guess it is/was used for testing?

Twey: :facepalm: yes it's from testing some other work I was doing — must have snuck in via the stash.  Removed.

PR #6118, opened 2026-04-23
Add certificate size and signer-count metrics to worker

ndr-ds:
## Motivation

Today we have no observability into certificate size or signer-count distributions. The KV-layer size histogram in `linera-views/src/backends/metering.rs` observes every value written to storage, but mixes cert bytes with every other KV write (chain state, blobs, events) and is not filterable. There is no metric that answers "how big are the certificates we are actually processing?"

Two reasons this is worth closing now:

**1. Regression monitoring for silent cert-size growth.** Certificate size is a direct input to cross-chain bandwidth and ScyllaDB storage cost. Changes to block structure, signature layout, or cert-carried payloads can silently grow certs (the #5944 to #6046 return-semantics refactor is the kind of change where an unnoticed size drift would cost real money). A byte-size histogram surfaces such regressions in Grafana instead of letting them show up as mystery latency or storage bills weeks later.

**2. Quantitative baseline for the eventual post-quantum signature migration.** Naive PQ signature migration inflates per-validator signatures 10x (Falcon-512) to 38x (ML-DSA-44). The real bandwidth and storage impact on Linera depends on two distributions that are currently unobserved:
- how big the signature component is relative to the cert payload
- how many validators sign each cert in practice

Without these, we would be projecting PQ cost from a back-of-envelope estimate (signers x 64 bytes). With these, we can project from the measured distributions and make the Category 1 vs Category 3 scheme decision on data rather than intuition. Same two metrics cover both use cases.

## Proposal

Adds two `HistogramVec` metrics in `linera-core/src/worker.rs` under the existing `#[cfg(with_metrics)]` gate:

- `linera_certificate_bytes{kind}` — BCS-serialized certificate size in bytes, bucketed exponentially from 128 B to 256 KiB
- `linera_certificate_signer_count{kind}` — number of validator signatures per certificate, bucketed exponentially from 1 to 100

Both are labelled by `kind` in `{confirmed, validated, timeout}` so the three certificate flavors can be separated in queries.

Observation is via a helper `metrics::observe_certificate(&cert, kind)` invoked at the top of the three worker ingress handlers (`handle_confirmed_certificate`, `handle_validated_certificate`, `handle_timeout_certificate`). Serialized size uses `bcs::serialized_size`, which walks the value without allocating an intermediate buffer, so the hot-path cost is minimal.

No new dependencies. Uses the existing `linera_base::prometheus_util` helpers and the `with_metrics` feature gate, matching every other metric in this module.

## Test Plan

CI.

## Release Plan

## Links

### DIFF
--- a/linera-core/src/worker.rs
+++ b/linera-core/src/worker.rs
@@ -102,7 +102,10 @@ mod metrics {
         exponential_bucket_interval, register_histogram, register_histogram_vec,
         register_int_counter, register_int_counter_vec,
     };
-    use linera_chain::{data_types::MessageAction, types::ConfirmedBlockCertificate};
+    use linera_chain::{
+        data_types::MessageAction,
+        types::{CertificateValue, ConfirmedBlockCertificate, GenericCertificate},
+    };
     use prometheus::{Histogram, HistogramVec, IntCounter, IntCounterVec};
 
     pub static NUM_ROUNDS_IN_CERTIFICATE: LazyLock<HistogramVec> = LazyLock::new(|| {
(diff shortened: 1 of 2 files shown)


ma2bd (approved):

linera-core/src/worker.rs
```diff
@@ -181,6 +184,45 @@ mod metrics {
...
+    });
+
+    pub static CERTIFICATE_SIGNER_COUNT: LazyLock<HistogramVec> = LazyLock::new(|| {
+        register_histogram_vec(
+            "certificate_signer_count",
+            "Number of validator signatures attached to each processed certificate. \
+             Combined with certificate_bytes, lets the signature component be isolated \
+             from the payload component.",
+            &["kind"],
+            exponential_bucket_interval(1.0, 100.0),
```
deuszx: We know we have <20 validators and don't plan on having much more for now. It's more important to know – for now at least – whether we have 3 or 5 signatures rather than 1/2/4/8.. etc

linera-core/src/worker.rs
```diff
@@ -181,6 +184,45 @@ mod metrics {
         )
     });
 
+    pub static CERTIFICATE_BYTES: LazyLock<HistogramVec> = LazyLock::new(|| {
+        register_histogram_vec(
+            "certificate_bytes",
+            "Serialized size of certificates processed by the worker, in bytes. \
+             Labelled by certificate kind so confirmed/validated/timeout can be separated.",
+            &["kind"],
+            exponential_bucket_interval(128.0, 262_144.0),
```
deuszx: This is only 262kB (0.2MB) which is quite small, isn't it? I'd expect our certs are mor than that.

ma2bd: let's find out! Blobs are not part of the certs, though.

PR #6122, opened 2026-04-24
Shared committee map (2nd attempt)

afck:
## Motivation

#6089 introduced a shared committee map, but was reverted due to a bug:

`ChainClient::local_committee` catches `BlobsNotFound` but ignores `chain_info()` returning `EventsNotFound`.

## Proposal

Catch it and sync the admin chain, so the required committee is available.

(Also: Avoid some more unnecessary committee cloning.)

## Test Plan

A regression test was added.

## Release Plan

- The fix should be ported to `main`.
- Release SDK.

## Links

- First attempt: #6089


ma2bd (approved):
nice!

PR #6134, opened 2026-04-26
Main clippy synchronization

MathieuDutSik:
## Motivation

Revert some of the clippy corrections in order to have coherence between `testnet_conway` and `main`.

## Proposal

The entries were removed from the `Cargo.toml`. The changes to the `linera-sdk` were reverted in order
to have the same API.

## Test Plan

CI

## Release Plan

This is done in https://github.com/linera-io/linera-protocol/pull/5982

## Links

None


linera-sdk/src/contract/runtime.rs
```diff
@@ -424,8 +424,8 @@ where
     }
 
     /// Creates a new data blob and returns its hash.
-    pub fn create_data_blob(&mut self, bytes: &[u8]) -> DataBlobHash {
-        let hash = contract_wit::create_data_blob(bytes);
+    pub fn create_data_blob(&mut self, bytes: Vec<u8>) -> DataBlobHash {
```
afck, line 427: (That one really did make sense… anyway, we can revisit that later. Feels like we should slowly move towards `needless_pass_by_value`, without enforcing it everywhere yet.)

afck (approved):

linera-sdk/src/contract/runtime.rs
MathieuDutSik: It totally made sense, but it is also breaking the `linera-sdk`.

PR #6135, opened 2026-04-27
Add panicker example

deuszx:


deuszx:
Seems like we're not interested in having this example.

PR #6136, opened 2026-04-27
Remove unused code, fix naming.

afck:
## Motivation

* A constant is imported with an alias for no reason.
* `Clock::sleep` and `Block::outcome_matches` are unused.
* `#[allow(…)]` attributes don't alert us if they're unneeded.

## Proposal

* Import `EPOCH_STREAM_NAME` without an alias.
* Remove unused code.
* Use `#[expect(…)]`; remove it if it fails.

## Test Plan

CI

## Release Plan

## Links


deuszx:
> * `Block::outcome_matches` are unused.

Was this really unused? B/c it seems like you just moved the `assert!(outcome_matches(_))` to `assert_outcome_matches!` macro.

deuszx (approved):

afck:
> Was this really unused?

It was unused in production, but not marked test-only.

deuszx:
> > Was this really unused?
> 
> It was unused in production, but not marked test-only.

It was marked with ` #[cfg(with_testing)]`

afck:
Ah, sorry, you're right!
Anyway, I still find it cleaner if that code lives in a test-only _module_ then.

MathieuDutSik (approved):

PR #5984, opened 2026-04-13
Migrate Docker image builds from Cloud Build to self-hosted OVH runners

ndr-ds:
## Motivation

Docker image builds OOM during Rust LTO linking on GCP Cloud Build's `e2-highcpu-32` (32
vCPU, 32 GB RAM). Cloud Build also costs ~$700/mo at current volume. The CI cluster
already has OVH c3-64 nodes (32 vCPU, 64 GB RAM) — same CPU count, 2x the RAM, ~10x
cheaper per build-hour.

## Proposal

**`docker_image.yml` — build locally on the runner instead of Cloud Build:**
- Runs on `linera-io-self-hosted-builder` (dedicated ARC runner with 56 GB DinD memory,
deployed via linera-io/linera-infra#893)
- Replaces `gcloud builds submit` with `docker build` + `docker push` directly on the
runner's DinD sidecar
- Adds `gcloud auth configure-docker` step for Artifact Registry authentication
- Keeps parallel builds (5 images as background processes, logs collected per-image)
- Fixes injection risk: `github.head_ref` / `github.ref_name` moved to `env:` block
- Timeout increased to 60 min to account for DinD overhead
- Adds Docker layer pruning step to prevent disk ustion

**`build_image.yml` — new `workflow_dispatch` workflow:**
- Replaces `gcloud builds submit` as the generic "submit a build" interface
- Anyone with repo access can trigger via `gh workflow run` — no kubeconfig or gcloud
needed
- Inputs: `image_type` (main/indexer/explorer/exporter/bridge/all), `image_tag`,
`build_mode` (release/debug), `ref` (git ref to build)
- Used by `lineractl build-image` (migrated in linera-io/linera-infra#893)
- All user inputs passed via `env:` blocks for injection safety

## Test Plan

After merge, trigger a test build:
```bash
gh workflow run build_image.yml --repo linera-io/linera-protocol --ref main \
-f image_type=exporter -f image_tag=test-builder -f build_mode=release
Then test the main image (the one that was OOMing):
gh workflow run build_image.yml --repo linera-io/linera-protocol --ref main \
-f image_type=main -f image_tag=test-builder -f build_mode=release
Verify images appear in Artifact Registry with correct tags. Once irmed, backport to
testnet_conway to unblock the broken docker_image workflow.

Release Plan

- These changes should be backported to the latest testnet branch, then
- be released in a validator hotfix.

PR #6138, opened 2026-04-27
Build release binaries on host for Docker Compose CI to avoid in-container OOM

### LINKED ISSUE #5909: Docker Compose CI: OOM during in-container release build
`ci-compose.sh` builds binaries from scratch inside Docker with `cargo build --release` (`lto=thin`), which OOM-kills `rustc` during LTO linking (signal 9, SIGKILL).

The Dockerfile already supports `--build-arg binaries=<path>` to skip in-container compilation.

**Fix:** Build release binaries on the host (which has enough memory), then pass them to Docker via `--build-arg binaries=target/release`. Also increase ScyllaDB `start_period` from `60s` to `120s` in `docker/docker-compose.yml` for CI runners that are slow to initialize.

**Evidence:**
- [Run 23879379363](https://github.com/linera-io/linera-protocol/actions/runs/23879379363) — `rustc` killed at ~664s during LTO linking
- [Run 23867309806](https://github.com/linera-io/linera-protocol/actions/runs/23867309806) — `rustc` killed at ~617s during LTO linking

**Files:** `docker/ci-compose.sh`, `.github/workflows/docker-compose.yml`, `docker/docker-compose.yml`

ndr-ds:
## Motivation

The Docker Compose CI workflow OOM-kills `rustc` during thin-LTO linking when the Dockerfile rebuilds release binaries from scratch inside the DinD container on `linera-io-self-hosted-ci`. The runner host has 64 GB of RAM, but the DinD container is memory-limited and a `cargo build --release` of `linera-server` with thin-LTO does not fit. Closes #5909.

The workflow today also wastes work: step `Build client binary` already runs a `cargo install --debug` on the host before `docker build` discards it and recompiles in release inside the container.

Evidence: runs [23879379363](https://github.com/linera-io/linera-protocol/actions/runs/23879379363) and [23867309806](https://github.com/linera-io/linera-protocol/actions/runs/23867309806) — `rustc` killed at ~664s and ~617s during LTO linking.

## Proposal

Use the existing `binaries` build-arg in `docker/Dockerfile` to skip the in-container compilation entirely:

- `.github/workflows/docker-compose.yml`: in step `Build client binary`, drop `--debug`, add `--bin linera-proxy` and `--features scylladb,metrics,jemalloc` so the host build matches the Dockerfile, then stage the three binaries from `target/release/` into `docker/binaries/` (the docker build context, since `target/` is in `.dockerignore`). On the `Run Compose` step, set `LINERA_BINARIES=docker/binaries` and `DOCKER_BUILDKIT=1` (the latter matches the convention in `build_image.yml` and `docker_image.yml` and ensures BuildKit prunes the unused `builder` stage out of the build graph; classic builder would still execute it and re-OOM).
- `docker/ci-compose.sh`: when `LINERA_BINARIES` is set, pass `--build-arg binaries=$LINERA_BINARIES` to `docker build`. Default (local dev) behavior is byte-for-byte unchanged — running `./ci-compose.sh` with no env var still triggers the in-container build.

Net CI work: one release build on the host instead of debug-host + release-LTO-DinD, so this should be strictly faster than the path we last saw pass, not just unstuck.

## Test Plan

CI.

## Release Plan

## Links

- Closes #5909
- Prior art: #5984 — same OOM class on a different workflow, addressed by routing to a runner pool with more DinD memory.

### DIFF
--- a/.github/workflows/docker-compose.yml
+++ b/.github/workflows/docker-compose.yml
@@ -43,6 +43,10 @@ jobs:
         with:
           version: "34.1"
           repo-token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Install libunwind-dev
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y libunwind-dev
       # TODO(#2709): Uncomment once this workflow runs in a custom runner
       # - name: Update aio-max-nr
       # run: echo 1048576 > /proc/sys/fs/aio-max-nr
--- a/.github/workflows/docker-compose.yml
+++ b/.github/workflows/docker-compose.yml
@@ -48,13 +48,20 @@ jobs:
       # run: echo 1048576 > /proc/sys/fs/aio-max-nr
       - name: Build client binary
         run: |
-          cargo install --path linera-service --bin linera --bin linera-server --debug --locked
+          cargo install --path linera-service \
+            --bin linera --bin linera-proxy --bin linera-server \
+            --features scylladb,metrics,jemalloc --locked
+          mkdir -p docker/binaries
+          cp target/release/linera target/release/linera-proxy target/release/linera-server docker/binaries/
       # TODO(#2709): Remove this step once this workflow runs in a custom runner
       - name: Patch Docker compose file to run ScyllaDB in developer mode
         run: |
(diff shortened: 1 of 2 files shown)


ma2bd (approved):

ndr-ds:
Closing — this approach doesn't actually work cleanly. Going with a DinD memory bump in linera-infra instead.

PR #6141, opened 2026-04-27
Add test of coherence of execution outcome.

MathieuDutSik:
## Motivation

The client works by executing the proposal twice: Once with all `IncomingBundle`  and another one with the failing `IncomingBundle`s marked as rejected.
We did not check whether that was coherent.

## Proposal

Add a simple check for the coherence of the outcome of both executions.

## Test Plan

CI

The scenario that this test intends to catch should not occur in `main` but did occur in other work on Block execution.

## Release Plan

This could be backported to `testnet_conway`.

## Links

None

### DIFF
--- a/linera-core/src/client/chain_client/mod.rs
+++ b/linera-core/src/client/chain_client/mod.rs
@@ -328,6 +328,12 @@ pub enum Error {
          The committed certificate hash is {0}"
     )]
     Conflict(CryptoHash),
+
+    #[error(
+        "Execution outcome mismatch: AutoRetry and committed execution produced \
+         different outcomes for the same block"
+    )]
+    ExecutionOutcomeMismatch,
 }
 
 impl From<Infallible> for Error {
@@ -1508,10 +1514,11 @@ impl<Env: Environment> ChainClient<Env> {
                 skipped.insert(origin);
             }
         }
-        let (proposed_block, _) = block.clone().into_proposal();
+        let (proposed_block, auto_retry_outcome) = block.clone().into_proposal();
         *proposal_guard = Some(PendingProposal {
             block: proposed_block,
             blobs,
+            auto_retry_outcome: Some(auto_retry_outcome),
         });
         Ok(block)
     }
@@ -1930,6 +1937,7 @@ impl<Env: Environment> ChainClient<Env> {
             // Otherwise we are free to propose our own pending block.
             let proposed_block = pending.block.clone();
             let blobs = pending.blobs.clone();
+            let auto_retry_outcome = pending.auto_retry_outcome.clone();
             let round = self.round_for_oracle(&info, &owner).await?;
             let (block, _, _) = self
                 .client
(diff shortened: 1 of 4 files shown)


linera-core/src/client/chain_client/mod.rs
```diff
@@ -1913,6 +1920,7 @@ impl<Env: Environment> ChainClient<Env> {
             // Otherwise we are free to propose our own pending block.
             let proposed_block = pending.block.clone();
             let blobs = pending.blobs.clone();
+            let auto_retry_outcome = pending.auto_retry_outcome.clone();
```
afck: Do we really need to clone for this? Isn't a reference enough?

linera-core/src/client/chain_client/mod.rs
```diff
@@ -1923,6 +1931,16 @@ impl<Env: Environment> ChainClient<Env> {
                     BundleExecutionPolicy::committed(),
                 )
                 .await?;
+            // Sanity check: the committed execution should produce the same outcome
+            // as the initial AutoRetry execution. A mismatch indicates a fuel
+            // accounting divergence between the two execution paths.
+            if let Some(auto_retry_outcome) = auto_retry_outcome {
+                let (_, committed_outcome) = block.clone().into_proposal();
+                ensure!(
+                    auto_retry_outcome == committed_outcome,
```
afck: Maybe we need to bring back `outcome_matches` for this. I removed it in https://github.com/linera-io/linera-protocol/pull/6136, but here it might help us avoid cloning the `block`.

linera-core/src/client/chain_client/mod.rs
```diff
@@ -1923,6 +1931,16 @@ impl<Env: Environment> ChainClient<Env> {
                     BundleExecutionPolicy::committed(),
                 )
                 .await?;
+            // Sanity check: the committed execution should produce the same outcome
+            // as the initial AutoRetry execution. A mismatch indicates a fuel
+            // accounting divergence between the two execution paths.
```
afck: It's not necessarily fuel-related.

MathieuDutSik: Yes, corrected.

linera-core/src/client/chain_client/mod.rs
```diff
@@ -1913,6 +1920,7 @@ impl<Env: Environment> ChainClient<Env> {
             // Otherwise we are free to propose our own pending block.
             let proposed_block = pending.block.clone();
             let blobs = pending.blobs.clone();
+            let auto_retry_outcome = pending.auto_retry_outcome.as_ref();
```
afck: (Maybe let's call it `staging_outcome`?)

afck (approved):
Looks good, but I'd rename "auto-retry" to "staging" here.

linera-core/src/client/chain_client/mod.rs
MathieuDutSik: No, we can avoid it.

linera-core/src/client/chain_client/mod.rs
MathieuDutSik: Ok, done for this use case.

linera-core/src/client/chain_client/mod.rs
MathieuDutSik: Ok.

PR #6146, opened 2026-04-28
Backport gRPC metrics observability fixes (#6119 + #6120)

ndr-ds:
## Motivation

Backport of two gRPC metrics observability fixes from main into testnet_conway, so the same diagnostics are available on the live testnet validators:

- linera-io/linera-protocol#6119: raise gRPC latency histogram ceiling from 2s to 5s
- linera-io/linera-protocol#6120: record cancelled gRPC requests via Drop guard on server middleware

Both fixes are required to fully diagnose the timeout-storm pattern observed on testnet_conway 2026-04-18 (~40% of `PreviousEventBlocks` requests silently dropped from server-side metrics during timeout spikes; ~80% of those requests piled up in the `+Inf` latency bucket because the histogram capped at 2000ms).

## Proposal

Cherry-pick #6119 and #6120 onto testnet_conway. No conflicts; the diff is identical to the sum of the two main PRs.

- `linera-rpc/src/grpc/server.rs`: histogram ceiling 2s → 5s + new `server_request_cancelled` counter + Drop-based cancellation guard
- `linera-service/src/proxy/grpc.rs`: histogram ceiling 2s → 5s

## Test Plan

CI. Same plan as the original PRs — post-deploy on testnet_conway, validate that during a timeout storm `proxy_request_count` approximately equals `server_request_count + server_request_cancelled` per method.

## Release Plan

## Links

- Backports: linera-io/linera-protocol#6119, linera-io/linera-protocol#6120

### DIFF
--- a/linera-rpc/src/grpc/server.rs
+++ b/linera-rpc/src/grpc/server.rs
@@ -94,6 +94,14 @@ mod metrics {
         )
     });
 
+    pub static SERVER_REQUEST_CANCELLED: LazyLock<IntCounterVec> = LazyLock::new(|| {
+        register_int_counter_vec(
+            "server_request_cancelled",
+            "Server requests whose handler future was dropped before completion (e.g. client-side timeout / disconnect)",
+            &[METHOD_NAME_LABEL, TRAFFIC_TYPE_LABEL],
+        )
+    });
+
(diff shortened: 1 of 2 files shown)


ndr-ds:
Clean backport, I'm just gonna merge

PR #6147, opened 2026-04-28
Push Arc deeper through certificate and block paths

ndr-ds:
## Motivation

Follow-up to #5944. The previous PR added Arc-based caching, but consumers still
called `Arc::unwrap_or_clone` at many sites because function signatures took
owned values. This PR pushes Arc deeper through those signatures and consolidates
the worker-side block cache shape.

## Proposal

- `Storage::read_blob`, `read_confirmed_block`, `read_certificate` and friends
  now return `Arc<...>` instead of cloning out of the cache.
- `ChainWorkerState::block_values` and `WorkerState::block_cache` now store
  `Arc<ValueCache<CryptoHash, ConfirmedBlock>>` (previously `Hashed<Block>`),
  so cache hits return the consumer-shaped `Arc<ConfirmedBlock>` directly with
  no per-hit conversion.
- `ValueCache::insert_hashed` is now generic: `insert_hashed<T>(Cow<Hashed<T>>)`
  on `ValueCache<CryptoHash, V>` where `V: From<Hashed<T>>`.
- Added `impl From<Hashed<Block>> for ConfirmedBlock` and `ValidatedBlock`.
- `WorkerState::process_confirmed_block` keeps owned `ConfirmedBlockCertificate`
  to satisfy the `chain_write` static-lifetime closure introduced in #6090; the
  inner `ChainWorkerState::process_confirmed_block` takes `&cert` so internal
  callers (recovery paths) skip clones entirely.
- `Client::receive_sender_certificate` now takes `Arc<ConfirmedBlockCertificate>`
  so the local-storage hit path in `download_sender_block_with_sending_ancestors`
  and `download_event_bearing_blocks` keeps the Arc end-to-end (saves a full cert
  clone per cached hit).
- Removed an `Arc::unwrap_or_clone` and a redundant full-cert clone in
  `chain_client::handle_certificates` by keeping `Arc<Cert>` through the
  iteration.

Net: workspace-wide `Arc::unwrap_or_clone` calls drop from 38 to 21 (-45%). The
remaining 21 are at API boundaries (network/RPC wire format, async-graphql
output, `LiteCertificate::with_value`, oracle response variants, test
utilities) where eliminating them would cascade across trait or wire contracts.

## Test Plan

CI.

## Release Plan

- Should be backported to `testnet_conway`

## Links

- Builds on #5944

### DIFF
--- a/linera-base/src/data_types.rs
+++ b/linera-base/src/data_types.rs
@@ -1383,6 +1383,12 @@ impl From<Blob> for BlobContent {
     }
 }
 
+impl From<Arc<Blob>> for BlobContent {
+    fn from(blob: Arc<Blob>) -> BlobContent {
+        blob.content().clone()
+    }
+}
+
 /// A blob of binary data, with its hash.
 #[derive(Debug, Hash, PartialEq, Eq, Clone, Allocative)]
 pub struct Blob {
--- a/linera-cache/src/value_cache.rs
+++ b/linera-cache/src/value_cache.rs
@@ -77,7 +77,6 @@ where
     }
 
     /// Inserts a pre-wrapped `Arc<V>` into the cache, returning the canonical `Arc`.
-    #[cfg(with_testing)]
     pub fn insert_arc(&self, key: &K, value: Arc<V>) -> Arc<V> {
         self.dedup_insert(key, value)
     }
@@ -204,14 +203,16 @@ where
     }
 }
 
-impl<T: Send + Sync + 'static> ValueCache<CryptoHash, Hashed<T>> {
-    /// Inserts a [`Hashed<T>`] into the cache, returning the canonical `Arc`.
+impl<V: Clone + Send + Sync + 'static> ValueCache<CryptoHash, V> {
+    /// Inserts a value constructed from a [`Hashed<T>`] into the cache, keyed
+    /// by its hash, returning the canonical `Arc<V>`.
     ///
     /// The `value` is wrapped in a [`Cow`] so that it is only cloned if it
     /// needs to be inserted in the cache.
-    pub fn insert_hashed(&self, value: Cow<Hashed<T>>) -> Arc<Hashed<T>>
+    pub fn insert_hashed<T>(&self, value: Cow<Hashed<T>>) -> Arc<V>
     where
         T: Clone,
+        V: From<Hashed<T>>,
     {
         let hash = (*value).hash();
         // Fast path: already in bounded cache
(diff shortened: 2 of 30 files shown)


linera-cache/src/value_cache.rs
```diff
@@ -77,7 +77,6 @@ where
     }
 
     /// Inserts a pre-wrapped `Arc<V>` into the cache, returning the canonical `Arc`.
-    #[cfg(with_testing)]
     pub fn insert_arc(&self, key: &K, value: Arc<V>) -> Arc<V> {
         self.dedup_insert(key, value)
     }
```
ma2bd, line 82: nit: do we need two distinct functions then?

linera-base/src/data_types.rs
```diff
@@ -1383,6 +1383,12 @@ impl From<Blob> for BlobContent {
     }
 }
 
+impl From<Arc<Blob>> for BlobContent {
```
afck, line 1386: Putting `Blob`s in `Arc`s actually doesn't buy us much, because the blob _content_ already has all its data in an `Arc` anyway, so we'd just need to deduplicate those.

(But if we prefer this, we could change the _inner_ `Arc` to `Box`, of course. But currently I think there are also still places where we clone the blob _content_, so the `Arc` there is still useful?)

linera-core/src/client/chain_client/mod.rs
```diff
@@ -2096,9 +2099,12 @@ impl<Env: Environment> ChainClient<Env> {
             .client
             .finalize_block(&committee, certificate.clone())
             .await?;
+        let certificate = Arc::new(certificate);
         self.update_validators(Some(&committee), Some(certificate.clone()))
             .await?;
-        Ok(ClientOutcome::Committed(Some(certificate)))
+        Ok(ClientOutcome::Committed(Some(Arc::unwrap_or_clone(
```
afck, line 2105: …or should even client outcomes contain `Arc`s?

linera-core/src/client/chain_client/mod.rs
```diff
@@ -3217,12 +3223,11 @@ impl<Env: Environment> ChainClient<Env> {
                         .await?
                         .into_iter()
                         .flatten()
-                        .map(Arc::unwrap_or_clone)
                         .collect();
                     remote_node.upload_blobs(missing_blobs).await?;
                     remote_node
                         .handle_confirmed_certificate(
-                            certificate,
+                            Arc::unwrap_or_clone(certificate),
```
afck: Should we pass in the `Arc` here, too? I imagine ultimately the remote node will just have to serialize it—no need to clone it for that?

linera-core/src/client/mod.rs
```diff
@@ -757,7 +757,7 @@ impl<Env: Environment> Client<Env> {
                 Ok((certificates, unresolved, validator_key)) => {
                     for certificate in certificates {
                         let mode = ReceiveCertificateMode::AlreadyChecked;
-                        self.receive_sender_certificate(certificate, mode, None)
+                        self.receive_sender_certificate(Arc::new(certificate), mode, None)
```
afck: Should they be put in the cache here?

linera-service/src/node_service.rs
afck: …or could GraphQL handle an `Option<Arc<ConfirmedBlock>>`, too?

afck (approved):
Looks great!

I think this could be taken even further (possibly in a follow-up PR).

Also, I'm not sure what the best approach to blobs is: They already contain an `Arc` internally, and also contain their own ID indirectly. 🤔

linera-base/src/data_types.rs
ndr-ds: I can try this as a follow up PR

linera-core/src/client/chain_client/mod.rs
ndr-ds: I can try this as a follow up PR

afck:
> Nothing to do / These changes follow the usual release cycle.

I think we should backport this.

ndr-ds:
Yes, 100%, sorry, I forgot to update the description

PR #6148, opened 2026-04-28
Attempt at using the protoc from github.

MathieuDutSik:
## Motivation

We are still facing `protoc` rate limits. 

## Proposal

Switch to using `github` itself as source.

## Test Plan

CI

## Release Plan

## Links

None


deuszx (commented):
Let's extract this to a custom github action (in our repo) where you hide the details of the `run` part and we only pass the version as an argument?

deuszx (approved):

PR #6153, opened 2026-04-28
Sync keyvaluestore

### LINKED ISSUE #3026: Have a version of `linera_views` that is synchronous

MathieuDutSik:
## Motivation

The `KeyValueStore` is an async library that is being used for implementing the views. When used for
protocol code that is fine. However, contract code is actually sync, so using `KeyValueStore` is actually
using the wrong tools.

Fixes https://github.com/linera-io/linera-protocol/issues/3026

## Proposal

Introduce a `SyncKeyValueStore` that does the operation in a sync way. The translation is direct.
After this the justification for having `async fn execute_operation(_)` collapses, and we can instead write
`fn execute_operation(_)`. That asyncness was anyway entirely artificial since we have in fact a `.block_wait()`
in the `linera-sdk` code.

A similar change cannot be made for the `Service` trait since it uses the `async-graphql` functionality, which does
not have a sync variant.

## Test Plan

CI

The test `test_wasm_end_to_end_matching_engine_benchmark` runs in 762.849ms in main and in 683.950916ms for this branch.

## Release Plan

That can actually be backported to `testnet_conway` since all the artificial `.blocking_wait()` calls
are inside of the `linera-sdk`, not of the protocol.

But of course, the documentation would have to be changed.

## Links

None


linera-views/src/backends/memory.rs
```diff
@@ -118,6 +121,17 @@ pub struct MemoryStore {
     max_stream_queries: usize,
 }
 
+/// A synchronous in-memory store.
+#[derive(Clone)]
+pub struct SyncMemoryStore {
```
ma2bd, line 114: I don't see any difference with `MemoryStore`. Seems like you could just implement all traits for `MemoryStore`.

linera-views/src/backends/memory.rs
```diff
@@ -260,6 +278,193 @@ impl MemoryStore {
...
+        }
+    }
+}
+
+impl SyncReadableKeyValueStore for SyncMemoryStore {
+    const MAX_KEY_SIZE: usize = usize::MAX;
+
+    fn max_stream_queries(&self) -> usize {
+        self.max_stream_queries
+    }
```
ma2bd: I don't think we need this in a `SyncReadableKeyValueStore` implementation.

linera-views/src/backends/memory.rs
```diff
@@ -260,6 +278,193 @@ impl MemoryStore {
...
+
+    fn root_key(&self) -> Result<Vec<u8>, MemoryStoreError> {
+        Ok(self.root_key.clone())
+    }
+
+    fn read_value_bytes(&self, key: &[u8]) -> Result<Option<Vec<u8>>, MemoryStoreError> {
+        let map = self
+            .map
+            .read()
+            .expect("SyncMemoryStore lock should not be poisoned");
```
ma2bd: nit: `.unwrap()` is accepted in this case

linera-views/src/backends/memory.rs
ma2bd, line 195: Could all the async methods simply call their synchronous sister instead of duplicating the code?

linera-views/src/backends/memory.rs
```diff
@@ -260,6 +278,193 @@ impl MemoryStore {
...
+    pub fn new_for_testing() -> Self {
+        Self {
+            map: Arc::default(),
+            root_key: Vec::new(),
+            max_stream_queries: SYNC_MEMORY_MAX_STREAM_QUERIES,
+        }
+    }
+}
+
+impl SyncReadableKeyValueStore for SyncMemoryStore {
```
ma2bd, line 225: Note that RocksDb would be a happy `SyncReadableKeyValueStore` (although I don't quite have a usecase for it right now).

linera-views/src/backends/memory.rs
MathieuDutSik: You are right, the code is simplified.

linera-views/src/backends/memory.rs
MathieuDutSik: Ok, done.

linera-views/src/backends/memory.rs
MathieuDutSik: Eliminated.

linera-views/src/backends/memory.rs
MathieuDutSik: Yes, the `MemoryStore` was anyway sync, so nest a `SyncMemoryStore` in it.

linera-views/src/backends/memory.rs
MathieuDutSik: Ok, done.

linera-views-derive/src/lib.rs
```diff
@@ -167,59 +198,89 @@ fn generate_view_code(input: &ItemStruct, root: bool) -> Result<TokenStream2, Er
...
+    let load_fn = match mode {
+        ViewMode::Async => quote! {
+            async fn load(context: #context) -> Result<Self, linera_views::ViewError> {
+                use linera_views::{context::Context as _, store::ReadableKeyValueStore as _};
+                #load_metrics
+                if Self::NUM_INIT_KEYS == 0 {
+                    Self::post_load(context, &[])
+                } else {
+                    let keys = Self::pre_load(&context)?;
+                    let values = context.store().read_multi_values_bytes(&keys).await?;
```
afck: Is this `.await` the only difference?

Would it make sense have a single variable `maybe_await` that is either `""` or `".await"` instead? Then we could deduplicate this and this line would become:


```suggestion
                    let values = context.store().read_multi_values_bytes(&keys)#maybe_await?;
```

linera-views-derive/src/lib.rs
```diff
@@ -167,59 +198,89 @@ fn generate_view_code(input: &ItemStruct, root: bool) -> Result<TokenStream2, Er
...
+                    let values = context.store().read_multi_values_bytes(&keys)?;
+                    Self::post_load(context, &values)
+                }
+            }
+        },
+    };
+
+    let has_pending_changes_fn = match mode {
+        ViewMode::Async => quote! {
+            async fn has_pending_changes(&self) -> bool {
```
afck: …and a `maybe_async` would allow deduplicating this and others.

linera-views/src/backends/memory.rs
```diff
@@ -36,6 +36,9 @@ pub struct MemoryStoreConfig {
 #[cfg(with_testing)]
 const TEST_MEMORY_MAX_STREAM_QUERIES: usize = 10;
 
+/// The number of streams for the sync test store
+const SYNC_MEMORY_MAX_STREAM_QUERIES: usize = 10;
```
afck: What kind of streams? Isn't this only used in the async case?

linera-views/src/backends/memory.rs
```diff
@@ -107,9 +110,13 @@ impl MemoryDatabases {
 static MEMORY_DATABASES: LazyLock<Mutex<MemoryDatabases>> =
     LazyLock::new(|| Mutex::new(MemoryDatabases::default()));
 
-/// A virtual DB client where data are persisted in memory.
+/// A synchronous in-memory store.
+///
+/// This is the primary type holding the actual fields and sync trait
+/// implementations, since the underlying storage (BTreeMap behind
```
afck:
```suggestion
/// implementations, since the underlying storage (`BTreeMap` behind
```

(Or "B-tree map", I guess.)

linera-views/src/context.rs
```diff
@@ -137,6 +140,52 @@ where
...
 }
 
+/// The context in which a sync view is operated.
+pub trait SyncContext: Clone
+where
+    crate::ViewError: From<Self::Error>,
+{
+    /// The type of the key-value store used by this context.
+    type Store: SyncReadableKeyValueStore
+        + SyncWritableKeyValueStore
```
afck, line 150: That's the only difference compared to `Context`… I also noticed that some (all?) of the view types are actually identical to their async versions.

I'm wondering if it would be better to, instead of having sync versions of all these structs and traits, just add sync (i.e. blocking) methods to the existing structs and traits?

In some impls (memory, RocksDB) the async methods would call the more natural sync ones; in others (ScyllaDB, DynamoDB) it would be the other way round? For views, some of the methods would still have two full implementations, of course, but e.g. all the mutation methods would not, since they are already sync!

linera-views-derive/src/lib.rs
MathieuDutSik: In that case, it is not the only difference. There is also
```rust
use linera_views::{context::Context as _, store::ReadableKeyValueStore as _};
```
vs
```rust
use linera_views::{context::SyncContext as _, store::SyncReadableKeyValueStore as _};
```

linera-views-derive/src/lib.rs
MathieuDutSik: Yes, we can do that.

linera-views/src/backends/memory.rs
MathieuDutSik: Yes, we had a leakage of abstraction:
* Yes, the `SYNC_MEMORY_MAX_STREAM_QUERIES` is not relevant to the `SyncMemoryStore`.
* But the `MemoryStore` was nesting a `SyncMemoryStore` hence that structure.

Addressed.

linera-views/src/context.rs
MathieuDutSik: Using blocking method like `.blocking_wait()` is exactly what we have now in the `linera-sdk`.
I think the benchmarks demonstrate that this has costs.

We certainly could provide Sync and async functionalities for all the storage. However, I do not think it is a good idea since it oversteps the crucial question: What is actually being used?
And the situation is the following:
* For protocol code, async code is most natural.
* For smart contract code, sync code is most natural.

I think that is a better approach than looking at having the most general code.

linera-views-derive/src/snapshots/linera_views_derive__tests__test_generate_view_code_C.snap
```diff
@@ -104,7 +104,8 @@ where
         Ok(Self { register, collection })
     }
     async fn load(context: C) -> Result<Self, linera_views::ViewError> {
-        use linera_views::{context::Context as _, store::ReadableKeyValueStore as _};
+        use linera_views::context::Context as _;
+        use linera_views::store::ReadableKeyValueStore as _;
```
afck, line 108: (That change seems unnecessary.)

linera-views/src/sync_views/lazy_register_view.rs
```diff
@@ -0,0 +1,181 @@
...
+/// A synchronous view that supports modifying a single value of type `T`,
+/// loaded lazily on first access.
+///
+/// This is a thin wrapper around [`LazyRegisterView`] that implements [`SyncView`]
+/// instead of [`View`](crate::views::View).
+#[derive(Debug, Allocative)]
+#[allocative(bound = "C, T: Allocative")]
+pub struct SyncLazyRegisterView<C, T>(
+    /// The inner async lazy register view whose state and logic we reuse.
+    pub(crate) LazyRegisterView<C, T>,
```
afck: Should this be private? Do we access it elsewhere?

linera-views/src/sync_views/log_view.rs
```diff
@@ -0,0 +1,332 @@
...
+
+/// A synchronous view that supports logging values of type `T`.
+///
+/// This is a thin wrapper around [`LogView`] that implements [`SyncView`]
+/// instead of [`View`](crate::views::View).
+#[derive(Debug, Allocative)]
+#[allocative(bound = "C, T: Allocative")]
+pub struct SyncLogView<C, T>(
+    /// The inner async log view whose state and logic we reuse.
+    pub(crate) LogView<C, T>,
```
afck: (Private?)

linera-views-derive/src/snapshots/linera_views_derive__tests__test_generate_view_code_C.snap
MathieuDutSik: I think it is fine.
The code generated via the derive function is suboptimal in the way it is presented. And I think that is fine for computer-generated code. Better that than being incorrect.

If we have computer generated code, then I think what matters is the generating code and its clarity.

linera-views/src/sync_views/log_view.rs
MathieuDutSik: Yes.

linera-views/src/sync_views/lazy_register_view.rs
MathieuDutSik: Yes.

linera-views-derive/src/snapshots/linera_views_derive__tests__test_generate_view_code_C.snap
afck: Sure, I'm not against it. I just meant it's unrelated to the rest of this PR? Not important; not a blocker!

MathieuDutSik: It is related to the rest of the PR.
With the abstraction about the trait, what we have is the following construction in `linera-views-derive/src/lib.rs`:
```rust
            use #context_trait as _;
	    use #read_kv_trait as _;
```

I think that this nicer overall.

But if we want to avoid that, then yes actually we can do with
```rust
            use {#context_trait as _, #read_kv_trait as _};
```

So, in the end yes we can keep that with minimal effort.

ma2bd:
continued in https://github.com/linera-io/linera-protocol/pull/6608

PR #6156, opened 2026-04-28
[backport] Add streaming DownloadBlobs RPC endpoint (#5976)

ndr-ds:
## Motivation

Backport of #5976 to `testnet_conway`.

Adds the streaming `DownloadBlobs` gRPC and simple-transport endpoints
that yield blobs as a stream (one `BlobContent` per requested ID, in
order). Callers are not rewired in this PR; they continue to use the
parallel unary `download_blob` path. The endpoint must land and be
deployed on the validators first, so the follow-up client-side rewire
can be tested against validators that actually serve it.

## Proposal

Adds the new RPC method to:

- `linera-rpc/proto/rpc.proto` (gRPC method definition)
- `linera-rpc/src/grpc/client.rs` and `linera-service/src/proxy/grpc.rs`
  (gRPC client + server)
- `linera-rpc/src/simple/client.rs`, `simple/server.rs`,
  `simple/transport.rs`, `message.rs`, `client.rs`
  (simple transport: dedicated connection per streaming request)
- `linera-core/src/node.rs` (`ValidatorNode` trait method)
- `linera-core/src/remote_node.rs` (`RemoteNode::download_blobs`)
- `linera-storage/src/db_storage.rs` (server-side blob fetching helper)
- `linera-rpc/tests/snapshots/format__format.yaml.snap` (wire-format
  snapshot adds the new variant)

Per the variant-index constraint of the simple transport, the new
`DownloadBlobs` variant of `RpcMessage` is placed at the end of the
enum on `testnet_conway` instead of in its main-branch position
(between `DownloadBlob` and `DownloadPendingBlob`). This preserves the
wire-format indices of all existing variants. A comment on the variant
documents this constraint for future backporters.

## Test Plan

CI.

## Release Plan

### DIFF
--- a/linera-rpc/src/message.rs
+++ b/linera-rpc/src/message.rs
@@ -70,6 +70,13 @@ pub enum RpcMessage {
 
     PreviousEventBlocks(Box<(ChainId, Vec<StreamId>)>),
     PreviousEventBlocksResponse(Box<BTreeMap<StreamId, (BlockHeight, CryptoHash)>>),
+
+    // New variant added by the streaming DownloadBlobs RPC backport.
+    // IMPORTANT: this variant must remain at the end of the enum to preserve
+    // wire-format compatibility with testnet_conway validators (the
+    // serde_reflection format used by the simple transport encodes the variant
+    // index, so existing variants must keep their existing indices).
+    DownloadBlobs(Vec<BlobId>),
 }
 
 impl RpcMessage {
@@ -101,6 +108,7 @@ impl RpcMessage {
             | UploadBlob(_)
             | UploadBlobResponse(_)
             | DownloadBlob(_)
+            | DownloadBlobs(_)
             | DownloadBlobResponse(_)
             | DownloadPendingBlobResponse(_)
             | DownloadConfirmedBlock(_)
@@ -132,6 +140,7 @@ impl RpcMessage {
             | NetworkDescriptionQuery
             | UploadBlob(_)
             | DownloadBlob(_)
+            | DownloadBlobs(_)
             | DownloadConfirmedBlock(_)
             | BlobLastUsedBy(_)
             | BlobLastUsedByCertificate(_)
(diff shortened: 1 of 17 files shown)


linera-rpc/src/simple/transport.rs
```diff
@@ -79,6 +86,16 @@ pub trait ConnectionPool: Send {
 #[async_trait]
 pub trait MessageHandler: Clone {
     async fn handle_message(&mut self, message: RpcMessage) -> Option<RpcMessage>;
+
+    /// Handle a batch blob download request by streaming one
```
afck:
```suggestion
    /// Handles a batch blob download request by streaming one
```

linera-rpc/src/simple/transport.rs
```diff
@@ -79,6 +86,16 @@ pub trait ConnectionPool: Send {
...
+    /// `RpcMessage::DownloadBlobResponse` per requested blob ID.
+    /// Returns `None` if not supported.
+    async fn handle_download_blobs(
+        &mut self,
+        _blob_ids: Vec<BlobId>,
+    ) -> Option<Pin<Box<dyn Stream<Item = RpcMessage> + Send>>> {
+        None
```
afck: (I guess we could just return a `Vec<Blob>` and turn that into a stream?)

linera-rpc/src/message.rs
```diff
@@ -70,6 +70,13 @@ pub enum RpcMessage {
 
     PreviousEventBlocks(Box<(ChainId, Vec<StreamId>)>),
     PreviousEventBlocksResponse(Box<BTreeMap<StreamId, (BlockHeight, CryptoHash)>>),
+
+    // New variant added by the streaming DownloadBlobs RPC backport.
+    // IMPORTANT: this variant must remain at the end of the enum to preserve
+    // wire-format compatibility with testnet_conway validators (the
+    // serde_reflection format used by the simple transport encodes the variant
+    // index, so existing variants must keep their existing indices).
```
afck: (Not sure if this doc is necessary; it kind of documents the commit history more than the enum variant itself.)

afck (approved):

linera-rpc/src/message.rs
ma2bd: I noticed that Claude loves to leave these breadcrumbs in comments.

PR #6161, opened 2026-04-29
Make process_inbox optimistic with prepare_chain fallback

ndr-ds:
## Motivation

`process_inbox` unconditionally calls `prepare_chain` first, which fans out `synchronize_chain_state` to a quorum of validators. Each call forces every validator to hydrate the chain `ChainStateView` from Scylla, which lands as reads on the chain single partition (one shard per replica). Under burst load (many chains processing inbox concurrently, e.g. during a worker restart catch-up or routine activity bursts) this saturates per-shard reader permits and produces validator p99 latency spikes.

In the common case where the worker has been running and processing notifications, the local view is already current and the defensive sync is wasted work.

## Proposal

Try the inbox-processing loop optimistically without `prepare_chain`. If it fails with an error consistent with our local view being stale, fall back to the full `prepare_chain` + retry, which preserves the previous semantics on the slow path.

The staleness predicate `Error::is_likely_stale_local_state` is intentionally conservative: false negatives bubble the error up (same as today), false positives only cost one wasted validator round-trip on an error path. Initial candidates: `WalletSynchronizationError`, `BlockProposalError`, and the `WorkerError` variants whose `is_local()` returns false for staleness reasons (`UnexpectedBlockHeight`, `InvalidEpoch`, `InvalidBlockChaining`, `EventsNotFound`). Open to expanding the list during review.

## Test Plan

CI for now. Considered adding a unit test simulating stale local state but deferred until the predicate set is finalised. Happy to add tests in a follow-up commit if reviewer prefers.

## Release Plan

Backport to `testnet_conway`.

### DIFF
--- a/linera-core/src/client/chain_client/mod.rs
+++ b/linera-core/src/client/chain_client/mod.rs
@@ -346,6 +346,35 @@ impl Error {
     pub fn signer_failure(err: impl signer::Error + 'static) -> Self {
         Self::Signer(Box::new(err))
     }
+
+    /// Returns `true` if this error is consistent with the local view of the
+    /// chain being out-of-date relative to the network.
+    ///
+    /// Used by [`ChainClient::process_inbox`] to decide whether to retry after
+    /// a quorum-wide `synchronize_chain_state` call. The list is intentionally
+    /// conservative: a false negative (real staleness mistakenly treated as
+    /// fatal) just bubbles the error up — the same behaviour as before the
+    /// optimistic path was introduced — while a false positive only costs one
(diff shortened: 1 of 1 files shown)


afck:
~Makes sense, but it would be better to apply this to all block proposals, not just the `process_inbox` ones.~

Ah, on the other hand, it's not clear if an operation _should_ still be applied, once the chain state has changed.

afck:
> Nothing to do

Let's backport it to `testnet_conway`?

ndr-ds:
The test failure is very suspicious, so maybe I'll just close this for now

PR #6163, opened 2026-04-29
Merge `confirmed_log` and `preprocessed_blocks` into `block_hashes`

### LINKED ISSUE #5069: Merge `confirmed_log` and `preprocessed_blocks`.
With checkpoints, the `confirmed_log` won't necessarily start at height 0 anymore, so in a sense the whole chain can be "sparse" even if it _is_ being executed.

We should merge the two collections that track a chain's block hashes.

afck:
## Motivation

We are planning to implement checkpointing for chains, which will mean that even the _executed_ part of a chain will not necessarily start at block 0. This makes the artificial distinction between `confirmed_log` and `preprocessed_blocks` even more complicated, since the former now has an offset, i.e. index is not equal to block height anymore.

## Proposal

Replace `confirmed_log: LogView<CryptoHash>` and `preprocessed_blocks: MapView<BlockHeight, CryptoHash>` with a single `block_hashes: CustomMapView<BlockHeight, CryptoHash>` that holds executed blocks (`height < next_block_height`) and preprocessed blocks (`height >= next_block_height`) together.

Add a big-endian `CustomSerialize` impl for `BlockHeight` so that lexicographic key order matches numeric order; this also fixes a latent bug in `next_height_to_preprocess`, which relied on `indices().last()` returning the numerically highest height. (Fixed by #6162 on `testnet_conway`.)

Collapse three near-duplicate `WorkerError` variants (`ConfirmedLogEntryNotFound`, `PreprocessedBlocksEntryNotFound`, `ConfirmedBlockHashNotFound`) into a single `BlockHashNotFound`.

## Test Plan

## Release Plan

## Links

- Closes https://github.com/linera-io/linera-protocol/issues/5069.


deuszx:
> Add a big-endian `CustomSerialize` impl for `BlockHeight` so that lexicographic key order matches numeric order; this also fixes a latent bug in `next_height_to_preprocess`, which relied on `indices().last()` returning the numerically highest height. (Fixed by #6162 on `testnet_conway`.)

Can you add a test for this, similar to the on in #6162  ?

afck:
@deuszx: Done in https://github.com/linera-io/linera-protocol/pull/6163/changes/7fcb7255f0d0c07fa60e64cdf3c1d2203e65abe1.

bart-linera (approved):

linera-chain/src/chain.rs
```diff
@@ -482,10 +481,13 @@ where
     ///
     /// The "+ 1" is so that it can be used in the same places as `next_block_height`.
     pub async fn next_height_to_preprocess(&self) -> Result<BlockHeight, ChainError> {
-        if let Some(height) = self.preprocessed_blocks.indices().await?.last() {
-            return Ok(height.saturating_add(BlockHeight(1)));
+        let next_block_height = self.tip_state.get().next_block_height;
+        if let Some(height) = self.block_hashes.indices().await?.last() {
+            if *height >= next_block_height {
+                return Ok(height.saturating_add(BlockHeight(1)));
+            }
```
deuszx: For chains with many executed blocks and no preprocessed ones, this is an extra cost on every call. Probably negligible since `indices().last()` on a sorted map should be `O(log n)`, but worth confirming the storage backend doesn't first load the full list.

ma2bd: `.indices()` most certainly loads the full list

afck: Good catch! We'd need a `last_key` method, I guess.

afck: I added a field to memoize the max height.

Later we could do something like https://github.com/linera-io/linera-protocol/pull/6171, and remove that field.

linera-chain/src/chain.rs
```diff
@@ -212,9 +212,10 @@ where
     /// The incomplete sets of blobs for upcoming proposals.
     pub pending_proposed_blobs: ReentrantCollectionView<C, AccountOwner, PendingBlobsView<C>>,
 
-    /// Hashes of all certified blocks for this sender.
-    /// This ends with `block_hash` and has length `usize::from(next_block_height)`.
-    pub confirmed_log: LogView<C, CryptoHash>,
+    /// Hashes of all known blocks in this chain, indexed by their height. A block at
+    /// `height < next_block_height` is executed; a block at `height >= next_block_height`
+    /// is preprocessed (verified but not yet executed) and may not be contiguous.
```
ma2bd, line 217: (maybe related) Is this still the case that we eagerly execute sender blocks if they happen to be contiguous?

afck: Yes

PR #6166, opened 2026-04-29
Only put _fast_ pending blocks in the wallet.

afck:
## Motivation

Storing the pending block in the wallet is mainly important for _fast_ blocks, where liveness depends on the client not making any conflicting proposal.

Blocks proposals can also quite large, and take up a lot of space in the wallet.

## Proposal

Only persist _fast_ proposals.

## Test Plan

Tests were added to `linera-client`.

## Release Plan

- Release SDK.

## Links

### DIFF
--- a/linera-client/src/client_context.rs
+++ b/linera-client/src/client_context.rs
@@ -430,12 +430,15 @@ impl<Env: Environment> ClientContext<Env> {
             .map_err(error::Inner::wallet)?
             .and_then(|chain| chain.owner);
 
-        let pending_proposal = client
+        // Only persist proposals that were made in the fast round: they need to be
+        // remembered across sessions even if the chain has since progressed past the
+        // fast round, so the proposer can retry them.
+        let pending_fast_proposal = client
             .pending_proposal()
             .await
             .filter(|p| p.round.is_some_and(|r| r.is_fast()));
         let new_chain = wallet::Chain {
-            pending_proposal,
+            pending_fast_proposal,
             owner: existing_owner,
             ..info.as_ref().into()
         };
@@ -1019,7 +1022,7 @@ impl<Env: Environment> ClientContext<Env> {
             for chain_client in chain_clients {
                 let info = chain_client.chain_info().await?;
                 let client_owner = chain_client.preferred_owner();
-                let pending_proposal = chain_client
+                let pending_fast_proposal = chain_client
                     .pending_proposal()
                     .await
                     .filter(|p| p.round.is_some_and(|r| r.is_fast()));
@@ -1027,7 +1030,7 @@ impl<Env: Environment> ClientContext<Env> {
                     .insert(
                         info.chain_id,
                         wallet::Chain {
-                            pending_proposal,
+                            pending_fast_proposal,
                             owner: client_owner,
                             ..info.as_ref().into()
                         },
--- a/linera-client/src/client_context.rs
+++ b/linera-client/src/client_context.rs
@@ -431,8 +431,7 @@ impl<Env: Environment> ClientContext<Env> {
             .and_then(|chain| chain.owner);
 
         // Only persist proposals that were made in the fast round: they need to be
-        // remembered across sessions even if the chain has since progressed past the
-        // fast round, so the proposer can retry them.
+        // remembered across sessions to make sure there are no conflicting fast proposals.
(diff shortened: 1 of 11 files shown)


linera-client/src/client_context.rs
```diff
@@ -430,8 +430,12 @@ impl<Env: Environment> ClientContext<Env> {
             .map_err(error::Inner::wallet)?
             .and_then(|chain| chain.owner);
 
+        let pending_proposal = client
```
deuszx: Can you add a short comment here explaining why we filter out non-fast rounds.

linera-client/src/client_context.rs
```diff
@@ -430,8 +430,12 @@ impl<Env: Environment> ClientContext<Env> {
...
+            .pending_proposal()
+            .await
+            .filter(|p| p.round.is_some_and(|r| r.is_fast()));
         let new_chain = wallet::Chain {
-            pending_proposal: client.pending_proposal().await,
+            pending_proposal,
```
deuszx: Is the pending proposals in the Chain ONLY for fast rounds, and never for regular ones? If so, maybe rename the field

afck: Done in https://github.com/linera-io/linera-protocol/pull/6166/commits/cf6534eb7df8d6ca240e9f6d5f0ccd57068014cb.

linera-client/src/client_context.rs
afck: Done in https://github.com/linera-io/linera-protocol/pull/6166/commits/30226428f51620bd95572b3c4e02289c108f5526.

linera-client/src/unit_tests/client_context.rs
```diff
@@ -0,0 +1,159 @@
...
+    test_utils::{FaultType, MemoryStorageBuilder, TestBuilder},
+    worker::{DEFAULT_BLOCK_CACHE_SIZE, DEFAULT_EXECUTION_STATE_CACHE_SIZE},
+};
+use linera_rpc::node_provider::DEFAULT_MAX_BACKOFF;
+
+use crate::{client_context::ClientContext, config::GenesisConfig};
+
+/// Builds a production [`ClientContext`] with a fresh in-memory wallet, sharing validators
+/// with the given [`TestBuilder`].
+async fn make_context(
```
deuszx, line 31: Personally, I try to put the helpers _after_ the test. Rule being the test is more interesting to the reader

deuszx (approved):

linera-client/src/unit_tests/client_context.rs
afck: But aren't we doing the opposite in most test files currently?

deuszx: I try not to, but sometimes it slips in.

afck: I had Claude look at other files:

```
Surveyed the project's test files. The convention is consistently helpers/setup before test cases:                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                                                                           
  - linera-client/src/unit_tests/chain_listener.rs — local ClientContext impl (line 33) before all tests (starting line 91)                                                                                                                                                                                                                                                
  - linera-core/src/unit_tests/worker_tests.rs — TestEnvironment struct/impl + helper functions (lines 101–682) all before the first test (line 686)                                                                                                                                                                                                                       
  - linera-core/src/unit_tests/wasm_client_tests.rs — ChainClientExt impl (line 71) before tests (line 86)                                                                                                                                                                                                                                                                 
  - linera-views/tests/random_container_tests.rs, hashable_tests.rs — interleaved (helpers near tests that use them), but each helper precedes its first user
```

So I'd keep it like this for now?

deuszx: So out of all files with tests it found only 4 that _define-before-use_ ? I'd expect this to be a vast minority.

Anyway, go ahead – not a blocker

afck:
```
The project is actually split fairly evenly: roughly 20 files with helpers-first vs. 15+ with tests-first.
```

I don't think it's that clear-cut. That was also my own impression.

I'm happy to follow any convention and fix all the tests once the team agrees on it.

PR #6171, opened 2026-04-29
`find_first_key_by_prefix`, `find_last_key_by_prefix`

afck:
## Motivation

As discussed in #6163, it's useful to be able to read the last key of a map view.

## Proposal

Add `find_first_key_by_prefix`, `find_last_key_by_prefix`, and the value variants to `ReadableKeyValueStore`, with default impls falling back to `find_keys_by_prefix(...).first()/.last()`. Native overrides:

- memory: `BTreeMap` range, O(log N)
- RocksDB
- DynamoDB: `Query` with `Limit=1` and `ScanIndexForward=true/false`
- IndexedDB: `IDBCursor` with `Next`/`Prev` direction
- Wrappers (journaling, value_splitting, lru_caching, metering, dual): forward to the inner store; metering adds latency histograms; value_splitting strips the 4-byte segment index from the result.

Adds view-level helpers:
- `ByteMapView::first_key`, `last_key`, `first_key_value`, `last_key_value`
- `MapView` / `CustomMapView::first_index`, `last_index`,
  `first_index_value`, `last_index_value`
- `ByteSetView::first_key`, `last_key`
- `SetView` / `CustomSetView::first_index`, `last_index`

The view methods take a fast path straight to the store when there are no staged changes; otherwise they fall back to the existing `for_each_key_while` / `for_each_index` machinery, which already handles the merge of in-memory updates with persisted state.

## Test Plan

Tests were added.

## Release Plan

## Links

### DIFF
--- a/linera-views/src/backends/lru_caching.rs
+++ b/linera-views/src/backends/lru_caching.rs
@@ -348,6 +348,16 @@ where
         &self,
         key_prefix: &[u8],
     ) -> Result<Option<Vec<u8>>, Self::Error> {
+        if let Some(cache) = self.get_exclusive_cache() {
+            let mut cache = cache.lock().unwrap();
+            if let Some(keys) = cache.query_find_keys(key_prefix) {
+                #[cfg(with_metrics)]
+                metrics::FIND_KEYS_BY_PREFIX_CACHE_HIT_COUNT
+                    .with_label_values(&[])
+                    .inc();
+                return Ok(keys.into_iter().next());
+            }
+        }
         self.store.find_first_key_by_prefix(key_prefix).await
     }
 
@@ -355,6 +365,16 @@ where
         &self,
         key_prefix: &[u8],
     ) -> Result<Option<Vec<u8>>, Self::Error> {
+        if let Some(cache) = self.get_exclusive_cache() {
+            let mut cache = cache.lock().unwrap();
+            if let Some(keys) = cache.query_find_keys(key_prefix) {
+                #[cfg(with_metrics)]
+                metrics::FIND_KEYS_BY_PREFIX_CACHE_HIT_COUNT
+                    .with_label_values(&[])
+                    .inc();
+                return Ok(keys.into_iter().next_back());
+            }
+        }
         self.store.find_last_key_by_prefix(key_prefix).await
     }
 
@@ -362,6 +382,16 @@ where
         &self,
         key_prefix: &[u8],
     ) -> Result<Option<(Vec<u8>, Vec<u8>)>, Self::Error> {
+        if let Some(cache) = self.get_exclusive_cache() {
+            let mut cache = cache.lock().unwrap();
+            if let Some(key_values) = cache.query_find_key_values(key_prefix) {
+                #[cfg(with_metrics)]
+                metrics::FIND_KEY_VALUES_BY_PREFIX_CACHE_HIT_COUNT
+                    .with_label_values(&[])
+                    .inc();
+                return Ok(key_values.into_iter().next());
+            }
+        }
         self.store.find_first_key_value_by_prefix(key_prefix).await
     }
 
@@ -369,6 +399,16 @@ where
         &self,
         key_prefix: &[u8],
     ) -> Result<Option<(Vec<u8>, Vec<u8>)>, Self::Error> {
+        if let Some(cache) = self.get_exclusive_cache() {
+            let mut cache = cache.lock().unwrap();
+            if let Some(key_values) = cache.query_find_key_values(key_prefix) {
+                #[cfg(with_metrics)]
+                metrics::FIND_KEY_VALUES_BY_PREFIX_CACHE_HIT_COUNT
+                    .with_label_values(&[])
+                    .inc();
+                return Ok(key_values.into_iter().next_back());
+            }
+        }
         self.store.find_last_key_value_by_prefix(key_prefix).await
     }
 }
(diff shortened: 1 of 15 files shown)


linera-views/src/backends/lru_caching.rs
```diff
@@ -343,6 +343,74 @@ where
...
+    ) -> Result<Option<Vec<u8>>, Self::Error> {
+        if let Some(cache) = self.get_exclusive_cache() {
+            let mut cache = cache.lock().unwrap();
+            if let Some(keys) = cache.query_find_keys(key_prefix) {
+                #[cfg(with_metrics)]
+                metrics::FIND_KEYS_BY_PREFIX_CACHE_HIT_COUNT
+                    .with_label_values(&[])
+                    .inc();
+                return Ok(keys.into_iter().next());
+            }
```
MathieuDutSik, line 359: If you match on the `query_find_key_values` then you can also prematurely end the search.

linera-views/src/backends/lru_caching.rs
```diff
@@ -343,6 +343,74 @@ where
...
+    ) -> Result<Option<Vec<u8>>, Self::Error> {
+        if let Some(cache) = self.get_exclusive_cache() {
+            let mut cache = cache.lock().unwrap();
+            if let Some(keys) = cache.query_find_keys(key_prefix) {
+                #[cfg(with_metrics)]
+                metrics::FIND_KEYS_BY_PREFIX_CACHE_HIT_COUNT
+                    .with_label_values(&[])
+                    .inc();
+                return Ok(keys.into_iter().next_back());
+            }
```
MathieuDutSik, line 376: Same.

linera-views/src/backends/lru_caching.rs
```diff
@@ -343,6 +343,74 @@ where
...
+            if let Some(key_values) = cache.query_find_key_values(key_prefix) {
+                #[cfg(with_metrics)]
+                metrics::FIND_KEY_VALUES_BY_PREFIX_CACHE_HIT_COUNT
+                    .with_label_values(&[])
+                    .inc();
+                return Ok(key_values.into_iter().next());
+            }
+        }
+        self.store.find_first_key_value_by_prefix(key_prefix).await
+    }
```
MathieuDutSik, line 396: I guess we do not need a corresponding cache entry such as `cache.query_find_first_key` or similar just yet.

linera-views/tests/views_tests.rs
```diff
@@ -646,6 +671,191 @@ async fn test_byte_map_view() -> Result<()> {
...
+        view.set.insert(&256u128)?;
+        view.save().await?;
+    }
+    {
+        let view = CustomSetStateView::load(context.clone()).await?;
+        assert_eq!(view.set.first_index().await?, Some(1u128));
+        assert_eq!(view.set.last_index().await?, Some(256u128));
+    }
+    Ok(())
+}
```
MathieuDutSik, line 857: I would prefer actually to have all those tests in a single  function `fn unit_tests_first_last_map_view`.

I would also add tests to `run_map_view_mutability`. It is a fuzzing test and it could catch some regression.

More importantly, we should have tests that exercise the pain points of the implementation. And that can be for example the `ValueSplitting` problems show up with large values. But there is a `LimitedTestMemoryStore` that can help dealing with that.

MathieuDutSik (commented):
It is a nice feature.
But it should be noted, that while it is not exactly a subset, it is strongly related to the PR https://github.com/linera-io/linera-protocol/pull/4975

linera-views/src/backends/value_splitting.rs
```diff
@@ -252,6 +252,39 @@ where
...
+        // 4-byte segment index.
+        Ok(self
+            .store
+            .find_first_key_by_prefix(key_prefix)
+            .await?
+            .map(|mut key| {
+                key.truncate(key.len() - 4);
+                key
+            }))
+    }
```
MathieuDutSik, line 271: Unfortunately, this could be a leftover key, so I am not sure that even this construction works.

linera-views/src/backends/value_splitting.rs
```diff
@@ -252,6 +252,39 @@ where
...
+    ) -> Result<Option<Vec<u8>>, Self::Error> {
+        // The largest stored key under any user-key prefix is `K|n` for the largest
+        // user key `K` and its highest segment index `n`. Strip the trailing 4 bytes.
+        Ok(self
+            .store
+            .find_last_key_by_prefix(key_prefix)
+            .await?
+            .map(|mut key| {
+                key.truncate(key.len() - 4);
+                key
```
MathieuDutSik, line 285: That is also broken by the leftover keys.
Scenario is the following:
* Write a big value that uses several keys.
* Remove that key with a `DeleteKey` in a batch. Since removing the other segment keys requires doing a read, they are left over.
* Then those keys get a wrong result.

The solution is to:
* Read the index of the key in question. If zero, it is good, return.
* If non-zero, we know it is a segment key (leftover or not, no way to know with an async operation). So, revert to a `find_keys_by_prefix`.

linera-views/src/backends/value_splitting.rs
```diff
@@ -252,6 +252,39 @@ where
...
+            }))
+    }
```
MathieuDutSik, line 287: The implementation is actually using the default trait implementation, which uses a `find_key_values_by_prefix`.

That makes the implementation correct. However, in my opinion, it also makes it slow. All the containers are using this `ValueSplitting`. So, this is just about the same as using the `find_keys_by_prefix`.

Instead, I think that a better approach is:
* Look for the first (key, value). If reading it allows to prove it is not segment, then return it.
* Barring that, use `find_key_values_by_prefix`.

afck:
Closing in favor of https://github.com/linera-io/linera-protocol/pull/6183.

PR #6175, opened 2026-04-29
Clean up bridge ABI duplication and tighten typed fields

MathieuDutSik:
## Motivation

Looking at the code reveals some issue with the `linera-bridge`.

## Proposal

Following corrections are done:
* Eliminate the code duplication for `BridgeOperation` and `BridgeParameters`.
* Make the types explicit: `B256` for Ethereum, `ChainId`, `CryptoHash` and so for the Linera world.
* Use ˙BlockHeight` for the Linera block heights.
* Simplify the `.map_err(_)`.
* Remove the `let mut handle = handle` and similar.
* Bug fix: remove a `rustls::crypto::ring::default_provider().install_default()`.
* Eliminate `String` for address and the type `Address`.
* Use `Duration` for duration instead of seconds.

## Test Plan

CI

## Release Plan

## Links

None.

### DIFF
--- a/examples/Cargo.lock
+++ b/examples/Cargo.lock
@@ -3860,6 +3860,7 @@ dependencies = [
  "alloy-rlp",
  "alloy-trie",
  "anyhow",
+ "linera-base",
  "serde",
 ]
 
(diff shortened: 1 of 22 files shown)


deuszx (changes requested):
Can you push this branch to `linera-io` repository? Otherwise I can't trigger the `bridge-e2e` tests and I'm pretty sure it will fail.

deuszx (approved):

PR #6183, opened 2026-04-30
First last key

MathieuDutSik:
## Motivation

This is a continuation of the existing PR https://github.com/linera-io/linera-protocol/pull/6171
by implementing the corrections that I asked.

## Proposal

Add find_first_key_by_prefix, find_last_key_by_prefix, and the value variants to ReadableKeyValueStore, with default impls falling back to find_keys_by_prefix(...).first()/.last(). Native overrides:

* memory: BTreeMap range, O(log N)
* RocksDB
* DynamoDB: Query with Limit=1 and ScanIndexForward=true/false
* IndexedDB: IDBCursor with Next/Prev direction

For the wrappers:
* Journaling, Metering, Dual are straightforward.
* The separate `linera-storage-service` is also straightforward.
* For the LRU caching, we use the already cached `find_keys_by_prefix`, `find_key_values_by_prefix` to return result when possible.

For the `ValueSplitting` this is harder since we can split some keys and anyway the remaining key could be leftover from previous big values that were deleted:
* We cannot distinguish leftover keys from legitimate existing segments.
* So, if the `find_first_key_by_prefix` returns a key that has index 0, we know it is ok, and we return. Otherwise, we do a `find_keys_by_prefix` which is expensive but should be very rare.
* For the `find_first_key_value_by_prefix`, if the obtained key has index 0 and the obtained value a count of 1, we know it is a legitimate one and we can return. Otherwise, we do an expensive but rare `find_key_values_by_prefix`.

## Test Plan

CI

The test `map_view_mutability` was modified to use a Value splitting store. The way the keys are used has changed, but this is only because the generated key-set should be prefix-free.

## Release Plan

## Links

None

### DIFF
--- a/linera-storage-service/proto/key_value_store.proto
+++ b/linera-storage-service/proto/key_value_store.proto
@@ -90,6 +90,46 @@ message ReplyFindKeyValuesByPrefix {
 }
 
 
+message RequestFindFirstKeyByPrefix {
+  bytes key_prefix = 1;
+}
+
+message ReplyFindFirstKeyByPrefix {
+  optional bytes key = 1;
+}
+
+
+message RequestFindLastKeyByPrefix {
+  bytes key_prefix = 1;
+}
+
+message ReplyFindLastKeyByPrefix {
+  optional bytes key = 1;
+}
+
+
+message RequestFindFirstKeyValueByPrefix {
+  bytes key_prefix = 1;
+}
+
+message ReplyFindFirstKeyValueByPrefix {
+  optional KeyValue key_value = 1;
+  int64 message_index = 2;
+  int32 num_chunks = 3;
+}
+
+
+message RequestFindLastKeyValueByPrefix {
+  bytes key_prefix = 1;
+}
+
+message ReplyFindLastKeyValueByPrefix {
+  optional KeyValue key_value = 1;
+  int64 message_index = 2;
+  int32 num_chunks = 3;
+}
+
+
 message RequestWriteBatchExtended {
   repeated Statement statements = 1;
 }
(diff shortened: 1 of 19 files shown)


linera-storage-service/proto/key_value_store.proto
afck, line 199: (Why the `Process` prefix everywhere?)

linera-storage-service/src/client.rs
```diff
@@ -279,6 +281,126 @@ impl ReadableKeyValueStore for StorageServiceStoreInternal {
...
+        let mut client = StorageServiceClient::new(channel);
+        let _guard = self.acquire().await;
+        let response = client
+            .process_find_last_key_by_prefix(request)
+            .make_sync()
+            .await?;
+        let response = response.into_inner();
+        let ReplyFindLastKeyByPrefix { key } = response;
+        Ok(key)
+    }
```
afck, line 335: These methods share a lot of code; can that be deduplicated? Maybe they'd just pass in one or two closures to the same general method: what method to call on the client, and possibly what to do with the result?

linera-views/src/backends/dual.rs
```diff
@@ -191,6 +191,70 @@ where
...
+        match self {
+            Self::First(store) => store
+                .find_last_key_value_by_prefix(key_prefix)
+                .await
+                .map_err(DualStoreError::First),
+            Self::Second(store) => store
+                .find_last_key_value_by_prefix(key_prefix)
+                .await
+                .map_err(DualStoreError::Second),
+        }
```
afck, line 256: It seems this could also be deduplicated with a single generic method that does the `match`ing and `map_err`?

PR #6184, opened 2026-04-30
Enable the anvil_deposit_proof.

MathieuDutSik:
## Motivation

Following previous work, it appeared that the `anvil_process_proof` test was not enabled.

## Proposal

Do so.

## Test Plan

CI is the point.

## Release Plan

Could be backported if it works.

## Links

None

### DIFF
--- a/.github/workflows/bridge-e2e.yml
+++ b/.github/workflows/bridge-e2e.yml
@@ -108,3 +108,31 @@ jobs:
         env:
           RUSTFLAGS: ""
         run: cargo test -- --ignored --nocapture
+
+      - name: Install Foundry (provides `anvil`)
+        uses: foundry-rs/foundry-toolchain@v1
+
+      - name: Ensure Solc Directory Exists
+        run: mkdir -p /home/runner/.solc
+      - name: Cache Solc
+        id: cache-solc
+        uses: actions/cache@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/bridge-e2e.yml
```diff
@@ -108,3 +108,31 @@ jobs:
...
+      - name: Add Solc to PATH
+        run: echo "/home/runner/.solc" >> $GITHUB_PATH
+      - name: Create a Solc symbolic link
+        run: ln -sf /home/runner/.solc/v0.8.33/solc-static-linux /home/runner/.solc/solc
+
+      - name: Run anvil deposit-proof test
+        working-directory: linera-bridge
+        env:
+          RUSTFLAGS: ""
+        run: cargo test -p linera-bridge --test anvil_deposit_proof -- --ignored --nocapture
```
deuszx, line 138: I think we can drop the `--test anvil_deposit_proof` part in case we add more tests in the future.

MathieuDutSik: If we were to do that, then cargo test -p linera-bridge -- --ignored would also run generate_base_block_fixture — which would either be slow/flaky
  (public Base RPC) or just plain fail in CI.

PR #6185, opened 2026-04-30
Address a bug in `linera-bridge` and restructure the way we handle by using a Write Ahead Log.

MathieuDutSik:
## Motivation

The EVM and Linera scan loops used bounded mpsc channels (capacity 64) to hand pending requests to the retry/processor loop. When the channel filled up, the scanner used try_send and silently dropped the request, then advanced its watermark — making the loss unrecoverable on the next pass.

## Proposal

Tather than reintroduce backpressure, remove the channels entirely. SQLite (already a write-through store via MonitorState) becomes the work queue. A tokio::sync::Notify provides low-latency wakeup so the consumer doesn't have to wait a full poll interval after a new item arrives. As a side effect, this also fixes a separate bug — pending requests in flight at shutdown were never reloaded into memory on startup.

## Test Plan

CI

## Release Plan

## Links

None

### DIFF
--- a/linera-bridge/src/monitor/evm.rs
+++ b/linera-bridge/src/monitor/evm.rs
@@ -7,7 +7,7 @@
 use std::{sync::Arc, time::Duration};
 
 use alloy::providers::Provider;
-use tokio::sync::RwLock;
+use tokio::sync::{Notify, RwLock};
 
 use super::{MonitorState, PendingDeposit};
 use crate::{
@@ -16,18 +16,18 @@ use crate::{
 };
 
 /// Background task that polls EVM for `DepositInitiated` events and checks
-/// Linera for completion.
+/// Linera for completion. Newly-discovered deposits are written to
+/// `MonitorState` (and SQLite) and the consumer is woken via `notify`.
 pub async fn evm_scan_loop<E: linera_core::environment::Environment + 'static>(
     monitor: Arc<RwLock<MonitorState>>,
     evm_client: Arc<EvmClient<impl Provider + 'static>>,
     linera_client: Arc<LineraClient<E>>,
-    pending_deposit_tx: tokio::sync::mpsc::Sender<PendingDeposit>,
+    deposit_notify: Arc<Notify>,
     scan_interval: Duration,
-    max_retries: u32,
 ) {
     loop {
         let (scan_result, completion_result) = tokio::join!(
-            evm_scan_iteration(&monitor, &evm_client, &pending_deposit_tx),
+            evm_scan_iteration(&monitor, &evm_client, &deposit_notify),
             check_deposit_completion(&monitor, &linera_client),
         );
 
@@ -36,22 +36,6 @@ pub async fn evm_scan_loop<E: linera_core::environment::Environment + 'static>(
         }
         if let Err(error) = completion_result {
             tracing::warn!(error = ?error, "Deposit completion check failed");
-        }
-
-        // Re-enqueue deposits that are eligible for retry.
-        {
-            let state = monitor.read().await;
-            for d in state.deposits_ready_for_retry(max_retries) {
-                if let Err(error) = pending_deposit_tx.try_send(PendingDeposit {
-                    key: d.value.key.clone(),
-                    tx_hash: d.value.tx_hash,
-                    depositor: d.value.depositor,
-                    amount: d.value.amount,
-                    nonce: d.value.nonce,
-                }) {
-                    tracing::warn!(?error, "Failed to enqueue deposit for retry");
-                }
-            }
         }
 
         let summary = monitor.read().await.status_summary();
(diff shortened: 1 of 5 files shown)


linera-bridge/src/monitor/mod.rs
```diff
@@ -330,6 +330,63 @@ struct GqlRequest {
...
+
+/// Symmetric counterpart of [`enqueue_or_track_deposit`] for the Linera→EVM
+/// burn pipeline.
+pub(crate) async fn enqueue_or_track_burn(
+    state: &RwLock<MonitorState>,
+    pending_tx: &tokio::sync::mpsc::Sender<PendingBurn>,
+    pending: PendingBurn,
+) {
+    use tokio::sync::mpsc::error::TrySendError;
+    match pending_tx.try_send(pending) {
```
deuszx: Nice find! Maybe we should remove the channels entirely and use the storage as the synchronization layer 🤔 i.e. write to the `state` (`track_burn`/`track_deposit`) on one side and read from it on the processing side. That way we'd defacto implement a WAL but use it as the queue as well.

MathieuDutSik: Done.

linera-bridge/src/monitor/evm.rs
```diff
@@ -64,30 +50,30 @@ pub async fn evm_scan_loop<E: linera_core::environment::Environment + 'static>(
...
-                }
-                d.last_retry_at = Some(std::time::Instant::now());
-            } else {
-                state.track_deposit(pending.clone()).await;
+    loop {
+        let pending = monitor.read().await.next_deposit_for_retry(max_retries);
+        let Some(pending) = pending else {
+            tokio::select! {
+                _ = notify.notified() => {}
+                _ = tokio::time::sleep(poll_interval) => {}
```
deuszx: Let's add some logging here:
```rust
_ = notify.notified() => {
    trace!("EVM deposits processor notified about new pending item")
}
```

and something similar for the `tokio` sleep arm. So that we can debug it bit easier when we need.

linera-bridge/src/monitor/linera.rs
```diff
@@ -69,34 +53,29 @@ pub async fn linera_scan_loop<E: linera_core::environment::Environment + 'static
...
-                    continue;
-                }
-                b.last_retry_at = Some(Instant::now());
-            } else {
-                state.track_burn(pending).await;
+    loop {
+        let pending = monitor.read().await.next_burn_for_retry(max_retries);
+        let Some(pending) = pending else {
+            tokio::select! {
+                _ = notify.notified() => {}
```
deuszx: Same `TRACE` level logging as in the `evm.rs` please.

linera-bridge/src/monitor/evm.rs
```diff
@@ -185,24 +170,29 @@ async fn evm_scan_iteration(
...
+            .await;
+        tracked_any |= was_new;
     }
 
-    let mut state = monitor.write().await;
-    state.last_scanned_evm_block = current_block;
+    {
+        let mut state = monitor.write().await;
+        state.last_scanned_evm_block = current_block;
+    }
```
deuszx: Do we need those `brackets`? They weren't necessary before.

linera-bridge/src/monitor/mod.rs
```diff
@@ -516,4 +573,161 @@ mod tests {
...
+        assert!(state.next_burn_for_retry(10).is_none());
+
+        // Once forwarded, the item is no longer offered for retry.
+        state.complete_burn(height, 0).await;
+        assert!(state.next_burn_for_retry(10).is_none());
+    }
+
+    /// Tombstone reference for the historical channel-full bug. With the
+    /// channels removed there is no failure mode to pin; this test is a
+    /// no-op that documents what used to be regression-tested.
```
deuszx: Should we remove it now then?

deuszx:
Please update the PR description

linera-bridge/src/monitor/evm.rs
MathieuDutSik: Ok, done.

linera-bridge/src/monitor/linera.rs
MathieuDutSik: Ok, done.

linera-bridge/src/monitor/evm.rs
MathieuDutSik: Not needed, thanks.

linera-bridge/src/monitor/mod.rs
MathieuDutSik: Yes.

deuszx (approved):
Thanks for this. As a follow-up now maybe we should split the table into "pending" and "finished" tasks. Otherwise the `SELECT`s will start getting slower and slower.

PR #6186, opened 2026-04-30
[testnet] admin_committee: chain-state fallback on storage miss

ndr-ds:
## Motivation

PR #6122 changed `Client::admin_committee` and `Client::admin_committees` to read committees via `Storage::get_or_load_committee` (shared cache, then `EPOCH_STREAM_NAME` events plus committee blob), instead of the pre-#6122 path of reading from the admin chain `system::committees` view in chain state. On a `testnet_conway` faucet PVC the new path returns `None` (one of `NetworkDescription`, the epoch event, or the committee blob is not present in this storage), so the function bails with `LocalNodeError::InactiveChain(admin_chain_id)` and the faucet crash-loops on every startup.

The pre-#6122 chain-state path still works on this storage. The rolled-back image proves it. PR #6114 (which removes `system::committees` from chain state) is on `main` but not on `testnet_conway`, so the chain-state field is still populated here.

## Proposal

In `Client::admin_committee` and `Client::admin_committees`, try the shared-cache and storage path first. On a miss, fall back to the admin chain `system::committees` view (the pre-#6122 source) and seed the shared cache from it so subsequent calls hit the fast path. The healthy case pays no extra cost; the slow path only fires when the cache and storage path both miss.

This is a transitional fix. It depends on the admin chain `system::committees` view still being populated. Before #6114 is ported to `testnet_conway`, this fallback must be replaced by an admin-chain backfill mechanism that fetches the missing events and committee blobs from validators.

## Test Plan

CI

## Release Plan

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -812,6 +812,11 @@ impl<Env: Environment> Client<Env> {
 
     /// Returns all committees known to the process at epochs up to and including the admin
     /// chain's current epoch, together with that epoch.
+    ///
+    /// On a miss in the shared cache / storage path (introduced in #6122), falls back to the
+    /// admin chain's own `system::committees` view in chain state and seeds the shared cache
+    /// from it. Transitional: depends on chain-state committees still being populated, so
+    /// must be revisited if/when #6114 is ported to testnet.
     #[instrument(level = "trace", skip_all)]
     async fn admin_committees(
         &self,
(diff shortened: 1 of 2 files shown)


linera-core/src/client/mod.rs
```diff
@@ -812,6 +812,11 @@ impl<Env: Environment> Client<Env> {
 
     /// Returns all committees known to the process at epochs up to and including the admin
     /// chain's current epoch, together with that epoch.
+    ///
+    /// On a miss in the shared cache / storage path (introduced in #6122), falls back to the
+    /// admin chain's own `system::committees` view in chain state and seeds the shared cache
+    /// from it. Transitional: depends on chain-state committees still being populated, so
+    /// must be revisited if/when #6114 is ported to testnet.
```
afck: We're not porting that to testnet, and `committees` will stay, so that's fine.

afck (approved):
I imagine that should work around it.

But we should add logging to find out what's actually missing in storage: the event, the committee… or the network description?

afck (approved):

PR #6188, opened 2026-04-30
Replace `let _ = ...` with explicit handling and deny the lints

afck:
## Motivation

We try to avoid `let _ =` in the code, but it's still present in a few places.

There's no Clippy lint for that in general, but for some special cases.

## Proposal

Deny the `let_underscore_untyped` and `let_underscore_must_use` Clippy lints.

Fix all `let _ =` places: log on failure where actionable (channel sends, oneshot responses, races in tests), `unwrap()` in benches, and document the "this can't fail / failure is expected" cases at the discard site.

## Test Plan

CI

## Release Plan

## Links


linera-service/src/controller.rs
```diff
@@ -176,7 +176,8 @@ where
             pending_services
                 .retain(|_, pending_service| pending_service.start_block_height > height);
             if pending_services.is_empty() {
-                let _ = self.pending_services_notifications.remove(&chain_id);
+                // We don't need the removed Notify handle; an absent entry is fine too.
+                drop(self.pending_services_notifications.remove(&chain_id));
```
ma2bd: do we need this `drop`?

Also, `drop` being polymorphic, doesn't it have the same caveat as `let _ =` ?

afck: Right, here we could probably just have neither. This would be fine, would it?

```suggestion
                self.pending_services_notifications.remove(&chain_id);
```

ma2bd: I think so. And again do we ever want `drop` ?

afck: I removed them.

linera-bridge/src/relay/mod.rs
```diff
@@ -502,7 +502,9 @@ async fn serve_loop<E: linera_core::environment::Environment + 'static>(
                             let (certs, _) = chain_client.process_inbox().await?;
                             Ok(certs)
                         }.await;
-                        let _ = response.send(result);
+                        if response.send(result).is_err() {
+                            tracing::debug!("ProcessInbox response receiver dropped");
+                        }
```
MathieuDutSik, line 507: Isn't there an explicit error here that can be printed?

linera-bridge/src/relay/mod.rs
```diff
@@ -546,7 +548,9 @@ async fn serve_loop<E: linera_core::environment::Environment + 'static>(
                             };
                             Ok(())
                         }.await;
-                        let _ = response.send(result);
+                        if response.send(result).is_err() {
+                            tracing::debug!("ProcessDeposit response receiver dropped");
+                        }
```
MathieuDutSik, line 553: Same

linera-core/src/worker.rs
```diff
@@ -948,12 +948,15 @@ where
...
-        let _ = pin.remove_if(&chain_id, |_key, future| {
+        let removed = pin.remove_if(&chain_id, |_key, future| {
             future
                 .peek()
                 .and_then(|r| r.clone().ok())
                 .is_some_and(|weak| weak.ptr_eq(&weak_poisoned))
         });
+        if removed.is_err() {
+            tracing::trace!(%chain_id, "Poisoned worker entry already replaced; skipping eviction");
+        }
```
MathieuDutSik, line 959: Same

MathieuDutSik (approved):
Approved since it is unconditionally better.
I think it would be better to print the errors when we have some, but maybe there are arguments against.

linera-bridge/src/relay/mod.rs
afck: No: `send(x)` returns `Ok(())` if successful or `Err(x)` if the receiver was dropped. There's no error type, only the returned payload.

PR #6196, opened 2026-05-01
Change what `uncompressedKeys` are being sent

MathieuDutSik:
## Motivation

The `uncompressedKeys` are being sent unordered, which leads to a reordering being done
onchain and a `O(n^2)` algorithm onchain, which is just bad.

## Proposal

Do, the reordering offchain before submitting it.

## Test Plan

CI

## Release Plan

## Links

None


deuszx (dismissed):

linera-bridge/src/solidity/LightClient.sol
```diff
@@ -224,30 +224,6 @@ contract LightClient {
...
-            if (used[j]) continue;
-            if (uncompressedKeys[j].length != 64) continue;
-            // Compare x-coordinate: blob[keyPos+1..keyPos+33] vs uncompressed[0..32]
-            bool match_ = true;
-            for (uint256 k = 0; k < 32; k++) {
-                if (blob[keyPos + 1 + k] != uncompressedKeys[j][k]) {
-                    match_ = false;
-                    break;
-                }
-            }
```
deuszx, line 245: OK, I wanted to verify this is not getting removed. We still do it in `_verifyKeyCompression`:
```
// Check x-coordinate matches (bytes 1..33 of compressed == bytes 0..32 of uncompressed)
for (uint256 i = 0; i < 32; i++) {
     require(blob[keyPos + 1 + i] == uncompressed[i], "key x-coordinate mismatch");
}
```

deuszx (approved):

linera-bridge/src/solidity/LightClient.sol
MathieuDutSik: It is correct as the error would be caught later in `_setCommittee committee.weights[validators[i]]) == 0`.

deuszx: Not necessarily, it could just match a different key. The checks we do in _verifyKeyCompression cover this though.

PR #6198, opened 2026-05-01
[testnet] Enable application introspection

MathieuDutSik:
## Motivation

It would be useful to be able to see the content of the messages, events and operations when using `linera-explorer`
instead of having it only for 

## Proposal

The whole difficulty is in making the data "Formats" available to `linera-explorer`. This is a PR against `testnet_conway`
so the design space is limited and application already exist.

The chosen design is to have a registry which contains the data:
* It is a very fragile design. It requires passing a chain ID and an application ID to provide to the `linera-explorer`.
* But it is compatible with `testnet_conway` and with existing applications. No need for additional deployments. And the registry chain can be its own.
* The data is stored as a `MapView<ModuleId, Vec<u8>>`. We cannot store as Blobs since we cannot return `None` on Blob access. Anyway, not an issue, the formats are small stuff.
* It is of course an antidesign since we have a map from `ModuleId` something Linera wide to a specific chain, but we do with the situation.

Alternative Designs:
* Add an entry `__formats` to the GraphQL design so that we can get the "Formats" from the service. Compatible with `testnet_conway` but not with existing applications.
* Add more blobs to the construction of the app. Currently, we have `Contract` and `Service`. We could have `Formats`.

From `linera-explorer` things are easier:
* The data is read from the registry.
* The `json_controller` is used to show the result.
* The JSON is parsed so that adresses and chain IDs are shown in hexadecimal.
* If absent for an app, only the hexadecimal. If present, show the JSON.

## Test Plan

CI

Now, doing local tests show the following.

In the "Applications" screen we now have an entry Formats. For two applications (social and fungible) we have a Formats. For one, ironically the register, we do not have a format:
Formats1_presence

Now, decoding of an application operation:
Formats2_decoded_operation

The `ChainId` and `Address20/Address32` are shown in the same way as in Linera (No "0x" for chain IDs, "0x" for addresses):
Formats6_chain_id_address

Operations done from the `linera-explorer` show exactly what was done:
Formats5_Mutations

Now, an event. First not clicked upon:
Formats3_events_unclicked

Then clicking on the short event shows what we have:
Formats4_events_clicked

## Release Plan

This is for `testnet_conway`. How to design it for `main` is a whole other question entirely.
There are many moving entries to consider.

## Links

None


deuszx:
If I'm seeing this correctly, we're not deserializing the names of the enums/variants of the messages/operations, etc but we should be able to do that as they are part of the format yaml
```
WrappedFungibleOperation:
  ENUM:
    0:
      Balance:
        STRUCT:
          - owner:
              TYPENAME: AccountOwner
    1:
      TickerSymbol: UNIT
    2:
      Approve:
        STRUCT:
          - owner:
              TYPENAME: AccountOwner
          - spender:
              TYPENAME: AccountOwner
          - allowance:
              TYPENAME: Amount
    ...
```

MathieuDutSik:
> If I'm seeing this correctly, we're not deserializing the names of the enums/variants of the messages/operations, etc but we should be able to do that as they are part of the format yaml
> 
> ```
> WrappedFungibleOperation:
>   ENUM:
>     0:
>       Balance:
>         STRUCT:
>           - owner:
>               TYPENAME: AccountOwner
>     1:
>       TickerSymbol: UNIT
>     2:
>       Approve:
>         STRUCT:
>           - owner:
>               TYPENAME: AccountOwner
>           - spender:
>               TYPENAME: AccountOwner
>           - allowance:
>               TYPENAME: Amount
>     ...
> ```

In the plot that I put in the PR description for `Event` it seems to match quite well with
```rust
pub enum Event {
    /// A new post was created
    Post { post: OwnPost, index: u32 },
    /// A user liked a post
    Like { key: Key },
    /// A user commented on a post
    Comment { key: Key, comment: String },
}
```

So that seems to match pretty well to an `Event::Post` entry.

deuszx:
> In the plot that I put in the PR description for `Event` it seems to match quite well with
> 
> ```rust
> pub enum Event {
>     /// A new post was created
>     Post { post: OwnPost, index: u32 },
>     /// A user liked a post
>     Like { key: Key },
>     /// A user commented on a post
>     Comment { key: Key, comment: String },
> }
> ```
> 
> So that seems to match pretty well to an `Event::Post` entry.

You're right. I misread the screenshots. They could be presented more nicely but it can be done in a separate PR.

linera-client/src/client_context.rs
```diff
@@ -926,6 +926,222 @@ impl<Env: Environment> ClientContext<Env> {
     }
 }
 
+#[cfg(all(feature = "fs", not(web)))]
+impl<Env: Environment> ClientContext<Env> {
+    /// Publishes a module along with the JSON-encoded `Formats` description loaded
+    /// from `snap_path`. The module publication and the formats-registry write
+    /// happen atomically in a single block.
+    pub async fn publish_bcs_module(
```
ma2bd: Adding `_bcs_` in the middle of the function name is not helpful. Here and below, I would suggest something like `publish_module_with_formats`

ma2bd: This being said, we would ideally avoid duplicating the publishing code in the first place. (Perhaps this can be done in a later PR)

linera-client/src/client_context.rs
```diff
@@ -926,6 +926,222 @@ impl<Env: Environment> ClientContext<Env> {
...
+        &mut self,
+        chain_client: &ChainClient<Env>,
+        contract: PathBuf,
+        service: PathBuf,
+        vm_runtime: VmRuntime,
+        snap_path: PathBuf,
```
ma2bd: nit: `snap_path` => `formats` ?

examples/formats-registry/src/state.rs
```diff
@@ -0,0 +1,14 @@
...
+use linera_sdk::{
+    linera_base_types::ModuleId,
+    views::{linera_views, MapView, RootView, ViewStorageContext},
+};
+
+#[derive(RootView)]
+#[view(context = ViewStorageContext)]
+pub struct FormatsRegistryState {
+    /// Maps a `ModuleId` to the bytes of its registered formats description.
+    pub formats: MapView<ModuleId, Vec<u8>>,
```
ma2bd, line 13: Perhaps a comment should clarify that we expect the format of a module to be "registered" on the same chain that created the module.

ma2bd: Alternatively, another design would be to use cross-chain messages to group all the format definitions of (a family of) applications on the same chain (likely the creation chain of this app).

linera-sdk/src/formats.rs
```diff
@@ -31,3 +34,191 @@ pub trait BcsApplication {
...
+/// and the container `registry`.
+pub fn bcs_to_json(
+    bytes: &[u8],
+    format: &Format,
+    registry: &Registry,
+) -> bcs::Result<serde_json::Value> {
+    let context = DeserializationContext {
+        format: format.clone(),
+        registry,
+        environment: &EmptyEnvironment,
```
ma2bd, line 48: If we use an `EmptyEnvironment`, primitive types like `AccountOwner` will be displayed as a vector of bytes. Perhaps that's fine for now

CLI.md
```diff
@@ -29,6 +29,7 @@ This document contains the help content for the `linera` command-line program.
 * [`linera service`↴](#linera-service)
 * [`linera faucet`↴](#linera-faucet)
 * [`linera publish-module`↴](#linera-publish-module)
+* [`linera publish-bcs-module`↴](#linera-publish-bcs-module)
```
ma2bd: publish-module-with-formats

ma2bd: (but once again we should not create another command ideally)

linera-sdk/src/lib.rs
```diff
@@ -33,7 +33,7 @@ pub mod contract;
 #[cfg(feature = "ethereum")]
 pub mod ethereum;
 mod extensions;
-#[cfg(not(target_arch = "wasm32"))]
+#[cfg(feature = "format-decode")]
```
ma2bd: No need to create a new feature for this.

linera-explorer/src/lib.rs
```diff
@@ -95,6 +96,17 @@ pub struct Config {
     indexer: String,
     node: String,
     tls: bool,
+    /// Hex-encoded chain id where the formats registry application is deployed.
+    /// Must be set together with [`Self::formats_registry_app_id`]. Stored as a
+    /// raw string so the JS-side `v-model` round-trips cleanly through
+    /// `serde_wasm_bindgen`; we validate the format only when actually using
+    /// the value.
+    #[serde(default)]
+    formats_registry_chain: Option<String>,
```
ma2bd, line 105: Actually here, it seems like we assume that all the formats go to a unique chain. (See also comment in `examples/formats-registry`)

examples/formats-registry/src/contract.rs
```diff
@@ -0,0 +1,71 @@
...
+                self.state
+                    .formats
+                    .insert(&module_id, value)
+                    .expect("storage");
+            }
+        }
+    }
+
+    async fn execute_message(&mut self, _message: ()) {
+        panic!("formats-registry does not support cross-chain messages");
```
ma2bd, line 62: See other comments. It would appear that we need messages to co-locate all formats on a single chain.

examples/formats-registry/src/state.rs
ma2bd: Ok so apparently we assume the latter

linera-sdk/src/formats.rs
MathieuDutSik: I thought about the problem of the conversion for `AccountOwner`, `ChainId`, and so on.

I did not want to modify the formats nor change `serde-reflection` because this expose to some complexity.

Instead, since it is really about showing up the JSON, the changes to the JSON are being done in the frontend. I think that this is the easiest place for that.
So, the following heuristic is applied:
* If an array has length N of integer between 0 and 255 with N in 20, 32, and 64 then we do the conversion to hexdecimal. I do not think it can ever hurt.
* If, additionally, the entry is a `Address20` or `Address32` then we add as prefix a "0x".

I think this is the right place for doing the modification. It is a frontend issue, so putting it in the frontend seems logical.

---------

Next question, could we replace the `EmptyEnvironment` with the relevant environment in order to force the behavior we want?
Yes, in principle. But some changes to `serde-reflection` would be needed. Currently the behavior of `json_converter.rs` is:
```rust
            TypeName(name) => {
                if let Some(container_format) = self.registry.get(&name) {
                    .....
                } else {
                    Ok(self.environment .......)
                }
            }
```

Since the `AccountOwner` is present in the registry, the environment would never be called.

So, more works needed for that approach to work. It is a nice fix, but not at all urgent.

CLI.md
MathieuDutSik: Ok, for `publish-module-with-formats`.

Now, about the end-design, this is another story.
Conceivably, we could have 2 blobs for ordinary applications and 3 blobs for applications with a `Formats`. That design would map fairly nicely with a `publish-module` / `publish-module-with-formats`.

But this is for further PRs.

linera-explorer/src/lib.rs
MathieuDutSik: Yes, we work with a single chain for the formats.
I think this is kind of appropriate for `testnet_conway`. Not perfect by any means but major advantages:
* No need for app deployment again because we only need to create a formats-chain with the entries
* Fully compatible with `testnet_conway`.

So, I am afraid that we have two extremes:
* Clumsy but backward compatible (that is `testnet_conway`)
* Not clumsy but breaking (that is `main` with blobs)

I concentrated here on the first, and I do not think we can do much better.

examples/formats-registry/src/contract.rs
MathieuDutSik: The formats are accessed by queries from `linera-explorer`. Since we cannot implement messages in service queries, I am not sure how that would help.

examples/formats-registry/src/state.rs
MathieuDutSik: Well, no.
Basically, I assume we would have a chain "format-registry".

linera-sdk/src/lib.rs
MathieuDutSik: Corrected.

linera-client/src/client_context.rs
MathieuDutSik: Done.

examples/formats-registry/src/state.rs
ma2bd: But you can't assume that everyone is an owner of that chain

linera-sdk/src/formats.rs
ma2bd: > I did not want to modify the formats nor change serde-reflection because this expose to some complexity.

That's fine for this PR but this is exactly the plan.

ma2bd: > some changes to serde-reflection would be needed.

I very much doubt it

ma2bd: > Since the AccountOwner is present in the registry, the environment would never be called.

It's trivial to remove `AccountOwner` etc from the registry. This will also saves space on-chain

examples/formats-registry/src/contract.rs
ma2bd: There will be a message from the Module-publishing chain to the registry-chain

ma2bd (approved):
Let's land this as it is and fix forward in follow-up PRs.

PR #6201, opened 2026-05-01
Stream all available blobs in DownloadBlobs and skip missing

ndr-ds:
## Motivation

The streaming `DownloadBlobs` RPC added in #5976 terminates the response
stream on the first missing blob, dropping any blobs that would have
been yielded after it:

- gRPC (`linera-service/src/proxy/grpc.rs`): the handler yields
  `Err(Status::not_found(...))` for a missing blob, which terminates
  the gRPC response stream. Subsequent blobs queued for that batch are
  silently lost.
- Simple transport (`linera-service/src/proxy/main.rs`): the handler
  interleaves `RpcMessage::Error(BlobsNotFound([single_id]))` items
  among `DownloadBlobResponse` items. The current consumer in
  `RemoteNode::download_blobs` propagates the first error via `?`,
  which drops the rest of the batch on the floor.

A batch of N blobs where blob K is missing returns at most K successful
blobs to the caller, even if the validator has all of blobs K+1..N
available. The follow-up #6157 wires this RPC into the per-peer
fallback path and relies on partial progress being reported correctly,
which the current broken server semantics prevent.

## Proposal

Both server handlers now yield only the blobs that are present, in
request order, and silently skip missing blobs. The client compares
the IDs it received against the IDs it asked for and re-requests any
missing IDs from another peer.

This preserves the existing wire format on both transports.
Storage-level errors that occur before any blob has been yielded
continue to surface as a top-level error on the response (gRPC) or as
an early `RpcMessage::Error` (simple transport).

Also addresses the review comment from afck on #6105 / #6155: the
type annotation on `created_blob_ids` in
`Client::process_certificates` moves from the `let` binding to the
`.collect()` call (turbofish style).

## Test Plan

CI.

## Release Plan

  alongside #5976. #6156 is the WIP backport of #5976 and should be
  updated to include this fix before being merged.

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -843,15 +843,15 @@ impl<Env: Environment> Client<Env> {
         // block bodies, so they don't need to be fetched from a validator. The
         // chain worker resolves them from `Block::created_blobs()` during
         // `handle_certificate`.
-        let created_blob_ids: BTreeSet<BlobId> = certificates
+        let created_blob_ids = certificates
             .iter()
             .flat_map(|certificate| certificate.value().block().created_blob_ids())
-            .collect();
-        let required_blob_ids: Vec<_> = certificates
+            .collect::<BTreeSet<BlobId>>();
+        let required_blob_ids = certificates
             .iter()
             .flat_map(|certificate| certificate.value().required_blob_ids())
             .filter(|blob_id| !created_blob_ids.contains(blob_id))
-            .collect();
+            .collect::<Vec<_>>();
 
         match self
             .local_node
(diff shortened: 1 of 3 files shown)


linera-core/src/client/mod.rs
```diff
@@ -786,10 +788,10 @@ impl<Env: Environment> Client<Env> {
         // block bodies, so they don't need to be fetched from a validator. The
         // chain worker resolves them from `Block::created_blobs()` during
         // `handle_certificate`.
-        let created_blob_ids: BTreeSet<BlobId> = certificates
+        let created_blob_ids = certificates
             .iter()
             .flat_map(|certificate| certificate.value().block().created_blob_ids())
-            .collect();
+            .collect::<BTreeSet<BlobId>>();
         let required_blob_ids: Vec<_> = certificates
```
afck: (Here's another type annotation.)

afck (approved):

PR #6202, opened 2026-05-01
Implement forward and reverse iterators for the `find_{keys,key_values}_by_prefix`

MathieuDutSik:
## Motivation

The `find_{keys,key_values}_by_prefix` functions are building sets that are intrinsically big.
Async iterators are actually the right solution here.

## Proposal

Add the functions in the trait and a default implementation. The default implementation is used for the `linera-storage-service`, `indexed-db`, `System-api` and `ViewContainer`.

The types are inspired from https://github.com/linera-io/linera-protocol/pull/4975
The reverse iterators are inspired from https://github.com/linera-io/linera-protocol/pull/6183 which is itself inspired by https://github.com/linera-io/linera-protocol/pull/6171

For `RocksDB`, `DynamoDB`, `ScyllaDB`, the functionality essentially already exists in the code.
The `ValueSplitting is the more challenging part.

## Test Plan

CI

Tests have been added to the `run_reads` functions.

## Release Plan

## Links

None

### DIFF
--- a/linera-views/src/backends/journaling.rs
+++ b/linera-views/src/backends/journaling.rs
@@ -201,12 +201,11 @@ where
         &'a self,
         key_prefix: &'a [u8],
     ) -> FindKeysStream<'a, Self::Error> {
-        Box::pin(async_stream::stream! {
-            let mut stream = self.store.find_keys_by_prefix_iter(key_prefix);
-            while let Some(item) = stream.next().await {
-                yield item.map_err(JournalingError::Inner);
-            }
-        })
+        Box::pin(
+            self.store
+                .find_keys_by_prefix_iter(key_prefix)
+                .map(|item| item.map_err(JournalingError::Inner)),
+        )
     }
 
     async fn find_key_values_by_prefix(
@@ -220,36 +219,33 @@ where
         &'a self,
         key_prefix: &'a [u8],
     ) -> FindKeyValuesStream<'a, Self::Error> {
-        Box::pin(async_stream::stream! {
-            let mut stream = self.store.find_key_values_by_prefix_iter(key_prefix);
-            while let Some(item) = stream.next().await {
-                yield item.map_err(JournalingError::Inner);
-            }
-        })
+        Box::pin(
+            self.store
+                .find_key_values_by_prefix_iter(key_prefix)
+                .map(|item| item.map_err(JournalingError::Inner)),
+        )
     }
 
     fn find_keys_by_prefix_rev_iter<'a>(
         &'a self,
         key_prefix: &'a [u8],
     ) -> FindKeysStream<'a, Self::Error> {
-        Box::pin(async_stream::stream! {
-            let mut stream = self.store.find_keys_by_prefix_rev_iter(key_prefix);
-            while let Some(item) = stream.next().await {
-                yield item.map_err(JournalingError::Inner);
-            }
-        })
+        Box::pin(
+            self.store
+                .find_keys_by_prefix_rev_iter(key_prefix)
+                .map(|item| item.map_err(JournalingError::Inner)),
+        )
     }
 
     fn find_key_values_by_prefix_rev_iter<'a>(
         &'a self,
         key_prefix: &'a [u8],
     ) -> FindKeyValuesStream<'a, Self::Error> {
-        Box::pin(async_stream::stream! {
-            let mut stream = self.store.find_key_values_by_prefix_rev_iter(key_prefix);
-            while let Some(item) = stream.next().await {
-                yield item.map_err(JournalingError::Inner);
-            }
-        })
+        Box::pin(
+            self.store
+                .find_key_values_by_prefix_rev_iter(key_prefix)
+                .map(|item| item.map_err(JournalingError::Inner)),
+        )
     }
 }
 
--- a/linera-views/src/backends/journaling.rs
+++ b/linera-views/src/backends/journaling.rs
@@ -19,6 +19,7 @@
 //! time the data in a block are written, the journal header is updated in the same
 //! transaction to mark the block as processed.
 
+use futures::stream::StreamExt;
 use serde::{Deserialize, Serialize};
 use static_assertions as sa;
 use thiserror::Error;
@@ -26,8 +27,8 @@ use thiserror::Error;
 use crate::{
     batch::{Batch, BatchValueWriter, DeletePrefixExpander, SimplifiedBatch},
     store::{
-        DirectKeyValueStore, KeyValueDatabase, KeyValueStoreError, ReadableKeyValueStore,
-        WithError, WritableKeyValueStore,
+        DirectKeyValueStore, FindKeyValuesStream, FindKeysStream, KeyValueDatabase,
+        KeyValueStoreError, ReadableKeyValueStore, WithError, WritableKeyValueStore,
     },
     views::MIN_VIEW_TAG,
 };
@@ -196,11 +197,35 @@ where
         Ok(self.store.find_keys_by_prefix(key_prefix).await?)
     }
 
+    fn find_keys_by_prefix_iter<'a>(
+        &'a self,
+        key_prefix: &'a [u8],
+    ) -> FindKeysStream<'a, Self::Error> {
+        Box::pin(async_stream::stream! {
+            let mut stream = self.store.find_keys_by_prefix_iter(key_prefix);
+            while let Some(item) = stream.next().await {
+                yield item.map_err(JournalingError::Inner);
+            }
+        })
+    }
+
     async fn find_key_values_by_prefix(
         &self,
         key_prefix: &[u8],
     ) -> Result<Vec<(Vec<u8>, Vec<u8>)>, Self::Error> {
         Ok(self.store.find_key_values_by_prefix(key_prefix).await?)
+    }
+
+    fn find_key_values_by_prefix_iter<'a>(
+        &'a self,
+        key_prefix: &'a [u8],
+    ) -> FindKeyValuesStream<'a, Self::Error> {
+        Box::pin(async_stream::stream! {
+            let mut stream = self.store.find_key_values_by_prefix_iter(key_prefix);
+            while let Some(item) = stream.next().await {
+                yield item.map_err(JournalingError::Inner);
+            }
+        })
     }
 }
 
--- a/linera-views/src/backends/journaling.rs
+++ b/linera-views/src/backends/journaling.rs
@@ -227,6 +227,30 @@ where
             }
         })
     }
+
+    fn find_keys_by_prefix_rev_iter<'a>(
+        &'a self,
+        key_prefix: &'a [u8],
+    ) -> FindKeysStream<'a, Self::Error> {
+        Box::pin(async_stream::stream! {
+            let mut stream = self.store.find_keys_by_prefix_rev_iter(key_prefix);
+            while let Some(item) = stream.next().await {
+                yield item.map_err(JournalingError::Inner);
+            }
+        })
+    }
+
+    fn find_key_values_by_prefix_rev_iter<'a>(
+        &'a self,
+        key_prefix: &'a [u8],
+    ) -> FindKeyValuesStream<'a, Self::Error> {
+        Box::pin(async_stream::stream! {
+            let mut stream = self.store.find_key_values_by_prefix_rev_iter(key_prefix);
+            while let Some(item) = stream.next().await {
+                yield item.map_err(JournalingError::Inner);
+            }
+        })
+    }
 }
 
 impl<D> KeyValueDatabase for JournalingKeyValueDatabase<D>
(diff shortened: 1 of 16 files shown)


ma2bd:
If I had to choose, I think I prefer this one over #6183.

To which extend do you the tests cover the new code? Especially value-splitting and LRU-caching?

MathieuDutSik:
> If I had to choose, I think I prefer this one over #6183.
> 
> To which extend do you the tests cover the new code? Especially value-splitting and LRU-caching?

The test `test_lru_cache_serves_find_by_prefix` does the test when the entry is present in the cache.

The test `test_value_splitting4_find_key_iters_with_leftovers` does the test of `ValueSplitting` when there are some leftover keys.

The `run_reads` does the systematic check of writing `(key-values)` to storage and exercising the read functionality. It is in my opinion fairly good, but does not have perfect coverage.

Note that PR #6183 also implements some function for `MapView` that are not in this PR.

linera-views/src/test_utils/mod.rs
```diff
@@ -202,6 +205,36 @@ pub async fn run_reads<S: KeyValueStore>(store: S, key_values: Vec<(Vec<u8>, Vec
             }
         }
         assert_eq!(set_key_value1, set_key_value2);
+        // Streaming variants must agree with the eager methods.
+        let keys_iter: Vec<Vec<u8>> = store
```
afck: The policy is to put type annotations like these on the method instead (`try_collect` in this case); also below.

linera-views/src/backends/journaling.rs
```diff
@@ -196,12 +197,60 @@ where
...
+    fn find_keys_by_prefix_iter<'a>(
+        &'a self,
+        key_prefix: &'a [u8],
+    ) -> FindKeysStream<'a, Self::Error> {
+        Box::pin(async_stream::stream! {
+            let mut stream = self.store.find_keys_by_prefix_iter(key_prefix);
+            while let Some(item) = stream.next().await {
+                yield item.map_err(JournalingError::Inner);
+            }
+        })
```
afck: Why not just `stream.map(|item| item.map_err(JournalingError::Inner))`?
(Also below.)

linera-views/src/backends/lru_caching.rs
```diff
@@ -317,6 +321,38 @@ where
...
+                    return;
+                }
+                #[cfg(with_metrics)]
+                metrics::FIND_KEYS_BY_PREFIX_CACHE_MISS_COUNT
+                    .with_label_values(&[])
+                    .inc();
+            }
+            let mut stream = self.store.find_keys_by_prefix_iter(key_prefix);
+            while let Some(item) = stream.next().await {
+                yield item;
```
afck: Why not cache them?

linera-views/src/backends/rocks_db.rs
```diff
@@ -198,6 +205,32 @@ impl RocksDbStoreExecutor {
         iter
     }
 
+    /// Returns an iterator positioned at the largest key starting with `prefix`,
+    /// iterating in reverse. `total_order_seek` is enabled because the database
+    /// uses a fixed-length prefix extractor; without it, `seek_for_prev` would
+    /// only search within the bloom-prefix scope and could miss keys whose
+    /// extractor-prefix differs from the seek target.
+    fn get_find_prefix_reverse_iterator(
```
afck: (Elsewhere we use `rev_iter` rather than `reverse_iterator`.)

linera-views/src/backends/rocks_db.rs
```diff
@@ -539,6 +572,126 @@ impl ReadableKeyValueStore for RocksDbStoreInternal {
...
+        key_prefix: &[u8],
+        direction: IterDirection,
+    ) -> FindKeysStream<'_, RocksDbStoreInternalError> {
+        let executor = self.executor.clone();
+        let key_prefix = key_prefix.to_vec();
+        Box::pin(async_stream::stream! {
+            if let Err(error) = check_key_size(&key_prefix) {
+                yield Err(error);
+                return;
+            }
```
afck: Does this work, too?

```suggestion
        Box::pin(async_stream::try_stream! {
            check_key_size(&key_prefix)?;
```

linera-views/src/backends/scylla_db.rs
```diff
@@ -180,6 +191,34 @@ impl ScyllaDbClient {
...
+            .prepare(format!(
+                "SELECT k FROM {KEYSPACE}.\"{namespace}\" \
+                 WHERE root_key = ? AND k >= ? ORDER BY k DESC"
+            ))
+            .await?;
+
+        let find_keys_by_prefix_rev_bounded = session
+            .prepare(format!(
+                "SELECT k FROM {KEYSPACE}.\"{namespace}\" \
+                 WHERE root_key = ? AND k >= ? AND k < ? ORDER BY k DESC"
```
afck, line 204: Shouldn't these be ascending instead of descending?

linera-views/src/backends/value_splitting.rs
```diff
@@ -252,6 +269,110 @@ where
...
+                let key = big_key;
+                let continues = match &state {
+                    Some((current_key, top, segs)) => {
+                        let segs_len = segs.len() as u32;
+                        *current_key == key && segs_len <= *top && index == *top - segs_len
+                    }
+                    None => false,
+                };
+                if continues {
+                    state.as_mut().unwrap().2.push(value);
```
afck: Maybe this `unwrap` can be avoided if it's done inside the `match` arm.

linera-views/src/backends/value_splitting.rs
```diff
@@ -252,6 +269,110 @@ where
...
+                    }
+                    let mut big_value = Vec::new();
+                    for (rev_pos, val) in segs.iter().rev().enumerate() {
+                        let idx = rev_pos as u32;
+                        if idx == 0 {
+                            big_value.extend_from_slice(&val[4..]);
+                        } else if idx < count {
+                            big_value.extend_from_slice(val);
+                        }
+                    }
```
afck:
```suggestion
                    let big_value = vec![segment_zero_value[4..].to_vec()];
                    for val in segs[1..count].iter().rev() {
                        big_value.extend_from_slice(val);
                    }
```

linera-views/src/store.rs
```diff
@@ -76,12 +95,83 @@ pub trait ReadableKeyValueStore: WithError {
...
 
+    /// Returns a boxed stream that yields the keys matching the prefix one by one.
+    /// The prefix is not included in the returned keys. The stream may be dropped
+    /// early to stop the iteration. The default implementation downloads the full
+    /// list eagerly via `find_keys_by_prefix` and yields the entries afterwards;
+    /// backends should override this to stream incrementally whenever possible.
+    fn find_keys_by_prefix_iter<'a>(
+        &'a self,
+        key_prefix: &'a [u8],
+    ) -> FindKeysStream<'a, Self::Error> {
```
afck, line 106: I think @Twey had the idea to make the return type `impl Stream<Item = Result<Vec<u8>>>`; would that make it `Send` or `!Send` automatically as needed?

linera-views/src/store.rs
```diff
@@ -76,12 +95,83 @@ pub trait ReadableKeyValueStore: WithError {
...
+        &'a self,
+        key_prefix: &'a [u8],
+    ) -> FindKeyValuesStream<'a, Self::Error> {
+        Box::pin(async_stream::stream! {
+            let key_values = self.find_key_values_by_prefix(key_prefix).await?;
+            for key_value in key_values.into_iter().rev() {
+                yield Ok(key_value);
+            }
+        })
+    }
```
afck, line 169: Could these all use `map` instead of `async_stream::stream!`?

linera-views/src/backends/scylla_db.rs
MathieuDutSik: The lexicographic order is the ascending one.
So, I think this is correct.

afck: (Ah, sorry, I missed that all four of those are `DESC`!)

linera-views/src/store.rs
MathieuDutSik: With the design that I propose (and exercised first on the `read_multi_values_bytes_iter`) it compiles on `Wasm32`.
So, I am not sure what problem is left to resolve.

afck: Just that it would be simpler (if it works!) and use less conditional compilation.

linera-views/src/backends/lru_caching.rs
MathieuDutSik: The iterator can be prematurely ended (for example when doing a `find_first_key_in_prefix`). So, it may not be possible to cache it. But let me look at that, maybe we can do it.

linera-views/src/backends/rocks_db.rs
MathieuDutSik: So, what would you prefer?
For myself, `rev_iter`.

linera-views/src/backends/journaling.rs
MathieuDutSik: That works.

linera-views/src/backends/rocks_db.rs
afck: I think I'd prefer the short form, too, yes. :+1:

linera-views/src/backends/lru_caching.rs
MathieuDutSik: So, yes we can.

linera-views/src/store.rs
MathieuDutSik: For the trait implementation, we can improve but that is not so simple since for a start the `_iter` function is not async.

linera-views/src/backends/rocks_db.rs
MathieuDutSik: Ok, done. Use `try_stream` when possible.

linera-views/src/backends/value_splitting.rs
MathieuDutSik: Ok, changed.

linera-views/src/test_utils/mod.rs
MathieuDutSik: Ok, corrected.

linera-views/src/backends/lru_caching.rs
```diff
@@ -325,31 +325,43 @@ where
...
-                metrics::FIND_KEYS_BY_PREFIX_CACHE_MISS_COUNT
-                    .with_label_values(&[])
-                    .inc();
-            }
-            let mut stream = self.store.find_keys_by_prefix_iter(key_prefix);
-            while let Some(item) = stream.next().await {
-                yield item;
             }
+            let mut cache = cache.lock().unwrap();
+            cache.insert_find_keys(key_prefix.to_vec(), &accumulated);
```
afck, line 357: OK, maybe that was a bad idea: A lot of time could have passed since we loaded all those values; couldn't they have changed in the meantime?

linera-views/src/backends/value_splitting.rs
MathieuDutSik: Yes, done.

linera-views/src/backends/lru_caching.rs
MathieuDutSik: The caching of `find_key_values_by_prefix` is only done when we have exclusive access. So, not with blobs, events, and similar.

So, could the state have changed meanwhile? The answer is yes: The construction of the iterator does not create a lock, in general. It certainly does not for `DynamoDB` and `ScyllaDB`. But if some keys are inserted or deleted during the operations then your iterator will be unstable. It cannot guarantee that it works correctly.

So, my verdict: Yes, it is fine to save the keys because yes, if that has changed, then your code would anyway not have worked correctly.

But this makes me realize another concern. Could it be that establishing an iterator prevents other operations from working? In other words, is the following code valid:
```rust
let mut iter = store.find_keys_by_prefix_iter(&key)?;
let mut values = Vec::new();
while let Some(key) = iter.next().await? {
  let value = store.read_value(&key).await?;
  values.push(value);
}
```

And the answer is that it is a problem for ScyllaDB. So, do not acquire a semaphore for the creation of an iterator so as not to introduce a deadlock. Correction done. I also added a test for that deadlocking in `run_reads`.

bart-linera: Maybe if the iterator can be unstable when the underlying state changes while it iterates, it should actually prevent such changes? But I guess that depends on what the use cases are.
Would it make the most sense if only operations on the keys that match the iterator's prefix were blocked? But perhaps it is not easy to achieve.

linera-views/src/backends/value_splitting.rs
```diff
@@ -252,6 +269,114 @@ where
...
+                    // Either no prior state, or the prior state belongs to a
+                    // different key / a non-contiguous index — drop it (those
+                    // segments are leftovers from a deleted big value) and
+                    // start a fresh accumulation.
+                    _ => Some((key, index, vec![value])),
+                };
+                if index == 0 {
+                    let (key, top, segs) = state.take().unwrap();
+                    let count = Self::read_count_from_value(segs.last().unwrap())?;
```
ndr-ds, line 359: Do we want these `unwrap`s? I see a few

PR #6205, opened 2026-05-02
Remove some `unreachable` and `.unwrap()`.

MathieuDutSik:
## Motivation

There are some `unreachable` / `.unwrap()` statements that we can remove without changing
any of the behavior of the code.

## Proposal

The corrections I found only in `linera-views` and `linera-execution`.

## Test Plan

CI

## Release Plan

## Links

None

### DIFF
--- a/linera-views/src/backends/dynamo_db.rs
+++ b/linera-views/src/backends/dynamo_db.rs
@@ -1067,7 +1067,6 @@ impl DynamoDbStoreInternalError {
 
     fn type_description_of(value: &AttributeValue) -> String {
         match value {
-            AttributeValue::B(_) => unreachable!("creating an error type for the correct type"),
             AttributeValue::Bool(_) => "a boolean",
             AttributeValue::Bs(_) => "a list of binary blobs",
             AttributeValue::L(_) => "a list",
--- a/linera-views/src/views/bucket_queue_view.rs
+++ b/linera-views/src/views/bucket_queue_view.rs
@@ -603,22 +603,23 @@ impl<C: Context, T: DeserializeOwned + Clone, const N: usize> BucketQueueView<C,
         let Some(bucket) = self.stored_buckets.back() else {
             return Ok(None);
         };
-        if !bucket.is_loaded() {
-            let key = self.get_bucket_key(bucket.index)?;
-            let data = self.context.store().read_value(&key).await?;
-            let data = match data {
-                Some(data) => data,
-                None => {
-                    return Err(ViewError::MissingEntries("BucketQueueView::back".into()));
-                }
-            };
-            self.stored_buckets.back_mut().unwrap().state = State::Loaded { data };
+        match &bucket.state {
+            State::Loaded { data } => Ok(Some(data.last().unwrap().clone())),
+            State::NotLoaded { .. } => {
+                let key = self.get_bucket_key(bucket.index)?;
+                let data: Vec<T> = self
+                    .context
+                    .store()
+                    .read_value(&key)
+                    .await?
+                    .ok_or_else(|| {
+                        ViewError::MissingEntries("BucketQueueView::back".into())
+                    })?;
+                let result = data.last().unwrap().clone();
+                self.stored_buckets.back_mut().unwrap().state = State::Loaded { data };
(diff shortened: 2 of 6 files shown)


linera-views/src/backends/dynamo_db.rs
```diff
@@ -1067,7 +1067,6 @@ impl DynamoDbStoreInternalError {
 
     fn type_description_of(value: &AttributeValue) -> String {
         match value {
-            AttributeValue::B(_) => unreachable!("creating an error type for the correct type"),
```
afck, line 1070: (Why was that even here? 🤨)

linera-views/src/views/bucket_queue_view.rs
```diff
@@ -603,22 +603,21 @@ impl<C: Context, T: DeserializeOwned + Clone, const N: usize> BucketQueueView<C,
...
-                None => {
-                    return Err(ViewError::MissingEntries("BucketQueueView::back".into()));
-                }
-            };
-            self.stored_buckets.back_mut().unwrap().state = State::Loaded { data };
+        match &bucket.state {
+            State::Loaded { data } => Ok(Some(data.last().unwrap().clone())),
+            State::NotLoaded { .. } => {
+                let key = self.get_bucket_key(bucket.index)?;
+                let data: Vec<T> = self
```
afck: Type annotation

linera-views/src/views/bucket_queue_view.rs
```diff
@@ -603,22 +603,21 @@ impl<C: Context, T: DeserializeOwned + Clone, const N: usize> BucketQueueView<C,
...
+                    .context
+                    .store()
+                    .read_value(&key)
+                    .await?
+                    .ok_or_else(|| ViewError::MissingEntries("BucketQueueView::back".into()))?;
+                let result = data.last().unwrap().clone();
```
afck: This also shouldn't be an `unwrap`: It could be triggered by a faulty database.

linera-views/src/views/register_view.rs
```diff
@@ -221,10 +221,7 @@ where
         self.delete_storage_first = false;
         match &mut self.update {
             Some(value) => value,
-            update => {
-                *update = Some(self.stored_value.clone());
-                update.as_mut().unwrap()
-            }
+            update => update.insert(self.stored_value.clone()),
         }
```
afck: Or just `self.update.get_or_insert_with(|| self.stored_value.clone())`?

linera-views/src/backends/dynamo_db.rs
MathieuDutSik: It was.

linera-views/src/views/bucket_queue_view.rs
MathieuDutSik: Actually, that one is needed.

linera-views/src/views/bucket_queue_view.rs
MathieuDutSik: I am not sure about that:
* The `BucketQueueView` has a number of invariants that are expressed by having some `.unwrap()` statements in the code.
* It is, of course, possible that any of those invariants are broken by having a faulty database. But I think that this is a dangerous path to take. Most likely we have an error.

In this PR, I did not want to change the behavior, that is replacing unreachable/unwrap/expect by an explicit error.

What I can propose is instead to replace the `.unwrap()` by explicit statement of the broken invariants.

linera-views/src/views/register_view.rs
MathieuDutSik: Yes, it works.

linera-views/src/views/bucket_queue_view.rs
afck: Not a blocker from my side, but I think our policy would be that it's below instead, something like `read_value::<Result<Vec<_>, _>>`

afck (approved):

linera-views/src/views/bucket_queue_view.rs
MathieuDutSik: Right, I had missed that.

PR #6206, opened 2026-05-02
[testnet] Remove some `unreachable` and `.unwrap()`.

MathieuDutSik:
## Motivation

There are some unreachable / .unwrap() statements that we can remove without changing
any of the behavior of the code.

Backport https://github.com/linera-io/linera-protocol/pull/6205

## Proposal

The corrections I found only in `linera-views` and `linera-execution`.

## Test Plan

CI

## Release Plan

This is the backport.

## Links

None.

### DIFF
--- a/linera-views/src/views/register_view.rs
+++ b/linera-views/src/views/register_view.rs
@@ -221,10 +221,7 @@ where
         self.delete_storage_first = false;
         match &mut self.update {
             Some(value) => value,
-            update => {
-                *update = Some(self.stored_value.clone());
-                update.as_mut().unwrap()
-            }
+            update => update.insert(self.stored_value.clone()),
         }
     }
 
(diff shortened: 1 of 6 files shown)


linera-views/src/views/register_view.rs
```diff
@@ -221,10 +221,7 @@ where
         self.delete_storage_first = false;
         match &mut self.update {
             Some(value) => value,
-            update => {
-                *update = Some(self.stored_value.clone());
-                update.as_mut().unwrap()
-            }
+            update => update.insert(self.stored_value.clone()),
```
afck: Didn't we use `get_or_insert` here? Would be good to agree with `main` as much as possible.

MathieuDutSik: Yes, sorry for that.

afck (approved):

PR #6209, opened 2026-05-02
Re-enable anvil_deposit_proof in CI with stabilization fixes.

MathieuDutSik:
## Motivation

PR #6184 enabled this test in CI and was reverted in #6197 because the runs failed consistently with `code: 3, data: 0x` (empty-data revert).

## Proposal

This re-enables it with three fixes that together address the failure.

* Pre-fill `gas` and `chain_id` on every `TransactionRequest` and cache the chain id once after spawning Anvil. This skips the per-tx `eth_estimateGas` round-trip, which on a contended runner was racing against `latest` and surfacing as `extcodesize == 0` reverts (empty data) on typed external calls.

* Pin the solc target to `evmVersion: cancun` and switch Anvil from `--hardfork shanghai` to `--hardfork cancun`. Cancun matches the EVM Base currently runs, and pinning makes the bytecode independent of whatever default future solc releases ship with. The obsolete comment about Anvil v0.2.0 missing Cancun header fields is removed. We set it up to `cancun` because this is the version used by `base`.

* Pin `foundry-toolchain@v1` to `version: v1.7.0` instead of drifting with `stable`.

Also drops the orphan `registerFungibleApplicationId` ABI binding and call left over after #6173 made `fungibleApplicationId` an immutable constructor argument and removed the setter.

## Test Plan

CI is the point.

## Release Plan

Can be backported to `testnet_conway` when we are confident that it is correct.

## Links

None

### DIFF
--- a/linera-bridge/tests/anvil_deposit_proof.rs
+++ b/linera-bridge/tests/anvil_deposit_proof.rs
@@ -73,8 +73,6 @@ contract MockERC20 {
 alloy_sol_types::sol! {
     function approve(address spender, uint256 amount) external returns (bool);
 
-    function registerFungibleApplicationId(bytes32 _fungibleApplicationId) external;
-
     function deposit(
         bytes32 target_chain_id,
         bytes32 target_application_id,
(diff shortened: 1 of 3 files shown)


linera-bridge/tests/anvil_deposit_proof.rs
```diff
@@ -100,13 +98,19 @@ fn compile_contract(source_code: &str, file_name: &str, contract_name: &str) ->
...
+
 #[tokio::test]
 #[ignore] // Requires `anvil` and `solc`
 async fn test_deposit_proof_generation() -> Result<(), Box<dyn std::error::Error>> {
-    // 1. Spawn Anvil
-    // Use Shanghai hardfork to avoid header field mismatches with older Anvil versions
-    // (Anvil v0.2.0 doesn't return all Cancun header fields via JSON-RPC).
-    let anvil = Anvil::new().arg("--hardfork").arg("shanghai").try_spawn()?;
+    // 1. Spawn Anvil targeting the Cancun EVM (Base's current hardfork).
+    let anvil = Anvil::new().arg("--hardfork").arg("cancun").try_spawn()?;
```
deuszx, line 113: Can you also update the `docker-compose.bridge-test.yml` ? It also uses `shanghai` at the moment.

deuszx: Ping here @MathieuDutSik

deuszx (approved):

PR #6210, opened 2026-05-02
Use the `Epoch` type for linera epochs.

MathieuDutSik:
## Motivation

Linera epochs have their own types, "Epoch".
We can use them in `linera-bridge`.

## Proposal

Do the change

## Test Plan

CI

## Release Plan

## Links

None

### DIFF
--- a/linera-bridge/src/relay/committee.rs
+++ b/linera-bridge/src/relay/committee.rs
@@ -55,9 +55,8 @@ pub async fn relay_committee<P: Provider>(
 }
 
 /// Catches up the LightClient with any committee updates missed while offline.
-/// Scans admin chain blocks from height 0 and relays committees newer than the
-/// LightClient's current epoch. Must succeed before the relay enters the main
-/// serve loop.
+/// Scans admin chain blocks and relays committees newer than the LightClient's
+/// current epoch. Must succeed before the relay enters the main serve loop.
 pub async fn catch_up<S, P>(
     storage: &S,
     evm_client: &EvmClient<P>,
@@ -69,13 +68,12 @@ where
     P: Provider,
 {
     let current_epoch = match evm_client.get_current_epoch().await {
-        Ok(epoch) => Epoch(epoch),
+        Ok(epoch) => epoch,
         Err(e) => {
             tracing::info!("LightClient not initialized yet, skipping catch-up: {e:#}");
             return Ok(());
         }
     };
-
     tracing::info!(
         %current_epoch,
         %admin_chain_id,
@@ -83,12 +81,12 @@ where
         "Checking for missed committee updates"
     );
 
-    if admin_chain_height == BlockHeight(0) {
+    let heights: Vec<BlockHeight> = (0..admin_chain_height.0).map(BlockHeight).collect();
+
+    if heights.is_empty() {
         tracing::info!("No admin chain blocks to scan");
         return Ok(());
     }
-
-    let heights: Vec<BlockHeight> = (0..admin_chain_height.0).map(BlockHeight).collect();
 
     let certs = storage
         .read_certificates_by_heights(admin_chain_id, &heights)
--- a/linera-bridge/src/relay/committee.rs
+++ b/linera-bridge/src/relay/committee.rs
@@ -55,8 +55,9 @@ pub async fn relay_committee<P: Provider>(
 }
 
 /// Catches up the LightClient with any committee updates missed while offline.
-/// Scans admin chain blocks and relays committees newer than the LightClient's
-/// current epoch. Must succeed before the relay enters the main serve loop.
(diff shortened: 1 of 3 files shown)


linera-bridge/src/monitor/db.rs
```diff
@@ -104,9 +104,65 @@ impl BridgeDb {
             .execute(&self.pool)
             .await?;
 
+        sqlx::query(
+            "CREATE TABLE IF NOT EXISTS scan_state (
+                key   TEXT PRIMARY KEY,
+                value INTEGER NOT NULL
```
deuszx: This seems too ambiguous – what are the `key` and `value`? Let's be more specific. I can see that you want to be "flexible" but I would argue YAGNI.

linera-bridge/src/monitor/db.rs
```diff
@@ -104,9 +104,65 @@ impl BridgeDb {
...
+    /// LightClient `currentEpoch` observed at the time `last_scanned_admin_height` was written.
+    /// Used to detect EVM rollbacks and force a full rescan.
+    pub async fn get_last_known_evm_epoch(&self) -> Result<Option<u32>> {
+        Ok(self
+            .get_scan_state("last_known_evm_epoch")
+            .await?
+            .map(|v| v as u32))
+    }
+
+    pub async fn set_last_known_evm_epoch(&self, epoch: u32) -> Result<()> {
```
deuszx: Why not pass in the `epoch: Epoch`?

deuszx: Ah, it's on the EVM!.

linera-bridge/src/relay/committee.rs
```diff
@@ -74,20 +80,42 @@ where
...
+            tracing::warn!(
+                current_epoch,
+                prev_epoch,
+                last_scanned = %h,
+                "LightClient epoch regressed since last scan; rescanning admin chain from 0"
+            );
+            BlockHeight(0)
+        }
+        _ => BlockHeight(0),
+    };
```
deuszx: I don't understand this logic. Why are we checking EVM epochs and then somehow using that in the result of Linera comparison?

linera-bridge/src/monitor/db.rs
deuszx: We still could use `Epoch` type here.

linera-bridge/src/relay/committee.rs
```diff
@@ -74,20 +80,42 @@ where
             return Ok(());
         }
     };
+
+    let persisted_height = db.get_last_scanned_admin_height().await?;
+    let persisted_epoch = db.get_last_known_evm_epoch().await?;
+    let start_height = match (persisted_height, persisted_epoch) {
+        (Some(h), Some(prev_epoch)) if current_epoch >= prev_epoch => h,
```
deuszx: we should update what's persisted then

linera-bridge/src/monitor/db.rs
MathieuDutSik: Correct.
The epoch is a linera epoch, so it makes sense to use the `Epoch` type.

linera-bridge/src/relay/committee.rs
MathieuDutSik: The "EVM epoch" are actually Epoch from linera as implemented in EVM.
I changed the structure to the following:
* If the current_epoch is decreasing, then we have an error, and we report so.
* If any of the epoch or block height has failed, then there is an error and we need to restart from Block height 0.

linera-bridge/src/relay/committee.rs
MathieuDutSik: I am not sure what you mean. The `catch_up` function is persisting the height and epoch before exiting.

linera-bridge/src/relay/committee.rs
deuszx: I think persisting this in the local DB is unnecessary – we now have 3 sources of this data (this complicates the logic as we can see here) but we are really interested in two:
1. Admin chain as the reference (source of truth)
2. EVM stored (what we need to update)

A simper variant is IMO the following:
On startup, do the "forced" sync of the admin chain using local Linera client. If our client is empty, it will start from scratch (but it shouldn't be empty, i.e. have no info about the admin chain) – we need to do this anyway. Then, send all committee updates to the LightClient instance as we sync (or filter out epochs that the EVM already knows about).

deuszx: This comment still stands @MathieuDutSik . It's not optional.

MathieuDutSik: Done. The change is now reverted. But yes, it is better to have only one source of data.
The changes now reduce to using the type "Epoch" for the linera epochs.

deuszx (approved):

PR #6219, opened 2026-05-05
Separate tables for pending and finished relay tasks

deuszx:
## Motivation

Tracking both pending and finished tasks in the same table means the queries will become slower with time.

## Proposal

Split into separate tables.

## Test Plan

CI

## Release Plan

## Links

https://github.com/linera-io/linera-protocol/pull/6185#pullrequestreview-4218334836

### DIFF
--- a/linera-bridge/src/monitor/db.rs
+++ b/linera-bridge/src/monitor/db.rs
@@ -394,14 +394,14 @@ mod tests {
 
     static FILE_COUNTER: AtomicU32 = AtomicU32::new(0);
 
-    async fn open_db(use_file: bool) -> BridgeDb {
+    async fn open_db(use_file: bool) -> Result<BridgeDb> {
         if use_file {
             let n = FILE_COUNTER.fetch_add(1, Ordering::Relaxed);
             let path = std::path::PathBuf::from(format!("/tmp/bridge_db_test_{n}.sqlite3"));
             std::fs::remove_file(&path).ok(); // Best-effort cleanup; NotFound is the common case.
-            BridgeDb::open(&path).await.unwrap()
+            BridgeDb::open(&path).await
         } else {
-            BridgeDb::open_in_memory().await.unwrap()
+            BridgeDb::open_in_memory().await
         }
     }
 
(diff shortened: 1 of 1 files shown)


linera-bridge/src/monitor/db.rs
```diff
@@ -345,29 +426,124 @@ mod tests {
...
-    async fn test_complete_deposit(use_file: bool) {
+    async fn test_insert_deposit_idempotent(use_file: bool) {
+        let db = open_db(use_file).await;
+        db.insert_deposit(&test_deposit()).await.unwrap();
+        db.insert_deposit(&test_deposit()).await.unwrap();
+
+        let (count,): (i64,) = sqlx::query_as("SELECT COUNT(*) FROM pending_deposits")
+            .fetch_one(&db.pool)
+            .await
+            .unwrap();
```
MathieuDutSik: I would like this function to be part of `BridgeDb`.
Because the function could be of general interest.

linera-bridge/src/monitor/db.rs
```diff
@@ -345,29 +426,124 @@ mod tests {
...
+        .fetch_one(&db.pool)
+        .await
+        .unwrap();
+        assert_eq!(raw, raw_bytes);
+    }
+
+    #[test_case(false; "in_memory")]
+    #[test_case(true; "file_backed")]
+    #[tokio::test]
+    async fn test_insert_and_query_burn(use_file: bool) {
```
MathieuDutSik: I prefer the test functions `anyhow::Result<()>`. Using `?` is preferable to `.unwrap()`.

MathieuDutSik (commented):
I would like to see a test where we start with say 5 pending, we are processing 2, and then we are left with 3 pending ones and 2 finished.

linera-bridge/src/monitor/db.rs
deuszx: Done in [Add DB methods returning count of pending burns and deposits](https://github.com/linera-io/linera-protocol/pull/6219/commits/60390852baed2e23b34fdcd518bdaa6e5dccb20c)

deuszx:
> I would like to see a test where we start with say 5 pending, we are processing 2, and then we are left with 3 pending ones and 2 finished.

Done in [Add a test with partial processing finish](https://github.com/linera-io/linera-protocol/pull/6219/commits/e45ef510af8033f21bf6431b63b79af82971ae69)

linera-bridge/src/monitor/db.rs
deuszx: Done in [monitor/db.rs tests return Result instead of unwrap](https://github.com/linera-io/linera-protocol/pull/6219/commits/f4dfb3f5867d80cea0a2ea73954d430d79606a87)

PR #6220, opened 2026-05-05
`HistoricallyHashableView::set_stored_hash`

afck:
## Motivation

In the future we want checkpoint chain states, and restore them from checkpoints.

The execution state hash is the hash of the _change history_, not the contents of the execution state itself. We want to keep that, so checkpoints should not affect hashing.

Thus when starting a chain from a checkpoint, we will need to explicitly initialize the historically hashable view.

## Proposal

Let a caller override the next save's stored hash without mixing the inner view's pending batch into it. After save, the view resumes from that hash and subsequent updates extend it normally.

## Test Plan

Tests were added.

## Release Plan

## Links

### DIFF
--- a/linera-views/src/views/historical_hash_wrapper.rs
+++ b/linera-views/src/views/historical_hash_wrapper.rs
@@ -788,36 +788,4 @@ mod tests {
         Ok(())
     }
 
-    #[tokio::test]
-    async fn test_set_stored_hash_overwrites_existing_chain() -> Result<(), ViewError> {
-        // Verify that once a chain has accumulated history, set_stored_hash truly replaces
-        // it (rather than mixing in).
-        let context = MemoryContext::new_for_testing(());
-        let mut view =
-            HistoricallyHashableView::<_, RegisterView<_, u32>>::load(context.clone()).await?;
-
-        view.set(7);
-        let mut batch = Batch::new();
-        view.pre_save(&mut batch)?;
-        context.store().write_batch(batch).await?;
-        view.post_save();
-
-        let chained = view.historical_hash().await?;
-        let forced = HasherOutput::from([4u8; 32]);
-        assert_ne!(chained, forced);
-
-        view.set_stored_hash(forced);
-        let mut batch = Batch::new();
-        view.pre_save(&mut batch)?;
-        context.store().write_batch(batch).await?;
-        view.post_save();
-
-        let mut reloaded =
-            HistoricallyHashableView::<_, RegisterView<_, u32>>::load(context.clone()).await?;
-        assert_eq!(reloaded.historical_hash().await?, forced);
-        // Inner state is unaffected.
-        assert_eq!(*reloaded.get(), 7);
-
-        Ok(())
-    }
 }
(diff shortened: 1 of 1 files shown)


MathieuDutSik (approved):
Yes, that makes sense if the checkpoint forces us to set up the hash. That is logical. And also kind of terrifying.

But my question is: Do the use of checkpoint automatically means that we have to believe by default because there is no other way than to replay the whole set of blocks? I would like to know.

ma2bd:
I'm not entirely fond of this. Could restoring a checkpoint be seen as a write operation?

afck:
> Do the use of checkpoint automatically means that we have to believe by default because there is no other way than to replay the whole set of blocks? I would like to know.

No other _fast_ way, yes: If we have a very long chain we want to avoid having to replay it entirely in every new node (client or validator).

> Could restoring a checkpoint be seen as a write operation?

I'm not sure I understand: It does write to storage, so of course it's a write operation?

afck:
I changed `set_stored_hash` to `reset_with_stored_hash`, which calls `inner.clear()`: That way it's enforced that the view has to be written from scratch.

Apart from that I don't have any better ideas. The only alternative I see is to hash the _content_ instead, but then all other nodes have to explicitly reset the hash of the _existing_ view so it matches the content instead of the history now. I think that's worse?

bart-linera:
Maybe the execution of the `Checkpoint` operation could reset the hash to the hash of the content? So _creating_ a checkpoint would replace the hash, and then continue historical hashing from there; and restoring a checkpoint would restore the state _and_ set the hash to the hash of the content, and continue from there. It would be another way of keeping things consistent, and the hash could be used to verify the execution state as well.

afck:
Yes, that's what I meant in my second paragraph above. But I'd prefer not to:
* It means that if you already have the chain, you still have to reset your hash at each checkpoint, even if you're not using the checkpoint.
* We already _do_ have hashes of the content anyway: The block contains the hashes of the blobs containing the execution state.

ma2bd:
> Yes, that's what I meant in my second paragraph above. But I'd prefer not to:
> 
> * It means that if you already have the chain, you still have to reset your hash at each checkpoint, even if you're not using the checkpoint.
> 
> * We already _do_ have hashes of the content anyway: The block contains the hashes of the blobs containing the execution state.

Yes I was suggesting to hash the content. The same operation should restore the state from the blob bytes. It should be fairly cheap to hash the content while we're restoring from the blob.

There is even a world in which the hash of the content and the hash of the blob bytes are the same thing.

afck:
@ma2bd: So you'd prefer that nodes that don't _use_ the checkpoint reset their execution state hash to the content-based one when they execute that block?
But then we'd _have_ to revert https://github.com/linera-io/linera-protocol/pull/6220/commits/ce62a5760a0dac1ce1b54bd6b2fc0478afa1aa7d, right? Those nodes would then actually overwrite the hash in an existing execution state hash without overwriting the content.

> the hash of the blob bytes are the same thing

That's difficult in general, due to the blob size limits.

afck:
As discussed, I will attempt the alternative approach on a new PR.

afck:
Superseded by https://github.com/linera-io/linera-protocol/pull/6242.

PR #6234, opened 2026-05-05
Tighten URI validation in extract_grpc_method_name.

ndr-ds:
## Motivation

The `linera-proxy` and validator gRPC server record the gRPC method name as a Prometheus label (`linera_proxy_request_*{method_name=...}` and the symmetric server-side metrics). The previous `extract_grpc_method_name` only required `parts[1]` to contain a `.`, so any path with a dotted segment passed through. Internet bot scans probing for `.env` files (path like `/.env.local/.env.production/.env.staging/...`) made it through that check, and the entire tail of the path landed in `parts[2]` and was used verbatim as the label value.

Mimir rejects values longer than its `max_label_value_length` (default 2048) with `err-mimir-label-value-too-long` (HTTP 400), which trips `prometheus_remote_storage_samples_failed_total` and the `AlloyRemoteWriteFailures` alert. Even without the length cap, attacker-shaped paths are a cardinality bomb in `method_name`.

## Proposal

Replace the `parts[1].contains(.)` heuristic with a strict gRPC URI parse:
- Path must be `"/" Service-Name "/" {method name}`.
- Service-Name must be one or more proto identifiers joined by dots, with at least one dot (so `package.Service` qualifies, but `.env.local` does not — a leading dot makes the first split component empty and fails the identifier check).
- Method name must be a single proto identifier.
- Both length-capped at 128 bytes as belt-and-braces.

Anything outside this shape returns `"non_grpc"`, the existing fallback. External contract, return type, and callers are unchanged.

Side effect: per-call `Vec<&str>` allocation removed (zero-alloc destructured `splitn` iterator).

## Test Plan

CI. Added 9 unit tests covering the bot-scan input, leading-dot service, extra path segments, invalid characters, empty segments, consecutive dots, an overlong method, and a missing leading slash. Existing positive tests stay green.

## Release Plan

## Links

### DIFF
--- a/linera-rpc/src/grpc/mod.rs
+++ b/linera-rpc/src/grpc/mod.rs
@@ -56,7 +56,7 @@ pub const ERROR_TYPE_LABEL: &str = "error_type";
 /// input being recorded verbatim as a Prometheus label value.
 const MAX_PROTO_IDENT_LEN: usize = 128;
 
-/// Returns `true` if `s` is a valid protobuf identifier (`[A-Za-z_][A-Za-z0-9_]*`)
+/// Returns `true` if `s` is a valid protobuf identifier (`[A-Za-z][A-Za-z0-9_]*`)
 /// within the length cap.
 fn is_proto_identifier(s: &str) -> bool {
     if s.is_empty() || s.len() > MAX_PROTO_IDENT_LEN {
@@ -64,10 +64,7 @@ fn is_proto_identifier(s: &str) -> bool {
     }
     let mut bytes = s.bytes();
     let first = bytes.next().expect("non-empty checked above");
-    if !first.is_ascii_alphabetic() && first != b'_' {
-        return false;
-    }
-    bytes.all(|byte| byte.is_ascii_alphanumeric() || byte == b'_')
+    first.is_ascii_alphabetic() && bytes.all(|b| b.is_ascii_alphanumeric() || b == b'_')
 }
 
 /// Returns `true` if `s` is a fully-qualified proto Service-Name: two or more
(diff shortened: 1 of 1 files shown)


linera-rpc/src/grpc/mod.rs
```diff
@@ -51,15 +51,47 @@ pub const TRAFFIC_TYPE_LABEL: &str = "traffic_type";
...
+    let first = bytes.next().expect("non-empty checked above");
+    if !first.is_ascii_alphabetic() && first != b'_' {
+        return false;
+    }
+    bytes.all(|byte| byte.is_ascii_alphanumeric() || byte == b'_')
+}
+
+/// Returns `true` if `s` is a fully-qualified proto Service-Name, i.e. one or more
+/// proto identifiers joined by dots with at least one dot (`package.Service` or
+/// `outer.inner.Service`).
```
deuszx: I don't see this code checking if input string contains `package.Service` or `outer.inner.Service` – am I missing something?

ndr-ds: These are just examples of the type of thing we accept

deuszx: So the only thing we can assert here for is the existence of the dots?

ndr-ds: Ok, let me make it a bit better

linera-rpc/src/grpc/mod.rs
```diff
@@ -51,15 +51,53 @@ pub const TRAFFIC_TYPE_LABEL: &str = "traffic_type";
...
+
+/// Returns `true` if `s` is a valid protobuf identifier (`[A-Za-z_][A-Za-z0-9_]*`)
+/// within the length cap.
+fn is_proto_identifier(s: &str) -> bool {
+    if s.is_empty() || s.len() > MAX_PROTO_IDENT_LEN {
+        return false;
+    }
+    let mut bytes = s.bytes();
+    let first = bytes.next().expect("non-empty checked above");
+    if !first.is_ascii_alphabetic() && first != b'_' {
```
deuszx: so the first byte must be an underscore?

linera-rpc/src/grpc/mod.rs
```diff
@@ -51,15 +51,53 @@ pub const TRAFFIC_TYPE_LABEL: &str = "traffic_type";
...
+        return false;
+    }
+    bytes.all(|byte| byte.is_ascii_alphanumeric() || byte == b'_')
```
deuszx: so is this a correct proto identifier: `________________` ?

ndr-ds: Good point

deuszx (approved):

PR #6236, opened 2026-05-06
Remove the per-chain committee map.

afck:
## Motivation

Revoking an epoch means the committee is not trusted anymore. This should not be delayed by each chain having to acknowledge the revocation.

## Proposal

Remove the per-chain committee map; revocation takes effect immediately.

## Test Plan

CI; tests have been updated where necessary.

## Release Plan

## Links

- First step of https://github.com/linera-io/linera-protocol/issues/6237.


linera-core/src/chain_worker/state.rs
```diff
@@ -187,6 +187,72 @@ where
         &self.chain
     }
 
+    /// Filters bundles destined for this chain to drop ones already received and to refuse
+    /// ones whose epoch has been revoked on the admin chain.
```
deuszx, line 191: So it's possible now, for an unlucky senderm, to send a message in epoch `n`, then admin chain is migrated and recipient will reject the incoming message b/c we're at epoch `n+1`?

afck: Yes: That's why we should never create epoch `n + 1` and immediately revoke epoch `n`. In a multi-chain system it's necessary to leave enough time for all chains to migrate and process their messages before revoking an epoch.

ma2bd (approved):

PR #6242, opened 2026-05-06
Add checkpointing functions to historically hashable views

afck:
## Motivation

In the future we want checkpoint chain states, and restore them from checkpoints.

That means the "historical" hash of the execution state view will not be the hash of the written batches since block 0, but the hash of the written batches since the most recent checkpoint and the checkpoint itself.

The historically hashable views currently don't support that.

## Proposal

Give historically hashable views the ability to:
* reset their contents to a value given as the serialized view, and
* reset the hash to the hash of those serialized bytes.

## Test Plan

Tests were added.

## Release Plan

## Links

### DIFF
--- a/linera-views/src/views/historical_hash_wrapper.rs
+++ b/linera-views/src/views/historical_hash_wrapper.rs
@@ -365,8 +365,16 @@ fn decode_key_values(bytes: &[u8]) -> Result<Vec<(Vec<u8>, Vec<u8>)>, ViewError>
 }
 
 fn hash_bytes(bytes: &[u8]) -> HasherOutput {
-    use crate::views::Hasher as _;
+    // Domain-separation tag: ensures the SHA3 input here cannot equal the SHA3 input of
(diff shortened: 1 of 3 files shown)


linera-views/src/views/historical_hash_wrapper.rs
```diff
@@ -225,6 +247,130 @@ impl<W: View> HistoricallyHashableView<W::Context, W> {
...
+        if window[1].0 <= window[0].0 {
+            return Err(ViewError::MalformedContent(
+                "keys must be in strictly increasing order",
+            ));
+        }
+    }
+    Ok(entries)
+}
+
+fn hash_bytes(bytes: &[u8]) -> HasherOutput {
```
ma2bd, line 367: Let's make sure that the output of `hash_bytes` can never be made to collide with the output of `make_hash`.

afck: Done. Happy to change the domain tag if you prefer something else.

ma2bd (approved):

PR #6249, opened 2026-05-06
Backport to testnet_conway: Push Arc deeper through certificate and block paths (#6147)

ndr-ds:
## Motivation

Backport of #6147 to `testnet_conway`. Reduces unnecessary cloning of certificates and blocks by passing `Arc<ConfirmedBlockCertificate>` and `Arc<ConfirmedBlock>` through the call stack rather than owned values.

## Proposal

Cherry-pick of `3287f869cc0` onto `testnet_conway` with conflict resolution:

- Kept testnet_conway's `read_certificates_for_heights` fallback path (via `write_certificate_height_indices`), adapting it to return `Vec<Arc<...>>` using `cache_certificate` instead of cloning
- Kept testnet_conway's `if all_found` / fallback guard in `proxy/grpc.rs`, using `certificate.as_ref().into()` instead of `Arc::unwrap_or_clone`
- Kept testnet_conway's `download_certificates_for_events` structure (group-by-chain with `sync_events_from_node`) which uses a different approach than main
- Fixed a duplicate `update_validators` call introduced during conflict resolution, inserting `cache_certificate` before the single remaining call

## Test Plan

CI

## Release Plan

## Links

- #6147

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -70,6 +70,29 @@ Contributions should generally follow the [Rust API guidelines](https://rust-lan
   unawaited futures or unhandled errors. `let _x =` should only be used for RAII guards.
 
 
+## Hash-consed types and `Arc<T>`
+
+Any content-addressed immutable data (also known as hash-consed data) — for
+example `Block`, `Blob`, `ConfirmedBlockCertificate` — should be cached and
+passed around as `Arc<T>`.
+
+**Never construct `Arc::new(value)` for a hash-consed type.** Always obtain
+the canonical `Arc<T>` from the dedup cache. Concretely:
+
+- For freshly-constructed `ConfirmedBlockCertificate`s (e.g. from network or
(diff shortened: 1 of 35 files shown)


CONTRIBUTING.md
```diff
@@ -70,6 +70,29 @@ Contributions should generally follow the [Rust API guidelines](https://rust-lan
...
 
 
+## Hash-consed types and `Arc<T>`
+
+Any content-addressed immutable data (also known as hash-consed data) — for
+example `Block`, `Blob`, `ConfirmedBlockCertificate` — should be cached and
+passed around as `Arc<T>`.
+
+**Never construct `Arc::new(value)` for a hash-consed type.** Always obtain
+the canonical `Arc<T>` from the dedup cache. Concretely:
```
ma2bd: I wonder if this might be enforceable using a new type `linera_storage::Arc` and the `Deref` trait.

ma2bd (approved):

PR #6255, opened 2026-05-07
[testnet] Change what uncompressedKeys are being sent

MathieuDutSik:
## Motivation

The `uncompressedKeys` are being sent unordered, which leads to a reordering being done
onchain and a `O(n^2)` algorithm onchain, which is just bad.

Backport https://github.com/linera-io/linera-protocol/pull/6196

## Proposal

Do, the reordering offchain before submitting it.

## Test Plan

CI

## Release Plan

This is the backport to `testnet_conway`.

## Links

None.


deuszx (approved):
I prefer when the backport is pure cherry-pick – it makes the reviewing a bit easier to do as it's clearer that the diff matches the original.

PR #6257, opened 2026-05-07
Implement the access to the transaction index.

### LINKED ISSUE #2411: Allow enforcing that a transaction is the first one on the chain.
Details TBD. We could allow applications to get the transaction index in addition to the block height. Or just add an assertion that this is the first one.

MathieuDutSik:
## Motivation

The transaction index is a nice invariant to have available when running a smart contract.

Fixes https://github.com/linera-io/linera-protocol/issues/2411

## Proposal

In the killed PR https://github.com/linera-io/linera-protocol/pull/5606 a random number generator
was proposed but this was not the right design. Here we expose the transaction index which allows building
random number generators.

RNG have their use since using random number in some API queries can disable their caching.

## Test Plan

CI
A fixture test has been added which demonstrates the two use cases:
* That a transaction is the first one.
* That a RNG can be built on that.

## Release Plan

This is I think breaking with `testnet_conway`.

## Links

None.

### DIFF
--- a/linera-execution/src/lib.rs
+++ b/linera-execution/src/lib.rs
@@ -942,6 +942,13 @@ pub trait ContractRuntime: BaseRuntime {
     /// The authenticated owner for this execution, if there is one.
     fn authenticated_owner(&mut self) -> Result<Option<AccountOwner>, ExecutionError>;
 
+    /// The index of the current transaction within its block.
+    ///
+    /// Combined with the chain ID, application ID and block height, this can be used by
+    /// smart contracts to derive a deterministic seed (e.g. for pseudo-random number
+    /// generation).
+    fn transaction_index(&mut self) -> Result<u32, ExecutionError>;
+
     /// If the current message (if there is one) was rejected by its destination and is now
     /// bouncing back.
     fn message_is_bouncing(&mut self) -> Result<Option<bool>, ExecutionError>;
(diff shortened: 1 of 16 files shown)


linera-execution/src/lib.rs
```diff
@@ -942,6 +942,13 @@ pub trait ContractRuntime: BaseRuntime {
     /// The authenticated owner for this execution, if there is one.
     fn authenticated_owner(&mut self) -> Result<Option<AccountOwner>, ExecutionError>;
 
+    /// The index of the current transaction within its block.
+    ///
+    /// Combined with the chain ID, application ID and block height, this can be used by
+    /// smart contracts to derive a deterministic seed (e.g. for pseudo-random number
```
deuszx: I would remove the part about the usage for implementation of PRNG. I don't even think that's correct.

MathieuDutSik: Ok, to removing the PRNG usage.
But why do you think this is wrong?

deuszx: If you can predict the number generated (and here you can since as the chain owner and block proposer you control all the variables) then it's not a very good source of randomness – you can predict the numbers it generates and then every game is "rigged" (no randomness).

MathieuDutSik: I do not make any claim of cryptographic security.

The use case I was thinking is for `coinbase` access. If we send a request that got rate-limited, then sending it again can get you the cache, so the same values. If we add a random entry, then we can bypass the cache.

Now, can coinbase or cloudflare use this kind of information to predict the next call and avoid it? Yes, in principle. Would they do it. I kind of doubt it.

ma2bd:
not urgent and would need a rebase after #6551

PR #6258, opened 2026-05-07
Correct the `check_burn_completion`.

MathieuDutSik:
## Motivation

`check_burn_completion` in `monitor::linera` is meant to detect burns whose containing block was already accepted on EVM (e.g. by a previous relay instance before a crash) and mark them complete *without* spending gas on a redundant `addBlock` submission. The implementation was broken in two ways:

  1. **Filter mismatch with the deployed contract.** It queried `Transfer(address,address,uint256)` logs filtered by `address = bridge_addr`. `FungibleBridge.sol` in this repo is not itself an ERC-20 — it inherits only `Microchain`, holds an external `IERC20 token`, and the burn-release path is `token.transfer(target, …)`. The `Transfer` event is therefore emitted with `log.address == address(token)`, not `bridge_addr`, so the filter returned the empty set every iteration. (If the bridge were ever rebuilt as an ERC-20 itself, the original filter would be correct — but for the deployed contract on this branch, it is mismatched.)
  2. **Mark-by-existence semantic.** Independently of (1): the code marked a pending burn complete based on the existence of *any* matching log. Two burns to the same recipient (same amount or otherwise) are indistinguishable since the EVM event carries no per-burn identifier under either contract design.

  The check was therefore dead code that provided false confidence. Burns were being reconciled only via the `forward_cert` "already verified" string-match path inside `process_pending_burns`, which costs gas (the `addBlock` tx is sent and reverts) and depends on a brittle revert-string match.

  ## Proposal

The logs approach is broken since it cannot distinguish between different operations.
Using the hash exposes to the same problem of authentication since a block can contain several transactions either in Linera or EVM.

The proposed solution is then to write whether the entries are processed or not directly in the chain. This is fairly radical but avoids problems in the future.

## Test Plan

CI

The test `is_burn_processed_returns_true_after_release` was introduced.

## Release Plan

## Links

None

### DIFF
--- a/linera-bridge/src/monitor/linera.rs
+++ b/linera-bridge/src/monitor/linera.rs
@@ -2,13 +2,12 @@
 // SPDX-License-Identifier: Apache-2.0
 
 //! Linera-side monitoring: scans for BurnEvent stream events (auto-burns),
-//! forwards certificates to EVM, checks EVM for completion via ERC-20
-//! Transfer events, and retries unforwarded burns.
+//! forwards certificates to EVM, checks EVM for already-processed burns via
+//! `FungibleBridge.processedBurns`, and retries unforwarded burns.
 
 use std::{sync::Arc, time::Duration};
 
 use alloy::{primitives::Address, providers::Provider};
-use linera_base::data_types::BlockHeight;
 use tokio::sync::{Notify, RwLock};
 
 use super::{MonitorState, PendingBurn};
@@ -237,16 +236,23 @@ async fn linera_scan_iteration<E: linera_core::environment::Environment>(
     Ok(())
 }
 
+/// Detects burns that were already released on EVM (by a prior `addBlock`
+/// from us before a crash, or by another relayer). Queries the per-burn
+/// `processedBurns` mapping on FungibleBridge so the answer is intrinsic to
+/// the burn and survives any future change to the contract's per-block
+/// processing semantics. Marks each such burn complete *without* sending an
+/// EVM transaction, so `process_pending_burns` doesn't waste gas re-submitting
+/// blocks that would just revert with "block already verified".
 async fn check_burn_completion(
     monitor: &RwLock<MonitorState>,
     evm_client: &EvmClient<impl Provider>,
 ) -> anyhow::Result<()> {
-    let pending: Vec<(BlockHeight, usize, Address)> = {
+    let pending: Vec<(linera_base::data_types::BlockHeight, usize)> = {
         let state = monitor.read().await;
         state
             .pending_burns()
             .into_iter()
-            .map(|b| (b.value.height, b.value.burn_index, b.value.evm_recipient))
+            .map(|b| (b.value.height, b.value.burn_index))
             .collect()
     };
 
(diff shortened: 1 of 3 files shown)


linera-bridge/src/monitor/linera.rs
```diff
@@ -237,33 +259,162 @@ async fn linera_scan_iteration<E: linera_core::environment::Environment>(
...
-    for (height, burn_index, recipient) in pending {
-        let logs = evm_client.get_transfer_logs(recipient).await?;
-        if !logs.is_empty() {
-            monitor
-                .write()
-                .await
-                .complete_burn(height, burn_index)
-                .await;
+    for (height, burn_indices) in by_height {
+        let cert = match read_certificate_at_height(linera_client, height).await {
```
deuszx: If we add the block's hash to the `PendingBurn` struct we will not need this `read_certificate_at_height` or `confirmed_signed_hash`.

deuszx (changes requested):
Good catch. https://github.com/linera-io/linera-protocol/pull/6258/changes#r3202225843 should simplify the PR considerably.

linera-bridge/src/monitor/mod.rs
```diff
@@ -57,12 +57,22 @@ pub struct PendingDeposit {
...
 
 /// A pending burn detected by the Linera scanner, sent to the retry loop.
+///
+/// `signed_hash` is the validator-vote hash used as the EVM
+/// `verifiedBlocks` key (see `linera::confirmed_signed_hash`). It is computed
+/// at discovery from the burn's containing certificate and cached here so
+/// `check_burn_completion` can poll EVM without re-reading the cert. Burns
+/// recovered from the SQLite WAL after a restart carry `None` and are simply
+/// skipped by `check_burn_completion`; they still get forwarded normally by
+/// `process_pending_burns`, which tolerates "already verified".
```
deuszx: I don't understand this design choice. If we recover it from the DB they also can contain the hash of the certificate. Or put differently - a `PendingBurn` can't exist without the certificate hash. We don't need the `Option`.

linera-bridge/src/monitor/linera.rs
```diff
@@ -237,33 +269,136 @@ async fn linera_scan_iteration<E: linera_core::environment::Environment>(
...
+    sol! {
+        function verifiedBlocks(bytes32 signedHash) external view returns (bool);
+    }
+
+    /// Pinning test: the Rust-side `confirmed_signed_hash` must produce the
+    /// same value the EVM contract uses as a `verifiedBlocks` key. We verify
+    /// this by submitting a cert via `addBlock` and then checking that the
+    /// hash we computed locally is the one the contract recorded.
+    #[test]
+    fn confirmed_signed_hash_matches_microchain_verified_blocks() {
```
deuszx: I'm not convinced this Is the best test for the change this PR is making. The certificate hash equality isn't what we want to verify as the code doesn't manipulate the certificate. We should be testing that `addBlock` called with a different cert isn't "completing an unrelated `PendingBurn`.

deuszx: This test wouldn't catch the original bug either.

linera-bridge/src/monitor/linera.rs
deuszx, line 217: `BridgeDb::insert_burn` needs to be updated to store the certificate hash in `pending_burns` table.

deuszx (commented):
Please update the PR description as I can see that the code no longer matches what is being written there.

linera-bridge/src/relay/evm.rs
```diff
@@ -19,9 +20,22 @@ sol! {
     interface IFungibleBridge {
         function addBlock(bytes calldata data) external;
         function lightClient() external view returns (address);
+        function processedBurns(bytes32 burnKey) external view returns (bool);
```
deuszx: `bytes32` is very opaque – let's be more specific in the API.

linera-bridge/src/solidity/FungibleBridge.sol
```diff
@@ -36,7 +36,21 @@ contract FungibleBridge is Microchain {
...
 
-    constructor(address _lightClient, bytes32 _chainId, address _token, bytes32 _fungibleApplicationId)
+    /// Per-burn dedup mapping. Set inside `_onBlock` after each burn's
+    /// `token.transfer` succeeds. The key is `keccak256(abi.encode(height,
+    /// burnIndex))` where `burnIndex` is the 0-based position of the burn
+    /// in the flat sequence of matched burn events across the block (same
+    /// counter the off-chain relayer uses). The off-chain monitor reads
+    /// this mapping per pending burn to decide whether release has already
+    /// happened, without inspecting block-level state.
+    mapping(bytes32 => bool) public processedBurns;
```
deuszx: I was meaning to do it as a follow-up to your PR. I agree the certificate hash isn't gonna work if there are multiple `Burn` events in a single block and a key that includes the index of transaction is better. Let's be more specific though. IIRC, we can do a `mapping(blockheight => (mapping(txnIndex => bool))` (pseudocode) and then in the API just expect `(blockheight, txnIndex)` instead of the keccak.

MathieuDutSik: Yes, I will update the description.

Indeed the hash suffer from essentially the same problem as the previous code.

linera-bridge/src/monitor/mod.rs
MathieuDutSik: Yes, this is a bad design.

linera-bridge/src/monitor/linera.rs
MathieuDutSik: Correct. The hash does not address it.
We need some entries indicating precisely what has been burned or not. The log shortcut does not work. The hash also has issues. Instead we indicate that directly in the chain.

MathieuDutSik:
> Please update the PR description as I can see that the code no longer matches what is being written there.

Done.

linera-bridge/src/solidity/FungibleBridge.sol
MathieuDutSik: I think the `mapping(bytes32, _)` is adequate. The `mapping(_, mapping(_, _))` has its own issues.

deuszx:
> 1. so the filter returned the empty set every iteration.

I am not sure this is true - if that was the case then the relayer would never mark any PendingBurn as completed as the returned logs vector would have been empty. And it was processing Burns and marking them as completed.

It doesn't matter that much as this PR replaces the code with a new, better version.

deuszx:
I will have a closer look next week as I think there's still an issue higher up - the `Microchain` tracks already-processed (Linera) blocks and atores the hashes, rejecting the transaction early if it had already seen the Linera hash. It means that the new version is disjoint from `Microchain` one and - because they track different type of "keys" - they're incompatible.

PR #6263, opened 2026-05-07
Add linera-storage-decode-key binary to inspect RootKey partition keys

ndr-ds:
## Motivation

When investigating ScyllaDB large partitions via `system.large_partitions` or `nodetool toppartitions`, the partition keys are raw hex blobs. There was no existing tool to decode these back into the `RootKey` enum variants (`ChainState`, `BlockExporterState`, `BlockByHeight`, etc.), making it hard to identify which logical entities are consuming the most disk/CPU.

## Proposal

**Make `RootKey` and its `bytes()` method public** so external code (including binaries in the same crate) can roundtrip encode/decode.

**Add a new binary `linera-storage-decode-key`** that:
- Accepts hex-encoded partition key blobs as positional args or from stdin (one per line)
- Tolerates a leading `0x` prefix
- Skips non-hex tokens, so raw `cqlsh` table output or `nodetool toppartitions` output can be piped in directly
- Decodes each blob by BCS-deserializing it into `RootKey` and printing the `Debug` representation
- Supports `--scylla` (shorthand for `--strip-bytes 1`) because the ScyllaDB backend prepends a single `0x00` tag byte via `get_big_root_key`

Example usage against a cqlsh dump:

```
cqlsh -e "SELECT partition_key FROM system.large_partitions WHERE ks='kv' AND cf='table_linera';" \
  | linera-storage-decode-key --scylla
```

Each input token is printed back with its decoded variant tab-separated:

```
00022f3a...  ChainState(ChainId(CryptoHash(...)))
```

**A backport to `testnet_conway`** has been prepared on `ndr-ds/root-key-decoder-conway` and will be submitted as a separate PR once this one is reviewed.

## Test Plan

CI. Additionally validated manually against real `system.large_partitions` output from a `testnet_conway` validator, confirming all 31 large-partition keys decoded correctly as `ChainState`, `BlockByHeight`, and `BlockExporterState` entries.

## Release Plan

## Links


ma2bd (approved):
check CI though

Issue #5320, opened 2026-01-25
Add value introspection in the block explorer(s)


ma2bd:
Prerequisites:
- [ ] #5321
- [ ] Define a standard "environment" for custom (de)serialization of cryptographic primitives (e.g. `AccountOwner` should be serialized as "0x..." strings)
- [ ] [testnet only] Create a `log-format` example application that can notarize the formats of a given `ModuleId` (most likely as a BCS-encoded blob)

Extensions:
- [ ] Consider logging the formats for blobs (most apps only use a few types, so the explorer could brute-force it)
- [ ] Decide if we also extend the local (old) block explorer
- [ ] [main only] Consider extending `ModuleId` with one optional field `formats` (and maybe `source_code_url` while we're at it?).

Note: We could decide to develop first in `testnet_conway` then backport to `main`.

ma2bd:
#6546
#6547
#6549

PR #6264, opened 2026-05-08
[Frontport] Implement application introspection

MathieuDutSik:
## Motivation

The application introspection is one nice feature to have.

Frontport https://github.com/linera-io/linera-protocol/pull/6198
Frontport https://github.com/linera-io/linera-protocol/pull/5319

This is related to issue https://github.com/linera-io/linera-protocol/issues/5320 but does not address all the entries of this issue.

## Proposal

The original branches were done on `testnet_conway` which limited what could be done.
The idea there was to have a `formats-registry` that stored the description.

On `main` we have a larger design space and the design is to add another format blob that is
optional. The `publish_module` takes an optional format argument. This is significantly simpler
since we do not need to pass the `ChainId` and the `ApplicationId` to the `linera-explorer`.

The rest of the implementation is done similarly to the `testnet_conway` commit.

## Test Plan

CI

Running locally we could get the following.

The Operation Response decoded:
Intro1_transaction

The MUTATIONS decoded:
Intro2_mutations

The in message decoded:
Intro3_in_messages

The out message decoded:
Intro4_out_messages

The events:
Intro5_event

## Release Plan

This is the frontport from `testnet_conway`.

## Links

None


ma2bd (approved):

linera-execution/src/resources.rs
```diff
@@ -83,9 +83,13 @@ pub const RUNTIME_CRYPTO_HASH_SIZE: u32 = 32;
...
-/// This includes: `ModuleId` (2 hashes + VmRuntime) + `ChainId` + `BlockHeight` + `u32`.
-/// Variable parts (`parameters` and `required_application_ids`) are calculated separately.
-pub const RUNTIME_CONSTANT_APPLICATION_DESCRIPTION_SIZE: u32 = 2 * RUNTIME_CRYPTO_HASH_SIZE + RUNTIME_VM_RUNTIME_SIZE  // ModuleId
+///
+/// This includes: `ModuleId` (2 hashes + VmRuntime + Option<CryptoHash> discriminator)
+/// + `ChainId` + `BlockHeight` + `u32`. Variable parts (`parameters`,
+///   `required_application_ids`, and the optional formats blob hash payload) are
+///   calculated separately.
+pub const RUNTIME_CONSTANT_APPLICATION_DESCRIPTION_SIZE: u32 = 2 * RUNTIME_CRYPTO_HASH_SIZE + RUNTIME_VM_RUNTIME_SIZE  // ModuleId core
+    + 1                                                      // formats_blob_hash discriminator
```
ma2bd: `3 * RUNTIME_CRYPTO_HASH_SIZE` ?

MathieuDutSik: Yes, correct.

PR #6267, opened 2026-05-08
Move rpc_endpoint setting (EVM rpc node) from bridge parameters to instantiation.

deuszx:
## Motivation

`rpc_endpoint` on the EVM bridge is currently part of `BridgeParameters`,
which is the immutable Linera application-parameters slot — the value is
baked into the ApplicationId hash and can never change. Rotating an RPC
provider, switching from a public to a private node, or pointing at a
different mirror all force a fresh deployment with a new ApplicationId,
which cascades into wallet, relay, and frontend reconfiguration.

This PR moves `rpc_endpoint` from immutable parameters to mutable
per-deployment state and adds an operation to update it.

## Proposal

- New `BridgeInstantiationArgument { rpc_endpoint: String }` carrying
  the value at `create_application` time. Empty string keeps the
  existing skip-finality semantics.
- `evm-bridge` contract stores `rpc_endpoint` in a `RegisterView<String>`
  on its state. `instantiate` writes the argument's value (after the
  existing chain-ID sanity check); `process_deposit` and
  `verify_block_hash` read from state instead of `application_parameters`.
- New `BridgeOperation::SetRpcEndpoint { rpc_endpoint: String }` updates
  the state field. Gated on `authenticated_signer().is_some()` —
  chain-owner-only, mirrors `RegisterFungibleApp` /
  `RegisterFungibleBridge`. Idempotent; empty string allowed (drops back
  to skip-finality mode).
- `rpc_endpoint` is removed from `BridgeParameters`. Callers
  (`setup.sh`, the e2e tests) pass it via `--json-argument` /
  `BridgeInstantiationArgument` instead.
- Service GraphQL exposes the new state field; existing
  `isBlockHashFinalized` reads `rpc_endpoint` from state too.
- New integration test `set_rpc_endpoint`: chain-owner submits the
  operation, GraphQL query confirms state updated, then resets to empty
  string. (TestValidator hardcodes the signer, so the unauth-panic case
  isn't covered by the harness — would need an SDK-level helper.)

## Test Plan

- CI
- manually confirmed that UI demo works.

## Release Plan

- These changes should be backported to `main` 

## Links

- Stacked on top of [`evm-bridge-relocation`](https://github.com/linera-io/linera-protocol/compare/testnet_conway...evm-bridge-relocation)

### DIFF
--- a/linera-bridge/src/abi.rs
+++ b/linera-bridge/src/abi.rs
@@ -45,12 +45,7 @@ pub enum BridgeOperation {
     SetRpcEndpoint { rpc_endpoint: String },
 }
 
-/// Initial mutable state passed at `create_application`.
-///
-/// Distinct from `BridgeParameters` (immutable, part of the ApplicationId
-/// hash). Fields here are written to the contract's state during
-/// `instantiate` and may be mutated later by operations such as
-/// `SetRpcEndpoint`.
+/// Initial state passed at `create_application`.
 #[derive(Clone, Debug, Default, Deserialize, Eq, PartialEq, Serialize)]
 pub struct BridgeInstantiationArgument {
     /// JSON-RPC endpoint of the source EVM chain for finality verification.
--- a/linera-bridge/src/abi.rs
+++ b/linera-bridge/src/abi.rs
@@ -44,6 +44,9 @@ pub enum BridgeOperation {
     /// Can only be called once, by the chain owner.
     /// Must be set before ProcessDeposit can succeed.
     RegisterFungibleBridge { address: [u8; 20] },
+    /// Replace the bridge's RPC endpoint. Chain-owner-only.
+    /// Empty string disables finality verification.
+    SetRpcEndpoint { rpc_endpoint: String },
 }
 
 /// Initial mutable state passed at `create_application`.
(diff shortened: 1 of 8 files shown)


linera-bridge/src/abi.rs
```diff
@@ -44,4 +40,40 @@ pub enum BridgeOperation {
...
+    /// Empty string disables finality verification.
+    SetRpcEndpoint { rpc_endpoint: String },
+}
+
+/// Initial mutable state passed at `create_application`.
+///
+/// Distinct from `BridgeParameters` (immutable, part of the ApplicationId
+/// hash). Fields here are written to the contract's state during
+/// `instantiate` and may be mutated later by operations such as
+/// `SetRpcEndpoint`.
```
ma2bd: This description is confusing. IIRC this is just an initial value that will be overwritten later. I would actually suggest to remove it and use `None` as initial value.

deuszx: The initial value is whatever the constructor sets it to. It can be set to empty string which effectively means we're not verifying finality at all.

deuszx: Simplified in [Modify comment on BridgeInstantiationArgument](https://github.com/linera-io/linera-protocol/pull/6267/commits/7083142e765c68d376c88338138daca9607c8f47)

PR #6269, opened 2026-05-08
[Testnet] Implement `native-fungible` natively.

MathieuDutSik:
## Motivation

This is related to the PR https://github.com/linera-io/linera-protocol/pull/6256 for main for the same problem but with a different design.

## Proposal

Instead of separating the Native from VM applications, put the Native at the same level. This gets us the following:
```rust
/// The virtual machine runtime
pub enum VmRuntime {
    /// The Wasm virtual machine
    #[default]
    Wasm,
    /// The Evm virtual machine
    Evm,
    /// A runtime-native application implemented directly in `linera-execution`. The
    /// associated bytecode hashes in the `ModuleId` are sentinel values and are not
    /// expected to resolve to any blob.
    Native(NativeApplicationKind),
}
```

The tests `cargo test -p linera-service wasm` are passing. Several hacks being done:
* Use `AccountArg` in `native`. This is because there is no unification in `testnet_conway` for the `Account` type and we should use the one of `linera-sdk`. But since `linera-sdk` depends on `linera-execution`, that is a problem.
* There is no `allowances` in `testnet_conway` for `native-fungible` thought that could be introduced.
* More code being inserted.

## Test Plan

CI

## Release Plan

This 

## Links

None


ma2bd:
~~Thanks @MathieuDutSik for the PR. Did you close because of the CI failure? (He does seem like we would need to roll this out in several steps indeed)~~

Never mind, I just saw your comment on #6256

PR #6273, opened 2026-05-11
[testnet] Handle `EventsNotFound` on admin chain `EPOCH_STREAM_NAME` in committee lookup and cert processing

ma2bd:
## Motivation

The web client surfaces this error from `chain.synchronize_from_validators()`:

```
chain client error: Failed to communicate with a quorum of validators:
Local error handling validator response:
Local node operation failed: Events not found:
    [EventId { chain_id: <admin>, stream_id: StreamName(00), index: 46 }]
```

`stream_name: 00` is `EPOCH_STREAM_NAME`. `index: 46` means the admin chain's `NewCommittee` event for epoch 46 isn't in local storage. After the shared committee map landed in #6122 / #6131 and the chain-state fallback in #6186, there are still two `EventsNotFound` paths that bubble straight out instead of recovering:

1. `Client::admin_committee` / `admin_committees` call `handle_chain_info_query(admin_chain_id)` directly. If the admin chain has never been initialized locally and its description's `min_active_epoch..=max_active_epoch` range needs events that aren't synced, this errors out without trying to bootstrap.
2. `Client::process_certificates` only retries on `BlobsNotFound`. When applying a downloaded certificate for the user's chain, the worker calls `initialize_chain` → `get_committees(min..=max_active_epoch)` and raises `EventsNotFound` for admin `EPOCH_STREAM_NAME` events covering the chain's active epochs. That bubbles out of `synchronize_chain_state_from` and is wrapped as `NodeError::ResponseHandlingError` by `communicate_with_quorum` — matching the stack trace above.

The retry in `ChainClient::local_committee` from #6122 doesn't help with either: (1) is reached through callers like `validator_nodes()` that go through `admin_committee` directly, and (2) happens after `admin_committee` has already returned successfully, deep inside `communicate_with_quorum`.

## Proposal

**Commit 1 — `admin_committee` / `admin_committees` bootstrap.** Route both through a new helper `admin_chain_info(with_committees)` that catches `LocalNodeError::EventsNotFound` when **all** missing event IDs are on the admin chain's `EPOCH_STREAM_NAME`. On a hit it calls `bootstrap_admin_chain_from_genesis`, then retries the query.

`bootstrap_admin_chain_from_genesis` reads the genesis committee directly from storage via `Storage::get_or_load_committee(Epoch::ZERO)` — that path resolves through `NetworkDescription::genesis_committee_blob_hash` and does not require any synced events. It then calls `synchronize_chain_state_from_committee(admin_chain_id, genesis_committee)` (deliberately **not** `synchronize_chain_state`, which would re-enter `admin_committee` and recurse).

Both call sites in `admin_committees` — the initial query and the chain-state-fallback `chain_info_with_committees(admin_chain_id)` — share this retry.

Return types of `admin_committee` / `admin_committees` change from `LocalNodeError` to `chain_client::Error` so we can call `synchronize_chain_state_from_committee` (which uses that error type). The wrapper `ChainClient::admin_committee` was updated to match; all other callsites already used `?` to convert.

**Commit 2 — `EventsNotFound` retry in `process_certificates`.** Mirror the existing `EventsNotFound` retry from `stage_block_execution`: when `handle_certificate` raises `LocalNodeError::EventsNotFound`, call `download_certificates_for_events` to fetch the publisher-chain certificates that contain those events, then retry the handler. A per-iteration `downloaded_events` set prevents looping forever if the events truly can't be fetched.

This is the change that addresses the reported web-client error directly.

## Test Plan

- `cargo check --workspace --exclude linera-web` passes.
- The existing regression test `test_local_committee_syncs_admin_chain` (added in #6122) still passes against the new return types.
- Manual verification in the web client by the reporter is required to confirm the original `EventsNotFound` no longer surfaces.

A dedicated regression test for the `process_certificates` path (downloading admin-chain certificates to satisfy `NewCommittee` event lookups during certificate processing) and for the `admin_committee` bootstrap path (uninitialized admin chain) are not included yet and can be added in a follow-up.

## Release Plan

- These changes should be backported to the latest \`testnet\` branch, then
    - be released in a new SDK,

## Links

- Original PR: #6122 (backport #6131)
- Follow-up chain-state fallback: #6186

### DIFF
--- a/web/pnpm-workspace.yaml
+++ b/web/pnpm-workspace.yaml
@@ -1,4 +1,6 @@
 packages:
-- '@linera/client'
-- '@linera/metamask'
+  - '@linera/client'
+  - '@linera/metamask'
+allowBuilds:
+  esbuild: true
 publishBranch: testnet_conway
--- a/linera-core/src/client/chain_client/mod.rs
+++ b/linera-core/src/client/chain_client/mod.rs
@@ -712,7 +712,7 @@ impl<Env: Environment> ChainClient<Env> {
 
     /// Obtains the committee for the latest epoch on the admin chain.
     #[instrument(level = "trace")]
-    pub async fn admin_committee(&self) -> Result<(Epoch, Arc<Committee>), LocalNodeError> {
+    pub async fn admin_committee(&self) -> Result<(Epoch, Arc<Committee>), Error> {
         self.client.admin_committee().await
     }
 
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -747,16 +747,33 @@ impl<Env: Environment> Client<Env> {
                     break;
                 }
             }
-            info = Some(
+            // Track event IDs we've already tried to download to avoid looping
+            // forever when a publisher chain refuses to serve them.
+            let mut downloaded_events = HashSet::<EventId>::new();
+            let response = loop {
                 match self.handle_certificate(certificate.clone()).await {
                     Err(LocalNodeError::BlobsNotFound(blob_ids)) => {
                         self.download_blobs(remote_nodes, &blob_ids).await?;
-                        self.handle_certificate(certificate).await?
+                        continue;
                     }
-                    x => x?,
+                    Err(LocalNodeError::EventsNotFound(event_ids)) => {
+                        let new_events = event_ids
+                            .iter()
+                            .filter(|id| !downloaded_events.contains(id))
+                            .cloned()
+                            .collect::<Vec<_>>();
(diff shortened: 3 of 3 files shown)


web/pnpm-workspace.yaml
```diff
@@ -1,4 +1,6 @@
 packages:
-- '@linera/client'
-- '@linera/metamask'
+  - '@linera/client'
+  - '@linera/metamask'
+allowBuilds:
+  esbuild: true
```
afck: Is this change related?

PR #6275, opened 2026-05-11
Checkpoints MVP

afck:
## Motivation

Long chains take too long to synchronize. We need checkpoints so that new clients and validators don't need the full chain.

## Proposal

As a first step, implement an MVP: Checkpointing, but only for chains with no messages or events, whose execution state fits into a single blob.

## Test Plan

Tests were added; in particular a `client_tests` case where a new client synchronizes using a checkpoint, and doesn't download the block before the checkpoint.

## Release Plan

## Links

- Part of #460


linera-chain/src/block_tracker.rs
```diff
@@ -169,15 +181,19 @@ impl<'resources, 'blobs> BlockExecutionTracker<'resources, 'blobs> {
     }
 
     /// Returns a new TransactionTracker for the current transaction.
-    fn new_transaction_tracker(&self) -> Result<TransactionTracker, ChainError> {
-        Ok(TransactionTracker::new(
+    fn new_transaction_tracker(&mut self) -> Result<TransactionTracker, ChainError> {
```
ma2bd: nit: Maybe rename to create_new_transaction_tracker since this has a side effect now. (Also curious if this is needed)

linera-core/src/chain_worker/state.rs
```diff
@@ -901,32 +902,84 @@ where
             .map(|blob| (blob.id(), blob))
             .collect::<BTreeMap<_, _>>();
 
-        // If this block is higher than the next expected block in this chain, we're going
-        // to have a gap: do not execute this block, only update the outboxes and return.
+        // If this block is higher than the next expected block in this chain, we have a
+        // gap. There are two ways to bridge it:
+        //   * If the block is a checkpoint, install its execution-state snapshot from
```
ma2bd, line 907: Maybe not in this PR but I feel like we should explicitly support a sparse mode of chain execution for which we never execute blocks or restore checkpoints, even if we could. Restoring a large execution state for an untracked (say sender) chain by mistake on the client could really hurt UX.

linera-core/src/client/mod.rs
```diff
@@ -833,6 +833,39 @@ impl<Env: Environment> Client<Env> {
     /// If `until_block_time` is `Some`, stops before processing any certificate whose
     /// block timestamp is greater or equal than the given value.
     #[instrument(level = "trace", skip_all)]
+    /// Downloads the checkpoint certificate at `checkpoint_height` from `remote_node`
+    /// and processes it locally, if our chain isn't already past that height. The
+    /// worker's `process_confirmed_block` recognises the gap-plus-checkpoint case and
+    /// installs the chain's execution state from the checkpoint blob before re-running
+    /// the certificate.
```
ma2bd, line 834: This comment is weirdly positioned. Did you mean to replace the previous one?

linera-core/src/client/mod.rs
```diff
@@ -833,6 +833,39 @@ impl<Env: Environment> Client<Env> {
...
+        remote_node: &RemoteNode<Env::ValidatorNode>,
+        chain_id: ChainId,
+        checkpoint_height: BlockHeight,
+    ) -> Result<(), chain_client::Error> {
+        let local_next = self
+            .local_node
+            .chain_info(chain_id)
+            .await
+            .map(|info| info.next_block_height)
+            .unwrap_or(BlockHeight::ZERO);
```
ma2bd: Did we mean to silently catch all types of errors here?

linera-core/src/client/mod.rs
```diff
@@ -1714,7 +1747,20 @@ impl<Env: Environment> Client<Env> {
...
         let remote_info = remote_node.handle_chain_info_query(query).await?;
+
+        // If the validator advertises a checkpoint and our local tip is below it, fetch
+        // the checkpoint cert and blob and apply the cert directly. The chain worker's
+        // `process_confirmed_block` recognises the gap-plus-checkpoint case and installs
+        // the chain's execution state from the blob before re-executing the cert. The
+        // subsequent `download_certificates_from` call then resumes from the post-
+        // checkpoint height and skips downloading any pre-checkpoint blocks.
+        if let Some(checkpoint_height) = remote_info.requested_latest_checkpoint_height {
+            self.bootstrap_chain_from_checkpoint(remote_node, chain_id, checkpoint_height)
```
ma2bd, line 1764: Are we worried that validators may return an old checkpoint instead of the latest?

linera-core/src/chain_worker/state.rs
afck: OK, I made a note; but I agree it shouldn't be part of this PR, since it's a bigger change.

linera-chain/src/block_tracker.rs
afck: > Also curious if this is needed

It was just to avoid cloning the blob… but actually, blob data is in an `Arc`, so it's cheap anyway. I'll make it `&self` again…

linera-core/src/client/mod.rs
afck: Not really: If we call this function for every validator and one of them returned an old checkpoint, even if we restored that first, once we got to the later checkpoint, we'd skip the blocks in between and restore that.

It would be an optimization to ask for the checkpoints first, wait for a quorum, and then use only the latest one, to avoid the unnecessary work, but it's not a correctness issue. I'd leave this for later, since `synchronize_chain_state_from` has several call sites, not all of which involve asking multiple validators.

ma2bd (approved):

PR #6276, opened 2026-05-11
Fix the bug in relayer logic responsible for marking Linera `PendingBurn` txn as finished on EVM

deuszx:
## Motivation

The relayer marked `PendingBurn` as completed even when the on-chain `_onBlock` did not release any tokens for that burn. Two coupled bugs:

- `process_pending_burns` treated `forward_cert` Ok as proof of completion. `addBlock` returning Ok only proves the EVM tx didn't revert; it does not prove that this specific burn's `token.transfer` ran inside `_onBlock`. If the contract's event match silently rejects the burn (e.g. due to an `applicationId` mismatch between the scanner and the deployed `FungibleBridge`), tokens are never released yet the relayer claims success.
- `check_burn_completion`, the recovery path that was supposed to mark burns complete based on observed on-chain state, queried `Transfer` logs filtered by `bridge_addr`. `FungibleBridge` does not emit `Transfer`; the underlying ERC-20 token does, since the bridge releases burns via `token.transfer(target, …)`. The filter therefore returned an empty set every iteration and the recovery path was dead.

## Proposal

- Add an end-to-end test (`burn_completion_requires_on_chain_evidence.rs`) that deploys `FungibleBridge` with a synthetic `fungibleApplicationId` so `_onBlock` rejects every burn the scanner forwards. The test waits for the relayer to attempt forwarding, then asserts both that the recipient's ERC-20 balance is zero and that `linera_bridge_burns_completed` did not increment. On the buggy code the second assertion fails.
- Stop marking burns complete on `forward_cert` Ok in `process_pending_burns`. The function now mirrors `process_pending_deposits`: forward, log, and always increment the retry counter. Per-burn completion is decided exclusively by `check_burn_completion` polling on-chain state.
- Fix `check_burn_completion`'s log filter to target the bridged token contract with `from = bridge_addr` (topic1) and `to = recipient` (topic2). The token address is queried once via `FungibleBridge.token()` and cached in a `OnceCell`, mirroring the existing `light_client_addr` plumbing.

## Test Plan

- New e2e test under `linera-bridge/tests/e2e/tests/`. Run with `cargo test --manifest-path linera bridge/tests/e2e/Cargo.toml -- --ignored`. Verified locally against the docker-compose stack: passes with the
fix, fails on the prior code with `linera_bridge_burns_completed = 1` while `token.balanceOf(recipient) = 0`.
- CI (existing burn coverage in `auto_deposit_scan` continues to pass).

## Release Plan

- These changes should be backported to the `main` branch

## Links

- Supersedes #6258.

### DIFF
--- a/linera-bridge/src/relay/evm.rs
+++ b/linera-bridge/src/relay/evm.rs
@@ -20,6 +20,7 @@ sol! {
     interface IFungibleBridge {
         function addBlock(bytes calldata data) external;
         function lightClient() external view returns (address);
+        function token() external view returns (address);
     }
 }
 
@@ -43,6 +44,7 @@ pub struct EvmClient<P> {
     relayer_addr: Address,
     deposit_event_sig: B256,
     light_client_addr: tokio::sync::OnceCell<Address>,
+    token_addr: tokio::sync::OnceCell<Address>,
 }
 
 impl<P: Provider> EvmClient<P> {
@@ -62,6 +64,7 @@ impl<P: Provider> EvmClient<P> {
             relayer_addr,
             deposit_event_sig: deposit_event_signature(),
             light_client_addr,
+            token_addr: tokio::sync::OnceCell::new(),
         }
     }
 
@@ -96,14 +99,36 @@ impl<P: Provider> EvmClient<P> {
         Ok(all_logs)
     }
 
-    /// Queries ERC-20 Transfer events from the bridge to a recipient.
+    /// Queries ERC-20 `Transfer` events emitted by the bridged token
+    /// contract, with the bridge as the sender (`from`) and `recipient`
+    /// as the receiver (`to`). The events come from `address(token)`,
+    /// not from the bridge itself, because `FungibleBridge._onBlock`
+    /// releases tokens by calling `token.transfer(target, ...)`.
     pub async fn get_transfer_logs(&self, recipient: Address) -> Result<Vec<Log>> {
+        let token_addr = self.get_token_address().await?;
         let transfer_sig = alloy::primitives::keccak256("Transfer(address,address,uint256)");
         let filter = Filter::new()
-            .address(self.bridge_addr)
+            .address(token_addr)
             .event_signature(transfer_sig)
+            .topic1(B256::left_padding_from(self.bridge_addr.as_slice()))
             .topic2(B256::left_padding_from(recipient.as_slice()));
         Ok(self.provider.get_logs(&filter).await?)
+    }
+
+    /// Discovers the bridged ERC-20 token address from the FungibleBridge.
+    pub async fn get_token_address(&self) -> Result<Address> {
+        self.token_addr
+            .get_or_try_init(|| async {
+                let bridge = IFungibleBridge::new(self.bridge_addr, &self.provider);
+                let addr = bridge
+                    .token()
+                    .call()
+                    .await
+                    .context("failed to query FungibleBridge.token()")?;
+                Ok(addr)
+            })
+            .await
+            .copied()
     }
 
     /// BCS-serialize and forward a certified block to FungibleBridge on EVM.
--- a/linera-bridge/tests/e2e/tests/burn_completion_requires_on_chain_evidence.rs
+++ b/linera-bridge/tests/e2e/tests/burn_completion_requires_on_chain_evidence.rs
@@ -1,0 +1,395 @@
+// Copyright (c) Zefchain Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Demonstrates that the current relayer marks a `PendingBurn` as completed
+//! when `addBlock` returns Ok, even though the on-chain `_onBlock` did not
+//! release any tokens for that burn. The misconfiguration here — deploying
+//! `FungibleBridge` with a `fungibleApplicationId` that does not match the
+//! real wrapped-fungible app the scanner watches — makes `_onBlock` skip
+//! every burn event silently while still letting `addBlock` succeed.
+//!
+//! Expected outcome on the current code: the test fails on the
+//! `linera_bridge_burns_completed == 0` assertion because the relayer
+//! treats `forward_cert` Ok as evidence of completion. Post-fix (when
+//! completion is gated on a per-burn on-chain dedup query) the test passes:
+//! no `Transfer` happened, so the relayer must not claim the burn is done.
+
+#![recursion_limit = "512"]
+
+use std::{collections::BTreeMap, path::PathBuf, time::Duration};
+
+use alloy::{primitives::U256, providers::ProviderBuilder, sol};
+use anyhow::Context as _;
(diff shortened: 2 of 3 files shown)


linera-bridge/src/relay/evm.rs
```diff
@@ -96,16 +99,38 @@ impl<P: Provider> EvmClient<P> {
...
         let filter = Filter::new()
-            .address(self.bridge_addr)
+            .address(token_addr)
             .event_signature(transfer_sig)
+            .topic1(B256::left_padding_from(self.bridge_addr.as_slice()))
             .topic2(B256::left_padding_from(recipient.as_slice()));
         Ok(self.provider.get_logs(&filter).await?)
     }
 
+    /// Discovers the bridged ERC-20 token address from the FungibleBridge.
```
afck, line 118:
```suggestion
    /// Discovers the bridged ERC-20 token address from the `FungibleBridge`.
```

linera-bridge/tests/e2e/tests/burn_completion_requires_on_chain_evidence.rs
```diff
@@ -0,0 +1,395 @@
...
+
+/// `bytes32` value with no relation to any real wrapped-fungible application
+/// description hash. Used as the `_fungibleApplicationId` in the deployed
+/// `FungibleBridge` so that `_onBlock`'s app-id check rejects every burn
+/// event the scanner forwards. Non-zero to satisfy the constructor guard.
+const SYNTHETIC_APP_ID_BYTES32: &str =
+    "0xdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeef";
+
+#[tokio::test]
+#[ignore] // Requires pre-built docker images, Wasm, and relay binary.
```
afck, line 65: Does it run in CI anyway?

deuszx: Yes, in bridge-e2e job.

linera-bridge/tests/e2e/tests/burn_completion_requires_on_chain_evidence.rs
```diff
@@ -0,0 +1,395 @@
...
+    {
+        use linera_persistent::Persist;
+        let mut ks = linera_persistent::File::new(&keystore_path, signer.clone())?;
+        ks.persist().await?;
+    }
+    linera_wallet_json::PersistentWallet::create(&wallet_path, relay_genesis_config)?;
+
+    let relay_port = 3003u16;
+    let bridge_addr_str = format!("{bridge_addr}");
+    // The relayer needs an evm-bridge app id even though this test never
```
afck:
```suggestion
    // The relayer needs an evm-bridge app ID even though this test never
```

linera-bridge/tests/e2e/tests/burn_completion_requires_on_chain_evidence.rs
```diff
@@ -0,0 +1,395 @@
...
+    }
+    if !detected {
+        relay_handle.abort();
+        anyhow::bail!("Relayer never detected the burn");
+    }
+
+    // Allow time for `process_pending_burns` to run `forward_cert`. With a
+    // 2s monitor_scan_interval and exponential backoff starting at 5s, two
+    // 5s ticks after detection are plenty for the first attempt to land.
+    tokio::time::sleep(Duration::from_secs(15)).await;
```
afck: Is there a way to avoid this, too, and check in a loop instead?

linera-bridge/tests/e2e/tests/burn_completion_requires_on_chain_evidence.rs
```diff
@@ -0,0 +1,395 @@
...
+        burns_detected,
+        burns_completed,
+        burns_failed,
+        burns_pending,
+        ?token_balance,
+        "Final state after relayer cycle"
+    );
+
+    // Sanity: the misconfiguration scenario is reproduced. The contract's
+    // `_onBlock` rejected every burn event (app id mismatch), so no
```
afck:
```suggestion
    // `_onBlock` rejected every burn event (app ID mismatch), so no
```

linera-bridge/tests/e2e/tests/burn_completion_requires_on_chain_evidence.rs
```diff
@@ -0,0 +1,395 @@
...
+    // `token.transfer` ran. Holds true under both current and post-fix
+    // relayer code; if it fails, the test setup itself is wrong.
```
afck: Don't document "current" code that's being changed in this PR.

afck (approved):

linera-bridge/tests/e2e/tests/burn_completion_requires_on_chain_evidence.rs
deuszx: Ah, I thought I removed all that. This is a leftover from when I write the test first and then the fix.

linera-bridge/tests/e2e/tests/burn_completion_requires_on_chain_evidence.rs
deuszx: I can rewrite the test so that it waits with a timeout.

linera-bridge/tests/e2e/tests/burn_completion_requires_on_chain_evidence.rs
deuszx: Done in [Address review comments](https://github.com/linera-io/linera-protocol/pull/6276/commits/f17217a910d21e0943e41e50063c8f11462a498e).

linera-bridge/tests/e2e/tests/burn_completion_requires_on_chain_evidence.rs
deuszx: Done in [Address review comments](https://github.com/linera-io/linera-protocol/pull/6276/commits/f17217a910d21e0943e41e50063c8f11462a498e).

linera-bridge/tests/e2e/tests/burn_completion_requires_on_chain_evidence.rs
deuszx: Done in [Address review comments](https://github.com/linera-io/linera-protocol/pull/6276/commits/f17217a910d21e0943e41e50063c8f11462a498e).

linera-bridge/tests/e2e/tests/burn_completion_requires_on_chain_evidence.rs
deuszx: Done in [Replace sleep with awaiting txn count](https://github.com/linera-io/linera-protocol/pull/6276/commits/f84d7eb7695f9e746160e364cd0ac76be7793ef6)

PR #6278, opened 2026-05-11
Retry on `EventsNotFound`

afck:
## Motivation

There are still some places missing where we retry on `BlobsNotFound` but not on `EventsNotFound`, specifically when it comes to missing epoch events.

## Proposal

Fix those; deduplicate some of the code.

## Test Plan

A regression test was added that fails on `main`.

## Release Plan

- Release SDK.

## Links


linera-core/src/client/mod.rs
```diff
@@ -2010,6 +2076,17 @@ impl<Env: Environment> Client<Env> {
     }
 }
 
+/// Returns the items in `ids` that are not yet present in `already_downloaded`.
+fn filter_new<T: Clone + Eq + std::hash::Hash>(
```
deuszx, line 2038: nit: I'd say it's `return_new` as it returned copies of the elements from the `ids` slice. `filter` – IIRC – does not copy?

linera-core/src/local_node.rs
```diff
@@ -72,6 +74,27 @@ pub enum LocalNodeError {
...
 }
 
+impl LocalNodeError {
+    /// Returns true if this error means the chain is not (yet) initialized in
+    /// the local node: its `ChainDescription` blob, an admin-chain epoch event
+    /// needed to load its initial committee, or its activation state is
+    /// missing. Callers that fall back to a "no information yet" default
+    /// should treat all three the same — but unrelated `BlobsNotFound` /
+    /// `EventsNotFound` errors must still surface as real failures.
+    pub fn is_chain_uninitialized(&self) -> bool {
```
deuszx: All those three variants mean exactly the same and can be handled/recovered in the same way? I'm worried this shows that we have 3 error variants to represent the same thing or that the caller of this method will incorrectly do some post-processing (b/c it won't check the specific error).

linera-core/src/client/mod.rs
afck: Not sure… almost every function "returns" something. Maybe `collect_new`? OR `new_blob_ids`?

linera-core/src/client/chain_client/mod.rs
```diff
@@ -657,15 +657,11 @@ impl<Env: Environment> ChainClient<Env> {
                 self.synchronize_chain_state(self.chain_id).await?;
                 self.client.local_node.chain_info(self.chain_id).await?
             }
-            Err(LocalNodeError::EventsNotFound(event_ids))
-                if event_ids
-                    .iter()
-                    .all(|event_id| event_id.stream_id == StreamId::system(EPOCH_STREAM_NAME)) =>
-            {
+            Err(LocalNodeError::EventsNotFound(event_ids)) => {
```
deuszx: Previously we assumed that syncing only `EPOCH_STREAM_NAME` event stream was safe? Now we want to sync all of the streams?

linera-core/src/local_node.rs
afck: That's true. Reorganized this a bit in https://github.com/linera-io/linera-protocol/pull/6278/commits/52ca4a1b523f02be4e0ad9bd5cefa2b330c94664. I think that's cleaner? At the source we know _why_ certain blobs or events weren't found, i.e. whether its an "inactive chain" situation etc.

linera-execution/src/system.rs
```diff
@@ -851,9 +851,19 @@ where
...
+        // only data init reads. If either is missing, the chain isn't yet
+        // initialized locally — surface that uniformly as `InactiveChain`,
+        // rather than as a generic missing-blob / missing-event error that
+        // downstream retry logic would mistake for operation-time data.
         let description_blob = self
             .read_blob_content(BlobId::new(chain_id.0, BlobType::ChainDescription))
-            .await?;
+            .await
+            .map_err(|err| match err {
+                ExecutionError::BlobsNotFound(_) => ExecutionError::InactiveChain(chain_id),
```
ma2bd: Do we need the remapping? If yes, can we create a proper error variant (something like `ChainInitializationFailed` ?) instead of recycling `ExecutionError::InactiveChain`?

(And if this leads to removing `ExecutionError::InactiveChain` or treating it as a non-retryable internal error that's even better.)

linera-core/src/client/mod.rs
ma2bd: "new" looks like a constructor.

This is a set exclusion (or substraction) so I would suggest: `exclude(_set)` or `substract(_set)`

linera-core/src/client/chain_client/mod.rs
ma2bd: nit: also possible `maybe_local_chain_info`

afck: Done in https://github.com/linera-io/linera-protocol/pull/6278/commits/3bfc283f5a4d27e161b3fdee33af2627e74b58dc.

linera-execution/src/system.rs
afck: I removed it, and cleaned up error handling some more.

linera-core/src/client/chain_client/mod.rs
```diff
@@ -657,17 +657,6 @@ impl<Env: Environment> ChainClient<Env> {
...
-            Err(LocalNodeError::EventsNotFound(event_ids))
-                if event_ids
-                    .iter()
-                    .all(|event_id| event_id.stream_id == StreamId::system(EPOCH_STREAM_NAME)) =>
-            {
-                // `initialize_and_save_if_needed` couldn't start the chain because
-                // the admin chain's epoch events aren't synced yet.
-                self.synchronize_chain_state(self.client.admin_chain_id)
-                    .await?;
-                self.client.local_node.chain_info(self.chain_id).await?
```
deuszx, line 669: I can't find where you remove flows that return `EventsNotFound` when committee is not found so how come it's OK to remove this branch now? How are you making sure the new code would never return `EventsNotFound` anymore?

afck: As far as I can tell those flows were never there in the first place!

If you have the chain description blob, you must already have verified the certificate that created or used it, which is signed by that committee, so you must have the event.

linera-core/src/client/mod.rs
```diff
@@ -692,12 +691,31 @@ impl<Env: Environment> Client<Env> {
...
 
+    /// If any of `event_ids` are not in `already_downloaded`, downloads the
+    /// publisher certificates that contain them, marks them as downloaded, and
+    /// returns true. Returns false (without downloading) if every reported event
+    /// has already been downloaded — that prevents an infinite retry loop when
+    /// the events are genuinely unavailable.
+    pub(crate) async fn download_new_events_into(
+        &self,
+        event_ids: &[EventId],
+        already_downloaded: &mut HashSet<EventId>,
```
ma2bd: nit: This signature is a tell that `HashSet<EventId>` should be a newtype struct (say Something `struct EventSetDownloader(HashSet<EventId>)`) in `&mut self` position. Then you could probably inline `filter_new`.

ma2bd: (And probably the same for blobs by the way)

afck: Done for events. For blobs it would actually have only one call site, so I don't think it's worth it.

ma2bd (approved):

PR #6280, opened 2026-05-11
Fix burn completion false positive via per-burn on-chain dedup

deuszx:
## Motivation

`check_burn_completion` used to mark a `PendingBurn` complete when ANY ERC-20 `Transfer` to the recipient existed. Multiple burns to the same recipient (a common pattern — a user repeatedly burning to their own
EVM address) collapsed to a single signal: once one transfer landed, every pending burn for that recipient flipped to completed regardless of whether its own `token.transfer` had actually run inside `_onBlock`. A
UI demo reproduced the failure exactly — 11 burns submitted, only 3 actually transferred on EVM, all 11 reported complete by the relayer.

## Proposal

- Add an on-chain per-burn flag. `FungibleBridge` writes `processedBurns[keccak256(abi.encode(height, evt.index))] = true` after each successful `token.transfer` inside `_onBlock`, and exposes an
`isBurnProcessed(uint64 height, uint32 eventIndex)` view.
- The relayer's `check_burn_completion` queries `isBurnProcessed` per pending burn instead of inspecting ERC-20 `Transfer` logs by recipient. Mark-by-existence is gone.
- `eventIndex` is the underlying Linera `Event.index` — the same value the off-chain scanner sees. `PendingBurn.burn_index: usize` (a synthetic flat counter) is renamed to `event_index: u32` and sourced from
`Event.index`. SQLite column, in-memory `HashMap` key, and metric labels updated.
- `Microchain.verifiedBlocks` is untouched. Multi-tx burn submission (splitting a single Linera block across several `addBlock` calls) needs that gate relaxed and is left for a follow-up.

Two new e2e tests reproduce the bug end-to-end and verify the fix:

- `multiple_burns_same_recipient_across_blocks` — 5 burns to the same EVM recipient pre-populated as separate Linera blocks before the relayer is spawned. Pre-fix: only some `addBlock` calls land;
`check_burn_completion`'s recipient-only filter false-marks every pending burn; the ERC-20 balance falls short of `5 * amount`. Post-fix: per-burn dedup forces each `addBlock`, balance reaches the full amount.
- `multiple_burns_same_block` — 4 burns (one recipient repeated) bundled into a single chain B block and absorbed into one chain A block carrying all 4 BurnEvents. Pre-fix: `_onBlock` transfers correctly but the
off-chain `check_burn_completion` never marks the burns complete (`burns_completed` stays at 0). Post-fix: per-burn `isBurnProcessed` returns true for every flag written in the same `_onBlock` call.

## Test Plan

- New e2e tests under `linera-bridge/tests/e2e/tests/`. Run with `cargo test --manifest-path linera-bridge/tests/e2e/Cargo.toml -- --ignored`. Both verified locally against the docker-compose stack: each fails on
the prior code with the expected diagnostic (`balance == 2e18, expected 5e18` and `burns_completed == 0, expected 4`) and passes with the fix.
- Existing `linera-bridge` unit tests (73 tests) continue to pass.
- CI.

## Release Plan

- These changes should be backported to the `main` branch

## Links

- Supersedes #6258 and stacks on #6276.

### DIFF
--- a/linera-bridge/src/monitor/linera.rs
+++ b/linera-bridge/src/monitor/linera.rs
@@ -81,8 +81,8 @@ pub(crate) async fn process_pending_burns<E: linera_core::environment::Environme
         };
 
         let credit_height = pending.height;
-        let burn_index = pending.burn_index;
-        tracing::info!(?credit_height, burn_index, "Processing burn...");
+        let event_index = pending.event_index;
+        tracing::info!(?credit_height, event_index, "Processing burn...");
 
         // Read the certificate at the burn's block height (already contains the auto-burn).
         let cert = match async {
@@ -106,13 +106,13 @@ pub(crate) async fn process_pending_burns<E: linera_core::environment::Environme
             Err(e) => {
                 tracing::warn!(
                     ?credit_height,
-                    burn_index,
+                    event_index,
                     "Failed to read certificate: {e:#}"
                 );
                 monitor
                     .write()
                     .await
-                    .mark_burn_retried(credit_height, burn_index);
+                    .mark_burn_retried(credit_height, event_index);
                 continue;
             }
         };
@@ -122,12 +122,12 @@ pub(crate) async fn process_pending_burns<E: linera_core::environment::Environme
             bcs::to_bytes(&cert).expect("failed to BCS-serialize ConfirmedBlockCertificate");
         if let Some(db) = monitor.read().await.db() {
             if let Err(e) = db
-                .store_burn_raw(credit_height, burn_index, &cert_bytes)
+                .store_burn_raw(credit_height, event_index, &cert_bytes)
                 .await
             {
                 tracing::warn!(
                     ?credit_height,
-                    burn_index,
+                    event_index,
                     "Failed to store burn raw bytes: {e:#}"
                 );
             }
@@ -141,15 +141,15 @@ pub(crate) async fn process_pending_burns<E: linera_core::environment::Environme
         // on-chain state. Here we only forward and retry.
         match evm_client.forward_cert(&cert).await {
             Ok(()) => {
-                tracing::info!(?credit_height, burn_index, "Burn forwarded to EVM");
+                tracing::info!(?credit_height, event_index, "Burn forwarded to EVM");
                 relay::update_balance_metrics(evm_client, linera_client).await;
             }
             Err(e) => {
                 let msg = format!("{e:#}");
                 if msg.contains("already verified") {
-                    tracing::trace!(?credit_height, burn_index, "Block already verified on EVM");
+                    tracing::trace!(?credit_height, event_index, "Block already verified on EVM");
                 } else {
-                    tracing::warn!(?credit_height, burn_index, "EVM forwarding failed: {e:#}");
+                    tracing::warn!(?credit_height, event_index, "EVM forwarding failed: {e:#}");
                 }
             }
         }
@@ -157,7 +157,7 @@ pub(crate) async fn process_pending_burns<E: linera_core::environment::Environme
         monitor
             .write()
             .await
-            .mark_burn_retried(credit_height, burn_index);
+            .mark_burn_retried(credit_height, event_index);
     }
 }
 
@@ -194,10 +194,10 @@ async fn linera_scan_iteration<E: linera_core::environment::Environment>(
     for block in &blocks {
         let height = block.block().header.height;
         let burn_events = find_burn_events(&block.block().body.events, fungible_app_id);
-        for (burn_index, burn_event) in burn_events.into_iter().enumerate() {
+        for (event_index, burn_event) in burn_events {
             new_burns.push((
                 height,
-                burn_index,
+                event_index,
                 Address::from(burn_event.target),
                 burn_event.amount,
             ));
@@ -205,14 +205,14 @@ async fn linera_scan_iteration<E: linera_core::environment::Environment>(
     }
 
     let mut tracked_any = false;
-    for (height, burn_index, recipient, amount) in &new_burns {
-        tracing::info!(?height, burn_index, %recipient, %amount, "Discovered burn");
+    for (height, event_index, recipient, amount) in &new_burns {
+        tracing::info!(?height, event_index, %recipient, %amount, "Discovered burn");
         let was_new = monitor
             .write()
             .await
             .track_burn(PendingBurn {
                 height: *height,
-                burn_index: *burn_index,
+                event_index: *event_index,
                 evm_recipient: *recipient,
                 amount: *amount,
             })
@@ -232,16 +232,20 @@ async fn linera_scan_iteration<E: linera_core::environment::Environment>(
     Ok(())
 }
 
+/// Polls the FungibleBridge per pending burn and marks any that the EVM
+/// has already released. Per-burn rather than per-recipient — the previous
+/// `Transfer`-log filter would mark every pending burn for a recipient
+/// complete the moment ANY transfer to that recipient was observed.
 async fn check_burn_completion(
     monitor: &RwLock<MonitorState>,
     evm_client: &EvmClient<impl Provider>,
 ) -> anyhow::Result<()> {
-    let pending: Vec<(BlockHeight, usize, Address)> = {
+    let pending: Vec<(BlockHeight, u32)> = {
         let state = monitor.read().await;
         state
             .pending_burns()
             .into_iter()
-            .map(|b| (b.value.height, b.value.burn_index, b.value.evm_recipient))
+            .map(|b| (b.value.height, b.value.event_index))
             .collect()
     };
 
@@ -249,14 +253,23 @@ async fn check_burn_completion(
         return Ok(());
     }
 
-    for (height, burn_index, recipient) in pending {
-        let logs = evm_client.get_transfer_logs(recipient).await?;
-        if !logs.is_empty() {
-            monitor
-                .write()
-                .await
-                .complete_burn(height, burn_index)
-                .await;
+    for (height, event_index) in pending {
+        match evm_client.is_burn_processed(height, event_index).await {
+            Ok(true) => {
+                monitor
+                    .write()
+                    .await
+                    .complete_burn(height, event_index)
+                    .await;
+            }
+            Ok(false) => {}
+            Err(e) => {
+                tracing::warn!(
+                    ?height,
+                    event_index,
+                    "is_burn_processed query failed: {e:#}"
+                );
+            }
         }
     }
 
--- a/linera-bridge/src/monitor/linera.rs
+++ b/linera-bridge/src/monitor/linera.rs
@@ -233,9 +233,7 @@ async fn linera_scan_iteration<E: linera_core::environment::Environment>(
 }
 
 /// Polls the FungibleBridge per pending burn and marks any that the EVM
-/// has already released. Per-burn rather than per-recipient — the previous
-/// `Transfer`-log filter would mark every pending burn for a recipient
-/// complete the moment ANY transfer to that recipient was observed.
+/// has already released.
 async fn check_burn_completion(
     monitor: &RwLock<MonitorState>,
     evm_client: &EvmClient<impl Provider>,
--- a/linera-bridge/src/monitor/mod.rs
+++ b/linera-bridge/src/monitor/mod.rs
@@ -57,10 +57,15 @@ pub struct PendingDeposit {
 }
 
 /// A pending burn detected by the Linera scanner, sent to the retry loop.
+///
+/// `event_index` is the underlying Linera `Event.index` — the position of
+/// the burn event within its stream ("burns" on the configured fungible
+/// application). Used both as the dedup key off-chain and as the value
+/// the `FungibleBridge.isBurnProcessed(height, eventIndex)` view consumes.
 #[derive(Debug, Clone, serde::Serialize)]
 pub struct PendingBurn {
     pub height: BlockHeight,
-    pub burn_index: usize,
+    pub event_index: u32,
     pub evm_recipient: Address,
     pub amount: Amount,
 }
@@ -96,7 +101,7 @@ pub type TrackedBurn = Tracked<PendingBurn>;
 /// In-memory monitoring state shared across scan loops and HTTP handlers.
 pub struct MonitorState {
     pub(crate) deposits: HashMap<DepositKey, TrackedDeposit>,
-    pub(crate) burns: HashMap<(BlockHeight, usize), TrackedBurn>,
+    pub(crate) burns: HashMap<(BlockHeight, u32), TrackedBurn>,
     pub(crate) last_scanned_evm_block: u64,
     pub(crate) last_scanned_linera_height: BlockHeight,
     db: Option<db::BridgeDb>,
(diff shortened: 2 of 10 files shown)


linera-bridge/src/monitor/linera.rs
```diff
@@ -232,31 +232,44 @@ async fn linera_scan_iteration<E: linera_core::environment::Environment>(
     Ok(())
 }
 
+/// Polls the FungibleBridge per pending burn and marks any that the EVM
+/// has already released. Per-burn rather than per-recipient — the previous
+/// `Transfer`-log filter would mark every pending burn for a recipient
+/// complete the moment ANY transfer to that recipient was observed.
```
afck: Let's not document removed code.

linera-bridge/src/monitor/mod.rs
```diff
@@ -57,10 +57,15 @@ pub struct PendingDeposit {
...
+///
+/// `event_index` is the underlying Linera `Event.index` — the position of
+/// the burn event within its stream ("burns" on the configured fungible
+/// application). Used both as the dedup key off-chain and as the value
+/// the `FungibleBridge.isBurnProcessed(height, eventIndex)` view consumes.
 #[derive(Debug, Clone, serde::Serialize)]
 pub struct PendingBurn {
     pub height: BlockHeight,
-    pub burn_index: usize,
+    pub event_index: u32,
```
afck, line 68: (I'm beginning to wonder whether we should make the event index 64 bits rather than 32, just to be safe.)

linera-bridge/src/relay/linera.rs
```diff
@@ -142,11 +142,14 @@ impl<E: linera_core::environment::Environment> Clone for LineraClient<E> {
     }
 }
 
-/// Find all BurnEvents in a block's event streams for a given application.
+/// Finds all BurnEvents in a block's event streams for a given application,
```
afck:
```suggestion
/// Finds all `BurnEvent`s in a block's event streams for a given application,
```

linera-bridge/tests/e2e/src/lib.rs
```diff
@@ -328,3 +328,164 @@ pub async fn start_compose(compose_file: &std::path::Path, project_name: &str) -
...
+        .nth(3)
+        .context("manifest dir has fewer than 3 ancestors")?
+        .to_path_buf();
+    let wasm_dir = repo_root.join("examples/target/wasm32-unknown-unknown/release");
+    let wf_contract = Bytecode::load_from_file(wasm_dir.join("wrapped_fungible_contract.wasm"))?;
+    let wf_service = Bytecode::load_from_file(wasm_dir.join("wrapped_fungible_service.wasm"))?;
+    let (wf_module_id, _) = chain_client
+        .publish_module(wf_contract, wf_service, VmRuntime::Wasm)
+        .await?
+        .expect("publish wrapped-fungible module committed");
```
afck, line 380: (Isn't this re-implementing test helpers we already have elsewhere?)

linera-bridge/tests/e2e/tests/burn_completion_requires_on_chain_evidence.rs
```diff
@@ -182,19 +144,13 @@ async fn relayer_does_not_mark_burn_complete_when_token_was_not_transferred() ->
...
-             {erc20_addr} \
-             'transfer(address,uint256)(bool)' \
-             {bridge_addr} \
-             500000000000000000000"
-        ),
         project_name,
         &compose_file,
+        erc20_addr,
+        bridge_addr,
+        500_000_000_000_000_000_000u128,
```
afck: (Or `500 * 10u128.pow(18)`?)

linera-bridge/tests/e2e/tests/multiple_burns_same_block.rs
```diff
@@ -0,0 +1,334 @@
+// Copyright (c) Zefchain Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Verifies that multiple BurnEvents in a single Linera block are all
```
afck:
```suggestion
//! Verifies that multiple `BurnEvent`s in a single Linera block are all
```

linera-bridge/tests/e2e/tests/multiple_burns_same_block.rs
```diff
@@ -0,0 +1,334 @@
...
+//! relayed and completed on the EVM side. The user chain submits one
+//! block carrying N Transfer operations to N distinct Address20
```
afck:
```suggestion
//! block carrying N `Transfer` operations to N distinct `Address20`
```

linera-bridge/tests/e2e/tests/multiple_burns_same_block.rs
```diff
@@ -0,0 +1,334 @@
...
+//! recipients on the bridge chain. The test process drives the inbox
+//! processing on the bridge chain so the resulting Credit messages
```
afck:
```suggestion
//! processing on the bridge chain so the resulting `Credit` messages
```

linera-bridge/tests/e2e/tests/multiple_burns_same_block.rs
```diff
@@ -0,0 +1,334 @@
...
+//! all land in one chain-A block — a single cert with N BurnEvents.
```
afck:
```suggestion
//! all land in one chain-A block — a single cert with N `BurnEvent`s.
```

linera-bridge/tests/e2e/tests/multiple_burns_same_block.rs
```diff
@@ -0,0 +1,334 @@
...
+//! After the relayer is spawned, exactly one `addBlock` call should
+//! release all N tokens (one `token.transfer` per burn in `_onBlock`)
+//! and the relayer must mark every pending burn complete.
+//!
+//! Expected outcome on the pre-fix code: `_onBlock` atomically
```
afck: Let's not document "the pre-fix code".

linera-bridge/tests/e2e/tests/multiple_burns_same_block.rs
```diff
@@ -0,0 +1,334 @@
...
+        "0x3C44CdDdB6a900fa2b585dd299e03d12FA4293BC".parse()?,
+        "0x90F79bf6EB2c4f870365E785982E1f101E93b906".parse()?,
+    ];
+    let burn_amount = Amount::from_tokens(BURN_AMOUNT_TOKENS);
+
+    // Bundle every Transfer into a SINGLE chain-B block, then drive
+    // chain A's inbox once. The N Credit messages travel together
+    // and `process_inbox` materialises one chain-A block containing
+    // all N BurnEvents.
+    let operations: Vec<Operation> = recipients
```
afck:
```suggestion
    let operations = recipients
```

linera-bridge/tests/e2e/tests/multiple_burns_same_block.rs
```diff
@@ -0,0 +1,334 @@
...
+    let height_before_inbox = cc_a.chain_info().await?.next_block_height;
+    cc_a.process_inbox().await?;
+    let height_after_inbox = cc_a.chain_info().await?.next_block_height;
+    assert_eq!(
+        height_after_inbox.0,
+        height_before_inbox.0 + 1,
+        "chain A must produce exactly ONE block carrying all {NUM_BURNS} BurnEvents \
+         (before={}, after={})",
+        height_before_inbox.0,
+        height_after_inbox.0,
```
afck:
```suggestion
         (before={height_before_inbox}, after={height_after_inbox})",
```

linera-bridge/tests/e2e/tests/multiple_burns_same_block.rs
```diff
@@ -0,0 +1,334 @@
...
+    .await
+    {
+        relay_handle.abort();
+        return Err(error);
+    }
+    // One addBlock processes all N burns in `_onBlock`; allow the
+    // relayer enough time to call it and for the post-tx settle path
+    // (check_burn_completion / per-burn isBurnProcessed query) to run
+    // at least one iteration.
+    tokio::time::sleep(Duration::from_secs(30)).await;
```
afck: Shouldn't we just get the metrics in a loop with much shorter sleep time per iteration, until they match the expectation?

linera-bridge/tests/e2e/tests/multiple_burns_same_recipient_across_blocks.rs
```diff
@@ -0,0 +1,299 @@
...
+//! block). When the relayer starts, its first scan iteration finds
+//! all 5 pending burns at once — the multi-pending shape required to
+//! reproduce the mark-by-existence false positive.
+//!
+//! Expected outcome on the pre-fix code: only some `addBlock` calls
+//! complete before `check_burn_completion` false-marks every pending
+//! burn complete; the recipient's ERC-20 balance ends up below
+//! `5 * amount` while `linera_bridge_burns_completed == 5`. The
+//! balance assertion fails. Post-fix: per-burn dedup forces the
+//! relayer to forward every burn, balance reaches the full amount.
```
afck: Same here: I'd rather explain simply what the current code is expected to do, rather than refer to "the UI-demo bug" and "pre-fix code".

linera-bridge/tests/e2e/tests/multiple_burns_same_recipient_across_blocks.rs
```diff
@@ -0,0 +1,299 @@
...
+    // means at least one `PendingBurn` was marked complete without
+    // its `token.transfer` actually landing on-chain — the UI-demo bug.
+    assert_eq!(
+        token_balance,
+        expected_balance,
+        "recipient must accumulate every burn; got {token_balance}, expected {expected_balance}"
+    );
+
+    Ok(())
+}
```
afck, line 291: Could this simply be merged into the other test by having _one_ of the recipients repeat? I.e. instead of having one looong test with four different recipients and one looong test with one repeated recipient, why not have one with recipients `[A, B, C, D, D, D, D]`?

linera-bridge/src/monitor/mod.rs
deuszx: Yeah, why not.

linera-bridge/tests/e2e/src/lib.rs
deuszx: I don't think we have the helpers for this, no. We can reuse this new helper in many of tests in `linera-bridge` though.

linera-bridge/tests/e2e/tests/multiple_burns_same_block.rs
deuszx: I'll run a `cargo fmt` in a separate PR. I didn't notice it was never ran (in CI) for those e2e tests – they're a standalone workspace.

linera-bridge/tests/e2e/tests/multiple_burns_same_recipient_across_blocks.rs
deuszx: One test is for multiple linera blocks, the other is for one linera block.

linera-bridge/tests/e2e/tests/multiple_burns_same_block.rs
afck: I don't think `cargo fmt` would have made this change?

linera-bridge/src/monitor/linera.rs
deuszx: Done in [Address review comments](https://github.com/linera-io/linera-protocol/pull/6280/commits/c0e94d105babc73d523ff1cf10b2ab29a769bdb2)

linera-bridge/src/relay/linera.rs
deuszx: Done in [Address review comments](https://github.com/linera-io/linera-protocol/pull/6280/commits/c0e94d105babc73d523ff1cf10b2ab29a769bdb2)

linera-bridge/tests/e2e/tests/burn_completion_requires_on_chain_evidence.rs
deuszx: Done in [Address review comments](https://github.com/linera-io/linera-protocol/pull/6280/commits/c0e94d105babc73d523ff1cf10b2ab29a769bdb2)

linera-bridge/tests/e2e/tests/multiple_burns_same_block.rs
deuszx: Done in [Address review comments](https://github.com/linera-io/linera-protocol/pull/6280/commits/c0e94d105babc73d523ff1cf10b2ab29a769bdb2)

linera-bridge/tests/e2e/tests/multiple_burns_same_block.rs
deuszx: Done in [Address review comments](https://github.com/linera-io/linera-protocol/pull/6280/commits/c0e94d105babc73d523ff1cf10b2ab29a769bdb2)

linera-bridge/tests/e2e/tests/multiple_burns_same_block.rs
deuszx: Done in [Address review comments](https://github.com/linera-io/linera-protocol/pull/6280/commits/c0e94d105babc73d523ff1cf10b2ab29a769bdb2)

linera-bridge/tests/e2e/tests/multiple_burns_same_block.rs
deuszx: Done in [Address review comments](https://github.com/linera-io/linera-protocol/pull/6280/commits/c0e94d105babc73d523ff1cf10b2ab29a769bdb2)

linera-bridge/tests/e2e/tests/multiple_burns_same_block.rs
deuszx: Done in [Address review comments](https://github.com/linera-io/linera-protocol/pull/6280/commits/c0e94d105babc73d523ff1cf10b2ab29a769bdb2)

linera-bridge/tests/e2e/tests/multiple_burns_same_block.rs
deuszx: Done in [Address review comments](https://github.com/linera-io/linera-protocol/pull/6280/commits/c0e94d105babc73d523ff1cf10b2ab29a769bdb2)

linera-bridge/tests/e2e/tests/multiple_burns_same_block.rs
deuszx: Done in [Address review comments](https://github.com/linera-io/linera-protocol/pull/6280/commits/c0e94d105babc73d523ff1cf10b2ab29a769bdb2)

linera-bridge/tests/e2e/tests/multiple_burns_same_block.rs
deuszx: Done in [Address review comments](https://github.com/linera-io/linera-protocol/pull/6280/commits/c0e94d105babc73d523ff1cf10b2ab29a769bdb2)

linera-bridge/tests/e2e/tests/multiple_burns_same_recipient_across_blocks.rs
deuszx: Done in [Address review comments](https://github.com/linera-io/linera-protocol/pull/6280/commits/c0e94d105babc73d523ff1cf10b2ab29a769bdb2)

afck (approved):

PR #6282, opened 2026-05-12
Measure max Linera Burns in single EVM transaction.

deuszx:
## Motivation

Measure how many `BurnEvent`s in a single Linera block can fit into one `addBlock` EVM transaction under different chain gas limits, so the relayer can be tuned with empirical numbers rather than guesswork and regressions in per-burn EVM cost get caught in CI.

## Proposal

- New parametrised e2e test `linera-bridge/tests/e2e/tests/burns_per_evm_tx.rs` using `test_case` (one row per chain). For each `ChainSpec`, it binary-searches the largest `N` such that
`eth_estimateGas(bridge.addBlock(cert_N))` is `<= chain_spec.block_gas_limit`. Estimation is a dry run, so the bridge's `processedBurns` mapping stays untouched and a single bridge instance handles every
iteration.
- Two chain rows: `ethereum` (30M gas limit, floor 50) and `base` (240M, floor 190). `min_expected_burns` set conservatively — about 5–10% below observed `max_n` — so unrelated changes that significantly bump
per-burn EVM cost get caught.

## Test Plan

Test is `#[ignore]`-gated and requires the docker stack + wrapped-fungible Wasm artifacts. Run via:
```
cd linera-bridge/tests/e2e
cargo test --test burns_per_evm_tx -- --ignored --test-threads=1 --nocapture
```
Verified locally:

| chain    | max_n | gas at max_n | block gas limit | floor |
|----------|-------|--------------|-----------------|-------|
| ethereum | 52    | 29 320 900   | 30 000 000      | 50    |
| base     | 199   | 238 782 494  | 240 000 000     | 190   |

## Release Plan

## Links

- Stacked on #6280.

### DIFF
--- a/linera-bridge/tests/e2e/src/lib.rs
+++ b/linera-bridge/tests/e2e/src/lib.rs
@@ -199,6 +199,34 @@ pub async fn deploy_linera_token(
     parse_broadcast_address(compose, project_name, compose_file, "DeployLineraToken.s.sol").await
 }
 
+/// Same as [`deploy_linera_token`] but overrides the initial token supply
+/// minted to the anvil deployer via the script's `TOKEN_SUPPLY` env var.
+/// `supply_attos` is the raw atto amount — pass `tokens * 10u128.pow(18)`
+/// for whole-token amounts.
+pub async fn deploy_linera_token_with_supply(
+    compose: &DockerCompose,
+    project_name: &str,
+    compose_file: &std::path::Path,
+    supply_attos: u128,
+) -> anyhow::Result<Address> {
+    exec_ok(
+        compose,
+        "foundry-tools",
+        &format!(
+            "env TOKEN_SUPPLY={supply_attos} \
+             forge script /contracts/script/DeployLineraToken.s.sol \
+             --root /contracts \
+             --rpc-url http://anvil:8545 \
+             --private-key {ANVIL_PRIVATE_KEY} \
+             --broadcast"
+        ),
+        project_name,
+        compose_file,
+    )
+    .await;
+    parse_broadcast_address(compose, project_name, compose_file, "DeployLineraToken.s.sol").await
+}
+
 /// Deploys FungibleBridge via the `DeployFungibleBridge.s.sol` forge
 /// script and returns the deployed contract address.
 pub async fn deploy_fungible_bridge(
@@ -517,17 +545,8 @@ where
     E: linera_core::environment::Environment,
 {
     use anyhow::Context as _;
-    chain_client
-        .synchronize_from_validators()
-        .await
-        .map_err(|e| anyhow::anyhow!(e))?;
-    let info = chain_client
-        .chain_info()
-        .await
-        .map_err(|e| anyhow::anyhow!(e))?;
+    chain_client.synchronize_from_validators().await?;
+    let info = chain_client.chain_info().await?;
     let hash = info.block_hash.context("chain has no blocks yet")?;
-    chain_client
-        .read_certificate(hash)
-        .await
-        .map_err(|e| anyhow::anyhow!(e))
+    Ok(chain_client.read_certificate(hash).await?)
 }
--- a/linera-bridge/tests/e2e/src/lib.rs
+++ b/linera-bridge/tests/e2e/src/lib.rs
@@ -489,3 +489,45 @@ where
     }
     anyhow::bail!("wait_for_relay_metrics timed out after {timeout:?}")
 }
+
+/// Sets the anvil block gas limit via the `evm_setBlockGasLimit` JSON-RPC
+/// method. Lets the test choose a ceiling well above the chain spec being
+/// measured so `eth_estimateGas` never aborts on anvil's own limit — the
+/// numeric comparison happens in Rust against `ChainSpec::block_gas_limit`.
+pub async fn set_anvil_block_gas_limit(
+    provider: &impl alloy::providers::Provider,
+    limit: u64,
+) -> anyhow::Result<()> {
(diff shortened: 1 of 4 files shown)


linera-bridge/tests/e2e/src/lib.rs
```diff
@@ -489,3 +489,45 @@ where
...
+/// method. Lets the test choose a ceiling well above the chain spec being
+/// measured so `eth_estimateGas` never aborts on anvil's own limit — the
+/// numeric comparison happens in Rust against `ChainSpec::block_gas_limit`.
+pub async fn set_anvil_block_gas_limit(
+    provider: &impl alloy::providers::Provider,
+    limit: u64,
+) -> anyhow::Result<()> {
+    use std::borrow::Cow;
+    let hex_limit = format!("0x{limit:x}");
+    let _: bool = provider
```
afck, line 531:
```suggestion
    provider
```

linera-bridge/tests/e2e/src/lib.rs
```diff
@@ -489,3 +489,45 @@ where
...
+        .raw_request(Cow::Borrowed("evm_setBlockGasLimit"), (hex_limit,))
+        .await?;
+    Ok(())
+}
+
+/// Fetches the `ConfirmedBlockCertificate` for the chain's current head
+/// block. Mirrors the `sync -> chain_info -> read_certificate` walk in
+/// `linera-bridge/src/monitor/linera.rs::process_pending_burns` but specialised
```
afck, line 539:
```suggestion
/// `linera-bridge/src/monitor/linera.rs::process_pending_burns` but specialized
```

linera-bridge/tests/e2e/src/lib.rs
```diff
@@ -489,3 +489,45 @@ where
...
+    chain_client: &linera_core::client::ChainClient<E>,
+) -> anyhow::Result<linera_chain::types::ConfirmedBlockCertificate>
+where
+    E: linera_core::environment::Environment,
+{
+    use anyhow::Context as _;
+    chain_client
+        .synchronize_from_validators()
+        .await
+        .map_err(|e| anyhow::anyhow!(e))?;
```
afck: Is this necessary? Doesn't `?` do that automatically?

linera-bridge/tests/e2e/tests/burns_per_evm_tx.rs
```diff
@@ -0,0 +1,366 @@
+// Copyright (c) Zefchain Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! For each `ChainSpec` (Ethereum / Base / Base Sepolia / …), binary-searches
+//! the largest `N` such that `eth_estimateGas(bridge.addBlock(cert_N))` is
+//! `<= chain_spec.block_gas_limit`, where `cert_N` is a Linera block carrying
+//! exactly `N` `BurnEvent`s. `eth_estimateGas` runs the EVM in dry-run mode,
```
afck, line 7: Are all `BurnEvent`s really the exact same size? If we e.g. have ULEB-encoded amounts, we should use a value that uses the maximum number of bytes for these tests.

deuszx: I don't know but we don't estimate – we construct the actual Linera block and upload it to the EVM.

afck: Yes, but the actual Linera block will contain BCS-serialized `Amount`s, won't it? And larger amounts actually use more bytes? So we should use sufficiently large ones here, I think.

deuszx: Those values aren't going to be hardcoded as ceiling for the relayer – it's more of a "good to know" thing. We'll need to make the relayer AND contracts support cases where we need to submit more `Burn`s than fit into a single EVM block

afck: But still: Don't we want these numbers to be pessimistic, i.e. work for all types of `Burn`s rather than only those with small amounts?

deuszx: Sure but this shouldn't matter. We have to deserialize the whole Linera block, verify signatures, pull in all the contracts' codes (on EVM), iterate over all events, etc. The few numbers having few zeros more won't change anything.

afck: (OK; not a blocker. I just thought it would be a simple change to make, to be on the safe side.)

afck (approved):

linera-bridge/tests/e2e/src/lib.rs
deuszx: Ah, yes `ChainClient::Error` and `LocalNodeError` both derive `thiserror::Error`.

linera-bridge/tests/e2e/tests/burns_per_evm_tx.rs
deuszx: with
```rust
const BURN_AMOUNT_TOKENS: u128 = 10u128.pow(15);
```
it's still
```
2026-05-12T13:24:25.096934Z  INFO burns_per_evm_tx: RESULT chain="base" max_n=199 gas_at_max_n=238796312 block_gas_limit=240000000
2026-05-12T13:25:22.533506Z  INFO burns_per_evm_tx: RESULT chain="ethereum" max_n=52 gas_at_max_n=29331838 block_gas_limit=30000000
```

PR #6285, opened 2026-05-12
`walrus-meter`

### LINKED ISSUE #4742: Applications don't load with Rust 1.87 or later

Twey:
## Motivation

We'd like upgrade our Rust version.  Currently we're being held back by `wasm-instrument`, which depends on `parity-wasm`, which is unmaintained and doesn't support some of the newer extensions to Wasm that the Rust compiler now emits by default.

## Proposal

Replace it with the new crate `walrus-meter`, which does the same thing but built on the maintained `walrus` library.

As a bonus, `walrus` intends to eventually support preserving DWARF debug info through its transformations.

## Test Plan

`walrus-meter` includes tests for compatibility with (how we use) `wasm-instrument`.  We have existing tests in this codebase, run in CI, that ensure that fuel costs don't change.  Those tests have in fact changed in this PR because `wasm-instrument` overcharges in some corner cases:

```
;; wasm-instrument          ;; walrus-meter
block  ;; label = @3        block  ;; label = @3
  i64.const $FOO_BAR_COST |   i64.const $FOO_COST
  call $spend                 call $spend
  block  ;; label = @4        block  ;; label = @4
    block  ;; label = @5        block  ;; label = @5
      ;; … FOO …                  ;; … FOO …
      br_if 0 (;@5;)              br_if 0 (;@5;)
      unreachable                 unreachable
    end                         end
    br 1 (;@3;)                 br 1 (;@3;)
  end                         end
                          >   i64.const $BAR_COST
                          >   call $spend
  ;; … BAR …                  ;; … BAR …
end                         end
```

This is a bug in the `wasm-instrument` case: the `br 1` will, if reached, exit `@3` causing `BAR` to be jumped over, but `wasm-instrument` has already merged the cost of it into the minimum cost of block `@3`, meaning `wasm-instrument` is charging for instructions that aren't executed.

In `linera-counter` the cost of `BAR` is 42, resulting in a 168 fuel discrepancy, i.e. 42×4 (the test does 4 iterations).

This may result in this PR being incompatible with Testnet Conway, if we assume that fuel costs don't change.

## Release Plan

## Links

- fixes https://github.com/linera-io/linera-protocol/issues/4742

### DIFF
--- a/linera-execution/src/wasm/mod.rs
+++ b/linera-execution/src/wasm/mod.rs
@@ -26,7 +26,6 @@ use linera_base::data_types::Bytecode;
 #[cfg(with_metrics)]
 use linera_base::prometheus_util::MeasureLatency as _;
 use thiserror::Error;
-use wasm_instrument::{gas_metering, parity_wasm};
 #[cfg(with_wasmer)]
 use wasmer::{WasmerContractInstance, WasmerServiceInstance};
 #[cfg(with_wasmtime)]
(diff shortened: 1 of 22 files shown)


linera-execution/src/wasm/mod.rs
```diff
@@ -209,45 +208,14 @@ impl UserServiceModule for WasmServiceModule {
...
-        parity_wasm::deserialize_buffer(&bytecode.bytes)?,
-        gas_metering::host_function::Injector::new(
-            "linera:app/contract-runtime-api",
-            "consume-fuel",
-        ),
-        &WasmtimeRules,
+    let instrumented_module = walrus_meter::instrument(
+        &bytecode.bytes,
+        walrus_meter::costs::Wasmtime,
+        ("linera:app/contract-runtime-api", "consume-fuel"),
```
deuszx, line 225: So how are costs set for walrus? seems like the defaults are all zeros?

Twey: I built our ‘Wasmtime costs’ into `walrus-meter` as `walrus_meter::costs::Wasmtime`.

Mind you our Wasmtime costs don't actually represent Wasmtime because Wasmtime counts falling off the end of a function as a `return`, which we don't do (with `wasm-instrument` or `walrus-meter`).  With that in mind I might rename it.

deuszx: Does it mean that if we ever want to change the costs we need to modify the `walrus-meter` crate?

deuszx: Can we do it like it was with parity-wasm? I.e. have `walrus-meter` expose an interface and then implement it on our end?

Twey: That is how it works, it just comes with some default implementations.

deuszx: OK but why not make it clear what actual costs are? I think hiding it behind the default implementation is confusing. Especially that it's no longer how it used to be.

Twey: Sure, done. It adds a dependency on `walrus` but probably that's okay.

deuszx: Thank you.

ma2bd (approved):

deuszx (approved):

PR #6289, opened 2026-05-13
Multi-leader jitter

### LINKED ISSUE #1666: Retry with random delays in multi-leader mode.
In a chain in multi-leader mode, when retrying to propose a block, we should add a random delay.

Otherwise, if multiple clients keep retrying immediately, they will keep conflicting.

afck:
## Motivation

If there are conflicting proposals in the first multi-leader round that probably means that several clients are trying to make a proposal at the same time. When they see the conflict, they would currently propose immediately in the next round, and likely cause a conflict again.

## Proposal

Wait for a random amount of time instead.

This can be disabled with a CLI option.

## Test Plan

A unit test was added for the delay computation. (Happy to add a higher-level case to `client_tests` if necessary.)

## Release Plan

## Links

- Closes https://github.com/linera-io/linera-protocol/issues/1666.


linera-base/src/ownership.rs
```diff
@@ -229,8 +227,77 @@ impl ChainOwnership {
...
         self.owners.is_empty() && self.super_owners.contains(owner)
     }
+
+    /// Returns whether `owner` has the lowest `hash(owner, round)` among the eligible
+    /// multi-leader proposers, and should therefore propose immediately rather than wait
+    /// out a jitter delay. Returns `false` for `open_multi_leader_rounds`, where the set
+    /// of proposers is unbounded.
+    ///
+    /// This is a gentle-clients convention; it is not enforced by the protocol.
+    pub fn is_preferred_multi_leader_proposer(
```
deuszx: This doesn't need to be `pub` it seems.

linera-core/src/client/chain_client/mod.rs
```diff
@@ -2175,6 +2188,41 @@ impl<Env: Environment> ChainClient<Env> {
...
+        let round_start = if round == info.manager.current_round {
+            match (info.manager.round_timeout, ownership.round_timeout(round)) {
+                (Some(end), Some(duration)) => end.saturating_sub(duration),
+                _ => now,
+            }
+        } else {
+            now
+        };
+        let propose_at = round_start.saturating_add(delay);
+        (propose_at > now).then_some(propose_at)
```
deuszx, line 2264: What is `propose_at == now`? Shouldn't we return `Some(_)` as well?

afck: No, we only return `Some` if we want to wait rather than propose now.

deuszx (approved):
Consider adding support to linera web client. Other than that – LGTM

afck:
This exposes a problem with how we handle `Committed` vs. `Conflict` outcomes. Need to think about this; back to draft.

deuszx (approved):

linera-core/src/client/chain_client/mod.rs
```diff
@@ -1415,6 +1419,45 @@ impl<Env: Environment> ChainClient<Env> {
         }
     }
 
+    /// Returns `Committed` if the committed block reflects our request — same
+    /// authenticated owner, and our `operations`. All other transactions must be ones that
+    /// are added automatically, to process messages, streams or new epochs.
+    fn classify_committed(
+        &self,
+        certificate: ConfirmedBlockCertificate,
+        operations: &[Operation],
```
ma2bd, line 1428: @afck This is great. It seems like we're progressing towards a model where we have a number of parallel queues (say operations, bundles, events) feeding new transactions inside blocks and we could start tracking the fate of user operations with callbacks (and support cancelation while in the queue).

It seems very reasonable to index operations by (AuthenticatedOwner, Bytes) by the way (and maybe other data that I'm forgetting that can plausibly affect execution). We could decide that applications are responsible for
* refined tracking of user operations with additional metadata inside the bytes, if needed.
* not depending on transient data like block height (or leaky global state after https://github.com/linera-io/linera-protocol/pull/6032)

PR #6294, opened 2026-05-13
Drop unused `PostedMessage::index` field.

afck:
## Motivation

Since messages are bundled, the index of an individual message in a sender block is not used anymore.

## Proposal

Remove it.

## Test Plan

CI

## Release Plan

## Links


deuszx (approved):

deuszx:
Wait @afck. Your solution to `memory-safe` cannot be merged as-is. We cannot be confident that all replacements of `assembly` with `*memory-safe` are correct. This needs to be fixed in the Solidity generator.

deuszx (changes requested):
Blocking this until we can get rid of `memory-safe` wrapper. Needs to be addressed in `serde-refleaction`.

deuszx:
Git forks won't work, I think. We can't publish to crates if we have git dependencies

afck:
I know, I just wanted to make sure it passes CI now! I'll make this "Ready for review" only when we have a new `serde-reflection` version.

ma2bd:
@afck I made a new release of the package `serde-generate` (0.34.0)

deuszx (approved):

PR #6295, opened 2026-05-13
Support splitting Linera block into multiple EVM transactions.

deuszx:
## Motivation

A Linera block can carry enough burns to exceed any EVM chain's block gas limit (~30M on Ethereum, ~240M on Base), in which case `FungibleBridge.addBlock(cert)` does not fit in one EVM transaction. Today the
relayer cannot make progress on such a block: `Microchain.verifiedBlocks` blocks a retry, and there is no per-burn entry point. This PR makes the bridge multi-tx-friendly so the relayer can split such a block
into several smaller EVM transactions.

## Proposal

- `Microchain.verifiedBlocks` mapping dropped; `addBlock` is now idempotent on the subclass's per-burn dedup (`processedBurns` keyed on `(height, Event.index)`). Documented requirement that subclasses MUST be idempotent under repeated calls.
- New `FungibleBridge.processBurns(bytes cert, uint32 txIndex, uint32[] eventPositionsInTx)` entry point: verifies the cert once and releases the subset of burns at the requested positions within a single transaction. Direct `body.events[txIndex][pos]` access — no nested-loop scan.
- `processBurns` is idempotent like `_onBlock`: positions already in `processedBurns` are skipped silently rather than reverted, so overlap with a prior `addBlock` (or a racing/retrying `processBurns`) doesn't poison the whole chunk.
- Shared helpers `_burnKey`, `_isMatchingBurn`, `_releaseBurn` extracted from `_onBlock` + `processBurns` to keep the dedup-key invariant in one place and tighten the matching logic. `_releaseBurn` applies checks-effects-interactions (sets the dedup flag BEFORE `token.transfer`) for both entry points — a reentrancy defense in depth.
- Off-chain `PendingBurn` gains `tx_index` and `event_pos_in_tx` fields alongside the existing `event_index` (still the on-chain dedup key). The scanner's `find_burn_events` widens its return tuple; SQLite `pending_burns` / `finished_burns` add two columns.
- `process_pending_burns` does per-height dispatch (cert fetch + persist + decide addBlock vs chunked) with branches factored into helpers: `submit_addblock` or `submit_chunked`.
- Per-burn retry accounting: cert-fetch and gas-estimate failures are infrastructure-level and no longer consume any burn's retry budget. `addBlock` success leaves retry counts alone (completion is observed asynchronously by `check_burn_completion`); `addBlock` submit failure bumps retry for every burn at the height. Chunked submit failure bumps retry only for that chunk's burns; remaining chunks still attempt submission.
- Permanently-oversized burns (single position still doesn't fit) are marked `failed` and drop out of subsequent snapshots, instead of poisoning their tx group on every retry pass.

## Test Plan

- New Forge tests on `linera-bridge/src/solidity/`:
  - `MicrochainIdempotencyTest::test_addBlock_can_be_called_repeatedly_for_same_cert` — re-entry on the same cert is now allowed.
  - Eight `FungibleBridgeProcessBurnsTest` cases — `single_position_marks_processed`, `multi_position_marks_both_processed`, `already_processed_skips` (idempotent no-op + no double-release), `tx_index_out_of_range_reverts`, `event_pos_out_of_range_reverts`, `non_burn_event_reverts`, `empty_positions_reverts`, `partial_overlap_releases_remaining` (settle pos 1 then [0,1] → pos 0 settles, pos 1 silently skipped).

- Rust unit tests under `cargo test -p linera-bridge --features relay`:
  - `relay::settlement::tests::estimate_fits_*` — 8 cases covering `Ok`, `gas required exceeds`, `exceeds block gas limit`, anvil's `Out of gas: gas required exceeds allowance: N` (empirically verified against
anvil 1.6 for both calldata-too-large and infinite-loop constructors), real contract revert with non-empty `data` (bubbles up), bare `out of gas` (bubbles up — not a fits signal), unrelated RPC error, and
transport-layer error.

- Existing `evm::microchain::tests::test_microchain_*` flipped to assert the new idempotent behaviour.

- New e2e test `linera-bridge/tests/e2e/tests/multi_tx_burn_chunking.rs`:
  - Anvil's block gas limit is dialled down via `evm_setBlockGasLimit` after the bridge deploys.
  - Chain B submits a single Linera block carrying 8 burns to 8 distinct recipients.
  - The relayer's `addBlock(cert)` estimate exceeds the limit and falls back to chunked `processBurns` per tx; all 8 burns settle, balances match `1 * 10^18`, `linera_bridge_burns_completed = 8`.

- Existing `multiple_burns_same_block` and `multiple_burns_same_recipient_across_blocks` e2e tests still cover the per-burn `isBurnProcessed` semantics from #6276 and remain unchanged.

## Release Plan

## Links

### DIFF
--- a/linera-bridge/src/monitor/linera.rs
+++ b/linera-bridge/src/monitor/linera.rs
@@ -87,145 +87,179 @@ pub(crate) async fn process_pending_burns<E: linera_core::environment::Environme
 
         for super::PendingBurnsAtHeight {
             height,
-            event_indices: event_indices_at_height,
+            event_indices,
             by_tx,
         } in groups
         {
-            // `event_indices_at_height` are the stream-index values for every
-            // pending burn at this height under the same `max_retries`
-            // snapshot as `by_tx`. `mark_burn_retried` / `mark_burn_failed`
-            // key on event_index; the per-tx chunking driven below uses
-            // (tx_index, pos_in_tx) and maps back through
-            // `MonitorState::event_index_for_pos` only on failure paths.
-
+            // Infrastructure-level failures (cert fetch, gas estimate)
+            // skip the height without consuming any burn's retry budget.
             let cert = match fetch_cert_at_height(linera_client, height).await {
                 Ok(cert) => cert,
                 Err(e) => {
-                    // Cert fetch is an infrastructure-level failure (e.g. node
-                    // sync). Skip the height — no burn was attempted, so don't
-                    // consume any retry budget.
                     tracing::warn!(?height, "Failed to read certificate: {e:#}");
                     continue;
                 }
             };
 
-            persist_cert_bytes(monitor, height, &event_indices_at_height, &cert).await;
+            persist_cert_bytes(monitor, height, &event_indices, &cert).await;
 
-            // Dry-run addBlock. If it fits, one tx settles everything in the block.
             match estimate_fits(evm_client.estimate_add_block_gas(&cert).await) {
-                Ok(true) => match evm_client.forward_cert(&cert).await {
-                    Ok(()) => {
-                        // Completion is async via `check_burn_completion`, so
-                        // we don't mark the burns completed here — and don't
-                        // bump retry either, because the call succeeded.
-                        tracing::info!(
-                            ?height,
-                            count = event_indices_at_height.len(),
-                            "Burns forwarded via addBlock"
-                        );
-                        relay::update_balance_metrics(evm_client, linera_client).await;
-                    }
-                    Err(e) => {
-                        // addBlock attempted every burn at this height — all of
-                        // them just took a failed attempt, so bump retry once
-                        // per burn.
-                        tracing::warn!(?height, "addBlock submission failed: {e:#}");
-                        let mut state = monitor.write().await;
-                        for ei in &event_indices_at_height {
-                            state.mark_burn_retried(height, *ei);
-                        }
-                    }
-                },
-                Ok(false) => {
-                    for (tx_index, positions) in &by_tx {
-                        // Iterative LIFO split-to-fit per tx group. The
-                        // algorithm is inlined (rather than factored into a
-                        // pure helper that takes an async predicate) because
-                        // an `AsyncFn` closure capturing `&EvmClient` across
-                        // awaits trips an HRTB Send bound under `tokio::spawn`.
-                        let mut stack: Vec<Vec<u32>> = vec![positions.clone()];
-                        let mut chunks: Vec<Vec<u32>> = Vec::new();
-                        // Positions that cannot fit even on their own — these
-                        // get marked `failed` so they drop out of subsequent
-                        // `pending_burns_by_height_and_tx` snapshots instead
-                        // of poisoning their tx group forever.
-                        let mut oversized: Vec<u32> = Vec::new();
-                        while let Some(slice) = stack.pop() {
-                            let est = evm_client
-                                .estimate_process_burns_gas(&cert, *tx_index, &slice)
-                                .await;
-                            let fits = estimate_fits(est).unwrap_or(false);
-                            if fits {
-                                chunks.push(slice);
-                                continue;
-                            }
-                            if slice.len() == 1 {
-                                oversized.push(slice[0]);
-                                continue;
-                            }
-                            let mid = slice.len() / 2;
-                            let (left, right) = slice.split_at(mid);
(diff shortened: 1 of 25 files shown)


linera-bridge/src/monitor/linera.rs
```diff
@@ -78,86 +83,235 @@ pub(crate) async fn process_pending_burns<E: linera_core::environment::Environme
...
+async fn mark_oversized_failed(
+    monitor: &RwLock<MonitorState>,
+    height: BlockHeight,
+    tx_index: u32,
+    oversized: &[u32],
+) {
+    if oversized.is_empty() {
+        return;
+    }
+    let to_fail: Vec<u32> = {
```
afck: (Type annotation should go on `collect` below.)

linera-bridge/src/monitor/linera.rs
```diff
@@ -78,86 +83,235 @@ pub(crate) async fn process_pending_burns<E: linera_core::environment::Environme
...
+            .iter()
+            .filter_map(|&pos| state.event_index_for_pos(height, tx_index, pos))
+            .collect()
+    };
+    let mut state = monitor.write().await;
+    for ei in to_fail {
+        tracing::error!(
+            ?height,
+            tx_index,
+            event_index = ei,
```
afck: (Why not `for event_index in to_fail`?)

linera-bridge/src/monitor/linera.rs
```diff
@@ -78,86 +83,235 @@ pub(crate) async fn process_pending_burns<E: linera_core::environment::Environme
...
             .await
-            .mark_burn_retried(credit_height, event_index);
+        {
+            tracing::warn!(
+                tx_index,
+                ?events_chunk,
+                ?error,
+                "processBurns submission failed"
+            );
+            let to_bump: Vec<u32> = {
```
afck: (Type annotation)

linera-bridge/tests/e2e/tests/multiple_burns_same_block.rs
```diff
@@ -274,13 +272,13 @@ async fn relayer_processes_every_burn_in_one_block() -> anyhow::Result<()> {
 
     // Each occurrence in `recipients` is one expected transfer. A
     // recipient listed twice should accumulate two transfers, etc.
-    let mut expected_per_recipient: std::collections::BTreeMap<
-        alloy::primitives::Address,
-        U256,
-    > = std::collections::BTreeMap::new();
+    let mut expected_per_recipient: std::collections::BTreeMap<alloy::primitives::Address, U256> =
```
afck: (Type annotation could move to next line; avoids `BTreeMap` duplication.)

linera-bridge/src/relay/linera.rs
```diff
@@ -142,25 +142,27 @@ impl<E: linera_core::environment::Environment> Clone for LineraClient<E> {
...
+/// - `event_index`: `Event.index` — the stream index, used as the on-chain dedup key
 pub(crate) fn find_burn_events(
     events: &[Vec<Event>],
     fungible_app_id: ApplicationId,
-) -> Vec<(u32, wrapped_fungible::BurnEvent)> {
+) -> Vec<(u32, u32, u32, wrapped_fungible::BurnEvent)> {
     let mut result = Vec::new();
-    for tx_events in events {
-        for event in tx_events {
+    for (i, tx_events) in events.iter().enumerate() {
```
afck: `(0u32..).zip(&events)` would avoid the type cast below.

linera-bridge/src/monitor/mod.rs
```diff
@@ -98,6 +114,42 @@ impl<T: Clone> Tracked<T> {
...
+
+impl Ord for PendingBurnsAtHeight {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        self.height.cmp(&other.height)
+    }
+}
+
+impl PartialOrd for PendingBurnsAtHeight {
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        Some(self.cmp(other))
```
afck, line 157: (That doesn't really satisfy the `PartialOrd` properties documented in the standard library: `Eq` should match whether this returns `Some(Equal)`: https://doc.rust-lang.org/std/cmp/trait.PartialOrd.html)

afck (commented):
(No blockers so far, but I might need a walkthrough in a meeting to approve. 😅)

PR #6297, opened 2026-05-13
Add Scylla Wasm UDFs to decode RootKey partition keys in cqlsh

ndr-ds:
## Motivation

`linera-storage-decode-key` (#6263) is great for offline decoding of `RootKey` partition keys, but you still have to copy hex out of `cqlsh`, pipe it through, and copy results back. Moving the decoder into ScyllaDB itself as a Wasm UDF eliminates the pipe and unlocks queries like:

```sql
SELECT decode_root_key(partition_key) AS what, partition_size
FROM system.large_partitions
WHERE keyspace_name = 'linera'
ORDER BY partition_size DESC
LIMIT 20;
```

and

```sql
SELECT decode_root_key(root_key), length(k), length(v)
FROM linera."ns_42"
WHERE root_key IN (
  encode_root_key('ChainState',       '7a3f...'),
  encode_root_key('BlockByHeight',    '7a3f...'),
  encode_root_key('Event',            '7a3f...'),
  encode_root_key('EventBlockHeight', '7a3f...')
);
```

## Proposal

New crate `linera-storage/udf/` (workspace member, not in `default-members`) targeting `wasm32-wasip1`. Two `#[scylla_udf::export_udf]` functions: `decode_root_key(blob) -> text` and `encode_root_key(text, text) -> blob`. Both transparently handle the leading `0x00` tag byte that `linera-views::backends::scylla_db::get_big_root_key` prepends.

`RootKey` is mirrored locally because compiling `linera-storage` to `wasm32-wasip1` would drag in the validator dependency tree. A drift-detection test (`linera-storage/tests/root_key_drift.rs`) asserts the mirrored enum BCS-encodes identically to the canonical one — it will fail loudly if `RootKey` changes here without the mirror being updated.

Build script at `linera-storage/udf/build.sh` produces the optimized `.wasm` (~145 KB) and the `.wat` ready to embed in a CQL `CREATE FUNCTION` statement. Full registration workflow documented in `linera-storage/udf/README.md`.

Includes a custom `[profile.wasm-udf]` in the workspace `Cargo.toml` (inherits release; strip; LTO; opt-level z; panic abort) because WAT source size matters: it is embedded as a string literal in the CQL statement.

Will be backported to `testnet_conway` (which has a different `RootKey` schema) as a separate PR.

## Test Plan

CI. Additionally locally:

- `cargo check -p linera-storage-udf`
- `cargo clippy -p linera-storage-udf --all-targets -- -D warnings`
- `cargo clippy -p linera-storage --all-targets -- -D warnings`
- `cargo test -p linera-storage-udf` — 11/11 passing
- `cargo test -p linera-storage --test root_key_drift` — 1/1 passing
- `RUSTFLAGS=... cargo build -p linera-storage-udf --target wasm32-wasip1 --profile wasm-udf` produces a 148 KB binary
- `wasm2wat` produces 55,587 lines of WAT ready for CQL embedding

End-to-end registration against a real Scylla instance is deferred until the operator workflow is in place; ScyllaDB Wasm UDFs are still experimental and require `udf` in `experimental_features` in `scylla.yaml`.

## Release Plan

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -129,7 +129,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4195a29a4b87137b2bb02105e746102873bc03561805cf45c0e510c961f160e6"
 dependencies = [
  "alloy-primitives",
- "num_enum",
+ "num_enum 0.7.4",
  "strum 0.27.2",
 ]
 
@@ -1811,6 +1811,17 @@ checksum = "85b6598a2f5d564fb7855dc6b06fd1c38cff5a72bd8b863a4d021938497b440a"
 dependencies = [
  "serde",
  "thiserror 1.0.69",
+]
+
+[[package]]
+name = "bigdecimal"
+version = "0.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d1e50562e37200edf7c6c43e54a08e64a5553bfb59d9c297d5572512aa517256"
+dependencies = [
+ "num-bigint 0.3.3",
+ "num-integer",
+ "num-traits",
 ]
 
 [[package]]
@@ -6291,6 +6302,7 @@ dependencies = [
  "linera-chain",
  "linera-execution",
  "linera-storage",
+ "linera-storage-udf",
  "linera-views",
  "papaya",
  "prometheus",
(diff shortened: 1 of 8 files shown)


ndr-ds:
hmm this requires experimental features on all Scylla nodes, which sounds a bit scary. Maybe let's park this for now

PR #6300, opened 2026-05-13
monitoring: add CPU steal time panel to validator General dashboard

ndr-ds:
## Motivation

CPU steal time is a hypervisor-level signal invisible to cgroup metrics like CFS throttling. It tells us whether the node is being starved by the hypervisor, which is distinct from our own CPU pressure.

## Proposal

Adds a "CPU Steal Time (per validator cluster)" panel to the **Cluster Info** section of the General validator dashboard, half-width (w=12) alongside the existing "Shards FS written bytes per second" panel. Queries \`node_cpu_seconds_total{mode="steal", cluster=~"gke-testnet-conway-validator-.*"}\` averaged per cluster.

The panel was verified live in Grafana (General dashboard version 33) before committing the JSON.

## Test Plan

CI.

## Release Plan


ndr-ds:
Wrong head branch — re-opened as a correct PR.

PR #6309, opened 2026-05-14
linera-sdk: crate trait and derive macro `StableEnum`

ma2bd:
## Motivation

We want enum variant tags on the wire to be stable under variant reordering, similar to Solidity's 4-byte function selectors.
* Today, BCS serializes the serde variant *index* (0, 1, 2, …, ULEB128-encoded), which changes whenever variants are reordered.
* This is especially a problem when we want an ABI to extend several existing ones. (Typically different flavors of `fungible`). After this PR, we could occasionally merge enums with solutions like https://github.com/rafaeltheraven/extendable-data.

## Proposal

Add a `#[derive(StableEnum)]` macro in `linera-sdk-derive` that emits, in one shot:

* `Serialize` / `Deserialize` impls calling `serialize_*_variant(name, tag, …)` / matching `tag` from `EnumAccess`, where the tag is derived at compile time from the variant name:
  ```
  tag = (be_u32(Keccak-256(name)[..4]) & 0x07FF_FFFF) | 0x0800_0000
  ```
  Masking the top 5 bits to `00001` constrains the tag to `[2^27, 2^28)`, which is exactly the range whose ULEB128 encoding is **always 4 bytes** — so every variant tag on the wire takes exactly 4 bytes.
* A `StableEnumTrace` impl exposing the `(name, tag)` pairs and a `trace_all_variants` method that drives `serde_reflection::Tracer` without caller-supplied samples. (The default `Tracer::trace_type` cannot terminate on non-contiguous u32 indices like ours, so we provide our own tracing entry point.)

Companion additions in `linera-sdk::formats`:

* `trait StableEnumTrace` (with `const STABLE_VARIANTS: &[(&str, u32)]` and `fn trace_all_variants`).
* `trait StableEnum: StableEnumTrace + Serialize + DeserializeOwned {}` blanket-implemented — usable as a single bound.
* `TracerExt::trace_stable_enum_type::<T>(&Samples) -> Result<Format>` extension method on `Tracer`.
* `StableEnumInCrate` variant of the derive (paths via `crate::…` instead of `::linera_sdk::…`) for use inside `linera-sdk` itself.

The `StableEnumTrace` impl is `#[cfg(not(target_arch = \"wasm32\"))]` so applications keep building for wasm without `serde-reflection`.

The `counter` example is migrated as a reference; other examples still use plain `#[derive(Serialize, Deserialize)]` and can be migrated in follow-up PRs.

Note: Importantly, the variant index used by Serde is different from the "discriminant" used by Rust in memory. Assigning new tag values using `repr(u32)` and the Rust syntax only changes the discriminant.

## Test Plan

* New unit test `formats::tests::stable_enum_round_trip` in `linera-sdk` covering:
  * BCS roundtrip for unit / newtype / tuple / struct variants
  * Every variant tag is exactly 4 ULEB128 bytes on the wire (continuation-bit pattern asserted explicitly)
  * Unknown tags are rejected on deserialization
  * `STABLE_VARIANTS` constant matches what BCS emits
  * `trace_stable_enum_type` populates the registry with hashed keys
  * End-to-end `bcs_to_json` against the reflected registry
* `counter` snapshot test updated: `CounterOperation::Increment` is now registered under tag `186772159` (`0x0b21eabf`) instead of index `0`.
* All other counter tests (`single_chain`, contract/service unit tests) continue to pass.

## Release Plan

This should backported to `testnet_conway`.


linera-sdk-derive/src/stable_enum.rs
```diff
@@ -0,0 +1,454 @@
...
+    /// dependency just to use `#[derive(StableEnum)]`.
+    fn serde_reflection_path(&self) -> TokenStream2 {
+        match self {
+            CrateRoot::LineraSdk => quote! { ::linera_sdk::formats::__private::serde_reflection },
+            CrateRoot::Crate => quote! { crate::formats::__private::serde_reflection },
+        }
+    }
+}
+
+/// Compute the stable tag for a variant name.
```
afck:
```suggestion
/// Computes the stable tag for a variant name.
```

Also below.

linera-sdk-derive/src/stable_enum.rs
```diff
@@ -0,0 +1,454 @@
...
+                quote! {
+                    #tag => {
+                        let __inner = #serde::de::VariantAccess::newtype_variant::<#ty>(__variant)?;
+                        ::core::result::Result::Ok(#enum_ident::#variant_ident(__inner))
+                    }
+                }
+            }
+            Fields::Unnamed(fields) => {
+                let len = fields.unnamed.len();
+                let types: Vec<_> = fields.unnamed.iter().map(|f| &f.ty).collect();
```
afck:
```suggestion
                let types = fields.unnamed.iter().map(|f| &f.ty).collect::<Vec<_>>();
```

Also below.

afck (approved):

examples/counter/src/lib.rs
```diff
@@ -47,7 +49,7 @@ pub mod formats {
             let samples = Samples::new();
 
             // Trace the ABI types
-            let (operation, _) = tracer.trace_type::<CounterOperation>(&samples)?;
+            let operation = tracer.trace_stable_enum_type::<CounterOperation>(&samples)?;
```
deuszx, line 52: Why do we need to change this? Can't it work transparently with `StableEnum`s ?

linera-sdk-derive/src/lib.rs
```diff
@@ -26,6 +27,34 @@ pub fn derive_mutation_root_in_crate(input: TokenStream) -> TokenStream {
     generate_mutation_root_code(input, "crate").into()
 }
 
+/// Derive `linera_sdk::formats::StableEnum` for an `enum`. Expands to:
+///
+/// * `serde::Serialize` / `serde::Deserialize` impls in which the variant tag
+///   is the first 4 bytes of `Keccak-256(variant_name)` (read big-endian as
```
deuszx, line 33: what's the reasoning behind usage of keccak here? we won't be trying to re-generate the hash on EVM chain, right? if so, using a faster function should be better.

examples/counter/src/lib.rs
ma2bd: `trace_type` assumes that the variant indices are contiguous.

linera-sdk-derive/src/lib.rs
ma2bd: good question. I was imitating EVM

deuszx:
Changing the type signature of the function should - in my opinion - change its function discriminant as well. This is basically an API change and should affect the selector. Similarly for events and other types.

When we get to "contract API extension" – how do you see this working when two different "extensions" have the same event name?

linera-sdk-derive/src/stable_enum.rs
```diff
@@ -0,0 +1,454 @@
...
+}
+
+/// Computes the stable tag for a variant name.
+fn compute_tag(variant_name: &str) -> u32 {
+    let hash = Keccak256::digest(variant_name.as_bytes());
+    let val = u32::from_be_bytes([hash[0], hash[1], hash[2], hash[3]]);
+    (val & 0x07FF_FFFF) | 0x0800_0000
+}
+
+/// Compute all variant tags, returning an error on a (vanishingly unlikely) collision.
```
afck:
```suggestion
/// Computes all variant tags, returning an error on a (vanishingly unlikely) collision.
```

linera-sdk-derive/src/stable_enum.rs
```diff
@@ -0,0 +1,454 @@
...
+    if !input.generics.params.is_empty() || input.generics.where_clause.is_some() {
+        return Err(Error::new(
+            input.generics.span(),
+            "#[derive(StableEnum)] does not yet support generic enums",
+        ));
+    }
+    Ok(())
+}
+
+/// Emit the combined `Serialize` + `Deserialize` + `StableEnumTrace` impls.
```
afck:
```suggestion
/// Emits the combined `Serialize` + `Deserialize` + `StableEnumTrace` impls.
```

linera-sdk-derive/src/stable_enum.rs
```diff
@@ -0,0 +1,454 @@
...
+                        #(
+                            #serde::ser::SerializeTupleVariant::serialize_field(&mut __tv, #idents)?;
+                        )*
+                        #serde::ser::SerializeTupleVariant::end(__tv)
+                    }
+                }
+            }
+            Fields::Named(fields) => {
+                let len = fields.named.len();
+                let field_idents: Vec<&Ident> = fields
```
afck, line 149: There are more type annotations, also below.

ma2bd: At this point, we should probably give up or create a new clippy rule

ma2bd: there is this one actually: https://rust-lang.github.io/rust-clippy/rust-1.89.0/index.html#redundant_type_annotations

PR #6314, opened 2026-05-15
ABIs: derive `StableEnum` for all ContractAbi Operation/Response enums

ma2bd:
## Motivation

Follow-up to #6309: now that `#[derive(StableEnum)]` exists, switch every `ContractAbi` operation/response enum in the repo to use it. This locks each variant tag to a Keccak-derived 4-byte ULEB128 value, decoupling the wire format from declaration order so reorderings/insertions don't silently break consumers.

## Proposal

Apply `#[derive(StableEnum)]` (or `StableEnumInCrate` for code inside `linera-sdk`) to every `Operation` and enum-typed `Response`:

- **Example contracts:** `amm`, `call-evm-counter`, `counter-no-graphql`, `crowd-funding`, `ethereum-tracker`, `evm-bridge` (via `linera-bridge`), `gen-nft`, `hex-game` (both `Operation` and the `HexOutcome` response), `matching-engine`, `non-fungible`, `rfq`, `social`, `task-processor`, `how-to/perform-http-requests`.
- **In-crate `linera-sdk` ABIs:** `fungible::FungibleOperation` + `FungibleResponse`, `controller::Operation`, `wrapped_fungible::WrappedFungibleOperation`.
- **`linera-sdk` fixtures:** `contract-call`, `cost-tracking`, `publish-read-data-blob`, `event-emitter`, `create-and-call`, `time-expiry`, `event-subscriber`.

Other changes pulled in:

- `linera-bridge` gets a direct (non-optional) `linera-sdk` dependency so `BridgeOperation` can use the derive. This was already a transitive dep via `wrapped-fungible` in the `relay` feature.
- Each `formats` module now traces the stable enum via `tracer.trace_stable_enum_type::<T>(&samples)?` (importing `TracerExt`), since `Tracer::trace_type` cannot enumerate the non-contiguous Keccak tags.
- Removed unused `scalar!(Operation)` calls in `amm` and `rfq`. `scalar!` registers an async-graphql `ScalarType` impl backed by `serde_json`, but StableEnum's serde impls use a numeric variant tag that serde_json cannot deserialize from a string variant name. Nothing in the codebase actually invokes them — `GraphQLMutationRoot` exposes individual fields rather than the whole enum — so the impls are dead code that would silently break if exercised.

Skipped because they have no enum to convert: `evm.rs` (Operation = `Vec<u8>`), `llm` (Operation = `()`), `meta-counter` (Operation is a struct), `track-instantiation` (Operation = `()`).

## Test Plan

- `cargo check --workspace` and `cargo check` on the examples workspace.
- All 18 example `format` snapshot tests regenerated with the new Keccak tags and pass.
- `linera-sdk`'s `formats::tests` (including `stable_enum_round_trip`) pass.
- `cargo +nightly fmt --check` and `cargo clippy --workspace --all-targets` are clean.

## Release Plan

Could be backported to `testnet_conway`

## Links

- Stacked on #6309


linera-service/tests/fixtures/evm_call_wasm_example_counter.sol
```diff
@@ -2,12 +2,13 @@
 // SPDX-License-Identifier: Apache-2.0
 pragma solidity ^0.8.0;
 
-// The BCS conversion code used below has been generated
-// by using the solidity framework of the crate
-// serde-reflection.
+// The serde-json conversion code has been generated by hand.
```
afck, line 5: ("generated by hand" 😆)
(Nothing wrong with it; I just find the formulation funny.)

afck (approved):

linera-service/tests/fixtures/evm_call_wasm_example_counter.sol
ma2bd: It is!

PR #6321, opened 2026-05-15
linera-metrics: expose stable tokio runtime metrics on /metrics

ndr-ds:
## Motivation

Operators need to answer "is the async runtime saturated?" without guessing. Currently there is no visibility into tokio scheduler health — worker utilisation, idle vs. busy ratios, queue depth, or task count.

## Proposal

Add a `TokioRuntimeCollector` that implements `prometheus::core::Collector` and pulls six metrics from tokio's stable `RuntimeMetrics` API on every `/metrics` scrape (pull-on-demand; no background task):

| Metric | Type | Description |
|--------|------|-------------|
| `linera_tokio_workers` | gauge | Number of worker threads |
| `linera_tokio_alive_tasks` | gauge | Number of alive tasks |
| `linera_tokio_global_queue_depth` | gauge | Global injection queue depth |
| `linera_tokio_worker_busy_seconds_total{worker}` | counter | Cumulative busy time per worker |
| `linera_tokio_worker_parks_total{worker}` | counter | Park count per worker |
| `linera_tokio_worker_park_unparks_total{worker}` | counter | Combined park+unpark count per worker (odd → currently parked) |

The three per-worker counters are gated on `#[cfg(target_has_atomic = "64")]` matching tokio's own gate — **no `tokio_unstable` flag required**.

`TokioRuntimeCollector::register()` is called once (via `OnceLock`) at the top of `start_metrics()`, so every binary that already exposes `/metrics` — `linera-server`, `linera-proxy`, and any future server — gets these for free with no per-binary changes.

The key derived metric for dashboards is:

```
rate(linera_tokio_worker_busy_seconds_total[1m])
```

A value near 1.0 on any worker means that thread is saturated.

## Test Plan

Built with `--features metrics` and ran a local single-validator network:

```
cargo build -p linera-service --bin linera-server --bin linera-proxy --bin linera --features metrics
linera net up --validators 1 --shards 1
curl -s http://127.0.0.1:11001/metrics | grep linera_tokio  # shard
curl -s http://127.0.0.1:12001/metrics | grep linera_tokio  # proxy
```

All 6 metric families present with real values on both endpoints. The collector contains no logic beyond atomic reads and proto construction; there is no unit-testable behaviour.

## Release Plan

## Links

### DIFF
--- a/linera-metrics/src/runtime_metrics.rs
+++ b/linera-metrics/src/runtime_metrics.rs
@@ -1,0 +1,167 @@
+use std::{collections::HashMap, sync::OnceLock};
+
+use prometheus::{
+    core::{Collector, Desc},
+    proto::{Counter, Gauge, LabelPair, Metric, MetricFamily, MetricType},
+};
+use tokio::runtime::Handle;
+
+pub(crate) struct TokioRuntimeCollector {
+    handle: Handle,
+    descs: Vec<Desc>,
+}
+
+const IDX_WORKERS: usize = 0;
+const IDX_ALIVE_TASKS: usize = 1;
+const IDX_GLOBAL_QUEUE: usize = 2;
+const IDX_BUSY_SECONDS: usize = 3;
+const IDX_PARKS: usize = 4;
+const IDX_PARK_UNPARKS: usize = 5;
+
+impl TokioRuntimeCollector {
+    fn new(handle: Handle) -> Self {
+        let descs = vec![
+            Desc::new(
+                "linera_tokio_workers".into(),
+                "Number of worker threads in the tokio runtime.".into(),
+                vec![],
+                HashMap::new(),
+            )
+            .expect("static metric descriptor is always valid"),
+            Desc::new(
+                "linera_tokio_alive_tasks".into(),
+                "Number of tasks currently alive in the tokio runtime.".into(),
+                vec![],
+                HashMap::new(),
+            )
+            .expect("static metric descriptor is always valid"),
+            Desc::new(
+                "linera_tokio_global_queue_depth".into(),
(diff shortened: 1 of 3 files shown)


linera-metrics/src/runtime_metrics.rs
```diff
@@ -0,0 +1,170 @@
...
+    handle: Handle,
+    descs: Vec<Desc>,
+}
+
+const IDX_WORKERS: usize = 0;
+const IDX_ALIVE_TASKS: usize = 1;
+const IDX_GLOBAL_QUEUE: usize = 2;
+const IDX_BUSY_SECONDS: usize = 3;
+const IDX_PARKS: usize = 4;
+const IDX_PARK_UNPARKS: usize = 5;
```
ma2bd: what is this? should we import the constants from an external library?

ndr-ds: Sorry, this is pretty ugly indeed, I'll try to improve it

linera-metrics/src/runtime_metrics.rs
```diff
@@ -0,0 +1,171 @@
...
+    fn new(handle: Handle) -> Self {
+        Self {
+            handle,
+            workers: Desc::new(
+                "linera_tokio_workers".into(),
+                "Number of worker threads in the tokio runtime.".into(),
+                vec![],
+                HashMap::new(),
+            )
+            .expect("static metric descriptor is always valid"),
```
ma2bd: nit: maybe move the expressions before the `Self { .. } ` with `let workers = ... ;` etc

ma2bd (approved):

PR #6328, opened 2026-05-18
Only preprocess sender blocks, even if contiguous

afck:
## Motivation

Currently chain workers fully execute new blocks whenever they can, i.e. if they are the next one to be executed according to the chain tip state. This is wasteful for e.g. sender chains.

## Proposal

Make the mode explicit, and on the client side specify explicitly whether to fully execute or only preprocess.

Validators still execute whenever they can, since they are expected to execute every chain anyway.

## Test Plan

CI

## Release Plan

- Could be backported to `testnet_conway`.

## Links


linera-core/src/client/mod.rs
```diff
@@ -872,27 +872,38 @@ impl<Env: Environment> Client<Env> {
...
             }
+            // This helper is only called when downloading our own chain's history
+            // (sender-chain certificates flow through `receive_sender_certificate`),
+            // so we know each certificate is contiguous with the local tip.
             let response = self
-                .handle_certificate_with_retry(&certificate, remote_nodes)
+                .handle_certificate_with_retry(
+                    &certificate,
+                    remote_nodes,
+                    ProcessConfirmedBlockMode::Execute,
```
deuszx: Do we have to thread the mode manually? This should be part of the client's wallet ?

linera-core/src/client/mod.rs
```diff
@@ -1205,8 +1234,12 @@ impl<Env: Environment> Client<Env> {
         } else {
             self.validator_nodes().await?
         };
-        self.handle_certificate_with_retry(&certificate, &nodes)
-            .await?;
+        self.handle_certificate_with_retry(
+            &certificate,
+            &nodes,
+            ProcessConfirmedBlockMode::Preprocess,
```
deuszx: We should be able to tell whether it's the sender or not based on the `chain_id` in the `certificate` and compare it with the client's wallet. I think.

linera-core/src/chain_worker/state.rs
```diff
@@ -137,6 +137,23 @@ pub enum BlockOutcome {
...
+/// current tip.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum ProcessConfirmedBlockMode {
+    /// Execute if contiguous; preprocess if there is a gap. This is what validators
+    /// always use, and is also the default for paths that don't care.
+    Auto,
+    /// Execute the block. Fails with [`WorkerError::UnexpectedBlockHeight`] if the
+    /// block is not contiguous with the current tip. Used by the client for blocks
+    /// of its own chain, where contiguity is a correctness invariant.
+    Execute,
```
deuszx, line 152: I don't understand why we need both `Auto` and `Execute`. IMO the `Execute` should behave as `Auto` does – if it's contiguous -> execute.

afck: My thinking was that the _client_ usually already knows whether this is for a sender chain (don't execute) or for our own chain: In the latter case, we know we want to execute the block, so failing to do so should be an error.

But maybe you're right and this is unnecessarily complicated. Happy to change it, if you prefer.

deuszx: I'm just not convinced it's necessary. We didn't need it before so I don't see how this PR should need it either (i.e. the "expectations" are still the same IMO).

linera-core/src/local_node.rs
```diff
@@ -128,6 +129,24 @@ where
...
+    pub async fn handle_confirmed_certificate(
+        &self,
+        certificate: ConfirmedBlockCertificate,
+        mode: ProcessConfirmedBlockMode,
+        notifier: &impl Notifier,
+    ) -> Result<ChainInfoResponse, LocalNodeError> {
+        Ok(Box::pin(
+            self.node
+                .state
+                .fully_handle_confirmed_certificate_with_notifications(certificate, mode, notifier),
```
deuszx, line 144: If the `mod` is `Preprocess`-only then we won't be able to create all notifications? i.e. we definitely shouldn't be sending `BlockExecuted`.

afck: Of course: The worker decides which notifications to send, based on what actually happened. (`BlockExecuted` seems to be sent from within the _client_, though?)

linera-core/src/worker.rs
```diff
@@ -1526,12 +1561,11 @@ where
...
             Either::Left(confirmed) => {
-                Box::pin(
-                    self.handle_confirmed_certificate(
-                        confirmed,
-                        notify_when_messages_are_delivered,
-                    ),
-                )
+                Box::pin(self.handle_confirmed_certificate(
+                    confirmed,
+                    ProcessConfirmedBlockMode::Execute,
```
deuszx: We never call `handle_lite_certificate` for chains that we're only preprocessing?

afck: I think we only send lite certificates _to_ the validators, but always download full certificates _from_ them.

linera-core/src/chain_worker/state.rs
```diff
@@ -902,84 +915,82 @@ where
...
-        //   * Otherwise, only preprocess: update outboxes and event streams and return,
-        //     so this block becomes available to clients via cross-chain syncing.
+        // Decide whether to fully execute or only preprocess this block. We
+        // preprocess when the caller explicitly asked for `Preprocess`, or when
+        // there's a gap that we can't bridge. A checkpoint block lets us bridge
+        // a gap in `Execute` mode by installing its execution-state snapshot;
+        // in `Preprocess` mode we deliberately skip the restore — the caller
+        // doesn't want this chain's state advanced.
+        let preprocess_only = matches!(mode, ProcessConfirmedBlockMode::Preprocess)
+            || (tip.next_block_height < height && !block.starts_with_checkpoint());
```
deuszx: If `block` starts with the checkpoint, shouldn't we always execute it? regardless of the result of the `tip.next_block_height < height` comparison?

afck: Yes, that's what we're doing: If `block.starts_with_checkpoint()` is `true`, then `(tip.next_block_height < height && !block.starts_with_checkpoint())` is `false`, so `preprocess_only` is `matches!(mode, ProcessConfirmedBlockMode::Preprocess)`.

deuszx: Hmm, but if someone sets the `mode` to `Preprocess` and we have the block that has the checkpoint then it will be `preprocess_only == true` as well? Is that correct?

afck: Yes! If we don't want to execute a chain, we don't want to restore the state from the checkpoint. Just because a sender block happens to contain a checkpoint that shouldn't make us do unnecessary work.

deuszx: OK, sounds good. I thought for a moment that checkpoints should always be restored but you even had it in the comment that's not the case.

linera-core/src/chain_worker/state.rs
```diff
@@ -902,84 +915,82 @@ where
...
+            self.register_delivery_notifier(height, &actions, notify_when_messages_are_delivered)
+                .await;
+            return Ok((
+                self.chain_info_response().await?,
+                actions,
+                BlockOutcome::Preprocessed,
+            ));
+        }
+
+        // If we're here with a gap, the block must be a checkpoint: install its
```
deuszx: I don't like the code that requires the reader to know the whole context. We're 130 lines into the function and we have this revelation _"oh, if we're here then the block must be the checkpoint"_. It's very fragile, coding like this. This function is now 300+ lines long and very imperative/low-level. Can't we start the `process_confirmed_block` by checking which mode we're in (preprocess or execute) and then call separate functions for those?

afck: I split the function into three in https://github.com/linera-io/linera-protocol/pull/6328/commits/9798d8b20470a8026495f4663909d74ee2bfced6, but it's not as simple as one function per mode, because that would duplicate a lot of code.

linera-core/src/chain_worker/state.rs
ma2bd: @afck @deuszx The point of this PR was to be explicit and avoid surprises. I kinda like the 3 modes. Afaict validators would use only `Auto` and clients (aka sparse nodes) would use only `Preprocess` and `Execute`.

deuszx: The three modes makes the code more confusing as the two are overlapping and require specific checks. Since the change to two modes the code diff is smaller and simpler..

deuszx: For the validators it's always Execute, isn't it? Ie they try to execute when the chain is contiguous but still accept existence of the sparse chains. So I'd go even further and say that validators don't even need it at all but since the same code is used both in clients and validators we end up having to handles scenarios that don't really exist.

ma2bd: No, validators are auto.

> they try to execute when the chain is contiguous but still accept existence of the sparse chains.

That's the definition of Auto

deuszx: That's the definition of Execute because you can't execute a single block out of order/in isolation.

ma2bd: Validators must accept sparse blocks as well for liveness reasons. So the strict Execute mode is not enough there

deuszx: But they still accept it – this mode is about what to do when we get the continuous chain of blocks. Validators should always execute, some clients don't have to (sender chains).

afck: If I go back to the three states, `Execute` will fail if not contiguous (or checkpoint), `Preprocess` will never execute, and `Auto` will never fail and execute if contiguous (or checkpoint).

afck: (But it's not straighforward, because then we can't, on the client side, determine the mode purely based on the chain's `ListeningMode` anymore.)

afck: Done.

ma2bd (approved):

PR #6330, opened 2026-05-18
Add `linera_cache::Arc<T>` newtype enforcing one-allocation-per-content invariant

ndr-ds:
## Motivation

Follow-up to #6147 (reviewed by @ma2bd). The previous PR pushed `Arc` deeper through the certificate/block paths but left the door open for callers to bypass the dedup guarantee by constructing a fresh `Arc::new(value)` and inserting it — the type alone couldn't enforce "this came from the cache".

## Proposal

Introduce `linera_cache::Arc<T>`: a `#[repr(transparent)]` newtype over `std::sync::Arc<T>` with no public constructor. The only way to obtain one is through `ValueCache::insert`, `ValueCache::insert_hashed`, or `ValueCache::get`. This makes the dedup invariant structural: if you hold a `CacheArc<T>`, it provably came from the cache.

API surface:
- `into_std(self) -> std::sync::Arc<T>` — for crossing trait/network/serde boundaries
- `as_std(&self) -> &std::sync::Arc<T>` — for read-only boundary crossings
- `unwrap_or_clone(this: Self) -> T` — mirrors `Arc::unwrap_or_clone`
- `Deref<Target = T>`, `Clone`, `From<CacheArc<T>> for std::sync::Arc<T>`

Re-exported from `linera-storage` as `pub use linera_cache::Arc` (aliased as `CacheArc` inside consuming crates).

## Test Plan

CI — existing tests exercise the cache paths and the type changes are mechanical (boundary conversions via `into_std()`/`as_std()`).

## Release Plan

## Links

- #6147 — the PR this follows up on

### DIFF
--- a/linera-cache/src/arc.rs
+++ b/linera-cache/src/arc.rs
@@ -1,0 +1,108 @@
+// Copyright (c) Zefchain Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! A provenance-tracking wrapper around [`std::sync::Arc`].
+//!
+//! [`Arc<T>`] is identical to [`std::sync::Arc<T>`] at runtime but has no
+//! public constructor. The only way to obtain one is through
+//! [`ValueCache::insert`], [`ValueCache::insert_hashed`], or
+//! [`ValueCache::get`]. This makes the "one allocation per content" invariant
+//! structurally enforced: callers cannot bypass the cache by calling
+//! `Arc::new` directly.
+
+use std::{
+    fmt,
+    hash::{Hash, Hasher},
+    ops::Deref,
+    sync::Arc as StdArc,
+};
+
+/// A reference-counted pointer that can only be constructed through a
+/// [`crate::ValueCache`].
+///
+/// `Arc<T>` is `#[repr(transparent)]` over [`std::sync::Arc<T>`] and
+/// implements `Deref<Target = T>`, `Clone`, `Debug`, `Display`, `PartialEq`,
+/// `Eq`, and `Hash` identically to [`std::sync::Arc<T>`].
+///
+/// Use [`Arc::into_std`] or [`Arc::as_std`] to interoperate with APIs that
+/// require [`std::sync::Arc<T>`] explicitly.
+#[repr(transparent)]
+pub struct Arc<T>(pub(crate) StdArc<T>);
+
+impl<T> Arc<T> {
+    /// Returns a reference to the underlying [`std::sync::Arc<T>`].
+    pub fn as_std(&self) -> &StdArc<T> {
+        &self.0
+    }
+
+    /// Converts into the underlying [`std::sync::Arc<T>`].
+    pub fn into_std(self) -> StdArc<T> {
+        self.0
+    }
+
+    /// Unwraps the inner value if this is the only strong reference;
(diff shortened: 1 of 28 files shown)


linera-cache/src/arc.rs
```diff
@@ -0,0 +1,108 @@
...
+/// A reference-counted pointer that can only be constructed through a
+/// [`crate::ValueCache`].
+///
+/// `Arc<T>` is `#[repr(transparent)]` over [`std::sync::Arc<T>`] and
+/// implements `Deref<Target = T>`, `Clone`, `Debug`, `Display`, `PartialEq`,
+/// `Eq`, and `Hash` identically to [`std::sync::Arc<T>`].
+///
+/// Use [`Arc::into_std`] or [`Arc::as_std`] to interoperate with APIs that
+/// require [`std::sync::Arc<T>`] explicitly.
+#[repr(transparent)]
```
ma2bd: To be clear, this line only matters for `unsafe` transmute operations, which we don't use, correct?

ma2bd (approved):
nice

linera-cache/src/arc.rs
ndr-ds: Good point, let me remove

linera-cache/src/value_cache.rs
```diff
@@ -67,18 +67,13 @@ where
...
     /// reused and the new value is dropped.
-    pub fn insert(&self, key: &K, value: V) -> Arc<V> {
-        self.dedup_insert(key, Arc::new(value))
-    }
-
-    /// Inserts a pre-wrapped `Arc<V>` into the cache, returning the canonical `Arc`.
-    pub fn insert_arc(&self, key: &K, value: Arc<V>) -> Arc<V> {
-        self.dedup_insert(key, value)
+    pub fn insert(&self, key: &K, value: V) -> crate::Arc<V> {
+        crate::Arc(self.dedup_insert(key, Arc::new(value)))
```
afck: (Wouldn't it be easier to keep `crate::Arc`s in the map, rather than (std) `Arc`s? Then we wouldn't have to explicitly create `crate::Arc`s everywhere.)

linera-core/src/client/chain_client/mod.rs
```diff
@@ -3220,10 +3222,14 @@ impl<Env: Environment> ChainClient<Env> {
                 .await?
                 .into_iter()
                 .flatten()
+                .map(|b| b.into_std())
                 .collect();
             remote_node.upload_blobs(missing_blobs).await?;
             remote_node
-                .handle_confirmed_certificate(certificate, CrossChainMessageDelivery::NonBlocking)
+                .handle_confirmed_certificate(
+                    certificate.into_std(),
```
afck: Wouldn't it be safer (i.e. less prone to accidental duplicates) if `handle_confirmed_certificate` also required a `CacheArc`?

afck (approved):

deuszx (approved):
It'd just drop the `*Arc` suffix – maybe just call it `Cached` or `CacheValue`. No need to mention the `Arc` part.


Also, I couldn't find the place where we'd violate the intended usage of the `Arc` obtained from the cache (as you write in the PR description). Can you point it out? Or is this just a precaution measure?

Issue #912, opened 2023-07-24
Setup basic A/B testing capabilities


andresilva91:

ndr-ds:
Initial version of this is https://github.com/linera-io/linera-protocol/issues/2934. Fake traffic might not be necessary if we just run multiple deterministic tests in the same order or benchmarks

PR #6332, opened 2026-05-18
Add cross-key worker boot test (Phase 0a, #6244)

### LINKED ISSUE #6244: [A/B] Phase 0a: cross-key worker integration test (validator boots on peer's DB backup)
## Context

Part of the A/B testing infrastructure epic (#912). Before we invest in snapshotting prod ScyllaDB and booting a test validator on it (Phase 2), we need empirical confirmation that a validator can boot on storage from a *different* validator without panicking or corrupting state.

## Problem statement

The testnet has 4 validators. Each has its own validator keypair. The ScyllaDB backups contain chain state (blocks, certificates, chain manager state) that includes the other validators' public keys as committee members, but the backed-up state was created with validator N's secret key.

Can validator M (different secret key, or no key at all) boot on validator N's backup and do useful work?

Code review (`linera-chain/src/manager.rs:498/533/373`, `linera-core/src/chain_worker/config.rs`) suggests yes — `ChainWorkerConfig::with_key_pair(None)` causes vote creation paths to early-return `Ok(None)` rather than panicking, making an "observer mode" validator feasible. But this has never been tested empirically.

## Task

Write an integration test that:

1. Starts a local multi-validator network (e.g. `TestBuilder` or equivalent)
2. Runs some blocks so state accumulates in the ScyllaDB (or RocksDB) backend
3. Takes a **backup** of validator N's DB (using RocksDB's `BackupEngine` API, not a file copy — the `BackupEngine` produces a consistent snapshot even under live writes)
4. Restores the backup into a fresh DB directory
5. Boots a validator instance on the restored DB with a **different keypair** (or `with_key_pair(None)`)
6. Asserts the validator starts cleanly — no panics, no `unwrap` on missing keys, service responds to gRPC health check

A passing test is the empirical gate for Phase 2's "snapshot prod, restore to test validator" approach.

## Non-goals

- Not testing catch-up (that's Phase 0b)
- Not testing production ScyllaDB (this can run on RocksDB to keep it local and fast)
- Not testing that the validator can produce valid votes (it won't, since it has no key)

## Acceptance criteria

- [ ] Integration test added under `linera-service/tests/` or `linera-core/tests/`
- [ ] Test uses `BackupEngine` (not file copy) for the backup step
- [ ] Test passes with `with_key_pair(None)` and with a mismatched keypair
- [ ] `cargo test` passes without `RUST_LOG=error` suppression (i.e., no unexpected panics)

## Related

- Parent epic: #912
- Phase 0b (stale backup catch-up): [TODO link]
- Phase 2 (deployment harness): [TODO link]
- #5965 (local validator sandbox — same fundamental question, answered here)
- `linera-core/src/chain_worker/config.rs` — `ChainWorkerConfig::with_key_pair`
- `linera-chain/src/manager.rs:498,533,373` — vote creation early-returns on `None` key

ndr-ds:
## Motivation

Part of the A/B testing infrastructure epic (#912). Before we can hot-swap validator databases, we need to confirm empirically that a `WorkerState` can start on a RocksDB backup written by a *different* validator — with no keypair (observer mode) or a mismatched keypair — without panicking or corrupting state, and that a validator restored from a stale backup can catch up by applying the blocks it missed.

## Proposal

Adds a `DatabaseBackup` test-only trait (`#[cfg(with_testing)]`) in `linera-views` with a single `backup_to(&self, dir: &Path) -> anyhow::Result<()>` method, implemented for `RocksDbDatabaseInternal` via the RocksDB backup API, and threaded through the wrapper stack via blanket impls on `LruCachingDatabase`, `ValueSplittingDatabase`, and `MeteredDatabase`. A matching `backup_to` method and a new `connect_for_testing` constructor are added to `DbStorage`. `TestBuilder` gains a `pub validator_key_pairs` field (mirroring `validator_storages`) so tests can recover a validator's own secret key.

Three tests in `linera-core` exercise the invariants:

- `test_no_key_boots_on_cross_key_backup` — observer mode (`key_pair = None`): backs up validator A's storage, restores it, boots a `WorkerState` with no key, and applies a block committed by the original network via `fully_handle_certificate_with_notifications`.
- `test_mismatched_key_boots_on_cross_key_backup` — same as above but with a freshly generated keypair that does not match the backup source.
- `test_stale_backup_catches_up_with_own_key` — stale-backup recovery: backs up one validator's storage, lets the network commit two more blocks, restores the backup with the validator's *own* key, and applies both missed blocks sequentially.

## Test Plan

```
cargo test -p linera-core --features rocksdb worker_backup -- --nocapture
```

All three tests pass.

## Release Plan

## Links

- Closes #6244 (Phase 0a of #912)


ma2bd (approved):

ma2bd:
I wonder if the test could be written with `memory`

ndr-ds:
With merge this and follow up with memory and scylladb tests

PR #6334, opened 2026-05-19
[testnet] ScyllaDB: eliminate per-batch read in write_batch via two-phase writes

ma2bd:
## Motivation

Previously, every ScyllaDB `write_batch` containing a `DeletePrefix` that overlapped a `Put` in the same batch paid a `find_keys_by_prefix` read inside `UnorderedBatch::expand_colliding_prefix_deletions`. The expansion existed because all statements in a CQL unlogged batch share one write timestamp by default, and at equal timestamps a range tombstone shadows a same-batch insert. This fired on the hot path (notably `reset_chain_manager` clearing `pending_*_blobs` views).

## Proposal

Split the write into two sequential unlogged batches: phase 1 issues the prefix-deletes only, phase 2 issues the puts and single-key deletes. The `.await` between phases plus token-aware routing gives phase 2 a strictly later coordinator timestamp, so LWW resolves the intended ordering without any read. In exclusive mode we additionally set explicit client timestamps `T` and `T+1` (via `Batch::set_timestamp`) to keep ordering robust against client/coordinator clock interactions; the per-store timestamp floor is seeded once on first write by reading `WRITETIME` of a reserved sentinel row (empty clustering key, which is unused by views and journaling alike), then advanced monotonically by 2 µs per batch. Shared mode uses coordinator-assigned timestamps.

Drop the now-unused `expand_colliding_prefix_deletions`. Flatten `UnorderedBatch` to carry `key_prefix_deletions` + `deletions` + `insertions` directly; its `SimplifiedBatch::from_batch` becomes pure (`Ok(batch.simplify())`). `SimpleUnorderedBatch` is unchanged and continues to be used by DynamoDB, which still needs full expansion since it has no native range-delete primitive.

## Test Plan

CI

## Release Plan

front-port to `main`

### DIFF
--- a/linera-views/src/backends/scylla_db.rs
+++ b/linera-views/src/backends/scylla_db.rs
@@ -851,9 +851,9 @@ impl ScyllaDbStoreInternal {
             .duration_since(UNIX_EPOCH)
             .map(|d| d.as_micros() as i64)
             .unwrap_or(0);
-        // Reserve room for `T` and `T+1` of the upcoming first batch. If
-        // another caller raced ahead and set a higher floor already, leave it.
-        let seed = now_us.max(writetime + 2);
+        // `writetime` is the last batch's `T + 1`, i.e. the highest timestamp it
+        // consumed; that is exactly what `ts_floor` tracks, so seed it directly.
+        let seed = now_us.max(writetime);
         if self
             .ts_floor
             .compare_exchange(0, seed, Ordering::Relaxed, Ordering::Relaxed)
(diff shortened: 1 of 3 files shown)


ndr-ds (approved):

linera-views/src/backends/scylla_db.rs
```diff
@@ -703,12 +789,103 @@ impl DirectWritableKeyValueStore for ScyllaDbStoreInternal {
...
+            .await?;
+        store
+            .write_batch_data(
+                &self.root_key,
+                deletions,
+                insertions,
+                t_data,
+                write_sentinel,
+            )
+            .await?;
```
ma2bd, line 830: We can't call write_batch twice in exclusive mode

PR #6338, opened 2026-05-19
[testnet] multi-leader jitter and conflict-detection fix

afck:
Backport of the client-side parts of #6289

## Motivation

If there are conflicting proposals in the first multi-leader round that probably means that several clients are trying to make a proposal at the same time. When they see the conflict, they would currently propose immediately in the next round, and likely cause a conflict again.

Sometimes retried proposals are considered "conflicts", even though they match what the user was attempting to do.

## Proposal

- Clients randomly delay re-proposing in multi-leader rounds to reduce the chance of conflicting again. Disabled with `--disable-multi-leader-jitter` (or `LINERA_DISABLE_MULTI_LEADER_JITTER` in the web client).
- Better distinguish `Committed` vs. `Conflict` outcomes.

## Test Plan

A unit test was added for the delay computation.

## Release Plan

- Release SDK.

## Links

- PR to `main`: #6289

### DIFF
--- a/linera-base/src/ownership.rs
+++ b/linera-base/src/ownership.rs
@@ -16,6 +16,7 @@ use serde::{Deserialize, Serialize};
 use thiserror::Error;
 
 use crate::{
+    crypto::{BcsHashable, CryptoHash},
     data_types::{Round, TimeDelta},
     doc_scalar,
     identifiers::AccountOwner,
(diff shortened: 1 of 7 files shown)


linera-base/src/ownership.rs
```diff
@@ -216,8 +217,79 @@ impl ChainOwnership {
...
+            .unwrap_or(TimeDelta::ZERO);
+        if round_duration == TimeDelta::ZERO
+            || self.is_preferred_multi_leader_proposer(owner, round_index)
+        {
+            return Some(TimeDelta::ZERO);
+        }
+        let priority = multi_leader_priority(owner, round_index);
+        let prefix = <[u8; 8]>::try_from(&priority.as_bytes().as_slice()[..8])
+            .expect("hash is at least 8 bytes long");
+        let hash_u64 = u64::from_le_bytes(prefix);
```
deuszx, line 269: nit: some of this could be hidden in `multi_leader_priority` itself. I don't see other uses of that function.

deuszx (approved):

linera-base/src/ownership.rs
afck: That's true. I'll take a look at `main` first, then backport it. (I'd prefer to merge this PR as it is, so it's close to #6289.)

PR #6344, opened 2026-05-20
[testnet] ScyllaDB: eliminate per-batch read in write_batch via two-phase writes (take 2)

ma2bd:
## Motivation

Reland of #6334 (merged then reverted), now fixing the atomicity regression that the first attempt introduced.

Previously, every ScyllaDB `write_batch` containing a `DeletePrefix` that overlapped a `Put` in the same batch paid a `find_keys_by_prefix` read inside `UnorderedBatch::expand_colliding_prefix_deletions`. The expansion existed because all statements in a CQL unlogged batch share one write timestamp by default, and at equal timestamps a range tombstone shadows a same-batch insert. This fired on the hot path (notably `reset_chain_manager` clearing `pending_*_blobs` views).

## Proposal

Resolve the in-batch prefix/insert collision with explicit write timestamps instead of a read, while preserving `write_batch` atomicity.

- **Exclusive mode** (single writer per partition): we own the write timestamps, so we issue the whole write as a single atomic unlogged batch with explicit per-statement `USING TIMESTAMP` — prefix-deletions at `T`, single-key deletions, insertions, and a sentinel at `T+1`. The higher data timestamp keeps a range tombstone from shadowing a same-batch insert, and the intended ordering is pinned by the timestamps rather than by send order. The per-store timestamp floor is seeded once on first write by reading the `WRITETIME` of a reserved sentinel row (empty clustering key, unused by views and journaling alike), then advanced monotonically.
- **Shared mode** (coordinator-assigned timestamps): we split the write into two sequential unlogged batches — phase 1 the prefix-deletes, phase 2 the puts and single-key deletes.

This supersedes the reverted #6334: that version also split the **exclusive** write into two sequential batches, which broke atomicity — a failure between phases could leave prefixes deleted but their replacement data missing. Since exclusive mode controls the timestamps, the ordering no longer depends on issuing two batches, so the whole write is now one atomic batch.

Drop the now-unused `expand_colliding_prefix_deletions`. `UnorderedBatch`'s `SimplifiedBatch::from_batch` becomes pure (`Ok(batch.simplify())`), so the per-batch read disappears at the journaling-layer boundary. `SimpleUnorderedBatch` is unchanged and still used by DynamoDB, which needs full expansion as it has no native range-delete primitive.

## Test Plan

CI, plus new exclusive-mode tests in `linera-views/tests/store_tests.rs`:
- `test_scylla_db_writes_from_state_exclusive` — routes the `DeletePrefix`-overlapping-`Put` cases through the explicit `T`/`T+1` path.
- `test_scylla_db_exclusive_seed_after_restart` — checks that a reconnected store reseeds its timestamp floor from the persisted sentinel and resumes strictly above prior data.

## Release Plan

front-port to `main`

### DIFF
--- a/linera-views/src/backends/scylla_db.rs
+++ b/linera-views/src/backends/scylla_db.rs
@@ -342,6 +342,18 @@ impl ScyllaDbClient {
         Ok(())
     }
 
+    /// Validates a key supplied by a caller's batch. Besides the size limit, the
+    /// key must be non-empty: the empty (zero-length) key is `WRITETIME_SENTINEL_KEY`,
+    /// reserved for the per-store timestamp sentinel that exclusive mode writes
+    /// internally. Prefix scans now deliberately hide that key, so any caller
+    /// content stored there would be silently invisible to reads. DynamoDB
+    /// likewise forbids zero-length keys.
+    fn check_batch_key(key: &[u8]) -> Result<(), ScyllaDbStoreInternalError> {
+        Self::check_key_size(key)?;
+        ensure!(!key.is_empty(), ScyllaDbStoreInternalError::ZeroLengthKey);
+        Ok(())
+    }
+
     fn check_batch_len(batch: &UnorderedBatch) -> Result<(), ScyllaDbStoreInternalError> {
         ensure!(
             batch.len() <= MAX_BATCH_SIZE,
@@ -524,13 +536,13 @@ impl ScyllaDbClient {
         let mut batch_values = Vec::new();
         let q_deletion = &self.write_batch_deletion;
         for key in batch.deletions {
-            Self::check_key_size(&key)?;
+            Self::check_batch_key(&key)?;
             batch_values.push(vec![root_key.to_vec(), key]);
             batch_query.append_statement(q_deletion.clone());
         }
         let q_insertion = &self.write_batch_insertion;
         for (key, value) in batch.insertions {
-            Self::check_key_size(&key)?;
+            Self::check_batch_key(&key)?;
             Self::check_value_size(&value)?;
             batch_values.push(vec![root_key.to_vec(), key, value]);
             batch_query.append_statement(q_insertion.clone());
@@ -599,7 +611,7 @@ impl ScyllaDbClient {
         // Single-key deletions, insertions, and the sentinel at timestamp `t + 1`.
         let t_data = t + 1;
         for key in deletions {
-            Self::check_key_size(&key)?;
+            Self::check_batch_key(&key)?;
             batch_values.push(vec![
                 CqlValue::BigInt(t_data),
                 CqlValue::Blob(root_key.to_vec()),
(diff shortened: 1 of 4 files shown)


linera-views/src/backends/scylla_db.rs
```diff
@@ -399,47 +452,185 @@ impl ScyllaDbClient {
...
+        &self,
+        root_key: &[u8],
+        key_prefix_deletions: Vec<Vec<u8>>,
+    ) -> Result<(), ScyllaDbStoreInternalError> {
+        if key_prefix_deletions.is_empty() {
+            return Ok(());
+        }
+        let session = &self.session;
+        let mut batch_query = scylla::statement::batch::Batch::new(BatchType::Unlogged);
+        let mut batch_values: Vec<Vec<Vec<u8>>> = Vec::new();
```
afck: Or, if we want to avoid the annotation (also below):

```suggestion
        let mut batch_values = Vec::<Vec<Vec<u8>>>::new();
```

linera-views/src/backends/scylla_db.rs
```diff
@@ -399,47 +452,185 @@ impl ScyllaDbClient {
...
+            .await
+            .map_err(ScyllaDbStoreInternalError::WriteBatchExecutionError)?;
+        Ok(())
+    }
+
+    /// Issues an unlogged batch containing the single-key deletions and the
+    /// insertions, letting the coordinator assign the write timestamp. Phase 2
+    /// of the shared-mode two-phase write; the `.await` between the phases lets
+    /// the coordinator give this batch a strictly later timestamp than the
+    /// prefix-deletes, so a range tombstone never shadows these insertions.
```
afck: (I don't really understand this: Which `await`? Aren't the insertions and deletions supposed to be distinct in this case, and should have the same timestamp?)

linera-views/src/backends/scylla_db.rs
```diff
@@ -517,6 +708,15 @@ pub struct ScyllaDbStoreInternal {
     semaphore: Option<Arc<Semaphore>>,
     max_stream_queries: usize,
     root_key: Vec<u8>,
+    /// Whether this store was opened with `open_exclusive`. When true, `write_batch`
+    /// resolves in-batch prefix/insert collisions via per-statement `USING TIMESTAMP`;
+    /// when false, it splits the batch into two sequential sub-batches with
+    /// server-side timestamps to preserve ordering across writers.
```
afck, line 731: (I wonder if it would be safer to just disallow batches that contain both prefix deletes and inserts/updates, if it's `false`?)

afck (approved):

linera-views/src/backends/scylla_db.rs
ma2bd: The comment is not amazing. Basically, we can't use a client-side timestamp in shared mode because there are several writers. To force the right sequence delete-by-prefix, then insertions we just use 2 batches (which will get an automatic server-side timestamp).

linera-views/src/backends/scylla_db.rs
ma2bd: Idk. In general we probably don't need atomicity in the first place

PR #6359, opened 2026-05-21
Increase default `multi_leader_rounds` from 2 to 5

afck:
## Motivation

In practice, most chain owners cooperate, if there even are more than a single owner, so multi-leader rounds are more useful than single-leader rounds.

## Proposal

Increase the default from 2 to 5 multi-leader rounds.

## Test Plan

CI; no logic was changed

## Release Plan

- Release a new SDK.

## Links

### DIFF
--- a/linera-base/src/identifiers.rs
+++ b/linera-base/src/identifiers.rs
@@ -1343,7 +1343,7 @@ mod tests {
         );
         assert_eq!(
             description.id().to_string(),
-            "6699e88680fe07056777caa26522cea7493c9cbe0611aa0e5c9d4918469d9e82"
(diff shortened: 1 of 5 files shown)


deuszx:
What's the reason for this change?

afck:
> What's the reason for this change?

I talked to @ma2bd about this: It's basically what I wrote in the description, but also that e.g. for the PM app we now create delegate users with weight 0, that can't participate in single-leader rounds at all.

deuszx:
> > What's the reason for this change?
> 
> I talked to @ma2bd about this: It's basically what I wrote in the description, but also that e.g. for the PM app we now create delegate users with weight 0, that can't participate in single-leader rounds at all.

Yes. I understood the description I just wondered if we saw the need for it or it's just a hunch.

deuszx (approved):

Issue #4690, opened 2025-09-29
Deduplicate handling of different error types returned from remote nodes.


afck:
There are different places in the updater and client where we handle errors returned by remote nodes, e.g.:
* missing blobs published by the proposal,
* missing blobs used in the certificate,
* missing chain description blobs,
* missing events,
* missing incoming messages,
* wrong block height,
* wrong round.

In all these places we need to update the remote node and try again, which currently is done it some duplicated code.

We should try and create a single function (as far as possible) that maybe takes a closure, executes it, and on error updates the remote node and tries again.

Issue #3033, opened 2024-12-16
Convert proxy errors to `NodeError`


afck:
[Unlike the server](https://github.com/linera-io/linera-protocol/blob/a3cce00ffedd93b89323c0a7c24a9b07b2ba09c8/linera-rpc/src/grpc/server.rs#L569), the proxy [doesn't convert errors to `NodeError`](https://github.com/linera-io/linera-protocol/blob/5fcbc190be2b0e4f796590ec4cdd6082b9bd1aac/linera-service/src/proxy/grpc.rs#L436). So they become a gRPC error, on which the client automatically retries. Retrying is not appropriate for many error cases, e.g. `BlobsNotFound`.

Issue #2875, opened 2024-11-13
Distinguish local and remote errors in `update_local_node_with_blobs_from`.


afck:
Currently we ignore errors returned by `update_local_node_with_blobs_from`: It's not clear which ones can be caused by a validator and which ones are really local. Only in the latter case should `communicate_with_quorum` abort and return. In the former case it should count the votes for the `NodeError`.

However, `update_local_node_with_blobs_from` calls `receive_certificate`, which in turn may download further certificates.

We should make sure that these methods return a `NodeError` exactly in all cases where the error could be caused by a faulty _remote_ node.

Issue #2857, opened 2024-11-11
Distinguish local and remote errors in `communicate_with_quorum`


afck:
`communicate_with_quorum` currently takes a call operator that returns a `Result<_, NodeError>` (async). There are several categories of node errors:

1. The validator itself encounters an error, serializes it and sends it as a response.
2. Something goes wrong with networking or serialization, e.g. gRPC.
3. The validator's response is invalid: A certificate for the wrong chain, an inconsistent chain info, etc.
4. Something goes wrong in the client, e.g. with local storage or processing certificates in the local node.

* https://github.com/linera-io/linera-protocol/issues/258 is about separating 2 from 1.
* https://github.com/linera-io/linera-protocol/issues/2856 is about case 3: the validator should be flagged as faulty, and at least be ignored for a while.

This issue is only about separating 4 from the rest. `communicate_with_quorum` should return an error rather than count the "votes for that error" if the error is local.

PR #6360, opened 2026-05-21
Fix chain manager issues with locking block and current round. (#6343)

afck:
Port of #6343.

## Motivation

There are a few bugs in several consensus protocol edge cases:

- `NewRound` notifications for round `r` can be relevant for a client even if it is already in round `>= r`, since they could contain locking blocks. The client currently ignores them.
- When updating a validator about a new round, we only send our local locking block if it's from our own current round. It can still be relevant for the validator, however, if it's from an earlier round.
- `update_signed_proposal` isn't called if `check_proposed_block` fails, e.g. due to a conflicting locked block. But the point of `signed_proposal` is only to make a note of the proposal (whether we accepted it or not) that was signed by a chain owner, because for multi-leader rounds the presence of such a proposal determines the current round.
- The worker fails to send a `NewRound` notification if the reason for the new round is an updated `signed_proposal`, because in that case it returns an error and doesn't `create_network_actions`.

## Proposal

- Only ignore `NewRound` notifications if we're already at a greater _height_, not just _round_.
- Always send the locking block to a validator if it's at least in _their_ current round (even if it's below _our_ current round).
- Call `update_signed_proposal` if `check_proposed_block` fails.
- `create_network_actions` even in the error case.

## Test Plan

Two regression tests were added.

(The second and third point above fix the same scenario; I'd say the third point is the proper fix and the second one more of a nice-to-have. However, the second point should unblock users in some cases even before the validators have been updated.)

## Release Plan

- Nothing to do.

## Links

- `testnet_conway` version: #6343

### DIFF
--- a/linera-rpc/src/simple/server.rs
+++ b/linera-rpc/src/simple/server.rs
@@ -205,13 +205,15 @@ where
     async fn handle_message(&mut self, message: RpcMessage) -> Option<RpcMessage> {
         let reply = match message {
             RpcMessage::BlockProposal(message) => {
-                match self.server.state.handle_block_proposal(*message).await {
-                    Ok((info, actions)) => {
-                        // Cross-shard requests
-                        self.handle_network_actions(actions);
-                        // Response
-                        Ok(Some(RpcMessage::ChainInfoResponse(Box::new(info))))
-                    }
+                let (result, actions) = self.server.state.handle_block_proposal(*message).await;
+                // Dispatch actions whether or not the proposal was accepted: a
+                // rejected proposal can still advance the manager's `current_round`
+                // (via `update_signed_proposal` on the `HasIncompatibleConfirmedVote`
+                // recovery path), and subscribers need the resulting `NewRound`
+                // notification.
+                self.handle_network_actions(actions);
+                match result {
+                    Ok(info) => Ok(Some(RpcMessage::ChainInfoResponse(Box::new(info)))),
                     Err(error) => {
                         self.log_error(&error, "Failed to handle block proposal");
                         Err(error.into())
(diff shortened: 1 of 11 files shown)


ma2bd (approved):

linera-rpc/src/simple/server.rs
```diff
@@ -205,13 +205,15 @@ where
...
-                match self.server.state.handle_block_proposal(*message).await {
-                    Ok((info, actions)) => {
-                        // Cross-shard requests
-                        self.handle_network_actions(actions);
-                        // Response
-                        Ok(Some(RpcMessage::ChainInfoResponse(Box::new(info))))
-                    }
+                let (result, actions) = self.server.state.handle_block_proposal(*message).await;
+                // Dispatch actions whether or not the proposal was accepted: a
+                // rejected proposal can still advance the manager's `current_round`
```
ma2bd, line 210: Should rejected proposals even be an error in the first place?

afck: Yes: We need to tell the client why we didn't accept it. In some cases (missing blobs, missing cross-chain messages, etc.) the client needs to do something about it and retry.

ma2bd: I guess that's the same categorization issue as usual with: return values / normal (non-internal) errors / retryable internal errors / non-retryable internal errors.

afck: I agree, but that's a bigger topic, and we have open issues for that: https://github.com/linera-io/linera-protocol/issues/3033, https://github.com/linera-io/linera-protocol/issues/2857, https://github.com/linera-io/linera-protocol/issues/2875, https://github.com/linera-io/linera-protocol/issues/4690

PR #6362, opened 2026-05-21
docker: build linera-tests without metrics feature

ndr-ds:
## Motivation

The `linera-tests` image was being built with the `metrics` feature enabled (the Dockerfile default: `scylladb,metrics,jemalloc`). With `metrics`, `linera service` requires `--metrics-port` as a mandatory argument. The test harness (`run_node_service_with_all_options` in `cli_wrappers/wallet.rs`) spawns `linera service` without passing `--metrics-port`, so every test that starts a node service was failing.

This broke the `testnet-conway-healthcheck` CronJob: 21/29 E2E tests failed.

## Proposal

Add an optional 7th parameter `BUILD_FEATURES` to the `build_and_push` shell function in `docker_image.yml`. When set, it overrides the Dockerfile's default `build_features` ARG via `--build-arg`.

For `linera-tests`, pass `scylladb,jemalloc` (dropping `metrics`). The `linera` binary embedded in the test image no longer requires `--metrics-port`, so the test harness can spawn it without changes.

All other images (`linera`, indexer, explorer, exporter) continue using the Dockerfile default and are unaffected.

## Test Plan

NOT YET END-TO-END VERIFIED — the fix needs to land on `testnet_conway`, trigger a CI image rebuild, and then the health-check CronJob re-run to confirm all 29 tests pass.

Logic verification:
- `build_and_push` with no 7th arg: `BUILD_FEATURES` is empty, `${BUILD_FEATURES:+...}` expands to nothing → Dockerfile default unchanged ✓
- `build_and_push ... "scylladb,jemalloc"`: adds `--build-arg build_features=scylladb,jemalloc` → `metrics` excluded → `--metrics-port` not required ✓

## Release Plan

(This PR already targets `testnet_conway`. Also needs a forward-port to `main`.)

## Links

- Companion infra fix: linera-io/linera-infra#1195 (validator check image tag)

### DIFF
--- a/.github/workflows/docker_image.yml
+++ b/.github/workflows/docker_image.yml
@@ -89,6 +89,7 @@ jobs:
             local IMAGE_LONG=$4
             local BUILD_NAME=$5
             local TARGET=$6
+            local BUILD_FEATURES=${7:-}
             local LOG_FILE="/tmp/${BUILD_NAME}.log"
 
             {
@@ -106,6 +107,7 @@ jobs:
                 --build-arg "build_date=${BUILD_DATE}" \
                 --build-arg "build_flag=--release" \
                 --build-arg "build_folder=release" \
+                ${BUILD_FEATURES:+--build-arg "build_features=${BUILD_FEATURES}"} \
                 .
 
               echo "Pushing ${BUILD_NAME} images..."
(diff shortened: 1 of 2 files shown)


ma2bd:
> With metrics, linera service requires --metrics-port as a mandatory argument.

But why? This sounds like this is the real problem that needs to be fixed

ndr-ds:
The `--metrics-port` doesn't necessarily need to be mandatory, indeed, and we can follow up on that. But the bigger issue here is that a client binary used just to run some tests doesn't need to be built with the metrics feature... we're not gonna be interested in the metrics of that client binary 😅
So I think this fix makes sense. The metrics port part can be a follow up

ma2bd (approved):

PR #6364, opened 2026-05-21
Ensure checkpoint-bootstrapped nodes have all referenced blobs.

afck:
## Motivation

The dumped execution state restores the chain's `used_blobs` set by ID, but the corresponding blob contents live in shared blob storage and may be absent on a bootstrapping node.

## Proposal

List the chain's `used_blobs` in `OracleResponse::Checkpoint` and surface them through `oracle_blob_ids` so the existing required-blob plumbing fetches or fails as usual.

## Test Plan

CI

## Release Plan

## Links

### DIFF
--- a/linera-indexer/lib/src/db/postgres/mod.rs
+++ b/linera-indexer/lib/src/db/postgres/mod.rs
@@ -450,10 +450,14 @@ impl PostgresDatabase {
                 }
                 OracleResponse::Checkpoint {
                     execution_state_blobs,
+                    used_blobs,
                 } => {
-                    let serialized = bincode::serialize(execution_state_blobs).map_err(|e| {
-                        PostgresError::Serialization(format!("Failed to serialize checkpoint: {e}"))
-                    })?;
+                    let serialized =
+                        bincode::serialize(&(execution_state_blobs, used_blobs)).map_err(|e| {
+                            PostgresError::Serialization(format!(
+                                "Failed to serialize checkpoint: {e}"
+                            ))
+                        })?;
                     ("Checkpoint", None, Some(serialized))
                 }
             };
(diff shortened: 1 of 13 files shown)


linera-indexer/lib/src/db/postgres/mod.rs
```diff
@@ -450,10 +450,14 @@ impl PostgresDatabase {
                 }
                 OracleResponse::Checkpoint {
                     execution_state_blobs,
+                    used_blobs,
                 } => {
-                    let serialized = bincode::serialize(execution_state_blobs).map_err(|e| {
-                        PostgresError::Serialization(format!("Failed to serialize checkpoint: {e}"))
-                    })?;
+                    let serialized = bincode::serialize(&(execution_state_blobs, used_blobs))
```
ma2bd: Why bincode? Are we not using BCS everywhere in the protocol?

ma2bd (approved):

linera-indexer/lib/src/db/postgres/mod.rs
afck: Right, good point! I'll change that.

afck: Actually, in the indexer we seem to use `bincode` for everything, so maybe it better to not change that in this PR?

deuszx: It doesn't matter here. We just want to store it as raw bytes. We are not sending it raw anywhere. And since bincode is faster I used that.

ma2bd: I mean "protocol" in the strict sense of the "Linera decentralized protocol" that theoretically should have public specifications, alternative implementations etc.

(The indexer is a separate tool that can be changed anytime.)

PR #6366, opened 2026-05-21
ScyllaDB: quote the namespace in the timestamped queries

ma2bd:
## Motivation

The exclusive-mode timestamped queries introduced in #6342 reference the table
without quoting the namespace, unlike every other statement and the
`CREATE TABLE`. CQL folds *unquoted* identifiers to lowercase, while the table
is created with a *quoted* (case-sensitive) name. So for any namespace
containing uppercase letters (allowed by `check_namespace`: ASCII alphanumeric
+ `_`, up to 48 chars), these queries would target a different, lowercased
table that does not exist — failing the exclusive-mode write/seed path. The
bug is masked today because test namespaces and `DEFAULT_NAMESPACE` are
lowercase.

## Proposal

Quote the namespace consistently (`{KEYSPACE}."{namespace}"`) in the five
affected statements:

- `SELECT WRITETIME(v) …` (the sentinel write-time read)
- the three `USING TIMESTAMP` `DELETE`s (unbounded prefix, bounded prefix,
  single key)
- the `USING TIMESTAMP` `INSERT`

## Test Plan

- `cargo check -p linera-views --features scylladb` passes.
- No behavioral change for the lowercase namespaces used by the existing
  ScyllaDB test suite; the fix only affects mixed-case namespaces.

## Release Plan

to be backported

## Links

- Follow-up to #6342.

### DIFF
--- a/linera-views/src/backends/scylla_db.rs
+++ b/linera-views/src/backends/scylla_db.rs
@@ -140,7 +140,7 @@ impl ScyllaDbClient {
 
         let read_writetime = session
             .prepare(format!(
-                "SELECT WRITETIME(v) FROM {KEYSPACE}.{namespace} WHERE root_key = ? AND k = ?"
+                "SELECT WRITETIME(v) FROM {KEYSPACE}.\"{namespace}\" WHERE root_key = ? AND k = ?"
             ))
             .await?;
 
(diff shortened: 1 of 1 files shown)


ma2bd:
same as #6367

PR #6370, opened 2026-05-22
[testnet] Increase default `multi_leader_rounds` from 2 to 5 (#6359)

afck:
Backport of #6359.

## Motivation

In practice, most chain owners cooperate, if there even are more than a single owner, so multi-leader rounds are more useful than single-leader rounds.

## Proposal

Increase the default from 2 to 5 multi-leader rounds.

## Test Plan

CI; no logic was changed

## Release Plan

- Release SDK.

## Links

- PR to `main`: #6359

### DIFF
--- a/linera-base/src/identifiers.rs
+++ b/linera-base/src/identifiers.rs
@@ -1231,7 +1231,7 @@ mod tests {
         );
         assert_eq!(
             description.id().to_string(),
-            "fe947fddf2735224d01eb9d56580109f2d9d02397dc5ddd748ef9beeb38d9caa"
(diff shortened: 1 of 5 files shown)


ma2bd:
How does this roll out concretely? We need update the faucet, then new chains will benefit from the option, correct?

ma2bd (approved):

afck:
> We need update the faucet, then new chains will benefit from the option, correct?

Yes, exactly, the faucet uses `ChainOwnership::single`, which now sets the number to 5.

PR #6371, opened 2026-05-22
Refuse to re-propose a pending block signed by a different owner

ma2bd:
## Motivation

Solve a client-side error seen in production, likely happening when we change the preferred owner while a pending proposal exists from the previous owner.

## Proposal

* Clear the pending block if we are about to return `InvalidSigner` any way.

Fortunately, we don't persist pending blocks in storage any more so users could reload the page and continue.

## Test Plan

CI


deuszx (approved):
Seems harmless indeed.

PR #6372, opened 2026-05-22
Introduce `U128(u128)` type for contracts and use in `WrappedFungible`.

deuszx:
## Motivation

  The bridge accepted any ERC-20 regardless of decimals, but treated every wrapped balance as a Linera `Amount` (hardcoded to 18 decimals). For a 6-decimal ERC-20 like USDC the on-chain semantics silently diverged from the EVM token's actual scale: a Linera-side transfer of `Amount::from_tokens(1)` would release `10^18` raw units on EVM — `10^12` whole tokens, not 1. The bridge had no mechanism for the operator to even declare the source token's decimals, let alone verify both sides agreed.

  ## Proposal

  - **`U128(pub u128)`** newtype in `linera-base::data_types` for raw token sub-units in the source ERC-20's decimal scale. Custom GraphQL `Scalar` impl (string in human-readable formats / GraphQL, bare
  `u128` in BCS) — works around `u128` lacking `async_graphql::InputType` without forcing Linera-native `Amount` on callers.
  - **e2e tests** exercising 6-decimal ERC-20s in both directions (EVM→Linera and Linera→EVM), making the off-by-decimals failure observable.
  - **`WrappedParameters.decimals: u8`** — the wrapped-fungible application now carries the source ERC-20's decimal scale alongside its ticker.
  - **Startup assertion in the relayer**: queries `WrappedFungibleService.decimals` via GraphQL and `ERC20.decimals()` via the EVM client, refuses to start on mismatch.
  - **`setup.sh --token-decimals N`** propagates the chosen decimals to both the deployed `LineraToken` and the wrapped-fungible parameters, so the demo's deploy is internally consistent.
  - **WrappedFungible rewritten to use `U128`** end-to-end: every `WrappedFungibleOperation` variant, `BurnEvent`, `InitialState`, `FungibleResponse`, the cross-chain `Message`, and the on-chain state
  (`MapView<AccountOwner, U128>` for accounts and allowances).
  - **EVM side**: regenerated Solidity types use `uint128` for amounts; `FungibleBridge._releaseBurn` and forge tests adjusted accordingly.
  - Frontend (`examples/bridge-demo/index.html`) converts the user-entered decimal via `ethers.parseUnits(amount, decimals).toString()` before posting the GraphQL mutation.

  ## Test Plan

  - `cargo test -p linera-bridge --features relay` — 130 tests pass
  - `cargo test -p wrapped-fungible` — mint_burn 7/7 pass
  - `forge test` in `linera-bridge/src/solidity` — 12/12 pass
  - `cargo check --tests` in `linera-bridge/tests/e2e` — compiles
  - Manual verification using UI demo with a 6-decimal token

  ## Release Plan

  ## Links


examples/wrapped-fungible/src/contract.rs
```diff
@@ -271,3 +273,16 @@ impl WrappedFungibleTokenContract {
...
 }
+
+/// Parses a `Transfer*` amount string into an [`Amount`].
+///
+/// The amount field is a `String` over the wire (so it can serialize as JSON for
+/// GraphQL clients), but represents a raw `u128` of token sub-units in the source
+/// ERC-20's decimal scale. We map straight into `Amount::from_attos` so the rest
+/// of the contract continues to operate on `Amount`.
+fn parse_token_amount(amount: &str) -> Amount {
+    let value: u128 = amount
```
afck:
```suggestion
    let value = amount
```

linera-base/src/data_types.rs
```diff
@@ -104,6 +104,73 @@ impl TryFrom<U256> for Amount {
...
+/// carries no implicit decimal scale: it is whatever the source token's own
+/// `decimals()` says it is. Used at application boundaries that bridge non-18-
+/// decimal tokens (e.g. wrapped USDC at 6) so callers don't accidentally read
+/// the value through Linera's 18-decimal lens.
+///
+/// Wire format: decimal string in human-readable serializers (JSON / GraphQL,
+/// since `u128` has no `async_graphql::InputType`), bare `u128` in binary
+/// (BCS) — identical bytes to [`Amount`] over BCS.
+#[derive(Clone, Copy, Debug, Default, Eq, Ord, PartialEq, PartialOrd, Hash)]
+pub struct TokenAmount(pub u128);
```
afck: Wouldn't you say "one token" is the big, subdivisible unit, i.e. "one coin"? E.g. is `Amount::from_tokens` (which multiplies everything by 10<sup>18</sup>) misnamed?
What else would you call it?

If you do agree that "one token" is the big unit, should we call the small one "subtoken", "subunit", "atomic unit", …?

deuszx: No, one token is the smallest not divisible unit.

afck: Then we should rename `Amount::from_tokens` to `Amount::from_coins`, and `from_attotokens` to `from_tokens` or something like that? (Not sure I agree, though; let's discuss with the team.)

deuszx: I simply wouldn't mix Amount type (native, fixed decimals) with contracts.

ma2bd: As discussed, it would be desirable to find another name because this is typically not a number of user-facing "tokens" (something that `Amount` did try to solve, making even more confusing potentially). `RawAmount` comes to mind. Maybe `U128` is also fine (and would also justify implementing `Deref` and `DerefMut`)

linera-base/src/data_types.rs
```diff
@@ -1659,6 +1726,10 @@ impl MessagePolicy {
 
 doc_scalar!(Bytecode, "A WebAssembly module's bytecode");
 doc_scalar!(Amount, "A non-negative amount of tokens.");
+doc_scalar!(
+    TokenAmount,
+    "Raw u128 token sub-units in the source token's decimal scale."
```
ma2bd: 99% of the time when the name and the description of a thing are not aligned, this means the name is not good. (Here TBF the description is also a bit cryptic)

examples/wrapped-fungible/src/state.rs
```diff
@@ -0,0 +1,141 @@
...
+        }
+        let owner_spender = OwnerSpender::new(owner, spender);
+        let allowance = self
+            .allowances
+            .get(&owner_spender)
+            .await
+            .expect("Failed allowance access")
+            .unwrap_or_default();
+        let new_allowance = allowance.0.checked_sub(amount.0).unwrap_or_else(|| {
+            panic!("Spender {spender} does not have sufficient allowance from owner {owner}; allowance={allowance} amount={amount}")
```
afck: (Line too long; maybe break the string literal.)

linera-base/src/data_types.rs
```diff
@@ -104,6 +104,78 @@ impl TryFrom<U256> for Amount {
     }
 }
 
+/// A `u128` newtype that serialises as a decimal string in human-readable
```
```suggestion
/// A `u128` newtype that serializes as a decimal string in human-readable
```

linera-base/src/data_types.rs
```diff
@@ -104,6 +104,78 @@ impl TryFrom<U256> for Amount {
...
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl std::ops::DerefMut for U128 {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.0
+    }
+}
```
afck: We're already using `derive_more`; I wonder if some of this boilerplate could be removed using its macros?

afck (approved):

linera-base/src/data_types.rs
deuszx: So you're saying Claude can't into english? 🤔

afck: Only British, it seems!

afck (approved):

PR #6373, opened 2026-05-23
Add NonCanonicalBTreeMap to skip BCS map re-sorting in value position

ma2bd:
## Motivation

BCS serializes a `BTreeMap` in *canonical* form: on every `serialize` call it re-sorts the
entries by their serialized-key bytes (an `O(n log n)` sort), and on `deserialize` it re-checks
that ordering. For maps kept in *value* position of a `RegisterView`/`MapView`, this extra work is unnecessary.

## Proposal

Introduce a small family of map/set wrappers in `linera-base` that make the (de)serialization
behavior explicit, and use them where it matters:

- `NonCanonicalBTreeMap<K, V>`: a `BTreeMap` newtype whose `serde` impl treats it as a
  `Vec<(K, V)>` — entries are serialized as a plain sequence in the map's existing key order, with
  no canonical re-sort. It derefs to `BTreeMap`, so it is a drop-in replacement.
- `NonCanonicalBTreeSet<T> = BTreeSet<T>`: an alias (sets already serialize as a sequence, so BCS
  never re-sorts them).
- `CanonicalBTreeMap<K, V> = BTreeMap<K, V>`: an alias (the canonical encoding is exactly what
  keys need).
- `CanonicalBTreeSet<T>`: a `BTreeSet` newtype that (de)serializes as a `BTreeMap<T, ()>`, so BCS
  sorts the elements canonically by their serialized bytes — for use in key position.

Applied changes:

- `ChainStateView`: `outbox_counters` → `NonCanonicalBTreeMap`; `nonempty_outboxes` /
  `nonempty_inboxes` → `NonCanonicalBTreeSet`.
- `ChainManager`: `fallback_owners` → `NonCanonicalBTreeMap`.
- Re-export the four types from `linera-sdk` (also reachable via
  `linera_sdk::linera_base_types::*`).
- Adopt `NonCanonicalBTreeSet` in the example app states that hold a set in value position
  (`hex-game`, `gen-nft`, `non-fungible`, `matching-engine`). For sets this is a no-op on the
  wire; it just makes the intended usage explicit in the reference applications.

## Test Plan

CI

## Release Plan

Backporting to testnet conway is possible but the upgrade will be a one-way door.

### DIFF
--- a/linera-base/src/data_types.rs
+++ b/linera-base/src/data_types.rs
@@ -2033,17 +2033,32 @@ mod tests {
 
         // `256u32` is chosen so that its little-endian BCS bytes sort *before* `1u32`'s,
         // i.e. the canonical (serialized-byte) order differs from the numeric `Ord` order.
-        let map =
-            NonCanonicalBTreeMap::from(BTreeMap::from([(1u32, 10u8), (256u32, 20u8), (2u32, 30u8)]));
+        let map = NonCanonicalBTreeMap::from(BTreeMap::from([
+            (1u32, 10u8),
+            (256u32, 20u8),
+            (2u32, 30u8),
+        ]));
 
         // It serializes as a plain `Vec<(K, V)>` in the map's `Ord` key order, with no canonical
         // re-sorting.
-        let entries = map.iter().map(|(k, v)| (*k, *v)).collect::<Vec<(u32, u8)>>();
-        assert_eq!(bcs::to_bytes(&map).unwrap(), bcs::to_bytes(&entries).unwrap());
+        let entries = map
+            .iter()
+            .map(|(k, v)| (*k, *v))
+            .collect::<Vec<(u32, u8)>>();
+        assert_eq!(
+            bcs::to_bytes(&map).unwrap(),
+            bcs::to_bytes(&entries).unwrap()
+        );
 
         // ... which differs from the canonical `BTreeMap` encoding that re-sorts by serialized key.
-        let canonical = map.iter().map(|(k, v)| (*k, *v)).collect::<BTreeMap<u32, u8>>();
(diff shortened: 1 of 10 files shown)


linera-base/src/data_types.rs
```diff
@@ -1779,6 +2025,71 @@ mod tests {
...
+            .iter()
+            .map(|(k, v)| (*k, *v))
+            .collect::<BTreeMap<u32, u8>>();
+        assert_ne!(
+            bcs::to_bytes(&map).unwrap(),
+            bcs::to_bytes(&canonical).unwrap()
+        );
+
+        // It round-trips.
+        let deserialized: NonCanonicalBTreeMap<u32, u8> =
```
afck, line 2064: (Type annotation; also below.)

linera-base/src/data_types.rs
```diff
@@ -1779,6 +2025,71 @@ mod tests {
...
+        // That canonical order differs from a plain `BTreeSet`'s sequence encoding, which keeps
+        // the numeric `Ord` order.
+        let plain = set.iter().copied().collect::<BTreeSet<u32>>();
+        assert_ne!(bcs::to_bytes(&set).unwrap(), bcs::to_bytes(&plain).unwrap());
+
+        // It round-trips.
+        let deserialized: CanonicalBTreeSet<u32> =
+            bcs::from_bytes(&bcs::to_bytes(&set).unwrap()).unwrap();
+        assert_eq!(set, deserialized);
+    }
```
afck, line 2091: (I'd find these tests a bit clearer if they compared to explicitly specified orderings, rather than checking that some `iter` results are equal or unequal to each other.)

afck (approved):

PR #6375, opened 2026-05-23
feat(validator): linera validator benchmark — pre-onboarding tool

eldios:
## Motivation

Before adding a validator to the active committee, the only check today is `linera validator query`, which verifies connectivity and a chain's tip but says nothing about speed, sustained-load behaviour, network path quality, or configuration soundness. This adds a richer pre-onboarding check, run manually from an operator VM.

Tracking: linera-io/linera-infra#1198.

## Proposal

New `linera validator benchmark <ADDRESS>` subcommand (in `linera-service`). It targets a single candidate validator and emits a structured report; it does not compare against the committee in-process (baselines are built by running the same tool on known-good validators and diffing JSON offline).

Layers, in execution order:
- **L1 preflight** — version, network description, baseline RTT
- **L2 partial sync (seed)** — opt-in `--deep`; seeds a bounded run of blocks **before** the read layers so they exercise real data; stateful side effect on the candidate
- **L3 read baseline** — sequential `handle_chain_info_query` (latency floor)
- **L4 read stress** — concurrency ramp; peak sustained throughput + saturation point
- **L5 bulk download** — `download_certificates_by_heights`; MB/s & certs/s
- **L6 tip-lag** — candidate tip vs max committee tip (read from the local genesis config); convergence trend

A not-yet-committee candidate may hold no blocks, which makes the read layers (L3–L5) shallow; the tool warns when a `--chain` is not held, and `--deep` seeds it first. Output: `--output json|yaml|md|brief` (repeatable, `,`/`+`-separated; default `md` to stdout). Markdown carries a summary recap + per-layer tables; `brief` is a headline list. Files are flushed after each layer (interrupt-safe; `metadata.complete` marks partials). Charm-style progress UI (`indicatif`) on a TTY, `--no-progress` to disable. Every RPC has a timeout (`--rpc-timeout-secs`, default 30) so a hung validator never blocks. No PASS/FAIL verdict — the operator interprets the numbers. Module docs (`README.md`, `REFERENCE.md`) and an example report are included under `linera-service/src/cli/validator_benchmark/`.

## Test Plan

- 36 unit tests in `validator_benchmark` (percentiles, `OutputSpec` parsing, bulk range resolution, tip-lag trend classification, sample merging, brief + markdown rendering).
- Integration test `test_validator_benchmark` on a local net (`storage-service`): runs all read-side layers against a live validator and asserts the brief on stdout plus a well-formed JSON report with every layer present and `partial_sync` absent without `--deep`.
- `cargo clippy --all-targets -- -D warnings` clean; `cargo fmt -- --check` clean on the pinned nightly; `CLI.md` regenerated to match `linera validator help-markdown`.
- Manually run against the live `k8s-validator-1.infra.linera.net` from an OVH US-EAST VM (full run, 493s, 0 errors); the captured output is the committed example.

## Release Plan

- Targets `main`. Additive operator CLI only — no transaction/WASM/storage format change. A companion PR ports the same tool to `testnet_conway` (#6374).


eldios:
Companion PR against `testnet_conway`: #6374

ndr-ds (changes requested):
I think calling this `benchmark` oversells it quite a bit... maybe we should call it `check`?

ndr-ds:
We should define "pass criteria" also as a follow up: how much worse do we allow the p99 of these measurements to be, for the validator to still be allowed in the committee?

eldios:
> I think calling this `benchmark` oversells it quite a bit... maybe we should call it `check`?

I get what you're saying but it actually is a benchmark, naming it a check would be even more off.

Usually a check has a successful outcome and a criteria to meet it.

This is a measurement tool that gives numbers which reflects how fast, what latency, etc... which is literally the definition of a benchmark.

Now, is it a thorough benchmark? absolutely not. It's just a quick pre onboarding benchmark, and it's totally fine if it's incomplete 😬 

Unless we plan on creating a subcommand that has the same name at the same place (`linera`->`validator` level) then I wouldn't worry too much about it

And if needed we can always expand more in future

eldios:
> We should define "pass criteria" also as a follow up: how much worse do we allow the p99 of these measurements to be, for the validator to still be allowed in the committee?

That's still in flight, but it doesn't belong to the command itself I think... that's more of a decision we internally take based on a number of other factors, including the numbers (but not limited to them).

PR #6377, opened 2026-05-25
Make "genesis" blob origin explicit.

afck:
## Motivation

There is another spurious "blobs not found" error: When we request a timeout certificate on a _root_ chain that does not have any blocks yet. That's because root chain descriptions are the only blobs that don't have a blob _state_.

## Proposal

Add a blob state for them, and make the "genesis" origin explicit.

## Test Plan

A regression test was added.

## Release Plan

- Would be nice to fix this on `testnet_conway` somehow, too. On the other hand, it's a very rare edge case.

## Links

### DIFF
--- a/linera-storage/src/lib.rs
+++ b/linera-storage/src/lib.rs
@@ -263,8 +263,13 @@ pub trait Storage: linera_base::util::traits::AutoTraits + Sized {
         ChainRuntimeContext<Self>: ExecutionRuntimeContext,
     {
         let id = description.id();
-        // Store the description blob.
-        self.write_blob(&Blob::new_chain_description(&description))
+        // Store the description blob and a `Genesis` blob state for it. The blob
+        // is not published by any block, so its provenance is the genesis config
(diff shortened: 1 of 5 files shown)


linera-storage/src/lib.rs
```diff
@@ -263,9 +263,21 @@ pub trait Storage: linera_base::util::traits::AutoTraits + Sized {
...
+        let description_blob = Blob::new_chain_description(&description);
+        let description_blob_id = description_blob.id();
+        self.write_blob(&description_blob).await?;
+        self.maybe_write_blob_states(
+            &[description_blob_id],
+            BlobState {
+                origin: BlobOrigin::Genesis,
+                last_used_by: None,
+                epoch: None,
+            },
```
deuszx: Or simpler, `BlobState::genesis()` as the helper. Unless we don't need it anywhere else. Actually, it could be some kind of `const`.

afck: Done.

deuszx (approved):

PR #6378, opened 2026-05-25
Use the correct signer for pending blocks.

afck:
## Motivation

`process_pending_block_without_prepare` calls `self.identity()` _again_, rather than taking the identity from the pending block. This can make the client submit a block signed by the wrong key, which is invalid.

## Proposal

Sign with the correct key.

## Test Plan

A regression test was added.

## Release Plan

- Could be backported, but https://github.com/linera-io/linera-protocol/pull/6371 already works around it.

## Links

### DIFF
--- a/linera-core/src/client/chain_client/mod.rs
+++ b/linera-core/src/client/chain_client/mod.rs
@@ -1937,12 +1937,12 @@ impl<Env: Environment> ChainClient<Env> {
         {
             return self.finalize_locking_block(info).await;
         }
-        let owner = self.identity().await?;
+        let identity = self.identity().await?;
 
         let local_node = &self.client.local_node;
         // Otherwise we have to re-propose the highest validated block, if there is one.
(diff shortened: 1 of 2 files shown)


linera-core/src/client/chain_client/mod.rs
```diff
@@ -1971,9 +1971,42 @@ impl<Env: Environment> ChainClient<Env> {
                     debug!("Retrying locking block from fast round.");
                     (block, blobs)
                 }
-            }
+            };
+            (block, blobs, identity)
```
deuszx, line 1975: What if the locking block had a different identity from `self.identity().await?;`?

afck: That's fine: In that case, we don't have to sign a _new_ block proposal.

When re-proposing a locked block, that block already has either a validated block _certificate_ or (if from the fast round) a signature from a super owner. We are now only signing that we are re-proposing this, but the authenticated signer remains the original proposer.

deuszx (approved):

PR #6383, opened 2026-05-26
[testnet] Add pending-proposal, round, and ownership APIs to `@linera/client`

ma2bd:
## Motivation

The web client (`@linera/client`) exposes only a small slice of chain operations.
Frontends that build wallet / chain-management UIs need to inspect and manage a chain's
consensus state — clear a stuck pending proposal, tell which round a block would be
proposed in, and query or adjust chain ownership — none of which were reachable from
JavaScript.

## Proposal

Add five methods to the `Chain` class, and un-gate the one `ChainClient` method they rely
on (`clear_pending_proposal`, previously `#[cfg(with_testing)]`):

- **`clearPendingProposal()`** — discards the queued block proposal and refreshes the copy
  persisted in the wallet (fast-round proposals are persisted across sessions, so the
  in-memory clear alone would let them resurrect on reload).
- **`nextRound()`** — returns `{ kind, number, leader, canPropose }` describing the round a
  block would currently be proposed in (`"fast"` / `"multiLeader"` / `"singleLeader"` /
  `"validator"`). `leader` is `undefined` when any eligible owner may propose (fast and
  multi-leader rounds), and `canPropose` reflects the connected identity.
- **`setMultiLeaderRounds(n)`** — changes only `multi_leader_rounds`, committing a
  `ChangeOwnership` while preserving the rest of the ownership configuration.
- **`isOwner(owner)`** — whether an address is a (super or regular) owner. Useful before
  `addOwner`, which silently overwrites an existing owner's weight rather than failing.
- **`ownerWeight(owner)`** — the weight of a regular owner, or `undefined` for super owners
  (which have no weight) and non-owners.

Weights cross the wasm boundary as `number`, consistent with the existing
`addOwner({ weight })`.

## Test Plan

- Added browser integration tests in `web/@linera/client/tests/Chain.test.ts` covering each
  method, using the same faucet harness as `Client.test.ts`. These run via `pnpm test`
  against a faucet (`LINERA_FAUCET_URL`).
- Locally verified that `linera-web` builds for `wasm32-unknown-unknown`, that
  `cargo clippy -p linera-web` is clean, and that `build.bash` regenerates the TypeScript
  bindings (`clearPendingProposal`, `nextRound`, `setMultiLeaderRounds`, `isOwner`,
  `ownerWeight`) with the expected signatures.

## Release Plan

to be ported to `main`


web/@linera/client/src/chain/mod.rs
```diff
@@ -135,6 +152,152 @@ impl Chain {
...
+    /// not a regular owner (it may be a super owner, which has no weight, or not an owner
+    /// at all).
+    ///
+    /// The weight determines how often the owner is selected as the leader in
+    /// single-leader and validator rounds.
+    ///
+    /// # Errors
+    /// If the chain ownership cannot be retrieved.
+    #[wasm_bindgen(js_name = ownerWeight)]
+    pub async fn owner_weight(&self, owner: AccountOwner) -> JsResult<Option<f64>> {
```
deuszx, line 201: Why f64 and not u64?

web/@linera/client/src/chain/mod.rs
```diff
@@ -135,6 +152,152 @@ impl Chain {
...
+    }
+
+    /// Discards any pending block proposal on this chain.
+    ///
+    /// When a proposal fails to reach a quorum (for example because the client went
+    /// offline mid-round) it stays queued and is retried before any new block. Call this
+    /// to drop it so a fresh block can be proposed instead.
+    ///
+    /// Importantly, this should never be used to clear a proposal already submitted in
+    /// the fast round.
```
deuszx: It'd be safer to return an error if we try to clear the proposal for the fast round. Client has the knowledge about this.

web/@linera/client/src/chain/mod.rs
```diff
@@ -135,6 +152,152 @@ impl Chain {
...
+    /// proposals to a single owner) and whether this client's identity may propose.
+    ///
+    /// `leader` is `undefined` in the fast and multi-leader rounds, where any eligible
+    /// owner may propose; in the single-leader and validator rounds it is the owner
+    /// currently allowed to propose.
+    ///
+    /// # Errors
+    /// If the chain information cannot be retrieved.
+    #[wasm_bindgen(js_name = nextRound)]
+    pub async fn next_round(&self) -> JsResult<RoundInfo> {
```
deuszx: nit: rename to `next_round_info`.

deuszx (approved):
LGTM but I'd consider returning an error if caller tries to clear the pending proposal on `Fast` round.

web/@linera/client/src/chain/mod.rs
```diff
@@ -135,6 +152,152 @@ impl Chain {
...
+        let identity = self.chain_client.identity().await?;
+        let can_propose = match manager.leader {
+            Some(leader) => leader == identity,
+            None => match round {
+                Round::Fast => manager.ownership.super_owners.contains(&identity),
+                _ => manager
+                    .ownership
+                    .can_propose_in_multi_leader_round(&identity),
+            },
+        };
```
afck: Maybe this could be a method of `ChainOwnership`?

afck (approved):

web/@linera/client/src/chain/mod.rs
ma2bd: `addOwner({ weight })` already takes a `number` (f64). Kept as `f64` for consistency.

web/@linera/client/src/chain/mod.rs
ma2bd: Done in d1bfc47582 — `clearPendingProposal` now fetches `chain_info` and returns an error if `current_round == Round::Fast`, instead of only warning in the docs.

web/@linera/client/src/chain/mod.rs
ma2bd: Done in d1bfc47582 — renamed to `nextRoundInfo` (both the JS name and the Rust fn), and updated the tests.

web/@linera/client/src/chain/mod.rs
ma2bd: Done — added `ChainOwnership::can_propose(owner, round, leader)` in 1688917fe2 and call it here, so the leader / fast / multi-leader branching lives on `ChainOwnership`.

ma2bd: Actually it was not great in `ChainOwnership` so I moved it again

PR #6385, opened 2026-05-26
Add WebCryptoEd25519 signer + Address32 owner bridge dispatch

deuszx:
## Motivation

The published `@linera/client` only ships `PrivateKey` as a JS `Signer` implementation, and its own docstring warns it is "intended only for testing or development." Consumer apps that need a persistent in-browser signer (e.g. autosigner / session-key patterns) end up storing raw secp256k1 hex in `localStorage`. That key is plaintext at rest, readable by any same-origin JS, and exfiltratable.

The Linera protocol already supports both secp256k1 (EVM) and Ed25519 `AccountOwner` variants. Web Crypto exposes Ed25519 with non-extractable `CryptoKey` handles persistable in IndexedDB (Chrome 137+, Firefox 129+, Safari 17+). This PR plugs that browser-native primitive into the existing `Signer` boundary so app authors have a production-grade alternative to the dev-only `PrivateKey`.

## Proposal

- `WebCryptoEd25519` (new): a `Signer` impl backed by a non-extractable Ed25519 `CryptoKey` from `crypto.subtle.generateKey({name:"Ed25519"}, false, ["sign","verify"])`, with the handle persisted in
IndexedDB across sessions. Private key bytes never enter JS.
- `WebCryptoEd25519.generate()` + `persist(recordKey)` split lets callers defer durable storage until an out-of-band setup step (e.g. on-chain `addOwner`) succeeds.
- `accountOwnerFromEd25519PublicKey(pubBytes)` exposed via wasm-bindgen so the JS signer derives `AccountOwner::Address32` without reimplementing BCS + Keccak256.
- wasm `Signer` bridge (`web/@linera/client/src/signer/mod.rs`) now dispatches on the `AccountOwner` variant: `Address20` -> `AccountSignature::EvmSecp256k1` (unchanged), `Address32` ->
`AccountSignature::Ed25519` (new). Defense-in-depth: the bridge re-derives `AccountOwner::from(public_key)` and rejects a returned signature whose pubkey does not match the requested owner.
- `Signer` TS interface gains a required `getPublicKey(owner): Promise<string>` method so the bridge can fetch the Ed25519 pubkey for the `AccountSignature::Ed25519 { signature, public_key }` construction.
Updated impls in this repo: `PrivateKey` (already conformant), `Composite` (delegates), `@linera/metamask` (throws — EVM owners never reach this path). This is a breaking change for external `Signer` impls.
- `InvalidAccountOwnerType` Display message tightened (no longer claims "Expected AccountOwner::Address20").

Same-origin policy already prevents cross-origin access to either storage backend. The improvement is against same-origin attackers (XSS, supply-chain compromise of any dependency, malicious extension content script): the previous `localStorage` hex was exfiltratable, the new non-extractable `CryptoKey` is not. An attacker who lands in the page can still ask the key to sign while the tab is open, but cannot ship the key off the device for later or remote use.

## Test Plan

- New `@vitest/browser` (Chromium) suite `tests/WebCryptoEd25519.test.ts` — 11 cases:
  - Wasm helper `accountOwnerFromEd25519PublicKey` matches a Rust-side known vector pinned in `linera-base`.
  - `create` -> `load` roundtrip preserves the owner address and signs verifiably against the exported public key via `crypto.subtle.verify`.
  - `loadOrCreate` idempotency.
  - Owner-mismatch rejection on `sign` and `getPublicKey`.
  - `generate` (in-memory only) leaves `load` returning `null`; `generate` then `persist` makes it loadable.
- New Rust unit `linera-base/src/identifiers.rs::ed25519_public_key_to_account_owner_known_vector` pins the derivation hex against silent drift; the JS test references the same vector.
- All existing `@linera/client` tests continue to pass (`Client.test.ts` failures present before this branch require a live faucet and are unrelated).
- Manual reproduction:
  ```
  cd web/@linera/client
  bash build.bash --release
  pnpm test -- WebCryptoEd25519     # 11/11 pass
  cd ../../..
  cargo test --package linera-base --lib ed25519_public_key_to_account_owner_known_vector
  ```

## Release Plan

## Links

### DIFF
--- a/web/@linera/client/src/signer/mod.rs
+++ b/web/@linera/client/src/signer/mod.rs
@@ -79,6 +79,12 @@ extern "C" {
 
     #[wasm_bindgen(catch, method, js_name = "containsKey")]
     async fn contains_key(this: &Signer, owner: AccountOwner) -> Result<JsValue, JsValue>;
+
+    #[wasm_bindgen(catch, method, js_name = "getPublicKey")]
+    async fn get_public_key(
+        this: &Signer,
+        owner: AccountOwner,
+    ) -> Result<js_sys::JsString, JsValue>;
 }
 
 impl linera_base::crypto::Signer for Signer {
--- a/web/@linera/client/src/signer/mod.rs
+++ b/web/@linera/client/src/signer/mod.rs
@@ -36,10 +36,7 @@ impl Display for Error {
                 write!(f, "Unexpected signature format received from JS")
             }
             Error::InvalidAccountOwnerType => {
-                write!(
-                    f,
-                    "Invalid account owner type provided. Expected AccountOwner::Address20"
-                )
+                write!(f, "Invalid account owner type provided")
             }
             Error::Unknown => write!(f, "An unknown error occurred"),
         }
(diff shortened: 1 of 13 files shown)


web/@linera/client/src/signer/mod.rs
```diff
@@ -93,21 +93,51 @@ impl linera_base::crypto::Signer for Signer {
...
+                    address: *address,
+                })
+            }
+            AccountOwner::Address32(_) => {
+                let pub_str: String = self.get_public_key(*owner).await?.into();
+                let public_key = parse_ed25519_public_key(&pub_str)
+                    .ok_or(Error::PublicKeyParse)?;
+                let signature = parse_ed25519_signature(&sig_str)
+                    .ok_or(Error::UnexpectedSignatureFormat)?;
+                // Defense in depth: a malicious or buggy signer might return a valid
```
ma2bd: Afaict that's an early reject, not really "defense in depth".

deuszx: but this is our code that validates what the client side (`Signer` implementation on the JS client) returned.

deuszx: [Improve comment](https://github.com/linera-io/linera-protocol/pull/6385/commits/cc4ca57466cef58edbddf61b55588e8e186916e1)

web/@linera/metamask/src/signer.ts
```diff
@@ -79,6 +79,16 @@ export default class Signer implements SignerInterface {
     }
   }
 
+  async getPublicKey(_owner: string): Promise<string> {
+    // MetaMask signs only for `Address20` (EVM secp256k1) owners. The wasm bridge
+    // never calls `getPublicKey` on the Address20 path — EVM signatures carry the
+    // signer's address inline in `AccountSignature::EvmSecp256k1`. If we get here,
```
ma2bd, line 85: We shouldn't talk about the wasm bridge here.

ma2bd (approved):

ma2bd:
note: This kind of features would be better developed on `main` first.

deuszx:
> note: This kind of features would be better developed on `main` first.

if I start by developing this on `main` I have no (easy) way to test it. By developing it here I can use testnet PM app deployment and make sure this works where it matters the most.

Another thing – do we release npm packages from `main` or `testnet_conway`?

web/@linera/metamask/src/signer.ts
deuszx: Why not? Wasm bridge here means the wasm<>host boundary

PR #6389, opened 2026-05-26
linera-storage: add in-memory caches for immutable DB reads

ndr-ds:
## Motivation

`DbStorage` pays a full Scylla/RocksDB round-trip on every call to four read methods that return permanently immutable data. On busy validators this contributes directly to Scylla read pressure (one node sustained ~7-8s p99 read spikes, with cache storms making it worse).

The four uncached paths:
- `read_certificate_hashes_by_heights` — `(ChainId, BlockHeight) → CryptoHash`
- `read_event_block_heights` — `EventId → BlockHeight`
- `read_events_from_index` — per-event bytes (keys still require a DB scan, but values now hit the cache)
- `read_network_description` — single `NetworkDescription` value

All four are permanently immutable once written. They are safe to cache indefinitely.

## Proposal

Add three new caches to `StorageCaches`:

- `block_hash_by_height: Arc<ValueCache<(ChainId, BlockHeight), CryptoHash>>` — S3-FIFO bounded, consistent with the existing five caches
- `event_block_height: Arc<ValueCache<EventId, BlockHeight>>` — same
- `network_description: Arc<OnceLock<NetworkDescription>>` — single-value, zero overhead after first read

Read path: check-then-populate pattern (same as `read_certificates_raw`). Misses batch to DB; hits short-circuit. Write path: caches are populated in `write_blobs_and_certificate` at write time so write-then-read (the common path) is immediately cached.

`read_events_from_index` still scans DB for the key list (unavoidable), but now checks the existing `event` cache before bulk-reading values.

New CLI flags with conservative defaults (1 000 entries), overridden to larger values in the validator `values.yaml`:
- `--block-hash-by-height-cache-size` (validator: 100 000)
- `--event-block-height-cache-size` (validator: 50 000)

New Prometheus counters: `READ_BLOCK_HASH_BY_HEIGHT_COUNTER`, `READ_EVENT_BLOCK_HEIGHT_COUNTER` (labeled `[CACHE]`/`[DB]`). `read_network_description` gets `[CACHE]`/`[DB]` on the existing `READ_NETWORK_DESCRIPTION_COUNTER`.

## Test Plan

- `cargo clippy` clean on `--all-targets --all-features` and `--no-default-features`
- Covered by existing integration tests exercising `read_certificate_hashes_by_heights` and `read_events_from_index` for the full round-trip

## Release Plan

- These changes should be backported to the latest `testnet` branch, then be released in a validator hotfix.

## Links

- Prerequisite backport (index to testnet_conway): #6393

### DIFF
--- a/linera-bridge/tests/e2e/tests/burns_per_evm_tx.rs
+++ b/linera-bridge/tests/e2e/tests/burns_per_evm_tx.rs
@@ -99,6 +99,8 @@ async fn burns_per_evm_tx(
             certificate_cache_size: 1000,
             certificate_raw_cache_size: 1000,
             event_cache_size: 1000,
+            block_hash_by_height_cache_size: 1000,
+            event_block_height_cache_size: 1000,
             cache_cleanup_interval_secs: linera_storage::DEFAULT_CLEANUP_INTERVAL_SECS,
         },
     )
--- a/linera-bridge/tests/e2e/tests/burns_per_evm_tx.rs
+++ b/linera-bridge/tests/e2e/tests/burns_per_evm_tx.rs
@@ -23,7 +23,7 @@ use linera_client::{chain_listener::ClientContext as _, client_context::ClientCo
 use linera_core::environment::wallet::Memory;
 use linera_execution::{Operation, WasmRuntime};
 use linera_faucet_client::Faucet;
-use linera_storage::{DbStorage, StorageCacheConfig};
+use linera_storage::DbStorage;
 use linera_views::backends::memory::{MemoryDatabase, MemoryStoreConfig};
 use test_case::test_case;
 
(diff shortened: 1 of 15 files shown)


linera-bridge/tests/e2e/tests/burns_per_evm_tx.rs
```diff
@@ -94,6 +94,8 @@ async fn burns_per_evm_tx(
             certificate_cache_size: 1000,
             certificate_raw_cache_size: 1000,
             event_cache_size: 1000,
+            block_hash_by_height_cache_size: 1000,
+            event_block_height_cache_size: 1000,
             cache_cleanup_interval_secs: linera_storage::DEFAULT_CLEANUP_INTERVAL_SECS,
         },
```
ma2bd: @deuszx should we deduplicating this?

linera-storage/src/db_storage.rs
```diff
@@ -1384,13 +1529,25 @@ where
...
-        let maybe_value = store.read_value(NETWORK_DESCRIPTION_KEY).await?;
+        let maybe_value: Option<NetworkDescription> =
+            store.read_value(NETWORK_DESCRIPTION_KEY).await?;
         #[cfg(with_metrics)]
         metrics::READ_NETWORK_DESCRIPTION
             .with_label_values(&[metrics::DB])
             .inc();
+        if let Some(ref desc) = maybe_value {
+            // Ignore the error: another thread may have populated the cache concurrently.
+            drop(self.caches.network_description.set(desc.clone()));
```
ma2bd: nit: `drop` is like `let _ =`. (unless I'm missing some new clippy rules?) We can't tell if this is dropping a future. Perhaps the right pattern is: `if let Err(error) = ... { debug!(%error, "something"); }`

ma2bd (approved):

linera-storage/src/db_storage.rs
```diff
@@ -1349,24 +1470,48 @@ where
...
+                    index,
+                };
+                self.caches.event.insert(&event_id, event_bytes.clone());
+                returned_values.push(IndexAndEvent {
+                    index,
+                    event: event_bytes,
+                });
+            }
         }
+        returned_values.sort_unstable_by_key(|ie| ie.index);
```
ma2bd: could we avoid this?

PR #6390, opened 2026-05-26
Backport #5710 (read_event via event-block-height index) + the event-block-height cache to testnet_conway

ndr-ds:
## Motivation

`read_event` on `testnet_conway` can only resolve an event if the client has
already subscribed and walked the publisher chain's `previous_event_blocks`
list. main got a faster, more general path in #5710: validators maintain an
`EventBlockHeight` index, and a client missing an event asks a validator for the
exact publishing block height, downloads only that block, and retries.

This is the `testnet_conway` backport of #5710, plus the fourth `DbStorage`
cache from #6389 (`event_block_height`), which was deferred out of the #6474
backport precisely because it had no index to read until now.

## Proposal

1. **Storage** — `RootKey::EventBlockHeight(ChainId)` partition, written for
   every event at certificate-write time; `read_event_block_heights` reads it.
2. **RPC** — new `EventBlockHeights` validator method end-to-end (proto, gRPC +
   simple clients/server, proxy-local handler, node trait, test validator). The
   `RpcMessage` variants are appended at the end of the enum so existing wire
   indices are unchanged.
3. **Client** — `download_certificates_for_events` first queries the index and
   downloads exactly the publishing blocks (verifying each block actually
   contains the expected events, rejecting a lying validator), then **falls back
   to the existing `previous_event_blocks` walk** on any failure. Events that
   predate the index, or validators that don't yet serve the RPC, degrade to
   exactly the pre-PR behavior and error shape.
4. **Cache** — `event_block_height: Arc<ValueCache<EventId, BlockHeight>>`
   (the 4th cache from #6389), populated write-through, with its CLI flag.

Fully additive: no existing event-resolution path is removed.

## Backwards compatibility

- `RootKey::EventBlockHeight` is appended last in the enum — every existing
  on-disk partition key keeps its index. A node rolled back to old code simply
  leaves the index partition orphaned (never read); a new node reading old
  history finds the index empty (not backfilled) and the client falls back to
  the walk.
- `RpcMessage` variants appended last (wire indices unchanged); the new gRPC
  method is additive. During a rolling upgrade, querying an old validator that
  doesn't serve the RPC just errors that one attempt, and the client falls back
  to the walk.

## Test Plan

- `cargo check --workspace --all-targets` (excl. wasm-only `linera-web`, per CI) + clippy clean
- RPC format snapshot regenerated (append-only — existing variant indices unchanged)
- `test_memory_read_event_downloads_publisher_chain` (ported from #5710) exercises
  the full index path: `UpdateStreams` without sync → `EventsNotFound` → index
  lookup → targeted certificate download → retry succeeds

## Release Plan

- Backport to be released in a validator hotfix.

## Links

- Upstream feature: #5710
- Main caches PR (the 4th cache deferred from here): #6389
- Index-free caches backport (the other 3): #6474


afck:
I guess we'll either need migration (create index entries for existing events) or fallback code (if no index entry found, don't assume that the event doesn't exist)?

ndr-ds:
Yes, this PR is still very much incomplete 😅 I'll mark it as ready for review once it's ready

linera-service/src/proxy/grpc.rs
```diff
@@ -956,6 +956,28 @@ where
...
     }
+
+    #[instrument(skip_all, err(level = Level::WARN), fields(
+        method = "event_block_heights"
+    ))]
+    async fn event_block_heights(
+        &self,
+        request: Request<api::EventBlockHeightsRequest>,
+    ) -> Result<Response<api::EventBlockHeightsResponse>, Status> {
+        let event_ids: Vec<linera_base::identifiers::EventId> = request
```
afck: (Type annotation should be on `try_into`?)

linera-storage/src/db_storage.rs
```diff
@@ -1759,6 +1804,73 @@ where
...
+        &self,
+        event_ids: &[EventId],
+    ) -> Result<Vec<Option<BlockHeight>>, ViewError> {
+        if event_ids.is_empty() {
+            return Ok(Vec::new());
+        }
+
+        let mut results = vec![None; event_ids.len()];
+        // Check cache first; collect misses.
+        let mut misses: Vec<usize> = Vec::new();
```
afck: (Type annotation)

linera-core/src/client/mod.rs
```diff
@@ -773,11 +773,172 @@ impl<Env: Environment> Client<Env> {
     /// using the event block height index on validators. Queries a validator for
     /// the block heights, downloads those certificates, and processes them — all
     /// as one atomic unit per validator attempt, with staggered fallback.
+    ///
+    /// The index has no backfill, so events published before the validators were
+    /// upgraded resolve to `None`; callers fall back to walking
+    /// `previous_event_blocks` whenever this fails.
```
afck: Is this documenting the call site rather than the method itself?

linera-core/src/client/mod.rs
```diff
@@ -792,11 +792,173 @@ impl<Env: Environment> Client<Env> {
...
+        {
+            Ok(()) => return Ok(()),
+            Err(error) => {
+                debug!(
+                    %error,
+                    "event block height index lookup failed; \
+                     falling back to walking previous event blocks",
+                );
+            }
+        }
```
afck: If it returned `NodeError::EventsNotFound(remaining_event_ids)`, should we now replace `event_ids` with `remaining_event_ids`, so we don't download the other ones again?

afck (approved):

PR #6395, opened 2026-05-27
Require docs on 7 small library crates with #![deny(missing_docs)]

ma2bd:
## Motivation

We want to progressively require documentation on the library crates of the main
workspace via `#![deny(missing_docs)]`. This is the first batch, covering the smallest
crates. The work is mirrored on `testnet_conway` (see Links) to keep the two branches in
sync and ease backports.

## Proposal

Add `#![deny(missing_docs)]` and document every public item in seven small library crates:

- `linera-indexer-graphql-client`
- `linera-sdk-derive`
- `linera-faucet-client`
- `linera-indexer-plugins`
- `linera-persistent`
- `linera-metrics`
- `linera-wallet-json`

Generated public items that cannot carry doc comments are exempted narrowly, with an
explanatory comment:
- `#[derive(GraphQLQuery)]` / async-graphql bindings → a module-level `#![allow(missing_docs)]`
  on the generated-binding file (a struct-level `allow` cannot reach the generated sibling
  modules).
- `thiserror_context::impl_context!` → the macro invocation is wrapped in a private
  `#[allow(missing_docs)] mod error { … }` and re-exported (an `allow` on the macro
  invocation is ignored by the compiler).

One non-doc change: `linera-sdk-derive`'s `pub mod tests` becomes a private `mod tests`
(the `pub` was gratuitous and tripped `missing_docs` under `--all-targets`).

## Test Plan

For each crate: `cargo check -p <crate> --all-features --all-targets` is clean under the new
lint, and `cargo +nightly fmt -p <crate> -- --check` passes.

## Release Plan

- Nothing to do / These changes follow the usual release cycle.

## Links

- testnet_conway counterpart: (to be linked)
- [reviewer checklist](https://github.com/linera-io/linera-protocol/blob/main/CONTRIBUTING.md#reviewer-checklist)

PR #6396, opened 2026-05-27
Require docs on 7 small library crates with #![deny(missing_docs)] (testnet_conway)

ma2bd:
## Motivation

`testnet_conway` counterpart of #6395. We are progressively requiring documentation on the
library crates via `#![deny(missing_docs)]`, mirroring the work on both `main` and
`testnet_conway` to keep the branches in sync and ease backports.

## Proposal

Same seven crates as #6395, documented identically where the code is identical:

- `linera-indexer-graphql-client`, `linera-sdk-derive`, `linera-faucet-client`,
  `linera-indexer-plugins`, `linera-persistent`, `linera-metrics`, `linera-wallet-json`.

The first seven commits mirror #6395 (cherry-picked). Where the branches' code already
diverges, docs were adjusted to match `testnet_conway`'s actual API rather than `main`'s
(e.g. `faucet-client`'s `ErrorInner` has no `ArithmeticError` variant here; `wallet-json`'s
`forget_keys` has a different signature). A final commit documents the public items that
exist only on `testnet_conway` (`forget_chain`, `start_metrics_with_extras`,
`owned_chain_ids`), isolated so the cherry-picked commits stay a clean mirror of `main`.

Same codegen conventions as #6395 (`GraphQLQuery`/async-graphql → module-level allow;
`impl_context!` → wrapper module + re-export).

## Test Plan

For each crate: `cargo check -p <crate> --all-features --all-targets` is clean under the new
lint, and `cargo +nightly fmt -p <crate> -- --check` passes.

## Release Plan

## Links

- `main` counterpart: #6395

### DIFF
--- a/linera-faucet/client/src/lib.rs
+++ b/linera-faucet/client/src/lib.rs
@@ -38,8 +38,9 @@ pub use error::Error;
 
 mod error {
     // `impl_context!` generates a public `Error` newtype (with accessors) that cannot carry
-    // doc comments, so this wrapper module is exempt from the crate's `missing_docs` policy.
-    #![allow(missing_docs)]
+    // doc comments, so this wrapper module is exempted from the crate's `missing_docs` policy.
+    // `expect` (rather than `allow`) flags this if the macro ever stops generating such items.
+    #![expect(missing_docs)]
 
     use thiserror_context::Context;
 
--- a/linera-faucet/client/src/lib.rs
+++ b/linera-faucet/client/src/lib.rs
@@ -3,6 +3,8 @@
 
 //! The client component of the Linera faucet.
 
+#![deny(missing_docs)]
+
 // TODO(#3362): generate this code
 
 use std::collections::BTreeMap;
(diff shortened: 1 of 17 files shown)


linera-faucet/client/src/lib.rs
```diff
@@ -17,18 +19,34 @@ use linera_execution::{committee::ValidatorState, Committee, ResourceControlPoli
...
     Http(#[from] reqwest::Error),
 }
 
-thiserror_context::impl_context!(Error(ErrorInner));
+pub use error::Error;
+
+mod error {
+    // `impl_context!` generates a public `Error` newtype (with accessors) that cannot carry
+    // doc comments, so this wrapper module is exempt from the crate's `missing_docs` policy.
+    #![allow(missing_docs)]
```
afck: (We could use `expect` here.)

afck (approved):
Indexer schema needs to be regenerated.

(Otherwise I only glanced over it, assuming there are no major differences compared to the `main` version?)

ma2bd:
Addressed review: regenerated `gql/operations_schema.graphql` (the new GraphQL descriptions come from documenting the operations plugin's async-graphql types), and switched the generated-code exemptions to `#[expect(missing_docs)]` to match #6395. There are no differences from the main PR beyond the spots where the two branches' code already diverges (noted in the description).

PR #6398, opened 2026-05-27
Simplify `communicate_concurrently` to return per-validator errors; log

afck:
## Motivation

In many `communicate_concurrently` call sites, we simply replace the validators' errors with a single new error. This makes it hard to debug the root cause.

## Proposal

Drop the `err` closure parameter. `communicate_concurrently` can only fail when all validators returned an error, so the natural return type is `Result<V, Vec<(ValidatorPublicKey, E)>>` — let each call site own the conversion (and warn-logging) it needs.

This also surfaces an empty-validators case explicitly. The two scheduler sites previously used `errors.last().unwrap()`, which would panic if peers was empty; they now map that to `NodeError::NoValidators`.

Add logging.

## Test Plan

CI

## Release Plan

- We could backport the refactoring to `testnet_conway`, if we prefer it over #6391.

## Links

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -11,7 +11,7 @@ use std::{
 
 use custom_debug_derive::Debug;
 use futures::{
-    future::Future,
+    future::{Future, TryFutureExt as _},
     stream::{self, AbortHandle, FuturesOrdered, FuturesUnordered, StreamExt},
 };
 #[cfg(with_metrics)]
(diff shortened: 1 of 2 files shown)


linera-core/src/client/mod.rs
```diff
@@ -836,7 +830,14 @@ impl<Env: Environment> Client<Env> {
...
                     remaining_event_ids = unresolved;
                 }
-                Err(_) => {
+                Err(errors) => {
+                    for (validator, error) in &errors {
+                        warn!(
+                            %validator,
+                            %error,
+                            "failed to download event certificates from validator",
+                        );
```
deuszx, line 839: This will just repeat the same error we logged in `communicate_concurrently`, right?

deuszx (approved):

PR #6399, opened 2026-05-27
Checkpoint with messages

afck:
## Motivation

Checkpoints are currently only allowed for chains without messages.

## Proposal

Enable them for chains with messages.

## Test Plan

Tests were added.

## Release Plan

## Links

- Part of #460.


linera-bridge/src/solidity/BridgeTypes.sol
```diff
@@ -920,6 +920,37 @@ library BridgeTypes {
         return value;
     }
 
+    struct Cursor {
```
ma2bd, line 923: @deuszx Does this kind of change add fees / code size to the Solidity bytecode (or is it stripped?). Should part of the protocol stay opaque to solidity?

linera-core/src/node.rs
```diff
@@ -279,6 +279,9 @@ pub enum NodeError {
     #[error("Blobs not found: {0:?}")]
     BlobsNotFound(Vec<BlobId>),
 
+    #[error("Blocks not found: {0:?}")]
+    BlocksNotFound(Vec<CryptoHash>),
```
ma2bd, line 283: haha this is not confusing at all (compared to `BlobsNotFound`)

ma2bd: (but fine ofc if this is the way it needs to be)

linera-bridge/src/solidity/BridgeTypes.sol
deuszx: Definitely. It's a difficult problem to solve – i.e. dynamically detect which parts should be stripped away and which should be included during generation, as it's all "caller dependent".

linera-core/src/node.rs
afck: I regularly mix up "blocks" and "blobs" when talking about this stuff. :sweat_smile: :dizzy_face: 
The only solution would be to find a different word for "blob".

ma2bd (approved):

ndr-ds:
FYI I just merged a couple PRs via the merge queue and the `bridge-e2e` tests were not an issue (both merged first try), so this error might be real 🤔
cc @deuszx

ma2bd:
> FYI I just merged a couple PRs via the merge queue and the `bridge-e2e` tests were not an issue (both merged first try), so this error might be real 🤔 cc @deuszx

Indeed. @deuszx I made a temporary patch to allow this PR to go through https://github.com/linera-io/linera-protocol/pull/6399/commits/eb6e9f168b63fc103cbd6a8c7c218b81a11b3804 but we're going to need to address the issue at some point

PR #6401, opened 2026-05-27
Wrap papaya hashmaps with an `Arc` – avoid cloning.

deuszx:
## Motivation

`GrpcConnectionPool::channels`, `GrpcNodeProvider::subscription_cooldowns`, and `GrpcClient::subscription_cooldowns` were all `papaya::HashMap` stored **by value**. Because all three structs `#[derive(Clone)]`, every clone called `papaya::HashMap::clone()`, which allocates a fresh `seize::Collector` and per-thread slot table — regardless of whether the map has entries.

On wasm this produces O(validators × chains) `seize::Collector` allocations — every chain subscription spawns a `GrpcClient` clone, every committee refresh calls `make_nodes` cloning `subscription_cooldowns` once
per validator. This is a suspected contributor to `JAVASCRIPT-REACT-69`; the root cause was not confirmed by reproduction.

## Proposal

Wrap each `papaya::HashMap` field in `Arc`. `Clone` becomes a refcount bump; no `seize::Collector` is allocated. The maps are now genuinely shared, which is what the existing doc comments already claimed.

**Side-effect fix**: previously `GrpcNodeProvider::make_node` passed `self.subscription_cooldowns.clone()` (by value) to each `GrpcClient`, so every client silently had its own isolated copy — cooldown state was
never actually shared across chains. Now all clients from the same provider share one map, as intended.

## Test Plan

`linera-rpc/tests/grpc_clone_alloc.rs` — two allocation-counting tests backed by a `#[global_allocator]`:

- `cloning_arc_papaya_hashmap_does_not_allocate`: 1000 `Arc<papaya::HashMap>` clones → <10 allocations
- `cloning_papaya_hashmap_by_value_allocates`: 100 bare `papaya::HashMap` clones → ≥100 allocations (documents the bug pattern; will signal if `papaya` ever internalises `Arc` itself)

`linera-rpc/tests/transport.rs` — updated call site to pass `Arc::new(papaya::HashMap::new())`.

## Release Plan

## Links


linera-rpc/src/grpc/client.rs
```diff
@@ -77,8 +77,11 @@ pub struct GrpcClient {
     max_backoff: Duration,
     /// Shared across all `GrpcClient` instances created by the same `GrpcNodeProvider`.
     /// Tracks when each validator address last had a subscription failure, so that
-    /// other chains don't independently retry the same dead validator.
-    subscription_cooldowns: papaya::HashMap<String, Instant>,
+    /// other chains don't independently retry the same dead validator. Wrapped in
+    /// `Arc` so the map is actually shared on `Clone` (and so cloning a `GrpcClient`
+    /// does not allocate a fresh `seize::Collector` per call — the alloc that drove
+    /// wasm OOMs).
```
afck: Not sure if comments should refer to past bugs.

afck (commented):
Good catch! I agree with the fix, of course!

But let's not refer to past bugs or past states of the code in comments on the current code.

(And I'm not even sure it deserves a regression test at all?)

linera-rpc/src/grpc/client.rs
ma2bd: Def not. This is AI trying to make us waste future tokens :)

linera-rpc/tests/grpc_clone_alloc.rs
```diff
@@ -0,0 +1,134 @@
+// Copyright (c) Zefchain Labs, Inc.
```
ma2bd: Indeed, this test is a bit crazy

linera-rpc/src/grpc/pool.rs
```diff
@@ -9,14 +11,17 @@ use super::{transport, GrpcError};
 #[derive(Clone, Default)]
 pub struct GrpcConnectionPool {
     options: transport::Options,
-    channels: papaya::HashMap<String, transport::Channel>,
+    // Wrapped in `Arc` so `Clone` shares the channel cache instead of allocating a
+    // fresh `papaya::HashMap` (and its `seize::Collector`) per clone. The latter
+    // exhausts wasm linear memory when many validator nodes are made per chain.
```
ma2bd: remove

deuszx: Done in [linera-rpc: remove bug-history prose from comments; drop regression test](https://github.com/linera-io/linera-protocol/pull/6401/commits/2791a7fc940c84ad03f1723e07cb09d271ce2150).

linera-rpc/tests/grpc_clone_alloc.rs
deuszx: Done in [linera-rpc: remove bug-history prose from comments; drop regression test](https://github.com/linera-io/linera-protocol/pull/6401/commits/2791a7fc940c84ad03f1723e07cb09d271ce2150).

linera-rpc/src/grpc/client.rs
deuszx: Done in [linera-rpc: remove bug-history prose from comments; drop regression test](https://github.com/linera-io/linera-protocol/pull/6401/commits/2791a7fc940c84ad03f1723e07cb09d271ce2150).

ma2bd (approved):

ndr-ds (approved):
Thanks for the fix!

PR #6408, opened 2026-05-28
Sync chain state after rejected proposal; revert proactive sync

afck:
## Motivation

https://github.com/linera-io/linera-protocol/pull/6343 fixed liveness issues by proactively downloading and uploading locked blocks. This is bad for performance and causes redundant requests.

## Proposal

- Revert the `NewRound` notification change: really only sync if the notification's round is strictly greater than ours.
- Revert the unconditional locking block upload.
- Instead, synchronize chain state from validator if they reject our proposal. Detect changes in our own chain manager after a failed proposal, and retry if appropriate.

## Test Plan

A new regression test was added to make sure that the liveness issue is still addressed.

## Release Plan

- Port to `main`.
- Release SDK.

## Links

- Performance regression introduced in #6343.


linera-core/src/client/chain_client/mod.rs
```diff
@@ -1959,14 +1996,55 @@ impl<Env: Environment> ChainClient<Env> {
...
+    ) -> Result<ClientOutcome<Option<ConfirmedBlockCertificate>>, Error> {
+        let mut snapshot = self.consensus_state_snapshot().await?;
+        let result = self
+            .process_pending_block_inner(proposal_guard, &mut snapshot)
+            .await;
+        match result {
+            Ok(outcome) => Ok(outcome),
+            Err(err) => {
+                if let Ok(current) = self.consensus_state_snapshot().await {
+                    if current != snapshot {
```
deuszx: Personally, I'd find it cleaner if `process_pending_block_inner` returned a new snapashot, alongside the error. It could be optional – `Some(_)` only when the update did happen – which would signal the caller that something might have changed and it should compare the snapshot. If it's `None` then it means that the error didn't update the state.

deuszx: Or is it just too complicated to implement it this way?

afck: That would replace every single `?` in the function… I imagine that would make it pretty unreadable?

linera-core/src/client/chain_client/mod.rs
```diff
@@ -281,6 +281,12 @@ pub enum Error {
     #[error("The state of the client is incompatible with the proposed block: {0}")]
     BlockProposalError(&'static str),
 
+    #[error(
+        "A validator absorbed new chain manager state during the proposal; \
+         retry with the refreshed local state"
+    )]
+    LocalConsensusStateAdvanced,
```
deuszx: What's confusing is that the error variant name is talking about local state that is supposdely changed (past tense of _advanced_) but the error message talks about validator's state change.

deuszx: Is that "validator" in the message about this client's knowledge about what validator has?

afck: Good point, that comment is wrong.

linera-core/src/client/chain_client/mod.rs
```diff
@@ -2125,6 +2203,11 @@ impl<Env: Environment> ChainClient<Env> {
                 }
             }
         }
+        // The local handle of our own proposal can set a Fast lock, advance
+        // `proposed`/`signed_proposal`, and thereby bump `current_round`. None of that
+        // is "new info from a validator", so fold it into the snapshot before the
+        // network calls below.
+        *snapshot = self.consensus_state_snapshot().await?;
```
deuszx: This method is over 200 lines long. I have no idea why you chose these particular places to overwrite the snapshot argument, seems arbitrary. I'd very much prefer if this method was more declarative and the places where the snapshot needs to be updated arose cleanly.

I'll have to trust the test covers this.

afck: Because `local_node.handle_block_proposal` above changes the chain manager state but should _not_ count as new info from validator. (As the comment tries to explain… but admittedly it's complicated.)

I'm not sure how to simplify this in a way that makes it clearer.

deuszx (approved):

linera-core/src/client/chain_client/mod.rs
```diff
@@ -1959,14 +1993,66 @@ impl<Env: Environment> ChainClient<Env> {
...
+            {
+                Ok(outcome) => return Ok(outcome),
+                Err(err) => err,
+            };
+            let Some(snapshot) = snapshot else {
+                return Err(err);
+            };
+            let Ok(current) = self.consensus_state_snapshot().await else {
+                return Err(err);
+            };
```
ma2bd, line 2029: Perhaps a bit more tight:
```
let current = self.consensus_state_snapshot()?;
```

And otherwise, should we at least log the error returned by `self.consensus_state_snapshot()` ?

PR #6431, opened 2026-05-29
Fix multi-leader jitter target receding instead of anchoring to round start

ndr-ds:
## Motivation

On 2026-05-29 several PM market chains froze: each was stuck in a multi-leader round, never committing a block, which crash-looped the workers via the idle-chain liveness probe. Disabling multi-leader jitter on the workers recovered them, pinpointing the jitter path.

Root cause is in `multi_leader_jitter_target` (`linera-core/src/client/chain_client/mod.rs`) added by #6338. It anchors the proposal delay to the **current time** whenever the round start cannot be derived from a round timeout:
- `ChainOwnership::round_timeout(round)` returns `None` for every multi-leader round **except the last** (`linera-base/src/ownership.rs`). The affected chains run `multi_leader_rounds = 1000`, so this is essentially every round.
- It also falls back to `now` when proposing in a round different from the manager’s current round (the conflict-bump case).

With the anchor at `now`, `propose_at = now + delay` is recomputed on every call. On a busy chain `process_inbox` runs constantly — each inbox notification interrupts the wait and re-evaluates — so the target recedes by `delay` every time and the non-preferred sole proposer **never reaches it**. `round_for_new_proposal` returns `WaitForTimeout` forever, the worker never proposes, and the chain wedges. (Observed: worker logs showing `…times out at T` where `T` is always ~`delay` ahead of the current log time, at a fixed round/height.)

## Proposal

When the round start cannot be derived from a timeout, anchor to the time the round was **first observed** (remembered per round in a new in-memory `jitter_round_anchor` field) instead of `now`. Then `propose_at` is a fixed point the clock catches up to: the worker waits out the jitter delay once and proposes. The `round_timeout − duration` derivation is unchanged where it already worked (single-leader rounds, final multi-leader round).

The pure timing logic is extracted into a free `multi_leader_jitter_target` function (the method becomes a thin adapter reading the clock + anchor) so it can be unit-tested.

**Backward compatible:** this changes only local proposal *timing* — no wire format, consensus rule, or persisted state changes. Mixed-version clients/validators interoperate. It also keeps jitter working (unlike the `--disable-multi-leader-jitter` operational stopgap currently on the workers), so that flag can be removed once this ships.

## Test Plan

- New regression test `multi_leader_jitter_target_is_anchored_not_receding`: in a non-final multi-leader round, the target stays fixed as the clock advances, and the owner proposes once the clock reaches it. **Verified it fails on the pre-fix behavior** (target recedes) and passes with the fix.
- `cargo clippy -p linera-core` clean; `cargo fmt --check` clean.

## Notes / follow-ups

- original PR: #6338
- **Single- vs multi-proposer:** the client-side first-seen anchor fully fixes the sole-proposer case (all affected PM chains). For multiple *active* proposers the staggering is approximate (each client anchors to its own first-seen time rather than a shared authoritative round start). A principled follow-up would track the round start in the chain manager and expose it via `ChainInfo` so all proposers anchor to the same value — filed separately.
- Frontport to `main` to follow (the bug exists there too).

### DIFF
--- a/linera-core/src/client/chain_client/mod.rs
+++ b/linera-core/src/client/chain_client/mod.rs
@@ -38,7 +38,7 @@ use linera_chain::{
         BlockProposal, BundleExecutionPolicy, BundleFailurePolicy, ChainAndHeight, IncomingBundle,
         ProposedBlock, Transaction,
     },
-    manager::LockingBlock,
+    manager::{ChainManagerInfo, LockingBlock},
     types::{
         Block, ConfirmedBlock, ConfirmedBlockCertificate, Timeout, TimeoutCertificate,
         ValidatedBlock,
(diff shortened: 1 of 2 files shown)


ndr-ds:
Closing for now, since we ended up just reverting the original jitter PR

PR #6447, opened 2026-06-03
[testnet] Filter client outbox indices to tracked chains

ma2bd:
## Motivation

A client's `nonempty_outboxes` and `outbox_counters` are `RegisterView`s (each re-serialized whole on every block). They only drop an entry when the local node *delivers* to that target, and a client only delivers to chains its wallet tracks (`is_full`). High-fan-out chains send to many untracked targets, whose entries are inserted but never removed, so the two blobs grow without bound and every block re-serializes a large value to local storage — stalling block production. Validators (which track all chains) keep these indices near-empty.

## Proposal

Keep the two indices filtered to tracked targets on clients (validators, `chain_modes == None`, are unchanged):

- `process_outgoing_messages` still schedules every target's outbox *queue*, but only indexes/counts targets the wallet tracks (new `tracked: Option<&BTreeSet<ChainId>>` parameter).
- New `ChainStateView::outbox_index_tracked_hash: RegisterView<Option<CryptoHash>>` (appended **last** to keep storage-key prefixes stable) stamps the hash of the tracked set; `reconcile_outbox_index` rebuilds the indices **additively** from the tracked chains' retained outbox queues when the hash changes — `O(|tracked|)`, no scan of the full fan-out, correct for migration / restart / shrink / growth.
- `mark_messages_as_received` only enforces the "counter must be present" invariant for validators and tracked targets, tolerating untracked ones (fixes a `CorruptedChainState` reachable when a target is un-tracked between sending an update and receiving its confirmation).
- The chain worker derives the tracked set from `chain_modes` and threads it through block application and `confirm_updated_recipient`.

Outboxes are local node state (not in the consensus `state_hash`), so validators and clients legitimately differ here; the change is consensus-safe.

A subtractive `O(|delta|)` fast path can be layered on later once the worker supplies the added/removed sets; until then the additive `O(|tracked|)` rebuild is used.

## Test Plan

- `linera-chain` unit tests: reconcile migration / shrink / re-track / no-op; un-track race tolerated; corruption still caught for tracked targets.
- `linera-core` worker-level test: full `confirm_updated_recipient` → `mark_messages_as_received` path for an untracked recipient (regresses the `CorruptedChainState`).
- `cargo clippy --all-targets --all-features` clean on `linera-chain` and `linera-core`; full `linera-chain` suite green.

## Release Plan

- These changes add a trailing field to `ChainStateView` (storage-format change) and so require a new deployment. Targeting `testnet_conway`.

## Links

### DIFF
--- a/linera-core/src/chain_worker/state.rs
+++ b/linera-core/src/chain_worker/state.rs
@@ -400,6 +400,37 @@ where
         })
     }
 
+    /// Returns the set of chains tracked in full mode, or `None` on a validator (which tracks all
+    /// chains and never filters its outbox indices).
+    fn tracked_full_chains(&self) -> Option<BTreeSet<ChainId>> {
+        let chain_modes = self.chain_modes.as_ref()?;
+        let chain_modes = chain_modes
+            .read()
+            .expect("Panics should not happen while holding a lock to `chain_modes`");
+        Some(
+            chain_modes
+                .iter()
+                .filter(|(_, mode)| mode.is_full())
+                .map(|(id, _)| *id)
+                .collect(),
+        )
+    }
+
+    /// Reconciles the chain's `nonempty_outboxes`/`outbox_counters` indices with the current set
+    /// of fully-tracked chains, and returns that set to be passed to block application so that
+    /// newly scheduled messages are only indexed for tracked targets. Returns `None` on a
+    /// validator (no filtering).
+    async fn reconcile_tracked_outboxes(&mut self) -> Result<Option<BTreeSet<ChainId>>, WorkerError> {
+        let Some(full_chains) = self.tracked_full_chains() else {
+            return Ok(None);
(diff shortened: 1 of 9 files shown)


linera-core/src/chain_worker/state.rs
ma2bd: Oh nice, I think we also don't need this any more!

ma2bd: Done in 3a72be73e6 — removed the loop entirely. With both indices now tracked-only, the `outbox_counters` fast path (`all_messages_delivered_up_to`) is already the complete answer for clients too, so the whole function collapsed to that call and I inlined it at the `confirm_updated_recipient` call site. clippy clean; `linera-chain` lib (51) + the confirm/cross-chain `linera-core` worker tests green.

PR #6448, opened 2026-06-04
Drive the client's coordination time off the node clock

ma2bd:
## Motivation

The client-side coordination layer in `linera-core` — the requests scheduler, the
notification circuit breaker, and `communicate_concurrently` — measured elapsed time and slept
using the wall clock (`std::time::Instant` / `tokio::time::sleep`) rather than the node's
`Storage::Clock`. That is a second, uncontrollable notion of time, separate from the one the
worker/consensus already run on (`TestClock`). As a result these paths cannot be simulated
deterministically, and their tests are flaky and slow — e.g. `test_staggered_parallel_retry_on_failure`
was previously "fixed" by inflating real delays (#5525).

This is the first step toward making `linera-core` a self-contained, fully
simulatable implementation of the protocol (fake scheduler / clock / network).

## Proposal

Route all client-side coordination time through the existing `Storage::Clock`, so the whole
node shares one (simulatable) notion of time:

- Add `ClockOf<Env>` plus two small helpers in `linera-core::client`: `duration_to_delta` and
  `sleep_for(clock, duration)`, which builds a relative sleep on top of `Clock::sleep_until`.
- **Requests scheduler**: holds a clock handle; latency-EMA measurement, staggered retries,
  cache TTL and in-flight TTL now read it. `RequestsCache` and `InFlightTracker` take
  `now: Timestamp` instead of calling `Instant::now()` internally.
- **Notification circuit breaker**: probe scheduling uses `Timestamp` from the node clock.
- **`communicate_concurrently`**: now uses a hedged, **failure-responsive** fan-out. Moving it
  onto a (manually-stepped) clock surfaced a pre-existing flaw: it launched every node on a fixed
  staggered schedule and *waited out that schedule even after earlier nodes had already failed*.
  It now starts the next node immediately on a failure, and only sleeps the hedge delay to race
  an extra request against a node that is still in flight (the slow request is hedged, not
  cancelled). Since the value is a hedge delay, not a timeout, the misnamed `timeout` is renamed
  to `hedge_delay` consistently — the function parameter, the scheduler's download-method
  parameters, the `chain_client::Options` fields (`*_download_timeout` →
  `*_download_hedge_delay`), and the CLI flags (`--blob-download-timeout-ms` →
  `--blob-download-hedge-delay-ms`, `--cert-batch-download-timeout-ms` →
  `--cert-batch-download-hedge-delay-ms`); `CLI.md` is regenerated accordingly. (The CLI flags
  are renamed because the project is pre-1.0 and the help text already described hedging.)

- **Unify the two staggered fan-out primitives.** `communicate_concurrently` and the scheduler's
  `try_staggered_parallel` were two copies of the same "race attempts, hedge the slow one, fail
  over fast" loop. They now share one generic engine, `hedged_fan_out`, and each caller is a thin
  adapter that supplies only what differs: the peer source (a shuffled slice vs. the in-flight
  tracker's dynamic alternative queue), the error reduction (collect all vs. keep the last), and
  the hedge schedule. The duplicated loop and a needless `AtomicU32` counter are gone (net −60
  lines in this commit).

No new abstraction at the `Environment` level — this reuses the clock the worker already uses. In
production (`WallClock`) behaviour is preserved: the hedge schedule is passed per-caller as
`hedge_schedule: impl Fn(usize) -> Duration`, kept at `|k| staggered_delay·k` for the scheduler
and `|k| hedge_delay·k²` for `communicate_concurrently` to match `main`. The one intended
behaviour change versus `main` is that `communicate_concurrently` no longer idles after a failure
(it fails over immediately, strictly faster).

## Test Plan

- Scheduler/cache unit tests converted to virtual time (`TestClock`); they no longer sleep on
  real time and now run in ~0.04s.
- `test_staggered_parallel_retry_on_failure` (previously flaky, see #5525) rewritten to assert
  on retry *order* under auto-advancing virtual time — deterministic.
- New unit tests for `communicate_concurrently` assert it fails over to a working node and
  collects all errors **without advancing the (frozen) clock** — i.e. it never waits after a
  failure and cannot deadlock under a manually-stepped clock.
- New `hedged_fan_out` unit tests drive the shared fan-out engine directly on a manually-stepped
  `TestClock`, covering the hedging paths the wall clock made untestable: the slow-peer hedge
  fires and does **not** cancel the original attempt, the hedge schedule alone determines each
  peer's start time, a frozen clock disables hedging, and a failure fails over without waiting
  out the hedge.
- Integration subset exercising the affected paths (transfers with notifications, too-many-faults,
  voting-rights changes, finalize-validated, message policy, follow-chain) passes locally in ~2s.
- `cargo fmt` / `cargo clippy -p linera-core --tests` / `cargo check -p linera-core --tests` clean.

## Release Plan

## Links

- Part of a larger effort toward a fake-scheduler-simulatable `linera-core` (subsequent steps:
  introduce a spawn seam, move the chain-listener into core).


linera-client/src/client_options.rs
```diff
@@ -208,20 +208,20 @@ pub struct Options {
...
     /// The delay when downloading a batch of certificates, after which we try a second validator,
     /// in milliseconds.
     #[arg(
-        long = "cert-batch-download-timeout-ms",
+        long = "cert-batch-download-hedge-delay-ms",
         default_value = "1000",
         value_parser = util::parse_millis
     )]
-    pub certificate_batch_download_timeout: Duration,
+    pub certificate_batch_download_hedge_delay: Duration,
```
afck, line 224: (Not in this PR, but: I'm not sure these two options need to be separate at all. It doesn't really matter whether it's a blob or certificates that we're trying to download.)

linera-core/src/client/chain_client/mod.rs
```diff
@@ -3237,6 +3237,9 @@ impl<Env: Environment> ChainClient<Env> {
             .options
             .notification_circuit_breaker_initial_probe_interval;
         let max_probe_interval = self.options.notification_circuit_breaker_max_probe_interval;
+        // Read the (possibly simulated) node clock once, so circuit-breaker probe scheduling can
+        // be driven deterministically in tests instead of depending on the wall clock.
```
afck, line 3241: I don't think we need that comment in every line where we call `current_time`.

linera-core/src/client/mod.rs
```diff
@@ -2365,41 +2376,139 @@ impl<'a, Env: Environment> EventSetDownloader<'a, Env> {
...
+/// either *immediately* when an in-flight attempt fails (no point waiting once we know an attempt
+/// failed), or after the in-flight attempt has run for `hedge_schedule(started)` without answering
+/// — hedging against a slow peer by racing an extra request; the slow attempt is **not** cancelled.
+/// `hedge_schedule` maps the number of attempts started so far to the delay before starting the
+/// next one (e.g. `|k| delay * k` for a linearly-growing stagger). All sleeping is on `clock`, so
+/// this runs in (possibly simulated) time; freezing the clock disables the slow-peer hedge while
+/// still advancing on every failure.
+pub(crate) async fn hedged_fan_out<Peer, T, Err, NextPeer, NextFut, Op, OpFut>(
+    first_peer: Peer,
+    mut next_peer: NextPeer,
```
afck, line 2399: Why not just `impl FnMut() -> NextFut`? Is there a reason some are explicit and in the `where` clause, and some use `impl`?

afck (approved):

afck:
Actually, are all the `client_*tests` and `worker_*tests` still passing deterministically with this?
I imagine that some of them rely on the fact that one faulty validator is tolerated, but if `communicate_concurrently` now picks the faulty one, and the test doesn't explicitly advance the clock, it might fail?

ma2bd:
> Actually, are all the `client_*tests` and `worker_*tests` still passing deterministically with this? I imagine that some of them rely on the fact that one faulty validator is tolerated, but if `communicate_concurrently` now picks the faulty one, and the test doesn't explicitly advance the clock, it might fail?

I don't see what could be less deterministic, do you?

afck:
> I don't see what could be less deterministic, do you?

If a test has 3 honest and 1 faulty validators, and a client there tries to download something, the test could rely on the client succeeding: If the first attempt was the faulty validator, it would retry and download from an honest one.

But now it won't necessarily retry anymore, will it? It will wait for the test clock to advance before retrying?

Anyway, seems to work fine. We'll see if any test becomes flaky.

PR #6457, opened 2026-06-05
Forward-port reactive chain-state sync on rejected proposals (#6432 + #6454) to main

ma2bd:
## Motivation

The reactive "lazy pull" chain-state sync on proposal rejection — and its snapshot-based retry refactor of `process_pending_block_without_prepare` — lives only on `testnet_conway` (#6408, reapplied as #6432). `main` still uses the older proactive-sync model. This forward-ports that work, together with the follow-up reliability fix #6454, so the two branches share the same consensus-client sync behavior.

This was prompted by review of #6454: that fix only makes sense on top of #6432, which `main` lacked, so the whole rework is ported as a unit.

## Proposal

Two changes, ported faithfully from `testnet_conway`:

1. **#6432 — reactive sync on rejection + snapshot retry.**
   - `updater.rs`: on a `ChainError` proposal rejection, pull manager values from that validator (`synchronize_chain_state_from`) so the local node absorbs whatever justified the rejection; tighten `sync_consensus_round` to push only the *current-round* locking certificate (`validated.round == target_round`).
   - `chain_client/mod.rs`: split `process_pending_block_without_prepare` into a retry `loop` + `process_pending_block_inner`, comparing a `consensus_state_snapshot()` baseline so a per-validator updater absorbing new state triggers a retry.
   - This reverts the proactive locking-block sync from #6343 in favor of the reactive path.

2. **#6454 — fallback sync.** When a rejected proposal did not advance the local consensus snapshot (the lazy pull was raced out by `communicate_with_quorum` dropping in-flight futures), fall back **once** to an explicit quorum `synchronize_chain_state` — guaranteed by quorum intersection to reach a lock-holder — and retry if it absorbed anything. Bounded to at most one fallback per call.

### Notes on the port

- The cherry-pick of #6432 auto-merged the production code; `main`'s independent owner-key / `outcome_matches` logic in the propose path was preserved.
- `main`'s divergent clock-skew (`InvalidTimestamp`) handling in `send_block_proposal` is unrelated to #6432 and is kept as-is.
- #6432 replaced two proactive-sync tests with `test_lazy_pull_absorbs_locking_block_on_proposal_rejection`. Both old tests are removed here too: `test_sync_consensus_round_pushes_locking_below_current_round` was kept temporarily and confirmed to **fail** under the new behavior (it asserts a below-current-round lock gets pushed, which #6432 deliberately stops doing), then removed — matching #6432 exactly.
- #6448 (a test-only clock seam, foundation for the not-yet-written deterministic test #6453) is **not** included; it isn't needed for the production fix and isn't on `testnet_conway`.

## Test Plan

- `cargo test -p linera-core --lib` (memory): **145 passed, 0 failed**.
- Targeted: `test_lazy_pull_absorbs_locking_block_on_proposal_rejection` ✅ and `test_finalize_validated` ✅.
- `cargo clippy -p linera-core --all-targets` clean; `cargo fmt` clean.
- The ported `process_pending_block_without_prepare` is byte-identical to `testnet_conway`'s plus exactly the #6454 fallback; the `updater.rs` lazy-pull branch is identical to `testnet_conway`.

## Release Plan

- These changes follow the usual release cycle. (Already on `testnet_conway` via #6432; #6454 is its in-review reliability follow-up.)

## Links

- Ports: #6432 (reapply of #6408) and #6454
- Related: #6430 (revert of #6408), #6343 (the proactive sync this reverts), #6448 / #6453 (deferred test infra)

### DIFF
--- a/linera-core/src/client/chain_client/mod.rs
+++ b/linera-core/src/client/chain_client/mod.rs
@@ -142,6 +142,16 @@ struct CircuitBreakerState {
     probe_interval: Duration,
 }
 
+/// A fingerprint of the chain manager's observable consensus state, used to detect
+/// whether a per-validator updater absorbed new info during a proposal attempt.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+struct ConsensusStateSnapshot {
+    next_block_height: BlockHeight,
+    current_round: Round,
+    lock_round: Option<Round>,
+    timeout_round: Option<Round>,
+}
+
 #[cfg(with_testing)]
 impl Options {
     pub fn test_default() -> Self {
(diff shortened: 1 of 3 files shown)


linera-core/src/client/chain_client/mod.rs
```diff
@@ -1911,14 +1945,99 @@ impl<Env: Environment> ChainClient<Env> {
...
+                // the lazy pull to miss needs a clock-driven per-validator response delay in the test
+                // harness (building on #6448); until then it is only covered indirectly by the (now
+                // non-flaky) `test_lazy_pull_absorbs_locking_block_on_proposal_rejection`.
+                if did_fallback_sync {
+                    return Err(err);
+                }
+                did_fallback_sync = true;
+                if self.synchronize_chain_state(self.chain_id).await.is_err() {
+                    return Err(err);
+                }
```
deuszx: This is different from the one on `testnet_conway`
```rust
if let Err(error) = self.synchronize_chain_state(self.chain_id).await {
  tracing::debug!(%error, "fallback sync failed after rejected proposal"); 
  return Err(err);
}
```

deuszx (approved):

PR #6470, opened 2026-06-08
Validators sign over `BlockHeader`, not the whole serialized `Block`.

deuszx:
## Motivation

A `BlockHeader` already commits to every body field through per-field hashes, yet a certificate signed the hash of the *whole* serialized block. Verifying a block therefore meant relaying its entire body — defeating the point of those commitments. Signing the header's hash instead lets a party holding only the header verify the block and prove any single body field against it, without the rest of the body. This is what makes a lean cross-chain block proof possible, which the EVM light client then builds on.

## Proposal

This branch lands the change end to end, in three layers:

- **Core commitment (this PR).** Certificates sign `CryptoHash::new(&header)` rather than the whole block. The wire format is unchanged — the header is recomputed from the body on deserialize, so a tampered body yields a different header hash and
fails signature verification. A `BlockHeader::verifies(field)` API checks a single body field against its matching header hash, giving holders of just the header a per-field inclusion proof.

- **EVM bridge adoption (#6472).** The on-chain light client verifies a block from its header and signatures alone — never its body — then records the block's `events_hash`. Callers prove individual events against that registered commitment, so
burns settle in chunks without re-verifying the certificate or shipping the block body.

- **Committee rotation from events (#6482).** Validator-set rotations are detected from the system epoch event Linera emits, not by parsing operations out of a transaction. Installing a new committee reuses the *same* register-then-prove path as
burns: register the admin block, then prove the single epoch event's inclusion against it. One inclusion primitive serves both flows, and dropping the transaction-parsing path removes a large type subtree from the bridge's EVM codegen.

## Test Plan

- `linera-chain` unit tests reproduce a certificate's signed hash from the header alone and prove a body field via `BlockHeader::verifies` without the rest of the body; storage round-trips run through header-consistent blocks.
- Bridge coverage (Solidity + revm + the `committee_rotation` e2e) lives in #6472 / #6482.
- CI runs the full suite across the touched crates.

## Release Plan

- These changes follow the regular release cycle.

## Links

- Folds in #6472 (EVM bridge: header-verified block proofs) and #6482 (committee rotation from events).
- [reviewer checklist](https://github.com/linera-io/linera-protocol/blob/main/CONTRIBUTING.md#reviewer-checklist


linera-storage/src/db_storage.rs
```diff
@@ -1766,39 +1815,7 @@ mod tests {
         // Create a test certificate at a specific height
         let chain_id = ChainId(CryptoHash::test_hash("test_chain"));
         let height = BlockHeight(5);
-        let block = Block {
```
ma2bd, line 1955: So what was the problem with this data-structure when it comes to proving the value of a body field?

deuszx: The signature was over the serialized `Block`, not over the `BlockHeader` (the same `Serialize for Block` was used for storage/wire and the proving)

ma2bd: Why not change just that?

deuszx: This is exactly what I'm doing here.

linera-base/src/hashed.rs
```diff
@@ -30,6 +30,13 @@ impl<T> Hashed<T> {
         Self { value, hash }
     }
 
+    /// Creates a [`Hashed`] from a value and a precomputed hash, without recomputing it.
+    ///
+    /// The caller is responsible for the hash being the canonical hash of `value`.
+    pub fn with_hash(value: T, hash: CryptoHash) -> Self {
```
ma2bd, line 36: I would call it `new_unchecked` or `with_hash_unchecked`

ma2bd: It feels a bit wrong to use `Hashed` in the first place if you need this method but maybe that's fine. After all, there is no `verify` method.

deuszx: The idea is that whatever is inside can have a hash computed and it's memoized.

ma2bd: Ok. In that case, perhaps we want the container to be agnostic then: `new` should be rename `with_crypto_hash` and the manual hash (yours) `with_unchecked_hash`.

deuszx: The container is agnostic – it accept any type as long as it implements `BcsHashable`. It's a textbook smart constructor. I will rename the new method to `with_unchecked_hash`.

ma2bd: If the constructor is called `new` then it looks like the default one for this container.

linera-chain/src/block.rs
```diff
@@ -29,14 +29,26 @@ use crate::{
...
 
 /// Wrapper around a `Block` that has been validated.
-#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize, Allocative)]
-#[serde(transparent)]
+#[derive(Debug, PartialEq, Eq, Clone, Allocative)]
 pub struct ValidatedBlock(Hashed<Block>);
 
+impl Serialize for ValidatedBlock {
+    fn serialize<S: serde::Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        self.0.serialize(serializer)
```
ma2bd, line 37: Note that this will be seen as `Block` instead of a `ValidatedBlock` by `serde-reflection` (same below).

linera-chain/src/block.rs
```diff
@@ -75,10 +87,21 @@ impl ValidatedBlock {
...
 
 /// Wrapper around a `Block` that has been confirmed.
-#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize, Allocative)]
-#[serde(transparent)]
+#[derive(Debug, PartialEq, Eq, Clone, Allocative)]
 pub struct ConfirmedBlock(Hashed<Block>);
 
+impl Serialize for ConfirmedBlock {
+    fn serialize<S: serde::Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        self.0.serialize(serializer)
```
ma2bd, line 95: same. Apparently, this was the case already

linera-storage/src/db_storage.rs
```diff
@@ -1678,6 +1684,49 @@ mod tests {
         DbStorage, Storage, TestClock,
     };
 
+    /// Builds a block populated with one item of each body kind, with values derived from the
+    /// height so blocks are distinct. The header is computed from the body via `Block::new`, so
+    /// the block round-trips through storage (the block hash commits to that header).
+    fn populated_block(chain_id: ChainId, height: u64) -> Block {
```
ma2bd, line 1876: isn't this for testing only?

ma2bd (approved):

linera-chain/src/block.rs
deuszx: Yes, I think it was due to `#[serde(transparent)]` marking.

linera-storage/src/db_storage.rs
deuszx: It is but this is inside `mod test`.

linera-chain/src/block.rs
deuszx: I guess that's why we also include the `CertificateKind` in `VoteValue` (apart from the `CryptoHash`) – to differentiate between validated* and confirmed* hashes for the same `Block`.

linera-base/src/hashed.rs
deuszx: Right, but I think `Hashed::new` is still the default constructor for this container. It's just that for the two types (confirmed and validated certificates) we have to do it differently.

afck (commented):
Wouldn't it be better to use something other than `T: BcsHashable<'de>` instead?
* This solution here changes the meaning of `Hashed<T>`: Up to now it was really always meant to contain the hash of `T` as a `BcsHashable`. Now it can instead contain a "user-defined hash" of `T` that disagrees with it!
* If we either changed the `BcsHashable` impl for `Block` (I don't think that's possible with the current trait?) or used a new, different `LineraHashable` trait (equals `BcsHashable` for all other types, but for `Block` it's only the hash of the _header_), the `Hashed<T>` type would semantically make more sense to me.

deuszx:
> Now it can instead contain a "user-defined hash" of T that disagrees with it!

Yes, if you use the "unchecked" constructor but dangerous things can happen in many places if developer doesn't read the docs. For this particular case, it's impossible to construct `Hashed<ConfirmedBlocKCertificate>` any other way (b/c there's no `BcsHashable for Block`).


This diff was the smallest one that achieved the goal – we can iterate on the implementation details in the future I guess w/o breaking the protocol now

PR #6471, opened 2026-06-09
Label block-execution metrics by phase

ndr-ds:
## Motivation

Block execution happens in three protocol-distinct contexts, but they all recorded into the *same* Prometheus time series, so a latency or resource sample gave no way to tell which path produced it:

- **staging a block proposal** — a proposer building its own block,
- **handling a received block proposal** — a validator validating it,
- **handling a confirmed certificate** — a validator committing it.

All three funnel through `ChainStateView::execute_block`, where ten execution metrics are recorded with no discriminating label (`&[]`).

## Proposal

Add a `phase` label to the ten execution metrics, with values `stage_proposal` / `handle_proposal` / `handle_confirmed`:

`block_execution_latency`, `num_blocks_executed`, `wasm_fuel_used_per_block`, `evm_fuel_used_per_block`, `vm_num_reads_per_block`, `vm_bytes_read_per_block`, `vm_bytes_written_per_block`, `message_execution_latency`, `operation_execution_latency`, `state_hash_computation_latency`.

A new `BlockExecutionPhase` enum (no `Default`, exactly three variants) is threaded as a **required** argument through `execute_block` -> `execute_block_inner` -> `BlockExecutionTracker`, plus the worker-level `execute_block` wrapper. Because it is required, a new execution path cannot compile without naming a phase — there is no ambient default that could silently mislabel a sample. The three call sites set it explicitly:

- `stage_block_execution` -> `StageProposal`
- `try_handle_block_proposal` -> `HandleProposal`
- `process_confirmed_block` -> `HandleConfirmed`

The unrelated outbox metrics (`num_outboxes`, `outbox_counters_size`) are untouched. The `phase` field on `BlockExecutionTracker` is compiled out when metrics are disabled.

Existing flat queries keep working (they aggregate over `phase`); new dashboards can split per phase, e.g.:

```
histogram_quantile(0.99, sum by (le, phase) (rate(block_execution_latency_bucket[5m])))
```

## Test Plan

- `cargo clippy -p linera-core -p linera-chain --features "metrics test" --all-targets` — clean, 0 warnings.
- `cargo build -p linera-chain -p linera-core` (default features, i.e. `with_metrics` off) — clean; confirms the cfg-gated `phase` field path compiles.
- `cargo test -p linera-chain --features "metrics test" --lib` — 51 passed (exercises `stage_proposal`).
- `cargo test -p linera-core --features "metrics test" --lib test_handle_block_proposal test_handle_certificate_*` — 13 passed (exercises `handle_proposal` and `handle_confirmed`).

Prometheus panics at runtime on a label-count mismatch, so the suites passing under `--features metrics` is direct evidence the labels are wired consistently across all three paths. Not yet verified by scraping `/metrics` on a running node.

## Release Plan

- Nothing to do / These changes follow the usual release cycle. (Metrics-only change; no protocol or storage format change.)

## Links

### DIFF
--- a/linera-chain/src/block_tracker.rs
+++ b/linera-chain/src/block_tracker.rs
@@ -23,7 +23,7 @@ use tracing::instrument;
 #[cfg(with_metrics)]
 use crate::chain::metrics;
 use crate::{
-    chain::EMPTY_BLOCK_SIZE,
+    chain::{BlockExecutionPhase, EMPTY_BLOCK_SIZE},
     data_types::{
         IncomingBundle, MessageAction, OperationResult, PostedMessage, ProposedBlock, Transaction,
     },
@@ -35,6 +35,10 @@ use crate::{
 #[derive(Debug)]
 pub struct BlockExecutionTracker<'resources, 'blobs> {
     chain_id: ChainId,
+    /// The protocol phase this block is executed in (staging a proposal, validating a
+    /// received proposal, or executing a confirmed certificate). Recorded on execution
+    /// spans and used to label execution metrics.
+    phase: BlockExecutionPhase,
     block_height: BlockHeight,
     timestamp: Timestamp,
     authenticated_owner: Option<AccountOwner>,
@@ -78,6 +82,7 @@ impl<'resources, 'blobs> BlockExecutionTracker<'resources, 'blobs> {
         local_time: Timestamp,
         replaying_oracle_responses: Option<Vec<Vec<OracleResponse>>>,
         proposal: &ProposedBlock,
+        phase: BlockExecutionPhase,
     ) -> Result<Self, ChainError> {
         resource_controller
             .track_block_size(EMPTY_BLOCK_SIZE)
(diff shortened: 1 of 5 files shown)


linera-chain/src/block_tracker.rs
```diff
@@ -35,6 +35,10 @@ use crate::{
 #[derive(Debug)]
 pub struct BlockExecutionTracker<'resources, 'blobs> {
     chain_id: ChainId,
+    /// The protocol phase this block is executed in. Only used to label execution metrics,
+    /// so it is compiled out when metrics are disabled.
+    #[cfg(with_metrics)]
+    phase: BlockExecutionPhase,
```
ma2bd: Just include it all the time? We could probably use this information for some of the execution logs

linera-chain/src/chain.rs
```diff
@@ -1201,6 +1248,7 @@ where
         published_blobs: &[Blob],
         replaying_oracle_responses: Option<Vec<Vec<OracleResponse>>>,
         policy: BundleExecutionPolicy,
+        phase: BlockExecutionPhase,
```
afck, line 1251: (This can now represent a lot of illegal inputs. I wonder if we should add some assertions or merge some types at some point, but I'm not sure how:
* The oracle responses are always `None` in `StageProposal`, and always `Some` in `HandleConfirmed`.
* The on-failure policy is always `Abort` in `HandleProposal` and `HandleConfirmed`.

Anyway, needs discussion first, and can be done in a later PR.)

afck (approved):

PR #6472, opened 2026-06-09
Verify block proofs from the header on the EVM bridge

deuszx:
## Motivation

With #6470 a confirmed block's signed hash is the `BlockHeader` hash, which already commits to every body field. The EVM bridge can therefore verify a block from just the header and the validator signatures, plus whichever body field it actually needs (events for burns, transactions for committee changes), instead of relaying and re-hashing the whole `ConfirmedBlockCertificate`. This shrinks the on-chain payload and makes per-burn settlement cheap.

## Proposal

- Add `BlockProof` — a confirmed block reduced to its signed commitment: the header, round, and validator signatures. It is derived from a `ConfirmedBlockCertificate`; the body never travels in the proof.
- The EVM light client recomputes the block hash as `keccak256("BlockHeader::" ++ BCS(header))`, verifies the committee quorum's signatures over it, and checks the body fields it needs against the header's
per-field hashes — reimplementing Linera's `hash_vec` / `hash_vec_vec` folds in Solidity.
- Body fields travel as separate arguments and are checked against the header: `addBlock` / `verifyBlockFromEvents` take per-event BCS plus per-transaction event counts and recompute `events_hash`; `addCommittee`
takes per-transaction BCS and recomputes `transactions_hash`. No block body is relayed.
- `registerBlock` + `verifyEventInclusion`: register a block once (verify the header quorum, store `events_hash`/height/chain id), then settle individual burns with a cheap per-event inclusion proof folded
against the registered `events_hash`. This replaces the old chunked `processBurns`, which re-ran full certificate verification (~2.49M gas) and re-hashed every event on each chunk. `addBlock` keeps the
whole-block fast path.
- `EvmLightClient::add_committee` takes the certificate and derives the proof internally; the committee relay and the e2e tests send a `BlockProof` instead of a full certificate.
- `BridgeTypes.sol` is regenerated to carry the lean `BlockProof`; the serialized-proof argument is named `blockProof` and the signed value `blockHash`.

## Test Plan

- CI

## Release Plan

- Backporting is not possible but we may want to deploy a new `devnet` and release a new SDK soon.

## Links

- Depends on #6470 (block header commitment) — this PR stacks on it and targets its branch.


deuszx:
This now passes bridge-e2e tests https://github.com/linera-io/linera-protocol/actions/runs/27374903991

linera-bridge/src/relay/mod.rs
```diff
@@ -480,24 +479,34 @@ async fn serve_loop<E: linera_core::environment::Environment + 'static>(
...
-                        let scan_upto = linera_base::data_types::BlockHeight(height.0 + 1);
-                        if let Err(e) = committee::catch_up(
-                            chain_client.storage_client(),
-                            &evm_client,
-                            admin_chain_id,
-                            scan_upto,
-                            max_retries,
-                        )
-                        .await
+                        tracing::debug!(%height, "New admin chain block, checking for committee update");
```
deuszx: Changes below are wrong – bad rebase. revert

linera-bridge/src/monitor/linera.rs
```diff
@@ -174,17 +176,43 @@ async fn submit_addblock<P: Provider>(
...
+    cert: &ConfirmedBlockCertificate,
     height: BlockHeight,
     by_tx: &[(u32, Vec<u32>)],
     max_retries: u32,
 ) {
+    // Chunked settlement proves each burn against the block's registered `events_hash`, so the
+    // block must be registered before any gas estimate or `processBurns` — both revert otherwise.
+    if let Err(error) = evm_client.register_block(cert).await {
+        tracing::warn!(?height, ?error, "registerBlock submission failed");
+        let to_bump: Vec<u32> = {
```
afck, line 135: (Type annotation)

linera-bridge/src/evm/light_client.rs
```diff
@@ -269,175 +255,156 @@ mod tests {
...
-                &mut light_client.db,
-                light_client.deployer,
-                light_client.contract,
-                &expireEpochsBelowCall { newMinEpoch: 1 },
-            )
-            .is_err(),
-            "cannot expire at epoch 0"
-        );
+        // `eventBcs` are exactly the events at `positions` within `tx_index`.
+        let good_bcs: Vec<Bytes> = positions
```
afck, line 392: (Type annotation; also below)

linera-bridge/src/relay/evm.rs
```diff
@@ -161,60 +178,125 @@ impl<P: Provider> EvmClient<P> {
...
-        let estimate = bridge.addBlock(cert_bytes.into()).estimate_gas().await;
-        tracing::debug!(?estimate, cert_size, "addBlock gas estimate");
-        estimate
+        bridge
+            .addBlock(proof, event_bcs, events_per_tx)
+            .estimate_gas()
+            .await
+    }
+
+    /// Registers a block on the LightClient from its header and signatures alone, so its events can
```
afck, line 111:
```suggestion
    /// Registers a block on the `LightClient` from its header and signatures alone, so its events can
```

Or:

```suggestion
    /// Registers a block on the light client from its header and signatures alone, so its events can
```

(Or "Light Client" if we consider it a proper name, I guess.)

linera-bridge/src/relay/evm.rs
```diff
@@ -161,60 +178,125 @@ impl<P: Provider> EvmClient<P> {
...
+        tx_index: u32,
+        positions: &[u32],
+    ) -> ProcessBurnsArgs {
+        let events = &cert.block().body.events;
+        let proof = EventInclusionProof::new(events, tx_index as usize, positions);
+        let event_bcs = positions
+            .iter()
+            .map(|p| {
+                Bytes::from(
+                    bcs::to_bytes(&events[tx_index as usize][*p as usize])
```
afck: Would it simplify things if the proven events were part of the `BlockProof`? It seems one is not useful without the other?

deuszx: I'll consider that after #6482 where the committee rotation is also proved via events.

PR #6474, opened 2026-06-09
linera-storage: cache immutable block-hash, event, and network-description reads (testnet_conway)

ndr-ds:
## Motivation

`DbStorage` pays a full Scylla/RocksDB round-trip on every call to several read
methods that return permanently immutable data. On the OVH `testnet_conway`
validators this is a live source of read pressure (sustained multi-second p99
read spikes).

This is the `testnet_conway` backport of #6389, restricted to the three caches
that need **no new storage index** — so it ships now without waiting on the
`EventBlockHeight` index + RPC backport (#5710), which the fourth cache depends on.

## Proposal

Add three caches to `StorageCaches`, mirroring the existing five:

- `block_hash_by_height: Arc<ValueCache<(ChainId, BlockHeight), CryptoHash>>` —
  backs `read_certificate_hashes_by_heights` (uses the existing `BlockByHeight` index)
- `network_description: Arc<OnceLock<NetworkDescription>>` — single-value, zero
  overhead after first read
- `read_events_from_index` now checks the existing `event` cache before bulk-reading values

Read path: check-then-populate (same pattern as `read_certificates_raw`). Write
path: `block_hash_by_height` is populated in `write_blobs_and_certificate` at write time.

One new CLI flag with a conservative default (1 000), overridden via lineractl/linera-infra:
- `--block-hash-by-height-cache-size`

New Prometheus counter `read_block_hash_by_height` (labeled `[cache]`/`[db]`);
`read_network_description` gains the same split on its existing counter.

Deferred to a follow-up (needs the #5710 index + RPC backport first): the
`event_block_height` cache (the 4th cache in #6389).

## Test Plan

- `cargo clippy --locked --all-targets --all-features --workspace --exclude linera-web`
  clean (matches CI's host-clippy invocation; `linera-web` is wasm-only and built
  by CI's dedicated wasm job)
- Covered by existing integration tests exercising `read_certificate_hashes_by_heights`
  and `read_events_from_index`

## Release Plan

- Backport to be released in a validator hotfix.

## Links

- Main version: #6389
- Deferred index/RPC backport (blocks the 4th cache): #5710

### DIFF
--- a/linera-bridge/tests/e2e/src/lib.rs
+++ b/linera-bridge/tests/e2e/src/lib.rs
@@ -549,9 +549,11 @@ where
     E: linera_core::environment::Environment,
 {
     use linera_execution::Operation;
-    let bytes = bcs::to_bytes(&wrapped_fungible::WrappedFungibleOperation::RegisterAuthorizedCaller {
-        app_id: bridge_app_id,
-    })?;
+    let bytes = bcs::to_bytes(
+        &wrapped_fungible::WrappedFungibleOperation::RegisterAuthorizedCaller {
+            app_id: bridge_app_id,
+        },
+    )?;
     chain_client
         .execute_operations(
             vec![Operation::User {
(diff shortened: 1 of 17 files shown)


linera-bridge/tests/e2e/src/lib.rs
```diff
@@ -549,9 +549,11 @@ where
     E: linera_core::environment::Environment,
 {
     use linera_execution::Operation;
-    let bytes = bcs::to_bytes(&wrapped_fungible::WrappedFungibleOperation::RegisterAuthorizedCaller {
-        app_id: bridge_app_id,
-    })?;
+    let bytes = bcs::to_bytes(
+        &wrapped_fungible::WrappedFungibleOperation::RegisterAuthorizedCaller {
+            app_id: bridge_app_id,
+        },
```
ma2bd, line 555: why add a comma, tho?

ma2bd (approved):

linera-bridge/tests/e2e/src/lib.rs
deuszx: it was here already, I guess it's `cargo fmt`?

PR #6475, opened 2026-06-09
Export ValueCache occupancy and capacity gauges

ndr-ds:
## Motivation

The `ValueCache` instances (certificate, raw-certificate, blob, confirmed-block, event, worker block cache) only export `value_cache_hit` / `value_cache_miss` counters. A low hit rate alone is ambiguous: it can mean the cache is **saturated and evicting** entries that would be re-requested (so a size bump would help), or that the working set is **cold / low-reuse** (so a bigger cache changes nothing). With no occupancy signal we can only infer which, which has made cache-sizing decisions guesswork.

Additionally, the existing counters are labelled only by `[key_type, value_type]`, which is not unique per instance: `DbStorage::confirmed_block` and `WorkerState::block_cache` are both `ValueCache<CryptoHash, ConfirmedBlock>` and live in the same server process, so their hit/miss counts were conflated.

## Proposal

Export two gauges per cache, alongside the existing counters:

- `value_cache_entries` — number of entries in the bounded cache, sampled at each cleanup sweep (`quick_cache::len()`).
- `value_cache_capacity` — the cache's maximum entry count, reported from `quick_cache::capacity()` (the real, post-rounding value, so occupancy = entries/capacity cannot exceed 1).

All four metrics (hit, miss, entries, capacity) now carry a `cache` label: a stable per-instance name passed to `ValueCache::new` (`storage_blob`, `storage_confirmed_block`, `storage_certificate`, `storage_certificate_raw`, `storage_event`, `worker_block`). This disambiguates instances sharing the same key/value types — without it the two `ConfirmedBlock` caches would clobber each other's gauge series (`.set()` is last-writer-wins), and it also fixes the pre-existing hit/miss conflation.

A cache that is at capacity yet still high-miss is churning (size-bound); one well below capacity and high-miss is cold / low-reuse (size bumps won't help).

The entries gauge is refreshed inside the existing 30s weak-index cleanup task (after the sweep's guard is dropped), so there is no cost on the cache hot path. The bounded cache is wrapped in `Arc` so the task can read its length. The metrics handle is `#[cfg(with_metrics)]`-gated, so there is no unused parameter when metrics are disabled. Also adds public `name()` / `len()` / `is_empty()`.

## Test Plan

`cargo clippy --all-targets -- -D warnings` passes for `linera-cache` (both with `--features metrics` and `--no-default-features`), `linera-storage`, and `linera-core`. `cargo test -p linera-cache --features metrics` passes (18 tests), including `test_len_and_is_empty` asserting occupancy tracks inserts and never exceeds capacity under overflow (verified empirically that `quick_cache::len()` is exactly the requested size at small capacities and rounds up at larger ones — hence capacity is reported from `capacity()`, not the requested size).

## Release Plan

### DIFF
--- a/linera-cache/src/value_cache.rs
+++ b/linera-cache/src/value_cache.rs
@@ -34,6 +34,9 @@ pub const DEFAULT_CLEANUP_INTERVAL_SECS: u64 = 30;
 /// A background task periodically sweeps dead `Weak` entries from the index
 /// to prevent unbounded memory growth.
 pub struct ValueCache<K, V> {
+    /// Stable instance name distinguishing this cache in metrics; several
+    /// caches may share the same key/value types within one process.
+    name: &'static str,
     cache: Arc<Cache<K, crate::Arc<V>>>,
     weak_index: Arc<papaya::HashMap<K, Weak<V>>>,
 }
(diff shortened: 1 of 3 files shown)


ma2bd (approved):

linera-cache/src/value_cache.rs
```diff
@@ -34,7 +34,10 @@ pub const DEFAULT_CLEANUP_INTERVAL_SECS: u64 = 30;
 /// A background task periodically sweeps dead `Weak` entries from the index
 /// to prevent unbounded memory growth.
 pub struct ValueCache<K, V> {
-    cache: Cache<K, crate::Arc<V>>,
+    /// Stable instance name distinguishing this cache in metrics; several
+    /// caches may share the same key/value types within one process.
+    name: &'static str,
```
ma2bd, line 46: Any constraints to be documented on the acceptable values here?

PR #6479, opened 2026-06-10
Use uncompressed encoding for secp256k1 validator public keys

### LINKED ISSUE #5702: [main] Protocol changes to simplify bridging
While working on the bridge compatible with testnet conway, some suboptimal design decisions had to be made to make it work for the protocol compatible with testnet  conway. When developing for `main` (and next testnet release) we can change the protocol to simplify:
1. Validator secp256k1 keys (verifying/protocol keys, not accounts) are in _compressed_ form. This makes the signature verification more costly on the EVM. We should post uncompressed keys and simplify verification.
2.

deuszx:
## Motivation

Flamegraphs of validator nodes showed significant time spent in `Secp256k1PublicKey::from_bytes`. Parsing a 33-byte compressed SEC1 key requires recovering the `y` coordinate via a modular square root — a full field exponentiation — and this cost is paid for every validator key in every deserialized certificate, vote, and committee blob.

The compressed encoding also complicated the Ethereum bridge: `LightClient.addCommittee` needed callers to supply uncompressed keys out-of-band (Ethereum addresses are derived from the uncompressed form), and then had to verify on-chain that each supplied key matched the compressed key in the committee blob (x-coordinate match, y-parity check, on-curve verification).

## Proposal

Store secp256k1 validator public keys in the 65-byte uncompressed SEC1 encoding (`0x04 || x || y`) instead of the 33-byte compressed one.

- **`linera-base`**: `as_bytes()`, BCS serialization, and the WIT layout of `Secp256k1PublicKey` now use the uncompressed encoding. Parsing an uncompressed key only validates the curve equation instead of computing a square root: measured **6.11 µs → 68 ns per key (~90x)** in release mode. Hex/JSON parsing remains tag-based and still accepts compressed keys, so existing human-readable configs stay readable. EVM account keys (`EvmPublicKey`) are unaffected and stay compressed.
- **`linera-bridge` (LightClient)**: `addCommittee(data, committeeBlob)` drops the caller-supplied `bytes[] validators` parameter. `_parseCommitteeBlob` derives each validator's Ethereum address directly via `keccak256` of the key's `x || y`
bytes from the hash-committed blob, so `_verifyKeyCompression` (and the field-prime constant) is removed entirely — there is no untrusted key material left to cross-check. The relay and `EvmLightClient` callers drop the now-unneeded key-extraction plumbing.

⚠️ This is a **breaking protocol change**: the key's serialized bytes feed `CryptoHash`, so derived account owners and committee blob hashes change. Existing networks cannot read old certificates; a new deployment is required.

## Test Plan

- CI
- Performance was measured with a release-mode microbenchmark calling `from_bytes` 10,000 times on the same key in each encoding (6.11 µs/iter compressed vs 68 ns/iter uncompressed).
- A multi-agent security review of the three commits (Solidity/EVM, Rust crypto, and cross-layer consistency lenses with adversarial false-positive filtering) found no newly introduced vulnerabilities.

## Release Plan

- Backporting is not possible but we may want to deploy a new `testnet` and release a new SDK.

## Links
- Closes https://github.com/linera-io/linera-protocol/issues/5702

### DIFF
--- a/linera-bridge/src/evm/client.rs
+++ b/linera-bridge/src/evm/client.rs
@@ -18,7 +18,6 @@ use alloy::{
 };
 use alloy_sol_types::SolCall;
 use linera_base::crypto::ValidatorPublicKey;
-use linera_execution::committee::Committee;
 use url::Url;
 
 use super::light_client::addCommitteeCall;
@@ -63,9 +62,6 @@ impl EvmLightClient {
 
     /// Calls `LightClient.addCommittee()` on the EVM chain.
     ///
-    /// Extracts uncompressed validator keys from the committee blob internally,
-    /// then submits the transaction to the LightClient contract.
-    ///
     /// - `certificate_bytes`: BCS-serialized `ConfirmedBlockCertificate`
     /// - `committee_blob`: raw committee blob bytes (BCS-serialized `Committee`)
     pub async fn add_committee(
@@ -73,12 +69,9 @@ impl EvmLightClient {
         certificate_bytes: &[u8],
         committee_blob: &[u8],
     ) -> anyhow::Result<TxHash> {
-        let validator_keys = extract_validator_keys(committee_blob)?;
-
         let call = addCommitteeCall {
             data: Bytes::copy_from_slice(certificate_bytes),
             committeeBlob: Bytes::copy_from_slice(committee_blob),
-            validators: validator_keys.into_iter().map(Bytes::from).collect(),
         };
 
         let tx = TransactionRequest::default()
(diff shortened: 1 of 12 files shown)


deuszx:
bridge e2e tests passed for this PR https://github.com/linera-io/linera-protocol/actions/runs/27350927830

linera-bridge/src/evm/client.rs
```diff
@@ -63,22 +62,16 @@ impl EvmLightClient {
...
-    /// then submits the transaction to the LightClient contract.
-    ///
     /// - `certificate_bytes`: BCS-serialized `ConfirmedBlockCertificate`
     /// - `committee_blob`: raw committee blob bytes (BCS-serialized `Committee`)
     pub async fn add_committee(
         &self,
         certificate_bytes: &[u8],
         committee_blob: &[u8],
     ) -> anyhow::Result<TxHash> {
-        let validator_keys = extract_validator_keys(committee_blob)?;
```
ma2bd: nice

ma2bd (approved):

PR #6481, opened 2026-06-11
Sync the admin chain and retry when receiving a certificate from an unknown epoch

ndr-ds:
## Motivation

A client whose local view of the admin chain is stale (e.g. a returning web
wallet whose committee has since rotated) can get permanently stuck: when it
downloads a blob's publishing certificate from a validator, the local
`check_certificate` returns `FutureEpoch` and the recovery fails — silently,
identically for every node — and `update_local_node_with_blobs_from` then
reports a fabricated `BlobsNotFound` for the originally requested blob.

This was observed in production on linera.market: a returning user's wallet
failed to connect with `Blobs not found: [ApplicationDescription …]` even
though every quorum validator served both the blob and its certificate
correctly (verified via direct gRPC). The wallet's local epoch never advances
on its own: nothing reacts to `FutureEpoch`, and the existing sync paths only
fan out to the stale committee, whose members may be dead or lagging. The only
user-side remedy was clearing site data.

## Proposal

- In `receive_sender_certificate`, react to `CheckCertificateResult::FutureEpoch`
  instead of failing: synchronize the admin chain — first from the node(s) that
  served the certificate, since they evidently know the newer epoch even if our
  own committee view is stale or unreachable, then falling back to the known
  committee — and check the certificate again. A per-node sync only counts if
  it actually made the epoch known.
- When blob recovery fails against every validator, surface a dedicated
  `chain_client::Error::BlobRecoveryFailed(BlobId)` rather than fabricating a
  `NodeError::BlobsNotFound` on the client. The per-validator failure reasons
  (committee-epoch errors, connection failures, genuine absence) are already
  logged at `warn!`; the surfaced error no longer fabricates a node error and
  does not depend on any single validator's response.

Known siblings deliberately left for follow-up (same pattern, off the incident
path): `download_and_process_sender_chain` silently drops `FutureEpoch` sender
certificates from its retry queue, and `download_certificates_for_events`
fabricates `EventsNotFound` over per-node committee failures. Under contention
several concurrent recovery closures may each trigger an admin sync; the syncs
are idempotent, so this is redundant work rather than a hazard.

## Test Plan

- New regression test `test_stale_client_reads_blob_published_in_future_epoch`:
  a client pinned at epoch 0 reads a data blob whose publishing certificate was
  signed by the epoch-1 committee. Before the fix it failed with
  `RemoteNodeError(BlobsNotFound(…))` while the per-validator warnings showed
  the real cause ("Cannot accept a certificate from an unknown committee in
  the future"); with the fix it passes on all storage backends.
- `test_finalize_locked_block_with_blobs` updated: reading an unpublished blob
  now surfaces `BlobRecoveryFailed` instead of the fabricated
  `RemoteNodeError(BlobsNotFound)`.
- `cargo test -p linera-core client_tests`: 67 passed, 0 failed.
- `cargo clippy -p linera-core --all-targets`: no warnings.

## Release Plan

    - be released in a new SDK,

The self-heal half landed on `testnet_conway` via #6486; the `BlobRecoveryFailed`
change is backported separately.

## Links

- Error-visibility predecessors (logging only, masking unchanged): #6391, #6398

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -1398,7 +1398,41 @@ impl<Env: Environment> Client<Env> {
     ) -> Result<(), chain_client::Error> {
         // Verify the certificate before doing any expensive networking.
         if let ReceiveCertificateMode::NeedsCheck = mode {
-            self.check_certificate(&certificate).await?.into_result()?;
+            let mut check_result = self.check_certificate(&certificate).await?;
+            if matches!(check_result, CheckCertificateResult::FutureEpoch) {
+                // The certificate is from an epoch our local view of the admin chain
+                // hasn't caught up to yet. Catch up and check again instead of failing.
+                // Prefer the nodes that gave us the certificate: they evidently know
+                // the newer epoch even if our own committee view is stale or its
+                // members are unreachable. A sync only counts if it actually made the
+                // epoch known; otherwise fall back to the known committee.
+                let admin_chain_id = self.admin_chain_id;
+                let epoch = certificate.block().header.epoch;
+                info!(
+                    %epoch,
+                    "certificate is from an unknown epoch; synchronizing the admin chain"
+                );
+                for node in nodes.iter().flatten() {
+                    match Box::pin(self.synchronize_chain_state_from(node, admin_chain_id)).await {
+                        Ok(()) => {
+                            check_result = self.check_certificate(&certificate).await?;
+                            if !matches!(check_result, CheckCertificateResult::FutureEpoch) {
+                                break;
+                            }
+                        }
+                        Err(error) => warn!(
+                            %error,
+                            validator = %node.public_key,
+                            "failed to synchronize the admin chain from validator",
+                        ),
+                    }
+                }
+                if matches!(check_result, CheckCertificateResult::FutureEpoch) {
+                    Box::pin(self.synchronize_chain_state(admin_chain_id)).await?;
+                    check_result = self.check_certificate(&certificate).await?;
+                }
+            }
+            check_result.into_result()?;
         }
         // Recover history from the network.
         let nodes = if let Some(nodes) = nodes {
@@ -2200,7 +2234,7 @@ impl<Env: Environment> Client<Env> {
                         "failed to download certificate-for-blob from validator",
                     );
                 }
-                chain_client::Error::from(NodeError::BlobsNotFound(vec![blob_id]))
+                blob_recovery_error(errors, blob_id)
             })
         }))
         .buffer_unordered(self.options.max_joined_tasks)
@@ -2406,6 +2440,26 @@ impl CheckCertificateResult {
     }
 }
 
+/// Chooses the error to surface when every validator failed to provide the certificate
+/// introducing a blob. A committee-epoch failure is local and node-independent: if any
+/// node ran into it, it is the real cause — surface it rather than fabricating a
+/// `BlobsNotFound` for a blob the network may well have.
+fn blob_recovery_error(
+    mut errors: Vec<(ValidatorPublicKey, chain_client::Error)>,
+    blob_id: BlobId,
+) -> chain_client::Error {
+    if let Some(position) = errors.iter().position(|(_, error)| {
+        matches!(
+            error,
+            chain_client::Error::CommitteeSynchronizationError
+                | chain_client::Error::CommitteeDeprecationError
+        )
+    }) {
+        return errors.swap_remove(position).1;
+    }
+    chain_client::Error::from(NodeError::BlobsNotFound(vec![blob_id]))
+}
+
 /// Creates a compressed Contract, Service and bytecode, plus an optional
 /// `ApplicationFormats` blob built from the JSON-encoded `Formats` description
 /// bytes.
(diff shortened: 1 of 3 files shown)


linera-core/src/client/mod.rs
```diff
@@ -2406,6 +2440,26 @@ impl CheckCertificateResult {
     }
 }
 
+/// Chooses the error to surface when every validator failed to provide the certificate
+/// introducing a blob. A committee-epoch failure is local and node-independent: if any
+/// node ran into it, it is the real cause — surface it rather than fabricating a
+/// `BlobsNotFound` for a blob the network may well have.
+fn blob_recovery_error(
```
ma2bd: Do we need this? This is breaking the current (defensive) design where we don't trust a single validator's error (it could be fake).

ma2bd: If we want to log an `error` or a `warn` (in a somewhat escalatory way) that's fine.

EDIT: looks like we already do

linera-core/src/client/mod.rs
```diff
@@ -2406,6 +2440,26 @@ impl CheckCertificateResult {
...
+    if let Some(position) = errors.iter().position(|(_, error)| {
+        matches!(
+            error,
+            chain_client::Error::CommitteeSynchronizationError
+                | chain_client::Error::CommitteeDeprecationError
+        )
+    }) {
+        return errors.swap_remove(position).1;
+    }
+    chain_client::Error::from(NodeError::BlobsNotFound(vec![blob_id]))
```
ma2bd: The previous code was doing it but IMHO it is really confusing to construct a fake `NodeError` on the client.

linera-core/src/client/mod.rs
```diff
@@ -2200,7 +2234,7 @@ impl<Env: Environment> Client<Env> {
                         "failed to download certificate-for-blob from validator",
                     );
                 }
-                chain_client::Error::from(NodeError::BlobsNotFound(vec![blob_id]))
```
ma2bd, line 2203: why don't we return a special `chain_client::Error` ?

linera-core/src/client/mod.rs
```diff
@@ -1398,7 +1398,41 @@ impl<Env: Environment> Client<Env> {
...
+                // the newer epoch even if our own committee view is stale or its
+                // members are unreachable. A sync only counts if it actually made the
+                // epoch known; otherwise fall back to the known committee.
+                let admin_chain_id = self.admin_chain_id;
+                let epoch = certificate.block().header.epoch;
+                info!(
+                    %epoch,
+                    "certificate is from an unknown epoch; synchronizing the admin chain"
+                );
+                for node in nodes.iter().flatten() {
```
afck, line 1415: ~Below we use `validator_nodes` if this is `None`. I think `None` is supposed to mean "as _all_ validators".~

(Edit: I see, you try again below anyway.)

afck, line 1415: Elsewhere we try to avoid iterating over all validators and asking them one-by-one, because if the first one is slow, that blocks the client for a while. We use `communicate_concurrently` instead; maybe that would be a better choice here, too?

linera-core/src/client/mod.rs
```diff
@@ -1398,7 +1398,41 @@ impl<Env: Environment> Client<Env> {
...
+                    match Box::pin(self.synchronize_chain_state_from(node, admin_chain_id)).await {
+                        Ok(()) => {
+                            check_result = self.check_certificate(&certificate).await?;
+                            if !matches!(check_result, CheckCertificateResult::FutureEpoch) {
+                                break;
+                            }
+                        }
+                        Err(error) => warn!(
```
afck: We try to avoid WARN-level logs that can be caused by a single faulty validator.

linera-core/src/client/mod.rs
```diff
@@ -2451,3 +2505,64 @@ pub async fn create_bytecode_blobs(
...
+        )];
+        assert_matches::assert_matches!(
+            blob_recovery_error(errors, blob_id),
+            chain_client::Error::RemoteNodeError(NodeError::BlobsNotFound(ids)) if ids == vec![blob_id]
+        );
+        assert_matches::assert_matches!(
+            blob_recovery_error(Vec::new(), blob_id),
+            chain_client::Error::RemoteNodeError(NodeError::BlobsNotFound(ids)) if ids == vec![blob_id]
+        );
+    }
```
afck: (Not sure if these tests carry their weight; `blob_recovery_error` is pretty simple, and ideally if it didn't work that should surface in other, higher-level tests?)

linera-core/src/unit_tests/client_tests.rs
```diff
@@ -1399,6 +1399,57 @@ where
...
 }
 
+/// Tests that a client whose local view of the admin chain is stale can still use a blob
+/// whose publishing certificate was signed by a committee from an epoch the client has
+/// not heard of yet.
+///
+/// The blob-recovery path downloads the publishing certificate from a validator, but
+/// validating it locally yields `CheckCertificateResult::FutureEpoch`. The client must
+/// react by catching up on the admin chain and retrying — instead of failing, let alone
+/// misreporting the failure as `BlobsNotFound` for the originally requested blob.
```
afck: That seems to document the behavior before the fix.

linera-core/src/client/mod.rs
```diff
@@ -1398,7 +1398,46 @@ impl<Env: Environment> Client<Env> {
...
+                    let certificate = &certificate;
+                    let _ = communicate_concurrently(
+                        nodes,
+                        async move |node| {
+                            Box::pin(self.synchronize_chain_state_from(&node, admin_chain_id))
+                                .await?;
+                            match self.check_certificate(certificate).await? {
+                                CheckCertificateResult::FutureEpoch => {
+                                    Err(chain_client::Error::CommitteeSynchronizationError)
+                                }
```
afck: What if it's `OldEpoch`?

ma2bd (approved):

ndr-ds:
To use `communicate_concurrently` here we need to box the recursion, and it's a bigger refactor. I'm gonna just do it as a follow up

PR #6490, opened 2026-06-15
Anchor multi-leader jitter target to the current round's start.

afck:
## Motivation

The jitter delay used `now` as the round-start anchor whenever the proposed round differed from `manager.current_round` (the conflict-bump case) or `manager.round_timeout` was missing. On a busy chain where `process_inbox` interrupts repeatedly, `propose_at = now + delay` was recomputed each retry and the target receded by `delay` every time, so a non-preferred client never reached it and never proposed.

## Proposal

Derive the anchor instead from data the chain manager already tracks: `manager.round_timeout - ownership.round_timeout(current_round)`. If either is `None`, propose immediately rather than wait against a stale anchor.

## Test Plan

A regression test was added.

## Release Plan

- Release SDK.

## Links

- A different approach, storing the round start in the client instead of chain manager: https://github.com/linera-io/linera-protocol/pull/6484
- Alternatively revert multi-leader jitter entirely: https://github.com/linera-io/linera-protocol/pull/6455

### DIFF
--- a/linera-core/src/client/chain_client/mod.rs
+++ b/linera-core/src/client/chain_client/mod.rs
@@ -2415,9 +2415,9 @@ impl<Env: Environment> ChainClient<Env> {
     /// the owner should propose immediately (either because the round is not a multi-leader
     /// round, the owner is the preferred proposer, or the jitter target is already in the past).
     ///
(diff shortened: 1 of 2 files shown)


afck:
Closing in favor of #6455.

PR #6491, opened 2026-06-15
[testnet] Return a dedicated CannotDownloadBlob error instead of fabricating BlobsNotFound

ndr-ds:
## Motivation

Follow-up to #6486 (the self-heal backport) addressing the review feedback on
#6481. `testnet_conway` currently still fabricates a `NodeError::BlobsNotFound`
on the client when blob recovery fails against every validator, which is
misleading: the blob may well exist on the network, and the real per-validator
causes (committee-epoch mismatch, connection failures, genuine absence) are
hidden behind a fake node error.

## Proposal

When blob recovery fails against every validator, surface a dedicated
`chain_client::Error::BlobRecoveryFailed(BlobId)` rather than fabricating a
`NodeError::BlobsNotFound` on the client. The per-validator failure reasons are
already logged at `warn!`; the surfaced error no longer fabricates a node error
and does not depend on any single validator's response.

This is the `BlobRecoveryFailed` change that #6486 deliberately excluded, now
applied on top of it. Equivalent to the corresponding change in #6481.

## Test Plan

- `test_finalize_locked_block_with_blobs` updated: reading an unpublished blob
  now surfaces `BlobRecoveryFailed` instead of the fabricated
  `RemoteNodeError(BlobsNotFound)`.
- `cargo test -p linera-core client_tests`: 62 passed, 0 failed.
- `cargo clippy -p linera-core --all-targets`: no warnings.

## Release Plan

    - be released in a new SDK,

## Links

- Backport of the error-surfacing nit from #6481
- Follows #6486 (self-heal)

### DIFF
--- a/linera-core/src/client/chain_client/mod.rs
+++ b/linera-core/src/client/chain_client/mod.rs
@@ -317,7 +317,7 @@ pub enum Error {
     },
 
     #[error("No validator provided a usable certificate registering blob {0}")]
-    BlobRecoveryFailed(BlobId),
+    CannotDownloadBlob(BlobId),
 
     #[error(transparent)]
     BcsError(#[from] bcs::Error),
--- a/linera-core/src/client/chain_client/mod.rs
+++ b/linera-core/src/client/chain_client/mod.rs
@@ -316,6 +316,9 @@ pub enum Error {
         target_next_block_height: BlockHeight,
     },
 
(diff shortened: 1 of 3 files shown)


ma2bd (approved):

ma2bd:
Not sure "recovery" is the right word. Maybe "fetching".

ma2bd:
E.g. we use the word "recovery" in the code for chain repairs.

Issue #6492, opened 2026-06-16
Guard against apps failing in `summarize_events`.


afck:
If an app's `summarize_events` code fails, that should not prevent the checkpoint from happening.

PR #6493, opened 2026-06-16
Checkpoints with events

afck:
## Motivation

Events currently disable checkpoints.

## Proposal

Allow checkpoints on chains with events, as long as they are not _system_ events. (I.e. the admin chain still cannot be checkpointed for now.)

To keep checkpoints from growing indefinitely, events from before the latest checkpoints are not guaranteed to be available anymore. Applications are given the opportunity to publish summary events at each checkpoint.

If an application's summary code fails, the checkpoint block fails. This will be addressed later: https://github.com/linera-io/linera-protocol/issues/6492

## Test Plan

A test was added.

## Release Plan

## Links

- Part of #460.

### DIFF
--- a/linera-execution/src/execution.rs
+++ b/linera-execution/src/execution.rs
@@ -304,6 +304,7 @@ pub enum UserAction {
     Operation(OperationContext, Vec<u8>),
     Message(MessageContext, Vec<u8>),
     ProcessStreams(ProcessStreamsContext, Vec<StreamUpdate>),
+    SummarizeEvents(ProcessStreamsContext, Vec<StreamUpdate>),
 }
 
 impl UserAction {
@@ -312,6 +313,7 @@ impl UserAction {
             UserAction::Instantiate(context, _) => context.authenticated_owner,
             UserAction::Operation(context, _) => context.authenticated_owner,
             UserAction::ProcessStreams(_, _) => None,
+            UserAction::SummarizeEvents(_, _) => None,
             UserAction::Message(context, _) => context.authenticated_owner,
         }
     }
@@ -321,6 +323,7 @@ impl UserAction {
             UserAction::Instantiate(context, _) => context.height,
             UserAction::Operation(context, _) => context.height,
             UserAction::ProcessStreams(context, _) => context.height,
+            UserAction::SummarizeEvents(context, _) => context.height,
             UserAction::Message(context, _) => context.height,
         }
     }
@@ -330,6 +333,7 @@ impl UserAction {
             UserAction::Instantiate(context, _) => context.round,
             UserAction::Operation(context, _) => context.round,
             UserAction::ProcessStreams(context, _) => context.round,
+            UserAction::SummarizeEvents(context, _) => context.round,
             UserAction::Message(context, _) => context.round,
         }
     }
@@ -339,6 +343,7 @@ impl UserAction {
             UserAction::Instantiate(context, _) => context.timestamp,
             UserAction::Operation(context, _) => context.timestamp,
             UserAction::ProcessStreams(context, _) => context.timestamp,
+            UserAction::SummarizeEvents(context, _) => context.timestamp,
             UserAction::Message(context, _) => context.timestamp,
         }
     }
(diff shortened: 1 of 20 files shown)


linera-execution/src/execution.rs
```diff
@@ -127,16 +127,24 @@ where
         &mut self,
         maximum_blob_size: u64,
     ) -> Result<Vec<Blob>, ExecutionError> {
-        let mut had_event_block = false;
+        // User event streams are summarized and pruned by the checkpoint itself (see
+        // `ExecutionStateActor`'s checkpoint handler), so they do not block checkpointing.
+        // System event streams (e.g. the epoch streams on the admin chain) are not
+        // summarized, so a chain that published any is still not allowed to checkpoint.
+        let mut had_system_event_block = false;
         self.previous_event_blocks
```
ma2bd, line 135: Is this including event blocks since the beginning of time or since the latest checkpoint?

linera-execution/src/execution_state_actor.rs
```diff
@@ -985,7 +1043,9 @@ where
         );
         let is_free = matches!(
             &action,
-            UserAction::Message(..) | UserAction::ProcessStreams(..)
+            UserAction::Message(..)
+                | UserAction::ProcessStreams(..)
+                | UserAction::SummarizeEvents(..)
```
ma2bd, line 1048: IMHO both `SummarizeEvents` and `ProcessStreams` (which is now skippable) should be metered.

ma2bd: Why is `UserAction::Message(..)` free btw?

linera-execution/src/execution_state_actor.rs
```diff
@@ -903,6 +904,63 @@ where
...
+            };
+            let next_index = self
+                .state
+                .system
+                .stream_event_counts
+                .get(&stream_id)
+                .await?
+                .unwrap_or(0);
+            // A summary is an absolute-state snapshot, so the application is not handed an
+            // incremental range to fold in; `previous_index` is left at 0.
```
ma2bd, line 935: Actually, it would be nice to receive the index of the latest summary (if any) instead of 0.

linera-execution/src/execution.rs
afck: With this change, it will only be since the latest checkpoint, to keep it from growing indefinitely.

linera-execution/src/execution_state_actor.rs
afck: Actually, all three of those are only free if the app is marked as free. See the `&&` below.

linera-execution/src/execution_state_actor.rs
afck: I don't think it's worth adding another map ("first event of this epoch") just for that.

So the alternative would be to look at the previous checkpoint and see which events were emitted there, and use 0 if none. However:
* That would set it to `0` even if there were already events in an earlier epoch, just not the previous one.
* It's another input that would need to become its own oracle value, because someone restoring from _this_ checkpoint probably doesn't _have_ the previous one.

Also, applications could easily keep track of this data themselves, if they need to.

ma2bd: * for nontrivial summaries, isn't it always going to be old-summary + updates -> new-summary ?
* (unrelated) we're not synchronizing epochs and checkpoints so far, are we?
* (provided that summaries only happen during checkpoints) would the block-hash (or height) of the last checkpoint (if any) be sufficient? This height is probably already visible in (or required by) the truncation of the confirmed log.

afck: * Probably, but that's for each app to decide?
* No; I'd do that in a separate PR, because it also opens some more questions to discuss, I think.

ma2bd: Indeed, we may not want to synchronize erochs and checkpoints. (was just reacting to your response)

ma2bd: We also have `self.previous_event_blocks` at our disposal by the way

ma2bd (approved):

linera-execution/src/execution_state_actor.rs
afck: Yes, but `previous_event_blocks` points to the most recent event, not the first event ~in the last epoch~ since the previous checkpoint.

(Sorry, yes, I really meant "interval between checkpoints" here.)

PR #6496, opened 2026-06-17
Add historical execution-state hashing behind FLAG_HISTORICAL_HASH(_SHADOW)

ma2bd:
## Motivation

`testnet_conway` currently runs with `FLAG_ZERO_HASH`, which skips execution-state hashing
entirely (the reported `state_hash` is all-zeros). We want to switch the network to *historical
hashing* — a rolling hash that is seeded once from a full content hash and then extended cheaply
from each block's write batch — without breaking the existing storage format and without a
network-wide migration or reset.

Wrapping `ExecutionStateView` directly in `HistoricallyHashableView` is not viable here: that
wrapper re-tags the inner view under `KeyTag::Inner`, shifting every existing execution-state key
by one prefix byte, which would require migrating every chain's storage.

## Proposal

Add historical hashing to `ExecutionStateView` *in place*, gated by new committee-policy flags, so
the storage layout of the existing fields is unchanged:

- New flags (parallel to `FLAG_ZERO_HASH`):
  - `FLAG_HISTORICAL_HASH` — enforced historical hashing.
  - `FLAG_HISTORICAL_HASH_SHADOW` — compute, persist and log the rolling hash, but keep reporting
    an all-zeros hash so it is **not** enforced in consensus. This lets validators cross-check
    their historical hashes (by comparing the logged values) before enforcement.
- `ExecutionStateView` now derives `View` (instead of `HashableView`) and gains an appended
  `historical_hash: RegisterView<C, Option<HasherOutput>>`. Appending the field gives it a fresh
  key tag, so existing chains load it as `None` with no migration and no movement of the other
  fields' keys. `HashableView` is hand-written over only the content fields so the rolling hash
  never feeds back into the content hash that seeds it.
- `crypto_hash_mut` selects a mode from the current committee's policy:
  - **Zero** (`FLAG_ZERO_HASH`) — unchanged.
  - **Historical** — on the first block after activation (`historical_hash == None`) the hash is
    seeded from the full content hash via `HashableView::hash_mut` (which already reflects this
    block's changes); every later block extends the stored hash with `SHA3-256(stored || bcs(batch))`
    over only the content fields' pending writes. In shadow mode the value is computed, stored and
    logged but a zero hash is reported.
  - **Legacy** — the previous full-content hash (value unchanged from before).
  - In both **Zero** and **Legacy** modes the stored rolling hash is dropped if present, so a chain
    that is deactivated and later re-activated re-seeds from a fresh full-content hash rather than
    extending a stale anchor that predates the blocks executed while it was inactive. Chains that
    were never activated store nothing (a `None` register writes no key).

The hashing convention mirrors `main` (same `ExecutionStateViewHash` wrapping and `make_hash`
chaining), with the one deliberate difference that the seed reuses the existing `HashableView`
content hash rather than `main`'s `dump_content` byte-scan.

## Test Plan

New unit tests in `linera-execution` (all green; `cargo clippy` and nightly `cargo fmt` clean):

- `historical_hashing_enforce_seeds_then_chains` — first block seeds (non-zero, populates the
  register); a later content change extends the chain to a different non-zero hash.
- `historical_hashing_shadow_reports_zero_but_advances` — shadow mode reports zero but still
  advances the stored rolling hash.
- `historical_hashing_is_deterministic` — identical state + operations agree; a differing
  operation diverges.
- `historical_hashing_resets_when_deactivated` — activate→seed, deactivate (rolling hash dropped),
  re-activate (re-seeded), confirming reactivation does not extend a stale anchor.
- `historical_hashing_persists_and_chains_across_reload` — e2e through the save/reload lifecycle:
  an existing chain loads the appended field as `None` (format compatibility), then the rolling
  hash is seeded, persisted, reloaded, chained across the reload boundary, and persisted again.
- `make_historical_hash_matches_recorded_value` — golden test pinning the chaining primitive to a
  recorded value, so any change to the algorithm or `Batch` serialization (i.e. any
  consensus-visible hash change) fails deliberately.

Rollout sequence on the live network: deploy → reconfigure to `FLAG_HISTORICAL_HASH_SHADOW` and
compare the logged `historical_state_hash` across validators for an epoch (this also surfaces any
latent execution-state divergence as a log mismatch rather than a chain halt) → reconfigure to
`FLAG_HISTORICAL_HASH` to enforce.

## Release Plan

- This PR targets `testnet_conway` directly, so there is nothing to backport. The new flags are
  testnet-only (like `FLAG_ZERO_HASH`, they should not survive to mainnet), so they are **not**
  intended to be forward-ported to `main`.
- No new SDK or storage migration is required: the appended `historical_hash` field loads as
  `None` on existing chains and stores nothing until activated.
- Activation is gated behind a committee-policy flag and changes the consensus-visible `state_hash`
  once enabled, so it is rolled out network-wide via a validator reconfiguration (shadow first,
  then enforce).

## Links

### DIFF
--- a/linera-execution/src/execution.rs
+++ b/linera-execution/src/execution.rs
@@ -175,24 +175,11 @@ where
                 self.system.pre_save(&mut batch)?;
                 self.users.pre_save(&mut batch)?;
                 self.stream_event_counts.pre_save(&mut batch)?;
-                Self::make_hash(Some(stored), &batch)?
+                make_historical_hash(Some(stored), &batch)?
             }
         };
         self.historical_hash.set(Some(hash));
         Ok(hash)
-    }
-
-    /// Extends a stored hash with a batch: `SHA3-256(stored || bcs(batch))`. An empty batch
-    /// leaves the hash unchanged. Mirrors `HistoricallyHashableView::make_hash`.
-    fn make_hash(stored: Option<HasherOutput>, batch: &Batch) -> Result<HasherOutput, ViewError> {
-        let stored = stored.unwrap_or_default();
-        if batch.is_empty() {
-            return Ok(stored);
-        }
-        let mut hasher = linera_views::sha3::Sha3_256::default();
-        hasher.update_with_bytes(stored.as_ref())?;
-        hasher.update_with_bcs_bytes(&batch)?;
-        Ok(hasher.finalize())
     }
 
     /// Wraps a raw 32-byte hash into a domain-separated [`CryptoHash`], matching the convention
@@ -203,6 +190,23 @@ where
         impl BcsHashable<'_> for ExecutionStateViewHash {}
         CryptoHash::new(&ExecutionStateViewHash(hash.into()))
     }
+}
+
+/// Extends a stored hash with a batch: `SHA3-256(stored || bcs(batch))`. An empty batch leaves the
+/// hash unchanged. Mirrors `HistoricallyHashableView::make_hash`. Changing this function changes
+/// consensus-visible execution-state hashes — see `make_historical_hash_matches_recorded_value`.
+fn make_historical_hash(
+    stored: Option<HasherOutput>,
+    batch: &Batch,
+) -> Result<HasherOutput, ViewError> {
+    let stored = stored.unwrap_or_default();
+    if batch.is_empty() {
+        return Ok(stored);
+    }
+    let mut hasher = linera_views::sha3::Sha3_256::default();
+    hasher.update_with_bytes(stored.as_ref())?;
+    hasher.update_with_bcs_bytes(&batch)?;
+    Ok(hasher.finalize())
 }
 
 impl<C: Context, C2: Context> ReplaceContext<C2> for ExecutionStateView<C> {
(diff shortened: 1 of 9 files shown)


afck:
> These changes should be backported

This is already targeting `testnet_conway`.
(And I don't think we should forward-port them?)

linera-execution/src/execution.rs
```diff
@@ -49,6 +53,56 @@ pub struct ExecutionStateView<C> {
...
+        Ok(hasher.finalize())
+    }
+
+    async fn hash(&self) -> Result<HasherOutput, ViewError> {
+        use std::io::Write as _;
+        let mut hasher = Self::Hasher::default();
+        hasher.write_all(self.system.hash().await?.as_ref())?;
+        hasher.write_all(self.users.hash().await?.as_ref())?;
+        hasher.write_all(self.stream_event_counts.hash().await?.as_ref())?;
+        Ok(hasher.finalize())
```
afck, line 116: It would be safer to destructure `self` in these methods, so the compiler would notice if we added a field and forgot to hash it here.

linera-execution/src/execution.rs
```diff
@@ -57,27 +111,117 @@ where
...
+            None => self.hash_mut().await?,
+            Some(stored) => {
+                // Hash only the content fields' pending writes. The `historical_hash` register is
+                // intentionally excluded so the rolling hash never depends on its own value (it
+                // may still be a pending, unsaved write at this point). This mirrors `main`, where
+                // the wrapper's hash key sits outside the hashed inner view.
+                let mut batch = Batch::new();
+                self.system.pre_save(&mut batch)?;
+                self.users.pre_save(&mut batch)?;
+                self.stream_event_counts.pre_save(&mut batch)?;
```
afck: Here destructuring (`let Self { … } = self`) would also be safer.

afck (approved):

linera-execution/src/execution.rs
ma2bd: Done in 42d050b8b2 — destructured `self` (with `historical_hash: _`) so a new field forces a compile error here.

linera-execution/src/execution.rs
ma2bd: Done in 42d050b8b2 — destructured `self` (with `historical_hash: _`) so a new field forces a compile error here.

PR #6501, opened 2026-06-17
Make `process_batches` chandle corruption/rollback errors correctly

afck:
This is #6498 plus one commit.

## Motivation

`process_batch` is not using `chain_write`, so it doesn't handle poison or inconsistent chain state errors.

## Proposal

Route the cross-chain batch driver's write step through `chain_write` so it inherits the same guarantees as every other write path, instead of acquiring its own lock and saving directly.

- `process_batch` now returns `Result<(), WorkerError>`. On a save failure it still rolls back and replies `BatchRolledBack` to all callers, but now also returns the save error so the caller can act on it.
- The batch driver (`ChainBatchRequestProcessor::create`) no longer calls `handle::write_lock` + `process_batch` itself. It drains the channel to coalesce requests, then calls `worker.chain_write(chain_id, |guard| guard.process_batch(requests))`. This gives the batch save cancellation safety and recovery.
- The driver captures the `WorkerState` and chain ID rather than a pinned `ChainWorkerArc`, so it no longer blocks TTL reclaim — `chain_write` re-resolves the worker per batch, which is also why a retry after eviction picks up a fresh worker.
- `enqueue_and_drive` maps a dropped result sender (meaning `chain_write` failed before `process_batch` ran, e.g. an already-poisoned-and-now-evicted worker) to `PoisonedWorker` instead of panicking.

The second commit closes one more gap: `process_batch` only surfaced save failures to `chain_write` and swallowed errors raised during processing.
- A _read_ can return a `must_reload_view` store error or a `CorruptedChainState` error before `save` is ever reached. Previously that error went only to the originating request's sender and `process_batch` returned `Ok`, so `chain_write` never evicted/reset the worker — the chain stays wedged until some unrelated write path tripped the same error.
- New `classify_processing_error` helper: if a processing error reports `must_reload_view()` or `indicates_corrupted_chain_state()`, it's captured in `recovery_error` (handed to `chain_write` for eviction/reset) and the caller gets `BatchRolledBack` to retry against a fresh worker. Any other processing error is unchanged — caller still gets the exact error and the worker stays healthy.
- The rollback return became `save_error.or(recovery_error)`, so both failure sources flow to recovery.

A caller whose request triggered a recoverable error now sees `BatchRolledBack` (retryable) rather than the raw `ViewError`, matching what the save-failure path already does to all callers.

## Test Plan

CI
It's hard to fully test issues like cancellation safety and DB corruptions in general.

## Release Plan

- Port to `main`.
- Release SDK.
- Hotfix validators.

## Links

- Based on #6498.

### DIFF
--- a/linera-core/src/chain_worker/state.rs
+++ b/linera-core/src/chain_worker/state.rs
@@ -1280,7 +1280,10 @@ where
     ///
     /// Both update and confirmation requests are handled together so that a
     /// single write-lock acquisition covers all pending work for the chain.
-    pub(crate) async fn process_batch(&mut self, requests: Vec<BatchRequest>) {
+    pub(crate) async fn process_batch(
+        &mut self,
+        requests: Vec<BatchRequest>,
+    ) -> Result<(), WorkerError> {
         let mut update_results = Vec::new();
         let mut confirm_results = Vec::new();
         let mut need_save = false;
@@ -1348,10 +1351,12 @@ where
                 }
             }
         }
+        let mut save_error = None;
         if !need_rollback && need_save {
             if let Err(error) = self.save().await {
                 tracing::error!(%error, "failed to save batch; rolling back");
                 need_rollback = true;
+                save_error = Some(error);
             }
         }
         if need_rollback {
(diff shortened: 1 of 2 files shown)


linera-core/src/chain_worker/state.rs
```diff
@@ -1366,11 +1371,14 @@ where
...
-            // Surface a *save* failure to the caller so `chain_write` can evict or
-            // reset the worker if it was left poisoned or corrupted. Processing
-            // errors that also trigger a rollback were already reported to their
-            // individual senders and leave the worker healthy, so they return `Ok`.
-            return match save_error {
+            // Surface an error that left the worker poisoned or the chain state
+            // corrupted so `chain_write` can evict or reset it. This covers both a
+            // failed `save` and a processing step that hit inconsistent state in
+            // storage (e.g. a journal-resolution failure while lazily loading a
+            // view). Ordinary processing errors (bad height, validation failures)
```
ma2bd: Note that during loading a journal resolution error will fail the `load`, so it doesn't need extra poisoning.

ma2bd (approved):

PR #6503, opened 2026-06-18
Inform subscriber chain: which events still readable?

afck:
## Motivation

Events before the publisher chain's previous checkpoints are not guaranteed to be readable anymore, but the subscriber chain doesn't know which events _are_ the ones after the checkpoint.

## Proposal

Add a `first_index` (readable floor) to event streams so a subscriber's `process_streams` knows where pre-checkpoint history ends.

Also, track the floor at the worker level, not in the execution state, to make this work with sparse blocks.

## Test Plan

CI; extended unit tests.

## Release Plan

## Links


linera-chain/src/chain.rs
```diff
@@ -1759,45 +1775,56 @@ where
...
             .next_expected_events
             .multi_get_pairs(stream_ids)
             .await?
             .into_iter()
-            .zip(list_indices)
+            .zip(ranges)
         {
-            let initial_index = if stream_id == StreamId::system(EPOCH_STREAM_NAME) {
-                // we don't expect the epoch stream to contain event 0
-                1
```
ma2bd, line 1792: what happened with this special case?

linera-chain/src/chain.rs
```diff
@@ -213,6 +213,20 @@ impl std::ops::Deref for ChainIdSet {
     }
 }
 
+/// The event indices we track for a stream, maintained whenever a block is processed
+/// (executed or merely preprocessed).
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq, Serialize, Deserialize, Allocative)]
+#[cfg_attr(with_graphql, derive(async_graphql::SimpleObject))]
+pub struct StreamCounts {
+    /// The lowest event index still guaranteed to be readable: the index of the first event
```
ma2bd: I suppose `first_index == next_index` is possible, meaning there is no event yet, after the last checkpoint?

ma2bd (approved):

ma2bd (approved):

linera-chain/src/chain.rs
afck: That's now handled by the logic below: The special case of the system stream starting at index 1 now behaves just as if the first index after a checkpoint happened to be 1.

linera-chain/src/chain.rs
afck: Clarified in https://github.com/linera-io/linera-protocol/pull/6503/changes/deb6aee4876b7bffe9403b3ea379c7f2fb34fdde.

PR #6515, opened 2026-06-19
Classify storage-service write-batch failures as must_reload_view

ndr-ds:
## Motivation

`must_reload_view()` marks errors after which the in-memory view can no longer be trusted —
notably an **ambiguous write**: the batch may or may not have been executed, and the client
cannot tell which.

A storage-service `write_batch` is sent over gRPC. If that call fails with a timeout, an
`UNAVAILABLE`, or a lost response, the server may already have applied the batch. The client
has no way to distinguish that from "never applied", so the view has to be reloaded. Today the
backend does not override the trait default (`false`) and the error is treated as benign.

ScyllaDB already classifies the same condition this way
(`ScyllaDbStoreInternalError::WriteBatchExecutionError`).

## Proposal

Add `StorageServiceStoreError::WriteBatchError`, map the `process_write_batch_extended`
failure to it in `submit_statements`, and return `must_reload_view() == true` for that variant
only.

Read RPCs keep producing `GrpcError` and are unaffected — a reload on a failed read would be
costly, since `classify_processing_error` turns these errors into `BatchRolledBack` and
discards a whole cross-chain batch.

## Test Plan

`cargo clippy -p linera-storage-service --all-targets` is clean. The variant is only produced
by a live gRPC failure against a running storage-server, so there is no unit test; the
classification is a one-line `matches!` mirroring the ScyllaDB one.

## Release Plan

## Links

- Same area as #6507 / #6508 (`JournalingError::must_reload_view`).


linera-views/src/backends/rocks_db.rs
```diff
@@ -711,6 +713,12 @@ pub enum RocksDbStoreInternalError {
     #[error("RocksDB error: {0}")]
     RocksDb(#[from] rocksdb::Error),
 
+    /// RocksDB error while writing a batch. Unlike [`RocksDbStoreInternalError::RocksDb`]
+    /// (which also covers read failures), this error means a batch write may or may not
+    /// have been applied, so the in-memory view must be reloaded from storage.
```
ma2bd: Because it runs in the same process, I don't think rocksdb has ambiguous writes. Please link the doc if you found something.

linera-storage-service/src/common.rs
```diff
@@ -74,6 +74,15 @@ impl From<Status> for StorageServiceStoreError {
...
 impl KeyValueStoreError for StorageServiceStoreError {
     const BACKEND: &'static str = "service";
+
+    fn must_reload_view(&self) -> bool {
+        // A `write_batch` is sent as one or more gRPC calls. If the round-trip fails
+        // (a gRPC status such as `DEADLINE_EXCEEDED`/`UNAVAILABLE`, or a transport-level
+        // error) the server may or may not have applied the batch, so the in-memory view
+        // must be reloaded from storage. These variants can also surface on read RPCs,
+        // where a reload is unnecessary but harmless; we err on the side of reloading.
+        matches!(self, Self::GrpcError(_) | Self::TransportError(_))
```
ma2bd: this makes sense!

linera-views/src/backends/rocks_db.rs
ndr-ds: Sorry, this was supposed to be a draft. A lot of this still needs to be verified

ma2bd: Np. (Probably) just revert this part

ma2bd:
> This mirrors ScyllaDB's existing WriteBatchExecutionError
(linera-views/src/backends/scylla_db.rs), whose batches are likewise not atomic.

WriteBatch operations should always be atomic. The question is whether some reported errors create ambiguous writes -- meaning that we don't know if the batch was (atomically) executed or not --, or unresolved journals -- meaning the state is not safe to read.

PR #6542, opened 2026-06-20
Snapshot example services' GraphQL schemas as insta tests

ma2bd:
## Motivation

We already record each application's serde/BCS formats as an `insta` snapshot (via the `BcsApplication` trait) so that unintended changes to the wire format of operations, messages and events show up as a reviewable diff. The GraphQL schema exposed by an application's *service* has no equivalent guard: today nothing flags an accidental change to the query/mutation surface that clients depend on.

## Proposal

Establish a lightweight, **purely conventional** pattern (no new trait, no macro) for snapshotting a service's GraphQL schema, and apply it to `hex-game` as the reference implementation.

Unlike `BcsApplication::formats()`, which is type-only, building an `async-graphql` `Schema` needs owned instances of the roots and a `ServiceRuntime` — so the natural home is a `#[cfg(test)]` unit test in `service.rs`, where `ServiceRuntime` resolves to `MockServiceRuntime`. Concretely, for `hex-game`:

- Extract `HexService::schema()` and have `handle_query` delegate to it, so the snapshotted SDL cannot drift from what the service actually serves.
- Add a `schema_sdl` test: `insta::assert_snapshot!(service.schema().sdl())`, reusing the existing mock-runtime setup.
- Commit the resulting SDL snapshot.
- Apply the pattern the other GraphQL example services.

## Test Plan

- `cargo test -p hex-game --bin hex_game_service` — both `query` and the new `schema_sdl` pass.
- Re-ran `schema_sdl` against the committed snapshot without `INSTA_UPDATE` to confirm the SDL is deterministic (the `MapView` hash-suffixed type names are stable across runs).
- `cargo fmt -- --check` (pinned nightly) and `cargo clippy -p hex-game --tests` both clean.

## Release Plan

## Links

### DIFF
--- a/examples/amm/src/service.rs
+++ b/examples/amm/src/service.rs
@@ -63,7 +63,7 @@ impl AmmService {
     }
 }
 
-#[cfg(test)]
+#[cfg(all(test, not(target_arch = "wasm32")))]
 mod tests {
     use linera_sdk::{util::BlockingWait, views::View, ServiceRuntime};
 
--- a/examples/amm/src/service.rs
+++ b/examples/amm/src/service.rs
@@ -10,7 +10,7 @@ use std::sync::Arc;
 use amm::{Operation, Parameters};
 use async_graphql::{EmptySubscription, Request, Response, Schema};
 use linera_sdk::{
-    graphql::GraphQLMutationRoot as _, linera_base_types::WithServiceAbi, views::View, Service,
+    graphql::GraphQLMutationRoot, linera_base_types::WithServiceAbi, views::View, Service,
     ServiceRuntime,
 };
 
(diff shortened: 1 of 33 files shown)


deuszx (commented):
I think the description is outdated(?) since you added this pattern for all examples?

Also, why not add the `fn schema()` to `trait ServiceAbi`?

ma2bd:
> I think the description is outdated(?) since you added this pattern for all examples?

Yes thx, I keep reminding Claude to do that

> Also, why not add the `fn schema()` to `trait ServiceAbi`?

Because in theory, an app doesn't have to use GraphQL. It's just a convention that most (Wasm) apps use (like BCS) but it's not part of the core protocol.

afck:
(I guess we could have a `GraphQlServiceAbi: ServiceAbi` trait, so some apps could use that one.)

deuszx:
> Because in theory, an app doesn't have to use GraphQL. It's just a convention that most (Wasm) apps use (like BCS) but it's not part of the core protocol.

But we have (for example) `impl Service for HexService` that returns `GraphQL` schema. If we already have all(?) apps use it then we could make it "official" and easier to discover

ma2bd:
> (I guess we could have a `GraphQlServiceAbi: ServiceAbi` trait, so some apps could use that one.)

Yes this could work. And also `BcsContractAbi` I suppose.

PR #6547, opened 2026-06-21
Decode well-known linera-base primitives as human-readable JSON in the BCS-to-JSON converter

ma2bd:
## Motivation

When the formats-registry / explorer decodes BCS payloads to JSON via `serde_reflection`'s `json_converter`, well-known linera-base primitives come out in their *binary* shape — a `CryptoHash` as a 32-element byte array, an `Amount` as a number, an `AccountOwner` as `{Address32: [...]}` — because the traced registry describes the BCS form, not the human-readable one each type defines via `is_human_readable()`. We also lacked the inverse direction (JSON → BCS), which is useful for tooling and tests.

## Proposal

Provide a "better" `json_converter` environment so those primitives round-trip through their human-readable representation, using **route (a)**: the well-known names are removed from the stored registry so (de)serialization falls back to the environment.

- **`LineraEnvironment`** (`linera-sdk/src/formats.rs`): implements both `DeserializationEnvironment` and `SerializationEnvironment`. On decode it turns a known primitive's BCS bytes into the concrete type and emits `serde_json::to_value` of it (human-readable); on encode it reads the human-readable JSON back into the concrete type and serializes its BCS form. It only fires for names *absent* from the registry, so it is a no-op on un-pruned registries.
- **`Formats::prune_known_primitives`** (non-wasm): before removing anything, it traces each primitive's canonical BCS `ContainerFormat` and compares it to the registry entry, returning `PruneError::Mismatch { name }` (and pruning nothing) on a collision/drift; only then does it remove the verified entries.
- **`BcsApplication::pruned_formats`**: a default trait method returning `formats()` with the primitives pruned — the single, drift-free entry point for the publishable/human-readable form. `formats()` stays the faithful, self-contained reflection.
- **Encoding API**: `Formats::{encode_operation, encode_response, encode_message, encode_event_value}` as inverses of the existing `decode_*` helpers. The internal `bcs_to_json` / `json_to_bcs` are now private.
- A single macro is the source of truth for the `name => Type` list, generating the decoder, the encoder, and the canonical-format tracer so they can't drift.
- The 19 example `tests/format.rs` snapshot tests now use `App::pruned_formats()`, so the published formats render primitives in human-readable form (12 snapshots updated; the rest reference no such primitive).

### Scope / limitations (documented in-code)

Supported, verifiable primitives: `CryptoHash, AccountOwner, Amount, Epoch, BlobId, StreamId, ModuleId, ApplicationId`. Excluded for now:
- **Crypto keys/signatures** (`Secp256k1PublicKey`, `EvmPublicKey`, …): their `Deserialize` validates curve-point validity, so `serde_reflection` cannot trace them from dummy bytes and we cannot verify their format before pruning. They also don't appear in any application ABI. Supporting them needs sample-based tracing (future work).
- **`U128`** (anonymous `u128` in BCS) and **`Blob`** (serializes as `BlobContent`) — not addressable by registry name.

## Test Plan

- `cargo test -p linera-sdk --lib formats` — 9/9 pass, including an end-to-end oracle (`decode(bcs) == serde_json::to_value(original)` after pruning), the inverse check (`encode(decode(bcs)) == bcs`), encode/decode helper round-trips, a mismatch-rejection test, and a canonical-registry build check.
- `cargo +nightly fmt --check` and `cargo clippy -p linera-sdk` clean.
- Regenerated all example `format` snapshots (`INSTA_UPDATE=always cargo test -p <example> --test format`), and re-verified them in compare mode after switching to `pruned_formats()`. `gen-nft` and `llm` can't be built in the dev sandbox (their ML deps need a CPU `fullfp16` feature); `llm`'s prune is a verified no-op (its ABI references no such primitive), and `gen-nft`'s snapshot was updated by hand to match the deterministic prune output (CI builds both and will verify).

## Release Plan

## Links

### DIFF
--- a/linera-sdk/src/formats.rs
+++ b/linera-sdk/src/formats.rs
@@ -53,7 +53,7 @@ pub trait BcsApplication {
     }
 }
 
-/// Decode BCS-serialized `bytes` into a [`serde_json::Value`], guided by `format`
+/// Decodes BCS-serialized `bytes` into a [`serde_json::Value`], guided by `format`
 /// and the container `registry`.
 ///
 /// Well-known linera-base primitives that are absent from `registry` (because they
@@ -72,7 +72,7 @@ fn bcs_to_json(
     bcs::from_bytes_seed(context, bytes)
 }
 
-/// Encode a [`serde_json::Value`] into BCS `bytes`, guided by `format` and the
+/// Encodes a [`serde_json::Value`] into BCS `bytes`, guided by `format` and the
 /// container `registry`. This is the inverse of [`bcs_to_json`].
 ///
 /// Well-known linera-base primitives that are absent from `registry` are read from
@@ -252,7 +252,7 @@ impl Formats {
         bcs_to_json(bytes, &self.event_value, &self.registry)
     }
 
-    /// Encode a JSON operation value into its BCS bytes. Inverse of
+    /// Encodes a JSON operation value into its BCS bytes. Inverse of
     /// [`decode_operation`](Self::decode_operation).
     pub fn encode_operation(&self, value: &serde_json::Value) -> bcs::Result<Vec<u8>> {
         json_to_bcs(value, &self.operation, &self.registry)
--- a/linera-sdk/src/formats.rs
+++ b/linera-sdk/src/formats.rs
@@ -109,7 +109,7 @@ where
     T: serde::Serialize + serde::de::DeserializeOwned,
     S: serde::Serializer,
 {
-    let value: T = serde_json::from_value(value.clone()).map_err(serde::ser::Error::custom)?;
+    let value: T = T::deserialize(value).map_err(serde::ser::Error::custom)?;
     value.serialize(serializer)
 }
 
--- a/linera-sdk/src/formats.rs
+++ b/linera-sdk/src/formats.rs
@@ -5,9 +5,11 @@
 
 use serde::{Deserialize, Serialize};
 use serde_reflection::{
-    json_converter::{DeserializationContext, EmptyEnvironment},
+    json_converter::{DeserializationContext, DeserializationEnvironment, SymbolTableEnvironment},
     Format, Registry,
 };
+#[cfg(not(target_arch = "wasm32"))]
+use serde_reflection::{Samples, Tracer, TracerConfig};
 
 /// The serde formats used by an application. The exact serde encoding in use must be
 /// known separately.
@@ -37,6 +39,10 @@ pub trait BcsApplication {
 
 /// Decode BCS-serialized `bytes` into a [`serde_json::Value`], guided by `format`
 /// and the container `registry`.
+///
+/// Well-known linera-base primitives that are absent from `registry` (because they
+/// were removed by [`Formats::prune_known_primitives`]) are decoded into their
+/// human-readable representation by [`LineraEnvironment`].
 pub fn bcs_to_json(
     bytes: &[u8],
     format: &Format,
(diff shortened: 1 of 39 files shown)


linera-sdk/src/formats.rs
```diff
@@ -31,25 +36,201 @@ pub trait BcsApplication {
...
+    /// meant to be published to a formats registry.
+    #[cfg(not(target_arch = "wasm32"))]
+    fn pruned_formats() -> Result<Formats, PruneError> {
+        let mut formats = Self::formats()?;
+        formats.prune_known_primitives()?;
+        Ok(formats)
+    }
 }
 
 /// Decode BCS-serialized `bytes` into a [`serde_json::Value`], guided by `format`
```
afck:
```suggestion
/// Decodes BCS-serialized `bytes` into a [`serde_json::Value`], guided by `format`
```

linera-sdk/src/formats.rs
```diff
@@ -31,25 +36,201 @@ pub trait BcsApplication {
...
     let context = DeserializationContext {
         format: format.clone(),
         registry,
-        environment: &EmptyEnvironment,
+        environment: &LineraEnvironment,
     };
     bcs::from_bytes_seed(context, bytes)
 }
 
+/// Encode a [`serde_json::Value`] into BCS `bytes`, guided by `format` and the
```
afck:
```suggestion
/// Encodes a [`serde_json::Value`] into BCS `bytes`, guided by `format` and the
```

linera-sdk/src/formats.rs
```diff
@@ -70,6 +251,65 @@ impl Formats {
     pub fn decode_event_value(&self, bytes: &[u8]) -> bcs::Result<serde_json::Value> {
         bcs_to_json(bytes, &self.event_value, &self.registry)
     }
+
+    /// Encode a JSON operation value into its BCS bytes. Inverse of
```
afck:
```suggestion
    /// Encodes a JSON operation value into its BCS bytes. Inverse of
```

(Also below.)

linera-sdk/src/formats.rs
```diff
@@ -70,6 +251,65 @@ impl Formats {
...
+        let expected = expected_primitive_registry()?;
+        // Verify everything first so a mismatch never leaves the registry half-pruned.
+        for name in KNOWN_PRIMITIVE_NAMES {
+            let (Some(actual), Some(expected_format)) =
+                (self.registry.get(*name), expected.get(*name))
+            else {
+                continue;
+            };
+            if actual != expected_format {
+                return Err(PruneError::Mismatch {
```
afck, line 296: So if my app has a custom `Amount` type but uses linera-base's `AccountOwner`, the latter will also not be shown in a human-readable way?

Why not only exclude `Amount` itself in that case?

afck (approved):

deuszx (commented):
I don't see anti-regression tests 👀 i.e. take JSON encoding of those primitive types and hardcode expected deserializations into Linera types (and BCS). And vice versa.

linera-sdk/src/formats.rs
```diff
@@ -221,4 +476,105 @@ mod tests {
...
+        let expected = serde_json::to_value(&value).unwrap();
+        assert_eq!(decoded, expected);
+        // Sanity-check the human-readable shape: a hex hash and a decimal amount string.
+        assert_eq!(decoded["hash"], json!("ab".repeat(32)));
+        assert_eq!(decoded["amount"], json!(value.amount.to_string()));
+
+        // Encoding is the inverse of decoding: the human-readable JSON re-encodes to
+        // exactly the original BCS bytes.
+        let reencoded = formats.encode_operation(&decoded).unwrap();
+        assert_eq!(reencoded, bytes);
```
ma2bd, line 546: @deuszx Were you looking for this?

deuszx: AFAIU this is a roundtrip test, it doesn't protect from breaking changes. It tests that we can encode what we decoded.

linera-sdk/src/formats.rs
ma2bd: You can't have a custom `Amount` type. Names will collide

deuszx: I can have a type that serializes to the same `JSON` as built-in `Amount` though?

ma2bd: to the same binary(BCS) schema, yes

ma2bd: I believe the human-readable(e.g. JSON) serialization of `Amount` is `String`

linera-sdk/src/formats.rs
ma2bd: I see what you mean

ma2bd: Claude claims that the 8 primitive types are already pinned in the block format here:
https://github.com/linera-io/linera-protocol/blob/testnet_conway/linera-rpc/tests/snapshots/format__format.yaml.snap

deuszx: I guess it's equivalent.

PR #6552, opened 2026-06-22
Race serving nodes concurrently when syncing the admin chain for an unknown epoch

ndr-ds:
## Motivation

Follow-up to #6481. When a client with a stale view of the admin chain hits a
`FutureEpoch` certificate, `receive_sender_certificate` catches up by syncing
the admin chain from the node(s) that served the certificate. #6481 did this
with a sequential loop over those nodes; every other place that takes a list of
nodes (`download_blobs`, `download_pre_checkpoint_blocks`) races them with
`communicate_concurrently` instead, so a slow or unresponsive node doesn't
block the others. This makes the admin-chain self-heal consistent with that
pattern.

## Proposal

- Replace the sequential per-node loop in `receive_sender_certificate` with
  `communicate_concurrently` over the serving nodes (first node that actually
  advances the epoch wins; the rest are aborted), keeping the known-committee
  `synchronize_chain_state` fallback unchanged.
- Convert `receive_sender_certificate` from an `async fn` to a function
  returning `Pin<Box<dyn Future + Send>>`. This is required, not cosmetic: the
  admin-chain sync recurses back into `receive_sender_certificate`, and without
  boxing the future at this boundary the `Send` bound needed by
  `communicate_concurrently` either overflows (`E0275`) or fails at the
  recursion site. Boxing the recursive future once at the function boundary is
  the minimal fix (no `async_recursion` dependency, no call-site casts).

Note on scope: today the only `NeedsCheck` caller passes a single serving node,
so the concurrency is latent — behavior is unchanged for the current callers.
The change makes the function honor a multi-node list correctly (and matches the
rest of the file) rather than relying on callers never passing more than one.

## Test Plan

- Existing `test_stale_client_reads_blob_published_in_future_epoch` still passes
  (behavior preserved on the single-node path).
- New `test_stale_client_reads_future_epoch_blob_with_offline_validator`: a stale
  client catches up and reads a future-epoch blob while one validator is offline,
  exercising recovery against a partially-unreachable committee.
- `cargo test -p linera-core client_tests`: all pass.
- `cargo clippy -p linera-core --all-features --all-targets`: no warnings.

## Release Plan

    - be released in a new SDK,

## Links

- Follow-up to #6481 (addresses @afck's review suggestion to use `communicate_concurrently`)


linera-core/src/client/mod.rs
```diff
@@ -1401,68 +1400,86 @@ impl<Env: Environment> Client<Env> {
...
-                            }
-                        }
-                        Err(error) => debug!(
-                            %error,
-                            validator = %node.public_key,
-                            "failed to synchronize the admin chain from validator",
-                        ),
-                    }
-                }
+    ) -> ReceiveSenderCertificateFuture<'a> {
```
afck: I think you could remove the `'a`s above, and use the implicit `self` lifetime?

```suggestion
    ) -> ReceiveSenderCertificateFuture<'_> {
```

afck: Alternatively, you could keep it an `async fn` (i.e. don't change the signature at all) and just add an `.await` after `Box::pin(…)`? Then you don't have to do the `Send` dance above.

afck (approved):

linera-core/src/client/mod.rs
ndr-ds: I still got an overflow when trying that :/
I got the lifetime elided though

PR #6556, opened 2026-06-23
Batch validator catch-up via an aggregated MissingCrossChainUpdates error

ma2bd:
## Motivation

A client that owns a busy hub chain (one receiving cross-chain messages from many sender chains) periodically froze for minutes and tripped the chain-idle liveness probe, restarting the process.

(Likely) Root cause: when a validator is behind on the sender chains a block consumes, it rejects the proposal **one missing sender at a time** (`MissingCrossChainUpdate`), and the client brings it up to date with **one blocking round-trip per sender**. Accepting a single high-fan-out block therefore takes ~N sequential `propose → sync-one-sender → retry` round-trips per validator, and because per-chain block production is a single task gated on a quorum, this serializes into multi-minute stalls (observed `process_inbox` p99 ≈ 8.7 min during a freeze). An operational mitigation (`maxPendingMessageBundles=1`) is already deployed; this is the protocol-level fix.

## Proposal

Add an aggregated error, `MissingCrossChainUpdates { chain_id, bundles }` — the batched form of `MissingCrossChainUpdate` — that lets a validator declare **every** missing incoming bundle it needs to validate a block proposal in a single response, and have the client fetch them all in one batch before retrying.

Concretely:
- When validating a proposal, the validator now **collects all** missing cross-chain bundles instead of bailing on the first, and returns them together as `(origin, height)` pairs.
- The client handles `MissingCrossChainUpdates` on the block-proposal path, syncing every reported origin chain in a single batch (via `send_chain_info_up_to_heights`) and retrying once. The local node's "download missing sender blocks from a validator" path is batched the same way: all reported senders are fetched, then the proposal is retried once.

This deliberately targets **proposals only**. Confirmed certificates don't have the same pathology: missing incoming bundles are tolerated (`must_be_present=false`), missing blobs are already discovered up front from `required_blob_ids` and returned as a single `BlobsNotFound`, and the only relevant missing events (the committee/epoch stream) are already batched into one `EventsNotFound`. So there is no O(N)-round-trip problem to fix there, and the confirmed-certificate path is left on its existing batched errors.

Because clients and validators are upgraded independently, the change is **backward compatible**. The wire error is appended without disturbing existing ones, and a new `supports_aggregated_missing` field on the proposal / confirmed-certificate requests lets a client advertise that it understands the aggregated error. The validator adapts its reply accordingly at the gRPC boundary: capable clients receive every missing sender as `MissingCrossChainUpdates`; older clients are downgraded to the legacy per-sender `MissingCrossChainUpdate` (the first missing sender) and recover one round-trip at a time, exactly as before. All four old/new client/validator combinations interoperate.

A separate follow-up will retire the legacy one-at-a-time path once every deployed node understands the aggregated error.

## Test Plan

- A worker-level test asserts that a proposal consuming several not-yet-received bundles is rejected with a single `MissingCrossChainUpdates` listing **all** of them (rather than bailing on the first).
- A client-level integration test (`test_proposal_batches_missing_dependency_catch_up`) drives the full path end to end: a block consuming bundles from three sender chains is proposed to a validator that is behind on **all** of them; the validator reports one aggregated error, the client catches every sender up in a single batch, and the block is confirmed. The test also confirms the loop terminates.
- Unit tests cover the gRPC error adaptation in both directions: capable clients get `MissingCrossChainUpdates` unchanged; older clients get it downgraded to the legacy per-sender `MissingCrossChainUpdate`.
- Unit tests cover the wire round-trip of the new error (bincode-in-protobuf) and the chain-to-node error mapping.
- `cargo clippy --all-targets` (incl. `--features server`), `cargo +nightly fmt`, and the affected crates' test suites pass.

## Release Plan

Targets `testnet_conway` for now. The new error variant is additive and inert until validators are upgraded; the capability handshake keeps the rollout safe for any mix of upgraded and non-upgraded clients and validators. Changing the wire error set requires a coordinated validator/client deployment.

### DIFF
--- a/linera-chain/src/chain.rs
+++ b/linera-chain/src/chain.rs
@@ -699,6 +699,10 @@ where
         }
         let origins = bundles_by_origin.keys().copied().collect::<Vec<_>>();
         let inboxes = self.inboxes.try_load_entries_mut(&origins).await?;
+        // When the bundles must already be present (block proposals), collect *every* missing
+        // `(origin, height)` rather than bailing on the first, so the caller can be told the
+        // full set of cross-chain updates to fetch in a single round-trip.
+        let mut missing_bundles = Vec::new();
         for ((origin, bundles), mut inbox) in bundles_by_origin.into_iter().zip(inboxes) {
             tracing::trace!(
                 "Removing [{}] from inbox for {origin}",
@@ -714,15 +718,8 @@ where
                     .remove_bundle(bundle)
                     .await
                     .map_err(|error| (chain_id, origin, error))?;
-                if must_be_present {
-                    ensure!(
-                        was_present,
-                        ChainError::MissingCrossChainUpdate {
-                            chain_id,
-                            origin,
-                            height: bundle.height,
-                        }
-                    );
+                if must_be_present && !was_present {
+                    missing_bundles.push((origin, bundle.height));
                 }
             }
             inbox.observe_size_metric();
@@ -732,6 +729,15 @@ where
                 }
             }
         }
+        ensure!(
+            missing_bundles.is_empty(),
+            ChainError::MissingDependencies {
+                chain_id,
+                bundles: missing_bundles,
+                events: Vec::new(),
+                blobs: Vec::new(),
+            }
+        );
         Ok(())
     }
 
--- a/linera-chain/src/chain.rs
+++ b/linera-chain/src/chain.rs
@@ -731,11 +731,9 @@ where
         }
         ensure!(
             missing_bundles.is_empty(),
-            ChainError::MissingDependencies {
+            ChainError::MissingCrossChainUpdates {
                 chain_id,
                 bundles: missing_bundles,
-                events: Vec::new(),
-                blobs: Vec::new(),
             }
         );
         Ok(())
--- a/linera-core/src/unit_tests/client_tests.rs
+++ b/linera-core/src/unit_tests/client_tests.rs
@@ -1082,6 +1082,91 @@ where
 #[cfg_attr(feature = "rocksdb", test_case(RocksDbStorageBuilder::new().await; "rocks_db"))]
 #[cfg_attr(feature = "scylladb", test_case(ScyllaDbStorageBuilder::default(); "scylla_db"))]
 #[test_log::test(tokio::test)]
+async fn test_proposal_batches_missing_dependency_catch_up<B>(
+    storage_builder: B,
+) -> anyhow::Result<()>
+where
+    B: StorageBuilder,
+{
+    // Regression test for the hub-chain freeze fixed by the aggregated `MissingDependencies`
+    // error. A block that consumes incoming bundles from several sender chains is proposed to a
+    // validator that is behind on *all* of those senders. The validator must report every
+    // missing sender at once and the client must catch them up in a single batch, instead of
+    // the legacy one-rejection-and-round-trip-per-sender loop that serialized into multi-minute
+    // stalls. The test exercises the client-side batching/termination logic in `updater.rs`,
+    // which the worker- and gRPC-level unit tests in this PR do not cover.
+    let signer = InMemorySigner::new(None);
+    // Zero default-faulty validators so the only unavailable validators are the ones this test
+    // deliberately takes offline below; the quorum is still three out of four.
+    let mut builder = TestBuilder::new(storage_builder, 4, 0, signer).await?;
+
+    // Three independent sender chains and one recipient ("hub") chain. The committee has four
+    // validators with a quorum of three.
+    let sender1 = builder.add_root_chain(1, Amount::from_tokens(2)).await?;
+    let sender2 = builder.add_root_chain(2, Amount::from_tokens(2)).await?;
+    let sender3 = builder.add_root_chain(3, Amount::from_tokens(2)).await?;
+    let recipient = builder.add_root_chain(4, Amount::ZERO).await?;
+    let recipient_id = recipient.chain_id();
+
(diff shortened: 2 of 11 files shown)


linera-chain/src/chain.rs
```diff
@@ -732,6 +729,13 @@ where
                 }
             }
         }
+        ensure!(
```
deuszx, line 732: IMO `ensure!` does more harm than good. It wraps a VERY simply construct that is now less readable (`return *` is assumed and condition is reverted to a form that is less obvious):
```rust
if !missing_bundles.is_empty() {
  return ...
}
```
is more readable

linera-core/src/updater.rs
```diff
@@ -477,7 +477,11 @@ where
         clock_skew_sender: mpsc::UnboundedSender<ClockSkewReport>,
     ) -> Result<Box<ChainInfo>, chain_client::Error> {
         let chain_id = proposal.content.block.chain_id;
+        // `sent_cross_chain_updates` tracks per-origin progress for the legacy per-sender
+        // `MissingCrossChainUpdate` path (a non-upgraded validator); `synced_cross_chain_updates`
```
deuszx, line 481: so is it a client or validator? Here you write _validator_ but below in `adapt_dependency_error` (and other places) you wrote _client_.

linera-rpc/proto/rpc.proto
```diff
@@ -278,6 +278,12 @@ message BlockProposal {
   // A lite certificate for a validated block, or a fast block proposal, that
   // justifies the proposal in this round.
   optional bytes original_proposal = 6;
+
+  // Whether the sender understands the aggregated `MissingCrossChainUpdates` error. When
+  // false (e.g. an older client that predates it), the validator downgrades that error
+  // to the legacy per-sender `MissingCrossChainUpdate` before replying, for backward
+  // compatibility.
+  bool supports_aggregated_missing = 7;
```
deuszx, line 286: I don't understand why we need this. A simpler approach is to update client code to understand both error variants, update them first, then update validators to return the new error variant, then (after a while) remove the old code. We control the apps and their release cycles.

linera-core/src/unit_tests/client_tests.rs
```diff
@@ -1077,6 +1077,92 @@ where
...
+
+    // Phase B: validator 3 comes back, but validator 2 — one of the validators that *does* have
+    // the sender blocks — goes offline. The recipient's block now needs validator 3's vote to
+    // reach a quorum {0, 1, 3}, so the updater cannot route around validator 3: it must catch it
+    // up on every missing sender before validator 3 can accept the proposal.
+    builder.set_fault_type([3], FaultType::Honest);
+    builder.set_fault_type([2], FaultType::Offline);
+
+    // Consume all three bundles in a single block. Validator 3 rejects the proposal with an
+    // aggregated `MissingCrossChainUpdates` listing all three senders; the client syncs them in
```
deuszx, line 1141: I mean, we don't really know which error variant was returned. This test probably duplicates other tests that already exercise the catch-up.

linera-core/src/unit_tests/worker_tests.rs
```diff
@@ -1546,6 +1546,61 @@ where
...
+            .unwrap_err();
+        let WorkerError::ChainError(chain_error) = error else {
+            panic!("unexpected error: {error}");
+        };
+        match *chain_error {
+            ChainError::MissingCrossChainUpdates { bundles, .. } => assert_eq!(
+                bundles,
+                vec![
+                    (chain_1, BlockHeight::ZERO),
+                    (chain_1, BlockHeight::from(1))
```
deuszx, line 1598: we should test with different chains as well.

linera-core/src/updater.rs
afck: That makes sense, doesn't it?
* This code here is in the client. When the client receives a `MissingCrossChainUpdate` (singular!) from a validator, that means that _validator_ is not updated yet.
* `adapt_dependency_error` below runs on the server, i.e. _validator_ side, and converts `MissingCrossChainupdates` to `MissingCrossChainUpdate` to send to a client that is not updated yet.

afck: ~(But I agree it might be nicer to keep the code simple instead, and just first update the clients, then the validators.)~

(On the other hand, I guess this approach simply allows for a faster rollout.)

afck:
I really think this should ultimately be merged with `BlobsNotFound` and `EventsNotFound`, so only one round-trip is needed even if blobs, events _and_ incoming messages are missing?

linera-rpc/proto/rpc.proto
ma2bd: that's 3 PRs instead of 2

linera-chain/src/chain.rs
ma2bd: This code has been using `ensure!` to state runtime invariants since day 1.

linera-rpc/proto/rpc.proto
deuszx: As discussed on Slack - two PRs, three deployments. Just like this one.

linera-core/src/unit_tests/worker_tests.rs
ma2bd: I'd say we're covering all the changes in `remove_bundles_from_inboxes` already.

linera-rpc/proto/rpc.proto
ma2bd: This is a more flexible deployment strategy than what we usually do.

afck (approved):

PR #6560, opened 2026-06-24
Fix faucet_balance_amount gauge overflow

ndr-ds:
## Motivation

The `faucet_balance_amount` Prometheus gauge is broken: it exports garbage/negative values (e.g. `7999411257572950000`, `-3232187779200037000` on testnet-conway). The faucet balance is an `Amount` (a `u128` scaled by `10^18`), cast straight to the gauge i64:

```rust
.set(u128::from(balance) as i64);   // 200M tokens = 2e26, overflows i64 (~9.2e18)
```

A `#[expect(clippy::cast_possible_truncation)]` was silencing the lint that flagged exactly this. The metric is unusable (e.g. you cannot build a low-balance alert on it).

## Proposal

Export the balance in whole tokens (`balance / 10^18`), which fits i64 for any realistic faucet balance, and drop the silencer:

```rust
.set(i64::try_from(u128::from(balance) / u128::from(Amount::ONE)).unwrap_or(i64::MAX));
```

Whole-token granularity is plenty for a faucet holding millions of tokens (and for a low-balance alert). Help text updated to say "in whole tokens".

## Test Plan

`cargo clippy -p linera-faucet-server --features metrics --all-targets -- -D warnings` passes (silencer no longer needed). Formatting verified with rustfmt. Gauge now reports a sane positive token count instead of an overflowed i64.

## Release Plan
- These changes should be backported to the latest `testnet` branch (so the testnet-conway faucet metric is usable / alertable).

## Links
- Follow-up: faucet low-balance alert once this metric is fixed and deployed.

### DIFF
--- a/linera-faucet/server/src/lib.rs
+++ b/linera-faucet/server/src/lib.rs
@@ -127,7 +127,7 @@ mod metrics {
     pub static FAUCET_BALANCE: LazyLock<IntGaugeVec> = LazyLock::new(|| {
         register_int_gauge_vec(
             "faucet_balance_amount",
-            "Current balance of the faucet chain",
+            "Current balance of the faucet chain, in whole tokens",
             &[],
         )
     });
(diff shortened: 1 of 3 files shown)


linera-faucet/server/src/lib.rs
```diff
@@ -812,14 +812,15 @@ where
...
+        // overflowing.
         #[cfg(with_metrics)]
-        #[expect(
-            clippy::cast_possible_truncation,
-            reason = "faucet balance metric — i64 is the prometheus gauge type"
-        )]
         metrics::FAUCET_BALANCE
             .with_label_values(&[])
-            .set(u128::from(balance) as i64);
+            .set(u128::from(balance) as f64 / u128::from(Amount::ONE) as f64);
```
ma2bd: Should we implement `From<Amount>` for `f64` ?

afck (approved):

PR #6561, opened 2026-06-24
Add a `describe-application` CLI command

ma2bd:
## Motivation

Tooling around the `formats-registry` (and the explorer) needs to map an
application ID to its `ModuleId` — e.g. to register an existing application's
formats after the fact. Today there is no way to fetch an application's
description from the CLI; `applications(chainId:)` in the node service only lists
applications already registered on a *specific* chain, so an arbitrary
application can't be looked up from another chain.

## What this does

Adds `linera describe-application <application-id> [reader]`, which prints the
application's `ApplicationDescription` (module ID, creator chain, parameters and
required dependencies) as JSON:

```console
$ linera describe-application <APP_ID> | jq -r .module_id
<MODULE_ID>
```

The description blob is content-addressed and fetched from the validators by
blob ID, so the **reader chain need not host the application** — the wallet's
default chain works.

## Implementation

- `Client::get_application_description_blob` / `get_application_description`
  (linera-core) — read the description blob, downloading it from the validators
  if it is not available locally, mirroring the existing
  `get_chain_description{,_blob}` helpers.
- `ChainClient::get_application_description` — thin delegating wrapper.
- `DescribeApplication` CLI command + dispatch.

### DIFF
--- a/linera-service/src/cli/command.rs
+++ b/linera-service/src/cli/command.rs
@@ -963,6 +963,15 @@ pub enum ClientCommand {
         reader: Option<ChainId>,
     },
 
+    /// Describe an existing application: print its `ApplicationDescription` (module
+    /// ID, creator chain, parameters and required dependencies) as JSON. The
+    /// description is content-addressed and fetched from the validators, so the
+    /// application need not be registered on the wallet's default chain.
+    DescribeApplication {
+        /// The ID of the application to describe.
+        application_id: ApplicationId,
+    },
+
     /// Create an application.
     CreateApplication {
         /// The module ID of the application to create.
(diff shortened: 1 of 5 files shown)


linera-service/src/cli/command.rs
```diff
@@ -963,6 +963,19 @@ pub enum ClientCommand {
...
+    /// ID, creator chain, parameters and required dependencies) as JSON. The
+    /// description is content-addressed and fetched from the validators, so the
+    /// application need not be registered on the reader chain.
+    DescribeApplication {
+        /// The ID of the application to describe.
+        application_id: ApplicationId,
+
+        /// An optional chain ID to use as the reader. The default chain of the wallet
+        /// is used otherwise.
+        reader: Option<ChainId>,
```
afck: That seems to be completely irrelevant, since the request is just passed through to the underlying `Client`. I'd remove this argument.

afck (approved):

linera-service/src/cli/command.rs
ma2bd: ok

PR #6562, opened 2026-06-24
Encode the application formats blob as BCS instead of JSON

ma2bd:
## Motivation

On `testnet_conway`, #6555 migrated the formats-registry blob from JSON to BCS so the
encoding matches the documented intent ("the BCS serialization of an application's
`Formats`"). Most of #6555 targets the formats-registry *example application* and its
BCS-publication flow, which do not exist on `main` — `main` instead publishes the
application `Formats` as a plain data blob bundled with the module bytecode, and the
explorer fetches it by hash via the node's `applicationFormats` GraphQL query.

This PR brings #6555's *intent* to `main`'s own design: the formats blob was still being
JSON-encoded on the publish side and JSON-decoded by the explorer, even though it is
described everywhere as a BCS-encoded `Formats`. This makes the encoding match the docs
and the registry's own BCS conventions.

## Proposal

- Encode the formats blob as BCS at the publish site
  (`client_context::load_formats_from_snap`: `serde_json::to_vec` -> `bcs::to_bytes`).
- Decode it as BCS in the explorer
  (`linera-explorer/src/formats.rs`: `serde_json::from_slice` -> `linera_sdk::bcs::from_bytes`).
- Update the now-inaccurate doc comments / CLI help that described the blob as
  "JSON-encoded `Formats`" to "BCS-encoded `Formats`" across `linera-base`,
  `linera-core`, `linera-execution`, `linera-sdk`, `linera-service`, and `CLI.md`.

The GraphQL/JS-interop boundaries that are genuinely JSON (the node query envelope, the
explorer response parsing) are intentionally left as JSON. Only the blob payload changes.

This does **not** port #6555's formats-registry example app, registry ABI, or
`prepare_bcs_publication` flow, none of which exist on `main` (that registry-application
architecture is testnet-only; see the #6547 -> #6549 split for the same dividing line).

## Test Plan

- `cargo check -p linera-client -p linera-explorer` — clean.
- `cargo clippy -p linera-client -p linera-explorer` — clean.
- `cargo +nightly fmt --check` (with the CI's unstable opts) — clean.
- The blob is opaque bytes end-to-end except at the two endpoints changed here
  (encode in `linera-client`, decode in `linera-explorer`); BCS round-trips through the
  same `linera_sdk::formats::Formats` type.

## Release Plan

## Links

- Ports the blob-encoding migration from #6555 (testnet_conway) to `main`.
- Sibling SDK-layer port: #6549 (port of #6547).

### DIFF
--- a/linera-client/src/client_context.rs
+++ b/linera-client/src/client_context.rs
@@ -961,8 +961,10 @@ impl<Env: Environment> ClientContext<Env> {
 }
 
 /// Reads an insta SNAP file containing a YAML-encoded `Formats` value, parses
-/// it, and returns the canonical JSON encoding to publish as the application
-/// formats blob.
+/// it, and returns the canonical BCS encoding to publish as the application
+/// formats blob. BCS matches the documented intent (the blob is "the BCS
+/// serialization of an application's `Formats`") and the encoding the explorer
+/// decodes with.
 #[cfg(feature = "fs")]
 fn load_formats_from_snap(path: &std::path::Path) -> Result<Vec<u8>, Error> {
     let content = fs::read_to_string(path).map_err(|e| {
@@ -980,13 +982,7 @@ fn load_formats_from_snap(path: &std::path::Path) -> Result<Vec<u8>, Error> {
             format!("failed to parse SNAP body in {path:?} as Formats: {e}"),
         )
     })?;
-    serde_json::to_vec(&formats).map_err(|e| {
-        std::io::Error::new(
-            std::io::ErrorKind::InvalidData,
-            format!("failed to serialize Formats as JSON: {e}"),
-        )
-        .into()
-    })
+    Ok(bcs::to_bytes(&formats)?)
 }
 
 #[cfg(feature = "fs")]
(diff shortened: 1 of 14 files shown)


linera-client/src/client_context.rs
```diff
@@ -980,13 +982,7 @@ fn load_formats_from_snap(path: &std::path::Path) -> Result<Vec<u8>, Error> {
...
         )
     })?;
-    serde_json::to_vec(&formats).map_err(|e| {
-        std::io::Error::new(
-            std::io::ErrorKind::InvalidData,
-            format!("failed to serialize Formats as JSON: {e}"),
-        )
-        .into()
-    })
+    Ok(bcs::to_bytes(&formats)?)
```
deuszx: it'd be clearer (and safer) if this method simply returned `Formats` type.

deuszx (approved):

linera-client/src/client_context.rs
ma2bd: yes indeed

deuszx (approved):

PR #6473, opened 2026-06-09
Add observability for message and block rejections

ndr-ds:
## Motivation

We have almost no observability when messages or blocks are rejected. Most rejection
paths are silent (no log, no metric), so rejection rates can't be measured reliably and
silent rejections are invisible in production. Closes #6459.

## Proposal

Add logging at every currently-silent rejection decision site, plus rate-correct
Prometheus counters, reusing the `error_type()` labelling convention established in #5951.

- **Message rejections:** logs at the policy decision (`IncomingBundle::apply_policy`,
  INFO), the closed-chain auto-reject (`prepare_chain_info_response`, INFO), and the
  validator-side execution of a rejection (`BlockExecutionTracker`, DEBUG — see Log levels).
  The execution-failure path already logged. No new message counter — the rejection rate
  stays on the existing commit-time pair `rejected_bundle_count` / `incoming_bundle_count`;
  the reason (which can't be recovered from a committed block) lives in the logs.
- **Validator proposal rejections:** `block_proposals_received_total` (unconditional) +
  `block_proposals_rejected_total{error_type}` at the worker's `handle_block_proposal`,
  plus a DEBUG log (see Log levels). `error_type` reuses `WorkerError::error_type()`,
  surfacing `WorkerError::InvalidOwner`, `ChainError::WrongRound`, etc.
- **Client staging failures:** `block_staging_failures_total{error_type}` + INFO log in
  `ChainClient::execute_block`, reusing `execute_block_latency`'s count as the denominator.
  Adds `error_type()` to the client `Error` and `LocalNodeError`, delegating into
  `WorkerError` / `ChainError` exactly like `WorkerError::error_type` from #5951.

Each rate lives within one observation point + process role, so numerator and denominator
emit from the same site and metric names can't be summed across points.

### Log levels (why the two hottest sites are DEBUG, not INFO)

The three low-volume decision sites are INFO; the two high-volume sites (proposal rejected,
message rejected) are DEBUG. Rationale from prod (`testnet-conway`,
`rate(linera_server_request_error{method_name="handle_block_proposal"}[1h])`):

- Proposal rejections run at **~10/sec cluster-wide** (~1.7/sec per shard), but **~92% is
  routine consensus churn** — `UnexpectedBlockHeight` (4.0/s), `InvalidTimestamp` (3.7/s),
  `MissingCrossChainUpdate` (1.6/s). The genuinely adversarial rejections (`InvalidOwner`,
[...]

PR #6565, opened 2026-06-26
Add observability for message and block rejections (backport of #6473)

ndr-ds:
## Motivation

Backport of #6473 to `testnet_conway`. We have almost no observability when messages or
blocks are rejected. This adds logging and rate-correct Prometheus counters at the
previously-silent rejection sites. The metrics are most useful on the deployed network
(`testnet_conway`), where all the rejection traffic that justified the change lives.

## Proposal

Cherry-pick of #6473 (its squash commit), reusing the `error_type()` labelling convention
from #5966 (the `testnet_conway` backport of #5951).

- **Message rejections:** logs at the policy decision (`apply_policy`, INFO), the
  closed-chain auto-reject (`prepare_chain_info_response`, INFO), and the validator-side
  execution of a rejection (`block_tracker`, DEBUG). No new message counter.
- **Validator proposal rejections:** `block_proposals_received_total` +
  `block_proposals_rejected_total{error_type}` at `handle_block_proposal`, plus a DEBUG log.
- **Client staging failures:** `block_staging_failures_total{error_type}` + INFO log in
  `execute_block`.

The two highest-volume sites (proposal rejected, message rejected) are DEBUG, not INFO:
on conway they run at ~10/sec, ~92% routine consensus churn (`UnexpectedBlockHeight`,
`InvalidTimestamp`, `MissingCrossChainUpdate`); genuinely adversarial rejections are
<0.05/sec. The three low-volume sites are INFO. Per-reason rates live in the metrics.

## Test Plan

- Cherry-picked #6473's squash commit. Conflicts resolved where conway has diverged from
  main: conway lacks main's `impl From<ExecutionError> for LocalNodeError` (kept conway's
  layout, added only `error_type()`), and refactored the closed-chain loop to use
  `self.chain` / `origins_and_inboxes` (kept conway's structure, added the `is_closed`
  capture for the closed-chain log).
- `cargo clippy -p linera-core -p linera-chain --features metrics -- -D warnings` — clean.
- `error_type()` delegation unit tests — pass.
- `cargo fmt` — no changes.
- Full CI on this PR.

## Release Plan

- Observability only; no protocol or storage format change. This is the `testnet` branch;
  the change reaches validators on the next deploy after merge.

## Links

- Backport of #6473
- `error_type` labelling from #5966 (originally #5951 on `main`)

### DIFF
--- a/linera-chain/src/block_tracker.rs
+++ b/linera-chain/src/block_tracker.rs
@@ -18,7 +18,7 @@ use linera_execution::{
     FLAG_FREE_REJECT,
 };
 use linera_views::context::Context;
-use tracing::{debug, instrument};
+use tracing::instrument;
 
 #[cfg(with_metrics)]
 use crate::chain::metrics;
@@ -244,11 +244,6 @@ impl<'resources, 'blobs> BlockExecutionTracker<'resources, 'blobs> {
                         origin: incoming_bundle.origin,
                         posted_message: Box::new(posted_message.clone()),
                     }
-                );
-                debug!(
-                    chain_id = %self.chain_id,
-                    origin = %incoming_bundle.origin,
-                    "Rejecting incoming message"
                 );
                 let mut actor =
                     ExecutionStateActor::new(chain, txn_tracker, self.resource_controller);
--- a/linera-chain/src/block_tracker.rs
+++ b/linera-chain/src/block_tracker.rs
@@ -18,7 +18,7 @@ use linera_execution::{
     FLAG_FREE_REJECT,
 };
 use linera_views::context::Context;
-use tracing::instrument;
+use tracing::{debug, instrument};
 
 #[cfg(with_metrics)]
 use crate::chain::metrics;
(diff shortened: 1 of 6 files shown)


linera-chain/src/block_tracker.rs
```diff
@@ -245,6 +245,11 @@ impl<'resources, 'blobs> BlockExecutionTracker<'resources, 'blobs> {
                         posted_message: Box::new(posted_message.clone()),
                     }
                 );
+                debug!(
+                    chain_id = %self.chain_id,
+                    origin = %incoming_bundle.origin,
+                    "Rejecting incoming message"
```
afck: This is now printed whenever we execute a block containing a rejected message.

The other log, in `data_types`, has similar wording but is where we make the _decision_ to reject a message.

There is at least one more place where we make that decision, I think, which should probably also be logged: When executing a message fails during staging.

afck: I see you removed this one, but I think there are still other places where we _decide_ to choose `Action::Reject`, and we should either log at all or none of those places?

ma2bd (approved):
LGTM

PR #6571, opened 2026-06-30
Stop tracking publisher chains after unsubscribing.

### LINKED ISSUE #4260: Stop tracking chains after unsubscribing. (If that was the only reason for tracking.)

afck:
## Motivation

The chain listener currently keeps listening for updates on publisher chains even after the last subscriber was removed.

## Proposal

Stop listening if the only reason for listening to a chain is that it's a publisher, and the last subscriber unsubscribed.

## Test Plan

Tests were added.

## Release Plan

## Links

- Fixes https://github.com/linera-io/linera-protocol/issues/4260

### DIFF
--- a/linera-client/src/unit_tests/chain_listener.rs
+++ b/linera-client/src/unit_tests/chain_listener.rs
@@ -436,6 +436,168 @@ async fn test_chain_listener_admin_chain() -> anyhow::Result<()> {
     Ok(())
 }
 
+/// Tests that the chain listener stops tracking an event publisher chain once the subscriber
+/// chain unsubscribes from it (#4260).
+#[cfg(any(feature = "wasmer", feature = "wasmtime"))]
+#[cfg_attr(feature = "wasmer", test_case::test_case(linera_execution::WasmRuntime::Wasmer; "wasmer"))]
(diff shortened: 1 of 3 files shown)


linera-client/src/unit_tests/chain_listener.rs
```diff
@@ -436,6 +436,168 @@ async fn test_chain_listener_admin_chain() -> anyhow::Result<()> {
     Ok(())
 }
 
+/// Tests that the chain listener stops tracking an event publisher chain once the subscriber
```
deuszx, line 439: I was expecting that the test would publish to the publisher chain and then assert that the _now-unsubscribed-client_ no longer sees the event.

ma2bd: Wasn't that the case before this PR?

deuszx: I see, so we were already ceasing to listen for events – we just never removed it from the chain listener?

afck: Yes, we would already not have called `process_streams`. This is just removing the listening task, too, now.

deuszx (approved):

Issue #6237, opened 2026-05-06
Epoch revocation handling


afck:
Epoch revocations are currently handled per-chain, and there is no logic in place for the client updating validators about blocks from revoked epochs using indirect proofs (i.e. descendant blocks or even recipient blocks referring to the old block by hash). We should:

* [x] Remove the per-chain committee map; revocations should be immediate.
* [ ] Add logic for indirect proofs. Add a test where a client joins a network with revoked epochs and then updates validators using an indirect proof. (https://github.com/linera-io/linera-protocol/pull/6591)
* [ ] #6594

PR #6572, opened 2026-06-30
Make safety faults attributable based on conflicting certificates.

afck:
## Motivation

Only including the final `ConfirmedBlock` votes (_C-votes_) isn't enough in general: A validator that signed two conflicting blocks in two _different_ rounds isn't necessarily violating the rules of the protocol!

## Proposal

`ValidatedBlock` votes (_V-votes_) in round s now include an optional round number l < s that means:
* I am voting to validate this block in round s.
* If l is specified: It got a quorum of V-votes in round l and I have not signed a C-vote for a different block in any round r in between (l ≤ r < s).
* If there is no l (`None`): I have not signed any C-vote for a different block in any r < s.

A `ConfirmedBlock` certificate now contains not only a quorum of C-votes in some round s, but also a quorum of V-votes in the same round. If those V-votes have a round l, it also contains a quorum of V-votes from l, etc., until it reaches a quorum without l.

That way, each pair of conflicting `ConfirmedBlock` certificates will contain evidence that a third of the validators violated either of two rules:
* They signed two conflicting C-votes in the same round.
* They signed a V-vote with l in round s, but did sign a conflicting C-vote in a round l < r < s.

As an optimization for the happy path (confirmation in the first round), the C-vote payload now contains a `first_round` flag, which is set simply if these C-votes are in the first round. (Which one the first round is depends on the chain's ownership config.) That way, in the first round we can always omit the V-vote quorum. However, a new type of equivocation is:
* A validator signed a C-vote with `first_flag` in round s, but _also_ a C-vote in an earlier round.

## Test Plan

Tests were added and updated.

## Release Plan

## Links

- This is preliminary work for #6237


linera-chain/src/data_types/mod.rs
```diff
@@ -624,7 +688,15 @@ pub struct LiteVote {
     pub value: LiteValue,
     /// The consensus round in which the vote was cast.
     pub round: Round,
-    /// The validator's signature over the value hash, round and certificate kind.
+    /// The lock round `ℓ` this vote signed (see [`VoteValue`]). Only `ValidatedBlock` votes
```
ma2bd: Please do not leak useless greek letters in the comments. As far as the code concerned, the field name is `lock`.

afck: Done in https://github.com/linera-io/linera-protocol/pull/6572/changes/3a128e846fb84c526d5079e54459ad7dd1553a31.

linera-chain/src/certificate/generic.rs
afck, line 18: (We should consider removing this type entirely, but probably in a separate PR.)

linera-chain/src/justification.rs
```diff
@@ -0,0 +1,494 @@
...
+                // means `0`). The confirmation contradicts it only if it falls in that window,
+                // i.e. `unlocking_round ≤ confirmed_round < validated_round`. A confirmation at or
+                // after `validated_round` is a legitimate later switch, not a violation.
+                ensure!(
+                    *confirmed_round < *validated_round
+                        && validated_unlocking_round
+                            .is_none_or(|unlocking_round| *confirmed_round >= unlocking_round),
+                    ChainError::EquivocationProofNoLockViolation
+                );
+                let confirmed = VoteValue(
```
ma2bd, line 444: [Claude says] **Bug: proofs involving first-round confirmations never verify.** This reconstructs the confirmed `VoteValue` with `first_round: false` (same in `DoubleVote::check` at L311), but real first-round C-votes sign `first_round: true` (`manager.rs`: `Vote::new_with_first_round(value, round, round == ownership.first_round(), …)`). So:

- `double_confirm` for two fast-round confirmations — the exact case the fast path exists for — produces a proof whose signature check fails against genuine votes;
- `walk_chain` against a first-round-attested confirmation (e.g. block A confirmed in `MultiLeader(0)`, fork B validated later with an unlocking round reaching back) produces a `LockViolation` that likewise fails `check()`.

The unit tests pass only because the `sign`/`confirmed_signatures` helpers in `justification_tests.rs` sign with `first_round: false`, which real validators never do in a first round. Suggest: carry the attested bit in `JustifiedConfirmation` (available as `certificate.first_round()`) and in both proof variants, use it when rebuilding the `VoteValue`, and add a test whose confirmed signatures are made with `Vote::new_with_first_round(…, true, …)`.

linera-chain/src/justification.rs
```diff
@@ -0,0 +1,494 @@
...
+            }
+        }
+    }
+}
+
+/// Extracts a proof of equivocation from two `ConfirmedBlock` certificates that finalize
+/// *different* blocks at the same height.
+///
+/// Returns `None` only if the inputs do not actually conflict (same block) or are malformed;
+/// for two genuinely conflicting, well-formed certificates a proof always exists.
```
ma2bd: [Claude says] **This claim is false as written, and the gap is the first-round optimization's own fault type.** Consider A confirmed at round r, B confirmed at round s > r with `first_round = true` and an empty chain: `walk_chain(a, b)` finds nothing (empty chain), `walk_chain(b, a)` finds nothing (A's chain tiles only `[0, r)`), and `double_confirm` bails on unequal rounds — `None`, even though the two C-quorums intersect in validators that signed both a first-round attestation at s and a C-vote at r < s. That's the "new type of equivocation" named in the PR description, but `EquivocationProof` has no variant for it and extraction can't produce it.

If this is deferred to #6237, the doc comment should say so explicitly; otherwise attribution is incomplete for exactly the case the attestation introduces. More generally, stating and proving the LibraBFT §9.2-style propositions (safe detection / complete detection) for the three fault types would have surfaced both this and the `first_round`-reconstruction bug mechanically — worth doing in this PR or the follow-up.

linera-chain/src/certificate/lite.rs
```diff
@@ -71,31 +129,55 @@ impl LiteCertificate<'_> {
...
                 signature,
             },
         ) = votes.next()?;
         let mut signatures = vec![(public_key, signature)];
         for (validator_key, vote) in votes {
-            if vote.value.value_hash != value.value_hash || vote.round != round {
+            if vote.value.value_hash != value.value_hash
+                || vote.round != round
+                || vote.unlocking_round != unlocking_round
+                || vote.first_round != first_round
```
ma2bd, line 128: [Claude says] **One Byzantine vote can poison certificate aggregation.** `communicate_with_quorum` groups votes by `(value_hash, round)` only, but this now requires *unanimity* on `unlocking_round` and `first_round` and returns `None` otherwise — surfacing in `communicate_chain_action` as `InternalError("Vote values or rounds don't match; this is a bug")`. A single validator returning a validly-signed vote with a flipped `first_round` or divergent `unlocking_round` (the signature check in `send_chain_update` passes — it signed that value) fails the whole aggregation even though a clean quorum exists.

Suggest either including the two fields in the quorum grouping key, or checking each vote's fields against the expected values in `send_chain_update` and treating a mismatch as that validator's error (like any other bad response).

linera-core/src/chain_worker/state.rs
```diff
@@ -972,10 +972,29 @@ where
...
         certificate.check(&committee)?;
 
+        // The first-round attestation is checkable only where we hold the chain's ownership at
+        // this height: when we are about to execute the block in order (no gap) on an already
+        // active chain, the current ownership is the configuration the block was proposed under,
+        // giving the correct first round. A block above our tip, on a sparsely-tracked chain, or
+        // on a chain we have not yet initialized (its genesis, before we have applied its
+        // description) leaves us without that ownership, so we trust the bit and defer the check
+        // to wherever the chain is executed in order.
+        if !gap && certificate.first_round() && self.chain.is_active().await? {
```
ma2bd: [Claude says] For a chain's **first block** this check is skipped forever: at genesis the chain isn't active yet, so the `is_active()` guard defers "to wherever the chain is executed in order" — but this *is* the in-order execution path, and it never runs for height 0. The initial ownership is derivable from the chain description being applied; consider checking the attestation post-initialization, or documenting explicitly that height-0 attestations are trusted (they're only exploitable alongside a ≥⅓-equivocation fork, but that fork is then unattributable until the missing proof variant exists).

linera-chain/src/certificate/confirmed.rs
```diff
@@ -2,19 +2,198 @@
...
+        if self.quorum().first_round() {
+            ensure!(
+                matches!(
+                    self.round(),
+                    Round::Fast
+                        | Round::MultiLeader(0)
+                        | Round::SingleLeader(0)
+                        | Round::Validator(0)
+                ),
+                ChainError::JustificationUnlockingRoundMismatch
```
ma2bd: [Claude says] Nit: this is the "attestation in a round that can't be a first round" case, so `ChainError::FalseFirstRoundAttestation` (which exists and is used for the execution-time check) fits better than `JustificationUnlockingRoundMismatch` — as-is, a debugging session would point at the wrong invariant.

linera-chain/src/certificate/validated.rs
```diff
@@ -2,17 +2,173 @@
...
+        self.below
+            .append(self.quorum.round(), self.quorum.signatures().clone())
+    }
+
+    /// Verifies the certificate: the quorum's signatures against its unlocking round, the
+    /// justification chain, and that the unlocking round matches the top of the chain.
+    pub fn check(&self, committee: &Committee) -> Result<(), ChainError> {
+        self.quorum.check(committee)?;
+        self.below.verify(self.hash(), committee)?;
+        ensure!(
```
ma2bd: [Claude says] Hardening nit: nothing here enforces `below.top_unlocking_round() < self.round()` — the chain verifies internal ordering and the unlocking round matches the top link, but a cert whose top link is at or above its own round isn't rejected. Not exploitable (honest validators never sign `l ≥ s`, and forging the quorum requires ≥⅓ equivocators), but it's a one-line `ensure!` of defense-in-depth, and `ConfirmedBlockCertificate::check` has the analogous `top == round` check.

linera-chain/src/certificate/confirmed.rs
```diff
@@ -2,19 +2,198 @@
...
+/// quorums for the block, making it self-contained evidence for fault attribution.
+#[derive(Clone, Debug, Allocative)]
+#[cfg_attr(with_testing, derive(Eq, PartialEq))]
+pub struct ConfirmedBlockCertificate {
+    /// The signed quorum of `ConfirmedBlock` votes. Its unlocking round is always `None`.
+    quorum: GenericCertificate<ConfirmedBlock>,
+    /// The full chain of validated quorums for the block, rising from the grounding round to its
+    /// top link in the round the block was confirmed. Empty iff the block was confirmed in the
+    /// fast round.
+    validated: JustificationChain,
```
ma2bd: nit: `validated` is a weird name for a field. We expect a noun, preferrably in line with the type `JustificationChain`.

linera-chain/src/certificate/lite.rs
ma2bd: This one looks simple to fix.

linera-chain/src/justification.rs
afck: Fixed in https://github.com/linera-io/linera-protocol/pull/6572/changes/ae6b6a02736a393190b669885d3dcb00df5f3c59.

linera-core/src/chain_worker/state.rs
afck: Fixed in https://github.com/linera-io/linera-protocol/pull/6572/changes/aca7c2dbb3dad1b34a4000723878467a795d1bba.

linera-chain/src/certificate/lite.rs
afck: Fixed in https://github.com/linera-io/linera-protocol/pull/6572/commits/7ed72b33ef2d78aeecccb647a24923389333df5b.

linera-chain/src/certificate/validated.rs
afck: Fixed in https://github.com/linera-io/linera-protocol/pull/6572/changes/e68da08523e8c1f5962014268c9cf70ef780160e.

linera-chain/src/certificate/confirmed.rs
afck: Done: https://github.com/linera-io/linera-protocol/pull/6572/changes/e68da08523e8c1f5962014268c9cf70ef780160e

linera-chain/src/justification.rs
afck: Done: https://github.com/linera-io/linera-protocol/pull/6572/commits/ae6b6a02736a393190b669885d3dcb00df5f3c59

linera-chain/src/certificate/confirmed.rs
afck: Done: https://github.com/linera-io/linera-protocol/pull/6572/changes/e68da08523e8c1f5962014268c9cf70ef780160e

linera-chain/src/data_types/mod.rs
```diff
@@ -951,7 +1034,13 @@ impl<'a, T: CertificateValue> SignatureAggregator<'a, T> {
     where
         T: CertificateValue,
     {
-        let hash_and_round = VoteValue(self.partial.hash(), self.partial.round, T::KIND);
+        let hash_and_round = VoteValue(
```
ma2bd, line 1094: [Claude says] minor: `SignatureAggregator::append` still hardcodes `(None, false)` in the `VoteValue`, so it can no longer aggregate first-round-attested confirmation votes or justified validated votes — any such use would fail signature verification. If it has no remaining callers after this PR (the bridge tests, SDK, and worker tests all moved off it), consider removing it; otherwise it's a footgun waiting for the first caller that touches a non-default vote.

linera-chain/src/certificate/confirmed.rs
```diff
@@ -2,19 +2,203 @@
...
+
+    /// Returns the [`LiteCertificate`] corresponding to this certificate, carrying the chain.
+    pub fn lite_certificate(&self) -> LiteCertificate<'_> {
+        let mut lite = self.quorum.lite_certificate();
+        lite.justification = self.justification.clone();
+        lite
+    }
+}
+
+impl Deref for ConfirmedBlockCertificate {
```
ma2bd, line 138: [Claude says] minor: with this `Deref`, `cert.quorum().lite_certificate()` (or any call that resolves through the deref to `GenericCertificate::lite_certificate`) silently produces a lite certificate with an *empty* justification chain, while `cert.lite_certificate()` carries the real one. Nothing prevents a future call site from taking the wrong path and shipping chainless certificates that fail `check()` at the receiver — a retry-loop liveness bug rather than a compile error. Consider making `GenericCertificate::lite_certificate` private (or renaming it to something like `lite_certificate_without_justification`) so the ambiguity can't compile.

linera-chain/src/justification.rs
```diff
@@ -0,0 +1,629 @@
...
+
+    /// Returns the unlocking round that a certificate sitting on top of this chain signed: the
+    /// round of the highest link, or `None` if the chain is empty.
+    pub fn top_unlocking_round(&self) -> Option<Round> {
+        self.links.last().map(|link| link.round)
+    }
+
+    /// Verifies that every link is a quorum of `ValidatedBlock` votes for `value_hash` and
+    /// that the rounds strictly increase, so each link is properly justified by the previous one.
+    pub fn verify(&self, value_hash: CryptoHash, committee: &Committee) -> Result<(), ChainError> {
```
ma2bd: [Claude says] minor: chain length is unbounded and verification is one batch signature check per link, per receiving node, on the certificate-handling hot path — and certificate size grows the same way. Growth requires a real quorum per contested round, so it's not attacker-cheap, but a height that fights through many rounds could push certificates toward transport limits (with a possible feedback loop: bigger certs → slower finalization → more rounds). A cheap guard: reject chains with more links than the certificate round permits (rounds strictly increase, so `links.len()` is bounded by the round's position in the ladder), plus a metric on observed chain length so contention shows up in monitoring before it becomes an incident.

ma2bd (commented):
[Claude says] Follow-up after re-reviewing the fix commits (`ae6b6a02`, `aca7c2db`, `7ed72b33`, `e68da085`): all three blocking findings and the two smaller inline ones from the earlier review are properly addressed — including details done better than suggested (full signed tuple as the aggregation grouping key; the attestation check moved to `execute_contiguous_block` where it also covers block 0 and checkpoint restores; `FirstRoundViolation` allowing equal blocks with the rationale documented). The remaining points below are all minor.

Two that don't anchor to a line:

- **minor (PR description):** the body still describes the unlocking-round claim window as `l < r < s` (twice), but the code implements — correctly — `l ≤ r < s`: `EquivocationProof::check` uses `confirmed_round >= unlocking_round`, matching the strict `vote.round < certificate.round` manager check, and the inclusive lower bound is what makes consecutive links' windows tile without a gap at each link boundary. Worth updating the text so nobody later "fixes" the `>=`.

- **minor (release plan):** changing `VoteValue` invalidates all existing signatures and the BCS layout of stored/transmitted certificates, so this is a flag-day upgrade — a mixed-version committee cannot form quorums. Fine for `main`'s cycle, but worth a sentence in the release plan (and confirming no backport to `testnet_conway` is intended).

ma2bd (approved):

PR #6577, opened 2026-07-01
Clean up in-flight dedup entries when the request owner is dropped

ndr-ds:
## Motivation

The client's request scheduler deduplicates concurrent requests for the same
data (`RequestsScheduler::deduplicated_request`): the first caller becomes the
*owner* and executes the request, while later callers *subscribe* and wait on a
broadcast channel for the result. `InFlightTracker` holds the map of in-flight
entries; only `complete_and_broadcast` ever removed an entry.

This is not cancel-safe. If the owning task is dropped **before** it completes —
which happens routinely: `synchronize_up_to` runs `download_certificates_from`
against every validator under `communicate_with_quorum`, and once a quorum
responds the remaining (losing) futures are dropped mid-flight — the entry is
never removed. Its broadcast sender stays alive, so every subscriber blocks
**forever** on `receiver.recv().await` (there is no timeout on that wait). A
subscriber living in a different task tree (e.g. the spawned download pipeline)
then starves the whole sync: the chain never advances, the client sits at 0%
CPU with no pending timers, and only a process restart recovers it.

The race is timing-dependent (the subscription join window is
`max_request_ttl`, 200 ms by default), so it is latent under today's request
timing. It is reliably reproducible under workloads that line the cert-download
tasks up in lockstep — see Test Plan.

## Proposal

Make the in-flight tracker cancel-safe with an ownership guard:

- `insert_new` now returns an `InFlightGuard` that owns the entry. The owner
  holds it across the cancellation-prone `try_staggered_parallel().await`.
- Completing via `InFlightGuard::complete_and_broadcast` broadcasts the result
  and removes the entry (unchanged success path).
- If the guard is instead **dropped without completing** (the owning task was
  cancelled), its `Drop` removes the entry. Dropping the entry drops the
  broadcast sender, so subscribers wake with `RecvError::Closed` and fall
  through to executing the request themselves — a path the code already had but
  could never reach while the entry leaked.
- A per-entry `generation` (from an `AtomicU64`) means a guard only ever removes
  *its own* entry, never a newer owner's entry that reused the same key after
  this one went stale and was replaced. Both `complete_and_broadcast` and `Drop`
  are generation-checked, making them idempotent and race-safe.
- When a new owner takes over a key that still has an entry (stale owner or
  insertion race), it adopts that entry's broadcast sender, so existing waiters
  receive the new owner's result instead of being disconnected and retrying.

To let `Drop` clean up synchronously, the entries map moves from
`tokio::sync::RwLock` to `std::sync::Mutex`; it is never held across an `.await`
(the per-entry `alternative_peers` list is cloned out from under the lock before
awaiting). This matches the existing `std::sync::Mutex` usage and discipline in
`linera-core`.

Client-side only; no protocol, storage, or validator change.

## Test Plan

- **New unit regression tests** `test_owner_drop_wakes_subscribers` (an owner
  registers an in-flight request, a second caller subscribes, the owner guard is
  dropped without completing — the subscriber observes `RecvError::Closed`
  rather than hanging, and the entry is gone) and
  `test_new_owner_adopts_existing_waiters` (a waiter subscribed to a replaced
  entry receives the new owner's result; the stale owner can neither broadcast
  nor remove the new entry). `cargo test -p linera-core --lib
  requests_scheduler` → 29 passed; `cargo clippy --all-targets` clean.
- **End-to-end, real workload.** Reproduced the hang with a fresh-wallet,
  empty-DB cold sync of a blob-heavy testnet_conway market chain, and measured a
  client built with vs. without this fix, interleaved under identical network
  conditions (12 cold syncs each, per-run hang watchdog):

  | Build | Cold syncs | Hung |
  |-------|-----------|------|
  | without this fix | 12 | **5 (~42%)** |
  | with this fix | 12 | **0** |

  Every time the unfixed build hung, the very next fixed run on the same
  conditions completed in ~12 s. Stacks/logs from the hung runs confirm the
  root cause above (subscriber parked on `recv().await`, leaked entry, no
  pending timer).

## Release Plan

- These changes should be backported to the latest `testnet` branch (companion
  backport: #6579). Client-side only, so it ships with the usual client/SDK
  release — no validator hotfix.

## Links

- Companion backport: #6579
- Exposed by, and a prerequisite for, the streaming `DownloadBlobs` caller
  change (#6157 / #6476), which reorders request timing enough to make this
  latent race fire on ~42% of cold syncs.

### DIFF
--- a/linera-core/src/client/requests_scheduler/in_flight_tracker.rs
+++ b/linera-core/src/client/requests_scheduler/in_flight_tracker.rs
@@ -27,7 +27,9 @@ use crate::node::NodeError;
 ///
 /// The entries map is guarded by a synchronous [`Mutex`] (never held across an `.await`)
 /// so that an owner's [`InFlightGuard`] can remove its entry from `Drop`, which is what
-/// makes deduplication cancel-safe — see [`InFlightTracker::insert_new`].
+/// makes deduplication cancel-safe — see [`InFlightTracker::insert_new`]. A `Mutex` is
+/// preferred over an `RwLock` because every critical section is only a few map
+/// operations: too short for reader parallelism to pay for the more expensive lock.
 #[derive(Debug, Clone)]
 pub(super) struct InFlightTracker<N> {
     /// Maps request keys to in-flight entries containing broadcast senders and metadata
@@ -105,24 +107,37 @@ impl<N: Clone> InFlightTracker<N> {
     /// dropping the broadcast sender so subscribers wake with `RecvError::Closed`
     /// and execute the request themselves rather than waiting forever.
     ///
+    /// If an entry already exists for this key (its owner went stale or lost an
+    /// insertion race), the new entry adopts its broadcast sender: the existing
+    /// subscribers are waiting for exactly this data, so they receive the new
+    /// owner's result instead of being disconnected. The previous owner's guard
+    /// is invalidated by the generation bump.
+    ///
     /// # Arguments
     /// - `key`: The request key to insert
     /// - `now`: The current time, recorded as the entry's start time
     pub(super) fn insert_new(&self, key: RequestKey, now: Timestamp) -> InFlightGuard<N> {
-        let (sender, _receiver) = broadcast::channel(1);
         let generation = self.next_generation.fetch_add(1, Ordering::Relaxed);
-        self.entries
+        let mut in_flight = self
+            .entries
             .lock()
-            .expect("in-flight tracker mutex poisoned")
-            .insert(
-                key.clone(),
-                InFlightEntry {
-                    sender,
-                    started_at: now,
-                    generation,
-                    alternative_peers: Arc::new(tokio::sync::RwLock::new(Vec::new())),
-                },
-            );
+            .expect("in-flight tracker mutex poisoned");
+        // The owner never listens on this channel — it produces the result itself.
+        // Subscribers create their receivers later via `Sender::subscribe`.
+        let sender = match in_flight.remove(&key) {
+            Some(previous) => previous.sender,
+            None => broadcast::channel(1).0,
+        };
+        in_flight.insert(
+            key.clone(),
+            InFlightEntry {
+                sender,
+                started_at: now,
+                generation,
+                alternative_peers: Arc::new(tokio::sync::RwLock::new(Vec::new())),
+            },
+        );
+        drop(in_flight);
         InFlightGuard {
             entries: self.entries.clone(),
             key,
--- a/linera-core/src/client/requests_scheduler/in_flight_tracker.rs
+++ b/linera-core/src/client/requests_scheduler/in_flight_tracker.rs
@@ -1,7 +1,14 @@
 // Copyright (c) Zefchain Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-use std::{collections::HashMap, fmt::Debug, sync::Arc};
+use std::{
+    collections::HashMap,
+    fmt::Debug,
+    sync::{
+        atomic::{AtomicU64, Ordering},
+        Arc, Mutex,
+    },
+};
 
 use linera_base::{data_types::Timestamp, time::Duration};
 use tokio::sync::broadcast;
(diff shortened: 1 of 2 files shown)


deuszx:
> `synchronize_up_to` runs `download_certificates_from` against every validator under `communicate_with_quorum`, and once a quorum responds

I'd imagine that – if this happens – this also means that particular blob/cert has been downloaded. I.e. we requested a data `X` as part of `synchronize_up_to`, we complete the process (`communicate_*` finishes) so that data must have been received and if so all the "subscribers" should get it as well.

> If the guard is instead dropped without completing (the owning task was cancelled), its `Drop` removes the entry. Dropping the entry drops the broadcast sender

Ideally we'd retry immediately using data from other callers/subscribers instead of going through the error-handling loop. At least that was the plan – that an error/"no data" result uses another caller's endpoint to get the data.

Also, I don't really understand why we need the `generation` – if caller drops the handle it doesn't mean the request is invalid (or that the endpoint/validator doesn't have the data). We don't have to remove the entry – if there are still awaiting callers we can drive it to completion.

afck:
(This seems to be the reason why CI is flaky on https://github.com/linera-io/linera-protocol/pull/6593.)

afck:
So I believe this will fix https://github.com/linera-io/linera-protocol/issues/6596.

Claude:
> This fixes the recurring storage-service-tests CI hang (seen twice in a row on #6593): when a deduplicated request's performing future is dropped — which `communicate_with_quorum` and `hedged_fan_out` do routinely to stragglers — the leaked in-flight entry keeps the broadcast sender alive, so subscribers wait forever; in unit tests the staleness TTL never expires because the storage clock is a frozen `TestClock`, so every later request for the key joins the dead entry and the client wedges silently (`test_execute_block_retries_when_chain_advances` then hangs until the job timeout). Reproduced pre-fix at roughly 1-in-6 per run pinned to a single CPU; with an equivalent fix, 60/60 clean runs under the same contention. Note it's not test-only: with a wall clock, subscribers that join an orphaned entry within the 200 ms TTL window also hang forever. One suggestion: wrap the two-owner race in `test_execute_block_retries_when_chain_advances` in a generous `tokio::time::timeout` so any future wedge fails fast instead of consuming the CI job.

I agree with adding the `timeout`!

linera-core/src/client/requests_scheduler/in_flight_tracker.rs
```diff
@@ -17,10 +24,17 @@ use crate::node::NodeError;
...
 /// information and alternative data sources.
+///
+/// The entries map is guarded by a synchronous [`Mutex`] (never held across an `.await`)
+/// so that an owner's [`InFlightGuard`] can remove its entry from `Drop`, which is what
+/// makes deduplication cancel-safe — see [`InFlightTracker::insert_new`].
 #[derive(Debug, Clone)]
 pub(super) struct InFlightTracker<N> {
     /// Maps request keys to in-flight entries containing broadcast senders and metadata
-    entries: Arc<tokio::sync::RwLock<HashMap<RequestKey, InFlightEntry<N>>>>,
+    entries: Arc<Mutex<HashMap<RequestKey, InFlightEntry<N>>>>,
```
afck, line 36: (Why not `RwLock`? It looks like there are still places where we lock for reading only?)

linera-core/src/client/requests_scheduler/in_flight_tracker.rs
```diff
@@ -81,61 +95,39 @@ impl<N: Clone> InFlightTracker<N> {
...
-                waiters = waiter_count,
-                "request completed; broadcasting result to waiters",
+        let generation = self.next_generation.fetch_add(1, Ordering::Relaxed);
+        self.entries
+            .lock()
+            .expect("in-flight tracker mutex poisoned")
+            .insert(
+                key.clone(),
+                InFlightEntry {
+                    sender,
```
afck: I forgot how this works: If there already is an entry under this `key`, wouldn't it make sense to keep its sender, too, since it's waiting for the same data? (No blocker, since it was already working this way.)

linera-core/src/client/requests_scheduler/in_flight_tracker.rs
```diff
@@ -81,61 +95,39 @@ impl<N: Clone> InFlightTracker<N> {
...
+    /// losing branch of `communicate_with_quorum` — its `Drop` removes the entry,
+    /// dropping the broadcast sender so subscribers wake with `RecvError::Closed`
+    /// and execute the request themselves rather than waiting forever.
     ///
     /// # Arguments
     /// - `key`: The request key to insert
-    pub(super) async fn insert_new(&self, key: RequestKey, now: Timestamp) {
+    /// - `now`: The current time, recorded as the entry's start time
+    pub(super) fn insert_new(&self, key: RequestKey, now: Timestamp) -> InFlightGuard<N> {
         let (sender, _receiver) = broadcast::channel(1);
```
afck: (I also forgot why we drop this `_receiver`: At this point there must already be someone who requested that data and is waiting for it?)

afck (approved):

linera-core/src/client/requests_scheduler/in_flight_tracker.rs
ndr-ds: The owner gets its result directly from executing the request, the channel only serves later subscribers via `sender.subscribe()`

linera-core/src/client/requests_scheduler/in_flight_tracker.rs
ndr-ds: For short critical sections, plain `Mutex` is actually faster
https://eventual-consistency.vercel.app/posts/write-locks-faster

PR #6580, opened 2026-07-01
linera-chain: fold execute_block inputs into a BlockExecution enum

ndr-ds:
## Motivation

Follow-up to #6471, addressing the review comment from afck: after #6471, `ChainStateView::execute_block` took `replaying_oracle_responses`, `policy`, and `phase` as three separate arguments, but their legal combinations are fully determined by the phase. The type therefore permitted illegal inputs, and one coupling was guarded at runtime by an `assert!`:

- oracle responses are `None` when staging or validating a proposal, `Some` when executing a confirmed certificate;
- the bundle-failure policy `AutoRetry` is only legal when staging a proposal (the block is still being built); validating or confirming must `Abort`.

## Proposal

Merge the three arguments into one `BlockExecution` enum whose variants carry exactly the legal data:

- `StageProposal { policy }` — caller-chosen policy (may be `AutoRetry`); never replays.
- `HandleProposal` — always `Abort`; never replays.
- `HandleConfirmed { oracle_responses }` — always `Abort`; replays the recorded responses.

`execute_block` now takes a single `BlockExecution` and derives the oracle-responses/policy internally. Payoff:

- the illegal `(AutoRetry, Some(oracle_responses))` combination is now unconstructable, so the `assert!(replaying_oracle_responses.is_none())` in `execute_block_inner` is **deleted**;
- `execute_block` drops its `#[expect(clippy::too_many_arguments)]` (down to 5 args);
- the two abort call sites no longer pass `committed()` explicitly.

`round` is deliberately left a positional parameter: its `None`-for-confirmed invariant is exclusion-only (both proposal phases carry a round), so folding it in would duplicate the field across variants and delete no guard.

Pure refactor, no behavior change. The `Copy` `BlockExecutionPhase` metric label is kept and derived via `BlockExecution::phase()`.

## Test Plan

- `cargo clippy -p linera-core -p linera-chain --features "metrics test" --all-targets` — 0 warnings.
- `cargo build -p linera-chain -p linera-core` (default features, metrics off) — clean.
- `cargo doc -p linera-chain` with `RUSTDOCFLAGS=-D warnings` — clean (intra-doc links on the new enum).
- `cargo test -p linera-chain --features "metrics test" --lib` — 57 passed.
- `cargo test -p linera-core --features "metrics test" --lib -- test_handle_block_proposal test_handle_certificate` — 20 passed.
- nightly fmt clean.

## Release Plan

- Nothing to do / These changes follow the usual release cycle. (Pure refactor; no protocol or storage format change.)

## Links

- Follow-up to #6471.


linera-chain/src/chain.rs
ma2bd: nit: Even in the age of AI, I would still generate this kind of boilerplate with derive macros (`strum` or `derive_more` comes to mind).

ma2bd (approved):

linera-chain/src/chain.rs
ndr-ds: Using `strum` for it (which we already use in other parts of the code)

PR #6586, opened 2026-07-03
Concurrent, deduplicated sender catch-up in the proposal loop

ma2bd:
## Motivation

Follow-up to #6556 on `testnet_conway` (which already batches validator catch-up via the aggregated `MissingCrossChainUpdates` error). This tightens the **local-node** side of the proposal catch-up loop, which still downloaded the reported missing sender blocks one origin at a time.

## Proposal

In the proposal loop that brings the local node up to date after a `MissingCrossChainUpdates` (`client/mod.rs`):

- Drop the dead `if !bundles.is_empty()` guard (the error is only produced with a non-empty set).
- Deduplicate the reported bundles to the **highest missing height per origin**. `download_sender_block_with_sending_ancestors` walks each origin's message-bearing blocks back from the given height, so the max subsumes the lower ones.
- Download the independent origins **concurrently**, bounded by `max_joined_tasks`, instead of sequentially.

### What this PR deliberately does *not* do

An earlier revision also tried to make the **validator-push** catch-up "sparse" — replacing `send_chain_info_up_to_heights` with `send_chain_info_at_heights` so it would send only the specifically-reported sender heights. That was **reverted** because it breaks the "leave gaps on the validator side" design (#4181), as caught by `test_update_validator_sender_gaps`:

- `send_chain_info_up_to_heights` sources the blocks to send from the client's **local storage** (all message-bearing blocks up to the target); for a hub that only receives from a sender, that's already exactly the message-bearing blocks and nothing else, so it's already sparse in practice.
- `send_chain_info_at_heights` sourced them from the **error's reported heights**, which — by design — omit any message-bearing **ancestor** whose bundle was consumed by an earlier block. A gap block's bundle can only be scheduled once its `previous_message_block` has been executed on the validator, so dropping the ancestor leaves the bundle undeliverable and the retry fails.

The push path is therefore left on `send_chain_info_up_to_heights`, and a memory-level regression test (`test_proposal_catch_up_with_sender_gap`) is added to lock this in.

## Test Plan

- `test_proposal_catch_up_with_sender_gap` (new): a sender that messages the recipient at heights 0 and 2 but not 1, consumed across two recipient blocks, with a lagging validator that must be caught up across the gap. Fails with the reverted sparse push, passes with `send_chain_info_up_to_heights`.
- `test_proposal_batches_missing_dependency_catch_up` and `test_handle_block_proposal_with_incoming_bundles` (aggregated-error coverage) pass, as does the broader cross-chain/inbox suite.
- `cargo clippy -p linera-core --lib --tests` and `cargo +nightly fmt` are clean.

## Release Plan

    - be released in a new SDK,

## Links

- Follow-up to #6556. Main counterpart: #6573.

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -2155,33 +2155,41 @@ impl<Env: Environment> Client<Env> {
                         continue;
                     }
                 }
-                // The local node reports *every* missing sender in a single
-                // `MissingCrossChainUpdates`, so we download them all in one pass and retry once;
-                // there is no need to iterate the discovery one sender at a time.
+                // The local node reports every missing sender bundle in a single
+                // `MissingCrossChainUpdates`, so we download them all in one pass and retry once.
                 if let LocalNodeError::WorkerError(WorkerError::ChainError(chain_err)) = &err {
                     if let ChainError::MissingCrossChainUpdates { chain_id, bundles } = &**chain_err
                     {
                         let chain_id = *chain_id;
-                        // Clone to end the borrow of `err` before we reassign it below.
-                        let bundles = bundles.clone();
-                        if !bundles.is_empty() {
-                            for (origin, height) in bundles {
-                                self.download_sender_block_with_sending_ancestors(
-                                    chain_id,
-                                    origin,
-                                    height,
-                                    remote_node,
-                                )
-                                .await?;
-                            }
-                            if let Err(new_err) =
-                                Box::pin(self.local_node.handle_block_proposal(proposal.clone()))
-                                    .await
-                            {
-                                err = new_err;
-                            } else {
-                                continue 'proposal_loop;
-                            }
+                        // `download_sender_block_with_sending_ancestors` walks each origin's
+                        // message-bearing blocks back from the given height, so the highest missing
+                        // height per origin subsumes the lower ones. Deduplicate to that (also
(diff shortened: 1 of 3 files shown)


linera-core/src/client/mod.rs
```diff
@@ -2155,33 +2155,41 @@ impl<Env: Environment> Client<Env> {
...
-                                err = new_err;
-                            } else {
-                                continue 'proposal_loop;
-                            }
+                        // `download_sender_block_with_sending_ancestors` walks each origin's
+                        // message-bearing blocks back from the given height, so the highest missing
+                        // height per origin subsumes the lower ones. Deduplicate to that (also
+                        // ending the borrow of `err` so we can reassign it below), then download the
+                        // independent origins concurrently, bounded by `max_joined_tasks`.
+                        let mut origin_heights: BTreeMap<ChainId, BlockHeight> = BTreeMap::new();
```
afck, line 2169: (Type annotation; but maybe we stop enforcing that?)

afck (approved):

linera-core/src/client/mod.rs
ma2bd: yes it's an uphill battle until someone figures how to include a default skill and/or CI checks

ma2bd: on the plus side it's also easy to refactor later

PR #6589, opened 2026-07-04
Retain the block owner's proposal signature alongside certificates (#456)

### LINKED ISSUE #456: Do not erase signature of block owners
Needed for auditing

Bonus point for not hashing the signature as part of the certificate.

ma2bd:
## Motivation

Fixes #456. When a block proposal is turned into a certificate, the chain owner's signature is discarded — a confirmed block retains `authenticated_owner` but nothing backs it. As a result, the authorization of a block (and of the `authenticated_owner` claims on its outgoing messages) is attested only by the validator quorum: an auditor replaying the chain, or any party distrusting a (possibly revoked) committee, cannot verify that a chain owner ever authorized the block.

## Proposal

Make the owner's signature over the block a first-class object, retained on the block itself, and make it a validity condition for blocks that declare an `authenticated_owner`:

- New `OwnerAuthorization { round, signature }`: the owner's signature over the *outcome-less* `ProposalContent` of the block. Every block's first proposal has `outcome: None`, and that signature is exactly the one validators check against `authenticated_owner` at proposal time. Given the block, anyone can reconstruct the signed content (`Block::to_proposed()` + the recorded round) and verify it — no extra data needed.
- **Proposals carry it directly** (per review feedback): `BlockProposal` gains an `owner_authorization` field and a `validated_certificate` field, replacing the `OriginalProposal` enum — `OriginalProposal::Fast(signature)` was byte-for-byte an `OwnerAuthorization { round: Fast, signature }`, and `Regular` was just the lite certificate. When the proposer is the authenticated owner and the signed content has no outcome (the common case), the proposal signature doubles as the authorization and the field stays empty. The worker's three-way proposal match collapses into one uniform rule; in this PR, an explicit authorization is accepted only on retries (fast-round or validated), so proposal semantics are unchanged.
- **The certified block retains it** (per review feedback): `Block` gains an `owner_authorization` field, sitting *outside* `BlockHeader` and therefore outside `Block::hash()`, which commits to the header only. The certificates carry nothing extra: `ConfirmedBlock::new(validated.block().clone())` moves the evidence from the validated to the confirmed certificate for free, and storage — which decomposes a certificate into (lite certificate bytes, block bytes) — preserves it without help.
- **Why the block hash must not cover it.** `Block` stays a pure function of `(ProposedBlock, BlockExecutionOutcome)`, which several rules silently depend on. `ChainManager::check_proposed_block` implements "without a certificate, only a retry of the very block we confirmed in the fast round is acceptable" as `matches_proposed_block`, comparing the seven `ProposedBlock` fields; that is sound only because same proposed block + deterministic execution ⟹ same hash. If the round-dependent signature entered the hash, a validator that confirmed a block in the fast round would accept a re-proposal carrying a freshly signed authorization as a *different* block and vote to validate it with `unlocking_round: None` — a claim contradicted by its own confirmation, i.e. exactly the attributable fault `extract_equivocations` mines. Two semantically identical blocks would then look like a fork. The authorization is also not covered by the proposer's signature (it cannot be, without circularity), so it is malleable in transit; keeping it out of the hash makes that harmless.
  Correspondingly, `PartialEq`/`Eq`/`Hash` on `Block` compare the header and body only, so equality agrees with `hash()`: a block re-proposed in a later round with a different but equally valid authorization is the *same* block.
- **Validity rule**: a certified block with `authenticated_owner: Some(owner)` is rejected (`ChainError::MissingOwnerAuthorization`) unless it retains a valid signature by `owner`. Enforced in `process_validated_block`, `process_confirmed_block`, and on retry proposals — so invalid or missing evidence can't enter locking blocks. Blocks without an authenticated owner may omit the signature, but if present it must verify. This rule is deliberately self-contained: it needs only the block, so anyone holding one can apply it.
- **Entitlement stays an execution-state question**, checked at proposal time rather than by the rule above. Whether the signer was allowed to authorize in the round they claim depends on the chain's ownership at that height, which `check_block_authorization` has no access to. So the worker checks it next to the existing `can_propose` call: an authorization claiming `Round::Fast` must be signed by a super owner, since only super owners propose in the fast round. That gate is what lets the manager rebuild a `LockingBlock::Fast` from such an authorization. An auditor replaying the chain re-derives ownership and can apply the same rule for any round.
- Wire format: `optional bytes owner_authorization` on the gRPC `BlockProposal` message; the `Certificate` and `LiteCertificate` messages need no such field, since the authorization travels inside the serialized block. BCS formats updated accordingly (`formats.yaml` regenerated) — the evidence now appears once, on `Block`, rather than on each of the three certificate types.

Note: this changes the stored and transmitted block and proposal formats, so it requires a new deployment; previously stored certificates do not deserialize. Every block also grows by the `Option` tag byte, so `EMPTY_BLOCK_SIZE` — which the `maximum_block_size` policy counts against — goes from 94 to 95.

The authorization is **exposed over GraphQL** along with the rest of the block: `Round` was already a scalar, and `AccountSignature` becomes one via `doc_scalar!` — the serde-based macro, like the per-scheme signature scalars it wraps. That symmetry is load-bearing rather than cosmetic: clients deserialize GraphQL responses with serde (the indexer goes through `post_graphql::<Block, _>`), so a scalar whose `ScalarType` encoding differs from its serde one cannot be read back. `bcs_scalar!` would be such a case here — it is only symmetric for types like `AccountOwner` that also carry a matching human-readable serde impl, which `AccountSignature` does not. `OwnerAuthorization` is a `SimpleObject` over the two scalars, so `block { ownerAuthorization { round signature } }` resolves, and blocks read back through `linera-service-graphql-client` carry their evidence.

One consequence of this placement is worth calling out: the evidence is non-hash-committed payload on a hash-addressed object, so `db_storage`'s block cache — keyed by block hash — cannot distinguish two copies of a block that differ only in their authorization. Validity is always judged against the certificate's own deserialized block, never the cache, and in honest operation every copy carries the same signature, so this is benign; it is the structural cost of retaining the evidence on the block rather than on the certificate.

## Test Plan

- Unit test `owner_authorization_verifies_against_block`: signature round-trip, wrong round, wrong signer vs. `authenticated_owner`, missing authorization rejected for authenticated blocks and accepted otherwise, and that attaching the authorization changes neither the block's hash nor its equality with the unauthorized block.
- Worker test `test_handle_certificate_without_owner_authorization`: stripping the authorization from a valid confirmed certificate leaves its hash and validator signatures intact, yet the worker rejects it with `MissingOwnerAuthorization`; the same certificate with the authorization retained is accepted.
- Client tests assert end-to-end retention: the fast-path certificate's block carries a verifying authorization, and in `test_finalize_locked_block_with_blobs` the block finalized by client 2B still verifies against owner 2A's signature (carried across the validated certificate and the retry) — with no client-side carry-over code, since the block itself moves from the validated to the confirmed certificate.
- `check_that_validators_have_certificate` compares the retained authorization across validators explicitly, since certificate equality — which compares value hashes — does not cover it.
- `account_signature_graphql_scalar_matches_serde` pins the invariant above for all three signature schemes: the scalar's `to_value` equals its serde value, and `parse` round-trips.
- The exported GraphQL schema (`service_schema.graphql`) is regenerated and matches `linera-schema-export`.
- Worker tests use an owner-key registry in `TestEnvironment` so self-certified test blocks carry valid authorizations; fast retries exercise the unified proposal path.
- The GraphQL round-trip is exercised end to end by `test_end_to_end_operations_indexer`, which fetches every block through `post_graphql::<Block, _>` and deserializes it with serde — the test that catches a scalar whose GraphQL encoding drifts from its serde one.
- `cargo test -p linera-base -p linera-chain -p linera-core -p linera-rpc -p linera-storage -p linera-sdk` all pass; clippy clean including `--all-features`; formatted with the pinned CI nightly.

## Release Plan

- Backporting is not possible but we may want to deploy a new `devnet` and release a new SDK soon.

## Links

- Closes #456
- Follow-up: #6590 (allow any round-eligible owner to propose an authorized block)


linera-chain/src/certificate/generic.rs
```diff
@@ -20,6 +23,11 @@ pub struct GenericCertificate<T: CertificateValue> {
     /// The round in which the value was certified.
     pub round: Round,
     signatures: Vec<(ValidatorPublicKey, ValidatorSignature)>,
+    /// The chain owner's signature over the certified block's proposal content.
+    /// Not covered by the certified value's hash or the validator signatures.
+    /// Required for blocks with an `authenticated_owner`; optional otherwise.
+    /// Always `None` for timeouts.
+    owner_authorization: Option<OwnerAuthorization>,
```
afck: I'm a bit surprised this is in `GenericCertificate` and not in `Block`. It certainly doesn't apply to timeout certificates, at least?

linera-chain/src/certificate/lite.rs
```diff
@@ -28,6 +28,10 @@ pub struct LiteCertificate<'a> {
     pub round: Round,
     /// Signatures on the value.
     pub signatures: Cow<'a, [(ValidatorPublicKey, ValidatorSignature)]>,
+    /// The retained chain owner's signature over the certified block's proposal
+    /// content, if available. Carried here so that reconstructing a full
+    /// certificate from a cached value does not lose it.
+    pub owner_authorization: Option<OwnerAuthorization>,
```
afck: …then we also wouldn't need it here, would we?

linera-chain/src/data_types/mod.rs
```diff
@@ -437,9 +437,96 @@ pub enum OriginalProposal {
...
+    /// propose in `round` under the chain's ownership at that height can only be
+    /// checked against the chain's execution state and is left to the caller.
+    pub fn verify(&self, block: &Block) -> Result<AccountOwner, ChainError> {
+        self.verify_proposed_block(block.to_proposed())
+    }
+
+    /// Checks the authorization rule for a certified block: a block with an
+    /// `authenticated_owner` is only valid together with that owner's signature over
+    /// the block's proposal content. For blocks without an authenticated owner the
+    /// signature is optional, but must be valid if present.
```
afck, line 484: (I wonder if we should just make it non-optional? There's always a signature by the original proposer.)

linera-core/src/client/chain_client/mod.rs
```diff
@@ -2289,13 +2294,23 @@ impl<Env: Environment> ChainClient<Env> {
         let certificate = if round.is_fast() {
             let hashed_value = ConfirmedBlock::new(block);
             self.client
-                .submit_block_proposal(committee.clone(), proposal, hashed_value)
+                .submit_block_proposal(
+                    committee.clone(),
+                    proposal,
+                    hashed_value,
+                    owner_authorization,
```
afck: Why not make the owner authorization part of the block proposal itself?

afck: If I'm reading that correctly, we do that in #6590 anyway?

linera-chain/src/data_types/mod.rs
ma2bd: My plan is that some chain owners will not be allowed to be "authenticated owners" (which is cleaner semantically that forbidding operations)

linera-chain/src/certificate/generic.rs
ma2bd: The original github issue was suggesting not to hash the user signature, but we can pick what makes the code simpler.

linera-core/src/client/chain_client/mod.rs
ma2bd: indeed

linera-chain/src/data_types/mod.rs
afck: OK.
But it might still make sense to preserve their signature anyway, even if it doesn't get passed on to any contracts as the authenticated owner? Just to document who put together this particular set of messages and operations?

ma2bd: 1) If the authenticated owner is not set, any valid block proposer can re-propose the same block from scratch. The block doesn't "belong" to anybody in my view.
2) If we keep the signature, we must verify it
3) Currently, the authenticated owner is always set

deuszx: I thought this PR/change is about retaining the signature of the `BlockProposal` signer. One can't propose a block on a chain w/o signing (proving ownership of the chain). How is it going to work if there will be no signature? Who is authorising and being authenticating?

linera-chain/src/certificate/generic.rs
afck: Right; I don't see a downside with hashing it, and simply making it part of the `Block`. And it would make more sense to me conceptually, rather than having it as part of the certificate.

linera-chain/src/data_types/mod.rs
afck: Of course, the proposer still signs it. The question here is just whether to preserve the signature when the block gets committed.

linera-chain/src/certificate/lite.rs
```diff
@@ -56,6 +56,10 @@ pub struct LiteCertificate<'a> {
     pub justification: Cow<'a, JustificationChain>,
     /// Signatures on the value.
     pub signatures: Cow<'a, [(ValidatorPublicKey, ValidatorSignature)]>,
+    /// The retained chain owner's signature over the certified block's proposal
+    /// content, if available. Carried here so that reconstructing a full certificate
+    /// from a cached value does not lose it.
+    pub owner_authorization: Option<OwnerAuthorization>,
```
afck: (I was referring to this one earlier: Could this be in `Block` rather than `[Lite]Certificate`?)

PR #6591, opened 2026-07-06
Recertify ancestor blocks; re-enable epoch revocation.

afck:
## Motivation

  We need to allow revoking old epochs. Otherwise validators are required to be
  trusted and keep their keys safe forever, without any time limit.

  Revocation also makes parts of the received log dead weight: entries pointing
  at revoked-epoch certificates are useless to clients, since those certificates
  can no longer be verified on their own. Without pruning, the log grows without
  bound.

  ## Proposal

  In the chain worker, don't trust certificates from revoked epochs unless
  something we already trust vouches for them. Vouching is tracked in a
  `vouched_blocks` set (generalizing the checkpoint trust marks): whenever a
  block is accepted, the blocks it commits to via `previous_block_hash` and
  `previous_message_blocks` are recorded as vouched, unless they are already
  known. Accepting a vouched block consumes the mark and vouches for *its*
  references in turn, so trust propagates backwards transitively. If a trusted
  block commits to a different hash than one already accepted at the same
  height, this is evidence of equivocation and gets logged.

  Also, stop signing on chains in revoked epochs.

  In the client, re-establish trust in a rejected revoked-epoch block by
  processing newer blocks first, in decreasing height order. For sender chains
  this follows the `previous_message_blocks` links, so only the message-bearing
  blocks are downloaded; walking all contiguous descendants remains the fallback
  when the sender never sent to us again from a live epoch.

  Finally, keep the received log partitioned by the epoch of the sender block's
  certificate, ordered numerically in storage:

  - Clients request all live epochs' logs in a single query (a map from epoch to
    tracker offset), and validators serve them under one total entry bound per
    response.
  - Received-certificate trackers are kept per validator *and* epoch, in a map
    view instead of one ever-growing register blob.
  - Validators drop a revoked epoch's log wholesale the next time a client asks
    for a live epoch's log on that chain, and tracker entries for revoked epochs
    are removed whenever trackers are updated. Since revocation is monotone,
    both prunings need only a single revocation check.

  ## Test Plan

  Tests were added, including: re-certification of a revoked-epoch sender block
  via `previous_message_blocks` without the worker ever seeing the intermediate
  non-sender block; and the sparse-sender-chain client test now crosses a
  revoked epoch and asserts that recovery does not fall back to contiguous
  downloads and that validators prune the revoked epoch's log.

  ## Release Plan

  - These changes are protocol-breaking (chain-state layout, RPC query format)
    and must not be backported to the current testnet; they follow the usual
    release cycle for the next network.

  ## Links

  - Part of #6237.

### DIFF
--- a/linera-chain/src/chain.rs
+++ b/linera-chain/src/chain.rs
@@ -352,8 +352,9 @@ where
     /// Hashes of blocks that are vouched for by data this chain already trusts, but
     /// have not been processed here yet. A hash is inserted when a trusted source
     /// commits to it: a checkpoint cert's `outbox_block_hashes`, or an accepted
-    /// block's `previous_block_hash` or `previous_message_blocks` links (unless the
-    /// referenced block is already in `block_hashes`). The worker accepts a cert
+    /// block's `previous_block_hash`, `previous_message_blocks` or
+    /// `previous_event_blocks` links (unless the referenced block is already in
+    /// `block_hashes`). The worker accepts a cert
     /// whose hash is in this set regardless of its epoch — the epoch may be revoked,
     /// but the hash commitment makes the block trustworthy — and removes the entry
     /// upon acceptance.
@@ -1538,8 +1539,9 @@ where
     }
 
     /// Records the block references an accepted block commits to — its parent via
-    /// `previous_block_hash` and the latest message block per recipient via
-    /// `previous_message_blocks` — in `vouched_blocks`, unless the referenced block
+    /// `previous_block_hash`, the latest message block per recipient via
+    /// `previous_message_blocks` and the latest event block per stream via
+    /// `previous_event_blocks` — in `vouched_blocks`, unless the referenced block
     /// is already in `block_hashes`. The accepted block is trusted, so these hash
     /// commitments remain a sufficient basis for accepting the referenced blocks
     /// even after the epochs that certified them are revoked.
@@ -1556,7 +1558,12 @@ where
         if let Some(parent_hash) = block.header.previous_block_hash {
             references.push((block.header.height.try_sub_one()?, parent_hash));
         }
-        for (hash, height) in block.body.previous_message_blocks.values() {
+        for (hash, height) in block
+            .body
+            .previous_message_blocks
+            .values()
+            .chain(block.body.previous_event_blocks.values())
+        {
             references.push((*height, *hash));
         }
         let known_hashes = self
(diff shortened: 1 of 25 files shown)


linera-core/src/chain_worker/state.rs
```diff
@@ -328,6 +328,53 @@ where
...
+        }
+        let child_height = header.height.try_add_one()?;
+        if let Some(child_hash) = self.chain.block_hashes.get(&child_height).await? {
+            let child = self.storage.read_confirmed_block(child_hash).await?;
+            if child
+                .is_some_and(|child| child.block().header.previous_block_hash == Some(block_hash))
+            {
+                return Ok(());
+            }
+        }
```
ma2bd: Why are we testing if an immediate child block is in storage?

afck: Because that means we have already accepted (i.e. verified at some time in the past) the child block, so we know this block is correct, too, even if its committee isn't trusted anymore.

deuszx: So if we get a child in epoch `N+1` when epoch `N` got revoked, then we can trust `N` epoch block? If so, why only the latest one and not all blocks in that epoch?

afck: Exactly: All blocks in the earlier epoch that are directly or indirectly recertified by the later block!

In fact, even if the child is in epoch `N`, but we checked it while `N` was still trusted, that means we should now also trust all ancestors of the child.

But the only way to verify those ancestors now is to check one hash after another. We can't skip any.

afck:
Note that the sparse re-certification only works for messages, not events, for now. But since the PR is already quite large, I'd add that as a follow-up later.

linera-chain/src/chain.rs
```diff
@@ -339,14 +349,15 @@ where
...
-    pub pre_checkpoint_block_trust: SetView<C, CryptoHash>,
+    /// Hashes of blocks that are vouched for by data this chain already trusts, but
+    /// have not been processed here yet. A hash is inserted when a trusted source
+    /// commits to it: a checkpoint cert's `outbox_block_hashes`, or an accepted
+    /// block's `previous_block_hash` or `previous_message_blocks` links (unless the
+    /// referenced block is already in `block_hashes`). The worker accepts a cert
+    /// whose hash is in this set regardless of its epoch — the epoch may be revoked,
+    /// but the hash commitment makes the block trustworthy — and removes the entry
+    /// upon acceptance.
+    pub vouched_blocks: SetView<C, CryptoHash>,
```
ma2bd, line 361: should this be in shared storage so that all chains benefit from it?

afck: How would other chains benefit from this chain's vouched blocks, if we don't have the blocks yet?

afck:
I split this up into three parts.
Closing in favor of #6601, #6603 and #6604.

PR #6595, opened 2026-07-07
`ConfirmedBlock` vote ledger

afck:
## Motivation

To defend against long-range attacks where validators make up chains that were never legitimately created, we need all validators to commit to the set of `ConfirmedBlock` votes they made in each epoch.

As an optimization, we can leave out votes for blocks that got confirmed, if we already made a vote for one of their descendants.

Currently, votes are ephemeral entries in the chain manager that get deleted after each block height.

## Proposal

Store the votes in shared storage.

Use 256 buckets (depending on chain ID); with that, ScyllaDB should support hundreds of millions of chains.

## Test Plan

Tests were added.

## Release Plan

## Links

- Preparation for https://github.com/linera-io/linera-protocol/issues/6594.


afck:
Let's split this into smaller parts.
Closing in favor of #6610.

PR #6601, opened 2026-07-13
Re-certify revoked-epoch blocks via their descendants

afck:
## Motivation
  
Once an epoch is revoked, its committee's signatures are no longer trustworthy: a certificate from a revoked epoch can't be verified on its own anymore. But nodes still need to accept old blocks — a sender's block with an undelivered message, for example — long after the epoch that certified it is gone. That is why epoch revocation is currently not usable.

  ## Proposal

Accept a confirmed block from a revoked epoch only if something the node already trusts vouches for it. A block certified in a live epoch commits to its parent's hash, so its descendants re-certify it: the worker records the hash of a block it trusts but hasn't seen yet, and accepts a certificate for that hash regardless of its epoch.

Clients re-certify accordingly: when a validator rejects a block as belonging to a revoked epoch, the client downloads that block's descendants (racing validators, so a single faulty one can't stall the sync) and submits them first. Validators also refuse to sign new blocks in a revoked epoch.

  ## Test Plan

CI, plus new unit tests in `linera-core` covering worker rejection and acceptance of revoked-epoch blocks and the client's descendant-based recovery.

## Release Plan

## Links

  - First of three stacked PRs. Followed by #6603 and #6604.

### DIFF
--- a/linera-core/src/chain_worker/state.rs
+++ b/linera-core/src/chain_worker/state.rs
@@ -375,6 +375,36 @@ where
         })
     }
 
+    /// Ensures that this worker may still create signatures for blocks in the given
+    /// epoch: once an epoch has been revoked on the admin chain, its committee must
+    /// not sign anything new. A chain that failed to migrate to a newer epoch before
+    /// the revocation is frozen.
+    async fn ensure_epoch_signable(
+        &self,
+        epoch: Epoch,
+        height: BlockHeight,
+    ) -> Result<(), WorkerError> {
+        let is_revoked = self
+            .storage
+            .is_epoch_revoked(epoch)
+            .await
+            .map_err(|error| {
+                WorkerError::ChainError(Box::new(ChainError::ExecutionError(
+                    Box::new(error),
+                    ChainExecutionContext::Block,
+                )))
+            })?;
+        ensure!(
+            !is_revoked,
+            WorkerError::VoteInRevokedEpoch {
+                chain_id: self.chain_id(),
+                epoch,
+                height,
+            }
+        );
+        Ok(())
+    }
+
     /// Returns whether this chain is known to be active (initialized).
     pub(crate) fn knows_chain_is_active(&self) -> bool {
         self.knows_chain_is_active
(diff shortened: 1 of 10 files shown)


linera-core/src/chain_worker/state.rs
```diff
@@ -328,6 +328,53 @@ where
...
+                WorkerError::ChainError(Box::new(ChainError::ExecutionError(
+                    Box::new(error),
+                    ChainExecutionContext::Block,
+                )))
+            })?;
+        if !is_revoked {
+            return Ok(());
+        }
+        let block_hash = certificate.hash();
+        if self.chain.block_hashes.get(&header.height).await? == Some(block_hash) {
```
deuszx, line 359: If we check this first (before `storage.is_epoch_revoked`) and it returned `Some(block_hash)` – doesn't it mean that we don't need to check the storage? Or put differently – can't we do just this, w/o the storage check.

linera-core/src/chain_worker/state.rs
```diff
@@ -328,6 +328,53 @@ where
...
+        }
+        let child_height = header.height.try_add_one()?;
+        if let Some(child_hash) = self.chain.block_hashes.get(&child_height).await? {
+            let child = self.storage.read_confirmed_block(child_hash).await?;
+            if child
+                .is_some_and(|child| child.block().header.previous_block_hash == Some(block_hash))
+            {
+                return Ok(());
+            }
+        }
```
deuszx, line 370: In theory, we might have some "later child" (non-direct) which would re-certify the `certificate` as well?

afck: But a later descendant probably doesn't mention this block by hash. At least not yet! That's part of https://github.com/linera-io/linera-protocol/pull/6603.

deuszx: hm, right. Somehow I thought we can trust the `certificate` b/c child referes to it indirectly but if that was the case we'd need to have those intermediate blocks as well.

linera-core/src/chain_worker/state.rs
afck: Actually, the most common case (new block at the tip) is that this `get` (which is also a DB read) returns `None` and we still have to check the epoch then, which passes. So for now it's better to do the epoch check first.

Anyway, this will also be simplified considerably in #6603.

PR #6603, opened 2026-07-13
Vouch for the blocks a block refers to

afck:
## Motivation

Re-certifying a revoked-epoch block by walking its descendants means downloading every block in between, even though almost none of them are needed: on a sparse chain, a receiver may care about two message blocks a thousand heights apart.

## Proposal

Blocks already commit to the hashes of the blocks they refer to: `previous_message_blocks` points at the last block that sent a message to a given recipient, `previous_event_blocks` at the last block that emitted an event on a given stream. Treat those the same way as the parent hash: when a block is accepted, vouch for everything it commits to. A receiver can then re-certify an old sender block by downloading only the message-bearing blocks, and a subscriber only the event-bearing ones — preprocessing them newest-first, so each block vouches for the next-older one.

A block that commits to a different hash than the one already accepted at that height is proof that a committee equivocated. Reject it before it is written to storage, and log the conflict at ERROR level.

## Test Plan

CI, plus unit tests that re-certify a revoked-epoch block through a `previous_message_blocks` link and through a `previous_event_blocks` link, asserting in both cases that the blocks in between are never downloaded. Conflicting certified blocks are covered by a worker test.

## Release Plan

## Links

  - Stacked on #6601.

### DIFF
--- a/linera-chain/src/chain.rs
+++ b/linera-chain/src/chain.rs
@@ -93,10 +93,11 @@ pub(crate) mod metrics {
     use std::sync::LazyLock;
 
     use linera_base::prometheus_util::{
-        exponential_bucket_interval, register_histogram_vec, register_int_counter_vec,
+        exponential_bucket_interval, register_histogram_vec, register_int_counter,
+        register_int_counter_vec,
     };
     use linera_execution::ResourceTracker;
-    use prometheus::{HistogramVec, IntCounterVec};
+    use prometheus::{HistogramVec, IntCounter, IntCounterVec};
 
     pub static NUM_BLOCKS_EXECUTED: LazyLock<IntCounterVec> = LazyLock::new(|| {
         register_int_counter_vec(
@@ -203,6 +204,14 @@ pub(crate) mod metrics {
             "Number of entries in the outbox_counters map (in-flight message heights)",
             &[],
             exponential_bucket_interval(1.0, 1_000_000.0),
+        )
+    });
+
+    pub static NUM_CONFLICTING_CERTIFIED_BLOCKS: LazyLock<IntCounter> = LazyLock::new(|| {
+        register_int_counter(
+            "num_conflicting_certified_blocks",
+            "Number of detected pairs of conflicting certified blocks, each proving that \
+             a committee equivocated",
         )
     });
 
@@ -1526,13 +1535,11 @@ where
         Ok(updated_streams)
     }
 
-    /// Records the block references an accepted block commits to — its parent via
-    /// `previous_block_hash`, the latest message block per recipient via
-    /// `previous_message_blocks` and the latest event block per stream via
-    /// `previous_event_blocks` — in `vouched_blocks`, unless the referenced block
-    /// is already in `block_hashes`. The accepted block is trusted, so these hash
-    /// commitments remain a sufficient basis for accepting the referenced blocks
-    /// even after the epochs that certified them are revoked.
+    /// Records the blocks an accepted block commits to ([`Block::vouches_for`]) in
+    /// `vouched_blocks`, unless the referenced block is already in `block_hashes`.
+    /// The accepted block is trusted, so these hash commitments remain a sufficient
+    /// basis for accepting the referenced blocks even after the epochs that
+    /// certified them are revoked.
     ///
     /// A chain has one certified block per height, so a block committing to a
     /// different hash than the one already accepted at that height is evidence of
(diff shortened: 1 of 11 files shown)


linera-chain/src/chain.rs
```diff
@@ -1522,6 +1526,77 @@ where
...
+            .values()
+            .chain(block.body.previous_event_blocks.values())
+        {
+            references.push((*height, *hash));
+        }
+        let known_hashes = self
+            .block_hashes
+            .multi_get(references.iter().map(|(height, _)| height))
+            .await?;
+        for (known, (height, hash)) in known_hashes.into_iter().zip(references) {
```
deuszx, line 1556: it's not immedaiately obvious to me that it's safe to zip `known_hashes` with `references` here and they will be matching height-wise. It'be cleaner if `references` was a map from height to hash and `known_hashes` where also the map. Then we can take the intersection of those two sets to compare the values at duplicated heights.

linera-chain/src/chain.rs
```diff
@@ -1522,6 +1526,77 @@ where
...
+        witness_block: CryptoHash,
+    ) -> ChainError {
+        tracing::error!(
+            chain_id = %self.chain_id(),
+            %height,
+            %existing_block,
+            %conflicting_block,
+            %witness_block,
+            "conflicting certified blocks at the same height: a committee equivocated",
+        );
```
deuszx: I heard once that for every error log line there should be a corresponding metric. And that alerts should be made on that metric, not the log line. I know that we're not using this technique yet but it seems to me that this place is especially important and we should have metrics about it right away.

linera-chain/src/chain.rs
```diff
@@ -1522,6 +1526,77 @@ where
...
+    /// commitments remain a sufficient basis for accepting the referenced blocks
+    /// even after the epochs that certified them are revoked.
+    ///
+    /// A chain has one certified block per height, so a block committing to a
+    /// different hash than the one already accepted at that height is evidence of
+    /// equivocation — two conflicting certificates signed by that height's
+    /// committee. In that case the block is rejected with
+    /// [`ChainError::ConflictingCertifiedBlocks`], which names the certificate
+    /// pair that proves the fault.
+    async fn vouch_for_block_references(&mut self, block: &Block) -> Result<(), ChainError> {
```
deuszx, line 1550: maybe `Block` could get a method like `vouches_for(&self)` that returns an iterator over the `(height, Cryptohash)` ? Just an idea, not sure whether the `Block` should "know" about this concept or not. Feel free to ignore.

linera-chain/src/chain.rs
```diff
@@ -1506,6 +1509,7 @@ where
         }
         let updated_streams = self.process_emitted_events(block).await?;
         self.process_outgoing_messages(block, tracked).await?;
+        self.vouch_for_block_references(block).await?;
```
deuszx, line 1521: Isn't it the case that we always first `preprocess_block` before we `apply_confirmed_block`?

linera-core/src/client/mod.rs
```diff
@@ -2145,6 +2185,9 @@ impl<Env: Environment> Client<Env> {
             certificates.insert(current_height, certificate);
         }
 
+        self.recertify_if_epoch_revoked(&certificates, remote_node)
```
deuszx, line 2211: Can you add a short comment why we're calling this method here? 

also for line 2052.

linera-core/src/chain_worker/state.rs
```diff
@@ -1021,20 +990,17 @@ where
...
         // cert without advancing the tip.
-        let in_trust_set = self
-            .chain
-            .pre_checkpoint_block_trust
-            .contains(&block_hash)
-            .await?;
+        let in_trust_set = self.chain.vouched_blocks.contains(&block_hash).await?;
         if in_trust_set {
-            self.chain.pre_checkpoint_block_trust.remove(&block_hash)?;
+            self.chain.vouched_blocks.remove(&block_hash)?;
```
deuszx, line 983: So we remove ONLY when we (fully) process that certificate? Is there no other scenario when we can remove the block hash from this set? Can't we also do it when we pre-process a block? We don't accept (pre-process) certificates that are from untrusted epochs

linera-chain/src/chain.rs
afck: Isn't this basically what we do everywhere where we use `multi_get`? IIUC it always returns the values in the same order. (I just noticed that the documentation doesn't state that clearly, which it probably should.)

linera-core/src/chain_worker/state.rs
afck: No, this is for both fully and preprocessed blocks. See `ProcessConfirmedBlockMode` below in this method.

(I guess we could rename it to `process_or_preprocess_confirmed_block`, or `handle_confirmed_block` or something? But I'd prefer to not do this change in this PR, since it's already quite complicated and it didn't introduce that name.)

linera-chain/src/chain.rs
afck: No, that's `execute_contiguous_block` instead of `preprocess_block` then.

linera-core/src/client/mod.rs
afck: Done in https://github.com/linera-io/linera-protocol/pull/6603/changes/555619828bc8c70bc377f970af918048c47cffc1.

linera-chain/src/chain.rs
afck: Done in https://github.com/linera-io/linera-protocol/pull/6603/changes/555619828bc8c70bc377f970af918048c47cffc1.

linera-chain/src/chain.rs
afck: Done in https://github.com/linera-io/linera-protocol/pull/6603/changes/555619828bc8c70bc377f970af918048c47cffc1.

PR #6604, opened 2026-07-13
Keep the received log per sender epoch

afck:
## Motivation

Validators keep every chain's received log forever, including entries that point at certificates from revoked epochs — which clients cannot verify on their own anymore. The log can't simply be truncated: clients track their position in it by index, so removing entries would shift every index behind them.

## Proposal

Partition the received log by the epoch of the sender's certificate, so a revoked epoch's log can be dropped wholesale, and its client-side tracker with it. Validators prune revoked epochs when a client asks for a live epoch's log. Clients keep one tracker per validator and epoch, and request all live epochs in a single query, so a network with many live epochs doesn't pay for one round trip per epoch.

Since a sync then only sees live epochs' entries, an undelivered message from a pruned epoch could otherwise block the messages behind it forever. The client therefore processes the lowest listed block together with its sending ancestors, which re-certifies the pruned-epoch message through the links introduced in the previous PR.

## Test Plan

CI, plus unit tests covering per-epoch log serving and pruning, the batched multi-epoch query, and a regression test for bulk-syncing across a revoked epoch.

## Release Plan

    SDK soon.

## Links

  - Stacked on #6603.

### DIFF
--- a/linera-chain/src/chain.rs
+++ b/linera-chain/src/chain.rs
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use std::{
-    collections::{BTreeMap, BTreeSet, HashMap, HashSet},
+    collections::{BTreeMap, BTreeSet, HashSet},
     sync::Arc,
 };
 
@@ -25,7 +25,7 @@ use linera_execution::{
     ServiceRuntimeEndpoint, TransactionTracker,
 };
 use linera_views::{
-    collection_view::CollectionView,
+    collection_view::CustomCollectionView,
     context::Context,
     log_view::LogView,
     map_view::{CustomMapView, MapView},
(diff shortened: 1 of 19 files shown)


linera-chain/src/chain.rs
```diff
@@ -297,10 +298,18 @@ where
...
+    /// by epoch so that a revoked epoch's log can be dropped wholesale: clients
+    /// cannot verify those certificates on their own anymore, and blocks that still
+    /// matter are re-certified via `previous_block_hash` /
+    /// `previous_message_blocks` links instead.
+    pub received_log: CustomCollectionView<C, Epoch, LogView<C, ChainAndHeight>>,
+    /// The number of `received_log` entries we have synchronized, per validator and
+    /// per sender epoch. Node-local synchronization bookkeeping, not exposed via
+    /// GraphQL.
+    #[cfg_attr(with_graphql, graphql(skip))]
+    pub received_certificate_trackers: MapView<C, (ValidatorPublicKey, Epoch), u64>,
```
ma2bd: This is unbounded in size :-|

ma2bd: unless we clean it for revoked epochs?

ma2bd: we do!

ma2bd: Could this be `MapView<C, ValidatorPublicKey, Vec<(Epoch, u64)>>` to spare DB keys?

afck: Done in https://github.com/linera-io/linera-protocol/pull/6604/commits/d1e968056878b1ada1e64ff04b2620d7333c29f5.

But note that this field is only nonempty on the client side. Not sure how important it is to save DB keys there?

PR #6605, opened 2026-07-13
Require timeout certificates to leave all non-Fast rounds

afck:
## Motivation

Multi-leader rounds can be skipped by any valid proposal in a later round. Supporting that requires tracking the highest signed proposal (`signed_proposal`), a recovery path that advances the round on rejected proposals, and special round-comparison rules — logic that complicates the protocol.

## Proposal

Now every non-`Fast` round ends only via a timeout certificate for that round (or a locking block from a higher round, which proves a quorum already reached it). Proposals must match the current round exactly. `Fast` proposals remain exempt: the existing vote and locking-block checks prevent them from conflicting with later-round blocks.

- Add `TimeoutConfig::multi_leader_round_duration` (default 1 second): multi-leader rounds are meant to be cooperative, so contention should be rare, and a short round keeps retry latency low. Single-leader rounds keep `base_timeout` plus `timeout_increment`.
- The fast round times out iff `fast_round_duration` is configured, regardless of whether there are regular owners. A chain with only super owners and no fast-round duration relies on them for progress.
- Remove the `signed_proposal` register, its recovery path, and `requested_signed_proposal` from `ChainManagerInfo`; `Worker::handle_block_proposal` returns `Result` again since a rejected proposal can no longer produce notifications.
- Expose the new duration as `--multi-leader-round-ms` in the CLI and `multiLeaderRoundMs` in GraphQL, and fix the `base_timeout` docs.

## Test Plan

Tests were updated and extended.

## Release Plan

## Links


afck:
Closing this, since it would probably make some use cases for open multi-leader rounds impossible.

PR #6607, opened 2026-07-14
[testnet] Backport #6606: TokenAmount and configurable-precision fungible

ma2bd:
## Motivation

Backport of #6606 (`TokenAmount<T>` + `Amount` alias + configurable-precision fungible token)
to `testnet_conway`.

## Proposal

Cherry-picks the change onto `testnet_conway`, reconciled with this branch's divergences:

- `linera-base` — `TokenAmount<T>`, the `Amount` type alias, and all conversions. The generic
  `TryFrom<U256>` uses this branch's `ArithmeticError` (not `AmountConversionError`, which
  `testnet_conway` had already removed).
- `linera-sdk` — the `branded_token!` macro, the shared `FungibleTokenAbi` moved to `U128`
  amounts + a `decimals` parameter (default 18), and the WIT type rename `amount` →
  `token-amount` (a rename only; the flat layout is unchanged, so it stays ABI-compatible).
- Examples on the shared `FungibleTokenAbi` (fungible, amm, matching-engine, crowd-funding,
  rfq) adapt at the `U128` boundary.

**Scope note:** `testnet_conway` splits the ABI into `FungibleTokenAbi` and a separate
`NativeFungibleTokenAbi` (still on `Amount`) that `main` never had. The `native-fungible`
example uses that separate ABI, so it is **left on `Amount`** here — only its shared
`InitialState`/`FungibleResponse`/`Parameters` touch points are converted. The generalized
`From<U128>`/`.into()` cleanup from the tail of #6606 is not included and can follow separately.

## Test Plan

CI. Regenerated the WIT files, the `linera-service` GraphQL schema, and the fungible /
native-fungible format snapshots for this branch; `CLI.md` and `binary_formats` are unaffected.

## Release Plan

## Links

- Backport of #6606

### DIFF
--- a/examples/amm/src/contract.rs
+++ b/examples/amm/src/contract.rs
@@ -675,7 +675,7 @@ impl AmmContract {
         let token = self.fungible_id(token_idx);
         let operation = FungibleOperation::Transfer {
             owner: source_owner,
-            amount,
+            amount: amount.into(),
             target_account,
         };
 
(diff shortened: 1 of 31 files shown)


ma2bd:
closing for now

PR #6616, opened 2026-07-15
Reimplement `ChainClient::sync_validator` on top of ValidatorUpdater

afck:
## Motivation

`ChainClient::sync_validator` is a from-scratch reimplementation of "bring a validator up to date with a chain".

## Proposal

Replace its body with a call to `ValidatorUpdater::send_chain_information`, which also gains it upload batching, the lite-certificate fast path, checkpoint push, admin-chain healing, round-evidence sync, and response verification.

The verification requires the validator's public key: `linera validator sync` now resolves it from the current committee by network address, or takes it via a new optional `--public-key` argument for validators that are not (yet) committee members.

## Test Plan

CI

## Release Plan

## Links

### DIFF
--- a/linera-core/src/client/chain_client/mod.rs
+++ b/linera-core/src/client/chain_client/mod.rs
@@ -75,7 +75,7 @@ use crate::{
         ValidatorNodeProvider as _,
     },
     remote_node::RemoteNode,
-    updater::{communicate_with_quorum, CommunicateAction, CommunicationError},
+    updater::{communicate_with_quorum, CommunicateAction, CommunicationError, ValidatorUpdater},
     worker::{Notification, Reason, WorkerError},
 };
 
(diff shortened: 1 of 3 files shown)


linera-core/src/client/chain_client/mod.rs
```diff
@@ -3462,67 +3462,31 @@ impl<Env: Environment> ChainClient<Env> {
...
-                .collect();
-            remote_node.upload_blobs(missing_blobs).await?;
-            remote_node
-                .handle_confirmed_certificate(certificate, CrossChainMessageDelivery::NonBlocking)
-                .await?;
-        }
-
-        Ok(())
+        let mut updater = ValidatorUpdater {
+            remote_node: RemoteNode { public_key, node },
```
deuszx, line 3478: Why is `ValidatorUpdater` not working w/o a public key? It's just about sending data to the remote node.

afck: I think we're still signing all chain info responses, and then checking the signature. But I agree it's not really necessary. :thinking:

deuszx: it seems that `ChainInfoResponse::new` takes the `key_pair` as `Option` type. So it shouldn't be strictly necessary.

deuszx: If at construction it's optional, it should also be optional during reception:
1. It's optional in `client_context.rs`
2. It's required for `RemoteNode::check_and_return_info`

afck: Done. (Unfortunately that made the change a lot bigger; touches lots of places in the code. :confused:)

linera-core/src/remote_node.rs
```diff
@@ -26,14 +26,61 @@ use crate::{
...
+            NodeIdentity::Unknown => f.write_str("(unknown)"),
+        }
+    }
+}
+
+/// A validator node together with its identity.
 #[derive(Clone, Debug)]
 pub struct RemoteNode<N> {
-    pub public_key: ValidatorPublicKey,
+    pub identity: NodeIdentity,
```
deuszx: Why not just `Option<ValidatorPublicKey>`?

linera-service/src/cli/validator.rs
```diff
@@ -236,6 +236,11 @@ pub struct Sync {
     /// Verify validator is online before syncing
     #[arg(long)]
     check_online: bool,
+    /// Public key of the validator, used to verify its responses. Defaults to the key
+    /// registered for this network address in the current committee; if the node is
+    /// not a committee member and no key is given, responses are not verified.
+    #[arg(long)]
+    public_key: Option<ValidatorPublicKey>,
```
deuszx, line 243: Do we need this? Why can't the `Sync` work w/o knowing the public key?

linera-core/src/client/chain_client/mod.rs
```diff
@@ -1026,14 +1026,19 @@ impl<Env: Environment> ChainClient<Env> {
             |_| (),
             |remote_node| {
                 let client = &self.client;
-                let tracker = trackers.get(&remote_node.public_key).copied().unwrap_or(0);
+                let trackers = &trackers;
                 let received_log_batches = Arc::clone(&received_log_batches);
                 Box::pin(async move {
+                    // `communicate_with_quorum` only queries nodes with a known public key.
```
deuszx: Shouldn't `communicate_with_quorum` simply filter out `nodes` that we don't have the public key for?

linera-core/src/client/chain_client/mod.rs
deuszx: I agree it looks net worse but I also think it's overengineered (unless I'm missing something). If not then you can revert that change and keep it non-optional.

linera-core/src/remote_node.rs
afck: It makes it more explicit, and harder to accidentally add code where the public key is dropped. Not sure; I can use `Option` if you prefer.

linera-service/src/cli/validator.rs
afck: If we prepare a soon-to-be-validator, it _does_ make sense to check that everything works correctly and e.g. that they're using the expected public key, doesn't it?

linera-core/src/client/chain_client/mod.rs
afck: I'd say it's an error anyway to call `communicate_with_quorum` with such nodes in the first place. So ideally it would be unreachable. Not sure how I'd change the signature of `communicate_with_quorum` to enforce that statically, though.

linera-core/src/client/chain_client/mod.rs
afck: Taking a step back, I'd say:
* `sync_validator` duplicated functionality of `ValidatorUpdater`, but without using the public key.
* So far, `sync_validator` never used the key, and `ValidatorUpdater` always did (because it was always dealing with _validators_ from a known committee).

This PR is trying to deduplicate this, by using the `ValidatorUpdater` for `sync_validator`, too. So we have to do one of the following now:
* _Require_ that the public key is specified on the command line if it's not known from the current committee.
* Change all the `ValidatorUpdater` code to deal with _optional_ public keys. That's a sizable change, and I don't think the overengineering with the `NodeIdentity` type is the only reason for that.

linera-core/src/client/chain_client/mod.rs
deuszx: Some types of data can be verified w/o the quorum (like a blob/cert requested by hash). That way we'd make sure that even if someone misuses `communicate_with_quorum` it will still work

linera-core/src/client/chain_client/mod.rs
deuszx: I agree that the diff looks too big and complicated now. We can revert that

afck: OK, reverted.

I created https://github.com/linera-io/linera-protocol/issues/6618 for making it optional.

deuszx (approved):

linera-core/src/client/chain_client/mod.rs
afck: But for those use cases, shouldn't one always use `communicate_concurrently` instead?

PR #6617, opened 2026-07-15
Unify the missing-dependency retry ladders in the client

afck:
## Motivation

Four call sites hand-roll the same loop — run a local operation, catch `BlobsNotFound`/`EventsNotFound`/`BlocksNotFound`, download the missing item, retry with a per-call dedup set — with slightly different download strategies: `handle_certificate_with_retry`, `try_process_locking_block_from`, `stage_block_execution and query_application`.

## Proposal

Replace them (and `EventSetDownloader`, which covered only the events case) with a single `DependencyDownloader`, parameterized by a `BlobSource` that captures the three blob download strategies: raw blobs from given nodes, pending blobs from one node, or certificates from the current validators.

Intentional behavior changes:
- All four sites now handle `BlocksNotFound` (missing pre-checkpoint sender blocks); previously only `handle_certificate_with_retry` did.
- `stage_block_execution` now downloads each missing blob at most once instead of retrying `BlobsNotFound` without a dedup set, so a blob that remains missing after a download fails instead of looping.

## Test Plan

CI

## Release Plan

## Links

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -1080,8 +1080,8 @@ impl<Env: Environment> Client<Env> {
         Ok(info)
     }
 
-    /// Calls `handle_confirmed_certificate`, retrying with any missing blobs (downloaded
-    /// from `nodes`) and any missing events (downloaded from the publisher
+    /// Calls `handle_confirmed_certificate`, retrying with any missing blobs and blocks
+    /// (downloaded from `nodes`) and any missing events (downloaded from the publisher
     /// chains via the current validators).
     async fn handle_certificate_with_retry(
         &self,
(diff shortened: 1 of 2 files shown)


linera-core/src/client/mod.rs
```diff
@@ -2509,39 +2470,127 @@ fn filter_new<T: Clone + Eq + std::hash::Hash>(
...
-/// `EventsNotFound` and continue the loop only if it returns `true`.
-pub(crate) struct EventSetDownloader<'a, Env: Environment> {
+/// How missing blobs are obtained when retrying an operation via a
+/// [`DependencyDownloader`].
+pub(crate) enum BlobSource<'a, Env: Environment> {
+    /// Download and process the certificates that registered the blobs, from the
+    /// current validators.
+    Certificates,
+    /// Download the raw blobs from the given nodes and store them.
+    Storage(&'a [RemoteNode<Env::ValidatorNode>]),
```
deuszx, line 2483: The comment on the enum says it's about _how they're obtained_ but this is about what to do with them after we download it?

afck: Hmm… it's both. Not sure how best to describe it:
* The first case is: we need the blobs _and_ proof for that they have been published (i.e. the certificates).
* The second: We already have the proof in our storage, so we only need the blobs now.
* The third: These blobs belong to a pending block: we need to ask the validator for a _pending_ blob explicitly, and then also tell our local node to add it to the pending proposal.

Let's see if Claude can come up with a better formulation…

afck: Done in https://github.com/linera-io/linera-protocol/pull/6617/changes/64b492e5b5cbd56cef3653f1301b7065be77a38f.

deuszx (commented):
TBH I am not sure this is an improvement. The new enum encodes at least two dimensions (source and way of processing) in a single-dimension (enum variant): it's trying to cover this by using variant's name for one thing and fields for other but even that is not consistent. Will leave a decision to others.

Issue #4095, opened 2025-06-11
Revisit and optimize the ubounded queues in the block exporter


usagi32:
And discuss whether we want to use a queue vs a map like in the original design document after evaluating pros and cons of both the approaches

PR #6624, opened 2026-07-25
linera-core: extract the confirmed-certificate sender from ValidatorUpdater

ndr-ds:
## Motivation

The plan is for the chain workers in the server binary to push each block they execute straight to
the other validators: at that moment the certificate and its blobs are already in memory, so
nothing has to be re-read from storage or relayed through another process. Those workers need the
client's existing dependency-upload logic — uploading blobs the destination is missing, and pushing
the admin chain when the destination does not yet know the committee that signed the certificate.

That logic currently lives inside `ValidatorUpdater`, which owns an `Arc<Client>`. A chain worker
has no client, no wallet and no signer, and cannot get one. So the choice is to duplicate the
recovery logic in the server or to extract it; this extracts it.

## Proposal

Extract the confirmed-certificate synchronization path out of `ValidatorUpdater` into a reusable
primitive, `ConfirmedCertificateSender<S, N>` (new `linera-core/src/updater/confirmed_sender.rs`),
holding a `RemoteNode`, a `LocalNodeClient`, the admin chain id and the upload batch size — no
client, wallet, or signer.

Moved, and preserved verbatim: `send_confirmed_certificate` (the dependency-upload retry loop),
the height sync, `read_certificates_for_heights`, `initialize_new_chain_on_validator`,
`update_admin_chain`, `send_chain_info_for_blobs`, and the consensus-round sync. Blob and
certificate reads that went through the local node become direct storage reads.

`ValidatorUpdater::send_chain_information` is now a straight delegation to the primitive, which is
why `updater.rs` gets ~124 lines shorter.

**Chain state views must not be loaded outside the chain worker** (`scripts/check_chain_loads.sh`:
loading one elsewhere races with the worker and can corrupt data). The two chain-level reads this
path needs — the admin chain's local tip, and the block-hash fallback — therefore go through
`LocalNodeClient`, i.e. worker-mediated, exactly as `ValidatorUpdater` did before. The server's
chain workers hold one too, so both consumers use the same mechanism.

Two behaviours are opt-in rather than unconditional, because they are correct for the client and
wrong for a chain worker:

- `with_height_index_backfill()` — writes `(chain_id, height) -> hash` indices back for
  certificates found only through the block-hash-list fallback. Only for a consumer that owns the
  storage it reads.
- `with_consensus_round_sync()` — after height sync, pushes the evidence that justifies a later
  consensus round (locking certificate, timeout certificate, or proposal). That is consensus
  participation rather than replication. The client enables both, preserving its behaviour exactly.

## Test Plan

No behaviour change is intended for the client: it enables both flags, so every path it takes is
the one it took before. The refactor is covered by the existing `linera-core` suite.

```
cargo clippy --locked --all-targets --all-features --workspace --exclude linera-web
cargo clippy --locked -p linera-core --no-default-features
RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --all-features
cargo test -p linera-core --lib      # 158 passed, 0 failed
cargo +nightly fmt -- --check
bash scripts/check_chain_loads.sh
cargo machete
```

One fix worth calling out, found while reviewing this: the `read_certificates_for_heights`
block-hash fallback was initially dropped. Storage written before the `(chain_id, height) -> hash`
index existed can only be read through that fallback, so a deep catch-up against older storage
would have silently skipped legacy certificates. It is restored, with the index write-back behind
the flag above.

## Release Plan

## Links

- #6639 builds on this: in-worker export of executed blocks
- #6641 relays that export through the proxy, so shards need no internet access
- #4095 — the unbounded exporter partition this line of work removes

### DIFF
--- a/linera-core/src/remote_node.rs
+++ b/linera-core/src/remote_node.rs
@@ -29,7 +29,9 @@ use crate::{
 /// A validator node together with the validator's name.
 #[derive(Clone, Debug)]
 pub struct RemoteNode<N> {
+    /// The validator's public key.
     pub public_key: ValidatorPublicKey,
+    /// The client used to communicate with the validator node.
     #[debug(skip)]
     pub node: N,
 }
@@ -197,6 +199,8 @@ impl<N: ValidatorNode> RemoteNode<N> {
         Ok(())
     }
 
+    /// Downloads a blob from the validator, returning `None` if the validator does not have it
+    /// or returns an invalid blob.
     #[instrument(level = "trace")]
     pub async fn download_blob(&self, blob_id: BlobId) -> Result<Option<Blob>, NodeError> {
         match self.node.download_blob(blob_id).await {
--- a/linera-core/src/updater.rs
+++ b/linera-core/src/updater.rs
@@ -12,7 +12,7 @@ use std::{
 
 use futures::{future, Future, StreamExt};
 use linera_base::{
-    crypto::ValidatorPublicKey,
+    crypto::{CryptoHash, ValidatorPublicKey},
     data_types::{BlockHeight, Round, TimeDelta},
     ensure,
     identifiers::{BlobId, ChainId},
@@ -35,11 +35,50 @@ use crate::{
     environment::Environment,
     node::{CrossChainMessageDelivery, NodeError, ValidatorNode},
     remote_node::RemoteNode,
-    updater::confirmed_sender::ConfirmedCertificateSender,
+    updater::confirmed_sender::{ConfirmedCertificateSender, LocalChainState},
     LocalNodeError,
 };
 
 pub(crate) mod confirmed_sender;
+
+/// Answers chain-level queries through the local worker, which owns the chain state views.
+struct WorkerChainState<Env: Environment> {
+    client: Arc<Client<Env>>,
+}
+
+impl<Env: Environment> Clone for WorkerChainState<Env> {
+    fn clone(&self) -> Self {
+        WorkerChainState {
+            client: self.client.clone(),
+        }
+    }
+}
+
+impl<Env: Environment> LocalChainState for WorkerChainState<Env> {
+    async fn next_block_height(
+        &self,
+        chain_id: ChainId,
+    ) -> Result<BlockHeight, chain_client::Error> {
+        Ok(self
+            .client
+            .local_node
+            .chain_info(chain_id)
+            .await?
+            .next_block_height)
+    }
+
+    async fn block_hashes(
+        &self,
+        chain_id: ChainId,
+        heights: Vec<BlockHeight>,
+    ) -> Result<Vec<CryptoHash>, chain_client::Error> {
+        Ok(self
+            .client
+            .local_node
+            .get_block_hashes(chain_id, heights)
+            .await?)
+    }
+}
 
 /// The default amount of time we wait for additional validators to contribute
 /// to the result, as a fraction of how long it took to reach a quorum.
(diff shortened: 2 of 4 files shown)


linera-core/src/updater.rs
```diff
@@ -15,29 +15,30 @@ use linera_base::{
...
     client::{chain_client, Client},
     data_types::{ChainInfo, ChainInfoQuery},
     environment::Environment,
     node::{CrossChainMessageDelivery, NodeError, ValidatorNode},
     remote_node::RemoteNode,
-    LocalNodeError,
+    updater::confirmed_sender::ConfirmedCertificateSender,
 };
 
+pub(crate) mod confirmed_sender;
```
ma2bd, line 40: nit: maybe `updater.rs` should become `updater/mod.rs` then?

linera-core/src/remote_node.rs
```diff
@@ -29,7 +29,9 @@ use crate::{
 /// A validator node together with the validator's name.
 #[derive(Clone, Debug)]
 pub struct RemoteNode<N> {
+    /// The validator's public key.
```
ma2bd, line 32: Damn it looks like I never added `#![deny(missing_docs)]` to `linera-core`?

linera-core/src/updater.rs
```diff
@@ -785,11 +639,12 @@ where
                         .await?;
 
                     tracing::debug!("Sending chains for missing blobs");
-                    self.send_chain_info_for_blobs(
-                        &missing_blob_ids,
-                        CrossChainMessageDelivery::NonBlocking,
-                    )
-                    .await?;
+                    self.confirmed_certificate_sender()
```
ma2bd, line 642: Is there a reason to re-create `confirmed_certificate_sender` every time?

linera-core/src/updater/confirmed_sender.rs
```diff
@@ -0,0 +1,645 @@
+// Copyright (c) Zefchain Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! A lightweight, reusable primitive for pushing confirmed-block certificates to a single
+//! validator.
+//!
+//! [`ConfirmedCertificateSender`] extracts the confirmed-certificate synchronization path out of
```
ma2bd, line 7: nit: `Uploader` feels like a more accurate concept than `Sender` here

linera-core/src/updater.rs
ma2bd: unless the context is different actually

linera-core/src/updater/confirmed_sender.rs
ma2bd: ok but I can see this was in the existing code

linera-core/src/updater/confirmed_sender.rs
```diff
@@ -0,0 +1,645 @@
...
+    /// `[validator_next_height, target_next_block_height)`, in order. Only heights actually present
+    /// in storage are sent; missing heights are silently skipped. This is what makes the "leave gaps on the validator
+    /// side" behavior (#4181) work: a chain we merely *receive* from is stored only at its
+    /// message-bearing heights, so exactly those are pushed. The validator executes the contiguous
+    /// prefix and preprocesses any block that sits above a gap.
+    ///
+    /// For an existing chain (`target_next_block_height > 0`) it optimistically sends the last
+    /// certificate first, then any earlier locally-held certificates in order. For a new chain
+    /// (`target_next_block_height == 0`) it sends the chain description and dependencies first, then
+    /// queries the validator's state.
```
ma2bd, line 152: nit: IMHO this is too verbose. We generally only want the first sentence.

linera-core/src/updater/confirmed_sender.rs
ma2bd: It is a little weird to shorten `ConfirmedCertificateSender` into `confirmed_sender` by the way

linera-core/src/updater.rs
ma2bd, line 87: It's a bit sad that we need a client here in the first place. Perhaps we should try to fix this and use only a local node instead.

ma2bd (changes requested):
I suggest a more targeted approach in #6645

PR #6626, opened 2026-07-27
Filter event-stream publishers by the message policy

ma2bd:
## Motivation

A client that tracks a chain with many recorded event subscriptions (e.g. a user
wallet whose default chain accumulated subscriptions in old blocks) opens one
notification stream **and** one `previous_event_blocks` sync per publisher chain.
Against validators fronted by a load balancer that caps HTTP/2 at 100 concurrent
streams per connection (e.g. a GCP Application LB / GFE), this fan-out saturates the
single connection the client keeps per validator, and unrelated requests — block
proposals, chain-info queries, subscription handshakes — start timing out client-side.

The message-policy options `--process-events-from-application-ids` and
`--restrict-chain-ids-to` existed, but only filtered which events were *processed*
(`collect_stream_updates`); the set of publishers *followed and synced* ignored them,
so they could not reduce the fan-out.

## Proposal

Make the whitelist prune the follow/sync set, not just processing:

- Add `MessagePolicy::accepts_event_stream(chain_id, stream_id)` as the single source
  of truth for both filters (`restrict_chain_ids_to` + `process_events_from_application_ids`).
- Apply it in `event_stream_publishers` (notification subscriptions) and
  `synchronize_publisher_chains` (the sparse `previous_event_blocks` sync).
- Refactor `collect_stream_updates` (event processing) onto the same helper.

The admin chain stays followed and synced unconditionally, so epoch/committee
migration is unaffected. With no whitelist set, `accepts_event_stream` returns `true`
for everything, so behavior is unchanged. Also corrects the now-inaccurate docs on
`restrict_chain_ids_to` and `process_events_from_application_ids` (CLI help and the
`MessagePolicy` struct), which described them as message-only.

With this change, `--process-events-from-application-ids ""` makes a service follow
only the admin chain.

## Test Plan

- Extended the event-streams test to assert `event_stream_publishers` returns
  admin-only when the whitelist is empty, and both senders plus admin when the app is
  whitelisted.
- `cargo test -p linera-core --features wasmer -- event` -> 5 passed (incl.
  `test_synchronize_downloads_admin_chain_events` and
  `test_memory_read_event_downloads_publisher_chain`, which cover the sync/admin paths).
- Nightly `cargo fmt -- --check` and `cargo clippy --all-targets --features wasmer` clean.

## Release Plan

- Client-side only; no protocol, storage, or validator change. Follows the usual
  `testnet_conway` release cycle (SDK). A port to `main` will follow.

## Links

### DIFF
--- a/linera-client/src/util.rs
+++ b/linera-client/src/util.rs
@@ -53,13 +53,18 @@ pub fn parse_chain_set(s: &str) -> Result<HashSet<ChainId>, CryptoError> {
 
 /// Parses a comma-separated list of application IDs into a set.
 pub fn parse_app_set(s: &str) -> anyhow::Result<HashSet<GenericApplicationId>> {
-    s.trim()
-        .split(",")
-        .map(|app_str| {
-            GenericApplicationId::from_str(app_str)
-                .or_else(|_| Ok(ApplicationId::from_str(app_str)?.into()))
-        })
-        .collect()
+    match s.trim() {
+        // An empty set is meaningful (e.g. `--process-events-from-application-ids ""` to follow
+        // only the admin chain), so accept it here instead of failing to parse an empty id.
+        "" => Ok(HashSet::new()),
+        s => s
+            .split(",")
+            .map(|app_str| {
+                GenericApplicationId::from_str(app_str)
+                    .or_else(|_| Ok(ApplicationId::from_str(app_str)?.into()))
+            })
+            .collect(),
+    }
 }
 
 /// Returns after the specified time or if we receive a notification that a new round has started.
@@ -91,3 +96,15 @@ macro_rules! impl_from_infallible {
 }
 
 pub(crate) use impl_from_infallible;
+
+#[cfg(test)]
+mod tests {
+    use super::parse_app_set;
+
+    #[test]
+    fn parse_app_set_accepts_empty() {
+        assert!(parse_app_set("").unwrap().is_empty());
+        assert!(parse_app_set("   ").unwrap().is_empty());
+        assert!(parse_app_set("not-an-application-id").is_err());
+    }
+}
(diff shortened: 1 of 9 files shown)


ndr-ds (commented):
Should we reuse some of the code in `event_stream_publishers` with `update_publisher_chain_modes`? Or is that one not supposed to use the `accepts_event_stream` filter?

linera-client/src/util.rs
```diff
@@ -91,3 +96,15 @@ macro_rules! impl_from_infallible {
...
+mod tests {
+    use super::parse_app_set;
+
+    #[test]
+    fn parse_app_set_accepts_empty() {
+        assert!(parse_app_set("").unwrap().is_empty());
+        assert!(parse_app_set("   ").unwrap().is_empty());
+        assert!(parse_app_set("not-an-application-id").is_err());
+    }
+}
```
ndr-ds: Should this get moved to `src/unit_tests/`?

ndr-ds (approved):
Accepting to unblock, but got a couple questions. Also you need to regenerate `CLI.md` to make CI green

linera-client/src/util.rs
ma2bd: Claude says: Done — moved to `linera-client/src/unit_tests/util.rs` in ed0bee0. Also regenerated `CLI.md` (3f751ef) to green up CI.

ma2bd:
Claude says: Intentional — `update_publisher_chain_modes` is on `Client`, which holds no `message_policy` (each `ChainClient` carries its own), so it can't consult `accepts_event_stream`. It also only records in-memory `EventsOnly` modes via `extend_chain_mode` — no network I/O, and `EventsOnly` isn't `is_tracked` (that needs `is_full`), so it doesn't cascade into tracking/sync. The actual fan-out is gated by the two methods that reach validators — `event_stream_publishers` and `synchronize_publisher_chains` — both now filtered. Documented this in ed0bee0. I kept the publisher-enumeration duplicated rather than sharing it, since the two differ (policy vs no-policy, `ChainClient` vs `Client`); happy to extract the plain enumeration if you'd prefer.

ndr-ds (approved):

PR #6639, opened 2026-07-31
linera-core: export executed blocks to the committee from the chain workers, relayed through the proxy

ndr-ds:
## Motivation

The committee destination of the block exporter is the only validator-to-validator block
dissemination we have: consensus alone guarantees only that a *quorum* holds any given block, so
without it a validator is not a complete replica. It is disabled on every conway validator today
because the exporter persists its own unbounded `canonical_state` log under a single Scylla
partition (#4095).

Rather than fixing that partition, this removes the reason for it. The chain workers already hold
the certificate and its blobs in memory at the moment they execute a block, so the server can push
them to the other validators directly — no notification hop, no re-read from storage, no second
process, and no exporter state to grow.

## Proposal

**One export queue task per server process** (`linera-core/src/chain_worker/export.rs`), modeled
on `forward_cross_chain_queries`. Chain workers hand it each block they execute and otherwise own
nothing. Two invariants drove the design, both the result of review findings:

1. **Nothing on the export path may touch a chain worker.** A task that does resets the worker's
   keep-alive clock, so exporting workers never expire and worker memory grows without bound
   (`--chain-worker-ttl-ms` becomes a no-op). Everything here reads storage only; the export-side
   sender (`BlockSender`) is deliberately not the client's `RemoteNodeUpdater`, which holds a
   local node and is now client-only again, back to its #6645 shape.
2. **Destination state is shared, not per chain.** One connection, one backoff, and one AIMD
   window per destination, shared by all chains — with per-chain tasks, one struggling peer is
   independently discovered and hammered by every chain.

**The queue.** Bounded (`--block-export-queue-size`); a full queue drops the block rather than
blocking the worker. Dropping is safe because it is repaired: every `export` call announces the
chain's tip to a shared map *before* trying the queue, so even a dropped block raises the target
that catch-up measures destinations against. Per-chain height ordering is kept explicit: at most
one in-flight send per (chain, destination) pair.

**Rate limiting (AIMD).** Each destination has an in-flight window: +1 per success up to
`--block-export-max-in-flight` (default 8), halved per transport failure, floor 1. N validators
backfilling one newcomer converge on its real capacity the way N TCP flows share a bottleneck —
no protocol change, works against unmodified conway peers.

**Failure scoping.** Three scopes, because the three failures have nothing to do with each
other. A timeout or transport error is about the *destination*: it halves that destination's
window, backs it off exponentially, and re-resolves its node (which rotates to the next proxy on a
relayed transport). `EventsNotFound`/`BlobsNotFound` are about one *chain* on that destination and
back off only that pair — the admin chain is exported
like any other chain, and a newcomer's missing admin history must not throttle the very
admin-chain export that repairs it. There is no admin-chain replay inside another chain's push.
And a failure to read our *own* storage is neither: it leaves the peer's window untouched and
halves a queue-wide in-flight budget (`--block-export-max-in-flight-total`, default 64, restored
one slot per success). Classifying it per-pair throttles one pair at a time while every other pair
keeps reading at full rate — the control loop cannot see the bottleneck it is creating, which is
the shape that turns storage pressure into a cascade.

**Catch-up.** A tick fires every `--block-export-idle-interval-ms` *even under load* — off a
deadline carried across iterations, since a timer rebuilt per iteration never expires under
sustained traffic. It folds announced tips in, gives every destination a cursor on every tracked
chain whenever the destination set changed (so a validator admitted after a chain's last block is
still caught up on it), and periodically scans storage for committees no block has announced.
Rounds are bounded by `--block-export-max-catch-up-blocks` so a freshly joined validator's
backfill cannot hold up live blocks. Converged chains are forgotten after
`--block-export-converged-retention-ms`, so queue memory tracks recent and lagging chains, not
every chain the process ever served.

Each destination carries the set of chains it is behind on, updated as pairs fall behind and
converge. Membership *is* the "needs work" flag, so there is no second bit to fall out of step
with it, and a drain costs what there is to send rather than what the process is holding — the
tick previously rediscovered that set by scanning every (chain, destination) pair at 5 Hz, which
at a million lagging chains took **longer than the tick interval itself**, stopping export during
exactly the outage it exists to survive. Drains resume from a rotating cursor, because a set is
ordered and walking it from the start would hand the window to the same lowest chain ids forever.

**Scale.** Per-chain state is keyed by a dense destination index rather than by
`ValidatorPublicKey`, which is 88 bytes and whose `Ord` re-encodes the curve point on *every
comparison* — 36 ns against about 1 ns for an integer ([`elliptic-curve`'s own
`TODO(tarcieri): more efficient implementation?`](https://docs.rs/elliptic-curve/0.13.8/src/elliptic_curve/public_key.rs.html#356-368)).
Per-chain state is the one place that cost is multiplied by the number of chains a down
destination leaves behind. Measured against the real key type:

| tracked chains | committee | memory/chain | convergence sweep |
| --- | --- | --- | --- |
| 10k | 3 | 1654 B → **246 B** | 4.0 ms → **0.1 ms** |
| 100k | 3 | 1643 B → **235 B** | 89.8 ms → **10.2 ms** |
| 1M | 3 | 1669 B → **261 B** (1.67 GB → 261 MB) | 1800 ms → **68.5 ms** |
| 100k | 20 | 4939 B → **1323 B** | 2842 ms → **72.7 ms** |

The sweep also bounds how many chains it forgets per pass, since each removal takes the mutex
every chain worker uses on every block, and returns the map's capacity once a burst has drained.

Two full passes remain O(tracked chains × destinations) and are deliberately left that way, with
the numbers above as the justification: the sweep (68.5 ms at a million chains, once every 5 s,
against a 200 ms tick) and the refill that gives a newly admitted validator a cursor on every
chain (585 ms at a million chains, once per committee change). Both are ~10 ms at the scale a
shard actually reaches, and bounding them needs iteration resumable across ticks — state that can
desync from the sets it mirrors, which is the bug class this design removed.

**Cursors.** `ChainStateView::exported_heights` (appended last, so positional `RootView` tags
keep existing databases readable) persists how far each destination acknowledged, folded in by
the worker on its normal save — merge-max, so a late fold cannot regress it. Going the other way,
`seed_missing_cursors` fills any cursor the queue does not have from it, on every block and
whatever created the record; a stale seed costs one skipped re-offer, since the cursor follows
the destination's own reported height in both directions. Blocks a destination already holds are
never re-sent (`reset_and_reexecute_chain` re-offers a chain's whole history).

**Committees.** Exported blocks carry only their epoch; the queue loads committees from storage
(`get_or_load_committee`) and probes the admin chain's epoch event on a cadence, so a validator
admitted while every chain is idle still becomes a destination. Reading the full set of *active*
epochs (and the union of their validators) from storage is a natural follow-up.

**Shards never touch the internet.** A shard holds the validator's secret key, so instead of
dialing peers it sends each request to this validator's own proxy — on the same internal endpoint
shards already use for notifications — naming the intended validator, and the proxy performs it:

- a `ValidatorRelay` service on the proxy's existing internal listener (no new port, no infra
  change), limited to the four requests export makes, forwarding bytes as they arrived so the
  shard sees exactly the peer's answers, errors included;
- the proxy refuses to relay to any address not in a committee it can see (checked live: without
  the guard it happily dialed an arbitrary host), scans epochs outside its lock, and reports scan
  failures as `unavailable` rather than disguising them as `permission_denied`;
- unilateral and conway-compatible: the peer sees an ordinary inbound gRPC call on its public
  port. `--block-export-transport direct` bypasses the relay for operators who need to take the
  sending proxy out of the path (trade-off documented on the flag).

**Scheduling reads the storage clock**, not the wall clock: every cursor, backoff deadline and
sweep timestamp is a `Timestamp` from `Storage::clock()`, and the tick sleeps through it. Under a
`TestClock` the queue can therefore be driven in virtual time — the idle-catch-up tests now
advance the clock instead of sleeping through real milliseconds. The one exception is the
queue-latency histogram, which stays on the wall clock because a simulated clock would report a
delay no operator waited.

**Enablement.** Off by default, behind `--export-blocks-to-committee`. Metrics: queue depth and
blob bytes, dropped blocks, export/send latency, per-destination lag and AIMD window, tracked
chains (the work-list a down peer grows), chain-scoped backoffs (a destination that cannot accept
some chain yet), acknowledged sends per destination (attempts live in the latency histogram;
success rate needs its own numerator), the number of resolved destinations (zero with export
enabled is otherwise indistinguishable from healthy-and-idle), lagging chain-destination *pairs*
(what the memory actually tracks — a chain behind on ten destinations costs ten times one behind
on one), the queue-wide in-flight budget, and the proxy's relayed-request count. Per-validator series are removed when a
validator leaves, so a departed peer does not leave a gauge frozen at its last value.

**Review.** This went through fourteen adversarial rounds — @ma2bd's inline review and gist, plus
multi-agent passes carrying correctness *and* performance lenses. 61 confirmed defects fixed, 15
refuted by evidence rather than argument. The scheduler is largely a product of that loop:

| Round | Fixed | Shape |
| --- | --- | --- |
| ma2bd inline + gist | 16 + 14 | Design defects: the chain-worker TTL leak, the frozen foreign admin worker, a tick that could not fire under load |
| 2 | 8 | Three were regressions from round 1's own fixes |
| 3 | 3 | Two were regressions from round 2's fixes |
| 4 | 5 | Dominated by a regression in the fix meant to make the class impossible |
| 5 | 2 | Same seeding bug, third creation path |
| 6 | 0 confirmed | One split, adjudicated against the code: real mechanism, wrong consequence |
| 7 | 0 | Clean on correctness |
| 8 | 3 | Performance lenses added. Two findings measured and *declined* as below the threshold; one fix was a regression in round 7's own fairness fix, invisible because the test that named the mechanism was a copy of it |
| 9 | 0 | Scoped to round 8's fix commit |
| 10 | 6 | Full-breadth rerun of 9, plus distributed-semantics and test-adequacy lenses that had never run: a missing `shrink_to_fit` on the progress map (the sibling of one fixed eleven lines earlier), a wasted per-block lookup, a comment claiming drop coverage a counter proved false (0 drops), and two net-up tests whose every assertion a dead export path satisfied |
| 11 | 3 | All three were regressions in round 10's own fixes: the new shrink ran *under the mutex* whose hold `MAX_FORGET_PER_SWEEP` bounds (10–26 ms measured by three independent benchmarks); both strengthened tests still passed with the path they name broken, because attempts and totals are not successes. Fixed with a per-destination acknowledged-sends counter both tests now assert on |
| 12 | 3 | The round-11 shrink still freed the peak-sized table under the mutex (`forget_chains` now hands it back for freeing outside the lock, pinned by a unit test that fails against both historical shapes); the dead-proxy baseline raced the kill (now snapshotted after it); a peer oscillating its reported height could hold a pair in a zero-delay re-send loop (a regression now backs the pair off — a real restore pays one delay once) |
| 13 | 1 | Aborted on a session limit (9 of 10 agents died); its one surviving lens found that round 12's own regression backoff never escalated against an *alternating* peer, because the advance half reset the shared counter. Fixed with a regression counter an advance cannot clear. Re-running at full breadth |
| 13 (full) | 2 | Six of seven lenses independently caught that round 13's own fix commit had *silently deleted* the shrink budget added one commit earlier — a hunk its message never mentioned, and one the existing test could not distinguish. Also: a destination reporting a height above our tip was acknowledged at that height and the claim persisted, permanently ending export to that pair |
| 14 | 2 | Round 13's clamp turned out to be half a fix: the acknowledgement was clamped, the in-memory cursor was not, so an over-report still stranded the pair for the process lifetime (no "behind" predicate matches, and only a completed send rewrites a cursor). Clamped once at entry instead. The test was also satisfied by "always acknowledge tip - 1", so it gained the honest-report direction |
| 15 | 0 | First clean full-breadth round: all seven lenses ran and found nothing in the diff |
| post-approval | — | @ma2bd's review comments plus the actionable items from #6715: storage-pressure backpressure, the two gauges that issue asks for, and the storage clock |

Two findings were **refuted with evidence**: tokio's `Child::kill()` on a reaped child returns
`Ok`, so `kill_proxy` never broke `terminate()` (the vec-removal is hygiene); and
`make_nodes_from_list(iter::once(..))` is deliberate, because the test provider resolves by
public key, which only the list API carries — `make_node` would panic every export test.

The persisted-cursor seeding regressed three times, each time because it was tied to record
*creation* while a second path also creates records. It is now `ChainRecord::seed_missing_cursors`,
applied on every block, idempotent and path-independent — pinned by a test with a positive
control, which is what the previous two fixes lacked.

## Test Plan

Unit tests (in-process four-validator `TestBuilder`, all positive-controlled — each named test
fails when its mechanism is disabled):

- `test_chain_workers_expire_while_export_is_enabled` — a worker with export on expires at its
  TTL; failed with `1 chain workers still resident 10s after the last activity` before the fix.
- `test_blocks_are_exported_to_the_committee` / `test_blocks_are_not_exported_by_default`.
- `test_export_catches_a_lagging_validator_up_while_idle` and
  `test_export_catches_up_a_backlog_larger_than_the_bound` (queue_size 2, the smallest the config
  accepts; instrumenting the drop path showed the sequential burst never actually drops, so
  drop-and-repair rests on the announce-before-enqueue ordering, not on a test) — both fail with
  the idle requeue disabled.
- `test_catch_up_sends_at_most_the_bound_per_round` — the bound arithmetic directly; a
  convergence-only test passes with the bound ignored, this one fails (`left: 11, right: 3`).
- `test_send_block_skips_a_block_the_destination_already_has` — asserted against an *offline*
  destination, so skipping succeeds precisely because nothing was sent.
- `a_height_above_our_tip_is_acknowledged_only_up_to_it` — a peer claiming `u64::MAX` is
  acknowledged at our tip *and stored at our tip*; the second half matters because a cursor above
  the tip satisfies every "converged" predicate and no "behind" one, and only a completed send
  rewrites a cursor. It also pins the other direction — an honest report below the tip is taken at
  its word — since "always acknowledge tip - 1" would otherwise satisfy the clamp.
- `forgetting_a_drained_burst_returns_its_table` — exercises the state where the two halves of
  the shrink guard *disagree* (a table several times oversized while still holding more than one
  sweep's budget). Its previous shape returned `None` under both, which is how the budget went
  missing for a commit.
- `an_oscillating_peer_escalates_but_a_restored_one_does_not` — the regression penalty doubles
  per lie (`[100ms, 200ms, 400ms, 800ms]`) while a restore pays one delay and has it cleared by
  its next advance; against the previous shape the same sequence measures
  `[100ms, 100ms, 100ms, 100ms]`. Paired with a `size_of::<ChainDest>()` assertion, since the
  counter has to stay free at one per (chain, destination).
- `draining_rotates_through_the_backlog` and `a_round_that_visits_nothing_keeps_its_place` — the
  drain's fairness, both against the real `drain_candidates`/`advance_cursor`. The second pins the
  case that matters: a saturated round must not reset the rotation. An earlier version of the
  first tested a *copy* of the production ordering and so passed against a live starvation bug.
- `test_export_survives_an_unreachable_destination`, config validation per knob, backoff
  escalation.

Integration tests over a real network (`block_export_tests.rs`, wired into the
`storage-service-tests` CI job; each takes `INTEGRATION_TEST_GUARD`):

- `test_block_export_through_the_proxy` — proxies' relayed-request counters are the load-bearing
  assertion; blocks alone prove nothing since the client talks to every validator anyway.
- `test_export_catches_up_a_newly_added_validator` — includes a chain left **idle since before
  the admission**, so only the queue's own bookkeeping can deliver it; fails
  (`idle chain 0/3`) with the newcomer fill-in disabled.
- `test_export_survives_a_dead_proxy` — a killed proxy's traffic moves to the rotation's next.
  The load-bearing assertion is **every destination's acknowledged-send counter growing after the
  kill**, with blocks produced inside the polling loop: chain convergence is produced by the
  client broadcast whether export works or not, and the surviving proxy's total also grows with a
  destination stranded, since the others round-robin onto it. A frozen per-destination counter
  *is* the stranding.
- `test_relay_refuses_a_non_committee_destination` — the SSRF guard, asserted directly.
- `test_block_export_direct_bypasses_the_proxy` — the exact mirror of the through-proxy test:
  relay counters at zero, and every destination's acknowledged-send counter nonzero. Successes,
  not attempts: the latency histogram counts failed sends too, so a transport that exported
  nothing — or failed every send — would otherwise pass.

```
cargo clippy --locked --all-targets --all-features --workspace --exclude linera-web
cargo clippy --locked --no-default-features
cargo clippy --locked --target wasm32-unknown-unknown --no-default-features --features web-default …
RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --all-features
cargo test -p linera-core --lib          # 183 passed, 0 failed
cargo test -p linera-service --test block_export_tests --features storage-service,metrics -- --ignored
cargo +nightly fmt -- --check
diff <(cargo run --locked --bin linera-schema-export) linera-service-graphql-client/gql/service_schema.graphql
bash scripts/check_chain_loads.sh
cargo machete
```

**Accepted limitations**, each documented at the code that owns it rather than left implicit:

- Catch-up covers chains seen since process start; a chain that never produces another block is
  not re-scanned from storage.
- The `exported_heights` fold is throttled, so a burst's tail may not be persisted — the cost is
  one query per destination when the cursor is rebuilt, against a whole-register write per block.
- A destination that regresses (restored from a backup) while we believe it converged waits for
  the chain's next block; probing converged pairs would cost a query per chain per destination
  forever, to catch something only a restore causes. A regressed report is believed but backs the
  pair off, so an oscillating (Byzantine) peer throttles itself instead of holding a re-send loop
  open. The penalty escalates on a regression counter an advance cannot clear: alternating
  advance with regression would otherwise reset the delay to its base value on every second
  answer — roughly two full catch-up rounds a second, indefinitely, every one of them counted as
  a success. A genuine restore regresses once and pays one delay.
- A pair deferred for a missing committee after a restart waits for the admin chain's next block
  or for client traffic, which pushes the admin chain on the same error. It retries indefinitely,
  and `block_export_chain_scoped_backoffs` makes it visible.
- Per-chain state is O(committee), not O(lagging pairs): every tracked chain carries a cursor for
  every destination, ~8.3 KB per chain at a committee of 100. Making it sparse would mean "absent
  = converged", which collides with today's "absent = ask the destination".
- `ProgressMap`'s removals and its shrink are both budgeted; its *growth* is not, so the map's
  own doubling rehash holds the progress mutex for ~10 ms at a million entries. That is inherent
  to the hash map and amortized over the inserts that caused it.
- A destination that advances one block per reply keeps its backoff cleared, so it can hold a
  pair re-reading catch-up windows. Bounded per round by `--block-export-max-catch-up-blocks` and
  across chains by the AIMD window, but not by the destination's rate of progress.
- A down destination retains the work-list of every chain active during its outage — that set is
  what its catch-up needs; `block_export_tracked_chains` exposes it.
- The tick's storage reads (the committee scan, every ~2 s) run on the queue task, so in-flight
  sends are not polled for the duration of one read — milliseconds against send timeouts measured
  in seconds. Driving the scan concurrently would need completions applied to shared state from
  two places; not worth it at that ratio.
- The `exported_heights` fold filters through the *chain's own* committee, while destinations come
  from the process-latest one. A newcomer the chain's epoch does not know yet is exported to, but
  its cursor is not persisted until the chain adopts the newer epoch — after a restart that pair
  pays one height query to rebuild it. Persisting it would mean sharing the live destination set
  across the progress mutex to keep pruning of departed validators exact.

Not covered: real-network scale. Nothing here has carried conway traffic; confirming there
(values-only deploy via `shardArgs`, no chart or infra change) is the step before the old
exporter code is deleted.

## Release Plan

## Links

- #4095 — the unbounded exporter partition this line of work removes
- #3668 — committee destination
- #6645 — the updater refactor this builds on (and returns `RemoteNodeUpdater` to)

### DIFF
--- a/linera-core/src/chain_worker/export.rs
+++ b/linera-core/src/chain_worker/export.rs
@@ -675,9 +675,16 @@ impl<N> DestState<N> {
     }
 
     /// Moves the cursor past the last chain considered, so the next round advances.
+    ///
+    /// A round that considered nothing — a saturated destination breaks before looking at its
+    /// first candidate — leaves the cursor alone. Clearing it there restarts the next drain at
+    /// the lowest chain id, which is the starvation the cursor exists to prevent, and a busy
+    /// destination is saturated on almost every tick.
     fn advance_cursor(&mut self, last_considered: Option<ChainId>) {
-        self.lagging_cursor =
-            last_considered.and_then(|id| self.lagging.range(id..).nth(1).copied());
+        let Some(last) = last_considered else {
+            return;
+        };
+        self.lagging_cursor = self.lagging.range(last..).nth(1).copied();
     }
 }
 
@@ -1499,7 +1506,7 @@ where
         jobs: &mut FuturesUnordered<JobFuture>,
         now: Instant,
     ) {
-        if dest.retry_at.is_some_and(|at| at > now) {
+        if dest.retry_at.is_some_and(|at| at > now) || dest.in_flight >= dest.window {
             return;
         }
         // Only as far as the window allows, so the cost is the sends we are about to make and
@@ -1511,6 +1518,7 @@ where
         // a backlog of ineligible entries cannot turn this back into a full scan.
         let ordered = dest.drain_candidates(dest.window.saturating_mul(LAGGING_SCAN_FACTOR));
         let mut spawn = Vec::new();
+        let mut stale = Vec::new();
         let mut last_visited = None;
         for chain_id in ordered {
             if dest.in_flight + spawn.len() >= dest.window {
@@ -1518,6 +1526,9 @@ where
             }
             last_visited = Some(chain_id);
             let Some(record) = chains.get(&chain_id) else {
+                // The chain was forgotten under us; drop the entry rather than walk past it on
+                // every drain from here on.
+                stale.push(chain_id);
                 continue;
             };
             let Some(chain_dest) = record.dest(index) else {
@@ -1532,6 +1543,9 @@ where
             }
         }
         dest.advance_cursor(last_visited);
+        for chain_id in stale {
+            dest.lagging.remove(&chain_id);
+        }
         for chain_id in spawn {
             let Some(record) = chains.get_mut(&chain_id) else {
                 continue;
@@ -1867,45 +1881,66 @@ mod tests {
 
     /// The drain rotates through the backlog instead of always serving its lowest chain ids.
     ///
-    /// The maintained set replaced a FIFO, and a set is *ordered* — iterating it from the start
+    /// The maintained set replaced a FIFO, and a set is *ordered* — walking it from the start
     /// every tick hands the whole window to the same few chains and starves everything after
     /// them, which at a large backlog means those chains are never exported at all.
     #[test]
     fn draining_rotates_through_the_backlog() {
         let mut dest = test_dest_state();
-        dest.window = 2;
-        let ids = (0..6u8)
-            .map(|i| ChainId(CryptoHash::test_hash(format!("chain{i}"))))
-            .collect::<Vec<_>>();
-        dest.lagging = ids.iter().copied().collect();
+        dest.lagging = test_chain_ids(6).into_iter().collect();
 
-        // Two passes with the window's worth of work each; the second must move on.
-        let first = drained(&dest, 2);
-        dest.lagging_cursor = first
-            .last()
-            .and_then(|id| dest.lagging.range(*id..).nth(1).copied());
-        let second = drained(&dest, 2);
+        // Two rounds of a two-chain window; the second must move past the first.
+        let first = dest.drain_candidates(2);
+        dest.advance_cursor(first.last().copied());
+        let second = dest.drain_candidates(2);
 
         assert_eq!(first.len(), 2);
         assert_eq!(second.len(), 2);
         assert!(
             first.iter().all(|id| !second.contains(id)),
-            "the second pass repeated the first: {first:?} then {second:?}",
+            "the second round repeated the first: {first:?} then {second:?}",
         );
     }
 
-    /// The candidate ordering `drain_ready` walks, without the spawning around it.
-    fn drained(dest: &DestState<()>, budget: usize) -> Vec<ChainId> {
-        match dest.lagging_cursor {
-            Some(cursor) => dest
-                .lagging
-                .range(cursor..)
-                .chain(dest.lagging.iter().take_while(|id| **id < cursor))
-                .copied()
-                .take(budget)
-                .collect(),
-            None => dest.lagging.iter().copied().take(budget).collect(),
-        }
+    /// A round that considered nothing must not throw the rotation away.
+    ///
+    /// `drain_ready` returns before looking at a single candidate when the destination is
+    /// already at its window, which for a destination with a backlog is almost every tick. If
+    /// that round reset the cursor, the next one would restart at the lowest chain id and the
+    /// tail of the backlog would never be reached — the starvation the cursor exists to prevent,
+    /// reintroduced by the cursor's own bookkeeping.
+    #[test]
+    fn a_round_that_visits_nothing_keeps_its_place() {
+        let mut dest = test_dest_state();
+        let ids = test_chain_ids(6);
+        dest.lagging = ids.iter().copied().collect();
+
+        let first = dest.drain_candidates(2);
+        dest.advance_cursor(first.last().copied());
+        let parked = dest.lagging_cursor;
+        assert!(parked.is_some(), "the first round must park the cursor");
+
+        // The saturated round: no candidate is visited, so nothing was considered.
+        dest.advance_cursor(None);
+
+        assert_eq!(
+            dest.lagging_cursor, parked,
+            "a round that visited nothing moved the cursor",
+        );
+        let resumed = dest.drain_candidates(2);
+        assert!(
+            resumed.iter().all(|id| !first.contains(id)),
+            "the drain restarted at the front of the backlog: {first:?} then {resumed:?}",
+        );
+    }
+
+    /// Chain ids in the order `lagging` holds them, so a test can name "the front" of a backlog.
+    fn test_chain_ids(count: usize) -> Vec<ChainId> {
+        let mut ids = (0..count)
+            .map(|i| ChainId(CryptoHash::test_hash(format!("chain{i}"))))
+            .collect::<Vec<_>>();
+        ids.sort_unstable();
+        ids
     }
 
     /// Destinations indexed the way `sync_destinations` assigns them: in registration order.
--- a/linera-core/src/chain_worker/export.rs
+++ b/linera-core/src/chain_worker/export.rs
@@ -40,7 +40,7 @@
 //! full history of a chain that never produces blocks again is out of scope here.
 
 use std::{
-    collections::{BTreeMap, HashMap, VecDeque},
+    collections::{BTreeMap, BTreeSet, HashMap},
     iter,
     sync::{Arc, Mutex},
 };
@@ -309,10 +309,26 @@ impl Clone for BlockExportHandle {
     }
 }
 
+/// A destination's dense index, assigned once per public key and never reused.
+///
+/// All per-chain state is keyed by this rather than by `ValidatorPublicKey`, which is 88 bytes
+/// and whose `Ord` re-encodes the curve point on *every comparison* — measured at 36 ns, against
+/// about 1 ns for an integer. Per-chain state is the one place that cost is multiplied by the
+/// number of chains a down destination leaves behind.
+type DestIndex = u32;
+
 /// The highest height each destination has acknowledged, per chain, written by the queue task
 /// and folded into each chain's `exported_heights` by its worker on save. Entries are removed
 /// when a chain converges, so this holds lagging chains only.
-type SharedProgress = Arc<Mutex<HashMap<ChainId, BTreeMap<ValidatorPublicKey, BlockHeight>>>>;
+#[derive(Default)]
+struct ProgressMap {
+    /// What the indices below refer to. Append-only: an index keeps its meaning for the life of
+    /// the process, so a validator that leaves and rejoins cannot inherit another's cursors.
+    validators: Vec<ValidatorPublicKey>,
+    heights: HashMap<ChainId, Vec<(DestIndex, BlockHeight)>>,
+}
+
+type SharedProgress = Arc<Mutex<ProgressMap>>;
 
 /// The height after each chain's newest announced block. Written on every `export` call before
 /// the queue is tried, so a block the full queue drops still raises the repair target the tick
@@ -407,13 +423,18 @@ impl BlockExportHandle {
             .progress
             .lock()
             .expect("progress mutex is never poisoned");
-        let Some(chain_progress) = progress.get(&chain_id) else {
+        let Some(chain_progress) = progress.heights.get(&chain_id) else {
             return BTreeMap::new();
         };
         chain_progress
             .iter()
-            .filter(|(validator, _)| committee.validators().contains_key(*validator))
-            .map(|(validator, height)| (*validator, *height))
+            .filter_map(|(index, height)| {
+                let validator = progress.validators.get(*index as usize)?;
+                committee
+                    .validators()
+                    .contains_key(validator)
+                    .then_some((*validator, *height))
+            })
             .collect()
     }
 }
@@ -454,8 +475,10 @@ where
         committee_dirty: false,
         admin_chain_id: None,
         ticks_until_scan: 0,
+        ticks_until_sweep: TICKS_PER_CONVERGENCE_SWEEP,
         chains: HashMap::new(),
-        destinations: HashMap::new(),
+        destinations: BTreeMap::new(),
+        dest_indices: BTreeMap::new(),
         next_generation: 0,
         announced_epoch: None,
         scan_attempted_for: None,
@@ -485,7 +508,10 @@ struct ChainRecord {
     tip: BlockHeight,
     /// When this chain last saw a block or a completed send, for the convergence sweep.
     last_activity: Instant,
-    dests: BTreeMap<ValidatorPublicKey, ChainDest>,
+    /// Sorted by index, so lookups binary-search integers. A flat vector rather than a map
+    /// because this is allocated per tracked chain: a `BTreeMap` pays for an eleven-slot leaf
+    /// whatever the committee size.
+    dests: Vec<(DestIndex, ChainDest)>,
 }
 
 impl ChainRecord {
@@ -497,24 +523,25 @@ impl ChainRecord {
     /// path used to miss the seeding, and nothing in the suite could see it.
     fn new<N>(
         now: Instant,
-        destinations: &HashMap<ValidatorPublicKey, DestState<N>>,
+        destinations: &BTreeMap<DestIndex, DestState<N>>,
         exported_heights: &BTreeMap<ValidatorPublicKey, BlockHeight>,
     ) -> Self {
         ChainRecord {
             tip: BlockHeight::ZERO,
             last_activity: now,
+            // Already in index order, which is the order `dests` must keep.
             dests: destinations
-                .keys()
-                .map(|validator| {
+                .iter()
+                .map(|(index, dest)| {
                     // Seeded here rather than by a later fill: a cursor pre-populated as `None`
                     // and then only `or_insert`-ed would silently swallow the persisted height,
                     // leaving `exported_heights` write-only and costing a query per destination
                     // on every restart.
                     let next_height = exported_heights
-                        .get(validator)
+                        .get(&dest.validator)
                         .and_then(|height| height.try_add_one().ok());
                     (
-                        *validator,
+                        *index,
                         ChainDest {
                             next_height,
                             ..ChainDest::default()
@@ -523,6 +550,34 @@ impl ChainRecord {
                 })
                 .collect(),
         }
+    }
+
+    fn dest(&self, index: DestIndex) -> Option<&ChainDest> {
+        let at = self
+            .dests
+            .binary_search_by_key(&index, |(at, _)| *at)
+            .ok()?;
+        Some(&self.dests[at].1)
+    }
+
+    fn dest_mut(&mut self, index: DestIndex) -> Option<&mut ChainDest> {
+        let at = self
+            .dests
+            .binary_search_by_key(&index, |(at, _)| *at)
+            .ok()?;
+        Some(&mut self.dests[at].1)
+    }
+
+    /// The cursor for `index`, created empty if this record does not have one yet.
+    fn dest_entry(&mut self, index: DestIndex) -> &mut ChainDest {
+        let at = match self.dests.binary_search_by_key(&index, |(at, _)| *at) {
+            Ok(at) => at,
+            Err(at) => {
+                self.dests.insert(at, (index, ChainDest::default()));
+                at
+            }
+        };
+        &mut self.dests[at].1
     }
 }
 
@@ -539,14 +594,14 @@ impl ChainRecord {
     /// the cursor in either direction.
     fn seed_missing_cursors<N>(
         &mut self,
-        destinations: &HashMap<ValidatorPublicKey, DestState<N>>,
+        destinations: &BTreeMap<DestIndex, DestState<N>>,
         exported_heights: &BTreeMap<ValidatorPublicKey, BlockHeight>,
     ) {
-        for validator in destinations.keys() {
+        for (index, dest) in destinations {
             let seed = exported_heights
-                .get(validator)
+                .get(&dest.validator)
                 .and_then(|height| height.try_add_one().ok());
-            let chain_dest = self.dests.entry(*validator).or_default();
+            let chain_dest = self.dest_entry(*index);
             if chain_dest.next_height.is_none() && chain_dest.in_flight.is_none() {
                 chain_dest.next_height = seed;
             }
@@ -564,8 +619,6 @@ struct ChainDest {
     /// the generation is what stops a stale job's completion from clearing a *newer* job's flag
     /// and breaking the one-send-per-pair ordering invariant.
     in_flight: Option<u64>,
-    /// Whether this chain already sits in the destination's ready list, to keep it there once.
-    queued: bool,
     /// Backoff for *chain-scoped* failures — the destination is healthy but cannot accept this
     /// chain yet, e.g. it lacks the committee and the admin chain's export has not reached it.
     retry_at: Option<Instant>,
@@ -575,6 +628,8 @@ struct ChainDest {
 /// One destination validator: its connection and the health state every chain shares.
 struct DestState<N> {
     node: N,
+    /// Kept here because per-chain state refers to this destination by index, not by key.
+    validator: ValidatorPublicKey,
     address: String,
     /// Bumped whenever this state is rebuilt, so a send started against a previous incarnation
     /// cannot corrupt the new one's accounting when it completes.
@@ -587,8 +642,43 @@ struct DestState<N> {
     /// Backoff for *destination-scoped* failures: transport errors and timeouts.
     retry_at: Option<Instant>,
     failures: u32,
-    /// Chains with work for this destination, drained as the window allows.
-    ready: VecDeque<ChainId>,
+    /// Chains this destination is behind on, maintained as pairs fall behind and converge
+    /// rather than rediscovered by scanning every chain each tick — that scan was O(tracked
+    /// chains x destinations) at 5 Hz, which at a million lagging chains took longer than the
+    /// tick interval itself and starved the queue of everything else.
+    ///
+    /// Membership *is* the "needs work" flag, so there is no separate `queued` bit to fall out
+    /// of step with it.
+    lagging: BTreeSet<ChainId>,
+    /// Where the next drain resumes in `lagging`. Without it the set's ordering hands the window
+    /// to the same lowest chain ids every time and starves the rest of the backlog — the
+    /// fairness the previous FIFO had for free.
+    lagging_cursor: Option<ChainId>,
+}
+
+impl<N> DestState<N> {
+    /// The chains to consider this round, resuming where the last one stopped and wrapping around.
+    ///
+    /// The rotation is the point: `lagging` is ordered, so walking it from the start every tick
+    /// would hand the window to the same lowest chain ids forever and starve the rest.
+    fn drain_candidates(&self, budget: usize) -> Vec<ChainId> {
+        match self.lagging_cursor {
+            Some(cursor) => self
+                .lagging
+                .range(cursor..)
+                .chain(self.lagging.iter().take_while(|id| **id < cursor))
+                .copied()
+                .take(budget)
+                .collect(),
+            None => self.lagging.iter().copied().take(budget).collect(),
+        }
+    }
+
+    /// Moves the cursor past the last chain considered, so the next round advances.
+    fn advance_cursor(&mut self, last_considered: Option<ChainId>) {
+        self.lagging_cursor =
+            last_considered.and_then(|id| self.lagging.range(id..).nth(1).copied());
+    }
 }
 
 /// How one send ended, scoped to what the error tells us about.
@@ -623,8 +713,14 @@ where
     admin_chain_id: Option<ChainId>,
     /// Ticks until the next storage scan for a committee no block has carried yet.
     ticks_until_scan: u32,
+    /// Ticks until the next sweep of converged chains.
+    ticks_until_sweep: u32,
     chains: HashMap<ChainId, ChainRecord>,
-    destinations: HashMap<ValidatorPublicKey, DestState<P::Node>>,
+    destinations: BTreeMap<DestIndex, DestState<P::Node>>,
+    /// Every validator ever registered as a destination, and the index its per-chain state uses.
+    /// Never pruned, so a validator that rejoins reuses its index rather than taking a departed
+    /// one's.
+    dest_indices: BTreeMap<ValidatorPublicKey, DestIndex>,
     /// The last destination generation handed out; never reused within this queue's lifetime.
     next_generation: u64,
     /// The newest epoch any exported block has announced; the tick loads its committee when it
@@ -648,6 +744,18 @@ where
 
 /// How many ticks apart the queue probes storage for committees no block has announced.
 const TICKS_PER_COMMITTEE_SCAN: u32 = 10;
+
+/// How many times the window a single drain may look past before giving up for this tick, so a
+/// backlog of ineligible entries cannot turn the drain back into a full scan.
+const LAGGING_SCAN_FACTOR: usize = 4;
+
+/// How many ticks apart converged chains are swept. The window they are held for is minutes, so
+/// this only decides how promptly the memory comes back.
+const TICKS_PER_CONVERGENCE_SWEEP: u32 = 25;
+
+/// How many chains one sweep may forget, bounding how long it holds the progress mutex that every
+/// chain worker takes on every block.
+const MAX_FORGET_PER_SWEEP: usize = 4096;
 
 impl<S, P> BlockExportQueue<S, P>
 where
@@ -752,13 +860,14 @@ where
         record.tip = record.tip.max(tip);
         record.last_activity = now;
         record.seed_missing_cursors(&self.destinations, &block.exported_heights);
-        let validators = self.destinations.keys().copied().collect::<Vec<_>>();
-        for validator in validators {
+        let indices = self.destinations.keys().copied().collect::<Vec<_>>();
+        for index in indices {
             let record = self.chains.get_mut(&chain_id).expect("inserted above");
-            let chain_dest = record.dests.entry(validator).or_default();
+            let record_tip = record.tip;
+            let chain_dest = record.dest_entry(index);
             let dest = self
                 .destinations
-                .get_mut(&validator)
+                .get_mut(&index)
                 .expect("iterating destinations");
             let contiguous = chain_dest.next_height == Some(height);
             let can_send_now = chain_dest.in_flight.is_none()
@@ -773,21 +882,20 @@ where
                     &self.storage,
                     &self.config,
                     chain_id,
-                    validator,
+                    index,
                     dest,
                     chain_dest,
-                    record.tip,
+                    record_tip,
                     Some((block.certificate.clone(), block.blobs.clone())),
                 );
-            } else if !chain_dest.queued {
-                chain_dest.queued = true;
-                dest.ready.push_back(chain_id);
+            } else if chain_dest.next_height.is_none_or(|next| next < record_tip) {
+                dest.lagging.insert(chain_id);
                 if can_send_now {
                     Self::drain_ready(
                         &mut self.chains,
                         &self.storage,
                         &self.config,
-                        validator,
+                        index,
                         dest,
                         jobs,
                         now,
@@ -800,20 +908,21 @@ where
     /// Folds one finished send back into the destination's and the chain's state.
     fn on_done(
         &mut self,
-        (chain_id, validator, generation, outcome): JobDone,
+        (chain_id, index, generation, outcome): JobDone,
         jobs: &mut FuturesUnordered<JobFuture>,
     ) {
         let now = Instant::now();
-        let Some(dest) = self.destinations.get_mut(&validator) else {
+        let Some(dest) = self.destinations.get_mut(&index) else {
             return; // The validator left the committee while its send was in flight.
         };
+        let validator = dest.validator;
         if dest.generation != generation {
             // The send ran against a previous incarnation of this destination; its slot was
             // never counted here and its result must not touch the fresh state.
             if let Some(chain_dest) = self
                 .chains
                 .get_mut(&chain_id)
-                .and_then(|record| record.dests.get_mut(&validator))
+                .and_then(|record| record.dest_mut(index))
             {
                 // Only the flag this very job set — the pair may since carry a newer job's.
                 if chain_dest.in_flight == Some(generation) {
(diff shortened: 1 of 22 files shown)


deuszx:
Just wanted to flag one thing: this changes the behavior of the exporter from its initial design. The idea was that there's never a roundtrip style of communication between the exporter and validator (send-response-analyse). Exporter figures out the correct order in which the elements need to be pushed and does that. Now there's a request-response style: exporter sends a thing, validator responds what is missing, exporter supplies it (rinse and repeat).

ma2bd:
> Just wanted to flag one thing: this changes the behavior of the exporter from its initial design. The idea was that there's never a roundtrip style of communication between the exporter and validator (send-response-analyse). Exporter figures out the correct order in which the elements need to be pushed and does that. Now there's a request-response style: exporter sends a thing, validator responds what is missing, exporter supplies it (rinse and repeat).

Yes, we're trying a different approach.

linera-core/src/chain_worker/export.rs
```diff
@@ -0,0 +1,718 @@
...
+                // so a destination that cannot be caught up trickles rather than spins.
+                Err(_) => self.catch_up_round().await,
+            }
+        }
+        debug!("Chain worker dropped; stopping block export");
+    }
+
+    /// Brings the destination set in line with the committee as the worker sees it now. Asked of
+    /// the worker rather than read from the chain view, which the worker owns while running.
+    async fn sync_destinations_from_local_node(&mut self) {
```
ma2bd: [Claude:] **The idle catch-up round pins the chain worker forever.**

This queries the task's *own* chain through the local node, so it re-enters `WorkerState::handle_chain_info_query` -> `chain_write` -> `guard.touch()` every `idle_catch_up_interval` (200 ms). `spawn_keep_alive` only drops a worker after `ttl` elapses with no touch, so the worker is touched 5x/s forever - and since the worker owns the `ChainExporter`, the task keeps the worker alive and the worker keeps the task alive. Every chain that has executed even one block then keeps a resident `ChainWorkerState` + `ChainStateView` + task for the process lifetime, at 5 exclusive write-locks/s each.

This also invalidates the reasoning behind `max_retry_delay.min(chain_worker_ttl)` in `server.rs` ("the task would be gone before the destination came up for retry") - it never is.

Separately, I don't think this query can return anything new. It reads the chain's *own* committee, which changes only when that chain executes a block (`CreateCommittee` and `ProcessNewEpoch` both `insert` and `epoch.set` in the same operation), and `export_block` already passes `committee` with every block. So the "a validator admitted while this chain happens to be idle" case can't arise from this chain's own committee - I think the call can be deleted rather than made cheaper.

linera-core/src/updater.rs
```diff
@@ -792,11 +804,128 @@ where
...
+        if to_read.is_empty() {
+            return Ok(blobs);
+        }
+        let read = self.local_node.read_blobs_from_storage(&to_read).await?;
+        blobs.extend(read.ok_or(NodeError::BlobsNotFound(to_read))?);
+        Ok(blobs)
+    }
+
+    pub(crate) async fn update_admin_chain(&mut self) -> Result<(), chain_client::Error> {
         let local_admin_info = self.local_node.chain_info(self.admin_chain_id).await?;
```
ma2bd, line 799: [Claude:] **The admin chain is read through the shard's own local node.**

Shards are sharded by chain id, so on a multi-shard validator every shard that does *not* own the admin chain instantiates its own `ChainWorkerState` for it here. That copy never executes admin blocks, so it never advances: `target = local_admin_info.next_block_height` is frozen at whatever it was when the shard first loaded it, and a destination that needs a newer epoch keeps being replayed a short admin chain and keeps failing.

Normally the staleness would be bounded by `chain_worker_ttl` - but every export round touches that foreign worker, so combined with the TTL issue in `export.rs` it never expires and the staleness is permanent.

Reading the admin chain's tip from storage instead avoids both the foreign worker and the staleness.

linera-core/src/chain_worker/export.rs
```diff
@@ -0,0 +1,718 @@
...
+        let result = self
+            .updater
+            .send_missing_blocks(
+                chain_id,
+                target,
+                self.next_height,
+                config.max_catch_up_blocks,
+            )
+            .await;
+        let outcome = self.record_outcome(result, target, now, config);
```
ma2bd: [Claude:] **The "succeeded but advanced nothing" backoff never escalates.**

`record_outcome`'s `Ok` arm sets `self.failures = 0`, and only then does `catch_up` call `back_off`, which reads `self.failures` for the shift. So the exponent is always `2^0` and `failures` oscillates 0<->1: `max_retry_delay` is unreachable on this path, and a destination that can never be caught up is retried every `retry_delay` (1 s) forever, each round paying a `read_certificates_for_heights` over up to `max_catch_up_blocks`.

There's a second effect: `sync_destinations_with` treats `failures > 0` as stuck, so such a destination has its connection torn down and rebuilt on every round, advancing the proxy rotation each time.

Computing the delay from the pre-`record_outcome` failure count (or incrementing before backing off) fixes both.

linera-service/src/proxy/grpc.rs
```diff
@@ -249,6 +288,76 @@ where
...
+            .contains(address)
+        {
+            return true;
+        }
+        let mut destinations = self.0.relay_destinations.write().await;
+        // Another task may have refreshed while we waited for the lock.
+        if destinations.addresses.contains(address) {
+            return true;
+        }
+        while let Ok(Some(committee)) = self
```
ma2bd: [Claude:] **One unloadable epoch permanently hides every later committee.**

`next_epoch` advances only inside the `Ok(Some(_))` arm, so the scan stops at the first epoch it cannot load and never resumes. `get_or_load_committee` returns `Ok(None)` whenever the `NewCommittee` event (or the network description, or the committee blob) is not in this validator's storage - e.g. a validator whose admin-chain history is incomplete. If epoch *k* is missing while *k+1...* exist, the current committee's addresses are never inserted and every relayed request fails with `permission_denied("... not a member of any known committee")`, with nothing naming the real cause.

The `while let Ok(...)` also collapses `Err` into the same path, so a transient storage error reaches the shard as an authorization refusal - indistinguishable from the genuine rejection the new integration test asserts on.

linera-service/src/server.rs
```diff
@@ -657,6 +810,36 @@ async fn run(options: ServerOptions) {
                 allow_revert_confirm,
                 reset_on_corrupted_chain_state_mins,
                 recovery_whitelist: recovery_whitelist.map(HashSet::from_iter),
+                block_export_config: export_blocks_to_committee.then(|| {
+                    // The export task dies with its chain worker, so a backoff longer than the
+                    // worker TTL would never elapse — the destination would then only be
+                    // reconsidered if the chain produced another block.
+                    let max_retry_delay = block_export_max_retry_delay.min(chain_worker_ttl);
```
ma2bd: [Claude:] `chain_worker_ttl` of zero legitimately means "no TTL" (`util::non_zero_duration` turns it into `ttl: None`), so `--chain-worker-ttl-ms 0` makes `max_retry_delay` zero here. `BlockExportConfig::check()` validates `certificate_upload_batch_size`, `max_catch_up_blocks`, `idle_catch_up_interval` and `retry_delay`, but not `max_retry_delay`, so this passes. `back_off` then computes a zero delay and sets `retry_at = Some(now)`, which `is_due` accepts immediately - a dead validator is retried on every block and every catch-up round with no pause, which is what `check()`'s own comments say must be impossible.

Worth asserting `max_retry_delay > 0` and `retry_delay <= max_retry_delay` in `check()`.

linera-service/src/cli_wrappers/local_net.rs
```diff
@@ -318,6 +328,20 @@ impl Validator {
         self.proxies.push(proxy)
     }
 
+    /// Kills one of this validator's proxies, leaving the rest running.
+    #[cfg(with_testing)]
+    async fn kill_proxy(&mut self, proxy_id: usize) -> Result<()> {
```
ma2bd, line 333: [Claude:] **This leaves the dead child in `self.proxies`, which breaks `terminate()`.**

`kill()` here is `start_kill()` + `wait()`, so the tokio `Child` ends up in `FusedChild::Done` but stays in the vec. `Validator::terminate` then loops over *all* proxies calling `proxy.kill().await.context(...)?`, and `start_kill` on an exited child returns `Err(InvalidInput, "can't kill an exited process")`.

`test_export_survives_a_dead_proxy` ends in `net.terminate().await?`, so it should fail at teardown even when the recovery it tests works - and this suite is now wired into `.github/workflows/rust.yml`. `terminate_server` already handles this by removing the child from the vec; the same fix applies here.

linera-rpc/src/grpc/relay.rs
```diff
@@ -0,0 +1,353 @@
...
+        let channel = self
+            .pool
+            .channel(relay.to_owned())
+            .map_err(|error: GrpcError| NodeError::GrpcError {
+                error: format!("error creating channel to the relay {relay}: {error}"),
+            })?;
+        Ok(RelayClient {
+            destination: address.to_owned(),
+            address: network.http_address(),
+            client: ValidatorRelayClient::new(channel),
```
ma2bd: [Claude:] This is the only gRPC client in the tree built without the message-size limits - `client.rs`, `server.rs` (x4) and `proxy/grpc.rs` (x4) all set `.max_encoding_message_size(GRPC_MAX_MESSAGE_SIZE).max_decoding_message_size(GRPC_MAX_MESSAGE_SIZE)`. So the shard side of the relay decodes at tonic's 4 MiB default while both hops around it accept 16 MiB; a relayed `ChainInfoResult` in between fails with a decode error that the export cursor logic reads as a generic failure.

linera-service/src/proxy/grpc.rs
```diff
@@ -249,6 +288,76 @@ where
...
+            .0
+            .relay_destinations
+            .read()
+            .await
+            .addresses
+            .contains(address)
+        {
+            return true;
+        }
+        let mut destinations = self.0.relay_destinations.write().await;
```
ma2bd: [Claude:] The write guard is held across the whole storage-backed epoch scan below. tokio's `RwLock` blocks new readers once a writer is queued, so every concurrent `peer()` call from every shard - including ones whose destination is already cached - stalls behind it. And because a miss leaves `next_epoch` on an epoch that returns `None`, each subsequent request for that same unknown address repeats the `read_network_description()` + `read_event()` round trip under the lock.

Running the scan outside the lock and taking the write guard only to merge the result would avoid it.

linera-core/src/chain_worker/export.rs
```diff
@@ -0,0 +1,718 @@
...
+        let stuck = self
+            .destinations
+            .iter()
+            .filter(|(_, destination)| destination.failures > 0)
+            .map(|(validator, destination)| (*validator, destination.address.clone()))
+            .collect::<Vec<_>>();
+        if stuck.is_empty() {
+            return;
+        }
+        let nodes = match self.node_provider.make_nodes_from_list(stuck) {
```
ma2bd: [Claude:] `make_nodes_from_list` collects into `Result<Vec<_>>` and fails the whole batch on the first bad address, so one unresolvable stuck destination blocks reconnecting all the others. `sync_destinations_with` resolves per-validator specifically to avoid that; this path still has it.

linera-core/src/chain_worker/export.rs
```diff
@@ -0,0 +1,718 @@
...
+
+    /// Pushes one bounded chunk of missing blocks to every destination that is behind.
+    ///
+    /// Runs only when nothing is queued, so live blocks always take priority over backfill.
+    async fn catch_up_round(&mut self) {
+        // Pick up committee changes too. A validator admitted while this chain happens to be idle
+        // would otherwise never become a destination at all — `sync_destinations` runs off an
+        // exported block, and there is no next block to carry the new committee.
+        self.sync_destinations_from_local_node().await;
+        self.refresh_stuck_destinations();
```
ma2bd: [Claude:] `sync_destinations_from_local_node()` on the line above already reaches `sync_destinations_with`, which rebuilds every destination with `failures > 0` - so this rebuilds the same set a second time. With `RelayNodeProvider` each rebuild calls `next_relay()` and bumps the shared counter, so a stuck destination advances the round-robin by two per round; with two proxies that lands it back on the one it just failed on, which is the opposite of what the doc intends.

The doc also says it runs "on every new block", but `export()` only calls `sync_destinations`.

ma2bd (commented):
[Claude:] Review of `ndr-ds/in-worker-block-export` @ 8f4da97 against merge-base 29e35a7. 16 inline comments; this is the summary.

**Two I'd want resolved before merge:**

1. **The idle catch-up round defeats the chain-worker TTL** (`chain_worker/export.rs`, `sync_destinations_from_local_node`). The task queries its own chain through the local node, touching the worker 5x/s, so no exporting chain worker is ever evicted and the task/worker pair keep each other alive. I also don't think the query can ever return a new committee, so it may be deletable outright.
2. **The admin chain is read through the shard's own local node** (`updater.rs`, `update_admin_chain`). On a multi-shard validator the non-owning shards materialise a foreign copy that never advances, so the catch-up target is frozen - and (1) makes that permanent rather than TTL-bounded.

**Medium:** the no-progress backoff never escalates; the proxy's epoch scan stops at the first unloadable epoch (and reports storage errors as `permission_denied`); `max_retry_delay` can be zero; `kill_proxy` leaves a dead child so `terminate()` errors and the new CI-wired test fails at teardown; `RelayClient` is the only gRPC client without the message-size limits.

**Low:** eight cleanups - batch node resolution, a duplicated stuck-destination rebuild, blob retention vs. the queue metric, a resend to destinations that are ahead, a dead `sync_consensus_rounds` knob, per-shard `RelayNodeProvider`, and a `LocalNodeClient` built even when export is off.

**Checked and cleared,** so nobody re-litigates them: `next_back()` on the committees map cannot outrun `current_committee()` (both mutation paths `insert` and `epoch.set` together, and `open_chain` seeds the child from the parent's own epoch); restoring `next_height` from `progress` on a stuck rebuild is safe, and `retry_at`/`failures` are preserved across it; `send_missing_blocks` bounds and `record_progress`'s `try_sub_one` are right; no deadlock between `export_block` and the export task; `NodeOptions::default()` really does zero `max_retries`; `exported_heights` is last in `ChainStateView`, so the positional `RootView` tags are preserved.

Generated by Claude Code.

linera-core/src/chain_worker/export.rs
```diff
@@ -0,0 +1,718 @@
...
+/// how to reach another validator.
+pub type ChainExporterFactory<S> = Arc<dyn Fn(ChainExportSetup<S>) -> ChainExporter + Send + Sync>;
+
+/// A block a chain worker has executed, on its way to the other validators.
+struct ExportedBlock {
+    certificate: CacheArc<ConfirmedBlockCertificate>,
+    /// The block's required blobs, so that a destination missing them — which is always the case
+    /// for a blob this very block publishes — is served without a read from storage. Held as the
+    /// storage cache's pointers, so queued blocks share the allocations rather than copying them.
+    blobs: Vec<CacheArc<Blob>>,
```
ma2bd, line 351: [Claude:] These are strong refs into the storage dedup cache for every required blob (published + read), so a backlogged queue keeps whole blob payloads resident past the LRU's own bound. Since `export()` awaits all destinations before dequeuing, one slow destination on a chain publishing large blobs grows shard RSS by bytes - while `block_export_queue_size` counts only blocks, so the metric the unbounded queue's safety rests on under-reports the real pressure.

linera-core/src/chain_worker/export.rs
```diff
@@ -0,0 +1,718 @@
...
+            // connection is replaced, so a validator that is itself down keeps backing off.
+            let previous = self.destinations.remove(&validator);
+            self.destinations.insert(
+                validator,
+                Destination {
+                    updater,
+                    address,
+                    // An acknowledged height was reported by the validator itself, so the block
+                    // after it is the one the validator needs next.
+                    next_height: acknowledged
```
ma2bd: [Claude:] Seeding from `acknowledged` is safe (those heights were reported by the validator itself), but on a chain-worker reload it comes from the persisted `exported_heights`, which is documented as a *lower* bound. If the stored value is 100 and the destination is really at 150, the first push passes a cursor of 101, so `send_missing_blocks` skips the query and re-reads + re-sends certificates 101..150 to a validator that already has them - per chain, per restart.

linera-core/src/updater.rs
```diff
@@ -792,11 +804,128 @@ where
...
+        let next_height = if destination_next_height == Some(height) {
+            height
+        } else {
+            self.send_missing_blocks(chain_id, height, destination_next_height, max_catch_up)
+                .await?
+        };
+
+        // Still below this block: the gap was larger than one chunk, so leave the rest for the
+        // next round rather than sending a block that would only be preprocessed.
+        if next_height < height {
```
ma2bd: [Claude:] The `is_some_and(|next| next > height)` guard above fires only when the caller passed a cursor, but the cursor is `None` on the first push to a destination and after every failure. In that case `send_missing_blocks` queries the destination, finds an empty range and returns the destination's own height - say 200 for a block at 150 - and `next_height < height` is false, so this falls through and sends a certificate the peer already executed.

`if next_height != height { return Ok(next_height); }` covers both directions.

linera-core/src/updater.rs
```diff
@@ -137,23 +136,33 @@ impl CommunicateAction {
...
-    pub remote_node: RemoteNode<Env::ValidatorNode>,
-    pub local_node: LocalNodeClient<Env::Storage>,
+    pub remote_node: RemoteNode<N>,
+    pub local_node: LocalNodeClient<S>,
     pub admin_chain_id: ChainId,
     pub certificate_upload_batch_size: u64,
+    /// Whether to also bring the remote node's *consensus round* up to ours once its height
+    /// matches. The client wants this; export does not — pushing consensus evidence is taking
+    /// part in consensus rather than replicating it, so export only ever moves heights.
+    pub sync_consensus_rounds: bool,
```
ma2bd: [Claude:] As far as I can tell this is always `true` in production: the flag only guards an early return in `send_chain_information`, and the export path never reaches it - `send_block`/`send_missing_blocks` don't call it, and `update_admin_chain` takes the `max_admin_catch_up_blocks.is_some()` branch and returns first. The remaining call sites are all client-side where it's `true`, so `false` is exercised only by the two new unit tests.

linera-service/src/server.rs
```diff
@@ -112,10 +120,74 @@ impl ServerContext {
...
+                    .map(|proxy| proxy.internal_address(&internal_network.protocol))
+                    .collect::<Vec<_>>();
+                assert!(
+                    !relay_addresses.is_empty(),
+                    "relaying exported blocks needs at least one proxy to relay through, but this \
+                     validator's internal network configures none; use \
+                     `--block-export-transport direct` to send without a proxy",
+                );
+                let node_provider =
+                    Arc::new(grpc::RelayNodeProvider::new(relay_addresses, node_options));
```
ma2bd: [Claude:] `chain_exporter_factory` is called once per shard from `make_shard_state`, so each shard builds its own `RelayNodeProvider` with its own `GrpcConnectionPool` and its own `Arc<AtomicUsize>`. That's S pooled connections per peer rather than the one the doc comment promises, and every shard's rotation starts at index 0, so the first destination of each shard is pinned to the same proxy instead of the egress being spread. Building it once in `ServerContext` would match the doc.

linera-core/src/worker.rs
```diff
@@ -1270,6 +1288,8 @@ where
             chain_id,
             service_runtime_endpoint,
             service_runtime_task,
+            self.chain_exporter_factory.clone(),
+            LocalNodeClient::new(self.clone()),
```
ma2bd: [Claude:] This is built for every chain worker even when `chain_exporter_factory` is `None`, so validators running without `--export-blocks-to-committee` pay an `Arc` allocation and a full `WorkerState` clone per chain for a field only block export reads. Constructing it lazily (or only when the factory is set) removes the cost for the default path.

ma2bd:
[Claude:] Following up on my two comments about the chain-worker TTL and the stale admin chain — I think they share one root cause that's worth naming separately, because fixing it collapses several of the other findings at once: **`ChainExportTask` takes a `LocalNodeClient`, and it doesn't really need one.**

Where `local_node` is actually used on the export path, at `8f4da97`:

| Use site | Genuinely needs a worker? |
|---|---|
| `updater.rs:596`, `updater.rs:1118` — `storage_client()` | No — unwraps straight to storage |
| `updater.rs:910` — `read_blobs_from_storage` | No — `local_node.rs:200-206` is a two-line passthrough to `storage.read_blobs` |
| `updater.rs:1006` — `handle_chain_info_query` | Not reachable from export (it's inside `send_chain_information`, which the export path never enters) |
| `export.rs:362` — the chain's own committee | Yes, as written |
| `updater.rs:916` — the admin chain's tip | Yes, as written |

So four of the six are storage in a trench coat, and the dependency really exists for two calls. The justification in the doc comment — *"asked of the worker rather than read from the chain view, which the worker owns while running"* — is a real constraint in general, but I don't think it applies to either one.

**The chain's own committee (`export.rs:362`).** The export task is *spawned by* the worker that owns that view, and `export_block` already hands it `committee` and `admin_chain_id` with every block. So the task takes a full round trip back into `WorkerState` → `get_or_create_chain_worker` → its own parent worker's write lock to fetch what its parent had in hand a moment earlier. That re-entrancy is exactly what pins the worker: `chain_write` touches `last_access`, so the TTL never elapses.

And I don't think the query can return anything new. It reads *this chain's own* committee, which changes only when this chain executes a block — `CreateCommittee` and `ProcessNewEpoch` both `insert` into `committees` and `epoch.set` in the same operation. An idle chain's committee is frozen by construction, so the "a validator admitted while this chain happens to be idle" case can't arise from this source. If that's right, the call can be deleted rather than made cheaper, which removes the TTL leak and the ~5 write-locks/s per idle chain together.

**The admin chain's tip (`updater.rs:916`).** Here the local node isn't just unnecessary, it's the wrong source: that's a chain the shard generally doesn't own, so `chain_info` materialises a second, foreign `ChainWorkerState` that never executes admin blocks and therefore never advances. The race the doc comment worries about is precisely what this creates — and the stale copy is the failure mode, not a hypothetical.

The updater already reaches storage directly in two other places via `self.local_node.storage_client()`, so reading the admin chain's tip the same way would fix this without changing `RemoteNodeUpdater`'s shape at all.

**What that buys.** With those two gone, `ChainExportSetup` no longer needs a `LocalNodeClient` (`export.rs:162`) — its only direct use is the committee query — which in turn means `ChainWorkerState` doesn't need one either, and the unconditional `LocalNodeClient::new(self.clone())` per chain worker in `worker.rs:1292` disappears for everyone, including validators running without `--export-blocks-to-committee`. Three of the comments I left above resolve as a side effect, and the export task ends up depending only on storage, which is the level it actually operates at.

Generated by Claude Code.

linera-core/src/chain_worker/state.rs
```diff
@@ -1015,6 +1021,10 @@ where
         tip: ChainTipState,
         notify_when_messages_are_delivered: Option<oneshot::Sender<()>>,
     ) -> Result<(ChainInfoResponse, NetworkActions, BlockOutcome), WorkerError> {
+        // Hold the certificate behind the storage cache's shared pointer rather than unwrapping
+        // the block out of it: the export task pushes this very certificate, so it never has to
+        // read the block back from storage.
```
ma2bd: nit: just keep the beginning "Hold the certificate behind the storage cache's shared pointer"

linera-core/src/chain_worker/state.rs
```diff
@@ -1142,6 +1156,61 @@ where
...
+        };
+        // The committee *after* applying the block: a validator that this very block admits has to
+        // be exported to from now on.
+        let (epoch, committee) = match self.chain.current_committee().await {
+            Ok((epoch, committee)) => (epoch, committee),
+            Err(error) => {
+                warn!(%error, "Not exporting a block of a chain with no current committee");
+                return;
+            }
+        };
```
ma2bd, line 1193: We now have `storage.get_or_load_committee(epoch)` so the exporter only needs the epoch (for starters -- ideally we could also read the set of active epochs, hence validators from storage).

linera-core/src/chain_worker/export.rs
```diff
@@ -0,0 +1,1418 @@
...
+                    Ok(received) => Wake::Block(received),
+                    Err(_) => Wake::Tick,
+                }
+            } else {
+                // `futures::select_biased` rather than `tokio::select`, which does not compile
+                // for the web target. Completions first, then fresh blocks, then the tick.
+                futures::select_biased! {
+                    done = jobs.next() => Wake::Done(done.expect("jobs is not empty")),
+                    received = receiver.recv().fuse() => Wake::Block(received),
+                    _ = linera_base::time::timer::sleep(self.config.idle_catch_up_interval)
```
ma2bd: needs to use the "storage clock" for simulatabilty

linera-core/src/chain_worker/state.rs
```diff
@@ -1142,6 +1167,76 @@ where
...
+
+        // Through the cache rather than `Arc::new`: these blobs are already in storage's cache
+        // in the common case, and the cache is what keeps one allocation per blob content.
+        let blobs = published_blobs
+            .into_iter()
+            .chain(read_blobs.into_values())
+            .map(|blob| self.storage.cache_blob(blob))
```
ma2bd, line 1200: should this function take a `Vec<Arc<Blob>>` already?

linera-core/src/chain_worker/export.rs
```diff
@@ -0,0 +1,2231 @@
...
+                    Ok(received) => Wake::Block(received),
+                    Err(_) => Wake::Tick,
+                }
+            } else {
+                // `futures::select_biased` rather than `tokio::select`, which does not compile
+                // for the web target. Completions first, then fresh blocks, then the tick.
+                futures::select_biased! {
+                    done = jobs.next() => Wake::Done(done.expect("jobs is not empty")),
+                    received = receiver.recv().fuse() => Wake::Block(received),
+                    _ = linera_base::time::timer::sleep(until_tick).fuse() => Wake::Tick,
```
ma2bd: shouldn't we use the Storage clock for simulatability?

ma2bd: (I can also do it in a follow-up)

ma2bd (approved):
See also https://github.com/linera-io/linera-protocol/issues/6715

PR #6659, opened 2026-08-06
Stop a panicking transport from permanently wedging a cross-chain message queue

ma2bd:
Stacked on #6657 — please review that first; this PR's own diff is the last commit.

## Motivation

If `handle_request` panics, the affected cross-chain message queue stops delivering **for the
lifetime of the process**, silently.

The forwarder tracks one job per queue in `job_states`, and relies on every spawned step
reporting back the queue it belongs to:

```rust
Some(Ok((queue, action))) = steps.join_next() => { ... }
```

A step that ends in a panic reports nothing — `join_next` yields `Some(Err(JoinError))`, the
pattern does not match, and the `select!` arm is skipped. The queue's entry stays in
`job_states` with no running step, so every later request for that `(sender, recipient)` pair
takes the `Entry::Occupied` path, overwrites the entry, and is never spawned. The forwarder
itself keeps running and other queues are unaffected, so nothing surfaces except the missing
messages.

Panics are not reachable through the current gRPC closure, which uses `?` throughout — though it
does index `network.shard(shard_id)` on a `Vec`. This is a latent fault rather than an active
one; the reason to fix it is that the failure mode is invisible and permanent, and this is the
code path behind the worker stalls we have chased in production.

Found while writing the tests in #6657.

## Proposal

Catch the panic at the spawn boundary and turn it into an ordinary failed attempt, so the queue
stays under the existing retry and give-up logic: it is retried up to `max_retries`, then the
job is removed and the queue is released. Panics are logged at `error!` alongside the default
panic hook's output, so they stay visible.

The wrapper shadows `run_action`, so both spawn sites pick it up without further changes. 24
lines of production code; no change to the behaviour of any non-panicking path.

Retrying (rather than dropping the request immediately) keeps a panic and a returned `Err`
handled identically, which seemed easier to reason about than a third code path — happy to
switch to dropping the job if reviewers prefer.

## Test Plan

Two regression tests, on top of the harness added in #6657:

- `test_retries_after_a_panicking_transport` — one panic, then success.
- `test_queue_recovers_after_giving_up_on_a_panicking_request` — panics exhaust the retries,
  and a later request on the same queue is still delivered. This is the wedge itself.

```
cargo test -p linera-rpc                                      # 48 passed
cargo clippy -p linera-rpc --all-targets \
    --features server,test,simple-network,metrics             # clean
cargo +nightly fmt -- --check                                 # clean
```

Verified that both new tests fail with the fix reverted and the tests kept, and that no other
test changes result:

```
test test_queue_recovers_after_giving_up_on_a_panicking_request ... FAILED
test test_retries_after_a_panicking_transport ... FAILED
test result: FAILED. 13 passed; 2 failed
```

## Release Plan

`testnet_conway` has the same defect with a different blast radius: it drives its steps with
`futures::stream::FuturesUnordered` rather than a `JoinSet`, so the panic is not contained at
all — it unwinds the forwarder task and that shard stops forwarding cross-chain messages
entirely. The same fix applies. I will open the backport once this lands.

## Links

- Stacked on #6657.

### DIFF
--- a/linera-rpc/src/cross_chain_message_queue.rs
+++ b/linera-rpc/src/cross_chain_message_queue.rs
@@ -124,7 +124,12 @@ pub(crate) async fn forward_cross_chain_queries<F, G>(
     // later request for it is only recorded there and never sent — that queue stops
     // delivering for the lifetime of the process. Catching the panic turns it into an
     // ordinary failed attempt, leaving the queue under the usual retry and give-up logic.
-    let run_action = move |action, queue: QueueId, state| {
+    let run_action = move |action, queue: QueueId, state: JobState| {
+        // Kept out of `state`, which the step consumes, so that the panic can be reported
+        // with the same fields as an ordinary send failure.
+        let nickname = state.nickname.clone();
+        let to_shard = state.task.shard_id;
+        let retries = state.retries;
         let step = run_action.clone()(action, queue, state);
         async move {
             AssertUnwindSafe(step)
@@ -132,7 +137,10 @@ pub(crate) async fn forward_cross_chain_queries<F, G>(
                 .await
                 .unwrap_or_else(|_| {
                     error!(
+                        nickname,
+                        retry = retries,
                         from_shard = this_shard,
+                        to_shard,
                         sender = %queue.sender,
                         recipient = %queue.recipient,
                         "Panic while sending a cross-chain query; treating it as a failed attempt",
--- a/linera-rpc/src/cross_chain_message_queue.rs
+++ b/linera-rpc/src/cross_chain_message_queue.rs
@@ -262,3 +262,477 @@ impl JobState {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use std::sync::{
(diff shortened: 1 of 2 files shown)


linera-rpc/src/cross_chain_message_queue.rs
```diff
@@ -117,6 +118,30 @@ pub(crate) async fn forward_cross_chain_queries<F, G>(
...
+    // delivering for the lifetime of the process. Catching the panic turns it into an
+    // ordinary failed attempt, leaving the queue under the usual retry and give-up logic.
+    let run_action = move |action, queue: QueueId, state| {
+        let step = run_action.clone()(action, queue, state);
+        async move {
+            AssertUnwindSafe(step)
+                .catch_unwind()
+                .await
+                .unwrap_or_else(|_| {
+                    error!(
```
ndr-ds, line 139: Should we also print `to_shard` and `nickname` here, like above?

ndr-ds (approved):

PR #6660, opened 2026-08-07
Report panics through tracing and a metric, and document our panic policy

ma2bd:
## Motivation

We build with unwinding, and Tokio catches a panic at the task boundary rather than stopping
the runtime. So a panicking task does not stop the process — and nothing in the codebase reacts
to it. The default hook writes the message and backtrace to standard error and that is all:
there is no `tracing` event carrying the panic through whatever subscriber the binary set up,
and no metric, so a panic cannot be alerted on and does not show up in any dashboard.

That matters because the *consequence* of a contained panic is usually invisible too. Whatever
the task was responsible for has simply stopped, and how much that costs depends entirely on who
was joining it — which today is often nobody (see #6659, and the `JoinSetExt` change that
follows this PR).

## Proposal

Add `linera_base::panic_hook::init`, which installs a hook that reports each panic through
`tracing::error!` (thread, location, message) and increments a new `linera_panics_total`
counter, then delegates to the hook that was installed before it, so the standard message and
the `RUST_BACKTRACE` backtrace are still printed. It is idempotent and does not change what a
panic *does* — only what is recorded about it.

It is installed from `linera_service::tracing::init` and its OpenTelemetry counterpart, which
every service binary reaches, rather than from each `main`; `linera-exporter` sets up its own
subscriber, so it calls the hook directly.

Also adds a **Panics** section to `CONTRIBUTING.md`, which had nothing on the subject. It writes
down the policy this PR and its follow-up implement:

- report every panic (this PR);
- prefer crashing over running degraded, for tasks whose death silently disables a subsystem;
- except for work scoped to one request or one chain, where a panic reachable from user input
  must not become a way for one crafted message to stop every validator at once;
- never swallow a panic silently;
- `panic!` / `unwrap` / `expect` are for our own invariants, not for input from peers, clients
  or applications.

## Test Plan

```
cargo test -p linera-base --features metrics panic_hook     # 1 passed
cargo clippy -p linera-base -p linera-service \
    -p linera-exporter --all-targets                        # clean
cargo +nightly fmt -- --check && taplo fmt --check          # clean
```

The unit test covers idempotence — installing twice must not chain the hook to itself, which
would make every panic report grow a line per call.

Manual check that the wiring works end to end is worth doing at review time by panicking a task
in a locally-running validator and confirming both the `error!` event and the counter; I have
not done that here.

## Release Plan

Recommended rather than required: it is observability only, no behaviour change, and
`linera_panics_total` is exactly the signal that is missing when diagnosing stalls on the live
network.

## Links

- Follow-up: `JoinSetExt::await_all_tasks` re-raising panics for supervisory task sets.
- Context: #6659, where a contained panic wedges a cross-chain message queue permanently.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -119,14 +119,13 @@ happens next is decided entirely by whoever joins that task. Choose deliberately
 
 * **Prefer crashing over running degraded.** When a task's death silently disables a
   subsystem — a cross-chain message forwarder, a notification fan-out, a chain listener —
-  the process should exit and be restarted rather than keep serving without it. Await such
-  tasks with `JoinSetExt::await_all_tasks`, which re-raises the panic.
+  the process should exit and be restarted rather than keep serving without it. Whoever
+  joins such a task must re-raise its panic rather than discard it.
 
 * **Except for work that is scoped to one request or one chain.** Handling a single RPC,
   or executing a block for a single chain, must not be able to take the process down: a
   panic reachable from user input would otherwise be a way for one crafted message to stop
-  every validator at once. Contain those panics and let the caller retry, using
-  `JoinSetExt::await_all_tasks_logging_panics` for sets of such tasks.
+  every validator at once. Contain those panics, log them, and let the caller retry.
 
 * **Never swallow a panic silently.** `catch_unwind` is acceptable at a boundary where the
   containment is the point, but log at `error!` and say in a comment why continuing is
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -106,6 +106,37 @@ Two things to keep in mind:
   unawaited futures or unhandled errors. `let _x =` should only be used for RAII guards.
 
 
+## Panics
+
+We build with unwinding, and Tokio catches a panic at the task boundary rather than
+stopping the runtime. A panicking task therefore does not stop the process by itself: what
+happens next is decided entirely by whoever joins that task. Choose deliberately.
+
+* **Report every panic.** `linera_base::panic_hook::init` is installed by
+  `linera_service::tracing::init`, so panics reach the log as a `tracing` event and
+  increment `linera_panics_total`. Do not add a panic hook of your own to a binary that
+  already calls it.
+
+* **Prefer crashing over running degraded.** When a task's death silently disables a
+  subsystem — a cross-chain message forwarder, a notification fan-out, a chain listener —
+  the process should exit and be restarted rather than keep serving without it. Await such
+  tasks with `JoinSetExt::await_all_tasks`, which re-raises the panic.
+
+* **Except for work that is scoped to one request or one chain.** Handling a single RPC,
+  or executing a block for a single chain, must not be able to take the process down: a
+  panic reachable from user input would otherwise be a way for one crafted message to stop
+  every validator at once. Contain those panics and let the caller retry, using
+  `JoinSetExt::await_all_tasks_logging_panics` for sets of such tasks.
+
+* **Never swallow a panic silently.** `catch_unwind` is acceptable at a boundary where the
+  containment is the point, but log at `error!` and say in a comment why continuing is
+  safer than exiting.
+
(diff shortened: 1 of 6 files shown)


linera-base/src/panic_hook.rs
```diff
@@ -0,0 +1,122 @@
...
+}
+
+/// A panic hook, in the form [`std::panic::take_hook`] returns it.
+type PanicHook = Box<dyn Fn(&PanicHookInfo<'_>) + Sync + Send>;
+
+/// The hook that was installed before ours, which we delegate to so that the standard
+/// message and the `RUST_BACKTRACE` backtrace are still printed.
+static PREVIOUS_HOOK: Mutex<Option<PanicHook>> = Mutex::new(None);
+
+static INIT: Once = Once::new();
```
ndr-ds, line 54: Maybe a `OnceLock` here would be better?

linera-base/src/panic_hook.rs
```diff
@@ -0,0 +1,122 @@
...
+
+/// Installs a panic hook that reports panics through `tracing` and the metrics registry
+/// before delegating to the hook that was previously installed.
+///
+/// Calling this more than once has no further effect. It does not change what a panic
+/// *does* — the process still unwinds the panicking task and keeps running — only what is
+/// recorded about it.
+pub fn init() {
+    INIT.call_once(|| {
+        *PREVIOUS_HOOK.lock().expect("hook mutex is never poisoned") =
```
ndr-ds: Might be worth doing
```
LazyLock::force(&metrics::PANICS);
```
here, to make sure the counter is always initialized to 0, and the counter always has data even if no panics are firing

CONTRIBUTING.md
```diff
@@ -106,6 +106,37 @@ Two things to keep in mind:
...
+* **Prefer crashing over running degraded.** When a task's death silently disables a
+  subsystem — a cross-chain message forwarder, a notification fan-out, a chain listener —
+  the process should exit and be restarted rather than keep serving without it. Await such
+  tasks with `JoinSetExt::await_all_tasks`, which re-raises the panic.
+
+* **Except for work that is scoped to one request or one chain.** Handling a single RPC,
+  or executing a block for a single chain, must not be able to take the process down: a
+  panic reachable from user input would otherwise be a way for one crafted message to stop
+  every validator at once. Contain those panics and let the caller retry, using
+  `JoinSetExt::await_all_tasks_logging_panics` for sets of such tasks.
```
ndr-ds: This is introduced in the next PR, right?

ndr-ds (approved):

linera-base/src/panic_hook.rs
ma2bd: Done — `PREVIOUS_HOOK` is now a `OnceLock<PanicHook>` instead of `Mutex<Option<_>>`, so reading it from inside a panic takes no lock and cannot be contended. Kept the `Once` to sequence the swap: `OnceLock::set` alone would not work, because a second `init()` would still call `take_hook()` (uninstalling our hook) before discovering the cell was already full.

linera-base/src/panic_hook.rs
ma2bd: Good catch — done. Without it the series only appears on the first panic, so a dashboard reads "no data" rather than 0 and there is nothing for an alert to sit on. Added a `metrics::register()` that calls `LazyLock::force(&PANICS)`, invoked from `init()`. Put it inside the `#[cfg(with_metrics)] mod metrics` rather than forcing it at the call site, so no `LazyLock` import leaks into builds without the feature.

CONTRIBUTING.md
ma2bd: Yes — both `await_all_tasks` re-raising and `await_all_tasks_logging_panics` come from #6661. Thanks, that was a forward reference. I have reworded the two bullets to state the rule without naming the methods, so this PR stands on its own, and I will add the method names to this section in #6661 once this lands.

PR #6722, opened 2026-08-18
Report the error a Web signer threw instead of an opaque enum code

### LINKED ISSUE #5150: Support throwing arbitrary errors from the Web `Signer` interface
Currently we squash them into an enum, discarding error information.  Instead we should be using an error type (here and in other Web interfaces) similar to `web-thread`'s to capture the error on the Rust side (so that it implements `std::error::Error`).

ma2bd:
## Motivation

Fixes #5150.

The Web `Signer` bridge squashed every JavaScript throw into a `#[repr(u8)]` enum:

```rust
impl From<JsValue> for Error {
    fn from(value: JsValue) -> Self {
        match value.as_f64().and_then(num_traits::cast) {
            Some(0u8) => Error::MissingKey,
            ...
            _ => Error::Unknown,
        }
    }
}
```

`as_f64()` is `Some` only for a JS *number*, and none of our signers throw numbers — `PrivateKey`, `Composite`, `WebCryptoEd25519` and `@linera/metamask` all throw `Error`s, and `WebCryptoEd25519` additionally surfaces `DOMException`s from IndexedDB and Web Crypto. So the match always fell through to `Error::Unknown`, and every failure — a rejected MetaMask prompt, a locked wallet, a wrong connected account, a blocked IndexedDB upgrade — reached the application as the same three words.

It was worse than the enum's own `Display` suggests, because `chain_client::Error` formatted the boxed error with `{0:?}`, so what a developer actually saw in the browser console was:

```
chain client error: signer error: Unknown
```

The `MissingKey`, `SigningError`, `JsConversion` and `Unknown` variants were only ever reachable through that conversion, so nothing in the tree constructed them.

## Proposal

Capture the thrown value instead of classifying it, following the shape of `web-thread`'s error type as the issue suggests.

`error::Thrown` records the `name`, `message` and `cause` of whatever was thrown, recursively, and implements `std::error::Error` with `source()` wired to the `cause` chain. Properties are read reflectively rather than through `dyn_ref::<js_sys::Error>()`, so `DOMException`s, user-defined error classes and errors thrown in another realm are all captured — none of those are `instanceof Error` on our side. Values that are not error-like at all (`throw 42`, `throw "boom"`, `throw {code: 3}`) are reported as themselves, via a total ladder of string → `JSON.stringify` → `String(value)`, the last bound with `catch` so a `toString` that throws cannot unwind through wasm. The `cause` chain is followed to a bounded depth, since nothing stops JavaScript from making an error its own cause.

It lives in `error` rather than `signer` so that the other Web interfaces the issue mentions can use it as they arrive.

`signer::Error` becomes an ordinary `thiserror` enum wrapping `Thrown`. The remaining variants are the ones the bridge itself raises, and two of them get sharper: `InvalidAccountOwnerType` previously covered both "the signer signed with the wrong key" and "this owner has no signature scheme", which are unrelated conditions, and `UnexpectedSignatureFormat` now carries the `AccountOwner` whose scheme the value failed to parse as. The mismatch variant reports both owners.

`chain_client::Error::Signer` switches from `{0:?}` to `{0}`, without which none of the above is visible.

`JsResult<T> = Result<T, JsError>` is gone: it bypassed `error::Error` entirely and was the alias every `Chain`, `Application` and `Formats` method used, including `addOwner`. Those methods now return `Result<T>` like the rest of the crate, so there is one boundary conversion instead of two.

The JavaScript stack of the throw is **not** carried across the boundary. An earlier revision captured it and restored it onto the rethrown error, but `linera_client::Error` truncated its own `source()` chain to a single node, putting the `Thrown` out of reach. That is fixed by #6747, so restoring the stack is now viable — as a follow-up, not here.

The same MetaMask rejection now reads:

```
chain client error: signer error: MetaMask signature request failed: User rejected the request.
```

### Note for reviewers

This removes `SignerError` from `@linera/client`'s exports. It was a Rust-side enum published so that JavaScript could throw its numeric codes back at us; no signer in this repo ever did, nothing in the tree reads it, and `excludeNotDocumented` kept it out of the generated docs. Flagging it because it is nonetheless a removal from the package's public surface.

## Test Plan

`web/@linera/client/tests/SignerError.test.ts` claims one faucet chain, connects to it with a signer whose `sign` throws, and proposes a block per case — the only operation that asks for a signature. Signing is the last step, so a refusal commits nothing and the chain is reusable, which keeps the round-trip cost to one for the whole file. Cases: an `Error`'s message; a `DOMException`'s `name:` prefix; a three-deep `cause` chain; a `cause` that points at itself, asserting both termination and the truncation marker; and the non-error fallbacks (`42`, `{code: 3}`, `undefined`, the last also asserting no `JsValue` debug form leaks). All of these reach `Error::Unknown` on `main`, so the suite should fail there.

These run in the existing `web` CI job, which is the only place they can run — they need the local validator network and the Playwright browser that the job sets up, so I have not run them locally. Verified locally:

- `cargo check` / `cargo clippy` for `linera-web` on `wasm32-unknown-unknown` with the pinned nightly, from `web/` so the target's `rustflags` apply: clean, the only warning being the pre-existing `missing_errors_doc` in `crypto.rs`.
- `cargo build --lib` followed by `wasm-bindgen 0.2.100` with `build.bash`'s flags: the glue generates, the `String` extern binds to the global as intended, and `SignerError` is gone from the emitted `.d.ts` with the rest of the API unchanged.
- `cargo check -p linera-core`: clean.
- `cargo +nightly fmt --check` and `taplo fmt --check`: clean.

## Release Plan

## Links

- Fixes #5150.
- linera-io/web-thread#2 fixes an inverted `Option::filter` in the reference implementation, which dropped every real `cause` and appended a spurious node in its place. The version here does not share the bug.


ma2bd:
Review notes on this branch. Nothing below is a correctness bug — the change is a clear improvement over the numeric-code funnel — but the first three are worth resolving before merge.

## 1. `stack` and `cause` are captured but never surface in the browser

`Thrown` records the stack and the whole `cause` chain and wires `source()`, but the only bridge from a Rust error to JS is wasm-bindgen's `impl<E: std::error::Error> From<E> for JsError`, which is literally `JsError::new(&error.to_string())`. `Display` for `Thrown` is `f.write_str(&self.description)` (`web/@linera/client/src/error/mod.rs:128-132`) — description only. So:

- `Thrown::stack()` (`error/mod.rs:68-72`) has **zero call sites** in the tree. The stack is read on every throw and dropped.
- The `cause` chain is invisible on the normal `Result` → `JsValue` path. `new Error("Request failed", { cause: rpcError })` still prints only `Request failed` — exactly the MetaMask shape the motivation section describes. The one place the extra data does appear today is `listener.rs:31`'s `{error:?}`.

Reading `.stack` is also not free — on V8 it is a lazy getter that formats the trace on access.

Two ways to make it land: fold the chain into `Display` (`{description}: {cause}`), or, better, stop going through `JsError` for this and rebuild a real JS error with `js_sys::Error` + `set_cause`/`set_stack`, so the browser console renders the chain natively. `error/index.ts`'s `LockError(message, stack?)` shows we already reach for that idea — though its Rust binding only passes `message`, so the `stack` parameter is unused there too.

## 2. `JsValue(...)` leaks into user-facing messages

`format!("{value:?}")` at `error/mod.rs:94` and `error/mod.rs:108-112` uses wasm-bindgen's `Debug`, which is `write!(f, "JsValue({})", self.as_debug_string())`. So a `Promise.reject()` with no argument — not exotic — produces:

```
chain client error: signer error: JsValue(undefined)
```

Same for `throw Symbol("x")` and BigInts (`JSON.stringify` throws on those). The `JsValue(...)` wrapper is a Rust internal in a message aimed at a JS developer; `value.js_typeof()` or a stripped debug string reads better.

## 3. The intricate half of `capture` is the untested half

The three TS tests cover `Error`, `DOMException` and a bare string — all of which land in the flat, straightforward paths. Untested: the `cause` recursion, the `MAX_CAUSE_DEPTH` guard, the cyclic-cause case the comment calls out, and the JSON/debug fallbacks. That is the code most likely to be wrong, and it is precisely what the linked `web-thread` bug was.

`wasm-bindgen-test` is already a workspace dep with browser precedent (`linera-views/tests/store_tests.rs`, `linera-rpc/tests/transport.rs` both use `wasm_bindgen_test_configure!(run_in_browser)`). `Thrown::capture` is a pure `JsValue → String` function; unit-testing it there would cover all of the above in milliseconds. As written, each of the three tests claims a faucet chain and carries a 150 s timeout to exercise one string-formatting path — a lot of validator round-trips per assertion.

If we would rather keep it in TS, at minimum `const e = new Error("loop"); e.cause = e;` is a cheap regression test for the guard.

## 4. Smaller points

- **`SignatureFormat { scheme: &'static str }`** (`signer/mod.rs:24`) is stringly-typed in a public error enum. Two variants, or a small enum, would let callers match on it. Minor, since only `Display` consumes it today.
- **Silent truncation at depth 8** (`error/mod.rs:114`): a chain deeper than `MAX_CAUSE_DEPTH` is cut with no marker, so a truncated chain is indistinguishable from a complete one. An ellipsis node would fix it — though this only matters once (1) makes the chain visible at all.
- **`Thrown` holding `String`s rather than a `JsValue`** is the right call and worth keeping deliberate: it keeps the error `Send + Sync`, so it satisfies `TaskSendable` on both sides of the `cfg(web)` split rather than only the web one.

## 5. Checked and fine

- **`{0:?}` → `{0}` on `chain_client::Error::Signer`** affects every signer, not just the web one. The other two implementors (`InMemorySigner`, and `Keystore`, which aliases its error) both use `thiserror` with real messages — `Display` is strictly better than `Debug` for them. No regression.
- **`SignerError` removal**: no references anywhere in the tree — no `.ts`, `.js`, `.rs`, example, or README — and nothing in the committed `docs/` output, so no typedoc regeneration needed. Worth the changelog note, but nothing breaks in-repo.
- **`num-traits` drop** from `web/@linera/client/Cargo.toml`: no remaining uses in the crate; `lint-cargo-machete` agrees.
- **Test setup** matches `Chain.test.ts`'s established shape, and the fake signer satisfies the full `Signer` interface (`sign`/`getPublicKey`/`containsKey`).

ma2bd:
All three blocking points addressed, plus both actionable smaller ones.

## 1. `stack` and `cause` now reach the browser

Took the second option — the bridge builds a real JavaScript error rather than going through `JsError`.

`error::Error::Other` now holds a `js_sys::Error`, and `to_js_error` constructs it. Every layer of a Rust error chain already prints the layer below it, so the message needs nothing extra; what it cannot carry is the JavaScript stack, so `to_js_error` walks the `source` chain for the first `Thrown` and sets `.stack` from it. `Thrown::stack()` is its only caller — the accessor is no longer dead.

Two things had to change to make that walk arrive:

- `JsResult<T> = Result<T, JsError>` bypassed `error::Error` entirely, and it was the alias every `Chain` and `Application` method used — including `addOwner`, the operation in the motivating example. It is gone; those methods now return `Result<T>` like the rest of the crate, so there is one boundary conversion instead of two.
- `signer::Error::Thrown` can no longer be `#[error(transparent)]`. thiserror expands `transparent` to `Error::source(inner.as_dyn_error())`, which skips the `Thrown` and returns *its* cause, so the stack was unreachable. `#[error("{0}")]` with `#[from]` gives the same `Display` and keeps the `Thrown` itself as the source. There is a note in the code, since the two attributes look interchangeable here.

Separately, `Thrown`'s `Display` now renders its `cause` chain (`cannot sign: the vault will not open: the key is missing`) rather than just its own description — matching how every other error type in the stack composes, and making the chain visible on the logging and metric paths too, not only at the JS boundary.

Agreed on the `LockError(message, stack?)` observation; its Rust binding still passes only the message. Left alone, since lock contention has no JavaScript throw behind it to take a stack from.

## 2. No more `JsValue(...)`

Non-error throws now go through a ladder: the string itself, else `JSON.stringify`, else `String(value)` bound as an extern. So `Promise.reject()` gives `undefined`, `throw 10n` gives `10`, `throw Symbol("x")` gives `Symbol(x)`, and a cyclic object gives `[object Object]` instead of a Rust-internal debug string. `format!("{value:?}")` is gone from the module.

## 3. Coverage

Restructured rather than moved. The three tests shared nothing; they now share one faucet chain claimed in `beforeAll`, with the thrown value swapped per case and `clearPendingProposal` between them, so the round-trip cost is paid once instead of once per assertion. That made the intricate paths cheap enough to cover:

- the `cause` recursion, asserting the full chain in the message
- the cyclic case, `e.cause = e`, asserting both that it terminates and that the truncation marker appears
- the JSON and `String` fallbacks (`42`, `{code: 3}`, `undefined`)
- the stack propagation from (1), asserting the thrown error's `stack` arrives verbatim

That is everything listed, though not at millisecond speed. I did not try `wasm-bindgen-test` here: the crate only builds under `web/.cargo/config.toml` (`build-std`, `+atomics`, `--shared-memory`, `--import-memory`), so the harness would need its own flag set and CI step, and I could not have validated either locally. Worth doing as a follow-up if the browser round-trips become the bottleneck — happy to file it.

## 4. Smaller points

- **`SignatureFormat`**: the `&'static str` is gone. The variant now carries the `AccountOwner`, which is matchable, determines the scheme anyway, and says which owner failed.
- **Silent truncation**: a chain deeper than `MAX_CAUSE_DEPTH` now ends in a `(further causes omitted)` node, so a cut chain is distinguishable. Covered by the cyclic test.
- **Owned `String`s over a `JsValue`**: made deliberate in the doc comment, with the `TaskSendable` reason.

One cost worth naming since you raised it: `.stack` is still read on every capture, before we know whether anything will ask for it. It is one lazy getter per failed signature, which seemed the wrong thing to optimise against losing the trace, but say the word and I will make it lazy.

ma2bd:
Two CI failures since the last comment, and the second one retracts part of what I claimed there.

## Compile failure

#6717 landed `ownerBalance` while this was open, written against the `JsResult` alias I had just deleted, so the merge of branch and main did not compile. Ported it to `Result<String>` like its neighbours.

Merged `origin/main` in rather than rebasing: publishing a rebase needs `--force-with-lease`, which my sandbox hard-blocks. Since this repo squash-merges, the merge commit disappears on landing and does not show in the PR diff — but say so and I will get it rebased instead.

## The stack cannot reach the boundary — reverted

Every other new test passed, including the `cause` chain, the cyclic guard and the JSON/`String` fallbacks. `carries the stack of the thrown error` failed, and the reason is not fixable from here:

```rust
// thiserror-context 0.1.2, src/lib.rs:324
impl std::error::Error for $out {}
```

`linera_client::Error` is that generated type, and that impl is empty — so its `source()` is the default `None`. It is the first node `to_js_error` was handed, so the walk terminated immediately and never reached the `Thrown`. The failure output shows it plainly: the returned error carried its own freshly-constructed wasm stack rather than the signer's.

`Display` is unaffected, since the macro routes it through `AsRef` instead of `source` — which is why the message chain works and only the stack did not. And `Inner` is `pub(crate)`, so linera-web cannot step past the newtype either.

So the stack capture is gone, along with `to_js_error`, the `js_sys::Error` change and the `'static` bound that existed only to serve it. That leaves your original observation resolved in the other direction: nothing reads the stack, and now nothing pays the lazy-getter cost to collect it. The `cause` chain still surfaces, through `Display`, with a passing test.

What survives from that work and is worth keeping:

- `JsResult` is still gone. It was a second boundary conversion that bypassed `error::Error` entirely; folding it in is a cleanup rather than a requirement now, so split it out if you would rather this PR stayed narrow.
- `signer::Error::Thrown` stays `#[error("{0}")]` rather than `#[error(transparent)]`. It no longer matters for the stack, but `transparent` forwards `source` past the `Thrown` to its cause, so a caller walking the chain would never see the throw itself. The code comment now says that instead of naming `to_js_error`.

The empty `Error` impl looks worth its own issue: it silently truncates the source chain of *every* `linera_client::Error`, so anything walking `source()` — not just this bridge — sees one node and stops. Happy to file it, and the stack propagation becomes a small follow-up once it is fixed.

ma2bd:
Re-reviewed at `7613c494c7`. CI is green (33 pass, 8 skipping), and all three blocking points are resolved. One new defect, introduced by the fix for (2), plus some refinements.

## New: `js_string` can throw, in the one place that must not

The `describe` ladder ends at `js_string`, bound without `catch`:

```rust
#[wasm_bindgen(js_name = "String")]
fn js_string(value: &JsValue) -> String;
```

`String(value)` indeed does not throw for `undefined`, symbols or bigints — the doc comment is right about those. But it throws `TypeError: Cannot convert object to primitive value` for an object with no `toString`, and propagates whatever a throwing `toString` raises. Without `catch`, that exception unwinds straight through the wasm frames instead of returning an `Err`.

It is reachable, though only via the last rung — the value has to defeat `JSON.stringify` first:

```js
const o = Object.create(null); o.self = o; throw o;   // stringify: cyclic → String(o): TypeError
throw new Proxy({}, { get() { throw 1 }, ownKeys() { throw 1 } });
throw { toJSON() { throw 1 }, toString() { throw 1 } };
```

Narrow, but this is the function whose entire contract is "describe anything, never fail", and an uncatchable throw in the fallback contradicts that — the `format!("{value:?}")` it replaced could not do it. `#[wasm_bindgen(catch)] fn js_string(value: &JsValue) -> Result<String, JsValue>` with `.unwrap_or_else(|_| "(unrepresentable value)".to_owned())` closes it.

## The retraction checks out

Pulled the crate rather than take it on trust. `thiserror-context-0.1.2/src/lib.rs`:

```rust
318:  impl std::fmt::Display for $out {
319:      fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
320:          f.write_fmt(format_args!("{}", self.as_ref()))
321:      }
322:  }
323:
324:  impl std::error::Error for $out {}
```

Exactly as described: empty impl, so `source()` is the default `None`, while `Display` routes through `AsRef` instead — which is precisely why the message chain survives and only the stack did not. `Inner` is `pub(crate)` (`linera-client/src/error.rs:16`), so the generated `AsRef<Inner>` is unusable from linera-web and there is no way past the newtype from outside. Dropping the stack capture was the right call.

Please do file the follow-up. It is broader than this PR: every `linera_client::Error` truncates its own source chain to a single node, so anything walking `source()` — error reporters, `anyhow`-style `{:#}` formatting, future bridges — silently sees one level. Worth an upstream PR to `thiserror-context` too; the fix there is a one-line `fn source()` delegating to `self.as_ref()`.

## Smaller points

- **The `undefined` assertion does not guard what it was written for.** `expect((await refusing(undefined)).message).toContain("undefined")` passes just as well against the old `JsValue(undefined)` output — the string it is meant to rule out contains the string it asserts. An `expect(message).not.toContain("JsValue")` alongside it would actually pin the fix.
- **`SignatureFormat` lost the scheme from its text.** Carrying the `AccountOwner` is the better shape, but the message went from "not a valid secp256k1 signature" to "not a valid signature for owner 0x…", which asks a JavaScript developer to know that Address20 implies secp256k1. Both fit: keep the owner in the variant, name the scheme in the string.
- **`Display` chaining plus `source()` now double-report** for any consumer that walks the chain and prints each node. Both were deliberate and it matches how the rest of the stack composes, and it is moot on the web path while `linera_client::Error` truncates — just worth knowing the redundancy is by construction rather than by accident.
- **Merge rather than rebase**: fine by me. The repo squash-merges, the merge commit does not survive landing, and the three-dot diff confirms main's changes are not in the PR diff. Not worth a force-push fight with the hook.
- **`JsResult` removal**: no longer motivated by the stack work, but it stands on its own — it was a second boundary conversion bypassing `error::Error`, and folding it in is a genuine simplification. I would keep it here. Worth noting it also means `Chain` and `Application` methods can now surface `LockError` rather than a plain `Error`; nothing does today, since `lock::Error` deliberately does not implement `std::error::Error`, but the door is open in a way it was not before.

Nothing above blocks except the `catch`, which is one attribute and one `unwrap_or_else`.

ma2bd:
Fixed at `432211452b`. You were right about `js_string`, and the reachability argument is exact.

## The `catch`

`String(value)` throwing was a real hole, and worse than "narrow" in one respect: `describe` is reached from `Thrown::capture`, which is on the `?` path of every `Signer` call, so an unwinding throw there would escape a function whose whole job is to *describe* a failure that already happened.

Bound with `catch` and made the ladder total:

```rust
#[wasm_bindgen(catch, js_name = "String")]
fn js_string(value: &JsValue) -> Result<String, JsValue>;
```

```rust
    .or_else(|| js_string(value).ok())
    .unwrap_or_else(|| "(unrepresentable value)".to_owned())
```

Verified in the generated glue rather than assumed — the binding now emits `handleError(function (...) { const ret = String(arg1); ... }, arguments)`, so the throw becomes an `Err` instead of unwinding. The doc comment no longer claims `String` never throws; it says which cases it covers and that the result is caught.

## Smaller points

- **The `undefined` assertion**: correct, it was self-satisfying — `JsValue(undefined)` contains `undefined`. Added `expect(undef).not.toContain("JsValue")` next to it, which is what actually pins the change.
- **`SignatureFormat`**: both, as suggested. The variant keeps the `AccountOwner`, and the message names the scheme through a `signature_scheme(.0)` helper in the format args, so it reads "not a valid secp256k1 signature for owner 0x…" again without the field going back to `&'static str`.
- **`Display` chaining plus `source()` double-reporting**: agreed it is by construction. Noting for the record that it is the same shape as `#[error("chain error: {0}")] Chain(#[from] ChainError)` elsewhere in the stack, so a chain-walking consumer would already see this pattern from other error types, not just this one.
- **`JsResult` removal**: keeping it, and thank you for the `LockError` observation — I had not thought about the widened return surface. Nothing constructs a `lock::Error` on those paths today, and `lock::Error` not implementing `std::error::Error` is what keeps the blanket `From` from swallowing it, so the door only opens if someone adds a lock-taking call to a `Chain`/`Application` method. Worth a comment on `lock::Error` if you want that invariant written down rather than inferred.

## Follow-up filed

#6735, with the `impl_context!` expansion, the three affected types in this repo, and the one-line upstream fix. `thiserror-context` is third-party (tmyers273/error-context), so the choice is an upstream PR plus a version bump versus doing it locally — I left that open in the issue rather than picking for you.

Unrelated but adjacent, from cross-checking the Sentry data: `linera-faucet/client/src/lib.rs:33` has `#[error("HTTP error: {0:?}")] Http(#[from] reqwest::Error)`, and on wasm `reqwest::Error`'s `Debug` embeds a `JsValue(...)` carrying a whole JS stack trace. That message is the top wasm-related error in the web client's Sentry project — ~13.2k events across 5 issues in 90 days, the largest alone 10,932 events / 1,951 users, all "cannot reach the conway faucet". The embedded stack varies by browser and bundle hash, which is also why one failure is split across five issues. Same `{0:?}`-on-a-JsValue mistake as `chain_client::Error::Signer`. Happy to take it, though it wants a check that reqwest's `Display` keeps the underlying `TypeError` text on wasm first, so we do not trade noise for lost detail.

ma2bd:
Checked `432211452b`. CI green (34 pass, 8 skipping). All three fixes are correct. The faucet plan at the end has a problem, though.

## The fixes

**`js_string`** — `catch` plus `.or_else(|| js_string(value).ok()).unwrap_or_else(|| "(unrepresentable value)".to_owned())` makes the ladder total. Correct.

**`signature_scheme(.0)`** — I checked this one empirically rather than by eye, because in a plain `format!` the mix is a trap: `"not a valid {} signature for owner {0}"` would have `{}` and `{0}` both consume positional arg 0, printing "not a valid secp256k1 signature for owner secp256k1". thiserror binds field references separately, so it is fine — built it against thiserror 1.0.69 and got:

```
the signer returned a value that is not a valid secp256k1 signature for owner 0xA20-7
the signer returned a value that is not a valid Ed25519 signature for owner 0xA32-9
```

Worth recording, since it compiles either way and nothing tests `SignatureFormat` — the collision would have shipped silently if thiserror behaved like `format!`.

**Test assertion** and **#6735** both good. Confirmed `linera_faucet_client::Error` is genuinely an `impl_context!` type (`linera-faucet/client/src/lib.rs:52`), so the issue's list of three is right.

## The faucet change will not do what you want

The mechanism you identified is real, but `{0:?}` → `{0}` will not remove the stack trace, because the stack is not in reqwest's `Debug` framing — it is inside the source string, and `Display` prints that too. Three links, all verified:

1. **reqwest 0.11.27** — the workspace pins `0.11.24` and the lock resolves 0.11.27, not 0.12 — `src/error.rs:283-285`:

```rust
pub(crate) fn wasm(js_val: wasm_bindgen::JsValue) -> BoxError {
    format!("{js_val:?}").into()
}
```

2. **wasm-bindgen 0.2.100**'s generated `debugString`, `wasm-bindgen-cli-support/src/js/mod.rs:4248-4250`:

```js
if (val instanceof Error) {
    return `${val.name}: ${val.message}\n${val.stack}`;
}
```

3. **reqwest's `Display`**, `src/error.rs:209-211`, appends `: {e}` for the source — as does `Debug`, `src/error.rs:177-179`.

So both renderings carry it:

```
{0:?}  reqwest::Error { kind: Request, url: "https://faucet…", source: "JsValue(TypeError: Failed to fetch\n<stack>)" }
{0}    error sending request for url (https://faucet…): JsValue(TypeError: Failed to fetch\n<stack>)
```

The switch only strips the `reqwest::Error { … }` framing. The per-browser, per-bundle-hash variance survives, and Sentry keeps splitting one failure across issues.

It is also not the same mistake as `chain_client::Error::Signer`. There `{0:?}` printed *our* enum, whose `Debug` was the word `Unknown`, so `Display` was strictly more informative. Here the noise sits below both formats, inside a third-party crate's source string.

A fix has to stop printing reqwest's source at all — build the message from the stable predicates (`is_timeout()`, `is_connect()`, `is_status()`, `status()`, `url()`) and drop the rest, so "cannot reach the faucet" renders identically everywhere. Sentry fingerprint rules would be the stopgap. Either way it is a different change from the one-character one, which argues for the separate PR you suggested rather than folding it in here.

I cannot check the Sentry figures themselves — no access to that project — so I am taking the 13.2k / 5-issue split on trust. The mechanism that would produce exactly that fragmentation is confirmed.

## Minor

- `signature_scheme`'s `Reserved(_) => "unsupported"` arm is unreachable in practice: that owner returns `ReservedOwner` before a `SignatureFormat` can be built. Harmless as a total match, it would just read oddly if it ever printed.
- On the `lock::Error` invariant: yes, worth one line on the type. "Deliberately does not implement `std::error::Error`, so the blanket `From` in `error` cannot swallow it into `Other`" is the kind of thing that gets helpfully fixed by a future derive.

Nothing blocking here — this is good to land as far as I am concerned.

ma2bd:
Thanks — and the faucet correction is right. I checked it rather than take it on trust, since it was my analysis that was wrong, and every link holds.

## The faucet plan was wrong

`reqwest-0.11.27/src/error.rs:283-285`, exactly as you quote:

```rust
pub(crate) fn wasm(js_val: wasm_bindgen::JsValue) -> BoxError {
    format!("{js_val:?}").into()
}
```

and the `debugString` in this repo's own generated glue — I still had the `wasm-bindgen 0.2.100` output from verifying the `String` extern — confirms where the stack enters:

```js
if (val instanceof Error) {
    return `${val.name}: ${val.message}\n${val.stack}`;
}
```

So the trace is baked into a `String` before reqwest ever formats it, and `Display` appends it via `if let Some(e) = &self.inner.source { write!(f, ": {e}")?; }` at `src/error.rs:209-211` just as `Debug` does at `177-179`. Switching the format specifier strips the `reqwest::Error { … }` framing and nothing else. The per-browser, per-bundle-hash variance survives, so the Sentry fragmentation survives with it.

I had also called it "the same one-character mistake" as `chain_client::Error::Signer`. It is not, and your framing of why is the useful part: there `{0:?}` printed *our* enum, whose `Debug` was the word `Unknown`, so `Display` was strictly better. Here the noise is below both formats, inside a third-party crate's source string, and no choice of specifier reaches it.

So the real fix is to stop printing reqwest's source and build the message from the stable predicates — `is_timeout()`, `is_connect()`, `is_status()`, `status()`, `url()` — so that "cannot reach the faucet" renders identically in every browser. That is a different change from the one I described, and it wants its own PR with its own reasoning about what detail we are willing to lose. I will not fold it in here.

On the Sentry figures: fair to treat as unverified from your side. For the record they came from the `sentry-infinity` datasource in Grafana (`monitoring.infra.linera.net`), querying `/api/0/organizations/linera-io/issues/` for project `4511190365110272` over `statsPeriod=90d` — the same path the `web-ui-errors` dashboard uses. Anyone with a Grafana service-account token can reproduce them; there is no separate Sentry access needed.

## `signature_scheme(.0)`

Good instinct to check that empirically. I had reasoned it from thiserror's documented trailing-args form rather than run it, and "compiles either way, nothing tests it" is exactly the shape of thing that ships wrong. Worth noting the collision would also have been invisible in review, since both spellings look correct.

## Minor points, both taken

- `signature_scheme`'s `Reserved` arm now carries a line saying reserved owners are rejected as `ReservedOwner` before any signature is parsed, so the arm exists only to keep the match total.
- `lock::Error` now documents the invariant on the type: it deliberately does not implement `std::error::Error`, because the blanket `From` in `error` would otherwise absorb it into `Error::Other` and the JavaScript side would lose the `LockError` class it matches on.

Pushed as `dbadd45a47`. Both are comment-only; clippy and fmt still clean.

ma2bd:
Checked `dbadd45a47` and the merge at `ee4f752f3c`, which is green (34 success, 8 skipped, no failures). The `signature_scheme` note is right. One correction on the other one.

## The `lock::Error` comment states the wrong consequence

> the blanket `From` impl in `crate::error` would otherwise absorb it into `Error::Other`

It cannot absorb it. The explicit `From<lock::Error>` and the blanket `From<E: std::error::Error>` overlap, so adding the impl is a hard compile error. Reproduced the exact shape from `error/mod.rs` plus `lock.rs`:

```
error[E0119]: conflicting implementations of trait `From<lock::Error>` for type `Error`
  --> src/main.rs:32:1
   |
26 | impl From<lock::Error> for Error {
   | -------------------------------- first implementation here
...
32 | impl<E: std::error::Error> From<E> for Error {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ conflicting implementation for `Error`
```

Without the impl it builds and routes to `Error::Lock`, as today.

This matters for a comment whose whole job is to record an invariant. As written it reads as a silent behavioural hazard needing vigilance; it is actually compiler-enforced, which is both stronger and more reassuring.

The flip side is what the comment misses. The protection is the **pair** — the explicit `From<lock::Error>` together with the absent `Error` impl. Drop the explicit `From` in some future refactor and the blanket one would then quietly absorb `lock::Error` into `Other`, exactly as the comment describes, with nothing to stop it. That is the failure mode worth writing down, and the current wording obscures it by attaching it to the wrong trigger.

Something along the lines of: adding `std::error::Error` here is E0119 against the explicit `From` in `crate::error`; that explicit impl is what keeps `lock::Error` out of `Error::Other` and preserves the `LockError` class JavaScript matches on.

Nothing else outstanding from my side.

ma2bd:
Right on both counts, and the second point is the more useful one. Reproduced it rather than take it on trust, with the `error` + `lock` shapes and the `std::error::Error` impl behind a feature flag:

```
--- without the impl ---
routed to: Error::Lock

--- with the impl ---
error[E0119]: conflicting implementations of trait `From<lock::Error>` for type `Error`
```

So the comment had it backwards in the way that matters: it described something to stay vigilant about, when the compiler already refuses it outright. And it attached the hazard to the wrong trigger — the fragile half is the explicit `From<lock::Error>`, not the missing `Error` impl. Deleting that `From` in a future refactor is what would silently route `lock::Error` into `Other` and lose the `LockError` class, and nothing would complain.

Rewritten as:

```rust
/// Adding a [`std::error::Error`] impl here is an E0119 conflict against the explicit
/// `From<lock::Error>` in [`crate::error`]. That explicit impl is what keeps this type out
/// of `Error::Other`: drop it and the blanket `From` absorbs this type, losing the
/// `LockError` class that JavaScript matches on.
```

Pushed to both this PR (`3ffcd1fbc3`) and #6780 (`5d0524ae1f`), so the branches stay in sync. Comment-only; fmt clean on each branch's own pinned nightly.

One thing that needs you rather than another push: this PR has **no formal review** — everything so far has come through as issue comments, so GitHub reports `mergeable_state=blocked`. It needs an approval submitted through the review UI before it can land.

PR #6737, opened 2026-08-20
Load a collection entry in a single round trip

ma2bd:
## Motivation

Loading an entry from a collection cost two **sequential** round trips. `try_load_entry` asked
whether the entry exists, and only then loaded the subview:

```rust
if self.context.store().contains_key(&key_index).await? {   // round trip 1
    let view = W::load(context).await?;                     // round trip 2
```

The existence marker is `Index`-tagged and the subview's initialization keys are
`Subview`-tagged, so the second read could not start until the first returned. The batch
variants had the same shape: `contains_keys`, then `read_multi_values_bytes`.

Both `ReentrantByteCollectionView` and `ByteCollectionView` were affected, in both their
single-entry and batch paths — four sites in total.

**The highest-frequency caller is application storage, not the chain state.**
`execution_state_actor.rs` calls `state.users.try_load_entry(&id)` on *every* `ContainsKey`,
`ReadValueBytes`, `ReadMultiValuesBytes`, `FindKeysByPrefix`, `FindKeyValuesByPrefix` and
`WriteBatch` an application issues. `users` is a
`ReentrantCollectionView<_, _, KeyValueStoreView<_>>` and `KeyValueStoreView::NUM_INIT_KEYS` is
1, so every application storage operation goes from two sequential round trips to one query of
two keys. `ChainStateView::inboxes` and `outboxes` benefit too, but far less often.

## Proposal

Read the index marker and the subview's initialization keys in one `read_multi_values_bytes`,
then use the marker to decide whether to build the subview.

The marker cannot be eliminated or inferred — it is what distinguishes an absent entry from a
present-but-empty one. A subview that was opened and never written stores no bytes of its own
(`RegisterView::pre_save` only emits a key when the register was assigned), so the `Index` key
with its empty value is the sole record of membership. It can, however, be fetched in the same
batch. Presence and not content is what decides, so the marker is read back as `Some(vec![])`
and tested with `is_some()`.

The key layout both collections share now lives in one private module,
`views::collection_entry`: the `KeyTag` enum (which the two files declared identically), the
`index_key` / `subview_key` / `subview_context` derivations, and `entry_keys` /
`post_load_entry`. That also collapses 17 inline `base_tag_index(KeyTag::… as u8, …)` sites
across the two files, so the derivation now has one spelling rather than four.

`entry_keys` returns the subview context it derived rather than taking it as a parameter, so a
caller cannot pass a context and short key that disagree. `post_load_entry` treats a short read
as `ViewError::MissingEntries` rather than as an absent entry, since fewer values than keys is a
broken store contract and not a missing entry.

### Cost

Never worse in round trips, and better whenever the entry exists:

| | before | after |
|---|---|---|
| entry exists | 2 round trips (1 key, then `NUM_INIT_KEYS`) | 1 round trip (`1 + NUM_INIT_KEYS` keys) |
| entry absent | 1 round trip (1 key) | 1 round trip (`1 + NUM_INIT_KEYS` keys) |

On a miss the extra initialization keys resolve to `None`, so the cost is a wider batched query
rather than an additional round trip.

For the two hot **batch** callers the hit rate is 100% by construction and enforced:
`chain_worker/state.rs` iterates `chain.nonempty_inboxes` and turns a `None` into
`ChainError::InternalError`, and `chain.rs`'s `load_outboxes` collects into `Option<Vec<_>>` and
errors with `CorruptedChainState`. Those are pure 2→1 wins with no widening at all.

Where a miss widens most is the single-entry paths on chain state:
`InboxStateView::NUM_INIT_KEYS` is 5 (three registers, two queues), so an inbox miss in
`get_inbox_next_height` — a sender asking a receiver that has never heard from it — goes from 1
key to 6. `OutboxStateView` is 2, so `get_tip_state_and_outbox_info` goes 1→3. Both are
per-chain and infrequent. Note also that a batch above 99 keys is split by ScyllaDB's
`read_multi_values_bytes` into concurrent sub-queries under a single semaphore permit.

## Test Plan

`test_collection_view_try_load_entries_with_absent_entries` and its reentrant twin cover a batch
that mixes hits and misses with a hole in the middle — entries 1 and 5 hold values, 3 exists but
was never written to (index marker only), 0/2/4/6 are absent — which is exactly what the
chunk-and-zip alignment has to get right, and which no existing test reached. They have teeth:
moving the marker to the end of each entry's key run in `entry_keys` fails both.

- `cargo test -p linera-views --features metrics` — 377 passing.
- `cargo test -p linera-chain` — 80 passing; this exercises the inbox and outbox collections.
- `cargo clippy --locked --all-targets --all-features --workspace` — clean.
- Nightly `cargo fmt --check`.

## Release Plan

Operational note for whoever deploys this: `contains_key` volume drops sharply and reappears as
wider `read_multi_values_bytes` batches, so any dashboard or alert thresholded on
`*_contains_key_latency` rates will shift.

## Links

- #6744 — backport to `testnet_conway`.

### DIFF
--- a/linera-views/src/views/collection_entry.rs
+++ b/linera-views/src/views/collection_entry.rs
@@ -1,0 +1,87 @@
+// Copyright (c) Zefchain Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! The key layout of a collection entry, shared by the collection views.
+//!
+//! An entry is stored as an index marker recording that it exists, plus — under a separate tag —
+//! the key space of its subview. The marker is what distinguishes an absent entry from one that
+//! is present but has written no bytes of its own, so it cannot be inferred from the subview's
+//! keys. It is read in the same batch instead, which is what makes loading an entry cost a single
+//! round trip whether or not the entry exists.
+
+use crate::{
+    context::Context,
+    views::{View, ViewError, MIN_VIEW_TAG},
+};
+
+/// The prefixes of the keys of a collection.
+#[repr(u8)]
+pub(crate) enum KeyTag {
+    /// Prefix for specifying an index and serves to indicate the existence of an entry in the
+    /// collection.
+    Index = MIN_VIEW_TAG,
+    /// Prefix for specifying as the prefix for the sub-view.
+    Subview,
+}
+
+/// The key marking the existence of the entry at `short_key`.
+pub(crate) fn index_key<C: Context>(context: &C, short_key: &[u8]) -> Vec<u8> {
+    context
+        .base_key()
+        .base_tag_index(KeyTag::Index as u8, short_key)
+}
+
+/// The base key of the subview stored at `short_key`.
+pub(crate) fn subview_key<C: Context>(context: &C, short_key: &[u8]) -> Vec<u8> {
+    context
+        .base_key()
+        .base_tag_index(KeyTag::Subview as u8, short_key)
+}
+
+/// The context of the subview stored at `short_key`.
+pub(crate) fn subview_context<C: Context>(context: &C, short_key: &[u8]) -> C {
+    context.clone_with_base_key(subview_key(context, short_key))
+}
+
+/// The number of keys [`entry_keys`] produces for one entry.
+///
+/// Never zero, because of the marker: that is what makes `chunks_exact` safe when a batch of
+/// entries is split back up, even for a subview type with no initialization keys.
+pub(crate) fn entry_len<W: View>() -> usize {
+    1 + W::NUM_INIT_KEYS
+}
+
+/// The subview's context, and the keys that decide whether the entry at `short_key` exists and,
+/// if it does, load it: the index marker followed by the subview's initialization keys.
+pub(crate) fn entry_keys<W: View>(
+    context: &W::Context,
+    short_key: &[u8],
+) -> Result<(W::Context, Vec<Vec<u8>>), ViewError> {
+    let subview_context = subview_context(context, short_key);
+    let mut keys = Vec::with_capacity(entry_len::<W>());
+    keys.push(index_key(context, short_key));
+    keys.extend(W::pre_load(&subview_context)?);
+    Ok((subview_context, keys))
+}
+
+/// Builds the subview out of the values read for [`entry_keys`], or `None` when the index marker
+/// is absent.
+///
+/// The marker is stored with an empty value, so its presence and not its content is what decides.
+pub(crate) fn post_load_entry<W: View>(
+    subview_context: W::Context,
+    values: &[Option<Vec<u8>>],
+) -> Result<Option<W>, ViewError> {
+    let Some((marker, init_values)) = values.split_first() else {
+        // The store returned fewer values than we gave it keys, which is a broken contract
+        // rather than an absent entry.
+        return Err(ViewError::MissingEntries(format!(
+            "{:?}",
+            subview_context.base_key().bytes
+        )));
+    };
+    if marker.is_none() {
+        return Ok(None);
+    }
+    Ok(Some(W::post_load(subview_context, init_values)?))
+}
--- a/linera-views/src/views/collection_entry.rs
+++ b/linera-views/src/views/collection_entry.rs
@@ -8,6 +8,11 @@
 //! is present but has written no bytes of its own, so it cannot be inferred from the subview's
 //! keys. It is read in the same batch instead, which is what makes loading an entry cost a single
 //! round trip whether or not the entry exists.
+//!
+//! The tags are what keep those two key spaces apart. Sub-views in a collection share a common
+//! key prefix, like in other view types, so appending a sub-view's own keys straight to that
+//! prefix would leave a child sub-view's key indistinguishable from a grandchild's — consider a
+//! collection stored inside a collection.
 
 use crate::{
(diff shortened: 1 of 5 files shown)


ndr-ds (approved):

ma2bd:
Reviewed both files at the head of the branch. The change is correct and behaviour-preserving; the four rewritten sites are equivalent to the old ones in every branch I could construct. I pushed one commit adding test coverage for a gap (details below), and the same commit to #6744.

### Correctness

- **Marker semantics.** `post_load_entry` keys off `marker.is_some()`, matching the old `contains_key`. All three states — absent, present-and-empty, present-with-data — behave as before.
- **`delete_storage_first`.** The early return in `ByteCollectionView::try_load_entry` is equivalent to the old `if !delete_storage_first && contains_key`; dropping the `updates` guard doesn't matter on the `None` path.
- **`chunks_exact` vs `chunks_exact_or_repeat`.** Safe *because* `entry_len = 1 + NUM_INIT_KEYS >= 1` — the marker is what makes the `chunks_exact(0)` panic unreachable. Worth knowing that's load-bearing. `chunks_exact_or_repeat` is still used at four other sites, so nothing goes dead.
- **Mixing tags in one batch.** New here: `Index`- and `Subview`-tagged keys now travel in the same `read_multi_values_bytes`. I checked the wrappers that could care — `DualStore` routes per store rather than per key, and `LruCachingStore` uses one shared budget. No layer treats the tags differently.

### Test gap (fixed in `606b03f`)

I instrumented `try_load_entries` in both files to panic when a single storage-backed batch contains both a hit and a miss, then ran the lib tests, all four integration test files, and the doctests. Zero hits. The doctests reach the all-miss branch, and `random_container_tests` only queries indices it knows exist — so a batch with holes *in the middle*, which is exactly what the new chunk-and-zip alignment has to get right, was never exercised.

The commit adds `test_collection_view_try_load_entries_with_absent_entries` and its reentrant twin: entries 1 and 5 hold values, 3 exists but was never written to (index marker only), 0/2/4/6 are absent; the collection is reopened so everything is served from storage, then both the batch and single-entry paths are checked. The probe fires on both tests now, and they have teeth — moving the marker to the end of each entry's key run in `entry_keys` fails them. `cargo test -p linera-views --features metrics` is 377 passing, clippy and nightly `cargo fmt --check` clean, on both branches.

### Nits, all optional

1. `subview_context`, `entry_keys` and `post_load_entry` are byte-for-byte identical across the two files (40 lines). They're associated fns that never touch `self`, so free functions generic over `W: View` in a shared private module would work as-is.
2. `entry_keys` takes a derivable parameter: `subview_context` is a function of `context` and `short_key`, both of which it already receives, so a caller can pass a mismatched pair and get silently wrong keys. `fn entry_keys(context, short_key) -> Result<(W::Context, Vec<Vec<u8>>)>` removes the hazard and shortens all four call sites.
3. `post_load_entry` turns a short read into "absent": the `let Some(..) else { return Ok(None) }` can only fire if the store returned fewer values than keys, and that contract violation currently surfaces as a silently missing entry rather than an error. A `debug_assert` or a `ViewError` reads better.
4. The same two-line comment appears four times; once on the helper, referenced from the call sites, would carry it.
5. The new helper isn't applied to the rest of the file. `wrapped_view` was converted, but `ByteCollectionView::load_entry_mut` still derives the subview context inline twice, and there are ~11 more inline `base_tag_index(KeyTag::Subview ...)` sites plus the older `get_subview_key` method — so this adds a fourth spelling of the same derivation without collapsing the others.

### On the hit rate you couldn't measure

For the two hot **batch** sites it's 100% by construction, and enforced:

- `chain_worker/state.rs` iterates `chain.nonempty_inboxes`, and a `None` is `ChainError::InternalError("Missing inbox for origin ...")`.
- `chain.rs` `load_outboxes` collects into `Option<Vec<_>>` and errors with `CorruptedChainState("Missing outboxes")`.

Both are pure 2→1 wins with no widening cost, so the batch paths don't need a counter — only the single-entry ones do.

Two things the description undersells:

- **The biggest win isn't inboxes/outboxes.** `execution_state_actor.rs` calls `state.users.try_load_entry(&id)` on *every* `ContainsKey`, `ReadValueBytes`, `ReadMultiValuesBytes`, `FindKeysByPrefix`, `FindKeyValuesByPrefix` and `WriteBatch` from an application. `users` is a `ReentrantCollectionView<_, _, KeyValueStoreView<_>>` with `NUM_INIT_KEYS = 1`, so every application storage operation goes from two sequential round trips to one query of two keys. That's a much higher-frequency path than inbox access.
- **Where the miss widens most.** `InboxStateView::NUM_INIT_KEYS` is 5 (three registers, two queues), so a single-entry inbox miss goes from 1 key to 6 — relevant in `get_inbox_next_height`, where a sender asking a receiver that has never heard from it is a genuine miss. `OutboxStateView` is 2, so `get_tip_state_and_outbox_info` goes 1→3. Both are per-chain and infrequent, so nothing to act on; but note that once a batch exceeds 99 keys, ScyllaDB's `read_multi_values_bytes` splits it into concurrent sub-queries under a single semaphore permit.

One operational note for the release: `contains_key` volume will drop sharply and reappear as wider `read_multi_values_bytes` batches, so any dashboard or alert thresholded on `*_contains_key_latency` rates will shift.

PR #6739, opened 2026-08-20
[testnet_conway] Make background certificate sync measurable: widen latency buckets, add volume metrics (#6736)

ma2bd:
Backport of #6736 to `testnet_conway`.

## Motivation

The measurements that justify #6736 were all taken *on this branch*, against the `pm-infra`
node services running `testnet_conway` in production, so the branch that most needs the fix
is this one.

Two blind spots make the chain listener's background sync unmeasurable today.

**The client-side latency histograms top out far below the real values.** They all use
`exponential_bucket_latencies(10_000.0)`, so the last finite bucket is 10 s. Over 7 days on
the pm-infra workers:

| histogram | share of calls above the 10 s bucket | mean |
|---|---|---|
| `find_received_certificates_latency` | **47.4%** | 352 s |
| `synchronize_chain_state_latency` | 7.1% | 2.4 s |
| `process_inbox_latency` | 0.14% | 169 ms |
| `execute_block_latency` | 0.003% | 91 ms |

`find_received_certificates` has a valid p50 (3.7 s) and nothing usable above p52.5 — every
quantile that matters lands in `+Inf`. Only `_sum / _count` is meaningful.

**Nothing records how much work a sync call did.** A 500 s call that fetched 5 certificates
and one that fetched 20,000 are indistinguishable: the chain and certificate counts exist
only as `debug!` fields, which are off in production. On the validator side, a received-log
query is just a `ChainInfoQuery` with `request_received_log_excluding_first_n` set, so it
cannot be separated from the ~57 req/s of ordinary chain info queries the shards serve.

## Proposal

Identical to #6736:

- `find_received_certificates_latency`, `synchronize_chain_state_latency`: 10 s → 10 min
- `process_inbox_latency`, `prepare_chain_latency`: 10 s → 1 min
- `execute_block_latency` unchanged (0.003% overflow)
- new client metrics `linera_find_received_certificates_log_entries`,
  `linera_find_received_certificates_sender_chains`,
  `linera_sender_certificates_discovered_total`, `linera_sender_certificates_missing_total`
- new validator metric `linera_received_log_query_entries`, whose `_count` separates
  received-log queries from every other `ChainInfoQuery`

No behavioural change; all new observations are behind `cfg(with_metrics)`.

## Test Plan

The cherry-pick applied cleanly except for one hunk in `chain_worker/state.rs`: this branch
still reads `self.chain.received_log` where `main` binds a local `chain`. Resolved by keeping
this branch's form and appending the same `observe` call. `client/mod.rs` and
`chain_client/mod.rs` needed no adjustment — #6721 and #6728 already brought
`declare_metrics!` and the metrics CI checks over, so the two branches agree here.

Locally, in the sandbox, on this branch's pinned toolchains (1.86.0 / nightly-2025-04-03):

```
cargo clippy -p linera-core --lib -- -D warnings                      -> exit 0
cargo clippy -p linera-core --lib --features metrics -- -D warnings   -> exit 0
cargo clippy -p linera-core --all-targets -- -D warnings              -> exit 0
cargo +nightly-2025-04-03 fmt --all -- --check                        -> exit 0
scripts/check_metrics_registration.sh                                 -> exit 0
scripts/check_metrics_features.sh                                     -> exit 0
```

CI on #6736 is green (34 pass, 8 skipped, 0 fail).

## Release Plan

`linera_received_log_query_entries` only appears once validators run a build containing this
change; the four client-side metrics only once the `pm-infra` node services do.

## Links

- #6736 — the `main` PR this backports
- #6721, #6728 — the metrics registration and CI-check backports this builds on

### DIFF
--- a/linera-core/src/client/mod.rs
+++ b/linera-core/src/client/mod.rs
@@ -76,9 +76,10 @@ mod validator_trackers;
 #[cfg(with_metrics)]
 pub(crate) mod metrics {
     use linera_base::prometheus_util::{
-        exponential_bucket_latencies, register_histogram_vec, register_int_counter_vec,
+        exponential_bucket_interval, exponential_bucket_latencies, register_histogram_vec,
+        register_int_counter, register_int_counter_vec,
     };
-    use prometheus::{HistogramVec, IntCounterVec};
+    use prometheus::{HistogramVec, IntCounter, IntCounterVec};
 
     linera_base::declare_metrics! {
         pub static PROCESS_INBOX_WITHOUT_PREPARE_LATENCY: HistogramVec =
@@ -86,7 +87,7 @@ pub(crate) mod metrics {
                 "process_inbox_latency",
                 "process_inbox latency",
                 &[],
-                exponential_bucket_latencies(10_000.0),
+                exponential_bucket_latencies(60_000.0),
             );
 
         pub static PREPARE_CHAIN_LATENCY: HistogramVec =
@@ -94,7 +95,7 @@ pub(crate) mod metrics {
                 "prepare_chain_latency",
                 "prepare_chain latency",
                 &[],
-                exponential_bucket_latencies(10_000.0),
+                exponential_bucket_latencies(60_000.0),
             );
 
         pub static SYNCHRONIZE_CHAIN_STATE_LATENCY: HistogramVec =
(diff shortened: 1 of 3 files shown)


linera-core/src/client/mod.rs
```diff
@@ -118,7 +119,35 @@ pub(crate) mod metrics {
                 "find_received_certificates_latency",
                 "find_received_certificates latency",
                 &[],
-                exponential_bucket_latencies(10_000.0),
+                exponential_bucket_latencies(600_000.0),
```
ndr-ds: Same comment as the `main` PR:

Claude looked at the testnet data and says we can bump this up even more. Since it's exponential buckets, it'll add just 2 more buckets so it's cheap
```suggestion
                exponential_bucket_latencies(3_600_000.0),
```

ndr-ds (approved):

PR #6760, opened 2026-08-22
linera-core: push a checkpoint when the catch-up window cannot reach it

ndr-ds:
## Motivation

Block export catches a lagging validator up by replaying blocks from storage, at most
`--block-export-max-catch-up-blocks` (default 200) per round. `send_missing_blocks` skips heights
whose certificates it cannot read — correct for a chain we only receive from, which is stored only
at its message-bearing blocks.

On a **checkpointed** chain that skipping stops being harmless, because a validator can legitimately
not hold its own early history. `execute_block_with_checkpoint_restore` installs state from the
checkpoint's `execution_state_blobs` and acquires only the checkpoint block plus the pre-checkpoint
blocks named in `outbox_block_hashes` — this chain's blocks at heights with unacknowledged outbox
entries (`collect_unfinalized_block_hashes`, reading `self.block_hashes`). It never fetches the rest
of `0..C`.

So a validator that bootstrapped from a checkpoint, acting as an export **sender** to a destination
below `C`, reads heights it does not hold, skips them, sends nothing, and the height the destination
reports never moves. `record_reached` sees no advance, backs the pair off, and the next round
retries the same empty window. It does not converge.

**This is prospective, not a live bug.** Nothing creates checkpoints automatically —
`ChainClient::checkpoint()`'s only non-test caller is `cli/main.rs` — and `testnet_conway` has no
checkpoint machinery at all. The point is to be correct before that changes, not to fix an
outage.

*(An earlier version of this description said pre-checkpoint certificates "may be pruned". That
overstates: pruning is a property the spec permits — `linera-chain/src/proof/checkpoints.rs`
describes a checkpoint as pruning "this chain's *history* — its older blocks" — but nothing in
`linera-storage/src` deletes certificates today. The reachable mechanism is the one above.)*

## Proposal

When the window cannot reach the target, offer the chain's latest checkpoint certificate first. The
destination installs the execution state, its reported height jumps past the checkpoint, and the
remaining heights are sent from there. The receiving side already dispatches on
`(mode, gap, starts_with_checkpoint)` and restores rather than preprocesses; the client's
`RemoteNodeUpdater::push_checkpoint_if_useful` already does this. Only export did not.

**The push is best-effort.** A validator is free to refuse someone else's execution state, so a
rejected push falls through to replaying blocks rather than failing the round. Propagating there
would leave a refusing destination permanently behind — strictly worse than never offering the
checkpoint, since the replay below never runs. Thanks @ma2bd for raising this; the first version
propagated.

**The checkpoint height is announced by the chain's own worker, never read by export.**
`Storage::load_chain` is not a read: it opens the chain partition exclusively, and two live
`ChainStateView`s for one chain are documented as causing "invalid states and data corruption". The
export queue is deliberately not serialized with the chain worker, so it takes the height along the
existing tip announcement, where the worker already holds `self.chain`:

```
state.rs  →  ChainAnnouncement { tip, checkpoint }  →  ChainRecord.checkpoint
          →  spawn_job  →  send_block / send_missing_blocks  →  push_checkpoint_if_useful
```

Both fields merge by maximum, as the tip already did.

`send_confirmed_certificate` also gains a once-only `BlocksNotFound` arm, because a checkpoint push
provokes it: the destination trust-marks the pre-checkpoint sender blocks it lacks and refuses the
restore until it has them. A hash the destination named but we do not hold is reported as **that
destination's** error, not a read failure — `is_local_scoped` matches `ReadCertificatesError` and
halves the queue's *shared* in-flight budget, so mapping it the way the client does would let one
peer throttle export to every other validator while the log blamed our storage.

### Known trade-off

A destination that always refuses checkpoints costs one rejected push per catch-up round. Recording
the refusal would avoid that, but it means per-(chain, destination) state, and `ChainDest` is
allocated per pair — 246 B/chain at the sizes this queue is built for. One extra RPC against a round
that already sends up to 200 certificates seemed the better trade; happy to revisit.

## Test Plan

```
cargo test -p linera-core --lib
cargo clippy --locked --workspace --all-targets --all-features --exclude linera-web -- -D warnings
cargo +nightly fmt
./scripts/check_chain_loads.sh
```

Three new tests, each verified to **fail** against the unfixed code rather than merely pass:

- `test_send_missing_blocks_pushes_a_checkpoint_past_the_window` — 5-block chain checkpointed at
  height 2, destination at 0, window 2. Reaches the tip in one round; without the push it returns
  `BlockHeight(2)`.
- `test_a_destination_refusing_checkpoints_still_replays` — new `FaultType::RejectCheckpoints`
  refuses checkpoint blocks and is honest otherwise. Passes at `BlockHeight(2)` via replay; with the
  push propagating it fails with `refusing to install a checkpoint`. The fault is injected on the
  **full**-certificate path as well as the lite one — the first attempt wired it only into the lite
  path, where export never sends, and the test passed for the wrong reason.
- `a_destinations_missing_block_is_scoped_to_that_destination` — calls the production constructor,
  so reverting the mapping to `ReadCertificatesError` fails it; carries a positive control that a
  genuine read failure is still local, and that the sibling `BlobsNotFound` keeps its own scope.

**What the tests cannot show.** All are `MemoryStorageBuilder`-only, and memory's `clear_journal` is
a no-op, so nothing here exercises the journaling behaviour that made the original `load_chain`
approach unsafe. That is why the fix is structural — export has no `load_chain` call at all — rather
than something a test guards.

Main-only: `testnet_conway` has neither `push_checkpoint_if_useful` nor the restore dispatch, so
there is nothing to backport.

## Release Plan

## Links

- #6639 / #6726 — block export, and its front-port to `main`

### DIFF
--- a/linera-core/src/chain_worker/export.rs
+++ b/linera-core/src/chain_worker/export.rs
@@ -49,7 +49,7 @@ use futures::{stream::FuturesUnordered, FutureExt as _, StreamExt as _};
 #[cfg(with_metrics)]
 use linera_base::prometheus_util::MeasureLatency as _;
 use linera_base::{
-    crypto::ValidatorPublicKey,
+    crypto::{CryptoHash, ValidatorPublicKey},
     data_types::{Blob, BlockHeight, Epoch, TimeDelta, Timestamp},
     identifiers::{BlobId, ChainId, StreamId},
     time::{timer::timeout, Duration},
@@ -405,10 +405,20 @@ impl ProgressMap {
 
 type SharedProgress = Arc<Mutex<ProgressMap>>;
 
-/// The height after each chain's newest announced block. Written on every `export` call before
-/// the queue is tried, so a block the full queue drops still raises the repair target the tick
-/// measures destinations against.
-type SharedTips = Arc<Mutex<HashMap<ChainId, BlockHeight>>>;
+/// What the worker announces about a chain on every `export` call, before the queue is tried, so
+/// a block the full queue drops still raises the repair target the tick measures destinations
+/// against.
+#[derive(Clone, Copy)]
+struct ChainAnnouncement {
+    /// The height after the chain's newest announced block.
+    tip: BlockHeight,
+    /// The chain's latest checkpoint, read from the view the *worker* owns. Export must never
+    /// load the chain itself: `Storage::load_chain` opens the chain partition exclusively, and a
+    /// second live `ChainStateView` racing the worker's is documented as corrupting.
+    checkpoint: Option<BlockHeight>,
+}
+
+type SharedTips = Arc<Mutex<HashMap<ChainId, ChainAnnouncement>>>;
 
 impl BlockExportHandle {
     /// Queues a block for export and returns immediately. A full queue drops the block — never
@@ -420,6 +430,7 @@ impl BlockExportHandle {
         blobs: Vec<CacheArc<Blob>>,
         epoch: Epoch,
         exported_heights: BTreeMap<ValidatorPublicKey, BlockHeight>,
+        latest_checkpoint: Option<BlockHeight>,
     ) {
         // Announced before the queue is tried: a dropped block must still raise the repair
         // target, or a chain whose *last* block was dropped would never be repaired at all.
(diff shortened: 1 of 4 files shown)


ma2bd:
@ndr-ds Note that some validators may not accept checkpoints in the future.

> On a checkpointed chain that is not merely slow — it can stall outright.

how so?

ndr-ds:
> @ndr-ds Note that some validators may not accept checkpoints in the future.
> 
> > On a checkpointed chain that is not merely slow — it can stall outright.
> 
> how so?

That's just Claude's BS, sorry, will edit the comment

ma2bd:
included in https://github.com/linera-io/linera-protocol/pull/6792

PR #6761, opened 2026-08-22
Index each certificate at the height in its own block header

ndr-ds:
## Motivation

`ChainStateView::block_hashes` documents that it returns *"the hashes of all blocks we have in the
given range"* — it drops the heights we do not hold. Two callers zipped that deliberately compacted
list against the **full** list of heights they had requested, and persisted the pairs as a
height→hash index:

```rust
// linera-core/src/updater.rs — and the same shape in linera-service/src/proxy/grpc.rs
let indices: Vec<_> = heights.into_iter().zip(hashes.clone()).collect();
storage.write_certificate_height_indices(chain_id, &indices).await?;
```

`zip` truncates rather than erroring, so a single missing height shifts every later pair:

```
block_hashes([5,6,7,8,9])  ->  [h5, h9]            // 2 entries for 5 requested
zip([5,6,7,8,9], [h5,h9])  ->  [(5,h5), (6,h9)]    // height 6 now maps to block 9's hash
```

**The holes are routine, not exceptional.** `preprocessed_blocks` is a sparse `MapView`, unlike the
dense append-only `confirmed_log`, so the gap can be *interior*. And `send_chain_information` says
outright that this is intended (#4181):

> Only certificates that are actually in our local storage are sent; heights we don't have are
> silently skipped. […] a chain we merely *receive* from is stored only at its message-bearing
> heights, so we push exactly those.

So for a chain we merely receive from, the requested range is expected to be sparse, `all_found` is
false, the fallback runs, and the zip mispairs.

**It does not converge, and it can lock out its own repair.** Take a sender we hold at `{5,6,8,9}`,
queried over `[5..10)`. The old code persists `5→h5, 6→h6, 7→h8, 8→h9, 9→h9`: the row for height 8
is **overwritten** with block 9's hash, destroying the correct `8→h8` that `add_certificate` had
written from the block's own header. Compounding it, the `block_hash_by_height` cache in
`db_storage.rs` was invalidated by neither write path, so a cached bad pair outlived even a correct
rewrite.

**The write is silent.** `zip` truncates without a diagnostic, `write_certificate_height_indices`
validated nothing, and the only related metric (`READ_BLOCK_HASH_BY_HEIGHT_COUNTER`) cannot
distinguish a poisoned row from a healthy one — a poisoned entry actually *raises* the hit rate.

### What this is, and is not, observed to be doing

Stated plainly so reviewers can weigh it: the defect is unambiguous from the code, but I have **no
production evidence that it is currently causing damage.** Over 7 days on `testnet-conway`,
`pm-infra-backend` and `pm-infra-faucet`:

- `Unexpected certificate value` / `Missing certificates for chain` / `Too many certificates
  returned` — **0 hits across 24.8M log lines** (positive control on the same selector returns
  24,274).
- The index itself is hot: ~485M cache and ~377M DB reads (`linera_read_block_hash_by_height`).

**Read that zero carefully — it is weaker than it looks.** `GrpcClient::download_certificates_by_heights`
(`linera-rpc/src/grpc/client.rs:591-635`) dispatches only to `download_raw_certificates_by_heights`,
and its `while !missing.is_empty()` loop removes from `missing` the heights the returned blocks
*actually* have. Against a validator holding mispaired rows, the response is non-empty but carries
blocks at the wrong heights, so `missing` never shrinks, the `received.is_empty()` break never
fires, and there is no progress check or iteration cap. The client **spins** rather than erroring —
and `RemoteNode`'s per-certificate height assertion sits *above* that loop, so it is never reached.
A poisoned row therefore produces a hang, not a log line. The zero above is consistent both with
"no poisoned rows exist" and with "they exist and wedge clients silently"; it cannot distinguish
them.

That loop is a pre-existing weakness independent of this defect — a Byzantine validator can pin a
client the same way — and it is fixed here too, in **4** below, because a fix for "the index can
mispair" that leaves the consumer of that index spinning is not a fix.

The case for merging is correctness: the bad row is **persisted**, nothing removes it, and the read
path cannot tell it from a good one. Another consumer worth naming is `client/mod.rs:1793-1811`,
which emits `ChainAndHeight` using the **requested** height while taking `chain_id` from the
certificate, then drops that height from `still_needed` — so a client can conclude it already holds
a message-bearing block it never downloaded and stop asking for it. Consumers that re-derive the
true height from the block header (`client/mod.rs:1896`, `chain_worker/export.rs:2033-2035`)
self-correct.

### Which endpoints could serve a wrong certificate

Both by-height gRPC endpoints on the proxy, and only those two. Each resolves height→hash through
`RootKey::BlockByHeight` and, before this change, returned whatever certificate the row named
without checking the block's height:

| endpoint | reads | validated before this PR |
|---|---|---|
| `download_certificates_by_heights` | `read_certificates_by_heights` | no |
| `download_raw_certificates_by_heights` | `read_certificates_by_heights_raw` | no |
| `DownloadCertificatesByHeights` (simple transport, `proxy/main.rs`) | asks the shard for hashes, then `read_certificates` **by hash** | n/a — cannot mispair |

The simple-transport path is the pre-optimization shape and is unaffected: it can return *fewer*
certificates than requested, because the shard drops heights it does not hold, but never a wrong
one. **The by-height index is what introduced the exposure** — it replaced "ask the shard, read by
hash" with "trust a persisted height→hash row."

The precondition is a mispaired row, so this cannot happen for a chain we fully track: every
requested height is below the tip, lands in the dense `confirmed_log`, and the fallback that writes
the bad row never runs. It needs a chain we merely receive from.

Local consumers of the same index — the block exporter (`chain_worker/export.rs`), the bridge relay
(`linera-bridge/src/relay/committee.rs`) and `client/mod.rs` — were equally unvalidated, but they
`flatten()` and then take height and contents from the certificate itself rather than from the
height they asked for, so they cannot mispair; at worst they act on an unexpected but internally
consistent block.

**Not the cause of the `CorruptedChainState` seen on one `testnet_conway` chain.** That was two live
`ChainStateView`s racing after a spurious eviction, fixed by #6776 / #6778. Those structures derive
every height from the block header, never from a requested height.

### Why this targets `testnet_conway` and not `main`

Normally we land on `main` and backport, and the one piece of this that applied to `main` did go
that way — the `show-block` panic, fixed by #6771 (merged). The rest cannot: the defect is in code
that has only ever existed on `testnet_conway`.

| ingredient | `main` | `testnet_conway` |
|---|---|---|
| a caller pairing hashes to heights by position | absent | `updater.rs`, `proxy/grpc.rs` |
| persisting that pairing | `write_certificate_height_indices` does not exist | 9 call sites |

`git log upstream/main -S "write_certificate_height_indices"` is **empty** — it was never on `main`;
it arrived on `testnet_conway` via #5233 and #5258. On `main`, `read_certificates_for_heights` reads
the height-keyed index and returns what it found, with no `block_hashes` fallback and no write-back,
so there is nothing to mispair.

## Proposal

Fix it where the pair is persisted, so no caller can get it wrong, and make the persisted index
repair itself — a length check at each call site would leave both the footgun and the existing rows.

**1. `write_certificate_height_indices` takes certificates, and derives each height from the block
header.** The caller no longer supplies heights, so it cannot supply wrong ones. This is the only
fix that reaches the proxy, where a call-site fix is impossible: the shard's response arrives over
the wire with no marker for the heights it dropped, and the requested heights are whatever a third
party put in the request. It takes `&[Arc<ConfirmedBlockCertificate>]` so neither caller deep-clones
— `updater.rs` caches the certificates it just resolved and passes those `Arc`s. A certificate
belonging to another chain is logged and skipped.

**2. `read_certificates_by_heights_raw` rejects and repairs a row whose block is not the one it was
indexed at.** The check lives in the **raw** read because that is what
`download_raw_certificates_by_heights` serves and, per the loop above, the only path a gRPC client
reaches; anywhere else would leave the one endpoint that puts a mispairing on another node's wire
unprotected. It compares **chain and height** — heights collide across chains constantly, so a
height-only check would pass chain B's block at the same height. Each distinct block is decoded once
however many requested heights resolved to it, and the decoded block is cached, so the non-raw path
does not decode it again. A block that will not decode drops the *answer* but not the *row*:
unreadable bytes do not prove the mapping wrong, and one bad block must not fail the batch.

Deletion is **compare-and-delete**: the hash that was validated may have come from the cache, so the
row is re-read and dropped only if it still holds that hash. A blind delete would let a stale cache
destroy a correct row a concurrent writer had just put there. A failure to clean up is logged and
does not discard the certificates already resolved — the row survives, so the next read retries.

**3. The height index is a *correctable* mapping, so it needs cache primitives that can correct it.**
`caches.block_hash_by_height` lives in `ValueCache`, which is built for immutable content-addressed
values: `insert` deduplicates (on a live entry it keeps the cached value and drops the new one) and
`remove` deliberately preserves the weak index, which `get` then uses to resurrect the value. Both
are right for `hash -> block`, where a key's value can never change; both are wrong for a mapping
that must be fixable — `insert` strands the stale hash on every write path, and `remove` lets it come
back. This adds `ValueCache::replace` (overwrites the weak index too) and `ValueCache::invalidate`
(drops both tiers), and moves every `block_hash_by_height` site onto them — including
`write_blobs_and_certificate`, which is where `add_certificate` records a correct row, and where an
`insert` would have let a stale cache entry survive and make the new repair delete that correct row.

**4. `GrpcClient::download_certificates_by_heights` returns only what was asked for.** Its loop
re-requests outstanding heights, removing from `missing` the heights the returned blocks *actually*
have. A validator serving mispaired rows answers with real certificates at heights nobody asked for,
so `missing` never shrank, the existing `received.is_empty()` break never fired, and the identical
request was re-sent forever. It now keeps only certificates on the right chain at a requested height,
stops when a round makes no progress, and ends with an `ensure!` like its sibling
`download_certificates` — so it cannot return a partial or padded result to callers that index
straight into it (`cli/validator.rs`).

**5.** Also carries the `testnet_conway` half of #6771: `linera chain show-block --height <unknown>`
indexed `[0]` into a vector that is empty for a height we do not hold, aborting with
index-out-of-bounds. It now reports `Chain {chain_id} has no block at height {height}`.

Together these close the three sides of the same by-height path: the write can no longer mispair,
the read rejects and repairs a row that already did, and the consumer fails loudly rather than
spinning or answering with the wrong block.

### Deliberately not in this PR

- **`read_certificate_hashes_by_heights` is still unvalidated.** It is the third public consumer of
  the index — `web/@linera/client`'s `blockHashAtHeight` calls it directly — and it returns a hash
  without the block, so validating there means fetching the block and inverting the layering. Worth
  doing, but as its own change.
- **`delete_certificate_height_indices` writes through a raw `store.write_batch`** rather than
  `MultiPartitionBatch`, so index repairs are invisible to the `batch_size` instrumentation covering
  the rest of the storage layer. `MultiPartitionBatch` has no delete helper today.

## Test Plan

`linera-storage/src/db_storage.rs`, at the layer that actually persists the index:

- `test_write_certificate_height_indices_uses_each_blocks_own_height` — a sender whose message
  blocks are at 3, 7 and 9, queried over `[0..10)`. Asserts the **persisted rows** (read directly
  out of `RootKey::BlockByHeight`, past the caches, since it is the rows that outlive the process)
  hold hashes only at 3, 7 and 9. Under the old code the three hashes landed at heights 0, 1 and 2.
- `test_read_certificates_by_heights_deletes_a_mispaired_index_row` — plants a row pointing at a
  block with a different height, asserts the read reports `None` for it, and that the row is gone
  from storage afterwards so the next fallback can rewrite it.
- `test_read_certificates_by_heights_raw_drops_a_mispaired_index_row` — the same, against the raw
  read directly, since that is the one that answers other nodes.
- `test_height_index_rejects_a_foreign_chain_block` — same height, different chain, which only the
  `chain_id` comparison can catch, on both the write and the read side.
- `test_write_certificate_height_indices_invalidates_a_cached_stale_hash` — plants the row the old
  zip would have written, reads it so the stale hash is cached, then corrects the row and asserts
  the corrected hash is what comes back.

Two of these were checked against the unfixed code rather than assumed to be meaningful: dropping
the `chain_id` comparison makes the foreign-chain test fail, and `insert`-style cache maintenance
makes the stale-hash test fail by returning the stale hash. Green on a test that cannot go red
proves nothing.

**Not covered:** the `GrpcClient` changes. `linera-rpc`'s test module has no server harness, so
exercising them means standing up a mock validator that deliberately serves mispaired or padded
responses over gRPC — a larger harness than the fix — and the pre-change failure mode was a hang, so
the test would need a timeout to avoid hanging CI itself. Stating that rather than implying
coverage; the storage tests cover the condition that produces the bad response, not the client's
reaction to it.

```
cargo test -p linera-storage -p linera-cache --lib    # 33 passed, 0 failed
cargo clippy -p linera-storage -p linera-core -p linera-service --all-targets
cargo clippy -p linera-rpc -p linera-cache --all-targets
cargo clippy ... --all-features                       # 0 warnings in every configuration
cargo fmt --all
```

## Release Plan

  This targets `testnet_conway` directly, where the defect lives; see the table under Motivation.
  `main` never had the fallback or `write_certificate_height_indices`, so there is nothing to land
  there first. The one main-applicable piece went out separately as #6771.

## Links

- Storage index affected: `RootKey::BlockByHeight`, written by `write_certificate_height_indices`
  and by `MultiPartitionBatch::add_certificate`.
- The behaviour the holes come from: #4181, described in `send_chain_information`'s doc comment.
- #6771 — the `show-block` panic, fixed on `main`.


ma2bd:
Read through this carefully — the core move is right and I'd merge it after a couple of things. Ranked by what I'd actually want changed.

## 1. The raw read is left unprotected, and it's the path that serves other nodes

`read_certificates_by_heights` now validates that a block really has the height it was indexed at, drops it, and deletes the row. `read_certificates_by_heights_raw` deliberately doesn't — the new doc says so, "because that would mean deserializing it."

Its one non-test caller is the proxy's `download_raw_certificates_by_heights` (`linera-service/src/proxy/grpc.rs:1053`), which does `.flatten()` and then `all_found = raw_certificates_by_height.len() == heights.len()`. A mispaired row is *present*, so it survives the flatten, the count looks right, `all_found` is true, and the proxy serves another node the wrong certificate for that height — with no validation and no self-healing delete.

That's this PR's own argument applied verbatim: "reads must reject such a row and delete it, or the corruption outlives every fix." It holds for the raw path too, and arguably harder there, since that one puts the bad pairing on the wire.

I haven't traced whether the receiving side re-checks the height, so this may well be caught downstream — worth confirming rather than assuming. If it is caught, the asymmetry still deserves a sentence in the raw method's doc saying who is responsible for it.

## 2. `write_certificate_height_indices` re-reads what the caller just read

At `linera-core/src/updater.rs:1005-1011`:

```rust
let certificates = storage.read_certificates(&hashes).await?;
match ResultReadCertificates::new(certificates, hashes.clone()) {
    ResultReadCertificates::Certificates(certs) => {
        storage.write_certificate_height_indices(chain_id, &hashes).await?;
```

and the first thing the new `write_certificate_height_indices` does is `self.read_certificates(hashes).await?`. `certs` is right there in scope and unused for the index. Same shape in the proxy.

Taking `&[ConfirmedBlockCertificate]` instead of `&[CryptoHash]` keeps the property this PR is built on — the height still comes from the block header, so a caller still cannot supply a wrong one — while dropping the second read, and the "not in storage" warn branch with it.

Being fair to it: the certificate cache probably absorbs most of this, so it's a design smell more than a hot-path cost. But it is on the catch-up path, which is exactly when a node is already behind.

## 3. The fix reintroduces the pattern it's fixing, one level down

```rust
for (height, maybe_raw) in heights.iter().zip(raw_certificates) {
```

`read_certificates_by_heights_raw`'s doc promises one element per input height, so this is contractual rather than wrong. But a PR whose entire subject is "`zip` truncates rather than erroring, so a single missing height shifts every later pair" is the one place I'd want that contract enforced instead of assumed — an `ensure!` on the lengths costs nothing.

## Smaller notes

- **A read now writes.** `read_certificates_by_heights` deletes rows. It's failure-tolerant (warn and continue), so it degrades rather than breaks, but components that are otherwise read-only — the block exporter, and the proposer delegate in #6779 — call this and will now issue deletes. Worth naming in the doc.
- **Self-healing is access-triggered.** A poisoned row at a height nobody queries stays put. Harmless while unread, but "outlives every fix" is solved on access, not swept.
- **Cache invalidation race.** Between `write_batch` and `caches.block_hash_by_height.remove`, a reader that fetched the old value pre-write could insert it post-remove, and `ValueCache::insert` deduplicates so it would stick. The new height check catches it on the next read, so it self-heals — nit only.

## What's good

The central move is the right shape: the caller no longer supplies heights, so it cannot supply wrong ones. That fixes the class, not the instance.

The tests are better than most — `test_write_certificate_height_indices_uses_each_blocks_own_height` reads the **persisted rows** rather than going through the cache, with a comment explaining exactly why ("it is the rows that outlive the process"). That is the difference between a test that would have caught this bug and one that would not.

And the "What this is, and is not, observed to be doing" section — stating plainly that there is no production evidence of current damage, with the log counts behind it — is the right way to present a fix for a defect that is unambiguous in code but unobserved in the field.

## One logistical note

This edits `read_certificates_for_heights` in `updater.rs`, which is the call path #6789 restructures (it moves proposal and validated-certificate upload onto `BlockSender`). No textual conflict today, since `main` has no `write_certificate_height_indices` at all — but whoever forward-ports this will be merging against moved code, so worth coordinating on which lands first.

ma2bd:
Reviewed this closely — the diagnosis and the shape of the fix look right to me, but I think the fix is incomplete on both sides of the cache. Grouping by what I'd want addressed before merge.

## Correctness

**1. The other write path still uses `insert` (`linera-storage/src/db_storage.rs:1800`)**

`write_blobs_and_certificate` still does `caches.block_hash_by_height.insert(&(chain, height), hash)`. That's the exact dedup-doesn't-overwrite hazard this PR documents and then fixes only in `write_certificate_height_indices`.

Concretely: a poisoned row `8 -> h9` gets read once, so the cache holds `(chain, 8) -> h9` with a live strong `Arc`. The real block at height 8 then arrives, `add_certificate` writes the correct row `8 -> h8` to disk, and `insert` hits `dedup_insert`, finds the live weak entry, returns `Operation::Abort(existing)` and re-seats `h9`. Every subsequent cache-first read still returns `h9` — and now the new repair in `read_certificates_by_heights_raw` compares `h9`'s height (9) against 8, mismatches, and **deletes the correct row that `add_certificate` just wrote**. The reasoning in the comment at `db_storage.rs:2207` applies verbatim here; this path needs `remove` too.

**2. The repair deletes blind off a cache-first read (`db_storage.rs:2126`)**

`read_certificate_hashes_by_heights` is cache-first, so a stale cache entry sitting over a *correct* DB row makes the loop mismatch and `delete_certificate_height_indices` unconditionally drop the good key. That window exists between `write_batch` and the `remove` loop at :2210, and deterministically via (1). The result is `None` for a height the node actually holds, which costs a chain-info round trip on the proxy path and pushes the height back into `still_needed` in `client/mod.rs:1793`. Suggest making this a compare-and-delete against the hash that was actually validated rather than a blind delete of the key.

**3. `ValueCache::remove` doesn't durably invalidate (`db_storage.rs:2210`)**

`ValueCache::remove` (`linera-cache/src/value_cache.rs:126`) drops only the bounded-cache entry — "The weak index entry is intentionally kept." If any consumer still holds a live `Arc<CryptoHash>` for `(chain, H)`, the next `get` hits tier 2, `weak.upgrade()` succeeds, and `get` re-inserts the stale value into the bounded cache (`value_cache.rs:145`). The stale hash comes back. `test_write_certificate_height_indices_invalidates_a_cached_stale_hash` can't catch this because it only exercises the case where the bounded cache is the sole strong ref. Either `remove` needs to clear the weak index too, or this wants a compare-and-set insert.

**4. Read-side check ignores `chain_id` (`db_storage.rs:2131`)**

The write side now rejects foreign-chain certificates (:2186) but the read validates height only. A row under `RootKey::BlockByHeight(chain_A)` at height H pointing at chain B's block at the same H passes. Heights collide across chains constantly, so I'd expect this to be the most common surviving mispairing. Downstream, `client/mod.rs:1793-1811` builds `ChainAndHeight` with `chain_id` from the certificate and `height` from the request, then drops H from `still_needed` — so the client believes it holds a message-bearing block on chain A that it never downloaded. The block is already decoded at that point, so `header.chain_id == chain_id` is free.

**5. `read_certificate_hashes_by_heights` is left unvalidated (`db_storage.rs:1959`)**

Third public consumer of the same index, no validation and no repair. `web/@linera/client/src/chain/mod.rs:714` (`blockHashAtHeight`) calls it directly and hands the hash to JS, so with a poisoned row it returns `h9` as "the hash of the block at height 8" — and since that path never reaches `read_certificates_by_heights_raw`, the row is never detected or repaired. That weakens the "no migration needed, the first read repairs it" claim for any height only ever queried this way. (`linera-storage/src/lib.rs:231` also doesn't get the doc note its two siblings got.) I think the check belongs where the hash is resolved rather than only in the raw read.

**6. One bad blob fails the whole batch (`db_storage.rs:2131`)**

The raw read never deserialized anything before — that's why it exists and why `download_raw_certificates_by_heights` uses it. Now `confirmed_block_height`'s `?` propagates a BCS decode failure out of `read_certificates_by_heights_raw`, so a single undecodable `confirmed_block` (corruption, or a block written under a different schema version) fails the entire N-height request with `Status::invalid_argument` instead of returning the N-1 healthy certificates. I'd treat a decode failure like a mispaired row: log, `result[index] = None`, continue.

**7. `download_certificates_by_heights` breaks without an `ensure!` (`linera-rpc/src/grpc/client.rs:638`)**

The sibling `download_certificates` (:583) ends with `ensure!(missing_hashes.is_empty(), NodeError::MissingCertificates(..))`; this loop has no equivalent, so after the break it returns `Ok` with a partial result that may include certificates at heights nobody asked for. `RemoteNode` is protected, but `linera-service/src/cli/validator.rs:744` calls this on a raw `ValidatorNode` and then does `certificates[0].inner()` at :749 — against a validator with a mispaired row it silently prints block 9 as the block at height 8, and against one that returns nothing it panics on index 0. That's the same footgun this PR fixes at `linera-service/src/cli/main.rs:1832`. `NodeError::MissingCertificatesByHeights` already exists; using it here would cover all callers instead of one.

## Performance

**8. Every block is now deserialized twice, and not cached on the hot path (`db_storage.rs:2131`)**

`confirmed_block_height` consults only `caches.confirmed_block`, but the raw read populates `caches.certificate_raw` (:1932) and never `confirmed_block`. So on `download_raw_certificates_by_heights` — the ~377M reads/7d path from the PR description — every request now fully decodes every block, discards it uncached, and does it again next time. On the non-raw path it's worse: :2131 decodes and throws away, then `deserialize_and_cache_certificate` (:2425) decodes the same bytes again. Two full block deserializations per certificate on the updater/catch-up path. Having `confirmed_block_height` return the decoded block (or call `cache_confirmed_block`) would fix both.

**9. Proxy fallback deep-clones every certificate (`linera-service/src/proxy/grpc.rs:661`)**

`read_certificates(&hashes).into_iter().flatten().map(CacheArc::unwrap_or_clone)` — the `certificate` cache retains a strong `Arc` for everything it just inserted, so `unwrap_or_clone` is always the clone branch (`linera-cache/src/arc.rs:43`), deep-copying full block bodies just to read `header.height` and `hash()`. Unlike `collect_raw_certificates_by_hashes`, which chunks by 100 and applies a `GrpcMessageLimiter`, this has no chunk size and no limiter, and `heights` is remote-supplied with no cap anywhere in `CertificatesByHeightRequest` handling. Taking `&[CacheArc<ConfirmedBlockCertificate>]` avoids the clone entirely.

**10. Duplicate heights aren't deduplicated (`db_storage.rs:2126`)**

`heights` arrives off the wire unsorted and uncapped. For `heights = [8, 8, 8, ...]` against one mispaired row, `confirmed_block_height` fully decodes the same block once per duplicate and `stale_heights` accumulates the same height N times, producing a delete batch of N identical keys. The `indices: HashMap<CryptoHash, Vec<usize>>` built at :2087 already groups these — one decode per distinct hash.

## Minor

- `db_storage.rs:2164` — the `ensure!(raw_certificates.len() == heights.len())` looks like dead code. `read_certificates_by_heights_raw` is in the same impl, builds `vec![None; heights.len()]` at :2117 and only ever assigns into existing indices; it never pushes, truncates or resizes. There's only one `Storage` impl, so the cross-crate contract the comment describes doesn't exist.
- `db_storage.rs:2451` — `delete_certificate_height_indices` goes through a raw `store.write_batch` while every other mutation in the file goes through `MultiPartitionBatch` + `self.write_batch`, so index repairs are invisible to the `batch_size` instrumentation covering the rest of the storage layer. `MultiPartitionBatch` has no delete helper today; adding one would keep this to a single write path.
- `db_storage.rs:2186` — no test covers the foreign-chain skip, the one branch of the new write path that silently discards input. If every certificate is foreign, `key_values.is_empty()` returns `Ok(())`, indistinguishable from success. Combined with (4), inverting or dropping that comparison would pass the whole suite. `test_certificate` already takes a `chain_id`.
- `linera-rpc/src/grpc/client.rs:639` — the new `warn!` re-records `chain_id` and `address`, both already on the enclosing `#[instrument]` span (:590 skips only `self` and sets `address` explicitly). Only `outstanding` is new.

